PR #5760, opened 2015-04-08
Fix #5672 . Read latest and non-empty condor log for classAd information

### LINKED ISSUE #5672: Problems reading condor logs with PyCondorPlugin
Two parts of the code to fix:
- We have to read the condor logs from the last retry of the job. We should organize by creation date/timestamp rather than by condor job number. 
- Why a condor.x.x.log can be empty? probably some weird case? anyway, the plugin need to handle this.

tsarangi:


tsarangi:
@ticoann, @lucacopa 
have a look and merge this !

lucacopa:
@ticoann @tsarangi looks good to me now

PR #5773, opened 2015-04-13
Changes for Logdb configuration

vkuznet:


src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
```diff
@@ -26,6 +27,10 @@ def __init__(self, **kwargs):
         #this will break all in one test
         self.reqMgr2 = ReqMgr(kwargs.get("reqmgr2_endpoint", None))
         self.previous_state = {}
+        localurl = kwargs.get('logdb_url', '')
```
ticoann: This only used for GQ so there will be no logdb_url

src/python/WMCore/Services/LogDB/LogDB.py
```diff
@@ -6,86 +6,91 @@
...
+            url = url[:-1]
+        else:
+            break
+    return url
+
+def url_db(url):
+    "Split given url into couchdb url and db name"
+    pos = remove_last_slash(url).rfind('/')
+    return remove_last_slash(url[:pos]), remove_last_slash(url[pos+1:])
+
```
ticoann: Maybe, we can use the following function since it is already there.
https://github.com/vkuznet/WMCore/blob/logdb2/src/python/WMCore/Lexicon.py#L554

src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
```diff
@@ -26,6 +27,10 @@ def __init__(self, **kwargs):
...
+        identifier = kwargs.get('hostName', '')
+        centralurl = kwargs.get('central_logdb_url', '')
+        self.logdb = LogDB(localurl, identifier, centralurl, logger=self.logger)
```
ticoann: since ther is no localurl, shouldn't this be 
self.logdb = LogDB(centralurl, identifier, None, logger=self.logger)

src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
```diff
@@ -26,6 +27,10 @@ def __init__(self, **kwargs):
         #this will break all in one test
         self.reqMgr2 = ReqMgr(kwargs.get("reqmgr2_endpoint", None))
         self.previous_state = {}
+        localurl = kwargs.get('logdb_url', '')
+        identifier = kwargs.get('hostName', '')
```
ticoann: do you prefer the  hostName over log_reporter? :smile: 
I can change that but for GQ and reqmgr is not really a host name. 
https://github.com/dmwm/deployment/pull/176/files#diff-0d7567a7c6e08d29a72ba8e06afacc5fR63

PR #5774, opened 2015-04-13
does not allow to auto-approve MSS subscriptions

amaltaro:
quick fix/hack not to auto-approve MSS subscriptions.
For the master branch, it's better if we check it during assignment.


amaltaro:
Updated the code to also flip the auto-approve flag for those subscriptions created in WMBS as well.
I'm applying this patch to submit1 to check things are ok :-) @ticoann , you may want to take a look

amaltaro:
Sorry Seangchan, since the patch is working fine. I just applied it to the agents :-P

PR #5775, opened 2015-04-14
Forbids auto-approval to MSS endpoint

amaltaro:
Don't accept assignment if user provides a MSS endpoint in the "AutoApproveSubscriptionSites" list.
Will take effect only when we deploy a new ReqMgr...
@ticoann , please take a look.


ticoann:
Patch looks good. created reqmgr2 side of validation #5796

PR #5787, opened 2015-04-15
add missing method

ticoann:


ticoann:
ran the test on private vm

PR #5788, opened 2015-04-15
move test under Service_t

ticoann:
@vkuznet, Valentin, I move the LogDB unittest under Service_t for consistency of the directory structure.
Is that OK?


ticoann:
Also I get following failure sometimes. I think it is due to the time stamp getting the same time. So I added the sleep between the call

F.....
FAIL: Test LogDB APIs
  File "/Users/sryu/WorkArea2/WMCore/test/python/WMCore_t/Services_t/LogDB_t/LogDB_t.py", line 69, in test_apis
    self.assertEqual(docs[0]['msg'], 'msg4')
AssertionError: 'msg2' != 'msg4'

Ran 6 tests in 1.029s

FAILED (failures=1)

vkuznet:
ok, both actions seem appropriate to me

PR #5789, opened 2015-04-16
fix summary call to have one id per thread

ticoann:
Don't merge, this is just update for further discussion with Valentin.

### DIFF
--- a/src/couchapps/LogDB/views/requests/map.js
+++ b/src/couchapps/LogDB/views/requests/map.js
@@ -1,5 +1,5 @@
 function(doc) {
     if (doc.request) {
-        emit(doc.request, doc);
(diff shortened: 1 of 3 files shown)


ticoann:
@vkuznet, Hi Valentin,  I think we had a little miscommunication how the summary will be set. This patch is what I thought initially. summary db will have unique id per (request + agent + thread + message_type).
But I might misunderstood what you were initially intend to do. Anyway, we can discuss tomorrow. Thanks

vkuznet:
Seangchan, this patch looks good to me. I don't mind that summary db will have request-agent-thread-messagetype ids.

ticoann:
This breaks the unittest. I will update the unit test separately.

PR #5792, opened 2015-04-16
Add pre-defined LogDB Message types and associative unit test

vkuznet:


ticoann:
Thanks Valentin, patch looks good. merging it.

PR #5797, opened 2015-04-17
change JobPrio to int value

ticoann:
from the report, https://cms-logbook.cern.ch/elog/Workflow+processing/19692

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -855,10 +855,10 @@ class PyCondorPlugin(BasePlugin):
         sd = condor.Schedd()
         if 'taskPriority' in kwargs and 'requestPriority' in kwargs:
             # Do a priority update
-            priority = (int(kwargs['requestPriority']) + int(kwargs['taskPriority'])*self.maxTaskPriority)
+            priority = (int(kwargs['requestPriority']) + int(kwargs['taskPriority'] * self.maxTaskPriority))
             try:
                 sd.edit('WMAgent_JobID =!= "UNDEFINED" && WMAgent_SubTaskName == %s && WMAgent_RequestName == %s'% (classad.quote(str(task)),classad.quote(str(workflow))),
-                        "JobPrio", classad.ExprTree('"%s"'% priority))
+                        "JobPrio", classad.Literal(int(priority)))
             except:
                 msg = "Couldn\'t edit classAd to change job Priority for WMAgent_SubTaskName=%s, WMAgent_RequestName=%s " % (classad.quote(str(task)), classad.quote(str(workflow)))
                 logging.debug(msg)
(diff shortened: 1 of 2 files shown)


ticoann:
@bbockelm, Hi Brian, please review.

amaltaro:
We could cast again priority to integer, but it shouldn't be needed.
Pushing this patch to the production nodes since JobUpdater is winning a lot.
What surprises me is that we did not hit this problem in testbed agents (which are also connected to the global pool...)

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -858,7 +858,7 @@ def updateJobInformation(self, workflow, task, **kwargs):
             priority = (int(kwargs['requestPriority']) + int(kwargs['taskPriority'])*self.maxTaskPriority)
             try:
                 sd.edit('WMAgent_JobID =!= "UNDEFINED" && WMAgent_SubTaskName == %s && WMAgent_RequestName == %s'% (classad.quote(str(task)),classad.quote(str(workflow))),
-                        "JobPrio", classad.ExprTree('"%s"'% priority))
+                        "JobPrio", classad.ExprTree('"%d"'% priority))
```
bbockelm: I think this is wrong.  I think it should be along the lines of:

classad.Literal(int(priority))

The current patch casts the priority to a string.

ticoann:
Thanks Brian, I updated the patch. Could you point me the condor python wrapper api document? I have no idea, I guess it is time for me to learn. :smile:

bbockelm:
Hi,

The python wrapper API docs are here:

http://research.cs.wisc.edu/htcondor/manual/v8.3/6_7Python_Bindings.html

They are in the HTCondor manual format (which is a bit strange way to do python docs, but it is what we have...) and are reasonably complete.  The online python help function is probably the most accurate.

Brian

ticoann:
Thanks Brian

ticoann:
Actually the problem seems to be this line.
https://github.com/dmwm/WMCore/blob/1.0.1/src/python/WMCore/BossAir/Plugins/CondorPlugin.py#L801

``` python
priority = (int(kwargs['requestPriority']) + int(kwargs['taskPriority']) * self.maxTaskPriority)
```

default self.maxTaskPriority is 1e7 which is float and give the priority to float.
And it seems we were using default value as well before. (There is not maxTaskPriority is set)
https://github.com/dmwm/WMCore/blob/1.0.1/etc/WMAgentConfig.py

However, that code hasn't changed between versions either.  I am not sure why it only causing the problem now.

ticoann:
Alan, do you want me to create new patch for 1.0.5 version?

amaltaro:
I think it would be more clear to have this patch in that branch as well.
OR, we could accumulate a few PR and then cherry pick all these changes for a patch2 for example.

PR #5800, opened 2015-04-17
improve force-complete, 1.0.5 version

amaltaro:
1.0.5_wmagent version of #5799

PR #5802, opened 2015-04-17
Shuffle pileup

### LINKED ISSUE #5791: Implement shuffling of PU files
As a follow up of discussions all around and this summary from @chwissing 
"in the last few weeks we have constantly trouble with processing
workflows, because each and every job opens the _same_ pileup file for
discovering the data products. That causes some issues at some sites."

we should shuffle the pileup files to avoid opening always the same one by each job.

Here is my understand of the code and how it works in WMAgent. WMAgent will create a list of PU files available at the site whitelisted (based only on *completed* blocks available there) and this list is then shipped together with the job (in a json file), which is then fed into the PSet.py configuration during runtime in the worker node (such to avoid DBS calls from the WN).
I guess @ericvaandering is taking over from here ;-)

ericvaandering:
Fixes #5791

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -6,6 +6,7 @@ Create a CMSSW PSet suitable for running a WMAgent job.
 """
 
 import os
+import random
 import types
 import socket
 import traceback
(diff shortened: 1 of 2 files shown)


ticoann:
Do we need this patch for 1.0.5 version?

amaltaro:
Yes, I'll run a simple test on Monday and if everything is ok, then I can create a PR against 105_wmagent.

amaltaro:
And it works like a charm shuffling files under 
```
        fileNames = cms.untracked.vstring( ('/store/mc/Summer12/MinBias_TuneZ2star_8TeV-pythia6/GEN-SIM/START50_V13-v3/0002/0E783FD7-E76A-E111-8DF8-0030487F932D.root', 
```
I will push this patch to all the production nodes tomorrow (together with #5800). Thanks guys!

PR #5804, opened 2015-04-18
change JobPrio to int value

ticoann:
1.0.5 version of #5797


ticoann:
This patched already in production agent, merging it.

PR #5806, opened 2015-04-19
Fix listFilesInBlockWithParents since parent lfn is a list

amaltaro:
There were 2 problems in a row:
1. 'f['parent_logical_file_name']' was a list and so cannot be a dict key
2. then a wrong argument name for _getLumiList
   @ticoann please review, though I'm going to push it to the other agents right now :-D

I will create the same patch for 1.0.5_wmagent and will also create a json template to exercise this includeParents part, my fault not to have gotten it during validation :-(


ticoann:
Thanks Alan, it is definitely not your fault. (And you know whose fault it is. :smile:) Sorry about the sloppy patch.

PR #5808, opened 2015-04-19
fix read settings for repack jobs

hufnagel:
Repack jobs have as input a streaming format. There is some ambiguity on the CERN site read settings (lazy-download or not), hard code an override for the repack read settings.


hufnagel:
Opened this by mistake, this is super-seeded by another patch.

hufnagel:
Actually, its better to keep this separate, the other patch has more changes that aren't ready yet.

PR #5812, opened 2015-04-20
add/change merge job disk and wall estimates

hufnagel:
Double disk estimate for merge jobs to take into account lazy-download. Add a wall time estimate for merge jobs assuming 5 min job initialization plus times for actual merge and the double checksum calculation and stageout at the end (at 5MB/s).

### DIFF
--- a/src/python/WMCore/JobSplitting/ParentlessMergeBySize.py
+++ b/src/python/WMCore/JobSplitting/ParentlessMergeBySize.py
@@ -65,7 +65,7 @@ class ParentlessMergeBySize(JobFactory):
 
         return fileGroups
 
-    def createMergeJob(self, mergeableFiles):
+    def createMergeJob(self, mergeFiles):
         """
         _createMergeJob_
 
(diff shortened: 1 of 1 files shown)


amaltaro:
Dirk, just to refresh my mind. lazy-download is only used by T0, right? Is it the one that completely downloads the files to the worker node and only then starts merging them?
According to your algo, it would take ~3500secs to merge 4 GB of data and 8GB of disk, time is probably overcalculated, but I guess not a big deal.

hufnagel:
No, lazy-download is also used by other sites, AFAIK RAL and maybe some T2. I would have taken this into account, but unfortunately this isn't really possible. At job splitting time you don't know if lazy-download is going to be used, this is a runtime decision.

Yes, I know that time estimates are on the high side. With virtual cloud based resources not so much though, local disk isn't that fast for these. I wanted to be conservative here, because over-estimating doesn't cost much (if the job finishes sooner, the pilot can just schedule another), but under-estimating could eventually cause problems (if we ever enable pilot matching to take wall time estimates into account).

hufnagel:
Two potential negative effects of these over-estimates:

Merge job with >10GB of data can't be scheduled on a 20GB slot because they request >20GB data. Is this a valid use case ? These jobs would only be able to run on multicore pilots where they take up multiple slots.

If a majority of jobs in our system over-estimate their wall-time, we'll have problems keeping pilots filled as they approach their end-of-life. There will be lots of jobs that can in principle still run in the pilot, but can't be matched because their wall time estimates are too high. Not a concern for merges, they are fast (even with the high estimate here) and a small fraction of jobs. This is a concern if we expand this to all processing jobs though.

hufnagel:
Actually, just thought of a tweak for the merge disk estimate. Instead of twice input I could use once input plus largest input file. That reflects lazy-download more correctly (it caches file by file).

amaltaro:
That's true, lazy-download is a runtime decision. I completely agree, it's better to over-estimate it than the other way around.

> Merge job with >10GB of data can't be scheduled on a 20GB slot because they request >20GB data. Is this a valid use case ?

It is. But AFAIK the job matching does not take into account the RequestDisk (it only uses it for monitoring and killing the job when it goes beyond the limits). 
In addition to that, it should not be a problem since default merge thresholds are set between 2 and 4GB, so only real exception should hit us here.

hufnagel:
@ticoann , please don't merge. I am going to change this to reduce the lazy-download overhead.

hufnagel:
Tested the latest version, seems to work.

ericvaandering:
This commit broke a bunch of unit tests which probably just need to be updated to new numbers.

PR #5816, opened 2015-04-21
Refactor LogDB to use one back-end and use deterministic ids

vkuznet:
This PR includes code to use single back-end for LogDB. It relies on the following deterministic id schema: for agent based requests the id is generated as request-agent-thread-type, while for user based comments id is generated as request-userdn-thread-type. All documents have standard structure: `{_id:id, data:[{msg:msg, ts:123, type:mtype}, ...]}`. The `data` part is always a list which allows to have consistent structure for messages. The agent-based messages are updated in place, while user based one are growing. The LogDB post API uses LogDBBackend user_update and agent_update APIs based on LogDB identifier (which either set for agent or uses userdn). We'll create LogDB object for every user request. The data in LogDB are minimized to have only msg/ts/mtype dict, while LogDB get API decompose id into request/agent/thread/type and update retrieved documents accordingly such that end-user get the following structure: `{request:bla, identifier:bla, thr: bla, msg:bla, ts:123, mtype:bla}`. I also provided first draft of LogDBReport class. All unit tests are adjusted accordingly.

### DIFF
--- a/src/python/WMCore/Services/LogDB/LogDBBackend.py
+++ b/src/python/WMCore/Services/LogDB/LogDBBackend.py
@@ -5,15 +5,26 @@ LogDBBackend
 Interface to LogDB persistent storage
 """
 
+# syste modules
 import time
 import datetime
+import hashlib
 
+# WMCore modules
 from WMCore.Database.CMSCouch import CouchServer, CouchNotFoundError, CouchConflictError
 from WMCore.Wrappers import JsonWrapper as json
 from WMCore.Services.LogDB.LogDBExceptions import LogDBError
 
 # define full list of supported LogDB types
 LOGDB_MSG_TYPES = ['info', 'error', 'warning', 'comment']
+
+def gen_hash(key):
+    "Generate hash for given key"
+    if  not isinstance(key, basestring):
+        raise NotImplementedError
+    keyhash = hashlib.md5()
+    keyhash.update(key)
+    return keyhash.hexdigest()
 
 def tstamp():
     "Return timestamp with microseconds"
@@ -31,8 +42,9 @@ def clean_entry(doc):
 
 def design_doc():
     """Return basic design document"""
-    rmap = dict(map="function(doc){ if(doc.request) emit(doc.request, doc)}", reduce="_count")
-    tmap = dict(map="function(doc){ if(doc.ts) emit(doc.ts, null)}")
+    rmap = dict(map="function(doc){ if(doc.request) emit(doc.request, null)}",
+            reduce="_count")
+    tmap = dict(map="function(doc){ if(doc.comments) for(i=0;i<doc.comments.length;i++) emit(doc.comments[i].ts, null)}")
     views = dict(requests=rmap, tstamp=tmap)
     doc = dict(_id="_design/LogDB", views=views)
     return doc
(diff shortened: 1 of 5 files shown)


src/python/WMCore/Services/LogDB/LogDBBackend.py
```diff
@@ -86,13 +96,35 @@ def prefix(self, mtype):
             mtype = 'agent-%s' % mtype
         return mtype
 
-    def post(self, request, msg='', mtype="comment"):
-        """Post new entry into LogDB for given request"""
+    def agent_update(self, request, msg='', mtype="comment"):
```
ticoann: default value shouldn't be comment, maybe info, or we don't have to have default value.
``` python
agent_update(self, request, msg='', mtype="info"):
```

src/python/WMCore/Services/LogDB/LogDBBackend.py
```diff
@@ -86,13 +96,35 @@ def prefix(self, mtype):
...
+            # this means document is not exist so we will just insert
+            pass
+        finally:
+            res = self.db.commitOne(doc)
+        return res
+
+    def user_update(self, request, msg, mtype='comment'):
+        """Update user info in LogDB for given request"""
+        data = {"ts":tstamp(), "msg":msg, "type":mtype}
+        doc = {"_id": self.docid(request, mtype), "data": [data]}
```
ticoann: In case we have view by type, it might be better to take the type out from data since type will be only 'comment'.
Or are you thinking about extending to other message type?

```
data = {"ts":tstamp(), "msg":msg}
 doc = {"_id": self.docid(request, mtype), "comments": [data], "type": "comment"}
```

src/python/WMCore/Services/LogDB/LogDBBackend.py
```diff
@@ -86,13 +96,35 @@ def prefix(self, mtype):
...
+        try:
+            exist_doc = self.db.document(doc["_id"])
+            doc["_rev"] = exist_doc["_rev"]
+            doc["data"] += exist_doc["data"]
```
ticoann: if we change to above, (just name change)
``` python
doc["comments"] += exist_doc["comments"]
```

ticoann:
Hi Valentin, Although id contains all the information (request, agent, thread, type), I am not sure whether it is good idea to rely on the id scheme to store the information instead of explicit key value pair. (I know it saves data space). 
If we ever changes id generation scheme (i.e. use hash or some for some reason),  - maybe I am paranoid. :smile: (more realistically - some of the parameters might contain '--'. I put '--' to distinguish the 
between the items. But there is no guarantee tha request name wouldn't contain '--'.

src/python/WMCore/Services/LogDB/LogDBBackend.py
vkuznet: On  0, ticoann <notifications@github.com> wrote:
> > +        data = {"ts":tstamp(), "msg":msg, "type":mtype}
> > +        doc = {"_id": self.docid(request, mtype), "data": [data]}
> > +        try:
> > +            exist_doc = self.db.document(doc["_id"])
> > +            doc["_rev"] = exist_doc["_rev"]
> > +        except CouchNotFoundError:
> > +            # this means document is not exist so we will just insert
> > +            pass
> > +        finally:
> > +            res = self.db.commitOne(doc)
> > +        return res
> > +
> > +    def user_update(self, request, msg, mtype='comment'):
> > +        """Update user info in LogDB for given request"""
> > +        data = {"ts":tstamp(), "msg":msg, "type":mtype}
> > +        doc = {"_id": self.docid(request, mtype), "data": [data]}
> 
> In case we have view by type, it might be better to take the type out from data since type will be only 'comment'.
> Or are you thinking about extending to other message type?
> 
> ```
> data = {"ts":tstamp(), "msg":msg}
>  doc = {"_id": self.docid(request, mtype), "comments": [data], "type": "comment"}
> ```

Hmm, I think we don't need type anymore, since id contains this info. We can
safely remove type from data and let LogDB get API add it to the doc since it
is decomposing id anyway to get request/agent/thr info. Removing type from docs
here will save space of DB. If you agree I'll re-factor code further.

vkuznet:
I think you're correct about being paranoid and it is better to have
request/agent/thread/type part of doc. And, I agree having long keys may lead
to another problem (index can grow), so using a hash is better idea.
I'll change the code accordingly today.

Issue #5817, opened 2015-04-21
JobUpdater crashing with CouchConflictError


amaltaro:
This component is crashing all the time with couch conflict errors, this is the traceback
```
2015-04-20 19:16:03,500:INFO:JobUpdaterPoller:Priorities were synchronized, wait until the next cycle
2015-04-20 19:18:03,688:INFO:JobUpdaterPoller:Synchronizing priorities with ReqMgr...
2015-04-20 19:18:48,555:ERROR:JobUpdaterPoller:Error in JobUpdater:
CouchConflictError - reason: Conflict, data: None result: None

Traceback (most recent call last):
  File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMComponent/JobUpdater/JobUpdaterPoller.py", line 80, in algorithm
    self.synchronizeJobPriority()
  File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMComponent/JobUpdater/JobUpdaterPoller.py", line 137, in synchronizeJobPriority
    self.workqueue.updatePriority(workflow, priorityCache[workflow])
  File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMCore/Services/WorkQueue/WorkQueue.py", line 138, in updatePriority
    self.updateElements(*elementsToUpdate, Priority = priority)
  File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMCore/Services/WorkQueue/WorkQueue.py", line 108, in updateElements
    answer = self.db.makeRequest(uri = thisuri, type = 'PUT')
  File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMCore/Database/CMSCouch.py", line 116, in makeRequest
    getattr(e, "reason", None), data)
[...]

PR #5818, opened 2015-04-21
Do not crash JobUpdater in case of couch issues

amaltaro:
Created a new CouchConflictError exception. If that happens to JobUpdater, just skip these updates to the next cycle. I'll try to run basic tests, but I cannot simulate the exception since we have a single agent in the same team in testbed.
@ticoann , please review.


ticoann:
Thanks Alan, Actually, I think you need to import 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/CMSCouch.py#L1004
instead of creating new Exception class. Isn't that the one is actually raised? Otherwise, patch looks good.

amaltaro:
Oups, I didn't see that exception, sorry. Ok, just updated the patch. Please take one more look at it :-)
If everything is ok, then I'll create it to the master branch as well.

ticoann:
Everything looks fine to me. Thanks,

amaltaro:
Ok, production agents have this patch in now. Thanks

amaltaro:
Fixes #5817 , feel free to merge it

PR #5820, opened 2015-04-21
remove local logdb

ticoann:


ticoann:
@vkuznet, I changed the name "comments" -> "messages".  Initially I thought it will be used only for comment type. but you made it uniform way to other message type (list) - which is nice since it has the same format. I added the sleep in test since it produces exact same timestamps.
Please review

vkuznet:
looks fine to me

Issue #5681, opened 2015-03-13
add new summary page for given request in wmstats


ticoann:
need to define what kind of information need to be displayed (i.e. workflow progress)

drkovalskyi:
Can we either get a separate webpage or just a view in the DB that won't require to load all the data that WMStats reads? 
I think this image shows all that we need to know:
![screen shot 2015-03-13 at 13 39 09](https://cloud.githubusercontent.com/assets/4704624/6640564/8ae254fa-c998-11e4-9352-6f4f28ba2f22.png)
If some information is not trivial to achieve, here is the bare minimum:
- lumi progress
- event progress
- failure rate
- queue injection
- job progress
"Estimated completion" is very useful as well, but if it's hard to get, we can live without.
Thanks,
          Dima

vlimant:
will go with an effort to renew the wm monitoring IMO

PR #5821, opened 2015-04-21
Add temporary access to request information including lastest job.

ticoann:
intermediate step towards #5681 This this grant temporary access to the api through reqmgr2 api. But this need to move to wmstats api when server cherrrypy backend is created.


ticoann:
tested in vm, merging it.

PR #5822, opened 2015-04-22
Handle couch exceptions in JobUpdater

amaltaro:
master version of #5818


ticoann:
Merging it since this only affects WMAgent

PR #5823, opened 2015-04-22
workaround for duplicate alcaharvest output in fjr

hufnagel:
This should never happen (but it has before), but if for some reason there is a difference in processing history for different data in the same run, the AlcaHarvest will write duplicate payload records into the FJR. Ignore all duplicates and only process each file once, no matter how many records there are for it.

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -1049,11 +1049,20 @@ class Report:
             return []
 
         analysisFiles = stepReport.analysis.files
-        result = []
-        for fileNum in range(analysisFiles.fileCount):
-            result.append(getattr(analysisFiles, "file%s" % fileNum))
 
-        return result
+        results = []
+        for fileNum in range(analysisFiles.fileCount):
+            results.append(getattr(analysisFiles, "file%s" % fileNum))
+
+        # filter out duplicates
+        fileNames = []
+        filteredResults = []
+        for result in results:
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann , can you review ? Tested in the normal error-free case, works fine for that. Really difficult to trigger the problem its supposed to avoid, but then, the logic is pretty simple...

ticoann:
Sorry I have a little knowledge of this.
Is it allowed to have duplicate output (generating duplicate output is not the error case)?
Is it only alcaharvest case which shouldn't have duplicate?
Otherwise, isn't it better to remove duplicate earlier?
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/FwkJobReport/Report.py#L1040

hufnagel:
AlcaHarvest is the only job type that looks at analysis output files from CMS. There can be one to many different analysis file records in the AnalysisFiles section. Duplicates is not a valid use case, as these files are all on the same file system, so more than one file with the same name does not make sense. This is a case of a double-reporting that causes problems laters when duplicate file records are passed to Oracle and fail with a unique violation.

hufnagel:
Let me look at removing them earlier... I don't think it matters here, but it would be a bit "cleaner".

hufnagel:
Ok, this works too in the "no duplicate" case. Again, can't really test the "duplicate" case, but the code again is pretty straight-forward. @ticoann , what do you think ?

ticoann:
Thanks Dirk,

PR #5824, opened 2015-04-22
remove crc checksum calculation, fixes #5632

### LINKED ISSUE #5632: Remove calculation of checksums other than adler32
If I understand the CompOps notes from today correctly, we are OK to do this from the site side. @hufnagel anything to add?

hufnagel:
The CRC checksum is not needed by any of our systems, everything uses adler32. Remove it.

Also remove MD5 checksum calculation for logArchives (no idea why that is done) and remove the code and unit test for MD5 checksum calculation as its not used anywhere else.

This does NOT remove all code that supports checksums of a different type than adler32. DBS for instance still supports CRC and MD5 checksums, we just don't set them anymore.

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/Views/HTMLList/WMStats.JobDetailList.js
+++ b/src/couchapps/WMStats/_attachments/js/Views/HTMLList/WMStats.JobDetailList.js
@@ -124,8 +124,7 @@ WMStats.namespace('JobDetailList');
                     htmlstr += jobDoc.output[i].location[j] + " ";
                 }*/
                 htmlstr += "</li>";
-                htmlstr += "<li><b>checksums: adler32:</b> " + jobDoc.output[i].checksums.adler32 + 
-                            ",<b>  cksum:</b> " + jobDoc.output[i].checksums.cksum + "</li>";
+                htmlstr += "<li><b>checksums: adler32:</b> " + jobDoc.output[i].checksums.adler32 + "</li>";
                 htmlstr += "<li><b>size:</b> " + jobDoc.output[i].size + "</li>";
                 htmlstr += "</ul>";
                 htmlstr += "</ul>";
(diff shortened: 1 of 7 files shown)


hufnagel:
@ticoann , @ericvaandering ,  @amaltaro , please review.

Not sure about the couchapp changes. They look obvious, but I don't know where they apply. Local wmagent only or does this need to go somewhere else ?

ericvaandering:
Looks OK to me, but I can't review the WMStats stuff.

ticoann:
Thanks Dirk, I was wondering whether we should clean up dbsbuffer side as well since we won't push to dbs with other checksums. For dbs reader we still should be able to read all types as you said. However not removing from dbs buffer table wouldn't cause the problem.
I will merge this after the production deployment since it contains wmstats changes.

hufnagel:
Lets hold of on this for a while. I have a suspicion this code breaks DBS upload (server checking for the CRC checksum even though its optional). Doing some testing now...

hufnagel:
Ok, confirmed. With this patch, the

'check_sum': '3479246129'

record is not in the file information passed to the DBS insertBlock call and that makes DBS throw a

HTTPError: HTTP Error 400: Invalid data when insert files.

I could make up a fake checksum, but given that DBS supports different checksum types and the CRC checksum *should* be optional, I'd rather have this fixed in DBS. @yuyiguo , any comment on this ?

ericvaandering:
Please open a DBS ticket.

yuyiguo:
This is not something DBS should be fixed. Check sum is required in DBS schema. There are adler32 or MD5 in DBS and these are optional. As I understood, check sum as check_sum in dbs is always required.

Thanks,
Yuyi

ericvaandering:
Let's follow this issue on the DBS ticket: https://github.com/dmwm/DBS/issues/468

PR #5826, opened 2015-04-22
seperate complete workflow and deleting workflow

### LINKED ISSUE #5411: separate workflow completion from cleaning up db.
to check the workflow completion. TaskArchive checks it child workflow (A workflow either using the same input file or output file as input file status). These cause a long delay of a given workflow to be completed some times. We need have different query for marking the subscription finish and cleaning up database.

ticoann:
This need intense testing, Please don't merge. fixes #5411

### DIFF
--- a/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
+++ b/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
@@ -7,7 +7,7 @@ The actual taskArchiver algorithm
 
 Procedure:
 a) Find and marks a finished all newly finished subscriptions
-      This is defined by the Subscriptions.GetAndMarkNewFinishedSubscriptions DAO
+      This is defined by the Subscriptions.MarkNewFinishedSubscriptions DAO
 b) Look for finished workflows as defined in the Workflow.GetFinishedWorkflows DAO
 c) Upload couch summary information
 d) Call WMBS.Subscription.deleteEverything() on all the associated subscriptions
@@ -298,6 +298,7 @@ class TaskArchiverPoller(BaseWorkerThread):
         try:
             self.findAndMarkFinishedSubscriptions()
             self.archiveTasks()
+            self.deleteWorkflowFromWMBSAndDisk()
         except WMException:
             myThread = threading.currentThread()
             if getattr(myThread, 'transaction', False) \
@@ -327,7 +328,7 @@ class TaskArchiverPoller(BaseWorkerThread):
 
         #Get the subscriptions that are now finished and mark them as such
         logging.info("Polling for finished subscriptions")
-        finishedSubscriptions = self.daoFactory(classname = "Subscriptions.GetAndMarkNewFinishedSubscriptions")
+        finishedSubscriptions = self.daoFactory(classname = "Subscriptions.MarkNewFinishedSubscriptions")
         finishedSubscriptions.execute(self.stateID, timeOut = self.timeout)
         logging.info("Finished subscriptions updated")
 
@@ -351,7 +352,7 @@ class TaskArchiverPoller(BaseWorkerThread):
         finishedwfs = finishedWorkflowsDAO.execute()
 
         #Only delete those where the upload and notification succeeded
-        logging.info("Found %d candidate workflows for deletion: %s" % (len(finishedwfs),finishedwfs.keys()))
+        logging.info("Found %d candidate workflows for archiving: %s" % (len(finishedwfs),finishedwfs.keys()))
         # update the completed flag in dbsbuffer_workflow table so blocks can be closed
         # create updateDBSBufferWorkflowComplete DAO
         if len(finishedwfs) ==0:
@@ -374,7 +375,6 @@ class TaskArchiverPoller(BaseWorkerThread):
             logging.error("we will try again when remote couch server comes back\n%s" % str(ex))
         
         if centralCouchAlive:
-            wfsToDelete = {}
             for workflow in finishedwfs:
                 try:
                     #Upload summary to couch
@@ -424,10 +424,9 @@ class TaskArchiverPoller(BaseWorkerThread):
                                 self.reqmgrSvc.updateRequestStatus(workflow, newState)
                             
                         logging.info("status updated to '%s' : %s" % (newState, workflow))
-    
-                    wfsToDelete[workflow] = {"spec" : spec, "workflows": finishedwfs[workflow]["workflows"]}
-    
+        
                 except TaskArchiverPollerException as ex:
+
                     #Something didn't go well when notifying the workqueue, abort!!!
                     logging.error("Something bad happened while archiving tasks.")
                     logging.error(str(ex))
@@ -438,14 +437,10 @@ class TaskArchiverPoller(BaseWorkerThread):
                     msg = "Couldn't upload summary for workflow %s, will try again next time\n" % workflow
                     msg += "Nothing will be deleted until the summary is in couch\n"
                     msg += "Exception message: %s" % str(ex)
-                    print traceback.format_exc()
+                    msg += "\nTraceback: %s" % traceback.format_exc()
                     logging.error(msg)
                     self.sendAlert(3, msg = msg)
                     continue
-
-            logging.info("Time to kill %d workflows." % len(wfsToDelete))
-            self.killWorkflows(wfsToDelete)
-
         return
 
     def notifyWorkQueue(self, subList):
@@ -470,6 +465,47 @@ class TaskArchiverPoller(BaseWorkerThread):
 
         return
 
+    
+    def deleteWorkflowFromWMBSAndDisk(self):
+        #Get the finished workflows, in descending order
+        deletableWorkflowsDAO = self.daoFactory(classname = "Workflow.GetDeletableWorkflows")
+        deletablewfs = deletableWorkflowsDAO.execute()
+
+        #Only delete those where the upload and notification succeeded
+        logging.info("Found %d candidate workflows for deletion: %s" % (len(deletablewfs),deletablewfs.keys()))
+        # update the completed flag in dbsbuffer_workflow table so blocks can be closed
+        # create updateDBSBufferWorkflowComplete DAO
+        if len(deletablewfs) ==0:
+            return
+        safeStatesToDelete = ["completed", "aborted-completed", "rejected", 
+                              "normal-archived", "aborted-archived", "rejected-archived"]
+        wfsToDelete = {}
+        for workflow in deletablewfs:
+            try:
+                spec = retrieveWMSpec(wmWorkloadURL = deletablewfs[workflow]["spec"])
+                if not spec:
+                    raise Exception(msg = "Couldn't load spec for %s" % workflow)
+                # check workfow is alredy completed or archived (which means
+                
+                #This is used both tier0 and normal agent case
+                result = self.centralCouchDBWriter.getStatusAndTypeByRequest(workflow)
+                wfStatus = result[workflow][0]
+                if wfStatus in safeStatesToDelete:
+                    wfsToDelete[workflow] = {"spec" : spec, "workflows": deletablewfs[workflow]["workflows"]}
+                else:
+                    logging.error("%s is in %s, will be deleted later" % (workflow, wfStatus))
+            
+            except Exception, ex:
+                #Something didn't go well on couch, abort!!!
+                msg = "Couldn't delete %s " % workflow
+                msg += "Exception message: %s" % str(ex)
+                msg += "\nTraceback: %s" % traceback.format_exc()
+                logging.error(msg)
+                continue
+
+        logging.info("Time to kill %d workflows." % len(wfsToDelete))
+        self.killWorkflows(wfsToDelete)
+        
     def killWorkflows(self, workflows):
         """
         _killWorkflows_
--- a/src/python/WMCore/WMBS/MySQL/Subscriptions/MarkNewFinishedSubscriptions.py
+++ b/src/python/WMCore/WMBS/MySQL/Subscriptions/MarkNewFinishedSubscriptions.py
@@ -1,15 +1,15 @@
 #!/usr/bin/env python
 """
-_GetAndMarkNewFinishedSubscriptions_
+_MarkNewFinishedSubscriptions_
 
-MySQL implementation of Subscriptions.GetAndMarkNewFinishedSubscriptions
+MySQL implementation of SubscriptionsMarkNewFinishedSubscriptions
 """
 
 import time
 
 from WMCore.Database.DBFormatter import DBFormatter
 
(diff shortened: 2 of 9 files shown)


ticoann:
@amaltaro, Alan,  Please review.

src/python/WMCore/WMBS/MySQL/Subscriptions/MarkNewFinishedSubscriptions.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+"""
+_GetAndMarkNewFinishedSubscriptions_
```
amaltaro: Remote "GetAnd". Same for 2 lines below.

src/python/WMCore/WMBS/MySQL/Workflow/GetDeletableWorkflows.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
+"""
+_GetDeletableWorkflows_
+
+MySQL implementation of Workflows.GetFinishedWorkflows
```
amaltaro: Fix class name here (and maybe date+author below xD)

src/python/WMCore/WMBS/MySQL/Workflow/GetFinishedWorkflows.py
```diff
@@ -14,9 +14,7 @@
 class GetFinishedWorkflows(DBFormatter):
     """
     A workflow is finished when all subscriptions associated to it are finished
-    and all the other workflows using the same spec are finished as well,
-    this DAO searches such workflows. It assumes that a spec is not shared by
-    two workflows of the same name.
+    It assumes that a spec is not shared by two workflows of the same name.
```
amaltaro: Is it possible to have two workflows with the same name?

src/python/WMCore/WMBS/MySQL/Workflow/GetDeletableWorkflows.py
```diff
@@ -0,0 +1,81 @@
...
+                                      LEFT OUTER JOIN wmbs_file_parent ON
+                                         wmbs_file_parent.parent = wmbs_fileset_files.fileid
+                                      LEFT OUTER JOIN wmbs_fileset_files child_fileset ON
+                                         child_fileset.fileid = wmbs_file_parent.child
+                                      LEFT OUTER JOIN wmbs_subscription child_subscription ON
+                                         child_subscription.fileset = child_fileset.fileset AND
+                                         child_subscription.finished = 0
+                                      LEFT OUTER JOIN wmbs_workflow child_workflow ON
+                                         child_subscription.workflow = child_workflow.id AND
+                                         child_workflow.name != wmbs_subscription.name
```
amaltaro: There is no name column in wmbs_subscription

src/python/WMCore/WMBS/MySQL/Workflow/GetDeletableWorkflows.py
```diff
@@ -0,0 +1,81 @@
...
+                                  GROUP BY wmbs_workflow.name
+                                  HAVING COUNT(wmbs_subscription.workflow) = 0  AND COUNT(child_workflow.name) = 0
```
amaltaro: This "COUNT(child_workflow.name) = 0" only evaluates to true when the workflow has been completely deleted from WMBS, which will never be the case here (I guess killWorkflow() does that job), right?

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -470,6 +463,47 @@ def notifyWorkQueue(self, subList):
 
         return
 
+    
+    def deleteWorkflowFromWMBSAndDisk(self):
```
amaltaro: Don't you need to call this method in the algorithm()? Otherwise it won't run ever

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -470,6 +463,47 @@ def notifyWorkQueue(self, subList):
...
+                spec = retrieveWMSpec(wmWorkloadURL = deletablewfs[workflow]["spec"])
+                if not spec:
+                    raise Exception(msg = "Couldn't load spec for %s" % workflow)
+                # check workfow is alredy completed or archived (which means
+                
+                #This is used both tier0 and normal agent case
+                result = self.centralCouchDBWriter.getStatusAndTypeByRequest(workflow)
+                wfStatus = result[workflow][0]
+                if wfStatus in safeStatesToDelete:
+                    wfsToDelete[workflow] = {"spec" : spec, "workflows": deletablewfs[workflow]["workflows"]}
```
amaltaro: For clarity, can we replace "deletablewfs[workflow]["workflows"]" by "workflow["workflows"]"?

src/python/WMCore/WMBS/Oracle/Subscriptions/MarkNewFinishedSubscriptions.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+_GetAndMarkNewFinishedSubscriptions_
+
+Oracle implementation of Subscription.GetAndMarkNewFinishedSubscriptions
+"""
+
+from WMCore.WMBS.MySQL.Subscriptions.GetAndMarkNewFinishedSubscriptions \
+        import GetAndMarkNewFinishedSubscriptions as MySQLGetAndMarkNewFinishedSubscriptions
+
+class GetAndMarkNewFinishedSubscriptions(MySQLGetAndMarkNewFinishedSubscriptions):
```
amaltaro: Rename class, import and fix docString to "MarkNewFinishedSubscriptions"

src/python/WMCore/WMBS/Oracle/Workflow/GetDeletableWorkflows.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+"""
+_GetDeletableWorkflows_
+
+Oracle implementation of Workflow.GetFinishedWorkflows
```
amaltaro: Fix class name here

amaltaro:
In addition to those quick comments, a few more general:
 1. Remove 'Subscriptions.GetAndMarkNewFinishedSubscriptions' DAO since it is/was only called here (also update tests).
 2. (GetAnd)MarkNewFinishedSubscriptions was the DAO I'd like to improve, it's super hard to understand IMO and we have no idea whether it updated anything or not in the end...
 3. Why "subWithUnfinishedJobSQL" query is  so much different between the oracle and mysql version? I know this code was already like that, but ...

 4. I think the initial goal is to get this feature in, then we can start improving the code

Adding @yuyiguo to evaluate these SQL queries, since this code looks to be quite more expensive now (and I'm not a DB expert at all :D)

src/python/WMCore/WMBS/MySQL/Workflow/GetFinishedWorkflows.py
ticoann: No, but I removed it since it is obvious.

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
ticoann: Sorry, I didn't follow this. deletablewfs is dict with {workflow_name: {'workflows': {workflow id: [....]}
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMBS/MySQL/Workflow/GetFinishedWorkflows.py#L60

I didn't want to change the return format since it gets feed to other method.

ticoann:
> In addition to those quick comments, a few more general:
> 
>     Remove 'Subscriptions.GetAndMarkNewFinishedSubscriptions' DAO since it is/was only called here (also update tests).

Yes I was planing to do that. 

>     (GetAnd)MarkNewFinishedSubscriptions was the DAO I'd like to improve, it's super hard to understand IMO and we have no idea whether it updated anything or not in the end...

That was improved version :-) It was dead slow before. I agree that this still need to be improved for understanding the code (I would break down to few calls instead of making one gigantic call. But I will do in separate issue after checking simpler modification works first.

> 
>     Why "subWithUnfinishedJobSQL" query is so much different between the oracle and mysql version? I know this code was already like that, but ...

Because of the performance reason. Oracle support 'with' syntax and mysql is not so we would like to improve the performance in Oracle a little more. Since the reason rewriting this query was due to the performance.

> 
>     I think the initial goal is to get this feature in, then we can start improving the code
Yes, I agree.

> Adding @yuyiguo to evaluate these SQL queries, since this code looks to be quite more expensive now (and I'm not a DB expert at all :D)
I was planning to bother Yuyi anyway. I will rewrite GetDeletableWorkflows sql first.

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
amaltaro: You don't need to change the return format. In the line 481 you are iterating over this deletablewfs, that's why it looks - to me ;-) - more clear if you only reference only workflow['workflows']. Anyways that makes no difference in the end, so you can skip it

ticoann: Hi Alan, workflow is string, workflow['workflows'] will give the syntax error. Or am I misunderstanding something?

amaltaro: da~~~, you're right! Sorry for the noise :(

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
ticoann: Thanks, I added in algorithm()

ticoann:
@amaltaro, I updated the patch, Could you check it again? I ran the query for syntax error but I can't really test against production since production agent doesn't have the same meaning of finished workflow any more. (In the patch finished means only the current subscription is finished not including child subscription). I will set up the test and run the workflows for this. 
@yuyiguo, Could you review the sql part? 

I will update the unittest in separate commit.

ticoann:
@yuyiguo, as you suggested I updated the query. I ran some test but in terms of the performance there was not difference  - however the data size wasn't so big. Anyway, it is easier to understand query with your suggestion. Thanks for the review.

amaltaro:
Sorry Seangchan. I'll review it on Monday, hope you can wait :-)

ticoann:
Thanks Alan, not urgent.

ticoann:
Basic test (at least closes workflow properly and delete the database) is performed and updated the bug fix. However, it need to be tested that actually workflow with child workflow is run. (Merging it first and do further test)

hufnagel:
Ok, this breaks the Tier0, TaskArchiver dies with

2015-05-07 17:41:02,374:INFO:TaskArchiverPoller:Polling for finished subscriptions
2015-05-07 17:41:02,641:ERROR:BaseWorkerThread:Error in worker algorithm (1):
Backtrace:
  <WMComponent.TaskArchiver.TaskArchiverPoller.TaskArchiverPoller instance at 0x1e58f38> TaskArchiverPollerException
Message: Caught exception in TaskArchiver
(DatabaseError) ORA-00904: "CHILD_WORKFLOW"."NAME": invalid identifier
 ' UPDATE wmbs_subscription\n              SET wmbs_subscription.finished = 1, wmbs_subscription.last_update = :timestamp\n              WHERE wmbs_subscription.id IN ( \n                    \n                   WITH\n                    complete_subscription\n                   AS\n                    (SELECT distinct wmbs_subscription.id,\n                                    wmbs_subscription.fileset,\n                                    wmbs_workflow.name\n                        FROM wmbs_subscription\n                        INNER JOIN wmbs_fileset ON\n                            wmbs_fileset.id = wmbs_subscription.fileset AND\n                            wmbs_fileset.open = 0\n                        INNER JOIN wmbs_workflow ON\n                            wmbs_workflow.id = wmbs_subscription.workflow AND\n                            wmbs_workflow.injected = 1\n                        LEFT OUTER JOIN wmbs_sub_files_available ON\n                            wmbs_sub_files_available.subscription = wmbs_subscription.id\n                        LEFT OUTER JOIN wmbs_sub_files_acquired ON\n                            wmbs_sub_files_acquired.subscription = wmbs_subscription.id\n                      WHERE wmbs_subscription.finished = 0 AND\n                            wmbs_sub_files_available.subscription is Null AND\n                            wmbs_sub_files_acquired.subscription is Null\n                     )\n                    \n                    \n                    SELECT complete_subscription.id\n                        FROM complete_subscription \n                    WHERE complete_subscription.id\n                        NOT IN (\n                          \n                    SELECT complete_subscription.id \n                          FROM complete_subscription\n                          INNER JOIN wmbs_jobgroup ON\n                            wmbs_jobgroup.subscription = complete_subscription.id\n                          INNER JOIN wmbs_job ON\n                            wmbs_job.jobgroup = wmbs_jobgroup.id AND\n                            wmbs_job.state_time > :maxTime AND\n                            wmbs_job.state != 4\n                     \n                        )\n                    GROUP BY complete_subscription.id\n                    HAVING COUNT(child_workflow.name) = 0) ' {'timestamp': 1431013262, 'maxTime': 1431009662}

Let me see if I can fix it...

hufnagel:
If you have invasive large patches like this, can we run them in the Tier0 before merging them ? Tier0 has probably the deepest workflow stack for any requests in the system.

hufnagel:
I can't make heads or tails of this. @amaltaro , can you take a look ? Also, why do we have two versions of MarkNewFinishedSubscriptions ? Can't we unify the MySQL and Oracle versions ?

hufnagel:
Btw, I see that the Oracle version of MarkNewFinishedSubscriptions is obviously wrong, looks almost like some copy&paste mishap. This never could have worked. No idea how exactly to fix it though.

ticoann:
Sorry Dirk It is my fault, I keep forgetting about Tier0 is using head directly and I only tested on MySQL agent The reason it is different from MySQL is for the performance reason. Anyway, I make the two query the same for now.
https://github.com/dmwm/WMCore/pull/5869
Please let me know if you had still problem.

hufnagel:
I switched my tests back to use 1.0.6 plus the checksum patch, that worked. @amaltaro , we should review if the performance argument for having separate MySQL/Oracle queries still applies. Separate queries are a pain to maintain (as shown by this mishap where only one query was tested).

ticoann:
Hi Dirk, yes I think you are right. It is pain to maintain two different queries especially with this complicate sql. If the performance gain is not that much, we should keep them same. I will try to run some test.
Besides, it is not Alan's fault. It is my fault (Seangchan) this hassle. Sorry about that.

amaltaro:
sigh... I forgot to patch the testbed agent to validate this fix. So it won't be available tomorrow since it requires careful validation.

ticoann:
FYI, Alan, this is just agent changes. So no need to validate by HG1506 dead line.

Issue #5767, opened 2015-04-10
ACDC EventBased is also broken


amaltaro:
For my record :), Julian mentioned ACDCs using EventBased job splitting are creating duplicate lumis. It seems it also creates less jobs than what it should. I'll investigate it

amaltaro:
It seems we need to change how we upload information to the ACDC server, if we want to make it to work for MC from scratch. This is the - important - information currently uploaded:
 fileset['events'] which is the same value as 'RequestNumEvents'. Instead, it should contain EventsPerJob for this job.
 fileset['first_event'] which is always 1.
 fileset['last_event'] which is always 0.
 fileset['runs'] which also contains a list of lumis. Since this can change, it seems useless to me.

If we properly upload the first and last event, then it should be enough to recover the correct events.

amaltaro:
Ok, I did some more progress on this (btw, feel free to assign it to me and don't bother reading it for now :)).

Here is the place that we get the right information about first/last things:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py#L193
e.g.:
```
{'LastRun': 1L, 'FirstRun': 1L, 'inclusivemask': True, 'runAndLumis': {1L: [[529L, 533L]]}, 'LastEvent': 133001L, 'FirstEvent': 132001L, 'LastLumi': 533L, 'FirstLumi': 529L}
```

Then it gets inside this if
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ACDC/CouchFileset.py#L106

which uses a "wrong" info from oracle:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py#L354
e.g.:
```
>>> tmpJob['input_files']
[...]

PR #5831, opened 2015-04-23
Correctly upload and create jobs for ACDC EventBased

amaltaro:
Fixing 3 things:
1. Fix for MC from scratch
2. Fixes when there is input dataset with EventBased (note: one cannot run the original workflow in EventBased and run ACDC in LumiBased, vice-versa. It only works if the same job splitting is kept.
3. Do not create ACDC docs for cleanup and logcollect jobs

Don't merge yet. I need to run some LumiBased wfs just to make sure it does not brake anything.

### DIFF
--- a/src/python/WMCore/ACDC/CouchFileset.py
+++ b/src/python/WMCore/ACDC/CouchFileset.py
@@ -99,9 +99,22 @@ class CouchFileset(Fileset):
         _add_
 
         Add files to this fileset
+
+        Note: if job was lumi based splitted, then we do not have
+        reliable events information. If job was event based splitted,
+        then we do not have reliable lumi information.
         """
         filteredFiles = []
         if mask:
+            for f in files:
+                # There is no LastEvent for last job of a file
+                if mask['LastEvent'] and mask['FirstEvent']:
+                    f['events'] = mask['LastEvent'] - mask['FirstEvent']
+                    f['first_event'] = mask['FirstEvent']
+                elif mask['FirstEvent']:
+                    f['events'] -= mask['FirstEvent']
+                    f['first_event'] = mask['FirstEvent']
+
             maskLumis = mask.getRunAndLumis()
             if maskLumis != {}:
                 # Then we actually have to do something
@@ -111,6 +124,7 @@ class CouchFileset(Fileset):
                         f['runs'] = newRuns
                         filteredFiles.append(f)
             else:
+                # Likely real data with EventBased splitting
                 filteredFiles = files
         else:
             filteredFiles = files
--- a/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
+++ b/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
@@ -131,6 +131,9 @@ class ErrorHandlerPoller(BaseWorkerThread):
         for job in jobList:
             job.failInputFiles()
 
+        # Do not build ACDC for utilitarian job types
+        jobList = [ job for job in jobList if job['type'] not in ['LogCollect','Cleanup'] ]
+
         self.handleACDC(jobList)
 
         return
--- a/src/python/WMCore/JobSplitting/EventBased.py
+++ b/src/python/WMCore/JobSplitting/EventBased.py
@@ -55,7 +55,7 @@ class EventBased(JobFactory):
                 logging.error(msg)
                 return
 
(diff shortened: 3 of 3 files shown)


amaltaro:
Just found out that this line
https://github.com/dmwm/WMCore/pull/5831/files#diff-92e2c20e60f6397a2a6930a7d09717a4R60

breaks this patch when there is input dataset and the same file in different acdc docs.
Maybe there is no need at all to call that method (the location is available in the acdc doc 'locations').

I'll fix this patch this weekend or on Monday, don't merge please.

amaltaro:
Just updated the patch, hopefully the last change. I basically isolated acdc job creation to that specific method.
Since it touches base code, I'll run some integration workflows over the night.

amaltaro:
Ok, it passes all my tests. @ticoann please review. If it's ok, I'll push it right away to the production agents.

ticoann:
Alan, this is just for agent update, right?

amaltaro:
That's right, we do not need to update central services. BTW, it fixes #5767 
If it's ok, then I'll create a master version tomorrow.

src/python/WMCore/ACDC/CouchFileset.py
```diff
@@ -99,9 +99,22 @@ def add(self, files, mask = None):
...
+        Note: if job was lumi based splitted, then we do not have
+        reliable events information. If job was event based splitted,
+        then we do not have reliable lumi information.
         """
         filteredFiles = []
         if mask:
+            for f in files:
+                # There is no LastEvent for last job of a file
+                if mask['LastEvent'] and mask['FirstEvent']:
+                    f['events'] = mask['LastEvent'] - mask['FirstEvent']
```
ticoann: Is mask inclusive? (I am not sure but just checking). In that case, shouldn't be 
 f['events'] = mask['LastEvent'] - mask['FirstEvent'] + 1

amaltaro: Yes, inclusivemask was True for all my tests. Good catch!
Let me update this code.

amaltaro: Actually, this code is only used to create the acdc docs (by ErrorHandler) and it's used to calculate the number of events we need to recover, example from my logs:
```
mask {'LastRun': 1L, 'FirstRun': 1L, 'inclusivemask': True, 'runAndLumis': {1L: [[649L, 653L]]}, 'LastEvent': 163001L, 'FirstEvent': 162001L, 'LastLumi': 653L, 'FirstLumi': 649L} for job id 527
```
So if we ( + 1), in the end we will be producing 1 event more than what it should. Does it make sense for you?

ticoann: What is the event to recover in the example?
Are those events inclusive here? 
'LastEvent': 163001L, 'FirstEvent': 162001L,
In that case, shouldn't +1?

In case of below,  event to be recovered is 1 or 0?
'LastEvent': 1, 'FirstEvent': 1,

amaltaro: The events to recover are 162001,162002...163000 (AFAIK the lastevent is excluded).
> In case of below, event to be recovered is 1 or 0?
> 'LastEvent': 1, 'FirstEvent': 1,

I think it's not a valid case, since it would mean EventsPerJob = 0. It would recover 0 events.

I'm pretty sure this patch is Ok, but let me retrieve a few logArchives and double check we are recovering the correct events.

amaltaro: I was investigating one of my test cases where we had 2 failures (500 evts each), the ACDC recovered exactly the same events that failed in the original one. You can see some of the logs here:
http://amaltaro.web.cern.ch/amaltaro/forSeangChan/GH_5831.log

ticoann: It seems above example is when last event wasn't specified (0). So it doesn't fall to that condition statement. Is there a case with last even is specified? But it you think +1 is not needed, we can merge this.

amaltaro: We don't care uploading the last event to couch since we have already firstevent and # of events.
In this example lastevent was not in couch, but was used to build the adcd docs, the only case it's not available is for the last job of the workflow (or perhaps last job of the fileset)... I still think we don't need +1 here :-)

PR #5833, opened 2015-04-24
Change CustodialSubType dafault to Replica instead of Move

amaltaro:
Workflow team was sometimes not setting this argument, and then it was defaulting to Move and making a mess for multi-steps/multi-tasks workflows.

@ticoann review. In principle we skip it only for the other cycle (unless we need to fix something still this week :-))


ticoann:
There need is some fix need to get in for reqmgr2. So we will include this one.

PR #5836, opened 2015-04-26
simplify how CMSSW is called

hufnagel:
The CMSSW executor class is calling subprocess in a complex and strange way, simplify it. Tested in Tier0 replays, works fine. An earlier version of this with most of the changes (removing the while loop) has been exercised in Tier0 replays for months now).


PerilousApricot:
+1 I'm the "AMM" in the comment. I still don't know what that block is
from. Only thing I can think is that subprocess was less fully-featured in
older versions of Python.

-- 
Sent from Gmail Mobile

PerilousApricot:
Also, that's from 5 years ago, I feel old.

-- 
Sent from Gmail Mobile

ericvaandering:
Dealing with subprocesses seems to have been reworked multiple times in earlier pythons but now seems stable.

PR #5842, opened 2015-04-29
Fix except to use python 2.6 recommendation

ericvaandering:
Update our use of exceptions to python 2.6 standard. Tested with all unit tests, checks out OK.


hufnagel:
I guess I should do the same in the T0 code base ? Of if you have a script to do so please feel free to do the same there...

ericvaandering:
I can do it easily. I can't test it, though. :-)

This is super easy though. 

easy_install futurize
futurize -f except . (or similar)

hufnagel:
That's fine, if you make a pull request I (or the operators) can test it in the next replay.

PR #5851, opened 2015-04-30
Only change jobPrio when it has a diff prio

amaltaro:
Quick patch provided by Brian. @bbockelm @ticoann iplease review. We need to push it to production asap.

There is still one thing we have to fix, it seems we first submit jobs to condor queue and only then we change its prio (creating one condor_qedit more). But this will have to wait.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/CondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/CondorPlugin.py
@@ -803,16 +803,17 @@ class CondorPlugin(BasePlugin):
         if 'taskPriority' in kwargs and 'requestPriority' in kwargs:
             # Do a priority update
             priority = (int(kwargs['requestPriority']) + int(kwargs['taskPriority'] * self.maxTaskPriority))
-            command = 'condor_qedit -constraint \'WMAgent_SubTaskName == "%s" && WMAgent_RequestName == "%s"\' ' %(task, workflow)
-            command += 'JobPrio %d' % priority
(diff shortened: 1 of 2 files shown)


bbockelm:
There's a second condor_qedit below this that also needs to be changed.  Can you update it?

ericvaandering:
Are we using condor plugin or pycondor plugin?

amaltaro:
@bbockelm code updated.
@ericvaandering no, but move to pycondor is my plan for the next wmagent production release (beginning of june)

amaltaro:
Eric, trying to answer you again :-) we are still using condor plugin. It seems pycondor is much more stable now (I know T0 guys ran a few replays using it).

ericvaandering:
So do we need to make sure the same sort of change gets into PyCondorPlugin or does it not suffer from the problem?

bbockelm:
Yup - same sort of change needs to be ported.

ticoann:
I guess it was causing the problem when it try to update the priority with same priority, right?
Since I don't know the syntax, if Brian said it looks OK, it is good to go. 
Do you want me to updated the production agent? You can include the PyCondor version on master patch. Thanks,

bbockelm:
Yeah - even if the classad is unchanged, it apparently sends an "update pending" message which is causing a problem.

amaltaro:
PyCondor has been updated as well. I guess we can now go ahead and patch the production agents (?)

amaltaro:
Pushed to production. I'll create the master version tomorrow or on Monday :-)

ticoann:
Thanks Alan, merging it since this is against WMAgent version. I will create the master version.

ticoann:
#5852 is created

amaltaro:
:+1:

PR #5854, opened 2015-05-01
remove unnecessary views from wmagent wmstats view

### LINKED ISSUE #4526: change wmstats agent side deploy script to only push necessary view

ticoann:
fixes #4526. @amaltaro, Alan please review. (Don't merge, need to be tested)

### DIFF
--- a/src/couchapps/WMStatsAgent/validate_doc_update.js
+++ b/src/couchapps/WMStatsAgent/validate_doc_update.js
@@ -1,0 +1,17 @@
+function(newDoc, oldDoc, userCtx) {
+   // Determines the doc operation type
+   var DOCOPS = { modif:0, creat:1, delet:2 };
+   var docOp = oldDoc ? (newDoc._deleted === true ? DOCOPS.delet : DOCOPS.modif)
+                      : DOCOPS.creat;
+
+   if (newDoc._deleted === true && !oldDoc) {
+     throw({forbidden: 'Do not create deleted docs'});
+   }
+
+   var allowed = (userCtx.name === null);
+   //---------------------------------
+     // Throw if user not validated
+   if(!allowed) {
+      throw {forbidden: "User not authorized for action. only local update allowed"};
+   }
+};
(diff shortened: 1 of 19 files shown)


amaltaro:
Just to make sure I understood:
 - WMStatsAgent contains wmstats views required only by the agent
 - WMStats contains wmstats views only required by WMStats.

I noticed the new WMStatsAgent files (updates, views, etc) are the same as those under WMStats. Shouldn't we then remove the same files under WMStats?

ticoann:
Most updates to wmstats happens in WMAgent. And most views are used from central WMStats.
Problem we have was the views, although most views (except one 'time' view) was pushed to wmagent deployment which cause a lot of disk space. Current patch kept only one view for agent deployment.
The problem is it makes the duplicate code base for maintenance.  We could do something in deployment script. (while deploying the wmagent delete the files unused instead of having the duplicate code base) - one problem with this is it is not very obvious that which code are used for WMAgent unless you look at the deployment script.  - If you prefer that way, I can change the code.
Updates codes are kept for central wmstats although some of them are not used. For the following reason.
  - It allows separate script for manipulating the data. This is needed for some recovery script.
  - It doesn't contribute to increase of the data size

src/couchapps/WMStatsAgent/validate_doc_update.js
```diff
@@ -0,0 +1,54 @@
...
+   var isGlobalAdm = (userCtx.name === null)
+                     || matchesRole("admin","")
+                     || matchesRole("_admin","")
+                     || matchesRole("-admin","group:couchdb");
+
+   //---------------------------------
+   // Authorization rules for Myapp DB
+
+   // The following rule aplies for all operation types
+   var allowed = isGlobalAdm || matchesRole("production-operator","group:dataops")
```
amaltaro: Just trying to simplify, do we need to validate the user roles for agent updates? Couch is accessed via localhost with no certificate, maybe we could drop this and the one above.

amaltaro:
I agree with you, IMO it's better to have a bit more code to maintain and have these things clearer.

src/couchapps/WMStatsAgent/validate_doc_update.js
ticoann: You are right, I updated the permissions.

PR #5855, opened 2015-05-01
do both adler32 and crc checksum calculation in one file read

hufnagel:
To improve performance and reduce disk IO on the worker node calculate the adler32 and crc checksums in one file read.

### DIFF
--- a/src/python/WMCore/Algorithms/BasicAlgos.py
+++ b/src/python/WMCore/Algorithms/BasicAlgos.py
@@ -10,7 +10,46 @@ Python implementations of basic Linux functionality
 import os
 import stat
 import time
-import hashlib
+import zlib
+import subprocess
+
+def calculateChecksums(filename):
+    """
+    _calculateChecksums_
+
+    Get the adler32 and crc32 checksums of a file. Return None on error
+
+    Process line by line and adjust for known signed vs. unsigned issues
+      http://docs.python.org/library/zlib.html
+
+    The cksum UNIX command line tool implements a CRC32 checksum that is
+    different than any of the python algorithms, therefore open cksum
+    in a subprocess and feed it the same chunks of data that are used
+    to calculate the adler32 checksum.
+
+    """
+    adler32Checksum = 1 # adler32 of an empty string
+    cksumProcess = subprocess.Popen("cksum", stdin = subprocess.PIPE, stdout = subprocess.PIPE)
+
+    # the lambda basically creates an iterator function with zero
+    # arguments that steps through the file in 4096 byte chunks
+    with open(filename, 'rb') as f:
+        for chunk in iter((lambda:f.read(4096)),''):
+            adler32Checksum = zlib.adler32(chunk, adler32Checksum)
+            cksumProcess.stdin.write(chunk)
+
+    cksumProcess.stdin.close()
+    cksumProcess.wait()
+
+    cksumStdout = cksumProcess.stdout.read().split()
+    cksumProcess.stdout.close()
+
+    # consistency check on the cksum output
+    filesize = os.stat(filename)[stat.ST_SIZE]
+    if len(cksumStdout) != 2 or int(cksumStdout[1]) != filesize:
+        raise RuntimeError("Something went wrong with the cksum calculation !")
+
+    return ("%x" % (adler32Checksum & 0xffffffff), "%s" % cksumStdout[0])
 
 
 def tail(filename, nLines = 20):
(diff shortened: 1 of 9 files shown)


hufnagel:
New pull request as it completely changes the goal of the changes. @ticoann , I still removed the crc checksum from wmstats, as IMO we should only deal with adler32 everywhere as we can (crc checksum will just stay around for legacy reasons.

But I can remove the wmstats changes if that makes validating and eventually merging this easier...

hufnagel:
@PerilousApricot , also would appreciate a look at my checksum calculation code. Its a (much) simpler version than you presented in the other issue.

Also, please note that cksum.stdin.write is highly discouraged by the python documentation, but is the only way this works. The suggested cksum.communicate() cannot deal with passing input data in chunks at a time.

hufnagel:
Also, do not merge yet, this is untested !

hufnagel:
Also included passing NOTSET as adler32 or md5 checksum to DBS if they are not defined. Instead don't pass anything to DBS.

ticoann:
For WMStats, it doesn't really matter. It is just a display purpose. If information is not needed we don't have there. Validating and merging is the same unless this need to be merged before May production deployment. But if this has to be merged before that we can do that as well.

hufnagel:
Patch has been tested in a Tier0 replay. I looked at the DBS file inject records, it only passed the adler32 and crc checksums now, not md5 anymore (and the injection still works). I also took the checksums passed to DBS and compared them to interactively calculated checksums from the same file copied from EOS to local disk.

hufnagel:
Lets leave the wmstats in place then. We keep crc checksums, because a) we save ourselves some headache with DBS and b) we get them basically for free now. They are still there, but we don't use them prominently anymore. Who knows, maybe they'll go away in DBS4 :-).

As for wmstats, as the checksum is still there, the change can be merged whenever. Its not critical that the wmstats change is deployed at the same time as the other changes, this can happen in any order.

hufnagel:
@ericvaandering @ticoann @PerilousApricot , please review/comment.

PerilousApricot:
:+1:  - Sucks that we have to spawn off a process for checksums, but like you noticed, there's something different between the command line and what zlib does :/. Rest of it looks good to me, esp. if you verified that it all looks the same interactively

hufnagel:
Btw, my interactive tests were mostly for the crc checksum calculated by cksum. If we do get adler32 wrong at the end of the job stageout will fail immediately because stageout actually checks that file written to EOS has the samer adler32 as what we pass with the file back to the agent.

ericvaandering:
So is this the proposed end state for this issue or do we move forward with the DBS issue and simplify this code in June/July?

src/python/WMCore/Algorithms/BasicAlgos.py
```diff
@@ -10,7 +10,48 @@
...
+    The cksum UNIX command line tool implements a CRC32 checksum that is
+    different than any of the python algorithms, therefore open cksum
+    in a subprocess and feed it the same chunks of data that are used
+    to calculate the adler32 checksum.
+
+    """
+    adler32Checksum = 1 # adler32 of an empty string
+    cksumProcess = subprocess.Popen("cksum", stdin = subprocess.PIPE, stdout = subprocess.PIPE)
+
+    f = open(filename, 'rb')
```
ericvaandering: You can use a context manager here (with open as f)

hufnagel:
Well, yes, removing the crc checksum from the WMCore core will simplify the code somewhat (removing like 10 lines of code in one file and a line of code each from two other files). But there really isn't much of performance benefit anymore (a few seconds for the crc algorithm to calculate from data already in memory) and with all these files in DBS already having the CRC checksum, there is an argument to be made to keep supplying it for consistency.

I think it depends on what happens on the DBS side, which is largely driven by what the policy is. Mandatory crc checksum and other checksums optional doesn't make much sense as a policy, but it is what is implemented. In effect we always pass both from WMAgent (not sure what CRAB3 does, they should probably also get these code updates...).

More a question for you though, do we want to spend time on what has stopped being a performance issue and has moved to be a policy clarification and implementation cleanup to be more consistent with a possible new policy ?

Even if we decide we still want to move forward on the DBS side, I would suggest to validate/test/merge this code update as-is and then make the other changes later once we know what we want to do on the DBS side.

src/python/WMCore/Algorithms/BasicAlgos.py
hufnagel: Just read up on it. No functional difference and the old style syntax isn't deprecated or anything like that either. There are no benefits for a simple example like this here, although I can see it being useful if what you ar doing in the file processing section is more complex.

hufnagel:
@ericvaandering , update the file open and loop construct. Works interactively, although I am also running another Tier0 replay with it.

hufnagel:
Replay worked with the latest code as well, spot check on a produced file and the cksum passed to DBS shows it to be identical to the same cksum calculated from another file copy from EOS.

ticoann:
@hufnagel, Dirk, this is ready to be merged, right?

hufnagel:
From my side definitely. I am using it in all Tier0 replays I am doing.

ericvaandering:
Merge it. Ignore my Jenkins prototype which is this "Waiting to hear" thing

PR #5863, opened 2015-05-07
Fix useSiteListAsLocation usage

amaltaro:
trustSiteLists was being set to True regardless the value of useSiteListAsLocation


amaltaro:
Added a protection in case it's set as a string during assignment. @ticoann please review
This one requires a new cmsweb upgrade, but we can wait until the next cycle

amaltaro:
tested in my VM

Issue #5857, opened 2015-05-04
Review AlertProcessor component


amaltaro:
We just got a message from CERN IT complaining that one of our agents (vocms055 - testbed) were sending too many mails to wmagentalerts@gmail.com, which slows down delivery of messages to gmail and impacts people at CERN.

Looking at the component log, it seems it retries this operation each 5 secs.

For now I have commented out a few lines in the config.py (fromAddr, smtpServer and toAddr). That should stop it.

Then we investigate how to improve it in the near future.

[1]
2015-05-04 06:38:04,487:WARNING:EmailSink:Server disconnected, reconnecting ...
2015-05-04 06:38:09,491:ERROR:Processor:Sending alerts failed (soft) on Connection unexpectedly closed, reason: EmailSink
Traceback (most recent call last):

   File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMCore/Alerts/ZMQ/Processor.py", line 88, in handleSoft
    target.send(alertBuffer)

   File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.5.patch1/lib/python2.6/site-packages/WMCore/Alerts/ZMQ/Sinks/EmailSink.py", line 50, in send
    self.smtp.sendmail(self.fromAddr, self.toAddr, msg)

   File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/external/python/2.6.8-comp6/lib/python2.6/smtplib.py", line 697, in sendmail
    self.rset()

   File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/external/python/2.6.8-comp6/lib/python2.6/smtplib.py", line 438, in rset
    return self.docmd("rset")

   File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/external/python/2.6.8-comp6/lib/python2.6/smtplib.py", line 363, in docmd
    return self.getreply()

   File "/data/srv/wmagent/v1.0.5.patch1/sw.pre/slc6_amd64_gcc481/external/python/2.6.8-comp6/lib/python2.6/smtplib.py", line 340, in getreply
    raise SMTPServerDisconnected("Connection unexpectedly closed")

[...]

PR #5866, opened 2015-05-07
Try to improve Alerts thresholds; changed email address

amaltaro:
Changing the email address to myself (I hope I don't regret it...) and also tweaking a few thresholds based on my "feeling" basically.
This won't solve #5857, but should mitigate the amount of messages.

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -389,7 +389,7 @@ config.AlertProcessor.critical.sinks.section_("email")
 # undeliverable since the wmagent machine doesn't run any email service,
 # with noreply@ it goes to /dev/null 
 config.AlertProcessor.critical.sinks.email.fromAddr = "noreply@cern.ch"
-config.AlertProcessor.critical.sinks.email.toAddr = ["wmagentalerts@gmail.com"] # add more in the list
+config.AlertProcessor.critical.sinks.email.toAddr = ["alan.malta@cern.ch"] # add more in the list
 config.AlertProcessor.critical.sinks.email.smtpServer = "cernmx.cern.ch"
 config.AlertProcessor.critical.sinks.email.smtpUser = None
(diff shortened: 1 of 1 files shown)


ticoann:
I think we should disable the email options. You can check first how much junk mail you get first by just updating one agent. If the message is very useful. We will propagate to log db or something.

amaltaro:
Yep! I will use this PR as a reference for my first checks. Don't merge it

amaltaro:
Recording here some of the messages this component sends out:
 1. "Could not find jobCacheDir /data/..." for JobArchiverPoller. It seems it happens when a workflow is aborted, so a flood of messages
 2. "Stopping retries for job 3470" for ErrorHandlerPoller. Again seems to happen when a workflow is aborted -> flood of messages.
 3. Components load in general. "Details: {u'numMeasurements': 5, u'threshold': u'10%', u'average': u'14.76%', u'period': 300}" for CPUPoller
 4. "Job 2973 exhausted due to a bad exit code (set([40, 50660, 99999]))" for ErrorHandlerPoller.

No wonder why it was flooding gmail with these emails, it's impossible to follow up all of them.
I'll try to:
 1. disable only those related to jobs (component alarms may be useful, though thresholds should be increased)
 2. otherwise, completely disable emails (but then there is no more meaning to have these components wasting resources there ...)

ericvaandering:
retest this please

ericvaandering:
retest this please

amaltaro:
... it reminds me I have to really fix this issue instead of putting it under the carpet :-(

ericvaandering:
I'm using it for testing some new "improvements" to Jenkins.

ticoann:
Alan, I will close this. since we will remove AlertSystem, Is that OK?

amaltaro:
Go for it!

PR #5868, opened 2015-05-07
Fixx issue with pycurl

vkuznet:


src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -10,6 +10,7 @@
 underlying data-services.
 """
 
+import json
```
ticoann: Valentin, this import won't be used 
Since there is following on line 18
from WMCore.Wrappers import JsonWrapper as json
https://github.com/vkuznet/WMCore/blob/pycurl-fix/src/python/WMCore/Services/pycurl_manager.py#L18

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -78,6 +79,10 @@ def set_opts(self, curl, url, params, headers,
         curl.setopt(pycurl.FOLLOWLOCATION, self.followlocation)
         curl.setopt(pycurl.MAXREDIRS, self.maxredirs)
 
+        if params:
+            if  isinstance(params, dict):
+                params = json.dumps(params)
+
         if  verb == 'GET':
             encoded_data = urllib.urlencode(params, doseq=doseq)
```
ticoann: If you pass the json string here as params I think you will get an error or wrong result.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -128,9 +133,11 @@ def parse_body(self, data, decode=False):
             try:
                 res = json.loads(data)
             except ValueError as exc:
-                msg = 'Unable to load JSON data\n%s' % data
-                raise ValueError(exc)
-            return res
+                msg = 'Unable to load JSON data, err=%s, data type=%s' \
+                        % (str(exc), type(data))
+                print msg
```
ticoann: Maybe better to log, instead of print?

src/python/WMCore/Services/pycurl_manager.py
vkuznet: I don't think I should care about string content. I add the code which should convert dict to proper json stream. But if user will pass wrong parameter the code should not do the validation here, it is up to the client.

src/python/WMCore/Services/pycurl_manager.py
vkuznet: The class itself does not depend on logger and neither accept logger parameter. I can import the logging module and redirect print statement to warning stream. But I don't know how WMCore handles logging modules. A long time ago Lassi convinced me that it is much better to use print to stdout/stderr instead of logging module (due to overhead, concurrency issues, etc). But I can add it if you want, similar to how we did in LogDB.

src/python/WMCore/Services/pycurl_manager.py
ticoann: Hi Valenitn, what I am saying is if the dictionary is passed for GET call then it is changed to json string urlencode will give the error.
May be json dumps should be apply only non GET call?
https://github.com/vkuznet/WMCore/blob/pycurl-fix/src/python/WMCore/Services/pycurl_manager.py#L91
https://github.com/vkuznet/WMCore/blob/pycurl-fix/src/python/WMCore/Services/pycurl_manager.py#L95

``` python
>>> urllib.urlencode({'a':1, 'b':2})
'a=1&b=2'
>>> urllib.urlencode(json.dumps({'a':1}))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib.py", line 1312, in urlencode
    raise TypeError
TypeError: not a valid non-string sequence or mapping object
```

src/python/WMCore/Services/pycurl_manager.py
ticoann: Ah, I didn't know about that. Thanks for explaining. I was just wondering when we move to python 2.7, didn't we say that we will remove the print statement, The problem with 'print' in WMAgent, it doesn't log to ComponentLog.

src/python/WMCore/Services/pycurl_manager.py
vkuznet: You're right, it's stressful day and my brain non working properly. I fixed the
code in PR.

ticoann: Sorry to hear that. Hope everything is fine with you. Thanks for updating the code.

vkuznet:
Seangchan, how should we move with this PR. Do you want me to modify print statement and turn it into logging?

ticoann:
Yes, please. The problem is print statement goes to stdout.log not ComponentLog which ops never look at. Could you squash the commit as well? Thank you.

vkuznet:
Ok, I changed the printout to logging, please review

ticoann:
Thanks, Patch looks good.

PR #5870, opened 2015-05-07
Fix workflow prio comparison

amaltaro:
That explains why we are updating workflow priority all the time, we were comparing long with int.
@ticoann please review


ticoann:
Looks good to me.

PR #5872, opened 2015-05-08
Fix workflow prio comparison - 1.0.5

amaltaro:
1.0.5_wmagent version of #5870 
I'm going to apply this patch very soon to all the agents.


ticoann:
Thanks Alan

PR #5877, opened 2015-05-08
Change order of table attributes, mandatory fields will be shown at top of the table

vkuznet:


ticoann:
Thanks Valentin,

ticoann:
I will set up the test in reqmgr2-dev and let you know.

Issue #5739, opened 2015-03-30
JobAccountant cannot find TotalEventCPU in fwjr


amaltaro:
Recording this error here to investigate in the next days. I've seen this error quite often in the agents, usually it happens to taskchain multicore jobs, but this case is a data taskchain
```
2015-03-30 21:45:48,050:INFO:AccountantWorker:Handling /data/srv/wmagent/v1.0.5.pre9/install/wmagent/JobCreator/JobCache/amaltaro_700_pre11_RunTau2012A_HG1504d_Ag105_150330_141000_8004/HLTD/RECODreHLT/Job
Collection_243_0/job_3934/Report.0.pkl
2015-03-30 21:45:49,188:ERROR:ChangeState:Error updating job in couch: 'TotalEventCPU'
2015-03-30 21:45:49,203:ERROR:ChangeState:Traceback (most recent call last):
  File "/data/srv/wmagent/v1.0.5.pre9/sw.pre.sryu/slc6_amd64_gcc481/cms/wmagent/1.0.5.pre9/lib/python2.6/site-packages/WMCore/JobStateMachine/ChangeState.py", line 156, in propagate
    self.recordInCouch(jobs, newstate, oldstate, updatesummary)
  File "/data/srv/wmagent/v1.0.5.pre9/sw.pre.sryu/slc6_amd64_gcc481/cms/wmagent/1.0.5.pre9/lib/python2.6/site-packages/WMCore/JobStateMachine/ChangeState.py", line 319, in recordInCouch
    updateSummaryDB(self.statsumdatabase, job)
  File "/data/srv/wmagent/v1.0.5.pre9/sw.pre.sryu/slc6_amd64_gcc481/cms/wmagent/1.0.5.pre9/lib/python2.6/site-packages/WMCore/JobStateMachine/SummaryDB.py", line 157, in updateSummaryDB
    sum_doc = fwjr_parser(document)
  File "/data/srv/wmagent/v1.0.5.pre9/sw.pre.sryu/slc6_amd64_gcc481/cms/wmagent/1.0.5.pre9/lib/python2.6/site-packages/WMCore/JobStateMachine/SummaryDB.py", line 84, in fwjr_parser
    pdict['totalEventCPU'] += float(perf['cpu']['TotalEventCPU'])
KeyError: 'TotalEventCPU'
``` 
I saved this pkl file in current of vocms009.
Feel free to assign this one to me.

amaltaro:
[...]

PR #5893, opened 2015-05-11
workaround to be able to handle cmssw74 performance data

hufnagel:
CMSSW_7_4 changes the format of the performance data in the FJR, implement a workaround to be able to use it.

@ticoann , you have to judge how urgent this is. Without this updating the job info in couch fails for every 74X job. I have no idea what the impact of that is.

### DIFF
--- a/src/python/WMCore/JobStateMachine/SummaryDB.py
+++ b/src/python/WMCore/JobStateMachine/SummaryDB.py
@@ -14,15 +14,14 @@ from WMCore.Database.CMSCouch import CouchNotFoundError
 
 def fwjr_parser(doc):
     """
-    Parse FWJR document and yeild the following structure:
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan, do we need to apply this in production agent? Are we running any CMSSW_7_4 jobs?
I don't think applying this patch will hurt anything though.

ericvaandering:
We for sure are running RelVals on 7.4 and the release is out now, IIRC and will start making requests soon.

ticoann:
Thanks Eric, 
Alan, I create the 1.0.5 version of patch. Could you apply that in production agents?

amaltaro:
I did not investigate this issue #5739 further, but it's quite possible this PR is fixing that, let's see ...
@ticoann yes, I'll take care of it now.

PR #5896, opened 2015-05-11
workaround to be able to handle cmssw74 performance data

ticoann:
CMSSW_7_4 changes the format of the performance data in the FJR, implement a workaround to be able to use it.


ticoann:
1.0.5 version of #5893, Alan could apply this in production agents?

amaltaro:
Sure, all the production agents have this patch now! Thanks

PR #5899, opened 2015-05-13
Do not shuffle PU files for deterministicPileup

amaltaro:
Shuffles pileup file list for all cases but data+deterministicPileup.
@ericvaandering please review

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -121,7 +121,7 @@ class EventAwareLumiBased(JobFactory):
 
                 #Do average event per lumi calculation
                 if f['lumiCount']:
-                    f['avgEvtsPerLumi'] = float(f['events'])/f['lumiCount']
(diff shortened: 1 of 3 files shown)


ericvaandering:
@amaltaro did you run the existing unit tests? Can you write a unit test for this new case?

ericvaandering:
And I don't think this solves the problem as we want it solved: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/LumiBased.py#L337 should have an offset which is quasi random to pick a spot inside the pre-mixed dataset.

amaltaro:
the time has come :-) Let me give it a try at the unittest setup notes you guys provided this week.
This indeed solves only the shuffling problem, it does not touch the designed deterministic PU in place.

amaltaro:
@ericvaandering as we discussed, I'm rounding the estimated # of events per lumi to avoid overlap of events. Notice it will be applied to any wf using LumiBased or EventAware, not only to deterministic pileup cases.

ericvaandering:
I think that's good. Did this not get merged yet? I suspect the ultimate resolution of this problem is going to remove most of this patch and just put a change of the seed in. :-) But this can be merged IMO.

PR #5903, opened 2015-05-14
add options on unittest script

ticoann:
will have option to deploy certain version, from certain repository and branch.
Also added code for only updating the source code when previously deployed unittest exist
https://github.com/dmwm/WMCore/wiki/Setup-wmcore-unittest

### DIFF
--- a/test/deploy/deploy_unittest.sh
+++ b/test/deploy/deploy_unittest.sh
@@ -1,13 +1,82 @@
 #!/bin/bash
-git clone git@github.com:dmwm/deployment.git
-(cd deployment; curl https://github.com/dmwm/deployment/commit/3309be9ce3b628efe39e1db30b6a1f865e1fa3e4.patch | git am)
-curl -s https://raw.githubusercontent.com/dmwm/WMCore/master/test/deploy/init.sh > init.sh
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, @ericvaandering, Please review

test/deploy/deploy_unittest.sh
```diff
@@ -1,13 +1,69 @@
...
\ No newline at end of file
+
+VERSION=1.0.7.pre3
+REPOSITORY=dmwm
+BRANCH=
+UPDATE=false
+
+deploy_agent() {
+
+    git clone git@github.com:dmwm/deployment.git
```
amaltaro: Can you change it to https (e.g.: https://github.com/USER/REPO.git). One password less :)
The same for the git clone below

amaltaro:
Then 2 more comments here not to pollute the other "view" :-)
 1. in the case statement, you should shift twice (-v VERSION)
 2. maybe add an example line on the top and that's all

ticoann:
Thanks Alan, I updated as you suggested, about the shift there is one one shift outside the case so it should work fine.

ticoann:
Merging it now for the Alex and Tapas to test. Alan, please review anyway I will fix if you find anything more.

Issue #5171, opened 2014-06-06
need update output dataset whenever version is changed.


ticoann:
https://cmslogbook.cern.ch/elog/Workflow+processing/14981

amaltaro:
Fix for this is already in place for ReqMgr2, thus I'm closing this issue. Feel free to reopen it if you wish

PR #5905, opened 2015-05-14
TaskChain fix

### LINKED ISSUE #5847: outputDataset broken for TaskChains
@vlimant just reported that both the ReqMgr API [1] and a couch view [2] is broken when:
 1. it's a TaskChain request with no AcquisitionEra (probably ProcessingString as well) in the main dict
 2. the request is still not assigned.

These views are being used to check whether there will be dataset collision (different requests creating the same output dataset).

It's quite important, but since today is the last day of validation and we've lived with that since ever, I would push it only for HG1506.

ticoann:
fixes #5847 and #5171

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
@@ -242,10 +242,9 @@ class TaskChainWorkloadFactory(StdBase):
             parentTask = None
             if parent in self.mergeMapping:
                 parentTask = self.mergeMapping[parent][parentTaskModule(taskConf)]
-
+                
             task = self.makeTask(taskConf, parentTask)
-            #sets the prepID for the task (overwrite initial PrepID if it is set by workflow
-            task.setPrepID(taskConf["PrepID"])
+            
             if i == 1:
                 # First task will either be generator or processing
                 self.workload.setDashboardActivity("relval")
@@ -268,9 +267,6 @@ class TaskChainWorkloadFactory(StdBase):
 
         self.workload.ignoreOutputModules(self.ignoredOutputModules)
 
-        # setting the parameters which need to be set for all the tasks
-        # sets acquisitionEra, processingVersion, processingString
-        self.workload.setTaskPropertiesFromWorkload()
         return self.workload
 
 
@@ -289,6 +285,21 @@ class TaskChainWorkloadFactory(StdBase):
             task = parentTask.addTask(taskConf['TaskName'])
         return task
 
+    def _updateCommonParams(self, task, taskConf):
+    # sets the prepID  all the properties need to be set by
+            # self.workload.setTaskPropertiesFromWorkload manually for the task
+        task.setPrepID(taskConf.get("PrepID", self.workload.getPrepID()))
+        task.setAcquisitionEra(taskConf.get("AcquisitionEra", self.workload.acquisitionEra))
+        task.setProcessingString(taskConf.get("ProcessingString", self.workload.processingString))
+        task.setProcessingVersion(taskConf.get("ProcessingVersion", self.workload.processingVersion))
+        lumiMask = taskConf.get("LumiList", self.workload.lumiList)
+        if lumiMask:
+            task.setLumiMask(lumiMask)
+        
+        if taskConf["PileupConfig"]:
+            self.setupPileup(task, taskConf['PileupConfig'])
+        
+        
     @ParameterStorage
     def setupGeneratorTask(self, task, taskConf):
         """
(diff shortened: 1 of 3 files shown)


ticoann:
@amaltaro, Alan,  please review (still need to be tested).

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -242,10 +242,18 @@ def __call__(self, workloadName, arguments):
...
-            #sets the prepID for the task (overwrite initial PrepID if it is set by workflow
-            task.setPrepID(taskConf["PrepID"])
+            
+            
+            # sets the prepID  all the properties need to be set by
+            # self.workload.setTaskPropertiesFromWorkload manually for the task
+            task.setPrepID(taskConf.get("PrepID", self.workload.getPrepID()))
+            task.setAcquisitionEra(taskConf.get("AcquisitionEra", self.workload.acquisitionEra))
+            task.setProcessingString(taskConf.get("ProcessingString", self.workload.processingString))
+            task.setProcessingVersion(taskConf.get("ProcessingString", self.workload.processingVersion))
```
amaltaro: "ProcessingVersion" instead of "ProcessingString"

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -242,10 +242,18 @@ def __call__(self, workloadName, arguments):
...
+            task.setLumiMask(taskConf.get("LumiList", self.workload.lumiList))
```
amaltaro: I guess it's better to first check whether there is a lumimask, no?

amaltaro:
Ok, this patch touches quite a lot of the base common code, we need to be careful merging it.
Seangchan, are all these changes really needed to fix the real problem? For example, why do we need to build the full (un)mergedLFN? Why do we care about runs?

Sorrry, I'm just scared of pushing these changes to testbed without having it in private machines first :-(

ticoann:
Hi Alan, Actually the code in WMCore is not really broken. However it assumes set the acquisition era to set like this, 

``` python
"AcquisitionEra" : {"Task1" :  "acquision1",  "Task1" :  "acquision1" }
```
That should set AcquisitionEra correctly in creation time. However, I think it is not good to set that way.
The rule I implied is, if lower level value is not set, select parent value. other wise select whatever it is set in the spec for that task.

The change seems a lot but mostly moving the function to task level from workflow level, so it will allow the properties by task if the different task has different values for that property.

Yes I agree that we shouldn't merge this before we test in vm. I am updating the code in reqmgr2-dev.cern.ch

ticoann:
What is the different between setupGeneratorTask and setupTask methods.
The reason I asked whether LumiList value need to be set for both or setupTask, it seems currently set in setupTask

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
ticoann: setupLumiMask do nothing if mask is {} but I guess we can check before that to save the unnecessary function call.
 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMTask.py#L1246

But I was wondering whether this need to set for both setupGeneratorTask and setupTask? (comment below)

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
ticoann: Thanks I will update that

amaltaro:
Yes, it makes sense to move to task level. About setupGeneratorTask, it creates a task without input dataset, like MC from scratch, usually the first task of a TaskChain request, but not always the case.

ticoann:
> Yes, it makes sense to move to task level. About setupGeneratorTask, it creates a task without input dataset, like MC from scratch, usually the first task of a TaskChain request, but not always the case.

So we don't need the lumi mask for that, right? But is it ok to move the setLumiMask outside the setTask?
It looks more general and lumi mask is only controlled by the parameter regard less the type of the Task.

amaltaro:
Exactly, we don't need lumiMask for setupGenerator. If the parameter is provided inside tasks, doesn't it make more sense to setup lumiMask inside the tasks?
LumiList parameter can also be provided in the main dict (outside tasks), but maybe we can simply drop that...

amaltaro:
My tired eye cannot see anymore problems here :) Let me know when you have the final version in a VM and agent, so I can fire a bunch of requests for the weekend.

ticoann:
You are right, I moved the setting up inside the setupTask. Also update the code in reqmgr2-dev.cern.ch
Feel free to make a request and test. Thanks,

ticoann:
To check,
First, we need to created request but not assigned first.
Then fireup bunch of request whether it brakes anything. I thought your are taking a rest. :smile: Don't have to run it now.  Enjoy the weekend.

amaltaro:
So far so good. A TaskChain workflow created without any complaints. Now a bunch assigned.
... and now, time to chill out :)

amaltaro:
Ok, test workflows successfully ran without any apparent issues. Thanks!

ticoann:
Thanks Alan for testing.

PR #5355, opened 2014-09-19
don't replicate the deleted documents for WQ

ticoann:
It seems deleting document happens locally (WorkQueueManager and cronjob) so this seems not necessary and only creates the overhead. However, this needs further test. (Don't merge yet)

PR #5906, opened 2015-05-14
don't replicate the deleted documents

ticoann:
1.0.5_agent version of #5355


amaltaro:
thanks Seangchan! I suggest to keep this one opened for the next week or so, until we run a few more tests in testbed.

ticoann:
The original patch was not cleanly applied due to other changed. Definitely we should keep this open until we test. (There might be a bug)

amaltaro:
Also applied to production agents, feel free to merge it.

Issue #5908, opened 2015-05-15
Add support for a total_events limit in EventAwareLumiBased splitting.


AndresTanasijczuk:
In LumiBased and FileBased splittings there is support for total_lumis and total_files limits respectively. Need to add a similar limit in EventAwareLumiBased splitting. The limit should be on events.

PR #5909, opened 2015-05-15
Add total_events flag in EventAwareLumiBased splitting. Fix counting of average number of events in job per file to include only good lumis. Fixes #5907 #5908

### LINKED ISSUE #5907: When using EventAwareLumiBased splitting, some jobs analyze less lumis than expected
I am testing CRAB3 with EventAwareLumiBased splitting (I am adding the feature of allowing to specify a "total_events" limit to stop creating new jobs; similar as I did for LumiBased and FileBased splitting in the past) and I observed that some jobs analyze less lumis than expected, in the sense that some jobs could analyze more lumis than they did and still fit within the requirement to analyze (approximately and at most) N (expected) events. I have identified the problem as coming from the fact that, when changing input files, the counting of how many events are already included in the job in line
https://github.com/dmwm/WMCore/blob/1.0.7.pre3/src/python/WMCore/JobSplitting/EventAwareLumiBased.py#L291
may use f['lumiCount'], which includes all lumis in the file, not only the good lumis filtered by the lumi mask. There is a min(lumisInJob, f['lumiCount']) in the counting, so when lumisInJob <= f['lumiCount'] there is not problem, since lumisInJob does count only good lumis, but when f['lumiCount'] < lumisInJob and therefore f['lumiCount'] is used in the counting, this includes all lumis in the file (https://github.com/dmwm/WMCore/blob/1.0.7.pre3/src/python/WMCore/JobSplitting/EventAwareLumiBased.py#L112-119), not only the good ones.

For my own information, the task where I observed the problem is 150513_154848:atanasi_crab_test_totalUnits_lumimask_7K_30K_without_final_fix (http://glidemon.web.cern.ch/glidemon/jobs.php?taskid=1093404) in particular job 2.

AndresTanasijczuk:
This pull has two commits, one to fix #5907 and one for #5908 
I tested both commits in a CRAB3 workflow: 150513_172334:atanasi_crab_test_totalUnits_lumimask_7K_30K_with_final_fix_2 (http://glidemon.web.cern.ch/glidemon/jobs.php?taskid=1093519) which used these CRAB configuration parameters:

Data.inputDataset = '/SingleMu/Run2012B-13Jul2012-v1/AOD'
Data.lumiMask = 'https://cms-service-dqm.web.cern.ch/cms-service-dqm/CAF/certification/Collisions12/8TeV/Prompt/Cert_190456-208686_8TeV_PromptReco_Collisions12_JSON.txt'
Data.runRange = '193093-193999'
Data.splitting = 'EventAwareLumiBased'
Data.unitsPerJob = 7000
Data.totalUnits = 30000

The task resulted in 5 jobs. In the post-job logs one can see which lumis and input files were analyzed by each job. Taking from DAS the information of how many lumis and events each file contains, I could get the average number of events per lumi for each file (f['avgEvtsPerLumi']) and then count the average number of events the task had analyzed. The result was 29880.37... So the total units limit from the second commit is working. 
Similarly I could count the average number of events each job has analyzed, and it was:
job 1: 6922
job 2: 6624.95....
job 3: 6746.62....
job 4: 6654.73....
job 5: 2932.07....
summing up the 29880.37.... in total. So one can see that each job analyzed less but almost 7000 average events. In all cases, adding one more lumi to the jobs would have caused to exceed the 7000 limit. I also counted that, without the fix of commit 1, the job 2 analyzes 5004.95.... average events from '193834': [3, 4, 29, 30, 33, 34]; (see http://glidemon.web.cern.ch/glidemon/jobs.php?taskid=1093404). While with the fix of commit 1, job 2 analyzes 6624.95... average events from '193834': [3, 4, 9, 10, 29, 30, 33, 34]; that is, it includes 2 more lumis (9 and 10) from the next input file. These 2 lumis are analyzed in job 3 when the fix of commit 1 is not included.

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -41,6 +41,7 @@ class EventAwareLumiBased(JobFactory):
 
         avgEventsPerJob = int(kwargs.get('events_per_job', 5000))
         eventLimit      = int(kwargs.get('max_events_per_lumi', 20000))
+        totalEvents     = int(kwargs.get('total_events', 0))
         splitOnFile     = bool(kwargs.get('halt_job_on_file_boundaries', True))
         ignoreACDC      = bool(kwargs.get('ignore_acdc_except', False))
         collectionName  = kwargs.get('collectionName', None)
@@ -138,7 +139,9 @@ class EventAwareLumiBased(JobFactory):
         firstLumi      = None
         lastRun        = None
         lumisInJob     = 0
+        totalAvgEventCount = 0
         currentJobAvgEventCount = 0
+        stopTask = False
         for location in locationDict:
 
             # For each location, we need a new jobGroup
@@ -275,9 +278,15 @@ class EventAwareLumiBased(JobFactory):
                         lastLumi = lumi
                         stopJob = False
                         lastRun = run.run
+                        totalAvgEventCount += f['avgEvtsPerLumi']
 
                         if self.currentJob and not f in self.currentJob['input_files']:
                             self.currentJob.addFile(f)
+
+                        # We stop here if there are more total events than requested.
+                        if totalEvents > 0 and totalAvgEventCount >= totalEvents:
+                            stopTask = True
+                            break
 
                     if firstLumi != None and lastLumi != None:
                         # Add this run to the mask
@@ -290,7 +299,16 @@ class EventAwareLumiBased(JobFactory):
                         firstLumi = None
                         lastLumi = None
 
+                    if stopTask:
+                        break
+
                 if not splitOnFile:
                     currentJobAvgEventCount += f['avgEvtsPerLumi'] * lumisInJobInFile
+
+                if stopTask:
+                    break
+
+            if stopTask:
+                break
 
         return
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -152,6 +152,7 @@ class EventAwareLumiBased(JobFactory):
                         parent = File(lfn = lfn)
                         f['parents'].add(parent)
 
+                lumisInJobInFile = 0
                 updateSplitOnJobStop = False
                 failNextJob          = False
                 #If the number of events per lumi is higher than the limit
@@ -252,6 +253,7 @@ class EventAwareLumiBased(JobFactory):
                             failNextJob = False
                             firstLumi = lumi
                             lumisInJob = 0
+                            lumisInJobInFile = 0
                             currentJobAvgEventCount = 0
                             totalJobs += 1
 
@@ -269,6 +271,7 @@ class EventAwareLumiBased(JobFactory):
                                     lumisPerJob = f['lumiCount']
 
                         lumisInJob += 1
+                        lumisInJobInFile += 1
                         lastLumi = lumi
                         stopJob = False
                         lastRun = run.run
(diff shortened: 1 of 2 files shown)


ericvaandering:
Please make sure the unit test for this method works and add a new one to test your new case.

AndresTanasijczuk:
Added the unit tests.

$ python EventAwareLumiBased_t.py.......
Ran 7 tests in 0.110s

OK


If I revert back  commit 1, the new testG_LumiMask gives:

$ python EventAwareLumiBased_t.py......F
FAIL: testG_LumiMask (__main__.EventAwareLumiBasedTest)
  File "EventAwareLumiBased_t.py", line 591, in testG_LumiMask
    self.assertEqual(jobs[0]['mask'].getRunAndLumis(), {1L: [[10L, 14L]], 2L: [[20L, 21L]], 4L: [[40L, 40L]]})
AssertionError: {1: [[10, 14]], 2: [[20, 21]]} != {1L: [[10L, 14L]], 2L: [[20L, 21L]], 4L: [[40L, 40L]]}

Ran 7 tests in 0.131s

FAILED (failures=1)

This is same to what I described about my CRAB3 test task. In testG_LumiMask, each job should analyze 8 lumis (each lumi has 100 events and I set events_per_job to 850). But job 1 got 7 lumis (lumi 4L: [40L, 40L] is given to job 2 instead of job 1). This is because when switching from fileB to fileC, lumisInJob = 7 while f['lumiCount'] = 4 (f = fileB). So min(lumisInJob, f['lumiCount']) = f['lumiCount'] = 4 which means that the estimate of the average number of events in job 1 is 100 * 5 (from fileA) + 100 * 4 (from fileB) = 900 instead of 100 * 5 + 100 * 2 (only 2 from the 4 in fileB are good lumis) = 700. With the wrong estimate of avgEvts = 900 there is no room for another lumi in job 1; with the correct estimate of avgEvts = 700 there is room for 1 more lumi.

AndresTanasijczuk:
I made a modification regarding total_events: consider total_events limit as a lower limit (total average events count must be >= total_events) instead as an upper limit. 
I also changed the unit tests accordingly.
```
$ python EventAwareLumiBased_t.py.......
Ran 7 tests in 1.086s

OK
```

hufnagel:
@lucacopa , how does this affect the Tier0 problem we had with the EventAwareLumiBases event limits ?

lucacopa:
@hufnagel This doesn't really affect us concerning that problem. Even if we set totalEvents limit, the job will end up in creation failed. The one that is triggering our problem is eventLimit = max_events_per_lumi, given an input file with a single lumi and a lot of events.

mmascher:
Can we merge this? It should be a transparent change for everything but CRAB3: WMAgent is not passing `total_events` so the code will behave as before.

ticoann:
Sorry Marco for the late merging I missed this pull request.

amaltaro:
Well, there is a change here that affect WMAgent (currentJobAvgEventCount).
I'll make sure it's Ok for WMagent in the next week

mmascher:
No problem! alan, are you sure you are not mixing currentJobAvgEventCount and totalJobAvgEventCount? I do not see what is changing when totalEvents is 0.

AndresTanasijczuk:
@mmascher Alan is right. There is this commit https://github.com/AndresTanasijczuk/WMCore/commit/8ce8872c82f61b6ecfcd876c5a5179035e4afd41 which fixes the counting of events in the current job, and has nothing to do with total_events. Was a bug that I found while working on the total_events, and decided to fix in the same pull (but different commit). The description of the bug was in the issue https://github.com/dmwm/WMCore/issues/5907

ticoann:
Bug fix for the issue #5907 looks good to me. Actually I got confused that this won't affect anything else other than agent but actually this affects GQ. I shouldn't have merged since we have the new tag in testbed. Anyway, I will try to set up vm and test this patch. (with Alan's help. :smile: )

PR #5910, opened 2015-05-15
Fix DQMHarvest spec to create Harvesting job instead of merge one

amaltaro:
I started working in the wrong branch, anyways we may need to patch 1.0.5 with this fix.
Creating the master version soon.


ticoann:
This doesn't have to be merged in 1.0.5 version since this only affects reqmgr. But I am merging it anyway.

PR #5911, opened 2015-05-15
DQMHarvest spec remake; Template updated

### LINKED ISSUE #5849: DQMHarvest spec creates merge job instead of harvesting
Everything with the new DQMHarvest spec "succeeds" until the job starts running on the worker node, where it fail with a segmentation fault or some kind.

The problems seems to be that this spec is creating a merge job instead, and so the PSet is missing the whole configuration basically.

amaltaro:
Fixes #5849.
Creates harvesting job(s) and logCollect. Tested in my VM and works Ok.
@ticoann please review.

I'm creating another issue to create unittests for this spec file.


ticoann:
Looks fine to me Thanks,

PR #5915, opened 2015-05-15
allow nested dict/list structure in Config

### LINKED ISSUE #5798: LumiList not added in the workload for TaskChain
@ticoann , I need your help with this issue since I failed miserably while trying to fix it :-(

The problem is that LumiList does not get added to the workload.request.schema section, since we call this method with an empty dict
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L802

I applied a couple of changes in the TaskChain.py spec file, basically
 - calling self.workload.setLumiList() from inside setupTask(), when the code sees the lumiList provided
 - also added LumiList argument under getTaskArguments() (I believe this one is not needed)
but that didn't work.

BTW, in the current production reqmgr (and in my VM) I get this error in the ReqMgr logs
```
ERROR:root:Invalid Value: Complex Value type in sequence:<type 'list'>
for name: LumiList and value: {'160960': [[277, 287], [79, 85], [167, 177], [226, 232]]}
Added to WMAgent Configuration
```
Does it mean we cannot have this structure under the request workload section?

This is a problem because, since LumiList is not available in the spec file - under the request.schema section, cloning TaskChain requests is not a real clone since the new request won't have this info.

Thanks a lot!

ticoann:
fixes #5798

### DIFF
--- a/src/python/WMCore/Configuration.py
+++ b/src/python/WMCore/Configuration.py
@@ -26,13 +26,15 @@ _SimpleTypes = [
     types.IntType,
     ]
 
-_SupportedTypes = [
+_ComplexTypes = [
     types.DictType,
     types.ListType,
     types.TupleType,
     ]
 
+_SupportedTypes = []
 _SupportedTypes.extend(_SimpleTypes)
+_SupportedTypes.extend(_ComplexTypes)
 
 
 def format(value):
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, Alan, please review. Not sure about side effect of this though. I updated the patch in reqmgr2-dev.cern.ch. Could you try to create the clone request which initially gives you the error?
Thanks

ticoann:
One thing, I am concern is that there must be reason we didn't allow this before. So need to be careful check things before merge this.

amaltaro:
Looks good to me. I'll create a few requests soon. Thanks!!

amaltaro:
Workflows also ran without any issues. However, we need to make sure service logs are not complaining about anything from these changes. Let's check this out during testbed validation, we may find the reason for that :-)

PR #5918, opened 2015-05-18
Allow AgentStatusWatcher to force sites do Down

amaltaro:
Added a configuration parameter so we can specify sites that we want to keep in 'Down' state.
This is needed for HLT, since FNAL schedds are not allowed to start/connect to pilots at T2_CH_CERN_HLT. I'll also fix the deployment script.

@ticoann or @lucacopa please review. I plan to push this patch in this afternoon.


amaltaro:
BTW, this needs to be added into config.py to all FNAL agents:
config.AgentStatusWatcher.forcedSiteDown = ['T2_CH_CERN_HLT]

lucacopa:
@amaltaro looks good to me

ticoann:
Alan, can I merge this?

amaltaro:
Yes, please. I'll patch the agents tomorrow.

PR #5919, opened 2015-05-18
Futurize: migrate dict has_key() to in

amaltaro:
@ticoann @ericvaandering how about we take one more step for this cmsweb cycle?

This PR _only_ migrates has_key() to in. That should be quite safe, however, we could build RPMs, deploy and run a few tests for all these services. The testbed deadline is tomorrow evening.


ericvaandering:
Sounds good to me. I've gotten pulled off into other things for the moment, otherwise I might have done this one already.

amaltaro:
and for the record, this was the procedure (most part of it in the twiki page)
easy_install future
...
futurize -f has_key src/* --no-diffs -w --nobackups
futurize -f has_key test/* --no-diffs -w --nobackups

ticoann:
May the force be with you. :smile:

amaltaro:
Indeed, but you are on the same boat, so I can always bug you ;)

PR #5920, opened 2015-05-18
create the separate log for the each thread

ticoann:
https://github.com/dmwm/deployment/pull/190 need to be updated as well


ticoann:
Tested in my vm

PR #5921, opened 2015-05-19
update needed for useSiteListAsLocation

amaltaro:
strToBool() cannot handle it if we pass "true". Should I also update these lines
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkloadTools.py#L54-56
?


ticoann:
Alan, can I merge this?

amaltaro:
Yes, please.

PR #5925, opened 2015-05-19
Debugging session to find out why PyCondor and BossAir lacking commun…

tsarangi:
…ication during job tracking

Syntax: call self explicitly

jobInfo was missing. Fixed

Global declaration of exitcodemap

Remove extra debug messages. Track how many changing/completeing/running from PyCondorPlugin

job aborted from condor log points to jobs that are removed from the queue. Mark them as complete


lucacopa:
@tsarangi looks good to me, I will be using it in the tier0 test instances and will let you know if I find another issue

PR #5926, opened 2015-05-19
add missing view

ticoann:
also fixes AgentStatusWatcher for tier0


ticoann:
Tested in vocms001 merging it.

PR #5928, opened 2015-05-20
add thread for moving to archived status

ticoann:
https://github.com/dmwm/deployment/pull/190 need to be updated as well


ticoann:
Tested in my vm

PR #5932, opened 2015-05-21
enable lazydownload for CERN merge jobs

hufnagel:
All merge jobs running at CERN force lazy-download.

Originally forced the new mode Brian implemented, which works, but has currently such big performance problems that it can't be turned on. Still want to leave the code in (disabled with an "or False) because the code technically works and will be needed again if/when the performance problems get fixed.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -36,7 +36,7 @@ def fixupGlobalTag(process):
     if hasattr(process, "GlobalTag"):
         if not hasattr(process.GlobalTag, "globaltag"):
             process.GlobalTag.globaltag = cms.string("")
-
+    return
 
 def fixupGlobalTagTransaction(process):
     """
@@ -52,7 +52,7 @@ def fixupGlobalTagTransaction(process):
     if hasattr(process, "GlobalTag"):
         if not hasattr(process.GlobalTag.DBParameters, "transactionId"):
             process.GlobalTag.DBParameters.transactionId = cms.untracked.string("")
-
+    return
 
 def fixupFirstRun(process):
     """
@@ -63,7 +63,7 @@ def fixupFirstRun(process):
     """
     if not hasattr(process.source, "firstRun"):
         process.source.firstRun = cms.untracked.uint32(0)
-
+    return
 
 def fixupLastRun(process):
     """
@@ -74,7 +74,7 @@ def fixupLastRun(process):
     """
     if not hasattr(process.source, "lastRun"):
         process.source.lastRun = cms.untracked.uint32(0)
-
+    return
 
 def fixupLumisToProcess(process):
     """
@@ -85,7 +85,7 @@ def fixupLumisToProcess(process):
     """
     if not hasattr(process.source, "lumisToProcess"):
         process.source.lumisToProcess = cms.untracked.VLuminosityBlockRange()
-
+    return
 
 def fixupSkipEvents(process):
     """
@@ -96,6 +96,7 @@ def fixupSkipEvents(process):
     """
     if not hasattr(process.source, "skipEvents"):
         process.source.skipEvents = cms.untracked.uint32(0)
+    return
 
 def fixupFirstEvent(process):
     """
@@ -106,7 +107,7 @@ def fixupFirstEvent(process):
     """
     if not hasattr(process.source, "firstEvent"):
         process.source.firstEvent = cms.untracked.uint32(0)
-
+    return
 
 def fixupMaxEvents(process):
     """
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann @ericvaandering would appreciate a review.

Without this any merge job run on CERN Cloud will potentially suffer from a factor 10 to 20 performance degradation...

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -546,7 +546,28 @@ def __call__(self):
...
+
+            cmsswVersion = os.environ['CMSSW_VERSION']
+            if cmsswVersion.startswith("CMSSW_7_5") and False:
+                print "Using fastCloning/lazydownload"
+                self.process.add_(
+                    cms.Service("SiteLocalConfigService",
+                                overrideSourceCloneCacheHintDir=cms.untracked.string("lazy-download")
+                                )
+                    )
+            elif funcName == "merge":
```
ericvaandering: Why don't you just use an OR?

hufnagel: OR for what ? These are two different configuration changes !

hufnagel: One sets lazy-download, the sets "if the job uses fast cloning -> use lazy-download"

ericvaandering:
I'd also suggest moving the deterimination of lazy download into a separate function. That main function si getting pretty long.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -546,7 +546,28 @@ def __call__(self):
...
+                print "Using lazydownload"
+                self.process.add_(
+                    cms.Service("SiteLocalConfigService",
+                                overrideSourceCacheHintDir=cms.untracked.string("lazy-download")
```
ericvaandering: Ah, I missed the absence of "Clone" in this line.

ticoann:
I don't know anything about this, so I can review. :smile: 
But I have a question.
So lazy down load will only trigger in cern machine and only merge job?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -546,7 +546,28 @@ def __call__(self):
...
         self.fixupProcess()
 
-        # repack is only run at CERN and has a streaming format as input that should never use lazy-download
+        # for CERN:
+        #   enable lazy-download for fastCloning for all CMSSW_7_5 jobs (currently off)
+        #   enable lazy-download for all merge jobs at CERN
+        if socket.getfqdn().endswith("cern.ch"):
+
+            cmsswVersion = os.environ['CMSSW_VERSION']
+            if cmsswVersion.startswith("CMSSW_7_5") and False:
```
ticoann: Is this just reminder (to remove 'and False') for later since this always False? right?

ericvaandering: Right. The fix from Brian *should* land in 7.5 but it's not there yet.

hufnagel:
Yes, the whole code block only triggers on CERN machines and only on merge jobs. Reason being that due to the weird CERN internal network layout (with two locations being 30ms apart), we cannot afford not to run lazy-download for merges but OTOH also cannot really afford to use lazy-download for all jobs due to how much disk space this uses on the local disk.

Brians patch would allow configuring this in the site-local-config and then every CMSSW job would automatically switch on lazy-download if the job was a merge job or did "merge-like" data handling.

So the second code block (that's currently disabled with the False) would test this behavior before we configure it in site-local-config.

@ericvaandering , I'll look at factoring this out into a separate function.

hufnagel:
@ericvaandering , you like this better ? Tried a bit of code cleanup, unifying naming conventions, adding return statements to method, that sort of thing. Still not 100% internally consistent, but I didn't want to change too much either.

hufnagel:
Tested in a Tier0 replay, repack and merge jobs behave in the correct way and all other jobs at least work as well (but I didn't check the logs for them).

ericvaandering:
Yeah. Just two comments. First, I don't have an issue with starting a function name with underscore if it's not supposed to be called by anything but inside the class. This is kind of a convention we have and I don't think it's just us. 

Second, you modified where the lazy download fixups are handled to only be when a process is created. I hope that's what you want.

In any case, I am fine to merge this.

hufnagel:
I didn't mind the convention, but I do mind a convention that is inconsistently applied. It doesn't make much sense to follow one convention for methods A, B, C and another for methods D, E, F. Plus from what I have seen of at least WMCore code, we rarely use the _method convention. It's still not completely consistent in this file, but as I said, I didn't want to change everything :-)

I moved the repack and merge modifications because for now they only apply to Config.DP created repack and merge processes. If Brian ever gets his patch working properly we'll have to revisit this anyways (best case is this method will disappear altogether and move to be configured from site-local-conf).

hufnagel:
If there are no objections to the code, can we get this merged soon. I have a bunch of changes piled up and would like to build a new T0 release soon, preferably with this in it.

hufnagel:
@ticoann ?

PR #5933, opened 2015-05-21
do the phedex recovery block by block

hufnagel:
With too many block/files in the recovery attempt, the call to the PhEDEx data service can time out (as it did for a recent example where we had 2500 blocks with a total of 38k files). Split it to run block by block. Recovery should be rare, we don't care about doing it fast.

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
@@ -8,7 +8,6 @@ Poll the DBSBuffer database and inject files as they are created.
 import threading
 import logging
 import traceback
-from collections import defaultdict
 from httplib import HTTPException
 
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
@@ -64,7 +63,7 @@ class PhEDExInjectorPoller(BaseWorkerThread):
         #    self.sendAlert will be then be available
         self.initAlerts(compName = "PhEDExInjector")
 
-        self.filesToRecover = None
+        self.blocksToRecover = None
 
     def setup(self, parameters):
         """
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann please review. Not sure how to test this. I hot-patched the production Tier0 where we had this problem, but not with this exact same patch (patched in place, then re-wrote the same patch again from memory for the pull request).

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -331,18 +335,27 @@ def recoverInjectedFiles(self):
...
-        self.setStatus.execute(injectedFiles, 1)
-        myThread.transaction.commit()
-        # when files are recovered set the self.file 
-        self.filesToRecover = None
-        return injectedFiles
+
+        # recover one block at a time
+        for block in self.blocksToRecover:
+
+            injectedFiles = self.phedex.getInjectedFiles(block)
```
ticoann: If this calls fails (due to some temporary network error), should we just log the error and keep doing for other block, although most likely subsequent call would fail in that case.

ticoann:
Hi Dirk, Patch looks good to me. I made a comment for a minor thing which I am not sure it is necessary.
For the test, we can add some unittest for existing files in PhEDEx but I think if we test normal case in the agent test, it should be OK. I will first test that. Thanks,

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
hufnagel: Well, its recovery. If it fails during recovery, I am not sure it makes sense to add a recovery option for the recovery :-). Let it fail and someone needs to investigate. In any case, it will crash now same was as it crashed before.

hufnagel:
I guess this is a bit on the back burner now with all the other problems...

ticoann:
@hufnagel, Dirk do you want me to merge this one? I will create a new tag tomorrow.

hufnagel:
Not urgent, as the production Tier0 is hot-patched and this is not really relevant for the replay instances. Merge at your convenience...

PR #5936, opened 2015-05-22
don't make unecessary dbs call for pileup data

ticoann:
This patch contains 3 things.
1. Increase QueueDepth default value to 1.
2. Don't stop the subscription creation if one of the workflow (block) fails to create. send the error messge to LogDB.
3. Don't make the duplicate dbs call for pile up dataset  only one per step not every block.

### DIFF
--- a/src/python/WMCore/WorkQueue/WMBSHelper.py
+++ b/src/python/WMCore/WorkQueue/WMBSHelper.py
@@ -12,6 +12,7 @@ Use WMSpecParser to extract information for creating workflow, fileset, and subs
 import copy
 import logging
 import threading
+import traceback
 from collections import defaultdict
 
 from WMCore.WMRuntime.SandboxCreator import SandboxCreator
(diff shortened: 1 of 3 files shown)


ticoann:

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -403,7 +404,14 @@ def createSubscriptionAndAddFiles(self, block):
         self.beginTransaction()
 
         self.createTopLevelFileset()
-        sub = self.createSubscription(self.topLevelTask, self.topLevelFileset)
+        try:
+            sub = self.createSubscription(self.topLevelTask, self.topLevelFileset)
+        except Exception, ex:
```
ericvaandering: Hey, lets do exception "as" since we fixed this everywhere. :-)

ticoann: Sorry old habit, fixed now.  Thanks for reviewing.

amaltaro:
Good catch about the PU calls! Patch looks good to me.

Issue #5935, opened 2015-05-21
Make sure MaxRSS and Memory are settable by Task


ericvaandering:
In the same way we need to schedule cores task by task we need to do the same with the Memory requested and the MaxRSS used for killing jobs on the WN.

ericvaandering:
This is the commit for the multicore stuff

https://github.com/ericvaandering/WMCore/commit/204a1f7074b8b0bf2774f26e886f712f778c79d1

PR #5937, opened 2015-05-22
Add test and code for propogating Memory setting

ericvaandering:
@ticoann @amaltaro does this make sense as a solution to #5935 ?

The unit tests pass. Do I have to add for MaxRSS too?

Or is there an easier way to get settings from TaskChain to take things in the task first and in the workload if they are not present?


ticoann:
Patch looks good. For MaxRSS and others (MaxVSize, SoftTimeout,  GracePeriod), it got set only one time and propagated to task level.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py#L349
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1541

I think it has to be a slightly different. (a little more change.) :smile:

ericvaandering:
The problem isn't that it's not getting passed to the task level, it is. What we want is to, like in this request, https://github.com/dmwm/WMCore/blob/fix_task_memory/test/data/ReqMgr/requests/Integration/TaskChain_Task_Multicore.json set values at the top but have them overriden if (and only if) they appear inside one of the tasks.

ticoann:
Yes that is what I meant. It shouldn't be overwritten by workload level setting. It should be set in task level. Unlike memory case (setupProcessingTask is called to set the memory value), there is no calls made to set the task level attribute. So that is why I said there will be a little more change than the memory case.

amaltaro:
Patch looks good to me. MaxRSS and MaxVSize are set only during assignment, we could cover these 2 as we do for AcquisitionEra for TaskChains, setting a dict like {'TaskName1': 'Blah', 'TaskName2': 'Blah2'}

About GracePeriod and SoftTimeout, I believe we can keep these at workload level only.

ticoann:
@ericvaandering, Eric can I merge this one first. and we will do separate for other parameters,  you can assign to me.

ericvaandering:
Sure


> On May 27, 2015, at 16:02, ticoann <notifications@github.com> wrote:
> 
> @ericvaandering <https://github.com/ericvaandering>, Eric can I merge this one first. and we will do separate for other parameters, you can assign to me.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/dmwm/WMCore/pull/5937#issuecomment-106056750>.
> 




Eric W. Vaandering 
Computing Physics Developer

Scientific Computing Division
630 962 7260 office
www.fnal.gov <http://www.fnal.gov/>

PR #5938, opened 2015-05-26
Don't make unnecessary dbs call for pileup data, 105 version

amaltaro:
1.0.5_wmagent version of #5936


amaltaro:
Since it was partially tested in the past week, and testing it is a bit tricky, I'm pushing it now to agents not very loaded (cmssrv214 and vocms0311). If everything is fine, we push it to others.

ticoann:
It took long time to create the subscription on cmsgwms-submit2. Patching that machine to see whether there will be improvement. This patch can be further improve by caching Pileup dataset record regardless which workflow uses. (First need to check how this work)

amaltaro:
I think it's safer to have a query per PU dataset (workflow), it's only 1 in the end and we avoid having stale caching which may badly affect the jobs downstream.

PR #5939, opened 2015-05-27
remove MultiProcessing and MultiProduction type

ticoann:
still need to clean up unittest


ticoann:

amaltaro:
seems good to me. However I did not check whether there are other places we need to do this cleanup(?)

ticoann:
There are unittest which need to be clean up. I will have separate commit. Only thing actually need to get in this time is removing multicore task type setting from ReReco workflow. Other things are just clean up. I am not completely sure whether this will break though.

amaltaro:
I would not backport this fix. We push this fix to reqmgr testbed (this week) and we apply it to the next WMAgent version (should come in a couple of weeks). So we'll have time to validate it ;-)

ticoann:
Yes, there is no need to backport this. This only affect reqmgr change. Backporting agent doesn't make sense.

amaltaro:
Hmm, I hope to learn that once and for all by the end of Run2 :-)

PR #5946, opened 2015-05-28
add rejected transion from announced

### LINKED ISSUE #5944: Having a state transition from announced/normal-archived to rejected
For anything that needs to be rejected after it has been announced, it would be less confusing and better for later book-keeping to have things rejected being in status rejected.

ticoann:
fixes #5944 @amaltaro, please review


amaltaro:
Looks good. Just wondering, don't we have a place in wmstats code also doing/allowing these transitions?

ticoann:
Good catch, Alan. I completely forgot. But I don't think we have consider that since we are dropping wmstats request db. I will remove the code updating wmstats request db.

PR #5947, opened 2015-05-29
fix JobCreator connection fail to Oracle database

hufnagel:
JobCreator sometimes looses the connection to the Oracle database. There is a workaround for it, but this didn't trigger anymore, probabaly because the error message changed with the new sqlalchemy version.


hufnagel:
@ticoann , please review. I in any case will tell Tier0Ops to deploy immediately and start using this in replays as we have suffered from it quite a bit.

ticoann:
Hi Dirk, Do you need a new tag?

hufnagel:
Eventually, but not urgently. I am not planning to build a new release just for this. I'll look at where you are with WMCore patches when I have something on the T0 side that warrants a new release.

ticoann:
OK. Thanks, I will just create when we test new agent, then.

PR #5948, opened 2015-05-30
Fix phedex GQ data location to fetch block location

amaltaro:
If we want to also retrieve block data location from 'subscriptions' API, this collapse=n needs to be provided.
I ran interactive tests and it's fetching the correct data location now.
@ticoann this needs to go into testbed in the beginning of this week


ticoann:
Thanks Alan, I will create a new tag and request another update this weekend.

amaltaro:
Thanks Seangchan! From the standard set of workflows, I ran almost all of them and I've not seen any issues. So there is nothing else from my side for today at least :-)

PR #5949, opened 2015-06-01
Tweak a few config parameters

amaltaro:
Instead of changing these parameters all the time in the deployment script, let's change it on the root level.

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -43,7 +43,7 @@ workloadSummaryURL = couchURL
 
 # Information for the workqueue, email of the administrator and the team names
 # for this agent.
-userEmail = "OP EMAIL"
+userEmail = "cms-comp-ops-workflow-team@cern.ch"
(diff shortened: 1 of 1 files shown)


etc/WMAgentConfig.py
```diff
@@ -61,9 +61,13 @@
 globalDBSUrl = "https://cmsweb.cern.ch/dbs/prod/global/DBSReader"
 globalDBSVersion = "DBS_2_0_8"
 
-# Job retry information.  This includes the number of times a job will tried and
-# how long it will sit in cool off.
-maxJobRetries = 3
+# List of SE for T1 _Disk endpoints (TODO clean this up at some point)
+diskSites = ['storm-fe-cms.cr.cnaf.infn.it','srm-cms-disk.gridpp.rl.ac.uk',
```
ericvaandering: Make sure @alexanderrichards knows about this and can adjust in what he's doing.

PR #5951, opened 2015-06-01
change default arch value and remove couch patch

ticoann:
@amaltaro, please review

### DIFF
--- a/bin/deploy-wmagent.sh
+++ b/bin/deploy-wmagent.sh
@@ -39,8 +39,8 @@ OP_EMAIL=cms-comp-ops-workflow-team@cern.ch
 HOSTNAME=`hostname`
 
 # These values may be overwritten by the arguments provided in the command line
-WMA_ARCH=slc5_amd64_gcc461
-REPO="comp=comp"
+WMA_ARCH=slc6_amd64_gcc481
+REPO="comp=comp.pre”
 AG_NUM=0
 FLAVOR=mysql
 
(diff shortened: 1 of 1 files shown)


bin/deploy-wmagent.sh
```diff
@@ -191,12 +191,6 @@ unzip -q deployment.zip
 cd deployment-$CMSWEB_TAG
 set +e 
 
-### Patching Couchdb1.6
-echo -e "\n*** Applying deployment patches (couchdb1.6, etc) ***"
-wget -nv https://github.com/dmwm/deployment/pull/162.patch -O - | patch -p 1
-chmod 600 /data/certs/service{cert,key}.pem
```
amaltaro: I would keep this line to make sure certs are with the right permissions, otherwise couch1.6 will crash.

amaltaro:
Sounds good! I kept the couch patch because once I start deploying agents with an HG deployment tag, I don't like changing it.

bin/deploy-wmagent.sh
ticoann: Hmm, I thought this was affecting only central site... or is it vice versa? When I ran the test without it wasn't make couch crash. Could you check as well?

amaltaro: I had problems with servicekey.pem set to 400 in the past, for local agent.

ticoann:
@amaltaro, I updated the patch leaving the file permission change. (I think you are right my permission was already changed for test before the script was run) Maybe that is why it worked. :smile: 
Please review.

ticoann:
> Sounds good! I kept the couch patch because once I start deploying agents with an HG deployment tag, I don't like changing it.

Sorry I don't follow, current tag contains that couch patch, doesn't it?

amaltaro:
Yes, current HG tag contains that. When I started deploying the 1.0.5 agents, I was using 1504x tag which did not have the patch in, since new changes got in as well, I didn't change the HG tag for deploying the same agent version.

Summary :-) feel free to remove the patch here. It would need to be removed anyways in a week or two :)

amaltaro:
BTW, since you are updating it, can you update line 43 to (since we still don't have comp for slc6 scram)? 
REPO="comp=comp.pre"

Thanks!

ticoann:
Thanks Alan, It is updated now. I will merge that patch then.

PR #5952, opened 2015-06-02
fix JobCreator crashes due to loosing Oracle connection v2

hufnagel:
Second version, copy&paste error, exception matching string was slightly wrong.

### DIFF
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -411,7 +411,7 @@ class JobCreatorPoller(BaseWorkerThread):
                    and getattr(myThread.transaction, 'transaction', False):
                 myThread.transaction.rollback()
             # Handle temporary connection problems (Temporary)
-            if "InterfaceError: not connected" in str(ex):
+            if "(InterfaceError) not connected" in str(ex):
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann @amaltaro

amaltaro:
Sounds good Dirk. I'll apply it to all the production nodes soon. Thanks!

hufnagel:
This time it should work. I looked at the code where it printed the exception, than looked for the matching line in the component log and based the check on that printout.

ticoann:
Hi Dirk, would it be too loose fitting just check "not connected" string (with case insensitive) or matching ORA code if there is one, (I am not sure whether the message is from Oracle or somewhere else).

hufnagel:
The full printout for the exception is "(InterfaceError) not connected None None" and it most likely comes out of the Oracle python bindings. As long as we don't update those we should be fine. The last version we used for 5 years I believe :-)

hufnagel:
In any case, I make a note to look at this error a bit deeper. It should not happen in the first place. What we did here is a bandaid, not a real fix.

ticoann:
OK. If we have to pay Oracle we are not going to upgrade another 5 years. :smile:  Merging it, since it is affect only agent.

PR #5953, opened 2015-06-02
Multiple DQMGui Urls; Keep JobCreator running if Oracle connection is…

amaltaro:
… lost
Backporting 2 fixes:
- #5807 to support uploading root files to more than one DQMGui (needed for the DQMGui server commissioning). IMO we should make it a list, but anyways this fix will vanish in a couple of months...
- #5952 not to crash JobCreator when oracle connection is lost

@ticoann please don't merge this yet. I want to run it for a few days in testbed first.


hufnagel:
Yeah, agree making it a list would have been cleaner. But it also would have meant having to change the type of the parameter you are passing along, potentially breaking other code that relies on this being a string. Was just lower impact doing it this way...

amaltaro:
Applied to production agents too. Feel free to merge it

PR #5954, opened 2015-06-02
Shuffle pileup files deterministically if needed

ericvaandering:
Don't merge just yet. @amaltaro want to test this? for deterministic cases, you should see the same file list every time.

No one has said "This is what we want" but I think it's going to be the eventual answer.

Behavior should be the same for the non-deterministic case.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -434,14 +434,14 @@ class SetupCMSSWPset(ScriptInterface):
                     if requestedPileupType == 'data':
                         baggage = self.job.getBaggage()
                         if getattr(baggage, 'skipPileupEvents', None) is not None:
+                            # For deterministic pileup, we want to shuffle the list the
+                            # same for every job in the task and skip events
+                            random.seed(self.job['task'])
+                            print "Skipping %d pileup events for deterministic data mixing" % baggage.skipPileupEvents
                             inputTypeAttrib.skipEvents = cms.untracked.uint32(int(baggage.skipPileupEvents) % eventsAvailable)
                             inputTypeAttrib.sequential = cms.untracked.bool(True)
-                        else:
-                            # we do not want to shuffle when it's deterministicPileup
-                            random.shuffle(inputTypeAttrib.fileNames)
-                    else:
-                        # Prevent each worker at a site from requesting the same file to find products
-                        random.shuffle(inputTypeAttrib.fileNames)
+                    # Shuffle according to the seed above or randomly
+                    random.shuffle(inputTypeAttrib.fileNames)
         return
 
     def _getPileupMixingModules(self):


amaltaro:
Thanks Eric! I got a few questions and comments:
 1. these single quotes seem weird, are they copy/pasted from word or something like that?
 2. can you remind me what task in a job is? If it's the taskName, then I'm afraid we'd always use the same seed (StepOneProc), unless it's prefixed by the WF name.
Otherwise it looks fine and I can give it a try by the end of the week.

ericvaandering:
1. I don't know what you are talking about. They look fine to me. :-) I fixed them, they were cut and pasted out of our Slack discussion.

2. Good question. I had assumed it was the task with workflow/dataset names but perhaps not.

amaltaro:
Indeed single quotes look 'normal' now :-)
Ok, I'll try to figure out what this task is tomorrow.

ericvaandering:
Tracing it back, it should be set here: 
        self.currentJob["task"] = self.subscription.taskName()

I'm assuming the "job" created by the job splitting is the same thing that ends up on the worker node since that's how the baggage parameters are translated.

If that's not good, then we have to use something else that uniquely identifies the (sub)-workflow on the worker node

ericvaandering:
Don't ever accuse me of using MS Word again.....

amaltaro:
Ok, then we are good, it contains the wf name as well.
No hard feelings please, it was the first thing that came up to my mind :-P

ericvaandering:
:-) I meant to give a reference: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/JobFactory.py#L122

ticoann:
@ericvaandering, Eric, can this be merged? And could you update the patch? Sorry I waited to long for this and there is conflict now.

ericvaandering:
I can update it, but lets not merge it until Physics & PPD decide this is what they want. (This will be what they want....)

ticoann:
Thanks Eric, I will wait then.

ericvaandering:
Patch updated. Hopefully they will make up their mind before it diverges again.

ericvaandering:
@amaltaro can you give this a test? What you should see is that two jobs in the same task/workflow should have the same list of pileup files but two jobs in different workflows should have the files in a different order.

amaltaro:
Sure, I'll take care of it on Tuesday/Wednesday. Will get back here with the outcome. Thanks

ericvaandering:
Hey @amaltaro . Any news on this? Apparently they are getting close to submitting the workflow that will generate the pre-mixed sample and after that will want to use it. We need this patch in place for the 2nd part. 

@sextonkennedy can follow directly here if she likes.

amaltaro:
There was a castor intervention during the whole day and I did not manage to retrieve the logs to x-check everything is in place. I should have updates tomorrow early in the morning

ericvaandering:
Great!

amaltaro:
Sigh, I'm so stupid... I managed to retrieve the logs from castor only an hour ago and, then I found out the list of files is completely different ... of course, I forgot to inject ReDigi workflows with 'DeterministicPileup'...
Sorry, rerunning these tests.

sextonkennedy:
@ericvaandering Thanks for adding me to the thread.

amaltaro:
Bad news, the files order is different inside the same workflow. I don't know why, since the code looks correct for me. I'll investigate it further on Monday. 

Another option would be to shuffle the json file only once (still in the agent) and then don't shuffle in the WN when it has skipEvents.

ericvaandering:
Crap. Can you add a couple of printouts making sure the task name is filled and is being used to see the random number generator?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -434,14 +434,13 @@ def _processPileupMixingModules(self, pileupDict, seLocalName, modules, requeste
                     if requestedPileupType == 'data':
                         baggage = self.job.getBaggage()
                         if getattr(baggage, 'skipPileupEvents', None) is not None:
+                            # For deterministic pileup, we want to shuffle the list the
+                            # same for every job in the task and skip events
+                            random.seed(self.job['task'])
```
ericvaandering: Like right here, @amaltaro ?

amaltaro: Sure, that will be my first attempt! The painful part is that this logging is only available in the wrapper/worker node, which is not returned in the logarchive. I'll keep you up-to-date ;-)

amaltaro:
Eric, I think I know what's happening here, let me know if I'm right or not.
This is the workflow configuration I'm using:
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_ReDigi_cmsRun2_Agent108_Validation_Deter_v14_150622_192003_138

which uses this python config for the DIGI step
https://cmsweb.cern.ch/couchdb/reqmgr_config_cache/13781a5a5bbb4736e868a76467505c48/configFile

Looking at this config, it has only MixingModule (?) and then it has an empty dataMixModules
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L464

which never goes inside this if
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L434

since there are no dataMixModules and we only have deterministic shuffling for data pileup.

Is this investigation right? Any reason to perform the deterministic shuffling only for 'data' pileup?

amaltaro:
Just confirmed with more logging that 'data' pileup has no modules for our test workflow. So questions are:
 - should we apply deterministic to 'data' pileup only?
 - perhaps the config for that workflow is too old, and newer DIGI config (output from cmsDriver.py) always come with both MixingModules?

sextonkennedy:
Hi Alan,

See my comment below:
On Jun 22, 2015, at 9:29 PM, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


Eric, I think I know what's happening here, let me know if I'm right or not.
This is the workflow configuration I'm using:
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_ReDigi_cmsRun2_Agent108_Validation_Deter_v14_150622_192003_138

which uses this python config for the DIGI step
https://cmsweb.cern.ch/couchdb/reqmgr_config_cache/13781a5a5bbb4736e868a76467505c48/configFile

Looking at this config, it has only MixingModule (?) and then it has an empty dataMixModules
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L464

which never goes inside this if
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L434

since there are no dataMixModules and we only have deterministic shuffling for data pileup.

Is this investigation right? Any reason to perform the deterministic shuffling only for 'data' pileup?

Short answer: Yes, deterministic shuffling is only for data pileup
Longer answer: the non-data pileup mixing is what is used to create the pre-mixed data sample libraries that we use over and over again with different signal samples.  For this we want “classic mixing”.  For the second step where you read in the premixed pileup, the dataMixing module is used.

I hope this helps.

Cheers, Liz

amaltaro:
Thanks Liz, I think it does!
Now I'm trying to map it to WMAgent, since flipping the deterministic flag to True won't give the results unless we have the correct config and input PU dataset.

I'm creating a new wf template based on the old production premixed workflow we handled (alahiff_TOP-Spring14premixdr-00001_00003_v0__140731_152029_8397).

ericvaandering:
Sorry, I was offline yesterday dealing with stuff at home. It seems like you know how to test this now and were just using a config with a regular mixing instead of the data/pre-mixing config we will be using?

amaltaro:
No problem! Yep, I thought I knew :) but my morning test with all the correct config and etc did not work again.
I made a premixing template (from an original premix ReDigi) and will give it a try at this fix:
https://github.com/amaltaro/WMCore/commit/d3ce8036ca1a16e2dad8ebd5867a489423870dc4

It basically fails because we do not have 'data' pileup in CMSSW listSections (only 'mc') and, to get this type over there, one needs to provide DataPileup in the request instead of MCPileup.

BTW, this is the new DIGI config (in case you want to x-check)
https://cmsweb.cern.ch/couchdb/reqmgr_config_cache/36bd365447c65b7bef5213faed096689/configFile

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
amaltaro: @ericvaandering here is the print baggage.skipPileupEvents location I was talking about (which gets printed in scramOutput.log, for future reference...).
Once you add this print statement, @ticoann  will merge it and cut a new release.

ericvaandering:
Ok, ready to merge. @amaltaro I added the print statement. @ticoann should be good to merge and cherry-pick as needed.

PR #5955, opened 2015-06-03
add MaxRSS etc per task in task chain workflow

### LINKED ISSUE #5940: Create task chain params for MaxRSS, etc
similar to #5937

ticoann:
fixes #5940. Initial patch - need to be tested. Also need to update unittest. (Please don't merge yet)
For task chain, workflow level performance monitoring shouldn't be defined in assignment stage, otherwise it will be overwritten. The value need to be defined in creation stage defining assignment stage doesn't work.

### DIFF
--- a/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
+++ b/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
@@ -328,7 +328,7 @@ class Assign(WebAPI):
         helper.setProcessingVersion(kwargs["ProcessingVersion"])
         helper.setAcquisitionEra(kwargs["AcquisitionEra"])
         helper.setProcessingString(kwargs.get("ProcessingString", None))
-
+        
         # Now verify the output datasets
         outputDatasets = helper.listOutputDatasets()
         for dataset in outputDatasets:
@@ -346,10 +346,10 @@ class Assign(WebAPI):
         helper.setMergeParameters(int(kwargs.get("MinMergeSize", 2147483648)),
                                   int(kwargs.get("MaxMergeSize", 4294967296)),
                                   int(kwargs.get("MaxMergeEvents", 50000)))
-        helper.setupPerformanceMonitoring(int(kwargs.get("MaxRSS", 2411724)),
-                                          int(kwargs.get("MaxVSize", 20411724)),
-                                          int(kwargs.get("SoftTimeout", 129600)),
-                                          int(kwargs.get("GracePeriod", 300)))
+        helper.setupPerformanceMonitoring(kwargs.get("MaxRSS", None),
+                                          kwargs.get("MaxVSize", None),
+                                          kwargs.get("SoftTimeout",None),
+                                          kwargs.get("GracePeriod", None))        
 
         # Check whether we should check location for the data
         useAAA = strToBool(kwargs.get("useSiteListAsLocation", False))
@@ -391,6 +391,10 @@ class Assign(WebAPI):
                                      blockCloseMaxEvents, blockCloseMaxSize)
 
         helper.setDashboardActivity(kwargs.get("Dashboard", ""))
+        # set Task properties if they are exist
+        # TODO: need to define the task format (maybe kwargs["tasks"]?)
+        helper.setTaskProperties(kwargs)
+        
         Utilities.saveWorkload(helper, request['RequestWorkflow'], self.wmstatWriteURL)
         
         # update AcquisitionEra in the Couch document (#4380)
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -228,7 +228,8 @@ class StdBase(object):
                             userSandbox = None, userFiles = [], primarySubType = None,
                             forceMerged = False, forceUnmerged = False,
                             configCacheUrl = None, timePerEvent = None, memoryReq = None,
-                            sizePerEvent = None, useMulticore = True, applySiteLists = True):
+                            sizePerEvent = None, useMulticore = True, applySiteLists = True,
+                            taskConf = {}):
         """
         _setupProcessingTask_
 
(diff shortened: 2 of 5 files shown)


ticoann:
@amaltaro, does these value (MaxRSS etc) need to be defined in assignment stage?

amaltaro:
Hi Seangchan, the correct thing would be defining these values during assignment, since it's up to CompOps to decide them (not PPD, though we use their Memory provided to define that...).
It can also get confused if we have these settings at creation level only for TaskChain and at assignmnent for all the others.

In addition to that, I really think we do _not_ need to have 'SoftTimeout' and 'GracePeriod' defined for each task, it should not change among tasks. So we need to update only MaxRSS and MaxVSize.

ticoann:
Thanks Alan, you are right. I will change in assignment stage. (It is a little more complicated to do that. I guess do it write is more important). I think we need to have general rule. 
If a parameter is set in workflow level it will be set for all the tasks, but if task level is set it will overwrite initial workflow level values. I will create the new patch.

ticoann:
How about MaxMemory? Should that be in create stage?

amaltaro:
This applies only for TaskChain workflows and the general rule should be:
 - if a parameter is set for the whole workflow but not for any task, then use it and set it for all the tasks
 - if a parameter is set only for task, well, then we don't really need it for the whole workflow, just use those provided by task
 - if a parameter is given both for whole workflow and task, use the task one

About MaxMemory, hmmmm, it sounds like a leftover from some old cleanup to me. I really don't remember anything about it. Maybe it's used only by T0?
PS.: Go sleep! :P

amaltaro:
Seangchan, a good hint to develop this feature (change) :-)
In the current assignment of TaskChain workflows, we provide a dict for a few fields, examples are AcquisitionEra and ProcessingString, e.g.:
"AcquisitionEra": {"DIGIPU1": "CMSSW_7_0_0_pre11", "RECOPU1": "CMSSW_7_4_2"}

I believe this is what we would need to aim here, e.g.:
"MaxRSS" : {"DIGI": 8000000, "RECO" : 4000000}

ticoann:
Thanks Alan, I changed that to put the params (AcquistionEra, etc) under tasks. (Actually supports both)
Isn't something like following more consistent, since we are setting other task level params that way?
Task1: {'MaxRSS': ...},
Task2: {'MaxRSS': ...}

amaltaro:
Hi Seangchan,
> I changed that to put the params (AcquistionEra, etc) under tasks.

some of them should be under tasks already (at least those I mentioned), so no changes are needed here.

About your suggestion, it sound more clear to me, though in this case there would be breaking changes (at least on the assignment side) and probably more changes to make eveything task-based work like that, it's better to change the least possible at this point IMO.

ticoann:
> About MaxMemory, hmmmm, it sounds like a leftover from some old cleanup to me. I really don't remember anything about it. Maybe it's used only by T0?

Sorry I meant actually Memory params, Is that need to be set in assigned phase as well?

amaltaro:
Ahh, no no. Memory is set during request creation only, which BTW is working perfectly fine in testbed now :-)

ticoann:
Alan, could you take a quick look at it? I still need to test and add the unittest. (Set the value for assign stage)
``` python
"assignRequest":
{  

"Tasks": {"ProdMinBias":  {"MaxRSS": 1234}
                  "
    "SiteWhitelist": [],
    "Team": "Team-OVERRIDE-ME",
    "AcquisitionEra": {"ProdMinBias" : "AcquisitionEra-OVERRIDE-ME",
                       "DIGIPROD1" : "AcquisitionEra-1-OVERRIDE-ME",
                       "RECOPROD1" : "AcquisitionEra-2-OVERRIDE-ME"},
    "ProcessingString": {"ProdMinBias" : "ProcessingString-OVERRIDE-ME",
                         "DIGIPROD1" : "ProcessingString-1-OVERRIDE-ME",
                         "RECODreHLT" : "ProcessingString-2-OVERRIDE-ME"},
    "Dashboard": "Dashboard-OVERRIDE-ME",
    "SiteBlacklist": [],
    "MergedLFNBase": "/store/relval",
    "UnmergedLFNBase": "/store/unmerged",
    "MinMergeSize": 2147483648,
    "MaxMergeSize": 4294967296,
    "MaxMergeEvents": 100000,
    "ProcessingVersion": 1,
    "MaxRSS": 3072000,
    "MaxVSize": 20294967,
    "SoftTimeout": 129600,
    "GracePeriod": 1000,
    "CustodialSites": [],
    "NonCustodialSites": [],
    "AutoApproveSubscriptionSites": [],
    "SubscriptionPriority": "Low",
    "CustodialSubType" : "Move",
    "BlockCloseMaxWaitTime" : 72000,
    "BlockCloseMaxFiles" : 500,
    "BlockCloseMaxEvents" : 200000000,
    "BlockCloseMaxSize" : 5000000000000
}
```

amaltaro:
I'd choose the first format too. However, I'd really prefer the same format as AcquisitionEra
```
"MaxRSS": {"ProdMinBias" : 2000,
                     "DIGIPROD1" : 3000,
                     "RECOPROD1" : 2500}
```
But I believe this is out of question ... :-D

ticoann:
Whatever you want. :smile:  dict format is supported now.

ticoann:
I think this need to be cleanup later though. But I will try to test and request to put in the testbed today

ticoann:
Merging it for the testbed update (change the code to support only 

``` python
"assignRequest":
{"MaxRSS": 3072000,
  "MaxVSize": 20294967,
  "SoftTimeout": 129600,
  "GracePeriod": 1000,
  "Task1": {"TaskName": "ProdMinBias", 
                 "MaxRSS": 1234},
  "Task2": {"TaskName":  "DIGIPROD1" :
                 "MaxVSize: 3456},
  "TaskChain": 2
   }
```
and

```
"MaxRSS": {"ProdMinBias" : 2000,
                     "DIGIPROD1" : 3000,
                     "RECOPROD1" : 2500}
```

PR #5956, opened 2015-06-03
Fix Gfal2 plugin to use proxy and register module

juztas:
I have tested this at CERN, but seems with castor and adler32 it always fail with Segmentation fault.
Without adler32 it works as expected, and as far as I understand, CERN will not change to gfal2.
Tapas, I have agent fully set up with this patch, and let me know if you have some time to test.
Everything what was needed for me is to put site-local-config.xml in afs directory and run jobs at CERN. Can you put somewhere modified site-local-config.xml that it is accessible in WN?
Changes which I did to this file are:

```
   <local-stage-out>
     <command value="gfal2"/>
     <option value=""/>
     <catalog url="trivialcatalog_file:/afs/cern.ch/cms/SITECONF/T2_CH_CERN/PhEDEx/storage.xml?protocol=srmv2"/>
```

### DIFF
--- a/src/python/WMCore/Storage/Backends/GFAL2Impl.py
+++ b/src/python/WMCore/Storage/Backends/GFAL2Impl.py
@@ -20,7 +20,8 @@ class GFAL2Impl(StageOutImpl):
 
     def __init__(self, stagein=False):
         StageOutImpl.__init__(self, stagein)
-
+        self.removeCommand = "env -i X509_USER_PROXY=$X509_USER_PROXY gfal-rm -vvv %s"
+        self.copyCommand = "env -i X509_USER_PROXY=$X509_USER_PROXY gfal-copy -t 2400 -T 2400 -p -vvv"
 
     def createSourceName(self, protocol, pfn):
         """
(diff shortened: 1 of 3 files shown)


hufnagel:
As long as you only need to test stageout settings, for WMAgent there is a environment variable WMAGENT_SITE_CONFIG_OVERRIDE where you can define a path to a custom site-local-config.
Again, this doesn't concern CMSSW, which will use the regular SITECONF.

juztas:
This is what I did and pointed to my own SITECONF for CERN.  If we want to test it on Wisconsin, all I need is location to new site-local-config with change to use gfal2. @tsarangi said that they have gfal2 installed and this site is good for testing ;)

amaltaro:
Justas, how about you create a
deleteCommand = "env -i X509_USER_PROXY=$X509_USER_PROXY gfal-rm -vvv"
to have this command only in one place? Easier for maintenance :)

juztas:
fixed by @amaltaro comment

juztas:
tested with CRAB3 and direct stageout. It works! :)

amaltaro:
@ticoann don't merge it yet. Justas is going to run a few more tests at SPRACE and we (actually, Justas) need to sort out something at CERN, I guess related to adler32..

ticoann:
Thanks Alan, just let me know when it gets tested

juztas:
From my side good to go. Tested on T2_BR_SPRACE with 500 jobs and tested with CRAB3 on site which is accepting only gfal2
Request: jbalcas_MonteCarlo_150615_102600_1935

ticoann:
Thanks Justas, I am merging it, then.

PR #5966, opened 2015-06-04
Run func/methodattrs mixups

ericvaandering:
Another part of the python 2.7/3.x cleanup, this one having to do with special functions in the python data model


ticoann:
I am not sure who is using PlotFairy or it is ever used but merging it.

PR #5967, opened 2015-06-04
remove Memory args for TaskChain basic argument

ticoann:
also typo


ticoann:
tested in vm merging it

ericvaandering:
Age old manager trick. Do something wrong so that someone will do it right.

ticoann:
Eric, it is not your fault. It is completely my fault I added to check Memory parameter setting but completely messed up since it doesn't have to be set in TaskChain.

PR #5972, opened 2015-06-05
fix for reqmgr2 script

ticoann:
add reqmgr2 assignment script and add option for inject-test-wf script.
setting option --reqmgr2 will inject only reqmgr2 db using reqmgr2 REST API.


ticoann:

PR #5973, opened 2015-06-08
make workqueue handle both reqmgr2 and reqmgr

ticoann:
@amaltaro, Alan, please review.

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
+++ b/src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
@@ -5,6 +5,7 @@
 from WMCore.Services.RequestManager.RequestManager import RequestManager
 from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
 from WMCore.Services.LogDB.LogDB import LogDB
+from WMCore.ReqMgr.DataStructs import RequestStatus
 from WMCore.WorkQueue.WorkQueueExceptions import WorkQueueWMSpecError, WorkQueueNoWorkError
 from WMCore.Database.CMSCouch import CouchError
 from WMCore.Database.CouchUtils import CouchConnectionError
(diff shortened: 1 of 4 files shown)


amaltaro:
Looks good to me, though I have a couple of questions/comments. Will make it directly in the code.

src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
```diff
@@ -279,8 +282,13 @@ def _workQueueToReqMgrStatus(self, status):
                          'Done' : 'completed'
                          }
         if status in statusMapping:
+            # if wq status passed convert to reqmgr status
             return statusMapping[status]
+        elif status in RequestStatus.REQUEST_STATE_LIST:
```
amaltaro: Do we ever pass ReqMgr status to this method? If not, then we do not need to validate against ReqMgr2 statuses.

src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
```diff
@@ -251,11 +249,16 @@ def reportRequestStatus(self, request, status, message = None):
         """
         if message:
             self.sendMessage(request, str(message), 'info')
```
amaltaro: What is this message? When is it used? Is it the messages we see in ReqMgr when there is some failure? Depending on the answer, we could use logdb for this one as well :)

ticoann: sendMessage actually log to logDB it is wrapper function for reqmgr1 and reqmgr2 common calls.

src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
ticoann: It does. line 180 above. Thanks for double checking.

PR #5974, opened 2015-06-09
repack jobs should always use a single core in CMSSW

hufnagel:
Tier0 uses an override to run multicore repack jobs to take advantage of multiple 20GB local disk slots. We do not want to run the actual repacking multicore though, add an override that forces repack jobs to only use a single core in CMSSW.


hufnagel:
@ticoann , @ericvaandering 

Bit funny way to implement it, first explicitly switch repacking to use multicore and then disable it again at runtime. Problem is this was the only clean way to do it because the information for how many core are needed is only available at job splitting and the only way to pass it is as a normal multicore job.

We don't currently support to set numberOfCores for glideinWMS and numberOfCores for CMSSW separately in a generic fashion, plus I am not sure this would be all that useful either in the long run.

ericvaandering:
Looks OK to me. I thought this is what was happening already (you were just lying to condor).

PR #5976, opened 2015-06-09
Added force sites to Down; RC improvements related to condor

amaltaro:
Hey @ticoann @lucacopa , I wanted to share with you this patch ASAP, it's not final yet, so please do not merge.

These are the main changes:
- removed forcedSiteList (just saw there is a leftover..., in principle nobody is using it)
- only change site state when there is actually a change (so fetching much less condor classAds now)
- only change site thresholds (and tasks) when there is actually a change
- minimum limit for site thresholds is set to 50
- minimum limit for taskIO thresholds is set to 25

The only thing left - I guess - is that I want to make site thresholds to CPUBound+IOBound, to always have room to submit merge/IO jobs. Of course, some logging cleanup is also needed, I want to run some final tests.
Feel free to review and make suggestions, I shall finish it tomorrow.


amaltaro:
I updated this PR. It should be the final version now.
I'm running the validation of the new agent (1.0.7.pre13) with this patch in.

PR #5977, opened 2015-06-09
fixes abort and force-complete for reqmgr2

ticoann:
Need to be called by reqmgr2 api. Or reqmgr2 GUI

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -25,6 +25,7 @@ from WMCore.ReqMgr.Utils.Validation import validate_request_create_args, \
                validate_request_update_args
 
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
+from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
 
 class Request(RESTEntity):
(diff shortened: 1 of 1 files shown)


ticoann:
tested in vm. @amaltaro, Alan please review.

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -34,6 +35,7 @@ def __init__(self, app, api, config, mount):
         self.reqmgr_db_service = RequestDBWriter(self.reqmgr_db, couchapp = "ReqMgr")
         # this need for the post validtiaon 
         self.reqmgr_aux_db = api.db_handler.get_db(config.couch_reqmgr_aux_db)
+        self.gq_service = WorkQueue("%s/%s" % (config.couch_host, config.couch_workqueue_db))
```
amaltaro: Minor comment, but :) why do you append these variables? Can't you simply pass the host and db to be used?
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WorkQueue/WorkQueue.py#L13

ticoann: You are right, I was think we will all use url format with db name attached, but it looks silly. I will change that.

PR #5984, opened 2015-06-11
Break processing of jobs into smaller transactions

amaltaro:
It seems we were rolling back only in case of issues with the last worker(?)
@ticoann please review.

### DIFF
--- a/src/python/WMComponent/JobAccountant/JobAccountantPoller.py
+++ b/src/python/WMComponent/JobAccountant/JobAccountantPoller.py
@@ -5,20 +5,14 @@ _JobAccountantPoller_
 Poll WMBS for complete jobs and process their framework job reports.
 """
 
-
-
-
 import time
 import threading
 import logging
 
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
-
 from WMCore.Agent.Harness import Harness
 from WMCore.DAOFactory import DAOFactory
-
 from WMComponent.JobAccountant.AccountantWorker import AccountantWorker
-
 from WMCore.WMException import WMException
 
 class JobAccountantPollerException(WMException):
@@ -34,8 +28,7 @@ class JobAccountantPoller(BaseWorkerThread):
     def __init__(self, config):
         BaseWorkerThread.__init__(self)
         self.config = config
-        self.accountantWorkSize = getattr(self.config.JobAccountant,
-                                          'accountantWorkSize', 100)
+        self.accountantWorkSize = getattr(self.config.JobAccountant, 'accountantWorkSize', 100)
         # initialize the alert framework (if available - config.Alert present)
         #    self.sendAlert will be then be available
         self.initAlerts(compName = "JobAccountant")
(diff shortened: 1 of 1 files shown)


src/python/WMComponent/JobAccountant/JobAccountantPoller.py
```diff
@@ -70,43 +63,31 @@ def algorithm(self, parameters = None):
...
 
-        while len(completeJobs) > self.accountantWorkSize:
+        while len(completeJobs) > 0:
             try:
                 jobsSlice = completeJobs[:self.accountantWorkSize]
                 completeJobs = completeJobs[self.accountantWorkSize:]
                 self.accountantWorker(jobsSlice)
+                if len(completeJobs) <= 100:
+                    self.accountantWorker(completeJobs)
+                    break
```
ticoann: I think these three lines just need to be removed. (from 74 to 76). Considering initial completeJobs = 90

src/python/WMComponent/JobAccountant/JobAccountantPoller.py
```diff
@@ -70,43 +63,31 @@ def algorithm(self, parameters = None):
...
+                logging.info("Remaining completed jobs to process: %d" % len(completeJobs))
             except WMException:
+                myThread = threading.currentThread()
+                if getattr(myThread, 'transaction', None) != None:
+                    myThread.transaction.rollback()
```
ticoann: Actually it seem it is better to rollback in AccountWorker.py  __call__ method since explicit transaction start and commit happens there

ticoann:
I am not sure I completely understand the code but transaction start here 
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobAccountant/AccountantWorker.py#L278
and commits here
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobAccountant/AccountantWorker.py#L336

So the patch won't really break down the transaction. it would have the same effect. (So title of the issue is a little confusing). But I think what you said about rolling back is the correct observation. Did you experience any problem with this?

src/python/WMComponent/JobAccountant/JobAccountantPoller.py
amaltaro: I don't know why I thought the last slice would remain in the loop forever.. I'll remove it.

amaltaro:
I agree the title is confusing, sorry. When I started working on this patch, I thought we were calling the last worker with the full list of jobs, which was not true but then I didn't amend the commit :( Nope, there was no problem with these transaction from what I saw, in the end I basically removed the dup code.

About the transaction, keeping it where it's now seems more clear IMO, otherwise we'd need to add a try/except for the whole call method in AccountWorker basically. What do you think?

ticoann:
Ah, I see I thought you are trying to break down the transaction. 

The reason I think AccountWorker "__call__" method contains rollback is better was that I couldn't find where the transaction starts and commit happens until I looked at AccountWorker module.

I think in general, for the roll back function, it is better to have that with in the same function when transaction begins and committed. Otherwise, whenever some routines call that function, we  will need to try and catch and do the roll back (and has to remember to do that as well).
However, in this particular case there is only one call is made (I mean it is just in one place - not counting the loop) so if you think it is more clearer. That would be fine with me.

amaltaro:
I agree with you, however, for now I kept it in the JobAccountant to make sure I won't break what is working :-)
BTW, there is this alert framework that I know nothing about, maybe we can chat later today. Patch is updated.

ticoann:
Thanks, Alan patch looks good. Merging it.

PR #5986, opened 2015-06-12
Log db view

vkuznet:


vkuznet:
@ticoann, I added logdb.js function and show how to use it in example2.html. Now I think it can be plugged into wmstats page (possibly we need to add overlap CSS elsewhere that this div element will overlay any other html content).

PR #5987, opened 2015-06-13
fix open block upload in DBS

ticoann:
fixes uploading open dbs blocks when dbs insert block call fails


ticoann:
@amaltaro, please review. I will create the new patch for 1.0.5_wmagent and 1.0.6_wmagent

amaltaro:
And now we are safe :-) Good to me! Thanks

PR #5958, opened 2015-06-03
Merging

alexanderrichards:
@ticoann note this is not ready to merge yet, just here to collect all patched in one place

Issue #5340, opened 2014-09-16
Add PNN to JDL


ticoann:
When submit the job the location has to be added in JDL.
This need further discussion since we only assign PSN currently to jobs

alexanderrichards:
@ticoann this is in the PNN patch so can be closed

ticoann:
Thanks Alex

PR #5989, opened 2015-06-15
PNN transition combined patch (Ready to be merged)

alexanderrichards:
updated version of #5958 rebased on master

Note this is not ready to me merged yet as we are still debugging an issue.


src/python/WMComponent/JobCreator/JobCreatorPoller.py
```diff
@@ -160,6 +164,10 @@ def creatorProcess(work, jobCacheDir):
         swVersion    = work.get('swVersion', None)
         agentNumber  = work.get('agentNumber', 0)
         numberOfCores = work.get('numberOfCores', 1)
+        inputDataset = work.get('inputDataset', None)
+        inputDatasetLocations = work.get('inputDatasetLocations', None)
+        logging.warning("ALEX JobCreator 5: inputDataset %s" % inputDataset)
```
ericvaandering: Make sure these are eventually removed, even if it's some time after this goes into production. If you want to put it into production like this, please open a ticket now to remind yourself.

alexanderrichards: I will remove these debugging lines once this is working. I have made this PR to allow Seangchan to easily merge the patches into his testing branch so he can help me debug a problem. I'll open a ticket though ;-)

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -548,64 +549,66 @@ def lfnsInBlock(self, fileBlockName):
...
 
-        blockInfo = {}
+        locations = {}
+        node_filter = set(['UNKNOWN', None])
         if not dbsOnly:
             try:
-                blockInfo = self.phedex.getReplicaSEForBlocks(phedexNodes=phedexNodes, block=blockNames, complete='y')
-            except Exception as ex:
+                blocksInfo = self.phedex.getReplicaPhEDExNodesForBlocks(block=fileBlockNames, complete='y')
+            except Exception, ex:
```
ericvaandering: Exception as ex is preferred (and now enforced in our code base)

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -548,64 +549,66 @@ def lfnsInBlock(self, fileBlockName):
...
             try:
-                for block in blockNames:
-                    res = self.dbs.listBlockOrigin(block_name = block)
-                    if res:
-                        blockInfo[block] = [res[0]['origin_site_name']]
-            except dbsClientException as ex:
-                msg = "Error in DBS3Reader: self.dbs.listBlockOrigin(block_name=%s)\n" % fileBlockName
+                for block in fileBlockNames:
+                    blocksInfo.update( dict(((block, blockInfo[0]['origin_site_name']) for blockInfo in self.dbs.listBlockOrigin(block_name = block) if blockInfo)) )
+            except dbsClientException, ex:
```
ericvaandering: as ex here too

src/python/WMCore/Services/DBS/DBS3Reader.py
alexanderrichards: oops yes this has come from rebasing my changes onto the head looks like it's a change I've introduced. I'll ensure that this is changed back before this branch is pulled.

src/python/WMCore/Services/DBS/DBS3Reader.py
alexanderrichards: ditto :smiley:

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -314,23 +315,44 @@ def refreshCache(self):
                 possibleLocations = set()
 
                 # all files in job have same location (in se names)
-                rawLocations = loadedJob["input_files"][0]["locations"]
+                #rawLocations = loadedJob["input_files"][0]["locations"]
+                #rawLocations = loadedJob.get("inputDatasetLocations") or []
+                #possibleLocations.update(rawLocations)
+                from pprint import pformat
+                rawLocations = loadedJob.get("inputDatasetLocations") or []
```
ticoann: inputDataLocations only set for initial input dataset is set when the jobs comes back and location is set from fwjr this will be empty.

alexanderrichards: but we know that the PNN now comes back with FWJR so locations should be picking this up but is not. I guess this will be solved with the resolution of #6004

ticoann: Yes I think all the code creating inputDatasetLocations and inputDataset need to be reverted and just use

``` python
rawLocations = loadedJob["input_files"][0]["locations"]
```

ericvaandering:
retest this please

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
alexanderrichards: commit 58ef77570f5759b66c1e4d46262a91b50afed21e fixes #6004

bin/wmagent-resource-control
```diff
@@ -109,10 +109,11 @@ def getCMSSiteInfo(pattern):
     for seName in siteDB.cmsNametoList(pattern, "SE", "result.json"):
         cmsNames = siteDB.seToCMSName(seName)
```
ticoann: I think this misses cmsNames FNALDisk some how.
https://cmsweb.cern.ch/sitedb/data/prod/site-names
We should create the function directly maps psn - pnn. using above call. I think still site db need to be updated. I will check with Diego and Nicolo.

src/python/WMCore/Services/DBS/DBS3Reader.py
alexanderrichards: commit bd7044ecbb061ac523b7f738468f84d8f0eacffc fixes this

src/python/WMCore/Services/DBS/DBS3Reader.py
alexanderrichards: commit bd7044ecbb061ac523b7f738468f84d8f0eacffc fixes this

src/python/WMComponent/JobCreator/JobCreatorPoller.py
alexanderrichards: commit 0f8dea95017feea9a3d290f893b857edf2f4d6fc fixes this

ticoann:
Hi Alex, are we still using inputDataset, inputDatasetLocations? 
It seems we are not using those due to the reason we discussed. (file location != dataset locaction)
Could you clean up inputDataset and inputDatastetLocations?

ticoann:
The reason submit failed in T1_US_FNAL, there is no mapping between T1_US_FNAL_Buffer and T1_US_FNAL in the call. (*)

https://cmsweb.cern.ch/sitedb/data/prod/data-processing 
I will ask Diego, Nicolo.

Seangchan

(*)
2015-07-22 09:21:28,306:WARNING:JobSubmitterPoller:ALEX_JobSubmitter3.2: rawLocations= set(['T1_US_FNAL_Buffer'])
2015-07-22 09:21:28,307:WARNING:JobSubmitterPoller:ALEX_JobSubmitter3.3: rawlocations(input_files)= set(['T1_US_FNAL_Buffer'])
2015-07-22 09:21:28,307:WARNING:JobSubmitterPoller:ALEX_JobSubmitter3.5: rawlocations(inputDatasetLocations)= None
2015-07-22 09:21:28,643:WARNING:JobSubmitterPoller:ALEX_JobSubmitter5: possibleLocations= set([])

alexanderrichards:
Hi Seangchan,
Those were put into the code for the classad. It was for one of the issues that Eric raised, see https://github.com/dmwm/WMCore/issues/5340

Should I still delete it?
Cheers
Alex

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -315,16 +315,25 @@ def refreshCache(self):
 
                 # all files in job have same location (in se names)
                 rawLocations = loadedJob["input_files"][0]["locations"]
+                
+                sitedb = SiteDB()
+ #               sitedb_names = set()
+                for l in rawLocations:
+#                    sitedb_names.update(sitedb.PNNtoPSN(l) or [])
+                    possibleLocations.update(sitedb.PNNtoPSN(l) or [])
```
ticoann: One problem getting the mapping from sitedb at this point, if sitedb goes down temporary or there is network problem.
The job submission will fail. I think we should get this information from wmbs_location_senames table instead.
We can create the memory cache as well we don't create call db every time.

alexanderrichards: Once we rename wmbs_location_senames  to wmbs_location_pnn then this will only be the pnn. I think that we still need to map that to the PSN at this point right?

--Alex

bin/wmagent-resource-control
ticoann: actually we can call siteDB.PSNtoPNN call directly to get this mapping. (Although the list is not complete there but I think site db need be updated for that)
under lining call is 
https://cmsweb.cern.ch/sitedb/data/prod/data-processing

hufnagel:
@ticoann , T1_US_FNAL_Buffer is not in the mapping because _Buffer locations are not valid input locations for processing. Submission for data on _Buffer should fail.

ticoann:
> @ticoann , T1_US_FNAL_Buffer is not in the mapping because _Buffer locations are not valid input locations for processing. Submission for data on _Buffer should fail.

Actually it is not submitted to the _Buffer. This patch gets the PNN from FWJR (site conf) when jobs ran,
So the processing job stageout to _Buffer and when it create the merge job to submit to site, it need to change to processing site from PNN.

And it looks this sitedb api. to get that map.
https://cmsweb.cern.ch/sitedb/data/prod/data-processing

However there is no PhEDEx name - T1_US_FNAL_Buffer. So it can't get the processing site name back. I was just wondering why there is no map between T1_US_FNAL_Buffer (PNN) and T1_US_FNAL (PSN)

hufnagel:
Processing jobs don't stageout to _Buffer. They stageout to _Disk.

ticoann:
> Processing jobs don't stageout to _Buffer. They stageout to _Disk.

Then there must be a problem getting the location information from fwjr (siteconf). It might be getting the location where the jobs ran instead of the stageout location.

ticoann:
Hi, Dirk
Actually, it is getting file location not job location.
https://github.com/dmwm/WMCore/pull/5989/files#diff-65c9ca1a1a49a1aad291e5b964e36d94R443
https://github.com/dmwm/WMCore/pull/5989/files#diff-b8ec91dd68e7734817b4844183f62243R121

And that returns T1_US_FNAL_Buffer instead of T1_US_FNAL_Disk.

This seems the call actually sets the PNN value. 
https://github.com/dmwm/WMCore/pull/5989/files#diff-b8ec91dd68e7734817b4844183f62243R212

Does this mean site conf is not correctly configured or we have to call some other function to get the correct stageout place?

hufnagel:
A possible explanation could be that Fermilab has a really screwy siteconf setup. There is a site T1_US_FNAL, which has only tape and a site T1_US_FNAL_Disk, which has only disk. No other T1 does that, they all only have a single site with tape and disk managed together.

hufnagel:
Btw, that code can never work. The parameter name in the siteconf isn't 'pnn', its 'phedex-node'.

ticoann:
> Btw, that code can never work. The parameter name in the siteconf isn't 'pnn', its 'phedex-node'.

Hmm, that is weird it returned 'T1_US_FNAL_Buffer'. Let me change and test again.

amaltaro:
Seangchan, if this location is being taken from the site-local-config.xml, then we should be seeing T1_US_FNAL_Disk instead. This is their configuration
https://cmsweb.cern.ch/gitweb/?p=siteconf/.git;a=blob;f=T1_US_FNAL_Disk/JobConfig/site-local-config.xml;h=44566c0b0c1fe0c1133c3de161d3b344814ca5ec;hb=HEAD

I'd suggest you to connect to a FNAL worker node and check which file/path the job is opening. Then you x-check the phedex-node available in that file.

ticoann:
Thanks Alan, How I can access worker node? I can't access directly using ssh.  It is only allowed when job is running using  condor_ssh_to_job?

ticoann:
Actually there seems to be conversion happened here (phedex-node -> pnn).

https://github.com/dmwm/WMCore/pull/5989/files#diff-d338938ebf2a8cae6acee606f31ef559R280
https://github.com/dmwm/WMCore/pull/5989/files#diff-d338938ebf2a8cae6acee606f31ef559R303

So that is why we probably get the value.
Alex, is it hard to change back the name to phedex-node instead of pnn in fwjr and localConfig? It seems all other params keep the original key. 

Thanks

amaltaro:
To access FNAL worker nodes, you need to:
 1. run these jobs from a FNAL agent (feel free to take cmssrv113 if you want)
 2. condor_ssh_to_job JOBID

Unless this conversion you mentioned is what is messing with the mapping :-)

bin/wmagent-resource-control
alexanderrichards: commit ef9c9fb846f0c1041879630d48d63de3c935888f fixes this

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
alexanderrichards: commit bbe2c1d7f01ed29c0f89a9535b508fdefeb427a4 fixes this

alexanderrichards:
> Alex, is it hard to change back the name to phedex-node instead of pnn in fwjr and localConfig? It seems all other params keep the original key.

commit 3e5d2994e4618dc9b623e840d5376e2e18c09f21 fixes this

ticoann:
@alexanderrichards, Hi Alex, sorry I had to merge other patches and that cause the conflict on your patch. Most likely with this patch. https://github.com/dmwm/WMCore/pull/6096
Could you update this patch? Since this patch has a lot of changes in a lot of files. I think we should try to merge this sooner than later. I will help you with unittest fix.

ericvaandering:
retest this please

(trying to get you an updated list of failures, @alexanderrichards )

ericvaandering:
retest this please

ericvaandering:
retest this please

bin/wmagent-resource-control
```diff
@@ -95,32 +95,19 @@ def createOptionParser():
...
-
+    pattern = pattern.replace('*', '.*').replace('%', '.*')
     if options.addDisk:
-        nameSEMapping = dict((k, list(set(v))) for k, v in nameSEMapping.iteritems()
-                        if not k.endswith("_Disk"))
-    else:
-        nameSEMapping = dict((k, list(set(v))) for k, v in nameSEMapping.iteritems())
-
-    return nameSEMapping
+        return SiteDBJSON().PSNtoPNNMap('^(?!.*_Disk\s*$)' + pattern)
```
ticoann: Shouldn't this be for opposite condition options.addDisk == False?
I think we always set the addDisk options True now so it might not be needed to check this condition at all.

alexanderrichards: The logic hasn't changed from before, it's just set up in a strange way such that if on --addDisk is given then it defaults to True otherwise it is False. Hence the backwards looking logic but it does work. I didn't change the logic as that wasn't crucial to the patch that I'm  implementing also I assumed that there must have been a reason it was done in this backwards way.

ticoann:
retest this please

ticoann:
retest this please

ericvaandering:
retest this please

ericvaandering:
@alexanderrichards most of the suggested patch stuff (isinstnace) is not on you and I would suggest NOT changing that. This branch is already a major change and to conflate it with possible type problems could really muddle things.

The only parts you have to change are the first two: future imports on NEW files and the one has_key -> in

alexanderrichards:
@ericvaandering thanks eric, I wasn't looking forward to it coming back with more of those syntax changes ;-). I'll remove the last commit then with the exception of the NEW files __future__ imports and the has_key which I notice is changed in the master upstream but as it's a modified line in my patch this needs to be carried forward.

ericvaandering:
retest this please

ericvaandering:
retest this please

alexanderrichards:
retest this please

alexanderrichards:
retest this please

ericvaandering:
Looks like this is improving.

alexanderrichards:
@ericvaandering I'm getting there, I think there were like over 200 or so at the start ;-)

alexanderrichards:
retest this please

alexanderrichards:
retest this please

mmascher:
Quick question about this `DESIRED_CMSDatasetLocations` https://github.com/alexanderrichards/WMCore/commit/fc3c8ad5d714178dc86534c552067f42a37e9214#diff-ea1f7bf61ef33fe7102fc3142a26fa56R1038

Are'nt these pnn's associated to blocks intead of dataset? Should we change the name of the classad to something respecting the truth (and less misleading) as `DESIRED_CMSBlockLocations`?

ericvaandering:
That seems smart to me. @jamesletts want to comment? I presume nothing in GlideinWMS has started with this, so we can modify the name as we wish as long as CRAB and WMAgent agree.

amaltaro:
Can someone explain the purpose of this new classAd. How will it be used by glideins?

belforte:
@amaltaro the purpose is described e.g.in the CRAB issue Marco referred to
https://github.com/dmwm/CRABServer/issues/4856
in one line: make data location available for HTC so to optimize scheduling decisions

juztas:
Does WMA plan to use overflow ?

ericvaandering:
It does not matter what WMA plans to do or does not plan to do. This would give GlideinWMS a way to do something *should it choose to* . Perhaps we will never use this, but we sat down and worked out a comprehensive plan which this is a part of and there is no reason to backtrack on it.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
```diff
@@ -563,6 +571,13 @@ def pollSubscriptions(self):
 
 
                 # Assemble a dict of all the info
+                inputDataset = None
+                locations = None
+                dbsurl = wmTask.inputDatasetDBSURL()
```
ericvaandering: This is where your problem comes in. wmTask in the unit test has that weird string for DBS which is invalid. I guess the changes here are related to the changes for Condor? 

I'll keep looking, but I'm trying to understand why you added this code in the first place.

I suspect just changing the string will make everything go smoothly.

ericvaandering:
#6257 shows my patch to your branch had no unintended effects.

ericvaandering:
Let me know if you have other unit tests that don't seem to be related to the SE->PNN stuff. I'm happy to help get this out the door.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
alexanderrichards: I don't recall now why the code here is as it is. Your right that the changes were related to the changes in Condor. When you suggest changing the string, to what should it be changed, the dbsurl used in production? I assume that the wmTask in unittest has a dummy string in order to decouple from actually using the production url right?

alexanderrichards:
@ericvaandering Thanks for this and please ignore my in-line comment. I addressed these posts in the order they came into my email so dealt with it before I saw your next posts :smile:

ericvaandering:
This is looking good. Now the only transitions are "error to failure" Maybe this is good to merge or do we need to investigate the error to failure issues?

alexanderrichards:
@ericvaandering, @ticoann I will differ to Seangchan to answer this as he said he was looking at the Workqueue tests.

amaltaro:
Sorry for my stupid question, but is it a breaking change meaning that as soon as it gets merged, we must migrate to PNN/PSN mode? Just wondering whether it should be merged into master or a new branch would be better.

ticoann:
Thanks Alan for reminding us to be careful. As far as I remember, i think this change will only affects the agent but I think we should test against current testbed to make sure whether it breaks anything or not.
Since the new testbed deploy deadline is today, I think it is better to wait for the merge until we make sure it is compatible with current reqmgr, workqueue, tier0, etc.

ticoann:
And thanks Alex for completing this big change.

amaltaro:
The list of files changed does not open in my laptop/browser :-) but AFAIR there are database changes, risn't that right @alexanderrichards ? Or did you manage to deploy the current production agent version and simply applied this patch?
Ok Seangchan, I'll run our battery of tests this week.

alexanderrichards:
the tables and keys in the database remain the same but the content now have pnns.

amaltaro:
Interesting, I received an email about a bunch of pylint warnings (like"Line 13 W0611 Unused import os") but that message is not here in GH.

Alex, once you get this patch in a final state, could you please go through the pylint warnings and clean up these unused imports (at least)? This is another case that deserves clean up here "Line 2 E0012 Bad option value 'W6501'".

alexanderrichards:
@amaltaro the pylint warnings did show up in this PR but I just removed the commend as I wanted to focus on the testing and it was impossibly long. As you say I have it on email anyway.

amaltaro:
Ok, thanks!

alexanderrichards:
FYI these unused imports in many places are not to do with this patch. Do you still want me to go through and fix them or is this best done as a separate task?

amaltaro:
Pylint will create a report per file, regardless whether you changed that piece of code or not :-)

If that's not too many files (I doubt!), I think it would be a good opportunity to push cleaner code upstream (and this patch is probably catching quite a lot of them).

ericvaandering:
I think this patch is so big and intrusive already, and has taken so long, that we don't want to continue to delay it with trivialities like imports. Those things are really harmless. Maybe we can save the message, and make a follow up issue that can be addressed by some house cleaning patches.

amaltaro:
Yep, sounds better indeed. Just created an issue not to forget it.

ticoann:
Alex, can I merge this now? Are you adding DESIRED_CMSDataset patch here? If you haven't added here we can have a separate patch. Just let me know. Thanks

alexanderrichards:
@ticoann The classad patch is included in this one. I think that this can be merged now. Eric did mention wanting Alan to run a final test but as far as I'm concerned it's ready.

ticoann:
Thanks Alex, I will merge this. My hands are shaking. :smile:

ericvaandering:
LOL

alexanderrichards:
your not alone :fearful:

hufnagel:
This is triggering many rebase demands for other pull requests (as expected).

I am not going to work on that until I am sure this patch will stay in, so we need some fast validation/feedback. Maybe not to the level "this works to 100%", but to the level "this works well enough that we keep it merged and will fix the remaining issues".

ericvaandering:
Fair enough. 

I did a check of the ~10 most recent pull requests and 3 need to be re-based.

amaltaro:
So you merge and I handle the bomb?!? :-P
Seangchan, can you cut a new tag please. Then I'll build an RPM and we can do some functional and wider testing.

ticoann:
Thanks, Alan. I was planning to run some test today on vocms008. So don't over work in the evening. :-) I will ask your help if I had some trouble on the test. Tag is created (1.0.11.pre2)

ticoann:
> This is triggering many rebase demands for other pull requests (as expected).

Sorry about that. I will run the test as soon as possible and let you know. (Since the fix after initial testing was mostly for the unittest fix, I won't expect any big problems)

ericvaandering:
Hey, @alexanderrichards had to rebase many times because of the rest of us. A bit of payback is totally justified. :-)

hufnagel:
I don't mind the work (its normal and expected), just only want to do it if we are certain this will not get unmerged again...

amaltaro:
Thanks Seangchan! Once you are done with DMWM tests, if things are ok, we can declare this patch good. Then I'll run a larger test in several sites (>10).

ticoann:
Unfortunately, all the log collect job came as failure which is my fault.
I create the patch and running the test again. 
https://github.com/dmwm/WMCore/pull/6318

Other than that it seems fine, Dirk, if you want you can go ahead and rebase your patch. Thanks

ericvaandering:
In case it was missed, there is an issue with the Tier0 and this code reported here: https://hypernews.cern.ch/HyperNews/CMS/get/wmDevelopment/719/1/1.html

To my amateur eyes, it looks like an instance where the MySQL code was updated but the corresponding Oracle code was not. @alexanderrichards can you have a look at the corresponding Oracle code for all the MySQL bits that changed and make any changes needed? I imagine @hufnagel will be looking at this too. 

If it's as simple as it appears to be, maybe it will all be solved before @ticoann and I wake up in the morning.

ticoann:
Eric, Alex, I am so sorry the I missed this problem. I should have caught this earlier. Anyway, the change seems pretty simple. So I made the patch. (But couldn't run the test yet - I didn't set up the Oracle version of agent)
https://github.com/dmwm/WMCore/pull/6320
Hopefully Dirk can run some test by the morning.

amaltaro:
I completely forgot the oracle version as well! I'm getting the pre2 version + eos + oracle patch and I will setup an oracle instance soon

ticoann:
Thanks Alan. I think Jenkins should set Oracle test as well, if possible.

ericvaandering:
We can try that. It will double the testing time and I would need to know how to configure it to do such a thing. I don't know if @PerilousApricot had that working before or not.

ticoann:
I think if we have to choose, I would choose Oracle test in Jenkins, since most individual unittest uses MySQL (MariaDB) and most of the agents are using Oracle (except fnal agents).

PR #5998, opened 2015-06-15
increase event limits for PromptReco

hufnagel:
Increase the max allowed number of events in a lumi before a job is immediately set to CreateFailed.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
@@ -259,7 +259,7 @@ class PromptRecoWorkloadFactory(StdBase):
                 self.eventsPerJob = int((8.0 * 3600.0) / self.timePerEvent)
             self.procJobSplitArgs["events_per_job"] = self.eventsPerJob
             if self.procJobSplitAlgo == "EventAwareLumiBased":
-                self.procJobSplitArgs["max_events_per_lumi"] = 20000
+                self.procJobSplitArgs["max_events_per_lumi"] = 100000
         elif self.procJobSplitAlgo == "LumiBased":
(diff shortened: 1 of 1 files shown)


hufnagel:
Please do not merge, we have to evaluate how this behaves in production first.

ticoann:
@lucacopa, Hi Luis, can this be merged, is this causing any problem?
@amaltaro, Alan will this cause any problem in our side?

amaltaro:
It should affect only PromptReco, so from our side it's Ok.

lucacopa:
@ticoann I haven't seen problems so far, but I think we should move forward to make this event limit configurable.

hufnagel:
It is configurable, at least as far as the job splitter code is concerned. You cannot currently set it as a spec parameter, but that's easy to add. The more interesting question is, which spec will actually need it and how is this going to be configured. If you just set it to a very large number independent of the input conditions anyways, it doesn't really matter where you hardcode things.

So for PromptReco, how do you want to configure this ? Tier0 config setting per dataset ? Even that isn't enough, because I really don't want Tier0Ops to maintain these limits, they would have to be set by Physics. In all lilkelyhood we would again just get one large number for everything, at which point this would all be a big effort for nothing.

lucacopa:
I was thinking about having a default and only use the configuration in case we are requested to increase that limit. As in the low-lumi runs with 2.5 kHz, this request was only for some datasets. 
I wonder if hardcoding it to 100k may end on 'really' long jobs when the luminosity rises.

amaltaro:
For central production, it seems there is not that much gain in making it configurable, since nobody
ever mentioned about it's being too small or too big.
So whatever you guys decide and develop, it's fine with me (maybe a configuration overwritten by an agent parameter...)  :-)

PR #6003, opened 2015-06-16
clean up dbs3 configuration code

ticoann:
This still need to be tested. This patch removed some codes for the more clarity of the code.
1. dbs3UploadOnly option and code using it. (It is always False now and never True).
2. status3 is removed from dbsbuffer_block schema(not used anymore)
3. InGlobalDBS status is removed, 'InDBS' is what is used now.

DBS block transition follows. 
Open (block is created in agent) -> Pending (block is closed in agent) -> InDBS (closed block is uploaded in dbs3) -> Closed (block is closed in PhEDEx) (PhEDEx injection started before block is closed)

Still need to do 
DBSBuffer and DBSUpload package need to be removed as well as unittest code, Fix the unittest for PhEDExInjector which still uses DBSBuffer

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -94,7 +94,7 @@ config.section_("General")
 config.General.workDir = workDirectory
 config.General.logdb_name = logDBName
 config.General.logdb_url = "%s/%s" % (couchURL, config.General.logdb_name)
-config.General.central_logdb_url = "need to get from secrete file"
+config.General.central_logdb_url = "need to get from secrets file"
 
 config.section_("JobStateMachine")
 config.JobStateMachine.couchurl = couchURL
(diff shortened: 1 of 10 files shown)


ticoann:
@amaltaro, Alan please review

etc/WMAgentConfig.py
```diff
@@ -147,8 +147,7 @@
 config.DBS3Upload.workerThreads = 1
 config.DBS3Upload.pollInterval = 100
 #"https://cmsweb.cern.ch/dbs/prod/global/DBSWriter" - production one
-config.DBS3Upload.dbsUrl = "OVER_WRITE_BY_SECETES" 
-config.DBS3Upload.dbs3UploadOnly = False
+config.DBS3Upload.dbsUrl = "OVER_WRITE_BY_SECETES"
```
ericvaandering: We could choose to spell SECRETS correctly since we are messing with this.

ericvaandering:
Will be nice to clean this up.

etc/WMAgentConfig.py
ticoann: Thanks, updated the typo

etc/WMAgentConfig.py
```diff
@@ -147,8 +147,7 @@
 config.DBS3Upload.workerThreads = 1
 config.DBS3Upload.pollInterval = 100
 #"https://cmsweb.cern.ch/dbs/prod/global/DBSWriter" - production one
-config.DBS3Upload.dbsUrl = "OVER_WRITE_BY_SECETES" 
-config.DBS3Upload.dbs3UploadOnly = False
+config.DBS3Upload.dbsUrl = "OVERWRITE_BY_SECRETES"
```
ericvaandering: Um, it's secrets, not secretes. :-)

ticoann: Sorry, I didn't sleep well last night. :-) Updated

amaltaro:
Marvelous cleanup Seangchan! Patch looks good to me.

PR #6020, opened 2015-06-22
move setting the processing task memory into StdBase

hufnagel:
The default for memory, timePerEvent and sizePerEvent should actually be used in the base class and not HAVE to be overridden in every single spec.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
@@ -322,8 +322,10 @@ class TaskChainWorkloadFactory(StdBase):
                                               configCacheUrl = self.configCacheUrl,
                                               splitArgs = splitArguments, stepType = cmsswStepType,
                                               seeding = taskConf['Seeding'], totalEvents = taskConf['RequestNumEvents'],
-                                              forceUnmerged = forceUnmerged, timePerEvent = self.timePerEvent,
-                                              memoryReq = taskConf.get('Memory', None), sizePerEvent = self.sizePerEvent,
+                                              forceUnmerged = forceUnmerged,
+                                              timePerEvent = taskConf.get('TimePerEvent', None),
+                                              sizePerEvent = taskConf.get('SizePerEvent', None),
+                                              memoryReq = taskConf.get('Memory', None),
                                               taskConf = taskConf)
 
         # this need to be called after setpuProcessingTask since it will overwrite some values
@@ -400,9 +402,9 @@ class TaskChainWorkloadFactory(StdBase):
                                               splitArgs = taskConf["SplittingArguments"],
                                               stepType = cmsswStepType,
                                               forceUnmerged = forceUnmerged,
-                                              timePerEvent = self.timePerEvent,
+                                              timePerEvent = taskConf.get('TimePerEvent', None),
+                                              sizePerEvent = taskConf.get('SizePerEvent', None),
                                               memoryReq = taskConf.get("Memory", None),
-                                              sizePerEvent = self.sizePerEvent, 
                                               taskConf = taskConf)
         
         # this need to be called after setpuProcessingTask since it will overwrite some values
--- a/src/python/WMCore/WMSpec/StdSpecs/MonteCarlo.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/MonteCarlo.py
@@ -52,10 +52,7 @@ class MonteCarloWorkloadFactory(StdBase):
                                               configCacheUrl = self.configCacheUrl,
                                               seeding = self.seeding,
                                               totalEvents = self.totalEvents,
-                                              eventsPerLumi = self.eventsPerLumi,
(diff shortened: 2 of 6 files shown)


hufnagel:
@tiocann , please take a look.

The relevant change is in StdBase. All other changes are just cleanup of now redundant code.

hufnagel:
I guess the question here is:

Should the Memory spec parameter always apply to all processing tasks created via setupProcessingTask() unless something else is specified in the call, or should it only apply to the tasks were we pass a value to setupProcessingTask() and not be specified at all otherwise ?

I assumed the former, but actually implemented was the latter. This patch changes this.

amaltaro:
We should apply Memory to all the processing tasks *only* when there is no Memory parameter specified at the task level.

That means for TaskChain spec, we actually want to pass memory, timePerEvent and sizePerEvent when calling setupProcessingTask().

hufnagel:
Yeah, and I left these calls intact. They override whatever is specified at the spec level with 'Memory'. My concern is with the other case. If you pass nothing to setupProcessingTask, should it configure nothing or should it configure the spec default 'Memory' ?

The previous implementation was nothing, this patch changes it to the spec default 'Memory'.

Similar for timePerEvent and sizePerEvent.

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -400,9 +399,7 @@ def setupTask(self, task, taskConf):
                                               splitArgs = taskConf["SplittingArguments"],
                                               stepType = cmsswStepType,
                                               forceUnmerged = forceUnmerged,
-                                              timePerEvent = self.timePerEvent,
                                               memoryReq = taskConf.get("Memory", None),
-                                              sizePerEvent = self.sizePerEvent,
```
amaltaro: timePerEvent and sizePerEvent needs to passed in this call. The same applies to the call from setupGeneratorTask()

amaltaro:
If you pass nothing to setupProcessingTask(), then it means there are no taks level parameters and it should use the one provided in the workload (notice it is *not* their default values, but the value provided for the whole schema).

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
hufnagel: No, they don't. Look at the StdBase changes. Passing them here is redundant.

hufnagel:
Yes, I meant the spec level default you configured for the whole spec (not a code default).

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
amaltaro: Ok, finally got it! One thing that confused me is that we were passing the timePerEvent and sizePerEvent from the spec (instead of the one set in the task). The current code in production seems to be working fine, but weren't we supposed to pass
timePerEvent = taskConf.get('TimPerEvent', None)
instead? The same for sizePerEvent.

hufnagel: I have no idea. If the exist, then yes, it would make sense to pass those. Should I change the pull request accordingly ?

amaltaro: Yes, please. I can try to test it in my VM today, don't guarantee :-)

hufnagel:
@amaltaro , please take a look now, I adjusted the TaskChain to pass per task timePerEvent and sizePerEvent as well.

amaltaro:
looks good to me!

ticoann:
Hi Dirk this it a little weird. Without the patch it should do the same thing. It should set the default value from StdSpecs if the inherited spec doesn't specify the value.
Here it where it happens

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L72

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py#L2

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/ReReco.py#L152

And default argument is inherited from 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py#L40

then
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py#L40

So it should set the default values  if the child spec doesn't specify that.
I will try to check where it got broken.

hufnagel:
@ticoann , we are talking past each other. self.memory is of course defined in StdBase, the problem is that it's not actually USED anywhere to set the performance parameters for a processing task.

ticoann:
@hufnagel, Ah, so you don't want to explicitly pass those parameters in the setupProcessingTask method. Instead of passing those params in setupProcessingTask in Repack and Express, you want to clean up other specs, correct?  Yes, I think that make sense. Sorry for the confusion.

PR #6016, opened 2015-06-19
Wmstats logdb

vkuznet:

PR #6021, opened 2015-06-23
add LogDB tab and retrievie only request data

ticoann:
Refactor log db code in WMStats, contains #6016
1. support tier0
2. retrieve only active request not all the request.
3 separate data retriever and view creation


ticoann:
@vkuznet, Valentin, Please review. It is somewhat tested (at the point it doesn't break any other code).
But needs further test. I am merging it for the testbed deployment but please review and check the code.

vkuznet:
Seangchan, code seems fine to me, I saw that you re-factored quite a bit, so I
can't tell about the logic since it's new to me.

PR #6022, opened 2015-06-24
fix bug in phedex recovery code

hufnagel:
You can't add a dict to a set (there is no way to compare dicts). Change to a list which should work the same. Triggered this in a replay, old code crashed due to trying to add a dict to a set, new code worked.


amaltaro:
Thanks Dirk! Pushing it to the new production agents as well.

hufnagel:
@ericvaandering , why did the merge build fail ?

PR #6023, opened 2015-06-24
Added premixing template; bugfix for wmagent deployment script

amaltaro:
For the record, in order to make deterministic pileup really work, we need at least:
- set 'DataPileup' instead of 'MCPileup'
- set 'DeterministicPileup' to True
- have the correct DataMixingModules in the job configuration (proper ConfigCacheID)

Eric will talk to Liz just to make sure we properly understood it and to have everyone on the same page.
BTW, all my jobs failed with "RootInputFileSequence::readOneSequential(): no input files specified for secondary input source.", that's another thing to investigate now.

Anyways, @ticoann feel free to merge it.


amaltaro:
Updated the PR with LumiList for the DQMHarvest template

PR #6027, opened 2015-06-26
fixes force-complete and aborted

ticoann:
There seems to be a bug on workqueue which doesn't set the workqueue elements to CancelRequested when request gets canceled (aborted and force-complete) which cause jobs to run.
However, I am not completely sure why this hasn't been found so long - it seems the bug was there all the time. (only case this doesn't get affected when cancelGraceTime is set to 0)

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -543,10 +543,18 @@ class WorkQueue(WorkQueueBase):
             # ensure all elements receive cancel request, covers case where initial cancel request missed some elements
             # without this elements may avoid the cancel and not be cleared up till they finish
             elements_not_requested = [x for x in elements if x['ChildQueueUrl'] and (x['Status'] != 'CancelRequested' and not x.inEndState())]
-            if elements_to_cancel or elements_not_requested:
-                self.logger.info("""Canceling work for workflow(s): %s""" % (requestNames))
-                self.logger.info("Canceling element(s) %s" % str([x.id for x in elements]))
-            self.backend.updateElements(*[x.id for x in elements_to_cancel], Status = 'Canceled')
+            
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan, please review. I am still running the test. So I will merge this if test is successful.

amaltaro:
The code was weird, indeed. Patch looks good to me.
About elements_not_requested, the reason we load the document again is to make sure we know the latest rev and so to avoid conflicts, which does not happen in updateElements(), right?

amaltaro:
Seangchan, I tested it in my VM and it seems to be working fine in the GQ, elements get updated to CancelRequested and then they are Canceled.

I'm now looking at the LQ, there seems to be an issue canceling work from the agent side (though I see work 'being' canceled in the WQ logs...). I'm investigating this, but feel free to merge this one and cut a new tag.

ticoann:
> About elements_not_requested, the reason we load the document again is to make sure we know the latest rev and so to avoid conflicts, which does not happen in updateElements(), right?

Actually what happens is when cancel request is fired, it checks whether workqueue element is already in LQ or not. (GQ case)
0. It sets workqueue_inbox element to CancelRequested
1. If WE is NOT in LQ. it sets WE to Canceled in workqueue db.
2. If WE is in LQ, it sets WE in workqueue db to CancelRequested. - The reason doing this is it will replicate that WE in LQ inbox and that should trigger the canceling process from LQ and agent. 
When Agent is done canceling the work it will set the LQ element status to Canceled and replicate to GQ.

The problem was it never sets the GQ element to CancelRequested when there is positive value for cancelGraceTime. I am not sure how this ever worked before.

There are 2 case it could work if the cancelGraceTime is set to 0 or less, Or cancelGraceTime is expired so it canceled the work eventually.

If there is problem on killing the jobs in WMAgent, there must be a bug in how LQ is canceled and triggers job killing. I will look at that today.  However that could be only agent side problem so I will merge this code.

PR #6032, opened 2015-06-28
initial patch for the logdb fix

ticoann:
Handles error filtering in client side (remove the reduced view)
1 first gets the information except messages from logdb for all active requests.
2. fetch the messages for only error and warning cases


ticoann:
@vkuznet, Valentin, please review. I change the css. Sorry, I couldn't find the css from your pull request.
Still need to improve user interface but we can probably do next release. (Screen shot attached)

screen shot 2015-06-28 at 1 46 57 am

ticoann:
That is weird. This shouldn't break any unittest since this only contains javascript change. (other than script creating minified version of javascript). Still need to change to use Valentin's css but merging it for testbed release. (cosmetic change will be done next release).

vkuznet:
Yes, since all reported unit tests failures are not from Logdb, e.g. StoreResults. Seangchan, how they are generated, do we have access to those logs to see what actually failed?

Issue #5971, opened 2015-06-05
Create trust PU site location flag


amaltaro:
Since the 'useSiteWhiteListAsLocation' (or trustSiteList...) applies only to the input dataset, we should also create a flag to trust the PU site location, such that agents would pull down work even if PU sample is not available.

This is mainly to allow running redigi at opportunistic sites. With the SDSC, NERSC and AWS flying around, we should mark it as high prio for the next cmsweb cycle.

hufnagel:
Instead of creating a new specific flag for this, we could also add the needed flags to use the "Tier0 style site override". That one is a direct override in the JobSubmitter, it doesn't rely on faking input data location. As such, it wouldn't care about secondary input location either.

Possible downside is that this needs a check and possible change of any location aware intermediate code layers. For instance, job splitting groups input by location, which can be suboptimal if you then submit the job to other sites anyways.

amaltaro:
What do you mean by "it does not fake input data location". I guess the correct thing here is to get all the files available from the PU dataset and consider (fake) them as available at the site whitelist, no?
Maybe you can point me to a commit/PR containing these changes?

hufnagel:
I mean that the regular site override fakes the data to be present at all sites it runs on. That has consequences for file location aware later code (like job splitting).

The "Tier0 mode" override is a direct override in the job submission, location information is not touched.

hufnagel:
By design your current way of doing things has advantages in that you get the behavior of "run everything as if the data is also at site C, even if it isn't for free. But it has big disadvantages too because this mode by design only allows you to do remote xrootd reads for the very first task in a workflow. You can't override the later stages.

That was exactly the reason I needed something else for the Tier0.

hufnagel:
This is activated with this parameter in a spec

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L902

and then applied per task like this

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L274

amaltaro:
For most of the cases - up to now - faking the data location for the first task only would be all we wanted.
However, from the AWS discussions I've seen so far, we want to run one full workflow starting from scratch and going up to (MINI)AOD., so the Tier0 override is what needs to be done here.

[...]

PR #6035, opened 2015-06-29
Skip PU location if trustSiteList enabled

amaltaro:
Fix for #5971.
I think we should break up this problem into 2 steps:
1. make sure wq elements get replicated to the agent even when sitewhitelist hasn't the input
2. make sure things work Ok on the agent side.

With this PR, we would basically replicate any wq element containing NoLocationUpdate=True.
This way we can run ReDigi workflows anywhere during the next 2 months, at least. 
What do you guys thinkg @ticoann @ericvaandering @hufnagel ?

### DIFF
--- a/src/couchapps/WorkQueue/lists/workRestrictions.js
+++ b/src/couchapps/WorkQueue/lists/workRestrictions.js
@@ -48,20 +48,30 @@ function(head, req) {
             break;
         }
 
+        var ele = row["doc"]["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
+
+        // check work is for a team in the request
+        if (teams.length && ele["TeamName"] && teams.indexOf(ele["TeamName"]) === -1) {
+            continue;
+        }
+
+        // skip if we only want work from certain wf's which don't include this one.
+        if (wfs.length && wfs.indexOf(ele["RequestName"]) == -1) {
+            continue;
+        }
+
+        // Don't check anything if useSiteListAsLocation/trusSiteLists is enabled
+        if (ele['NoLocationUpdate']) {
+            // subtract element jobs from site resources
+            if (first !== true) {
+                send(",");
+            }
+            send(toJSON(row["doc"])); // need whole document, id etc...
+            first = false; // from now on prepend "," to output
+            continue; // we have work, move to the next element
+        }
+
         for (var site in resources) {
-            var ele = row["doc"]["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
-
-            // TODO: probably move this to a standalone function
-
-            // check work is for a team in the request
-            if (teams.length && ele["TeamName"] && teams.indexOf(ele["TeamName"]) === -1) {
-                continue;
-            }
-
-            // skip if we only want work from certain wf's which don't include this one.
-            if (wfs.length && wfs.indexOf(ele["RequestName"]) == -1) {
-                break;
-            }
 
             // skip if in blacklist
             if (ele["SiteBlacklist"].indexOf(site) != -1) {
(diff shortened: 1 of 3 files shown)


amaltaro:
And of course there is a drawback in this implementation, if someone assigns a workflow with 'useSiteListAsLocation':True AND the PU data is not in any of the sites whitelisted, then jobs will probably fail with
```RootInputFileSequence::readOneRandom(): no input files specified for secondary input source.```

That's why we also have to implement the step2, faking PU data location (inside the agent) to all the sites whitelisted.

**Comment by amaltaro** [comment]
Ok, found @davidlange6 GH user name :-)
David, do you know whether we can read PU data (secondary input) via xrootd, fallback mechanism?

IF not, then we simply cannot pull work when PU location does not correspond to site whitelist.

**Comment by davidlange6** [comment]
afaik, yes - but actually using it would be an disaster for our PU via secondary input workflows

Sorry - behind on my inbox today..

**Comment by amaltaro** [comment]
@ericvaandering @ticoann as we discussed in the meeting, this fix the GQ <--> LQ replication.
Then we still need to properly implement how the secondary source is fed inside the worker node.

**Comment by ericvaandering** [comment]
test this please

File: src/couchapps/WorkQueue/lists/workRestrictions.js [hunk 1]
```diff
@@ -52,6 +52,16 @@ function(head, req) {
...
+            // Don't check anything if useSiteListAsLocation/trusSiteLists is enabled
+            if (ele['NoLocationUpdate']) {
+                // subtract element jobs from site resources
+                if (first !== true) {
+                    send(",");
+                }
+                send(toJSON(row["doc"])); // need whole document, id etc...
+                first = false; // from now on prepend "," to output
+                break; // we have work, move to next element (break out of site loop)
+            }
```
ticoann: Since you are fixing it, maybe we should move the not related to site code to out side of the for loop. (line 52 - 74).
Also team and request checking ahead of if (ele['NoLocationUpdate'])

src/python/WMCore/WorkQueue/DataLocationMapper.py
```diff
@@ -246,6 +246,8 @@ def updatePileupLocation(self, fullResync = False):
             for data, locations in dataMapping.items():
                 elements = self.backend.getElementsForPileupData(data)
                 for element in elements:
+                    if element.get('NoLocationUpdate', False):
+                        continue
```
ticoann: When the flag is True, shouldn't location to be set to site whitelist?

amaltaro: I think so, let me refresh my mind on why I just skipped it :)

src/couchapps/WorkQueue/lists/workRestrictions.js
amaltaro: Good suggestions! I'm changing it, thanks!

src/python/WMCore/WorkQueue/DataLocationMapper.py
amaltaro: I think I know why, because the PU data location makes no difference at this level since it's handled differently (job creation does not depend on PU data, and it's created by PileupFetcher.py querying directly dbs).

The only thing we need to make sure is that elements will get replicated (and for this case, don't waste time updating pileup location).

amaltaro:
I'll try to test this new commit tomorrow.
The second patch I want to push is about the pileup.json file, but that will take me a couple of days to go through and we can make a new PR.

ericvaandering:
retest this please

ericvaandering:
retest this please

amaltaro:
I don't know why this test started crashing only know
```
WMSpecFactoryException Message: Validation failed: IncludeParent flag is True but inputdataset /MinimumBias/ComissioningHI-v1/RAW doesn't have parents  ModuleName : WMCore.WMSpec.WMWorkloadTools  MethodName : validateInputDatasSetAndParentFlag
```
I'll try to fix this unittest.

amaltaro:
It's not that simple to fix that unittest, so I'll leave that to another time (besides, it's not related to this patch) :-)
@ticoann I tested it in my VM today and it works fine. Feel free to merge it.

ticoann:
retest this please

PR #6041, opened 2015-07-01
Change to builtin types as preferred since python 2.2

ericvaandering:
Please merge when Jenkins is OK.

### DIFF
--- a/src/python/WMCore/Configuration.py
+++ b/src/python/WMCore/Configuration.py
@@ -23,7 +23,6 @@ _SimpleTypes = [
     str,
     int,
     type(None),
-    int,
     ]
 
 _ComplexTypes = [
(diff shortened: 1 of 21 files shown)


src/python/WMCore/Configuration.py
```diff
@@ -17,19 +17,19 @@
...
-    types.LongType,
-    types.NoneType,
-    types.IntType,
+    bool,
+    float,
+    bytes,
+    str,
+    int,
+    type(None),
+    int,
```
amaltaro: not a big deal, but 'int' twice here

ericvaandering: Yeah, because Long became int. It happened one other place too. Let me fix it.

ericvaandering:
Really this whole commit points out problems that really should be solve with isintance, I think.

ericvaandering:
test this please

ericvaandering:
These unit tests breakages are legit. Do not merge.

PR #6042, opened 2015-07-01
wrap up error messages for humans to read

### LINKED ISSUE #6038: Format LogDB error messages in WMStats (add new lines)

ticoann:
fixes #6038

### DIFF
--- a/src/couchapps/WMStats/_attachments/css/main_layout.css
+++ b/src/couchapps/WMStats/_attachments/css/main_layout.css
@@ -544,3 +544,11 @@ legend {
     font-size: 1em;
     z-index: 100;
 }
+
+pre {
+    white-space: pre-wrap;       /* CSS 3 */
+    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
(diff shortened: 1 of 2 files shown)


ticoann:
@vkuznet, Valentin, please review. I would like to format the messages in generic way in the below code.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/LogDB/LogDB.py#L47

But I am not sure whether that is possible. The css wrapping up will at least wrap up the long strings.

vkuznet:
Changes to CSS looks fine to me, and in JS we can do something like split the input message with white space separator, loop over entries, sum up the word length and insert new line when length will exceed certain length.

ticoann:
Hi Valentin, I think css change should wrap up the messages within div boundary. (So basically JS function you suggested). I was more thinking about content oriented format (on line for title and more importantly error messages breaks up by line in right places). It seems some of the error are already in that format but I am not sure how it can be done).

vkuznet:
yes, agreed. Let's post-pone for a bit and see how real messages will appear on web interface.

PR #6047, opened 2015-07-02
Try to update types leaving out bytes and JSON Thunker

ericvaandering:
A less aggressive version of the patch that seemed to break a lot of JSON thunker

### DIFF
--- a/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
+++ b/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
@@ -262,7 +262,7 @@ class JSONThunker:
         """
         _unthunk - does the actual work for unthunk
         """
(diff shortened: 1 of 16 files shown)


src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
```diff
@@ -262,7 +262,7 @@ def _unthunk(self, jsondata):
         """
         _unthunk - does the actual work for unthunk
         """
-        if (type(jsondata) == types.UnicodeType):
+        if (type(jsondata) == str):
```
PerilousApricot: I'm responsible for this catastrophe of thunk/untunk code (sorry). Assuming we want to move from python 2.X to 3.X, this needs to be changed to isinstance() to handle unicode properly

Issue #6045, opened 2015-07-02
wmagent archivedelayhours is not working


AndrewLevin:
In the wmagent config.py I see

config.TaskArchiver.archiveDelayHours = 336

but workflows are disappearing from wmstats after 1 day. For example:

woodson_HLT_newconditions0_RelVal_246951_150630_140936_4826

AndrewLevin:
Actually, I see that workflow now. But this workflow is missing:

woodson_HLT_reference_RelVal_248038_150630_161538_7152

amaltaro:
When did you change the TaskArchiver configuration? The agent was redeployed on 29 of June with the default value (24h). It may be that this workflow was archived before your change (+ TaskArchiver restart, of course).

AndrewLevin:
I did not change it. 336 was the original value when it was deployed.

amaltaro:
Ok, it was in the deployment script. As I told you in the meeting, I think I know what the problem is and I'm cooking a fix for that.

ticoann:
@amaltaro, Hi Alan, this can be closed, right?

amaltaro:
Hi Seangchan, that's right. Please close this issue.

AndrewLevin:
Wait, Alan, I have a question. If you have two agents that are both on the same team, with different archivedelayhours, then does your fix (https://github.com/dmwm/WMCore/pull/6048) still work?

amaltaro:
it makes no sense to have two agents in the same team with different archiveDelay settings.

AndrewLevin:
But it is still possible to make a mistake. For relval there is only one agent, but for some teams there are like 3 or 4 agents, right? And agents are being added/removed/redeployed pretty frequently, right? So, it is probably not a good idea to require that the settings between different agents be synchronized. I recommend we set this per workflow, not per agent. Or we just set it to globally in one place.

amaltaro:
There is a stable deployment procedure and agents belonging to the same team will always have the same archiveDelay (unless someone changes it by hand).

We don't have finer granularity to control it and IMO making this setting by workflow is completely overkill (besides more work on us). You are the only 'client' interested/using this feature and it will work perfectly fine for you.

AndrewLevin:
[...]

PR #6048, opened 2015-07-03
Archive only workflows belonging to the same team

amaltaro:
Also implemented a method for the "byteamandstatus" ReqMgr view, which is currently broken. I'll make a new PR to fix that.
In addition to that, I made a few cosmetic fixes to let pylint happy :)
@ticoann please review (but only in the next week!)

PS.: this cannot be applied to current production agents since this view is not working there, I'll create a temp dirty patch for these agents....


amaltaro:
Fixes #6045 , though it also depends on #6049

PR #6050, opened 2015-07-06
Temp patch for archiveDelay of workflows - 1.0.8 version

amaltaro:
This is a workaround to let any non-relval agent to archive only non-relval workflows. The correct fix was made against master #6048 but it will also require a new cmsweb, so we stick with this one for now.
Temp fix for #6045


amaltaro:
It seems build failed and looking at "Console Details" under
https://cmsjenkins01.cern.ch/jenkins/job/DMWM-PR-test/105/

I see this message "DMWM-WMCore-PR-pylint #77 completed. Result was FAILURE"
@ericvaandering should I worry about it? Actually, all my PR's tests are failing since last Thursday or so...

PR #6055, opened 2015-07-08
Add a destroy command to wmcore-db-init

mmascher:
Hi,

I added an utility command to wmcore-db-init that allows to destroy a database so it can be recreated from scratch.

Can you have a look at it please?

### DIFF
--- a/bin/wmcore-db-init
+++ b/bin/wmcore-db-init
@@ -24,7 +24,7 @@ from sqlalchemy.engine.url import make_url
 def usage():
 
     msg = """
-Usage: wmcore-db-init [--config] --create <options>
+Usage: wmcore-db-init [--config] (--create|--destroy) <options>
 
 You must either set the WMAGENT_CONFIG environment variable or specify the config file with
 --config 
@@ -36,7 +36,7 @@ need to be created.
     print(msg)
 
 
-valid = ['config=', 'create', 'modules=']
+valid = ['config=', 'create', 'destroy', 'modules=']
 
 try:
     opts, args = getopt.getopt(sys.argv[1:], "", valid)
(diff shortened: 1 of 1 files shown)


bin/wmcore-db-init
```diff
@@ -121,7 +128,18 @@ def create(config):
...
+
+def destroy(config):
+    params = {}
+
+    if hasattr(config.CoreDatabase, "tablespaceName"):
+        params["tablespace_table"] = config.CoreDatabase.tablespaceName
+    if hasattr(config.CoreDatabase, "indexspaceName"):
+        params["tablespace_index"] = config.CoreDatabase.indexspaceName
+
+    wmInit.clearDatabase(modules)
```
ticoann: params and modules are not really used in clearDabase method. You can just do wmInit.clearDatabase() which will delete all the data. (Unless you want to delete specific modules, in that case WMInit clearDatabase method need to be modified.

mmascher: I will remove them

ticoann:
Hi Marco, if you want to add this in command line tool, could you add another safety check.
i.e. ask one more input to user with information, what is going to delete and type explicitly yes.
Such as following

```
Are you sure you want to delete all the wmagent database, it is not reversible. (yes, no)?
```

mmascher:
Hi @ticoann , I have added an `askConfirmation` function to really check and make sure what the user is doing. FYI, this is gonna be used by CRAB3 devs to scratch their development database isntance.

bin/wmcore-db-init
```diff
@@ -121,7 +128,18 @@ def create(config):
...
+    return
+
+def destroy(config):
+    params = {}
+
+    if hasattr(config.CoreDatabase, "tablespaceName"):
+        params["tablespace_table"] = config.CoreDatabase.tablespaceName
+    if hasattr(config.CoreDatabase, "indexspaceName"):
+        params["tablespace_index"] = config.CoreDatabase.indexspaceName
+
```
ticoann: Is setting 'params' needed? It doesn't seems to be used.

ticoann:
Thanks, Marco. It looks good. I just made one comment on the destroy method.

bin/wmcore-db-init
mmascher: Ops, I forgot to remove it! Done

mmascher:
Thank you for your quick feedback!

ticoann:
Thanks for the quick fix. :smile:

PR #6056, opened 2015-07-08
Allow to get non-valid files from DBS in listDatasetFileDetails.

AndresTanasijczuk:
This pull is because in CRAB we only care about files status in DBS and not about dataset status.

When the function listDatasetFileDetails was introduced by @bbockelm it was retrieving the details for all files in the dataset, no matter the dataset/files status. CRAB is then filtering the valid files. Then the "validFileOnly" flag was introduced here and hardcoded to 1. The problem with this is that in this case one gets the valid files only if the dataset status itself is VALID or PRODUCTION, but for any other dataset status one gets no files. And this is not what we want in CRAB. For example, if the dataset is in status INVALID or DEPRECATED, but there are some files in valid status, we want to be able to retrieve them. 

So I am introducing in this pull (first commit) the possibility to set validFileOnly = 0. In CRAB we then take care to filter the valid files. 

I tested the first commit with CRAB on a private dataset (/GenericTTbar/atanasi-140429_131619_crab_AprilTest_3000jbsOf3hrs-95e5dc29a1ac0766eb8514eb5d4ff77a/USER at phys03) that I set to INVALID and left 1 file in valid status (I used the procedure documented in https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCrabForPublication#Invalidate_a_dataset_in_DBS) calling listDatasetFileDetails with validFileOnly = 0. I got all the files in the dataset as I wanted, but I was not getting the lumis. I tried then changing listFileLumis->listFileLumiArray (and therefore getting the lumis per file instaed of per block and btw doing the same for the parents) and I was able to get the lumis. So this is what I changed in the second commit.

### DIFF
--- a/src/python/WMCore/JobSplitting/JobFactory.py
+++ b/src/python/WMCore/JobSplitting/JobFactory.py
@@ -205,8 +205,8 @@ class JobFactory(WMObject):
             fileset = self.loadFiles(size = self.limit)
             logging.debug("Loaded %i files" % (len(fileset)))
         else:
-            fileset = self.subscription.availableFiles(limit = self.limit, doingJobSplitting = True)
             logging.debug("About to load files by DAO")
+            fileset = self.subscription.availableFiles(limit = self.limit, doingJobSplitting = True)
 
         for file in fileset:
             locSet = frozenset(file['locations'])
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -221,7 +221,7 @@ class DBS3Reader:
         return [ x['logical_file_name'] for x in self.dbs.listFileArray(dataset = datasetPath)]
 
 
-    def listDatasetFileDetails(self, datasetPath, getParents=False):
+    def listDatasetFileDetails(self, datasetPath, getParents=False, validFileOnly=1):
         """
         TODO: This is completely wrong need to be redone. or be removed - getting dataset altogether
         might be to costly
@@ -236,36 +236,38 @@ class DBS3Reader:
               'Parents': [],
               'Checksum': '22218315',
               'Adler32': 'a41a1446',
-              'FileSize': 286021145
+              'FileSize': 286021145,
+              'ValidFile': 1
             }
 
         """
-        fileDetails = self.dbs.listFileArray(dataset = datasetPath, validFileOnly = 1, detail=True)
+        fileDetails = self.dbs.listFileArray(dataset = datasetPath, validFileOnly = validFileOnly, detail=True)
         blocks = set() #the set of blocks of the dataset
         #Iterate over the files and prepare the set of blocks and a dict where the keys are the files
         files = {}
         for f in fileDetails:
             blocks.add(f['block_name'])
             files[f['logical_file_name']] = remapDBS3Keys(f, stringify = True)
+            files[f['logical_file_name']]['ValidFile'] = f['is_file_valid']
             files[f['logical_file_name']]['Lumis'] = {}
             files[f['logical_file_name']]['Parents'] = []
 
         #Iterate over the blocks and get parents and lumis
-        #TODO this part is completely wrong need to be redone
         for blockName in blocks:
             #get the parents
             if getParents:
                 parents = self.dbs.listFileParents(block_name=blockName)
                 for p in parents:
-                    if p['logical_file_name'] in files: #invalid files are not there
+                    if p['logical_file_name'] in files: #invalid files are not there if validFileOnly=1
                         files[p['logical_file_name']]['Parents'].extend(p['parent_logical_file_name'])
             #get the lumis
-            file_lumis = self.dbs.listFileLumis(block_name=blockName, validFileOnly = 1)
+            file_lumis = self.dbs.listFileLumis(block_name=blockName)
             for f in file_lumis:
-                if f['run_num'] in files[f['logical_file_name']]['Lumis']:
-                    files[f['logical_file_name']]['Lumis'][f['run_num']].extend(f['lumi_section_num'])
-                else:
-                    files[f['logical_file_name']]['Lumis'][f['run_num']] = f['lumi_section_num']
+                if f['logical_file_name'] in files: #invalid files are not there if validFileOnly=1
+                    if f['run_num'] in files[f['logical_file_name']]['Lumis']:
+                        files[f['logical_file_name']]['Lumis'][f['run_num']].extend(f['lumi_section_num'])
+                    else:
+                        files[f['logical_file_name']]['Lumis'][f['run_num']] = f['lumi_section_num']
 
         return files
 


amaltaro:
Andres, did you check whether listFileLumis() and listFileLumiArray() return the same information in the same format? The former queries by block, so there may be side effects to the other WMCore services (WMAgent).

AndresTanasijczuk:
Alan, I did not explicitly check, because I assumed they should, given that in _getLumiList() https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L63 both methods are used and no distinction on the returned list is made.

belforte:
I think this change is totally off-track. The problem is that DBS server is doing something not documented, not expected and IMHO totally non sensible. See
https://github.com/dmwm/CRABServer/pull/4851#issuecomment-119705437

mmascher:
I talked to @amaltaro , since the `listDatasetFileDetails` function is only used by CRAB3 so there are no concerns about the impact of this pull on WMAgent.

That said his comment is a good one, Andres, ar you sure that `listFileLumis` and `listFileLumisArray` return the same information? Also, with this change we are going from a DBS query per block to a DBS query per file.

I am not sure who wrote the TODO `#TODO this part is completely wrong need to be redone` , but to me querying per block seems the right choiche.. At least that was the implementation Manuel Giffels suggested long time ago.

mmascher:
Ok, I would ignore the TODO and remove it given that I already discussed this with @ticoann : https://github.com/dmwm/WMCore/commit/2afc01ae571390f5fa009dd258be757adac89c28#diff-374b7a6640288184175057234e393e1cR224

And I would continue querying per block.

mmascher:
Back to @belforte 's comment. I agree that the semantic used by DBS is confusing now. If I ask for valid files the DBS API should return the valid files, and having an internal logic that turn valid files to invalid files if the dataset is not in PRODUCTION or VALID is confusing.

That said Andres' pull request address a WMCore limitation in the DBS3Reader wrapper: there is no possibility to get the invalid files as the `validFileOnly` flag is forced to true here `https://github.com/dmwm/WMCore/pull/6056/files#diff-374b7a6640288184175057234e393e1cL243`

In my opinion we can proceed with this pull which is generic enough to be merged independently from what we decide to do in DBS. On the other hand I will wait before merging the "CRAB3 related pull" as it is a "workaround" to the aforementioned DBS behavior.

belforte:
@mmascher OK with addressing WMCore limitation, but indeed  the loop on blocks is replaced by a loop on files. This must be avoided, in case delaying the fix unitl we have found an agreement in DBS, if in the meanwhile CRAB3 can not deal with datasets not VALID nor PRODUCTION... so be it.

ericvaandering:
Rather than go back and forth on this, I think we should schedule a discussion on this with @yuyiguo  when she is back from vacation. That means this can find its way into the September release at the earliest. 

I forget the actual reason for this "feature" only that we discussed it and I asked for the current behavior. From my perspective, if you ask for valid files that match some pattern, you shouldn't by default get files in invalid datasets. If you look at the DBS issue list, you see a lot of related comments about how this is done. It seems like some more flexibility is needed. 

@vkuznet should be in this discussion too to make sure we do the right things for DAS.

mmascher:
This patch makes sense indepentently of what we decide about the DBS logic. It's the related CRAB3 patch that should/should not be merged depending on the decision. But yes, let's wait for Yuyi before taking any decision.

That said, @AndresTanasijczuk can you explain why did you changed and went from the perblock call to the perfile call (not sure I understood it all from our private chat)?

AndresTanasijczuk:
Alan, Marco: I have tested running listFileLumis and listFileLumiArray on 3 datasets (the INVALID one I pointed out above, a dataset of mine with multiple runs per file /SingleMu/atanasi-CRAB3-tutorial_Data-analysis_LumiList-f765a0f4fbf582146a505cfe3fd08f3e/USER, this other dataset /QCD_Pt-1800_Tune4C_13TeV_pythia8/Spring14dr-castor_PU_S14_POSTLS170_V6-v1/GEN-SIM-RECODEBUG) and in all 3 cases the returned dictionary is exactly the same between listFileLumis and listFileLumiArray.
Also, the doc string in both methods for what they return is the same: https://github.com/dmwm/DBS/blob/master/Client/src/python/dbs/apis/dbsClient.py#L829 and https://github.com/dmwm/DBS/blob/master/Client/src/python/dbs/apis/dbsClient.py#L853 It says ":returns: List of dictionaries containing the following keys (lumi_section_num, logical_file_name, run)" which coincides exactly with what I see.
Having said that, I think we actually don't need to use listFileLumiArray. I will make a separate comment on that later. I will also address the question why I changed to query per file instead of per block. Let me only add here that listFileLumiArray allows one to query passing a list of files, so I could in principle make Num_blocks lists of files_per_block and make one query for each of these lists. Of course the question is why would I do that? What is the different between doing that and calling listFileLumis with the block name (as currently done)? Wait for my later comment...

belforte:
maybe better wait until you can get help from Yuyi on what exactly happens server side in the various cases ? The point is not moving a loop on file names on one side or another of a WMCore or DBSclient function but 1 vs 1k calls to Oracle.

AndresTanasijczuk:
Why I changed from perblock to perfile query? 
In my original comment I tried to explain why. I said: "I tested the first commit with CRAB on a private dataset (/GenericTTbar/atanasi-140429_131619_crab_AprilTest_3000jbsOf3hrs-95e5dc29a1ac0766eb8514eb5d4ff77a/USER at phys03) that I set to INVALID and left 1 file in valid status (I used the procedure documented in https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideCrabForPublication#Invalidate_a_dataset_in_DBS) calling listDatasetFileDetails with validFileOnly = 0. I got all the files in the dataset as I wanted, but I was not getting the lumis."
One detail that I forgot to add in that explanation is that on top of the first commit I added also the following change: 
file_lumis = self.dbs.listFileLumis(block_name=blockName, validFileOnly = 1) --> 
file_lumis = self.dbs.listFileLumis(block_name=blockName, validFileOnly = validFileOnly)
I didn't include that change in the first commit, because it would conflict with the second commit.

So the point was that I was not able to get the lumis information when querying perblock using listFileLumis, not matter if I use validFileOnly = 0 or 1 in listFileLumis.

I then suspected (just a guess) that the reason could be that querying per block requires all the files in the block to be in 'valid' state for listFileLumis to return me the lumis information. So I said to myself "let's try querying per file". And I could have kept using listFileLumis querying per file, but listFileLumiArray is more flexible, because it allows one to make a query with a list of files, something I only tested today and works fine.

Now let me give more details of what I have found today.
Just by chance I tried removing the argument validFileOnly when calling listFileLumis and keep doing the query per block. And to my surprise in that case I did get the lumis information. So it seems there is a bug in listFileLumis: not passing validFileOnly gives me the lumis info for the files in the block, while passing it with either value 1 or 0 doesn't give me the lumis info. I would have expected that passing validFileOnly = 0 is the same as not passing it, given the comment in the doc string https://github.com/dmwm/DBS/blob/master/Client/src/python/dbs/apis/dbsClient.py#L827 Again, I am talking about my particular use case example of an INVALID dataset with only 1 file in 'valid' state.
Given this finding, and without caring if this is a bug or not, I could use listFileLumis querying per block and without passing the validFileOnly argument, and add an if statement like in https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L260. This way we will keep querying per block and not need to change to listFileLumiArray. Or I can change to listFileLumiArray and make a query with the lists of files per block. Any of these are IMHO valid alternatives that do not need to wait for any change/bug fix in DBS.

Let me give more details about when the listFileLumis and listFileLumiArray methods did or didn't return me the lumis information in my example dataset:

listFileLumiArray(logical_file_name=LFN, validFileOnly = 1) NO (not even for the valid file)
listFileLumiArray(logical_file_name=LFN, validFileOnly = 0) YES (for all the files)
listFileLumiArray(logical_file_name=LFN) YES (for all the files)

listFileLumiArray(logical_file_name=list(LFNs in block), validFileOnly = 1) NO (not even for the valid file)
listFileLumiArray(logical_file_name=list(LFNs in block), validFileOnly = 0) YES (for all the files)
listFileLumiArray(logical_file_name=list(LFNs in block)) YES (for all the files)

listFileLumis(logical_file_name=LFN, validFileOnly = 1) NO (not even for the valid file)
listFileLumis(logical_file_name=LFN, validFileOnly = 0) YES (for all the files)
listFileLumis(logical_file_name=LFN) YES (for all the files)

listFileLumis(block_name=blockName, validFileOnly = 1) NO (not even for the valid file)
listFileLumis(block_name=blockName, validFileOnly = 0) NO (not even for the valid file)  <---- This is wrong IMO; should give me the info for all files.
listFileLumis(block_name=blockName) YES (for all the files)

I don't want to argue if whether when validFileOnly = 1 I should get the lumis info for the file I have in 'valid' state in my INVALID dataset. That is the discussion Eric, Stefano, Yuyi and others will have about what should DBS do in this case. My point here is just that listFileLumis(block_name=blockName, validFileOnly = 0) should have given me the lumis info as it does listFileLumis(block_name=blockName).

yuyiguo:
Hi, All:

The difference between listFileLumis and listFileLumiArray is that listFileLumiArray can handle a long list of LFNs, but listFileLumis can only handle a short list because the underline methods are different for these two APIs. If you are using list of LFNs, please use listFileLumiArray. Everything listFileLumis can do, listFileLumiArrray should be able to do. I have a bug in listFileLumiArray client API that prevent it to take block name as input. I will fix it in next release.

I open a github issue https://github.com/dmwm/DBS/issues/484 on the problem Andrew reported. All the calls should return lumi list. Looking into the code, I could not see why some of the calls did not return lumi list as Andrew tested. Andrew, could you please update the issue with the block/lfns you used for the test, I will see if I can find the problem using the same data.

In general, it is fast for DBS to get results back if one can use block name instead of a list of LFNs if all the data is in the same block.

The valid file in DBS is defined as 1. The file itself is valid.  2. The dataset that the file belong too has access type as VALID or PRODUCTION. If you want to use “INVALID” file, just get all the files w/o the validFileOnly switch.

Cheers,
Yuyi

AndresTanasijczuk:
Yuyi, I will give the details again of the dataset I used for my test in the DBS issue you opened.

Once you fix listFileLumiArray to accept block name:
1) which of the two methods (listFileLumis or listFileLumiArray) would you recommend us to use when querying with block name?
2) would there be any case where using listFileLumis is recommended instead of listFileLumiArray?

AndresTanasijczuk:
I will change back my second commit to query per block.

AndresTanasijczuk:
Ok, done. Went back to query per block, went back to use listFileLumis instead of listFileLumiArray sine the last doesn't accept block name as input, but removed the validFileOnly argument from listFileLumis so that now it will return the lumis information for all files in the block and I added an if statement that will filter out invalid files in case validFileOnly = 1.

yuyiguo:
Thanks Andres,
listFileLumi can be used same way as  listFileLumiArray as long as you don’t use a long list of LFN when I fixed the block name with listFileLumiArray. The underline code (query) for both APIs are the same so there is no difference for the block name call. For the simpleness of your code, maybe just use on API everywhere. That said if you don’t call list of lfn,  any of the two is fine. If you use list, then just use lisFileArray everywhere.
Cheers,
Yuyi

ericvaandering:
I'm going to merge this as requested in the CRAB3 meeting. It may not be needed down the road.

PR #6067, opened 2015-07-21
Check output datatiers during assignment

### LINKED ISSUE #6062: ReqMgr2 -> Prevent when a request tries to create an invalid data tier on DBS
With current ReqMgr:
- If a request comes with a non-existent datatier in DBS, it will run but the events won't be uploaded to DBS, hence creating errors inside DBS3Upload.
What we need:
- if a request comes with a non-existent datatier in DBS, it will be failed after being assinged (or after being created).
@vlimant

amaltaro:
Fixes #6062 . Also fixed a bunch of unittests for DBS
@ticoann I remember some time ago we didn't want to make DBS calls during assignment, but for this case the api call should be very efficient. Anyways, if you know another way to validate it, let me know :-)

I ran a few tests in my VM and unittests and everything looks fine. Please review.

### DIFF
--- a/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
+++ b/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
@@ -8,8 +8,6 @@ Handles site whitelist/blacklist info as well.
 """
 
 import types
-import copy
-import logging
 import cherrypy
 import threading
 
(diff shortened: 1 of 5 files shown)


ericvaandering:
retest this please

ericvaandering:
On Jul 22, 2015, at 11:51, DMWM Autobot <notifications@github.com<mailto:notifications@github.com>> wrote:


  *   Score for test/python/WMCore_t/Services_t/DBS_t/DBSReader_t.py decreased from 9.84 to 10.00


Well that’s interesting. I’m probably doing a string comparison




Eric W. Vaandering
Computing Physics Developer

Scientific Computing Division
630 840 6678 office

amaltaro:
@ticoann as we discussed a week ago, this should work Ok, but in the near future we better
check these datatiers during job configuration injection (configCache) and cache the output of the dbs api. This way it would cover both reqmgr and reqmgr2

PR #6069, opened 2015-07-22
Let jobs retry when exit code 61102

lucacopa:
The logic of WMJobPermanentSystemErrors is incompatible with the Retry plugin deciding when to fail a job. The list of exit codes that may cause a job to be failed should be at least configurable, not harcoded. In any case, if a site is down, it'll go through retries pretty quickly, so the gain is minimal. With this pull request the behavior for Ops is not changed, but in the Tier0 we can configure the list of exit codes in https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py#L237 at convenience and nothing else would overwrite it. 
@hufnagel @amaltaro @ticoann

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -234,7 +234,7 @@ config.ErrorHandler.logLevel = globalLogLevel
 config.ErrorHandler.maxRetries = maxJobRetries
 config.ErrorHandler.pollInterval = 240
 config.ErrorHandler.readFWJR = True
-config.ErrorHandler.failureExitCodes = [50660, 50661, 50664]
+config.ErrorHandler.failureExitCodes = [50660, 50661, 50664, 61102]
 config.ErrorHandler.maxFailTime = 120000
 config.ErrorHandler.maxProcessSize = 30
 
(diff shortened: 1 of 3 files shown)


hufnagel:
Discussed this with Luis, I second this proposal. I see no reason to hardcode something that is also configurable, especially as for the Tier0 we need it to be configurable.

amaltaro:
Yep! Makes no sense to have configurable exit codes but one exit code hardcoded :)

amaltaro:
retest this please

amaltaro:
hmm, it seems the git clone failed
"error: The requested URL returned error: 403 Forbidden while accessing https://github.com/dmwm/WMCore.git/info/refs"
and opening that link we get a message to upgrade git client. @ericvaandering did I look at the right log?

ericvaandering:
Yeah, I don't know if its a transient problem with GitHub or if its a problem with one of the nodes. We had the deployment fail for the first time in months this morning due to this, but a retry worked.

ericvaandering:
Test this please

ericvaandering:
test this please

PR #6076, opened 2015-07-27
do not bulk submit jobs with different numberofcore to HTCondor

hufnagel:
HTCondor does not support bulk submission of jobs with different cpu requests. Multicore repacking is the first use case triggering this, put in a protection.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -385,68 +385,76 @@ class PyCondorPlugin(BasePlugin):
         # unordered list of jobs with random sandboxes.
         # We intend to sort them by sandbox.
 
+        # We also need to do a secondary sort by number of cores
+        # because we can have jobs with different number of cores
+        # and the same sandbox and they cannot be submitted together.
+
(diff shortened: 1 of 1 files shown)


amaltaro:
That's bad! Perhaps @bbockelm could comment whether it would be possible to get it fixed in HTCondor instead(?) Thus we can avoid over-complicating job submission...

hufnagel:
As far as complicating job submission goes, I think this is peanuts. Instead of sorting jobs by sandbox (which is basically workflow/task), we sort by sandbox and numberOfCores. In practical terms this currently only affects multicore repacking, no other workflow has a variable numberOfCores within the same workflow/task.

amaltaro:
There are taskChain workflows with mixed numberOfCores, though we're still not using PyCondor plugin in production.
Anyways, as you said, it's not a big deal since it's only a different sorting and you made the job already :).

hufnagel:
TaskChain workflows with different numberOfCores should be supported already as the sandbox should be different from task to task. I *think*...

ticoann:
@hufnagel, Dirk is this tested? Can I merge this?

hufnagel:
We've been using this in the Tier0 for while now...

PR #6084, opened 2015-07-30
TaskArchiver will not parse all jobs for aborted wfs

amaltaro:
At some point we started creating workloadSummary for aborted workflows. Since this is a very heavy operation, I'm cutting this process at 5k jobs for aborted-\* workflows.
Also fixed some pylint complaints.
@ticoann review when you are back from vacation please.

### DIFF
--- a/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
+++ b/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
-#pylint: disable=W6501, W0142
-# W6501: pass information to logging using string arguments
+#pylint: disable=W0142
 # W0142: Some people like ** magic
 """
 The actual taskArchiver algorithm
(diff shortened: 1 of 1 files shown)


ticoann:
Hi Alan, if it is aborted workflow, does failed jobs  have to be even in ACDC?

amaltaro:
Nope, I don't think it should create ACDC docs. Is ErrorHandler creating them?

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -690,6 +689,16 @@ def archiveWorkflowSummary(self, spec):
...
             workflowData['output'][dataset]['tasks']  = outputList.get(dataset, {}).keys()
 
+        # If the workflow was aborted, then don't parse all the jobs, cut at 5k
+        try:
+            reqDetails = self.centralCouchDBWriter.getRequestByNames(workflowName)
+            wfStatus = reqDetails[workflowName]['RequestTransition'][-1]['Status']
+            if wfStatus in ["aborted", "aborted-completed", "aborted-archived"]:
+                logging.info("Workflow %s in status %s with a total of %d jobs, capping at 5000" % (
+                    workflowName, wfStatus, len(failedJobs)))
+                failedJobs = failedJobs[:5000]
```
ticoann: Should we need to keep 5000 data? I am just wondering what is the right number. How many failed job was there when this was causing problem?

ticoann:
I thought it was TaskArchiever but you are right, it is ErrorHandler. (So my comment is unrelated. sorry about that. :smile: )

amaltaro:
I mixed them up all the time as well :)

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
amaltaro: There were 2M jobs in that case, which would take 5 days, at least. I don't remember now how long does it take to process 250 (standard slice size), but I would guess a minute.
I just wanted to be conservative with this 5k jobs, but if you think we can go down, I can set it to 1k jobs only (?) :)

ticoann: Thanks Alan, let's try this and see what happens

PR #6085, opened 2015-07-30
Swith to compatible xrd client for old dcache version at FNAL

lucacopa:


ericvaandering:
How urgent is this?

lucacopa:
there will be an update in all the FNAL workernodes, we need this patch to be in place for new workflows before that happens. Maybe Alan need to fix the workflow sandboxes for some workflows already created in the wmagents too.

ericvaandering:
Fixing sandboxes sounds like a crazy idea to me. I'm pretty sure @amaltaro  has better things to do with his time.

If FNAL can't figure out a way around this, then we can drain the site and start over.

amaltaro:
I guess that's the best option for us. We've updated the sandboxes in the past (very dangerous but possible :))

PR #6086, opened 2015-07-30
Mark flipping tests as integration

ericvaandering:
At least this should take out some of the tests that insist on flipping back and forth and making noise.


ericvaandering:
I'm going to merge this. There are more flipping tests to be taken care of, though.

PR #6092, opened 2015-08-02
stop error messages on dashboard reporting of logarchives

hufnagel:
Get rid of spurious error messages in the dashboard reporting for logArchive, LogCollect and Sqlite output.

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -767,26 +767,18 @@ class Report:
         """
 
         stepReport = self.retrieveStep(step = step)
-
         if not stepReport:
             logging.debug("Asked to retrieve files from non-existant step %s" % step)
             return []
 
+        # steps with no outputModules is normal
         listOfModules = getattr(stepReport, 'outputModules', None)
-
         if not listOfModules:
-            logging.debug("Asked to retrieve files from step %s with no outputModules" % step)
-            logging.debug("StepReport: %s" % stepReport)
             return []
 
         listOfFiles = []
-
         for module in listOfModules:
-            tmpList = self.getFilesFromOutputModule(step = step, outputModule = module)
-            if not tmpList:
-                continue
-            listOfFiles.extend(tmpList)
-
+            listOfFiles.extend( self.getFilesFromOutputModule(step = step, outputModule = module) )
 
         return listOfFiles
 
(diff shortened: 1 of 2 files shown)


hufnagel:
please test again

amaltaro:
It will be great to get rid of these messages in the component :)
Dirk, it seems we don't need to initialize dataTier and procDataset, we can also remove the ifs in L255 and L257.

hufnagel:
Restructured a bit more, should be easier to read now. I can see why the if not dataTier and if not procDataset statement where there, but it makes the code harder to read and the if not doesn't really save anything compared to getting the value again from the file info. Plus we usually do not have multiple files per output module, so I changed that too to make it more readable. Take a look.

amaltaro:
Sounds good to me.  Would you mind running pylint in these 2 files? Some of its suggestions would be better to follow and move towards standardization of the code :-) E.g.:
```
pylint --rcfile=standards/.pylintrc src/python/WMCore/Services/Dashboard/DashboardReporter.py
```

hufnagel:
I thought the DMWMBot is running pylint.

hufnagel:
In general, I'll follow pylint if it complains about specific changed in my patch. I am NOT going to rework the rest of the code, that should be done in a separate issue.

amaltaro:
It does. But you can always catch some error/warnings and good practices before committing your code e.g.(spaces here and there, trailing whitespaces, etc). 

You don't need to rework the whole code, just avoid injecting more warnings into the code. Variable names and line length are two cases that we cannot avoid, in general.

ericvaandering:
It does run pylint, but it only complains if the code is really bad or if your patch makes it worse, I think. Of course it can be made to do something different if you feel it’s helpful. It also ignores anything that is a “C” warning.

Eric

On Aug 3, 2015, at 07:21, Dirk Hufnagel <notifications@github.com<mailto:notifications@github.com>> wrote:


I thought the DMWMBot is running pylint.

—
Reply to this email directly or view it on GitHub<https://github.com/dmwm/WMCore/pull/6092#issuecomment-127216391>.




Eric W. Vaandering
Computing Physics Developer

Scientific Computing Division
630 840 6678 office

hufnagel:
Had to change the patch slightly as it was conflicting with some other changes (and I didn't want to integrate everything into a single pull request).

hufnagel:
Ok, I might have to rework this a bit further. Please don't merge yet.

hufnagel:
retest this please

hufnagel:
So how can I see the details of why the checks failed ? The 'Details' link doesn't work.

ericvaandering:
Yeah, the “details” points to the wrong URL, but I can’t change that and if Shazad changes it, it breaks Offline’s tests. So we leave it for now. That’s why I put that first message there, which should give you the right “Details”.

However, when Jenkins actually fails, it’s because the tests actually fail to run, not because of some criteria on the tests. For instance, if you follow this through you can end up here: https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=0,label=slc6_amd64-cloud/103/console (the console log for one slice for the test that failed). In this case, CherryPy refused to shut down…. But the last two tests have had a lot of failures and one looked like a deployment failure, so maybe something is up with a node.

I’ve tried a couple of things to try to get failing slices to re-run, but no luck so far…

Eric



On Aug 3, 2015, at 4:58 PM, Dirk Hufnagel <notifications@github.com<mailto:notifications@github.com>> wrote:


So how can I see the details of why the checks failed ? The 'Details' link doesn't work.

hufnagel:
retest this please

hufnagel:
Will rework and extend this, closing for now.

PR #6099, opened 2015-08-05
Added unique constraint to oracle dbsbuffer_file lfn

amaltaro:
As discussed in this elog
https://cms-logbook.cern.ch/elog/Workflow+processing/21225
this is the simple patch.

However, I noticed quite a bit of inconsistencies between the mysql and oracle db schema. They basically are:
- oracle has much more not null constraints
- oracle missing unique constraint for dbsbuffer_file_parent
- oracle with additional fk for dbsbuffer_block
- oracle with additional fk for dbsbuffer_file_runlumi_map
- mysql missing a pk for dbsbuffer_algo_dataset_assoc
  you can see the whole story here:
  https://github.com/amaltaro/WMCore/commit/6b1379a5a37ff52fe3c6b69ec071bb96d6956e2a

@ticoann what do you think? should we make them consistent? replicate which schema (mysql -> oracle or vice-versa?)? 
Otherwise, someone could make some diagrams and investigate what's really correct :-)

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/Oracle/Create.py
+++ b/src/python/WMComponent/DBS3Buffer/Oracle/Create.py
@@ -317,10 +317,6 @@ class Create(DBCreator):
                (CONSTRAINT dbsbuffer_algo_unique UNIQUE (app_name, app_ver,
                                                          app_fam, pset_hash) %s)""" % tablespaceIndex
 
-        self.indexes["01_pk_dbsbuffer_file"] = \
-          """ALTER TABLE dbsbuffer_file ADD
-               (CONSTRAINT dbsbuffer_file_pk PRIMARY KEY (id) %s)""" % tablespaceIndex
-
         self.indexes["01_pk_dbsbuffer_file_parent"] = \
           """ALTER TABLE dbsbuffer_file_parent ADD
                (CONSTRAINT dbsbuffer_file_parent_pk PRIMARY KEY (child, parent) %s)""" % tablespaceIndex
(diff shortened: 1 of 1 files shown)


amaltaro:
BTW, this is the updated version of my local patch (with lots of changes):
https://github.com/amaltaro/WMCore/commit/13e461dabf08aeb1a91dcc5663f275c10caffbd1

Maybe @yuyiguo could help us in this task? :-)

amaltaro:
Sorry for one more update guys, I left one comma (,) behing and it was breaking MySQL agent deployment: here is the up-to-date commit: https://github.com/amaltaro/WMCore/commit/6019cbaab21b827695a0cbd2b4b4ce56d5473ee2
@yuyiguo pinging you again to make sure you won't miss it :-)

yuyiguo:
Alan:
Here is my comments/questions:
1. what does it mean removing NOT NULL in the map TABLE dbsbuffer_algo_dataset_assoc of Oracle?

2. Are these changes happened to newly created DB only or to existing DB too. If to existing DB when you changed dbsbuffer_file_checksums_uk to dbsbuffer_file_checksums_uq, did you drop the old *uk* index? if not, there would end up two unique constraints there. Same to fk_dbsfilechecksums_cktype, fk_dbsfilechecksums_file, dbsbuffer_dataset_unique and ....  

3. self.indexes["01_uq_dbsbuffer_file_parent"] = \
           """ALTER TABLE dbsbuffer_file_parent ADD
-               (CONSTRAINT dbsbuffer_file_parent_pk PRIMARY KEY (child, parent) %s)""" % tablespaceIndex
+               (CONSTRAINT dbsbuffer_file_parent_uq UNIQUE (child, parent) %s)""" % tablespaceIndex

The  oracle primary key is guaranteed to be unique, Why change it? Then you need to create another key as primary key in principle.  
Yuyi

amaltaro:
Thanks for looking into it Yuyi. I guess I forgot to give you some background on these changes I made.
I basically took the schema we currently have for the mysql db and applied them to oracle (besides the implicit enabled not null for oracle), the idea is to have them consistent.

Now answering your questions:
 1. there is no not null constraint for those columns in mysql (and in principle they should always be an integer). I believe removing constraints that we don't really need will make db operations more performant, no?
 1.a. we will not change agents on the fly, it will only apply to newly deployed agents.
 
> The oracle primary key is guaranteed to be unique, Why change it? Then you need to create another key as primary key in principle.

I didn't know that :) Anyways, child, parent are not supposed to be PK, in principle.

amaltaro:
@ticoann how about we merge this PR for now and we open another issue to review - at least - the dbsbuffer database with @yuyiguo ? 
Yuyi, would you have time this week to check that out? Or only after testbed deadline? Do you want a meeting with Seangchan and me (if that's needed)?

yuyiguo:
All the changes you made were OK. Anything could be primary key as long as they were supposed to be unique. Also I think it is a good idea to keep map tables have not null that is the guarantee of the data integrity no mater what happens in the code. However if you want to put this kind of data integrity check inside the code, it is OK to remove "Not Null " in the DB schema.

If I would you, I would change the mysql schema to match Oracle for the first and last comments I had. But what you did was OK. I think you could go head to merger the code.

hufnagel:
I am very much against removing "not null" constraints if the row indeed is never supposed to be null. Apart from adding another protection layer against screwups in code, "not null" constraints can speed up database performance by giving the database hints about content of a table.

Yes, inserts are a bit slower, but then, they are blindingly fast anyways. Complicates queries with many joins can potentially benefit from "not null" constraints.

Also, there is so much effort spend on databases in terms of manpower and code optimizations, anytime someone says: "oh, we do do this in code because it's faster", I really really don't believe them.

amaltaro:
Thanks Yuyi, Dirk! Just to be clear, I basically removed only the "enabled" from the "not null" constraints because it's enabled by default. But looking at the code, indeed a few not null were removed.

I'll update that code and create a PR, so we can follow up and adjust a few details from there. Thanks!

PR #6100, opened 2015-08-05
Remove imports that no longer exist

ericvaandering:
This won't actually fix any tests. Still trying to work out how to do that with CherryPy in our setup. But these imports disappeared in CherryPy 3.2 and the method of testing changed. So it seems to me you can merge this and then let things fail on something other than the missing import.


ericvaandering:
retest this please

ericvaandering:
retest this please

ericvaandering:
Retest this please

PR #6102, opened 2015-08-06
Disable onlySSB; increased pending thresholds

amaltaro:
We still need to change the default site state when resource-control gets popupated from normal to down...
The pending thresholds will slightly increase for central production nodes sharing the same team, but for single agent team, it will be a very big step :)

@julianbadillo FYI


amaltaro:
tested and good to go

PR #6103, opened 2015-08-06
add debug output and consistency checks for file registration

hufnagel:
Also restructure/cleanup the job accounting code for job handling.

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -196,26 +196,6 @@ class AccountantWorker(WMConnectionBase):
 
         return jobReport
 
-    def didJobSucceed(self, jobReport):
-        """
-        _didJobSucceed_
-
-        Get the status of the jobReport.  This will loop through all the steps
-        and make sure the status is 'Success'.  If a step does not return
-        'Success', the job will fail.
-        """
-        if not hasattr(jobReport, 'data'):
-            return False
-
-        if not hasattr(jobReport.data, 'steps'):
-            return False
-
-        if not jobReport.taskSuccessful():
-            return False
-
-
-        return True
-
     def isTaskExistInFWJR(self, jobReport, jobStatus):
         """
         If taskName is not available in the FWJR, then tries to
@@ -237,7 +217,9 @@ class AccountantWorker(WMConnectionBase):
             else:
                 logging.info("TaskName '%s' successfully recovered and added to fwjr id %s." % (jobReport.getTaskName(),
                                                                                                 jobReport.getJobID()))
-        
+
+        return
+
     def __call__(self, parameters):
         """
         __call__
@@ -254,25 +236,16 @@ class AccountantWorker(WMConnectionBase):
             # Load the job and set the ID
             fwkJobReport = self.loadJobReport(job)
             fwkJobReport.setJobID(job['id'])
-            jobSuccess = None
             
-            if not self.didJobSucceed(fwkJobReport):
-                logging.error("I have a bad jobReport for %i" %(job['id']))
-                self.handleFailed(jobID = job["id"],
-                                  fwkJobReport = fwkJobReport)
-                jobSuccess = False
-            else:
-                self.isTaskExistInFWJR(fwkJobReport, "success")
-                self.handleSuccessful(jobID = job["id"],
-                                      fwkJobReport = fwkJobReport,
-                                      fwkJobReportPath = job['fwjr_path'])
-                jobSuccess = True
+            jobSuccess = self.handleJob(jobID = job["id"],
+                                        fwkJobReport = fwkJobReport)
 
             if self.returnJobReport:
                 returnList.append({'id': job["id"], 'jobSuccess': jobSuccess,
                                    'jobReport': fwkJobReport})
             else:
                 returnList.append({'id': job["id"], 'jobSuccess': jobSuccess})
+
             self.count += 1
 
         self.beginTransaction()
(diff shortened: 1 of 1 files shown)


hufnagel:
@amaltaro , @ticoann , this will need some review/testing on the agent side. Also, as I said, the list of exceptions for the comparison of expected job output and actual job output is not complete. What I have works for the Tier0, but not for other multi-step agent workflows for instance.

I am going to push to integrate this into the next Tier0 prod deployment if possible...

hufnagel:
Added different code path for multi-step CMSSW Processing and Production jobs, they are handled correctly now (essentially we skip the consistency check for them).

ericvaandering:
retest this please

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -493,49 +463,126 @@ def handleSuccessful(self, jobID, fwkJobReport, fwkJobReportPath = None):
...
+            elif jobType == "Express" and set(outputMap.keys()).difference(outputModules) == set(['write_RAW']):
+                pass
+            else:
+                failJob = True
+                if jobType in [ "Processing", "Production" ]:
+                    cmsRunSteps = 0
+                    for step in fwkJobReport.listSteps():
+                        if step.startswith("cmsRun"):
+                            cmsRunSteps += 1
+                    if cmsRunSteps > 1:
```
amaltaro: Dirk, why do you fail Processing/Production jobs with > 1 cmsRuns?
I understood the idea was not to fail them, but to skip the expected outputmodules check.

hufnagel: I do not fail them, I skip failing them. Check the logic: default is to fail the job. If it is a Production or Processing job with more than one cmsRun step, this isn't done.

amaltaro: Wrong comment, sorry. I'll delete it.

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -493,49 +463,126 @@ def handleSuccessful(self, jobID, fwkJobReport, fwkJobReportPath = None):
...
+            fileList = fwkJobReport.getAllFiles()
+
+            # consistency check comparing outputMap to fileList
+            # they should match except for some limited special cases
+            outputModules = set([])
+            for fwjrFile in fileList:
+                outputModules.add(fwjrFile['outputModule'])
+            if set(outputMap.keys()) == outputModules:
+                pass
+            elif jobType == "LogCollect" and len(outputMap.keys()) == 0 and outputModules == set(['LogCollect']):
```
amaltaro: Why don't you simply pass the LogCollect check (by jobType only)? Can it ever be wrong (since there is no output besides the tarball to castor)?

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -493,49 +463,126 @@ def handleSuccessful(self, jobID, fwkJobReport, fwkJobReportPath = None):
...
+                    fileList = fwkJobReport.getAllFilesFromStep(step = 'logArch1')
+                else:
+                    logging.debug("Job %d , list of expected outputModules does not match job report, accepted for multi-step CMSSW job", jobID)
+        else:
+            fileList = fwkJobReport.getAllFilesFromStep(step = 'logArch1')
+
+        if jobSuccess:
+            logging.info("Job %d , handle successful job", jobID)
+        else:
+            logging.error("Job %d , bad jobReport, failing job",  jobID)
```
amaltaro: Is this really an error (bad job report)? It could be that the job didn't have the same output modules and in this case it's a failed job but not bad job report, right?
In other words, are we going to print an ERROR in logs for any/every job that has exit code != 0? If the answer is yes, the we should change this logging.

src/python/WMComponent/JobAccountant/AccountantWorker.py
hufnagel: Its a consistency check, I just put in the maximal constraint I can think off. In practical terms just checking for LogCollect job type would be enough too, but as I have the other information available anyways...

src/python/WMComponent/JobAccountant/AccountantWorker.py
hufnagel: Well, if the output in the job report doesn't match the expected output, the job report is bad in some way. The previous line will have the error printout about the non-matching output, so I figured there was no need to print out different failure reports in addition to it. If you look at where jobSuccess is initially set, it already groups multiple failure modes into one variable and we don't print out different messages for these too.

hufnagel: Btw, we already had a printout for every failed job

https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobAccountant/AccountantWorker.py#L260

This is just the equivalent as the other line was removed.

src/python/WMComponent/JobAccountant/AccountantWorker.py
amaltaro: Ok, it does not hurt having all this info there. I usually prefer not overcomplicating what is not needed :-)

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -470,20 +443,17 @@ def _mapLocation(self, fwkJobReport):
...
-        wmbsJob.load()
-        wmbsJob["outcome"] = "success"
-        wmbsJob.getMask()
-        outputID = wmbsJob.loadOutputID()
+        Figure out if a job was successful or not, handle it appropriately
+        (parse FWJR, update WMBS) and return the success status as a boolean
 
-        wmbsJob["fwjr"] = fwkJobReport
+        """
+        jobSuccess = hasattr(fwkJobReport, 'data') \
```
amaltaro: BTW, since you are changing this code... how about adding the 2 hasattr() inside this method 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/FwkJobReport/Report.py#L1010
?

hufnagel: Actually, report always has a data attribute and that always has a steps attribute. They are created in the init method. So at most I can remove the checks on them here if you want, but I am not going to move the 2 hasattr checks into the Report class (as then I would have to add a lot more internal checks in the report class).

amaltaro: Answering here as well. If they are initialized in the init method, then I would simply remove those checks.

hufnagel:
Let me also get rid of the pre-2.6 python idioms jenkins is complaining about...

amaltaro:
Dirk, are you still cooking something for this PR or is it ready to go? I want to test it with the newest tag (and the logCollect fix, please confirm whether that is ready as well). Thanks

hufnagel:
Its ready to go.

PR #6111, opened 2015-08-10
Fake SE for PU when trustSiteLists

### LINKED ISSUE #5971: Create trust PU site location flag

amaltaro:
Fixes #5971, complementing #6035. I ran a couple of tests today and it seems to be working fine, I just did a minor change and want to check this out before getting it merged.

@ticoann @ericvaandering I know we discussed something a bit different some time ago. AFAIR, it was something about a new PUSiteWhitelist that the assigner would set when setting trustSiteLists to True.

Instead of creating a new sitewhitelist, I simply used the usual SiteWhitelist and forced those SE's into the pileupconf.json file (so the job can try a file opening fallback).


amaltaro:
Tested and good to go! Please include this one in today's tag

PR #6117, opened 2015-08-11
Fix local couch cleaning calls

amaltaro:
We were calling the wrong db (actually, couchapp/design doc) and querying a view that is not available in local wmstats.
I tested this fix in testbed and several workflows were properly archived.

@ticoann please review asap, so I can push it to all the production agents (and also create a master version, needs to be slightly different).


ericvaandering:
retest this please

ericvaandering:
Ah, this is failing because it's a pull against 1.0.8 instead of master. Ignore it I guess. I should try to figure this out at some point, though.

ticoann:
Thanks Alan, I think I made the mistake when I cleaned up the views in wmagent. Proper fix should be reverting the view back. I will create the patch.

ticoann:
Hi Alan, #6119 is created

amaltaro:
Cool, thanks Seangchan! Feel free to close this one at your will. Otherwise I can close it once the other gets merged and in place

PR #6120, opened 2015-08-11
Disable four more flapping tests

ericvaandering:


ericvaandering:
@ticoann although there is probably more to come, please merge this.

PR #6121, opened 2015-08-11
fix complete transiton update

ticoann:
move the Oracle/MySQL db complete transition after all other data (couchdb) is updated.
There could be a race condition if the couchdb update fails (wmbs data could get deleted before couchdb update)


ticoann:
@amaltaro, Alan, could you review.

amaltaro:
Seems the right thing to do. Thanks Seangchan!

ticoann:
retest this please

PR #6125, opened 2015-08-11
making logcollect jobs run better

hufnagel:
LogCollect jobs have a gazillion problems. This is fixing some of them.
- try to stage the logArchive in once, with a timeout at 10 minutes, then give up
- do not actually delete them unless they were staged in and added to the tarball
- stage in one file, add it to the tarball and then delete it from local disk and mass storage
- deletion from mass storage is optional, we try it for 5 minutes and then give up (and not fail the job)

This version seems to work, but probabaly needs some work, open questions:
- should I delete all logArchive if any of them can be processed and therefore the LogCollect job succeeds ?
- what about the fixed timeouts for stagein, is it ok to not retry ?
- what about the fixed timeout for deletions, should it be longer (we should never retry deletions, its not worth waiting for it considering this is just unmerged cleanup at that point) ?

This is fixing two immediate problems we have observed in the Tier0:
- LogCollect jobs ran out of local disk because it staged in all files and then build the tarball instead of doing it file by file
- LogCollect job took hours to retry stagein of a non-existent file, for each input file, leading to 10+hour LogCollect jobs that ultimately failed anyways

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -14,16 +14,16 @@ import socket
 
 from WMCore.WMSpec.Steps.WMExecutionFailure import WMExecutionFailure
 
-from WMCore.WMSpec.Steps.Executor           import Executor
-from WMCore.FwkJobReport.Report             import Report
+from WMCore.WMSpec.Steps.Executor import Executor
 
-import WMCore.Storage.FileManager
-import WMCore.Storage.StageOutMgr as StageOutMgr
-import WMCore.Storage.StageInMgr  as StageInMgr
-import WMCore.Storage.DeleteMgr   as DeleteMgr
+from WMCore.Storage.StageOutMgr import StageOutMgr
+from WMCore.Storage.StageInMgr import StageInMgr
+from WMCore.Storage.DeleteMgr import DeleteMgr
+
 from WMCore.Storage.StageOutError import StageOutFailure
 
 from WMCore.Algorithms.Alarm import Alarm, alarmHandler
+
 
 class LogCollect(Executor):
     """
@@ -40,7 +40,7 @@ class LogCollect(Executor):
         Pre execution checks
 
         """
-        #Are we using an emulator?
+        # Are we using an emulator?
         if (emulator != None):
             return emulator.emulatePre( self.step )
 
@@ -52,7 +52,7 @@ class LogCollect(Executor):
         _execute_
 
         """
-        #Are we using emulators again?
+        # Are we using emulators again?
         if (emulator != None):
             return emulator.emulate( self.step, self.job )
 
@@ -66,12 +66,11 @@ class LogCollect(Executor):
         lfnPrefix = overrides.get('lfnPrefix', "srm://srm-cms.cern.ch:8443/srm/managerv2?SFN=/castor/cern.ch/cms")
         lfnBase   = overrides.get('lfnBase',   "/store/user/jsmith")
         userLogs  = overrides.get('userLogs',  False)
-        dontStage = overrides.get('dontStage', False)
 
         stageOutParams = {"command": "srmv2-lcg",
                           "se-name": seName,  "lfn-prefix": lfnPrefix}
         
-        #Okay, we need a stageOut Manager
+        # Okay, we need a stageOut Manager
         useNewStageOutCode = False
         if getattr(self.step, 'newStageout', False) or \
             ('newStageOut' in overrides and overrides.get('newStageOut')):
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann @amaltaro , comments ? Mainly on policy, once I know what we want I can implement it.

hufnagel:
Btw, question, what is the retry policy for LogCollect jobs ? I remember that we switched them to not be retried at all at some point, which means I should take that into account inside the LogCollect job and maybe at least for the stageIn allow one (quick) retry.

amaltaro:
Dirk, logCollects are retried at least once (maybe even twice for some cases - RelVals probably)

> - deletion from mass storage is optional, we try it for 5 minutes and then give up (and not fail the job)

we no longer store anything in mass storage, it's all in disk. Unless I misunderstood something here.

> - LogCollect jobs ran out of local disk because it staged in all files and then build the tarball instead of doing it file by file

I thought you were trying to delete files from the storage :) Now it's clear you meant files staged in to the worker nodes only.
BTW, are logCollect jobs deleting anything from the storage at all? I guess only Cleanup jobs delete any data from the site, right?

hufnagel:
No, LogCollect delete the input files (logArchives) from the unmerged space when they are done. No cleanup jobs for LogCollect, it all happens inside the LogCollect job.

At the moment I am only deleting the files that actually make it into the LogCollect job, not the ones that are skipped. And skipping is a valid option, a LogCollect job is considered successful if at least one of the input files an be staged in and used. I didn't change that logic.

ticoann:
retest this please

ericvaandering:
test this please

amaltaro:
As Dirk and I discussed in the office, the policy Dirk is taking about is more about the step itself (reading in files, deleting, etc). With this patch
 - files that failed to be read will just be left behind (retries 0)
 - files that failed the deletion will also stay behind (retries 0)
 - 10min of timeout to read in files (maybe you could define an attr/var for that, so we have a single and clear number in the code)
 - 5min for file deletion (same comment as above, use a var please)

Patch looks good to me. Is it tested in T0?

hufnagel:
Minimally tested in a Tier0 replay. Need more testing though and I also need confirmation on the policy, especially how this should go together with the LogCollect job retries (as some of what the job does internally can make the job retry redundant and vice-versa).

amaltaro:
I'd like to add this PR in the new release candidate WMAgent (>= 1.0.9.pre6). Let us get back here in the next week :-)

hufnagel:
Changed the code a bit. Biggest functional change is that in case of a succeeding job, ALL input files are deleted (not just the ones that could be staged in and used). Also made some minor mods to make pylint a bit more happy.

hufnagel:
Anyone have an idea what this "* is not callable error" error is about ? The code works as is.

ticoann:
HI Dirk, 
StageOutMgr is a module not class in the following statement (so it is not callable)

``` python
import WMCore.Storage.StageOutMgr as StageOutMgr
```

You can change that

``` python
from WMCore.Storage.StageOutMgr import StageOutMgr
``` 
Same for 

StageInMgr
DeleteMgr

Seangchan

ericvaandering:
Except that we have this __call__ method in these things that does make them callable. We seem to do this a lot and I don't really understand the rationale behind it, but it confuses pylint since you appear to be treating a class as a function. I don't think pylint goes deep enough into the code to know that we are doing this.

hufnagel:
Plus what Seangchan suggested doesn't work, that code crashes. The code is correct as-is, its just pyling that can't figure that out.

ticoann:
Hi Dirk, sorry I didn't check the whole file when I looked at patch. (When you changed import statement, I think it crashes because following line (as well as line 81, 82)

https://github.com/hufnagel/WMCore/blob/logcollect-optimize/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py#L83

also has to be 

StageOutMgr(**stageOutParams) not StageOutMgr.StageOutMgr(**stageOutParams)

Or if you leave the import statement as it is, following line has to be 

https://github.com/hufnagel/WMCore/blob/logcollect-optimize/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py#L87

StageOutMgr(retryPauseTime  = self.step.retryDelay,
                                          numberOfRetries = self.step.retryCount,
                                          **overrides)
->

StageOutMgr.StageOutMgr(retryPauseTime  = self.step.retryDelay,
                                          numberOfRetries = self.step.retryCount,
                                          **overrides)

It seems there are 2 places using StatgeMgr class but one use the module and the other use the class.
I think current code works because you are not using new plug in but pylint is complaining the code in the new plug in block.

Seangchan

hufnagel:
Ok, let me give that a try...

hufnagel:
This is also ready to go.

PR #6126, opened 2015-08-12
Sync Mysql schema w.r.t. Oracle

amaltaro:
As discussed in #6099, here is the PR. Changes are:
- as suggested by @yuyiguo , not null was put back for a few cases
- set back primary key for dbsbuffer_file_parent (though I don't get why we would need both, I guess only child would be enough)
- synced other constraints from oracle to mysql

I also think we should remove the uniqueness from those 4 columns in dbsbuffer_algo, they make no sense for me (though I kept them here).

@yuyiguo @hufnagel you may want to take a look here

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
@@ -24,15 +24,14 @@ class Create(DBCreator):
         self.create["01dbsbuffer_dataset"] = \
               """CREATE TABLE dbsbuffer_dataset
                         (
-                           id              BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
+                           id              BIGINT UNSIGNED PRIMARY KEY AUTO_INCREMENT,
                            path            VARCHAR(500) COLLATE latin1_general_cs UNIQUE NOT NULL,
                            processing_ver  VARCHAR(255),
                            acquisition_era VARCHAR(255),
                            valid_status    VARCHAR(20),
                            global_tag      VARCHAR(255),
                            parent          VARCHAR(500),
-                           prep_id         VARCHAR(255),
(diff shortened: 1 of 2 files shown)


yuyiguo:
Alan:
The patch looked OK to me. I don't have the completed two create files to compare, but I am sure you already did  the sync.

self.indexes["01_pk_dbsbuffer_file_parent"] = \
          """ALTER TABLE dbsbuffer_file_parent ADD
               (CONSTRAINT dbsbuffer_file_parent_pk PRIMARY KEY (child, parent) %s)""" % tablespaceIndex

The above PK is needed, only child can not be used as primary key as a primary key requires uniqueness. This is a map table so only one pair of child-parent should be in the table. That is good to be used as PK and one does not have to define an extra PK. 

 Did you mean to remove below?
self.indexes["02_pk_dbsbuffer_algo"] = \
          """ALTER TABLE dbsbuffer_algo ADD
               (CONSTRAINT dbsbuffer_algo_unique UNIQUE (app_name, app_ver,
                                                         app_fam, pset_hash) %s)""" % tablespaceIndex

An Algo is defined as a group of app_name, app_ver, app_fam and pset_hash. The uniqueness of the group is make sure there is only one group with the same 4 variables will be in the DB. if it is removed, then the same group may be added in the table multiple times. The unique key will make the table smaller and quick searching.  

Yuyi

amaltaro:
Thanks for reviewing and explaining these things to me Yuyi :-)
Dirk also taught me a few things this afternoon.
I had a few constraint name issues this afternoon, but after fixing a few of them, I managed to deploy an agent (oracle) and so far so good.
I'll test myssql tomorrow and let you guys know about the outcome.

amaltaro:
Everything is fine for mysql flavor as well. From my side, we're good to go.

PR #6130, opened 2015-08-13
Old exception syntax snuck in

ericvaandering:
This is the only python 2.7 regression. :-)


amaltaro:
Only 1, not bad at all!! :)

ericvaandering:
+1

PR #6133, opened 2015-08-14
Update to use isinstance vs. type

ericvaandering:
Don't merge this, but I want to see what happens. This is an optional part of futurize


ericvaandering:
retest this please

ericvaandering:
No unit test failures, but I we'll skip this for now. Works fine even in python3 even though it's no longer recommended.

PR #6142, opened 2015-08-18
Fix POSTFIELDS settings

vkuznet:


vkuznet:
It seems to me that now unit tests complains about different from pycurl issues.

ericvaandering:
Yes. Anything that wants port 8888 is unstable. Either I try to figure out what it is, or we turn off all those tests in Jenkins.

ticoann:
Hi Eric, changing port number would help? What do you mean by unstable? connection drops?

ericvaandering:
If you look into the tests you will see things like “port 8888 unavailable”. I assume something else is occupying the port. I don’t know if changing ports to something else static would help.

I also see oscillating tests where #’s come back incorrectly. (792 != 800). I have no idea what could cause that.

Eric



On Aug 18, 2015, at 11:30, ticoann <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi Eric, changing port number would help? What do you mean by unstable? connection drops?

—
Reply to this email directly or view it on GitHub<https://github.com/dmwm/WMCore/pull/6142#issuecomment-132267159>.




Eric W. Vaandering
Computing Physics Developer

Scientific Computing Division
630 840 6678 office

mmascher:
Hi all,

about this change, the CRAB3 code is using the pycurl_manager and this breaking change that has an impact in a lot of places (client, task worker, etc).

Not a big deal, we can change the code and remove `urllib.urlencode` when we call a `post`. But since we are changing this may I ask you why `params` is encoded for `POST` and `GET` but not for `DELETE` and `PUT` here https://github.com/vkuznet/WMCore/blob/4638c5cd17de15e32e1663a296e47609c45a5241/src/python/WMCore/Services/pycurl_manager.py#L96?

Can't we use `encoded_data` here https://github.com/vkuznet/WMCore/blob/4638c5cd17de15e32e1663a296e47609c45a5241/src/python/WMCore/Services/pycurl_manager.py#L96 instead of `params` so that clients of this library will always have to pass a dictionary and never worry to call `urllib.encode`? Otherwise we go from bad to bad instead of improving the situation.

Maybe we should add a unit test and mention in the comments that this code is used by CRAB3 and any change in the interface of the library has an impact there as well?

FYI @emaszs @juztas @AndresTanasijczuk you will get something like that if you update WMCore:

```
2015-08-27 10:49:03,348:ERROR:MasterWorker:Task Worker could not update a task status: not a valid non-string sequence or mapping object
Configuration parameters={'status': 'HOLDING', 'getstatus': 'KILL', 'limit': 10, 'subresource': 'process', 'workername': 'erupeika0'}
  File "/afs/cern.ch/user/e/erupeika/private/github/repos/CRABServer/src/python/TaskWorker/MasterWorker.py", line 160, in _lockWork
    self.server.post(self.restURInoAPI + '/workflowdb', data = urllib.urlencode(configreq))
  File "/afs/cern.ch/user/e/erupeika/private/github/repos/CRABServer/src/python/RESTInteractions.py", line 69, in post
    return self.makeRequest(uri = uri, data = data, verb = 'POST')
  File "/afs/cern.ch/user/e/erupeika/private/github/repos/CRABServer/src/python/RESTInteractions.py", line 113, in makeRequest
    capath=caCertPath)#, verbose=True)# for debug
  File "/afs/cern.ch/user/e/erupeika/private/github/repos/WMCore/src/python/WMCore/Services/pycurl_manager.py", line 156, in request
    ckey, cert, capath, verbose, verb, doseq, cainfo)
  File "/afs/cern.ch/user/e/erupeika/private/github/repos/WMCore/src/python/WMCore/Services/pycurl_manager.py", line 87, in set_opts
    encoded_data = urllib.urlencode(params, doseq=doseq)
  File "/afs/cern.ch/cms/slc6_amd64_gcc491/external/python/2.7.6-cms/lib/python2.7/urllib.py", line 1315, in urlencode
    raise TypeError
TypeError: not a valid non-string sequence or mapping object
```

vkuznet:
I fixed issue with pycurl for PUT/POST/DELETE request in this PR: https://github.com/dmwm/WMCore/pull/6176

mmascher:
Thanks Valentin, I will test this, but it should work.

Another comment, any good reason to change the type of the exception that
is raised in case of errors from httplib.HTTPException to
urllib2.HTTPError[1]?

This is a bit painful for CRAB3: apart from all the places where we are
catching HTTPException that would have to be changed (requires some time
but not a big deal actually), but we are advertizing it in a twiki [2] for
users who want to use the crab client as a library that we can raise
HTTPException in case of issues, and this is something I would like to stay.

We could always catch urllib2.HTTPError and re-raise httplib.HTTPException
, or use a crab3 version of the pycurlmanager, but that would be ugly in my
opinion.

[1]
https://github.com/vkuznet/WMCore/commit/a6002ffa91124fa5da813a0b9c35aa5d0f6a456d#diff-1120a2d619920d0b79a217f9ee2b164eL163
[2]
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRABClientLibraryAPI#Debugging_errors_with_crabComman

vkuznet:
Marco,
the reason I changed httplib.HTTPException to urllib2.HTTPError since later
provides concise error where I can initialize it with code,url,params,headers,etc.
fields. The former only allow to create a message for the exception.
But I understand the legacy issue and will revert changes to use
HTTPException instead.
Best,
Valentin.

mmascher:
Thanks!

PR #6145, opened 2015-08-19
fix the lumi calcuation

### LINKED ISSUE #5761: wrong total lumi calculation in task chain.
Review the code below.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py#L58

Following gives double total lumis than what it supposed to be.
https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/pdmvserv_task_TOP-RunIIWinter15wmLHE-00015__v1_T_150404_183555_8973

ticoann:
fixes #5761 - hopefully

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
@@ -5,7 +5,7 @@ WorkQueue splitting by block
 """
 __all__ = []
 
-
+from __future__ import division
 
 from WMCore.WorkQueue.Policy.Start.StartPolicyInterface import StartPolicyInterface
 from WMCore.WorkQueue.WorkQueueExceptions import WorkQueueWMSpecError, WorkQueueNoWorkError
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan could you review? I found the suspicious logic. It seems remainderEvents got added twice when calculating the number of lumis. (But no idea this is the cause of the problem)

ericvaandering:
This looks like a good place to start using `from __future__ import division`

amaltaro:
I had no idea that the lumi calculation happened here :-) Patch looks good to me!
Did you manage to test it using the same request provided in the issue?

ticoann:
> This looks like a good place to start using from __future__ import division

Thanks Eric, I added that. There is no auto changing for this? :smile: 
But one thing I am concerned is if this module get imported in other package, won't it break the division there? (If some code actually intended to get  integer (3 / 2 = 1)

ticoann:
> Did you manage to test it using the same request provided in the issue?

No I didn't test. should I clone the same workflow do you know any test spec which I could test?

ericvaandering:
The problem with "fixing" division is two-fold. First, python2 is broken. That's why they are fixing it. Unlike a strongly typed language where what you get is predictable, this is not simply because it depends on what types you may pass into a function. Second, there is a way to "fix" python3 to do the braindead thing python2 does, but we should be smarter than that.

That's why I want every new file to include the future division (Jenkins will enforce this) and it's a good idea to go back to old ones that do division and *think* about what to do and how to future proof it. Les work in the future and probably fewer bugs now. That's why I suggest this.

The import will only change the behavior of this one file, not the whole python interpreter.

PR #6146, opened 2015-08-19
remove dbs2 code

ticoann:
removing some of DBS2 code. There are more in WMBSFeeder and DBSUpload. but I think those need to be deleted all together. I am checking whether anyone is using them.


ticoann:
retest this please

ericvaandering:
retest this please

PR #6150, opened 2015-08-20
srmcp requires to create directory before copy. Fixes #6149

### LINKED ISSUE #6149: Possible bug with SRM (v2?) plugin
It seems SRM plugin tries to stage out files before explicitly creating the target directory. At least this is what Justas reported about recent issues at Caltech when the stage out fallbacks to srm protocol.

juztas:
tested with pdb and copied site-local-config of caltech. So seems is never calling this function, so now before copying file it will try to check if all needed directories exists.

Test code:

```
touch /tmp/justas.txt
voms-proxy-init -voms cms

python >>
import os
import WMCore.Storage.StageOutMgr as StageOutMgr
from WMCore.Storage.Registry import retrieveStageOutImpl



local_stageout_mgr = StageOutMgr.StageOutMgr()
impl = retrieveStageOutImpl("srmv2")
local_stageout_mgr.numberOfRetries = 1
local_stageout_mgr.retryPauseTime = 1

file_for_transfer = {'LFN': '/tmp/justas.txt', 'PFN': '/store/temp/user/jbalcas/justas.txt'}
local_stageout_mgr(file_for_transfer)
```

### DIFF
--- a/src/python/WMCore/Storage/Backends/SRMV2Impl.py
+++ b/src/python/WMCore/Storage/Backends/SRMV2Impl.py
@@ -55,7 +55,6 @@ class SRMV2Impl(StageOutImpl):
 
         """
         targetdir = os.path.dirname(targetPFN)
-
         if self.stageIn:
             # stage in to local directory - should exist but you never know
             if not os.path.exists(targetdir):
@@ -122,6 +121,7 @@ class SRMV2Impl(StageOutImpl):
         Build an srmcp command
 
         """
+        self.createOutputDirectory(targetPFN)
         result = "#!/bin/sh\n"
         result += "REPORT_FILE=`pwd`/srm.report.$$\n"
         result += "srmcp -2 -report=$REPORT_FILE -retry_num=0"


hufnagel:
Please change this. createOutputDirectyory() is already called upstream

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L171

you should check why it has no effect there.

hufnagel:
Worse, createOutputDirectory() has no business being in the createStageoutCommand method as that circumvents the stageIn checks.

juztas:
I checked now and it did not called that backend plugin function, still looking which function it called.

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L170

hufnagel:
That line is called here

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutMgr.py#L299

juztas:
ok, founded.
So during local stageout https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L170 this evaluates to true, because 
```
(Pdb) self.deferDirectoryCreation()
False
(Pdb) self.stageIn
False
```
But during fallback stageout it evalueates to false, and directory creation is not called:
```
(Pdb) self.deferDirectoryCreation()
True
(Pdb) self.stageIn
False
(Pdb) 
```

hufnagel:
Fallback stageout does not initialize self.directoryErrorCodes correctly ?

juztas:
Fallback:
```
(Pdb) self.directoryErrorCodes
(1,)
(Pdb) self.deferDirectoryCreation()
True
(Pdb) self.stageIn
False
```
and on fallback https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L170 this `if` evaluates to false and it is not calling directory creation.

juztas:
Maybe ~6 years ago this was not in place for fallback stageout to different se?

hufnagel:
Why did you merge this ? This is basically a hack, it should not work like this. Plus it now calls createOutputDirectory() for all stageins, which I don't know what it'll do but probably nothing good.

hufnagel:
Ie. this should be unmerged abd fixes properly.

juztas:
I agree with Dirk.

hufnagel:
Let me take a look, this might be as easy as the initialization being skipped somehow for fallback stageout...

hufnagel:
Ok, the way this code is designed is that the underlying assumption here is that the output directory will already exist in most cases and therefore always calling a createDir for it is expensive and should not be done.

If a stageout method allows detecting whether a stageout failed due to a non-existing output directory (that's what the error codes are for), we use that to detect a missing output dir on the initial stageout failure and call createOutputDirectory().

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L197

As you can see this isn't even a normal stageout retry, this is done immediately. So this actually works as designed here, the interesting question is why you don't get this retry on the fallback stageout.

And another question, why is the local stageout not using this as intended in the plugin ? It also should not call createOutputDir(). I checked the way local and fallback stageout are called and they look identical to me

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutMgr.py#L270

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutMgr.py#L231

Justas, in what context are you calling this ? Standard WMAgent or in some other way ?

And as a last note, the SRM stageout plugin has the error code 1 for detecting a missing output dir, which is bit iffy because it assumes every stageout failure is due to a missing output dir and will retry accordingly. That's not very efficient either. If we want to have a quick fix here we could just remove the list of error codes from the SRM plugin, ie. it'll always call createOutputDir().

juztas:
It is a real production job. So what happens is that local stageout uses smrv2-lcg and it is timing out, here comes fallback, which is srmv2 plugin, but it is not calling directory creation. Julian provided full log file, and this happens. 
Whenever local stageout fails, it goes to fallback stageout and this skips directory creation due to this if: https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L170

hufnagel:
Ah, so local stageout and fallback stageout use different implementations ! Ok, so we can ignore the local stageout here and just look at the fallback.

As I said before, skipping directory creation for srmv2 is intentional, someone designed it that way. It is supposed to be retried immediately though with directory creation

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/StageOutImpl.py#L197

You don't see that happening in the log ? The only reason I can imagine why this doesn't happen is that the error code 1 is actually incorrect, ie. the srmv2 stageout throws a different error code if the directory doesn't exist. In which case we should find out what that error code is and add it to the list of error codes. Also, that 1 really shouldn't be there, but I am bit hesitant to remove it as a lot of sites use the srmv2 plugin and removing the 1 could cause problems if it actually was the correct error code for some versions of the srmv2 stageout.

In the end it might just be easier to completely remove the error codes in the srmv2 plugin, which would make it always call createOutputDir().

hufnagel:
@juztas , I think we should just go ahead and remove the error code list

https://github.com/juztas/WMCore/blob/11cf54da5978b48bd81212be5f6d071af48d83c6/src/python/WMCore/Storage/Backends/SRMV2Impl.py#L31

completely and always call the directory creation when using srmv2 stageout. Having a 1 there is really is bad. This is supposed to be a specific error code that is only and always returned when the target directory does not exist. If no such error code exists, this shouldn't just be abused as a general purpose retry, we have the regular stageout retries for that.

PR #6155, opened 2015-08-21
Fixed TotalInputLumis calculation for Block policy

amaltaro:
From all the use cases, ReReco with parents was being miscalculated (though it seemed to me a general issue). Also fixed a few pylints complaints.

@ticoann , I tested it in my VM and now we get the right number :) Nevertheless, I want to leave a few tests running this night.
Please include it in the next testbed tag (for Monday)


ticoann:
Just a reminder. This need to be added for the HG1509 deployment

PR #6156, opened 2015-08-21
ignore null response

ticoann:
Firefox get null response sometimes, but not in Chrome.
Not sure this will fix the problem since it doesn't happen in my vm but at least it doesn't break the current code.


ericvaandering:
retest this please

PR #6082, opened 2015-07-29
Add unit tests to FWJR generator and dashboard reporting on local vs. fallback

ericvaandering:
Add FWJRs for local and remote reads of two file workflows, unit tests to FWJR generator and dashboard reporting on local vs. fallback

PR #6158, opened 2015-08-23
fix indentation typo in dashboard reporting

hufnagel:
Fixes a an indentation problem introduced with #6082 

@ericvaandering please review as that pull request was yours.


hufnagel:
Btw, at the very least this means for 1.0.9.pre3 to 1.0.9.pre5 the dashboard reporting code will throw an error for every single job in the JobAccountant. Didn't investigate further if its only cosmetic or if dashboard reporting is completely broken for these tags or if there are further consequences also for overall job accounting.

PR #6160, opened 2015-08-24
HDFS is a case-sensitive comparison for checksum (GFAL2)

juztas:


ericvaandering:
Do we know this wouldn't break other file systems?

hufnagel:
Question is, do we have any non-HDFS filesystems where we use the GAFL2Impl ? All of them potentially could have problem if we change adler32 to ADLER32...

juztas:
sprace, cern succeeds except all US sites. crab3 gfal2 with implemented checksum succeeds on non US sites.
About other: https://jira.opensciencegrid.org/browse/SOFTWARE-2006

juztas:
about pylint complains, afaik all sites uses backend but not plugins (SAM uses plugins?) and I remember that @hufnagel wrote on hypernews about rewriting plugins if no one is using it.

juztas:
So only US sites (T2) were failing and where is a fix in new gridftp-hdfs repo from osg-testing. Checked several letter combinations for checksum and with/without it and all succeeded. Please see: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/884/1/1/1/1/1/1/1/1/2/1/1/1/3/1/1.html
Closing as not needed to change.

ericvaandering:
Thanks Juztas.

PR #6161, opened 2015-08-24
Fixes ACDC AcquisitonEra, etc setting

### LINKED ISSUE #6098: Problem with ACDC to recover merge failures for TaskChain
The issue is the following: we normally set AcqEra and ProcStr to a dict made of TaskName:Value. The problem assigning wfs with these settings is that we are not specifying the AcqEra/ProcStr for the merge task (since it's out of our control), and then merge jobs will create files with None AcqEra and ProcStr.

The workaround for this issue would be setting AcqEra and ProcStr to a simple string instead. However it won't work for all the cases and, one of these cases would be if we need to recover *merge* jobs for Task1, but the workflow has a Task2 that must have != AcqEra and/or ProcStr.

ticoann:
fixes #6098 (Don't merge yet - still have a problem)

### DIFF
--- a/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
+++ b/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
@@ -316,7 +316,7 @@ class Assign(WebAPI):
         if kwargs.get("ProcessingString", None):
             processedDatasetParts.append("ProcessingString")
         for field in processedDatasetParts:
-            if type(kwargs[field]) == dict:
+            if isinstance(kwargs[field], dict):
                 for value in kwargs[field].values():
                     self.validate(value, field)
             else:
@@ -325,9 +325,9 @@ class Assign(WebAPI):
         # Set white list and black list
         whiteList = kwargs.get("SiteWhitelist", [])
         blackList = kwargs.get("SiteBlacklist", [])
-        if type(whiteList) != list:
+        if not isinstance(whiteList, list):
             whiteList = [whiteList]
-        if type(blackList) != list:
+        if not isinstance(blackList, list):
             blackList = [blackList]
         helper.setSiteWildcardsLists(siteWhitelist = whiteList, siteBlacklist = blackList,
                                      wildcardDict = self.wildcardSites)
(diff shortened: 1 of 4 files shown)


ticoann:
@amaltaro, Please review. Still need more test. Please don't merge yet.

ticoann:
@amaltaro, Alan, I am not sure how to create the task chain workflow and only to make merge job fail. Could you help me with the test? Thanks,

amaltaro:
Seangchan, AFAIK creation of Resubmission workflows is working fine and inheriting the right information. From the quick look I had, it fixes only request creation but not assignment, right?
If so, the problem happens during assignment when merge tasks default to None.

About the test, we can use this template
https://github.com/dmwm/WMCore/blob/master/test/data/ReqMgr/requests/DMWM/TaskChain_MC.json
but then we need to watch those jobs from the agent and kill the merge jobs when they make it to condor (the hard way :)).

ticoann:
Hi Alan, do you know which script they are using to assign Resubmission workflow?
So when they assign it they don't pass acquisition era, etc. The way it (merge task acdc) get assigned different from other tasks (acdc)?

It is weird that only Merge task ACDC has the problem not others.
Anyway, I updated the code I will try to test it. (reqmgr2-dev.cern.ch) Could you run the test there as well when you get a chance.  Thanks.

Seangchan

amaltaro:
Sure, I'll run a test and kill some jobs as soon as I get at CERN.

ACDCs are usually assigned via webpage, but not TaskChain since we need to provide:
"AcquisitioEra": {'Task1Name': 'Tas1AcqEraValue', 'Task2Name': 'Tas2AcqEraValue', ..}
and the same for ProcessingString. (and that's when the merge task is lost)

ticoann:
Thank you very much Alan,

> ACDCs are usually assigned via webpage, but not TaskChain since we need to provide:
> "AcquisitioEra": {'Task1Name': 'Tas1AcqEraValue', 'Task2Name': 'Tas2AcqEraValue', ..}
> and the same for ProcessingString. (and that's when the merge task is lost)

Do you know which script is used for assigning the TaskChain ACDC?
For ACDC, why  do we need above in assignment? ACDC workflow is not TaskChain anymore, isn't it?
(since ACDC runs over only one task.)

amaltaro:
Can you believe I ran 3 workflows and I failed to kill a single merge job?!?! :-(
I even set this cronjob to get them automatically removed 
*/3 * * * * condor_rm -const 'CMS_JobType =?= "Merge"'
but did not work. I'll inject more requests.

Now answering your last message, I believe this is what they use to assign ACDC TaskChain
https://github.com/CMSCompOps/WmAgentScripts/blob/master/assignProdTaskChain.py

>  For ACDC, why do we need above in assignment? ACDC workflow is not TaskChain anymore, isn't it?

This is not true for TaskChains :) It will still have the same structure as a TaskChain, well, actually starting the task where it was resubmitted from. So exactly the same if recovering Task1, slightly different if recovering Task2. See this example:
https://reqmgr2-dev.cern.ch/reqmgr/view/showWorkload?requestName=amaltaro_ACDC_TaskChain_Task_Multicore_Seangchan_Bugfix6161_TEST_v3_150825_163959_4395

ticoann:
fixed by #6266

PR #6166, opened 2015-08-25
Make dir creation correct in SRMV2 impl

juztas:
@hufnagel please, review 
For others: https://github.com/dmwm/WMCore/pull/6150#issuecomment-134251674

### DIFF
--- a/src/python/WMCore/Storage/Backends/SRMV2Impl.py
+++ b/src/python/WMCore/Storage/Backends/SRMV2Impl.py
@@ -28,7 +28,6 @@ class SRMV2Impl(StageOutImpl):
 
     def __init__(self, stagein=False):
         StageOutImpl.__init__(self, stagein)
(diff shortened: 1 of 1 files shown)


hufnagel:
You don't need the line defining self.directoryErrorCodes at all, it is already defined in the base class.

juztas:
updated

hufnagel:
Looks good now

ericvaandering:
retest this please

ticoann:
retest this please

amaltaro:
Justas, Dirk, did you manage to run any real test for this patch? Otherwise I'll simply run the standard tests and then push it to production.

hufnagel:
I don't think a test is needed. This effectively removes on line of code and that line of code did something (skip directory creation for stageout). Now with the line removed we simply don't skip it anymore.

But Justas might have tested it too.

PR #6167, opened 2015-08-25
Handle requests without prep_id

amaltaro:
Diego G. reported a problem with WMDataMining in testbed. From the traceback, it seems to fail when a request does not have Prep_Id, it should fix that issue

@ericvaandering please review. I could not test it in my VM because it fails to generate a SSO cookie (?)
otherwise I'll wipe my VM in make a fresh installation this evening just to make sure there won't be another bug after this one.

### DIFF
--- a/src/python/WMCore/WMDataMining/Utils.py
+++ b/src/python/WMCore/WMDataMining/Utils.py
@@ -145,7 +145,7 @@ def gatherWMDataMiningStats(wmstatsUrl, reqmgrUrl, wmMiningUrl,
                 log.error("Could not find any status from workflow: %s" % wf) # Should not happen but it does.
 
             # Remove a single  task_ from the start of PREP ID if it exists
-            if prep_id.startswith('task_'):
+            if prep_id and prep_id.startswith('task_'):
                 prep_id.replace('task_', '', 1)
 
             # Can be an empty list, full list, empty string, or non-empty string!


ericvaandering:
Is the failure at this exact place? Like it's trying to do "starts with" on a None?

ericvaandering:
In any case, looking at this more carefully, I think it's better to move this test and action after line 92 (before we try to contact McM)

amaltaro:
That's right Eric, it happens in this exact line
```
2015-08-25 10:27:26,257:ERROR:CherryPyPeriodicTask:     if prep_id.startswith('task_'):
2015-08-25 10:27:26,257:ERROR:CherryPyPeriodicTask: AttributeError: 'NoneType' object has no attribute 'startswith'
```

ericvaandering:
Alright, then this is a good patch. @ticoann please merge.

ericvaandering:
retest this please

ticoann:
retest this please

ericvaandering:
@ticoann this is good to merge. I'll address the idiom thing in another PR since Alan did this one.

amaltaro:
Sorry guys, I completely forgot to update this one before going to bed :(

ticoann:
Ah, you mean pylint recommendation? Don't worry about it. :smile: we can always add that later.

amaltaro:
Yes, I see Eric updated it in 6171. Thanks!

PR #6168, opened 2015-08-25
Query requests by state in clean acdc thread

amaltaro:
Fixes the couchCleanup issue reported by Diego G. today (in testbed).
It was failing to get the couch view (I did not find the reason), however, I changed it to get requests by status (also removed 'announced' from the list, so we keep only final state).

@ticoann please review. I'll test it in my VM now.


ticoann:
Hi Alan, If you get the view by status with archived status there will be too many request ~40K retrieved that is why it is geting the original request in ACDC first. That is why it is getting the byrequest view first and filters within the retrieved data. If the view fails to get, there must be some config settings messed up. I will take a look.

amaltaro:
Hmmm, I didn't test against production. Doing so now takes around 3 seconds to get a response back.

If getCouchView() takes a second or more to respond, then I guess we don't need to worry about it. Moreover, requestByStatus returns a clean list of requests to just be used.

Anyways, if you don't manage to find the root cause, we have a fix available :-D

ericvaandering:
retest this please

ticoann:
Alan, I found the problem and create the new patch #6170. closing this one.

PR #6170, opened 2015-08-25
fix cleanup Couch

ticoann:
fixes #6168  Somehow stale option passing is not working with post call. removing the option for now.


ticoann:
retest this please

ericvaandering:
Are you sure these errors aren't real?

PR #6176, opened 2015-08-28
Fix issues with pycurl manager for POST/PUT/DELETE requests

vkuznet:
Change exception on failed requests; use future print statement; require params to be passed as dictionary


vkuznet:
Some unit tests may require the recipe from this PR: https://github.com/dmwm/WMCore/pull/6180 since it assign test_authz_key object for test server.

PR #6181, opened 2015-08-28
Assign test_authz_key in unit test

vkuznet:
Fix other tests who inherit test_authz_key problem.


ticoann:
@vkuznet, Valentin, this can be closed without merging it as well, right?

vkuznet:
yes, please close now I re-factor code properly to avoid globals.

PR #6184, opened 2015-08-31
pycurl_manager is no longer decoding the body of HTTP responses

mmascher:
The bug was found by @jmarra13 during the validation of the crab3 September release (https://github.com/dmwm/CRABServer/issues/4954#issuecomment-136322439)


mmascher:
Since this is quite urgent I will patch the crabclient WMCore "by hand", no need to hurry up with a tag

ticoann:
Thanks Marco

PR #6188, opened 2015-09-01
Adapt error message in StatusPoller

amaltaro:
Decreased timeout for jobs that go into an error state.
Import error message from a single place
Aesthetic changes and pylint fixes in general

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -1,4 +1,7 @@
 #/usr/bin/env python
+#pylint: disable=E1101,C0103
+#E1101: ignore no member complaints here (created dynamically)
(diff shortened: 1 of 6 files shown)


amaltaro:
Perhaps we should disable this E1101 pylint check in etc/WMAgentConfig.py ?

ericvaandering:
I think we can. You can also re-enable it later in the code if there is a need. Clearly PyLint does not understand our configs. :-)

amaltaro:
Much better now :-) It should be good to go, but I'll give you my green light tomorrow Seangchan

amaltaro:
Also ok to be merged now. Let me know the new branch to make a PR there.

Issue #5785, opened 2015-04-15
Standardize on JSON libraries


ericvaandering:
As explained here: https://docs.python.org/dev/whatsnew/2.7.html the python standard library for JSON got much faster in python 2.7 so there is no advantage from cjson or simplejson. We should standardize and simplify or remove the JsonWrapper class.

PR #6191, opened 2015-09-02
All set of changes to make Api_t/Simple_t/Daemon_t unit test to work

vkuznet:
PLEASE DO NOTE MERGE until the work is done

### DIFF
--- a/test/python/WMCore_t/REST_t/Api_t.py
+++ b/test/python/WMCore_t/REST_t/Api_t.py
@@ -1,17 +1,26 @@
-from cherrypy.test import helper
-from cherrypy import expose, response, config as cpconfig
+# system modules
+import cjson, re, zlib
+import cherrypy
+from cherrypy.test import webtest
+from cherrypy import response
+from multiprocessing import Process
+
+# WMCore modules
 from WMCore.REST.Server import RESTApi, RESTEntity, restcall, rows
 from WMCore.REST.Test import setup_test_server, fake_authz_headers
+from WMCore.REST.Test import fake_authz_key_file
 from WMCore.REST.Validation import validate_num, validate_str
 from WMCore.REST.Error import InvalidObject
 from WMCore.REST.Format import RawFormat
 from WMCore.REST.Tools import tools
 import WMCore.REST.Test as T
-import cjson, re, zlib
 
 gif_bytes = ('GIF89a\x01\x00\x01\x00\x82\x00\x01\x99"\x1e\x00\x00\x00\x00\x00'
              '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
              '\x00,\x00\x00\x00\x00\x01\x00\x01\x00\x02\x03\x02\x08\t\x00;')
+
+FAKE_FILE = fake_authz_key_file()
+PORT = 8888
 
 class Simple(RESTEntity):
     def validate(self, *args): pass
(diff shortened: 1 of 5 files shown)


ericvaandering:
retest this please

ericvaandering:
Ok, @vkuznet this should be your limited test coming up. There are a couple of things from Alan ahead of you in the queue, so it may be a while before it completes.

test/python/WMCore_t/REST_t/Api_t.py
```diff
@@ -1,18 +1,27 @@
-from cherrypy.test import helper
-from cherrypy import expose, response, config as cpconfig
+# system modules
+import cjson, re, zlib
```
ericvaandering: I think json is every bit as good as cjson now (or actually is cjson)

vkuznet: I can change it to json, but I think it needs to be clarified by Seangchan.

vkuznet:
Seangchan, as far as I concern the changes are in good shape, but I didn't touch cjson since it was present in a code. Therefore you can merge these changes or let me know if you want me to convert cjson to json and then we'll merge this PR.

ericvaandering:
Let's go as is. I missed that cjson was original.

On Sep 8, 2015, at 10:59, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:


Seangchan, as far as I concern the changes are in good shape, but I didn't touch cjson since it was present in a code. Therefore you can merge these changes or let me know if you want me to convert cjson to json and then we'll merge this PR.

ticoann:
Is json is as good as cjson in python 2.6.8? Or after python 2.7?
@vkuznet, this can be merged right? (Just double checking :smile: )

vkuznet:
The cjson is still better performance wise, and, yes we can merge this one.

ericvaandering:
I might have been confusing simplejson with json (lots of confusion to be had) but this suggests there is little difference anymore: http://stackoverflow.com/questions/706101/python-json-decoding-performance

Takeaway is we shouldn't be writing any new code that uses anything other than the standard library.

vkuznet:
Eric,
you may read this thread on python mailing list I participated long time ago
http://bugs.python.org/issue6594
The speed is not an issue among the two and rather the memory footprint,
especially loading nested JSON's such as provided by PhEDEx APIs.

But I agree apart from some corner cases we should/must stick to standard
library.
Valentin.

ticoann:
One other thing I remember was cjson has bug which convert all the unit code to str. However that actually served as a feature in WMCore. So it might break some of the code in WMCore if we switch to json.  Yes I agree to move to standard library as well unless there is big advantage using cjson

PerilousApricot:
Not only that, JSON requires that all strings (including dict keys) are unicode. cjson (used to, at least) convert dict keys to 'string' and keep all other strings as unicode, which was a catastrophe when we put it against the HTTP frontend validation.

ericvaandering:
Fortunately our usage seems to be about 80% json, 10% cjson, and 10% simplejson.

vkuznet:
Simplejson should just go, out of questions.

PerilousApricot:
Oh yeah. All of the simplejson/WMCore.JsonWrapper stuff is almost certainly legacy from when we were on 2.4 and json wasn't in stdlib.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Wrappers/JsonWrapper/__init__.py

ericvaandering:
Yeah, I know. We might even have an issue to get rid of that crap. If not, we should

ericvaandering:
#5785

PR #6195, opened 2015-09-02
Fix memory leak

ticoann:


ticoann:
Sorry, Wrong version closing it

PR #6196, opened 2015-09-02
fix the memory leak in wmstats

ticoann:
1.0.9_cmsweb version of patch (#6194)


ticoann:
retest this please

ticoann:
@ericvaandering, Eric, is this fail valid? It shouldn't fail any unittest since there is None for this.

ericvaandering:
That's the result of my pull request being on master. Don't worry about it

PR #6077, opened 2015-07-28
add option for EventAwareLumiBase to observe a 47h wall time estimate cap

hufnagel:
The Tier0 has very unreliable timePerEvent and sizePerEvent estimates as running conditions can change greatly, leading to huge differences within the same dataset from run to run. To not create jobs that can never run, cap the wall time and disk estimates.

PR #6200, opened 2015-09-03
Apply glideIn restrictions to resource estimation

### LINKED ISSUE #5738: Do not submit jobs asking for too much resources
Now that glideInWMS is enforcing "MaxWallTimeMins" and "RequestDisk", we should change WMAgent NOT to submit greedy jobs to condor, otherwise those will be pending forever.
Instead, these jobs should be marked as jobfailed (or submitfailed) with a clear message.

It should be configurable and I guess it needs further discussion, but here is a first idea:
 - don't submit if job asks for more than 48h MaxWallTimeMins
 - don't submit if job asks for more than (27GB * RequestCpus) RequestDisk

amaltaro:
After being beat by PyCharm, I went back to the terminal :-)
Fixes #5738 We basically define in the agent min/max limits for diskUsage and
wallTime and these are going to be enforced. 

First test was Ok, will run more using different algos/cores.

@hufnagel do you want me to remove your previous patch (#6077)?

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -194,6 +194,10 @@ config.JobCreator.pollInterval = 120
 config.JobCreator.jobCacheDir = config.General.workDir + "/JobCache"
 config.JobCreator.defaultJobType = "Processing"
 config.JobCreator.workerThreads = 1
+# glidein restrictions used for resource estimation
+config.JobCreator.GlideInRestriction = {"MinWallTimeSecs": 1 * 60 * 60,    # lower limit (or default)
+                                        "MaxWallTimeMins": 47 * 60 * 60,   # pilot lifetime is usually 48h
+                                        "RequestDiskKB": 20 * 1024 * 1024} # unit KB, site limit at 27GB
 
 config.component_("JobSubmitter")
 config.JobSubmitter.namespace = "WMComponent.JobSubmitter.JobSubmitter"
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-#pylint: disable=W0613, W6501
 """
 The JobCreator Poller for the JSM
 """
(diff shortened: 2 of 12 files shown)


hufnagel:
The cap for requested disk is nCores dependent, ie. nCores * MaxRequestDiskKB. Otherwise the logic looks ok and usable for the Tier0. Thefore you can remove some code I added in my previous patch.

 - capResourceEstimates method from DataStructs.Job
 - capJobTime and capJobDisk code in the EventAwareLumiBased job splitter
 - capJobTime and capJobDisk code in the PromptReco spec

amaltaro:
> The cap for requested disk is nCores dependent,

Oups, I missed that detail. Will fix it soon.
About the previous patch, ok, I'll push it all together here. Thanks!

amaltaro:
Code updated, I'll make a new commit later on with pep8 fixes.

ericvaandering:
So is there not a pull against master for this? Jenkins is failing because of the merge against 109_wmagent issue. Should have it fixed now.

amaltaro:
Sorry, not yet Eric. Should I always make a PR for master first?

ericvaandering:
No, just wondering.

ericvaandering:
retest this please

hufnagel:
Reviewed the code, never use an empty dict as the default value for a parameter in the function definition. Pylint should complain about it too.

Reason its dangerous is that any modifications to said default parameter will survive to the next function call, ie. if you then still use the default, it'll now be the value that parameter had at the end of the previous function call. The dictionary is not reinitialized between function calls.

You wouldn't be exposed to this with your code here, but please change the default value to None.

amaltaro:
sigh... I wrongly rebased this PR and now it;s a mess! I will create a new one and close this one

amaltaro:
Yeah, it did complain. But I kept an empty dict for clarity, so if someone look at that var, he knows right away that it's supposed to be a dict.

hufnagel:
No, please don't do that. None is a perfectly appropriate value for any parameter.

hufnagel:
The problem here is that dicts are passed by reference, not value. So no deep copy. Which means your {} default is initialized once when the class is loaded and then is just reused and reused and reused.

amaltaro:
Yes, but in this case that dict will only change when the component gets restarted (class loaded) and no code is supposed to update it. Anyways I will initialize it to None then.

amaltaro:
New PR made, closing this one. Sorry for the mess

hufnagel:
I wouldn't insist if we weren't bitten by this before :-). I can tell you, this is a total pain to debug if you don't expect it.

Plus the code might be ok now, but you never know how its going to evolve. I'd rather not leave booby traps lying around...

amaltaro:
Thanks Dirk! I updated master as well.

ericvaandering:
Could not agree more with Dirk. You have the same issue with lists as parameters. So while it's ugly to have 

```python

if myDict is None:
    myDict = {}

```
it's the only safe way.

amaltaro:
Ok, will consider it from now on :-)

PR #6203, opened 2015-09-03
New Redigi template to exercise other spec param

amaltaro:
Running it now in testbed to check whether TpE and SpE works in a step base.


amaltaro:
Ready to be merged, no hurry though :)

PR #6218, opened 2015-09-07
Adding phedex-node to localStageOut report.

emaszs:
We need to have the phedex-node value of site-local-config.xml in order to avoid querying PhEDEx if possible. This is the issue: https://github.com/dmwm/CRABServer/issues/4895.

Also added an integration test to see if site-local-configs actually match what PhEDEx is providing.

### DIFF
--- a/test/python/WMCore_t/Storage_t/SiteLocalConfig_t.py
+++ b/test/python/WMCore_t/Storage_t/SiteLocalConfig_t.py
@@ -11,8 +11,11 @@ import unittest
 from WMQuality.TestInit import TestInit
 from WMCore.WMBase import getTestBase
 
-from WMCore.Storage.SiteLocalConfig import SiteLocalConfig
+from WMCore.Storage.SiteLocalConfig import SiteLocalConfig, SiteConfigError
 from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig
+from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
+
+from nose.plugins.attrib import attr
 
 class SiteLocalConfigTest(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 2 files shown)


test/python/WMCore_t/Storage_t/SiteLocalConfig_t.py
```diff
@@ -156,6 +159,37 @@ def testLoadingConfigFromOverridenEnvVarriable(self):
...
+        for node in nodes:
+            node_map[str(node[u"se"])] = str(node[str(u"name")])
+        
+        for d in os.listdir("/cvmfs/cms.cern.ch/SITECONF/"):
+            # Only T0_, T1_... folders are needed
+            if d[0] == "T":
+                os.environ['WMAGENT_SITE_CONFIG_OVERRIDE'] ='/cvmfs/cms.cern.ch/SITECONF/%s/JobConfig/site-local-config.xml' % (d)
+                try:
+                    slc = loadSiteLocalConfig()
+                except Exception as e:
```
ericvaandering: Please take out this try/except. You are guaranteeing this unit test will succeed and spew a bunch of junk to stdout where no one will ever see it.

mmascher: Hi Eric,

my intention with this is to use it as an `integration` unit test (run by hand) to discover faulty/broken SLC at sites (see the email I sent yesterday to the site support team). In that sense I would even improve the number of printouts/type or defect discovered by the unit test (I would also like to know which sites have not set the phedex-node attribute in their SLC).

I know these things belongs to the SST and a SAM test would be more appropriate, but imho the codebase should stay here since WMAgent and CRAB3 uses the WMCore code at runtime, and it makes also sense to test the code run by them. Maybe I can talk to Ali next week and see if they can recycle the unit test in some way?

ericvaandering: Sorry, then trap the specific errors you want or let it fail. Trapping every exception is bad form.

ericvaandering:
@alexanderrichards can you take a look at the meat of this patch (not the test) and see if this is something that would conflict with what you are working on?

alexanderrichards:
@ericvaandering Yes this patch duplicates changes in my patch.

mmascher:
Which patch is it? Did you call this `phedex-node`?

alexanderrichards:
yes but my patch has not yet been merged, the link I gave was for a PR.

alexanderrichards:
It will probably be fine to merge this anyway as the overlap is small and I will just rebase on top.

ericvaandering:
So they are called exactly the same thing (exact same code changes to SiteLocalConfig)? Otherwise I would rather have @alexanderrichards break out just a tiny bit of his massive patch and give that to the CRAB3 guys so they can start using it now.

alexanderrichards:
the patch here makes a two line change to SiteLocalConfig plus multi line change to the associated test file. My patch doesn't touch the test file but make the same two lines change to SiteLocalConfig plus a bit more.

ericvaandering:
Ok, then it's OK (I did not expect you to have the test).

@ticoann please merge

emaszs:
I'll modify the exception to be more specific as per Eric's comment, give me a bit of time.

emaszs:
Done, exception caught is now SiteConfigError instead of plain Exception.

test/python/WMCore_t/Storage_t/SiteLocalConfig_t.py
```diff
@@ -156,6 +159,37 @@ def testLoadingConfigFromOverridenEnvVarriable(self):
...
+            if d[0] == "T":
+                os.environ['WMAGENT_SITE_CONFIG_OVERRIDE'] ='/cvmfs/cms.cern.ch/SITECONF/%s/JobConfig/site-local-config.xml' % (d)
+                try:
+                    slc = loadSiteLocalConfig()
+                except SiteConfigError as e:
+                    print e.args[0]
+                phedexNode = slc.localStageOut.get("phedex-node")
+                # If slc is correct, perform check
+                if "se-name" in slc.localStageOut and slc.localStageOut["se-name"] in node_map and phedexNode != None:
+                    assert phedexNode == node_map[slc.localStageOut["se-name"]], \
```
ericvaandering: Also please use the assert classes from unittest here, not the bare assert from python. self.assertEqual is probably what you want.

emaszs: Indeed, I've modified that as well.

PR #6221, opened 2015-09-10
Changes to support generator data as json stream

vkuznet:


vkuznet:
Please do not merge this PR yet, since we're experimenting with Yuyi to reduce memory footprint of DBS code. There are other changes which may be required.

src/python/WMCore/WebTools/RESTFormatter.py
```diff
@@ -26,7 +28,21 @@ def __init__(self, config):
 
         TemplatedPage.__init__(self, config)
 
+    def genstreamer(self, data):
+        yield "["
+        firstItem  = True
+        for item in data:
+            if not firstItem:
+                yield ","
```
ericvaandering: Is this going to be generally safe? Some JS stuff has a problem with trailing commas as well.

ericvaandering: Ah, wait, you are also guaranteed to yield the actual value, right? So no problem

vkuznet: Yes, this will basically create string representation of list of dicts, e.g.
[{dict1},{dict2}]

This is what DBS returns and in fact this is what
src/python/WMCore/Database/DBFormatter.py is doing in formatDict which I intend
to convert into generator as well. But first we need to test how RESTModel
class will handle generator response from db layer and unwind it into
string representation of list of dicts for a client.

vkuznet:
@ticoann, Seangchan, I think we can merge this, since changes I made will only reflect DBS code which may switch to generator. I provided unit tests anyway and fixed some issues with it suggested by pylint. Yuyi claimed that usage of generator already improve single DBS API by 30% memory wise and in order to DBS use generators we'll need this change in WMCore. Later I may refactor WMCore again to move this part directly to the exposed layer of public API.

ericvaandering:
retest this please

PR #6223, opened 2015-09-11
Fix cmssw

ericvaandering:
Fix up the scram emulator to let a few unit tests succeed.


ericvaandering:
@ticoann please merge this

PR #6227, opened 2015-09-11
Convert DBInterface to use generators while dealing with database calls; this eliminates ResultSet class usage (even though I left it); remove thread from ResultSet class anyway

vkuznet:
Please keep this PR open until comprehensive set of integration tests is done. By default I left current behavior, i.e. return list from processData. But clients can start using generators via explicit option useGenerator=True in processData.


ticoann:
retest this please

ericvaandering:
retest this please

vkuznet:
Please discard this PR in favor of #6284

Issue #6241, opened 2015-09-15
Inconsistent usage of max_list_len


amaltaro:
@yuyiguo @ticoann this is the issue I mentioned yesterday, which is hitting us in production right now.

It seems we have an inconsistent usage of this max_list_len (for lumi_list for instance) between dbs client and server.

In the dbsClient code, we check the length of the list regardless whether it's a plain or a nested list (cannot be bigger than 1000, otherwise we slice it):
https://github.com/dmwm/DBS/blob/d58e4fe7be87a6f84da4f246523641c9988ba1ef/Client/src/python/dbs/apis/dbsClient.py#L1009

on the server side, we first expand this lumi_list in a plain list and then AND only then we check it against
the max length
https://github.com/dmwm/DBS/blob/00f44b9b32cefd79888eb1ded0ab8e359ab47965/Server/Python/src/dbs/web/DBSReaderModel.py#L910

where the following nested list
[[248, 494], [534, 546], [638, 650], [794, 1313], [1327, 1339], [1353, 1482], [1496, 1573], [1652, 1664], [1743, 1755], [1860, 1872]]

has 1053 elements, thus DBSServer returns a Bad Request.

Before putting more effort on it, I'd like to know whether we should fix it on the DBS side
 a. checking their length before any expansion
 b. if passes previous test but it's bigger, either go ahead or slice on the client side
or fix it in WMCore
 a. always expand from range of lumis to a plain lumi list and make POST calls with this plain list.

ticoann:
Hi Alan, it seem this should fixed in DBS side, it sounds like dbs client call slices the parameters assuming which should accept by server but DBS server rejects it due to Oracle constraint. Is that correct?
Since it is hitting production now, so we should have temporary patch in WMAgent side but I think it should be eventually fixed in DBS side. Otherwise, it will have the problem whatever application which uses the same client api. @yuyiguo, Yuyi, what do you think?

amaltaro:
Almost :) It does not really reach oracle. In the client side we have a list of lists where its length is smaller than 1000, so no slicing is needed here. However, the first thing done on the server side is expanding this list of lists in a plain list, which goes beyond the 1000 limit. Since the server does not apply any slices, it simply raises an exception in this case:
https://github.com/dmwm/DBS/blob/00f44b9b32cefd79888eb1ded0ab8e359ab47965/Server/Python/src/dbs/web/DBSReaderModel.py#L929

[...]

PR #6242, opened 2015-09-15
Accept plain list of lumis

amaltaro:
On the same topic as #6241 , I just noticed we can not workaround that issue by providing a plain list of lumis since it's broken in production :-(

I also noticed we are not properly setting splitting.lumis (a few lines below) when it's a list of lists, since we don't expand it. Since things work fine since the beginning, I assume it's not a problem, otherwise let me know and I can update it.

Request injection works on my VM, however I still want to completely run one workflow.

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -1322,8 +1322,11 @@ class WMTaskHelper(TreeHelper):
         for run, runLumis in lumiMask.items():
             runs.append(int(run))
             lumiList = []
-            for lumi in runLumis:
-                lumiList.extend([str(l) for l in lumi])
+            # check whether it's a plain or nested list
+            if isinstance(runLumis[0], int):
+                lumiList.extend([str(l) for l in runLumis])
+            else:
+                lumiList.extend([str(l) for lumi in runLumis for l in lumi])
             lumis.append(','.join(lumiList))
 
         self.data.input.splitting.runs = runs
--- a/test/data/ReqMgr/requests/Integration/MonteCarloFromGEN_LumiMask.json
+++ b/test/data/ReqMgr/requests/Integration/MonteCarloFromGEN_LumiMask.json
@@ -1,7 +1,7 @@
 { 
 "createRequest":
 {         
-    "Comments": "MonteCarloFromGEN with LumiMask, 1 run and 42l in total. Half an hour opened",
+    "Comments": "MonteCarloFromGEN with LumiMask, 1 run and 15l in total. Half an hour opened",
     "Campaign": "Campaign-OVERRIDE-ME",
     "RequestString": "RequestString-OVERRIDE-ME",
     "CMSSWVersion": "CMSSW_5_3_19",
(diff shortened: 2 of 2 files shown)


ericvaandering:
I'm a little confused about where this is going. But I want to point out that we have DataStructs/LumiList in WMCore now which handles lumi lists in different formats and converts between them. And is able to do math operations on them. So just keep in mind to not re-invent the wheel every time we have to deal with these lists (since they are easily prone to off-by-one errors).

amaltaro:
LumiList.py does not provide what we need in this piece of code. The problem I'm fixing here is that setLumiMask() can receive (from requestor) either a compactList or a runsAndLumis dict, where in the latter it would just fail since int object is not iterable.

Nevertheless, I'm creating a new method in LumiList that I'm sure would be used somewhere in the code

amaltaro:
Mhh, I wanted to improve LumiList, but actually there are several bits to change. I'll create another issue for that.
This PR should fix the problem where requestors can provide either a range of lumis or a list of individual lumis.

amaltaro:
@ericvaandering , I believe something is wrong with jenkins, no? It's the third test done for this PR without any results yet.

ericvaandering:
There were two failures of just one slice (different each time). Basically could not install the software, probably a hiccup with comp.pre.

It happens occasionally.

amaltaro:
@ericvaandering while updating this PR, I noticed a plain list may break other stuff (buildLumiMask for instance).
My question is simple, can we support only range of lumis for WMAgent? I'm not even sure whether the initial idea was to support both or only one.

ericvaandering:
Originally, as I recall it, they did not really want to support lumi ranges at all. When it was supported, it was often supplied as a two lists, one of runs (not unique) and one (twice as long) with the start/end lumis.

So I think, in general, doing this the way the rest of CMS does it is a good idea. When you say "can we support only range of lumis for WMAgent” at what level are you talking about? Requests or deeper in the system. Because deeper in the system, things may have been translated into one of these other forms.

Eric


On Sep 16, 2015, at 5:37 AM, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


@ericvaandering<https://github.com/ericvaandering> while updating this PR, I noticed a plain list may break other stuff (buildLumiMask for instance).
My question is simple, can we support only range of lumis for WMAgent? I'm not even sure whether the initial idea was to support both or only one.

amaltaro:
At request creation level only, this way we avoid over-complicating something that may never be used in real life.

In the end, providing range of lumis seems the most reasonable IMO.
If we can, then I just close this PR and there is nothing to fix here.

ericvaandering:
Yes. We can require the compact format at request generation.

On Sep 16, 2015, at 07:55, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


At request creation level only, this way we avoid over-complicating something that may never be used in real life.

In the end, providing range of lumis seems the most reasonable IMO.
If we can, then I just close this PR and there is nothing to fix here.

amaltaro:
Then I'm closing this PR since there is nothing to merge here. Thanks and sorry for the noise!

PR #6244, opened 2015-09-15
Fix the case where a lumi in the list is completely inside the previous

ericvaandering:
Chris pointed out an edge case


ericvaandering:
@ticoann please merge

PR #6245, opened 2015-09-15
Remove exception in tearDown, get further with pileup fetcher

ericvaandering:
This fixes an exception in the teardown method which may allow other tests to run. It changes where an exception gets thrown in the test it intends to fix, now at the point where it tries to find a logger for the current thread.


ericvaandering:
@ticoann please merge

PR #6246, opened 2015-09-15
Fix service tests

ericvaandering:
Fixes four Service_t tests which fail because they reference an old CVS server at CERN


ericvaandering:
@ticoann please merge

PR #6250, opened 2015-09-16
fix key name in reqmgr2 client

amaltaro:
also applied some pep8 fixes


amaltaro:
@ticoann good to be merged

PR #6254, opened 2015-09-17
Making DBS3Upload logs more friendly

amaltaro:
Since this task completed 101 days in my todo list, I thought I should finally make these changes :-)
Moving that log to debug, since that line can become as bigger as >100 lines length and surfing through the logs is a PITA.

Also took advantage and cleaned some pylint complaints.


amaltaro:
@ticoann good to be merged.

PR #5588, opened 2015-01-23
Create indexes to wmbs_job and fix column lengths

amaltaro:
Fixes a couple of issues:
 1) column length for unique constraint cannot be > 767. Fix for #5465 
 2) partial fix for #5213 by adding 4 indexes to 4 fk in bl_runjob table (applied Yuyi's fix). 
The deadlock still happens, but _should_ be less often now.

I've just deployed a MariaDB backend agent and the deployment went fine and the 4 indexes are there [1]
Since my knowledge in DB is quite weak :-(, I may be missing something.
 @yuyiguo @hufnagel can you please review?

```
[1]
MariaDB [wmagent]> show index from bl_runjob;
+-----------+------------+------------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table     | Non_unique | Key_name               | Seq_in_index | Column_name  | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+-----------+------------+------------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| bl_runjob |          0 | PRIMARY                |            1 | id           | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| bl_runjob |          0 | retry_count            |            1 | retry_count  | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| bl_runjob |          0 | retry_count            |            2 | wmbs_id      | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| bl_runjob |          1 | idx_bl_runjob_wmbs     |            1 | wmbs_id      | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| bl_runjob |          1 | idx_bl_runjob_location |            1 | location     | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| bl_runjob |          1 | idx_bl_runjob_status   |            1 | sched_status | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
| bl_runjob |          1 | idx_bl_runjob_users    |            1 | user_id      | A         |           0 |     NULL | NULL   | YES  | BTREE      |         |               |
+-----------+------------+------------------------+--------------+--------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
```

PR #6260, opened 2015-09-18
Clean-up database usage in unittests, major upgrade do not merge

vkuznet:
Major upgrade: remove reallyDeleteMyDatabaseAfterEveryTest bot option and replace it with setUp/tearDown approach to setup/delete unittest databases for every unit test; re-factor DBCore class to use generators and be compatible with WMCore functionality; re-factor ResultSet to keep data instead of resultproxy objects; close all resultproxy objects once data is read from them; change logic of processData to use transaction (it as reversed); adjust all unit tests to create its own database for their test purposes

### DIFF
--- a/src/python/WMCore/Database/DBFormatter.py
+++ b/src/python/WMCore/Database/DBFormatter.py
@@ -38,6 +38,18 @@ class DBFormatter(WMObject):
         t = datetime.datetime.now()
         return self.convertdatetime(t)
 
+# VK this is example of format method re-factoring if
+# we'll use ResultSetExampleDictConverter class
+#    def format(self, result):
+#        """
+#        Some standard formatting, put all records into a list
+#        """
+#        out = []
+#        for res in result:
+#            for row in res.fetchall():
+#                out.append(row)
+#            res.close()
+#        return out
     def format(self, result):
         """
         Some standard formatting, put all records into a list
@@ -61,6 +73,15 @@ class DBFormatter(WMObject):
             for i in r.fetchone():
                 out.append(i)
         return out
+
+# VK this is example of formatDict method re-factoring if
+# we'll use ResultSetExampleDictConverter class
+#    def formatDict(self, result):
+#        out = []
+#        for row in result:
+#            for item in row.fetchall():
+#                out.append(item)
+#        return out
 
     def formatDict(self, result):
         """
@@ -86,6 +107,16 @@ class DBFormatter(WMObject):
 
         return dictOut
 
+# VK this is example of formatOneDict method re-factoring if
+# we'll use ResultSetExampleDictConverter class
+#    def formatOneDict(self, result):
+#        """
+#        Return a dictionary representing the first record
+#        """
+#        if (len(result) == 0):
+#            return {}
+#        return result[0]
+
     def formatOneDict(self, result):
         """
         Return a dictionary representing the first record
@@ -109,7 +140,7 @@ class DBFormatter(WMObject):
         Use fetchmany(size = default arraysize = 50)
 
         """
-	if type(cursor.keys) == types.MethodType:
+        if type(cursor.keys) == types.MethodType:
             keys = [x.lower() for x in cursor.keys()]
         else:
             keys = [x.lower() for x in cursor.keys]
--- a/src/python/WMCore/Services/DBS/DBSReader.py
+++ b/src/python/WMCore/Services/DBS/DBSReader.py
@@ -1,15 +1,13 @@
 #!/usr/bin/env python
 """
 _DBSReader_
-
 Readonly DBS Interface
-
 """
 import urlparse
 
 from WMCore.Services.DBS.DBSErrors import DBSReaderError
 from WMCore.Services.EmulatorSwitch import emulatorHook
-from WMCore.Services.DBS.DBS3Reader import DBS3Reader
+
 
 @emulatorHook
 def DBSReader(endpoint, **kwargs):
@@ -17,13 +15,17 @@ def DBSReader(endpoint, **kwargs):
     
     msg = ''
     try:
-        dbs = DBS3Reader(endpoint, **kwargs)
+        dbs = _getDBS3Reader(endpoint, **kwargs)
         # if this doesn't throw endpoint is dbs3
         dbs.dbs.serverinfo()
         return dbs
     except Exception as ex:
         msg += 'Instantiating DBS3Reader failed with %s\n' % str(ex)
-    raise DBSReaderError("Can't contact DBS at %s, kwargs %s, got errors %s" \
-            % (endpoint, kwargs, msg))
+    raise DBSReaderError("Can't contact DBS at %s, got errors %s" % (endpoint, msg))
+
+
+def _getDBS3Reader(endpoint, **kwargs):
+    from WMCore.Services.DBS.DBS3Reader import DBS3Reader as DBSReader
+    return DBSReader(endpoint, **kwargs)
 
 __all__ = [DBSReader]
--- a/src/python/WMCore/Services/DBS/DBSReader.py
+++ b/src/python/WMCore/Services/DBS/DBSReader.py
@@ -9,7 +9,7 @@ import urlparse
 
 from WMCore.Services.DBS.DBSErrors import DBSReaderError
 from WMCore.Services.EmulatorSwitch import emulatorHook
-
+from WMCore.Services.DBS.DBS3Reader import DBS3Reader
 
 @emulatorHook
 def DBSReader(endpoint, **kwargs):
@@ -17,17 +17,13 @@ def DBSReader(endpoint, **kwargs):
     
     msg = ''
     try:
-        dbs = _getDBS3Reader(endpoint, **kwargs)
+        dbs = DBS3Reader(endpoint, **kwargs)
         # if this doesn't throw endpoint is dbs3
         dbs.dbs.serverinfo()
         return dbs
     except Exception as ex:
         msg += 'Instantiating DBS3Reader failed with %s\n' % str(ex)
-    raise DBSReaderError("Can't contact DBS at %s, got errors %s" % (endpoint, msg))
-
-
-def _getDBS3Reader(endpoint, **kwargs):
-    from WMCore.Services.DBS.DBS3Reader import DBS3Reader as DBSReader
-    return DBSReader(endpoint, **kwargs)
+    raise DBSReaderError("Can't contact DBS at %s, kwargs %s, got errors %s" \
+            % (endpoint, kwargs, msg))
 
 __all__ = [DBSReader]
--- a/src/python/WMCore/WMBS/CreateWMBSBase.py
+++ b/src/python/WMCore/WMBS/CreateWMBSBase.py
@@ -66,7 +66,7 @@ class CreateWMBSBase(DBCreator):
         self.create["01wmbs_fileset"] = \
           """CREATE TABLE wmbs_fileset (
              id          INTEGER      PRIMARY KEY AUTO_INCREMENT,
-             name        VARCHAR(700) NOT NULL,
+             name        VARCHAR(255) NOT NULL,
              open        INT(1)       NOT NULL DEFAULT 0,
              last_update INTEGER      NOT NULL,
              UNIQUE (name))"""
@@ -74,7 +74,7 @@ class CreateWMBSBase(DBCreator):
         self.create["02wmbs_file_details"] = \
           """CREATE TABLE wmbs_file_details (
              id           INTEGER      PRIMARY KEY AUTO_INCREMENT,
-             lfn          VARCHAR(700) NOT NULL,
+             lfn          VARCHAR(255) NOT NULL,
              filesize     BIGINT,
              events       INTEGER,
              first_event  BIGINT       NOT NULL DEFAULT 0,
@@ -148,9 +148,9 @@ class CreateWMBSBase(DBCreator):
         self.create["07wmbs_workflow"] = \
           """CREATE TABLE wmbs_workflow (
              id           INTEGER          PRIMARY KEY AUTO_INCREMENT,
-             spec         VARCHAR(700)     NOT NULL,
-             name         VARCHAR(700)     NOT NULL,
-             task         VARCHAR(700)     NOT NULL,
+             spec         VARCHAR(70)     NOT NULL,
+             name         VARCHAR(70)     NOT NULL,
+             task         VARCHAR(70)     NOT NULL,
              type         VARCHAR(255),
              owner        INTEGER          NOT NULL,
              alt_fs_close INT(1)           NOT NULL,
@@ -278,15 +278,15 @@ class CreateWMBSBase(DBCreator):
           """CREATE TABLE wmbs_job (
              id           INTEGER       PRIMARY KEY AUTO_INCREMENT,
              jobgroup     INTEGER       NOT NULL,
-             name         VARCHAR(255),
+             name         VARCHAR(70),
              state        INTEGER       NOT NULL,
              state_time   INTEGER       NOT NULL,
              retry_count  INTEGER       DEFAULT 0,
              couch_record VARCHAR(255),
              location     INTEGER,
              outcome      INTEGER       DEFAULT 0,
-             cache_dir    VARCHAR(767)  DEFAULT 'None',
-             fwjr_path    VARCHAR(767),
+             cache_dir    VARCHAR(70)  DEFAULT 'None',
+             fwjr_path    VARCHAR(70),
              FOREIGN KEY (jobgroup)
              REFERENCES wmbs_jobgroup(id) ON DELETE CASCADE,
              FOREIGN KEY (state) REFERENCES wmbs_job_state(id),
--- a/test/python/WMCore_t/WMBS_t/JobSplitting_t/WMBSMergeBySize_t.py
+++ b/test/python/WMCore_t/WMBS_t/JobSplitting_t/WMBSMergeBySize_t.py
@@ -30,7 +30,8 @@ class WMBSMergeBySize(unittest.TestCase):
         """
         self.testInit = TestInit(__file__)
         self.testInit.setLogging()
-        self.testInit.setDatabaseConnection()
+        self.testDB = 'unittest_%s' % self.__class__.__name__
+        self.testInit.prepareDatabase(self.testDB)
         self.testInit.setSchema(customModules = ["WMCore.WMBS"],
                                 useDefault = False)
 
@@ -38,7 +39,6 @@ class WMBSMergeBySize(unittest.TestCase):
         self.daoFactory = DAOFactory(package = "WMCore.WMBS",
                                      logger = myThread.logger,
                                      dbinterface = myThread.dbi)
-        return
 
     def tearDown(self):
         """
@@ -46,8 +46,7 @@ class WMBSMergeBySize(unittest.TestCase):
 
         Clear out WMBS.
         """
-        self.testInit.clearDatabase()
-        return
+        self.testInit.destroyDatabase(self.testDB)
 
     def stuffWMBS(self, injected = True):
         """
--- a/setup_test.py
+++ b/setup_test.py
@@ -139,7 +139,9 @@ if can_nose:
         # if you set this to true, I will really delete your database
         #  after every test
         # WARNING WARNING WARNING
-        user_options = [\
+        user_options = [('reallyDeleteMyDatabaseAfterEveryTest=',
+                         None,
+                         'If you set this I WILL DELETE YOUR DATABASE AFTER EVERY TEST. DO NOT RUN ON A PRODUCTION SYSTEM'),
                          ('buildBotMode=',
                           None,
                           'Are we running inside buildbot?'),
--- a/setup_test.py
+++ b/setup_test.py
@@ -139,9 +139,7 @@ if can_nose:
         # if you set this to true, I will really delete your database
         #  after every test
         # WARNING WARNING WARNING
-        user_options = [('reallyDeleteMyDatabaseAfterEveryTest=',
-                         None,
-                         'If you set this I WILL DELETE YOUR DATABASE AFTER EVERY TEST. DO NOT RUN ON A PRODUCTION SYSTEM'),
+        user_options = [\
                          ('buildBotMode=',
                           None,
                           'Are we running inside buildbot?'),
@@ -258,13 +256,6 @@ if can_nose:
             else:
                 quickTestArg = []
                 
-            if self.reallyDeleteMyDatabaseAfterEveryTest:
-                print "#### WE ARE DELETING YOUR DATABASE. 3 SECONDS TO CANCEL ####"
-                print "#### buildbotmode is %s" % self.buildBotMode
-                sys.stdout.flush()
-                import WMQuality.TestInit
-                WMQuality.TestInit.deleteDatabaseAfterEveryTest( "I'm Serious" )
-                time.sleep(4)
             if self.workerNodeTestsOnly:
                 args = [__file__,'--with-xunit', '-m', '(_t.py$)|(_t$)|(^test)','-a','workerNodeTest',self.testingRoot]
                 args.extend( quickTestArg )
--- a/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
@@ -25,12 +25,12 @@ class Create(DBCreator):
               """CREATE TABLE dbsbuffer_dataset
                         (
                            id              BIGINT UNSIGNED PRIMARY KEY AUTO_INCREMENT,
-                           path            VARCHAR(500) COLLATE latin1_general_cs UNIQUE NOT NULL,
+                           path            VARCHAR(255) COLLATE latin1_general_cs UNIQUE NOT NULL,
                            processing_ver  VARCHAR(255),
                            acquisition_era VARCHAR(255),
                            valid_status    VARCHAR(20),
                            global_tag      VARCHAR(255),
-                           parent          VARCHAR(500),
+                           parent          VARCHAR(255),
                            prep_id         VARCHAR(255)
                         ) ENGINE=InnoDB"""
 
@@ -58,7 +58,7 @@ class Create(DBCreator):
                            app_name varchar(100),
                            app_ver  varchar(100),
                            app_fam  varchar(100),
-                           pset_hash varchar(700),
+                           pset_hash varchar(255),
                            config_content LONGTEXT,
                            in_dbs int, 
                            UNIQUE (app_name, app_ver, app_fam, pset_hash)
@@ -80,8 +80,8 @@ class Create(DBCreator):
         self.create["03dbsbuffer_workflow"] = \
           """CREATE TABLE dbsbuffer_workflow (
                id                           INTEGER PRIMARY KEY AUTO_INCREMENT,
-               name                         VARCHAR(700),
-               task                         VARCHAR(700),
+               name                         VARCHAR(255),
+               task                         VARCHAR(255),
                block_close_max_wait_time    INTEGER UNSIGNED,
                block_close_max_files        INTEGER UNSIGNED,
                block_close_max_events       INTEGER UNSIGNED,
@@ -93,7 +93,7 @@ class Create(DBCreator):
         self.create["04dbsbuffer_file"] = \
           """CREATE TABLE dbsbuffer_file (
              id           INTEGER      PRIMARY KEY AUTO_INCREMENT,
-             lfn          VARCHAR(500) NOT NULL,
+             lfn          VARCHAR(255) NOT NULL,
              filesize     BIGINT,
              events       INTEGER,
              dataset_algo BIGINT UNSIGNED   NOT NULL,
--- a/src/python/WMComponent/DBS3Buffer/MySQL/InsertWorkflow.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/InsertWorkflow.py
@@ -31,7 +31,7 @@ class InsertWorkflow(DBFormatter):
     def execute(self, requestName, taskPath,
                 blockMaxCloseTime, blockMaxFiles,
                 blockMaxEvents, blockMaxSize,
-                conn = None, transaction = False):
+                conn = None, transaction = True):
         """
         _execute_
 
@@ -43,7 +43,7 @@ class InsertWorkflow(DBFormatter):
                  'blockMaxEvents' : blockMaxEvents,
                  'blockMaxSize' : blockMaxSize}
 
-        self.dbi.processData(self.sql, binds, conn = conn,
+        result = self.dbi.processData(self.sql, binds, conn = conn,
                              transaction = transaction)
 
         binds = {'name': requestName, 'task': taskPath}
--- a/src/python/WMCore/Database/DBCore.py
+++ b/src/python/WMCore/Database/DBCore.py
@@ -7,10 +7,12 @@ Core Database APIs
 
 """
 
+import sqlalchemy
 
 
 from WMCore.DataStructs.WMObject import WMObject
 from WMCore.Database.ResultSet import ResultSet
+from WMCore.WMException import WMException
 from copy import copy
 import WMCore.WMLogging
 
@@ -28,9 +30,6 @@ class DBInterface(WMObject):
         Test the hell out of it
         Support executemany()
     """
-
-    logger = None
-    engine = None
 
     def __init__(self, logger, engine):
         self.logger = logger
@@ -63,14 +62,9 @@ class DBInterface(WMObject):
             resultProxy = connection.execute(s)
         else:
             resultProxy = connection.execute(s, b)
-
-        if returnCursor:
-            return resultProxy
-
-        result = ResultSet()
-        result.add(resultProxy)
-        resultProxy.close()
-        return result
+        yield resultProxy
+        if not returnCursor:
+            resultProxy.close()
 
     def executemanybinds(self, s=None, b=None, connection=None,
                          returnCursor=False):
@@ -92,27 +86,18 @@ class DBInterface(WMObject):
 
         s = s.strip()
         if s.lower().endswith('select', 0, 6):
-            """
-            Trying to select many
-            """
+            # Trying to select many
             if returnCursor:
-                result = []
                 for bind in b:
-                    result.append(connection.execute(s, bind))
+                    result = connection.execute(s, bind)
+                    yield result
             else:
-                result = ResultSet()
                 for bind in b:
                     resultproxy = connection.execute(s, bind)
-                    result.add(resultproxy)
+                    yield resultproxy
                     resultproxy.close()
-
-            return self.makelist(result)
-
-        """
-        Now inserting or updating many
-        """
         result = connection.execute(s, b)
-        return self.makelist(result)
+        yield result
 
     def connection(self):
         """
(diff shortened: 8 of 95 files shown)


PerilousApricot:
FWIW, the 'really tear down database'  option is there because otherwise it
literally will blow away the entire database. I didn't want someone to
accidentally run the command and have it wipe whatever databases were
around.

ericvaandering:
Is there no way we can do this more gradually? Also, why are you changing the DBS3Reader code?

src/python/WMCore/WMBS/CreateWMBSBase.py
```diff
@@ -278,15 +278,15 @@ def __init__(self, logger = None, dbi = None, params = None):
...
              state        INTEGER       NOT NULL,
              state_time   INTEGER       NOT NULL,
              retry_count  INTEGER       DEFAULT 0,
              couch_record VARCHAR(255),
              location     INTEGER,
              outcome      INTEGER       DEFAULT 0,
-             cache_dir    VARCHAR(767)  DEFAULT 'None',
-             fwjr_path    VARCHAR(767),
+             cache_dir    VARCHAR(70)  DEFAULT 'None',
+             fwjr_path    VARCHAR(70),
```
amaltaro: I believe you mistakenly changed the length of these columns... it's a big NO for this change, we store the whole path here and even 767 chars is not enough for TaskChain.
The same comment probably applies for other columns in other tables that you changed.

ericvaandering:
Jenkins is going to flail on this:

 error: option --reallyDeleteMyDatabaseAfterEveryTest not recognized

If you want to see a test, commit something that allows this option but doesn't do anything with it.

vkuznet:
Good point, I put it back as an option.

src/python/WMCore/WMBS/CreateWMBSBase.py
vkuznet: Alan,
maria DB does not support fields more than 255, and that was the reason for
change. Otherwise I can't proceed with changes.

For unit tests there is no need to insert gigantic names. If we can't be
flexible we'll not move forward with anything.

V.

vkuznet:
The short answer is no, I changed internals of WMCore which affect
almost entire universe. But changes are isolated to WMCore and affect unit
tests.

The changes for DBS3 are minor to comply with overall
change, i.e. specify transactions, use prepareDatabase method in unit tests
and changes for column names in tables which were restricted by maria db.

src/python/WMCore/WMBS/CreateWMBSBase.py
hufnagel: MariaDB supports fields longer than 255. You have some problems with combined unique constraints, that's all. These columns CANNOT be that small, otherwise everything we do will break (we store very long strings there all the time).

vkuznet:
Please don't post comments without context. First, read the changes.  The hack option
"reallyDeleteMyDatabaseAfterEveryTest" is replaced with correct setUp/tearDown
procedure. Every unit test creates its own DB for itself and there is not need
anymore for reallyDeleteMyDatabaseAfterEveryTest option and, this option didn't
used in production either.

src/python/WMCore/WMBS/CreateWMBSBase.py
vkuznet: Dirk, you're correct, the problem with unique constraints not the fields by
attributes. I was not clear on this. Anyway, I had enough headache to
change the code that I would worry right away about mariadb tweaking.

If someone knows how to tweak maria DB setup please speak up. Otherwise I'll
come back to this issue later when time permit. So far my goal was to
get things running with WMCore changes.

PerilousApricot:
Hi @vkuznet - I didn't just post off the hip, the option was there for a reason and I checked to look at the code. First the history of why it was there:

If the setUp/tearDown works right, then you always get your database back to a clean state. If not, the database is left inconsistent, and next time you run the test, the setUp blows up.

If you're a developer working on your own machine, you can sort it out and do the right thing with the database, but what ended up happening the first time around with jenkins is a test case would be broken, not tear down properly and then break every single test case after. Worse, it would also break every single test that came after.

The option to delete the database doesn't simply call the Delete DAOs after each test like you've done here:

https://github.com/vkuznet/WMCore/blob/a1c508fbb43a089e86c1d01e4d1129a80fe106d7/src/python/WMCore/WMInit.py#L211

via

https://github.com/vkuznet/WMCore/blob/a1c508fbb43a089e86c1d01e4d1129a80fe106d7/src/python/WMQuality/TestInit.py#L269

It actually does a DROP DATABASE; CREATE DATABASE to ensure the database is truly empty. Otherwise Eric or whomever will have to periodically clean out the mysql databases to deal with Jenkins blowing up.

src/python/WMCore/WMBS/CreateWMBSBase.py
hufnagel: Just pointing out you can't commit any of these short column length because it'll break everything. If this is just an intermediate stage to get some tests to run, fine. But we can't forget about changing this back to longer columns.

PerilousApricot:
Sorry, I needed to go one level deeper. You do blow away the whole database, but now it's the default, so if someone who happens to have their WMAGENT_CONFIG set to production stuff runs test code, it deletes everything for them without warning...

ericvaandering:
"If the setUp/tearDown works right, then you always get your database back to a clean state. If not, the database is left inconsistent, and next time you run the test, the setUp blows up."

Python 2.7 includes (as @alexanderrichards reminded us) an .addCleanup method in setup that will run even if tearDown does not, so that can be used to solve that particular problem later this year.

ericvaandering:
> On Sep 18, 2015, at 15:50, Andrew Melo <notifications@github.com> wrote:
> 
> Sorry, I needed to go one level deeper. You do blow away the whole database, but now it's the default, so if someone who happens to have their WMAGENT_CONFIG set to production stuff runs test code, it deletes everything for them without warning…
> 
I presumed this is what you were worried about in the first place. I’m afraid I don’t really know what the solution is to have guaranteed clean unit tests but also not blow away everything is some well meaning person runs a unit test on a production server.

PerilousApricot:
@ericvaandering It's not just a matter of if the tearDown doesn't fire, which is obnoxious. People also had an annoying habit of changing their Create DAOs and not their Delete DAOs.

In either case, I think that having the default behavior to be to completely drop and recreate the database is a BAD idea.

PerilousApricot:
@ericvaandering The solution is to require people to opt in with a very clear command line argument. You CAN delete the whole database, but you have to certify that you're a consenting adult.

Newer versions of binutils won't let you ```rm -rf /```, you have to pass an additional argument to say "yes, I'm really sure what I want to do is recursively delete everything. You don't lose the functionality, but it keeps most of the trivial accidents from ruining your day.

src/python/WMCore/Services/DBS/DBSReader.py
```diff
@@ -17,17 +17,13 @@ def DBSReader(endpoint, **kwargs):
     
     msg = ''
     try:
-        dbs = _getDBS3Reader(endpoint, **kwargs)
+        dbs = DBS3Reader(endpoint, **kwargs)
```
ericvaandering: This is the set of changes that I am wondering why you need to make it. Granted, I want to get rid of all this with what @jha2 is doing, but I don't see what this has to do with MariaDB.

ericvaandering:
I would like to get away from having to run tests with setup_test.py (or at least having to run them that way to get the same results that Jenkins gets). Requiring `export REALLY_DROP_DATABASES=I_AGREE` in the environment may be safe enough.

ericvaandering:
Or even writing a plugin to add a command line option to nose itself: http://nose.readthedocs.org/en/latest/plugins/writing.html

vkuznet:
Andrew,

> If the setUp/tearDown works right, then you always get your database back to a clean state. If not, the database is left inconsistent, and next time you run the test, the setUp blows up.

the prepareDataset which is invoked in setUp does five steps. If dataset left
in unclean case it will be dropped first, then new dataset will be created,
schema will be populated and connection will be reset to fresh new dataset for
test to run. All this steps are done at every setUp.

Moreover, every tests runs with its own databases, i.e. one unit tests can't run with
database from another test. The current design is opposite. All unit tests
are run within the same dataset and if something left over there we have
failures. This is why the current design is not unittests per-se it is
integration tests.

> 
> If you're a developer working on your own machine, you can sort it out and do the right thing with the database, but what ended up happening the first time around with jenkins is a test case would be broken, not tear down properly and then break every single test case after. Worse, it would also break every single test that came after.

The purpose of unit tests is to have clean DB populate with its own stuff and
run tests with it. If test fails and tearDown is not invoked, e.g. due to
exception you have:
- a unique dataset to look for error
- no other tests contribute to this database which is nice to isolate the
  problem

> 
> The option to delete the database doesn't simply call the Delete DAOs after each test like you've done here:
> 
> https://github.com/vkuznet/WMCore/blob/a1c508fbb43a089e86c1d01e4d1129a80fe106d7/src/python/WMCore/WMInit.py#L211
> 
> via
> 
> https://github.com/vkuznet/WMCore/blob/a1c508fbb43a089e86c1d01e4d1129a80fe106d7/src/python/WMQuality/TestInit.py#L269
> 
> It actually does a DROP DATABASE; CREATE DATABASE to ensure the database is truly empty. Otherwise Eric or whomever will have to periodically clean out the mysql databases to deal with Jenkins blowing up.

I made only separation of methods and nothing else.

The current behavior is in this class
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/MySQL/Destroy.py

which does all steps at once, drop database, create database and use it.

What I did is separate the logic into individual classes. One class destroy
database and another create it.

src/python/WMCore/WMBS/CreateWMBSBase.py
vkuznet: Dirk,
we discussed on a meeting that we WILL NOT do the actual merge until we resolve
all issues and we agreed to do gradual adjustments.
The PR title reflect that:
Clean-up database usage in unittests, major upgrade do not merge

It does say *do not merge*.

vkuznet:
I want to understand the logic here.

I repeat again, the commit made ALL unittests to create their own unique
databases. How they can affect your WMAGENT_CONFIG database?

For example DBCore_t.py test creates unitttest_DBCore, the JobAccounting_t.py
creates unittest_JobAccounting database, etc. Please explain how it affects
your database which you define in WMAGENT_CONFIG?

Again, if you need integration tests, then you can and will use WMAGENT_CONFIG
to point to YOUR database and we can adjust code to write over there.

vkuznet:
Why person is allowed to run anything on production server?

The unit tests are not supposed to be run on production server. For that people
must use pre-production server.

Again, all unit tests creates their own DB and CAN NOT affect any production
database by definition now. Isn't it a good thing?

vkuznet:
On  0, Andrew Melo <notifications@github.com> wrote:
> @ericvaandering It's not just a matter of if the tearDown doesn't fire, which is obnoxious. People also had an annoying habit of changing their Create DAOs and not their Delete DAOs.

why do you accept code which modifies Create DAOs and does not do proper stuff
to Delete DAOs. I don't see ANY REASON why code should be accepted in a first
place. Do not try to hide a real problem.

> In either case, I think that having the default behavior to be to completely drop and recreate the database is a BAD idea.

I'll not repeat myself here, please read previous comments.

src/python/WMCore/Services/DBS/DBSReader.py
vkuznet: Eric,
the code is redundant by itself, nobody calls _getDBS3Reader except
getDBS3Reader on line 20
https://github.com/dmwm/WMCore/pull/6260/files#diff-274504b8b0136ef92367790e745914b5L20

I just cleared it up because on my previous commits you asked explicitly to
move import/from statements in top of the files. I made it and remove
extra callback. But this change is indeed irrelevant for purpose of database
discussion.

PerilousApricot:
@vkuznet I guess there's just a difference of opinion.

I think there should be a safety on the trigger that blows away the whole database. You think that nobody would ever accidentally run the test harness on an important system, so the safety is unnecessary.

vkuznet:
Eric and Alexander are correct here. Right now I modified setUp to take care
of left overs by dropping DB and re-creating it. Next logical step will
be to use .addCleanup method. But in order to use it the changes I made
to separate databases across all unit tests are required.

vkuznet:
Yes, I stand on position that unittests should not be run on production nodes.
For that purpose we must use pre-production setup.  The code should run in
sandbox environment where nobody can fish.  And, all tests should setup their
own environment and don't fool with each other. We don't need special triggers
if code is correctly designed. Triggers leads to hacks, abandoning
the system and end-less if-else-if-else-if statements in a code (which is what
I see over there).

PerilousApricot:
Like you said, as long as nobody screws up and runs the code in the wrong place, everything is gravy :)

vkuznet:
I said that code now designed to setup and run stuff in sandbox environments.
If you accessing production system and have all rights over there nothing
can stop you to make any mistake and screw the system, triggers is not
a remedy for admin rights.

ericvaandering:
Can you guys stop going back and forth until I can find time to weigh in?

On Sep 18, 2015, at 17:18, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:

I said that code now designed to setup and run stuff in sandbox environments.
If you accessing production system and have all rights over there nothing
can stop you to make any mistake and screw the system, triggers is not
a remedy for admin rights.

src/python/WMCore/Services/DBS/DBSReader.py
ericvaandering: Lets back this one bit out of it, please. My comment on where imports were was not about such a special case.

On Sep 18, 2015, at 16:58, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/python/WMCore/Services/DBS/DBSReader.py<https://github.com/dmwm/WMCore/pull/6260#discussion_r39905400>:

> @@ -17,17 +17,13 @@ def DBSReader(endpoint, **kwargs):
>
>      msg = ''
>      try:
> -        dbs = _getDBS3Reader(endpoint, **kwargs)
> +        dbs = DBS3Reader(endpoint, **kwargs)


Eric, the code is redundant by itself, nobody calls _getDBS3Reader except getDBS3Reader on line 20 https://github.com/dmwm/WMCore/pull/6260/files#diff-274504b8b0136ef92367790e745914b5L20 I just cleared it up because on my previous commits you asked explicitly to move import/from statements in top of the files. I made it and remove extra callback. But this change is indeed irrelevant for purpose of database discussion.
…
On 0, Eric Vaandering <notifications@github.com<mailto:notifications@github.com>> wrote: > @@ -17,17 +17,13 @@ def DBSReader(endpoint, **kwargs): > > msg = '' > try: > - dbs = _getDBS3Reader(endpoint, **kwargs) > + dbs = DBS3Reader(endpoint, **kwargs) This is the set of changes that I am wondering why you need to make it. Granted, I want to get rid of all this with what @jha2<https://github.com/jha2> is doing, but I don't see what this has to do with MariaDB. --- Reply to this email directly or view it on GitHub: https://github.com/dmwm/WMCore/pull/6260/files#r39900934

vkuznet: Done.

ericvaandering:
Ok, I am now free from shuttling our speaker back and forth. I'm going to lay out my understanding and then you can both (all?) tell me where I am wrong.

First, I did not appreciate that Valentin's patch uses a differently named database for every single test. Given that, I cannot see how blowing away that specific database could cause an issue as long as we have a guarantee that the testing code never ever creates, writes to, or drops a database with a different, universal name.

Second, I think we all agree that no one should run tests on "production" servers. However, it has been standard practice to run tests on developer's personal servers. While inadvertently blowing away a database in that case will not cause dataloss to CMS, it could certainly ruin what a developer is trying to do. Furthermore, just because something should not be done does not mean it will not happen. WMAgent doesn't just have knowledgeable core developers, there are casual developers and ops people as well. So a "belt and suspenders" approach is warranted.

The same is true of code review type questions. Yes, we try to catch mistakes. No one can guarantee they are all caught.

Finally, it seems that python has given us better tools (certainly it will with 2.7) to manage our testing framework. We should definitely be looking at how we can use them.

ericvaandering:
At the risk of opening another can of worms, how do we solve similar issues for Couch and not just MySQL/MariaDB?

ericvaandering:
The tests are failing because the Jenkins jobs are taking much longer (a few times more than usual) for this PR. I've bumped up the timeouts to 4x longer than the average successful job. Hopefully that will help.

PerilousApricot:
@ericvaandering I wasn't trying to start a whole thing -- I was just trying explain why it was there. Like you said, it's a belt and suspenders thing.

On one hand, it's good to mitigate the chances of fumbling a command and blowing away something important. On the other hand, if you're testing locally on your machine, it should fail loudly if your code doesn't tear down the database properly and not paper over that failure (IMO). If the local code doesn't Do The Right Thing and tear things down right, the tests won't fail and whoever reviews the code will accept it, meaning that faulty behavior lives on.

That's the point I was (unsuccessfully) trying to articulate.

PerilousApricot:
As an aside, I'm kinda curious about ```On 0, Andrew Melo <notifications@github.com<mailto:notifications@github.com>> wrote:```. Did I time-travel to jan 1, 1970 (or whenever the unix epoch is)?

ericvaandering:
Not having had a clue about how this really worked before today.... Your point is (rephrasing) that if the DeleteDAO and CreateDAO don't match, we won't know if we are dropping the DB?

ericvaandering:
Maybe today doesn't exist yet in the Julian calendar. :-)

PerilousApricot:
Unless I'm misreading the code, the database is blown away unconditionally during test setUp here: https://github.com/vkuznet/WMCore/blob/a1c508fbb43a089e86c1d01e4d1129a80fe106d7/src/python/WMQuality/TestInit.py#L269

..without letting the user/tester know if there were scraps sitting around in the database from previous tests.

test/python/WMCore_t/WMBS_t/JobSplitting_t/WMBSMergeBySize_t.py
```diff
@@ -30,7 +30,8 @@ def setUp(self):
         """
         self.testInit = TestInit(__file__)
         self.testInit.setLogging()
-        self.testInit.setDatabaseConnection()
+        self.testDB = 'unittest_%s' % self.__class__.__name__
+        self.testInit.prepareDatabase(self.testDB)
```
ericvaandering: A more limited comment. Why not move these two new lines into the setup() of the base class and use super() where needed? Same question with stuff that is repeatedly used in tearDown() 

I just satisfied my own curiously that `self.__class__.__name__` is the name of the derived class, even when accessed in the base class. Also, to be "nice" self.testDB should be defined in `__init__`, not here.

ericvaandering:
"without letting the user/tester know if there were scraps sitting around in the database from previous tests."

What happens in unittest if we were to check if the DB is empty in the tearDown and throw an exception if it wasn't? If that works, it might actually do a better job of localizing these errors. As it is, we only get problems if we run a test *after* the one that failed and then the reasons are obscured.

PerilousApricot:
You're right, that would be ideal. One way or another, waving a flag and saying, "this is not good" helps keep the riff-raff out :)

ericvaandering:
My question can be simplified to "what happens if we throw an exception in tearDown()". Is that an OK thing to do?

PerilousApricot:
I'm going to claim ignorance in how it behaves in 2.7+. In 2.6 at least, you have the problem where if you want to do something like:

```python
def tearDown(self):
    cleanA()
    cleanB()
```

and cleanA() throws, cleanB() won't fire. Depending on what cleanB() does, it might or might not matter, but you could see a couchDB or a temporary directory being torn down in that case.

src/python/WMCore/WMBS/CreateWMBSBase.py
amaltaro: Valentin, to avoid forgetting that we do need to update these columns back to their original values before merging this PR, can you please update them now? It will probably be merged eventually, so better update the code now then let it go to production. Thanks

ericvaandering:
@vkuznet Even after increasing the timeout, many of the slices of this patch cannot complete. The test suite runs some 5-6 times slower I'm guessing. Also, the slices that do complete show a large increase in the number of errors (triple or so as you can see here: https://cmssdt.cern.ch/jenkins/view/DMWM/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud/523/testReport/)

src/python/WMCore/Database/DBFormatter.py
```diff
@@ -109,7 +140,7 @@ def formatCursor(self, cursor, size=10):
         Use fetchmany(size = default arraysize = 50)
 
         """
-	if type(cursor.keys) == types.MethodType:
+        if type(cursor.keys) == types.MethodType:
```
ericvaandering: This appears to be a genuine bug

ericvaandering:
Let me make an alternate proposal and you can tell me if it makes sense:

First, this PR also appears to include the generator from databases stuff that Valentin has been telling us about the last couple of weeks. But isn't that orthogonal to testing harness issues?

Can we leave the existing mechanism in place with it's repeated dropping and creating of a shared database and instead start with a *new* RESTBaseUnitTestWithDBBackend.py (add a 2 if need be) that does the setup/teardown the way Valentin suggests, uses unique database names, maybe gets us ready for the nice thing(s) in 2.7, etc. Then we can do a staged migration to this? 

From the test results its also clear that this modified class is not picking up all the database usage (I doubt it was intended to) and we may need a "UnitTestWithDB" class as well for things like the WMBS tests.

I assume the "reallyDropDatabases" flag on the legacy database and the new behavior on uniquely named databases can co-exist.

vkuznet:
Hi,
to clarify changes I made I want to point out to the existing WMCore code behavior.  For that I'll use existing DBCore_t.py file [1] in master head of WMCore as an example. All other WMCore tests do EXACTLY the same, therefore logic is applied to every database related unittest.

Code [1] does few things,
For a DATABASE environment variable user set before hand of the tests the
[1] does few things:
- in setUp it calls self.testInit.setDatabaseConnection() method
  - this cause a WMinit call to setDatabaseConnection, see [3], which creates
  MySQL object via DBFactory

- in tearDown it calls self.testInit.clearDatabase() which will call WMinit
  clearDatabase() method which by itself calls Destroy class of MySQL object [2]
  The Destroy class does 3 things DROP the database, CREATE the database and
  USE the database, see [2].

- the setup_test.py uses reallyDeleteMyDatabaseAfterEveryTest flag to flip
  global variable in [5] which if it sets will blow up database via [2] anyway.
  Moreover it is done in setDatabaseConnection of [5].

To sum-up, the current design does blow up database if reallyDeleteMyDatabaseAfterEveryTest is set. If it is not the case the database content is carried over from one test to another.

What I did is separated [2] logic into Destroy and Create classes whose names clearly identify their action over DB. Remove global variables and reallyDeleteMyDatabaseAfterEveryTest and set every unittest to use its own DB. I also align logic of transactions in a code

If WMCore team would like to preserve current behavior then the following
changes can be made on top of my patch:
- set testDB to unique database name, i.e. read it from environment variable
- destroy DB content conditionally, i.e. read a glad from environment variable
It can be easily accommodated in a clean way.

Best,
Valentin.

[1] https://github.com/dmwm/WMCore/blob/master/test/python/WMCore_t/Database_t/DBCore_t.py
[2] https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/MySQL/Destroy.py
[3] https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMInit.py
[4] https://github.com/dmwm/WMCore/blob/master/setup_test.py
[5] https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/TestInit.py

vkuznet:
Andrew, 
Of course you're right on this subject and I didn't changed anything in current
code behavior. Current tests are also have tearDown issue you pointed out, I
neither invent or had any intention to change it.
Here is example from current WMCore (all other tests do the same):
https://github.com/dmwm/WMCore/blob/master/test/python/WMComponent_t/JobAccountant_t/JobAccountant_t.py#L96

and  self.testInit.clearDatabase() will call MySQL Destroy class
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/MySQL/Destroy.py

which drops and create database again.

If the exception is thrown in self.testInit.clearDataset(), yes the logic will be the same in
current code too, i.e. the self.testInit.tearDownCouch() will not be called.

vkuznet:
Eric,
I'm not sure what you meant by that, but currently couch setup is more correct
design wise. We connect to Couch, then create db there and tear it down.
With my PR I shown that it can be down in similar way for MySQL/MariaDB.
V.

src/python/WMCore/Database/DBFormatter.py
vkuznet: Yes, and I had no intention to fix all bugs. But this specific one is captured
by pylint.

src/python/WMCore/WMBS/CreateWMBSBase.py
vkuznet: Alan, you're correct and I'll revert this once I'll figure out why MariaDB
can't handle unique names longer than 255. If anyone knows the fix please speak
up.

test/python/WMCore_t/WMBS_t/JobSplitting_t/WMBSMergeBySize_t.py
vkuznet: Eric, in order to do that every single unittests classes need to add __init__
method and properly initialize it from unittest.TestCase class.
Moving self.testDB into __init__ would be proper thing to do too.
But in order to do that we need to agree what WMCore team wants to use
for database, use single database in all tests or use individual database
within unit tests.

vkuznet:
Eric, I can explain time in my design proposal, every unit tests drop database,
create database and create all tables. In current system I don't really know
what's going on, since it depends how BuildBot is run (does it uses
reallyDeleteMyDatabaseAfterEveryTest flag or not), does global variables
changes throw the tests (I didn't inspect all tests classes and have no idea if
someone didn't overwrite globals), etc.  The increase of failures are related
to two facts. I trip down database column names (Alan's comment) due to
constrain from MariaDB and I don't know if shared data is left among tests. My
point is that in current system I can't even say for sure what's going on.

vkuznet:
> First, this PR also appears to include the generator from databases stuff that Valentin has been telling us about the last couple of weeks. But isn't that orthogonal to testing harness issues?

this can go into separate PR indeed.

> 
> Can we leave the existing mechanism in place with it's repeated dropping and creating of a shared database and instead start with a *new* RESTBaseUnitTestWithDBBackend.py (add a 2 if need be) that does the setup/teardown the way Valentin suggests, uses unique database names, maybe gets us ready for the nice thing(s) in 2.7, etc. Then we can do a staged migration to this? 

I am not sure I'm following. Please review the current WMCore test behavior I
posted and really say what logic of TestInit setUp and tearDown should do.
Since I managed to separate Dropping DB and its creation, adjust method to do
what initial logic was meant to do, now we can put things together as plan will
clearly define. I am not sure I understand the purpose of new
RESTBaseUnitTestWithDBBackend.py and how it fits in unittest picture (I think
we're talking about all database releated unittests here).

> 
> From the test results its also clear that this modified class is not picking up all the database usage (I doubt it was intended to) and we may need a "UnitTestWithDB" class as well for things like the WMBS tests.

Here I agree, if WMBS wants to perform integration tests we need new class
which they should use for that. The new class will setup once a database
and let re-use it within integration tests. The setup of database will be done in __init__/__del__
methods rather than in setUp/tearDown.

> 
> I assume the "reallyDropDatabases" flag on the legacy database and the new behavior on uniquely named databases can co-exist.

This can be accomplished in new design proposal via usage of environment
variables rather than globals and hacks in setup_tests.

ericvaandering:
Let me be more clear on my proposal. I am not saying to leave this as it is forever. But I'm leery of big patches that change everything. This is especially difficult since we have another "big patch" outstanding and the two are bound to collide.

So what I would rather see, if possible, is a small patch that adds a new class that "does it right" and just a few tests that use it. Then we can work on migrating the rest until they are all done. That new class should really use inheritance to move as much of the "action" as possible into the new class.

I realize this probably makes TestInit (I think that was the code) a bit more complex in the mean time since it has to have methods to support the new way of doing things and the old, but IMO, this is a more deliberate, conservative way to approach this.

I realize what we have been doing may not be optimal, but it has worked for us, more or less, for several years. Taking 6 months to transition over is not going to kill us.

I'll come back to some of the more minor points next week.

ticoann:
Hi all, Sorry for joining the conversation late.

Here are some of my opinion discussed here.

1. This patch should only concentrate on the unittest setting. (For generator and fix for transaction and connection handling should be separate patches - each)

2. I think it is good idea creating database and destroying database is good idea for each unittest. It is a cleaner solution than keep tracking the tables and drop the necessary tables. I understand the concern about destroying unwanted database but we can add more safety measure.
   a. for the production deployment doesn't contain unittest code (currently implemented).
   b. we absolutely avoid creating the same database name as production database wmagent, reqmgr, etc. (We can have a specific rule for unittest db as Valentin showed - unittest prefix and unittest will auto generate the db name when we set up the test. Also for tearDown, we can always do the Lexicon matching to check whether db contains the prefix before delete them)

3. I think initial patch should just include the testInit changes. Basically make this function to do following.
https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/TestInit.py#L158
   a.  if there is db exist for particular unittest delete. ( this can be skipped if we use python 2.7 addcleanup)
   b. create the database
   c. create the db connection

And tearDown method will drop the db.
That way I don't think individual unittest code need to be changed initially and avoid conflict with other unittest fix. After unittest fix are upto the certain standard we can change to proper method name or divide to multiple methods.

Misc:
I think we can move  the unittest rpm setting to python 2.7 now (and use addcleanup - but not just for that), since we have the repository. (There is concern that we test the code against python 2.7 and production runs 2.6 but we need to do this eventually anyway and current code is python 2.7 compatible. - One concern is Tier0 code which shares the same unittest strtucture. But I think it is not much to change Tier0 to make python 2.7 ready)

ticoann:
One more thing: As others said, we can't include the schema changes here. (reducing the variable length).  I get the same error Valentin gets. (I set up unittest env manually not following the wiki in local not following.) :smile:  But we use the same set up in production (MariaDB with the same schema). I wonder why we don't have the same problem. (We had but I thought Alan fixed)

https://github.com/dmwm/WMCore/pull/5588

vkuznet:
Seangchan,
if I read changes correctly, Alan's patch does not really address issue with unique indexes.
It is something else on DB setup which needs to be done (may be at compile time). If it is
configuration parameter before compiling the database it explains difference in VM
deployment versus local setup we had.
Valentin.

vkuznet:
> 1. This patch should only concentrate on the unittest setting. (For generator and fix for transaction and connection handling should be separate patches - each)

agreed, I'll provide separate patches.

> 
> 2. I think it is good idea creating database and destroying database is good idea for each unittest. It is a cleaner solution than keep tracking the tables and drop the necessary tables. I understand the concern about destroying unwanted database but we can add more safety measure.
>    a. for the production deployment doesn't contain unittest code (currently implemented).
>    b. we absolutely avoid creating the same database name as production database wmagent, reqmgr, etc. (We can have a specific rule for unittest db as Valentin showed - unittest prefix and unittest will auto generate the db name when we set up the test. Also for tearDown, we can always do the Lexicon matching to check whether db contains the prefix before delete them)
> 
> 3. I think initial patch should just include the testInit changes. Basically make this function to do following.
> https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/TestInit.py#L158
>    a.  if there is db exist for particular unittest delete. ( this can be skipped if we use python 2.7 addcleanup)
>    b. create the database
>    c. create the db connection
> 
> And tearDown method will drop the db.
> That way I don't think individual unittest code need to be changed initially and avoid conflict with other unittest fix. After unittest fix are upto the certain standard we can change to proper method name or divide to multiple methods.
> 

For this plan to work we need to re-factor TestInit internals several times. In
current implementation there are two methods:
- all unit tests call setDatabaseConnection() in their setUp() methods
  therefore we need to implement your a/b/c steps over there
- all unit tests call clearDatabase() in their tearDown() methods
  therefore we need to implement deletion of database over there

I would assume that we'll create proper internal methods in TestInit which can
be used in re-factoring I mentioned above. Then we'll need to re-factor
TestInit again to replace setDatabaseConnection() and clearDatabase() with
proper structure. Am I correct?

> Misc:
> I think we can move  the unittest rpm setting to python 2.7 now (and use addcleanup - but not just for that), since we have the repository. (There is concern that we test the code against python 2.7 and production runs 2.6 but we need to do this eventually anyway and current code is python 2.7 compatible. - One concern is Tier0 code which shares the same unittest strtucture. But I think it is not much to change Tier0 to make python 2.7 ready)
> 

I'm not sure that we need to mix that.

vkuznet:
Eric, here is abstract class which may fit the bill.

```
#!/usr/bin/env python
#-*- coding: utf-8 -*-
#pylint: disable=
"""
File       : WMCoreUnitTest.py
Description: Base class for WMCore database related unit tests.
             It defines test database name based on user preferences specified via
             environment variables and setup logic to setup/tear down actions
             based on database.
"""

# system modules
import os
import sys
import unittest

class WMCoreUnitTest(unittest.TestCase):
    """
    Abstract class to implement proper setup/tear down methods for communicating
    with WMCore database back-end in unittests.
    """
    def __init__(self, methodName='runTest'):
        super(WMCoreUnitTest, self).__init__(methodName)
        self.dbName = os.environ.get('WMCORE_TEST_DATABASE', \
                'unittest_%s' % self.__class__.__name__)
        print("WMCoreTestDB test %s database" % self.dbName)
        enforce = os.environ.get('WMCORE_TEST_DATABASE_DELETE', False)
        if  enforce:
            print("WMCORE_TEST_DATABASE_DELETE=%s, will delete %s" % (enforce, self.dbName))
            self.deleteDatabase()
            self.createDatabase()

    def setUp(self):
        """Setup database for unittests"""
        if  self.dbName.startswith('unittest'):
            self.deleteDatabase()
            self.createDatabase()

    def tearDown(self):
        """Tear down database"""
        if self.dbName.startswith('unittest'):
            self.deleteDatabase()

    def createDatabase(self):
        """Abstract method to create test database, must be implemented in sub-classes"""
        print("Create %s" % self.dbName)

    def deleteDatabase(self):
        """Abstract method to delete test database, must be implemented in sub-classes"""
        print("Delete %s" % self.dbName)
```

And the logic I followed:
- read database name from WMCORE_TEST_DATABASE environment, if it is not set define
  database name wrt class name.
- re-create database (i.e. delete and create) if its name starts with unittest
- delete database in tear down if its name starts with unittest
- define abstract methods to delete and create databases
- use enforcement mechanism to re-create database ONCE in constructor of the class, this can be used
  to address Andrew's worries about integration scenario. The database will be deleted
  and created again only once for entire unittest (not for every method).

Then I foresee that we can make TestInit to inherit from this class and modify
its logic to use base class methods.

Best,
Valentin

ticoann:
> For this plan to work we need to re-factor TestInit internals several times. In current implementation there are two methods: - all unit tests call setDatabaseConnection() in their setUp() methods therefore we need to implement your a/b/c steps over there - all unit tests call clearDatabase() in their tearDown() methods therefore we need to implement deletion of database over there I would assume that we'll create proper internal methods in TestInit which can be used in re-factoring I mentioned above. Then we'll need to re-factor TestInit again to replace setDatabaseConnection() and clearDatabase() with proper structure. Am I correct?

Yes, that is correct. But I think refactoring shouldn't be that much hassle.
As you said 
1. create internal method to to do a/b/c. and 
2. make those calls in side setDatabaseConnection method
3. change clearDatabase call to delete the database.

When all the test is done, we just need to change the name of setDatabaseConnection to proper name and replace all unittest. (Assuming a/b/c need to be called always together in set up).

The point I wanted to make is, if possible, we keep the changes only in TestInit initially.
Thanks,

vkuznet:
> 1. This patch should only concentrate on the unittest setting. (For generator and fix for transaction and connection handling should be separate patches - each)

Here is PR for transaction logic: https://github.com/dmwm/WMCore/pull/6274

Once it is in place I can provide another patch on top of it to fix connection
issues. Then I can patch on top of that data structure logic, i.e. convert code
to use generators.

ticoann:
> Here is PR for transaction logic: https://github.com/dmwm/WMCore/pull/6274 Once it is in place I can provide another patch on top of it to fix connection issues. Then I can patch on top of that data structure logic, i.e. convert code to use generators.

Thanks Valentin, Sorry I didn't know you have to create patch in sequence. The transaction and connection patch doesn't require other parts of code change but generator patch does, correct?

In that case, we can have one patch for transaction and connection (two commits will be fine) but generator patch separately. Transaction and connection patch won't be merged at least the cmsweb web validation is finished. (We can run the test after that, Is it OK with you?).

ericvaandering:
Commenting on you proposed subclass:

I don't get " use enforcement mechanism to re-create database ONCE in constructor of the class, this can be used
  to address Andrew's worries about integration scenario. The database will be deleted
  and created again only once for entire unittest (not for every method)." I think @PerilousApricot  concern was only with databases that were not left totally empty by the delete DAO. I don't understand how this helps that scenario.

Second question, why does deleting/creating the database have to be implemented in the subclass? Some or all of what needs to be done could be implemented in the base class, right?

Comment: we have some (many) unit tests who have nothing to do with underlying databases. Maybe they don't benefit at all from subclassing our own test case. If so, maybe just a name change from WMCoreUnitTest is appropriate.

vkuznet:
On  0, ticoann <notifications@github.com> wrote:
> > Here is PR for transaction logic: https://github.com/dmwm/WMCore/pull/6274 Once it is in place I can provide another patch on top of it to fix connection issues. Then I can patch on top of that data structure logic, i.e. convert code to use generators.
> 
> Thanks Valentin, Sorry I didn't know you have to create patch in sequence. The transaction and connection patch doesn't require other parts of code change but generator patch does, correct?

Since transaction, connection and improper usage of data structures/database
proxyresults objects are within processData() method fixing them
separately and merging those patches will cause conflict. That's why they
should go in sequence.  So either you accept sequence of patches or I can do
everything in one shot, see src/python/WMCore/Database/DBCore.py in
https://github.com/dmwm/WMCore/pull/6227/files PR.

vkuznet:
> 
> I don't get " use enforcement mechanism to re-create database ONCE in constructor of the class, this can be used
>   to address Andrew's worries about integration scenario. The database will be deleted
>   and created again only once for entire unittest (not for every method)." I think @PerilousApricot  concern was only with databases that were not left totally empty by the delete DAO. I don't understand how this helps that scenario.

In current design, if I do
1. python setup.py test --buildBotMode=true --reallyDeleteMyDatabaseAfterEveryTest=true ...
and
2. python test/python/.../<SomeFile>_t.py
are doing different things. In first case, database will be destroyed after
every test and it will be re-created. While this is not true in second case.  I
provided details that this is due to usage of globals and setup_test.py sets
the trashDatabases in TestInit.py This is why it is REALLY bad, since we can't
reproduce unittest behavior using *1* and *2* way to run unit tests.  Well, we
can make *2* to be identical to *1* if we don't use
reallyDeleteMyDatabaseAfterEveryTest flag, so we even have more complex
scenario.

In new design, using environment we will achieve that all aforementioned cases
will run in identical environment. For example, we set environment
and run *1* or *2* and they will deal with database in the same unique way
(whatever it is). Either database will be created for every
unit tests with unique name, or we'll use one common database for all tests.
And, we will have control to either delete database or not. But both *1*
and *2* way to run unit tests will behave identically from database
deletion/creation point of view and regardless how we run unit tests (via
buildbot or not).

Regarding left overs from DAO. Starting from clean db is a way to go. Doing
this way we guarantee that we delete previous context and created fresh DB.  If
user need to understand why/what was left after previous test they can inspect
left-overs and decide if they want to keep them or not.  The control can be
done via setting environment rather through globals and flags.

> 
> Second question, why does deleting/creating the database have to be implemented in the subclass? Some or all of what needs to be done could be implemented in the base class, right?

it depends on how you want to design to go. It can be either be part of base
class or delegated to another level.

> 
> Comment: we have some (many) unit tests who have nothing to do with underlying databases. Maybe they don't benefit at all from subclassing our own test case. If so, maybe just a name change from WMCoreUnitTest is appropriate.

Then don't inherit from this class. Probably WMCoreUnitTest is not proper name
and should be WMCoreDatabaseUnitTest or similar.

vkuznet:
@amaltaro , do we need to keep this open. It is already 7 years old and was done for py2 era.

amaltaro:
@vkuznet you are the main contributor of this PR, if you are okay with closing this out, I am all in favor.

As you said, it's a 7 years old code proposal, with lots of discussion and good content, but if we haven't converged on this so far, it might be better to come back with a clean proposal when we deem such development necessary again. 

It seems to be addressing many things in one PR as well. If you feel like there is something that must be changed, please try to isolate it and create a new GH issue. 

I am closing this PR, please reopen it in case I am missing anything. Thank you all for your contribution and rich discussion!

PR #6261, opened 2015-09-18
Add DBSRESTFormatter

yuyiguo:


ericvaandering:
This is going to fail our checks because we now make every new file contain (at the beginning)

```python
from __future__ import (division, print_function)
```

even though this module will probably never print or divide anything. :-)

src/python/WMCore/WebTools/DBSRESTFormatter.py
```diff
@@ -0,0 +1,22 @@
...
+A basic REST formatter. The formatter takes the data from the API call, turns it into the
+appropriate format and sets the CherryPy header appropriately.
+
+Could add YAML via http://pyyaml.org/
+"""
+from WMCore.WebTools.RESTFormatter import RESTFormatter
+
+class DBSRESTFormatter(RESTFormatter):
+    RESTFormatter.__init__(self, config)
+    def __init__(self, config):
```
ericvaandering: Don't you want the order of these two lines (13 & 14) swapped?

ericvaandering: And no need for "self" in `RESTFormatter.__init__(self, config)`

ticoann: Actually self is needed, isn't it (since is calling using class not instance)?

ticoann:
@yuyiguo, Hi Yuyi, you said this is needed urgently, right? Could you update your patch with what Eric pointed out

1. move  following line inside the __init__.
```
RESTFormatter.__init__(self, config)
```
2. add future import at the very top of the module. ( This should be imported first than any other import)
```
from __future__ import (division, print_function)
```
Then, I will create the new tag with this.
Thanks

src/python/WMCore/WebTools/DBSRESTFormatter.py
```diff
@@ -0,0 +1,23 @@
...
+appropriate format and sets the CherryPy header appropriately.
+
+Could add YAML via http://pyyaml.org/
+"""
+from __future__ import (division, print_function)
+from WMCore.WebTools.RESTFormatter import RESTFormatter
+
+class DBSRESTFormatter(RESTFormatter):
+    def __init__(self, config):
+	RESTFormatter.__init__(self, config)
```
ericvaandering: a) indentation is wrong
b) self is not needed. You will get an error the first time you try to use this.

yuyiguo:
Indentation is correct. I though it was wrong at the beginning too. That was because the display. If you click on  view code https://github.com/yuyiguo/WMCore/blob/master/src/python/WMCore/WebTools/DBSRESTFormatter.py
It is correctly indented.

ericvaandering:
Very odd. Are you sure that line 15 is not spaced with two tabs instead of 8 spaces? That could do this to diff. And spacing with tabs is evil. :-)

src/python/WMCore/WebTools/DBSRESTFormatter.py
ticoann: Hi Eric,  I think actually self is needed. I think it is the other way around (it will get an error without self).

ericvaandering: You are correct. Yeah, different than calling a method. And I have still not sorted out if this form or one using super() is preferred.

yuyiguo:
@ticoann Thanks Seangchan for committing and tagging.
@ericvaandering, You are right, Eric. I like using tab with vi when coding python. This code was copied and edited from the original RESTFormatter.java . So mixed. :-(  

The tools used to migrate to python 2.7/3 will fix these tab and space mixing?  If not, we are going to have a lot of manual work to do. python3 does not like to mix tab and space at what level, file level or package level?  I corrected the code with 4 space and pushed it to github,  but github indicated that everything was already updated. So github would not consider this was a change and update the repository?

ericvaandering:
Doesn't vi have a way to use the tab key but it actually inserts spaces? This is the way I've used every editor for many years. 

I assume it will be on the file level that you may not mix.

yuyiguo:
This is what I used, but still there is mixing the two ...

PR #6265, opened 2015-09-21
Should be set from blacklist, not whitelist

ericvaandering:


ericvaandering:
@ticoann please merge

PR #6267, opened 2015-09-21
direct update agent status for Tier0 agent as well

### LINKED ISSUE #6256: Tier0 wmstats agent document is not updated properly
When we restart an agent, the agent info section in wmstats is not correctly updated. It shows there is an error with AnalyticsDataCollector component, when it is not the case and agent is running properly. @ticoann

ticoann:
fixes #6256


ticoann:
@ebohorqu, Hi Eliana,  could you patch tier0 test agent whether this fixes the issue?

PR #6271, opened 2015-09-22
Update groups:roles authorized

amaltaro:
Not sure we really care about these changes, since it's six of one and half a dozen of the other :-)

I removed dataops:production-manager role, since nobody has such role in SiteDB right now.
and added reqmgr:data-manager with the goal to start moving PPD/PdmV folks into this group instead of giving them reqmgr:admin role. Right now they seem to have exactly the same privileges, but at least we can start better organizing people in sitedb.

@ticoann what do you think? No hurry, it's not urgent


ticoann:
Thanks Alan for clean up the roles.

PR #6272, opened 2015-09-22
Enable double harvesting upload in the json templates

amaltaro:
Broen asked me to start uploading root files for both servers (including a new dedicated relval node) such that they can better validate and have reliable testing in the new Guis.


amaltaro:
changes only json templates. @ticoann merge it please

PR #6277, opened 2015-09-23
LogDB has to use environment settings

ericvaandering:


ericvaandering:
@ticoann please merge. I leave the last one to you that the thread names are not matching up.

ticoann:
Thanks Eric,

PR #6282, opened 2015-09-24
don't crash if the condor log has missing fields

hufnagel:
Treat the condor log as empty in this case, which marks a job as complete and this will trigger checks for the job report.

Reason for this is a condor log from an old job that was killed before it went to running in condor. Lots of stuff is missing from the condor log in that case.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -1049,7 +1049,7 @@ class PyCondorPlugin(BasePlugin):
                 if results[i].get("JobStatus",0)==3:
                     continue
                 else :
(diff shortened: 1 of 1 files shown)


hufnagel:
This is deployed in the production Tier0 already.

hufnagel:
@ticoann , can you review this or have someone else review this ? Should be safe, as I said, we actually are running this code in the production Tier0.

ticoann:
That sounds fine to me, This jobs will eventually marked (killed or failed) from JobAccountant anyway, right?

hufnagel:
Correct. Also, that code path to just mark the job as complete and let the JobAccountant deal with it is already used for the case of an empty condor log file, I am just reusing it also for the case of a condor log file with missing information.

PR #6283, opened 2015-09-24
New base class to deal with database(s) in unit tests

vkuznet:


vkuznet:
I didn't implemented deleteDatabase and createDatabase methods since further changes will be required for that. In particular, changes to MySQL Create class and decouple it into Create/Destroy, see https://github.com/vkuznet/WMCore/blob/fix-unittests/src/python/WMCore/Database/MySQL/Create.py and https://github.com/vkuznet/WMCore/blob/fix-unittests/src/python/WMCore/Database/MySQL/Destroy.py

ericvaandering:
In your "Destroy" at lines 15-18, do I read this correct that the idea is just to destroy the currently being-used database?

src/python/WMCore/Database/DBTest.py
```diff
@@ -0,0 +1,48 @@
...
+import unittest
+
+class DBTest(unittest.TestCase):
+    def __init__(self, methodName='runTest'):
+        super(WMCoreUnitTest, self).__init__(methodName)
+        self.dbName = os.environ.get('WMCORE_TEST_DATABASE', \
+                'unittest_%s' % self.__class__.__name__)
+        print("WMCoreTestDB test %s database" % self.dbName)
+        enforce = os.environ.get('WMCORE_TEST_DATABASE_DELETE', False)
+        if  enforce:
```
ericvaandering: Ok, so I think I understand this point now. This is used to crop and recreate the database once per task iff person running it is setting the database name themselves (line 22) and has this flag set.

src/python/WMCore/Database/DBTest.py
```diff
@@ -0,0 +1,48 @@
...
+            print("WMCORE_TEST_DATABASE_DELETE=%s, will delete %s" % (enforce, self.dbName))
+            self.deleteDatabase()
+            self.createDatabase()
+
+    def setUp(self):
+        """Setup database for unittests"""
+        if  self.dbName.startswith('unittest'):
```
ericvaandering: Why did you choose here to not start with unittest_ or even just check for `'unittest_%s' % self.__class__.__name__`?

src/python/WMCore/Database/DBTest.py
```diff
@@ -0,0 +1,48 @@
...
+            self.createDatabase()
+
+    def tearDown(self):
+        """Tear down database"""
+        if self.dbName.startswith('unittest'):
+            self.deleteDatabase()
+
+    def createDatabase(self):
+        """Abstract method to create test database, must be implemented in sub-classes"""
+        print("Create %s" % self.dbName)
```
ericvaandering: Unless I am missing something, you should be able to write these as well and then, if needed, have the derived class override it or call this and then do whatever else might be needed.

vkuznet:
Yes, the problem is two-fold. First, we need properly fix Destroy class.
Second, we need to agree what DATABASE environment will points to.
Let me elaborate on later. Currently the DATABASE points to
mysql://localhost/SOME_DATABASE

So it is has full urn, i.e. the name of the database to connect to.
If this is the case we need to get current database in Destroy class.

We can switch to use short urn, e.g. mysql://localhost
then we can pass around dbName in Destroy and Create.

Again, we can adjust to whatever scenario, and I'm willing to modify
Destroy accordingly, but we need to agreed how we setup things.

My view, that we should have consistencies among databases, e.g.
MySQL and CouchDB. In that case we should use short urn.
It has other advantages. We can connect to DB regardless of database usage.
If no database exists or we destroy it we still will have a connection.
This is how CouchDB behvaior is done.

src/python/WMCore/Database/DBTest.py
vkuznet: Yes, dbName would be either what user provides or unittest_bla.
User can provide additional environment to enforce deletion of database.

src/python/WMCore/Database/DBTest.py
vkuznet: Well, probably this will serve better:

```
if self.dbName == 'unittest_%s' % self.__class__.__name__:
```

if you worry about collision from environment.

src/python/WMCore/Database/DBTest.py
vkuznet: Yes, I can, but I can't use current
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/MySQL/Destroy.py

since its logic should be fixed. This is why I suggest to create
two new classes: CreateDatabase.py and DeleteDatabase.py which will
do proper things. If you'll agree I can create them as I shown them in
my original PR.

src/python/WMCore/Database/DBTest.py
PerilousApricot: Using the "reply-to-email" feature and not snipping previous replies makes these notifications an absolute catastrophe.

vkuznet: > Using the "reply-to-email" feature and not snipping previous replies makes these notifications an absolute catastrophe.

I'll try to reply via web interface instead of my email client next time.

vkuznet:
> Imports for Python3 compatability missing in new files. Please fix this:
> * New file src/python/WMCore/Database/DBTest.py does not use python 3 division. Please add `from __future__ import division`.

I don't think it is relevant, but I don't mind to add division.

ericvaandering:
I'm out of the office for the afternoon, but I would do a new method for create and destroy as well. Might as well make a clean break

And yes, you will never divide in the database, but we are making this requirement  and yes, you will never divide in the database code, but we are making this requirement on all new Python  files

vkuznet:
Eric, what did you mean by divide?

vkuznet:
Eric, I understood now that by divide you meant divide function from __future__
module.

vkuznet:
I've added new Create/DeleteDatabase classes and use them in DBTests following
previous WMInit logic. Please review.

amaltaro:
test this please

vkuznet:
I found that there many places where dbi object is defined. Moreover since it attached (shared) to main thread it seems that in order to make all unit tests work and talk to different DBs we need to fix all instances where dbi object is assigned. Here is a full list of all places:

```
src/python/WMCore/Agent/Harness.py
186-
187:                myThread.dbi = myThread.dbFactory.connect()
188-                myThread.transaction = Transaction(myThread.dbi)
--
191-
192:                myThread.dbi = myThread.config.CoreDatabase.connectUrl
193-                myThread.sql_transaction = False

src/python/WMCore/HTTPFrontEnd/RequestManager/Admin.py
36-        # Get it from the DBFormatter superclass
37:        myThread.dbi = self.dbi
38-

src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py
108-        # Get it from the DBFormatter superclass
109:        myThread.dbi = self.dbi
110-

src/python/WMCore/HTTPFrontEnd/RequestManager/BulkOperations.py
25-        # Get it from the DBFormatter superclass
26:        myThread.dbi = self.dbi
27-

src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrBrowser.py
81-        # Get it from the DBFormatter superclass
82:        myThread.dbi = self.dbi
83-

src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py
199-        # Get it from the DBFormatter superclass
200:        myThread.dbi = self.dbi
201-

src/python/WMCore/HTTPFrontEnd/RequestManager/WebRequestSchema.py
39-        # Get it from the DBFormatter superclass
40:        myThread.dbi = self.dbi
41-

src/python/WMCore/ThreadPool/ThreadSlave.py
109-        # thread so we have only one engine in the application.
110:        myThread.dbi = myThread.dbFactory.connect()
111-        logging.info("THREAD: Initializing default transaction")

src/python/WMCore/WMInit.py
134-        myThread.dbFactory = DBFactory(logging, dbConfig, options)
135:        myThread.dbi = myThread.dbFactory.connect()
136-

src/python/WMCore/WorkQueue/WorkQueue.py
517-                        myThread = threading.currentThread()
518:                        myThread.dbi = self.conn.dbi
519-                        myThread.logger = self.logger

src/python/WMCore/WorkerThreads/BaseWorkerThread.py
117-        logging.info("Initialising default database")
118:        myThread.dbi = myThread.dbFactory.connect()
119-        logging.info("Initialising default transaction")
```

@ticoann, I need to know your opinion on this subject before making any changes to classes.

vkuznet:
@amaltaro , do we need to keep this open. It is already 7 years old and was done for py2 era.

amaltaro:
@vkuznet Valentin, if we haven't converged on this yet, then indeed I'd suggest to close it. 

If you feel like something needs to be followed up and fixed this or in the coming year, please open a new GH issue.
I am closing it, but feel free to reopen it in case I am missing anything.

PR #6284, opened 2015-09-24
Fix transaction logic; connection usage; use generator objects in DBCore.

vkuznet:
Invert transaction logic; change connection usage; convert processData to use generator objects; adjust ResultSet class

### DIFF
--- a/src/python/WMCore/Database/DBCore.py
+++ b/src/python/WMCore/Database/DBCore.py
@@ -7,8 +7,6 @@ Core Database APIs
 
 """
 
-
-
 from WMCore.DataStructs.WMObject import WMObject
 from WMCore.Database.ResultSet import ResultSet
 from copy import copy
@@ -29,21 +27,21 @@ class DBInterface(WMObject):
         Support executemany()
     """
 
-    logger = None
-    engine = None
-
     def __init__(self, logger, engine):
+        super(DBInterface, self).__init__(self)
         self.logger = logger
         self.logger.info ("Instantiating base WM DBInterface")
         self.engine = engine
         self.maxBindsPerQuery = 500
 
-    def buildbinds(self, sequence, thename, therest=[{}]):
+    def buildbinds(self, sequence, thename, therest=None):
         """
         Build a list of binds. Can be used recursively, e.g.:
         buildbinds(file, 'file', buildbinds(sename, 'location'), {'lumi':123})
         TODO: replace with an appropriate map function
         """
+        if  not therest:
+            therest = [{}]
         binds = []
         for r in sequence:
             for i in self.makelist(therest):
@@ -63,14 +61,9 @@ class DBInterface(WMObject):
             resultProxy = connection.execute(s)
         else:
             resultProxy = connection.execute(s, b)
-
-        if returnCursor:
-            return resultProxy
-
-        result = ResultSet()
-        result.add(resultProxy)
-        resultProxy.close()
-        return result
+        yield resultProxy
+        if not returnCursor:
+            resultProxy.close()
 
     def executemanybinds(self, s=None, b=None, connection=None,
                          returnCursor=False):
@@ -92,27 +85,19 @@ class DBInterface(WMObject):
 
         s = s.strip()
         if s.lower().endswith('select', 0, 6):
-            """
-            Trying to select many
-            """
+            # Trying to select many
             if returnCursor:
-                result = []
                 for bind in b:
-                    result.append(connection.execute(s, bind))
+                    result = connection.execute(s, bind)
+                    yield result
             else:
-                result = ResultSet()
                 for bind in b:
                     resultproxy = connection.execute(s, bind)
-                    result.add(resultproxy)
+                    yield resultproxy
                     resultproxy.close()
-
-            return self.makelist(result)
-
-        """
-        Now inserting or updating many
-        """
+            return
         result = connection.execute(s, b)
-        return self.makelist(result)
+        yield result
 
     def connection(self):
         """
@@ -120,85 +105,95 @@ class DBInterface(WMObject):
         """
         return self.engine.connect()
 
+    def processData(self, sqlstmt, binds=None, conn=None,
+                    transaction=True, returnCursor=False):
+        """Wrapper around _processData generator"""
+        if  not binds:
+            binds = {}
+        connection = conn if conn else self.connection()
+#        print("### processData engine obj=%s, engine=%s, sql=%s, binds=%s, conn=%s, trainsaction=%s" \
+#                % (hex(id(self.engine)), self.engine.__dict__['engine'], sqlstmt, binds, conn, transaction))
+        gen = self._processData(sqlstmt, binds, connection, transaction, returnCursor)
+        rset = ResultSet()
+        for resultproxy in gen:
+            rset.add(resultproxy)
+        if not conn:
+            connection.close()
+        return [rset]
 
-    def processData(self, sqlstmt, binds={}, conn=None,
-                    transaction=False, returnCursor=False):
+    def _processData(self, sqlstmt, binds=None, connection=None,
+                    transaction=True, returnCursor=False):
         """
         set conn if you already have an active connection to reuse
         set transaction = True if you already have an active transaction
 
         """
-        connection = None
-        try:
-            if not conn:
-                connection = self.connection()
-            else:
-                connection = conn
+        if  not binds:
+            binds = {}
+        # Can take either a single statement or a list of statements and binds
+        sqlstmt = self.makelist(sqlstmt)
+        binds = self.makelist(binds)
+        if len(sqlstmt) > 0 and (len(binds) == 0 or (binds[0] == {} or binds[0] == None)):
+            if transaction:
+                trans = connection.begin()
 
-            result = []
-            # Can take either a single statement or a list of statements and binds
-            sqlstmt = self.makelist(sqlstmt)
-            binds = self.makelist(binds)
-            if len(sqlstmt) > 0 and (len(binds) == 0 or (binds[0] == {} or binds[0] == None)):
-                # Should only be run by create statements
-                if not transaction:
-                    #WMCore.WMLogging.sqldebug("transaction created in DBInterface")
-                    trans = connection.begin()
+            for i in sqlstmt:
+                gen = self.executebinds(i, connection=connection,
+                                      returnCursor=returnCursor)
+                for rec in gen:
+                    yield rec
 
-                for i in sqlstmt:
-                    r = self.executebinds(i, connection=connection,
-                                          returnCursor=returnCursor)
-                    result.append(r)
+            if transaction:
+                trans.commit()
+        elif len(binds) > len(sqlstmt) and len(sqlstmt) == 1:
+            #Run single SQL statement for a list of binds - use execute_many()
+            if transaction:
+                trans = connection.begin()
+            while(len(binds) > self.maxBindsPerQuery):
+                gen = self._processData(sqlstmt, binds[:self.maxBindsPerQuery],
+                                               connection=connection, transaction=True,
+                                               returnCursor=returnCursor)
+                for rec in gen:
+                    yield rec
+                binds = binds[self.maxBindsPerQuery:]
 
-                if not transaction:
-                    trans.commit()
-            elif len(binds) > len(sqlstmt) and len(sqlstmt) == 1:
-                #Run single SQL statement for a list of binds - use execute_many()
-                if not transaction:
-                    trans = connection.begin()
-                while(len(binds) > self.maxBindsPerQuery):
-                    result.extend(self.processData(sqlstmt, binds[:self.maxBindsPerQuery],
-                                                   conn=connection, transaction=True,
-                                                   returnCursor=returnCursor))
-                    binds = binds[self.maxBindsPerQuery:]
+            for i in sqlstmt:
+                gen = self.executemanybinds(i, binds, connection=connection,
+                                                    returnCursor=returnCursor)
+                for rec in gen:
+                    yield rec
 
-                for i in sqlstmt:
-                    result.extend(self.executemanybinds(i, binds, connection=connection,
-                                                        returnCursor=returnCursor))
-                if not transaction:
-                    trans.commit()
-            elif len(binds) == len(sqlstmt):
-                # Run a list of SQL for a list of binds
-                if not transaction:
-                    trans = connection.begin()
+            if transaction:
+                trans.commit()
+        elif len(binds) == len(sqlstmt):
+            # Run a list of SQL for a list of binds
+            if transaction:
+                trans = connection.begin()
 
-                for i, s in enumerate(sqlstmt):
-                    b = binds[i]
+            for i, s in enumerate(sqlstmt):
+                b = binds[i]
 
-                    r = self.executebinds(s, b, connection=connection,
-                                          returnCursor=returnCursor)
-                    result.append(r)
+                gen = self.executebinds(s, b, connection=connection,
+                                      returnCursor=returnCursor)
+                for rec in gen:
+                    yield rec
 
-                if not transaction:
-                    trans.commit()
-            else:
-                self.logger.exception(
-                    "DBInterface.processData Nothing executed, problem with your arguments")
-                self.logger.exception(
-                    "DBInterface.processData SQL = %s" % sqlstmt)
-                WMCore.WMLogging.sqldebug('DBInterface.processData  sql is %s items long' % len(sqlstmt))
-                WMCore.WMLogging.sqldebug('DBInterface.processData  binds are %s items long' % len(binds))
-                assert_value = False
-                if len(binds) == len(sqlstmt):
-                    assert_value = True
-                WMCore.WMLogging.sqldebug('DBInterface.processData are binds and sql same length? : %s' % (assert_value))
-                WMCore.WMLogging.sqldebug('sql: %s\n binds: %s\n, connection:%s\n, transaction:%s\n' %
-                                           (sqlstmt, binds, connection, transaction))
-                WMCore.WMLogging.sqldebug('type check:\nsql: %s\n binds: %s\n, connection:%s\n, transaction:%s\n' %
-                                           (type(sqlstmt), type(binds), type(connection), type(transaction)))
-                raise Exception("""DBInterface.processData Nothing executed, problem with your arguments
-                Probably mismatched sizes for sql (%i) and binds (%i)""" % (len(sqlstmt), len(binds)))
-        finally:
-            if not conn and connection != None:
-                connection.close() # Return connection to the pool
-        return result
+            if transaction:
+                trans.commit()
+        else:
+            self.logger.exception(
+                "DBInterface.processData Nothing executed, problem with your arguments")
+            self.logger.exception(
+                "DBInterface.processData SQL = %s" % sqlstmt)
+            WMCore.WMLogging.sqldebug('DBInterface.processData  sql is %s items long' % len(sqlstmt))
+            WMCore.WMLogging.sqldebug('DBInterface.processData  binds are %s items long' % len(binds))
+            assert_value = False
+            if len(binds) == len(sqlstmt):
+                assert_value = True
+            WMCore.WMLogging.sqldebug('DBInterface.processData are binds and sql same length? : %s' % (assert_value))
+            WMCore.WMLogging.sqldebug('sql: %s\n binds: %s\n, connection:%s\n, transaction:%s\n' %
+                                       (sqlstmt, binds, connection, transaction))
+            WMCore.WMLogging.sqldebug('type check:\nsql: %s\n binds: %s\n, connection:%s\n, transaction:%s\n' %
+                                       (type(sqlstmt), type(binds), type(connection), type(transaction)))
+            raise Exception("""DBInterface.processData Nothing executed, problem with your arguments
+            Probably mismatched sizes for sql (%i) and binds (%i)""" % (len(sqlstmt), len(binds)))
--- a/src/python/WMCore/Database/DBFormatter.py
+++ b/src/python/WMCore/Database/DBFormatter.py
@@ -46,7 +46,7 @@ class DBFormatter(WMObject):
         for r in result:
             for i in r.fetchall():
                 row = []
-                for j in i:
+                for j in i.values():
                     row.append(j)
                 out.append(row)
             r.close()
@@ -68,22 +68,8 @@ class DBFormatter(WMObject):
         """
         dictOut = []
         for r in result:
-            descriptions = r.keys
             for i in r.fetchall():
-                #WARNING: this can generate errors for some stupid reason
-                # in both oracle and mysql.
-                entry = {}
-                for index in xrange(0,len(descriptions)):
-                    # WARNING: Oracle returns table names in CAP!
-                    if type(i[index]) == unicode:
-                        entry[str(descriptions[index].lower())] = str(i[index])
-                    else:
-                        entry[str(descriptions[index].lower())] = i[index]
-
-                dictOut.append(entry)
-
-            r.close()
-
+                dictOut.append(i)
         return dictOut
 
     def formatOneDict(self, result):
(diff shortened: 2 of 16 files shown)


ericvaandering:
This appears to be the patch that is causing things to time out, running at least 4 times as slow on many of the slices. Can you do some local tests and see if you expect that?

vkuznet:
Eric,
I need more input on that. Could you please post exact command how you got your
4 times increase in time. I don't see it in my local stand-alone test.

I created local test, see code here:
https://gist.github.com/vkuznet/33202de85bc3a36bd2f0

It is based on DBCore_t.py except that I didn't used TestInit/WMInit classes
and rather used their logic how to connect to DB.

I run test with my patch and without. In both cases I got ~10seconds,

```
# use generator/connection/transaction patch
real0m10.443s
user0m0.905s
sys0m0.213s
```

```
# use masthead of WMCore
real0m10.452s
user0m0.907s
sys0m0.205s
```

This is stand-alone program, it uses WMCore core modules, but it does not
uses unittest classes. The program drops/create database and insert/select
stuff from it. I can increase number of rows if necessary. From stand-alone
test I conclude that the overhead is really in unittest framework and not
in a patch itself.

ericvaandering:
Ok, here is a test slice that succeeded: 

https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/577/consoleFull
In the log you will find every command issued and this line:
Total time to test slice 2: 1261
and a report of the tests including their durations: https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/lastCompletedBuild/testReport/

Now to the same slice for a different pull request: 
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/565/consoleFull
Total time to test slice 2: 678
and https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/565/testReport/



So twice as long. But I see other slices that are being killed because they take more than 4 times as long as a successful slice.

vkuznet:
Eric,
I'm trying to reproduce buildBot numbers and I far from it.

Here is what I did:
- I followed instructions from [1] to setup my test bed
- I run two tests DBSBufferFile_t.py and JobAccountant_t.py tests
- I used stock head from WMCore and patched with my PR

The numbers I got quite differ from buildBot, see [2,3].
For DBSBufferFile_t.py the stock head runs without problems but it
run slower than with my PR (which had 1 failure).
The JobAccountant_t.py tests failed miserably in both scenarios
and that's why its time 5 times more (in both cases) then
the one I read from buildBot outputs. Moreover in buildBot
it runs without problem [4,5] while in my setup it failed with many
exceptions. I also see that buildBot shows only 2 tests for
JobAccountant_t.py while I have 15 tests.

Therefore, still there are difference between VM setup from [1]
and what is done on buildBot test machines. I'm willing to help
and debug stuff, but this misbehavior in VM setup is killing all
efforts and I have no clue how to setup things correctly and
be sure that what is run on buildBot is identical to what is written in [1].
I hope Seangchan can help to identify the difference and setup
proper environment.

So far, based on DBSBuffer database test I don't see any slow down
from my PR.

Valentin.

[1] https://github.com/dmwm/WMCore/wiki/Setup-wmcore-unittest
[2] https://gist.github.com/vkuznet/548aba4220fa86b4cd8d
[3] https://gist.github.com/vkuznet/241c9eb1ac3e195d10a4
[4] https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/lastCompletedBuild/testReport/
[5] https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/565/testReport/

ericvaandering:
I have not had a chance to try this in my setup yet, but I can make two comments. 

Rather than trying to run a whole test, can you run just a  single one like this:

nosetests JobCreator_t.py:JobCreatorTest.testE_TestNonProxySplitting

WMCore_t.WMBS_t.Job_t seems to be the worst offender in these tests (biggest difference). Now this machine is probably running a CMSSW IB as well, so I wouldn't trust the reproducibility of any wall time # within a factor of two, but....

Of course buildbot shows 2/15 tests, this is a test run for just a slice (1/10) and since there are repeated failures, it's not able to aggregate the results for your PR. You can choose a different slice too if you like.

One point is that [4] is going to change next time Jenkins tests something.
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/577/testReport/ is the static URL

PerilousApricot:
Hi Val,

I had to fight this exact sort of heisenbug all the damned time when I was actively maintaining the tests (which is a big reason I got frustrated with it). 

Someone would post code, it would break, they would say it worked on their machine, I would say how to replicate it, they would do something different (which worked) then they would consider the matter closed. A few years of this steady accretion of bugs people didn't fix is why the test suite ended up with a lot of the tests that are marked "@integration"

Instead of running what you ran in [2] [3], what do you see if you run the command Eric pointed to:

```python code/setup.py test --buildBotMode=true --reallyDeleteMyDatabaseAfterEveryTest=true --testCertainPath=code/test/python --testTotalSlices=10 --testCurrentSlice=2```

The resulting .xml that get spits out will have timing information you can use to correlate.

-Andrew

**Comment by vkuznet** [comment]
Some unit tests failed due to issue discussed in #6292 PR.

**Comment by vkuznet** [comment]
Yes, what a joy to run these unittests :)

If I run buildBotMode the process just freeze after this:

```
ok
#446 test_simple_json_deflate (Api_t.Tester) ... [28/Sep/2015:15:54:00] ENGINE
Bus STARTING
[28/Sep/2015:15:54:00] ENGINE Started monitor thread '_TimeoutMonitor'.
[28/Sep/2015:15:54:00] ENGINE Serving on 127.0.0.1:8888
[28/Sep/2015:15:54:00] ENGINE Bus STARTED
```

and hangs forever. I have no idea what's doing and I need to manually kill it.
It seems to me it is another issue since I see plenty of tests which passed by.
I don't really observe any long latencies in those tests.

Any other suggestions?
Valentin.

**Comment by ericvaandering** [comment]
Did you try my suggestion and run just a few of the tests that appear to increase the most?

Does it freeze on you in your pull request or on master? Freezing on the PR could cause the outright failures I see, but IIRC I could look in on the logs and see things moving forward, just slowly. 

I've only very rarely seen the tests actually lock up. Obviously if this happened regularly, we could not run our tests.

**Comment by vkuznet** [comment]
On  0, Eric Vaandering <notifications@github.com> wrote:
> Did you try my suggestion and run just a few of the tests that appear to increase the most?

yes, I did, and I posted that running local tests I still on par or even
better. For instance, I run DBSBuffer_t and Job_t tests. For example,
the Job_t tests takes almost 6m in both cases.

> 
> Does it freeze on you in your pull request or on master? Freezing on the PR could cause the outright failures I see, but IIRC I could look in on the logs and see things moving forward, just slowly. 

both, I see similar behavior, but may be I just impatient.

**Comment by ericvaandering** [comment]
I said this "Rather than trying to run a whole test, can you run just a single one like this:

nosetests JobCreator_t.py:JobCreatorTest.testE_TestNonProxySplitting

WMCore_t.WMBS_t.Job_t seems to be the worst offender in these tests (biggest difference). "

which I did not see a response to.

**Comment by vkuznet** [comment]
Eric,
I run 

```
time python test/python/WMCore_t/WMBS_t/Job_t.py
```

in both cases the time is on par, 5m42s/5m48s for master/PR, respectively

When I run 

```
nosetests JobCreator_t.py:JobCreatorTest.testE_TestNonProxySplitting
```

the time is 20.272sec/20.619sec, respectively.

I don't see any large factors, although I adjust format function to reflect
ResultSet class changes.

**Comment by ericvaandering** [comment]
That command is only an example of *how* to run a single test. The particular test was just whatever was in my history. Can you please go back to the comment of https://github.com/dmwm/WMCore/pull/6284#issuecomment-143272945 and find and run a few *specific tests* that show a significant slowdown? Using, of course 
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/577/testReport/  instead of the URL that says "latest"

**Comment by ericvaandering** [comment]
test this please

**Comment by vkuznet** [comment]
Eric,
I just submit another compatibility change, so we'll need to retest it again
with that commit. Sorry that you fired it early, if by any chance you can stop
current test and re-start it again it would be nice.
V.

**Comment by vkuznet** [comment]
The old build logs is gone and I can't compare.
But I finally managed to run DBSBuffer_t.py without any tweaks in DBS code,
i.e. I made my ResultSet class to be compatible with current one which
stores sqlalchemy resultproxy/resultrow objects.

With that change this unit tests shows in master 1m46s, while in my PR 1m44s

I want to see how my last change will affect buildBot outcome.
Valentin.

**Comment by ericvaandering** [comment]
I fired it off to get recent timing stuff (I'm looking at this too). Problem is that the two sets of tests are going to overlap now unless I manage to kill the first one.

**Comment by vkuznet** [comment]
Eric, can we test this again. I'm not sure where it was left yesterday, since it seems test was going on once I submit my changes. My understand that I just need to post "test this please" message to github to fire up test again, is it sufficient?

**Comment by ericvaandering** [comment]
I guess it was. I thought you needed that all by itself in a message, but I guess not.

**Comment by vkuznet** [comment]
Eric,
please help me here to correctly read the results from unit tests.
As far as I understood this time unittest run all of them, no slices.

If I look at particular test from
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-test/702/testReport/WMComponent_t.DBS3Buffer_t.DBSBufferFile_t/DBSBufferFileTest/

and compare it to previous results from
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=2,label=slc6_amd64-cloud-tiny/lastCompletedBuild/testReport/WMComponent_t.DBS3Buffer_t.DBSBufferFile_t/DBSBufferFileTest/

I see that in every tests new PR beats the benchmark. But of course the total
time is larger now since all tests from DBSBufferFileTest were run.

Am I right? Do you see any slowdown this time?
Valentin.

**Comment by ericvaandering** [comment]
The tests are always run in slices, but they are aggregated if everything succeeds. Once Jenkins posts to GH, the aggregation is done and all the sub jobs have succeeded. You can look at the slices if you want, but there is little point.

Also, Jenkins wants to compare your test run with the last test run, but that's totally useless. Instead you need to compare with the baseline and that's what the messages I post do. 

OK, long winded explanation out of the way, compare this 
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-test/702/testReport/WMComponent_t.DBS3Buffer_t.DBSBufferFile_t/DBSBufferFileTest/ with this
https://cmssdt.cern.ch/jenkins/view/DMWM/job/DMWM-WMCore-UnitTests/lastCompletedBuild/testReport/WMComponent_t.DBS3Buffer_t.DBSBufferFile_t/DBSBufferFileTest/ which is aggregated the same way. You'll see the times are comparable.

**Comment by vkuznet** [comment]
Ok, it means that latest commit did fix the issue with timing. I think because
I missed couple of issues the tests were gradually failed which cause cascade
of delays and overall time blow up.

Now, I can walk through failed tests and fix additional issue which I already
observe, e.g.

https://cmssdt.cern.ch/jenkins/job/DMWM-PR-test/702/testReport/junit/WMCore_t.WMBS_t.Fileset_t/FilesetTest/testFileCreate/

**Comment by ericvaandering** [comment]
Here's a slice from the first round of testing that commit which did time out. The log sure looks like a freeze to me: https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/SLICE=9,label=slc6_amd64-cloud-tiny/620/console

**Comment by ericvaandering** [comment]
Is still the monolithic PR referenced in the subject, or is it now isolated to fixing one issue? If you've changed the goal of the commit, you can update the title of the PR.

**Comment by vkuznet** [comment]
It is still monolithic PR. But I did internal code re-factoring again to
support methods expected from resultproxy/resultrow objects used in current
code. And, as test show I still miss some methods.

**Comment by vkuznet** [comment]
I added in latest commit get method which seems to be required by many WMCore
unit tests. This completes ResultRow class to imitate its resultrow buddy from
sqlalchemy. I hope this will fix plenty of failed/regressed unit tests now.

**Comment by amaltaro** [comment]
@ericvaandering it seems this one is in an infinite looping firing Jenkins. Perhaps it's the amount of commits to this PR(?)

**Comment by ericvaandering** [comment]
Not infinite, but it will try four times per commit. Or maybe five. There is a problem with this pull request that makes Jenkins timeout, so that means about two hours per test.

**Comment by amaltaro** [comment]
Ok, the retry is per commit then and not per PR. Valentin, any reason not to squash all these commits?

**Comment by vkuznet** [comment]
Alan,
I'll squash once I figure out all tiny details of this conversion and reason
for jenkins failures/timeouts.
V.

**Comment by vkuznet** [comment]
Eric,
I still need your help to find proper numbers.
From various failed jenkins builds I found that SLICE=1 timeout all the time,
e.g.
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/680/SLICE=1,label=slc6_amd64-cloud-tiny/

But I can't find anywhere breakdown of timing for unit tests.

If I compare this to SLICE=1 from successful build, e.g.
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/669/SLICE=1,label=slc6_amd64-cloud-tiny/testReport/

I can see from test report the numbers each test consumed.

In order to understand timeouts I need to know the corresponding numbers for
failed build above. How I can get them?

V.

**Comment by ericvaandering** [comment]
What I mean is that if the PR is updated with a new commit, it will fire again.

On Oct 7, 2015, at 07:51, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:

Alan,
I'll squash once I figure out all tiny details of this conversion and reason
for jenkins failures/timeouts.
V.

**Comment by ericvaandering** [comment]
The numbers are not available if things timeout because nose is killed before it can generate the report. Can you not reproduce this in your own environment? 

If the tests are really running and not frozen I can let it run longer. Or I can make more slices. Both temporary of course.

**Comment by vkuznet** [comment]
Eric,
I think the tests jenkins use for my PR are different for tests used in
successful build and that explain the difference in time.

For example, here is successful SLICE=1
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/680/SLICE=1,label=slc6_amd64-cloud-tiny/consoleText

and here is consoleText from SLICE=1 of unsuccessful build
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/669/SLICE=1,label=slc6_amd64-cloud-tiny/consoleText

The former has this list of tests
15 16 40 41 46 48 50 56 58 70 71 73 80 102 107 115 117 118 126 142 150 158 161
169 173 175 185 191 225 245 248 251 252 255 258 282 283 300 304 319 321 330 336
355 359 365 369 380 384 385 386 394 397 400 425 426 432 452 454 463 466 473 479
489 522 527 537 548 550 551 557 562 581 596 599 600 625 626 631 640 645 647 656
664 673 715 718 733 745 756 772 775 788 791 792 796 800 804 812 814 817 823 829
833 855 856 876 880 882 894 905 909 912 918 923 928 934 939 940 942 944 948 953
956 966 967 973 980 986 1016 1018 1030 1033 1037 1058 1064 1067 1099

while later has the following:
102 107 115 117 118 126 142 150 158 161 169 173 175 185 191 225 245 248 251 252
255 258 282 283 300 304 319 321 330 336 355 359 365 369 380 384 385 386 394 397
400 425 452 454 463 466 473 479 489 522 527 537 548 550 551 557 562 580 595 598
599 624 625 630 639 644 646 655 1015 1017 1029 1032 1036 1057 1063 1066 1098

As you can see the test ids are quite different [1], e.g. unsuccessful build
had tests 1098, 1066, 1057, 1036, 1032, 1029, etc. which are not present in
successful build.

I don't know how jenkins perform slicing, but from consolteText outputs
I conclude that we don't compare apples to apples.

Can we do the build with persistent slicing ids? Both benchmark and my PR?

V.

[1] I used the following command once I downloaded consoleText files into
680.txt and 669.txt

```
cat 680.txt | grep ^#[0-9] | awk '{print $1}' | sed -e "s,#,,g" | sort -n | awk '{ORS=" "; print $1}'
```

**Comment by ericvaandering** [comment]
@vkuznet I'm in all day training and can't look in detail at this. But IIRC from my looking at this earlier, the slicing is supposed to be done on mod (some hash of the signature of the function), while the numbering is done sequentially. Only if the hash can't be determined is mod(sequence) used. So take a look at the *names* of the tests and see if you are right. We could change how the mod is behaving if needed.

**Comment by ericvaandering** [comment]
This slicing is determined in setup_test.py as I recall.

**Comment by PerilousApricot** [comment]
That's right -- the idea was to try and keep the hashing/slicing as stable as possible even when adding/removing tests. It's not perfect, but it helps keeping everything from jumping to different slices constantly.

**Comment by ericvaandering** [comment]
BTW, if you look at the slices you will see that the order they are run in is something like 1000-end, then 100-999, then 1-99 which explains the discrepancies you see in the first part of your list.

**Comment by ericvaandering** [comment]
And I see that somewhere between 400 and 600 the number diverges. But the names in both sets are the same as far as I can tell.

**Comment by vkuznet** [comment]
Yes, I see now. Next tricky question. I found that this _testLoadFile_ unittest
is hanging (by looking at last visible unittest in failed SLICE).
The question is where it comes from, i.e. from which unittest file.
Just using find unit command there are four unittest files with such unittest
name. The jenkins log does not show the file name where unittest comes from.
Can you fix that?

And, the problem seems to be with locking of MySQL. I run this SLICE=1
in my local environment and I confirmed jenkins output. The SLICE hangs after
_testFilesetClosing_ ERROR tests
and when next _testLoadFile_ test should run.

The interesting part is that now I can't drop wmcore_unittest database in MySQL
and this seems to me a problem. What I did is the following. I saw that my
python setup.py test --buildBotMode=true
--reallyDeleteMyDatabaseAfterEveryTest=true --testCertainPath=test/python
--testTotalSlices=10 --testCurrentSlice=1
hangs. I killed it. And then I tried to do
mysql -u unittestagent -p -S $DBSOCK -e "drop database wmcore_unittest; create
database wmcore_unittest;"

to recreate the database, but it hangs too. I login into mysql and manually
tried to drop database and it hangs too.

So, I need to figure out what cause MySQL to come up with such state that
unit tests can't clean it up because it hangs on dropping database.

V.

**Comment by ericvaandering** [comment]
Quick solution is to add a commit that changes the doc string for the four tests in question

**Comment by ericvaandering** [comment]
I'm a bit confused. `git grep testLoadFile` gives me no hits, but `git grep testLoad` gives me 6 tests by that name and a few more similar. In any case, my immediate suggestion is the same, change the test names and docstrings to be unique. I can see if there is a way to get nose to spit out the module/class/function name, same as what is actually in the report since that's a unique identifier.

**Comment by vkuznet** [comment]
Eric, my fault, I mistyped the failing test is testLoadData and here what we have for it

```
find test/python/ -name "*.py" -print | awk '{print "echo "$1"; grep
testLoadData "$1""}' | /bin/sh
```

gave

```
test/python/WMCore_t/WMBS_t/File_t.py
    def testLoadData(self):
        _testLoadData_
test/python/WMCore_t/WMBS_t/Fileset_t.py
    def testLoadData(self):
        _testLoadData_
test/python/WMCore_t/WMBS_t/JobGroup_t.py
    def testLoadData(self):
        _testLoadData_
...
test/python/WMCore_t/WMBS_t/Subscription_t.py
    def testLoadData(self):
        _testLoadData_
test/python/WMCore_t/WMBS_t/Workflow_t.py
test/python/WMCore_t/WMBS_t/__init__.py
test/python/WMCore_t/WMBS_t/Job_t.py
    def testLoadData(self):
        _testLoadData_
```

So far I don't know which test is failing.

vkuznet:
Found two more issue with this transition change. First, some unit tests relies explicitly on int data type when checking integers. It is not always true since database can return long data type based on value and DB internals. The SQLAlchemy does internal casting from long to int and our code does not care about int/long difference. Second, the new ResultRow class thrown exception when result proxy keys mismatched with ones it was set. Such exception thrown in sub-thread cause a deadlock (thread dies) which block the database. I replaced thrown exception with print statement so far yielding WARNING message and need to look at SQL queries to find out when that happens (possibly when we select out of select). But with this changes I was able to run Fileset_t.py unit tests which hang before in jenkins.

vkuznet:
Wow, hangs are fixed now and I'm only 6 unittest off from success :)

ticoann:
Thanks, Valentin. Don't worry about WorkQueue failure.

vkuznet:
I need to know which tests I should worry, otherwise it is never ending story.
If I correct, now we have all except 6 unit tests pass through the PR changes
meaning that they're working as expected (except those 6 unit tests). Which
ones should I look and try to understand/fix?

ericvaandering:
We have a list of unstable unit tests in WMCore/test/etc

This list will still be checked, but if it fails it will be noted as unstable and will not cause the PR status to go red. If you are convinced a test fits in that category, please make a PR against that file to add it.

ericvaandering:
Valentin, none of the rest of those look familiar to me as things that pop up. You can also verify these things for yourself but using git to switch between master and your branch and testing those tests (maybe more than once in case you suspect they are not repeatable).

vkuznet:
test this please

vkuznet:
I found that I can't reproduce jenkins results in my local build anymore.
One of the recent failed builds is
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-test/829/
and here is one slice which failed
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/746/SLICE=3,label=slc6_amd64-cloud-tiny/console

if I run this slice on my local node
time python setup.py test --buildBotMode=true
--reallyDeleteMyDatabaseAfterEveryTest=true --testCertainPath=test/python
--testTotalSlices=10 --testCurrentSlice=3

it succeed with the following results

FAILED (SKIP=2, errors=3, failures=3)
30.99user 2.78system 17:47.98elapsed 3%CPU

and tests which failed, _testCloseSettingsPerWorkflow_, _FailJobs_, etc. run fine in my local build.

And neither I can reproduce stand-alone test results, e.g.
https://cmssdt.cern.ch/jenkins/job/DMWM-PR-unittests/746/SLICE=5,label=slc6_amd64-cloud-tiny/testReport/junit/WMCore_t.WMBS_t.Fileset_t/FilesetTest/testAddBulkByLFN/

shows (OperationalError) (1065, 'Query was empty'), but I don't have it in my
local build.

Any suggestions?
Valentin.

vkuznet:
test this please

ericvaandering:
This new test will be no different. It's picking up a trial run of python 2.7 as the baseline and it's all fubar. I will retest this one when a real baseline is in place.

ericvaandering:
test this please

vkuznet:
Eric,
am I correct that now we got a clean build?
V.

ericvaandering:
Yeah, I think so, and these two tests should probably be in the unstable list.

Let me look at the timing...

ericvaandering:
Timing seems roughly OK to me too. This took a total of 1 hr 4 min to run and the last successful job was 49 minutes. Could easily just be job scheduling.

amaltaro:
As we just discussed in the meeting, the idea is to merge this PR only after the testbed deployment/validation. We already got the PNN changes in, with a few issues (maybe only one) being addressed and we better isolate major changes, IMO.

ticoann:
Yes we will merge this after production deployment and agent deployment (Nov 3)

vkuznet:
Sure, I'll squash the changes and re-test with this once again to be on a safe
side next week.

vkuznet:
while squashing this PR I found that due to performed unittest I had modification to some test json files. Even though I reverted the changes I was not able to patch them back to a master in a clean way. Therefore I pull out my changes into another PR, https://github.com/dmwm/WMCore/pull/6337 and will check its jenkins. If 6337 will pass tests, it will replace this PR. I hope I didn't miss anything.

ticoann:
new one exists #6337

PR #6286, opened 2015-09-24
Add phedexOnly flag in listFileBlockLocation and listDatasetLocation DBS3Reader APIs.

AndresTanasijczuk:
Motivation: In CRAB we want to get the locations only from PhEDEx or only from DBS (no default fallback to DBS if locations not found in PhEDEx).
This PR should not change anything if phedexOnly = False (the default).

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -557,14 +557,17 @@ class DBS3Reader:
             raise DBSReaderError(msg)
 
 
-    def listFileBlockLocation(self, fileBlockNames, dbsOnly = False):
+    def listFileBlockLocation(self, fileBlockNames, dbsOnly = False, phedexOnly = False):
         """
         _listFileBlockLocation_
 
         Get origin_site_name of a block
 
         """
-        
+        if dbsOnly and phedexOnly:
+            msg = "Error in DBS3Reader: arguments dbsOnly and phedexOnly should not be set together."
+            raise DBSReaderError(msg)
+
         singleBlockName = None
         if isinstance(fileBlockNames, basestring):
             singleBlockName = fileBlockNames
@@ -589,10 +592,11 @@ class DBS3Reader:
                     locations[name] = list(valid_nodes)
 
             fileblockNames = set(fileBlockNames) - set(locations) #get the blocks we did not find information in phedex
-            if fileblockNames:#if we couldnt get data location from PhEDEx, try to look into origin site location from dbs
-                dbsOnly = True
+            if not fileblockNames:
+                phedexOnly = True
+            #if we couldnt get data location from PhEDEx, we can look into origin site location from dbs
 
-        if dbsOnly:
+        if not phedexOnly:
             blocksInfo = {}
             try:
                 for block in fileBlockNames:
@@ -744,13 +748,17 @@ class DBS3Reader:
         pathname = blocks[-1].get('dataset', None)
         return pathname
 
-    def listDatasetLocation(self, datasetName, dbsOnly = False):
+    def listDatasetLocation(self, datasetName, dbsOnly = False, phedexOnly = False):
         """
         _listDatasetLocation_
 
         List the origin SEs where there is at least a block of the given
         dataset.
         """
+        if dbsOnly and phedexOnly:
+            msg = "Error in DBS3Reader: arguments dbsOnly and phedexOnly should not be set together."
+            raise DBSReaderError(msg)
+
         self.checkDatasetPath(datasetName)
 
         locations=set()
@@ -762,13 +770,13 @@ class DBS3Reader:
                 msg += "%s\n" % str(ex)
                 raise Exception(msg)
 
-            if not blocksInfo: # if we couldnt get data location from PhEDEx, try to look into origin site location from dbs
-                dbsOnly = True
-            else:
+            if blocksInfo:
                 for blockSites in blocksInfo.values():
                     locations.update(blockSites)
+                phedexOnly = True
+            # if we couldnt get data location from PhEDEx, we can look into origin site location from dbs
 
-        if dbsOnly:
+        if not phedexOnly:
             try:
                 blocksInfo = self.dbs.listBlockOrigin(dataset = datasetName)
             except dbsClientException as ex:
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -579,7 +579,7 @@ class DBS3Reader:
             try:
                 blocksInfo = self.phedex.getReplicaPhEDExNodesForBlocks(block=fileBlockNames, complete='y')
             except Exception as ex:
-                msg = "Error while getting block location from PhEDEx for block_name=%s)\n" % fileBlockName
+                msg = "Error while getting block location from PhEDEx for block_name=%s)\n" % fileBlockNames
                 msg += "%s\n" % str(ex)
                 raise Exception(msg)
 
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -29,7 +29,7 @@ def remapDBS3Keys(data, stringify = False, **others):
                'block_name': 'BlockName', 'lumi_section_num': 'LumiSectionNumber'}
 
     mapping.update(others)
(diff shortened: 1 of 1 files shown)


src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -577,11 +581,12 @@ def listFileBlockLocation(self, fileBlockName, dbsOnly = False, phedexNodes=Fals
                 msg += "%s\n" % str(ex)
                 raise Exception(msg)
 
-            if not blockInfo or len(blockInfo) != len(blockNames): #if we couldnt get data location from PhEDEx, try to look into origin site location from dbs
-                dbsOnly = True
-                blockNames = set(blockNames) - set(blockInfo) #get the blocks we did not find information in phedex
+            if blockInfo and len(blockInfo) == len(blockNames):
+                phedexOnly = True
```
ericvaandering: It's early for me to try to follow the logic here, but I don't understand why you are changing phedexOnly, an input parameter.

AndresTanasijczuk: I also don't like to change input parameters. I do it because that's what the current version is already doing (is changing the dbsOnly input parameter), so I thought I will keep the same "philosophy". Otherwise I would introduce a new boolean variable (e.g. "allLocationsFound") to signal that all locations were successfully obtained from PhEDEx (or not) and add a check on this variable: `if not phedexonly:` --> `if not phedexonly and not allLocationsFound:`.

mmascher: @ericvaandering , does it ever make sene to mix locations from phedex and the one from dbs? I suggest we also remove that no matter of the value of the input parameters

AndresTanasijczuk: From yesterday's CRAB3 meeting, I understood @ericvaandering was not in favor of changing the default behavior of the code, arguing that merge jobs may rely on it.

ericvaandering: I said it has to be tested thoroughly in WMAgent. @amaltaro may have time to do that.

AndresTanasijczuk: Are we talking about the same thing? I was referring to the question @mmascher made above (if it makes sense to have mixed locations from phedex and dbs). If @ericvaandering is in favor of testing it, then I first have to implement it. Should I? Or should we leave it for another issue.

ericvaandering: I don't think it ever makes sense to mix the locations from DBS and Phedex for a single block, but ANY changes to this code should be tested in WMAgent.

AndresTanasijczuk: The question is not for a single block. The question is when you have a list of blocks in `listFileBlockLocation()`, if it makes sense to return locations from phedex for some blocks and from dbs for other blocks. The current code (with and without my current patch) allows this by default. I.e. it tries to get the locations from phedex first, and if there are some blocks for which it could not get them, it tries to get them from dbs (only for those blocks).

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -557,13 +557,17 @@ def lfnsInBlock(self, fileBlockName):
...
-    def listFileBlockLocation(self, fileBlockName, dbsOnly = False, phedexNodes=False):
+    def listFileBlockLocation(self, fileBlockName, dbsOnly = False, phedexNodes=False, phedexOnly = False):
         """
         _listFileBlockLocation_
 
         Get origin_site_name of a block
 
         """
+        if dbsOnly and phedexOnly:
+            msg = "Error in DBS3Reader: arguments dbsOnly and phedexOnly should not be set togheter."
```
amaltaro: together misspelled here

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -592,8 +597,8 @@ def listFileBlockLocation(self, fileBlockName, dbsOnly = False, phedexNodes=Fals
                 msg += "%s\n" % formatEx3(ex)
                 raise DBSReaderError(msg)
 
-            if not any(blockInfo.values()): # no data location from dbs
-                return list()
+        if not any(blockInfo.values()): # no data location
```
amaltaro: any(blockInfo) would be enough here Andres, we can get rid of the values method.
Otherwise, changes to this method sounds good to me.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -745,15 +750,21 @@ def blockToDatasetPath(self, blockName):
...
-    def listDatasetLocation(self, datasetName, dbsOnly = False):
+    def listDatasetLocation(self, datasetName, dbsOnly = False, phedexOnly = False):
         """
         _listDatasetLocation_
 
         List the origin SEs where there is at least a block of the given
         dataset.
         """
+        if dbsOnly and phedexOnly:
+            msg = "Error in DBS3Reader: arguments dbsOnly and phedexOnly should not be set togheter."
```
amaltaro: also misspelled here.

amaltaro:
Let me finish the testbed validation and then I can run a few quick workflows to test it.
Looking at the code, it seems its functionality will be exactly the same, for WMAgent at least.

src/python/WMCore/Services/DBS/DBS3Reader.py
AndresTanasijczuk: I don't want to do changes that are not necessary for the purpose of my PR (I do it in CRAB a lot, but I try to not do it in WMCore).
But since you tell me... I will remove `.values()`. This is assuming you can never have `blockInfo[blockname] = []`.

AndresTanasijczuk: Actually, the `any()` is also not needed, right? I can just do `if not blockInfo:`.

amaltaro: In principle, yes :) But it would be better to check what is calling this method (a few places) to make sure it would handle an empty list the same way as a dict with empty list values. Otherwise, just keep it as it was before and we take the safe side

AndresTanasijczuk: Let's take the safe side :-) Will not modify it.

AndresTanasijczuk:
I just fixed the misspelled "togheter" -> "together"

amaltaro:
retest this please

ericvaandering:
Needs to be rebased

AndresTanasijczuk:
The new DBS3Reader.py in master branch has a couple of bugs in the listFileBlockLocation() function. Should I fix them? Or will someone first test the current code and fix it and then I come with this PR?

amaltaro:
Andres, you meanthe DBS3Reader.py that just came from the PNN patch? Or the file that was in master before that (yesterday, for instance). If you know what the bug is and how to fix, I think you can go ahead (create a new PR please).

AndresTanasijczuk:
Alan, I mean the new code that just came from the PNN patch. I know what the bugs are. I can make a separate PR.

AndresTanasijczuk:
Mhh, actually what I thought is a bug, it is not. It is actually a change in the function behavior (that is not mentioned in the commit comment https://github.com/dmwm/WMCore/commit/157ad855b7b51a176d33f9c30a147d1803c6582e#diff-374b7a6640288184175057234e393e1cR591). Let me clarify.

Before the PNN patch, the function listFileBlockLocation() with dbsOnly=False was behaving as follows:
1) It gets as argument a list of N blocks for which it has to get their locations.
2) It first tries to get the locations from phedex.
3) If it can not get the locations from phedex for M blocks, it tries to get the origin site from DBS for these M blocks. 
4) The final dictionary of block locations returned by the function can therefore be a mix: some blocks can have locations from phedex, some from DBS.

After the PNN patch the behavior changed as follows:
3) If it can not get the locations from phedex for M blocks, it tries to get the origin site from DBS for the N blocks.
4) The final dictionary of block locations returned by the function contains either all phedex locations or all DBS locations.

Can you confirm this change is intentional? We were actually asking @ericvaandering about if this behavior was ok or should be changed. See for example this comment above from this PR: https://github.com/dmwm/WMCore/pull/6286#discussion_r40799605

amaltaro:
The new behavious sounds ok to me, maybe even better.

AndresTanasijczuk:
Ok, then I am done with the rebase.

ticoann:
test this please

PR #6287, opened 2015-09-25
create the view for MCPileup and DataPileup

### LINKED ISSUE #6083: Create a couch view for pileup datasets
This will be needed for DDM (possible other "customers" as well) to properly manage DataOps space.

ticoann:
Fixes #6083. also update prep id view with StepChain


ticoann:

amaltaro:
Sounds good to me Seangchan. Just one question, instead of emitting null as value, how about we emit doc.RequestName ? This way we have the mapping right away.

amaltaro:
Seangchan, nevermind my comment, the requestName is the doc id itself, so we have the mapping in place already :-)

ticoann:
added reqmgr2 api and unittest as well

PR #6288, opened 2015-09-26
simplify/rewrite dbs3upload

hufnagel:
Current code was not compatible with the Tier0. It somehow worked so far, but never really correctly (mostly causing internal multiplication of work) and with the huge backlog we accumulated due to the few deadlocks and component restarts it completely broke it. I tried to keep the changes minimal. Didn't quite succeed, but this is about as minimal as I can get it.

I needed to remove the workflow parameter from DBSBufferBlock as workflow is not a property of a block. Replaced it with a workflows set, which is filled when files are added to the DBSBufferBlock. Then for block closing I look at whether any of the workflows in a block is completed.

Dataset_algo is also not a property of a block, but it was used as a key for a block cache and to split blocks into smaller groups to process. Removed the whole cache based on dataset_algo as it wasn't needed and replaced the splitting to use datasetpath instead. That actually is a property of a block, but I had to add it to DBSBufferBlock as it wasn't stored.

As I had to touch some queries for these datasetpath and workflow changes anyways, optimized these queries when I saw something obvious. Saw plenty more to be optimized there, but constrained myself...

Reworked the transaction code in DBSBufferUtil, read methods don't need transactions, they just slow things down. Removed them. Left the transactions intact for writes.

Tested this code in a short Tier0 replay, both in live mode and in the mode where DBSUpload component is restarted and has to pick up from open blocks stored in the database. Worked.

Currently running a larger Tier0 replay with Express and Bulk processing and automatic transfers from T0_CH_CERN_Disk to T2_CH_CERN and automatic deletions from T0_CH_CERN_DIsk. If that works too I am going to pull the trigger and deploy this in the production Tier0.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
@@ -32,13 +32,15 @@ import Queue
 import traceback
 import multiprocessing
 
+from WMCore.DAOFactory import DAOFactory
+
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.Services.UUID                  import makeUUID
 from WMCore.WMException                    import WMException
 from WMCore.Algorithms.MiscAlgos           import sortListByKey
 
 from WMComponent.DBS3Buffer.DBSBufferUtil  import DBSBufferUtil
-from WMComponent.DBS3Buffer.DBSBufferBlock import DBSBlock
+from WMComponent.DBS3Buffer.DBSBufferBlock import DBSBufferBlock
 
 from dbs.apis.dbsClient import DbsApi
 
@@ -78,11 +80,11 @@ def createDictionaryFromConfig(configSection):
     return final
 
 
-def uploadWorker(input, results, dbsUrl):
+def uploadWorker(workInput, results, dbsUrl):
     """
     _uploadWorker_
 
-    Put JSONized blocks in the input
+    Put JSONized blocks in the workInput
     Get confirmation in the output
     """
 
@@ -94,7 +96,7 @@ def uploadWorker(input, results, dbsUrl):
     while True:
 
         try:
-            work = input.get()
+            work = workInput.get()
         except (EOFError, IOError):
             crashMessage = "Hit EOF/IO in getting new work\n"
             crashMessage += "Assuming this is a graceful break attempt.\n"
@@ -163,15 +165,19 @@ class DBSUploadPoller(BaseWorkerThread):
 
         # This is slightly dangerous, but DBSUpload depends
         # on DBSInterface anyway
-        self.dbsUrl           = self.config.DBS3Upload.dbsUrl
+        self.dbsUrl = self.config.DBS3Upload.dbsUrl
 
         self.dbsUtil = DBSBufferUtil()
 
+        myThread = threading.currentThread()
+        self.daoFactory = DAOFactory(package = "WMComponent.DBS3Buffer",
+                                     logger = myThread.logger,
+                                     dbinterface = myThread.dbi)
 
         self.pool   = []
         self.blocksToCheck = []
-        self.input  = None
-        self.result = None
+        self.workInput  = None
+        self.workResult = None
         self.nProc  = getattr(self.config.DBS3Upload, 'nProcesses', 4)
         self.wait   = getattr(self.config.DBS3Upload, 'dbsWaitTime', 2)
         self.nTries = getattr(self.config.DBS3Upload, 'dbsNTries', 300)
@@ -189,7 +195,6 @@ class DBSUploadPoller(BaseWorkerThread):
 
         # Setting up any cache objects
         self.blockCache = {}
-        self.dasCache   = {}
 
         self.filesToUpdate = []
 
@@ -211,14 +216,15 @@ class DBSUploadPoller(BaseWorkerThread):
             # Then something already exists.  Continue
             return
 
-        self.input  = multiprocessing.Queue()
-        self.result = multiprocessing.Queue()
+        self.workInput  = multiprocessing.Queue()
+        self.workResult = multiprocessing.Queue()
 
         # Starting up the pool:
         for _ in range(self.nProc):
             p = multiprocessing.Process(target = uploadWorker,
-                                        args = (self.input,
-                                                self.result, self.dbsUrl))
+                                        args = (self.workInput,
+                                                self.workResult,
+                                                self.dbsUrl))
             p.start()
             self.pool.append(p)
 
@@ -242,7 +248,7 @@ class DBSUploadPoller(BaseWorkerThread):
         terminate = False
         for _ in self.pool:
             try:
-                self.input.put('STOP')
+                self.workInput.put('STOP')
             except Exception as ex:
                 # Something very strange happens here
                 # It's like it raises a blank exception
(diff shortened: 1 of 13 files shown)


hufnagel:
Tier0 deployment does not need this to be merged btw. I still would appreciate any comments on the code and the deployment plans. @ticoann @amaltaro @ericvaandering

hufnagel:
Larger replay has problem, seems I need to keep debugging this. Also will address the problems found in the jenkins test. Fixed all unit test problems apart from the last two as I don't understand why it fails there now. If someone has a suggestion...

hufnagel:
retest this

hufnagel:
please retest this

amaltaro:
retest this please

test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
```diff
@@ -245,24 +245,24 @@ def testLoad(self):
...
-               "ERROR: File id is not an integer type."
-        assert type(testFileC["size"]) == int or type(testFileC["size"]) == long, \
-               "ERROR: File size is not an integer type."
-        assert type(testFileC["events"]) == int or type(testFileC["events"]) == long, \
-               "ERROR: File events is not an integer type."
+            "ERROR: File load by ID didn't work"
+
+        assert isinstance(testFileB["id"], int) or isinstance(testFileB["id"], long), \
+            "ERROR: File id is not an integer type." )
+        assert assert isinstance(testFileB["size"], int) or isinstance(testFileB["size"], long), \
```
ericvaandering: Dirk, Jenkins is failing because this file has some syntax error in it, it seems, and it can't parse it. I'm guessing its the "assert assert" here.

If you are re-writing things rather than just modifying them a bit, please use self.assertTrue or self.assertEqual instead of bare asserts which are not part of the unit testing.

hufnagel:
restest this please

hufnagel:
retest this please

test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
hufnagel: I am not rewriting the unit test, the only reason I touched that line was because jenkins told me to.

ericvaandering: It's not telling you to, it's telling you to consider it. You can ignore Jenkins on that one.

test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
```diff
@@ -245,24 +245,24 @@ def testLoad(self):
...
-        assert type(testFileC["id"]) == int or type(testFileC["id"]) == long, \
-               "ERROR: File id is not an integer type."
-        assert type(testFileC["size"]) == int or type(testFileC["size"]) == long, \
-               "ERROR: File size is not an integer type."
-        assert type(testFileC["events"]) == int or type(testFileC["events"]) == long, \
-               "ERROR: File events is not an integer type."
+            "ERROR: File load by ID didn't work"
+
+        assert isinstance(testFileB["id"], int) or isinstance(testFileB["id"], long), \
+            "ERROR: File id is not an integer type." )
```
ericvaandering: The close parenthesis does not belong here. That's causing the syntax error. Removing that means things run.

ericvaandering:
test this please

ticoann:
> Tier0 deployment does not need this to be merged btw. I still would appreciate any comments on the code and the deployment plans. @ticoann @amaltaro @ericvaandering


Hi Dirk, sorry I haven't review the code yet. How was the testing going with large data? As far as for the merging to branch it won't be merged until next week since we need to run agent test as well.
I will try to review the code, it helps that you summarized the changes you made. Thanks for that.

hufnagel:
@tiocann, I made some further changes to clean up transaction handling and hopefully fix the crash that caused the whole mess with the backlog in the first place. That crash is still there btw, we saw it in the Tier0 over the weekend.

I am going to deploy this change in the Tier0 either way, but do you prefer it as part of this pull request or should I make a new one ? If the latter I need to wait until this is merged (there are conflicts between them, I can make a clean new pull request that also works together with this one).

ticoann:
Hi Dirk, you can add that here and part of same request. Thanks,

hufnagel:
Ok, thanks, will do once the replay testing the new code is done.

hufnagel:
Additional changes are to remove transactions within transactions (which some calls to DBSBufferUtil were doing). No idea what the system is actually doing then, but the crashes we saw were in one of these calls. I not only cleaned up that call but also other areas with the same pattern.

hufnagel:
Also remove a bunch of methods from DBSBufferUtil that weren't used anywhere and remove the transaction from a simple SELECT call (not needed for that).

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -415,91 +402,57 @@ def loadFiles(self):
...
+        _checkBlockCompletion_
 
-        Loop all Open blocks and mark them as Pending if they have timed out.
+        Mark Open blocks as Pending if they have timed out or their workflows have completed
         """
         completedWorkflows = self.dbsUtil.getCompletedWorkflows()
         for block in self.blockCache.values():
             if block.status == "Open":
-                if block.workflow in completedWorkflows:
+                if ( block.getTime() > block.getMaxBlockTime() ) or any (key in completedWorkflows for key in block.workflows):
```
amaltaro: Just a minor comment, what about separating the block timeout check from the block/workflow completion check in 2 ifs? It would be much more clear IMO.
Well, another comment, you are not adding these blocks to the blockCache anymore. Did I miss something here?

hufnagel: I combined them because logically its a single condition. If (block either has waited over the max wait time or if any of the workflows producing files in the blocks has completed) then close block.

And why would I need to assign the block to the block cache ? Its already in there, I am iterating over the blocks in the block cache !

amaltaro: It's arguable, one thing is a block timeout, the other is a workflow completed condition, anyways it just sounds cleaner than that almost ugly line.

I was misguided by the previous code that was making the blockCache redundant assignment.

hufnagel: Mine is cleaner because that's what the logic already is. The alternative would be:

if "blocks is too old":
    closeBlock()
elif "at least one workflow is finished in block":
    closeBlock()

You can't just have two independent ifs because these are not independent, you are only supposed to close the block once.

amaltaro: that's what I meant with "2 ifs" (if elif), sorry for not being clear :(

hufnagel: Still like mine better :-). Closer to the programming logic (evaluate the full logic condition first, then call what you want to call). But that comes down to personal preference to some degree...

ericvaandering:
Needs to be rebased

hufnagel:
So this is rebased and the last jenkins test worked. @ticoann , can you review this please ?

hufnagel:
@amaltaro , please include this in your testing.

amaltaro:
Ok. I guess @ticoann and I will cut a new tag today after we merge all these stuff.

Issue #6273, opened 2015-09-22
Job deleting files after a successful stage out


amaltaro:
As reported in this HN:
https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/2536.html

it seems the stage out plugin is firing an srmRm for files that failed the
first stage out attempt and succeeded in the second one (still in the same
job).

It happened at T2_IT_Pisa that uses srmv2 protocol. More info in the HN itself.
debugging...

amaltaro:
For the record, from the list of tickets I got from the transfer team, these are the sites and their local stage-out list:
T2_US_Vanderbilt using "vandy" command, so VandyImpl
T2_FI_HIP using srmv2 command, so SRMV2Impl
and all the others (T2_IT_PISA, T1_IT_CNAF, T2_US_Caltech, T2_EE_Estonia, T2_BE_IIHE, T2_DE_RWTH, T2_FR_GRIF_LLR) using srmv2-lcg, so LCGImpl

ericvaandering:
Does this have anything to do with this? https://ggus.eu/index.php?mode=ticket_info&ticket_id=118307

amaltaro:
The transfer team claims to still see files missing from time to time, mainly at KIT. I asked them to get in touch with us in the last CompOps, if we don't hear anything by beginning of March, we shall close this annoying issue.

amaltaro:
After more investigation and debugging with the KIT folks, we came to the conclusion that the Alarm signal is "responsible" for these missing files, since it leaves behind a subprocess still trying to stage out:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py#L219

 and we cannot kill that subprocess because - at this point - we have no idea what command/call we are running, we didn't build the stage out command nor we know which protocol to use.

However, if we move the Alarm handling to this code (2 layers below the previous one):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/StageOutImpl.py#L173

then we could use check_output from subprocess, get the PID and kill it in case the Alarm is raised.

A second option would be removing the Alarm handling from the StageOut.py, we already have a timeout option for srm and lcg (probably for gfal as well), thus we don't need to have two layers of timeout for the stage out code.

hufnagel:
[...]

PR #6291, opened 2015-09-28
Fix stage out deleting unwanted files

amaltaro:
Attempt to fix (or mitigate) #6273 

I could not find any flaw in the code, then as discussed with Seangchan, I'm removing as much as possible file deletion commands. In summary:
- files cleanup (when failures happen) will be done mainly at top level (cleanSuccessfulStageOuts call)
- since other stage out implementation don't care for file size and checksum, I also removed it from LCGImp such that we rely solely on the exit status of lcg-cp command.
- removed deferDirectoryCreation since it was always returning False
- trying to place all wmagent exit codes in a single place

I discussed this with @hufnagel today and he does not like the removal of additional checks. However there is not much we can do here since logs are not available after the workflow is archived. Moreover, I don't expect lcg-cp returning exit status 0 when transfer faces problems (!= file size or checksum).

Pylint cleanup (and my ALAN logs) were left for a bit later. For now only sharing this code for comments/review.

### DIFF
--- a/src/python/WMCore/Storage/Backends/LCGImpl.py
+++ b/src/python/WMCore/Storage/Backends/LCGImpl.py
@@ -124,12 +124,10 @@ class LCGImpl(StageOutImpl):
             echo "lcg-cp exit status: $EXIT_STATUS"
             if [[ $EXIT_STATUS != 0 ]]; then
                echo "Non-zero lcg-cp Exit status!!!"
-               echo "Cleaning up failed file:"
-                %s
-               exit 60311
+               exit $EXIT_STATUS
             fi
 
-            """ % self.createRemoveFileCommand(targetPFN)
+            """
 
         if self.stageIn:
             remotePFN, localPFN = sourcePFN, targetPFN.replace("file:", "", 1)
@@ -139,8 +137,6 @@ class LCGImpl(StageOutImpl):
         result += "FILE_SIZE=`stat -c %s"
         result += " %s `\n" % localPFN
         result += "echo \"Local File Size is: $FILE_SIZE\"\n"
-
-        removeCommand = self.createRemoveFileCommand(targetPFN)
 
         useChecksum = (checksums != None and 'adler32' in checksums and not self.stageIn)
 
@@ -153,14 +149,12 @@ class LCGImpl(StageOutImpl):
                         exit 0
                     else
                         echo "Error: Checksum Mismatch between local and SE"
-                        echo "Cleaning up failed file"
-                        %s
                         exit 60311
                     fi
                 else
                     exit 0
                 fi
-            """ % (localAdler32, removeCommand)
+            """ % localAdler32
         else:
             checksumCommand = "exit 0"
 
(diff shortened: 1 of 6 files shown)


hufnagel:
Ok, I went through the original discussion of the problem and looked at your changes. I don't think this has anything to do with the WMAgent code, otherwise we would have seen this with other sites. We use immediate removal of files after stageout failures for CERN for instance as xrdcp absolutely needs it. Tier0 never has seen a problem with a file being staged out and then deleted and we create and stageout an absolute ENORMOUS number of files.

I also have nothing against removing some protections if we can show that they indeed are the problem here. Making changes without understanding what is going on most likely means we don't fix the problem and remove protections that were there for a reason.

Thing is, as lcg-cp happily overwrites existing files, removing the removeCommand calls on size and checksum comparison failures makes sense, PROVIDED the last stageout retry that fails cleans up after itself somehow. If it doesn't we leave dark data at the site.

Also, why would you remove the size and checksum comparison ? We don't need to delete on them, but we should fail on them. There are reasons we in general don't trust exit codes. Also has nothing to do with this issue really.

hufnagel:
I have one question, is lcg-del (which seems to be the delete command used here), run synchronously or asynchronously (ie. does it delete right away or does it queue the file for deletion somewhere ) ?

If the latter it could explain the problem we are seeing...

amaltaro:
These commands seem to be synchronous, since we have the srm-timeout option. BTW, I'll add a 10min timeout for lcg-del commands. I couldn't find anything more direct than that.

> PROVIDED the last stageout retry that fails cleans up after itself somehow. If it doesn't we leave dark data at the site.

if the last retry fails, then a cleanup is in place (was not changed)
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py#L217

> Also, why would you remove the size and checksum comparison ? We don't need to delete on them, but we should fail on them. There are reasons we in general don't trust exit codes. Also has nothing to do with this issue really.

I could add them back and only exit in case of failures, no deletion. I simply removed it because others implementation don't really care.

hufnagel:
I think if it is working we should leave it in. Someone spend some time making it work and we should not just remove it without a good reason. Some implementation use them and in general these checks are useful.

amaltaro:
Ok, I'll investigate it further this week then. But this "Someone spend some time making it work and we should not just remove it without a good reason." is not really a good argument to keep code around as well :). We know there are code in WMCore that is no longer used (or that was never used), or that has not been used during the last year or so. Thus removing them seems the most reasonable thing to do IMO, otherwise it only overcomplicates the code, makes it harder to maintain and etc (v2 of stage out plugins are a good example).

hufnagel:
Size and checksum comparison for stageout is in general a really good idea. We had cases where the exit code of the copy program could simply not be trusted. In addition, comparing the checksum can catch errors that the client exit code cannot tell you. Corruption of data in-flight should be rare, but cannot be detected unless you compare what went in (checksum you know the file is supposed to have) to what came out (checksum of the data as written on the other end).

amaltaro:
Updated the code with the file size and checksum checks again, but this time not deleting anything in place in case it fails.
Also added srm-timeout to ls and del commands, such that we remove them from srm queue in case it takes too long.

Moreover, this Alarm
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py#L214
can brings us problems in case it is triggered and we don't cancel the srm requests (also leaving garbage  at the site). We could keep only the srm timeouts here, but this class is used by all plugins and then probably not a good idea.

juztas:
@amaltaro Just quick test with 2 sites xmls, and both failed to do fallback stageout. If 'srmv2' is used as fallback, it requires to create directories before executing cp and now this code is not calling directory creation.

amaltaro:
Thanks for checking this out Justas. I'd probably miss the fallback test. I'll check the rest tomorrow

ericvaandering:
test this please

amaltaro:
@ticoann do not merge please. It still has my loggings and I'd prefer discussing it in tomorrow's meeting.

ericvaandering:
Needs to be rebased.

amaltaro:
@ticoann let's hold on this one for a bit more please. Anyways it's on the agent side only.

amaltaro:
Closing it in favor of #6695

PR #6292, opened 2015-09-28
Fix lumis list to store unique values

vkuznet:
While testing #6284 I came across several unit tests which failed due to improper logic of Run DataStructure. The Run class stores lumis in a list. But comparison function does not account for identical lumi sections. Therefore unittests such as DBSBufferFile_t, Job_t had several failures when I switched to usage of generators in WMCore. I think that caused because nobody check uniqueness of results from SQL (or SQL didnt' require unique values). When generators are in place they don't care about results and if no one specify unique SQL clause explicitly we may end-up with duplicates. The proposed changes takes care of internal data structure without any outside modification. I added property lumis, which always return unique list, while internal lumis are still stored in list. Once this fix was applied aforementioned unit tests with generator usage in WMCore were completed ok.

### DIFF
--- a/src/python/WMCore/FwkJobReport/XMLParser.py
+++ b/src/python/WMCore/FwkJobReport/XMLParser.py
@@ -233,7 +233,7 @@ def runHandler():
                           if "ID" in lumi.attrs]
 
                 runInfo = Run(runNumber = runId)
-                runInfo.lumis.extend(lumis)
+                runInfo._lumis.extend(lumis)
 
                 Report.addRunInfoToFile(fileSection, runInfo)
 
--- a/src/python/WMCore/DataStructs/Mask.py
+++ b/src/python/WMCore/DataStructs/Mask.py
@@ -107,10 +107,10 @@ class Mask(dict):
 
         Add a run object
         """
-        run.lumis.sort()
-        firstLumi = run.lumis[0]
-        lastLumi  = run.lumis[0]
-        for lumi in run.lumis:
+        lumis = sorted(list(run.lumis))
+        firstLumi = lumis[0]
+        lastLumi  = lumis[0]
+        for lumi in lumis:
             if lumi <= lastLumi + 1:
                 lastLumi = lumi
             else:
@@ -144,7 +144,7 @@ class Mask(dict):
               with duplicate lumis.
         """
 
-        if not type(lumis) == list:
+        if not isinstance(lumis, list):
             lumis = list(lumis)
 
         if not run in self['runAndLumis'].keys():
--- a/src/python/WMCore/DataStructs/Run.py
+++ b/src/python/WMCore/DataStructs/Run.py
@@ -21,8 +21,18 @@ class Run(WMObject):
     def __init__(self, runNumber = None, *newLumis):
         WMObject.__init__(self)
         self.run = runNumber
-        self.lumis = []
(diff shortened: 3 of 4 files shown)


ericvaandering:
Clever. I did not know about @property.

vkuznet:
My only worry is a size of lumi list. If it is less than 1k such tweak will
work just fine, if we anticipate larger size we may need to re-factor code to use set
instead of list. But that can affect other APIs which may rely on list
data-structure, e.g. someone can create obj=Run() and do obj.lumis.append(1)
If lumis will be stored as set data-structure such code will break.

ericvaandering:
Lumis are 23s long. 1000 corresponds to a run of 16 hours. So that is close to the maximum for data (2k may be possible).

For montecarlo, it's the total # of jobs in a workflow, so 1k may be too few.

vkuznet:
In that case we should be fine with such change.

ericvaandering:
What about all the unit tests that blew up here?

vkuznet:
Some of them are just silly, e.g.

```
WMCore_t.DataStructs_t.Mask_t.MaskTest.testFilter
AssertionError: [9, 2] != [2, 9]
```

which basically means that no ones carry about sorting when comparing the list
objects. Other I don't understand and need to look at. I'm still battling with
gen/transaction/connection PR.

vkuznet:
I fixed one place in Run which I missed and I adjusted Mask to properly handle
lumis as property (sort should be done locally).

ericvaandering:
Ok, so you return a sorted list which is as it should be. The unit tests then should be modified to expect one.

vkuznet:
I found it empirically, since I don't know initial logic about lumis or unit
test expectations. From the code it is not clear should lumis be a sorted list
or not. The unit test assumes it is, therefore I added extra layer (even though
it is not strictly required) to avoid additional failures.

vkuznet:
I think I found last missing parts which deal with lumis property in Run class. I was able to get fixes for some of the reported unit tests. But I still had issue with job splitting. Once tests are finished I'll need a help from expert on that issue (@ticoann).

src/python/WMCore/FwkJobReport/XMLParser.py
```diff
@@ -233,7 +233,7 @@ def runHandler():
                           if "ID" in lumi.attrs]
 
                 runInfo = Run(runNumber = runId)
-                runInfo.lumis.extend(lumis)
+                runInfo._lumis.extend(lumis)
```
ericvaandering: Help me understand this. I would have thought that with the @property decorator would allow you to just lumis?

vkuznet: Yes, that a problem. The property allows to access internal member data, but it
does not mean that it allows applying methods of that member data type. So our
internal member data is self._lumis which is a list type. The lumis property
points to it, but does not allow to use list APIs (e.g. append, extend, etc).
What should be properly done is the following. We need to add method
extendLumis to Run class and use it instead of accessing internal data member.
But I don't know how many places in WMCore "relies" on direct internal member
data access and moreover using its type APIs.

ticoann:
test this please

amaltaro:
test this please

amaltaro:
Changes from 2015, with lack of activity and likely no longer relevant.

PR #6293, opened 2015-09-29
Fix typo validate_doc

amaltaro:
Sorry for being so stupid and making this mistaken :-(


ticoann:
Sorry, I should have caught that when I review. Don't worry I need to make a new tag anyway. There seems to a problem with t0_wmstats. Thanks.

PR #6296, opened 2015-09-30
add block level subscription to PhEDEx from wmagent blocks (DO NOT MERGE)

### LINKED ISSUE #5945: Inject Blocks in Phedex with group=DataOps
If this https://github.com/dmwm/PHEDEX/issues/989 get solved, we should get the blocks injected to phedex under dataops group.

ticoann:
fixes #5945 

This is initial patch for the review.
As far as I know about the issue, we will make block level subscription when block is injected. (After discussing with Nicolo, I changed to make subscription when block is closed so it can have less duplicate subscription called)

If one of the call fails (either injecting/closing the block or subscribing the block) both will be retried in next cycle)

Here are things still need to be clarified and need to be done. 
1. Should this (block level subscription) be optional for Tier0?
2. Still need to add unittest and test it.
3. When PNN patch is merged this need to be changed to get the correct location

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
@@ -14,6 +14,7 @@ from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 
 from WMCore.Services.PhEDEx import XMLDrop
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
+from WMCore.Services.PhEDEx.DataStructs.SubscriptionList import PhEDExSubscription
 
 from WMCore.DAOFactory import DAOFactory
 
@@ -79,7 +80,7 @@ class PhEDExInjectorPoller(BaseWorkerThread):
 
         self.getUninjected = daofactory(classname = "GetUninjectedFiles")
         self.getMigrated = daofactory(classname = "GetMigratedBlocks")
-
+        
         daofactory = DAOFactory(package = "WMComponent.DBS3Buffer",
                                 logger = self.logger,
                                 dbinterface = myThread.dbi)
@@ -248,7 +249,35 @@ class PhEDExInjectorPoller(BaseWorkerThread):
             myThread.transaction.commit()
 
         return
+    
+    def subscribeBlocksInDataset(self, datasetPaths, node, xmlData):
+        """
+        _subscribeBlocksInDataset_
+        :arg list datasetPaths: list of datasetPath
+        :arg str node: site location
+        :arg str xmlData: xml formatted data
+        """
+        
+        subs = []
 
+        # Create the subscription objects and add them to the list
+        # The list takes care of the sorting internally 
+        for datasetPath in datasetPaths:
+            phedexSub = PhEDExSubscription(datasetPath, node, self.group, level = 'block',
+                                           move='n', custodial='n', static='y',
+                                           request_only='n', no_mail='y')
+            # Add it to the list
+            subs.append(phedexSub)
+
+        for subscription in subs:
+            
+            msg = "Subscribing: %s to %s, with options: " % (subscription.getDatasetPaths(), subscription.getNodes())
+            msg += "Move: %s, Custodial: %s, Request Only: %s, Level: %s,  No mail: %s" % (
+                                subscription.move, subscription.custodial, subscription.request_only, 
+                                subscription.level, subscription.no_mail)
+            logging.debug(msg)
+            self.phedex.subscribe(subscription, xmlData)
+            
     def closeBlocks(self):
         """
         _closeBlocks_
@@ -286,6 +315,8 @@ class PhEDExInjectorPoller(BaseWorkerThread):
             myThread.transaction.begin()
             try:
                 xmlData = self.createInjectionSpec(migratedBlocks[siteName])
+                # inject and subscribe when closing the block
+                subRes = self.subscribeBlocksInDataset(migratedBlocks[siteName].keys(), location, xmlData)
                 injectRes = self.phedex.injectBlocks(location, xmlData)
                 logging.info("Block closing result: %s" % injectRes)
             except HTTPException as ex:
--- a/src/python/WMCore/Services/PhEDEx/DataStructs/SubscriptionList.py
+++ b/src/python/WMCore/Services/PhEDEx/DataStructs/SubscriptionList.py
@@ -24,8 +24,9 @@ class PhEDExSubscription(object):
     PhEDEx subscription data service
     """
     def __init__(self, datasetPathList, nodeList, group,
-                 level = 'dataset', priority = 'normal', move = 'n', static = 'n',
-                 custodial = 'n', request_only = 'y', blocks = None, subscriptionId = -1):
+                 level='dataset', priority='normal', move='n', static='n',
+                 custodial='n', request_only ='y', no_mail='n', 
+                 blocks=None, subscriptionId = -1):
         """
         Initialize PhEDEx subscription with default value
         """
@@ -44,6 +45,7 @@ class PhEDExSubscription(object):
         self.group = group
         self.custodial = custodial.lower()
         self.request_only = request_only.lower()
+        self.no_mail = no_mail.lower()
         self.requesterID = None
         self.status = "New"
 
--- a/src/python/WMCore/Services/PhEDEx/PhEDEx.py
+++ b/src/python/WMCore/Services/PhEDEx/PhEDEx.py
@@ -103,6 +103,7 @@ class PhEDEx(Service):
         args['custodial'] = subscription.custodial
         args['group'] = subscription.group
         args['request_only'] = subscription.request_only
+        args['no_mail'] = subscription.no_mail
 
         return self._getResult(callname, args = args, verb = "POST")
 


ticoann:
@amaltaro, Alan could you review the logic? This still need to be tested.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -85,6 +86,7 @@ def setup(self, parameters):
                                 dbinterface = myThread.dbi)
         self.setStatus = daofactory(classname = "DBSBufferFiles.SetPhEDExStatus")
         self.setBlockClosed = daofactory(classname = "SetBlockClosed")
+        self.getSubscribionInfoByDataset = daofactory(classname = "GetSubscribionInfoByDataset")
```
amaltaro: I think you want to use the PhEDEx DAOFactory (above), no? And there is a typo in the class name

amaltaro:
Hmm, looking at the query I understand it will subscribe data (blocks) only when subscription information is given, am I right? And then, the block subscription will only be made against the sites provided in the subscription info during assignment, right?

If the answer is yes for both of my questions above, then I don't understand this PR. Otherwise, my understand from that issue is that ASA a block is injected into PhEDEx against a site XXX, then the agent should also make a block subscription at the same site, with the goal (only goal) to have a better phedex group accounting while datasets are being produced.

ticoann:
> If the answer is yes for both of my questions above, then I don't understand this PR. Otherwise, my understand from that issue is that ASA a block is injected into PhEDEx against a site XXX, then the agent should also make a block subscription at the same site, with the goal (only goal) to have a better phedex group accounting while datasets are being produced.

Yes, you are completely right. I don't know what I was thinking. I will change that. So the subscription is always "non custodial" and "auto approved", right since the data is already there, right?

amaltaro:
Always non-custodial, unless it was produced in the custodial site (but I think it's impossible, only MSS can be custodial).
and auto-approved without sending notification (request-only?).
This fix should subscribe only blocks IMO and let the final dataset subscription to be handled normally.

ticoann:
Thank you very much Alan for the quick review. I updated the patch. Could you check again?

hufnagel:
If I understand correctly that this subscribes blocks to the location where they are injected, then yes, for the Tier0 this needs to be optional. In principle I could adjust the auto-delete code to work around it, but this would potentially complicate things.

Timing is a general problem here. We have to make sure this doesn't interfere with having an automatic move subscription as part of the workflow for instance. What happens if we have a dataset level move subscription in place and then make a block subscription for a block in that dataset we just injected ? Is PhEDEx going to still move the block or not ?

Also, what happens if the block injection succeeds and the block subscription doesn't ? Do we retry the block subscription ? How does that affect all the questions about timing and other move subscriptions ?

This needs some thought...

ticoann:
Thanks Dirk for the feedback.

> If I understand correctly that this subscribes blocks to the location where they are injected, then yes, for the Tier0 this needs to be optional. In principle I could adjust the auto-delete code to work around it, but this would potentially complicate things.
>

I can make it optional for Tier0. Or do you want me to completely disable for Tier0?

> Timing is a general problem here. We have to make sure this doesn't interfere with having an automatic move subscription as part of the workflow for instance. What happens if we have a dataset level move subscription in place and then make a block subscription for a block in that dataset we just injected ? Is PhEDEx going to still move the block or not ?

If there is automatic move subscription, we shouldn't make block subscription at all. (I think I should check that condition and don't make subscription for those requests). What do you think?
> 
> Also, what happens if the block injection succeeds and the block subscription doesn't ? Do we retry the block subscription ? How does that affect all the questions about timing and other move subscriptions ?
> 

Actually, it will make always the subscription call first, if it fails, it will be retried and in case of success, injection call will be tried. If injection call fails, it will retry both calls. 
With the move subscription concern and timing, see above comment.

> This needs some thought...

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -248,7 +249,35 @@ def injectFiles(self):
...
+        """
+        
+        subs = []
+
+        # Create the subscription objects and add them to the list
+        # The list takes care of the sorting internally 
+        for datasetPath in datasetPaths:
+            phedexSub = PhEDExSubscription(datasetPath, node, self.group, level = 'block',
+                                           move='n', custodial='n', static='y',
+                                           request_only='n', no_mail='y')
```
ticoann: @nikmagini, Hi Nicolo, what option should we use for requset_only?  This is block level subscription we talked.
Basically  the blocks will be already in the site just we want to link the group to data.

amaltaro:
I think Dirk's points express why I was reluctant of handling these things on the WMAgent side...

Adding on Seangchan and Dirk's comments, the only case we would not make a block subscription is when there is an automatic dataset MOVE subscription, which are basically extinct nowadays.

So, even if there is an automatic subscription, we should be subscribing blocks produced against their origin site.

> Actually, it will make always the subscription call first, if it fails, it will be retried and in case of success, injection call will be tried. If injection call fails, it will retry both calls. 

Don't we need to first inject data first? You probably mixed them up here :)

hufnagel:
> > Timing is a general problem here. We have to make sure this doesn't interfere with having an automatic move subscription as part of the workflow for instance. What happens if we have a dataset level move subscription in place and then make a block subscription for a block in that dataset we just injected ? Is PhEDEx going to still move the block or not ?

> If there is automatic move subscription, we shouldn't make block subscription at all. (I think I should check that condition and don't make subscription for those requests). What do you think?

Hm, but then you have the same issue as before, until the move subscription kicked in that block will be present at the injection site and it won't have any owner. OTOH, you can make a block level subscription to the injection site either as that block won't be moved then (at least as far as I understand move subscriptions).

Not making the block subscription in the presence of an active move subscription is an option, the time between the block showing up at the injection site and it being moved should be small. Still, you set yourself up for race condition problems. I think we need to make sure DDM can't screw this up from their end too.

hufnagel:
> > If I understand correctly that this subscribes blocks to the location where they are injected, then yes, for the Tier0 this needs to be optional. In principle I could adjust the auto-delete code to work around it, but this would potentially complicate things.

> I can make it optional for Tier0. Or do you want me to completely disable for Tier0?

Some settings that I can control from the Tier0 side would be best. Even better would be if this is configurable per output somehow. Can we somehow configure this as part of the normal SubscriptionInfo object ? Or how were you planning to do this ?

hufnagel:
> > Also, what happens if the block injection succeeds and the block subscription doesn't ? Do we retry the block subscription ? How does that affect all the questions about timing and other move subscriptions ?

> Actually, it will make always the subscription call first, if it fails, it will be retried and in case of success, injection call will be tried. If injection call fails, it will retry both calls.
With the move subscription concern and timing, see above comment.

As Alan already pointed out, you can subscribe a block within it first being injected into PhEDEx :-)

The problem is that you can't tie these together, as they are two separate PhEDEx data service calls. If the block injection works, the block is in PhEDEx. All you can do then is to retry the block subscription until it succeeds, but that would be blocking if there is a problem. Which means in practice this needs to be implemented asynchronously. Which means you need to start worrying about what happens if one works and the other doesn't.

As I said, this needs some though. What Jean-Roch asked us to do is in practice impossible. You always have to have protections against the case of a block being injected but not subscribed, And if we have to have these protections anyways, why are we doing this at all ???

hufnagel:
So summary, even if we do this, this doesn't mean there won't be unsubscribed blocks created by production that DDM under not circumstances is allowed to touch.

ticoann:
> As Alan already pointed out, you can't subscribe a block without it first being injected into PhEDEx :-)

Actually when subscribe call is made, there should be already blocks injected. Actual initial injection happens https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L212

The following injection call after subscription call is should be just closing the blocks.  (I will double check)

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
nikmagini: Hi Seangchan,

I think you want request_only='n' because this will auto-approve the subscription.
no_mail='y' is to suppress the mail notifications.

Cheers
N

ticoann:
> Some settings that I can control from the Tier0 side would be best. Even better would be if this is configurable per output somehow. Can we somehow configure this as part of the normal SubscriptionInfo object ? Or how were you planning to do this

I was planning to do by just agent not per each request. Assuming this should be done by all the requests. But it has be done by requests we can add in normal subscription Info object - need to change the schema to hold the information.

ticoann:
> Hm, but then you have the same issue as before, until the move subscription kicked in that block will be present at the injection site and it won't have any owner. OTOH, you can make a block level subscription to the injection site either as that block won't be moved then (at least as far as I understand move subscriptions).
> 
> Not making the block subscription in the presence of an active move subscription is an option, the time between the block showing up at the injection site and it being moved should be small. Still, you set yourself up for race condition problems. I think we need to make sure DDM can't screw this up from their end too.

Dirk, Alan, I probably don't understand the core of this issue (Why they need owners for each blocks). Would you mind explaining it or point me the link discussion occured? 
I thought owner ship is only needed when data is present on the site. So if the move subscription is made to other site (move subscription means "move and delete", right?), owner ship for deleted data is gone as well.
What does DDM do with that information? Do they need to do something with transient data (which move subscription is made initially so data will be in the created site temporary?

nikmagini:
Hi all,

concerning Move subscriptions:

I agree with the proposal to not subscribe the block to the source when you have a Move subscription to the destination.

Otherwise with the current implementation of Move in PhEDEx the final result for the source block (deletion, staying, or retransfer) would depend on the order of the approvals of the source vs destination requests.

Cheers
N.

ticoann:
After talking to Alan, (He explained this is need for accounting - removing the data without owners)
So I would to following.
1. If there is location contains MSS, won't make the block subscription to that. - I don't think there is any case we do that unless we had big screw up in the code but for the safe side we can add this checking).
2. If there is move subscription made (check only from request spec) for the given dataset, agent won't make the block subscription for the dataset. (As Dirk said there is race condition if accounting happens while data is moving but eventually accounting will be correct - once all the data is moved).

@hufnagel, Dirk is that OK with you, or do you still think there will be problem?

ticoann:
As far as making block subscription optional, we can  make per agent configuration or per request configuration. Dirk I didn't quite catch it from last release meeting. Did you want this per request configuration?  Or what you meant was only to owner ship (currently it is configured per agent not per request)

hufnagel:
If we make it per request, we can more easily skip making these subscriptions for moves. We know what we are moving after all if we make the move subscriptions ourselves. Could simplify this a lot because we don't have to check the PhEDEx API for move requests.

And before someone brings it up, if someone outside WMAgent makes a move subscriptions for a dataset that WMAgent is actively working on, we are in deep trouble anyways and nothing will work correctly.

PS: Yes, I know the statement is that we don't make move subscriptions from the WMAgent anymore, but this is per policy. Policies can change.

ticoann:
Hi Dirk, Actually, I meant we only consider the move subscription made (more precisely, intended to be made) by agent. (I wasn't planning to check subscription directly made in PhEDEx).
So we just check the wmbs table whether specific workflow has a move subscription, then we don't make the block subscription. (Assuming move subscription made independently to PhEDEx, It is only made when all the injection is finished. - I guess that is not realistic assumption).
If  you need to handle the move subscription made directly to PhEDEx (not from agent), I guess we can't avoid race condition.

The only possible solution I can think of in that case. PhEDEx need to implement specific flag for subscribe api. (If flag is set it creates the subscription but not actually try to move or copy the data over. The block subscription created by agent always set that flag since the data is already there and when a move subscription is created it won't try to move back the data to original location).

hufnagel:
Hm, when you setup a workflow/request in the agent, part of the information is subscription information for the output. I was thinking since at that time you know whether you are moving the data or not, you write the "local" subscription record at the same time (or not in the presence of move subscription). It would be easy for me to also at the same time provide an override flag that triggers the same behavior from the Tier0 (if the subscription type is move or the "noLocalSubscrption" flag is True => don't create "local" subscription record).

From the technical side this would mean that you have to write some subscription record in wmbs that triggers "local" subscription of blocks. Not sure if we can overload the subscription table this way, the problem here is the site row which currently expects a real site (and that isn't know until the data exists for these "local" subscriptions).

The alternative is to always trigger this "local" subscription automatically and make a dedicated check for move subscription records inside wmagent before. I just see that being much harder to override for the Tier0 as we don't really have a hook into that place.

I would not touch anything in PhEDEx for this. Move subscriptions aren't really made outside WMAgent anyways (and hardly inside at the moment).

ticoann:
> The alternative is to always trigger this "local" subscription automatically and make a dedicated check for move subscription records inside wmbs before. I just see that being much harder to override for the Tier0 as we don't really have a hook into that place.

That is what I had in mind initially. Basically put the check right before following.
https://github.com/dmwm/WMCore/pull/6296/files#diff-73e9dd1a91792c7023eeec392d2e28b3R319

From following tables,
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/DBS3Buffer/MySQL/Create.py#L28
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/DBS3Buffer/MySQL/Create.py#L45

We just need to create new DAO getting move subscription status from above tables for given dataset. (If it is to many db access we can have cache - but I don't think it is necessary).

I was just wondering whether this wouldn't work for tier0. Does tier0 works differntly?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -286,6 +315,8 @@ def closeBlocks(self):
             myThread.transaction.begin()
             try:
                 xmlData = self.createInjectionSpec(migratedBlocks[siteName])
+                # inject and subscribe when closing the block
+                subRes = self.subscribeBlocksInDataset(migratedBlocks[siteName].keys(), location, xmlData)
```
hufnagel: I would not make two remote API calls in the same try/except block. Not without thinking about what would happen if the first call suceeds and the first fails. First here is block subscriptions, is it safe to retry that even if it already succeeded ? Even if it is, two try/except blocks would be better strictly to know in the execption handler which call did in fact fail.

PR #6302, opened 2015-10-01
_|dummy used in the end for dummy variables

juztas:


standards/.pylintrc
```diff
@@ -88,7 +88,7 @@ function-rgx=([a-z][a-zA-Z0-9]*$)|([a-z][a-zA-Z0-9]*_$)
 method-rgx=([a-z][a-zA-Z0-9]*$)|([_]{2}[a-z]+[_]{2}$)|([a-z][a-zA-Z0-9]*_$)
 
 # Regular expression which should only match correct argument names
-argument-rgx=[a-z][a-zA-Z0-9]*$
+argument-rgx=[a-z][a-zA-Z0-9]*_?$
```
ericvaandering: What does adding underscore to the end do here?

juztas: I was testing it half day only with `dummy-variables-rgx=_$|dummy` and it was not working. So I guess pylint first makes `argument-rgx` check and adds if name is valid or not, if variable is not used it checks `dummy-variables-rgx`. Also this helped: http://stackoverflow.com/questions/19395165/pylint-isnt-ignoring-dummy-variables

ericvaandering:
So @ticoann this is OK to merge in my opinion, although others can weigh in. I know @annawoodard does not care for this but it allows you to do things like

```python
_host, port = blah()
```

which stops pylint from complaining that you did not use _host. The only current way to get pylint to be quiet is to say

```python
_, port = blah() 
```

ericvaandering:
Actually it looks like dummyHost should work and I think that's more clear and not likely to be confused with an "internal only" variable. Suggest we try that before approving this PR.

standards/.pylintrc
```diff
@@ -135,7 +135,9 @@ enable-variables=yes
 # Tells wether we should check for unused import in __init__ files.
 init-import=no
 
-
+# A regular expression matching the name of dummy variables (i.e. expectedly
+# not used).
+dummy-variables-rgx=_$|dummy
```
ticoann: are we allowing only _ and dummy for dummy variable or any thing start with dummy prefix?

ericvaandering: I thought as a compromise we could allow dummy*. Does that work with the current or not?

On Oct 7, 2015, at 14:27, ticoann <notifications@github.com<mailto:notifications@github.com>> wrote:


In standards/.pylintrc<https://github.com/dmwm/WMCore/pull/6302#discussion_r41434557>:

> @@ -135,7 +135,9 @@ enable-variables=yes
>  # Tells wether we should check for unused import in __init__ files.
>  init-import=no
>
> -
> +# A regular expression matching the name of dummy variables (i.e. expectedly
> +# not used).
> +dummy-variables-rgx=_$|dummy


are we allowing only _ and dummy for dummy variable or any thing start with dummy prefix?

juztas:
Updated, Now dummyVar, dummyHost and anything which starts with dummy should work

ticoann:
Thanks Justas, I will merge this,

ericvaandering:
Retest this please

Just to understand flipping tests.

ericvaandering:
Retest this please

ericvaandering:
@juztas you need to amend this to explicitly allow _

See the pylint report for #6322

PR #6307, opened 2015-10-02
add default format as json

### LINKED ISSUE #6306: Change RESTFormatter to accept */* and return json as default.
Need to add under following. 

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Server.py#L627

ticoann:
fixes #6306

### DIFF
--- a/src/python/WMCore/REST/Server.py
+++ b/src/python/WMCore/REST/Server.py
@@ -625,7 +625,8 @@ class MiniRESTApi:
         self.compression_chunk = 64 * 1024
         self.compression = ['deflate']
         self.formats = [ ('application/json', JSONFormat()),
-                         ('application/xml', XMLFormat(self.app.appname)) ]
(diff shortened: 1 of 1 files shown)


amaltaro:
I'll probably confuse you more than review. Do I understand it right that if a user asks for xml, he gets XML formatted data, if he wants json, he gets json and if he wants plain text, he would get json?

ericvaandering:
Those look real to me, @ticoann .

ticoann:
Actually this is not necessary and seems it is intentional to require Accept header. Closing it without merge.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Server.py#L748

ticoann:
Sorry about the buzz. Antanas was asking make more flexible to accept request if Accept header is not specified. I didn't know it was intentional doing this way until I saw Lassi's comment above. So I just sent email to Antanas that client need to specify the Accept header. And Alan, yes what you said is correct.

PR #6311, opened 2015-10-02
create wmstats link according to request origin

### LINKED ISSUE #6034: Clean up WQ and other caches created by Request.py in Service package

ticoann:
fixes #6034 - tested in my vm

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/minified/global.min.js
+++ b/src/couchapps/WMStats/_attachments/js/minified/global.min.js
@@ -31,7 +31,8 @@ WMStats.Globals = function($){
 		   "TotalInputLumis": "input_lumis",
 		   "TotalInputFiles": "input_num_files",
 		   "Run": "run",
-		   "AgentJobInfo": "AgentJobInfo"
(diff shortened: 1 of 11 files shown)


ticoann:

src/couchapps/WMStats/_attachments/js/minified/global.min.js
```diff
@@ -66,6 +67,16 @@ WMStats.Globals = function($){
     
     function getReqDetailPrefix () {
         if (_dbVariants[dbname] == "tier1") {
```
amaltaro: What is this tier1 dbname? Is it still in use? The same for the analysis one below.

amaltaro:
Patch looks good to me. Thanks!

src/couchapps/WMStats/_attachments/js/minified/global.min.js
ticoann: I am not sure analysis one is used (may be not)  but we have to distinguish between tier0 and production.

PR #6314, opened 2015-10-05
reqmgr2 auth fix

ticoann:
This is need for agent update the status (agent role is production-operator)
However we need to define roles and groups and their permissions.

### DIFF
--- a/src/couchapps/ReqMgr/validate_doc_update.js
+++ b/src/couchapps/ReqMgr/validate_doc_update.js
@@ -41,7 +41,6 @@ function(newDoc, oldDoc, userCtx) {
    var allowed = isGlobalAdm || matchesRole("admin", "group:reqmgr")
                              || matchesRole("data-manager", "group:reqmgr")
                              || matchesRole("web-service", "group:facops")
-                             || matchesRole("production-operator", "group:dataops")
                              || matchesRole("production-operator", "group:dataops");
    
    // Throw if user not validated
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, @ericvaandering , Alan, Eric, do you have any suggestion  on this. 
Which role/group should have permission to do specific things.

i.e. production-operation group have permission to change the status but not create the request, etc.

amaltaro:
Seangchan, I was thinking something like:
 1. "admin", "group:reqmgr" AND "web-service", "group:facops": root acces, can do anything :)
 2. "data-manager", "group:reqmgr": PdmV, PPD, McM folks, they should be able to inject configs, create, approve AND reject requests.
 3. "production-operator", "group:dataops": workflow team, should be able to inject configs, create, approve, abort/reject and make any other status change (perhaps not transitions meant to be done automatically)?
What do you think?

ericvaandering:
I think you guys know this much better than me.

ticoann:
> "data-manager", "group:reqmgr": PdmV, PPD, McM folks, they should be able to inject configs, create, approve AND reject requests.
> "production-operator", "group:dataops": workflow team, should be able to inject configs, create, approve, abort/reject and make any other status change (perhaps not transitions meant to be done automatically)? What do you think?

Thanks Alan, I was expecting  more fine control that that. i.e PPD, etc can only approve and reject there own request not others.

ticoann:
This need a further discussion but for now merging this since current code won't work without it.

PR #6315, opened 2015-10-06
Log components activity

### LINKED ISSUE #6209: Make sure ComponentLog always have some activity
As discussed yesterday during the meeting, we should make sure each component logs at least one single line per cycle, such that a ComponentLog will always have some logging in for the last X minutes (not more than 30min maybe).

amaltaro:
Fixes #6209


amaltaro:
@ticoann merge please

PR #6319, opened 2015-10-07
create acdc request depending on the original request creation

### LINKED ISSUE #6312: Switch ACDC request creation to create the reqmgr2 request
Currenly when ACDC is created, It is creating reqmgr1 request. We can either make this optional or change to reqmgr2 request creation. I would say we switch to create reqmgr2 request. (And fix the bugs if we can find) Making this optional will can be done but not sure that will confuse ops more.
@amaltaro, Alan, what do you think?

ticoann:
fixes #6312, If the original request is from reqmgr2 create reqmgr2 acdc, other wise reqmgr acdc as Alan suggested

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
+++ b/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
@@ -420,7 +420,7 @@ WMStats.ViewModel = (function (){
                 summary.ACDCDatabase = acdcServiceUrl.couchdb;
             }
             
-            requestInfo = reqMgrRequest.getData();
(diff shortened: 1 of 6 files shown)


ticoann:
@amaltaro, Please review. (Test in vm)

amaltaro:
Sorry Seangchan, I missed this PR. Can you tell me what's the difference between SoftwareVersions and CMSSWVersion? Is the first one a list of all cmssw releases set in a workflow (TaskChain/StepChain)?

ticoann:
I think it is the same thing SoftwareVersions is legacy code used by ReqMgr. we should remove it eventually.

PR #6322, opened 2015-10-08
Make dataset workunit for DQMHarvest instead of blocks

### LINKED ISSUE #6217: Investigate DQMHarvest with less events
To make sure I won't lose track of it after vacations, this issue requires investigation:
https://hypernews.cern.ch/HyperNews/CMS/get/dataopsrequests/7950/1/1/1/1/1/1/1/1/1/1/3/2.html

amaltaro:
The problem is that WorkQueueManager was creating one element (and fileset) per block, so for datasets with > 1 blocks, we were not harvesting full stats in the same job.

DO NOT MERGE IT YET.
I want to check whether removing the lfnbase stuff won't be a problem here.
Fixes #6217 . Pylint cleaning will come later as well as the unittests

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/DQMHarvest.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/DQMHarvest.py
@@ -22,10 +22,7 @@ class DQMHarvestWorkloadFactory(StdBase):
         self.workload.setDashboardActivity("harvesting")
         self.reportWorkflowToDashboard(self.workload.getDashboardActivity())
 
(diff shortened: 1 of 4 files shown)


amaltaro:
Took advantage that the PNN super patch got in and went a bit aggressive in the pylint corrections :)

ericvaandering:
Don't modify the two unused _ variable issues.

amaltaro:
Yep, I'm done with my changes. @ticoann it won't pass pylint tests, feel free to review the first commit and merge it

PR #6324, opened 2015-10-09
implement a copy-delete type subscription mode

hufnagel:
Normally there are only two subscription modes supported, copy or move. This implements a delayed move, consisting of a copy first, followed by a delete from the injection site once the workflow creating the data is done and the block is fully transferred to all sites the WMAgent made a subscription for.

If there is an active (outside the agent) subscription to the injection site, the delete is skipped.

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorSubscriber.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorSubscriber.py
@@ -35,6 +35,7 @@ from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 
 from WMCore.Services.PhEDEx import XMLDrop
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
+from WMCore.Services.PhEDEx.DataStructs.PhEDExDeletion import PhEDExDeletion
 from WMCore.Services.PhEDEx.DataStructs.SubscriptionList import PhEDExSubscription, SubscriptionList
 
 from WMCore.DAOFactory import DAOFactory
@@ -63,6 +64,7 @@ class PhEDExInjectorSubscriber(BaseWorkerThread):
         #    self.sendAlert will be then be available
         self.initAlerts(compName = "PhEDExInjector")
 
+        return
 
     def setup(self, parameters):
         """
@@ -76,6 +78,8 @@ class PhEDExInjectorSubscriber(BaseWorkerThread):
                                 logger = self.logger,
                                 dbinterface = myThread.dbi)
 
+        self.findDeletableBlocks = daofactory(classname = "GetDeletableBlocks")
+        self.markBlocksDeleted = daofactory(classname = "MarkBlocksDeleted")
         self.getUnsubscribed = daofactory(classname = "GetUnsubscribedDatasets")
         self.markSubscribed = daofactory(classname = "MarkDatasetSubscribed")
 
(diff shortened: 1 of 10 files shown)


hufnagel:
Btw, as a comment, which I wrote this because I needed it in the Tier0, the implementation is completely generic and can be used for general WMAgent operation if so desired...

hufnagel:
Basically, this allows you to run a Taskchain style request where a later step reads a previous steps output, while at the same time supporting aggregating the requests output at a site of your choice and not leave any remains at the site or sites where the request is actually running.

hufnagel:
@amaltaro , please include this in your testing.

hufnagel:
Btw, this is currently a no-op for WMAgent operations as you don't use the directive that enables this. It would be good to see that it really does nothing though.

hufnagel:
Ok, let me address these issues...

src/python/WMComponent/PhEDExInjector/PhEDExInjectorSubscriber.py
```diff
@@ -91,15 +95,148 @@ def algorithm(self, parameters):
...
+
+            if blockName in subscriptions and location in subscriptions[blockName]:
+                logging.debug("Block %s subscribed to %s, skip deletion" % (blockName, location))
+                binds = { 'DELETED' : 2,
+                          'BLOCKNAME' : blockName }
+                self.markBlocksDeleted.execute(binds, transaction = False)
+            else:
+                blockInfo = []
+                try:
+                    blockInfo = self.phedex.getReplicaInfoForBlocks(block = blockName, complete = 'y')['phedex']['block']
```
amaltaro: instead of calling the replicas api, can't you use the subscription one? AFAIK it's a lighter phedex api

hufnagel: I can't because I actually need to know that the replica is complete at the target site. For the subscriptions I wouldn't need to call PhEDEx, I could look that up inside WMBS.

hufnagel:
Ok, the broken unit tests are real, will fix them. Also will fix the logging to the new style.

What about the missing python3 imports ? Looks bogus to me as the files in question don't even use logging or divisions (and it says so in the message). So why should I add the imports ???

ericvaandering:
Cause you might add those things later and we want people to get used to the new ways of writing this code. Please add them. (This requirement is only for new files)


> On Oct 9, 2015, at 15:10, Dirk Hufnagel <notifications@github.com> wrote:
> 
> Ok, the broken unit tests are real, will fix them. Also will fix the logging to the new style.
> 
> What about the missing python3 imports ? Looks bogus to me as the files in question don't even use logging or divisions (and it says so in the message). So why should I add the imports ???
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/dmwm/WMCore/pull/6324#issuecomment-146975529>.
> 




Eric W. Vaandering 
Computing Physics Developer

Scientific Computing Division
630 840 6678 office
www.fnal.gov <http://www.fnal.gov/>

hufnagel:
retest this please

ericvaandering:
This is definitely something that either got checked into master or an unstable test.

See https://github.com/dmwm/ContinuousIntegration/issues/6#issuecomment-146986328 or Slack which will keep a running tally now of changes in tests.


> On Oct 9, 2015, at 20:12, DMWM Autobot <notifications@github.com> wrote:
> 
> WMCore_t.WorkQueue_t.WorkQueue_t.WorkQueueTest:testProductionMultiQueue changed from success to failure
> 




Eric W. Vaandering 
Computing Physics Developer

Scientific Computing Division
630 840 6678 office
www.fnal.gov <http://www.fnal.gov/>

hufnagel:
Ok, so that failed unit test should not stop us from merging this. I am following up on more pylint complains though...

hufnagel:
@ticoann , are we waiting on anything to merge this ?

ticoann:
No Just your green light. :smile:  I will merge this.

PR #6325, opened 2015-10-09
work around EOS auth problems

hufnagel:
There is a problem with the gsi authentication plugin in xrootd. Effect is that jobs running at CERN loose their EOS privileges after a while and the job dies as it can't read any input anymore.

This patch works around this problem by increasing an internal xrootd timeout.

This has been open for a while, the xrootd team didn't really provide a real fix apart from "please upgrade to the latest xrootd client and check if this is still an issue". Even if the latest xrootd client fixes this, it'll take a while until we stop using CMSSW releases that use older clients.

Making a pull request because a) don't want to run with this custom patch in the Tier0 forever and b) CompOps production jobs at CERN should also be affected by this.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
@@ -10,7 +10,7 @@ import logging
 import os
 import subprocess
 import sys
-import time
+import socket
 
 from WMCore.FwkJobReport.Report import addAttributesToFile
 from WMCore.WMRuntime.Tools.Scram import Scram
@@ -24,7 +24,7 @@ def analysisFileLFN(fileName, lfnBase, job):
     Construct an LFN for a user file
     """
 
-    dummy, base = os.path.split(fileName)
+    base = os.path.split(fileName)[1]
     root, ext = os.path.splitext(base)
 
     newBase = '{base}_{count:04d}{ext}'.format(base=root, ext=ext, count=job['counter'])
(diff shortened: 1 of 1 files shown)


hufnagel:
Ok, addressing these issues too...

amaltaro:
Dirk, is this 86400 the maximum length a job can run at CERN without facing these xrootd disconnections? If so, then we should increase it to 2 days at least, no?

hufnagel:
24h is what was recommended, I actually don't know exactly what this timeout does...

Also fixed the last spurious import.

hufnagel:
Btw, I feel a bit uncomfortable merging this, but I also feel uncomfortable leaving it out as it seems to ne absolutely required for Tier0 operations (or also CompOps operations) at CERN. I started a HN thread about whether this is safe to merge

https://hypernews.cern.ch/HyperNews/CMS/get/edmFramework/3572.html

maybe we should push a bit more to get an answer on this...

amaltaro:
Yeah, to be honest I'm not really confident this problem happens in central production at CERN. Is Tier0 being locally mapped to cmsprd or cmsprod?

ericvaandering:
Maybe on the timeout question, ask on the wan-access HN.

hufnagel:
Maybe it doesn't happen to CompOps because they don't run much on the CERN cloud resources. But once you do you'll be inside the same pilots as the Tier0 jobs, I very much doubt that you won't be affected by this. CompOps using T2_CH_CERN/LSF is a different question, but that use should go down to almost zero over time.

hufnagel:
Anyways, the way this was explained to me was a problem between CMSSW, the gsi plugin authentication handling inside xrootd and EOS, so to first order *everyone* running at CERN should be affected.

hufnagel:
retest this please

hufnagel:
retest this please

ericvaandering:
test this please

ericvaandering:
test this please

ticoann:
test this please

hufnagel:
Modify the patch to only implement the workaround at CERN.

hufnagel:
@ticoann @ericvaandering , this is ready to be merged, can you review it please ?

ticoann:
patch looks fine to me.

Issue #6299, opened 2015-10-01
Add a few other fields to workload_cache reqmgr2


amaltaro:
We should add the following fields (provided during assignment) to the reqmgr_workload_cache document, in reqmgr2:
dashboardActivity, maxRSS, maxVSize, softTimeout and hardTimeout.

ticoann:
Hi Alan, those are already in the reqmgr_workload_cache. But with CamalCase except hardTimeout

amaltaro:
I didn't find any of them here Seangchan
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_ReDigi_cmsRun2_PNN_Oracle_151008_114823_2481
maybe we are talking about different docs?

ticoann:
Ah, this only works with reqmgr2 request.
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_StepChain_ReDigi3_ReqMgr2_Validation_151008_003414_7581

However, it seems hardTimeout is just softTimeout + gracePeriod.  I can add the field (HardTimeout) though.

amaltaro:
Seangchan, it seems they are not available even for ReqMgr2 requests only, see
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_TaskChainPyquenZeemumuJets_PU_ReqMgr2_Validation_151008_003253_907

amaltaro:
Sorry Seangchan, I picked a workflow that has not been assigned, that's why those fields are not there. This one was assigned:
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_TaskChainTTbar_ReqMgr2_Validation_151008_003315_396

maxRSS and maxVSize are there, but not the dashboardActivity. Can you make this change as well?

ticoann:
Actually, the name is Dashboard instead of DashboardActivity. Not sure why we change the name in json doc.

amaltaro:
All these fields are now available for reqmgr2.

PR #6326, opened 2015-10-09
add missing output dataset in reqmgr2

ticoann:
@amaltaro, Alan, thanks finding this bug. Please review. Is there any other record missing in reqmgr2 couchdb or want to add?

Patch is tested in my vm.

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -363,6 +363,9 @@ class Request(RESTEntity):
         
         dn = cherrypy.request.user.get("dn", "unknown")
         
+        if ('SoftTimeout' in request_args) and ('GracePeriod' in request_args): 
+            request_args['HardTimeout'] = request_args['SoftTimeout'] + request_args['GracePeriod']
+        
         if "total_jobs" in request_args:
             # only GQ update this stats
             # request_args should contain only 4 keys 'total_jobs', 'input_lumis', 'input_events', 'input_num_files'}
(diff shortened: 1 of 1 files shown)


src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -414,6 +414,23 @@ def delete(self, request_name):
...
+        Update request_args here if additional information need to be put in couchdb
+        """
+        request_args['RequestWorkflow'] = sanitizeURL("%s/%s/%s/spec" % (request_args["CouchURL"], 
+                                            request_args["CouchWorkloadDBName"], workload.name()))['url']
+            
+        request_args['OutputDatasets'] = []
+        # Add the output datasets if necessary
+        # for some bizarre reason OutpuDatasets is list of lists
+        for ds in workload.listOutputDatasets():
+            request_args['OutputDatasets'].append(ds)
```
amaltaro: how about extend() instead of append? that should give us a simple list

amaltaro:
hi Seangchan, since you are asking, want to take a look at this one 
https://github.com/dmwm/WMCore/issues/6299
? :-D

ticoann:
Thanks Alan, Now I remember we had another issue.  I will try to update that.

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: Thanks Alan, acutally can just do following (I have to check what exactly listOutputDatasets() returns but it seems it returns list of str.
```
request_args['OutputDatasets'] = workload.listOutputDatasets()
```

ticoann:
Hi Alan, I added HardTimeout

PR #6330, opened 2015-10-13
fix a PhEDExInjector deadlock problem

hufnagel:
Should fix an occasional Oracle deadlock crash.

Transaction included too much unrelated code and isn't needed in any case.


hufnagel:
@ticoann , can you review ? I am running a Tier0 replay right now, seems to be working alright. We likely will deploy this the next time the PhEDExInjector crashes in the prod Tier0.

ticoann:
Thanks Dirk, This looks fine too, me. Once unittest finishes I will merge that and include in the tag.

PR #6331, opened 2015-10-15
Update WQ to handle PhEDExNodeNames; fixed remote read

amaltaro:
Fixes 2 things. 
1. My last fix for DQMHarvest :(, since I followed the SE convention instead of PNNs
2. setting to read data remotely would fail since updating a set with string will break the string down in chars and update the set with each elements

Do NOT merge yet please, needs further test.

### DIFF
--- a/src/python/WMCore/WorkQueue/WMBSHelper.py
+++ b/src/python/WMCore/WorkQueue/WMBSHelper.py
@@ -737,10 +737,13 @@ class WMBSHelper(WMConnectionBase):
                 continue
 
             # Create a LumiList from the WMBS info
-            lumis = []
+            runLumis = {}
             for x in f['LumiList']:
-                lumis.append([str(x['RunNumber']), x['LumiSectionNumber']])
-            fileLumiList = LumiList(lumis=lumis)
+                if x['RunNumber'] in runLumis:
+                    runLumis[x['RunNumber']].extend(x['LumiSectionNumber'])
+                else:
+                    runLumis[x['RunNumber']] = x['LumiSectionNumber']
+            fileLumiList = LumiList(runsAndLumis=runLumis)
 
             if lumiMask:
                 if fileLumiList & lumiMask:  # At least one lumi from file is in lumiMask
--- a/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
+++ b/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
@@ -24,7 +24,7 @@ def mapSitetoSE(sites):
     fakeSEs = []
     rControl = ResourceControl()
     for site in sites:
-        fakeSEs.extend(rControl.listSiteInfo(site)['se_name'])
+        fakeSEs.extend(rControl.listSiteInfo(site)['pnn'])
     return fakeSEs
 
 class PileupFetcher(FetcherInterface):
--- a/src/python/WMCore/WorkQueue/DataLocationMapper.py
+++ b/src/python/WMCore/WorkQueue/DataLocationMapper.py
@@ -47,6 +47,14 @@ def timeFloor(number, interval = UPDATE_INTERVAL_COARSENESS):
     from math import floor
     return floor(number / interval) * interval
 
+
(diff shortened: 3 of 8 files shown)


ericvaandering:
test this please

amaltaro:
The LumiList change should fix the mystery of subscriptions without actual files in WMBS. The problem is that lumiList is a list of lists which then fails to be sorted here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/DataStructs/LumiList.py#L87
giving as result an empty compactList, which later on does not match in WMBSHelper..

Let's see whether it breaks any unit tests.

ericvaandering:
Please don't change lumilist. I think we're using it wrong if it doesn't work as is.

src/python/WMCore/DataStructs/LumiList.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, filename = None, lumis = None, runsAndLumis = None, runs = No
                 run = str(run)
                 if run not in runsAndLumis:
                     runsAndLumis[run] = []
-                runsAndLumis[run].append(lumi)
+                runsAndLumis[run].extend(lumi)
```
ericvaandering: The list of lists is not a list of lumis, it is a list of ranges. So if you need to do this, you need to extend the list *outside* of this class.

In other words [[1,5], [8,9]] is equivalent to [1,2,3,4,5,8,9]

amaltaro:
Ok, I reverted the changes in LumiList. It was indeed being initialized in the wrong way. Now instead of passing "lumis" (list of list) we pass a dict with runAndLumis.

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -737,10 +737,10 @@ def validFiles(self, files):
                 continue
 
             # Create a LumiList from the WMBS info
-            lumis = []
+            runLumis = {}
             for x in f['LumiList']:
-                lumis.append([str(x['RunNumber']), x['LumiSectionNumber']])
-            fileLumiList = LumiList(lumis=lumis)
+                runLumis[str(x['RunNumber'])] = x['LumiSectionNumber']
```
ericvaandering: I'm not sure this will work either. Typically the agent has run/lumi pairs and here you will overwrite the each run as a new lumi comes in. (In other words, RunNumber is often duplicated)

amaltaro: That's why I would prefer having LumiList sorting all these things out for us, otherwise we need to parse the N different types of run/lumi information everywhere in the code to make sure we're properly instantiating a lumilist object.
Anyways, it should work in this case since f['LumiList'] is in the following format:
[{'RunNumber': 203777, 'LumiSectionNumber': [2, 23, 63, 48]]}]

where each element of the list would be mapped to a run and an extended list.
To be on the safe side, I think it's better to protect and not overwrite the run value in case it already exists in the dict.

ericvaandering: Lumi list should now sort things. If not, I will make it do so. But yeah, you will want to do a conditional extend.

I have not looked at why some tests are still failing.

Eric

amaltaro:
Changing globals to fix most of the WMBS unit tests, still some to go, but I mainly want to see whether this change will break other tests or not...

ticoann:
test this please

amaltaro:
test this please

amaltaro:
Tried to fix a few more unit tests, though it seems Dataset_t is using the old DBSReader (DBS2) emulator, which is completely outdated. I still have to look at the WorkQueue issues, I'm not sure they are related to this change or Seangchan's change.

I'll look at it later in the evening @ticoann if possible wait for me before merging it

ericvaandering:
"though it seems Dataset_t is using the old DBSReader (DBS2) emulator, " no surprise there. That code is everywhere.

amaltaro:
@ticoann just deployed this branch and run the Harvest and JobGroup tests and they ran without any issues. Please review and merge it for the new tag.

PR #6333, opened 2015-10-16
make the jobsplitting trustsitelists aware

hufnagel:
If the TrustSiteLists paramater is set, change the job splitting to only look at sitewhitelists and siteblacklists and not real locations.

Currently running a Tier0 replay (PromptReco at T1) to test the basics.

@ticoann @amaltaro does the change look alright ?

### DIFF
--- a/src/python/WMCore/JobSplitting/JobFactory.py
+++ b/src/python/WMCore/JobSplitting/JobFactory.py
@@ -6,11 +6,8 @@ _JobFactory_
 
 import logging
 import threading
-import gc
 
 from WMCore.DataStructs.WMObject import WMObject
-from WMCore.DataStructs.Fileset  import Fileset
-from WMCore.DataStructs.File     import File
 from WMCore.Services.UUID        import makeUUID
 from WMCore.WMBS.File            import File as WMBSFile
 from WMCore.DAOFactory           import DAOFactory
@@ -27,11 +24,11 @@ class JobFactory(WMObject):
     def __init__(self,
                  package='WMCore.DataStructs',
                  subscription=None,
-                 generators=[],
+                 generators=None,
                  limit = 0):
         self.package = package
         self.subscription  = subscription
-        self.generators    = generators
+        self.generators    = generators if generators else []
         self.jobInstance   = None
         self.groupInstance = None
         self.jobGroups     = []
@@ -46,12 +43,18 @@ class JobFactory(WMObject):
         self.daoFactory    = None
         self.timing = {'jobInstance': 0, 'sortByLocation': 0, 'acquireFiles': 0, 'jobGroup': 0}
 
-        if package == 'WMCore.WMBS':
+        if package == "WMCore.WMBS":
+
             myThread = threading.currentThread()
+
             self.daoFactory = DAOFactory(package = "WMCore.WMBS",
                                          logger = myThread.logger,
                                          dbinterface = myThread.dbi)
             self.getParentInfoAction  = self.daoFactory(classname = "Files.GetParentInfo")
+
+            self.pnn_to_psn = self.daoFactory(classname = "Locations.GetPNNtoPSNMapping").execute()
+
+        return
 
     def __call__(self, jobtype = "Job", grouptype = "JobGroup", *args, **kwargs):
         """
@@ -86,10 +89,9 @@ class JobFactory(WMObject):
         map(lambda x: x.start(), self.generators)
 
 
-        self.limit        = int(kwargs.get("file_load_limit", self.limit))
+        self.limit = int(kwargs.get("file_load_limit", self.limit))
         self.algorithm(*args, **kwargs)
         self.commit()
-        #gc.collect()
 
         map(lambda x: x.finish(), self.generators)
         return self.jobGroups
@@ -102,17 +104,18 @@ class JobFactory(WMObject):
         subscription and splits them into jobs and inserts them into job groups.
         The algorithm must return a list of job groups.
         """
-        self.newGroup(args, kwargs)
+        self.newGroup()
         self.newJob(name='myJob')
+        return
 
-    def newGroup(self, *args, **kwargs):
+    def newGroup(self):
         """
         Return and new JobGroup
         """
         self.appendJobGroup()
         self.currentGroup = self.groupInstance(subscription=self.subscription)
         map(lambda x: x.startGroup(self.currentGroup), self.generators)
-
+        return
 
     def newJob(self, name=None, files=None, failedJob=False, failedReason=None):
         """
@@ -124,15 +127,6 @@ class JobFactory(WMObject):
         self.currentJob["jobType"] = self.subscription["type"]
         self.currentJob["taskType"] = self.subscription.workflowType()
         self.currentJob["owner"] = self.subscription.owner()
-
-        # only put into job object if relevant
-        # JobSubmitter can deal with absence
-        if len(self.siteBlacklist) > 0:
-            self.currentJob["siteBlacklist"] = self.siteBlacklist
-        if len(self.siteWhitelist) > 0:
-            self.currentJob["siteWhitelist"] = self.siteWhitelist
-        if self.trustSitelists:
-            self.currentJob["trustSitelists"] = self.trustSitelists
 
         # All production jobs must be run 1
         if self.subscription["type"] == "Production":
@@ -147,7 +141,7 @@ class JobFactory(WMObject):
         for gen in self.generators:
             gen(self.currentJob)
         self.currentGroup.add(self.currentJob)
-
+        return
 
     def appendJobGroup(self):
         """
@@ -163,7 +157,6 @@ class JobFactory(WMObject):
 
         return
 
-
     def commit(self):
         """
         Bulk commit the JobGroups all at once
@@ -173,29 +166,57 @@ class JobFactory(WMObject):
         if len(self.jobGroups) == 0:
             return
 
-        logging.debug("About to commit %i jobGroups" % (len(self.jobGroups)))
-        logging.debug("About to commit %i jobs" % (len(self.jobGroups[0].newjobs)))
+        logging.debug("About to commit %i jobGroups", len(self.jobGroups))
+        logging.debug("About to commit %i jobs", len(self.jobGroups[0].newjobs))
 
         if self.package == 'WMCore.WMBS':
+
+            for jobGroup in self.jobGroups:
+
+                for job in jobGroup.newjobs:
+
+                    if self.trustSitelists:
+                        locSet = set(self.siteWhitelist) - set(self.siteBlacklist)
+                    else:
+                        locSet = set([])
+                        for pnn in job['input_files'][0]['locations']:
+                            locSet.update(self.pnn_to_psn.get(pnn, []))
+
+                        if len(self.siteWhitelist) > 0:
+                            locSet = locSet & set(self.siteWhitelist)
+                        if len(self.siteBlacklist) > 0:
+                            locSet = locSet - set(self.siteBlacklist)
+
+                    job['possiblePSN'] = locSet
+
+                # now after the jobs are created, remove input file locations
+                # they are no longer needed and just take up space
+                for job in jobGroup.newjobs:
+                    for fileInfo in job['input_files']:
+                        fileInfo['locations'] = set([])
+
             self.subscription.bulkCommit(jobGroups = self.jobGroups)
+
         else:
-            # Then we have a DataStructs job, and we have to do everything
-            # by hand.
+
+            # we have a DataStructs job and have to do everything by hand
             for jobGroup in self.jobGroups:
                 jobGroup.commit()
                 for job in jobGroup.jobs:
                     job.save()
             self.subscription.save()
 
-        #gc.collect()
         return
 
     def sortByLocation(self):
         """
         _sortByLocation_
 
-        Sorts the files in the job by location and passes back a dictionary of files, with each key corresponding
-        to a set of locations
+        Retrieve available files and return them sorted by loacation.
+        The keys in the dict correspond to a set of locations.
+
+        If TrustSiteLists is set ignore the file locations and treat all files
+        as being present in the same place (use key "AAA" location).
         """
 
         fileDict = {}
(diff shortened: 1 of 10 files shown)


amaltaro:
I think you'll need to cast them to a set first.
Then we'd also need to update the workqueue code which also does data location, though I didn't check where in the code.

hufnagel:
We can ignore the workqueue for now, the mechanism there is completely different and orthogonal to this. All we need is to propagate the trigger for TrustSiteLists to the request level so that it can actually be used. Then we will have two different parameters that do the different ways to do xrootd overrides. We can keep them both for now.

hufnagel:
Actually, this approach doesn't work. Problem is that file['locations'] are PNN and sitewhitelist and siteblacklist are PSN. Of the top of my hat the only way around that is to add a PSN to PNN lookup in the job splitter to translate the sitewhitelist/siteblacklist into PNN.

This is a bit silly then though as we do a PSN to PNN translation in the JobCreator just to have a reverse PNN to PSN translation in the JobSubmitter.

I see two ways to avoid this silliness:

1) If trustsitelist is used override the job file locations with a special single "OVERRIDE' location. That will ensure that the job splitting algorithm will consider them all at the same place and do the right thing. The JobSubmitter doesn't care either because it anyways takes the possible sites to submit to from the sitewhitelist and siteblacklist, ie. it doens't even look at the job file locations.

2) Remove the PNN to PSN translation from the JobSubmitter altogether and have the JobCreator determine the PSN for a job based on the file location or the sitewhitelist/siteblacklist if trustsitelists is used.

Method 2 is a bit cleaner but has the disadvantage that the PNN to PSN matching is only done once at job creation time. This assumes though that the JobSubmitter in its current implementation would actually take into account changes over time in the PNN to PSN mapping. Does it ?

Comments ?

amaltaro:
I'm confused now. At the JobCreator level, what we need to have (at least for this case) is the PSN only, since we do not care about the PNN for trustSitelists. Said that, can't we put the PSN in the job 'location' for THIS specific case?

hufnagel:
No. The JobCreator uses input file locations to pre-sort files for job splitting. The assumption here is that jobs at different sites should never be assigned to the same job.

This isn't true anymore when you use trustsitelists. In this case the jobsplitter needs to consider these files to all be in the same location. The way the code is written, if I want to accomplish this correctly, I need to take the sitewhitelist - siteblacklist, convert that list of PSN to PNN, update the jobs input file location with these PNN and then pass them to the job splitting algorithm.

hufnagel:
So that brings us to the two options I mentioned.

Skip the PSN to PNN conversion I mentioned and just set the PNN for the jobs input file to some random identical value so the job splitting works correctly. We pass a non-valid PNN to the JobSubmitter as part of the job record, but it doesn't matter because the JobSubmitter skips the PNN to PSN conversion if trustsitelists is used.

Or do the PSN to PNN conversion in the JobSubmitter, but if we have to have the code to do that there anyways, I would rather do everything there and just determine where we can submit jobs to in the JobCreator and pass that list of possible PSN to the JobSubmitter (and remove the PNN to PSN conversion from the JobSubmitter).

amaltaro:
> The assumption here is that jobs at different sites should never be assigned to the same job.

You meant files in different sites should not be assigned to the same job, right? However, this is AAA world, we don't know and we *should* not care about data/file location.

hufnagel:
Well, the job splitter cares !

ticoann:
I think  Dirk's method 2 is cleaner. 
By converting file location (supposed to be PNN) to PSN here. (Not completely sure that is the only place)
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/JobFactory.py#L212
and remove the conversion from 
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L323

So split job only handles PSN. I don't think we should concern about PSN and PNN mapping change between job creation and job submission.  That mapping doesn't change often write?

However current patch should work as well. It will simply create the splitting correctly when the trusted sites are set but that location is not used for submission anyway since we that location is completely ignored.

Personally I prefer cleaner changes (Dirk's 2nd method) if the change is simple. But we have to be careful whether there will be any hidden side effect.

ticoann:
Of course this assumes we have the site populated in wmbs_location table. (Current code won't populate the sites without pnn in site db).

hufnagel:
I am testing my method 1 right now because it was faster to create the code. But I also prefer method 2 and will look into it.

ticoann:
Dirk, do you really need to do fake site "aaa"? Doesn't your initial patch work? Either you set the fake location or not it should be ignored, isn't it?

hufnagel:
I need to because the available files are sorted according to locSet. They need to have the same locSet for the splitting algorithm to all process them together.

ticoann:
I mean instead of using 'aaa', use actual white list site although it is not PNN but PSN.

hufnagel:
Functionally it should be the same, if its not and these file locations are actually accessed in the trustsitelists case I want to find this out with this test (as they shouldn't be).

hufnagel:
Ok, I am implementing method 1 now. One thing I noticed, the system assumes that site whitelist/blacklist are in WMAgent siteName, which is internally different than cmsName. I think in practice everyone assumes that site whitelist/blacklist is in cmsName, ie. PSN anyways, so I removed that conversion code from the JobSubmitter.

Way I implemented this to keep the JobCreator code simple is to take the file locations, convert them to PSN, then sort by them but also update the file locations to the PSN so they get included with the job record that is written out later and passed to the JobSubmitter. Then the JobSubmitter just interprets these file locations as possible PSN it can run at and just modified according to sites that are in draining or down.

First try, almost expect some unit tests to break, but want to see what jenkins makes of it.

hufnagel:
Ok, that doesn't work. The Tier0 doesn't use sortByLocation() because it has to do its own data discovery queries, therefore anything implemented there doesn't work. I need to look at what the JobCreator does with the jobgroups and jobs it gets back from the splitting algorithm.

Which means for sortByLocation() I am going back to just using "AAA" as a possible location.

hufnagel:
retest this please

hufnagel:
Ok, that last commit worked (the testProductionMultiQueue failure is bogus). I removed the site whitelis and blacklist functionality test from the JobSubmitter unit test though, need to add at least one test for this to one of the job splitting tests to make sure it works correctly. Will use the FileBased splitter unit test (doing it in one is enough).

hufnagel:
I also still want to remove the job input files location data from the job object that is passed to the JobSubmitter. It currently breaks some of the unit tests because the File objects are reused multiple times (first test works then, later break because the location is gone). Will need to do some refactoring to avoid that reuse.

hufnagel:
retest this please

ericvaandering:
test this please

hufnagel:
Ok, great, unit tests are happy :-). @ticoann @amaltaro , can you review please ?

I am currently running another Tier0 replay, but its almost done and I didn't see any problems.

hufnagel:
One question I still have, what is the difference between siteName and cmsName ? Is there one in practice. It seems to me we use siteName as PSN (because that is what the JobSubmitter is using as location. The site whitelist/blacklist were assumed to be cmsNames though.

Are they different anywhere in our system ? I though what the requestors put into site whitelist/blacklist is actually a PSN, ie. a location jobs are submitted to...

hufnagel:
This patch changes the old behavior in that it treats site whitelist/blacklist as siteName now, ie. they are interpreted directly as "sites" that are used by the submission system and do not go through a cmsName ->siteName conversion first. Would be good if someone can confirm that this won't cause problems.

amaltaro:
I see we have both cms_name and site_name in the wmbs_location table, though all the 125 rows/sites in my agent have exactly the same value for them. I don't think there will be an issue here.

Now the question could be whether their definition is the same or not, I honestly have no idea.

hufnagel:
I think the design has a separation of what a "site" is called within WMAgent and what a CMS "site" name is, that's why we have a conversion from the cmsName in site whitelist/blacklist to site name. In practice though that same site name is then used directly for job submission.

So while you could setup a site with siteName "site1" with cmsName "T1_US_FNAL", the old code would correctly handle the conversion of "T1_US_FNAL" to "site1", in practice this is useless because this concept of "site1" does not extend to gWMS. If you look at

WMCore/src/python/WMCore/WMBS/MySQL/Locations/GetPSNtoPNNMapping.py

the siteName is directly interpreted as a PSN and we later put it into the JDL to request what sites a job can run at.

So I think in practice cmsName is useless. Agreed ?

src/python/WMCore/JobSplitting/JobFactory.py
```diff
@@ -27,11 +24,11 @@ class JobFactory(WMObject):
     def __init__(self,
                  package='WMCore.DataStructs',
                  subscription=None,
-                 generators=[],
+                 generators=None,
                  limit = 0):
         self.package = package
         self.subscription  = subscription
-        self.generators    = generators
+        self.generators    = generators if generators else []
```
amaltaro: Besides shutting up pylint, it makes no sense. Why don't we just default to [] since lambda function is called on it later on?

hufnagel: Well, the usual argument about not having empty lists or dicts as default arguments applies. We discussed this before, its just in general a really bad habit and you should never do it, even if in the specific case it doesn't hurt you.

amaltaro: That's why I made my comment, because implicitly the default value is still []  
If we default to [] 3 lines below the previous location, then just leave it as it was before :)

hufnagel: They are different. The new code creates a new empty list every time the function is called. With the [] in the call definition that is not the case.

amaltaro:
> So I think in practice the separation between siteName and cmsName is useless. Agreed ?

I think so, though I still see cms_name being used in a few other places here. Perhaps we should also update them to site_name to avoid confusion?

hufnagel:
So I think in practice the separation between siteName and cmsName is useless. Agreed ?

    I think so, though I still see cms_name being used in a few other places here. Perhaps we should also update them to site_name to avoid confusion?

Yes, agreed. But I would open a new issue/PR for that, this PR is already long enough.

hufnagel:
I just noticed a comment I need to change, but I want to wait until the rest of the code is reviewed and has been signed off on. I can then fix the comment, update the PR and we can merge after the jenkins run completes again.

amaltaro:
Hmm, it seems the underlying plugins are using either "locations" or "possibleSites" key to get the PSN (target site list) for submission, see
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/CondorPlugin.py#L1010

how did your test work if it was removed from JobSumitterPoller code (?)

hufnagel:
If you look at the JobSubmitter code, it generates possibleSites from possiblePSN.

https://github.com/hufnagel/WMCore/blob/a4e274b1fe677622b1c5439aae271de915d0327e/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L292

https://github.com/hufnagel/WMCore/blob/a4e274b1fe677622b1c5439aae271de915d0327e/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L369

https://github.com/hufnagel/WMCore/blob/a4e274b1fe677622b1c5439aae271de915d0327e/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L665

Seems to me that possibleSites is directly based on possiblePSN (maybe we can unify somehow ?) and potentialLocations is based on that plus the filter on down/drained sites.

hufnagel:
Look here

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/CondorPlugin.py#L1018

DESIRED_Sites gets all possible sites ignoring site down/drain and ExtDESIRED_Sites gets the list of sites that includes site down/drain ? Does this make sense ?

ticoann:
> DESIRED_Sites gets all possible sites ignoring site down/drain and ExtDESIRED_Sites gets the list of sites that includes site down/drain ? Does this make sense ?

I might be wrong but I don't think ExtDESIRED_Sites is used in Glidein at least for now. Do you know how ExtDESIRED_Sites is used?

hufnagel:
Well, if we don't use EstDESIRED_Sites, we don't use site down/drain status during job submission and we might as well rip that code out... It might have been used ages ago when we were using other submission plugins, but for the CondorPlugin the sit down/drain status seems to only affect ExtDESIRED_Sites.

hufnagel:
But again, we should open a new issue for that :-). My changes don't touch this part.

ticoann:
Hi Dirk,
One thing I am concerning is that now the trusted/white site list matching/assign is happening in job creation instead of job submission. One outstanding issue we had is that add ability to make site whitelist change after workflow is running status. I think it wouldn't make big difference if the time between job creation and submission is small but it would be better that matching happens in submission time than creation time.

ticoann:
I think we eventually clean up all other location handling other (i.e. CMS site)  than PSN and PNN.  But we could handle that in separate issue.

hufnagel:
> One thing I am concerning is that now the trusted/white site list matching/assign is happening in job creation instead of job submission. One outstanding issue we had is that add ability to make site whitelist change after workflow is running status. I think it wouldn't make big difference if the time between job creation and submission is small but it would be better that matching happens in submission time than creation time.

Is that really a problem ? You anyways will have to deal with jobs that are already running/pending, there is no way to change them unless you condor edit them. So changing the whitelist or blacklist will never be instantaneous. 

I could change this back, but it will make the code less clean because you have to propagate this information further down and can't deal with all of these dependencies in one place. My changes you evaluate jobs input file locations, the associated PSN, the site whitelist/blacklist and whether we should ignore file location and trust the site whitelist/blacklist in a single piece of code. If you apply the site whitelist/blacklist during job submissions, you do it in two places.

Also would have to revert/change a few unit tests that rely on this being done in the JobCreator now.

Its doable, but is it worth it, given that even with this change you still won't be able to switch site whitelist/blacklist immediately, there will always be a backlog of jobs that you can only edit in condor ?

hufnagel:
Btw, if you want to change the site whitelist/blacklist for already created jobs, you actually would have to edit the job sandbox that is written to disk. This information isn't passed via database, its persistent in files on disk.

Are we really doing this ? If we don't, we might as well do everything in the JobCreator because that is anyway as far as an updated site whitelist/blacklist will reach.

hufnagel:
Just had a chat with @ericvaandering about this, I don't think we want to support automatic patching of job sandboxes. And for manual Ops procedures/scripts you can still change the possiblePSN parameter in the job sandbox.

ticoann:
Hi Dirk,  All your concern is valid. that is why we didn't implemented yet. 

The main reason this was issue was when workflow is created site was fine but ops want to be able to change the site whitelist when they found the problem in the middle of running. (better site comes out in terms of resource) 

> Is that really a problem ? You anyways will have to deal with jobs that are already running/pending, there is no way to change them unless you condor edit them. So changing the whitelist or blacklist will never be instantaneous.

running jobs are out of control until it fails and comes back. For pending jobs, we can use condor_edit to change but I really don't want to do that.  

> Btw, if you want to change the site whitelist/blacklist for already created jobs, you actually would have to edit the job sandbox that is written to disk. Every single job sandbox.

Yes, we have to change that as well.

So initial plan was we don't  care about already submitted jobs. But we change this before submission.
Anyway, since Eric said not supporting this. :smile:  I am fine with that.

hufnagel:
Its not that we will never support this. But in the current design it doesn't make sense to have an automatic mechanism that modifies every single job sandbox on disk. We might still have to support that functionality, but not in an automated fashion, but only for emergency "these jobs won't run anywhere if we don't patch the sandbox" manual interventions. And for these you can always modify possiblePSN and directly specify where jobs should be run, no need to fiddle with whitelist or blacklist.

When we do the planned redesign with the division of "unit of work" and job, we should keep this use case in mind though, maybe there is an easier way to accomplish this...

hufnagel:
Ok, as there weren't any further code comments I am fixing the one comment and push the changes. Once this goes through another jenkins test we can decide when this should be merged.

ticoann:
test this please

ericvaandering:
test this please

PR #6334, opened 2015-10-16
fix WorkQueue_t test

ticoann:
1. fixes unittests caused by changing behavior of cancel and done. - WQ elements are not deleted until request status is end status.
2. fixes unittest caused by PNN, PSN transition patch
3. fixes unittest caused by replication is no done before the following call is made.

Still need to emulate DBS - currently it calls production dbs for spec validation which can fail if the server is not available.
Also need to emulate ReqMgr2 call to test performCleanupAction correctly.
To makes sure replication is finished on time, add the time.sleep(2). Not completely sure that is optimal number to wait.

All the tests in local are passed at least

### DIFF
--- a/test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py
+++ b/test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py
@@ -13,7 +13,8 @@ import time
 
 from WMCore.Configuration import Configuration
 from WMCore.WorkQueue.WorkQueue import WorkQueue, globalQueue, localQueue
-from WMCore.WorkQueue.WorkQueueExceptions import *
+from WMCore.WorkQueue.WorkQueueExceptions import WorkQueueWMSpecError, \
+     WorkQueueNoMatchingElements, WorkQueueNoWorkError
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue as WorkQueueService
 from WMCore.Lexicon import sanitizeURL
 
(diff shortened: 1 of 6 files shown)


ticoann:
@ericvaandering, @amaltaro, Eric, Alan please review

amaltaro:
Nice one Seangchan! src changes look good to me :)

ericvaandering:
test this please

test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py
```diff
@@ -602,44 +611,43 @@ def testQueueChaining(self):
                                        {})
         self.assertEqual(0, len(self.localQueue))
         self.assertEqual(2, len(work))
-
+
```
ericvaandering: What's going on here? Your editor seems to be inserting a blank space in a lot of empty lines. Can you update the patch stripping all trailing space?

ericvaandering:
test this please

test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py
ticoann: When return key is hit it automatically indent the line with space. I will clean that up

ticoann: Actually it is not  that, it seems it added only one space for each line. Not sure how that happens. Anyway, fixed and updated.

ticoann:
updated some pylint warnings in WorkQueue_t, but didn't fix Emulater pylint warningings since this will be replaced with mock.

PR #6338, opened 2015-10-18
Allow _ as a variable name and anything that starts with dummy

ericvaandering:
Fixed up the variable names that are allowed to be unused:

```
_
dummy
dummySomething
```

anything else will trigger a warning. The idea behind `dummySomething` is that you can easily note the return values of a function without using them, perhaps for future use.


ericvaandering:
Merging this. Will stop some pylint complaints on others.

PR #6351, opened 2015-10-21
add cascade option for announced workflow as well

ticoann:
@amaltaro, Alan please review


amaltaro:
Sounds good to me Seangchan. To reduce the line lenght, how about
if req_status in ["closed-out", "announced"] etc etc?

ticoann:
Thanks Alan, patch is updated as you suggested

PR #6352, opened 2015-10-22
Fix status type view

### LINKED ISSUE #6220: fixing bystatusandtype
could this view emit the status and type, and not the workflow name in addition ?

ticoann:
fixes #6220, Ops script and McM script needs to be changed accordingly, Don't merge until it is agreed upon


ticoann:
@vlimant, Jean-Roch, please review whether this is what you want,

vlimant:
looks like ok to me.
I forgot about the https://github.com/cms-PdmV/stats/blob/master/tools/statsMonitoring.py#L183 but it seems they changed to something else ; can be adapted anyways

ticoann:
replaced by #6359

PR #6354, opened 2015-10-23
some PhEDExInjector improvements

hufnagel:
Two problems addressed in one pull request:

Retrieving the node map fails very often. It doesn't help that both worker thread do this at the same time when they are starting up. Move the retrieval out of the worker threads into the base component and also at least try it 3 times with a few second sleep in between the tries. This also fixes the problem where only one of the worker threads fails in the node map retrieval and the other worker thread keeps "working". Component log looks like the component is working, but it isn't really because half its functionality is dead.

The other issue is frequent Oracle deadlocks in the file status update. Already optimized the transaction handling as much as possible, but it still happens (although less frequent). My suspicion is a race condition between the two worker threads in accessing/modifying the same table/rows in different ways. If this is the case, the proper fix is to get rid of the two worker threads and only use one. Don't want to tackle this now (too big a change and also not sure if there weren't performance reasons why we are using two), but add a single retry after 1s instead. If we are really looking at a race condition that should probably "fix" the problem.

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjector.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjector.py
@@ -8,9 +8,11 @@ DBS.
 
 import threading
 import logging
+import time
 
 from WMCore.Agent.Harness import Harness
-from WMCore.WMFactory import WMFactory
+
+from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
 
 from WMComponent.PhEDExInjector.PhEDExInjectorPoller import PhEDExInjectorPoller
 from WMComponent.PhEDExInjector.PhEDExInjectorSubscriber import PhEDExInjectorSubscriber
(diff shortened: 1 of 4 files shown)


hufnagel:
@ticoann please review

hufnagel:
retest this please

hufnagel:
retest this please

hufnagel:
Ok, it worked finally. @ticoann can you do a review and merge it if you think its ok. I would like to have this included in the next Tier0 release as well.

ericvaandering:
So where there unit tests you were breaking or things that are unstable. If you have found new unstable ones, please update https://github.com/dmwm/WMCore/blob/master/test/etc/UnstableTests.txt.

You can even do a pull request on a simple text file like this directly from the web UI.

hufnagel:
Mostly it was just the unstable tests that gave me problems, but there might have been one more WorkQueue test as well. If I see it again I'll update the list of unstable tests.

ticoann:
@hufnagel, patch looks fine to me. FYI, if one of the thread fails, even though ComponentLog looks fine (error message is already somewhere but log keeps moving), WMStats monitor should catch that error. If it (one of the thread failure) doesn't show in agent info page in wmstats, there must something wrong with the code. Let me know if that is the case.

hufnagel:
You are right, WMStats catches that error. It's just confusing when you look at the log, because the error is not at the end of the log, depending on your log level and the activity it can be quite far back (and hard to find).

PR #6356, opened 2015-10-23
Fixes for the PNN update patch

### LINKED ISSUE #6349: Handle legacy se names from dbs
We need to handle that DBS returns se names in 'origin_site_name'. Until dbs migrates the all the data to PNN after agent is upgrade, we need to handle this.

The following could return se or PNN
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L603
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L791

### LINKED ISSUE #6350: WorkQueue block location is set for  PNN or PSN
I am not sure this is real issue. But looking at the code it seems sometime the location is set to PSN (when trusted site is set) other cases PNN.  I wonder why this doesn't cause the problem yet. (*)
I think all needs to change to PSN

(*)
PSN:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/Block.py#L44
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/Block.py#L185

I think following just have an error, shouldn't this return PNN not se names? (Although current dbs has se names so changing this might actually break things but we have to support both PNN, se)
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/Block.py#L46
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/ResubmitBlock.py#L109
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/ResubmitBlock.py#L141
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/ResubmitBlock.py#L164

PNN: - this need to be convert to PSN
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/Block.py#L187

If element containing site as location followings need to be changed as well.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueue.py#L400

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueue.py#L421

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueue.py#L440

alexanderrichards:
These commts should fix #6350 and also fix #6349

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -5,6 +5,7 @@ _DBSReader_
 Readonly DBS Interface
 
 """
+import re
 from collections import defaultdict
 from dbs.apis.dbsClient import DbsApi
 from dbs.exceptions.dbsClientException import *
@@ -14,6 +15,7 @@ from WMCore.Services.EmulatorSwitch import emulatorHook
 
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
 from WMCore.Lexicon import cmsname, slicedIterator
+pnn_regex = re.compile(r'^T[0-3%]((_[A-Z]{2}(_[A-Za-z0-9]+)*)?)')
 
 def remapDBS3Keys(data, stringify = False, **others):
     """Fields have been renamed between DBS2 and 3, take fields from DBS3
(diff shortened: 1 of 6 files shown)


src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -14,6 +15,7 @@
 
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
 from WMCore.Lexicon import cmsname, slicedIterator
+pnn_regex = re.compile(r'^T[0-3%]((_[A-Z]{2}(_[A-Za-z0-9]+)*)?)')
```
alexanderrichards: @ericvaandering I have reverted this pattern to match the one for cmsname in https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L129

So that we can have these fixes in place I include it here as a module const until we think of a better approach. We can't use the method from Lexicon unless we want to be catching AssertionErrors everywhere., see https://github.com/alexanderrichards/WMCore/commit/a25af89d1eb6c1ac003672907db105a0d6b023ed#commitcomment-13953990 for the discussion

ericvaandering: Sounds good. Maybe a comment there that the same string appears in Lexicon so we know to change one if we change the other? Also a comment on this. I guess T1_UK passes this? Maybe we have that for wild carding? Anyhow, as this re does not allow "." I think we are safe.

ticoann:
closing this add some more fixes and created #6358

PR #6361, opened 2015-10-27
fix Dataset_t

ticoann:
I am not completely sure why this is broken it seems emulator returns only with run number 1 so I had to set the black list to run nubmer 2.
@amaltaro, Alan, could you review


amaltaro:
Jenkins did not report any changes to that test. I may have missed this test in this change:
https://github.com/dmwm/WMCore/pull/6331/files

I'll take a look at it tomorrow morning.

amaltaro:
Ok, just run these tests on master and on this branch and indeed it fixes a test. Thanks Seangchan!

PR #6327, opened 2015-10-12
Added Pileup support to MCFromGEN

amaltaro:
Fixes #6219 which was already available in MonteCarlo but not in MonteCarloFromGEN.

I also added the deterministicPileup capability to these 2 specs and TaskChain.

@ticoann do NOT merge it yet. I ran basic tests in my VM and workload gets properly filled, I just need to check a couple other things (and unittests + pylint).

TODO: I noticed EventBased splitting is not capable to handle it yet.. I will create another issue to get it fixed there as well.

PR #6362, opened 2015-10-28
Added Pileup support to MCFromGEN - 1.0.9 version

amaltaro:
1.0.9 version of #6327 (without code cleanup)


ticoann:
this is against branch unittest won't pass anyway. merging it

PR #6363, opened 2015-10-28
Fix data location for dataset case - 1.0.9 version

amaltaro:
Part of the fix made in #6331 (+ another fix).


ticoann:
this is against branch unittest won't pass anyway. merging it

PR #6364, opened 2015-10-28
Fix a fix for DataLocation

amaltaro:
Fixes a bug introduced by me in #6331. This is needed for current testbed, sorry :(


ticoann:
@amaltaro, Hi Alan, was there a bug? It seems it was just variable name changes.

amaltaro:
there was :( I was accessing dataItem in L154 while the for var was item. Sorry about that.
This is good to be merged btw.

ticoann:
Ah, yes, I miss that. Thanks Alan and don't be sorry. I make a lot more mistakes. :smile:

PR #6368, opened 2015-10-30
add capability for highio jobs

hufnagel:
Switch merge, cleanup and logcollect jobs to highio jobs. By default this is off, but it can be enabled for a select list of sites in the JobSubmitter config. I want to include this for T0_CH_CERN by default as all Tier0 and WMAgent operations there should make use of highIOjobs. What's the best way to do this ?
Add it in WMCore/etc/WMAgentConfig.py ?

@ticoann any advice ?

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -259,11 +259,11 @@ config.JobArchiver.namespace = "WMComponent.JobArchiver.JobArchiver"
 config.JobArchiver.componentDir  = config.General.workDir + "/JobArchiver"
 config.JobArchiver.pollInterval = 240
 config.JobArchiver.logLevel = globalLogLevel
+config.JobArchiver.numberOfJobsToCluster = 1000
 # This is now OPTIONAL, it defaults to the componentDir
 # HOWEVER: Is is HIGHLY recommended that you do NOT run this on the same
 # disk as the JobCreator
 #config.JobArchiver.logDir = config.General.workDir + "/JobArchives"
-config.JobArchiver.numberOfJobsToCluster = 1000
 
 config.component_("TaskArchiver")
 config.TaskArchiver.namespace = "WMComponent.TaskArchiver.TaskArchiver"
@@ -554,8 +554,3 @@ config.ArchiveDataReporter.pollInterval = 300
 config.ArchiveDataReporter.WMArchiverURL = None
 config.ArchiveDataReporter.numDocsRetrievePerPolling = 1000 # number of documents needed to be polled each time
 config.ArchiveDataReporter.numDocsUploadPerCall = 200 # number of documents upload each time in bulk to WMArchive
-# This is now OPTIONAL, it defaults to the componentDir
-# HOWEVER: Is is HIGHLY recommended that you do NOT run this on the same
-# disk as the JobCreator
-#config.JobArchiver.logDir = config.General.workDir + "/JobArchives"
-config.JobArchiver.numberOfJobsToCluster = 1000
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -14,7 +14,6 @@ import threading
 import os.path
 import cPickle
 
-# WMBS objects
 from WMCore.DAOFactory        import DAOFactory
 from WMCore.WMExceptions      import WM_JOB_ERROR_CODES
 
@@ -25,7 +24,7 @@ from WMCore.DataStructs.JobPackage            import JobPackage
 from WMCore.FwkJobReport.Report               import Report
 from WMCore.WMException                       import WMException
 from WMCore.BossAir.BossAirAPI                import BossAirAPI
-#from WMCore.Services.SiteDB.SiteDB            import SiteDBJSON as SiteDB
+
 
 def siteListCompare(a, b):
     """
@@ -363,6 +362,11 @@ class JobSubmitterPoller(BaseWorkerThread):
                 numberOfCores = getattr(baggage, "numberOfCores", 1)
             loadedJob['numberOfCores'] = numberOfCores
 
+            # check if job should be submitted as highIO job
+            highIOjob = False
+            if newJob['type'] in ["Repack", "Merge", "Cleanup", "LogCollect"]:
+                highIOjob = True
+
             # Now that we're out of that loop, put the job data in the cache
             jobInfo = (jobID,
                        newJob["retry_count"],
@@ -387,7 +391,8 @@ class JobSubmitterPoller(BaseWorkerThread):
                        newJob['task_id'],
                        loadedJob.get('inputDataset', None),
                        loadedJob.get('inputDatasetLocations', None),
-                       loadedJob.get('allowOpportunistic', False))
+                       loadedJob.get('allowOpportunistic', False),
+                       highIOjob)
 
             self.jobDataCache[workflowName][jobID] = jobInfo
 
@@ -708,7 +713,8 @@ class JobSubmitterPoller(BaseWorkerThread):
                                'taskID' : cachedJob[20],
                                'inputDataset' : cachedJob[21],
                                'inputDatasetLocations' : cachedJob[22],
-                               'allowOpportunistic' : cachedJob[23]}
+                               'allowOpportunistic' : cachedJob[23],
+                               'highIOjob' : cachedJob[24]}
 
                     # Add to jobsToSubmit
                     jobsToSubmit[package].append(jobDict)
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -385,72 +385,75 @@ class PyCondorPlugin(BasePlugin):
         # unordered list of jobs with random sandboxes.
         # We intend to sort them by sandbox.
 
-        # We also need to do a secondary sort by number of cores
-        # because we can have jobs with different number of cores
-        # and the same sandbox and they cannot be submitted together.
+        # Do a secondary sort for numberOfCores and also for
+        # highIO jobs, they need to be submitted separately.
 
         submitDict = {}
         nSubmits = 0
         for job in jobs:
             sandbox = job['sandbox']
-            numberOfCores = job.get('numberOfCores', 1)
-            if not sandbox in submitDict.keys():
+            numberOfCores = job['numberOfCores']
+            highIOjob = job['highIOjob']
+            if sandbox not in submitDict.keys():
                 submitDict[sandbox] = {}
-            if not numberOfCores in submitDict[sandbox]:
-                submitDict[sandbox][numberOfCores] = []
-            submitDict[sandbox][numberOfCores].append(job)
+            if numberOfCores not in submitDict[sandbox]:
+                submitDict[sandbox][numberOfCores] = {}
+            if highIOjob not in submitDict[sandbox][numberOfCores]:
+                submitDict[sandbox][numberOfCores][highIOjob] = []
+
+            submitDict[sandbox][numberOfCores][highIOjob].append(job)
 
 
         # Now submit the bastards
         queueError = False
         for sandbox in submitDict.keys():
             for numberOfCores in submitDict[sandbox].keys():
-                jobList = submitDict[sandbox][numberOfCores]
-                if queueError:
-                    # If the queue has failed, then we must not process
-                    # any more jobs this cycle.
-                    break
-                for jobsReady in grouper(jobList, self.jobsPerWorker):
-                    idList = [x['id'] for x in jobsReady]
-                    jdlList = self.makeSubmit(jobList=jobsReady)
-                    if not jdlList:
-                        # Then we got nothing
-                        logging.error("No JDL file made!")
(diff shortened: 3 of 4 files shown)


amaltaro:
Yes Dirk, WMCore/etc/WMAgentConfig.py  is the place to set an standard value for this new parameter.

BTW, what this highIO job will represent to the system/glidein? Will it be used by condor to know how many highIO payloads are inside a pilot? Why setting it to a some sites and not to others?

hufnagel:
Only the Tier0 sites support it right now, its a property of the factory configuration for a site.

hufnagel:
If you send highIO jobs to a site that doesn't have them, they'll never run.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -385,76 +385,79 @@ def submit(self, jobs, info=None):
...
             numberOfCores = job.get('numberOfCores', 1)
+            highIOjob = job.get('highIOjob', False)
             if not sandbox in submitDict.keys():
                 submitDict[sandbox] = {}
             if not numberOfCores in submitDict[sandbox]:
-                submitDict[sandbox][numberOfCores] = []
-            submitDict[sandbox][numberOfCores].append(job)
+                submitDict[sandbox][numberOfCores] = {}
+            if not highIOjob in submitDict[sandbox][numberOfCores]:
+                submitDict[sandbox][numberOfCores][highIOjob] = []
```
amaltaro: Why don't you keep it a flat list instead of making it a dict with False/True key name? This way we can get rid of of an additional for loop in the code below.

Another comment on this block, if sandbox is not in the dict, then numberOfCores won't be there as well (nor highIO, if you want to use it), so it could be replaced by:
if not sandbox in submitDict.keys():
    submitDict[sandbox] = {numberOfCores: []}

hufnagel: Tried to keep it flat, but the result looks hacky, so I just added another level to the dict.

Don't understand the second comment. This is a sort, I am checking numberOfCores for each sandbox and highIOjob for each sandbox/numberOfCores combination.

hufnagel:
We might want to hold off on this for now as its unclear if highio jobs will be used in the form they exist now. Need to talk to glideInWMS developers and see how we best implement this. We might use this differently.

ticoann:
@hufnagel, Hi Dirk, what is the status of this patch?

hufnagel:
Patch is on hold, we are waiting for changes on the glideInWMS side and then have to check if there are any changes needed in this patch (which is a bit unclear until we see the new implementation in glideInWMS).

amaltaro:
It seems Jenkins are not happy with this PR.
@hufnagel shall we close it and then you create a new PR when you have consolidate how we proceed (it needs rebase anyways)

hufnagel:
It is rebased, there are no conflicts anymore. Lets see what jenkins does now.

hufnagel:
Ok, jenkins is happy and my first Tier0 tests worked as well. I can confirm that highio jobs now behave like regular single core jobs if the resource does not provide highio job slots (T0_CH_CERN). Next test is on T2_CH_CERN_AI, but I'll wait until tomorrow with this so Farrukh can keep an eye on things while the test is running.

hufnagel:
test this please

hufnagel:
Why is it complaining about a file that isn't even in the patch ?!?

ticoann:
I think it is unrelated error.

ticoann:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
Unit test failures are bogus. @ticoann @amaltaro can you review the code ?

I am going to remove the sitelist, ie. there is no reason not to have all Repack, Merge, LogCollect and Cleanup jobs be highio jobs everywhere and always. Just wanted to hear from you if you prefer testing this code version first or if I should remove it right away here or if I should do that later.

hufnagel:
A small minimal non-functional change to adhere to code conventions that were in place already in PyCondorPlugin. Not going to change anything, but will trigger another jenkins check.

etc/WMAgentConfig.py
```diff
@@ -210,6 +210,7 @@
 config.JobSubmitter.jobsPerWorker = 100
 config.JobSubmitter.maxJobsPerPoll = 1000
 config.JobSubmitter.cacheRefreshSize =10000 # minimum size of the cache before refresh. prevent frequent cache update
+config.JobSubmitter.highIOjobSites = [ "T0_CH_CERN", "T2_CH_CERN_AI" ]
```
amaltaro: Talking in adhere to python conventions :-), you can start by removing the trailing whitespaces inside the list.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -363,6 +363,13 @@ def refreshCache(self):
                 numberOfCores = getattr(baggage, "numberOfCores", 1)
             loadedJob['numberOfCores'] = numberOfCores
 
+            # check if job should be submitted as highIO job
+            highIOjob = False
+            if self.highIOjobSites and isinstance(self.highIOjobSites, list):
+                if potentialLocations.issubset(set(self.highIOjobSites)) \
+                        and newJob['type'] in [ "Repack", "Merge", "Cleanup", "LogCollect" ]:
```
amaltaro: trailing whitespaces here as well

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -385,72 +385,75 @@ def submit(self, jobs, info=None):
...
             sandbox = job['sandbox']
             numberOfCores = job.get('numberOfCores', 1)
+            highIOjob = job.get('highIOjob', False)
             if not sandbox in submitDict.keys():
                 submitDict[sandbox] = {}
             if not numberOfCores in submitDict[sandbox]:
-                submitDict[sandbox][numberOfCores] = []
-            submitDict[sandbox][numberOfCores].append(job)
+                submitDict[sandbox][numberOfCores] = {}
+            if not highIOjob in submitDict[sandbox][numberOfCores]:
```
amaltaro: we better start using `if highIOjob not in submitDict[sandbox][numberOfCores]:`. Can you fix the two ifs above as well?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: Depending on how hacky it gets, it's better than increasing the complexity of this component from O(n^2) to O(n^3). We already have a limit on the number of jobs we can submit in parallel due to the poor performance

amaltaro:
BTW, what's the definition of a highIOslot? How many do we have per worker node?

In addition to that, are you sorting by highIO flag (as well) to avoid too much auto-clustering? Or there is another reason for that?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: There will always be a complexity of O(n^3) now, this is unavoidable. You cannot submit jobs in parallel that differ in sandbox, numberOfCores or the highIO flag. So you need to sort your jobs according to all three criteria. In practice it shouldn't be so bad though because the parameters your sort by are (sometimes highly) correlated. Like almost all highIO jobs are single core (except for multi core repacking in the Tier0, which is rare).

hufnagel: Also, the balance from non highIO to highIO jobs should be (much) more than 10 to 1, so again this means the performance impact should be small.

hufnagel:
From none (most sites now as they don't configure for it) to as many as you want. For the usual 8 core multicore slots we want to configure one for the Tier0, this will be enough to run repack, merge, logcollect and cleanup. It's also not a fixed slot that exists all the time, it's created on demand from the partitionable pilot for jobs that have the highIO flag. If the site have it configured, the pilot will run highIO flagged jobs as normal single core jobs (no difference to now).

I am sorting by highIO flag because I thought the Requestioslots ClassAdd

https://github.com/hufnagel/WMCore/blob/highio-jobs/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py#L979

applied to all jobs in a parallel submit. If this isn't true and the ClassAdd can apply to some and not other jobs in the parallel submit, then I could also avoid the extra sort.

hufnagel:
Fixed the minor issues commented on. Can't do anything about the triple sort, as I expected, if I don't sort by highIO flag, all jobs in a bulk submit will get the flag as long as a single job in the group has it (which is not what we want).

hufnagel:
test this please

hufnagel:
Ok, final review and still the question on whether I should remove the site restriction immediately or you want to test it first with the ability to restrict this to certain sites.

hufnagel:
Discussed it with @amaltaro , will remove the current site limits.

hufnagel:
Ok, so this latest version is working too, just tested it. Therefore I am asking this to be merged and tested at larger scales. I want to get this into the next Tier0 version and we also need it in all production agents (or all agents that can send jobs to T0_CH_CERN, which is or will be the same thing).

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -970,6 +973,11 @@ def customizePerJob(self, job):
         if job.get('inputDatasetLocations'):
             jdl.append('+DESIRED_CMSDataLocations = "%s"\n' % ','.join(job['inputDatasetLocations']))
 
+        # HighIO jobs
+        highIOjob = job.get('highIOjob')
+        if highIOjob:
+            jdl.append('+Requestioslots = 1\n')
```
ticoann: So it is trying to match high io slot first among the possible site then if there no site with high io slots it will still match the other possible sites, correct?

hufnagel: I have no idea actually what it'll do if there are multiple sites it can match, some with and some without highIO job slots. In practical terms, if you look at the job types this is active for, the possible sites they can run at is only a single site anyways, so not sure it matters what the exact behavior is. At least I can't think of a reason why we would care where it matches first in such a case.

ticoann: Ah,  your right. Those types has to have one location. I am just wondering whether it still can get submitted site without highIO sites (if that one site doesn't have highIO slot).

hufnagel: I tested that case with a single site. Works fine, jobs run as if the flag isn't set (ie. the pilot ignores the flag if it doesn't have highIO job slots).

hufnagel: Now admittedly I haven't tried matching a single core highIO job to a single core pilot. It matches fine as a standard single core job to a multi core partitionable pilot without highIO job slots though.

amaltaro:
@hufnagel I'm applying this patch to a testbed agent, we run a couple of tests and if everything is fine, Seangchan will merge and cut a new tag with it. Is that ok?

hufnagel:
Yes, that's fine.

amaltaro:
Sent those workflows to the same node with the resize patch and lots of jobs went into held status... starting tests again in another node (vocms0231)

amaltaro:
Ok, I ran around 10 workflows at T2_CH_CERN and FNAL and I there were no issues, I think it's good to be merged then. We'll exercise it a bit more in the next WMAgent release anyways.

amaltaro:
@ticoann please merge

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -363,6 +362,11 @@ def refreshCache(self):
                 numberOfCores = getattr(baggage, "numberOfCores", 1)
             loadedJob['numberOfCores'] = numberOfCores
 
+            # check if job should be submitted as highIO job
+            highIOjob = False
+            if newJob['type'] in ["Repack", "Merge", "Cleanup", "LogCollect"]:
```
amaltaro: Dirk, should we consider 'Harvesting' jobs as highio as well?

hufnagel: Was thinking about that but decided against it. They really don't use much IO and can use non-negligible cpu, so I decided they are better off as standard single core jobs.

amaltaro: Yes, in the end I believe it's a case by case. Super big workflows would create a IObound harvest job. Small ones wouldn't. Anyways it's merged already LoL

hufnagel: Even for large workflows reading in DQMIO should be quick. Plus you still spend some cpu combining he DQMIO input into a single small output file. So the IO, if it is there, is mostly on the storage element, not on the local disk. So not really an issue.

PR #6369, opened 2015-10-30
Fix typo in SiteDB

amaltaro:
WorkQueue in testbed is failing to digest an ACDC workflow and the reason is a typo in a list method (extendS).
@ticoann lets wait for jenkins and then cut a new release.


amaltaro:
Damnit! It's Friday... (I never thought I'd say something like that :P) I'll ask Diego and Bruno to check whether they can patch workqueue in testbed for the weekend

ticoann:
Thanks Alan, how come we missed this, I think we need to change SiteDB unittest to test this. (Currently using emulator.) I will fix that.

amaltaro:
That patch was way too big, almost impossible to catch all the cases :-)

PR #6372, opened 2015-10-30
add BlockWhitelist validation

ticoann:
(need further test - do not include in Nov tag) @amaltaro, Alan please review


amaltaro:
Looks good! How about adding the same check for BlockBlacklist?

PR #6383, opened 2015-11-05
dbsbuffer changes to work around race condition between dbs and phedex updates

hufnagel:
Both PhEDExInjector and DBS3Buffer components update dbsbuffer_file records. They do it in a way that the two updates can deadlock each other. General slow performance of dbsbuffer related code makes this quite likely to happen, therefor rework and simplify the schema (add some indexes, remove triggers, general cleanup).

No functional changes and the cleanup and indexes are only done on Oracle for now as some of it is very Oracle specific. We could sync the MySQL version later if so desired.

Also put in recovery logic to at least retry once in each component if deadlock occurs.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
@@ -89,7 +89,7 @@ def uploadWorker(workInput, results, dbsUrl):
     """
 
     # Init DBS Stuff
-    logging.debug("Creating dbsAPI with address %s" % dbsUrl)
+    logging.debug("Creating dbsAPI with address %s", dbsUrl)
     dbsApi = DbsApi(url = dbsUrl)
 
 
@@ -112,7 +112,7 @@ def uploadWorker(workInput, results, dbsUrl):
 
         # Do stuff with DBS
         try:
-            logging.debug("About to call insert block with block: %s" % block)
+            logging.debug("About to call insert block with block: %s", block)
             dbsApi.insertBulkBlock(blockDump = block)
             results.put({'name': name, 'success': "uploaded"})
         except Exception as ex:
@@ -120,9 +120,9 @@ def uploadWorker(workInput, results, dbsUrl):
             if 'Block %s already exists' % name in exString:
                 # Then this is probably a duplicate
                 # Ignore this for now
-                logging.error("Had duplicate entry for block %s. Ignoring for now." % name)
-                logging.debug("Exception: %s" % exString)
-                logging.debug("Traceback: %s" % str(traceback.format_exc()))
+                logging.error("Had duplicate entry for block %s. Ignoring for now.", name)
+                logging.debug("Exception: %s", exString)
+                logging.debug("Traceback: %s", str(traceback.format_exc()))
                 results.put({'name': name, 'success': "uploaded"})
             elif 'Proxy Error' in exString:
                 # This is probably a successfully inserton that went bad.
@@ -136,7 +136,7 @@ def uploadWorker(workInput, results, dbsUrl):
                 msg += exString
                 logging.error(msg)
                 logging.error(str(traceback.format_exc()))
-                logging.debug("block: %s \n" % block)
+                logging.debug("block: %s \n", block)
                 results.put({'name': name, 'success': "error", 'error': msg})
 
     return
@@ -320,8 +320,8 @@ class DBSUploadPoller(BaseWorkerThread):
         Find all blocks; make sure they're in the cache
         """
         openBlocks = self.dbsUtil.findOpenBlocks()
-        logging.info("Found %d open blocks." % len(openBlocks))
-        logging.debug("These are the openblocks: %s" % openBlocks)
+        logging.info("Found %d open blocks.", len(openBlocks))
+        logging.debug("These are the openblocks: %s", openBlocks)
 
         # Load them if we don't have them
         blocksToLoad = []
@@ -332,14 +332,14 @@ class DBSUploadPoller(BaseWorkerThread):
         # Now load the blocks
         try:
             loadedBlocks = self.dbsUtil.loadBlocks(blocksToLoad)
-            logging.info("Loaded blocks: %s" % loadedBlocks)
+            logging.info("Loaded blocks: %s", loadedBlocks)
         except WMException:
             raise
         except Exception as ex:
             msg =  "Unhandled exception while loading blocks.\n"
             msg += str(ex)
             logging.error(msg)
-            logging.debug("Blocks to load: %s\n" % blocksToLoad)
+            logging.debug("Blocks to load: %s\n", blocksToLoad)
             raise DBSUploadException(msg)
 
         for blockInfo in loadedBlocks:
@@ -352,14 +352,14 @@ class DBSUploadPoller(BaseWorkerThread):
             # Now we have to load files...
             try:
                 files = self.dbsUtil.loadFilesByBlock(blockname = blockname)
-                logging.info("Have %i files for block %s" % (len(files), blockname))
+                logging.info("Have %i files for block %s", (len(files), blockname))
             except WMException:
                 raise
             except Exception as ex:
                 msg =  "Unhandled exception while loading files for existing blocks.\n"
                 msg += str(ex)
                 logging.error(msg)
-                logging.debug("Blocks being loaded: %s\n" % blockname)
+                logging.debug("Blocks being loaded: %s\n", blockname)
                 raise DBSUploadException(msg)
 
             # Add the loaded files to the block
@@ -394,7 +394,7 @@ class DBSUploadPoller(BaseWorkerThread):
                 msg =  "Unhandled exception while loading uploadable files for DatasetPath.\n"
                 msg += str(ex)
                 logging.error(msg)
-                logging.debug("DatasetPath being loaded: %s\n" % datasetpath)
+                logging.debug("DatasetPath being loaded: %s\n", datasetpath)
                 raise DBSUploadException(msg)
 
             # Sort the files and blocks by location
@@ -581,8 +581,8 @@ class DBSUploadPoller(BaseWorkerThread):
                 msg =  "Unhandled exception while writing new blocks into DBSBuffer\n"
                 msg += str(ex)
                 logging.error(msg)
-                logging.debug("Blocks for DBSBuffer: %s\n" % createInDBSBuffer)
-                logging.debug("Blocks for Update: %s\n" % updateInDBSBuffer)
+                logging.debug("Blocks for DBSBuffer: %s\n", createInDBSBuffer)
+                logging.debug("Blocks for Update: %s\n", updateInDBSBuffer)
                 raise DBSUploadException(msg)
             else:
                 myThread.transaction.commit()
@@ -608,7 +608,7 @@ class DBSUploadPoller(BaseWorkerThread):
                 msg =  "Unhandled exception while setting blocks in files.\n"
                 msg += str(ex)
                 logging.error(msg)
-                logging.debug("Files to Update: %s\n" % self.filesToUpdate)
+                logging.debug("Files to Update: %s\n", self.filesToUpdate)
                 raise DBSUploadException(msg)
             else:
                 myThread.transaction.commit()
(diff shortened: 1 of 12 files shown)


hufnagel:
As @ticoann is on vacation this week, can you start reviewing this @amaltaro ?

As I said, there shouldn't be any functional changes. What I did change:

 - moving primary key and unique constraints into the CREATE TABLE call (easier to read)
 - adding indexes on all FK columns
 - removing most triggers and adjusting the INSERT calls accordingly
 - remove some weird duplicate catching code/sql in NewAlgo, do it the "standard" way

Also put in a retry for the calls that can fail with the deadlock, one after a 5s sleep, the other after a 10s sleep.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -718,16 +718,28 @@ def retrieveBlocks(self):
...
-                myThread.transaction.rollback()
-                raise
             except Exception as ex:
                 myThread.transaction.rollback()
-                msg =  "Unhandled exception while finished closed blocks in DBSBuffer\n"
-                msg += str(ex)
-                logging.error(msg)
-                logging.debug("Blocks for Update: %s\n" % loadedBlocks)
-                raise DBSUploadException(msg)
+                # possible deadlock with PhEDExInjector, retry once after 10s
```
amaltaro: how about adding a logging.warning here just to log a simple message (and saying it will be retried in 10secs)?

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -718,16 +718,28 @@ def retrieveBlocks(self):
...
+                                            transaction = True)
+                except Exception as ex:
+                    myThread.transaction.rollback()
+                    msg =  "Unhandled exception while finished closed blocks in DBSBuffer\n"
+                    msg += str(ex)
+                    logging.error(msg)
+                    logging.debug("Blocks for Update: %s\n", loadedBlocks)
+                    raise DBSUploadException(msg)
+                else:
+                    myThread.transaction.commit()
```
amaltaro: this line could be completely outside of the try/except (and the 2 lines below removed). Since in case of exception it will raise it upstream.

src/python/WMComponent/DBS3Buffer/Oracle/NewDataset.py
```diff
@@ -12,6 +12,13 @@
 
 class NewDataset(MySQLNewDataset):
 
-    sql = """INSERT INTO dbsbuffer_dataset (path, processing_ver, acquisition_era, valid_status, global_tag, parent, prep_id)
-               SELECT :path, :processing_ver, :acquisition_era, :valid_status, :global_tag, :parent, :prep_id FROM DUAL
-               WHERE NOT EXISTS (SELECT path FROM dbsbuffer_dataset WHERE path = :path)"""
+    sql = """INSERT INTO dbsbuffer_dataset
+             (id, path, processing_ver, acquisition_era, valid_status, global_tag, parent, prep_id)
+             SELECT dbsbuffer_dataset_seq.nextval, :path, :processing_ver, :acquisition_era,
```
amaltaro: Stupid question, doesn't the database automatically creates the next sequential id? Or is it redundant?

src/python/WMComponent/DBS3Buffer/MySQL/NewAlgo.py
```diff
@@ -11,19 +11,11 @@
 from WMCore.Database.DBFormatter import DBFormatter
 
 class NewAlgo(DBFormatter):
-    """
-    _NewAlgo_
 
-    Check for existence of algo, then add if not exists
-    """
-    existsSQL = """SELECT id FROM dbsbuffer_algo WHERE app_name = :app_name AND
```
amaltaro: This code will break if we try to insert a set of values that already exist in the database. Any reason to remove this check?

amaltaro:
I made a few comments/questions. Besides the removal of the check for an algo, code looks good, though I did no look at every line.

Just a suggestion for the next PRs, try to split pylint fixes in a different commit please. It makes the reviewer's life much easier and a more reliable review ;)

amaltaro:
> No functional changes and the cleanup and indexes are only done on Oracle for now as some of it is very Oracle specific. We could sync the MySQL version later if so desired.

The best would be to sync them in this PR, otherwise it will probably be forgotten. I can help you validating the mysql changes.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
hufnagel: Good idea, will do that.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
hufnagel: Good idea, will do that.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
hufnagel: I could do that, but with the else you make it clear in the code that this part of the whole try/except construct. I think it makes the whole thing a bit easier to read.

src/python/WMComponent/DBS3Buffer/Oracle/NewDataset.py
hufnagel: Not on Oracle. There you create a sequence, but you have to either use them in the insert or (what was done before here) define an Oracle trigger that does it for you on each insert.

src/python/WMComponent/DBS3Buffer/MySQL/NewAlgo.py
hufnagel: No, it won't. That's what we have the INSERT IGNORE for. We use INSERT IGNORE everywhere else in MySQL inserts, no idea why it had a special check here. Maybe because this code was ancient.

hufnagel:
> Just a suggestion for the next PRs, try to split pylint fixes in a different commit please. It makes the reviewer's life much easier and a more reliable review ;)

I don't like it because it makes it harder to rebase further changes later. Example: commit 1 is functional, commit 2 is pylint, commit 3 is more functional. I have no idea how to squash 3 into 1 at that stage). Anyone have a suggestion how to do this ?

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
amaltaro: Arguable :-) I'm all in favor of avoiding duplicate code as much as possible (even 2 lines only).

src/python/WMComponent/DBS3Buffer/MySQL/NewAlgo.py
amaltaro: I told you it was a stupid question ;)

amaltaro:
I think you cannot, you could squash 3 into 2, but not 3 into 1. Anyways, a new PR for the mysql part only won't be a problem IMO.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
hufnagel: Well, that whole section of code is one big duplication because we do the retry if it fails in the exception. I rather duplicate the whole blocks of code here because this is ugly to start with, rather show the complete functional blocks to show what I am doing.

Another reason I want to leave it in is that the else here isn't needed due to the raise in the except. But there can be cases where you don't raise (just log the error and move on), then you absolutely need the else.

I am using this as a general purporse try/except/else construct everywhere in my code.

hufnagel:
This won't get merged anyways until Seangchan is back. Will see what I can with the MySQL changes. I have no way to test them though, but I guess jenkins will do that for me.

hufnagel:
@amaltaro , please review. MySQL changes are in and they also triggered some minor changes in the Oracle schema to get them in sync. Basically primary keys and unique constraints are always defined in the CREATE TABLE statement itself, any FK constraint was moved out to the separate ALTER TABLE statement. I did the latter because if you have the FK constraints in the table definition then you become sensitive to the order you create tables at (as they depend on each other).

Added all the new Oracle indexes also to MySQL.

hufnagel:
@ticoann @amaltaro is there a reason this isn't merged yet ?

ticoann:
Hi Dirk, we are running some other test for the wmagent new version. I will include this patch for the quick test once current test is done. I will merge it in a couple of days. Sorry for the delay.

hufnagel:
No problem. We use this in the Tier0, just wanted to know what's going on as there wasn't any activity for a while.

amaltaro:
@hufnagel are you using this code in Tier0?
I deployed the newest agent today (1.0.12.pre1) and I'm getting oracle issues (cannot insert NULL into) in a couple components. Didn't you see these issues in T0?

The WorkQueueManager error is:
```
(IntegrityError) ORA-01400: cannot insert NULL into ("CMS_WMBS_PREPROD3"."DBSBUFFER_ALGO"."ID")
 'INSERT INTO dbsbuffer_algo (app_name, app_ver, app_fam, pset_hash,\n                                         config_content, in_dbs)\n               SELECT :app_name, :app_ver, :app_fam, :pset_hash,\n                 :config_content, 0 FROM DUAL WHERE NOT EXISTS\n                   (SELECT * FROM dbsbuffer_algo WHERE app_name = :app_name AND\n                      app_ver = :app_ver AND app_fam = :app_fam AND\n                      pset_hash = :pset_hash)' {'config_content': 'Unknown', 'app_fam': 'Unknown', 'app_ver': 'Unknown', 'pset_hash': 'Unknown', 'app_name': 'cmsRun'}
```

and the JobACcountant also fails with:
```
(IntegrityError) ORA-01400: cannot insert NULL into ("CMS_WMBS_PREPROD3"."DBSBUFFER_FILE_PARENT"."PARENT")
 'INSERT INTO dbsbuffer_file_parent (child, parent) VALUES\n               ((SELECT id FROM dbsbuffer_file WHERE lfn = :child),\n               (SELECT id FROM dbsbuffer_file WHERE lfn = :parent))' [{'par
ent': '/store/mc/Fall13/BprimeJetToBZ_M800GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM/POSTLS162_V1-v1/20000/CA435697-9155-E411-B6DB-00266CFFBED8.root', 'child': '/store/backfill/1/DMWM_Test/BprimeJetToBZ_M80
0GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM-RAW/ReDigi_3steps_Agent1012_Validation_TEST_Alan_v1-v1/00000/D0CCC454-6E8F-E511-98D5-003048FFD7BE.root'}, {'parent': '/store/mc/Fall13/BprimeJetToBZ_M800GeV_Tune4
C_13TeV-madgraph-tauola/GEN-SIM/POSTLS162_V1-v1/20000/C89B633D-DE54-E411-BEBB-0025907B50FC.root', 'child': '/store/backfill/1/DMWM_Test/BprimeJetToBZ_M800GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM-RAW/ReDig
i_3steps_Agent1012_Validation_TEST_Alan_v1-v1/00000/D0CCC454-6E8F-E511-98D5-003048FFD7BE.root'}]BindList: [{'parent': '/store/mc/Fall13/BprimeJetToBZ_M800GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM/POSTLS162
_V1-v1/20000/CA435697-9155-E411-B6DB-00266CFFBED8.root', 'child': '/store/backfill/1/DMWM_Test/BprimeJetToBZ_M800GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM-RAW/ReDigi_3steps_Agent1012_Validation_TEST_Alan_v
1-v1/00000/D0CCC454-6E8F-E511-98D5-003048FFD7BE.root'}, {'parent': '/store/mc/Fall13/BprimeJetToBZ_M800GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM/POSTLS162_V1-v1/20000/C89B633D-DE54-E411-BEBB-0025907B50FC.r
oot', 'child': '/store/backfill/1/DMWM_Test/BprimeJetToBZ_M800GeV_Tune4C_13TeV-madgraph-tauola/GEN-SIM-RAW/ReDigi_3steps_Agent1012_Validation_TEST_Alan_v1-v1/00000/D0CCC454-6E8F-E511-98D5-003048FFD7BE.roo
t'}]
```

The interesting thing is that both tables are empty(?)

ticoann:
Hmm, I didn't get this error on my test either.

amaltaro:
Maybe it's something to oracle only? Did you test it in mysql?

ticoann:
Ah, yes I tested in MariaDB

hufnagel:
This is used just fine in the Tier0, no problems there.

I would say the first problem looks like a missing modification for an insert call. I thought I got them all but apparently not.

The second problem seems like broken code to me. It tries to insert file parentage if there is no file parentage, which shouldn't be done.

Can you point at lines of code where the problem occurs, not just the error message ?

amaltaro:
I couldn't identify any problem in your patch. However, in the end it seems to have exposed another problem where we were using a python module (and so dao) from DBSBuffer instead of DBS3Buffer.

I'm testing a patch and will make a PR if that's really stable.

PR #6385, opened 2015-11-06
Add timestamp to stageout; remove ls retries; minor changes

amaltaro:
As reported in the CompOps (and personally by @bbockelm), when a dcache server is under high load the lcg-ls retries may make things even more unstable. So I removed those 10 retries we had before per each lcg-cp attempt.

Timestamp was also added to the stage out code such that it makes our life a bit easier to debug any thing there.

Directory creation was evaluating to True all the time, so I simply removed the code checking for that (and error codes).

Finally, lcg-cp logs will be spit to the _condor_stdout now, even when it's successful.

@hufnagel please have a look.
I successfully ran several tests, but I put a few more changes in this commit that needs one more test at least.

### DIFF
--- a/src/python/WMCore/Storage/Backends/LCGImpl.py
+++ b/src/python/WMCore/Storage/Backends/LCGImpl.py
@@ -95,11 +95,11 @@ class LCGImpl(StageOutImpl):
             options = None
             useCVMFS = True
 
-        copyCommand = "lcg-cp -b -D srmv2 --vo cms --srm-timeout 2400 --sendreceive-timeout 2400 --connect-timeout 300 --verbose "
+        copyCommand = "lcg-cp -b -D srmv2 --vo cms --srm-timeout 2400 --sendreceive-timeout 2400 --connect-timeout 300 --verbose"
         if options != None:
             copyCommand += " %s " % options
         copyCommand += " %s " % sourcePFN
-        copyCommand += " %s \n" % targetPFN
+        copyCommand += " %s 2> stageout.log" % targetPFN
 
         # for CVMFS skip the usual environment setup and use grid.cern.ch instead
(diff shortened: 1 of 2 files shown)


bbockelm:
For what it's worth, it's possible to salvage the retries on the lcg-ls (although since there are retries on the higher level, it doesn't really seem necessary).  It's just that we would need to do an exponential backoff to avoid congestive-collapse-like effects.

But this is good for now, I think.

amaltaro:
test this please

amaltaro:
test this please

hufnagel:
Last time you removed the special directory creation code, it broke some of @juztas tests. Are we sure this won't happen anymore ? Also, you should remove the directoryErrorCodes from

src/python/WMCore/Storage/Plugins/Examples/SRMV2Impl.py

as well because its useless with your changes.

amaltaro:
I don't know which tests Justas usually run and AFAIK he's leaving on vacation today. BTW, if CRAB uses the stage out v2, then these changes should have no impact on them.

About
src/python/WMCore/Storage/Plugins/Examples/SRMV2Impl.py

it is (should be) just an example and it uses StageOutImplV2.py instead, thus this PR should not have any impact.

amaltaro:
updated a few messages trying to make logs a bit better looking

amaltaro:
Hmm, these unittests changes cannot be related to this PR. I just fixed the last 2 warnings that I haven't noticed.

amaltaro:
test this please

amaltaro:
Voila, unittests have finally cleared out. @ticoann it's ready to be merged. I'd like to put it in in the new agent release, so we may have a chat before merging it to discuss how to proceed. Thanks

PR #6386, opened 2015-11-09
Set run number 999999 to multiRun harvesting

### LINKED ISSUE #5643: Name for output root file for multiRun DQM harvesting
I believe the definition of the root file created out of the harvesting jobs does not come from WMAgent, but from cmsDriver command itself used to create the configuration. Anyways recording it here so we can investigate it later on.

I noticed the file name in DQM Gui has only one run number even when we are running "multiRun" harvesting mode, meaning several runs will be present in the same output root file, example:
DQM_V0001_R000202209__JetHT__CMSSW_7_2_0-PRE_R_72_V10A_RelVal_jetHT2012c_TEST_HG1503_Validation_multiRun_v2-v1__DQMIO.root  

If it relies on WMAgent, then we should investigate a proper way to name it.

### LINKED ISSUE #6299: Add a few other fields to workload_cache reqmgr2

amaltaro:
Fixes #5643
Still need to be tested, pylint clean up and a new unit test added. 
Don't merge yet, please.

### DIFF
--- a/src/python/WMCore/JobSplitting/Harvest.py
+++ b/src/python/WMCore/JobSplitting/Harvest.py
@@ -160,6 +160,7 @@ class Harvest(JobFactory):
                 if f not in self.currentJob['input_files']:
                     self.currentJob.addFile(f)
 
+        self.currentJob.addBaggageParameter("multiRun", True)
         if endOfRun:
             self.currentJob.addBaggageParameter("runIsComplete", True)
         self.mergeLumiRange(self.currentJob['mask']['runAndLumis'])
--- a/src/python/WMCore/JobSplitting/Harvest.py
+++ b/src/python/WMCore/JobSplitting/Harvest.py
@@ -4,7 +4,6 @@ _Harvest_
 
 """
 import threading
-import os
 
 from WMCore.JobSplitting.JobFactory import JobFactory
 from WMCore.Services.UUID import makeUUID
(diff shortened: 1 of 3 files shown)


amaltaro:
test this please

amaltaro:
Hold on on this one, it seems there is yet another place that we need to fix (otherwise it looks for the wrong file to upload).

amaltaro:
Do not merge this one Seangchan. I'm waiting for CMSSW/DQM guys to tell me what's wrong on their side.

amaltaro:
I discussed this issue in a HN and it turns to be a bug in cmssw, see
https://hypernews.cern.ch/HyperNews/CMS/get/dqmDevel/2296.html

in summary, a fix is being made against 80x and will be backported to 75x and 76x cycles.
However, it makes very hard to put this feature/fix in in WMCore because it would work for some releases and fail for others, unless we instrument the code to add the CMSSW version as a job baggage and check it during runtime.

Honestly, I don't like that approach. We may just keep it out for a year or two...

amaltaro:
Before I forget, @ericvaandering and I discussed this issue/fix on Friday and we decided not to push this change into WMCore right now. Once people start using multiRun harvesting in production AND they ask us for this feature/fix, then we can handle that.

Keep it ON HOLD please (and let's see for how long :D)

amaltaro:
Closing it since it's fixed in a new PR

PR #6389, opened 2015-11-12
Remove 2 wmstats views that are available in erlang

amaltaro:
These views are already available in erlang (WMStatsErl). I x-checked their results and it's all good.


amaltaro:
@ticoann review and feel free to merge please

PR #6393, opened 2015-11-16
add some more PhysicsSkims

hufnagel:
Needed for pp reference run and the heavy ion run. Tier0 is going to be patched, please merge at your convenience.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -127,7 +127,19 @@ class StdBase(object):
                                        'TopMuEG' : 'RAW-RECO',
                                        'EcalActivity' : 'RAW-RECO',
                                        'CosmicSP' : 'RAW-RECO',
-                                       'CosmicTP' : 'RAW-RECO'
+                                       'CosmicTP' : 'RAW-RECO',
+                                       'ZMM' : 'RAW-RECO',
+                                       'Onia' : 'RECO',
+                                       'HighPtJet' : 'RAW-RECO',
+                                       'D0Meson' : 'RECO',
+                                       'Photon' : 'AOD',
+                                       'ZEE' : 'AOD',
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann , please review and merge (after jenkins is done with it)

amaltaro:
Dirk, out of curiosity, why do you have this T0 specific code in the StdBase class? StdBase should carry common information only. Can't you add it to the PromptReco spec?

hufnagel:
Can't go into PromptReco because it is needed in the determineOutputModules() method.

hufnagel:
Its not Tier0 specific code anyways, its code dealing with Config.DP built configuration. These are actually generic, but we only use them in the Tier0.

hufnagel:
Ok, need to update the list of skims.

amaltaro:
It's not T0 specific but only T0 has ever used it :-) Any production workflow uses configcacheID.
Just to separate settings from the actual code, how about you create another python file with this mapping and simply import it into StdBase? Something like:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/RequestManager/RequestDB/Settings/RequestStatus.py

I'm just trying to simplify it and to make StdBase a real Base schema for all the other requests, without several use cases here.

hufnagel:
Configuration.DataProcessing was always an integral part of WMCore configuration building. All merge jobs use it for instance. This is absolutely needed to make PromptReco Config.DataProcessing work.

Don't really see the point in importing another python file into StdBase. Whether imported or included the code is still going to be there.

In any case, this code is already deployed in the Tier0. I am not going to refactor it at least until the end of data taking this year. If it can't get merged like this then we'll have to wait until then for version 2.0.

amaltaro:
The big advantage is that you change only a "settings" file and don't risk messing around with the code :-)
Anyways, not a big deal and not a reason not to merge it, not from me at least...

hufnagel:
@ticoann , could this get merged before we forget about it ?

PR #6216, opened 2015-09-04
Update to use both psutil 0.6.1 and 3.1.1

ericvaandering:
Fixes #6153

PR #6394, opened 2015-11-16
stop endless spamming cpu alarms

hufnagel:
Seems like the psutil.Process object has no get_cpu_percent method in the python version that we ship with WMAgent (and the Tier0). Not sure if this is a recent change or of this was always broken, but this fixes it. Without it the AlertGenerator goes nuts, sending a cpu alert every polling cycle.

### DIFF
--- a/src/python/WMComponent/AlertGenerator/Pollers/System.py
+++ b/src/python/WMComponent/AlertGenerator/Pollers/System.py
@@ -32,7 +32,7 @@ class ProcessCPUPoller(object):
         """
         try:
             # raises: psutil.error.AccessDenied, psutil.error.NoSuchProcess
-            pollProcess = lambda proc: proc.get_cpu_percent(PeriodPoller.PSUTIL_INTERVAL)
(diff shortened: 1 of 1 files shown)


hufnagel:
@ticoann , please review and merge

ticoann:
Thanks Dirk, that code hasn't changed for a couple of years. I guess the bug always exists. I will merge this when unittest finishes although this shouldn't affect any of the test. 
Anyway, we will retire AlertSystem soon, LogDB will replace AlertSystem.

ericvaandering:
I must have missed this in https://github.com/dmwm/WMCore/pull/6216/files

amaltaro:
Dirk, can you also fix get_memory_percent()? it seems it got replaced by memory_percent(). Thanks

ticoann:
> Dirk, can you also fix get_memory_percent()? it seems it got replaced by memory_percent(). Thanks

#6401 created

amaltaro:
Thanks Seangchan!

Issue #6382, opened 2015-11-04
Migrate wmstats views to erlang


amaltaro:
We've seen some issues with wmstats views in the last days and it's not managing to keep things up-to-date. Migrating its views from javascript to erlang would help us to mitigate these issues.

In addition to that, we have the same views in erlang and in javascript (agentInfo and requestHistory).
We could remove the javascript ones.

amaltaro:
@juztas since you were playing with ASO views, would you like to take this issue? :-)
If not, @BrunoCoimbra is willing to do so.

geneguvo:
Hi guys,
I have some good news about this.

I've checked my old notes from 1 year and a half ago, when we were having problems with the wmstats views still in the old couch, old machines.

At that time I wrote most of the views in erlang (12 out of 16 if I remember correctly), but only moved  to production 8 of those (the other 4 we left to be included later). In the meantime, you guys have cleaned up some views that were not anymore needed. So right now, wmstats has:
```
agentInfo
cooledoffRequests
jobsByStatusWorkflow
latestAgentSite
latestRequest
requestAgentUrl
requestHistory
time
```
From these, agentInfo and requestHistory are already there in erlang and being used, so we should be able  to drop their corresponding javascript version.

From the others, I have cooledoffRequests, latestAgentSite, latestRequest and requestAgentUrl from my old notes. I've double checked and their code haven't changed since then, so we should be ready to use them.

After that, there's only "time" and "jobsByStatusWorkflow" remaining. "time" is easy to implement, however "jobsByStatusWorkflow" seems to be the most complicated one of all the views for the map function. Its reduce function is easy since is the same already in use in other views.

If we could write "jobsByStatusWorkflow" in erlang, then we could drop completely the javascript version, otherwise even if we keep this last one things shall already improve by dropping the others.

Cheers,
Diego.

juztas:
Go ahead @BrunoCoimbra Let me know if you need any help

geneguvo:
I'm doing some performance tests using the current views we already have in erlang vs javascript. I'm using a snapshot of the database we had on the day of the issue in production.

Depending on the results, we may stay with the remaining javascript views. Otherwise, Bruno is going to work on the remaining views.

geneguvo:
I've finalized my performance tests comparing the 6 (out of 8) views  we have written both in javascript and erlang. The erlang views performed ~1.7 times faster than the javascript ones with the wmstats production database we have atm.
[...]

PR #6399, opened 2015-11-18
Add erlang views for wmstats

geneguvo:
This is the set of erlang views we already have written for wmstats. The two remaining ones @BrunoCoimbra is working on.

I tested these erlang views with a snapshot of the production database and did various performance tests. I haven't tested the rewrite rules, though. I just added them similarly to the ones that already existed.

We also need to do corresponding changes for the frontend rules in deployment (https://github.com/dmwm/deployment/blob/master/frontend/app_couch_ssl.conf) once these views have been included in the next reqmon release.

This is work done as agreed in #6382.


geneguvo:
Please review/test and include it in the upcoming release HG1512.

ticoann:
Thanks a lot, Diego. I will run the test. If it works fine we will drop js code completeley.

geneguvo:
Dont drop the js code yet. We need it for the validation of the new views. We should also keep both for a month or two. In case of any problem, we could easily set to use the js views instead.

ticoann:
Thanks Diego, that is good point. I will keep the js code for a while then. I was talking to Alan earlier, though. What happen if we keep js view. I mean isn't it affect the view creation time?  What is the performance increase with erlang over javascript. It is one view creation(update) or view accessing or both?

ticoann:
All view created in erlang gives the same result as js view in my vm test. merging the patch.

PR #6402, opened 2015-11-19
cleanup archieved workflows from various agent couchdb

ticoann:
Test in my vm, @amaltaro, Alan please review.


amaltaro:
Seangchan, what's the difference between a "jobsummary" and a "agent_request" doc?

BTW, how about we set the polling cycle for the CleanCouchPoller much higher now? Just to avoid increasing cmsweb calls too much. Maybe once every 4 hours would be enough (unless it does more stuff that I don't know).

ticoann:
> Seangchan, what's the difference between a "jobsummary" and a "agent_request" doc?

There are two different types of documents stored in wmstats. First one is job summary, second one is summary of requests.

> 
> BTW, how about we set the polling cycle for the CleanCouchPoller much higher now? Just to avoid increasing cmsweb calls too much. Maybe once every 4 hours would be enough (unless it does more stuff that I don't know).

Acutaully it doesn't increase the number of call that much. Just one call is added. (Depending on the number of workflows reside in one agent, it might be a call which require more data transfer but It shouldn't be more than 200K data transfer per each call.

ticoann:
Sorry, previous comments are not entirely true. It could make many more calls when there are a lot of archived workflows within the polling cycle, but I think that is OK. Anyway, if you think polling cycle need to be increased we can do that as well.

amaltaro:
As you said, it's only one view call, so no really a big deal.
However, I'm always in favor of not doing calls when not needed :-D Since this thread doesn't do anything else than cleaning up archived workflows, 4h polling seems quite reasonable to me. Or 6h, or 2h, anything is better than < 600 secs

ticoann:
I think it is depends on how often workflow is archived if the workflow archived more or less uniformly every hour we should clean up every hour. If workflow archived every 4 hours we should clean up every 4 hours. However I don't think the calls are too costly compare to cleaning up a lot of things at one time. 

Anyway, I tested in my vm. Please review. If you prefer we can still increase the polling cycle. Also it seems this wasn't that big of the problem since there was deleting function called by time limit and wmstats only keeps the error jobs, it shouldn't affect the replication much even without this patch.
The benefit of this patch is cleaning up job database when workflow is archived. (But those db is not replicated anyway).

amaltaro:
Yeah, let's keep the standard polling cycle and in case we see bad things then we can adapt. As you said, in the end it would probably be better than cleaning up lots of things in one shot.

> The benefit of this patch is cleaning up job database when workflow is archived. (But those db is not replicated anyway).

Doesn't it clean up the wmagent_job database? Isn't that the database that gets replicated to wmstats? That is the most critical couch db that is giving trouble at the moment.

ticoann:
> Doesn't it clean up the wmagent_job database? Isn't that the database that gets replicated to wmstats? That is the most critical couch db that is giving trouble at the moment.

It cleans up wmagent_summary database which contains jobsummary type documents. However we are only keeping jobsummary when jobs failed. So it shouldn't be that many documents unless there are tons of failures.

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -221,3 +227,26 @@ def cleanAllLocalCouchDB(self, workflowName):
...
+        loop through the workflows in couchdb, if archived delete all the data in couchdb
+        """
+        
+        localWMStats = self.wmstatsCouchDB.getDBInstance()
+        options = {"group_level": 1, "reduce": True}
+        try:
+            results = localWMStats.loadView("WMStatsAgent", "allWorkflows", options = options)['rows']
+            requestNames = [x['key'] for x in results]
+        except Exception as ex:
+            errorMsg = "Error on loading workflow lists %s" % requestNames
```
amaltaro: I noticed in my tests that it was trying to use requestNames before assignment (because I mistakenly applied the patch - couch part :-(). So you may remove it from the except block

ticoann: Thanks Alan. Good catch. I updated the code.

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -221,3 +227,31 @@ def cleanAllLocalCouchDB(self, workflowName):
...
+        
+        localWMStats = self.wmstatsCouchDB.getDBInstance()
+        options = {"group_level": 1, "reduce": True}
+        try:
+            results = localWMStats.loadView("WMStatsAgent", "allWorkflows", options = options)['rows']
+            requestNames = [x['key'] for x in results]
+            logging.info("There are %s workfows to check for archived status" % len(requestNames))
+            
+            workflowDict = self.centralRequestDBReader.getStatusAndTypeByRequest(requestNames)
+            numDeletedRequests = 0
```
amaltaro: Just a minor suggestion, how about move this line to L236 and then move the return line (L249) after the try/except block?

ticoann: Thanks Alan, I updated as you suggested.

amaltaro:
It looks awesome now :-D Thanks Seangchan!

ticoann:
Sound like you are complementing yourself. :smile:

ericvaandering:
Complimenting, not complementing. Sorry, just finished reviewing a paper. :-)

amaltaro:
Relax Seangchan, I got your message. LoL

ticoann:
> Complimenting, not complementing. Sorry, just finished reviewing a paper. :-)

Thanks Eric, updated the comment. :-)

PR #6404, opened 2015-11-20
fix priority update for workqueue from reqmgr/reqmgr2

### LINKED ISSUE #6310: Make change Priority work with ReqMgr2
When changes the priority need to update workqueue priority as well. (check whether we can get the directly from reqmgr db instead of updating workqueue)

ticoann:
fixes #6310, @amaltaro, Alan it seem reqmgr update workqueue priority when it changes its priority. Only reqmgr2 doesn't do that. It might happen that if update fails it doesn't retry, though. (The case you pointed). Don't merge yet. running the test now.


ticoann:
Actually reqmgr priory update is not propagating to WQ as well I am still debugging.

ticoann:
It is a little hacky way to fix for reqmgr but this seems to be the simplest way. We will live with it since we will drop reqmgr (hopefully) soon. @amaltaro, Alan, please review. (tested in my vm)

amaltaro:
Sounds good Seangchan! And with this change it will update the GQ priority for all the elements regardless of their status.

PR #6406, opened 2015-11-21
Use DBSBufferFile module from DBS3 package

amaltaro:
WorkQueueManager was failing to completely create request subscriptions and that was also affecting JobAccountant (which I thought were due to the recent sql schema changes). Turns out the problem was happening because it was using dao that did not consider the new schema.

@ticoann this seems to fix the WMAgent issue I saw yesterday. Anyways, better waiting until tomorrow or Monday to merge it.

### DIFF
--- a/src/python/WMComponent/DBSUpload/Database/Interface/UploadToDBS.py
+++ b/src/python/WMComponent/DBSUpload/Database/Interface/UploadToDBS.py
@@ -14,7 +14,7 @@ from WMCore.WMFactory        import WMFactory
 from WMCore.DAOFactory       import DAOFactory
 from WMCore.WMConnectionBase import WMConnectionBase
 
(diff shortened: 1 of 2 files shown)


amaltaro:
BTW, with these changes 
DBSBuffer.Database.Interface.DBSBufferFile
should be deletable now. I'll try to clean more DBSBuffer code in the coming week.

amaltaro:
In case you want to merge it today for a new tag, go ahead, tests look ok. Otherwise we talk tomorrow ;)

ticoann:
Wow, that is a big mistake from my part. Thanks.

ticoann:
One thing though, following file is not used anywhere, right? 
 src/python/WMComponent/DBSUpload/Database/Interface/UploadToDBS.py

PR #6408, opened 2015-11-23
Advertise group name for workflow.

bbockelm:
This pulls the group name out of the workflow, if present, and advertises it separately in the classad.


bbockelm:
@PerilousApricot - you might want to watch this one.  It would allow Vanderbilt to pull in HI work over PP, regardless of the PP priority.

ticoann:
Thanks Brian.

PR #6409, opened 2015-11-23
Avoid random distribution of pending jobs.

bbockelm:
It seems that random distribution was pushing some sites way over their
pending limit for low priority jobs.  This would then cause the over-threshold
sites to be unable to pick up jobs.

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -484,15 +484,18 @@ class JobSubmitterPoller(BaseWorkerThread):
             self.cachedJobs         = {}
             self.jobDataCache       = {}
 
-        #Sort the sites using the following criteria:
-        #T1 sites go first, then T2, then T3
-        #After that we fill first the bigger ones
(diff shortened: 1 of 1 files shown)


bbockelm:
@ticoann @ericvaandering @amaltaro

amaltaro:
Sounds good, thanks Brian! Initially, I did not understand why start submitting first at the T3s, but now it seems to make no difference at all (besides the limit of jobs to submit per component cycle...). Am I missing something here?

bbockelm:
It's a second-order effect, nowhere nearly as big as the other part.  Basically, we're making the assumption it is better to fill up the 'pending' jobs on smaller T3s/T2s because they are less likely to receive workflows with very exclusive whitelists.

Example: suppose we have two workflows, A and B.  A has a whitelist of 50 sites and 10k idle jobs, while B has a whitelist of 1 site, T1_US_FNAL.

If we submit to T1_US_FNAL last, we might have submitted all the jobs in workflow A to other sites and submit workflow B.  If we submit to T1_US_FNAL first, we might only submit only workflow A.

_But_ this makes a particular assumption (T1s receive workflows with more restrictive whitelists) and won't make a big difference on steady state.

PR #6410, opened 2015-11-24
Convert SE to PNN for old ACDC requests

amaltaro:
And this should fix the last known issue for the SE->PNN transition.
@ticoann please review and merge it for pre2


ticoann:
Thanks Alan, good catch!

PR #6412, opened 2015-11-25
delete WMStats document locally

ticoann:
This still need to be tested and deployment script also need to be updated. 
(Also, more log message need to be added)
Following it what it supposed to do. (Try to avoid replicating deleted documents)
1. Documents will be deleted both agent and central wmstats independently. (Before central wmstats rely on replication for deleting its own documents.
2. There are 2 deletion is set. (One for summary documents periodically, and deleting everything including job documents when workflow is archived).
3. Tier0 should updated the same time since it shares replication code.


ticoann:
@amaltaro, Alan, please review. Still need to test. Hopefully I can push this today and create a new tag for the testbed.

src/python/WMCore/Services/WMStats/WMStatsWriter.py
```diff
@@ -208,3 +208,37 @@ def replaceRequestTransitionFromReqMgr(self, docs):
             self.couchDB.updateDocument(requestName, self.couchapp,
                         'generalFields',
                         fields={'general_fields': JSONEncoder().encode(requestTransition)})
+    
+    def getArchivedRequests(self):
```
amaltaro: Since it's a reading method, how about we move this method to the Reader class?

src/python/WMCore/Services/WMStats/WMStatsWriter.py
```diff
@@ -208,3 +208,37 @@ def replaceRequestTransitionFromReqMgr(self, docs):
...
+        delete all wmstats docs for a given requestName
+        """
+        view = "allWorkflows"
+        options = {"key": requestName, "reduce": False}
+        docs = self.couchDB.loadView(self.couchapp, view, options = options)['rows']
+   
+        for j in docs:
+            doc = {}
+            doc["_id"]  = j['value']['id']
+            doc["_rev"] = j['value']['rev']
```
amaltaro: I think the keys under 'value' are '_id' and '_value', no?

src/python/WMCore/Services/WMStats/WMStatsWriter.py
ticoann: Thanks I moved to Reader class

src/python/WMCore/Services/WMStats/WMStatsWriter.py
ticoann: actually key is set to id and rev.
https://github.com/dmwm/WMCore/pull/6412/files#diff-78a3679504ec44c511e66b686ee0ec1aR4

amaltaro: Ahh, sorry, you are right Seangchan!

ticoann:
tested in vm it worked fine now. but need to test in Tier0 instance.

PR #6413, opened 2015-11-25
Workaround for emitting doc null in workqueue

amaltaro:
All the production agents were facing issues while pulling work down (for almost 48h) because a row had doc: null, and then it was breaking the list/filter that is used by the agents for data acquisition.

Investigating the production workqueue views, it had 174 row with doc:null, like:
{"id":"039150ae7098cb4749a90123c014d135","key":"Canceled","value":{"_id":"039150ae7098cb4749a90123c014d135"},"doc":null},
{"id":"41122e9339257b37c8e8a719b84fa7fe","key":"CancelRequested","value":{"_id":"41122e9339257b37c8e8a719b84fa7fe"},"doc":null},
{"id":"47ede14085856e62907acffd101b70ee","key":"CancelRequested","value":{"_id":"47ede14085856e62907acffd101b70ee"},"doc":null},
{"id":"d91f4dd62d8088cc0db5d5f974844568","key":"Done","value":{"_id":"d91f4dd62d8088cc0db5d5f974844568"},"doc":null},
{"id":"15c67cd97353c5dfafaf60dcbf74c402","key":"Negotiating","value":{"_id":"15c67cd97353c5dfafaf60dcbf74c402"},"doc":null},
{"id":"5cd8d5183f2b97f4fcf4dbdcae395815","key":"Running","value":{"_id":"5cd8d5183f2b97f4fcf4dbdcae395815"},"doc":null},

which then breaks here:
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/lists/filter.js#L16

Talking to Diego and Bruno, we don't really understand how it can help besides a race condition where the document is still valid while the view is parsing it but before emitting the doc, it gets deleted and so the null value.

Tested in my VM. @ticoann please review. I'm not sure whether it can affect other things.
If you don't like this fix, another option would be fixing the list/filter.js code to simply skip the doc in case it's null, however in this case we would still keep the null doc in the workqueue views.

It's not critical to get this fix in, but if such problem happens during christmas, it stops the whole production from acquiring new work.

### DIFF
--- a/src/couchapps/WorkQueue/views/elementsByStatus/map.js
+++ b/src/couchapps/WorkQueue/views/elementsByStatus/map.js
@@ -1,8 +1,10 @@
 function(doc, site) {
-    var ele = doc["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
-    // Can't get multiple keys to work - use one for now
-    //emit([ele["Status"], ele["RequestName"], doc._id, doc.timestamp, ele["ParentQueueId"]],
-    if (ele) {
-        emit(ele["Status"], {'_id' : doc['_id']});
+    if (doc) {
+        var ele = doc["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
+        // Can't get multiple keys to work - use one for now
+        //emit([ele["Status"], ele["RequestName"], doc._id, doc.timestamp, ele["ParentQueueId"]],
+        if (ele) {
+            emit(ele["Status"], {'_id' : doc['_id']});
+        }
     }
 }


ticoann:
Hi Alan,  how did you get the following data?
did you use include_docs=True?

{"id":"039150ae7098cb4749a90123c014d135","key":"Canceled","value":{"_id":"039150ae7098cb4749a90123c014d135"},"doc":null},
{"id":"41122e9339257b37c8e8a719b84fa7fe","key":"CancelRequested","value":{"_id":"41122e9339257b37c8e8a719b84fa7fe"},"doc":null},
{"id":"47ede14085856e62907acffd101b70ee","key":"CancelRequested","value":{"_id":"47ede14085856e62907acffd101b70ee"},"doc":null},
{"id":"d91f4dd62d8088cc0db5d5f974844568","key":"Done","value":{"_id":"d91f4dd62d8088cc0db5d5f974844568"},"doc":null},
{"id":"15c67cd97353c5dfafaf60dcbf74c402","key":"Negotiating","value":{"_id":"15c67cd97353c5dfafaf60dcbf74c402"},"doc":null},
{"id":"5cd8d5183f2b97f4fcf4dbdcae395815","key":"Running","value":{"_id":"5cd8d5183f2b97f4fcf4dbdcae395815"},"doc":null},

Anyway, if the I am not sure the patch above will fix the problem if list filtering is working like include_docs.
i.e. include_docs makes separate call after view call is made. (although it looks like one call). Which means there is always the chance that document gets deleted between the call. 
Although it is weird the the last 2 document status is not near end status ("Negotiating", "Running").

I think it is safer to has a protection 
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/lists/filter.js#L16
if that is the case.

amaltaro:
That's right Seangchan, this was my call:
```
/workqueue/_design/WorkQueue/_view/elementsByStatus?stale=ok&idOnly=true&reduce=false&include_docs=true
```

and this is an example call that was failing in central couch logs
```
POST /workqueue/_design/WorkQueue/_list/filter/elementsByStatus?filter=%7B%22Status%22%3A+%22Negotiating%22%2C+%22ChildQueueUrl%2
2%3A+%22http%3A%2F%2Fcmssrv217.fnal.gov%3A5984%22%7D&idOnly=true&reduce=false&include_docs=true
```

From what I understood, it calls a list that applies a filter (basically formatting) on the output of elementsByStatus. And what emits the doc is the view, so I don't think there is a chance to get a null in this list when filtering that view. Of course, the same problem may happen to other views, but I wanted to be conservative here and change only what we are certain it can fail.

What do you think? Do we keep this change or we move it to the list/filter code?

ticoann:
Hi Alan,
Which line the error occurs? in the view or in the filter.
The check should have added whichever code error is occurred. (I am guessing it is within the filter, 

I am afraid that this patch won't fix what you wanted to fix.
```
if (doc)
``` 
I think above will be always true at the time the view is called. (I might be wrong).


> From what I understood, it calls a list that applies a filter (basically formatting) on the output of elementsByStatus.

I think is not direct out put from the elementsByStatus I am suspecting that doc in list is from include_doc call.

direct output from view doesn't contain any doc detail information
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/views/elementsByStatus/map.js#L6

ticoann:
replaced by #6414

PR #6416, opened 2015-11-27
Fix exception keyword; do partial cleanup in case spec is not found

amaltaro:
Just pushed this patch (well, a 1.0.9 flavor) to submit1 that was facing problems, as follows:
1. fixed keyword argument passed to Exception class
2. if the spec file cannot be found, then proceed with the clean up that can be done

I still don't know why that agent was failing to find the spec pkl file in the local disk. My guess is that it happened because one of the WorkQueueManager threads (WorkQueueManagerWMBSFileFeeder) was down for 2 days (no activities in the log), and so it may have not completed some actions...

@ticoann please review and then we can discuss and maybe merge it in the next week

### DIFF
--- a/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
+++ b/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
@@ -375,14 +375,14 @@ class TaskArchiverPoller(BaseWorkerThread):
                 try:
                     #Upload summary to couch
                     spec = retrieveWMSpec(wmWorkloadURL = finishedwfs[workflow]["spec"])
-                    if not spec:
-                        raise Exception(msg = "Couldn't load spec for %s" % workflow)
-                    self.archiveWorkflowSummary(spec = spec)
+                    if spec:
+                        self.archiveWorkflowSummary(spec = spec)
+                        # Send Reconstruciton performance information to DashBoard
+                        if self.dashBoardUrl != None:
+                            self.publishRecoPerfToDashBoard(spec)
+                    else:
+                        logging.warn("Workflow spec was not found for %s", workflow)
                     
-                    #Send Reconstruciton performance information to DashBoard
-                    if self.dashBoardUrl != None:
-                        self.publishRecoPerfToDashBoard(spec)
-
                     #Notify the WorkQueue, if there is one
                     if self.workQueue != None:
                         subList = []
@@ -443,8 +443,7 @@ class TaskArchiverPoller(BaseWorkerThread):
                     continue
                 except Exception as ex:
                     #Something didn't go well on couch, abort!!!
-                    msg = "Couldn't upload summary for workflow %s, will try again next time\n" % workflow
-                    msg += "Nothing will be deleted until the summary is in couch\n"
+                    msg = "Problem while archiving tasks for workflow %s\n" % workflow
                     msg += "Exception message: %s" % str(ex)
                     msg += "\nTraceback: %s" % traceback.format_exc()
                     logging.error(msg)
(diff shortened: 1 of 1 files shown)


ticoann:
Alan, was this first time happening in production? I looked through the code and I am not sure (as you said) why the spec file is not found. It seems some process must have deleted before the status is completed. As far as dealing with it after the fact, we can't archive the workflow without spec file, although it seems we can get the information somewhere else (download the spec from reqmgr db).
As far as deleting other record, we might just delete everything else except cleaning up the spec directory (which already got deleted).

Instead of doing continue, we pass spec with None value 
https://github.com/amaltaro/WMCore/blob/taskarchiver-exception/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py#L496

and handles that within the killWorkflows method
https://github.com/amaltaro/WMCore/blob/taskarchiver-exception/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py#L582

However I am really wondering how this happen.

amaltaro:
AFAIR, yes, it's the first time I see an agent complaining about the spec file. It happened with submit1 and the long story is here:
https://cms-logbook.cern.ch/elog/Workflow+processing/22351

A long story short, I think it happened because WorkQueueManager was not happy and one of its threads was stuck/down (WorkQueueManagerWMBSFileFeeder), and probably keeping a lock for wmbs_workflow table.

Then the TaskArchiver kicked out the workflow cleanup, did all the removal and the last part was updating wmbs_workflow, which failed with a lock wait timeout. Since the workflow remained in the table, it was caught again in the next cycle and then it started breaking. See some logs from the node below:
```
2015-11-25 13:07:42,895:INFO:TaskArchiverPoller:Deleting workflow fabozzi_HARVEST-2015B-ZeroBias-16Nov2015_7414_151124_122804_3986
2015-11-25 13:07:50,836:INFO:CleanCouchPoller:Cleaning up the old request docs
2015-11-25 13:07:53,606:INFO:CleanCouchPoller:{'error': {'conflict': 342}, 'delete': 342} docs deleted
2015-11-25 13:07:53,606:INFO:CleanCouchPoller:Getting requests in 'announced' state for team 'production'
2015-11-25 13:07:54,694:INFO:CleanCouchPoller:Ready to archive normal 0 workflows
2015-11-25 13:07:54,694:INFO:CleanCouchPoller:archive normal 0 workflows
2015-11-25 13:07:54,811:INFO:CleanCouchPoller:Ready to archive aborted 0 workflows
2015-11-25 13:07:54,812:INFO:CleanCouchPoller:archive aborted 0 workflows
2015-11-25 13:07:54,927:INFO:CleanCouchPoller:Ready to archive rejected 0 workflows
2015-11-25 13:07:54,927:INFO:CleanCouchPoller:archive rejected 0 workflows
2015-11-25 13:09:24,071:ERROR:TaskArchiverPoller:Critical error while deleting workflow fabozzi_HARVEST-2015B-ZeroBias-16Nov2015_7414_151124_122804_3986
(OperationalError) (1205, 'Lock wait timeout exceeded; try restarting transaction') 'DELETE FROM wmbs_workflow WHERE id = %s AND\n           NOT EXISTS (SELECT id FROM wmbs_subscription WHERE workflow = %
s AND id != %s)' (1352L, 1352L, 3823)Traceback (most recent call last):
...
2015-11-25 13:13:32,547:INFO:TaskArchiverPoller:Deleting workflow fabozzi_HARVEST-2015B-ZeroBias-16Nov2015_7414_151124_122804_3986
2015-11-25 13:15:13,620:ERROR:TaskArchiverPoller:Critical error while deleting workflow fabozzi_HARVEST-2015B-ZeroBias-16Nov2015_7414_151124_122804_3986
(OperationalError) (1205, 'Lock wait timeout exceeded; try restarting transaction') 'DELETE FROM wmbs_workflow WHERE id = %s AND\n           NOT EXISTS (SELECT id FROM wmbs_subscription WHERE workflow = %s AND id != %s)' (1352L, 1352L, 3823)Traceback (most recent call last):
```

Ok, we can pass None instead. If you think we should fix only the exception and leave the other fix for the future (if that happens again), it's ok with me as well. (?)

ticoann:
Thanks Alan, I think we should fix both if there is possibility if this could happen again.

amaltaro:
@ticoann I updated the code, can you please check it again?
I still want to run a couple tests tomorrow with this patch in before merging it.

PR #6421, opened 2015-11-30
Newline after CMSGroups classAd

amaltaro:
My bad I did not catch this bug in testbed, the problem is that my test workflows don't have a group and so they never matched that regexp. Pushing it to the newest production agents right now.


amaltaro:
tested in production :stuck_out_tongue:

ticoann:
Thank you very much Alan, since this is only for agent side change, merging it.

amaltaro:
Thanks Seangchan!

PR #6422, opened 2015-11-30
add the structure to control modifiable status by status

ticoann:
Initial version to control the restriction of property for each state for request in reqmgr2 interface.
@vkuznet, Valentin, I will add the code to get this from config (or db). Please review  that this is what you needed.


vkuznet:
Seangchan, changes look fine to me.
Valentin

vkuznet:
Seangchan, please check changes I made in #6425 . I didn't test them yet, but the logic should be fine. I read your list of attributes, put on top of table visible attributes, followed by invisible ones. Both PRs should come together in order to work.

PR #6423, opened 2015-12-01
Adopt group regexp for new pdmvserv task name scheme.

bbockelm:
@amaltaro this is probably why you didn't hit the `CMSGroups` issue earlier - newer tasks have tweaked the naming scheme.  This appropriately matches those I could find.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/CondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/CondorPlugin.py
@@ -32,7 +32,7 @@ from WMCore.BossAir.Plugins.BasePlugin import BasePlugin, BossAirPluginException
 from WMCore.FwkJobReport.Report        import Report
 from WMCore.Algorithms                 import SubprocessAlgos
 
-GROUP_NAME_RE = re.compile("^[a-zA-Z0-9]+_([A-Z]+)-")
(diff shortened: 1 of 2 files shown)


amaltaro:
My tests use a different request name schema.
Brian, this regex won't match ReReco workflows, like
fabozzi_Run2015C_25ns-TOTEM_minBias2-05Oct2015_7414_151015_182441_9095

would you like to adapt this PR to take that into account as well?

Out of curiosity, what's the purpose of this new classAd? To be able to bump prio by physics group or just monitoring purpose?

bbockelm:
Indeed - I only want to pull this from workflows that have an easily-identifiable group.  So I'm OK if it ignores the one in your example.

I'm doing this so sites have an easy (but not 100% accurate) way to give resources toward the activity of a particular physics group.

As I said, it's not 100% accurate and that's OK by design - it's meant to be a second-order effect.  If you have a better idea how to grab this information, I'd be happy to hear!

amaltaro:
I see. We could check with PPD folks if it would be too hard for them to provide this info within the workflow schema (right now we have a too general 'Group' key). Since they receive requests from physics group and they make/approve requests in McM, it could be a possibility...

bbockelm:
That's a good idea, but let's see if many sites use this mechanism.  I figure if it is popular at 90% accuracy, then it's worth it to make it 100% accurate.  If no one uses it ... well, we can move on in a few months.

PR #6424, opened 2015-12-01
Don't create fileset for IgnoredOutputModules

### LINKED ISSUE #6395: Don't create subscriptions for IgnoredOutputModules
This IgnoredOutputModules is a feature available only for TaskChain an Resubmission and it's mainly used by the recovery script (to recover ReReco lumis).

Needs more investigation, but it seems WorkQueueManager creates a subscription even for those output modules that we want to ignore, which later on brings problems in JobAccountant that will x-check the outputModules expected (database call) against the job outputs (which fails in this case).

amaltaro:
Fixes #6395
It's needed for the resubmission workflows (rereco recoveries) where we set IgnoredOutputModules in the schema to avoid producing/staging some tiers. Without this patch, JobAccountant fails jobs since the expected outputmodules is != than the one returned in the Report.pkl.

I tested these changes on the agent side and seems to be working ok. I still want to test in a clean environment though

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -270,8 +270,19 @@ class WMTaskHelper(TreeHelper):
         outputModules = []
         for stepName in self.listAllStepNames():
             outputModules.append(self.getOutputModulesForStep(stepName))
-
         return outputModules
+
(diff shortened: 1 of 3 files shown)


amaltaro:
Feel free to merge it as soon as it passes jenkins.

PR #6425, opened 2015-12-02
Changes to make visible attributes in a table (not tested yet)

vkuznet:


vkuznet:
Please do not merge, I need to test it together with #6422

PR #6427, opened 2015-12-02
All Error codes of CMS jobs and Stageout related for ASO/CRAB3

juztas:
summarize:
Commented out some error codes, which are not clear for me if they still happen; Tomorrow will look at dashboard history for last year.
3 new exitCodes from CMSSW:

```
+        8030: "Exceeded maximum allowed VSize (ExceededResourceVSize).",
+        8031: "Exceeded maximum allowed RSS (ExceededResourceRSS).",
+        8032: "Exceeded maximum allowed time (ExceededResourceTime).",
```

Introduce 4 new exitCodes for CRAB3/ASO stageout related issues. There is an issue created for me to report it to dashboard in case of stageout failure:

```
+        60321: "Site related issue: no space, SE down, refused connection.",
+        60322: "User is not authorized to write to destination site.",
+        60323: "User quota exceeded.",
+        60324: "Other stageout exception.",
```

Also there are some comments there exit codes mismatch between WMA and CRAB3 and which one are not in right range of exit codes.

@amaltaro @belforte @hernand @ticoann can you review and if there any doubts on exitCode or error msg, make a comment.

Mostly all of them are taken from: https://twiki.cern.ch/twiki/bin/view/CMSPublic/JobExitCodes

### DIFF
--- a/src/python/WMCore/WMExceptions.py
+++ b/src/python/WMCore/WMExceptions.py
@@ -6,15 +6,15 @@ _WMExceptions_
 mappings to a human readable message.
 """
 
-WMEXCEPTION = {'WMCore-1' : 'Not allowed to instantiate ',
-               'WMCore-2' : 'Problem creating database table ',
-               'WMCORE-3' : 'Could not find in library class ',
-               'WMCORE-4' : 'Problem with loading the class ',
-               'WMCORE-5' : 'Dialect not specified in configuration ',
-               'WMCORE-6' : 'Message name is same as diagnostic message name! ',
-               'WMCORE-7' : 'Component name is reserved word. ',
-               'WMCORE-8' : 'No config section for this component! ',
-               'WMCORE-9' : 'Problem inserting a trigger flag. ',
+WMEXCEPTION = {'WMCore-1': 'Not allowed to instantiate ',
+               'WMCore-2': 'Problem creating database table ',
+               'WMCORE-3': 'Could not find in library class ',
+               'WMCORE-4': 'Problem with loading the class ',
+               'WMCORE-5': 'Dialect not specified in configuration ',
+               'WMCORE-6': 'Message name is same as diagnostic message name! ',
+               'WMCORE-7': 'Component name is reserved word. ',
+               'WMCORE-8': 'No config section for this component! ',
+               'WMCORE-9': 'Problem inserting a trigger flag. ',
                'WMCORE-10': 'Problem setting trigger action. ',
                'WMCORE-11': 'Security exception. ',
                'WMCORE-12': 'Database connection problem ',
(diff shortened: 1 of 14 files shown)


ericvaandering:
I don't think you should comment out anything. All of those under 256 are defined by the system, I think.

src/python/WMCore/WMExceptions.py
```diff
@@ -6,15 +6,15 @@
...
-               'WMCore-2' : 'Problem creating database table ',
-               'WMCORE-3' : 'Could not find in library class ',
-               'WMCORE-4' : 'Problem with loading the class ',
-               'WMCORE-5' : 'Dialect not specified in configuration ',
-               'WMCORE-6' : 'Message name is same as diagnostic message name! ',
-               'WMCORE-7' : 'Component name is reserved word. ',
-               'WMCORE-8' : 'No config section for this component! ',
-               'WMCORE-9' : 'Problem inserting a trigger flag. ',
+WMEXCEPTION = {'WMCore-1': 'Not allowed to instantiate ',
+               'WMCore-2': 'Problem creating database table ',
```
ericvaandering: We should try to understand why this case shift is here and fix it. WMCore-2 is widely used. WMCore-1 is not used at all.

juztas:
From: http://tldp.org/LDP/abs/html/exitcodes.html#AEN23549
`1 - 2, 126 - 165, and 255 [1] have special meanings, and should therefore be avoided for user-specified exit parameters.` So I guess these has to be change with their special meanings. Also here it describes more and some messages could be improved: http://help.phys.unsw.edu.au/doc/f77/9,2,1.html

juztas:
@HassenRiahi @mmascher also please take a look at stageout errors and which will be marked as permanent/not.

HassenRiahi:
We have already agreed on the permanent stage-out failures here: https://github.com/dmwm/AsyncStageout/commit/70582f477c86e63a538d1648305c008d4c4e8ed2#commitcomment-13490521. So your patch looks good to me.

amaltaro:
Justas, I don't see a need for these 3 new exit codes:
+        8030: "Exceeded maximum allowed VSize (ExceededResourceVSize).",
+        8031: "Exceeded maximum allowed RSS (ExceededResourceRSS).",
+        8032: "Exceeded maximum allowed time (ExceededResourceTime).",

They are already available in the code (and being used):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMExceptions.py#L28-L32
and L57.

juztas:
@amaltaro 8030, 8031, 8032 comes from newer cmsRun and cmsRun can exit with these condes and not for application(wma,crab3 wrapper) itself. I just learned about them yesterday that they are available

src/python/WMCore/WMExceptions.py
```diff
@@ -22,37 +22,274 @@
...
+        60408: "Failure to stage out of log files during log collection (WMAgent).",
+        60409: "Timeout in stage out of log files during log collection (WMAgent).",
+        60410: "Failure in deleting log files in log collection (WMAgent).",
+        60411: "Timeout in deleting log files in log collection (WMAgent).",
+        60450: "No output files present in the report",
+        60451: "Output file lacked adler32 checksum (WMAgent).",
+        60452: "No run/lumi information in file (WMAgent).", # Wrong exitCode
+        60453: "Unable to determine pset hash from output file (CRAB3).", # Wrong exitCode
+        60514: "PreScriptFailure.", # Wrong exitCode
+        60515: "PreScriptScramFailure.", # Wrong exitCode
```
ericvaandering: These two should be removed, right @belforte

src/python/WMCore/WMExceptions.py
```diff
@@ -22,37 +22,274 @@
...
+        60406: "Failure in staging in log files during log collection (WMAgent).",
+        60407: "Timeout in staging in log files during log collection (WMAgent).",
+        60408: "Failure to stage out of log files during log collection (WMAgent).",
+        60409: "Timeout in stage out of log files during log collection (WMAgent).",
+        60410: "Failure in deleting log files in log collection (WMAgent).",
+        60411: "Timeout in deleting log files in log collection (WMAgent).",
+        60450: "No output files present in the report",
+        60451: "Output file lacked adler32 checksum (WMAgent).",
+        60452: "No run/lumi information in file (WMAgent).", # Wrong exitCode
+        60453: "Unable to determine pset hash from output file (CRAB3).", # Wrong exitCode
```
ericvaandering: These two seem like issues with the FWJR and don't here.

src/python/WMCore/WMExceptions.py
juztas: you change to 50513, great, will remove them

src/python/WMCore/WMExceptions.py
juztas: 60452 move to 50452 and also change in code. same for 60453 in CRAB

src/python/WMCore/WMExceptions.py
belforte: right. But are all those actyally used in the code ?

ticoann: @juztas, I merged Eric's patch #6432. Feel free to remove 60515. But what is the replacement for 60514?

belforte: Eric already changed 60514 to 50513.
That's expedient, but I still think we want less fine grained but
more descriptive exit codes. OTOH there is no exit code now for
"WMA run time code did not work" when you do not know where else
to stick blame to.
IMHO most rational solution is to introduce a new
range of exit codes for failures in our own job wrappers
and try hard to use only a handful.
Keeping SCRAM failures for cases where it is clear scram failed,
since those usually point to a local CMSSW mount problem (corrupted CVMFS cache eg)
and it will help to be able to classify immediately as a site problem.

Different are the rare cases where WMA PreScript failes for reasons
that need expert debugging to figure out.

E.g at some point we introduced 80000 for "interal error in CRAB Job Wrapper",
but that was not a good pick (my fault).

We could e.g. use 7xxxx. But really only use 10 or so.
We are not in the 80's anymore.

juztas:
Finished grep through WMCore/CRAB3 code, and updated my pull request. It is just proposal for future, comments are welcome:
```
----------------
Range(1 - 512)
    standard ones in Unix and indicate a CMSSW abort that the cmsRun did not catch as exception
----------------
Range(7000 - 9000)
    cmsRun (CMSSW) exit codes. These codes may depend on specific CMSSW version
    https://github.com/cms-sw/cmssw/blob/CMSSW_5_0_1/FWCore/Utilities/interface/EDMException.h#L26
----------------
Range(10000 - 19999)
    Failures related to the environment setup
----------------
Range(50000 - 59999)
    Failures related executable file
----------------
Range(60000 - 69999)
    Failures related staging-OUT
----------------
Range(70000 - 79999)
    Failures related only for WMAgent. (which does not fit into ranges before)
----------------
Range(80000 - 89999)
    Failures related only for CRAB3. (which does not fit into ranges before)
----------------
Range(90000 - 99999)
    Other problems which does not fit to any range before.
----------------
```

WMAgent todos in the code and use new exit codes:
```
range(61101-61105) move to range(71101-71105)
range(61300-61305) move to range(71300-71305)
60318 change to 70318
60451 change to 70452
```

CRAB3:
```
60453: move to 80453
90000: TODO: Need to separate. new stageout errors for ASO stageout are 60321-60324 and other failure codes needed for file metadata.
```

juztas:
@belforte I think you commented and I pushed almost at same time. As I don't see you your comment here. @ticoann it is taken care already and I added more details in this comment: https://github.com/dmwm/WMCore/pull/6427#issuecomment-161911688

belforte:
@juztas you did fine. I like your llist of ranges.
I will review a bit CRAB exit code usage in light of that.
The only relevant part of my lost comment is that when @ericvaandering changed 60514 to 50513 it would have been better to pick an exit code in 7xxxx range. And keep "scram related" for things where we are sure scram failed since those are usually a site problem. It helps to tell immediately site problem from WMA/CRAB problem. Much finer granularity in exit codes is not useful in my experience.

ericvaandering:
50513 is the replacement for 60614 (the original code, as far as I know, but I have not done any archaeology on this code).

belforte:
yes. On me.

src/python/WMCore/WMExceptions.py
```diff
@@ -22,37 +22,278 @@
...
+----------------
+Range(80000 - 89999)
+    Failures related only for CRAB3. (which does not fit into ranges before)
+----------------
+Range(90000 - 99999)
+    Other problems which does not fit to any range before.
+----------------
+"""
+
+WM_JOB_ERROR_CODES = {
```
ericvaandering: Let's also get a link to the TWiki in here with a strong admonition that changes must be made in both places.

src/python/WMCore/WMExceptions.py
ericvaandering: I have a patch to remove them and replace everything scram related (that I found) with 50513. I think a lot of this 60xxx stuff snuck in as a typo.

juztas:
@ericvaandering this might to become too big. Should I create separate pull for pylint and unit tests?

In that case for this pull request I would add back range(61101 - 61105) and others which are used currently in the code with message 'Deprecated. TODO: Will be removed in next pull'

juztas:
@belforte twiki updated: https://twiki.cern.ch/twiki/bin/view/CMSPublic/JobExitCodes

juztas:
Also what about dashboard, from there it gets error messages?

ericvaandering:
General advice is to make one commit that changes what you want to change and fixes any broken unit tests. (real code changes). Then have a 2nd commit on the same PR that fixes the pylint stuff so people can skip over that when reviewing.

belforte:
from the twiki

juztas:
thanks Eric. will do.

juztas:
just for my knowledge, do I need any magic word to retest everything?

src/python/WMCore/WMExceptions.py
juztas: Some are used but clearly not all. I leave them all as it is so far

ericvaandering:
@ticoann and @amaltaro please review.

ericvaandering:
No, any time you push a new commit or change the commits on a PR, Jenkins should start anew.

ticoann:
@ericvaandering, Eric, There are some error code changes which affects not just CRAB3 and ASO. Do you want us to patch the current agent or can this wait until next deployment? How much will this be the problem if some agent use the old code and new agent is using new error code.

juztas:
Now I defined kind of policy (see below) and had to move some error code range from 6**** to 7**** for WMA, also for CRAB3. There is no hurry on this patch and it can wait until next deployment, which is in February. 
If you have agent without patch, it will just report not correct exit code to dashboard and will show not correct in wmstats.
If you have agent with patch, it will report correct exit code and also will show correct in wmstats.
I grep through all code base of WMCore and CRAB3 and also there is a comment next to error msg, which service is using it.
My global view and what we discussed yesterday in CRAB3 meeting, would be also to have identifier is it site issue, is it FATAL error, for example:
```
{exitCode: (ErrorMsg, SiteIssue, FatalError)}
{50660: ('Application terminated by wrapper because using too much RAM (RSS).', False, True)}
```
But this is not for this pull request and maybe in the future.

Range of error codes:
```
+----------------
+Range(1 - 512)
+    standard ones in Unix and indicate a CMSSW abort that the cmsRun did not catch as exception
+----------------
+Range(7000 - 9000)
+    cmsRun (CMSSW) exit codes. These codes may depend on specific CMSSW version
+    https://github.com/cms-sw/cmssw/blob/CMSSW_5_0_1/FWCore/Utilities/interface/EDMException.h#L26
+----------------
+Range(10000 - 19999)
+    Failures related to the environment setup
+----------------
+Range(50000 - 59999)
+    Failures related executable file
+----------------
+Range(60000 - 69999)
+    Failures related staging-OUT
+----------------
+Range(70000 - 79999)
+    Failures related only for WMAgent. (which does not fit into ranges before)
+----------------
+Range(80000 - 89999)
+    Failures related only for CRAB3. (which does not fit into ranges before)
+----------------
+Range(90000 - 99999)
+    Other problems which does not fit to any range before.
+----------------
```

ticoann:
Thanks Justas

ericvaandering:
As Justas said, waiting on this patch is no big deal. Dashboard will have to adjust, but i think if we are reporting two error codes for one error for a while (the ones that were changed) its not so bad.

ticoann:
Thanks we won't patch the agent then.

amaltaro:
Justas, one more missing part I can remember of. Can you also update this line
https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py#L242
by replacing 61102 for 71102 ?

Thanks. Once we merge it (well, better before), we need to check the whole code and make sure these exit codes were updated everywhere.

juztas:
So for WMA changes my console output:
```
$ grep -ir '61101'
$ grep -ir '61102'
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
CHANGES:  - Let jobs retry when exit code 61102
$ grep -ir '61103'
$ grep -ir '61104'
$ grep -ir '61105'
$ grep -ir '61300'
$ grep -ir '61301'
$ grep -ir '61302'
$ grep -ir '61303'
$ grep -ir '61304'
Binary file .git/objects/pack/pack-1de19b6f8abc735df5e6dbb257ae7c7a73d225ba.pack matches
$ grep -ir '61305'
test/python/WMComponent_t/JobAccountant_t/fwjrs/LoadTest69.pkl:S'a8c61318-3dd8-11df-9797-0030482c2dd0'
$ grep -ir '60318'
src/python/WMCore/WMExceptions.py:                      60318: "Internal error in Crab cmscp.py stageout script.", # (CRAB3)
$ grep -ir '60452'
$
```
Of cource double check does not hurt, and good to check. Thanks Alan, I found one place there I missed to change and fixed it

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -37,8 +37,10 @@ class DQMUpload(Executor):
     Execute a DQMUpload Step
 
     """
+    def __init__(self):
+        return
```
amaltaro: This is wrong. You're overriding the parent constructor
https://github.com/juztas/WMCore/blob/errors/src/python/WMCore/WMSpec/Steps/Executor.py#L62

and so skipping some classes attributes.

Why did you add it here? I guess it was a pylint complain... if so, next time make sure to make pylint changes only when you are absolutely sure it's correct, otherwise it's a nightmare to review and detect bugs later on.
That's what I do at least, if I'm unsure, I just skip it :)

juztas: That was the only place there I added. Good to know it now. Thanks Alan, will remove soon

amaltaro:
Patch sounds good to me. Thanks Justas! @ticoann please merge

PR #6434, opened 2015-12-05
Stop submission at pending threshold

bbockelm:
Now that we mark all submitted jobs as pending at the site we're currently looping over, make sure we don't submit past the threshold.


amaltaro:
We wanted to hold on changes to job submission and thresholds until we completely understand it.
But since this patch is already in half of the agents, I think we should merge it.
I'm pushing it to the other agents now.

ticoann:
replaced by #6496

Issue #6435, opened 2015-12-07
Wrong datatype for filesize


amaltaro:
As reported in this HN:
https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/2688.html

and this ggus:
https://ggus.eu/index.php?mode=ticket_info&ticket_id=118081

WMAgent is injecting data with a wrong filesize, which brings data transfer down
everywhere. 

It seems the bug was introduced in this patch:
https://github.com/dmwm/WMCore/pull/6383/files#diff-3ecf4489a3d8f3d00928958a1608f9b1L97

I'm working on a patch right now.

For the agents already deployed, we'll need to alter the tables.

PR #6436, opened 2015-12-07
Make filesize BIGINT in MySQL

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/6435 and changes need to be pushed to all new agents already deployed.

@ticoann ping me on gtalk when you see this PR please, I'll wait for your ack :)

PS.: I noticed we have also changed several columns (in oracle) from number(11) to integer. Integer - in oracle - is the same as number(38), which means the database space taken is increased and probably the performance is affected. Maybe we should set to integer only what we're sure of, otherwise keep the old and good working lengths.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
@@ -83,7 +83,7 @@ class Create(DBCreator):
                  block_close_max_wait_time    INTEGER,
                  block_close_max_files        INTEGER,
                  block_close_max_events       INTEGER,
-                 block_close_max_size         INTEGER,
+                 block_close_max_size         BIGINT,
                  completed                    INTEGER       DEFAULT 0,
                  PRIMARY KEY (id),
                  CONSTRAINT uq_dbs_wor UNIQUE (name, task)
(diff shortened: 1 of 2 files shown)


hufnagel:
Sorry, this one is on me :(. Short term fix is to apply this patch directly to the running agents and modify the schema so new files will have their size recorded correctly ?

hufnagel:
I could go through my previous patch and review interger/bigint for MySQL and integer/number(11) for Oracle...

ticoann:
Hi Dirk, don't worry about it. :-) we have plan to fix it thanks to Alan. 
And honestly it not on you, I should be caught that on the review.

amaltaro:
No problem Dirk, I went through the mysql changes and spotted only those 3 columns to roll back.
Actually, Eric noticed that this change
LastModificationDate  BIGINT,

is not really needed, unless we want to run WMAgent until 2038 :-P.
@ticoann I will remove the LastModificationDate change and also make unittests for it. Please hold on on merging this one for now.

ticoann:
Thank you very much Alan, I will wait for your green light.

amaltaro:
@hufnagel Dirk, if you insist :-) I do have one thing that I'd like you to start doing. Can you please make sure to make a different commit for pylint/unittest related changes? It makes code review and future code debug much much easier for everyone.

amaltaro:
Removed LastModificationDate change and made a filesize unittest (tested only in my oracle VM, I hope it will succeed in Jenkins :)

PR #6437, opened 2015-12-07
Fix regex support

HassenRiahi:
Pls review. It fixes the support of regex. that matches more that 1 pattern.


ericvaandering:
Can you give an example of what the problem is? I wonder if this is the thing that Tony was complaining about that WMAgent didn't really do things correctly.

You should also provide a new unit test that fails with the current code and succeeds now.

HassenRiahi:
Here is an example: https://cmsweb.cern.ch/phedex/datasvc/xml/prod/tfc?node=T2_CH_CERNBOX

I will provide a new unit test.

ericvaandering:
So I guess what this is doing (from your example) is mapping /store/user/ewv/ to /local/dir/e/ewv and /store/user/hassen to /local/dir/h/hassen ?

HassenRiahi:
Yup!

mmascher:
I have a general comment. We all know that we have two implementations of the LFN<=>PFN conversion in CMS: one that is in Perl in phedex (lfn2pfn API), and one in python here in WMCore. Unfortunately we do not have anything that checks that the two are the same, and I know for sure that on some particular iinput the two were not returning the same result.

I think it would be nice to have a unit test that takes some lfn that are used in WMCore and CRAB3, all the sites, and checks that the phedex lfn2pfn API, and the WMCore API return the same information.

HassenRiahi:
The algo. is adjusted. I tested it over all CMS TFCs and it provides the same results as the current code for LFN ---> PFN convesion. A new unit test has been included as well.

ericvaandering:
Did you mention that you had some LFNs that initial failed in your new code? Can you put those in as unit tests too (same test is OK) so that we prevent easy regressions?

ericvaandering:
test this please

ericvaandering:
You should definitely fix the indentation issue in TrivialCatalog

mmascher:
Also @HassenRiahi , can you make a pull request to the branch 1.0.12_crab of WMCore as well please? That's the branch we are going to use for the February release. The patch should have the title "1.0.12_crab version of Fix regex support"

HassenRiahi:
Patch updated. Pls @ericvaandering review.  
Ok @mmascher once we agree on this pull I will make another one against the branch.

ericvaandering:
Looks OK to me.

PR #6439, opened 2015-12-08
Allow JobSubmitter to go over max running limits in the case of

bbockelm:
Production / Processing jobs.

This limit has not been effective for as long as we have had multi-site
whitelists.  However, it has been able to cause priority-inversions in
practice.

We keep the concept of a threshold for IO-intensive jobs and for
purposes of pulling work into the agent.

@amaltaro @ticoann - this corresponds to the thread I started on CompOps
@hufnagel - please think about how this might affect the T0.  That's less-familiar territory to me.

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -564,13 +564,13 @@ class JobSubmitterPoller(BaseWorkerThread):
                 if state == 'Down':
                     continue
 
-                #If the number of running exceeded the running slots in the
-                #site then get out
-                if totalRunningSlots >= 0 and totalRunning >= totalRunningSlots:
-                    continue
-
-                #If the task is running more than allowed then get out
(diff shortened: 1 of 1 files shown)


hufnagel:
Shouldn't matter for the Tier0. We anyways in practice have been using limits that are factors higher than we can run due to multicore. We never adjusted max running down to account for that and it just works. Limits on other job types are absolutely needed though.

ticoann:
test this please

amaltaro:
@ticoann please hold on on this PR. I don't really like oversubmitting jobs that belong to Production and Processing tasks, I mean, I don't see a good reason for that (AND if that's the case, then we better find a way to get rid of these task thresholds).

amaltaro:
Since the priority inversion has been fixed in WMCore, I'm in favor of closing this PR.
Moreover, merging those 2 job types (https://github.com/dmwm/WMCore/issues/7118) is going to decrease even more the need for this oversubsubscription.
@ticoann thoughts?

amaltaro:
Closing it for lack of activity and sort of disagreement :)

PR #6440, opened 2015-12-08
skeleton stuff for WMArchive (Do not merge!)

ticoann:
@yuyiguo Yuyi, this need to be filled with Archiver client, but before that we need to agree on document format and rpm need to be included in WMAgent. Please review,


src/python/WMCore/Services/WMArchiver/WMArchiver.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import (division, print_function)
+#TODO: add nececarry WMArchiver client import (need to include rpm)
+#
+
+class WMArchiver(object):
+    """
+    This is sckelton class which need be implemented.
```
ericvaandering: skeleton

ticoann: Thanks. :smile:

ticoann:
Update the code to handle bulk submission and caching in WMAgent side. (New component is created to handle  the WMArchive update) Still need to add unittest and integration test.

ticoann:
@vkuznet, Hi, Valentin as we discussed yesterday this is the structure WMAgent will post to WMArchive. Let me know if you want to change the key name, etc.
This is additional field will be added to fwjr
https://github.com/dmwm/WMCore/pull/6440/files#diff-1be2f4f865842be142f794cf9c1528f2R64

current fwjr structure example is here
https://github.com/dmwm/WMArchive/issues/46

vkuznet:
Seangchan, everything seems fine to me.

vkuznet:
Ohh, one thing, if docs are fetched from couchdb they have id/_rev keys, e.g.
https://github.com/dmwm/WMArchive/blob/master/test/data/fwjr_processing.json#L3
Do you want to keep them? May be we need to rename them as couch_id, couch_rev.

ticoann:
Thanks Valentin, I don't think we need to keep the revision number. for couch id it is the job id I am passing. I am not (at least shouldn't) passing couch doc itself - only fwjr properties so it should be filtered.
I will test and fix the bug if it is passing the id and rev.

ticoann:
So are you assuming whole json below for WMArchive? Or just fwjr field?
https://github.com/dmwm/WMArchive/blob/master/test/data/fwjr_processing.json#L7

vkuznet:
I took whole json as is and used it in all tests.

ticoann:
@vkuznet, I set up vocms008 agent with this patch. I am running test now. wmarchive url is set to 
https://couch-dev1.cern.ch/wmarchive/data, is this correct one?

vkuznet:
Seangchan,
I didn't update couch-dev1 for a while, it runs already old code even though
services is working. You should switch to vocms013.cern.ch/wmarchive/data
It is up-to-date.
Valentin.

ticoann:
Hi Valentin, I was trying to access vocms013 initially but it seems I don't have access. (not just to access the machine also frontend with my cert - which matters this case). Could you add my DN on the frontend.

ticoann:
@vkuznet, Valentin, please review. assumes https://github.com/dmwm/WMArchive/issues/96 will be fixed

vkuznet:
Seangchan, overall PR looks good to me. Feel free to fix suggested python 2.6 idiom though.

ticoann:
Thanks, Valentin, I fixed pylint warnings and added the unit tests. I will merge it when jenkins unittest is run.

PR #6441, opened 2015-12-08
Initial WMArchive REST service

vkuznet:
Please take initial implementation of WMArchive REST service (proxy server). Here is its configuration file:

```
import socket
from WMCore.Configuration import Configuration
from os import path

HOST = socket.gethostname().lower()
ROOTDIR = __file__.rsplit('/', 3)[0]
config = Configuration()

main = config.section_("main")
srv = main.section_("server")
srv.thread_pool = 30
main.application = "wmarchive"
main.port = 8246 # main application port it listens on
#main.index = "ui"
main.index = "data"
# Defaults to allow any CMS authenticated user. Write APIs should require
# additional roles in SiteDB (i.e. "Admin" role for the "ReqMgr" group)
#main.authz_defaults = {"role": None, "group": None, "site": None}

#sec = main.section_("tools").section_("cms_auth")
#sec.key_file = "%s/auth/wmcore-auth/header-auth-key" % ROOTDIR
#sec.key_file = "/Users/vk/CMS/DMWM/GIT/WMCore/auth-key"

# this is where the application will be mounted, where the REST API
# is reachable and this features in CMS web frontend rewrite rules
app = config.section_(main.application) # string containing "reqmgr2"
app.admin = "cms-service-webtools@cern.ch"
app.description = "CMS data operations WMArchive."
app.title = "CMS Request Manager (WMArchive)"

views = config.section_("views")

# practical to have this kind of configuration values not in
# service related RPM (difficult/impossible to change in CMS web
# deployment) but in the deployment configuration for the service

# redirector for the REST API implemented handlers
data = views.section_("data")
data.object = "WMCore.WMArchive.Service.RestApiHub.RestApiHub"
```

and (assuming all environment is set) it can be run as following:

```
wmc-httpd -r -d /tmp -l wma.log ./wmarch_config.py
```

### DIFF
--- a/src/python/WMCore/WMArchive/Service/Methods.py
+++ b/src/python/WMCore/WMArchive/Service/Methods.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+#pylint: disable=
+"""
+File       : Methods.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
+Description: This module consists of all REST APIs required by WMArchive service
+             Every API is designed as a class with appropriate get/post/put/delete
+             methods, see RESTEntity class for more details.
+"""
+
+# system modules
+import re
+import json
+import cherrypy
+
+# WMCore modules
+import WMCore
+from WMCore.REST.Server import RESTEntity, restcall, rows
+from WMCore.REST.Tools import tools
+from WMCore.REST.Validation import validate_str
+from WMCore.REST.Format import JSONFormat
+
+# WMArchive modules
+from WMCore.WMArchive.Service.Manager import WMArchiveManager
+
+# global regexp
+PAT_EXAMPLE = re.compile(r"^[a-zA-Z]$")
+
+class WMAInfo(RESTEntity):
+    def __init__(self, app, api, config, mount):
+        RESTEntity.__init__(self, app, api, config, mount)
+        self.config = config
+        self.config = config
+        self.mgr = WMArchiveManager(config)
+                
+    def validate(self, apiobj, method, api, param, safe):
+        "Validate user input"
+        if  not param.args or not param.kwargs:
+            return False # this class does not need any parameters
+        return True
+
+    @restcall
+    @tools.expires(secs=-1)
+    def get(self):
+        wmcore_reqmgr_version = WMCore.__version__
+        result = {"WMArchive version": self.mgr.version()}
+        return rows([result])
+
+class WMAGet(RESTEntity):
+    def __init__(self, app, api, config, mount):
+        RESTEntity.__init__(self, app, api, config, mount)
+        self.config = config
+        self.mgr = WMArchiveManager(config)
+    
+    def validate(self, apiobj, method, api, param, safe):
+        """
+        Validate request input data.
+        Has to be implemented, otherwise the service fails to start.
+        If it's not implemented correctly (e.g. just pass), the arguments
+        are not passed in the method at all.
+        
+        """
+        print "VALIDATE", param
+        # here we implement what should be validated
+        validate_str("query", param, safe, PAT_EXAMPLE, optional=True)
+
+    @restcall(formats = [('application/json', JSONFormat())])
+    @tools.expires(secs=-1)
+    def get(self, query):
+        "GET request with given query, all the work is done by WMArchiveManager"
+        docs = self.mgr.read(query)
+        return docs
+
+class WMAPost(RESTEntity):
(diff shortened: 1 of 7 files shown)


src/python/WMCore/WMArchive/Service/RestApiHub.py
```diff
@@ -0,0 +1,41 @@
...
+        :arg str mount: API URL mount point; passed to all entities."""
+        
+        RESTApi.__init__(self, app, config, mount)
+        
+        cherrypy.log("WMArchive entire configuration:\n%s" % Configuration.getInstance())    
+        cherrypy.log("WMArchive REST hub configuration subset:\n%s" % config)
+        
+        self._add({
+            "info": WMAInfo(app, self, config, mount),
+            "write": WMAPost(app, self, config, mount),
```
ticoann: I think the name should be something related to the data not about the action ("write")

src/python/WMCore/WMArchive/Service/RestApiHub.py
```diff
@@ -0,0 +1,41 @@
...
+            "fetch": WMAGet(app, self, config, mount),
```
ticoann: Here as well

src/python/WMCore/WMArchive/Web/WMArchiveService.py
```diff
@@ -0,0 +1,41 @@
...
+        cherryconf.update({'tools.encode.on': True,
+                           'tools.gzip.on': True,
+                           'tools.gzip.mime_types': mime_types,
+                          })
+
+        # initialize rest API
+        statedir = '/tmp'
+        app = RESTMain(config, statedir) # REST application
+        mount = '/rest' # mount point for cherrypy service
+        api = RestApiHub(app, config.mgr, mount)
```
ticoann: The way mounting new REST class is a different from WMCore/WebTools,  Although I don't think you need to follow exactly but you can check how it is done in sitedb
i.e.)
Configuration is pointing which object to construct the front page.
https://github.com/dmwm/sitedb/blob/master/src/python/Config.py#L32

Front page construction
https://github.com/dmwm/sitedb/blob/master/src/python/FrontPage.py

src/python/WMCore/WMArchive/Service/Methods.py
```diff
@@ -0,0 +1,97 @@
...
+        validate_str("query", param, safe, PAT_EXAMPLE, optional=True)
+
+    @restcall(formats = [('application/json', JSONFormat())])
+    @tools.expires(secs=-1)
+    def get(self, query):
+        "GET request with given query, all the work is done by WMArchiveManager"
+        docs = self.mgr.read(query)
+        return docs
+
+class WMAPost(RESTEntity):
```
ticoann: Maybe we can have more data oriented name, I think the verb can define the operation.

src/python/WMCore/WMArchive/Service/Methods.py
```diff
@@ -0,0 +1,97 @@
...
+        return True
+
+    @restcall
+    @tools.expires(secs=-1)
+    def get(self):
+        wmcore_reqmgr_version = WMCore.__version__
+        result = {"WMArchive version": self.mgr.version()}
+        return rows([result])
+
+class WMAGet(RESTEntity):
```
ticoann: Maybe we can have more data oriented name, I think the verb can define the operation.

ticoann:
Hi Valentin, as Yuyi mentioned I think it might be better the code is hosted in different repository.
https://github.com/dmwm/WMArchive
Since this project will contain different set of dependencies than other REST applications (ReqMgr2, WMStats) and I expect this will be much bigger and complicated application than those.

In terms of the code, in general,
RESTEntity is designed to operates over specific data. It will be implemented one or more verbs on that that. (i.e. POST, PUT, GET, DELETE)
I think in some cases it doesn't really fits the use case. (I had some problems applying that concept on reqmgr2).
Anyway, this is Lassi's design concept, so sitedb will be the good example how he initially envision this.

https://github.com/dmwm/sitedb/blob/master/src/python/DataPeople.py

I added some in line comment

vkuznet:
I migrated code (and fixed it with respect of provided comments) to WMArchive repository. The new PR is https://github.com/dmwm/WMArchive/pull/1 I'll keep this open until we merge new PR. So far I only modified data classes, but didn't yet change service wrt sitedb layout. I need to think about it if it is appropriate here. This can be done later. Please review new PR.

ticoann:
This is moved to WMArchive repository, closing it

PR #6442, opened 2015-12-08
Fix DatabasePage class to load even without database configuration

vkuznet:
This is required for classes which do not need database, i.e. doesn't have database in configuration. For instance, if my class want to implement RESTApi, WebApi but does not need or implement database access, the current system enfornce through inheritance the database parameter presense in configuration file. Otherwise it does not load WebApi class which inherits from DatabasePage. I put DatabasePage config loading in try/except block to allow higher classes to load configuration even without database parameter.

### DIFF
--- a/src/python/WMCore/WebTools/DatabasePage.py
+++ b/src/python/WMCore/WebTools/DatabasePage.py
@@ -43,10 +43,13 @@ class DatabasePage(TemplatedPage, DBFormatter):
         exposed that passes the database connection to all the DAO's.
         """
         TemplatedPage.__init__(self, config)
-        dbConfig = ConfigDBMap(config)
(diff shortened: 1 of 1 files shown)


ticoann:
@vkuznet, Valentin, so you still need this patch? If that is the case, could you have more specific check on db, i.e.
Just putting 
```python
try:
        dbConfig = ConfigDBMap(config)
except AssertionError as ex:
        some warning
else:
       transaction code
```
Thanks,

vkuznet:
Seangchan,
yes, I still want this behavior, since client may not need a DB, i.e. not
supply it in configuration file, but still use framework. I'll adjust code
later.
Valentin.

vkuznet:
Seangchan,
I modified code as you suggested.
V.

PR #6446, opened 2015-12-09
Improve listFileArray usage; query by block+lumi instead of dataset+lumi

amaltaro:
Fixes (or partially) https://github.com/dmwm/DBS/issues/505 , at least from the WMCore side.

As previously discussed, querying for dataset + run + lumi_list is a full table scan and very expensive, which quite often times out (5min).

Changing it to, first we get a list of blocknames and then we query for block+run+lumi_list, which is much faster.

The slicing in WMCore seems to be still needed, if we query for 20k lumis (and block), it takes almost 5min sometimes. However, I increased the slicing to 1000 lumis, should be better than 50.

@ticoann DO NOT MERGE yet please. I ran only 1 interactive test and it seems fine, but better test it for real specially if we want Diego to patch/push it to WorkQueue production this week.

@ticoann @yuyiguo please review.

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/Block.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/Block.py
@@ -215,15 +215,18 @@ class Block(StartPolicyInterface):
         lumiMask = task.getLumiMask()
         taskMask = LumiList(compactList = lumiMask)
 
+        # for performance reasons, we first get all the blocknames
+        blocks = dbs.dbs.listBlocks(dataset=datasetPath)
+
         # Find all the files that have runs and lumis we are interested in,
         # fill block lfn part of maskedBlocks
-
         for run, lumis in lumiMask.items():
             files = []
-            for slumis in Lexicon.slicedIterator(lumis, 50):
-                slicedFiles = dbs.dbs.listFileArray(dataset=datasetPath, run_num=run,
-                                       lumi_list=slumis, detail=True)
-                files.extend(slicedFiles)
+            for slumis in Lexicon.slicedIterator(lumis, 1000):
+                for block in blocks:
+                    slicedFiles = dbs.dbs.listFileArray(block_name=block.get('block_name'), run_num=run,
+                                                        lumi_list=slumis, detail=True)
+                    files.extend(slicedFiles)
             for lfn in files:
                 blockName = lfn['block_name']
                 fileName = lfn['logical_file_name']
@@ -233,7 +236,6 @@ class Block(StartPolicyInterface):
                     maskedBlocks[blockName][fileName] = LumiList()
 
         # Fill maskedLumis part of maskedBlocks
-
         for block in maskedBlocks:
             fileLumis = dbs.dbs.listFileLumis(block_name=block, validFileOnly = 1)
             for fileLumi in fileLumis:


amaltaro:
test this please

amaltaro:
@ticoann I talked to Diego and I just applied this patch to workqueue testbed. I'll run some tests over the night and if everything is Ok, we can push it to production tomorrow morning.
If you or @yuyiguo don't think it's a good idea, please let me know before that :)

ticoann:
so actually, if it is called by blocks not dataset (in case run number is 1), this query is manageable?
For me it looks fine, in that case.

yuyiguo:
HI, Alan:
Seangchan talked me during the break. The code you had would not work because it still searching the lumi table with run number = 1. 

I would suggested you do below steps:
I think below steps may help. You may want to test it in a sample script.
1. find all blocks in the given dataset.
2. loop all blocks with listFileLuims(block_name=b)
3. compare your lumi list with what you get from 2 , then you will find the blocks you want to process.

These were the original steps I suggested in the DBS issue of this problem. BWT, I would throw exception for run num =1 for listFileArray API in next DBS release. Basically, we would not allow any search with run num =1 in lumi table unless you have lfn.

amaltaro:
Seangchan, I injected the same workflow that was failing in production (as reported in the DBS issue) with 35k lumi sections and the whole process in WorkQueue (central) took around 6min. My tests with the previous code (dataset instead of blocks) took 1h for 10k lumi sections only (when it worked). So I take it as a working fix :-)

Yuyi, I do think your suggestion is the most performant one, BUT in this case you move the whole processing to WMCore instead. And since DBS is providing APIs to handle lumi list, why would I bother of reinventing the wheel in WMCore.

> BWT, I would throw exception for run num =1 for listFileArray API in next DBS release. Basically, we would not allow any search with run num =1 in lumi table unless you have lfn.

Would you mind then creating a WMCore GH issue saying what *exactly* will not be supported anymore in DBS starting in the next deployment? I believe you'll need to make a broader announce for such change, since more services/people may be relying on it right now.

I'll assign some other workflows to testbed, just to make sure basic things are working as expected. If it is, then tomorrow we'll patch the production WorkQueue.
Sorry, Diego will be gone on Monday and we need to make this change tomorrow.

yuyiguo:
Alan, 
It is ok for now, but we have to change it again in February release.  Slowness of search file lumi table with run number 1 w/o lfn was well understood. There was a wide range discussion about 6 month or longer ago. The conclusion was stopping search file lumi table with run num 1for all DBS APIs. I updated other APIs, but missed this one for some reason. I don't think any other applications are doing this except for WMAgent. It looked like you missed the discussion. Please see the doc of one of the apis: https://cms-http-group.web.cern.ch/cms-http-group/apidoc/dbs3-client/current/dbs.apis.html?highlight=listfilelumis#dbs.apis.dbsClient.DbsApi.listFileLumis

I will update listFileArray to eliminate the run num for 1 search because sooner or later, you would see the same problem for block searching as you had for dataset. Eventually it is not the API allows or not, it is the data cannot be searched in the way we wish.

amaltaro:
Sigh, my lumiMask tests failed because I forgot to properly indent
files.extend(slicedFiles)
which must go inside the blocks loop ... Fixing it in a sec.

BTW, don't worry Yuyi, we'll make a proper fix for February. I just want to avoid making too many changes for critical production fixes, the less changes the better in these cases ;)

amaltaro:
Patch is working fine now and I'm pushing it to the central WorkQueue in production (just talked to Diego). Then we need to patch all the agents as well, will take care of it later this evening.

ticoann:
Thank you very much Alan, I will create the new tag with this (wmagent branch).

amaltaro:
Relax, I got it. Pay attention to your training ;)

amaltaro:
BTW, there is no need for another tag right now, let's wait and see whether we get more bug fixes in in the next few days :)

amaltaro:
@yuyiguo one more thing on the documentation you pointed out, that's not clear to me:
"run_num=1 is MC data and it will cause almost whole table scan, so run_num=1 will cause an input error."

Can we or we cannot use run_num=1 when we pass a block_name along the listFileLumis. IF it only accepts run_num when logical_file_name is provided, can you please update the documentation?

amaltaro:
@yuyiguo I'd like to create a new issue for January, can you update us on my last questions please?

yuyiguo:
we cannot use run_num=1 when we pass a block_name or dataset along the listFileLumis.

amaltaro:
@ticoann this PR will conflict with #6479
You may want to merge it, and then rebase your PR.

ticoann:
Thanks Alan, I will rebase that.

PR #6456, opened 2015-12-13
Add threadid to logfile.

bbockelm:
Without this information, it's fairly difficult to debug multithreaded
daemons (especially if multiple threads invoke the same function).


bbockelm:
For what it's worth, it would be more useful if we could log the Linux thread ID - but looking around on the web, it involves doing raw syscalls.  Not quite that desperate yet.

PR #6457, opened 2015-12-13
Cache pileup information.

bbockelm:
Pileup datasets shouldn't have open blocks or be replicating while
jobs are assigned.  Hence, it should be OK to stash the results for
a few hours.

(It might be possible to take an alternate approach and lookup
the entire dataset instead of block-by-block.)

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
+++ b/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
@@ -5,6 +5,8 @@ of pileup files in the job sandbox for the dataset.
 """
 
 import os
+import copy
+import time
 
 from WMCore.WMSpec.Steps.Fetchers.FetcherInterface import FetcherInterface
 import WMCore.WMSpec.WMStep as WMStep
@@ -26,6 +28,8 @@ def mapSitetoSE(sites):
     for site in sites:
         fakeSEs.extend(rControl.listSiteInfo(site)['pnn'])
     return fakeSEs
+
+g_pileup_cache = {}
 
 class PileupFetcher(FetcherInterface):
     """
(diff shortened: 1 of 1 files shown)


ticoann:
Thanks Brian, Do you know why the duplicate calls are made for the same Pileup dataset, I was wondering whether we can even cache final product (json file). 

https://github.com/dmwm/WMCore/blob/1c10c5f1c8e64201ebcc944d7ff6817bc6b29d54/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py#L71

Is there 2 (or more) different workflow has the same Pileup dataset or the different steps within the same workflow uses the same Pileup dataset?

bbockelm:
As far as I can tell, we build the pileup information twice within the same workflow - probably simply two steps that use it?

That said, do we really need to look up information block-by-block?  Are there not PhEDEx and DBS calls per dataset?

Would be nice if we could both cache the output and use bulk calls to build it in the first place.

amaltaro:
It's very common to see the Pileup dataset set in multiple places within the same workflow (usually in all tasks of a taskChain request).

The reason we look up that information block-by-block is that when the job gets into a WN, it will get all blocks (actually their files) that are present at that current site.

The way we could improve it would be:
```
dbs.listFileArray(dataset=dataset, validFileOnly = 1, detail=True)
```
and then we do some processing on the agent side. With that we get rid of the block-based calls for 'FileList' and 'NumberOfEvents'. The data location still needs to go through PhEDEx though and I haven't checked the available APIs.

amaltaro:
Oups, to complete my answer, we could implement a new DBSReader API that does that processing. Since listFileArray is a DBS API (not our own API).

ericvaandering:
test this please

ericvaandering:
These failures may be nothing. The first DBS check I looked at had an error like this "HTTPError:HTTP Error 503: Service Unavailable" which typically means exactly that: DBS is down. Just one of the reasons we need to properly mock DBS.

ericvaandering:
Don't merge this yet, at least on the master branch, if @amaltaro has other ideas how to do this.

ticoann:
I am not sure we can call this by dataset (not because there is no api but for the performance reason, I don't think we can query effectively for all the files in the dataset in dbs. For PhEDEx it might be fine to get all the block locations in one dataset call if that exists - it doesn't seems it exist though, looking at PhEDEx APIs. There is one https://cmsweb.cern.ch/phedex/datasvc/doc/data but I think this would be the same problem as DBS since all the node location is property of file) I think that is why it calls by block by block here. However we can/should run the test the calling by dataset both on dbs and PhEDEx.

As far as final json file cache, if the most of the case is duplicate call on Pileup dataset with in the same workflow, it should be relatively simple to have cache and cache will be deleted when workflow is done.

ticoann:
replaced by #6463

PR #6450, opened 2015-12-11
Split job work for JobArchiver

amaltaro:
fixes #6449
As reported by Brian here
https://cms-logbook.cern.ch/elog/Workflow+processing/22541

we can split the job state change in JobArchiver, especially for cases where a super big workflow is aborted and the component has to go through lots of jobs.

@ticoann please review, still needs to be tested in testbed.

PR #6458, opened 2015-12-13
Limit the work done by the archiver.

bbockelm:
Duplicate #6450 
In corner cases (particularly, large-scale kills), the JobArchiver
will try to pull all possible jobs - possibly more than it can
handle.

This limits the JobArchiver to 10k jobs per type (successful, killed,
exhausted).

### DIFF
--- a/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
+++ b/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
@@ -126,28 +126,30 @@ class JobArchiverPoller(BaseWorkerThread):
         Performs the archiveJobs method, looking for each type of failure
         And deal with it as desired.
         """
-        try:
-            self.archiveJobs()
-            self.pollForClosable()
-            self.markInjected()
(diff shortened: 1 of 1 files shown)


ticoann:
Thank you very much Brian, I guess you already applied this in cmssrv218, right?

ticoann:
test this please

amaltaro:
I've made the same patch a few days ago, just a slightly different approach: #6450 

To be honest, mine sounds cleaner, but I'm not fighting for it :-) Shall I close mine now?

bbockelm:
#6450 looks fine to me - I just needed something that would cause `cmssrv218` to start making progress.

This patch additionally causes the daemon to no longer wait if they hit the limits.  However, looking at logs, there's not much benefit: 10k jobs take about 3 hours to clear, so the 5 minutes this patch saves is fairly irrelevant.

ticoann:
Thanks Alan, it missed your earlier patch. I am closing this one, then

PR #6459, opened 2015-12-13
Sliced iterator that works on all iterables

### LINKED ISSUE #6453: slicedIterator does not belong in Lexicon and only works for sequences
@ticoann what do you think

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L570

does not seem to me to belong in Lexicon.py since it's instead just a general utility function. The other problem with it is that it does not work with an iterator, only with an actual list. I have a version of this that will work with iterators too. It's a bit slower, but the original will slice 12M items into 1000 item chunks in 72 ms and the new version is 200 ms.

I'd propose to replace this with a new version that is actually in src/python/Utils since it has *nothing* to do with WMCore. Here's my replacement code:

```python
#! /usr/bin/env python

import itertools
import time


class MyTimer(object):
    def __init__(self, verbose=False):
        self.verbose = verbose

    def __enter__(self):
        self.start = time.time()
        return self

    def __exit__(self, *args):
        self.end = time.time()
        self.secs = self.end - self.start
        self.msecs = self.secs * 1000  # millisecs
        if self.verbose:
            print 'elapsed time: %f ms' % self.msecs


def grouper(iterable, n):
    """
    :param iterable: List of other iterable to slice
    :type: iterable
    :param n: Chunk size for resulting lists
    :type: int
    :return: iterator of the sliced list
    """
    iterable = iter(iterable)
    return iter(lambda: list(itertools.islice(iterable, n)), [])


def slicedIterator(sourceList, sliceSize):
    """
    :param: sourceList: list which need to be sliced
    :type: list
    :param: sliceSize: size of the slice
    :type: int
    :return: iterator of the sliced list
    """
    start = 0
    end = 0

    while len(sourceList) > end:
        end = start + sliceSize
        yield sourceList[start: end]
        start = end


print "Make big list",
with MyTimer(verbose=True) as t:
    bigList = list(range(0, 12345678))
print "Grouping big list",
with MyTimer(verbose=True) as t:
    for i in grouper(bigList, 1000):
        pass
print "Grouping xrange",
with MyTimer(verbose=True) as t:
    for i in grouper(xrange(0, 12345678), 1000):
        pass
print "Slicing big list",
with MyTimer(verbose=True) as t:
    for i in slicedIterator(bigList, 1000):
        pass

tinyList = range(0, 11)
print "Original function"
for i in slicedIterator(tinyList, 3):
    print i
print "New function"
for i in grouper(tinyList, 3):
    print i
print "New function on xrange"
for i in grouper(xrange(0, 11), 3):
    print i
```

ericvaandering:
Fixes #6453. Don't merge yet.

### DIFF
--- a/src/python/Utils/IterTools.py
+++ b/src/python/Utils/IterTools.py
@@ -1,0 +1,18 @@
+#! /usr/bin/env python
+from __future__ import division, print_function
+
+from itertools import islice
+
+
+def grouper(iterable, n):
+    """
+    :param iterable: List of other iterable to slice
+    :type: iterable
+    :param n: Chunk size for resulting lists
+    :type: int
+    :return: iterator of the sliced list
+
+    Source: http://stackoverflow.com/questions/3992735/python-generator-that-groups-another-iterable-into-groups-of-n
+    """
+    iterable = iter(iterable)
+    return iter(lambda: list(islice(iterable, n)), [])
--- a/src/python/Utils/__init__.py
+++ b/src/python/Utils/__init__.py
@@ -1,0 +1,11 @@
+#! /usr/bin/env python
+
+"""
+_Utils_
+
+Generic code that has nothing to do with WMCore, just useful
+stuff to use. If this is a found snippet of code, please
+credit the source.
+"""
+
+__all__ = []
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -9,11 +9,12 @@ from collections import defaultdict
 from dbs.apis.dbsClient import DbsApi
 from dbs.exceptions.dbsClientException import *
 
+from Utils.IterTools import grouper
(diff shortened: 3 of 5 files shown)


ericvaandering:
@ticoann any ideas about this? The test is failing, but I think its just because something doesn't know about a new top level directory under python. Is there something that has to be added to allow a new one?

amaltaro:
I think you're missing a __init__.py under src/python/Utils . I slightly remember having the same issue in the past.

amaltaro:
I meant
```
__init__.py
```

ericvaandering:
Thanks @amaltaro , that was it.

amaltaro:
yw ;-)

ericvaandering:
I'm not adding future imports to ```__init__.py```

ericvaandering:
@ticoann this is also ready to be merged.

ticoann:
duplicate #6479

PR #6463, opened 2015-12-15
call by dataset instead of blocks and pileup dataset

ticoann:
related to #6457, (Thanks to Brian and Alan's suggestion), Call is made by dataset. (In this particular case with dataset /MinBias_TuneCUETP8M1_13TeV-pythia8/RunIISummer15GS-MCRUN2_71_V1-v2/GEN-SIM , It worked fine to query dataset level.  It took < 1 min for the test in fnal machine). Still need to add the cache (similar to #6457 but will try to add the final json file instead of memory cache).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,127 @@
+1.0.12 to 1.0.13.pre2:
+  - Merge pull request #6509 from ticoann/deployment_script
+  - Merge pull request #6514 from ticoann/fix_dependency
+  - change default arch to gcc493
+  - add Utils dependency in wmagent
+  - Merge pull request #6508 from hufnagel/missing-init
+  - Merge pull request #6504 from ticoann/remove_signal_import
+  - change the module import
+  - Merge pull request #6507 from vkuznet/fix-siginfo
+  - add a few missing init.py files
+  - Change SIGINFO to SIGUSR1 since the former only exists in newest version of python 2.7.11
+  - 1.0.13.pre1
+  - Merge pull request #6501 from ticoann/pycurl_revert
+  - Merge pull request #6502 from ticoann/update_wmcore_to_py27
+  - Merge pull request #6494 from hufnagel/condorstatusservice
+  - Merge pull request #6325 from hufnagel/eos-auth-problems
+  - Merge pull request #6337 from vkuznet/fix-tran-con-gen
+  - Cleaning pylint issues
+  - update to py2.7 and gcc493
+  - Merge pull request #6500 from dmwm/ericvaandering-patch-6
+  - Merge pull request #6499 from dmwm/ericvaandering-patch-5
+  - revert changes made by 6495
+  - Update env_unittest.sh
+  - Update env_unittest.sh
+  - Update deploy_unittest.sh
+  - work around EOS auth problems
+  - add CondorStatusService
+  - Merge pull request #6497 from dmwm/ericvaandering-patch-3
+  - Switch to new architecture, comp repo
+  - Merge pull request #6495 from vkuznet/fix_rest_reqmgr
+  - Merge pull request #6493 from vkuznet/fix_restfrontpage
+  - Merge pull request #6496 from ticoann/stop_submission_at_pending_theshold
+  - Add comments
+  - Stop submission at pending threshold
+  - Catch explicitly HTTPError
+  - Changes to fix RequestManager_t unittest and migration to pycurl; the later uses POST/PUT requests and send values in a body of the request instead of positional arguments
+  - Relax constrain on passed args (when no YUI is used)
+  - Merge pull request #6492 from ericvaandering/add_HIdata
+  - Add the HI dataset
+  - Merge pull request #6489 from ticoann/fix_reqmgr_client_unittest
+  - fix RequestManager unittest
+  - Merge pull request #6491 from ticoann/fix_dependency
+  - fix dependencies
+  - Merge pull request #6488 from ericvaandering/update_phedex_emulator
+  - Add a dumb function to the PhEDEx emulator to make mocked-DBS work
+  - Merge pull request #6475 from amaltaro/request-reqmgr2
+  - Fix pylint issues
+  - Merge pull request #6483 from amaltaro/deploy-gcc493
+  - Make deployment compatible with py2.7/gcc493
+  - Merge pull request #6437 from HassenRiahi/FixRegexSupport
+  - Merge pull request #6443 from vkuznet/error_msg_bug
+  - Merge pull request #6456 from bbockelm/log_threadid
+  - Merge pull request #6481 from jha2/using_mock2
+  - Merge pull request #6474 from ericvaandering/make_mock_json
+  - Merge pull request #6479 from ticoann/grouper_patch
+  - Implemented comments from pylint for #6481
+  - Merge remote-tracking branch 'upstream/master' into grouper_patch
+  - Merge pull request #6446 from amaltaro/dbs-lumiList
+  - Adding unit test using mock for DbsApi.
+  - Pylint complaints for DBSReader
+  - Sliced iterator that works on all iterables
+  - Update unit test and fix bad identation.
+  - Merge pull request #6450 from amaltaro/jobArchiver-sliced
+  - Fix status validation in reqmgr2
+  - Merge pull request #6470 from ericvaandering/dbs_mock_wo_gather
+  - A program to be run interactively to add to or replace the JSON files in test/data used to mock the DBS emulator
+  - Fix the TFC algo. and add unit test.
+  - Merge remote branch 'origin/FixRegexSupport'
+  - Merge remote branch 'upstream/master'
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, Alan please review. I will try to add cache also.

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
```diff
@@ -70,15 +74,21 @@ def _queryDbsAndGetPileupConfig(self, stepHelper, dbsReader, fakeSites):
             # each dataset input can generally be a list, iterate over dataset names
             blockDict = {}
```
ticoann: One thing I don't quite get is blockDict is outside of datasets loop. I am assuming there will be only one pileup dataset for each type. but if there are mulitple datasets in the same type, wouldn't this cause the problem by missing blocks for different dataset? But I leave as it is since that will require further down return data structure change.

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
```diff
@@ -70,15 +74,21 @@ def _queryDbsAndGetPileupConfig(self, stepHelper, dbsReader, fakeSites):
...
-                # DBS listBlocks returns list of DbsFileBlock objects for each dataset,
-                # iterate over and query each block to get list of files
-                for dbsBlockName in blockNames:
-                    blockDict[dbsBlockName] = {"FileList": sorted(dbsReader.lfnsInBlock(dbsBlockName)),
-                                               "PhEDExNodeNames": dbsReader.listFileBlockLocation(dbsBlockName),
-                                               "NumberOfEvents": dbsReader.getDBSSummaryInfo(block=dbsBlockName)['NumberOfEvents']}
-                    blockDict[dbsBlockName]['PhEDExNodeNames'].extend(x for x in fakeSE if x not in \
-                                                                      blockDict[dbsBlockName]['PhEDExNodeNames'])
+            
+                blockFileInfo = dbsReader.dbs.listFileArray(dataset=dataset, validFileOnly = 1, detail=True)
```
amaltaro: Typo here, "dbs." should be removed.

amaltaro: Actually, listFileArray is a dbs api, not dbsReader :)

ticoann: :-) Actually it might be better to wrap up listFileArray within DBSReader api for maintenance .  I will see whether that make more sense.

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
amaltaro: Good question, maybe @ericvaandering  can shed some lights on here. I believe there could be a case like "stepHelper.mc.pileup", no? I'm probably wrong, since it works for any type (mc, data) of pileup...

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
```diff
@@ -57,6 +58,9 @@ def _queryDbsAndGetPileupConfig(self, stepHelper, dbsReader, fakeSites):
         a subset of the blocks in a dataset will be at a site.
 
         """
+        # only production PhEDEx is connected (This can be moved to init method
+        phedex = PhEDEx()
+        node_filter = set(['UNKNOWN', None])
```
amaltaro: do we really need this node_filter? Why do we have this UNKNOWN?

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
ticoann: It divides by types so each type will have the groups of blocks but if any one type (MC or data) has multiple datasets blocks are mixed together with different dataset. I am not sure that would be the problem or Pileup is restricted only one dataset per type.

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
ticoann: I am not sure why but there were cases we got UNKNOWN or None before. (The test I did contain this case though)

ticoann:
@amaltaro, Alan please review. Add caching (Not sure how effective this will be - caching only within the workfow. we could improve that by setting cache directory in different location. But we need a policy for deleting cache. One other problem is caching is over whole pileup dataset if multiple dataset are defined. 
It also have hash collision possibility on cache file name.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -800,3 +800,15 @@ def checkBlockName(self, blockName):
...
             raise DBSReaderError("Invalid Block name: => %s <=" % blockName)
+        
+    def getFileListByDataset(self, dataset, detail=True):
+        
+        try:
+            fileList = self.dbs.listFileArray(dataset=dataset, validFileOnly = 1, detail=detail)
+            return fileList
+        except dbsClientException as ex:
+            msg = "Error in "
+            msg += "DBSReader.getFileDetailByDataset(%s)\n" % datasetName
```
amaltaro: bad copy&paste

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
ericvaandering: I don't know why you would have multiple datasets, but once you do having the dict of blocks outside of the loop seems correct to me. I'm probably missing something....

amaltaro:
BTW, it seems we do one PU lookup for each step (CMSSW) in the workflow, I noticed it today with a ReDigi wf with 2 steps.

ticoann:
Yes this patch should take care of that but this won't be much help if different workflows has the same PU.

amaltaro:
If the dataset query takes < 1min, then I think acquiring ~60 DR workflows per hour would be more than enough.

However, if we want to cover all the cases, then I'd suggest creating a cacheDir under WorkQueueManager/ and name the cache file with the PU dataset name (replacing / by__ or something like that). This issue is basically fixed in the current production agent, so we are not in a hurry to fix it in master :-)

ticoann:
It is little more complicated to have cache by dataset since it allows multiple dataset and multiple types. (mc, data) in the the same workflow. (Although I didn't see any of that case), and creates one json file.
Theoretically 
1. we have combine all those with one cache. which would be not very effective and sure it will go over file name limit 255 bytes.
2. We can create the cache for each dataset  create new json on the fly. (This could take care of the case when different workflows (each has multiple PU datasets) share some datasets but not all.  
I would implement for 2. but I am not sure that is legitimate use case.

amaltaro:
I'm not sure I understood your point.
> 2. We can create the cache for each dataset create new json on the fly. (This could take care of the case when different workflows (each has multiple PU datasets) share some datasets but not all.

Did you mean several steps/tasks using the same dataset?
In general, we better cover 2 use cases:
 a) same PU dataset in the same workflow, just different tasks/steps
 b) same PU dataset in different workflows (same campaign, etc)

The dataset type does not really matter (mc or data), and we don't even need to store it in any cache, only the json with the useful information.

ticoann:
I am considering case if there are multiple PU datasets are needed for Pileup (not sure whether there is the case). In that case we have cache not the final json which is saved under step directory but have to cache each dataset. and combine them when it gets saved in step.

i.e. Workflow A uses a, b PU datasets (if it is allowed to use multiple pileup datasets) for its processing.
If and workflow B uses a, c PU datasets. So when workflow B is procesed a wouldbe cached but not c. but final json file should contain both a, c info.

So it needs to support that case if it is legitimate use case. (In the case b) you mentioned)

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
```diff
@@ -70,18 +75,90 @@ def _queryDbsAndGetPileupConfig(self, stepHelper, dbsReader, fakeSites):
...
+        
+        directory = filePath.rsplit('/', 1)[0]
+        
+        if not os.path.exists(directory):
+            os.mkdir(directory)
+        try:
+            f = open(filePath, 'w')
+            f.write(json)
+        except IOError:
+            m = "Could not save pileup JSON configuration file: '%s'" % fileName
```
amaltaro: there is no fileName here

amaltaro:
BTW, the way it;s right now it caches the PU only for the same workflow, right? cacheDirectory seems to be something like
install/wmagent/WorkQueueManager/cache/WORKFLOW_NAME/WMSandbox/
?

I think we should make it a general cache under, e.g.:
install/wmagent/WorkQueueManager/cache/pileupCache

PR #6464, opened 2015-12-16
First draft at tracking blow-up factor.

bbockelm:
We take blow-up factor to be the number of additional jobs in a
task chain after the jobs in the first task.  The latter steps in
the task chain should affect what is acquired for work now - esp
as the blow-up factor may be >100 in the cases of LHE.

This approach is not foolproof, I think.  WQE's already in WMBS
may only count the jobs currently in WMBS still.  If nothing else,
this will slow the acquisition of such workflows to the point where
they are load-balanced across agents.  Not clear that it would
prevent an agent from taking on too much work if it's a single-agent
team.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
@@ -220,6 +220,20 @@ class TaskChainWorkloadFactory(StdBase):
         StdBase.__call__(self, workloadName, arguments)
         self.workload = self.createWorkload()
 
+        # Detect blow-up factor from first task in chain.
+        blowupFactor = 0
+        if (self.taskChain > 1) and 'TimePerEvent' in arguments["Task1"]:
+            origTpe = arguments["Task1"]['TimePerEvent']
+            if origTpe <= 0: origTpe = 1.0
+            sumTpe = 0
+            tpeCount = 0
+            for i in range(1, self.taskChain + 1):
+                if 'TimePerEvent' in arguments["Task%d" % i]:
+                    sumTpe += arguments["Task%d" % i]['TimePerEvent']
+                    tpeCount += 1
+            if tpeCount > 0:
+                blowupFactor = sumTpe / float(origTpe)
+
         for i in range(1, self.taskChain + 1):
 
             originalTaskConf = arguments["Task%d" % i]
@@ -252,13 +266,15 @@ class TaskChainWorkloadFactory(StdBase):
                 if isGenerator(arguments):
                     # generate mc events
                     self.workload.setWorkQueueSplitPolicy("MonteCarlo", taskConf['SplittingAlgo'],
-                                                          taskConf['SplittingArguments'])
+                                                          taskConf['SplittingArguments'],
+                                                          BlowupFactor=blowupFactor)
                     self.workload.setEndPolicy("SingleShot")
                     self.setupGeneratorTask(task, taskConf)
                 else:
                     # process an existing dataset
                     self.workload.setWorkQueueSplitPolicy("Block", taskConf['SplittingAlgo'],
-                                                          taskConf['SplittingArguments'])
+                                                          taskConf['SplittingArguments'],
+                                                          BlowupFactor=blowupFactor)
                     self.setupTask(task, taskConf)
                 self.reportWorkflowToDashboard(self.workload.getDashboardActivity())
             else:
--- a/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
@@ -71,7 +71,7 @@ class MonteCarlo(StartPolicyInterface):
                                  NumberOfEvents = nEvents,
                                  Jobs = jobs,
                                  Mask = copy(mask),
-                                 BlowupFactor = self.args['BlowupFactor']))
+                                 BlowupFactor = self.args['BlowupFactor'])
 
 
             if mask['LastEvent'] > (2**32 - 1):
--- a/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
@@ -28,6 +28,8 @@ class MonteCarlo(StartPolicyInterface):
         self.args.setdefault('SubSliceType', 'NumberOfEventsPerLumi')
         self.args.setdefault('SubSliceSize', self.args['SliceSize']) # events per lumi
         self.args.setdefault('MaxJobsPerElement', 250)  # jobs per WQE
+        self.args.setdefault('BlowupFactor', 0.0) # Estimate of additional jobs following tasks.
+                                                  # Total WQE tasks will be Jobs*(1+BlowupFactor)
 
         if not self.mask:
             self.mask = Mask(FirstRun = 1,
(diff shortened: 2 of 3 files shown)


src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
```diff
@@ -68,7 +70,9 @@ def split(self):
                                  NumberOfLumis = nLumis,
                                  NumberOfEvents = nEvents,
                                  Jobs = jobs,
-                                 Mask = copy(mask))
+                                 Mask = copy(mask),
+                                 BlowupFactor = self.args['BlowupFactor']))
```
amaltaro: It's failing in jenkins because the code is breaking here (additional closing parenthesis).

amaltaro:
There is one point that we need to carefully review here before merging which is the elements acquisition. If we have elements with a very big number of jobs (>25k, I foresee cases close to 100k), then it's possible that all the pending thresholds (including a queue_factor) won't give enough slots and so, the element will starve in GQ.

bbockelm:
The logic in `availableWork` is such that it will acquire new WQE's until _after_ it goes over.  It doesn't care if it goes over by a lot or a little.

So, if there's a WQE with 100k estimated jobs, a site with 99 pending jobs, and a limit of 100 pending jobs, it will acquire the WQE.

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -220,6 +220,20 @@ def __call__(self, workloadName, arguments):
         StdBase.__call__(self, workloadName, arguments)
         self.workload = self.createWorkload()
 
+        # Detect blow-up factor from first task in chain.
+        blowupFactor = 0
```
amaltaro: How about setting the default value to 1.0 instead?

src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
```diff
@@ -28,6 +28,8 @@ def split(self):
         self.args.setdefault('SubSliceType', 'NumberOfEventsPerLumi')
         self.args.setdefault('SubSliceSize', self.args['SliceSize']) # events per lumi
         self.args.setdefault('MaxJobsPerElement', 250)  # jobs per WQE
+        self.args.setdefault('BlowupFactor', 0.0) # Estimate of additional jobs following tasks.
```
amaltaro: How about setting the default value to 1.0 instead?

src/python/WMCore/WorkQueue/WorkQueueBackend.py
```diff
@@ -385,7 +385,7 @@ def availableWork(self, thresholds, siteJobCounts, teams = None, wfs = None):
                 elements.append(element)
                 if site not in siteJobCounts:
                     siteJobCounts[site] = {}
-                siteJobCounts[site][prio] = siteJobCounts[site].setdefault(prio, 0) + element['Jobs']
+                siteJobCounts[site][prio] = siteJobCounts[site].setdefault(prio, 0) + element['Jobs']*(1+element.get('BlowupFactor', 0.0))
```
amaltaro: How about setting the default value to 1.0 instead? Such that we get rid of the 1+ in this line

bbockelm: I'm reluctant to switch the defaults to 1.0; this way, there's less chance that coding errors would *shrink* the estimated number of jobs.

amaltaro: Sorry, I fail to see how it could be worse. Can you give me an example?

amaltaro:
test this please

amaltaro:
@bbockelm please fix the suggested "Pre-python 2.6 idioms" and this pylint
"Line 12 W0611 Unused CouchMonitor imported from WMCore.Database.CMSCouch"
message. Since it's one line change, you can squash all your changes in a single commit.

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -252,13 +266,15 @@ def __call__(self, workloadName, arguments):
...
+                                                          taskConf['SplittingArguments'],
+                                                          BlowupFactor=blowupFactor)
                     self.workload.setEndPolicy("SingleShot")
                     self.setupGeneratorTask(task, taskConf)
                 else:
                     # process an existing dataset
                     self.workload.setWorkQueueSplitPolicy("Block", taskConf['SplittingAlgo'],
-                                                          taskConf['SplittingArguments'])
+                                                          taskConf['SplittingArguments'],
+                                                          BlowupFactor=blowupFactor)
```
ericvaandering: I was going to comment on the case of this argument, but I guess it's passed through to MonteCarlo.py?

Upper case is definitely wrong, but the mistake is not yours.

amaltaro:
After some exhausting test, this is the summary:
 a) there are some minor fixes (blowUp calculation & jobCount calculation)
 b) this works for basically a single WQ cycle (but in practice it takes ~2 cycles - 20min), mainly during WQE negotiation. After the WQ has splitted and created subscriptions, the amount of slots taken goes to the standard value (considering only Jobs from WQE instead of * blowupFactor).

It won't solve this issue, but should at least make it much less annoying in our production agents.

I'm cooking an update patch and will make a PR later on.

bbockelm:
Yeah - that's where things get sticky!  It'll help spread out the workflow amongst several agents, but fundamentally too-big workflows will still get digested.

Do you think it would be possible to track the blow-up factor for the Jobs (i.e., once it's hit WMBS)?  That looked much more difficult.

amaltaro:
Possible it is :D, I just can't measure how much work it would take (maybe a new column in wmbs and etc ...).

amaltaro:
In case you want to take a look
https://github.com/amaltaro/WMCore/commit/a9802d2c15bc8068f6346fd99cb651f05025fd7b
should be final version, I hope.

bbockelm:
So, thinking over this, it seems tracking the blowup may just be a ton of work.  I have an idea of three simpler things we can do instead:

1. Refactor the way we count jobs per site (https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueBackend.py#L373).  Right now, if a work queue element has 500 jobs, the site has 99 jobs, and the site has a threshold of 100 jobs, then after the for-loop, the site will have 599 jobs.
   * Instead, we should add 1 job to the site and try to find locations for the 499.  We accept the block as long as at least 1 job was accepted (guaranteeing forward progress).  However, it is far less likely to overshoot its target by thousands of jobs (currently, it will take 1 WQE per iteration per site below-threshold).
   * Won't solve the current deficiency, but mitigate it.  Workflows with big blowup factors and big whitelists will pull down much slower on busy agents.
1.  Have a threshold for the number of jobs per agent in WMBS (all states, including cleanout); once this threshold is reached, stopped pulling work.  Based on experience, the threshold is somewhere around 1M for CERN agents and 500k for FNAL agents.
1.  Do something about the cleanout state.  Either get rid of jobs quicker or try a partition.

The latter two will help with all big workflows, regardless of whether they suffer from issues with blowup factors.

amaltaro:
Comments about your three points:
1. Sounds a good idea not to over-acquire work. However, WQE would be pulled all the time for > smaller sites since they would have a job or two finishing all the time. So it's not clear it would be that helpful.
2. Don't agree! We should not block data acquisition based on work done on that agent and waiting to be archived. Actually, we want to move the other way for quite a long time, which is getting rid of of the thresholds (tasks), so more thresholds will just over-complicate and give us trouble when debugging issues.

These are not simple changes and cannot be done today (for February CMSWEB). I'd suggest proceeding with the patch we've been working and validating and check how helpful it will be in real life.
IF it does not mitigate these issues, then we discuss and improve it again based on your comments.
@ericvaandering @ticoann what do you think?

amaltaro:
Since #6542 was merged, I'm closing this one.
Further discussion can be followed in https://github.com/dmwm/WMCore/issues/6544

PR #6465, opened 2015-12-16
1.0.12 wmagent

ticoann:
@amaltaro, Alan, could you double check the patch. I will create the new tag for agent

### DIFF
--- a/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
+++ b/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
@@ -29,6 +29,7 @@ from WMCore.DAOFactory        import DAOFactory
 from WMCore.WMBS.Fileset      import Fileset
 from WMCore.WMException       import WMException
 from WMCore.WMBS.Workflow     import Workflow
+from WMCore                   import Lexicon
 
 
 class JobArchiverPollerException(WMException):
@@ -146,7 +147,6 @@ class JobArchiverPoller(BaseWorkerThread):
                 myThread.transaction.rollback()
             raise JobArchiverPollerException(msg)
 
-
         return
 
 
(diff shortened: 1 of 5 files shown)


ticoann:
Alan, One more thing I would like to add is changing the block calls to dataset call in PileupFetcher class. I can add that one as on top of this. (We would still keep Brian's caching patch).
If you need to deploy new agent today, feel free to merge this and create a new tag 1.0.12.patch2 if the all the patch looks OK. Thanks

amaltaro:
Sounds good Seangchan! It includes: 
#6450 (without the last update to sendAlert I'll apply to master only. The right thing)
#6434 
#6457 

About the block to dataset, if you feel that's important for this release. Then we wait.

ticoann:
> About the block to dataset, if you feel that's important for this release. Then we wait.

The performance difference was 2h vs 1 minutes. It depends how many workflows left with this Pileup dataset.

amaltaro:
Ok, I think we can wait then :-P

ticoann:
added https://github.com/dmwm/WMCore/commit/61b216f62e8bf6a2fc3fe0196a2366c9a0dae2f5
Which is not tested. :smile: 
@amaltaro, please review and is possible to run test for this commit?

ticoann:
Also increase the cache expiration for 12 hours.

amaltaro:
Seangchan, this is the working version of your patch (https://github.com/dmwm/WMCore/commit/61b216f62e8bf6a2fc3fe0196a2366c9a0dae2f5):
https://github.com/amaltaro/WMCore/commit/b4388166d388bb22f7acddb60a87526e4a60a922

I made a few fixes, also backported Marco's fix for siteDB (#6466).

I compared the json created with the current version and the json with this patch using this ordered() method:
http://stackoverflow.com/questions/25851183/how-to-compare-two-json-objects-with-the-same-elements-in-a-different-order-equa
and other comparisons to make sure everything is ok.

I just removed my loggings and etc. And additional clean test shouldn't hurt :-)

ticoann:
Thanks Alan! I replace my  patch to your patch. Please check

amaltaro:
Hmmm, why my both redigi test workflows failed to find the secondary input... hold on on this one please Seangchan.

amaltaro:
Nevermind, it seems data got deleted from FNAL a few hours ago. I'm running another test, but I won't wait awake for the outcome :)
@ticoann feel free to merge it

ticoann:
Thanks Alan, 1.0.12.patch2 is created. https://github.com/dmwm/WMCore/tree/1.0.12.patch2

amaltaro:
Thanks Seangchan! I'm creating a PR for this patch for both comp_gcc481 and comp_gcc493 architectures.

PR #6469, opened 2015-12-16
Update dependencies for crabcache

mmascher:


ericvaandering:
Ok, you will also have to learn to ignore Jenkins tests on commits to branches other than master. On my list of things to figure out, but they don't seem to work well.

PR #6470, opened 2015-12-16
DBS Mock without the DBS Gather function

ericvaandering:
This PR should be merged if it goes OK. This uses mock and switches DBSReader to use it.

It's the basis for what @jha2 is working on.

I'll get to the bit that produces the JSON files later after I refactor it according to @vkuznet comments.


ericvaandering:
test this please

src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py
```diff
@@ -5,19 +5,74 @@
...
+            :param kwargs: named arguments it was called with
+            :return: the dictionary that DBS would have returned
+            """
+            if kwargs:
+                signature = '%s:%s' % (item, sorted(kwargs.iteritems()))
+            else:
+                signature = item
+
+            try:
+                if mockData[self.url][signature] == 'Raises HTTPError':
```
vkuznet: It would be better to check self.url and later signature keys exists in mockData. This is potential error when dict will not contains those keys.

ericvaandering: That should throw a KeyError. Which is caught 4 lines below.

ericvaandering:
test this please

ericvaandering:
@ticoann seems like all these errors are with WQ and are irrelevant. Can you merge?

PR #6474, opened 2015-12-16
Program to create mock DBS JSON files

ericvaandering:
This is what was called DBSGather in a previous incarnation of this setup. Valentin suggested (wisely) putting the parameters for the data that needed to be retrieved into JSON. Turns out that's problematic because son.load returns all strings as unicode and that might not match the function signatures (sometimes we call things with blocks and filenames in unicode and sometimes not).

So instead I factored out the data into some importable python so we at least have separation of code and data.


ericvaandering:
@ticoann @vkuznet @jha2 take a look at this if you like. Can be merged as it doesn't touch any libraries at all.

PR #6475, opened 2015-12-17
Fix status validation in reqmgr2

amaltaro:
Fix request api validation when > 1 status is provided.

Pylint was also complaining about
E:363,36: Instance of 'Request' has no 'wmstatsCouch' member (no-member)

Since nothing else in the code uses that method (and I actually don't think wmstats method belong to this class), I simply removed it. @ticoann let me know if we need to fix and add it back.


amaltaro:
If this code is Ok, I still want to do a pep-8 cleanup :)

ticoann:
Thanks Alan, patch looks good. For not used code, it is there because I was planing to have wmstats request record (job information) in the API. but I think it would be better to put in wmstats server and access reqmgr db record there. (So deleting is fine).

amaltaro:
Thanks Seangchan. Made another cleaning commit, but didn't remove some unused arguments because I believe you're either following some layout or just planning to use them in the future :-)

amaltaro:
@ticoann feel free to merge this one.

PR #6477, opened 2015-12-17
Fix data acquisition for TrustSitelists

amaltaro:
DO NOT MERGE yet, please.
When we assign work to opportunistic sites (aka., without PNN), we cannot force file wmbs registration against the site white list (because the mapping to PNN is empty and then it fails to acquire work).

I'm almost sure it will break our current "useSiteListAsLocation" mode, ideally we should merge them and live with a single case.

This parameter cannot be overriden during assignment (only request creation), so I'm going to fix that next within this patch.


hufnagel:
Yes, this will break useSiteListAsLocation, but why do you need to make this change ? Can't we treat useSiteListAsLocation and trsutSiteLists as separate things at the request level ? Do they really have to be tied up like this ?

amaltaro:
Yes, we can! However, since their results is basically the same, I'm in favor of simplifying the code and have (or try to) a single flag for these things.

amaltaro:
Closing it in favor of #6566

PR #6479, opened 2015-12-17
Grouper patch

ticoann:
replace #6459. @ericvaandering, Eric I add the dependency for rpm and JobArchiver change over your patch. Please review

### DIFF
--- a/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
+++ b/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
@@ -12,6 +12,7 @@ import shutil
 import tarfile
 import traceback
 
+from Utils.IterTools import grouper
 from WMComponent.TaskArchiver.TaskArchiverPoller import uploadPublishWorkflow
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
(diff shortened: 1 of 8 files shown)


src/python/WMComponent/JobArchiver/JobArchiverPoller.py
```diff
@@ -12,6 +12,7 @@
 import tarfile
 import traceback
 
+from Utils.IterTools import grouper
 from WMComponent.TaskArchiver.TaskArchiverPoller import uploadPublishWorkflow
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.JobStateMachine.ChangeState import ChangeState
```
amaltaro: remove the import Lexicon please

ticoann: Thanks Alan, it is removed now.

ericvaandering:
So, if you do it right you can make a PR against another developer's branch. :-)

ticoann:
> So, if you do it right you can make a PR against another developer's branch. :-)

Oh, I didn't think about that I will try next time.

PR #6481, opened 2015-12-17
Adding unit test using mock for DbsApi.

jha2:
Adding unit test using mock for DbsApi


ericvaandering:
This is @jha2 's unit test that actually checks DBS vs. the mock data. Eventually this should be the only test that uses DBS.

jha2:
Implemented the changes according to pylint.

ticoann:
test this please

PR #6483, opened 2015-12-18
Make deployment compatible with py2.7/gcc493

amaltaro:
Adjust the path for applying patches, basically.


amaltaro:
Only deployment script, merging

PR #6486, opened 2015-12-19
Do not use nCores for job wall time estimation

amaltaro:
At the moment, ReqMgr assumes the TimePerEvent provided for a workflow/task/step takes into account the number of cores. E.g., if we receive {'TimePerEvent': 4, 'Multicore': 8},  then it means one event will take 4 seconds to be processed in an 8 cores job, so the automatic 8h job splitting would set 7200 events per job.
However, when JobCreator is creating these jobs, it assumes _each_ core takes 4seconds to process 1 event, so calculating a ~2h wall time.

Almost sure unit tests need to be fixed...


amaltaro:
This is wrong, don't merge!

PR #6487, opened 2015-12-19
Add Signals dependency

amaltaro:
I just have problems deploying services in my VM
ImportError: No module named Signals
and I don't see this module under the services path.

First time I change this file, but I believe this is the right fix.
@ticoann please review


amaltaro:
And while workflow creation, reqmgr was complaining about import Utils as well.
This patch should fix that, though I didn't test and will probably not have time to test it in the coming days.

ticoann:
Thanks Alan, I moved the Signals.py under Utils and reset the dependency #6491

amaltaro:
Thanks Seangchan!

PR #6493, opened 2016-01-07
Relax constrain on passed args (when no YUI is used)

vkuznet:
When I worked with WMArchive I found that constraint in REST Service serving static content is too tight. It was done when we used YUI library. Now, I wanted to serve just set of static files, e.g. wmarchive/static/images/img.jpg, etc. and to do so I loose constrain on passed args.

### DIFF
--- a/src/python/WMCore/REST/Server.py
+++ b/src/python/WMCore/REST/Server.py
@@ -320,7 +320,7 @@ class RESTFrontPage:
 
         Serves assets as documented in :meth:`_serve`."""
         if len(args) > 1 or (args and args[0] != "yui"):
-            raise HTTPError(404, "No such file")
+            return self._serve(['/'.join(args)])
         paths = request.query_string.split("&")
         if not paths:
             raise HTTPError(404, "No such file")
--- a/src/python/WMCore/REST/Server.py
+++ b/src/python/WMCore/REST/Server.py
@@ -319,8 +319,12 @@ class RESTFrontPage:
         with 'yui/' to make them compatible with the standard combo loading.
 
         Serves assets as documented in :meth:`_serve`."""
+        # The code was originally designed to server YUI content.
+        # Modified to support any content by joining args into single path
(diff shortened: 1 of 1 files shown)


vkuznet:
@ticoann, Seangchan, I'm really puzzled by this test failure. The change I made
is for frontend page to serve static files. The failed test is for get jobs
from Retry Manager. Why it fails and how web interface is related to jobs
retrieval? Can you have a look at unit test failure.

ticoann:
Hi Valentin, you are right. This patch is not related to the failures reported. Must be unstable test.
Anyway, could you update the comment above the code as well which state
```
Path arguments must be empty, or consist of a single 'yui' string,
```

src/python/WMCore/REST/Server.py
```diff
@@ -320,7 +320,7 @@ def static(self, *args, **kwargs):
 
         Serves assets as documented in :meth:`_serve`."""
         if len(args) > 1 or (args and args[0] != "yui"):
```
ticoann: I am not quite under stand this condition (although this is not what you changed). are we just supporting multiple file loading only when "yui" keyword is provided in the arguments?

vkuznet: On  0, ticoann <notifications@github.com> wrote:
> > @@ -320,7 +320,7 @@ def static(self, *args, **kwargs):
> >  
> >          Serves assets as documented in :meth:`_serve`."""
> >          if len(args) > 1 or (args and args[0] != "yui"):
> 
> I am not quite under stand this condition (although this is not what you changed). are we just supporting multiple file loading only when "yui" keyword is provided in the arguments?

yes, I think condition reads as if multiple arguments are provided then only
those with yui will be served. This is what I found "weird", but when code was
developed we probably only had YUI. For WMArchive I wanted to support
js/file.js, css/file.css, image/file.png etc. content. That's why I changed the
code.

PR #6494, opened 2016-01-07
add CondorStatusService

hufnagel:
Requested by Brian, already in use for CRAB3. As far as I understand it makes CMSSW publish job progress information to Condor, that information can then be remotely queried.

Also exercised in Tier0 tests, no issues observed.

If desired I can factor this out into its own method, functionally wouldn't be any different though.

@ericvaandering and @ticoann , please review.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -6,6 +6,7 @@ Create a CMSSW PSet suitable for running a WMAgent job.
 """
 
 import os
+import re
 import random
 import socket
 import traceback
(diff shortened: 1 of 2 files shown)


src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -598,6 +598,11 @@ def __call__(self):
                 print traceback.format_exc()
                 raise ex
 
+            self.process.add_(
+                cms.Service("CondorStatusService"
+                            )
+                )
```
ericvaandering: This all could certainly go on one line.

hufnagel: Agreed

hufnagel:
Brian just made a good point. We have to check what happens with CMSSW releases that don't support the CondorStatusService.

hufnagel:
Modified the patch to be CMSSW version specific. Likely needs further testing now.

hufnagel:
retest this please

ericvaandering:
This is going to fail unit tests, I bet, since CMSSW_VERSION is not defined. I'd like to see protection against that variable being empty and a unit test which puts it into the environment and tests it.

hufnagel:
Moving the call to handleCondorStatusService() makes sense as we want it in all configs, not just the Config.DataProcessing created ones. Also added protections against CMSSW_VERSION not being in the environment. Unit test was also added, although I don't actually check what is done to the configuration, just that it works with various values of CMSSW_VERSION. Don't know how I can check if a process contains a certain service tbh.

ericvaandering:
" I don't actually check what is done to the configuration, just that it works with various values of CMSSW_VERSION. Don't know how I can check if a process contains a certain service tbh."

Good enough for me. It's probably possible, but anything to do with CMSSW configs outside of the CMSSW environment is a PITA.

hufnagel:
Ok, then this pull request is done. Not sure why I get the 'all checks have failed', maybe for the pylint complains ?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -574,21 +575,41 @@ def handleSpecialCERNMergeSettings(self, funcName):
         Enable lazy-download for fastCloning for all CMSSW_7_5 jobs (currently off)
         Enable lazy-download for all merge jobs
         """
-        cmsswVersion = os.environ['CMSSW_VERSION']
+        cmsswVersion = os.environ.get('CMSSW_VERSION', "")
         if cmsswVersion.startswith("CMSSW_7_5") and False:
```
ticoann: Hi Dirk, I know you didn't make change on this. Could you explain me again why this condition (always False) need to be set.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -574,21 +575,41 @@ def handleSpecialCERNMergeSettings(self, funcName):
...
+            self.process.add_(cms.Service("SiteLocalConfigService",
+                                          overrideSourceCacheHintDir = cms.untracked.string("lazy-download")))
+        return
+
+    def handleCondorStatusService(self):
+        """
+        _handleCondorStatusService_
+
+        Enable CondorStatusService for CMSSW releases that support it.
+
```
ticoann: Hi Dirk, so following are supported? (7_5_0 is not supported)?
If there is document available could you add the link on the comment, just for the reference?
```
    CMSSW_8_*_* and up 
     CMSSW_7_6_* and up 
     CMSSW_7_5_1 and up
     CMSSW_7_4 _7 and up
```

Maybe you could have condition like befow more clarity

``` python
verson = (result.group(1) *100 + result.group(2)* 10 + result.group(3))
if  version >= 747 and version != 750: 
    addCondorStatusService = True
```

ericvaandering: That's dangerous, 7.4.10 would be marked as not supported.

ericvaandering:
Right Dirk, SetupCMSPSet fails the pylint check outright (not got worse, just is so bad according to pylint).

So pylint should not hold up the merge in this case.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
ticoann: Ah, yes that is right

hufnagel: There is no document that lists what is supported, I just browsed the CMSSW code repository and checked the versions that have the CondorStatusService service plugin.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
hufnagel: It was Brians automatic "if job uses fastCloning then enable lazy download" feature which sounded great in theory, but wasn't actually working in practice (most likely because the check if a job was using fastCloning was too expensive and therefore it spend too much time on that before it had a chance to switch to lazy download).

I should probably remove this completely, although don't want to do this as part of this pull request.

PR #6360, opened 2015-10-26
Changes for pycurl, Requests and service classes to use pycurl by default (do not merge)

vkuznet:

PR #6495, opened 2016-01-07
Changes to fix RequestManager_t unittest and migration to pycurl

vkuznet:
Make sure that Requests that  use POST/PUT requests send values in a body of the request instead of positional arguments

### DIFF
--- a/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py
+++ b/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py
@@ -274,8 +274,8 @@ class ReqMgrRESTModel(RESTModel):
         else:
             try:
                 result = Utilities.requestDetails(requestName)
-            except: # if we fail to get a request return nothing
-                return
+            except cherrypy.HTTPError:
+                return # if we fail to get a request return nothing
             try:
                 teamNames = GetRequest.getAssignmentsByName(requestName)
                 result['teams'] = teamNames
--- a/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py
+++ b/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py
@@ -272,7 +272,10 @@ class ReqMgrRESTModel(RESTModel):
         if requestName == None:
             result = GetRequest.getRequests()
         else:
-            result = Utilities.requestDetails(requestName)
+            try:
+                result = Utilities.requestDetails(requestName)
+            except: # if we fail to get a request return nothing
+                return
             try:
                 teamNames = GetRequest.getAssignmentsByName(requestName)
                 result['teams'] = teamNames
@@ -611,7 +614,14 @@ class ReqMgrRESTModel(RESTModel):
 
     def putMessage(self, request):
         """ Attaches a message to this request """
-        message = JsonWrapper.loads( cherrypy.request.body.read() )
+        body = cherrypy.request.body.read()
+        obj = JsonWrapper.loads( body )
+        if isinstance(obj, dict):
+            message = obj.get('message', '')
+        elif isinstance(obj, basestring):
+            message = obj
+        else:
+            raise Exception("Unsupported message type, msg=%s, type=%s" % (obj, type(obj)))
         result = ChangeState.putMessage(request, message)
         return result
 
--- a/src/python/WMCore/WebTools/RESTModel.py
+++ b/src/python/WMCore/WebTools/RESTModel.py
@@ -185,6 +185,14 @@ class RESTModel(WebAPI):
             if len(args) != 0:
                 # store the method name
                 method = methodKey
+                # for PUT/POST requests data may be send in body of the request
+                # we check if this is the case (no args/kwargs) and read body
+                if  verb in ['PUT', 'POST'] and \
+                    not input_args and not input_kwargs:
+                    from WMCore.Wrappers import JsonWrapper
+                    body = cherrypy.request.body.read()
+                    if  body:
+                        input_kwargs = JsonWrapper.loads( body )
                 input_data = self._sanitise_input(input_args, input_kwargs, method)
                 # store the function
                 func = function
--- a/test/python/WMCore_t/Services_t/RequestManager_t/RequestManager_t.py
+++ b/test/python/WMCore_t/Services_t/RequestManager_t/RequestManager_t.py
@@ -64,7 +64,10 @@ class RequestManagerTest(RESTBaseUnitTest):
 
         try:
             r = self.jsonSender.put('request', schema)
-            self.requestName = r[0]['RequestName']
+            try:
+                self.requestName = r[0]['RequestName']
+            except:
+                self.requestName = r[0].values()[0]['RequestName']
         except Exception as ex:
             msg = traceback.format_exc()
             print "Exception during set up, reason: %s" % msg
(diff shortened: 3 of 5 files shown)


src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py
```diff
@@ -272,7 +272,10 @@ def getRequest(self, requestName=None):
         if requestName == None:
             result = GetRequest.getRequests()
         else:
-            result = Utilities.requestDetails(requestName)
+            try:
+                result = Utilities.requestDetails(requestName)
+            except: # if we fail to get a request return nothing
```
ticoann: Valentin, why this exception occurs?
Is it due to passing the argument in request  body?

Earlier message you send through email was following. 
```
Once I fixed that like this, I replaced

result = Utilities.requestDetails(requestName)

with

           try:
               result = Utilities.requestDetails(requestName)
               print("### getRequest, result", result)
           except Exception as exp:
               print("### getRequest, exp", str(exp))
               return

my PUT request proceeds and I got running unittest. It fails now in another place

DEBUG:cherrypy.error:call to PUT with args: ['message', 'Taizong_160106_101656_8214'] kwargs: {} resulted in cannot parse JSON description: error

                                             stack trace: Traceback (most recent call last):
 File "/mnt/disk1/vk/WMCoreTests/unittestdeploy/wmagent/1.0.8.pre7/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.8.pre7/lib/python2.6/site-packages/WMCore/WebTools/RESTApi.py", line 103, in _buildResponse
   args, kwargs)
 File "/mnt/disk1/vk/WMCoreTests/unittestdeploy/wmagent/1.0.8.pre7/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.8.pre7/lib/python2.6/site-packages/WMCore/WebTools/RESTModel.py", line 117, in handler
   data = methodCall(*params, **kwargs)
 File "/mnt/disk1/vk/WMCoreTests/unittestdeploy/wmagent/1.0.8.pre7/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.8.pre7/lib/python2.6/site-packages/WMCore/WebTools/RESTModel.py", line 192, in wrapper
   return func(**input_data)
 File "/mnt/disk1/vk/WMCoreTests/unittestdeploy/wmagent/1.0.8.pre7/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.8.pre7/lib/python2.6/site-packages/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py", line 629, in putMessage
   message = JsonWrapper.loads( cherrypy.request.body.read() )
 File "/mnt/disk1/vk/WMCoreTests/unittestdeploy/wmagent/1.0.8.pre7/sw.pre/slc6_amd64_gcc481/cms/wmagent/1.0.8.pre7/lib/python2.6/site-packages/WMCore/Wrappers/JsonWrapper/__init__.py", line 40, in loads
   raise ex
DecodeError: cannot parse JSON description: error
```

vkuznet: Seangchan,
The exception I got is HTTPERror 404, 'Given requestName not found'.
May be I should be explicit and catch HTTPError exception only.
But the point is that requestDetails call can fail for whatever reason and
since nobody cares in this call how to handle it the entire stack will fail
elsewhere. If you know better behavior for this let me know and I can update
PR.
V.

ticoann: I see so this is not just for pycurl patch but current code as well. Yes I think it is better to catch 404 error other error might be server side error.

vkuznet: Ok, let me change that in PR.

vkuznet: Done.

test/python/WMCore_t/Services_t/RequestManager_t/RequestManager_t.py
```diff
@@ -123,13 +126,12 @@ def testA_RequestManagerService(self):
         
         self.reqService.sendMessage(requestName,"error")
         self.reqService.putWorkQueue(requestName, "http://test_url")
-        self.reqService.reportRequestProgress(requestName)
         self.reqService.reportRequestProgress(requestName,
                         percent_complete = 100, percent_success = 90)
         
-        self.reqService.reportRequestStatus(requestName, "running-open")
+        self.reqService.updateRequestStatus(requestName, "running-open")
```
ticoann: Does this need to be changed to updateRequestStatus?

vkuznet: Well, by logic it should be update request status. By code I think now they do
the same (reportRequsetStatus and updateRequestStatus). You better look at
RequestManager code and decide which method to use and keep. The difference
is that updateRequestStatus uses request/<REQUESTNAME> call while
reportRequestStatus uses request call to REST server.

src/python/WMCore/WebTools/RESTModel.py
```diff
@@ -185,6 +185,14 @@ def wrapper(*input_args, **input_kwargs):
...
                 # store the method name
                 method = methodKey
+                # for PUT/POST requests data may be send in body of the request
+                # we check if this is the case (no args/kwargs) and read body
+                if  verb in ['PUT', 'POST'] and \
+                    not input_args and not input_kwargs:
+                    from WMCore.Wrappers import JsonWrapper
+                    body = cherrypy.request.body.read()
+                    if  body:
+                        input_kwargs = JsonWrapper.loads( body )
```
ticoann: Hi Valentin, actually this block is causing the problem.  when it is converted input or input_kwargs it is automatically checks the number of argument with the corresponding method. Basically params passed in the body need to be handled each method not in the RESTModel. This will break all the calls using WebTools/RESTModel.py (including DBS calls)

vkuznet: Seangchan,
I'm not sure I follow. Can you provide concrete example, just write it, instead
of pointing to some specific code.

The if statement checks that no input_args and input_kwargs is given. It means
that somebody doing the following call

PUT /path/bla

and provides in body some parameters, e.g. {"foo":1}

ticoann: Sorry I wasn't clear. What I meant is that following line checks the number of argument defined by method and compare them what is passed.

https://github.com/vkuznet/WMCore/blob/9a68173916e5ef994f0453cabbfce7077f322648/src/python/WMCore/WebTools/RESTModel.py#L241

For the parameters passed by body, might not defined in the function definition.
i.e.
https://github.com/vkuznet/WMCore/blob/9a68173916e5ef994f0453cabbfce7077f322648/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrRESTModel.py#L133

only take one argument but if it contains message body with other parameter (i.e. message) it will fail.

vkuznet: I think it is another issue with current implementation of RESTModel.
The POST/PUT requests can be done without any explicit arguments and client can
send request in a body.

The logic of RESTModel should adapt to this generic statement.

The line
https://github.com/vkuznet/WMCore/blob/9a68173916e5ef994f0453cabbfce7077f322648/src/python/WMCore/WebTools/RESTModel.py#L241

does not fit in generic REST model approach. It is custom to our
implementation. And, I think we should generalize it accordingly.
Moreover, line 242 and 243 is common source of misleader when something goes
wrong. The HTTP error is probably fine, but self.debug statements never
appears in log unless someone explicitly configure to use DEBUG level.

I would suggest to re-do implementation of _sanitise_input with respect to
generic REST model approach, i.e. client should be allowed to send
both POST/PUT requests with explicit arguments and using body.

My original change is at lowest level which allows to adjust
_sanitise_input implementation.

ticoann: Although I am not comfortable changing the RESTModel code since we don't have full control apis using this code.
I think relaxing the restriction on _sanitise_input will be OK.
I am still getting errors with this patch though with suggested change. Error messages are reported #6360

PR #6498, opened 2016-01-13
Update env_unittest.sh

ericvaandering:
We need to get the new WMCore directly appended to the path too. Otherwise it pulls from the RPM version of the source and won't reflect user changes.


ericvaandering:
Wrong way to solve this

PR #6504, opened 2016-01-14
signal.SIGINFO doesn't exist in signal module

ticoann:
For now remove the import. Also import should be import Utils.Signals

### DIFF
--- a/bin/wmc-httpd
+++ b/bin/wmc-httpd
@@ -1,4 +1,4 @@
 #!/usr/bin/env python
 from WMCore.REST.Main import main
-import WMCore.Signals
+import Utils.Signals
 main()


vkuznet:
Seangchan, please don't remove, it is problem of our outdated python version. The signals.SIGINFO exists in python 2.7. What should be done is to downgrade SIGINFO to SIGUSR1 instead. I'll make appropriate PR.

vkuznet:
Please use this PR: https://github.com/dmwm/WMCore/pull/6507 where I changed SIGINFO to SIGUSR1

ticoann:
Thanks Valentin, I put Signals back

PR #6508, opened 2016-01-14
add a few missing init.py files

hufnagel:


hufnagel:
@ticoann , can you please merge ?

ericvaandering:
You can ignore the future stuff for init.py, but you have to fix the relative import part.

ticoann:
I am not sure why this is causing the problem. since the package is not called. But merging this.

PR #6509, opened 2016-01-14
change default arch to gcc493

ticoann:


bin/deploy-wmagent.sh
```diff
@@ -37,8 +37,8 @@ OP_EMAIL=cms-comp-ops-workflow-team@cern.ch
 HOSTNAME=`hostname`
 
 # These values may be overwritten by the arguments provided in the command line
-WMA_ARCH=slc6_amd64_gcc481
-REPO="comp=comp.pre"
+WMA_ARCH=slc6_amd64_gcc493
+REPO="comp=comp”
```
amaltaro: this close double quotes looks weird

ticoann: Thanks Alan, good catch!. not sure I how it got there, updated now.

PR #6512, opened 2016-01-15
Do not call listFileArray with run_num=1

amaltaro:
Turns out the change is much easier than I initially thought - unless I got it wrong :)

Still needs to be tested, planning to test it with pre1 + Seangchan patches (pre2).
@yuyiguo @ticoann please review

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/Block.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/Block.py
@@ -5,11 +5,8 @@ WorkQueue splitting by block
 """
 __all__ = []
 
-from WMCore.WorkQueue.Policy.Start.StartPolicyInterface import StartPolicyInterface
-
 from math import ceil
-
-from Utils.IterTools import grouper
+from WMCore.WorkQueue.Policy.Start.StartPolicyInterface import StartPolicyInterface
 from WMCore.Services.SiteDB.SiteDB import SiteDBJSON as SiteDB
 from WMCore.DataStructs.LumiList import LumiList
 from WMCore.WorkQueue.WorkQueueExceptions import WorkQueueWMSpecError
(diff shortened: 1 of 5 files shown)


src/python/WMCore/WorkQueue/Policy/Start/Block.py
```diff
@@ -227,8 +227,13 @@ def getMaskedBlocks(self, task, dbs, datasetPath):
             
             for slumis in grouper(lumis, 1000):
                 for block in blocks:
-                    slicedFiles = dbs.dbs.listFileArray(block_name=block.get('block_name'), run_num=run,
-                                                        lumi_list=slumis, detail=True)
+                    # run_num=1 will not be allowed anymore in this API (only for LFNs, not blocks
+                    # nor dataset), starting on Feb/2016
+                    if run==1:
+                        slicedFiles = dbs.dbs.listFileArray(block_name=block.get('block_name'), lumi_list=slumis, detail=True)
```
ticoann: Yuyi, is this call allowed?

ticoann: Hi Alan, Yuyi told me this call wouldn't work. when lumi_list is specified, run number has to be specified.
How about something like following?
https://github.com/dmwm/WMCore/compare/master...ticoann:dbs_api_change?expand=1

amaltaro: I like your patch :-) I'll build pre4 and run a quick test by tomorrow. If things work fine, you can make a PR and close this one.
Thanks a lot!

ticoann: Thanks Alan, If the patch looked OK. I can run the test. Don't work too hard. :smile:

amaltaro: Don't worry Seangchan. Yes, I'm running a test against pre5 and found a couple of issues already, still a few more to sort out. I also want to test lfns with more than one run number, the current patch would break.

Maybe I can drop an email to the HTTP group and say we'll deliver new WMCore tags only tomorrow afternoon. It should not be an issue and our crab friends to it all the time :-D

ticoann: Yes, I think we need to do that. I will send them email. Thanks, Alan

amaltaro: Now stop working and enjoy what is left of the US holiday, there are only a few of them :-P

amaltaro:
I'm not sure, but I think these two Block tests failed due to the (un)mocking dbs calls. I ran the same tests in my node and all these tests pass.
I just updated one line in the Block test, but it will likely not fix that unittest...

amaltaro:
Thanks Seangchan! This will need some careful check during validation.

ticoann:
Thank to you Alan, I will create a new tag tonight.

ericvaandering:
FWIW, Alan and I had a Slack discussion that a different output format is not needed since the LumiList class defines L1+L2, L1-L2, L1|L2, and L1&L2 for operations.

PR #6513, opened 2016-01-15
remove an external used that doesn't work with python 2.7

hufnagel:
Showed up in my comp_gcc493 T0 version that uses python 2.7. Easy enough to fix as you really don't need that helper function.

### DIFF
--- a/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
+++ b/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
@@ -12,7 +12,6 @@ import shutil
 import tarfile
 import traceback
 
-from Utils.IterTools import grouper
 from WMComponent.TaskArchiver.TaskArchiverPoller import uploadPublishWorkflow
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.JobStateMachine.ChangeState import ChangeState
@@ -152,7 +151,9 @@ class JobArchiverPoller(BaseWorkerThread):
         logging.info("Found %i finished jobs to archive", len(doneList))
 
         jobCounter = 0
-        for slicedList in grouper(doneList, 10000):
+        chunkSize = 10000
+        for slicedList in [ doneList[i:i+chunkSize] for i in range(0, len(doneList), chunkSize) ]:
+
             self.cleanWorkArea(slicedList)
 
             successList = []


ericvaandering:
What's the problem? I just wrote that function and as far as I know it should work just fine on python2 and python3.

hufnagel:
Starting : JobArchiver
  File "/data/tier0/srv/wmagent/2.0.1/sw.hufnagel/slc6_amd64_gcc493/cms/wmagent/1.0.13.pre1/bin/wmcoreD", line 347, in <module>
    startup(config)
  File "/data/tier0/srv/wmagent/2.0.1/sw.hufnagel/slc6_amd64_gcc493/cms/wmagent/1.0.13.pre1/bin/wmcoreD", line 220, in startup
    componentObject = factory.loadObject(classname = namespace, args = config)
  File "/data/tier0/srv/wmagent/2.0.1/sw.hufnagel/slc6_amd64_gcc493/cms/wmagent/1.0.13.pre1/lib/python2.7/site-packages/WMCore/WMFactory.py", line 68, in loadObject
    module = __import__(module, globals(), locals(), [classname])
  File "/data/tier0/srv/wmagent/2.0.1/sw.hufnagel/slc6_amd64_gcc493/cms/wmagent/1.0.13.pre1/lib/python2.7/site-packages/WMComponent/JobArchiver/JobArchiver.py", line 17, in <module>
    from WMComponent.JobArchiver.JobArchiverPoller import JobArchiverPoller
  File "/data/tier0/srv/wmagent/2.0.1/sw.hufnagel/slc6_amd64_gcc493/cms/wmagent/1.0.13.pre1/lib/python2.7/site-packages/WMComponent/JobArchiver/JobArchiverPoller.py", line 15, in <module>
    from Utils.IterTools import grouper
ImportError: No module named Utils.IterTools

hufnagel:
Not 100% sure why it doesn't work, but as it really isn't needed at all...

hufnagel:
Hm, Utils.IterTools is supposed to be from our own code base ?

ericvaandering:
Yes. It is in our code base. The problem is that it's a new top-level directory, so maybe something in your deployment is keeping it from being in the path.  First check that you have it present, then check the PYTHONPATH. 

We had some issue with this at first in WMAgent too.

Please fix the actual problem as this was added for a reason.

ticoann:
@hufnagel, Dirk, I will create the new tag with #6514. But please hold on to the rpm creation until I run the test. I will let you know.  Thanks,

PR #6515, opened 2016-01-15
add cache for pile up data and call by dataset instead of block

ticoann:
replace #6463


ericvaandering:
Can you fix up some of the coding issues?

ticoann:
I will fix some of pylint issue. I am still running the test on this patch.

PR #6521, opened 2016-01-16
allow to configure group ownership for a subscription

hufnagel:
In the subscription information record allow to configure a phedex group override. If not specified it defaults to None which triggers using the default component parameters (so DataOps).

So far this is only used by the Tier0, if desired for WMAgent we would need to make this configurable from RequestManager and pass the information down to where subscriptions are setup.


hufnagel:
@ticoann , can you review ? Tested both cases in Tier0 replay. Not configuring the new option (which the WMAgent will do) leads to a None being passed, a NULL entry being written into the database and then the component default being used (DataOps). Configuring a value passes it along to PhEDEx.

amaltaro:
changes look good to me. We still have to check how reqmgr handles this, in principle we need to pass a new argument during workflow assignment "PhEDExGroup". I'll build new RPMs tomorrow and test this patch.

PR #6337, opened 2015-10-17
Fix transaction logic; connection usage; use generator objects in DBCore; clean patch replacement of 6284

vkuznet:

PR #6522, opened 2016-01-17
Revert "Fix transaction logic; connection usage; use generator objects in DBCore; clean patch replacement of 6284"

ticoann:
Reverts dmwm/WMCore#6337

### DIFF
--- a/src/python/WMCore/Database/DBCore.py
+++ b/src/python/WMCore/Database/DBCore.py
@@ -7,6 +7,8 @@ Core Database APIs
 
 """
 
+
+
 from WMCore.DataStructs.WMObject import WMObject
 from WMCore.Database.ResultSet import ResultSet
 from copy import copy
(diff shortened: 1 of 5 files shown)


ticoann:
Reverting back due to the errors found in Tier0 test.
Here is Dirk's finding.

Noticed two failures:

1) select queries break if two column names are identical

https://github.com/dmwm/T0/blob/master/src/python/T0/WMBS/Oracle/RunConfig/GetStreamDatasetTriggers.py

As you see, the two columns selected have the same name, but in 
different tables. This didn't work anymore, it somehow got confused by 
the identical names and returned (column2,column2) instead of 
(column1,column2).

2) array inserts seems to sometimes drop some binds

https://github.com/dmwm/T0/blob/master/src/python/T0/WMBS/Oracle/RunConfig/InsertLumiSection.py

Passed lumis 1 to 100 (same run), it always skips inserting lumi 46.

Probably would fail in other places too, these are the two things I 
noticed (and which killed the Tier0 tests).

Cheers

       Dirk

PR #6523, opened 2016-01-17
Fix transition logic (Do not merge)

ticoann:
Reverts dmwm/WMCore#6522 (There are some problems reported with this patch in Oracle) Need further investigation.


vkuznet:
Seangchan,
can we setup necessary unittests for Oracle db and implement behavior of T0
queries in them?
V.

PR #6524, opened 2016-01-19
fix CMSSW version check for CondorStatusService

hufnagel:
Comparing strings to numbers technically works in python, but it doesn't really give the right results. Therefore cast the output of the regular expression matching to int before using it for comparisons.


hufnagel:
@ticoann , can you merge ? Any tag that contained the previous version of the CondorStatusService code should be considered broken and not be used.

PR #6529, opened 2016-01-20
custodial and non-custodial phedex groups for subscriptions

hufnagel:
The Tier0 makes multiple subscriptions for some datasets, one to MSS and one to Disk. These subscriptions should use different PhEDEx groups, which isn't currently possible. Add this feature by allowing to configure different PhEDEx groups for custodial and nonCustodial subscriptions.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/NewSubscription.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/NewSubscription.py
@@ -25,9 +25,11 @@ class NewSubscription(DBFormatter):
     def _createPhEDExSubBinds(self, datasetID, subscriptionInfo, custodialFlag):
         
         # DeleteFromSource is not supported for move subscriptions
+        phedex_group = None
         delete_blocks = None
         if custodialFlag:
             sites = subscriptionInfo['CustodialSites']
+            phedex_group = subscriptionInfo['CustodialGroup']
             if subscriptionInfo['CustodialSubType'] == 'Move':
                 isMove = 1
             else:
@@ -36,6 +38,7 @@ class NewSubscription(DBFormatter):
                     delete_blocks = 1
         else:
             sites = subscriptionInfo['NonCustodialSites']
+            phedex_group = subscriptionInfo['NonCustodialGroup']
             if subscriptionInfo['NonCustodialSubType'] == 'Move':
                 isMove = 1
             else:
@@ -51,7 +54,7 @@ class NewSubscription(DBFormatter):
                            'auto_approve' : 1 if site in subscriptionInfo['AutoApproveSites'] else 0,
                            'move' : isMove,
                            'priority' : subscriptionInfo['Priority'],
-                           'phedex_group' : subscriptionInfo.get('PhEDExGroup', None),
+                           'phedex_group' : phedex_group,
                            'delete_blocks' : delete_blocks} )
         return binds
     
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
@@ -46,7 +46,6 @@ class PhEDExInjectorPoller(BaseWorkerThread):
         self.config = config
         self.phedex = phedex
         self.dbsUrl = config.DBSInterface.globalDBSUrl
-        self.group = getattr(config.PhEDExInjector, "group", "DataOps")
 
         # This will be used to map SE names which are stored in the DBSBuffer to
         # PhEDEx node names.  The first key will be the "kind" which consists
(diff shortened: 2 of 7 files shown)


hufnagel:
@ticoann @amaltaro this will change the parameter names that need to be passed down from the request. Its tested in Tier0 code, I can do what I need to do now.

hufnagel:
Ah, yes, also need to adjust one of the unit tests.

ericvaandering:
@hufnagel I think that failure is worth looking into: https://cmssdt.cern.ch/jenkins/job/DMWM-WMCore-PR-test/1395/testReport/WMCore_t.WMSpec_t.WMTask_t/WMTaskTest/testSubscriptionInformation/

Very helpful error message though. :-)

amaltaro:
I assume Custodial and NonCustodial groups default to DataOps (in the configuration), right? If not, can you make sure it defaults to DataOps please?

hufnagel:
Wouldn't DataOps for Custodial and AnalysisOps for NonCustodial be better ?

hufnagel:
Also, things auto-default to DataOps since they default to None and None triggers using the old-style component default, which has DataOps as default. If you want me to put in real defaults here I would simplify things and get rid of the component defaults. Opinions ?

amaltaro:
Not sure about the defaults, maybe we keep them as they are right now (it may be just a bit confusing if we have component and component config defaults).

Since these are new spec parameters that the assigner will be able to provide during assignment, we need to add them here as well (otherwise I don't see a way to propagate it downstream):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1010

and workload updateArgs needs to be updated as well:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1690

hufnagel:
Ok, I can add them in these two places. Still need to make changes for the unit test as it fails anyways. I can make the defaults use 'DataOps' for everything now. Then they will be the same as the component config parameter defaults, which isn't needed anymore then.

amaltaro:
I was going to ask you to remove the default from the agent config:
https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py

but it seems not to be there. It has been used only by T0, most likely.

hufnagel:
No, the default is in the component code itself. If he agent config isn't there it uses DataOps and that is what we have been using in every WMAgent and the Tier0 so far.

hufnagel:
Btw the defaults in StdBase for nonCustodialSubType is Replica, which in WMWorkload and WMTask it's Move. Should I change that to make it Replica everywhere ? AFAIK we don't use Move anymore...

hufnagel:
Also, for now I just require type str for the phedex group, no further validation. There are a lot of phedex groups and I don't think we have easy access to the list to check.

amaltaro:
WMWorkload and WMTask should be using the StdBase default, no? We define Replica as default in the top code StdBase and just let it propagate to the workload and task.
Yeah, input validation in this case would be tricky,  I follow you in leaving it str only.

hufnagel:
WMWorkload and WMTask use the StdBase defaults but only indirectly and only for WMAgent operations (as we pull the parameters from the spec during assignment). Tier0 calls these directly, we don't care about what is in the specs (and I am not planning to even set the spec parameters).

hufnagel:
test this please

hufnagel:
Any idea why jenkins is continuously failing ?

hufnagel:
Ok, figured it out myself, will fix the code.

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
@ticoann @amaltaro how about now ? I think I addressed all the brought up points.

hufnagel:
Ah, still found one typo. Need to wait for jenkins again...

hufnagel:
Ok, went through the code again, looks good to me now. @amaltaro @ticoann please review and let me know if all your issues are addressed now.

hufnagel:
This has become more urgent now. I am more or less done refactoring the PhEDExInjector into a single piece of code with a single working thread, but I can't finalize or test until the code in this PR has been merged (too many conflicts to resolve first).

ticoann:
looks good too me. Still need to run the test in integration test.

amaltaro:
@ticoann we shouldn't have merged this PR. It's validation phase and it touches base code for central services.
I had no time to test it before, but for such cases we better test a workflow or two in our VMs before merging.

ticoann:
Sorry, my mistake. I thought this only affects agent side but apparently not. Let me run the few test with this patch.

amaltaro:
don't worry Seangchan, I can do this test. Just rest and get better ;)

hufnagel:
Ok, keep me in the loop on how the tests go.

PR #6531, opened 2016-01-20
Consolidate all uses of cjson and simplejson to json in python 2.7

### LINKED ISSUE #5785: Standardize on JSON libraries

ericvaandering:
Fixes #5785


ericvaandering:
I'll add one more commit to take care of some pylint issues. But @ticoann and @vkuznet please comment. Next step will be to remove all traces of the JsonWrapper classes.

vkuznet:
Eric,
I agree on removing simplejson since json module replaces it, but why you want
get rid of cjson (and JsonWrapper)? The cjson performs significantly better
than json especially on deep nested docs and when code deals with lots of them.
It also has less memory footprint. But I didn't check recent python code,
so may be worth to write simple benchmark and load N FWJR and compare the
CPU/RAM usage between json and cjson.
V.

ericvaandering:
The reasoning is as follows. CJSON was last updated 18 months ago and does not support  python 3. It's difficult to find documentation for. All the recent benchmarking info I've seen suggests 2.7's JSON is about as fast as cjson (https://gist.github.com/lightcatcher/1136415, http://justinfx.com/2012/07/25/python-2-7-3-serializer-speed-comparisons/ ) So, why not move to the official version now that things have settled down?

The wrappers, of course, only exist to give access to the alternate representations and to map the cjson interface to the more familiar dump(s)/load(s). If we only have one implementation, a wrapper is unnecessary code and another layer of indirection.

If you are still concerned, you can send me some documents (FWJRs you are working with I presume) and I can benchmark it. Or you can do the same.

vkuznet:
Well, let's take simple FWJR document
https://github.com/dmwm/WMArchive/blob/master/test/data/fwjr_processing.json

and pass it through benchmark code
https://gist.github.com/lightcatcher/1136415

will get

```
JSON Benchmark
2.7.11 (default, Dec  5 2015, 23:52:42)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]
-----------------------------
ENCODING
cjson: 2.697387s
json: 2.852638s

DECODING
cjson: 1.526323s
json: 1.112545s
```

So far so good, but what if someone gave you 100 of those FWJR in a single list, e.g.
web server sends a stream ```[doc1, doc2, doc3, ...]```. To imitate this use case
I generated docs.json with 100 jsons in a list:

```
>>> import json
>>> rec = json.load(open('fwjr.json'))
>>> ostream = open('docs.json', 'w')
>>> docs = [dict(rec) for _ in range(100)]
>>> ostream.write(json.dumps(docs))
>>> ostream.close()
```

If I pass this docs.json now to benchmark I get quite different results:

```
JSON Benchmark
2.7.11 (default, Dec  5 2015, 23:52:42)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]
-----------------------------
ENCODING
cjson: 150.986326s
json: 354.160647s

DECODING
cjson: 201.610148s
json: 159.415197s
```

The bottom line is that benchmark numbers really depends on use case.

ericvaandering:
That was exactly the use case I was envisioning. But JSON is within a factor of two for a pathological case (and that code is doing the encode/decode 10K times). The claims of 100s of times faster for CJSON are no longer valid. And considering that it is going away anyhow, I’d be inclined to bite the bullet now.

ericvaandering:
If we find we have performance issues, a better approach will be to certify ultrajson and do 

import ultrajson as json

vkuznet:
Fine with me if we're doing this consciously.

ericvaandering:
Obviously this will have to be tested in the real world before we really accept it. Let's discuss briefly in 20 minutes.

ericvaandering:
Ok, this is done. But as we discussed we should test this for real a) looking for performance issues and b) looking for places where we took advantage of a bug in cjson that did not respect unicode

ericvaandering:
Rebased onto recent master

ericvaandering:
@ticoann can we not forget about this please? I think you had some production tests in mind?

ericvaandering:
@ticoann @amaltaro

amaltaro:
Perhaps we should get rid of the duplicate module:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Wrappers/JsonWrapper/__init__.py
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Utils/jsonwrapper/__init__.py
?

amaltaro:
and since Savannah is dead, how about we completely remove this test:
 test/data/ReqMgr/SavannahRequestQuery.py
?

ericvaandering:
I had planned to get rid of both of those JsonWrapper modules as the next phase of the cleanup. No need to keep wrapping a standard library. But first I want to see everything running with the standard JSON. For the Savannah thing, yeah. :-)

amaltaro:
@ericvaandering should we merge it in the coming days and target it for the next cmsweb cycle?

ericvaandering:
Please, @amaltaro

amaltaro:
Thanks, Eric. Tests look good in both VM services and in the agent.

PR #6535, opened 2016-01-22
initial attempt to change fwjr structure (Do not merge!)

ticoann:
This is initial attempt to switch FWJR format to support archiver (putting in hadoop)
This code won't change the current structure unless newFormat is set to True.
1. Removing arbitrary key value from dictionary. 
   output = {name: value}  -> output = [{'outputModule': name, 'value': value}
   runs = {runNumber: runNumberList} -> runs = [{'runNumber': runNumber, 'value': runNumberList} 
2. for input, left as it is since inputSource is alwas 'source'

It would break the code when newFormat is set to True. - need to be fixed and updated.
Make sure this change wouldn't break any code outside WMCore. Especially CRAB - need to check.

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -156,7 +156,7 @@ class Report:
             raise FwkJobReportException(msg)
 
 
-    def jsonizeFiles(self, reportModule):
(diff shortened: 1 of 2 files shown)


ticoann:
closing it, we decided not to change fwjr format.

PR #6538, opened 2016-01-25
refactor phedexinjector, fixes #6451

### LINKED ISSUE #6451: we should switch the PhEDExInjector to use a single worker thread instead of two
At the moment the PhEDExInjector runs two worker threads. One deals with file injections and block closing, the other deals with subscriptions and block deletions. As far as I can tell, we don't really need two threads here, they would be able to run sequentially.

These worker threads run on a different polling cycle, but this could be implemented within one cycle by just doing part of the work every Nth cycle.

Reason I am pushing for this now is that we have big problems in the Tier0 with garbage response to all kind of PhEDEx data service calls and we can't seem to figure out where they come from. And these problem only appear when the PhEDExInjector is running two worker threads, not when it runs just one of them (either of them).

There is another advantage to switching to a single thread. With two threads sometimes log output from them are overlapping which makes debugging from the logs difficult. Also, if one of the threads dies, the other keeps writing logs, which makes finding the relevant failure messages in the logs difficult.

hufnagel:
Move all PhEDExInjector functionality into a single thread. Also remove all transactions and cleanup/simplify code (two separate threads used different conventions before, unify them).

Still being tested, but @amaltaro maybe take a look at the "big picture" changes already.


hufnagel:
This can only be tested against master, not the latest tag. At the very least, the recently merged subscription code changes are needed.

ericvaandering:
Testing happens against the latest thing tagged JENKINS_xxxx This tracks master with up to a 12 hour lag.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -36,17 +63,35 @@ class PhEDExInjectorPoller(BaseWorkerThread):
...
 
         Initialise class members
         """
         BaseWorkerThread.__init__(self)
-        self.config = config
-        self.phedex = phedex
         self.dbsUrl = config.DBSInterface.globalDBSUrl
 
+        self.pollCounter = 0
+        self.subFrequency = None
```
ericvaandering: Why None for a default here where a # is clearly what is needed later?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -36,17 +63,35 @@ class PhEDExInjectorPoller(BaseWorkerThread):
...
+        if getattr(config.PhEDExInjector, "subscribeDatasets", False):
+            pollInterval = config.PhEDExInjector.pollInterval
+            subInterval = config.PhEDExInjector.subscribeInterval
+            self.subFrequency = max(1, int(round(subInterval/pollInterval)))
+            logging.info("SubscribeDataset and deleteBlocks will run every %d polling cycles", self.subFrequency)
+
+        # retrieving the node mappings is fickle and can fail quite often
+        self.phedex = PhEDEx({"endpoint": config.PhEDExInjector.phedexurl}, "json")
+        try:
```
ericvaandering: We really should get the retry package as part of our distribution to get rid of silly patterns like this. @ticoann ?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -60,6 +105,11 @@ def __init__(self, config, phedex, nodeMappings):
             self.seMap[node["kind"]][node["se"]] = node["name"]
             self.nodeNames.append(node["name"])
 
+        self.phedexNodes = {'MSS':[], 'Disk':[]}
+        for node in nodeMappings["phedex"]["node"]:
+            if node["kind"] in [ "MSS", "Disk" ]:
```
ericvaandering: pylint would claim there is extraneous space here. :-)

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -92,6 +148,33 @@ def setup(self, parameters):
...
+        _algorithm_
+
+        Poll the database for uninjected files and attempt to inject them into
+        PhEDEx.
+        """
+        logging.info("Running PhEDEx injector poller algorithm...")
+
+        self.pollCounter += 1
+
+        if len(self.blocksToRecover) > 0:
```
ericvaandering: Is len() and a comparison necessary here or would ```if self.blocksToRecover:``` suffice?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -241,42 +323,35 @@ def injectFilesPhEDExCall(self, location, injectData, lfnList):
...
-            raise PhEDExInjectorPassableError(msg)
-
-        logging.info("Injection result: %s", injectRes)
-
-        if "error" in injectRes:
-            msg = ("Error injecting data %s: %s" %
-                   (injectData, injectRes["error"]))
-            logging.error(msg)
-            self.sendAlert(6, msg = msg)
+            if isinstance(ex, HTTPException):
```
ericvaandering: I don' understand why you got rid of the try/except/except pattern here. Seems more clear to me than testing the type of the exception.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -316,48 +387,30 @@ def closeBlocks(self):
...
-            except HTTPException as ex:
-                # If we get an HTTPException of certain types, raise it as an error
-                if ex.status == 400:
-                    msg =  "Received 400 HTTP Error From PhEDEx: %s" % str(ex.result)
-                    logging.error(msg)
-                    self.sendAlert(6, msg = msg)
-                    logging.debug("Blocks: %s", migratedBlocks[siteName])
-                    logging.debug("XMLData: %s", xmlData)
-                    raise
+            except Exception as ex:
```
ericvaandering: Same comment here about exception pattern

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -373,50 +426,221 @@ def recoverInjectedFiles(self):
...
+          If yes => request block deletion, approve request, set deleted=1
+          If no => do nothing (check again next cycle)
+        """
+        logging.info("Starting deleteBlocks method")
 
-        Poll the database for uninjected files and attempt to inject them into
-        PhEDEx.
+        blockDict = self.findDeletableBlocks.execute(transaction = False)
+
+        if len(blockDict) == 0:
```
ericvaandering: does ```if blockDict``` work?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -373,50 +426,221 @@ def recoverInjectedFiles(self):
...
+                            if dataset not in deletableEntries[location]:
+                                deletableEntries[location][dataset] = set()
+                            deletableEntries[location][dataset].add(blockName)
+
+
+        binds = []
+        for blockName in skippableBlocks:
+            binds.append( { 'DELETED' : 2,
+                            'BLOCKNAME' : blockName } )
+        if len(binds) > 0:
```
ericvaandering: ```if binds```

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -373,50 +426,221 @@ def recoverInjectedFiles(self):
...
-        except Exception:
-            # Guess we should roll back if we actually have an exception
-            if getattr(myThread, 'transaction', None):
-                myThread.transaction.rollbackForError()
-            raise
+            response = self.phedex.delete(deletion, xmlData)
+            requestId = response['phedex']['request_created'][0]['id']
+            # auto-approve deletion request
+            self.phedex.updateRequest(requestId, 'approve', location)
+        except Exception as ex:
```
ericvaandering: Same comment

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -373,50 +426,221 @@ def recoverInjectedFiles(self):
...
+            logging.info("Subscribing: %s to %s, with options: Move: %s, Custodial: %s, Request Only: %s",
+                         subscription.getDatasetPaths(),
+                         subscription.getNodes(),
+                         subscription.move,
+                         subscription.custodial,
+                         subscription.request_only)
+
+            try:
+                self.phedex.subscribe(subscription, xmlData)
+            except Exception as ex:
```
ericvaandering: And again

ericvaandering:
This means you added a test that fails the compile step. Often a missing import or perhaps a syntax error.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
hufnagel: Because the default is to skip the second polling cycle. Using None to indicate that avoids having to use a second parameter.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
hufnagel: Originally it got rid of some code duplication in the except clauses, but when I further cleaned this up now it really doesn't anymore. Will switch back to the double except again.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
hufnagel: Using a None as a boolean is quite 'pythonic" :-)

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
hufnagel: Ok, ok, I am doing it the 'python way'. Either works...

hufnagel:
I didn't touch any of the unit test code in this PR...

hufnagel:
Ok, jenkins is happy now, all comments are dealt with, this is done. As it relies on the subscription changes merged earlier I would wait until the other patch is tested and validate though.

@amaltaro @ticoann this should be reviewed under functional aspects too, Erics comments were more on coding style.

amaltaro:
Ok, Seangchan is validating the other patch (already merged).

I'll review and test this patch once I'm done with all the testbed stuff that is in my plate, possible only on Monday.

hufnagel:
Any news on this ?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -36,17 +63,35 @@ class PhEDExInjectorPoller(BaseWorkerThread):
...
-        self.config = config
-        self.phedex = phedex
         self.dbsUrl = config.DBSInterface.globalDBSUrl
 
+        self.pollCounter = 0
+        self.subFrequency = None
+        if getattr(config.PhEDExInjector, "subscribeDatasets", False):
+            pollInterval = config.PhEDExInjector.pollInterval
+            subInterval = config.PhEDExInjector.subscribeInterval
+            self.subFrequency = max(1, int(round(subInterval/pollInterval)))
```
amaltaro: I didn't get why we need another frequency/polling yet. Why don't we simply use the configuration subscribeInterval for that cycle of subscribe+delete?

amaltaro: Nevermind, I understand why it's needed now.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -92,6 +148,33 @@ def setup(self, parameters):
...
+
+        Poll the database for uninjected files and attempt to inject them into
+        PhEDEx.
+        """
+        logging.info("Running PhEDEx injector poller algorithm...")
+
+        self.pollCounter += 1
+
+        if self.blocksToRecover:
+            logging.info("""PhEDExInjector Recovery:
```
amaltaro: We better make it a warning logging and rephrase the message, "we" actually don't need to check it, but the agent will.

amaltaro:
There is one logic that I'd like to get rid of here, the list of diskSites (that was used when we were comissioning Disk/Tape separation:
https://github.com/hufnagel/WMCore/blob/phedexinjector-refactor/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L259

Is T0 injecting files against Buffer or MSS endpoints? If not, then we could simplify that logic and get rid of of the configuration for diskSites.

amaltaro:
I couldn't spot any issues in this PR, though it's not easy to review it. 
I'm patching an agent and I'll run a battery of tests.

BTW, we should get extra spaces removed before merging it, there are quite some of them that pylint would complain about.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
hufnagel: I don't want to touch this PR if not absolutely needed. As that log message is exactly the same as before, can we just create another issue later ?

hufnagel:
Same for the diskSites, I would prefer to handle this in a separate PR (actually was looking at this already before and decided against removing it). In any case, this feature is useless now that we keep location as PNN, once we remove all the old SE based code it'll go too.

Where do you see extra spaces ? Wouldn't the pylint in jenkins have complained about them ?

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -373,50 +424,224 @@ def recoverInjectedFiles(self):
...
+            logging.error("PhEDEx block delete/approval failed with HTTPException: %s %s", ex.status, ex.result)
+        except Exception as ex:
+            logging.error("PhEDEx block delete/approval failed with Exception: %s", str(ex))
+            logging.debug("Traceback: %s", str(traceback.format_exc()))
+        else:
+            binds = []
+            for dataset in blocksToDelete:
+                for blockName in blocksToDelete[dataset]:
+                    binds.append( { 'DELETED' : 1,
+                                    'BLOCKNAME' : blockName } )
```
amaltaro: These spaces, inside the dict and inside the parenthesis. Also in other places like this one

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
amaltaro: I see. Since you are touching this code, it would be better to make these cosmetic fixes inside this PR. Of course, I see no reason why not to make a new one for those changes as well.

amaltaro:
> Wouldn't the pylint in jenkins have complained about them ?

I don't think Jenkins complain on Convention cases. Only Errors and Warnings (and I guess Refactoring).
You need to manually run pylint to spot those coding styles/conventions.

hufnagel:
Changed it.

ericvaandering:
It complains. That's the XX comments on code style. I assume if you are going to fix pylint stuff you're going to run pylint yourself to see what the whole list of complaints is, but I don't include the "C" stuff in the mail/web post because it's too long.

amaltaro:
Agreed, Eric. The list would be super mega giant for some cases :)

Dirk, Seangchan, I ran all the DMWM tests over the last night, checked all of them with a script that checks couch, dbs and phedex and everything looks Ok. Component logs also look fine to me.

The only improvement I'd still like to have is regarding the subscribeDatasets call. It would be helpful to run that in the first cycles of the component, instead of waiting for pollCounter to catch up on subFrequency. Thus, we could - when config subscribeDatasets is True - initialize self.pollCounter to something like self.subFrequency -1. If you have question, just pass by my office.

hufnagel:
Adjusted to call subscribeDatasets in the first polling cycle.

amaltaro:
Thanks Dirk! The only change was on L82-83, right? If so, then these changes look Ok.

amaltaro:
@hufnagel can you confirm the changes please?
@ticoann once Dirk confirm it, I think we should merge it before @ericvaandering 's avalanche of print updates :-D

ericvaandering:
Definitely. And breaking those up to try to minimize conflicts.

On Feb 10, 2016, at 15:20, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


@hufnagel<https://github.com/hufnagel> can you confirm the changes please?
@ticoann<https://github.com/ticoann> once Dirk confirm it, I think we should merge it before @ericvaandering<https://github.com/ericvaandering> 's avalanche of print updates :-D

-

hufnagel:
Yeah, only changed a couple lines (and the order of a comparison statement later, which doesn't matter).

PR #6542, opened 2016-01-27
Track blow-up factor in TaskChain requests

amaltaro:
Overseed #6464. Minor fixes for pylint, pre 2.6 idioms and upper case. Mainly fixed the siteJobCounts calculation that had an extra +1.

Still to decide whether we merge it now for this week's testbed or not.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
@@ -84,6 +84,7 @@ Example initial processing task
      "SplittingArguments"    : {"files_per_job" : 1 }, Size of jobs in terms of splitting algorithm
  },
 """
+from __future__ import division
 
 from WMCore.Lexicon import identifier, couchurl, block, primdataset, dataset
 from WMCore.WMSpec.StdSpecs.StdBase import StdBase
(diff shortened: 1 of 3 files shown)


amaltaro:
testPrioritiesWorkPolling test is unstable waiting for the DBS mock. I'll make another PR to mark that as unstable for now.

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -471,13 +489,13 @@ def modifyTaskConfiguration(self, taskConf,
...
-            taskConf["RequestNumEvents"] = int(taskConf.get("RequestNumEvents", 0) / \
-                                               taskConf.get("FilterEfficiency"))
+            taskConf["RequestNumEvents"] = taskConf.get("RequestNumEvents", 0) // \
+                                               taskConf.get("FilterEfficiency")
             taskConf["SizePerEvent"] = taskConf.get("SizePerEvent", self.sizePerEvent) * \
                                        taskConf.get("FilterEfficiency")
 
         if taskConf["EventsPerJob"] is None:
-            taskConf["EventsPerJob"] = int((8.0 * 3600.0) / (taskConf.get("TimePerEvent", self.timePerEvent)))
+            taskConf["EventsPerJob"] = (8.0 * 3600.0) // taskConf.get("TimePerEvent", self.timePerEvent)
```
ericvaandering: Are you sure these are equivalent? regardless I kind of prefer the outer int because it's clearer (to me)

amaltaro: It returns the floor division in a float format, like 8.5 // 4 = 2.0. 
int() guarantees to return a INT floor division, like int(8.5/4.0) = 2
Should I revert the casting then?

amaltaro:
I read comparison of different object types will raise an exception in py3, so I reverted those 2 lines back (from // to int()), to avoid possible headaches in the future.

src/python/WMCore/WMSpec/StdSpecs/TaskChain.py
```diff
@@ -477,7 +495,7 @@ def modifyTaskConfiguration(self, taskConf,
                                        taskConf.get("FilterEfficiency")
 
         if taskConf["EventsPerJob"] is None:
-            taskConf["EventsPerJob"] = int((8.0 * 3600.0) / (taskConf.get("TimePerEvent", self.timePerEvent)))
+            taskConf["EventsPerJob"] = int((8.0 * 3600.0) / taskConf.get("TimePerEvent", self.timePerEvent))
```
ericvaandering: Don't change this, but of course there is no reason to use 8.0 vs 8, etc. That was done to force the old code into being float division.

amaltaro:
@ticoann good to be merged.

PR #6549, opened 2016-01-28
Don.t call listFileArray with run=1 - 1.0.12 version

amaltaro:
1.0.12 version of #6512

Still need to be tested. It needs to be applied not after the cmsweb upgrade.


amaltaro:
Good to go, I'm applying this patch to all the production agents by tomorrow.

PR #6094, opened 2015-08-04
Make unique file entries for ACDC MCFakeFiles

amaltaro:
It seems several acdc entries (jobs) using the same MCFakeFile causes an Oracle IntegrityError

```
2015-08-03 07:25:55,051:ERROR:WorkQueue:jbadillo_ACDC_SUS-RunIIWinter15pLHE-00011_00010_v0_STEP0ATCERN_150708_150727_7499, /acdc/jbadillo_ACDC_SUS-RunIIWinter15pLHE-00011_00010_v0_STEP0ATCERN_150708_150727_7499/:pdmvserv_SUS-RunIIWinter15pLHE-00011_00010_v0_STEP0ATCERN_150704_201230_7297:Production/0/26 creating subscription failed in LQ: (IntegrityError) (1062, "Duplicate entry 'MCFakeFile-pdmvserv_SUS-RunIIWinter15pLHE-00011_00010_v0_STEP0AT' for key 'lfn'") 'insert into dbsbuffer_file(lfn, filesize, events, dataset_algo, status, workflow)\n                values (%s, %s, %s, %s, %s, %s)' [('MCFakeFile-pdmvserv_SUS-RunIIWinter15pLHE-00011_00010_v0_STEP0ATCERN_150704_201230_7297-Production-b9ea0183968f67ba39db2a964f0565e6', 0, 500000, 1L, 'GLOBAL', 19221L), ('MCFakeFile-pdmvserv_SUS-RunIIWinter15pLHE-00011_00010_v0_STEP0ATCERN_150704_201230_7297-Production-b9ea0183968f67ba39db2a964f0565e6', 0, 200000, 1L, 'GLOBAL', 19221L)]
```

cmssrv217 was raising this exception for a couple of weeks.
I have applied this patch there (agent is basically drained) and then WorkQueue went through.

@ticoann please do not merge it yet, I need to go deeper here to make sure it's not going to bug down in the chain (I guess job creation for acdc is retrieved from couch, not from oracle, so in principle we should be safe here)...

PR #6551, opened 2016-01-30
1.0.12 version of Make unique file entries for ACDC MCFakeFiles

amaltaro:
Exactly the same patch as #6094.

I'll push it soon to the agents.


ticoann:
Thanks Alan, new tag is created with this and #6549. 1.0.12.patch3

PR #6557, opened 2016-02-03
Update PhEDEx groups during assignment

amaltaro:
We missed this piece of code in Dirk's PR #6529 already merged.

I was testing it this week and noticed it always defaults to DataOps group, so won't make any harm to central production.

I still need to test it, don't merge please.

@hufnagel I got two questions though:
1. IMO, we should not be exposing DeleteFromSource to anyone. In the end it's coupled to Move subscriptions, isn't it? If so, we forget about this parameter and simply add this param when making Move sub.
2. what's the reason behing having PrimaryDataset for PhEDEx subscriptions? Is it used in T0?

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/NewSubscription.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/NewSubscription.py
@@ -10,6 +10,7 @@ Created on May 2, 2013
 
 from WMCore.Database.DBFormatter import DBFormatter
 
+
 class NewSubscription(DBFormatter):
     """
     _NewSubscription_
@@ -23,7 +24,7 @@ class NewSubscription(DBFormatter):
           """
 
     def _createPhEDExSubBinds(self, datasetID, subscriptionInfo, custodialFlag):
-        
+
         # DeleteFromSource is not supported for move subscriptions
         delete_blocks = None
         if custodialFlag:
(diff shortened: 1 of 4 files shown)


hufnagel:
DeleteFromSource is not for Move subscriptions, it's for Copy subscriptions. Default is False and then it does nothing. If it is set to True, it'll block delete the produced data from the injection site when these conditions are set:

- data was subscribed to another site or sites and the block has been fully transferred
- all workflows producing files in the block have finished
- there is no active subscription to the injection site

It's basically a delayed move to account for TaskChain workflows and their need to keep the data at the injection site for a while. Tier0 has been using it exclusively for month.

You don't have to use it for WMAgent Ops (as I said, default is False), but the option is there, there is nothing Tier0 specific in the implementation.

hufnagel:
PrimaryDataset is used in the Tier0 for Repack workflows. They have multiple outputs with different primary datasets and the same data tier but completely different subscription.

hufnagel:
PS: Feel free to change DeleteFromSource from a request parameter to an assignment parameter. It should be configurable with the other subscription parameters though because it is there and usable.

amaltaro:
> data was subscribed to another site or sites and the block has been fully transferred
> all workflows producing files in the block have finished
> there is no active subscription to the injection site 

I'm not sure I understood it completely. In the end, is it a replica subscription or a deletion request?
Does the agent watch over subscriptions to see when it's fully transferred?

hufnagel:
It's a replica subscription, the agent watches things and then makes block deletion requests. This is only usable if the agent uses a cert that can approve block deletion requests though, which is the case for the Tier0, but at the moment not really for other agents/sites.

So this is usable in principle, but for operational reason not in practice now. I still would add the option as we are touching this code anyways.

amaltaro:
test this please

amaltaro:
Ok, everything is awesome now. Just a summary of changes in this PR:
1. Custodial and NonCustodialType defaults to Replica (done before, actually)
2. Custodial and NonCustodialGroup defaults to DataOps
3. DeleteFromSource defaults to False (not currently used by the WF team)
4. Pushed these new params and Dashboard,  UseSiteListAsLocation and AllowOpportunistic to the workload_cache couch doc.
5. All these params can be overwritten during assignment.

I'll make another commit with the pylint/pep-8 changes.

hufnagel:
About 4, what about TrustSiteLists ? Doesn't that also exist ?

amaltaro:
Thanks Dirk, just added TrustSitelists flag as well. I'm done with my changes and testing.
@hufnagel @ticoann please review

hufnagel:
Looks good to me.

amaltaro:
@ticoann just ran a final test with all the changes (including cosmetic ones) in place and we are good to go! Please merge

PR #6556, opened 2016-02-02
Throw an exception if there is an attempt to access DBSReader - never merge

ericvaandering:
without having EmulatedUnitTest imported

PR #6485, opened 2015-12-18
Cause old DBS emulator not to function - Don't merge ever!

ericvaandering:
This patch causes the old DBS2 emulator to not function. Failing unit tests are good candidates for @jha2 to reimplement using new Mocked DBS

PR #6558, opened 2016-02-03
Throw an exception if we try to access the old DBS3 emulator - never merge

ericvaandering:


ericvaandering:
Like #6485 but for the DBS3 version of the emulator. 
PR #6556 does something similar trying to trap actual DBS access

ericvaandering:
test this please

ericvaandering:
test this please

PR #6559, opened 2016-02-03
Workqueue resource monitor

ticoann:
using 
[gloabqueue instance].status("Available")

It can retrieve workqueue element information which has higher priority (or equal) within the same site group.

i.e.

```
For this workflow,
Sites for riahi_TEST_HELIX_0911-T1_UK_RALBackfill_151119_190209_6101: priority 50000, set(['T1_UK_RAL'])

There are few workflow ahead of this for given site.

riahi_TEST_HELIX_0911-T1_UK_RALBackfill_151209_220229_2632: priority: 90000 job: 10010 sites: 
set(['T1_UK_RAL'])
riahi_TEST_HELIX_0911-T1_UK_RALBackfill_151123_214806_3068: priority: 90000 job: 10010 sites: set(['T1_UK_RAL'])
```

Still need to be refined.
1. factor in creation time.
2.  Need to define what we should do when there are different blocks in the same workflow which has different site restriction.


ticoann:
@amaltaro, Alan, please review. I will add separate patch on diagnosis tool using this and local queue information.

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
```diff
@@ -237,4 +237,33 @@ def passesSiteRestriction(self, site):
...
+        """
+        if len(a) == 0 or len(b) == 0:
+            return a | b
+        else:
+            return a & b
 
+    def possibleSites(self):
+        
+        if self['NoLocationUpdate']:
+            return "all"
```
amaltaro: Shouldn't we return self['SiteWhitelist'] in this case?

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElementResult.py
```diff
@@ -23,6 +23,8 @@ def __init__(self, **kwargs):
                             sum([x['EventsWritten'] for x in self['Elements']]))
             self.setdefault('FilesProcessed',
                             sum([x['FilesProcessed'] for x in self['Elements']]))
+            self.setdefault('Jobs',
```
amaltaro: Don't we need to consider the blowupFactor here?
And a silly question, what's the difference in this if-else? Is the ParentQueueElement the workqueue_inbox doc that is created for each workflow, and which spawns > 1 WQE?

test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElementsSummary_t.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+"""
+    WorkQueueElement unit tests
+"""
+from __future__ import print_function
+import unittest
+from WMCore.Wrappers import JsonWrapper
+from WMCore.WorkQueue.DataStructs.WorkQueueElementsSummary import WorkQueueElementsSummary
```
amaltaro: Isn't it supposed to be WorkQueueElementResult class? Or did you forget to push this new class?

test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElementsSummary_t.py
```diff
@@ -0,0 +1,47 @@
...
+        self.gqElements = []
+        for ele in gqData:
+            self.gqElements.append(WorkQueueElement(**ele))
+
+
+    
+    def testElementsWithHigherPriorityInSameSites(self):
+        """Id calculated correctly"""
+        
+        gqSummary = WorkQueueElementsSummary(self.gqElements)
```
amaltaro: same comment here and below (same for methods called).

test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElementsSummary_t.py
ticoann: Thanks for got to add new class. It is added now.

ticoann:
test this please

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
ticoann: You are right. Thanks Alan,

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElementResult.py
ticoann: We can't add blowupFactor here. That will be double blowup. :smile: But for correct calculation of how many jobs will be created, you are right we have to add the same blowup factor right before the display. (If we add here it will be calculated again in the LQ WorkQueueManager.


> And a silly question, what's the difference in this if-else? Is the ParentQueueElement the workqueue_inbox doc that is created for each workflow, and which spawns > 1 WQE?

ParentQueueElement is a global workqueue element for given local workqueue elements.
So global workqueue elements doesn't have ParentQueueElement

vlimant:
How does one access that information and see what is inside the agents ?

PR #6565, opened 2016-02-05
delete WMStats document locally

ticoann:
Since production wmstats is update to clean up db using thread, this needs to be patched in production agents. (Sorry I forgot to add this earlier)
@amaltaro Alan, please review.

### DIFF
--- a/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
+++ b/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
@@ -213,8 +213,10 @@ class WorkQueueElement(dict):
     def passesSiteRestriction(self, site):
         """Takes account of white & black list, and input data to work out
         if site can run the work"""
-        # Don't check anything if useSiteListAsLocation/trusSiteLists is enabled
-        if self['NoLocationUpdate']:
+        # Don't check anything if trustSitelists is enabled and SiteWhitelist is empty or site is 
+        # is site whitelist
+        if self['NoLocationUpdate'] and \
+           (not self['SiteWhitelist'] or site in self['SiteWhitelist']):
             return True
         # data restrictions - all data must be present at site
         for locations in self['Inputs'].values():


src/python/WMCore/Services/WMStats/WMStatsReader.py
```diff
@@ -326,5 +335,21 @@ def getRequestSummaryWithJobInfo(self, requestName):
...
+        """
+        get list of archived workflow in wmstats db.
+        """
+        
+        options = {"group_level": 1, "reduce": True}
+        
+        results = self.couchDB.loadView(self.couchapp, "allWorkflows", options = options)['rows']
+        requestNames = [x['key'] for x in results]
+        
+        workflowDict = self.reqDB.getStatusAndTypeByRequest(requestNames)
```
amaltaro: Aren't we going to fetch way too many requests here? Maybe we could gather all the workflows newer than 3 months (is there a view by request creation date?), filter the archived ones and proceed with that small set only?

src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
```diff
@@ -0,0 +1,45 @@
...
+                              reqdbCouchApp=config.reqdb_couch_app)
+
+    def setConcurrentTasks(self, config):
+        """
+        sets the list of functions which
+        """
+        self.concurrentTasks = [{'func': self.cleanUpOldRequests, 'duration': (config.DataKeepDays * 24 * 60 * 60)},
+                                {'func': self.cleanUpArchivedRequests, 'duration': config.archivedCleanUpDuration}]
+
+    def cleanUpOldRequests(self, config):
```
amaltaro: Hmm, what are these old docs? I thought this was needed only on the agent side.

src/python/WMCore/Services/WMStats/WMStatsReader.py
ticoann: It shouldn't be that many, 
1. it use reduce so only workflows it will return the number of workflows not number of docs.
2. and once the document related workflow is deleted it is gone from db, so next polling cycle there will be less documents to retrieve.

i.e. initially there are 2000 archived workflows (maybe 2 million docs) and 1000 active workflows (with million doc).
after deleting the will be only 1000 active workflows but 2 million docs will be deleted.
Next cycle there might be few more workflows  are archived. (I run this twice a day, it shouldn't be creating too many archived workflows in have a day)
https://github.com/dmwm/deployment/blob/master/reqmon/config.py#L106

src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
ticoann: This patch includes WMStats in cmsweb as well which is already in production last week

ericvaandering:
"nose.failure.Failure:runTest was added. Status is error"

This, combined with all the deleted tests suggest there is an import missing somewhere that is causing a whole lot of failures.

ticoann:
closing this one. New agent will update this patch

PR #6567, opened 2016-02-08
Removing DBSUpload and DBSBuffer

### LINKED ISSUE #4949: CleanUp all the dbs2 and dls related code.
remove dbs3UploadOnly flag
remove the sql set by the flog
remove DBSUpload, DBSBuffer package.
need to be careful since some of code used outside those packages (i.e. PhEDExInjector)

alexanderrichards:
This PR fixes https://github.com/dmwm/WMCore/issues/4949


alexanderrichards:
retest this please

ericvaandering:
At a minimum I am worried about these two.

Eric

alexanderrichards:
So the first of these:

> WMComponent_t.DBS3Buffer_t.DBSBufferFile_t.DBSBufferFileTest:testProperties changed from success to error

comes from:
[WMComponent_t.DBS3Buffer_t.DBSBufferFile_t.py](https://github.com/dmwm/WMCore/blob/master/test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py#L972)

where it is clearly using dbs2 daoFactory2 defined:
https://github.com/dmwm/WMCore/blob/master/test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py#L39

So the problem here is that there is no dbs3 equivalent of  LoadInfoFromDAS.py in MySQL. Do we therefore want to keep this test if it was originally designed to test dbs2 functionality that we have now removed? or is there an equivalent dbs2 sql in a different filename?

ticoann:
test this please

bin/wmcoreD
```diff
@@ -8,12 +8,12 @@ and control the daemons:
...
 # generates a config file based on the root dirs (it looks for DefaultConfig.py files):
 # the component sections need a namespace attribute to the class can be loaded and daemonized.
-wmcore-new-config --roots=src/python/WMComponent/ErrorHandler/,src/python/WMComponent/DBSBuffer/ --output=WMCoreConfig.py
+wmcore-new-config --roots=src/python/WMComponent/ErrorHandler/,src/python/WMComponent/DBS3Buffer/ --output=WMCoreConfig.py
 
 #edit the generated config file to fit your needs.
 
 # loads default tables and tables from a specific module. The backend is automatically selected based on the config file.
-wmcore-db-init --config=WMCoreConfig.py --create --modules=WMComponent.DBSBuffer.Database 
+wmcore-db-init --config=WMCoreConfig.py --create --modules=WMComponent.DBS3Buffer.Database
```
ticoann: It is just comments but should be following

wmcore-db-init --config=WMCoreConfig.py --create --modules=WMComponent.DBS3Buffer

alexanderrichards: yes, thanks good spot

PR #6570, opened 2016-02-10
Escape single quote if execute in clean env

juztas:


juztas:
https://github.com/dmwm/CRABClient/pull/4594

juztas:
@mmascher @jmarra13 this should solve for users with older release. I tried with 5_3_19, 7_0_6, 7_6_0 and it works ok.

mmascher:
@juztas can you create the "1.0.12_crab version of" this pull request please (like https://github.com/dmwm/WMCore/pull/6478)

Then we will need a WMCore tag if we want to deploy this quickly (depending if the user still has problems)

juztas:
@mmascher https://github.com/dmwm/WMCore/pull/6571

src/python/WMCore/Credential/Proxy.py
```diff
@@ -159,6 +159,8 @@ def setEnv(self, cmd):
         if self.uisource is not None and len(self.uisource) > 0:
             ret = 'source ' + self.uisource + ' && ' + ret
         if self.cleanEnvironment:
+            # Need to escape ' in cmd
+            ret = ret.replace("'", "'\\''")
```
ticoann: isn't this following?
```python
ret = ret.replace("'", "\\'")
or 
ret = ret.replace("'", r"\'")
```

juztas:
updated in both pull requests to use r

src/python/WMCore/Credential/Proxy.py
```diff
@@ -159,6 +159,8 @@ def setEnv(self, cmd):
         if self.uisource is not None and len(self.uisource) > 0:
             ret = 'source ' + self.uisource + ' && ' + ret
         if self.cleanEnvironment:
+            # Need to escape ' in cmd
+            ret = ret.replace("'", r"'\''")
```
ticoann: Hi Justas, 

I meant whether there need additional sing quote?
``` python
ret = ret.replace("'", r"'\''") # this produces "Tom'\\''s house"
ret = ret.replace("'", r"\'") # this produces "Tom\\'s house"
```
I was just wondering the first one is what you want (I thought it should be the second case)
first one produces

PR #6571, opened 2016-02-10
Escape single quote if execute in clean env for crab release

juztas:


src/python/WMCore/Credential/Proxy.py
```diff
@@ -159,6 +159,8 @@ def setEnv(self, cmd):
         if self.uisource is not None and len(self.uisource) > 0:
             ret = 'source ' + self.uisource + ' && ' + ret
         if self.cleanEnvironment:
+            # Need to escape ' in cmd
+            ret = ret.replace("'", "'\\''")
```
ticoann: isn't this 
```python
ret = ret.replace("'", "\\'")
or 
ret = ret.replace("'", r"\'")
```

mmascher:
@juztas Can you edit the title and call this "1.0.12_crab version of " and then the original title? That's the protocol for pull requests over branches.

src/python/WMCore/Credential/Proxy.py
mmascher: The two instructions are equivalent, of course using `r'string'` is the standard way of telling python "do not escape the string", not escaping the escape character :)

Let's do the merge in this crab branch, and make sure the code is changed in the master pull.

juztas: This is correct, see : http://wiki.bash-hackers.org/syntax/quoting
`r''` will only make one backslash less
```
In practise that means, to produce a text like Here's my test… as a single-quoted string, you have to leave and re-enter the single quoting to get the character "'" as literal text:

# WRONG
echo 'Here's my test...'

# RIGHT
echo 'Here'\''s my test...'

# ALTERNATIVE: It's also possible to mix-and-match quotes for readability:
echo "Here's my test"
```

juztas: ok, Marco convinced me, 1 min will add r

src/python/WMCore/Credential/Proxy.py
```diff
@@ -159,6 +159,8 @@ def setEnv(self, cmd):
         if self.uisource is not None and len(self.uisource) > 0:
             ret = 'source ' + self.uisource + ' && ' + ret
         if self.cleanEnvironment:
+            # Need to escape ' in cmd
+            ret = ret.replace("'", r"'\''")
```
ticoann: Hi Justas, 

I meant whether there need additional sing quote?
``` python
ret = ret.replace("'", r"'\''") # this produces "Tom'\\''s house"
ret = ret.replace("'", r"\'") # this produces "Tom\\'s house"
```
I was just wondering the first one is what you want (I thought it should be the second case)
first one produces

juztas: This is what I want, see below
```
In practise that means, to produce a text like Here's my test… as a single-quoted string, you have to leave and re-enter the single quoting to get the character "'" as literal text:

# WRONG
echo 'Here's my test...'

# RIGHT
echo 'Here'\''s my test...'

# ALTERNATIVE: It's also possible to mix-and-match quotes for readability:
echo "Here's my test"
```

PR #6585, opened 2016-02-12
Exclude LogCollect and CleanUp task for completed state transition

ticoann:
This is the first part of the patch. 
1. It will make the complete status without checking LogCollect and CleanUp tasks
2. workload summary will be only created all the subscription is finished 
3. request will be moved to archived status only after LogCollect and CleanUp tasks are finished.

I will create the second patch separately which will do following.
1. update LogCollect and CleanUp task status to the  central couchdb (reqmgr2 db) by agent
2. update db clean status  to the  central couchdb (reqmgr2 db) by agent (for moving to archived status)
3. create the cherrypy thread which collects all the status information from agent and update reqmgr2 db request status. (including archived status)


ticoann:
@amaltaro, Alan please review. We can discuss in detail tomorrow. Thanks,

src/python/WMComponent/TaskArchiver/DataCache.py
```diff
@@ -0,0 +1,12 @@
...
+class DataCache(object): 
+    _finishedWFs = [] # Global cache holding workflow lists 
+    
+    @staticmethod
+    def getFinishedWorkflows():
+        return DataCache._finishedWFs;
+
+    @staticmethod
+    def setFinishedWorkflows(workflowLists):
+        DataCache._finishedWFs = workflowLists;
```
amaltaro: this will overwrite the cache. How about we use a set() for the class attr and then we update() here?
Then, in the get method it could cast to a list if needed.

ticoann: cache need to be overwritten each polling cycle.
https://github.com/dmwm/WMCore/pull/6585/files#diff-748e5dcfaf8a48d0e9774b3066780e42R298

I am passing set for the value, sorry it my mistake set the default value as list. I will fix that.

ticoann:
Refactor the code  move all archive and clean up code to CleanCouchPoller.  Still could have race condition if the couchdb clean up failed for some reason and retried. Had to rewrite the sql. 
@amaltaro, Alan, please review. I will fix the unittest and some of pylint warnings.

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -252,3 +450,686 @@ def cleanAlreadyArchivedWorkflows(self):
...
+                        subscription.load()
+                        subscription.deleteEverything()
+
+                    #Check that the workflow is gone
+                    if wmbsWorkflow.exists():
+                        #Something went bad, this workflow
+                        #should be gone by now
+                        msg = "Workflow %s, Task %s was not deleted completely" % (wmbsWorkflow.name,
+                                                                                   wmbsWorkflow.task)
+                        raise TaskArchiverPollerException(msg)
```
amaltaro: This exception does not belong here :-) Same a few lines below

amaltaro:
Seangchan, I noticed the workloadSummary will only be updated when all tasks are done (including logCollect and cleanUp), and the workflow will stay in the same completed status, right? So, this one will be tricky to identify (besides checking...) when the workloadSummary is available or not. 

About TaskArchiver thread, it will basically a) mark subscriptions as Done; b) populate the DataCache with 100% completed workflows; c) change workflows to a completed status and d) mark WQE as Done

The CleanCouch thread will a) change workflow status to a final archived status; b) clean up local couch (central couch cleans by itself?) c) clean up wmbs and dbsbuffer tables; d) upload workload summary and e) calculate and upload a bunch of performance data.

It seems to me there are a lot of stuff we could remove from this code (maybe even some of those dashboard reports/performance). It's worth revisit it in the near future :-)

I didn't spot any issues, though it's easy to review it. Let's put it in an agent and fullly validate in the coming weeks.

PR #6586, opened 2016-02-12
change the default log level

ticoann:
I am not sure what is the causing this, but it seems some of the log (wmstats) has default level WARNING, instead of NOSET. 
According to the documentation it said "root logger"'s default value is WARNING but not sure how it affects the wmstats logging.
https://docs.python.org/2/library/logging.html

The problem was first self.logger.info prints out the log but not the second one.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py#L41

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py#L30


ticoann:
@ericvaandering, @amaltaro  I am not sure why this happens (it is not overwriting logging level anywhere I could find). Please review

Issue #6345, opened 2015-10-21
Stepchain: support trident configurations


amaltaro:
Stepchain should support parameters like input step and input module such that we can chain the steps as we want. For instance, 
 - Step1 produces GEN-SIM-RECO
 - Step2 reads the GEN-SIM-RECO from Step1 and produces AODSIM
 - Step3 *also* reads the GEN-SIM-RECO from Step1 and produces AODSIM
Differences between Step2 and Step3 could be a != release, or GT or config or etc

amaltaro:
Setting milestone to January, feel free to change it in case I'm wrong.

ticoann:
This will allow each intermediate output to be stored, correct?

ericvaandering:
That's a different, probably simpler, issue.

amaltaro:
Fixed by #6589

PR #6589, opened 2016-02-17
Allow StepChain to output intermediate samples

### LINKED ISSUE #6344: Stepchain: allow intermediate stage out
As discussed in yesterday's meeting, we need to have the flexibility of saving intermediate outputs in Stepchain. Right now it only saves (stages out) the files produced by the last step (cmsRun).

amaltaro:
Fixes #6344 and #6345

A few notes about this fix:
1. **a new patch still needs to be created for the agent**, more in [*]
2. Breaks the current behaviour of StepChain. Added 'KeepOutput' parameter to each task (which defaults to True), such that it MUST be set to False in case we don't want outputs of a/some step(s)
3. The last step in the schema is supposed to be the last step, so it MUST have KeepOutput=True, otherwise validation will fail.
4. It does not support two generation steps (two initial independent steps). Can be added in the future if needed.

@ticoann sorry for the delay, it took me a long time to find the best way to fix it - and stop making stupid mistakes - and still not in a completely working state ... :(
Let me run a one more test in my VM.

[*] Merge jobs for the same output module still fall into the same fileset, in the agent. E.g.: in our MC template, we have RAWSIMoutput for Step1 (GEN-SIM tier) and for Step2 (GEN-SIM-RAW tier). The agent creates different wmbs subscriptions for them, but it associates them to the same wmbs fileset

```
| wf_id | sub_id | fset_id | wf_task                                                                                                                                             | fset_name                                                                                                                      |
|   410 |    414 |     923 | /amaltaro_StepChain_MC_DMWM_TEST_ALAN_NoDrop_v19_160217_084415_4570/ProdMinBias/DIGIPROD1MergeRAWSIMoutput                                          | /amaltaro_StepChain_MC_DMWM_TEST_ALAN_NoDrop_v19_160217_084415_4570/ProdMinBias/unmerged-RAWSIMoutput                          |
|   407 |    411 |     923 | /amaltaro_StepChain_MC_DMWM_TEST_ALAN_NoDrop_v19_160217_084415_4570/ProdMinBias/ProdMinBiasMergeRAWSIMoutput                                        | /amaltaro_StepChain_MC_DMWM_TEST_ALAN_NoDrop_v19_160217_084415_4570/ProdMinBias/unmerged-RAWSIMoutput                          |
```

which means, the merge job will try to merge both GEN-SIM and GEN-SIM-RAW files, into the same job...

I have no idea how to fix it in the agent, will need to find out by the cmsweb production upgrade :-D


amaltaro:
test this please

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
```diff
@@ -421,16 +438,16 @@ def validateSchema(self, schema):
         Go over each step and make sure it matches validation parameters.
         """
         numSteps = schema['StepChain']
-        couchUrl = schema.get("ConfigCacheUrl", None) or schema["CouchURL"]
+        couchUrl = schema.get("ConfigCacheUrl", None) or schema.get("CouchURL", None)
```
ericvaandering: What happens after this if couchUrl is None?

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
```diff
@@ -445,6 +462,11 @@ def validateSchema(self, schema):
                                                couchDBName=schema["CouchDBName"],
                                                getOutputModules=True)
 
+        if 'KeepOutput' in schema[stepName] and strToBool(schema[stepName]['KeepOutput']) is False:
```
ericvaandering: '''
and not strToBool
'''

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
```diff
@@ -445,6 +462,11 @@ def validateSchema(self, schema):
...
+            msg = "Dropping the output of the last step is prohibited.\n"
+            msg += "Set the 'KeepOutput' value to True and try again."
+            self.raiseValidationException(msg=msg)
```
ericvaandering: Did not look at the base class, but do we really need a function to raise an exception?

test/python/WMCore_t/WMSpec_t/StdSpecs_t/StepChain_t.py
```diff
@@ -7,11 +7,13 @@
 import json
 import os
 import unittest
-
+from copy import copy
```
ericvaandering: Just verify that you don't need deep copy instead of copy.

amaltaro: Yep, we only copy simple data structure/objects here. So, copy should be fine.

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
amaltaro: Actually, couchUrl cannot be None, since it's a mandatory parameter:
https://github.com/amaltaro/WMCore/blob/stepchain-intermediate/src/python/WMCore/WMSpec/StdSpecs/StepChain.py#L331

so in the end, the change in this line was useless.

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
amaltaro: I knew there was something weird when I wrote it!

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
amaltaro: This is the code of the method:
`logging.error("About to raise exception %s" % msg)`
`raise WMSpecFactoryException(message=msg)`

Should we get rid of of this method and simply raise the WMSpec exception? If so, then I can make it in a new PR, sounds cleaner.

amaltaro:
Just pushed another commit to fix the unittest. IF it passes the unittests, feel free to merge it.

amaltaro:
@ticoann please merge

ticoann:
Thanks a lot, Alan

PR #6593, opened 2016-02-17
remove unused code

ticoann:


amaltaro:
I like these pull requests!!! :-)

ericvaandering:
Me too. What is the HTTPFrontEnd?

ticoann:
> Me too. What is the HTTPFrontEnd?

It is the component version of WMCore/HTTPFrontEnd, but never used in anywhere. 
WMCore/HTTPFrontEnd need to be deleted but still several services running with it. (ReqMgr, WMBS, Agent) The last 2 are the REST services which can provide component and wmbs information but I don't think any one is used due to they are under firewall. It could be useful for monitoring though. )

PR #6597, opened 2016-02-18
Take nCores into account for automatic event job splitting

### LINKED ISSUE #6448: Automatic job splitting must take multicore into account
The default job splitting aims at 8h jobs, but when multicore is enabled that would give us 8h/nCores.

amaltaro:
Fixes #6448 (?)
This issue popped up just before Xmas break, which was discussed here to some extent:
https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/2704/2/2/1/2/1/1/1/1/1/2/1/1/1/1/2.html

The main issue was the definition of TimePerEvent, which apparently McM considers to be something like: (Time it took to process them) / (N events processed). E.g. it took 60secs to process 10 evts (for 4 cores), that would give us an average of 6s per event.

As Brian mentioned, TimePerEvent should be the time each event takes to process, but we do N events at a time. E.g. 6s per event would give us: 4 \* 60 / 6 = 40 evts in total.

The idea is that we make it consistent, of course.
@ericvaandering @ticoann what do you think? Maybe we can quickly discuss it during the meeting.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
@@ -5,7 +5,7 @@ _DataProcessing_
 Base module for workflows with input.
 """
 
-from WMCore.Lexicon import dataset, couchurl, identifier, block
+from WMCore.Lexicon import dataset, block
 from WMCore.WMSpec.StdSpecs.StdBase import StdBase
 from WMCore.WMSpec.WMWorkloadTools import makeList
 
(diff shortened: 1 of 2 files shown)


amaltaro:
@drkovalskyi Dima, please see my comment above. 
The idea with this fix is that McM will inject a TimePerEvent (the time it takes to process one event, in single core) and the number of cores (Multicore parameter). Then, ReqMgr would use these two info provided in order to do the right 8h automatic job splitting.

However, if they cannot provide this information and they stick to what the give us "time it takes to process X events with my multicore setup", then we need to adapt the agent to properly handle it.

Could you please get this info for us, such that we avoid this issue from popping up in production again?

ericvaandering:
I'll say what I said in our meeting. What Alan is suggesting goes along with the viewpoint we should be evolving towards that the job of McM is to tell us how long things run and how many cores it can efficiently do so (the two numbers Alan is requesting). How many cores are actually used should eventually be in the domain of computing. We may run fewer cores do to operational issues.

ericvaandering:
So if what Alan suggests can be done easily, we should do that. If not, WMAgent will adapt.

vlimant:
There is no issue with having a consistent treatment of TimePerEvent, as long as the new convention is clearly established and advertised with PPD, ahead of time, and that they have a way to measure it from the fwk job report (I think it's trivial available) @anorkus.
There will be required changes also in wmagentscripts @mcremone @areinsvo when this happens.

Not sure whether the injected Memory parameter should follow such a paradigm shift too. Best would be to be provided BaselineMemory and MemoryPerCore. Where baseline memory is for whatever Multicore is provided at injection time.

amaltaro:
This PR is no longer relevant, see https://github.com/dmwm/WMCore/wiki/TimePerEvent-convention
and https://github.com/dmwm/WMCore/pull/8415

PR #6602, opened 2016-02-19
Add {un}mergedLFNBase to the workload cache

amaltaro:
It seems this info is available in workload_cache docs only when it's provided during request creation.
If it's not, but only during assignment, then we don't get these key/value pairs in the cache doc.

It will always be available from now on, forever ;)

@ticoann please review. I just did a test in my VM and it works. Let us push this to current testbed please.


amaltaro:
@ticoann please merge and at it to the next tag (1.0.14.pre3?)

PR #6603, opened 2016-02-19
Fix a problem when comparison is done on empty lists

vkuznet:


amaltaro:
Valentin, I don't really get this fix. It does not fix anything, it will only change from a set to a list assigned to 'res' in the end.

If both whitelist and blacklist are empty lists, it returns an empty set, which has length 0, Which does not go inside the if statement and does not raise an exception, as it's expected to.

vkuznet:
Alan, the tricky part is the following:

when code parse the web entry, which is empty string what you get is
```
whiteList = ['']
blackList = ['']
res = (set(whiteList) & set(blackList))
```

and here ```res=['']``` and it has non zero length.

While the fix I made does the correct thing it collects ONLY sites from
white/black lists which are NON EMPTY strings. Therefore I fix the problem when
no sites is provided, i.e. len(['']).

src/python/WMCore/WMSpec/WMWorkloadTools.py
```diff
@@ -215,7 +215,7 @@ def validateSiteLists(arguments):
         whiteList = [whiteList]
     if not isinstance(blackList, list):
         blackList = [blackList]
-    res = (set(whiteList) & set(blackList))
+    res = [s for s in list(set(whiteList) & set(blackList)) if s]
```
ericvaandering: what does list(set & set) give you if its empty?

ticoann:
I think it has to be handle before the it hits this code client (browser) shouldn't send [''] but [] (jsonized)

src/python/WMCore/WMSpec/WMWorkloadTools.py
vkuznet: Eric, it's not about empty list, it is about the case when lists contains empty
strings, e.g. whiteList=[''] and blackList=[''].

The result of set operations would be set(['']) and which is a list with single
element and that element is empty string. Once you take a len(set['']) it gives
1. The logic below which check len(res) is wrong because both whiteList and
blackList contain an empty string ''.

vkuznet:
That I agree. But code does not protect against such case.

src/python/WMCore/WMSpec/WMWorkloadTools.py
ericvaandering: Ah, thanks. I understand now. Lines 215 and 217 are the real problem. Could add "and blackList" and else blackList = []

vkuznet: On  0, Eric Vaandering <notifications@github.com> wrote:
> > @@ -215,7 +215,7 @@ def validateSiteLists(arguments):
> >          whiteList = [whiteList]
> >      if not isinstance(blackList, list):
> >          blackList = [blackList]
> > -    res = (set(whiteList) & set(blackList))
> > +    res = [s for s in list(set(whiteList) & set(blackList)) if s]
> 
> Ah, thanks. I understand now. Lines 215 and 217 are the real problem. Could add "and blackList" and else blackList = []

yes, exactly. And, it is matter of preferences, either we need to modify
215/217 lines as you suggested or define res as I propose. In addition,
Seangchan should fix code which pass arguments here and properly json'ise
arguments.

amaltaro:
Ok, I see now, thanks!

src/python/WMCore/WMSpec/WMWorkloadTools.py
ericvaandering: IMO, fixing 215/217 makes it more obvious what you are protecting against, but I'm not asking you to change it.

ticoann:
replaced by #6611

PR #6607, opened 2016-02-20
remove GlobalMonitor from rpm

ticoann:


src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrConfiguration.py
```diff
@@ -118,24 +115,4 @@ def reqMgrConfig(
     active.create.requestor = user
     active.create.cmsswDefaultVersion = 'CMSSW_5_2_5'
 
-    if addMonitor:
```
ericvaandering: Can you remove addMonitor from the function signature too?

ericvaandering: git grep says you can.

PR #6610, opened 2016-02-21
Fix the rest of the print statements that are not in the src/python 

ericvaandering:
and test/python directories


ericvaandering:
test this please

PR #6613, opened 2016-02-22
Force LHE templates to be assigned to CERN

amaltaro:
they always fail going to FNAL, so lets make sure they don't.


amaltaro:
It's only test injection, so I'm merging it.

ticoann:
Thanks Alan

PR #6614, opened 2016-02-22
convert frwj report for wmarchive

ticoann:
1. remove the layer, ("steps",  "checksum", "dataset", "sources")
2. convert name in output =  {"OutputPFN": "outputPFN", "inputPath": "inputDataset",
                                "lfn": "outputLFN", "input": "inputLFNs", "inputpfns": "inputPFNs"}
3. remove the field from output  ["InputPFN", "pfn"] - not useful information
4. convert the dataset pattern {"outputDataset": ["primaryDataset", "processedDataset", "dataTier"]}
5. change the run structure  {'runNumber': 123, lumis: [123, 124]}

### DIFF
--- a/src/python/WMCore/Services/WMArchiver/WMArchiver.py
+++ b/src/python/WMCore/Services/WMArchiver/WMArchiver.py
@@ -1,39 +1,7 @@
 from __future__ import (division, print_function)
 
-import socket
-import time
-import copy
-import WMCore
 from WMCore.Wrappers import JsonWrapper
 from WMCore.Services.Service import Service
-
-def _changeRunStruct(dictData, sourceKey):
-    if sourceKey in dictData:
-        # convert output module
-        for key, valueList in dictData[sourceKey].items():
-            for valueDict in valueList:
-                if "runs" in valueDict:
-                    newValue = [{"runNumber": run, "lumis": lumis}  for run, lumis in valueDict["runs"].items() ]
-                    valueDict["runs"] = newValue
-    return
-
-def _convertDictToList(dictData, sourceKey, keyName, valueName):
-    _changeRunStruct(dictData, sourceKey)
-    newData = [{keyName: key, valueName: value}  for key, value in dictData[sourceKey].items() ]
-    dictData[sourceKey] = newData
-
-def convertToArchiverFormat(fwjr):
-    """
-    """
-    newFWJR = copy.deepcopy(fwjr)
-    if newFWJR.get("steps", False):
-        steps = newFWJR['steps']
-        for key in steps:
-            if "output" in steps[key]:
-                _convertDictToList(steps[key], 'output', "outputModule", "value")
-            if "input" in steps[key]:
-                _changeRunStruct(steps[key], 'input')
-    return newFWJR
     
 class WMArchiver(Service):
     """
@@ -57,18 +25,7 @@ class WMArchiver(Service):
         self.encoder = JsonWrapper.dumps
         Service.__init__(self, httpDict)        
     
-    def createArchiverDoc(self, job_id, fwjr):
-        """
-        job_id is jobid + retry count same as couch db _id
-        """
-        newfwjr = convertToArchiverFormat(fwjr)
-        # append meta data in fwjr
-        newfwjr['meta_data'] = {'agent_ver': WMCore.__version__,
-                             'host': socket.gethostname().lower(),
-                             'fwjr_id': job_id,
-                             'ts': int(time.time())
-                             }
-        return newfwjr
+
     
     def archiveData(self, data):
         return self["requests"].post('', {'data': data})[0]['result']
--- a/src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
+++ b/src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
@@ -6,6 +6,7 @@ from __future__ import (division, print_function)
 import logging
 import traceback
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
+from WMCore.Services.WMArchiver.DataMap import createArchiverDoc
 from WMCore.Services.WMArchiver.WMArchiver import WMArchiver
 from WMCore.Services.FWJRDB.FWJRDBAPI import FWJRDBAPI
 
@@ -44,7 +45,7 @@ class ArchiveDataPoller(BaseWorkerThread):
             jobIDs = []
             archiverDocs = []
             for job in data:
-                doc = self.wmarchiver.createArchiverDoc(job["id"], job['doc']["fwjr"])
+                doc = createArchiverDoc(job["id"], job['doc']["fwjr"])
                 archiverDocs.append(doc)
                 jobIDs.append(job["id"])
                 
--- a/src/python/WMCore/Services/WMArchiver/DataMap.py
+++ b/src/python/WMCore/Services/WMArchiver/DataMap.py
@@ -1,0 +1,215 @@
+from __future__ import (division, print_function)
+
+import socket
+import time
+import copy
+import WMCore
+
+# From top level
+
+WMARCHIVE_TOP_LEVEL_REMOVE_OUTSIDE_LAYER = ["steps"]
+# convert data format under stpes["cmsRun1"/"logArch1"/"stageOut1"]["output"]
+WMARCHIVE_REMOVE_OUTSIDE_LAYER = ["checksum", "dataset"]
+# convert to list from str
+WMARCHIVE_CONVERT_TO_LIST = ["OutputPFN", "lfn"]
+
+WMARCHIVE_DATA_MAP = {"OutputPFN": "outputPFNs", "inputPath": "inputDataset",
+                      "lfn": "outputLFNs", "input": "inputLFNs", "inputpfns": "inputPFNs"}
+
+
+WMARCHIVE_REMOVE_FIELD = ["InputPFN", "pfn", "user_dn", "user_vogroup", "user_vorole"]
+WMARCHIVE_COMBINE_FIELD = {"outputDataset": ["primaryDataset", "processedDataset", "dataTier"]}
+
+WMARCHIVE_FILE_REF_KEY = ["lfn", "pfn", "files"]
+
+PERFORMANCE_TYPE = {'cpu': {'AvgEventCPU': float,
+                            'AvgEventTime': float,
+                            'MaxEventCPU': int,
+                            'MaxEventTime': float,
+                            'MinEventCPU': int,
+                            'MinEventTime': float,
+                            'TotalEventCPU': int,
+                            'TotalJobCPU': float,
+                            'TotalJobTime': float},
+                    'memory': {'PeakValueRss': int,
+                               'PeakValueVsize': int}
+                    }
+                                       
+def combineDataset(dataset):
+    dataset["outputDataset"] = "/%s/%s/%s" % (dataset["primaryDataset"], dataset["processedDataset"], dataset["dataTier"])
+    del dataset["primaryDataset"]
+    del dataset["processedDataset"]
+    del dataset["dataTier"]
+    return dataset
+    
+def changeRunStruct(runDict):
+    return [{"runNumber": int(run), "lumis": lumis}  for run, lumis in runDict.items() ]
+    
+
+def convertInput(inputList):
+    for inputDict in inputList:
+        if "runs" in inputDict:
+            inputDict['runs'] = changeRunStruct(inputDict["runs"])
+    return inputList
+
+def typeCastPerformance(performDict):
+    for key in performDict:
+        for param in performDict[key]:
+            if key in PERFORMANCE_TYPE:
+                if param in PERFORMANCE_TYPE[key]:
+                    if performDict[key][param] in ["-nan", "nan", "inf"]:
+                        performDict[key][param] = PERFORMANCE_TYPE[key][param](-1)
+                    else:
+                        performDict[key][param] = PERFORMANCE_TYPE[key][param](performDict[key][param])
+    return performDict
+            
+    
+def convertOutput(outputList):
+    newOutputList = []
+    for itemList in outputList:
+        newOutputList.extend(itemList)
+    
+    for outDict in newOutputList:
+        for field in WMARCHIVE_REMOVE_FIELD:
+            if field in outDict:
+                del outDict[field]
+                
+        for field in WMARCHIVE_CONVERT_TO_LIST:
+            if field in outDict and isinstance(outDict[field], basestring):
+                outDict[field] = [outDict[field]]
+            
+        for oldKey, newKey in WMARCHIVE_DATA_MAP.items():
+            if oldKey in outDict:
+                outDict[newKey] = outDict[oldKey]
+                del outDict[oldKey]
+                
+        if "runs" in outDict:
+            outDict['runs'] = changeRunStruct(outDict["runs"])
+        
+        if "checksums" in outDict:
+            outDict.update(outDict["checksums"])
+            del outDict["checksums"]
+        
+        if "dataset" in outDict:
+            outDict.update(combineDataset(outDict["dataset"]))
+            del outDict["dataset"]
+            
+        for key, value in outDict.items():
+            # set the default value for None to empty string
+            if value == None or value == "None":
+                outDict[key] = ""
+            
+            
+    return newOutputList
+
+def convertStepValue(stepValue):
+    if "input" in stepValue:
+        # remove source layer
+        stepValue['input'] = convertInput(stepValue['input'].get('source', {}))
+    if "output" in stepValue:
+        # remove output module name layer
+        stepValue['output'] = convertOutput(stepValue['output'].values())
+    
+    if "performance" in stepValue:
+        typeCastPerformance(stepValue["performance"])
+        
+    return stepValue
+            
+def convertSteps(steps):
+    stepList = []
+    for key, value in steps.items(): 
+        stepItem = {}
+        stepItem['name'] = key
+        stepItem.update(convertStepValue(value))
+        stepList.append(stepItem)
+    return stepList
+
+def convertToArchiverFormat(fwjr):
+    """
+    """
+    newFWJR = copy.deepcopy(fwjr)
+    if "steps" in newFWJR:
+        newFWJR["steps"] = convertSteps(newFWJR["steps"])
+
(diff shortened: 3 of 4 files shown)


ticoann:
The example is updated https://github.com/dmwm/WMArchive/wiki/fwjr.json
@yuyiguo, @vkuznet, please review.
There still need to be set the default values correctly (please point if you find any problems)

amaltaro:
@ticoann are you trying to shrink this data by removing only outer dict level? Or could it be a few inner keys only?

I believe data volume will be critical here, so we should get rid of any useless and repeated information, right?

vkuznet:
Seangchan,
the schema is improving, but still has many holes in it.
Basically all empty dicts or lists should be filled with whatever structure
they may contain. Here are they are:

```
{'cmsRun1': {'analysis': {},
             'cleanup': {},
             'errors': [],

....

'fallbackFiles': [],
 'logArch1': {'analysis': {},
              'cleanup': {},
              'errors': [],
              'input': {},
              'logs': {},
			  'output':[...
				'runs': [],
				...
			  ],
			  'parameters': {},
              'performance': {'cpu': {},
                              'memory': {},
                              'multicore': {},
                              'storage': {}},

'skippedFiles': [],
 'stageOut1': {'analysis': {},
               'cleanup': {},
               'errors': [],
               'input': {},
               'logs': {},
               'output': [],
               'parameters': {},
               'performance': {'cpu': {},
                               'memory': {},
                               'multicore': {},
                               'storage': {}},
....
```

ALL of those empty structures MUST be filled with some fake info,
otherwise the schema generation tool will not know which data types
to expect in empty dicts/lists.


But for some attributes, e.g. performance dict, I don't think
I understand why cpu/memory/multicore/storage should be a dict types?

Best,
Valentin.

ticoann:
> @ticoann are you trying to shrink this data by removing only outer dict level? Or could it be a few inner keys only?

As far as i understood, nested structure give more problem on the data structure (searching) used by hadoop. Valentin would know better about this.

> I believe data volume will be critical here, so we should get rid of any useless and repeated information, right?
I'm not sure how it would be critical with data volume. But yes, we will try to reduce the duplicate information as much as possible.  We talked Dirk briefly some of the data would be exactly the same but has different meaning depend on the section it resides.
i.e. run and lumi mask in input section has different meaning than the same info in output section.
Input is what we try to intend and output is what we get. It is the since the unit is a job. But it could be different when we break the job in smaller unit.

But mostly I am sure what all the information is in framework job report how this will evolve. So please advice if you find any information which could be removed.

ticoann:
Hi Valentin, I updated information with some fake data in  https://github.com/dmwm/WMArchive/wiki/fwjr.json.

Basically all the sections originally in steps has the same structure.
So cmsRun1, logArch1, stageOut1 will have the same structure. (Although some ot them are always empty in logArch1 and stageOut1)

For performance data it is configurable so key values are arbitrary (but with in the list of the choices)
I am not sure what are the list of choices for each group. But the example is what is set in agent currently - which never changed. Although it could change, but I am not sure what would be added. So we have to change the schema when that happens.

For 
```
'analysis': {}, 
'cleanup': {}, 
'logs': {},  
'parameters': {}
```

I am not sure about the structure of those field.  As far as I checked, I couldn't fine any data there. So I am not completely sure what the structure would be. We can completely remove those but not sure whether that field will be field up some times.  My guess is analysis part would be used in crab, but not completely sure.

Thanks

ticoann:
test this please

amaltaro:
My comment will completely depend on whether we prefer saving disk space or cpu cycles. In case we want to save space, here are my comments (based on the json example):
- under step / errors, we have a list of errors. Do you know whether we are aggregating that info? Or if we have 10 errors exactly the same, that list length will be 10? 
 - under step / input, 'guid' is already party of the lfn, so it could be removed (unless we want to use it to optimize database lookup).
 - under step / output, maybe we could also remove 'async_dest', 'catalog', 'guid' and 'cksum' (aren't we using only adler32 now?)? Still here, I'm not convinced we should keep 'user_dn', 'user_vogroup' and 'user_vorole'.

vkuznet:
Alan and others,
since we're talking about long term storage and backed up by HDFS we don't need
to worry about size to some extend. If you add more fields that's fine. We also
PLAN to do aggregation therefore anything useful from terms of aggregated
statistics should go to the report. If you need to find who generated samples
then user DN we must have. Basically, your questions should be converted to
what you want to do with stored document. Any queries you foresee must have
appropriate information in FWJR.

But as Seangchan mentioned we should try to avoid nested structure, i.e.
dict contains list of dicts, which by itself contains other dict/lists.
This structure is easy when you construct the document by really hard to use in
searching. The flatten structure is better for storage and search.
Ideally I would prefer to have one level of nestness, e.g. a dict with
key/value pairs, but not a dict with another dict as a value for some keys.
Such flatten structure is easy to extend and that is allowed, e.g. we can
expand schema, but we should not shrink it.

V.

vkuznet:
Seangchan,
even though I understand as human your statement

> Basically all the sections originally in steps has the same structure.
> So cmsRun1, logArch1, stageOut1 will have the same structure. (Although some ot them are always empty in logArch1 and stageOut1)
> 

we MUST fill out those structure with some fake values for the schema file.
Please do so. You need to understand that once we put schema in code it will
check it at RUN TIME and if schema file has empty data-structure and at RUN
TIME it will be provided with "the same structure" the code will just BREAK (or
worse seg-fault/throw exceptions).  Don't think in python way, think in terms
of C/C++ code. You can't have undefined variable or change data type of the
variable in C/C++ code.

So, fill out those "identical" structures in cmsRun1/logArch1/stageOut1.

> For performance data it is configurable so key values are arbitrary (but with in the list of the choices)
> I am not sure what are the list of choices for each group. But the example is what is set in agent currently - which never changed. Although it could change, but I am not sure what would be added. So we have to change the schema when that happens.

I rather prefer to change code that you WILL NOT assign arbitrary stuff. The
schema change is a consequence of code structure and it is better to let code behave
correctly. Unfortunately this is "bad behavior of python" which allows to assign
at run-time anything and change data-type of variables at run-time.

Think about this way. We'll be running in auto mode where each agent will feed
FWJR into WMArchive. If one agent decided to add some fields on a fly at
run-time it means ALL docs will be failed during injection into WMArchive. Can
you imaging how much time we will spend cleaning up such mess, verifying which
docs were correctly written and which are not. To avoid that it is much better
to only allow agents to change FWJR with scheduled upgrade. Before the software
upgrade we change schema and up-load it and then we change agents.

Please, please don't put us into the mess because of these lazy statements.
Change the code to make it predictable.

> 
> For 
> ```
> 'analysis': {}, 
> 'cleanup': {}, 
> 'logs': {},  
> 'parameters': {}
> ```
> 
> I am not sure about the structure of those field.  As far as I checked, I couldn't fine any data there. So I am not completely sure what the structure would be. We can completely remove those but not sure whether that field will be field up some times.  My guess is analysis part would be used in crab, but not completely sure.

Hmm, but there is a code which generates at least keys and associate with them
empty dicst, right? Where this code? We MUST find it. Don't remove
data-structures, remove the code which generates them. If they will be required
then we'll need a code which will generate them properly. Here again, the
change of schema will be a consequence of agent code change which will bring
these data-structures in controlled way. For example, new release will create
them and before deployment of new release we'll be able to generate new schema
with them and put it in place before we'll acquire the data.

src/python/WMCore/Services/WMArchiver/WMArchiver.py
```diff
@@ -7,32 +7,82 @@
...
+def convertSteps(steps):
+    for key, value in steps.items(): 
+        steps[key] = convertStepValue(value)
+    return steps
 
 def convertToArchiverFormat(fwjr):
     """
     """
     newFWJR = copy.deepcopy(fwjr)
     if newFWJR.get("steps", False):
```
ericvaandering: why not ```if 'steps' in newFWJR```?

src/python/WMCore/Services/WMArchiver/WMArchiver.py
```diff
@@ -7,32 +7,82 @@
...
+            outDict.update(outDict["checksums"])
+            del outDict["checksums"]
+        if "dataset" in outDict:
+            outDict.update(combineDataset(outDict["dataset"]))
+            del outDict["dataset"]
+            
+    return newOutputList
+
+def convertStepValue(stepValue):
+    if stepValue.get("input", False):
```
ericvaandering: Why not use ```in``` per the comment below?

src/python/WMCore/Services/WMArchiver/WMArchiver.py
ticoann: Yes, that is more clear I will fix it.

vkuznet:
Seangchan,
here is more stuff to fix:
- fix data-types, e.g. 
"PeakkValueRss": "0", should be "PeakValueRss": 0.0 
"PeakValueVsize": "0" should be "PeakValueVsize": 0.0
"TotalJobCPU": "0.39894", should be "TotalJobCPU": 0.39894
....

- identify correct data type, e.g.
'size': 0, should it be int or float?

'location': None should be 'location: "some string"

'user_dn': None, should be 'user_dn': "some string"

etc.

amaltaro:
My argument for removing the user DNs is that it was made available and - I think - never used. Probably with the assumption that it would be eventually used. Since it has passed a few years without using it, I think it'd be good to get rid of it.
BTW, there is no single person responsible/owner of the datasets, it all comes through McM with the same user/cert. The Physics group is encoded in the request names.

test/python/WMCore_t/Services_t/WMArchiver_t/WMArchiver_t.py
```diff
@@ -140,13 +140,13 @@ class WMArchiver_t(unittest.TestCase):
 
     def testConvertToArchiverFormat(self):
         newData = convertToArchiverFormat(SAMPLE_FWJR)
-        outputModules = set([a['outputModule'] for a in newData['steps']['cmsRun1']['output']]) 
-        outModules = set(SAMPLE_FWJR['steps']['cmsRun1']['output'].keys())
-        self.assertEqual(outputModules - outModules, set())
+        self.assertEqual(newData.get("steps", None), None)
```
ericvaandering: This could be simply assertFalse('steps' in newData)

yuyiguo:
@ticoann , @vkuznet 
Shengchan, can you merge this PR? Also, can you copy recent  fwjr to vocms013 that has bunch of free storage? 

Valentin, 
couchDB seems not to include in the software sets on vocms013. Instructions on installing couchDB cmsweb based node? 

Thanks,
Yuyi

vkuznet:
Yuyi, why do we need couch on a node where we're not going to use it?

yuyiguo:
@vkuznet , @ticoann 

If I remembered correctly, the fwjr copied from production nodes is in couchDB and I need to read them from couchDb in order to insert into MongoDB using WMArchive format.  Seangchan, please correct me if I am wrong.

ticoann:
@yuyiguo, yes, you are right we need couch server to retract the data from fwjr. Didn't we get the machine for this at fnal before?

vkuznet:
Yuyi,
I still don't understand why do you need couchdb RPMs? We don't need
couch DB on vocms013 since you can read docs from couch via HTTP(s) anyhow.
So, I would assume your python (pseudo)-code will do that:

```
import urllib, urllib2
url = "YOUR COUCH URL"
params = {} # your dict of what you search for
req = urllib2.Request(url, urllib.urlencode(params))
response = urllib2.urlopen(req)
result = json.load(response.read())

import pymongo
client = pymongo.MongoClient()
# some code to insert data into mongo
```

This code will read data from couch, but you don't really need a couchdb.

ticoann:
Hi Valentin, since this is production data we don't want to directly read from couch. Doesn't want to  overload the production agent. So I was suggesting copy the file directly to other machine and run the couch server there.  @yuyiguo, Yuyi whether are you going to transfer the data (to Mongo db) vocms013?
We can use one of the fnal test machine how ever if you need to transfer to cern machine firewall has to be open.

yuyiguo:
@ticoann , @vkuznet 

Seangchan, 
The MongoDB is on vocms013. It would be nice to copy the data to vomcms013  and setup a CouchDb on vomcs013 directly. FNAL now has very restrict rules on open a port to outside world. Since our cmsdag machines are not supposed to open ports to outside FNAL. We may need a lot of paper work to do so. I thought it is easy to just install a couchDB on vocms013 directly.

vkuznet:
Hmm, why not to use reqmgr2-dev then? Of course we can build WMArchive with couchdb
But a lot of stuff we don't really need, e.g. erlang, spidermonkey, couch.
Then we need to configure node to setup couch service. Can we just use couch
on existing dev VMs, e.g. reqmgr2-dev?

vkuznet:
Yuyi,
the couch is a back-end and it comes as dependency for some service.
In order to install couch, I need to make WMArchive depend on it.
Then I need to build WMArchive RPMs with it, install them.
Start separately couch. As I wrote I rather prefer to use existing
VMs where couch comes naturally, e.g. reqmgr2-dev.

My understanding we only need couch once to read the data.

yuyiguo:
reqmgr2-dev is fine with me assuming it has enough storage to hold the data.

ticoann:
reqmgr2-dev doesn't have enough space. Yuyi, haven't you set up MongoDB at FNAL machine before?

ticoann:
some more unittest still need to be updated (will create the new issue) but since Yuyi need the code merging it.

ticoann:
> The MongoDB is on vocms013. It would be nice to copy the data to vocms013 and setup a CouchDb on vomcs013 directly. FNAL now has very restrict rules on open a port to outside world. Since our cmsdag machines are not supposed to open ports to outside FNAL. We may need a lot of paper work to do so. I thought it is easy to just install a couchDB on vocms013 directly.

We don't need to open the port, do we? We will just push the data so vocms013 has to open the port but not FNAL machine.

vkuznet:
I don't like any idea of opening ports. Why do we need to open vocms013 ports?

We should not fail security measurements, otherwise we'll be panished and node
will be revoked.

I can install manually everything, e.g. erlang and couch, if you need it to run
on vocms013.

ticoann:
> I can install manually everything, e.g. erlang and couch, if you need it to run on vocms013.

I know you can, just not to give you too much extra work. :smile:

yuyiguo:
@ticoann, @vkuznet 

To me the best way is to install couchDB on vocms013, then we don't have security or space issue. Valenint, do you have instructions on install a stand alone couchDB? If so, please point them  to me. If you think it is easier to do it yourself instead of explaining them to me, go head please.

vkuznet:
Yuiy, Seangchan,
please read vocms013:/data/wma/couchdb/HOWTO and build/compile/install/run your
own couch without messing around with WMArchive environment.

I encourage you to learn these basic skills to install manually any unix
package since it becomes handy in many circumstances. The basic idea is to
grab tarbal, untar, configure, make and make install.

For steps in HOWTO I added couple of system packages, e.g. ncurses and icu,
which were required by couch build.

Once you have your own couch in your own area you can put data in it.
Once you're done you may easily clean-up your own area without touch
WMArchive environment since I want to keep it clean (it will take more of
my time to clean messy environment or reinstall the node than build local
couch on my own).

Best,
Valentin.

yuyiguo:
Ok, I will build it .
Yuyi

Sent from my iPhone

On Mar 2, 2016, at 4:13 PM, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:

Yuiy, Seangchan,
please read vocms013:/data/wma/couchdb/HOWTO and build/compile/install/run your
own couch without messing around with WMArchive environment.

I encourage you to learn these basic skills to install manually any unix
package since it becomes handy in many circumstances. The basic idea is to
grab tarbal, untar, configure, make and make install.

For steps in HOWTO I added couple of system packages, e.g. ncurses and icu,
which were required by couch build.

Once you have your own couch in your own area you can put data in it.
Once you're done you may easily clean-up your own area without touch
WMArchive environment since I want to keep it clean (it will take more of
my time to clean messy environment or reinstall the node than build local
couch on my own).

Best,
Valentin.

On 0, Yuyi Guo <notifications@github.com<mailto:notifications@github.com>> wrote:
> @ticoann, @vkuznet
>
> To me the best way is to install couchDB on vocms013, then we don't have security or space issue. Valenint, do you have instructions on install a stand alone couchDB? If so, please point them to me. If you think it is easier to do it yourself instead of explaining them to me, go head please.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/dmwm/WMCore/pull/6614#issuecomment-191440196

-

yuyiguo:
@vkuznet , @ticoann 

Thanks for the building instructions, Valentin.

Now couchdb is running on vocms013 under my area.

yuyiguo:
@ticoann 
SeangChan, Below is the data from the production fwjr.
Besides PeakValueRss and PeakValueVsize needed to change the data type, EventThroughput is mssing in the schema as well as storage and multicore.

u'performance': {u'cpu': {u'AvgEventTime': u'10.3341',
                                                             u'EventThroughput': u'0.0952297',
                                                             u'MaxEventTime': u'48.3678',
                                                             u'MinEventTime': u'4.46389',
                                                             u'TotalJobCPU': u'1963.63',
                                                             u'TotalJobTime': u'2101.53',
                                                             u'TotalLoopCPU': u'1962.28'},
                                                    u'memory': {u'PeakValueRss': u'1528.52',
                                                                u'PeakValueVsize': u'1931.56'},
                                                    u'multicore': {},
                                                    u'storage': {u'readAveragekB': 77.8474891246,
                                                                 u'readCachePercentageOps': 0.0,
                                                                 u'readMBSec': 0.0438598972596,
                                                                 u'readMaxMSec': 4832.84,
                                                                 u'readNumOps': 97620.0,
                                                                 u'readPercentageOps': 1.00032780168,
                                                                 u'readTotalMB': 7423.792,
                                                                 u'readTotalSecs': 0,
                                                                 u'writeTotalMB': 357.624,
                                                                 u'writeTotalSecs': 575158.0}},

vkuznet:
Seangchan,
we also discussed with Yuyi that it is better to split fileArray into two
arrays. One will keep LFNs and another PFNs. The reason for that is simple
we mostly will look-up data by LFNs, therefore we can reduce its file array
by factor of 2 which will significantly impact on index size in DB.
Valentin.

ticoann:
Thanks Yuyi and Valentin, I created the new issue. #6670

PR #6615, opened 2016-02-23
Promote block names to unicode to help mock DBS

ericvaandering:
Otherwise we have to double the size of the data file for u'block' and 'block'


ericvaandering:
@ticoann, please merge

PR #6616, opened 2016-02-24
Allow data-manager role to create requests in reqmgr2; check group me…

amaltaro:
First issue found by Antanas, their DN was not able to make a POST call to reqmgr2/data/request (create a new request), I had to temporarily give him admin privileges:
https://cmsweb.cern.ch/sitedb/prod/people/pdmvserv%40vocms081.cern.ch

This PR changes basically two things:
 a) we always check the group membership of the DN
 b) we allow mainly "data-manager" and "production-operator" to also make requests.

@ticoann please review. I still need to test it in my VM.

### DIFF
--- a/src/python/WMCore/ReqMgr/Auth.py
+++ b/src/python/WMCore/ReqMgr/Auth.py
@@ -3,28 +3,22 @@ Define Authentiction roles and groups for the access permission for writing the 
 WMCore.REST.Auth authz_match
 """
 ALL_ROLES = ['developer', 'admin', 'data-manager', 'production-operator', 'web-service']
+ALL_GROUP = ['reqmgr', 'facops', 'dataops']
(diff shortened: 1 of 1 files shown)


amaltaro:
BTW, there is also this file:
https://github.com/dmwm/WMCore/blob/6f3570a741779d209f0f720647642d51b64845da/src/python/WMCore/WMStats/Auth.py

with the same code. Git grep does not find any one using it. Shall I remove that code?

ticoann:
Thanks Alan, this give all the permission for every group, Is McM allowed to only creating and approving requests? Anyway, I will make this configurable so we can control this run time.  But lets push this for the testing. Do you want me to make new tag and make the pull request for testbed.

amaltaro:
Thanks! I suggest we cut a new tag later this evening FNAL timezone :-)
I'll try to improve the authz in the coming weeks.

Issue #6590, opened 2016-02-17
Add basic FJR attributes to job classad


bbockelm:
Since we're scraping `condor_history` and `condor_q` outputs, it would be useful if we could complement the job information with some more data about the job run.

WMRuntime can scalably set attributes in the classad by doing:

```
condor_chirp set_job_attr_delayed key value
```

where `value` is quoted according to classad rules.  This sends the additional key/value info along with the next scheduled update (so there aren't scalability issues here).

Particularly, I'd like to get the following pieces of information:
- Exit code of the job.
- Number of MB transferred from the last CMSSW step.
- Events process from the last CMSSW step.
- Log archive LFN.

bbockelm:
@hufnagel also suggested that he may have time to tackle this.

ticoann:
Thanks Dirk. :smile:

hufnagel:
Started looking at this. WMAgent runtime code is highly modular, it's a sequence of steps that are executed, the global wrapper code is very short and minimal functionality. As such, there isn't really a good place to put this as a global implementation, it's better off in the step executors. Would have to touch both the cmssw and logarchive executor implementations. Likely want to have the chirp call implemented as a function that is callable from both executors instead of implementing it twice.

@bbockelm , this causes a problem though because this way you won't get the exit code of a job, just the exit code of the cmssw step(s). You also won't just get the info from the LAST cmssw step, but the infos from ALL cmssw steps (we can distinguish them by key name).

The only pieces of code that really looks at all these things together is the JobAccountant, ie. a central daemon. We don't aggregate this information at runtime.

bbockelm:
Hi Dirk,

Can't we have both?  I.e., have the CMSSW executor update attributes name CmsFooExitCode and CmsExitCode?

I suggested the last step because I figured that was easiest.  If we can get all, that's even better.

Do note that classad attributes aren't json - there are various quoting rules you might hit if reporting each step.  What do the step names look like?

Brian

hufnagel:
Sure we can do both, one set of attributes per cmssw step, the other common so the last step wins in setting it. Step names for cmssw are usually cmsRun1, cmsRun2 etc.

hufnagel:
Another thing, what do you want the logarchive LFN for ? The one from the processing job is kind of useless as it is a temporary LFN in unmerged space. It survives maybe order one or two days at most.

hufnagel:
[...]

PR #6618, opened 2016-02-24
add support to call condor_chirp at runtime

hufnagel:
We want to publish certain information (as key/value pairs) into HTCondor during the runtime of a job. Add support to do so in the Executor base classe, so this can be called from different step types. Also implement publishing the cmsRun exit code.

We probably want to test this first before merging. I tested that it doesn't impact agent functionality, but haven't check that it actually does what it is supposed to do.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executor.py
+++ b/src/python/WMCore/WMSpec/Steps/Executor.py
@@ -7,7 +7,10 @@ Interface definition for a step executor
 
 """
 
+import os
 import sys
+import json
+import subprocess
 
 from WMCore.FwkJobReport.Report import Report
 from WMCore.WMSpec.WMStep import WMStepHelper
@@ -25,7 +28,6 @@ def getStepSpace(stepName):
     you use it
 
     """
-
     modName = "WMTaskSpace"
     if modName in sys.modules.keys():
         taskspace = sys.modules[modName]
@@ -47,7 +49,6 @@ def getStepSpace(stepName):
         msg += str(ex)
         raise RuntimeError(msg)
     return stepSpace
-
 
 
 class Executor:
@@ -104,11 +105,6 @@ class Executor:
             self.emulator.initialise(self)
             self.emulationMode = True
 
-
-
-
-
-
         return
 
 
@@ -123,7 +119,6 @@ class Executor:
         return
 
 
-
     def pre(self, emulator = None):
         """
         _pre_
@@ -136,6 +131,7 @@ class Executor:
 
         """
         return None
+
 
     def execute(self, emulator = None):
         """
@@ -163,3 +159,26 @@ class Executor:
 
         """
         return None
+
+
+    def setCondorChirpAttrDelayed(self, key, value):
+        """
+        _setCondorChirpAttrDelayed_
+
+        Util to call condor_chirp and publish the key/value pair
+
+        """
+        # construct condor_chirp binary location from CONDOR_CONFIG
+        condor_chirp_bin = None
+        condor_config = os.getenv('CONDOR_CONFIG', None)
+        if condor_config:
+            condor_config_dir = os.path.dirname(condor_config)
+            condor_chirp_bin = os.path.join(condor_config_dir, 'main/condor/libexec/condor_chirp')
+            if not os.path.isfile(condor_chirp_bin):
+                condor_chirp_bin = None
+
+        if condor_chirp_bin:
+            args = [ condor_chirp_bin, 'set_job_attr_delayed', key, json.dumps(value) ]
+            returncode = subprocess.call(args)
+
+        return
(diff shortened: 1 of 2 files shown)


src/python/WMCore/WMSpec/Steps/Executor.py
```diff
@@ -163,3 +158,26 @@ def post(self, emulator = None):
...
+        condor_chirp_bin = None
+        condor_config = os.getenv('CONDOR_CONFIG', None)
+        if condor_config:
+            condor_config_dir = os.path.dirname(condor_config)
+            condor_chirp_bin = os.path.join(condor_config_dir, 'main/condor/libexec/condor_chirp')
+            if not os.path.isfile(condor_chirp_bin):
+                condor_chirp_bin = None
+
+        if condor_chirp_bin:
+            args = [ condor_chirp_bin, 'set_job_attr_delayed', str(key), str(value.__repr__()) ]
```
ericvaandering: why str of repr? (and you can do repr(value) to get the same thing)

hufnagel: Is repr(value) guaranteed to return a string ? Each element of the argument list *has* to be a string, otherwise the subprocess.call() method will fail. Will change to repr(value) though.

hufnagel: Ok, reading the docs, repr(object) does return a string. So will try that and see if it works.

ericvaandering:
Ok. Didn't know if you were worried you would get unicode or something.

ericvaandering:
>>> a = 'foo'
>>> b = 2
>>> c = repr(a)
>>> c
"'foo'"
>>> a == c
False
>>> repr(b)
'2'

ericvaandering:
So rear is going to give you a string with quotes as the first and last character of the string. Don't know if that's what you want.

hufnagel:
There is enough ambiguity with how intermediate code layers in condor or the subprocess call will interpret quotes that I am just going to try it one way and see what I get.

hufnagel:
Tested, works. Having some problems, but they are confirmed to be related to a HTCondor bug in the version I tested on, not related to the code. @amaltaro @ticoann please review and merge.

Partial fix for #6590

bbockelm:
Hm - playing around, I think `repr` isn't doing what I wanted.  ClassAd quoting rules only use double-quotes; yet above Eric ended up with single quotes.

Before merge, can you make sure to post the resulting attributes from `condor_history`?

hufnagel:
cmst1@vocms001:/data/tier0 $ condor_history -l 186053.6 | grep Chirp
Chirp_WMCore_1_ExitCode = 0
Chirp_WMCore_cmsRun_ExitCode = 0
Chirp_WMCore_cmsRun1_ExitCode = 0
Chirp_WMCore_2_ExitCode = 0

bbockelm:
Yes - suspiciously, there's no strings.

Looks like json's formatting rules will avoid the problematic single-quotes.  Try:
```
>>> foo = "bar"
>>> json.dumps(foo)
'"bar"'
>>> foo = 'ba"z'
>>> json.dumps(foo)
'"ba\\"z"'
```

hufnagel:
That doesn't work for numbers. If a do a json.dumps() on a number I get 'number'.

hufnagel:
I can do json.dumps(str(temp)). Works the same for numbers and strings.

bbockelm:
Actually, that's right.  When the interpreter return `'2'`, it's actually a string of length 1 with the ASCII character for `2`.  Maybe this illustrates better:

```
>>> print json.dumps(2)
2
>>> print json.dumps("foo")
"foo"
>>> print json.dumps("f\"oo")
"f\"oo"
```

All of those are handled correctly in terms of classads.  However, 

```
>>> print json.dumps(str(2))
"2"
```
That's not correct.  ClassAds will interpret this as a string instead of an integer.

hufnagel:
Oh, so you still want to get the number as a number. Fine then, json.dumps(value) it is.

hufnagel:
Will test and then update the pull request...

hufnagel:
Tested. Also passed a string as a value and that seems to work too. Output of condor_history:

Chirp_WMCore_Test_ExitCode = "teststring"

Once jenkins is done this is ready to be reviewed and merged.

amaltaro:
What's the difference between Chirp_WMCore_cmsRun_ExitCode and Chirp_WMCore_cmsRun1_ExitCode (when stepName is cmsRun1 e.g.)?

It seems nothing will happen when we run jobs at sites without a condor batch system. Still, I'd prefer to have it tested - in a non-T0 env - before it gets merged.

hufnagel:
The difference is only for jobs with multiple CMSSW steps. Chirp_WMCore_cmsRun_ExitCode always contains the exit code from the last CMSSW step (with multiple calls to condor_chirp setting the same attribute the last call wins).

This has nothing to do with condor batch systems, it works at all sites that we access via gWMS, ie. all sites. Also has nothing to do with Tier0 environment. But go ahead and test it, just don't delay it too long, there is some pressure (from @bbockelm ) to have this merged and then also to make progress on the other half of the request.

amaltaro:
Ok, that's why CONDOR_CONFIG should be always set then, because it relies on gWMS instead of the local batch system. Got it!
Well, it looks there won't be any delays to get it merged ... :)

PR #6620, opened 2016-02-24
Change status, fix issue6619

vkuznet:


ticoann:
Sorry it was my mistake, I gave you wrong status. patch looks good. I will merge after running the test

PR #6627, opened 2016-02-24
Assign splitting button for individual requests

vkuznet:


vkuznet:
Seangchan,
I tested web UI, but I didn't check if POST requests came through ok.
Hopefully you did that, or at least will try before putting this into
production.
Valentin.

ticoann:
Hi Valentin,  I will test before cut the tag. The one more patch I have to put in so I will run the all of them together. Thanks.

PR #6635, opened 2016-02-25
Fix request assignment with reqmgr2 client

amaltaro:
@ticoann is this client working Ok for your tests (request creation + approval + assignment in the same command)? I had issues these days where it fails to find the request_name attribute in the config instance. 

It works with these changes, but AFAIK it used to work in the past :)


ticoann:
Hi Alan, it works ok with me create + approved + assign combined but it seems broken if I used -i option. Which is just create + approved. I had to have similar changes you have here to make it work.

amaltaro:
Thanks, Seangchan!

PR #6643, opened 2016-02-26
Add a rudimentary phedex emulator

ericvaandering:
This does something similar to the current emulator and DataBlockGenerator: it makes things up as it goes along. :-)


ericvaandering:
test this please

PR #6646, opened 2016-02-29
Reject assignment if AcqEra is None

amaltaro:
For those TaskChain ACDC cases that we end up with AcqEra None.
@ticoann once it passes the tests, we should push it into the next testbed tag.


ticoann:
Thanks Alan

PR #6648, opened 2016-03-01
Update workflow arguments for requests assigned in reqmgr2.

### LINKED ISSUE #6647: ReqMgr2 does not take the site white list into consideration
I see jobs running everywhere and an empty site whitelist (in the workload) list for the test workflows injected into testbed (anorkus_*).

amaltaro:
Fixes #6647

@ticoann please review and if possible run a test in your VM, mine is too messy to be confident :-)
BTW, I've just assigned a MC workflow and I see that GQ has the proper SiteWhiteList value, the same for the workload Production task, so it should fix the issue indeed.

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -374,43 +374,42 @@ class Request(RESTEntity):
 
         dn = cherrypy.request.user.get("dn", "unknown")
 
-        if ('SoftTimeout' in request_args) and ('GracePeriod' in request_args):
-            request_args['HardTimeout'] = int(request_args['SoftTimeout']) + int(request_args['GracePeriod'])
-
         if 'RequestPriority' in request_args:
             self.gq_service.updatePriority(workload.name(), request_args['RequestPriority'])
 
+        if 'SoftTimeout' in request_args and 'GracePeriod' in request_args:
+            request_args['HardTimeout'] = int(request_args['SoftTimeout']) + int(request_args['GracePeriod'])
+
         if "total_jobs" in request_args:
-            # only GQ update this stats
-            # request_args should contain only 4 keys 'total_jobs', 'input_lumis', 'input_events', 'input_num_files'}
+            # only GQ update these statistics: {'total_jobs', 'input_lumis', 'input_events', 'input_num_files'}
             report = self.reqmgr_db_service.updateRequestStats(workload.name(), request_args)
-        # if is not just updating status
-        else:
-            req_status = request_args.get("RequestStatus", None)
 
-            if len(request_args) >= 1 and req_status == None:
-                try:
-                    workload.updateArguments(request_args)
-                except Exception as ex:
-                    msg = traceback.format_exc()
-                    cherrypy.log("Error for request args %s: %s" % (request_args, msg))
-                    raise InvalidSpecParameterValue(str(ex))
+        req_status = request_args.get("RequestStatus", None)
+        if len(request_args) > 1 and req_status == "assigned":
+            try:
+                workload.updateArguments(request_args)
+            except Exception as ex:
+                msg = traceback.format_exc()
+                cherrypy.log("Error for request args %s: %s" % (request_args, msg))
+                raise InvalidSpecParameterValue(str(ex))
                 
-                # legacy update schema to support ops script
-                loadRequestSchema(workload, request_args)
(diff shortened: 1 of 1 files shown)


ticoann:
So this is the case when workflow is assigned and at the same time the parameters are updated, right?

amaltaro:
Yes, we must update args during assignment, at least a few them are mandatory (well, it was in reqmgr one) :)

ticoann:
Hi Alan, I found a couple of other bugs while I was looking at the code. I created the new patch. #6649

amaltaro:
@ticoann this is the patch I'm testing in my VM now (I basically merged your changes from #6649 + a few fixes).

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,42 @@ def _updateRequest(self, workload, request_args):
...
+                workload.updateArguments(request_args)
+            except Exception as ex:
+                msg = traceback.format_exc()
+                cherrypy.log("Error for request args %s: %s" % (request_args, msg))
+                raise InvalidSpecParameterValue(str(ex))
+                
+            # legacy update schema to support ops script
+            loadRequestSchema(workload, request_args)
+            # if we don't update the Team with this call, GQ does not see the Team name
+            report = self.reqmgr_db_service.updateRequestProperty(workload.name(), request_args, dn)
```
amaltaro: It was not clear to me, but if we don't call this update property, GQ fails to acquire work since it does not see any "Team" for that request.

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,42 @@ def _updateRequest(self, workload, request_args):
...
+        elif len(request_args) == 1 and req_status in ["aborted", "force-complete"]:
+            self.gq_service.cancelWorkflow(workload.name())
 
-            # If it is aborted or force-complete transition call workqueue to cancel the request
-            else:
-                if req_status == "aborted" or req_status == "force-complete":
-                    self.gq_service.cancelWorkflow(workload.name())
-                report = self.reqmgr_db_service.updateRequestProperty(workload.name(), request_args, dn)
+        # request assigned are already updated
+        if not req_status in [ None, "assigned"]:
```
amaltaro: We have to skip assigned here because the request change is already taken care by updateArguments (above)

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: You are right. 
` workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)` should update all the argument but team is not the part of workload property.  
we can use this and remove
`workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)`

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,42 @@ def _updateRequest(self, workload, request_args):
...
-                
-                # legacy update schema to support ops script
-                loadRequestSchema(workload, request_args)
-                # trailing / is needed for the savecouchUrl function
-                workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)
 
-            elif (req_status in ["closed-out"  "announced"]) and request_args.get("cascade", False):
-                cascade_list = self._retrieveResubmissionChildren(workload.name)
+        req_status = request_args.get("RequestStatus", None)
+        if len(request_args) > 1 and req_status == "assigned":
```
ticoann: We have to check the condition when req_status == None and len(request_args) >=1  otherwise splitting parameter won't be  updated

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: updateArguments won't update status change but updateRequestProperty would.

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: actually  following is needed since it updates the spec. Sorry for the noise
'workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)'

ticoann:
Agent side is not tested but GQ side works fine. Merging it.

ticoann:
Sorry wrong patch merge.

PR #6649, opened 2016-03-01
fix updating request parameters

### LINKED ISSUE #6647: ReqMgr2 does not take the site white list into consideration

ticoann:
fixes #6647
1. it won't allow any argument update other then assignment-approved or assigned status.
2. After assigned status it can't change any values in spec except Priority.
3. It still can update the arguments when it is not associated with status change. (This option might need to be eventually removed but need to support splitting parameter update for now which gets updated without status change)

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -374,43 +374,56 @@ class Request(RESTEntity):
 
         dn = cherrypy.request.user.get("dn", "unknown")
 
-        if ('SoftTimeout' in request_args) and ('GracePeriod' in request_args):
-            request_args['HardTimeout'] = int(request_args['SoftTimeout']) + int(request_args['GracePeriod'])
-
-        if 'RequestPriority' in request_args:
-            self.gq_service.updatePriority(workload.name(), request_args['RequestPriority'])
+        # allow Priority update any time.
+        if 'RequestPriority' in request_args: 
+            self.gq_service.updatePriority(workload.name(), int(request_args['RequestPriority']))
 
         if "total_jobs" in request_args:
             # only GQ update this stats
             # request_args should contain only 4 keys 'total_jobs', 'input_lumis', 'input_events', 'input_num_files'}
             report = self.reqmgr_db_service.updateRequestStats(workload.name(), request_args)
-        # if is not just updating status
+            
+        if ('SoftTimeout' in request_args) and ('GracePeriod' in request_args):
+            request_args['HardTimeout'] = int(request_args['SoftTimeout']) + int(request_args['GracePeriod'])
+        
+        
+        req_status = request_args.get("RequestStatus", None)
+
+        if (len(request_args) >= 1 and req_status == None) or \
+            (len(request_args) > 1 and req_status in ["assigned", "assignment-approved"]):
+            #Only allow extra value update for assigned status
+            try:
+                workload.updateArguments(request_args)
+            except Exception as ex:
+                msg = traceback.format_exc()
+                cherrypy.log("Error for request args %s: %s" % (request_args, msg))
+                raise InvalidSpecParameterValue(str(ex))
+            
+            # legacy update schema to support ops script
+            loadRequestSchema(workload, request_args)
+            # trailing / is needed for the savecouchUrl function
+            report = self.reqmgr_db_service.updateRequestProperty(workload.name(), request_args, dn)
+            workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)
+        
+        elif len(request_args) == 2 and req_status in ["closed-out", "announced"] and \
+             request_args.get("cascade", False):
+            # check whehter it is casecade option
+            cascade_list = self._retrieveResubmissionChildren(workload.name())
+            for req_name in cascade_list:
+                report = self.reqmgr_db_service.updateRequestStatus(req_name, req_status, dn)
+            # update original workflow status
+            report = self.reqmgr_db_service.updateRequestStatus(workload.name(), req_status, dn)
+            
+        elif len(request_args) == 1 and req_status != None:
+            # incase only the state transition happens
+            if req_status == "aborted" or req_status == "force-complete":
+                # cancel the workflow first
+                self.gq_service.cancelWorkflow(workload.name())
+            #update the request status in couchdb    
+            report = self.reqmgr_db_service.updateRequestStatus(workload.name(), req_status, dn)
         else:
-            req_status = request_args.get("RequestStatus", None)
-
-            if len(request_args) >= 1 and req_status == None:
-                try:
-                    workload.updateArguments(request_args)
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan please review. I will run the test in my vm. This should fix the sitewhite list problem as well

ticoann:
test this please

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,55 @@ def _updateRequest(self, workload, request_args):
...
-                cascade_list = self._retrieveResubmissionChildren(workload.name)
+            
+        if ('SoftTimeout' in request_args) and ('GracePeriod' in request_args):
+            request_args['HardTimeout'] = int(request_args['SoftTimeout']) + int(request_args['GracePeriod'])
+        
+        
+        req_status = request_args.get("RequestStatus", None)
+
+        if (len(request_args) >= 1 and req_status == None) or \
+            (len(request_args) > 1 and (req_status in ["assigned", "assignment-approved"])):
```
amaltaro: Ok, we need 'assigned' here for updating params during assignment and the 'assignment-approved' probably is for changes affecting the job splitting, right?
BTW, how about we make this expression
`if len(request_args) >= 1 and req_status in ["assigned", "assignment-approved", None]:`
in case we can drop the len() > 1.

amaltaro:
Sounds good, I'll push it to my VM later on to run a few tests. Thanks!

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,55 @@ def _updateRequest(self, workload, request_args):
...
-            elif (req_status in ["closed-out"  "announced"]) and request_args.get("cascade", False):
-                cascade_list = self._retrieveResubmissionChildren(workload.name)
+            
+        if ('SoftTimeout' in request_args) and ('GracePeriod' in request_args):
+            request_args['HardTimeout'] = int(request_args['SoftTimeout']) + int(request_args['GracePeriod'])
+        
+        
+        req_status = request_args.get("RequestStatus", None)
+
+        if (len(request_args) >= 1 and req_status == None) or \
```
ericvaandering: Don't need surrounding parentheses AND the continuation mark. One or the other is good

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,55 @@ def _updateRequest(self, workload, request_args):
...
+            
+            # legacy update schema to support ops script
+            loadRequestSchema(workload, request_args)
+            # trailing / is needed for the savecouchUrl function
+            workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)
+            
+            if req_status == "assigned":
+                report = self.reqmgr_db_service.updateRequestStatus(workload.name(), req_status, dn)
+        
+        elif (len(request_args) == 1 and req_status != None):
```
ericvaandering: No need for parentheses.

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,55 @@ def _updateRequest(self, workload, request_args):
...
+            if (req_status in ["closed-out"  "announced"]) and request_args.get("cascade", False):
```
ericvaandering: Same here

amaltaro:
I found a few more issues in this fix, I'm testing them right now.

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: Actually assignment-approved doesn't change anything now. but since something could change in assignment-approved possibly, I just added.

when there is req_status and len(request_args) == 1 we don't have to load workload and change its params. That is why I seperate the condition, for that case it will be handled below.

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: Ah, yes. I will remove one. Thanks

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,52 @@ def _updateRequest(self, workload, request_args):
...
+                cherrypy.log("Error for request args %s: %s" % (request_args, msg))
+                raise InvalidSpecParameterValue(str(ex))
+            
+            # legacy update schema to support ops script
+            loadRequestSchema(workload, request_args)
+            # trailing / is needed for the savecouchUrl function
+            report = self.reqmgr_db_service.updateRequestProperty(workload.name(), request_args, dn)
+        
+        elif len(request_args) == 1 and req_status != None:
+            if req_status in ["closed-out"  "announced"] and request_args.get("cascade", False):
```
amaltaro: typo in this list

amaltaro: and if request_args==1, then cascade is not there for sure

ticoann: Thanks Alan, fixed now

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -374,43 +374,53 @@ def _updateRequest(self, workload, request_args):
...
+                msg = traceback.format_exc()
+                cherrypy.log("Error for request args %s: %s" % (request_args, msg))
+                raise InvalidSpecParameterValue(str(ex))
+            
+            # legacy update schema to support ops script
+            loadRequestSchema(workload, request_args)
+            # trailing / is needed for the savecouchUrl function
+            report = self.reqmgr_db_service.updateRequestProperty(workload.name(), request_args, dn)
+        
+        elif len(request_args) == 2 and req_status in ["closed-out", "announced"] and \
```
amaltaro: this does not cover the case that we closed-out workflows without using the cascade option. We should check both for length 1 or 2, no?

ticoann: actually it covered by next condition. in case there is only status update only thing need to do is update the couch db.
```
elif len(request_args) == 1 and req_status != None:
```

amaltaro: Sorry, you are right! And in this cascade switch, does cascade_list have the workload.name() in its list? if not, then we still need to update the status for the original/parent workflow.

ticoann: Ah, good point. I don't think it returns original workflow name. I will update the patch. Thanks Alan, but go to sleep. :smile:

ticoann:
Thanks to Alan who fixed all my buggy code, patch seems to be working for Team and SiteWhitelist, still casecade part need to be tested but merging it first. (I have to create ACDC for testing the merge)

PR #6650, opened 2016-03-02
remove '@' from request name

### LINKED ISSUE #6637: ReqMgr2 request name structure
Antanas created a new reqmgr2 request today:
pdmvserv@vocms081.cern.ch_SUS-RunIIWinter15pLHE-00232_00146_v1__160226_103421_890

using this DN:
https://cmsweb.cern.ch/sitedb/prod/people/pdmvserv%40vocms081.cern.ch

Since we don't have the same username registration as we do for reqmg (oracle), I believe we are somehow getting the sitedb "username" or so, to prepend to the request name.

ticoann:
fixes #6637

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -18,6 +18,7 @@ TODO/NOTE:
 import time
 import cherrypy
 from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_START_STATE
+from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
 
 def initialize_request_args(request, config, clone = False):
     """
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, please review

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -70,8 +70,12 @@ def generateRequestName(request):
     current_time = time.strftime('%y%m%d_%H%M%S', time.localtime(time.time()))
     seconds = int(10000 * (time.time() % 1.0))
     request_string = request.get("RequestString", "")
+
     if request_string != "":
         request["RequestName"] = "%s_%s" % (request["Requestor"], request_string)
```
amaltaro: Instead changing the fina request name, how about we change this line to:
`request["RequestName"] = "%s_%s" % (request["Requestor"].split('@')[0], request_string)`
?

One more unrelated comment, I'd enforce a RequestString to be provided. Don't ask me for a reason, but it sounds more clear and... organized than having a "amaltaro_160302_1233_43" :-)

ticoann: I the example case, Requestor is pdmvserv@vocms081.cern.ch,  so pdmvserv is enough information that it is from host cert, right? I can change that.
 
Changing final case would be the same result unless there are other string includes '@' (Which shouldn't be the case anyway) Or we can created the lexicon for RequestString and prevent getting '@'. I guess you prefer second option

amaltaro: Yes, pdmvserv would be enough.

> Or we can created the lexicon for RequestString and prevent getting '@'. I guess you prefer second option
Sounds good to me as well. But we can make it for the next deployment IMO, a) enforce a RequestString and b) prevent such chars (@)

ticoann:
@amaltaro Alan, I updated the code. Please review.
I couldn't force RequestString in spec since tier0 doesn't use that. \
Also I just check "@' for validation, we can update lexicon if you know exactly what is allowed. I think it shouldn't allow '_'.  Let me know.

amaltaro:
Fix looks good Seangchan. RequestString is not enforced in the spec, but it's mandatory here generating the request name. I'm not sure whether it will cause problems to Tier0 then?

ticoann:
It shouldn't cause the problem in Tier0 since this function is only used in reqmgr2. What I said was if we enforce this in StdSpec.py, it might cause the problem in Tier0.

ticoann:
tested in vm

PR #6656, opened 2016-03-03
switch to PyCondor plugin

ticoann:
@amaltaro, Alan please review


amaltaro:
Sounds good, Seangchan! I didn't create this patch before because we may still need to deploy 1.0.12 agent out there.

ticoann:
Ah yes, I didn't think about that. You are right. We should merge this only we will switch to new Agent. Thanks

amaltaro:
I guess we can merge it now, no hurry though :-)

Issue #6662, opened 2016-03-03
ReqMgr2 does not accept MaxRSS per task


amaltaro:
```
  File "/data/srv/HG1603e/sw/slc6_amd64_gcc493/cms/reqmgr2/1.0.14.pre6/lib/python2.7/site-packages/WMCore/ReqMgr/Service/Request.py", line 394, in _updateRequest
    workload.updateArguments(request_args)
  File "/data/srv/HG1603e/sw/slc6_amd64_gcc493/cms/reqmgr2/1.0.14.pre6/lib/python2.7/site-packages/WMCore/WMSpec/WMWorkload.py", line 1739, in updateArguments
    self.setupPerformanceMonitoring(int(kwargs["MaxRSS"]),
TypeError: int() argument must be a string or a number, not 'dict'
```

ticoann:
When does this happen? How kwargs["MaxRSS"] is dict not number?

ticoann:
Never mind I found the example. :smile: I will try to fix this on time.

ticoann:
fix is added #6666

ticoann:
Alan, does this fixed now in testbed?

amaltaro:
Yep, worked like a charm! Thanks, Seangchan!

amaltaro:
Oups, I tested in reqmgr (one), not reqmg2. Leave it open until I get that test done in reqmgr2 please :)

amaltaro:
Thanks, Seangchan! It's fixed!

PR #6666, opened 2016-03-03
fix overwritting site whitelist and refactor code

### LINKED ISSUE #6658: SiteWhiteList is overwritten after initial creation in reqmgr2
It seems sitewhite list is initially created correctly but gets overwritten later with empty list. 

https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/sryu_MonteCarloFromGEN_Sryu_reqmgr2_160303_071533_4152

ticoann:
fixes #6658 and #6662


ticoann:
Still running the test. @amaltaro, Alan please review.
Splitting updates has different api so only update params happens assignment state (just added assignment approved as well since we could update there too. Although we are not doing that)

PR #6667, opened 2016-03-04
make team setting mendatory in assign

ticoann:
@amaltaro, Alan please review. (Although I probably need to merge before your review)
Tested in vm

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -387,8 +387,18 @@ class Request(RESTEntity):
 
     def _handleAssignmentStateTransition(self, workload, request_args, dn):
         
+        req_status = request_args["RequestStatus"]
+        
+        if not request_args.get('Team', '').strip() and \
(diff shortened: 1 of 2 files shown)


src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -387,8 +387,18 @@ def _handleNoStatusUpdate(self, workload, request_args):
 
     def _handleAssignmentStateTransition(self, workload, request_args, dn):
         
+        req_status = request_args["RequestStatus"]
+        
+        if not request_args.get('Team', '').strip() and \
+           not request_args.get('Teamd', []) and req_status == "assigned":
```
amaltaro: Typo here in Teamd

amaltaro: Nevermind, I guess this arg comes from the web assignment, no?

ticoann: Thanks I will fix that. Yes args get from web interface and I set only for Team not Teams so we should be fine for now.

PR #6676, opened 2016-03-07
Don.t expand job classAds into a plain list; remove default values

amaltaro:
And I also replaced the classAd "MachineAttrGLIDEIN_CMSSite0" by "MATCH_EXP_JOBGLIDEIN_CMSSite", which is what we use in the Condor plugin and sounds more familiar to everyone.

@hufnagel was there any motivation for keeping default values for what is retrieved from condor?
@ticoann I still need to run tests on this PR

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -21,13 +21,13 @@ import fnmatch
 
 import WMCore.Algorithms.BasicAlgos as BasicAlgos
 
-from WMCore.Credential.Proxy           import Proxy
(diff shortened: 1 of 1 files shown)


hufnagel:
I have no idea what you are asking me about here ? Did I write the previous code ?

amaltaro:
I thought you could know about it since you are the main PyCondor user (and also made a few fixes to it).
No problem, I'm patching now the testbed agent and we can see whether it will work.

amaltaro:
I had to put back some of those default values, otherwise the job tracking was failing.

amaltaro:
test this please

amaltaro:
Good to be merged

PR #6678, opened 2016-03-07
Locally update the PNNs location when reusing the pileup cache file

amaltaro:
This fix is mandatory for the new release. When we use TrustSitelists AND the dataset was already available in the cache, it was not getting the right PNN location for the blocks. So, in this case, we copy
the cache to the new sandbox location and update (actually, just add the site white list) the pileupconf.json for that request.

@ticoann please review but do not merge yet


amaltaro:
Enough with tests for this. Tested and works Ok.
@ticoann please merge it and the other 2 requests (6676 and 6682), then cut a new 1.0.14 tag and also create a new branch 1.0.14_wmagent out of it.

PR #6680, opened 2016-03-07
add api to get team name

ticoann:
add list of team names for assigning the request.
team names are only determined by agents registered to wmstats. (Not sure this will cause the problem).
Otherwise we have to have the interface to add the team name like in reqmgr1 
@vkuznet, @amaltaro, Please review

### DIFF
--- a/src/python/WMCore/Services/WMStats/WMStatsReader.py
+++ b/src/python/WMCore/Services/WMStats/WMStatsReader.py
@@ -248,19 +248,19 @@ class WMStatsReader(object):
         
         return result
     
-    def agentsByTeam(self, ignoreDrain = True):
+    def agentsByTeam(self, filterDrain=False):
         """
         return a dictionary like {team:#agents,...}
         """
         result = self._getAgentInfo()
         response = dict()
+
         for agentInfo in result["rows"]:
-            
             teams = agentInfo['value']['agent_team'].split(',')
             for team in teams:
-                if team not in response.keys():
-                    response[team] = 0
-            if ignoreDrain:
+                response.setdefault(team, 0)
+                
+            if filterDrain:
                 if not agentInfo['value']['drain_mode']:
                     for team in teams:
                         response[team] += 1
--- a/src/python/WMComponent/AgentStatusWatcher/ResourceControlUpdater.py
+++ b/src/python/WMComponent/AgentStatusWatcher/ResourceControlUpdater.py
@@ -121,7 +121,7 @@ class ResourceControlUpdater(BaseWorkerThread):
         """
         agentsByTeam = {}
         try:
-            agentsByTeam = self.centralCouchDBReader.agentsByTeam()
+            agentsByTeam = self.centralCouchDBReader.agentsByTeam(filterDrain=True)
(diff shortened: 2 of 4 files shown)


ticoann:
REST api will be (not deployed)
https://cmsweb.cern.ch/wmstatsserver/data/teams
Test vm is set
https://reqmgr2-dev.cern.ch/wmstatsserver/data/teams

vkuznet:
Seagnchan,
how this API will work? I mean does teams change over time or depends on
request? When should I query this API, every time when somebody
click on assign link, can we cache teams?

I think you really need to clarify the behavior for that.
So far it is just REST API which as far as I can tell will return
list of teams.

Thanks,
Valentin.

ticoann:
Hi Valentin,
In production teams are not changed, we very occasionally add teams but in general we have only a few teams and doesn't change. Although call is pretty light weight and repeated call won't have much overhead but I think caching would be fine. Are you trying to caching in browser? I think some of the browser caches ajax call automatically,  you can do manual caching if you want.

FYI. team is something that related to agent.  Basically team is used to pull down the request to certain agent.  So if a team is assigned to given request only agent belongs that team can pull that request.
We were planning to remove team concept completely, since request should run all the agent due to the not perfect priority and resource control we maintain a couple of teams for different agent.

When agent gets deployed, it registers (actually just reports) the team in wmstats so that is why I created wmstats REST api.

vkuznet:
Seangchan,
I changed https://github.com/dmwm/WMCore/pull/6681/
with respect to discussion we had. getTeams local API now fetch teams
all the time and fall backs into using cached values.
Please review, since I didn't test it on vm.
Best,
Valentin.

src/python/WMCore/Services/WMStats/WMStatsReader.py
```diff
@@ -248,19 +248,19 @@ def _getAgentInfo(self):
         
         return result
     
-    def agentsByTeam(self, ignoreDrain = True):
+    def agentsByTeam(self, filterDrain=True):
```
amaltaro: I'd default it to False instead.

src/python/WMCore/Services/WMStats/WMStatsReader.py
```diff
@@ -248,19 +248,19 @@ def _getAgentInfo(self):
...
         """
         return a dictionary like {team:#agents,...}
         """
         result = self._getAgentInfo()
         response = dict()
+        import pprint
```
amaltaro: you forgot this pprint here :)

src/python/WMCore/Services/WMStats/WMStatsReader.py
ticoann: Actually, in initial usage, we filtered out drain agent for calculating the resource distribution in AgentWatcher component. I can change those if it makes more sense including drain agent for team.

src/python/WMCore/Services/WMStats/WMStatsReader.py
ticoann: Thanks, updated

PR #6681, opened 2016-03-08
Changes to support team in assign interface

vkuznet:


src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -211,6 +211,27 @@ def __init__(self, app, config, mount):
         identifier = cdict.get("log_reporter", "reqmgr2")
         self.logdb = LogDB(centralurl, identifier)
 
+        # local team cache which will request data from wmstats
+        self.wmstatsurl = config.reqmgr.wmstats_url
```
amaltaro: How about keeping the same style as in the previous lines, e.g.:
self.wmstatsurl = cdict.get('wmstats_url', '')

I'd also get rid of the if in the line below, we simply assume it will be in the configuration file.
Otherwise, we better than check whether all the previous URLs are also available in the config file.

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -211,6 +211,27 @@ def __init__(self, app, config, mount):
...
+        if  not self.wmstatsurl:
+            raise Exception('ReqMgr2 configuration file does not provide wmstats url')
+        self.team_update_interval = 3600 # interval we update team cache
+        if  hasattr(config.reqmgr, 'team_update_interval'):
+            self.team_update_interval = int(config.reqmgr.team_update_interval)
```
amaltaro: and these 3 lines could be replaced by:
`self.team_update_interval = cdict.get('team_update_interval', 3600)`

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
vkuznet: Alan,
the problem is that config is not a pure python dict. It is WMCore object.
I didn't inspect if it supports get method and I provide a safe way
to fetch values from config. Of course if config would be a python
dict I'll use the way you described.
Valentin.

amaltaro: Well, just look at the previous 10 lines and you'll guarantee this config object supports get method. If it does not, then the whole code is doomed to fail :-)
Sorry, I'm all in favor of keeping the code consistent.

ticoann:
Valentin, it doesn't seems to be working. code is updated in reqmgr2-dev. Not sure why. I will check tomorrow

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -211,6 +212,27 @@ def __init__(self, app, config, mount):
...
+        teams = self.team_cache
+        try:
+            url = '%s/data/teams' % self.wmstatsurl
+            params = None
+            headers = {'Content-type':'application/json'}
+            data = getdata(url, params, headers)
+            teams = data.get('result', [])
+            self.team_cache = teams
+        except:
+            pass
```
ticoann: May be we should pass some warning messages here

vkuznet: Seangchan,
I fixed issue with fetching teams content and added warning printouts.
Please review. The reqmgr2-dev is working now.
Valentin.

ticoann:
Thanks Valentin, it is working. I will add another patch adding wmstats url in config and close this one,

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -211,6 +212,33 @@ def __init__(self, app, config, mount):
...
+        params = {}
+        headers = {'Accept':'application/json'}
+        try:
+            data = getdata(url, params, headers)
+            if  'error' in data:
+                print("WARNING: fail to get teams from %s" % url)
+                print(data)
+            teams = data.get('result', [])
+            self.team_cache = teams
+        except Exception exp:
```
vkuznet: Ohh, sorry typo on my end.

ticoann:
@vkuznet, Valentin, I added wmstats_url in the config.
https://github.com/dmwm/deployment/pull/295/files

Noe 
self.wmstatsurl= config.reqmgr.wmstats_url

vkuznet:
ok, that looks good. I think we have everything now in place and you can merge
different PRs.

PR #6682, opened 2016-03-08
clean wmbs data only workflow is archived

### LINKED ISSUE #6673: announced workflows not getting archived
with the new agent 1.0.14.pre8, example:
amaltaro_ReRecoSkim_HG1603gbis_Validation_v3_160305_001049_2496

I still have to debug to see what's going on.

ticoann:
fixes #6673

### DIFF
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -257,14 +257,14 @@ class CleanCouchPoller(BaseWorkerThread):
             self.archiveSummaryAndPublishToDashBoard(finishedwfsWithLogCollectAndCleanUp)
             logging.info("%s workload summary is created" % len(finishedwfsWithLogCollectAndCleanUp))
             
+            logging.info("Cleaning up couch db")
+            self.cleanCouchDBAndChangeToArchiveStatus()
+            logging.info("Done: cleaning up couch db")
+            
             logging.info("Cleaning up wmsbs and disk")
             self.deleteWorkflowFromWMBSAndDisk()
             logging.info("Done: cleaning up wmsbs and disk")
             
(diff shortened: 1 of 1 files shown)


ticoann:
fixes #6673

ticoann:
@amaltaro, please review, still need to tested in vm.
The problem was there is a race condition, to get moved the request to archived condition it needs to get the finished request first. But wmbs db gets deleted before archived status. 
1. We could have pull the all requests ran for that agent and compare and check whether it is exist in that agent,  
2. Or we delete the db only after it is archived. 

I did 2nd option since it is simpler change but maybe we should do 1st option for cleaner solution and clean the wmbs as soon as possible.
What do you think?

I will test this.
This need to be tested on the new request if requests are already deleted the patch doesn't work.

ticoann:
Actually, I think I have to go with option 1. It seems it will cause the problem updating multiple workload summary

amaltaro:
I agree, option 1 sounds better. The sooner we clean up the agent/db, the better.
BTW, you can patch vocms009 and test this fix there, I have workflows running and I'll run more in the coming days.

ticoann:
Hi Alan, test went OK with this patch. I looked into option 1. The problem with that is we don't have efficient view to get requests in a certain status and agent. We could get that information using current view but it would be a little complicated and not very efficient. 
Since sending multiple workload summary won't be critical problem, I would suggest to use this patch until next cmsweb deployment and I can create a new patch for next deployment. What do you think?

amaltaro:
Seangchan, sounds good to me. This data cleanup is getting a bit confusing for me :(, maybe we should first put transitions and their dependencies on a paper and then we change the code accordingly.

ticoann:
Thank Alan, that is good idea. i will do that. I will update that has to occur between the transactions.
I will create a new tag for agent with two other patches you created

PR #6686, opened 2016-03-09
Fix priority update in reqmgr2

amaltaro:
Without this fix, it goes through the args validation for workflow assignment, which clearly fails then.
@ticoann I tested in my VM and the workload_cache doc gets updated, however the workload config is not. Is the priority from the workload used anywhere in the end?

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -14,7 +14,7 @@ from WMCore.WMFactory import WMFactory
 from WMCore.ReqMgr.Auth import getWritePermission
 from WMCore.ReqMgr.DataStructs.Request import initialize_request_args
 from WMCore.ReqMgr.DataStructs.RequestStatus import check_allowed_transition
-from WMCore.ReqMgr.DataStructs.RequestError import InvalidStateTransition
+from WMCore.ReqMgr.DataStructs.RequestError import InvalidStateTransition, InvalidSpecParameterValue
 from WMCore.ReqMgr.Tools.cms import releases, architectures
 
 def loadRequestSchema(workload, requestSchema):
@@ -63,7 +63,7 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
     param and safe structure is RESTArgs structure: named tuple
     RESTArgs(args=[], kwargs={})
     
-    validate post request
+    validate post/put request
     1. read data from body
     2. validate the permission (authentication)
     3. validate state transition (against previous state from couchdb)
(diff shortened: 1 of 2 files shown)


amaltaro:
and forcing it to be an integer

ticoann:
I don't think it is used in this case. When priority changes JobUpdater change the priority in WMBS.
So it is not looking at the spec, With this patch it would only change the spec in central couchdb not the one downloaded in wmagent. Does this causing any problem?

amaltaro:
Yes, I think you are right! Nope, it shouldn't cause any problem.
I'm just pushing pylint fixes and then you can merge it :)

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -100,40 +103,46 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
     else:
         args_without_status = request_args
 
-    if len(args_without_status) > 0 and not workqueue_stat_validation(args_without_status):
+    if len(args_without_status) == 1 and 'RequestPriority' in args_without_status:
+        if not isinstance(args_without_status['RequestPriority'], int):
```
ticoann: Could we just try to convert and if gives value error then through the exception? Is it passing string?

ticoann:
Thanks Alan, I will merge this after creating the  new tag since this contains reqmgr changes I want to have consistency with what we have in production now

amaltaro:
Yes, I was considering cast it to int as well. Ok, I will update it, but first... bad news, GQ uses the priority from the spec file, just noticed it in my test... I'll fix it later on, anyways we are not in hurry with this one :-)

ticoann:
Thanks Alan, could you point me the code GQ uses priority from spec. I think I would like to change that to use from couch not spec if possible.

amaltaro:
This seems to be the GQ code:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/StartPolicyInterface.py#L115

ticoann:
Ah, you are right but for the GQ spec I think we are OK,
Following call 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L375

calls this line which updates GQ spec priority.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WorkQueue/WorkQueue.py#L143

However I think GQ has its own spec copied, so I would like to point that to reqmgr spec. But this could be dealt in a separate issue.

amaltaro:
Ok, I'll leave the change of spec url for another PR then.
Updated the patch, and tested in my VM. I also added the range validation to keep it in sync with StdBase.

PR #6692, opened 2016-03-11
fix site restriction checking

ticoann:
Not sure whether this was always problem or just happened. 
This need to be applied all the agents. I will create the patch for 1.0.12_wmagent and 1.0.14_wmagent

### DIFF
--- a/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
+++ b/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
@@ -213,8 +213,10 @@ class WorkQueueElement(dict):
     def passesSiteRestriction(self, site):
         """Takes account of white & black list, and input data to work out
         if site can run the work"""
-        # Don't check anything if trustSitelists is enabled
-        if self['NoLocationUpdate']:
+        # Don't check anything if trustSitelists is enabled and SiteWhitelist is empty or site is 
+        # is site whitelist
+        if self['NoLocationUpdate'] and \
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan please review

amaltaro:
Seangchan, I didn't get the problem. Is it an issue when there is an empty site white list or what?

ticoann:
Hi Alan, The issue is when site white list is not empty and TrustSites flag is set to True

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/Block.py#L57

In that case all the site is in the resource control  pass the restriction instead of only sites in the white list, so it pulls down many more workqueue elements than it supposed to be.

As far as code goes I think we always had this problem (although there was code change here this release) I guess that was some times we so there are too many jobs to pulled in one agent. (Not completely sure though).

Anyway, I ran the test on cmssrv production machine which Dave was running the step chain test. (That was the reason I looked at the problem). If you think it is OK, I will update to all production machine.

amaltaro:
That was a great catch, Seangchan! Yeah, fix sounds good! Would you mind creating the same fix for 1.0.14_wmagent and 1.0.12_wmagent branches? For 1.0.14, don't cut a new tag yet, I will have two more fixes for Monday or Tuesday. Thanks!

ticoann:
Thanks for the review Alan, I create the branch patch and I applied all the agents except draining agents (We will move to new version right). Updated the twiki as well. Don't work on weekend. :smile:

PR #6693, opened 2016-03-11
fix site restriction filter

ticoann:
1.0.12 version of #6692, applied to cmssrv271, cmssrv273, cmssrv277. 
If everything is OK I will apply to all the agent.
Problem only occurs when TrustSite is True. Seems I like we had this problem for a long time.


ticoann:
@amaltaro, I will merge this, I will create a new tag .patch3. Although we probably won't deploy that version. Is that OK?

amaltaro:
Sounds good to me. Maybe don't even bother creating a patch4 tag

PR #6695, opened 2016-03-13
Solve the missing files mystery during stage out

amaltaro:
Listing the main changes here:
- print date and time just before an storage operation, for the most critical implementations;
- use storage transfer/connection timeout when available by the client (to avoid the trigger of Alarm)
- increase the waiting time to trigger Alarm for 2h per attempt (worst case is 10h for RFCP impl)
- tried to simplify a bit the stageout command for SRM flavors

@ticoann still needs to be tested, quite a bit of testing actually :(
@hufnagel a replay writing to castor/eosT0 would be interesting to have before it gets merged, though you may want to wait for a green light from my side (in case there is some stupid error from my side)


amaltaro:
BTW, I was scanning SITECONF to see what implementations we have out there, these are the results
```
amaltaro@alan-wmacloud:/data/amaltaro/siteconf $ cat alan 
    101 <command value="srmv2-lcg"/>
     45 <command value="srmv2"/>
     15 <command value="cp" />
     14 <command value="srm"/>
      5 <command value="gfal2"/>
      5 <command value="rfcp-CERN"/>
      3 <command value="rfcp"/>
      3 <command value="stageout-xrdcp-fnal"/>
      1 <command value="rfcp-2"/>
      1 <command value="rfcp-RAL"/>
      1 <command value="rfio"/>
      1 <command value="vandy" />
      1 <command value="xrdcp"/>
```

I didn't filter by local or fallback stage out. As we can see, there are still 14 sites using the super old "srm" implementation (though they may be T3 sites...)

src/python/WMCore/Storage/Backends/LCGImpl.py
```diff
@@ -105,7 +120,9 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
             result += "export LD_LIBRARY_PATH=`echo $LD_LIBRARY_PATH | sed -e 's+:*/cvmfs/cms.cern.ch/[^:]*++'g`\n"
 
             result += "echo Sourcing CVMFS UI setup script\n"
-            result += ". /cvmfs/grid.cern.ch/emi-ui-2.9.0-1_sl5v1/etc/profile.d/setup-cvmfs-ui.sh\n"
+            # this version dates back to Aug 2013. So, let's get the symlink to the latest one instead
+            #result += ". /cvmfs/grid.cern.ch/emi-ui-2.9.0-1_sl5v1/etc/profile.d/setup-cvmfs-ui.sh\n"
+            result += ". /cvmfs/grid.cern.ch/emi3ui-latest/etc/profile.d/setup-ui-example.sh\n"
```
amaltaro: It's worth to mention that we were using a quite old EMI client. Updated it to the latest one in cvmfs, though I'm always afraid of pointing to the latest available version...

hufnagel:
We can't do a replay to castor here at CERN because we don't use castor anymore (the rfcp-CERN stageout plugin is all xrdcp these days). If you want to test castor you would have to run a test workflow at one of the sites using the rfcp, rfcp-2, rfcp-RAL or rfio stageout plugins.

amaltaro:
If we don't use castor, then we don't need to bother testing it :)
Yes, I ran ~10k jobs in several sites exercising most of these plugins, e.g.:
```
    101 <command value="srmv2-lcg"/>		T1_DE_KIT	T1_ES_PIC
     45 <command value="srmv2"/>			T1_RU_JINR	T2_BR_SPRACE
     15 <command value="cp" />				T2_ES_IFCA	T2_US_Florida
     14 <command value="srm"/>				T2_CH_CSCS	
      5 <command value="gfal2"/>			T2_UK_SGrid_Bristol	T2_US_Purdue
      5 <command value="rfcp-CERN"/>		T2_CH_CERN
      3 <command value="rfcp"/>				T2_FR_IPHC	T2_TH_CUNSTDA
      3 <command value="stageout-xrdcp-fnal"/>	FNAL
```

Tests seems to be Ok and we shall merge/backport this fix by the end of the day, though I still need to look carefully at the outcome of my tests.

ticoann:
Thanks Alan, Since I don't know much about command changes that I could review, since test is OK and Dirk can review, that will be good for me.  This is great that you fixed it.

amaltaro:
If the Alarm was the real responsible for these missing files, then yes, it should be fixed.
But don't merge it, I want to make sure it works for all those protocols I listed above, some of my tests are kind of stuck :(

ticoann:
Alan, this is ready to be merged, right?

amaltaro:
Yes, please.

PR #6696, opened 2016-03-13
JobStartDate may be eventually missing in the classAd

amaltaro:
The component crashed a couple of days ago, because that ad was missing in the Ads.
At some point we should review this plugin, it has way too many possible states and assumptions flying over the code :(

@ticoann don't merge yet, I wan't to make sure no more crashes will show up.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -167,10 +167,21 @@ class PyCondorPlugin(BasePlugin):
                        2: "Running",
                        3: "Removed",
                        4: "Complete",
-                       5: "Held",
-                       6: "Running"}
+                       5: "Held"}
 
         return exitCodeMap
+
+    @staticmethod
+    def logToScheddExitCodeMap(x):
+        """
+        JobStatus shows the last status of the job
+        Get TriggerEventTypeNumber which is the current status of the job
+        Map it back to Schedd Status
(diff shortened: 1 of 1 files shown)


amaltaro:
Did some rework on this code mainly to make it more "readable", some improvements and also fixed a bug in the updateSiteInformation method.

I don't really have time to look at the log stuff right now, but will try to come back to it in the coming weeks.

Needs test, don't merge please.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -343,6 +356,8 @@ def submit(self, jobs, info=None):
         # If we're here, then we have submitter components
         self.scriptFile = self.config.JobSubmitter.submitScript
         self.submitDir = self.config.JobSubmitter.submitDir
+        logging.info("ALAN self.submitDir %s and self.config.JobSubmitter.submitDir %s", self.submitDir,
```
amaltaro: hmm, and of course I need to remove this line before it gets merged :)

ticoann: I think you can keep it. :smile:

amaltaro: LoL. I couldn't find where we set this var (and it's not in the config), that's why I wanted to check on it :-)

amaltaro:
@ticoann good to go

PR #6700, opened 2016-03-14
return list of dictionary with common structure when common_dict=1

ticoann:
all reqmgr2 get call have common_dict param. When it is set to 1 it returns list of dictionary instead of dictionary with request name(RequestName) as a key.

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -20,6 +20,9 @@ import cherrypy
 from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_START_STATE
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
 
+#TODO: this need to be update from DB instead of hard coding for more flexibility
+RESULT_MASK_FOR_DAS = ["RequestName", "PrepID", "CMSSWVersion", "GlobalTag"]
+
 def initialize_request_args(request, config, clone = False):
     """
     Request data class request is a dictionary representing
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -16,12 +16,12 @@ from WMCore.REST.Validation import validate_str
 from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 
 import WMCore.ReqMgr.Service.RegExp as rx
-from WMCore.ReqMgr.DataStructs.Request import initialize_request_args
-from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_STATE_LIST
-from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_STATE_TRANSITION
+from WMCore.ReqMgr.DataStructs.Request import initialize_request_args, RESULT_MASK_FOR_DAS
+from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_STATE_LIST,\
+    REQUEST_STATE_TRANSITION, ACTIVE_STATUS
 from WMCore.ReqMgr.DataStructs.RequestType import REQUEST_TYPES
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
-from WMCore.ReqMgr.Utils.Validation import validate_request_create_args, \
+from WMCore.ReqMgr.Utils.Validation import validate_request_create_args,\
     validate_request_update_args, loadRequestSchema
 
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
@@ -61,7 +61,7 @@ class Request(RESTEntity):
             param.args.pop()
             return
         
-        no_multi_key = ["detail", "_nostale", "date_range"]
+        no_multi_key = ["detail", "_nostale", "date_range", "common_dict"]
         for key, value in param.kwargs.items():
             # convert string to list
             if key not in no_multi_key and isinstance(value, basestring):
@@ -245,7 +245,53 @@ class Request(RESTEntity):
         workload = spec.factoryWorkloadConstruction(clone_args["RequestName"],
                                                     clone_args)
         return (workload, clone_args)
-
+    
+    def _maskTaskStepChain(self, masked_dict, req_dict, chain_name, mask_key):
+        
+        mask_exist = False
+        num_loop = req_dict["%sChain" % chain_name]
+        for i in range(num_loop):
+            if mask_key in req_dict["%s%s" % (chain_name, i+1)]:
+                mask_exist = True
+                break
+        if mask_exist: 
+            defaultValue = masked_dict[mask_key]
+            masked_dict[mask_key] = []
+            # assume mask_key is list if the condition doesn't meet.
+            
+            for i in range(num_loop):
+                chain = req_dict["%s%s" % (chain_name, i+1)]
+                if mask_key in chain:
+                    chain_key = "%sName" % chain_name
+                    masked_dict[mask_key].append({chain_key: chain[chain_key], mask_key: chain[mask_key]})
+                else:
+                    if isinstance(defaultValue, dict):
+                        value = defaultValue.get(chain_key, None)
+                    else:
+                        value = defaultValue
+                    masked_dict[mask_key].append({chain_key: chain[chain_key], mask_key: chain[mask_key]})
+        return
+    
+    def _mask_result(self, mask, result):
+        
+        if len(mask) == 1 and mask[0] == "DAS":
+            mask = RESULT_MASK_FOR_DAS
+        
+        if len(mask) > 0:
+            masked_result = {}
+            for req_name, req_info in result.items():
+                masked_result.setdefault(req_name, {})
+                for mask_key in mask:
+                    masked_result[req_name].update({mask_key: req_info.get(mask_key, None)})
+                    if "TaskChain" in req_info:
+                        self._maskTaskStepChain(masked_result[req_name], req_info, "Task", mask_key)
+                    elif "StepChain" in req_info:
+                        self._maskTaskStepChain(masked_result[req_name], req_info,"Step", mask_key)
+                        
+            return masked_result
+        else:
+            return result
+    
     @restcall(formats=[('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
     def get(self, **kwargs):
         """
@@ -255,7 +301,7 @@ class Request(RESTEntity):
         If jobInfo is True, returns jobInfomation about the request as well.
 
         TODO:
-        stuff like this has to filtered out from result of this call:
+        stuff like this has to masked out from result of this call:
             _attachments: {u'spec': {u'stub': True, u'length': 51712, u'revpos': 2, u'content_type': u'application/json'}}
             _id: maxa_RequestString-OVERRIDE-ME_130621_174227_9225
             _rev: 4-c6ceb2737793aaeac3f1cdf591593da4
@@ -275,7 +321,11 @@ class Request(RESTEntity):
         team = kwargs.get("team", [])
         mc_pileup = kwargs.get("mc_pileup", [])
         data_pileup = kwargs.get("data_pileup", [])
+        mask = kwargs.get("mask", [])
         detail = kwargs.get("detail", True)
+        # set the return format. default format has requset name as a key
+        # if is set to one it returns list of dictionary with RequestName field.
+        common_dict = int(kwargs.get("common_dict", 0))
         if detail in (False, "false", "False"):
             option = {"include_docs": False}
         else:
@@ -286,7 +336,9 @@ class Request(RESTEntity):
             self.reqmgr_db_service._setNoStale()
 
         request_info = []
-
+        
+        if len(status) == 1 and status[0] == "ACTIVE":
+            status = ACTIVE_STATUS
         if status and not team and not request_type:
             request_info.append(self.reqmgr_db_service.getRequestByCouchView("bystatus", option, status))
         if status and team:
@@ -317,9 +369,20 @@ class Request(RESTEntity):
             request_info.append(self.reqmgr_db_service.getRequestByCouchView("bydatapileup", option, data_pileup))
         # get interaction of the request
         result = self._intersection_of_request_info(request_info)
+        
         if len(result) == 0:
             return []
-        return rows([result])
+        
+        result = self._mask_result(mask, result)
+        # If detail is set to False return just list of request name
+        if not option["include_docs"]:
+            return result.keys()
+        
+        if common_dict == 1:
+            response_list = result.values()
+        else:
+            response_list = [result] 
+        return rows(response_list)
 
     def _intersection_of_request_info(self, request_info):
         requests = {}
--- a/src/python/WMCore/WMStats/Service/ActiveRequestJobInfo.py
+++ b/src/python/WMCore/WMStats/Service/ActiveRequestJobInfo.py
@@ -7,7 +7,7 @@ from WMCore.REST.Server import RESTEntity, restcall, rows
 from WMCore.REST.Tools import tools
 from WMCore.WMStats.DataStructs.DataCache import DataCache
 
-from WMCore.REST.Format import JSONFormat
+from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 
 class ActiveRequestJobInfo(RESTEntity):
     """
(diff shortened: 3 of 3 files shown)


ticoann:
https://reqmgr2-dev.cern.ch/reqmgr2/data/request?outputdataset=/Pythia8_Dijet15_pp502_TuneCUETP8M1/Integ_Test-MonteCarlo_Ext_reqmgr1_test-v1/GEN-SIM&common_dict=1

Previous key is under RequestName field.

ticoann:
@vkuznet, Valentin, please review.

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -319,7 +322,12 @@ def get(self, **kwargs):
         result = self._intersection_of_request_info(request_info)
         if len(result) == 0:
             return []
-        return rows([result])
+        
+        if common_dict == 1:
+            response_list = result.values()
```
vkuznet: Does result values guaranteed to have identical structure? What about request names, they will not appear in result.values() since they're the keys of result dict

ticoann: It doesn't have identical structure by nature. i.e. TaskChain has different structure than MC. If you need exact same structure, we first need a survey what information they need which has common for all the request and send only those.  RequestName field is in the value with current api.

vkuznet: Well, if there is no identical structure DAS cannot handle it efficiently.
Moreover if structure is unknown it means that we can't write efficient
parser either. This is a big drawback to deal with unstructured data.
As you know searching, parsing of data who change their structure
requires a lot of effort, both CPU and RAM wise.

ticoann: Yes, we need to know what exact information People need by inputdatasets, outputdatasets, etc. We can put some minimal information first as Alan suggested. If users need more we can add accordingly.

ticoann:
@vkuznet,
Hi Valentin, I updated the patch so return structure is controlled by requestor. filter query will contol the return structure.

https://reqmgr2-dev.cern.ch/reqmgr2/data/request?outputdataset=/Pythia8_Dijet15_pp502_TuneCUETP8M1/Integ_Test-MonteCarlo_Ext_reqmgr1_test-v1/GEN-SIM&common_dict=1&filter=RequestName&filter=EventsPerLumi

Will return,

{"result": [
 {
  "RequestName": "sryu_MonteCarlo_Ext_Sryu_reqmgr1_160304_080525_8176", 
  "EventsPerLumi": 100
}]}

vkuznet:
Seangchan,
even though this particular URL works, this creates another 2 problems:

1. client should to figure out somehow which fields are available to query.
I think it can be resolved if you'll provide API to look-up all filter fields, e.g.
https://reqmgr2-dev.cern.ch/reqmgr2/data/filters will return list of acceptable
filters.

2. with large list of filter parameters GET request can easily reach its limit,
whatever it is, and so far I can't query via POST, see below. We should be able
to pass a list of parameters in a request, rather specify each filter individually
as you provided in your example.

```
scurl -X POST -H "Content-type:application/json" -H "Accept:application/json" -d '{"outputdataset":"/Pythia8_Dijet15_pp502_TuneCUETP8M1/Integ_Test-MonteCarlo_Ext_reqmgr1_test-v1/GEN-SIM", "common_dict":1, "filter":["RequestName","EventsPerLumi"]}' "https://reqmgr2-dev.cern.ch/reqmgr2/data/request"
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"></meta>
    <title>400 Bad Request</title>
    <style type="text/css">
    #powered_by {
        margin-top: 20px;
        border-top: 2px solid black;
        font-style: italic;
    }

    #traceback {
        color: red;
    }
    </style>
</head>
    <body>
        <h2>400 Bad Request</h2>
        <p>Invalid spec parameter value: RequestString need to be specified</p>
        <pre id="traceback"></pre>
    <div id="powered_by">
    <span>Powered by <a href="http://www.cherrypy.org">CherryPy 3.2.2</a></span>
    </div>
    </body>
</html>
```

ticoann:
Hi Valentin,

For 1.  We can provide list of keys they can filter.  But I don't mean DAS will support flexible query from a large set of filter. I think users need to specify what they want ahead. If they want to know what information need to get providing inputdataset. They have to spell it out before this gets to add DAS.
Or if you want to get DAS more flexible. We can give the minimal set of filter with the api your suggested. But still the number of filters they can search need to be controlled, not give them all possible numbers and they can choose among them. If user really wants to do that they can directly query reqmgr2.

For 2. post method for the API is currently used by inserting the data. We could provide post call later if necessary. But for the filter, I would like to see what information they need to extract first. If we end up needing bunch of filters. We can send those information by default or one separate flag.

vkuznet:
Seangchan,
the filters you introduced is posterior information. Users can't know a-priory
what is available unless you tell them though. DAS will not manipulate filters,
I'll put them into DAS maps and that what DAS will return all the time for that
or other APIs. Users will get records with this info.

How you gonna handle it is up to you and your best shot is to provide
minimum information and add gradually. Nobody will provide you any response,
as we learnt from many times everybody is too busy. But people will speak
if they miss something only when they query something, not another way around.

ticoann:
Actually, that is my point. I am not sure why initially it was needed and I wonder after you removed DAS map for this due to the performance problem, did any one complain about the information they can't find?
If they don't need the information, we don't need to provide. When they complain about missing information, we will provide that information. We can go with minimal set we might think useful (maybe 0 initially :smile:), but I don't think we need to start with all possible information first.

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -20,6 +20,9 @@
 from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_START_STATE
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
 
+#TODO: this need to be update from DB instead of hard coding for more flexibility
+RESULT_FILTER_FOR_DAS = ["RequestName", "OutputDatasets", "InputDatasets", "PrepID"]
```
ticoann: This the data will be returned when filter=DAS is set. @amaltaro, Alan please review what information we should add for this. I have no idea. :smile:

ticoann:
Hi Valentin,
I hard coded what information to return when filter=DAS is set. I currently have these four keys but I will consult with Alan what to add.

["RequestName", "OutputDatasets", "InputDatasets", "PrepID"]

https://reqmgr2-dev.cern.ch/reqmgr2/data/request?status=completed&request_type=MonteCarlo&common_dict=1&filter=DAS

returns
```
{"result": [
 {
  "InputDatasets": [], 
  "PrepID": "SUS-Summer12pLHE-00144", 
  "RequestName": "sryu_MonteCarlo_LHE_Sryu_reqmgr1_160304_080530_5998", 
  "OutputDatasets": [
    "/SMS-T5wg_2J_mGl-1550to1600_mLSP-25to1575_TuneZ2star_8TeV-madgraph-tauola/Integ_Test-MonteCarlo_LHE_reqmgr1_test-v1/GEN"
  ]
}]}
```

src/python/WMCore/ReqMgr/DataStructs/Request.py
amaltaro: I'd add the following three keys: Campaign, GlobalTag and CMSSWVersion. I'd say we should also add the ConfigCacheID (unless DAS is somehow fetching the configuration that produced that dataset already?), but if you prefer, this can wait until people explicitly ask for this information.

BTW, three questions:
a) why InputDatasets (plural)? Is it for TaskChain cases? Or it adds the Pileup sample?
b) giving a second thought, OutputDatasets sounds useless, since the user using DAS will know already which dataset or block or file he wants to look at
c) PrepID, GlobalTag and CMSSWVersion can be different for TaskChain (and StepChain) requests, are you considering it?

ticoann: > a) why InputDatasets (plural)? Is it for TaskChain cases? Or it adds the Pileup sample?

It is just the key, initially we are supposed to support multiple inputdatasets so that is why we have Inputdatasets altough we have only one elements,

> b) giving a second thought, OutputDatasets sounds useless, since the user using DAS will know already which dataset or block or file he wants to look at

I will remove that,

> c) PrepID, GlobalTag and CMSSWVersion can be different for TaskChain (and StepChain) requests, are you considering it?

I think DAS has different ways to get the ConfigID and PrepID.
 None of Taskchain property are considered here. Only top level. If we add TaskChain properties I will add the list instead of string.

ticoann:
@amaltaro, I updated patch as you suggested. TaskChain and StepChain are handled duplicate values allowed for that which Task has the value in order (Task1, Task2, ...)

https://reqmgr2-dev.cern.ch/reqmgr2/data/request?status=ACTIVE&filter=DAS&request_type=TaskChain
{"result": [
 {
  "PrepID": [
    "task_EXO-RunIIFall15DR76-03799", 
    "EXO-RunIIFall15DR76-03799", 
    "EXO-RunIIFall15DR76-03799", 
    "EXO-RunIIFall15MiniAODv1-02471"
  ], 
  "RequestName": "sryu_McMTaskChain2_Sryu_reqmgr2_160304_080733_673", 
  "GlobalTag": [
    "76X_mcRun2_asymptotic_v12", 
    "76X_mcRun2_asymptotic_v12", 
    "76X_mcRun2_asymptotic_v12", 
    "76X_mcRun2_asymptotic_v12"
  ], 
  "InputDatasets": null, 
  "Campaign": "reqmgr2-validation", 
  "CMSSWVersion": [
    "CMSSW_7_6_1", 
    "CMSSW_7_6_1", 
    "CMSSW_7_6_1", 
    "CMSSW_7_6_1"
  ]
}]}

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -275,7 +315,11 @@ def get(self, **kwargs):
         team = kwargs.get("team", [])
         mc_pileup = kwargs.get("mc_pileup", [])
         data_pileup = kwargs.get("data_pileup", [])
+        filter = kwargs.get("filter", [])
```
amaltaro: You are redefining the filter builtin, better picking another name.

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -20,6 +20,9 @@
 from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_START_STATE
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
 
+#TODO: this need to be update from DB instead of hard coding for more flexibility
+RESULT_MASK_FOR_DAS = ["RequestName", "InputDatasets", "PrepID", "Campaign"]
+
```
ticoann: @amaltaro, Alan could you advice what needs to be included here?

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: changed to mask

src/python/WMCore/ReqMgr/DataStructs/Request.py
amaltaro: Sure, by scanning one workload_cache, I'd say we need to provide to DAS:
["RequestName", "PrepID", "CMSSWVersion", "GlobalTag"]

not sure about (since people can look at reqmgr to get this info):
["MCPileup", "Memory", "ScramArch", "TimePerEvent", "SizePerEvent"]

I removed "Campaign" because it's encoded in the dataset name (which should be the same as AcquisitionEra in the end).

As I said during the meeting, some fields don't need to be a list (should not). But we need to properly find these keys in a TaskChain or StepChain request:
PrepID, CMSSWVersion, GlobalTag (and Memory, ScramArch, TimePerEvent, SizePerEvent)

amaltaro:
And as we discussed during the meeting, the use case for DAS should be always the same, as:
 - user provides a dataset as input, and ReqMgr has to provide information regarding THIS dataset, not the workflow itself. The keys were already discussed. What is not provided by ReqMgr2, is provided by PhEDEx or DBS, so it should be Ok.

src/python/WMCore/ReqMgr/DataStructs/Request.py
ticoann: I will start with minimum. Actually, I didn't get about those property should not be a list. 
If the prepID is different for each Task/Step, how would you organize?

amaltaro:
> I will start with minimum. Actually, I didn't get about those property should not be a list.
> If the prepID is different for each Task/Step, how would you organize?

We have two use cases here:
 a) a DAS query: in this case the user knows exactly which dataset he wants to look at, so we do not need to provide a list of PrepID, that will only confuse him and not give a real answer. For DAS case, we need to fetch the correct PrepID, GlobalTag and etc
 b) general case (anyone asking reqmgr2 for some information, a request information, etc): in this case, there is no dataset in the call (usually), so the user does not want info for a specific dataset, but instead for a whole request and thus we need some info to be in a list format.

If you prefer we can discuss over vidyo today

ticoann:
Thanks to Alan, I now understand more about DAS and general uses cases. However, current reqmgr document doesn't specify which task creates which outputdataset. OutputDataset is the just list outside the Task. We need to create the map in reqmgr document in the request document.

Right now only thing I can do from quest which passes output dataset. is following

https://reqmgr2-dev.cern.ch/reqmgr2/data/request?outputdataset=/ZprimeToA0hToA0chichihbb_2HDM_MZp-2500_MA0-300_13TeV-madgraph/RunIIFall15DR76-PU25nsData2015v1_76X_mcRun2_asymptotic_v12-v0/AODSIM&request_type=TaskChain&mask=DAS&common_dict=1

returns 

```
{"result": [
 {
  "PrepID": [
    {
      "TaskName": "EXO-RunIIFall15DR76-03799_0", 
      "PrepID": "EXO-RunIIFall15DR76-03799"
    }, 
    {
      "TaskName": "EXO-RunIIFall15DR76-03799_1", 
      "PrepID": "EXO-RunIIFall15DR76-03799"
    }, 
    {
      "TaskName": "EXO-RunIIFall15MiniAODv1-02471_0", 
      "PrepID": "EXO-RunIIFall15MiniAODv1-02471"
    }
  ], 
  "RequestName": "sryu_McMTaskChain2_Sryu_reqmgr2_160318_050204_3263", 
  "GlobalTag": [
    {
      "TaskName": "EXO-RunIIFall15DR76-03799_0", 
      "GlobalTag": "76X_mcRun2_asymptotic_v12"
    }, 
    {
      "TaskName": "EXO-RunIIFall15DR76-03799_1", 
      "GlobalTag": "76X_mcRun2_asymptotic_v12"
    }, 
    {
      "TaskName": "EXO-RunIIFall15MiniAODv1-02471_0", 
      "GlobalTag": "76X_mcRun2_asymptotic_v12"
    }
  ], 
  "CMSSWVersion": [
    {
      "TaskName": "EXO-RunIIFall15DR76-03799_0", 
      "CMSSWVersion": "CMSSW_7_6_1"
    }, 
    {
      "TaskName": "EXO-RunIIFall15DR76-03799_1", 
      "CMSSWVersion": "CMSSW_7_6_1"
    }, 
    {
      "TaskName": "EXO-RunIIFall15MiniAODv1-02471_0", 
      "CMSSWVersion": "CMSSW_7_6_1"
    }
  ]
}]}
```
If we need, we can add the task and output dataset map in reqmgr document but we have update all old document as well.

@amaltaro, Alan, is there any other way we can map between task property and output dataset name?

ticoann:
@vkuznet, Valentine what is the exact reqmgr2 query in DAS which caused the problem?

vkuznet:
Andrew Levin asked for this dataset
/SingleMu/Run2012D-ZMu-PromptSkim-v1/RAW-RECO

DAS had a call to ReqMgr2 inputdataset
https://cmsweb.cern.ch/reqmgr2/data/request?inputdataset=/SingleMu/Run2012D-ZMu-PromptSkim-v1/RAW-RECO

and ReqMgr2 returns large JSON which DAS struggles to parse in real time on web
interface.

ticoann:
Thanks Valentin, I thing we have to remove inputdataset call from DAS. DAS it doesn't distinguish whether dataset is fed as inputdataset or outputdataset right?
The same dataset can be either input or output dataset.
And the result it returns are different depending on whether it is input dataset or output dataset
I think it is more meaningful if das only queries output dataset.

https://cmsweb.cern.ch/reqmgr2/data/request?outputdataset=/SingleMu/Run2012D-ZMu-PromptSkim-v1/RAW-RECO

Anyway, we will still restrict the return values as we discussed here but I have to solve the case with Task and Step chain.

Thanks

vkuznet:
Seangchan,
this is what I did to solve the problem, I remove inputdataset API call from
DAS and right now DAS only query outputdataset API.
V.

ticoann:
Yes, I know. I am just saying we shouldn't put it back even if we make this call small and consistent (no arbitrary key). You still want consistent dictionary format for output dataset result, right?

vkuznet:
yes

ticoann:
merging this. If need to change the return format will do next iteration.

PR #6705, opened 2016-03-15
Avoid crashing ErrorHandler in case of Couch/HTTP issues

amaltaro:
I'm running this patch in submit1 and submit2 (well, the exception handling part only), since they were crashing all the time.

@ticoann please review and if that's Ok, I'll create two more PRs for 1.0.14 and 1.0.12 branches.

### DIFF
--- a/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
+++ b/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
@@ -1,6 +1,4 @@
 #!/usr/bin/env python
-#pylint: disable=W0613, W6501
-# W6501: It doesn't like string formatting in logging messages
 """
 The actual error handler algorithm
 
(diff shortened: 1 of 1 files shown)


ticoann:
Hi Alan, patch looks good. I was just wondering whether we can catch more specific errors than HTTPException. i.e. if check the status of the exception etc? If it can't be, go ahead create the patch for other version. Thanks,

amaltaro:
I could make another except block for HTTPException only and print the status and reason for that case, though I'm not sure whether we should take different actions for different status (and which action) :(

Let's see, if jenkins complain about something else, than I could change this PR according, otherwise I'll just backport it.

ticoann:
I was just concern about too catching general exception. (including other exceptions which might be better to make component crash), not about the print out messages. Anyway, current patch will be fine.

amaltaro:
And this master PR we don't merge it, I'll bypass some status code that come to my mind (400, 404, 502, 503) and raise HTTPException otherwise. I also want to do some pylint cleanup.

amaltaro:
Sorry, I'll update and test my changes by Monday.

amaltaro:
I removed all the sendAlert pieces from this component and also removed the outer transaction rollback, we already create them in the two inner methods.
Still need to be tested

amaltaro:
test this please

amaltaro:
@ticoann if unittests run Ok and the code looks Ok for you (I'm running this code in vocms074, but no exception happened so far...), then merge please.

I'll use the new tag for the new agent validation.

amaltaro:
test this please

PR #6713, opened 2016-03-16
Backporting again Solve the missing files mystery during stage out

amaltaro:
1.0.12 version of #6695


amaltaro:
@ticoann also good to go. There is no need for a new tag, I refuse to install this agent version again :)

PR #6714, opened 2016-03-17
add WMArchive related views and field

ticoann:
add the view for checking upload is done by given workflow.
add jobtype and jobstatus for fwjr


ticoann:
@vkuznet, @amaltaro, Please review

src/python/WMCore/Services/WMArchiver/DataMap.py
```diff
@@ -247,6 +252,8 @@ def createArchiverDoc(job_id, fwjr, version=None):
     newfwjr['meta_data'] = {'agent_ver': version,
                          'host': socket.gethostname().lower(),
                          'fwjr_id': job_id,
-                         'ts': int(time.time())
+                         'jobtype': jobtype,
```
vkuznet: @ticoann , Seangchan does it means we need to tweak WMArchive schema again and add jobtype and jobstate to it, see https://github.com/dmwm/WMArchive/blob/master/test/data/samples/fwjr_production.py#L4

ticoann: Yes, I will update new schema after running the test. With this patch I can extract all different kinds of fwjr by type and its status. I will change necessary schema changes and create the new pull request in WMArchive.

ticoann:
tested now in vm, merging it

PR #6715, opened 2016-03-17
Backport  1.0.14 add WMArchive related views and field

ticoann:
1.0.14_wmagent version of #6714


ticoann:
@amaltaro, Alan, this is not really tested but need to be added if we ever want to switch to put the data in WMArchiver. So I would like to test while we are doing new version of agent testing. Sorry about that.

PR #6721, opened 2016-03-21
Use dummy PNNs for PSNs without a PNN

### LINKED ISSUE #6433: populate resource table with PSN which doesn't have PNN
In first step use the fake PNN to populate wmbs_location_senames table.

i.e. T3_US_OSG doesn't have PNN

alexanderrichards:
This pull request resolves https://github.com/dmwm/WMCore/issues/6433 by adding into the resources table sites that have no PNN with a dummy entry of 'Site_Dummy'

### DIFF
--- a/src/python/WMCore/Services/SiteDB/SiteDB.py
+++ b/src/python/WMCore/Services/SiteDB/SiteDB.py
@@ -311,7 +311,7 @@ class SiteDBJSON(Service):
             pnns.update(pnn_list)
         return list(pnns)
 
-    def PSNtoPNNMap(self, psn_pattern=''):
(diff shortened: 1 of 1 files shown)


ticoann:
@alexanderrichards, Sorry Alex for late reply. As we discussed in the meeting. 
I was initially thought not updating wmbs_location_senames  if there is no PNN,  but will cause a lot of changes, current code set "None" as default value and I think we should leave as it is.

If you allow to have the option to have site name without PNN.  What we need to check is

1. find all the queries using inner join with wmbs_location_senames
2. and check whether the return pnn name is used. 

Or we can just set run the test with current code and see what is braking.

ticoann:
Not relavant anymore.

PR #6723, opened 2016-03-21
add special subscription support for skims

hufnagel:
When defining the subscriptions for workflow output, allow to define different subscriptions for skims. Needed by Tier0. Basically, skims can have the same data tier as regular workflow output but need different subscriptions. As such, the usual selection by primary dataset and data tier isn't enough.

@ticoann please review

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -897,6 +897,7 @@ class WMTaskHelper(TreeHelper):
                                    custodialSubType = "Replica", nonCustodialSubType = "Replica",
                                    custodialGroup = "DataOps", nonCustodialGroup = "DataOps",
                                    priority = "Low", primaryDataset = None,
+                                   useSkim = False, isSkim = False,
                                    dataTier = None, deleteFromSource = False):
         """
         _setSubscriptionsInformation_
@@ -926,9 +927,18 @@ class WMTaskHelper(TreeHelper):
             outputDataset = entry["outputDataset"]
             outputModule = entry["outputModule"]
 
-            primDs = outputDataset.split('/')[1]
-            tier = outputDataset.split('/')[3]
+            dsSplit = outputDataset.split('/')
+
+            primDs = dsSplit[1]
+
+            procDsSplit = dsSplit[2].split('-')
+            skim = ( len(procDsSplit) == 4 )
+
+            tier = dsSplit[3]
+
             if primaryDataset and primDs != primaryDataset:
+                continue
+            if useSkim and isSkim != skim:
                 continue
             if dataTier and tier != dataTier:
                 continue
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -1150,6 +1150,7 @@ class WMWorkloadHelper(PersistencyHelper):
                                             custodialSubType="Replica", nonCustodialSubType="Replica",
                                             custodialGroup="DataOps", nonCustodialGroup="DataOps",
                                             priority="Low", primaryDataset=None,
+                                            useSkim = False, isSkim = False,
                                             dataTier=None, deleteFromSource=False):
         """
         _setSubscriptionInformationWildCards_
@@ -1192,6 +1193,8 @@ class WMWorkloadHelper(PersistencyHelper):
                                         nonCustodialGroup=nonCustodialGroup,
                                         priority=priority,
                                         primaryDataset=primaryDataset,
+                                        useSkim=useSkim,
+                                        isSkim=isSkim,
                                         dataTier=dataTier,
                                         deleteFromSource=deleteFromSource)
 
@@ -1200,6 +1203,7 @@ class WMWorkloadHelper(PersistencyHelper):
                                    custodialSubType="Replica", nonCustodialSubType="Replica",
                                    custodialGroup="DataOps", nonCustodialGroup="DataOps",
                                    priority="Low", primaryDataset=None,
+                                   useSkim = False, isSkim = False,
                                    dataTier=None, deleteFromSource=False):
         """
         _setSubscriptionInformation_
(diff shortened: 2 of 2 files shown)


src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -926,10 +927,19 @@ def setSubscriptionInformation(self, custodialSites = None, nonCustodialSites =
...
             outputModule = entry["outputModule"]
 
-            primDs = outputDataset.split('/')[1]
-            tier = outputDataset.split('/')[3]
+            dsSplit = outputDataset.split('/')
+
+            primDs = dsSplit[1]
+
+            procDsSplit = dsSplit[2].split('-')
+            skim = ( len(procDsSplit) == 4 )
```
amaltaro: Why 4? Is AcquisitionEra supposed to be "Skim"? If so, then directly check against it, you have the number 4 hardcodded here anyways.
I also don't see the need for two flags,

amaltaro:
And where is the special subscription handling when isSkim/useSkim is True?

src/python/WMCore/WMSpec/WMTask.py
hufnagel: You need two flags. I could have done the same with a trinary operator but as that doesn't exist in python, two booleans was the best choice (thought briefly about using True, False and None but that would have been just super ugly and dangerous).

4 is the number of '-' separated fields in the processed datasets name. The 4th is the filterName and is only used for skims. So if you have 4, the dataset in question is a "skim" by some definiiton. Good enough for the Tier0 at least, as all Tier0 output strictly follows processed dataset naming conventions.

I need two booleans because I need to be able to set subscription info

a) where it completely ignores whether the dataset is a skim or not (same as before)

b) differently based on whether the dataset is a skim or not

hufnagel:
There is no special subscription handling. This is just a way to select when subscription info should be used and when not. For an example see

https://github.com/hufnagel/T0/blob/skim-subscription/src/python/T0/RunConfig/RunConfigAPI.py#L810

src/python/WMCore/WMSpec/WMTask.py
hufnagel: Basically a) provides backwards compatibility. If it weren't for a) you would have to make every setSubscriptionInfo call twice everywhere, once for skims and once for non-skims.

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -926,10 +927,19 @@ def setSubscriptionInformation(self, custodialSites = None, nonCustodialSites =
...
+
+            tier = dsSplit[3]
+
             if primaryDataset and primDs != primaryDataset:
                 continue
+            if useSkim and isSkim != skim:
```
hufnagel: This is the only line these two booleans are actually used.

src/python/WMCore/WMSpec/WMTask.py
amaltaro: > 4 is the number of '-' separated fields in the processed datasets name.

Dahhh, of course! I overlooked it.

The two flags are still not clear to me... how do you interact with them btw, don't you need to provide this parameter from the spec down to the workload and wmtask?

One more comment, pylint won't like this style:
`skim = ( len(procDsSplit) == 4 )`
it should be
`skim = len(procDsSplit) == 4`

hufnagel: That's a failure in pylint IMO, I find the former easier to read (as it more clearly separates the assignment and comparison operator.

Two flags because one is for "should I care if this is a skim or not", the second is for "now that I care, do you want me to configure for a skim or a non-skim".

And yes, the two flags propagate all the way down from the top, that's what most of this PR does. Tier0 calls setSubscriptionInfo on the workload. To be honest, I don't see a need for WMAgent to ever use this, your workflows are much simpler and don't need this.

hufnagel: But as useSkim has default False, you'll just use compatibility mode, no change.

hufnagel:
retest this please

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -897,6 +897,7 @@ def setSubscriptionInformation(self, custodialSites = None, nonCustodialSites =
                                    custodialSubType = "Replica", nonCustodialSubType = "Replica",
                                    custodialGroup = "DataOps", nonCustodialGroup = "DataOps",
                                    priority = "Low", primaryDataset = None,
+                                   useSkim = False, isSkim = False,
```
ticoann: This two parameter need to be moved to at the end. Following uses positional arguments
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1224
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1230

ticoann: Sorry, never mind. You already changed below.

src/python/WMCore/WMSpec/WMTask.py
ticoann: > 4 is the number of '-' separated fields in the processed datasets name. The 4th is the filterName and is only used for skims. So if you have 4, the dataset in question is a "skim" by some definiiton. Good enough for the Tier0 at least, as all Tier0 output strictly follows processed dataset naming conventions.

So for non Tier0 workflow,  useSkim is always False, right? 
Also, maybe it is good to have some lexicon check when Tier0 ProcDS is created so it checks the convention (if it doesn't have it yet)

hufnagel: Yes, only Tier0 should ever need to use the two new parameters. Everyone else will get the defaults. Only Tier0 workflows have these specific requirements about putting one type of output in one place and another type of output in another place.

Tier0 processed dataset names are constructed using standard WMCore code in StdBase. I would assume any kind of lexicon check is done there.

hufnagel:
retest this please

PR #6725, opened 2016-03-21
fix acdc name if it contains @

### LINKED ISSUE #6703: ReqMgr2: ACDC creation via webpage - naming needs to be fixed
If you don't change the request name during ACDC creation, it offers the same original request name without the timestamp part, such that ACDCing
pdmvserv_EXO-RunIISummer15GS-05060_00274_v2__160308_110629_7026

creates:
amaltaro_pdmvserv_EXO-RunIISummer15GS-05060_00274_v2__160315_102225_1898

so, just prepending the user name and appending the timestamp.

My suggestion is to keep the same behaviour as in reqmgr, so replacing the user name by "ACDC" and removing the timestamp part, such that it creates:
amaltaro_ACDC_EXO-RunIISummer15GS-05060_00274_v2__160315_102225_1898

ticoann:
fixes #6703

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/WMStats.Utils.js
+++ b/src/couchapps/WMStats/_attachments/js/WMStats.Utils.js
@@ -182,7 +182,8 @@ WMStats.Utils.splitCouchServiceURL = function(url) {
 
 
 WMStats.Utils.acdcRequestSting = function(originalRequest, requestor) {
-    var stripRequestor = originalRequest.replace(new RegExp('^'+ requestor), "ACDC");
+	// also stript strings if it contains '@'
+    var stripRequestor = originalRequest.replace(new RegExp('^'+ requestor.split('@')[0]), "ACDC");
     return stripRequestor.replace(/_\d+_\d+_\d+$/, "");
 };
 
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, please review

amaltaro:
Seangchan, this fixes the issue when the original request has @ in the requestor name, right?
If so, besides our test wfs and the short period that we allowed it to happen, it should never ever happend again, right? If so (again), then there is no real need to push this change IMO.

ticoann:
Yes, the only thing is we have the code to handle this in creating the request side. Since request name contains "@" is handled one side, maybe we should add this. Or remove both for consistency.

amaltaro:
I guess I misunderstood this case. The requestor name filtered here does not come from the original request (it's not recycled from the original request), but it's added to the new workflow created via wmstats and that's why we need to filter it here as well. Did I understand it now?

ticoann:
Yes Alan, you are correct. 
To create new ACDC request name from original request name, what it does is filter out requestor name and date part from original requestname.
Normal structure for request name is username_requestString_date.
So it is filteringout username date part and create the new requestSring by add ACDC.
This case username is not the same as requsestor name. Since we added the code to strip @cern.ch
So actual user name is username@cern.ch but in the request we just prefix username not whole string. So when we extract requestsString we need to consider that. We can't parse by '_',  since some of the username contains '_'.

Issue #6720, opened 2016-03-21
Change reqmgr2 splitting page same as reqmgr interface


ticoann:
current template
https://github.com/dmwm/WMCore/blob/master/src/templates/WMCore/WebTools/RequestManager/Splitting.tmpl

Basically
organize parameter by task name. 
With <input>

Allow parameter update by task as well as all the task.

vkuznet:
$ticoann, Seangchan I need more input on that. Do you expect me to visualize
the page from the Javascript code?

- Please provide example of web page how it was done on reqmgr interface.
- Please be specific about splitting page. Do you refer to splitting tab from
  assign page or something else. If former, you explicitly asked to put a full
  JSON in splitting pop-up window. If latter, I have no idea what you're
  talking about.
- Which parameter update are you taking about, need example (again, it would be
  nice if you'll point me to existing web page or make a sketch/photo).

ticoann:
@vkuznet, Hi Valentin,
Here is how it is done in reqmgr1
https://cmsweb-testbed.cern.ch/reqmgr/view/splitting/sryu_TaskChain_Multicore_wmagent_14_160223_225215_4928

splitting page is what we have how the pop up window.  (Yes which I asked to have json page, initially)
Corresponding JSON pop up is here (when click the Splitting tab) as well as in assignment page. (It is using the same code right?)
https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=sryu_TaskChain_Multicore_wmagent_14_160223_225215_4928

I don't think the interface should be exactly same. (reqmgr1 allows the algorithm change)

Here is the detail.
1. Organize the splitting parameter by task name. (The older of task name goes to shorter to longer. not by length but by section i.e /a/bcd goes ahead than /a/b/c.) Secondary order can be by task type (Processing/Merge/Harvest/Skim/LogCollect/Cleanup) but I won't bother it to implement that now. Task type can be added later

2. Only need to display split parameter section from json (make algorithm and type not editable.
```
"splitParams": {
      "min_merge_size": 2147483648, 
      "algorithm": "ParentlessMergeBySize", 
      "initial_lfn_counter": 0, 
      "max_merge_size": 4294967296, 
      "max_merge_events": 400000, 
      "max_wait_time": 86400, 
      "trustSitelists": false, 
      "type": "Merge"
    },
```

So  interface is something like this.

```
[...]

PR #6726, opened 2016-03-22
add algorithm edit list

ticoann:
related to #6720

Just copied the list of algorithm and editable field from reqmgr1.

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
+++ b/src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
@@ -6,10 +6,112 @@ from WMCore.REST.Validation import validate_str
 from WMCore.ReqMgr.Utils.Validation import get_request_template_from_type
 
 import WMCore.ReqMgr.Service.RegExp as rx
-from WMCore.REST.Format import JSONFormat
+from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 
 from WMCore.Wrappers import JsonWrapper
 from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
+
+#TODO: need to get algorithm proper way
+ALGO_DICT = {"FileBased": {"files_per_job": ''},
+             "TwoFileBased": {"two_files_per_job": ''},
+             "LumiBased": {"lumis_per_job": '', 
+                          "halt_job_on_file_boundaries": True},
+             "EventAwareLumiBased": {"avg_events_per_job": '',
+                                     "max_events_per_lumi": '',
+                                     "halt_job_on_file_boundaries": True},
+             "EventBased": {"events_per_job": '',
+                            "events_per_lumi": '',
+                            "lheInputFiles": False},
+             "Harvest": {"periodic_harvest_interval": ''},
+             "ParentlessMergeBySize": {"min_merge_size": '',
+                                       "max_merge_size": '',
+                                       "max_merge_events": '',
+                                       "max_wait_time": ''},
+             "WMBSMergeBySize": {"min_merge_size": '',
+                                 "max_merge_size": '',
+                                 "max_merge_events": '',
+                                 "max_wait_time": ''},
+             "MergeBySize": {"min_merge_size": '',
+                             "max_merge_size": '',
+                             "max_merge_events": '',
+                             "max_wait_time": ''}
+             }
+
+ALGO_LIST_BY_TYPES = {"Processing": ["LumiBased", "EventAwareLumiBased", 
+                             "EventBased", "FileBased"],
+              "Production": ["EventBased"],
+              "Skim": ["FileBased", "TwoFileBased"],
+              "Harvesting": ["Harvest"],
+              "Merge": ["ParentlessMergeBySize", 
+                        "WMBSMergeBySize",
+                        "MergeBySize"],
+              "Cleanup": ["FileBased"],
+              "LogCollect": ["FileBased"]
+              }
+
+def format_algo_web_list(task_name, task_type, split_param):
+    fdict = {"taskName": task_name}
+    fdict["taskType"] = task_type
+    default_algo = split_param["algorithm"]
+    algo_list = ALGO_LIST_BY_TYPES[task_type]
+    param_list = []
+    
+    if default_algo in algo_list:
+        new_param = {"algorithm": default_algo}
+        for key, value in split_param.items():
+            if key in ALGO_DICT[default_algo]:
+                new_param[key] = value
+        param_list.append(new_param) 
+    elif default_algo == "":
+        raise cherrypy.HTTPError(400, "Algorithm name is empty: %s" % split_param)
+    else:
+        param_list.append(split_param)
+    
+    # If task type is merge don't allow change the algorithm
+    if fdict["taskType"] != "Merge":
+        for algo in algo_list:
+            if algo != default_algo:
+                param = {"algorithm": algo}
+                param.update(ALGO_DICT[algo])
+                param_list.append(param)
+                
+    fdict["splitParamList"] = param_list
+    return fdict
+
+def create_web_splitting_format(split_info):
+    web_form = []
+    for sp in split_info:
+        # skip Cleanup and LogCollect: don't allow change the param
+        if sp["taskType"] not in ["Cleanup", "LogCollect"]:
+            web_form.append(format_algo_web_list(sp["taskName"], sp["taskType"], 
+                                               sp["splitParams"]))
+    return web_form
+
+def _validate_split_param(split_algo, split_param):
+    """
+    validate param for editing, also returns param type
+    """
+    valid_params = ALGO_DICT[split_algo]
+    if split_param in valid_params:
+        if isinstance(valid_params[split_param], bool):
+            cast_type = bool
+        else:
+            cast_type = int
+        return (True, cast_type)
+    else:
+        return (False, None)
+            
+def _assign_if_key_exsit(key, original_params, return_params, cast_type):
+    if key in original_params:
+        if cast_type == None:
+            return_params[key] = original_params[key]
+        elif cast_type == bool:
+            if str(original_params[key]).lower() == "true":
+                return_params[key] = True
+            else:
+                return_params[key] = False
+        else:
+            return_params[key] = cast_type(original_params[key])
 
 class RequestSpec(RESTEntity):
     
@@ -24,7 +126,7 @@ class RequestSpec(RESTEntity):
         validate_str("name", param, safe, rx.RX_REQUEST_NAME, optional=False)
 
 
-    @restcall(formats = [('application/json', JSONFormat())])
+    @restcall(formats = [('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
     @tools.expires(secs=-1)
     def get(self, name):
         """
@@ -61,10 +163,11 @@ class WorkloadConfig(RESTEntity):
         are not passed in the method at all.
         
         """
+        
         self._validate_args(param, safe)
 
 
-    @restcall(formats = [('application/json', JSONFormat())])
+    @restcall(formats = [('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
     @tools.expires(secs=-1)
     def get(self, name):
         """
@@ -79,7 +182,7 @@ class WorkloadConfig(RESTEntity):
         try:
             helper.loadSpecFromCouch(self.reqdb_url, name)
         except Exception:
-            raise cherrypy.HTTPError(404, "Cannot find workload: % "+ name)
+            raise cherrypy.HTTPError(404, "Cannot find workload: %s" % name)
         
         return str(helper.data)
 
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, Alan, please review. Especially please check the algorithm list and editable splitting parameters whether is missing parameters or missing algorithm. Thanks,

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,51 @@
...
 import WMCore.ReqMgr.Service.RegExp as rx
-from WMCore.REST.Format import JSONFormat
+from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 
 from WMCore.Wrappers import JsonWrapper
 from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
 
+#TODO: need to get algorithm proper way
+ALGO_DICT = {"FileBased": {"files_per_job": 0,
+                           "include_parents": False},
```
amaltaro: I tend to say "include_parents" should not be a splitting changeable parameter. It's defined during creation time with "IncludeParents" and that's it.

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,51 @@
...
+             "LumiBased":{"lumis_per_job": 0, 
+                          "halt_job_on_file_boundaries": True,
+                           "include_parents": False},
+             "EventAwareLumiBased": {"avg_events_per_job": 0,
+                                     "max_events_per_lumi": 0,
+                                     "halt_job_on_file_boundaries": True,
+                                     "include_parents": False},
+             "EventBased": {"events_per_job": 0,
+                            "events_per_lumi": 0,
+                            "lheInputFiles": 0,
```
amaltaro: this lhe is boolean

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -208,4 +248,45 @@ def post(self, name):
...
+    def __init__(self, app, api, config, mount):
+        # main CouchDB database where requests/workloads are stored
+        RESTEntity.__init__(self, app, api, config, mount)
+        self.reqdb_url = "%s/%s" % (config.couch_host, config.couch_reqmgr_db)
+    
+    
+    def _validate_args(self, param, safe):
+        # TODO: need proper validation but for now pass everything
+        args_length = len(param.args)
+        if args_length > 0 :
```
amaltaro: Isn't it `if args_length < 1:`?
BTW, for my education, besides handling different validations (which is not the case for this class), what's the motivation not to put the actual validation code into the validate() method?

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,51 @@
...
+    fdict = {"taskType": task_type}
+    param_list = []
+    for algo in algo_list:
+        param = {"alogrithm": algo}
+        param.update(ALGO_DICT[algo])
+        param_list.append(param)
+    fdict["splitParamList"] = param_list
+    return fdict
+        
+SPILTT_ALGOS = [format_algo_list("Processing", PROCESSING_ALGO),
```
amaltaro: Should read `SPLIT_ALGOS`, the same in the class get method

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,51 @@
...
+                format_algo_list("Procuction", PRODUCTION_ALGO),
+                format_algo_list("Skim", SKIM_ALGO),
+                format_algo_list("Harviest", HARVEST_ALGO),
```
amaltaro: Harvest

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,51 @@
...
+    param_list = []
+    for algo in algo_list:
+        param = {"alogrithm": algo}
+        param.update(ALGO_DICT[algo])
+        param_list.append(param)
+    fdict["splitParamList"] = param_list
+    return fdict
+        
+SPILTT_ALGOS = [format_algo_list("Processing", PROCESSING_ALGO),
+                format_algo_list("Procuction", PRODUCTION_ALGO),
```
amaltaro: Production

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,51 @@
...
+PROCESSING_ALGO = ["LumiBased", "EventAwareLumiBased", "EventBased", "FileBased"]
+PRODUCTION_ALGO = ["EventBased"]
+SKIM_ALGO = ["FileBased", "TwoFileFileBased"]
+HARVEST_ALGO = ["Harvest"]
+
+def format_algo_list(task_type, algo_list):
+    fdict = {"taskType": task_type}
+    param_list = []
+    for algo in algo_list:
+        param = {"alogrithm": algo}
```
amaltaro: algorithm

amaltaro:
Besides the few comments, it sounds good. It would return something like:
```
{'splitParamList': [{'files_per_job': 0, 'include_parents': False, 'alogrithm': 'FileBased'}, {'two_files_per_job': 0, 'include_parents': False, 'alogrithm': 'TwoFileFileBased'}], 'taskType': 'Skim'}
```
for Skim

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
ticoann: It is raising exception if args is passed so basically it says it shouldn't pass the args.
The reason  it separates _validate_args from validation function is in general, there are different validation requires for different verbs (GET, POST etc) so I would like to separate that. However in this case we have only GET so we could move the code in validation. But I am thinking to add PUT method so we can add new Algorithm on the fly if necessary.  (Maybe never happen. :smile:)

ticoann:
@amaltaro Thank you very much Alan, I fixed things you pointed out. I shouldn't create the patch at night. :smile: Please review again. We have a lot more algorithms exist in WMCore. So the others are not used?
https://github.com/dmwm/WMCore/tree/master/src/python/WMCore/JobSplitting

ticoann:
@amaltaro, After talking to Valentin, I changed the API to give all filtered result for splitting params for web editing. Please review.

https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/web_form/sryu_TaskChain_Task_Multicore_Sryu_reqmgr2_160318_050229_8964

To get all the params.
https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/sryu_TaskChain_Task_Multicore_Sryu_reqmgr2_160318_050229_8964

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
```diff
@@ -6,11 +6,72 @@
...
+             "LumiBased": {"lumis_per_job": '', 
+                          "halt_job_on_file_boundaries": True},
+             "EventAwareLumiBased": {"avg_events_per_job": '',
+                                     "max_events_per_lumi": '',
+                                     "halt_job_on_file_boundaries": True},
+             "EventBased": {"events_per_job": '',
+                            "events_per_lumi": '',
+                            "lheInputFiles": False},
+             "Harvest": {"periodic_harvest_interval": ''},
+             "ParentlessMergeBySize": {"min_merge_size": '',
```
amaltaro: We shouldn't touch Merge algos. AFAIK WMCore enforces it according to the task type and job spliting for the Processing/Production task

amaltaro:
The special handling for default_algo (the initial algo for that workflow) is to show the original job splitting values, isn't it?
The Merge splitting we _usually_ do not touch, I'm not sure we should show them in the Gui... and AFAIK we use mostly ParentlessMergeBySize and WMBSMergeBySize.

For the other algos, if they are not used by T0, then they are legacy: FixedDelay, MergeBySize (not sure), MinFileBased, RunBased (hmm, sounds interesting), SiblingProcessingBased (what?), SizeBased, SplitFileBased, TwoFileAndEventBased ...oh well, where these ideas come from...

src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
amaltaro: Nevermind this comment, I'm deleting it

amaltaro:
Ok, looking at your web_form:
https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/web_form/sryu_TaskChain_Task_Multicore_Sryu_reqmgr2_160318_050229_8964

I changed my mind, there may be use cases where we want to change the Merge splitting parameters, but I still think the algo itself should be unchangeable by the user.

About Celanup and LogCollect, I'd keep the web form clean and simply not show them. This would be one complain that I have about the reqmgr splitting page, it has way too much info...

ticoann:
Thanks, Alan, I changed the api as you suggested.
1. remove cleanup and log collect update
2. For merge only parameter changes but not algorithm.

https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/web_form/sryu_TaskChain_Task_Multicore_Sryu_reqmgr2_160318_050229_8964

Please review.

ticoann:
Hi Alan, I refactored code and updated.

Looking at the test TaskChain, I see something weird.

https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/sryu_McMTaskChain2_Sryu_reqmgr2_160318_050204_3263

There is on Merge is set with "EventAwareLumiBased",
Is that a mistake?
Also there empty algorithm, I guess in that case it takes default algorithm
```
{
  "splitAlgo": "EventAwareLumiBased", 
  "splitParams": {
    "algorithm": "EventAwareLumiBased", 
    "initial_lfn_counter": 0, 
    "max_events_per_lumi": 20000, 
    "trustSitelists": false, 
    "type": "Merge"
  }, 
  "taskName": "/sryu_McMTaskChain2_Sryu_reqmgr2_160318_050204_3263/EXO-RunIIFall15DR76-03799_0/EXO-RunIIFall15DR76-03799_1/EXO-RunIIFall15DR76-03799_1MergeDQMoutput", 
  "taskType": "Merge"
}
```

amaltaro:
There is something wrong here, both first task and a merge task have an empty list of algos (besides a Merge task showing up a Processing algo).

I injected a good test case:
https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/amaltaro_TaskChain_MC_DMWM_TEST_SplitAPI_160324_113023_1915
Task1: should default to EventBased, 300 evts per Job
Task2: should default to EventAwareLumiBased, 600 evts per job (6l)
Task3: should be LumiBased, as enforced in the schema, 1000 evts per job (10l)

ticoann:
Yes something is definitely wrong.
From schema, The SplitAlgo is missing for Task1 and Task2

https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_TaskChain_MC_DMWM_TEST_SplitAPI_160324_113023_1915

amaltaro:
BTW @vkuznet I connected to reqmgr2-dev and I like it much more now, it's more compact and seems cleaner ;)

ticoann:
@amaltaro, Alan, could you point me the json file you used to create
amaltaro_TaskChain_MC_DMWM_TEST_SplitAPI_160324_113023_1915?

Is this one?
https://github.com/dmwm/WMCore/blob/master/test/data/ReqMgr/requests/DMWM/TaskChain_MC.json

Or something else. It seems above one has algorithm specified.

Anyway, it seems the problem is not in this patch but (it might be the way we are setting default)
I updated the code so instead of allowing empty algorithm, we will through error.

Tested with above json and seems to work fine. I am not sure how I created my workflow.

amaltaro:
It's basically the same template, but with a few tweaks. You can see it here:
http://amaltaro.web.cern.ch/amaltaro/forSeangChan/TaskChain_MC.json

BTW, the way we set default works just fine, it must be something related to reqmgr2 only.

ticoann:
Hi Alan,
I created two workflows  with your json.
reqmgr1 - sryu_TaskChain_MC_alan_Sryu_reqmgr1_160324_185923_7359
reqmgr2 - sryu_TaskChain_MC_alan_Sryu_reqmgr1_160324_185907_1165

Both seems to be same and fine. Not sure could you try to create again.

https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/sryu_TaskChain_MC_alan_Sryu_reqmgr1_160324_185923_7359

https://reqmgr2-dev.cern.ch/reqmgr2/data/splitting/sryu_TaskChain_MC_alan_Sryu_reqmgr1_160324_185907_1165

ticoann:
Merging it now. Will further debug if the problem occurs

PR #6735, opened 2016-03-23
remove non-functional phedex inject override

hufnagel:
Set and get methods are useless because the underlying functionality has been removed long ago.


hufnagel:
Ok, let me fix that one unit test.

PR #6736, opened 2016-03-23
fix task archiver for tier0 case

ticoann:
tested in vocms001. @hufnagel, Dirk, please review


hufnagel:
Hm, tier0 has different team names for replay and production. Does this matter ?

ticoann:
It doesn't matter as long as it uses different database. If there is plan for using two different team (agents) running for tier0, we need to change things.

1. put team in the request document
2. create the view for team and status.

hufnagel:
Replay tier0 instances should never publish information in the same databases as the production instance. We can have two production instances running in parallel though and these would write to the same central database.

ticoann:
with the same team name or different team name? Anyway, it it not completely necessary to have team name for this case It just reduces what request pulled down. I added this since there are few different teams in production agents.

hufnagel:
Same team name I believe, although you might want to check with John or Eliana.

ticoann:
OK I will do that.

PR #6738, opened 2016-03-23
Use uuid from python

### LINKED ISSUE #6737: UUID code included as external repeats UUIDs from time to time
Let's try removing this code and using the actual python library

ericvaandering:
Fixes #6737 (I hope)


ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
Let's merge this. We were seeing a UUID collision in the unit tests every week or two. We can see if we ever see one again

ticoann:
OK, I was just wondering what is the advantages and disadvantages between different algorithms. uuid(1,2,3,4)

ericvaandering:
Well, UUID4() is a completely random UUID and should not collide during the life of the universe. But it could. UUID1 is based on the MAC ID of the machine and the time so will *never* collide assuming the request rate is less than a certain amount. But if there are issues with the clock, it can go pear-shaped. So I think that's what happened. Hopefully the C based algorithm will do its job. If not, we should switch to UUID4().

ericvaandering:
uuid2 and 3 are not really relevant. I think one or both of them require you to supply some name instead of the MAC address used in uuid1. 

uuid1 is the thing that gives us similar strings at the end of file names. Why we didn't go for uuid4 in the beginning, I'm not really sure except that there is a very, very small probability that it's not unique (assuming your random # generator doesn't have a problem)

PR #6741, opened 2016-03-24
1.0.14_crab version of Make sure httplib2 is not required if using pycurl

mmascher:


amaltaro:
I'm not a big fan of imports in the middle of the code (nor PEP-8 likes it), but I'm merging this one that is specific to CRAB and then we can discuss on the master one.

amaltaro:
Daahh, Seangchan merged it already :(

mmascher:
Me neither, but I am also not a big fan of httplib2 since it is not a
standard package. And I don't see why we should depend on it if there is
the option to use pycurl.

amaltaro:
AFAIK pycurl is not production ready yet, there were some issues to be sorted out, but I didn't really follow up on those discussions and tests.

mmascher:
The CRAB3 client has always been using it, but it does not use Service.py
and Requests.py if not for contacting phedex/sitedb.

PR #6742, opened 2016-03-24
Make sure httplib2 is not required if using pycurl

mmascher:


ericvaandering:
What's the reasoning behind this? Just reducing a dependency for the CRAB client?

mmascher:
Yep, and having the crabclient run in the CMSSW environment sourcing the
minimal amount of stuff

ericvaandering:
Ah, I see longer discussion in the branch version and Alan saying exactly what I would have said: "I'm not a big fan of imports in the middle of the code"

ericvaandering:
I think the goal here is important enough to have a sub-optimal coding practice, but I'd like to see an issue attached to the pycurl issue to investigate whether we can completely get rid of http2lib in this case.

mmascher:
Ok, I will open a github issue. Which one is the pycurl issue?

amaltaro:
I don't think we need an issue for that, there is a PR already in place for it:
https://github.com/dmwm/WMCore/pull/6360/files

and by reading the final comments, it seems we are waiting for reqmgr (one) to go
away in order to make pycurl default.

mmascher:
Can we merge this one?

amaltaro:
test this please

amaltaro:
Hmmm, I wonder why it suggested
`if isinstance(encoder, (type(self.get)) or isinstance(encoder, type(f)):`
instead of
`if isinstance(encoder, (type(self.get), type(f))):`
?

@mmascher , can you make another commit to fix the pre-python 2.6 comments and some pylint complaints (like imports unused and other things that are clear)?

amaltaro:
ping, now is a good moment to get it fixed and merged

mmascher:
done

mmascher:
test this please

mmascher:
retriggering tests since the unit test error seemed temporary: `DBS is not responding: https://cmsweb.cern.ch/dbs/prod/global/DBSReader`

amaltaro:
test this please

amaltaro:
pushing this one to my VM as well.

amaltaro:
@mmascher can you squash the last two commits please? If you need to change any additional bits, let me please. Otherwise this patch is good to be merged.

mmascher:
I split them in two commits on purpose but if you want I can still squash them.

Reason is that pylint changes are safe (just variable renaming, removed unused variables ect), while strictly speaking fixes for Pre-python 2.6 idioms change the semantic of the code (isinstance(a,b) is different than type(a)==b because it checks subclasses as well). So I put the "dangerous" commit alone (although it's really unlikely it will cause issues and I consider this safe as well :) )

amaltaro:
Let's be optimistic, they both are quite safe. For the sake of going through the history, I think it's better to squash them :-)

amaltaro:
Thanks, Marco

PR #6745, opened 2016-03-24
1.0.14_crab version of WMCore Import fix

emaszs:
A fix for a small bug which throws missing module errors because of an incorrect import location. 

Would also like to have a new tag for this branch after a pull request is merged, 1.0.14_crab_4. Thanks.

### DIFF
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -183,8 +183,6 @@ class Requests(dict):
         as a string.
 
         """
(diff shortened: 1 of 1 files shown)


mmascher:
My bad, I tested things with the import in the right place, then tried another change that did not work, and finally put back the import in the wrong place. I have already fixed the PR for the amster branch.

amaltaro:
Bad job, Marco! I'm on it

amaltaro:
tag created 1.0.14_crab_4

PR #6746, opened 2016-03-24
patch before init-agent

ticoann:
In some case,  patch need to be applied before init-agent. since that uses WMCore code to create the db.


ticoann:

amaltaro:
I don't remember if we have all the files in place before the init-agent (or is it activate-agent...).
If you don;t want to wait, just merge it. Otherwise I'll test it on Tuesday

amaltaro:
and before you say 'stop working', I'm stopping now :-D

PR #6747, opened 2016-03-24
fix LocalQueueProfile_t

ticoann:
@ericvaandering, Eric, please review.

### DIFF
--- a/src/python/WMQuality/Emulators/DBSClient/DBSReader.py
+++ b/src/python/WMQuality/Emulators/DBSClient/DBSReader.py
@@ -67,7 +67,7 @@ class DBSReader:
         self.dataBlocks = DataBlockGenerator()
         args = { "url" : url, "level" : 'ERROR', "version" : 'DBS_2_0_9'}
         self.dbs = _MockDBSApi(args)
-
+        
     def getFileBlocksInfo(self, dataset, onlyClosedBlocks = True,
                           blockName = '*', locations = True):
 
--- a/test/python/WMCore_t/WorkQueue_t/LocalQueueProfile_t.py
+++ b/test/python/WMCore_t/WorkQueue_t/LocalQueueProfile_t.py
@@ -14,7 +14,7 @@ from WMCore.Services.EmulatorSwitch import EmulatorHelper
 from WMQuality.Emulators.DataBlockGenerator import Globals
 from WMQuality.Emulators.WMSpecGenerator.WMSpecGenerator import WMSpecGenerator
 from WMCore.WorkQueue.WorkQueue import localQueue
-from .WorkQueueTestCase import WorkQueueTestCase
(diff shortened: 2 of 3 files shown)


test/python/WMCore_t/WorkQueue_t/LocalQueueProfile_t.py
```diff
@@ -14,7 +14,7 @@
 from WMQuality.Emulators.DataBlockGenerator import Globals
 from WMQuality.Emulators.WMSpecGenerator.WMSpecGenerator import WMSpecGenerator
 from WMCore.WorkQueue.WorkQueue import localQueue
-from .WorkQueueTestCase import WorkQueueTestCase
+from WorkQueueTestCase import WorkQueueTestCase
```
ericvaandering: This is not correct python3. You need to add WMCore_t.WorkQueue_t. in front. In general, relative imports are bad, especially if we don't need them.

ericvaandering: Hopefully Jenkins will catch this. Let's see.

src/python/WMQuality/Emulators/DBSClient/DBSReader.py
```diff
@@ -9,19 +9,19 @@
 
 class _MockDBSApi():
     """Mock dbs api"""
-    def __init__(self, args):
```
ericvaandering: I don't see how any of the changes here relate to the other changes or to any issue.

ericvaandering:
Yeah, "nose.failure.Failure:runTest was added. Status is error" is your import error.

src/python/WMQuality/Emulators/DBSClient/DBSReader.py
ticoann: args are dict type below line 68, but DbsApi first argument is url str. which gives the error. 
Also,  DbsApi init signature is (url="", proxy=None, key=None, cert=None, verifypeer=True, debug=0, ca_info=None, userAgent="") so I just to want to make compatible.

test/python/WMCore_t/WorkQueue_t/LocalQueueProfile_t.py
ticoann: fixed it now.

ticoann:
> Yeah, "nose.failure.Failure:runTest was added. Status is error" is your import error.

Power of jenkins test.:smile: Thanks

ticoann:
test this please

ticoann:
Hmm, the errors doesn't seems to be related to this patch.

ticoann:
test this please

ericvaandering:
@ticoann we have a winner.

ticoann:
Yes, you are right, I don't know how that could affected the failing test though. Thanks, I will merge this.

ericvaandering:
Probably just an instability we haven't noted yet.

PR #6748, opened 2016-03-25
Make username and group consistent among resubmit tests

ericvaandering:
Spec and ACDC had different user/group names and fetch was being attempted with the wrong pair.


ericvaandering:
@ticoann please merge. The four tests were throwing exceptions. Now they are just disagreeing about sites where things can run. @jha2 can fix that up when moving it to the mock emulator.

PR #6752, opened 2016-03-26
Split trust location flag for input and for pileup data

amaltaro:
Disclaimer: it's a non working version :)

A summary of this patch is:
- get rid of the 'NoLocationUpdate' WQE parameter
- added a new 'NoInputUpdate' WQE parameter, set by "TrustSitelists" flag to bypass only the input dataset
- added a new 'NoPileupUpdate' WQE parameter, set by "TrustPUSitelists" flag to bypass only the pileup dataset

I had a few typos in the javascript code, that was fixed and doesn't break WQ anymore, however, when I set a) only the PU flag or b) when I set both, somehow blocks don't get replicated from GQ to LQ.

@ticoann if you spend time looking at this, check only the workRestrictions.js and DataStructs/WorkQueueElement.py. Otherwise, I'll sort it out on Tuesday.

PR #6755, opened 2016-03-29
Split trust location flag for input and for pileup data

ticoann:
replace #6752

### DIFF
--- a/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
+++ b/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
@@ -72,7 +72,7 @@ class WorkQueueElement(dict):
         # When was the last time we found new data (not the same as when new data was split), e.g. An open block was found
         self.setdefault('TimestampFoundNewData', 0)
         # Trust initial input and pileup location or not
-        #self.setdefault('NoLocationUpdate', False)
+        self.setdefault('NoLocationUpdate', False)
         # Trust initial input dataset location only or not
         self.setdefault('NoInputUpdate', False)
         # Trust initial pileup dataset location only or not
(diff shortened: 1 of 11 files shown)


ticoann:
@amaltaro, with this patch it seems it is pulling down the work to LQ,
Please review. I won't include in the tag now until you review.

There are a couple of minor issues.
1. when this is updated agent if the all the job are not create from previous workflow following code change is needed in WMTask - so wmagent branch patch need to have this.

``` python
    def getTrustSitelists(self):
        """
        _getTrustSitelists_

        Get the input and pileup flag for 'trust site lists' in the task.
        """
        
        if hasattr(self.data.constraints.sites, "trustPUlists"):
            trustPUlists =self.data.constraints.sites.trustPUlists
        else:
            trustPUlists = False
        return (self.data.constraints.sites.trustlists, trustPUlists)
```
2. unittest need to be fixed

Thanks,

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
```diff
@@ -71,8 +71,12 @@ def __init__(self, **kwargs):
         self.setdefault('OpenForNewData', False)
         # When was the last time we found new data (not the same as when new data was split), e.g. An open block was found
         self.setdefault('TimestampFoundNewData', 0)
-        # Should we check the location of the inputs, or trust the initial values?
+        # Trust initial input and pileup location or not
         self.setdefault('NoLocationUpdate', False)
```
amaltaro: Just remove (or comment it out for now) this NoLocationUpdate line please. It will be redundant and the less flags we have, the better.

ticoann: Hi Alan, we have to remove completely 'NoLocationUpdate', or keep this line.Otherwise it fails in line 277.
I am fine with completely removing it. But I think you left it other places for backward compatibility. So I think we need to keep 'NoLocationUpdate' in wmagent branch.

amaltaro: I'm in favor of removing it for any new WQE, though we still have to check whether old WQE have this flag enabled.

amaltaro: I guess that explains why several tests failed in the other PR :-(

ticoann: I think it might be safer to keep it for this deployment and remove it all together next deployment although I think in most cases, it will be fine. (except something is assigned before the GQ updated. but anyway for agent side it will be safer to include it.  (So I would keep it now and remove it next deployment all together)

ticoann:
Alan, feel free to close this patch if you updated yours.

amaltaro:
Thanks a lot for your help Seangchan, I'm closing this one and we merge #6752. Thanks!

PR #6756, opened 2016-03-29
Fix GFAL2 and environment problem

juztas:
semicolon in env -i leads to execute two diff commands with different environment. Fix a bug which was reported in several threads:
https://ggus.eu/index.php?mode=ticket_info&ticket_id=120327
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1478/1/2/1.html

P.S. This affects only WMAgent as CRAB3 branch is not using this pull request yet: https://github.com/dmwm/WMCore/commit/68f64875567501bc83ee63fb6543f84319f85611

### DIFF
--- a/src/python/WMCore/Storage/Backends/GFAL2Impl.py
+++ b/src/python/WMCore/Storage/Backends/GFAL2Impl.py
@@ -20,10 +20,13 @@ class GFAL2Impl(StageOutImpl):
 
     def __init__(self, stagein=False):
         StageOutImpl.__init__(self, stagein)
-        self.setups = "env -i X509_USER_PROXY=$X509_USER_PROXY; "
-        self.setups += 'date "+%Y-%m-%dT%H:%M:%S";'
-        self.removeCommand = "%s gfal-rm -vvv -t 600 " % self.setups
-        self.copyCommand = "%s gfal-copy -vvv -t 2400 -T 2400 -p " % self.setups
+        # If we want to execute commands in clean shell, we can`t separate them with ';'.
+        # Next commands after separation are executed without env -i and this leads us with 
+        # mixed environment with COMP and system python.
+        # GFAL2 is not build under COMP environment and it had failures with mixed environment.
+        self.setups = "env -i X509_USER_PROXY=$X509_USER_PROXY sh -c '%s'"
+        self.removeCommand = self.setups % 'printenv; date; gfal-rm -vvv -t 600 %s '
+        self.copyCommand = self.setups % 'printenv; date; gfal-copy -vvv -t 2400 -T 2400 -p %(checksum)s %(options)s %(source)s %(destination)s'
 
     def createSourceName(self, protocol, pfn):
         """
@@ -58,11 +61,11 @@ class GFAL2Impl(StageOutImpl):
                Command is interrupted if time expires before it finishes
         """
         if pfn.startswith("file:"):
-            return self.removeCommand + pfn
+            return self.removeCommand % pfn
         elif os.path.isfile(pfn):
             return "/bin/rm -f %s" % os.path.abspath(pfn)
         else:
-            return self.removeCommand + pfn
+            return self.removeCommand % pfn
 
 
     def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=None):
@@ -81,14 +84,15 @@ class GFAL2Impl(StageOutImpl):
 
         useChecksum = (checksums != None and 'adler32' in checksums and not self.stageIn)
 
-        copyCommand = self.copyCommand
+        copyCommandDict = {'checksum': '', 'options': '', 'source': '', 'destination': ''}
         if useChecksum:
-            copyCommand += " -K adler32 "
+            copyCommandDict['checksum'] = "-K adler32"
         if options != None:
-            copyCommand += " %s " % options
-        copyCommand += " %s " % sourcePFN
-        copyCommand += " %s \n" % targetPFN
+            copyCommandDict['options'] = options
+        copyCommandDict['source'] = sourcePFN
+        copyCommandDict['destination'] = targetPFN
 
+        copyCommand = self.copyCommand % copyCommandDict
         result += copyCommand
 
         if _CheckExitCodeOption:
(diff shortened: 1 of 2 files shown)


juztas:
For me it failed in different location. AFAIK SEs should not be used anymore from sitedb or I missed this, but I added one print to check this.

src/python/WMCore/Storage/Backends/GFAL2Impl.py
```diff
@@ -20,10 +20,18 @@ class GFAL2Impl(StageOutImpl):
...
+        # are executed without env -i and this leads us with mixed environment with COMP and system python.
+        # GFAL2 is not build under COMP environment and it had failures with mixed environment.
+        self.setups = "env -i X509_USER_PROXY=$X509_USER_PROXY sh -c '%s'"
+        # %s in the end is used for adding file location which is needed to be removed
+        self.removeCommand = self.setups % 'printenv; date; gfal-rm -vvv -t 600 %s '
+        # %(checksum) is adding a checksum option
+        # %(option) is adding options which are defined in local site conf
+        # %(source) is source file location
+        # %(destination) is destination file location
+        self.copyCommand = self.setups % 'printenv; date; gfal-copy -vvv -t 2400 -T 2400 -p %(checksum)s %(options)s %(source)s %(destination)s'
```
ericvaandering: I think this code is over documented. (Probably the first time I've said this). If you are doing something tricky, by all means document it, but I don't think you need to specify all the parameters since that's pretty obvious.

test/python/WMCore_t/Services_t/SiteDB_t/SiteDB_t.py
```diff
@@ -114,7 +115,7 @@ def testPNNtoPSN(self):
     
     def testCMSNametoList(self):
         result = self.mySiteDB.cmsNametoList("T1_US*", "SE")
-        self.assertTrue(result == [u'cmssrm.fnal.gov', u'cmssrmdisk.fnal.gov'])
+        self.assertTrue(sorted(result) == [u'cmssrm.fnal.gov', u'cmssrmdisk.fnal.gov'])
```
ericvaandering: Try assertItemsEqual here, no need of sorted() or to get the second list right

juztas:
mh... ok, sitedb was just a noise. Will change what Eric suggested ;)

src/python/WMCore/Storage/Backends/GFAL2Impl.py
juztas: I took seriously what Stefano was saying about documenting the code ;)

juztas:
test this please

juztas:
3 runs and different results? Mh.. Now its time to set up the wma environment for testing unittests. Regarding this pull request it is good to go. It can't affect Workqueue.

amaltaro:
Justas, no need for that. These unit tests are unrelated. Actually, I don't think we have Storage/plugin unit tests at all.
Are you done with these changes? If so, I'll patch my test agent and run a couple of workflows at Purdue.

juztas:
I need for another pull request, which requires wma env. This one is good to go. Thanks Alan

src/python/WMCore/Storage/Backends/GFAL2Impl.py
```diff
@@ -20,10 +20,13 @@ class GFAL2Impl(StageOutImpl):
...
-        self.setups = "env -i X509_USER_PROXY=$X509_USER_PROXY; "
-        self.setups += 'date "+%Y-%m-%dT%H:%M:%S";'
-        self.removeCommand = "%s gfal-rm -vvv -t 600 " % self.setups
-        self.copyCommand = "%s gfal-copy -vvv -t 2400 -T 2400 -p " % self.setups
+        # If we want to execute commands in clean shell, we can`t separate them with ';'.
+        # Next commands after separation are executed without env -i and this leads us with 
+        # mixed environment with COMP and system python.
+        # GFAL2 is not build under COMP environment and it had failures with mixed environment.
+        self.setups = "env -i X509_USER_PROXY=$X509_USER_PROXY sh -c '%s'"
+        self.removeCommand = self.setups % 'printenv; date; gfal-rm -vvv -t 600 %s '
```
amaltaro: Justas, can you keep the same date format as before please.

juztas: I wanted to skip all these nasty things. See here:
```
>>> command = 'date "+%Y-%m%dT%H:%M:%S"; gfal-copy %s'
>>> command
'date "+%Y-%m%dT%H:%M:%S"; gfal-copy %s'
>>> command % 'test'
ValueError: unsupported format character 'Y' (0x59) at index 8
>>> command = 'date "+%%Y-%%m%%dT%%H:%%M:%%S"; gfal-copy %s'
>>> command % 'test'
'date "+%Y-%m%dT%H:%M:%S"; gfal-copy test'
```

amaltaro: To me, this format string (%s) assigned to a variable is already nasty enough (and that's why I got rid of it in my previous patch).
The good thing of keeping that date format is that it will be consistent with most part - if not all - of WMCore code, having different timestamps format is painful.

juztas: It's just a taste of code, I think ;) I can add it, but why adding a date everywhere in each plugin, while you can add it before executing the command, which is already done. Why you need this in multiple places?
https://github.com/dmwm/WMCore/blob/0b3086a3b359bdd2d1a75461c7f7f3f8a14ea4bf/src/python/WMCore/Storage/StageOutImpl.py#L175

amaltaro: The timestamp in the stage out wrapper has been proven not to be good enough (during the KIT missing files investigation). Sometimes a print just before a copy/rm comes in handy. Keep it as it is then, if needed we can update it in the future, of course.

amaltaro:
Indeed GFAL2 is broken in the old agent (1.0.12), all my jobs failed over the last night without this patch.
Thanks for this fix, Justas.

PR #6757, opened 2016-03-29
Allow runtime environment to resize the cores in the job.

bbockelm:
This allows WMRuntime to resize the number of threads in the pset
_if_ the job is in the condor runtime environment (`_CONDOR_MACHINE_AD`
env var is set and the slot's ad is in the referenced file), the
`Cpus` attribute is set to an integer, and `WMCore_ResizeJob` is set to
`true`.

So, by default this does nothing.  However, Ops can then explicitly
enable job resizing (and setting `RequestCpus` and `RequestMemory`
appropriately to reflect this) for a given workflow or subtask in
condor.  Since I suspect we'll need to experiment a bit with the
correct values for `RequestMemory` / `RequestCpus`, I don't currently
change those in the `CondorPlugin`.

This just gives us the flexibility to explore the phase space.

Cases of the `resizeCores` function tested by hand:
- When `_CONDOR_MACHINE_AD` env var is missing.
- When machine ad is unparseable.
- When resize job attribute is missing, true, or false.
- Non-parseable values for `Cpus` or `WMCore_ResizeJob`
- `Cpus` is greater than, equal to, or less than numCores

### DIFF
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -10,6 +10,8 @@ process/config but does not depend on any CMSSW libraries. It needs to stay like
 from __future__ import print_function
 import logging
 import pickle
+import traceback
+import os
 
 from PSetTweaks.PSetTweak import PSetTweak
 from PSetTweaks.PSetTweak import parameterIterator, psetIterator
@@ -518,3 +520,95 @@ def makeOutputTweak(outMod, job):
     #      output module based on the settings in the section
 
     return result
+
+
+def readAdValues(attrs, adname, castInt=False):
+    """
+    A very simple parser for the ads available at runtime.  Returns
+    a dictionary containing
+    - attrs: A list of string keys to look for.
+    - adname: Which ad to parse; "job" for the $_CONDOR_JOB_AD or
+      "machine" for $_CONDOR_MACHINE_AD
+    - castInt: Set to True to force the values to be integer literals.
+      Otherwise, this will return the values as a string representation
+      of the ClassAd expression.
+
+    Note this is not a ClassAd parser - will not handle new-style ads
+    or any expressions.
+
+    Will return a dictionary containing the key/value pairs that were
+    present in the ad and parseable.
+
+    On error, returns an empty dictionary.
+    """
+    adfile = None
+    if adname == 'job':
+        adfile = os.environ.get("_CONDOR_JOB_AD")
+    elif adname == 'machine':
+        adfile = os.environ.get("_CONDOR_MACHINE_AD")
+    else:
+        print("Invalid ad name requested for parsing: %s" % adname)
+        return {}
+    if not adfile:
+        print("%s adfile is not set in environment." % adname)
+        return {}
+    attrs = [i.lower() for i in attrs]
+    retval = {}
+    try:
+        with open(adfile) as fd:
+            for line in fd:
+                info = line.strip().split("=", 1)
+                if len(info) != 2:
+                    continue
+                attr = info[0].strip().lower()
+                val = info[1].strip()
+                if attr in attrs:
+                    if castInt:
+                        try:
+                            retval[attr] = int(val)
+                        except ValueError:
+                            print("Error parsing %s's %s value: %s", (adname, attr, val))
+                    else:
+                        retval[attr] = val
+    except IOError:
+        print("Error opening %s ad:" % adname)
+        print(traceback.format_exc())
+        return {}
+
+    return retval
+
+
+def resizeResources(resources):
+    """
+    _resizeResources_
+
+    Look at the job runtime environment and determine whether we are allowed
+    to resize the core count.  If so, change the resources dictionary passed
+    to this function according to the information found in $_CONDOR_MACHINE_AD.
+    The following keys are changed:
+     - cores -> uses value of Cpus from the machine ad.
+     - memory -> Memory
+
+    This only works when running under HTCondor, $_CONDOR_MACHINE_AD exists,
+    and WMCore_ResizeJob is true.
+          - WMCore_ResizeJob is 'true'
+
+    No return value - the resources directory is changed in-place.
+    Should not throw an exception - on error, no change is made and a message
+    is printed out.
+    """
+    if readAdValues(['wmcore_resizejob'], 'job').get('wmcore_resizejob', 'false').lower() != "true":
+        print("Not resizing job")
+        return
+
+    print("Resizing job.  Initial resources: %s" % resources)
+    adValues = readAdValues(['memory', 'cpus'], 'machine', castInt=True)
+    machineCpus = adValues.get('cpus')
+    machineMemory = adValues.get('memory')
+    if machineCpus and (machineCpus > 0) and ('cores' in resources):
+        resources['cores'] = machineCpus
+    if machineMemory and (machineMemory > 0) and ('memory' in resources):
+        resources['memory'] = machineMemory
+    print("Resizing job.  Resulting resources: %s" % resources)
+
+
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -16,7 +16,7 @@ import pickle
 from WMCore.WMRuntime.ScriptInterface import ScriptInterface
 from WMCore.Storage.TrivialFileCatalog import TrivialFileCatalog
 from PSetTweaks.PSetTweak import PSetTweak
-from PSetTweaks.WMTweak import applyTweak
+from PSetTweaks.WMTweak import applyTweak, resizeResources
 from PSetTweaks.WMTweak import makeOutputTweak, makeJobTweak, makeTaskTweak
 from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig
 from WMCore.Wrappers.JsonWrapper import JSONDecoder
@@ -664,8 +664,13 @@ class SetupCMSSWPset(ScriptInterface):
         self.fixupProcess()
 
         try:
-            if int(self.step.data.application.multicore.numberOfCores) > 1:
-                numCores = int(self.step.data.application.multicore.numberOfCores)
+            origCores = int(self.step.data.application.multicore.numberOfCores)
+            resources = {'cores': origCores}
+            resizeResources(resources)
+            numCores = resources['cores']
+            # When we are using multiple cores - or the pset might have
+            # a default of greater than 1 - we must override the provided pset.
+            if (numCores > 1) or (origCores > 1):
                 options = getattr(self.process, "options", None)
                 if options == None:
                     self.process.options = cms.untracked.PSet()
--- a/src/python/WMCore/WMRuntime/Watchdog.py
+++ b/src/python/WMCore/WMRuntime/Watchdog.py
@@ -5,6 +5,8 @@ _Watchdog_
 
 This cleverly named object is the thread that handles the monitoring of individual jobs
 """
+from __future__ import division
+from __future__ import print_function
 
 import os
 import os.path
@@ -14,6 +16,7 @@ import traceback
 
 from WMCore.WMFactory   import WMFactory
 from WMCore.WMException import WMException
+from PSetTweaks.WMTweak import resizeResources
 
 class WatchdogException(WMException):
     """
@@ -78,6 +81,46 @@ class Watchdog(threading.Thread):
                 #This should be a config section
                 monitorArgs = getattr(task.data.watchdog, monitor)
                 args = monitorArgs.dictionary_()
+                # Scale resources according to the HTCondor runtime environment.
+                origCores = 1
+                for stepName in task.listAllStepNames():
+                    sh = task.getStepHelper(stepName)
+                    origCores = max(origCores, sh.getNumberOfCores())
+                resources = {'cores': origCores}
+                origMaxRSS = args.get('maxRSS')
+                if origMaxRSS:
+                    origMaxRSS = int(origMaxRSS/1024.) # HTCondor expects MB; we get KB.
+                    resources['memory'] = origMaxRSS
+                # Actually parses the HTCondor runtime
+                resizeResources(resources)
+                # We decided to only touch Watchdog settings if the number of cores changed.
+                # (even if this means the watchdog memory is wrong for a slot this size).
+                changedCores = origCores != resources['cores']
+                # HTCondor doesn't explicitly scale VSize; it's also not clear what
+                # resources this manages (as we already watch the memory use) or how
+                # it should relate to other resources (such as memory or cores used).
+                # Hence, we simply remove it if we change anything about the memory.
+                if changedCores and origMaxRSS and (resources['memory'] != origMaxRSS) and ('maxVSize' in args):
+                    del args['maxVSize']
+                # Scale our timeouts according to the scaling of the cores.  *NOTE*
+                # the job ad contains the real scaling here - unfortunately, it uses
+                # ClassAd expressions and we can only parse values.  Something for the
+                # future.
+                # As the scaling is not perfectly linear (efficiency is closer to 95%),
+                # we add a somewhat-arbitrarily-chosen 60 minutes (timeouts are in seconds).
+                if changedCores:
+                    coreScaling = resources['cores'] / origCores
+                    if 'softTimeout' in args:
+                        args['softTimeout'] = int(args['softTimeout']/coreScaling) + 3600
+                    if 'hardTimeout' in args:
+                        args['hardTimeout'] = int(args['hardTimeout']/coreScaling) + 3600
+                # If we did base maxRSS off the memory in the HTCondor slot, subtract a bit
+                # off the top so watchdog triggers before HTCondor does.
+                if changedCores and origMaxRSS and (resources['memory'] != origMaxRSS):
+                    args['maxRSS'] = 1024*(resources['memory'] - 50) #Convert back to KB
+                logging.info("Final Watchdog settings:")
+                for k, v in args.iteritems():
+                    logging.info("  %s: %r", k, v)
             # Actually initialize the monitor variables
             mon.initMonitor(task = task, job = wmbsJob,
                             logPath = self.logPath, args = args)
--- a/src/python/WMCore/WMRuntime/Tools/Scram.py
+++ b/src/python/WMCore/WMRuntime/Tools/Scram.py
@@ -268,6 +268,10 @@ class Scram:
             self.procWriter(proc, 'export VO_CMS_SW_DIR=%s/cmssoft/cms\n'%os.environ['OSG_APP'])
         if os.environ.get('CMS_PATH', None) != None:
             self.procWriter(proc, 'export CMS_PATH=%s\n'%os.environ['CMS_PATH'])
+        if os.environ.get('_CONDOR_JOB_AD'):
+            self.procWriter(proc, 'export _CONDOR_JOB_AD=%s\n'%os.environ['_CONDOR_JOB_AD'])
+        if os.environ.get('_CONDOR_MACHINE_AD'):
+            self.procWriter(proc, 'export _CONDOR_MACHINE_AD=%s\n'%os.environ['_CONDOR_MACHINE_AD'])
 
         if hackLdLibPath:
             self.procWriter(proc, self.library_path)


bbockelm:
@hufnagel - this is the idea we discussed over email.

amaltaro:
Brian, I'm probably not seeing the whole picture here, thus I have a few questions:
1. where does the "WMCore_ResizeJob" var (job classAd?) come from? It's not set anywhere in the agent, is it some site condor configuration?
2. same question for 'cpus' attribute, where does it come from?
3. At which stage - exactly - Ops would override these values (RequestCpus and RequestMemory)? I believe while the job is still pending on the local schedd(?)

To be honest, I'd rather see the agent making these changes during assignment. This flexibility should be on the WMAgent hands, not on Ops, IMO. 
In addition to that, it's strongly coupled to the job splitting, which is one more reason to make this flexibility the right way to avoid future issues and making our life 'miserably' for debugging issues.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -616,6 +616,53 @@ def handleCondorStatusService(self):
...
+        Returns the resized number of cores; if resizing is disabled, returns
+        the value of the `numCores` input parameter.
+        """
+        adFile = os.environ.get('_CONDOR_MACHINE_AD')
+        if not adFile:
+            return numCores
+        resizeJob = None
+        machineCpus = None
+        try:
+            fd = open(adFile)
```
ericvaandering: Please use "with" here

bbockelm:
Hi @amaltaro -

1.  `WMCore_ResizeJob` would come from the job ad... actually, that's a bug.  I only look at the machine ad.  I'll fix that with the other items @ericvaandering mentions.
2.  `Cpus` comes from machine ad - this is set my HTCondor.
3.  Ops would decide whether they want to resize jobs while the job is still pending in the queue.  In fact, talking to Jean-Roch, he prefers the other direction - starting everything with single core and then selectively marking as multicore.  This way, there's less chance we overrun the job runtime estimates.

Basically, we, I would prefer that the WMAgent would mark the workflow as "single core but use multicore as available".  However, I think this will take some tuning to get right.  Hence, I would like to:
1.  Add this to the runtime so Ops has the ability to set "single core but use multicore as available" per-job.
2.  Figure out the correct tunings - particularly, we'll need to get `RequestMemory` and `MaxWallTimeMins` correct.
3.  Once we decide we're happy, put it back into the agent as an option at the _workflow_ level.

I don't see a strong advantage to change both the runtime code and the BossAir code in a single step.

hufnagel:
Please keep in mind that due to the way the code is written, this doesn't allow you to convert singlecore workflows to multicore. The code path you modified is completely bypassed for singlecore due to backwards compatibility. We do not not override numberOfCores for 1 for singlecore jobs, we don't do anything and rely on the singlecore CMSSW default. We either have to give up on the default and always override or we have to restructure the code a bit.

bbockelm:
@ericvaandering - can this roll in yet?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -616,6 +616,67 @@ def handleCondorStatusService(self):
...
+        try:
+            with open(jobAdFile) as fd:
+                for line in fd:
+                    line = line.strip()
+                    info = line.split("=", 1)
+                    if len(info) != 2:
+                        continue
+                    attr = info[0].strip().lower()  # Need case-insensitive compares.
+                    val = info[1].strip()
+                    if attr == 'wmcore_resizejob':
```
amaltaro: Why can't we use case sensitive? It would be much easier to map what we read here and what is added by the plugin.

ericvaandering: Yeah, this bit of code is not nice looking.  :-)

ericvaandering:
No, we can put this in next week once Seangchan is back and presumably all the tests and comments have been resolved.

ericvaandering:
(not unit tests, the current failures are noise)

ticoann:
This need to be updated current production agent, right? If Alan's test is good we will make the the new patch for 1.0.14_wmagent and apply to agent.
So for this test and for temporary, Ops will do condor_q edit fo change WMCore_ResizeJob to true for individual Job, correct?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
bbockelm: HTCondor ClassAds are not case sensitive, hence we can't use a case-sensitive compare here.

bbockelm:
@ticoann - yes, for now Ops will take responsibility for enabling `WMCore_ResizeJob`.

ericvaandering:
We can put this in a patch release when it's validated by Alan's tests. Since we're not gaining much with this there is no "drop everything" rush.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -518,3 +520,98 @@ def makeOutputTweak(outMod, job):
...
+        print("Error opening job ad:")
+        print(traceback.format_exc())
+        return
+    if machineCpus and (machineCpus > 0) and ('cores' in resources):
+        resources['cores'] = machineCpus
+    if machineMemory and (machineMemory > 0) and ('memory' in resources):
+       resources['memory'] = machineMemory
+
+
+if __name__ == '__main__':
```
amaltaro: Probably a leftover, you can remove this block

amaltaro:
I'd merge shouldResizeResources function into resizeResources, since half of that code is again in resizeResources.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -518,3 +520,90 @@ def makeOutputTweak(outMod, job):
...
+    machineMemory = None
+    try:
+        with open(machineAdFile) as fd:
+            for line in fd:
+                info = line.strip().split("=", 1)
+                if len(info) != 2:
+                    continue
+                attr = info[0].strip().lower()
+                val = info[1].strip()
+                if attr == 'cpus':
```
amaltaro: How about you make a dict of these two attributes? Less code duplication.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -518,3 +520,90 @@ def makeOutputTweak(outMod, job):
     #      output module based on the settings in the section
 
     return result
+
+
+def shouldResizeResources():
```
amaltaro: As commmented before (in the main PR), I'd suggest merging this code into resizeResources function, 90% of it is again in that func.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -518,3 +520,90 @@ def makeOutputTweak(outMod, job):
...
+    except IOError as ioe:
+        print("Error opening job ad:")
+        print(traceback.format_exc())
+        return False
+    return bool(resizeJob)
+
+
+def resizeResources(resources):
+    """
+    _resizeCores_
```
amaltaro: you meant _resizeResources_ :)

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,35 @@ def setupMonitors(self, task, wmbsJob):
                 #This should be a config section
                 monitorArgs = getattr(task.data.watchdog, monitor)
                 args = monitorArgs.dictionary_()
+                # Scale resources according to the HTCondor runtime environment.
+                origCores = 1
+                try:
+                    origCores = int(task.data.application.multicore.numberOfCores)
+                except AttributeError:
+                    print("No value for numberOfCores. Not setting")
```
amaltaro: "Not setting" --> "Setting default to 1" or so...

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,35 @@ def setupMonitors(self, task, wmbsJob):
...
+                resizeResources(resources)
+                # HTCondor doesn't explicitly scale VSize; hence we scale it
+                # according to the scaling of maxRSS between the workflow and runtime.
+                if origMaxRSS and ('maxVSize' in args):
+                    args['maxVsize'] *= resources['memory'] / origMaxRSS
+                # Scale our timeouts according to the scaling of the cores.  *NOTE*
+                # the job ad contains the real scaling here - unfortunately, it uses
+                # ClassAd expressions and we can only parse values.  Something for the
+                # future.
+                coreScaling = resources['cores'] / origCores
```
amaltaro: I'm not sure we should scale the timeout values. A 4 core job meant to run in single core mode won't run 4 times faster, I think it would stay between 3.0 and 3.5 times faster, but depending on the step it could be even worse.

amaltaro:
Finally, if it's not asking too much. A unit test to validate that function would be awesome :-)

src/python/WMCore/WMRuntime/Watchdog.py
bbockelm: So, in measurements of applications, things are sufficiently close to linear - and the timeouts are sufficiently loose - that 4.0 would be fine for a 4 core job.

I'll add an extra 60 minutes to be sure though.

bbockelm:
Ok, I think I got all your comments incorporated.  Let's see what pylint says.

I'll be happy to do a unit test ... in a separate PR (file a ticket as a reminder?).  It should actually be straightforward to do, but might first take me awhile to figure out how to do this in WMCore in the first place -- I'd rather not hold up this work for that.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -518,3 +520,91 @@ def makeOutputTweak(outMod, job):
...
+    On error, returns an empty dictionary.
+    """
+    adfile = None
+    if adname == 'job':
+        adfile = os.environ.get("_CONDOR_JOB_AD")
+    elif adname == 'machine':
+        adfile = os.environ.get("_CONDOR_MACHINE_AD")
+    else:
+        print("Invalid ad name requested for parsing: %s" % adname)
+        return {}
```
amaltaro: I think you want to move this `return {}` 2 lines below (inside the if).

bbockelm: Yup, that works too.

bbockelm: Actually - I think I want both, since the two error cases are somewhat distinct.

amaltaro:
test this please

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,37 @@ def setupMonitors(self, task, wmbsJob):
...
+                resources = {'cores': origCores}
+                origMaxRSS = args.get('maxRSS')
+                if origMaxRSS:
+                    resources['memory'] = origMaxRSS
+                # Actually parses the HTCondor runtime
+                resizeResources(resources)
+                # HTCondor doesn't explicitly scale VSize; hence we scale it
+                # according to the scaling of maxRSS between the workflow and runtime.
+                if origMaxRSS and ('maxVSize' in args):
+                    args['maxVsize'] *= resources['memory'] / origMaxRSS
```
amaltaro: Before I forget, "maxVSize" not "maxVsize"

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,37 @@ def setupMonitors(self, task, wmbsJob):
...
+                resources = {'cores': origCores}
+                origMaxRSS = args.get('maxRSS')
+                if origMaxRSS:
+                    resources['memory'] = origMaxRSS
+                # Actually parses the HTCondor runtime
+                resizeResources(resources)
+                # HTCondor doesn't explicitly scale VSize; hence we scale it
+                # according to the scaling of maxRSS between the workflow and runtime.
+                if origMaxRSS and ('maxVSize' in args):
+                    args['maxVSize'] *= resources['memory'] / origMaxRSS
```
amaltaro: I just had an example where the maxVSize was set to 30GB for single core, the job matched to 2 cpus and this value was reduced to ~3.7GB, a very likely value to be reached by the job.

I suggest we change it taking into consideration the `coreScaling` value, something like:
```
if 'maxVSize' in args:
  args['maxVSize'] *= coreScaling
```

ericvaandering: Haven't we in the past decided that VSize is worthless so we just keep it set at some very large value?

hufnagel: Correct. Tier0 sets it to 100GB. My vote is to completely remove the VSize kill code.

amaltaro: Yep, we decided not to kill anymore on VSize. We used to have 20GB per core in production.
I believe some protection is still good to have, in case things go very very bad.

ericvaandering: The problem is, I'm not sure there is a "very bad" when it comes to VSize. The application can reserve an enormous amount of memory and in 64 bit space that's probably in the EB range. But as long as it doesn't fill it with anything, RSS stays in bounds.

bbockelm: For now, I think I'll set:  "If we change max RSS for resizable jobs, disable max VSize monitoring".
- There's no clear reason why max VSize should scale with the number of CPUs.  If max vsize is 4GB for 4 cores, it is not necessarily 1GB for 1 core.
- The only alternate I can think of is take the maximum RSS and add a flat number (i.e., 2GB).

Just removing the cap feels less hacky though.

bbockelm:
As I mentioned to Alan, I verified the runtime jobs now look correct.

amaltaro:
Brian, I just took a quick look at those two *.ad files and the key names seems to be written according to what we set in PyCondor plugin. Why do you want to always lower() those keys (besides a possible change in gWMS that would break the agent...)?

For clarity reasons, I'd rather have the exact key matching over there.

bbockelm:
ClassAd keys are not case sensitive (and we don't control some of the keys, such as `memory` or `cpus`).  We have indeed seen cases where the underlying implementation changes slightly and case-sensitive comparisons break.

Doing a case-sensitive comparison for a case-insensitive language makes no sense, even if it increases readability.

amaltaro:
Ok, we have what we have then :)

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,42 @@ def setupMonitors(self, task, wmbsJob):
...
+                # we add a somewhat-arbitrarily-chosen 60 minutes (timeouts are in seconds).
+                coreScaling = resources['cores'] / origCores
+                if 'softTimeout' in args:
+                    args['softTimeout'] = int(args['softTimeout']/coreScaling) + 3600
+                if 'hardTimeout' in args:
+                    args['hardTimeout'] = int(args['hardTimeout']/coreScaling) + 3600
+                # If we did base maxRSS off the memory in the HTCondor slot, subtract a bit
+                # off the top so watchdog triggers before HTCondor does.
+                if origMaxRSS and (resources['memory'] != origMaxRSS):
+                    args['maxRSS'] = 1000*(resources['memory'] - 50) #Convert back to KB
```
amaltaro: 'memory' contains the machineMemory argument, which is what exactly? Is that the total available memory in the worker node?
We have an inconsistency between the RequestMemory ad (which changes in paces of 500MB) and this one, that is completely dependent on the machine. If we request more 500MB, then the watchdog must have an additional 500MB.

Or would the batch system kill this job in case it goes 1MB  beyond resources['memory']?

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,42 @@ def setupMonitors(self, task, wmbsJob):
...
+                # Hence, we simply remove it if we change anything about the memory.
+                if origMaxRSS and (resources['memory'] != origMaxRSS) and ('maxVSize' in args):
+                    del args['maxVSize']
+                # Scale our timeouts according to the scaling of the cores.  *NOTE*
+                # the job ad contains the real scaling here - unfortunately, it uses
+                # ClassAd expressions and we can only parse values.  Something for the
+                # future.
+                # As the scaling is not perfectly linear (efficiency is closer to 95%),
+                # we add a somewhat-arbitrarily-chosen 60 minutes (timeouts are in seconds).
+                coreScaling = resources['cores'] / origCores
```
amaltaro: Moreover, we should not change any of these watchdog parameters in cases where the # of cores is unchanged

amaltaro:
Another concern that I have is with the max/min limits defined by the agent:
https://github.com/dmwm/WMCore/blob/5f0e0751bbeb6ec6b2f7d2fdf492e5a8d55fe33c/etc/WMAgentConfig.py#L201

I've seen jobs requesting (actually matching to pilots) 60h of walltime (and could be worse). 
Can we assume that
`MaxWallTimeMins = WMCore_ResizeJob ? ( 10800 / RequestCpus + 15 ) : 2700`, if such job run in single core, the pilot lifetime would guarantee at least 10815mins (otherwise it would not match against a single core, I suppose...)?

Then another obstacle is with the Timeouts defined for the JobStatusLite:
https://github.com/dmwm/WMCore/blob/5f0e0751bbeb6ec6b2f7d2fdf492e5a8d55fe33c/etc/WMAgentConfig.py#L226

which by default considers pilot lifetime of 48h, at the best case. Of course we can tweak this number and increase it, however these settings are in an agent basis and we should target the most common lifetime out there.

src/python/WMCore/WMRuntime/Watchdog.py
bbockelm: `memory` is the amount of memory allocated to the job, not the memory in the machine itself.  It's approximately equal to the `RequestMemory` (there's various rounding stuff, but let's set that aside for now).

So, if the slot has `3` cores it might request `3500`MB RAM; correspondingly, the contents of `memory` attribute in `$_CONDOR_MACHINE_AD` should be `3500`.  The watchdog limit is then set to `3450`MB.  I believe we have set condor to not kill jobs when they exceed their requested memory.

src/python/WMCore/WMRuntime/Watchdog.py
bbockelm: If the `cores` is unchanged, the scale factor should be 1, right?

src/python/WMCore/WMRuntime/Watchdog.py
amaltaro: In order to have a better job tracking (and future debugging), how about we change `resources['memory']` from the machine memory allocated to the exact `RequestMemory` value?

On the watchdog side, it makes more sense to kill a job only after it goes *beyond* the RequestMemory, so I'd get rid of the `-50`, maybe we add 50 actually.

src/python/WMCore/WMRuntime/Watchdog.py
amaltaro: Yes, but then we are adding an hour to the timeouts. This is my point.

src/python/WMCore/WMRuntime/Watchdog.py
bbockelm: We can't use the `RequestMemory` as that isn't necessarily an integer - and we'd need a full-fledged ClassAd library to parse the attribute.

In the case where the `RequestMemory` is actually equal to the memory allocated by the batch system (common for single-core glideins), if we go above the limit, the batch system may do a hard kill - no glidein or job logs are returned.  That's why we've traditionally kept the watchdog slightly below the actual batch allocation.

bbockelm:
Correct - if the lifetime on a single core works out to be 60 hours, you need a pilot with at least 60 hours remaining (actually, the way things are setup, probably closer to 72 hours).

While all this is good to keep in mind, it's also a driving reason why we plan to inject as single-core and increase the core count (rather than decreasing).  Hence, the jobs should always get shorter - it avoids some of the job lifetime issues.

src/python/WMCore/WMRuntime/Watchdog.py
amaltaro: Nope, the watchdog was always kept slightly above the Memory request. Of course it does not mean it was always correct.

On one hand, we don't want to let the batch system kill these jobs before the agent. On the other, it sounds weird that a job was meant to use up to 3000MB but we kill it even before the requirements.

We could add the 500MB extra/less here, but then it's one more place that we would need to synchronize in case we change that default. So I guess we keep it as it is

src/python/WMCore/WMRuntime/Watchdog.py
amaltaro: @bbockelm I still would like to see a change where *nothing* gets changed in the watchdog in cases where there is no job resizing. Can you push this `coreScaling = resources['cores'] / origCores` line above and just `continue` the loop in case there was no change?

amaltaro:
As an additional comment, there will be cases not behaving well and taking too much resources. E.g.:
```
Resizing job.  Initial resources: {'cores': 8}
Resizing job.  Resulting resources: {'cores': 1}
...
2016-04-19 16:45:19,654:INFO:Watchdog:  maxRSS: 4558000
2016-04-19 16:45:19,654:INFO:Watchdog:  softTimeout: 1040400
2016-04-19 16:45:19,654:INFO:Watchdog:  hardTimeout: 1048400
```

It was originally requesting 8GB of RAM (for 8 cores), then running in single core it requests 4.5GB.
This job actually surprises me that it found glideins providing ~1M secs of lifetime (?) Is it something that we should worry about?

bbockelm:
Yup - there's upper-level logic needed to avoid somewhat-hopeless cases like the one above.

Indeed, there are glideins (particularly, at CERN) that last for 30 days.

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,46 @@ def setupMonitors(self, task, wmbsJob):
...
                 args = monitorArgs.dictionary_()
+                # Scale resources according to the HTCondor runtime environment.
+                origCores = 1
+                for stepName in task.listAllStepNames():
+                    sh = task.getStepHelper(stepName)
+                    origCores = max(origCores, sh.getNumberOfCores())
+                resources = {'cores': origCores}
+                origMaxRSS = args.get('maxRSS')
+                if origMaxRSS:
+                    origMaxRSS = int(origMaxRSS/1000.) # HTCondor expects MB; we get KB.
```
amaltaro: To avoid future confusion, how about we adopt powers-of-two for the `origMaxRSS`, as we do for all the rest of the WMCore code (I think). Same comment valid for ~30 lines below.

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,46 @@ def setupMonitors(self, task, wmbsJob):
...
+                    resources['memory'] = origMaxRSS
+                # Actually parses the HTCondor runtime
+                resizeResources(resources)
+                # We decided to only touch Watchdog settings if the number of cores changed.
+                # (even if this means the watchdog memory is wrong for a slot this size).
```
amaltaro: Perhaps we should then change responsibilities on who controls the maxRSS? Instead of letting the assigner (Unified) set it, we just leave it to HTCondor. What do you think?

bbockelm: I think there's already plenty of things going on in this patch already ... let's refactor that later.

amaltaro:
@bbockelm squash please

amaltaro:
@ticoann do it! :-P ready to be merged

amaltaro:
@ticoann sorry for messing with yes and no... but we better keep this and the condor resize PR on hold.
I added you to a private email reporting issues in frontend.

amaltaro:
superseeded by #7363

PR #6758, opened 2016-03-29
Import fix for CRAB3

emaszs:
Similar to #6745, need to import httplib2 because otherwise we don't have it.


emaszs:
Sorry, it's an incomplete PR, please ignore.

PR #6767, opened 2016-03-31
Fix print regressions

ericvaandering:
Had a couple of regressions of our rules on print function vs. statement. I will try to adjust Jenkins to find these.


ericvaandering:
@ticoann please merge

PR #6773, opened 2016-04-04
1.0.14_crab version of Fix logic for CondorStatusService availability check

emaszs:
There was an issue discovered during validation that the CondorStatusService is being added incorrectly. Discussion can be found here https://github.com/dmwm/CRABServer/issues/5186#issuecomment-204324695. Since the service is available from CMSSW_7_6_0 onwards (https://github.com/dmwm/CRABServer/issues/5186#issuecomment-204381799), a fix is needed. After the merge we would also like a new tag, 1.0.14_crab_5 for our branch.


ericvaandering:
You need to fix the print (or remove them).

emaszs:
Interesting. Removed the print.

ericvaandering:
Ah, sorry, I looked too briefly. The print was not your problem so you should not have had to fix it. But thanks anyhow. (it stuck around from the commit you branched from).

amaltaro:
@emaszs tag 1.0.14_crab_5 has been just created

PR #6777, opened 2016-04-05
Fix logic for CondorStatusService availability check

emaszs:
Same PR as for our 1.0.14_crab branch (#6773).

There was an issue discovered during validation that the CondorStatusService is being added incorrectly. Discussion can be found here https://github.com/dmwm/CRABServer/issues/5186#issuecomment-204324695. Since the service is available from CMSSW_7_6_0 onwards (https://github.com/dmwm/CRABServer/issues/5186#issuecomment-204381799), a fix is needed.


hufnagel:
You probably want to close this in favor of #6778

emaszs:
Ah, understood. Thanks for pointing it out.

PR #6778, opened 2016-04-05
tag chirp updates from CMSSW, fixes #6739

### LINKED ISSUE #6739: Tag chirp updates from CMSSW
The `CondorStatusService` can add a tag to all of its update attributes (i.e., `ChirpCMSSWMaxEvents` becomes `ChirpCMSSWDIGIMaxEvents`).  This can be used to allow HTCondor to record statistics for each step in a stepchain separately.  This is enabled by setting the `tag` to an appropriate untracked string (see https://github.com/cms-sw/cmssw/blob/7e2655ffaf3e2a7fa9ab823e788d5059c4e2353d/FWCore/Services/plugins/CondorStatusUpdater.cc#L422).

Without this, only the last step in a stepchain is recorded.

hufnagel:
Tag the CMSSW chirp updates with the step name to support multi-step workflows. Disable chirp updates for releases before CMSSW76X since these didn't support the tagging mechanism.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -604,15 +604,13 @@ class SetupCMSSWPset(ScriptInterface):
                 elif int(result.group(1)) == 7:
                     if int(result.group(2)) >= 6:
                         addCondorStatusService = True
-                    elif int(result.group(2)) == 5 and int(result.group(3)) >= 1:
-                        addCondorStatusService = True
-                    elif int(result.group(2)) == 4 and int(result.group(3)) >= 7:
-                        addCondorStatusService = True
             except ValueError:
                 pass
 
         if addCondorStatusService:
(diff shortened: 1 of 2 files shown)


ericvaandering:
Please add an issue for the far future to remove this check when CMSSW 7 is no longer used.

hufnagel:
Why should we remove the check ? It's harmless (although a bit ugly, I give you that).

ericvaandering:
For that reason only. (Back in the days of CRAB2 and rapidly changing CMSSW, we had a bit of a spaghetti code even with removing truly obsolete version.)

amaltaro:
Yes, let us avoid keeping spaghetti in the code :-D

hufnagel:
Ok, put in an issue and this code is ok now too, unit tests are happy. Please consider merging.

ericvaandering:
Spaghetti is best confined to a dinner plate. Not great in code or on your shirt.

ticoann:
merging it, only affects agent side

PR #6779, opened 2016-04-05
CRAB3 is failing untar for older cmssw releases

juztas:
Before we were using python2.6 and now with this enforcement to python2 makes it worse and fails with old releases (as example tested with CMSSW_5_3_22)

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
@@ -357,15 +357,22 @@ export SCRAM_ARCH=$SCRAM_ARCHIT
 echo "Retrieving SCRAM project..."
 # do the actual executing
 $SCRAM_COMMAND project $SCRAM_PROJECT $CMSSW_VERSION
-if [ $? -ne 0 ]; then echo "Scram failed"; exit 71; fi
+EXIT_STATUS=$?
+if [ $EXIT_STATUS -ne 0 ]; then echo "Scram failed with exit code: $EXIT_STATUS"; exit 71; fi
 cd $CMSSW_VERSION
-if [ $? -ne 0 ]; then echo "***\nCouldn't chdir: $?\n"; exit 72; fi
+EXIT_STATUS=$?
+if [ $EXIT_STATUS -ne 0 ]; then echo "***\nCouldn't chdir: $EXIT_STATUS\n"; exit 72; fi
 
 eval `$SCRAM_COMMAND runtime -sh`
-if [ $? -ne 0 ]; then echo "***\nCouldn't get scram runtime: $?\n*"; exit 73; fi
+EXIT_STATUS=$?
+if [ $EXIT_STATUS -ne 0 ]; then echo "***\nCouldn't get scram runtime: $EXIT_STATUS\n*"; exit 73; fi
 if [ -n "$USER_TARBALL" ] ; then
     python2 -m WMCore.WMRuntime.Scripts.UnpackUserTarball $USER_TARBALL $USER_FILES
-    if [ $? -ne 0 ]; then echo "***\nCouldn't untar sandbox: $?\n"; exit 74; fi
+    EXIT_STATUS=$?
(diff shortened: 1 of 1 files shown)


juztas:
This is what @jmarra13 reported during CRAB3 meeting. after change of his jobs to use this behave all seems to be fine. I think there still should be some backward compatibility to allow users to run with old releases.

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -365,7 +365,18 @@ def post(self, emulator=None):
 if [ $? -ne 0 ]; then echo "***\nCouldn't get scram runtime: $?\n*"; exit 73; fi
```
juztas: There are more problems, line like this Couldn't get scram runtime: $?\n will return `Couldn't get scram runtime: 0`. 

`To check the exit code we can simply print the $? special variable in bash. This variable will print the exit code of the last run command.` In this case last run command is not eval....

ericvaandering:
So versions of python for old CMSSW releases don't have python2 defined but python26 is defined?

juztas:
There is something more fishy... When I checked it was using system python, while I would think it should use cmssw, so python2 failed to `import os` and whenever I added this more complex if it succeeded with python26. I could even print more env if needed for more debug.

ticoann:
test this please

amaltaro:
> Before we were using python2.6 and now with this enforcement to python2 makes it worse and fails with old releases (as example tested with CMSSW_5_3_22)

Are you sure it isn't a site issue? In our battery of tests, we run workflows using 5_3_16, 5_3_19, 7_5_1, etc etc and I didn't spot a single failure.

juztas:
In that case mostly all T2s are with this bug :) 
```
======== CMSSW OUTPUT STARTING ========
NOTICE: lines longer than 3000 characters will be truncated
== CMSSW: Beginning CMSSW wrapper script
== CMSSW:  slc6_amd64_gcc472 scramv1 CMSSW
== CMSSW: Performing SCRAM setup...
== CMSSW: Completed SCRAM setup
== CMSSW: Retrieving SCRAM project...
== CMSSW: ***
== CMSSW: Couldn't untar sandbox: 0
== CMSSW:
======== CMSSW OUTPUT FINSHING ========
```
And I was connected on WN and tested this patch for specific task which uses CMSSW_5_3_22. I think we are affected due to fact that we are in mixed environment. WMA mostly moved to python2.7 everywhere, CRAB3 is still has in some place to use. For this specific part again CRAB3 and WMA does differently before, if I read both codes correctly. 

Is there anything what you afraid of this? If you succeed to untar with python2 it will not affect prod. If you fail, you will try with python2.6. There are more things with respect to env and move to python2.7 solve in CRAB3, but this requires more time to validate from Jadir. 

I could create a patch only to CRAB3 branch in that case until some scram archs/python fixes are done in CRAB3. Let me know ;)

amaltaro:
> I could create a patch only to CRAB3 branch in that case until some scram archs/python fixes are done in CRAB3. Let me know ;)

I'd rather not over-complicate things in WMCore unless it's completely necessary.

So yes, try to fix/validate it in CRAB3 land/branch only please. If it cannot be fixed there, well, then we have no choice but to patch it in WMCore.

ericvaandering:
I don't want CRAB3 only versions of WMCore. We are just getting out of that mess. 

I don't understand why this bug exists, but if it does, and CRAB3 needs a change to WMCore to solve it, then we need to have it in the WMCore mainline.

juztas:
Hi Eric, I proposed to have same in main branch (It should not affect production tarball untar) until we solve this issue with CRAB3. As a quick discussion with @belforte we need for CRAB3 to define all list of issues related to python migration as it is not done, mainly schedd/WN and on WN use COMP python and not CMSSW on some places. Also environment initialization can be in one place. Now there is 3 diff places with almost same code: WMA, CRAB3, SAM...
Maybe Stefano will comment more ;)

belforte:
good point. Seems an item for Eric to ponder. I am off this afternoon, sorry.

PR #6785, opened 2016-04-06
validate location from fwrj

ticoann:
just prevent situation happened using T0_CERN_CH

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -79,7 +79,8 @@ class AccountantWorker(WMConnectionBase):
         self.getJobInfoByID          = self.daofactory(classname = "Jobs.LoadFromID")
         self.getFullJobInfo          = self.daofactory(classname = "Jobs.LoadForErrorHandler")
         self.getJobTaskNameAction    = self.daofactory(classname = "Jobs.GetFWJRTaskName")
-
+        self.pnn_to_psn              = self.daofactory(classname = "Locations.GetPNNtoPSNMapping").execute()
+        
         self.dbsStatusAction       = self.dbsDaoFactory(classname = "DBSBufferFiles.SetStatus")
         self.dbsParentStatusAction = self.dbsDaoFactory(classname = "DBSBufferFiles.GetParentStatus")
         self.dbsChildrenAction     = self.dbsDaoFactory(classname = "DBSBufferFiles.GetChildren")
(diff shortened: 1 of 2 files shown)


ticoann:

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -794,7 +795,13 @@ def handleWMBSFiles(self, wmbsFilesToBuild, parentageBinds):
                 runLumiBinds.append({'lfn': lfn, 'runs': wmbsFile['runs']})
 
             if len(wmbsFile.getLocations()) > 0:
-                fileLocations.append({'lfn': lfn, 'location': wmbsFile.getLocations()[0]})
+                outpnn = wmbsFile.getLocations()[0]
+                if self.pnn_to_psn.get(outpnn, None):
```
amaltaro: None is redundant here.

amaltaro:
Code looks Ok, I'm just not sure whether we should really break the component if this case happens again. I'd instead print a warning and fail the job (possible adding a new error/message in WMExceptions).

ticoann:
I think we should crash the component. If job runs somewhere we don't have the record of, 
1. it is completely screwed up
2. Or we don't have the proper site record in our table.
I think it would be mostly 2 case (although proper way is to have SSB record), then we need add the site in the table manually and restart the component.

If we just print error and move on, we need to do manual surgery when case 2 cause the problem.
I think it actually shouldn't happen (at least very often). But it happens it might be easier to fix by crashing Component

ticoann:
I will update the patch to fix unit tests.

src/python/WMComponent/JobAccountant/AccountantWorker.py
ticoann: Yes, I just make consistent with other code in the module. :smile:

amaltaro:
> If we just print error and move on, we need to do manual surgery when case 2 cause the problem.

we do not move on. We permanently fail this job, report failures to wmstats and someone  fix either a) the assignment or some black magic script that was screwing with jobs or b) fix resource control/sitedb.

I'm just trying to get rid of ALL the unneeded component crashes.
On the other hand, I agree that this should happen not more than once/twice per year, so it's ok'ish to crash the component if you prefer (at least we would be able to make a hot fix and not waste cpu time)

ticoann:
> e do not move on. We permanently fail this job, report failures to wmstats and someone fix either a) the assignment or some black magic script that was screwing with jobs or b) fix resource control/sitedb.

I see, I think a) case it makes sense to failed the jobs  but b) case, it maybe better to fix resource and not redo the whole processing.  I am guessing most case will be b). but I might be wrong.

amaltaro:
Patch looks good, though I'm running a couple quick workflows just to make sure nothing breaks.
If it's Ok, I think this one deserves to be backported to 1.0.14

amaltaro:
@ticoann feel free to merge this one

ticoann:
test this please

PR #6786, opened 2016-04-06
Change PyCondorPlugin to be able to dynamically resize jobs.

bbockelm:
This changes `Cpus`, `Memory`, and `MaxWallTimeMins` to be appropriate for any number of cores that might be available at matchmaking time.

For these to be effective, one will need to set `WMCore_ResizeJob` to be `True` in the ad.  This defaults to `False` right now and would have to be enabled by hand.  A separate update will allow this to be specified within the workflow.


bbockelm:
@ericvaandering @amaltaro here's the next piece of the resizable jobs.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -267,7 +267,7 @@ def __init__(self, config):
             self.proxy = self.setupMyProxy()
 
         # Build a request string
-        self.reqStr = "(Memory >= 1 && OpSys == \"LINUX\" ) && (Arch == \"INTEL\" || Arch == \"X86_64\") && stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS==\"any\") || (GLIDEIN_REQUIRED_OS==REQUIRED_OS))"
```
amaltaro: Hmm, I'd expect any pilots to provide at least 1MB of memory, any motivation to explicitly ask for it?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +972,29 @@ def customizePerJob(self, job):
 
         # Performance estimates
         if job.get('estimatedJobTime'):
-            jdl.append('+MaxWallTimeMins = %d\n' % int(job['estimatedJobTime'] / 60.0))
+            estimatedMins = int(job['estimatedJobTime'] / 60.0)
+            origCores = int(job.get('numberOfCores', 1))
+            estimatedMinsSingleCore = estimatedMins*origCores
```
amaltaro: Don't you want to make a division instead?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +972,29 @@ def customizePerJob(self, job):
...
+            jdl.append('+MaxWallTimeMins = WMCore_ResizeJob ? (%d/RequestCpus + 60) : %d\n' % (estimatedMinsSingleCore, estimatedMins))
+            
         if job.get('estimatedMemoryUsage'):
-            jdl.append('request_memory = %d\n' % int(job['estimatedMemoryUsage']))
+            extraMem = int(job.get('extraMemoryPerCore', 500))
```
amaltaro: Where we define extraMemoryPerCore? Perhaps we should make it configurable in the agent setup - for the moment - since it may fluctuate quite a bit at this stage of multicore workflows. Moreover, the current 4 cores jobs usually request 6GB, so I'd default to 1GB instead.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +972,29 @@ def customizePerJob(self, job):
...
-        if job.get('numberOfCores', 1) > 1:
-            jdl.append('machine_count = 1\n')
-            jdl.append('request_cpus = %s\n' % job.get('numberOfCores', 1))
+        jdl.append('machine_count = 1\n')
+        minCores = int(job.get('minCores', 1))
+        maxCores = int(job.get('maxCores', 4))
+        origCores = int(job.get('numberOfCores', 1))
+        jdl.append('+MinCores = %d\n' % minCores)
+        jdl.append('+MaxCores = %d\n' % maxCores)
+        jdl.append('+RequestCpus = WMCore_ResizeJob ? (isUndefined(Cpus) ? MinCores : ((Cpus>MaxCores) ? MaxCores : Cpus)) : %d\n' % origCores)
```
amaltaro: I have to be honest, I failed to expand this (quaternary?) expression. How is (isUndefined(Cpus) evaluated?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
ericvaandering: I presume you are calculating minutes only to improve the match making, but please just convert to minutes at the end. Also, make sure any division is going to also give reasonable values in python3

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
ericvaandering: this would be a good place for a comment explaining what's going on for those not indoctrinated to the HTCondor high priesthood.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
ericvaandering: Brian is preparing us for the future where where this is given us by McM and passed through request manager. If you want to make this a configuration parameter (not a bad idea) you can make a PR against Brian's branch.

hufnagel: 500MB extra per core sounds about right and is based on feedback from the CMSSW team. Tier0 uses 2500 + (n-1) * 500MB and we never had problems in (proton proton) PromptReco. HeavyIon is a different matter though and I would expect some upgrade workflows might need more too.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: This is now automatically added by condor.  In fact, having `Memory >= 1` causes `condor_submit` to throw a warning and condor will ignore this.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +972,29 @@ def customizePerJob(self, job):
...
+            estimatedMins = int(job['estimatedJobTime'] / 60.0)
+            origCores = int(job.get('numberOfCores', 1))
+            estimatedMinsSingleCore = estimatedMins*origCores
+            jdl.append('+MaxWallTimeMins = WMCore_ResizeJob ? (%d/RequestCpus + 60) : %d\n' % (estimatedMinsSingleCore, estimatedMins))
+            
         if job.get('estimatedMemoryUsage'):
-            jdl.append('request_memory = %d\n' % int(job['estimatedMemoryUsage']))
+            extraMem = int(job.get('extraMemoryPerCore', 500))
+            baseMem = int(job['estimatedMemoryUsage'])
+            jdl.append('+RequestMemory = WMCore_ResizeJob ? (%d + RequestCpus*%d) : %d\n' % (baseMem, extraMem, baseMem))
```
hufnagel: As I said during your presentation, this should be

jdl.append('+RequestMemory = WMCore_ResizeJob ? (%d + (RequestCpus-1)*%d) : %d\n' % (baseMem, extraMem, baseMem))

since baseMem will already include the relevant extraMem portion for a singlecore job.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: Nope - this is the number of minutes this is estimated to run on one core.  For example, if a job estimate is provided of 60 minutes on 2 cores, then `estimatedMinSingleCore` should be 120.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: Sure, that's fine.  I would like to see this as configurable per agent - or, rather, in the workflow definition.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: Or saying it another way: If you allow job resizing and match to a single core, your memory request should be baseMem and not baseMem+extraMem.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +972,35 @@ def customizePerJob(self, job):
 
         # Performance estimates
         if job.get('estimatedJobTime'):
-            jdl.append('+MaxWallTimeMins = %d\n' % int(job['estimatedJobTime'] / 60.0))
+            estimatedMins = int(job['estimatedJobTime'] / 60.0)
```
ericvaandering: Actually this is the only division in the file. Please add "from __future__ import division" to the top so this is one less thing we have to audit in a few years.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: It looks like I commented the removal part thinking you were adding it :-)

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -1001,15 +1008,21 @@ def customizePerJob(self, job):
...
         # for monitoring and accounting.  Defaults to 0; once matched, it'll
         # put an attribute in the job  MATCH_EXP_JOB_GLIDEIN_Cpus = 4
         jdl.append('+JOB_GLIDEIN_Cpus = "$$(Cpus:0)"\n')
         # Make sure the resize request stays within MinCores and MaxCores.
         jdl.append('+RequestResizedCpus = (Cpus>MaxCores) ? MaxCores : ((Cpus < MinCores) ? MinCores : Cpus)\n')
+        # Advertise the original CPU setting, in case someone needs this for monitoring
+        jdl.append('+OriginalCpus = %d\n' % origCores)
+        # If the job is running, then we should report the matched CPUs in RequestCpus - but only if there are sane
+        # values.  Otherwise, we just report the original CPU request
+        jdl.append('+JobCpus = ((JobStatus =!= 1) && (JobStatus =!= 5) && !isUndefined(MATCH_EXP_JOB_GLIDEIN_Cpus) && (int(MATCH_EXP_JOB_GLIDEIN_Cpus) isnt error)) ? int(MATCH_EXP_JOB_GLIDEIN_Cpus) : OriginalCpus\n')
```
amaltaro: `isnt error` is it right? This is a nasty expression

bbockelm: Yeah - I don't want things to go haywire because some future person sets the value to "foo" instead of "1".  If it's not parseable, we don't want to use its value.

You can now see why I broke up the logic for `RequestCpus` into several expressions :)

bbockelm:
As I mentioned to Alan, the test jobs look fine.  It's ready from my side.

ticoann:
Thank you very much Brian. Could you squash the commit? 
We will create the wmagent branch patch and if Alan gives green light as well, we will deploy to production agent

bbockelm:
@ticoann @amaltaro - commit squashed.

ericvaandering:
Where is the logic that prevents this from being applied to cleanup jobs? And how are we eventually going to have this optionally put in depending on a request setting?

bbockelm:
@ericvaandering - well, we hardcode resizing off for the time being for _all_ jobs.  Differentiating job types - and making decisions on a request-by-request basis - will be part of a separate set of work.

ericvaandering:
So right now, we put all of this in the JDL for all job types but we (externally) guarantee that we don't turn it for cleanup jobs.

bbockelm:
Correct!

When testing, I've purposely been sloppy and turn on resizing for cleanup jobs to verify mistakes don't have any particularly bad issues (other than the loss in efficiency).

ericvaandering:
OK. Thanks.

amaltaro:
test this please

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +975,59 @@ def customizePerJob(self, job):
 
         # Performance estimates
         if job.get('estimatedJobTime'):
-            jdl.append('+MaxWallTimeMins = %d\n' % int(job['estimatedJobTime'] / 60.0))
+            estimatedMins = int(job['estimatedJobTime'] / 60.0)
+            origCores = int(job.get('numberOfCores', 1))
+            estimatedMinsSingleCore = estimatedMins*origCores
+            # For now, assume a 15 minute job startup overhead -- condor will round this up further
+            jdl.append('+EstimatedSingleCoreMins = %d\n', estimatedMinsSingleCore)
+            jdl.append('+OriginalMaxWallTimeMins = %d\n', estimatedMins)
```
amaltaro: Interesting that we did not spot it before... Replace `,` by `%` please.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -972,16 +975,59 @@ def customizePerJob(self, job):
...
+            jdl.append('+MaxWallTimeMins = WMCore_ResizeJob ? (EstimatedSingleCoreMins/RequestCpus + 15) : OriginalMaxWallTimeMins\n')
+            
         if job.get('estimatedMemoryUsage'):
-            jdl.append('request_memory = %d\n' % int(job['estimatedMemoryUsage']))
+            baseMem = int(job['estimatedMemoryUsage'])
+            jdl.append('+OriginalMemory = %d\n', baseMem)
```
amaltaro: Interesting that we did not spot it before... Replace `,` by `%` please.

amaltaro:
@bbockelm squash please

amaltaro:
@bbockelm sorry Brian, it seems you need to rebase this PR and then it's ready to be merged

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -982,16 +985,59 @@ def customizePerJob(self, job):
...
+        # Set up JDL for multithreaded jobs.
+        # By default, RequestCpus will evaluate to whatever CPU request was in the workflow.
+        # If the job is labelled as resizable, then the logic is more complex:
+        #  - If the job is running in a slot with N cores, this should evaluate to N
+        #  - If the job is being matched against a machine, match all available CPUs, provided
+        #    they are between min and max CPUs.
+        #  - Otherwise, just use the original CPU count.
+
+        jdl.append('machine_count = 1\n')
+        minCores = int(job.get('minCores', 1))
```
amaltaro: @bbockelm as discussed in todays meeting, we shouldn't have a full range of resizing from multicore to 1 core.
So the initial idea is to resize down to half of the original number of cores (8 cores job should run with 4 cores, at least).
Can you please change this line to something like
`minCores = int(job.get('minCores', max(origCores/2, 1)))`

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -982,16 +985,59 @@ def customizePerJob(self, job):
...
+            estimatedMinsSingleCore = estimatedMins*origCores
+            # For now, assume a 15 minute job startup overhead -- condor will round this up further
+            jdl.append('+EstimatedSingleCoreMins = %d\n' % estimatedMinsSingleCore)
+            jdl.append('+OriginalMaxWallTimeMins = %d\n' % estimatedMins)
+            jdl.append('+MaxWallTimeMins = WMCore_ResizeJob ? (EstimatedSingleCoreMins/RequestCpus + 15) : OriginalMaxWallTimeMins\n')
+            
         if job.get('estimatedMemoryUsage'):
-            jdl.append('request_memory = %d\n' % int(job['estimatedMemoryUsage']))
+            baseMem = int(job['estimatedMemoryUsage'])
+            jdl.append('+OriginalMemory = %d\n' % baseMem)
```
amaltaro: And Farrukh just mentioned Merge jobs don't carry OriginalMemory classAd, now I see why.

Perhaps we should always set these classAds to a reasonable number when estimated numbers are not found, like 1250 MB

amaltaro:
Superseeded by #7362

PR #6798, opened 2016-04-12
Untar with python2.6 for older release

juztas:
Patch to CRAB3 branch only until CRAB3 will finish all movement to python2.7 and also fix environment issues.

PR #6799, opened 2016-04-12
Make sure we print correct exit code in the log files

juztas:
In light of: https://github.com/dmwm/WMCore/pull/6779

This is just to make print correct exit code for specific command which was executed.


juztas:
@amaltaro after your comments I created two pull requests:
one for CRAB3 until all movement to python2.7 is solved and also which python to use where. https://github.com/dmwm/WMCore/pull/6798/files
second which is this one is to print correct exit code.

juztas:
Added exit Codes and messages, whenever it will be merged, will update it on twiki.

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -357,15 +357,22 @@ def post(self, emulator=None):
 echo "Retrieving SCRAM project..."
 # do the actual executing
 $SCRAM_COMMAND project $SCRAM_PROJECT $CMSSW_VERSION
-if [ $? -ne 0 ]; then echo "Scram failed"; exit 71; fi
+EXIT_STATUS=$?
+if [ $EXIT_STATUS -ne 0 ]; then echo "Scram failed with exit code: $EXIT_STATUS"; exit 71; fi
```
amaltaro: Why not checking `$?` directly? As it was before.

juztas:
```
$ ls bla
ls: cannot access bla: No such file or directory
$ if [ $? -ne 0 ]; then echo "Command Failed: $?"; fi
Command Failed: 0
$ ls bla
ls: cannot access bla: No such file or directory
$ EXIT_STATUS=$?
$ if [ $EXIT_STATUS -ne 0 ]; then echo "CommandFailed: $EXIT_STATUS"; fi
CommandFailed: 2
```

amaltaro: Hmm, interesting :) I don't like repeating this `EXIT_STATUS=$?` line 1 million times, but at least it's very clear.

PR #6809, opened 2016-04-18
add defualt ProcessingString(Version)

### LINKED ISSUE #6384: ReqMgr2 -> request info without ProcessingString or Processing version
@ticoann @amaltaro 
When an acdc is created, the information provided in this API:
https://cmsweb.cern.ch/reqmgr2/data/request?name=jbadillo_ACDC_Run2015BHINPFJet100-16Oct2015_7414_151106_101855_6024
Comes without ProcessingString or ProcessingVersion.
In contrast with the old API:
https://cmsweb.cern.ch/reqmgr/reqMgr/request?requestName=jbadillo_ACDC_Run2015BHINPFJet100-16Oct2015_7414_151106_101855_6024

ticoann:
fixes #6384

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -8,7 +8,7 @@ import logging
 
 from WMCore.Cache.WMConfigCache import ConfigCache, ConfigCacheException
 from WMCore.Configuration import ConfigSection
-from WMCore.Lexicon import lfnBase, identifier, acqname, cmsswversion, cmsname, couchurl, block
+from WMCore.Lexicon import lfnBase, identifier, acqname, cmsswversion, cmsname, couchurl, block, procstring
 from WMCore.Services.Dashboard.DashboardReporter import DashboardReporter
 from WMCore.WMSpec.WMSpecErrors import WMSpecFactoryException
 from WMCore.WMSpec.WMWorkload import newWorkload
@@ -939,8 +939,8 @@ class StdBase(object):
                      "ScramArch": {"default": "slc5_amd64_gcc462", "optional": False},
                      "GlobalTag": {"null": True},
                      "GlobalTagConnect": {"null": True},
-                     "ProcessingVersion": {"default": 1, "type": int},
-                     "ProcessingString": {"null": True},
+                     "ProcessingVersion": {"default": 1, "type": int, "assign_optional": False},
+                     "ProcessingString": {"default": "", "validate": procstring, "assign_optional": False},
                      "LumiList": {"default": {}, "type": makeLumiList},
                      "SiteBlacklist": {"default": [], "type": makeList,
                                        "validate": lambda x: all([cmsname(y) for y in x])},
--- a/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrWebTools.py
+++ b/src/python/WMCore/HTTPFrontEnd/RequestManager/ReqMgrWebTools.py
@@ -421,6 +421,8 @@ def buildWorkloadAndCheckIn(webApi, reqSchema, couchUrl, couchDB, wmstatUrl, clo
                       "SizePerEvent",
                       "PrepID",
                       "RequestNumEvents",
+                      "ProcessingString",
+                      "ProcessingVersion",
                       ]
     
     couchDb = Database(reqDetails["CouchWorkloadDBName"], reqDetails["CouchURL"])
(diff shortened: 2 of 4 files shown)


src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -939,8 +939,8 @@ def getWorkloadArguments():
                      "ScramArch": {"default": "slc5_amd64_gcc462", "optional": False},
                      "GlobalTag": {"null": True},
                      "GlobalTagConnect": {"null": True},
-                     "ProcessingVersion": {"default": 1, "type": int},
-                     "ProcessingString": {"null": True},
+                     "ProcessingVersion": {"default": 1, "type": int, "assign_optional": False},
+                     "ProcessingString": {"default": "None", "validate": procstring, "assign_optional": False},
```
amaltaro: I suggest we default it to the same default value as AcquisitionEra, so "FAKE". Easier to keep track of what to expect on the assignment side.

ticoann: Thanks Alan, updated

ticoann:
@amaltaro, Alan, please review. It sets default value for both reqmgr1 and reqmgr2. (This sets all other default values for reqmgr2 as well - not sure it is desired but more consistent). Tested in my vm.

ticoann:
test this please

ticoann:
Also ProcessingString is mandatory now. (Is there is case ProcessingString is skipped?)

amaltaro:
Yes, I agree we should make AcqEra and ProcStr mandatory, at least.
However, we have one specific case (ACDC workflows for TaskChain type, merge task) that we cannot provide AcqEra and ProcStr, that's how an issue was fixed in the past and it's in my TODO list to change that... :(

hufnagel:
ProcessingString is optional according to the naming convention. RAW datasets don't use a processing string (there is no processing there, so no ProcessingString).

amaltaro:
But this case applies only to T0, right? And since T0 does not use ReqMgr, I do think we should enforce it (but then, we of course need to fix the TaskChain case I mentioned before)

hufnagel:
Yes, only applies to Tier0. ReqMgr only checks to make it mandatory would work I guess, changing the Lexicon checks or anything deeper that is also triggered by Tier0 is not fine.

ticoann:
Although ProcessString is checked in ReqMgr, it sets up the default value, when it is not set. I will change back to not have default value. Which means it won't show in creation time, or I can have empty string as default value. (Alan, do what is the exact problem, if the process string is not showing before it gets assigned?)

ticoann:
@amaltaro, Alan for making mandatory for these doesn't actually check whether each Task should have these fields set. It only checks in request level, But I guess this is a still problem for ACDC for TaskChain (for merge task). However aquisitionera is always mandatory since it was updated few months ago. Is that causing the problem?

amaltaro:
Hmm, I initially thought AcqEra was not mandatory for Resubmission (it's not in the spec), but looking at a few workflows in production, they all have it (and they were not assigned yet).

Giving a second thought to it, I don't think we can enforce it at request level, since TaskChain requests could carry this information at the task level only. OR, we do enforce it at request level and make sure that McM or others always provide it at request AND task level, to avoid issues.

> However aquisitionera is always mandatory since it was updated few months ago. Is that causing the problem?

It would be for the ACDC case mentioned before, but in the end it's not really enforced because I was working on a script with Paola in the last week and we did not provide those keys (acqera and procstr) at all.

ticoann:
Alan, they are not enforced on creation but in assignment. (But I think this is enforced only for reqmgr2 request).

ticoann:
Change the default value for ProcessingString so Tier0 won't have the FAKE in the request name.

amaltaro:
I was thinking about this issue and, actually I think we should start enforcing these strings during request creation AND disable the override during assignment (our validation code works only for creation anyways). Another argument is that McM/Offline is defining the output dataset names for quite some time already. A third argument is that it's annoying to read the original values and to provide them again during assignment.

@ericvaandering , Seangchan, what do you think? I'm not saying we have to fix it right now, but maybe for the next month (thus, you don't spend time on it trying to solve it in a different way).

ticoann:
If ProcessingString and AcquisitonEra is defined by McM, I agree we should enforce this in creation. The fix won't be difficult to add (only for reqmgr2)

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

PR #6820, opened 2016-04-22
add requestor view

ticoann:
Jean-Roch's request. query using status and requestor (doesn't allow only by requestor which could return to many data) Also, archived status is not allowed.
@amaltaro, Alan please review (tested in vm)

### DIFF
--- a/src/couchapps/ReqMgr/views/bystatusandrequestor/map.js
+++ b/src/couchapps/ReqMgr/views/bystatusandrequestor/map.js
@@ -1,0 +1,3 @@
+function(doc) {
+  emit([doc.RequestStatus, doc.Requestor], null);
+}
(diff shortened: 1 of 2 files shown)


amaltaro:
It looks good Seangchan.

However, I couldn't find an issue for it, I thought he initially wanted to pass only the requestor name.
Since it does not return value, do you still think it would be too heavy to return archived requests as well (not the whole doc)?

ticoann:
Hi Alan, if it just return request names archived request query will be fine. 
@vlimant, Jean-Roch, is it OK to return just request names if for querying archived request?
That means it will only accept detail = False if status contains archived status

ticoann:
Thanks, Alan, I update the patch so reqmgr2 will accept the query containing status=*-archived only detail=false explicitly.

PR #6822, opened 2016-04-22
No need these deps in apmon

juztas:
Tested with crab3: http://dashb-cms-job.cern.ch/dashboard/templates/task-analysis/#user=crab&refresh=0&table=Jobs&p=1&records=25&activemenu=2&status=&site=&tid=160422_074307%3Ajbalcas_crab_testapmon

### DIFF
--- a/src/python/WMCore/Services/Dashboard/DashboardAPI.py
+++ b/src/python/WMCore/Services/Dashboard/DashboardAPI.py
@@ -3,9 +3,6 @@
 """
 This is the Dashboard API Module for the Worker Node
 """
-from __future__ import absolute_import
-from __future__ import print_function
-
 from WMCore.Services.Dashboard import apmon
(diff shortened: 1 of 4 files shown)


amaltaro:
Justas, all the new modules should have these imports
`
`

This PR goes on the other direction that we want to go. Including defining classes on the old way.

juztas:
So in this case you need to have future package on WN

amaltaro:
WMAgent certainly ships it with the job. If CRAB does not, then we have to solve it in a different way instead of demoting the code to python 2.4.

juztas:
Ok, tested now. Missread what was the problem and it is not future. Problem is builtins and past.utils. 
for crab it is another problem, so not for this discussion. 

Will create new one. Thanks Vago ;)

PR #6825, opened 2016-04-22
No need these deps in apmon

juztas:


juztas:
@amaltaro can you take a look ?

ticoann:
test this please

src/python/WMCore/Services/Dashboard/ProcInfo.py
```diff
@@ -32,17 +32,14 @@
...
-from past.utils import old_div
 from WMCore.Services.Dashboard.Logger import Logger
 import socket
 import os
 import re
 import time
 
 
-class ProcInfo(object):
+class ProcInfo:
```
ticoann: If you want you can still new class inheriting object without importing builtins object. But it is up to you.

ticoann:
Justas, Ideally it would be better to ship all the dependencies on WorkerNode. (in this case future package). But for now this would be fine. Thanks

amaltaro:
Justas, I'm in favor of using the new style class definition, so
`class Logger(object):`

as Seangchan said, there is no need to import from builtins.
If it's using the old style, ok'ish not to update it. But if it has the new style and you make it old, then ... well, not good.

amaltaro:
@juztas if you have time, can you please update the class definition only? Otherwise, Seangchan will fix it in another commit in an hour or two.

juztas:
@amaltaro have you did that already? if not I will change it now. Please, let me know and sorry for late reply

amaltaro:
Nope, otherwise this one would be merged

ticoann:
@juztas, Don't worry about it. I will create additional patch. Thanks,

PR #6828, opened 2016-04-23
Increase the WQE size for MC policy

amaltaro:
Since we just had problems aborting a monstrous workflow (that brought central couchdb down twice - so far), let's increase the WQE size for MonteCarlo workflows such that we can cut the amount of docs by 4.

@ticoann besides agents acquiring more work than they should, do you foresee any problem with this change?


ticoann:
Hi Alan, I think it was set like that to spread out the work. I am not sure whether this is relevant now. 
If the site white list is set for a specific site, it will require more slots open to pull the work.

amaltaro:
test this please

PR #6834, opened 2016-04-26
Remove WMStats js views that are available in erlang already

amaltaro:
These views were made available in erlang 2 or 3 months ago
https://github.com/dmwm/WMCore/tree/master/src/couchapps/WMStatsErl/views

and everything works fine, so I suggest we remove their javascript versions.


ticoann:
test this please

PR #6835, opened 2016-04-26
Change fileset creation to allow multiple output from StepChain

### LINKED ISSUE #6787: Stepchain: sort out duplicate outputModules
As Stepchain is now, it does not support KeepOutput:True for different steps using exactly the same outputModule (not matter the datatier used). The problem is that merge jobs are created for all files created under a specific outputModule (including all datatiers).

amaltaro:
Once this patch is complete, it will fix #6787 

@ticoann as we discussed yesterday, this is my current NON-fully working patch. 

It seems to be working for all type of requests, however there is still a problem that I didn't manage to fix yet. For this template https://github.com/dmwm/WMCore/blob/master/test/data/ReqMgr/requests/DMWM/StepChain_MC.json ,
it creates 2 merge jobs for 'DIGIPROD1MergeRAWSIMoutput' and another 2 merge jobs for 'ProdMinBiasMergeRAWSIMoutput'.

Needless to say, do NOT merge it :-D

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -16,7 +16,6 @@ import threading
 import logging
 import gc
 import collections
-from pprint import pformat
 
 from WMCore.FwkJobReport.Report  import Report
 from WMCore.DAOFactory           import DAOFactory
@@ -560,7 +559,6 @@ class AccountantWorker(WMConnectionBase):
         expected output map for that specific job
         """
         jobSuccess = True
-        logging.info(pformat(fileList))
 
         # there are a few special cases that we just pass as success
         outputModules = set([])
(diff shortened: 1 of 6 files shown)


ticoann:
Thanks Alan, I will try to understand the patch. :-)

amaltaro:
Only look at it if you don't have anything better to do :-)
The most important changes are:
 1. in WMBSHelper, changing the name of the fileset to append the datatier.
 2. JobAccountant, to properly check the output mapping with what came back from the job.

Changes are very easy... but this part of the code is quite complex :(

amaltaro:
Problem solved with the duplicated merge jobs (and dup records in workflow and fileset). Now onto the next problem :) why the logcollect and cleanup workflow entries were not created

amaltaro:
New PR created in #7998

PR #6836, opened 2016-04-26
Translated two more WMStats views to Erlang

amaltaro:
Making this PR on behalf of @BrunoCoimbra , thanks vagabundo!

Bruno has performed a validation/diff of the output between these new views and the old in javascript and everything is Ok. However, @ticoann would you mind to do another x-check please? :-D

If everything is Ok, then we could push it to testbed this week.

@ticoann besides the frontend map rules, is there anything else we need to update?


ticoann:
test this please

Issue #6837, opened 2016-04-27
Write WMStats allWorkflows view in erlang


amaltaro:
This is the last bit missing to have all the WMStats views written in erlang.
https://github.com/dmwm/WMCore/tree/master/src/couchapps/WMStats/views/allWorkflows

Since @BrunoCoimbra agreed to take over this task... have fun! :-D

PR #6839, opened 2016-04-27
Include WMStats allWorkflows view in erlang

BrunoCoimbra:
This PR addresses the issue #6837.

I translated the WMStats view allWorkflows to Erlang.
After a diff comparison, everything seems to be working just as it should. :)
@amaltaro and @ticoann, could you please double check?


ticoann:
Thanks Bruno

ticoann:
Test this please

ticoann:
Actually this view is used by agent which doesn't have Erlang version set for couch app.  So I will hold on to this until we change the agent to set up the couch app.

ticoann:
test this please

PR #6841, opened 2016-04-29
Remove logging lines hitting couch

amaltaro:
There was this stupid guy nicknamed amaltaro that forgot these 2 logging lines in the code. Thanks to that brilliant mind, couchdb logs were increasing 100GB per day..

@ticoann please add it to Today's testbed tag

PS.: Diego removed these two lines in vocms0140 and vocms0131 already.


ticoann:
Thanks Alan, you need to work less and relax more. :smile:

PR #6848, opened 2016-05-04
implement repack output straight to error dataset

hufnagel:
Repack jobs stageout directly to merged for files larger than 2GB. If these files are larger than 12GB we do not want them to be included in the regular RAW output dataset and we also don't want them to be prompt reconstructed. Therefore implement an override in the JobAccountant that changes the DBS Primary Dataset for such files and also skips adding them to the filesets used for PromptReco.

And yes, it's a bit hacky, but it's the only way I could think of to make this work in a half-way sane way.

@ticoann @amaltaro please review


ticoann:
So this big file still need to uploaded to dbs with new dataset (-Error)?

hufnagel:
Yeah, we don't want to completely drop it (users should still be able to find it), we just don't want it in the normal output dataset as the PhEDEx subscriptions are different.

PR #6849, opened 2016-05-04
Changing to xrd rm for the delete method, since there is no NFS mount…

DAMason:
This fixes the cleanup jobs in the FNAL stageout plugin.  Functionally tested successfully on a test agent.

### DIFF
--- a/src/python/WMCore/Storage/Backends/FNALImpl.py
+++ b/src/python/WMCore/Storage/Backends/FNALImpl.py
@@ -12,6 +12,7 @@ import logging
 from WMCore.Storage.Registry import registerStageOutImpl
 from WMCore.Storage.StageOutImpl import StageOutImpl
 from WMCore.Storage.Backends.LCGImpl import LCGImpl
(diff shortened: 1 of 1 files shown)


ticoann:
test this please

ticoann:
From Dirk's comment.

Please don't introduce a dependency on RFCPCERNImpl . If there is code in that plugin that is useful in more than one place it should be factored out and put into some base class somewhere. RFCPCERNImpl will likely go away at some point (whenever I find some free time sigh) since it's a horrible mixed Castor/EOS stageout plugin for a site that only uses EOS these days.

@hufnagel, Thanks Dirk. I will fix remove the dependency. Since I have to update the production, wmagent version will be merged but will fix the master.

amaltaro:
@DAMason if you don't manage to look at it in the coming days, then I'll jump into it in the next week.

amaltaro:
@amaltaro reminder to myself, it needs to be backported to the new wmagent branch

amaltaro:
@DAMason thanks for this fix. I created another patch based on Dirk's suggestions, see #6915 
I'm closing this one.

PR #6853, opened 2016-05-05
Changing to xrd rm for the delete method, since there is no NFS mount…

ticoann:
… to rm from anymore. Introduces a dependency on RFCPCERNImpl to steal splitPFN. Functionally tested successfully.


hufnagel:
Please don't introduce a dependency on RFCPCERNImpl . If there is code in that plugin that is useful in more than one place it should be factored out and put into some base class somewhere. RFCPCERNImpl will likely go away at some point (whenever I find some free time sigh) since it's a horrible mixed Castor/EOS stageout plugin for a site that only uses EOS these days.

PR #6856, opened 2016-05-11
fix priority problem

ticoann:
attempt to fix the priority problem (lower priority taking over the slots) - need to be tested

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -75,7 +75,9 @@ class JobSubmitterPoller(BaseWorkerThread):
         self.changeState = ChangeState(self.config)
         self.repollCount = getattr(self.config.JobSubmitter, 'repollCount', 10000)
         self.maxJobsPerPoll = int(getattr(self.config.JobSubmitter, 'maxJobsPerPoll', 1000))
-        self.cacheRefreshSize = int(getattr(self.config.JobSubmitter, 'cacheRefreshSize', 10000))
+        self.cacheRefreshSize = int(getattr(self.config.JobSubmitter, 'cacheRefreshSize', 30000))
+        self.skipRefreshCount = int(getattr(self.config.JobSubmitter, 'skipRefreshCount', 20))
+        self.refreshPollingCount = 0
 
         # BossAir
         self.bossAir = BossAirAPI(config=self.config)
@@ -253,10 +255,13 @@ class JobSubmitterPoller(BaseWorkerThread):
 
         logging.info("Querying WMBS for jobs to be submitted...")
         logging.info("cachedJobIDs : %s" % len(self.cachedJobIDs))
-        if len(self.cachedJobIDs) < self.cacheRefreshSize:
+        if self.cacheRefreshSize == -1 or len(self.cachedJobIDs) < self.cacheRefreshSize or \
+           self.refreshPollingCount >= self.skipRefreshCount:
             newJobs = self.listJobsAction.execute()
+            self.refreshPollingCount = 0
             logging.info("Found %s new jobs to be submitted.", len(newJobs))
         else:
+            self.refreshPollingCount += 1
             newJobs = []
             dbJobs = self.cachedJobIDs
             logging.info("Skipping cache update to be submitted. (%s job in cache)" % len(dbJobs))
@@ -425,7 +430,6 @@ class JobSubmitterPoller(BaseWorkerThread):
                             except KeyError:
                                 # Already gone
                                 pass
-
         logging.info("Done pruning killed jobs, moving on to submit.")
         return
 
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -209,7 +209,8 @@ config.JobSubmitter.pollInterval = 120
 config.JobSubmitter.workerThreads = 1
 config.JobSubmitter.jobsPerWorker = 100
 config.JobSubmitter.maxJobsPerPoll = 1000
-config.JobSubmitter.cacheRefreshSize =10000 # minimum size of the cache before refresh. prevent frequent cache update
+config.JobSubmitter.cacheRefreshSize = 30000 # set -1 if cache need to refresh all the time.
+config.JobSubmitter.skipRefreshCount = 20 # (If above the threshold meet, cache will updates every 20 polling cycle) 120 * 20 = 40 minutes
 config.JobSubmitter.submitScript = os.path.join(os.environ["WMCORE_ROOT"], "etc/submit.sh")
 
 config.component_("JobTracker")
--- a/src/couchapps/WorkQueue/views/availableByPriority/map.js
+++ b/src/couchapps/WorkQueue/views/availableByPriority/map.js
@@ -1,9 +1,6 @@
 function(doc, site) {
     var ele = doc["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
     if (ele && ele["Status"] === "Available") {
-        // TODO: Make priority modifier configurable
-        // modifier so 1 hour queue time == +1 priority boost
-        var priority = ele["Priority"] - (doc.timestamp * (1. / 60 / 60))
-        emit(priority, {"_id" : doc["_id"]});
+        emit(ele["Priority"], {"_id" : doc["_id"]});
     }
 }
--- a/src/python/WMCore/WorkQueue/WorkQueueBackend.py
+++ b/src/python/WMCore/WorkQueue/WorkQueueBackend.py
@@ -326,6 +326,7 @@ class WorkQueueBackend(object):
         self.logger.info("Getting available work from %s/%s" %
                          (sanitizeURL(self.server.url)['url'], self.db.name))
         elements = []
+        sortedElements = []
 
         # We used to pre-filter sites, looking to see if there are idle job slots
         # We don't do this anymore, as we may over-allocate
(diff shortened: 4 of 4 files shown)


ticoann:
Looking at the code again. Actually this might not be the problem,  this patch won't change any functional behavior. I think problem might be in JobSubmitter code. (I can see if there are too many jobs created and high priority job is created after lower priority job is created, high priority jobs might not get to submitted first)

ticoann:
Test this please

ticoann:
test this please

ticoann:
@amaltaro, Alan, please review.  We need to have better solution for JobSubmitter cache but. I am not sure what can we do about this.

amaltaro:
As a general comment, refreshing the job submitter cache all the time is bad (and makes caching useless for this component :D). In cases the agent has >10k (or much more like 50k) jobs queued, this can take from several minutes to an hour or two.

Cache should be updated only when really needed (something in the cached jobs needs to be updated - sitewhitelist, priority, etc)

hufnagel:
Is refreshing the cache really needed or was this only done here because we don't understand how low priority jobs can jump the queue on high priority jobs ?

ticoann:
I don't know where this (not refreshing cache) is actually the problem or not. 
One case I can think of which could affect is the same case that Tier0 problem we had a couple of days ago (Express jobs are not submitted)

Here is the case.
1.  Jobs are created much more (30K) with lower priority, than JobSubmitter Cache threshold  (10K),
2.  When JobSubmitter cache is refreshed cache is filled with all 30K jobs (threshold means cache refresh threshold but there is no limit on cache size, it will take all created jobs in wmbs)
3.  Higher priority workflow is assigned  - this will still get to WMAgent and create the new jobs,
4.  However those jobs are not submitted until JobSubmitter cache is refreshed. (20K jobs need to be submitted before it gets refreshed).

Bottom line is JobSubmitter has its own queue (cache) so that could prevent higher priority jobs gets submitted first.
The problem is it take long time to refresh cache if there are too many jobs in created status.
I am not sure how we can solve this other than refreshing cache for each cycle. (Since we don't know whether new jobs created or not with higher priority or not)

ticoann:
test this please

amaltaro:
As we discussed in yesterdays meeting, the ideal would be to control how many jobs the agents creates, to make sure the agent cache won't grow above a specific number.

In addition to that, we could have another polling cycle in JobSubmitter that would refresh the cache every hour, this should be frequent enough to make everyone's happy.

ticoann:
Hi Alan, yes I agree. I will update the patch. I am just sending jenkins test message since it seems not working. (Merged build finished. No test results found.)

ticoann:
Test this please

ticoann:
test this please

amaltaro:
@ticoann I don't think limiting the # of jobs created by the JobCreator will be very simple.

And, since I'd like to have this patch in in the next WMAgent release candidate (tomorrow maybe), how about
 a) never refresh the cache if above 50k cached jobs
 b) either b1) refresh the cache every hour OR b2) refresh the cache every 20 cycles (~45min)

No hurry though, we can get it done by tomorrow :-P

hufnagel:
@amaltaro , the problem with these proposals is that in the Tier0 we need to guarantee that high priority jobs are submitted first, no matter how large the backlog in low priority jobs is. So any arbitrary cutoff is kind of unacceptable. Delaying things somewhat (not too long) is ok, stopping to accept new jobs and evaluating their priority compared to older jobs is not.

amaltaro:
Ok, then let me rephrase one of them
 a) never refresh the cache if above 50k (default setting) cached jobs, however keep it configurable such that Tier0 can set it to 0 (so, always refresh it).

hufnagel:
Why is this a problem really ? Once you are over a certain threshold, switch to evaluate the priority of he highest priority lets say 1k jobs in the cache. If your new jobs have higher or same priority add them to the cache (as the assumption is they should run soon) and if not don't add them.

hufnagel:
Ok, this doesn't take into account that jobs have different possible sites, but you can't have everything. At least you guarantee this way that high priority jobs go out first and aren't superseeded by lower priority jobs just because your agent is overloaded.

hufnagel:
Number can be tweaked of course, but the principle should work. Just needs two things:

A fast way to count jobs by priority we have in the cache.

Selectively being able to just add jobs with the highest priorities to the cache.

Should be doable, shouldn't it ?

ticoann:
How about the following.
We set the threshold for delay of refresh.
1. set threshold 30K.
2. If cash is more than a threshold updated cache every 20 polling cycle until size is smaller than threshold.

Or we can do a little bit more fine control by increase time proportionally  (10K - 20K 10 polling cycle, 20K - 30K, 20 polling cycle, etc)

amaltaro:
Seangchan, IMO it makes no real difference whether it's 10 or 20 polling cycle. So I suggest we keep it as simple as possible and just use a X (30k sounds Ok) threshold and update it every Y polling cycles (20 sounds Ok).

> A fast way to count jobs by priority we have in the cache.

This is kind of a new feature. I'd rather have it in another PR to get this one done and merged asap.

> Selectively being able to just add jobs with the highest priorities to the cache.

Only highest priority jobs get created, so in principle there is no effective way to do this. If there is, I'd rather keep it in another PR as well.

ticoann:
> A fast way to count jobs by priority we have in the cache.
It seems filling up the cache take more time than sorting the priority if the cache size if big. It is loading pickled file to get some information (location, etc)

> Selectively being able to just add jobs with the highest priorities to the cache.
This is tricky. We have to consider not just priority but also sites jobs can run with we get it from pickle file,  so it won't be just done by db call.

> Should be doable, shouldn't it ?
I think it is but may not be simple.

hufnagel:
If you set some sort of threshold and don't add jobs into the cache, you by design don't consider certain jobs for running. In such a system lower priority jobs in the cache can drown out higher priority jobs that are stuck in the db and not in the cache.

Corect, if you just add high priority jobs from the db to the cache, you will consider them for running preferentially, but at the expense of possible lower priority jobs that might have been able to run at sites where the higher priority jobs can't run.

But we are discussing here what to do in an overload situation. You pick your poison and I'd rather have looked at the high priority jobs and decided they can't run because of site restrictions than not to have looked at all and just ran FIFO.

Anyways, if you think it's too much for this PR please open another issue. But any kind of hard cutoff without taking into account priorities will inevitably result in some sort of complaint from Ops at some point: Why are all these low priority jobs being submitted if we have high priority work in the system ?

amaltaro:
Either I misunderstood this fix, or you got lost on the discussion.
I'm not saying we are not going to update the cache. I'm not saying that low priority jobs should run in front of the high prio either.

The original PR would update the cache ALL the time, such that high prio jobs *always* run in front of the low prio.

I'm just suggesting to not update it ALL the time, but instead define a polling cycle for these updates in case the cache is too big.

Having said that, the worst case scenario is that low prio jobs will run in front of the high prio within a cycle (less than an hour). I was hoping it would not be that bad to Tier0... (?) Certainly is not for central production.

hufnagel:
I understand. Was just some brain storming to see if there is a smarter alternative to "do it less often". Something like "only do the parts that are most important".

If we get a configurable timeout to reduce the impact of large job counts without affecting submission too badly (Tier0 we can probably live with 30 min cycle, 1h is stretching it) that would be fine for now.

Would still like to see an issue where we could maybe have some smarter way in the future...

ticoann:
@amaltaro, Alan, updated the patch and create the new issue #6864 
Updated patch is not test I will run the new test. 
I keep the 2 commit - one for WQ side change and the other for JobSubmitter change.
I will create the new tag tonight. I want to include one more patch for the new agent test. 
Please review.
Thanks,

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -254,10 +256,13 @@ def refreshCache(self):
 
         logging.info("Querying WMBS for jobs to be submitted...")
         logging.info("cachedJobIDs : %s" % len(self.cachedJobIDs))
-        if len(self.cachedJobIDs) < self.cacheRefreshSize:
+        if self.cacheRefreshSize == -1 or len(self.cachedJobIDs) < self.cacheRefreshSize or \
```
amaltaro: this check is not really needed `self.cacheRefreshSize == -1`

amaltaro:
Besides the last comment, patch looks good.
FOI., mysql agents seems to take only (?I'm surprised...) ~10secs to refresh 5k jobs. Oracle agents seem to take a second or two only (astonished, again...)

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: Just added in case someone wants to refresh every cycle. But we can remove that

ticoann:
I am not sure what would be the optimal number even with 30K tier0 didn't take long (Oracle) < 1min.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
amaltaro: In this case we just need to set 'cacheRefreshSize' to 0. It would refresh all the time :-)

amaltaro:
Relax, 30k is a good number in general ;)

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: Actually, if  it is set to 0, it will never get to be refreshed. since cache size has to go lower than the  cacheRefreshSize to be refreshed. :smile:

ticoann:
test this please

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
amaltaro: Stupid me! :-D

PR #6860, opened 2016-05-13
Fix site whitelist when provided in the command line

amaltaro:
Sorry, I removed a line that I shouldn't in the last commit.


amaltaro:
already tested.

PR #6861, opened 2016-05-13
Add glidein restrictions tweak for the relval agent

amaltaro:
Still need to be tested.


amaltaro:
Just updated the site thresholds as well, since we've the condor overflow mechanism in place, having more jobs in the local condor queue is useful (these thresholds are already in production for quite a bit)

amaltaro:
I also added the bash script to restart ErrorHandler (while we sort out that issue).
It's tested, so I'm merging it.

PR #6873, opened 2016-05-20
1.0.16_crab version of Use system environment for gfal2 stageout

emaszs:


mmascher:
That has already been mergen on master, but after the branch was made (where after refers to git history)

PR #6876, opened 2016-05-20
clean up status transition

ticoann:
@amaltaro, Alan please review. Mostly for preventing move to failed status automatically


src/python/WMCore/RequestManager/RequestDB/Settings/RequestStatus.py
```diff
@@ -35,19 +30,14 @@
...
- "running-open" : ["running-closed", "aborted", "failed", "epic-FAILED"],
- "running-closed" : ["force-complete", "completed", "aborted", "failed", "epic-FAILED"],
+ "acquired" : ["running-open", "completed", "acquired", "aborted"],
+ "running" : ["completed", "aborted"],
+ "running-open" : ["running-closed", "aborted"],
+ "running-closed" : ["force-complete", "completed", "aborted"],
  "force-complete" : ["completed"],
- "failed" : ["failed", "testing-approved", "assigned", "rejected"],
- "epic-FAILED" : ["epic-FAILED"],
+ "failed" : ["failed", "assigned", "rejected"],
```
amaltaro: how about we remove "failed" from "failed" status?

src/python/WMCore/RequestManager/RequestDB/Settings/RequestStatus.py
```diff
@@ -35,19 +30,14 @@
...
- "testing" : ["testing", "tested", "test-failed", "aborted"],
- "tested" : ["tested", "assignment-approved","failed", "rejected", "aborted"],
- "test-failed" : ["test-failed", "testing-approved", "rejected", "aborted"],
  "assignment-approved" : ["assignment-approved", "assigned", "rejected"],
  "assigned" : ["assigned","acquired", "aborted", "failed"],
- "acquired" : ["running-open", "failed", "completed", "acquired", "aborted"],
- "running" : ["completed", "aborted", "failed", "epic-FAILED"],
- "running-open" : ["running-closed", "aborted", "failed", "epic-FAILED"],
- "running-closed" : ["force-complete", "completed", "aborted", "failed", "epic-FAILED"],
+ "acquired" : ["running-open", "completed", "acquired", "aborted"],
```
amaltaro: Hmmm, I'm not really sure whether this is a legit use case... Can't an "acquired" workflow move to "failed" according to its input or some other constraint?

amaltaro:
Seangchan, do you know whether this state transition is done on GQ or on the agent side?
If it's on the agent side, then we'll be able to debug it in the next time it happen.

src/python/WMCore/RequestManager/RequestDB/Settings/RequestStatus.py
ticoann: I remember there was some problem doing that in ReqMgr2. I am not sure what it was. But for the safe side I left it there.

src/python/WMCore/RequestManager/RequestDB/Settings/RequestStatus.py
ticoann: I don't think anybody move the failed status manually. I think Ops use only reject an abort. Failed status moves automatically if the work is not pulled for a long time by workqueue, The problem is there is the case when only partial work is done and other element got stuck somewhere it still have possibility moving to failed status automatically

ticoann:
Alan, I will merge this and see whether it causes any further problem I haven't expected.

PR #6879, opened 2016-05-23
Update memory and cores requirements during assignment

amaltaro:
As we @ticoann @ericvaandering discussed some time ago, these are the required change to make "Multicore" and "Memory" parameters overwritable during assignment.

It's still not clear to me whether we should also update the reqmgr2 couch doc (reqmgr_workload_cache). If we do, then the original values are gone and any clones would use the new values. I'm inclined not to... I think (?)

Initial tests work Ok in my VM, though I still need to check the reqmgr2 part.


amaltaro:
Ok, all 4 use cases tested:
 1) update cores and memory in both reqmgr and reqmgr2
 2) cores and memory untouched during assignment  in both reqmgr and reqmgr2
 3) only cores updated  in both reqmgr and reqmgr2
 4) only memory updated in both reqmgr and reqmgr2

The job splitting was not touched and it would need to be manually changed, if needed.

amaltaro:
@ticoann btw, it's good to go from my side

PR #6885, opened 2016-05-24
Allow to use callables as configuration params

mmascher:
That's a very useful feature imho that exploits the true potential of having a configuration in python.

In particular I would like to have as a parameter of CRAB the function we use to pick up the schedd depending on parameters like memory usage, running jobs etc.

### DIFF
--- a/test/python/WMCore_t/Configuration_t.py
+++ b/test/python/WMCore_t/Configuration_t.py
@@ -7,10 +7,50 @@ import os
 
 from WMCore.Configuration import ConfigSection
 from WMCore.Configuration import Configuration
+from WMCore.Configuration import ConfigurationEx
 from WMCore.Configuration import loadConfigurationFile
 from WMCore.Configuration import saveConfigurationFile
 
-from WMQuality.TestInitCouchApp import TestInitCouchApp as TestInit
+from WMQuality.TestInit import TestInit
+
+
+class ConfigurationExTest(unittest.TestCase):
+    """
+    test case for Configuration object
+
+    """
+    def setUp(self):
+        """set up"""
+        self.testInit = TestInit(__file__)
+        self.testDir  = self.testInit.generateWorkDir()
+        self.functionSave = "%s/WMCore_Agent_Configuration_t_function.py" % self.testDir
+
+
+    def tearDown(self):
+        """clean up"""
+        self.testInit.delWorkDir()
+
+
+    def testCallableConfigParams(self):
+        """ctor"""
+        def f():
+            return True
+
+        config = Configuration()
+        config.section_("SectionF")
+        #creating field for the following test
+        config.SectionF.aFunction = ''
+        #Cannot set a function for plain Configuration objects
+        #config.SectionF.__setattr__('aFunction', f)
+        self.assertRaises(RuntimeError, config.SectionF.__setattr__, config.SectionF.aFunction, f)
+
+        config = ConfigurationEx()
+        config.section_("SectionF")
+        #No failures with configurationEx
+        config.SectionF.aFunction = f
+
+        #However ConfigurationEx instances cannot be saved
+        self.assertRaises(RuntimeError, saveConfigurationFile, config, self.functionSave)
 
 
 class ConfigurationTest(unittest.TestCase):
@@ -119,7 +159,7 @@ class ConfigurationTest(unittest.TestCase):
         self.assertRaises(
             RuntimeError, setattr,
             config.Section2, "BadDict", badDict)
-        
+
         goodDict = { "dict" : {}, "list": [], "tuple" : () }
         config.Section2.GoodDict = goodDict
 
--- a/test/python/WMCore_t/Configuration_t.py
+++ b/test/python/WMCore_t/Configuration_t.py
@@ -3,7 +3,6 @@
 
 
 import unittest
-import os
 
 from WMCore.Configuration import ConfigSection
 from WMCore.Configuration import Configuration
@@ -75,7 +74,7 @@ class ConfigurationTest(unittest.TestCase):
     def testA(self):
         """ctor"""
         try:
-            config = Configuration()
+            dummyConfig = Configuration()
         except Exception as ex:
             msg = "Failed to instantiate Configuration\n"
             msg += str(ex)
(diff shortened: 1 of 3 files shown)


ticoann:
Thanks Marco, would you mind updating pylint suggestion as well? It seems callable was removed in Python 3 - 3.1 but reinserted in Python 3.2. (I guess it would be fine since we are moving to latest Python3 in future.) 😄  @ericvaandering, Eric, do you have any comment on that as an Python 3 expert?

ericvaandering:
Yeah, we will never run Python 3.0 to 3.4.

ericvaandering:
Question on how this works, though. Don't we JSONize these files? How do you do that for a function? Anyhow, this case should have a unit test that demonstrates the functionality.

amaltaro:
@mmascher since the 1.0.16_crab version has been merged already, can you follow up on Eric's and Seangchan's comments please? Otherwise we may eventually forget to fix/merge it and CRAB will have issues again in the next cycle. Thanks

mmascher:
I improved the pull request taking care of all the comments

mmascher:
All the tests here are failing with:

```
Creating ApMon with static configuration
Wed Jun  8 15:12:50 2016: ApMon[ERROR]: Cannot open 127.0.0.1:8884
Wed Jun  8 15:12:50 2016: ApMon[ERROR]: IOError: [Errno 2] No such file or directory: '127.0.0.1:8884'
Wed Jun  8 15:12:50 2016: ApMon[WARNING]: Not sending parameters since no destination is defined.
Wed Jun  8 15:12:50 2016: ApMon[DEBUG]: Closing UDP socket on ApMon object destroy.
```

I don't think my changes can have an impact on that, could it be a fake positive (for this PR)?

juztas:
@mmascher go deeper, this is not an error and these messages are ignored.

mmascher:
Thanks @juztas , now I know I have to ignore those.

I think I know what is going on, thare are some configuration saved for testing that does not have the new skipChecks flag set, and when they are loaded you get a: `'ConfigSection' object has no attribute'_internal_skipChecks'`

I will fix this.

BTW, I also see some `Invalid WMSpec: 'ReRecoWorkload': Input data has no locations "/MinimumBias/ComissioningHI-v1/RAW"` messages, but I guess I can ignore them as well.

amaltaro:
In general we should only ignore tests marked as 'unstable'. If there are more tests out there unstable, I suggest to update our text file (if we have a good reason to let them unstable) such that people don't get in doubt about them.

About the `invalid wmspec`, I thought it was fixed with the last mock changes...

ericvaandering:
Of course, one has to verify that the test is unstable. Looking at the history in https://github.com/dmwm/ContinuousIntegration/issues/7 is the best way to do this. Free beer to anyone who permanently fixes one of these unstable tests.

mmascher:
Uhm.. now the only test failing is `testStripReport`.

The test load a report.xml and saves the report using Report.save. then it checks the size of the saved report with an hardcoded value: 7101. I think hardcoding such a value is pretty bad programming, and I also think this is a pretty useless check! Personally I'd like to remove this check in the test, what's the point of checking the exact size of the saved report, what it this test accomplishing by doing that? Apart that every time somebody do even a small change (as I did) you'd have to update these hardcoded size values...

Here there are the relevan portion of code and the stacktrace:

```python
    def testStripReport(self):
        """
        _testStripReport_

        Test whether or not we can strip input file information
        from a FWJR and create a smaller object.
        """

        myReport = Report("cmsRun1")
        myReport.parse(self.xmlPath)

        path1 = os.path.join(self.testDir, 'testReport1.pkl')
        path2 = os.path.join(self.testDir, 'testReport2.pkl')

        myReport.save(path1)
        info = BasicAlgos.getFileInfo(filename=path1)
        self.assertEqual(info['Size'], 7101)
```

```
  File "/build/cmsbuild/jenkins-workarea/workspace/DMWM-PR-unittests/SLICE/3/label/slc6_amd64-cloud-tiny/deploy/1.0.17.pre6-comp/sw/slc6_amd64_gcc493/external/python/2.7.11-comp/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/build/cmsbuild/jenkins-workarea/workspace/DMWM-PR-unittests/SLICE/3/label/slc6_amd64-cloud-tiny/code/test/python/WMCore_t/FwkJobReport_t/Report_t.py", line 683, in testStripReport
    self.assertEqual(info['Size'], 7101)
  File "/build/cmsbuild/jenkins-workarea/workspace/DMWM-PR-unittests/SLICE/3/label/slc6_amd64-cloud-tiny/deploy/1.0.17.pre6-comp/sw/slc6_amd64_gcc493/external/python/2.7.11-comp/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/build/cmsbuild/jenkins-workarea/workspace/DMWM-PR-unittests/SLICE/3/label/slc6_amd64-cloud-tiny/deploy/1.0.17.pre6-comp/sw/slc6_amd64_gcc493/external/python/2.7.11-comp/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
'7447 != 7101\n--------------------
```

ericvaandering:
Well, the good news is that the test has a good description which is that the striped report is smaller than the regular report. So instead of 7201 and 6210 later, just change the test so that the value reported 2nd is smaller than the first. 

Yes, many of our tests really aren't testing what we want to. We need to iteratively improve these tests as we find them.

mmascher:
Thanks for the suggestion Eric, I will do it. Like you say makes more sense and is more robust.

test/python/WMCore_t/Configuration_t.py
```diff
@@ -7,10 +7,50 @@
...
+        self.testDir  = self.testInit.generateWorkDir()
+        self.functionSave = "%s/WMCore_Agent_Configuration_t_function.py" % self.testDir
+
+
+    def tearDown(self):
+        """clean up"""
+        self.testInit.delWorkDir()
+
+
+    def testA(self):
```
ericvaandering: Can you name your test something more descriptive? When it fails, it would be nice to have a description and easily findable string to speed up debugging.

amaltaro:
@mmascher can you please update this PR as per Eric's suggestions?
Make sure to squash it as well :-D (unit tests and pylint fixes can be left in a separate commit)

mmascher:
Done and squashed. I left `Improve testStripReport unit test` in a single commit because it is improving a failing unit test completely unrelated to my PR.

PR #6887, opened 2016-05-24
Fix visible attributes assignment on web UI for reqmgr2

vkuznet:


ticoann:
Thanks Valentin, I tested the patch and worked fine.

PR #6888, opened 2016-05-24
add job detail REST api

### LINKED ISSUE #6858: aggregate error information per wf from wmstats
in order to get faster a full picture on whether a workflow has to be recovered or kill-cloned.
what I think we need is, for a given workflow (or subtask) is the list sites with error, error code, and the text attached to it (something very close to the workflow summary page, but with reliable information)

 @prozober @jenimal please update on what else is needed

ticoann:
fixes #6858

### DIFF
--- a/src/python/Utils/IterTools.py
+++ b/src/python/Utils/IterTools.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python
 from __future__ import division, print_function
 
+import collections
 from itertools import islice
 from itertools import chain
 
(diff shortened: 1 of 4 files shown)


ticoann:
test this please

ticoann:
test this please

ticoann:
Merging this for now (tested in vm) for testbed test. If Jean-Roch and Paola needs different return format or information I will update that in next round.

PR #6893, opened 2016-05-26
Refactor SiteDB into two parts to make it easy to mock

ericvaandering:
This is a change that will be part of @jha2 's branch for mocking SiteDB, so don't merge this. However you can look at it. The point is to split SiteDB into two parts. One which actually talks to SiteDB and the other which parses the dictionaries it returns. This makes it possible to use mock to just replace the lower level and leave the upper level alone.

@ticoann and @amaltaro , please feel free to review/comment.

### DIFF
--- a/src/python/WMCore/Services/SiteDB/SiteDB.py
+++ b/src/python/WMCore/Services/SiteDB/SiteDB.py
@@ -2,76 +2,26 @@
 """
 _SiteDB_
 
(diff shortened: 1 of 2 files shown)


amaltaro:
@ericvaandering this one needs to be rebased (sorry!) Should we push it in for tomorrow's testbed as well?

ericvaandering:
This was already part of Manoj's PR.

PR #6895, opened 2016-05-27
switch logcollect stagein to use edmCopyUtil, fixes #6740

### LINKED ISSUE #6740: Rewrite LogCollect jobs to use edmCopyUtil
Hi,

`edmCopyUtil` (added for CMSSW_8_0_x) takes a list of LFNs and downloads them to the local disk.  For example,

```
edmCopyUtil /store/log/foo1 /store/log/foo2 /store/log/foo3 .
```

Will copy the three files to the current directory.  This uses the currently configured LFN access method in `site-local-config.xml` instead of using SRM to download the files.

If we use this for log collect jobs, it'll be a huge load off the site SRM servers - and probably make a 10x speedup for these job types.

Brian

hufnagel:
Most obvious change is to not use the stagin method defined in site-local-config to copy logArchives to the local node, but edmCopyUtil. This requires setting up a Scram environment and calling edmCopyUtil via scram (and also some proxy forwarding). There is a hardcoded dependency on CMSSW_8_0_8, but since all we do here is copy logArchives to local disk, this should work as long as we don't retire CMSSW_8_0_8. Well, and as long as the chosen slc6_amd64_gcc493 architecture can run on all of our hardware.

At the moment edmCopyUtil is called for 10 files at a time, so if any of the 10 files can't be read, all 10 files are skipped.

I also fixed a bunch of design flaws and bugs in the previous code:

Input file deletion didn't work (at all) due to a wrong dictionary keyword, fixed that.

Input file deletion should only be done after LogCollect stageout and after job is considered successful.

The skipped files were added to the job report without an LFN, leading to error messages in the JobAccountant.

Reporting LogCollect output was done in a weird way. We reported the output file that was staged to Castor, but with an EOS location and immediately after the mandatory Castor stageout succeeded (and even if the optional EOS stageout failed). Changed that to only report the EOS stageout, since that is optional, you can have a LogCollect job that succeeds but doesn't report any output since the EOS stageout failed. Which is correct IMO if you care about output actually present on EOS.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -2,11 +2,12 @@
 """
 _Step.Executor.LogCollect_
 
-Implementation of an Executor for a StageOut step.
+Implementation of an Executor for a LogCollect step.
 """
 from __future__ import print_function
 
 import os
+import re
 import logging
 import signal
 import tarfile
@@ -24,6 +25,10 @@ from WMCore.Storage.DeleteMgr import DeleteMgr
 from WMCore.Storage.StageOutError import StageOutFailure
 
 from WMCore.Algorithms.Alarm import Alarm, alarmHandler
+
+from WMCore.WMRuntime.Tools.Scram import Scram
+
+from Utils.IterTools import grouper
 
 
 class LogCollect(Executor):
@@ -53,6 +58,10 @@ class LogCollect(Executor):
         _execute_
 
         """
+        scramCommand = self.step.application.setup.scramCommand
+        scramArch = self.step.application.setup.scramArch
+        cmsswVersion = self.step.application.setup.cmsswVersion
+
         # Are we using emulators again?
         if (emulator != None):
             return emulator.emulate( self.step, self.job )
@@ -62,38 +71,39 @@ class LogCollect(Executor):
             overrides = self.step.override.dictionary_()
 
         # Set wait to over an hour
-        waitTime  = overrides.get('waitTime', 3600 + (self.step.retryDelay * self.step.retryCount))
-        seName    = overrides.get('se-name', "srm-cms.cern.ch")
-        pnn       = overrides.get('phedex-node', "T2_CH_CERN")
-        lfnPrefix = overrides.get('lfn-prefix', "srm://srm-cms.cern.ch:8443/srm/managerv2?SFN=/castor/cern.ch/cms")
-        lfnBase   = overrides.get('lfnBase',   "/store/user/amaltaro")
-        userLogs  = overrides.get('userLogs',  False)
+        waitTime = overrides.get('waitTime', 3600 + (self.step.retryDelay * self.step.retryCount))
 
-        stageOutParams = {"command": "srmv2-lcg",
-                          "se-name": seName,  "phedex-node": pnn, "lfn-prefix": lfnPrefix}
+        # Pull out StageOutMgr Overrides
+        # switch between old stageOut behavior and new, fancy stage out behavior
+        useNewStageOutCode = False
+        if 'newStageOut' in overrides and overrides.get('newStageOut'):
+            useNewStageOutCode = True
 
-        # Set EOS stage out params
-        seEOSName    = "srm-eoscms.cern.ch"
-        lfnEOSPrefix = "srm://srm-eoscms.cern.ch:8443/srm/v2/server?SFN=/eos/cms"
-        
-        stageEOSOutParams = {"command": "srmv2-lcg",
-                             "se-name": seEOSName, "phedex-node": pnn, "lfn-prefix": lfnEOSPrefix}    
+        # hardcode CERN Castor T0_CH_CERN_MSS stageout parameters
+        castorStageOutParams = {}
+        castorStageOutParams['command'] = overrides.get('command', "srmv2-lcg")
+        castorStageOutParams['option'] = overrides.get('option', "")
+        castorStageOutParams['se-name'] = overrides.get('se-name', "srm-cms.cern.ch")
+        castorStageOutParams['phedex-node'] = overrides.get('phedex-node', "T2_CH_CERN")
+        castorStageOutParams['lfn-prefix'] = overrides.get('lfn-prefix', "srm://srm-cms.cern.ch:8443/srm/managerv2?SFN=/castor/cern.ch/cms")
 
-        # Okay, we need a stageOut Manager
+        # hardcode CERN EOS T2_CH_CERN stageout parameters
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = overrides.get('command', "srmv2-lcg")
+        eosStageOutParams['option'] = overrides.get('option', "")
+        eosStageOutParams['se-name'] = overrides.get('se-name', "srm-eoscms.cern.ch")
+        eosStageOutParams['phedex-node'] = overrides.get('phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = overrides.get('lfn-prefix', "srm://srm-eoscms.cern.ch:8443/srm/v2/server?SFN=/eos/cms")
+
+        # are we using the new stageout method ?
         useNewStageOutCode = False
         if getattr(self.step, 'newStageout', False) or \
             ('newStageOut' in overrides and overrides.get('newStageOut')):
             useNewStageOutCode = True
 
         try:
-            if not useNewStageOutCode:
-                # old style
-                stageOutMgr = StageOutMgr(**stageOutParams)
-                stageOutEOSMgr = StageOutMgr(**stageEOSOutParams)
-                stageInMgr = StageInMgr()
-                deleteMgr = DeleteMgr()
-            else:
-                # new style (is even working ???)
+            if useNewStageOutCode:
+                # is this even working ???
                 #logging.info("LOGCOLLECT IS USING NEW STAGEOUT CODE")
                 #stageOutMgr = StageOutMgr(retryPauseTime  = self.step.retryDelay,
                 #                          numberOfRetries = self.step.retryCount,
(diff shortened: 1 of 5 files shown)


hufnagel:
@ticoann @amaltaro @bbockelm feel free to comment

hufnagel:
Question, does the WMCore scram code allow us to leave the scram architecture with some default value, ie. scram itself picks what the best architecture is ? This would be preferable to hardcoding it.

hufnagel:
Ok, jenkins is happier now. Please don't merge yet, we still have the scram issue to resolve and I just posted a question on the output reporting in HN which I want to have an answer on (and likely this will lead to some modifications of the patch).

src/python/WMCore/WMBS/MySQL/Create.py
```diff
@@ -65,6 +65,26 @@ def __init__(self, logger = None, dbi = None, params = None):
           """CREATE UNIQUE INDEX uniq_wmbs_file_run_lumi on
              wmbs_file_runlumi_map (fileid, run, lumi)"""
 
+        #
+        # Insert LogCollect site to account for LogCollect output
+        #
+        self.inserts["99_location_logcollect_1"] = \
```
amaltaro: This is way too hacky IMO. Why are you adding a `LogCollect` *site_name* into the locations tables?

src/python/WMCore/WMRuntime/Tools/Scram.py
```diff
@@ -110,7 +110,9 @@ def preCommand(self):
 
         """
         result = ""
-        if self.architecture != None:
+        if self.architecture == None:
```
amaltaro: `is None` is the best practice to do this comparison.

src/python/WMCore/WMRuntime/Tools/Scram.py
```diff
@@ -139,7 +141,10 @@ def project(self):
         # send commands to the subshell utilising the process writer method
         self.procWriter(proc, self.library_path)
         self.procWriter(proc, self.preCommand())
-        self.procWriter(proc, "%s --arch %s project CMSSW %s\n" % (self.command, self.architecture, self.version))
+        if self.architecture == None:
```
amaltaro: Same here, `is None` please.

src/python/WMCore/WMRuntime/Tools/Scram.py
amaltaro: BTW, why you explicitly export SCRAM_ARCH (without value) now. Is it really needed?

amaltaro:
@ericvaandering can you have a look at this PR please? Jenkins are going crazy :)

hufnagel:
The reason that jenkins is going nuts is that I updated the pull request like twelve times after I made it, very frequently, always before the jenkins runs for the previous code version was done.

src/python/WMCore/WMRuntime/Tools/Scram.py
hufnagel: Tried to make it pick the architecture automatically. This works interactively, but not here unfortunately. Have HN thread open about it.

src/python/WMCore/WMBS/MySQL/Create.py
hufnagel: Haha, what we have now is way more hacky (registering a file that is present at T0_CH_CERN_MSS with the T2_CH_CERN PNN).

This is correct. We stage out to T0_CH_CERN_MSS, we should register the files as present at T0_CH_CERN_MSS and the only way to do that is to have a site that has that as PNN. No real processing site has it, so I needed to add one.

hufnagel: I could have done he same with adding a fake LogCollect site to resource control via the script but figured this was cleaner. It doesn't affect anything from what I can tell, except you can now register files with T0_CH_CERN_MSS PNN in WMBS.

amaltaro: I think the reason we report T2_CH_CERN PNN instead is because the logCollect tarball is also - optionally - staged to EOS, making it easier for anyone looking up for a PFN.

Having a hacky code does not mean you can make it worse. I'm totally against adding LogCollect to wmbs_location and over complicating what's already complicated. We better not have any report then (as we do for the last several years, without any issues).

hufnagel: Well, then you should actually report the tarBall stagout to T2_CH_CERN (and not report it when that optional stageout fails), which you don't do. As-is the LogCollect stageout and what it is reporting doesn't make sense. I am going to talk to @ticoann today to figure out what we are actually supposed to be reporting here.

And btw, we haven't done this for years, this reporting of LogCollect output is new, we have never done this before.

amaltaro: I'm confused on which report it is. Is it reporting the job processing to dashboard (I don't think so)? Is it reporting the stage out of this job to dashboard? or what? What is the benefit on this report (possible none, since it was never put in place)?

hufnagel: It adds the output file to the FJR, so the JobAccountant puts it into WMBS. That's one of the things I want to talk to @ticoann about, whether we actually use that for anything (only thing I can think off is that WMArchive wants to have it in the FJR). We certainly don't need it for any further processing, as far as WMBS is concerned this is dead weight.

hufnagel: Somewhere between 1.0.15.something and master this changed, we noticed this on the Tier0 side since we now have to add T2_CH_CERN to our resource control even though we never want to run jobs there. Just to be able to register the LogCollect output in WMBS.

hufnagel: As-is right now the JobAccountant writes one LogCollect output file with LFN="srm castor pfn" location at T2_CH_CERN, which is complete bogus on various levels. The LFN is wrong, it's a PFN and it's a PFN that doesn't actually represent a file on T2_CH_CERN, but on T0_CH_CERN_MSS.

hufnagel: I would prefer we get rid of the reporting altogether, but if we do have to report something, we should report correct information. Which is what my patch is trying to do...

hufnagel: That doesn't work either. You report not the site, but the PNN. And that is T0_CH_CERN_MSS, which is not the pnn for T0_CH_CERN. The real fix here is to break the association between PNN and site in the WMBS schema, but that is something that has potential consequences for all of resource control and how we do location handling in WMCore.

hufnagel: In the disk/tape separated environment, there is no processing site where the PNN is T0_CH_CERN_MSS. Which is why I added a fake LogCollect site.

amaltaro: Then T0_CH_CERN_Disk? It's not the correct storage, but at least you get rid of T2 on your side.

hufnagel: I want to talk to @ticoann first to check what this is actually used for before I make further changes.

ericvaandering:
It handles it eventually, but every commit ID is a different commit that has to be tested. Now why we don't get results for the test at all, I dunno....

hufnagel:
Descoped the PR, removed the change in stageout PNN and the LogCollect fake site. Also use the workflow cmssw and scram arch, but only if they are CMSSW_8_X or newer. Otherwise the standard StageIn method (file by file) is used.

hufnagel:
Still have to cleanup the Scram class, but as None is not an allowed value at all (scram binary cannot auto-detect arch in non-interactive mode), I might clean this up a bit more.

hufnagel:
Already talked this through with @ticoann , we agreed that the general design is ok.

@amaltaro , can you review again please ?

One thing I removed was a the self.report.addError() call on timing out the stageIn. That didn't make any sense to me since that call is to set the exit code and error string for a failing step and timing out a stageIn does not fail a step. The code is from 2011, Matt Norman wrote it, so no chance to figure out what it was for (if anything).

amaltaro:
Code looks good, Dirk. Did you run some tests? Since we have time to merge this PR, I'd like to run a couple of tests in the coming days before we merge it.

hufnagel:
I have run lost of tests, but with slightly different intermediate code versions. This included testing both code paths (edmCopyUtil and stageIn). Rerunning these now with the final code, will report back once that is done.

Also, run any tests you want since my testing is limited to the Tier0. As far as merging goes, we might also want to backport this to the current production agents ( @bbockelm should say on how urgent this is ).

amaltaro:
In principle we wait for the next production release, timescale of 2 weeks'ish. If we waited for > 5 years, I guess we can wait a couple of weeks more :)

bbockelm:
No back port is needed I think - sites have been putting up with the current behavior for many months at this point.

hufnagel:
Ok, final code version also works in both code paths. @amaltaro , can you review the alarm timeout settings ? Since LogCollect jobs can have up to 500 input files I was aggressive (don't want to have this take forever if the storage has problems). Especially for the edmCopyUtil or stageIn I only give it 1min per file. Should be enough really...

amaltaro:
In summary the alarms are set to:
1. 3600 secs (* retries) for stage out to castor or EOS
2. 60 secs per file for stage in to the worker node or for deletion from the storage
confirm?

BTW, I just noticed there are changes to StdBase, it will require a cmsweb upgrade to get it working for central production (I believe it worked in your tests because Tier0 creates the workflow specs locally)

hufnagel:
Correct about the timeouts. Since copy, stageIn or deletion work on local storage, I thought 1 min per file is plenty.

About StdBase, are you sure ? I didn't change any of the parameters and aren't the specs build when work is pulled from the local workqueue ? I had to change StdBase since LogCollect needed to get he cmssw version and scram arch from the workflow.

amaltaro:
> I had to change StdBase since LogCollect needed to get he cmssw version and scram arch from the workflow.

Exactly, this happens during workflow creation (as discussed in the RP meeting).

hufnagel:
Hm, does that mean you need a LogCollect executor code version that can work with both old and new cmsweb ? Or can you do a synced deployment ?

hufnagel:
@amaltaro , when is this going to be merged ?

amaltaro:
Provided everything goes right, this week :-) I'm merging some very intrusive patches from Eric, then I'll create new clean RPMs and start looking into this one (though I'll only verify the functional aspect)

hufnagel:
Not sure that will work. You need the StdBase changes for the runtime changes to work. If you run the new runtime code with an old StdBase, the LogCollect jobs will crash. That means this code only works with a new cmsweb deployment.

amaltaro:
Yep, I'll test it in my VM with "central services"

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -62,38 +71,39 @@ def execute(self, emulator = None):
...
         try:
-            if not useNewStageOutCode:
-                # old style
-                stageOutMgr = StageOutMgr(**stageOutParams)
-                stageOutEOSMgr = StageOutMgr(**stageEOSOutParams)
-                stageInMgr = StageInMgr()
-                deleteMgr = DeleteMgr()
-            else:
-                # new style (is even working ???)
+            if useNewStageOutCode:
```
amaltaro: The best part of this code (and actually the current master code) is that there is *no* difference at all when we use the new stage out code or not... at least not when we instantiate these classes, with exactly the same arguments.

Dirk, please remove the `if/else` and the dup lines

amaltaro:
I would also suggest not to touch the Create database modules, since there is actually no need for changes over there. Next time we need to update that code, then we could also apply pylint/cosmetic changes.

hufnagel:
I already modified the Create database module (as part of some larger modifications that were rolled back). Unless you see a problem with the remaining mods (which should all be cleanup/cosmetic), I'd rather leave them in since they do improve things and I don't want to reinvent them again next time. If I take them out here I would just make another PR immediately with the changes, which is kind of pointless since they are already in here.

I also don't want to touch the old/new stageout code dup lines. They have been like this in the code for a long time and I still have cleaning up old and new stageout code on my todo list. Also, while *currently* there is no difference, I don't want to remove the code that selects between them because in the future there might be a difference and then we would have to reinvent the code that figures out what code path to follow.

amaltaro:
Ok, I ran a few 5x, 7x and one 8_0_5 workflow and all logcollects succeeded. Basic tests are done and we can test it further during the testbed validation (which reminds me we need to backport this PR to 1.0.17_wmagent branch).

PR #6896, opened 2016-05-29
Fix DBS3+CherryPy5 issue which is a mismatch type problem between python and Oracle

vkuznet:


ticoann:
Test this please

ticoann:
test this please

ticoann:
Thanks Valentin, I will run the quick test beforw I merge it.

amaltaro:
Shouldn't we cast arguments into strings _only_ when arg is unicode?
At least one of the unittests failures is because it expects an integer and gets a string.

amaltaro:
BTW, it seems we don't need to rush with this patch right now because the cherrypy upgrade will most likely be postponed to the next month, when we have the whole month to properly validate all bits.

vkuznet:
Alan,
it is good point, but I need to understand why we pass an integer to REST
model. By definition REST model runs within web server, clients communicate
with REST server via HTTP interface, i.e. when you place an url call
you send string anyhow which later the code may translate to int/float types
if necessary.

Could you please point to unit test is failing.

I don't mind to change code to check basestring type for arguments though.

Best,
Valentin.

vkuznet:
Alan,
I made a change to cast only basestring to str.
Valentin.

amaltaro:
Thanks, Valentin! That was it, all unit tests are happy now :-)

Can you apply a few minor fixes/improvements please:
1. remove the double spacing in your both `if` statements
2. remove this pylint `Line 3 E0012 Bad option value 'W0142'`

If you prefer, you can run pylint yourself and cleanup this code as much as you can and is safe (if so, create an additional commit then please).

vkuznet:
Alan, I changed code based on your suggestions. Regarding pylint improvements, I think it should be done separately, let's not these things together.

src/python/WMCore/WebTools/RESTModel.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # I'm afraid *args, **kwargs magic is needed here
-# pylint: disable=W0142
+# pylint: disable=
```
amaltaro: You can completely remove this line

amaltaro:
Thanks, Valentin. Can you squash all these commits into a single one please (don't forget looking at my last comment about completely removing a line, made in the code)?

vkuznet:
Alan,
honestly I never understood the value of squash, commits are dedicated to see
how code evolved and in my mind all commits have real value (since someone
who may want to reproduce certain code change can do it) instead of
squashed one. Moreover the squash may lead to unwanted error and can
screw things up (unintentionally). Anyway, I merged stuff and removed the line with pylint.
Valentin.

amaltaro:
Sorry, I completely disagree (especially in this case). Any bug fix, feature change or new feature that we make (provided it's not that big that affects reviewing), should go in only *one* commit IMO, it makes navigation through the history of that file much easier and gives an specific pointer to what was changed, instead of navigating through several commits to see that a handful of lines were changed to fix a bug for instance.

However, we do keep pylint fix/improvements in a different commit, since those should be very safe and usually no review needed (making the real code review much easier and cleaner).

BTW, if you squash (s) you still keep the commit message history of your changes. If you fixup (f) then you really get rid of of those messages and completely loose track of your changes. Squashing should be very safe, different than rebasing your branch to fix conflicts, which I always suffer to handle :)

ticoann:
Thanks Valentin, Alan. It seems this fetch shouldn't affect current functionality and I would like to include for testbed validation. I know we won't have enough time to do new validation but I don't think we really need to validate all the workflows, but just performance aspect. 
The reason I want to include this is that reqmgr would have the same problem it is using cx_Oracle, but I guess since db call volume is much smaller than DBS. It might not be that noticeable. 

Alan, what do you think? If you think it would be too risky, we won't include for this release. 
for now, I didn't include this for new release.
https://gitlab.cern.ch/cms-http-group/doc/issues/31

@yuyiguo, @vkuznet, Yuyi, Valentin, However, if you want to have new release for DBS, I can merge this and create a new tag for DBS to use regardless what reqmgr will do

amaltaro:
Looks a safe patch to push now to testbed, Seangchan. In the worst case, we can roll it back by the end of the week.
I'll merge it and cut a new tag once I get to CERN.

ticoann:
Thanks Alan for quick reply. Since I am still awake I will do that. 😄

ticoann:
@vkuznet,  @yuyiguo, new tag is created including this patch 1.0.17.pre6. If you want to update DBS as well please use that tag. Thanks,

amaltaro:
That's a shame, working on a US holiday at 2 am... :P

ericvaandering:
Can you paste your thread with the oracle_cx list here? Does that call for a different solution?

vkuznet:
Following Eric's suggestion here is a link to cx_oracle thread I started on their mailing list:
https://sourceforge.net/p/cx-oracle/mailman/cx-oracle-users/?viewmonth=201606
which basically confirms my observation that unicode string should be treated as string when comes to oracle queries. I see a few solutions here:

- ask CERN DBS to convert DBS tables to proper NCHAR data type which will cover unicode. This may blow up DBS tables/indexes (unicode is more than str). Then our python code will use unicode to access the oracle tables
- request to add setinputsizes() call in SQLAlchemy which we use to cover the conversion for us
- use plain cx_oracle driver instead of SQLAlchemy (it can be painful to refactor WMCore code).

Out of these options I rather consider first one, which of course require DBS opinion and actual conversion of data in DBS data. Going unicode for all data should solve many problem down the road in a long run since all tools, python included, moving towards this direction.

ericvaandering:
Thanks @vkuznet. I think #3 is a non-starter since every year or two we are asked how difficult it would be to move away from Oracle. Question about #1 is if we did the DB change but left our code as is, would we incur a performance penalty until we had them lined up again. Also, where else do we suffer from this problem?

vkuznet:
If we change fields in oracle to support unicode we'll need to perform stress
testing to see how it will behave. In my opinion we may expect some slowness,
but it should be not dramatic.

Regarding your last question, all data-services which are written in python
and use oracle back-end will suffer from this problem, DBS, T0, etc.

As cx_oracle folks said the effect can be non well pronounced if tables are
small, since query will bypass the index. What we observe is a largest table in
DBS, the file table, which has O(10-100M) records.

ericvaandering:
Thanks, that makes sense. Let's test solution #1 if we can and quantify if/how much of a downtime or degraded performance period we need to rebuild indices. We should also look at the agent schemas.

hufnagel:
Solution 1 only really makes sense for fields where we actually benefit from unit tests, doesn't it ? We have plenty if strings used for things like status names etc, where there really isn't any benefit to using or supporting unicode.

vkuznet:
Dirk,
the problem is that we should either use strings or unicode. Otherwise (mix
case) we end-up with end-less and error prone if statements which should
distinguish data types for certain columns. In other words we need to
go over all columns in all tables and ensure in python code proper type conversion.

What I suggested is a perfect solution for python 2.x land, i.e.
cast basestrings (which is str or unicode) to strings. But it is not suitable
for python 3.x where I bet all web based frameworks will rely on unicode,
while our back-end needs strings. That's why I think moving towards
unicode for everything is a better and uniform solution in a long run.

Valentin.

hufnagel:
Ok, I can see going all in for unicode if we have to have it for some parameters makes sense for code maintenance reasons. It's not just Oracle though, this would need some campaign to convert all MySQL and Oracle "string" types in our schema to one supported unicode enabled type. Which needs a bunch of testing before we can do it...

yuyiguo:
DBS using varchar2 for all its string columns. Convert varchar2 to nvarchar2 may not be a bigger deal, but we have to rebuild all the indexes that may require downtime. This is not only  for LFN, but all of the varchar2 columns, such as dataset, blocks, data tiers  and so on.

I am not sure if nvarchar will cause any performance issues. Oracle recommended that people stay on varchar2 unless one has special characters that unicode has to be used. In our case, we don't have any special characters. 

Going to python 3, there will be no string at all?  If so,  we cannot even cast to str as we are doing now. I guess the question becomes if Oracle API can have something to indicate if this is a real unicode or a str and binding the DB accordingly.

ericvaandering:
@hufnagel I look at it differently. The world is moving to unicode, so while probably none of our file names will ever benefit from unicode (although I can imagine French diacriticals sneaking into dataset names), it seems to me we should start to run Oracle in unicode mode as well rather than always having to remember to try to downcast the native "string" in our tools to ASCII.

ericvaandering:
Python3 has no ASCII string.

hufnagel:
@ericvaandering , I agree if you look at this from a "our software stack is already 80% unicode enabled, lets make sure it's 100% and avoid any casting issues and potential problems down the road".

That is, provided this doesn't screw us for performance reasons...

yuyiguo:
I will contact with Kate to ask her what we will facing on the DB side if we go to nvarchar2.

hufnagel:
Interesting question is, if you pass in unicode to sqlalchemy and cx_oracle in a all python3 environment, write the data into a carchar2 field and then read back what you just wrote, are you guaranteed to get a correct result even ?

ericvaandering:
I would hope you'd get an exception on the insert.

vkuznet:
There is another solution to this problem which I found based on suggestion from cx_oracle folks. Here is a code snippet which works:
```
    cursor = connection.cursor()
    cursor.setinputsizes(logical_file_name=cx_Oracle.STRING)
    cursor.execute(sql, **params)
```

here we use a connection object (it is part of DBInterface in WMCore) and we explicitly instructs cursor to use cx_Oracle.STRING for input unicode values. But I can't find yet how we can implement such solution in WMCore. The problem is that we rely on DBInterface.processData API (it is used in DBS) and we need to extend it to pass casting parameters such as in shown example. Then we can still use either string or unicode for input parameters but must explicitly define its size via aforementioned cast.

yuyiguo:
Here is a suggestion from a fermi lab colleague who faced the same problem before.

I had the same problem with cherrypy strings sometimes being unicode. With cx_Oracle you can set a type converter function on each connection:

            def _OracleInputTypeHandler(cursor, value, numElements):
                """ Convert unicode objects to byte strings before sending them to Oracle """
                if isinstance(value, unicode):
                    encoding = cursor.connection.encoding # the lambda below must not capture the cursor as it creates a circular reference
                    return cursor.var(str, arraysize = numElements,
                        inconverter = lambda x: x.encode(encoding))

           connection.inputtypehandler = _OracleInputTypeHandler # connection is an instance of cx_Oracle.Connection

This will convert any unicode strings to the byte string encoding for the connection. Now no manual conversion of each string is needed: they’ll all be sent as byte strings (probably utf-8).

I never tried this with python 3.0 - it’s possible cx_Oracle also works differently there.

yuyiguo:
Here is Kate's replay.
 Hi Yuyi,  

I would propose to create a small table with the new column types and test if it behaves properly on insert/update.
Then you would need to test if the indexes are still working ok and after this is ok, I would suggest the conversion of the test DB. 
I’ve had some discussions with other DBAs and we have no real experience using nvarchar, so I would suggest running some tests in int2r (conversion and performance) to asses it. 

Cheers 
Kate

vkuznet:
Yuyi,
this solution is slightly better variant of what I proposed, but it
brings conversion closer to oracle driver. The problem I see with it that
we use generic WMCore "wrapper" around connection objects which may
not only be a Oracle ones, e.g. MySQL, MariaDB. Plus encode method
is slower than str cast. And, as your colleague correctly pointed out
we don't know if it's gonna work in python 3.x
Valentin.

yuyiguo:
Here is more info on how to use with SqlAlchemy.

What I did was to use the SQLAlchemy event system to add a listener on the connect event (called every time a new connection is created). I set up the cx_Oracle type converter there. Then it was all entirely transparent to users; whether they got the connection directly from the SQLAlchemy engine or indirectly through a Session object it was correctly set up.

You can see the code here https://cdcvs.fnal.gov/redmine/projects/sam-web/repository/revisions/oracle/entry/python/samdb/database.py - line 52

(None of this is necessary with PostgreSQL - it doesn’t have a separate wide character type like Oracle does and if necessary data is automatically converted from the client encoding to the server encoding).

vkuznet:
Based on Yuyi's input I made another PR: https://github.com/dmwm/WMCore/pull/6912 which takes elegant solution provided in sam-web. I adopted it to DBCore interface and revert back RESTModel patch I applied earlier. This is better approach then the one I originally proposed since it handles only Oracle DB, it closer to execution and let intact input parameters in REST model. Even though both solution fix the problem upon further investigation I incline to use #6912 

Yuyi, could you please take #6912 and test with your DBS code that will work properly with new Cherrypy. If so I'll suggest that we'll schedule a change in next WMCore upgrade.

yuyiguo:
OK, I will verify this when I build for next release.

PR #6897, opened 2016-05-30
Merge Justas' patch for untarring with python 2.6 for older releases

emaszs:
Patch for an issue discovered a few releases ago about untarring the sandbox with older CMSSW versions, for example CMSSW_5_3_31 [*]. This was fixed on a previous WMCore CRAB branch but no master PR was created.

1.0.16_crab PR here: #6898 

https://github.com/dmwm/CRABServer/issues/5229#issuecomment-221912050

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
@@ -371,7 +371,13 @@ if [ -n "$USER_TARBALL" ] ; then
     EXIT_STATUS=$?
     if [ $EXIT_STATUS -ne 0 ]; then
        echo "***\nCouldn't untar sandbox with python2: $EXIT_STATUS\n";
-       exit 74;
+       echo "***\nWill try with python2.6 as it might be an old CMSSW release!"
+       python2.6 -m WMCore.WMRuntime.Scripts.UnpackUserTarball $USER_TARBALL $USER_FILES
+       EXIT_STATUS=$?
+       if [ $EXIT_STATUS -ne 0 ]; then
+          echo "***\nCouldn't untar sandbox with python2.6: $EXIT_STATUS\n";
+          exit 74;
+       fi
     fi
 fi
 echo "Completed SCRAM project"


src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -371,7 +371,14 @@ def post(self, emulator=None):
     EXIT_STATUS=$?
     if [ $EXIT_STATUS -ne 0 ]; then
        echo "***\nCouldn't untar sandbox with python2: $EXIT_STATUS\n";
-       exit 74;
+       echo "***\nWill try with python2.6 as it is might be old CMSSW release!"
+       python2.6 -m WMCore.WMRuntime.Scripts.UnpackUserTarball $USER_TARBALL $USER_FILES
+       EXIT_STATUS=$?
+       if [ $EXIT_STATUS -ne 0 ]; then
+          echo "***\nCouldn't untar sandbox with python2: $EXIT_STATUS\n";
+          echo "***\nThat was last try... failing..."
```
amaltaro: In L374: "it is might be old CMSSW" --> "it might be an old CMSSW" ...

In L378: "sandbox with python2" --> "sandbox with python2.6"

In L379: I think we can remove this "That was last try" line

emaszs:
Good points @amaltaro , updated the commit with the fixes.

belforte:
I hate these environment issues.
IIUC, code calls python2 and if this fails, calls python2.6. Seems very fragile, and surely ugly as hell.
Is this something like some CMSSW releases point with python2 to e.g. older versions (2.4 ?) but python2.6 is always available (where ?)
Points in the direction that we run a python script in an environment where `python2` is defined, but nothing can be expected about it. Seems the wrong thing to do.
Couldn't we rather use these incidents to make the tool more robust ?

amaltaro:
The way I understand it is that python2 points to the latest python2.X version available in the ... environment (I guess we can read in the worker node?).
If this alias is not available in the worker node, then it fallback to python2.6. If this one is also not available, then just bail out and complain.

belforte:
yeah.. I can read that in the code, but why do we want to rely on or even try aliases that we do no know who defines, what they point to, and even if they exist ? We run a python script and have no idea which python version it runs with and why ? That's pretty dangerous IMHO, even if "it works".
At least do this in some initial setup, then downstream code can assume some sanity. 

I know.. thiis only a philosophical point, one should always try to solve problems at the root so that they do not appear, rather then workaround them in each stop they pop up. I know... there's never enough time to clean up things, only to patch and make things  even harder to cleanup.

sorry for bugging. End of rant.

amaltaro:
I should shut up because it's a looooong discussion, however, if CRAB were using the COMP repository for all crab "stuff" besides running the cmsRun, that would probably reduce these kind of problems to almost 0, no?

belforte:
yes. Or maybe one should call scram runtime after WMCore.WMRuntime.Scripts.UnpackUserTarball ?
https://github.com/emaszs/WMCore/blob/16a49e8085760e91e9d61ee0c87495128ec5b69c/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py#L366

PR #6904, opened 2016-06-02
make sure CMSSW always finds HOME in environment, fixes #6894

### LINKED ISSUE #6894: Exit Code 134 $HOME/.root.mimes
the frequent error code 134 seems to be related to  $HOME/.root.mimes

does wmcore make reference to this file, how does this get accessed by the job ?

https://hypernews.cern.ch/HyperNews/CMS/get/edmFramework/3654.html

this is causing major failures and inefficiencies in production.

hufnagel:
Progression in environment variable checks is based on what I observed in Tier0 job environments.

Tested the more generic code changes, but obviously can't test the 'HOME not set use case' since the Tier0 jobs have HOME set.

@amaltaro @vlimant please review and maybe try out in a production agent to see what it does. Can't be worse than CMSSW crashing because HOME not being set.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
@@ -218,11 +218,18 @@ class CMSSW(Executor):
                 cmsswArguments]
         logging.info("Executing CMSSW. args: %s", args)
 
-        # at CERN override the environment to work around problem with GSI authentication plugin and EOS
+        # possibly needed environment overrides for CMSSW call go here
+        envOverride = {}
+        # work around problem with GSI authentication plugin and EOS at CERN
         if socket.getfqdn().endswith("cern.ch"):
-            returncode = subprocess.call(args, stdout = stdoutHandle, stderr = stderrHandle, env=dict(os.environ, XRD_LOADBALANCERTTL='86400'))
-        else:
-            returncode = subprocess.call(args, stdout = stdoutHandle, stderr = stderrHandle)
+            envOverride['XRD_LOADBALANCERTTL'] = "86400"
+        # some libraries linked with CMSSW need HOME in the environment
+        if 'HOME' not in os.environ:
+            envOverride['HOME'] = os.environ.get('PWD', "/")
+
+        os.environ.update(envOverride)
+
+        returncode = subprocess.call(args, stdout = stdoutHandle, stderr = stderrHandle)
 
         self.setCondorChirpAttrDelayed('Chirp_WMCore_cmsRun_ExitCode', returncode)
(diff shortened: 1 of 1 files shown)


amaltaro:
Where this issue come from? Do we have a GH issue for that? Why would CMSSW not find $HOME variable in a few ... worker nodes but succeed in others?

If you can give me more details on where it failed, which workflow or configuration, we can setup it in *testbed* for testing.

hufnagel:
Check #6894 and the HN discussion. I just wrote the patch, @vlimant should follow up.

vlimant:
Hi Dirk, that's nice to force set it as an immediate fix.
I think we have to follow up with glidein to understand how it does not get set systematically.
Furthermore, we should find a way to report "some text" about the non-CMSSW exception (like that one, and segfault, and all) that result in 134, so that we can under the actual origin without digging up a log

hufnagel:
So it seems like it is set in a frontend validation script.

https://github.com/jamesletts/CMSglideinWMSValidation/blob/master/set_home_cms.source

Given this, should I simplify this patch look only at PWD with a fallback to / maybe as a safety feature ?

amaltaro:
Please do so, I feel like crying when I look at that line 230 :)

vlimant:
@amaltaro you don't have a taste for russian dolls ?

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -218,11 +218,20 @@ def execute(self, emulator=None):
                 cmsswArguments]
         logging.info("Executing CMSSW. args: %s", args)
 
-        # at CERN override the environment to work around problem with GSI authentication plugin and EOS
+        #
+        # possibly needed environment overrides for for CMSSW call go here
+        #
```
amaltaro: Can you reduce these 3 commented lines into 1 only? "for for" btw

amaltaro:
@vlimant since this is going to be a tricky validation, can you point me to the site where most of these errors came from? I'll send 2k jobs there and see whether we hit this 134 exit code again.

Moreover, I understood it depends on the library being used in CMSSW(?) So can you also point me the workflow name?

vlimant:
@amaltaro https://cms-logbook.cern.ch/elog/Workflow+processing/24500 pdmvserv_SUS-RunIISpring16FSPremix-00009_00008_v0__160526_002050_6726 ran on the HLT (admittedly using overflowing)

amaltaro:
Thanks, failures both at CERN and CERN_HLT. I'll clone this worfklow later on and run some jobs for the weekend.

amaltaro:
The clone workflow is about to finish, about 2/3 of the jobs failed but there was not a single exit code 134. I think we can consider this fix as done.

amaltaro:
Merging it now such that we can cut a new stable release.

PR #6905, opened 2016-06-02
Remove the need for a prompt when setting up unit tests

ericvaandering:
This change allows the script to run without any input from the user and, in turn, lets me set up a docker container ready to run tests. Please merge (or I will if @ticoann is on vacation)


amaltaro:
Seangchan should be flying at this time, so I just went ahead and merged

PR #6909, opened 2016-06-03
remove the dashboard reporting sleep call, fix #6899

### LINKED ISSUE #6899: time.sleep delaying Agent components
The polling cycle of some components in the WMAgent is taking too long. The cause is this line in the Dashboard Reporter:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Dashboard/DashboardAPI.py#L85
It is called every time that a job report is sent to Dashboard. Hundreds of jobs are reported per polling cycle when the agent has a big load, causing the delays that we observed.

hufnagel:


hufnagel:
@amaltaro please review :-)

amaltaro:
these are the king of PRs that I like reviewing :-D Thanks Dirk

PR #6910, opened 2016-06-07
Remove check for dup docs

amaltaro:
@vkuznet as we discussed in the past week, the server is supposed to always return the same list of ids, regardless whether there were dups or not, so I just removed that check.

Please review. I still need to test it in my VM

### DIFF
--- a/src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
+++ b/src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
@@ -2,7 +2,7 @@
 ArchiveDataPoller
 """
 
-from __future__ import (division, print_function) 
+from __future__ import (division, print_function)
 import logging
 import traceback
 from Utils.IterTools import grouper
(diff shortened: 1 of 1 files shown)


vkuznet:
Alan, I don't mind if you apply this patch, it is more a question to agent side though if it needs to know or report about duplicates. This line
https://github.com/dmwm/WMCore/pull/6910/files#diff-ecf84ade00ceea4c6584c57e768bb29fL65
may be important or not depending on an agent side.

amaltaro:
You said the server will always return the same length of ids (regardless of dups or not). That means that line is pointless and would always complain about dups. Thanks for the review.

vkuznet:
Alan,
the same length of ids does not mean all different ids, i.e.
[1,2] is not the same as [1,1].

So, the purpose of this patch was (probably) to see if agent sent
duplicates. The WMArchive will return the list with ids whose size
is a size of input docs. But if duplicates were presented in docs
you'll get some of them in returned list.

amaltaro:
If you wanted to say that the agent was - probably - checking whether it was sending dups within the
same bulk call, then of course this check should have been put in place before the upload attempt.

vkuznet:
It seems like Seangchan wanted to check if duplicates were send.
But it does seem to be useful since WMArchive provide such insurance
and it does check for duplicates before upload. That's why it reports
two types of duplicates in a log:
- the input had duplicates
- the STS had duplicates for provided input

amaltaro:
Seangchan, as discussed with Valentin when you were away. The server always return the same list of Ids, no matter whether there were dups or not. His suggestion was to actually check on the client side if we have dup to upload. This is what I did now, before uploading them to the server, check whether there are dups and make them unique if that's the case. I don't expect it to ever happen...

Please review and feel free to merge if that's Ok

amaltaro:
@ericvaandering do you know if there is a problem with Jenkins? I got this _weird_ message above and this PR 6973 is facing some kind of failure to run.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
@ticoann good to go from my side. Please review

ticoann:
Hi Alan, did you find any duplicate entry in the test. I think it shouldn't have duplicate entry in the couch at all since job id is used as id and it should be unique.
I think duplicate entry happens only if the call didn't get the response from WMArchive (actually succeeded) first cycle and try to push job id when it try to push them again next cycle.
In that case, we don't need to track that the list of duplicate entry but logging wouldn't hurt

src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
```diff
@@ -11,64 +11,68 @@
...
+
     def algorithm(self, parameters):
         """
         get information from wmbs, workqueue and local couch
         """
         try:
-            logging.info("Getting not archived data info from FWRJ db...")
-            data = self.fwjrAPI.getFWJRByArchiveStatus('ready', limit=1000)['rows']
-            
+            data = self.fwjrAPI.getFWJRByArchiveStatus('ready', limit=self.numDocsRetrievePerPolling)['rows']
```
ticoann: my mistake thanks. this should get in.

src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
```diff
@@ -11,64 +11,68 @@
...
                     archiverDocs.append(doc)
                     jobIDs.append(job["id"])
-                    
+
+                # probably useless... check whether we have duplicate docs to upload to WMArchive
+                duplicateIDs = set([x for x in jobIDs if jobIDs.count(x) > 1])
+                if duplicateIDs:
+                    logging.warn("Removing these duplicate entries before uploading them %s", duplicateIDs)
+                    archiverDocs = list(set(archiverDocs))
+
```
ticoann: I am not sure we will ever hit this condition true since job["id"] is unique value in couchdb. Did we find any case this is true?

amaltaro: Nope, I also think this test is useless (see my comment :-D), however this is where Valentin told me the check should be. See discussion in the beginning of the PR please

amaltaro:
General comment, Seangchan. My initial goal was to remove this message
```
There are duplicate entry set([])
```
that's printed over and over in the current produciton agents.

See comments above (and Valentin reply), but in summary WMArchive will always return the same list length as uploaded, no matter where a) all succeeded b) all failed or c) only some failed. @vkuznet please correct me if I'm wrong

src/python/WMComponent/ArchiveDataReporter/ArchiveDataPoller.py
ticoann: I read Valentin's comment but I think there is an little confusion. There shouldn't be any duplicate in the same cycle only in different cycle.
There was problem before but it is due to the hash collision which Valentin fixed.

amaltaro: Agreed! Do you want me to remove this check then (L55-60)?

vkuznet: > I read Valentin's comment but I think there is an little confusion. There shouldn't be any duplicate in the same cycle only in different cycle.
> There was problem before but it is due to the hash collision which Valentin fixed.

Seangchan is correct, we may only have duplicates if in first call WMArchive
fails and then agent make another call with duplicated docs.

amaltaro: > Seangchan is correct, we may only have duplicates if in first call WMArchive fails and then agent make another call with duplicated docs.

Which, on the client/agent side, there is no way to distinguish. Right?

vkuznet: On  0, Alan Malta Rodrigues <notifications@github.com> wrote:
> >              for slicedData in grouper(data, self.numDocsUploadPerCall):
> >                  jobIDs = []
> >                  archiverDocs = []
> >                  for job in slicedData:
> >                      doc = createArchiverDoc(job)
> >                      archiverDocs.append(doc)
> >                      jobIDs.append(job["id"])
> > -                    
> > +
> > +                # probably useless... check whether we have duplicate docs to upload to WMArchive
> > +                duplicateIDs = set([x for x in jobIDs if jobIDs.count(x) > 1])
> > +                if duplicateIDs:
> > +                    logging.warn("Removing these duplicate entries before uploading them %s", duplicateIDs)
> > +                    archiverDocs = list(set(archiverDocs))
> > +
> 
> > Seangchan is correct, we may only have duplicates if in first call WMArchive fails and then agent make another call with duplicated docs.
> 
> Which, on the client/agent side, there is no way to distinguish. Right?

On a client side you'll get after first try the status from the wmarchive.
If status is fail you know that you need to try again. But, yes, you don't know
which docs should be re-tried. We always said that client will "just retry
entire set". The WMArchive job is not to insert documents which are already
in archive, so from the client side it will be no difference if client
will sent 1, 2 or 10 times.

amaltaro:
Ok, given the previous comments and the inability to spot docs that succeeded or not uploading to WMArchive, I've just scratched the dup check.

It should be straight forward to add it back once (if) server report errors more accurately.

amaltaro:
patch tested in vocms0116 (hlt agent).
@ticoann please merge

PR #6913, opened 2016-06-08
Remove duplicate WMExceptions

amaltaro:
I know we wanted to make sure the error message would get exposed, but 3 exit codes with the same error message is a bit overkill :-D

Let's see whether there are unit tests using those exit code, grepping over tests was too painful.
Also deleted a binary file that somehow sneaked in.


juztas:
Please, dont. CMSSW can exit with 85, 84, 92. CRAB uses these exit codes and it changes to long one, but production does not change them.
Production last week top 10 exit codes:
```
0	2,884,831
-1	513,882
65	110,000
85	24,271
70	18,330
84	15,258
92	12,660
134	11,409
139	7,018
127	1,668
```
For last 3 months I do not see 151 in prod, but this should come from Stageout plugins, worth to check. Also, CRAB3 uses this exit code.

juztas:
Check with David or grep on CMSSW for 85, 84, 92. Others, let me google that for you:
```
Exit Code Number	Meaning	Example	Comments
1	Catchall for general errors	let "var1 = 1/0"	Miscellaneous errors, such as "divide by zero" and other impermissible operations
2	Misuse of shell builtins (according to Bash documentation)	empty_function() {}	Missing keyword or command, or permission problem (and diff return code on a failed binary file comparison).
126	Command invoked cannot execute	/dev/null	Permission problem or command is not an executable
127	"command not found"	illegal_command	Possible problem with $PATH or a typo
128	Invalid argument to exit	exit 3.14159	exit takes only integer args in the range 0 - 255 (see first footnote)
128+n	Fatal error signal "n"	kill -9 $PPID of script	$? returns 137 (128 + 9)
130	Script terminated by Control-C	Ctl-C	Control-C is fatal error signal 2, (130 = 128 + 2, see above)
255*	Exit status out of range	exit -1	exit takes only integer args in the range 0 - 255
```
So any 128+n is a kill, but with a different signal. (Whenever I added I kept what was on twiki and I think it was written there for a reason. Some have changed for sure)
147 - SIGSTOP
151 - SIGURG
Range from 1 - 512 is not the easiest way to debug, but CRAB already checks multiple things to make error codes from short to long and make clear messages to the user. Would be great to see the same from production ;)

ericvaandering:
You know about `git grep` I hope....

amaltaro:
Yes, I do. Though I used only `grep` :-D . Anyways, the outcome is the same and it returns gazillions of matches.

amaltaro:
Ok, if we have to keep it then we have to keep it. Thanks

PR #6925, opened 2016-06-10
make more efficient use of condor bulk submission

hufnagel:
Ok, it seems the only reason we needed to pre-sort for bulk submission was that some parameters were only selectively set and instead of using preset defaults, it would default to the value from a previous job within the same bulk submit. So bulk submits with some jobs setting the parameters and some jobs without setting the parameters lead to weird behavior.

This patch fixed that for the highIOjob flag (always set for either 0 or 1) and numberOfCores (always set, even for singlecore jobs which now specifically request 1 cpu). That in turn allows us to get rid of the pre-sorting by numberOfCores and highIOjob.

I tested this change for numberOfCores with a bulk submit of 26 jobs all requesting between 4 and 8 cpus and it worked fine, the jobs running in Condor showed up the correct number of cpus it requested.

There are still some parameters that are selectively set, but since we don't pre-sort by them, nothing is changed to how things work now. There is a chance we might even be able to get rid sorting by sandbox and just bulk submit everything without caring about sandbox, but that needs more work and also will need us to address these last outstanding parameters that are selectively set.

Example for how to deal with such a parameter would be what I did for MaxWallTime. Old code set it selectively, new code defaults to 12h if it is not set. Of course, if we say that the expected wall time should always be set for a job, we can also simplify this.

I'll follow up in HN with the list of parameters. We need to make sure that they are always set, either by always having the information available to set it or by providing some sensible default.

@amaltaro @ticoann , would appreciate feedback and comments on this patch (as it exists now).

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -210,7 +210,6 @@ class PyCondorPlugin(BasePlugin):
         self.scriptFile = None
         self.submitDir = None
         self.removeTime = getattr(config.BossAir, 'removeTime', 60)
-        self.useGSite = getattr(config.BossAir, 'useGLIDEINSites', False)
         self.submitWMSMode = getattr(config.BossAir, 'submitWMSMode', False)
         self.errorThreshold = getattr(config.BossAir, 'submitErrorThreshold', 10)
         self.errorCount = 0
@@ -380,80 +379,56 @@ class PyCondorPlugin(BasePlugin):
         if not os.path.exists(self.submitDir):
             os.makedirs(self.submitDir)
 
-
-        # Now assume that what we get is the following; a mostly
-        # unordered list of jobs with random sandboxes.
-        # We intend to sort them by sandbox.
-
-        # Do a secondary sort for numberOfCores and also for
-        # highIO jobs, they need to be submitted separately.
-
-        submitDict = {}
+        # Submit the jobs
         nSubmits = 0
-        for job in jobs:
-            sandbox = job['sandbox']
-            numberOfCores = job['numberOfCores']
-            highIOjob = job['highIOjob']
-            if sandbox not in submitDict.keys():
-                submitDict[sandbox] = {}
-            if numberOfCores not in submitDict[sandbox]:
-                submitDict[sandbox][numberOfCores] = {}
-            if highIOjob not in submitDict[sandbox][numberOfCores]:
-                submitDict[sandbox][numberOfCores][highIOjob] = []
-
-            submitDict[sandbox][numberOfCores][highIOjob].append(job)
-
-
-        # Now submit the bastards
         queueError = False
-        for sandbox in submitDict.keys():
-            for numberOfCores in submitDict[sandbox].keys():
-                for jobList in submitDict[sandbox][numberOfCores].values():
-                    if queueError:
-                        # If the queue has failed, then we must not process
-                        # any more jobs this cycle.
-                        break
-                    for jobsReady in grouper(jobList, self.jobsPerWorker):
-                        idList = [x['id'] for x in jobsReady]
-                        jdlList = self.makeSubmit(jobList=jobsReady)
-                        if not jdlList:
-                            # Then we got nothing
-                            logging.error("No JDL file made!")
-                            return {'NoResult': [0]}
-                        jdlFile = "%s/submit_%i_%i.jdl" % (self.submitDir, os.getpid(), idList[0])
-                        with open(jdlFile, 'w') as handle:
-                            handle.writelines(jdlList)
-                        jdlFiles.append(jdlFile)
+        for jobsReady in grouper(jobs, self.jobsPerWorker):
 
-                        # Now submit them
-                        logging.info("About to submit %i jobs", len(jobsReady))
-                        if self.glexecPath:
-                            command = 'CS=`which condor_submit`; '
-                            if self.glexecWrapScript:
-                                command += 'export GLEXEC_ENV=`%s 2>/dev/null`; ' % self.glexecWrapScript
-                            command += 'export GLEXEC_CLIENT_CERT=%s; ' % self.glexecProxyFile
-                            command += 'export GLEXEC_SOURCE_PROXY=%s; ' % self.glexecProxyFile
-                            command += 'export X509_USER_PROXY=%s; ' % self.glexecProxyFile
-                            command += 'export GLEXEC_TARGET_PROXY=%s; ' % self.jdlProxyFile
-                            if self.glexecUnwrapScript:
-                                command += '%s %s -- $CS %s' % (self.glexecPath, self.glexecUnwrapScript, jdlFile)
-                            else:
-                                command += '%s $CS %s' % (self.glexecPath, jdlFile)
-                        else:
-                            command = "condor_submit %s" % jdlFile
+            if queueError:
+                # If the queue has failed, then we must not process any more jobs this cycle.
+                break
 
-                        try:
-                            self.inputQueue.put({'command': command, 'idList': idList})
-                        except AssertionError as ex:
-                            msg = "Critical error: input pipeline probably closed.\n"
-                            msg += str(ex)
-                            msg += "Error Procedure: Something critical has happened in the worker process\n"
-                            msg += "We will now proceed to pull all useful data from the queue (if it exists)\n"
-                            msg += "Then refresh the worker pool\n"
-                            logging.error(msg)
-                            queueError = True
-                            break
-                        nSubmits += 1
+            idList = [x['id'] for x in jobsReady]
+            jdlList = self.makeSubmit(jobList=jobsReady)
+            if not jdlList:
+                # Then we got nothing
+                logging.error("No JDL file made!")
+                return {'NoResult': [0]}
+            jdlFile = "%s/submit_%i_%i.jdl" % (self.submitDir, os.getpid(), idList[0])
+
+            with open(jdlFile, 'w') as handle:
+                handle.writelines(jdlList)
+            jdlFiles.append(jdlFile)
+
+            # Now submit them
+            logging.info("About to submit %i jobs", len(jobsReady))
+            if self.glexecPath:
+                command = 'CS=`which condor_submit`; '
+                if self.glexecWrapScript:
+                    command += 'export GLEXEC_ENV=`%s 2>/dev/null`; ' % self.glexecWrapScript
+                command += 'export GLEXEC_CLIENT_CERT=%s; ' % self.glexecProxyFile
+                command += 'export GLEXEC_SOURCE_PROXY=%s; ' % self.glexecProxyFile
+                command += 'export X509_USER_PROXY=%s; ' % self.glexecProxyFile
+                command += 'export GLEXEC_TARGET_PROXY=%s; ' % self.jdlProxyFile
+                if self.glexecUnwrapScript:
+                    command += '%s %s -- $CS %s' % (self.glexecPath, self.glexecUnwrapScript, jdlFile)
+                else:
+                    command += '%s $CS %s' % (self.glexecPath, jdlFile)
+            else:
+                command = "condor_submit %s" % jdlFile
+
+            try:
+                self.inputQueue.put({'command': command, 'idList': idList})
+            except AssertionError as ex:
+                msg = "Critical error: input pipeline probably closed.\n"
+                msg += str(ex)
+                msg += "Error Procedure: Something critical has happened in the worker process\n"
+                msg += "We will now proceed to pull all useful data from the queue (if it exists)\n"
+                msg += "Then refresh the worker pool\n"
+                logging.error(msg)
+                queueError = True
+                break
+            nSubmits += 1
 
         # Now we should have sent all jobs to be submitted
         # Going to do the rest of it now
@@ -559,7 +534,7 @@ class PyCondorPlugin(BasePlugin):
         logging.info("Done submitting jobs for this cycle in PyCondorPlugin")
         return successfulJobs, failedJobs
 
-    def track(self, jobs, info=None):
+    def track(self, jobs):
         """
         _track_
 
@@ -663,7 +638,7 @@ class PyCondorPlugin(BasePlugin):
                     try:
                         os.remove(reportName)
                         condorReport.save(filename=reportName)
-                    except Exception as ex:
+                    except:
                         logging.error("Cannot remove and replace empty report %s", reportName)
                         logging.error("Report continuing without error!")
             else:
@@ -717,7 +692,7 @@ class PyCondorPlugin(BasePlugin):
 
         return jobtokill
 
-    def kill(self, jobs, info=None):
+    def kill(self, jobs):
         """
         _kill_
 
@@ -801,8 +776,7 @@ class PyCondorPlugin(BasePlugin):
         jdl.append("Log = condor.$(Cluster).$(Process).log\n")
 
         jdl.append("+WMAgent_AgentName = \"%s\"\n" % (self.agent))
-        jdl.append(
-            "+JOBGLIDEIN_CMSSite= \"$$([ifThenElse(GLIDEIN_CMSSite is undefined, \\\"Unknown\\\", GLIDEIN_CMSSite)])\"\n")
+        jdl.append("+JOBGLIDEIN_CMSSite= \"$$([ifThenElse(GLIDEIN_CMSSite is undefined, \\\"Unknown\\\", GLIDEIN_CMSSite)])\"\n")
 
         # Required for global pool accounting
         jdl.append("+AcctGroup = \"%s\"\n" % (self.acctGroup))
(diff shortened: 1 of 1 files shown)


hufnagel:
This patch will have to be reworked once the "jdl magic" patches are merged, but that's ok. It's good enough for code review and discussions as-is.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -976,22 +962,23 @@ def customizePerJob(self, job):
             jdl.append('+DESIRED_CMSDataLocations = "%s"\n' % ','.join(job['inputDatasetLocations']))
 
         # HighIO jobs
-        highIOjob = job.get('highIOjob')
-        if highIOjob:
+        if job.get('highIOjob'):
```
amaltaro: Possible a useless suggestions, but since this code is polluted with `ifs`, how about something like:
`jdl.append('+Requestioslots = %d\n' % job.get('highIOjob', 0))`
?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -976,22 +962,23 @@ def customizePerJob(self, job):
...
-        # Set up JDL for multithreaded jobs
-        if job.get('numberOfCores', 1) > 1:
-            jdl.append('machine_count = 1\n')
-            jdl.append('request_cpus = %s\n' % job.get('numberOfCores', 1))
+        # Max wall time estimate (12h if undefined)
+        jdl.append('+MaxWallTimeMins = %d\n' % int(job.get('estimatedJobTime', 12*3600))/60)
+
+        # Set up JDL for number of cores
+        jdl.append('machine_count = 1\n')
+        jdl.append('request_cpus = %s\n' % job.get('numberOfCores', 1)
```
amaltaro: missing close parenthesis

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -976,22 +962,23 @@ def customizePerJob(self, job):
...
-        if highIOjob:
+        if job.get('highIOjob'):
             jdl.append('+Requestioslots = 1\n')
+        else:
+            jdl.append('+Requestioslots = 0\n')
 
         # Performance estimates
-        if job.get('estimatedJobTime'):
-            jdl.append('+MaxWallTimeMins = %d\n' % int(job['estimatedJobTime'] / 60.0))
         if job.get('estimatedMemoryUsage'):
```
amaltaro: From your HN, condor sets it to 1250MB in case it's not provided (logcollect, cleanup, harvesting(?)).
How about we default it to 1000MB?

For `estimatedDiskUsage`, I'd expect worker nodes to provide *at least* 4 GB, how about we default to that number (I don't remember the unit here though...)

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -976,22 +962,23 @@ def customizePerJob(self, job):
...
             jdl.append('request_memory = %d\n' % int(job['estimatedMemoryUsage']))
         if job.get('estimatedDiskUsage'):
             jdl.append('request_disk = %d\n' % int(job['estimatedDiskUsage']))
 
-        # Set up JDL for multithreaded jobs
-        if job.get('numberOfCores', 1) > 1:
-            jdl.append('machine_count = 1\n')
-            jdl.append('request_cpus = %s\n' % job.get('numberOfCores', 1))
+        # Max wall time estimate (12h if undefined)
+        jdl.append('+MaxWallTimeMins = %d\n' % int(job.get('estimatedJobTime', 12*3600))/60)
```
amaltaro: And here you changed the logic according to the original code `int(job['estimatedJobTime'] / 60.0)`

hufnagel: No, same logic if estimatedJobTime is defined for the job already. Just set it to 12h if it's not.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: Good idea. Any processing/production job that needs more should really define it elsewhere and pass it along with the job.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: Good idea.

amaltaro:
Dirk, for wallTime and requestDisk, I suggest you set the default value to the minimum in these settings. So, 1h and 1GB.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: Ah, good catch.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: About disk, the VO card says 20GB per core, should we use that instead ?

hufnagel:
I have a problem with that because that is the optimal way to make sure the jobs get run on a pilot with not much wall time left and will end up being rescheduled.

hufnagel:
And the disk could be a problem too. If you ask for 1GB and use 20GB, you can cause problems for other payloads running in the same pilot.

hufnagel:
My take here is that only a very small fraction of jobs won't have the performance estimates set inside the job (correct ?). As such, asking for a bit more resources won't affect the big picture much, but will make sure these jobs have a chance to run correctly.

If my initial assumption is wrong, maybe we should make sure that most jobs actual have these estimates set ?

hufnagel:
The impact of guessing too high is small, mainly not quite as efficient job scheduling as could be possible. The impact of guessing too low can be lots of job retries and internal Condor rescheduling and/or using up all the disk space in the pilot.

amaltaro:
Valid comments, Dirk. From my experience, only utilitarian jobs (cleanup, logcollect, possibly harvest) don't have job estimates.That's why I was shooting at lower default values to ease job scheduling.
That's why 12h estimated for these job types may not be optimal... but I agree that if it's too low, we have much more problems on the agent side.

So it would be 12h walltime and 20gb for disk, right?

hufnagel:
Yeah, 12h wall time (still only a quarter of pilot lifetime, so should be easy to schedule), 20GB disk per core (which is what each site should provide anyways) and the 1000MB for memory.

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
test this please

hufnagel:
Ok, this seems to work. Also just tested in a Tier0 replay. Jobs with different cpu requests and different highIOjon flags were bulk submitted together and it worked fine.

Now looking at the outstanding jdl parameters

         if job.get('taskName'):
             jdl.append('+WMAgent_SubTaskName = "%s"\n' % job['taskName'])

         if job.get('taskType'):
             jdl.append('+CMS_JobType = "%s"\n' % job['taskType'])

How about I just add an "Unknown" as default ? I still think this should actually always be set, but I don't want to dig through the code base to confirm this...

         if job.get('inputDataset'):
             jdl.append('+DESIRED_CMSDataset = "%s"\n' % 
job['inputDataset'])
         if job.get('inputDatasetLocations'):
             jdl.append('+DESIRED_CMSDataLocations = "%s"\n' % 
','.join(job['inputDatasetLocations']))

How about and empty string "" as default ?

         if self.useGSite:
             jdl.append('+GLIDEIN_CMSSite = \"%s\"\n' % (jobCE))

No idea. Is this guaranteed to be always set or do we need a default ?

         if job.get('proxyPath'):
             jdl.append('x509userproxy = %s\n' % job['proxyPath'])

Do we even still use this ? If potentially yes, how about an empty string "" again as default ?

         if job.get('requestName'):
             jdl.append('+WMAgent_RequestName = "%s"\n' % 
job['requestName'])
             m = GROUP_NAME_RE.match(job['requestName'])
             if m:
                 jdl.append('+CMSGroups = %s\n' % 
classad.quote(m.groups()[0]))

Default "Unknown" again ?

hufnagel:
@bbockelm any comments on how some of these parameters are used in the HTCondor side and what sensible defaults are ?

bbockelm:
`"Unknown"` is a fine default, I suppose.

We in fact _do_ use `x509userproxy` in the global pool from WMAgent: perhaps we don't use it for the T0 pool?

I have no clue what `GLIDEIN_CMSSite` is for or why we would set it.

Everything else looks familiar.

hufnagel:
We do use x509userproxy , but it is set by the schedd based on the X509 environment, not because the agent puts it there. We have never set this variable from the agent.

hufnagel:
Ok, so proposal:

Default "Unknown" for +WMAgent_SubTaskName, +CMS_JobType, +WMAgent_RequestName and +CMSGroups.

Remove setting x509userproxy (btw, CRAB3 doesn't use the agent submission plugins, does it ?).

What about +DESIRED_CMSDataset and +DESIRED_CMSDataLocations ? Empty string as suggested ok ? What is this used for ? Just monitoring or something functional ?

About +GLIDEIN_CMSSite, I am going to ask in HN about it what this is for...

amaltaro:
Don't remove `x509userproxy` please. We do set it in the agent environment and AFAIK we need this setting in both the agent code (to add the path/DN to the classAd) and to the condor schedd config.

I think we should not give `Unknown` as default value for:
WMAgent_SubTaskName, +CMS_JobType, +WMAgent_RequestName
instead let it break the component. It should always be present. If it's not, we need someone investigating it.

About `+CMSGroups`, when @bbockelm put it in WMCore code it was meant to be temporary until gWMS was able to handle it itself. Otherwise, Unknown should work be sufficient here.

hufnagel:
@amaltaro , you are not correct on this. No agent sets this. I checked.

amaltaro:
And how the schedd knows which file has to be shipped with these jobs?
I thought we needed to have X509 var in the system before starting the agent, such that the plugin would inherit this env var, e.g.:
```
cmst1@vocms0230:/data/srv/wmagent/current $ cat /data/admin/wmagent/env.sh | grep X509_USER
export X509_USER_CERT=/data/certs/servicecert.pem
export X509_USER_KEY=/data/certs/servicekey.pem
export X509_USER_PROXY=/data/certs/myproxy.pem
```

but I failed to find it in the code. If you are 100% certain it's not used, then go ahead. I'm not.

hufnagel:
I am 100% certain. The schedd picks it out of the environment the condor_submit is run in and you need a special schedd config parameter to do this (which is set for production agent machines schedd).

The agent itself has never cared what the X509 environment is set to.

hufnagel:
The x509userproxy code remnants are code leftovers from when CRAB was developed to run inside an agent. It would have been a request level parameter, passing the user proxy for the user submitting the request.

hufnagel:
Ok, following the discussion in HN and here, these are the next changes:

+WMAgent_SubTaskName, +CMS_JobType, +WMAgent_RequestName are always set using the information in the job. If the job doesn't have the info the component crashes (and we should check if this is a valid use case or not).

In the case +CMSGroups wasn't set, set it to undefined now. This is not a string value, but a literal with special meaning.

Same thing for +DESIRED_CMSDataset and +DESIRED_CMSDataLocations, instead of not setting them they will be set to undefined now.

I left x509userproxy alone for now (since frankly it doesn't matter one way or the other). If we ever clean this up we will have to remove a whole bunch of dead code, not just that two lines.

hufnagel:
test this please

hufnagel:
Next set of changes:

Had to change how request_memoy and request_disk are set. Some jobs have it set to None, which is not a valid value. For consistency reasons I also changed how +MaxWallTimeMins is set.

This version also gets rid of the sorting by sandbox. Tested that first and it seems to work.

hufnagel:
test this please

hufnagel:
@amaltaro , this is ready for some more testing on the agent side

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -954,44 +925,44 @@ def customizePerJob(self, job):
...
-        if job.get('numberOfCores', 1) > 1:
-            jdl.append('machine_count = 1\n')
-            jdl.append('request_cpus = %s\n' % job.get('numberOfCores', 1))
+        jdl.append('+Requestioslots = %d\n' % job.get('highIOjob', 0))
+
+        # Performance and resource estimates
+        numberOfCores = job.get('numberOfCores', 1)
+        requestMemory = int(job['estimatedMemoryUsage']) if job.get('estimatedMemoryUsage', None) else 1000
+        requestDisk = int(job['estimatedDiskUsage']) if job.get('estimatedDiskUsage', None) else 20*1000*1000*numberOfCores
+        maxWallTimeMins = int(job['estimatedJobTime'])/60.0 if job.get('estimatedJobTime', None) else 12*60
```
amaltaro: `requestDisk` and `maxWallTimeMins` are not reasonable at all. 
I think all the most important task types would have these estimates, leaving these default numbers only for logcollect, cleanup, etc. How about we set the default WT to 4h (maybe 8h then) and the disk to 5GB (10GB if you prefer).

amaltaro:
I'm gonna test it now. Please make a new commit if you make any changes besides the default values I suggested, for the moment.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
hufnagel: We had exactly he same arguments before :-). My take here is that without the job specifying things, you simply don't know and have to guess. Guessing correct isn't the goal, but guessing in a way to minimize the impact of getting it wrong.

If you guess too low on the disk, you run the risk of the job taking all the disk and if other jobs on the node are disk hungry too you could run out of disk. While guessing too high on disk really has not much of an impact up to the 20GB per core limit that we should always have anyways.

Same with the wall time. Yes, these jobs are probably short. But guessing too low here can mean your job gets scheduled on a pilot that can't run the job, it gets killed, re-scheduled etc. The 12h is reasonable long, but still only a quarter of the pilot lifetime. About 75% of all our pilots should be able to match a 12h job, so there isn't much of a downside to ask for 12h.

amaltaro: Hmm, I had this impression :) I agree with your first argument and playing safe is better. However, it will cause pilot inefficiencies (25% cause by walltime) and may be even worse for HLT farm and future opportunistic resources.

I'm just wondering whether it would be worthy to have 1 problematic job out of 1M, but taking the most efficiency approach. OR, having 0 problems but injecting not negligible efficiency to gwms... (?)

aperezca:
Hi Dirk, Alan,

I have been reading the discussion here about default parameter values. I would like to have a better understanding of what kind of jobs would carry those default values, and as this would affect payload to pilot scheduling performance, I'd prefer this be discussed together with the SI team. What do you think, should we try next Thu?

Antonio.

hufnagel:
There is not going to be an effect since almost all jobs will have non-default values configured. We are only talking about utility jobs here... Anyways, I won't be available Thursday.

aperezca:
Ok, it does not have to be this Thu. of course, and even we can discuss this by email or here… just to keep SI people aware of the potential impact of this.

So don't we have some properly defined values for utility jobs? Even more, have we been monitoring what the typical actual used values are?
 
Another question out of curiosity, I saw some defaults depending on number of cores, are we running any utility jobs using more than one?

hufnagel:
Merges can be multi-core. AlcaHarvesting and repack too, but they aren't strictly utility jobs.

hufnagel:
And I agree, all job types should declare how long they'll take. This is insurance. If we are sure that all utility jobs do declare their wall time and disk needs we could just remove the defaults and extract these from the job without fallback.

hufnagel:
Anyways, this will blow up the scope of this ticket. At the moment if a job doesn't declare wall time or disk, we don't put it into the job and HTCondor picks whatever default it wants (no idea what these are, do you Antonio ?). I think we can make a better job than HTCondor, so we should put something halfway sensible now and maybe open another issue to track down all jobs that don't declare it.

aperezca:
Dirk, my point was precisely that this discussion should happen outside the context of this ticket, with broader audience, given its potential impact. I think we'd need to
-identify the key scheduling parameters (cores, disk, memory, walltime, anything else?)
-know what type of jobs (and how many) do not request them explicitly (why?)
-find out what HTCondor is doing in that case
-propose hopefully better values for the defaults than those currently in use by HTCondor. We could even define different sets of defaults depending on the type of jobs.

hufnagel:
That discussion should not hold up this ticket. We pick sensible, maybe a bit too large defaults and we can discuss to our hearts content what the perfect values should be. Once we have them we can update the defaults in the code. I am not going to willing to wait for it. Anyways, feel free to discuss it in the SI meeting, you don't really need me for it at this stage.

I put my sensible (bit likely a bit on the high side) defaults in the code, I think they'll work just fine.

aperezca:
Ok, I agree, ideal values should not prevent us from setting reasonable values now.

amaltaro:
Changes look good to me. I also ran a bunch of workflows (including a large enough multicore request) together and didn't spot any issues.

@ticoann it's good to be merged from my side

amaltaro:
test this please

PR #6943, opened 2016-06-17
Update deprecated cherrypy key

amaltaro:
`ENGINE WARNING: Use of engine.autoreload_on is deprecated and will be removed in a future version. Use engine.autoreload.on instead.`

So updating it now.


amaltaro:
Tested in my VM.

PR #6944, opened 2016-06-20
make checksum use with gfal2 optional, fixes #6591

### LINKED ISSUE #6591: Stage-out into S3
S3 does not support adler32 checksum and so the stage-out into the data bridge for CMS@home is failing. As far as I can see, the checksum check is hardcoded in the gfal plugin (https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/Plugins/GFAL2Impl.py#L44). 

Does it make sense to enable adler32 checksum based on the protocol? for e.g if it s3, disable the check.

hufnagel:
Add a '--nochecksum' option that can be set in site-local-config and which triggers skipping using the adler32 checksum in gfal-copy. All other options are ignored and passed on to the gfal-copy directly.

### DIFF
--- a/src/python/WMCore/Storage/Backends/GFAL2Impl.py
+++ b/src/python/WMCore/Storage/Backends/GFAL2Impl.py
@@ -4,6 +4,9 @@ _GFAL2Impl_
 Implementation of StageOutImpl interface for gfal-copy
 """
 import os
+import argparse
+import logging
+
 from WMCore.Storage.Registry import registerStageOutImpl
 from WMCore.Storage.StageOutImpl import StageOutImpl
 from WMCore.Storage.Execute import runCommandWithOutput as runCommand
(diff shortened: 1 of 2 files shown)


hufnagel:
@amaltaro , I have no way to test this. Tested as much as I could replicating in standalone tests, but actual gfal2 stageout tests would need to be done by you I guess. Also, tests for the new feature to skip the checksum would have to be done by @HassenRiahi or whoever is taking over the cms@home tests...

amaltaro:
@hufnagel WMAgent does not use these `Plugins` v2 implementation, instead it uses the `Backends` one.
In addition to that, if you want to use the `options` field in the site-local-config.xml, why do you handle this option as a command line argument? I failed to see how that would work. Instead, we should probably parse this option in this method:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/StageOutMgr.py#L80

and then propagate it down to the actual stage out implementation.

hufnagel:
I handle them with the argument parser, because that's what the option flag is supposed to be, arguments to be passed to the stageout command. In this case here it's just that one of the arguments isn't passed directly, it affects how we call gfal-copy. Think of it as an extra argument to gfal-copy implemented on top of gfal-copy in our code that calls gfal-copy.

The arguments are highly implementation specific, they have nothing to do with the general stageout implementation, they are always very command specific.

I'll add the an implementation for the normal v1 stageout plugins (ie. backend) as well though.

amaltaro:
Now that critical problems are mostly sorted out, giving this patch a try before it gets merged. I think Purdue is a good candidate for this test

amaltaro:
Patch looks good. I ran 3 workflows in Purdue and got no failures (but logcollect).
@ticoann please merge

amaltaro:
test this please

PR #6942, opened 2016-06-16
Remove admin interface

vkuznet:

PR #6946, opened 2016-06-20
Don.t share db handler object among APIs; retrieve single doc always …

### LINKED ISSUE #6764: request search intermitently give error 500
https://cmsweb.cern.ch/reqmgr2/data/request?status=assignment-approved&details=true

 is failing very often

amaltaro:
…when possible
Together with #6942, partially fixes #6764 

As spotted and suggested by Valentin, create a different db api object for each Rest API.
Second change is to avoid making a POST call to couchdb/_all_docs when the user is requesting a single doc.

The `500 Internal Server Error` is still showing up in the logs, though in a much smaller rate (< 15%)...
still investigating ...

@vkuznet please review
I still need to test it with the new tag in a clean environment

### DIFF
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -21,18 +21,18 @@ import tempfile
 import traceback
 import urllib
 import urlparse
+import types
 from httplib import HTTPException
 from json import JSONEncoder, JSONDecoder
 
 from WMCore.Algorithms import Permissions
+from WMCore.Lexicon import sanitizeURL
 from WMCore.WMException import WMException
 from WMCore.Wrappers.JsonWrapper.JSONThunker import JSONThunker
-
 try:
     from WMCore.Services.pycurl_manager import RequestHandler, ResponseHeader
 except ImportError:
     pass
-from WMCore.Lexicon import sanitizeURL
 
 def check_server_url(srvurl):
     """Check given url for correctness"""
@@ -134,7 +134,6 @@ class Requests(dict):
         """
         Wrapper around request helper functions.
         """
-
         if  self.pycurl:
             result = self.makeRequest_pycurl(uri, data, verb, incoming_headers,
                          encoder, decoder, contentType)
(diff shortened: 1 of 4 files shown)


src/python/WMCore/Services/Requests.py
```diff
@@ -236,7 +227,7 @@ def f():
 
         headers["Content-length"] = str(len(encoded_data))
 
-        assert isinstance(encoded_data, type('string')), \
+        assert isinstance(encoded_data, str), \
```
vkuznet: please always use basestring instead of str, it covers both use case a string and unicode.

vkuznet:
Alan, changes are fine, but does not really solve the problem. For instance you may further improve your PR by converting makeRequest_httplib and  _getURLOpener methods https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L358
to a stand-alone functions.

vkuznet:
oops, here is link to makeRequest_httplib:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L168

src/python/WMCore/Services/Requests.py
amaltaro: From the previous check, I assumed we'd be interested only in 8-bit strings, not in unicode strings. Ok, I'll update the code soon

amaltaro:
Giving it a try at pycurl...

amaltaro:
several of them failed while locating data from PhEDEx.
I saw a couple others failing with http code 411, e.g.:
https://cmssdt.cern.ch/jenkins/job/DMWM-WMCore-PR-test/2492/testReport/junit/WMCore_t.RequestManager_t.ReqMgr_t/ReqMgrTest/testK_CheckRequestFailsInjectionForbiddenInputArg/

however, I do see `Content-Length` being provided in the headers, according to the same logs above.

I'll remove the last commit and keep only the initial changes proposed in this PR. Further changes will go in a new PR.

amaltaro:
As discussed with Valentin, further improvements to make it thread safe will be done in a new PR (which will require much more testing), they are basically the split up of some of these methods unsafe into functions.

vlimant:
Will this go in July deployment ?

amaltaro:
Yes

PR #6947, opened 2016-06-22
Make sure to always retrieve the long hostname

amaltaro:
It seems gethostname() may return a different output depending on the OS version.
Testing right now


amaltaro:
Tested during WMAgent deployment today

PR #6958, opened 2016-06-28
Minor improvements on prio and status in reqmgr2

### LINKED ISSUE #6956: ReqMgr2: status update via  web UI is broken
It returns a status Ok, however no changes were made to the request itself. The reason seems to be coming from the fact that you always _update_ the request priority as well, this is what is sent downstream from web interactions:
`{"RequestStatus":"aborted","RequestPriority":"250000"}`

which then does not end in an statusUpdate in Request.py, e.g.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L508

I'm still checking on how to fix it.

amaltaro:
Was meant to fix #6956, but in the end I don't think changing the status and the priority is a valid use case (besides when a request is created and assigned).

Anyways, fix the RequestPriority type and handleAssignment method to be called only when a request is being assigned.

Needs to be tested. @ericvaandering @vkuznet feel free to review :)

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -72,17 +72,14 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
     
     TODO: rasie right kind of error with clear message 
     """
-
-    request_name = request_args["RequestName"]
-    # this need to be deleted for validation
-    del request_args["RequestName"]
+    # this needs to be deleted for validation
+    request_name = request_args.pop("RequestName")
     couchurl =  '%s/%s' % (config.couch_host, config.couch_reqmgr_db)
     workload = WMWorkloadHelper()
-    # param structure is RESTArgs structure.
     workload.loadSpecFromCouch(couchurl, request_name)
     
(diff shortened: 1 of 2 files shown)


src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -72,17 +72,14 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
     
     TODO: rasie right kind of error with clear message 
     """
-
-    request_name = request_args["RequestName"]
-    # this need to be deleted for validation
-    del request_args["RequestName"]
+    # this needs to be deleted for validation
+    request_name = request_args.pop("RequestName")
```
vkuznet: The pop method will throw exception if key does not exists, therefore it's better to be safe and but try/except KeyError around it. But try/except is costly, and it's better just to get request name and then delete it.

amaltaro: I didn't worry because we were directly accessing that key before, an exception would have happened way before.
Anyways, this validate func always receive a RequestName key/value.

amaltaro:
Good in my VM

PR #6959, opened 2016-06-29
Fix issue 6956

vkuznet:


amaltaro:
I was about to report that it was not working, but I just noticed the doc.tmpl patch didn't apply.
It was applied and things are working, thanks!

amaltaro:
Works well in my VM

PR #6965, opened 2016-07-06
add BPHSkim to StdBase

hufnagel:


src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -138,7 +138,8 @@ def determineOutputModules(self, scenarioFunc=None, scenarioArgs=None,
                                       'SingleTrack': 'AOD',
                                       'MinBias': 'AOD',
                                       'OniaUPC': 'RAW-RECO',
-                                      'HighMET': 'RECO'
+                                      'HighMET': 'RECO',
+                                      'BPHSkim': 'USER'
                                      }
                     dataTier = skimToDataTier.get(physicsSkim, 'USER')
```
ericvaandering: Why isn't this line sufficient for what you are doing in line 142? Also, a little trick, if you add a comma to the end of line 142 it does not impact the python but if you then need to add another skim later, you don't need to "delete" a line.

ericvaandering:
@amaltaro this 1L change you made is going to "fail" every single futurize check on every PR from now on. Why not use the actual type instead of type(1L)?

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
hufnagel: Trick works but looks a little ugly IMO. Plus do we really care if a patch changes one or two lines ?

Same sentiment for the default to tier USER. In principle correct, but I still like to keep a list of all know physics skims in one place. Also, most physics skims don't have data tier USER and I don't want to get into the habit of "we can just add the new physics skim to the config without adding it to the code". That default is to catch mistakes, not as something intentionally used.

ericvaandering: I disagree about the style, but as you wish. 

Agree on the BPHSkim part.

PR #6968, opened 2016-07-06
Fix hiding split window

vkuznet:


amaltaro:
Thanks, Valentin. It seems that was the issue.

amaltaro:
Ok, merging to get it into testbed tomorrow.

PR #6572, opened 2016-02-10
initial patch for abort and force-complete handling

ticoann:
Need to add WQ element manipulation which distinguishes aborted from force complete work

PR #6970, opened 2016-07-08
Complete/kill everything for an aborted workflow

amaltaro:
Well, I noticed only a minute ago that there is a patch being discussed already for this case, @ticoann made #6572 ... :(

I didn't notice killWorkflow() was called for both 'aborted' and 'force-complete' workflows. It would be easier to just kill/complete everything for workflows having this state transition, as well as less error prone. However, I do agree that it may be useful to have `cleanup` and `logcollect` jobs running for `force-completed` workflows, though I don't know whether people would really care if they lose some of these remaining jobs...

@ericvaandering @ticoann what do you guys think?


amaltaro:
Since Seangchan decoupled workflow completion from these utilitarian jobs, cleanup and logcollect became less annoying, though there may be requests staying in the system if sites go out of Normal and jobs need to run there.

amaltaro:
I think this PR is no longer relevant, so closing it.

PR #6971, opened 2016-07-10
make sure AlcaHarvest CMSSW jobs only use a single core

hufnagel:
The SiPixelAli AlcaHarvest jobs are assigned 4 cores since they need lots of disk space. The AlcaHarvest CMSSW job itself can only run single-threaded though, so make sure it only uses one core.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -568,6 +568,20 @@ class SetupCMSSWPset(ScriptInterface):
 
         return
 
+    def handleAlcaHarvestSettings(self):
+        """
+        _handleAlcaHarvestSettings_
+
+        AlcaHarvest jobs should only use one core in CMSSW
+        """
+        try:
+            if int(self.step.data.application.multicore.numberOfCores) > 1:
+                self.step.data.application.multicore.numberOfCores = 1
+        except AttributeError:
+            pass
+
+        return
+
     def handleSpecialCERNMergeSettings(self, funcName):
         """
         _handleSpecialCERNMergeSettings_
@@ -642,6 +656,8 @@ class SetupCMSSWPset(ScriptInterface):
                 self.handleRepackSettings()
             elif funcName == "merge":
                 self.handleMergeSettings()
+            elif funcName == "alcaHarvesting":
+                self.handleAlcaHarvestSettings()
 
             if socket.getfqdn().endswith("cern.ch"):
                 self.handleSpecialCERNMergeSettings(funcName)


amaltaro:
Code looks good. However, I'd suggest making a single `handleNumberOfCores` method for `if funcname in ['merge', 'repack', 'alcaHarvesting']. Then from this `if` you make another call for the repack tweak (without the multicore code).

Eventually we should fix it in the source (spec file) and set it to 1 for any job types or tasks that need to run as single core.

hufnagel:
We can't "fix" it in the spec. The spec has them as single core jobs, but that is overridden in the job splitting, from that point on they are real multicore jobs. Multicore jobs whose CMSSW step only uses a single core, hence the needed override in the runtime CMSSW code.

Also, since you suggest a functional change/refactoring, can you open another issue on it later ? Or I can just make the change later. I do not want to change this version since I already pointed Tier0Ops at it and AFAIK it is actively being used.

amaltaro:
Ahh, this alcaHarvesting is one step in a multi-steps job? Or is it a single job but one that you want to deceive gWMS as multicore?

About this patch, isn't the change simple enough that you can make it right from the beginning? Sorry for being picky, but the more I look at the WMCore code, the more I want to get rid of complications, duplications, very specific exceptions and patches everywhere.

hufnagel:
Correct on the latter. I want gWMS to schedule it multicore, but the CMSSW step can only run single-threaded so far.

This is complicated, but it can't be done any other way at the moment. We would have to change the disk space kill expressions to change this (kill on what a job reserved, not the 20GB*N + 7GB rule). Problem is that disk space estimates are so unreliable that this would be dangerous. For our workflows to function at all we rely on them having plenty of local disk space and that is allocated by the famous 20GB per core rule we ask of the sites. I just don't see how we can fit single core jobs that ask for and use 100GB of local disk into our job mix without total screwups and messes.

hufnagel:
Which effectively means if you have a job that you know needs 100GB of local disk, you need to submit it as a 5 core gWMS job, no matter what the actual payload uses in terms of cores. Only way to do it.

hufnagel:
@ticoann this seems to have been forgotten somewhat ?

ticoann:
I think Alan just wanted to make duplicated code to simple since the 3 methods are almost the same. (But I might be wrong). For me, I was just wondering 
following part
```
exception:
    pass
```
Is there nothing need to be done when exception occurs?

hufnagel:
If there is an exception it means that self.step.data.application.multicore.numberOfCores doesn't exist, which is fine, since then the job is already singlecore (that parameter not existing is historically interpreted as singlecore in our code).

hufnagel:
We can wait merging this until Monday, I can review this and see if I can simplify the methods...

ticoann:
test this please

ticoann:
Dirk, do you need a new tag with this?

hufnagel:
Ok, looked at the code and I can simplify this. I'd rather do it in a new commit and pull request though since I don't want to modify this one (we are already using it in the Tier0). Which means this should get merged and I'll need a new tag.

ticoann:
Hi Dirk, I will merge it and create the new tag then. In general we freeze the merge during testbed validation since code might affect in applications in cmsweb. (reqmgr, etc) 
But in this case, it won't. Anyway, I think we should maintain Tier0 branch for T0 agent upgrade as we briefly discussed in the release meeting.

PR #6973, opened 2016-07-11
reduce RECO merges by allowing PromptReco to stageout to merged

hufnagel:
Sigh. Really thought I had removed this earlier. This was needed when we were still doing event splitting in PromptReco, now it just completely blows up the number of (and data volume) we pipe through RECO merges.


amaltaro:
test this please

hufnagel:
Closing this, it should no longer be considered for merge. We are currently using this in PromptReco but it has problems. About to replace it with a different pull request.

PR #6980, opened 2016-07-14
Make sure file:// is added before source/dest pfn in gfal2 plugin

juztas:
1. Created a new function createFinalPFN, which will take care of preparing pfn and appending file:// if needed
   a) Also added a check if pfn starts with '/' and in that case append file://
2. any function `createStageOutCommand` or `createRemoveCommand` or `removeFile` uses new function for preparing pfn. In any case if WMA or CRAB or whatever would call this Stageout plugin without correct pfns, it enforces pfn modification by calling createFinalPFN function.

### DIFF
--- a/src/python/WMCore/Storage/Backends/GFAL2Impl.py
+++ b/src/python/WMCore/Storage/Backends/GFAL2Impl.py
@@ -28,17 +28,35 @@ class GFAL2Impl(StageOutImpl):
         self.removeCommand = self.setups % '. $JOBSTARTDIR/startup_environment.sh; printenv; date; gfal-rm -vvv -t 600 %s '
         self.copyCommand = self.setups % '. $JOBSTARTDIR/startup_environment.sh; printenv; date; gfal-copy -vvv -t 2400 -T 2400 -p %(checksum)s %(options)s %(source)s %(destination)s'
 
-    def createSourceName(self, protocol, pfn):
+
+    def createFinalPFN(self, protocol, pfn):
         """
-        _createSourceName_
-        GFAL2 uses file:/// urls
+        _createFinalPFN_
+        GFAL2 requires file:/// for any direction transfers
         """
         if pfn.startswith('file:'):
             return pfn
         elif os.path.isfile(pfn):
             return "file://%s" % os.path.abspath(pfn)
-        else:
-            return pfn
+        elif pfn.startswith('/'):
+            return "file://%s" % os.path.abspath(pfn)
+        return pfn
+
+
+    def createSourceName(self, protocol, pfn):
+        """
+        _createSourceName_
+        GFAL2 uses file:/// urls
+        """
+        return self.createFinalPFN(protocol, pfn)
+
+
+    def createTargetName(self, protocol, pfn):
+        """
+        _createTargetName_
+        GFAL2 uses file:// urls
+        """
+        return self.createFinalPFN(protocol, pfn)
 
 
     def createOutputDirectory(self, targetPFN):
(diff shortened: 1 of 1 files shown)


src/python/WMCore/Storage/Backends/GFAL2Impl.py
```diff
@@ -28,17 +28,35 @@ def __init__(self, stagein=False):
...
+    def createFinalPFN(self, protocol, pfn):
         """
-        _createSourceName_
-        GFAL2 uses file:/// urls
+        _createFinalPFN_
+        GFAL2 requires file:/// for any direction transfers
         """
         if pfn.startswith('file:'):
             return pfn
         elif os.path.isfile(pfn):
```
amaltaro: I thought isfile() would work only for regular files, not for remote ones. Can you x-check this works as expected?
BTW, isn't your `elif pfn.startswith('/'):` line below doing exactly the same check as above?

amaltaro:
These 3 new methods don't use `protocol` at all, so I'd suggest to remove it

src/python/WMCore/Storage/Backends/GFAL2Impl.py
juztas: What you mean by remote file? with hostname or only /store/mc/... ?
Not the same.
```
os.path.isfile(path)
Return True if path is an existing regular file. This follows symbolic links, so both islink() and isfile() can be true for the same path.
```
'/' needed if you would do gfal-copy srm://....file.root /store/user/... in that case the destination file does not exist and you need to append file:// in beginning. isfile will be false in this case, but '/' not and will add file:// 

In the end, any source or destination file which starts with:
srm:// or gridftp:// or anyOtherProcol:// will be returned as it is
file: will be returned as it is
fileName.root will return '"file://%s"  % abspath(fileName.root)
/store/fileName.root will return 'file:///store/fileName.root'

and what LogCollect does it sets destination like that:
```
gfal-copy -vvv -t 2400 -T 2400 -p   gsiftp://red-gridftp.unl.edu//mnt/hadoop/user/uscms01/pnfs/unl.edu/data4/cms/store/unmerged/logs/prod/2016/6/30/pdmvserv_task_HIG-RunIISpring16DR80-01259__v1_T_160620_125903_751/HIG-RunIISpring16DR80-01259_0/0012/0/54670436-3c68-11e6-900f-001e67abefa8-19-0-logArchive.tar.gz /var/lib/condor/execute/dir_2916738/glide_2t8wVK/execute/dir_40989/job/WMTaskSpace/logCollect1/54670436-3c68-11e6-900f-001e67abefa8-19-0-logArchive.tar.gz
```
So now it is not correct and with this fix it will have file:// in beginning of dest file.

juztas:
you cant from two, only from one new one which is createFinalPFN.

hufnagel:
Btw, @amaltaro , please merge #6944 before this since there are likely conflicts otherwise.

juztas:
yes, missed that. I will rebase after that, just ping me.

amaltaro:
Testing it on vocms0230

amaltaro:
Justas, can you test it in CRAB as well?
I've just ran a few workflows at Purdue and everything looks good. Thanks.

@ticoann good to merge from my side

amaltaro:
test this please

amaltaro:
Justas, as foreseen before, can you rebase this PR. Please, also get rid of the `protocol` unused argument on your 3 new methods.

amaltaro:
@juztas can you rebase this PR please?

juztas:
yes, today. I was on holiday

PR #6985, opened 2016-07-15
Catch deadlock on oracle as well...

amaltaro:
Sigh, sorry for my handicap fix. I've just noticed it catches deadlock on mariadb only. Code updated once again, but IMO we don't need to backport it again, I suggest we just wait for the next release.

@ticoann please review :-)

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
@@ -338,7 +338,7 @@ class PhEDExInjectorPoller(BaseWorkerThread):
                 try:
                     self.setStatus.execute(lfnList, 1)
                 except Exception as ex:
-                    if 'Deadlock found when trying to get lock' in str(ex):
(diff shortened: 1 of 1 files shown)


ticoann:
Thanks Alan

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -338,7 +338,7 @@ def injectFilesPhEDExCall(self, location, injectData, lfnList):
                 try:
                     self.setStatus.execute(lfnList, 1)
                 except Exception as ex:
-                    if 'Deadlock found when trying to get lock' in str(ex):
+                    if 'Deadlock found' or 'deadlock detected' in str(ex):
```
ericvaandering: Except this doesn't work (well, it will always trigger):

```
>>> a = 'blah'
>>> 'blah' or 'Blah' in a
'blah'
```

amaltaro: That's why I left on vacation ... LoL 
See #7029

PR #6990, opened 2016-07-16
create a working xrdcp stageout plugin, fixes #6989

### LINKED ISSUE #6989: Stop using EOSCMS SRM for LogCollect stageout
I've been talking with the EOSCMS team about working to eliminate the SRM endpoint.  (They'd like to get rid of SRM and OSG would like to retire support for bestman2).

In the agent, we currently hardcode the stageout to EOSCMS SRM.  We should remove this.  The two obvious options are:
1.  Switch to `gfal2` and `gsiftp://` stageout.
2.  Switch to `xrdcp` stageout for LogCollect.

I think Dirk is considering (2), but I wanted to record this as a ticket.

@amaltaro @hufnagel

hufnagel:
Create a basic xrdcp stageout plugin. Use it for LogCollect stageout to EOS and Castor (the latter needs special options). Also fix a bug in the StageOutMgr exception handling. In addition convert the top-level splitPFN function in StageOutImpl into a static class method (had problems with the top-level function).

@amaltaro , @ticoann , please review.

Patch has been tested (both for general stageout with some hacking and for LogCollect), although only at CERN (only resources I have easy access to). Would need to be tested elsewhere too (especially Fermilab since I modified the splitPFN call in their stageout plugin).

Still leaving the wrong PNN for the castor stageout in, we can look at that later.

### DIFF
--- a/src/python/WMCore/Storage/Backends/FNALImpl.py
+++ b/src/python/WMCore/Storage/Backends/FNALImpl.py
@@ -8,7 +8,7 @@ Implementation of StageOutImpl interface for FNAL
 from __future__ import print_function
 import os
 from WMCore.Storage.Registry import registerStageOutImpl
-from WMCore.Storage.StageOutImpl import StageOutImpl, splitPFN
+from WMCore.Storage.StageOutImpl import StageOutImpl
 from WMCore.Storage.Backends.LCGImpl import LCGImpl
 
 _CheckExitCodeOption = True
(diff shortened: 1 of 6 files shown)


hufnagel:
Tests at CERN were on Tier0 VM, which don't use glexec or kerberos. Straight X509 authentication, so really shouldn't be any different from running this anywhere else.

hufnagel:
The old xrdcp stageout plugin was completely non-functional and no one used it. One greek site had it in it's site-local-conf, but that only worked because they also had a fallback rule. Site support contacted them and they elevated their fallback to be the default stageout rule.

hufnagel:
Slight change, remove the use of a legacy (ie. deprecated) option for castor support and do it a different way. Still need to run a test on it though.

hufnagel:
Ok, new version is tested as well, works.

hufnagel:
test this please

amaltaro:
Testing in vocms0230 as well. Dirk, make a new commit if you need to make further changes please

amaltaro:
@ticoann it's also good to go.
I ran a bunch of workflows at CERN/FNAL - including some at Purdue, that was failing logcollect jobs - and it all looks good, files can be seen in EOS and etc.

amaltaro:
test this please

PR #6993, opened 2016-07-19
Validate RequestName for reqmgr2 as well

amaltaro:
As reported here: https://cms-logbook.cern.ch/elog/Workflow+processing/25012

there is no validation for RequestName field in ReqMgr2, which allows it to get in but fails during workqueue PUT request (to update the status).

With this fix, it should use the same validation function as used by ReqMgr. We probably don't need to have this limit anymore in ReqMgr2, since Oracle is gone over there. However, while we have both reqmgr in production, it sounds better to keep them in sync.

@ticoann please review. I already tested in my VM and it throws a

```
        <h2>400 Bad Request</h2>
        <p>Invalid spec parameter value: 'amaltaro_RVCMSSW_8_0_13RunVRZeroBiasErr2016C__IterMedianAlgo_rsb_RelVal_ZBiasErr2016C_160719_135037_5457' does not match regular expression [a-zA-Z0-9\s\.\-_]{1,100}$</p>
        <pre id="traceback"></pre>
```

if the check fails.

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -2,10 +2,10 @@
 Unlike ReqMgr1 defining Request and RequestSchema classes,
 define just 1 class. Derived from Python dict and implementing
 necessary conversion and validation extra methods possibly needed.
-    
+
 TODO/NOTE:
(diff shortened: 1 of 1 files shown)


ticoann:
patch looks good to me too.

PR #6994, opened 2016-07-19
more fine-grained options to force unmerged and use them in PromptReco

hufnagel:
Allowing to override all outputs to either merged or unmerged is not enough. Add a way to go with the default (straight to merge enabled for all outputs) but force specific output modules to always write unmerged. Use this to force PromptReco to write ALCARECO output always unmerged.

@ticoann @amaltaro please review. I think this will need some intense testing since it touches the guts of the system. AFAIK WMAgent doesn't use this functionality though, so we have time for testing. Tier0 testing and deployment can follow a different schedule.

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -1052,29 +1052,33 @@ class Report:
 
         return filteredResults
 
-
     def getAllFileRefsFromStep(self, step):
         """
         _getAllFileRefsFromStep_
 
-        Retrieve a list of all files produced in a step.  The files will be in
-        the form of references to the ConfigSection objects in the acutal
-        report.
+        Retrieve all files produced in a step.  Return it as a dictionary keyed
+        of the output module since the stageout code has some output module
+        specific settings (force unmerged output).
+
+        The files will be in the form of references to the ConfigSection
+        objects in the acutal report.
         """
         stepReport = self.retrieveStep(step=step)
         if not stepReport:
-            return []
+            return {}
 
         outputModules = getattr(stepReport, "outputModules", [])
-        fileRefs = []
+        fileRefs = {}
         for outputModule in outputModules:
+
+            fileRefs.setdefault(outputModule, [])
+
             outputModuleRef = self.getOutputModule(step=step, outputModule=outputModule)
 
             for i in range(outputModuleRef.files.fileCount):
-                fileRefs.append(getattr(outputModuleRef.files, "file%i" % i))
+                fileRefs[outputModule].append(getattr(outputModuleRef.files, "file%i" % i))
 
-        analysisFiles = self.getAnalysisFilesFromStep(step)
-        fileRefs.extend(analysisFiles)
+        fileRefs['analysis'] = self.getAnalysisFilesFromStep(step)
 
         return fileRefs
 
@@ -1236,9 +1240,10 @@ class Report:
 
         fileRefs = []
         for step in self.data.steps:
-            tmpRefs = self.getAllFileRefsFromStep(step=step)
-            if len(tmpRefs) > 0:
-                fileRefs.extend(tmpRefs)
+            fileDict = self.getAllFileRefsFromStep(step=step)
+            for tmpRefs in fileDict.values():
+                if len(tmpRefs) > 0:
+                    fileRefs.extend(tmpRefs)
 
         return fileRefs
 
--- a/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
@@ -62,7 +62,8 @@ class PromptRecoWorkloadFactory(StdBase):
                                                splitAlgo = self.procJobSplitAlgo,
                                                splitArgs = self.procJobSplitArgs,
                                                stepType = cmsswStepType,
-                                               forceUnmerged = True)
+                                               forceUnmerged = ["write_ALCARECO"] if 'ALCARECO' in self.writeTiers else False)
+
         if self.doLogCollect:
             self.addLogCollectTask(recoTask)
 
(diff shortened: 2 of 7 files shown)


hufnagel:
Ok, since I made changes to the JobReport class I think these JobReport unit test failures could be real, have to investigate...

hufnagel:
Ok, unit tests look better now. The other failures there are bogus.

hufnagel:
Just rebased this to be compatible with HEAD again (there was a conflict with the SE->PNN changes)

amaltaro:
This will have to wait for at least a few days more because I don't want to mix too many new things with testbed validation as well.

hufnagel:
Ok, Sep 2 is more than a few days ago. What's the hold-up with this ? We have been using this in the Tier0 since a couple months now.

src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -1049,29 +1049,34 @@ def getAnalysisFilesFromStep(self, step):
...
-            return []
+            return {}
 
         outputModules = getattr(stepReport, "outputModules", [])
-        fileRefs = []
+        fileRefs = {}
         for outputModule in outputModules:
+
+            if not outputModule in fileRefs:
+                fileRefs[outputModule] = []
```
amaltaro: These 2 lines could be replaced by `fileRefs.setdefault(outputModule, [])`

src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -1233,9 +1238,10 @@ def getAllFileRefs(self):
 
         fileRefs = []
         for step in self.data.steps:
-            tmpRefs = self.getAllFileRefsFromStep(step=step)
-            if len(tmpRefs) > 0:
-                fileRefs.extend(tmpRefs)
+            fileDict = self.getAllFileRefsFromStep(step=step)
+            for tmpRefs in fileDict.values():
+                if len(tmpRefs) > 0:
```
amaltaro: And I think this if is not really needed. In the worst case it will extend []

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
```diff
@@ -55,14 +55,24 @@ def buildWorkload(self):
...
-                                               forceUnmerged = True)
+        if 'ALCARECO' in self.writeTiers:
+            recoOutMods = self.setupProcessingTask(recoTask, taskType, self.inputDataset,
+                                                   scenarioName = self.procScenario,
+                                                   scenarioFunc = "promptReco",
+                                                   scenarioArgs = scenarioArgs,
+                                                   splitAlgo = self.procJobSplitAlgo,
+                                                   splitArgs = self.procJobSplitArgs,
+                                                   stepType = cmsswStepType,
+                                                   forceUnmerged = [ "write_ALCARECO" ])
```
amaltaro: To avoid duplicating this code, how about you do something like:
`forceUnmerged = ["write_ALCARECO"] if 'ALCARECO' in self.writeTiers else False)`

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -367,12 +367,14 @@ def setupProcessingTask(self, procTask, taskType, inputDataset=None, inputStep=N
         procTaskCmsswHelper.setOverrideCatalog(self.overrideCatalog)
         procTaskCmsswHelper.setErrorDestinationStep(stepName=procTaskLogArch.name())
 
-        if forceMerged:
+        if forceMerged and isinstance(forceMerged, bool):
             procTaskStageHelper.setMinMergeSize(0, 0)
-        elif forceUnmerged:
+        elif forceUnmerged and isinstance(forceUnmerged, bool):
```
amaltaro: It seems `forceMerged` and `forceUnmerged` are supposed to be either `False` or a list. Which makes me think we should not check for boolean type here. 

Instead, I'd keep the simple checks and default these args to None in the method signature. The same for `def addOutputModule`. The code will look more readable IMO.

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
```diff
@@ -129,102 +129,98 @@ def execute(self, emulator = None):
...
+                                                              step = step,
+                                                              outputName = outputName)
                         except Exception as ex:
-                            logging.error("Encountered error while handling LFN for merge due to size.\n")
+                            logging.error("Encountered error while handling LFN for merge.\n")
                             logging.error(str(ex))
                             logging.debug(fileName)
-                            logging.debug("minMergeSize: %s" % self.step.output.minMergeSize)
+                            logging.debug("minMergeSize: %s" % getattr(self.step.output, 'minMergeSize', None))
+                            logging.debug("maxMergeEvents: %s" % getattr(self.step.output, 'maxMergeEvents', None))
```
amaltaro: Since we lack of good logging/report on the stageout land, it's likely a good idea to move these two min/maxMerge debug logging to the beginning of the exception block AND make it an info log level.

src/python/WMCore/FwkJobReport/Report.py
hufnagel: No, this is needed since self.getAllFileRefsFromStep returns a dict now, not a list.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
hufnagel: How is that supposed to work ? Boolean or List for forceUnmerged triggers completely different code paths, so I need to check for type at some point...

hufnagel:
@amaltaro , made some changes and replied to two of the comments where I do not want to make changes.

hufnagel:
@amaltaro , can you review again ? We've been using this in the Tier0 this year, but OTOH merging this now won't make a difference for this years Tier0 processing and building a new Tier0 release for next year isn't all that urgent. So if you still have some major concerns or ideas how to do this different;y/simpler, I am willing to rework this (it's functional and I need it, but I don't like it all that much either...).

hufnagel:
Coming back to this, I want to address this issue one way or another before we start data taking in 2017.

hufnagel:
Superseeded by #7934

PR #6995, opened 2016-07-19
Remove reference to 1L which futurize does not like

ericvaandering:


amaltaro:
Just to keep it consistent, how about we do the same for the string, int and double types?

ericvaandering:
Well, I think this is not really "our" code, so the fewer changes we make the better.

ericvaandering:
test this please

amaltaro:
Looks good then.

PR #7004, opened 2016-07-23
JobSubmitter refactoring; always submit higher prio first

amaltaro:
Initial patch for the JobSubmitter refactoring. As discussed in several occasions, I made changes such that highest priorities jobs will be submitted first (following by its timestamp).

I also removed a bunch of things that looked useless and overcomplicated. Hopefully this will solve the last bit of the so annoying/called 'job inversion'...

@ticoann very basic tests look ok. Running all the Integration tests now.

### DIFF
--- a/bin/wmagent-resource-control
+++ b/bin/wmagent-resource-control
@@ -167,13 +167,12 @@ def printThresholds(resourceControl, desiredSiteName):
         print "%s - %d running, %d pending, %d running slots total, %d pending slots total%s:" % (
                         siteName, runningJobs, pendingJobs, runningSlots, pendingSlots, stateMsg)
 
-        for task in siteThresholds['thresholds']:
+        for task, thres in siteThresholds['thresholds'].iteritems():
             print "  %s - %d running, %d pending, %d max running, %d max pending, priority %s" % \
-                  (task['task_type'], task["task_running_jobs"],
-                   task['task_pending_jobs'], task["max_slots"],
-                   task["pending_slots"], str(task.get("priority", 1)))
-            if task["task_running_jobs"] > 0:
-                taskWorkloads = resourceControl.listWorkloadsForTaskSite(task['task_type'], siteName)
+                  (task, thres["task_running_jobs"], thres['task_pending_jobs'], thres["max_slots"],
+                   thres["pending_slots"], str(thres.get("priority", 1)))
+            if thres["task_running_jobs"] > 0:
+                taskWorkloads = resourceControl.listWorkloadsForTaskSite(task, siteName)
                 for t in taskWorkloads:
                     print "    %d - %s" % (t["running"], t["task"])
 
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -13,6 +13,7 @@ import logging
 import threading
 import os.path
 import cPickle
+from operator import itemgetter
 
 from WMCore.DAOFactory        import DAOFactory
 from WMCore.WMExceptions      import WM_JOB_ERROR_CODES
@@ -24,22 +25,6 @@ from WMCore.DataStructs.JobPackage            import JobPackage
 from WMCore.FwkJobReport.Report               import Report
 from WMCore.WMException                       import WMException
 from WMCore.BossAir.BossAirAPI                import BossAirAPI
-
-
-def siteListCompare(a, b):
-    """
-    _siteListCompare_
-
-    Sites are stored as a tuple where the first element is the SE name and the
-    second element is the number of free slots.  We'll sort based on the second
-    element.
-    """
-    if a[1] > b[1]:
-        return 1
-    elif a[1] == b[1]:
-        return 0
-
-    return -1
 
 
 class JobSubmitterPollerException(WMException):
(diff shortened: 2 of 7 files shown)


bin/wmagent-resource-control
```diff
@@ -167,13 +167,12 @@ def printThresholds(resourceControl, desiredSiteName):
         print "%s - %d running, %d pending, %d running slots total, %d pending slots total%s:" % (
                         siteName, runningJobs, pendingJobs, runningSlots, pendingSlots, stateMsg)
 
-        for task in siteThresholds['thresholds']:
+        for task, thres in siteThresholds['thresholds'].iteritems():
```
hufnagel: iteritems is deprecated, please use items

hufnagel:
"Useless and over-complicated" ? Could you explain what you removed ? Will want to review this patch, but some explanation of what features you removed would be good...

bin/wmagent-resource-control
amaltaro: Interesting. I'll update this patch tomorrow. Still, depending on the length of the dictionary, it may be useful to keep iteritems() until the very last moment towards the python 3 migration

amaltaro:
Don't spend time reviewing it right now Dirk, there are all those unit tests that I have to go through before.

This was more a heads up such that Seangchan does not work on the same thing. I'll give you all the details tomorrow, my quota of work in the weekend is way exceeded now :)

bin/wmagent-resource-control
ericvaandering: I don't think so. In python 3 items() is the same as what iteritems() is in python 2. In python 2 both exist and items() actually builds the list. Use iteritems() for now, the python 2to3 converter will make it items later. (It will translate ```items()``` to ```[items()]```)

hufnagel: Thing is, we (almost) never use the items or values or keys functions to access the list directly, we always use it for iteration.

I have started to use items() instead of iteritems() exclusively since it is python3 compatible syntax and frankly, for the list sizes I am dealing with building the list in memory doesn't really matter.

As such, converting all instances of items() to [items()] is absolutely not what I want. My guess would be that 100% of items() calls can stay items() calls.

ericvaandering: But you are missing the point. If you use items() now, then the actual 2to3 converter which we will have to run is going to make them into [items()] and then we'll have to go through and remove them. If you use iteritems() now, you will be *doing* what python3 does with items() and the conversion will generate the right code.

ericvaandering: https://docs.python.org/2/library/2to3.html#2to3fixer-dict

hufnagel: I get your point :-). My point is that using items() is the more "pythonic" way to write this (hence the change in python3). This has nothing to do with functionality, this has to do with what looks natural in code. And "for a,b in items()" look natural, so natural that the majority of our existing code is using it for iterations. Blindly converting it to "for a,b in [items()]" is wrong.

Before we run the converter we should have a campaign to switch all items() call that we can (as I said, I expect 100%) to iteritems() and only then run the converter.

hufnagel: Anyways, my original objection is wthdrawn. If we are doing a global converted run anyways during he switch from python2 to python3, using iteritems() now is indeed functionally correct.

amaltaro: Ok, I took my change to items() out then

ericvaandering: Right, we have to do a global converter run, so we should program for the functionality we want and look forward to the glorious future where the right functionality also looks the prettiest. :-)

amaltaro:
Now all unit tests should be fixed.

Summary of changes:
* changed the output format of listThresholdsForSubmit()
* got rid of the alert system and all those caching for workflows, prios, sorted sites, etc
* cache refresh mechanism was not changed
* populate the cachedJobs following *only* the final job priority (which takes request and task priority into account)
* no more check for job type, what matters is a) absolute job priority and b) their creation - in wmbs - timestamp
* merged the long tuple and job dict into a single job dict. A double check here is welcome :-)
* no longer check running thresholds at job submitter. Let us care only about the pending thresholds. It means, even if the site is full of running jobs but still has pending slots available, then submit jobs to it.

@hufnagel @ticoann feel free to review it. Heavy testing is needed for it yet.

amaltaro:
These WMBS failures don't come from this PR (I've just fixed them in 7005)

amaltaro:
Just fixed some nasty key mistakes in `ListThresholdsForSubmit`. Now the component is no longer crashing :)

amaltaro:
I was wondering this morning... and I'm not really sure whether it's a good idea not to check the running thresholds here. The only problem I can see is that an agent will always keep jobs pending, no matter how many jobs a site is running (which, from my initial point of view, it was good). For instance:
```
Site ABC running_slots 1000, pending_slots 500
0 running jobs, 500 pending jobs
500 running jobs, 0 pending jobs
500 running jobs, 500 pending jobs
1000 running jobs, 0 pending jobs
1000 running jobs, 500 pending jobs
etc etc etc
```

if the fair share imposed by HTCondor/gWMS works reasonably Ok, then we should not worry about running thresholds at JobSubmitter level. Otherwise, we better have that back to limit total number of jobs submitted for a site.

@ericvaandering @ticoann @hufnagel comments?

hufnagel:
I think there are different considerations here. For processing and production jobs running limits are not very useful. A site should control how much resources they are willing to give us.

For merge, cleanup, logcollect, ie. jobs that put a relatively high load on the io systems, the situation is different. Normally you keep up with those, but if for some reason you fall behind, you don't want to start flooding a site with only such jobs. So running limits are very much needed.

Which brings me to conclude that we should maintain running limits overall and per task type in WMCore and pass the bucket to Ops. We can simply eliminate running limits in general and for production and processing jobs by setting the running limits for a site to 100000, which at the same time keeping the running limits for other task type at something small. This is an operational decision that can be implemented without us making sweeping changes in the code.

amaltaro:
Thanks, Dirk! I made the changes you suggested in my 3rd commit, though I didn't add any special handling for specific job types. I'd avoid that as much as I can, if possible. Moreover, Ops can always play with thhresholds settings and make them unlimited, as you mentioned before.

I'll test it at larger scale in the next days, meanwhile, more comments and review are welcome :-)

ericvaandering:
Just a comment on this from @hufnagel :

"For merge, cleanup, logcollect, ie. jobs that put a relatively high load on the io systems, the situation is different. Normally you keep up with those, but if for some reason you fall behind, you don't want to start flooding a site with only such jobs. So running limits are very much needed."

We should be working towards pushing these limits off onto GlideinWMS in the Global Pool so that these limits can be enforced globally, not on a per agent basis.

amaltaro:
@ticoann good to go as well. If there are remaining problems, it will be only found at large scale :)

PR #6365, opened 2015-10-28
cleaning up after PNN patch

alexanderrichards:
This pull includes change to the wmbs_location_senames table as well as change to the column name se_name.

This pull fixes https://github.com/dmwm/WMCore/issues/6066

PR #7005, opened 2016-07-26
Fix a some WMBSHelper unit tests - SE to PNN transition

amaltaro:
Fix at least 2 tests that fail since we merged this PR #6365

@ericvaandering @jha2 please review. I'm not sure this is the correct way...


amaltaro:
test this please

ericvaandering:
Which unit tests? You're actually editing the old SiteDB emulator to get these to pass, which is OK, but we are trying to get rid of that one.

amaltaro:
It was meant to fix these 2
```
WMCore_t.WorkQueue_t.WMBSHelper_t.WMBSHelperTest:testDuplicateSubscription changed from error to success
WMCore_t.WorkQueue_t.WMBSHelper_t.WMBSHelperTest:testMCFakeFileInjection changed from error to success
```
but it seems I got several others for free... beers are accepted ;-)

ticoann:
Thanks Alan, I thing we should merge this one first even though eventually will change to use sitedb mock. What do you think?

ericvaandering:
Yes, I agree

amaltaro:
Thanks!

jha2:
Alan, this is concerning the pull request #7017 . Here, I also added the meta data for fake sites like 'T2_XX_SiteB'.  The meta data are different for the pull request #7005 and #7017.  The type of meta data depends on the service (or call)  which we want to test.

PR #7009, opened 2016-07-28
Make sure file:// is added before source/dest pfn in gfal2 plugin

juztas:
Created a new function createFinalPFN, which will take care of preparing pfn and appending file:// if needed 
a) Also added a check if pfn starts with '/' and in that case append file:// any function createStageOutCommand or createRemoveCommand or removeFile uses new function for preparing pfn. 
In any case if WMA or CRAB or whatever would call this Stageout plugin without correct pfns, it enforces pfn modification by calling createFinalPFN function.
Some details here: https://github.com/dmwm/WMCore/pull/6980


juztas:
@amaltaro please review

amaltaro:
Thanks, Justas.
I'm not going to test it again, so @ticoann please merge

PR #7013, opened 2016-07-29
Protect WorkQueueManager thread against DBS/PhEDEx faulty calls

amaltaro:
There was a StoreResults request that was breaking vocms0304 when calling this method `_getDBSBlock()`. Since it is not enclosed in a try/except block, the whole thread was failing and no work was getting into wmbs.

Taking advantage, I also set some defaults to the logDB instantiation.
@ticoann please review, I still have to test it

### DIFF
--- a/src/python/WMCore/Services/LogDB/LogDB.py
+++ b/src/python/WMCore/Services/LogDB/LogDB.py
@@ -5,7 +5,6 @@ https://github.com/dmwm/WMCore/issues/5705
 """
 
 # standard modules
-import os
 import re
 import logging
 import threading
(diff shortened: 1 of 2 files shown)


ericvaandering:
We had a StoreResults request!?!?!

amaltaro:
Yep! Paola was struggling the last weeks to get it working (there are some ggus tickets waiting for dataset elevation for months now). Do you know whether we still (should) support it?

ericvaandering:
We should still support it.

amaltaro:
@ticoann sorry Seangchan, I forgot this PR. Are you able to patch your services/agent and let a couple of workflows go into acquired? It would be good to have it in testbed.

Otherwise each time data handling fails in WorkQueueManagerWMBSFileFeeder, we lose the whole cycle (not sure it affects central workqueue though).

ticoann:
Thanks Alan, I will test this and patch the production agent as well. Stop working and enjoy your vacation!

amaltaro:
Stopping at this exact minute ;-) Thanks!

PR #7014, opened 2016-07-29
Send multiple job report to dashboard with the same object

amaltaro:
I found out some agents spending several minutes to send information to dashboard. One of them was actually even failing to resolve the name into an ip address from time to time (since we get a new instance for each report).

This patch solves that problem by passing a list of jobs to report to the same method that retrieves the apmon instance. I also tried to get rid of some of those standalone functions and make it all in the `DashboardAPI`  class (so I had to update the WMRuntime report as well).

I still want to a) make it report a group of jobs in `handleSteps` instead of one by one, b) run a larger test to make sure it behaves Ok and c) test the ReqMgr report

@ticoann @ericvaandering please review
@mmascher @juztas @emaszs I noticed there is a 'copy' of the apmon code in the CRABServer repo. Is CRAB completing relying on the WMCore code for the dashboard report. There are several different functions/reports here that I don't really know if they are used at all (certainly not by WMAgent)

### DIFF
--- a/src/python/WMCore/Services/Dashboard/DashboardAPI.py
+++ b/src/python/WMCore/Services/Dashboard/DashboardAPI.py
@@ -32,19 +32,27 @@ class DashboardAPI(object):
         self.defaultParams = {'sys_monitoring': 0, 'general_info': 0, 'job_monitoring': 0}
         self.server = server if server else DASHBOARDURL
         self.logger = logr if logr else logging.getLogger()
+        self.apmon = None
 
-    def getApMonInstance(self):
+    def __enter__(self):
+        self.apmon = self._getApMonInstance()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.apmon.free()
+
+    def _getApMonInstance(self):
         """
         _getApMonInstance_
 
-        Return an instance of ApMon
+        Private method that returns an instance of ApMon
         """
         self.logger.debug("Creating ApMon with static configuration")
         apMonConf = {self.server: self.defaultParams}
         try:
             return apmon.ApMon(apMonConf, self.logger)
         except Exception as ex:
-            msg = "Errror while getting an instance of ApMon under: %s\n" % self.server
+            msg = "Error while getting an instance of ApMon under: %s\n" % self.server
             msg += str(ex)
             self.logger.error(msg)
         return None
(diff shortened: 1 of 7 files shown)


test/python/WMCore_t/Services_t/Dashboard_t/DashboardAPI_t.py
```diff
@@ -16,10 +16,8 @@
 
 from WMCore.WMBase import getTestBase
 from WMCore.Services.Dashboard.Logger import Logger
-from WMCore.Services.Dashboard.DashboardAPI import DashboardAPI
-from WMCore.Services.Dashboard.DashboardAPI import apmonFree, apmonSend, parseAd
-from WMCore.Services.Dashboard.DashboardAPI import APMONINSTANCE, APMONINIT, APMONCONF
-from WMCore.Services.Dashboard.DashboardAPI import reportFailureToDashboard, logger, getApmonInstance
+from WMCore.Services.Dashboard.DashboardAPI import DashboardAPI, parseAd, \
+    reportFailureToDashboard, logger
```
ericvaandering: I prefer ```from blah import (foo, bar)``` to line continuations.

test/python/WMCore_t/Services_t/Dashboard_t/DashboardAPI_t.py
```diff
@@ -34,22 +32,23 @@ def testApmonInstance(self):
 
         Just test initialization of apmon Instance
         """
-        print("Apmon Configuration %s" % APMONCONF)
-        apmon = getApmonInstance(apmonServer=APMONCONF)
-        self.assertTrue(apmon.initializedOK())
+        dashboard = DashboardAPI()
```
ericvaandering: This pattern looks an awful lot like something that could be done as a context manager. Maybe we should think about adding those methods to DashboardAPI

amaltaro:
@ericvaandering please have a look at the third commit, it should have fixed both of your comments.
I took me some reading on context managers, hopefully I got it right and didn't overcoded :)

I still need to test it in real life.

test/python/WMCore_t/Services_t/Dashboard_t/DashboardAPI_t.py
```diff
@@ -32,12 +31,9 @@ def testApmonInstance(self):
...
         """
-        dashboard = DashboardAPI()
-        apmonInst = dashboard.getApMonInstance()
-        self.assertTrue(apmonInst.initializedOK())
-        # Free up apmon instance and check if it was successfull
-        apmonInst.free()
-        self.assertEqual(None, apmonInst)
+        with DashboardAPI() as dashboard:
+            self.assertTrue(dashboard.apmon.initializedOK())
+        self.assertTrue(dashboard.apmon.freedOK())
```
ericvaandering: Does this make sense? I would guess "dashboard" is only valid within the context. But maybe you can do this.

src/python/WMCore/Services/Dashboard/DashboardAPI.py
```diff
@@ -32,19 +32,26 @@ def __init__(self, monitorId=None, jobMonitorId=None, lookupUrl=None,
         self.defaultParams = {'sys_monitoring': 0, 'general_info': 0, 'job_monitoring': 0}
         self.server = server if server else DASHBOARDURL
         self.logger = logr if logr else logging.getLogger()
+        self.apmon = self._getApMonInstance()
 
-    def getApMonInstance(self):
+    def __enter__(self):
+        return self
```
ericvaandering: I'm having a hard time following this, but __enter__ should create the resource. Or is it still created in __init__?

test/python/WMCore_t/Services_t/Dashboard_t/DashboardAPI_t.py
amaltaro: Good catch! It worked in my local unittest, but I was probably just lucky. Removing the last assert in a bit

src/python/WMCore/Services/Dashboard/DashboardAPI.py
amaltaro: The real resource is created in the init method (self.apmon). It could be created under enter method, but the pylint would not be happy with an instance attribute being created outside of the init method.

amaltaro:
Ok, I'm done with my changes and I'd like to have it in the next tag for the next production candidate release.
Hopefully I ran all the tests this time, including performance wise. I also x-checked this info with the Dashboard team.

I kept the 3rd commit alone because it has the context manager usage and it touches both srv and test code, so squashing is going to be tricky.

@ticoann please merge if @ericvaandering has no additional comments

src/python/WMCore/Services/Dashboard/DashboardAPI.py
ericvaandering: Not to say that you should change it, but the way around that is self.apmon = None in the init()

amaltaro:
Thanks, Eric. I updated the enter() method to make it more clear.

PR #7016, opened 2016-08-01
Store site lists in the proper list format

amaltaro:
It seems ReqMgr2 accepts (and stores) site lists in a string format (as provided), which then cause issues later on while casting it to a list. Elog:
https://cms-logbook.cern.ch/elog/Workflow+processing/25185

```
>>> a = "T1_DE_KIT"
>>> for site in a:
...   site
... 
'T'
'1'
'_'
'D'
'E'
'_'
'K'
'I'
'T'
>>> 
```

@ticoann please review. I just injected a test workflow and it was properly cast to a list in the workload doc
I'll make sure Andrew L. has this fix on his assignment script until we get it into production


amaltaro:
@ticoann this is the second PR I'd like to get in for the current testbed (the other is 7013)

PR #7020, opened 2016-08-02
add robust merge optional

ticoann:
@hufnagel, Dirk please review, Tier0 workflow spec needs to be updated to make robust merge false.


ticoann:
this need to be tested

PR #7028, opened 2016-08-05
use dbs client to check the dbs status instead of REST call

ticoann:
@ericvaandering, Eric please review. reqmgr2 spec is already included to use dbs3 client

### DIFF
--- a/src/python/WMCore/WMSpec/WMWorkloadTools.py
+++ b/src/python/WMCore/WMSpec/WMWorkloadTools.py
@@ -62,32 +62,6 @@ def strToBool(string):
         raise WMSpecFactoryException("Can't convert to bool: %s" % string)
 
 
-def _verifyDBSCall(dbsURL, uri):
-    try:
-        # from WMCore.Services.DBS.DBS3Reader import DBS3Reader
-        # DBS3Reader(dbsURL).dbs.serverinfo()
-        from WMCore.Services.Requests import JSONRequests
-        jsonSender = JSONRequests(dbsURL)
-        result = jsonSender.get("/%s" % uri)
-        if not result[1] == 200:
-            raise WMSpecFactoryException("DBS is not connected: %s : %s" % (dbsURL, str(result)))
-    except:
-        raise WMSpecFactoryException("DBS is not responding: %s" % dbsURL)
-
-    return result[0]
-
-
-def checkDBSURL(dbsURL):
-    # use the import statement here since this is packed and used in RunTime code.
-    # dbs client is not shipped with it.
-
-    if dbsURL:
-        _verifyDBSCall(dbsURL, "serverinfo")
-        return True
-
-    return True
-
-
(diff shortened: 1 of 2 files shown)


ericvaandering:
When I said "take a look at it" I didn't mean "don't go to lunch until you've fixed it"!

src/python/WMCore/WMSpec/WMWorkloadTools.py
```diff
@@ -64,13 +64,8 @@ def strToBool(string):
...
 def _verifyDBSCall(dbsURL, uri):
     try:
-        # from WMCore.Services.DBS.DBS3Reader import DBS3Reader
-        # DBS3Reader(dbsURL).dbs.serverinfo()
-        from WMCore.Services.Requests import JSONRequests
-        jsonSender = JSONRequests(dbsURL)
-        result = jsonSender.get("/%s" % uri)
-        if not result[1] == 200:
-            raise WMSpecFactoryException("DBS is not connected: %s : %s" % (dbsURL, str(result)))
+        from WMCore.Services.DBS.DBS3Reader import DBS3Reader
```
ericvaandering: Two comments. First, is there a reason to have the import here instead of in the headers? Second, the function actually makes a DBS call but I guess does not return anything? So the purpose was just to make sure that DBS was online?

What you've written now will pass with the mock emulator no problem which should solve @alexanderrichards is.

ericvaandering:
Check line 154 and line 170. There real calls to DBS are being executed. Why not just convert all three usages of _verifyDBSCall to real calls to DBSReader3?

ericvaandering:
Well, that went *boom*. :-)

ticoann:
The import is inside the function call due to the problem the application which import this module doesn't have dependency on dbs client.  I am not sure which one was that. It might be Tier0 WMStats. Let me check more and I will update this patch to change all the dbs call uses  dbs api.

ericvaandering:
nose.failure.Failure:runTest was added. Status is error and the deleted tests are a classic sign of a busted import

ticoann:
Eric, why does it say many unittest are deleted?
I am also wondering whether we need to add the mock api for listDatasets and listDatasetParents.

ericvaandering:
That's what I was pointing out on my comment. There is an import that is broken, I bet. You can also follow the link and find out WHAT tests are broken and what the error message is. You should get used to doing that.

ticoann:
Hi Eric, I was just wondering why the error message shows the test is deleted. Sorry, I didn't know it is related to a broken import or something else. I updated the patch with the broken import.

ericvaandering:
Ah, it shows that because those tests all get reported as general nose failures. I think because it can't even get far enough to do the discovery of which "test*" exist. Or something..

ericvaandering:
Ok, these look like deficiencies in the mock emulator now. @ticoann I will try to take it from here.

ticoann:
replaced by #7030

PR #7030, opened 2016-08-08
Building on Seangchan's changes to remove direct calls to DBS

ericvaandering:


ericvaandering:
This should supersede #7028 assuming it works

ericvaandering:
Let's see how these go. Some of these look like new failures against the actual DBS exposed by Seangchan's changes. The mock-related ones may be all fixed now. If that's the case, I'd suggest we move this in and I can work to get more tests using the mock emulator.

ericvaandering:
@ticoann the four failures in Dataset_t, Block_t, and WMCore_t.WorkQueue_t.WorkQueue_t.WorkQueueTest.testProcessingWithPileup are real. These are nonexistent data sets that are throwing errors from your new check where they had not been thrown before. I can see two possible options: move the behavior back to what it was where these checks just checked for a valid regex or fix the test to use real dataset names.

For instance, here is a report: https://cmssdt.cern.ch/jenkins/job/DMWM-WMCore-PR-test/2721/testReport/WMCore_t.WorkQueue_t.WorkQueue_t/WorkQueueTest/testProcessingWithPileup/

ticoann:
replaced by #7036

PR #7036, opened 2016-08-09
Dbs wq fix

ticoann:
replaces #7030


ticoann:
It introduces other unittest failure in WorkQueueSimpleTest. Will check later.

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
```diff
@@ -96,6 +96,8 @@ def _queryDbsAndGetPileupConfig(self, stepHelper, dbsReader, fakeSites):
                 blockReplicasInfo = phedex.getReplicaPhEDExNodesForBlocks(dataset=dataset, complete='y')
                 for block in blockReplicasInfo:
                     nodes = set(blockReplicasInfo[block]) - node_filter | set(fakePNNs)
+                    print("DDDD %s" % blockDict.keys())
+                    print(blockReplicasInfo.keys())
```
amaltaro: Don't forget removing these lines before merging it :-D

src/python/WMCore/WMSpec/WMWorkloadTools.py
```diff
@@ -153,13 +127,18 @@ def _validateArgumentOptions(arguments, argumentDefinition, optionKey):
...
     dbsURL = arguments.get("DbsUrl", None)
     if inputdataset != None and dbsURL != None:
-        result = _verifyDBSCall(dbsURL, "datasets?&dataset_access_type=*&dataset=%s" % inputdataset)
-        if len(result) == 0:
-            msg = "Inputdataset %s doesn't exist on %s" % (inputdataset, dbsURL)
-            raise WMSpecFactoryException(msg)
+        #import DBS3Reader here, since Runtime code import this module and worker node doesn't have dbs3 client 
+        from WMCore.Services.DBS.DBS3Reader import DBS3Reader
+        from WMCore.Services.DBS.DBSErrors import DBSReaderError
+        try:
```
amaltaro: I don't follow it. Why we want to contact DBS from the worker node? Or are you referring to the jenkins node that does not have dbs client?
Sorry if I missed something, but can't we mock these checks instead of adding these imports in the middle of the code?

ericvaandering: We don't want to access it from the worker node. There are methods from this file that are being called on the worker node, so putting the imports at the top would not work. The methods that need to contact DBS are in ReqMgr(?).

amaltaro: I'm not saying we should, but how about we then create a method in this file that interacts with DBS server?
We can have the specific import in a single place and handle the api to be called from there.

src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
ticoann: Thanks I will remove it.

src/python/WMCore/WMSpec/WMWorkloadTools.py
ticoann: I was planning to move the functions in different module and not include the package packed but this module is not directly accessed in the WN but through WMSpec. So I imported when DBSService with in the module. I could create the function for import and call that function with in other function but one method need 2 import and the other need just one.  I know it is ugly but couldn't improve much with the simplest change.

ticoann:
@ericvaandering, I tried to cherry-pick your commit from #7034 over this patch, but there a conflict on test/data/Mock/DBSMockData.json although this patch doesn't contain any changes in DBSMockData.json.
And when I just overwrites test/data/Mock/DBSMockData.json with files in your patch it gives me new set of errors.
I was just wondering how do you want to proceed.

ericvaandering:
Sorry, hadn't seen your not when I went to talk to you. I'm taking care of this. Just to tell you what I'm doing: 
1) Cherry pick everything but the JSON changes
2) Re-run MakeDBSMock to recreate the JSON
3) Check in new JSON

ticoann:
I though you already saw this when we talked. :smile: Thanks, Eric.

PR #7037, opened 2016-08-10
Get proper data location for StoreResults requests

amaltaro:
Do not use PhEDEx for data location of StoreResults requests, since the input dataset is not available in TMDB. This change should be transparent to anything else.

Paola helped me to test it in testbed and it seems to be working fine.
However, I made a last minute change and I still want to check whether it's Ok.

@ticoann please review

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -624,9 +624,12 @@ class DBS3Reader:
 
         return locations
 
-    def getFileBlock(self, fileBlockName):
+    def getFileBlock(self, fileBlockName, dbsOnly=False):
         """
         _getFileBlock_
+
+        dbsOnly flag is mostly meant for StoreResults, since there is no
+        data in TMDB.
 
         return a dictionary:
         { blockName: {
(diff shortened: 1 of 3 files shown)


amaltaro:
Just fixed another issue for cases that do not use global DBS for data location mapper.
It was casting a string to set, so again a `set(['C', 'E', 'H', 'N', '2', 'T', 'R', '_'])` like.

ticoann:
Thanks Alan, patch looked good to me. Are you still testing or can I merge?

ticoann:
Is this urgent? Do we need production cmsweb upgrade?

amaltaro:
I don't think it's need for cmsweb because we're running tests in testbed and in principle it's all working fine. Though I did not check the workqueue logs to see whether it "breaks" with those site white list.

Changes look isolated enough that we could merge them today for the next tag. Anyways I'm running the new agent validation with this patch

amaltaro:
@ticoann please merge as well

vlimant:
@ticoann @amaltaro is this going to be in the next deployment ? it seems to me that we need this in order to process the current StoreResult requests in the system.

amaltaro:
It's in the new agent release. If I'm not wrong Paola mentioned that she assigned these workflows to the `production` team, that was running a mix of new and old WMAgent release.
If so, abort/clone and assign it again.

vlimant:
Is that now deployed on all agents ?

amaltaro:
Yep!

PR #7038, opened 2016-08-10
Dbs wq plus mock2

ericvaandering:
My version of #7036


ticoann:
It seems good enough to merge. There are 2 errors and 1 failure in Block_t. But Eric promised to take care of that. Thanks 😄

PR #7039, opened 2016-08-10
add checking for wmarchiver upload status

ticoann:
@amaltaro, Alan please review. (I will include in the tag for the test but please review)


ticoann:
Include name change for WMArchiver -> WMArchive to match the application name

PR #7045, opened 2016-08-12
1.0.17 version of Make sure file:// is added before source/dest pfn i…

amaltaro:
…n gfal2 plugin

1.0.17 branch version of #7009 
It's required because there are more sites joining the gfal stage out plugin and T2_CH_CSCS is one example site that just got back into production and requires this patch to work.


amaltaro:
Pylint is happy, so I'm pushing it to all the production agents.

ticoann:
Thanks Alan

PR #7046, opened 2016-08-12
Robust comparison and use for DBS

ericvaandering:
I wrote a new kind of unit test assert() to check lists of dictionaries of lists (all the way down) or dictionaries as a starting point. The difference is all the lists are sorted at each point.

This is better for dealing with DBS data since depending on the day of the week, the list of lumis, etc, come out in random order. So it should fix this intermittent error with the test of the DBS mock emulator.


ericvaandering:
@ticoann: please merge

ticoann:
Thanks Eric,

PR #7048, opened 2016-08-12
use the adler32 checksum in gfal2 stageout

hufnagel:
Our runtime code calculate the adler32 checksum before calling the stageout code. Actually use the checksum in the gfal2 stageout (like we already do for xrdcp stageout).

I have no way to test this myself, syntax rules are based on the xrdcp stageout, which supports the same argument in the same fashion.

@amaltaro , please review and test as convenient

### DIFF
--- a/src/python/WMCore/Storage/Backends/GFAL2Impl.py
+++ b/src/python/WMCore/Storage/Backends/GFAL2Impl.py
@@ -103,6 +103,8 @@ class GFAL2Impl(StageOutImpl):
 
         copyCommandDict = {'checksum': '', 'options': '', 'source': '', 'destination': ''}
 
+        useChecksum = (checksums is not None and 'adler32' in checksums and not self.stageIn)
+
         if not options:
             options = ''
 
@@ -111,7 +113,11 @@ class GFAL2Impl(StageOutImpl):
         args, unknown = parser.parse_known_args(options.split())
 
         if not args.nochecksum:
-            copyCommandDict['checksum'] = "-K adler32"
+            if useChecksum:
+                checksums['adler32'] = "%08x" % int(checksums['adler32'], 16)
+                copyCommandDict['checksum'] = "-K adler32:%s" % checksums['adler32']
+            else:
+                copyCommandDict['checksum'] = "-K adler32"
 
         copyCommandDict['options'] = ' '.join(unknown)
 


amaltaro:
Dirk, what's the difference between L118 and L120? Both use checksum adler32 but only the former does the checksum validation after the transfer? If so, can we simply remove that `-k adler32` in L120?

BTW, is it failing at some site(s) or it's "just" to have this additional check on files?

hufnagel:
The difference between L118 and L120 is how the reference adler32 checksum is determined. In one case it calculates at source, calculates at destination and compares. The other has a reference specified, which it compares to the destination.

Since we calculate the adler32 checksum in WMAgent runtime code before calling stageout, L118 uses that adler32 checksum. It gives you a tiny bit better protection (since the checksum is calculated earlier) and it should in principle reduce the cpu overheads for the stageout (but that effect is likely somewhat negligible).

L120 is for stageIn, where we do not know the adler32 checksum of the source. It would also trigger if there is no checksum passed to the stageout code (which should never happen).

hufnagel:
@amaltaro , good enough feedback ? Is this going to be included in one of the next tests ? Should see some testing at a variety of sites using gfal2 stageout if possible. No real hurry to merge it, would prefer this to be well tested.

ticoann:
@amaltaro, Alan how did we test this before I remember Justas was testing. 
If Dirk is confident, we can merge it to include in the new tag and run the test, what do you think?

amaltaro:
We better leave it out for now, I didn't test it yet. I know there are more sites out there, but those certainly using gfal are Purdue and CSCS.

It seems production doesn't urgently need it, so I suggest we leave it until me or someone else can have a look at it in real life.

hufnagel:
I agree there is no real urgency to merge this and we should test this before. OTOH, I also don't want it to be forgotten and then we make other changes to the gfal2 plugin in the mean time and have to rebase etc.

src/python/WMCore/Storage/Backends/GFAL2Impl.py
```diff
@@ -103,6 +103,8 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
 
         copyCommandDict = {'checksum': '', 'options': '', 'source': '', 'destination': ''}
 
+        useChecksum = (checksums != None and 'adler32' in checksums and not self.stageIn)
```
amaltaro: Use `is` or `is not` for comparison with `None` please.

amaltaro:
Adding this patch to vocms0230 right now, such that we can exercise these changes.

amaltaro:
Dirk, are you aware of any site that has this option already enabled? Otherwise I'll just send a couple hundred jobs to Purdue and CSCS

hufnagel:
The only site I know about that needs the optons is cms@home stageout to the data bridge and we are already using this patch there.

Your tests should just confirm that the patch works for sites that don't use the new option. The stageout command and xrdcp parameters also changed for these sites.

amaltaro:
and I had no issues in a couple of workflows, so it's good to get out of my todo list :-D
@ticoann it affects only agent side, feel free to merge it

ticoann:
Thanks Alan

PR #7049, opened 2016-08-14
Convert leftover se_name to pnn

amaltaro:
John found (thanks!) this problem in a replay, the mysql class was updated but not the oracle one.


ticoann:
Thanks John, Alan

PR #7051, opened 2016-08-15
specify the timeout error

### LINKED ISSUE #6830: Differentiate error code for time out.
There are 3 cases WMAgent kills the job when timeout occurs. current code is 71304
Andrew, pointed that it would  be helpful for debugging to have different error code for each cases (Error, Running, Pending)
Is there any rules for setting the error code.
I could set 
71304 - Running
71305 - Pending
71306 - Error

@amaltaro, @ericvaandering, any comments?

ticoann:
fixes #6830

### DIFF
--- a/src/python/WMCore/BossAir/StatusPoller.py
+++ b/src/python/WMCore/BossAir/StatusPoller.py
@@ -11,7 +11,9 @@ JobStatusAir
 import time
 import logging
 import threading
+from collections import defaultdict
 
+from Utils.IterTools import flattenList
 from WMCore.WMException                    import WMException
 from WMCore.WMExceptions                   import WM_JOB_ERROR_CODES
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
@@ -101,7 +103,7 @@ class StatusPoller(BaseWorkerThread):
             return
 
         # Look for jobs that need to be killed
-        jobsToKill = []
+        jobsToKill = defaultdict(list)
 
         # Now check for timeouts
         for job in runningJobs:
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, @AndrewLevin, Alan, Andrew please review. (still need to be tested)

src/python/WMCore/BossAir/StatusPoller.py
```diff
@@ -116,14 +118,18 @@ def checkStatus(self):
...
                     logging.info("Killing job %i because it has exceeded timeout for status '%s'", job['id'], globalState)
                     job['status'] = 'Timeout'
-                    jobsToKill.append(job)
-
+                    jobsToKill[globalState].append(job)
+        
+        timeOutCodeMap = {"Running": 71304, "Pending": 71305, "Error": 71306}            
         # We need to show that the jobs are in state timeout
         # and then kill them.
+        jobsToKillList = itertools.chain.from_iterable(jobsToKill.values())
```
amaltaro: This would break in BossAirAPI because it's an iterable and you cannot get its length. Perhaps you could use directly this util https://github.com/dmwm/WMCore/blob/master/src/python/Utils/IterTools.py#L22 (or just cast it here to list :))

ticoann: Thanks Alan, I forgot that Eric wrote this function already. Code updated

ericvaandering: @amaltaro wrote it, not me.

amaltaro:
Seangchan, did you manage to test this patch or I better have a look at it before it goes live?

PR #7057, opened 2016-08-15
Switch PhEDEx injector tests to mock for SiteDB and DBS

ericvaandering:
Don't merge yet


ericvaandering:
Ok @ticoann this can be merged too, it looks like. Aside from the big one, which is WorkQueue, this completes the SiteDB portion of this. @alexanderrichards is working on WorkQueue right?

ticoann:
Yes Alex is working on WorkQueue. Thanks,

PR #7064, opened 2016-08-17
Fix resource control tests

ericvaandering:
The real change is that we no longer have a constant # of sites in ResourceControl since we're using the mock emulator. (assertGreaterEqual).  Rest of the changes are just cosmetic.


ericvaandering:
Looks good to me, @ticoann

PR #7076, opened 2016-08-20
Add only unique file/location rows into dbsbuffer

amaltaro:
Sigh... why Oracle cannot have the handy and simple `ignore` clause...

This fix is supposed to fix this problem [1]
@hufnagel @yuyiguo can you review it please?
@ticoann I'll apply this patch to the production nodes as soon as I get it reviewed (vocms0309 is complaining about it)

```
[1]
(IntegrityError) ORA-00001: unique constraint (CMS_WMBS_PROD2.PK_DBS_FIL_LOC) violated
 'INSERT INTO dbsbuffer_file_location (filename, location)\n               SELECT df.id, dl.id\n               FROM dbsbuffer_file df, dbsbuffer_location dl\n               WHERE df.lfn = :lfn\n               AND dl.se_name = :sename' [{'sename': 'T3_US_Vanderbilt_EC2', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/492/00000/C0B873F9-5393-E511-A0ED-02163E014704.root'}, {'sename': 'T2_US_Vanderbilt', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/492/00000/C0B873F9-5393-E511-A0ED-02163E014704.root'}, {'sename': 'T3_US_Vanderbilt_EC2', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/494/00000/FA899139-5493-E511-899F-02163E01433B.root'}, {'sename': 'T2_US_Vanderbilt', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/494/00000/FA899139-5493-E511-899F-02163E01433B.root'}, {'sename': 'T3_US_Vanderbilt_EC2', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/495/00000/4A753A00-5493-E511-8EEC-02163E011B5A.root'}, {'sename': 'T2_US_Vanderbilt', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/495/00000/4A753A00-5493-E511-8EEC-02163E011B5A.root'}, {'sename': 'T3_US_Vanderbilt_EC2', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/496/00000/C820070A-5493-E511-962B-02163E0134BD.root'}, {'sename': 'T2_US_Vanderbilt', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/496/00000/C820070A-5493-E511-962B-02163E0134BD.root'}  ... displaying 10 of 500 total bound parameter sets ...  {'sename': 'T3_US_Vanderbilt_EC2', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/768/00000/8C431DDE-9995-E511-87C9-02163E014130.root'}, {'sename': 'T2_US_Vanderbilt', 'lfn': '/store/hidata/HIRun2015/HIHardProbes/RAW/v1/000/262/768/00000/8C431DDE-9995-E511-87C9-02163E014130.root'}]
```

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetLocation.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetLocation.py
@@ -11,7 +11,7 @@ MySQL implementation of DBSBuffer.SetLocation
 from WMCore.Database.DBFormatter import DBFormatter
 
 class SetLocation(DBFormatter):
-    sql = """INSERT INTO dbsbuffer_file_location (filename, location)
+    sql = """INSERT IGNORE INTO dbsbuffer_file_location (filename, location)
                VALUES (:fileid, :locationid)"""
 
     def execute(self, binds, conn = None, transaction = None):
--- a/src/python/WMComponent/DBS3Buffer/Oracle/DBSBufferFiles/SetLocation.py
+++ b/src/python/WMComponent/DBS3Buffer/Oracle/DBSBufferFiles/SetLocation.py
@@ -12,4 +12,11 @@ from WMComponent.DBS3Buffer.MySQL.DBSBufferFiles.SetLocation import SetLocation 
      MySQLSetLocation
 
 class SetLocation(MySQLSetLocation):
-    pass
+    """
+    Insert unique lfn, location row
+    """
+    sql = """INSERT INTO dbsbuffer_file_location (filename, location)
+               SELECT :fileid, :locationid FROM DUAL WHERE NOT EXISTS
+                 (SELECT filename FROM dbsbuffer_file_location
+                 WHERE filename= :fileid and location= :locationid)
+               """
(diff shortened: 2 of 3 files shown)


yuyiguo:
This was the correct DB behavior.  You may catch the exception and ignore it. But  looks like you were inserting more than one rows. You would have to figure out which one was the duplicated and remove that one. Then try again.

ticoann:
@yuyiguo, Yuyi, what we want is just ignore if the same file is already registered. This happens since different workflow can have the same input files. (In this case it is slight different cases with ACDC). In MySQL (MariaDB), it is faster just ignore the insert since it doesn't require the round trip to filter those files out.
So we were just wondering whether query is OK in Oracle to have the same effect as IGNORE in MariaDB. (These are the bulk insert, I would think it would be faster this way than checking the exception and filter out the duplicate data)

hufnagel:
This is the usual way how to do this on Oracle. It's still not 100% safe vs. having multiple parallel processes calling this query, but I don't think we have that ?

yuyiguo:
The query only works when the underneath binding are one by one not bulk. I think one-by-one binding is what  wmcore does, right?

amaltaro:
It seems we use up to 500 bindings together:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/DBCore.py#L175

but I was mostly worried about the syntax.
BTW, it was tested in a couple old agents that were having problems :-)

hufnagel:
The query works fine with bulk array binds.

PR #7079, opened 2016-08-22
Allow transition from force-complete to abort

### LINKED ISSUE #7078: Allow force-complete to abort transition

amaltaro:
Fixes #7078 
@ticoann please review


ticoann:
@amaltaro, Alan, patch looks good. One think I am concerned is that force-complete and aborted statust is not just status change but it triggered underlining actions. (canceling WQ and killing jobs, etc) I am not sure how this will be worked together when status change occurs

amaltaro:
It should be Ok, but we better x-check this case indeed.

amaltaro:
no longer needed, closing.

PR #7080, opened 2016-08-22
Do not upload skipped files for LogCollect/Cleanup to ACDC Server

### LINKED ISSUE #7077: LogCollect docs being uploaded to the ACDC Server (?)
For more details, see #7074 

The only thing that comes to my mind is a possible `skipBadFiles` usage by the logcollect jobs.... anyways, it needs debugging.

amaltaro:
Fixes #7077

ErrorHandler has this code already, so those files must be coming from skipped files.
@ticoann please review , still needs to be tested though

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
-#pylint: disable=E1103, W6501, E1101, C0301
+#pylint: disable=E1103, E1101, C0301
 #E1103: Use DB objects attached to thread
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan, I think you are right. I wonder where this could happen. Good Job!
Patch looks good and I will merge this.

ticoann:
We can run the test with other patches. (But I can't hardly see this can go wrong)

amaltaro:
I was only worried of not getting these skipped files into the completed (or failed) wmbs_sub_files_* table, but it seems they would go into the complete regardless.

ticoann:
test this please

PR #7081, opened 2016-08-22
implement a simple condor plugin

hufnagel:
Based on PyCondorPlugin, but with many many changes.
- get rid of worker pool with multiple threads submitting in parallel
- just do condor_submit serially (but bulk submitting lots of jobs together) 
- get rid of using agent jobid for anything here and use condor jobid instead
- get rid of common schedd query function and only query for what is needed
- status times come from the component, not the schedd

Basics (submit and track) tested but not sure how to test some of the utility functions that are not called by the agent but only by Ops scripts. Also, I still need to duplicate the relevant PyCondorPlugin unit tests.

@amaltaro @ticoann good enough for some code review and early feedback though.

@bbockelm , is this going along the lines of what we discussed ?

Anything else on the wishlist for this code ? I might as well address it all in one go.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -1,0 +1,569 @@
+#!/usr/bin/env python
+"""
+_SimpleCondorPlugin_
+
+"""
+from __future__ import print_function
+from __future__ import division
+
+import os
+import os.path
+import re
+import time
+import logging
+import threading
+
+import WMCore.Algorithms.BasicAlgos as BasicAlgos
+
+from WMCore.DAOFactory import DAOFactory
+from WMCore.WMInit import getWMBASE
+from WMCore.BossAir.Plugins.BasePlugin import BasePlugin
+from WMCore.FwkJobReport.Report import Report
+from WMCore.Credential.Proxy import Proxy
+from Utils.IterTools import grouper
+
+##  python-condor stuff
+import htcondor
+import classad
+
+
+class SimpleCondorPlugin(BasePlugin):
+    """
+    _SimpleCondorPlugin_
+
+    Condor plugin for glide-in submissions
+    """
+
+    @staticmethod
+    def stateMap():
+        """
+        For a given name, return a global state
+        """
+        stateMap = {'1': 'Pending', #Idle
+                    '2': 'Running', #Running
+                    '3': 'Error', #Removed
+                    '4': 'Complete', #Completed
+                    '5': 'Running', #Held
+                    '6': 'Running', #Transfering output
+                    '7': 'Error', #Suspended
+                    '100': 'Error'} #Unknown
+
+        return stateMap
+
+    @staticmethod
+    def exitCodeMap():
+        """
+        Exit Codes and their meaing
+        https://htcondor-wiki.cs.wisc.edu/index.cgi/wiki?p=MagicNumbers
+        """
+        exitCodeMap = {0: "Unknown",
+                       1: "Idle",
+                       2: "Running",
+                       3: "Removed",
+                       4: "Complete",
+                       5: "Held"}
+
+        return exitCodeMap
+
+    @staticmethod
+    def logToScheddExitCodeMap(x):
+        """
+        JobStatus shows the last status of the job
+        Get TriggerEventTypeNumber which is the current status of the job
+        Map it back to Schedd Status
+        Mapping done using the exit codes from condor website,
+        https://htcondor-wiki.cs.wisc.edu/index.cgi/wiki?p=MagicNumbers
+        """
+        logExitCode = {0: 1, 1: 1, 2: 0, 3: 2, 4: 3, 5: 4, 6: 2, 7: 0, 8: 0, 9: 4, 10: 0, 11: 1, 12: 5, 13: 2}
+        return logExitCode.get(x, 100)
+
+    def __init__(self, config):
+        BasePlugin.__init__(self, config)
+
+        self.locationDict = {}
+
+        myThread = threading.currentThread()
+        daoFactory = DAOFactory(package="WMCore.WMBS", logger=myThread.logger,
+                                dbinterface=myThread.dbi)
+        self.locationAction = daoFactory(classname="Locations.GetSiteInfo")
+
+        self.packageDir = None
+
+        if os.path.exists(os.path.join(getWMBASE(),
+                                       'src/python/WMCore/WMRuntime/Unpacker.py')):
+            self.unpacker = os.path.join(getWMBASE(),
+                                         'src/python/WMCore/WMRuntime/Unpacker.py')
+        else:
+            self.unpacker = os.path.join(getWMBASE(),
+                                         'WMCore/WMRuntime/Unpacker.py')
+
+        self.agent = getattr(config.Agent, 'agentName', 'WMAgent')
+        self.sandbox = None
+
+        self.scriptFile = config.JobSubmitter.submitScript
+
+        self.defaultTaskPriority = getattr(config.BossAir, 'defaultTaskPriority', 0)
+        self.maxTaskPriority = getattr(config.BossAir, 'maxTaskPriority', 1e7)
+        self.jobsPerSubmit = getattr(config.JobSubmitter, 'jobsPerSubmit', 200)
+
+        # Required for global pool accounting
+        self.acctGroup = getattr(config.BossAir, 'acctGroup', "production")
+        self.acctGroupUser = getattr(config.BossAir, 'acctGroupUser', "cmsdataops")
+
+        # Build a requirement string
+        self.reqStr = "stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS=?=\"any\") || (GLIDEIN_REQUIRED_OS=?=REQUIRED_OS)) && (TARGET.Cpus >= RequestCpus)"
+        if hasattr(config.BossAir, 'condorRequirementsString'):
+            self.reqStr = config.BossAir.condorRequirementsString
+
+        # x509 proxy handling
+        proxy = Proxy({'logger': myThread.logger})
+        self.x509userproxy = proxy.getProxyFilename()
+        self.x509userproxysubject = proxy.getSubject()
+
+        return
+
+    def submit(self, jobs, info=None):
+        """
+        _submit_
+
+
+        Submit jobs for one subscription
+        """
+        successfulJobs = []
+        failedJobs = []
+
+        if len(jobs) == 0:
+            # Then was have nothing to do
+            return successfulJobs, failedJobs
+
+        schedd = htcondor.Schedd()
+
+        # Submit the jobs
+        for jobsReady in grouper(jobs, self.jobsPerSubmit):
+
+            cluster_ad = self.getClusterAd()
+            proc_ads = self.getProcAds(jobsReady)
+
+            logging.debug("Start: Submitting %d jobs using Condor Python SubmitMany" % len(proc_ads))
+            try:
+                clusterId = schedd.submitMany(cluster_ad, proc_ads)
+            except Exception as ex:
+                logging.error("SimpleCondorPlugin job submission failed.")
+                logging.error("Moving on the the next batch of jobs and/or cycle....")
+                logging.exception(ex)
+
+                condorErrorReport = Report()
+                condorErrorReport.addError("JobSubmit", 61202, "CondorError", str(ex))
+                for job in jobsReady:
+                    job['fwjr'] = condorErrorReport
+                    failedJobs.append(job)
+            else:
+                logging.debug("Finish: Submitting jobs using Condor Python SubmitMany")
+                for index,job in enumerate(jobsReady):
+                    job['gridid'] = "%s.%s" % (clusterId, index)
+                    job['status'] = '1'
+                    successfulJobs.append(job)
+
+        # We must return a list of jobs successfully submitted and a list of jobs failed
+        logging.info("Done submitting jobs for this cycle in SimpleCondorPlugin")
+        return successfulJobs, failedJobs
+
+    def track(self, jobs):
+        """
+        _track_
+
+        Track the jobs while in condor
+        This returns a three-way ntuple
+        First, the total number of jobs still running
+        Second, the jobs that need to be changed
+        Third, the jobs that need to be completed
+        """
+
+        changeList = []
+        completeList = []
+        runningList = []
+
+        # get info about all active and recent jobs
+        logging.debug("SimpleCondorPlugin is going to track %s jobs", len(jobs))
+
+        schedd = htcondor.Schedd()
+
+        logging.debug("Start: Retrieving classAds using Condor Python XQuery")
+        try:
+            itobj = schedd.xquery("WMAgent_AgentName == %s" % classad.quote(self.agent),
+                                  ['ClusterId', 'ProcId', 'JobStatus', 'MATCH_EXP_JOBGLIDEIN_CMSSite'])
+        except Exception as ex:
+            logging.error("Query to condor schedd failed in SimpleCondorPlugin.")
+            logging.error("Returning empty lists for all job types...")
+            logging.exception(ex)
+            return runningList, changeList, completeList
+        else:
+            logging.debug("Finish: Retrieving classAds using Condor Python XQuery")
+            jobInfo = {}
+            for jobAd in itobj:
+                gridId = "%s.%s" % (jobAd['ClusterId'], jobAd['ProcId'])
+                jobStatus = str(jobAd.get('JobStatus', 100))
+                if jobStatus not in SimpleCondorPlugin.stateMap():
+                    jobStatus = '100'
+                location = jobAd.get('MATCH_EXP_JOBGLIDEIN_CMSSite', None)
+                jobInfo[gridId] = (jobStatus, location)
+
+            logging.debug("SimpleCondorPlugin retrieved %s classAds from condor schedd", len(jobInfo))
+
+        # now go over the jobs and see what we have
+        for job in jobs:
+
+            # if the schedd doesn't know a job, consider it complete
+            # doing any further checks is not cost effective
+            if job['gridid'] not in jobInfo:
+                (newStatus, location) = ('4', None)
+            else:
+                (newStatus,location) = jobInfo[job['gridid']]
+
+            # check for status changes
+            if newStatus != job['status']:
+
+                # update location info for Idle->Running transition
+                if newStatus == '2' and job['status'] == '1':
+                    if location:
+                        job['location'] = location
+                        logging.debug("JobAdInfo: Job location for jobid=%i gridid=%s changed to %s", job['jobid'], job['gridid'], location)
+
+                job['status'] = newStatus
+                job['status_time'] = int(time.time())
+                logging.debug("JobAdInfo: Job status for jobid=%i gridid=%s changed to %s", job['jobid'], job['gridid'], job['status'])
+                changeList.append(job)
+
+            job['globalState'] = SimpleCondorPlugin.stateMap()[newStatus]
+
+            # stop tracking finished jobs
+            if job['globalState'] in [ 'Complete', 'Error' ]:
+                completeList.append(job)
+            else:
+                runningList.append(job)
+
+        logging.debug("SimpleCondorPlugin tracking : %i/%i/%i (Executing/Changing/Complete)",
+                      len(runningList), len(changeList), len(completeList))
+
+        return runningList, changeList, completeList
+
+    def complete(self, jobs):
+        """
+        Do any completion work required
+
+        In this case, look for a returned logfile
+        """
+
+        for job in jobs:
+
+            if job.get('cache_dir', None) is None or job.get('retry_count', None) is None:
+                # Then we can't do anything
+                logging.error("Can't find this job's cache_dir or retry count: %s", job)
+                continue
+
+            reportName = os.path.join(job['cache_dir'], 'Report.%i.pkl' % job['retry_count'])
+            if os.path.isfile(reportName) and os.path.getsize(reportName) > 0:
+                # everything in order, move on
+                continue
+            elif os.path.isdir(reportName):
+                # Then something weird has happened. Report error, do nothing
+                logging.error("The job report for job with id %s and gridid %s is a directory", job['id'], job['gridid'])
+                logging.error("Ignoring this, but this is very strange")
+            else:
+                logging.error("No job report for job with id %s and gridid %s", job['id'], job['gridid'])
+
+                if os.path.isfile(reportName):
+                    os.remove(reportName)
+
+                # create a report from scratch
+                condorReport = Report()
+                logOutput = 'Could not find jobReport\n'
+
+                if os.path.isdir(job['cache_dir']):
+                    condorOut = "condor.%s.out" % job['gridid']
+                    condorErr = "condor.%s.err" % job['gridid']
+                    condorLog = "condor.%s.log" % job['gridid']
+                    for condorFile in [ condorOut, condorErr, condorLog ]:
+                        condorFilePath = os.path.join(job['cache_dir'], condorFile)
+                        if os.path.isfile(condorFilePath):
+                            logTail = BasicAlgos.tail(condorFilePath, 50)
+                            logOutput += 'Adding end of %s to error message:\n' % condorFile
+                            logOutput += '\n'.join(logTail)
+                    condorReport.addError("NoJobReport", 99303, "NoJobReport", logOutput)
+                else:
+                    msg = "Serious Error in Completing condor job with id %s!\n" % job['id']
+                    msg += "Could not find jobCache directory %s\n" % job['cache_dir']
+                    msg += "Creating a new cache_dir for failed job report\n"
+                    logging.error(msg)
+                    os.makedirs(job['cache_dir'])
+                    condorReport.addError("NoJobReport", 99304, "NoCacheDir", logOutput)
+
+                condorReport.save(filename=reportName)
+
+                logging.debug("Created failed job report for job with id %s and gridid %s", job['id'], job['gridid'])
+
+        return
+
+    def updateSiteInformation(self, jobs, siteName, excludeSite):
+        """
+        _updateSiteInformation_
+
+        Allow or disallow jobs to run at a site.
+        Called externally by Ops scripts if a site enters or leaves Down, Draining or Aborted.
+
+        Kill job if after removing site from allowed sites it has nowhere to run.
+
+        Parameters:    excludeSite = False when moving to Normal
+                       excludeSite = True when moving to Down, Draining or Aborted
+        """
+        schedd = htcondor.Schedd()
+
+        jobtokill = []
+        try:
+            itobj = schedd.xquery('WMAgent_AgentName =?= %s && JobStatus =?= 1' % classad.quote(self.agent),
+                                  ['ClusterId', 'ProcId', 'DESIRED_Sites', 'ExtDESIRED_Sites'])
+        except Exception as ex:
+            logging.error("Failed to query condor schedd.")
+            logging.exception(ex)
+            return jobtokill
+        else:
+            jobInfo = {}
+            for jobAd in itobj:
+                gridId = "%s.%s" % (jobAd['ClusterId'], jobAd['ProcId'])
+                jobInfo[gridId] = jobAd
+            for job in jobs:
+                jobAd = jobInfo.get(job['gridid'], None)
+                if jobAd:
+                    desiredSites = jobAd.get('DESIRED_Sites').split(',')
+                    extDesiredSites = jobAd.get('ExtDESIRED_Sites').split(',')
+                    if excludeSite:
+                        # Remove siteName from DESIRED_Sites if job has it
+                        if siteName in desiredSites:
+                            if len(desiredSites) > 1:
+                                desiredSites.remove(siteName)
+                                desiredSites = ','.join(desiredSites)
+                                try:
+                                    schedd.edit([job['gridid']], 'DESIRED_Sites', classad.ExprTree('"%s"' % desiredSites))
+                                except Exception as ex:
+                                    logging.error("Failed to edit sites for job %s" % job['gridid'])
+                                    logging.exception(ex)
+                            else:
+                                jobtokill.append(job)
+                    else:
+                        # Add siteName to DESIRED_Sites if ExtDESIRED_Sites has it (moving back to Normal)
+                        if siteName not in desiredSites and siteName in extDesiredSites:
+                            desiredSites.append(siteName)
+                            desiredSites = ','.join(sorted(desiredSites))
+                            try:
+                                schedd.edit([job['gridid']], 'DESIRED_Sites', classad.ExprTree('"%s"' % desiredSites))
+                            except Exception as ex:
+                                logging.error("Failed to edit sites for job %s" % job['gridid'])
+                                logging.exception(ex)
+
+        return jobtokill
+
+    def kill(self, jobs):
+        """
+        _kill_
+
+        Kill a list of jobs based on the WMBS job names.
+        Kill can happen for schedd running on localhost... TBC.
+        """
+        logging.info("Killing %i jobs from the queue", len(jobs))
+
+        schedd = htcondor.Schedd()
+        gridIds = [job['gridid'] for job in jobs]
+        try:
+            schedd.act(htcondor.JobAction.Remove, gridIds)
+        except RuntimeError:
+            logging.warn("Error while killing jobs on the schedd: %s", gridIds)
+
+        return
+
+    def killWorkflowJobs(self, workflow):
+        """
+        _killWorkflowJobs_
+
+        Kill all the jobs belonging to a specific workflow.
+        """
+        logging.info("Going to remove all the jobs for workflow %s", workflow)
+
+        schedd = htcondor.Schedd()
+
+        try:
+            schedd.act(htcondor.JobAction.Remove, "WMAgent_RequestName == %s" % classad.quote(workflow))
+        except RuntimeError:
+            logging.warn("Error while killing jobs on the schedd: WMAgent_RequestName=%s", workflow)
+
+        return
+
+    def updateJobInformation(self, workflow, task, **kwargs):
+        """
+        _updateJobInformation_
+
+        Update job information for all jobs in the workflow and task,
+        the change will take effect if the job is Idle or becomes idle.
+
+        The currently supported changes are only priority for which both the task (taskPriority)
+        and workflow priority (requestPriority) must be provided.
+        """
+        schedd = htcondor.Schedd()
+
+        if 'taskPriority' in kwargs and 'requestPriority' in kwargs:
+            newPriority = int(kwargs['requestPriority']) + int(kwargs['taskPriority'] * self.maxTaskPriority)
+            try:
+                constraint = "WMAgent_SubTaskName =?= %s" % classad.quote(task)
+                constraint += " && WMAgent_RequestName =?= %s" % classad.quote(workflow)
+                constraint += " && JobPrio =!= %d" % newPriority
+                schedd.edit(constraint, 'JobPrio', classad.Literal(newPriority))
+            except Exception as ex:
+                logging.error("Failed to update JobPrio for WMAgent_SubTaskName=%s", task)
+                logging.exception(ex)
+
+        return
+
+    def getClusterAd(self):
+        """
+        _initSubmit_
+
+        Return common cluster classad
+
+        scriptFile & Output/Error/Log filenames shortened to
+        avoid condorg submission errors from >256 chars paths
+
+        """
+        ad = classad.ClassAd()
+
+        #ad['universe'] = "vanilla"
+        ad['Requirements'] = classad.ExprTree(self.reqStr)
+        ad['ShouldTransferFiles'] = "YES"
+        ad['WhenToTransferOutput'] = "ON_EXIT"
+        ad['UserLogUseXML'] = True
+        ad['JobNotification'] = 0
+        ad['Cmd'] = self.scriptFile
+        ad['Out'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".out")')
+        ad['Err'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".err")')
+        ad['UserLog'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".log")')
+
+        ad['WMAgent_AgentName'] = self.agent
+
+        ad['JOBGLIDEIN_CMSSite'] = classad.ExprTree('isUndefined(GLIDEIN_CMSSite) ? Unknown : GLIDEIN_CMSSite')
+
+        ad['JobLeaseDuration'] = classad.ExprTree('isUndefined(MachineAttrMaxHibernateTime0) ? 1200 : MachineAttrMaxHibernateTime0')
+
+        # Required for global pool accounting
+        ad['AcctGroup'] = self.acctGroup
+        ad['AcctGroupUser'] = self.acctGroupUser
+
+        # Customized classAds for this plugin
+        ad['DESIRED_Archs'] = "INTEL,X86_64"
+
+        ad['x509userproxy'] = self.x509userproxy
+        ad['x509userproxysubject'] = self.x509userproxysubject
+
+        ad['Rank'] = 0.0
+        ad['TransferIn'] = False
+
+        # TODO: remove when 8.5.7 is deployed
+        params_to_add = htcondor.param['SUBMIT_ATTRS'].split() + htcondor.param['SUBMIT_EXPRS'].split()
+        params_to_skip = ['accounting_group', 'use_x509userproxy', 'PostJobPrio2', 'JobAdInformationAttrs']
+        for param in params_to_add:
+            if (param not in ad) and (param in htcondor.param) and (param not in params_to_skip):
+                ad[param] = classad.ExprTree(htcondor.param[param])
+
+        return ad
+
+    def getProcAds(self, jobList):
+        """
+        _getProcAds_
+
+        Return list of job specific classads for submission
+
+        """
+        classAds = []
+        for job in jobList:
+            ad = classad.ClassAd()
+
+            ad['Iwd'] = job['cache_dir']
+            ad['TransferInput'] = "%s,%s/%s,%s" % (job['sandbox'], job['packageDir'],
+                                                   'JobPackage.pkl', self.unpacker)
+            ad['Arguments'] = "%s %i" % (os.path.basename(job['sandbox']), job['id'])
+
+            ad['TransferOutput'] = "Report.%i.pkl" % job["retry_count"]
+
+            ad['JobMachineAttrs'] = "GLIDEIN_CMSSite"
+            ad['JobAdInformationAttrs'] = "JobStatus,QDate,EnteredCurrentStatus,JobStartDate,DESIRED_Sites,ExtDESIRED_Sites,WMAgent_JobID,MATCH_EXP_JOBGLIDEIN_CMSSite"
+
+            sites = ','.join(sorted(job.get('possibleSites')))
+            ad['DESIRED_Sites'] = sites
+
+            sites = ','.join(sorted(job.get('potentialSites')))
+            ad['ExtDESIRED_Sites'] = sites
+
+            ad['WMAgent_RequestName'] = job['requestName']
+
+            match = re.compile("^[a-zA-Z0-9_]+_([a-zA-Z0-9]+)-").match(job['requestName'])
+            if match:
+                ad['CMSGroups'] = match.groups()[0]
+            else:
+                ad['CMSGroups'] = classad.Value.Undefined
+
+            ad['WMAgent_JobID'] = job['jobid']
+            ad['WMAgent_SubTaskName'] = job['taskName']
+            ad['CMS_JobType'] = job['taskType']
+
+            # Handling for AWS, cloud and opportunistic resources
+            ad['AllowOpportunistic'] = job.get('allowOpportunistic', False)
+
+            if job.get('inputDataset'):
+                ad['DESIRED_CMSDataset'] = job['inputDataset']
+            else:
+                ad['DESIRED_CMSDataset'] = classad.Value.Undefined
+
+            if job.get('inputDatasetLocations'):
+                sites = ','.join(sorted(job['inputDatasetLocations']))
+                ad['DESIRED_CMSDataLocations'] = sites
+            else:
+                ad['DESIRED_CMSDataLocations'] = classad.Value.Undefined
+
+            # HighIO jobs
+            ad['Requestioslots'] = int(job.get('highIOjob', False))
+
+            # Performance and resource estimates
+            numberOfCores = job.get('numberOfCores', 1)
+            ad['RequestCpus'] = numberOfCores
+            ad['RequestMemory'] = int(job['estimatedMemoryUsage']) if job.get('estimatedMemoryUsage', None) else 1000
+            ad['RequestDisk'] = int(job['estimatedDiskUsage']) if job.get('estimatedDiskUsage', None) else 20*1000*1000*numberOfCores
+            ad['MaxWallTimeMins'] = int(job['estimatedJobTime'])/60.0 if job.get('estimatedJobTime', None) else 12*6
+
+            taskPriority = job.get('taskPriority', self.defaultTaskPriority)
+            try:
+                taskPriority = int(taskPriority)
+            except ValueError:
+                logging.error("Job taskPriority %s not an int, using default", taskPriority)
+                taskPriority = self.defaultTaskPriority
+
+            priority = job.get('priority', 0)
+            try:
+                priority = int(priority)
+            except ValueError:
+                logging.error("Job priority %s not an int, using 0", priority)
+                priority = 0
+
+            ad['JobPrio'] = int(taskPriority + priority * self.maxTaskPriority)
+
+            postJobPrio1 = -1 * len(job.get('potentialSites', []))
+            postJobPrio2 = -1 * job['taskID']
+
+            ad['PostJobPrio1'] = int(postJobPrio1)
+            ad['PostJobPrio2'] = int(postJobPrio2)
+
+            # Add OS requirements for jobs
+            if job.get('scramArch') is not None and job.get('scramArch').startswith("slc6_"):
+                ad['REQUIRED_OS'] = "rhel6"
+            else:
+                ad['REQUIRED_OS'] = "any"
+
+            classAds.append((ad,1))
+
+        return classAds


src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+    if 'ERROR: proxy has expired\n' in error:
+        errorCondition = True
+        errorMsg = 'CRITICAL ERROR: Your proxy has expired!\n'
+    elif '999100\n' in error:
+        errorCondition = True
+        errorMsg = "CRITICAL ERROR: Failed to build submit command!\n"
+    elif 'Failed to open command file' in error:
+        errorCondition = True
+        errorMsg = "CONDOR ERROR: jdl file not found by submitted jobs!\n"
+    elif 'It appears that the value of pthread_mutex_init' in error:
```
bbockelm: Dump the glexec code.  This was never used.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+            jdlFile = "%s/submit_%i_%i.jdl" % (self.submitDir, os.getpid(), idList[0])
+
+            with open(jdlFile, 'w') as handle:
+                handle.writelines(jdlList)
+            jdlFiles.append(jdlFile)
+
+            # Now submit them
+            logging.info("About to submit %i jobs", len(jobsReady))
+
+            command = "condor_submit --terse -disable %s" % jdlFile
```
bbockelm: Can we switch to pure python so we don't have to resort to parsing the stdout/err?

bbockelm: In particular, look at `submit` with the `transaction` object or `submitMany` for an auto-transaction.  If you want to really be lazy, use the latest release and the standalone `Submit` object.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+                logging.error("Went to check on error report for job %i.  Found a directory instead.\n", job['id'])
+                logging.error("Ignoring this, but this is very strange.\n")
+
+            # If we're still here, we must not have a real error report
+            logOutput = 'Could not find jobReport\n'
+            # But we don't know exactly the condor id, so it will append
+            # the last lines of the latest condor log in cache_dir
+            # TODO: Alan - how about we look at the .out or .err file instead of the .log?
+            genLogPath = os.path.join(job['cache_dir'], 'condor.*.*.log')
+            logPaths = glob.glob(genLogPath)
```
bbockelm: Globs are relatively expensive.  Don't we already know the precise file name?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+            logging.debug("No returning job report for job %i", job['id'])
+
+        return
+
+    def updateSiteInformation(self, jobs, siteName, excludeSite):
+        """
+        _updateSiteInformation_
+
+        Modify condor classAd for all Idle jobs for a site if it has gone Down, Draining or Aborted.
+        Kill all jobs if the site is the only site for the job.
```
bbockelm: Why should we kill the job for draining?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+                                        logging.debug("Failed to edit sites for job %s" % job['gridid'])
+                                else:
+                                    jobtokill.append(job)
+                        else:
+                            # Add siteName to DESIRED_Sites if ExtDESIRED_Sites has it (moving back to Normal)
+                            if siteName not in desiredSites and siteName in extDesiredSites:
+                                desiredSites.append(siteName)
+                                desiredSites = ','.join(desiredSites)
+                                try:
+                                    schedd.edit([job['gridid']], 'DESIRED_Sites', classad.ExprTree('"%s"' % desiredSites))
```
bbockelm: Sort the `DESIRED_Sites` first to prevent combinatorics of auto-clusters.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        except RuntimeError:
+            logging.warn("Error while killing jobs on the schedd: %s", gridIds)
+
+        return
+
+    def killWorkflowJobs(self, workflow):
+        """
+        _killWorkflowJobs_
+
+        Kill all the jobs belonging to a specif workflow.
```
bbockelm: typo

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        jdl.append("Output = condor.$(Cluster).$(Process).out\n")
+        jdl.append("Error = condor.$(Cluster).$(Process).err\n")
+        jdl.append("Log = condor.$(Cluster).$(Process).log\n")
+
+        jdl.append("+WMAgent_AgentName = \"%s\"\n" % (self.agent))
+        jdl.append("+JOBGLIDEIN_CMSSite= \"$$([ifThenElse(GLIDEIN_CMSSite is undefined, \\\"Unknown\\\", GLIDEIN_CMSSite)])\"\n")
+
+        # Required for global pool accounting
+        jdl.append("+AcctGroup = \"%s\"\n" % (self.acctGroup))
+        jdl.append("+AcctGroupUser = \"%s\"\n" % (self.acctGroupUser))
```
bbockelm: Use provided quoting methods.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        jdl.append("when_to_transfer_output = ON_EXIT\n")
+        jdl.append("log_xml = True\n")
+        jdl.append("notification = NEVER\n")
+        jdl.append("Executable = %s\n" % self.scriptFile)
+        jdl.append("Output = condor.$(Cluster).$(Process).out\n")
+        jdl.append("Error = condor.$(Cluster).$(Process).err\n")
+        jdl.append("Log = condor.$(Cluster).$(Process).log\n")
+
+        jdl.append("+WMAgent_AgentName = \"%s\"\n" % (self.agent))
+        jdl.append("+JOBGLIDEIN_CMSSite= \"$$([ifThenElse(GLIDEIN_CMSSite is undefined, \\\"Unknown\\\", GLIDEIN_CMSSite)])\"\n")
```
bbockelm: Use provided quoting method to avoid this nasty expression.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+
+        # Required for global pool accounting
+        jdl.append("+AcctGroup = \"%s\"\n" % (self.acctGroup))
+        jdl.append("+AcctGroupUser = \"%s\"\n" % (self.acctGroupUser))
+        jdl.append("+AccountingGroup = \"%s.%s\"\n" % (self.acctGroup, self.acctGroupUser))
+
+        # Customized classAds for this plugin
+        jdl.append('+DESIRED_Archs = \"INTEL,X86_64\"\n')
+        jdl.append('+REQUIRES_LOCAL_DATA = True\n')
```
bbockelm: I doubt this works - looks like an abandoned attempt at doing overflow?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+                    logging.error("Not setting priority")
+                    logging.debug("Priority: %s", job['priority'])
+                except Exception as ex:
+                    logging.error("Got unhandled exception while setting priority for job %i\n", job['id'])
+                    logging.error(str(ex))
+                    logging.error("Not setting priority")
+
+            jdl.append("priority = %i\n" % (task_priority + prio * self.maxTaskPriority))
+
+            jdl.append("+PostJobPrio1 = -%d\n" % len(job.get('potentialSites', [])))
```
bbockelm: Check with unified to see what the current scheme is for this.  I'm pretty sure Unified fixes up this later to simplify negotiation.  Might as well do the correct thing on submit now.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+
+        # dataset info
+        if job.get('inputDataset'):
+            jdl.append('+DESIRED_CMSDataset = "%s"\n' % job['inputDataset'])
+        else:
+            jdl.append('+DESIRED_CMSDataset = undefined\n')
+        if job.get('inputDatasetLocations'):
+            jdl.append('+DESIRED_CMSDataLocations = "%s"\n' % ','.join(job['inputDatasetLocations']))
+        else:
+            jdl.append('+DESIRED_CMSDataLocations = undefined\n')
```
bbockelm: It would be very, very, very useful to also provide the block being processed.  One of the bigger limitations now...

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        requestMemory = int(job['estimatedMemoryUsage']) if job.get('estimatedMemoryUsage', None) else 1000
+        requestDisk = int(job['estimatedDiskUsage']) if job.get('estimatedDiskUsage', None) else 20*1000*1000*numberOfCores
+        maxWallTimeMins = int(job['estimatedJobTime'])/60.0 if job.get('estimatedJobTime', None) else 12*60
+        jdl.append('request_memory = %d\n' % requestMemory)
+        jdl.append('request_disk = %d\n' % requestDisk)
+        jdl.append('+MaxWallTimeMins = %d\n' % maxWallTimeMins)
+
+        # How many cores job is using
+        jdl.append('machine_count = 1\n')
+        jdl.append('request_cpus = %s\n' % numberOfCores)
```
bbockelm: Why is this a string?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        numberOfCores = job.get('numberOfCores', 1)
+        requestMemory = int(job['estimatedMemoryUsage']) if job.get('estimatedMemoryUsage', None) else 1000
+        requestDisk = int(job['estimatedDiskUsage']) if job.get('estimatedDiskUsage', None) else 20*1000*1000*numberOfCores
+        maxWallTimeMins = int(job['estimatedJobTime'])/60.0 if job.get('estimatedJobTime', None) else 12*60
+        jdl.append('request_memory = %d\n' % requestMemory)
+        jdl.append('request_disk = %d\n' % requestDisk)
+        jdl.append('+MaxWallTimeMins = %d\n' % maxWallTimeMins)
+
+        # How many cores job is using
+        jdl.append('machine_count = 1\n')
```
bbockelm: What is this for?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        jdl.append('request_cpus = %s\n' % numberOfCores)
+
+        # Add OS requirements for jobs
+        if job.get('scramArch') is not None and job.get('scramArch').startswith("slc6_"):
+            jdl.append('+REQUIRED_OS = "rhel6"\n')
```
bbockelm: Is this code relevant any more?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        jdl.append("Error = condor.$(Cluster).$(Process).err\n")
+        jdl.append("Log = condor.$(Cluster).$(Process).log\n")
+
+        jdl.append("+WMAgent_AgentName = \"%s\"\n" % (self.agent))
+        jdl.append("+JOBGLIDEIN_CMSSite= \"$$([ifThenElse(GLIDEIN_CMSSite is undefined, \\\"Unknown\\\", GLIDEIN_CMSSite)])\"\n")
+
+        # Required for global pool accounting
+        jdl.append("+AcctGroup = \"%s\"\n" % (self.acctGroup))
+        jdl.append("+AcctGroupUser = \"%s\"\n" % (self.acctGroupUser))
+        jdl.append("+AccountingGroup = \"%s.%s\"\n" % (self.acctGroup, self.acctGroupUser))
```
bbockelm: None of this is idiomatic any more.  Check the `condor_submit` man page for the current usage of accounting groups.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        """
+        jdl = []
+        jobCE = job['location']
+        if not jobCE:
+            # Then we ended up with a site that doesn't exist?
+            logging.error("Job for non-existant site %s", job['location'])
+            return jdl
+
+        if self.submitWMSMode and len(job.get('possibleSites', [])) > 0:
+            strg = ','.join(map(str, job.get('possibleSites')))
```
bbockelm: Sort please.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+            jdl.append('+DESIRED_Sites = \"%s\"\n' % strg)
+        else:
+            jdl.append('+DESIRED_Sites = \"%s\"\n' % (jobCE))
```
bbockelm: Is this really live code?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+        jdl = []
+        jobCE = job['location']
+        if not jobCE:
+            # Then we ended up with a site that doesn't exist?
+            logging.error("Job for non-existant site %s", job['location'])
+            return jdl
+
+        if self.submitWMSMode and len(job.get('possibleSites', [])) > 0:
+            strg = ','.join(map(str, job.get('possibleSites')))
+            jdl.append('+DESIRED_Sites = \"%s\"\n' % strg)
```
bbockelm: Use provided quoting methods.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,720 @@
...
+
+        jdl.append('+WMAgent_RequestName = "%s"\n' % job['requestName'])
+        match = GROUP_NAME_RE.match(job['requestName'])
+        if match:
+            jdl.append('+CMSGroups = %s\n' % classad.quote(match.groups()[0]))
+        else:
+            jdl.append('+CMSGroups = undefined\n')
+
+        jdl.append('+WMAgent_SubTaskName = "%s"\n' % job['taskName'])
+        jdl.append('+CMS_JobType = "%s"\n' % job['taskType'])
```
bbockelm: Use provided quoting methods on above two lines.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: Will do.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: Will look into it. Also avoids creating a temporary jdl file, although the HTCondor manual talks about different syntax for the jdl and the classads for the submission, so will have to investigate.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: We should, will look into it.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: All the idle jobs, should correct the comment. Does that make more sense ? Although why we would only kill the idle jobs and not the running jobs either is a bit of a question to me. Either way they'll lead to job failures in the agent and it's not like a running job is guaranteed to finish anytime soon. Could have just started...

@ticoann @ericvaandering want comment ?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: Looked into it. There isn't enough documentation to do this. I took an example jdl file and dumped it into classads as described in the documentation. Where I am stuck is this:

"""Then, load the resulting contents of <filename> into Python. """

How ? I guess I need to convert this into a classad instance somehow, but

a) I don't want to go via a file if I can avoid it
b) with bulk submission there are lots of duplicate parameters since they differ for different jobs

Any example code how to do this ?

hufnagel: Also, submit needs a transaction ? Documentation doesn't mention it...

ericvaandering:
You have a syntax error: RefactoringTool: Can't parse src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py: ParseError: bad input: type=8, value=u')', context=('', (374, 92))

Most likely line 374

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,683 @@
...
+                if os.path.isdir(job['cache_dir']):
+                    # TODO: Alan - how about we look at the .out or .err file instead of the .log ?
+                    condorLog = os.path.join(job['cache_dir'], "condor.%s.log" % job['gridid'])
+                    if os.path.isfile(condorLog):
+                        logTail = BasicAlgos.tail(condorLog, 50)
+                        logOutput += 'Adding end of condor.log to error message:\n'
+                        logOutput += '\n'.join(logTail)
+                    condorReport.addError("NoJobReport", 99303, "NoJobReport", logOutput)
+                else:
+                    msg = "Serious Error in Completing condor job with id %s!\n" % job['id'])
```
ericvaandering: Yeah, here it is. A trailing parenthesis

hufnagel:
Thanks for noticing. Fixed it, lets see if it goes through now.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ticoann: It only kills the jobs when all the desired sites are in drain status since in that case jobs can't be running any sites.
Other cases, it just modifies classAd for Desided_sites list to remove the draining site.
I am not completely sure why we keep the running jobs as it is. (I think we hope it might run if not will at least and be failed/comes back) But maybe it is better to kill those as well.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
bbockelm: Yeah - we can talk about this on Thursday.  The documentation meant to just do the `-dryrun` once to help figure out the attribute names (by hand, but do pure python otherwise - no writing out files at all).

For the transaction stuff: you can either do:
```
  with schedd.transaction() as txn:
     schedd.submit(...)
     schedd.submit(...)
     schedd.submit(...)
```
or, somewhat equivalently,
```
schedd.submitMany(cluster_ad, [(ad1, 1), (ad2, 1), (ad3, 1)], ...)
```

Multiple `submit`s inside a transaction does one cluster per `submit` while `submitMany` does a single cluster.  Both forms do a single transaction in terms of schedd resources; `submitMany` may use less memory since the schedd only saves the differences in the proc ad compared to the cluster ad.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
bbockelm: What's the benefit of killing the jobs?  (Particularly, as opposed to keeping them in the queue for when the site comes back online?)

For one-hour downtimes, it seems like overkill - especially if the downtime is only for the storage system and jobs will continue to run.

amaltaro: To complement what Seangchan said. We only kill pending jobs when the DESIRED_Sites has only one site (the one that is going to Down/Drain), and I agree with this behavior.
On the running jobs, I think it's better to give them a chance and let them run.

ticoann: Brian, you are right if the site comes back in an hour this is overkill.
However, this code is not triggered by agent. So there is no automatic kill for this situation. This is used by Ops' script. So I think it is only used when Ops decides when the situation is right.

bbockelm: @hufnagel - in that case, can you bring this up at the next CompOps meeting to get guidance on what their desire is for this behavior?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: Question, if I have 1000 jobs to submit, does it matter if I break this into 5 individual submits within the same transaction or 5 (adN, 1) blocks in a submitMany ? Or can I just make a single 1000 job (or 10000 job for that matter) submit or submitMany call. The current code assumes it's bad to submit too many jobs in one go, even if that is just a single bulk submit call.

hufnagel: Oh, and that also means: is it bad if there are too many jobs in a single cluster ?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: I would just remove the limit on Idle jobs again here (I added it). Then it's up to the Ops scripts that call this whether they select only Idle jobs or Idle and Running jobs.

As far as it goes, IMHO if you want to evacuate a site, you want to evacuate immediately, just kill everything.

There is no functional difference between killing Idle jobs and Running jobs, both cause the same problems on the agent side (since Idle == Running as far as the agent is concerned).

I don't really attend CompOps meetings anymore, but if this could be put on the agenda, I could call in (next Monday).

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ericvaandering: Also just use a list expression instead of map(). That's preferred now.

ericvaandering:
@hufnagel can you comment on what the motivation for doing this? More than what you wrote in the initial post which is mostly a list of what you want to change without much "why"? Or a pointer to somewhere this is discussed (quite likely I missed a mail or something).

Thanks!

ericvaandering:
I didn't notice the syntax error, Jenkins did. Failures in the pylint test are a good indication. I suppose I could make it notice these and post back to the issue.

ticoann:
That would be nice that if Jenkins can report back syntax error separate from pylint (warnings/score).

hufnagel:
@ericvaandering , I always hated the current implementation of the Condor BossAir plugins in that it tracked by agent job id and not Condor job id. I was in the the original design meetings when we discussed BossAir and there were and still are very good reasons not to track by agent job id (basically an agent job is different from a specific instance of that job) . Commented on it a few times already during the evolution of the plugins but my comments were ignored. Getting that aspect of the plugin changed had been on my internal todo list for a couple years maybe.

Then when I was in Nebraska I talked with Brian quite a bit about the shortcomings of the current Condor BossAir plugin and we made a wish list for changes.

In addition I had the code fresh in my mind since I recently improved on the bulk submission, so I just gave it a go, took only a few days non-continuous work (had some time to kill in between working on NERSC and waiting for pilots to run) to have a first version, at which point I started asking questions in HN and kept improving the code.

hufnagel:
As for where this was discussed before, comments in various tickets related to PyCondorPlugin and I mentioned it a few times in release planning meetings (the last one I think as well). Other than that this started exploratory and is now basically "write a somewhat clean implementation of an all python based Condor BossAir plugin and remove accumulated cruft". No push from my side to use this in operations anytime soon. Given the amount of validation such a changeover involves, I'd rather change everything at once in one go and then we can start testing slowly. If we see a clear improvement (in performance) to what we are using now we can always accelerate the schedule later.

hufnagel:
Also the reason I implemented a new plugin instead of patching PyCondorPlugin. With a new plugin we can have it in the code and for tests (or in the Tier0) it can be selected in the resource setup and agent configuration.

ericvaandering:
Thanks @hufnagel . I'll be at CERN next week so maybe we can chat over drinks. I'm not opposed at all just trying to understand.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: @ericvaandering added it with some multicore changes. It translates into MaxHosts=1 for the classads. Is that really needed though ? That's the default and only supported option, isn't it ? Or can you request multiple hosts on HTCondor for MPI jobs ?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: According to the man page, AcctGroup and AcctGroupUser is correct and AccountingGroup does not exist, so I removed the last line.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: Yes. In practice self.submitWMSMode is always True though, since the agents are configured that way. I think this flag was intended for direct condor submissions if someone would want to setup a site where we submit directly to the batch system.

IMO we can get rid of that since in that mode a lot of other code in this plugin makes no sense whatsoever. This is pretty much a gWMS only plugin at this point.

Also not sure if the job submitter even passes jobs for submission that have no possible or potential sites. I think it throws an error on these and immediately fails them for submission. Would allow us to simplify this code here quite a bit.

hufnagel: Yes, the JobSubmitter checks for empty possibleLocations and immediately fails the job without even submitting it. So that check for empty possibleLocations is pointless. For potentialLocations too for that matter (if it is empty then possibleLocations is empty too and the plugin never sees the job).

As far as jobCE is concerned, this is badly named, but live code. The JobSubmitter "assigns" a job to one of the possible locations just to have mapping of job to site for its internal accounting (before the job goes to running and this is updated with the correct site).

hufnagel: @ericvaandering @amaltaro @ticoann you ok with me removing self.submitWMSMode and the length checks ?

hufnagel:
Ok, worked through all the comments. For some of them we are still discussing things, but due to changes around close code lines the diffs will be outdated. In that case just comment on the new commit.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ericvaandering: I'm sure I got a recipe from someone if I did this. I claim no knowledge anymore of condor settings.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
bbockelm: The schedd has to do one file operation per new job (write to the user log).  So, O(1k) is probably fine: I would prefer to avoid O(20k) jobs.  Somewhere around the 10s of thousands level we can probably expect the writes to start blocking on disk IO.

I can't think of any reason why a given strategy of breaking up into multiple submit calls would be better than another.  There's certainly a small amount of memory savings to have as few attributes in the proc ad as possible (since the cluster ad is shared among all processes).  This seems pretty minimal though.

I do recommend holding the transaction open for as short time as possible: build up all the dictionaries first, then do submit.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
bbockelm: This would be something you'd only set for the HTCondor parallel universe (which we definitely aren't using).  Let's remove it.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: @bbockelm , would you think that schedd.submit() on individual jobs as long as all share the same transaction is doable performance wise ? That I would know how to do, just not sure how to pack multiple jobs into a single submit call.

bbockelm: I think that should be fine, but it's worth to check the timing in testing.  Maybe log the submit rate so we can notice performance issues in the future?

What part were you having issues with in packing into `submit` / `submitMany` calls?

hufnagel: I wanted to know how to submit multiple jobs with very different classads via single submit() call. Maybe that isn't possible ? Do I need submitMany() for this ? Just talked to Farrukh, seems we are using HTCondor 8.5.4, which should support submitMany().

bbockelm: Yes - I would lean toward `submitMany` for that.

hufnagel: Ok, I have an implementation with submitMany(), but I can't use it since the condor rpm we deploy with the agents is 8.3.1 and we cannot update to 8.5.4 due to build errors. See #7127

hufnagel:
Ok, pending code review and new information this is my final version. This only works with a new condor 8.5.4 package, we are still working out how to get that into COMP exactly.

Good enough for code review code. @ticoann @amaltaro @bbockelm

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+                logging.exception(ex)
+
+                condorErrorReport = Report()
+                condorErrorReport.addError("JobSubmit", 61202, "CondorError", str(ex))
+                for job in jobsReady:
+                    job['fwjr'] = condorErrorReport
+                    failedJobs.append(job)
+
+                # Report submission error to operators
+                # TODO - Dirk: Do we really need this ?
```
hufnagel: @ticoann , do we still need the alerts ? Weren't you deprecating or changing this ?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+
+                if os.path.isfile(reportName):
+                    os.remove(reportName)
+
+                # create a report from scratch
+                condorReport = Report()
+                logOutput = 'Could not find jobReport\n'
+
+                if os.path.isdir(job['cache_dir']):
+                    # TODO: Alan - how about we look at the .out or .err file instead of the .log ?
```
hufnagel: @amaltaro , can we decide what we want to do here and then either implement the change or leave things as they are ? Leaving this comment here for (more) years isn't really buying us anything...

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+        ad['Out'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".out")')
+        ad['Err'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".err")')
+        ad['UserLog'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".log")')
+
+        ad['WMAgent_AgentName'] = self.agent
+
+        ad['JOBGLIDEIN_CMSSite'] = classad.ExprTree('isUndefined(GLIDEIN_CMSSite) ? Unknown : GLIDEIN_CMSSite')
+
+        ad['JobLeaseDuration'] = classad.ExprTree('isUndefined(MachineAttrMaxHibernateTime0) ? 1200 : MachineAttrMaxHibernateTime0')
+
```
hufnagel: I implemented this to follow the last PyCondorPlugin commit (addresses HLT use case). Also used it for JOBGLIDEIN_CMSSite since it's a similar construct and the new syntax is easier to read.

@bbockelm , is this syntax correct ? When testing it seems to "work" (job is submitted and runs to completion, not sure that says anything though).

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
amaltaro: Feel free to remove them. They have vanished from almost all components by now.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
amaltaro: I'm not sure whether we get anything back from condor into .out or .err in cases where the job fails during the bootstrap (< 1min). We need to watch those 2 logs to be able to answer :)

hufnagel: Ok, but what's the logic here then ? Look at out and err files, if there is something there include the last 50 lines and fall back to the log file if we can't get anything out of the out and err files ? From experience the .out and .err files are too large and a relevant error could be anywhere in the middle of it.

There either is an automated way to extract useful information from these files or there isn't. If the former we should implement it and remove this comment. If the latter we should just remove the comment.

A comment like this is useless on it's own.

amaltaro: Providing the last 50 lines of the log is as useless as the comment, but the comment at least points there is something not right :-)

> Ok, but what's the logic here then ? Look at out and err files, if there is something there include the last 50 lines and fall back to the log file if we can't get anything out of the out and err files ? From experience the .out and .err files are too large and a relevant error could be anywhere in the middle of it.

Yes, I'd get the last 50 lines of the err file. If empty, then get it from .out. If empty as well, then go to the useless (for my eyes) .log file.
You have a good point, the error can be anywhere in the out file, depending at which stage the failure happened.

hufnagel: So should we get more than 50 lines ? Is there a size limit for embedding this informaton in the report ?

amaltaro: I leave this question to @ticoann Right now he's collecting the last 25 lines in this patch:
https://github.com/dmwm/WMCore/pull/7117/files

May be not to overload wmstats with too much information... ?

hufnagel: The thing is, this information is only collected if the job got killed without generating a job report from CMSSW. This should be a very small fraction of jobs and for these jobs you want to have all the debugging information you need...

hufnagel: That patch is not that relevant here since CMSSW usually does print the problem at the end of its log output. The Condor logs on the other hand...

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
bbockelm: When you set this, what's the resulting line in the ClassAd?

I'm not sure, but I would have guessed that you need to set `+JobLeaseDuration` since you are giving it an expression, not an integer.

amaltaro: Brian is right. Look at the original PR in #7166

hufnagel: Discussed it with Brian, this seems to work as-is. It seems with the classAd and python bindings you don't need the +Param syntax at all, in fact, I am not using it anywhere.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ticoann: I think 50 line is fine. It shouldn't make a big difference unless we propagate whole log file.

ticoann: Sorry I miss understood discussion. If the meaning full messages are anywhere in the out, log, error file. We need to some how extract the infromation from the file, By merely increasing the number of lines to extract won't get us much. 
We need to collect the files and look at those and whether we can find some patterns.  (string match with error, traceback, etc)
But we can do this in separate patch (issue).

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+        logging.debug("SimpleCondorPlugin is going to track %s jobs", len(jobs))
+
+        schedd = htcondor.Schedd()
+
+        logging.debug("Start: Retrieving classAds using Condor Python XQuery")
+        try:
+            itobj = schedd.xquery("WMAgent_AgentName == %s" % classad.quote(self.agent),
+                                  ['ClusterId', 'ProcId', 'JobStatus', 'MATCH_EXP_JOBGLIDEIN_CMSSite'])
+        except:
+            logging.debug("Query to condor schedd failed in SimpleCondorPlugin")
```
amaltaro: Don't report errors in `debug` level please. For this case, you could:
`logging.exception("Query to condor schedd failed in SimpleCondorPlugin: ")`

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+    def complete(self, jobs):
+        """
+        Do any completion work required
+
+        In this case, look for a returned logfile
+        """
+
+        for job in jobs:
+
+            if job.get('cache_dir') is None or job.get('retry_count') is None:
```
amaltaro: Perhaps we should separate this `if` into two, one for `cache_dir` and one for `retry_count` with the proper error message. I'd also print the job `id` and/or `gridid`, if possible all in one line.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+        Parameters:    excludeSite = False when moving to Normal
+                       excludeSite = True when moving to Down, Draining or Aborted
+        """
+        schedd = htcondor.Schedd()
+
+        jobtokill = []
+        try:
+            itobj = schedd.xquery("WMAgent_AgentName == %s" % classad.quote(self.agent),
+                                  ['ClusterId', 'ProcId', 'JobStatus', 'DESIRED_Sites', 'ExtDESIRED_Sites'])
+        except:
```
amaltaro: add a `logging.exception` here please. No exceptions should pass unnoticed.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+        else:
+            jobInfo = {}
+            for jobAd in itobj:
+                gridId = "%s.%s" % (jobAd['ClusterId'], jobAd['ProcId'])
+                jobInfo[gridId] = jobAd
+            for job in jobs:
+                jobAd = jobInfo.get(job['gridid'], None)
+                if jobAd:
+                    # only deal with Idle jobs
+                    if str(jobAd.get('JobStatus', 100)) == '1':
```
amaltaro: How about we remove this line and instead add a constraint like `JobStatus == 1` or so to the `xquery` command?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+                        if excludeSite:
+                            # Remove siteName from DESIRED_Sites if job has it
+                            if siteName in desiredSites:
+                                if len(desiredSites) > 1:
+                                    desiredSites.remove(siteName)
+                                    desiredSites = ','.join(desiredSites)
+                                    try:
+                                        schedd.edit([job['gridid']], 'DESIRED_Sites', classad.ExprTree('"%s"' % desiredSites))
+                                    except:
+                                        logging.debug("Failed to edit sites for job %s" % job['gridid'])
```
amaltaro: `logging.error` please

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+                                    jobtokill.append(job)
+                        else:
+                            # Add siteName to DESIRED_Sites if ExtDESIRED_Sites has it (moving back to Normal)
+                            if siteName not in desiredSites and siteName in extDesiredSites:
+                                desiredSites.append(siteName)
+                                desiredSites = ','.join(sorted(desiredSites))
+                                try:
+                                    schedd.edit([job['gridid']], 'DESIRED_Sites', classad.ExprTree('"%s"' % desiredSites))
+                                except:
+                                    logging.debug("Failed to edit sites for job %s" % job['gridid'])
```
amaltaro: `logging.error` please

amaltaro (commented):
It would be great to have some performance tests as well in order to compare it with PyCondor (job_id vs gridid handling).
Condor may timeout when it's under heavy load, SQL db usually doesn't.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+                                                      preAlert=preAlert)
+                    sendAlert(6, msg="Condor submission failed. Check condor status.")
+                    sender.unregister()
+                except:
+                    pass
+
+            else:
+
+                logging.debug("Finish: Submitting jobs using Condor Python SubmitMany")
+                for index,job in enumerate(jobsReady):
```
hufnagel: I am pretty sure, but @bbockelm might want to comment. It's the only way to do this anyways since submitMany doesn't return anything except the clusterId.

The assumption here is that submitMany will submit jobs with procId 0...N in an unbroken sequences in the same order the jobs are passed in the submitMany call.

bbockelm: Yes - proc ids are allocated contiguously, starting at 0.  `submitMany` will further only create a single job cluster.

This is a guaranteed semantic of the interface and not set to change in future versions.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: I can do that but these lines are quite useless as they are. If they ever trigger something seriously has gone wrong.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
bbockelm: Agreed.  We should have the schedd return the minimal rows possible.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+        # Customized classAds for this plugin
+        ad['DESIRED_Archs'] = "INTEL,X86_64"
+
+        ad['x509userproxy'] = self.x509userproxy
+        ad['x509userproxysubject'] = self.x509userproxysubject
+
+        ad['Rank'] = 0.0
+        ad['TransferIn'] = False
+
+        params_to_add = htcondor.param['SUBMIT_ATTRS'].split() + htcondor.param['SUBMIT_EXPRS'].split()
```
hufnagel: We have to add them all since when using condor_submit they are all added (ie. this is the current behavior).

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: Will do, but someone needs to test the new syntax later.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+        """
+        ad = classad.ClassAd()
+
+        #ad['universe'] = "vanilla"
+        ad['Requirements'] = classad.ExprTree(self.reqStr)
+        ad['ShouldTransferFiles'] = "YES"
+        ad['WhenToTransferOutput'] = "ON_EXIT"
+        ad['UserLogUseXML'] = True
+        ad['JobNotification'] = 0
+        ad['Core'] = 0
```
hufnagel: It was set peviously, so I just forwarded it. Looking through the HTCondor docs it doesn't seem to exist though, so removing it.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: At some point in the future this will be integrated so these are also set automatically if you submit jobs via python bindings.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: In light of that I am keeping the one check, but I am printing out the whole job object if we hit it.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
hufnagel: More interesting here is params_to_skip, parameters that the schedd will set for you but it doesn't really need to since we already set them ourselves. Bug, historic leftovers or why are they there ?

hufnagel: It's 5 lines of code, factoring it out would double it :-). I'll add the comment though...

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -0,0 +1,590 @@
...
+    def submit(self, jobs, info=None):
+        """
+        _submit_
+
+
+        Submit jobs for one subscription
+        """
+
+        # If we're here, then we have submitter components
+        self.scriptFile = self.config.JobSubmitter.submitScript
```
hufnagel: How do I get rid of outdated comments now ? This review is now orphaned after my last force-push, but it doesn't disappear like the old code comments did.

hufnagel:
The latest version addresses all the concerns from the review.

amaltaro:
Thanks, Dirk! Since it's a new file, can you please run the pep8 code reformat to get rid of all possible pylint complaints?

hufnagel:
Where is this documented ? pep8 is not available by default on our voboxes.

amaltaro:
If you have the wmagent unittest environment, then you have it, though I don't know the pep8 syntax for that, I just use it from the IDE (PyCharm in my case, learned from Eric :))

hufnagel:
If you want the developers to follow some syntax convention, check for that convention in the jenkins tests. If I get suggestions from the jenkins tests I will follow them. I am not going to do anything more myself though.

amaltaro:
Jenkins reports only a summary for that, otherwise it would be way way too much verbose and it would be impossible to browse GH PRs. E.g.
```
AlanWorkingBook:WMCore amaltaro$ alias pylt
alias pylt='pylint --rcfile=/Users/amaltaro/PycharmProjects/WMCore/standards/.pylintrc'
AlanWorkingBook:WMCore amaltaro$ pylt SimpleCondorPlugin.py 
Warning: option comment is obsolete and it is slated for removal in Pylint 1.6.
Warning: option ignore-iface-methods is obsolete and it is slated for removal in Pylint 1.6.
************* Module SimpleCondorPlugin
W:467, 0: TODO: remove when 8.5.7 is deployed (fixme)
C:114, 0: Line too long (172/160) (line-too-long)
C:162, 0: Exactly one space required after comma
                for index,job in enumerate(jobsReady):
                         ^ (bad-whitespace)
C:221, 0: Exactly one space required after comma
                (newStatus,location) = jobInfo[job['gridid']]
                          ^ (bad-whitespace)
C:240, 0: No space allowed after bracket
            if job['globalState'] in [ 'Complete', 'Error' ]:
                                     ^ (bad-whitespace)
C:240, 0: No space allowed before bracket
            if job['globalState'] in [ 'Complete', 'Error' ]:
                                                           ^ (bad-whitespace)
C:286, 0: No space allowed after bracket
                    for condorFile in [ condorOut, condorErr, condorLog ]:
                                      ^ (bad-whitespace)
C:286, 0: No space allowed before bracket
                    for condorFile in [ condorOut, condorErr, condorLog ]:
                                                                        ^ (bad-whitespace)
C:495, 0: Line too long (167/160) (line-too-long)
C:567, 0: Exactly one space required after comma
            classAds.append((ad,1))
                               ^ (bad-whitespace)
R: 30, 0: Too many instance attributes (15/7) (too-many-instance-attributes)
C:144,12: Invalid variable name "cluster_ad" (invalid-name)
C:145,12: Invalid variable name "proc_ads" (invalid-name)
W:147,12: Specify string format arguments as logging function parameters (logging-not-lazy)
W:150,19: Catching too general exception Exception (broad-except)
W:125,27: Unused argument 'info' (unused-argument)
W:195,15: Catching too general exception Exception (broad-except)
R:250, 4: Method could be a function (no-self-use)
W:325,15: Catching too general exception Exception (broad-except)
W:347,39: Catching too general exception Exception (broad-except)
W:348,36: Specify string format arguments as logging function parameters (logging-not-lazy)
R:322, 8: Too many nested blocks (7/5) (too-many-nested-blocks)
W:359,35: Catching too general exception Exception (broad-except)
W:360,32: Specify string format arguments as logging function parameters (logging-not-lazy)
R:365, 4: Method could be a function (no-self-use)
R:383, 4: Method could be a function (no-self-use)
W:419,19: Catching too general exception Exception (broad-except)
C:468, 8: Invalid variable name "params_to_add" (invalid-name)
C:469, 8: Invalid variable name "params_to_skip" (invalid-name)
```

though mine complains about "invalid-name", which I don't care because we usually use camel case naming. Maybe jenkins is not using the rc file under standards(?)

If you prefer I can reformat the code and send it to you. But please, have a look at it and setup your development environment, needless to say, this improves the quality of the code and makes the code more readable.

PerilousApricot:
@amaltaro You can have jenkins report it as an additional "check" that links to the output of pylint from the "details" link (instead of having jenkins update it as a comment)

bbockelm:
+1,000,000 to the above suggestion: the Jenkins bot causes so much noise and email that I tend to  not read tickets carefully.  I'd much rather it give either a X or a checkmark and a link to where I can read more.

The Jenkins bot makes DMWM tickets hard to read.

ericvaandering:
I'm always willing to entertain suggestions, but I guarantee you that no one is going to follow a link if we don't fail the PR. Should I fail every PR that doesn't fully follow PEP-8? You will come after me with pitchforks. 

We also can't actually run the pylint plugin for Jenkins. It caused some problem some time ago in the CMSSDT setup and Shahzad is loathe to re-enable it. 

Personally I click the little X on PR that I am responsible for to delete out of date comments, but maybe I am the only one that can do that.

bbockelm:
When I work with CVMFS, they absolutely reject PRs don't follow their code standards.  I suppose you could turn them into two separate checks and decide, when merging, whether you care if the second check fails.  In that case, the Github UI would say that one check failed and one succeeded.

To turn things around: why would you accept a PR that doesn't meet code standards?

hufnagel:
+1

Accept a PR and move on or don't accept it and tell me why and what I need to fix. I don't even mind if I need to fix a lot of stuff that has nothing to do with my changes if it improves the code and makes it more adherent to standards.

amaltaro:
I think these coding conventions are not being enforced because we cannot really isolate what comes from the new changes and what was already there. I personally would not like to add 10lines to a code and rework 1000 lines of reformatting (that sometimes cannot be completely automated).

In addition to that, we have a small enough group of WMCore contributors/developers that I was hoping we wouldn't hear "if you want me to do, then enforce it", knowing that those changes make the code better.
Just my 2c.

hufnagel:
@amaltaro on your last comment, we either push for code convention changes or we don't. This half-way of "we want to change things but it's not supposed to be any effort or hurt" is mostly just annoying :-). Maybe we can have a campaign at some point... Only way I see to make progress on this.

I wanted to check for what pep8 tells me, but it's not even shipped with the python version we build ourselves in COMP. That should be fixed.

And yes, pep8 is in the wmagent unittest environment, but that's heavy to setup to just check code conventions. For me to use this all the time the tools need to be available in the environment I develop in, not in a separate setup. I don't use an IDE, I use shell plus emacs. Always have.

PR #7083, opened 2016-08-23
Changes needed for "Automatic" splitting in CRAB

matz-e:
These are changes needed for "Automatic" splitting in CRAB. In detail:
1. Allow to specify the runtime limitation parameter `maxSecondsUntilRampdown` in the pset (`CMSSW_7_4_0` and later)
   - this allows us to run a splitting "probe" on the schedd that will return after a specific amount of time
   - allows to cap the runtime of regular jobs to have more of a "task completion time" guarantee (ASO excluded)
2. Adjust the dashboard API to accept non-integer job IDs
   - the "capped" regular jobs have completion jobs with IDs of the form `1-1`, `1-2`,… to complete unprocessed lumis from job `1` etc…
3. Allow to set a hard job limit for splitting, raising an exception when the total number of jobs exceeds this limit
   - the splitting probe runs the splitting algorithm in the the postjob on the schedd. When splitting parameters turn out to be unreasonable, creating more jobs than we ask for will bog down the schedd. Returning with an error state before this happens prevents the postjob from consuming excessive amounts of memory.

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -40,6 +40,7 @@ class EventAwareLumiBased(JobFactory):
         """
 
         avgEventsPerJob = int(kwargs.get('events_per_job', 5000))
+        jobLimit        = int(kwargs.get('job_limit', 0))
         eventLimit      = int(kwargs.get('max_events_per_lumi', 20000))
         totalEvents     = int(kwargs.get('total_events', 0))
         splitOnFile     = bool(kwargs.get('halt_job_on_file_boundaries', True))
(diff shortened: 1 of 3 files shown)


src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -40,6 +40,7 @@ def algorithm(self, *args, **kwargs):
         """
 
         avgEventsPerJob = int(kwargs.get('events_per_job', 5000))
+        jobLimit        = int(kwargs.get('job_limit', -1))
```
ericvaandering: Using -1 for this is kind of a C thing. Normally you'd use None although you want an int, so how about 0? Then the next bit just becomes ```if jobLimit and ...```

matz-e: I'll fix that. Been doing too much C stuff lately…

amaltaro:
Hi @matz-e , in case you still want to have this merged into WMCore, can you please rebase and squash the commits? Thanks

matz-e:
@amaltaro did that…

amaltaro:
@ticoann I think we can merge this one. I didn't test it, but code looks Ok and should be transparent for production. (branch PR was merged already)

PR #7084, opened 2016-08-23
Avoid using test in function names

amaltaro:
Fix this unittest failure

```
Simple_t.SimpleTest:test_auth_fail changed from success to error
```

that is popping up in #6094 

Googling a bit, the problem come from the fact that nose uses a regex to search for tests that needs to be executed. Some more info here:
http://stackoverflow.com/questions/16475980/nose-runs-test-on-function-in-setup-when-no-suite-specified

I hope it fixes the problem.

### DIFF
--- a/src/python/WMCore/REST/Test.py
+++ b/src/python/WMCore/REST/Test.py
@@ -63,7 +63,7 @@ def fake_authz_key_file(delete=True):
     t.seek(0)
     return t
 
-def setup_test_server(module_name, class_name, app_name = None, authz_key_file=None, port=8888):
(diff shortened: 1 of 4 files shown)


ericvaandering:
I dunno. I don't find that function listed here: https://cmssdt.cern.ch/jenkins/view/DMWM/job/DMWM-WMCore-UnitTests/lastCompletedBuild/testReport/

But a good idea anyhow.

amaltaro:
Wow, it seems it's hitting jenkins very very hard...

amaltaro:
BTW, this is the default regex used by nosetests `(?:\b|_)[Tt]est`
@ticoann feel free to merge this one (otherwise the MCFake patch will be bugged)

PR #7085, opened 2016-08-23
1.0.19 wmagent version of 

ticoann:
#7080   Do not upload skipped files for LogCollect/Cleanup to ACDC Server
#7076   Add only unique file/location rows into dbsbuffer
#6094   Make unique file entries for ACDC MCFakeFiles


ticoann:
sorry wrong pull request.

PR #7097, opened 2016-08-24
Use the correct priority var between JobSubmitter and RunJob

amaltaro:
To bear in mind that the job object built at JobSubmitter level is not exactly what we pass downstream do PyCondor plugin. Instead, we build a RunJob object from the "JobSubmitter job object".

Tested in testbed and it looks good.
@ticoann please review

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -329,13 +329,13 @@ class JobSubmitterPoller(BaseWorkerThread):
             if newJob['type'] in ["Repack", "Merge", "Cleanup", "LogCollect"]:
                 highIOjob = True
 
-            # Create a job dictionary object and put it in the cache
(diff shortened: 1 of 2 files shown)


ticoann:
HI Alan, I am not completely sure what this change would do.
Isn't finial priority is the combination of requestPriority and taskPriority?

There are a couple of more place expecting requestPriority.

https://github.com/amaltaro/WMCore/blob/6d79c234092a3af4568e24a1b9d51759758ce407/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py#L738

https://github.com/amaltaro/WMCore/blob/6d79c234092a3af4568e24a1b9d51759758ce407/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py#L740

amaltaro:
This update function was not touched in my previous patch, so that key is correct. Just x-checking:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobUpdater/JobUpdaterPoller.py#L148

sounds good indeed.

ticoann:
test this please

PR #7099, opened 2016-08-24
1.0.19 version of Use the correct priority var between JobSubmitter and RunJob

amaltaro:
1.0.19_wmagent version of #7097


ticoann:
patched in production agent - merging it.

PR #7100, opened 2016-08-24
Add support for jinja templates

vkuznet:


ericvaandering:
test this please

ericvaandering:
Jenkins could not reach GitHub

src/python/WMCore/ReqMgr/Web/tools.py
```diff
@@ -24,9 +24,18 @@
...
+    from Cheetah.Template import Template
+    CHEETAH=True
+except:
+    pass
+try:
+    import jinja2
+    JINJA = True
+    CHEETAH=False
+except:
+    JINJA = False
```
amaltaro: Two comments here:
1. the code will try to import both modules. I think it should try import jinja2 first and in case it fails, then import Cheetah (no need for import in the latter)
2. these globals don't look good. How about we check whether jinja was imported in the __init__ method and save it in an instance attribute, e.g.
```
self.jinja = True if 'jinja2' in sys.modules.keys() else False
```
and then properly use it in the other places.

vkuznet: > Two comments here:
> 1. the code will try to import both modules. I think it should try import jinja2 first and in case it fails, then import Cheetah (no need for import in the latter)

Alan, first we need to visually inspect all aspects of web pages generated by
templates. That's why I prefer that code should be able to work with both
templates engines. This will allow to perform comparison.

> 2. these globals don't look good. How about we check whether jinja was imported in the __init__ method and save it in an instance attribute, e.g.
> ```
> self.jinja = True if 'jinja2' in sys.modules.keys() else False
> ```
> and then properly use it in the other places.

this is a good solution indeed and I'll adjust the code.

ticoann:
To remind me, flip the flag and change spec to have a jinja dependency

PR #7116, opened 2016-08-26
Raise error if we try to access DBS directly in tests

ericvaandering:


ericvaandering:
New version of #6556

ericvaandering:
Need to figure out why LocalWorkQueueProfileTest seems not to be using mock.

ericvaandering:
Irrelevant now. May need to find a different way to do this

Issue #7075, opened 2016-08-18
Incomplete info for dbs block insertion (vocms026)


amaltaro:
It's the second time we have this issue with the relval agent. DBS3Upload is failing to update some blocks because they are either missing the acquisition_era or the start_date information.

Small part of the component log
```
2016-08-18 14:28:32,621:140404496717568:ERROR:DBSUploadPoller:Error trying to process block /RelValQCD_Pt_80_120_13_HI/CMSSW_8_1_0_pre8-81X_mcRun2_HeavyIon_v1-v1/GEN-SIM#312e69c2-42fe-11e6-836c-02163e016c
c2 through DBS.
HTTP Error 400: BlockInsert: acquisition_era_name and start_date are required.                                             NULL was received from user input. Please correct your data.
2016-08-18 14:28:32,622:140404496717568:ERROR:DBSUploadPoller:Traceback (most recent call last):
  File "/data/srv/wmagent/v1.0.17.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.0.17.patch1/lib/python2.7/site-packages/WMComponent/DBS3Buffer/DBSUploadPoller.py", line 116, in uploadWorker
    dbsApi.insertBulkBlock(blockDump = block)
  File "/data/srv/wmagent/v1.0.17.patch1/sw/slc6_amd64_gcc493/cms/dbs3-client/3.3.134/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 317, in insertBulkBlock
    return self.__callServer("bulkblocks", data=blockDump, callmethod='POST' )
  File "/data/srv/wmagent/v1.0.17.patch1/sw/slc6_amd64_gcc493/cms/dbs3-client/3.3.134/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 194, in __callServer
    self.__parseForException(http_error)
  File "/data/srv/wmagent/v1.0.17.patch1/sw/slc6_amd64_gcc493/cms/dbs3-client/3.3.134/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 221, in __parseForException
    raise HTTPError(http_error.url, data['exception'], data['message'], http_error.header, http_error.body)
HTTPError: HTTP Error 400: BlockInsert: acquisition_era_name and start_date are required.                                             NULL was received from user input. Please correct your data.
```

amaltaro:
Two more important details. There is no acquisition_era in the dbsbuffer_dataset table:
```
[...]

PR #7120, opened 2016-08-29
Port assignment validation code to ReqMgr2

amaltaro:
While investigating #7075, I stumbled upon a few flaws that are fixed here, they are:
- fixed a validation issue in Assign.py
- ported the output datasets lexicon validation to ReqMgr2
- ported the datatiers validation to ReqMgr2
- fixed the validation of assignment dict arguments in ReqMgr2
- BUT, did not fix the AcqEra issue as reported in that issue :-(

@ticoann please review but don't merge please, I still have to test it.

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -10,12 +10,13 @@ from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
 from WMCore.WMSpec.WMWorkloadTools import loadSpecByType, loadSpecClassByType, setArgumentsWithDefault
 from WMCore.REST.Auth import authz_match
 from WMCore.WMFactory import WMFactory
-
+from WMCore.Services.DBS.DBSReader import DBSReader
 from WMCore.ReqMgr.Auth import getWritePermission
 from WMCore.ReqMgr.DataStructs.Request import initialize_request_args
 from WMCore.ReqMgr.DataStructs.RequestStatus import check_allowed_transition
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidStateTransition, InvalidSpecParameterValue
 from WMCore.ReqMgr.Tools.cms import releases, architectures
+from WMCore.Lexicon import procdataset
 
 def loadRequestSchema(workload, requestSchema):
     """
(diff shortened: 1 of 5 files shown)


src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -186,3 +187,35 @@ def get_request_template_from_type(request_type, loc="WMSpec.StdSpecs"):
...
+    _validateDatatier(datatier, dbsUrl)
+
+def _validateDatatier(datatier, dbsUrl):
+    """
+    _validateDatatier_
+
+    Provided a list of datatiers extracted from the outputDatasets, checks
+    whether they all exist in DBS already.
+    """
+    dbsReader = DBSReader(dbsUrl)
```
ticoann: Maybe we should use cache (memory cache) to save Tier name since there . We probably don't want to make this validation fail when there is network or DBS glitch

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -186,3 +187,35 @@ def get_request_template_from_type(request_type, loc="WMSpec.StdSpecs"):
...
+    """
+    datatier = []
+    for dataset in outDsets:
+        tokens = dataset.split("/")
+        procds = tokens[2]
+        datatier.append(tokens[3])
+        try:
+            procdataset(procds)
+        except AssertionError as ex:
+            msg = "Bad output dataset name, check the processed dataset.\n %s" % str(ex)
```
ticoann: It seems there is a case procdataset returns False ("" or None)

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L199

Shoudn't we catch that case as well?

amaltaro: Thanks, Seangchan! I create a new issue for that: #7121
Where we can discuss details and refactor the code.

src/python/WMCore/ReqMgr/Utils/Validation.py
amaltaro: Good idea! Let me update this code

amaltaro:
@ticoann as we discussed over chat, I sent all my tests to a sick schedd :-(
Changes related to reqmgr2 looks ok, though all the jobs failed to be submitted on the agent side.

IMO, we can merge it and then we properly test and fix whatever problems arise during testbed validation.

PS.: I'll have a look at a proper caching this week

PR #7125, opened 2016-08-30
Cache list of datatiers from DBS

amaltaro:
Complement #7120, such that we only fetch the list of datatiers from DBS once every two hours.

@ticoann please review. I still need to test it.

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -17,16 +17,17 @@ from WMCore.Services.DBS.DBSErrors import DBSReaderError, formatEx3
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
 from WMCore.Services.SiteDB.SiteDB import SiteDBJSON as SiteDB
 
-def remapDBS3Keys(data, stringify = False, **others):
+
+def remapDBS3Keys(data, stringify=False, **others):
     """Fields have been renamed between DBS2 and 3, take fields from DBS3
     and map to DBS2 values
     """
-    mapping = {'num_file' : 'NumberOfFiles', 'num_files' : 'NumberOfFiles', 'num_event' : 'NumberOfEvents',
-               'num_block' : 'NumberOfBlocks', 'num_lumi' : 'NumberOfLumis',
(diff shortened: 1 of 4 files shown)


amaltaro:
test this please

amaltaro:
@ticoann it's good to go from my side. I tested it against both reqmgr and reqmgr2
I hope this caching goes along with what we were discussing yesterday :-)

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -240,15 +242,36 @@ def listDatasetFiles(self, datasetPath):
...
         _listDatatiers_
 
         Get a list of datatiers known by DBS.
         """
-        return [ tier['data_tier_name'] for tier in self.dbs.listDataTiers() ]
+        timenow = int(time.time())
+        if DBS3Reader._datatiers and timenow - 7200 < DBS3Reader._datatiers['ts']:
+            return DBS3Reader._datatiers['tiers']
+
+        DBS3Reader._setDatatiersCache(timenow, dbsUrl)
```
ticoann: may be we should try catch here, so if  cache updates fails return what is in the cache unless it is empty.
Also dbsUrl if mandatory paramater. it should check here not in _setDatatiersCache

amaltaro:
I applied your suggestions to the code. Let us see if Jenkins will be happy.

amaltaro:
@ticoann sounds good, but please have a final look

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -244,15 +246,42 @@ def listDatasetFiles(self, datasetPath):
         Get list of files for dataset
 
         """
-        return [ x['logical_file_name'] for x in self.dbs.listFileArray(dataset = datasetPath)]
+        return [x['logical_file_name'] for x in self.dbs.listFileArray(dataset = datasetPath)]
 
-    def listDatatiers(self):
+    @staticmethod
+    def listDatatiers(dbsUrl=None):
```
ticoann: since there dbsUrl is mandatory value it shouldn't have default value set. Anyway, this is minor. I will merge the patch

PR #7129, opened 2016-08-30
check parent file exist

### LINKED ISSUE #7128: dbsReader listFilesInBlockWithParents crash on RAW : dataset with no parents
Whilte trying to bring https://github.com/CMSCompOps/WmAgentScripts/blob/master/recover/recoverMissingLumis.py to have parent info

Calling 

blockFileParents = dbsReader.listFilesInBlockWithParents(blockName)

one gets 

/DoubleEG/Run2015D-v1/RAW#fceb85c4-5ce3-11e5-824b-001e67ac06a0

  File "recover/recoverMissingLumis.py", line 666, in <module>
    sys.exit(main())
  File "recover/recoverMissingLumis.py", line 663, in main
    val.group, val.dbsUrl, val.fake, datasetInformation)
  File "recover/recoverMissingLumis.py", line 410, in defineRequests
    blockBlacklist, blockWhitelist, dbsUrl, fakeLocation=fakeLocation)
  File "recover/recoverMissingLumis.py", line 246, in getFiles
    blockFileParents = dbsReader.listFilesInBlockWithParents(blockName)
  File "/data/srv/wmagent/v1.0.17.pre4/sw/slc6_amd64_gcc493/cms/wmagent/1.0.17.pre4/lib/python2.7/site-packages/WMCore/Services/DBS/DBS3Reader.py", line 528, in listFilesInBlockWithParents
    parentLumis = self._getLumiList(lfns = parentsLFNs)
  File "/data/srv/wmagent/v1.0.17.pre4/sw/slc6_amd64_gcc493/cms/wmagent/1.0.17.pre4/lib/python2.7/site-packages/WMCore/Services/DBS/DBS3Reader.py", line 82, in _getLumiList
    for lumisItem in lumiLists:
UnboundLocalError: local variable 'lumiLists' referenced before assignment

admittedly, the block has no parent, it might not be a reason to crash.

ticoann:
fixes #7128

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -74,6 +74,10 @@ class DBS3Reader:
                 lumiLists = []
                 for slfn in grouper(lfns, 50):
                     lumiLists.extend(self.dbs.listFileLumiArray(logical_file_name = slfn))
+            else:
+                # shouldn't call this with both blockName and lfns empty
+                # but still returns empty dict for that case
+                return {}
         except dbsClientException as ex:
             msg = "Error in "
             msg += "DBSReader.listFileLumiArray(%s)\n" % lfns
@@ -531,7 +535,14 @@ class DBS3Reader:
             # Probably a child can have more than 1 parent file
             for fp in f['parent_logical_file_name']:
                 childByParents[fp].append(f['logical_file_name'])
+        
         parentsLFNs = childByParents.keys()
+        
+        if len(parentsLFNs) == 0:
+            msg = "Error in "
+            msg += "DBSReader.listFilesInBlockWithParents(%s)\n There is no parents files" % (
+                fileBlockName)
+            raise DBSReaderError(msg)
 
         parentFilesDetail = []
         #TODO: slicing parentLFNs util DBS api is handling that.


ticoann:
Not sure whether it is right approach, if parent files doesn't exist, it will just return without that information. @amaltaro, Alan please review

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -74,6 +74,10 @@ def _getLumiList(self, blockName = None, lfns = None, validFileOnly = 1):
                 lumiLists = []
                 for slfn in grouper(lfns, 50):
                     lumiLists.extend(self.dbs.listFileLumiArray(logical_file_name = slfn))
+            else:
+                # shouldn't call this with both blockNam and lfns empty
```
amaltaro: Just a typo "blockName"

amaltaro:
How about we check the result from `listFileParents()` in L523 and raise an DBSError in case it's empty? I'm not sure we should return something in case parent files are missing, this can lead to much worse behavior in the end.
However, I also don't know if it's possible to have a dataset with parent but parent missing for a block or two ( @ericvaandering do you know?)

ticoann:
I don't think there is case if there is a parent dataset but missing parent blocks.

ticoann:
> How about we check the result from listFileParents() in L523 and raise an DBSError in case it's empty? I'm not sure we should return something in case parent files are missing, this can lead to much worse behavior in the end.

Yes, I think that would be safer since we have different function to only child data. I will update that.

ticoann:
test this please

ticoann:
test this please

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -531,7 +535,14 @@ def listFilesInBlockWithParents(self, fileBlockName, lumis = True, validFileOnly
             # Probably a child can have more than 1 parent file
             for fp in f['parent_logical_file_name']:
                 childByParents[fp].append(f['logical_file_name'])
+        
         parentsLFNs = childByParents.keys()
+        
+        if len(parentsLFNs) == 0:
+            msg = "Error in "
+            msg += "DBSReader.listFilesInBlockWithParents(%s)\n There is no patents files" % (
```
amaltaro: parents file, not patents - ewv

amaltaro:
Seangchan, I discussed this issue with Eric and his take on this is that we should not raise an exception if `parentsLFNs` is an empty list. So, we make the DBS service return whatever it knows and then we see if this would cause problems upstream (in the agent or so).

ericvaandering:
I'm going just by the name of the function: return file data *with* parents. If there are no parents, why should I be barred from calling this function?

ericvaandering:
test this please

ticoann:
> I'm going just by the name of the function: return file data with parents. If there are no parents, why should I be barred from calling this function?

Yes, I agree. The question is what would we do when caller pass the blocks without parent.
Precondition for this function is only passes the parameters with parent exists.
So who(what)ever calls this function has to makes sure it is passing blocks/etc with parent files exist.
I think in this case it is correct behavior to raise errors instead of find something break down the stream by accepting empty values from DBS.

vlimant:
I had not paid too much attention to the fix.
How is one supposed to use this eventually ?
How do I check if a block has a parent before passing it to listFilesInBlockWithParents or not ?

ticoann:
You just need to check this flag in the workflow spec.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L975

Even if there is parent blocks for input dataset it doesn't mean we process with parents. (only when the flag is set).

PR #7130, opened 2016-08-30
Properly set AcqEra for any task level - TaskChain

### LINKED ISSUE #6941: AcquisitionEra/ProcessingString for TaskChain Merge ACDC
it is mandatory right now to drop AcquisitionEra and ProcessingString in assigning merge acdc of taskchain wf, while it has to be passed for all other acdc.
We should either
    * not have to pass any of those for ACDC : which would make sense because recovery should recover on the original dataset
    * not make an exception for taskchain-merge-ACDC

bonus : make also the ProcessingVersion a useless parameter in assignment so that it cannot be set to anything different than the original wf version.

amaltaro:
Fixes #6941

Makes AcqEra parameter mandatory for assignment in reqmgr (it already is, I hope, in reqmgr2).
About the assign_optional validation, I could not find anything validating it in ReqMgr. Should we call this validation function being used by reqmgr2:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkloadTools.py#L241
?

For ACDCs of TaskChain Merge jobs, we sometimes don't have the Processing task in the workflow, meaning AcqEra key/value would not find that TaskName. I added a workaround such that Merge tasks will inherit the value from its parent according to one string contains another.

@ticoann I still have to test it, but feel free to make comments :)

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -1349,6 +1349,13 @@ class WMTaskHelper(TreeHelper):
         """
         if isinstance(era, dict):
             taskEra = era.get(self.name(), parentAcquisitionEra)
+            if taskEra is None:
+                # We cannot properly set AcqEra for ACDC of TaskChain Merge
+                # failures, so we should look up for a similar taskname in
+                # the acqera dict passed from the requestor
+                for taskname in era:
+                    if taskname in self.name():
+                        taskEra = era[taskname]
         else:
             taskEra = era
 


src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -1349,6 +1349,13 @@ def setAcquisitionEra(self, era, parentAcquisitionEra=None):
         """
         if isinstance(era, dict):
             taskEra = era.get(self.name(), parentAcquisitionEra)
+            if taskEra is None:
+                # We cannot properly set AcqEra for ACDC of TaskChain Merge
+                # failures, so we should look up for a similar taskname in
+                # the acqera dict passed from the requestor
+                for taskname in era:
```
ticoann: Can we look at the parent task for acquisition era?

amaltaro: Hmm, I had that in mind as well. But in this specific case, I don't think it has a parent task, since the Merge task would be the first Task in the ACDC request.

amaltaro:
@ticoann Seangchan, code is updated.
Please wait for Jenkins test results. I cannot really test it this evening because I would need to have TaskChain/merge failures in my VM first.

The fix looks pretty simple though, so I don't expect any problems.

PR #7149, opened 2016-09-04
remove absolute paths for xrdcp binaries

hufnagel:
They aren't needed and they make the FNAL stageout plugin unusable on some opportunistic sites.


amaltaro:
@ticoann please merge

PR #7150, opened 2016-09-04
remove event limits for merges

hufnagel:
Don't want to remove the code completely, but setting the default to 100M events is effectively doing the same thing.

### DIFF
--- a/test/python/WMComponent_t/JobCreator_t/JobCreator_t.py
+++ b/test/python/WMComponent_t/JobCreator_t/JobCreator_t.py
@@ -524,11 +524,11 @@ class JobCreatorTest(unittest.TestCase):
                       first_event = 1024, locations = set(["somese.cern.ch"]))
         fileII.addRun(Run(2, *[46]))
         fileII.create()
-        fileIII = File(lfn = "fileIII", size = 1024, events = 102400,
+        fileIII = File(lfn = "fileIII", size = 1024, events = 1024,
                        first_event = 2048, locations = set(["somese.cern.ch"]))
         fileIII.addRun(Run(2, *[46]))
         fileIII.create()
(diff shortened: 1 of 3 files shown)


hufnagel:
test this please

hufnagel:
Interesting, seems the parameters passed to the job splitter in the

WMComponent_t.JobCreator_t.JobCreator_t.JobCreatorTest:testE_TestNonProxySplitting

unit test actually were never successfully passed to the job splitter. This unit test "accidentally" worked and now that I changed the defaults it doesn't anymore.

hufnagel:
Ok, this works now. Ready for review @amaltaro

test/python/WMComponent_t/JobCreator_t/JobCreator_t.py
```diff
@@ -538,6 +538,9 @@ def stuffWMBS(self, workflowURL, name):
             mergeFileset.addFile(file)
             bogusFileset.addFile(file)
 
+#        mergeFileset.markOpen(False)
+#        bogusFileset.markOpen(False)
```
amaltaro: Dirk, can you remove these lines please?

amaltaro:
The change looks completely safe, however, do you need to have it in testbed this week and production in the next one? If not, then I'd rather merge it only after the production upgrade to avoid any possible side effect in the validation ...

hufnagel:
There is no hurry with this. Reason I am making the change is since the discussion about event limits came up for the Tier0 and the statement was "we do not need a limit for events per file". We already implemented this for the Tier0, I just want to change the defaults for WMAgent merges to be consistent (eventually).

amaltaro:
Ok.
@ticoann from my side, I'd merge it only after the production upgrade is done (next Tuesday)

amaltaro:
test this please

Issue #7025, opened 2016-08-04
problem with workflow sandbox tarball being zero size


hufnagel:
We observed a problem in the Tier0 with the workflow sandbox tarball being zero size. We looked at both the Tier0Feeder logs (since we call wmbsHelper.createSubscription() there) and the JobCreator logs and didn't see anything out of the ordinary.

Any idea what could have happened here ?

ticoann:
which version of agent are you using? Is this first  time observing the problem?

hufnagel:
Tier0 production deployment, so WMCore version is 1.0.17.pre2. I think this is the first time we have seen this, but @ebohorqu might want to clarify.

amaltaro:
Perhaps you should start considering to follow the same path we do for central production, which is only using "production ready" releases. Cycle 1.0.17 production release would be at least *.pre6, that would certainly save you from some headaches.

hufnagel:
You really think that is the problem here ?! Workflow sandbox creation code hasn't been touched in ages AFAIK.

amaltaro:
Nope, I just made a general comment that sooner or later will hit T0 :)

amaltaro:
BTW, is it a very recent workflow? Or it was created more or less in parallel with the disk space issues the T0 agent was facing a week or so ago?

hufnagel:
Could be related to the disk space issues, in fact, that's the most likely explaination. Still a problem though since it means the workflow sandbox creation ignores an error in creating the tarball.

amaltaro:
Fixed by #7151

PR #7151, opened 2016-09-05
change sandbox tar creation code, fixing #7025

hufnagel:
Two reasons. First the code is old-style and uses some syntax that is deprecated in python2.7.

Second there were some problems with creating zero size sandboxes, which should never happen. New code should be more efficient and I also added some logging.

### DIFF
--- a/src/python/WMCore/WMRuntime/SandboxCreator.py
+++ b/src/python/WMCore/WMRuntime/SandboxCreator.py
@@ -6,12 +6,12 @@
 """
 
 import os
-import re
 import shutil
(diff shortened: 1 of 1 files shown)


amaltaro:
Looks good. Did you manage to test it?

hufnagel:
Yep, ran a Tier0 replay. Worked correctly, sandbox was created and I got the log output.

amaltaro:
I also ran all the DMWM and Integration workflows and no problems were noticed.
@ticoann please merge

PR #7159, opened 2016-09-06
Fix assignment-only with reqmgr2 client

amaltaro:
Still checking whether it does not break other options.


amaltaro:
I tested request creation+assignment and assignment-only and it all works Ok.
@ticoann please merge

PR #7162, opened 2016-09-06
Set ProcStr and ProcVer for any task level - TaskChain

amaltaro:
Should have been made along #7130

Sorry @ticoann I was not expecting to hit problems here as well. I'll try to test it in my VM today.
If I don't manage, then I can cut a new tag tomorrow early in the morning, just send me an email with the PRs we need to merge

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -1277,6 +1277,10 @@ class WMTaskHelper(TreeHelper):
         """
         if isinstance(procVer, dict):
             taskProcVer = procVer.get(self.name(), parentProcessingVersion)
+            if taskProcVer is None:
+                for taskname in procVer:
+                    if taskname in self.name():
+                        taskProcVer = procVer[taskname]
         else:
             taskProcVer = procVer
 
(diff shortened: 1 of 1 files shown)


amaltaro:
Darn it! I was running the wrong TaskChain workflow in my VM.
As we discussed, if you prefer you can leave this one and I'll finish the validation tomorrow morning and cut a new tag. Otherwise, you can merge and cut a new tag and I'll x-check before Bruno deploys it into testbed. Thanks, Seangchan!

ticoann:
Thanks  Alan, It seems the patch works fine.

Here is original workflow.

https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/sryu_ACDC_TaskChain_RelVal_Multicore_reqmgr2_160907_050639_3367

Here is ACDC created from merge failure.

https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/sryu_ACDC_TaskChain_RelVal_Multicore_reqmgr2_160907_050639_3367

amaltaro:
Yes, just for completeness. I tested the following cases:
a) AcqEra and ProcStr as dict, different values
b) AcqEra and ProcStr as string (though they shouldn't use this case)
c) no key/value for AcqEra and ProcStr provided at all.
they all worked ok.

PR #7165, opened 2016-09-07
Double destroy log

ericvaandering:
Look at Alex's change with the cleanup logging enabled.


amaltaro:
Wow, this one must be a real total destruction :-P

PR #7166, opened 2016-09-07
Additional condor classAd for HLT usage

amaltaro:
One of the condor tweaks for better usage of the HLT resources. Farrukh knows more details, but what we discussed is that it can be applied upstream to all the agents and it will only make difference when jobs land on HLT worker nodes. So it has no side effects.

We have this patch already in place for the HLT agent (vocms0116).

@Farrukh-Aftab, do I get it right that before we starting running production in HLT, all the agents have to have this patch?

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -871,6 +871,9 @@ class PyCondorPlugin(BasePlugin):
             jdl.append("+WMAgent_JobID = %s\n" % job['jobid'])
             jdl.append("job_machine_attrs = GLIDEIN_CMSSite\n")
 
+            # As discussed with Farrukh on Sept 6, this knob is needed for HLT usage
+            jdl.append("+JobLeaseDuration = isUndefined(MachineAttrMaxHibernateTime0) ? 1200 : MachineAttrMaxHibernateTime0\n")
+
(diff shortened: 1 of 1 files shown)


Farrukh-Aftab:
You can run at HLT from all the agents right now too. What this classAd does is support the 24 hour hibernation of VMs.

What we are essentially doing here is asking (to put it at a very higher level) the glidein/VM/pilot for its hibernation time. If the hibernation time is being set by the glidein/VM/pilot, we use it. Otherwise we default to 20 mins.

On the HLT side this is being set to 24 hours. So a job will stay in *running* state on the schedd waiting for the VM to come back up for 24 hours. VMs are going to suspend their state to disk and then restore it allowing jobs to continue from where they had left off.

ticoann:
@Farrukh-Aftab, Hi Farrukh, So MachineAttrMaxHibernateTime0 is set by glidein/vm/pilot? So what JDL specifies is merely turn on the function? 

@amaltaro, Alan, isn't the agent kill the job if it stays too long in running status?

Farrukh-Aftab:
@ticoann , yep! We add a configuration knob to the schedd (SYSTEM_MACHINE_ATTR) listing the attributes we want fetched from the glidein/vm/pilot and that information is then populated into MatchineAttr* classAds for the jobs once they start using that glidein/vm/pilot.

amaltaro:
@ticoann please merge

PR #7168, opened 2016-09-07
support use of xrdcp-old

hufnagel:


hufnagel:
Will be needed if we ever want to use this for FNAL (at least for right now).

amaltaro:
Dirk, can you give us more details please. Why FNAL keeps this xrdcp-old and why not the main binary? Are we using this right now? If you are adding this just in case we ever want to use it for FNAL in a far future, then I'm in favor of keeping this out and instead we make sure the correct setup is working when the time comes.

hufnagel:
FNAL is using it now, their storage (the dCache part) doesn't work with xrdcp for some reason. Which means any opportunistic resource that wants to stage out to FNAL also had to use xrdcp-old.

Anyways, this is an option. If FNAL fixed their storage we can stop using it. For now we need it.

hufnagel:
And yes, I want to use the generic xrdcp plugin for opportunistic resources that stageout to FNAL and if that works well I am going to push FNAL to change over to the generic xrdcp plugin as well.

hufnagel:
Eventual plan is to retire all stageout plugin except for generic xrdcp and generic gfal2. Lets see if we can make it happen.

amaltaro:
@ticoann please merge

amaltaro:
Well, since I had to push other patches to the production agents. I've just pushed this one as well.
I'm creating the wmagent branch request as well.

PR #7175, opened 2016-09-12
Condor_chirp stageout information

### LINKED ISSUE #6855: condor_chirp/crab3 and add stageout reports
Hey,
From crab jobs I see interesting numbers for stageout related failures (local, fallback, direct), but would be better to keep the track of each stageout separate and see the percentage also for prod and crab together,separate.  which now we can't filter easily and it is not possible.

So I have few questions:
Can we move https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Executor.py#L164-L184 to better place, something like src/python/WMCore/Services/Condor/Condor.py ? Maybe in the future, crab would use the same python function to do chirp for other things or prod.
With one of previous pull requests Stageout module already prepares a stageoutPolicyReport with all tried stageouts and can we do condor_chirp, something like this?

```
for itemNum in range(len(stageoutPolicyReport)):
    stageOutRow = stageoutPolicyReport[itemNum]
    for reportType in ['PNN', 'Command', 'ExitCode']:
        if reportType in stageOutRow and stageOutRow[reportType] is not None:
            key = Chirp_Stageout_%s_%s_%s" % (stageoutType, itemNum, reportType)
            self.setCondorChirpAttrDelayed(key, stageOutRow[reportType])
```

@bbockelm is there any penalty for a job on schedd or anything else of doing more condor_chirp commands?

juztas:
Fixes #6855 and I did not wanted to use `setCondorChirpAttrDelayed` method from Execute class. I think it should be outside. All of this data is already reported to dashboard and we asked them to plot it, but... So lets do condor chirp and we will have it in cms-gwmsmon & Brians monitoring page.

### DIFF
--- a/setup_dependencies.py
+++ b/setup_dependencies.py
@@ -4,48 +4,48 @@ Manage dependancies by declaring systems here.
 A system can depend on packages or other systems.
 If a package ends with a + include all subpackages.
 """
-dependencies = {'wmc-rest':{
-                        'bin': ['wmc-dist-patch', 'wmc-dist-unpatch','wmc-httpd'],
-                        'packages' : ['WMCore.REST'],
+dependencies = {'wmc-rest': {
+                        'bin': ['wmc-dist-patch', 'wmc-dist-unpatch', 'wmc-httpd'],
+                        'packages': ['WMCore.REST'],
                         'modules': ['WMCore.Configuration'],
-                        'systems':['wmc-base']
+                        'systems': ['wmc-base']
                         },
-                'wmc-base':{
+                'wmc-base': {
                         'bin': ['wmc-dist-patch', 'wmc-dist-unpatch'],
-                        'packages' : ['Utils', 'WMCore.DataStructs'],
+                        'packages': ['Utils', 'WMCore.DataStructs'],
                         'modules': ['WMCore.WMFactory', 'WMCore.WMException', 'WMCore.Configuration',
                                     'WMCore.WMExceptions', 'WMCore.WMFactory', 'WMCore.Lexicon',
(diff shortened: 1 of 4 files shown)


amaltaro:
I don't know where, but these changes certainly don't belong to the DashboardAPI module.
Maybe we should create another Condor module under Services? Or maybe add these under StageOut executor?

The same would apply to other functions that are already in the DashboardAPI, but that's another story.

amaltaro:
BTW, any chances to have a unit test for it?

juztas:
I would create another Condor module under Services as we might want to add more in the future, so it will not be only for Stageout. Also currently it reports also for cmsRun exitCode. Let me know what you decide.  Also `parseAd` I would move from DashboardAPI.
For unit test, not sure what to test. condor_chirp is not available as we are removing it during rpm build, so it is available on WN only

juztas:
@amaltaro

amaltaro:
> I would create another Condor module under Services as we might want to add more in the future

I agree. Can you update this patch already in the new module please?

juztas:
@amaltaro just to be sure, here: https://github.com/dmwm/WMCore/blob/master/setup_dependencies.py#L32 I need to add WMCore.Services.Condor+ ?

amaltaro:
Yes, I think that's the place.

juztas:
test this please

juztas:
test this please

amaltaro:
@juztas can you update this PR please? I'm not sure it will have again conflicts with https://github.com/dmwm/WMCore/pull/7660 :_

amaltaro:
@juztas Justas, we appreciate your contribution, but given that this PR has been inactive for around 7 years now, we are closing this out. Thanks again!

PR #7186, opened 2016-09-15
close subscription even if there is available element in GQ in Drain

ticoann:


ticoann:
Applied in sumbit1 and seems to be working.

PR #7187, opened 2016-09-15
close subscription even if there is available element in GQ in Draining agent

ticoann:
This allows logCollect and cleanup job to be created when there is GQ elements in available state


ticoann:
updated production agents.

PR #7188, opened 2016-09-15
upload wmbs info to wmstats

ticoann:

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -545,7 +545,6 @@ config.AgentStatusWatcher.forceSiteDown = [] # List of sites to be forced to Dow
 config.AgentStatusWatcher.onlySSB = True # Set thresholds for sites only in SSB (Force all other to zero/down)
 config.AgentStatusWatcher.enabled = True # switch to enable or not this component
 config.AgentStatusWatcher.agentPollInterval = 300
-config.AgentStatusWatcher.monitorPollInterval = 3  # number of component polling cycles
 config.AgentStatusWatcher.defaultAgentsNumByTeam = 5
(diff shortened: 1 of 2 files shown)


amaltaro:
Thanks, Seangchan! And how do we retrieve this info from wmstats? I will check the final data structure uploaded tomorrow.

amaltaro:
BTW, don't worry with testing it. I'll have a close look at it tomorrow

ticoann:
Alan it seems it works fine.
Data is uploaded and more importantly nothing breaks. 😄 

https://reqmgr2-dev.cern.ch/couchdb/wmstats/vocms008.cern.ch%3A9999
https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?wmstats/vocms008.cern.ch%3A9999

@amaltaro, let me know if you think it is good to merge.

amaltaro:
Well, you're da man Seangchan!

Just one suggestion, I think we should keep that `if` because running all those queries in a heavy loaded agent can take some time (and adds even more load), that's why I instrumented it to collect `WMBS_INFO` only once every 3 component cycles.

amaltaro:
I didn't really time that, but the aggregation wmbs_job queries may take several secs AFAIR (<30secs). To be on the safe side, I think we can get that info once every 15min as it was before.

ticoann:
As we discussed I added
1. query time. (in sec)
2. remove the monitoring config
3. add json file back
4. change log to debug.

Test seems fine. Just query seems to take less then 1 sec. 
https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?wmstats/vocms008.cern.ch%3A9999

amaltaro:
Thank you very much, Seangchan! Changes look good.

amaltaro:
> Alan, how long does the query takes on high load?

BTW, just got some numbers for the CERN nodes and it's 3secs in avg, though their load is quite small.

PR #7191, opened 2016-09-16
Split by lumi (by work) with event awareness

ericvaandering:
So this now works including the unit tests which come from EventAwareLumiBased.

I changed some numbers because I think I'm actually doing something a little different in that I'm trying to get close the request # of events/job, not stay under it. In other words if I have a 100 event lumi and the job already has 420 events and the target is 500 events, it will add it since 520 is closer to 500 than 400 is.

Otherwise this mostly behaves the same except the correction for split lumis is built in since it constructs jobs by lumi instead of file. The `algorithm()` is about half the size of the EventAwareLumiBased since I'm using LumiList objects and did some other refactoring.

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
@@ -1,0 +1,333 @@
+#!/usr/bin/env python
+"""
+_EventAwareLumiByWork_
+
+Lumi based splitting algorithm that will chop a fileset into
+a set of jobs based on lumi sections, failing jobs with too many
+events in a lumi and adapting the number of lumis per job
+according to the average number of events per lumi in the files.
+
+This is adapted from EventAwareByLumi but does its work on lumi sections directly, not files.
+No correction for lumis split across files is needed, it's automatic. Makes use of the LumiList
+class to simplify the code
+
+"""
+
+from __future__ import (division, print_function)
+
+import logging
+import traceback
+from collections import defaultdict
+
+from WMCore.ACDC.DataCollectionService import DataCollectionService
+from WMCore.DataStructs.LumiList import LumiList
+from WMCore.DataStructs.Run import Run
+from WMCore.JobSplitting.JobFactory import JobFactory
+from WMCore.WMBS.File import File
+
+
+class EventAwareLumiByWork(JobFactory):
+    """
+    Split jobs by lumis taking into account events per lumi
+    """
+
+    locations = []
+
+    def __init__(self, package='WMCore.DataStructs', subscription=None, generators=None, limit=0):
+        super(EventAwareLumiByWork, self).__init__(package, subscription, generators, limit)
+
+        self.loadRunLumi = None  # Placeholder for DAO factory if needed
+        self.perfParameters = {}
+        self.deterministicPU = False
+        self.maxLumis = 1  # Maximum lumis seen in a job (needed for deterministic pileup only)
+        self.maxEvents = 1  # Maximum events seen in a lumi (needed for deterministic pileup only)
+
+        # Job accumulators
+        self.eventsInJob = 0
+        self.jobLumis = []
+        self.jobFiles = set()
+        self.eventsInLumi = 0
+
+        # Location accumulators. Using these two objects lets us automatically fix up the cases where a lumi
+        # is split across two files
+        self.lumisProcessed = set()  # Which run/lumi pairs already have jobs created for this location
+        self.filesByLumi = {}  # Map by lumi of which files contain that lumi
+
+    def algorithm(self, *args, **kwargs):
+        """
+        _algorithm_
+
+        Split files into a number of lumis per job
+        Allow a flag to determine if we split files between jobs
+        """
+
+        # TODO: This has one possible weakness. If two files (blocks) contain the same run/lumi section but are
+        # TODO: split across sites two different jobs will read the same lumi
+
+        avgEventsPerJob = int(kwargs.get('events_per_job', 5000))
+        lumiEventLimit = int(kwargs.get('max_events_per_lumi', 20000))
+        totalEventLimit = int(kwargs.get('total_events', 0))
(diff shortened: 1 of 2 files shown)


ericvaandering:
@ticoann @amaltaro @hufnagel am interested in your thoughts.

@mmascher you might be interested too.

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
retest this please

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
```diff
@@ -0,0 +1,276 @@
...
+        if collectionName:
+            lumiMask = self.lumiMaskFromACDC(couchURL=kwargs.get('couchURL'), couchDB=kwargs.get('couchDB'),
+                                             filesetName=kwargs.get('filesetName'), collectionName=collectionName,
+                                             owner=kwargs.get('owner'), group=kwargs.get('group'))
+
+        lDict = self.sortByLocation()
+
+        # First we need to load the data
+        if self.package == 'WMCore.WMBS':
+            loadRunLumi = self.daoFactory(classname="Files.GetBulkRunLumi")
```
hufnagel: This doesn't seem to be used anywhere...

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
```diff
@@ -0,0 +1,276 @@
...
+                    for lumiNum in runLumi.lumis:
+                        lumi = str(lumiNum)
+                        filesByLumi[run][lumi].append(fileObj)
+                        eventsByLumi[run][lumi] = round(eventsInFile / lumisInFile)
+
+            # Use fast construction to build a LumiList for the FileSet and apply the mask determined earlier
+            lumisByLocation[locationSet] = LumiList(runsAndLumis=runsAndLumis)
+            if lumiMask:
+                lumisByLocation[locationSet] &= lumiMask
+            lumisToProcess += lumisByLocation[locationSet]
```
hufnagel: Wouldn't it be better to apply the mask earlier, before you build up the full filesByLumi and eventsByLumi dictionaries. If you do sparse processing of large filesets this can make quite a difference...

hufnagel:
@ericvaandering , so this basically resorts your input so the splitting algorithm operates on lumis in the input instead of files ? Not sure how useful this is, since the problem is that we want jobs to sequentially read from input files. We do not reorder lumis.

For example, how will this work in the presence of completely unordered input ?

File A : lumi 1,3,5,7,9
File B : lumi 2,4,6,8,10

Each lumi is 1000 events, splitting is set to 2000 events. What kind of jobs is this going to create ?

5 jobs each reading one lumi from file A and another from file B ?

hufnagel:
Btw, since our standard WMAgent merge logic doesn't care about lumi order and merges whatever comes first in whatever order it finds, this example case is actually more the norm than the exception.

ericvaandering:
Thanks for looking at it Dirk.

In the pathological case you outline, yeah, that's what it would do. Basically when I looked at this I had three goals:

1) Operate on work not files
2) Get rid of one-off code mucking around with lumi lists and move to DataStructs/LumiList
3) Get rid of the post-hoc correction for lumis split across files and build that functionality in from the start.

Ordering was a by product and I think I could rewrite this without much trouble to preserve file order without much trouble.

Re: your 2nd comment, it seems to me (I have not looked at the details yet) that to reduce the # of ACDCs and help with recovery workflows, it would be useful if each step, including merges, could have it's own lumi white list. That way if one output stream is missing set A and another is missing set B, you can re-run the workflow with the union of the two and then automatically filter out what you don't need.

hufnagel:
My problem is that this isn't a pathological case, this is the normal case. If you have workflows that run at multiple sites and job assignment is random and then merges happen whenever jobs are done and only at sites, you end up with highly discontinuous assignments of lumis to files.

hufnagel:
This job splitter will work ok for workflows that split at single lumi level. At the multiple lumi level the practical effect will be that you end up with jobs that open/read multiple input files and only use a fraction of each file.

hufnagel:
You basically will increase the number of file opens across the whole system by a factor that is almost the same as the number of lumis a job is processing on average.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
ericvaandering: What do you mean? That the code I have doesn't do anything with it or that this is actually not needed because we don't use this mode of operations? I didn't investigate it, this is just a snippet of code left over from EventAwareByLumi.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
ericvaandering: I can look into that. Good point.

ericvaandering:
test this please

ericvaandering:
@hufnagel care to take another look? I refactored this a bit to work on files first, addressing your concerns. It's functional as a job splitter, but not on the ACDC stuff yet.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
```diff
@@ -0,0 +1,289 @@
...
+            lumisProcessed = set()
+
+            filesByLumi, lumisByFile, eventsByLumi = self.fileLumiMaps(filesAtLocation=filesAtLocation, getParents=getParents, lumiMask=lumiMask)
+            for f in filesAtLocation:
+                lfn = f['lfn']
+                if lfn not in lumisByFile:
+                    continue  # There are no lumis of interest in the file
+                for run, lumi in lumisByFile[lfn].getLumis():
+                    if (run, lumi) in lumisProcessed:
+                        continue  # We already saw this lumi and it got included in an earlier job
```
hufnagel: Is this really needed ? You populate this in the same loop, so this skips if you iterate over the same element twice ? Isn't that impossible ?

PR #7195, opened 2016-09-19
fix typo in sandbox creation code

hufnagel:
Added the PSetTweaks directory twice and didn't added the Utils directory at all. Fixed it.


amaltaro:
I wonder how it didn't break all the tests we ran in testbed ... anyways.

hufnagel:
Yeah, was surprised that worked at all :-). Really only seemed to have affected LogCollect.

ticoann:
It broke new agent test from last week. :smile: I think it is only added to the new version.
I was going to look at that. Thank you very much for the fix, Dirk.

PR #7207, opened 2016-09-21
add underlining logDB call from component

ticoann:
@amaltaro Alan, please review.

Usuage is like following.

get logDB instance anywhere component thread using following function.
getLogDBInstanceFromThread()

start logging the message.  If it is error, we need to make sure to delete it in next cycle if the error is gone

### DIFF
--- a/src/python/WMCore/WorkerThreads/BaseWorkerThread.py
+++ b/src/python/WMCore/WorkerThreads/BaseWorkerThread.py
@@ -142,6 +142,15 @@ class BaseWorkerThread:
             # to get the right name
             myThread = threading.currentThread()
 
+            if hasattr(self.component.config, "General") and \
+               hasattr(self.component.config.General, "central_logdb_url") and \
+               hasattr(self.component.config, "Agent"):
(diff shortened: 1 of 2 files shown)


ticoann:
at least this didn't brake anything on the test in vm. Merging it for further test.

src/python/WMCore/WorkerThreads/BaseWorkerThread.py
```diff
@@ -142,6 +142,15 @@ def __call__(self, parameters):
             # to get the right name
             myThread = threading.currentThread()
 
+            if hasattr(self.component.config, "General") and \
+               hasattr(self.component.config.General, "central_logdb_url") and \
+               hasattr(self.component.config, "Agent"):
+                from WMCore.Services.LogDB.LogDB import LogDB
```
amaltaro: I still have to digest the usage of this logDB :) but my first comment would be about the location of this import. Can't we import it from the top of this module?

ticoann: I am not sure why I put import here. :-) I will move to the top. Thanks,

PR #7208, opened 2016-09-21
add skipped file view

ticoann:
@amaltaro, Alan please review, still need to add wmstats side of change. which checked which contains whether workflow contains skipped file or not.


ticoann:
Tested and data is updated correctly. Will make the separate patch for wmstats part change. since it cause the conflict. Also this can be patched separately in wmagent.

ticoann:
Created separate commit for wmstats part and agent part. 1.0.19 agent version need to be created only agent part

ticoann:
Actually, Dirk told me that we can't get the event and lumis list from fwrj for skipped files.
I would need to add new views on acdc and get it from there but this could be expensive to get for all the request. I will create the link in wmstats so it only gets when it is needed.

PR #7210, opened 2016-09-22
Remove condor log parsing from PyCondor

amaltaro:
As discussed with @ticoann and @hufnagel some time ago. Parsing the condor logs is quite inefficient and it finds information for 1 out of 10 jobs, at the best. Thus I'm removing that piece of code.

If a job vanishes from condor, we mark it as `Removed`. If it remains in `Removed` status for 2min, then we mark it as `Completed` and let JobAccountant check what happened.

I've also removed the alert system and updated some logging level.
I still have to test it.

### DIFF
--- a/src/python/WMCore/BossAir/BossAirAPI.py
+++ b/src/python/WMCore/BossAir/BossAirAPI.py
@@ -509,7 +509,7 @@ class BossAirAPI(WMConnectionBase):
                 jobsToReturn.extend(localRunning)
                 jobsToChange.extend(localChanges)
                 jobsToComplete.extend(localCompletes)
-                logging.info("Running/changing/completing %i/%i/%i jobs in plugin %s.\n", len(localRunning), len(localChanges), len(localCompletes), plugin)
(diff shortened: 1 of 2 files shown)


amaltaro:
test this please

amaltaro:
test this please

bbockelm:
Copying from the other ticket - it's almost impossible for the condor log parsing to function as written.

So it is probably indeed a good idea to remove this.

amaltaro:
BTW, any thoughts on how to handle cases where condor is not responsive and it returns no classAds at all? Can we assume that if 0 classAds are returned for 30min consecutive, that all jobs are gone from the condor pool?

bbockelm:
`Schedd.xquery` should always throw an exception on failure (a bit different from the `CondorPlugin`, as `condor_q` has historically had issues returning 1 on failure).

amaltaro:
@ticoann I ran some tests with this patch and it looks good. I suggest we merge it now, it will have to be tested for the new agent again anyways.

amaltaro:
Actually, I'll just fix a couple of loggings.

amaltaro:
@ticoann this one is good to go. Please merge and we'll further test it in the next agent release. BTW, it only affects agent side.

PR #7213, opened 2016-09-23
Log reqmgr2 assignment parameters

amaltaro:
Should be useful for debugging mysterious reqmgr2 request issues.


amaltaro:
@ticoann good to be merged

ticoann:
Alan, What was the error?

amaltaro:
One of the cases was with the Trust flag being used during creation and NOT during assignment.
I'm certain there will be other problems where these parameters (from the logs, not from what operator says) will be useful.

PR #7218, opened 2016-09-23
fix site information in wmstats

### LINKED ISSUE #7217: Data location inconsistency between job logs and wmstats
@vlimant was asking me about a very weird case where wmstats reports that a job was running at T2_US_Nebraska (failed job) while the condor logs in the agent vocms0303 reports that this job ran at T1_US_FNAL (notice it's the first attemp, so retry 0).

I forgot to add a link to a wmstats screenshot and a tarball of this job, from the JobCreator:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/Issue_7217/

@ticoann do you have any idea of what can have happened?

ticoann:
fixes #7217

### DIFF
--- a/src/python/WMCore/JobStateMachine/ChangeState.py
+++ b/src/python/WMCore/JobStateMachine/ChangeState.py
@@ -318,7 +318,10 @@ class ChangeState(WMObject, WMConnectionBase):
                     jobState = "jobfailed"
                 else:
                     jobState = newstate
-                    
+                
+                # there is race condition updating couch record location and job is completed.
+                # for the fast fail job, it could miss the location update
+                job["location"] = job["fwjr"].getSiteName() or job.get("location", "Unknown")     
                 # complete fwjr document
                 job["fwjr"].setTaskName(job["task"])
                 jsonFWJR = job["fwjr"].__to_json__(None)


ticoann:
@amaltaro, Alan, please review. (I still need to test) but it looks OK, I will create 1.0.19_wmagent branch version

src/python/WMCore/JobStateMachine/ChangeState.py
```diff
@@ -318,7 +318,10 @@ def recordInCouch(self, jobs, newstate, oldstate, updatesummary = False):
                     jobState = "jobfailed"
                 else:
                     jobState = newstate
-                    
+                
+                # there is race condition updating couch record location and job is completed0
+                # for the fast fail job, it could miss the location update
+                job["location"] = job["fwjr"].getSiteName() or job.get("location", "UnKnown")
```
amaltaro: So, if there is a site in the fwjr, it gets assigned. Otherwise it gets the original location for the job. Is that right? Can this job have multiple locations?

I think it's better to avoid these capital letters, so `UnKnown` to `Unknown`. There is a typo in the `completed0` comment as well

ticoann: Job can't have multiple location at this point although default value for getSiteName() is {}.
siteName comes from siteConfig, AFAIK, it is just one site.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Bootstrap.py#L235

What could happen is that job might failed with no fwjr. in that case it will show the location which agent selected. I am not sure whether that is desired or we should set unknown. I didn't set unknown because in normal case when JobTracker tracks down the running job it would have correct site even though fwjr doesn't exist.

ticoann: Updated typo as well. Thanks,

ticoann: This will give the fast fail job (missed by JobTracker) correct site name if fwjr is created. So we might not need to look at schedd log to track those jobs although we can't have correct information if fwjr is not generated

amaltaro:
test this please

ticoann:
Hmm, It seems that errors reported are not related to the change.

ticoann:
test this please

PR #7225, opened 2016-09-26
Rely on EnteredCurrentStatus instead of JobStartDate classAd

### LINKED ISSUE #7203: Exit code 99303 to be more informative
Looks to me that the error message could be more informative of the location of the error.
Looks like the job exited properly in condor, but the agent is enable to complete the post-processing : getting the JobReport.*.pkl which might have failed to transfer back due to routing issue. 
It looks as if we should be able to get more informative information. Like "job succeeded but I cannot get the pkl file" or "job status was X, and no pkl file"

amaltaro:
Fixes #7203 (and maybe #7227)
As discussed in https://github.com/dmwm/WMCore/issues/7203 and in gtalk, there are cases where a job is running but it does not have `JobStartDate` classAd, especially on heavily loaded schedds.

Brian suggested to replace that by `EnteredCurrentStatus`, which is always there.

@bbockelm @ticoann please review. Still need to be tested and backported.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -993,33 +993,23 @@ class PyCondorPlugin(BasePlugin):
             logging.debug(msg)
             return None, None
         else:
-            for slicedAds in grouper(itobj, 1000):
-                for jobAd in slicedAds:
-                    ### This condition ignores jobs that are Removed, but stay in the X state
-                    ### For manual condor_rm removal, job wont be in the queue \
-                    ### and status of the jobs will be read from condor log
-                    if jobAd["JobStatus"] == 3:
-                        continue
-                    else:
-                        ## For some strange race condition, schedd sometimes does not publish StartDate for a Running Job
-                        ## Get the entire classad for such a job
-                        ## Do not crash WMA, wait for next polling cycle to get all the info.
-                        if jobAd["JobStatus"] == 2 and jobAd.get("JobStartDate") is None:
-                            logging.debug("THIS SHOULD NOT HAPPEN. JobStartDate is MISSING from the CLASSAD.")
-                            logging.debug("Could be caused by some race condition. Wait for the next Polling Cycle")
-                            logging.debug("%s", str(jobAd))
-                            continue
-
-                        tmpDict = {}
-                        tmpDict["JobStatus"] = int(jobAd.get("JobStatus", 100))
-                        tmpDict["stateTime"] = int(jobAd["EnteredCurrentStatus"])
-                        tmpDict["runningTime"] = int(jobAd.get("JobStartDate", 0))
-                        tmpDict["submitTime"] = int(jobAd["QDate"])
-                        tmpDict["DESIRED_Sites"] = jobAd["DESIRED_Sites"]
-                        tmpDict["ExtDESIRED_Sites"] = jobAd["ExtDESIRED_Sites"]
-                        tmpDict["runningCMSSite"] = jobAd.get("MATCH_EXP_JOBGLIDEIN_CMSSite", None)
-                        tmpDict["WMAgentID"] = int(jobAd["WMAgent_JobID"])
-                        jobInfo[tmpDict["WMAgentID"]] = tmpDict
+            for jobAd in itobj:
+                ### This condition ignores jobs that are Removed, but stay in the X state
+                ### For manual condor_rm removal, job wont be in the queue \
+                ### and status of the jobs will be read from condor log
+                if jobAd["JobStatus"] == 3:
+                    continue
+                else:
+                    tmpDict = {}
+                    tmpDict["JobStatus"] = int(jobAd.get("JobStatus", 100))
+                    tmpDict["stateTime"] = int(jobAd["EnteredCurrentStatus"])
+                    tmpDict["runningTime"] = int(jobAd.get("JobStartDate", 0))
+                    tmpDict["submitTime"] = int(jobAd["QDate"])
+                    tmpDict["DESIRED_Sites"] = jobAd["DESIRED_Sites"]
+                    tmpDict["ExtDESIRED_Sites"] = jobAd["ExtDESIRED_Sites"]
+                    tmpDict["runningCMSSite"] = jobAd.get("MATCH_EXP_JOBGLIDEIN_CMSSite", None)
+                    tmpDict["WMAgentID"] = int(jobAd["WMAgent_JobID"])
+                    jobInfo[tmpDict["WMAgentID"]] = tmpDict
 
             logging.info("Retrieved %i classAds", len(jobInfo))
 
@@ -1052,30 +1042,29 @@ class PyCondorPlugin(BasePlugin):
 
         try:
             logging.debug("Opening condor job log file: %s", logFile)
-            logfileobj = open(logFile, "r")
+            lastEvent = None
+            with open(logFile) as logfileobj:
+                for event in condor.read_events(logfileobj, 1):
+                    lastEvent = event
         except:
             logging.debug('Cannot open condor job log file %s', logFile)
         else:
-            tmpDict = {}
-            cres = condor.read_events(logfileobj, 1)
-            ulog = list(cres)
-            if len(ulog) > 0:
-                if all(key in ulog[-1] for key in ("TriggerEventTypeNumber", "QDate", "JobStartDate",
-                                                   "EnteredCurrentStatus", "MATCH_EXP_JOBGLIDEIN_CMSSite",
-                                                   "WMAgent_JobID")):
-
-                    _tmpStat = int(ulog[-1]["TriggerEventTypeNumber"])
+            if lastEvent:
+                if all(key in lastEvent for key in ("TriggerEventTypeNumber", "QDate", "EnteredCurrentStatus",
+                                                    "MATCH_EXP_JOBGLIDEIN_CMSSite", "WMAgent_JobID")):
+                    tmpDict = {}
+                    _tmpStat = int(lastEvent["TriggerEventTypeNumber"])
                     tmpDict["JobStatus"] = PyCondorPlugin.logToScheddExitCodeMap(_tmpStat)
-                    tmpDict["submitTime"] = int(ulog[-1]["QDate"])
-                    tmpDict["runningTime"] = int(ulog[-1]["JobStartDate"])
-                    tmpDict["stateTime"] = int(ulog[-1]["EnteredCurrentStatus"])
-                    tmpDict["runningCMSSite"] = ulog[-1]["MATCH_EXP_JOBGLIDEIN_CMSSite"]
-                    tmpDict["WMAgentID"] = int(ulog[-1]["WMAgent_JobID"])
+                    tmpDict["submitTime"] = int(lastEvent["QDate"])
+                    tmpDict["stateTime"] = int(lastEvent["EnteredCurrentStatus"])
+                    tmpDict["runningTime"] = int(lastEvent.get("JobStartDate", 0))
+                    tmpDict["runningCMSSite"] = lastEvent["MATCH_EXP_JOBGLIDEIN_CMSSite"]
+                    tmpDict["WMAgentID"] = int(lastEvent["WMAgent_JobID"])
                     jobLogInfo[tmpDict["WMAgentID"]] = tmpDict
                 else:
-                    logging.debug('%s is CORRUPT', str(logFile))
+                    logging.debug('%s does not have all the classAds', str(logFile))
             else:
-                logging.debug('%s is EMPTY', str(logFile))
+                logging.debug('%s is empty', str(logFile))
 
         logging.info("Retrieved %i Info from Condor Job Log file %s", len(jobLogInfo), logFile)
 


src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -875,7 +875,7 @@ def makeSubmit(self, jobList):
             jdl.append("+JobLeaseDuration = isUndefined(MachineAttrMaxHibernateTime0) ? 1200 : MachineAttrMaxHibernateTime0\n")
 
             ### print all the variables needed for us to rely on condor userlog
-            jdl.append("job_ad_information_attrs = JobStatus,QDate,EnteredCurrentStatus,JobStartDate,DESIRED_Sites,ExtDESIRED_Sites,WMAgent_JobID,MATCH_EXP_JOBGLIDEIN_CMSSite\n")
+            jdl.append("job_ad_information_attrs = JobStatus,QDate,EnteredCurrentStatus,DESIRED_Sites,ExtDESIRED_Sites,WMAgent_JobID,MATCH_EXP_JOBGLIDEIN_CMSSite\n")
```
bbockelm: Replace `MATCH_EXP_JOBGLIDEIN_CMSSite` with `JOBGLIDEIN_CMSSite`.

bbockelm: Note that `EnteredCurrentStatus` could be dropped: it's going to be present already.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -984,7 +984,7 @@ def getClassAds(self):
             logging.debug("Start: Retrieving classAds using Condor Python XQuery")
             itobj = schedd.xquery(
                 'WMAgent_JobID =!= "UNDEFINED" && WMAgent_AgentName == %s' % classad.quote(str(self.agent)),
-                ["JobStatus", "EnteredCurrentStatus", "JobStartDate", "QDate", "DESIRED_Sites",
+                ["JobStatus", "EnteredCurrentStatus", "QDate", "DESIRED_Sites",
```
bbockelm: Keep `JobStartDate` here.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -993,33 +993,23 @@ def getClassAds(self):
...
+                tmpDict = {}
+                ### This condition ignores jobs that are Removed, but stay in the X state
+                ### For manual condor_rm removal, job wont be in the queue \
+                ### and status of the jobs will be read from condor log
+                if jobAd["JobStatus"] == 3:
+                    continue
+                else:
+                    tmpDict["JobStatus"] = int(jobAd.get("JobStatus", 100))
+                    tmpDict["stateTime"] = int(jobAd["EnteredCurrentStatus"])
+                    tmpDict["runningTime"] = tmpDict["stateTime"] if jobAd['JobStatus'] == 2 else 0
```
bbockelm: Change this to:
```
int(jobAd.get("JobStartDate", jobAd["EnteredCurrentStatus"]))
```

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -1060,15 +1050,14 @@ def readCondorLog(self, job):
             cres = condor.read_events(logfileobj, 1)
             ulog = list(cres)
             if len(ulog) > 0:
-                if all(key in ulog[-1] for key in ("TriggerEventTypeNumber", "QDate", "JobStartDate",
-                                                   "EnteredCurrentStatus", "MATCH_EXP_JOBGLIDEIN_CMSSite",
-                                                   "WMAgent_JobID")):
+                if all(key in ulog[-1] for key in ("TriggerEventTypeNumber", "QDate", "EnteredCurrentStatus",
```
bbockelm: Can you drop the `all` here?  Particularly, `MATCH_EXP_JOBGLIDEIN_CMSSite` is not guaranteed to exist.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -1060,15 +1050,14 @@ def readCondorLog(self, job):
...
+                                                   "MATCH_EXP_JOBGLIDEIN_CMSSite", "WMAgent_JobID")):
 
                     _tmpStat = int(ulog[-1]["TriggerEventTypeNumber"])
                     tmpDict["JobStatus"] = PyCondorPlugin.logToScheddExitCodeMap(_tmpStat)
                     tmpDict["submitTime"] = int(ulog[-1]["QDate"])
-                    tmpDict["runningTime"] = int(ulog[-1]["JobStartDate"])
                     tmpDict["stateTime"] = int(ulog[-1]["EnteredCurrentStatus"])
+                    tmpDict["runningTime"] = tmpDict["stateTime"] if tmpDict["JobStatus"] == 2 else 0
```
bbockelm: Replace with `int(ulog[-1].get("JobStartDate", 0)`

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -1060,15 +1050,14 @@ def readCondorLog(self, job):
             cres = condor.read_events(logfileobj, 1)
             ulog = list(cres)
```
bbockelm: Note that this line is way overkill if we only use the last entry: this is really an iterator and keeping the results in memory is somewhat costly.  You could do something like:
```
event = None
for event in cres:
   pass
if event:
   ...
```

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -1060,15 +1050,14 @@ def readCondorLog(self, job):
...
+                if all(key in ulog[-1] for key in ("TriggerEventTypeNumber", "QDate", "EnteredCurrentStatus",
+                                                   "MATCH_EXP_JOBGLIDEIN_CMSSite", "WMAgent_JobID")):
 
                     _tmpStat = int(ulog[-1]["TriggerEventTypeNumber"])
                     tmpDict["JobStatus"] = PyCondorPlugin.logToScheddExitCodeMap(_tmpStat)
                     tmpDict["submitTime"] = int(ulog[-1]["QDate"])
-                    tmpDict["runningTime"] = int(ulog[-1]["JobStartDate"])
                     tmpDict["stateTime"] = int(ulog[-1]["EnteredCurrentStatus"])
+                    tmpDict["runningTime"] = tmpDict["stateTime"] if tmpDict["JobStatus"] == 2 else 0
                     tmpDict["runningCMSSite"] = ulog[-1]["MATCH_EXP_JOBGLIDEIN_CMSSite"]
```
bbockelm: Replace with:
```
ulog[-1].get("JOBGLIDEIN_CMSSite", ulog[-1].get("MATCH_EXP_JOBGLIDEIN_CMSSite", ""))
```

bbockelm (changes requested):
If `JobStartDate` is available, it's probably the best thing to use for the `runningTime` attribute - we should keep that.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: `JOBGLIDEIN_CMSSite` seems to be an expression using `GLIDEIN_CMSSite` which turns to be undefined. E.g. from vocms0303:
```
cmst1@vocms0303:/data/srv/wmagent/current $ condor_q -af:h CMS_JobType JobStatus MATCH_EXP_JOBGLIDEIN_CMSSite JOBGLIDEIN_CMSSite GLIDEIN_CMSSite -limit 2
CMS_JobType JobStatus MATCH_EXP_JOBGLIDEIN_CMSSite JOBGLIDEIN_CMSSite                                                         GLIDEIN_CMSSite
Processing  2         T2_FR_IPHC                   $$([ifThenElse(GLIDEIN_CMSSite is undefined, "Unknown", GLIDEIN_CMSSite)]) undefined      
Production  2         T1_RU_JINR                   $$([ifThenElse(GLIDEIN_CMSSite is undefined, "Unknown", GLIDEIN_CMSSite)]) undefined      
```

About `EnteredCurrentStatus`, I'll keep it around just to be explicit that it's supposed to be in the condor log.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: Adding it back

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: This code considers that `event` will leak out of the for-loop scope. Though stack overflow says it's a safe construction :)

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: I think the best we can do is to get this Ad and if it's not available, fallback to the first site in the `DESIRED_Sites`. Actually, it may spoil  the BossAir db in case it already had the right site information.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: Actually it's `runningTime`. So we either set it to 0 in case `JobStartDate` is not available or we do some sort of `int(time.time()) - jobAd["EnteredCurrentStatus"]`. Keeping it with 0 for the moment.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: done

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: See comment above

hufnagel:
Just as a comment, the SimpleCondorPlugin doesn't use any of those. It seemed to me the PyCondorPlugin code wasn't actually doing anything useful with all the ClassAds it retrieves...

amaltaro:
Sigh... my test jobs are not really running in ITB. 

How about @bbockelm and @ticoann have a last review (I prefer to be as conservative as possible here) and then we push this patch to one production agent only and watch it for some hours a day before rolling it everywhere?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: Yeah - that's precisely why the original log code doesn't work.

The _schedd_ evaluates the expression and saves the resulting string as `MATCH_EXP_JOBGLIDEIN_CMSSite`.  That's what you see in the output of `condor_q` or `condor_history`.  However, when it sends the ClassAd to the `condor_shadow` (the `condor_shadow` writes the log), it saves the resulting string to `JOBGLIDEIN_CMSSite`.

Hence the shadow, when writing the log, sees something like `T2_US_Nebraska` instead of `$$([ifThenElse(...)])`.

(If you can't tell, this isn't the recommended way to get this information into the log - it's a bit roundabout and seems prone to issues.)

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: Ah, OK.  I was a bit surprised that it was a Unix timestamp and not a relative value.  Were we just doing this incorrectly for multiple years?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: That's fine too.  Just the assumption that _all_ data was present (particularly, `JOBGLIDEIN_CMSSite` is only relevant if the job was running - might not be true for held or idle jobs) was causing the code to not work.

Having slightly wrong data may be the best we can do, unfortunately.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: Very confusing indeed. Just to make sure I understood you.

No matter what we set to `job_ad_information_attrs`, either `MATCH_EXP_JOBGLIDEIN_CMSSite` or `JOBGLIDEIN_CMSSite` would map to the same `JOBGLIDEIN_CMSSite` in the condor log (string), for running jobs. Is that correct?

Hmm, having a look at a couple of logs from production - for running jobs - one of them has the MATCH_blah classAd and the other doesn't.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: `runningTime` seems to be correct. Either 0 or JobStartDate. We probably mixed it with `stateTime`

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: We could make it safer like
```
lastEvent= None
for event in logEvents:
   lastEvent = event
```
though from a quick test, it seems to be ~60% slower (for a sample of 10M integers) than the code in the PR.

ticoann:
The main fix for this patch is just removing following condition, right?

https://github.com/dmwm/WMCore/blob/b06fea845524a910522a077dc46d0d312d6b31e7/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py#L1007

I was just wondering whether we can keep the Complete(or Removed) status a little longer in schedd, (Maybe twice of the JobTracker polling cycle.) It that too much burden for schedd or agent to keep it that wrong?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: Almost.  

For running jobs, `MATCH_EXP_JOBGLIDEIN_CMSSite` should show up in the condor log if the job was previously run and is being restarted a second time; I believe `MATCH_EX_JOBGLIDEIN_CMSSite` should show the site where the job previously run.

`JOBGLIDEIN_CMSSite` should point at the correct site for running jobs.

Since the job is not necessarily running when this code executes, maybe this is the most bulletproof variant:
```
if ad.get("JOBGLIDEIN_CMSSite", "").startswith("$$"): #JOBGLIDEIN_CMSSite present but not evaluated
  site = ad.get("MATCH_EXP_JOBGLIDEIN_CMSSite", "Unknown")
else:
  site = ad.get("JOBGLIDEIN_CMSSite", "Unknown")
```

hufnagel:
You can. It's a classadd you need to set in the jobs, it's on my todo list to test this.

In all of this, I really don't want us spending a lot of effort to optimize PyCondorPlugin. I think all of this information we retrieve is a bit useless anyways, SimpleCondorPlugin doesn't look at any of it.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
bbockelm: Well yes - in fact, I would have expected to be even slower than that for iterating through a list of 10M integers.

However, in the above example, we don't have a list of events - each iteration through the loop is a call to a HTCondor library.  The "optimization" from the for loop means that we don't keep all the events in memory - just the last one.

bbockelm:
@ticoann - Yes - the most important part of the patch is removing that conditional.  It was triggering the use of the log-reading code - particularly under heavy load.

The rest is just fixing up the log reading code (which we hope never is invoked) which appears to have only functioned in rare cases.

When that condition was hit and the log reading code didn't work, then the jobs would be considered complete by the agent.  As they were still running in HTCondor, they would not have a FJR returned, causing the issue in #7203.

ticoann:
Yes, Dirk. The goal is we will just apply  Alan's patch to production agent and start to use SimpleCondorPlugin for next deployment. (We will start  testing on that)

ticoann:
@bbockelm, Thanks Brian, it is clear now. I was wondering is there any documentation of list of parameter (i.e EnteredCurrentStatus) and their usage/definition?

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: Ok, got it!

amaltaro:
Provided we don't want to spend too much effort on this plugin. I'm in favor of finishing the validation of this patch - as it is - and then we push it to the agents tomorrow or when that's finished.

ticoann:
Alan, yes I agree.

bbockelm:
@ticoann - there is indeed documentation on most of the attributes:

http://research.cs.wisc.edu/htcondor/manual/v8.5/12_Appendix_A.html

However, it's pretty lengthy - not for pleasant reading - and doesn't always cover the historical bugs / semantics.  It'll be good for about 90% of the use cases, but dropping me a line for the last 10% of sticky situations may be a good idea.

ticoann:
Thanks Brian for the pointer.

amaltaro:
Fixed a crash in the code while using the event iterator over a file object outside the with context :(

And put back the `1` argument to the `read_events` api, otherwise the component was reading logs at 1Hz instead of the usual ~10. I assume it reads only one event, right?

Brian, the documentation mentions only 2 args, file object and xml flag. Is it because we're using an old htcondor release in comp?

amaltaro:
test this please

amaltaro:
Tests are looking good now. I'm rolling it to production.

amaltaro:
@ticoann please merge
@bbockelm can you enlighten me on the `read_events` question I asked above?

bbockelm:
Hi Alan,

Yes - `read_events` appears to need to explicitly be told whether the logfile is XML or the custom format.  I don't quite understand why - it should autodetect - but let's keep `read_events(logobj, 1)` for now.

Brian

PR #7216, opened 2016-09-23
Do not report stage-out se-name to dashboard - 1.0.19v

amaltaro:
I stumbled upon this issue while investigating a wmstats issue.

This se-name code was added in 20 Apr, https://github.com/dmwm/WMCore/pull/6810 and it surprises me that we hit it only now, maybe because this problem happened in the exception block (and some sites still have se-name in their site-local-config).

This is the traceback from the job logs:

```
ERROR:root:Exception occured when executing step
ERROR:root:Exception is 'se-name'
ERROR:root:Traceback:
ERROR:root:Traceback (most recent call last):
  File "/data/srv/wmagent/v1.0.19.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.0.19.patch1/lib/python2.7/site-packages/WMCore/WMSpec/Steps/ExecuteMaster.py", line 140, in doExecution
    executionObject.execute()
  File "/tmp/22098895.1.cms-sl6/glide_EN6oF7/execute/dir_2659860/job/WMCore.zip/WMCore/WMSpec/Steps/Executors/StageOut.py", line 200, in execute
    manager(fileForTransfer)
  File "/tmp/22098895.1.cms-sl6/glide_EN6oF7/execute/dir_2659860/job/WMCore.zip/WMCore/Storage/StageOutMgr.py", line 205, in __call__
    fileToStage = stageoutPolicyReport(fileToStage, self.siteCfg.localStageOut['se-name'],
KeyError: 'se-name'
```

@ticoann @hufnagel please review, it still needs to be tested

PR #7231, opened 2016-09-28
crab version of #7216

ticoann:
@juztas, please review

### DIFF
--- a/src/python/WMCore/Services/Dashboard/DashboardAPI.py
+++ b/src/python/WMCore/Services/Dashboard/DashboardAPI.py
@@ -289,7 +289,7 @@ def reportFailureToDashboard(exitCode, ad=None, stageOutReport=None):
     apmonFree()
     return exitCode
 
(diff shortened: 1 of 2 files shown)


juztas:
@ticoann it will not work and will fail earlier. There are more places there se-name is defined...
For example this: https://github.com/dmwm/WMCore/commit/6f08f43d50c909a0e6762c3aee3cef912a9472d1 is needed for sure to be safe.
Let's wait for @emaszs @mmascher if they are ready to move to new WMCore (take master and go with validation). So far CRAB requires both and se-name and PNN defined in site-local-config...

ticoann:
Hmm, so it not just Alan's patch. In that case it might be better to move to new version of WMAgent and validate. I just tried the other patch. there are many conflicts and it probably need validation with all the patches.

ticoann:
closing this in favor of using newer version of WMCore for crab

PR #7232, opened 2016-09-28
sleep and not overload agents

juztas:
Same hack as in CRAB as per @vlimant request

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -10,6 +10,25 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
 mv tmp_env.sh startup_environment.sh
 export JOBSTARTDIR=$PWD
 
+
+# Saving START_TIME and when job finishes, check if runtime is not lower than 20m
+# If it is lower, sleep the difference. We don`t want to overload agents with too fast jobs...
+START_TIME=$(date +%s)
+function finish {
+  END_TIME=$(date +%s)
+  DIFF_TIME=$((END_TIME-START_TIME))
+  echo "Job Running time in seconds: " $DIFF_TIME
+  if [ $DIFF_TIME -lt 1200];
+  then
+    SLEEP_TIME=$((1200 - DIFF_TIME))
+    echo "Job runtime is less than 20minutes. Sleeping " $SLEEP_TIME
+    sleep $SLEEP_TIME
+  fi
+}
+# Trap all exits and execute finish function
+trap finish EXIT
+
+
 # should be a bit nicer than before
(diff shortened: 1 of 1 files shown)


juztas:
for reference:
dashb-cms-job.cern.ch/dashboard/templates/web-job2/#user=&refresh=0&table=Jobs&p=1&records=25&activemenu=0&usr=&site=&submissiontool=&application=&activity=&status=&check=submitted&tier=&date1=2016-09-28+18%3A59&date2=2016-09-28+20%3A59&sortby=site&scale=linear&bars=20&ce=&rb=&grid=&jobtype=&submissionui=&dataset=&submissiontype=&task=wmagent_pdmvserv_HIG-RunIISummer15wmLHEGS-00597_00162_v0_TTbbWithttHFGenFilter_160923_122154_89&subtoolver=&genactivity=&outputse=&appexitcode=&accesstype=&inputse=&cores=
https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/3262.html

amaltaro:
This will certainly mitigate a few issues we can have in production.
This `trap` command is interesting, I wonder what happens if user condor_rm a job, would that be killed or that signal would hang on the `finish` function until the 20min has passed?

amaltaro:
The sigkill timeout is configurable per workflow, but it defaults to 5min in production as well. My concern is that these jobs cannot be killed within 20min. Another possible problem would be in HLT, since jobs have to be checkpointed/hibernate.

Last but not least, for very stable resources and workflows, this 20min will add wastage of resources, as for the T0 case @hufnagel ?

juztas:
@amaltaro they can be killed, and sigkill is not catchable. one of the linux signal which you can`t catch. 
also, what about this:
add to each schedd a configuration knob (you can skip T0 schedds if you want): 
```
MIN_JOB_RUNTIME_SECONDS = 1200
SUBMIT_ATTRS = $(SUBMIT_ATTRS) MIN_JOB_RUNTIME_SECONDS
```
Now before sleep, check all jobad's and if it has MIN_JOB_RUNTIME_SECONDS, it will use it, otherwise it will skip sleeping. So no wastage on T0 and printing jobad's in job log file would improve debugging for sure. Another thing why 20mins, job matching and etc and if this slot is somehow bad (cvmfs, scratch space, etc) currently it would take one job after another quickly and would fail them constantly.

juztas:
also too many short jobs was one of scalability issues for crab.

juztas:
Close. new is here with more improvements: https://github.com/dmwm/WMCore/pull/7236
and with possible solution for T0.

Issue #4716, opened 2013-07-08
Monitoring fix for late binding


dballesteros7:
If jobs fail out too fast after moving to running state the JobStatusLite component has no chance to updating the running site to the real site instead of the randomly selected one in submission. This causes problems in the monitoring since we get failures appointed to the wrong sites in WMStats and Dashboard.

Possible solution is to retrieve condor_history on "missing" jobs and making sure of the running site when declaring them as failed. It should be top priority.

amaltaro:
I think it has been fixed in the last couple of months, so closing it.

Issue #7224, opened 2016-09-26
Access to condor logs


vlimant:
As we discussed with @juztas @vlimant and this hypernews [https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/3207.html], this requires wider communication and this can be reported on Monday meeting and discussed. Here are the steps which needs to be done if people agree. Also, this is just my suggestion, please feel free to accept it, ignore, modify...

Cheers,
Justas

[1]
a) Modify JobArchiver that whenever it tarballs condor*{err,out,log} files, it does not remove them, but moves to new location, which is defined in WMAgent configuration. Keep directory structure the same as JobCache, which would allow easily to map on Brian's ES, WMStats, Dashboard.
b) For wmagent puppet configuration reuse CRAB code for HTTP and cleanup script. Cleanup script is just simple cronjob with controlled date. Lets say keep last 30days in beginning and lare can be modified depending on the usage.
c) Get additional 1TB mounted for each schedd. Just to make sure that in case there are huge IO read for logs, it does not interfere with mounted volume where WMAgent or Schedd is running.
d) Rewrite rules in cmsweb, that it allows to access it from outside CERN. This will make sure that all users are authenticated and have certificate available. Now URL could be for example: https://cmsweb.cern.ch/prodlogs/vocms0303/taskname.../.../../../
e) Report to dashboard for each job a location to log file.
f) Start to improve the logging and what information it keeps.

P.S. Problem which I foresee is FNAL schedds, but this also could be solved, but would require FNAL people to open a port 80 to all cmsweb machines.

If an agreement is found, I volunteer myself for a and e, if you need help. For b, c, d there are a lot of people in CatA's office who know all details and can make it good and f is just longer term and would improve in time.

juztas:
@amaltaro @ticoann comments?

amaltaro:
We discussed this issue in the last WMAgent meeting and we'd like to explore an approach like what we have in CrabCache (something like a dropbox).

The idea would be to upload the wrapper logs to this area in cmsweb and report the url for these logs in wmstats and dashboard. 

Of course we cannot upload every single log to cmsweb, so the initial idea would be to upload a few logs (5?) for `cooloff` or `failed` jobs only, for each set of `workflow` + `agent` + `task` + `site` + `exitcode`.

amaltaro:
[...]

PR #7236, opened 2016-09-29
Improve submit.sh logging, exitCodes check

juztas:
a) Make sure script runtime is not less than set by a scheduler. Also it will force to sleep up to 5mins if this bootstrap script fails, which mainly means something is wrong with this WN or transferring files to it failed for unknown reason. Re-sending other jobs so fast to same worker node lead to same failure most often. if this MIN_JOB_RUNTIME_TIME_SECONDS is not set, will use 0. If Job exited 0 and bootstrap script also 0, it will not sleep any time and will exit.
b) Improve logging and separate what is where... Also to addition print the condor job ad content for debug reasons and also print environment more nicely.
c) output file variable is a string separated with spaces. Each of word is treated as a separate argument, so this shows a failure: `[: -eq: unary operator expected`

So this would be one of the steps for better debugging, one of the items here: https://github.com/dmwm/WMCore/issues/7224
Also I would propose to add a check on 68 & 137 jobrc code and print some system information, like ulimit, disk space, memory usage, but this would overload this pull request.

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -1,4 +1,15 @@
 #!/bin/bash
+# should be a bit nicer than before
+echo "======== WMAgent bootstrap STARTING at $(TZ=GMT date) ========"
+echo "User id:    $(id)"
+echo "Local time: $(date)"
+echo "Hostname:   $(hostname -f)"
+echo "System:     $(uname -a)"
+echo "Arguments:  $@"
+
+# Saving START_TIME and when job finishes END_TIME.
+WMA_MIN_JOB_RUNTIMESECS=300
+START_TIME=$(date +%s)
 
 # On some sites we know there was some problems with environment cleaning
 # with using 'env -i'. To overcome this issue, whenever we start a job, we have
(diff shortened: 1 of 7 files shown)


ticoann:
test this please

amaltaro:
It's only the bash code, I don't think there are tests for it.
I'll have a look at it in the next week, once we get closer to completing the testbed validation.

juztas:
test this please

juztas:
I saw (I think) that Eric fixed whenever no python files are modified. lets see test run

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
...
+    sleep $SLEEP_TIME
+  fi
+  echo "======== WMAgent final job runtime checks FINISH at $(TZ=GMT date) on $(hostname) ========"
+}
+# Trap all exits and execute finish function
+trap finish EXIT
+
 # should be a bit nicer than before
-echo "WMAgent bootstrap : `date -u` : starting..."
+echo "======== WMAgent bootstrap STARTING at $(TZ=GMT date) on $(hostname) ========"
```
amaltaro: Can you move these 4 lines to the extreme top of this file, right after /bin/bash.
Also print the username id please

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
...
+echo "Local time : $(date)"
+echo "Current system : $(uname -a)"
+echo "Arguments are $@"
 
 # We need to create the expected output file in advance just in case
 # some problem happens during the job bootstrap
 outputFile="Report.0.pkl Report.1.pkl Report.2.pkl Report.3.pkl"
-if [ -n "$_CONDOR_JOB_AD" ]; then
-    outputFile=`grep ^TransferOutput $_CONDOR_JOB_AD | awk -F'=' '{print $2}' | sed 's/\"//g'`
+MIN_JOB_RUNTIME_TIME_SECONDS=0
```
amaltaro: Set it to 300 secs please and rename it to `WMA_MIN_JOB_RUNTIMESECS`, just to avoid any var name collision.

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
 mv tmp_env.sh startup_environment.sh
 export JOBSTARTDIR=$PWD
 
+# Saving START_TIME and when job finishes END_TIME.
+# Multiple checks are done:
+# 1) Check submit.sh bootstrap exit code.
+#    If it is not 0 sleep MIN_JOB_RUNTIME_SECONDS and in case it is 0, force to sleep 5 mins total runtime
```
amaltaro: Adapt this comment

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
...
+#    If this bootstrap script is exiting non 0 exit code, something is wrong on this WN...
+# 2) If Job exited 0, do not sleep any timeout
+# 3) Otherwise, sleep up to MIN_JOB_RUNTIME_TIME_SECONDS seconds of runtime.
+#    By default MIN_JOB_RUNTIME_TIME_SECONDS will be set to 0, if this is not passed through agent schedd configuration
+# Fast finishing jobs (failed specially) might overload schedulers and 
+# restarting jobs on same bad WN leads to more failures
```
amaltaro: Remove these 3 last lines of comment

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
...
+START_TIME=$(date +%s)
+function finish {
+  exitCode=$?
+  echo "======== WMAgent final job runtime checks STARTING at $(TZ=GMT date) on $(hostname) ========"
```
amaltaro: use `hostname -f` instead please. I've seen cases where just `hostname` gives us the short name, not fqdn. Update all the others as well.

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
...
+MIN_JOB_RUNTIME_TIME_SECONDS=0
+if [ "X$_CONDOR_JOB_AD" != "X" ];
+then
+   echo "======== HTCONDOR JOB SUMMARY START at $(TZ=GMT date) on $(hostname) ========"
+   outputFile=`grep ^TransferOutput $_CONDOR_JOB_AD | awk -F'=' '{print $2}' | sed 's/\"//g'`
+   JOB_CMSSite=`grep '^JOB_CMSSite =' $_CONDOR_JOB_AD | tr -d '"' | awk '{print $NF;}'`
+   MIN_JOB_RUNTIME_TIME_SECONDS=`grep '^MIN_JOB_RUNTIME_TIME_SECONDS =' $_CONDOR_JOB_AD | tr -d '"' | awk '{print $NF;}'`
+   echo "Execution site: $JOB_CMSSite"
+   echo "Current hostname: $(hostname)"
+   echo "Output files: $outputFile"
```
amaltaro: I think outputFile is always going to be the same (condor logs), so I'd get rid of it.

etc/submit.sh
```diff
@@ -10,37 +10,100 @@ sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
...
+   echo "==== HTCONDOR JOB AD CONTENTS START ===="
+   while read i; do
+       echo "== JOB AD: $i"
```
amaltaro: How about just print the ad itself `echo "$i"`

amaltaro (commented):
It may have become way too much verbose now, but I want to see a real case first

etc/submit.sh
juztas: what do you mean by that? It is

etc/submit.sh
juztas: it is just a comment to understand why it was done.

etc/submit.sh
juztas: why here 300 ?

etc/submit.sh
juztas: in any case it will be in condor job ads

juztas:
at which point it is too much verbose? additionally to previous version it will print condor job ads and also will separate nicely what is where...

etc/submit.sh
amaltaro: Because you'll set it to 300, so it will not be 0 and there is not `if` :)

etc/submit.sh
amaltaro: Yep, but the motivation for that is clear and you already put it in the PR.

etc/submit.sh
amaltaro: just to update the default value from 0 to 300, if we ever want to change it. Then we change and the changes take effect right away.

amaltaro:
BTW, fixes #4716

etc/submit.sh
```diff
@@ -1,4 +1,13 @@
 #!/bin/bash
+# should be a bit nicer than before
+echo "======== WMAgent bootstrap STARTING at $(TZ=GMT date) on $(hostname -f) ========"
+echo "User id : $(id)"
+echo "Local time : $(date)"
+echo "Current system : $(uname -a)"
+echo "Arguments are $@"
+echo "Current hostname: $(hostname -f)"
```
amaltaro: Can you update this line to `echo "Current hostname: $(hostname -f) at $GLIDEIN_CMSSite"`?

amaltaro: Actually, the hostname is already in line 3. So you can remove this line 8 and update line 3 with `at $GLIDEIN_CMSSite` as well.

amaltaro:
@juztas please make a new commit if you make any further changes to this patch. I'm going to start its validation as it is new and a new commit helps to find out what were the latest changes.

amaltaro:
I have a few more suggestions on this patch. I'll just make a PR against your branch instead of asking for changes and waiting for them to get applied.

juztas:
great. go ahead.

amaltaro:
Meanwhile, what's the difference between `Job_CMSSite` and `GLIDEIN_CMSSite`? I asked Farrukh and he's not aware of the former.

juztas:
Mh... Brian knows better. I think (what I see) that GLIDEIN_CMSSite is a pilot attribute and 'JOB_CMSSite' is only crab specific. I am not sure why this is needed.
Prod:
```
JOB_GLIDEIN_CMSSite = "$$(GLIDEIN_CMSSite:Unknown)"
JOBGLIDEIN_CMSSite = "$$([ifThenElse(GLIDEIN_CMSSite is undefined, \"Unknown\", GLIDEIN_CMSSite)])
```
while analysis has:
```
JOB_CMSSite = "$$(GLIDEIN_CMSSite:Unknown)"
JOB_GLIDEIN_CMSSite = "$$(GLIDEIN_CMSSite:Unknown)"
JOBGLIDEIN_CMSSite = "$$([ifThenElse(GLIDEIN_CMSSite is undefined, \"Unknown\", GLIDEIN_CMSSite)])"
```
Now this is defined in schedd configuration, so I am not sure if all 3 are needed.

amaltaro:
If it helps, a job running in a HLT worker node failed (was empty) to print both classAds/vars :-(

juztas:
hlt is special. I also have to do hacks for it in cms-gwmsmon. Can you give me full log file?

amaltaro:
cmst1@vocms0224:/data/srv/wmagent/current $ vim install/wmagent/JobArchiver/logDir/a/amaltaro_ReDigi_cmsRun3_Agent1021_Validation_161012_185344_7929/JobCluster_16/Job_16692.tar.bz2
though the JOB_CMSSite ad is there, the other GLIDEIN is not.

juztas:
as I thought. hlt pilot startup does not set required value, so for production, use always: JOB_GLIDEIN_CMSSite. Analysis does not run on hlt, so no problem

amaltaro:
@juztas can you squash it all please? :)

amaltaro:
Justas, if you prefer you can create a new clean PR. Just in case, you can find the latest file here:
http://amaltaro.web.cern.ch/amaltaro/forAlan/submit.sh

juztas:
Closed. New is here: https://github.com/dmwm/WMCore/pull/7294 Thanks Alan

PR #7241, opened 2016-09-30
add assign params in acdc

### LINKED ISSUE #7157: ReqMgr2: ACDCs do not inherit assignment information
you have to fill the form from scratch. I'm temporarily placing a screenshot here to make it clear:
https://www.dropbox.com/s/95wes2ujzrdhhyq/assignment.png?dl=0

ticoann:
fixes #7157  Also add the dropdown menu for some of the properties

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
+++ b/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
@@ -433,6 +433,36 @@ WMStats.ViewModel = (function (){
             summary.SoftwareVersions = requestInfo.SoftwareVersions;
             summary.CMSSWVersion = requestInfo.SoftwareVersions[0];
             summary.Campaign = requestInfo.Campaign;
+            // assign more value for acdc assignment
+            summary.Team = requestInfo.Teams[0];
+            //TODO get the site white list from acdc db. (site might be overflowed)
+            //also change templete to handle multiple selection
+            //summary.SiteWhitelist = requestInfo.SiteWhitelist.join();
+            //summary.SiteBlacklist = requestInfo.SiteBlacklist.join();
+            summary.AcquisitionEra = requestInfo.AcquisitionEra;
+            summary.ProcessingString = requestInfo.ProcessingString;
+            summary.ProcessingVersion = requestInfo.ProcessingVersion;
+            summary.Dashboard = requestInfo.Dashboard;
+            summary.MergedLFNBase = requestInfo.MergedLFNBase;
+            summary.TrustSitelists = false;
+            summary.UnmergedLFNBase = requestInfo.UnmergedLFNBase;
+            summary.MinMergeSize = requestInfo.MinMergeSize;
+            summary.MaxMergeSize = requestInfo.MaxMergeSize;
+            summary.MaxMergeEvents = requestInfo.MaxMergeEvents;
+            summary.MaxVSize = requestInfo.MaxVSize;
+            summary.BlockCloseMaxWaitTime = requestInfo.BlockCloseMaxWaitTime;
+            summary.BlockCloseMaxFiles = requestInfo.BlockCloseMaxFiles;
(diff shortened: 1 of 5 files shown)


ticoann:
@amaltaro, Tested in my vm so I will merge for the new tag. But please review.

src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
```diff
@@ -433,6 +433,36 @@ WMStats.ViewModel = (function (){
...
+            //TODO get the site white list from acdc db. (site might be overflowed)
+            //also change templete to handle multiple selection
+            //summary.SiteWhitelist = requestInfo.SiteWhitelist.join();
+            //summary.SiteBlacklist = requestInfo.SiteBlacklist.join();
+            summary.AcquisitionEra = requestInfo.AcquisitionEra;
+            summary.ProcessingString = requestInfo.ProcessingString;
+            summary.ProcessingVersion = requestInfo.ProcessingVersion;
+            summary.Dashboard = requestInfo.Dashboard;
+            summary.MergedLFNBase = requestInfo.MergedLFNBase;
+            summary.TrustSitelists = false;
```
amaltaro: shouldn't it be `requestInfo.TrustSitelists` instead of false

src/python/WMCore/ReqMgr/Web/utils.py
```diff
@@ -264,3 +271,18 @@ def sort(docs, sortby):
...
+def reorder_list(org_list, first):
+    """
+    if the first is in the list.
+    move the first in front of the list
+    if not, add first to the list
+    """
+    new_list = list(org_list)
+    try:
+        i = new_list.index(first)
+        new_list[0], new_list[i] = new_list[i],  new_list[0]
```
amaltaro: I'd say this cannot work!! But it seems the way python evaluates this expression makes the trick...

amaltaro: BTW, I'm not sure the case `if not, add first to the list` can ever happen. And if it does, that would be for a good reason (like cmssw that got deprecated, and in this case I'd be in favor of not adding new values on top of the original source of information)

amaltaro (commented):
Good job!

src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
ticoann: For acdc, it shouldn't always set the TrustSitelist false? since there is  only one location for unmerged input. Except when the first step fails. - In that case requestor can overwrite. I thought default value should be false.

hufnagel: TrustSiteList only applies to the original top-level task. If you rerun jobs from that one, you need to inherit from the request. If you are not (like rerunning some merges), you want to use false.

src/python/WMCore/ReqMgr/Web/utils.py
ticoann: The code does is swapping the items on the list. So it will have correct default selected options in the html options. (Code itself works, I tested).
I know that condition won't be ever happen except the case you describe or if we add remove certain options between the deployment and acdc is created from the previous workflow. However, I wasn't sure whether remove parameter still need to be set for previous workflow. In that case we can run the acdc at all using GUI interface. The chances would be close to none but I thought would be safer to do.

src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
ticoann: This only affects default value. At assign ops have review and change if they needed. I guess I can check whether it is for the top level task here and set the default value accordingly.
```
if toplevel_task:
   summary.TrustSitelists = requestInfo.TrustSitelists;
else:
   summary.TrustSitelists = false;
```
I will create the new issue.

PR #7242, opened 2016-09-30
Make sure it sets StopTime

juztas:
@ticoann you were too fast. So I checked the code and in case there is any WMExecutionFailure or Exception in execution, endTime will not be set and that is why this key error was happening. Now this one for sure should add endTime for each execution.


ticoann:
Hi Justas, are you talking about #7242. I just merged, since it won't break anything and I am creating the new tag.  If this guarantees _ExeWCTime key would exist. We need to put the change back or have more clear message than setting that with default value (0)

amaltaro:
@ticoann the change sounds correct. I also ran some workflows with this patch in and it looks ok.
I suggest we merge it and then the last bits should have been fixed in https://github.com/dmwm/WMCore/pull/7257

ticoann:
Merging it, only affects WMAgent.

PR #7244, opened 2016-09-30
Gfal2 plugin is too much verbose...

juztas:
Removing printenv, as it will always use job startup environment and also remove -vvv, as it is too much verbose... In case there is an issue on site, this can be turned on through site-local-config option.


juztas:
will fix this unit test now

amaltaro:
I ran some tests and it looks good.
@ticoann please merge, it affects only the agent.

amaltaro (approved):

PR #7248, opened 2016-09-30
fix trust site list default value

ticoann:
Set the acdc TrustedSite default value according to the task. (First task inherit, others set to false)
@amaltaro, please review


amaltaro:
Looks good! Though I would prefer to make it default to False and let the P&R handle that...

ticoann:
@amaltaro, Alan I thought you said you won't work weekend. Stop replying. 😄 If you prefer current setting we don't have to add. I will check with ops as well.

PR #7147, opened 2016-09-02
Fix ReqMgr2 validation and spec args definition - take2

amaltaro:
Superseeds #7139 

These are the changes:
- Fix the validation code used by ReqMgr2 (for both spec creation and assignment)
- Use the same validation used by ReqMgr2 in ReqMgr
- Fix some spec arguments and their definition

PR #7250, opened 2016-10-03
Inherit MaxRSS and Sitelists for ReqMgr2 ACDCs

amaltaro:
I found some issues in my VM and this PR tries to address them, as:
- `MaxRSS` was not getting inherited for Resubmission wfs;
- when `CustodialSites`, `NonCustodialSites` or `AutoApproveSubscriptionSites` is inherited. The dropdown menu comes already selected and you cannot change that selection (only that option available).
- provides a multiple dropdown menu for `SiteWhitelist` and `SiteBlacklist`, instead of an input text box.

This does not fix the site list inherited for Resubmission though


amaltaro:
@ticoann please review. I tested all of this in my VM for ACDCs and new requests and it looks good.

ticoann:
sites() and pnn() probably should keep the cache so it won't fail in temporary failure for the siteDB call.  The problem with multiple choice and section window, it won't keep the all  original one ones in current code. I am not sure whether we can set multiple default values for select tag.

ticoann:
replaced by #7147

PR #7238, opened 2016-09-29
keyerror %d_ExeWCTime in DashboardInterface

juztas:
I have seen this printed for a long long time already... Lets fix it.

```
ERROR:root:Traceback (most recent call last):
  File "/condor/condor/dir_3852736/glide_yXLdYX/execute/dir_165130/job/WMCore.zip/WMCore/WMRuntime/Monitors/DashboardMonitor.py", line 135, in stepEnd
    stepReport = stepReport)
  File "/condor/condor/dir_3852736/glide_yXLdYX/execute/dir_165130/job/WMCore.zip/WMCore/WMRuntime/DashboardInterface.py", line 339, in stepEnd
    self.WrapperWCTime += data['%d_ExeWCTime' % self.stepCount]
KeyError: '2_ExeWCTime'
```

PR #7257, opened 2016-10-04
Better handling for step performance keys

amaltaro:
Complement #7242 and https://github.com/dmwm/WMCore/pull/7238

### DIFF
--- a/src/python/WMCore/WMRuntime/DashboardInterface.py
+++ b/src/python/WMCore/WMRuntime/DashboardInterface.py
@@ -9,13 +9,14 @@ import socket
 import subprocess
 import re
 
-from WMCore.WMSpec.WMStep     import WMStepHelper
+from WMCore.WMSpec.WMStep import WMStepHelper
 from WMCore.WMSpec.WMWorkload import getWorkloadFromTask
 from WMCore.Services.Dashboard.DashboardAPI import DashboardAPI
 from WMCore.Services.Dashboard.DashboardReporter import unicodeToStr
 from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig, SiteConfigError
 
-def getSyncCE(default = socket.gethostname()):
+
+def getSyncCE(default=socket.gethostname()):
     """
     _getSyncCE_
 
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

src/python/WMCore/WMRuntime/DashboardInterface.py
```diff
@@ -257,34 +262,34 @@ def stepStart(self, step):
...
-        data['MessageTS']   = time.strftime(self.tsFormat, time.gmtime())
-        data['taskId']      = self.taskName
-        data['jobId']       = self.jobName
-        data['%d_stepName' % self.stepCount]    = helper.name()
-        data['%d_ExeStart' % self.stepCount]    = helper.name()
+        data['MessageTS'] = time.strftime(self.tsFormat, time.gmtime())
+        data['taskId'] = self.taskName
+        data['jobId'] = self.jobName
+        data['%d_stepName' % self.stepCount] = helper.name()
+        data['%d_ExeStart' % self.stepCount] = helper.name()
```
ticoann: Why did we need the same value for different key? I just realized, should we need to add time faction instead of helper.name()?

ticoann (changes requested):

ticoann:
test this please

src/python/WMCore/WMRuntime/DashboardInterface.py
amaltaro: Looked like a bug for me, indeed. However, looking at the documentation/twiki page:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/InstrumentationOfWMAgentForDashboardReporting

it seems this is the correct input they expect..

ticoann: You are right. Although it is weird that we need two different key, I guess this is what was expected. Thanks for checking into

ticoann (approved):

amaltaro:
I'm just updating this error message to warning instead:
`logging.error('Failed to retrieve step cpu performance. Defaulting to 0')`

since performance info doesn't seem to be available for stageOut and logArch steps.

amaltaro:
@ticoann this one is also good to be merged

PR #7258, opened 2016-10-04
Parse out/err files instead of condor.log

amaltaro:
I finally found a concrete example that supports reading the tail of the `.out` and .`err` instead of the condor `.log` file.

I hit the same `99303 NoJobReport` exit code in testbed and looking at the condor log files:
- log extension: besides some condor classAds, it has no clear error message
- out extension: in general "good" output from the submit.sh, like

```
WMAgent bootstrap : Tue Oct  4 10:38:21 UTC 2016 : starting...
WMAgent bootstrap : Tue Oct  4 10:38:21 UTC 2016 : arguments validated...
```
- err extension: errors from the submit.sh, like

```
WMAgent bootstrap : Tue Oct  4 10:38:21 UTC 2016 : Error: VO_CMS_SW_DIR, OSG_APP, CVMFS environment variables were not set and /cvmfs is not present
WMAgent bootstrap : Tue Oct  4 10:38:21 UTC 2016 : Error: Because of this, we can't load CMSSW. Not good.
```

Thus, I suggest we pick 50/50, some out and some err lines, just to be on the safe side.
I still have to test it (and be lucky to see it happening again)

@ticoann should we pick only the last 25lines of each file instead? or 50 lines is Ok for each?
@hufnagel you may update the new plugin you're working on.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -128,6 +128,24 @@ def parseError(error):
     return errorCondition, errorMsg
 
 
+def parseCondorLogs(logfile, extension):
+    """
+    Retrieve the last X lines of the log file
+    """
+    errLog = None
+    logOut = ''
+
+    logPaths = glob.glob(logfile)
+    if len(logPaths):
+        errLog = max(logPaths, key=lambda path: os.stat(path).st_mtime)
+    if errLog is not None and os.path.isfile(errLog):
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, I think 50 is OK for that. However, I remember that some times condor.log has valuable message as well.

hufnagel:
Adjusted the SimpleCondorPlugin (printing out last 50 lines of the out, err and log file).

amaltaro:
I'm going to merge this one as well, since it affects only agent side.

PR #7262, opened 2016-10-05
Default EventsPerJob to None

amaltaro:
My bad, sorry! By setting the default value to 1000, it makes the automatic event splitting to be skipped.

This `None` would fail the `int` casting, however that is still not validated during request creation.
I'm planning to revisit that code in October anyways.

@ticoann please review. I still have to test it in my VM and see if Jenkins will complain


src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
```diff
@@ -17,6 +17,17 @@ class DataProcessing(StdBase):
     for this kind of WMSpec.
     """
 
+    def __init__(self):
```
amaltaro: They are, but pylint cannot see it and it complains that we are accessing these members "before" their definition. I'd rather keep it not to have false alarms in jenkins.

ticoann: I think in this case, we have ignore pylint. It seems all other Specs has the same structure and if we change only this one. It looks more puzzled when we look at the code.

ticoann:
Alan patch looks good to me.

amaltaro:
@ticoann my test workflows are still running, but I see that the job splitting is correct now.
Can you merge it - at your convenience but *not* at midnight please :) - and create a pre6 tag for cmsweb? I'll make the cmsdist PR tomorrow morning

PR #7265, opened 2016-10-06
fix assignment-approved reqmgr2

ticoann:
@amaltaro, Alan, please review. tested in my vm


amaltaro:
Thanks, Seangchan!
Patch looks good, I also ran some tests and it's ok, besides one issue that we do not need to fix right now. I'll create a new issue for that

PR #7267, opened 2016-10-06
 Set Input(Files,events,sizes,lumiCount) to the job attrributes. Fixes #7246

### LINKED ISSUE #7246: Set Input_(Files,events,sizes) to the job attributes where available
would be good to add this into job classads:
```
INPUT_FILES = "/store/data/Run2016B/DisplacedJet/RAW/v2/000/274/157/00000/DE308367-FE24-E611-A134-02163E011A12.root,N+1, N+2, N..."
INPUT_EVENTS = "6479,N+1,N+2, N..."
INPUT_FirstEvent = "0,N+1,N+2, N..."
INPUT_LastEvent = "0,N+1,N+2, N..."
INPUT_LumiCount = "10,N+1,N+2, N..."
INPUT_Sizes = "4270814155,N+1,N+2, N...
```
And also the block name, which job is analyzing, but it is separate ticket here: https://github.com/dmwm/WMCore/issues/6972

juztas:


amaltaro:
Justas, to give some context, what's the motivation for this change? What's this feature needed for?
For most of the cases it will not carry useful info since `first_event` and `last_event` is not set for lumi based splitting. Moreover, any utilitarian job will likely blow up the maximum jdl size.

Don't we already have this information (well, close to it) being reported by CondorService, from cmssw?

hufnagel:
Seems overkill to me as well.

hufnagel:
And yes, we have jobs with thousands of input files, this code won't work there.

juztas:
ok, thanks, lets close. will come back with block or are there also splitting between thousands of input blocks?

PR #7270, opened 2016-10-06
Handle workflows without inbox elements in JobArchiver

amaltaro:
I noticed there was a workflow in testbed not moving from aborted to aborted-completed and the reason is taht JobArchiver was failing to check the WMBS injection status (there was no inbox element for that request(?)).

In case the inbox element is gone, then we set this workflow as injected in the agent and get it out of the system.

@ticoann please review


amaltaro:
@ticoann please review and merge before cutting the 1.0.21 tag

PR #7271, opened 2016-10-06
Log the hostname and user in the condor logs

amaltaro:
wmagent branch patch to mitigate black hole node issues seen in production. Just move some prints right to the beginning of execution.

@ticoann please review, though it was tested in cmssrv214 already

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# should be a bit nicer than before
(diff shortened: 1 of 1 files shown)


amaltaro:
BTW, this is the patch we just discussed that we should push to production today.

ticoann:
Alan patch looks good.

amaltaro:
Thanks! I'm pushing it to production in 5min or so.

PR #7272, opened 2016-10-06
Parse out/err files instead of condor.log

ticoann:
#7258 for WMAgent branch


ticoann:
@amaltaro Alan,  I updated this to all production agents.
and restarted JobStatusLite. Thanks,

PR #7276, opened 2016-10-07
fix task chain assign page

ticoann:
One restriction it has is that when type the dictionary parameters for ProcessSting, etc. It should be valid json (cannot use single quote only double quote)


ticoann:
@amaltaro, Alan please review. Tested in my vm

amaltaro:
It looks good, Seangchan. Thanks!

PR #7283, opened 2016-10-12
Always override the PSet number of threads

amaltaro:
If requestors upload the PSet config to couch with a hardcoded numberOfThreads AND they do not provide `Multicore` parameter in the schema (or it's 1), then it is not overriden and those jobs will be created with the PSet config.

This makes sure that jobs will run with the correct number of cores, as provided in the schema or setting it to single core.

I'm applying it to cmssrv214, since Dave is running a test over there.
@hufnagel not sure it impacts any of the cores manipulations made by T0


amaltaro:
@ticoann it's good to be merged. We still have to backport it though

PR #7287, opened 2016-10-12
Support run whitelist/blacklist in Harvest algo

amaltaro:
While cooking new test templates, I noticed run whitelist/blacklist has no effect at all in the DQMHarvest workflows.

I still have some loggings on the Harvest.py code because I want to have a closer look at it tomorrow, just to make sure jobs are getting properly created.

Tests need to be more robust, specially about `multiRun` harvesting.


amaltaro:
@ticoann @hufnagel please review and merge (already backported to the agent branch)

PR #7288, opened 2016-10-12
add location filter to wmarchive format

ticoann:
fix what Valentine reported. 
@amaltaro, @vkuznet, Alan, Valentin, please review

```
and look what's wrong with the record you'll see that it is
SCHEMA ERROR PATH: "/steps/output/location" TYPE: "Basic data type" AREC: ["string", "null"] DATA: ["T2_CH_CERN"]

The message should be read as following:
- in a path: steps -> output -> location we have a data ["T2_CH_CERN"], but its
 avro type is either string or null.

If you'll look at our production schema:
https://github.com/dmwm/WMArchive/blob/master/src/python/WMArchive/Schemas/FWJRProduction.py#L60

the steps->output->location should be a string type not a list !!!
```

### DIFF
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -214,6 +214,12 @@ def convertOutput(outputList):
         if "dataset" in outDict:
             outDict.update(combineDataset(outDict["dataset"]))
             del outDict["dataset"]
+            
+        if "location" in outDict and isinstance(outDict["location"], list):
+            if len(outDict["location"]) > 0:
+                outDict["location"] = outDict["location"][0]
+            else:
+                outDict["location"] = ""
         
         _validateTypeAndSetDefault(outDict, STEP_DEFAULT['output'][0])            
             


yuyiguo:
@ticoann
While you apply this patch to all servers, can you add a patch to fix the time issue? 
@vkuznet Could you change some of the time values in HDFS docs systemically?

vkuznet:
Yuyi, I'm not sure what you mean by changing values in HDFS docs
systematically. Are you asking to change values in FWJR docs?
I don't think it is doable because we need to read doc(s) and write a new one.
We already have 35M docs on HDFS, are you asking to walk through all of
them and change some values in these docs?

The files on HDFS are files and it is not a DB where we can manipulate tables,
therefore if we need to change something it means:
- read a file
- change a value for all records in that file
- write new file
- delete original file

yuyiguo:
Thanks Valentin for the explanation. Let's talk this on the meeting.

amaltaro:
@ticoann even if it's not the final fix, feel free to merge it. It was backported to the wmagent branch

PR #7294, opened 2016-10-14
Improve submit.sh logging, exitCodes check

juztas:
a) Make sure script runtime is not less than set by a scheduler. Also it will force to sleep up to 5mins if this bootstrap script fails, which mainly means something is wrong with this WN or transferring files to it failed for unknown reason. Re-sending other jobs so fast to same worker node lead to same failure most often. if this MIN_JOB_RUNTIME_TIME_SECONDS is not set, will use 0. If Job exited 0 and bootstrap script also 0, it will not sleep any time and will exit.
b) Improve logging and separate what is where... Also to addition print the condor job ad content for debug reasons and also print environment more nicely.
c) output file variable is a string separated with spaces. Each of word is treated as a separate argument, so this shows a failure: [: -eq: unary operator expected

So this would be one of the steps for better debugging, one of the items here: #7224
Also I would propose to add a check on 68 & 137 jobrc code and print some system information, like ulimit, disk space, memory usage, but this would overload this pull request.

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -1,4 +1,15 @@
 #!/bin/bash
+# should be a bit nicer than before
+echo "======== WMAgent bootstrap STARTING at $(TZ=GMT date) ========"
+echo "User id:    $(id)"
+echo "Local time: $(date)"
+echo "Hostname:   $(hostname -f)"
+echo "System:     $(uname -a)"
+echo "Arguments:  $@"
+
+# Saving START_TIME and when job finishes END_TIME.
+WMA_MIN_JOB_RUNTIMESECS=300
+START_TIME=$(date +%s)
 
 # On some sites we know there was some problems with environment cleaning
 # with using 'env -i'. To overcome this issue, whenever we start a job, we have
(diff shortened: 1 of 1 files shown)


juztas:
More details here: https://github.com/dmwm/WMCore/pull/7236

amaltaro:
Fine work you got here Justas :) Seems to be working just fine on bad bootstrap cases as well:
```
======== WMAgent bootstrap STARTING at Mon Oct 17 10:50:35 GMT 2016 ========
...
======== WMAgent Python boostrap starting at Mon Oct 17 10:50:35 GMT 2016 ========
======== WMAgent final job runtime checks STARTING at Mon Oct 17 10:50:35 GMT 2016 ========
Mon Oct 17 10:50:35 GMT 2016: Job Runtime in seconds:  0
Mon Oct 17 10:50:35 GMT 2016: Job bootstrap script exited:  3
Mon Oct 17 10:50:35 GMT 2016: Job execution exited:
Mon Oct 17 10:50:35 GMT 2016: Job runtime is less than 300 seconds. Sleeping  300
======== WMAgent final job runtime checks FINISHED at Mon Oct 17 10:55:35 GMT 2016 ========
```

amaltaro:
@ticoann please merge

PR #7302, opened 2016-10-18
Import the stage out plugins

amaltaro:
Looking again at the error logs - for the new agent - we can see it's not a problem with xrdcp in the path, but instead with the plugins that were never registered in Registry class. Well, of course the xrdcp is an issue as well, but this is getting addressed somewhere else.

@emaszs I guess you need it for the *21 CRAB branch as well, no?


emaszs:
@amaltaro Yes, we do, I had a patch for this in crab but might aswell have it in the proper place.

amaltaro:
@ticoann please merge

PR #7310, opened 2016-10-19
only add document if the job doc exists

ticoann:
@amaltaro, Alan please review.  This will remove all the Junk document from wmstats.
We were pushing the association for all successful jobs.

### DIFF
--- a/src/couchapps/WMStats/updates/jobLogArchiveLocation.js
+++ b/src/couchapps/WMStats/updates/jobLogArchiveLocation.js
@@ -1,7 +1,6 @@
 function (doc, req) {
   if (!doc) {
-    doc = {};
-    doc._id = req.id;
+    return [null, 'ERROR'];
   };
   if (doc.logArchiveLFN === undefined) {
       doc.logArchiveLFN = {};


ticoann:
Before rotates the wmstats and tier0 wmstats it should be updated to all the agents to take the effect.

ticoann:
test this please

amaltaro:
I'm not sure I can review it :-)
Which junk information we were pushing before? Can you point me to an example?
And about `!doc`, is it a deleted doc that you are verifying here? I assume successful jobs have a deleted doc, don't they?

ticoann:
Hi Alan, we already discussed in the meeting but for the record.
Junk documents like this.
https://cmsweb.cern.ch/couchdb/_utils/document.html?wmstats/0000054c-2eac-11e6-b490-a0369f23cf8a-0

It is not deleted docs it is more like missing docs since we don't propagate successful job documents

amaltaro:
I see, so each of the 500 jobs that were collected in that logCollected job will have that mapping (and it's pushed to wmstats).
Question on the structure of this document, why a dictionary of `logArchiveLFN`? Can it have `False` value? I was just wondering whether it wouldn't be more efficient to make it either a simple string (if it's guaranteed there will be no more than 1 logArchive per job/doc) or a list.

ticoann:
I am not completely sure why the structure is dictionary (Although I wrote the code. :smile)
One thing is that structure is feed to javascript code which expect dictionary format (that code is used in many other data). I think I just didn't want to write the function for too many formatting function.) We can change that but we also rewrite the formatting function.

PR #7320, opened 2016-10-21
Handle workflows without inbox elements in JobArchiver - 1.0.19 wmagent version

amaltaro:
1.0.19_wmagent branch version of #7270 

I see vocms0304 getting bugged exactly with the same issue, so I'm pushing it there now and will check the other agents later to see if this needs to be pushed everywhere.


ticoann:
Don't we need 1.0.21 version?

amaltaro:
No need, the master version made it before we cut 1.0.21. Removing that label

PR #7322, opened 2016-10-24
AcqEra/ProcStr/ProcVer required during creation

### LINKED ISSUE #5373: make acquistion era, process string, processing version mandatory in reqmgr.
Currently this is not mandatory and created None string when it is not set. 
Disable default values and make this mandatory field for some one who assigned workflows

amaltaro:
Fixes #5373 
... and let the fun (unittests) begin.

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -229,15 +229,28 @@ def procversion(candidate):
     """ Integers """
     if isinstance(candidate, dict):
         for candi in candidate.values():
-            check(r'^[0-9]+$', candi)
+            check(r'^[0-9]+$', str(candi))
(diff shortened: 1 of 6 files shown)


amaltaro:
It seems these changes trigger some other DBS (or phedex) calls where we do not have data mocked.

I'm going to separate the dbs summary (for Dataset) and run list into two calls instead to see if that gets better.

amaltaro:
test this please

amaltaro:
It works fine in my VM and it enforces those parameters in ReqMgr2 (not reqmgr).

Note, however, I still have to create another PR for the assignment part. I noticed TaskChain workflows fail to get assigned if AcqEra is not provided.

Issue #7133, opened 2016-08-31
Make a list of accepted parameters for creation and another for assignment


amaltaro:
and enforce this list. Otherwise we end up with key/values in the schema (from creation time) that don't really set all the need properties (or for all the tasks). Then, if you don't provide them during assignment, it would default to the default values instead of those injected during creation.

In the end it brings us to a messy environment :(

ticoann:
@amaltaro, Alan if you could create the list, I can add in the validation.
Basically it is controlled by
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L882
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L884

vlimant:
+1

PR #7329, opened 2016-10-25
fix acdc creation

### LINKED ISSUE #7325: ACDC creation reqmgr1 vs reqmgr2
Trying to create a reqmgr2 ACDC for a worfklow in reqmgr1.
pdmvserv_EXO-RunIISpring16DR80-04597_00669_v0__160930_150303_8913
Using the same schema at injection
{
  "InitialTaskPath": "/pdmvserv_EXO-RunIISpring16DR80-04597_00669_v0__160930_150303_8913/StepOneProc", 
  "OriginalRequestName": "pdmvserv_EXO-RunIISpring16DR80-04597_00669_v0__160930_150303_8913", 
  "ACDCServer": "https://cmsweb.cern.ch/couchdb", 
  "Group": "ppd", 
  "RequestPriority": 70000, 
  "TimePerEvent": 3, 
  "ACDCDatabase": "acdcserver", 
  "RequestType": "Resubmission", 
  "SizePerEvent": 2500, 
  "Memory": 8192, 
  "RequestString": "ACDC_EXO-RunIISpring16DR80-04597_00669_v0_", 
  "PrepID": "EXO-RunIISpring16DR80-04597"
}

One ended up being 
https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/vlimant_ACDC_EXO-RunIISpring16DR80-04597_00669_v0__161025_154757_6084 (reqmgr2)
the other one
https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/vlimant_ACDC_EXO-RunIISpring16DR80-04597_00669_v0__161025_162428_4125 (reqmgr1)

so far so good. At assignment time, we try to get processingstring from the workload cache, and it is not present in the reqmgr2 case.

ticoann:
fixes #7325, #7333 @amaltaro, Alan please review. (Still need to test)

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -22,6 +22,9 @@ from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_START_STATE
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
 from WMCore.Lexicon import identifier
 
+ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST = \
+    ["_id", "_rev", "Requestor", "ReqMgr2Only", "VoRole", "RequestTransition", "RequestStatus", "DN", 
+     "TotalEstimatedJobs", "TotalInputEvents", "TotalInputLumis", "TotalInputFiles"]
 
 def initialize_request_args(request, config, clone=False):
     """
@@ -70,6 +73,15 @@ def initialize_request_args(request, config, clone=False):
         if "InputDataset" in request:
             request["InputDatasets"] = [request["InputDataset"]]
 
+def initialize_resubmission(request_args, config, reqmgr_db_service):
+    request_args["OriginalRequestCouchURL"] = '%s/%s' % (config.couch_host,
+                                                         config.couch_reqmgr_db)
+    requests = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
+    resubmission_args = requests.values()[0]
+    for arg in resubmission_args:
+        if (arg not in request_args) and (arg not in ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST):
+            request_args[arg] = resubmission_args[arg]
+    return request_args
 
 def generateRequestName(request):
     currentTime = time.strftime('%y%m%d_%H%M%S', time.localtime(time.time()))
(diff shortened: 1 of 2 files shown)


ticoann:
test this please

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -22,6 +22,9 @@
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
 from WMCore.Lexicon import identifier
 
+ARGS_TO_REMOVE_FROM_ORIGINAL_REUEST = \
```
amaltaro: typo

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -70,6 +73,15 @@ def initialize_request_args(request, config, clone=False):
         if "InputDataset" in request:
             request["InputDatasets"] = [request["InputDataset"]]
 
+def initialize_resubmission(request_args, config, reqmgr_db_service):
+    request_args["OriginalRequestCouchURL"] = '%s/%s' % (config.couch_host,
+                                                         config.couch_reqmgr_db)
+    requests = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
+    resubmission_args = requests.values()[0]
+    for arg in resubmission_args:
+        if (arg not in request_args) and (arg not in ARGS_TO_REMOVE_FROM_ORIGINAL_REUEST):
```
amaltaro: same here

amaltaro (commented):
And then we have to see how it's going to handle TaskChain requests... It will basically clone the original request but still make a Resubmission workflow, so should be fine.

amaltaro:
It seems this patch causes the same problem caused by the resubmit.py client used by P&R team, some time ago.
It inherits basically all the parameters, however some of those are supposed to be set only during assignment (e.g. Trust* flags, MaxRSS, etc). So if CompOps don't provide the Trust flags, they would expect it to be `False` (default value), but it could be `True` if original workflow was assigned with this set to True.

Two ways to fix it in a short term:
 1. expand the blacklist to include some of these keys (not all!):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/RequestStatus.py#L78
 2. instead of a black list, make it a white list of what to inherit (a bit tricky for TaskChains)

Longer term: https://github.com/dmwm/WMCore/issues/7133

PR #7331, opened 2016-10-26
1.0.21_crab version of Add DBS to CRABClient package dependencies

emaszs:
We're missing the DBSReader module because the prototype re-implementation of crab report queries DBS from the client rather than rely on the server to do the same thing.

master branch PR here: https://github.com/dmwm/WMCore/pull/7332


ticoann:
@emaszs, Emil 1.0.21.crab2 tag is created

emaszs:
Thanks!

PR #7343, opened 2016-10-31
Do not update workload cache couch doc with assignment dict parameters

### LINKED ISSUE #6881: ReqMgr2: filter out some keys before persisting them in the reqmgr2 couch doc
E.g.
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_TaskChain_MC_HG1604_Validation_ReqMgr2_160410_111249_3022

gets overwritten during assignment. For instance, we usually provide a dict for AcquisitionEra and ProcessingString, which then overwrite the basic string value by the dict value.

### LINKED ISSUE #6906: Upload new values to Multicore, Memory and Performance params to workload doc
This actually needs to be discussed whether we can completely overwrite the original values or whether we should somehow provide an API containing the exact parameters provided during request creation.

amaltaro:
Fixes #6906 
Fixes #6881

The idea is to skip dict values like AcquisitionEra, ProcessingString, ProcessingVersion, MaxRss and MaxVSize from the reqmgr2 db update, e.g.:
https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/amaltaro_TaskChainZTT_DeterPU_HG1611_Validation_161028_163752_4462

as you can see, `AcquisitionEra` was updated to a dict object value. If we, at any moment, need to clone this request, it would fail the create validation because that parameter is supposed to be a string.

@ticoann please review. Tested in my VM.

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -489,7 +489,11 @@ class Request(RESTEntity):
         # legacy update schema to support ops script
         loadRequestSchema(workload, request_args)
 
-        report = self.reqmgr_db_service.updateRequestProperty(workload.name(), request_args, dn)
+        # Do not update dict values to reqmgr2 workload cache, it's not properly formated.
+        # More info in the PR: #7343
+        reqArgsNoDict = {k: v for k, v in request_args.iteritems() if not isinstance(v, dict)}
+
+        report = self.reqmgr_db_service.updateRequestProperty(workload.name(), reqArgsNoDict, dn)
         workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)
         return report
 


ticoann:
Were we accepting dict before? As far as I remember, we accept both dictionary and string for task chain. (TaskChain could have 2 options (adding in Task or having in dictionary format)). Since the validation wasn't working properly before it was just passing through). I am fine with only accepting string but that means all the creation template and clone script is has to be changed accordingly) 
In that case, instead of filtering dict type we should throw the error if those parameters is updated with wrong format.

amaltaro:
Both string and dict are still accepted, the only difference is that we're not going to update the couchdb doc with the (new) values passed during workflow assignment. So, if anyone need to clone a TaskChain workflow, it will have the same original information as injected by McM.

And from your comment, I just noticed we should improve the validation and only allow it to be a dict for TaskChain requests. But this can be taken care in another issue/PR :)

amaltaro:
And this is why it was Ok in ReqMgr:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/HTTPFrontEnd/RequestManager/Assign.py#L422

we were not updating the couch doc with any possible new values. We were just reusing/getting the string value and sending it to the update function again.

ticoann:
Hi Alan, Actually the problem seems to be on due to double casting. validation cast correct type but __call__ in StdBase again casting dict type to str again. This doesn't seem to be related to assigning the workflow. I updated the patch #7347. And that seems to fix the test case you gave me.

The only thing is validation has to be run before it gets to pass to __call__ method to get the correct type casting. I am not sure Tier0 does that. (I think it is safe in Tier0 as well since there is no web interface on that).

amaltaro:
Seangchan made a proper fix in another PR. Closing this one

PR #7345, opened 2016-10-31
Make team name required during assignment; proper validation

### LINKED ISSUE #6917: Request assignment to > 1 team should fail
According to this constraint:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py#L238

requests will fail to get acquired by GQ. Instead, we should fail the assignment of such requests.
This is one example that was bugging GQ:
https://cmsweb.cern.ch/reqmgr2/data/request?name=mewu_RVCMSSW_8_0_10PhotonJets_Pt_10_13_160602_114021_8987

amaltaro:
Fixes #6917

Make sure a single team name (string) is set during the workflow assignment.

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -463,9 +463,8 @@ class Request(RESTEntity):
         
     def _handleAssignmentStateTransition(self, workload, request_args, dn):
         
-        req_status = request_args["RequestStatus"]
-        if req_status == "assigned" and not request_args.get('Team', '').strip():
-            raise InvalidSpecParameterValue("Team must be set during workflow assignment: %s" % request_args)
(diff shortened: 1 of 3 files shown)


amaltaro:
This is a tricky business...
The way the whole spec arg validation works is the following:
 a) it validates the assign_optional flag 
 b) then it validates whether the arg can be None or not
 c) then it tries to cast the original value
 d) then it performs the validation func on *the casted* value

At this moment, a possible list of teams became a string, which passes all further validations :-(

So my idea is to cast only the very basic data types (that wouldn't give problems), otherwise raise an exception. 

@ticoann do you think this is a reasonable approach? We actually may have to extend that cast function to other arguments in the near future, just to avoid problems when work is already in the agent or so.

ticoann:
duplicate

PR #7347, opened 2016-11-01
fix the validation for dictionary type

ticoann:
replace #7343


ticoann:
Current code allows dict type for all the parameters.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkloadTools.py#L157

I think we should allow only specific parameters for dict type.
i.e  we can create the new type dictOrStr so which accept str or dict format and type cast correctly.
What do you think?

amaltaro:
Creating a hybrid type would be a solution. Then we use this new type override only in the workload factory that is supposed to support it (TaskChain).

I've done the same in this open PR:
https://github.com/dmwm/WMCore/pull/7345/files#diff-16f3ba05fc754377cbc456e9ef920e38R65

if you have a better place to put it, just let me know.

PR #7352, opened 2016-11-01
Fix default task value for Multicore

### LINKED ISSUE #7350: Fix unitest.
WMCore_t.WMSpec_t.StdSpecs_t.TaskChain_t.TaskChainTests:testMultithreadedTasksTaskChain changed from success to failure

amaltaro:
Fixes #7350

It's a legit case, the test caught a bug that was going into production (thanks Seangchan!)

This fixes the problem, but it shouldn't be passing the validation function... which leads me to think we've never had a proper validation at task level. I'm debugging it further.


ticoann:
OK you did it first. :-)

PR #7321, opened 2016-10-24
Fix type in DQMHarvest; correct GQ stats accounting

amaltaro:
Several minor changes to the DQMHarvest spec:
- fixed typo (black_blacklist)
- use the correct splitting algo for the spec and start policy
- gather correct statistics at GQ level (not for events)
- added dataset name to the WMTask attributes (to avoid using PD,ProcStr,Tier all the time)

DQMHarvesting tests look Ok in my VM, still a wider test missing.
@ticoann please review

PR #7354, opened 2016-11-02
Fix type in DQMHarvest; correct GQ stats accounting - wmagent branch

amaltaro:
1.0.21_wmagent branch version of #7321
It needs to be pushed to production during cmsweb upgrade.


ticoann:
We need a new tag for WMAgent with this, right?

amaltaro:
Yes, but I guess we can wait a bit and try to add another bug fix for Pycondor.

amaltaro:
Seangchan, I'm merging it now since it was already pushed to all the production nodes.

PR #7361, opened 2016-11-03
make Campaign optional, Acquistion and ProcessingString mandatory

ticoann:
Alan, please review. (Tested in my vm)

1. If Campaign is set by McM it will take that data
2. If Campaign is not set by McM it will have the same value as AcquisitionEra unless AcquisitionEra is dictionary. In that case Campaign won't be set.

Also revert back to make AcquisitionEra and ProcessingString mandatory


src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -937,18 +937,16 @@ def getWorkloadArguments():
                                          "attr": "priority"},
                      "VoGroup": {"default": "unknown", "attr": "owner_vogroup"},
                      "VoRole": {"default": "unknown", "attr": "owner_vorole"},
-                     "Campaign": {"optional": False},
-                     "AcquisitionEra": {"default": "FAKE", "validate": acqname,
-                                        "assign_optional": False},
+                     "Campaign": {"optional": True}, # default value will be AcquisitionEra
```
amaltaro: This will default to `'default': None, 'null': False`, which means it cannot be None. Did you try a case whether there is no `Campaign` in the create dict?

src/python/WMCore/WMSpec/WMWorkloadTools.py
```diff
@@ -319,6 +319,12 @@ def setArgumentsWithDefault(arguments, argumentDefinition):
     for argument in argumentDefinition:
         if argument not in arguments and "default" in argumentDefinition[argument]:
             arguments[argument] = argumentDefinition[argument]["default"]
+            
+    # set the Campaign default value to the same as AcquisitionEra if Campaign is not specified
+    if ("AcquisitionEra" in arguments) and \
+       isinstance(arguments["AcquisitionEra"], basestring) and \
+        not arguments.get("Campaign", ""):
+        arguments["Campaign"] = arguments["AcquisitionEra"]
```
amaltaro: This line looks ugly :) How about:
```
if not arguments.get("Campaign"):
    if "AcquisitionEra" in arguments and isinstance(arguments["AcquisitionEra"], basestring):
```
?

I guess we cannot directly test `isinstance(arguments["AcquisitionEra"], basestring)` because during assignment this arg may not be there, right?

amaltaro: LoL

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
ticoann: Technically default value is not None but AcquisitionEra. 
I thought about it for the TaskChain case with AcquisitionEra different for each task, I think in that case it shouldn't set the default value to "" or None. I am not sure what would be the intention on that but I thought we shouldn't assume anything for that. So I thought it would be safer either they specify Campaign other wise not allow create the requests

ticoann:
test this please

amaltaro:
@ticoann , it seems all the problems are coming from my question above. Can you update the `Campaign` definition to:
`{'null': True}`
? that should fix the problem

PS.: These are the classic moments that I'm proud of saying "nothing is *just*"

ticoann:
Thanks Alan, I just set up the default value to empty string for dict type then.

amaltaro:
It looks good, thanks!

PR #7362, opened 2016-11-04
Dynamically resize jobs in PyCondor

amaltaro:
Superseeds #6786 (read the whole discussion/suggestions over there)

Cherry-picking Brian's changes was too problematic, so I applied that patch directly and updated what had to be updated.
The new changes are:
* applied some default values Dirk put in place a couple of months ago
* `Multicore` provided in the request is supposed to be the max numberOfCores. The min is supposed to be half of it (or 1)
* Fixed `UserLog` in the SimpleCondor plugin, it didn't have the absolute path.

@bbockelm please review just to make sure I did not make any mistakes.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -5,6 +5,8 @@ _PyCondorPlugin_
 Example of Condor plugin
 For glide-in use.
 """
+from __future__ import division
+
 import os
 import os.path
 import re
@@ -234,6 +236,7 @@ class PyCondorPlugin(BasePlugin):
         self.maxTaskPriority = getattr(config.BossAir, 'maxTaskPriority', 1e7)
         self.jobsPerWorker = getattr(config.JobSubmitter, 'jobsPerWorker', 200)
         self.deleteJDLFiles = getattr(config.JobSubmitter, 'deleteJDLFiles', True)
+        self.extraMem = getattr(config.JobSubmitter, 'extraMemoryPerCore', 500)
 
         # Required for global pool accounting
         self.acctGroup = getattr(config.BossAir, 'acctGroup', "production")
@@ -282,7 +285,7 @@ class PyCondorPlugin(BasePlugin):
             self.proxy = self.setupMyProxy()
 
         # Build a request string
-        self.reqStr = "(Memory >= 1 && OpSys == \"LINUX\" ) && (Arch == \"INTEL\" || Arch == \"X86_64\") && stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS==\"any\") || (GLIDEIN_REQUIRED_OS==REQUIRED_OS))"
+        self.reqStr = "(OpSys == \"LINUX\" ) && (Arch == \"INTEL\" || Arch == \"X86_64\") && stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS==\"any\") || (GLIDEIN_REQUIRED_OS==REQUIRED_OS))"
         if hasattr(config.BossAir, 'condorRequirementsString'):
             self.reqStr = config.BossAir.condorRequirementsString
 
@@ -486,7 +489,7 @@ class PyCondorPlugin(BasePlugin):
                 logging.error(msg)
                 continue
 
-            if not exitCode == 0:
+            if exitCode != 0:
                 logging.error("Condor returned non-zero.  Printing out command stderr")
                 logging.error(error)
                 errorCheck, errorMsg = parseError(error=error)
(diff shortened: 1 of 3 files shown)


ticoann:
This patch also need to added SimpleCondor as well, right?

amaltaro:
That's right! And possible another one that we still have to validate. I'd suggest to test SimpleCondor plugin as is now and then we can add these new features/fixes later

hufnagel:
Fine with me, but we shouldn't merge any changes that only touch PyCondorPlugin, just makes it harder to get both plugins synced later.

amaltaro:
@hufnagel @bbockelm I've just ported this feature into SimpleCondor plugin as well. Can you please x-check it to see if there was no mistake.

BTW, I noticed we have different requirement strings between those plugins.
PyCondor:
`"(OpSys == \"LINUX\" ) && (Arch == \"INTEL\" || Arch == \"X86_64\") && stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS==\"any\") || (GLIDEIN_REQUIRED_OS==REQUIRED_OS))"`

SimpleCondor
`"stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS=?=\"any\") || (GLIDEIN_REQUIRED_OS=?=REQUIRED_OS)) && (TARGET.Cpus >= RequestCpus)"`

It seems `OpSys` is not really used. Desired_Archs always have the same value, so in principle it's useless as well. How about the RequestCpus? Won't it evaluate the requirement expression before the `RequestCpus` one?

amaltaro:
test this please

amaltaro:
Ok, I fixed a couple of issues in the SimpleCondor plugin. Both plugins were tested and we have the resizing working just fine. Dashboard is also getting the proper values.

@bbockelm @hufnagel your last chance to make any suggestions on this PR :-)
@ticoann you're welcome to review it as well. Otherwise, please merge it by the end of your day.

amaltaro:
BTW, to get SimpleCondor plugin functional, we need #7390 as well

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -456,7 +457,7 @@ def getClusterAd(self):
         ad['Cmd'] = self.scriptFile
         ad['Out'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".out")')
         ad['Err'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".err")')
-        ad['UserLog'] = classad.ExprTree('strcat("condor.", ClusterId, ".", ProcId, ".log")')
+        ad['UserLog'] = classad.ExprTree('strcat(Iwd, "/condor.", ClusterId, ".", ProcId, ".log")')
```
bbockelm: For simplicity of comprehension, it might be a good idea to either apply the `Iwd` prefix to all three of `Out`, `Err`, and `UserLog` or to none.

hufnagel: @amaltaro , I had it for none of them and that worked as far as I remembered. Why did you add it for the UserLog ? Did you see problems with it ?

amaltaro: Not really a problem, but our `condorq` alias prints the `UserLog` ad in its output. Without the absolute path, we cannot easily see the condor logs location for that job.

I'll just set the absolute path to all of them then.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -953,18 +956,56 @@ def customizePerJob(self, job):
...
+        jdl.append('+EstimatedSingleCoreMins = %d\n' % estimatedMinsSingleCore)
+        jdl.append('+OriginalMaxWallTimeMins = %d\n' % estimatedMins)
+        jdl.append('+MaxWallTimeMins = WMCore_ResizeJob ? (EstimatedSingleCoreMins/RequestCpus + 15) : OriginalMaxWallTimeMins\n')
+
         requestMemory = int(job['estimatedMemoryUsage']) if job.get('estimatedMemoryUsage', None) else 1000
-        requestDisk = int(job['estimatedDiskUsage']) if job.get('estimatedDiskUsage', None) else 20*1000*1000*numberOfCores
-        maxWallTimeMins = int(job['estimatedJobTime'])/60.0 if job.get('estimatedJobTime', None) else 12*60
-        jdl.append('request_memory = %d\n' % requestMemory)
+        jdl.append('+OriginalMemory = %d\n' % requestMemory)
+        jdl.append('+RequestMemory = OriginalMemory + %d * (WMCore_ResizeJob ? (RequestCpus-OriginalCpus) : 0)\n' % self.extraMem)
```
amaltaro: that's actually not a bad idea :)

amaltaro:
And this is why I didn't set `Iwd` to `Out` and `Err`:
```
HoldReasonCode = 12
HoldReason = "Error from slot1@glidein_963163_862482878@cmswn2120.fnal.gov: STARTER at 131.225.189.168 failed to send file(s) to <128.142.141.177:4080>; SHADOW at 128.142.141.177 failed to write to file /data/srv/wmagent/v1.0.23.pre3/install/wmagent/JobCreator/JobCache/amaltaro_MonteCarlo_eff_JDLMagic_161207_193837_5536/Production/JobCollection_1183_0/job_5401/condor.12579.18.err: (errno 2) No such file or directory"
HoldReasonSubCode = 2
```

It looks for that file to transfer back from the scratch area, thus the absolute path would fail (hmmm, though there is no such file in the worker node, but it works)

amaltaro:
reading the error message, I may be wrong...

bbockelm:
@amaltaro - let's not block this issue in understanding the error message.  I would have expected that to happen only if the workflow was killed (removing the parent directory) but condor jobs not removed correctly - but let's move on.

It's also acceptable to me to make a comment in the code describing the problem and referencing your comment above!  I just want to make sure future coders to understand what we did.

amaltaro:
Ok, comment added. ExtraMemory ad added in both plugins.

Let's merge it and then we can fix what needs to be fix in the next release candidate.

bbockelm:
Indeed - any further work inside this PR will quickly become counterproductive.

PR #7363, opened 2016-11-04
Allow runtime environment to resize the cores in the job

amaltaro:
Superseeds #6757 (read the whole discussion over there)

Manual fix was only needed for SetupCMSSWPset.py , then I made a couple other aesthetic changes. This code is very verbose, but I think it's better to keep it all in the code for a while.

@bbockelm please review just to make sure there was no mistake on my side.

### DIFF
--- a/src/python/WMCore/WMRuntime/Watchdog.py
+++ b/src/python/WMCore/WMRuntime/Watchdog.py
@@ -105,18 +105,6 @@ class Watchdog(threading.Thread):
                 if changedCores and origMaxRSS and (resources['memory'] != origMaxRSS):
                     args.pop('maxVSize', None)
                     args['maxRSS'] = 1024 * (resources['memory'] - 50)  # Convert back to KB
-                # Scale our timeouts according to the scaling of the cores.  *NOTE*
-                # the job ad contains the real scaling here - unfortunately, it uses
-                # ClassAd expressions and we can only parse values.  Something for the
-                # future.
-                # As the scaling is not perfectly linear (efficiency is closer to 95%),
-                # we add a somewhat-arbitrarily-chosen 60 minutes (timeouts are in seconds).
-                if changedCores:
-                    coreScaling = resources['cores'] / origCores
-                    if 'softTimeout' in args:
-                        args['softTimeout'] = int(args['softTimeout']/coreScaling) + 3600
-                    if 'hardTimeout' in args:
-                        args['hardTimeout'] = int(args['hardTimeout']/coreScaling) + 3600
 
                 logging.info("Watchdog modified: %s. Final settings:", changedCores)
                 for k, v in args.iteritems():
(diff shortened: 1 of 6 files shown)


amaltaro:
I do have one concern though about the timeouts in the watchdog. When these jobs get pulled by a pilot, does it already considers a resizing related to the number of cores and pilot lifetime?

For instance, there is a 4 cores job (asking for walltime=24h) that is pulled by a 2 cores pilot, so running with 2 cores and theoretically for 48h. Is this pilot certain to have >48h of lifetime?

If the softTimeout was initially set to 36*3600 secs, then with the coreScaling the new timeout settings in the watchdog would be twice this value + 3600 secs (way beyond the pilot lifetime).

hufnagel:
Doesn't the matching to a pilot take into account the wall time adjustment due to matching to more or less cores than what was asked for ? I would consider this essential to deploy any kind of adjustable number of cores scheme...

hufnagel:
As regarding to the softTimeout, I always considered this a global limit, ie. to be independent of job settings, core count etc. It's a maximum we allow a job to run and we have to create jobs that are shorter than that (including all possible effects, of which adjusting the core count is just one).

amaltaro:
> Doesn't the matching to a pilot take into account the wall time adjustment due to matching to more or less cores than what was asked for ?

I'm pretty sure it does, Dirk. Anyways, better getting a confirmation from Brian or @Farrukh-Aftab

> As regarding to the softTimeout, I always considered this a global limit, ie. to be independent of job settings, core count etc. 

Exactly, that's my opinion as well.

bbockelm:
I can confirm that the pilot is responsible for making sure there is sufficient wall time, assuming that this code is in place:

https://github.com/dmwm/WMCore/pull/7362/files#diff-9b070481bcf1072545075872a49c7968R936

amaltaro:
Timeout removed in the second commit.

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -78,6 +81,34 @@ def setupMonitors(self, task, wmbsJob):
...
+                # off the top so watchdog triggers before HTCondor does.
+                if changedCores and origMaxRSS and (resources['memory'] != origMaxRSS):
+                    args.pop('maxVSize', None)
+                    args['maxRSS'] = 1024 * (resources['memory'] - 50)  # Convert back to KB
+
+                logging.info("Watchdog modified: %s. Final settings:", changedCores)
+                for k, v in args.iteritems():
+                    logging.info("  %s: %r", k, v)
             # Actually initialize the monitor variables
             mon.initMonitor(task = task, job = wmbsJob,
```
bbockelm: It seems that Dashboard gets its cores count from `src/python/WMCore/WMRuntime/DashboardInterface.py` which, in turn, pull the information from the task description (the `task` parameter here).

Unfortunately, the task description doesn't get its core count updated until later (`SetupCMSSWPset.py`).

I think we could set, after L107, an override for the number of cores.  This would be passed through `DashboardMonitor.py` to the class in `DashboardInterface.py` and override the `maxCores` member.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Ok, reporting to dashboard is fixed now, see: https://goo.gl/Ns0Yjl

If there are no further comments/suggestions on the cores logic (min, max, etc) (?), I'll port this fix to SimpleCondorPlugin tomorrow

amaltaro:
Same here, it's tested and seems to be working just fine.
@bbockelm @hufnagel @ticoann please review. You can merge it as well by the end of the day.

PR #7365, opened 2016-11-04
Adapt last templates to use the new required args during creation

amaltaro:
This PR is meant to address 2 things:
1) new required parameters during creation (acqera/procstr/etc)
2) updating releases that have been recently deprecated, see:
https://hypernews.cern.ch/HyperNews/CMS/get/relAnnounce/1783.html

This description will get updated as I go through these templates and make new commits:
*) properly handle the required creation args
*) delete templates using very old releases (actually, pre releases)
*) creating and validating new templates (TaskChain_ProdMinBias)
*) RelVal multicore template updated to a valid release
*) All DMWM templates were now updated
*) first half of the Integration templates (DQMHarvest, MC, MCFromGEN and ReDigi)
*) second half of the integration templates done (ReReco and TaskChain)

I think we have two missing templates for now, TaskChain reading lhe files and reco with skims.

BTW, a couple of these templates have a very very big PU sample - thus it was not migrated to dbs testbed - though the requests and central services work just fine.


amaltaro:
For the record, list of releases deprecated this week:
https://hypernews.cern.ch/HyperNews/CMS/get/relAnnounce/1783.html

I've just removed all templates using 700_pre releases. They will be replaced in the coming days.

test/data/ReqMgr/requests/DMWM/StepChain_MC.json
```diff
@@ -1,74 +1,82 @@
...
+            "StepName": "GENSIM"
+        }, 
+        "Step2": {
+            "CMSSWVersion": "CMSSW_8_0_21", 
+            "ConfigCacheID": "f7e311f2c6b5a0884faea133990edcbf", 
+            "GlobalTag": "80X_mcRun2_asymptotic_2016_TrancheIV_v6", 
+            "InputFromOutputModule": "RAWSIMoutput", 
+            "InputStep": "GENSIM", 
+            "KeepOutput": false, 
+            "MCPileup": "/Neutrino_E-10_gun/RunIISpring15PrePremix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v2-v2/GEN-SIM-DIGI-RAW",
```
amaltaro: I just noticed this awesome Stepchain template is using the evil PU sample :( I'm updating it soon.

amaltaro:
That should be basically it (besides the taskchain lhe). Now I have to update a couple of unittests that are using these json templates (I guess I'll just decouple unittests from them) and that's it.

amaltaro:
Cool! In the end I had to keep the evil Neutrino PU dataset (> 100k files), though I didn't import it to dbs testbed. I haven't seen any problem with the agents while parsing/querying for that PU dataset as well.

We likely have to update some minor settings/comments, but I think we should merge it and if needed I can create another PR addressing those few issues. @ticoann what do you think?

ticoann (approved):

PR #7368, opened 2016-11-04
Adapting inject-test script to required args

amaltaro:
Mostly tested, but better not merge it now because I still have to make my mind on how to proceed with assignment. I think we have to make ~half of the Integration templates with the assignment information (AcqEra,...) and the other half just using what was injected at creation....


amaltaro:
Code updated and tested.
@ticoann please merge

PR #7376, opened 2016-11-07
Fix T0 ProcessingString regex

amaltaro:
ProcString can be an empty string for T0, this regex should accept that.
It also fixes these two tests
```  WMComponent_t.AnalyticsDataCollector_t.Plugins_t.Tier0Plugin_t.Tier0PluginTest:testC_PromptRecoStates changed from success to error
    WMCore_t.WMSpec_t.StdSpecs_t.PromptReco_t.PromptRecoTest:testPromptReco changed from success to error
```
that failed in #7375 

I tested it interactively and online, but @ticoann , please review because my regex skills are not good :)

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -253,10 +253,10 @@ def procstringT0(candidate):
     """
     if isinstance(candidate, dict):
         for candi in candidate.values():
-            check(r'[a-zA-Z0-9_]{1,100}$', candi)
+            check(r'^$|[a-zA-Z0-9_]{1,100}$', candi)
         return True
     else:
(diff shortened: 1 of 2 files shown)


ticoann:
Hi Alan, But it can't be empty for Production and Processing, right? I think we should have separate lexicon (validation) for Tier0.

amaltaro:
Yes, we created a different lexicon a couple of weeks ago `procstringT0` :-)

Interesting that all these other tests changed from error from success, even so they are not affected at all by this change(?)

ticoann:
That is weird. Was there a case some test workflow doesn't have process string on failed case. I will investigate. So is this patch needed except unittest fix?

amaltaro:
test this please

PR #7378, opened 2016-11-08
Allow jobs to request repack slots

### LINKED ISSUE #7372: Creation of Repack slot
Hello:

We want to have a particular slot to run Repack jobs (similar to what we do for the HighIO slots). Farrukh already implemented and tested this on the Condor side. 

The name of the new classAd is 'RequestRepackslots'.

Please remove Repack from the list of jobs sent to the HighIO slots (https://github.com/dmwm/WMCore/blob/6f580f444dca6ed8d3f1c1c34680b197bfeaf574/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L332). 

Also, Repack should be the only type of job using the Repack slot.

To see the details of the condor implementation, please check:
https://cms-logbook.cern.ch/elog/GlideInWMS/4562

We would like to have these changes ASAP, in order to test and deploy them. 

Thanks a lot.

amaltaro:
Fixes #7372

Since it relies only on the job type, I added this logic only to the plugins instead of adding it to the RunJob object structure. We could actually do the same thing with the highio flag IMO.

@johnhcasallasl please review and test it in a replay

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -327,11 +327,6 @@ class JobSubmitterPoller(BaseWorkerThread):
                 numberOfCores = getattr(baggage, "numberOfCores", 1)
             loadedJob['numberOfCores'] = numberOfCores
 
-            # check if job should be submitted as highIO job
-            highIOjob = False
-            if newJob['type'] in ["Repack", "Merge", "Cleanup", "LogCollect"]:
-                highIOjob = True
-
             # Create a job dictionary object and put it in the cache (needs to be in sync with RunJob)
             jobInfo = {'id': jobID,
                        'requestName': newJob['request_name'],
@@ -341,7 +336,6 @@ class JobSubmitterPoller(BaseWorkerThread):
                        'priority': newJob['wf_priority'],
                        'taskID': newJob['task_id'],
                        'retry_count': newJob["retry_count"],
-                       'highIOjob': highIOjob,
                        'taskPriority': None,                                # update from the thresholds
                        'custom': {'location': None},                        # update later
                        'packageDir': batchDir,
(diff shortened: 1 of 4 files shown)


hufnagel:
Looks good. As for simplifying the code, as you have to modify JobSubmitter files anyways, any reason not to simplify it completely here ? Are all the taskTypes we need available in the plugin ?

amaltaro:
taskType is an attribute of the object. So its value is always there.
Are you saying that we could get rid of the highio attribute and just do this handling on the plugin side? If so, sure I can put it here as well.

hufnagel:
If it works for repackslots I see no reason it wouldn't work for ioslots. Two different ways to implement the same thing is just confusing...

amaltaro:
@johnhcasallasl let me know the outcome of your replay. Then I can proceed and make these minor changes to highio, as suggested by Dirk.

johnhcasallasl:
@amaltaro, It was tested yesterday and seems to be working fine. Please do the changes for the next test that you want to do and let us know.

amaltaro:
@johnhcasallasl I have just pushed a second commit with the highio changes. Can you apply the whole patch (both commits) and run another replay please. If it succeeds, then I'll squash it and we're done.

johnhcasallasl:
The tests showed that the slots are working properly. In the replays everything was fully processed and we didn't observed any misbehaviour.

amaltaro:
John confirms it works, so I'm going to squash it and it's ready to be used in T0 and merged into master.

amaltaro:
@ticoann @hufnagel please review and merge
@johnhcasallasl feel free to apply it to the T0 node at your convenience (since you already tested it, I didn't :-))

hufnagel:
Looks good ! And it should be safe to apply this patch as it'll work with already created jobs that were created with the old JobSubmitter code. The Plugin will just ignore the existing highio field and look at the taskType instead.

PR #7379, opened 2016-11-09
fix setup-dependencies for ASO deployment

dciangot:


dciangot:
ASO deployment with this patch is working without issues on test VM and then on ASO testbed machine. So as far as I can tell it is ready to be merged.

PR #7382, opened 2016-11-11
Fix for WMBS job mask in oracle

### LINKED ISSUE #7377: Upload of MC ACDC docs broken
I've just found a MC workflow with 1 out of 20 jobs failed. The ACDC workflow produced exactly the same lumi + number of events. The problem lies on the jobfailed information uploaded to couchdb acdcserver:
https://cmsweb-testbed.cern.ch/couchdb/acdcserver/_design/ACDC/_view/byCollectionName?key=%22amaltaro_MonteCarlo_PhEDEx_HG1611_Validation_pre6_161105_103000_3637%22&include_docs=true&reduce=false

where one can see the list of lumis correspond to ALL the lumis this workflow is supposed to create.
The same applies to the file nunmber of events (10k, so total number of events), though it's known and expected already..

Needs to be investigated further, on the agent side.

amaltaro:
Fixes #7377
The real fix was just the `job = :jobid` to the select query. This means if there was any other workflow (job) using exactly the same job splitting and having exactly the same job mask, we were not adding a new job for it.

It shouldn't cause too much trouble in production since there is a variety of workflows, but in our tests we recycle workflows, that's why it was spotted.

@ticoann @yuyiguo please review. I'm gonna run one more test with these final changes.

### DIFF
--- a/src/python/WMCore/WMBS/Oracle/Masks/Save.py
+++ b/src/python/WMCore/WMBS/Oracle/Masks/Save.py
@@ -7,33 +7,30 @@ Oracle implementation of Masks.Save
 
 __all__ = []
 
-
-
 from WMCore.WMBS.MySQL.Masks.Save import Save as SaveMasksMySQL
 
+
 class Save(SaveMasksMySQL):
-    sql = """INSERT INTO wmbs_job_mask (job, firstevent, lastevent, firstrun,
-               lastrun, firstlumi, lastlumi, inclusivemask)
-             SELECT :jobid, :firstevent, :lastevent, :firstrun, :lastrun, :firstlumi,
-               :lastlumi, :inclusivemask FROM dual
-             WHERE NOT EXISTS (SELECT job FROM wmbs_job_mask wjm2 WHERE wjm2.firstevent = :firstevent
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

amaltaro:
@ticoann please review and merge. Already tested in testbed

ticoann:
@amaltaro, Alan, I was wondering what cause this duplicate record happens in the first place?

amaltaro:
There was "duplicate" records because the `job` id was not included in the wmbs_job_mask select, meaning that any other workflow using exactly the same job splitting (e.g. MC from scratch) would have the same "*event", "*run" and "*lumi" information.

Now duplicates are not going to happen because we also check if the job id + mask already exist.

ticoann:
I mean I don't understand the cause why duplicate is happening with the same job id. 
Is there something wrong with splitting algorithm or is it some inconsistency can happen with db update (not properly rolls back).

amaltaro:
I don't really know. This uniqueness probably precedes my life within wmcore. And since it works "just fine", I'm happy with that :P

PR #7391, opened 2016-11-15
Change PU file list behavior during WMRuntime/TrustPUSitelists

### LINKED ISSUE #7387: Bypass any sites when TrustPUSitelists True
At the moment, this flags adds all the PNNs (mapped from the SiteWhitelist value) to the pileup json file.

We could change this feature such that if jobs are overflowed to other sites, the runtime scripts would see those blocks as present at that sites as well. 

We likely want to add a job baggage and treat it at SetupCMSSW script.

amaltaro:
Fixes #7387

Summary of the changes and the expected behaviour is as follows:
* pileup json file contains only real data/location, no more fake pnns.
* `TrustPUSitelists` changes the behaviour of two things, when set to `True`.
  - used for bypassing GQ to LQ replication.
  - makes sure **all** the PU files are added to the PSet as secondary input, no matter the site and/or data location
* if job lands on a site that contains only 1 PU block out of 100, the PSet will have only files that belong to that block (unles Trust flag was set to True) 
* Needless to say, if the PU dataset is wiped out from the storage by DDM (or anything else), ALL the jobs will read the PU data via AAA (depending on the workflow, it can cause a very bad impact on the grid resources (AAA, storage)) 

I'm not testing it before I get a confirmation whether this is the desired behaviour or not.
So @ticoann @hufnagel @vlimant @ericvaandering please share your thoughts (Dirk, only short comments please :), otherwise it's impossible to have a productive follow up)

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -12,7 +12,7 @@ import random
 import re
 import socket
 import traceback
-from json import JSONDecoder
+import json
 
 import FWCore.ParameterSet.Config as cms
 
@@ -405,10 +405,13 @@ class SetupCMSSWPset(ScriptInterface):
         If other pileup types are specified by the user, the method doesn't
         modify anything.
 
-        The method considers only files which are present on this local
-        PNN (PhEDExNodeName). The job will use only those. Dataset, divided into
-        blocks, may not have all blocks present on a particular SE. However,
-        all files belonging into a block will be present when reported by DBS.
+        The method considers only files which are present on this local PNN.
+        The job will use only those, unless it was told to trust the PU site
+        location (trustPUSitelists=True), in this case ALL the blocks/files
+        will be added to the PSet and files will be read via AAA.
+        Dataset, divided into blocks, may not have all blocks present on a
+        particular PNN. However, all files belonging into a block will be
+        present when reported by DBS.
 
         The structure of the pileupDict: PileupFetcher._queryDbsAndGetPileupConfig
 
@@ -418,6 +421,8 @@ class SetupCMSSWPset(ScriptInterface):
         each type of modules instances can have either "secsource"
         or "input" attribute, so need to probe both, one shall succeed.
         """
+        baggage = self.job.getBaggage()
+
         for m in modules:
             for pileupType in self.step.data.pileup.listSections_():
                 # there should be either "input" or "secsource" attributes
@@ -427,18 +432,16 @@ class SetupCMSSWPset(ScriptInterface):
                     continue
                 inputTypeAttrib.fileNames = cms.untracked.vstring()
                 if pileupType == requestedPileupType:
-                    # not all blocks may be stored on the local SE, loop over
-                    # all blocks and consider only files stored locally
                     eventsAvailable = 0
+                    useAAA = True if getattr(baggage, 'trustPUSitelists', False) else False
                     for blockName in sorted(pileupDict[pileupType].keys()):
                         blockDict = pileupDict[pileupType][blockName]
-                        if PhEDExNodeName in blockDict["PhEDExNodeNames"]:
+                        if PhEDExNodeName in blockDict["PhEDExNodeNames"] or useAAA:
                             eventsAvailable += int(blockDict.get('NumberOfEvents', 0))
                             for fileLFN in blockDict["FileList"]:
                                 # vstring does not support unicode
                                 inputTypeAttrib.fileNames.append(str(fileLFN['logical_file_name']))
                     if requestedPileupType == 'data':
-                        baggage = self.job.getBaggage()
                         if getattr(baggage, 'skipPileupEvents', None) is not None:
                             # For deterministic pileup, we want to shuffle the list the
                             # same for every job in the task and skip events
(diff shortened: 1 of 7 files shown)


hufnagel:
Ok for me ! Short enough :-)

amaltaro:
Added a last bullet, likely the most important one.

hufnagel:
Btw, this will change the behavior to what we do now. With faking the location using TrustPUSitelists you always had all blocks being present in full at the sites you submitted the workflow to, right ? Which meant it would use the complete list of pileup data for the job, whether it was locally available or not.

Only using what is really locally available is better in terms of efficiency, but is this really what Ops wants ? Question for @vlimant I guess.

amaltaro:
test this please

ericvaandering:
"Only using what is really locally available is better in terms of efficiency, but is this really what Ops wants ? "

Same question for me. Should a lot of jobs land at a site with only a small amount of pileup, you will skew the statistics by reusing those events more than they should be on average.

vlimant:
The desired behavior is, as I thought was described in the original issue
Trust=true job see all files of the PU dataset
Trust=false the file location is determined and things are ran from local

hufnagel:
Ok, but how do you want to handle overflow ?  For Trust=true even overflowed jobs see all the pileup ? For Trust=false never allow overflow (since the overflowed site would not have anything local) ?

ticoann:
> The desired behavior is, as I thought was described in the original issue
> Trust=true job see all files of the PU dataset
> Trust=false the file location is determined and things are ran from local

It sounds like Trust flag controls whether it takes all the files in PU or only files in the local  site.
(So Trust flag is not just used in GQ ->LQ, it has to be also used in Runtime to determine to get all the files or local files)
There are 2 cases we have to determine the behavior.
1. Trust flag is false but PU data is not exists in the local site. (not overflow case) 
2. As Dirk said (overflow case)

If the flag mean that way, how ops determines whether flag is set to True or False.
Do they know ahead whether sample PU size is big enough in local sites?

amaltaro:
> The desired behavior is, as I thought was described in the original issue
> Trust=true job see all files of the PU dataset
> Trust=false the file location is determined and things are ran from local

The second commit should give you this behaviour.

Jean-Roch should reply here, but I think Ops/Unified does know when it will need to read PU remotely and in such cases the TrustPU flag is set to True.
IF it's not and there is no local PU data, jobs will fail.

This is not tested, but I'm happy to hear comments/review :)

vlimant:
Hi @ticoann , we do check presence at sites in phedex beforehand and do decide how to use reqmgr/wmagent accordingly.

Hi @hufnagel , trust=true all jobs sees all pu files. trust=false, jobs sees the local file, and if overflown it should see the files that are local at the site it is overflown to, and see this dynamically of possible (use case is replicating the PU after assignment to overflow)

vlimant:
@amaltaro is that last commit making overflowing on premix possible ? i.e. if a job lands at a site outside of the whitelist, does it still see the proper list of PU.

ticoann:
@amaltaro, Alan, I know you are still updating the patch as the discussion goes. 

for overflowing case and Trust flag is false.
i.e Trust =False and no data exist in current site.
We randomly pick the site with in the which PU is originally existed and only add the blocks belongs to that instead of all the blocks. Is that good enough?

hufnagel:
I think if you want to overflow job with trust = false and have them be able to read local pileup, you need to attach the full mapping of pileup to PNN to the jobs.

OTOH, for workflows with trust = true, since we don't use the mapping at all, we could just leave it out in principle (unless that causes problems with some other code ?).

vlimant:
>  i.e Trust =False and no data exist in current site.
>  We randomly pick the site with in the which PU is originally existed and only add the blocks belongs to that instead of all the blocks. Is that good enough?

on ops end (modulo mistakes) we would only overflow trustpu=False to site where we know the PU is at, either prior to assignment, or after assignment. so what you discribed would be good enough, best of course would be to have the mapping be more dynamic.

hufnagel:
Looks like we have agreement :-)

hufnagel:
@vlimant , some/most of these changes will only affect new workflows though. Is that going to be a problem ?

amaltaro:
Unless we have super ultra critical workflows in production AND that will reamain running for several weeks (> 2 weeks at least), I'm not going to mess around with the workflow sandboxes in ALL the agents.

> at, either prior to assignment, or after assignment. so what you discribed would be good > enough, best of course would be to have the mapping be more dynamic.

**prior** assignment. Once the block/location pileup json is created, it goes into the sandbox and it will not consider any new PU relocation.

vlimant:
ops will assess and restart those in critical situation, no need to tamper with sandbox for now.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -405,10 +405,13 @@ def _processPileupMixingModules(self, pileupDict, PhEDExNodeName, modules, reque
...
         modify anything.
 
-        The method considers only files which are present on this local
-        PNN (PhEDExNodeName). The job will use only those. Dataset, divided into
-        blocks, may not have all blocks present on a particular SE. However,
-        all files belonging into a block will be present when reported by DBS.
+        The method considers only files which are present on this local PNN.
+        The job will use only those, unless there isn't a single block locally
+        available, in this case ALL the blocks/files will be added to the PSet
+        and files will be read via AAA.
```
bbockelm: The above comment no longer matches the implemented code.

amaltaro: That's why comments are usually evil. Good catch

amaltaro:
@ticoann please review and merge (the agent version was just pushed to the agents)

PR #7369, opened 2016-11-04
add some Physics Skims for 2016 PA run

hufnagel:
Will be deployed in the Tier0 within the next few days (pending confirmation that everything works in replays).

PR #7392, opened 2016-11-15
Added physics skims for 2016 PA run; separate skim mapping

amaltaro:
Superseeds #7369

@hufnagel please review
@ticoann please merge


hufnagel:
Looks good, but likely won't be used in the HI run since my previous patch is already used in the Tier0.

amaltaro:
Dirk, I've just read "Should not be merged until code has been validated in Tier0.".
So, shouldn't we merge it?

hufnagel:
Code has been validated in replays and works, this is ok. We use the old PR, but this new code also looks good.

PR #7402, opened 2016-11-19
treat all the request as reqmgr2

ticoann:
#replace 7401 (need to create the 1.0.22_cmsweb branch and 1.0.21_wmagent branch)

PR #7407, opened 2016-11-21
1.0.21_wmagent version of #7402

ticoann:


ticoann:
updated all the production agents.

PR #7409, opened 2016-11-22
Fix multicore settings for PromptReco

### LINKED ISSUE #7408: fix PromptReco AlcaSkim task to use single core
https://github.com/dmwm/WMCore/pull/7395

changed how number of cores has to be set inside a specs for individual tasks. Needs to use taskConf now.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py#L85

Also should remove the useMulticore parameter from setupProcessingTask since it is no longer used.

amaltaro:
Fixes #7408

My goal with the latest changes in the Multicore settings is that a `Multicore` key/value at the top level should be propagated to all task types except LogCollect/Cleanup/Merge/Harvesting.
Since alcaSkim has to be single core, then we provide a new task value instead of using the spec `multicore` attribute.

Dirk, as a bonus you got a unit test (testPromptRecoWithSkims) fixed :-) It was looking for a task that isn't created in the spec.
@hufnagel @ticoann please review. I still want to write similar unit tests to the other specs though

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
@@ -75,7 +75,7 @@ class PromptRecoWorkloadFactory(StdBase):
 
             else:
                 alcaTask = recoTask.addTask("AlcaSkim")
-
+                alcaTaskConf = {'Multicore': 1}
                 scenarioArgs = { 'globalTag' : self.globalTag,
                                  'skims' : self.alcaSkims,
                                  'primaryDataset' : self.inputPrimaryDataset }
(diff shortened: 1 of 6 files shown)


amaltaro:
The only WMCore specs without Memory/Multicore unit tests now are StoreResults and DQMHarvest. I noticed we don't set cores nor memoryRequirements for Harvest tasks. I'm not sure we'll ever have multicore harvesting jobs and or requiring more than the default amount of memory (1000MB AFAIK).

@ticoann @hufnagel @ericvaandering any insights here?

amaltaro:
@ticoann please review and merge. I didn't run too many tests, but basic stuff seems Ok.

hufnagel:
Looks good to me too.

PR #7412, opened 2016-11-22
switch repack to lazy-download and multi-core

hufnagel:
Configure repack to use lazy-download (repacking high event rate streams is very slow from EOS) and allow repack to use multi-core in case it already has multiple core slots assigned to it (because of disk space).

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -533,14 +533,12 @@ class SetupCMSSWPset(ScriptInterface):
         """
         _handleRepackSettings_
 
-        Disable lazy-download for repacking (no benefit on streamer files).
+        Repacking small events is super inefficient reading directly from EOS.
         """
         print("Hardcoding read/cache strategies for repack")
         self.process.add_(
             cms.Service("SiteLocalConfigService",
-                        overrideSourceCacheHintDir = cms.untracked.string("storage-only"),
-                        overrideSourceReadHint = cms.untracked.string("read-ahead-buffered"),
-                        overrideSourceTTreeCacheSize = cms.untracked.uint32(20*1024*1024)
+                        overrideSourceCacheHintDir = cms.untracked.string("lazy-download")
                         )
             )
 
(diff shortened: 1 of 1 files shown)


src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -633,7 +631,7 @@ def __call__(self):
             if funcName == "repack":
                 self.handleRepackSettings()
 
-            if funcName in ["repack", "merge", "alcaHarvesting" ]:
+            if funcName in ["merge", "alcaHarvesting" ]:
```
amaltaro: Dirk, you can actually remove these two lines. I put a generic core override (no matter the task, it will always set numberOfCores) 
https://github.com/hufnagel/WMCore/blob/e8d45081f5995a4934499e09dd8c45ced4bdac52/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L658

and missed this one.

hufnagel: Ok, I'll need to change your code then. There are merge and alcaHarvesting jobs that are multicore for HTCondopr, but should be singlecore for CMSSW.

amaltaro: Can you remind me why they are multicore for HTCondor but must run single core? Is it about job requirements that are put in the jdl? If so, then we should try to make these changes at spec level and make sure the job/performance requirements are set since the beginning and avoid hacking around during job submission or runtime.

hufnagel: They need more than 20GB of disk. There is nothing else we can do, they need to get multicore slots for that.

hufnagel: Periodic remove expression in the pilots has a ceiling of 20GB per core and nothing in the jobs classads can change that.

hufnagel: Ok, reviewed code, nothing to change. It overrides the right place, your code sees the changes value.

amaltaro:
Code looks good to me. @ticoann feel free to review and merge.

hufnagel:
Can you hold off merging for a bit. I am testing this code in Tier0 replays and it works fine, but performance gains are not as expected. Need to discuss this a bit before I am comfortable merging this...

hufnagel:
On second thought, this is much much better than before already, so feel free to review and merge. We won't use it in the Tier0 until next year anyways, there will be a lot of testing before then...

amaltaro:
@ticoann it's also good to be merged from my side. In the worst case it will cause problems to repack jobs only (T0) :-)

PR #7420, opened 2016-11-24
fix wrong equation for JobPrio calculation in SimpleCondorPlugin

hufnagel:
Accidentally swapped two parameters, fix this.


hufnagel:
@amaltaro , please review and merge. Simple fix, don't think it needs much discussion :-)

amaltaro:
Good catch!

PR #7390, opened 2016-11-14
Condor classAd doesn.t like unicode

amaltaro:
I had a couple of issues in my first test with this plugin:
1. python args not matching the c++ signature in the `convertFromUnicodeToStr` func, see [1] for more details
2. cannot set any classad parameter with a unicode value, see [2]. Python3 will default to unicode utf-8, so I wonder whether condor should be able to treat these cases, @bbockelm ?

So my initial proposal here was to make a python dict object with the classAds and only in the end we cast any unicode to ascii string. 

I'm happy to further discuss and improve it though, I just needed to get my tests ongoing.

[1]
```
2016-11-14 16:49:04,965:139808333096704:INFO:JobSubmitterPoller:Done assigning site locations.
2016-11-14 16:49:04,991:139808333096704:ERROR:BossAirAPI:Unhandled exception while submitting jobs to plugin: SimpleCondorPlugin
Python argument types in
    ClassAd.__init__(ClassAd, list)
did not match C++ signature:
    __init__(_object*, boost::python::dict)
    __init__(_object*, std::string)
    __init__(_object*)
2016-11-14 16:49:05,067:139808333096704:ERROR:BaseWorkerThread:Error in worker algorithm (1):
Backtrace:
  <WMComponent.JobSubmitter.JobSubmitterPoller.JobSubmitterPoller instance at 0x7f27ab43a4d0> BossAirException
Message: Unhandled exception while submitting jobs to plugin: SimpleCondorPlugin
Python argument types in
    ClassAd.__init__(ClassAd, list)
did not match C++ signature:
    __init__(_object*, boost::python::dict)
    __init__(_object*, std::string)
    __init__(_object*)
        ModuleName : WMCore.BossAir.BossAirAPI
        MethodName : submit
        ClassInstance : None
        FileName : /data/srv/wmagent/v1.0.22.pre10/sw/slc6_amd64_gcc493/cms/wmagent/1.0.22.pre10/lib/python2.7/site-packages/WMCore/BossAir/BossAirAPI.py
        ClassName : None
        LineNumber : 433
        ErrorNr : 0

Traceback: 
  File "/data/srv/wmagent/v1.0.22.pre10/sw/slc6_amd64_gcc493/cms/wmagent/1.0.22.pre10/lib/python2.7/site-packages/WMCore/BossAir/BossAirAPI.py", line 420, in submit
    info=info)

[...]

PR #7421, opened 2016-11-24
SimpleCondorPlugin job submission Unicode problems

hufnagel:
@ticoann unicode conversion code doesn't work for me, I get job submission errors:

Python argument types in
    ClassAd.__init__(ClassAd, list)
did not match C++ signature:
    __init__(_object*, boost::python::dict)
    __init__(_object*, std::string)
    __init__(_object*)

Don't want to merge it like this, but lets discuss what needs to be done in the PR and I'll modify.


hufnagel:
This change makes submission work in Tier0 replays, without it I get the errors mentioned.

amaltaro:
I have a similar fix here:
https://github.com/dmwm/WMCore/pull/7390/files

that was actually waiting for your comment :-)

If it didn't fail for you, then `sites` in the T0 setup must be a ascii string instead of unicode

hufnagel:
Closed in favor of #7390

PR #7423, opened 2016-11-24
SimpleCondorPlugin misses an acct group classad for submits

hufnagel:
Please review and merge, just found in testing (and discussion with Farrukh).


amaltaro:
Thanks, Dirk!

PR #7426, opened 2016-11-25
Allow PrimaryDataset default to be None for MCFromGEN

### LINKED ISSUE #7410: Fix PrimaryDataset for MonteCarloFromGEN
it can be None, since this spec always has input dataset.

amaltaro:
Fixes #7410 

@ticoann please review. Tested in my VM.

The next step (to be eventually done) would be to not save - in the workload_cache doc - parameters with None value.


amaltaro:
Tests are Ok on my vm. @ticoann please review and merge.

PR #7442, opened 2016-12-03
fix aborted and force complete request not killing the jobs

ticoann:
This is not tested and still has some issues.
Here is possible issues (might not be critical though).

1. There is race condition if some workflow is not cleaned up by LQ but it moved to aborted-archived status. (This should be rare but need to be fixed by how we propagate aborted-completed and       *-archived status from multiple agent).
2. Jobs still can be created and submitted between the time user aborted and killing job happens.
    a. currently aborted workflow check is made for wmbs population to prevent further pulling down the work
    b. But we can also add the check in JobCreator and JobSubmitter. (Not sure how much we will gain this - If we need to choose JobSubmitter check will be more effective)

3. killWork can be called concurrently (but it should cause the deadlock on db)

@amaltaro, Alan, don't need to review yet. I will update the code.

### DIFF
--- a/src/python/WMCore/Cache/GenericDataCache.py
+++ b/src/python/WMCore/Cache/GenericDataCache.py
@@ -1,5 +1,7 @@
 from __future__ import print_function, division
 import time
+import traceback
+import logging
 
 class MemoryCacheStruct(object):
     """
@@ -27,10 +29,16 @@ class MemoryCacheStruct(object):
             return True
         return False
     
-    def getData(self):
+    def getData(self, noFail=True):
         if self.isDataExpired():
-            self.data = self.func(**self.kwargs)
-            self.lastUpdate = int(time.time())
+            try:
+                self.data = self.func(**self.kwargs)
+                self.lastUpdate = int(time.time())
+            except Exception as ex:
+                if noFail:
+                    logging.error(traceback.format_exc())
+                else:
+                    raise
         return self.data
 
         
--- a/src/python/WMCore/Services/ReqMgr/ReqMgr.py
+++ b/src/python/WMCore/Services/ReqMgr/ReqMgr.py
@@ -52,9 +52,7 @@ class ReqMgr(Service):
             
         data = result['result']    
         if len(data) == 0:
-            return {}
-        elif len(data) == 1:
-            return data[0]
+            return []
         else:
             return data
     
@@ -100,7 +98,7 @@ class ReqMgr(Service):
         callname = 'request?%s' % query
         return self._getResult(callname, verb = "GET")
 
-    def getRequestByStatus(self, statusList):
+    def getRequestByStatus(self, statusList, detail=True):
         """
         _getRequestByStatus_
         
@@ -118,7 +116,7 @@ class ReqMgr(Service):
         TODO: need proper error handling if status is not 200 from orignal reporting.
         """
         
-        query = self._createQuery({'status': statusList})
+        query = self._createQuery({'status': statusList, 'detail': detail})
         callname = 'request?%s' % query
         return  self._getResult(callname, verb = "GET")
         
@@ -215,4 +213,17 @@ class ReqMgr(Service):
         propDict["RequestName"] = request
         return self["requests"].put('request/%s' % request, propDict)[0]['result']
     
+    
+    def getAbortedAndForceCompleteRequestsFromMemoryCache(self):
+        """
+        _getAbortedAndForceCompleteRequestsFromMemoryCache_
+        
+        """
+        # imports here to avoid the dependency not using this function
+        from WMCore.Cache.GenericDataCache import MemoryCacheStruct
+        
+        #TODO remove "aborted-completed status when state transition happens in reqmgr2
+        maskStates = ["aborted", "aborted-completed", "force-complete"]
+        return MemoryCacheStruct(expire=0, func=self.getRequestByStatus, 
+                                            kwargs={'statusList': maskStates, "detail": False})
         
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -28,6 +28,7 @@ from WMCore.DataStructs.JobPackage            import JobPackage
 from WMCore.FwkJobReport.Report               import Report
 from WMCore.WMException                       import WMException
 from WMCore.BossAir.BossAirAPI                import BossAirAPI
+from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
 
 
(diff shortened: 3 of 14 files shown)


ticoann:
Since clean up polling cycle is random there are considerable jobs are submitted after aborted and force complete happens. Although they are removed eventually, it is better to add the code not to submit those jobs. (Also somehow the killWorflow doesn't seem to kill pending jobs. I will update the patch.

amaltaro:
> Since clean up polling cycle is random there are considerable jobs are submitted after aborted and force complete happens. 

You mean the WorkQueueManager polling cycle to locally abort the workflow (cancel wq elements, kill jobs, etc)? If the delay between the request status transition (to aborted or force-completed) and killing the jobs and not submitting anything new is < 30min, then it should be good enough.

> (Also somehow the killWorflow doesn't seem to kill pending jobs. I will update the patch.

I was force-killing workflows during the last week (which in the end call killWorkflow) and I didn't have any issues. From the PyCondor point of view, there is no distinction between JobStatus.

Now getting back to your initial comments in this PR. The worst case scenario I can see is a race condition between locally killing the workflow and adding new subscriptions to wmbs_subscription (pulling work from LQ to WMBS). To avoid that problem, I think we should split this killing in two steps (two different polling cycles):
1) marking wq elements to Cancel (in all possible couch databases :-))
2) then completing work on SQL db (including what is created - in the JobSubmitter queue) and killing jobs from condor.

This would likely prevent some work that was already being processed by WorkQueueManager.

Sorry for being so verbose, checking your code now :)

src/python/WMCore/Cache/GenericDataCache.py
```diff
@@ -27,10 +28,16 @@ def isDataExpired(self):
...
-    def getData(self):
+    def getData(self, noFail=True):
         if self.isDataExpired():
-            self.data = self.func(**self.kwargs)
-            self.lastUpdate = int(time.time())
+            try:
+                self.data = self.func(**self.kwargs)
+                self.lastUpdate = int(time.time())
+            except Exception as ex:
+                if noFail:
```
amaltaro: It seems this `noFail` argument is not used (yet at least). How about we remove it? In the end it could just mask a real problem in case it is misused (as we had in the workqueue <--> policy interface). Otherwise, how about using `logging.exception` instead?

src/python/WMCore/Services/ReqMgr/ReqMgr.py
```diff
@@ -118,7 +118,7 @@ def getRequestByStatus(self, statusList):
         TODO: need proper error handling if status is not 200 from orignal reporting.
         """
         
-        query = self._createQuery({'status': statusList})
+        query = self._createQuery({'status': statusList, 'detail': detail})
```
amaltaro: Doesn't it return details by default? Maybe it would be better to accept a mask, such that it returns only the RequestName and RequestStatus to reduce the memory footprint.

amaltaro (changes requested):
Now that I went through the code, it seems you've added the protection I mentioned before to avoid getting new subscriptions into wmbs.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
amaltaro: Oh, I thought you were retrieving details as well.

ticoann:
Test in vm, it seems working OK.

ticoann:
test this please

ticoann:
@amaltaro, Hi Alan. I think patch is ready.
I am not sure why following test fails, it doesn't seem to be related.

```
WMCore_t.WorkQueue_t.WorkQueue_t.WorkQueueTest:testGlobalBlockSplitting changed from success to failure
```

ticoann:
test this please

ticoann:
@ericvaandering, Hi Eric, The report looks nice but one thing is missed compare to before is there is no stack trace where the failure occurs.

ticoann:
test this please

ticoann:
> @ericvaandering, Hi Eric, The report looks nice but one thing is missed compare to before is there is no stack trace where the failure occurs.

Never mind, Sorry now I found the link for that (default one)

ericvaandering:
Yeah, there is still the full Jenkins stuff under the "default details". This is just meant to replace the e-mails.

amaltaro:
And now I may know why wmbs was still creating jobs for aborted/force-completed workflows.
It seems we only update the workqueue db that is replicated (so global workqueue and local workqueue_inbox). The local workqueue is probably (I didn't find the code) updated by one of the WorkQueueManager threads and it only updates those elements to Done when all the local workqueue_inbox are marked as Canceled (which may as well be gone from the agent by there), see WQ data for a force-completed workflow [1].

Relying on the reqmgr status is certainly better than on the element status in couch. However, if we were marking as Canceled (or Done) local workqueue elements here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueue.py#L515

I think it would have avoided most of these problems.

[1]
```
cmst1@vocms0230:/data/srv/wmagent/current $ python getWQStatusByWorkflow.py amaltaro_MonteCarlo_eff_December_Patches_161218_182044_5572
Summary for https://cmsweb-testbed.cern.ch/couchdb/workqueue and request amaltaro_MonteCarlo_eff_December_Patches_161218_182044_5572
{u'Canceled': {u'http://vocms0230.cern.ch:5984': 3}, 'numberOfElements': 3}
Summary for http://localhost:5984/workqueue and request amaltaro_MonteCarlo_eff_December_Patches_161218_182044_5572
{u'Running': {None: 3}, 'numberOfElements': 3}
Summary for http://localhost:5984/workqueue_inbox and request amaltaro_MonteCarlo_eff_December_Patches_161218_182044_5572
{u'Canceled': {u'http://vocms0230.cern.ch:5984': 3}, 'numberOfElements': 3}
```

amaltaro:
@ticoann I made this patch to test another approach (without depending on changes in cmsweb): https://github.com/amaltaro/WMCore/commit/bc46eaf3a9113b58bf4e06ece653e76115e58fd2

In plain english, it should behave like:
1) once WorkQueueManager finds a workflow aborted (CancelRequested)
1.a) it moves the inbox elements to Canceled (but only elements pulled by the agent or available in GQ)
1.b) it moves ALL the local workqueue elements to Canceled (avoiding wmbs to pick up work)
2) TaskArchiver (in any agent) will only move the request status to {aborted-}completed if ALL the local workqueue_inbox elements (= global workqueue elements) for that workflow were Canceled
2.a.) meaning, all the agents working on that request have to go through and cancel work on their local queue.

First test looks promising. Ping me on slack please and we can discuss it over vidyo.

PR #7453, opened 2016-12-08
Add human readable printout in case when json.dumps failed to do its job

vkuznet:


src/python/WMCore/REST/Format.py
```diff
@@ -244,6 +244,7 @@ def stream_chunked(self, stream, etag, preamble, trailer):
                 trailer = None
                 raise
             finally:
+                print("ERROR, json.dumps failed to serialize %s, type %s" % (obj, type(obj)))
```
vkuznet: Alan, there is no general except: block here, the code only contains except GeneratorExit: which will be triggered ONLY when generator will exhaust. And, I don't know if except is desired here. But finally: will certainly called if exception was thrown. That's why I thought it is appropriate place.

Regarding cherrypy.log I think it is overhead anyhow, prints work more reliably and allows to get what you want without web backend library.

vkuznet: Let's Seangchan @ticoann to decide :)

amaltaro: If we don't catch the exception raised, finally block will re-raise it. So I think we should create a general except block, print the error message and raise it again.
Otherwise, this print will just show up for success or exception cases.

ticoann: I think Alan is right, doesn't finally block runs both exception and success case?

vkuznet:
Alan, Seangchan, please review, I made own exception block.

ticoann:
Hi Valentin, I think  following needs to be added.

vkuznet:
Done.

ticoann:
@vkuznet, Valentin, could you squash the commit for this? Thanks

vkuznet:
@ticoann, enjoy the squashed PR.

ticoann:
@vkuznet, Valentin, it is not squashed. Now there are more commits. 😄

vkuznet:
I HATE these squashes, things CAN GO WRONG with them and it is waste of time. Please discard this PR and take new one: https://github.com/dmwm/WMCore/pull/7479

PR #7455, opened 2016-12-09
Dont complete the same job twice in BossAir

amaltaro:
And I finally figured out why I always see plugin complaints in WorkQueueManager every time we abort or force-complete a workflow. We were completing killed jobs twice, 1st at BossAir level (properly handling the cache_dir and fwjr error) then again at the plugin level (which was not carrying the cache_dir info, thus those errors in the logs).

Tests were Ok, but I just made a few more changes. So let's see.


amaltaro:
Ok, I couldn't spot any side-effects on all my gazillions tests. I'm requesting reviews following the new smart GH way :)

PR #7456, opened 2016-12-09
Missing places in SimpleCondor still using unicode

amaltaro:
Complement #7390 . There were these 2 other functions that could be passing unicode string to the classad library that doesn't accept it.

Code already tested, but let's wait until the whole validation is over, just in case there are additional fixes over here.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -404,7 +404,7 @@ class SimpleCondorPlugin(BasePlugin):
         schedd = htcondor.Schedd()
 
         try:
-            schedd.act(htcondor.JobAction.Remove, "WMAgent_RequestName == %s" % classad.quote(workflow))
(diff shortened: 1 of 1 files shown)


amaltaro:
There is still one issue to sort out here. Jobs going into status 5 (Hold) are not getting removed from condor, though they seem to be getting completed in BossAir and wmbs... @hufnagel if you have any clue, I'm all ears :)

amaltaro:
That was just a line change `5: "Error"` instead of Running. However, I made further changes here #7459

hufnagel:
Hm, do we interpret 'Hold' as just a temporary job glitch and the job can recover or is it game over ? I thought the former, hence mapping it to running. If the latter, then yes, mapping to Error is correct.

amaltaro:
I think they have to be removed for most of the cases (so error). Those jobs that I experience these last days were in status 5 (Held) for 50k secs.

hufnagel:
Ok, so even if there might be a chance to get some of them back, in general it's not worth the wait and we should just move on and submit another job attempt. Agree with the change then.

amaltaro:
It's also good to go.

PR #7448, opened 2016-12-06
Return empty location if block is not complete anywhere

amaltaro:
Partially fixes #7446

Make sure to always return a list of block locations (even if block is still not completed anywhere).
TBD how to handle blocks with only invalid files.

Issue #7441, opened 2016-12-02
Handling Blocks with only Invalid Files


vlimant:
https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_BTV-RunIISummer16MiniAODv2-00029_00039_v0__161122_005312_2045

is in assigned, with one block 
/QCD_Pt-15to20_MuEnrichedPt5_TuneCUETP8M1_13TeV_pythia8/RunIISummer16DR80Premix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v2/AODSIM#63bcf038-ae2a-11e6-97bb-02163e0184a6
not complete anywhere. Is this the reason for not moving to acquired and will this move if the block gets wiped out? 

More info in #7466

amaltaro:
It will get acquired once we have this block complete in one of the expected locations:
https://cmsweb.cern.ch/phedex/datasvc/xml/prod/blockreplicas?block=/QCD_Pt-15to20_MuEnrichedPt5_TuneCUETP8M1_13TeV_pythia8/RunIISummer16DR80Premix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v2/AODSIM%2363bcf038-ae2a-11e6-97bb-02163e0184a6

Though I also think that piece of code could be improved and not crash in GQ.

vlimant:
the block is doomed IMO as that block has been deleted somehow. If the block disappears from TMDB, would that work ?

ticoann:
> the block is doomed IMO as that block has been deleted somehow. If the block disappears from TMDB, would that work ?

No, it wouldn't work. So we don't have any replica for this block?

Alan, Was this crashing the GQ?

amaltaro:
As Seangchan said, if data is not in TMDB, then there is no way to get it working.

Seangchan, yes, you can have a look at today's workqueue log (vocms0140). Otherwise I'll check it in the next week

ticoann:
HI Alan, I tried to look at vocms0140 log but it seems there is no work queue log /srv-log/vocms0140

amaltaro:
Seangchan, you can have a look at vocms022:/build/srv-logs/vocms0140/workqueue/workqueue-20161201.log

in any case, I paste it here before that gets lost/archived:
```
[...]

PR #7457, opened 2016-12-09
Take2 - return empty location if block is not complete anywhere

amaltaro:
My previous fix in #7448 was not completely correct.


amaltaro:
We still have to handle #7441 at WorkQueue level, but this one is good to go.

PR #7458, opened 2016-12-09
Remove origin_site_name conversion from SE to PNN

amaltaro:
As double-checked with Yuyi, `origin_site_name` reports the PNN instead of SE name for quite a while now. Thus I'm making these changes now before we forget it again.

Needs to be tested in VM and agent.
To be merged *after* #7457


ericvaandering:
+3/-81 Awesome.

amaltaro:
I'm getting good at these deletions LoL

PR #7459, opened 2016-12-09
Removed scheddLog mapping; updated stateMap to string status

amaltaro:
While investigating why Held jobs (condor status 5) are not removed from the condor queue. I noticed a couple strange things:
a) `bl_status` table uses the plugin `stateMap` keys to populate the schedd status `name` column (not very readable [1])
b) Held jobs were treated as Running jobs (according to stateMap, thus never killed)

I have also removed another mapping function that was no longer used. Last but not least, I moved the iteration over the schedd jobs to the try block, since the resource is still associated to the iterator while it's yielding values (thus it can timeout).

I need to test it in a fresh new agent, since it will change the values in `bl_status`.
I'll ask for reviews once I get the first tests running with no issues.

[1]
```
MariaDB [wmagent]> select * from bl_status;
+----+------+
| id | name |
+----+------+
|  1 | 1    |
|  8 | 100  |
|  3 | 2    |
|  2 | 3    |
|  5 | 4    |
|  4 | 5    |
|  7 | 6    |
|  6 | 7    |
|  9 | New  |
+----+------+
9 rows in set (0.00 sec)
```

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -189,18 +189,6 @@ class PyCondorPlugin(BasePlugin):
 
         return exitCodeMap
 
-    @staticmethod
-    def logToScheddExitCodeMap(x):
-        """
-        JobStatus shows the last status of the job
-        Get TriggerEventTypeNumber which is the current status of the job
-        Map it back to Schedd Status
-        Mapping done using the exit codes from condor website,
-        https://htcondor-wiki.cs.wisc.edu/index.cgi/wiki?p=MagicNumbers
-        """
-        logExitCode = {0: 1, 1: 1, 2: 0, 3: 2, 4: 3, 5: 4, 6: 2, 7: 0, 8: 0, 9: 4, 10: 0, 11: 1, 12: 5, 13: 2}
-        return logExitCode.get(x, 100)
-
     def __init__(self, config):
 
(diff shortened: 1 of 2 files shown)


hufnagel:
Why did you change the SimpleCondorPlugin stateMap ? It should key off numbers since that is what is returned from the python API. These have nothing to do with agent states. Have you actually tried this change ? I doubt it'll work.

amaltaro:
I changed it to have meaningful values in the bossair status table. However, giving it a second thought, it may be better indeed to keep the exact condor job status. We cannot have everything, so we just have to pick one... would you keep the numerical schedd_status or the english meaning?

This PR - as is - doesn't work :). I missed two other places that were locally fixed and it works now.

hufnagel:
For internal tracking and this is what this is (we don't expose these states to the outside, just what agent state they map to, right) I would stay with whatever the underlying implementation uses as state. Makes code maintenance easier in the long run IMO.

OTOH, it's not a huge change and there is always that pesky 'New' state that is there anyways for not yet submitted jobs...

amaltaro:
And ... this one is also good to go. I couldn't find any BAD side-effects of this change. I kept an eye on the bossair and wmbs tables and all the tracking and status seems to be Ok.

Nonetheless, @ticoann and @hufnagel please review. Feel free to give feedback even after it gets merged, I'll run more tests with the new tag that we'll cut tomorrow or so before pushing it to production.

PR #7465, opened 2016-12-12
Adding StompAMQ interface

stiegerb:


src/python/WMCore/Services/StompAMQ/StompAMQ.py
```diff
@@ -0,0 +1,175 @@
...
+        successfully_sent = []
+        for notification in data:
+            body = self._send_single(conn, notification)
+            if body:
+                successfully_sent.append(body)
+
+        if conn.is_connected():
+            conn.disconnect()
+
+        self._logger.warning('Sent %d docs to %s' % (len(successfully_sent), repr(self._host_and_ports)))
```
amaltaro: I suggest we make it an `info` log for the moment.

stiegerb: Ok, fixed this. And most of the pylint warnings.

amaltaro:
Benjamin, can you remind me what's the stomp version you're using? We may need to update this spec (I'll take care of that):
https://github.com/cms-sw/cmsdist/blob/comp_gcc493/py2-stomp.spec

Moreover, can you please squash all these commits in a single one?

stiegerb:
I was testing with the latest version (stomp.py 4.1.15).

For squashing, do you want me to open a new PR or force push here? I also just learned that you should actually be able to "squash & merge" directly on Github: https://github.com/blog/2141-squash-your-commits

amaltaro:
There is no need for a new PR. You'll basically merge them all together and push only one commit.
See the procedure here: http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request

This "squash & merge" from GH is kind of evil because if you (e.g. Eric :)) are tracking this repo with some graphical tools (like sourceTree), then you cannot see when this branch got merged into master.

stiegerb:
Ok, done.

amaltaro:
@stiegerb can you please add these import statements as suggested above?
I discussed this with Seangchan and even if we're not using it - yet - it's safer to add it now to get used to python3 behavior.

stiegerb:
Ok, done.

ticoann:
Thanks Benjamin, I am merging this. If you get a chance, could you created the unit test for it. 
Not sure how it would test though without revealing any secrets (user id, etc) or connecting production service. Maybe some kind of emulator? #7473

stiegerb:
Thanks Seangchan. I can have a look at the unittest, but it will probably have to wait until January.

amaltaro:
For the record, I was trying to build stomp rpms with 4.1.15 but that's failing because the source code cannot be found in pypi. I suggest we use what we have already in the repo (3.1.3) and in case of issues, then we try to update it. 

BTW, ASO uses stomp, so we need to ping them in case we update that package.

stiegerb:
Ok, this might be an issue.

I tested with stomp.py 3.1.7 (the closest to 3.1.3 I could find on pypi), and the interface changed a bit. I fixed it to run with that version and tested it a bit, see here:
https://github.com/stiegerb/WMCore/commit/ed1a0f7853149063a3705a9f21e084c384f0e763

We'll have to decide which version of stomp.py we want to use.

PR #7449, opened 2016-12-07
fix site location when location changed

ticoann:
fixes #7443

PR #7472, opened 2016-12-14
1.0.21 wmagent

ticoann:
WMAgent branch 1.0.21.

#7449 
#7442 
#7477

### DIFF
--- a/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
+++ b/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
@@ -217,7 +217,15 @@ class ErrorHandlerPoller(BaseWorkerThread):
                 if times is not None:
                     startTime = times['startTime']
                     stopTime = times['stopTime']
-
+                    
+                # correct the location if the original location is different from recorded in wmbs
+                # WARNING: we are not updating job location in wmbs only updating in couchdb by doing this.
+                # If location in wmbs needs to be updated, it should happen in JobAccountant.
+                locationFromFWJR = report.getSiteName()
+                if locationFromFWJR:
+                    job["location"] = locationFromFWJR
+                    job["site_cms_name"] = locationFromFWJR
+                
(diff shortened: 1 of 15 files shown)


ericvaandering:
I changed Jenkins to only test things that are going to the master branch. I don't know if I can trigger a test on this PR. Let me check.

ericvaandering:
test this please

ticoann:
That is good, it doesn't trigger. Thanks Eric

ericvaandering:
I was hoping it would trigger if asked. Oh well, @amaltaro has to be careful now building the branches. Doing them was a pain since the tests may change a lot between this and the reference.

amaltaro:
It's good not to see the unit tests results for branch PRs indeed, thanks!

I suggest we always run pylint checks locally for these branch PRs (especially after sorting conflicts out). If we can run only them here, it would be better, but otherwise it's not a big deal.

amaltaro:
Testing it in vocms0230. If everything goes well, I'll push this patch to all the agents tomorrow.

amaltaro:
@ticoann I forgot we also need this patch on the agent https://github.com/dmwm/WMCore/pull/7477
Can you please cherry-pick it as well? I'm running more tests with this patch in place now.

amaltaro (changes requested):
The site location patch looks good (though I looked only at a couple diff sites).

The abort patch is not ready for production, it will leave workflows stuck in force-complete. Which reminds me, we haven't made the changes to make this transition on GQ, have we?

ticoann:
> The abort patch is not ready for production, it will leave workflows stuck in force-complete. Which reminds me, we haven't made the changes to make this transition on GQ, have we?

As we discussed this is known fact and complimentary script is running as cronjob in vocms049. (every hour). I included #7477 with this patch and running the test. If everything goes fine, I will patch production agents.

ticoann:
All the production and relval agents are updated. (But not hepcloud agents)

PR #7477, opened 2016-12-15
Update WorkQueue Interface code for getting work by status

amaltaro:
Tricky returned data from ReqMgr apis. If details are set to False, then it returns a flat list.
If True, then it returns a list with a dict (plus several inner dictionaries).

BTW, I completely removed `_getRequestsByTeamsAndStatus` because it was used only by GQ and, GQ doesn't get/split work based on teams, it simply gets whatever is in "assigned" or "running-closed".

Finally, I've also removed the "Teams" check because we now validate the team data during assignment (must be a non empty string - tested it this afternoon).

@ticoann we need this patch in otherwise GQ will break nicely. Please review. 
Tests look good so far

### DIFF
--- a/src/python/WMComponent/JobUpdater/JobUpdaterPoller.py
+++ b/src/python/WMComponent/JobUpdater/JobUpdaterPoller.py
@@ -116,7 +116,8 @@ class JobUpdaterPoller(BaseWorkerThread):
         for workflow, priority in workflowsToCheck:
             if workflow not in priorityCache:
                 try:
-                    priorityCache[workflow] = self.reqmgr2.getRequestByNames(workflow)[workflow]['RequestPriority']
+                    result = self.reqmgr2.getRequestByNames(workflow)[0]
+                    priorityCache[workflow] = result[workflow]['RequestPriority']
                 except Exception as ex:
                     logging.error("Couldn't retrieve the priority of request %s", workflow)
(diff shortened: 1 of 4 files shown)


ticoann:
test this please

amaltaro:
test this please

ericvaandering:
What about: " TODO remove "aborted-completed status when state transition happens in reqmgr2"?

amaltaro:
@ticoann Seangchan can correct me, but I think the idea is to make the aborted-completed transition from ReqMgr/WorkQueue, instead of the agent. Fixing the other simple things.

ticoann:
Alan, yes you are right I am trying to create another patch for that - moving status transition to reqmgr. (It would take a while)

amaltaro:
@ticoann there were 2 other places affected by the different data type. JobUpdater and ReqMgrService.

I think it's all fixed now. Please review and merge

amaltaro:
Another unstable test failing.

ticoann:
I will add below to unstable test
WMCore_t.WorkQueue_t.WorkQueue_t.WorkQueueTest:testQueueChainingStatusUpdates

ericvaandering:
Once they are added you can ignore if they fail. The unstable tests are not currently fetched in real time.

ticoann:
test this please

ticoann (approved):

PR #7479, opened 2016-12-15
Hadle json.dumps failures in human readable form

vkuznet:


vkuznet:
Please take this PR instead of #7453

ericvaandering (changes requested):
Please at least fix the things in bold here:
https://cmssdt.cern.ch/jenkins/view/All/job/DMWM-WMCore-PR-test/3682/artifact/artifacts/PullRequestReport.html#pylint

vkuznet:
Eric,
I don't remember if we agreed to follow pylint suggestions to fix lines which
didn't related to PR changes. In this case pylint complains about lines which
I never touched in my PR and in my opinion it should be done in separate PR,
since changes will not reflect PR purpose.

Please clarify. I'll be happy to make a separate PR for pylint suggestions,
but it really defeats the purpose of PR itself.

vkuznet:
Plus some suggested (in bold) changes are tricky to classify and their
resolution may require major code re-factoring. Here is clear example,
from pylint bold suggestion:
W0602, line 425 in stream_compress: Using global for '_stream_compressor' but no assignment is done

Not only global is bad by design, but assignment should be done. I have no idea
to which variable.

ericvaandering:
Please fix them, but what we generally do is add a *2nd commit* to this PR to address those changes. That makes review easy. If you are unsure (in general) if the purpose of the PR may need changes and new commits, you can wait.  But before it's merged, we should fix at least the most egregious errors.

Major changes should be avoided, I agree. Usually those are not the case.

ticoann:
I don't know why global variable is used.

> W0602, line 425 in stream_compress: Using global for '_stream_compressor' but no assignment is done

assignment is done in 

https://github.com/vkuznet/WMCore/blob/785179ae4be0d175de5fddda2763b51f7b9aa3c5/src/python/WMCore/REST/Format.py#L411

One other thing is something like following is legit case if we inherit the class and doesn't need to overwrite __init__ method

W0232, line 312 in RawFormat: Class has no __init__ method

ericvaandering:
Description

Used when a class has no __init__ method, and neither does its parent classes.


Are you sure the parent class has one?

ericvaandering:
RestFormat has no init(). Adding one will clear up all the errors. Maybe even converting it to RESTFormat(object) will do it.

ticoann:
> Used when a class has no init method, and neither does its parent classes.
> 
> Are you sure the parent class has one?

You are right parents class doesn't have __init__ method. In this case 
https://github.com/vkuznet/WMCore/blob/785179ae4be0d175de5fddda2763b51f7b9aa3c5/src/python/WMCore/REST/Format.py#L37

So we can add following in parent class.
```
def __init__(self):
    pass
```
But is this the right thing to do? If we don't have anything to initialize?

ticoann:
> RESTFormat(object) will do it.

I think that is better

vkuznet:
While, you're writing emails, I done all the changes.

ericvaandering:
I'm sure pylint will show this, but:

```
Python 2.7.10 (default, Jul 30 2016, 18:31:42) 
>>> dir(object)
['__class__', '__delattr__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__']
```

so init is there to be picked up, I hope.

src/python/WMCore/REST/Format.py
```diff
@@ -35,6 +34,9 @@ def is_iterable(obj):
         return True
 
 class RESTFormat:
+    def __init__(self):
```
ericvaandering: Like we said in our emails, I think inheriting from object will fix this with no other changes

ticoann:
You are too fast. :-) Sorry Valentin, could update with new class definition with object. I think that is cleaner

vkuznet:
I bet you'll ask for squash, what a saga :)

ericvaandering:
Of the last two, probably. :-) But let Jenkins finish with it first if you don't mind. Having three running tests on the same bit of code may be confusing.

ericvaandering (approved):

ericvaandering:
Only thing holding up pylint is the global, which we agree is beyond the scope of this PR

ticoann:
And squashing the last 2 pylint related commit together. :-)

vkuznet:
Done

PR #7480, opened 2016-12-16
add method for tag collector

ticoann:


vkuznet:
+1

ericvaandering:
See #7483

ticoann:
I will factor out the tag collect stuff outside ReqMgr2 as Eric suggested but merge this one first.

PR #7481, opened 2016-12-16
More changes to handle the new json templates in 7365

amaltaro:
Changes needed for the the gazillions different ways of injecting and assigning workflows, like:
*) if the json template contains AcqEra/ProcStr/ProcVer in the assign dict, then we send post their key/value pairs during assignment
*) properly handle dict values in the assign template for AcqEra and ProcStr. ProcVer could be added as well, but I don't think we've ever used it.


ericvaandering:
test this please

ticoann (approved):

PR #7486, opened 2016-12-19
Remove deprecated workqueue argument: NoLocationUpdate

amaltaro:
This argument was replaced more than 6 months ago by `NoInputUpdate` and `NoPileupUpdate`.
Thus I'm removing this code.

I've also updated the json file with workqueue elements that had only the deprecated arg. Both args were added and kept the same default value (False).


amaltaro:
test this please

ericvaandering:
Why is the updated Jason file 60k lines longer?

amaltaro:
Because I pretty dumped it, in case someone needed to look at one "document" structure. Maybe it's not a good idea and I should just dump without indent (single line)?

ericvaandering:
It's fine. One thing I'd suggest is using options like

```json.dump(lookup, mockData, indent=1, separators=(',', ': '), sort_keys=True)```

which will keep the size a bit smaller and keep things sorted. Then future diffs won't be as large

amaltaro:
Json file updated as you suggested.

ericvaandering (approved):

amaltaro:
@ticoann good to be merged

ticoann (approved):

PR #7493, opened 2016-12-22
add query for checking draining status

ticoann:
@goughes, Erik, this is initial check for draining status. (shouldn't be merged - just start point the Erik and work on)

when agent is set to drain.

1. In AgentStatusWatcher, check the drain status and run this query to whether everything is read to go. (all the count has to be 0)
2. report this to wmstats (and email). 

There need to be a check for stuck workflows (jobs won't be run), Alan will tell what condition to check. (This should run before above check)

We can discuss in more detail but start with this patch. Also could you create the unittest for it.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/CheckDrainStatus.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/CheckDrainStatus.py
@@ -1,0 +1,46 @@
+from __future__ import print_function, division
+from WMCore.Database.DBFormatter import DBFormatter
+
+class CheckDrainStatus(DBFormatter):
+    """
+    _CheckDrainStatus_
+    
+    1. checks all the workflows are completed status (within the agent)
+    2. checks all the blocks are closed
+    3. checks all the files are injected to PhEDEx
+    4. checks all the files are uploaded to DBS
+
+    TODO: need to check archived statsus
+    """
+    completedSQL = "SELECT count(*) FROM dbsbuffer_workflow WHERE completed = 0"
+    
+    blockClosedSQL = "SELECT count(*) FROM dbsbuffer_block where status != 'Closed'"
+
+    phedexSQL = "SELECT count(*) FROM dbsbuffer_file WHERE in_phedex = 0"
+    
+    dbsSQL = "SELECT count(*) FROM dbsbuffer_file WHERE status != 'InDBS'"
+
+    def _executeQuery(self, sql, keyName, conn = None, transaction = False):
+        result = self.dbi.processData(self.sql, conn = conn, transaction = transaction)
+        count = self.format(result)[0]
+        return {keyName: count}
+        
+    def execute(self, conn = None, transaction = False):
+        result = {}
+        # number of workflows
+        completeStatus = self._executeQuery(self.completedSQL, "NotCompleted", conn, transaction)
+        result.update(completeStatus)
+        
+        # number of blocks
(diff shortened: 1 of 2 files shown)


src/python/WMComponent/DBS3Buffer/MySQL/CheckDrainStatus.py
```diff
@@ -0,0 +1,46 @@
...
+
+    TODO: need to check archived statsus
+    """
+    completedSQL = "SELECT count(*) FROM dbsbuffer_workflow WHERE completed = 0"
+    
+    blockClosedSQL = "SELECT count(*) FROM dbsbuffer_block where status != 'Closed'"
+
+    phedexSQL = "SELECT count(*) FROM dbsbuffer_file WHERE in_phedex = 0"
+    
+    dbsSQL = "SELECT count(*) FROM dbsbuffer_file WHERE status != 'InDBS'"
```
amaltaro: We should apply some filters to these two `dbsbuffer_file` queries, since we don't want to get hold by backfill or crappy files that we don't really care. Some examples here:
https://github.com/amaltaro/ProductionTools/blob/master/drainAgent.py#L63

amaltaro (approved):
Changes look good Seangchan. Though I don't think it will ever report the agent as clear to go, since we *always* have files marked as not in phedex.

We need to track that down and make sure the phedex status is correct for all the cases.

src/python/WMComponent/DBS3Buffer/MySQL/CheckDrainStatus.py
ticoann: Thanks Alan, I will update the query.  Other than unmerged files (I thought we are not putting unmerged files in dbsbuffer db) and backfill case there is the case files won't be injected to PhEDEx? I mean there is a legit case we can filter out?

amaltaro: In theory, any files should be injected into phedex. I put it in my script because I saw such cases in the past and we shouldn't bother about it.
Giving it a second though - this code will be in WMCore, not a dummy script :) - maybe we don't filter anything on this query, but try to track down those cases and make sure this table is in a clean and consistent state. What do you think?

ticoann: Yes I agree that we need to be more strict and find the problem and fix it. But we also have to filter legit cases - unmerged, backfill(?). Alan stop replying (you are on vacation. 😄 )

amaltaro: Now that vacation is over :-) How about we then filter out `/store/unmerged/` and
`/store/backfill/`? For files not in phedex and files not in dbs.
The backfill legit case would be Ops testing a new site that doesn't have a proper site setup or so...

amaltaro:
One more comment on it. I noticed you check whether files are in dbs with `status != 'InDBS'` while what I've been using is `status = 'NOTUPLOADED'`.

The third DBS status in dbsbuffer_file table is `GLOBAL`. Do I read the code correctly to say that:
`NOTUPLOADED`: new files created by a workflow, not injected into dbs
`GLOBAL`: input files for workflows (so already created by other workflow)
`InDBS`: new files created by a workflow AND injected into dbs.
?

ticoann:
replaced by #7757

PR #7496, opened 2016-12-23
Move GQ cronjob to cherrypy thread

ticoann:
This also requires deployment script change and manage script creation (https://github.com/dmwm/deployment/pull/424)


ericvaandering:
test this please

src/python/WMCore/GlobalWorkQueue/CherryPyThreads/LocationUpdateTask.py
```diff
@@ -0,0 +1,26 @@
...
+
+class LocationUpdateTasks(CherryPyPeriodicTask):
+
+    def __init__(self, rest, config):
+
+        CherryPyPeriodicTask.__init__(self, config)
+
+    def setConcurrentTasks(self, config):
+        """
+        sets the list of functions which
```
amaltaro: I suggest to either remove or correct this docstring :)

amaltaro (commented):
I x-checked the cronjobs against the cherrypy threads and they look consistent and correct.

Duration of these threads is the polling cycle relative to the last execution or absolute time? I'm just wonder if we could end up with two "same" threads running in parallel in case one takes too long (which is quite normal for GQ).

One suggestion though, can you reformat these codes to get the pylint results as clean as possible? Especially on new files.

Before merging it, maybe we could get the deployment changes/PR to check whether everything is fine.

ticoann:
Thanks Alan for reviewing, I am working on the deployment script. (Yes I will merge this after testing)
Thread duration is not absolute or precise, since it is running in the same thread. it doesn't start new thread if the thread is not finished. 

I will fix pylint and comments as well.

ticoann:
This patch seems to be working fine now. I will update pylint fix.

ticoann:
@amaltaro, Alan, please review I updated pylint as I can as well. It is tested. and seems to work.
I still need add to code to check heartbeat of thread but I will do in separate patch

PR #7497, opened 2016-12-29
Run certain paths outside of normal slice structure

ericvaandering:
~~For now, just some debugging~~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,12 +6,11 @@ from __future__ import print_function
 
 import os
 import os.path
-from distutils.core import setup, Command
+from distutils.core import Command, setup
 from os.path import join as pjoin
 
-from setup_build import BuildCommand, InstallCommand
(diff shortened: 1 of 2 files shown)


ericvaandering:
test this please

ericvaandering:
The goal of this pull request is to allow us to get the Api_t tests tested on their own because, for whatever reason, they can't be tested in a container with other tests. They can be tested by themselves, though. 

This adds an 11th slice to the tests which only does the Api_t tests. It seems to have the side effect of making them all succeed.

ericvaandering:
@ticoann this is as good as as I can really make it. Please review and merge. 

I removed all the pylint testing stuff since we do that on our own. Frankly, looking at this code, I don't really understand why this is part of setup_test.py. We could probably just do this in shell scripts.

ticoann (approved):

PR #7500, opened 2017-01-03
Add a logger to (old) SiteDB emulator to match Service.py

ericvaandering:
This may help with some of the WorkQueue tests


ericvaandering:
Tests are still failing, but with a more useful error about data not existing for acdc tests.

@ticoann please merge. I'm not fixing the many pylint warnings in this code since it's the old SiteDB emulator and we are moving towards replacing it with a Mock-ed emulator.

ticoann:
Thank Eric, I will merge that when Jenkins check is finished.

PR #7501, opened 2017-01-03
Move Simple_t into another slice

ericvaandering:
This and previous replace #6921


ericvaandering:
@ticoann please merge this. Looking back at my notes on what it takes to run tests in Docker these are problematic as well when run with other tests.

PR #7504, opened 2017-01-04
fix the job report when ther is no possible matching site

ticoann:
@amaltaro, Alan please review

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-#pylint: disable=W0102, C0301
+#pylint: disable=C0301
 # for the whitelist and the blacklist
 # C0301: I'm ignoring this because breaking up error messages is painful
 """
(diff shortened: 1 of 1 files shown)


ericvaandering (changes requested):
On line #3 we're turning off detection of one of the more dangerous features of python. Can you remove that and then fix any errors it actually turns up?

amaltaro (changes requested):
I'm not sure it will really fix the problem, but sounds good :)
Can you also update this code please:
https://github.com/ticoann/WMCore/blob/396abf16f8790398d996a2da66dc2a8b6188a059/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L432
?
We shouldn't have any old jobs out there and, to be on the safe side you could use `job.get()` in the `if`

ticoann:
@ericvaandering, @amaltaro,  Eric, Alan, please review again updated what you suggested.

ericvaandering (approved):

ticoann:
@amaltaro, Alan I updated the patch, could you review again?

amaltaro:
test this please

amaltaro (approved):

PR #7505, opened 2017-01-04
Throw an exception when accessing old SiteDB emulator

ericvaandering:
Run this as a test to figure out where we are using this still


amaltaro:
test this please

ericvaandering:
Not needed. Patch removing it is going in soon.

PR #7506, opened 2017-01-04
Wq with mock sitedb

ericvaandering:


ericvaandering:
@ticoann please merge

PR #7507, opened 2017-01-04
allow overflowing job to sites not in site white list

### LINKED ISSUE #7503: Submit failed to site that are not in drain
I was looking at https://cmst2.web.cern.ch/cmst2/unified/report/vlimant_task_JME-RunIISummer16DR80-00006__v1_T_170103_111641_3417 that failed submit from agent 303 and some of the sites are/were not in drain at any point other the past few days.

Could not trace anything in 303 logs, so there is something very fishy.

https://its.cern.ch/jira/browse/CMSCOMPPR-369

ticoann:
fixes #7503 @amaltaro, Alan please review

### DIFF
--- a/src/python/WMCore/JobSplitting/JobFactory.py
+++ b/src/python/WMCore/JobSplitting/JobFactory.py
@@ -190,6 +190,12 @@ class JobFactory(WMObject):
                         fileLocations = locSet
                         if len(self.siteWhitelist) > 0:
                             locSet = locSet & set(self.siteWhitelist)
+                            if len(locSet) == 0:
+                                # if there is no matching location in site whitelist,
+                                # select the current input file location
+                                # this allows overflowed jobs to be run again
+                                locSet = fileLocations
+                                
                         if len(self.siteBlacklist) > 0:
                             locSet = locSet - set(self.siteBlacklist)
 


ticoann:
test this please

amaltaro (commented):
Hi Seangchan, I don't think this change is needed and it shall not have any effect. The reasons are:
a) we only set the sitewhitelist for top level tasks, see https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L501 , thus there is no previous task that was overflowed and that if should never be true.
b) CompOps operates on this mode for quite a long time already and it was not an issue.

Is there a GH issue reporting an actual problem? I'd guess the problem is somewhere else (maybe sites in down/drain).

ticoann:
Thanks Alan, Now I understand why we didn't see the problem before. Somehow the cloned workflow setting site whitelist for child task and that was causing the problem. I think patch is safe since the site white list filtering will happen in WQ. but the better option is we have to make sure site white list is empty for child task.

amaltaro:
Seangchan, to avoid these cloning problems that happen from time to time, how about we define a new CLONE mask with only the create requests? Here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L280

This would be a minor change on the CompOps scripts

ticoann:
I think we can provide "clone" api to OPs. so Ops don't have to do any thing special.
However, if child task shouldn't have site white list and black list in all the cases. I think we should reset everything when site white list is set for the safe side.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L503

So setting every child tasks site white list to be empty

amaltaro:
Yes, a complete work would be providing a clone api that also accepts a json as input to overwrite values needed by Ops. Though the usability is not clear, since Ops would need to GET this request schema first, check what needs to be changed and post the changes to this reqmgr2 api.

> So setting every child tasks site white list to be empty

It's certainly a good idea to have this protection. However, I'd rather work on this issue:
https://github.com/dmwm/WMCore/issues/7133
to sort out all these possible cases.

amaltaro:
@ticoann based on the last comments, shall we close this one?

PR #7510, opened 2017-01-05
API is no longer unstable (I think). Storage one is

ericvaandering:


ericvaandering:
@ticoann please merge if you think this is OK. Considering that API is run in its own slice now, I imagine it will remain stable.

PR #7511, opened 2017-01-05
Remove old SiteDB emulator

ericvaandering:
No longer actually called. May turn up some places where it thinks it is.


ericvaandering:
Needs some work

ericvaandering:
@ticoann please merge when Jenkins comes back OK. (These are just squashed of the last run)

PR #7514, opened 2017-01-05
Move last use of old-style PhEDEx emulator to mock-based emulator

ericvaandering:


ericvaandering:
Please review. In all honesty, this unit test fails again and was failing before. But nothing else fails and it lets us completely remove the old PhEDEx emulator. Probably some actual work should go into a new emulator.

ticoann:
Eric, mock PhEDEx api still uses the fake data, right? (dataset name, site from DataBlockGenerator)

ericvaandering:
Somewhat, yes. Another thing I want to do is get rid of some of that. Will probably involve making a more robust PhEDEx emulator. Currently parts of the mock PhEDEx use the mock DBS to get correct names.

Eric

ticoann (approved):

ericvaandering:
@amaltaro can you review this please?

src/python/WMQuality/Emulators/PhEDExClient/MockPhEDExApi.py
```diff
@@ -17,16 +21,22 @@
...
 
 class MockPhEDExApi(object):
     """
     Version of Services/PhEDEx intended to be used with mock or unittest.mock
     """
 
-    def __init__(self, responseType="json"):
-        pass
+    def __init__(self, dict=None, responseType="json", secure=True):
+        dict = dict or {}
```
amaltaro: This `dict` arg is not really needed, is it?

amaltaro (approved):
Besides one minor comment in the code, changes look good.

I have a couple of questions comments though:
a) the old emulator - to be removed - is https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/Emulators/PhEDExClient/PhEDEx.py, right? How about we delete it in the same PR just to make sure it's not used anymore?
b) out of the scope of this PR, but I noticed we have DataBlockGenerator.py and DataBlockGenerator3.py. Is one meant to be used for DBS2 and the other DBS3? Maybe someone :-) would like to merge them together

src/python/WMQuality/Emulators/PhEDExClient/MockPhEDExApi.py
ericvaandering: Unfortunately, yes. This function has to have the same signature as the real PhEDEx service and since that one is sometimes called with named arguments, it even needs to be called dict. Not happy about that, but I don't really want to change the PhEDEx interface at this time.

PR #7492, opened 2016-12-22
add couch view to check wqe summary

ticoann:
tested the view but need to add unittest

PR #7516, opened 2017-01-06
fix rewrite rule

ticoann:
fixes WQ display of elements view


ticoann:
@amaltaro, Alan, my WQ new view patch broke the WQ page. Please review. #7492

amaltaro (approved):
You're my hero! That was it, I applied this patch to my VM and it works fine now. Thanks!

PR #7237, opened 2016-09-29
Fetch data only for a specific request in WorkQueue summary util

amaltaro:
I noticed it was not fetching information for the request name provided in the command line.
@ticoann please review, I ran a few tests and it looks Ok

PR #7519, opened 2017-01-07
WorkQueue view to get elements stats by Status

amaltaro:
Added one more couch view to get a summary based on the elements status.

A query like https://alan-cloud1.cern.ch/couchdb/workqueue/_design/WorkQueue/_view/jobsByStatus?reduce=true&group=true should return a nice summary as
```
{"rows":[
{"key":"Available","value":{"sum":102,"count":2,"min":33,"max":69,"sumsqr":5850}},
{"key":"Done","value":{"sum":466,"count":6,"min":2,"max":350,"sumsqr":127778}},
{"key":"Running","value":{"sum":871,"count":10,"min":0,"max":329,"sumsqr":197293}}
]}
```


ticoann:
Alan, are you getting the stats from LQ? GQ? or both? Don't you needs stats by request?

amaltaro:
It will be used to get stats from both queues. This is supposed to be a very high level summary.

Then later we'll need to have a distribution of jobs per site, as written in the wiki (section 2 and 4):
https://github.com/dmwm/WMCore/wiki/WMAgent-monitoring

where we could either use `getElementsStatusAndJobsByWorkflow` with reduce=false, or use this method:
https://github.com/dmwm/WMCore/pull/7237/files#diff-3cf0385be35cd9b5a8689054609f9d20R104

though it may be too heavy.

BTW, I fixed a problem in `getElementsStatusAndJobsByWorkflow` with the options and it seems it breaks all the workqueue unit tests now. Why would it break if we pass `update_after`?
PS.: please don't reply today :), we can discuss it on Monday.

amaltaro:
test this please

ticoann:
test this please

amaltaro:
And the next result should be cleaner...

amaltaro:
Ahaa! Apparently these workqueue tests are inserting a document into the database (queueWork) and then it starts processing them (processInbound), however in between there is a conflicts fix function (where it fails because it doesn't retrieve anything from couch).

@ticoann  so my theory is that the function `_stats` is not really happy about these elements with Jobs=None. Though I cannot see how one view could affect other views accesses... :(

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
@ticoann I had a look at most of the views and it seems changing the Jobs default value from None to 0 doesn't cause any problems (confirmed by the unit tests as well). Would you have any concerns with this change?

ticoann:
Alan, I don't think that will be the problem but let me check

amaltaro:
Seangchan, do you want me to take advantage that I'm working with these views and also update your latest one to `_stats`
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/views/elementsDetailByWorkflowAndStatus/reduce.js
?

ticoann:
Alan, Actually I will change https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/views/jobStatusByRequest/reduce.js

to _stats and remove what I added
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/views/elementsDetailByWorkflowAndStatus/reduce.js

ticoann:
Alan I was looking at the code where Jobs == None is treated differently than other case. i.e. GQ work queue_inbox doesn't have Jobs so it will be None. But I don't find anything that would affect the statistics.

I don't think following codes are harmful but  if the default value is 0  Jobs != None is not necessary.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElementResult.py#L27

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElementResult.py#L143

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElementResult.py#L150

ticoann:
Alan, I created the #7524 to use the _stats for completion checking view.  (This patch need to be merged first for unittest to pass for #7524)

amaltaro:
Thanks, Seangchan. I'm updating WorkQueueElementResult then.

amaltaro:
For some reason one WorkQueue unittest was failing to see the up-to-date docs after a updateElements. I had to load the view once again before getting completedWork.

@ticoann can you please review again
I suggest we cut a new tag after merging this one and #7524

ticoann (approved):

PR #7524, opened 2017-01-12
use jobStatusByRequest view for complete status

ticoann:
Alan, please review. #7519 has to be merged first for unit test to be succeeded


src/couchapps/WorkQueue/views/elementsDetailByWorkflowAndStatus/reduce.js
```diff
@@ -1,17 +0,0 @@
-function (key, values, rereduce) {
```
amaltaro: Since you deleted this reduce function, shouldn't you also rollback the rewrite arg:
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/rewrites.json#L7
?

src/python/WMCore/Services/WorkQueue/WorkQueue.py
```diff
@@ -61,7 +61,7 @@ def getJobStatusByRequest(self):
         data = self.db.loadView('WorkQueue', 'jobStatusByRequest',
                                 self.defaultOptions)
         return [{'request_name' : x['key'][0], 'status': x['key'][1],
-                 'jobs' : x['value']} for x in data.get('rows', [])]
+                 'jobs' : x['value']['sum']} for x in data.get('rows', [])]
```
amaltaro: Since we have all the stats, I would make a keyword arg for this method (e.g. fullStats=False) such that one could either retrieve number of jobs per request+status OR stats per request+status. e.g.:
`[{'request_name' : x['key'][0], 'status': x['key'][1], 'jobs' : x['value']['sum']}]`
or
`{'request_name' : x['key'][0], 'status': x['key'][1], 'stats' : x['value']}`

src/python/WMCore/Services/WorkQueue/WorkQueue.py
```diff
@@ -191,12 +191,12 @@ def getElementsStatusAndJobsByWorkflow(self, inboxFlag=False, stale=True):
         options = {'reduce' : True, 'group_level': 2}
         if stale:
             options['stale'] = 'update_after'
-        data = db.loadView('WorkQueue', 'elementsDetailByWorkflowAndStatus',
+        data = db.loadView('WorkQueue', 'jobStatusByRequest',
                                 {'reduce' : True, 'group_level': 2})
```
amaltaro: I have this line fixed in my PR 7519, it should be `options` instead of this dict of options.

src/python/WMCore/Services/WorkQueue/WorkQueue.py
```diff
@@ -191,12 +191,12 @@ def getElementsStatusAndJobsByWorkflow(self, inboxFlag=False, stale=True):
         options = {'reduce' : True, 'group_level': 2}
```
amaltaro: option name is `group` either true or false (though I think group_level would work here...)

amaltaro: And you may want to rename this method as well :)

amaltaro (changes requested):
After these comments, I actually think it would be better to merge those two methods (getJobStatusByRequest and getElementsStatusAndJobsByWorkflow) together. They have the basically the same functionality with different options :-)

src/couchapps/WorkQueue/views/elementsDetailByWorkflowAndStatus/reduce.js
ticoann: when reduce function doesn't exist, default is reduce=false. But I think it is better to have explicit since if we later add reduce function we don't have to change this again.

src/python/WMCore/Services/WorkQueue/WorkQueue.py
ticoann: I was suspecting this function is used somewhere outside WMCore. I couldn't find any usage outside unittest. 
So I didn't want to change the return type. But I couldn't find any usage from WmAgentScript. So I guess we can updated that.

src/python/WMCore/Services/WorkQueue/WorkQueue.py
ticoann: If group_level is set group is always true.   I will change the function name. :-)

src/python/WMCore/Services/WorkQueue/WorkQueue.py
ticoann: Thanks I will update that.

ticoann:
> After these comments, I actually think it would be better to merge those two methods (getJobStatusByRequest and getElementsStatusAndJobsByWorkflow) together. They have the basically the same functionality with different options :-)

The problem is we can't use _stats we have to use create reduce since the value is more than Jobs.

amaltaro:
I meant, merge those 2 methods that are loading the same view `jobStatusByRequest`, they are:
 def getElementsCountAndJobsByWorkflow(self, inboxFlag=False, stale=True):
 def getJobStatusByRequest(self):

ticoann:
Ah, I see. I will update that.

PR #7525, opened 2017-01-12
Change harvest multiRun based output file convention - run 999999

### LINKED ISSUE #5643: Name for output root file for multiRun DQM harvesting

amaltaro:
Fixes #5643
Superseeds #6386

It's all written in the issue ticket, but in summary:
 *) when it's `multiRun` harvesting type, we force dqmSaver to use run number `999999`
 *) it also appends the minimum and maximum range number to the processed dataset
 *) it's *only* supported for >= CMSSW_8_0_X. When not supported, the run number edges are still put in the output file name (but not the 999999 run)
 *) Bonus: I refactored the code which checks whether a release support a specific feature or not, such that it can be generally used by other services/features.

### DIFF
--- a/src/python/WMCore/JobSplitting/Harvest.py
+++ b/src/python/WMCore/JobSplitting/Harvest.py
@@ -172,6 +172,14 @@ class Harvest(JobFactory):
         if endOfRun:
             self.currentJob.addBaggageParameter("runIsComplete", True)
         self.mergeLumiRange(self.currentJob['mask']['runAndLumis'])
+
+        # now calculate the minimum and maximum run number, it has to go to the root name
+        minRun = min(self.currentJob['mask']['runAndLumis'].keys())
+        maxRun = max(self.currentJob['mask']['runAndLumis'].keys())
+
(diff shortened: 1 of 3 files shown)


amaltaro:
Just tested it for a real workflow and it works fine. Asking for review now..

amaltaro:
test this please

ticoann (approved):
I am not qualified to review this. :smile:

PR #7526, opened 2017-01-13
fix task and step chain mask

### LINKED ISSUE #7513: fix _maskTaskStepChain call
in src/python/WMCore/ReqMgr/Service/Request.py
related to
W0612, line 272 in Request._maskTaskStepChain: Unused variable 'value'

ticoann:
fixes #7513
Now it returns something like this.

https://reqmgr2-dev.cern.ch/reqmgr2/data/request?name=sryu_TaskChainRunJet2012C_multiRun_reqmgr2_170113_054413_4532&mask=InputDataset&mask=ProcessingString&mask=TaskName

```
{"result": [
 {
  "sryu_TaskChainRunJet2012C_multiRun_reqmgr2_170113_054413_4532": {
    "InputDataset": [
      {
        "Task1": "/JetHT/Run2012C-v1/RAW"
      }, 
      {
        "Task2": null
      }
    ], 
    "ProcessingString": [
      {
        "Task1": "RelVal_TEST_multiRunHarvesting"
      }, 
      {
        "Task2": "RelVal_TEST_multiRunHarvesting"
      }
    ], 
    "TaskName": [
      {
        "Task1": "HLTD"
      }, 
      {
        "Task2": "RECODreHLT"
      }
    ]
  }
}]}
````

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -263,16 +263,16 @@ class Request(RESTEntity):
             # assume mask_key is list if the condition doesn't meet.
             
             for i in range(num_loop):
-                chain = req_dict["%s%s" % (chain_name, i+1)]
+                chain_key = "%s%s" % (chain_name, i+1)
+                chain = req_dict[chain_key]
                 if mask_key in chain:
-                    chain_key = "%sName" % chain_name
-                    masked_dict[mask_key].append({chain_key: chain[chain_key], mask_key: chain[mask_key]})
+                    masked_dict[mask_key].append({chain_key: chain[mask_key]})
                 else:
                     if isinstance(defaultValue, dict):
                         value = defaultValue.get(chain_key, None)
                     else:
                         value = defaultValue
-                    masked_dict[mask_key].append({chain_key: chain[chain_key], mask_key: chain[mask_key]})
+                    masked_dict[mask_key].append({chain_key: value})
         return
                  
     def _mask_result(self, mask, result):


ticoann:
@amaltaro, @vlimant, Alan, Jean-Roch, is output format OK?

amaltaro (approved):
Looks good to me, thanks!

ticoann:
I have to leave one pylint error (accessing private value) as it is.

vlimant:
"InputDataset": [
      {
        "Task1": "/JetHT/Run2012C-v1/RAW"
      }, 
      {
        "Task2": null
      }
    ], 

 is not really practical and is breaking some code in Unified. If there is no strong reason to not have

"InputDataset": ["/JetHT/Run2012C-v1/RAW"]

  then we should stick to it.

hufnagel:
Question, do we allow InputDataset to be non-null for anything but top level tasks ? If we don't, the syntax @vlimant proposes makes sense. For the other things, what do you really gain from the Task1, Task2, etc keys ? Wouldn't a list make more sense ?

{"result": [
 {
  "sryu_TaskChainRunJet2012C_multiRun_reqmgr2_170113_054413_4532": {
    "InputDataset": ["/JetHT/Run2012C-v1/RAW", null], 
    "ProcessingString": ["RelVal_TEST_multiRunHarvesting", "RelVal_TEST_multiRunHarvesting"], 
    "TaskName": ["HLTD", "RECODreHLT"]
  }
}]}

amaltaro:
I'm not sure it works, but there should be the possibility to have two different InputDataset for different tasks.
Though I'm happy to declare it - right now - as not supported :)

amaltaro:
Actually, that's terrible argument :-) There is no way multiple input datasets would work in WorkQueue.
A good one may be that the code Seangchan wrote is generic enough, if we start having specific handling for diffeerent keys, it will quickly become an ugly and hard to maintain code.

A list for InputDataset, MCPileup and DataPileup, would most likely be ok.
But it's *not* ok for other properties like ProcessingString, GlobalTag, etc

ticoann:
@vlimant, Hi Jean-Roch. If in general (not just inputdatasets but other properties), it is not important to specify which tasks it belongs to, I can change the format as you suggested. But as Alan said, if it is just for the specific properties, it might not be good to have different rules for different properties.

Please let me know. format changing is simple.

hufnagel:
Why is a list not ok ?  What's the difference (for any parameter) between

[value1, value2, value3]

and

{ "Task1": value1, "Task2": value2, "Task3": value3 }

both contain the same information since the dict keys don't actually matter.

hufnagel:
Lists are ordered, so relationship between parameters are maintained (all list index 0 fields belong to the first task, all list index 1 fields belong to the second task and so on).

ticoann:
Hi Dirk, yes your right. Both contains the same information. One explicitly and one implicitly.
In that case, we have to allow null values and duplicate values.
If the parameter in Task1 and Task2 has same values we have listed twice and if parameter exists in Task1 but not Task2 we still need to fill up with null(None) value.

It is not a problem of implementing that. I just need an agreement which structure we will use. 
@vlimant, Jean-Roch, since you will be the user for the call, we will let you choose the format. 😄

hufnagel:
You can support tweaks on the format that could allow

"InputDataset": "/JetHT/Run2012C-v1/RAW"

or

"InputDataset": ["/JetHT/Run2012C-v1/RAW"],

it's all a matter in how this is interpreted. In this case it could mean first task only, all other don't have this set for instance. The dict syntax gives you more flexibility in encoding more information if needed, but in the examples shows so far I didn't see anything that needed that flexibility.

If @vlimant is the one using this information, then yes, he should say what he wants and as long as we can encode it that way without ambiguities, that's fine then.

amaltaro:
No Dirk, that's not correct. We may have properties in some tasks and not in others, so one cannot really say if the 2 value belongs to Task1 or Task4 for instance.

BTW, I'm against tweaks :-) We should provide one data format returned and the client handles this as he wants.

hufnagel:
That's why I said the examples shown didn't show why you needed a dict. It still doesn't, having a property not defined for a task can also be handled with None in the list. Anyways, I am not using this information, what matters is what the consumer of this info expects and can handle.

ticoann:
Yes, If we need to know the which property belong to which task. 
We either use 
1. dictionary (as above) or (we can remove None values since we have key).
2. list which could contain None or duplicate values.

If we don't care about task info. we can just have 
3. list with no duplicate values or None,

amaltaro:
@vlimant can you tell us which kind of information you want to see (according to what Seangchan suggests?)? 
a) just unique values regardless of their task/step location in the request
b) all precise values in the request and task/step level
?

vlimant:
there is something tricky indeed about this. will the output of 

https://cmsweb.cern.ch/reqmgr2/data/request?name=pdmvserv_task_B2G-RunIISummer16DR80Premix-01498__v1_T_170117_085932_5825

change to have an InputDataset a the lowest level (as opposed to having it in the Task<N> dict) ?

I don't know why you need that extra value in the dict, maybe for internal consumption, as only fixing #7513 (typo ?).

Please clarify. If for internal only, dict makes sense to me.

ticoann:
Hi Jean-Roch,
> there is something tricky indeed about this. will the output of
> 
> https://cmsweb.cern.ch/reqmgr2/data/request?name=pdmvserv_task_B2G-RunIISummer16DR80Premix-01498__v1_T_170117_085932_5825
> 
> change to have an InputDataset a the lowest level (as opposed to having it in the Task dict) ?

Return format of this API won't change. If inputdataset is defined in both place (workflow level and task level) it will show both. (Currently we are only talking about the mask option for Task and Step level property.

> I don't know why you need that extra value in the dict, maybe for internal consumption, as only fixing #7513 (typo ?).
> 
> Please clarify. If for internal only, dict makes sense to me.

We don't need extra value from dict. I just make it explicit. 
This mask option is not used internally. As I said, it is totally up to you that which format you would prefer, and we will change that. 

Please choose among these three options, (not just for inputdatasets though (for all possible task/step level properties - process string, prep id, etc)) - or if you have better format in mind, please let us know.

1. dictionary (as above) or (we can remove None values since we have key).
2. list which could contain None or duplicate values. (the order of the list shows the which Task/Step it belongs to)
3. list without duplicate values or None value, (If we don't care about associated task/step name)

vlimant:
Hi, I think there is something fishy here. I have not asked for the InputDataset to be made for TaskChain and StepChain. So *where is this change coming from in the first place* ?
#7513 looks unrelated by a long shot.

amaltaro:
InputDataset was just an example. This is all about how we parse the request dictionaries (especially on TaskChain/StepChain) on the server side and give you an useful output.

The simple question remains though, what's your use case using this api with mask.
a) just to know which values you have in the system (e.g., which CMSSW releases we have active in the system)
b) to know which values you have in the system AND where exactly are they placed in each request
?

PR #7528, opened 2017-01-15
move complete transtion in reqmgr2 thread

ticoann:
still need to test.  When cmsweb is upgrade all the agent has to be patched as well.

Remove the state transition functions out of WQ and add in reqmgr2 thread.
Remove the closeWork procedure. DBS3 doesn't allow open block so the function is outdated.
There is no distinction between running-open and running-closed before. 
Keep the both status but has different meaning
running-open  - if any work is in LQ
running-closed - if all the work is in WMBS

Still need to clean up the code.


ticoann:
test this please

ticoann:
Hi Alan, in basic test, it seems working OK. (I still need to fix(or remove) one unittest). If we keep the running-closed definition as in the patch, we have to allow running-open to force-complete transition since we can 't enforce the condition all the jobs are in wmbs before force-complete.

src/python/WMCore/Services/WorkQueue/WorkQueue.py
```diff
@@ -189,27 +220,35 @@ def getElementsCountAndJobsByWorkflow(self, inboxFlag=False, stale=True):
...
-                completedWFs.append(workflow)
-        return completedWFs
+            statusSet = set(data[workflow].keys())
+            status = convertWQElementsStatusToWFStatus(statusSet)
+            if status:
+                workflowsStatus[workflow] = status
+        return workflowsStatus
+    
+    
+    def getWorkflowReqMgrStatus(self, stale=True):
```
amaltaro: It actually gets the status from WorkQueue, so better renaming it

src/python/WMCore/Services/WorkQueue/WorkQueue.py
```diff
@@ -5,6 +5,37 @@
 from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
 
 
+def convertWQElementsStatusToWFStatus(elementsStatusSet):
```
amaltaro: Either I have mental problems, or this is an evil mapping :) It took me quite a bit to digest, I was even wondering if using the "english" methods like `issuperset` wouldn't be better, but I guess it's the same in the end.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -6,9 +6,76 @@
...
+            numOfArchived = 0
+            
+            for requestName, requestInfo in requestsDict.items():
+                
+                if requestInfo.getJobSummary().getTotalJobs() == 0:
+                    for nextStatus in nextStatusList: 
+                        reqDBWriter.updateRequestStatus(requestName, nextStatus)
+                    numOfArchived += 1
+            
+            print("Total %s-archieved: %d", (status, numOfArchived))
```
amaltaro: typo

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -6,9 +6,76 @@
...
+                print("%s in %s moved to %s" % (wf, status, "aborted-completed"))
+            else:
+                for j in range(i+1):
+                    reqDBWriter.updateRequestStatus(wf, nextStatus[j])
+                    print("%s in %s moved to %s" % (wf, status, nextStatus[j]))
+    return
+
+def moveToArchivedForNoJobs(wmstatsService, reqDBWriter):
+    
+    statusTransition = {"aborted-completed": ["aborted-archived"],
```
amaltaro: what about "announced" to "normal-archived" ?

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -6,9 +6,76 @@
...
+    for status, nextStatusList in statusTransition.items():
+        
+        requests = wmstatsService.getRequestByStatus([status], 
+                                            jobInfoFlag = True, legacyFormat = True)
+        
+        print("checking %s workflows: %d" % (status, len(requests)))
+        
+        if len(requests) > 0:
+        
+            requestCollection = RequestInfoCollection(requests)
```
amaltaro: BTW, archiving workflows seem to be much heavier than the other transitions since it has to get a bunch of information from wmstats. Maybe we should define a different polling cycle for archival (every hour or so).

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -6,9 +6,76 @@
 
 from Utils.CherryPyPeriodicTask import CherryPyPeriodicTask
 from WMCore.Services.WMStats.WMStatsReader import WMStatsReader
+from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
 from WMCore.Services.WMStats.DataStruct.RequestInfoCollection import RequestInfoCollection
 
+
+def moveForwardStatus(reqDBWriter, requestStatus):
```
amaltaro: I think a doc string would be welcomed here. As a recall that `requestStatus` contains workflow names and status *based* on the workqueue elements.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -6,9 +6,76 @@
...
+    
+    statusTransition = {"assigned": ["acquired", "running-open", "running-closed", "completed"],
```
amaltaro: Ok, I now see it's not the same allowed transitions mapping. Though it may be better to set this specific status transition in the same RequestStatus file.

amaltaro (changes requested):
Besides a few issues, it looks good so far. However, don't forget running a code reformat in the end (especially in the cherrypy thread). Comments could be free of typos as well ;)

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: Yes, it is not the sam state transition path (In this case there is one path which WorkQueue can control).
You are right about moving this to RequestStatus file. I will do that.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: I will add the comments.

ticoann (commented):
Yes, mapping is confusing that is why I added comment. For me it is more clear with operator then English method like issupperset since it shows which set is bigger. :smile: but if it is more clear to you I will change that.

The code still need to be cleaned up. The main thing I want your advice  is the mapping between WQE status and request status. I updated the comments. Hope this is more clear how the status is defined. (I am not sure whether that is good definition, though. If you don't agree about mapping we can discuss and modify)

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -5,10 +5,65 @@
...
+def moveForwardStatus(reqDBWriter, wfStatusDict):
+    
+    for status, nextStatus in AUTO_TRANSITION.iteritems():
+        requests = reqDBWriter.getRequestByStatus([status])
+        for wf in requests:
+            stateFromGQ = wfStatusDict.get(wf, None)
+            # special case for failed workflow
+            if stateFromGQ is None:
+                continue
+            elif stateFromGQ == "failed" and status == "assigned":
```
amaltaro: I'd add a `elif stateFromGQ == status: continue`. To avoid spending cycles when the status is unchanged.

amaltaro (commented):
Seangchan, besides the fragility of this WQ status into ReqMgr status, I cannot see any flaw in your logic.
Still, this is a legit case to have unit tests checking all sorts statusSet from WQ :-)

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: Actually it will be caught in the ValueError at the bottom in that case since there will be no next state, bu t i can add that for clarity.

PR #7530, opened 2017-01-17
Added workqueue views for childQueue, priority and status

amaltaro:
These views are going to be use for WorkQueue monitoring.

I've also removed a couple of childQueues* views that apparently are not being use anywhere in WMCore (hmm, there are a couple of tests for them).

@ticoann should we keep these views that I removed?


ticoann:
I think the childQueues can be gone, I see that you replaced one view. However, unittest need to be updated.
I was wondering whether we can combine all three views and do the calculation in the client.

 emit([ele["Status"], ele["ChildQueueUrl"], ele["Priority"],], ele["Jobs"]);

amaltaro:
Yes, the unittests need to be removed as well (or updated).

>  I was wondering whether we can combine all three views and do the calculation in the client.

Good question :) We could likely reduce these 4 new views into 1 or 2 only, though we'd have more python code to handle this info and set a proper format/stats for each metric we want to monitor.

I still think letting most of this computation to the `stats` builtin reduce function is cleaner. Since the workqueue db is not big (< 100k valid docs), these views shouldn't take too much space.

Anyways, if you think we should have a single view and then parse this information on the client, that's fine as well.

ticoann:
Hi Alan, yes you are right, the view in workqueue shouldn't be big. It might be cleaner and faster to have separate view.

amaltaro:
test this please

amaltaro:
@ticoann I think it's worth a review now. Just don't merge please, I still want to squash and run a pylint cleanup.

ticoann (changes requested):
I am not sure monitoredDetailsByStatus view is needed. It looks like it retrieves all the elements in db and organized by status.
If we really need all the data we don't have to have a view. We can just get all documents. I am not sure though how much memory is needed. (We can stream as well).  But I was wondering why we need whole elements for monitoring.

amaltaro:
I updated the monitoredDetails view to only return details for elements in Available/Acquired/Negotiating. I was fetching all the status before because I was not sure whether we would expand this monitoring to Running and etc in the near future. But if we need that, then we can change it later.

About the memory footprint, this view returns around 1/3 of the real element size. The average size in memory - of course it may vary quite a bit - seems to be around 700B. So 10k elements would make it 7MB in memory. I could make the API return a generator instead, but then we would need to fetch these elements twice, since we calculate stuff in two different ways in `getGlobalSiteStatusSummary`.

Let me know if you have further suggestions please.

ticoann:
Thanks Alan, I think it would be small enough memory then.
As far as the view goes, I was wondering whether we really need a new view.

We can use this view for the same purpose.
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/views/elementsByStatus/map.js

by passing the option keys=["Available", 'Negotiating', "Acquired"]&include_docs=True.

It will get a little more data but we don't need additional view and save workqueue db space as well.

amaltaro:
If we use this other view, then the memory footprint will be around 3x bigger, for retrieving the whole docs. Can I go ahead and update it or we stick with the specific view created?

ticoann:
I prefer to use existing view. I think memory is OK ~21M. One other thing is that it is more flexible to age other statistics if we need to gather other properties.

amaltaro:
test this please

amaltaro:
@ticoann it's ready for your review again and, if it's fine, then please merge.

PR #7532, opened 2017-01-18
Fix WorkQueueElement location/restriction methods

amaltaro:
Mostly fixed issues in `possibleSites()` method and make tons of tests for that method and the site passing restriction method.
Ideally we should have the same tests for https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/lists/workRestrictions.js , eventually :-)

### DIFF
--- a/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
+++ b/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
@@ -216,12 +216,13 @@ class WorkQueueElement(dict):
                     PercentSuccess=self['PercentSuccess'])
 
     def passesSiteRestriction(self, site):
-        """Takes account of white & black list, and input data to work out
-        if site can run the work"""
+        """
+        _passesSiteRestriction_
+
+        Given a site name, checks whether it passes the site and data restrictions.
+        """
         # Site list restrictions
-        if self['SiteWhitelist'] and site not in self['SiteWhitelist']:
-            return False
-        if site in self['SiteBlacklist']:
+        if site in self['SiteBlacklist'] or site not in self['SiteWhitelist']:
             return False
 
         # input data restrictions (TrustSitelists flag)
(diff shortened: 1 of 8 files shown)


amaltaro:
test this please

amaltaro:
I don't see why it was not failing before. However, now they fail basically because the elements don't have a SiteWhitelist set, thus the getWork, pullWork, getAvailableWork, etc calls no longer acquire those elements. I'm updating those unittests. 

@ticoann a workqueue element without any SiteWhitelist is not supposed to be acquired (meaning, workflow assigned to nowhere - or everywhere, though I'd rather be explicit -  right?

ticoann:
Workflow without site white list should be acquired unless trusted site flag is set to true (or no blocks available in any sites)

amaltaro:
Thanks, Seangchan! So the use cases would be:
 *) for MC, it would go to ALL resources that are available.
 *) Redigi/others, it would go to where the input block is available
?

I'm still not convinced this is a good idea, since you have no clue of where your jobs may end up.

ticoann (approved):
Patch looks good.

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
```diff
@@ -243,33 +240,20 @@ def passesSiteRestriction(self, site):
...
-    def intersectionWithEmptySet(self, a, b):
-        """
-        interaction of 2 sets but if one of the set is empty returns union
-        """
-        if len(a) == 0 or len(b) == 0:
-            return a | b
-        else:
-            return a & b
-
     def possibleSites(self):
```
amaltaro: @ticoann as we just discussed, we are/were going to make an empty SiteWhitelist as "consider ALL sites", as it's right now AFAICT.
After staring at the code for 10min :-) I noticed there is one use case that we cannot really distinguish:
 i) a MonteCarlo element would NOT have Inputs, ParentData nor PileupData, thus only the SiteWhitelist that would be empty. Value returned by this function is `[]` --> meaning run everywhere.
 ii) any other element with either Input/Parent/Pileup should follow the intersection of data location. Thus it could return `[]` --> meaning no possible location.

So we'll have this inconsistency :( Maybe we should just make SiteWhitelist mandatory and non empty during assignment. Checking with Jean-Roch to see if they have ever used it....

ticoann: Yes, you are right it might be better to make SiteWhitelist mandatory. Let's wait what Jean-Roch says.

amaltaro:
And this commit should fix several workqueue unit tests... I hope :)

ticoann:
Patch looks good to me. Let me know if you want to merge this. (I think a few pylint errors can be fixed).

amaltaro:
Thanks, Seangchan. I fixed some pylints and it can be merged now. We may need to update one of our other PRs due to conflicts though :)

PR #7533, opened 2017-01-19
add state transition information in wmstats

### LINKED ISSUE #6840: Show request transition in wmstats
in human readable way time.asctime(...) so that one can see what the workflow has gone through

ticoann:
fixes #6840


ticoann:
@amaltaro, @vlimant, Alan, Jean-Roch. Is that good enough or display somewhere else?

screen shot 2017-01-19 at 3 26 42 pm

vlimant:
cute. but then it's the only place one can read it, right ?

ticoann:
> cute. but then it's the only place one can read it, right ?

We can add other places, if you prefer somewhere else.

PR #7542, opened 2017-01-24
fix aborted-completed to aborted-archived

ticoann:
When there is no WQE exists. Tested in vm

### DIFF
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
@@ -6,7 +6,6 @@ from __future__ import (division, print_function)
 
 from Utils.CherryPyPeriodicTask import CherryPyPeriodicTask
 from WMCore.ReqMgr.DataStructs.RequestStatus import AUTO_TRANSITION
-from WMCore.Services.WMStats.WMStatsReader import WMStatsReader
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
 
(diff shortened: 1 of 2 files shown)


ticoann:
I don't know why following unittest fails. I will create the separate issue.

    WMCore_t.Services_t.LogDB_t.LogDB_t.LogDBTest:test_thread_name changed from error to failure
    WMCore_t.WorkQueue_t.WorkQueue_t.WorkQueueTest:testProcessingWithPileup changed from error to failure

amaltaro:
Don't we need to update the agents to clean all the information as well?

ticoann:
I  don't think we have to redeploy agent. It just affects if the workflows which are not already moved to aborted-completed.  Not applying patch means it doesn't wait for all the agent to kill the jobs before it moves to aborted-completed which current agents do.

amaltaro:
So the current agent version will cleanup the workflow in one agent only (the first one to move from aborted to aborted-completed), right? The other agents working on the same workflow will not clean any of the databases (?)

How about the new agent, when the db cleanup would happen? Once the agent sees that workflow in aborted?

ticoann:
> So the current agent version will cleanup the workflow in one agent only (the first one to move from aborted to aborted-completed), right? The other agents working on the same workflow will not clean any of the databases (?)

Other agent still cleans up and kill the jobs. (That is the patch we already applied. Periodically killing the job).  Only difference is that without this patch, aborted-completed doesn't mean all the jobs are killed in all the agent. (Only one of the agent killed all the jobs although eventually they will get killed).
DB clean up happens after aborted-completed, after db is cleaned up, it will move to aborted-archived. 

But aborted archived status still has the same problem. If one of the agent clean up db, it will move to aborted-archived status.

PR #7544, opened 2017-01-24
Fix down_component for agentInfo view

amaltaro:
Seangchan, thanks for spotting this issue.
In principle "down_components" has no usage for global workqueue. Though we could extend it - eventually - to also report possible problems with the threads.

If you prefer to keep this key around, then we just close this PR. Otherwise this would fix the problem.


amaltaro:
Sigh, I noticed this global_workqueue doc uploaded to couch was also breaking reqmgr2 when pulling the list of teams, my bad :(
This should fix it.

PR #7550, opened 2017-01-26
Fix getConfig in ReqMgr2 UI

amaltaro:
ReqMgr service always returns a list.
Also fixed a print problem in the cherrypy thread, it seems I'm not the only one mixing up those syntax :-)

Already tested in my VM.


ticoann (approved):

ticoann:
Yes, print and log syntax are confusing. Sorry. Anyway, I will remove all the print for the thread and replace with log, it seem print message goes to the main thread not the specific thread.

PR #7553, opened 2017-01-27
Handle Tier0 case for when aborted/force-complete is set

### LINKED ISSUE #7551: latest JobSubmitter broken for Tier0 use
Putting a check for ReqMgr2 into the JobSubmitter broke that component for Tier0 use.

https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L109

When you put any referral back to ReqMgr2 into deeper agent code it's a) quite bad design-wise and b) without some provisions for it to be optional it'll break the Tier0.

ticoann:
fixes #7551

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -105,9 +105,17 @@ class JobSubmitterPoller(BaseWorkerThread):
 
         # Keep a record of the thresholds in memory
         self.currentRcThresholds = {}
-
-        self.reqmgr2Svc = ReqMgr(self.config.TaskArchiver.ReqMgr2ServiceURL)
-        self.abortedAndForceCompleteWorkflowCache = self.reqmgr2Svc.getAbortedAndForceCompleteRequestsFromMemoryCache()
+        
+        self.useReqMgrForCompletionCheck = getattr(self.config.TaskArchiver, 'useReqMgrForCompletionCheck', True)
+        
+        if self.useReqMgrForCompletionCheck:
+            # only set up this when reqmgr is used (not Tier0)
+            self.reqmgr2Svc = ReqMgr(self.config.TaskArchiver.ReqMgr2ServiceURL)
+            self.abortedAndForceCompleteWorkflowCache = self.reqmgr2Svc.getAbortedAndForceCompleteRequestsFromMemoryCache()
+        else:
+            # Tier0 Case - just for the clarity (This private variable shouldn't be used
+            self.abortedAndForceCompleteWorkflowCache = None
+            
         return
 
     def getPackageCollection(self, sandboxDir):
@@ -229,7 +237,14 @@ class JobSubmitterPoller(BaseWorkerThread):
            self.refreshPollingCount >= self.skipRefreshCount:
             newJobs = self.listJobsAction.execute()
             self.refreshPollingCount = 0
-            abortedAndForceCompleteRequests = self.abortedAndForceCompleteWorkflowCache.getData()
+            
+            if self.useReqMgrForCompletionCheck:
+                # if reqmgr is used (not Tier0 Agent) get the aborted/forceCompleted record
+                abortedAndForceCompleteRequests = self.abortedAndForceCompleteWorkflowCache.getData()
+            else:
+                #T0Agent
+                abortedAndForceCompleteRequests = []
+                
             logging.info("Found %s new jobs to be submitted.", len(newJobs))
         else:
             self.refreshPollingCount += 1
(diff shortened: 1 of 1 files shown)


ticoann:
@hufnagel, @amaltaro, Dirk Alan please review.  Dirk Could you patch this when you run the Tier0 test. If it is OK, I will create the new tag on Monday.

ticoann:
test this please

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -105,9 +105,14 @@ def __init__(self, config):
...
-
-        self.reqmgr2Svc = ReqMgr(self.config.TaskArchiver.ReqMgr2ServiceURL)
-        self.abortedAndForceCompleteWorkflowCache = self.reqmgr2Svc.getAbortedAndForceCompleteRequestsFromMemoryCache()
+        
+        self.useReqMgrForCompletionCheck = getattr(self.config.TaskArchiver, 'useReqMgrForCompletionCheck', True)
+        
+        if self.useReqMgrForCompletionCheck:
+            # only set up this when reqmgr is used (not Tier0)
+            self.reqmgr2Svc = ReqMgr(self.config.TaskArchiver.ReqMgr2ServiceURL)
+            self.abortedAndForceCompleteWorkflowCache = self.reqmgr2Svc.getAbortedAndForceCompleteRequestsFromMemoryCache()
```
amaltaro: I think we should add a
```
else:
    self.abortedAndForceCompleteWorkflowCache = []
```

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -225,8 +230,8 @@ def refreshCache(self):
 
         logging.info("Refreshing priority cache with currently %i jobs", len(self.cachedJobIDs))
 
-        if self.cacheRefreshSize == -1 or len(self.cachedJobIDs) < self.cacheRefreshSize or \
-           self.refreshPollingCount >= self.skipRefreshCount:
+        if self.useReqMgrForCompletionCheck and self.cacheRefreshSize == -1 or \
+           len(self.cachedJobIDs) < self.cacheRefreshSize or self.refreshPollingCount >= self.skipRefreshCount:
             newJobs = self.listJobsAction.execute()
             self.refreshPollingCount = 0
             abortedAndForceCompleteRequests = self.abortedAndForceCompleteWorkflowCache.getData()
```
amaltaro: and then remove `self.useReqMgrForCompletionCheck` from the `if` above, otherwise T0 will never refresh the cache. Then add another `if self.useReqMgrForCompletionCheck` here to update `abortedAndForceCompleteRequests`

amaltaro (changes requested):

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: We can't do this 
 self.abortedAndForceCompleteWorkflowCache is not list it has method to .getData(), although self.abortedAndForceCompleteWorkflowCache will be never used in else case.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: Yes, you are right. I will update the patch

hufnagel:
Ok, I can't test this since something else was broken in the last day or so. Every time I try to start JobSubmitter now (with or without this patch) it dies with (in the component stderr)

INFO:root:JobSubmitter.preInitialization
INFO:root:Setting poll interval to 30 seconds
Configuration Error Line 58 while reading global config source /etc/condor/condor_config
unexpected keyword 'ifexist command into $(LOCAL_CONFIG_DIR)/49ec2-instance.config' after include

Completely unrelated I am sure, but it means I can't test this patch...

hufnagel:
Ok, problem with the new 8.5.7 condor shedd which doesn't seem to work with WMAgent... arukh implemented a workaround for now.

JobSubmitter starts up now with this patch, after I added to the config

config.TaskArchiver.useReqMgrForCompletionCheck = False

Should this be added to wmagent-mod-config ?

ticoann:
Thanks Dirk for testing. Why weekend though. 😄?

>config.TaskArchiver.useReqMgrForCompletionCheck = False
> Should this be added to wmagent-mod-config ?

I think it is already set.

https://github.com/dmwm/T0/blob/master/etc/Tier0Config.py#L31

ticoann:
@amaltaro, Alan, Dirk tested this patch with Tier0. If changes are OK (I updated bug you pointed out), I will merge this.

amaltaro:
Code looks good now. Thanks

ticoann:
test this please

ticoann:
test this please

PR #7554, opened 2017-01-27
_stats builtin reduce function requires a number

amaltaro:
And that's why workqueue_inbox views were breaking...

### DIFF
--- a/src/couchapps/WorkQueue/views/jobsByChildQueueAndPriority/map.js
+++ b/src/couchapps/WorkQueue/views/jobsByChildQueueAndPriority/map.js
@@ -1,6 +1,10 @@
 function(doc) {
     var ele = doc["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
     if (ele) {
-        emit([ele["ChildQueueUrl"], ele["Priority"]], ele["Jobs"]);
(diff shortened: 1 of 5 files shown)


amaltaro:
I'm merging it to cut a new tag.

ticoann:
Hmm, I though default is changed to 0. Is this workqueue element from previous injected workflow?
Anyway global queue inbox to have Job field doesn't make much sense. Maybe we need to skip as count than convert this to 0.

How about something like this it will filter out undefined case too
```
if (ele["Jobs"] > -1) {
     emit([ele["Status"], ele["Priority"]], ele["Jobs"]);
}
```

amaltaro:
Hi Seangchan, that's right. The new default value was changed to 0, but there were old elements in testbed (with null value) and that's why the view indexing was breaking, my bad :(

For global workqueue, we don't really use any information from workqueue_inbox, since all we care is in inbox.

I'm not sure, but I think checking if a var is null is processing cheaper than > -1. IMO I'd just leave it as it is (actually, in ~3 months from now we could simply remove that check, since ALL docs should have an integer value).

ticoann:
Of course, it is not your fault. :smile:. Yes you're right since it will be clean up, we can just remove the code later. Thanks,

PR #7555, opened 2017-01-27
Apply couchdb file descriptor patch; use SimpleCondorPlugin

amaltaro:
Patch couchdb with the file descriptor fixes.
Set up the agent to use SimpleCondorPlugin instead of PyCondor. Once we get to the moment that we have to deploy the new agent in production, we can update the WMAgent config file and remove one line from here.

@ticoann no need to merge it now, but in case you set up your VM or testbed, please use it to exercise as much as we can the new plugin.

### DIFF
--- a/bin/deploy-wmagent.sh
+++ b/bin/deploy-wmagent.sh
@@ -24,8 +24,8 @@
 ### Usage:               -n <agent_number> Agent number to be set when more than 1 agent connected to the same team (defaults to 0)
 ### Usage:
 ### Usage: deploy-wmagent.sh -w <wma_version> -c <cmsweb_tag> -t <team_name> [-s <scram_arch>] [-r <repository>] [-n <agent_number>]
-### Usage: Example: sh deploy-wmagent.sh -w 1.0.21.patch3 -c HG1610f -t production -p "7399" -n 2
-### Usage: Example: sh deploy-wmagent.sh -w 1.0.21 -c HG1610f -t testbed-cmssrv214 -s slc6_amd64_gcc493 -r comp=comp.amaltaro
+### Usage: Example: sh deploy-wmagent.sh -w 1.1.0.patch1 -c HG1702f -t production -p "7605" -n 2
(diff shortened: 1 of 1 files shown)


bin/deploy-wmagent.sh
```diff
@@ -230,9 +230,11 @@ cd -
 fi
 echo "Done!" && echo
 
-### Enabling couch watchdog:
+### Enabling couch watchdog; couchdb fix for file descriptors
 echo "*** Enabling couch watchdog ***"
 sed -i "s+RESPAWN_TIMEOUT=0+RESPAWN_TIMEOUT=5+" $CURRENT/sw*/$WMA_ARCH/external/couchdb*/*/bin/couchdb
+sed -i "s+exec 1>&-+exec 1>$CURRENT/install/couchdb/logs/stdout.log+" $CURRENT/sw*/$WMA_ARCH/external/couchdb*/*/bin/couchdb
+sed -i "s+exec 2>&-+exec 2>$CURRENT/install/couchdb/logs/stderr.log+" $CURRENT/sw*/$WMA_ARCH/external/couchdb*/*/bin/couchdb
```
ticoann: Hi Alan, since we will be deploying agent after cmsweb deployment.
I think it is ok to change it now.

```
exec 1>/tmp/couch-stdout.log
exec 2>/tmp/couch-stderr.log
```

ticoann (changes requested):

bin/deploy-wmagent.sh
amaltaro: And since that change was reverted back, I think I can just keep it as it is now :-)

bin/deploy-wmagent.sh
```diff
@@ -268,6 +270,7 @@ echo "Done!" && echo
 echo "*** Tweaking configuration ***"
 sed -i "s+team1,team2,cmsdataops+$TEAMNAME+" $MANAGE/config.py
 sed -i "s+Agent.agentNumber = 0+Agent.agentNumber = $AG_NUM+" $MANAGE/config.py
+sed -i "s+BossAir.pluginNames = \['PyCondorPlugin+BossAir.pluginNames = \['SimpleCondorPlugin+" $MANAGE/config.py
```
amaltaro: I'm going to update it directly in the default values, so it won't be needed at this stage.

ticoann (approved):

PR #7556, opened 2017-01-27
Add RHEL7 value for condor JDL.

bbockelm:


ericvaandering:
This is nice for Andrew or whoever to have but don't merge this.

amaltaro:
Since SL5 worker nodes are close to extinction, how about we map scram slc5_* to `rhel6`? Then the `any` value can disappear, since SL7 worker nodes cannot run anything but slc7_* jobs

bbockelm:
I think it's useful to keep the concept of `any` in the system; we have another year or so before we need to start thinking about RHEL8.

amaltaro:
My concern is about rhel7, not rhel8. If we want to keep `any` in the system, then I think it's definition will have to slightly change in glideinWMS/factory, i.e., any would map only to rhel5 and rhel6, since rhel7 cannot run any other jobs than slc7_*

amaltaro:
@bbockelm please map slc5 to rhel6 (can be another elif) and then we can push this patch to the relval agent and get it out of our plate, before workflows go in and the agent is not patched.

amaltaro:
Thanks, Brian. This patch was pushed in to the RelVal agent (vocms026)

amaltaro:
It seems some steps have None scramArch, which was breaking the relval agent. I've just updated the code in the node and here is a diff (Brian, can you update this PR to reflect that please?):
```
cmst1@vocms026:/data/srv/wmagent/current/sw/slc6_amd64_gcc493/cms/wmagent/1.0.21.patch1/lib/python2.7/site-packages/WMCore/BossAir/Plugins $ diff -rup PyCondorPlugin.py{.bkp,} | colordiff
--- PyCondorPlugin.py.bkp	2017-02-03 20:16:17.986246444 +0100
+++ PyCondorPlugin.py	2017-02-03 20:22:03.272911460 +0100
@@ -967,7 +967,9 @@ class PyCondorPlugin(BasePlugin):
         jdl.append('request_cpus = %s\n' % numberOfCores)
 
         # Add OS requirements for jobs
-        if job.get('scramArch', '').startswith("slc6_") or job.get('scramArch', '').startswith("slc5_"):
+        if not job.get('scramArch'):
+            jdl.append('+REQUIRED_OS = "any"\n')
+        elif job.get('scramArch').startswith("slc6_") or job.get('scramArch').startswith("slc5_"):
             jdl.append('+REQUIRED_OS = "rhel6"\n')
         elif job.get('scramArch', '').startswith("slc7_"):
             jdl.append('+REQUIRED_OS = "rhel7"\n')
```

amaltaro:
And since we have the new WMAgent production release in production now - which supports RHEL7 as well - there is no more need for this PR.
Thanks, Brian! I'm closing this one.

PR #7562, opened 2017-01-30
Remove unneeded/legacy templates; minor updates

amaltaro:


amaltaro:
As reported here:
https://github.com/dmwm/WMCore/issues/7572

Group is still needed in these requests otherwise it may fail ACDCs. Adding it here until that issue gets fixed.

PR #7565, opened 2017-01-30
remove port from agent url

ticoann:
This was initially added that assuming there could be multiple agents are running in the same machine. Which is not the case so far. Removing it to remove the dependency from WMBSService component.


amaltaro (commented):
So this agent_url is the id/key of the docs, right? If so, sounds good. We'll likely have to remove the old docs once this goes into production.

ticoann:
> So this agent_url is the id/key of the docs, right?  If so, sounds good. We'll likely have to remove the old docs once this goes into production.

Hi Alan, Yes and Yes,

PR #7567, opened 2017-01-31
prevent exceptions printout in TempDirectory

hufnagel:
Deletion of objects in python does not guarantee any particular order, so what is attempted here simply does not work reliably. In this particular case, the shutil module is already unloaded when TempDirectory.__del__ is called. Avoid the exception printed to stderr. There is no real way to fix this in a clean way.

I checked and we use __del__ overloading in a handful of places in WMCore. I would suggest to open an issue about this since none of them likely do what we think they do.

### DIFF
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -249,7 +249,7 @@ class Requests(dict):
             try:
                 response, result = conn.request(uri, method=verb,
                                                         body=encoded_data, headers=headers)
-            except AttributeError as ex:
+            except AttributeError:
                 msg = traceback.format_exc()
                 # socket/httplib really screwed up - nuclear option
                 conn.connections = {}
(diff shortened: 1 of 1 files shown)


hufnagel:
Oh, and btw, this only shows up for the Tier0 since the WMAgent components that use this all seem to have WMCORE_CACHE_DIR set and don't use this code path. The Tier0Feeder doesn't or at least some of the code it calls doesn't.

hufnagel:
Ok, still investigating some issues here...

hufnagel:
Ok, this looks better. Ready for review.

ericvaandering:
On possibility for this type of thing is to turn it into a context manager which has a defined __exit__ run at a defined time. In fact, I thought we had one already

hufnagel:
Yeah, looked at that too. But it wasn't obvious to me how this would apply to the current design where you create the temp directory in really low level code, pass a reference back that is then used in high level code.

If someone has an idea on how to do that...

hufnagel:
Yes, I would have to completely rewrite the design of this to use a context manager. Not going to do that but if someone wants to put in an issue feel free. General issue to review our use of __del__ overloading would likely also be a good idea. Please merge.

amaltaro:
@ticoann please merge

PR #7568, opened 2017-02-01
Fix abort cascade

amaltaro:
If we abort a workflow with cascade:true, then the wq elements are not canceled. This should fix it, I still need to test though.

@ticoann you may want to have a quick look at the second commit as well, besides pylint fixes, I have also removed a few functions that were dead (not being used anywhere).


amaltaro:
Tests are Ok. Seangchan, please review and merge it for today/tomorrow testbed.

ticoann (approved):

ticoann:
As mentioned above there is one thing I am not completely sure whether it works. But I will run the test with this first.

PR #7573, opened 2017-02-02
another debug message for workqueue thread crash (Never merge)

ticoann:


ticoann:
OK, Thanks to Alan, problem was found. Closing this.

PR #7574, opened 2017-02-02
Break run-lumi mapping into smaller slices for DB insertion

amaltaro:
Fix an issue with WorkQueueManager where one of its thread (WorkQueueManagerWMBSFileFeeder) stays forever trying to insert the run/lumi mapping rows into the database.

Creating the binds list is quite straight forward (took < 0.5 secs for 500k lumis in my interactive tests). This operation is likely heavier on Oracle because it first has to check whether that row already exists... so setting it to 10k only to be on the safe side (the case it was getting stuck was .5 million).

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/AddRunLumi.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/AddRunLumi.py
@@ -8,8 +8,8 @@ MySQL implementation of AddRunLumi
 from Utils.IterTools import grouper
 from WMCore.Database.DBFormatter import DBFormatter
 
-class AddRunLumi(DBFormatter):
 
+class AddRunLumi(DBFormatter):
     sql = """insert dbsbuffer_file_runlumi_map (filename, run, lumi)
             select id, :run, :lumi from dbsbuffer_file
             where lfn = :lfn"""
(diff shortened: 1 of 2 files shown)


ticoann (approved):

ticoann:
Thanks Alan, Patch looks good.
As we discussed we will invest how dbs handles this. but right now, we will update the agent with this patch.

https://github.com/dmwm/DBS/blob/9b5fdb8ae4f95ec39e63dcf4bc2485c8d6e7f338/Server/Python/src/dbs/utils/DBSDaoTools.py
https://github.com/dmwm/DBS/blob/9b5fdb8ae4f95ec39e63dcf4bc2485c8d6e7f338/Server/Python/src/dbs/utils/DBSDaoTools.py

amaltaro:
@ticoann @yuyiguo just for the record, I ran a workflow in testbed with this patch (and oracle backend) and all the rows were properly inserted into the table (10k slices). I double-checked this information in the table and it looks ok.

Looking at the WorkQueueManager log, it seems this operation (+ all the other wmbs operations, I didn't separate them) took around 20secs.
```
2017-02-07 10:04:10,973:139635463386880:INFO:WMBSHelper:"amaltaro_MonteCarlo_BadLumi_Agent110_Validation_170207_094401_7605" Injecting production 1:1:1 - 1:14074:140731 (run:lumi:event) into wmbs
2017-02-07 10:04:11,091:139635463386880:INFO:WMBSHelper:WMBS File: MCFakeFile-amaltaro_MonteCarlo_BadLumi_Agent110_Validation_170207_094401_7605-Production-b777c1ebc253ce7e2f17bbe261acdc57 on Location: set(...])
2017-02-07 10:04:33,115:139635463386880:INFO:WorkQueue:Created top level subscription 513 for amaltaro_MonteCarlo_BadLumi_Agent110_Validation_170207_094401_7605 with 1 files
```

I'm not sure whether the dbsbuffer information is also populated at this stage, if it's then the real time should be half, since we have the same run/lumi mapping inserted into dbsbuffer.

Bottom line is, it works as expected :-)

ticoann:
Thank you very much Alan. Great job! :-)

PR #7576, opened 2017-02-02
Retrieve all non-deprecated CMSSW releases in all ScramArchs available

amaltaro:
Shahzad has just created this arch option (anyarch) and now we can retrieve a CMSSW release in all its ScramArch (before it was showing only the production one).

This is needed if we want to run slc7_* workflows.

### DIFF
--- a/src/python/WMCore/Services/TagCollector/TagCollector.py
+++ b/src/python/WMCore/Services/TagCollector/TagCollector.py
@@ -7,14 +7,14 @@ from WMCore.Services.TagCollector.XMLUtils import xml_parser
 class TagCollector(Service):
     """
     Class which provides interface to CMS TagCollector web-service.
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

ticoann (approved):

ericvaandering:
So he added a new category of architecture? Is this really what we want as a permanent change? I suppose if we don't get the requests for them, it doesn't matter if they are there.

ticoann:
I think adding it won't hurt but requestor need to be careful not to assign that accidentally.

amaltaro:
The previous script was not able to yield the same CMSSW in different architectures (only the production one). So he added this option such that the script building this list can yield a CMSSW release in all the architectures it was built on.

Otherwise, if we want to inject a request using a new architecture - that has not been made a production arch yet - it's just not possible.

PR #7579, opened 2017-02-02
Scram by os

ericvaandering:
This pull request deals with the case where the spec has a list of arches that can be run.

First commit is renaming IterTools so it doesn't clash with the built-in itertools on OS X. 
Second commit is a context manager for temporary environments used in the testing
Third commit is the meat of the PR with unit tests.


ericvaandering:
@bbockelm (and @hufnagel) Can you check my change to WMTweak? My understanding is that the condor ad file is going to have string values encased in double quotes so I strip them off to make a nice python like dictionary out of the results. I actually don't understand why https://github.com/ericvaandering/WMCore/commit/73db6bb70f8fb7133cb4c3a27383e5b70123b822#diff-8d59623f30f8f60048722cb843d1bc8eR601 works (if it does).

ericvaandering:
test this please

ericvaandering:
@ticoann pending reviews, this can go in. Should be completely backwards compatible.

bbockelm:
@ericvaandering - ClassAd values are typed, meaning that some values will be a string (`"foo"`) but others won't (`true`).  In the line you reference, it's a boolean, meaning there is no double-quotes necessary.

Since the other callers of the `readAdValues` function expect the raw contents, I would suggest *not* stripping the double-quotes where you did and move that logic to `getSingleScramArch`

src/python/PSetTweaks/WMTweak.py
```diff
@@ -570,7 +567,7 @@ def readAdValues(attrs, adname, castInt=False):
                         except ValueError:
                             print("Error parsing %s's %s value: %s", (adname, attr, val))
                     else:
-                        retval[attr] = val
+                        retval[attr] = val.strip('"')
```
bbockelm: Not necessarily correct here.  Would advise to move this to the call site (`getSingleScramArch`).

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -95,6 +97,8 @@ def execute(self, emulator=None):
         logging.info('User files are %s', userFiles)
         logging.info('User sandboxes are %s', userTarball)
 
+        scramArch = self.getSingleScramArch(scramArch)
```
bbockelm: This may return `None`, right?  Is that handled correctly?  Is it better to fail the job instead?

ericvaandering: Yeah, it can. Not sure. I could also raise an exception.

bbockelm (changes requested):
Two small items for the code review.

Unrelated pylint formatting changes make it difficult to determine the intended functional changes and frustrating to review.  Would be helpful to separate those to a separate PR in the future.

ericvaandering:
They are in different commits for exactly that reason. That's how we roll in DMWM.

hufnagel:
Still not perfect since sometimes functional and related changes are spread across multiple commits. Guess it's hard or impossible to rebase against anything by the most recent commit.

hufnagel:
Also, this mode is not compatible with the github review process since that shows you all the changes at once. At least I didn't find a way to split them except to manually check all he individual commits. Which then leads to problem if functional changes are split across different commits.

Anyways, I never bother to use the review process anyways because of this. Code changes look ok.

ericvaandering:
<meta>No, I've had probably 15 different commits in my branch, reordered, grouped, inserted in between and squashed and so forth. You can rebase -I as much as you want.

ericvaandering:
OK, this addresses @bbockelm first issue, but not the return value. @amaltaro what do you think?

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
amaltaro: We could raise an exception, but honestly, if this was a problem then we'd have seen this problem at least once in the last 5 years :)
Though I think it would be good to have a logging line before the cmssw bootstrap (L121) just printing the scram command + release + arch

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -324,6 +328,33 @@ def post(self, emulator=None):
 
         return None
 
+    @staticmethod
+    def getSingleScramArch(scramArch):
+        """
+        Figure out which scram arch is compatible with both the request and the release
```
amaltaro: Did you mean "with both the request and the worker node OS"?

src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py
```diff
@@ -324,6 +328,33 @@ def post(self, emulator=None):
...
+            try:
+                ad = readAdValues(['glidein_required_os'], 'machine')
+                runningOS = ad['glidein_required_os'].strip('"')
+                validArches = sorted(OS_TO_ARCH[runningOS], reverse=True)
+                for requestedArch in sorted(scramArch, reverse=True):
+                    for validArch in validArches:
+                        if requestedArch.startswith(validArch):
+                            return requestedArch
+            except KeyError:
+                return sorted(scramArch)[-1]  # Give the most recent release if lookup fails
```
amaltaro: It may be worth to log this error, with runningOS and scramArch.
In addition to that, I'd log the case we have no match (return None).

amaltaro (changes requested):
I've just noticed I mixed up single comment with review :-)
Code looks good in general, though I have a couple of logging suggestions that will help us once we need to debug it.

ericvaandering:
Thanks @amaltaro . Updated with your suggestions.

amaltaro:
Testing it on top of pre10

amaltaro:
@ticoann also good to be merged

ticoann:
@ericvaandering, Eric, Sorry could you fix the conflicts?

ericvaandering:
Conflicts fixed. @amaltaro please mark as reviewed.

ericvaandering:
test this please

amaltaro (approved):

amaltaro:
blindly approved ;)

ericvaandering:
@ticoann please merge this. The pylint errors are coming from things where I just renamed IterTools. Any file I truly modified, I cleaned up

PR #7582, opened 2017-02-03
add heartbeat reporting

ticoann:
tested in my vm.  need to update with https://github.com/dmwm/deployment/pull/437

### DIFF
--- a/src/python/WMCore/Services/LogDB/LogDB.py
+++ b/src/python/WMCore/Services/LogDB/LogDB.py
@@ -8,6 +8,7 @@ https://github.com/dmwm/WMCore/issues/5705
 import re
 import logging
 import threading
+from collections import defaultdict
 
 # project modules
 from WMCore.Services.LogDB.LogDBBackend import LogDBBackend
@@ -58,12 +59,12 @@ class LogDB(object):
         try:
             if request == None:
                 request = self.default_user
-            if  self.user_pat.match(self.identifier):
+            if self.user_pat.match(self.identifier):
                 res = self.backend.user_update(request, msg, mtype)
             else:
                 res = self.backend.agent_update(request, msg, mtype)
         except Exception as exc:
-            self.logger.error("LogDBBackend post API failed, error=%s" % str(exc))
+            self.logger.error("LogDBBackend post API failed, error=%s", str(exc))
             res = 'post-error'
         self.logger.debug("LogDB post request, res=%s", res)
         return res
@@ -87,7 +88,7 @@ class LogDB(object):
                     rec.update({'request':request, 'identifier':identifier, 'thr': thr, 'type':mtype})
                     res.append(rec)
         except Exception as exc:
-            self.logger.error("LogDBBackend get API failed, error=%s" % str(exc))
+            self.logger.error("LogDBBackend get API failed, error=%s", str(exc))
             res = 'get-error'
         self.logger.debug("LogDB get request, res=%s", res)
         return res
@@ -100,7 +101,7 @@ class LogDB(object):
             for row in results['rows']:
                 res.append(row["key"])
         except Exception as exc:
-            self.logger.error("LogDBBackend get_all_requests API failed, error=%s" % str(exc))
+            self.logger.error("LogDBBackend get_all_requests API failed, error=%s", str(exc))
             res = 'get-error'
         self.logger.debug("LogDB get_all_requests request, res=%s", res)
         return res
@@ -116,7 +117,7 @@ class LogDB(object):
                 request = self.default_user
             res = self.backend.delete(request, mtype, this_thread)
         except Exception as exc:
-            self.logger.error("LogDBBackend delete API failed, error=%s" % str(exc))
+            self.logger.error("LogDBBackend delete API failed, error=%s", str(exc))
             res = 'delete-error'
         self.logger.debug("LogDB delete request, res=%s", res)
         return res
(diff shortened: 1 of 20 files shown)


ticoann:
test this please

ticoann:
test this please

ticoann:
@amaltaro, Alan please review. It seems to be working in my vm test.

src/python/WMCore/Services/LogDB/LogDB.py
```diff
@@ -126,5 +127,55 @@ def cleanup(self, thr, backend='local'):
...
+        if self.user_pat.match(self.identifier):
+            self.logger.error("User %s: doesn't allow this function", self.identifier) 
+            return report
+        
+        for row in self.backend.get(self.default_user, None, agent=True).get('rows', []):
+            identifier = row['doc']['identifier']
+            # wmstats thread can run in multiple boxed. 
+            if self.identifier == identifier or identifier.startswith(self.identifier):
+                #this will handle wmstats DataCacheUpdate thread with multiple machine
+                postfix = identifier.replace(self.identifier, "")
```
amaltaro: I think I got lost in this review. First, why checking if identifier startswith ... ? Isn't it supposed to be an unique key (cherrypy thread names, agent names, request names)?
What's the difference between identifier and doc.thr? This should tell me what is postfix for.

src/python/WMCore/Services/LogDB/LogDB.py
```diff
@@ -126,5 +127,55 @@ def cleanup(self, thr, backend='local'):
...
+                msg = row['doc']['messages'][-1]['msg']
+                if (thr in report) and ('ts' in report[thr]) and ts <= report[thr]['ts']:
+                    continue
+                else:
+                    report[thr]['type'] = mtype
+                    report[thr]['msg'] = msg
+                    report[thr]['ts'] = ts
+        return report
+    
+    def _append_down_component_detail(self, report, thr, msg, ts=0, state="error"):
```
amaltaro: Instead using ts=0, how about we make this argument mandatory? Otherwise we have no clue when the thread disappeared, do we?

amaltaro (changes requested):
I'm not sure I followed all these changes (that's why I think we could simplify a few things :)).
Code-wise looks Ok, though I suggest testing it in a clean environment with my suggestions, if you agree.

src/python/Utils/CherryPyPeriodicTask.py
```diff
@@ -26,8 +26,19 @@ def __init__(self, config, enableLogDB=False):
         """ 
         self.logger = getTimeRotatingLogger(config._internal_name, config.log_file)
         self.setConcurrentTasks(config)
+        self.logDB = None
+        
+        if enableLogDB:
```
ticoann: I think that is good idea. I will update that.

src/python/Utils/HeartbeatMonitorBase.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import (division, print_function)
+from Utils.CherryPyPeriodicTask import CherryPyPeriodicTask
+
+class HeartbeatMonitorBase(CherryPyPeriodicTask):
+    
+    def __init__(self, rest, config):
+        super(HeartbeatMonitorBase, self).__init__(config, enableLogDB=True)
+        
+        from WMCore.Services.WMStats.WMStatsWriter import WMStatsWriter
```
ticoann: The reason I put this in init method is I am not sure which application has dependence on Utils package. 
Those applications won't be necessary using this class. Or I could move this package some where else (under Services) but I thought this might be the better place.

src/python/Utils/HeartbeatMonitorBase.py
```diff
@@ -0,0 +1,36 @@
...
+        sets the list of function reference for concurrent tasks
+        """
+        self.concurrentTasks = [{'func': self.reportToWMStats, 'duration': config.heartbeatCheckDuration}]
+        
+    def thread_list(self):
+        # Need to implement in child class ie return ['CleanUpTask', 'LocationUpdateTask', 'ReqMgrInteractionTask', 'MonitoringTask']
+        raise NotImplementedError("Need to return list of Class Names for each cherrypy threads to moniotr")
+        
+    def reportToWMStats(self, config):
+        """
```
ticoann: I will add doctoring. :-)

src/python/Utils/HeartbeatMonitorBase.py
```diff
@@ -0,0 +1,36 @@
...
+  
+        """
+        from WMCore.Services.WMStats.WMStatsWriter import convertToServiceCouchDoc
```
ticoann: the same reason above

src/python/WMCore/GlobalWorkQueue/CherryPyThreads/MonitoringTask.py
```diff
@@ -4,11 +4,12 @@
 from WMCore.WorkQueue.WorkQueue import globalQueue
 from WMCore.Services.WMStats.WMStatsWriter import WMStatsWriter, convertToServiceCouchDoc
 
+GQ_THREADS = ['CleanUpTask', 'LocationUpdateTask', 'ReqMgrInteractionTask', 'MonitoringTask']
```
ticoann: I want to eventually have the same as other application HeartbeatMonitor but this one has additional information attached so GQ has a little different structure now.
The reason I explicitly  has the thread list instead of getting from config is that there might be the case we don't want to monitor the thread. (i.e. Heartbeat thread itself or thread running in multiple machines) but I think it wouldn't hurt to monitor everything.  I think it is good idea to get them from config. I will change that.

src/python/WMCore/Services/LogDB/LogDB.py
ticoann: This is the special case I had to handle. In wmstats, DataCacheUpdate thread not just run in one threads but all the threads reqmon was deployed. It might be just one reqmon instance now but we can't assume that future either.
In that case I added the postfix with host name to distinguish different thread in different box. (I still need to improve the detection if one thread is not running in one machine though - it won't detect if it never started.)

src/python/WMCore/Services/LogDB/LogDB.py
ticoann: If thread crashes we always have the record. if thread never started we won't know. I don't think it makes sense to add time (current time) if the thread never started.

ticoann:
@amaltaro, Hi Alan I updated the patch as you suggested. I tested in vm it seems working fine.
I also moved the package location for CherryPyPeriodicTask.py HeartbeatMonitorBase.py from Utils to WMCore.REST since it is belong to cherrypy server side code. But I am not sure whether that is the right place. If you think there is a better location for that package please let me know.
I still put the LogDB import with in the methods of HeartbeatMonitorBase class, since I am not sure whether some application imports REST packages but not Services packages.

amaltaro:
@ticoann I haven't seen any problem with the agent. I think we can merge it before cutting the WMAgent tag.

PR #7586, opened 2017-02-03
Make condor submitters able to handle a list of releases or a single release

ericvaandering:
I think this should work for the case we want using stringMemberOf as we do for sites.  More work needed once we branch out past x86 or have multi-os capable pilots.

Also removes the original condor plugin as agreed on Thursday with @ticoann and @amaltaro

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -3,28 +3,26 @@
 _SimpleCondorPlugin_
 
 """
-from __future__ import print_function
-from __future__ import division
+from __future__ import division, print_function
 
+import logging
 import os
 import os.path
 import re
-import time
-import logging
 import threading
+import time
+
+import classad
+import htcondor
 
 import WMCore.Algorithms.BasicAlgos as BasicAlgos
-
-from WMCore.DAOFactory import DAOFactory
-from WMCore.WMInit import getWMBASE
+from Utils.IterTools import convertFromUnicodeToStr, grouper
 from WMCore.BossAir.Plugins.BasePlugin import BasePlugin
-from WMCore.FwkJobReport.Report import Report
 from WMCore.Credential.Proxy import Proxy
-from Utils.IterTools import grouper, convertFromUnicodeToStr
+from WMCore.DAOFactory import DAOFactory
+from WMCore.FwkJobReport.Report import Report
+from WMCore.WMInit import getWMBASE
 
-##  python-condor stuff
-import htcondor
-import classad
 
 class SimpleCondorPlugin(BasePlugin):
     """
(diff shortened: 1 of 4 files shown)


hufnagel (commented):
For comparisons in class ads please use =?= instead of ==. I don't quite remember why, but I did change that after a chat with Brian, there was a reason I picked =?=.

ericvaandering:
@hufnagel are you requesting I change the old plugin as well for the lines I changed or just your shiny new plugin?

ericvaandering:
Removing the old plugin is not quite that simple, dropped that commit.

ericvaandering:
test this please

hufnagel:
I don't care about the old plugin :-).

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -103,7 +104,9 @@ def __init__(self, config):
         self.acctGroupUser = getattr(config.BossAir, 'acctGroupUser', "cmsdataops")
 
         # Build a requirement string
-        self.reqStr = "stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) && ((REQUIRED_OS=?=\"any\") || (GLIDEIN_REQUIRED_OS=?=REQUIRED_OS)) && (TARGET.Cpus >= RequestCpus)"
+        self.reqStr = ('stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) '
```
amaltaro: Nice truncation!

amaltaro: Nice truncation!

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -585,11 +588,24 @@ def getProcAds(self, jobList):
             ad['PostJobPrio2'] = int(-1 * job['taskID'])
 
             # Add OS requirements for jobs
-            if job.get('scramArch') is not None and job.get('scramArch').startswith("slc6_"):
-                ad['REQUIRED_OS'] = "rhel6"
+
+            requiredOSes = set()
```
amaltaro: How about making it a get function in BasePlugin?

amaltaro (approved):
I have only one suggestion - but it's really optional - the code looks good. However, before we merge it, let me test these changes this week

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ericvaandering: Yeah, I figured there had to be a pythonic way of doing it without continuation marks since there is an effort to reduce the number of places they are needed. Just have to be careful with spaces since they must be put in manually.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ericvaandering: Will do. Good idea and I can test it.

ericvaandering:
This has addressed all the comments, I believe.

amaltaro:
Thanks, Eric. Testing it in testbed now. BTW, I think this one will need to be merged before #7579, otherwise the IterTools change will conflict

ericvaandering:
@amaltaro yeah, probably. There may be an easily resolved conflict anyhow.

amaltaro:
I didn't test the list feature, but at least all the current functionality is working Ok.
@ticoann also good to be merged

PR #7587, opened 2017-02-03
don't call getElementsForParent if not necessary

ticoann:
as it explained in the title. reduce the unnecessary call when following gets called.
 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueue.py#L660


amaltaro (approved):
Nice catch! And this continuous splitting is only triggered for requests that are in running-open, right?

ticoann:
Yes

PR #7590, opened 2017-02-06
Remove user/group from ACDC collections

### LINKED ISSUE #7572: Remove user/group from ACDC collections
During testbed validation, we noticed some ACDCs were failing (failed status) because there was no work for collection. 

It turns out the original request didn't have a Group key/value, job failures (acdc docs) were somehow injected with DATAOPS group and then the ACDC was failing to fetch those docs, since the ACDC also didn't have a Group (actually its value was "unknown"). This is the query:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ACDC/DataCollectionService.py#L121

So the idea is to get rid of of at least user and group from this query. Only the filesetName should suffice (since it contains the collection name and task), but we have to investigate whether we use the collection name somewhere else in the code.

amaltaro:
Fixes #7572 
Initial tests worked without any issues on services and agent side.
Now let's see how many unittests it's going to break, once that's fixed, I'll ask for your review @ticoann


amaltaro:
test this please

amaltaro:
Now that we created a wmagent branch, feel free to review and merge.
Seangchan, if you prefer we can only update the views and leave all the owner/group code around, though IMO we better clear that out together.

amaltaro:
@ticoann please review or be silent forever :-)

src/python/WMCore/ACDC/CouchService.py
```diff
@@ -8,92 +8,19 @@
...
-from WMCore.GroupUser.User import User, makeUser
 import WMCore.Database.CouchUtils as CouchUtils
-import WMCore.Database.CMSCouch as CMSCouch
 
-class CouchService(Service):
+class CouchService(object):
 
     def __init__(self, **options):
-        Service.__init__(self, **options)
+        super(CouchService, self).__init__()
```
ticoann: Is this line needed?

ticoann (approved):

ticoann:
I am silent forever.:-) ..... 

Well... One thing, patch looks good but I think we need some careful testing. I will merge this.

amaltaro:
Very very careful :-) Thanks!

PR #7592, opened 2017-02-06
work around the fact that dCache only knows ADLER32, not adler32

hufnagel:
FNAL dCache only seems to know about ADLER32 checksums, the xrootd client only knows about adler32, neither accepts the other capitalization. The actual xrdcp stageout command still works (it prints a warning that it doesn't know what to do with the passed adler32 checksum), but the post-transfer checksum comparison fails since the regular expression expects an adler32 return and get ADLER32. Fix the regular expression.

In parallel also investigate why dCache gets it wrong, but this patch at least allows the plugin to work against FNAL dCache.

### DIFF
--- a/src/python/WMCore/Storage/Backends/XRDCPImpl.py
+++ b/src/python/WMCore/Storage/Backends/XRDCPImpl.py
@@ -140,7 +140,7 @@ class XRDCPImpl(StageOutImpl):
         if useChecksum:
 
             copyCommand += "echo \"Local File Checksum is: %s\"\n" % checksums['adler32']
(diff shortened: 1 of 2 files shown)


hufnagel:
@ticoann , since Ajits test worked, here is the pull request. Should probably go into a tag soon and would also be good to be back-ported to the current production agents and deployed there. I would like to use this plugin for FNAL stageout soon...

hufnagel:
test this please (don't believe the failure)

hufnagel:
Ok, fixed the unit test too so it shouldn't complain anymore (well, the first test will, the one after not).

ticoann:
Thanks Dirk, I will merge it when Jenkins finish the test.

PR #7593, opened 2017-02-07
Never return None from MemoryCacheStruct

amaltaro:
We can either change the default value (as done here) or have a check for None in all parts of the code that use this cache struct as an iterable. I think it's much cleaner this one, and this cache should always cache some sort of iterable object.

Just to give some background, this caused all the agents to break (JobSubmitter) during the cmsweb upgrade, since it can return None in case of exception not raised:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Cache/GenericDataCache.py#L38

this was the traceback:
```
2017-02-07 11:34:27,369:139733011392256:INFO:Harness:>>>Terminating worker threads
2017-02-07 11:34:27,380:139733011392256:ERROR:BaseWorkerThread:Error in event loop (2): <WMComponent.JobSubmitter.JobSubmitterPoller.JobSubmitterPoller instance at 0x7f16292a4cf8> JobSubmitterPollerException
Message: Fatal error in JobSubmitter:
argument of type 'NoneType' is not iterable


        ModuleName : WMComponent.JobSubmitter.JobSubmitterPoller
        MethodName : algorithm
        ClassInstance : None
        FileName : /data/srv/wmagent/v1.0.21.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.0.21.patch1/lib/python2.7/site-packages/WMComponent/JobSubmitter/JobSubmitterPoller.py
        ClassName : None
        LineNumber : 707
        ErrorNr : 0

Traceback: 
  File "/data/srv/wmagent/v1.0.21.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.0.21.patch1/lib/python2.7/site-packages/WMComponent/JobSubmitter/JobSubmitterPoller.py", line 691, in algorithm
    self.refreshCache()

  File "/data/srv/wmagent/v1.0.21.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.0.21.patch1/lib/python2.7/site-packages/WMComponent/JobSubmitter/JobSubmitterPoller.py", line 249, in refreshCache
    if (newJob['request_name'] in abortedAndForceCompleteRequests) and \


Backtrace:
  File "/data/srv/wmagent/v1.0.21.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.0.21.patch1/lib/python2.7/site-packages/WMCore/WorkerThreads/BaseWorkerThread.py", line 206, in __call__
    raise ex

2017-02-07 11:34:27,380:139733011392256:INFO:BaseWorkerThread:Worker thread <WMComponent.JobSubmitter.JobSubmitterPoller.JobSubmitterPoller instance at 0x7f16292a4cf8> terminated
```

### DIFF
--- a/src/python/WMCore/Cache/GenericDataCache.py
+++ b/src/python/WMCore/Cache/GenericDataCache.py
@@ -3,15 +3,17 @@ import time
 import traceback
 import logging
 
+
 class MemoryCacheStruct(object):
     """
     WARNING: This can be used in multi thread by registering on GenericDataCache
     But this cache is not thread saft.
     """
+
     def __init__(self, expire, func, initCacheValue=None, kwargs=None):
         """
         expire is the seconds which cache will be refreshed when cache is older than the expire.
-        func is the fuction which cache data is retrived
+        func is the fuction which cache data is retrieved
         kwargs are func arguments for cache data
         """
         self.data = initCacheValue
@@ -20,28 +22,28 @@ class MemoryCacheStruct(object):
         if kwargs == None:
             kwargs = {}
         self.kwargs = kwargs
-        self.lastUpdated  = -1
-        
+        self.lastUpdated = -1
+
     def isDataExpired(self):
         if self.lastUpdated == -1:
             return True
         if (int(time.time()) - self.lastUpdated) > self.expire:
             return True
         return False
-    
+
     def getData(self, noFail=True):
         if self.isDataExpired():
             try:
                 self.data = self.func(**self.kwargs)
                 self.lastUpdate = int(time.time())
-            except Exception as ex:
+            except Exception:
                 if noFail:
                     logging.error(traceback.format_exc())
                 else:
                     raise
         return self.data
 
-        
+
 class CacheExistException(Exception):
     def __init__(self, cacheName):
         Exception.__init__(self, cacheName)
@@ -51,6 +53,7 @@ class CacheExistException(Exception):
     def __str__(self):
         return "%s: %s" % (self.msg, self.error)
 
+
 class CacheWithWrongStructException(Exception):
     def __init__(self, cacheName):
         Exception.__init__(self, cacheName)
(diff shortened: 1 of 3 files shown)


ticoann:
The reason why default value is None is that it doesn't enforce the data type it stores, It could be dictionary, list, string, object or anything. However practically it would be either list or dict.
It might be better to caller to define default value and pass it to the function.

amaltaro:
The test that is failing is DQMHarvest (I'll fix it tomorrow). @ticoann can you check whether this is what you suggested? Or should we define the default value from the very top (from each of the components using this cache)?

src/python/WMCore/Cache/GenericDataCache.py
```diff
@@ -3,45 +3,47 @@
...
+    def __init__(self, expire, func, initCacheValue=None, kwargs=None):
         """
-        expire is the seconds which cache will be refreshed when cache is order than the expire.
-        func is the fuction which cache data is retrived
+        expire is the seconds which cache will be refreshed when cache is older than the expire.
+        func is the fuction which cache data is retrieved
         kwargs are func arguments for cache data
         """
-        self.data = None
+        self.data = initCacheValue
```
ticoann: It think we can set the default value here other wise make initCacheValue mandatory

```
self.data = [] if initCacheValue == None else initCacheValue
```

amaltaro: I'm not sure I followed your comment. We do set the default value according to `initCacheValue` and if not provided, it defaults to the old value `None`, thus just keeping the previous behaviour.

ticoann: I thought you wanted to have default value [] not None. Since old default mostly breaks when it is not overwritten, but if you are OK with None it would be fine.

ticoann:
test this please

src/python/WMCore/Cache/GenericDataCache.py
amaltaro: Yep, that's what I wanted and that's why I explicitly set it here:
https://github.com/dmwm/WMCore/pull/7593/files#diff-8468e9d008c1cdfe50bf6725d0eac87bR215

ticoann:
test this please

ticoann:
@ericvaandering, Eric it seems the Jenkins test got stuck on this one.

PR #7597, opened 2017-02-08
add logger for StatusChangeTasks

ticoann:


amaltaro (approved):
It looks good. Since you're touching this code, can you also fix that logging line (L61) that would print "Total aborted-completed-archived: X."? I noticed this weird "status" in the logs yesterday

ticoann:
Alan, I was thinking it might be the too many log. Should I change to debug level? Or stay in info?

PR #7603, opened 2017-02-08
Add nStreams to CMSSW template and manipulator

ericvaandering:


src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -584,14 +584,20 @@ def handleSingleCoreOverride(self):
...
+        Make sure job only uses one core and one stream in CMSSW
         """
         try:
             if int(self.step.data.application.multicore.numberOfCores) > 1:
                 self.step.data.application.multicore.numberOfCores = 1
         except AttributeError:
             pass
 
+        try:
+            if int(self.step.data.application.multicore.eventStreams) > 1:
```
amaltaro: Shouldn't it be `> 0` instead?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -680,15 +686,17 @@ def __call__(self):
 
         try:
             origCores = int(getattr(self.step.data.application.multicore, 'numberOfCores', 1))
+            eventStreams = int(getattr(self.step.data.application.multicore, 'eventStreams', 0))
             resources = {'cores': origCores}
-            resizeResources(resources)
+            if not eventStreams:  # Don't know how to handle resizing jobs with different nThreads and nStreams
```
amaltaro: Giving this a second thought. I think it's more dangerous not to resize than perform a non-ideal resize. Consider the following: a job originally requesting 8 cores (but resizable from 4 to 8) matches against a 4 cores slot. If we don't resize it, this job would spawn the original 8 cores and then the worker node is in trouble.

For the moment, how about we set it to 0 in case `origCores != numCores`?

amaltaro (changes requested):
Only those 2 pending comments, the rest looks good.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
ericvaandering: Either is going to be pathological. We have to make sure it never happens. If we fire up too many cores, yeah, we do something bad and we run out of memory. If we scale down to four course, it probably still runs out of memory since we were setting nStreams < nCores for a reason.

So maybe your suggestion is "less bad" but like I said, we have to make sure this does not happen.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
ericvaandering: Yeah, probably.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
amaltaro: That's the problem, we cannot control it :( Anyone can edit condor jobs and set resisable to true.

ericvaandering: OK, you are bringing me around. I will change the logic. I'll also put something in the log file.

ericvaandering:
Addressed @amaltaro comments. Should have two Jenkins tests outstanding, 2nd one is on the reordered, squashed commits.

amaltaro (approved):

amaltaro:
I have almost all changes from the first commit into the other PR now (7634), which seems to be functional.

How about we merge this one, then I rebase 7634 and start the real tests?
Eric, do you know after which CMSSW release EventStreams is supported?

ericvaandering:
Sounds good that setting is been there had a number of threads to set so anything we are running now in multithreaded mode.

On Feb 17, 2017, at 07:30, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


I have almost all changes from the first commit into the other PR now (7634), which seems to be functional.

How about we merge this one, then I rebase 7634 and start the real tests?
Eric, do you know after which CMSSW release EventStreams is supported?

ericvaandering:
Wow, Siri and a loud Starbucks garbled that message. The setting for nStreams has been in CMSSW as long as the multithreaded stuff has been there, so you can test this with anything we are currently running in multithreaded mod.

amaltaro:
@ticoann can you merge it, please? I'll then let some workflows running for the weekend.

amaltaro:
Thanks, you're the man!

ticoann:
That was easy to make you happy. 😄

PR #7604, opened 2017-02-09
Use env var when getting CMSSW version in CRAB

emaszs:
When running TweakPset.py in CRAB, `self.step.data.application` basically doesn't contain a setup attribute and causes an exception. It used to be the case that the CMSSW version was being retrieved from the env until this commit: https://github.com/dmwm/WMCore/commit/81b9910c4e22e9d79d3cd6dfa2d19c189d5deddc. 
There are also a couple other places in SetupCMSSWPSet.py where a similar change was made, here: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L551 and here: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L605, but I don't think these are used by CRAB, please let me know if anyone suspects otherwise or if I should modify those places as well for the sake of consistency.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -548,7 +548,11 @@ class SetupCMSSWPset(ScriptInterface):
         runIsComplete = getattr(baggage, "runIsComplete", False)
         multiRun = getattr(baggage, "multiRun", False)
         runLimits = getattr(baggage, "runLimits", "")
-        cmsswVersion = self.step.data.application.setup.cmsswVersion
+        try:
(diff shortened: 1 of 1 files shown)


amaltaro:
Thanks, Emilis! I think we could make it consistent and apply the same changes to the other 2 places.
Once that's done, can you test these changes before we merge it?

emaszs:
I've made the change consistent across the file. I also see test jobs finishing successfully with the patch, hopefully there won't be many more changes needed for CRAB :)

ticoann:
@emaszs Hi Emilis, if you need this patch with new tag for crab. Please make a pull request against 1.1.0_crab branch as well. Thanks.

PR #7605, opened 2017-02-09
Fix for agents by team call

amaltaro:
global_workqueue doc broke this code.


amaltaro:
In theory we should also backport it to 1.0.23, in practice I think we can just leave that as it is (using the default value).

ticoann:
Yes we don't have to back port. Let's just add on the new agent.

ticoann (approved):

ticoann:
@amaltaro, Alan this was added, right?

amaltaro:
Yes, it's already in the wmagent branch (#7616) and applied to all production agents.

Issue #7580, opened 2017-02-02
docker set up for unittest


ticoann:
When I set up the unittest using docker form the github wiki page and run the unittest.
I am getting a following error. Although it seems it is picking up correct python version 2.7.11

  File "/home/dmwm/unittestdeploy/wmagent/1.0.21.patch4/sw/slc6_amd64_gcc493/cms/wmagent/1.0.21.patch4/lib/python2.7/site-packages/Utils/itertools.py", line 5, in <module>
    from itertools import islice
ImportError: cannot import name islice

@ericvaandering, Eric, do you know what I am doing wrong?

ericvaandering:
This was fixed by #7579

PR #7606, opened 2017-02-09
Change the UUID module name

ticoann:
Due to the same reason as #7580. Also it is better not to use the built in module name although case is different.


ticoann:
There are a lot more places using this module than I thought. I am not sure whether there is better approach. 
If this patch is approved. We also need to change

https://github.com/dmwm/T0/search?utf8=%E2%9C%93&q=UUID

and following patch might need to be recreated.
https://github.com/cms-sw/cmsdist/blob/comp_gcc493/crabserver3-setup.patch

But if there is better option please advise.

ericvaandering (approved):
Looks good. Now you just have to address all the pylint comments. :-) Just kidding.

ticoann:
> Now you just have to address all the pylint comments

My heart stopped for a while.

amaltaro (approved):
Patch looks good.

Issue #7445, opened 2016-12-05
Request not going to completed


vlimant:
I came across https://cmsweb.cern.ch/couchdb/workqueue/_design/WorkQueue/_rewrite/elementsInfo?request=pdmvserv_B2G-RunIISummer16DR80Premix-00791_00064_v0__161109_162829_9818 for example that has all WQE Done, but is not completed.
At least not from where I look
FYI @prozober

amaltaro:
I have the impression this is related to the reqmgr (one) problem.

The next cmsweb upgrade (tomorrow) should have a fix for this issue. If you cannot wait, then I'll talk to Seangchan, I think he came up with a script to get these workflows moving on.

vlimant:
if that can be fixed tomorrow, it is fine.

ticoann:
Actually, it seems it is different problem we had with reqmgr1. 
The problem was wq_inbox element (request) is deleted before it changed the request to complete status. Not completely sure how this happened, 
But it might be the case workflow is all finished (reported from wmagent by reqplication between following calls)
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py#L60
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py#L66

That could happen. 
Anyway, I move the status manually and it is in completed status.
For the fix, instead of fixing this specific case, we are trying to change how the request status status is updated for complete. -archived status.

jenimal:
I found another 115 workflows from the EXO-RunIIWinter15pLHE camplaign that are stuck in running-closed:
https://cms-logbook.cern.ch/elog/Workflow+processing/25998

is this the same problem?  should I look for more?

vlimant:
@amaltaro @ticoann that's plenty of meat for debugging.

ticoann:
I didn't look at all the 115 workflows but it seems to be the same problem as above for the ones I looked at. I will investigate more.

vlimant:
Can you guys please do a magic to have them move along asap ?

ticoann:
Jean-Roch, I am testing the script to spot this and move the status. It should be done later today.

ticoann:
Total 122 requests are moved to completed. It seems the same problem as previous one. And I am pretty sure the cause is the race condition I mentioned above. We will create the patch for next deployment.

vlimant:
[...]

PR #7612, opened 2017-02-10
Emit workqueue elements with 0 Jobs

amaltaro:
As reported in the last comment of #7445
I was surprised I could not find any of those 2 requests in this view.

This problem is hitting us right now in production. We have to discuss with cmsweb team to see how we handle this case.

### DIFF
--- a/src/couchapps/WorkQueue/views/analyticsData/map.js
+++ b/src/couchapps/WorkQueue/views/analyticsData/map.js
@@ -1,11 +1,14 @@
 function(doc) {
    var ele = doc["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
-   if (ele && ele['Jobs']) {
+   if (ele && ele['Jobs'] >= 0) {
+       var jobs = ele['Jobs']
+       if (ele["Jobs"] === null)
+           jobs = 0
        // assumes request name and input data is one to one (None for MC workflow)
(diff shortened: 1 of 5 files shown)


amaltaro:
And the second commit addresses other WorkQueue views that could be affected by the Jobs value (besides a Available typo).

amaltaro:
test this please

ticoann:
Alan, except jobInjectStatusByRequest/map.js other code was not modified for this deployment, right? I was wondering how Jobs == 0 is legit case in first place.  Which splitting algo allows Jobs == 0?

ticoann:
Actually there is one missing view which cause this problem.  jobStatusByRequest. 
I will create the new branch patch on top of this

ticoann:
Never mind of above comments. You fixed all. Sorry.

ticoann:
I think this doesn't need to be merged. #7618 should guarantee valid WQE has to have jobs more than 1. If we get Jobs == 0 that would be error case and we should investigate.
We can check before the next deployment whether there is remaining WQE with Jobs == 0.

ticoann (commented):
This might not be needed anymore.

amaltaro:
But 7618 will go only in the next production upgrade and it may be that there are 0 jobs wqe in production already, though we can manually fix those. In the end I agree with in not adding code that should be cleaned in a month or two.

I'll make another PR to fix only the analytics view. I'm closing this one then.

PR #7617, opened 2017-02-11
change debug level to info for the status report

ticoann:
@amaltaro, I think this can be info level. There is no report for this function.


ticoann:
test this please

amaltaro:
And it should log only 3 lines per cycle, I agree it's a good idea.

ticoann:
It could be more than 3 lines since it logs every status if there is status transition. But it should be less then 10 lines per cycle.

PR #7618, opened 2017-02-12
allow 0 event file for job estimation

ticoann:
replace #7612.  Instead allowing  Jobs == 0 in workqueue elements, set the job 1 when number of events are 0 due to the file has 0 event.
Since there is legit case it shouldn't allow Job == 0 (in MC case)
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py#L113

@amaltaro, Alan please review

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/Block.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/Block.py
@@ -3,7 +3,7 @@
 WorkQueue splitting by block
 
 """
-__all__ = []
+from __future__ import print_function, division
 
 from math import ceil
 from WMCore.WorkQueue.Policy.Start.StartPolicyInterface import StartPolicyInterface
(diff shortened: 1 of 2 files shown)


src/python/WMCore/WorkQueue/Policy/Start/Block.py
```diff
@@ -45,15 +45,22 @@ def split(self):
...
                         parentList[dbsBlock["Name"]] = self.siteDB.PNNstoPSNs(dbsBlock['PhEDExNodeList'])
-
+            
+            if self.args['SliceType'] == 'NumberOfEvents' and block['NumberOfEvents'] == 0:
+                # if there is 0 events file and use EventAwareLumiBased algo
+                # set the estimate jobs to 1 (It is hard to get the realistic estimate 
+                # unless run the splitting alogrithm
+                estimateJobs = 1
+            else:
+                estimateJobs = ceil(float(block[self.args['SliceType']]) / float(self.args['SliceSize'])),
```
amaltaro: I'd suggest to just use this formula above in a `max(1, formula)`. For clarity, a line of comment saying that there may be 0 event files.

amaltaro (changes requested):
BTW, is it really replacing #7612? If it does, then we have to make sure there are no elements with Jobs=0 in the system when it goes into production.

src/python/WMCore/WorkQueue/Policy/Start/Block.py
ticoann: Do you mean replace whole thing with max or just NumberOfEvents case.
I think only legit case for the formula to be 0 is when NumberOfEvents is 0. If other value (.i.e NumberOfFiles or NumberOfLumis), make this formula 0 there is some thing wrong with it, isn't it?

ericvaandering: Are these args already integers or are you converting from string? A quick search shows that you could get rid of all kinds of casting to floats simply by importing division from the future.

amaltaro: I'd replace the whole thing. NumberOfFiles=0 is already taken care and the block is rejected if that happens.
For NumberOfLumis - Eric can correct me - but I don't think we can have a EDM file without a lumi section.

ericvaandering: I think you are right

ticoann:
Alan, Eric, I updated the patch as you suggested. (I don't think string value is passed, I changed to  __future__ division)

ericvaandering:
Would you mind quickly updating lines 135 and 183 as well?

ticoann:
Thanks Eric, I didn't check other places. Patch is updated.

ticoann:
test this please

ticoann:
test this please

amaltaro:
test this please

amaltaro:
test this please

ticoann:
There are 3 failing unittests.

WMCore_t.Agent_t.HarnessCouchBackend_t.HarnessCouchBackend_t:testAddWorker changed from success to failure
WMCore_t.Services_t.WorkQueue_t.WorkQueue_t.WorkQueueTest:testCompletedWorkflow changed from success to failure
WMCore_t.Services_t.WorkQueue_t.WorkQueue_t.WorkQueueTest:testWorkQueueService changed from success to failure

First one seems to be temporary error. The other 2 is due to the changes (previous Jobs count was 0 but new patch change those to 1) - updated the patch now. Hopefully it will pass the unittest now.

amaltaro (approved):

PR #7616, opened 2017-02-11
#7605 #7620 for  1.1.0 wmagent 

ticoann:
@amaltaro, Alan this need to be included in wmagent branch for the deployment, right?

PR #7620, opened 2017-02-14
Use machine attr for CMS site location in SimpleCondor

amaltaro:
Problem spotted and reported in:
https://cms-logbook.cern.ch/elog/GlideInWMS/5034

Summary, the `JOBGLIDEIN_CMSSite` expression we set on the submitter plugin is not correct (I don't know why the others are though...) and it causes `MATCH_EXP_JOBGLIDEIN_CMSSite` not to be set at all.

One of the solutions is to use the machine attribute - which is supposed to be more reliable.
For the record, the `0` in the end corresponds to the actual site. E.g., if job gets evicted, then it could have a `1` in the end with the previous site name.

I still have to test it, but comments are welcome (especially from @bbockelm )

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -183,11 +183,11 @@ class SimpleCondorPlugin(BasePlugin):
         logging.debug("Start: Retrieving classAds using Condor Python XQuery")
         try:
             itobj = schedd.xquery("WMAgent_AgentName == %s" % classad.quote(self.agent),
-                                  ['ClusterId', 'ProcId', 'JobStatus', 'MATCH_EXP_JOBGLIDEIN_CMSSite'])
+                                  ['ClusterId', 'ProcId', 'JobStatus', 'MachineAttrGLIDEIN_CMSSite0'])
             for jobAd in itobj:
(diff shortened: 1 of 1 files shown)


amaltaro:
This patch was applied to vocms0230 and I can properly see `MachineAttrGLIDEIN_CMSSite0` in the job classAds.
I can also see a proper location in wmbs_job for the job I'm looking at.

So, from my side it's good to go to production right away. Just waiting for unittests and a confirmation from @ticoann @hufnagel or @bbockelm

ticoann:
Alan, I have no idea on this. So if you get @bbockelm Brian's blessing, it is good to patch.  Thanks

amaltaro:
@Farrukh-Aftab since I was bugging you on gtalk, your review would be appreciated as well :)

amaltaro:
This was pushed to the 3 agents running the new release. If we get no further comments by tomorrow, then we can backport it to the wmagent branch on https://github.com/dmwm/WMCore/pull/7616

Farrukh-Aftab:
Looks okay to me. You can use the same syntax to fetch more information from the startd side too if needed.

amaltaro:
Thanks, Farrukh.
Seangchan, I guess we can merge it now.

hufnagel:
Not sure why this didn't work anymore, it did at some point for me. But if the new attribute is more reliable then switching to it is the better approach anyways, so don't think it's worth debugging this any further.

ticoann (approved):

PR #7621, opened 2017-02-14
fixes #7285 ACDC Memory value inheritance

### LINKED ISSUE #7285: Incorrect Memory value inherited for ACDC (TaskChain)
It requires further investigation, however running a ACDC for
amaltaro_TaskChain_RelVal_Multicore_Agent1021_Validation_161007_234412_2447

in testbed, for the second task (RECODreHLT). I see it has memoryRequirements 4000 instead of 8000. 

It's likely inheriting the value from the main dict instead of the task one...

ticoann:
Fixes #7285,
@amaltaro, Alan please review. I also added fix for PrepID, AcquisitionEra, ProcessString.
If there is more task specific values let me know.

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/DataStruct/T1/WMStats.RequestSummary.js
+++ b/src/couchapps/WMStats/_attachments/js/DataStruct/T1/WMStats.RequestSummary.js
@@ -137,3 +137,24 @@ WMStats.Requests = function(data) {
     };
     return tier1Requests;
 };
+
+
+WMStats.Requests.getPropertyByTask = function(taskPath, property, requestInfo) {
+	
+    if (requestInfo.TaskChain) {
+        var numTasks = requestInfo.TaskChain;
+        var taskSplit = taskPath.split("/");
+        var task = taskSplit[taskSplit.length -1];
+        for (var i=1; i <= numTasks; i++) {
+            if (requestInfo["Task" + i].TaskName == task) {
+                if (requestInfo["Task" + i][property] !== undefined) {
+                	return requestInfo["Task" + i][property];
+                } else {
+                	break;
+                }
+            }
+        }
+    }
+    
+    return requestInfo[property];
+};
--- a/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
+++ b/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
@@ -421,35 +421,36 @@ WMStats.ViewModel = (function (){
             }
             
             var requestInfo = reqMgrRequest.getData();
+            var getPropertyByTask = WMStats.Requests.getPropertyByTask;
             //from passing request info
-            summary.Memory = requestInfo.Memory;
-            summary.SizePerEvent = requestInfo.SizePerEvent;
-            summary.TimePerEvent = requestInfo.TimePerEvent;
             summary.DbsUrl = requestInfo.DbsUrl || "https://cmsweb.cern.ch/dbs/prod/global/DBSReader";
             summary.Group = requestInfo.Group;
             summary.RequestPriority = requestInfo.RequestPriority;
             summary.RequestString = WMStats.Utils.acdcRequestSting(summary.OriginalRequestName, requestInfo.Requestor);
-            summary.PrepID = requestInfo.PrepID;
-            summary.SoftwareVersions = requestInfo.SoftwareVersions;
-            summary.CMSSWVersion = requestInfo.SoftwareVersions[0];
-            summary.Campaign = requestInfo.Campaign;
+            summary.Dashboard = requestInfo.Dashboard;
+            summary.MergedLFNBase = requestInfo.MergedLFNBase;
             // assign more value for acdc assignment
             summary.Team = requestInfo.Teams[0];
+            
+            summary.Memory = getPropertyByTask(summary.InitialTaskPath, "Memory", requestInfo);
+            summary.SizePerEvent = getPropertyByTask(summary.InitialTaskPath, "SizePerEvent", requestInfo);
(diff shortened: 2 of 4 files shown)


ticoann:
This is only javascript change shouldn't be failing any unittest.

ticoann:
test this please

ticoann:
@amaltaro, Alan this seems to be working. But one thing I found is that there are task names like these.

LogCollectForMINIAODDreHLT,  LogCollectForMINIAODDreHLT

We are only filtering LogCollect jobs by checking whether LogCollect is at the end. i.e. HLTDRAWoutputMergeLogCollect

The filtering prevents creating the ACDC when it fails, should I extend that to any tasks containing LogCollect?

src/couchapps/WMStats/_attachments/js/DataStruct/T1/WMStats.RequestSummary.js
```diff
@@ -137,3 +137,24 @@ WMStats.Requests = function(data) {
...
+
+WMStats.Requests.getPropertyByTask = function(taskPath, property, requestInfo) {
+	
+    if (requestInfo.TaskChain) {
+        var numTasks = requestInfo.TaskChain;
+        var taskSplit = taskPath.split("/");
+        var task = taskSplit[taskSplit.length -1];
+        for (var i=1; i <= numTasks; i++) {
+            if (requestInfo["Task" + i].TaskName == task) {
+                if (requestInfo["Task" + i][property] !== undefined) {
```
amaltaro: I'm not sure if we should also check for null value here. It looks like the answer is no, since most of the properties we'd look for don't have a default value inside the tasks:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py#L554

src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
```diff
@@ -421,15 +421,16 @@ WMStats.ViewModel = (function (){
             }
             
             var requestInfo = reqMgrRequest.getData();
+            var getProperyByTask = WMStats.Requests.getPropertyByTask;
```
amaltaro: Haaaa, if there is no typo, it's not your PR... but I give you this, it's consistent at least :-)
Can you update it in all the places please?

amaltaro (changes requested):
Changes look good in general. However, I wonder why we don't have the same handle for `Multicore` property? Can you add it please?

In addition to that, can you remove the `TrustSitelists` and `TrustPUSitelists` handle please? We know these arguments have to be provided during assignment, otherwise it causes trouble.

Last but not least, you may want to update your first commit message saying that it also fixes #7285 (yes!!! 1 PR, 2 bugs fixed!)

src/couchapps/WMStats/_attachments/js/DataStruct/T1/WMStats.RequestSummary.js
ticoann: Yes, default value inside the task would be not good. That means it will overwrite the workflow level default. I think if something sets Null in task property most likely that is wrong on the config part.  But if some one said intentionally we have respect that as that property is set to Null for some reason.

src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
ticoann: I am always consistence since I copy and paste. :-) I will fix the typo.

ticoann:
> In addition to that, can you remove the TrustSitelists and TrustPUSitelists handle please? We know these arguments have to be provided during assignment, otherwise it causes trouble.

Didn't we decide to have default value for this in ACDC.  Since child test can't be set arbitrary for this?

amaltaro:
Yes, but depending on how it's used, it can have side effects. E.g. of inheriting a True value:
*) acdc created via wmstats and assigned via web --> OK
*) acdc created via wmstats and assigned via script (that doesn't set those parameters to True again) --> BAD

How about we properly discuss it tomorrow?

ticoann:
@amaltaro, Alan I updated what we discussed in the meeting. Could you check one more time?
Thanks,

src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
```diff
@@ -421,35 +421,36 @@ WMStats.ViewModel = (function (){
...
+            summary.MergedLFNBase = requestInfo.MergedLFNBase;
             // assign more value for acdc assignment
             summary.Team = requestInfo.Teams[0];
+            
+            summary.Memory = getPropertyByTask(summary.InitialTaskPath, "Memory", requestInfo);
+            summary.SizePerEvent = getPropertyByTask(summary.InitialTaskPath, "SizePerEvent", requestInfo);
+            summary.TimePerEvent = getPropertyByTask(summary.InitialTaskPath, "TimePerEvent", requestInfo);
+            summary.PrepID = getPropertyByTask(summary.InitialTaskPath, "PrepID", requestInfo);
+            // TODO remove SoftwareVersions and keep only CMSSWVersion
+            summary.SoftwareVersions = requestInfo.SoftwareVersions;
```
amaltaro: Why do we need `SoftwareVersions` here?

amaltaro (commented):
Looks good. Feel free to merge it now :-)
BTW, isn't it possible to get rid of all this dup code?

ticoann:
> BTW, isn't it possible to get rid of all this dup code?

I have to look I think originally we didn't have CMSSWVersion key, so maybe old document might not have that key. Also I don't know whether we use SoftwareVersions to store all the CMSSWVersion(s) since it is the list. Anyway, I will create the issue for this.

PR #7623, opened 2017-02-15
Accept either string or list of strings for ScramArch at spec level

### LINKED ISSUE #7578: Accept string and list of strings for ScramArch
We need to update the request specs in order to accept either a simple string architecture or a list of architectures.

Then we also have to update the plugins such that they can use a different requirements string for that.

amaltaro:
Fixes #7578
Deeper changes at CMSSW level done here https://github.com/dmwm/WMCore/pull/7579

We then validate and convert the user input into a python list of strings (even if user provided only a basic string), otherwise it's a bit more complicated to do the data sanitisation.

Still to be tested.

### DIFF
--- a/src/python/WMCore/WMSpec/WMWorkloadTools.py
+++ b/src/python/WMCore/WMSpec/WMWorkloadTools.py
@@ -153,7 +153,8 @@ def _validateArgFunction(argument, value, valFunction):
     if valFunction:
         try:
             if not valFunction(value):
-                raise WMSpecFactoryException("Argument %s, value: %s doesn't pass the validation function." % (argument, value))
+                raise WMSpecFactoryException(
+                    "Argument %s, value: %s doesn't pass the validation function." % (argument, value))
         except Exception as ex:
             # Some validation functions (e.g. Lexicon) will raise errors instead of returning False
             logging.error(str(ex))
@@ -185,11 +186,10 @@ def _validateArgumentOptions(arguments, argumentDefinition, optionKey=None):
 
 
 def _validateInputDataset(arguments):
-    
     inputdataset = arguments.get("InputDataset", None)
     dbsURL = arguments.get("DbsUrl", None)
     if inputdataset != None and dbsURL != None:
-        #import DBS3Reader here, since Runtime code import this module and worker node doesn't have dbs3 client 
+        # import DBS3Reader here, since Runtime code import this module and worker node doesn't have dbs3 client
         from WMCore.Services.DBS.DBS3Reader import DBS3Reader
         from WMCore.Services.DBS.DBSErrors import DBSReaderError
         try:
@@ -209,7 +209,7 @@ def validateInputDatasSetAndParentFlag(arguments):
         else:
             dbsURL = arguments.get("DbsUrl", None)
             if dbsURL != None:
-                #import DBS3Reader here, since Runtime code import this module and worker node doesn't have dbs3 client 
+                # import DBS3Reader here, since Runtime code import this module and worker node doesn't have dbs3 client
                 from WMCore.Services.DBS.DBS3Reader import DBS3Reader
                 result = DBS3Reader(dbsURL).listDatasetParents(inputdataset)
                 if len(result) == 0:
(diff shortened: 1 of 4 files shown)


ericvaandering:
Looks like this is blowing up in Jenkins...

ericvaandering:
WMCore_t.WMSpec_t.StdSpecs_t.MonteCarloFromGEN_t.MonteCarloFromGENTest.testMonteCarloFromGEN says scramArch is None

amaltaro:
Com'on!! It's supposed to be a simple change :(

amaltaro:
test this please

ericvaandering:
Almost there. :-)

amaltaro:
And it should get fixed with my last commit. I like it, it means we have a quite reasonable unit test coverage now (though there is a lot to improve, of course) :-)

amaltaro:
Most of the StdBase pylint warnings were fixed in 7619. In terms of spec level, it works as expected. I also took a change to look at a couple of condor jdl and a job runtime and everything looks fine (though we have to check it again in the near future).

From my side it's good to go.

src/python/WMCore/WMSpec/WMWorkloadTools.py
```diff
@@ -44,6 +44,20 @@ def makeList(stringList):
     raise WMSpecFactoryException("Can't convert to list %s" % stringList)
 
 
+def makeNonEmptyList(stringList):
```
ericvaandering: I'd rather see this in Utils (perhaps true of everything here) with a generic exception. But I think there is already exactly this in WMObject if you are inheriting from that. 

Having this general purpose stuff scattered around is a PITA.

test/python/WMCore_t/WMSpec_t/StdSpecs_t/StepChain_t.py
```diff
@@ -285,7 +285,7 @@ def testStepChainSingleStep(self):
         self.assertEqual(outModsAndDsets['outputDataset'],
                          '/PrimaryDataset-StepChain/AcquisitionEra_StepChain-ProcessingString_StepChain-v1/MINIAODSIM')
         self.assertEqual(task.getSwVersion(), 'CMSSW_7_1_25_patch2')
-        self.assertEqual(task.getScramArch(), 'slc6_amd64_gcc481')
+        self.assertItemsEqual(task.getScramArch(), ['slc6_amd64_gcc481'])
```
ericvaandering: Nicely done. :-)

test/python/WMCore_t/WMSpec_t/WMWorkloadTools_t.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def testMakeNonEmptyList(self):
+        """
+        Test the makeNonEmptyList function.
+        It has exactly the same behaviour as makeList, but throws an exception
+        for empty list or string.
+        """
+        self.assertRaises(WMSpecFactoryException, makeNonEmptyList, [])
+        self.assertRaises(WMSpecFactoryException, makeNonEmptyList, "")
+
```
ericvaandering: Since it has the same behavior, you should test that's true with a few of your other examples. Test make list against make non empty

ericvaandering (changes requested):

src/python/WMCore/WMSpec/WMWorkloadTools.py
amaltaro: Hmm, I'm not sure we can define (cast) the spec input parameters to a WMObject (if you have an idea, I'm all ears). Perhaps what we could do is to add these functions as WMObject methods and create some sort of wrapper functions in Utils (but then we have the Utils dependency on WMCore/DataStructs).

Or I could simply move these functions to Utils/Utilities or something like that.

test/python/WMCore_t/WMSpec_t/WMWorkloadTools_t.py
amaltaro: will do

src/python/WMCore/WMSpec/WMWorkloadTools.py
ericvaandering: Definitely no dependencies from Utils onto something else. The idea with Utils is that *any* python project might benefit just by grabbing it.

What would it take to just move both make*List functions to Utils?

ticoann:
test this please

amaltaro:
Eric, I changed the unit tests as you suggested. Can you check and approve it if it's Ok?
As we discussed in the meeting, I'll work on this issue https://github.com/dmwm/WMCore/issues/7633 once this one gets merged.

ericvaandering (approved):

amaltaro:
To be on the safe side, triggering jenkins once again.

amaltaro:
test this please

amaltaro:
@ticoann I think we can merge it now

ticoann (approved):

PR #7625, opened 2017-02-15
Make jobs auto-naming

ericvaandering:
Another stand-alone bit of what I need

### DIFF
--- a/src/python/WMCore/JobSplitting/JobFactory.py
+++ b/src/python/WMCore/JobSplitting/JobFactory.py
@@ -122,6 +122,8 @@ class JobFactory(WMObject):
         """
         Instantiate a new Job onject, apply all the generators to it
         """
+        if name is None:
+            name = self.getJobName()
         self.currentJob = self.jobInstance(name, files)
         self.currentJob["task"] = self.subscription.taskName()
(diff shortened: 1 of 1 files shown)


ericvaandering:
test this please

ticoann (approved):

ticoann:
Although we always the pass the job name but this would be safer protection. I will merge once the unittest finished.

ericvaandering:
New code I'm working decides not to name the jobs. :-)

amaltaro:
Sorry for the stupid question :), but what this job naming means? Is it simply the sequential job id assigned to a job (an integer number)? Or a real uid?
What's the impact of having this auto-naming?

ericvaandering:
A lot of times we do this:
  self.newJob(name = self.getJobName())
other times we use makeUUID as that parameter. 

I decided with the former it was best just to make that the default.

ticoann:
This is real uuid  for job. I am not completely sure why we even give the option to overwrite instead of using auto generated job name. Most of the cases does that anyway, except T0.

amaltaro:
Yep, it should have been auto-generated since day 0 then, it makes a lot of sense. Thanks

ericvaandering:
@ticoann please merge

PR #7628, opened 2017-02-16
Fix typo in analyticsData view

amaltaro:
Fix was originally in #7612


ticoann:
Is this view used any where. :-)

amaltaro:
That's a good question :-) I've certainly never used it

PR #7634, opened 2017-02-16
Support for number of event streams

### LINKED ISSUE #7515: Add support to CMSSW number of streams
It should have exactly the same behaviour as Multicore arg, thus:
*) can be set during request creation (defaults to 0)
*) can be overriden during request assignment
*) can have different values for different tasks in the same request

Disclaimer: It would be trick to handle it in resizable jobs, so we decided to set `numberOfStreams` to the same value as `numberOfThreads`, at least to begin with.

@ericvaandering @ticoann please correct/complement if I missed something.

amaltaro:
Fixes #7515

I started creating a unit test for it and noticed some steps have this attribute set to None (I was expecting it to be 0..) so this is a non-working version.

BTW, @ericvaandering I didn't notice you started implementing it at CMSSW level in #7603.
I'll have a look at your PR, and then we can merge that and rebase this one and resume this coding.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -367,10 +367,16 @@ class StdBase(object):
         procTaskCmsswHelper = procTaskCmssw.getTypeHelper()
         procTaskStageHelper = procTaskStageOut.getTypeHelper()
 
-        if 'Multicore' in taskConf and taskConf['Multicore'] > 0:  # used for StepChain overrides
-            procTaskCmsswHelper.setNumberOfCores(taskConf['Multicore'])
-        else:
-            procTaskCmsswHelper.setNumberOfCores(self.multicore)
+        # StepChain overrides
+        multicore = self.multicore
+        eventStreams = self.eventStreams
+        if 'Multicore' in taskConf and taskConf['Multicore'] > 0:
+            multicore = taskConf['Multicore']
+        if 'EventStreams' in taskConf and taskConf['EventStreams'] >= 0:
+            eventStreams = taskConf['EventStreams']
+
+        procTaskCmsswHelper.setNumberOfCores(multicore, eventStreams)
+
         procTaskCmsswHelper.setUserSandbox(userSandbox)
         procTaskCmsswHelper.setUserFiles(userFiles)
         procTaskCmsswHelper.setGlobalTag(globalTag)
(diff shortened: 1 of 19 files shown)


src/python/WMCore/WMSpec/WMStep.py
```diff
@@ -50,12 +50,24 @@ def getNumberOfCores(self):
...
             return 1
 
+    def getNumberOfStreams(self):
+        """
+        _getNumberOfStreams_
+
+        Return the number of event streams for the step in question
+        """
+        # FIXME: I'm not really sure it's actually used
+        print("HEY, GET OUT!!!")
```
amaltaro: Yes, it looks like we use this one in the unit tests. I have to remove this comment/print

amaltaro:
My tests haven't finished yet because there seems to be a problem with the ITB pool. Nonetheless, the specs look fine and it's ready for a review (I'm updating two templates in the other PR).

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -1013,6 +1019,7 @@ def getWorkloadArguments():
                      "IncludeParents": {"default": False, "type": strToBool},
                      "Multicore": {"default": 1, "type": int,
                                    "validate": lambda x: x > 0},
+                     "EventStreams": {"type": int, "validate": lambda x: x >= 0, "null": True},
```
ticoann: Shouldn't this have default value 0 and not allow null?

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -367,10 +367,16 @@ def setupProcessingTask(self, procTask, taskType, inputDataset=None, inputStep=N
...
-            procTaskCmsswHelper.setNumberOfCores(taskConf['Multicore'])
-        else:
-            procTaskCmsswHelper.setNumberOfCores(self.multicore)
+        # StepChain overrides
+        multicore = self.multicore
+        eventStreams = self.eventStreams
+        if 'Multicore' in taskConf and taskConf['Multicore'] > 0:
+            multicore = taskConf['Multicore']
+        if 'EventStreams' in taskConf and taskConf['EventStreams'] >= 0:
+            eventStreams = taskConf['EventStreams']
```
ticoann: If taskConf can over write that, (i.e, different task can have different EventStreams value,

Shouldn't following lines modify to add that?
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StepChain.py#L209
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StepChain.py#L435

How about the TaskChain? Doesn't it have to use this value as well?
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py#L644

ticoann: Sorry never mind at the coments. I was looking at the wrong code. It seems you merged already Eric's change here.

amaltaro: That's right. This PR was rebased on 7603

Issue #7627, opened 2017-02-16
Heartbeat monitoring loses track of threads


amaltaro:
I've seen it in testbed and I've just seen it again in production. It's like the heartbeat monitoring loses track of some threads as if they never ran. For instance, global_workqueue was completely green 15min ago, but now it shows the following red warning:
```
    agent: global_workqueue
    agent last updated: 2017/2/16 (Thu) 08:23:40 UTC : 0 h 3 m
    data last updated: N/A
    status: Components or Thread down
    team:
    component errors:
    HeartbeatMonitor
        HeartbeatMonitor
        status: error
        error: 1970/1/1 (Thu) 00:00:00 UTC
        error message:

        Thread not running

        pid: N/A
    LocationUpdateTask
        LocationUpdateTask
        status: error
        error: 1970/1/1 (Thu) 00:00:00 UTC
        error message:

        Thread not running

        pid: N/A
    data collect error: undefined
```

ticoann:
Alan I think it is related to the log db bug which is already fixed. However older version of agent (1.0.21) doesn't have this fix. I have temporary patch to all the production agents. #7635 
I will monitor whether that is the issue (or only issue).

ticoann:
already fixed in new agent. Current production is updated with #7635

PR #7635, opened 2017-02-16
temp fix for logdb delete - current version has realfix

ticoann:
related to #7627

Applied to all the production agent (only 1.0.21 version) and WorkQueueManagers are restarted.


ticoann:
It seems this was the problem, I am merging this although I don't think we will redeploy this version.

PR #7636, opened 2017-02-16
Move utilitarian function from WMWorkloadTools to Utils

amaltaro:
Fixes #7633

Moving any generic python functions that don't depend on WMCore code.
I guess we don't want to move functions that are too specific to WMCore (e.g. parsePileupConfig), do we?

PR #7637, opened 2017-02-17
Rename Utilities test class

amaltaro:
Complement to #7636

### DIFF
--- a/test/python/Utils_t/Utilities_t.py
+++ b/test/python/Utils_t/Utilities_t.py
@@ -6,11 +6,10 @@ Unittests for Utilities functions
 from __future__ import division, print_function
 
 import unittest
-
 from Utils.Utilities import makeList, makeNonEmptyList, strToBool, safeStr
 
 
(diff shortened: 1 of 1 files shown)


test/python/Utils_t/Utilities_t.py
```diff
@@ -75,7 +74,7 @@ def safeStr(self):
         self.assertEqual(safeStr(""), "")
 
         for v in [[1, 2], {'x': 123}, set([1])]:
-            self.assertRaises(safeStr(v), '123')
+            self.assertRaises(ValueError, safeStr, v)
```
ericvaandering: This is fine. In general I think the 
'''
with self.assertRaises...
    safeStr(v)
```
form is more readable.

ticoann: I guess I am not much pythonian yet. Both syntax are confusing to me. :-)

ticoann (approved):

PR #7640, opened 2017-02-17
fail MC workflows with too many lumis in one element

### LINKED ISSUE #7575: Fail request acquisition in GQ if there are way too many lumis
As a follow up of https://github.com/dmwm/WMCore/issues/7560 and private discussions.
There may be cases where we have gazillions of lumis in the same WQE (or file), e.g.
 pdmvserv_task_BPH-PhaseIIFall16GS82-00010__v1_T_170124_203731_9213

with 23M lumis per file (since it's MC). This just makes this workqueue thread `WorkQueueManagerWMBSFileFeeder` to get kind of stuck when inserting run/lumi into the local SQL database.

My idea is to use the ratio of `TotalInputLumis / TotalEstimatedJobs` and fail the work creation in global workqueue if beyond a specific threshold (200k lumis?) Ideally we should make it configurable.

ticoann:
fixes #7575

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/MonteCarlo.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/MonteCarlo.py
@@ -111,7 +111,8 @@ class MonteCarloWorkloadFactory(StdBase):
             self.eventsPerLumi = self.eventsPerJob
         self.prodJobSplitArgs = {"events_per_job": self.eventsPerJob,
                                  "events_per_lumi": self.eventsPerLumi,
-                                 "lheInputFiles": self.lheInputFiles}
+                                 "lheInputFiles": self.lheInputFiles,
+                                 "MaxLumisPerWQElement": self.maxLumisPerWQElement}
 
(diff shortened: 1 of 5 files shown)


ticoann:
@amaltaro, Alan please review. Tested in my vm, it fails the request if the WQ element contains more lumis that threashold set in spec. MC workflow is only affected by this.
Also I threshold is on WQE not at job. After going through the code, bulk lumis are inserted per (fake) input file which is one in WQE in MC, not by job.

amaltaro (approved):
Changes look good and it's interesting to provide the flexibility to change this threshold. BTW, it will also affect StepChain and TaskChain starting from scratch, thus we could also specify this value over there to get the proper type and validation function. Anyways, it's not a must IMO, so this one should be good to go.

ticoann:
Thanks Alan, I added threshold for TaskChain, StepChain as well. I will run the quick test.

ticoann:
replaced by #7641

PR #7643, opened 2017-02-21
Fix setup_build issue with empty PYTHONPATH and add wmc-wmarchive build key

vkuznet:


ericvaandering:
Can someone please go back and address the worst of the pylint reported problems here? Indentation with tabs and the wrong # of spaces could really bite us.

ticoann:
Sorry I will fix the indentation and other issue today.

PR #7650, opened 2017-02-22
Support Step level dataset args in StepChain

### LINKED ISSUE #7155: Assigning StepChain with dictionnary AcquisitionEra
I assigned 

https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/vlimant_task_BPH-RunIIFall15DR76-00057__v1_T_160905_213311_4199

with

{
  "SoftTimeout": 159600,
  "MaxVSize": 4394967000,
  "BlockCloseMaxEvents": 2000000,
  "MinMergeSize": 2147483648,
  "CustodialSites": [],
  "MaxMergeSize": 4294967296,
  "NonCustodialSites": [
    "T1_FR_CCIN2P3_Disk"
  ],
  "ProcessingVersion": 1,
  "MaxRSS": 2027520,
  "ProcessingString": {
    "BPH-RunIIFall15MiniAODv2-00057_0": "PU25nsData2015v1_76X_mcRun2_asymptotic_v12",
    "BPH-RunIIFall15DR76-00057_1": "PU25nsData2015v1_76X_mcRun2_asymptotic_v12",
    "BPH-RunIIFall15DR76-00057_0": "PU25nsData2015v1_76X_mcRun2_asymptotic_v12"
  },
  "Teamproduction": "checked",
  "Dashboard": "reprocessing",
  "SiteWhitelist": [
    "T1_FR_CCIN2P3",
    "T2_US_MIT"
  ],
  "TrustSitelists": false,
  "AutoApproveSubscriptionSites": [
    "T1_FR_CCIN2P3_Disk"
  ],
  "UnmergedLFNBase": "/store/unmerged",
  "CustodialSubType": "Replica",
  "TrustPUSitelists": false,
  "GracePeriod": 300,
  "MergedLFNBase": "/store/mc",
  "MaxMergeEvents": 200000,
  "checkboxvlimant_task_BPH-RunIIFall15DR76-00057__v1_T_160905_213311_4199": "checked",
  "SiteBlacklist": [],
  "action": "Assign",
  "AcquisitionEra": {
    "BPH-RunIIFall15MiniAODv2-00057_0": "RunIIFall15MiniAODv2",
    "BPH-RunIIFall15DR76-00057_1": "RunIIFall15DR76",
    "BPH-RunIIFall15DR76-00057_0": "RunIIFall15DR76"
  },
  "NonCustodialSubType": "Replica"
}

and 

"OutputDatasets":[
"/BdToKK_BMuonFilter_SoftQCDnonD_TuneCUEP8M1_13TeV-pythia8-evtgen/RunIIFall15DR76-PU25nsData2015v1_76X_mcRun2_asymptotic_v12-v1/AODSIM",
"/BdToKK_BMuonFilter_SoftQCDnonD_TuneCUEP8M1_13TeV-pythia8-evtgen/RunIIFall15DR76-PU25nsData2015v1_76X_mcRun2_asymptotic_v12-v1/DQMIO",
"/BdToKK_BMuonFilter_SoftQCDnonD_TuneCUEP8M1_13TeV-pythia8-evtgen/RunIIFall15DR76-PU25nsData2015v1_76X_mcRun2_asymptotic_v12-v1/MINIAODSIM"

they all have the same acquisitionera despite the chain-specific setting.
Is there anything special needed to assign StepChain ?

amaltaro:
Fixes #7155

Properly builds a StepChain request taking into consideration the Step level settings (AcqEra/ProcStr/ProcVer).
In case those values are overwritten during assignment, the code is ready to nicely handle and update those properties at step (and output module) level.

BTW, most of these changes should be transparent to anything but StepChain requests.

TODO: I still have to remove all my loggings.
TODO2: make sure the tests will work just fine. Spec-wise, changes seem Ok.
TODO3: squash the first 3 commits

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -203,7 +203,7 @@ class WMTaskHelper(TreeHelper):
         setattr(self.data.steps, "topStepName", stepName)
         return
 
-    def listAllStepNames(self):
+    def listAllStepNames(self, cmsRunOnly=False):
         """
         _listAllStepNames_
 
@@ -211,7 +211,10 @@ class WMTaskHelper(TreeHelper):
         """
         step = self.steps()
         if step:
-            return step.allNodeNames()
+            stepNames = step.allNodeNames()
+            if cmsRunOnly:
+                stepNames = [step for step in stepNames if step.startswith("cmsRun")]
+            return stepNames
         else:
             return []
 
@@ -1268,13 +1271,15 @@ class WMTaskHelper(TreeHelper):
                 IDs.append(ID)
         return IDs
 
-    def setProcessingVersion(self, procVer, parentProcessingVersion=0):
+    def setProcessingVersion(self, procVer, parentProcessingVersion=0, stepChain=False):
         """
         _setProcessingVersion_
 
         Set the task processing version
         """
-        if isinstance(procVer, dict):
+        if isinstance(procVer, dict) and stepChain:
+            taskProcVer = self._getStepValue(procVer)
+        elif isinstance(procVer, dict):
             taskProcVer = procVer.get(self.name(), parentProcessingVersion)
             if taskProcVer is None:
                 for taskname in procVer:
@@ -1285,7 +1290,7 @@ class WMTaskHelper(TreeHelper):
 
         self.data.parameters.processingVersion = int(taskProcVer)
         for task in self.childTaskIterator():
-            task.setProcessingVersion(procVer, taskProcVer)
+            task.setProcessingVersion(procVer, taskProcVer, stepChain)
         return
 
     def getProcessingVersion(self):
@@ -1296,13 +1301,15 @@ class WMTaskHelper(TreeHelper):
         """
         return getattr(self.data.parameters, 'processingVersion', 0)
 
-    def setProcessingString(self, procString, parentProcessingString=None):
+    def setProcessingString(self, procString, parentProcessingString=None, stepChain=False):
         """
         _setProcessingString_
 
         Set the task processing string
         """
-        if isinstance(procString, dict):
+        if isinstance(procString, dict) and stepChain:
+            taskProcString = self._getStepValue(procString)
+        elif isinstance(procString, dict):
             taskProcString = procString.get(self.name(), parentProcessingString)
             if taskProcString is None:
                 for taskname in procString:
@@ -1314,7 +1321,7 @@ class WMTaskHelper(TreeHelper):
         self.data.parameters.processingString = taskProcString
 
         for task in self.childTaskIterator():
-            task.setProcessingString(procString, taskProcString)
+            task.setProcessingString(procString, taskProcString, stepChain)
         return
 
     def getProcessingString(self):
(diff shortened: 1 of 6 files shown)


amaltaro:
And it seems to work now. But If I provide those values again during assignment, then it gets all screwed... investigating...

amaltaro:
The new unit test works, I'm proud of myself. :1st_place_medal: 
Well, it works until it gets assigned :)

amaltaro:
Yay!!! It seems to be working now. 
@ticoann this PR touches the baseline WMCore, but I tried to make it the most transparent (and less hacky) possible. Checking the spec built, it should be working just fine (there shouldn't be the need to update the agents BTW).

Just to be on the safe side, I've just started running the whole battery of tests DMWM + Integration. If everything is fine, we should be able to merge it later in the weekend or on Monday.

Nonetheless, could you please starting reviewing these changes (don't bother about those loggings)?

ticoann:
Hi Alan, personally, I would prefer we only allow setting the parameters in the nested structure and doesn't accept dictionary format. (I think we discussed this long time before for TaskChain. :-)
```
"Step1" : {"AcqEra": "blah", "StepName":"A"}

and not Allow

"AcqEra": {"A" : "blah"}
```

But I guess that ship has sailed. 😄 

As far as code goes, I am wondering whether change is only can be made in StepChain spec, not modifying StdSpec file. (I will try to look closely tomorrow). But if you are already tested and everything works fine you shouldn't change the the patch. We can refactor later. Anyway, I will review closely tomorrow.

Sorry for the late review.

amaltaro:
It seems that was discussed here: https://github.com/dmwm/WMCore/pull/5955 . 
According to the last comment, there are two cases that should be supported (at least since 2015), though we only use (and test) the latter one, since TaskChain day 1, I believe. 

IMO supporting both structures is just too much and source for problems, if I were to choose, I'd pick the first one, though that would be a breaking change and needs to be properly tested and sync'ed with Ops.

About isolating those changes to the StepChain spec, it can be done but then we'd have to override all those StdBase methods (addMerge, addOutput, setupProcessing...), meaning quite a bit of duplicated code.

My test workflows are still running, they should finish by tomorrow morning and then I can say whether the patch works fine or not :-)

ticoann:
Hi Alan, I agree that keeping only one structure, but the preference is different. I think if we keep the  nested structure, we don't have to change the current code.
Anyway, while I am going through patch, I was wondering whether we can do this different way.
By just converting the dictionary format to nested format before we apply to any other function.
I am not sure that will work, though. Could you check on the #7668 whether this will do the same thing as this patch? (I didn't test anything :-), I just want you to check the logic).

But I said, after this patch is tested, if you are confident you can merge and refactor later. 
Thanks,

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -1355,13 +1362,36 @@ def setNumberOfCores(self, cores, nStreams):
...
+        """
+        value = None
+        if self.taskType() == "Merge":
+            extractedTaskName = self.name().split("Merge")[0]
+            value = keyDict.get(extractedTaskName)
+        elif self.taskType() in ["Production", "Processing"]:
+            value = keyDict.get(self.name())
+
+        return value
+
```
ticoann: So task name is actually step name here?  When it is merge step there is post fix "Merge" on StepName?

ticoann:
Actually, you are right. It seems it is easier to keep the dictionary format in step chain, then nested format. I failed to make simpler code.  😄

src/python/WMCore/WMSpec/WMTask.py
amaltaro: That's correct! Have a look at the unit test example where it lists all task names and you can clearly see how the StepName value is mapped to an actual task (but in the end, what you said is correct ;)

amaltaro:
All the tests look Ok and I couldn't spot any issues. I'm removing all my loggings and squash it, and then it's good for a last review and merge, from my side.

ticoann:
Thank you very much Alan. I created new tag with this patch. 1.1.1.pre7

PR #7657, opened 2017-02-23
fix dn record for state transiton

### LINKED ISSUE #7648: RequestTransition records the wrong DN
https://cmsweb.cern.ch/reqmgr2/fetch?rid=prozober_task_EXO-RunIISummer16DR80Premix-07611__v1_T_170220_173516_7037

unless really mistake @prozober has not toggled the status into acquired, running, ...

ticoann:
fixes #7648, need to add deployment script https://github.com/dmwm/deployment/pull/443 as well


ticoann:
Alan please review. Test in my vm. (Although I am using the same DN for both cases. 😄 )

amaltaro (approved):
I'm not sure I know the difference between ReqMgr service and ReqDBWriter :-) I guess the latter access the couchdb documents directly (so possible no DN information?) 
Patch looks good though :)

ticoann:
Yes that is exactly the difference, (Also, ReqMgr goes through validation) So ReqDBWriter will be more performative (don't know what would be exact difference). 
In general within the ReqMgr or if there is no validation needed when application access the API, ReqDBReader/Writer can be used.

PR #7659, opened 2017-02-24
Check size of /data1 during WMA deployment

amaltaro:
Just to avoid the annoying question even when /data1 is quite small.


amaltaro:
Already tested, so I'm merging it.

PR #7660, opened 2017-02-24
Check pnn and command exist in SiteConfig fallback.

alexanderrichards:
This patch avoids a KeyError Exception being thrown when 'phedex-node'
and 'command' keys are not present in the fallback stage out definitions in
the site config. It explicitly checks for their absence and raises an
appropriate StageOutInitError.

	modified:   src/python/WMCore/Storage/StageOutMgr.py

### DIFF
--- a/src/python/WMCore/Storage/StageOutMgr.py
+++ b/src/python/WMCore/Storage/StageOutMgr.py
@@ -120,7 +120,19 @@ class StageOutMgr:
 
         msg += "There are %s fallback stage out definitions.\n" % len(self.fallbacks)
         for item in self.fallbacks:
-            msg += "\tFallback to : %s using: %s \n" % (item['phedex-node'], item['command'])
+            pnn = item.get("phedex-node")
+            command = item.get("command")
+            if pnn is None:
+                msg = "Unable to retrieve fallback phedex-node\n"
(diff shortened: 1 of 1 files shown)


amaltaro (commented):
Nice catch, Alex!
I was just wondering, how about we make these error messages a bit more generic like `Unable to retrieve fallback command/phedex-node blah...."`
?

Tha same would apply to the standard stage out settings (above).

In the end it's a trade-off, a cleaner code but a not very precise error message, thus I'm not sure :-)
@ticoann @hufnagel any thoughts on it?

ticoann:
I think precise error message is better. So we can see where it fails, although currently it seems stageout error is not send back to agent.
How about clean up a little.

```

msg = "Unable to retrieve fallback %s\n"
msg += "From site config file.\n"
msg += "Unable to perform StageOut operation"

if pnn is None:
    raise StageOutInitError(msg % "phedex")
if command is None:
    raise StageOutInitError(msg % "commend")

And for the same as other code which has the same message.
```

hufnagel:
Precise error message is better I think.

amaltaro:
@ticoann I suggest we merge this PR as it is. If needed, I can update this code later on, I have some changes on my private branch already.

PR #7662, opened 2017-02-24
Inject/Assign script can now handle step level assignment of StepChains

amaltaro:
There are three changes here:
1) handle StepChain assignment in a step basis, meant to handle cases as implemented here #7650
2) list of ScramArchs rolled back to a single OS, since frontend/factory is not ready to handle it.
3) update the validate script to keep going when a workflow doesn't have the workload summary couch doc. Also updated how we look for harvested datasets in order to handle the multirun harvesting case.


amaltaro:
Merging it such that people can use the up-to-date templates and inject script.

PR #7663, opened 2017-02-24
New tables for by work 

ericvaandering:


ericvaandering:
@ticoann this could be merged after I write the Oracle equivalents or I could continue with this PR to add the DAOs. Which approach would you prefer?

ericvaandering:
Can @ticoann or maybe @hufnagel take a look at (especially) the last commit? I need to add a primary key to an existing table and its associated sequence table. Did I do it right? The SQL for that insert was more complicated than I'm used to.

hufnagel:
What are you trying to do with this ? Looks fishy to me. Don't add some meaningless id to what is essentially an association table. Instead declare a combination of the existing field as PK.

Fileid, run, lumi already IS the PK for that table, just declare it as such and forget about the extra id that will just take up extra space and slow things down.

hufnagel:
In general, ids are only needed on tables that represent enumerable items that we create (jobs, files, datasets, blocks etc). Tables that represent relationships between these should not have ids since the relationship itself and the constraints on it define the table.

In this case, the combination of file, run and lumi is unique and is therefore perfect for a PK.

hufnagel:
Ok, I didn't read the other code changes. You are trying to introduce file/run/lumi as an enumerable item that can be referenced elsewhere ? Hmmm. I wouldn't do it.

If some other table needs file/run/lumi information just add file, run, lumi information. That's cleaner than inventing a new id that isn't really an independent id. Is there a reason you don't think that will work ?

Adding an index on file,run,lumi in the other table and having a PK for file, run, lumi will be equally fast compared to adding that new id.

ericvaandering:
Adding it to another table will work just fine. I wanted to a) keep things normalized and b) keep the data sizes down. Maybe since these tables eventually get cleaned up, that's not a big deal. It certainly would be easier to code that way. 

I think this came out of wanting an association table from work unit to FRL since the same RL can occur in different files. But I suppose rather than having an association table between two other tables, I can just have a work unit, file, run, lumi table that takes care of that problem in two tables instead of three. Sound better?

hufnagel:
You can still have an association table if you want. Have two tables

work unit, file, run, lumi

file, run, lumi 

Where the PK of the second one is FRL and the first one has a FK constraint (with index on the FK) on FRL in the second. Then any kind of joins between these tables will be as fast as if you had your new id.

These are just numbers in the end. Wouldn't treat strings this way, but for numbers the storage costs are manageable (and don't forget that adding that extra id costs storage too).

ericvaandering:
Thanks @hufnagel , but now I'm confused. file, run, lumi exists already. It's called wmbs_file_runlumi_map. If I repurpose my new table wmbs_file_workunit_assoc to be work unit, file, run, lumi then I don't see a need to ever join with wmbs_file_runlumi_map as the info is a complete duplicate of what's there already. Are you proposing the FK and constraint just so they are guaranteed to be in sync?

hufnagel:
Yes, to some degree. The FK relationships and indexes on top of them speed up joins, but they also constrain your data model and relationships. If you clearly define them, there is less of a chance that application code does the wrong thing. If you try, you will get a db error since you violate some constraint. That's preferable to have some three level down application code getting confused, having to debug it and then find out that your data in the db is internally inconsistent.

Also, to repeat advice I got from an Oracle admin at CERN, they more you constrain your data relationships, the better the optimizer works since it'll know how your data is related to each other and which optimizations will work and which won't.

yuyiguo:
From what I understood when talked with Eric, the file_run_lumi table is used to be a map table only between file and run/lumi. There is no external primary key needed, the primary key is simple the three(file, run, lumi) together. Now Eric wants to join file_run_lumi table with another table Work_unit. The map between file_lum_run and work_unit are many to many. So there will be a map table (Say M) in between. M will have a foreign key as file_run_lumi's primary, another foreign key as work_unit's primary key. Because the primary key of file_run_lumi will be a foreign key for M, it is better practice to create a external primary key instead of using the three key combination. Saving space is not a main factor here because they are all int after all. The data integrity and join is the main concern here. The relational database cannot be optimized without proper relationship. 

I  am not sure if I explained the on the "on delete cascade" clearly.  The example is that the BLOCKS table is the "parent" table, The FILES table has a foreign key (block_id) from BLOCKS' primary key. So the DDL is:
ALTER TABLE FILES ADD CONSTRAINT BK_FL 
    FOREIGN KEY (BLOCK_ID) REFERENCES BLOCKS (BLOCK_ID) ON DELETE CASCADE;
So when a block is deleted from the BLOCKS, all the files that has the block's primary id as foreign key will be deleted too. W/O ON DELETE CASCADE, one has to first delete the files, then the block may be deleted.

hufnagel:
As long as your join statement specifies all three fileid, run, lumi , it doesn't matter if you have a composite PK and a FK constraint on that composite, the join will work the same way as if you added an external id based PK.

hufnagel:
Anyways, as Eric pointed out, the join isn't even the issue here since he'll never do a join. Basically, as-is, the FRL table only exists to constrain the allowed FRL combinations. Which is useful in it's own right as it constrains the data relationships and what is allowed for other records that contain and reference FRL.

ericvaandering:
Ok, I've gotten a bit lost in the discussion and my SQL is not good enough to understand how to do this without an example. And I think I've got a mistake in my code as it is. 

So, lets say a work_unit is the run/lumi/part-of-lumi (that why those event counts are there, for future expansion). It has an ID. 

If each run/lumi was in one file, we could just add file/run/lumi (FRL) to that table and be done. But that's not always the case, so I need another table that associates work_unit_id to one more more file, run, lumi ID. Dirk's suggestion is to put that all in another table rather than what I was trying to do which is build an association map between wmbs_work_unit and wmbs_file_runlumi_map. OK, fine. I can do that.

What I don't know how to do and could not easily find an example of is how do I make sure that every FRL in wmbs_file_workunit_assoc is also in wmbs_file_runlumi_map and when deletes happen in wmbs_file_runlumi_map that the equivalent rows are deleted in wmbs_file_workunit_assoc? Actual SQL would be really helpful. 

And, of course, wmbs_file_workunit_assoc is probably not a correct name for this table anymore.

Thanks!

yuyiguo:
Eric,

wmbs_file_runlumi_map should be the associated table between  work_unit and FRL. The map table do exact the same as the associated table you was originally doing. The map/associated table connect the many to many relationship between FRL and work_init.

The deletion in map/associated table will not trigger any deletion in any other tables because the is the end tables.   

We can discuss with Dirk tomorrow.

ericvaandering:
wmbs_file_runlumi_map is the list of which RL pairs are in files. I'm not going to repurpose that nor could I.

ticoann:
Sorry for the joining the discussion late. I don't know enough to argue whether it is better to have new id for wmbs_file_runlumi_map or just add 3 keys to new table directly.
I would have ids for keeping the data integrity easily, but whatever other experts (Yuyi, Dirk) agree on. :-)

> @ticoann this could be merged after I write the Oracle equivalents or I could continue with this PR to add the DAOs. Which approach would you prefer?

I would prefer continue PR with DAO. I am not sure until changes are ready whether we can merge this. We could create a new branch for this and merge whole things later as well.

One question I have is that work unit only contains only one lumi? It can't have a range or lumis?

ericvaandering:
A work unit should only have one run/lumi pair, I think. A job has multiple work units. So I guess I could put that into the work unit and then just have a many to one relationship of files to work units.

ericvaandering:
Seems its best to wait until Friday when @hufnagel is back at FNAL (?) and have a sit-down discussion.

hufnagel:
Yes, that would be good. Only getting back late Thursday, will be at the lab Friday.

Multiple workunits can have the same file/run/lumi once you get into sub-lumi splitting. Not sure how much you want to already address this in the schema though...

ericvaandering:
From @hufnagel : 

On example primary keys:

https://github.com/hufnagel/T0/blob/master/src/python/T0/WMBS/Oracle/Create.py

Check for the lumi_section table creation and where that PK is referenced as a FK later to constrain another table. That syntax should work the same on MySQL.

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
@ticoann as we discussed, you can test and then merge this while I turn my attention to the basic classes that will underly this DB schema

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ticoann:
test this please

ticoann:
@ericvaandering, Eric, I was trying to test this under oracle and getting following error.

```
(DatabaseError) ORA-00904: : invalid identifier
 'CREATE TABLE wmbs_workunit( id INTEGER NOT NULL, taskid INTEGER NOT NULL, retry_count INTEGER DEFAULT 0, last_unit_count INTEGER NOT NULL, last_submit_time INTEGER NOT NULL, status INTEGER DEFAULT 0,)
```

It seems Oracle doesn't accept trailing ',' at the end. INTEGER DEFAULT 0,)

Could you update the patch?

ericvaandering:
@ticoann done

ticoann (approved):

ticoann:
Thanks Eric

src/python/WMCore/WMBS/Oracle/Create.py
```diff
@@ -650,6 +655,58 @@ def __init__(self, logger=None, dbi=None, params=None):
...
+                                                   '(CONSTRAINT wmbs_workunit_fk_taskid'
+                                                   ' FOREIGN KEY(taskid) REFERENCES wmbs_workflow(id)'
+                                                   ' ON DELETE CASCADE)')
+
+        # Association table between jobs and workunits, indices come from CreateWMBSBase.py
+        self.create["22wmbs_job_workunit_assoc"] = (
+            'CREATE TABLE wmbs_job_workunit_assoc ('
+            ' job    INTEGER NOT NULL,'
+            ' workunit INTEGER NOT NULL'
+        )
```
ticoann: Sorry there is missing ')'

ticoann:
@ericvaandering, Eric, there is one more problem. (Oracle table/index name should be 30 or less)
And below is 31.

```
CREATE INDEX idx_wmbs_job_workunit_assoc_job ON wmbs_job_workunit_assoc(job) 

(DatabaseError) ORA-00972: identifier is too long
```

ericvaandering:
test this please

ericvaandering:
@ticoann 18th time is the charm?

ticoann:
Sorry it seems constraint name should be short too.

ticoann:
Eric, Sorry, one more problem. Not sure how to fix it. It seems foreign key has to be whole primary key but composite primary key spread in different tables.

```
ALTER TABLE wmbs_frl_workunit_assoc ADD (CONSTRAINT wmbs_frl_workunit_assoc_fk_workunit FOREIGN KEY(workunit) REFERENCES wmbs_workunit(id) ON DELETE CASCADE)

(DatabaseError) ORA-00972: identifier is too long
```

ticoann:
unittest failure doesn't seems to be related. Merging it.

PR #7655, opened 2017-02-22
provide protected lfn api

ticoann:
fixes #7645 and #7651. Also change the mask return format to for task and step chain from the discussion with Alan.

PR #7665, opened 2017-02-24
Get unique size for /data1 partition

amaltaro:
For some reason the physical nodes return two records for /data1 partition (block device), thus it returns two partition sizes.

### DIFF
--- a/bin/deploy-wmagent.sh
+++ b/bin/deploy-wmagent.sh
@@ -154,8 +154,8 @@ else
   exit 1
 fi
 
-DATA_SIZE=`lsblk -bo SIZE,MOUNTPOINT | grep ' /data1' | awk '{print $1}'`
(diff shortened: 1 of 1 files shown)


amaltaro:
Just tested it in vocms0310, so I'm merging it :)

ticoann:
Sure go ahead. I think all unit test fails due to this patch.
https://github.com/dmwm/WMCore/pull/7655
I made a mistake in the dependency. I will fix that.

amaltaro:
Bad job! ;)

amaltaro:
Hehe, I got a failure merging it, it seems we pushed the button almost at the same time (but you won :))

ticoann:
It was my mistake. I was pushing it merge button. I was trying to reply your Bad job comment.

PR #7668, opened 2017-02-25
take dict param for stepchain

ticoann:
Possibly replace #7650. Not at all tested. @amaltaro, Alan, could take a quick look whether this is completely non sense or not?

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -441,7 +441,7 @@ class StdBase(object):
     def addOutputModule(self, parentTask, outputModuleName,
                         primaryDataset, dataTier, filterName,
                         stepName="cmsRun1", forceMerged=False,
-                        forceUnmerged=False):
+                        forceUnmerged=False, forceTaskName=None):
         """
         _addOutputModule_
 
@@ -452,7 +452,7 @@ class StdBase(object):
         haveProcString = (self.processingString != None and self.processingString != "")
         haveRunNumber = (self.runNumber != None and self.runNumber > 0)
 
-        taskName = parentTask.name()
+        taskName = forceTaskName or parentTask.name()
         acqEra = self._getDictionaryParams(self.acquisitionEra, taskName)
         procString = self._getDictionaryParams(self.processingString, taskName)
         procVersion = self._getDictionaryParams(self.processingVersion, taskName, 1)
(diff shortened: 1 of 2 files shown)


ticoann:
Also I have one question, For step chain, the first step is basically a Task, correct?

ticoann:
@amaltaro, Alan, don't review this one, I try to  make a little cleaner but getting messier. 😄

amaltaro:
First, stop working.
Second, too late, I already have a big text with some pointers :) I'll post them in a few minutes

ticoann:
Yes, it is hard to get the  Step level args, this patch only handles dict format. I initially thought nested level (Step level definition is automatically handled but it seems not the case).
And there are code I don't quite understand, (Merge case).

amaltaro:
This is the basic structure of a StepChain request:
 *) one production/processing task (let's say AllStepsTask), 
  -) inside this single task, we have the output modules section (they are not a task here). Each of them contain the definition of the unmerged files created (that would be inherited from the task, 1st problem!)
 **) for each output module in AllStepsTask, we also create a merge Task (the first merge task is created at the moment we create the top level task - so StdBase uses the workload properties without, 2nd problem!).
 ***) and once we get into the assignment business, it goes beyond the spec files and it gets much dirtier :-(

All in all, the task properties (AcqEra/ProcStr/ProcVer) doesn't mandate what a Task is going to produce - and it's supposed to be a simple string/integer value - thus we need to make sure the output modules section are updated accordingly.

Looking at this example
https://alan-cloud1.cern.ch/reqmgr2/config?name=amaltaro_StepChain_Tasks_StepChain_Validation_170224_131205_6908

these kind of parameters are not correct:
```
amaltaro_StepChain_Tasks_StepChain_Validation_170224_131205_6908.tasks.GENSIM.parameters.processingString = 'GENSIM_StepChain_Tasks_StepChain_Validation_TEST_Alan_v1'
amaltaro_StepChain_Tasks_StepChain_Validation_170224_131205_6908.tasks.GENSIM.parameters.acquisitionEra = 'Integ_TestStep1'
```
since they actually correspond to the first step (cmsRun1) only, but the further two steps would need different settings. Which's why we have to edit the output module sections. In that config page, look for `.transient =`. You'll find 4 True (creating unmerged files, running all in the only Production task) and then 4 False, corresponding to the merge tasks and creating the merged files. They need to match, otherwise I believe the agent could get lost merging them (and we don't want incorrect info between them). When looking at those transient stuff, a few lines below you should see `processedDataset`, AFAICS this is how we build the list of outputDatasets.

amaltaro:
(I stil didn't check my tests, first a coffee :-)) but I suggest we try to refactor it in the next month or so, if desired. Though I'm not sure we can make it much better, it's a nasty thing to do

ticoann:
I think refactor would be major change. We need a codefest for that. 😄

PR #7672, opened 2017-02-28
fix Timeout causing null in sched_status in bl_runjob table

ticoann:


ticoann:
updated in production agents and db entry is modified manually
vocms0116 - 85
vocms0130 - 95
cmsgwms-submit1 -1

PR #7673, opened 2017-02-28
Disallow sched_status to be null in the db schema

### LINKED ISSUE #7670: Losing tracks of the condor job.
Some of the jobs in bl_runjob table sched_status is null.

i.e. There are 97 of them in vocms0116 and 90 of them in vocms0130. Other agents has no such a case.

```
select count(*) from bl_runjob where sched_status is null;

```
This causes following query not to return records in this cases.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/MySQL/LoadRunning.py#L24

That seems to be the reason those jobs can't be completed.

I am not completely sure whether this is SimpleCondor plugin related or previously could happen. (Although it seems only new agent have this problem, I couldn't find anything wrong in SimpleCondor plugin).

Here is the place sched_status should be set

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/BossAirAPI.py#L167

and

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L189

But both cases it should be set at least "New" and "Unknown", it can't be setting something not known to db.

1	Unknown
2	Completed
3	Running
4	Held
5	Idle
6	TransferOutput
7	Suspended
8	New
9	Removed

amaltaro:
Fixes #7670 , which is basically that Timeout line in SimpleCondorPlugin.
However, I also removed the begin/end transaction from all get/select methods in BossAir.
In addition to that, we better have this protection in that table for the future plugins to come :)

@ticoann please review after unittests are finished (I still have to test it btw)

### DIFF
--- a/src/python/WMCore/BossAir/BossAirAPI.py
+++ b/src/python/WMCore/BossAir/BossAirAPI.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
-#pylint: disable=E1103, E1101, C0103
-#E1103: Use DB objects attached to thread
-#E1101: Create config sections
-#C0103: Internal methods start with _
+# pylint: disable=E1103, E1101, C0103
+# E1103: Use DB objects attached to thread
+# E1101: Create config sections
+# C0103: Internal methods start with _
 """
 _BossAirAPI_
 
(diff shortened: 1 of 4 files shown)


amaltaro:
Sorry @hufnagel , I saw your comments on the issue only now. I guess I was blind when I ruled out the Timeout issue. You may want to review this PR as well.

amaltaro:
unrelated test failure with logDB, but firing the test again just in case

amaltaro:
test this please

hufnagel:
Removing wrapping the read and select calls into transactions is the way to go, but please also pass explicitly False for the transaction to the DAO call then. No reason to check if there is an existing transaction.

hufnagel:
Actually, ignore my statement. If we ever call these functions from outside with an existing transaction checking for that could be useful.

ticoann:
Patch looks good to me. I think unittest failure is not relate.

ticoann:
test this please

ericvaandering:
You should be able to fix all the bolded warnings

ticoann (approved):

amaltaro:
pylint warnings were fixed as well. Since there is no hurry to merge it, let me first run a couple tests in the coming days.

amaltaro:
test this please

PR #7674, opened 2017-02-28
Allow different PrepID in StepChains

amaltaro:
It was setting PrepID - though using the workload level only - before my recent changes accepting Step level assignment. With those changes, it was not even setting that :(

This should fix this issue for good and properly set different PrepIDs if they were set inside a Step dict. 

There is one fix though for Cleanup task, which right now inherits PrepID from the production/processing job. I'll have to add the same `taskConf` parameter over there, I just don't want to do this at this stage of the validation.

I'll start running a few workflows soon, to make sure it doesn't break anything else.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -436,6 +436,10 @@ class StdBase(object):
 
             procTaskCmsswHelper.setDataProcessingConfig(scenarioName, scenarioFunc,
                                                         **scenarioArgs)
+        # only in the very end, in order to get it in for the children tasks as well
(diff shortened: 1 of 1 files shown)


ticoann (approved):

ticoann:
Alan, patch looks good to me. 
One thing I was wondering is - shouldn't "setPrepID" need to handle dictionary format?

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMTask.py#L1490

amaltaro:
It's always a string set during request creation and it doesn't change anymore. So no dict dirty playing needed there :)

amaltaro:
Tests look Ok, merging.

PR #7677, opened 2017-02-28
add function to get output unmerged lfnbase

### LINKED ISSUE #7645: List of protected paths
it might be simpler to expose the list of protected base lfn as made by
https://github.com/CMSCompOps/WmAgentScripts/blob/master/listProtectedLFN.py
directly from reqmgr so that we do not rely on some synchronization, sso, afs, ...
for the moment, it is 
https://cmst2.web.cern.ch/cmst2/unified/listProtectedLFN.txt
https://vocms049.cern.ch/unified/listProtectedLFN.txt

ticoann:
fixes #7645 @amaltaro, please review.

This only affects newly created workflows. So it would be better if this gets in this deployment and can be used from next deployment.

### DIFF
--- a/src/couchapps/ReqMgr/updates/updaterequest.js
+++ b/src/couchapps/ReqMgr/updates/updaterequest.js
@@ -65,6 +65,7 @@ function(doc, req) {
 	            key == "CustodialSites" ||
 	            key == "NoneCustodialSites" ||
 	            key == "AutoApproveSubscriptionSites" ||
+	            key == "OutputModulesLFNBases" ||
 	            key == "Teams") {
 	    		
 	    		doc[key] = JSON.parse(newValues[key]);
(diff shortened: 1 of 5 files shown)


ticoann:
Tested in vm
Add OutputModulesLFNBases in couch doc

https://reqmgr2-dev.cern.ch/couchdb/_utils/document.html?reqmgr_workload_cache/sryu_MonteCarlo_eff_reqmgr2_170301_061341_9060

Also provide the api

https://reqmgr2-dev.cern.ch/wmstatsserver/data/protectedlfns

amaltaro:
I was surprised `protectedlfns` was returning much less data than `protectedlfns_partial`, but the reason - I think - is that `protectectedlfns` will only fetch data info for requests created after this patch.

BTW, I found a typo in the updaterequest function and I'll create a fix for it after this one is merged. And I'm not sure I like (who cares if I do... :)) the API name and the class name, I may also add some docstring to distinguish them.

amaltaro (approved):
Thank you very much Seangchan!

amaltaro:
Seangchan, I renamed this new api here: #7678
instead of "protectedlfns_partial" it's "protectedlfns_final"

PR #7683, opened 2017-03-01
Remove user/group from ACDC collections - wmagent branch

amaltaro:
Backport of #7590 (first commit only)

It has to be applied to the production agents just after the production deployment. Testing in vocms0230


amaltaro:
BTW, we have to patch WMCore python code and couchapps, so
```
curl https://patch-diff.githubusercontent.com/raw/dmwm/WMCore/pull/7683.patch | patch -d /data/srv/wmagent/current/apps/wmagent/lib/python2.7/site-packages/ -p3
curl https://patch-diff.githubusercontent.com/raw/dmwm/WMCore/pull/7683.patch | patch -d /data/srv/wmagent/current/apps/wmagent/data/couchapps/ -p3
$manage execute-agent wmagent-couchapp-init
```
then stop/start the agent.

ticoann:
@amaltaro, Alan I will patch this for all (only) running production agent after production upgrade tomorrow.

ticoann:
Do we need to patch couch app? I think we only use the central couch for ACDC.

amaltaro:
Good point, Seangchan. I think we can skip the couchapps changes, as you said, it should be used only from central couch.

ticoann:
All agents are updated (created new patch for 1.0.21_agent, since there was conflict on this patch). And restarted

ticoann:
Alan, May be we should create the new tag for agent, let me know.

amaltaro:
Sounds good! Can you cut a patch3 tag please?

PR #7688, opened 2017-03-03
Apply futurize stage2

ericvaandering:
Just for kicks lets see how much this breaks


ericvaandering:
The answer is "a lot" but from a few problems so far.

PR #7689, opened 2017-03-03
Prepare this type checking code for python 3 compatibility. This does…

ericvaandering:
… not work as is with

the str from builtins

### DIFF
--- a/src/python/WMCore/Configuration.py
+++ b/src/python/WMCore/Configuration.py
@@ -25,20 +25,20 @@ _SimpleTypes = [
     long,  # PY3: Not needed in python3, will be converted to duplicate int
     type(None),
     int,
-    ]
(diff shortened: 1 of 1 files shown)


ericvaandering:
@ticoann please merge this. Maybe a good thing for some testing first since it is pretty basic in our stack. But before I fixed this, I had hundreds of unit test failures. So it's pretty well covered. :-)

ticoann:
Eric, can I merge this on Tuesday? We are still running the test in testbed. Although it is simple change,   just in case we have make other the fix for testbed it might be safer to not include.

ericvaandering:
Tuesday is fine. No rush. I'm going to keep going through futurize stuff and see if there is more future proofing to do.

PR #7690, opened 2017-03-03
Reorganize imports

ericvaandering:
Each of these files original format surrounding imports, docstrings, and comments caused the futurize program to put new imports in the wrong place breaking the code.

### DIFF
--- a/src/python/WMCore/ReqMgr/Tools/cms.py
+++ b/src/python/WMCore/ReqMgr/Tools/cms.py
@@ -1,16 +1,15 @@
 #!/usr/bin/env python
-#-*- coding: utf-8 -*-
-#pylint: disable=
+# encoding: utf-8
+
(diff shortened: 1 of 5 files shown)


ericvaandering:
@ticoann please merge

ticoann:
@ericvaandering, Is that a set of rules for this to work with futurize? Any other module needs to be corrected?

ericvaandering:
This is most of it, at least. I don't really understand what happens per se. If things are not done like this (and I don't know which parts are essential) then futurize likes to add imports above the ```__future``` imports. 

Thanks for merging. I'll try running futurize again and see if there are more problems

ericvaandering:
I'm not sure this even worked. I guess as we do these migrations we will just have to re-order things manually to avoid blowing stuff up.

PR #7693, opened 2017-03-06
Change the way to move to normal archived status.

### LINKED ISSUE #7671: Archive orphan announced workflows
For cases where the agent was shutdown/redeployed while workflows were not archived.
With the relval operations going to the Unified system, it could be that we have a single archiveDelay and then we can easily move it to the reqmgr2 cherrypy thread.

ticoann:
fixes #7671, need to update https://github.com/dmwm/deployment/pull/446

Also add input validation for the REST api

### DIFF
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -290,20 +290,10 @@ class CleanCouchPoller(BaseWorkerThread):
             if not self.isUploadedToWMArchive(workflowName):
                 continue
             if self.cleanAllLocalCouchDB(workflowName):
-                if self.useReqMgrForCompletionCheck:
-                    try:
-                        self.reqmgr2Svc.updateRequestStatus(workflowName, archiveState)
-                    except Exception as ex:
-                        msg = "%s : fail to update status with HTTP error: %s" % (workflowName, str(ex))
-                        msg += traceback.format_exc()
-                        logging.error(msg)
-                        raise ex
-
-                    updated += 1
-                    logging.debug("status updated to %s %s", archiveState, workflowName)
-                else:
-                    # tier0 update case
+                if not self.useReqMgrForCompletionCheck:
+                    #  only update tier0 case, for Prodcuction/Processing reqmgr will update status 
                     self.centralRequestDBWriter.updateRequestStatus(workflowName, archiveState)
+                updated += 0
         return updated
 
     def archiveSummaryAndPublishToDashBoard(self, finishedwfsWithLogCollectAndCleanUp):
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -13,7 +13,6 @@ import threading
 import traceback
 import time
 
-from httplib import HTTPException
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.Services.WMStats.WMStatsWriter import WMStatsWriter
 from WMCore.Services.RequestDB.RequestDBReader import RequestDBReader
@@ -282,7 +281,7 @@ class CleanCouchPoller(BaseWorkerThread):
         except Exception as ex:
             msg = traceback.format_exc()
             logging.error(msg)
-            logging.error("Error occurred, will try again next cycle")
+            logging.error("Error occurred, will try again next cycle: %s", str(ex))
 
     def archiveWorkflows(self, workflows, archiveState):
         updated = 0
@@ -291,7 +290,7 @@ class CleanCouchPoller(BaseWorkerThread):
                 continue
             if self.cleanAllLocalCouchDB(workflowName):
                 if not self.useReqMgrForCompletionCheck:
-                    #  only update tier0 case, for Prodcuction/Processing reqmgr will update status 
+                    #  only update tier0 case, for Prodcuction/Processing reqmgr will update status
                     self.centralRequestDBWriter.updateRequestStatus(workflowName, archiveState)
                 updated += 0
         return updated
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
@@ -3,14 +3,16 @@ Created on May 19, 2015
 """
 from __future__ import (division, print_function)
 
+import time
 from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
 from WMCore.ReqMgr.DataStructs.RequestStatus import AUTO_TRANSITION
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
 from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
+from WMCore.Services.WMStatsServer.WMStatsServer import WMStatsServer
 
 
 def moveForwardStatus(reqmgrSvc, wfStatusDict, logger):
-    
+
     for status, nextStatus in AUTO_TRANSITION.iteritems():
         count = 0
         requests = reqmgrSvc.getRequestByStatus([status], detail=False)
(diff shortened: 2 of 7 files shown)


ericvaandering:
[WMAgent does not crash on new data](https://trello.com/c/GVg53MEM/1-wmagent-does-not-crash-on-new-data)

ticoann:
test this please

ticoann:
@amaltaro, Alan basic test seems to be working. But it seems data doesn't clean up as fast as I initially thought. I need to investigate. Also, not tested against multiple agent. Validation function need to be updated.

ticoann:
@amaltaro, Alan test seems to working OK. Please review. I will fix some of pylint errors now.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -64,6 +64,39 @@ def moveToArchivedForNoJobs(reqmgrSvc, wfStatusDict, logger):
                 for nextStatus in nextStatusList:
                     reqmgrSvc.updateRequestStatus(wf, nextStatus)
                     count += 1
+        logger.info("Total %s-archived: %d", status, count)
```
amaltaro: don't you want to say `Total aborted-completed: %d` ?

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -64,6 +64,39 @@ def moveToArchivedForNoJobs(reqmgrSvc, wfStatusDict, logger):
...
+
+    return
+
+
+def moveToArchived(wmstatsSvc, reqmgrSvc, archiveDelayHours, logger):
+    """
+    Handle normal-archived trantion.
```
amaltaro: mispelled

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -64,6 +64,39 @@ def moveToArchivedForNoJobs(reqmgrSvc, wfStatusDict, logger):
...
+    
+    currentTime = int(time.time())
+    threshold = archiveDelayHours * 3600
+    count = 0
+    
+    statusTransition = {"announced": "normal-archived",
+                        "aborted-completed": "aborted-archived",
+                        "rejected": "rejected-archived"}
+    
+    inputConditon = {"RequestStatus": ["announced"], "AgentJobInfo": "NO_KEY"}
```
amaltaro: this line can be removed

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -64,6 +64,39 @@ def moveToArchivedForNoJobs(reqmgrSvc, wfStatusDict, logger):
...
+    outputMask = ["AgentJobInfo", "RequestTransition"]
+    
+    for status, nextStatus in statusTransition.items():
+        inputConditon = {"RequestStatus": [status], "AgentJobInfo": "NO_KEY"}
+        for reqInfo in wmstatsSvc.getFilteredActiveData(inputConditon, outputMask):
+            if reqInfo["RequestName"] and (not reqInfo["RequestTransition"] or
```
amaltaro: Isn't `reqInfo` a dict keyed by the request name? I wonder if we actually have the key `RequestName` in this inner dict.

amaltaro: BTW, is there any case that `RequestTransition` would be empty or None? Or is that just an additional protection in case sh** happens?

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
```diff
@@ -0,0 +1,90 @@
...
+    def _createQuery(self, queryDict):
+        """
+        _createQuery
+        :param queryDict: is the format of {name: values} fair. value can be sting, int or list
+        :type queryDict: dict
+        :returns: url query string
+        """
+        args = ""
+        for name, values in queryDict.items():
+            if isinstance(values, basestring) or isinstance(values, int):
```
amaltaro: I think `isinstance(values, (basestring, int))` works :)

amaltaro (approved):
Just to make sure I understood, all these calls use the cached data in wmstats, so there is no actual call to the reqmgr database (transitions are also stored in the wmstats, right?)

The `AgentJobInfo` thing is not that clear as well on how it's used in Request.py, but I guess I can ask you via vidyo.

Besides that, I have some minor suggestions, but in general it looks good.

ticoann:
> Just to make sure I understood, all these calls use the cached data in wmstats, so there is no actual call to the reqmgr database (transitions are also stored in the wmstats, right?)

That is correct since request are all against Active data, I am using the already cached information

> The 'AgentJobInfo' thing is not that clear as well on how it's used in Request.py, but I guess I can ask you via vidyo.

AgentJobInfo is only added when there is record for given request. Since WMAgent deletes the request from couch db. when workflow is done/archived this will get deleted. So instead of marking explicit status on wmagent-info type document in wmstats. It is just checking whether information got deleted from the agent.

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
ticoann: Thanks I didn't know.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: It returns list of dictionary in this case also it should always contain RequestName key

>BTW, is there any case that RequestTransition would be empty or None? Or is that just an additional protection in case sh** happens?

There shouldn't be for all active request. (Old request might not contain this but it shouldn't Active anymore)

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: This cannot be removed. This is condition param for wmstatsSvc.getFilteredActiveData(inputConditon, outputMask)

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: updated

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
ticoann: updated

ticoann:
Thank you for reviewing, Alan. I updated the code. If it gets though I will update the pylint error.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -64,6 +64,39 @@ def moveToArchivedForNoJobs(reqmgrSvc, wfStatusDict, logger):
...
+
+    currentTime = int(time.time())
+    threshold = archiveDelayHours * 3600
+    count = 0
+
+    statusTransition = {"announced": "normal-archived",
+                        "aborted-completed": "aborted-archived",
+                        "rejected": "rejected-archived"}
+
+    inputConditon = {"RequestStatus": ["announced"], "AgentJobInfo": "NO_KEY"}
```
amaltaro: My point was that you also set `inputConditon` in L91, so this one simply gets overwritten and never used.

ticoann: Sorry my mistake. :smile:

ticoann:
test this please

ticoann:
test this please

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -282,28 +281,18 @@ def algorithm(self, parameters=None):
...
-                        raise ex
-
-                    updated += 1
-                    logging.debug("status updated to %s %s", archiveState, workflowName)
-                else:
-                    # tier0 update case
+                if not self.useReqMgrForCompletionCheck:
+                    #  only update tier0 case, for Prodcuction/Processing reqmgr will update status
                     self.centralRequestDBWriter.updateRequestStatus(workflowName, archiveState)
+                updated += 0
```
amaltaro: BTW, this sum is funny (not to say buggy) :)

PR #7694, opened 2017-03-07
Don't sort list each iteration in FileBased splitting

emaszs:
Looks like I've discovered a minor bug which has a big performance impact for the CRAB TaskWorker component when processing large datasets (one dataset which has 160k files hangs a child process for hours). AFAICT, it makes no sense to sort the lfn list after each lfn addition.


amaltaro:
Good catch, Emilis. The failing test is actually unstable and not related to this fix. Anyways triggering it again.

amaltaro:
test this please

PR #7695, opened 2017-03-07
1.1.0_crab version of Don't sort list each iteration in FileBased splitting

emaszs:
Master PR here https://github.com/dmwm/WMCore/pull/7694
Could we also have a new tag created for our branch (1.1.0.crab3) after the merge? Thanks!


ticoann:
Thanks Emil for the fix. 1.1.0.crab3 tag is created.

PR #7696, opened 2017-03-07
Make Request compatible with Python3 

ericvaandering:
after futurize and reverting commented change

Another instance of type checking against str


ericvaandering:
Similar to what I did for Configuration. Should be tested, but should be OK.

PR #7697, opened 2017-03-08
1.0.21 wmagent version of 7683

ticoann:
ref #7683


ticoann:
All the agents are updated

amaltaro:
Thanks, Seangchan! I missed this detail that we still had very old agent releases out there that would need these patches (ScramArch list as well)

PR #7073, opened 2016-08-18
Remove gLite submission plugin

amaltaro:
Fixes #7068
because removing code is always easier.
@ericvaandering please give your blessing :)

PR #7716, opened 2017-03-14
Remove old/unused submitter plugins

### LINKED ISSUE #7588: remove CondorPlugIn and PyCondorPlugIn
We will switch to SimpleCondorPlugin in next agent deployment (Feburary)

amaltaro:
Fixes #7588
To be merged before #7073 (since it should clear some of the failing unit tests in the other PR).

I'm still keeping around the LSF, PyCondor and SimpleCondor plugins.


ticoann:
Not sure why WQ test is failing doesn't seems to be related.

ticoann:
test this please

amaltaro:
@ticoann yes, I don't think it's related to this PR, in any case I'm firing the tests again.
@hufnagel what do you think about removing the LSF plugin as well? If the chances we'll need that within 5 years or so is very small, I'd just go ahead and shoot that plugin as well.

amaltaro:
test this please

hufnagel:
I would leave the LSF Plugin. It is not obsolete in the sense that it supports a different submission system, removing it would remove functionality. SimpleCondor and PyCondor is different since they both address the same underlying submission system. Yes, we likely won't be using LSF directly anytime soon, but that is purely for operational reasons. From a code management point of view there is no reason to remove it.

Also, how many plugins will we have left after this removal ? Might be worthwhile to at least one other plugin to have an example of how things can be done differently for a different submission system.

amaltaro:
test this please

amaltaro:
We should keep the LSF, Simple and the test/Mock plugins :-)

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Only WorkQueue tests failing which I don't think is related to these changes.
Thus @ticoann please review and merge

ticoann (approved):

PR #7717, opened 2017-03-14
Add new test to make sure we don't crash on old FWJR format

ericvaandering:


ericvaandering:
@ticoann go ahead and merge. Next step is to actually extract the data out of this new XML

ticoann:
Thanks Eric,

PR #7718, opened 2017-03-14
Handle multi-scramArch in LogCollect

amaltaro:
And that's why LogCollect is failing everywhere. I thought it was the usual castor issue that we have from time to time.

Still needs to be tested, but it's probably good for review.

### DIFF
--- a/src/python/WMCore/WMRuntime/Tools/Scram.py
+++ b/src/python/WMCore/WMRuntime/Tools/Scram.py
@@ -30,6 +30,8 @@ import os.path
 import sys
 import subprocess
 import logging
+from PSetTweaks.WMTweak import readAdValues
+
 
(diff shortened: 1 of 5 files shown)


ericvaandering:
I approve of moving that function.

ericvaandering (approved):

ticoann:
So the log collect job currently failing due to this? Production agent need to be patched?

amaltaro:
That's right! And it will only work for new requests being acquired by the agent (new sandbox). 
I'll create the wmagent branch PR once I'm sure this works Ok.

ticoann (approved):

ticoann:
Looks good to me.

amaltaro:
Patch works fine. However, I'll try to fix the exit code 0 issue as well (when it fails).

amaltaro:
Actually, I'll fix that in another PR not to keep holding this one. That will likely require more testing anyways. @ticoann please merge. I'm creating the wmagent branch version soon.

PR #7725, opened 2017-03-15
typo in update count

ticoann:
Sorry my mistake. (not used anywhere but fixed anyway).


amaltaro (approved):
A complicated review.. hmm, let me think... I think it's Ok ;)

PR #7726, opened 2017-03-15
Remove duplicate logic.

bbockelm:
Except for `REQUIRED_OS`, the same Requirements logic is currently in both HTCondor `startd`s and WMAgent.  Remove the logic from the agent so we can make changes from a single place.

For `REQUIRED_OS`, we keep the logic but:
- Left a TODO to remove it once `T2_CH_CERN_HLT` has made the corresponding
  change.
- Add support for `GLIDEIN_REQUIRED_OS="any"` (for pilots that use singularity).

@ericvaandering @hufnagel @amaltaro - please note, as you've been doing `REQUIRED_OS` work recently.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
@@ -270,10 +270,10 @@ class PyCondorPlugin(BasePlugin):
         if self.serverCert and self.serverKey and self.myproxySrv:
             self.proxy = self.setupMyProxy()
 
-        # Build a request string
-        self.reqStr = ('(OpSys == "LINUX" ) && (Arch == "INTEL" || Arch == "X86_64") '
-                       '&& stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) '
-                       '&& ((REQUIRED_OS=="any") || stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS))')
+        # See notes about this in SimpleCondorPlugin.  TODO(bbockelm): remove once HLT is upgraded.
+        self.reqStr = ('(REQUIRED_OS=?="any") || '
+                       '(GLIDEIN_REQUIRED_OS=?="any") || '
+                       'stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS)')
         if hasattr(config.BossAir, 'condorRequirementsString'):
             self.reqStr = config.BossAir.condorRequirementsString
 
@@ -769,7 +769,8 @@ class PyCondorPlugin(BasePlugin):
         #    avoid condorg submission errors from > 256 character pathnames
 
         jdl.append("universe = vanilla\n")
-        jdl.append("requirements = %s\n" % self.reqStr)
+        if self.reqStr:
+            jdl.append("requirements = %s\n" % self.reqStr)
 
         jdl.append("should_transfer_files = YES\n")
         jdl.append("when_to_transfer_output = ON_EXIT\n")
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -102,10 +102,13 @@ class SimpleCondorPlugin(BasePlugin):
         self.acctGroup = getattr(config.BossAir, 'acctGroup', "production")
         self.acctGroupUser = getattr(config.BossAir, 'acctGroupUser', "cmsdataops")
 
-        # Build a requirement string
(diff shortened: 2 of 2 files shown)


bbockelm:
To work with existing pilots, would it be possible to add the following to the WMAgents:

```
config.BossAir.condorRequirementsString = '(REQUIRED_OS=?="any") || (GLIDEIN_REQUIRED_OS=?="any") || stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS)'
```

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -103,9 +103,10 @@ def __init__(self, config):
         self.acctGroupUser = getattr(config.BossAir, 'acctGroupUser', "cmsdataops")
 
         # Build a requirement string
-        self.reqStr = ('stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) '
-                       '&& ((REQUIRED_OS=?="any") || stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS))'
-                       '&& (TARGET.Cpus >= RequestCpus)')
+        # See notes about this in PyCondorPlugin.  TODO(bbockelm): remove once HLT is upgraded.
+        self.reqStr = ('((REQUIRED_OS=?="any") || '
```
amaltaro: Isn't it going to match against any sites? Since you removed the "DESIRED_Sites" logic from the requirements string.

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
```diff
@@ -270,10 +270,13 @@ def __init__(self, config):
         if self.serverCert and self.serverKey and self.myproxySrv:
             self.proxy = self.setupMyProxy()
 
-        # Build a request string
-        self.reqStr = ('(OpSys == "LINUX" ) && (Arch == "INTEL" || Arch == "X86_64") '
-                       '&& stringListMember(GLIDEIN_CMSSite, DESIRED_Sites) '
-                       '&& ((REQUIRED_OS=="any") || stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS))')
+        # Build a request string.  All CMS resources match DESIRED_Sites on the START
```
amaltaro: This plugin may soon be gone, so maybe put these comments on the other plugin instead.

amaltaro (changes requested):
I'm not sure I understood your comment. 
HTCondor (or is it glideinWMS?) already has a *default* requirement string expression. 

By setting it from the agent side, don't we have the capability to overwrite that and enforce this one to be evaluated during job matchmaking? 

If we remove it from the agent, then we no longer can say how the the production jobs are supposed to match to resources and what's taken into account (I mean, we are blindly trusting that SI will always have the correct  expr).

bbockelm:
The same site / OS matching logic is on both the worker node and in the WMAgent jobs.

Since both filters must pass, having the logic duplicated in two places makes deploying changes much, much longer and require more effort.

For example, the recent changes WMAgent has been trying with `REQUIRE_OS` require a new agent version and a corresponding change on the SI side.  If you didn't put it by default in the code, then it would only require a change on the SI side.

Put another way - has there ever been a case where the WMAgent defaults need to be more restrictive than the HTCondor defaults?

ticoann:
test this please

src/python/WMCore/BossAir/Plugins/PyCondorPlugin.py
amaltaro: @bbockelm "All CMS resources match DESIRED_Sites on the START expression side". Where is the START expression evaluated? Where is the second expression evaluated?

If any site is eligible to run jobs, then why not just remove this part of the expression and leave only the REQUIRED_OS + GLIDEIN_REQUIRED_OS validation?

amaltaro:
BTW, if we are not desperate to have this change in the global pool, I suggest we put it in only for the next agent release.

bbockelm:
@amaltaro - I think the only possible urgent need of this is if we need to do a lot of RHEL7 work before the next agent validation.  Jean-Roch asked about this earlier today and I will point him at this ticket for reference.

amaltaro:
@bbockelm one more question, once we remove the `Requirements` expression from these plugins, where and how can we find it?

amaltaro:
I asked these questions to James and Diego D. and I think it's more clear now.
Just in case, I'm running a few workflows in ITB before we push it to the production agents.

amaltaro:
@bbockelm can you squash these commits please? Thanks

amaltaro (approved):

amaltaro:
@ticoann please merge

PR #7728, opened 2017-03-15
Read new FWJR XML and store in WMAgent JobReport

ericvaandering:
First commit is just cleaning up the code before working on it

### DIFF
--- a/src/python/WMCore/DataStructs/Run.py
+++ b/src/python/WMCore/DataStructs/Run.py
@@ -1,24 +1,22 @@
 #!/usr/bin/env python
-# pylint: disable=W0104
+
 """
 _Run_
 
 container representing a run, and its constituent lumi sections
-
 """
 
-
-
 from WMCore.DataStructs.WMObject import WMObject
+
 
 class Run(WMObject):
     """
     _Run_
 
     Run container, is a list of lumi sections
-
     """
-    def __init__(self, runNumber = None, *newLumis):
+
+    def __init__(self, runNumber=None, *newLumis):
         WMObject.__init__(self)
         self.run = runNumber
         self.lumis = []
@@ -26,7 +24,6 @@ class Run(WMObject):
 
     def __str__(self):
         return "Run%s:%s" % (self.run, list(self.lumis))
-
 
     def __lt__(self, rhs):
         if self.run != rhs.run:
@@ -38,21 +35,12 @@ class Run(WMObject):
             return self.run > rhs.run
         return list(self.lumis) > list(rhs.lumis)
 
-
     def extend(self, items):
         self.lumis.extend(items)
         return
 
     def __cmp__(self, rhs):
-        if self.run == rhs.run:
-            return cmp(list(self.lumis), list(rhs.lumis))
-        if self.run > rhs.run:
-            return 1
-        if self.run < rhs.run:
-            return -1
-
-        return cmp(self, rhs)
-
+        return (self > rhs) - (self < rhs)  # Python3 equivalent of cmp()
 
     def __add__(self, rhs):
         """
@@ -63,27 +51,32 @@ class Run(WMObject):
             msg += "Run %s does not equal Run %s" % (self.run, rhs.run)
             raise RuntimeError(msg)
 
-        #newRun = Run(self.run, *self)
-        #[ newRun.append(x) for x in rhs if x not in newRun ]
-        [ self.lumis.append(x) for x in rhs.lumis if x not in self.lumis ]
+        # newRun = Run(self.run, *self)
+        # [ newRun.append(x) for x in rhs if x not in newRun ]
+        [self.lumis.append(x) for x in rhs.lumis if x not in self.lumis]
 
         return self
+
     def __iter__(self):
         return self.lumis.__iter__()
 
     def __next__(self):
         return self.lumis.__next__()
+
     def __len__(self):
         return self.lumis.__len__()
-    def __getitem__(self,key):
+
+    def __getitem__(self, key):
         return self.lumis.__getitem__(key)
-    def __setitem__(self,key,value):
-        return self.lumis.__setitem__(key,value)
-    def __delitem__(self,key):
+
+    def __setitem__(self, key, value):
+        return self.lumis.__setitem__(key, value)
+
+    def __delitem__(self, key):
         return self.lumis.__delitem__(key)
 
     def __eq__(self, rhs):
-        if not isinstance(rhs, Run) :
+        if not isinstance(rhs, Run):
             return False
         if self.run != rhs.run:
             return False
@@ -107,10 +100,10 @@ class Run(WMObject):
         Convert to JSON friendly format.  Include some information for the
         thunker so that we can convert back.
         """
-        return {"Run" : self.run, "Lumis" : self.lumis,
+        return {"Run": self.run, "Lumis": self.lumis,
                 "thunker_encoded_json": True, "type": "WMCore.DataStructs.Run.Run"}
 
-    def __to_json__(self, thunker = None):
+    def __to_json__(self, thunker=None):
         """
         __to_json__
 
--- a/src/python/WMCore/DataStructs/Run.py
+++ b/src/python/WMCore/DataStructs/Run.py
@@ -157,12 +157,13 @@ class Run(WMObject):
         Method to replace myRun.lumis.extend() which does not work with the property
         """
         for lumi in lumiList:
-            if isinstance(lumi, (list, tuple)) and self.eventsPerLumi[lumi[0]]:  # Already exists, add events
-                self.eventsPerLumi[lumi[0]] += lumi[1]
+            if isinstance(lumi, (list, tuple)) and lumi[0] in self.eventsPerLumi and self.eventsPerLumi[lumi[0]]:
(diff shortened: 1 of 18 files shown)


ticoann:
test this please

ticoann:
@ericvaandering, Eric do you want me to merge this first? Then you can create new patch for the actual code change.

ericvaandering:
No. Net yet.

ericvaandering:
test this please

ericvaandering:
test this please

src/python/WMCore/DataStructs/Run.py
```diff
@@ -101,6 +101,9 @@ def lumis(self):
     def lumis(self, lumiList):
         self.eventsPerLumi = list(lumiList)
 
+    def extendLumis(self, lumiList):
```
ericvaandering: @vkuznet for a further refactoring, I'm trying to turn lumis into a property rather than a list. But we have a few places that treat it as a list trying to directly extend or append to it.

Do you know of a way to automatically attach a method to the property so that I don't have to modify the other code. In other words, I want to still use myRun.lumis.append(42)

vkuznet:
Eric,
the problem is that you create a new list in lumis method, see
https://github.com/ericvaandering/WMCore/blob/75253e1b3ba0f9170a1e6cbd68145c7716970c77/src/python/WMCore/DataStructs/Run.py#L98

This code works:

```
class Run(object):
    def __init__(self):
        self.elumis = []

    @property
    def lumis(self):
        return self.elumis


r = Run()
print(r.lumis)
r.lumis.append(1)
print(r.lumis)
```

and it does treat lumis as property and allows to do list operation on it.

ericvaandering:
Right, but that's the way I'm going to have to implement it. Because eLumis is (later) going to be not a list, but a list of tuples. (lumi number, event count). But most of our legacy code cares about lumi numbers only, so I wanted to mimic that interface while having additional information under the hood.

src/python/WMCore/DataStructs/Run.py
```diff
@@ -1,58 +1,46 @@
...
     def __str__(self):
-        return "Run%s:%s" % (self.run, list(self.lumis))
-
+        return "Run%s:%s" % (self.run, list(self.eventsPerLumi))
 
     def __lt__(self, rhs):
         if self.run != rhs.run:
             return self.run < rhs.run
-        return list(self.lumis) < list(rhs.lumis)
+        return list(self.eventsPerLumi) < list(rhs.lumis)
```
vkuznet: Eric, this may become quite expensive operations since you create copies of existing lists. If you'll deal with lots of runs and their lumis your code may slow down substantially. This comment applies to all __op__ methods and properties you define.

vkuznet (commented):
The code also has another flaw. It does not deal with identical lumi values, since underlying structure is a list and not a set. It can create a problem when clients will add the same lumi multiple times, which will affect the hash function and therefore will treat runs differently.

vkuznet:
This will complicate things even more.

I think proper things to do would be:
- create new abstract class (AbstractLumi) to define lumi operations, such as __lt__, etc.
- create new Lumis class from AbstractLumi which will only give lumis
- create new EventLumis class from AbstractLumi which may overwrite __lt__
  methods according to new (event, lumi) data structure.
- modify Run class
  - replace lumis from list to AbstractLumi and modify all comparison methods
    to use AbstractLumis one
  - legacy code will instantiate Run with Lumis class
  - new code will instantiate Run with EventLumis

src/python/WMCore/DataStructs/Run.py
ericvaandering: I'm not sure this is ever called and will have to change for my plans anyhow. Usually the are pretty small lists as well for a particular file. (If we ran a whole day and put all the lumis into one file in some very sparse skim, we'd have a list 4000 items long.)

ericvaandering:
It's not legacy code creating a different structure. Structure will be the same, but with None as the number of events. I'll fix the problem you point out too since [(1,1000)].append((1,500)) has to give (1,1500)

ericvaandering:
I can (did) redo the few places that use extend/append with a new function name. I'll think about adding lumis as a sub-class and implementing those methods, though. That might be cleaner.

vkuznet:
The problem is that every time someone will make comparison the amount of
RAM Run class allocates will grow since all comparison operations create new
lists. It is potential memory leak, depending on a client code who may
or may not keep list of runs around. The more runs you have the more
damage may show up by doing comparison operation.

ericvaandering:
I don't understand. Those lists for comparisons are not made in client code, they are in the Run class code and should be able to be cleaned up automatically, right? Like I said, I'm not sure this code is ever exercised anyhow and if it is, how often it has to go to lumi level. I can also re-write it to iterate through two lists and return on the first difference. But in any case, this is not code I'm changing from what's there already (at least not yet).

Second point, I don't think that making lumis another class will work because for sure there are places in the code that do things like ```lumis = myRun.lumis``` and then carry on as if lumis is a list. Could be even more refactoring.

vkuznet:
> I don't understand. Those lists for comparisons are not made in client code, they are in the Run class code and should be able to be cleaned up automatically, right? Like I said, I'm not sure this code is ever exercised anyhow and if it is, how often it has to go to lumi level. I can also re-write it to iterate through two lists and return on the first difference. But in any case, this is not code I'm changing from what's there already (at least not yet).

well, the code which will instantiate Run class is a client code, in other
words whoever who use Run class is a client. You don't need to iterate, we just
don't need to create new lists.

> 
> Second point, I don't think that making lumis another class will work because for sure there are places in the code that do things like ```lumis = myRun.lumis``` and then carry on as if lumis is a list. Could be even more refactoring.

here probably you'll need to change muRun.lumis to a method of a lumi class
which will return list of lumis.

I don't know who are using Run class and depending on use case it may worth
using either list or class implementation of lumis.

ericvaandering:
I guess I haven't been clear about where I am going with this code. I guess just wait until I have a real version of it and then we can see.

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
There is something goofy going on here. Jenkins seems to be complaining about code that does not exist and/or that I already fixed. Trying to reproduce exactly what Jenkins is doing with git doesn't help.

ericvaandering:
Going to try closing this an opening another.

ericvaandering:
Grrr.

ericvaandering:
test this please

ericvaandering:
@ticoann I think this is ready for review. It changes the internal format of the Run and hence the job report as well as the JSON format and the "python format". 

The one failure I cannot figure out yet. As far as I can tell, the code it complains about does not exist. I don't see failures in my VM.

src/python/WMCore/DataStructs/Run.py
```diff
@@ -1,120 +1,197 @@
...
+        return sorted(self.eventsPerLumi.keys())
+
+    @lumis.setter
+    def lumis(self, lumiList):
+        """
+        Setter to allow for replacement of the lumis with a list or list of tuples
+        """
+        self.eventsPerLumi = {}  # Remove existing dictionary
+        for lumi in lumiList:
+            if isinstance(lumi, (list, tuple)):
```
ticoann: are we acceptation both list of list and list of tuples? Is there a particular reason to support both?

ericvaandering: Yeah. Lots of existing code uses just lists to create a run, I think with code like `myRun.lumis = [1,2,4,6]` Using this interface lets us intercept that use case and treat it as if they'd written `myRun.lumis = [(1,100), (2,150), ... ]`

ericvaandering: Each of the three main commits preserves backwards compatibility with the existing code.

ericvaandering: Sorry, I missed your actual question. List of tuples vs. list of list. No real reason, I tried to use list of tuples, but there is no reason we can't use either so why not? Kind of following the duck-typing concept of python.

ticoann:
@ericvaandering, Eric

For following line,

https://github.com/dmwm/WMCore/blob/b27c386bcd45a2f9447cd585535c85eb7c2ae697/src/python/WMCore/FwkJobReport/Report.py#L87

```
instead of 
setattr(fileSection.runs, str(singleRun.run), singleRun.lumis)
->
setattr(fileSection.runs, str(singleRun.run), singleRun.eventsPerLumi)
```

I think this has to be changed to make archive format work. But this change might break other parts of the code whichever uses FWJR to get the lumi information.

ericvaandering:
Seems correct. Let me add it to my other PR to make sure things don't blow up.

ticoann:
test this please

ticoann:
test this please

ericvaandering:
@ticoann the baseline of the tests is still borked. I'm going to do this current run and if I don't solve it, I will set things to the way they were yesterday.  

Things should be ready for you to trigger new tests after you have lunch. :-)

ticoann:
Thanks Eric, I think 2 of the test is not related but wasn't sure about the one. 😄

ericvaandering:
I maintain that WMCore_t.WMBS_t.JobSplitting_t.LumiBased_t.LumiBasedTest:testB_NoRunNoFileSplitting  is claiming to fail on code that does not exist. I've never understood what's going on there.

PR #7730, opened 2017-03-16
Remove some tweaks for the relval agent deployment

### LINKED ISSUE #7722: New relval agent setting
Following up from https://github.com/dmwm/WMCore/issues/7715 we should set the relval agent limits to [60 min, N min] so as to have utilitarian jobs ask for 60min and processing jobs maxed to N min.
N could be the default setting IMO, which would change ~nothing to the current situation.
We can live with N=2200 min for the time being.

We should enforce that TimePerEvent is set to pseudo-reasonable values in the matrix to avoid having jobs requesting much less than what they need ; @fabozzi may you please follow up on this

amaltaro:
Fixes #7722


amaltaro:
Since it's only deployment, I'm merging this one.

PR #7731, opened 2017-03-16
Set Agent to DrainMode=True if exceeds DiskUseThreshold.

### LINKED ISSUE #7577: Automatically set the agent to Drain when diskUseThreshold is reached
AnalyticsDataCollector has this configuration:
config.AnalyticsDataCollector.diskUseThreshold = 85

which is used to trigger a warning in wmstats when the disk usage is >= than that value.

The idea is to use this value in order to automatically set the agent into Drain. Once that is done, these 3 components should be restarted to see these changes:
WorkQueueManager,AnalyticsDataCollector,AgentStatusWatcher

alexanderrichards:
This PR closes https://github.com/dmwm/WMCore/issues/7577

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -141,6 +141,7 @@ config.WorkQueueManager.queueParams["ParentQueueCouchUrl"] = "https://cmsweb.cer
 # this has to be unique for different work queue. This is just place holder
 config.WorkQueueManager.queueParams["QueueURL"] = "http://%s:5984" % (config.Agent.hostName)
 config.WorkQueueManager.queueParams["WorkPerCycle"] = 100  # don't pull more than this number of elements per cycle
+config.WorkQueueManager.queueParams["DrainMode"] = False
 
 config.component_("DBS3Upload")
 config.DBS3Upload.namespace = "WMComponent.DBS3Buffer.DBS3Upload"
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -141,7 +141,7 @@ config.WorkQueueManager.queueParams["ParentQueueCouchUrl"] = "https://cmsweb.cer
 # this has to be unique for different work queue. This is just place holder
 config.WorkQueueManager.queueParams["QueueURL"] = "http://%s:5984" % (config.Agent.hostName)
 config.WorkQueueManager.queueParams["WorkPerCycle"] = 100  # don't pull more than this number of elements per cycle
-config.WorkQueueManager.queueParams["drain_mode"] = False
+config.WorkQueueManager.queueParams["DrainMode"] = False
 
 config.component_("DBS3Upload")
 config.DBS3Upload.namespace = "WMComponent.DBS3Buffer.DBS3Upload"
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -141,7 +141,7 @@ config.WorkQueueManager.queueParams["ParentQueueCouchUrl"] = "https://cmsweb.cer
 # this has to be unique for different work queue. This is just place holder
 config.WorkQueueManager.queueParams["QueueURL"] = "http://%s:5984" % (config.Agent.hostName)
 config.WorkQueueManager.queueParams["WorkPerCycle"] = 100  # don't pull more than this number of elements per cycle
-config.WorkQueueManager.queueParams["DrainMode"] = False
+config.WorkQueueManager.queueParams["drain_mode"] = False
 
 config.component_("DBS3Upload")
 config.DBS3Upload.namespace = "WMComponent.DBS3Buffer.DBS3Upload"
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -7,6 +7,7 @@ Perform general agent monitoring, like:
 """
 __all__ = []
 
+import os
 import threading
 import logging
 import time
@@ -200,6 +201,9 @@ class AgentStatusPoller(BaseWorkerThread):
             if float(disk['percent'].strip('%')) >= diskUseThreshold and \
                             disk['mounted'] not in self.config.AnalyticsDataCollector.ignoreDisk:
                 agentInfo['disk_warning'].append(disk)
+                agentInfo['drain_mode'] = True
+                self.config.WorkQueueManager.queueParams['DrainMode'] = True
+                os.system("$manage execute-agent wmcoreD --restart --components=WorkQueueManager,AnalyticsDataCollector,AgentStatusWatcher")
 
         # Couch process warning
         couchProc = numberCouchProcess()
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -38,6 +38,12 @@ class AgentStatusPoller(BaseWorkerThread):
         BaseWorkerThread.__init__(self)
         # set the workqueue service for REST call
         self.config = config
+        self.config_path = os.path.join(os.environ.get('config', ''), 'config.py')
+        if not os.path.exists(config_path):
+            self.config_path = '/data/srv/wmagent/current/config/wmagent/config.py'
+            if not os.path.exists(config_path):
+                self.config_path = None
+
         # need to get campaign, user, owner info
         self.agentInfo = initAgentInfo(self.config)
         self.summaryLevel = config.AnalyticsDataCollector.summaryLevel
@@ -107,12 +113,9 @@ class AgentStatusPoller(BaseWorkerThread):
         get information from wmbs, workqueue and local couch
         """
         try:
-            config_path = os.path.join(os.environ.get('config', ''), 'config.py')
-            if not os.path.exists(config_path):
-                config_path = '/data/srv/wmagent/current/config/wmagent/config.py'
-            if os.path.exists(config_path):
-                self.config = loadConfigurationFile(config_path)
-                self.config.path = config_path
+            if self.config_path is not None:
+                self.config = loadConfigurationFile(self.config_path)
+
             agentInfo = self.collectAgentInfo()
             self.checkProxyLifetime(agentInfo)
 
@@ -208,8 +211,8 @@ class AgentStatusPoller(BaseWorkerThread):
                             disk['mounted'] not in self.config.AnalyticsDataCollector.ignoreDisk:
                 agentInfo['disk_warning'].append(disk)
                 self.config.WorkQueueManager.queueParams['DrainMode'] = True
-                if hasattr(self.config, 'path'):
-                    saveConfigurationFile(self.config, self.config.path)
+                if self.config_path is not None and not isDrainMode(self.config):
+                    saveConfigurationFile(self.config, self.config_path)
 
         # Couch process warning
         couchProc = numberCouchProcess()
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -23,7 +23,7 @@ from WMComponent.AnalyticsDataCollector.DataCollectAPI import WMAgentDBData, \
     diskUse, numberCouchProcess
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue as WorkQueueDS
 from WMCore.WorkQueue.DataStructs.WorkQueueElementsSummary import getGlobalSiteStatusSummary
-
+from WMCore.Configuration import loadConfigurationFile, saveConfigurationFile
 
 class AgentStatusPoller(BaseWorkerThread):
     """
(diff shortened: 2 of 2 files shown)


src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -168,6 +169,9 @@ def collectAgentInfo(self):
             if float(disk['percent'].strip('%')) >= diskUseThreshold and \
                             disk['mounted'] not in self.config.AnalyticsDataCollector.ignoreDisk:
                 agentInfo['disk_warning'].append(disk)
+                agentInfo['drain_mode'] = True
+                self.config.WorkQueueManager.queueParams['DrainMode'] = True
+                os.system("$manage execute-agent wmcoreD --restart --components=WorkQueueManager,AnalyticsDataCollector,AgentStatusWatcher")
```
alexanderrichards: @ticoann @amaltaro 
I'm not sure this is the best place to do the restart of these services?

Also I wanted to use the functions directly from the wmcoreD script however it is not in a module dir so not importable and is written with global variables which are populated from command line args and as such it's not easy to use the code directly. I don't know if you would have preferred spawning a subprocess with the subprocess module, or using a separate thread or an exec (if we want to replace the service anyway) for this or if a simple `os.system` is sufficient? I guess it really depends on the behaviour your looking for. I'll let you decide.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -168,6 +169,9 @@ def collectAgentInfo(self):
             if float(disk['percent'].strip('%')) >= diskUseThreshold and \
                             disk['mounted'] not in self.config.AnalyticsDataCollector.ignoreDisk:
                 agentInfo['disk_warning'].append(disk)
+                agentInfo['drain_mode'] = True
```
amaltaro: This line is not needed.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -168,6 +169,9 @@ def collectAgentInfo(self):
...
+                self.config.WorkQueueManager.queueParams['DrainMode'] = True
```
amaltaro: As mentioned in the meeting, can you check whether this configuration change gets automatically persisted in disk (config/wmagent/config.py)?

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
amaltaro: Giving it some thought, I think it would be useful to have a ManageAgent or something like that under
https://github.com/dmwm/WMCore/tree/master/src/python/WMCore/Agent

with basically the same functionality as wmcoreD. I'm sure it would be used for other purposes once we get it.
@ticoann what do you think?

amaltaro (changes requested):
Minor things to update and check. Let's wait to see what Seangchan thinks about moving the wmcoreD functionality to somewhere that we can import it.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
alexanderrichards: So from the meeting we agree we will instead try to reread the config each polling cycle instead of actually restarting the components, a la...

https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/AgentStatusWatcher/ResourceControlUpdater.py#L89

alexanderrichards: hmmm it looks like we already update the config each time but then modify it to the original config which we get at instantiation time `self.agentInfo` [AgentStatusPoller#L148](https://github.com/alexanderrichards/WMCore/blob/80448e32ccf09c57d017fa7550fbb16435649d07/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py#L148). I wonder if it's sufficient to simply remove the last line below. Why do we update this based on the config in the `self.agentInfo`?

```python
        logging.info("Getting agent info ...")
        agentInfo = self.wmagentDB.getComponentStatus(self.config)
        agentInfo.update(self.agentInfo)
```


Actually it looks like this `self.agentInfo` is only a couple of fields anyway [DataCollectAPI#L443](https://github.com/alexanderrichards/WMCore/blob/80448e32ccf09c57d017fa7550fbb16435649d07/src/python/WMComponent/AnalyticsDataCollector/DataCollectAPI.py#L443). Does this mean that we are already reloading the config?

amaltaro: Sorry for the belated answer. `self.agentInfo` just adds a couple of key/value pairs which are later on uploaded to wmstats.

And looking at that code, it doesn't really update the configuration every cycle, since it uses the same config object that is passed when starting that thread :-(

We could use this function https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Configuration.py#L573 for reloading the agent config, but then how do we know the filename path? We can discuss it in the meeting.

ticoann:
@alexanderrichards, @amaltaro 

I was thinking more about this case. And I think we have clarify one thing first.

1. What make the agent go in drain? I guess there are 2 cases. 
    a. if certain conditions meet (disk usage) in the system by automatic detection
    b. manually set by ops. - currently using config file 

2. If we only do a. we can just remove config completely. (But I think it is not always the case, Ops might want to drain agent even the condition is not met and (maybe let it run) even if condition is met.

If there is conflict between two options we need to have specific rules.
i.e. which ever says drain = True, take the effect.  
So config set to False for drain but condition is met than agent will go automatically in drain mode. 
Also config is set to True and condition doesn't meet it will still go to drain mode.

ericvaandering:
test this please

alexanderrichards:
> We could use this function https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Configuration.py#L573 for reloading the agent config, but then how do we know the filename path? We can discuss it in the meeting.

looks like the configuration object doesn't know the filename/path. Do we know a standardised location that it should be? even if it's relative to something else?

amaltaro:
Yes, the standard location for all the agents (but T0) is:
/data/srv/wmagent/current/config/wmagent/config.py

maybe we start with this "default" location, you run some tests and later on we try to fetch this file location in a better way. What do you think?

amaltaro:
Alex, we have this line in the environment script (env.sh)
`export config=/data/srv/wmagent/current/config/wmagent`

How about you check if this $config var is in the environment, then check if the file exists and if so load it. Otherwise, get the default location mentioned above.

alexanderrichards:
sounds like a good idea

alexanderrichards:
> Alex, we have this line in the environment script (env.sh)
> export config=/data/srv/wmagent/current/config/wmagent

This config var is set to the directory as you said above or the actual config.py file itself?

amaltaro:
It's set to the directory because we use it again in the next line:
```
export config=/data/srv/wmagent/current/config/wmagent
export manage=$config/manage
```

though to be honest, that's a bad environment variable name. Maybe in the near future we should call it WMA_CONFIG or so. Besides $manage, I don't know if there is anything else using that var...

etc/WMAgentConfig.py
```diff
@@ -141,6 +141,7 @@
 # this has to be unique for different work queue. This is just place holder
 config.WorkQueueManager.queueParams["QueueURL"] = "http://%s:5984" % (config.Agent.hostName)
 config.WorkQueueManager.queueParams["WorkPerCycle"] = 100  # don't pull more than this number of elements per cycle
+config.WorkQueueManager.queueParams["drain_mode"] = False
```
amaltaro: It has to be named `DrainMode`, for the agent config.
 `drain_mode` is the key name in wmstats.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -200,6 +206,8 @@ def collectAgentInfo(self):
             if float(disk['percent'].strip('%')) >= diskUseThreshold and \
                             disk['mounted'] not in self.config.AnalyticsDataCollector.ignoreDisk:
                 agentInfo['disk_warning'].append(disk)
+                agentInfo['drain_mode'] = True
+                self.config.WorkQueueManager.queueParams['drain_mode'] = True
```
amaltaro: same here

amaltaro: same here

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -106,6 +107,10 @@ def algorithm(self, parameters):
         get information from wmbs, workqueue and local couch
         """
         try:
+            config_path = os.path.join(os.environ.get('config', ''), 'config.py')
+            if not os.path.exists(config_path):
+                config_path = '/data/srv/wmagent/current/config/wmagent/config.py'
```
amaltaro: this may be a problem for jenkins. Maybe we should check if the default path exists as well. If it does NOT, then we don't even load the config and we don't save.
Moreover, when you save the config you are also persisting it on disk (I believe...). So I'd only save it if it really changed.

amaltaro (changes requested):
I know you haven't asked for a review yet :) but since I saw almost 200 unit tests failing, I thought it wouldn't hurt having a look at it.
There are a few things that need to be fixed.

etc/WMAgentConfig.py
alexanderrichards: ah ha, I had that originally but changed it with this commit as I thought I only saw drain_mode. I'll change it back.

alexanderrichards:
@amaltaro what do you think of the latest changes?

amaltaro (changes requested):
Sorry, Alex, I have one more suggestion :(

Can you find the config.py file in the init method instead and set an instance variable with the actual file (or None if file does not exist). Bsically moving L110-L115 into the init method.

Then when you swap the flag and save the configuration file, just check whether self.config_path is not None.

One more comment, we should also check what 's the value of DrainMode in the config, in case it's already True, there is nothing to be done.

amaltaro:
@alexanderrichards please let me know if my comments are not clear.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -106,6 +107,12 @@ def algorithm(self, parameters):
         get information from wmbs, workqueue and local couch
         """
         try:
+            config_path = os.path.join(os.environ.get('config', ''), 'config.py')
```
amaltaro: Alex, can you also rename this env var to WMA_CONFIG please? We'll have to update the env script `/data/admin/wmagent/env.sh`.

amaltaro: Alex, scratch this comment. Let us keep `config` and that's it.

alexanderrichards: are you sure?

amaltaro: I've changed my mind 5 times this morning :) I leave it up to you, surprise me LoL

alexanderrichards: one surprise coming up...:wink:

alexanderrichards:
@amaltaro I think I have now implemented the requirements that you set out. Lets wait and see what the testing shows up but feel free to review and comment in the meantime.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -200,6 +210,9 @@ def collectAgentInfo(self):
             if float(disk['percent'].strip('%')) >= diskUseThreshold and \
                             disk['mounted'] not in self.config.AnalyticsDataCollector.ignoreDisk:
                 agentInfo['disk_warning'].append(disk)
+                self.config.WorkQueueManager.queueParams['DrainMode'] = True
+                if self.config_path is not None and not isDrainMode(self.config):
```
amaltaro: If we have more than 1 disk warning, then we'd be saving the config > 1 times. I suggest keeping this if after this for loop. See changes in the file I'm going to share.

amaltaro (changes requested):
Alex, there are good and bad news. Starting wit the good one, we're almost there, though there is at least one problem and a few other changes needed.
I took your changes and started testing/updating your patch, my final file (based on 1.1.2 cycle) follows:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/PR_7731/AgentStatusPoller.py

can you diff and apply those changes to this PR please? You can also squash all these commits.

Now the bad news is that... I was almost forgetting that we also need to reload the agent configuration on the WorkQueueManager component. Since we have to play again with the configuration location and etc, how about we lookup the config env var here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Agent/Configuration.py#L49
and apply the default path in case it fails to find that var.

It'll add an overhead since this function calls the other  `loadConfigurationFile` method, but at least it's under the Agent package and I *think* it's only used by agents.

Then the second change is on the WorkQueue code, I think it's easier if I take care of that in another PR.

Please let us know if something is not clear or doesn't look right.

ticoann:
replaced by #8011

PR #7734, opened 2017-03-17
Completely replace CouchURL by ConfigCacheUrl

### LINKED ISSUE #7732: Converge to a single ConfigCacheUrl parameter
Right now we look at:
1. ConfigCacheUrl
2. CouchURL
however, user can define ConfigCacheURL, which is meaningless since it's not supported in the spec files.

I think we enforce only `ConfigCacheUrl` and completely remove any support to `CouchURL` from the code. @ticoann @hufnagel is that Ok?

BTW, this issue should be fixed in line with https://github.com/dmwm/WMCore/issues/7133

amaltaro:
Fixes #7732 

Since this is possible a breaking change, we need to:
a) review/update the CompOps scripts
b) talk to/announce this change to McM
c) talk to/announce this change to RelVal managers

I also had to remove `getWorkloadArgumentsWithReqMgr` and merge those arguments into `getWorkloadARguments`. Then when a spec doesn't want to use them (T0), it should override it. It's also part of the create/assign argument separation that I'm going to work on after this.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/DQMHarvest.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/DQMHarvest.py
@@ -56,17 +56,14 @@ class DQMHarvestWorkloadFactory(StdBase):
             msg = "DQMConfigCacheID parameter has not been provided in the request"
             raise WMSpecFactoryException(message=msg)
 
-        couchUrl = schema.get("ConfigCacheUrl", None)
         self.validateConfigCacheExists(configID=schema["DQMConfigCacheID"],
-                                       couchURL=couchUrl,
+                                       configCacheUrl=schema['ConfigCacheUrl'],
                                        couchDBName=schema["CouchDBName"],
                                        getOutputModules=False)
 
     @staticmethod
     def getWorkloadArguments():
         baseArgs = StdBase.getWorkloadArguments()
-        reqMgrArgs = StdBase.getWorkloadArgumentsWithReqMgr()
-        baseArgs.update(reqMgrArgs)
         specArgs = {"RequestType": {"default": "DQMHarvest"},
                     "InputDataset": {"default": None, "optional": False,
                                      "validate": dataset},
@@ -150,10 +147,7 @@ class DQMHarvestWorkloadFactory(StdBase):
                                           dqmHarvestUnit=dqmHarvestUnit)
 
         if self.dqmConfigCacheID is not None:
-            if getattr(self, "configCacheUrl", None) is not None:
-                harvestTaskCmsswHelper.setConfigCache(self.configCacheUrl, self.dqmConfigCacheID, self.couchDBName)
-            else:
-                harvestTaskCmsswHelper.setConfigCache(self.couchURL, self.dqmConfigCacheID, self.couchDBName)
+            harvestTaskCmsswHelper.setConfigCache(self.configCacheUrl, self.dqmConfigCacheID, self.couchDBName)
             harvestTaskCmsswHelper.setDatasetName(self.inputDataset)
         else:
             scenarioArgs = {'globalTag': self.globalTag,
--- a/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
@@ -38,8 +38,6 @@ class DataProcessing(StdBase):
     @staticmethod
     def getWorkloadArguments():
         baseArgs = StdBase.getWorkloadArguments()
-        reqMgrArgs = StdBase.getWorkloadArgumentsWithReqMgr()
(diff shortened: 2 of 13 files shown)


vlimant:
@anorkus for McM
@fabozzi for relval

amaltaro:
Thanks for initiating it, Jean-Roch.
Let us know if help is needed with the scripts or anywhere else.

amaltaro:
test this please

amaltaro:
@hufnagel I guess you'll have to test this PR in a replay as well.
In order to have a better coverage on these spec changes, we could actually migrate those other two T0 specs in
https://github.com/dmwm/T0/tree/master/src/python/T0/WMSpec/StdSpecs

to WMCore. How about that? PromptReco is already here anyways. BTW, I suggest creating some unit tests for them.

hufnagel:
These changes should be irrelevant for the Tier0, we don't use Couch in Express and Repack specs. Will be tested whenever I am trying to build the next T0 release, will make any necessary changes then.

There are unit tests for these specs, but Tier0 unit tests are more integrated, there is a big RunConfig unit tests that tests everything (with an actual real Tier0 config and real Express and Repack settings).

If we decide to move them to WMCore (and I admit there is an argument to be made to do so) I agree that they should get their own more specialized unit test.

amaltaro:
> These changes should be irrelevant for the Tier0, we don't use Couch in Express and Repack specs. 

I think we do the spec validation for any request spec and, if these specs are not overwritting the standard StdBase, then I think it will cause you trouble.

Ok, it's good to know you have something testing it besides a replay. I just want to avoid giving you bad surprises. BTW, stay tuned, more spec changes are on the way :)

anorkus:
Hi,

mcm and relvals were/are using both params, so transition is easy. To clean-up the code i just remove the "CouchURL" parameter from injection dictionary:
https://github.com/cms-sw/cmssw/pull/17988
and
https://github.com/cms-PdmV/wmcontrol/commit/8a4dbfab05aa1a0eafc36d66dbccc38b22266740

amaltaro:
Thank you, Antanas. Then the only part missing is with the CompOps scripts, but I think @vlimant and @prozober are taking care of that already.

amaltaro:
@ticoann tests in my VM look fine. Please review and merge.

ticoann:
Hi Alan, Sorry for the late review. 
There are 3 other places  using CouchURL

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/Request.py#L66

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L539

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L587

It looks like those (CouchURL and ConfigCacheUrl) have 2 different meanings. So technically we have keep them both. 😄 

However, I think current patch is fine in a sense it doesn't mix up the two parameters. And CouchURL is always set by system not by user (but we still keep both in the document)

There is one thing I didn't quite understand though. Why did you remove getWorkloadArgumentsWithReqMgr? I think that sill need to be exist to validate even though those are automatically generated by request manager. Also that shows the difference what parameters will be generated between T0 and Production/Processing workflows.

amaltaro:
Nice catch! I somehow overlooked those 3 places (how come?!?!)

About getWorkloadArgumentsWithReqMgr removal, I actually merged those arguments into getWorkloadArguments because some specs (if my memory doesn't fail, the *Chain specs) were actually not loading those ReqMgr args and that was making the ConfigCacheUrl fail the validate function.

Dirk already said it's not a problem for T0 :-) but if it does cause issues, then we have to override those parameters in the T0 specs.

I'm updating this patch in a bit.

ticoann:
Hi Alan, I wasn't saying Tier0 will fail with this patch since there is  default value for parameters. :-) But I is setting some values which is never needed. 
Anyway, I don't think you need to change anything on those three CouchURL code missed. Current patch looks fine.

amaltaro:
Seangchan, as we discussed, I still have to add `CouchURL` to StdBase, since that's going to be used to say where the spec is located (with default value and never overriden).
The trick part is going to find where it should be cmsweb, cmsweb-testbed, or our own VM hostname... Looking at it.

ticoann:
Alan, it is getting from the config. 

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/Request.py#L66

On a second thought I think we can skip spec since this won't be user defined, even if it is defined it will be overwritten. (Maybe just add the validation somehow)

amaltaro:
Thank you very much for the review, Seangchan.
I've just updated StdBase.py in the first commit. Can you check it again please?

Once tests are done, feel free to merge it.

ticoann (approved):

PR #7735, opened 2017-03-18
remove WMBSService from WMCore and WMAgent

ticoann:
Need to test. Also need to remove "WMBS_SERVICE_PORT" from T0/manage wmagent/manage script


ticoann:
@emaszs, Emilis, crabserver doesn't need HTTPFrontEnd dependency anymore, right? I couldn't find the code using that.

https://github.com/dmwm/WMCore/pull/7735/files#diff-d073738f4bf39da0585f5bb9c5f499efL115

ticoann:
@hufnagel,  do you need to get proxy from WMBSService config?

https://github.com/dmwm/T0/blob/d7b772a7425e06870f775f2bfb482bafed7d7bde/test/python/T0_t/WMBS_t/RunConfig_t/RunConfig_t.py#L53

hufnagel:
Latest Tier0 release has a patch that removes WMBSService from the agent configuration. We don't need it.

emaszs:
@ticoann Yeah, looks like we don't use it in CRAB anymore now that we've switched to the TagCollector service.

ticoann:
test this please

PR #7740, opened 2017-03-21
Explicitly add FQAN for X509 user proxy for job submission

amaltaro:
There is a random behaviour going on on the global pool where some jobs don't have this `x509userproxyfirstfqan` ad defined, which causes problems at RALPP.

Diego and I failed to find the issue and/or a pattern to what's going on. So the solution is to explicitly set it in the submitter plugin. Still to be tested

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -113,6 +113,7 @@ class SimpleCondorPlugin(BasePlugin):
         proxy = Proxy({'logger': myThread.logger})
         self.x509userproxy = proxy.getProxyFilename()
         self.x509userproxysubject = proxy.getSubject()
+        self.x509userproxyfqan = proxy.getAttributeFromProxy(self.x509userproxy)
 
         return
 
@@ -466,6 +467,7 @@ class SimpleCondorPlugin(BasePlugin):
 
         ad['x509userproxy'] = self.x509userproxy
         ad['x509userproxysubject'] = self.x509userproxysubject
+        ad['x509userproxyfirstfqan'] = self.x509userproxyfqan
 
         ad['Rank'] = 0.0
         ad['TransferIn'] = False


vlimant:
wait, can we have @bbockelm 2 cents on this ?
It does not make sense to add a parameter just because *one site* is using it. We will disable the site in production until further notice, but no need to set that classas ; there is enough junk already

amaltaro:
Ok, I tested this change in vocms0116 and i see the correct classad in the new jobs.
Since we want more feedback on this issue, I have removed the patch and restarted JobSubmitter once again.

It would be great if someone could follow this issue up on the elog @ddavila0 created in the previous week before people start shooting in the dark.

bbockelm:
Not much mystery here:
- Until latest HTCondor stable series (SI is behind), jobs submitted via python bindings don't get X509 attributes populated.
- `SimpleCondorPlugin.py` populates a few attribuets, but not this one.
- When jobs are started by the `condor_shadow`, it'll update the attributes.

So, this attribute is missing for all jobs on agents running `SimpleCondorPlugin` that have `NumShadowStarts=?=0`.

bbockelm (approved):
Looks good to me.

All the `x509userproxy` stuff can be ripped out once SI updates to the latest stable series, I believe.

vlimant:
ok @bbockelm, thanks. But we should not add this classad for all jobs if we do not need it, except for one site : unless that one site will be all sites shortly

bbockelm:
Nope - we shouldn't add it only for one site, we should add it for consistency between the jobs generated by this plugin and from the other means of submission.

Basically, we want the jobs to "look like" jobs from `condor_submit` and what will be generated when we update the HTCondor client.

vlimant:
OK then. sorry for the noise. thanks for the confirmation

amaltaro:
Thanks for the review, Brian. Would you or Diego @ddavila0  know what's the latest stable HTCondor release that will auto-magically add these x509 attributes? I assume we only have to keep `x509userproxy` when that happens?

bbockelm:
Automatic insertion of these attributes started in 8.5.8.  So, any release since December 2016 should be OK.

amaltaro:
@ticoann can you merge it please

PR #7744, opened 2017-03-21
fixes acdc clone assignment overwrite

### LINKED ISSUE #7692: Values inheritance for chained requests
@ticoann Seangchan, and there is yet another case that we didn't foresee :( Taking this request as example:
https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=amaltaro_TaskChain_MC_HG1703_Validation_170301_142616_3204

its ProcessingString was overwritten during assignment (that's why we have it in the top level dict as a dict object). The inner ProcessingString values are not overwritten in the couch workload doc though, see for more details: https://github.com/dmwm/WMCore/issues/6881

Then the problem is, the values - correctly - inherited are from the inner dicts instead of the dict at the top level.

I don't think it will be a real problem for Production, since the output dataset name  is defined at creation time by McM, so there shouldn't be any change between the values set by McM (at creation) and Unified (at assignment).
For the next cmsweb cycle, I suggest we simply clone the original request into a Resubmission spec and get over all this painful inheritance business.

ticoann:
fixes #7692

### DIFF
--- a/src/couchapps/ReqMgr/updates/updaterequest.js
+++ b/src/couchapps/ReqMgr/updates/updaterequest.js
@@ -29,13 +29,26 @@ function(doc, req) {
     // req.query is dictionary fields into the 
     // CMSCouch.Database.updateDocument() method, which is a dictionary
     function isEmpty(obj) {
-	    for(var prop in obj) {
-	        if(obj.hasOwnProperty(prop))
-	            return false;
-	    }
-    	return true;
-	}
-	
+        for(var prop in obj) {
+            if(obj.hasOwnProperty(prop))
+                return false;
+        }
+        return true;
+    }
+
+    function updateTaskStepChain(chainType, prop, value) {
+        if (doc[chainType + "Chain"]) {
+            var numChain = doc[chainType + "Chain"];
+            for (var i=1; i <= numChain; i++) {
+                if (doc[chainType + i] && doc[chainType + i][prop] && 
+                    value.hasOwnProperty(doc[chainType + i][chainType + "Name"])) {
+                    
+                    doc[chainType + i][prop] = value[doc[chainType + i][chainType + "Name"]];
+                }
+            }
+        }
+    }
+
     // req.query is dictionary fields into the 
     // CMSCouch.Database.updateDocument() method, which is a dictionary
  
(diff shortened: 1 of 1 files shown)


ticoann:
Still needs to test

ticoann:
tested in vm (reqmgr2-dev)

src/couchapps/ReqMgr/updates/updaterequest.js
```diff
@@ -79,11 +92,14 @@ function(doc, req) {
...
     	}
-       
+       	
         // If key is RequestStatus, also update the transition
         if (key == "RequestStatus") {
         	updateTransition();
         }
+        
+        updateTaskStepChain("Task", key, newValues[key]);
+        updateTaskStepChain("Step", key, newValues[key]);
```
amaltaro: Instead of calling this function twice, how about we iterate over ["Task", "Step"] inside the function itself? It should have a slightly better performance.

src/couchapps/ReqMgr/updates/updaterequest.js
```diff
@@ -36,6 +36,19 @@ function(doc, req) {
     	return true;
 	}
 	
+	function updateTaskStepChain(chainType, prop, value) {
+		if (doc[chainType + "Chain"]) {
+			var numChain = doc[chainType + "Chain"];
```
amaltaro: Don't we have to first test whether this `doc` contains this key (TaskChain or StepChain)? If not, just bail out.

amaltaro: BTW, this whole function is miss-indented. And so is L88-90. Maybe you can just reformat the whole file to make it more readable :)

amaltaro (changes requested):
Besides the possible undefined key, the logic seems correct.

src/couchapps/ReqMgr/updates/updaterequest.js
ticoann: In javascript we don't have to check, if it doesn't contain the key it returns "undefined" which is create as false.
My editor automatically sets tab to 4 space but I am not sure why it is more than that. I will fix the indentation.

src/couchapps/ReqMgr/updates/updaterequest.js
ticoann: It is a little ugly to iterate over in javascript. :-) I am not sure how much we will save with less function call. But as far as I know, modern javascript is pretty good at handling that so we probably shouldn't concern that much anymore. (Especially if the code is less cleaner)

```
a = ["Task", "Step"]

for (var i in a) {
    var chainType = a[i];
}

```

ticoann:
Alan, I updated code (fixes indentation) but keep the function as it is. 😄

amaltaro:
You can do better than that :-) You missed the lines I mentioned before
https://github.com/ticoann/WMCore/blob/253d291e20881e6a72964cd54407ec05575a8f2f/src/couchapps/ReqMgr/updates/updaterequest.js#L88-L90

Then your logic is completely correct and this PR is good to go ;)

ticoann:
Are you an "indentation" cop? :-) Now it should all correct. Thanks

amaltaro:
when your javascript skills are not that good, indentation matters :-)

PR #7745, opened 2017-03-22
Separate creation and assignment spec parameters

### LINKED ISSUE #7133: Make a list of accepted parameters for creation and another for assignment

amaltaro:
Fixes #7133

In short, this PR will enforce creation-only arguments to be set in the request when it's being created. IT will enforce assignment-only arguments when a request is being assigned. It also enforces/validate the task/step level arguments. Anything that's NOT defined in the specs, will fail the write operation.

And a more detailed list of changes is:
* specific methods for retrieving: creation, assignment and task/step level list of arguments and their definition;
* validation of unknown args, if there are any, then request creation will fail;
* RequestDN cannot be null and it's now mandatory
* Added Comments and SubRequestType to the list of allowed args for creation
* Fixed TaskChain tasks inheritance
* specific method to get test assignment default arguments
* allow merge settings during creation and assignment (ideally it should be used only at assignment (To problem?))
* Removed SoftwareVersions and InputDatasets args (were being added by ReqMgr2)
* Moved most of the validation to the spec level (like SiteWhitelist and Team)
* TimePerEvent and SizePerEvent are optional in a Task (shouldn't it be required?)
* FilterEfficiency only supported for MonteCarlo and chained requests (first generator task)
* do not over-validate chain specs/args
* Use mocked data for StepChain and DQMHarvest; a bunch of unit tests fixed
* created method to return assignment test arguments

TODO:
* remove commented out code from src/python/WMCore/ReqMgr/DataStructs/Request.py
* remove redundant default definition from the argument definition
* a couple of FIXME in StdBase.py
* investigate whether this method is actually ever executed: https://github.com/dmwm/WMCore/pull/7745/files#diff-0eae8d98a110117b2919a6fd50deeb7fL1749
* apply pep8 fixes


ticoann:
> *) if user provide any argument that is not defined in these methods (depending on the operation being done), the operation will fail with a WMSpec exception

For this part, I think we have to discuss with ops and make some parameters available. I think there are putting some new ones ie. 'SubRequestType'. So if necessary we can add that.
But I completely agreed we need to have control what parameters are allowed.

amaltaro:
This `SubRequestType` is already covered in this PR, and so is `Comments` that we use in our templates :-)
@vlimant is Unified adding any ad-hoc parameter to the requests dict?

@ticoann what concerns me the most is the T0 use cases for a) not using ReqMgr and b) setting ALL or most of the creation + assignment parameters at *creation* only, since I guess T0 doesn;t really have an assignment where the request spec is updated. @hufnagel feel free to tell me that I'm wrong :)

ticoann:
I am not sure how many parameters only needs to defined in assignment but not creation. (Not just T0 but also Production/Processing).
When clone and ACDC is made, most of inherited the parameters are set in creation, right?

amaltaro:
Clone should follow the same path as a brand new request.
ACDC may be a bit trickier, but one step at a time :) (and we want to do some rework on ACDCs anyways)

amaltaro:
@ericvaandering just to avoid creating more changes than needed. How did you dump data to this json file
https://github.com/dmwm/WMCore/blob/master/test/data/Mock/DBSMockData.json
?

I haven't used anything special besides indent
`json.dump(data, jo, indent=1)`

ericvaandering:
It's made by this piece of code: https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/Emulators/DBSClient/MakeDBSMockFiles.py

The way I keep the data minimal and sorted to minimize git diffs is this:

```
   with open(outFilename, 'w') as mockData:
        json.dump(lookup, mockData, indent=1, separators=(',', ': '), sort_keys=True)
```

amaltaro:
Thanks, Eric. I'm updating that json file. And I guess I also have to update MakeDBSMockFiles.py to make sure those data will not be removed from the json file. Now checking what were the data and which calls :-(

ericvaandering:
Definitely fix this one:

```
Unused GlobalParams imported from WMQuality.Emulators.DataBlockGenerator.Globals
```

That's code I'm trying to kill off.

amaltaro:
Sure. I'm saving these pep8 fixes and improvements for my last commit :-)
There are still some stuff to figure out though.

amaltaro:
There is still this template `TaskChain_MCRecyc.json` which yields a None AcqEra during assignment and fails the validation. And the fight continues...

amaltaro:
test this please

amaltaro:
you must be kidding!!!

amaltaro:
test this please

amaltaro:
And finally all my tests are working. 
Since GH is having a hard time loading the suggested reviewers, @ticoann please review. @ericvaandering and @hufnagel your input is also very welcome. (sorry for the very large code review)

Note: make sure to read the first comment in this PR, just to have a better background of the current changes and small things still to be done.

amaltaro:
Puzzled! How come my latest commit didn't break any unit tests? :-)
@ticoann as we discussed yesterday, you can find Resubmission changes in my latest commit. In summary they are:
* accept any create/assign arguments during creation time of a Resubmission request
* any other argument is removed (still using that static list of arguments)
* then we skip the arguments validation (type, optional/non-optional, null, validation function), since most of the arguments already come from the original request.

If we want, I think we can perform an ad-hoc arguments validation. I'm just not sure how to handle conflicting definitions between create and assign then.
Please have a look and share your thoughts.

amaltaro:
@ticoann can you please review it once again? I'd like to have it in tomorrow's testbed such that we have enough time to fix those bugs created by it :-)
Just don't merge it yet, allow me squash and get rid of a couple of comments.

ticoann:
Hi Alan, I think patch looks good. I still have some doubt on Resubmission validation but I might be wrong on those comment. If basic test ran without problem I would suggest to include in the tomorrow testbed deployment and do further tests. 
Thanks,

ticoann:
test this please

amaltaro:
Seangchan, can you remind me what are those concerns again? Or we can chat tomorrow.
On my way to rebase it now (I wish I could go back to the history of commits/changes after rebasing my branch :))

amaltaro:
This was a tense rebase with a few conflicts. I hope no mistakes were made during that process.
@ticoann once it passes unit tests, feel free to merge. If we have problems, then I can possible look at it over the night or tomorrow morning.

amaltaro:
test this please

ticoann (approved):

src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
```diff
@@ -48,23 +50,27 @@ def buildWorkload(self, originalRequestURL, arguments):
...
-        #TODO remove the None case when reqmgr is retired
-        return self.buildWorkload(arguments.get("OriginalRequestCouchURL", None), arguments)
+        return self.buildWorkload(arguments)
 
     @staticmethod
-    def getWorkloadArguments():
+    def getWorkloadCreateArgs():
+        baseArgs = DataProcessing.getWorkloadCreateArgs()
         specArgs = {"RequestType" : {"default" : "Resubmission"},
+                    "OriginalRequestName": {"null": False},
```
amaltaro: You're probably right. I have to check why it didn't complain in my VM

ticoann:
Hi Alan, too late. 😄 I made a pull request for the testbed. I think we can still run the test for other stuff in test bed. and update this later if it causes the problem. But feel free to make a new pull request if you think this will cause the problem for sure.

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -143,17 +144,27 @@ def validate_request_create_args(request_args, config, reqmgr_db_service, *args,
...
     specClass = loadSpecClassByType(request_args["RequestType"])
-    setArgumentsWithDefault(request_args, specClass.getWorkloadArguments())
 
-    # get the spec type and validate arguments
-    spec = loadSpecByType(request_args["RequestType"])
     if request_args["RequestType"] == "Resubmission":
-        initialize_resubmission(request_args, config, reqmgr_db_service)
+        # do not set default values for Resubmission since it will be inherited from parent
+        # both create & assign args are accepted for Resubmission creation
+        acceptedArgs = specClass.getWorkloadCreateArgs()
```
amaltaro: And this makes sense too. We can load the original spec definition and then update it with the resubmission one.
I'll make these changes today.

amaltaro:
Thanks for the review, Seangchan. I'll make the changes and test it with the new testbed RPMs later today.

ticoann:
Thanks you very much, Alan.

PR #7748, opened 2017-03-22
Generate WMArchive report with new key eventsPerLumi

ericvaandering:
@ticoann @vkuznet @yuyiguo 

I think this does what we discussed this morning. Add a 2nd line if the data is available. Is this what you need? Do I have to update the schema somewhere too? Would it be better if I filled eventsPerLumi with Nones in we don't have the info rather than leaving the value out of the JSON completely?

### DIFF
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -125,7 +125,18 @@ def combineDataset(dataset):
     return dataset
 
 def changeRunStruct(runDict):
-    return [{"runNumber": int(run), "lumis": lumis}  for run, lumis in runDict.items()]
+    runList = []
+    for run, lumis in runDict.iteritems():
+        singleRun = {"runNumber": int(run)}
+        if isinstance(lumis, dict):
+            # In this case, lumis is a dictionary with lumi numbers as the key, event counts as the value
+            singleRun.update({'lumis': [int(lumi) for lumi in lumis.keys()],
+                              'eventsPerLumi': lumis.values()})
+        elif isinstance(lumis, list):
+            singleRun.update({'lumis': lumis})
+        runList.append(singleRun)
+
+    return runList
 
 def _changeToFloat(value):
     if value in ["-nan", "nan", "inf", ""]:
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -335,8 +335,7 @@ def createFileArray(fwjr, fArray, fArrayRef):
                             fArray[fileType].add(fileName)
                     else: # this should be string
                         fArray[fileType].add(value)
-            else:
-                createFileArray(value, fArray, fArrayRef)
+            createFileArray(value, fArray, fArrayRef)
     elif isinstance(fwjr, list):
         for item in fwjr:
             createFileArray(item, fArray, fArrayRef)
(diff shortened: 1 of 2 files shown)


ericvaandering:
@ticoann can you look at the two bold warnings in this code? Seems to me that 
```
for _ in blah:
    blah
else:
    do something
```

is the same as 

```
for _ in blah:
    blah
do something
```

if you don't have a `break` in the loop

ticoann:
@ericvaandering, Eric, you are right please remove "else", Thanks.

ticoann:
For WMArchive.

https://github.com/dmwm/WMArchive/blob/master/src/python/WMArchive/Schemas/FWJRProduction.py#L47

'eventsPerLumi'  has to be added.  There might be more place to be changed but @vkuznet would know.

> Would it be better if I filled eventsPerLumi with Nones in we don't have the info rather than leaving the value out of the JSON completely?

I don't think you have to fill the field with Nones. We can leave the value out of the JSON completely. WMArchive can set the default value as [].

One other thing is FWJR has to be changed for this to work

https://github.com/dmwm/WMCore/blob/b27c386bcd45a2f9447cd585535c85eb7c2ae697/src/python/WMCore/FwkJobReport/Report.py#L87

I am not sure whether that will break any other places. I updated comment on #7728

ericvaandering:
PR made against WMArchive: https://github.com/dmwm/WMArchive/pull/325

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
Seeing if new Jenkins 2.0 setup picks this up

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

amaltaro:
@ericvaandering please remove the label once it's ready to be merged

ericvaandering:
Should go after #7728 but otherwise good to go. I did not put the label on there. :-)

amaltaro:
@vkuznet @yuyiguo 
Since I haven't seen any replies to Eric's question, I wanted to make sure everyone is on the same page regarding the run/lumi/event data and data format we upload to WMArchive.

If you look at the code, you will see it defines a dict for each run processed by that job. This dict also contains `lumis` and `eventsPerLumi` keys, where their values is a "sorted" list with the lumi numbers and number of events in that lumi. E.g. 
```
u'runs': [{'eventsPerLumi': [None, None, None],
                'lumis': [115, 121, 118],
                'runNumber': 283171}],
```

It's going to look weird for workflows processing way too many lumis in the same job (LHE), but on the other hand, I think that's the best way to have a consistent data format...

ericvaandering:
I'd say its an "ordered" list where the order in eventsPreLumi corresponds to the order in lumis, but neither list need be sorted.

vkuznet:
Alan,
do we have an estimate of number of lumis in LHE job?
In other words I need to know how large the FWJR will become with respect to
existing one (estimated at ~10KB).

ericvaandering:
It seems to me you should be able to answer that since the lumis are already there, it’s just adding the #/lumi.

amaltaro:
Eric, yes, I meant to say ordered and not sorted.
Valentin, as Eric said, the best would be if you could fetch one of these documents for a LHE request (the output files have LHE datatier). Otherwise, I think we can have up to 100k events per job, though I'm not sure if the lumi size is set to 1k or 300 events for them.

PR #7751, opened 2017-03-23
add DAO and method to check DBS and PhEDEx complete status

ticoann:
Initial call for checking DBS PhEDEx upload status.

### DIFF
--- a/bin/check-phedex-dbs-status
+++ b/bin/check-phedex-dbs-status
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+from __future__ import (print_function, division)
+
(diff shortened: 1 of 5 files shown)


ticoann:
@goughes, Erik, could you create the unittest for this.
You can create (Maybe for other functions in DBSBufferUtil.py 😄 )

WMCore/test/python/WMComponent_t/DBS3Buffer_t/DBSBufferUtil_t.py

You can look at the unittest here. It should be similar way.

https://github.com/dmwm/WMCore/blob/master/test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py

Thanks,

ticoann:
replaced by #7757

PR #7752, opened 2017-03-23
Make config cache list work in ACDC workflow

ticoann:
as described in the title. also change the config file label to config cache list


ticoann:
tested in vm

PR #7757, opened 2017-03-28
add DAO and method to check DBS and PhEDEx complete status and DBSBufferUtil unittest

goughes:
This includes PR #7751 from Seangchan and initial commit of the DBSBufferUtil unittest.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
@@ -4,8 +4,10 @@ _DBSBufferUtil_
 
 APIs related to using the DBSBuffer.
 """
+from __future__ import print_function
 
 import threading
+from collections import defaultdict
 
 from WMCore.DAOFactory import DAOFactory
 from WMComponent.DBS3Buffer.DBSBufferFile import DBSBufferFile
@@ -202,3 +204,38 @@ class DBSBufferUtil(WMConnectionBase):
         result = wfCompletedDAO.execute(transaction = False)
 
         return result
+    
+    def getPhEDExDBSStatusForCompletedWorkflows(self, summary=False):
+        
+        """
+        _gettPhEDxDBSStatuForCompletedWorkflows_
+
+        """
+        wfCompletedStatusDAO = self.daoFactory(classname = "CheckStatusForCompletedWorkflows")
+        result = wfCompletedStatusDAO.execute(transaction = False)
+        if summary:
+            result = self.summaryPhEDExDBSStatus(result)
+            
+        return result
+
+    def summaryPhEDExDBSStatus(self, data):
+        """
+        """
+        summary = defaultdict(dict)
+        for workflow, value in data.iteritems():
+            # only getting completed workflows
+            summary[workflow]["Completed"] = True
+            
+            if value["NotInPhEDEx"] == 0 and value["InPhEDEx"] > 0:
+                summary[workflow]["PhEDExInjected"] = True
+            else:
+                summary[workflow]["PhEDExInjected"] = False
+            
+            if value["NotInDBS"] == 0 and value["InDBS"] > 0:
+                summary[workflow]["DBSUploaded"] = True
+            else:
+                summary[workflow]["DBSUploaded"] = False
+        return summary
+                
+            
+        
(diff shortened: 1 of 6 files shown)


ticoann:
test this please

ticoann:
there is one error I made. But merging it and create additional patch

src/python/WMComponent/DBS3Buffer/MySQL/CheckStatusForCompletedWorkflows.py
```diff
@@ -0,0 +1,59 @@
...
+        results = self.formatDict(results)
+        resultByWorkflow = {}
+        for info in results:
+            workflow = info['name']
+            resultByWorkflow.setdefault(workflow, {})
+            resultByWorkflow[workflow].setdefault("InDBS", 0)
+            resultByWorkflow[workflow].setdefault("NotInDBS", 0)
+            resultByWorkflow[workflow].setdefault("InPhEDEx", 0)
+            resultByWorkflow[workflow].setdefault("NotInPhEDEx", 0)
+            resultByWorkflow[workflow].setdefault("EmptyTasks", 0)
```
amaltaro: I'm not sure we actually need this EmptyTasks. Is it the case all the jobs fail for a specific task, thus a task with empty block (aka no output files)? If that's the case, all the other flags with value 0 would mean the same thing.

src/python/WMComponent/DBS3Buffer/Oracle/CheckStatusForCompletedWorkflows.py
```diff
@@ -0,0 +1,17 @@
...
+Oracle implementation of DBSBuffer.CheckStatusForCompletedWorkflows
+"""
+from __future__ import print_function, division
+
+from WMComponent.DBS3Buffer.MySQL.CheckStatusForCompletedWorkflows \
+ import CheckStatusForCompletedWorkflows as MySQLCheckStatusForCompletedWorkflows
+
+class CheckStatusForCompletedWorkflows(MySQLCheckStatusForCompletedWorkflows):
+    """
+    _CountFiles_
```
amaltaro: hmm, not really this docstring :)

src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
```diff
@@ -202,3 +204,38 @@ def getCompletedWorkflows(self):
...
+        wfCompletedStatusDAO = self.daoFactory(classname = "CheckStatusForCompletedWorkflows")
+        result = wfCompletedStatusDAO.execute(transaction = False)
+        if summary:
+            result = self.summaryPhEDExDBSStatus(result)
+            
+        return result
+
+    def summaryPhEDExDBSStatus(self, data):
+        """
+        """
```
amaltaro: and you haven't fixed this docstring

src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
```diff
@@ -202,3 +204,38 @@ def getCompletedWorkflows(self):
...
+        summary = defaultdict(dict)
+        for workflow, value in data.iteritems():
+            # only getting completed workflows
+            summary[workflow]["Completed"] = True
+            
+            if value["NotInPhEDEx"] == 0 and value["InPhEDEx"] > 0:
```
amaltaro: or should it be `if value["NotInPhEDEx"] == 0 and value["InPhEDEx"] >= 0:`, for cases that no output files were produced? One could say the phedex injection is done (though there were no files :D). ... I don't know, it all depends on how you interpret it.
Same below for DBS

amaltaro (commented):
Looks good, just some minor questions/suggestions

src/python/WMComponent/DBS3Buffer/MySQL/CheckStatusForCompletedWorkflows.py
ticoann: It is the case when workflow is injected in the system but no files are produced yet. I don't think this will happen likely but I think we need to check this for drain status. That said, I think I have to update the summary function to include this value.

ticoann: Maybe I can change the name for clarity to "OutputNotProduced".

src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
ticoann: I think it needs be handled by EmptyTask below. Which means everything is failed.

ticoann:
Thanks Alan for the review. I will created the new patch with doc string and clearer comments.
As far as "EmptyTask" field, I think we need that to distinguish between no output is created vs. all the outputs are not yet injected. But I have to think more, you make me confused. :-)

ticoann:
Hi Alan, I think you are right on your advice. #7774

PR #7763, opened 2017-03-31
fix killing the jobs when race condition happens

### LINKED ISSUE #7762: Job left over from aborted workflow
I noticed 
https://cms-gwmsmon.cern.ch/prodview/fabozzi_RVCMSSW_9_0_0GluGluHToZZTo4L_M125_Pow_py8_Evt_13__PU25_reHLTcc7_170327_230233_6021

https://cmsweb.cern.ch/reqmgr2/fetch?rid=fabozzi_RVCMSSW_9_0_0GluGluHToZZTo4L_M125_Pow_py8_Evt_13__PU25_reHLTcc7_170327_230233_6021

which is aborted-archived. From the relval agent. Is that agent fully up to date on the release version ?

ticoann:
fixes #7762

### DIFF
--- a/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
+++ b/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
@@ -186,6 +186,15 @@ class TaskArchiverPoller(BaseWorkerThread):
                 finishedwfsWithLogCollectAndCleanUp[wf] = finishedwfs[wf]
         return (finishedwfs, finishedwfsWithLogCollectAndCleanUp)
 
+    def killCondorJobsByWFStatus(self, statusList):
+        if isinstance(statusList, basestring):
+            statusList = [statusList]
+        reqNames = self.centralCouchDBWriter.getRequestByStatus(statusList)
+        logging.info("There are %d requests in 'aborted' status in central couch.", len(reqNames))
+        for wf in reqNames:
+            self.workQueue.killWMBSWorkflow(wf)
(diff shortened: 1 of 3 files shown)


ticoann:
test this please

ticoann:
test this please

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -208,9 +217,8 @@ def completeTasks(self, finishedwfs):
 
         centralCouchAlive = True
         try:
-            abortedWorkflows = self.centralCouchDBWriter.getRequestByStatus(["aborted"])
-            logging.info("There are %d requests in 'aborted' status in central couch.", len(abortedWorkflows))
-
+            abortedWorkflows = self.killCondorJobsByWFStatus("aborted")
```
amaltaro: we should pass a list as arg to avoid a few extra cpu cycles casting it

amaltaro (commented):
I think you explained me why WorkQueueManager polling cycle affects it, but I cannot remember.

If it queries central couch for the last request transition, why do we care about local WQ cycle?

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
ticoann: I don't think that is a big difference given this will be polled every few minutes. But I will update that. :-)

PR #7771, opened 2017-04-05
Fix the unit test broken by #7728

ericvaandering:
Now that I can see it, I can fix it.


ericvaandering:
@ticoann please merge after tests

PR #7774, opened 2017-04-05
fixes what Alan suggested

ticoann:
From #7757 and #7763


amaltaro (approved):
Wonderful! Especially the subject ;)

PR #7775, opened 2017-04-06
add the DAO to check incomplete workflow in agent

ticoann:
For checking draining status, use this for the first check if this passes check the DBS, PhEDEx upload status.

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
@@ -213,8 +213,8 @@ class DBSBufferUtil(WMConnectionBase):
         _getPhEDxDBSStatuForCompletedWorkflows_
         Check the PhEDEx and DBS upload status for the completed workflow
         """
-        wfCompletedStatusDAO = self.daoFactory(classname = "CheckStatusForCompletedWorkflows")
-        result = wfCompletedStatusDAO.execute(transaction = False)
+        wfCompletedStatusDAO = self.daoFactory(classname="CheckStatusForCompletedWorkflows")
+        result = wfCompletedStatusDAO.execute(transaction=False)
         if summary:
             result = self.summaryPhEDExDBSStatus(result)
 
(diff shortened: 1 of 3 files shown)


src/python/WMComponent/DBS3Buffer/DBSBufferUtil.py
```diff
@@ -240,3 +240,11 @@ def summaryPhEDExDBSStatus(self, data):
             else:
                 summary[workflow]["DBSUploaded"] = False
         return summary
+
+    def isAllWorkflowCompleted(self):
+        """
+        check whether all the workflows are completed status in give agent (not globaly in reqmgr)
+        This should be used as preconditon for checking draining status
+        """
+        completeFWFlag = self.daoFactory(classname="isAllWorkflowsCompleted")
```
amaltaro: dao name is wrong

src/python/WMComponent/DBS3Buffer/Oracle/IsAllWorkflowsCompleted.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import print_function, division
+"""
+_IsAllWorkflowsCompleted_
+
+Oracle implementation of DBS3Buffer.IsAllWorkflowsCompleted
+"""
+
```
amaltaro: no need for all these blank lines :)

src/python/WMComponent/DBS3Buffer/Oracle/IsAllWorkflowsCompleted.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+
+from WMComponent.DBS3Buffer.MySQL.IsAllWorkflowsCompleted \
+    import IsAllWorkflowsCompleted as MySQLIsAllWorkflowsCompleted
+
+class IsAllWorkflowsCompleted(MySQLIsAllWorkflowsCompleted):
+    """
+    Oracle version
```
amaltaro: you probably forgot to close this docstring and pass this class

amaltaro (changes requested):
A few minor issues need to be addressed, otherwise the code looks good.

However, I think we should make this DAO a bit more generic to avoid creating gazillions of DAOs for the draining process. I thought we could make it return two different results:
1) just a boolean flag, as it is already implemented
2) return distinct names - for cases we want to know further what are the workflows still holding us on this draining.

So we could replace that SQL query by
`select distinct name from dbsbuffer_workflow WHERE completed=0;`
what do you think?

ericvaandering:
test this please

ticoann:
included in #7926

PR #7778, opened 2017-04-06
Add campaign to request list, this should enable sorting

### LINKED ISSUE #7768: remove campaign filter from other than approve pages in reqmgr2
Or fix the filter.

vkuznet:
fixes #7768


ticoann:
Thanks Valentin, I will run the test and merge this.

ticoann:
@vkuznet, Hi Valentin, It still has problem

https://reqmgr2-dev.cern.ch/reqmgr2/requests?status=new
campaign field still undefined and filter doesn't work

for the same request approve page works

https://reqmgr2-dev.cern.ch/reqmgr2/approve

vkuznet:
Seangchan,
but it looks like that campaign attribute is not defined for requests page, see
Campaign: undefined

while it is fine in approve interface, see Campaign: HG1704Test

Can we dump request JSON passed to requests API?

Valentin.

ticoann:
Those pages (approve and request) use different call to get the data?  could you point me the call which gets the data for each pages?

vkuznet:
Your request forced me to look further and I think I found a culprit.
I updated PR https://github.com/dmwm/WMCore/pull/7778/files
please redeploy and let's check what will come out.
The problem was that I didn't force Campaign in look-up attributes.

ticoann:
Thanks Valentin, it works now.

PR #7789, opened 2017-04-13
Use python 2.7.13 on the worker nodes

amaltaro:
WMAgent version of #7807
If we decide to go with the new COMP python build, then we need to add this PR to the agent.
However, before that we need to have it deployed on CVMFS and OSG sites.

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -1,4 +1,38 @@
 #!/bin/bash
+# Function to check the exit code of this bootstrap script and the job/python
+# wrapper exit code.
+# 1) If the bootstrap exit code is not 0, then something is wrong with the worker
+#    node and this script will sleep for WMA_MIN_JOB_RUNTIMESECS before it exits.
+# 2) If the job exit code is not 0, then again sleep for WMA_MIN_JOB_RUNTIMESECS
+# 3) If all exit codes are 0, then just quit
+finish() {
+  exitCode=$?
+  echo "======== WMAgent final job runtime checks STARTING at $(TZ=GMT date) ========"
+  END_TIME=$(date +%s)
+  DIFF_TIME=$((END_TIME-START_TIME))
+  echo "$(TZ=GMT date): Job Runtime in seconds: " $DIFF_TIME
+  echo "$(TZ=GMT date): Job bootstrap script exited: " $exitCode
+  echo "$(TZ=GMT date): Job execution exited: " $jobrc
+
+  if [ $exitCode -ne 0 ];
+  then
+    WMA_MIN_JOB_RUNTIMESECS=300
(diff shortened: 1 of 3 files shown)


ericvaandering:
Seems to me we can be smarter than having to update version numbers every time. :-) Can we auto discover this?

amaltaro:
We can always be smarter, though it depends on the cost-benefit, which for this case is very small :-) 
Anyways, I'll update this patch such that it looks for the latest python release and just use that. That will require special hooks for the moment we migrate to python 3 as well.

amaltaro:
Eric, the second commit has the changes you suggested. I still need to test it though.

amaltaro:
Tested both at FNAL and CERN and it works Ok, though I still want to test it in other US and EU sites. @ticoann good for a review though.

etc/submit.sh
```diff
@@ -1,4 +1,72 @@
...
+{
+  if [ -z "$prefix" ]; then
+    echo "ERROR: Python directory is empty"
+    exit 6
+  elif [ -z "$suffix" ]; then
+    echo "ERROR: Suffix for python directory is empty"
+    exit 7
+  fi
+
+  local latest='2.7.6'         # default release
```
ticoann: shouldn't we set default to 2.7.13 ?

ericvaandering:
test this please

ericvaandering:
(testing Jenkins, not the PR)

ericvaandering:
test this please

etc/submit.sh
amaltaro: It doesn't exist in cvmfs yet, that's why I made an existent one as default.

ticoann: I see. Thanks,

amaltaro:
[Fix #7789 - Use python 2.7.13 on the worker nodes](https://trello.com/c/eBjWafrE/29-fix-7789-use-python-2-7-13-on-the-worker-nodes)

amaltaro:
And I need to make sure it can properly handle a "version" directory like `2.7.11-comp`. I'll test it tomorrow.

amaltaro:
sigh... I rebased it with master instead of the wmagent branch!

amaltaro:
@vkuznet can you point me to the script you are talking about in the meeting?

vkuznet:
/cvmfs/cms.cern.ch/common/das_client

amaltaro:
Thanks, Valentin. Much easier indeed.

amaltaro:
Seems to be working fine. I tested it in a couple US sites, Vienna, Bari, Jinr and there were no issues.
I'm merging it now in order to cut a new wmagent tag. However, please feel free to review it.

PR #7783, opened 2017-04-10
add aux db update

ticoann:
https://github.com/dmwm/deployment/pull/470  needed  with this patch. Still need to test and update. Don't review yet.

Fixes #7759, #6072

PR #7790, opened 2017-04-13
Fix request archival logic

amaltaro:
That's why we have some workflows (either in my VM and testbed) not moving to archived.

I was going to say that I don't understand why we need to pass the `AgentJobInfo` as input condition
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py#L90

but I think the answer is that the agent may still be running logCollect/Cleanup, thus we don't want to archive it. 

@ticoann is that the case?

If so, then this one should be getting archived:
https://cmsweb-testbed.cern.ch/wmstatsserver/data/filtered_requests?RequestStatus=announced&mask=RequestTransition&mask=AgentJobInfo

since all the subscriptions are finished.

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -183,10 +183,8 @@ class RequestInfo(object):
         for key, value in filterDict.iteritems():
             # special case checks where key is not exist in Request's Doc.
             # It is used whether AgentJobInfo is deleted or not for announced status
-            if value == "NO_KEY" and key not in self.data:
+            if value == "NO_KEY":
                 continue
-            elif value == "NO_KEY" and key in self.data:
-                return False
 
             if isinstance(value, dict):
                 # TODO: need to handle dictionary comparison


ticoann:
Hi Alan, we are not just checking whether all the tasks (including CleanUp, LogArchive tasks), before request gets archived, all the data has to be cleaned up as well. That is why we are checking AgentInfo is None. Is this Agent redeployed without completely drained?

ericvaandering:
test this please

ericvaandering:
test this please

ticoann:
Hi Alan, the reason it is not gets to archived is that replication doesn't include deleted documents. 
I don't remember why we changed not to replicate the deleted document. Anyway, I will create a new patch.

ticoann:
FYI. I changed the replication on vocms0224 just for the testing.

ticoann:
Hi Alan, I created a initial patch for this #7797. But looking at the code again, current code still need to work.
This function should clean the document
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py#L26

And shouldn't be updated by agent once agent deleted the document

Could you reject and announce  a few workflows in testbed which ran in vocms0230?
I didn't want to meddle with what you are testing without your permission. 😄

amaltaro:
Seangchan, which documents are these? Is it the aggregated/summary doc the agent creates for each active workflow in the system every X minutes?
IIUC, while the agent has data for a workflow, this document will keep coming to wmstats, right?

amaltaro:
And I forgot to mention, these 3 workflows should be either in announced or rejected:
amaltaro_RelVal_Premix_HG1705_Validation_170418_232048_11
amaltaro_MonteCarlo_DeterPU_HG1705_Validation_170418_232212_5164
amaltaro_StepChain_MCStatic_HG1705_Validation_170418_231320_2172

they have all failed anyways :)

ticoann:
replaced by #7783

PR #7793, opened 2017-04-18
Split up a long running test into two parts

ericvaandering:
This test can take 10 minutes as is. It seems to be two independent parts that can be two tests.

### DIFF
--- a/test/python/WMCore_t/WMBS_t/JobSplitting_t/FileBased_t.py
+++ b/test/python/WMCore_t/WMBS_t/JobSplitting_t/FileBased_t.py
@@ -129,7 +129,7 @@ class FileBasedTest(unittest.TestCase):
         """
         testFileset = Fileset(name = "TestFilesetX")
         testFileset.create()
-        for i in range(5000):
+        for _ in range(5000):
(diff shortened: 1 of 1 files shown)


ericvaandering:
@ticoann please merge

test/python/WMCore_t/WMBS_t/JobSplitting_t/FileBased_t.py
```diff
@@ -129,7 +129,7 @@ def createLargeFileBlock(self):
         """
         testFileset = Fileset(name = "TestFilesetX")
         testFileset.create()
-        for i in range(5000):
+        for _ in range(5000):
```
ticoann: I wonder where 5000 comes from. but if it doesn't have to be 5000, whether we can reduce to 1000 if that helps reducing testing time significantly

ericvaandering: I don't know. I just tried to keep the logic exactly the same. 

The 5 or 10 minutes isn't *really* a problem, it just sets the scale of how short you can make the total testing cycle if you farm things out more widely.

PR #7795, opened 2017-04-19
Make central services report their version to wmstats

amaltaro:
otherwise it's just undefined for the moment.


ticoann:
@amaltaro,  Alan, I include this on the other fetch to avoid conflict. #7796

PR #7797, opened 2017-04-20
initial patch to handle archived

ticoann:
replace #7790

Still need to add config duration and archive delay params.

### DIFF
--- a/src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
+++ b/src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
@@ -3,6 +3,8 @@ from __future__ import (division, print_function)
 import traceback
 from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
 from WMCore.Services.WMStats.WMStatsWriter import WMStatsWriter
+from WMCore.Services.WMStats.DataStruct.RequestInfoCollection import RequestInfo
+from WMCore.WMStats.DataStructs.DataCache import DataCache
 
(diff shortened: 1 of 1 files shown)


src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
```diff
@@ -41,13 +57,18 @@ def cleanUpArchivedRequests(self, config):
...
-            except Exception as ex:
-                self.logger.error("deleting %s failed: %s", req, str(ex))
-                for line in traceback.format_exc().rstrip().split("\n"):
-                    self.logger.error(" " + line)
-            else:
-                self.logger.info("%s deleted", len(result))
+            self._deleteWMStatsDocsByRequest(self, req)
         return
+
+    def cleauUpAgentDocs(self, config):
```
amaltaro: method name misspelled (same above)

src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
```diff
@@ -41,13 +57,18 @@ def cleanUpArchivedRequests(self, config):
...
+        """
+        delete wmstats document if all the job is finished including logCollect and Clean up
+        It won't be completely sync with agent deletion, hence there is always chance for race conditoin
```
amaltaro: conditoin misspelled

amaltaro (commented):
Seangchan, I was wondering if we should change the cleanup logic on the agent side to only delete stuff based on the request status (archived). This way we also guarantee that all agents will delete all the data for the workflows they were working on.

PR #7803, opened 2017-04-20
fix moving to archived status

ticoann:
replace #7797, #7790


ticoann:
Tested in vm, merging it for the new tag.
@amaltaro, still please review.  I think the problem was I added PhEDEx and DBS report on the report and I think that never gets cleaned up.

PR #7801, opened 2017-04-20
fixes proxy info warning 1.1.2_wmagent version

ticoann:
#7796 wmagent branch version

PR #7804, opened 2017-04-21
fix clone api

ticoann:

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -30,8 +30,7 @@ ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST = \
      'Teams', 'TotalEstimatedJobs', 'TotalInputEvents', 'TotalInputFiles', 'TotalInputLumis',
      'TransientOutputModules', 'TrustPUSitelists', 'TrustSitelists', 'VoRole', '_id', '_rev']
 
-
-def initialize_request_args(request, config, clone=False):
+def initialize_request_args(request, config):
     """
     Request data class request is a dictionary representing
     a being injected / created request. This method initializes
@@ -55,14 +54,10 @@ def initialize_request_args(request, config, clone=False):
                                      "UpdateTime": int(time.time()), "DN": request["RequestorDN"]}]
     request["RequestDate"] = list(time.gmtime()[:6])
 
-    if clone:
-        # if it is clone parameter should contain requestName
-        request["OriginalRequestName"] = request["RequestName"]
-    else:
-        # update the information from config
-        request["CouchURL"] = config.couch_host
-        request["CouchWorkloadDBName"] = config.couch_reqmgr_db
-        request["CouchDBName"] = config.couch_config_cache_db
+    # update the information from config
+    request["CouchURL"] = config.couch_host
+    request["CouchWorkloadDBName"] = config.couch_reqmgr_db
+    request["CouchDBName"] = config.couch_config_cache_db
 
     generateRequestName(request)
 
@@ -84,6 +79,30 @@ def initialize_resubmission(request_args, reqmgr_db_service):
     # to be used later on for spec validation
     request_args["OriginalRequestType"] = parent_args["RequestType"]
 
+def _replace_cloned_args(clone_args, user_args):
+    """
+    replace original arguments with user argument.
+    If the value is dictionary format, overwrite only with specified arguments.
+    If the original arguemnt has simple value and user passes dictionary, completely replace to dictionary
+    """
+    for prop in user_args:
+        if isinstance(user_args[prop], dict) and isinstance(clone_args.get(prop), dict):
+            _replace_cloned_args(clone_args.get(prop, {}), user_args[prop])
+        else:     
+            clone_args[prop] = user_args[prop]
+    return
+    
+def initialize_clone(request_args, reqmgr_db_service):
+    """
+    Initialize a Clone arguments by inheriting and overwriting argument from OriginalRequest
+    """
+    requests = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
+    clone_args = requests.values()[0]
+    # TODO: need to validate new overwrite request_args here since it will skip the clone_args validtaiton
+    _replace_cloned_args(clone_args, request_args)
+    clone_args = {k: v for k, v in clone_args.iteritems() if k not in ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST}
+
+    return clone_args
 
 def generateRequestName(request):
     currentTime = time.strftime('%y%m%d_%H%M%S', time.localtime(time.time()))
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -15,17 +15,14 @@ from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 from WMCore.REST.Server import RESTEntity, restcall, rows
 from WMCore.REST.Validation import validate_str
 from WMCore.ReqMgr.DataStructs.ReqMgrConfigDataCache import ReqMgrConfigDataCache
-from WMCore.ReqMgr.DataStructs.Request import initialize_request_args
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
-from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_STATE_LIST, \
-    REQUEST_STATE_TRANSITION, ACTIVE_STATUS
+from WMCore.ReqMgr.DataStructs.RequestStatus import (REQUEST_STATE_LIST, 
+                                       REQUEST_STATE_TRANSITION, ACTIVE_STATUS)
 from WMCore.ReqMgr.DataStructs.RequestType import REQUEST_TYPES
-from WMCore.ReqMgr.Utils.Validation import validate_request_create_args, \
-    validate_request_update_args, loadRequestSchema, validateOutputDatasets
+from WMCore.ReqMgr.Utils.Validation import (validate_request_create_args, validate_request_update_args,
+        validate_clone_create_args, loadRequestSchema, validateOutputDatasets)
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
-from WMCore.WMSpec.WMWorkloadTools import loadSpecByType
-
 
 class Request(RESTEntity):
     def __init__(self, app, api, config, mount):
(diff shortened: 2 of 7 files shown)


ticoann:
I had to modify a little bit of your Resubmission validation code. Anyway, I still need to test this. So I won't include in the new tag. I will try to finish tomorrow.

ticoann:
test this please

amaltaro:
test this please

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -74,25 +71,21 @@ def _validateRequestBase(self, param, safe, valFunc, requestName=None):
...
-            # cherrypy.log(str(param.kwargs))
             request_args = {}
+        # In case key args are also passed and request body also exists.
+        # If the request.body is dictionally update the key args value as well
+        if isinstance(request_args, dict):
             for prop in param.kwargs:
                 request_args[prop] = param.kwargs[prop]
 
-            for prop in request_args:
+            for prop in param.kwargs.keys():
```
amaltaro: How about merging these two for loops into:
```
for prop in param.kwargs.keys():
                 request_args[prop] = param.kwargs.pop(prop)
```

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -181,22 +167,23 @@ def validate(self, apiobj, method, api, param, safe):
...
-            #                 if len(param.args) == 2:
-            #                     #validate clone case
-            #                     if param.args[0] == "clone":
-            #                         param.args.pop()
-            #                         return None, request_args
 
             if method == 'POST':
                 args_length = len(param.args)
-                if args_length == 1 and param.args[0] == "multi_update":
+                if len(param.args) == 2 and param.args[0] == "clone":
```
amaltaro: `args_length` can be used in this line :-)

src/python/WMCore/Services/ReqMgr/ReqMgr.py
```diff
@@ -214,3 +214,17 @@ def getAbortedAndForceCompleteRequestsFromMemoryCache(self):
...
         return MemoryCacheStruct(expire=0, func=self.getRequestByStatus, initCacheValue=[],
                                  kwargs={'statusList': maskStates, "detail": False})
+
+    def cloneRequest(self, requestName, overwrittenParams=None):
+        """
+        _updateRequestStatus_
+
+        :param requestName: request name which need to be cloned
+        :type reqeust: str
+        :param overwrittenParams: status of workflow to update (i.e. 'assigned')
```
amaltaro: This doc string needs to be updated (and typo above)

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -875,8 +840,17 @@ def factoryWorkloadConstruction(self, workloadName, arguments):
 
         Named this way so that nobody else will try to use this name.
         """
-        self.masterValidation(schema=arguments)
-        self.validateSchema(schema=arguments)
+        # TODO: eventually perform a complete argument validation for Resubmission workflows
```
amaltaro: these comments can be removed now, since we do have a spec validation for Resubmission in the spec `validateSchema`

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -875,8 +840,17 @@ def factoryWorkloadConstruction(self, workloadName, arguments):
...
+        # For now, we mostly trust the inherited values and the few args provided by the user
+        # Change to OriginalRequestName to handle clone workflow.
+        if 'OriginalRequestName' not in arguments:
+            self.masterValidation(schema=arguments)
+            self.validateSchema(schema=arguments)
+        elif arguments.get("RequestType") != "Resubmission":
```
amaltaro: Should be equals, not different.

amaltaro (changes requested):
Seangchan, a few updates need to be done to the code, but otherwise it looks good.

I do have a question about these requests though that always confuse me. It's about these GET/PUT/POST calls and their param.kwargs + body message. I don't get why we use some and delete others and etc. We can chat later today

amaltaro:
[Fix #7801 - fixes proxy info warning](https://trello.com/c/4NtIFuCu/30-fix-7801-fixes-proxy-info-warning)

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -85,6 +78,19 @@ def initialize_resubmission(request_args, acceptedArgs, reqmgr_db_service):
...
             request_args[arg] = parent_args[arg]
 
+def initialize_clone(request_args, reqmgr_db_service):
+    """
+    Initialize a Clone arguments by inheriting and overwriting argument from OriginalRequest
+    """
+    requests = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
+    clone_args = requests.values()[0]
+    # TODO: need to validate new overwrite request_args here since it will skip the clone_args validtaiton
+    for prop in request_args:
```
ticoann: You mean only some part of inner dict not the whole thing. I wonder whether we should allow removing some params as well.

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: Yes I will update that.

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: copy and past problem. will update.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
ticoann: Yes you are right. I wonder why it didn't fail when I run the clone test.

ticoann:
Alan, updated code and fix the conflict, please review again.

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -84,6 +79,30 @@ def initialize_resubmission(request_args, reqmgr_db_service):
     # to be used later on for spec validation
     request_args["OriginalRequestType"] = parent_args["RequestType"]
 
+def _replace_cloned_args(clone_args, user_args):
+    """
+    replace original arguments with user passed arguement.
```
amaltaro: A few things misspelled here.

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -84,6 +79,30 @@ def initialize_resubmission(request_args, reqmgr_db_service):
...
+    If the value is dictionary format, overwrite only specified arguments.
+    If the original arguemnt has simple value and user passes dictionary, completely replace to dictionary
+    """
+    for prop in user_args:
+        if isinstance(prop, dict) and isinstance(clone_args.get(prop), dict):
```
amaltaro: `prop` will always be a string here (initially). Don't you want to:
```
if isinstance(user_args[prop], dict) and isinstance(clone_args.get(prop), dict):
    _replace_cloned_args(clone_args.get(prop, {}), user_args[prop])
```

?

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -84,6 +79,30 @@ def initialize_resubmission(request_args, reqmgr_db_service):
...
+            clone_args[prop] = user_args[prop]
+    return
+    
+def initialize_clone(request_args, reqmgr_db_service):
+    """
+    Initialize a Clone arguments by inheriting and overwriting argument from OriginalRequest
+    """
+    requests = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
+    clone_args = requests.values()[0]
+    # TODO: need to validate new overwrite request_args here since it will skip the clone_args validtaiton
```
amaltaro: In the future we can get the createArguments for that specific requestType and filter out the clone_args + user_args. This way we can trigger the normal spec validation later on.

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -74,25 +71,20 @@ def _validateRequestBase(self, param, safe, valFunc, requestName=None):
...
             request_args = {}
-            for prop in param.kwargs:
+        # In case key args are also passed and request body also exists.
+        # If the request.body is dictionally update the key args value as well
+        if isinstance(request_args, dict):
+            for prop in param.kwargs.keys():
                 request_args[prop] = param.kwargs[prop]
-
-            for prop in request_args:
                 del param.kwargs[prop]
```
amaltaro: My suggestion was to use the `pop` dict method to get+remove that key/value. Two lines become a single one :)

amaltaro (changes requested):
Unless I misunderstood something, there should be still a couple of fixes to be applied.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -875,8 +840,14 @@ def factoryWorkloadConstruction(self, workloadName, arguments):
 
         Named this way so that nobody else will try to use this name.
         """
-        self.masterValidation(schema=arguments)
-        self.validateSchema(schema=arguments)
+        if 'OriginalRequestName' not in arguments:
```
amaltaro: One more comment that I noticed only now. We either have to
a) remove `OriginalRequestName` key/value from the request dictionary (only for clone, keep it for resubmission)
or
b) we have to add it to the StdBase create definition.

Otherwise it will fail the validation. If we go with b), then we may need to update this code and a few other places as well.

ticoann: For clone, we don't have validation function executed. (We will skip the validation so it should be fine). I think eventually I would like to add validation for only newly overwritten values not the ones which was cloned since that should be validated before (in original request creation). Are you actually seeing the validation fails. I didn't have problem when I clone the request.

src/python/WMCore/ReqMgr/DataStructs/Request.py
ticoann: Yes you're right. I will update that

src/python/WMCore/ReqMgr/Service/Request.py
ticoann: Sorry my mistake.

src/python/WMCore/ReqMgr/DataStructs/Request.py
ticoann: I was thinking just validate user_args. clone_args are already validated in the original workflow. Anyway, we can discuss later.

amaltaro: Validating only user_args makes sense to me. I'm just worried about legacy workflows (let's say previous release) that could have a data format/etc different that what we expect now. Though this would be a case in 10000 I guess...

PR #7805, opened 2017-04-21
Make a clear distinction between ReqMgr2 and user create arguments

amaltaro:
@ticoann I'm pretty sure you predicted this confusion... :-(

Besides this distinction, the only changes are:
*) RequestString is not a ReqMgr2 argument, but user has to provide it
*) RequestStatus is mandatory during assignment (maybe there was a reason I set it to optional...)

Once we merge it, we should point the master code to Antanas again.

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -134,16 +134,7 @@ def validate_request_create_args(request_args, config, reqmgr_db_service, *args,
     if request_args["RequestType"] == "Resubmission":
         # do not set default values for Resubmission since it will be inherited from parent
         # both create & assign args are accepted for Resubmission creation
-        acceptedArgs = {}
-        if 'OriginalRequestType' in request_args:
-            parentSpecClass = loadSpecClassByType(request_args['OriginalRequestType'])
-            acceptedArgs = parentSpecClass.getWorkloadCreateArgs()
-
-        acceptedArgs.update(specClass.getWorkloadCreateArgs())
-        assignArgs = specClass.getWorkloadAssignArgs()
-        acceptedArgs.update(assignArgs)
-
-        initialize_resubmission(request_args, acceptedArgs, reqmgr_db_service)
+        initialize_resubmission(request_args, reqmgr_db_service)
     else:
         # set default values for the request_args
         setArgumentsWithDefault(request_args, specClass.getWorkloadCreateArgs())
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -95,37 +95,21 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
     # validate the status
     if "RequestStatus" in request_args:
         validate_state_transition(reqmgr_db_service, request_name, request_args["RequestStatus"])
-        # delete request_args since it is not part of spec argument and validation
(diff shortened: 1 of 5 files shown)


amaltaro:
the reason it was optional during assignment comes from the fact that we remove that key/value during the arguments handling/validation. Don't merge it yet, if it's simple, I'll just fix the validation code, otherwise I make it optional in the definition again.

amaltaro:
2nd commit fixes that issue. Ive just need more time to make sure it won't break in a very specific case.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -1057,18 +1047,30 @@ def getWorkloadCreateArgs():
...
                     }
+
+        # these arguments are internally set by ReqMgr2 and should not be provided by the user
+        reqmgrArgs = {"Requestor": {"attr": "owner", "optional": False},
+                      "RequestorDN": {"attr": "owner_dn", "optional": False, "null": False},
+                      "RequestName": {"optional": False, "null": False, "validate": identifier},
+                      "RequestStatus": {"optional": False, "validate": lambda x: x == REQUEST_START_STATE},
+                      "RequestTransition": {"optional": False, "type": list},
+                      "RequestDate": {"optional": False, "type": list},
+                      "CouchURL": {"default": "https://cmsweb.cern.ch/couchdb", "validate": couchurl},
```
ticoann: This can be provide by user as optional value. Since we could use testbed for the this. Or Are we automatically set this?

ticoann: Never mind it seems it is always automatically set by ReqMgr

amaltaro:
@ticoann I tested all test workflows and resubmission for all request types. This time it should be working fine.
Please review and merge, if possible, before your clone fix :-)

amaltaro:
test this please

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -95,37 +95,21 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
...
-        # validate the arguments against the spec argumentSpecdefinition
-        # TODO: currently only assigned status allows any update other then Status update
-        workload.validateArgumentForAssignment(args_without_status)
-
-    # to update request_args with type conversion
-    request_args.update(args_without_status)
+        elif request_args["RequestStatus"] == 'assigned':
+            workload.validateArgumentForAssignment(request_args)
+
+    # TODO: fetch it from the assignment arg definition
```
ticoann: You mean 
```
reqPrioDef = workload.getWorkloadCreateArgs()["RequestPriority"]
request_args['RequestPriority'] = reqPrioDef["type"](request_args['RequestPriority'])
if not reqPrioDef["validate"](request_args['RequestPriority']):

```
Was lamdba function failing? That is why you didn't include?

ticoann:
Thanks Alan, Patch looks good to me. I'm merging it before clone patch. 😄

src/python/WMCore/ReqMgr/Utils/Validation.py
amaltaro: Yes, sort of. I'd like to eventually get the allowed actions
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/RequestStatus.py#L74
and call the validation functions (including type check) for all those arguments that are going to be updated. Just as we do for creation or assignment, but with less args defined.

PR #7809, opened 2017-04-24
Allow -fake- Campaign arg at chain level

amaltaro:
As requested by @vlimant and @anorkus, allow `Campaign` to be set at task/step level as well, during creation time. It has no meaning in the WMCore code, it's still a request argument and the system will only use the top level argument.

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -270,6 +270,14 @@ def acqname(candidate):
     else:
         return check(r'[a-zA-Z][a-zA-Z0-9_]*$', candidate)
 
+def campaign(candidate):
+    """
+    Check for Campaign name.
+    letters, numbers, underscores and dashes are allowed, up to 60 chars.
+    """
+    if not candidate:
+        return True
+    return check(r'^[a-zA-Z0-9-_]{1,60}$', candidate)
 
 def primdataset(candidate):
     """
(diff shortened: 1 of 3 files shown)


vlimant:
hey @amaltaro out of curiosity, how is it used at the top level ?

amaltaro:
Campaign gets propagated to WMArchive and AFAIR it's part of the dbs meta-data, at least.

vlimant:
if it is a dataset dbs-meta-data, then it makes sense to have it per task (same as prepid), acquisition era, ... i.e. it should not remain "fake"

amaltaro:
Nope, I was wrong, it's not part of the dbs meta-data.

@ticoann I did a few basic tests and everything looks fine. Once jenkins are happy, please merge. If something needs to be changed, please make a new PR or I'll fix it tomorrow morning.

ticoann:
I was just wondering why this need to set in book keeping purpose? Is due to breaking some McM or Ops code? If Campaign is only meaningful on request level not the chain level (only one campaign per request), we shouldn't set in the chain level as well. I think it could cause the confusion later on.
However, without this if something breaks, we can add this but eventually we should consider removing it again.

ticoann (approved):

vlimant:
On Ops side, configuration are done per campaign, and need to be specified per task, since they can belong to different campaigns. It's a very useful member, like SubRequestType, or others

ticoann:
So can campaign  be different by tasks? I thought all the tasks in the same workflow should have same campaign. If that is the case this patch  is necessary.

amaltaro:
This patch is now in testbed, tag pre5 :)

ticoann:
> This patch is now in testbed, tag pre5 :)

I know, I just wanted to revert back next release if Campaign is always request level property not varies by task. 😄

PR #7813, opened 2017-04-26
fix clone api removing original request param

ticoann:
@amaltaro, Alan please review

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -30,6 +30,11 @@ ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST = \
      'Teams', 'TotalEstimatedJobs', 'TotalInputEvents', 'TotalInputFiles', 'TotalInputLumis',
      'TransientOutputModules', 'TrustPUSitelists', 'TrustSitelists', 'VoRole', '_id', '_rev']
 
+ARGS_TO_REMOVE_FOR_CLONE = \
+    ['DN', 'RequestStatus', 'RequestTransition', 'RequestWorkflow', 'Requestor', 'RequestorDN',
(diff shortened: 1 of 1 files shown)


amaltaro (commented):
I'm not sure we need this patch. A clone is supposed to clone the original workflow (aka. create args), not clone the original workflow and its assignmnent settings.

Unless there is something breaking, but the new list seems to be a sublist of the former one, so we should be fine.

PS.: one thing that worries me is 'TransientOutputModules'... have to check the specs again.

ticoann:
I just wonder whether ops want to specify the values again for the cloned workflow. At least I thought they want to have previous parameter as default value. This patch is not critical since they and always specify in assignment. I was wondering whether we should treat the Resubmission the same way. What you do think?

amaltaro:
As we quickly discussed over vidyo. I think the Resubmission use case is different than clone, thus IMO:
*) Resubmission: inherits the assignment parameters such that one can easily assign it via ReqMgr2 web GUI
*) Clone: original request can be in any status (from assignment-approved up to completed) and it should be handled via script on the Ops side, so they are going to assign it with scripts as well. This way, cloning only the creation parameters make more sense to me.

But, I'm replying to mention something else :) I've just noticed we still have some assignment parameters in the cloned requests :( I know how to solve this problem, I just think we are way too late to fix it for this cycle.

ticoann:
I think assignment params in clone would be OK. Since that will be overwritten by assignment script. Actually that was the my point on this. If it is overwritten anyway, we might as well keep the default values as original one. So OpsScript doesn't have to overwritten values if it hasn't been changed in cloned workflow.

amaltaro:
That's the problem. It's _usually_ overwritten, but not mandatory. So we may end up in a funky situation. That's why I prefer not to have it at all during creation and let those stuff for assignment and/or default assignment values, just to make sure all the setter functions are properly called.

ticoann:
But if it is not overwritten it doesn't take the original value. (Just removed from the spec)?

ticoann:
replaced by Alan's clone patch

Issue #5410, opened 2014-10-17
Some files do not have the proper block_id association


amaltaro:
Every time we're finishing the draining of agents, we stumble on some files that were not injected into PhEDEx because the  block association is NULL (block_id).
The interesting part is that usually these files (and their block) is injected into DBS, so we have an inconsistent number of blocks and files between DBS and PhEDEx. I've not found the reason for that yet... :-(
@ticoann , this is what we've just chatted over gtalk.

PR #7814, opened 2017-04-26
Insert input files into dbsbuffer with in_phedex=1

### LINKED ISSUE #5223: Investigate why some files are not injected into PhEDEx
Creating this issue so we do not forget to investigate and fix it once and for all.
Everytime we need to redeploy a new agent, we find lots of files that are not injected into PhEDEx (it seems they are only injected into DBS), and then we need to run the well known phedex-fix.sh script.

This is the query that always returns lots of files not injected:

```
mysql> select lfn from dbsbuffer_file where in_phedex=0 and lfn not like '%unmerged%';
```

amaltaro:
Fixes #5223 (and possible #5410)

If I correctly understood the code :), these parentLFNs are the input datasets for workflows, that's why we mark them as status=GLOBAL. However, we should mark them in PhEDEx as well to make sure it's not going to bug us during agent draining.

PS.: I just noticed this `addParents` method is not called anywhere in the code...

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
@@ -13,13 +13,17 @@ from WMCore.WMBS.WMBSBase import WMBSBase
 
 
 class DBSBufferFile(WMBSBase, WMFile):
-    def __init__(self, lfn = None, id = -1, size = None,
-                 events = None, checksums = {}, parents = None, locations = None,
-                 status = "NOTUPLOADED", workflowId = None, prep_id = None):
+
+    def __init__(self, lfn=None, id=-1, size=None,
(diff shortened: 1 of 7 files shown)


ericvaandering:
test this please

ericvaandering:
test this please

amaltaro:
[Investigate why some files are not injected into PhEDEx](https://trello.com/c/CklWsYqw/36-investigate-why-some-files-are-not-injected-into-phedex)

amaltaro:
I tested this patch in my VM and it works fine, files from the input dataset get injected into `dbsbuffer_file` as `in_phedex=1` and what's produced is properly tagged in the database and injected later on on phedex/dbs.

Then after running all DMWM workflows, I see there is no a single file "not injected" into PhEDEx, finally :-D

PR #7816, opened 2017-04-26
Fix instances where indent is not a multiple of 4

ericvaandering:


ericvaandering:
@ticoann, please merge

PR #7818, opened 2017-04-26
New json templates for testing events per lumi functionality

amaltaro:
These two templates are using the same workflow from production, which was originally using 8_0_26 release, but I expect it to work without any issues. The idea is:
a) we run TaskChain_GENSIM.json and publish all the GEN-SIM-RAW data to DBS.
b) then we use TaskChain_EpL.json, with some modifications, to read the previously created dataset and make sure it works fine.

@ticoann I have just injected them into testbed, to check whether it works and whether the settings are Ok. Please go ahead and run them in your VM


ticoann:
Thanks Alan

PR #7819, opened 2017-04-26
Proxy timeleft return the smallest between proxy and voms

amaltaro:
Just noticed this issue in the FNAL testbed agent. I overlooked this voms timeleft detail that returns 0 if the proxy has a different expiration time from the voms one (which I don't really understand why it would work like that...).

@mmascher @emaszs CRAB is the heavier user of this class. Can you please check whether these changes are Ok for CRAB3?

Still need to be tested and backported.

### DIFF
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# pylint: disable=C0103,C0321
 """
 _Proxy_
 Wrap gLite proxy commands.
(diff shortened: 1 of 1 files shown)


amaltaro:
And it works just fine now.

amaltaro:
@ticoann can you review it please? if I don't hear anything from Marco or Emilis by tomorrow morning, I'll backport it and patch the production agents.

ticoann (approved):

ticoann:
Patch looks good to me.

ticoann:
test this please

emaszs:
The change should be fine for CRAB but do you think you could still print the message when the times are different? https://github.com/dmwm/WMCore/pull/7819/files#diff-e3ad52ab9a9b87ca6d434faca1ba5eb2L680

amaltaro:
Sure! I added the same debug logging level it had before, except by the unit that is just in secs now. Is that Ok? If so, @ticoann can you review it again and merge it please? Thanks!

emaszs:
I think that's perfectly fine, thanks Alan!

PR #7823, opened 2017-04-27
Clone only creation arguments and validate them

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/7846
@ticoann Seangchan, this is what I had in mind for cloning the creation arguments only. Which then can go through standard request creation validation as well.

The drawback is the fact that we validate all arguments again. Which is not that bad in cases we had a cmsweb upgrade that changed the data format/validate function for some of them :)
Second improvement to be done would be checking permissions before cloning the args, though we have basically 0 cases where the user cannot create a request.

If you think it would be good to have it in this testbed, let me know and I'll try to validate tomorrow and ask Lina for a new upgrade. Otherwise we can discuss and improve it for the next cycle.

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -31,6 +31,7 @@ ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST = \
      'Teams', 'TotalEstimatedJobs', 'TotalInputEvents', 'TotalInputFiles', 'TotalInputLumis',
      'TransientOutputModules', 'TrustPUSitelists', 'TrustSitelists', 'VoRole', '_id', '_rev']
 
+
 def initialize_request_args(request, config):
     """
     Request data class request is a dictionary representing
@@ -80,6 +81,7 @@ def initialize_resubmission(request_args, reqmgr_db_service):
     # to be used later on for spec validation
     request_args["OriginalRequestType"] = parent_args["RequestType"]
 
+
 def _replace_cloned_args(clone_args, user_args):
     """
     replace original arguments with user argument.
@@ -127,6 +129,7 @@ def initialize_clone(requestArgs, originalArgs, argsDefinition, chainDefinition=
 
     return cloneArgs
 
+
 def generateRequestName(request):
     currentTime = time.strftime('%y%m%d_%H%M%S', time.localtime(time.time()))
     seconds = int(10000 * (time.time() % 1.0))
@@ -134,8 +137,8 @@ def generateRequestName(request):
     request["RequestName"] = "%s_%s" % (request["Requestor"], request.get("RequestString"))
     request["RequestName"] += "_%s_%s" % (currentTime, seconds)
 
-def protectedLFNs(requestInfo):
 
+def protectedLFNs(requestInfo):
     reqData = RequestInfo(requestInfo)
     result = []
     if reqData.andFilterCheck(ACTIVE_STATUS_FILTER):
@@ -144,14 +147,16 @@ def protectedLFNs(requestInfo):
         for out in outs:
             dsn, ps, tier = out.split('/')[1:]
             acq, rest = ps.split('-', 1)
-            dirPath = '/'.join([ base, acq, dsn, tier, rest])
+            dirPath = '/'.join([base, acq, dsn, tier, rest])
             result.append(dirPath)
     return result
+
 
 class RequestInfo(object):
     """
     Wrapper class for Request data
     """
+
     def __init__(self, requestData):
         self.data = requestData
 
@@ -258,4 +263,3 @@ class RequestInfo(object):
         # cannot determin whether AgentJobInfo is cleaned or not when 'AgentJobInfo' Key doesn't exist
         # Maybe JobInformation is not included but since it requested by above status assumed it returns True
         return True
-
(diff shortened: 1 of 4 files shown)


amaltaro:
And I forgot to mention, it's totally untested :)

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -92,15 +92,12 @@ def _replace_cloned_args(clone_args, user_args):
...
     """
     Initialize a Clone arguments by inheriting and overwriting argument from OriginalRequest
     """
-    requests = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
-    clone_args = requests.values()[0]
-    # TODO: need to validate new overwrite request_args here since it will skip the clone_args validtaiton
-    _replace_cloned_args(clone_args, request_args)
-    clone_args = {k: v for k, v in clone_args.iteritems() if k not in ARGS_TO_REMOVE_FROM_ORIGINAL_REQUEST}
+    _replace_cloned_args(originalArgs, request_args)
+    clone_args = {k: v for k, v in originalArgs.iteritems() if k in argsDefinition}
```
amaltaro: Actually, we need the same mechanism as in:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkloadTools.py#L292

there must have a smarter way to handle it...

amaltaro:
test this please

amaltaro:
test this please

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -840,14 +840,16 @@ def factoryWorkloadConstruction(self, workloadName, arguments):
...
         """
-        if 'OriginalRequestName' not in arguments:
-            self.masterValidation(schema=arguments)
+        if arguments.get("RequestType") == "Resubmission":
+            # It accepts both creation & assignment, thus skip master validation
             self.validateSchema(schema=arguments)
-        elif arguments.get("RequestType") == "Resubmission":
-            # Don't validate schema for cloned workflow
-            # But Resubmission workflow has to be validated
+        elif clone:
```
ticoann: should we check clone before checking the Resubmission? In case resubmission request is cloned.

amaltaro: I don't expect clones of Resubmission, but that's a good point and I'll fix it tomorrow :)

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -92,15 +92,32 @@ def _replace_cloned_args(clone_args, user_args):
...
+        if k in argsDefinition:
+            clone_args[k] = v
+        elif requestType == "ReReco" and k.startswith("Skim"):
+            clone_args[k] = v
+        elif requestType == "StepChain" and k.startswith("Step"):
+            clone_args[k] = v
+        elif requestType == "TaskChain" and k.startswith("Task"):
+            clone_args[k] = v
+        elif requestType == "Resubmission" and k[:4] in ["Skim", "Step", "Task"]:
+            clone_args[k] = v
```
ticoann: why do we need this if, elif statement for all this conditions?

Can we just do following?
```
_replace_cloned_args(originalArgs, request_args)
```

amaltaro: Because I wanted to copy only creation & defined arguments, nothing else. The drawback is that we need to take into account those floating arguments from the chained/Rereco specs

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
When I thought it was ready, I just noticed we should do the same with Task/Step dictionaries. I'm going to update it

amaltaro:
Ok, validating the user args only is not going to work. It will require too much code duplication to validate Steps/Tasks configuration. E.g., StepChain has some important step validation (which also depends on whether the step is a generator or not), given all these subtle details, I think it's MUCH cleaner if we simply:
a) build the new spec
b) run the usual validation (masterValidation + validateSchema).

It should add a second more for clone creation, but it will save lots of complexity.

src/python/WMCore/ReqMgr/DataStructs/Request.py
```diff
@@ -88,31 +91,53 @@ def _replace_cloned_args(clone_args, user_args):
...
-    # TODO: need to validate new overwrite request_args here since it will skip the clone_args validtaiton
+    chainPattern = r'(Task|Step)\d{1,2}'
+    clone_args = {}
+    for topKey, topValue in originalArgs.iteritems():
+        if topKey in argsDefinition:
+            clone_args[topKey] = topValue
+        elif topKey.startswith(("Skim", "Step", "Task")):
+            # accept floating args from ReReco, StepChain and TaskChain
+            if re.match(chainPattern, topKey):
+                for innerKey in topValue:
```
amaltaro: btw, I don't like it. There is likely a one line code that does it all, I just need to figure it out :)

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
It passes all my tests, including some juggling with StepChain/TaskChain. Anyways, it needs to be validated again in testbed, especially with workflows injected by McM (usually unwanted args).

PR #7832, opened 2017-05-03
Allow TagCollector to set the logger

mmascher:


mmascher:
See the related thread named "default alarms' routing for vocmsweb/backend" in cms-service-webtools

ticoann:
test this please

PR #7836, opened 2017-05-04
Don't validate Resubmission specs

### LINKED ISSUE #7835: Resubmission crashes on "old" requests with unknown key args
As resubmission is handled now, it will break when the original request has arguments that it "should not", because we simply copy those over to the new request:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/Request.py#L74

issue reported here:
https://github.com/CMSCompOps/WmAgentScripts/issues/222

amaltaro:
Fixes #7835 (if you allow me to call it a fix :))

@ticoann this is the fix I suggest to be pushed into cmsweb as a critical patch release.
Then we can properly implement and test Resubmission creation/validation for the next cycle.
What do you think? If you agree, we can ask the HTTP team to push it to testbed even today and to production tomorrow morning.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -843,11 +843,7 @@ class StdBase(object):
         if 'OriginalRequestName' not in arguments:
             self.masterValidation(schema=arguments)
             self.validateSchema(schema=arguments)
-        elif arguments.get("RequestType") == "Resubmission":
-            # Don't validate schema for cloned workflow
-            # But Resubmission workflow has to be validated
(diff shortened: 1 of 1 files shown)


ticoann (approved):

ticoann:
I will create the branch patch and new tag then

amaltaro:
Thanks, Seangchan! I should've listened to you when you said "why don't we skip the resubmission validation?!?" :-)
BTW, Jenkins may fail since Eric is pushing a new container in.

ticoann:
Did I say that? 😄  I think maybe we should treat resubmission same way as clone for validation.

ericvaandering:
So little faith!

PR #7843, opened 2017-05-08
Use StdBase workload arguments as a baseline for fetched workflow ones

### LINKED ISSUE #7840: Retrieve assignment default arguments for assignment via web UI
Since we separated the creation from assignment arguments, the ReqMgr2 assignment form no longer shows the default assignment values for some of those arguments (merge and block args in general).

We should change it such that:
1) it retrieves the assignment arguments definition, as in:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1042
2) it retrieves the request dictionary (it already does)
3) then it overrides the default arguments by the request one
4) fill in the form with the pre-selected values.

vkuznet:
Fixes #7840


amaltaro:
I'm closing this one since the other PR has been merged. Thanks, Valentin!

PR #7848, opened 2017-05-12
Temp fix to get Held jobs removed from condor - old agent

amaltaro:
Temporary fix for 1.1.0._wmagent branch only. Not to be merged!

While draining production agents, I've been seeing several jobs in condor status 5 (Held) that are in the pool for several days. Apparently they are not getting removed because `status_time` value in `bl_runjob` is None, thus JobStatusLite never removes them.

I've seen Pending jobs lingering around as well for more than 1M secs.

Now why it's None is another story. I have an old branch where I was working on this job tracking and updates to boss air and wmbs tables. I'll have to revive and resume working on that.


amaltaro:
This actually doesn't work. Jobs are not active in boss air anymore. I'm creating an issue instead with comments.

PR #7850, opened 2017-05-13
Load default assignment values in the Gui

### LINKED ISSUE #7840: Retrieve assignment default arguments for assignment via web UI

amaltaro:
Fixes #7840
It replaces #7843 , I still want to fix the assign page as well.
If assignment values are not present in the request, then fetch the default ones from the assignment dictionary definition.
Also set Team attribute default value to `''` such that we don't see a None value in the web UI.

Valentin, besides the import, there was an issue with the prop override that was done too late. It has to be done before filtering out and handling the visible attr. 

@vkuznet please review, basic tests look good.

### DIFF
--- a/src/python/WMCore/ReqMgr/Web/ReqMgrService.py
+++ b/src/python/WMCore/ReqMgr/Web/ReqMgrService.py
@@ -7,42 +7,42 @@ __author__ = "Valentin Kuznetsov"
 """
 from __future__ import print_function
 
+import collections
+import json
 # system modules
 import os
-import sys
-import json
-import time
 import pprint
-import collections
+import sys
+import time
 from datetime import datetime
+from copy import deepcopy
 
 # cherrypy modules
 import cherrypy
+from cherrypy import config as cherryconf
 from cherrypy import expose, response, tools
 from cherrypy.lib.static import serve_file
-from cherrypy import config as cherryconf
 
-# ReqMgrSrv modules
-from WMCore.ReqMgr.Web.tools import exposecss, exposejs, TemplatedPage
-from WMCore.ReqMgr.Web.utils import json2table, json2form, genid, checkargs, tstamp, sort, reorder_list
-from WMCore.ReqMgr.Utils.url_utils import getdata
-from WMCore.ReqMgr.Tools.cms import releases, architectures, dashboardActivities
-from WMCore.ReqMgr.Tools.cms import web_ui_names, SITE_CACHE, PNN_CACHE
-from WMCore.ReqMgr.Tools.cms import lfn_bases, lfn_unmerged_bases
-from WMCore.ReqMgr.Tools.cms import site_white_list, site_black_list
-
-# WMCore modules
-from WMCore.ReqMgr.Utils.Validation import get_request_template_from_type
-from WMCore.ReqMgr.DataStructs.RequestStatus import get_modifiable_properties, get_protected_properties
-from WMCore.Services.LogDB.LogDB import LogDB
-from WMCore.ReqMgr.Web.utils import gen_color
 # import WMCore itself to determine path of modules
 import WMCore
-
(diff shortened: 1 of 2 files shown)


src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -521,10 +525,14 @@ def fetch(self, rid):
                 transitions.remove(status)
             visible_attrs = get_modifiable_properties(status)
             filterout_attrs = get_protected_properties()
-
             # extend filterout list with "RequestStatus" since it is passed separately
             filterout_attrs.append("RequestStatus")
 
+            args = self.assignArgs
+            for key, defin in self.assignArgs.iteritems():
```
vkuznet: Alan, replace iteritems() with items() which is only supported method in python 3.X

amaltaro:
test this please

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
amaltaro: I think Eric had a good argument to call it `iteritems()` instead of `items`, keeping in mind the python 3 migration. But I'll update it

amaltaro:
test this please

amaltaro:
test this please

vkuznet (approved):
Looks good to me

amaltaro:
@vkuznet would you mind reviewing it again, please?
I had to default `Team` to empty string otherwise we would see `None` in the web UI.
I also took the opportunity to remove some hard-coded arguments from this class.

Tests look good.

amaltaro:
test this please

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -521,10 +524,13 @@ def fetch(self, rid):
                 transitions.remove(status)
             visible_attrs = get_modifiable_properties(status)
             filterout_attrs = get_protected_properties()
-
             # extend filterout list with "RequestStatus" since it is passed separately
             filterout_attrs.append("RequestStatus")
 
+            for key, val in self.assignArgs.items():
+                if key not in doc or not doc.get(key):
```
vkuznet: Alan, I'm not 100% sure if "not doc.get(key)" will work, the problem is what do you want from it. If empty string or empty list are acceptable arguments, then the conditions will not give you what you want and it will overwrite doc[key] value with defaults. I think a proper way should be:
```
if not doc.get(key, None)
```
which does both tests, i.e. check that key in dict and if not it returns None which is not a proper data-type for assignment args. Compare this to empty string or empty list which may be proper values (and data-types) for assigned args.

vkuznet (approved):

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
amaltaro: Oh yes! I oversaw the first logic in the if. It will be `if not doc.get(key)`, which is the same as you suggested (None is returned by default in case key is not there).

And what I want is to go into the `if ` for cases like `''`, `[]`, `None`, `{}`. Such that we set it to the assignment default (which is going to be the same for most of the cases). I'm updating this patch soon.

amaltaro:
test this please

amaltaro:
Thanks, Valentin. Patch was updated and jenkins are happy now. I'm going to merge it within a few minutes.

PR #7853, opened 2017-05-15
Dictionary with all the create arguments for each request type

amaltaro:
Json file containing all the request arguments supported by each request type (including a few important fields of their definition).
There is also a json for StepChain and TaskChain inner chain (for both generator and non-generator case).
ReReco Skim stuff is a bit complicated too :(

I built these json with this script:
https://github.com/amaltaro/ProductionTools/blob/master/makeRequestSpecs.py

@ericvaandering @ticoann I'm not sure whether there is a better location for these files. Let me know if that's the case.

### DIFF
--- a/doc/createSpecs/DQMHarvest_createSpec.json
+++ b/doc/createSpecs/DQMHarvest_createSpec.json
@@ -1,0 +1,302 @@
+{
+    "AcquisitionEra": {
+        "default": null, 
+        "optional": false, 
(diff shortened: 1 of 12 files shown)


amaltaro:
test this please

amaltaro:
these failures make no sense, since this PR has only new json documents. I'm merging it to complete the wiki about request creation

PR #7863, opened 2017-05-22
Remove some default values for required spec arguments

### LINKED ISSUE #7861: MonteCarlo request RequestNumEvents default to 1000
as discussed in https://github.com/CMSCompOps/WmAgentScripts/pull/221 operation have had a hard time with clones defaulting to 1000 event to be processed.
Looks like that parameter should be made mandatory at creation time and fail creation if not provided, never fall back to 1000 by default.
@amaltaro this is a show-stopper for operation
FYI @prozober

amaltaro:
Fixes #7861 
I reviewed all the specs and the only additional thing we could remove is the default settings for the splitting units (FilesPerJob, LumisPerJob), but I'll leave it alone...

I'm sure it will impact jenkins, so let's see which tests will fail.


amaltaro:
test this please

amaltaro:
These failures are probably temporary, since I cannot reproduce then in my local node.
I also tested workflow injection and it didn't get affected by these changes. So I'm merging it and then we can take time validating it further this week and the next one.

amaltaro:
test this please

PR #7867, opened 2017-05-23
Fix up some typos

ericvaandering:
@yuyiguo and I were looking at this code while I was explaining what Jenkins did. Seemed like a good idea to fix the obvious problems.


ericvaandering:
@amaltaro please merge. Unit test failure is unrelated

PR #7868, opened 2017-05-23
add option to disable xrdcp write recovery

hufnagel:
Seems EOS has a write recovery feature that does really bad things sometimes (xrdcp exit code 0 only to delete the file later). Add an option to disable this, turn on that option for LogCollect stageout to EOS (but not Castor). For normal stageout this then needs to be activated in the SITECONF.

There is ongoing discussion on this, so I expect some revisions will be needed, just wanted to get it out here for others to see and comment on.

### DIFF
--- a/src/python/WMCore/Storage/Backends/XRDCPImpl.py
+++ b/src/python/WMCore/Storage/Backends/XRDCPImpl.py
@@ -62,6 +62,7 @@ class XRDCPImpl(StageOutImpl):
         parser = argparse.ArgumentParser()
         parser.add_argument('--cerncastor', action='store_true')
         parser.add_argument('--old', action='store_true')
+        parser.add_argument('--disablewriterecovery', action='store_true')
         args, unknown = parser.parse_known_args(options.split())
 
         copyCommandOptions = ' '.join(unknown)
@@ -110,6 +111,9 @@ class XRDCPImpl(StageOutImpl):
                 if all(os.path.isfile(initFile) for initFile in initFiles):
                     for initFile in initFiles:
                         copyCommand += "source %s\n" % initFile
+
+        if args.disablewriterecovery and not args.cerncastor:
+            copyCommand += "env XRD_WRITERECOVERY=0 "
 
         copyCommand += "%s --force --nopbar " % xrdcpExec
 
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -79,7 +79,7 @@ class LogCollect(Executor):
         # hardcode CERN EOS T2_CH_CERN stageout parameters
         eosStageOutParams = {}
         eosStageOutParams['command'] = overrides.get('command', "xrdcp")
-        eosStageOutParams['option'] = overrides.get('option', "")
+        eosStageOutParams['option'] = overrides.get('option', "--disablewriterecovery")
         eosStageOutParams['phedex-node'] = overrides.get('phedex-node', "T2_CH_CERN")
         eosStageOutParams['lfn-prefix'] = overrides.get('lfn-prefix', "root://eoscms.cern.ch//eos/cms")
 


hufnagel:
test this please

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -111,6 +112,9 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
                     for initFile in initFiles:
                         copyCommand += "source %s\n" % initFile
 
+        if args.disablewriterecovery and not args.cerncastor:
+            copyCommand += "env XRD_WRITERECOVERY=0 "
```
amaltaro: newline at the end?

hufnagel: no, why would there ? env is a command pre-processort, it changes the environment for the command it executes.

amaltaro: I thought this variable would be set for the sub-shell that gets created when running this stage out script. Instead, you have to sort of pass the command to be executed as an "argument". Thanks

amaltaro:
@hufnagel since you're using `parse_known_args` method, that means we could update the CERN site-local-config and run this test with that change in place, just to make sure we have the expected behaviour. What do you think? 

Are you able to update the site-local-config? or who is in charge of that?

hufnagel:
The problem is that unknown options are presumed to be xrdcp options and are used as such. Which means xrdcp would fail with an unknown option for all current production agents. Which means we cannot really edit the site-local-conf until all the agents are patched with the new code that knows about this option.

For testing just change line 115 to:

if not args.disablewriterecovery and not args.cerncastor:

Should be good enough, shouldn't it ?

And I can modify the site-local-config once all the agents are patched, takes a few minutes to make the change and half an hour to an hour to be propagated to the worker nodes.

amaltaro:
Given the fragility of these options, I wonder whether this is the best approach then... 
Having to sync WMCore stage out plugins with site changes is far from ideal, if this is the only option, then be it.

hufnagel:
Two options I can see:

1) We get rid of passing the unknown part of the option to xrdcp and just drop it.

Means if a site finds itself in the situation where they need special options in their xrdcp command, they'll need to wait until we add support for them in the xrdcp stageout plugin.

2) Clearly separate the part of options that should be used for xrdcp and in the plugin.

Something like renaming the current options --plugin-old --plugin-castorcern and then --plugin-disablewriterecovery. Any option that starts with plugin would not be passed on to xrdcp.

Doesn't help this time around (you still need to get that code deployed first), but would help if we ever have to do this again and add another option.

amaltaro:
I'm in favor of 2., though I'd rename it to --wma-old and etc just to be clear that it belongs to stage out specific settings and not to the executable options itself.

BTW, I fail to find `--old` option being used anywhere. Do you remember why we needed that? I guess for FNAL, but according to their site-local-config, that is gone.

amaltaro:
Ok, after tweaking that file locally I can see successful stage out using command like
```
LOCAL_SIZE=`stat -c%s "/srv/job/WMTaskSpace/cmsRun1/LHEoutput.root"`
echo "Local File Size is: $LOCAL_SIZE"
env XRD_WRITERECOVERY=0 xrdcp --force --nopbar --cksum adler32:7d9a64a3  "/srv/job/WMTaskSpace/cmsRun1/LHEoutput.root"  "root://eoscms.cern.ch//eos/cms/store/unmerged/DMWM_Test/SeesawTypeIII_SIGMAplusSIGMAminusHH_M-1180_13TeV-madgraph-pythia8/LHE/MonteCarlo_eff_HG1706_Validation_TEST_Alan_v112-v11/00000/1E69BDCC-DD45-E711-95FF-FA163E6AD580.root"
REMOTE_SIZE=`xrdfs 'eoscms.cern.ch' stat '/eos/cms/store/unmerged/DMWM_Test/SeesawTypeIII_SIGMAplusSIGMAminusHH_M-1180_13TeV-madgraph-pythia8/LHE/MonteCarlo_eff_HG1706_Validation_TEST_Alan_v112-v11/00000/1E69BDCC-DD45-E711-95FF-FA163E6AD580.root' | grep Size | sed -r 's/.*Size:[ ]*([0-9]+).*/\1/'`
echo "Remote File Size is: $REMOTE_SIZE"
echo "Local File Checksum is: 7d9a64a3"
REMOTE_XS=`xrdfs 'eoscms.cern.ch' query checksum '/eos/cms/store/unmerged/DMWM_Test/SeesawTypeIII_SIGMAplusSIGMAminusHH_M-1180_13TeV-madgraph-pythia8/LHE/MonteCarlo_eff_HG1706_Validation_TEST_Alan_v112-v11/00000/1E69BDCC-DD45-E711-95FF-FA163E6AD580.root' | grep -i adler32 | sed -r 's/.*[adler|ADLER]32[ ]*([0-9a-fA-F]{8}).*/\1/'`
echo "Remote File Checksum is: $REMOTE_XS"
if [ $REMOTE_SIZE ] && [ $REMOTE_XS ] && [ $LOCAL_SIZE == $REMOTE_SIZE ] && [ '7d9a64a3' == $REMOTE_XS ]; then exit 0; else echo "Error: Size or Checksum Mismatch between local and SE"; xrdfs 'eoscms.cern.ch' rm /eos/cms/store/unmerged/DMWM_Test/SeesawTypeIII_SIGMAplusSIGMAminusHH_M-1180_13TeV-madgraph-pythia8/LHE/MonteCarlo_eff_HG1706_Validation_TEST_Alan_v112-v11/00000/1E69BDCC-DD45-E711-95FF-FA163E6AD580.root ; exit 60311 ; fi
```

amaltaro:
@hufnagel to avoid changing it (stage out + site config) today and again in the next week or so, I've created a new PR https://github.com/dmwm/WMCore/pull/7887 with your suggestion 2.

Can you review it please? If you agree, I'll test it by tonight/tomorrow and then we push it to all the agents.

amaltaro:
Dirk, I'm closing this PR to avoid confusion. New one was made/reviewed/tested #7887

PR #7870, opened 2017-05-24
LumisPerJob doesnt belong to this template

amaltaro:


amaltaro:
keeping it open in case there are further minor updates.

amaltaro:
Merging these templates. @ticoann it won't break, but you may want to clone WMCore again before injecting your tests. Also a minor update to the inject-test script.

PR #7615, opened 2017-02-10
Add thread to check drain status + AgentStatusWatcher unittest

goughes:
Adding the skeleton thread for checking the drain status.  Need to figure out what database connections are needed and what drain info to write where.

PR #7871, opened 2017-05-24
No longer fetch/feed Resubmission args in WMStats to ReqMgr

### LINKED ISSUE #7872: Stop providing resubmission args from wmstats
Because we now clone the whole request in ReqMgr anyways, so no need to fetch all those arguments in javascript as well.

amaltaro:
Fixes #7872 
Since Resubmission is basically a clone workflow now (we copy the original creation + assignment requests into a new "truncated" request), there is no more need to load this info in WMStats javascript, do some parsing, and posting it to reqmgr. We need only the very basics, all the rest comes from the resubmission handling on ReqMgr.

Still need to be tested. I don't know if any of those were also used for the small pop-up when creating an ACDC via wmstats.

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
+++ b/src/couchapps/WMStats/_attachments/js/ViewModels/WMStats.ViewModel.js
@@ -411,6 +411,7 @@ WMStats.ViewModel = (function (){
         resubmission.formatResubmissionData = function(reqMgrRequest) {
             var summary = {};
             // from resubmission keys
+            summary.RequestType = "Resubmission";
             summary.OriginalRequestName = resubmission.keys().requestName;
             summary.InitialTaskPath = resubmission.keys().task;
             if (resubmission.keys().acdcURL) {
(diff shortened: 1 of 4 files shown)


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Failures in the unittests are not related to this PR.
@ericvaandering I know you're fixing a problem with containers shutdown, maybe these failures are related? For all these tests that I looked at, they fail talking to mysql (maybe it's stopped too soon?):
```
WMInitException
Message: Critical error while attempting to delete entire DB!
(OperationalError) (2002, 'Can\'t connect to local MySQL server through socket \'/home/dmwm/unittestdeploy/wmagent/current/install/mysql/logs/mysql.sock\' (2 "No such file or directory")') None NoneTraceback (most recent call last):
```
Can you please have a look? 
The error message is different, but this PR #7615 is also facing issues with the local relational db.

amaltaro:
Merging it to cut a new tag. All tests look Ok in my VM.

ericvaandering:
This often means a problem with the container if its repeatable. I started the process to rebuild them.

PR #7874, opened 2017-05-25
No need to promote to float with future division

ericvaandering:
Pylint --3k is warning about all these divisions. No idea why, but there is no reason to promote a variable to float to ensure floating point division anymore


ericvaandering:
@amaltaro please merge

amaltaro:
awesome!

PR #7875, opened 2017-05-25
prevent exceptions printout in TempDirectory

hufnagel:
Restore protections. The exception caught here has nothing to do with the actual call, it happens because the system is in shutdown and there is no shutil module available anymore. Therefore do not restrict the exceptions catch, it needs to catch everything.

### DIFF
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -577,5 +577,5 @@ class TempDirectory(object):
         try:
             # it'll likely fail, but give it a try
             shutil.rmtree(self.dir, ignore_errors=True)
-        except (IOError, OSError):
+        except Exception:
             pass


src/python/WMCore/Services/Requests.py
```diff
@@ -577,5 +577,5 @@ def __del__(self):
         try:
             # it'll likely fail, but give it a try
             shutil.rmtree(self.dir, ignore_errors=True)
-        except (IOError, OSError):
+        except:
```
ericvaandering: ```except Exception``` splitting hairs, but this is allowed by our pylint setup, specifying nothing is not. We assume you are being especially lazy in the latter case.

hufnagel: Well, I am not lazy, this is intentional. But ok, I understand the argument.

amaltaro:
Blindly passing any exceptions that occur is quite dangerous, but I think for this specific case it's Ok.

PR #7878, opened 2017-05-29
Set status_time when jobs get submitted in BossAir

### LINKED ISSUE #7849: Jobs completed in BossAir but still in condor pool
While draining the 1.1.0 agent, I've seen two major issues related to the job tracking:
1) pending condor jobs still being tracked by JobStatusLite even after the Timeout has exceeded (they were pending for more than 1M secs). The reason seems to be from the fact that they don't have `status_time` in the `bl_runjob` table.
2) held condor jobs not being tracked anymore, but still in condor for several days. Apparently because their `status` in `bl_runjob` is 0, meaning the job is considered completed and there is no more tracking.

This happened in several agents! I'll resume working on it soon before it causes more damage...

amaltaro:
Fixes #7849 (well, the held jobs still need to be addressed).

Problem being the lack of `status_time` on newly created jobs in BossAir (which happens after calling condor submit). I still need to test to make sure that was the only reason. Weird that we've found it only now though...

### DIFF
--- a/src/python/WMCore/BossAir/MySQL/NewJobs.py
+++ b/src/python/WMCore/BossAir/MySQL/NewJobs.py
@@ -5,7 +5,7 @@ _NewJobs_
 MySQL implementation for creating a new job
 """
 
-
+import time
 from WMCore.Database.DBFormatter import DBFormatter
 
 class NewJobs(DBFormatter):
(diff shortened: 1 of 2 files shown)


hufnagel (approved):
Looks good. Not sure what it actually fixes, but having the status_time information there can't possibly hurt, so approving it.

amaltaro:
It fixes the JobStatusLite state timeout mechanism, this line basically:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/StatusPoller.py#L116

otherwise it never kill jobs that are idle since ever.

amaltaro:
test this please

PR #7882, opened 2017-05-31
Automatically remove jobs in Held status after 10min

### LINKED ISSUE #7881: Implement PeriodicRemove expression for held jobs
It could either be set in the condor_schedd configuration or be added to the job classads.
We better ad it to every job such that we control the exact expression.

According to Stefano, it's just a matter of defining that expression:
https://cms-logbook.cern.ch/elog/GlideInWMS/5393

no further ads are needed.
For reference, this is how it's done for CRAB:
https://github.com/dmwm/CRABServer/blob/f8770011dcd03febc7218ecbc49e3b4b9b0c7684/src/python/TaskWorker/Actions/DagmanSubmitter.py#L488

though in our case I think it should be (plain english) `job is on Hold && currentStatusTime > 10minutes` or so.

amaltaro:
Fixes #7881

Since it's simpler than I thought, why not to fix it right now :)
I'm not sure we actually need the remove reason, since the job was already taken care by the agent, but anyways it won't hurt.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -461,6 +461,10 @@ class SimpleCondorPlugin(BasePlugin):
 
         ad['JobLeaseDuration'] = classad.ExprTree('isUndefined(MachineAttrMaxHibernateTime0) ? 1200 : MachineAttrMaxHibernateTime0')
 
+        ad['PeriodicRemove'] = classad.ExprTree('( JobStatus =?= 5 ) && ( time() - EnteredCurrentStatus > 10 * 60 )')
+        removeReasonExpr = 'PeriodicRemove ? "Job automatically removed for being in Held status" : ""'
+        ad['PeriodicRemoveReason'] = classad.ExprTree(removeReasonExpr)
+
         # Required for global pool accounting
         ad['AcctGroup'] = self.acctGroup
         ad['AcctGroupUser'] = self.acctGroupUser


src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -461,6 +461,9 @@ def getClusterAd(self):
 
         ad['JobLeaseDuration'] = classad.ExprTree('isUndefined(MachineAttrMaxHibernateTime0) ? 1200 : MachineAttrMaxHibernateTime0')
 
+        ad['PeriodicRemove'] = classad.ExprTree('( JobStatus =?= 5 ) && ( time() - EnteredCurrentStatus > 10 * 60 )')
+        ad['PeriodicRemoveReason'] = "Job automatically removed for being in Held status"
```
belforte: maybe only set the PeriodicRemoveReason when the job is actually removed ? Avoid confusing someone who looks at job ads one day and finds this reason set. The main reason to set a PeriodicRemoveReason is in case the PeriodicRemove is an OR or various conditions and you want to record which one triggered it, e.g. in CRAB we want to know if job was removed because it was using too much time, or too much memory or...

amaltaro:
Ok, I've just ran a test by setting a (running) job into hold status (condor_hold) and while the job was on hold, these were the Remove ads:
```
PeriodicRemove = ( JobStatus =?= 5 ) && ( time() - EnteredCurrentStatus > 10 * 60 )
PeriodicRemoveReason = PeriodicRemove ? "Job automatically removed for being in Held status" : ""
```

while after > 10min the job was automatically removed from condor and its Remove ads are:
```
RemoveReason = "Job automatically removed for being in Held status"
```

So this fix looks Ok, though a confirmation from @bbockelm won't hurt.

Now the same job in question is reported in wmstats as `NoJobReport (Exit Code: 99303)`, which is one of the issues reported here #7849, and that needs to be fixed!

amaltaro:
Since tests were ok, merging it for the next release candidate. We no longer need to bother about killing held jobs. This means we no longer need this silly cronjob btw:
https://github.com/dmwm/WMCore/blob/master/bin/deploy-wmagent.sh#L321

PR #7887, opened 2017-05-31
Add option to disable xrdcp write recovery

amaltaro:
Possible replaces #7868
Changes are:
* support a new wmagent-specific argument to disable XRD_WRITERECOVERY (workaround for a problem at T2_CH_CERN)
* rename wmagent-specific stage out arguments
* make sure not to add those arguments (options) to the actual stage out command
* remove support for xrdcp-old

### DIFF
--- a/src/python/WMCore/Storage/Backends/XRDCPImpl.py
+++ b/src/python/WMCore/Storage/Backends/XRDCPImpl.py
@@ -60,10 +60,13 @@ class XRDCPImpl(StageOutImpl):
             options = ''
 
         parser = argparse.ArgumentParser()
-        parser.add_argument('--cerncastor', action='store_true')
-        parser.add_argument('--old', action='store_true')
+        parser.add_argument('--wma-cerncastor', action='store_true')
+        parser.add_argument('--wma-old', action='store_true')
+        parser.add_argument('--wma-disablewriterecovery', action='store_true')
         args, unknown = parser.parse_known_args(options.split())
 
+        # strip out WMAgent specific options
+        unknown = [option for option in unknown if not option.startswith('--wma-')]
         copyCommandOptions = ' '.join(unknown)
 
         copyCommand = ""
@@ -74,13 +77,13 @@ class XRDCPImpl(StageOutImpl):
             remotePFN, localPFN = targetPFN, sourcePFN
             copyCommand += "LOCAL_SIZE=`stat -c%%s \"%s\"`\n" % localPFN
             copyCommand += "echo \"Local File Size is: $LOCAL_SIZE\"\n"
-            if args.cerncastor:
+            if args.wma_cerncastor:
                 targetPFN += "?svcClass=t0cms"
 
         useChecksum = (checksums != None and 'adler32' in checksums and not self.stageIn)
 
         xrdcpExec = "xrdcp"
-        if args.old:
+        if args.wma_old:
             xrdcpExec = "xrdcp-old"
 
         # check if xrdcp(-old) and xrdfs are in path
@@ -110,6 +113,9 @@ class XRDCPImpl(StageOutImpl):
                 if all(os.path.isfile(initFile) for initFile in initFiles):
                     for initFile in initFiles:
                         copyCommand += "source %s\n" % initFile
+
+        if args.wma_disablewriterecovery:
+            copyCommand += "env XRD_WRITERECOVERY=0 "
 
         copyCommand += "%s --force --nopbar " % xrdcpExec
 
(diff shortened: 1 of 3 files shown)


hufnagel (commented):
Hm, one very important thing you forgot. You nee to add a filter so that any unknown option that starts with --wma is not added to copyCommandOptions.

Without that we have the same mess again next time we want to add an option.

amaltaro:
> We need the --old option since Fermilab is still using it (in their old stageout plugin, but I am trying for them to change it). We also need to keep supporting both --old and --wma-old for now since a few SITECONF for opportunistic computing (NERSC for instance) used that to stageout to FNAL.

Ok, I'll add `--old` support back then.

> wait for old workflows with old sandboxes to clear out of the system

well, this makes it almost impossible to accomplish. It's a PITA to track all the workflows and make sure old sandboxes are gone.

How about we make two different developments them:
* master branch: this patch as is (+ old support)
* wmagent branch - to patch all the agents: we accept both `--wma-old` and `--old` for all those wmagent arguments (so 6 different arguments would be accepted).

Then we can patch all the agents and update any site-local-configs to --wma-XXX. How does it sound?

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -60,10 +60,12 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
...
 
         parser = argparse.ArgumentParser()
-        parser.add_argument('--cerncastor', action='store_true')
-        parser.add_argument('--old', action='store_true')
+        parser.add_argument('--wma-cerncastor', action='store_true')
+        parser.add_argument('--wma-disablewriterecovery', action='store_true')
         args, unknown = parser.parse_known_args(options.split())
 
+        # strip out WMAgent specific options
+        unknown = set(unknown) - set(['--wma-cerncastor', '--wma-disablewriterecovery'])
```
amaltaro: filter is here btw.

hufnagel: No, it's not. The filter I am talking about is on "--wma-*"

hufnagel: Future not yet know parameters we might want to add later. We want to be able to just add them to site-local-config without having to worry that older code will pick them up.

hufnagel:
OTOH, it doesn't really hurt to leave both --old and --wma-old in there for now (even for a few months or a year). Less possibility to mess up between agent branch and master branch. Up to you though since you have to manage that.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
amaltaro: Oh, I see!

amaltaro:
It does hurt! All these temp code usually becomes permanent (and source of confusion) :-)
Giving it more thought, we'll need to patch the sandboxes anyways. Otherwise workflows already in the system will break. 

Thus we patch agents + sandboxes today and tomorrow you update the site-local-configs, just to give some time for the running jobs to finish.

amaltaro:
Thanks, Dirk! Patch was update. Basically added back the '-old' support and filter any '--wma-*' options. Im' going to create the wmagent branch version and try to run a few tests over the night.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -74,16 +77,16 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
...
             copyCommand += "echo \"Local File Size is: $LOCAL_SIZE\"\n"
-            if args.cerncastor:
+            if args.wma_cerncastor:
                 targetPFN += "?svcClass=t0cms"
 
         useChecksum = (checksums != None and 'adler32' in checksums and not self.stageIn)
 
         xrdcpExec = "xrdcp"
-        if args.old:
+        if args.wma_old:
```
hufnagel: Shouldn't that be ?

if args.old or args.wma_old:

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -111,6 +114,9 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
                     for initFile in initFiles:
                         copyCommand += "source %s\n" % initFile
 
+        if args.wma_disablewriterecovery and not args.wma_cerncastor:
```
hufnagel: I know I wrote that line, but I think we can remove the "and not args.wma_cerncastor". It's redundant and makes the code harder to read. We only use castor mode from LogCollect anyways where we control the options we set.

hufnagel: I know I wrote that line, but I think we should remove the check on wma_cerncastor here. It's redundant since the only place where we use castor mode is in LogCollect and we control that code.

hufnagel (changes requested):

amaltaro:
test this please

PR #7888, opened 2017-05-31
Add option to disable xrdcp write recovery - wmagent branch

amaltaro:
wmagent version of #7887
Not exactly the same, since we support both old and new options here.
This one is also worth a review, Dirk :-)

### DIFF
--- a/src/python/WMCore/Storage/Backends/XRDCPImpl.py
+++ b/src/python/WMCore/Storage/Backends/XRDCPImpl.py
@@ -62,8 +62,14 @@ class XRDCPImpl(StageOutImpl):
         parser = argparse.ArgumentParser()
         parser.add_argument('--cerncastor', action='store_true')
         parser.add_argument('--old', action='store_true')
+        parser.add_argument('--wma-cerncastor', action='store_true')
+        parser.add_argument('--wma-old', action='store_true')
+        parser.add_argument('--wma-disablewriterecovery', action='store_true')
         args, unknown = parser.parse_known_args(options.split())
 
+        # strip out WMAgent specific options
+        unknown = [option for option in unknown if not option.startswith('--wma-')]
+        unknown = set(unknown) - {'--cerncastor', '--old'}
         copyCommandOptions = ' '.join(unknown)
 
         copyCommand = ""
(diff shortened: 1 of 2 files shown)


src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -62,8 +62,15 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
         parser = argparse.ArgumentParser()
         parser.add_argument('--cerncastor', action='store_true')
         parser.add_argument('--old', action='store_true')
+        parser.add_argument('--disablewriterecovery', action='store_true')
```
hufnagel: Why ? We don't have any old-style syntax in site-local-config to worry about for this parameter.

hufnagel: And no workflows (LogCollect) that set that either.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -111,6 +118,10 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
                     for initFile in initFiles:
                         copyCommand += "source %s\n" % initFile
 
+        if (args.wma_disablewriterecovery or args.disablewriterecovery) and not\
+                (args.wma_cerncastor or args.cerncastor):
```
hufnagel: Same comment as in the other PR, we can just remove that secondary check, cleaner that way.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -62,8 +62,15 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
...
         parser.add_argument('--old', action='store_true')
+        parser.add_argument('--disablewriterecovery', action='store_true')
+        parser.add_argument('--wma-cerncastor', action='store_true')
+        parser.add_argument('--wma-old', action='store_true')
+        parser.add_argument('--wma-disablewriterecovery', action='store_true')
         args, unknown = parser.parse_known_args(options.split())
 
+        # strip out WMAgent specific options
+        unknown = set(unknown) - {'--cerncastor', '--old', '--disablewriterecovery',
+                                  '--wma-cerncastor', '--wma-old', '--wma-disablewriterecovery'}
```
hufnagel: Any reason you don't use the more general removal of generic wma options here ?

hufnagel (changes requested):

amaltaro:
Thanks, Dirk. Code is updated according to your suggestions.

amaltaro:
[Fix #7888 - Add option to disable xrdcp write recovery; remove xrdcp-old](https://trello.com/c/fmWpPxyZ/36-fix-%237888---add-option-to-disable-xrdcp-write-recovery%3B-remove-xrdcp-old)

amaltaro:
My tests were successful. I'm going to push it to production agents ASAP such that @hufnagel  can update the site-local-config, before his day is over. Otherwise I leave these changes to a Friday, which is far from ideal.

amaltaro:
Dirk, all the production agents (+ relval + hepcloud) were just patched.

hufnagel:
Yeah, but the Tier0 hasn't been patched yet, right ? So as is I can update the SITECONF for T2_CH_CERN but not T0_CH_CERN.

amaltaro:
Correct, not the T0 node(s). Actually, I usually don't touch T0 agents. Can you patch it and make sure it's used for any future replays?

PR #7889, opened 2017-06-01
move Repack and Express Spec from T0

hufnagel:
Also change the whole parameter setup so it passes the WMCore validation. Add Repack and Express spec unit tests as well.

The Repack unit test is ok. The Express has problems, but I need feedback on it. In Express there is a TaskChain and AlcaSkim follows Express. AlcaSkim only uses one core, but seems to inherit the number of streams setting from Express, which it should not. I likely have to change something in the Express spec to fix this. Any suggestions ?

Another issue, if you check the last Express unit test, the cmsswVersion for step cmsRun1 is a list with a single entry while the cmsswVersion for step cmsRun2 is a string. This seems to work, but also seems wrong and should be fixed in the spec. Suggestions welcome as well.

@amaltaro @ticoann , please comment.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Express.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Express.py
@@ -1,0 +1,500 @@
+"""
+_Express_
+
+Express workflow
+
+express processing -> FEVT/RAW/RECO/whatever -> express merge
+                      (supports multiple output with different primary datasets)
+                   -> ALCARECO -> alca skimming / merging -> ALCARECO
+                                                          -> ALCAPROMPT -> end of run alca harvesting -> sqlite -> dropbox upload
+                   -> DQM -> merge -> DQM -> periodic dqm harvesting
+                                          -> end of run dqm harvesting
+"""
+
+from __future__ import division
+
+import WMCore.WMSpec.Steps.StepFactory as StepFactory
+
+from Utils.Utilities import makeList, makeNonEmptyList
+from WMCore.Lexicon import procstringT0
+
+from WMCore.ReqMgr.Tools.cms import releases, architectures
+
+from WMCore.WMSpec.StdSpecs.StdBase import StdBase
+
+
+class ExpressWorkloadFactory(StdBase):
+    """
+    _ExpressWorkloadFactory_
+
+    Stamp out Express workflows.
+    """
+    def __init__(self):
+        StdBase.__init__(self)
+
+        self.inputPrimaryDataset = None
+        self.inputProcessedDataset = None
+
+        return
+
+    def buildWorkload(self):
+        """
+        _buildWorkload_
+
+        Build the workload given all of the input parameters.
+
+        Not that there will be LogCollect tasks created for each processing
+        task and Cleanup tasks created for each merge task.
+
+        """
+        workload = self.createWorkload()
+        workload.setDashboardActivity("tier0")
+        self.reportWorkflowToDashboard(workload.getDashboardActivity())
+
+        cmsswStepType = "CMSSW"
+        taskType = "Processing"
+
+        # complete output configuration
+        for output in self.outputs:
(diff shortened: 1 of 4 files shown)


amaltaro:
> The Repack unit test is ok. The Express has problems, but I need feedback on it. In Express there is a TaskChain and AlcaSkim follows Express. AlcaSkim only uses one core, but seems to inherit the number of streams setting from Express, which it should not. I likely have to change something in the Express spec to fix this. Any suggestions ?

Are you talking about this part of the code
https://github.com/dmwm/WMCore/pull/7889/files#diff-4d394a69f5a7e33bc0b58c741df9e1f4R198
?

IF so, one way to make sure your task will be created with the parameters you want and not with the workload parameters (top level dict or so) is by passing a `taskConf` dictionary. This line:
https://github.com/dmwm/WMCore/pull/7889/files#diff-4d394a69f5a7e33bc0b58c741df9e1f4R206

Just noticed there is a small set of parameters that are handled in this method (overriden from taskConf):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L251


> Another issue, if you check the last Express unit test, the cmsswVersion for step cmsRun1 is a list with a single entry while the cmsswVersion for step cmsRun2 is a string. This seems to work, but also seems wrong and should be fixed in the spec. Suggestions welcome as well.

Dirk, you probably wanted to say that the ScramArch is a string instead of a list. The reason seems to be on the Express.getWorkloadCreateArgs arg definition.
`"RecoScramArch": {"optional": False, "null": True},`
when you should be using the same definition we have in StdBase:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L965

same comment applies to `RecoCMSSWVersion`. And you better double-check all the other arguments that you're defining to avoid data inconsistencies.

hufnagel:
Already had a taskConf for the AlcaSkim, just didn't include the EventStreams parameter. Fixed.

I also changed both RecoCMSSWVersion and RecoScramArch to be the same as CMSSWVersion and ScramArch in StdBase. This makes the unit test fail with validation errors and the Tier0 replay fails the same way. The error message is not very illuminating.

hufnagel:
Hm, maybe now the validation actually complains that it's a string and not a list. I still don't see why I can set ScramArch to a string and it comes out as a list in the spec and the same doesn't work for RecoScramArch.

amaltaro:
There is only 1 unit test failing now.
BTW, you can set scramArch to either a string or a list, though when you call the factory it's going to cast your data to `makeNonEmptyList`, where you get the list of scrams.

hufnagel:
I updated the unit tests and put in valid scram archs for the cmssw versions I am using. Lets see what the next (last as there might be a few queued for this) jenkins test says.

hufnagel:
Once the unit tests work I am looking at the other complaints/warnings.

hufnagel:
Ok, cleaned up things a bit and also added imports for the releases() and architectures() functions, maybe it works now.

hufnagel:
Ok, this looks good now. @amaltaro , can your review please ?

hufnagel:
Can this please be merged asap ? It's not testable by anything but a Tier0 replay and the unit tests work, so there shouldn't be any holdup. I want to have this in a new tag.

ticoann:
@hufnagel, Dirk, 1.1.4.pre5 is created

ericvaandering:
```
Pylint check: failed
12 warnings and errors that must be fixed
201 comments to review
Pycodestyle check: succeeded
372 comments to review
```
Please start paying some nominal attention to this stuff, guys.

ticoann:
Sorry I just checked the bold warnings. There are private values defined outside init unfortunately we are doing a lot of this in the code in WMCore. (This code is moved from Tier0 code not created newly)

hufnagel:
I did an fixed the "real" errors.

PR #7892, opened 2017-06-02
Remove default value for ProcessingVersion inside a task

### LINKED ISSUE #7891: Task ProcessingVersion default value shouldn't be 0
That change was made by me and there was a good reason to keep that set to 0 (I think to fallback to the workload level arg in case that was eval to false).

However, if one creates a TaskChain request with ProcessingVersion arg only at the top level, the task dict will override that to 0, thus the output dataset for that task will be `v0`.

We either set the inner task default to 1 or we don't set a default there, such that the top level (workload) is used instead. I think latter option is better and less error prone.

amaltaro:
Fixes #7891
Change is simple, I'm just not sure whether there will be bad side effects in some other random use case.


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Better, but I still need to test this and #7893. We can talk tomorrow and ask for a new testbed tag.

amaltaro:
Looks good in my tests and all suggestions were taken care of.

amaltaro:
Actually, suggestions were made on the other PR :-)

ticoann (approved):

PR #7893, opened 2017-06-02
Fix to jobsByChildQueue workqueue views; WQ monitoring data reformat

### LINKED ISSUE #7847: Review data structs for agent/services metrics
These are the metrics we upload to wmstats (agentInfo), here:
https://cmsweb.cern.ch/couchdb/_utils/database.html?wmstats/_design/WMStatsErl/_view/agentInfo

Benjamin discussed with Alberto (CERN IT) and we have to improve some of these metrics in order to have an easier data lookup and aggregation. 

BTW, dots '.' are not allowed in key names. @stiegerb can you please:
*) find out which regex is used to validate these key names
*) and comment here on a few cases - with examples - that should be changed from/to
?

As we discussed, I'm assigning this issue to you :) but if it becomes urgent and you don't manage to check it, we can always reassign it.

amaltaro:
and Reformat workqueue monitoring information

Fixes #7847

@stiegerb noticed we had only one priority dict per agent (the last one), this is what this PR really fixes.
Then on top of that, I've changed the data format for most of the workqueue monitoring information such that the real value we want to query for are in a value field instead of a key. E.g.:
```
"workByAgentAndPriority": {
    "http://vocms0309.cern.ch:5984": {
           "900000": {
               "count": 1,
               "min": 39374,
               "max": 39374,
               "sum": 39374,
               "sumsqr": 1550311876
           }
       },
```

should become
```
"workByAgentAndPriority": [
    {
        "agent_name": "vocms0309",
        "priority": "900000"
        "count": 1,
        "min": 39374,
        "max": 39374,
        "sum": 39374,
        "sumsqr": 1550311876
    },
```

Benjamin will still have to break that up into several smaller docs before pushing that into stompAMQ, but at least it should be a bit easier now and it will also be helpful once we start injecting data directly into stompAMQ.

@ticoann @stiegerb @vkuznet can you please review it.
Still needs to be tested.

### DIFF
--- a/src/python/WMCore/Services/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/Services/WorkQueue/WorkQueue.py
@@ -93,7 +93,8 @@ class WorkQueue(object):
         data = self.db.loadView('WorkQueue', 'jobsByChildQueueAndStatus', options)
         result = {}
         for x in data.get('rows', []):
-            result[x['key'][0]] = {x['key'][1]: x['value']}
+            result.setdefault(x['key'][0], {})
+            result[x['key'][0]][x['key'][1]] = x['value']
 
         return result
 
@@ -108,7 +109,8 @@ class WorkQueue(object):
         data = self.db.loadView('WorkQueue', 'jobsByChildQueueAndPriority', options)
         result = {}
         for x in data.get('rows', []):
-            result[x['key'][0]] = {x['key'][1]: x['value']}
+            result.setdefault(x['key'][0], {})
+            result[x['key'][0]][x['key'][1]] = x['value']
 
         return result
 
--- a/src/python/WMCore/Services/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/Services/WorkQueue/WorkQueue.py
@@ -85,16 +85,20 @@ class WorkQueue(object):
          3. minimum number of expected Jobs in an element
          4. maximum number of expected Jobs in an element
          5. sum of the squares of the expected Job in each element
+
+        Also reformat the output such that it's MONIT IT friendly and easier to aggregate.
         """
         options = {'reduce': True, 'group_level': 2}
         if stale:
             options['stale'] = 'update_after'
 
         data = self.db.loadView('WorkQueue', 'jobsByChildQueueAndStatus', options)
-        result = {}
+        result = []
         for x in data.get('rows', []):
-            result.setdefault(x['key'][0], {})
-            result[x['key'][0]][x['key'][1]] = x['value']
+            item = {'agent_name': self._getShortName(x['key'][0]),
+                    'status': x['key'][1]}
+            item.update(x['value'])
+            result.append(item)
 
         return result
 
@@ -107,12 +111,26 @@ class WorkQueue(object):
             options['stale'] = 'update_after'
 
         data = self.db.loadView('WorkQueue', 'jobsByChildQueueAndPriority', options)
-        result = {}
+        result = []
         for x in data.get('rows', []):
-            result.setdefault(x['key'][0], {})
-            result[x['key'][0]][x['key'][1]] = x['value']
+            item = {'agent_name': self._getShortName(x['key'][0]),
+                    'priority': int(x['key'][1])}
+            item.update(x['value'])
+            result.append(item)
 
         return result
+
+    def _getShortName(self, longQueueName):
+        """
+        Get a full workqueue queue name (full hostname + port) and return its short name,
+        otherwise it fails to get injected into elastic search. E.g.:
+            from "http://cmssrv217.fnal.gov:5984" to "cmssrv217"
+        """
+        if longQueueName is None:
+            return "AgentNotDefined"
+        shortName = longQueueName.split('//')[-1]
+        shortName = shortName.split('.')[0]
+        return shortName
 
     def getWMBSUrl(self):
         """Get data items we have work in the queue for"""
@@ -308,9 +326,11 @@ class WorkQueue(object):
         options = {'reduce': True, 'group': group, 'stale': 'update_after'}
 
         data = db.loadView('WorkQueue', 'jobsByStatus', options)
-        result = {}
+        result = []
         for x in data.get('rows', []):
-            result[x['key']] = x['value']
+            item = {'status': x['key']}
+            item.update(x['value'])
+            result.append(item)
 
         return result
 
@@ -330,9 +350,12 @@ class WorkQueue(object):
             options['stale'] = 'update_after'
 
         data = self.db.loadView('WorkQueue', 'jobsByStatusAndPriority', options)
-        result = {}
+        result = []
         for x in data.get('rows', []):
-            result[x['key'][0]] = {x['key'][1]: x['value']}
+            item = {'status': x['key'][0],
+                    'priority': int(x['key'][1])}
+            item.update(x['value'])
+            result.append(item)
 
         return result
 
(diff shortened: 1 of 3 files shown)


vkuznet:
Alan, it would be nice if you'll post entire doc you post. The changes you made looks much better with respect to what you had before, but I wonder about doc itself. For instance, why "priority": "900000" has value of string type? Should be int or string? Remember that when you'll do aggregation, sorting these things matter. For instance,
```
"9">"8"
True
```
which is not probably what you want.

ticoann:
It looks fine for me. The change is only affects monitoring and Benjamin will handle the format difference, correct. As far as priority type, it seems 'int' in the code I am not sure why example shows string.

amaltaro:
Valentin, as Seangchan already confirmed, priority is an integer value. Apparently, when we dump an integer key into a json object it's converted into a string. Anyways this PR should "fix" it.

You can find the current data format here:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/PR_7893/original.json
and the new format which will come out of this patch in:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/PR_7893/new.json

Disclaimer: the jobs by site was intentionally left alone because I wanted to preserve a single data struct for each workqueue status. Otherwise we'd have a list with the status scrambled in the elements/dict of this list.

vkuznet:
Alan,
how do you dump the JSON? To get proper data-types you must always use
json.dumps instead of print. If you got a string it means either your dump is
wrong or your actual value was a string data-type.

When I look at new.json from your PR:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/PR_7893/new.json

I see the following questionable values:

- agent name sometimes is "None", do you mean None python data-type or "None"
  as a string data-type, e.g. it is listed as
```
"workByAgentAndStatus": [
    {
      "agent_name": "None", 
```

- all available jobs are real numbers, instead of ints, e.g.
```
"Available": [
      {
        "Jobs": 14021.0,
        "NumElems": 15,
        "site_name": "T3_CH_Volunteer"
      }
```
while acquired jobs are ints:
```
"Acquired": [
      {
        "Jobs": 132,
        "NumElems": 5,
        "site_name": "T1_DE_KIT"
      }
```

Best,
Valentin.

amaltaro:
Thanks for the quick reply!
It's using json.dumps, see
```
>>> json.dumps({1: 123, '2': '234'})
'{"1": 123, "2": "234"}'
>>> json.loads('{"1": 123, "2": "234"}')
{u'1': 123, u'2': u'234'}
```

googling for a minute it tells me json objects cannot have integer keys, they are always strings.

About string None, I just wanted to make it consistent with all the other agent_name values, instead of having a None (or null) value over there. I know there are cases it only confuses it further, but IMO this is not one of them. If you guys think it should be None object, I can update it.

About real x integer, I cannot find that real value. Jobs is always an integer value (well, one of those metrics could be real, but I wanted to avoid juggling with decimal digits).

vkuznet:
On  0, Alan Malta Rodrigues <notifications@github.com> wrote:
> Thanks for the quick reply!
> It's using json.dumps, see
> ```
> >>> json.dumps({1: 123, '2': '234'})
> '{"1": 123, "2": "234"}'
> >>> json.loads('{"1": 123, "2": "234"}')
> {u'1': 123, u'2': u'234'}
> ```
> 
> googling for a minute it tells me json objects cannot have integer keys, they are always strings.

that's correct, and my point was that if you have a value of certain type, e.g.
int or string, the json.dumps will dump them correctly.

And, it is a good habit not to use int keys because of the issue you outlined.

> About string None, I just wanted to make it consistent with all the other agent_name values, instead of having a None (or null) value over there. I know there are cases it only confuses it further, but IMO this is not one of them. If you guys think it should be None object, I can update it.

In this case I rather prefer to see empty string, e.g. "" instead of "None".

> 
> About real x integer, I cannot find that real value. Jobs is always an integer value (well, one of those metrics could be real, but I wanted to avoid juggling with decimal digits).

well, it is your new.json file: http://amaltaro.web.cern.ch/amaltaro/forWMCore/PR_7893/new.json
see
```
{
  "possibleJobsPerSite": {
    "Acquired": [
      {
        "Jobs": 6,
        "NumElems": 2,
        "site_name": "T1_DE_KIT"
      },
      {
        "Jobs": 126,
        "NumElems": 3,
        "site_name": "T1_IT_CNAF"
      },
      {
        "Jobs": 1000,
        "NumElems": 1,
        "site_name": "T3_CH_Volunteer"
      },
      {
        "Jobs": 1000,
        "NumElems": 1,
        "site_name": "T2_US_Nebraska"
      },
      {
        "Jobs": 132,
        "NumElems": 5,
        "site_name": "T1_US_FNAL"
      }
    ],
    "Available": [
      {
        "Jobs": 14021.0,
        "NumElems": 15,
        "site_name": "T3_CH_Volunteer"
      }
    ]
  }
```
the last entry for Availabel contains a real value for Jobs.

amaltaro:
Benjamin has more experience than me with the monit it system, but I think having an empty string for the agent_name will be tricky to plot. It'll likely be an empty row under the plot as well, so not very clear to present that info. If Kibana and grafana will not have any issues ploting issue (keeping it clear), then empty string is clearly the best approach indeed.

amaltaro:
I'm now getting the ceiling of the Jobs estimative (per site). So it should be all float (until we move to python3, which then will return ints as far as I read).

ticoann:
test this please

stiegerb:
Hi Alan,
For the `"agent_name":"None"`, how about finding something more meaningful? If I understood you correctly the other day it could be something along the lines of `"not_yet_acquired"`?
Indeed empty strings would be cumbersome in Kibana/Grafana.

Cheers,
Benjamin

amaltaro:
Calling it "AgentNotDefined". The new.json example has been updated with the global workqueue output expected.

amaltaro:
@ericvaandering it seems these jenkins links are getting redirect to cms-conddb? Would you know if there is any outage or some other issue?

amaltaro:
test this please

ericvaandering:
Works for me, @amaltaro Do you still see this?

src/python/WMCore/Services/WorkQueue/WorkQueue.py
```diff
@@ -106,12 +111,27 @@ def getChildQueuesAndPriority(self, stale=True):
...
         return result
 
+    def _getShortName(self, longQueueName):
+        """
+        Get a full workqueue queue name (full hostname + port) and return its short name,
+        otherwise it fails to get injected into elastic search. E.g.:
+            from "http://cmssrv217.fnal.gov:5984" to "cmssrv217"
+        """
+        if longQueueName is None:
+            return "AgentNotDefined"
```
ticoann: This means element is not pulled to agent and still "Available" state in global lqueue, right?

amaltaro: Exactly!

amaltaro:
Eric, it still fwds my request to https://cms-conddb.cern.ch/docs/
However, it works if I open it in another browser.

ericvaandering:
Maybe try logging out. Sounds like something is busted with the round trip through CERN SSO

amaltaro:
Eric, indeed it works in another browser. Restarting firefox also sorted that issue out. Thanks

amaltaro:
Looks good in my tests and all suggestions were taken care of.

amaltaro:
Once these 2 are merged (7892 + 7893), please cut a new tag for testbed.

ticoann (approved):

PR #7897, opened 2017-06-06
[DO NOT MERGE] Full Job system monitoring for SENSE

juztas:
This is as discussed on email thread to enable full job monitoring at NERSC side.
It enables multiple ApMon destinations and publish all system and also job information to monalisa server. Default publishing to Dashboard stays as it is.
It is not suggested to include this into main stream and have all jobs to report system information.
@amaltaro @vlimant


amaltaro:
@hufnagel when I talked to Justas a month ago, I understood this was meant only for NERSC (or some other US opportunistic resource). Are you using it on your tests? Will we have to add it to the agents? If not, then I suggest we close it.

amaltaro:
@juztas @hufnagel ping.

hufnagel:
No idea honestly, we need feedback from @juztas what he wants to do with this.

PR #7898, opened 2017-06-06
disable xrootd write recovery for alcaharvest uploads

hufnagel:
Trivial change, just want to be consistent with how we treat other EOS uploads.


amaltaro:
merging it for the next tag.

PR #7873, opened 2017-05-24
Increment ProcessingVersion of clone requests

amaltaro:
Fixes #7862
and also add back OriginalRequestName that I had mistakenly removed from ACDCs.

PR #7902, opened 2017-06-07
Skip ProcVer bump for Resubmission

amaltaro:
I injected this bug in #7873 , where I should be checking the new request args as well, not only the original (clone) ones.
Already tested in my VM.


amaltaro:
@ticoann I'm just waiting for jenkins to finish and then I'd like to merge it, cut a new tag and update the cmsdist request. Would you have something else to be added?
I asked Lina and she should be able to run this testbed deployment still today.

ticoann (approved):

PR #7911, opened 2017-06-08
Properly filter out unknown keys for clone/resub

### LINKED ISSUE #7909: Buggy initialize_clone method
I know I wrote this code myself
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/Request.py#L75

but depending on the case it can raise `RuntimeError: dictionary changed size during iteration` :(

An easy fix would be replacing iteritems by items, but then it'll simply postpone this bug for when we migrate to python3.

I decided to make this issue just to draw attention that the same issue might/will happen somewhere else that we do the same dict handling.

BTW, I think it was not caught by the unit tests because I don't have invalid args in the sample dict (and that's why it passed all tests in my VM as well...).

amaltaro:
Fixes #7909 
Yet another PR on the clone/resubmit subject (there must be something wrong with me!)
Fixes a few issues, besides the one changing the dict that we're iterating over.
* Task1/Step1 are known args, so first check whether the key matches that regex and remove unsupported inner keys
* *really* remove inner keys :)
* ReReco has no inner chain definition, so it would check against None

Sorry about that. Still to be tested on my VM, but unit tests are Ok.

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/Request.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/Request.py
@@ -85,19 +85,22 @@ def initialize_clone(requestArgs, originalArgs, argsDefinition, chainDefinition=
     :return: dictionary with original args filtered out, as per the spec definition. And on
      top of that, user arguments added/replaced in the dictionary.
     """
+    chainDefinition = chainDefinition or {}
     chainPattern = r'(Task|Step)\d{1,2}'
+
     cloneArgs = {}
     for topKey, topValue in originalArgs.iteritems():
-        if topKey in argsDefinition:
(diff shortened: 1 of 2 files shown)


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
@ticoann those 2 jenkins results above ran on the same code, so I assume we are good to merge it (unless you want to wait for the next run).
I ran tests on my VM and it should be fine now, I hope :)

After you merge it, can you cut a new tag (pre8?)? I can then make a new cmsdist request tomorrow morning. Thanks!

ticoann:
test this please

ticoann (approved):

PR #7912, opened 2017-06-08
Cleanup JobCreator_t.py

ericvaandering:
including changing to set literals for @amaltaro

### DIFF
--- a/test/python/WMComponent_t/JobCreator_t/JobCreator_t.py
+++ b/test/python/WMComponent_t/JobCreator_t/JobCreator_t.py
@@ -1,50 +1,37 @@
 #!/bin/env python
-#pylint: disable=E1101, W0201, W0142, E1103
+# pylint: disable=E1101, W0201, E1103
 # E1101: reference config file variables
 # W0142: ** magic
 # W0201: Don't much around with __init__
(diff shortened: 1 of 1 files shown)


amaltaro:
wonderful! :)

amaltaro:
test this please

amaltaro:
@ericvaandering I guess there are still a few warnings/errors that you could get rid of? :)

ericvaandering:
Three of these tests are just commented out which is maybe the biggest source of warnings. But it seems they might be useful at some point?

amaltaro:
Hmm, yes, I think so. How about decorating them with `performance` then? If performance unittests are picked up by jenkins, then we could use `integration`, right?

ericvaandering:
Ok, this is as good as it will get. @ticoann please merge when Jenkins comes back.

ticoann:
test this please

amaltaro:
test this please

ticoann:
test this please

ticoann:
test this please

ericvaandering:
test this please

ericvaandering:
@ticoann please merge

PR #7913, opened 2017-06-08
New module to return status and trhead stack of the given process

vkuznet:


src/python/Utils/procstats.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+#-*- coding: utf-8 -*-
+#pylint: disable=
```
amaltaro: this line can be removed

src/python/Utils/procstats.py
```diff
@@ -0,0 +1,92 @@
...
+
+try:
+    import psutil
+except:
+    pass
+
+def timestamp(time0=None):
+    "Standard timestamp"
+    ttime = time.localtime(time0) if time0 else time.localtime()
+    tstamp = time.strftime('%d/%b/%Y:%H:%M:%S', ttime)
```
amaltaro: Not sure it makes too much difference, but one time format (suggested by the epoch converter page) that is clear, including the local time zone is:
`time.strftime("%a, %d %b %Y %H:%M:%S +0000", ttime)`
maybe we could update it?

src/python/Utils/procstats.py
```diff
@@ -0,0 +1,92 @@
...
+        tdict = {"thead":id2name.get(tid,""), "thead_id": tid}
+        stack_list = []
+        for filename, lineno, name, line in traceback.extract_stack(stack):
+            sdict = dict(filename=filename, line_number=lineno, name=name, line=line)
+            stack_list.append(sdict)
+        tdict.update({"stack":stack_list})
+        threads.append(tdict)
+    return dict(threads=threads)
+
+class OptionParser():
```
amaltaro: Since we're going to use it as a module, I believe this class and the `main` function are not really needed here, right?

amaltaro (commented):
In addition to comments made along the code, can you please update this module name to initial capital letters (like the others in the same package)?

I also suggest running a tool to reformat this code (lack of spaces here, extra spaces there, etc).

Since it's a new module, jenkins will also ask you to import a couple of functions from __future__ (division and print_function I think), for python3 migration purposes.

amaltaro:
BTW, it would be great to have some basic unit tests for these functions. An example is:
https://github.com/dmwm/WMCore/blob/master/test/python/Utils_t/IteratorTools_t.py#L19

src/python/Utils/procstats.py
vkuznet: Alan, I rather like to keep it since it allows to run this as standalone script and does not affect anything else.

src/python/Utils/procstats.py
vkuznet: I usually dislike having %a in timestamp since printing a day has little usage but in fact eat the space.

src/python/Utils/procstats.py
vkuznet: Well, as I replied already to your comment about #!/usr/bin/env, the script can be used stand-alone. I don't mind to remove it although.

vkuznet:
Alan, I doubt I can add meaningful unit test for this module since it capture details of running process in real-time. Even if add sys call the results (CPU or RAM) may differ since those should be captured at the same time. If you have specific idea about unit test please let me know, otherwise I don't see a point to writing unit test for it.

vkuznet:
@sry, @amaltaro, @ericvaandering how to tweak jenkins to skip testing non-existing file?
I renamed procstats.py to ProcessStats.py but jenkins keeps triggering error on procstats.py

ericvaandering:
Whew. :-)

vkuznet:
How do you want me to deal with this pylint error:
```
W0212, line 71 in threadStack: Access to a protected member _current_frames of a client class
```

in order to get thread info it is the only way to do it (via access of protected member) and there is no public API for that in sys module.

My understanding that this should be added as exception to CMS/DMWM pylint rules.

ticoann (commented):
> My understanding that this should be added as exception to CMS/DMWM pylint rules.

I think we can still keep this rule (but obviously we should ignore for this case since we are accessing sys module. Still we should avoid making "_variable" to public variable in our code. Although there are many already)

vkuznet:
@ticoann please review, your original comments refer to removed code.

vkuznet:
@ticoann please review this asap, I want to include this module for new WMArchive release and deploy status endpoint for it.

ticoann:
@vkuznet, Valentin, patch looks good to me. Could you squash the commit? I will merge that. Do you need the tag today?

vkuznet:
@ticoann, Seangchan, I squashed the changes, please merge.
I'll appreciate a tag today/tomorrow. This tag I can use to build WMArchive only.

ticoann:
Thanks Valentin, I will create the new tag tomorrow then.

ticoann:
test this please

Issue #7256, opened 2016-10-04
Set T0 workflows activity to "t0"


amaltaro:
As mentioned here #7253
Several dashboard activities are mapped to `t0`. So we should update it from `tier0` to `t0` in the T0 specs here and in the T0 repository as well. No hurry though

hufnagel:
@amaltaro , can you suggest where exactly in the Tier0 specs this should be changed ?

amaltaro:
First change (replace tier0 by t0 activity) should be in the setDashboardActivity call, thus:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py#L34

same for Express and Repack.

Second change is, if Tier0Feeder is now calling this workload method (as we discussed a week or two ago, which basically updates the workload object and calls a bunch of setter methods - used for wf assignment):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1633

then you need to pass the t0 value for `Dashboard` key.

PR #7920, opened 2017-06-09
fix dashboard activity for tier0

ticoann:
@hufnagel, Dirk please review. 

Jen found out that Tier0 workflows are reported as “production” activity in dashboard.

http://dashb-cms-job.cern.ch/dashboard/templates/web-job2/#user=&refresh=60&table=Jobs&p=1&records=25&activemenu=1&usr=&site=&submissiontool=&application=&activity=production&status=&check=submitted&tier=&date1=2017-06-08+19%3A53&date2=2017-06-09+19%3A53&sortby=task&scale=linear&bars=20&ce=&rb=&grid=&jobtype=&submissionui=&dataset=&submissiontype=&task=&subtoolver=&genactivity=&outputse=&appexitcode=&accesstype=&inputse=&cores=

It seems it is hard coded for express, repack and promptreco.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/Express.py#L51
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/Repack.py#L46
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py#L34

But it gets overwritten in assignment stage.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Express.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Express.py
@@ -16,7 +16,7 @@ from __future__ import division
 import WMCore.WMSpec.Steps.StepFactory as StepFactory
 
 from Utils.Utilities import makeList, makeNonEmptyList
-from WMCore.Lexicon import procstringT0
+from WMCore.Lexicon import procstringT0, activity
 
 from WMCore.ReqMgr.Tools.cms import releases, architectures
 
@@ -493,7 +493,7 @@ class ExpressWorkloadFactory(StdBase):
                     "MaxInputSize": {"type": int, "optional": False},
                     "MaxInputFiles": {"type": int, "optional": False},
                     "MaxLatency": {"type": int, "optional": False},
-
+                    "Dashboard": {"default": "tier0", "type": str, "validate": activity}
                     }
         baseArgs.update(specArgs)
         StdBase.setDefaultArgumentsProperty(baseArgs)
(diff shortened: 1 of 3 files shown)


amaltaro:
If Dirk has put a code in place to call the assignment arguments, then he has to provide the correct Dashboard activity, otherwise it will get the default "production" value:
https://github.com/ticoann/WMCore/blob/efbd47fe5105548f8e9807593a72114ada941724/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1114

and won't work even with this patch. 

I think this patch is not needed, T0 just needs to pass that value when tier0 workflows are getting updated/assigned. @hufnagel btw, make sure to set it to `t0`, not "tier0" nor something else, see this issue:
https://github.com/dmwm/WMCore/issues/7256

hufnagel:
@amaltaro I can fix this in the Tier0 specs, no problem.

ticoann:
@amaltaro, @hufnagel, I think it is better to set the Dashboard default value to "tier0" or "t0" which ever is the correct form, instead of hard coding in the code. (In case testingT0 workflow should have different activities, it can be controlled by spec parameter)

amaltaro:
Yes, it could be done either during creation (we'd then need to override the createArgs on every T0 spec) or during assignmnent (when T0 updates the workload). 

If there isn't any difference, then I suggest we leave Dashboard settings to the assignment stage, leaving it hard-coded for creation, as it is for all the other specs.

hufnagel:
@amaltaro suggested to change the creation and the assignment step, which I did. If there is a way to just change one default, that would be easier. These specs are Tier0 specific.

ticoann:
closing it. - replaced by #7924

PR #7923, opened 2017-06-12
Auto-detect drain mode at WorkQueueManager level

amaltaro:
Complement of #7731 (WorkQueueManager part)

I still have to figure how to handle the `getWMBSInjectionStatus` method, which is called from WorkQueueManager and from JobArchiverPoller. They create a queue instance and that no longer changes (and this method checks the DrainMode attr). We could pass the DrainMode value to this method, or we could override that attr in case it changed. I don't know what's the cleanest way of doing that.


ticoann (commented):
created a new patch for this as we discussed

ticoann:
replaced by #8011

PR #7924, opened 2017-06-12
report the correct dashboard activity from Tier0, fixes #7256

### LINKED ISSUE #7256: Set T0 workflows activity to "t0"

hufnagel:
Fixes #7256
There is also an updateArguments call from Tier0 code when the workflow is created.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Express.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Express.py
@@ -48,7 +48,7 @@ class ExpressWorkloadFactory(StdBase):
 
         """
         workload = self.createWorkload()
-        workload.setDashboardActivity("tier0")
+        workload.setDashboardActivity("t0")
         self.reportWorkflowToDashboard(workload.getDashboardActivity())
 
         cmsswStepType = "CMSSW"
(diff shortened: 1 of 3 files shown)


amaltaro:
test this please

ticoann:
Is "tier0" is not the correct dashboard activity? I thought that was, not "t0".

ticoann:
Due to following code. 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1633

Activity will be overwritten with default value "production". I think it might be better to set the default value for Repack, Express, PromptReco to "t0".

hufnagel:
No, it'll be overwritten with t0.

https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L676

hufnagel:
That's what Alan suggested. If there is a way to change the spec so it's default is t0 and I don't need to overwrite I'll take that though, makes the code a bit easier to manage. These specs are Tier0 specific.

PR #7926, opened 2017-06-13
Collect drain statistics and report to wmstats

goughes:
This will incorporate some statistics on the draining process and report them to wmstats with the other agent info.

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -22,7 +22,7 @@ from WMComponent.AnalyticsDataCollector.DataCollectAPI import WMAgentDBData, \
     diskUse, numberCouchProcess
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue as WorkQueueDS
 from WMCore.WorkQueue.DataStructs.WorkQueueElementsSummary import getGlobalSiteStatusSummary
-
+from WMComponent.AgentStatusWatcher.DrainStatusPoller import DrainStatusPoller
 
 class AgentStatusPoller(BaseWorkerThread):
     """
@@ -183,6 +183,9 @@ class AgentStatusPoller(BaseWorkerThread):
         if isDrainMode(self.config):
             logging.info("Agent is in DrainMode")
             agentInfo['drain_mode'] = True
+            agentInfo['drain_stats'] = DrainStatusPoller.getDrainInfo()
+            logging.info(agentInfo)
+
         else:
             agentInfo['drain_mode'] = False
 
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
@@ -1,0 +1,78 @@
+"""
+Monitoring the agent drain status using CheckDrainStatus
+"""
+__all__ = []
+
+import threading
+import logging
+import time
+import traceback
+import json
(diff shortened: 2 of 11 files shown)


src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -183,6 +183,9 @@ def collectAgentInfo(self):
         if isDrainMode(self.config):
             logging.info("Agent is in DrainMode")
             agentInfo['drain_mode'] = True
+            agentInfo['drain_stats'] = DrainStatusPoller.getDrainInfo()
+            logging.info(agentInfo)
```
vkuznet: this can be combined with logging.info from line 184.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -0,0 +1,69 @@
...
+from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
+from WMComponent.AnalyticsDataCollector.DataCollectAPI import isDrainMode
+from WMComponent.DBS3Buffer.DBSBufferUtil import DBSBufferUtil
+
+class DrainStatusPoller(BaseWorkerThread):
+    """
+    Collects information related to the agent drain status
+    """
+    # class variable that contains drain statistics
+    drainStats = {}
```
vkuznet: what's the purpose of having class variable instead of instance one?

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -0,0 +1,69 @@
...
+    def collectDrainInfo(self):
+        """
+        Call methods to check the drain status
+        """
+        results = {}
+        results['workflows_completed'] = self.dbsUtil.isAllWorkflowCompleted()
+        logging.info("Drain stats - workflows completed: " + str(results['workflows_completed']))
+        return results
+
+    @classmethod
```
vkuznet: why this method needs to be a classmethod since it only returns class variable without doing anything. Can you put drainStats into __int__ method and have it as a member of the class?

vkuznet (changes requested):
I also want to add to my comments that code use mixed syntax, in many places it collapse spaces between key and a values, e.g. replace key = value with key=value, but in others it does the opposite. I would recommend to run this PR through pylint and fix code syntax.

ericvaandering:
Has this code changed since Valentin's comments as I didn't notice the `key=value` issue he mentioned. In any case, `key=value` is correct in function calls, `variable = value` is correct in executable code. The pylint and pep8 reports will tell you that.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
goughes: AgentStatusWatcher spawns three threads based of the AgentStatusPoller, DrainStatusPoller and ResourceControlUpdater classes.  Each instance of those classes doesn't know about the other.  AgentStatusPoller needs to get variable contents from DrainStatusPoller, hence the class variable.  If there is a better way, I'm open to suggestions.

goughes:
I haven't changed anything.  I will do another review based on your comments.

goughes:
This references #7615 which has problems with the unittests.  Will open new PR based on #7955 and #7775.

ticoann:
are you replacing #7955 and #7775? If you do that could you close the replaced pull requests?

goughes:
I'll wait until #7955 is merged before making the new PR.  I will close #7775 after.

Issue #7901, opened 2017-06-07
No retries on error code 87


vlimant:
this kind of error is clearly not recoverable by a retry, so we should not even try.

amaltaro:
For bookkeeping, can you please post here the error message for this exit code?

vlimant:
Fatal Exception (Exit code: 8026)
An exception of category 'NotFound' occurred while

vlimant:
Fatal Exception (Exit code: 8023)
An exception of category 'MismatchedInputFiles' occurred while
[0] Calling InputSource::readFile_
Exception Message:
RootPrimaryFileSequence::nextFile() Branch 'GlobalObjectMapRecord_hltGtStage2ObjectMap__HLT.' is in file 'root://ccmsdlf.ads.rl.ac.uk//castor/ads.rl.ac.uk/prod/cms/store/unmerged/Run2016B/NoBPTX/RECO/18Apr2017_ver1-v1/10004/1086F426-9B47-E711-8D27-0242AC110004.root'
but not in previous files.

CMSSWStepFailure (Exit code: 87)

amaltaro:
Instead of failing jobs on exit code 87 (which I haven't found anywhere a description for it), shouldn't we fail jobs with 'MismatchedInputFiles' (8023)?

vlimant:
here goes the confusion of the many exit codes exposed in wmstats.
whichever is the error code that is being seen by the agent, this should not be retried. I'd say add 87 and 8023

amaltaro:
Calling it 8023 then.

PR #7932, opened 2017-06-14
Fail jobs with exit code 8023 or 8026

### LINKED ISSUE #7907: No retries for error code 8026
Fatal Exception (Exit code: 8026)
An exception of category 'NotFound' occurred while

is not likely to succeed on retry

amaltaro:
Fixes #7907 and #7901 (for the new agent only)

@vlimant I also noticed we fail jobs - in the RelVal agent only - that are Running for too long (71304), following this timeout:
https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py#L228

Should we a) keep that, b) add to all the other agents or c) remove that tweak for the relval agent?

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -242,7 +242,7 @@ config.ErrorHandler.logLevel = globalLogLevel
 config.ErrorHandler.maxRetries = maxJobRetries
 config.ErrorHandler.pollInterval = 240
 config.ErrorHandler.readFWJR = True
-config.ErrorHandler.failureExitCodes = [50660, 50661, 50664, 71102]
+config.ErrorHandler.failureExitCodes = [8023, 8026, 50660, 50661, 50664, 71102]
 config.ErrorHandler.maxFailTime = 120000
 config.ErrorHandler.maxProcessSize = 30
 


ticoann (approved):

etc/WMAgentConfig.py
```diff
@@ -242,7 +242,7 @@
 config.ErrorHandler.maxRetries = maxJobRetries
 config.ErrorHandler.pollInterval = 240
 config.ErrorHandler.readFWJR = True
-config.ErrorHandler.failureExitCodes = [50660, 50661, 50664, 71102]
+config.ErrorHandler.failureExitCodes = [87, 8026, 50660, 50661, 50664, 71102]
```
ticoann: I couldn't find 87 below.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMExceptions.py#L82
Should we need to add that?

amaltaro: Maybe we should fail on 8023 instead of 87? Getting back to the issues

amaltaro:
@ticoann PR was updated as per our discussion. Feel free to merge it when jenkins are happy.

amaltaro:
@hufnagel @johnhcasallasl these 2 additional exit codes will be present in the T0 configuration as well. In case you don't want them, can you please update the T0 deployment script? Thanks

PR #7934, opened 2017-06-14
more fine-grained options to force unmerged and use them in PromptReco

hufnagel:
Allowing to override all outputs to either merged or unmerged is not enough. Add a way to go with the default (straight to merge enabled for all outputs) but force specific output modules to always write unmerged. Use this to force PromptReco to write ALCARECO output always unmerged.

@ticoann @amaltaro please review. This super-seeds #6994, which was a much more invasive way to accomplish the same thing.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
@@ -61,7 +61,7 @@ class PromptRecoWorkloadFactory(DataProcessing):
                                                splitAlgo = self.procJobSplitAlgo,
                                                splitArgs = self.procJobSplitArgs,
                                                stepType = cmsswStepType,
-                                               forceUnmerged = True)
+                                               forceUnmerged = ["write_ALCARECO"] if 'ALCARECO' in self.writeTiers else False)
         if self.doLogCollect:
             self.addLogCollectTask(recoTask)
 
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -393,10 +393,12 @@ class StdBase(object):
 
         if forceMerged:
             procTaskStageHelper.setMinMergeSize(0, 0)
-        elif forceUnmerged:
+        elif forceUnmerged and not isinstance(forceUnmerged, list):
             procTaskStageHelper.disableStraightToMerge()
         else:
             procTaskStageHelper.setMinMergeSize(self.minMergeSize, self.maxMergeEvents)
+            if forceUnmerged and isinstance(forceUnmerged, list):
+                procTaskStageHelper.disableStraightToMergeForOutputModules(forceUnmerged)
 
         procTaskCmsswHelper.cmsswSetup(cmsswVersion,
                                        softwareEnvironment="",
--- a/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
@@ -7,6 +7,7 @@ Implementation of an Executor for a StageOut step
 """
 from __future__ import print_function
 
+import sys
 import os
 import os.path
 import logging
@@ -68,7 +69,7 @@ class StageOut(Executor):
         # this alarm leaves a subprocess behing that may cause trouble, see #6273
         waitTime = overrides.get('waitTime', 7200 * self.step.retryCount)
 
-        logging.info("StageOut override is: %s " % self.step)
+        logging.info("StageOut override is: %s ", self.step)
 
         # Pull out StageOutMgr Overrides
 
@@ -115,11 +116,10 @@ class StageOut(Executor):
                 #Don't try to parse your own report; it's not there yet
                 continue
             stepLocation = os.path.join(self.stepSpace.taskSpace.location, step)
-            logging.info("Beginning report processing for step %s" % (step))
+            logging.info("Beginning report processing for step %s", step)
             reportLocation = os.path.join(stepLocation, 'Report.pkl')
             if not os.path.isfile(reportLocation):
-                logging.error("Cannot find report for step %s in space %s" \
-                              % (step, stepLocation))
+                logging.error("Cannot find report for step %s in space %s", step, stepLocation)
                 continue
             # First, get everything from a file and 'unpersist' it
             stepReport = Report()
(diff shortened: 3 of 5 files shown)


ticoann:
test this please

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
```diff
@@ -61,7 +61,7 @@ def buildWorkload(self):
                                                splitAlgo = self.procJobSplitAlgo,
                                                splitArgs = self.procJobSplitArgs,
                                                stepType = cmsswStepType,
-                                               forceUnmerged = True)
+                                               forceUnmerged = ["write_ALCARECO"] if 'ALCARECO' in self.writeTiers else False)
```
ticoann: where self.writeTiers get assigned?

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -391,12 +391,14 @@ def setupProcessingTask(self, procTask, taskType, inputDataset=None, inputStep=N
         procTaskCmsswHelper.setOverrideCatalog(self.overrideCatalog)
         procTaskCmsswHelper.setErrorDestinationStep(stepName=procTaskLogArch.name())
 
-        if forceMerged:
+        if forceMerged and isinstance(forceMerged, bool):
```
ticoann: does it have to check whether it is bool or not?

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
hufnagel: Look at the spec parameters .

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
hufnagel: That check is actually not doing anything for now. I just put it in since the corresponding check for the forceUnmerged does something and distinguishes between bool and list. Figured I might as well check if forceMerged is a bool and not some random type.

Effectively this specifically checks for and ignores any type but bool and list for forceUnmerged. Figured I might as well check for and ignore any type but bool for forceMerged.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -391,12 +391,14 @@ def setupProcessingTask(self, procTask, taskType, inputDataset=None, inputStep=N
...
             procTaskStageHelper.setMinMergeSize(0, 0)
-        elif forceUnmerged:
+        elif forceUnmerged and isinstance(forceUnmerged, bool):
             procTaskStageHelper.disableStraightToMerge()
         else:
             procTaskStageHelper.setMinMergeSize(self.minMergeSize, self.maxMergeEvents)
+            if forceUnmerged and isinstance(forceUnmerged, list):
```
ticoann: In stead of passing list for forceUnmerged. 
I was wondering whether we can check, 
```
if hasattr(self, "writeTiers") and "ALCARECO" in self.writeTiers 
```
However, I am not sure which one is more evil. 😄

hufnagel: This doesn't work. The override is only active for that specific output module. There are other output modules in the same step that will use straight to merge normally. Which means the code that checks whether to do straight to merge needs to know the name of the output module, not just a simple boolean yes/no for the whole thing.

hufnagel: The reason for this is that ALCARECO output isn't valid on it's own, it's just some intermediate data that is the input for another processing step. Hence under no circumstance can we allow it to be stage out straight to merge. But we also don't want to force all the PromptReco output to be written unmerged, since that causes huge inefficiencies for RECO output.

hufnagel: You have one processing step here with multiple ouputs (RECO, AOD, MINIAOD, DQMIO, ALCARECO). All output modules should use normal straight to merge limits, *except* the ALCARECO output which should always be written unmerged.

amaltaro:
> Allowing to override all outputs to either merged or unmerged is not enough. Add a way to go with the default (straight to merge enabled for all outputs) but force specific output modules to always write unmerged. Use this to force PromptReco to write ALCARECO output always unmerged.

Dirk, are you saying you want ALCARECO tier always going to the unmerged area regardless of any of those merge thresholds? Meaning, there will jobs merging these ALCARECO files later on (?)

Same comment for the first sentence when you say "straight to merge". Are you saying you want to get outputs always in the merged area? Or that they should follow the merge thresholds and IF needed, be staged out to the merged area? Meaning, there can be or cannot be merge jobs for those tiers.

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
amaltaro: and maybe rename the argument from `forceUnmerged` to `forceUnmergedTiers` to make it more clear that it's not a simple bool.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
amaltaro: I follow Seangchan on that and I think it should be removed.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
amaltaro: Two comments here: 
1) as my previous comments, I suggest keeping only `if forceUnmerged:`, which will check for empty list or list with elements.
2) this code won't work! IF it goes into the `elif` in L396, it will never go into the `else` below. IF it goes into the `else` in L398, it will never have a true forceUnmerged list.

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
```diff
@@ -136,48 +135,40 @@ def execute(self, emulator = None):
             # for that step; or so I hope
             files = stepReport.getAllFileRefsFromStep(step = step)
             for fileName in files:
-                if not hasattr(fileName, 'lfn') and hasattr(fileName, 'pfn'):
-                    # Then we're truly hosed on this file; ignore it
-                    msg = "Not a file: %s" % fileName
+
+                # make sure the file information is consistent
+                if hasattr(fileName, 'pfn') and ( not hasattr(fileName, 'lfn') or not hasattr(fileName, 'module_label') ):
```
amaltaro: I'm not sure I follow this code (and the previous one). Don't we care about the 'lfn' attr only? Well, and maybe the 'module_label' for future jobs (or dbs info?)? Why do we need this check for 'pfn'?

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
```diff
@@ -136,48 +135,40 @@ def execute(self, emulator = None):
...
+                # Figuring out if we should do straight to merge
+                #  - should we do straight to merge at all ?
+                #  - is straight to merge disabled for this output ?
+                #  - are we over the size threshold
+                #  - are we over the event threshold ?
+                straightToMerge = False
+                if not getattr(fileName, 'merged', False) and hasattr(self.step.output, 'minMergeSize'):
+                    if fileName.module_label not in getattr(self.step.output, 'forceUnmergedOutputs', []):
+                        if hasattr(fileName, 'size'):
+                            if fileName.size >= self.step.output.minMergeSize:
```
amaltaro: How about we merge these 2 ifs into `if getattr(fileName, 'size', 0) >= self.step.output.minMergeSize:`?

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
amaltaro: Hmm, maybe the pfn has already applied the storage.xml rules and that's why we need it?
BTW, there are two unneeded spaces on that line.

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
```diff
@@ -136,48 +135,40 @@ def execute(self, emulator = None):
...
+                                straightToMerge = True
+
+                if straightToMerge:
+
+                    try:
+                        fileName = self.handleLFNForMerge(mergefile = fileName,
+                                                          step = step)
+                    except Exception as ex:
+                        logging.info("minMergeSize: %s", getattr(self.step.output, 'minMergeSize', None))
+                        logging.info("maxMergeEvents: %s", getattr(self.step.output, 'maxMergeEvents', None))
```
amaltaro: How about we move these 2 info lines to just above the `handleLFNForMerge` call?
2nd comment is on the addError, I noticed you updated the exit code. Which was and will not be precise, since we don't know exactly what triggered the direct-to-merge. How about we merge those exceptions:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMExceptions.py#L168
by removing the 'by size' or 'by event'? If you prefer to keep both, then we should update this code to provide the correct one.

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
```diff
@@ -264,16 +255,12 @@ def post(self, emulator = None):
...
-                if not hasattr(file, 'lfn') or not hasattr(file, 'location') or \
-                       not hasattr(file, 'guid'):
-                    continue
-
-                file.user_dn = getattr(self.step, "userDN", None)
-                file.async_dest = getattr(self.step, "asyncDest", None)
-                file.user_vogroup = getattr(self.step, "owner_vogroup", '')
-                file.user_vorole = getattr(self.step, "owner_vorole", '')
+            for fileInfo in files:
+                if hasattr(fileInfo, 'lfn') and hasattr(fileInfo, 'location') and hasattr(fileInfo, 'guid'):
```
amaltaro: +1

amaltaro (changes requested):
Last but not least :), next time please separate pylint/code reformat changes from real changes in a different commit.

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
hufnagel: I don't want to do that since the parameter is used elsewhere as just a boolean, these other places don't care about the tier functionality, they just want to turn this off period. This is a general parameter to force all output to unmerged (same as forceMerged is a general parameter to force all output to merged). All I want is to support that forcing to unmerged to be partial in this one case, not general. I do not want every other use of this parameter to be exposed to this list option, they should be able to keep using the boolean syntax.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
hufnagel: Ok, can get rid of it. This is only called from within specs, so we control the calling code. No real need for the type check.

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
amaltaro: hmm, boolean forceUnmerged is used even in TaskChain.
With this fix you're basically redefining this argument, which will be source of confusion. On the other hand, having yet another argument for that list is also not that great...

hufnagel: Yeah, exactly. Overloading the parameter so that the clients that use the boolean functionality don't need to care that you can also pass a list seemed the least invasive option to me. Makes the low level code a bit more complex, but only in one limited location.

hufnagel: Also, two parameters makes for ugly code as well when you have to make a decision based on them (do this based on parameter A but only if parameter B isn't set and has this value).

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
hufnagel: 1 is discussed above. 2 is the correct behavior. If we go into the elif at 396, it's because forceUnmerged is a boolean and is True. There is no need to go into the else, we just want to disable all straight to merge during stageout.

The else is only triggered if neither forceMerged and forceUnmerged is True, which triggers the normal case of straight to merge enabled with the usual thresholds. Only in that case do I want to allow an override for specific output modules that should always write unmerged.

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
hufnagel: I don't understand this check either. I just kept the original check and added one more for module_label since I am actually using it, so it should be there. Based on the original check it almost looks like we don't want files for stageout that have a pfn and no lfn. But files with no pfn and no lfn are ok ???

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
hufnagel: Can do. To have consistent code, how about I do the same for the event comparison ? Defaults 0 for events and 2B for maxMergeEvents  ?

src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
hufnagel: We basically never trigger direct to merge on events since we don't really have event limits. Which was the motivation for merging this code. Getting rid of one of the exceptions and adjusting the message for the other is a good idea though.

hufnagel: I don't want to move these two lines up, since then I get that printout for every file that goes straight to merge. These two lines are were they are to help debugging of problems, ie. to give additional information when something fails.

hufnagel:
I think I addressed everything brought up in the review (either made a chance or replied why I didn't). Anything else or can this be merged now ?

ticoann:
@amaltaro, @hufnagel, Alan, Dirk, is all the the issues  commented settled? Should I merge this now?

amaltaro:
Looks Ok. I assume Dirk is going to be the first one to spot any possible problems with the stage out code :)

PR #7945, opened 2017-06-20
seems like python3 doesn't know maxint

hufnagel:
Old version works just fine (and is functionally equivalent to the new version) on python2. So this should get merged at some point, but there is no practical impact at the moment. Well, jenkins is complaining, that is the impact :-).


ericvaandering:
@ticoann please merge

PR #7949, opened 2017-06-20
Add MicroService classes

vkuznet:


vkuznet:
Example of config file for micro-service:
https://gist.github.com/vkuznet/49d07398357d88e9b967b3c9722083de

vkuznet:
Script to run Micro-Service:
https://gist.github.com/vkuznet/6a5b77ec59a59d68577c637328dea164

ticoann:
@vkuznet, Valentin, could you fix pylint warnings (bold ones)

ericvaandering:
Not just bold warnings. Everything that can be fixed. No reason to have problems in completely fresh code. Also, why is UnifiedStageout in the MicroService category? I think what's micro and what is not going to be well defined and "Services" is not providing services, but APIs to the existing services. Should give a thought to organization here. Maybe chat tomorrow.

vkuznet:
Guys, I sketched the code to show you the idea. The UnifiedStageout may be
elsewhere outside of MicroService, that what we need to discuss.
The warnings will be fixed once we agree on approach.

ticoann:
@vkuznet, So Valentin, you don't need to include this one in today's tag?

vkuznet:
Yes, I don't to include this one in today's tag.

PerilousApricot:
I was reading through the PR for curiosity's sake, and tried to follow the following link https://github.com/dmwm/WMCore/pull/7949/files#diff-aac8b14095c9afe5b6234329c0b10f2eR65 but it's broken. Is that mirrored somewhere else?

vkuznet:
Andrew,
it works for me now, may be transient error.
Valentin.

vkuznet:
@ticoann could you please review and merge.

ticoann:
Sorry Valentin for the late review. Patch looks good. Just a couple of small things.

1. Could you send cherrypy.HTTError with specific status and messages, so if client gets knows wether something wrong with request or serverside error.
2. I think you can remove import statement from __init__ function although pylint complains, unless you are planning to wright code there.

If you can squash the commit that will be great too. Thanks,

vkuznet:
@ticoann, Seangchan, I squashed commits as well as provided proper HTTPError and removed import from init. Please review/test/merge.

ticoann:
Thanks Valentin,

PerilousApricot:
Sorry, @vkuznet -- I just now saw this notification. The link I was referring to is: https://github.com/knly/MicroServiceAggregation -- which currently returns a "not found"

PR #7950, opened 2017-06-21
Properly override tasks in a Resubmission request

### LINKED ISSUE #7917: Resubmission for taskchain using unique memory value
It should have been fixed with the latest changes to wmstats resubmission, but I probably overlooked one place that still sends a Memory value to reqmgr2. IT could be that the task arguments were not taken into account for the ACDC, but then I'd expect the ACDC to run with 4 cores as well, but instead it has the proper cores settings

E.g. with the following workflow (4cores/4gb, 8cores/8gb and 1core/2gb)
https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=amaltaro_TaskChain_RelVal_Multicore_HG1706c_Validation_170530_104756_668 

and acdc (correct cores but always requesting 4gb)
https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=amaltaro_ACDC2_HLTD_TaskChain_RelVal_Multicore_HG1706c_Validation_170608_101312_3924

amaltaro:
Fixes #7917
Still need to test and create unittests for those 2 workload methods.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
@@ -39,15 +39,12 @@ class ResubmissionWorkloadFactory(StdBase):
         helper.ignoreOutputModules(self.ignoredOutputModules)
 
         # override a couple of parameters, if provided by user
-        if 'Memory' in arguments:
-            helper.setMemory(arguments['Memory'])
-        if 'Campaign' in arguments and not helper.getCampaign():
-            helper.setCampaign(arguments["Campaign"])
         if 'RequestPriority' in arguments:
             helper.setPriority(arguments["RequestPriority"])
-        if 'TimePerEvent' in arguments:
-            for task in helper.taskIterator():
-                task.setJobResourceInformation(timePerEvent=arguments["TimePerEvent"])
+        if arguments['OriginalRequestType'] != 'TaskChain' or isinstance(arguments['Memory'], dict):
+            helper.setMemory(arguments['Memory'])
+        if arguments['OriginalRequestType'] != 'TaskChain' or isinstance(arguments.get('TimePerEvent'), dict):
+            helper.setTimePerEvent(arguments.get("TimePerEvent"))
 
         return helper
 
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -532,7 +532,7 @@ class WMTaskHelper(TreeHelper):
         the three key values are main memory usage, time per processing unit (e.g. time per event) and
         disk usage per processing unit (e.g. size per event).
         """
-        if self.taskType() in ["Merge", "Harvesting", "Cleanup", "LogCollect"]:
+        if self.taskType() in ["Merge", "Cleanup", "LogCollect"]:
             # don't touch job requirements for these task types
             return
 
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -532,26 +532,22 @@ class WMTaskHelper(TreeHelper):
         the three key values are main memory usage, time per processing unit (e.g. time per event) and
         disk usage per processing unit (e.g. size per event).
         """
+        if self.taskType() in ["Merge", "Harvesting", "Cleanup", "LogCollect"]:
+            # don't touch job requirements for these task types
+            return
+
         performanceParams = getattr(self.data.input.splitting, "performance")
+
+        timePerEvent = timePerEvent.get(self.name()) if isinstance(timePerEvent, dict) else timePerEvent
+        sizePerEvent = sizePerEvent.get(self.name()) if isinstance(sizePerEvent, dict) else sizePerEvent
+        memoryReq = memoryReq.get(self.name()) if isinstance(memoryReq, dict) else memoryReq
+
         if timePerEvent or getattr(performanceParams, "timePerEvent", None):
             performanceParams.timePerEvent = timePerEvent or getattr(performanceParams, "timePerEvent")
         if sizePerEvent or getattr(performanceParams, "sizePerEvent", None):
             performanceParams.sizePerEvent = sizePerEvent or getattr(performanceParams, "sizePerEvent")
-
-        # special handling for memory overwrite during assignment
-        if isinstance(memoryReq, dict):
-            taskMemory = memoryReq.get(self.name())
-        else:
-            taskMemory = memoryReq
-
-        if self.taskType() in ["Merge", "Harvesting", "Cleanup", "LogCollect"]:
-            # we don't want to touch memory for these task types
-            pass
-        elif taskMemory or getattr(performanceParams, "memoryRequirement", None):
-            performanceParams.memoryRequirement = taskMemory or getattr(performanceParams, "memoryRequirement")
-
-        for task in self.childTaskIterator():
-            task.setJobResourceInformation(memoryReq=memoryReq)
+        if memoryReq or getattr(performanceParams, "memoryRequirement", None):
+            performanceParams.memoryRequirement = memoryReq or getattr(performanceParams, "memoryRequirement")
 
         return
 
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -641,17 +641,55 @@ class WMWorkloadHelper(PersistencyHelper):
             task.setNumberOfCores(cores, nStreams)
         return
 
-    def setMemory(self, memory):
+    def setMemory(self, memory, initialTask=None):
         """
         _setMemory_
 
-        Update memory requirements for each task in the spec
+        Update memory requirements for each task in the spec, thus it
+        can be either an integer or a dictionary key'ed by the task name.
         """
         if not memory:
             return
 
-        for task in self.taskIterator():
-            task.setJobResourceInformation(memoryReq=memory)
+        if initialTask:
+            taskIterator = initialTask.childTaskIterator()
+        else:
+            taskIterator = self.taskIterator()
+
+        for task in taskIterator:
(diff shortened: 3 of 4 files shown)


src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
```diff
@@ -39,15 +39,18 @@ def buildWorkload(self, arguments):
...
         if 'RequestPriority' in arguments:
             helper.setPriority(arguments["RequestPriority"])
-        if 'TimePerEvent' in arguments:
-            for task in helper.taskIterator():
-                task.setJobResourceInformation(timePerEvent=arguments["TimePerEvent"])
+        if arguments['OriginalRequestName'] == 'TaskChain' and not isinstance(arguments['Memory'], dict):
+            # Memory must be a dict object if user wants to override TaskChain tasks
+            pass
+        else:
+            helper.setMemory(arguments['Memory'])
```
ticoann: How about changing to following? It is more compact but if you think it might be less clear, don't change.
```
if arguments['OriginalRequestName'] != 'TaskChain' or isinstance(arguments['Memory'], dict):
    helper.setMemory(arguments['Memory'])
```

src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
```diff
@@ -39,15 +39,18 @@ def buildWorkload(self, arguments):
...
+        if arguments['OriginalRequestName'] == 'TaskChain' and not isinstance(arguments.get('TimePerEvent'), dict):
+            # TimePerEvent must be a dict object if user wants to override TaskChain tasks
+            pass
+        else:
+            helper.setTimePerEvent(arguments["TimePerEvent"])
```
ticoann: Similar to above change.

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -537,22 +537,9 @@ def setJobResourceInformation(self, timePerEvent=None, sizePerEvent=None, memory
...
-        if isinstance(memoryReq, dict):
-            taskMemory = memoryReq.get(self.name())
-        else:
-            taskMemory = memoryReq
-
-        if self.taskType() in ["Merge", "Harvesting", "Cleanup", "LogCollect"]:
-            # we don't want to touch memory for these task types
-            pass
-        elif taskMemory or getattr(performanceParams, "memoryRequirement", None):
+        if memoryReq or getattr(performanceParams, "memoryRequirement", None):
```
ticoann: So we won't pass the memoryReq for the types above?

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -537,22 +537,9 @@ def setJobResourceInformation(self, timePerEvent=None, sizePerEvent=None, memory
             performanceParams.timePerEvent = timePerEvent or getattr(performanceParams, "timePerEvent")
         if sizePerEvent or getattr(performanceParams, "sizePerEvent", None):
             performanceParams.sizePerEvent = sizePerEvent or getattr(performanceParams, "sizePerEvent")
-
-        # special handling for memory overwrite during assignment
-        if isinstance(memoryReq, dict):
-            taskMemory = memoryReq.get(self.name())
-        else:
-            taskMemory = memoryReq
-
```
ticoann: Shouldn't we keep this one for checking dict type?

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -537,22 +537,9 @@ def setJobResourceInformation(self, timePerEvent=None, sizePerEvent=None, memory
...
-        if self.taskType() in ["Merge", "Harvesting", "Cleanup", "LogCollect"]:
-            # we don't want to touch memory for these task types
-            pass
-        elif taskMemory or getattr(performanceParams, "memoryRequirement", None):
+        if memoryReq or getattr(performanceParams, "memoryRequirement", None):
             performanceParams.memoryRequirement = taskMemory or getattr(performanceParams, "memoryRequirement")
 
-        for task in self.childTaskIterator():
-            task.setJobResourceInformation(memoryReq=memoryReq)
-
```
ticoann: Shoudn't we keep this one incase memoryReq is dict?

src/python/WMCore/WMSpec/WMTask.py
ticoann: Also add for timePerEvent and sizePerEvents as well if those need to be overwritten?

src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
amaltaro: Much much better! :)

src/python/WMCore/WMSpec/WMTask.py
amaltaro: Ideally not. The way I see it is that the Workload object could see a dictionary value, but it then passes a string to the task object. However, I think if we clone a taskChain it could be that setupProcessingTask
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L336

would be handling a dictionary value.
I'll add this task.name() back here for all three arguments, but I think there is no need to call it for all the children from here.

src/python/WMCore/WMSpec/WMTask.py
amaltaro: That's right. We skip those task types and they get some default values in JobCreator (or submitter plugin).

src/python/WMCore/WMSpec/WMTask.py
amaltaro: This method is called when we create a new processing/production task and, since the children tasks (logcollect/merge/cleanup) don't have this setting, I think we no longer need it.

amaltaro (commented):
Thanks for the review Seangchan, I'm updating the code with your suggestions.

amaltaro:
test this please

src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
```diff
@@ -39,15 +39,12 @@ def buildWorkload(self, arguments):
...
-        if 'Memory' in arguments:
-            helper.setMemory(arguments['Memory'])
-        if 'Campaign' in arguments and not helper.getCampaign():
-            helper.setCampaign(arguments["Campaign"])
         if 'RequestPriority' in arguments:
             helper.setPriority(arguments["RequestPriority"])
-        if 'TimePerEvent' in arguments:
-            for task in helper.taskIterator():
-                task.setJobResourceInformation(timePerEvent=arguments["TimePerEvent"])
+        if arguments['OriginalRequestName'] != 'TaskChain' or isinstance(arguments['Memory'], dict):
```
amaltaro: `OriginalRequestType`

src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
```diff
@@ -39,15 +39,12 @@ def buildWorkload(self, arguments):
...
+            helper.setMemory(arguments['Memory'])
```
amaltaro: And this guy sets Memory for ALL tasks. I'll have to filter only Production/Processing task types to have this set, do you agree @ticoann ?

amaltaro (commented):
Bad job! Doesn't work!

amaltaro:
All tests are fine! I've also updated a few resubmission templates in #7950 
@ticoann can you review it again please? Thanks!

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -532,26 +532,22 @@ def setJobResourceInformation(self, timePerEvent=None, sizePerEvent=None, memory
         the three key values are main memory usage, time per processing unit (e.g. time per event) and
         disk usage per processing unit (e.g. size per event).
         """
+        if self.taskType() in ["Merge", "Harvesting", "Cleanup", "LogCollect"]:
```
amaltaro: BTW, it looks like we have never set these requirements for Harvesting tasks. Which means, even DQMHarvest workflows would be missing this memory setup.

For taskchain, it would be actually set to None, since there is no taskname (it's not a TaskChain task, though it's a task of the WMWorkload). For DQMHarvest workflow it would work because this setting would be just an integer.
I'm removing this filter "Harvesting" in this line.

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -641,17 +641,55 @@ def setCoresAndStreams(self, cores, nStreams):
...
+
+        Update TimePerEvent requirements for each task in the spec, thus it
+        can be either an integer or a dictionary key'ed by the task name.
+        """
+        # don't set it for utilitarian/merge tasks
+        if not timePerEvent:
+            return
+
+        if initialTask:
+            taskIterator = initialTask.childTaskIterator()
```
ticoann: If we set this way, aren't we missing to set the initialTask value itself?

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -641,17 +641,55 @@ def setCoresAndStreams(self, cores, nStreams):
...
+        else:
+            taskIterator = self.taskIterator()
+
+        for task in taskIterator:
```
ticoann: Isn't taskIterator only contains top level task?  it seems workload.taskIterator() only returns top level task unlike task.taskIterator()

ticoann:
@amaltaro, I still think there is some problem with this patch as mentioned in line. However you said all the test went fine, I will merge this. We can run the test in testbed. Thanks.

src/python/WMCore/WMSpec/WMWorkload.py
amaltaro: Yes, but then we pass this task as argument and recursively call this method L691, which then recursively iterates over the children.

ticoann: Yes, you're right \as always.  recursive is my enemy. :-)  Thanks Alan,

PR #7952, opened 2017-06-21
add site info for skipped file report

ticoann:
fixes #7925

PR #7946, opened 2017-06-20
merge cmssw version mismatch fix

ticoann:
fixes #7900 (hopefully)

PR #7937, opened 2017-06-15
Accept plain list of lumis in the Run object

amaltaro:
Fixes #7941
Got this exception [1] in WorkQueueManager while validating 1.1.4.patch1.

This Run object is somehow tricky with its arguments, besides the run number argument, the others are positional arguments, which means the class constructor sees a tuple of data, which can even be a tuple of a list of tuples :-) (see unit tests)

I'm open for suggestions on a better way or a better performance for fixing it.

[1]
```
2017-06-15 20:53:42,464:140585872336640:INFO:WMBSHelper:"amaltaro_ReReco_Parents_Agent114_Validation_170615_150800_1173" Injecting block /Cosmics/Commissioning2015-PromptReco-v1/RECO#77f9d258-b5d6-11e4-8732-02163e00d625 (89 files) into wmbs
2017-06-15 20:53:42,466:140585872336640:ERROR:WorkQueue:amaltaro_ReReco_Parents_Agent114_Validation_170615_150800_1173, /Cosmics/Commissioning2015-PromptReco-v1/RECO#77f9d258-b5d6-11e4-8732-02163e00d625: 
creating subscription failed in LQ: 
list index out of range
Traceback (most recent call last):
  File "/data/srv/wmagent/v1.1.4.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.1.4.patch1/lib/python2.7/site-packages/WMCore/WorkQueue/WorkQueue.py", line 374, in getWork
    dbsBlock)
  File "/data/srv/wmagent/v1.1.4.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.1.4.patch1/lib/python2.7/site-packages/WMCore/WorkQueue/WorkQueue.py", line 442, in _wmbsPreparation
    sub, match['NumOfFilesAdded'] = wmbsHelper.createSubscriptionAndAddFiles(block=dbsBlock)
  File "/data/srv/wmagent/v1.1.4.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.1.4.patch1/lib/python2.7/site-packages/WMCore/WorkQueue/WMBSHelper.py", line 417, in createSubscriptionAndAddFiles
    addedFiles = self.addFiles(block)
  File "/data/srv/wmagent/v1.1.4.patch1/sw/slc6_amd64_gcc493/cms/wmagent/1.1.4.patch1/lib/python2.7/site-packages/WMCore/WorkQueue/WMBSHelper.py", line 449, in addFiles
[...]

PR #7933, opened 2017-06-14
filter exceptions on error log for wmstats

ticoann:
fixes #7919 #7798

@amaltaro, Alan, this is initial patch for propagating the more relevant error messages to WMStats.
1. add the stageout failure messages in the exception message string.
2. filter out exceptions to report to WMStats.

I am not sure whether it would extract more meaningful messages but I will try to run the test to see.
We can discuss further tomorrow (you would have any better (and simple) idea to extract meaningful messages from condor log),

PR #7953, opened 2017-06-22
1.1.4 wmagent for patch2

ticoann:
Include following patch

#7933 filter exceptions on error log for wmstats
#7937  Accept plain list of lumis in the Run object
#7946 Fix merge cmssw version mismatch
#7952 Add site info for skipped file report
#7955 Add drain thread
#7957 Rollback unfinished transactions if ErrorHandler hits an exception
#7959 Fix a problem with new lumilist format
#7964 Raise HTTPException instead of ServerNotFoundError


ticoann:
@amaltaro, Alan I created the branch patch for the test #7943 is not included. Also I didn't merge to master either. I will run the quick test first.

amaltaro:
7943 is not added yet to this bundle. It's better to make it clear in the first comment such that it doesn't slip through the cracks.

amaltaro (changes requested):
We'll likely have to cherry-pick again #7933 and possible #7615. Without mentioned #7943 is not picked yet.

amaltaro:
Seangchan, I believe you're not picking the test commits, are you? That would explain those unit test failures.
BTW, if you have a slightly concern on not cherry-picking everything that is needed, I suggest creating a new PR from scratch to avoid surprises in patch2

ticoann:
create against wrong branch. Bad job. :-( replaced by #7965

PR #7959, opened 2017-06-23
Fix a problem with new lumilist format

mmascher:


ticoann:
@mmascher, Marco, do you need a new tag with this patch?

ticoann:
@amaltaro, did we see any problem with this case. I assume it returns "string" format in some cases? I am asking whether we should create wmagent branch patch for this.

ericvaandering:
I think this has to do with how CRAB is constructing them that the lumi numbers are not guaranteed to be integers, probably because the lumi numbers are now dict keys.

What I find interesting is that this code works with the dict of events/lumi now too. I never envisioned that usage. :-)

amaltaro:
Nope, I haven't spotted any issues with LumiList. Even if it didn't cause a real code execution problem, I would expect problems in the final outcome, but there where none as well.

PR #7960, opened 2017-06-23
Resubmission templates following the new creation mode

amaltaro:
New resubmission templates following the new mode of creating ACDCs.
Note that only `RequestPriority`, `Memory` and `TimePerEvent` can be overwritten during request creation, according to this code being merged soon: #7950


amaltaro:
All good, so merging it.

PR #7962, opened 2017-06-23
fix python3 compatibility (sorted)

ticoann:


ericvaandering:
This one wasn't necessary (but welcome). I'm not sure why it kept showing up. I think there was a problem with a stale JENKINS_ tag.

ticoann:
Thanks Eric, I will merge it anyway, then.

PR #7967, opened 2017-06-26
Add last row inserted to ResultSet

ericvaandering:
I need to save the inserted primary key between inserts in MySQL. I don't see another way in SQLAlchemy to do this or examples of how we do it. I'm happy to be wrong, @amaltaro or @ticoann

### DIFF
--- a/src/python/WMCore/Database/ResultSet.py
+++ b/src/python/WMCore/Database/ResultSet.py
@@ -15,6 +15,7 @@ class ResultSet:
     def __init__(self):
         self.data = []
         self.keys = []
+        self.insertedID = None
 
     def close(self):
         return
@@ -33,6 +34,7 @@ class ResultSet:
         myThread = threading.currentThread()
 
         if resultproxy.closed:
+            self.insertedID = resultproxy.lastrowid
             return
         elif resultproxy.returns_rows:
             for r in resultproxy:
--- a/src/python/WMCore/Database/ResultSet.py
+++ b/src/python/WMCore/Database/ResultSet.py
@@ -6,18 +6,17 @@ SQLAlchemy result sets (aka cursors) can be closed. Make this class look as much
 like the SQLAlchemy class to minimise the impact of adding this class.
 """
 
-
-
-
 import threading
 
-class ResultSet:
+
+class ResultSet(object):
     def __init__(self):
         self.data = []
         self.keys = []
         self.insertedID = None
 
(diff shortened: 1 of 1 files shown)


ticoann:
I remember Yuyi needed this before. Not sure how we added though.

It seems mysql there is way to do this, 
https://dev.mysql.com/doc/refman/5.7/en/getting-unique-id.html

```
INSERT INTO foo (auto,text)
    VALUES(NULL,'text');         # generate ID by inserting NULL
INSERT INTO foo2 (id,text)
    VALUES(LAST_INSERT_ID(),'text');  # use ID in second table
```

For Oracle, use "returning into" clause. but I am not sure the oracle version we uses support this.
https://oracle-base.com/articles/misc/dml-returning-into-clause

@yuyiguo, Yuyi, did you add this feature some where in WMCore?

Anyway, if the change works I don't see anything is simpler than that though.

ticoann:
Actually for Oracle. we can just use .currval

From here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMBS/Oracle/Files/Add.py#L15

```
sql = """INSERT INTO wmbs_file_details (id, lfn, filesize, events,
                                            first_event, merged)
             VALUES (wmbs_file_details_SEQ.nextval, :lfn, :filesize, :events,
                     :first_event, :merged)"""

```

wmbs_file_details_SEQ.curval will have the value just inserted

ericvaandering:
Right, @hufnagel clued me in on how to do this in Oracle. In MySQL that should work, but @yuyiguo did not add it to WMCore (I grepped for that string). I can see other uses where the ID may be useful, so maybe this is not such a bad idea.

ericvaandering:
Let's leave this unmerged. Yuyi's method works great too and is simpler. If I need IDs for some other reason, we can come back to this.

ticoann:
Either way is fine with me.

hufnagel:
Using .currval is dangerous, since you need to make sure that no one is advancing the sequence in between your calls. So basically single row insert followed by a select within the same transaction. If you have a bulk insert you'll only get the last sequence value, not all of the ones being inserted. If you don't have a transaction some other operation could change the sequence before you have a chance to look at .currval.

If you need to do this in Oracle it's much safer to just combine all the operations you want to do with the sequence value into one operation and use .nextval for all of them (within the same operation that doesn't advance).

PR #7970, opened 2017-06-27
add failed wf  message to log db

### LINKED ISSUE #7808: Report the reason why a workflow failed to be acquired
When we hit a terminal exception in WorkQueue, we should somehow propagate this information up to CompOps. Maybe in the ReqMgr2 template where it has the state transitions and etc.

ticoann:
fixes #7808
It should display to wmstats.
Following could be added
1. add the display in reqmgr2 page
2. send email to ops when it happens


ticoann:
@amaltaro, Alan, tested in vm. It shows under wmstats error logs tab. 
I will try to implement point 2. (maybe 1 as well) in separate patch.

PR #7972, opened 2017-06-27
Workunit job association 

ericvaandering:
This is a work in progress for a job to make and associate work units when the job is created. 

Not sure at the time of job creation is the best place to ultimately make work units, but for now it is expedient.


ericvaandering:
@amaltaro and @ticoann I'd like comments on this. This now starts creating work units in the database when jobs are created and when workflows are removed.  Also comments on additional unit tests I should perform are welcome and I'd like to see it tested in the next round of @amaltaro 's agent testing. It will have a performance impact, the question is how much.

amaltaro:
Given you know there will be a performance penalty, how about you make some performance tests? :-)

ticoann:
Hey Alan, I thought you are on vacation. :-) As far as performance testing, I am not sure how we can test it isolated. I assume performance hit comes from the work_units being many more than jobs and which access the queries those tables will impact on performance. I would assume the limit we have ~50K(I am not sure this is the right number) jobs per agent mostly depends on DB capability. In that case, we might have much less jobs each agent can handle.

ticoann:
@ericvaandering, As far as testing goes, we have to use below job splitting, right?
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/EventAwareLumiByWork.py

This is backward compatible, right? (We can still run other job splitting with this patch) 

Before we start testing, we have to list what to test, first correctness and then performance.
I guess we need to test whether final product is the same (same events, lumis) when jobs are 100 successful.
@amaltaro, any good idea what to test and how? - Not expecting to get an answer until after your vacation. :-)

Anyway, I think we need to have stable version of agent before we test this one.

ericvaandering:
Nope, it doesn't have anything to do with job splitting. You should start to see these extra tables get populated with any jobs that are assigned to run on files with lumis.

ericvaandering:
What I want to do is make sure it doesn't blow up by violating any constraints in the database, for instance. Then I think the plan @PerilousApricot and I hatched was that we could start doing correctness tests in JobArchiver to verify that jobs and work units both agree about what the work to be done still is.

ericvaandering:
Exactly, it's going to insert a couple of rows/lumi section/workflow into the database. Deletes will be massive but only once per workflow, so that should be quick.

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

ticoann:
@ericvaandering, Hi Eric, finally found the problem. 😄 After following line. 
self.associateWorkUnits() need to be added. 

https://github.com/ericvaandering/WMCore/blob/3d15b7727b4b944578f2e749047d9dfad5f42695/src/python/WMCore/WMBS/Job.py#L117

ericvaandering:
Thanks @ticoann !

ericvaandering:
test this please

ticoann:
test this please

ericvaandering:
@ticoann if you want to try this on your test machine, I think it's ready. It is putting the correct info when called from Subscription

ticoann:
@ericvaandering, Thanks Eric, It is working fine. All the data is populated in the workunit tables.
Could you squash the commit? I will merge it, then I will try to run some performance test.

ericvaandering:
Ok, it looks like this is ready @ticoann

ticoann:
Thanks Eric, I will try to run some performance test.

PR #7975, opened 2017-06-28
Gettask type

ericvaandering:
This code is currently unused except in bin/paused-jobs, but it doesn't work with long types which is what we get from the databases


ericvaandering:
@ticoann please merge

PR #7993, opened 2017-07-10
Add option to not override PSet numThreads for CRAB

emaszs:
The SetupCMSSWPset class is used in CRAB3 to adjust the PSet coming from the user [1]. However, a change last October had been made in SetupCMSSWPset [2] to always override the number of threads in the existing user PSet with what's in the `self.step.data.application.multicore`, something that CRAB3 doesn't really define. In case the value is undefined, after the change [2] PSet contents are always overwritten with `1` as the number of threads to be used. Therefore, currently CRAB3 jobs always use `1` thread in CMSSW regardless of how many cores were requested by the user.

[1] https://github.com/dmwm/CRABServer/blob/master/scripts/TweakPSet.py#L63
[2] https://github.com/dmwm/WMCore/commit/4b34707247559453470ee3087fa949666ac3ec9a

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -205,8 +205,9 @@ class SetupCMSSWPset(ScriptInterface):
                  "process.source.lumisToProcess": fixupLumisToProcess,
                  "process.source.firstLuminosityBlock": fixupFirstLumi}
 
-    def __init__(self):
+    def __init__(self, crabPSet=False):
         ScriptInterface.__init__(self)
+        self.crabPSet = crabPSet
         self.process = None
 
     def createProcess(self, scenario, funcName, funcArgs):
@@ -556,14 +557,9 @@ class SetupCMSSWPset(ScriptInterface):
         runIsComplete = getattr(baggage, "runIsComplete", False)
         multiRun = getattr(baggage, "multiRun", False)
         runLimits = getattr(baggage, "runLimits", "")
-        try:
-            cmsswVersion = self.step.data.application.setup.cmsswVersion
-        except AttributeError:
-            # CRAB3 needs to use an environment var to get the version
-            cmsswVersion = os.environ.get("CMSSW_VERSION", "")
 
         self.process.dqmSaver.runIsComplete = cms.untracked.bool(runIsComplete)
-        if multiRun and isCMSSWSupported(cmsswVersion, "CMSSW_8_0_X"):
+        if multiRun and isCMSSWSupported(self.getCmsswVersion(), "CMSSW_8_0_X"):
             self.process.dqmSaver.forceRunNumber = cms.untracked.int32(999999)
         if hasattr(self.step.data.application.configuration, "pickledarguments"):
             args = pickle.loads(self.step.data.application.configuration.pickledarguments)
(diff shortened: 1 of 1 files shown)


belforte:
is this really a sensible solution, to take a supposedly common low level function and make it do different things for CRAB and non-CRAB ?
To me it seems only to make things more complicate, diffcult to understand and most likely nobody will properly document what exactly SetupCMSSWPset.py does.
Wouldn't it be better to make it clear the need and usage of the multicore attribute and set in in CRAB jobs e.g. from numcored in the JDL ?
I.e. spend some time cleaning up the design, not adding code.

emaszs:
It really doesn't make sense for CRAB3 to load the pset, extract the number of cores requested and set a parameter that WMCore understands - only to have that same value put back in the PSet in the end. With this patch, SetupCMSSWPset simply won't touch the cores in the CRAB3 PSet, which sounds a lot simpler to me.

I think the design in this case is fine, our use-cases are slightly different (CRAB3 trusts what's in the PSet instead of relying on some other outside variable) but most of the relevant code is reused.

belforte:
I disagree. If this function wants a number of cores as input to set
the number of threads, we shoud pass the one in the JDL. Not extract it
form the PSet. If we do not want WMCore to change the PSet we need
a function from WMCore with the correct semantic to begin with.

The main problem is that what this function should do is not documented.
And this change will not help.

A secondary problem is that the function behavior was changed w/o
checking with the users. Did @amaltaro explain why communication
broke like that ? What will prevent this from happening again ?

Should CRAB simply stop using WMCore ? I am not sure what we are gaining. @ericvaandering ?

emaszs:
@belforte I do see your point about using what's in the JDL instead of parsing the PSet. But then why did we introduce a check in CRAB3 to enforce that `numCores == numThreads` (in crabConfig and PSet respectively)? Ignoring what's in the user's PSet would solve that issue as well as this one I think.

amaltaro:
> is this really a sensible solution, to take a supposedly common low level function and make it do different things for CRAB and non-CRAB ?

I agree that this script should have the same outcome, no matter if it's an analysis or production job.
Now the problem is how we accomplish it, CRAB does not use the required data structure that needs to be passed to SetupCMSSW (the job/step object configuration in addition to the pset file).
That's why in WMAgent we check for those attributes, because we build all the required objects and ship them to the worker node for that parsing.
CRAB, on the other hand, rely on os environment variables, at least for the CMSSW config.

My suggestion was to always override the number of cores on the worker node, but Emilis mentioned there is already a validation in place and CRAB has never overriden those PSet settings, that's why I suggested this approach.

Now onto the miscommunication, I was actually not aware that CRAB was also using this module, sorry. And I didn't know it was CRAB that was relying on os.env variables. We have so many legacy in the code, that I simply decided to remove that.

This patch is good to me and it will nicely distinguish when we're setting up a crab or a wmagent job, instead of simply having try/except blocks where nobody has an idea why handle it in a different way in the except block...

PS.: I think there are other parts of the code where we have a different handle between production and analysis,
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L559
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L623
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L645
 maybe we should update those too?

belforte:
thanks @amaltaro 
I can imagine good reasons for running 1-thread cmsRun in a multicore pilot. The question is how to make sure that this is only done when it is good rather than by mistake or confusion.
Anyhow I clearly see that we have much more documentation to do than what this topic would address. 
Do what you see better.

amaltaro:
test this please

amaltaro (approved):

ticoann:
test this please

amaltaro:
test this please

PR #7994, opened 2017-07-10
check schedd limit before pulling the work

### LINKED ISSUE #7559: Preventing Schedd to run at their limit
following up on #7509 discussion.
We should think hard on how to prevent the agents to push the schedd to its limit, by refusing more work when a certain limit is reached.
@bbockelm provided the classad for when the schedd stops matching. Is there an access to the limit and the number of jobs running (the agent has it anyways) ?

My idea is that agent should stop getting new work when at 80% of the schedd limit, and send an alarm to wmstats for P&R to react. 
Of course, the already acquired work would have to be handled.

@scarletnorberg FYI

ticoann:
fixes #7559 
Initial version (We can still add what Jean-Roch requested if necessary)

>can we have a >N hours after hitting CurbMatchmaking "no work acquired" cooloff period so that it does not start right back ? or is there already such cooloff period on the schedd side @bbockelm ?
>can we record the total number of running jobs when CurbMatchmaking is met, and use 80% of that value to prevent work for getting in ?can we have a >N hours after hitting CurbMatchmaking "no work acquired" cooloff period so that it does not start right back ? or is there already such cooloff period on the schedd side @bbockelm ?
>can we record the total number of running jobs when CurbMatchmaking is met, and use 80% of that value to prevent work for getting in ?

### DIFF
--- a/src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
+++ b/src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
@@ -5,14 +5,14 @@ _WorkQueueManagerPoller_
 Pull work out of the work queue.
 """
 
-
-
-
 import time
 import random
 import traceback
+import threading
 
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
+from WMCore.Services.PyCondor.PyCondorAPI import isScheddOverloaded
+
 
 class WorkQueueManagerWorkPoller(BaseWorkerThread):
     """
@@ -55,17 +55,28 @@ class WorkQueueManagerWorkPoller(BaseWorkerThread):
         """
         _retrieveCondition_
         set true or false for given retrieve condion
-        i.e. thredshod on workqueue
+        TODO: only condition it checks now is schedd limit
+        But we can add other conditions.
+        i.e. user config for enabling schedd limit check or threshold on workqueue, etc
         """
-        return True
+
+        return (not isScheddOverloaded())
 
     def pullWork(self):
         """Get work from parent"""
         self.queue.logger.info("Pulling work from %s" % self.queue.parent_queue.queueUrl)
         work = 0
+
+        myThread = threading.currentThread()
+
         try:
             if self.retrieveCondition():
                 work = self.queue.pullWork()
+                myThread.logdbClient.delete("LocalWorkQueue_pullWork", "warning", this_thread=True)
+            else:
+                msg = "Workqueue didn't pass the retrieve condition: NOT pulling work"
+                self.queue.logger.warning(msg)
+                myThread.logdbClient.post("LocalWorkQueue_pullWork", msg, "warning")
         except IOError as ex:
             self.queue.logger.error("Error opening connection to work queue: %s \n%s" %
                                     (str(ex), traceback.format_exc()))
@@ -81,6 +92,6 @@ class WorkQueueManagerWorkPoller(BaseWorkerThread):
         try:
             self.queue.processInboundWork()
         except Exception as ex:
-            self.queue.logger.exception('Error during split')
+            self.queue.logger.exception('Error during split: %s' % str(ex))
         self.logger.info('Splitting finished')
         return
--- a/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
+++ b/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
@@ -1,0 +1,19 @@
+from __future__ import print_function, division
+
+try:
+    # This module has dependency with python binding for condor package (condor)
+    import htcondor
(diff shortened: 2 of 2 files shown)


src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
```diff
@@ -55,9 +56,12 @@ def retrieveCondition(self):
...
         _retrieveCondition_
         set true or false for given retrieve condion
-        i.e. thredshod on workqueue
+        TODO: only condition it checks now is schedd limit
+        But we can add other conditions.
+        i.e. user config for enabling schedd limit check or thredshod on workqueue, etc
         """
-        return True
+
+        return (not isJobLimitReachedInSchedd())
```
bbockelm: How frequently is this called?  It does reload a few things on each invocation, meaning it would be fine to do it every few seconds - but not at 10Hz.

ticoann: There is some randomness to the polling cycle but in any case it should be > 3 min.  Thanks for the feedback.

src/python/WMCore/BossAir/PyCondorAPI.py
```diff
@@ -0,0 +1,12 @@
+import htcondor
+
+def isJobLimitReachedInSchedd():
```
amaltaro: I'd name it `isScheddOverloaded`, since it will not rely only on the job limit but on the duty cycle and *anything* else that the glideinWMS team decides to configure for this `CurbMatchmaking` ad.

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
```diff
@@ -66,6 +70,8 @@ def pullWork(self):
         try:
             if self.retrieveCondition():
                 work = self.queue.pullWork()
+            else:
+                self.queue.logger.info("Retrieve condition is not true: won't pull work")
```
amaltaro: How about we say `Workqueue didn't pass the retrieve condition: NOT pulling work`. 
I was going to say we could make it a warning log level to catch more attention, but maybe it's not a good fit in here... (?)

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
```diff
@@ -55,9 +56,12 @@ def retrieveCondition(self):
         """
         _retrieveCondition_
         set true or false for given retrieve condion
-        i.e. thredshod on workqueue
+        TODO: only condition it checks now is schedd limit
+        But we can add other conditions.
+        i.e. user config for enabling schedd limit check or thredshod on workqueue, etc
```
amaltaro: typos :)

amaltaro (commented):
Looks good in general. I'm just not sure that function is in the best place... maybe we should add that function to the BossAirAPI? But not really sure :)

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
ticoann: I think actually good place to add warning log to logDB, Thanks Alan,

src/python/WMCore/BossAir/PyCondorAPI.py
ticoann: I will change the name. Thanks Alan.

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
ticoann: copy and paste. :-(

ticoann:
BossAirAPI is more generic API (Not depending on the plugins), But current location is might not be good either. I was thinking if there is any python binding code (from SImpleCondorPlugin) we can move here as wrapped api later.

ticoann:
test this please

vlimant:
Hi, when is this thing going in production, or when did it go in production.
is there a wiki explaining what's the limit ?

amaltaro:
This feature is going to be in the upcoming production release, meant to be deployed by the end of the next week.

There is no wiki explaining it because it's out of WMCore hands, we simply use a condor classad provided by gWMS team (which btw can change without any notice).

More useful details here:
https://github.com/dmwm/WMCore/issues/7559#issuecomment-313994707

but in summary, if this expressions is evaluated to True:
`( ShadowsRunning > 9.700000000000000E-01 * 22000 ) || ( RecentDaemonCoreDutyCycle > 9.800000000000000E-01 )`

then local WQ won't pull any work from GQ, no matter the work priority.

PR #7996, opened 2017-07-11
return correct exit code

### LINKED ISSUE #7990: Stageout executor handles exit codes wrong
Hi,

I was tracking down some false positives at Vanderbilt (where stageouts were thought to have succeeded but they actually failed) and found a painful WMCore bug. See the following snippet, which is a simplified version of the actual code (https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/Execute.py#L77)

```python
Python 2.7.13 (default, Dec 17 2016, 23:03:43) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> import os
>>> import subprocess
>>> child = subprocess.Popen("exit 127", shell=True)
>>> err = child.wait()
>>> print err
127
>>> os.WIFEXITED(err)
False
>>> os.WIFSIGNALED(err)
False
>>> os.WIFSTOPPED(err)
True
>>> os.WSTOPSIG(err)
0
```

The code actually returns `0` to its caller, instead instead of `127`! This isn't good! In particular, that's the exit code that bash uses if it can't find a script. I'm going to add a separate PR to fix the vandy specific stuff to test for this, but otherwise that code probably needs to be fixed up.

ticoann:
fixes #7990

### DIFF
--- a/src/python/WMCore/Storage/Execute.py
+++ b/src/python/WMCore/Storage/Execute.py
@@ -7,86 +7,30 @@ Run the stage out commands in a nice non-blocking way
 """
 from __future__ import print_function
 
-import fcntl
-import os
-import select
-import sys
 from subprocess import Popen, PIPE
 
 from WMCore.Storage.StageOutError import StageOutError
-
-
-def makeNonBlocking(fd):
-    """
-    _makeNonBlocking_
-
-    Make the file descriptor provided non-blocking
-
-    """
-    fl = fcntl.fcntl(fd, fcntl.F_GETFL)
-    try:
-        fcntl.fcntl(fd, fcntl.F_SETFL, fl | os.O_NDELAY)
-    except AttributeError:
-        fcntl.fcntl(fd, fcntl.F_SETFL, fl | fcntl.FNDELAY)
-
 
 def runCommand(command):
     """
     _runCommand_
 
-    Run the command without deadlocking stdou and stderr,
-    echo all output to sys.stdout and sys.stderr
+    Run the command without deadlocking stdout and stderr,
 
     Returns the exitCode
 
     """
     # capture stdout and stderr from command
-    child = Popen(command, shell=True, bufsize=1, stdin=PIPE, stdout=PIPE, stderr=PIPE, close_fds=True)
-    child.stdin.close()  # don't need to talk to child
-    outfile = child.stdout
-    outfd = outfile.fileno()
-    errfile = child.stderr
-    errfd = errfile.fileno()
-    makeNonBlocking(outfd)  # don't deadlock!
-    makeNonBlocking(errfd)
-    outdata = errdata = ''
-    outeof = erreof = 0
-    while True:
-        ready = select.select([outfd, errfd], [], [])  # wait for input
-        if outfd in ready[0]:
-            try:
-                outchunk = outfile.read()
-            except Exception as ex:
-                msg = "Unable to read stdout chunk... skipping"
-                print(msg)
-                outchunk = ''
-            if outchunk == '': outeof = 1
-            sys.stdout.write(outchunk)
-        if errfd in ready[0]:
-            try:
-                errchunk = errfile.read()
-            except Exception as ex:
-                msg = "Unable to read stderr chunk... skipping"
-                print(msg, str(ex))
-                errchunk = ""
-            if errchunk == '': erreof = 1
-            sys.stderr.write(errchunk)
-        if outeof and erreof: break
-        select.select([], [], [], .1)  # give a little time for buffers to fill
+    child = Popen(command, shell=True, bufsize=1, stdin=PIPE, close_fds=True)
 
-    err = child.wait()
-    if os.WIFEXITED(err):
-        return os.WEXITSTATUS(err)
-    elif os.WIFSIGNALED(err):
-        return os.WTERMSIG(err)
-    elif os.WIFSTOPPED(err):
-        return os.WSTOPSIG(err)
-    return err
+    child.communicate()
+    retCode = child.returncode
 
+    return retCode
 
 def runCommandWithOutput(command):
     """
-    _runCommand_
+    _runCommandWithOutput_
 
     Run the command without deadlocking stdout and stderr,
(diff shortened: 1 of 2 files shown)


ticoann:
Not tested yet.

src/python/WMCore/Storage/Execute.py
```diff
@@ -96,48 +23,14 @@ def runCommandWithOutput(command):
...
 
-    err = child.wait()
-    if os.WIFEXITED(err):
-        err = os.WEXITSTATUS(err)
-    elif os.WIFSIGNALED(err):
-        err = os.WTERMSIG(err)
-    elif os.WIFSTOPPED(err):
-        err = os.WSTOPSIG(err)
-    return err, output
+    retCode = child.wait()
```
amaltaro: According to the official documentation, this call can deadlock if piped output (stdout and stderr) is too large.
Maybe we could keep only the communicate call and in the next line retrieve the return code with `child.returncode`?

src/python/WMCore/Storage/Execute.py
```diff
@@ -96,48 +23,14 @@ def runCommandWithOutput(command):
...
+    sout, serr = child.communicate()
+
+    # If child is terminated by signal, err will be negative value. (Unix only)
+    sigStr = "Terminated by signal %s\n" % -retCode if retCode < 0 else ""
```
amaltaro: Are you making it positive because we check (somewhere else) whether return code is > 0?
Unless I misunderstood you, I'm in favor of not changing the exit code.

src/python/WMCore/Storage/Execute.py
```diff
@@ -96,48 +23,14 @@ def runCommandWithOutput(command):
...
+    output = "%sstdout: %s\nstderr: %s" % (sigStr, sout, serr)
+    return abs(retCode), output
```
amaltaro: and same here, unless there is a good reason to change it.
PS.: the ouput above won't be nicely formatted in case there is a sigStr value

amaltaro (commented):
Seangchan, I made a few comments along the code. Besides that, I like it, the code is much much cleaner now :)

src/python/WMCore/Storage/Execute.py
ticoann: yes, you're right. I will change to keep only communicate function. Thanks.

src/python/WMCore/Storage/Execute.py
ticoann: With the same reason, I can keep the negative value.  output should be OK since there is '\n', isn't it?

ticoann:
@amaltaro Alan I updated as you suggested. Please review again. Thanks,

amaltaro (approved):
Code looks good, thanks! As we discussed, I'm going to run some tests until tomorrow or Friday and then we can backport it

bbockelm (changes requested):
I like the general approach of using library code instead of WMCore-specific code.

I think this gets the return code handling correct.

However, this changes two things that I don't like:
1.  The `stdout` / `stderr` is buffered in memory instead of being dumped straight to the WMCore out/err.
2. `stdout`/`stderr` is no longer intermixed but printed out separately.

Is it possible to set the `Popen` object's `stdout` and `stderr` to `sys.stdout` and `sys.stderr`, respectively?

ticoann:
Thanks Brian for the review. 

> However, this changes two things that I don't like:
> 
> The stdout / stderr is buffered in memory instead of being dumped straight to the WMCore out/err.

I didn't quite understand about "WMCore out/err"? Are you saying instead of putting in memory and print out, just pass to sys.stdout? If that is the case, one problem I am having is that we save that output in error case and stick to the exception massage which later extracted then reported to wmstats. 

This code is used only for Stageout, Are we expecting a lot of output when commend succeed.

> stdout/stderr is no longer intermixed but printed out separately.

Hmm, I am not sure how to fix this with generic code.

bbockelm (approved):
Missed in my prior review that the intent of the function is to buffer the response in memory.  So, there's nothing we can improve on here.

Accordingly, setting my feedback to approve.

ticoann:
test this please

ticoann:
@amaltaro, Alan actually there are places still use runCommand method. I added back the code. Could you test again? Thanks,

amaltaro:
test this please

amaltaro:
Thanks, Seangchan! Looks much better now and all DMWM workflows ran without any issues (at CERN and FNAL).

I'm going to create a new issue to replace that call on all the stage out plugins. I see no reason to keep using one plugin that doesn't return the stdout/err.

ticoann:
test this please

amaltaro:
This specific patch works fine.

However, we still have an issue with the stage out commands themselves. I did a test with an nonexistent command
```
xrdcp-old-old -d 0 -f  /storage/local/data1/condor/execute/dir_1228949/glide_Fc7TQJ/execute/dir_1242379/job/WMTaskSpace/cmsRun1/RAWSIMoutput.root  root://cmsdcadisk01.fnal.gov//dcache/uscmsdisk/store/unmerged/DMWM_Test/Pythia8_Dijet15_pp502_TuneCUETP8M1/GEN-SIM/MonteCarlo_Ext_PhEDEx_July_Validation_TEST_Alan_v17-v11/00000/545E4D98-9F6B-E711-85F4-0025905B8606.root\?eos.targetsize=16075518\&eos.checksum=ae4f996a ; if [ $? -eq 0 ] ; then exit 0; else echo "Error: xrdcp exited with $?"; exit 60311 ; fi
```

and this was the error message:
```
Message: 2017-07-18T05:25:45 : Command exited non-zero ExitCode:151
Output: (stdout: Error: xrdcp exited with 1
  
stderr: /bin/sh: xrdcp-old-old: command not found
)
```

The problem lies on the shell variable `$?`, which always has the exit status from the latest command executed. Which means, when the code reaches the `else` clause, it has the exit status of the `if` statement, and the exit status of the xrdcp command is gone now.

IMO, we should first fix those stage out commands and then we patch the production agents. Anyways this execute issue is there for several months already.

PR #8000, opened 2017-07-12
Support DeterministicPileup in StepChain

### LINKED ISSUE #7977: StepChain: support DeterministicPileup
Apparently it's not supported in this spec. Or it's not supported for EventBased, I have to check ...

amaltaro:
Fixes #7977

Still need to be tested

To update doc specs and a few json templates.

### DIFF
--- a/doc/createSpecs/StepChain_createSpec.json
+++ b/doc/createSpecs/StepChain_createSpec.json
@@ -28,7 +28,12 @@
         "default": "https://cmsweb.cern.ch/couchdb", 
         "optional": true, 
         "type": "<type 'str'>"
-    }, 
+    },
+    "DeterministicPileup": {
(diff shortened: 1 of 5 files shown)


amaltaro:
Actually, it's going to be better if we make it a StepChain top level argument - just as it's for TimePerEvent and SizePerEvent - since it's part of the `SplittingArguments` settings. Otherwise we'd have to check every single step dictionary before we create the first step (task).

amaltaro:
@anorkus FYI, we're adding DeterministicPileup support in StepChains. You possibly have it already for TaskChains (in McM), but for StepChain it's a top level argument (first commit has that change).

amaltaro:

ticoann (approved):

Issue #8002, opened 2017-07-13
Add validation to MergedLFNBase/UnmergedLFNBase arguments


amaltaro:
We have these LFNs defined here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Tools/cms.py#L68

and unmerged just below it. I suggest we create a validate function in the Lexicon module and call that function from the StdSpecs validate function.

Or even better, we could have these MergedLFNBase lfns in the aux db

In addition to that:
* remove the assignment default values
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1092
* I don't know how correct are those "/store/results/" lfns, thus I'd suggest to validate it on `startswith('/store/results/')` and that's all, they are rarely used anyways and we can always enforce it to be assigned via script.

@ticoann what do you think?
BTW, slightly related to #7979

amaltaro:
BTW2: Ops does not rely on the default assignment values:
https://github.com/CMSCompOps/WmAgentScripts/issues/235

hufnagel:
The current validation breaks the Tier0.

AssertionError: '/store/backfill/1/data/Tier0_REPLAY_vocms001/HLTPhysics/RAW/v125' does not match regular expression /store/results/([a-zA-Z0-9\-_]+)/([a-zA-Z0-9\-_]+)/([a-zA-Z0-9\.\-_]+)/([A-Z\-_]+)/([a-zA-Z0-9\-_]+)

This prevents the latest T0 release from being used (it "worked" before since we didn't override the LFN at assignment and got the wrong/default values).

ticoann:
@amaltaro, Alan, all suggestions sounds good to me.
@hufnagel, Dirk, you are validating LFN base, right? Actually that is just last checking. For T0, the actual validation happens below.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L393

Is that the all the string? Is seems '(counter)' part is missing.

hufnagel:
The counter is not part of the base, never has been. It's injected by WMCore itself, like for any other workflow.

hufnagel:
Isn't the counter injected into the LFN in runtime stageout code ?

hufnagel:
Ah, no, these are the directories based on the run number. Let me check where they are added...

ticoann:
[...]

PR #8006, opened 2017-07-14
corrects Tier0 LFN bases during assignment, fixes #8001

### LINKED ISSUE #8001: Replays not writing to the backfill area on eos
Hello:

We have noticed that the output from the recent replays is not going to the
/eos/cms/tier0/store/backfill/1/data/ area, but to the /eos/cms/tier0/store/data/ area. We checked the Tier0  configuration and everything seems to be properly set up.  Operationally it not causing particular problems right now, but it is not the expected behavior. Discussing it with @hufnagel, it is likely related to the general parameter reorganization in StdBase.

--
Regards,

John

hufnagel:
Really ugly, but the only way to handle this right now. Underlying WMCore code should probably be restructured so this can be done in a less ugly way. Some related discussion in #8002

### DIFF
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -654,7 +654,7 @@ class WMWorkloadHelper(PersistencyHelper):
             task.setJobResourceInformation(memoryReq=memory)
         return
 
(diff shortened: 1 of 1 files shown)


hufnagel:
Superseeded by #8007

PR #8007, opened 2017-07-16
Remove assignment default value for un/mergedLFN bases

### LINKED ISSUE #8002: Add validation to MergedLFNBase/UnmergedLFNBase arguments

amaltaro:
Partially fixes #8002 
Possibly superseeds #8006 

I still have to test it.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -1088,8 +1088,8 @@ class StdBase(object):
                                               "validate": lambda x: x in ["Low", "Normal", "High"]},
                      "DeleteFromSource": {"default": False, "type": strToBool},
                      # merge settings
-                     "UnmergedLFNBase": {"default": "/store/unmerged"},
-                     "MergedLFNBase": {"default": "/store/data"},
+                     "UnmergedLFNBase": {"assign_optional": True},
+                     "MergedLFNBase": {"assign_optional": True},
                      "MinMergeSize": {"default": 2 * 1024 * 1024 * 1024, "type": int,
                                       "validate": lambda x: x > 0},
                      "MaxMergeSize": {"default": 4 * 1024 * 1024 * 1024, "type": int,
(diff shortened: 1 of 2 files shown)


hufnagel:
This is not going to fix the Tier0 problem since the setAcquisitionEra, setProcessingString and setProcessingVersion calls all reset the LFNBases for the tasks. Without these calls being run aware this is not going to work for the Tier0.

amaltaro:
But don't you set those during workflow creation? Then you simply don't set them during assignment again (as you're going to do for the LFN bases)

hufnagel:
Hm, so I would only pass a minimal set of parameters to updateArguments, like before. I can give that a try.

amaltaro:
That's right! Since T0 used to only "create" requests, I expect that to work out of the box and there is no need to override any of the output datasets related arguments during assignment.

Please let me know how it goes on the T0 side. I'm checking it on normal production env.

hufnagel:
Ok, your patch doesn't work. Now for figuring out why...

hufnagel:
Hm, let me test that some more. I think I might have looked at the wrong metrics plus there was interference from some other changes in that replay.

hufnagel (approved):
Seems to be working in Tier0 replays, code change also looks good. One thing, shouldn't we remove that FIXME comment line once we validate the changes ?

amaltaro:
We already know how to validate these LFNs bases, but it's still not in place.
Thanks for testing it! I still have to check things on my end.

hufnagel:
Found another problem. Can you also wrap the call to setMergeParameters and only call it if kwargs actually contains merge parameters ? This completely screws with any workflow that defines the forceMerged since that works by setting the merge parameters to 0 (so we always got to merged).

ticoann (approved):

amaltaro:
@ticoann from my side we can merge it. Feel free to go ahead and merge it please.

PR #8009, opened 2017-07-17
fix for forceMerged output

hufnagel:
Some changes in StdBase and workflow assignment broke the forceMerged output functionality.

### DIFF
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -974,7 +974,7 @@ class WMWorkloadHelper(PersistencyHelper):
         for task in taskIterator:
             for stepName in task.listAllStepNames():
                 stepHelper = task.getStepHelper(stepName)
-                if stepHelper.stepType() == "StageOut" and stepHelper.minMergeSize() != -1:
+                if stepHelper.stepType() == "StageOut" and stepHelper.minMergeSize() > 0:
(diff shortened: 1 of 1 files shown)


hufnagel:
Tested in Tier0 replay. Without this patch Express AlcaSkim output is written unmerged, even though the task was created with the forceMerged=True flag. With this patch things work as expected.

hufnagel:
@amaltaro please review

hufnagel:
Secondary issue, I thought we always use minMergeSize=0 to trigger forcing straight to merge. Do we really use -1 in other places for this. Longterm we should maybe unify this (if there is a discrepancy).

amaltaro:
Just to make sure I understood it.
When you create the alcaSkim task, you force it to merged which then calls it
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L394

but since we have default values during assignment (which are there regardless whether you pass them or not) it tries to set again those merge parameters. As `-1 != 0`, it sets those again.

Patch looks ok!

amaltaro (approved):

hufnagel:
Yes, exactly. With the default merge parameters back, alcaSkim output is again unmerged (since it's small) and since there is no merge task after alcaSkim, that output never makes it to merged and it never shows up in DBS and PhEDEx.

PR #8011, opened 2017-07-18
get draining parameter from aux db

ticoann:
replace #7731 #7923

Also #8023 need to be added (won't be merged until new agent is ready)

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Auxiliary.py
+++ b/src/python/WMCore/ReqMgr/Service/Auxiliary.py
@@ -158,17 +158,25 @@ class AuxBaseAPI(RESTEntity):
 
 
     def validate(self, apiobj, method, api, param, safe):
-        pass
+        args_length = len(param.args)
+        if args_length == 1:
+            safe.kwargs["subName"] = param.args.pop(0)
+            return
+        return
 
 
     @restcall(formats=[('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
-    def get(self):
+    def get(self, subName=None):
         """
         Return entire self.name document
-
+        subName is subcategory of document which is added as postfix string
         """
         try:
-            sw = self.reqmgr_aux_db.document(self.name)
+            if subName:
+                docName = "%s_%s" % (self.name, subName)
+            else:
+                docName = self.name
+            sw = self.reqmgr_aux_db.document(docName)
             del sw["_id"]
             del sw["_rev"]
         except CouchNotFoundError:
@@ -177,7 +185,7 @@ class AuxBaseAPI(RESTEntity):
         return rows([sw])
     
     @restcall
-    def post(self):
+    def post(self, subName=None):
         """
         post sofware version doucment
         """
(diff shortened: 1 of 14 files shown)


ticoann:

ticoann:
test this please

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -57,14 +57,76 @@ def _getDataFromMemoryCache(self, callname):
...
+        """
+        retrieve agent configuration reqmgr aux db.
+        """
+        return self._getDataFromMemoryCache('wmagentconfig/%s' % agentName)
+
+    def postWMAgentConfig(self, agentName, agentConfig):
+
+        return self["requests"].post('wmagentconfig/%s' % agentName, agentConfig)[0]['result']
+    
+    def isWMAgentDraining(self, agentName, agentDrainCondition=None):
```
amaltaro: Now that I finished reading this module. I think we should separate getters from setters. And this method is clearly a getter (by its name), but it also uploades the couch record.

Just an idea, whatever is calling `isDrainMode` should first call a `updateDrainMode` or something like that and only then call `isDrainMode`. 
Second option would be renaming this method to `updateDrainMode` and moving the final logic to `isDrainMode`.
Third option is something better that you'll probably suggest :)

amaltaro (changes requested):
A few minor things to fix and possible some other minor improvements. Until those are fixed, you have my red flag :-)
And sorry for taking so long to review it!

bin/deploy-wmagent.sh
```diff
@@ -312,6 +312,12 @@ else
 fi
 echo "Done!" && echo
 
+### Upload WMAgentConfig to AuxDB
+echo "*** Upload WMAgentConfig to AuxDB ***"
+cd $MANAGE
+./manage execute-agent wmagent-upload-config
+echo "Done!" && echo
```
ticoann: Yes, you're right. I will do that.

ticoann:
test this please

ticoann:
test this please

src/python/WMCore/ReqMgr/Service/Auxiliary.py
```diff
@@ -186,10 +195,41 @@ def post(self):
...
+        if subName:
+            docName = "%s_%s" % (self.name, subName)
+        else:
+            docName = self.name
+        doc = Document(docName, doc)
         result = self.reqmgr_aux_db.commitOne(doc)
         return result
 
+    @restcall(formats=[('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
+    def put(self, subName=None):
```
ticoann: Hmm, I didn't see previous reply for this. (I am pretty sure I did).  
The problem is there is "_"  prefix from subName so even if we have empty sting as default we still need a if else statement.

src/python/WMCore/ReqMgr/Service/Auxiliary.py
```diff
@@ -186,10 +195,41 @@ def post(self):
...
+            propertyDict = json.loads(data)
+
+        try:
+            result = None
+            if subName:
+                docName = "%s_%s" % (self.name, subName)
+            else:
+                docName = self.name
+
+            if propertyDict:
```
ticoann: Every property values is required explicitly. So if agent is out of drain, it should have explicitly say {"AgentDrainMode": False} and current code (with this patch) will do that.

PR #8014, opened 2017-07-21
Set PU caching to 0.5h instead of 24

### LINKED ISSUE #7894: Not all files of the MCPileup are added to the configuration
following up on https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/4066.html where the software needs all file of the secondary input to be listed in the mix module, @thongonary and I were looking at a specific job log

/eos/cms/store/logs/prod/2017/05/WMAgent/rverma_PR_newco_RelVal_276950_170428_191716_3771/rverma_PR_newco_RelVal_276950_170428_191716_3771-Task1MergeDQMoutputMergedEndOfRunDQMHarvestLogCollect-cmswn2203-1-logs.tar

/WMTaskSpace/logCollect1/2a1532f6-465e-11e7-a342-a0369f23d01e-488-1-logArchive.tar.gz

and the PSet has 2423 files in mixing module while the full dataset does have 5660 files in total, and is in full at the site showing errors about this workflow.

the suspicion is that that secondary file list was truncated at pset tweak time one way or the other.

It's the first time we (at least I) encouter such type of request requiring exactly all files of the secondary to be present.

amaltaro:
Fixes #7894

As discussed in the meeting, I'm simply reducing the cache lifetime to 30min instead of 24h. If we find that it still doesn't meet our needs, then we can think of something more complete/complicated.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
+++ b/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
@@ -141,7 +141,7 @@ class PileupFetcher(FetcherInterface):
         """
         cacheFile = self._getCacheFilePath(stepHelper)
 
-        if not self._isCacheExpired(cacheFile) and os.path.getsize(cacheFile) > 0:
+        if not self._isCacheExpired(cacheFile, delta=0.5) and os.path.getsize(cacheFile) > 0:
             # if file already exist don't make a new dbs call and overwrite the file.
             # just return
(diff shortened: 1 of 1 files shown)


amaltaro:
Just measured (from the logs) a pretty large PU dataset and it took around a minute to build the json file.
Also made sure it works. I'm going to backport it and apply to all the agents with your patch 8005.

amaltaro:
test this please

ticoann:
Thanks Alan!

PR #8015, opened 2017-07-21
Set PU caching to 0.5h instead of 24 - wmagent version

amaltaro:
wmagent version of #8014


amaltaro:
Seangchan, I'm going to patch the production agents now with this one and 8005 .

amaltaro:
[Fix #8015 - Set PU caching to 0.5h instead of 24](https://trello.com/c/ZjNpRI3s/51-fix-8015-set-pu-caching-to-05h-instead-of-24)

PR #8018, opened 2017-07-24
Inherit PrepID from the workload level for StepChain

### LINKED ISSUE #7978: StepChain: tricky merge PrepID inheritance (correct for TaskChains)
This is how we made StepChain support different PrepID settings in a Step basis:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L671

since it's a merge task, it should have the same arguments as the parentTask (production/processing). So far so good!

Problem is, it should inherit the **workload** value for StepChains. Following the standard behaviour: if it's not passed in the task, then inherit it from the top level (workload).
For StepChain, the Step1 settings is what defines the task level arguments.

Honestly, I think we should get rid of of this inheritance and make this stuff mandatory at Task/Step level. I believe McM already does that...

amaltaro:
Fixes #7978

Make sure StepChain inheritance comes from the workflow level and not from the task one (which is set according to the first step settings).

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
@@ -47,6 +47,7 @@ class StepChainWorkloadFactory(StdBase):
         self.sizePerEvent = None
         self.timePerEvent = None
         self.primaryDataset = None
+        self.prepID = None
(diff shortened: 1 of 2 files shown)


amaltaro:
test this please

ticoann (approved):

PR #8021, opened 2017-07-25
Update subscription methods to keep list of datasets under an output module

### LINKED ISSUE #7981: StepChain: setSubscription doesn't work with dup output modules
Apparently related to this issue https://github.com/dmwm/WMCore/issues/6787 , regardless whether one is set to KeepOutput=False.

This is the code: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMWorkload.py#L1181

amaltaro:
Fixes #7981 

Summary of changes are:
* set the dataset subscriptions at the very end of the workflow assignment (thus no need for updating them)
* removed the update subscription method (set subscription rely on the list of output datasets for a given task)
* subscriptions section are unique inside the same task (no longer using a possible dup output module)


amaltaro:
test this please

amaltaro:
@hufnagel with these changes (well, only the subscriptions change, previous branch has to be merged first), I assume setSubscriptionsInfo is called only once in a workflow lifetime, since it's an assignment argument.

Do you think it could cause problems for T0? Are you setting those subscription info while you create a workload?

hufnagel:
Tier0 sets subscription information directly between creation and "assignment". 

https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L664

https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L669

https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L1069

Take a look and decide yourself if your changes break it. One thing, Tier0 sets different subscriptions for the same primary dataset, depending on dataTier. For skim output it even sets different subscriptions for output that that identical primary dataset and data tier and just differs in the filter name (skim).

hufnagel:
A quick read of your changes I don't think this works. At all.

amaltaro:
But once T0 creates the workload, the output datasets are not changed (acqera, procstr, lfn, etc), are they?
And from what you said, you may call setSubscriptions several times for the same workload, such that you can get all the different subscriptions. I think there is a unit test - under WMTask - to check this case and it works, so this patch might work as is. Still need to test though

hufnagel:
Correct, once the workload is created, we don't change output modules or datasets. We call setsubscriptionInfo once, but we do pass a list of subscriptions. Can be a long list.

amaltaro:
From the unittests, it works for production but it would break the T0 use case, since T0 can have different subscription info for datasets under the same outputmodule section. Again a mapping that's no longer 1 to 1.

Since we're moving towards the multi-step tasks, we should name the subscriptions section after the dataset name instead of the output module. That's the "least hard" way for accomplishing multi duplicate output module with the required T0 flexibility, though that will have some significant changes on the agent side too :(

amaltaro:
Let's make breaking changes then :-) (I know I know it has to be backwards compatible)

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Seems to work fine. I'll just update this patch such that the agent can handle old workloads (current ReqMgr version)

amaltaro:
I was going to forget to squash these commits. Just did.

ticoann:
Thanks Alan, I will merge and create and create a new tag.

amaltaro:
You're the man! Thanks!

ticoann:
But what are you doing this late? Stop working Alan. 😄

ticoann:
test this please

ticoann (approved):

ticoann:
@amaltaro, Alan 1.1.6.pre3 is created

PR #8023, opened 2017-07-25
add deployment script

ticoann:
depends on #8011 Do not merge until new agent is ready

### DIFF
--- a/bin/deploy-wmagent.sh
+++ b/bin/deploy-wmagent.sh
@@ -312,6 +312,12 @@ else
 fi
 echo "Done!" && echo
 
+### Upload WMAgentConfig to AuxDB
+echo "*** Upload WMAgentConfig to AuxDB ***"
+cd $MANAGE
+./manage execute-agent wmagent-upload-config
+echo "Done!" && echo
+
 ###
 # set scripts and specific cronjobs
 ###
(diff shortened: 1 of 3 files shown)


amaltaro:
Seangchan, this bin script should have been merged with 8011. Otherwise we need to cut another tag when this deployment fix is merged.

ticoann:
Hi Alan, yes. But didn't we agree that we won't merge this until we need to deploy new agent with a new tag. Actually it seems we can just need to change the deploy instruction so it can work with different tags.

```
wget -nv https://raw.githubusercontent.com/dmwm/WMCore/master/bin/deploy-wmagent.sh
->
wget -nv https://raw.githubusercontent.com/dmwm/WMCore/1.1.6.pre4/bin/deploy-wmagent.sh
```

amaltaro:
Yes, for the deployment script. The other script (wmagent-upload-config) is sort of part of the WMCore base code, since we need to execute it during deployment (meaning, we'd have to patch the agent during deployment to make it work).

We could indeed update the deployment instructions, but then it's another thing that we need to keep track and keep updating. Taking the master one is always easier :)

ticoann:
Alan, I updated the patch do have default location in separate location. I am not sure that is the best location. Also I think we need to have some discussion on the structure of the config.

amaltaro (approved):
Patch looks good to me. As we discussed, feel free to merge it.

PR #8027, opened 2017-07-27
more selective assignment overrides

hufnagel:
Tier0 subscriptions are completely broken in the current WMCore since updateArguments overrides all subscriptions and sets default values. Tier0 does NOT use spec arguments to set subscriptions, we make the subscription calls ourselves.

Not sure about performance monitoring or block closing, but most likely they are also affected (if they use defaults). Tier0 also uses custom values here.

I don't necessarily want this version merged, but there is an urgency here since we want to deploy the latest Tier0 release ASAP. Suggest another patch if you don't like this version.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Express.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Express.py
@@ -231,11 +231,6 @@ class ExpressWorkloadFactory(StdBase):
                                            periodic_harvest_interval=self.periodicHarvestInterval,
                                            doLogCollect=True)
 
-        workload.setBlockCloseSettings(self.blockCloseDelay,
-                                       workload.getBlockCloseMaxFiles(),
-                                       workload.getBlockCloseMaxEvents(),
-                                       workload.getBlockCloseMaxSize())
-
         # setting the parameters which need to be set for all the tasks
         # sets acquisitionEra, processingVersion, processingString
         workload.setTaskPropertiesFromWorkload()
@@ -482,8 +477,6 @@ class ExpressWorkloadFactory(StdBase):
                     "AlcaHarvestDir": {"optional": False, "null": True},
                     "AlcaSkims": {"type": makeList, "optional": False},
                     "DQMSequences": {"type": makeList, "attr": "dqmSequences", "optional": False},
-                    "BlockCloseDelay": {"type": int, "optional": False,
-                                        "validate": lambda x: x > 0},
                     "Outputs": {"type": makeList, "optional": False},
                     "MaxInputRate": {"type": int, "optional": False},
                     "MaxInputEvents": {"type": int, "optional": False},
--- a/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
@@ -102,11 +102,6 @@ class PromptRecoWorkloadFactory(DataProcessing):
                     self.addMergeTask(alcaTask, self.procJobSplitAlgo, alcaOutLabel,
                                       doLogCollect=self.doLogCollect)
 
-        workload.setBlockCloseSettings(self.blockCloseDelay,
-                                       workload.getBlockCloseMaxFiles(),
-                                       workload.getBlockCloseMaxEvents(),
-                                       workload.getBlockCloseMaxSize())
-
         # setting the parameters which need to be set for all the tasks
         # sets acquisitionEra, processingVersion, processingString
         workload.setTaskPropertiesFromWorkload()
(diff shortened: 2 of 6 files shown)


ticoann:
Thanks Dirk. Sorry for the trouble. I think we can merge this. Ideally, we can change the updateArgument method  to depend on the spec definition later instead of assuming all the specs has the same parameters to be updated.

hufnagel:
@amaltaro , any suggestions ? For the performance monitoring and the block closing I think I can modify Tier0 code to pass the appropriate kwargs to updateArguments.

For the subscriptions this is impossible, the syntax does not support what we need. Is simple having a conditional on just CustodialSites enough ?

hufnagel:
Lets give @amaltaro a chance to reply, no need to rush merging this. As I said, I can at least remove the the conditional for the performance monitoring and block closing since I can actually pass the values from Tier0 code to the updateArguments call.

For the conditional on the subscription, should I check for more (all ?) values or is the one enough ?

ticoann:
> For the conditional on the subscription, should I check for more (all ?) values or is the one enough ?

I think you have to check all. Since those values seems to be optional and default value can be passed (empty list []).
How about instead of using .get("something")  

```
if "something" in kwargs:
```
I think if you do that you can just check one parameter. Assuming tier0 is not passing anything.

hufnagel:
I used the get() because that is what @amaltaro used for the check on the LFN bases...

amaltaro:
I think the subscription has to depend at least on `NonCustodialSites` and `CustodialSites`, relying on only one very specific arg and setting a bunch of others may cause issues in the future.
And if you think the block and rss can be tweaked on the T0 side, then I think we could take those `ifs` out.

Using get() is necessary here because those keys will always be there, with the default value. That's why we need to check for None, empty string, etc.

hufnagel:
@amaltaro , how about this version ?

amaltaro:
I'm only reviewing it once I get my review on 7998 :-)

ticoann:
> Using get() is necessary here because those keys will always be there, with the default value. That's why we need to check for None, empty string, etc.

I don't think that is true for Tier0 cases. So if we check the key directly we can filter out Tier0 case.

https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L676
https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L1075

ticoann:
Never mind, Alan. You are right it fills up the kwargs with default values. Sorry for the noise.

amaltaro (approved):
Now I can review it :-)
Changes look good, I created another issue for the TaskChain unittest that failed and I'll fix that by tomorrow.

amaltaro:
@hufnagel can you please rebase it?

hufnagel:
Which commit broke it ? If I rebase I'll need to add that commit to the Tier0 deploy instructions.

amaltaro:
don't you want it for the wmagent branch (1.1.4_wmagent)? You could probably cherry-pick it for the wmagent branch and then rebase this master version.

hufnagel:
Any news on this ? @amaltaro are you going to create the wmagent branch version or will I need to do that ?

amaltaro:
Patch looks good and safe for the wmagent branch, though you still have to rebase it.
Feel free to cherry-pick it for the wmagent branch and make another PR.

hufnagel:
@amaltaro , what's going on with the failing unit test ?

ticoann:
The failure seems to be real failure I am running the test again to make sure.

```
WMCore_t.WMSpec_t.StdSpecs_t.StepChain_t.StepChainTests:testDupOutputDataSettings changed from success to failure
WMCore_t.WMSpec_t.StdSpecs_t.StepChain_t.StepChainTests:testOutputDataSettings changed from success to failure
WMCore_t.WMSpec_t.StdSpecs_t.TaskChain_t.TaskChainTests:testOutputDataSettings changed from success to failure
```

ticoann:
test this please

amaltaro:
Your PR is changing the default behaviour, so the unit tests have to be updated too. Anyways, I have that fixed on these 2 files:
http://amaltaro.web.cern.ch/amaltaro/forDirk/TaskChain_t.py
http://amaltaro.web.cern.ch/amaltaro/forDirk/StepChain_t.py

fetch them and make a new commit for these unittest fixes and you should be good.
BTW, I'm closing the other issue I had opened.

amaltaro:
You should have separated it in two commits, as usual. One for real changes (src/*), second for unit tests and pylint fixes. BTW, you're welcome :)

hufnagel:
Thanks for the changes ;-)

PR #8033, opened 2017-08-03
'basestring' fix applied

vilnius-dev:
basestring fix applied to WMCore/src/python folder, converts **basestring** to **str**

Example:

python2
-        if isinstance(status, basestring):
python3
+        if isinstance(status, str):

### DIFF
--- a/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
+++ b/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
@@ -12,7 +12,7 @@ if sys.version.startswith('3.'):
 
     # Make types compatible, can also be removed once transition is over
     long = int
-    basestring = str
+    str = str
 
 
 class _EmptyClass(object):
@@ -46,7 +46,7 @@ class JSONThunker(object):
                                  str,
                                  bytes,
                                  unicode,
-                                 basestring
+                                 str
                                  )
         # objects that inherit from dict should be treated as a dict
         #   they don't store their data in __dict__. There was enough
@@ -167,7 +167,7 @@ class JSONThunker(object):
         toThunk = self.checkRecursion(toThunk)
         toThunk = self.checkBlackListed(toThunk)
 
-        if isinstance(toThunk, basestring):
+        if isinstance(toThunk, str):
             # things that got blacklisted
             return toThunk
         if hasattr(toThunk, '__to_json__'):
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -297,7 +297,7 @@ class PSetTweak:
         for param, value in self:
             psetName = param.rsplit(".", 1)[0]
             paramName = param.rsplit(".", 1)[1]
-            if isinstance(value, basestring):
+            if isinstance(value, str):
                 value = "\"%s\"" % value
             result += "setattr(%s, \"%s\", %s)\n" % (
                 psetName, paramName, value)
--- a/src/python/Utils/IteratorTools.py
+++ b/src/python/Utils/IteratorTools.py
@@ -43,7 +43,7 @@ def convertFromUnicodeToStr(data):
     code fram
     http://stackoverflow.com/questions/1254454/fastest-way-to-convert-a-dicts-keys-values-from-unicode-to-str
     """
-    if isinstance(data, basestring):
+    if isinstance(data, str):
         return str(data)
     elif isinstance(data, collections.Mapping):
         return dict(map(convertFromUnicodeToStr, data.iteritems()))
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -13,7 +13,7 @@ def makeList(stringList):
     """
     if isinstance(stringList, list):
         return stringList
-    if isinstance(stringList, basestring):
+    if isinstance(stringList, str):
         toks = stringList.lstrip(' [').rstrip(' ]').split(',')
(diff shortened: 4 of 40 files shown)


ericvaandering:
I think this is going to blow up, unfortunately. In python2 basestr also includes unicode while str does not.

vkuznet:
If you'll add my name explicitly I'll comment earlier. Eric is right, this PR will break the code since basestring covers both str and unicode. Therefore you (Roman) need to adjust code accordingly and change every statement to cover both use cases, e.g.
```
if (isinstance(status, str) or isinstance(status, unicode))
```
Please be aware that both conditions should be applied through OR operator and every if or if not should cover them accordingly.

ericvaandering:
Can be a bit simpler. Is instance takes a tuple as the 2nd parameter

On Aug 7, 2017, at 10:53, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:


If you'll add my name explicitly I'll comment earlier. Eric is right, this PR will break the code since basestring covers both str and unicode. Therefore you (Roman) need to adjust code accordingly and change every statement to cover both use cases, e.g.

if (isinstance(status, str) or isinstance(status, unicode))


Please be aware that both conditions should be applied through OR operator and every if or if not should cover them accordingly.

amaltaro:
Sorry for creating confusion, but I though all strings were unicode in python3 (all under the str class/data type), thus there is no unicode type. Am I wrong?

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
```diff
@@ -12,7 +12,7 @@
 
     # Make types compatible, can also be removed once transition is over
     long = int
-    basestring = str
+    str = str
```
vkuznet: this is ambiguous and should not be defined

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
```diff
@@ -46,7 +46,7 @@ def __init__(self):
                                  str,
                                  bytes,
                                  unicode,
```
vkuznet: unicode is not defined in python3 therefore should be removed manually. At the same time bytes is not strings anymore. Therefore I would review this class manually.

vkuznet (commented):
My fault, in python3 unicode and str are the same, so we need to test another way around, i.e.
if isinstance(s, unicode) should be converted to if isinstance(s, str). I also added couple of comments to code which should be revisited manually. I would suggest to write simple stand-alone code and test its logic in python2 and python3.

ericvaandering:
That's why I said initially that this fix only makes sense also when doing another one, which I think is "builtins". That adds a "str" class to python2 that is supposed to work like str in python3.  That mostly works, I think, but causes some problems

But look, we have to temper our expectations. We are not going to have code that really works on python2 and python3, I think. But what we can work towards is code that mostly does that and when we can really test things on python3, we can do an automated conversion of the last little bit and maintain both code bases that way. At least that is my thought. I don't expect we need to maintain both code bases for long when we finally decide to go for it.

vkuznet:
Eric,
we can do slightly differently. Replace
```if instance(s, basestring)```
into
```
from six import string_types
if instance(s, string_types)
```
which defines string_types differently in py2/py3.
In py2 string_types is a ```(<type 'basestring'>,)``` while in py3 it is ```(<class 'str'>,)```.
The problem is that in py2 we still need to cover unicode.

**Comment by ericvaandering** [comment]
No. No six. I'm open to version defined if statements if that is the last option.

On Aug 8, 2017, at 04:02, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:

Eric,
we can do slightly differently. Replace
```if instance(s, basestring)```
into
```
from six import string_types
if instance(s, string_types)
```
which defines string_types differently in py2/py3.
In py2 string_types is a ```(<type 'basestring'>,)``` while in py3 it is ```(<class 'str'>,)```.
The problem is that in py2 we still need to cover unicode.

ericvaandering:
But moreover, I'm going to stick to my original recommendation. The str/basestring/unicode mess is probably the hardest thing to do during this transition, so I think it's best to forget about or postpone this until the rest is in good shape. 

I sent @amaltaro and @vkuznet some ideas about the lowest hanging fruit. I'd suggest starting there (or maybe @vkuznet has other ideas). But I don't think this is the place to start with our code.

amaltaro:
I couldn't agree more. Honestly, if the py2 to py3 migration is straight forward (as it would be for this feature) AND the futurize procedure is not (or depends on third -party libraries), I'm all in favor of keep a bookkeeping of those changes (maybe a GH issue flagged as python3, with the exact change and/or command needed for the migration) and forgot it for now until we get to the point where we're about to migrate to python3.

Otherwise we may increase the dependencies, code performance, etc for a couple of years "just" for the sake of having a hybrid code that runs on py2 and py3. Just my 2 cents :)

mapellidario:
> The str/basestring/unicode mess is probably the hardest thing to do during this transition, so I think it's best to forget about or postpone this until the rest is in good shape.

I could not agree more! However, we decided not to split the work between easy (first) and hard (later), but with the gradual migration we deal with all changes (easy and hard) together only on a portion of dmwm/WMCore at a time.
In particular. dealing with `basestring` is part of our current efforts with string in py2 and py3 and it is taken care of. 

I would suggest to close this PR without merging

amaltaro:
Agreed! Thanks Roman and Dario for your effort! Closing it

PR #8037, opened 2017-08-04
handle wq job estimation failure

### LINKED ISSUE #8035: Workflow not moving from "assigned"
Something weird is happening, I created this ACDC two days ago
https://cmsweb.cern.ch/reqmgr2/fetch?rid=vlimant_ACDC1_task_HGC-PhaseIITDRSpring17DR-00023__v1_T_170802_143815_4017

it is sitting in "assigned", but, when I go through WMStats it says the workflow has already run some jobs.
The only WorkQueue element seems to be done
https://cmsweb.cern.ch/couchdb/workqueue/_design/WorkQueue/_rewrite/elementsInfo?request=vlimant_ACDC1_task_HGC-PhaseIITDRSpring17DR-00023__v1_T_170802_143815_4017

Could you please help us to take a look?

ticoann:
fixes #8035

### DIFF
--- a/src/couchapps/WorkQueue/views/jobStatusByRequest/map.js
+++ b/src/couchapps/WorkQueue/views/jobStatusByRequest/map.js
@@ -1,6 +1,6 @@
 function(doc) {
    var ele = doc["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
-   if (ele && ele['Jobs']) {
+   if (ele && ele['Jobs'] !== undefined) {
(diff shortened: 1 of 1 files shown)


ticoann:
Although all the WQ element has to have more than 0 jobs. There is a case job count update fails. (I guess some network problem)  Since there is no retry mechanism and it is handled by LQ even though estimate is off, we should handle this in state transition as well. (I think)

amaltaro (approved):
Looks good to me.

PR #8042, opened 2017-08-09
'long' fix applied

vilnius-dev:
Strips the `L` prefix on long literals and renames `long` to `int`.

### DIFF
--- a/src/python/WMCore/Configuration.py
+++ b/src/python/WMCore/Configuration.py
@@ -22,7 +22,6 @@ _SimpleTypes = [
     float,
     basestring,  # For py2/py3 compatibility, don't let futurize remove PY3 Remove when python 3 transition complete
     str,
-    long,  # PY3: Not needed in python3, will be converted to duplicate int
     type(None),
     int,
 ]
--- a/src/python/WMCore/ReqMgr/Web/utils.py
+++ b/src/python/WMCore/ReqMgr/Web/utils.py
@@ -37,8 +37,7 @@ def quote(data):
         res = data
     elif  isinstance(data, list):
         res = data
-    elif  isinstance(data, long) or isinstance(data, int) or\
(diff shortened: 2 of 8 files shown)


amaltaro:
The only way to have it working would be changing the SQL driver/class such that it **always** return integers. At the moment we always have `long` coming back from the DAOs.

Besides that, this patch has duplicate `int` types.

amaltaro:
Looking at the oracle documentation, could it be that we see `long` in the responses because the tables are created either with LONGINTEGER or NUMBER
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/DBS3Buffer/Oracle/Create.py#L111
?

Performing a quick test on the bossair tables (bl_runjob and bl_status), it seems there is no long data type returned when queried.

ericvaandering:
@vilnius-dev please remember to update the code in test/python as well for all of these patches.

ericvaandering:
@amaltaro I can imagine that SQLAlchemy is down sizing things to put it in the int type if it fits? 

But I thought we already did this as part of stage1. I'll have to look when I get into the office.

amaltaro:
> @amaltaro I can imagine that SQLAlchemy is down sizing things to put it in the int type if it fits? 

I'm not sure. I remember seeing run numbers like 1L, 2L, etc a month ago when I was fixing a few issues related to LumiList.

vilnius-dev:
Duplicate int types fixed

mapellidario:
This is taken care by the fixer `lib2to3.fixes.fix_long`, which is in the futurize stage 2 list.
In #9731 the file `src/python/Utils/MathUtils.py` contains such changes.

I would therefore suggest to close this PR without merging.

amaltaro:
Agreed! Thanks Roman and Dario for your effort! Closing it

PR #8043, opened 2017-08-09
'filter' fix applied

vilnius-dev:
Wraps filter() usage in a list call.
Second commit applied to "test" only and first applied to the rest files

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -88,7 +88,7 @@ class PSetLister:
         self.psets.append(psetPath)
         params = childParameters(psetPath, pset)
         self.parameters[psetPath] = params
-        map(self, childPSets(pset))
+        list(map(self, childPSets(pset)))
         self.queue.pop(-1)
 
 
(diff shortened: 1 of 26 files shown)


vilnius-dev:
@vkuznet

amaltaro:
Is it because filter and map return a generator in py3?

ericvaandering:
Filter and lambda functions in general are falling out of favor as well. But yeah, I bet filter itself is changing. @vilnius-dev are you also documenting these changes with an explanation in a Twiki?

vilnius-dev:
Usually I document all changes on Friday.

ericvaandering:
@ticoann this is a good candidate for testing and integration early in the process.

vkuznet:
@vilnius-dev , where are the documented changes on (T)wiki. I don't see them. Please make them available asap.

ericvaandering:
@ticoann this is good to go as far as I'm concerned. I guess I need to add this fixer to the python 3 checker now.

ticoann:
That is good idea. :-) Thanks.

PR #8045, opened 2017-08-10
Fix isDrainMode arguments in JobArchiver

### LINKED ISSUE #8044: request doesn't move to completed status
I think the problem is following query returns empty list.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMBS/MySQL/Subscriptions/MarkNewFinishedSubscriptions.py

I am not sure what caused this. I will investigate further.

amaltaro:
Fixes #8044

isDrainMode was missing an argument.


ticoann (approved):

ticoann:
Thanks Alan

Issue #8039, opened 2017-08-07
workload summary not created (or lacking info) in > 1.1.6.pre3


amaltaro:
Just started investigating what is happening, but I've seen two cases so far:
* workload summary is created but it's missing tons of information, including the output datasets and stats
* workload summary isn't created at all "Error: unknown workflow name"

amaltaro:
Sigh, I must have done something bad with all the thousands of rebase I had to do on PR 7998.
Problem is that only the unmerged filesets are getting created (and top level).

Unless tag pre2 is fine, then it means another PR screwed that up.

amaltaro:
My tests are still running, but everything seems to be fine when running on private central services and new agent.

There is still a problem though, that stepchain fix is not backwards compatible (there were a few fixes to the spec files), the agent then counts on some of those information in order to properly create the filesets. As we can see, there is a problem running "old" workflows (created on reqmgr 1.1.4 or 1.1.5) on the new agent. We have to discuss on how to proceed in the next upgrade.

I still have to test what happens when we run new requests (new reqmgr) on old agent (1.1.4), I believe it will work without any issues.

amaltaro:
That's it. Master works fine but it doesn't if we use the current production reqmgr (1.1.5.pre5).
I'm closing this issue now. We still have to discuss on how to proceed with these breaking changes during production deployment.

ticoann:
> I still have to test what happens when we run new requests (new reqmgr) on old agent (1.1.4), I believe it will work without any issues.

If that is the case we can  update cmsweb first and and deploy new agents after that.

PR #8046, opened 2017-08-10
Minor aesthetic fixes for TaskArchiver

amaltaro:
Since I was debugging #8039 and had these improvements in a branch already, why not to push them :-)
Mostly logging improvements, but I also deleted the FileEncoder class that was giving a pylint error. I couldn't find it being used anywhere in the code.

### DIFF
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -9,8 +9,8 @@ import shutil
 import httplib
 import urllib2
 import threading
-import traceback
 import time
+import urllib2
 
 from WMCore.Lexicon import sanitizeURL
 from WMCore.Algorithms import MathAlgos
@@ -30,22 +30,6 @@ from WMCore.WMBS.Workflow import Workflow
 from WMComponent.JobCreator.CreateWorkArea import getMasterName
 from WMComponent.JobCreator.JobCreatorPoller import retrieveWMSpec
 from WMComponent.TaskArchiver.DataCache import DataCache
-
-
-class FileEncoder(json.JSONEncoder):
-    """
-    JSON encoder to transform sets to lists and handle any object with a __to_json__ method
-    """
-
-    def default(self, obj):
-        """
-        Redefine JSONEncoder default method
-        """
-        if hasattr(obj, '__to_json__'):
-            return obj.__to_json__()
-        elif isinstance(obj, set):
-            return list(obj)
-        return json.JSONEncoder.default(self, obj)
 
 
 class CleanCouchPollerException(WMException):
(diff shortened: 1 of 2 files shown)


src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -928,19 +908,11 @@ def publishRecoPerfToDashBoard(self, workload):
...
+
         # Request will be only interesting for performance if it's a ReReco or PromptReco
-        (isReReco, isPromptReco) = (False, False)
-        if workload.getRequestType() == 'ReReco':
-            isReReco = True
-        # Yes, few people like magic strings, but have a look at :
-        # https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L718
-        # Might be safe enough
-        # FIXME: in TaskArchiver, add a test to make sure that the dataset makes sense (procDataset ~= /a/ERA-PromptReco-vVERSON/DQM)
-        if re.search('PromptReco_', workload.name()):
```
amaltaro: @hufnagel do you still want these performance stats for PromptReco?

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -226,33 +222,28 @@ def completeTasks(self, finishedwfs):
...
                     if self.workQueue != None:
                         subList = []
                         logging.info("Marking subscriptions as Done ...")
                         for l in finishedwfs[workflow]["workflows"].values():
                             subList.extend(l)
                         self.notifyWorkQueue(subList)
 
-                    #Now we know the workflow as a whole is gone, we can delete the information from couch
+                    # Tier-0 case, the agent has to mark it completed
                     if not self.useReqMgrForCompletionCheck:
```
amaltaro: and this code as well. It's only used by T0, right? since there is no aborted-completed status for T0 requests, I removed this logic as well

ticoann:
@hufnagel, Dirk could you check on Alan's comment?

ticoann (approved):

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
hufnagel: I don't think we ever used them. They were supposed to be used to determine future job splitting.

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
hufnagel: Correct, Tier0 doesn't ever abort a workflow. We always finish them (even if they have 100% failure).

hufnagel (approved):

PR #8047, opened 2017-08-10
Fix raw_input

vilnius-dev:
The input() function was fixed in Python 3 so that it always stores the user inputs as str objects. In order to avoid the dangerous behavior in Python 2 to read in other types than strings, we have to use raw_input() instead. 
@vkuznet

### DIFF
--- a/bin/wmstats-request-status-change.py
+++ b/bin/wmstats-request-status-change.py
@@ -39,7 +39,7 @@ if __name__ == "__main__":
         print("new status needs to be set")
         sys.exit(1)
 
-    answer = raw_input("%s change to %s in wmstats db (yes, no)?" % (options.request, options.newstatus))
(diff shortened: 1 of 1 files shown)


vkuznet (commented):
I don't expect any issues with that but it would be nice to confirm that changing long to int in current code will not break anything. The long used to read runinfo and I expect that client will not put actual long number since our run info values are far away from int/long range.

ericvaandering:
It looks like you made the raw_input branch as a subbranch of the one to fix long? If you just make that one change to the current master on a new branch (or re-base) I can't imagine there will be a problem.

vilnius-dev:
I rebased this branch, please have a look

vkuznet:
ok, looks good.

ericvaandering:
test this please

ericvaandering:
Good for me too. @ticoann please merge when you can

PR #8048, opened 2017-08-10
initial patch to get the log

### LINKED ISSUE #7224: Access to condor logs

ticoann:
fixes #7224 Initial patch, still add the cp code but added for the review

### DIFF
--- a/src/couchapps/WMStatsErl/views/jobsByStatusWorkflow/map.erl
+++ b/src/couchapps/WMStatsErl/views/jobsByStatusWorkflow/map.erl
@@ -25,7 +25,7 @@ fun({Doc}) ->
                OutAcc ++ [Type]
             end, [], StepVal)
           end, [], E)),
-          Emit([Workflow, Task, State, ExitCode, Site, AcdcUrl, AgentName, ListErrors], {[{id,Id},{rev,Rev}]})
+          Emit([Workflow, Task, State, ExitCode, Site, AgentName, AcdcUrl, ListErrors], {[{id,Id},{rev,Rev}]})
       end;
     _ -> ok
   end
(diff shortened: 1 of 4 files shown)


ticoann:
this will blow up unittest. I want to find out what it will be affected and I will fix those.

ticoann:
test this please

ticoann:
change to get all the log instead. replaced by #8456

PR #8049, opened 2017-08-10
Fix services logging for TagCollector and ReqMgrAuxDB

### LINKED ISSUE #8038: Cheerypy threads log rotation
From this HN discussion:
https://hypernews.cern.ch/HyperNews/CMS/get/webInterfaces/1555.html

the way we use and rotate the log files is not ideal. A thread keeps writing to the same file as of
when it was started, and the log rotation happens at midnight (I suppose) by appending the date to the log file, which is then also transferred to vocms022.

I guess we cannot use the logrotate for several of these WMCore services, since most of them spawn several threads, though I don't know exactly how that works :) Ideally our logs should follow the same behaviour as the mongdb service, for instance.

amaltaro:
Fixes #8038
This fix the problem reported by Lina where ReqMgr2 services are writing to two different log files:
* usual and correct /data/srv/logs/reqmgr2/*
* and forbidden directory /data/srv/state/reqmgr2/*

In addition to that, it also fixes a problem with the cherrypy log rotation that was using an incorrect logging name.

### DIFF
--- a/src/python/WMCore/REST/CherryPyPeriodicTask.py
+++ b/src/python/WMCore/REST/CherryPyPeriodicTask.py
@@ -35,7 +35,7 @@ class CherryPyPeriodicTask(object):
         if hasattr(config, "central_logdb_url"):
             # default set up for logDB config section need to contain propervalues
             from WMCore.Services.LogDB.LogDB import LogDB
-            self.logDB = LogDB(config.central_logdb_url, config.log_reporter,
+            self.logDB = LogDB(config.central_logdb_url, config.log_reporter, logger=self.logger,
                                thread_name=config.object.rsplit(".", 1)[-1])
         else:
             self.logDB = None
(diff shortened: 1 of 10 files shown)


src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -308,6 +308,7 @@ def __init__(self, app, config, mount):
 
         # initialize access to reqmgr2 APIs
         self.reqmgr_url = config.reqmgr.reqmgr2_url
+        # FIXME pass a logger here
```
amaltaro: Have to check it again, but I probably want to remove these comments...

amaltaro:
@ticoann tests look good and I see only caching and a couple of json files under state dir.

Feel free to review it. I still have to check those two FIXME to see if there is something else that needs to be improved, but it's likely to be the final version. (Oh, and I have to squash those commits)

amaltaro:
One problem that I didn't manage to solve is that the cherrypy thread logs go to both the thread log and the main logfile (reqmgr2). The amount of logging lines is quite small though.

ticoann (approved):

PR #8029, opened 2017-07-27
failed without retry for exit code 8006 and 70

ticoann:
fixes #8026

PR #8050, opened 2017-08-10
add exitCode for non retry list

ticoann:
replace #8029, still need to handle RelVal agent case.

### DIFF
--- a/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
+++ b/src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
@@ -38,6 +38,7 @@ from WMCore.JobStateMachine.ChangeState import ChangeState
 from WMCore.WMBS.Job import Job
 from WMCore.WMException import WMException
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
+from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
 
 
 class ErrorHandlerException(WMException):
@@ -85,6 +86,10 @@ class ErrorHandlerPoller(BaseWorkerThread):
 
         self.dataCollection = DataCollectionService(url=config.ACDC.couchurl,
                                                     database=config.ACDC.database)
+        if hasattr(self.config, "Tier0Feeder"):
+            self.reqAuxDB = None
+        else:
+            self.reqAuxDB = ReqMgrAux(self.config.TaskArchiver.ReqMgr2ServiceURL)
 
         return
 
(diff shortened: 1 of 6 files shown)


ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
```diff
@@ -233,8 +238,13 @@ def readFWJRForErrors(self, jobList):
...
                     continue
-
-                if len([x for x in report.getExitCodes() if x in self.exitCodes]):
+                
+                if self.reqAuxDB is None:
+                    exitCodesNoRetry = self.config.ErrorHandler.failureExitCodes
+                else:
+                    exitCodesNoRetry = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("NoRetryExitCodes", [])
+                
+                if len([x for x in report.getExitCodes() if x in exitCodesNoRetry]):
```
amaltaro: Apparently the cache duration for ReqMgrAuxDB is set to 0.

Since `self.exitCodes` is initialized in the init method (with the agent config), how about we do the following:
* in the `algorithm` method, we check whether `self.reqAuxDB` is not None and in that case we update `self.exitCodes` with the `NoRetryExitCodes` content. Inside a try/except block such that it doesn't break the component and - if it fails, then it uses the previous list of exit codes.

amaltaro (changes requested):
Once you update this code (if you agree with my suggestion), can you also squash the last 2 commits, please?

src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
ticoann: getWMAgentConfig is using MemoryCache class which mean it won't though the exception if call fails it will just return the previous result.

ticoann:
@amaltaro Alan, I squash the last 2 commits. Is it OK now?

src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
```diff
@@ -195,6 +200,12 @@ def readFWJRForErrors(self, jobList):
         cooloffJobs = []
         passJobs = []
         exhaustJobs = []
+
+        if self.reqAuxDB is None:
+            exitCodesNoRetry = self.config.ErrorHandler.failureExitCodes
```
amaltaro: Is T0 going to overwrite the default WMAgent configuration? Otherwise this line will crash since that config section won't have `failureExitCodes` attr. I'll make a new PR with my suggestions instead of writing them here :)

ticoann: Hi Alan, T0 was always overwriting that. 

https://github.com/dmwm/T0/blob/master/bin/tier0-mod-config#L95

But if T0 never used it I would prefer to make [] here. Although this is more flexible.

PR #8052, opened 2017-08-14
'itertools' fix applied

vilnius-dev:
In Python 2 the itertools module define variants of the global zip(), map(), and filter() functions that returned iterators instead of lists. In Python 3, those global functions return iterators, so those functions in the itertools module have been eliminated. @vkuznet

### DIFF
--- a/src/python/WMCore/DataStructs/LumiList.py
+++ b/src/python/WMCore/DataStructs/LumiList.py
@@ -120,7 +120,7 @@ class LumiList(object):
             if len(runs) <= 0 or len(lumis) != len(runs):
                 raise RuntimeError('Improper format for wmagentFormat. # of lumi lists must match # of runs')
 
-            for run, lumiString in itertools.izip(runs, lumis):
+            for run, lumiString in zip(runs, lumis):
                 runLumis = lumiString.split(',')
                 if not str(run) in self.compactList:
                     self.compactList[str(run)] = []
--- a/test/python/Utils_t/IteratorTools_t.py
+++ b/test/python/Utils_t/IteratorTools_t.py
@@ -24,7 +24,7 @@ class IteratorToolsTest(unittest.TestCase):
         listChunks = [i for i in grouper(list(range(0, 7)), 3)]  # Want list(range) for python 3
         iterChunks = [i for i in grouper(xrange(0, 7), 3)]  # xrange becomes range in python 3
 
-        for a, b in itertools.izip_longest(listChunks, iterChunks):
+        for a, b in itertools.zip_longest(listChunks, iterChunks):
             self.assertEqual(a, b)
 
         self.assertEqual(listChunks[-1], [6])


ericvaandering:
It seems to me that this fixer should be put in the list of things that we can automatically run to make a python3 code base, but not to apply now.

For one, as its done now, the LumiList one will not make an iterator python2 but a list. Even worse, "zip_longest" does not exist in itertoools on python2.

vkuznet:
@ericvaandering, Eric this is a nice example to discuss how we should handle DMWM code in general. Do we want to make it complaint with py2 and py3 or to keep it working for single version of python. I prefer the former, you can express your arguments. To resolve ambiguity of zip under py2/py3 I would suggest to make special DMWM module with such type of functions where we can clearly define behavior in py2/py3 and make them consistent. Then we can replace WMCore codebase with such cases to use our definition, e.g. dmwm.zip This will allow more smooth transition from py2/py3, and, in fact, will allow to perform side by side comparison of py2 and py3. Of course, we can put aside such changes but I worry that such growing list will slow down the transition and it would be harder to make py2/py3 comparison.

ericvaandering:
We should have an "in person" discussion, but here are my thoughts. We have a few different approaches:

1) Write separate code for python2 and python3 where we don't like the results coming from an auto-translator
2) Accept the results of the auto-translator even if they result in sub-optimal python (e.g. iterkeys() converted to keys() which performs less well)
3) Where we don't like the results and the conversion is straightforward, just make a note and hope that in the future we can easily make a python3 version in a completely automated way.

So my preference is as much of 3) as possible, a little bit of 1) and no 2). We're going to be running python2 for a couple of years still, I think, so I don't see the point in 2).

ericvaandering:
In any case with 1) we should make sure we can actually clean this up once we complete the transition. As I was discussing with Alan today, I don't foresee that we will maintain two code bases for longer than necessary (say a few months while we verify the python3 version). Once python3 is running and deployed, we don't have to keep looking back.

vkuznet:
I'm not insisting on #1, but I wonder how we can proof that py3 code runs as good as py2?
At least with #1 we can run both without branching and compare. And, I doubt that few months for
such large code base is feasible to maintain two branches. In my mind #1 gives option to
perform transition in logical steps, i.e. make compatible code, run py2 vs py3,
verify, make a switch, and then clean-up once completed. Let's discussed on
meeting/offline about transition steps.

ericvaandering:
My point with 3) is that I want that part to just be:

git checkout master
git checkout -b python3
futurize -f [list of deferred fixers]

In other words, a procedure you can apply to the tip of master anytime you want. For stuff that doesn't fit that mold, I think we need to do 3)

mapellidario:
Such changes are applied by the `lib2to3.fixes.fix_itertools` fixer, which futurize includes in the stage 2 list.
Right now, instead of applying a single stage 2 fixer on the whole dmwm/WMCore in a single PR, we are running all fixers (stage2 fixers included) one module at a time. 
For example, in #10057  the file `src/python/WMCore/Database/DBFormatter.py` contains such change.

I would therefore suggest to close this PR without merging.

amaltaro:
Agreed! Thanks Roman and Dario for your effort!

PR #8053, opened 2017-08-14
Update subscription methods to keep list of datasets under an output …

amaltaro:
wmagent branch version of #8021

Still testing, but new requests (once cmsweb gets upgraded) may carry a different subscription structure, so we also have to update the agents.

### DIFF
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -910,37 +910,39 @@ class WMTaskHelper(TreeHelper):
 
         Set the subscription information for this task's datasets
         The subscriptions information is structured as follows:
-        data.subscriptions.outputModules is a list of all output modules with configured datasets
-        data.subscriptions.<outputModule>.dataset
-        data.subscriptions.<outputModule>.custodialSites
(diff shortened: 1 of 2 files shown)


amaltaro:
Just tested it with the current reqmgr (1.1.5) and the newest one (1.1.6) and both works fine. So we can apply this patch at any time before the cmsweb upgrade.

ticoann:
@amaltaro, This need to be apply before cmsweb upgrade, right?

amaltaro:
Yes, I'll apply it in the coming days because it's backward compatible. If you want, feel free to merge it

ticoann:
Unless you are redeploying the agent. I will just leave it open until production agents are status so we can track. Thanks Alan,

amaltaro:
I've just pushed this patch in in all the agents. Feel free to merge it or just leave it open until we cut the next production release, by the next week.

amaltaro:
[Fix #8053 - Update subscription methods to keep list of datasets under an output](https://trello.com/c/Njabd2G0/53-fix-8053-update-subscription-methods-to-keep-list-of-datasets-under-an-output)

ticoann (approved):

PR #8055, opened 2017-08-14
add ReqMgrAux mock api

ticoann:
This will be applied in several components' unittests which uses the agent cofing from reqmgr_aux db.


ticoann:
@ericvaandering, Eric, please review. Only getWMAgentConfig will be used in component. put request is used in the deployment script.

ericvaandering (approved):
The general concept looks good. Mocking things and making sure they are called is tricky and I can't say I ever really understood it deeply, so if you haven't already, test this with a print statement in your mocked function just to be sure it's being called.

ticoann:
Wow, total blow up. I can't even copy and paste right. :-) updated patch.

PR #8063, opened 2017-08-17
'cmp' fix applied

vilnius-dev:
`past.builtins.cmp(x, y) → integer`

> Return negative if x<y, zero if x==y, positive if x>y.

[[source]](http://python-future.org/_modules/past/builtins/misc.html#cmp)

-----------------------------------------------------------------------------

> ### Python 2 only:
> `assert cmp('a', 'b') < 0 and cmp('b', 'a') > 0 and cmp('c', 'c') == 0`

> ### Python 2 and 3: alternative 1
> `from past.builtins import cmp`
> `assert cmp('a', 'b') < 0 and cmp('b', 'a') > 0 and cmp('c', 'c') == 0`

> ### Python 2 and 3: alternative 2
> `cmp = lambda(x, y): (x > y) - (x < y)`
> `assert cmp('a', 'b') < 0 and cmp('b', 'a') > 0 and cmp('c', 'c') == 0`

### DIFF
--- a/src/python/WMCore/DataStructs/Fileset.py
+++ b/src/python/WMCore/DataStructs/Fileset.py
@@ -6,6 +6,7 @@ Data object that contains a set of files
 
 """
 from __future__ import print_function
+from operator import itemgetter
 __all__ = []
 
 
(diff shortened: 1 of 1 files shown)


ericvaandering:
This is just changing the behavior of one line of code. Better to just rewrite that code in a better way.

ericvaandering:
I'd suggest following this recipe here and getting rid of cmp entirely

https://stackoverflow.com/questions/72899/how-do-i-sort-a-list-of-dictionaries-by-values-of-the-dictionary-in-python

vkuznet:
@vilnius-dev , any progress on this? Please take Eric's advise and re-write code to use specific key for comparison.

ericvaandering:
test this please

ericvaandering:
@ticoann should be good to merge. @vilnius-dev has removed the problematic code with a more pythonic way of doing this.

PR #8064, opened 2017-08-17
'execfile' fix applied

vilnius-dev:
Removes usage of execfile(). The argument to execfile() is wrapped in calls to open(), compile(), and exec().


ericvaandering:
@ticoann do we even need the file that this changes? Looks like it is from the ReqMgr1 -> 2 transition and can just be deleted.

ticoann:
You are right I will remove that code.

ticoann:
remove the package altogether #8086 - closing this.

PR #8066, opened 2017-08-17
'input' fix applied

vilnius-dev:
Converts `input(prompt)` to `eval(input(prompt))`

### DIFF
--- a/bin/wmstats-request-status-change.py
+++ b/bin/wmstats-request-status-change.py
@@ -39,7 +39,7 @@ if __name__ == "__main__":
         print("new status needs to be set")
         sys.exit(1)
 
-    answer = input("%s change to %s in wmstats db (yes, no)?" % (options.request, options.newstatus))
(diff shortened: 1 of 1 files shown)


vkuznet:
@vilnius-dev, this will fail in python2. You can make a simple test yourself, e.g.
```
a = eval(input("bla? "))
bla? yes
  File "<string>", line 1, in <module>
NameError: name 'yes' is not defined
```
or
```
a = eval(input("bla? "))
bla? "yes"
  File "<string>", line 1, in <module>
NameError: name 'yes' is not defined
```

Plus usage of eval function is a security flaw, at least you should use it as following:
```
eval(data, { "__builtins__": None }, {})
```
where you pass globals and locals.

Therefore I suggest to refactor this code explicitly for py2 and py3. The input can be used for py2 while eval can be used for py3. Or, leave this PR until we make a switch to py3.

ericvaandering:
This is re-changing the same code that was done in #8047 I think. 

The "dict" fix will do that too. Unfortunate.

ticoann:
This is not need we need to pass answer with quotation

PR #8070, opened 2017-08-17
'newstyle' fix applied

vilnius-dev:


ericvaandering:
Here too the import ordering is off in FWJRDBAPI.py

vkuznet:
@vilnius-dev , did you resolved issues with unit tests?

vilnius-dev:
It seems this fix changes results of Unit tests: https://cmssdt.cern.ch/jenkins/job/DMWM-WMCore-PR-test/5639/testReport/junit/WMCore_t.JobSplitting_t.Harvest_t/HarvestTest/testByRunAndRunBlacklist/

ericvaandering:
Let's get the import order fixed and run it again. That might be an unstable test and one we don't flag as such.

ericvaandering:
Hi @vilnius-dev from what I can tell, this is just an issue of your master not being correct and GH doesn't really know how to merge it. 

Our procedure for using git is shown here. https://github.com/dmwm/WMCore/wiki/Developing-against-WMCore It's the first four lines of "starting new development" that are where I think you went wrong. 

In short, always make sure your "master" is something that is or used to be DMWM's master. Never merge master into your development branches. Instead, use git rebase to update where your feature branch starts from.

@amaltaro should be able to help you next Wednesday if you can't get it yourself.

vilnius-dev:
I created a new PR here: https://github.com/dmwm/WMCore/pull/8199

vkuznet:
Roman,
are you sure that you need to load object from builtins?
I can do as simple as

```
Python 2.7.14 (default, Sep 22 2017, 00:06:07)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
>>> class Test(object):
...     def __init__(self):
...        pass
...
>>>
```

without `from builtins import object`.

ericvaandering:
But python3 has no "object" 

Code is here:

https://github.com/PythonCharmers/python-future/blob/master/src/future/builtins/__init__.py

and other things referenced therein

vkuznet:
Eric,
I'm not sure about object, here is an example:

```
 python3
Python 3.5.4 (default, Sep 22 2017, 08:33:07)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
>>> class Test(object):
...    def __init__(self):
...      pass
...
>>> print(Test)
<class '__main__.Test'>
>>> print(object)
<class 'object'>
>>> print(object.__dict__)
{'__hash__': <slot wrapper '__hash__' of 'object' objects>, '__subclasshook__': <method '__subclasshook__' of 'object' objects>, '__new__': <built-in method __new__ of type object at 0x10dc50100>, '__class__': <attribute '__class__' of 'object' objects>, '__format__': <method '__format__' of 'object' objects>, '__le__': <slot wrapper '__le__' of 'object' objects>, '__delattr__': <slot wrapper '__delattr__' of 'object' objects>, '__lt__': <slot wrapper '__lt__' of 'object' objects>, '__repr__': <slot wrapper '__repr__' of 'object' objects>, '__setattr__': <slot wrapper '__setattr__' of 'object' objects>, '__dir__': <method '__dir__' of 'object' objects>, '__sizeof__': <method '__sizeof__' of 'object' objects>, '__init__': <slot wrapper '__init__' of 'object' objects>, '__eq__': <slot wrapper '__eq__' of 'object' objects>, '__reduce__': <method '__reduce__' of 'object' objects>, '__reduce_ex__': <method '__reduce_ex__' of 'object' objects>, '__str__': <slot wrapper '__str__' of 'object' objects>, '__gt__': <slot wrapper '__gt__' of 'object' objects>, '__doc__': 'The most base type', '__getattribute__': <slot wrapper '__getattribute__' of 'object' objects>, '__ne__': <slot wrapper '__ne__' of 'object' objects>, '__ge__': <slot wrapper '__ge__' of 'object' objects>}
```

From this example I can conclude that `object` is known in python3 without any
additions and it is a class with buildin dict.

ericvaandering:
I wondered about that from looking at the code. Maybe all this is really doing is adding both python 2 & 3 methods to those objects (e.g. both next() and __next__() )

vilnius-dev:
What I have found regarding @vkuznet question:
https://stackoverflow.com/questions/36517648/in-python-why-import-object-from-builtins-module

**Python 2 only**
````python
class Upper(object):
    def __init__(self, iterable):
        self._iter = iter(iterable)
    def next(self):          # Py2-style
        return self._iter.next().upper()
    def __iter__(self):
        return self

itr = Upper('hello')
assert itr.next() == 'H'     # Py2-style
assert list(itr) == list('ELLO')
````

**Python 2 and 3:**
````python
from builtins import object

class Upper(object):
    def __init__(self, iterable):
        self._iter = iter(iterable)
    def __next__(self):      # Py3-style iterator interface
        return next(self._iter).upper()  # builtin next() function calls
    def __iter__(self):
        return self

itr = Upper('hello')
assert next(itr) == 'H'      # compatible style
assert list(itr) == list('ELLO')
````

vkuznet:
Well, if object from buildtins provides next/iter methods it does not mean
that we need to convert all our classes to be generator like object. In fact,
it is another way around, I would expect that almost all classes are not
generator objects and only for few exceptions we need this functionality.

Therefore, it is unclear to me that we do need to have
from builtins import object
in all modified code.

ericvaandering:
Do you think it hurts?

vkuznet:
It does not hurt but it will add functionality to an object which may not need it or it may be in-appropriate. In other words it changes object behavior, now it is not iterable while with new builtins it will be. Is it correct (and desired) behavior is an open question.

PR #8071, opened 2017-08-17
'object' fix applied

vilnius-dev:
Fixer that adds ``from builtins import object`` if there is a line like this:
    `class Foo(object):`

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -7,6 +7,7 @@ independent python structure
 
 """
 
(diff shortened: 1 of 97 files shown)


ericvaandering:
If you click on "details" you will see that a lot of the failures come from the wrong ordering of commits in MockDbsApi. Fix that and a lot will change, I think.

mapellidario:
This is covered by our gradual migration. I would suggest closing this PR without merging

amaltaro:
Thanks Roman and Dario for your effort! Closing it

PR #8073, opened 2017-08-17
'xrange_with_import' fix applied

vilnius-dev:
Renames `xrange()` to `range()` and wraps existing `range()`calls with list

### DIFF
--- a/bin/couch-thrash.py
+++ b/bin/couch-thrash.py
@@ -4,6 +4,7 @@ _couch-trash_
 
 """
 
+from builtins import range
 import os
 import random
 import time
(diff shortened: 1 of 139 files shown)


ericvaandering:
ReqMgrDocGenerator.py has imports in the wrong order

vkuznet:
@ticoann , this should be tested for performance impact since range and xrange are different data-types in py2. Changing to range (which is generator in py3) will impact py2 code since range is a list.

This may be an example to be applied later when we'll do py2/py3 conversion.

ticoann:
test this please

mapellidario:
This is covered by our gradual migration.

I would suggest closing this PR without merging, since it is superseded by our current efforts.

mapellidario:
Thanks Roman for your work! Closing this

PR #8076, opened 2017-08-18
Fix missing results function with json.dumps

vkuznet:


vkuznet:
Wrong move, I found that I need to return list instead of json. Please discard this PR.

PR #8077, opened 2017-08-18
adjust number of jobs per WQE by MaxLumi allowed

### LINKED ISSUE #8060: Adjust the size of the WQE for "monte carlo" request
looking at the failure for 

https://cmsweb.cern.ch/reqmgr2/fetch?rid=mcremone_task_TOP-RunIIWinter15wmLHE-00061__v1_T_170809_194110_9524

it looks like the size of the WQE should be smaller so as to not end up with a humengous amount of lumis in it.

ticoann:
fixes #8060

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
@@ -45,6 +45,16 @@ class MonteCarlo(StartPolicyInterface):
         mask = Mask(**self.mask)
 
         #First let's initialize some parameters
+        lumis_per_job = ceil(self.args['SliceSize'] / self.args['SubSliceSize'])
+        totalLumisPerElement = int(self.args['MaxJobsPerElement']) * lumis_per_job
+
+        if self.args['MaxLumisPerElement'] and totalLumisPerElement > int(self.args['MaxLumisPerElement']):
+            # If there are too many lumis in the WQ element. reduce the number of jobs per element
+            self.args['MaxJobsPerElement'] = floor(int(self.args['MaxLumisPerElement']) / lumis_per_job)
+            if self.args['MaxJobsPerElement'] == 0:
+                raise WorkQueueWMSpecError(self.wmspec, """Too many lumis (%s) in a job:
+                                           Change 'SliceSize' / 'SubSliceSize'""" % lumis_per_job)
+
         stepSize = int(self.args['SliceSize']) * int(self.args['MaxJobsPerElement'])
         total = mask['LastEvent'] - mask['FirstEvent'] + 1
         lastAllowedEvent = mask['LastEvent']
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, Alan please review

src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
```diff
@@ -45,6 +45,15 @@ def split(self):
         mask = Mask(**self.mask)
 
         #First let's initialize some parameters
+        lumis_per_job = ceil(self.args['SliceSize'] / self.args['SubSliceSize'])
+        totalLumisPerElement = int(self.args['MaxJobsPerElement']) * lumis_per_job
+
+        if self.args['MaxLumisPerElement'] and totalLumisPerElement > int(self.args['MaxLumisPerElement']):
+            # If there are too many lumis in the WQ element. reduce the number of jobs per element
+            self.args['MaxLumisPerElement'] = floor(totalLumisPerElement / lumis_per_job)
```
amaltaro: I knew something wasn't right here. After spending some pen+paper time, I think we have to actually replace this L53 by
`self.args['MaxJobsPerElement'] = floor(int(self.args['MaxLumisPerElement']) / lumis_per_job)`

this will give us a smaller number of jobs per element, which is then used to calculate `stepSize` below.

In addition to that, I can't see any use case for this `if` below.

amaltaro (changes requested):
This flexibility is much better than what I proposed, thanks!

Since I almost missed a problem in this patch, can you please write a unit test to test this case?

Also x-check whether my comment makes sense, from a few examples I wrote down, this code would actually make things worse.

src/python/WMCore/WorkQueue/Policy/Start/MonteCarlo.py
ticoann: Alan, you are completely right! That was I was intend to do. :-)
If statement below is needed, Of course I made another typo. It should be below.
```
if self.args['MaxJobsPerElement'] == 0:

```
I will update the patch and add the unittest. Thanks.

ticoann:
@amaltaro, Alan could you review again?

ticoann:
test this please

amaltaro (approved):

amaltaro:
Looks better now :) I still can't see a case where 'MaxJobsPerElement' would be 0, but it won't hurt if we keep it.

ticoann:
Yes that is for the safety guard. Hopefully, it will never happen but people make mistake.

test/python/WMCore_t/WorkQueue_t/Policy_t/Start_t/MonteCarlo_t.py
```diff
@@ -255,6 +255,51 @@ def testContinuousSplittingSupport(self):
         self.assertRaises(NotImplementedError, policyInstance.modifyPolicyForWorkAddition, *[None])
         return
 
+    def _getLHEProdictionWorkload(self, splitArgs):
```
amaltaro: typo

test/python/WMCore_t/WorkQueue_t/Policy_t/Start_t/MonteCarlo_t.py
```diff
@@ -255,6 +255,51 @@ def testContinuousSplittingSupport(self):
...
+
+        LHEProductionWorkload = monteCarloWorkload('MonteCarloWorkload', mcArgs)
+        LHEProductionWorkload.setJobSplittingParameters(
+            getFirstTask(LHEProductionWorkload).getPathName(),
+            'EventBased',
+            {'events_per_job': splitArgs['SliceSize'],
+             'events_per_lumi': splitArgs['SubSliceSize']})
+        getFirstTask(LHEProductionWorkload).setSiteWhitelist(['T2_XX_SiteA', 'T2_XX_SiteB'])
+        getFirstTask(LHEProductionWorkload).addProduction(totalEvents=totalevents)
+        getFirstTask(LHEProductionWorkload).setSiteWhitelist(['T2_XX_SiteA', 'T2_XX_SiteB'])
```
amaltaro: and duplicated line

amaltaro (commented):
Sorry for the belated review

ticoann:
Thanks Alan, created the patch

PR #8080, opened 2017-08-21
Run job splitter only on a reasonable amount of events/lumis

### LINKED ISSUE #7988: JobSplitting not matching the 8h target
Looking at 

https://cms-gwmsmon.cern.ch/prodview/pdmvserv_task_TOP-RunIISummer15wmLHEGS-00237__v1_T_170701_124014_5751
https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_task_TOP-RunIISummer15wmLHEGS-00237__v1_T_170701_124014_5751

the miniaodsim jobs do not seem to be coming out as expected.

{
    "splitAlgo": "EventAwareLumiBased", 
    "splitParams": {
      "trustPUSitelists": true, 
      "algorithm": "EventAwareLumiBased", 
      "max_events_per_lumi": 20000, 
      "deterministicPileup": false, 
      "lheInputFiles": false, 
      "trustSitelists": false, 
      "events_per_job": 28800, 
      "type": "Processing"
    }, 
    "taskName": "/pdmvserv_task_TOP-RunIISummer15wmLHEGS-00237__v1_T_170701_124014_5751/TOP-RunIISummer15wmLHEGS-00237_0/TOP-RunIISummer16DR80Premix-00266_0/TOP-RunIISummer16DR80Premix-00266_1/TOP-RunIISummer16MiniAODv2-00263_0", 
    "taskType": "Processing"
  }, 

is the splitting for the task

"Task4": {
    "KeepOutput": true, 
    "PrepID": "TOP-RunIISummer16MiniAODv2-00263", 
    "InputFromOutputModule": "AODSIMoutput", 
    "GlobalTag": "80X_mcRun2_asymptotic_2016_TrancheIV_v6", 
    "TimePerEvent": 1, 
    "FilterEfficiency": 1, 
    "SplittingAlgo": "EventAwareLumiBased", 
    "ProcessingString": "PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6", 
    "ScramArch": [
      "slc6_amd64_gcc530"
    ], 
    "SizePerEvent": 60, 
    "ConfigCacheID": "86329c92d317a6deffea363c322483d1", 
    "Memory": 5900, 
    "Multicore": 4, 
    "TaskName": "TOP-RunIISummer16MiniAODv2-00263_0", 
    "AcquisitionEra": "RunIISummer16MiniAODv2", 
    "PrimaryDataset": "TGJets_lept_4f_ckm_NLO_13TeV_amcatnlo_madspin_Herwigpp", 
    "Campaign": "RunIISummer16MiniAODv2", 
    "InputTask": "TOP-RunIISummer16DR80Premix-00266_1", 
    "CMSSWVersion": "CMSSW_8_0_21"
  },

which means 28800/60.=480 min of walltime desired.

This could be coming from the fact that jobs are made as soon as there are a couple or one file available in input, and therefore not accumulate enough number of input files to make a long enough job.

amaltaro:
Fixes #7988 

Changes how secondary tasks/steps create processing jobs. Instead of creating jobs as soon as files are available `wmbs_sub_files_available`, wait for enough files such that they can meet the job splitting arguments (amount of events or lumis).

In addition to that, check whether the (input) fileset is closed, in this case just proceed with the splitting no matter how many files/events/lumis are available (useful for the very last jobs to be created).

Also fixed a couple of issues with the EventAwareByWork splitting.

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -30,6 +30,12 @@ class EventAwareLumiBased(JobFactory):
 
     locations = []
 
+    def __init__(self, package='WMCore.DataStructs', subscription=None, generators=None, limit=0):
+        super(EventAwareLumiBased, self).__init__(package, subscription, generators, limit)
+
+        self.loadRunLumi = None  # Placeholder for DAO factory if needed
+        self.collectionName = None  # Placeholder for ACDC Collection Name, if needed
+
     def algorithm(self, *args, **kwargs):
         """
         _algorithm_
@@ -43,58 +49,59 @@ class EventAwareLumiBased(JobFactory):
         eventLimit = int(kwargs.get('max_events_per_lumi', 20000))
         totalEvents = int(kwargs.get('total_events', 0))
         splitOnFile = bool(kwargs.get('halt_job_on_file_boundaries', False))
-        collectionName = kwargs.get('collectionName', None)
+        self.collectionName = kwargs.get('collectionName', None)
         splitOnRun = kwargs.get('splitOnRun', True)
         getParents = kwargs.get('include_parents', False)
         runWhitelist = kwargs.get('runWhitelist', [])
         runs = kwargs.get('runs', None)
         lumis = kwargs.get('lumis', None)
         applyLumiCorrection = bool(kwargs.get('applyLumiCorrection', False))
+        deterministicPileup = kwargs.get('deterministicPileup', False)
 
         timePerEvent, sizePerEvent, memoryRequirement = \
             self.getPerformanceParameters(kwargs.get('performance', {}))
-        deterministicPileup = kwargs.get('deterministicPileup', False)
+
         eventsPerLumiInDataset = 0
 
-        if deterministicPileup and self.package == 'WMCore.WMBS':
-            getJobNumber = self.daoFactory(classname="Jobs.GetNumberOfJobsPerWorkflow")
-            jobNumber = getJobNumber.execute(workflow=self.subscription.getWorkflow().id)
-            self.nJobs = jobNumber
+        if self.package == 'WMCore.WMBS':
+            self.loadRunLumi = self.daoFactory(classname="Files.GetBulkRunLumi")
+            if deterministicPileup:
+                getJobNumber = self.daoFactory(classname="Jobs.GetNumberOfJobsPerWorkflow")
+                self.nJobs = getJobNumber.execute(workflow=self.subscription.getWorkflow().id)
 
         goodRunList = {}
         if runs and lumis:
             goodRunList = buildLumiMask(runs, lumis)
 
         # If we have runLumi info, we need to load it from couch
-        if collectionName:
+        if self.collectionName:
             try:
                 from WMCore.ACDC.DataCollectionService import DataCollectionService
                 couchURL = kwargs.get('couchURL')
                 couchDB = kwargs.get('couchDB')
                 filesetName = kwargs.get('filesetName')
-                collectionName = kwargs.get('collectionName')
 
                 logging.info('Creating jobs for ACDC fileset %s', filesetName)
                 dcs = DataCollectionService(couchURL, couchDB)
-                goodRunList = dcs.getLumiWhitelist(collectionName, filesetName)
+                goodRunList = dcs.getLumiWhitelist(self.collectionName, filesetName)
             except Exception as ex:
                 msg = "Exception while trying to load goodRunList. "
                 msg += "Refusing to create any jobs.\nDetails: %s" % str(ex)
                 logging.exception(msg)
                 return
 
-        lDict = self.sortByLocation()
+        lDict = self.getFilesSortedByLocation(avgEventsPerJob)
+        if not lDict:
+            logging.info("There are not enough events/files to be splitted. Trying again next cycle")
+            return
+
+
         locationDict = {}
-
-        # First we need to load the data
-        if self.package == 'WMCore.WMBS':
-            loadRunLumi = self.daoFactory(classname="Files.GetBulkRunLumi")
-
         for key in lDict.keys():
             newlist = []
             # First we need to load the data
-            if self.package == 'WMCore.WMBS':
-                fileLumis = loadRunLumi.execute(files=lDict[key])
+            if self.loadRunLumi:
+                fileLumis = self.loadRunLumi.execute(files=lDict[key])
                 for f in lDict[key]:
                     lumiDict = fileLumis.get(f['id'], {})
                     for run in lumiDict.keys():
@@ -218,7 +225,7 @@ class EventAwareLumiBased(JobFactory):
                             firstLumi = None
                             lastLumi = None
 
-                        if firstLumi == None:
+                        if firstLumi is None:
                             # Set the first lumi in the run
                             firstLumi = lumi
 
--- a/src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
@@ -7,7 +7,7 @@ a set of jobs based on lumi sections, failing jobs with too many
 events in a lumi and adapting the number of lumis per job
 according to the average number of events per lumi in the files.
 
-This is adapted from EventAwareByLumi but does its work on lumi sections directly, not files.
+This is adapted from EventAwareLumiBased but does its work on lumi sections directly, not files.
 No correction for lumis split across files is needed, it's automatic. Makes use of the LumiList
 class to simplify the code
 
@@ -18,7 +18,6 @@ from __future__ import (division, print_function)
 import logging
 from collections import defaultdict
 
-from WMCore.ACDC.DataCollectionService import DataCollectionService
 from WMCore.DataStructs.LumiList import LumiList
 from WMCore.DataStructs.Run import Run
 from WMCore.JobSplitting.JobFactory import JobFactory
@@ -36,6 +35,7 @@ class EventAwareLumiByWork(JobFactory):
         super(EventAwareLumiByWork, self).__init__(package, subscription, generators, limit)
 
         self.loadRunLumi = None  # Placeholder for DAO factory if needed
+        self.collectionName = None  # Placeholder for ACDC Collection Name, if needed
         self.perfParameters = {}
         self.deterministicPU = False
         self.maxLumis = 1  # Maximum lumis seen in a job (needed for deterministic pileup only)
@@ -67,7 +67,7 @@ class EventAwareLumiByWork(JobFactory):
         lumiEventLimit = int(kwargs.get('max_events_per_lumi', 20000))
         totalEventLimit = int(kwargs.get('total_events', 0))
         splitOnFile = bool(kwargs.get('halt_job_on_file_boundaries', False))
-        collectionName = kwargs.get('collectionName', None)
+        self.collectionName = kwargs.get('collectionName', None)
         splitOnRun = kwargs.get('splitOnRun', True)
         getParents = kwargs.get('include_parents', False)
         runWhitelist = kwargs.get('runWhitelist', [])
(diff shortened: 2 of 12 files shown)


amaltaro:
Another comment, I've just noticed this is not optimal because it guarantees we will have at least 1 job with a decent splitting, but it could be that for each decent job there will be one very short...

If this change is not enough, then we have to make (much?) more changes to the job splitting algos such that they consider only the lumi sections that are able to completely fill a job, leaving the remainder lumi sections for a future cycle. Still, a whole file has to either be fully splitted or not...

amaltaro:
Before I start the real tests, I'd like to hear your thoughts @ericvaandering @ticoann @hufnagel

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -309,3 +312,32 @@ def algorithm(self, *args, **kwargs):
...
+        If the fileset is closed, resume the splitting. Otherwise check whether
+        there are enough events in each of these locations, if events don't
+        match the desired events_per_job splitting parameter, then skip those
+        files until further cycles.
+        :param eventsPerJob: number of events desired in the splitting
+        :return: a dictionary of files, key'ed by a frozenset location
+        """
+        lDict = self.sortByLocation()
+
+        fileset = self.subscription.getFileset()
```
amaltaro: I guess CRAB relies on DataStruct, so I probably have to check whether it's using WMCore.WMBS first (?)

amaltaro:
test this please

amaltaro:
That's unfortunate... I've just found we have JobSplitting tests under
* WMCore/JobSplitting, for DataStructs
* WMCore/WMBS/JobSplitting, and for WMBS structs

working on that...

amaltaro:
Tests are looking good and it improves the overall job splitting in 50%, at the very very least.
Reviews are most welcomed at this stage.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
```diff
@@ -325,3 +328,29 @@ def populateFilesFromWMBS(self, filesByLocation):
             lumiDict = fileLumis.get(f['id'], {})
             for run in lumiDict.keys():
                 f.addRun(run=Run(run, *lumiDict[run]))
+
+    def getFilesSortedByLocation(self, eventsPerJob):
```
ericvaandering: Is this function the same as the one in the other splitting code? If so, put it in the base class or import it from one to the other.

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -309,3 +316,29 @@ def algorithm(self, *args, **kwargs):
         self.lumiChecker.closeJob(self.currentJob)
         self.lumiChecker.fixInputFiles()
         return
+
+    def getFilesSortedByLocation(self, eventsPerJob):
+        """
+        _getFilesSortedByLocation_
+
+        Retrieves a list of files available and sort them by location.
+        If the fileset is closed, resume the splitting. Otherwise check whether
```
ericvaandering: What do you do if the file set is not closed but there are not enough events? Is there a timeout? Is success guaranteed at some point?

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
```diff
@@ -258,6 +259,8 @@ def lumiListFromACDC(couchURL=None, couchDB=None, filesetName=None, collectionNa
         This is not implemented yet
         :return:
         """
+        from WMCore.ACDC.DataCollectionService import DataCollectionService
```
ericvaandering: Can this go to the top?

src/python/WMCore/JobSplitting/LumiBased.py
```diff
@@ -367,3 +361,39 @@ def algorithm(self, *args, **kwargs):
         self.lumiChecker.closeJob(self.currentJob)
         self.lumiChecker.fixInputFiles()
         return
+
+    def getFilesSortedByLocation(self, lumisPerJob):
```
ericvaandering: Same comment.

test/python/WMCore_t/WMBS_t/JobSplitting_t/EventAwareLumiBased_t.py
```diff
@@ -235,6 +246,17 @@ def testB_NoFileSplitNoHardLimit(self):
                                splitOnRun=False,
                                events_per_job=360,
                                performance=self.performanceParams)
+        self.assertEqual(len(jobGroups), 0, "There isn't enough events, so it should have 0 job groups")
```
ericvaandering: aren't enough events. :-)

test/python/WMCore_t/WMBS_t/Subscription_t.py
```diff
@@ -1065,29 +1065,20 @@ def testListIncompleteDAO(self):
 
         incompleteSubs = subIncomplete.execute()
 
-        assert len(incompleteSubs) == 2, \
-            "ERROR: Wrong number of incomplete subscriptions returned: %s" % len(incompleteSubs)
-        assert testSubscription["id"] in incompleteSubs, \
-            "ERROR: Original subscription is missing."
-
+        self.assertEqual(len(incompleteSubs), 2, "ERROR: Wrong number of incomplete subscriptions returned: %s" % len(incompleteSubs))
```
ericvaandering: I never see the point in putting "ERROR:" at the front of these. The assertion error kind of shows there is an error....

test/python/WMCore_t/WMBS_t/Subscription_t.py
```diff
@@ -1065,29 +1065,20 @@ def testListIncompleteDAO(self):
...
+        self.assertTrue(testSubscription["id"] in incompleteSubs, "ERROR: Original subscription is missing.")
+        print("Incomplete subs: %s" % incompleteSubs)
```
ericvaandering: You're going to remove this, right?

ericvaandering (changes requested):
I have one substantive question and one question about refactoring some of this code to avoid duplication

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
amaltaro: That's a good question. As mentioned in my first comment, I didn't add any timeout because that means I'd have to query every single file to be splitted and compare their insert time against the timeout value.
If there are not enough events, it will get merged once the fileset gets closed, which is when the last job for that task completes. I've just run further tests and they confirm it works.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
amaltaro: It was in the top but I moved it over here because - I think - it would break CRAB.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
amaltaro: The EventAware algos have the same code, not the LumiBased. I'm usually not in favor of having these x-imports, so I think I'll just add it to the base class then.

src/python/WMCore/JobSplitting/LumiBased.py
amaltaro: This one is different because it does all the work with lumis

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
ericvaandering: Do we have to worry about 12 or 24 hour cleanup of un-merged space?

test/python/WMCore_t/WMBS_t/JobSplitting_t/EventAwareLumiBased_t.py
amaltaro: who said that?!?! :P I'll fix it later.

test/python/WMCore_t/WMBS_t/Subscription_t.py
amaltaro: Agreed! Will fix it

test/python/WMCore_t/WMBS_t/Subscription_t.py
amaltaro: Yep :)

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
hufnagel: First, this job splitter should never run on unmerged input. Second, even our default merge job splitters don't have timeouts. If you have a processing workflow with two jobs and the two jobs finish two weeks apart, the output from the first job will just sit on unmerged space for two weeks (assuming they are too small to be elevated to merged immediately). WMAgent has always worked that way. I do have timeouts for Repack and RepackMerge, but that's mostly because people are super-paranoid about RAW (plus we have/had problems with the StorageManager bookkeeping that sometimes kept a Repack workflow in limbo for weeks).

amaltaro: Just a correction, these splitters can run on unmerged data for some specific taskchain setup (KeepOutput=False)

hufnagel: Ok, unmerged cleanup time is 2 weeks for most sites. "Normally" that should not be an issue but in case of workflow tails we definitely do run the risk of loosing these files.

Problem is, adding a safety timeout of lets say a week requires to run a more expensive data discovery query for the job splitter. Inert time isn't available from the wmbs_file_details table, you need to join with wmbs_fileset_files. That join requires another join with wmbs_subscription. Neither of these two joins are very expensive, so I think the performance implications should be minimal.

hufnagel: Thing is, if this is really a problem Ops should be seeing this already. Maybe they are just working around it ? Worth asking them and see how much these worries are justified ?

amaltaro:
@ericvaandering I've just updated this patch and all your comments/concerns should've been taken care of. Thanks!

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
vlimant: what could be the problem for Ops ? loosing unmerged files because site would be deleting files after 2 weeks ? the actual policy is not 2 weeks, but "2 weeks and not in the list of protected directories" which will protect any deletion of the unmerged output of on-going workflows. Unless sites don't play by the rule, we are safe.
If the timeout of 2 weeks is on collecting enough work for the subsequent job, will this impact the completion time significantly ?

amaltaro: It can impact in the case there is a last job or two going into cooloff over and over. The change proposed relies on:
1. enough events or lumis for at least one job, properly splitted
2. when the fileset is closed, so all the input files were made available

and not on a time limit for files available for splitting. That would slightly increase the complexity and load on the database.
Perhaps a better approach would be making an in-memory cache for subscriptions and first time a file was made available., if we deem necessary to have it. Otherwise this patch is good as is.

amaltaro:
From my side, it's good to go.
Feel free to review and merge.

ericvaandering (approved):

hufnagel:
@amaltaro , do you still need a review from me ? I skimmed through the code but I don't have time to really go through it. What are the functional changes ?

amaltaro:
Dirk, I have updated the PR description (1st comment).

hufnagel:
Thanks, this make it clearer to me. Looking at the code, if I have a job splitter that is NOT using the newly added getFilesSortedByLocation method to discover it's input, it would not be affected by these changes, right ?
So for the Tier0, this only changes how we do the PromptReco job splitting ?

amaltaro:
That's correct. It changes the behaviour only for LumiBased and EventAware* algos.

amaltaro:
@ticoann can you review and merge it? Then we can run further integration tests, though I expect no issues :)

hufnagel (approved):
Looks good for me

ticoann (approved):

PR #8082, opened 2017-08-21
'map' fix applied 

vilnius-dev:
Wraps `map()` in a list call. It also changes `map(None, x)` to `list(x)`.
Using from `future_builtins import map` disables this fixer.
Fix applied because `map()` don’t return lists anymore in Python 3.

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -88,7 +88,7 @@ class PSetLister:
         self.psets.append(psetPath)
         params = childParameters(psetPath, pset)
         self.parameters[psetPath] = params
-        map(self, childPSets(pset))
+        list(map(self, childPSets(pset)))
         self.queue.pop(-1)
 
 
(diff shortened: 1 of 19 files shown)


vkuznet:
@ticoann , we should look at if this PR will have some performance issues due to change of data-type for map. Could you please test results here.

ticoann:
test this please

ericvaandering:
On python 2 this shouldn't matter as map already produces a list. On python 3, we should get the same behavior as python 2 which might not be optimal (using iterators is nicer) but shouldn't be any worse.

mapellidario:
This is covered by our gradual migration. However, we follow the performance suggestions at https://python-future.org/compatible_idioms.html#map, which suggests

```python
from builtins import map  # in py2 "map = itertools.imap" , so that py2 and py3 behave the same
mynewlist = list(map(f, myoldlist))
assert mynewlist == [f(x) for x in myoldlist]
```

instead of

```python
# Python 2 and 3: option 1
# Idiomatic Py3, but inefficient on Py2
mynewlist = list(map(f, myoldlist))
assert mynewlist == [f(x) for x in myoldlist]
```

I would suggest closing this PR without merging, since it is superseded by our current efforts.

mapellidario:
Thanks Roman for your work! Closing this

PR #8083, opened 2017-08-22
Minor change to exitCodes in ErrorHandler

amaltaro:
Complement to #8050 

And T0 overrides that value to an empty list, so there is nothing to do for T0 agent.


amaltaro:
@ticoann patch has been updated, hopefully I got everything we discussed.

ticoann:
Thanks Alan, patch looks good. I will merge it when Jenkins comes back.

PR #8084, opened 2017-08-22
'division_safe' fix applied

vilnius-dev:
This fix used instead of libfuturize.fixes.fix_division.
  
**Division**  
Integer division (rounding down):  
```
# Python 2 only:
assert 2 / 3 == 0
```
```
# Python 2 and 3:
assert 2 // 3 == 0
```
“True division” (float division):
```
# Python 3 only:
assert 3 / 2 == 1.5
```
```
# Python 2 and 3:
from __future__ import division    # (at top of module)

assert 3 / 2 == 1.5
```

“Old division” (i.e. compatible with Py2 behaviour):
```
# Python 2 only:
a = b / c            # with any types
```
```
# Python 2 and 3:
from past.utils import old_div

a = old_div(b, c)    # always same as / on Py2
```

### DIFF
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -8,6 +8,8 @@ process/config but does not depend on any CMSSW libraries. It needs to stay like
 
 """
 from __future__ import print_function
+from __future__ import division
+from past.utils import old_div
 import logging
 import pickle
 import traceback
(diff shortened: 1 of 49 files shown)


ericvaandering:
This one, I think, requires some thought and we should make an attempt to discern which of these things are expecting ints and which are expecting floats as a result. 

This will take some effort, but will leave us with better, less ugly code.

vilnius-dev:
@ticoann  As we discussed on meeting this PR could be closed.  
I making changes "by hand", results can be viewed here: https://github.com/dmwm/WMCore/pull/8125

PR #8089, opened 2017-08-22
Fix check for number of agents in a team

### LINKED ISSUE #8087: adjust the number of pending jobs to number of agents in the a team
I think I saw a mechanism to adjust the number of pending jobs to the number of active agents. How does that work ?

amaltaro:
Fixes #8087

We must get min from the same data type...
I've also deleted the misleading code where variables are set every polling cycle.


amaltaro:
@ticoann tests look good. Please review and merge

ticoann (approved):

PR #8093, opened 2017-08-25
Implement minute heartbeat for the components and monitor it

### LINKED ISSUE #8109: Agent in red in wmstats but no components/details given
It looks like I finally found the reason for it - I think - this might be triggered when we have a component down but without any component details. This is an example of my agentInfo doc:
```
   "down_components": [
       "PhEDExInjector"
   ],
   "timestamp": 1504025561,
   "down_component_detail": [
   ],
   "agent_version": "1.1.6.pre7"
```

I'll debug it further tomorrow, just wanted to keep it recorded.

amaltaro:
Fixes #8109

Changes so far are:
* while the threads are sleeping, send a heartbeat every 60secs;
* monitor heartbeat (last_updated) based on update_threshold (decreased from 6000 to 1800). Ideally it should be smaller, like 10min, but I'm afraid there are components that may spend > 10min in a busy cycle;
* update the db with a (nice) error when a worker thread is gracefully terminated (TODO: update all the components overriding it).

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -462,6 +462,15 @@ config.AnalyticsDataCollector.summaryLevel = "task"
 config.AnalyticsDataCollector.couchProcessThreshold = 50
 config.AnalyticsDataCollector.pluginName = None
 
+config.component_("ArchiveDataReporter")
+config.ArchiveDataReporter.namespace = "WMComponent.ArchiveDataReporter.ArchiveDataReporter"
+config.ArchiveDataReporter.componentDir = config.General.workDir + "/ArchiveDataReporter"
+config.ArchiveDataReporter.pollInterval = 300
+config.ArchiveDataReporter.WMArchiveURL = None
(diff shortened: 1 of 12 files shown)


amaltaro:
@ticoann no need to review it yet. However, I do have a couple of questions about code that can be possible removed.

I can't find this stuff
https://github.com/dmwm/WMCore/tree/master/src/python/WMCore/Agent/Flow
being used anywhere. Apparently it comes from the ProdAgent era.

And second is about the database backends
https://github.com/dmwm/WMCore/tree/master/src/python/WMCore/Agent/Database
we have setup for both couchdb and sqllite.

Should we go ahead and remove all that code?

amaltaro:
Seangchan, I still have to fix the unit test and the terminate method, but would you mind having a look at this patch, please? Thanks

amaltaro:
Finally \o/

@ticoann since you haven't created a tag yet, how about you review&merge this PR and then cut a new tag for the weekend?

I still have work to do regarding these components monitoring, but I think it will be mostly agent side now.

ticoann (approved):

ticoann:
@amaltaro, Alan patch looks good.

PR #8094, opened 2017-08-25
condor monitor report fix

ticoann:
@goughes, Erik, I found some minor bug on drain monitoriing. please review.

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
+++ b/src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
@@ -2051,7 +2051,7 @@ WMStats.ViewModel = (function (){
             summary.RequestPriority = requestInfo.RequestPriority;
 
             //TODO get the site white list from acdc db. (site might be overflowed)
-            // assignment args
+            // assign more value for acdc assignment
             summary.Team = requestInfo.Teams[0];
             summary.TrustPUSitelists = false;
             summary.TrustSitelists = false;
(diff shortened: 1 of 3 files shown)


src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
```diff
@@ -4402,7 +4402,6 @@ WMStats.namespace('ResubmissionList');
             htmlstr += "<li><b>ACDC DB Name:</b><input type='text' name='ACDCDatabase' size=30  value='wmagent_acdc'/></li>";
         }
         htmlstr += "<li><b>RequestPriority:</b> " + summary.RequestPriority + "</li>";
-        htmlstr += "<li><b>RequestType:</b> " + summary.RequestType + "</li>";
```
goughes: This line always gets removed when running combineMinifyWMStats.py, I added it back in the drain stats PR.  Not sure if it is needed or not.

goughes:
Ahh, yes good catch.  I must have forgotten to change this after I added the CondorAPI class.  This patch looks good.  Check my comment about the minify script to see what you think.

src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
ticoann: Actually, it seems this code was changed manually before. (Not using combinedMinifyWMStats.py) That is why it is still there. This should be removed. Thanks for pointing out.

ticoann:
tested in vm

PR #8098, opened 2017-08-25
make sure no empty string in Teams workqueue elements

### LINKED ISSUE #8095: agent need to acquire in their team

ticoann:
fixes #8095

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -195,7 +195,7 @@ class WorkQueue(WorkQueueBase):
 
         if isinstance(self.params['Teams'], basestring):
             self.params['Teams'] = [x.strip() for x in \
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan please review

amaltaro:
Seangchan, I think we should catch any of these problems right in the beginning. If you look at StdBase, you see `Team` is properly validated, but couch update sort of messes that up later on.

I'm fixing the  root cause at the moment and then we can see how to proceed.

ticoann:
For root cause of fix, we should remove either Teams or Team and keep just one. If we never allow multiple teams we can just keep 'Team'. It looks like that is current assumption anyway (in WorkQueue)

amaltaro:
Seangchan, how about this patch. Has it been applied anywhere?

ticoann:
This patch is not necessary. Closing it.

PR #8099, opened 2017-08-25
Add initial port of unified scripts: utils and transferor (DO NOT MERGE)

vkuznet:
This implementation ports WmAgentScripts/Unified
- port utils.py and trasferor.py scripts
- code refactoring and clean-up
- new components:
  - Common.py # set of common utilities for Unified
  - SiteInfo.py # site information module
  - RequestInfo.py # request information module
  - Transferor.py # transferor.py script
  - Stageout.py # initial placeholder for stageout module
  - Injector.py # initial placeholder for Injector module
I run local pylint rules and adjusted all issues with original code.

@ticoann , @amaltaro , @ericvaandering Please review initial commit for transferor module. Please fire up jenkins tests too since I need its input for pylint/pycode/pep/etc. tests. Here is a link to its output: http://bit.ly/2wvqAJR

I heavily rely on pycurl multi fetch url and sso I committed earlier. I was able to get all transferor info (dbs, phedex, sites, monitoring, detox, etc.) under 400 sec. Now code is ready to start serious refactoring towards better throughput. So far, I didn't yet place any phedex requests though. But in aforementioned link you'll see times breakdown for various components. So far the major time contributor is SiteInfo cache and getting site list for all workflows.


vkuznet:
I applied few bugs and made another round of code testing. It recovered missing datasets and obviously took longer time. New log is here: http://bit.ly/2vw1aHq

vkuznet:
@ticoann , @amaltaro , @ericvaandering Please review new changes. I added:
- initial unit tests for MicroService and Unified components (will expand in a future to cover Unified functions)
- speed up code by factor of 2 by using multi-fetch request's specs
- code re-factoring, polishing to confirm with pylint/pycodestyle suggestions

Now code runs 150 sec per 50 workflows, making calls in parallel to dbs/phedex,reqmgr_cache, etc.  The initialization time is quite large to acquire all data from various sources. Later it is used extensively through the code. Therefore the init time for SiteInfo module affects unit test too.

vkuznet:
@ticoann, @ericvaandering  , Seangchan, Eric, I need a valid proxy file to perform tests, since SiteInfo class queries lots of cervices. Can you instruct me how to ensure that it is present on jenkins node (should I rely simply on X509_USER_PROXY).

And, for some sites I relies on cern-get-sso-cookie tool. Is it present on jenkin's node?

ericvaandering:
> On Aug 28, 2017, at 3:11 PM, Valentin Kuznetsov <notifications@github.com> wrote:
> 
> @ticoann <https://github.com/ticoann>, @ericvaandering <https://github.com/ericvaandering> , Seangchan, Eric, I need a valid proxy file to perform tests, since SiteInfo class queries lots of cervices. Can you instruct me how to ensure that it is present on jenkins node (should I rely simply on X509_USER_PROXY).
> 
Yeah, we have lots of tests that rely on certificates (not sure if there is a proxy, but we could alter things to make one if you require it).


> And, for some sites I relies on cern-get-sso-cookie tool. Is it present on jenkin's node?
> 
> 

Not sure. Shahzad, do we have this on the cms-dmwm-cc7 nodes? If not, can we install it? Otherwise, we could move these particular tests into a Docker container same as we tried in the past, but I’d prefer not to.


Cheers,

Eric

ericvaandering:
On 8/29/2017 1:15 AM, Eric W Vaandering wrote:

>
>> On Aug 28, 2017, at 3:11 PM, Valentin Kuznetsov 
>> <notifications@github.com <mailto:notifications@github.com>> wrote:
>>
>> @ticoann <https://github.com/ticoann>, @ericvaandering 
>> <https://github.com/ericvaandering> , Seangchan, Eric, I need a valid 
>> proxy file to perform tests, since SiteInfo class queries lots of 
>> cervices. Can you instruct me how to ensure that it is present on 
>> jenkins node (should I rely simply on X509_USER_PROXY).
>>
> Yeah, we have lots of tests that rely on certificates (not sure if 
> there is a proxy, but we could alter things to make one if you require 
> it).
>
>
>> And, for some sites I relies on cern-get-sso-cookie tool. Is it 
>> present on jenkin's node?
>>
>>
>
> Not sure. Shahzad, do we have this on the cms-dmwm-cc7 nodes?
cern-get-sso-cookie should be available on dmwm nodes on next puppet run 
(i.e. with in an hour).
Please note that tests can gain access to cookies generated by this 
tool, proxy certificates and ssh keys of cmsbuild user, so make sure 
that limited number of ppl can trigger tests after validating the changes.

--Shahzad
> If not, can we install it? Otherwise, we could move these particular 
> tests into a Docker container same as we tried in the past, but I’d 
> prefer not to.
>
>
> Cheers,
>
> Eric
>
>

vkuznet:
> > Not sure. Shahzad, do we have this on the cms-dmwm-cc7 nodes?
> cern-get-sso-cookie should be available on dmwm nodes on next puppet run 
> (i.e. with in an hour).
> Please note that tests can gain access to cookies generated by this 
> tool, proxy certificates and ssh keys of cmsbuild user, so make sure 
> that limited number of ppl can trigger tests after validating the changes.
> 
> --Shahzad

Shahzad, thanks for prompt action. Please note, the code I used relies on
python tempfile.NamedTemporaryFile module with delete=True option (by default),
therefore it wipes out a cookie file once test is done.

vkuznet:
@ericvaandering , @ticoann , what should I do with these unit tests?
```
WMCore_t.Lexicon_t.LexiconTest:testPhysicsGroup was added with status success
nose.failure.Failure:runTest was added with status error. Must be fixed
```
there are not part of this PR and I think should be excluded somehow.

ericvaandering:
The lexicon thing is going to go away. It's a race condition between the current master and the last time Jenkins took a reference point. I think the other thing is real and is a clear indication that an import statement is not working. You can what's broken on the Jenkins page itself.

vkuznet:
Eric, thanks for the tip. I look-up Jenkins build and it failed to look-up unifiedConfiguration.json file which I need to deal with in a code, I read it like this:
```
fname = os.getenv('UNIFIED_CONFIG_JSON', 'unifiedConfiguration.json')
```

The question is how I can supply this json file to jenkins? Where should I put it in a test area or code area. I don't know a-priory what is a cwd when code runs or what is a root area. If I know that I can put this config over there or I may keep it around with a source code. Please advise, then the nose run test will proceed.

src/python/WMCore/Services/MicroService/Unified/Common.py
```diff
@@ -0,0 +1,290 @@
...
+
+def monitoringUrl():
+    "Return monitoring url"
+    return 'http://cmsmonitoring.web.cern.ch/cmsmonitoring'
+
+def dbsUrl():
+    "Return DBS URL"
+    return 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader'
+
+def reqmgrUrl():
```
amaltaro: Not really a review, just skimming the code. 
Valentin, can you set ALL these variables in a configuration file instead? Having a function to simply return a hardcoded value doesn't look right to me.

amaltaro:
You're probably still working on it. But I see some functions/calls to DBS and PhEDEx that could be using the Services APIs instead.

vkuznet:
Alan,
the reason I didn't use DBS Service APIs is simple, they don't support
concurrent pattern, i.e. I can't places N calls at once. And, for Unified
it is critical since it needs to acquire information about many workflows
at once. Therefore I used pycurl_managet.getdata API which does that
and query DBS and others via its data-service interface. In
a code you'll many places where it has been used as multi_getdata(urls)
pattern where urls is a list of urls which should be placed concurrently.
This approach significantly reduces time required to obtain necessary
information from CMS data-services and scale much better.
Best,
Valentin.

vkuznet:
Alan,
that's the plan, once I finish code re-factoring. In fact, I'll combine both
approaches. The hard-coded values will be default, while config will provide
primary choice, e.g.
```
def dbsUrl():
    return config.get('dbsUrl', 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader')
```
This will allow to avoid missing configuration parameters as well as use
testbed or other urls.
Best,
Valentin.

ericvaandering:
For the config file you should be able to do like this: 

```
from WMCore.WMBase import getTestBase

basePath = getTestBase()
```

and then add whatever you need to find the exact file you need on the file system. No need to actually know the cwd

ericvaandering:
Valentin, what about refactoring the DBS and PhEDEx code out to something like Services/DBSConcurrent.py to give us a base for doing this stuff in a smarter way down the road? No need to implement more of the service than you need, but could give us a nice road map for the future.

vkuznet:
Eric, I would agree that such functionality should be introduced at lower
level that others can take advantages of common code base. In fact, it is
since it resides in pycurl module. Do we need to generalize it further, I
don't know. Since so far its pretty generic already, e.g.
```
urls = [] # user list of urls to query concurrently
```
What the urls are is irrelevant for this code, it can be DBS/Phedex or their
combinations. For other examples you can look in pycurl module directly.

So far, I'm concentrating on porting huge code base and diverging yet another
route will slow down this task significantly. I think we can do it later if
necessary and I need to understand what exactly do you meant by refactoring DBS
and PhEDEx code. If you want to modify dbs client APIs to provide concurrent
support it would be independent task to address, etc.

yuyiguo:
DBS Client does multiple tasks such as, reusing the connection for the same user, this will save a lot of time; it will split for multiple calls if the input url is longer than 8190 bytes.    I did not see the  multi_getdata function had these basic functions built in.

yuyiguo:
Valentin:

We studied the connection reuse while developing DBS and found that CMS front end authorization was time consuming.  When you have N calls to DBS simultaneously, you have to make N connect objects. I am not sure how this N objects are reused. I would like to develop something that can reuse the connection.  DBS client does support input list to make one DBS calls, if the list is too long, DBS client will split it into multiple calls in sequence. Client will combine the results together to delivery to the user.   I agree with you that we should do the test you mentioned. My guess is that when the N is a small number, DBS client may be faster since it does not need to make a lot of connections because these connection is time consuming. However, when N is huge, multiple call simultaneously would help.

vkuznet:
@ericvaandering Eric, I changed code to reuse existing Singleton class (class is identical of what I put earlier, may be I added it). But it resides in not optimal module either, in WMCore.Cache.WMConfigCache and I would prefer to have it in Utils since it is common class. Anyhow, I move slowly adding more logic and components we need for Unified.

ericvaandering:
I agree it makes more sense in Utils. Feel free to move it in this PR or in a dedicated one.

ticoann:
Valentin, I have look at the patch in depth for the logic. A couple of suggestions.
For Unified config. I think we can put it in reqmgr_aux db. so if that need to be changed, we don't have to access the file directly.  I will create the patch you can use.

About the getting cert and key we can reuse the code.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L397

I will take that about from Request class so it can be used stand alone function.

vkuznet:
I moved singleton class from WMCore.Cache.WMConfigCache into Utils.Utilities, see https://github.com/dmwm/WMCore/pull/8116

I'll need to rebase this code base once #8116 will be merged.

vkuznet:
I created a new branch #8122 without master merges during rebase.

PR #8106, opened 2017-08-29
don't allow empty string team

### LINKED ISSUE #8095: agent need to acquire in their team

ticoann:
fixes #8095

### DIFF
--- a/src/couchapps/WorkQueue/lists/workRestrictions.js
+++ b/src/couchapps/WorkQueue/lists/workRestrictions.js
@@ -55,7 +55,11 @@ function(head, req) {
         };
 		
         var ele = row["doc"]["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
-
+        // team shouldn't be empty string or undefined
+        if (teams.length === 0 || !ele["TeamName"]) {
+            continue;
+        }
+        
         // check work is for a team in the request
         if (teams.length && ele["TeamName"] && teams.indexOf(ele["TeamName"]) === -1) {
             continue;


src/couchapps/WorkQueue/lists/workRestrictions.js
```diff
@@ -55,7 +55,11 @@ function(head, req) {
         };
 		
         var ele = row["doc"]["WMCore.WorkQueue.DataStructs.WorkQueueElement.WorkQueueElement"];
-
+        // team shouldn't be empty string or undefined
+        if (teams.length === 0 || !ele["TeamName"]) {
```
amaltaro: The line below is performing almost the same checks. How about we either add `|| teams.indexOf(ele["TeamName"]) === -1` to this if and remove the if below or keep the if
below with only the indexOf check?

amaltaro (commented):
Functionality looks good. Made a comment in the code for an improvement though.
Perhaps you could also update the PR title to something like 'Don't pull work if team name is an empty string'

amaltaro:
Seangchan, is this the fix you applied to the hepcloud and relval team? I'm updating the twiki we forget it. Once this patch is updated, we should also apply this patch to the other relval agent (down atmo).

src/couchapps/WorkQueue/lists/workRestrictions.js
ticoann: Yes you are right! I can modify that but I need to check why all the unittest failed.

ticoann:
Unfortunately, this patch need to apply in GQ not just LQ. With LQ application it just prevent job to be created but not prevent pulling the job. I will do some damage control. I will stop the WorkQueueManagers in relval and hepcould for a while.

amaltaro:
Seangchan, as we quickly discussed, how about you also change this code:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py#L215

such that it considers only non-empty strings? Something like `request['Teams'] = [t for t in request['Teams'] if t]`

ticoann:
closing it replaced by #8108

PR #8108, opened 2017-08-29
temp fix for empty team name handling (never merge)

### LINKED ISSUE #8095: agent need to acquire in their team

ticoann:
fixes #8095 @amaltaro, Alan I think this should prevent pulling down the team with empty string


amaltaro:
Patch looks good, Seangchan.

ericvaandering:
test this please

amaltaro:
test this please

ericvaandering:
test this please

ticoann:
@amaltaro, The unittest failure is real.  Since unittest assumes 'TeamName' is optional in WQ. 
but that shouldn't be the case. Anyway, I am not going to merge this patch since it is temp patch for production agents. Alan's team patch should fix this problem.

ticoann:
real fix by Alan is added

PR #8116, opened 2017-08-31
Move Singleton from WMCore.Cache/WMConfigCache into Utils.Utilities

vkuznet:


src/python/Utils/Utilities.py
```diff
@@ -3,6 +3,16 @@
 import time
 import subprocess
 
+
+class Singleton(type):
```
ericvaandering: It would make sense to me to put a class into its own module/file.

vkuznet:
Eric, that was my original thought, but I thought that class is too tiny
to reside in a single module. I'm ok with either way. Let @ticoann express
his wishes.
Best,
Valentin.

ticoann:
Valentin, I am OK with having tiny module with just Singleton with Singleton name since it is very generic and will be used many place for visibility. However if you are considering to add some other related classes it would be fine with different name. (I guess that is why you make as "Pattern")
I think you forgot to push Pattern module in the patch. :-)

ticoann:
test this please

ticoann:
test this please

amaltaro:
test this please

vkuznet:
@ticoann should I correct pylint suggestion in src/python/WMCore/Cache/WMConfigCache.py? They are not related to my changes but cause pylint failure.

ericvaandering:
test this please

ericvaandering:
@vkuznet anything you can do easily and safely, please do. But don't refactor other code, for instance, to change the over-riding of id()

vkuznet:
Yup, I saw the id usage and understand that it can affect lots of code since it
is used in ctor. Now PR should be clean.

ticoann:
Thanks Valentin, could you squash the commit as well. One for actual logic change (not pylint change). It seems Jenkins is having some trouble. So I will just run the integration test directly and merge if I don't find a problem.

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -322,7 +323,8 @@ def loadByView(self, view, value):
 
         if len(viewRes['rows']) == 0:
             # Then we have a problem
-            logging.error("Unable to load using view %s and value %s" % (view, str(value)))
```
ericvaandering: This will stop the warning, but what it's actually asking you to do is 
```
logging.error('msg %s %s', view, value)
```
instead of 
```
logging.error('msg %s %s' % (view, value)'
```
The first way performs better by not having to construct the strings when the message will be dropped (especially good for debug messages).

ericvaandering:
Jenkins is OK now, I think

vkuznet:
Eric,
I adjusted code to the same style used in a code. All logging calls is done via
msg = # define
logging.method(msg)
But you're right the actual suggestion asked to use arguments. I thought having
consistent code style is better.

ericvaandering:
OK. No problem.

PR #8117, opened 2017-08-31
refactor getKeyCert method

ticoann:
move getKeyCert method from Request class. So it can be used in other code.

### DIFF
--- a/src/python/Utils/CertTools.py
+++ b/src/python/Utils/CertTools.py
@@ -1,0 +1,52 @@
+from __future__ import division, print_function
+import os
+
+
+def getKeyCertFromEnv():
+    """
+    gets key and certificate from environment variables
+    If no env variable is set return None, None for key, cert tuple
+
+    First preference to HOST Certificate, This is how it set in Tier0
+
+    """
+
+    envPairs = [('X509_HOST_KEY', 'X509_HOST_CERT'),  # First preference to HOST Certificate,
+                ('X509_USER_PROXY', 'X509_USER_PROXY'),  # Second preference to User Proxy, very common
+                ('X509_USER_KEY', 'X509_USER_CERT')]  # Third preference to User Cert/Proxy combinition
+
+    for keyEnv, certEnv in envPairs:
+        key = os.environ.get(keyEnv)
+        cert = os.environ.get(certEnv)
+        if key and cert and os.path.exists(key) and os.path.exists(cert):
+            # if it is found in env return key, cert
+            return key, cert
+
+    # TODO: only in linux, unix case, add other os case
+    # look for proxy at default location /tmp/x509up_u$uid
+    key = cert = '/tmp/x509up_u' + str(os.getuid())
+    if os.path.exists(key):
+        return key, cert
+
+    # Finary look for globaus location
+    if (os.environ.get('HOME') and
+        os.path.exists(os.environ['HOME'] + '/.globus/usercert.pem')  and
+        os.path.exists(os.environ['HOME'] + '/.globus/userkey.pem')):
+
+        key = os.environ['HOME'] + '/.globus/userkey.pem'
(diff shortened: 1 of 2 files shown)


ticoann:
@vkuznet, @amaltaro please review.
@ericvaandering, is it OK to put in Utils package?

ericvaandering:
How about Services? 

I intended Utils as things that are useful for just about any python project to deal with things like lists, etc. Certificates are something a bit less broadly useful. 

In the end I don't have a very strong opinion though.

vkuznet:
changes looks fine to me, and, I would probably move them from request module to utils. They may be used in Services and others modules but they are generic functions.

ticoann:
I cleaned up code a little more. So Eric prefers under Service, but Valentin prefers under Utils, right? 😄 , I guess it is Alan's call. 😄  @amaltaro

amaltaro:
I think it could go under Utils/Utilities.py
OR... we create it just under Services, like the UUID function.

ticoann:
Alan, you are not helping. :-)

amaltaro:
LoL. It's a hard pick, but seems to fit better under Utils, since it's generic enough and has not a single bit of WMCore code. Anyways, depending on Eric's offer, I can easily change my mind :P

Seangchan, I think we can rename the module as well (and get rid of that typo), how about CertTools or something like that?

ticoann:
CertTools is better. I was thinking that this module can be expanded if you have some common CMS related environment value handling but that might be not so common. Anyway, updated the module name to CertTools

ericvaandering:
Why do we have Utils/Utilities? As an uninformed person, what am I going to expect to find there?

ericvaandering:
Yes, I realize I am not helping either. :-)

vkuznet:
I treat Utils as a set of helpers. These helpers should be generic and free
from dependencies, therefore it will allow to use them anywhere in a codebase.
Based on this definition
```
def bla():
    return 1
```
is a helper function. You may certain structure within Utils to better address
the use cases. That's why my vote for Utils, would it be Utilities or Singleton
is a separate topic for discussion.

The Services on a contrary provides set of actions to codebase, they already
depend on independent utilities and represent more sophisticated set of tools
dedicated to an action.

Hope this helps.

ticoann:
> Why do we have Utils/Utilities? As an uninformed person, what am I going to expect to find there?

Yes, you are not helping. 😄  I agree we that we can use more specific module name there but that might require more changes  outside of WMCore

ticoann:
Valentin, that is valid point. I was just wondering whether certificate is only used for making a "Request" and nowhere else and whether the code itself is too specific (i.e. searching order of the certificates) to be used broadly outside Service package. Anyway, I will leave that in Utils then. Thanks.

ericvaandering:
@vkuznet I agree with you, but that's not my point. Everything in Utils is already a Utility. But lets not just throw everything in to an already poorly named file called "Utilities.py". So far Utilities.py contains a bunch of functions for generating lists, converting strings, benchmarking, counting processes, and doing a "df". What do they have in common? Nothing. Let's stop this madness now.

If certificate stuff is going to go here, make a file called "certificates". Singleton should go in Singleton.py or Patterns.py if we think there will be more.

vkuznet:
Good, we have a consensus. I agree with Certificates.py (since it may include
other cert related stuff or sign-ons) and Patterns.py to which Singleton and
others belongs.

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

ticoann:
@ericvaandering, Eric, could you check whether Jenkins has some problem when you get a chance?

ericvaandering:
test this please

ticoann:
test this please

PR #8119, opened 2017-09-01
fix wmstats agent info api

ticoann:


ticoann:
wrong patch closing it.

PR #8123, opened 2017-09-03
Fix worker sleep bug from 8093

amaltaro:
Apparently this logical operator was haunting me in the last week :(
Fix bug introduced in #8093.


amaltaro:
I might cut a pre11 tag tomorrow with this patch in. However, this is only present on the agent side, so we can also wait for a couple of days and see if further patches are needed or not for cmsweb.

ticoann:
Thanks Alan, I would never find this problem. This this the bug which make the thread super in steroid? 😄

amaltaro:
Yes, because worker threads were NOT sleeping and just running the whole algorithm again. It would have been a catastrophe to have it in production :)

ticoann:
I guess Agent would be super busy.  :-) Anyway, that was the problem I saw when I tested 1.1.6.pre10 agent. Now it seems working normally. However I am not sure why MemoryCache didn't work when I increase the timeout.

PR #8127, opened 2017-09-05
increase cache expire for aborted workflow check

ticoann:


src/python/WMCore/Cache/GenericDataCache.py
```diff
@@ -2,9 +2,11 @@
 import time
 import traceback
 import logging
+from future.utils import with_metaclass
 
+from Utils.Patterns import Singleton
 
-class MemoryCacheStruct(object):
+class MemoryCacheStruct(with_metaclass(Singleton, object)):
```
amaltaro: Does it mean there will be only one cache object (MemoryCacheStruct actually) shared among all the components in the agent? And another object shared in ReqMgr?
If the answer is yes, then how we make it thread safe?

src/python/WMCore/Services/ReqMgr/ReqMgr.py
```diff
@@ -206,15 +206,15 @@ def updateRequestProperty(self, request, propDict):
...
+    def getAbortedAndForceCompleteRequestsFromMemoryCache(self, expire=10):
         """
         _getAbortedAndForceCompleteRequestsFromMemoryCache_
         """
         # imports here to avoid the dependency not using this function
         from WMCore.Cache.GenericDataCache import MemoryCacheStruct
 
         maskStates = ["aborted", "aborted-completed", "force-complete"]
-        return MemoryCacheStruct(expire=0, func=self.getRequestByStatus, initCacheValue=[],
+        return MemoryCacheStruct(expire=expire, func=self.getRequestByStatus, initCacheValue=[],
```
amaltaro: 10 seconds should be good. I'm just afraid the cache might be stale for some workflows and, for instance, a request transition can be not allowed and fail. Or it can't happen?

amaltaro (commented):
A few questions before I give my final answer :)

src/python/WMCore/Cache/GenericDataCache.py
ticoann: No it means it will be shared among in one component.  This is used in WorkQueueManager and ErrorHandler. Each component will have its own Singleton object.

ticoann: It is not thread safe.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
ticoann: You are right. I think it it safer to set expire=0 as default

ticoann:
@amaltaro, Alan I updated the patch. I shouldn't use Singleton. It is supposed used with GenericDataCache (which supposed be Singleton). I am not sure this patch is actually taking effect. I will put the debugging statement whether it is honoring expire time.

ticoann:
Now it is tested. It works as expected.

amaltaro (approved):
It seems you've also found and fixed an old bug :-)
Looks good to me, Seangchan!

ticoann:
Yes, that was why it was always updating the cache. 😄

PR #8142, opened 2017-09-07
More attempts to get rid of bad pylint errors

ericvaandering:


ericvaandering:
Also looks good to merge. As with others, once we agree I will squash commits

src/python/WMCore/Storage/FileManager.py
```diff
@@ -7,22 +7,16 @@
...
-from WMCore.Storage.DeleteMgr import DeleteMgr
-from WMCore.Storage.Registry import retrieveStageOutImpl, RegistryError
+from WMCore.Storage.Registry import RegistryError, retrieveStageOutImpl
 from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig
+from WMCore.Storage.StageOutError import StageOutError, StageOutFailure
+
+log = logging
 
-import WMCore.Storage.Backends
-import WMCore.Storage.Plugins
```
amaltaro: We need to keep these 2 imports otherwise we don't get the plugins. Since I made this mistake already, how about you add a comment above to avoid accidental deletions?

amaltaro (changes requested):
@ticoann Since it touches the StageOut code, I'd rather keep this *out* of the next wmagent release as well.

src/python/WMCore/Storage/FileManager.py
ericvaandering: OK, @amaltaro this is fixed. Thanks!

ticoann:
@ericvaandering, Eric patch looks good to me. Could you squash the commits? Alan, I will exclude for this for WMAgent release as you said.

ericvaandering:
Squashed. Please merge @ticoann

ticoann (approved):

Issue #8141, opened 2017-09-07
Fix jinja_template function


ericvaandering:
https://github.com/dmwm/WMCore/blob/e32ebd4793c96e18ebd358db119abef978a76a76/src/python/WMCore/ReqMgr/Web/tools.py#L126 has an error. There is no variable ```quote```

vkuznet:
My mistake, it is missing quote function. Here is its code:
```
import cgi
def quote(data):
    """
    Sanitize the data using cgi.escape.
    """
    if  isinstance(data, int) or isinstance(data, float):
        res = data
    elif  isinstance(data, dict):
        res = data
    elif  isinstance(data, list):
        res = data
    elif  isinstance(data, long) or isinstance(data, int) or\
          isinstance(data, float):
        res = data
    elif  isinstance(data, ObjectId):
        res = str(data)
    else:
        try:
            if  data:
                res = cgi.escape(data, quote=True)
            else:
                res = ""
        except Exception as exc:
            print_exc(exc)
            print("Unable to cgi.escape(%s, quote=True)" % data)
            res = ""
    return res
```
Feel free to add it to the code. In jinja templates we can use certain functions. The way we pass it is via kwargs.update call as shown in Web/tool.py line 126

vkuznet:
ok, here is a fix https://github.com/dmwm/WMCore/pull/8145

ericvaandering:
#8145 fixed this

PR #8145, opened 2017-09-07
Fix for #8141, add quote function definition to jinja templates

vkuznet:


amaltaro:
Valentin, please define a meaningful PR subject and in the description you can say
```
Fixes #8141
and whatever you want to say, it doesn't hurt to provide some comments
```

amaltaro:
And the same is valid for the commit message :)

vkuznet:
Alan, enjoy new title :)

ericvaandering:
@vkuznet you do need the print function import. After that, it's good to merge.

ericvaandering (changes requested):
Please use the compact form of is instance and add the future import

src/python/WMCore/ReqMgr/Web/tools.py
```diff
@@ -34,6 +31,31 @@
...
+    """
+    Sanitize the data using cgi.escape.
+    """
+    if  isinstance(data, (int, float)):
+        res = data
+    elif  isinstance(data, dict):
+        res = data
+    elif  isinstance(data, list):
+        res = data
+    elif  isinstance(data, (int, long, float)):
```
amaltaro: Given all these if's have do the same thing, why don't we do a single `if ... else ...`? BTW, besides long, the last one is dup. 
Also, remove all these double space please. We follow the pep-8 convention and this will just create pylint complaints whenever someone change this file.

amaltaro (changes requested):

vkuznet:
Alan, you're correct and changes were applied to fix what you suggested.

amaltaro (approved):

amaltaro:
Thanks, Valentin!

amaltaro:
@ticoann since it affects ReqMgr only, I think we can also merge this one before the next wmagent release candidate.

PR #8146, opened 2017-09-07
Explicitly set LFNBase during creation - Chained requests

### LINKED ISSUE #8151: MergedLFNBase not taken into account during creation of chained requests
Both TaskChain and StepChain don't use the value provided for setting tasks/outputs/etc, they use the default value (/store/data). Even though the final request dict contains the correct value (the one provideed by the user).

amaltaro:
Fixes #8151
Since (un)mergedLFNBase is not mandatory during assignment (the same as AcqEra/ProcStr/ProcVer), we need to make sure the workload configuration (steps, output modules, etc) are coherent with the value passed during request creation.

These 2 spec files were missing that call, so even though the request json contains one MergedLFNBase, the workload config contains the default /store/data

Given that the workflow team always set those 2 parameters, I'd rather not merge and push it to the current testbed because I'm not sure whether it can have - bad - side effects, especially on StepChain.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
@@ -97,6 +97,8 @@ class StepChainWorkloadFactory(StdBase):
             self.setupNextSteps(firstTask, arguments)
 
         self.workload.setStepMapping(self.stepMapping)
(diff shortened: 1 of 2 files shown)


vlimant:
that will not make lfnbase required at request creation time, will it ?

amaltaro:
No, it won't. However, it should be taken into account if provided.

amaltaro:
test this please

PR #8147, opened 2017-09-07
Replace WMCore.ReqMgr.Utils.url_utils getdata with one from WMCore.Services.pycurl_manager and RequestHandler

vkuznet:


amaltaro:
@ticoann since it affects ReqMgr only, I think we can also merge this one before the next wmagent release candidate.

amaltaro (approved):

PR #8150, opened 2017-09-08
bulk update method

ticoann:
Initial patch, still need to create the unittest and workqueue part change.

### DIFF
--- a/src/python/WMCore/Database/CMSCouch.py
+++ b/src/python/WMCore/Database/CMSCouch.py
@@ -8,7 +8,7 @@ http://wiki.apache.org/couchdb/API_Cheatsheet
 
 NOT A THREAD SAFE CLASS.
 """
-from __future__ import print_function
+from __future__ import print_function, division
 
 import time
 import urllib
@@ -20,8 +20,8 @@ import traceback
 from httplib import HTTPException
 from datetime import datetime
 
+from Utils.IteratorTools import grouper, nestedDictUpdate
 from WMCore.Services.Requests import JSONRequests
-from WMCore.Lexicon import replaceToSantizeURL
 
 
 def check_name(dbname):
@@ -345,6 +345,45 @@ class Database(CouchDBRequests):
             updateUri = '/%s/_design/%s/_update/%s/%s' % \
                         (self.name, design, update_func, doc_id)
             return self.put(uri=updateUri, data=fields, decode=False)
+
+    def updateBulkDocuments(self, doc_ids, paramsToUpdate, updateLimits=1000):
+
+        uri = '/%s/_bulk_docs/' % self.name
+        conflictDocIDs = []
+        for ids in grouper(doc_ids, updateLimits):
+            # get original documens
+            docs = self.allDocs(options={"include_docs": True}, keys=ids)['rows']
+            data = {}
+            data['docs'] = []
+            for j in docs:
+                doc = {}
+                doc.update(j['doc'])
+                nestedDictUpdate(doc, paramsToUpdate)
+                data['docs'].append(doc)
+
+            if len(data['docs']) > 0:
+                retval = self.post(uri, data)
+                for result in retval:
+                    if result.get('error', None) == 'conflict':
+                        conflictDocIDs.append(result['id'])
+
+        return conflictDocIDs
+
+    def updateBulkDocumentsWithConflictHandle(self, doc_ids, updateParams, updateLimits=1000, maxConflictLimit=10):
+        """
+        param: doc_ids: list couch doc ids for updates, shouldn't contain any duplicate or empty string
+        param: updateParams: dictionary of parameters to be updated.
+        param: updateLimits: number of documents in one commit
+        pram maxConflictLimit: number of conflicts fix tries before we give up to fix it to prevent infinite calls
+        """
+        conflictDocIDs = self.updateBulkDocuments(doc_ids, updateParams, updateLimits)
+        if len(conflictDocIDs) > 0:
+            # wait a second before trying again for the confict documents
+            if maxConflictLimit == 0:
+                return conflictDocIDs
+            time.sleep(1)
+            self.updateBulkDocumentsWithConflictHandle(conflictDocIDs, updateParams, maxConflictLimit=maxConflictLimit - 1)
+        return []
 
     def putDocument(self, doc_id, fields):
         """
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -1089,19 +1089,8 @@ class WorkQueue(WorkQueueBase):
                         processedInputs = []
                         for unit in work:
                             processedInputs.extend(unit['Inputs'].keys())
-                        if processedInputs or rejectedWork:
-                            chunkSize = 20
-                            chunkProcessed = processedInputs[:chunkSize]
-                            chunkRejected = rejectedWork[:chunkSize]
-                            # TODO:Make this a POST update, instead of several PUT
-                            while processedInputs or rejectedWork:
-                                self.backend.updateInboxElements(inbound.id, ProcessedInputs=chunkProcessed,
-                                                                 RejectedInputs=chunkRejected,
-                                                                 options={'incremental': True})
-                                processedInputs = processedInputs[chunkSize:]
-                                rejectedWork = rejectedWork[chunkSize:]
-                                chunkProcessed = processedInputs[:chunkSize]
-                                chunkRejected = rejectedWork[:chunkSize]
+                        self.backend.updateInboxElements(inbound.id, ProcessedInputs=processedInputs,
+                                                                 RejectedInputs=rejectedWork)
                         # if global queue, then update workflow stats to request mgr couch doc
                         # remove the "UnittestFlag" - need to create the reqmgrSvc emulator
                         if not self.params.get("UnittestFlag", False):
--- a/src/python/WMCore/Services/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/Services/WorkQueue/WorkQueue.py
@@ -1,6 +1,6 @@
 import json
 from collections import defaultdict
-from WMCore.Database.CMSCouch import CouchServer
+from WMCore.Database.CMSCouch import CouchServer, CouchConflictError
 from WMCore.Lexicon import splitCouchServiceURL
 from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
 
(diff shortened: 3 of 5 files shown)


ticoann:
Updated unittests and WorkQueue changes. @amaltaro, Alan please review

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -346,6 +347,39 @@ def updateDocument(self, doc_id, design, update_func, fields={}, useBody=False):
...
+        conflictDocIDs = []
+        for ids in grouper(doc_ids, updateLimits):
+            # get original documens
+            docs = self.allDocs(options={"include_docs": True}, keys=ids)['rows']
+            data = {}
+            data['docs'] = []
+            for j in docs:
+                doc = {}
+                doc.update(j['doc'])
+                doc.update(paramsToUpdate)
```
amaltaro: I guess we don't have inner dicts to be updated, do we?

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -346,6 +347,39 @@ def updateDocument(self, doc_id, design, update_func, fields={}, useBody=False):
...
+                data['docs'].append(doc)
+
+            if len(data['docs']) > 0:
+                retval = self.post(uri, data)
+
+                for idx, result in enumerate(retval):
```
amaltaro: If there are several thousands documents and several conflicts, it might be that `idx` will have the same value as previous iterations. How about we store the doc id instead?

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -346,6 +347,39 @@ def updateDocument(self, doc_id, design, update_func, fields={}, useBody=False):
...
+                        conflictDocIDs.append(idx)
+
+        return conflictDocIDs
+
+    def updateBulkDocumentsWithConflictHandle(self, doc_ids, updateParams, updateLimits=1000):
+
+        conflictDocIDs = self.updateBulkDocuments(doc_ids, updateParams, updateLimits)
+        if len(conflictDocIDs) > 0:
+            # waite a second before trying again for the conflit documents
+            time.sleep(1)
```
amaltaro: A few comments on this block:
* do you think this sleep is crucial in here? given we fetch the docs with _all_docs, that doesn't require any views/lists and the delay on couch should be closer to 0.
* to avoid an infinite loop, how about we limit it to 5 attempts or so?
* last but not least, how about we log everytime we hit doc conflicts, saying the number of conflicts out of XX docs.

amaltaro (changes requested):
I think there is one thing that must be solved, otherwise the code looks good in general.

Seangchan, do you think it would be simple (like 30min) to check the performance difference between the current setup and the new one using bulk updates? Like with 5k docs or so.

src/python/WMCore/Database/CMSCouch.py
ticoann: I don't think there is nested dict to update in WQ

src/python/WMCore/Database/CMSCouch.py
ticoann: Idx is doc id. With the post call, it returns doc_id and status.

src/python/WMCore/Database/CMSCouch.py
ticoann: I am not sure whether delay is crucial. I think conflict happens when different updates happens at at the same time. Agent doesn't use bulk update yet, but if the too bulk updates happens at the same time it potentially can have a lot of conflicts so I thought we could wait a little to avoid constant update try. 

Other two points are very valid. I will add those.

ticoann:
Yes I will test performance between bulk and individual update. Still unittest failure seems to be real. Although I am sure what cause that. I am investigating

src/python/WMCore/Database/CMSCouch.py
amaltaro: What's the format of `retval`? I insist :), this `idx` will simply be a counter from 0 to len(retval), and, not the doc id.

ticoann: Sorry you're right. I thought I fixed this before. Apparently not. I will update the patch.

ticoann:
I ran some performance test which is not very accurate since it on connect local couch, document site is small, there could be other facts affecting this.
5000 document update. 
1. update individually - 42 sec
2 update bulk with 1000 chuck (5 calls) - 2 sec

amaltaro:
Wow! 2 secs for the 5k docs or for each slice? Not bad at all! :)

ticoann:
There are other factors involved though. The difference in between two other than bulk and one aspect.
The formal needs to retrieve all the content in the doc and the latter is in place change. 
So it will be less call for bulk but it will be much more data transfer. However it WQE is pretty small per doc and we can control size of the bulk. I think it will be overall helpful but might have some side effect.
Also the test skips authentication so it will be beneficial in that aspect in real situation.
Anyway, I have to fix the unittest in WQ which is really odd.

ticoann:
test this please

ticoann:
@amaltaro, Alan I updated the patch as you suggested. Could you review again? Thanks

ticoann:
test this please

ticoann:
@amaltaro, Alan finally fix all the unittest. There are 6 unittests failing in WQ but not because of this patch. I will try to look at them in separate patch. Please review.

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -346,6 +346,46 @@ def updateDocument(self, doc_id, design, update_func, fields={}, useBody=False):
...
+                doc = {}
+                doc.update(j['doc'])
+                nestedDictUpdate(doc, paramsToUpdate)
+                data['docs'].append(doc)
+
+            if len(data['docs']) > 0:
+                retval = self.post(uri, data)
+                for result in retval:
+                    if result.get('error', None) == 'conflict':
+                        conflictDocIDs.append(result['id'])
```
amaltaro: Seangchan, did you check whether the doc `id` is really reported back in case of a conflict?

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -346,6 +346,46 @@ def updateDocument(self, doc_id, design, update_func, fields={}, useBody=False):
...
+        param: updateLimits: number of documents in one commit
+        pram maxConflictLimit: number of conflicts to before we give up to fix it to prevent infinite call
+        """
+        conflictDocIDs = self.updateBulkDocuments(doc_ids, updateParams, updateLimits)
+        if len(conflictDocIDs) > 0:
+            # waite a second before trying again for the conflit documents
+            if maxConflictLimit == 0:
+                return conflictDocIDs
+            time.sleep(1)
+            self.updateBulkDocumentsWithConflictHandle(conflictDocIDs, updateParams, maxConflictLimit - 1)
```
amaltaro: Third argument has to be a kwd argument (maxConflictLimit=maxConflictLimit - 1). BTW, nice recursive function!

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -346,6 +346,46 @@ def updateDocument(self, doc_id, design, update_func, fields={}, useBody=False):
...
+
+        """
+        param: doc_ids: list couch doc ids for updates, shouldn't contain any duplicate or empty string
+        param: updateParams: dictionary of parameters to be updated.
+        param: updateLimits: number of documents in one commit
+        pram maxConflictLimit: number of conflicts to before we give up to fix it to prevent infinite call
+        """
+        conflictDocIDs = self.updateBulkDocuments(doc_ids, updateParams, updateLimits)
+        if len(conflictDocIDs) > 0:
+            # waite a second before trying again for the conflit documents
```
amaltaro: typo "waite" and "conflit". A few others in L378 and I think blank line in L373 can go away.

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -1089,19 +1089,8 @@ def processInboundWork(self, inbound_work=None, throw=False, continuous=False):
...
-                            while processedInputs or rejectedWork:
-                                self.backend.updateInboxElements(inbound.id, ProcessedInputs=chunkProcessed,
-                                                                 RejectedInputs=chunkRejected,
-                                                                 options={'incremental': True})
-                                processedInputs = processedInputs[chunkSize:]
-                                rejectedWork = rejectedWork[chunkSize:]
-                                chunkProcessed = processedInputs[:chunkSize]
-                                chunkRejected = rejectedWork[:chunkSize]
+                        self.backend.updateInboxElements(inbound.id, ProcessedInputs=processedInputs,
+                                                                 RejectedInputs=rejectedWork)
```
amaltaro: Cool! What about that "incremental" option?

amaltaro (changes requested):
Code looks good. There is only 1 minor comment that has to be updated and a few other aesthetic improvements.

I was going to ask what happens with those docs that hit a conflict in all attempts, but I see an error message is logged. I don't want to over-complicate it :), but a better warning would be writing it to LogDB.

Just for the record, I looked up at one WQ doc and `Mask` is one of the fields that could give us problems in this dict.update call. However, I don't think we ever update it, do we? We create that field with a specific value during creation and then it's never changed again (?)

amaltaro:
Ahh, I forgot mentioning. Since you created a unittest, how about creating a second one (and maybe a third one) that would hit a doc conflict with maxRetries>0 (and another with maxRetries=0)?

src/python/WMCore/WorkQueue/WorkQueue.py
ticoann: incremental option was needed since update handler was getting the data in request url since everything passed to request body now, there is no need for "incremental" option.

src/python/WMCore/Database/CMSCouch.py
ticoann: I haven't tested, I just check the document. (I am not sure how to create the conflict in test setting).

https://wiki.apache.org/couchdb/HTTP_Bulk_Document_API

One other option is we added option _all_or_nothing, but that will create the conflict.

src/python/WMCore/Database/CMSCouch.py
ticoann: thanks, I will update

src/python/WMCore/Database/CMSCouch.py
ticoann: good catch, thanks Alan,

src/python/WMCore/Database/CMSCouch.py
amaltaro: Don't we need the rev number for the doc update? If so, you could add the same doc id twice to the `data` to be updated. That might yield a conflict.

Yeah, that last line in this couch documentation is not very positive.

ticoann: rev number is not needed explicitly, The way update here woks is

1.  get all the documents (list of lds are given) first
2. rev is returned from first call and used for update.

ticoann:
test this please

PR #8154, opened 2017-09-08
add missing codecs import

### LINKED ISSUE #8038: Cheerypy threads log rotation

ticoann:
fixes #8038, I am not sure why we didn't get any errors on the test. (I guess doesn't fall into that path). @amaltaro, Alan I can merge and create a new tag but since it didn't break anything we might not ask redeploying the testbed. What do you think?

### DIFF
--- a/src/python/WMCore/WMLogging.py
+++ b/src/python/WMCore/WMLogging.py
@@ -6,6 +6,7 @@ Logging facilities used in WMCore.
 """
 import time
 import logging
+import codecs
 from logging.handlers import HTTPHandler, RotatingFileHandler, TimedRotatingFileHandler
 
 # a new log level which is lower than debug


ericvaandering:
The Pylint reports are gone now, but I'm sure it caught it and put it in bold. Please check this stuff.

amaltaro:
I guess self.encoding never evaluates to true(?). I agree, since everything looks fine in testbed and my VMs, we don't need to cut a new tag for testbed.
My bad that I missed it in the pylint report. Thanks for fixing it, Seangchan.

PR #8158, opened 2017-09-11
Add timer context manager

ericvaandering:


ericvaandering:
@ticoann please merge (unless we have this kind of thing already)

ticoann:
@ericvaandering, Are we keeping the default precision? It seems default is 1/10000. I was wondering we just keep 1/100 sec?

amaltaro:
We have this function/decorator already:
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/Utilities.py#L65

ericvaandering:
@ticoann it uses whatever is the base timer. I like the high precision since I was timing some stuff that was well under 1/100th of a second.

@amaltaro this kind of proves my point about Utils/Utilities. I want to keep the context manager as well as the decorator, but I'll do a little refactoring to pull them together into one file. I also don't like the name "timeit" since that's the name of a python library as well.

amaltaro:
Which point? That it has all sorts of utils? Yes, timeit was a bad choice indeed.

ericvaandering:
@amaltaro what do you think now?

ericvaandering:
That Utils/Utilities is a grab bag of unrelated stuff. :-)

amaltaro (approved):
Looks good, thanks!

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

ericvaandering:
That failing test is a newly unstable test. This is ready to merge.

ticoann:
OK, thanks. It doesn't seems to be related but just make sure.

PR #8162, opened 2017-09-13
Delete some request manager 1 code

### LINKED ISSUE #8153: Remove remaining ReqMgr code
Eric, we moved to ReqMgr2 service long time ago, but ReqMgr code is located at different places in WMCore. I proposed a long time ago to Seangchan to clean-up old code. It seems to me that we safely delete:
WMCore/ReqMgr/Tools/WorkflowTools.py
WMCore/ReqMgr/Utils/url_utils.py
The later (getdata) is still in use in WMCore/ReqMgr/Web/ReqMgrService.py but it can be easily replaced with getdata method of pycurl_manager.py

Let's hear @ticoann Seangchan opinion.

ericvaandering:
Fixes #8153


vkuznet (approved):
+1

ericvaandering:
@vkuznet in the same directory, can we get rid of reqMgrClient.py? It's throwing many many errors about dbs3 not being defined. Which suggests to me its also not used.

vkuznet:
Yes, I think so.

PR #8165, opened 2017-09-14
Store each worker thread cycle elapsed time and polling cycle

amaltaro:
In short, improve the worker threads monitoring and store more info into the database.
* has a database schema change (wm_workers)
* store the polling cycle settings for each worker thread
* time the `algorithm` execution for every worker thread and store it in the database
* retrieve any results from `algorithm` and store it in the database as well. Can be used in the future for keeping track of e.g. how many jobs JobCreator created in that cycle, jobs submitted by JobSubmitter, etc.

This means, any future worker thread (inheriting from BaseWorkerThread) has to use the `@timeFunction` decorator. 

I might add to this PR:
* recording when a thread is terminated and distinguish between a graceful stop and a crash

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -104,6 +104,7 @@ class AgentStatusPoller(BaseWorkerThread):
         self.localCouchMonitor = CouchMonitor(self.config.JobStateMachine.couchurl)
         self.setUpCouchDBReplication()
 
+    @timeFunction
     def algorithm(self, parameters):
         """
         get information from wmbs, workqueue and local couch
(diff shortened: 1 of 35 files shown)


amaltaro:
Ok, only the couchdb related heartbeat tests are failing now:
WMCore_t.Agent_t.HarnessCouchBackend

AFAIK we don't really use this couchdb agent monitoring:
https://github.com/dmwm/WMCore/tree/master/src/python/WMCore/Agent/Database/CouchDB
do we?

@ticoann if you want I can fix those tests, otherwise I'm in favor of removing this couchdb monitoring code.

ticoann:
Yes I agree that if we don't use we should remove them

amaltaro:
Thanks, Seangchan. I'll create a new PR deleting that couch-based monitoring.

@hufnagel @ticoann I have just resumed my tests, but since I'd like to add it to the upcoming release, can you please start reviewing it?

amaltaro:
test this please

amaltaro:
I ran some integration tests and everything looks fine. The failing unit tests belong to a part of the code that's being removed in #8277 . 
From my side, it's ready to go.

ticoann (approved):

PR #8167, opened 2017-09-14
Use 2to3 to remove deprecated unit test asserts

ericvaandering:
Some of our code uses old, out of date, names for unit test assertion methods


ericvaandering:
@ticoann or @amaltaro please merge

PR #8100, opened 2017-08-25
add new exit code for SYSTEM_PERIODIC_REMOVE

ticoann:
fixes #7906

PR #8178, opened 2017-09-19
wait 10 sec before checking the log file.

ticoann:
additional fix for #8100

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -272,6 +272,8 @@ class SimpleCondorPlugin(BasePlugin):
                 if os.path.isfile(reportName):
                     os.remove(reportName)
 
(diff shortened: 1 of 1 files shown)


ticoann:

amaltaro:
So, this condor logs are checked by the JobStatusLite component, which has a polling cycle of 60secs. AFAIK we don't want to raise that cycle because it's - sometimes - already too large.

Let me try to come up with another possible fix, this sleep is evil and it will eventually hurt us. Moreover, we are not sure this 10secs of grace time is going to be enough.

ticoann:
Yes, you are right. I could delay whole things although it only delays when FWJR are not found. 
We could put the sleep outside the for loop (hoping) all the log files were transferred. Or we can only wait if we couldn't find the "Reason" string from the log, hoping there is not many jobs failed and actually didn't produce fwjr.

ticoann:
as Alan pointed out this could give more harm than benefit. Closing it. We could wait a little in the upper level outside job loop but we will always have a race condition

PR #8185, opened 2017-09-21
fix the deployment script

ticoann:
fix the deployment script causing the problem in FNAL machines.  1.1.6.patch3 agent has fixed version

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -72,7 +72,7 @@ config = Configuration()
 config.section_("Agent")
 config.Agent.hostName = serverHostName
 config.Agent.contact = "cms-comp-ops-workflow-team@cern.ch"
-config.Agent.teamName = "cmsdataops"
+config.Agent.teamName = "REPLACE_TEAM_NAME"
 config.Agent.agentName = "WMAgentCommissioning"
 config.Agent.agentNumber = 0
 config.Agent.useMsgService = False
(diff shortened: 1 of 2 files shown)


etc/WMAgentConfig.py
```diff
@@ -72,7 +72,7 @@
 config.section_("Agent")
 config.Agent.hostName = serverHostName
 config.Agent.contact = "cms-comp-ops-workflow-team@cern.ch"
-config.Agent.teamName = "cmsdataops"
+config.Agent.teamName = "REPLACE_TEAM_NAME"
```
ticoann: This change is not really needed but just for the clarity.

amaltaro: Hmm, that was the problem then?!? Don't we have to update that sed above in order to get the team name replaced by the team name provided in the command line? Or are you planning to manually update it after the deployment?

ticoann: You are right. Since I thought agent works on vocms machines. I assumed that it will be set somewhere else. I looked at the code then it seems sed is the only command change the team.
I updated the patch now it should work. unfortunately which means we need to have a new tag for agent.

I could change the script as following to avoid a new tag, But I would prefer create a new tag.
```
sed -i "s+Agent.teamName = cmsdataops+Agent.teamName = $TEAMNAME+" $MANAGE/config.py
```
Anyway I thought you are in the airplane. Stop working.

amaltaro:
Thanks! So is this the only diff between patch2 and patch3?

ticoann:
yes

PR #8187, opened 2017-09-21
adjust some tier0 workflows to automatic data tier in output module

hufnagel:
Tier0 repack and express has some special output modules, whose handling got broken by the automatic inclusion of data tier in output modules. In the Express spec there was also a couple places where the change in the input references was simply forgotten.

@amaltaro @ticoann , please review

Without this patch the Tier0 can't use recent WMCore versions. We will also need to merge this patch into the 1.1.6 branch and make a new patch tag. Should I just make a pull request from the same branch into the 1.1.6 branch ? Or do you want to use separate source branch ?

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -478,12 +478,12 @@ class AccountantWorker(WMConnectionBase):
             elif jobType == "LogCollect" and len(outputMap.keys()) == 0 and outputModules == set(['LogCollect']):
                 pass
             elif jobType == "Merge" and set(outputMap.keys()) == set(
-                    ['Merged', 'MergedError', 'logArchive']) and outputModules == set(['Merged', 'logArchive']):
+                    ['MergedRAW', 'MergedErrorRAW', 'logArchive']) and outputModules == set(['MergedRAW', 'logArchive']):
                 pass
             elif jobType == "Merge" and set(outputMap.keys()) == set(
-                    ['Merged', 'MergedError', 'logArchive']) and outputModules == set(['MergedError', 'logArchive']):
+                    ['MergedRAW', 'MergedErrorRAW', 'logArchive']) and outputModules == set(['MergedErrorRAW', 'logArchive']):
                 pass
-            elif jobType == "Express" and set(outputMap.keys()).difference(outputModules) == set(['write_RAW']):
+            elif jobType == "Express" and set(outputMap.keys()).difference(outputModules) == set(['write_RAWRAW']):
                 pass
             else:
                 failJob = True
--- a/src/python/WMCore/WMSpec/StdSpecs/Express.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Express.py
@@ -181,7 +181,8 @@ class ExpressWorkloadFactory(StdBase):
                 alcaSkimTask = expressTask.addTask("%sAlcaSkim%s" % (expressTask.name(), expressOutLabel))
 
                 alcaSkimTask.setInputReference(expressTask.getStep(expressRecoStepName),
-                                               outputModule=expressOutLabel)
+                                               outputModule=expressOutLabel,
+                                               dataTier="ALCARECO")
 
                 alcaTaskConf = {'Multicore': 1,
                                 'EventStreams': 0}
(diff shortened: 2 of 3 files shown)


hufnagel:
Not going to fix the pylint warnings and errors since they are about code I am not going to touch. At some point we probably should fix them (or put in pylint directives to ignore them), but I am not going to do it in the context of this pull request.

hufnagel:
Any feedback ?

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -478,12 +478,12 @@ def handleJob(self, jobID, fwkJobReport):
             elif jobType == "LogCollect" and len(outputMap.keys()) == 0 and outputModules == set(['LogCollect']):
                 pass
             elif jobType == "Merge" and set(outputMap.keys()) == set(
-                    ['Merged', 'MergedError', 'logArchive']) and outputModules == set(['Merged', 'logArchive']):
+                    ['MergedRAW', 'MergedErrorRAW', 'logArchive']) and outputModules == set(['MergedRAW', 'logArchive']):
                 pass
             elif jobType == "Merge" and set(outputMap.keys()) == set(
-                    ['Merged', 'MergedError', 'logArchive']) and outputModules == set(['MergedError', 'logArchive']):
+                    ['MergedRAW', 'MergedErrorRAW', 'logArchive']) and outputModules == set(['MergedErrorRAW', 'logArchive']):
```
ticoann: I am not quite understand these conditions. Shouldn't original conditions also need to be kept (['Merged', 'MergedError'])? Is it only the Repack case? Maybe Alan should review. @amaltaro.
Also it seem above two conditions can be combined.

hufnagel: Original case doesn't exist anymore since the output modules contain the RAW now at the end (per @amaltaro changes). Could be combined, but a combination wouldn't be any more readable code, at least not while being constrained to this exact use case.

The exact use case is RepackMerge, with one of the two possible output modules being dropped.

src/python/WMCore/WMSpec/StdSpecs/Express.py
```diff
@@ -181,7 +181,8 @@ def buildWorkload(self):
                 alcaSkimTask = expressTask.addTask("%sAlcaSkim%s" % (expressTask.name(), expressOutLabel))
 
                 alcaSkimTask.setInputReference(expressTask.getStep(expressRecoStepName),
-                                               outputModule=expressOutLabel)
+                                               outputModule=expressOutLabel,
+                                               dataTier="ALCARECO")
```
amaltaro: Instead of hard-coding the datatier, don't you have it in another variable?

src/python/WMCore/WMSpec/StdSpecs/Express.py
```diff
@@ -383,7 +385,7 @@ def addAlcaHarvestTask(self, parentTask, parentOutputModuleName,
 
         harvestTaskConditionHelper = harvestTaskCondition.getTypeHelper()
         harvestTaskConditionHelper.setRunNumber(self.runNumber)
-        harvestTaskConditionHelper.setConditionOutputLabel(condOutLabel)
+        harvestTaskConditionHelper.setConditionOutputLabel(condOutLabel+"ALCAPROMPT")
```
amaltaro: same here

amaltaro (commented):
Thanks for fixing it. These changes make sense as per L475.

src/python/WMCore/WMSpec/StdSpecs/Express.py
hufnagel: We are in a conditional "if" code segment here that checks for the dataTier to be ALCARECO. Seemed redundant to put the variable again.

src/python/WMCore/WMSpec/StdSpecs/Express.py
hufnagel: Again, this is a conditional code path that is only called for ALCAPROMPT.

src/python/WMCore/WMSpec/StdSpecs/Express.py
amaltaro: If you ever change this tier name, then you have one more place to update, so chances to miss it are higher :-)

PR #8193, opened 2017-09-25
add log for acdc wmbs injection (never merge - only for temp fix)

ticoann:


amaltaro:
Was it applied anywhere?

ticoann:
Alan, welcome back!. Yes it is applied on vocms0310 and vocms0308 but reversed for now. I didn't want to patch to permanent since I need a proper fix on this. (Please don't restart WorkQueueManager on those machines)

bbockelm:
This PR looks old - can it be closed out?

PR #8195, opened 2017-09-26
add more cpu performance matrics in WMArchive

### LINKED ISSUE #8192: Add TotalInitTime from FWJR
@Dr15Jones tells me that TotalInitTime is set in more recent cmsRun executions in the same place as TotalJobTime which tracks the time during the initialization phase. Also need to get this into WMArchive

ticoann:
fixes #8192  - still need to be tested.


ticoann:
@ericvaandering, @Dr15Jones, could you check whether parameter names are correct?
I will update pylint errors later.

ericvaandering:
They are correct

PR #8204, opened 2017-09-28
Automatically set MaxRSS according to Memory requirements

### LINKED ISSUE #8091: Remove the need for setting MaxRSS at assignment time
There is too much confusion and complications in having both "Memory" and "MaxRSS" in the schema. I cannot think of a reason we should not have only one of the two, and the other one handled purely internally

amaltaro:
Fixes #8091
Basic changes are:
* given the Memory requirements for a workload/task, automatically set the MaxRSS watchdog to the same value as the Memory one.
* creation default MaxVSize to an "infinite" number (1TiB)
* also changed the MaxRSS and MaxVSize unit from KiB to MiB

plus the following required changes:
* PerformanceMonitor watchdog works with both KiB or MiB MaxRSS/MaxVSize settings (TODO: remove the backwards compatibility in a couple of months)
* kill a job with the first `reason`, not the last one in case of multiple reasons
* MaxRSS and MaxVSize default assignment values were removed.
* Merge/Cleanup/Logcollect tasks aren't affected by Memory requirements

That means Ops will have to provide the correct rss/vsize values during assignment, only for old workflows (created before cmsweb upgrade). Newer workflows should have the correct settings and rss/vsize parameters can be skipped during assignment.

TODO: It also means we have to update all WMAgents with the Watchdog changes (KiB vs MiB).

There will be a transition phase here, but once CMSWEB upgrade is completed (next Tuesday Oct 10th), this is the impact on the clients:
* clients *must* provide rss/vsize for old workflows;
* clients don't need to provide rss/vsize for new workflows (created after the workflow is finished). Though it's still fine if it's provided;
* redundant, but clients *can* keep providing these rss/vsize settings during assignment for old or new workflows;

Moreover, if memory requirements are overwritten during assignment, that will properly set the maxRSS as well.

### DIFF
--- a/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
+++ b/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
@@ -7,6 +7,7 @@ Monitor object which checks the job to ensure it is working inside
 the agreed limits of virtual memory and wallclock time, and terminate it
 if it exceeds them.
 """
+from __future__ import division
 
 import os
 import signal
@@ -195,28 +196,31 @@ class PerformanceMonitor(WMRuntimeMonitor):
             msg += "command = %s\n" % self.monitorCommand
             logging.error(msg)
             return
-        rss   = float(output[2])
-        vsize = float(output[3])
+        # FIXME: making it backwards compatible. Keep only the "else" block in HG1801
+        if self.maxRSS is not None and self.maxRSS >= (1024 * 1024):
+            # then workload value is still in KiB (old way)
+            rss = int(output[2])
+            vsize = int(output[3])
+        else:
+            rss   = int(output[2]) // 1024  # convert it to MiB
+            vsize = int(output[3]) // 1024  # convert it to MiB
         logging.info("Retrieved following performance figures:")
-        logging.info("RSS: %s;  VSize: %s; PCPU: %s; PMEM: %s" % (output[2], output[3],
-                                                                  output[4], output[5]))
+        logging.info("RSS: %s;  VSize: %s; PCPU: %s; PMEM: %s", output[2], output[3],
+                     output[4], output[5])
 
         msg = 'Error in CMSSW step %s\n' % self.currentStepName
-        if self.maxRSS != None and rss >= self.maxRSS:
+        if self.maxRSS is not None and rss >= self.maxRSS:
             msg += "Job has exceeded maxRSS: %s\n" % self.maxRSS
             msg += "Job has RSS: %s\n" % rss
             killProc = True
             reason = 'RSS'
-        if self.maxVSize != None and vsize >= self.maxVSize:
+        elif self.maxVSize is not None and vsize >= self.maxVSize:
             msg += "Job has exceeded maxVSize: %s\n" % self.maxVSize
             msg += "Job has VSize: %s\n" % vsize
             killProc = True
             reason = 'VSZ'
-
-        #Let's check the running time
-        currentTime = time.time()
-
-        if self.hardTimeout != None and self.softTimeout != None:
+        elif self.hardTimeout is not None and self.softTimeout is not None:
+            currentTime = time.time()
             if (currentTime - self.startTime) > self.softTimeout:
                 killProc = True
                 reason = 'Wallclock time'
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -193,12 +193,13 @@ class StdBase(object):
         _addDashboardMonitoring_
 
         Add dashboard monitoring for the given task.
+        Memory settings are defined in Megabytes and timing in seconds.
         """
-        # A gigabyte defined as 1024^3 (assuming RSS and VSize is in KiByte)
-        gb = 1024.0 * 1024.0
-        # Default timeout defined in CMS policy
-        softTimeout = 47.0 * 3600.0 + 40.0 * 60.0
-        hardTimeout = 47.0 * 3600.0 + 45.0 * 60.0
+        # Default settings defined by CMS policy
+        maxrss = 2.3 * 1024  # 2.3 GiB, but in MiB
+        vsize = 1 * 1024 * 1024  # 1 TiB, but in MiB
+        softTimeout = 47 * 3600  # 47h
+        hardTimeout = 47 * 3600 + 5 * 60  # 47h + 5 minutes
 
         monitoring = task.data.section_("watchdog")
         monitoring.interval = 300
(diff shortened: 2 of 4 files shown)


amaltaro:
Well, better than I thought. I still need to run a few real tests and fix those unit tests.
@ticoann Seangchan, let's keep this one out of the next tag. Maybe I can get it properly tested by Monday and then we go for a new tag.

amaltaro:
This patch seems to be working Ok, but there are still two issues to be addressed:
1. MaxRSS and MaxVSize are not supported during assignment anymore, which means, workflows created before this patch will not have the proper rss/vsize settings! And one cannot override it, so what to do...
2. apparently StepChain uses only the top level Memory value. Perhaps we should simply reject workflow creation if Memory is inside the Step dict (?)

amaltaro:
test this please

amaltaro:
@ticoann while jenkins doesn't come back, can you please start reviewing this PR. I'd like to add it in for tomorrow's testbed.
@vlimant since you created that issue, can you please have a look at as well

amaltaro:
Giving it a second thought, I'm missing an assignment overwrite of Memory or Multicore.
So I rather couple Memory changes to maxrss and Multicore to maxvsize. Updating it, don't look at it yet.

ericvaandering:
test this please

amaltaro:
test this please

ticoann:
test this please

src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
```diff
@@ -195,28 +196,31 @@ def periodicUpdate(self):
...
             msg += "Job has exceeded maxVSize: %s\n" % self.maxVSize
             msg += "Job has VSize: %s\n" % vsize
             killProc = True
             reason = 'VSZ'
-
-        #Let's check the running time
-        currentTime = time.time()
-
-        if self.hardTimeout != None and self.softTimeout != None:
+        elif self.hardTimeout is not None and self.softTimeout != None:
```
ticoann: self.softTimeout != None:
-> 
self.softTimeout is not None:

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -461,6 +456,12 @@ def setupProcessingTask(self, procTask, taskType, inputDataset=None, inputStep=N
         prepID = taskConf.get("PrepID") or self.prepID
         procTask.setPrepID(prepID)
 
+        # has to be done in the very end such that child tasks are set too
+        procTask.setPerformanceMonitor(maxRSS=memoryReq,
+                                       maxVSize=multicore * 20 * 1024,  # ~20GB per core
```
ticoann: shouldn't this be taking from MaxVSize from spec? instead of hard coding ( 20 * 1024 )

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -1365,6 +1388,8 @@ def setNumberOfCores(self, cores, nStreams):
             taskStreams = nStreams
 
         if taskCores:
+            # if # of cores is changed, then also change the MaxVSize values
+            self.setMaxVSize(taskCores * 20 * 1024)  # 20 GiB per core
```
ticoann: same as above  - shoudn't be hard coded?

ticoann (approved):

ticoann:
Hi Alan patch looks good, I have a couple of questions (commented) in line.
It my concern is wrong assessment, please feel free to merge and create a new tag.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
hufnagel: MaxVSize is supposed to be infinite or something equivalent since we do not really limit on VSize.

src/python/WMCore/WMSpec/WMTask.py
hufnagel: Same as above, hardcode to 1TB would also work, but 20GB per core is also something ridiculously high and should also work. They are all equivalent to "no limit".

src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
amaltaro: I'll fix it.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
amaltaro: You're right. I keep confusing it with ImageSize :( I'll just set it to a very high number (1000 * 1024 MiB) in StdBase and I'm in favor of completely removing it in the next month.

amaltaro:
Thanks Dirk and Seangchan. I'll update this patch, run some quick tests and cut a new tag by lunch time or so.

amaltaro:
@ticoann @hufnagel jenkins are having issues and we need @ericvaandering to have a look at it with Shahzad.
Not to delay testbed even more, I'm merging this one (with your comments applied) and I can follow up on any unseen issues later today/tomorrow.

vlimant:
great ! how can we track the deployment of this to edit the client side that is setting the MaxRSS value ?

amaltaro:
I'm updating the PR description with this info. FYI @thongonary @areinsvo and @mcremone

src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
```diff
@@ -195,28 +196,31 @@ def periodicUpdate(self):
...
+            vsize = int(output[3]) // 1024  # convert it to MiB
         logging.info("Retrieved following performance figures:")
-        logging.info("RSS: %s;  VSize: %s; PCPU: %s; PMEM: %s" % (output[2], output[3],
-                                                                  output[4], output[5]))
+        logging.info("RSS: %s;  VSize: %s; PCPU: %s; PMEM: %s", output[2], output[3],
+                     output[4], output[5])
 
         msg = 'Error in CMSSW step %s\n' % self.currentStepName
-        if self.maxRSS != None and rss >= self.maxRSS:
+        if self.maxRSS is not None and rss >= self.maxRSS:
```
vlimant: would you mind adding an explicit unit for next deployment

src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
```diff
@@ -195,28 +196,31 @@ def periodicUpdate(self):
...
             msg += "Job has exceeded maxRSS: %s\n" % self.maxRSS
             msg += "Job has RSS: %s\n" % rss
             killProc = True
             reason = 'RSS'
-        if self.maxVSize != None and vsize >= self.maxVSize:
+        elif self.maxVSize is not None and vsize >= self.maxVSize:
             msg += "Job has exceeded maxVSize: %s\n" % self.maxVSize
```
vlimant: would you mind adding an explicit unit for next deployment

PR #8206, opened 2017-09-28
Acdc unmerged filter

### LINKED ISSUE #8196: ACDC with too many parents
Currently following workflows are skipped in vocms0308 and vocms0310 due to too many parents files need to be retrieved. (60K) Which takes up all the processing in WorkQueueMananger
We need to check whether parentage is needed or make more effient.

```
"mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053538_530", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053544_9947", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053551_9974", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053600_3305", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053608_5854", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053617_8169", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053625_8322", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053632_9709", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053641_9747", "mcremone_ACDC0_task_EGM-RunIISummer17GS-00002__v1_T_170924_053649_646"
```

ticoann:
fixes #8196 
@amaltaro, Alan please review. Still need to be tested.
First commit handles when documents with umerged parents are already in acdc db
Second commit prevent putting the documents with unmerged parents  in acdc db.

We can only cherry pick first commit for current agent if second commit is not tested throughly

### DIFF
--- a/src/python/WMCore/ACDC/DataCollectionService.py
+++ b/src/python/WMCore/ACDC/DataCollectionService.py
@@ -91,10 +91,16 @@ class DataCollectionService(CouchService):
             fileset = CouchFileset(database=self.database, url=self.url,
                                    name=taskName)
             coll.addFileset(fileset)
+            inputFiles = job['input_files']
+            for fInfo in inputFiles:
+                if fInfo["merged"] and ("parents" in fInfo) and \
+                   len(fInfo["parents"]) and ("/store/unmerged/" in next(iter(fInfo["parents"]))):
+                    # remove parents files from acdc doucment if they are unmerged files
+                    fInfo["parents"] = []
             if useMask:
-                fileset.add(files=job['input_files'], mask=job['mask'])
+                fileset.add(files=inputFiles, mask=job['mask'])
             else:
-                fileset.add(files=job['input_files'])
+                fileset.add(files=inputFiles)
 
         return
 
--- a/src/python/WMCore/WorkQueue/WMBSHelper.py
+++ b/src/python/WMCore/WorkQueue/WMBSHelper.py
@@ -387,7 +387,7 @@ class WMBSHelper(WMConnectionBase):
 
         wmbsFile['inFileset'] = True  # file is not a parent
 
-        logging.info("WMBS File: %s on Location: %s", wmbsFile['lfn'], wmbsFile['newlocations'])
+        logging.info("WMBS MC Fake File: %s on Location: %s", wmbsFile['lfn'], wmbsFile['newlocations'])
 
         self.wmbsFilesToCreate.append(wmbsFile)
 
@@ -676,11 +676,19 @@ class WMBSHelper(WMConnectionBase):
         adds the ACDC files into WMBS database
         """
         wmbsParents = []
-        for parent in acdcFile["parents"]:
-            parent = self._addACDCFileToWMBSFile(DatastructFile(lfn=parent,
-                                                                locations=acdcFile["locations"]),
-                                                 inFileset=False)
-            wmbsParents.append(parent)
+        # TODO:  this check can be removed when ErrorHandler filters parents file for unmerged data
+        # If files is merged and has unmerged parents skip the wmbs population
+        if acdcFile.get("merged", 0) and len(acdcFile["parents"]) and ("/store/unmerged/" in next(iter(acdcFile["parents"]))) :
+            # don't set the parents
+            pass
+        else:
+            # set the parentage for all the unmerged parents
+            for parent in acdcFile["parents"]:
+                logging.info("WMBS ACDC Parent File: %s", parent)
+                parent = self._addACDCFileToWMBSFile(DatastructFile(lfn=parent,
+                                                                    locations=acdcFile["locations"]),
+                                                     inFileset=False)
+                wmbsParents.append(parent)
 
         # pass empty check sum since it won't be updated to dbs anyway
         checksums = {}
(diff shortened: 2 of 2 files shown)


ticoann:
test this please

ticoann:
test this please

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -676,11 +676,17 @@ def _addACDCFileToWMBSFile(self, acdcFile, inFileset=True):
         adds the ACDC files into WMBS database
         """
         wmbsParents = []
-        for parent in acdcFile["parents"]:
-            parent = self._addACDCFileToWMBSFile(DatastructFile(lfn=parent,
-                                                                locations=acdcFile["locations"]),
-                                                 inFileset=False)
-            wmbsParents.append(parent)
+        # TODO:  this check can be removed when ErrorHandler filters parents file for unmerged data
+        if acdcFile.get("merged", 0) and len(acdcFile["parents"]) and "/umerged/" in acdcFile["parents"][0]:
```
amaltaro: In plain english: if input file is unmerged and there are unmerged parents, skip it. Is it correct? Maybe you can add this comment here just so I don't ask you again :-)

In addition to that, I'd replace the last check by `and acdcFile["parents"][0].startswith("/store/unmerged")`, just to avoid any possible "unmerged" somewhere in the file name.

src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -91,10 +91,16 @@ def failedJobs(self, failedJobs, useMask=True):
             fileset = CouchFileset(database=self.database, url=self.url,
                                    name=taskName)
             coll.addFileset(fileset)
+            inputFiles = job['input_files']
+            for fInfo in inputFiles:
+                if fInfo["merged"] and ("parents" in fInfo) and \
```
amaltaro: Same comment as below. I think `parents` key is always present, no? Just to be safe, it could be replaced by
```
if fInfo["merged"] and len(fInfo.get("parents", []) and fInfo["parents"].startswith("/store/unmerged"):
```

Now thinking about it in plain english: if the input file is merged and there are unmerged parents, skip the parents. I wonder whether we should skip any parents that are unmerged?

amaltaro (changes requested):
Even if you don't like my suggestions :), there is a typo `"/umerged/"`.

Seangchan, can you also provide a clear and concise description at the very top of this PR? We discussed it last week, but apparently I got confused once again.

src/python/WMCore/WorkQueue/WMBSHelper.py
ticoann: actually the condition says 
If input filed is "merged" and parents are unmerged we skip it. I will update the comments.
Ah, so it is always /store/unmerged?  I was thinking about that but not sure whether /store/ can be changed later so that is why I did /unmerged/ instead of just unmerged.
Anyway I will update that.

src/python/WMCore/ACDC/DataCollectionService.py
ticoann: I think parents key should be always present but I just checked for the safe side. I will change that.

ticoann: I think there shouldn't be a case if there is unmerged input and parents are unmerged.
It there are I think we have a bigger problem (parentage will be lost) I just want to record this in ACDC if that happens.

amaltaro: Seangchan, as we discussed yesterday, our only concern is related to the unmerged parents file.
How about we remove the "merged" part in order to make it more clear.
BTW, do you know whether retrieving the first element of that iterator is faster than a simple `fInfo["parents"][0]`? Otherwise I'd just stick to the direct access to the first element.

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -676,11 +676,19 @@ def _addACDCFileToWMBSFile(self, acdcFile, inFileset=True):
...
         """
         wmbsParents = []
-        for parent in acdcFile["parents"]:
-            parent = self._addACDCFileToWMBSFile(DatastructFile(lfn=parent,
-                                                                locations=acdcFile["locations"]),
-                                                 inFileset=False)
-            wmbsParents.append(parent)
+        # TODO:  this check can be removed when ErrorHandler filters parents file for unmerged data
+        # If files is merged and has unmerged parents skip the wmbs population
+        if acdcFile.get("merged", 0) and len(acdcFile["parents"]) and ("/store/unmerged/" in next(iter(acdcFile["parents"]))) :
```
amaltaro: Same comment about merged. And the above comment is the contrary of this `if` line, did you mean if input is *not* merged, then skip wmbs population (?)

src/python/WMCore/ACDC/DataCollectionService.py
ticoann: As we chat earlier, I think we should just check whether input file is merged file or not. We don't have to check about parents' file status (merged or unmered) but for the safe side, we just check parent file status. If you think it is safe to just check merged == 1 or not, we can change it later.

PR #8207, opened 2017-09-29
Explicitly allow an empty list without generating an error message

ericvaandering:
I noticed a lot of error messages coming out of the work unit addition code. No need for those from subscription.

### DIFF
--- a/src/python/WMCore/WMBS/MySQL/Jobs/AddWorkUnits.py
+++ b/src/python/WMCore/WMBS/MySQL/Jobs/AddWorkUnits.py
@@ -55,6 +55,8 @@ class AddWorkUnits(DBFormatter):
             binds = self.getBinds(jobFileRunLumis)
         elif jobid and fileid and run and lumi:
             binds = DBFormatter.getBinds(self, jobid=jobid, fileid=fileid, run=run, lumi=lumi)
+        elif not jobFileRunLumis and isinstance(jobFileRunLumis, list):  # Empty list is ok
+            return
         else:
             logging.error('Jobs.AddWorkUnits called with insufficient arguments')
             return


ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
@ticoann this is good to merge

src/python/WMCore/WMBS/MySQL/Jobs/AddWorkUnits.py
```diff
@@ -55,6 +55,8 @@ def execute(self, jobid=None, fileid=None, run=None, lumi=None, jobFileRunLumis=
             binds = self.getBinds(jobFileRunLumis)
         elif jobid and fileid and run and lumi:
             binds = DBFormatter.getBinds(self, jobid=jobid, fileid=fileid, run=run, lumi=lumi)
+        elif not jobFileRunLumis and isinstance(jobFileRunLumis, list):  # Empty list is ok
```
amaltaro: Don't we reach the same result with the `isinstance` check only (since the if above takes care of element'ed list)?

ericvaandering: No. If jobFileRunLumis is None, this line will not be true. Only if its an empty list.

amaltaro: That's my point, at this stage the only thing you care about is the data type. None or [] has the same outcome in the first statement (here comes my suggestion to keep only the second one)

ericvaandering: It's an and not an or. If all the parameters are None, it falls through to the end and generates an error message.

amaltaro:
Just to make sure we're talking about the same use cases, see the example below:
```
In [39]: def eric(jobFileRunLumis):
    ...:     if jobFileRunLumis:
    ...:         print "first if"
    ...:     elif not jobFileRunLumis and isinstance(jobFileRunLumis, list):
    ...:         print "second if"
    ...:         

In [40]: 

In [40]: eric(None)

In [41]: eric([])
second if

In [42]: 

In [42]: 

In [42]: 

In [42]: def eric2(jobFileRunLumis):
    ...:   if jobFileRunLumis:
    ...:      print "first if"
    ...:   elif isinstance(jobFileRunLumis, list):
    ...:      print "second if"
    ...:      

In [43]: eric2(None)

In [44]: eric2([])
second if
```

so both have the same outcome and the first statement is useless (for those inputs that I tested). Maybe you're expecting a different value/type for `jobFileRunLumis` ?

ericvaandering:
I see what you are saying. There are four possible branches and you are saying the first condition on the third branch is redundant, not that any of the branches are redundant. 

I actually prefer to keep both conditions on the third branch because I think it reads more clearly, not because there is a logical difference. :-)

src/python/WMCore/WMBS/MySQL/Jobs/AddWorkUnits.py
ticoann: @ericvaandering, I think Alan is right it seems it give the same result.
But I am not sure why this is needed, do you want to avoid error message printing out?
Is there case this calls with everything is None. If we don't expect that we should raise an error. Not just printing out the log message.

amaltaro:
test this please

amaltaro:
@ticoann how about we merge this one?

ticoann:
test this please

amaltaro:
It looks like the VM ran out of space and jenkins failed right in the beginning with
`fatal: write error: No space left on device`

I think it's time to bug @ericvaandering :)

jenkins log in https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/lastBuild/console

amaltaro:
test this please

ericvaandering:
Thanks! Probably need to log into VM and cleanup docker stuff

Sent from a mobile device.

On Feb 8, 2018, at 11:03, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


It looks like the VM ran out of space and jenkins failed right in the beginning with
fatal: write error: No space left on device

I think it's time to bug @ericvaandering<https://github.com/ericvaandering> :)

jenkins log in https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/lastBuild/console

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

ticoann:
test this please

amaltaro:
@ticoann since it touches only the work unit code, I think we can trust Eric and merge it :)

PR #8208, opened 2017-09-29
use the cache for agent config to prevent crashing when cmsweb is down

ticoann:
@amaltaro, Alan please review

### DIFF
--- a/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
+++ b/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
@@ -12,7 +12,6 @@ class ReqMgrAux(Service):
     API for dealing with retrieving information from RequestManager dataservice
 
     """
-
     def __init__(self, url, header=None, logger=None):
         """
         responseType will be either xml or json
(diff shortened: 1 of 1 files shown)


src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -127,9 +127,12 @@ def isDrainMode(config):
 
     reqMgrAux = ReqMgrAux(config.TaskArchiver.ReqMgr2ServiceURL)
     agentConfig = reqMgrAux.getWMAgentConfig(config.Agent.hostName)
-
-    return agentConfig["UserDrainMode"] or agentConfig["AgentDrainMode"]
-
+    if "UserDrainMode" in agentConfig and "AgentDrainMode" in agentConfig:
+        AUXDB_AGENT_CONFIG_CACHE = agentConfig
```
amaltaro: Isn't it another variable/id? I think we'd need to declare it `global` if we want to overwrite its value in here, no?

amaltaro (changes requested):
Thanks, Seangchan. I had it in my todo list too, but you beat me to the punch :)
Another option would be to keep the drain status cached on the component side (attr of the component), but I agree having a global cache would be better.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
ticoann: Yes you are right. Thanks I will update that.

ticoann:
test this please

amaltaro:
test this please

amaltaro:
This test works just fine on my VM. Triggering it one last time.

amaltaro:
test this please

amaltaro:
And that's a different test, so I assume we are Okay with this PR.

amaltaro (approved):

PR #8209, opened 2017-09-29
filter WMArchive report for performance data

ticoann:
1. send only selected parameters to WMArchive from performance data.
2. use int for NumberOfThreads and NumberOfStreams

### DIFF
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -2,6 +2,7 @@ from __future__ import (division, print_function)
 
 import socket
 import copy
+from collections import defaultdict
 
 # From top level
 
@@ -38,8 +39,8 @@ PERFORMANCE_TYPE = {'cpu': {'AvgEventCPU': float,
                             'TotalLoopCPU': float,
                             'TotalInitTime': float,
                             'TotalInitCPU': float,
-                            'NumberOfThreads': float,
-                            'NumberOfStreams': float},
+                            'NumberOfThreads': int,
+                            'NumberOfStreams': int},
                     'memory': {'PeakValueRss': float,
                                'PeakValueVsize': float},
                     'storage': {'readAveragekB': float,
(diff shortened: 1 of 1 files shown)


ticoann:
@vkuznet, @amaltaro, Valentin, Alan please review.
If this is OK, I will update WMArchive schema for those parameters to int as well.

src/python/WMCore/Services/WMArchive/DataMap.py
```diff
@@ -192,12 +194,14 @@ def typeCastPerformance(performDict):
                         value = performDict[key][param]
                         if value in ["-nan", "nan", "inf", ""]:
                             value = -1
-                        performDict[key][param] = PERFORMANCE_TYPE[key][param](value)
+                        if PERFORMANCE_TYPE[key][param] == int:
```
vkuznet: that's fine, but I would suggest to add explicit comment here to clarify what code is doing. Something like:
```
# the received value comes from CMSSW FWJR and its type is string
# first we convert it to float because int(string) may lead to error, e.g. int("3.0") will not work
# then we call PERFORMANCE_TYPE[key][param] converter over the value, in this case
# PERFORMANCE_TYPE[key][param] is int. The full chain would be int(float(string))
```

vkuznet (commented):
As I mentioned I'll add explicit comment explaining logic of conversion.

ticoann:
Thanks Valentin, I updated the comments.  I wanted get Alan's opinion as well, If Alan agrees I will merge it and create WMAgent branch patch.

ticoann:
test this please

ticoann:
test this please

ticoann:
test please

ticoann:
test this please

ticoann:
test this please

amaltaro:
test this please

amaltaro (approved):
Changes look good, Seangchan.
However, I think we can boost a bit the performance of this function by first iterating over `PERFORMANCE_TYPE` dict, since it's supposed to be a smaller set of `performDict`. Difference should be almost negligible for a few docs, but certainly good enough for millions of docs.

Anyways, it's up to you whether we change it now or later.

ticoann:
Thanks Alan, I updated the for loop order as  you suggested. Please review the logic (I always confused with this)

amaltaro:
Looks good, Seangchan!

Just another observation though, this patch will change the performance metrics behaviour w.r.t. what is propagated by the agent.
 - at the moment, we propagate ALL metrics. And those "expected" metrics are properly casted.
 - with this patch, only "expected" metrics will be uploaded (properly casted).

I guess that is the correct behaviour that we want in the end, don't we?

ticoann:
As far as I know, without this patch we propagate all the information about CPU performance and properly casted what was defined here. Other values are not casted and pass as it is (mostly string)
However, WMArchive filter any variables not defined in its schema. 
https://github.com/dmwm/WMArchive/blob/master/src/python/WMArchive/Schemas/FWJRProduction.py#L104
So only those defined above will be recorded. @vkuznet, Valentin, right?

With this patch we also filters out the parameters which is not defined above when we send the docs to WMArchive. So both will have the same result but we are sending less data with this patch.

vkuznet:
yes, I think it is correct definition.

PR #8210, opened 2017-10-01
Ensure that pycurl URL is string data type

vkuznet:


vkuznet:
@ticoann , Seangchan please review and merge. I found that this is required especially when we load something from external source, like JSON, and then pass it to pycurl. In this case url may be passed as unicode which cause a problem. It seems to be that naturally to make conversion here that pycurl code can handle both str and unicode url values.

ticoann:
Thanks Valentin,

PR #8225, opened 2017-10-03
Update deploy_unittest.sh

ericvaandering:
@ticoann @amaltaro any reason we shouldn't do this? SLC7 build is already being used for unit tests and the test of the deployment, but this is the version that ends up in the docker containers we build.


amaltaro:
Nope, I think it's the perfect time to move to slc7 rpms.

ericvaandering:
@ticoann you can merge. The Jenkins test of this is not going to show anything. :-)

PR #8228, opened 2017-10-03
 Add schedd max jobs per user constraint to JobSubmitter 

### LINKED ISSUE #8224: Throttle JobSubmitter to 100k or MAX_JOBS_PER_OWNER
as per https://cms-logbook.cern.ch/elog/GlideInWMS/5638 we need JobSubmitter to go idled and put a warning sign when reaching 95% of MAX_JOBS_PER_OWNER schedd classad 
FYI @bbockelm

amaltaro:
Fixes #8224 

First commit has a merge of CondorAPI and PyCondorAPI service classes.
Second commit has the actual change and implementation of this extra check condition on the JobSubmitter component, such that it can skip that cycle if there are way too many jobs in condor pool


src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -709,9 +711,16 @@ def algorithm(self, parameters=None):
         2) Find jobs for all the necessary sites
         3) Submit the jobs to the plugin
         """
+        myThread = threading.currentThread()
+
+        if not self.passSubmitConditions:
+            msg = "JobSubmitter didn't pass the submit conditions. Skipping this cycle."
+            logging.warning(msg)
+            myThread.logdbClient.post("JobSubmitter_submitWork", msg, "warning")
```
amaltaro: @ticoann Seangchan, it's probably much easier for you. All the unit tests failed with this error
`'_MainThread' object has no attribute 'logdbClient'`

I thought this attr would always be set for the components, since their super class is:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkerThreads/BaseWorkerThread.py#L150

Before I spend too much time on this, would you know how we can get this logdb defined for this thread?

ticoann: It is defined below,

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkerThreads/BaseWorkerThread.py#L154

The problem was test configuration doesn't contain config.General.central_logdb_url
and that result in myThread.logdbClient  as None.
I think it is a little trick to fix the unittest. 
we have to create the logDB in unittest or mock it. (I prefer mocking it).
Anyway, it might be better to remove this log db update and I will create the separate issue for unittest to be passed.

amaltaro: I might remove these logdb entries from the component then. I updated the test code to set central_logdb_url and it still fails to find that attr. Thanks, SEangchan!

ticoann: Thanks Alan, I will try to create the mock patch for this. I think we need those if we start to putting the log from different components

ticoann:
Alan, could you remove the JobSubmitter_t changes? I think #8232 is working without it.

amaltaro:
Sure, I've just removed the _t commit. Thank you very much, Seangchan!

amaltaro:
I still have to add back the logdbclient lines though, but we're not in a hurry and I'll wait for your mock PR to be merged first.

amaltaro:
test this please

amaltaro:
Haaa, unit tests don't complain anymore. Thank you very much, Seangchan!
@ticoann please start reviewing but don't merge it yet. I still have to run real tests with it.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Only unstable tests...

ticoann (approved):

ticoann:
We decided not to port this in wmagent branch, correct?

amaltaro:
I think we can backport it later, if needed. BTW, I still have to run a real test, hopefully this didn't break anything :-)

ticoann:
Alan, do you know what is the value for MAX_JOB_BY_OWNER. someone said it is 100K, in that case , would this help?

amaltaro:
Yes, it's 100k. I still have to play with that ad, make it much smaller and run plenty of jobs just to make sure it works as expected.

vlimant:
this threshold should also propagate and prevent the agent to get any new WQE

PR #8229, opened 2017-10-04
make datatier retrieval for physicsskims work with patch releases

hufnagel:
Sorry for only sending this now, is there a chance to get this also into a 1.1.16.patch6 ?


ticoann:
Dirk, is this a bug on Tier0 or in general?

hufnagel:
Only affects Tier0 and it's validated there in a replay.

PR #8232, opened 2017-10-04
add mock logdb

ticoann:
@ericvaandering, please review.  

Unittest using logdb need to be updated with something like below.

```
from WMQuality.Emulators.EmulatedUnitTestCase import EmulatedUnitTestCase
class ErrorHandlerTest(EmulatedUnitTestCase):
```

### DIFF
--- a/src/python/WMQuality/Emulators/Cache/MockMemoryCacheStruct.py
+++ b/src/python/WMQuality/Emulators/Cache/MockMemoryCacheStruct.py
@@ -1,0 +1,13 @@
+from __future__ import (division, print_function)
+
+
+class MockMemoryCacheStruct(object):
+    """
+    Version of Services/PhEDEx intended to be used with mock or unittest.mock
+    """
+
+    def __init__(self, *args, **kwargs):
+        print("Using MockMemoryCache")
+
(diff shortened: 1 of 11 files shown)


ericvaandering:
This looks like a good start. One piece of advice is the use some prints or what not initially to really make sure that the mock is being called. Getting the mocked code to be called from all uses was my biggest challenge.

ticoann:
@ericvaandering, I added print statement for MockClasses. Also added ApMon mock.
However, it doesn't seem to take Mock class. Could check what I am doing wrong?

ticoann:
Never mind Eric. I made some stupid mistake. I takes Mock classes.

ticoann:
@amaltaro, Alan I think this is ready. I created a few Mock classes to speed up the test and fix some of the exceptions which wasn't affecting result.
What takes long now is profiling test. (Which is expected but Jenkins won't run it)

```
@attr('integration')
    def testF_PollerProfileTest(self)
```
If you are OK I will merge it when Jenkins comes back.

amaltaro:
Sure, go ahead Seangchan.

ticoann:
test this please

PR #8235, opened 2017-10-06
Pylint fixes on PerformanceMonitor

amaltaro:
I wanted to have it in #8204, but jenkins was down and I choose to postpone it.
So, here are only pylint fixes.


amaltaro:
test this please

PR #8241, opened 2017-10-11
fixes missing lumi in acdc wf

### LINKED ISSUE #8220: Unreported missing lumisections
I was looking at the 2 missing lumisection from 
https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=ReReco-Run2016H-FSQJets-07Aug17-0008
and the fact that only the DQMIO is missing something, while all seem to have succeeded on ACDC (there's a second order mystery on why     prozober_ACDC0_Run2016H-v1-FSQJets-07Aug17_8029_170927_144054_8719 has failures, but nothing missing in the output #8219)

Tracking this in the logs, the two lumisections were actually merged in AOD from the output of the original workflow, but cannot be found to be merge in DQMIO. Neither can they be found in the ACDC document of the original workflow (hence not in any ACDC).

The lumisection was processed (see all logs below) 1004fc1c-9eaa-11e7-a550-001e67abf518-512-1-logArchive.tar.gz and reported in the FrameworkJobReport.xml

I could not make sense of the content of stageOut1/Report.pkl to check the output file names in the storage and whatnot, so I cannot tell whether the file was stage-out properly and never used subsequently in a merge job, or if the stage-out failed but did not get reported as missing.

With Ops spending time and effort chasing last lumisections, and things getting delayed, I think it is important to track down, understand how the accounting went astray and fix it.

all logs from eos are in
vlimant@vocms049.cern.ch >ls /tmp/vlimant/fabozzi_Run2016H-v1-FSQJets-07Aug17_8029_170831_183904_6546_ALLLOGS

N.B. dqmio files are reported with 0 events in the FrameworkJobReport
1004fc1c-9eaa-11e7-a550-001e67abf518-512-1-logArchive.tar.gz_dir/cmsRun1/FrameworkJobReport.xml:<Run ID="282663">
1004fc1c-9eaa-11e7-a550-001e67abf518-512-1-logArchive.tar.gz_dir/cmsRun1/FrameworkJobReport.xml-   <LumiSection ID="742" NEvents="0"/>
and that is expected, since these are histogram files.

ticoann:
fixes #8220

### DIFF
--- a/src/python/WMCore/ACDC/DataCollectionService.py
+++ b/src/python/WMCore/ACDC/DataCollectionService.py
@@ -20,26 +20,36 @@ from WMCore.DataStructs.Run import Run
 from WMCore.WMException import WMException
 
 
-def mergeFakeFiles(chunkFiles):
+def mergeSameFiles(chunkFiles):
     """
-    _mergeFakeFiles_
+    _mergeSameFiles_
 
     Receive a list of dicts with acdc files info and merge them together
-    (sum up events) belonging to the same MCFakeFile.
+    (sum up events) belonging to the same File.
     """
-    # do not do anything for real files
-    if not chunkFiles[0]['lfn'].startswith('MCFakeFile'):
-        return chunkFiles
-
-    logging.info("Merging %d ACDC FakeFiles...", len(chunkFiles))
+    originalLen = len(chunkFiles)
+    logging.info("Merging %d ACDC files with the same lfn...", originalLen)
     mergedFiles = {}
     for acdcFile in chunkFiles:
         if acdcFile['lfn'] not in mergedFiles:
             mergedFiles[acdcFile['lfn']] = acdcFile
         else:
-            mergedFiles[acdcFile['lfn']]['events'] += acdcFile['events']
-            mergedFiles[acdcFile['lfn']]['runs'][0]['lumis'].extend(acdcFile['runs'][0]['lumis'])
-    logging.info("resulted in %d final ACDC FakeFiles.", len(mergedFiles))
+            # if it is fake file events add up for real file don't add up.
+            if chunkFiles[0]['lfn'].startswith('MCFakeFile'):
+                mergedFiles[acdcFile['lfn']]['events'] += acdcFile['events']
+
+            for runLumi in acdcFile['runs']:
+                containRunNum = False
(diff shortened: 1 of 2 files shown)


ticoann:

ticoann:
Hi Alan, since you don't trust this patch. :smile:
Here is the test code, (you can run with this with current code and with patch).
https://github.com/ticoann/WmAgentScripts/blob/wmstat_temp_test/test/ACDCTest.py

One thing I am not completely sure is how to handle events (first, last events) also location. 
I think location should be combined unless acdc contains all the locations for that file. Not sure about first, last events.

src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -20,26 +20,36 @@
...
-            mergedFiles[acdcFile['lfn']]['events'] += acdcFile['events']
-            mergedFiles[acdcFile['lfn']]['runs'][0]['lumis'].extend(acdcFile['runs'][0]['lumis'])
-    logging.info("resulted in %d final ACDC FakeFiles.", len(mergedFiles))
+            # if it is fake file events add up for real file don't add up.
+            if chunkFiles[0]['lfn'].startswith('MCFakeFile'):
+                mergedFiles[acdcFile['lfn']]['events'] += acdcFile['events']
+
+            for runLumi in acdcFile['runs']:
+                containRunNum = False
+                for mRunLumi in mergedFiles[acdcFile['lfn']]['runs']:
```
amaltaro: I think `runs` is a set here, right? Can't we simply add new Run objects to this set instead of trying to merge them?

ticoann: Actually here is the list of dict 
it wrapped run object later (below code)

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ACDC/DataCollectionService.py#L260

test/python/WMCore_t/ACDC_t/DataCollectionService_t.py
```diff
@@ -343,7 +343,7 @@ def notestMergeFakeFiles(self):
                                     'run_number': 1}],
```
amaltaro: Seangchan, this test is still not running in jenkins. Can you please update the method name from `notest...` to `test...`?

ticoann:
fixed by #8285

PR #8244, opened 2017-10-13
Temporary workaround to avoid LQ pulling too much work - DO NOT MERGE

amaltaro:
Master branch version not created because this will never get merged (upstream, at least).

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -791,7 +791,11 @@ class WorkQueue(WorkQueueBase):
         if (resources, jobCounts) == (False, False):
             return 0
 
-        self.logger.info("Pull work for sites %s: " % str(resources))
+        # FIXME temporary fix/hack to avoid too many queued jobs in created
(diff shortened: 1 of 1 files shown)


amaltaro:
Ok, I see the numbers I was expecting to in my testbed agent.
So I'm pushing it to the 3 newest agents (250, 251 and 253) and then bumping their thresholds a bit.

amaltaro:
@vlimant FYI

amaltaro:
[Fix #8244 - Temporary workaround to avoid LQ pulling too much work](https://trello.com/c/6U9scSnz/61-fix-8244-temporary-workaround-to-avoid-lq-pulling-too-much-work)

amaltaro:
And I've just noticed we could have used the configurable parameter `QueueDepth`, to accomplish the same thing. Not sure whether it would have any other impact though. 

MAybe we can configure that to 0.5 and remove this patch, but on Monday...

amaltaro:
Patch has been removed from the agents and config change (#8248) was applied (both have the same behaviour).

PR #8246, opened 2017-10-13
Add reqmgr2-ms into setup

vkuznet:


vkuznet:
@ticoann , Seangchan please merge this PR, it is required to build the service. I name it as reqmgr2-ms. I'll also need a new tag to make RPMs.

ticoann:
@vkuznet, Is this the best name :-)? Can we make something without '-'. It seems names with '-' are not stand alone application (which might not be the intention originally). How about just microservice? Or if you have better name without '-'?
@amaltaro, do you have any good name?

vkuznet:
Seangchan,
I stick with convention used in this file, e.g.
wmc-rest
wmc-wmarchive
wmc-base
etc.

The name is only used to build WMCore sub-parts and for nothing else.

But I don't mind to change it to whatever please you :)

ticoann:
Ah I see. but I thought we are doing stand alone micro service with separate spec file? reqmgr2-mc is just used to build the subpart where is the actual application dependency definitions in the file?

vkuznet:
The spec file and deploy/manage scripts are not part of WMCore. I'm preparing
them separately, e.g.
https://github.com/dmwm/deployment/pull/549
https://github.com/cms-sw/cmsdist/pull/3497

These are changes for WMCore setup to build appropriate code which later I used
in spec file:
https://github.com/cms-sw/cmsdist/pull/3497/files#diff-0c5b1fb35497b3a90f154adf4ef6e50cR13

ticoann:
So actually this is stand alone application. Shouldn't the name need to be matched?
Here in the dependency file "reqmgr2-mc" but in spec file says "reqmgr2_mc".

But actually my point was I thought we match the dependency name and the application name.
So there is no application for following things. (no wmc-rest.spec)

wmc-rest
wmc-wmarchive
wmc-base

wmc-wmarchive is exceptional since it's main part of the code is not in WMCore. but only has dependencies.
It looks like reqmgr-mc is solely depends on WMCore. If you have time, can we chat in Skype?

amaltaro:
I'm out of ideas. Maybe we could call it unified2... if so, then I think we need Jean-Rochs blessing first.

vkuznet:
Guys, you're nuts :) this is a name which only to be used to build snapshot of WMCore.
If you can't come up with anything leave as is and let's move on. I need this
change to proceed testing spec/RPMs which is much more time consuming.

ticoann:
"nuts" - you are a genius. How about we name it "nuts" (New Unified Transfer System).  :smile:
Valentin, how about just without "-". (reqmgr2ms) then it could be consistent with others, if you don't have strong objection.

vkuznet:
Done, it is now called reqmgr2ms and I like your sense of humor.

ticoann:
Thanks Valentin,

PR #8247, opened 2017-10-14
Wmstats site fix

### LINKED ISSUE #8240: Wrong site reported in wmstats
I thought it had been fixed, but looking at one case today I see it's not.
More details here:
https://cms-logbook.cern.ch/elog/GlideInWMS/5691

ticoann:
fixes #8240,

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -22,7 +22,7 @@ from WMCore.Credential.Proxy import Proxy
 from WMCore.DAOFactory import DAOFactory
 from WMCore.FwkJobReport.Report import Report
 from WMCore.WMInit import getWMBASE
-from WMCore.WMException import listWMExceptionStr
+from WMCore.Lexicon import getIterMatchObjectOnRegexp, WMEXCEPTION_REGEXP, CONDOR_LOG_FILTER_REGEXP
 
 
 class SimpleCondorPlugin(BasePlugin):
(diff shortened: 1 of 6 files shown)


ticoann:
@amaltaro, Alan, I know it is ugly. so please suggest better way.

ticoann:
test this please

amaltaro:
We right now read the condor.log 3 times. We should read it only once and then make all the regex searches once it's all loaded in memory.
I also believe we could fetch only the last X lines of that file and look up for those regex.

Both the `Reason` and the `MachineAttr...` should be at the very end of the file, if available.

Last but not least :-) for the master version, I'd actually move all those file handle related functions from BasicAlgo to Utils

ticoann:
> We right now read the condor.log 3 times. We should read it only once and then make all the regex searches once it's all loaded in memory.

Yes, you are right. I would prefer not to load whole thing in the memory but not sure how to do in first search case (since it has to store multiple lines).  I would either find way to do this or make regex to run against memory since we store them in first case anyway.

> I also believe we could fetch only the last X lines of that file and look up for those regex.

This is harder to decide which X we have to choose. (We have settings for 50 - 250 lines but some times it wasn't enough)

> Both the Reason and the MachineAttr... should be at the very end of the file, if available.
> 
> Last but not least :-) for the master version, I'd actually move all those file handle related functions from BasicAlgo to Utils

Yes, I think I have to make more generic, 

Thanks Alan

ticoann:
@amaltaro, Alan, could you review again. 
Here are the changes.
1. It won't put everything in the memory (mmap)
2. It will search the multiple regex (with or) at one time.
3. This assumes if there are multiple parts exists for Reason, and MachineAttrGLIDEIN_CMSSite0 the values (text part) will be all the same.

I have a question though, What does "MachineAttrGLIDEIN_CMSSite0" mean exactly? Even it is in HELD state this means job is assigned and sent to there?

ticoann:
test this please

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -293,13 +294,28 @@ def complete(self, jobs):
...
-                                        exitCode = 99400
-                                        exitType = "RemovedByGLIDEIN"
+                                # for condor log, search for the information
+                                for matchObj in getIterMatchObjectOnRegexp(condorFilePath, CONDOR_LOG_FILTER_REGEXP):
+                                    condorReason = matchObj.group("Reason")
+                                    if condorReason:
+                                        logOutput += condorReason
+                                        if "SYSTEM_PERIODIC_REMOVE" in condorReason or "via condor_rm" in condorReason:
+                                            exitCode = 99400
+                                            exitType = "RemovedByGLIDEIN"
```
amaltaro: Seangchan, I think we should break the loop if we find a 99400 exit code.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -293,13 +294,28 @@ def complete(self, jobs):
...
+                                        else:
+                                            exitCode = 99401
+
+                                    siteName = matchObj.group("Site")
```
amaltaro: I think we should search for the complete class ad `MachineAttrGLIDEIN_CMSSite0`, just to avoid issues.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ticoann: I am not sure there will be multiple different reasons. I was expecting to have the same reason if there is a duplicate "Reason". Anyway, I will add break statement

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ticoann: It does. This is just key I added for matching the regular expression.

https://github.com/dmwm/WMCore/pull/8247/files/5bd0793abb2c05de40ab0b9d6e39edcfa1c56ae0#diff-0ed75c44d8f6712727213d921d2f53e4R53

ticoann:
test this please

ticoann:
@amaltaro, Alan test ran OK. Could you review again?

ticoann:
Something is not right on Jenkins unittests

ticoann:
test this please

src/python/WMCore/Lexicon.py
```diff
@@ -656,3 +667,24 @@ def activity(candidate):
...
+    result = re.match(regex, source)
+    if result:
+        return result.group(1).strip()
+    else:
+        return None
+
+
+def getIterMatchObjectOnRegexp(filePath, regexp):
+    with open(filePath, 'r') as f:
+        mm = mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ)
```
amaltaro: Just to make sure I understood this mmap, is it really just a mapping (so memory contains position/address of where the file content is) or it actually loads the data into memory?

src/python/WMCore/Lexicon.py
```diff
@@ -656,3 +667,24 @@ def activity(candidate):
     if candidate in dashboardActivities:
         return True
     raise AssertionError("Invalid dashboard activity: %s should 'test'" % candidate)
+
+
+def getStringsBetween(start, end, source):
```
amaltaro: Are we planning to use this method or it's just a leftover from your initial work?

amaltaro (changes requested):
Looks good in general (still a required change though), we just need to make sure that we read logs from the worker nodes in a consistent way (commented in the code).

src/python/WMCore/Lexicon.py
ticoann: I thought it might be useful in some case to extract some thing between tag. I am not currently using is though. If you prefer I can remove it.

src/python/WMCore/Lexicon.py
```diff
@@ -656,3 +667,24 @@ def activity(candidate):
...
+    regex = r"\%s(.*?)\%s" % (start, end)
+    result = re.match(regex, source)
+    if result:
+        return result.group(1).strip()
+    else:
+        return None
+
+
+def getIterMatchObjectOnRegexp(filePath, regexp):
+    with open(filePath, 'r') as f:
```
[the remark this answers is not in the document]
ticoann: good point. I will change this.

src/python/WMCore/Lexicon.py
ticoann: It is the first case(I guess. :-)). It uses virtual memory. I am not sure how much it load to physical memory at one time (I guess it is depending on operating system)
Here is reference I got for this code.

https://stackoverflow.com/questions/14289421/how-to-use-mmap-in-python-when-the-whole-file-is-too-big

src/python/WMCore/Lexicon.py
amaltaro: I'm not a big fan of writing code that might one day be used (I guess usually they are not). Everyone has a different taste though, and it's small. we can give it a shot :)

amaltaro (approved):
Looks good!

PR #8248, opened 2017-10-16
Use half of the resources for pulling work from GQ

amaltaro:
Replaces #8244

The goal is to avoid creating a large backlog of GQ -> LQ elements (thus created jobs), such that jobs don't wait for too long in the agent queue, but instead get submitted to condor in a short time.

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -129,6 +129,7 @@ config.WorkQueueManager.queueParams["ParentQueueCouchUrl"] = "https://cmsweb.cer
 # this has to be unique for different work queue. This is just place holder
 config.WorkQueueManager.queueParams["QueueURL"] = "http://%s:5984" % (config.Agent.hostName)
 config.WorkQueueManager.queueParams["WorkPerCycle"] = 100  # don't pull more than this number of elements per cycle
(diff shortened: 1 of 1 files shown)


amaltaro:
[Fix #8248 - Use half of the resources for pulling work from GQ](https://trello.com/c/6U9scSnz/61-fix-8248-use-half-of-the-resources-for-pulling-work-from-gq)

amaltaro:
test this please

vlimant:
is that a parameter that goes in the couchdb config ?

amaltaro:
I'm afraid not. I'd rather have only settings that change at least once a year in couch, this one is certainly not one of those.
If we really start playing with it, then we can add it there.

ticoann:
I would even say once a year is not often enough to put in couch config. :smile: As Alan said we would like to put configs in the couch if ops need to change more often than not. I think we also need to add a web interface for view and edit. I will create the issue for that.

amaltaro:
@ticoann I think this one can be merged

ticoann (approved):

vlimant:
I think we should have as many of the agent configurations in couch with a proper interface. tuning the values as we go is the next step of the process, and they will have to accessible in couch.

PR #8250, opened 2017-10-16
Set a configurable fraction of the condor schedd job limit

amaltaro:
Fixes a few bugs inserted in the PR #8228
Further changes are:
* moved a general config attr to the General section (ReqMgr2ServiceURL)
* removed `reqmgr_url` from the wmagent deploy/setup
* removed a backward compatibility code for submitfailed

and mainly:
* given the schedd jobs limit `MAX_JOBS_PER_OWNER` and a configurable `CondorJobsFraction` value, in the ReqMgrAuxDB, proceed with job submission or not. As soon as a single slot gets free, the algo allows one more job to be submitted.

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -203,6 +203,7 @@ config.JobSubmitter.cacheRefreshSize = 30000  # set -1 if cache need to refresh 
 config.JobSubmitter.skipRefreshCount = 20  # (If above the threshold meet, cache will updates every 20 polling cycle) 120 * 20 = 40 minutes
 config.JobSubmitter.submitScript = os.path.join(os.environ["WMCORE_ROOT"], "etc/submit.sh")
 config.JobSubmitter.extraMemoryPerCore = 500  # in MB
+config.JobSubmitter.condorJobsFraction = 0.75  # fraction of the condor schedd limit
 
 config.component_("JobTracker")
 config.JobTracker.namespace = "WMComponent.JobTracker.JobTracker"
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -84,6 +84,7 @@ config.General.workDir = workDirectory
 config.General.logdb_name = logDBName
 config.General.logdb_url = "%s/%s" % (couchURL, config.General.logdb_name)
 config.General.central_logdb_url = "need to get from secrets file"
+config.General.ReqMgr2ServiceURL = "ReqMgr2 rest service"
 
 config.section_("JobStateMachine")
 config.JobStateMachine.couchurl = couchURL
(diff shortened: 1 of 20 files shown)


amaltaro:
Why I no longer see MockReqMgrAuxDB being called, I _simply_ changed the section where the ReqMgr2 attribute is...

amaltaro:
test this please

amaltaro:
Ok, after fixing the ReqMgrAux mock object, all these tests succeed. And it also succeeds getting the `MAX_JOBS_PER_OWNER` from htcondor. So there is likely no need to emulate that function for the htcondor parameters.

@bbockelm would you know how I can update `MAX_JOBS_PER_OWNER` htcondor parameter via python bindings? I tried something like `htcondor.param.update("MAX_JOBS_PER_OWNER", 5000)` but it expects a _Param object (?)

amaltaro:
test this please

etc/WMAgentConfig.py
```diff
@@ -202,6 +203,7 @@
 config.JobSubmitter.skipRefreshCount = 20  # (If above the threshold meet, cache will updates every 20 polling cycle) 120 * 20 = 40 minutes
 config.JobSubmitter.submitScript = os.path.join(os.environ["WMCORE_ROOT"], "etc/submit.sh")
 config.JobSubmitter.extraMemoryPerCore = 500  # in MB
+config.JobSubmitter.condorJobsFraction = 0.75  # fraction of the condor schedd limit
```
amaltaro: Remove it since it's already in the aux db

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -750,23 +747,36 @@ def passSubmitConditions(self):
...
         if maxScheddJobs is None:
-            passCond = False
-        elif int(executingJobs) + self.maxJobsPerPoll >= int(maxScheddJobs):
-            passCond = False
+            logging.warning("Failed to retrieve 'MAX_JOBS_PER_OWNER' from HTCondor")
+            return False
+
+        # fetch idle + running jobs in condor
+        executingJobs = int(self.countWMBSJobsByState.execute("executing"))
+        if executingJobs + self.maxJobsPerPoll >= int(maxScheddJobs * self.condorFraction):
```
amaltaro: @ticoann if we have less jobs than `self.maxJobsPerPoll`, then it skips the submission. For instance, if `MAX_JOBS_PER_OWNER` is set to 500 and `self.maxJobsPerPoll` is set to 1000, the component will never submit any jobs.

I think I'll change the code such that if there is 1 single slot available, then we try submitting that job.
What do you think? The case above isn't that bad because these condor limits are usually high, I'm just concerned that it can bug us in the very long future...

amaltaro:
@ticoann Seangchan, besides a few last-minute bits I just put in, this has been tested and I think it's ready for a review/merge.

amaltaro:
test this please

etc/WMAgentConfig.py
ticoann: Is this already in aux db?

ticoann: Never mind, I was reviewing commit by commit and didn't see the other commit. :-)

ticoann (approved):

ticoann:
test this please

ticoann:
I didn't see the part which removing Processing and Production threshold. Are you going to have different patch for that. (I am just asking since the issue was referenced here)

amaltaro:
Yes, I'd rather have that in a different patch. I created a new issue #8267

amaltaro:
test this please

PR #8260, opened 2017-10-18
Add WorkQueue debugging for ACDC - NEVER merge

amaltaro:
To debug #8220
@ticoann this is the patch I suggest we apply to the agents for a few weeks until we can have a better insight of the problem. Can you have a quick look? I'll also test it in testbed first, just to make sure it isn't broken.

### DIFF
--- a/src/python/WMCore/WMBS/File.py
+++ b/src/python/WMCore/WMBS/File.py
@@ -615,6 +615,17 @@ def addFilesToWMBSInBulk(filesetId, workflowName, files, isDBS = True,
                                 conn = conn,
                                 transaction = transaction)
     if len(runLumiBinds) > 0:
+        ### FIXME temporary debugging  ###
+        logging.info("AMR about to add %i non-unique files for fileset %i",
+                     len(runLumiBinds), filesetId)
+        lumisPerRun = {}
+        for entry in runLumiBinds:
+            if isinstance(entry['runs'], set):
+                for run in entry['runs']:
+                    lumisPerRun.setdefault(run.run, 0)
+                    lumisPerRun[run.run] += len(run.lumis)
+        logging.info("AMR run and # of lumis to be recovered: %s", lumisPerRun)
+        ### done debugging  ###
         setFileRunLumi.execute(file = runLumiBinds,
                                conn = conn,
                                transaction = transaction)
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -412,6 +412,16 @@ class WorkQueue(WorkQueueBase):
                                            acdcInfo['fileset'],
                                            splitedBlockName['Offset'],
                                            splitedBlockName['NumOfFiles'])
+            ### FIXME temporary debugging  ###
+            self.logger.info("AMR processing %d non-unique files for fileset %s",
+                             len(fileLists), acdcInfo['fileset'])
+            lumisPerRun = {}
+            for entry in fileLists:
+                for run in entry['runs']:
+                    lumisPerRun.setdefault(run.run, 0)
+                    lumisPerRun[run.run] += len(run.lumis)
+            self.logger.info("AMR run and # of lumis to be recovered: %s", lumisPerRun)
+            ### done with debugging
 
             block = {}
             block["Files"] = fileLists
--- a/src/python/WMCore/WMBS/MySQL/Files/AddRunLumi.py
+++ b/src/python/WMCore/WMBS/MySQL/Files/AddRunLumi.py
@@ -4,7 +4,7 @@ _AddRunLumi_
 
 MySQL implementation of AddRunLumi
 """
-
+import logging
 from Utils.IteratorTools import grouper
 from WMCore.Database.DBFormatter import DBFormatter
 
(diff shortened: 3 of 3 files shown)


src/python/WMCore/WorkQueue/WorkQueueBackend.py
```diff
@@ -416,7 +416,7 @@ def availableWork(self, thresholds, siteJobCounts, teams=None, wfs=None,
                 siteJobCounts[possibleSite][prio] = siteJobCounts[possibleSite].setdefault(prio, 0) + \
                                                     element['Jobs'] * element.get('blowupFactor', 1.0)
             else:
-                self.logger.info("No available resources for %s with doc id %s", element['RequestName'], element.id)
+                self.logger.debug("No available resources for %s with doc id %s", element['RequestName'], element.id)
```
ticoann: Alan, shouldn't this be separate patch and need to be applied not for the testing?

src/python/WMCore/WMBS/File.py
```diff
@@ -615,6 +615,8 @@ def addFilesToWMBSInBulk(filesetId, workflowName, files, isDBS = True,
                                 conn = conn,
                                 transaction = transaction)
     if len(runLumiBinds) > 0:
+        logging.info("AMR about to add run/lumi for %i files for fileset %i",
+                     len(runLumiBinds), filesetId)
```
ticoann: what is the structure of runLumiBinds? can we can the all the information we need?

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -412,6 +412,16 @@ def _getDBSBlock(self, match, wmspec):
...
                                            splitedBlockName['NumOfFiles'])
+            ### FIXME temporary debugging  ###
+            self.logger.info("AMR processing %d non-unique files for fileset %s",
+                             len(fileLists), acdcInfo['fileset'])
+            lumisPerRun = {}
+            for entry in fileLists:
+                for run in entry['runs']:
+                    lumisPerRun.setdefault(run.run, 0)
+                    lumisPerRun[run.run] += len(run.lumis)
+            self.logger.info("AMR run and # of lumis to be recovered: %s", lumisPerRun)
```
ticoann: do we need information for all the run lumis not by file?

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -412,6 +412,16 @@ def _getDBSBlock(self, match, wmspec):
                                            acdcInfo['fileset'],
                                            splitedBlockName['Offset'],
                                            splitedBlockName['NumOfFiles'])
+            ### FIXME temporary debugging  ###
+            self.logger.info("AMR processing %d non-unique files for fileset %s",
+                             len(fileLists), acdcInfo['fileset'])
```
ticoann: What kind of information we are tracking?

ticoann:
Hi Alan, what you observed was the difference between acdc record and wmbs record, correct?
I am not completely sure about what we can find out with debugging message.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
amaltaro: agreed! I'm making a new patch to apply with this one

src/python/WMCore/WMBS/File.py
amaltaro: it's a list of dictionaries, with 2 keys/values, lfn and Run object.
Ideally we should print everything, the problem is that there are files (like MCFake) with > 100k lumi sections.
Maybe we could print lfn, run and # of lumis inside each run. What do you think?

src/python/WMCore/WorkQueue/WorkQueue.py
amaltaro: It should be enough to know the run and amount of lumis recovered for each of them.

src/python/WMCore/WMBS/File.py
ticoann: So len(runLumiBinds) only prints number of lfns?

> Maybe we could print lfn, run and # of lumis inside each run. What do you think?

Yes I think that will give more information.

src/python/WMCore/WorkQueue/WorkQueue.py
ticoann: OK

amaltaro:
Thanks for the review, Seangchan. I think we need to track 3 things:
1. that we pull all the lumi sections from ACDCServer (thus all lfns/runs/lumis)
2. that we don't miss any of this information until we get to File.addFilesToWMBSInBulk
3. this is a bit more complicated, but we should be able to see what are the binds we use for wmbs insertion. Another problem here is that we have one entry per lumi, and there is no way to know whether we're handling it for ACDC or standard workflows.

Do you see anything missing?

ticoann:
Thanks Alan, That make sense. I was wondering since the missing lumi case examples has the same lfn entries with different lumis. we maybe should focus on that as well.  Printing out what lfn and has more than 1 entries in ACDC.

amaltaro:
Since WorkQueueManager didn't crash in my testbed agent, I'm pushing this one to the 3 new production nodes.

amaltaro:
This patch was already merged for 1.1.8.patch2

ticoann:
@amaltaro, Alan do we still need to keep this around?

amaltaro:
Absolutely not, it's from 2017 :)

PR #8261, opened 2017-10-18
Dont log No available resources in WorkQueue

amaltaro:
Now that the WorkQueueManager polling cycle is much more aggressive, this log causes the ComponentLog to get rotated once a day, if not twice...


amaltaro:
@ticoann I think this one can be merged

ticoann (approved):

PR #8265, opened 2017-10-20
fix the acdc parentage

### LINKED ISSUE #8257: Some files missing parentage
From this thread (well, actually the prep-ops one)
https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/3753.html

which might be coming from this fix pushed into the agents:
#8206

still to be debugged.

ticoann:
fixes #8257

### DIFF
--- a/src/python/WMCore/ACDC/DataCollectionService.py
+++ b/src/python/WMCore/ACDC/DataCollectionService.py
@@ -8,6 +8,8 @@ Copyright (c) 2010 Fermilab. All rights reserved.
 """
 
 import logging
+import threading
+from WMCore.DAOFactory import DAOFactory
 
 import WMCore.ACDC.CollectionTypes as CollectionTypes
 import WMCore.Database.CouchUtils as CouchUtils
@@ -93,10 +95,12 @@ class DataCollectionService(CouchService):
             coll.addFileset(fileset)
             inputFiles = job['input_files']
             for fInfo in inputFiles:
-                if fInfo["merged"] and ("parents" in fInfo) and \
-                   len(fInfo["parents"]) and ("/store/unmerged/" in next(iter(fInfo["parents"]))):
-                    # remove parents files from acdc doucment if they are unmerged files
+                if fInfo["merged"]:
                     fInfo["parents"] = []
+                elif ("parents" in fInfo) and len(fInfo["parents"]) and ("/store/unmerged/" in next(iter(fInfo["parents"]))):
+                    # parents are umerged files and input files are unmerged files - need to find merged ascendant
+                    fInfo["parents"] = list(getMergedParents(fInfo["parents"]))
+                # other case, fInfo["parents"] all or merged parents
             if useMask:
                 fileset.add(files=inputFiles, mask=job['mask'])
             else:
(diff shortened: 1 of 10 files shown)


ticoann:
This is just initial patch which was tested. 
Instead of using the already existing DAO, created a new DAO to handle the bulk query. 
Alan, could you just check the logic. This patch need to be revised but I just want to discuss whether I am in the right direction.

amaltaro:
Again, my mind just doesn't work for this acdc logic :( I'll write all possible use cases that come to my mind and then I should be able to review it.

amaltaro:
Now that I came up with all use cases (sort of regret that, there are so many... :)), the summary is:
* if parents are merged files, we're all good and there is nothing to worry about
* there can't be unmerged parents and in these cases we need to scan the heritage no matter if the input file is
  * merged (Processing job only)
  * or unmerged (Merge or Processing job)

From my large list of use cases, these would be the examples:
25. TaskChain from scratch with 3 tasks, save all tasks: Processing task3 failure, it has a merged input file and merged parent (output of task1)
37. TaskChain input with 4 tasks, save only task4: Processing task3 failure, unmerged input file and merged parent (has to be input dataset)
43. TaskChain MC with 4 tasks, save only task4: Processing task4 failure, unmerged input file and NO parent

So, we never know whether a job has to have a parent or not. We need to scan the file heritage.
Do you agree? Let me know if you want the whole list of use cases (that came to my mind)

ticoann:
Hi Alan, We can simplified use cases. slightly different from yours.

1. if the input file itself is merged file, we are good. - We don't even record parentage. (because input file will be new ACDC job output files parent and it should be in dbs)
2. if input file is unmerged file we need to go through its heritage until we find the merged file. (It there is no merged accendant, leave the parent files as empty list (no parents).

I think this should cover all the cases.

amaltaro:
Haaah, much better indeed. Yes, I think that covers all use cases.
I still have concerns about StepChain, but they are probably not related to this issue.

src/python/WMCore/WMBS/MySQL/Files/GetParentInfo.py
```diff
@@ -15,22 +13,17 @@
 
 class GetParentInfo(DBFormatter):
     sql = """SELECT wfp.id, wfp.lfn, wfp.merged,
```
amaltaro: last comma has to go away

src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -350,3 +353,32 @@ def getLumilistWhitelist(self, collectionID, taskName):
...
+    myThread = threading.currentThread()
+    daoFactory = DAOFactory(package="WMCore.WMBS", logger=myThread.logger,
+                                dbinterface=myThread.dbi)
+
+    getParentInfoAction = daoFactory(classname="Files.GetParentInfo")
+
+    parentsInfo = getParentInfoAction.execute(childLFNs)
+    newParents = set()
+    unmergedParents = set()
+    for parentInfo in parentsInfo:
```
amaltaro: Shouldn't we keep looking for parent and grand parents until one of them no longer exists (None)?
Eventually we might have taskChains with way too many tasks that could cause a lack of grandgrandgrand parentage information.

amaltaro (changes requested):

src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -93,10 +93,13 @@ def failedJobs(self, failedJobs, useMask=True):
             coll.addFileset(fileset)
             inputFiles = job['input_files']
             for fInfo in inputFiles:
-                if fInfo["merged"] and ("parents" in fInfo) and \
-                   len(fInfo["parents"]) and ("/store/unmerged/" in next(iter(fInfo["parents"]))):
-                    # remove parents files from acdc doucment if they are unmerged files
-                    fInfo["parents"] = []
+                if ("parents" in fInfo) and len(fInfo["parents"]) and ("/store/unmerged/" in next(iter(fInfo["parents"]))):
```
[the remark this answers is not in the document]
ticoann: updated the code, However, DAO can't be changed since this is used in JobAccountant as well. I am not sure how heavy the load is it just get parent file lfn.  Anyway, we need to first check what is the performance penalty on this.

src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -350,3 +353,32 @@ def getLumilistWhitelist(self, collectionID, taskName):
 
         lumiList = LumiList(compactList=self.getLumiWhitelist(collectionID, taskName))
         return lumiList
+
+
+def getMergedParents(childLFNs):
+    import threading
+    from WMCore.DAOFactory import DAOFactory
```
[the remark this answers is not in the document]
ticoann: Once it is imported it doesn't really load multiple times. I think it just check the dictionary where module is already loaded. Anyway, I was just thinking this can be just used as library.
But I moved to the top now.

src/python/WMCore/ACDC/DataCollectionService.py
ticoann: It does. it will look for first merged parents. If it gets unmerged parents it will keep look until it finds merged parents or no parents.

ticoann:
test this please

amaltaro:
test this please

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

PR #8268, opened 2017-10-23
Script to migrate input dataset and create a StoreResults request

amaltaro:
This script relies on a properly structured input from the user. It then makes a DBS migration request (asynchronous operation that might take a few minutes) and a StoreResults workflow is created.

Giving it a try in the relval team:
 amaltaro_EmbeddingRun2016B-MuTauFinalState-imputSep16DoubleMu_mirror_miniAOD-_171023_120525_5544

Seangchan, I think we should keep it under WMCore until we change how StoreResults are created (by the conveners, with proper sitedb roles, etc).

### DIFF
--- a/bin/createStoreResults.py
+++ b/bin/createStoreResults.py
@@ -1,0 +1,157 @@
+"""
+For more details about StoreResults workflows, please refer to:
+https://github.com/dmwm/WMCore/wiki/StoreResults-requests
+
+Pre-requisites:
+ 1. a valid proxy in your X509_USER_PROXY variable
+ 2. wmagent env: source /data/srv/wmagent/current/apps/wmagent/etc/profile.d/init.sh
+ 3. have the correct permissions in SiteDB, otherwise dataset migration won't work
+
+Expected input json file like:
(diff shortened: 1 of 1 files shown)


amaltaro:
BTW, in order to migrate datasets between different DBS instances, one needs to have one of these sitedb roles:
{'dbs': 'operator', 'dataops': 'production-operator'}

amaltaro:
FYI @thongonary @vlimant @mcremone

vlimant:
BTW, if we can move from dbs instances, why not stick to this and do anything further ? 
what is the actual purpose of storeresult in this case ? merge and /store/something ?

amaltaro:
The main purposes are:
1. make files bigger (merge files)
2. inject this dataset into phedex, otherwise no transfers can be made.

ericvaandering:
Exactly, @amaltaro . But out of curiosity, didn't we have a script like this already? Maybe a broken one?

amaltaro:
Hmm, maybe it was the reqmgr client scripts (reqmgr.py or reqmgr2.py), which takes a json template as input (the same we use for daily tests). Indeed we could have added the other static fields inside one of those clients and use them (the only part missing would be the dataset migration).

Issue #8284, opened 2017-10-30
Don't pull work from GQ if schedd is at its limit


amaltaro:
Specific issue related to #8129 
This issue has to be fixed together #8267 

Using htcondor param `MAX_JOBS_PER_OWNER`. Same as here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L771

amaltaro:
Fixed by #8288

Issue #8267, opened 2017-10-23
Remove Production/Processing thresholds on JobSubmitter


amaltaro:
Specific issue related to #8129

As discussed in previous discussions with @vlimant , we plan to get rid of the Production/Processing task thresholds and use instead the schedd limitation `MAX_JOBS_PER_OWNER`.

Initial developments were done https://github.com/dmwm/WMCore/pull/8228 and https://github.com/dmwm/WMCore/pull/8250

@ticoann didn't we have to change something somewhere else too?

amaltaro:
Fixed by #8288 , though we followed a slightly different/better idea, which is keeping the thresholds but over-submitting jobs that have a higher priority than anything else in the agent (20% of the thresholds only though).

PR #8288, opened 2017-10-30
submit higher priority jobs regardless the threshold

### LINKED ISSUE #8129: Job submitter bottleneck
Let me put this in a GH issue as I think this is going (was) to be impacting protection.
With large number of low priority (63k prio) pending on all agents, higher priority work WQE does get acquired, the jobs are created (regardless of thresholds according to @amaltaro ) but then jobs do not get submitted over threshold.
Example an agent with 40k job pending threshold could be holding 35k low priority job and de-facto is reduced to a 5k effective threshold for all the rest. That is, the low priority pending have no chance at running (because of higher priority work being ran otherwise from other agents) and will not free their pending slot, leaving very little room for other work that would deserve to get submitted and ran.

we should find a way to solve this

ticoann:
fixes #8129, #8284
The simplest way I can think of to solve this issue. (There are short comings though)

1. It will still submit higher priority workflow but only the highest one first.
   i.e. current highest priority is 5 and there are jobs in 7,8,9 in the wmbs queue. It will submit priority 9 only then once 9 is done, 8, etc
2. Not all the jobs with 9 will be submitted. only 20% of current slot. (This can be increased) - maybe will make configurable.

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
@@ -1,0 +1,24 @@
+from __future__ import print_function, division
+import logging
+from WMCore.DAOFactory import DAOFactory
+from WMCore.Services.PyCondor.PyCondorAPI import getScheddParamValue
+
+
+def availableScheddSlots(dbi, logger=logging, condorFraction=1):
+    """
+    check executing jobs and compare with condor limit.
+    return the difference -- executing jobs - (condor limit * condorFraction)
+    """
+    action = DAOFactory(package='WMCore.WMBS',
+                        logger=logger,
+                        dbinterface=dbi)(classname="Jobs.GetCountByState")
+    executingJobs = int(action.execute("executing"))
+
+    maxScheddJobs = getScheddParamValue("MAX_JOBS_PER_OWNER")
+
+    if maxScheddJobs is None:
+        logger.warning("Failed to retrieve 'MAX_JOBS_PER_OWNER' from HTCondor")
+        return 0
+
+    freeSubmitSlots = int(int(maxScheddJobs) * condorFraction - executingJobs)
+    return freeSubmitSlots
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -7,7 +7,7 @@ _JobSubmitterPoller_t_
 
 Submit jobs for execution.
 """
-
+from __future__ import print_function, division
 import logging
 import threading
 import os.path
@@ -30,8 +30,18 @@ from WMCore.FwkJobReport.Report               import Report
 from WMCore.WMException                       import WMException
 from WMCore.BossAir.BossAirAPI                import BossAirAPI
 from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
-from WMCore.Services.PyCondor.PyCondorAPI import getScheddParamValue
 from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
+
+from WMComponent.JobSubmitter.JobSubmitAPI import availableScheddSlots
+
+
+def jobSubmitCondition(jobStats):
+
+    for jobInfo in jobStats:
+        if jobInfo["Current"] >= jobInfo["Threshold"]:
+            return jobInfo["Condition"]
+
+    return "JobSubmitReady"
 
 
 class JobSubmitterPollerException(WMException):
@@ -73,7 +83,8 @@ class JobSubmitterPoller(BaseWorkerThread):
         self.packageSize = getattr(self.config.JobSubmitter, 'packageSize', 500)
         self.collSize = getattr(self.config.JobSubmitter, 'collectionSize', self.packageSize * 1000)
         self.maxTaskPriority = getattr(self.config.BossAir, 'maxTaskPriority', 1e7)
-        self.condorFraction = None  # update during every algorithm cycle
+        self.condorFraction = 0.75  # update during every algorithm cycle
+        self.condorOverflowFraction = 0.2
 
         # Additions for caching-based JobSubmitter
         self.cachedJobIDs = set()
@@ -102,9 +113,7 @@ class JobSubmitterPoller(BaseWorkerThread):
             logging.debug("Config: %s", config)
             raise JobSubmitterPollerException(msg)
 
-
         # Now the DAOs
-        self.countWMBSJobsByState = self.daoFactory(classname="Jobs.GetCountByState")
         self.listJobsAction = self.daoFactory(classname="Jobs.ListForSubmitter")
         self.setLocationAction = self.daoFactory(classname="Jobs.SetLocation")
         self.locationAction = self.daoFactory(classname="Locations.GetSiteInfo")
@@ -513,6 +522,69 @@ class JobSubmitterPoller(BaseWorkerThread):
 
         return
 
+    def _getJobSubmitCondition(self, jobPrio, siteName, jobType):
+        """
+        returns the string describing whether a job is ready to be submitted or the reason can't be submitted
+        Only jobs with "JobSubmitReady" return value will be added to submit job.
+        Other return values will indicate the reason jobs cannot be submitted.
+        i.e. "NoPendingSlot"  - pending slot is full with pending job
+        """
+        try:
+            totalPendingSlots = self.currentRcThresholds[siteName]["total_pending_slots"]
+            totalPendingJobs = self.currentRcThresholds[siteName]["total_pending_jobs"]
+            totalRunningSlots = self.currentRcThresholds[siteName]["total_running_slots"]
+            totalRunningJobs = self.currentRcThresholds[siteName]["total_running_jobs"]
+            highestPriorityInJobs = self.currentRcThresholds[siteName]['wf_highest_priority']
+
+            taskPendingSlots = self.currentRcThresholds[siteName]['thresholds'][jobType]["pending_slots"]
+            taskPendingJobs = self.currentRcThresholds[siteName]['thresholds'][jobType]["task_pending_jobs"]
+            taskRunningSlots = self.currentRcThresholds[siteName]['thresholds'][jobType]["max_slots"]
+            taskRunningJobs = self.currentRcThresholds[siteName]['thresholds'][jobType]["task_running_jobs"]
+
+            # set the initial totalPendingJobs since it increases in every cycle when a job is submitted
+            self.currentRcThresholds[siteName].setdefault("init_total_pending_jobs", totalPendingJobs)
+
+            # set the initial taskPendingJobs since it increases in every cycle when a job is submitted
+            self.currentRcThresholds[siteName]['thresholds'][jobType].setdefault("init_task_pending_jobs", taskPendingJobs)
+
+            initTotalPending = self.currentRcThresholds[siteName]["init_total_pending_jobs"]
+            initTaskPending = self.currentRcThresholds[siteName]['thresholds'][jobType]["init_task_pending_jobs"]
+
+        except KeyError as ex:
+            msg = "Invalid key for site %s and job type %s\n" % (siteName, jobType)
+            logging.exception(msg)
+            return "NoJobType_%s_%s" % (siteName, jobType)
+
+        if highestPriorityInJobs is None or jobPrio <= highestPriorityInJobs:
+            # there is no pending or running jobs in the system (None case) or
+            # priority of the job is lower or equal don't allow overflow
+            totalPendingThreshold = totalPendingSlots
+            taskPendingThreshold = taskPendingSlots
+            totalJobThreshold = totalPendingSlots + totalRunningSlots
+            totalTaskTheshold = taskPendingSlots + taskRunningSlots
+        else:
+            # In case the priority of the job is higher than any of currently pending or running jobs.
+            # Then increase the threshold by condorOverflowFraction * original pending slot.
+            totalPendingThreshold = max(totalPendingSlots, initTotalPending) + (
+                                    totalPendingSlots * self.condorOverflowFraction)
+            taskPendingThreshold = max(taskPendingSlots, initTaskPending) + (
+                                    taskPendingSlots * self.condorOverflowFraction)
+            totalJobThreshold = totalPendingThreshold + totalRunningSlots
+            totalTaskTheshold = taskPendingThreshold + taskRunningSlots
+
+        jobStats = [{"Condition": "NoPendingSlot",
+                     "Current": totalPendingJobs,
+                     "Threshold": totalPendingThreshold},
+                    {"Condition": "NoTaskPendingSlot",
+                     "Current": taskPendingJobs,
+                     "Threshold": taskPendingThreshold},
+                    {"Condition": "NoRunningSlot",
+                     "Current": totalPendingJobs + totalRunningJobs,
+                     "Threshold": totalJobThreshold},
+                    {"Condition": "NoTaskRunningSlot",
+                     "Current": taskPendingJobs + taskRunningJobs,
+                     "Threshold": totalTaskTheshold}]
+        return jobSubmitCondition(jobStats)
 
     def assignJobLocations(self):
         """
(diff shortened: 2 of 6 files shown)


ticoann:
@amaltaro, Alan please review. still need to combine Production and Processing priority but might do it with separate patch.

amaltaro:
Seangchan, did you add this new key `wf_highest_priority` just for the cases where the JobSubmitter cache is stale?

JobSubmitter already makes sure to submit the highest priority jobs first:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L539

and what we discussed was to simply remove the Production/Processing thresholds, not to over-submit based on the workflow priority (and respecting a factor that will never make everyone happy :)).

I think we can keep it simpler by implementing what is in https://github.com/dmwm/WMCore/issues/8267 , so, if the submitter passes the restrictions, we allow job submission. Of course, we still want to control the IO jobs.

ticoann:
Hi Alan, I must misunderstood. I thought removing Production/Processing is in addition to this change. The problem I see is that we can always have the same problem not submitting the jobs in JobSubmitter if we don't by pass the threshold limit for the higher priority. We can chat if you have a time.

ticoann:
@amaltaro, Alan, I ran the few tests it seems it is working as intended but code ends up a little complicate than I was hoping for. Please review.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -34,6 +35,15 @@
 from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
 
 
+def jobSubmitCondition(jobStats):
+
+    for jobInfo in jobStats:
+        if jobInfo["Current"] >= jobInfo["Threshold"]:
+            return jobInfo["Condition"]
+
+    return "JobSumitReady"
```
amaltaro: typo

amaltaro: Maybe also add a docstring saying that this method decides whether a job can be submitted or not.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +524,65 @@ def getThresholds(self):
 
         return
 
+    def _getJobSubmitCondition(self, jobPrio, siteName, jobType):
+        """
+        returns the string whehter job is read to submit or the reason can't be submitted
```
amaltaro: typo

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +524,65 @@ def getThresholds(self):
...
+
+        if highestPriorityInJobs is None or jobPrio <= highestPriorityInJobs:
+            # there is no pending or running jobs in the system (None case) or
+            # priority of the job is lower or equal don't allow overflow
+            totalPendingThreshold = totalPendingSlots
+            taskPendingThreshold = taskPendingSlots
+            totalJobThreshold = totalPendingSlots + totalRunningSlots
+            totalTaskTheshold = taskPendingSlots + taskRunningSlots
+        else:
+            # priorty of the job is higher than any off currently pending or running.
```
amaltaro: more typos

amaltaro (approved):
Logic looks good, Seangchan. It took me a while to digest all the details (evil jobStats...), but it seems to be doing what it's supposed to do.
Several aesthetic fixes though.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +524,65 @@ def getThresholds(self):
...
+            # increase the threshold by condorOverflowFraction * origianl pending slot.
+            pendingTotalOverflow = max(totalPendingSlots, initTotalPending) + (
```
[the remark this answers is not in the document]
ticoann: the reason I did max(totalPendingSlots, initTotalPending) to handle when 
totalPendingSlots >  initTotalPending. Lets say totalPendingSlots = 1000 and initTotalPending is 100 (which is current pending jobs in the system). We don't want to just submit just 20 percent in that case we want to submit up to the totalPendingSlots. 
I shouldn't add 20 percent extra in that case, but code will be more complicated and I think it is OK to overflow.

ticoann:
Thanks Alan, I will fix all the typos.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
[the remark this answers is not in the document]
amaltaro: From L543, isn't initTotalPending == totalPendingSlots ? Well, at least for the first time we iterate over a specific site. Actually, to me it seems it's always the same, since we don't do anything else with the pending slots (+1 or -1 etc) inside the submitter algo.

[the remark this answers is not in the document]
ticoann: Yes you're right. You found a bug. I was meant to put initiTotalPedning as initial totalPendingJobs not totalPendingSlots. Thanks I will update the patch.

ticoann:
Alan could you review again. I fixed what you pointed.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +524,74 @@ def getThresholds(self):
...
+            totalRunningSlots = self.currentRcThresholds[siteName]["total_running_slots"]
+            totalRunningJobs = self.currentRcThresholds[siteName]["total_running_jobs"]
+            highestPriorityInJobs = self.currentRcThresholds[siteName]['wf_highest_priority']
+
+            taskPendingSlots = self.currentRcThresholds[siteName]['thresholds'][jobType]["pending_slots"]
+            taskPendingJobs = self.currentRcThresholds[siteName]['thresholds'][jobType]["task_pending_jobs"]
+            taskRunningSlots = self.currentRcThresholds[siteName]['thresholds'][jobType]["max_slots"]
+            taskRunningJobs = self.currentRcThresholds[siteName]['thresholds'][jobType]["task_running_jobs"]
+
+            totalPendingJobs = self.currentRcThresholds[siteName]["total_pending_jobs"]
```
amaltaro: I failed to see any difference :-) Maybe next time make a new commit only to make it clear what changed (and merge after you get the green light).

Anyways, isn't this line L546 == L536? And L550 == L542?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +524,74 @@ def getThresholds(self):
...
+            taskPendingJobs = self.currentRcThresholds[siteName]['thresholds'][jobType]["task_pending_jobs"]
+            # set the initial taskPendingJobs since it increases in every cycle when a job is submitted
+            self.currentRcThresholds[siteName]['thresholds'][jobType].setdefault("init_task_pending_jobs", taskPendingJobs)
+
+            initTotalPending = self.currentRcThresholds[siteName]["init_total_pending_jobs"]
+            initTaskPending = self.currentRcThresholds[siteName]['thresholds'][jobType]["init_task_pending_jobs"]
+
+        except KeyError as ex:
+            msg = "Invalid key for site %s and job type %s\n" % (siteName, jobType)
+            msg += traceback.format_exc()
```
amaltaro: not sure if you forgot to replace it by logging.exception or just didn't want to. Ping'ing you again just to be sure.

amaltaro (commented):
The max thing is still not clear to me though

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: Of course I forgot. :-)

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: Ah I need to change my glasses.  The difference was before it was total_pending_slot. Now it is total_pending_jobs

ticoann:
About the max, I am still confused too. :-)
Here are the all the use cases we have handle.
1. We have to submit jobs regardless number of pending jobs with certain fraction of pending slots if the priority of jobs higher than pending jobs. 
2. We have to submit jobs up to pending slots if current pending jobs are fewer than pending slot regardless in fraction defined.

so without max(pending_job, pending_slot)
If we choose pending_slot (without max). it won't get to submit the jobs if pending jobs are more than 20 % of pending slots.

If we choose pending job (without max), it submits not enough jobs if the pending jobs are much smaller than the pending slots. (only 20% of pending slot) - Of course, next cycle it will put more jobs but we have to wait until next cycle.

amaltaro:
test this please

ticoann:
test this please

ticoann:
@amaltaro, Alan I added fix for #8284 as well. Please review.

ticoann:
@amaltaro, Alan commit is correctly updated now. Could you review?

src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import print_function, division
+import logging
+from WMCore.DAOFactory import DAOFactory
+from WMCore.Services.PyCondor.PyCondorAPI import getScheddParamValue
+
+
+def numOfJobsSubmitUptoScheddLimit(dbi, logger=logging, condorFraction=1):
+    """
+    check executiing jobs and campare with condor limit.
```
amaltaro: typos :)

src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
```diff
@@ -0,0 +1,24 @@
...
+    return the differece executing jobs - (condor limit * condorFraction)
+    """
+    action = DAOFactory(package='WMBS',
```
amaltaro: Isn't it WMCore.WMBS instead?

src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
```diff
@@ -0,0 +1,24 @@
...
+                        logger=logger,
+                        dbinterface=dbi)(classname="Jobs.GetCountByState")
+    executingJobs = int(action.execute("executing"))
+
+    maxScheddJobs = int(getScheddParamValue("MAX_JOBS_PER_OWNER"))
```
amaltaro: I wonder whether it really returns None in case it fails to fetch that parameter...
Anyways, my point is that this casting to int will crash if the returned value is None (my mistake btw :)).
How about we cast after the if below or in the L23?

src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import print_function, division
+import logging
+from WMCore.DAOFactory import DAOFactory
+from WMCore.Services.PyCondor.PyCondorAPI import getScheddParamValue
+
+
+def numOfJobsSubmitUptoScheddLimit(dbi, logger=logging, condorFraction=1):
```
amaltaro: Minor, but.... how about renaming it to something like 'availableScheddSlots'?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +525,71 @@ def getThresholds(self):
...
+            # there is no pending or running jobs in the system (None case) or
+            # priority of the job is lower or equal don't allow overflow
+            totalPendingThreshold = totalPendingSlots
+            taskPendingThreshold = taskPendingSlots
+            totalJobThreshold = totalPendingSlots + totalRunningSlots
+            totalTaskTheshold = taskPendingSlots + taskRunningSlots
+        else:
+            # In case priority of the job is higher than any of currently pending or running.jobs
+            # increase the threshold by condorOverflowFraction * origianl pending slot.
+            pendingTotalOverflow = max(totalPendingSlots, initTotalPending) + (
```
amaltaro: And now I see your point with this `max`. This is needed for cases where higher prio jobs come on top of other higher prio and so on. Still, considering 20% of the original pending slots threshold. You were right ;)
PS.: typos in the comments above.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +525,71 @@ def getThresholds(self):
...
+                                    totalPendingSlots * self.condorOverflowFraction)
+            pendingTaskOverflow = max(taskPendingSlots, initTaskPending) + (
```
amaltaro: just to minimize these assignments (that twists my brain), how about assigning this expression directly to `taskPendingThreshold`?
Same for the line above, assigning directly to `totalPendingThreshold` instead of `pendingTotalOverflow`.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -513,6 +525,71 @@ def getThresholds(self):
...
+        jobStats = [{"Condition": "NoPendingSlot",
+                     "Current": totalPendingJobs,
+                     "Threshold": totalPendingThreshold},
+                    {"Condition": "NoTaskPendingSlot",
+                     "Current": taskPendingJobs,
+                     "Threshold": taskPendingThreshold},
+                    {"Condition": "NoRunningSlot",
+                     "Current": totalPendingJobs + totalRunningJobs,
+                     "Threshold": totalJobThreshold},
+                    {"Condition": "NoTaskSlot",
```
amaltaro: Just to keep it consistent, how about renaming it from `NoTaskSlot` to `NoTaskRunningSlot`?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -728,6 +778,8 @@ def algorithm(self, parameters=None):
             if self.useReqMgrForCompletionCheck:
                 # only runs when reqmgr is used (not Tier0)
                 self.removeAbortedForceCompletedWorkflowFromCache()
+                agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
+                self.condorOverflowFraction = agentConfig.get("CondorOverflowFraction", 0.2)
```
amaltaro: Since we already get the aux db config in L815. How about we move this 
`self.condorOverflowFraction = agentConfig.get("CondorOverflowFraction", 0.2)`
to L817?

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
```diff
@@ -61,11 +63,16 @@ def passRetrieveCondition(self):
         False if the component should skip pulling work this cycle.
         
         For now, it only checks whether the agent is in drain mode or
-        if the condor schedd is overloaded.
+        if the condor schedd is overloaded or
+        MAX_JOBS_PER_OWNER is reached or if the condor schedd is overloaded.
```
amaltaro: dup msg

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
```diff
@@ -61,11 +63,16 @@ def passRetrieveCondition(self):
...
         """
+
         passCond = True
+        myThread = threading.currentThread()
         if isDrainMode(self.config):
```
amaltaro: Just to avoid us checking all these constraints, how about we add one info (or warning) log per condition? Then we can remove this log line I think:
https://github.com/ticoann/WMCore/blob/a2f3245d72b595e040bcbe2760ef69c9ba55e903/src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py#L94

amaltaro (approved):
Looks much better now, Seangchan!
I still have some suggestions though that will make the code cleaner and more readable, besides that, it's good to go (provided you did some simple tests :))

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: Actually L815 need to be moved here 781 otherwise Tier0 agent will crash.

ticoann:
Thank  you very much Alan for through review. I changed everything (I hope) as you suggested except one thing. I rearrange the aux db call so Tier0 Agent won't crash, also Tier0 is not checking the same condition.  Could you review again?

ticoann:
test this please

ticoann:
@amaltaro, Alan let me know 4th commit is OK. especially for Tier0 values

ticoann:
@hufnagel, Dirk, FYI, this change will effect Tier0 Agent (in minimal way)

https://github.com/dmwm/WMCore/pull/8288/files#diff-63fe4938fe7d571728e9e8e9d3e0a1afR770

Basically it won't submit the job when schedd limit is reached. MAX_JOBS_PER_OWNER
Is this OK?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -586,7 +581,7 @@ def _getJobSubmitCondition(self, jobPrio, siteName, jobType):
                     {"Condition": "NoRunningSlot",
                      "Current": totalPendingJobs + totalRunningJobs,
                      "Threshold": totalJobThreshold},
-                    {"Condition": "NoTaskSlot",
+                    {"Condition": "NoTaskRunnigSlot",
```
amaltaro: typo :)

src/python/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py
```diff
@@ -86,13 +85,13 @@ def pullWork(self):
         myThread = threading.currentThread()
 
         try:
-            if self.passRetrieveCondition():
+            cond = self.passRetrieveCondition()
+            if cond == "OK":
```
amaltaro: Yes, I like it!

src/python/WMComponent/JobSubmitter/JobSubmitAPI.py
```diff
@@ -4,21 +4,21 @@
 from WMCore.Services.PyCondor.PyCondorAPI import getScheddParamValue
 
 
-def numOfJobsSubmitUptoScheddLimit(dbi, logger=logging, condorFraction=1):
+def availableScheddSlots(dbi, logger=logging, condorFraction=1):
     """
-    check executiing jobs and campare with condor limit.
+    check executiing jobs and compare with condor limit.
```
amaltaro: and since you'll have to fix the typo below, how about "executiing" and "differece"

amaltaro (changes requested):
It looks like we are competing to see who makes more commits and get more comments on our PRs :)

Seangchan, code looks good, you just need to fix a typo in a dict key (and why not a couple more in the comments..) and we're good.

hufnagel:
What is the default for MAX_JOBS_PER_OWNER ? 100k ? If so there shouldn't be an issue for the Tier0.

amaltaro:
yes, 100k

ticoann:
test this please

ticoann:
test this please

PR #8290, opened 2017-10-31
Make sure job splitting parameters are positive

### LINKED ISSUE #8289: avoid infinite loop in EventBased splitting
see https://github.com/dmwm/CRABServer/issues/5599
this loop in EventBased.py will go into infinte loop if eventsPerJob is not strictly positive.
We will put a check in CRAB client, but it may be a good idea to protect the code here as well.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/EventBased.py#L148-L183

amaltaro:
Fixes #8289

StdBase already validates these inputs and WMAgent should be safe.
However, adding another protection at the job splitting layer is not going to hurt and it will protect any other services importing these somehow generic job splitting algorithms

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -63,6 +63,10 @@ class EventAwareLumiBased(JobFactory):
 
         eventsPerLumiInDataset = 0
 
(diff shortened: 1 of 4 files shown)


amaltaro:
test this please

amaltaro:
test this please

ticoann (approved):

PR #8291, opened 2017-10-31
add config to control higher priority workflow

ticoann:
Alan, it the name ok?

### DIFF
--- a/src/python/WMCore/Agent/DefaultConfig.py
+++ b/src/python/WMCore/Agent/DefaultConfig.py
@@ -11,6 +11,8 @@ DEFAULT_AGENT_CONFIG = {
     "IgnoreDisks": ["/lustre/unmerged"],
     # fraction of condor schedd limit, used for job submission
     "CondorJobsFraction": 0.75,
+    # In JobSumitter, submit jobs over the threshold if the priority is higher than current pending/running jobs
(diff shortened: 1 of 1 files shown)


src/python/WMCore/Agent/DefaultConfig.py
```diff
@@ -12,5 +12,7 @@
     # fraction of condor schedd limit, used for job submission
     "CondorJobsFraction": 0.75,
     # ExitCodes returned by jobs which doesn't need to be retried
-    "NoRetryExitCodes": [70, 73, 8001, 8006, 8009, 8023, 8026, 50660, 50661, 50664, 71102]
+    "NoRetryExitCodes": [70, 73, 8001, 8006, 8009, 8023, 8026, 50660, 50661, 50664, 71102],
+    # In JobSumitter, with this percentage, submit jobs over the threshold if the priority is higher than current pending/running jobs
+    "HigherPriorityOverflowPercent": 20  # default 20% over threshold
```
amaltaro: Sounds good, but how about we just call it `CondorOverflowFraction`, such that it's consistent with the other condor parameter (CondorJobsFraction)? 

In addition to that, we provide the fraction from 0 to 1 instead of the percentage. Or you could transform the previous one into percentage :-)

ticoann:
Thanks Alan,  I update the parameter.

PR #8057, opened 2017-08-15
fix fwjr error string

ticoann:
fixes #8056

PR #8295, opened 2017-11-02
Drop bytes that cannot be encoded with utf-8

### LINKED ISSUE #8056: JobAccountant crashing with undecoded bytes
Seeing repeat crashes on the JobAccountant on cmssrv237 -- attached (one hopes) is an exerpt of the log and the offending pkl file -- 

[JobAccountant.err.txt](https://github.com/dmwm/WMCore/files/1223271/JobAccountant.err.txt)
[Report.0.pkl.txt](https://github.com/dmwm/WMCore/files/1223273/Report.0.pkl.txt)

Tried removing that file to see if it moved on -- but crashed again...

amaltaro:
Fixes #8056
Replaces #8057

I think `findMagicStr` method is the reason why we started seeing these problems again, since it does not drop bytes that it cannot encode.

### DIFF
--- a/src/python/WMCore/JobStateMachine/SummaryDB.py
+++ b/src/python/WMCore/JobStateMachine/SummaryDB.py
@@ -71,7 +71,8 @@ def fwjr_parser(doc):
     pdict = dict(totalJobCPU=0, totalJobTime=0, totalEventCPU=0)
     ddict = {}
     for key, val in steps.items():
-        wrappedTotalJobTime += val['stop'] - val['start']
+        if val['stop'] is not None and val['start'] is not None:
+            wrappedTotalJobTime += val['stop'] - val['start']
         site_name = val['site']
         site_summary = dict(wrappedTotalJobTime=wrappedTotalJobTime,
                             inputEvents=0, dataset=ddict,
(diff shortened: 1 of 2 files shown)


amaltaro:
This one is going to be hard to reproduce, but I ran a few workflows with this patch in and everything looks Ok.

src/python/WMCore/JobStateMachine/SummaryDB.py
```diff
@@ -71,7 +71,8 @@ def fwjr_parser(doc):
     pdict = dict(totalJobCPU=0, totalJobTime=0, totalEventCPU=0)
     ddict = {}
     for key, val in steps.items():
-        wrappedTotalJobTime += val['stop'] - val['start']
+        if val['stop'] is not None and val['start'] is not None:
```
ticoann: shouldn't this be 'or' not 'and'?

amaltaro: I just copied your code from PR 8057 :-) But I think it's and, since we want to make sure both start/stop values are not null, before we subtract them.

ticoann: Sorry, I misread. I thought "is None or is None". My bad.

PR #8304, opened 2017-11-06
initial patch for worker node data update

### LINKED ISSUE #8279: Add worker node hostname in fwjr for WMArchive.
In meta_data field.
meta_data["wn"] or meta_data["worker_node"]

ticoann:
fixes #8279 
Still need to do
1. update correct value
2. agree on default value
3. agree on property name and format

### DIFF
--- a/src/python/WMCore/WMRuntime/Bootstrap.py
+++ b/src/python/WMCore/WMRuntime/Bootstrap.py
@@ -224,6 +224,21 @@ def createInitialReport(job, logLocation):
     report.data.siteName = getattr(siteCfg, 'siteName', 'Unknown')
     report.data.hostName = socket.gethostname()
     report.data.ceName = getSyncCE()
+
+
+    # TODO: need to check what format it returns and what features need to extract.
+    # currently
+    # $MACHINEFEATURES/hs06: HS06 score of the host
+    # $MACHINEFEATURES/total_cpu: number of configured job slots
+    # $JOBFEATURES/hs06_job: HS06 score available to your job
+    # $JOBFEATURES/allocated_cpu: number of allocated slots (=8 in case of a multicore job
+    report.data.machineFeatures = {}
+    report.data.machineFeatures['HS06_SCORE_HOST'] = os.environ.get('MACHINEFEATURES/hs06')
+    report.data.machineFeatures['TOTAL_CPU'] = os.environ.get('MACHINEFEATURES/total_cpu')
+    report.data.jobFeatures = {}
+    report.data.jobFeatures['HS06_SCORE_JOB'] = os.environ.get('JOBFEATURES/hs06_job')
+    report.data.jobFeatures['ALLOCATED_CPU'] = os.environ.get('JOBFEATURES/allocated_cpu')
+
     report.data.completed = False
     report.setTaskName(taskName=job.get('task', 'TaskNotFound'))
 
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -223,6 +223,7 @@ class Report(object):
         Create a JSON version of the Report.
         """
         jsonReport = {}
+        jsonReport["WorkerNodeInfo"] = self.getWorkerNodeInfo()
         jsonReport["task"] = self.getTaskName()
         jsonReport["steps"] = {}
         jsonReport["skippedFiles"] = self.getAllSkippedFiles()
@@ -1529,3 +1530,10 @@ class Report(object):
                     delattr(source.files, "file%d" % fileNum)
                 source.files.fileCount = 0
         return
+
+    def getWorkerNodeInfo(self):
+        wnInfo = {"HostName": self.data.hostName,
+                  "MachineFeatures": self.data.machineFeatures,
+                  "JobFeatures": self.data.jobFeatures}
+
+        return wnInfo
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -419,6 +419,7 @@ def createArchiverDoc(job, version=None):
     # append meta data in fwjr
     newfwjr['meta_data'] = {'agent_ver': version,
                          'host': socket.gethostname().lower(),
+                         'wn_info': fwjr["WorkerNodeInfo"],
                          'fwjr_id': job_id,
                          'jobtype': jobtype,
                          'jobstate': jobstate,


src/python/WMCore/WMRuntime/Bootstrap.py
```diff
@@ -224,6 +224,16 @@ def createInitialReport(job, logLocation):
...
     report.data.ceName = getSyncCE()
+
+
+    # TODO: need to check what format it returns and what features need to extract.
+    # currently
+    # $MACHINEFEATURES/hs06: HS06 score of the host
+    # $MACHINEFEATURES/total_cpu: number of configured job slots
+    # $JOBFEATURES/hs06_job: HS06 score available to your job
+    # $JOBFEATURES/allocated_cpu: number of allocated slots (=8 in case of a multicore job
+    report.data.machineFeatures = os.environ.get('MACHINEFEATURES', {})
```
vkuznet: here I think we need to scan all environment variables ($MACHINEFEATURES/hs06, $MACHINEFEATURES/total_cpu, etc). and then define our own dict which will provide key-value pairs.

vkuznet (commented):
We need an example from a concrete node to see what $MACHINEFEATURES variables refer to.

ticoann:
test this please

ticoann:
It gets the host name but not Job features and machine features. Valentin, @vkuznet  Valentin, do you know which site this env value is set?

amaltaro:
Seangchan, I think we should have left this one out of the current WMAgent release. The last jenkins test was done in the beginning of November and we have no idea how the job and machine features data looks like.
We should probably agree on the data format and exactly which data we need to fetch and publish to WMArchive.

aperezca:
Hi, yes, MJF is defined in, at least, this document:

http://hepsoftwarefoundation.org/notes/HSF-TN-2016-02.pdf

As a test site, you can use PIC, for example. Any job should have access to these files and variables, if you want to check.

amaltaro:
Does PIC allow ssh_to_job (global or itb pool?)? Such that I can play with these metrics and gather all info we need?

aperezca:
I'm not sure ssh_to_job works right now, but looking at one job currently running, I can provide the following info:

[root@td707 jobfeatures-43250153.pbs04.pic.es]# pwd
/tmp/mjf-cmpilot033/jobfeatures-43250153.pbs04.pic.es
[root@td707 jobfeatures-43250153.pbs04.pic.es]# ls -l
total 28
-rw-r--r-- 1 cmpilot033 cmpilot  1 Feb 18 22:23 allocated_cpu
-rw-r--r-- 1 cmpilot033 cmpilot  5 Feb 18 22:23 hs06_job
-rw-r--r-- 1 cmpilot033 cmpilot 21 Feb 18 22:23 job_id
-rw-r--r-- 1 cmpilot033 cmpilot 10 Feb 18 22:23 jobstart_secs
-rw-r--r-- 1 cmpilot033 cmpilot 11 Feb 18 22:23 max_swap_bytes
-rw-r--r-- 1 cmpilot033 cmpilot 12 Feb 18 22:23 scratch_limit_bytes
-rw-r--r-- 1 cmpilot033 cmpilot  6 Feb 18 22:23 wall_limit_secs

[root@td707 jobfeatures-43250153.pbs04.pic.es]# for i in `ls`; do echo $i; cat $i; done
allocated_cpu 8 
hs06_job 83.80 
job_id 43250153.pbs04.pic.es 
jobstart_secs 1518988993 
max_swap_bytes 33625731072 
scratch_limit_bytes 973222244352 
wall_limit_secs 385200 

Each variable is stored in a file named after the variable, which has the variable value as content.

amaltaro:
Thanks, Antonio!
I managed to connect to a T2_DE_DESY worker node (batch0670.desy.de) and this is the information I extracted.
```
[cmsplt003@batch0670 /var/lib/condor/execute/dir_5653/glide_RaR3wG/execute/dir_16494]$ echo $JOBFEATURES

[cmsplt003@batch0670 /var/lib/condor/execute/dir_5653/glide_RaR3wG/execute/dir_16494]$
[cmsplt003@batch0670 /var/lib/condor/execute/dir_5653/glide_RaR3wG/execute/dir_16494]$ echo $MACHINEFEATURES
/etc/batch/machinefeatures/
[cmsplt003@batch0670 /var/lib/condor/execute/dir_5653/glide_RaR3wG/execute/dir_16494]$ ll $MACHINEFEATURES
total 12
-rw-r--r-- 1 root root 3 Dec 11 14:14 hs06
-rw-r--r-- 1 root root 4 Sep 21  2016 hsW
-rw-r--r-- 1 root root 2 Sep 21  2016 total_cpu
[cmsplt003@batch0670 /var/lib/condor/execute/dir_5653/glide_RaR3wG/execute/dir_16494]$ for i in `ls $MACHINEFEATURES`; do echo $i; cat $MACHINEFEATURES/$i; echo ""; done
hs06
369
hsW
1.48
total_cpu
32
```

where one can see the worker node does not provide JOBFEATURES var/dir

amaltaro:
Would you know the precise htcondor classads? BTW, didn't I have you added in my gtalk? If you can find me there, it's probably easier to talk there then we can update a summary here

aperezca:
Not sure how you'd access $JOBFEATURES directly as the user. As job = pilot in this context, wouldn't you need to do ssh_to_job from the pilot factory, which is the schedd for the glidein, rather than from the payload schedd, to access the proper environment?

amaltaro:
I checked a job that ran at CERN and I saw both environment variables, though I couldn't list their content because I cannot connect to CERN worker nodes.

I created a new issue to fix these information and properly feed the monitoring tools. For now, this patch will always have an empty information (empty string?) for Job and Machine features.
I also ran a couple of tests just to make sure it was not going to break somewhere else.

vkuznet:
@ticoann, Seangchan it seems like you already resolved this issue without me.
But I don't really know which site to look at. I saw that Alan already provided
some info.

ticoann:
@vkuznet, Valentin, I discussed with Alan about the data structure. 
Is it OK to have worker node info in metadata top level and separate metrics for worker node (job/machine features).
For now we can just push host info and we can push job/machine info when we know more about what is need to be posted. (I will create the separate patch). Would that be OK?

vkuznet:
Yes, I think all of these metrics should be at top level since they remain
constant for entire document which represent workflow/steps running on a node.

bbockelm:
Hi folks,

Do we ever expect this to be useful for more than 20% of jobs?  Certainly the adoption in the US is zero and will always be that way for non-WLCG resources.

Seems to be a more widely applicable to pull the information from the HTCondor machine ad. SI always has some benchmark info there and also takes care of interacting with the batch environment.  We should let them handle this, rather than both teams parsing the info.

Brian

ticoann:
Sorry I wasn't aware there was other effort propagating this info. My understanding was this info was needed in WMArchive. Certainly, we don't have to duplicate the effort.

vkuznet:
Brian, Seangchan, I'll not close the case, since we should clarify what
end-users need. It's most likely that some information is present in one source
but not another and we need to have a common key to match the info in
two sources. If I recall correctly that was the case with DESY people
request, but we better clarify it with Manuel Giffels. Here what he asked me
originally in a series of emails (I only extract relevant parts):

I have one additional question concerning the data, because a computer science student recently started to work on a workload balancing simulation. The idea is to find the optimal balance between I/O-intense tasks and CPU-intense tasks running at GridKa. Therefore, we would be interested in performance data of WMAgent jobs.
...
We are interested in anything stored under the 'performance‘ key inside your FWJR Schema. In addition, we like to get also information about the WnHostName, JobType (production, processing, etc.) and the task (e.g. mcremone_ACDC0_task_TOP-RunIIFall17wmLHEGS-00056__v1_T_180122_144320_4391).
...
Do you have any hint on how one could match the jobs inside job monitoring with the fwjr’s in wmarchive?

Best,
Valentin.

aperezca:
Wait, let's recap; what information is needed and what is the intended source? Can you guys check what's already being collected by the pilots and propagated to job classads? 
Then we may consider if we need anything that's only provided via de Machine & Job Features mechanism, as the deployment of is is certainly scarce (because it's not requested by CMS).
If it's indeed needed, then we have to make mandatory for CMS sites, deploy, etc.

bbockelm:
Hi,

Indeed - I think this is already in the job ClassAds (and has been for many months).  Let's bring it up with Manuel maybe?

In addition, the SI pilots include DB12-based benchmarking far more uniformly than HS06 is deployed.  It's less accurate than HS06, but might be a reasonable tradeoff they want to consider.

Brian

PR #8313, opened 2017-11-09
Extract error

ticoann:
@amaltaro, Alan please review, it will extract this specific format of error and display in wmstats

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -314,6 +314,10 @@ class SimpleCondorPlugin(BasePlugin):
                                     if errMsg:
                                         logOutput += "\n\n%s\n" % errMsg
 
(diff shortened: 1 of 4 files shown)


src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -314,6 +314,10 @@ def complete(self, jobs):
                                     if errMsg:
                                         logOutput += "\n\n%s\n" % errMsg
 
+                                    errMsg = matchObj.group('ERROR')
```
amaltaro: Isn't it going to fetch part of the WMException message?

ticoann: I change the reg expression.
https://github.com/dmwm/WMCore/pull/8313/files#diff-0ed75c44d8f6712727213d921d2f53e4R50

So it will look for WMExeption or ERROR. 
This ERROR seems to be not part of WMException it fails before it create WMException (mostly syntax error)

PR #8280, opened 2017-10-27
Set PNN instead of PSNs for wmbs file locations

amaltaro:
Fix #8258 
First commit has some improvements/unused code changes, which I applied while debugging that issue.
Second commit changes all wmbs file setters to set the PNN instead of the location (PSN mapped from PNN).

I'm trying not to change any logic in the code, but only change what we store in the database and adapt getters/setters.

@ticoann and @hufnagel can you please check whether I'm going at the right direction

Code is completely untested btw.

PR #8314, opened 2017-11-09
Remove T0 legacy setup and option from resource-control insertion

amaltaro:
This dirty hack has to be removed once #8280 goes in
@ticoann please review

### DIFF
--- a/bin/wmagent-resource-control
+++ b/bin/wmagent-resource-control
@@ -50,8 +50,6 @@ def createOptionParser():
     myOptParser.add_option("-p", "--thresholds", dest = "printThresholds",
                            default = False, action = "store_true",
                            help = "Print out all known thresholds and site information.")
-    myOptParser.add_option("--add-T0", dest = "addT0", default = False, action = "store_true",
-                           help = "Add the CMS Tier-0 site to resource control.")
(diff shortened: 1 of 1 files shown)


ticoann:
This is just for the new agent not for the master,  right? I didn't know you were talking about this change (only in bin). I guess we don't need to patch the production agent just back port it, right?
Also we don't need to bother to have pylint change if we are only applying this to new agent.

amaltaro:
Yes, though I created another patch in #8315
I'll change this PR and simply remove a legacy code.

ticoann:
That's good, Thanks Alan.

amaltaro:
Patch was updated. I basically removed the T0 code that was legacy (still storage element?!?) and removed the option too, since one can use `--add-one-site`

amaltaro:
@hufnagel I doubt it can cause any harm to T0, but just in case ...

ticoann (approved):

ticoann:
test this please

hufnagel:
No problem, we haven't used that option in ages.

ericvaandering:
This commit seems to have broken Jenkins by causing two tests to time out. Please investigate it and fix it. Look at #8325 for the test names.

PR #8317, opened 2017-11-10
need to increase at least hold 7200

ticoann:


amaltaro:
Seangchan, was it breaking anything? these 6 digits can hold up to 999999 secs, I doubt any component cycle would take more than this time :)

ticoann:
Hi Alan, yes it was breaking some. How would I spot the problem otherwise? 😄 
It seems like 6 is combined number. I mean DECIMAL(6,4) only accepts 99.1234 but not 100.1234.
I thought the same as you too.

PR #8319, opened 2017-11-10
Temporary fix for sites with multi-pnns

ticoann:


amaltaro:
I don't get why it fails to run this script during the agent deployment
```
Executing wmagent-resource-control --add-T1s --plugin=SimpleCondorPlugin --pending-slots=50 --running-slots=50 --down ...
./manage: line 821: /data/srv/wmagent/v1.1.8.patch1/sw.sryu/slc7_amd64_gcc630/cms/wmagent/1.1.8.patch1/bin/wmagent-resource-control: Permission denied
```

it's not executable on the node, while it is in my local repo(?)

PR #8325, opened 2017-11-13
Make sure NANOAOD files go straight to UNmerged

### LINKED ISSUE #8278: Ban NANOAOD{SIM} files from going straight to merge
Otherwise we could end up with some EDM and non-EDM files under the same dataset. The idea is to never have NANOADO EDM format registered in DBS.

This hook can be added either with:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Templates/StageOut.py#L51

or checking the datatier here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py#L151

@hufnagel you use these mechanism in T0, let me know if there is a better place for adding this tweak.

amaltaro:
Fixes #8278

Still have to test though

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/StageOut.py
@@ -149,18 +149,20 @@ class StageOut(Executor):
                 #  - are we over the size threshold
                 #  - are we over the event threshold ?
                 straightToMerge = False
-                if not getattr(fileName, 'merged', False) and hasattr(self.step.output, 'minMergeSize'):
-                    if fileName.module_label not in getattr(self.step.output, 'forceUnmergedOutputs', []):
-                        if getattr(fileName, 'size', 0) >= self.step.output.minMergeSize:
+                if not getattr(fileName, 'merged', False):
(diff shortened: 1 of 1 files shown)


hufnagel:
Might be wrong, but don't merge jobs run through the same code path ? Ie. merge jobs are just processing jobs whose output is always straightToMerge ? Or is that done differently ? Haven't looked at this in a while...

amaltaro:
test this please

amaltaro:
I'm not sure... if that's the case, then it means straightToMerge is always False

amaltaro:
@ericvaandering it looks like jenkins don't want to run for this PR. Would you mind having a look at it once you have some spare cycles?

ericvaandering:
Looks like two tests might be timing out:

#487 _testInsertAllPNNs_
#489 _testInsertT0_ 

(from two different slices) both in test/python/WMCore_t/ResourceControl_t/ResourceControl_t.py

amaltaro:
Now it works :-) from the condor.err log (should be .out, but ...)
`INFO:root:NANOAOD and NANOAODSIM files never go straight to merge!`

amaltaro:
test this please

amaltaro:
@ticoann @hufnagel  it has been backported already to 1.1.8, but please review even if it's already merged :)

hufnagel:
Logic looks alright.

PR #8338, opened 2017-11-16
change to debug statement for parent file

ticoann:
This generates a lot of log message. @amaltaro, if you agree I will create agent branch patch as well.

PR #8344, opened 2017-11-17
Skip MCFakeFile parentage

### LINKED ISSUE #8337: Do not put MCFakeFile as parent file when ACDC is created.
Is it necessary to MCFakeFile as parents since it won't be added in DBS anyway.

amaltaro:
Fixes #8337
Also contains changes from #8338

@ticoann I think we'll need to have this into the new agent and also backported to the current production agents. I'm going to run a quick ACDC test, just to check whether it doesn't break anything.

### DIFF
--- a/src/python/WMCore/ACDC/DataCollectionService.py
+++ b/src/python/WMCore/ACDC/DataCollectionService.py
@@ -95,12 +95,16 @@ class DataCollectionService(CouchService):
             coll.addFileset(fileset)
             inputFiles = job['input_files']
             for fInfo in inputFiles:
-                if fInfo["merged"]:
+                if int(fInfo["merged"]) == 1:  # Looks like Oracle and MySQL return diff type
                     fInfo["parents"] = []
-                elif ("parents" in fInfo) and len(fInfo["parents"]) and ("/store/unmerged/" in next(iter(fInfo["parents"]))):
-                    # parents are umerged files and input files are unmerged files - need to find merged ascendant
-                    fInfo["parents"] = list(getMergedParents(fInfo["parents"]))
-                # other case, fInfo["parents"] all or merged parents
+                elif fInfo.get("parents", []):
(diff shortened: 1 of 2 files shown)


ticoann:
patch looks good to me. But we also need the patch fore ErrorHandling part so we wan't put the MCFakefile initially in the ACDC

amaltaro:
Good point Seangchan. I updated the patch. I'm also not sure whether we could have a third case where parents would not be set at all. Can you review it again please?

amaltaro:
If everything is fine, I'll patch vocms0253 some point tomorrow.

ticoann:
Thanks Alan patch looks good to me.

ticoann:
I am just wondering if this workflow failed so many jobs in the first step why would ops create ACDC instead of cloning?

amaltaro:
It wasn't that bad, it had 7k failures against 100k success jobs. Now, don't ask me where that 47k offset comes from :)

amaltaro:
The component took 10h to go through those wmbs operations. I suggest we push these changes in in the new agent only (and properly validate it as well).

ticoann:
If that agent went through this, yes I agree we put this on only new agents.

amaltaro:
Trying to figure why parents are no longer put in the acdc fileset docs... I can't see any issues with this patch nor with the one already in

src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -97,10 +97,14 @@ def failedJobs(self, failedJobs, useMask=True):
             for fInfo in inputFiles:
                 if fInfo["merged"]:
```
amaltaro: Haaa... it was this merged value that is a string, either '0' or '1', so it evaluates always to True :(

amaltaro:
@ticoann it has been backported already to 1.1.8, but please review even if it's already merged :)

PR #8349, opened 2017-11-20
Overwrite default WMAgent config couch doc

### LINKED ISSUE #8350: WMAgent couch config is not getting updated
I might be wrong, but when we try to commit a couch document with an already existent doc id, it doesn't add nor update the document. I noticed this problem while trying to upload the wmagent default configuration (the one we run during the agent deployment):
`$manage execute-agent wmagent-upload-config`

amaltaro:
Fixes #8350

Changes are:
* the script run during the deployment, will first try to remove the agent couch doc, then it uploads a new one
* if one uses the post method to upload a document that already exists, nothing happens
* if one uses the put method, then it updates the couch doc properties (like a dict.update)

### DIFF
--- a/bin/wmagent-upload-config
+++ b/bin/wmagent-upload-config
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 from __future__ import division, print_function
+
 import os
 from pprint import pformat
+
 from WMCore.Agent.DefaultConfig import DEFAULT_AGENT_CONFIG
 from WMCore.Configuration import loadConfigurationFile
 from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
--- a/bin/wmagent-upload-config
+++ b/bin/wmagent-upload-config
@@ -1,17 +1,24 @@
 #!/usr/bin/env python
 from __future__ import division, print_function
 import os
+from pprint import pformat
 from WMCore.Agent.DefaultConfig import DEFAULT_AGENT_CONFIG
 from WMCore.Configuration import loadConfigurationFile
 from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
 
-def insertWMAgentConfig(reqmgrURL, hostName, agentConfig):
-    reqMgrAux = ReqMgrAux(reqmgrURL)
-    reqMgrAux.postWMAgentConfig(wmConfig.Agent.hostName, agentConfig)
+
+def insertWMAgentConfig(reqMgrAux, agentName, agentConfig):
+    print("Uploading agent configuration to ReqMgrAux. Data is\n%s" % pformat(agentConfig))
+    reqMgrAux.postWMAgentConfig(agentName, agentConfig)
+
+
+def removeWMAgentConfig(reqMgrAux, agentName):
+    reqMgrAux.deleteWMAgentConfig(agentName)
+
 
 if __name__ == "__main__":
     wmConfig = loadConfigurationFile(os.environ['WMAGENT_CONFIG'])
-    
-    insertWMAgentConfig(wmConfig.General.ReqMgr2ServiceURL,
-                        wmConfig.Agent.hostName, DEFAULT_AGENT_CONFIG)
-    
+    reqMgrAux = ReqMgrAux(wmConfig.General.ReqMgr2ServiceURL)
+
+    removeWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName)
+    insertWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName, DEFAULT_AGENT_CONFIG)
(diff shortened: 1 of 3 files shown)


ticoann:
Actually, I was thinking for the post call, only insert when document doesn't exist instead of updating. It thought it might be safer to do that way. I would explicitly delete when we want to replace the whole document. I can add delete RESTApi and update the deployment script if that is OK with you.  But if you think that is not a safer, this patch looks fine.

amaltaro:
With this patch, post and put method will have the same behaviour/outcome. I think your suggestion would be better indeed, but we need to delete/create the document only for the WMAgent config.
Let me have a look at it...

ticoann:
There are still difference for post and put in here as well. Post will replace whole document, put will just add/update only specified the properties.  I am just thinking by separating post and delete we might make less mistake or we can even give different permission on that later. If you agreed, I can create the patch. Don't spend your valuable time on this. 😄

amaltaro:
If you commit a couch dob with the _rev field/value, it replaces the previous doc then? How about the revision, does it go to -2, -3, etc? Or it remains in -1?

Have a look at my 2nd commit, unless I completely misunderstood you, it should be what we were discussing. Still have to be tested though :)

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -99,7 +99,7 @@ def getWMAgentConfig(self, agentName):
         return agentConfig[0]
 
     def postWMAgentConfig(self, agentName, agentConfig):
-
+        self["requests"].delete('wmagentconfig/%s' % agentName)
```
ticoann: Oh I meant we will have separate delete method and only used in deployment script.  That way if anyone use postAgentConfig assuming there is no config for that, won't be accidentally overwriting the config.

amaltaro: Wise Seangchan!!!

ticoann:
> If you commit a couch dob with the _rev field/value, it replaces the previous doc then? How about the revision, does it go to -2, -3, etc? Or it remains in -1?

It replaces the previous document and revision number increase.

amaltaro:
I updated the 2nd commit again. Hopefully it addresses everything now. Still untested.

bin/wmagent-upload-config
```diff
@@ -1,17 +1,24 @@
...
 
 if __name__ == "__main__":
     wmConfig = loadConfigurationFile(os.environ['WMAGENT_CONFIG'])
-    
-    insertWMAgentConfig(wmConfig.General.ReqMgr2ServiceURL,
-                        wmConfig.Agent.hostName, DEFAULT_AGENT_CONFIG)
-    
+    reqMgrAux = ReqMgrAux(wmConfig.General.ReqMgr2ServiceURL)
+
+    removeWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName)
```
ticoann: What happens when remove fails (if document doesn't exists) will it through error or move on. (Of course, we should print message and move on)

ticoann:
Thanks Alan, patch looks good to me. I have only one comments inline. 😄

bin/wmagent-upload-config
amaltaro: I wanted to ask you this question but I forgot. 
Should we raise an exception as we do here: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L492
?

ticoann: I don't think we should raise error here. but catch error and print log and move on. (Since some cases - a brand new agent won't have document initially.

amaltaro: It will log the error and move on. I'm actually not sure whether we really need the `CouchError` exception class.

amaltaro:
@ticoann I did a few tests and everything looks Ok. Deleting a document that doesn't exist will generate the following log in reqmgr2:
```
[21/Nov/2017:15:13:13]  ERROR: failed to delete document: WMAGENT_CONFIG_vocms0262.cern.ch
Reason: CouchNotFoundError - reason: Object Not Found, data: {} result: None
[21/Nov/2017:15:13:13] alancc7-cloud1.cern.ch 188.184.88.70 "DELETE /reqmgr2/data/wmagentconfig/vocms0262.cern.ch HTTP/1.1" 200 OK [data: 3803 in 20 out 46940 us ] [auth: OK "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms040.cern.ch" "" ] [ref: "" "WMCore.Services.Requests/v001" ]
```

I've also merged both commits now.

ticoann (approved):

ticoann:
Thanks Alan,

PR #8351, opened 2017-11-20
Overwrite default WMAgent config couch doc - wmagent version

amaltaro:
wmagent branch patch of #8349

Slightly different approach such that it doesn't depend on ReqMgr/cmsweb changes.

### DIFF
--- a/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
+++ b/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
@@ -100,7 +100,7 @@ class ReqMgrAux(Service):
 
     def postWMAgentConfig(self, agentName, agentConfig):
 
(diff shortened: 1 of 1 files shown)


amaltaro:
tested and it works fine.

amaltaro:
[Fix #8351 - Overwrite default WMAgent config couch doc](https://trello.com/c/Uke9Toru/69-fix-8351-overwrite-default-wmagent-config-couch-doc)

ticoann:
Is it because agent gets redeployed it doesn't update the config? 
I think we just need to deleted that config. (Maybe manually for now but we can add in deploy script for deletion later - I will do that).

amaltaro:
That's right. If it's a new agent and there is no such document, then there are no problems.
So, deleting the document before uploading the new configuration, will have the same effect.

ticoann (approved):

PR #8354, opened 2017-11-20
kill all the jobs when force-complete, aborted

### LINKED ISSUE #8252: WMBSHelper creating work for force-completed workflows
As reported here:
https://its.cern.ch/jira/browse/CMSCOMPPR-1555

this workflow pdmvserv_task_JME-RunIIFall17GS-00005__v1_T_171013_161511_6259 was force-completed and even so a bunch of jobs were getting submitted and their numbers growing.

It only happened to vocms0253 (250 and 251 handled it properly).

Transitions from reqmgr2 were (notice -2h compared to agents local time):
```
force-complete: 2017-10-16 14:28:46 UTC
completed: 2017-10-16 15:03:10 UTC
```

which was properly seem by WorkQueueManager a minute later:
```
2017-10-16 16:29:45,871:140716091242240:INFO:WorkQueue:Canceling work for workflow(s): set([u'pdmvserv_task_JME-RunIIFall17GS-00005__v1_T_171013_161511_6259'])

2017-10-16 16:29:51,356:140716091242240:INFO:WMBSHelper:Killing 31461 jobs for workflow: pdmvserv_task_JME-RunIIFall17GS-00005__v1_T_171013_161511_6259
```

but several minutes later there were still a few LQE (out of a few hundred, I think)
```
2017-10-16 16:33:12,755:140716116420352:INFO:WorkQueueBackend:No available resources for pdmvserv_task_JME-RunIIFall17GS-00005__v1_T_171013_161511_6259 with doc id fdf6b0f5d5b8af8264c47a1373bd4ea3
...
```

which later on were acquired by the agent
```
2017-10-16 17:05:13,896:140716108027648:INFO:WMBSHelper:"pdmvserv_task_JME-RunIIFall17GS-00005__v1_T_171013_161511_6259" Injecting production 1:320001:350080001 - 1:321000:351174000 (run:lumi:event) into wmbs
```

and, that's the reason we saw so many jobs still getting submitted.

And I don't get why we see this log 5min after WMBSHelper was killing those (log above)
```
2017-10-16 16:35:08,510:140716091242240:INFO:SimpleCondorPlugin:Going to remove all the jobs for workflow pdmvserv_task_JME-RunIIFall17GS-00005__v1_T_171013_161511_6259
```

I think what triggered this problem was that not all LQE were canceled (not from the local/workqueue db at least), which were later on acquired by WMBSHelper.

ticoann:
fixes #8252

### DIFF
--- a/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
+++ b/src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
@@ -23,10 +23,10 @@ histogramLimit: Limit in terms of number of standard deviations from the
   average at which you cut the histogram off.  All points outside of that
   go into overflow and underflow.
 """
-__all__ = []
 import logging
 import threading
 import traceback
+
 from Utils.Timers import timeFunction
 from WMComponent.TaskArchiver.DataCache import DataCache
 from WMCore.DAOFactory import DAOFactory
@@ -77,7 +77,7 @@ class TaskArchiverPoller(BaseWorkerThread):
         self.config = config
         self.jobCacheDir = self.config.JobCreator.jobCacheDir
 
-        if getattr(self.config.TaskArchiver, "useWorkQueue", False) != False:
+        if not getattr(self.config.TaskArchiver, "useWorkQueue", False):
             # Get workqueue setup from config unless overridden
             if hasattr(self.config.TaskArchiver, 'WorkQueueParams'):
                 self.workQueue = localQueue(**self.config.TaskArchiver.WorkQueueParams)
@@ -192,7 +192,7 @@ class TaskArchiverPoller(BaseWorkerThread):
         if isinstance(statusList, basestring):
             statusList = [statusList]
         reqNames = self.centralCouchDBWriter.getRequestByStatus(statusList)
-        logging.info("There are %d requests in 'aborted' status in central couch.", len(reqNames))
+        logging.info("There are %d requests in %s status in central couch.", len(reqNames), statusList)
         for wf in reqNames:
             self.workQueue.killWMBSWorkflow(wf)
         return reqNames
@@ -214,8 +214,7 @@ class TaskArchiverPoller(BaseWorkerThread):
 
         centralCouchAlive = True
         try:
-            abortedWorkflows = self.killCondorJobsByWFStatus(["aborted"])
-            self.killCondorJobsByWFStatus(["force-complete"])
+            self.killCondorJobsByWFStatus(["force-complete", "aborted"])
         except Exception as ex:
             centralCouchAlive = False
             logging.error("we will try again when remote couch server comes back\n%s", str(ex))
@@ -224,7 +223,7 @@ class TaskArchiverPoller(BaseWorkerThread):
             for workflow in finishedwfs:
                 try:
                     # Notify the WorkQueue, if there is one
-                    if self.workQueue != None:
+                    if self.workQueue is not None:
                         subList = []
                         logging.info("Marking subscriptions as Done ...")
                         for l in finishedwfs[workflow]["workflows"].values():
--- a/src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
+++ b/src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
@@ -2,15 +2,15 @@
 """
 Perform cleanup actions
 """
-__all__ = []
-
-
-
 import time
 import random
+import threading
+
 from Utils.Timers import timeFunction
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
+from WMCore.DAOFactory import DAOFactory
+
 
 class WorkQueueManagerCleaner(BaseWorkerThread):
     """
@@ -24,8 +24,11 @@ class WorkQueueManagerCleaner(BaseWorkerThread):
         self.queue = queue
         self.config = config
         self.reqmgr2Svc = ReqMgr(self.config.General.ReqMgr2ServiceURL)
-        # state lists which shouldn't be populated in wmbs. (To prevent creating work before WQE status updated)
-        self.abortedAndForceCompleteWorkflowCache = self.reqmgr2Svc.getAbortedAndForceCompleteRequestsFromMemoryCache(expire=120)
+        myThread = threading.currentThread()
+        daoFactory = DAOFactory(package="WMCore.WMBS",
+                                logger=myThread.logger,
+                                dbinterface=myThread.dbi)
+        self.finishedWorflowCheck = daoFactory(classname="Subscriptions.CountFinishedSubscriptionsByWorkflow")
 
     def setup(self, parameters):
         """
@@ -47,9 +50,14 @@ class WorkQueueManagerCleaner(BaseWorkerThread):
             # this will clean up whatever left over from above clean up.
             # also if the wq replication has problem it won't delay the killing jobs in condor
             # and updating wmbs status
-            abortedAndForceCompleteRequests = self.abortedAndForceCompleteWorkflowCache.getData()
+            # state lists which shouldn't be populated in wmbs. (To prevent creating work before WQE status updated)
+            # added completed status in the list due to the race condition
+            requests = self.reqmgr2Svc.getRequestByStatusFromMemoryCache(["aborted", "aborted-completed", "force-complete", "completed"]).getData()
+            results = self.finishedWorflowCheck.execute(workflowNames=requests)
 
-            for wf in abortedAndForceCompleteRequests:
+            requestsToKill = [reqInfo["workflow"] for reqInfo in results if reqInfo["open"] > 0]
+
+            for wf in requestsToKill:
                 self.queue.killWMBSWorkflow(wf)
 
         except Exception as ex:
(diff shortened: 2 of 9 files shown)


ticoann:
@amaltaro, This is not most efficient solution but seems to be simplest. Still there is chance for race condition for closed-out, announced -archived. But it should be very unlikely.

The extra overhead would be ~1 sec per request when request is completed with all killed job. I am not sure whether that is two much or not current production reqmgr contains ~200 completed requests. If this is too much overhead, we can create the new query just checking all the requests are completed by just checking the subscription for available and acquired subscription.

ticoann:
test this please

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -214,8 +214,7 @@ def completeTasks(self, finishedwfs):
 
         centralCouchAlive = True
         try:
-            abortedWorkflows = self.killCondorJobsByWFStatus(["aborted"])
-            self.killCondorJobsByWFStatus(["force-complete"])
+            self.killCondorJobsByWFStatus(["force-complete", "aborted"])
```
amaltaro: Shouldn't we add aborted-completed too? There is still the possibility of having jobs for a workflow in completed though.
BTW, this will kill any job types for those workflows, right? In theory we should let LogCollect and Cleanup jobs running.

src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
```diff
@@ -47,9 +45,11 @@ def algorithm(self, parameters):
             # this will clean up whatever left over from above clean up.
             # also if the wq replication has problem it won't delay the killing jobs in condor
             # and updating wmbs status
-            abortedAndForceCompleteRequests = self.abortedAndForceCompleteWorkflowCache.getData()
+            # state lists which shouldn't be populated in wmbs. (To prevent creating work before WQE status updated)
+            # added completed list due to the race condition (The extra cost of the query is ~1sec per request)
+            requestsToKill = self.reqmgr2Svc.getRequestByStatusFromMemoryCache(["aborted", "aborted-completed", "force-complete", "completed"])
```
amaltaro: I guess this one will also kill logcollect/cleanup subscriptions for completed (force-completed) workflows, no?
I'm wondering whether it would be better to have a different approach for WorkQueueManager, just to make sure we'll only acquire work that's valid. Something like
1. fetch a list of acquired/running-open/running-closed workflows
2. acquire anything that's in that list
3. anything else that is in wmbs - except logcollect and cleanup - gets removed.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
```diff
@@ -210,12 +210,18 @@ def getAbortedAndForceCompleteRequestsFromMemoryCache(self, expire=0):
...
         """
+
+
+        maskStates = ["aborted", "aborted-completed", "force-complete"]
+        return self.getRequestByStatusFromMemoryCache(maskStates, expire)
+
+    def getRequestByStatusFromMemoryCache(self, statusList, expire=0):
+
         # imports here to avoid the dependency not using this function
         from WMCore.Cache.GenericDataCache import MemoryCacheStruct
```
amaltaro: can we move this import to the top of the module? I think it's better to have it all in the same place (even if it's not used all the time) than having it spread along the code. But that's just my opinion :)

amaltaro (commented):
Patch looks Ok. However, I'd rather have something that will not allow rogue jobs to keep running/acquiring.
In order to avoid condor scalability issues, I was even considering a new worker thread just to handle these cases, with a slightly longer polling cycle.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
ticoann: Services gets imported to  other projects wmarchive, crab, I was afraid break anything due to dependency. But I think it is OK for this.
https://github.com/dmwm/WMCore/blob/master/setup_dependencies.py
I will move the import up

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
ticoann: we only checked aborted since after this aborted  workflow supposed to be change to aborted-completed.
As far as I know we don't kill log collect and cleanup jobs.

ticoann:
test this please

ticoann:
@alexanderrichards, Alan could review again. I ran the test in vm and added unittest.

ticoann:
test this please

src/python/WMCore/WMBS/MySQL/Subscriptions/CountFinishedSubscriptionsByWorkflow.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class CountFinishedSubscriptionsByWorkflow(DBFormatter):
+    """
+    Gets count subscriptions given workflow but exculding cleanup and log collect subscription
+    """
+
+    sql = """SELECT ww.name as workflow,
+             COUNT(case when ws.finished = 1 then 1 else null end) as finished,
+             COUNT(case when ws.finished = 0 then 1 else null end) as open
```
amaltaro: how about we fetch only "open" subscriptions? In the end we don't really care about finished ones and it might buy us 1ns :-)

src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
```diff
@@ -47,9 +53,14 @@ def algorithm(self, parameters):
             # this will clean up whatever left over from above clean up.
             # also if the wq replication has problem it won't delay the killing jobs in condor
             # and updating wmbs status
-            abortedAndForceCompleteRequests = self.abortedAndForceCompleteWorkflowCache.getData()
+            # state lists which shouldn't be populated in wmbs. (To prevent creating work before WQE status updated)
+            # added completed list due to the race condition (The extra cost of the query is ~1sec per request)
```
amaltaro: I guess this comment (1sec per request) is no longer true, is it? We only fetch request names given a list of request status (thus a single reqmgr2 http call).

src/python/WMCore/WMBS/MySQL/Subscriptions/CountFinishedSubscriptionsByWorkflow.py
amaltaro: BTW, did you try this query in an oracle backend?

src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
```diff
@@ -47,9 +53,14 @@ def algorithm(self, parameters):
...
+            requests = self.reqmgr2Svc.getRequestByStatusFromMemoryCache(["aborted", "aborted-completed", "force-complete", "completed"]).getData()
+            results = self.finishedWorflowCheck.execute(workflowNames=requests)
+
+            requestsToKill = [reqInfo["workflow"] for reqInfo in results if reqInfo["open"] > 0]
 
-            for wf in abortedAndForceCompleteRequests:
+            for wf in requestsToKill:
                 self.queue.killWMBSWorkflow(wf)
```
amaltaro: Did you check whether it also kills logcollect and cleanup subscriptions/jobs?

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
amaltaro: And the reason we kill on force-completed and aborted only - at this stage - is because `killWMBSWorkflow` from WorkQueueManager will kill subscriptions and also jobs from condor. Is that correct or I'm missing something?

amaltaro (commented):
Seangchan, patch looks good. I did ask a few questions in the code though.
I also made a suggestion for the DAO.

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
ticoann: Yes, that is right. Then it will move to aborted-completed, completed state after that. 
In principle, it doesn't make sense to check aborted-completed, completed at this point.

src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
ticoann: Yes I will remove that. Thanks

src/python/WMCore/WMBS/MySQL/Subscriptions/CountFinishedSubscriptionsByWorkflow.py
ticoann: I think performance difference are negligible. and this can be used (with high priority). 😄 
Yes I ran this in Oracle.

src/python/WMComponent/WorkQueueManager/WorkQueueManagerCleaner.py
ticoann: Yes, it doesn't kill LogCollect or Cleanup Jobs. Although as I mentioned killed job is not reported in wmstats but that might be different issue.

ticoann:
test this please

ticoann:
test this please

ticoann:
test please

ticoann:
test this please

ticoann:
test this please

ticoann:
test this please

PR #8356, opened 2017-11-21
fix the priority handling

ticoann:
@amaltaro, please review. If it looks OK I will add to WMAgent branch patch

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -85,6 +85,7 @@ class JobSubmitterPoller(BaseWorkerThread):
         self.maxTaskPriority = getattr(self.config.BossAir, 'maxTaskPriority', 1e7)
         self.condorFraction = 0.75  # update during every algorithm cycle
         self.condorOverflowFraction = 0.2
+        self.ioboundTypes = ('LogCollect', 'Merge', 'Cleanup', 'Harvesting')
 
         # Additions for caching-based JobSubmitter
         self.cachedJobIDs = set()
@@ -337,7 +338,7 @@ class JobSubmitterPoller(BaseWorkerThread):
             # try to remove draining sites if possible, this is needed to stop
             # jobs that could run anywhere blocking draining sites
             # if the job type is Merge, LogCollect or Cleanup this is skipped
-            if newJob['type'] not in ('LogCollect', 'Merge', 'Cleanup', 'Harvesting'):
+            if newJob['type'] not in self.ioboundTypes:
                 nonDrainingSites = [x for x in possibleLocations if x not in self.drainSites]
                 if nonDrainingSites: # if >1 viable non-draining site remove draining ones
                     possibleLocations = nonDrainingSites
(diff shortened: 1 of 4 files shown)


amaltaro:
It seems to be working fine. However, I think we should have a protection for IOBound jobs such that we don't oversubscribe a site with them. In other words, we should not over-submit IOBound jobs.

How about we create a variable in the init method:
`self.ioboundTypes = ('LogCollect', 'Merge', 'Cleanup', 'Harvesting')`

and then add a `if jobType in self.ioboundTypes or ... ` to this line:
https://github.com/ticoann/WMCore/blob/a1ae58d4df872c32d5e892564be47d0d2222bcd9/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L558

?

amaltaro:
If you do so, also update the L340 with the var name instead of that tuple, please.

ticoann:
So for IOBound Jobs you don't want to put higher priority jobs submitted over the threshold?

ticoann:
Thanks Alan for review, I updated the code as you suggested.

amaltaro:
I think it's safer for sites, otherwise we might have way too much of them pending/running in a site and cause storage issues. They anyways have the highest possible priority 50 million or so, so they are always the first one to get submitted.

amaltaro:
And getting back to the lines we have had a previous discussion:
https://github.com/ticoann/WMCore/blob/87d9830479bac105a91979ea101d2ccaf0102993/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py#L570

it's oversubmitting jobs based on `jobs_pending * (1 + condorFraction)` when it's supposed to use `slots_pending * condorFraction + max(jobs_pending, slots_pending)`.

Example of the current situation:
slots_pending are 500 and so is jobs_pending, 20% more will be 600 slots_pending and a max of 600 jobs_pending.
Then a higher prio workflow comes in, it will oversubmit 120 jobs (20% of 600), so 720 now.
Then another higher prio, thus 144 additional jobs and so on.

If we really want to oversubmit based on the condor overflow fraction and the original pending slots, then we need to change it. I think it would be fine to keep it as it is, but in case we hit too many different priorities, we might hit the roof with this over submission.

ticoann:
Hmm, that is weird. I cannot see why it is doing that. I will try to fix it. Thanks Alan

ticoann:
Hi Alan, feel free to fix the problem if you can spot easily. (I couldn't see). If not, do bother now. As you said it shouldn't be a big deal. We can may be fix in next release.

ticoann:
Alan, I was wondering you were just looking at the Processing/Production jobs. Or are some of jobs are newly added IOBound jobs (20 of them)?

ticoann:
test this please

amaltaro:
Those were Production (or Processing), I didn't have any merge jobs in the system. It should be a simple change, I just didn't want to run my tests again :P

amaltaro:
My explanation above is exactly how you implemented... that's why you couldn't find a problem, there is none :) Sorry for the noise.

amaltaro (approved):

amaltaro:
Seangchan, maybe run a pylint cleanup. There are tons of lines in the report. I'm fixing the py3 compatibility in a different PR.

ticoann:
I will do py3 compatibility as well. Otherwise there will be conflict

PR #8299, opened 2017-11-03
add number of core info in performance error message

ticoann:
fixes #8271 @amaltaro, Alan please review

PR #8358, opened 2017-11-22
Fix for printing number of cores in performance kills

amaltaro:
Fix for #8299 and #8312
We don't use this cores arguments anywhere in the PerformanceMonitor, really just informational.

### DIFF
--- a/src/python/WMCore/WMRuntime/Watchdog.py
+++ b/src/python/WMCore/WMRuntime/Watchdog.py
@@ -77,8 +77,9 @@ class Watchdog(threading.Thread):
             logging.info(msg)
             mon = self.loadMonitor(monitor)
             args = {}
-            if hasattr(task.data.watchdog, monitor):
-                #This should be a config section
+            if hasattr(task.data.watchdog, monitor) and monitor == 'PerformanceMonitor':
+                # Apply tweaks to PerformanceMonitor only.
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

amaltaro:
Already tested and it's good to go.

ticoann (approved):

ticoann:
Thanks Alan

src/python/WMCore/WMRuntime/Watchdog.py
```diff
@@ -103,8 +104,8 @@ def setupMonitors(self, task, wmbsJob):
                 # If we did base maxRSS off the memory in the HTCondor slot, subtract a bit
                 # off the top so watchdog triggers before HTCondor does.
                 # Add the new number of cores to the args such that DashboardInterface can see it
+                args['cores'] = resources['cores']
```
ticoann: Actually I was wondering whether we should be put below instead.

args['cores'] = changedCores

Not sure which core number we should propagate.

amaltaro: changedCores is just a boolean value. Before calling `resizeResources`, `resources` has the original number of cores, once it comes back, it might have a different value (resources is passed by reference).

PR #8359, opened 2017-11-22
Add Phedex Mock object to immitate phedex subscription

vkuznet:


test/python/WMCore_t/Services_t/MicroService_t/Unified_t/Transferor_t.py
```diff
@@ -14,22 +14,25 @@
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
 from WMCore.Services.MicroService.Unified.Transferor import \
         RequestStore, RequestManager
-
+from WMQuality.Emulators.PhEDExClient.MockPhEDExApi import MockPhEDExApi
 
 class TransferorTest(unittest.TestCase):
```
amaltaro: in order to use the mocked phedex, you need to use the `EmulatedUnitTestCase` super class instead.
Then you call the setUp method from the super class too, just as Seangchan suggested in the GH issue.

amaltaro (changes requested):

test/python/WMCore_t/Services_t/MicroService_t/Unified_t/Transferor_t.py
vkuznet: and can you explain why do we need EmulatedUnitTestCase in order to use mocked phedex. I tried it and it works w/o EmulatedUnitTestCase and it works with EmulatedUnitTestCase. So my question is what's the point of using it?

amaltaro: It worked without it either because a) it is still making real calls to the data service or b) because you're explicitly passing a mockphedex object to the src class code.

With EmulatedUnitTestCase, it uses the mock.patch to remap calls to PhEDEx to a MockPhedex instead, such that we don't need to have a special handle for tests in the src/ code

vkuznet:
Alan,
if you'll look at my code you'll see that I need to use mock instance
internally, since I make explicit API calls, e.g.
https://github.com/dmwm/WMCore/pull/8359/files#diff-46afd98b16c505aaf350d42eca7c2bb5L29

In other words you shown me usage of Mock class implicitly, while I need to use
it explicitly. I doubt that mock.patch does it correctly since I just tested
it and it make real calls to PhEDEx data-service, e.g.

```
class TransferorTest(EmulatedUnitTestCase):
    def setUp(self):
        super(TransferorTest, self).setUp()
        self.phedex = PhEDEx()
        data = self.phedex.subscriptions(group=self.group)
        # and here I GOT real phedex data not the mock ones
```
In other words the mock.patch does not mock PhEDEx() instance. In fact this is
true if you'll read
https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/Emulators/EmulatedUnitTestCase.py
It does not create MockPhEDExAPI instance anywhere.

Therefore my point is still valid, why should I use it if it does not provide
functionality I need in this specific use case. Anyhow I need to use 
`self.phedex = MockPhEDExApi()` to fetch data first before I perform real tests.

amaltaro:
Since you import PhEDEx into this module:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/MicroService/Unified/Transferor.py#L93

you need to also add this module to the list of patches to be applied, here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/Emulators/EmulatedUnitTestCase.py#L61

plus the final tweaks to the mock class:
https://github.com/dmwm/WMCore/blob/master/src/python/WMQuality/Emulators/PhEDExClient/MockPhEDExApi.py#L160

unless I'm missing something, that should be all you need.

vkuznet:
Well, I did what you asked for:

```
--- a/src/python/WMQuality/Emulators/EmulatedUnitTestCase.py
+++ b/src/python/WMQuality/Emulators/EmulatedUnitTestCase.py
@@ -60,6 +60,7 @@ class EmulatedUnitTestCase(unittest.TestCase):
             self.phedexPatchers = []
             patchPhedexAt = ['WMCore.Services.PhEDEx.PhEDEx.PhEDEx', 'WMCore.WorkQueue.WorkQueue.PhEDEx',
                              'WMCore.Services.DBS.DBS3Reader.PhEDEx',
+                             'WMCore.Services.MicroService.Unified.Transferor'
                              'WMComponent.PhEDExInjector.PhEDExInjectorPoller.PhEDEx']
             for module in patchPhedexAt:
                 self.phedexPatchers.append(mock.patch(module, new=MockPhEDExApi))
```

and now it complains about the following

```
  File "test/python/WMCore_t/Services_t/MicroService_t/Unified_t/Transferor_t.py", line 24, in setUp
    super(TransferorTest, self).setUp()
  File "/Users/vk/CMS/DMWM/GIT/WMCore/src/python/WMQuality/Emulators/EmulatedUnitTestCase.py", line 67, in setUp
    self.phedexPatchers[-1].start()
  File "build/bdist.macosx-10.9-x86_64/egg/mock.py", line 1396, in start
    result = self.__enter__()
  File "build/bdist.macosx-10.9-x86_64/egg/mock.py", line 1252, in __enter__
    self.target = self.getter()
  File "build/bdist.macosx-10.9-x86_64/egg/mock.py", line 1414, in <lambda>
    getter = lambda: _importer(target)
  File "build/bdist.macosx-10.9-x86_64/egg/mock.py", line 1102, in _importer
    thing = _dot_lookup(thing, comp, import_path)
  File "build/bdist.macosx-10.9-x86_64/egg/mock.py", line 1091, in _dot_lookup
    __import__(import_path)
ImportError: No module named TransferorWMComponent
```

which means that I need to implement TransferorWMComponent???

I think it's too much headache already to emulate simple thing. I did it
explicitly and it works. If you want to do it through EmulateUnitTest class
then make it working that I should only import it and use it. So far
it is not working and you certainly missing something in a logic.

amaltaro:
I think you missed a comma in your code, that's why you got that ImportError.
BTW, I'm very picky with the WMCore code and against "it works so take it" :-)

We should try our best to keep the code consistent, maintainable and not reinvent the wheel.
And that's why I *explicitly* said in the issue that I was "not asking you to fix it"...

vkuznet:
Alan,
good catch with comma, but even though it fixed the code loading part it does
not mock, it still places calls to phedex because I initialize
`self.phedex = PhEDEx()` in a unit test not in my Transferor code.

And, I don't see anywhere in a ticket that you were not asking me to fix it
since except you and me no one in a ticket list. Are you going to fix it?

src/python/WMQuality/Emulators/EmulatedUnitTestCase.py
```diff
@@ -60,6 +60,7 @@ def setUp(self):
             self.phedexPatchers = []
             patchPhedexAt = ['WMCore.Services.PhEDEx.PhEDEx.PhEDEx', 'WMCore.WorkQueue.WorkQueue.PhEDEx',
                              'WMCore.Services.DBS.DBS3Reader.PhEDEx',
+                             'WMCore.Services.MicroService.Unified.Transferor',
```
ticoann: This need to be 
WMCore.Services.MicroService.Unified.Transferor.PhEDEx

test/python/WMCore_t/Services_t/MicroService_t/Unified_t/Transferor_t.py
```diff
@@ -14,22 +14,27 @@
...
-class TransferorTest(unittest.TestCase):
+class TransferorTest(EmulatedUnitTestCase):
     "Unit test for Transferor module"
     def setUp(self):
         "init test class"
+        super(TransferorTest, self).setUp()
         self.group = 'DataOps'
         self.interval = 2
-        self.rmgr = RequestManager(group=self.group, interval=self.interval, verbose=True)
         self.phedex = PhEDEx()
```
ticoann: This still need to use mock

self.phedex = MockPhEDExApi()

vkuznet:
Seangchan, Alan, it seems you didn't read or understand my comments. I changed the code that you'll see what's going on. Now it uses EmulatedUnitTestCase class and setup it properly. BUT (!!!) it still queries real Phedex data-services. Here is a call line:
https://github.com/dmwm/WMCore/pull/8359/files#diff-46afd98b16c505aaf350d42eca7c2bb5L29
and it returns this data:
```
{u'phedex': {u'request_date': u'2017-11-27 16:37:02 UTC', u'request_timestamp': 1511800622.52604, u'call_time': 0.00368, u'dataset': [{u'files': 37, u'name': u'/SingleMuon/Run2017G-ZMu-PromptReco-v1/RAW-RECO', u'bytes': 121933216, u'is_open': u'y', u'id': u'1092277', u'subscription': [{u'node': u'T1_US_FNAL_MSS', u'percent_files': 0, u'custodial': u'y', u'suspend_until': None, u'node_files': 0, u'level': u'DATASET', u'time_create': 1511757193, u'move': u'n', u'request': 1155203, u'percent_bytes': 0, u'group': u'DataOps', u'priority': u'high', u'time_update': 1511757193, u'node_id': 10, u'suspended': u'n', u'time_start': None, u'node_bytes': 0}]}], u'instance': u'prod', u'request_call': u'subscriptions', u'request_url': u'http://cmsweb.cern.ch:7001/phedex/datasvc/json/prod/subscriptions', u'request_version': u'2.4.0pre1'}}
```
which if I read it correctly the data coming from real Phedex data-service and not from a mock.

Please review the whole code, not just committed changes since it seems you missed the point I"m talking about. Event with EmulatedUnitTestCase I do see that code still calls real Phedex data-service.

ticoann:
Hi Valentin, thanks for updating the code. I found a couple of errors that is why it is not using the mockPhEDEx. Other than that code looks good.

vkuznet:
Code committed and unit test runs fine in my setup. I also squashed the changes.

ticoann:
Thanks Valentin. If Alan is OK with the code. I will merge it when unittest comes back from Jenkins.

amaltaro:
test this please

ticoann:
test this please

test/python/WMCore_t/Services_t/MicroService_t/Unified_t/Transferor_t.py
```diff
@@ -14,22 +14,26 @@
...
         self.rmgr = RequestManager(group=self.group, interval=self.interval, verbose=True)
-        self.phedex = PhEDEx()
 
         # get some subscriptions from PhEDEx to play with
         data = self.phedex.subscriptions(group=self.group)
+        print("### data", data)
         for datasetInfo in data['phedex']['dataset']:
             dataset = datasetInfo.get('name')
-            print("### dataset info from phedex, #files %s" % datasetInfo.get('files', 0))
+            print("### dataset info from phedex, dataset %s #files %s" \
```
ericvaandering: Is there a reason to keep all these print statements in the test now that it is presumably working OK?

vkuznet:
I still prefer to keep it since these tests are quite complex and it's better to
understand what's going on in them.

PR #8361, opened 2017-11-24
Mistakenly disabled dashboard monitor in 8358

amaltaro:
Fix a bug injected in #8358... I didn't notice args was used after the `if` block. This simple fix is haunting us...


amaltaro:
@ticoann it has been backported already to 1.1.8, but please review even if it's already merged :)

PR #8375, opened 2017-11-30
StepChain template with IncludeParents; fixed TaskChain one

amaltaro:
Because they never stop coming.... merge jobs are failing for these templates, though it's enough to test the IncludeParents feature.

TODO: update TaskChain_EpL.json, the input dataset no longer exists (after the migration of dbs prod into int).


amaltaro:
Second commit has new templates for ReDigi and StepChain, since their previous input dataset got deleted with the DBS prod -> int migration.
The NanoAOD template now uses the new merge settings from the CMSSW.

PR #8379, opened 2017-12-04
get campaign info from reqmgr aux db

### LINKED ISSUE #8132: Create the API to update Unified config to Aux db
created the api to read and write for unified cofig

ticoann:
fixes #8132 
Still need clean up and fix unitest (Just for the initial test)


ticoann:
@vkuznet, @vlimant, I am trying to move campaign  config to couch db so it can be added and updated easily. I have question about the config format.

https://github.com/ticoann/WMCore/blob/770efec8ded37fcf29381999486430a21873a8ae/src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/CAMPAIGN_CONFIG.py

Is there a reason to separate relval and production/processing campaign separately? It seems code combines that anyway before processing. 

Also some of the parameters in different level. i.e. SiteWhiteList can go in the top level,  can we have more consistent structure? Also it seems it supports wild cards for site list in the code but config doesn't have that cases, should we support wild card?

vkuznet:
This config in many ways is not ideal. Here I want to outline potentials
problems:
- we should never use dynamic keys, e.g. keys 'HIRun2015', 'TTI2023Upg14D'
have no meaning and such keys are "dynamic" since once we create
new value we create new key, while the key 'Campaign' is explicit and
"static" (its value is dynamic). Therefore I suggest to replace 
```
    "Summer16Geant4102": {
        "go": true,
        "parameters": {
            "EventsPerLumi": "x10"
        }
    }
```
to
```
    "Campaign": {
        "name": "Summer16Geant4102",
        ....
    }
```
- don't use dictionary when it is no necessary, e.g.
`"secondaries": {"/PYTHIA6_MinBias_TuneZ2star_14TeV/TTI2023Upg14GS-FLATBS15_DES23_62_V1-v1/GEN-SIM": {}...`
may have better represented with a list like
`"secondaries": ["X", "Y", ...]` where X, Y are dataset names.
- use better naming conventions and structures, e.g. instead of
`"overflow": {"PU": {"max": 10000, "pending": 0}`
I'll suggest
`"overflow": {"name": "PU", "max": 10000, "pending": 0}`
- I have no idea if this `"go": true` is required in all structures, it seems to
me that all of them use identical true value, therefore it can be lifted to
upper level.
- if SiteBlacklist is static across all datasets/workflows it is better to be
defined at higher level of the structure
- overall I see that there is a repetition in sub-structures, i.e.
```
"Run2016C": {
        "go": true,
        "labels": ["03Feb2017", "15Feb2017", "22Feb2017", "18Apr2017"],
        "fractionpass": 1.0,
        "lumisize": -1,
        "maxcopies": 1,
        "overflow": {"PRIM": {}}
    },
    "Run2016D": {
        "go": true,
        "labels": ["03Feb2017", "15Feb2017", "22Feb2017", "18Apr2017"],
        "fractionpass": 1.0,
        "lumisize": -1,
        "maxcopies": 1,
        "overflow": {"PRIM": {}}
    },
```
So, instead you better use list for that instead of dict key-values
sub-strcutures, e.g.
```
[
  {"campaign": "Run2016C", "parameters: {"go":true, ...}},
  {"campaign": "Run2016D", "parameters: {"go":true, ...}},
]
```
The advantage that it requires less space, have static structure (which implies
easy to parse and port), etc.
- I'm not sure why CAMPAIGN_CONFIG should contain relval config part
- it would be nice to settle on common schema for campaign and relval parts
- it seems to me that relval SiteWhiteList are common across all entries,
therefore it can be taken away from all sub-structures to higher level.

ticoann:
Hi Valentin, I agree with you. I am just wondering why current structure is set up as it is. Since we can control whole thing we should restructure the configuration and code accordingly.

vkuznet:
The short answer is that everybody busy and want to get things done, and
people forget to think about sustainability of their code, its portability
and effectiveness.

ticoann:
I am not blaming any one. :smile:  I just want to know whether I can change the data structure and don't want to miss some important point of current structure.

So if you agree I will change the structure as follows. (I don't think 'parameter' level is necessary)

[
  {"campaign": "Run2016C", "go":true, ...},
  {"campaign": "Run2016D", "go":true, ...},
]

vkuznet:
Yes, I agree to use such list structure.

ticoann:
Looking at the operation against this, actually there is advantage on dict data structure.
I assume there will be more update/delete operation on individual campaign than whole document replacement.
In that case, list structure is not very convenient. How about we keep the dict structure something like this

{"Run2016C": {"campaign": "Run2016C", "go":true, ...},
"Run2016D": {"campaign": "Run2016D", "go":true, ...},
}

But provide REST api which can return the list format as well. (This could be default format for REST api)

[
{"campaign": "Run2016C", "go":true, ...},
{"campaign": "Run2016D", "go":true, ...},
]

There is down size doing streaming on this structure but I don't think we will do over just one document anyway.

vkuznet:
And, what are advantages?

ticoann:
> And, what are advantages?

Update and delete.  Instead of searching the list every time we do.

amaltaro:
My vote goes to
```
[
{"campaign": "Run2016C", "go":true, ...},
{"campaign": "Run2016D", "go":true, ...},
]
```

even if that will cause a whole list scan. It's only one structure anyways and quite small O(10), I think.

vkuznet:
Seangchan, if list is small, e.g. < 1000, the effect of data look-up is
negligible. But update/deletion of a dictionary is in fact a larger
factor since dictionary size should shrink of grow and therefore
memory allocation will involve. Since the structure we have
is very nested the insert and update will be costly operation.

So, I would not worry about such issue unless you profile and prove
that this is the real issue.

ticoann:
The only difference for the list format and dict format is we have additional key (campaign) in terms of nested structure. I don't think is extra code when we use this format for update.
So for the outside operation which query this doc won't see the difference.

We can always return list format through API unless it specifies. 
[
{"campaign": "Run2016C", "go":true, ...},
{"campaign": "Run2016D", "go":true, ...},
] 

The reason I prefers dict format for the couch doc, since all our couch doc in Aux. is dict, hence we have general update function which handles document updates. This will have additional update function for the list format of document. 
I don't see particular advantage to storing document as list in that aspect. (return format will be list)

Since the vote is 2:1, I can still change what you guys prefer. :smile:

ticoann:
test this please

ticoann:
@vkuznet, Valentin, I reorganize the directory structure as others. Could you check on this?
I am not sure where is the right place for Mananager.py, WMCore.MicroService. or WMCore.MicroService.Service? Anyway, if you can just comment on the directory structure, that will be great. For the code wise, I still have to update config data structure.

corresponding deployment script changes are here.
https://github.com/dmwm/deployment/pull/572

src/python/WMCore/MicroService/Unified/RequestInfo.py
```diff
@@ -293,7 +254,7 @@ def getMulticore(request):
         mcores.extend([int(v) for v in mcoresCol.values()])
     return max(mcores)
 
-def getSiteWhiteList(request, siteInfo, reqSpecs=None, pickone=False, verbose=True):
+def getSiteWhiteList(request, siteInfo, campaignInfo, reqSpecs=None, pickone=False, verbose=True):
```
vkuznet: what is campaignInfo parameter looks like? Is new class CampaignConfig or something else?

vkuznet (commented):
I am not sure I understand the following changes:
- file rename
test/python/WMCore_t/Services_t/MicroService_t/Unified_t/__init__.py → __init__.py
test/python/WMCore_t/Services_t/MicroService_t/__init__.py → src/__init__.py
- why CampaignConfig is moved into WMCore/ReqMgr/Service/Auxiliary module?

ticoann:
> file rename
> test/python/WMCore_t/Services_t/MicroService_t/Unified_t/init.py → init.py
> test/python/WMCore_t/Services_t/MicroService_t/init.py → src/init.py

I think this was created when I set up the patch in IDE. I removed it.

> why CampaignConfig is moved into WMCore/ReqMgr/Service/Auxiliary module?

This is hold all the default config in Aux DB. The structure of config data still needs to be changed as we discussed.

src/python/WMCore/MicroService/Unified/RequestInfo.py
ticoann: It is CampaignConfig class which is the dictionary. I think it works with get but not parameters I will update the code

ticoann:
@vkuznet, Valentin, If the unittest comes back successful, I will merge this for further test unless you found some obvious problems. Aux db will contain each campaign config instead of multiple campaigns in one document as we discussed. Still need to add the validation for those docs though.

ticoann:
test this please

src/python/WMCore/ReqMgr/Service/Auxiliary.py
```diff
@@ -268,3 +269,12 @@ class PermissionsConig(AuxBaseAPI):
 
     def setName(self):
```
amaltaro: since you're changing these files, can you fix a typo in here `PermissionsConig`?

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -123,6 +123,59 @@ def updateAgentDrainingMode(self, agentName, drainFlag):
...
+              "T2_US_Nebraska",
+              "T2_US_UCSD",
+              "T2_US_Wisconsin"
+             ]
+           }
+          }
+
+        :return: None
+        """
+        return self["requests"].post('campaignconfig' % campaignName, campaignConfig)[0]['result']
```
amaltaro: Aren't you missing the document name here? Something like
` return self["requests"].post('campaignconfig/%s' % campaignName, campaignConfig)[0]['result'] `

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -123,6 +123,59 @@ def updateAgentDrainingMode(self, agentName, drainFlag):
                                    (drainFlag, resp))
             return False
 
+    def getCampaignConfig(self, campaignName):
+        """
+        get campaign config for transferor function in unified ReqMgr2MS.
+        """
+        return self._getDataFromMemoryCache('campaignconfig/%d' % campaignName)
```
amaltaro: %s instead?

src/python/WMCore/ReqMgr/Service/Auxiliary.py
```diff
@@ -268,3 +269,12 @@ class PermissionsConig(AuxBaseAPI):
...
         self.name = "PERMISSION_BY_REQUEST_TYPE"
+
+
+class CampaignConfig(AuxBaseAPI):
+    """
+    Config which used for Agent
+    """
+
+    def setName(self):
+        self.name = "CAMPAIGN_CONFIG"
```
amaltaro: won't it create a default configuration doc (name CAMPAIGN_CONFIG) that will never be used? Since all the other campaign docs are named after the campaign name.

amaltaro (commented):
I'm not sure I followed all your changes, even after reading the whole discussion again :)
Nonetheless, I made a few comments along the code, otherwise I think we can merge and proceed with further testing.

src/python/WMCore/ReqMgr/Service/Auxiliary.py
ticoann: Thanks Alan, I will update that.

src/python/WMCore/ReqMgr/Service/Auxiliary.py
ticoann: when post call is used subname is passed. So CAMPAIGN_CONFIG_[actual_campaign_name] will be uploaded

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
ticoann: Thanks, fixed it.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
ticoann: You are right, updated now.

ticoann:
Thank you very much Alan for the review. I updated the code when Jenkins comes back ok, I will merge/create a new tag and start testing.

Issue #8388, opened 2017-12-14
add 'reserved' for PhEDEx priority


ticoann:

amaltaro:
Fixed by #8390

PR #8389, opened 2017-12-14
temp fix for phedex priority

ticoann:
temp fix for #8388 - need to patch agents.

### DIFF
--- a/src/python/WMCore/Services/PhEDEx/DataStructs/SubscriptionList.py
+++ b/src/python/WMCore/Services/PhEDEx/DataStructs/SubscriptionList.py
@@ -57,7 +57,7 @@ class PhEDExSubscription(object):
             # Validation checks on the subscription
             for option in (self.static, self.custodial, self.request_only, self.move):
                 assert option in ('y', 'n')
-            assert self.priority in ('low', 'normal', 'high')
+            assert self.priority in ('low', 'normal', 'high', 'reserved')
             assert self.level in ('dataset', 'block')
             if self.level == 'block':
                 assert self.blocks is not None


amaltaro:
Seangchan, it's not really needed for the agent-side only. However, if you want to keep things consistent, we should update the validate function as well:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1116

Once this is tested and pushed in everywhere (if not yet), can you please merge this one and cut a new release (patch3 I think?). I'll probably have to deploy another agent or two in the coming days.

ticoann:
Yes this is a patch just for agent.
Real patch is here. https://github.com/dmwm/WMCore/pull/8390 :smile: 
It seems working fine so I will patch all the agents today and create a new tag. (I might need another agent patch for ACDC parentage).

And stop working on vacation - I can leave you a lot of work when you come back. 😄

amaltaro:
Don't worry, you can leave just a bit for me ;)

Question is, how Unified managed to assign workflows with this `Reserved` priority? Isn't the validate function working?

ticoann:
Unified update priority directly to PhEDEx, so it shouldn't affect agent. But we are pulling the existing subscription an matches it.
 
https://github.com/CMSCompOps/WmAgentScripts/blob/56b0ea9402d7a5b76e8fe2ce463005421f2a3843/Unified/closor.py#L42

This is the code where it is failing. It pulls existing subscription for check.
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L590

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/PhEDEx/DataStructs/SubscriptionList.py#L166

Then fails below.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/PhEDEx/DataStructs/SubscriptionList.py#L60

So when unified update DQM dataset subscription priority it starts failing the WMAgent.
I think the procedure need to be reviewed, and maybe update subscription should happen in request manager.

amaltaro:
Or perhaps we just want to hardwire DQM/DQMIO dataset subscriptions to the `reserved` queue?
Otherwise, we might explore how T0 subscriptions settings work and support the same for Production (specific settings by datatier).

hufnagel:
T0 subscriptions don't do anything special, specific settings by datatier (and some other parameters) are supported by the WMCore code already. You just need to pass the relevant parameters (and make multiple calls for DQMIO and the other tiers).

I don't know what would happen if you make a specific call for DQMIO and a generic call without data tier. We don't use it like that in the Tier0 (we make calls for every data tier we produce), although for you that seems like the best option (if it works correctly).

ticoann:
Actually, it seem it only move to reserved when all the DQM dataset is create not before that. but not sure what is the logic behind it. As far as I know reserved works as higher than high priority in basic but sounds more complicated when Natalia explained to me. - which means I don't understand what reserved priority does exactly.

amaltaro:
@ticoann is there any other patch for the agent branch? if not, can you cut a new release and make a cmsdist gcc630 PR please?

ticoann:
Hi Alan, there is one more patch. But I can merge that today.
https://github.com/dmwm/WMCore/pull/8397

Are you going to deploy new agents? If so, just let me know. You shouldn't really be working. :smile:

amaltaro:
Yes, I'm planning to leave another relval agent and another one or two production agents, just to be safe for the Xmas break ;)

amaltaro:
[Fix #8389 - temp fix for phedex priority](https://trello.com/c/f7Y7MIrK/74-fix-8389-temp-fix-for-phedex-priority)

PR #8393, opened 2017-12-15
fix multicore setting for acdc of acdc

### LINKED ISSUE #8392: Multicore in assigning ACDC of ACDC
looking at 
https://cmsweb.cern.ch/reqmgr2/config?name=pdmvserv_task_SMP-PhaseIITDRFall17wmLHEGS-00030__v1_T_171103_204004_8254
https://cmsweb.cern.ch/reqmgr2/config?name=mcremone_ACDC0_task_SMP-PhaseIITDRFall17wmLHEGS-00030__v1_T_171116_004028_825
https://cmsweb.cern.ch/reqmgr2/config?name=vlimant_ACDC1_task_SMP-PhaseIITDRFall17wmLHEGS-00030__v1_T_171122_180054_1020
https://cmsweb.cern.ch/reqmgr2/config?name=mcremone_ACDC2_task_SMP-PhaseIITDRFall17wmLHEGS-00030__v1_T_171213_210739_3479

it looks like ACDC0 maintained the numberOfCores = 8 in the spec, while the other ones switched to numberOfCores = 1

I cannot find any flaw in how things are assigned. From the logs of where the assignement was done, no Multicore parameter was passed on to the assign api, and yet numberOfCores = 1 even for the tasks that specify Multicore=8

somehow @ticoann @amaltaro the base directory value (set to 1 by default) is taken as the value that needs to go everywhere for ACDC of ACDC.

N.B. this is a showstopper for still cursed upgrade requests of highest priority.

FYI @prozober

ticoann:
fixes #8392

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -159,6 +159,13 @@ def validate_resubmission_create_args(request_args, config, reqmgr_db_service, *
         # ACDC of ACDC, we can't validate this case
         # simply copy the whole original dictionary over and accept all args
         createArgs = originalArgs
+        # TODO: This code can cause the problem when ACDC is created before this update which
+        # TODO: doesn't contain "TopLevelOriginalRequestType" so only uncomment this all the acdc are created after this updated
+        #request_args["TopLevelOriginalRequestType"] = originalArgs["TopLevelOriginalRequestType"]
+
+        # TODO: Remove this line when all the acdc are created after this patch is applied and uncomment above line.
+        # WARNING: This doesn't solve the problem when when this is ACDC of ACDC or more. ( > 2 level ACDC)
+        request_args["TopLevelOriginalRequestType"] = originalArgs.get("TopLevelOriginalRequestType") or originalArgs["OriginalRequestType"]
     else:
         # load arguments definition from the proper/original spec factory
         parentClass = loadSpecClassByType(originalArgs["RequestType"])
@@ -166,6 +173,7 @@ def validate_resubmission_create_args(request_args, config, reqmgr_db_service, *
         if originalArgs["RequestType"] in ('StepChain', 'TaskChain'):
             chainArgs = parentClass.getChainCreateArgs()
         createArgs.update(parentClass.getWorkloadCreateArgs())
+        request_args["TopLevelOriginalRequestType"] = originalArgs["RequestType"]
 
     request_args['OriginalRequestType'] = originalArgs["RequestType"]
     cloned_args = initialize_clone(request_args, originalArgs, createArgs, chainArgs)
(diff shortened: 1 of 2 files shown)


ticoann:
There is alternative implementation #8394 
This doesn't change the current meaning of OriginalRequestType which actually means parent request type.

ticoann:
test this please

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -159,13 +159,21 @@ def validate_resubmission_create_args(request_args, config, reqmgr_db_service, *
         # ACDC of ACDC, we can't validate this case
         # simply copy the whole original dictionary over and accept all args
         createArgs = originalArgs
+        # TODO: This code can cause the problem when ACDC is created before this update which
+        # TODO: doesn't contain "TopLevelOriginalRequestType" so only uncomment this all the acdc are created after this updated
+        #request_args["TopLevelOriginalRequestType"] = originalArgs["TopLevelOriginalRequestType"]
+
+        # TODO: Remove this line when all the acdc are created after this patch is applied and uncomment above line.
+        # WARNING: This doesn't solve the problem when when this is ACDC of ACDC or more. ( > 2 level ACDC)
+        request_args["TopLevelOriginalRequestType"] = originalArgs.get("TopLevelOriginalRequestType") or originalArgs["OriginalRequestType"]
```
amaltaro: I don't get your comment, if it's an ACDC of an ACDC, `TopLevelOriginalRequestType` value will be the original RequestType, thus `TopLevelOriginalRequestType` will always have the original request type, no matter which ACDC level it is. Am I missing something (besides the first ACDC being created before this patch)?

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -159,13 +159,21 @@ def validate_resubmission_create_args(request_args, config, reqmgr_db_service, *
...
     else:
         # load arguments definition from the proper/original spec factory
         parentClass = loadSpecClassByType(originalArgs["RequestType"])
         createArgs = parentClass.getWorkloadAssignArgs()
         if originalArgs["RequestType"] in ('StepChain', 'TaskChain'):
             chainArgs = parentClass.getChainCreateArgs()
         createArgs.update(parentClass.getWorkloadCreateArgs())
+        request_args["TopLevelOriginalRequestType"] = originalArgs["RequestType"]
```
amaltaro: In theory this parameter is only allowed for Resubmission specs, no? In this case, shouldn't we remove this line?

src/python/WMCore/ReqMgr/Utils/Validation.py
ticoann: Yes exactly, it doesn't fix if the first ACDC is created before the patch and we have to make the ACDC of that ACDC.

src/python/WMCore/ReqMgr/Utils/Validation.py
ticoann: This is still resubmission case (first leve ACDC) and the condition above is ACDC of ACDC.

ticoann:
Alan stop working and enjoy your vacation.

ticoann:
replaced by #8394

PR #8397, opened 2017-12-19
fix acdc parentage wmagent version 1.1.8

ticoann:
wmagent version of #8396


ticoann:
This only need to be applied none draining agents since agent won't pull the work but for the tails (in case work is pulled but job hasn't been created), I will apply all the draining agents as well.

ticoann:
I found that parent file from ACDC doesn't have any information about events or filesize or lumis either. Since those information is not used and never there, I wouldn't add it now.

amaltaro:
[Fix #8397 - fix acdc parentage](https://trello.com/c/Y6XUMm8U/75-fix-8397-fix-acdc-parentage)

PR #8398, opened 2017-12-20
fix creation failure

### LINKED ISSUE #7536: higher number of creation failures
https://cmsweb.cern.ch/reqmgr2/fetch?rid=fabozzi_Run2016B-v2-ZeroBias6-17Jan2017_8020_170117_170800_2500

https://cmsweb.cern.ch/reqmgr2/fetch?rid=fabozzi_Run2016B-v2-ZeroBias6-25Nov2016_8020_161125_152646_3953

have basically identical settings and yet, the 17Jan has had way more creationg failures. which is counter intuitive, has the code that generate that failure been changed much ?

ticoann:
fixes #7536

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -35,6 +35,8 @@ class EventAwareLumiBased(JobFactory):
 
         self.loadRunLumi = None  # Placeholder for DAO factory if needed
         self.collectionName = None  # Placeholder for ACDC Collection Name, if needed
+        #TODO this might need to be configurable instead of being hardcoded
+        self.defaultJobTimeLimit = 48 * 3600 # 48 hours
 
     def algorithm(self, *args, **kwargs):
         """
@@ -46,7 +48,7 @@ class EventAwareLumiBased(JobFactory):
 
         avgEventsPerJob = int(kwargs.get('events_per_job', 5000))
         jobLimit = int(kwargs.get('job_limit', 0))
-        eventLimit = int(kwargs.get('max_events_per_lumi', 20000))
+        jobTimeLimit = int(kwargs.get('job_time_limit', self.defaultJobTimeLimit))
         totalEvents = int(kwargs.get('total_events', 0))
         splitOnFile = bool(kwargs.get('halt_job_on_file_boundaries', False))
         self.collectionName = kwargs.get('collectionName', None)
@@ -161,9 +163,10 @@ class EventAwareLumiBased(JobFactory):
                 lumisInJobInFile = 0
                 updateSplitOnJobStop = False
                 failNextJob = False
-                # If the number of events per lumi is higher than the limit
+                # If estimated job time is higher the job time limit (condor limit)
                 # and it's only one lumi then ditch that lumi
-                if f['avgEvtsPerLumi'] > eventLimit and f['lumiCount'] == 1:
+                timePerLumi = f['avgEvtsPerLumi'] * timePerEvent
+                if timePerLumi > jobTimeLimit and f['lumiCount'] == 1:
                     failNextJob = True
                     stopJob = True
                     lumisPerJob = 1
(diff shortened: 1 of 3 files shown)


ticoann:
@amaltaro, @vlimant, please review.

I think it is better to make configurable for jobTimeLimit, should it be request level configuration or agent level. If it only consider pilot limit, it can be agent level configuration but if we need fine control per request - not sure about the use case, it can be request level configuration.

ticoann:
make it configurable in request level but haven't added the splitting parameter from web. It will only take the default value.

ticoann:
@amaltaro, Alan could you review this patch?

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -35,6 +35,8 @@ def __init__(self, package='WMCore.DataStructs', subscription=None, generators=N
 
         self.loadRunLumi = None  # Placeholder for DAO factory if needed
         self.collectionName = None  # Placeholder for ACDC Collection Name, if needed
+        #TODO this might need to be configuarble instead of hardcoded
```
amaltaro: typo

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -161,9 +163,9 @@ def algorithm(self, *args, **kwargs):
                 lumisInJobInFile = 0
                 updateSplitOnJobStop = False
                 failNextJob = False
-                # If the number of events per lumi is higher than the limit
+                # If estimated job time is higher the job time limt (condor limit)
```
amaltaro: typo

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -161,9 +163,9 @@ def algorithm(self, *args, **kwargs):
...
                 # and it's only one lumi then ditch that lumi
-                if f['avgEvtsPerLumi'] > eventLimit and f['lumiCount'] == 1:
+                if(f['avgEvtsPerLumi'] * timePerEvent) > jobTimeLimit and f['lumiCount'] == 1:
```
amaltaro: you missed a blank space in the beginning.
BTW, how about we add the job length as well to the error message on this line https://github.com/ticoann/WMCore/blob/ac9f6654cb85a69c37ba48f15302c6fc31792f3e/src/python/WMCore/JobSplitting/EventAwareLumiBased.py#L254

amaltaro (changes requested):
Sorry for the delay, Seangchan.

Patch looks good in general, though I still think it will cause more problems to us when P&R team has to process workflows with very high TimePerEvent...

There are a few places that need to be updated as well (we need to be careful with backwards incompatibility here):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/EDITABLE_SPLITTING_PARAM_CONFIG.py#L8

and also in a few StdSpecs

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
ticoann: Do you want to print out total number of jobs?

amaltaro:
How about we replace this message:
`"File %s has too many events (%d) in %d lumi(s)"`
by
`"File %s has a single lumi with too many events (%d) and it would take %s secs to run"`
?

ticoann:
> "File %s has a single lumi with too many events (%d) and it would take %s secs to run"

secs? not hours? I thought you are relaxing Friday night. 😄

vlimant:
what you have implemented is essentially making the max_events_per_lumi be 48h/timeperevent. Making the job_time_limit configurable is absolutely mandatory in this case. either we can set max_events_per_lumi or job_time_limit via script/interface.

I agree that in case of overestimated timeperevent, Ops would have to set job_time_limit to a very large value ; same as is done these days with setting max_events_per_lumi to an insanely large number.

@amaltaro isn't there a problem with avgEvtsPerLumi not beeing estimated properly when there is a lumi mask somewhere ? and hence leading to artificially large value of avgEvtsPerLumi, and these creation failures.

ticoann:
> what you have implemented is essentially making the max_events_per_lumi be 48h/timeperevent. Making the job_time_limit configurable is absolutely mandatory in this case. either we can set max_events_per_lumi or job_time_limit via script/interface.

job_time_limit can be overwritten when splitting algorithm is set. The question is whether this is request property (for each request)  or general property over all the request. (It sounds like you want to make it request property, right?

> isn't there a problem with avgEvtsPerLumi not beeing estimated properly when there is a lumi mask somewhere ?

I don't think avgEvtsPerLumi will be related to lumi mask unless masked lumis has very different (small or large) event numbers then others. Mask will be applied later.

amaltaro:
job_time_limit has to be a request property (actually step/task), just as TimePerEvent is.

About the lumi mask, no, there is no problem. Even if there was, the algorithm only fails job creation when the file has a single lumi section (where the avgEvtsPerLumi is precise on those cases)

ticoann:
test this please

amaltaro:
Seangchan, we need to change these parameters at ReqMgr level as well (and possible make it backwards compatible in the agent). I'll create another patch later today.

PR #8408, opened 2018-01-10
Remove legacy site black/white list parser

amaltaro:
One of these unit tests failed in my other PR. Having a quick look at it, it isn't used in WMCore/CRABServer/Unified, so I think it's safe to delete it.


amaltaro:
Just in case, @hufnagel @mmascher @ericvaandering you might know of another usage of those...

ericvaandering:
I also seem to have recall finding that this was not used.

hufnagel:
Never seen this, AFAIK we don't use it in the Tier0.

amaltaro:
Then I think we can merge it, @ticoann :)

ticoann (approved):

PR #8404, opened 2018-01-03
temporary fix for encoding problem (Never merge)

ticoann:
temporary fix for #8403  wmagent 1.1.8 version of #8058

PR #8410, opened 2018-01-12
Ensure StageOut error details contain only valid bytes

### LINKED ISSUE #8403: Another unicode encoding failure
It is in

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/Execute.py#L70

output contains "\x95" which cause encoding error
  
and

'utf8' codec can't decode byte 0xb2 in position 24451

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/8403
Replaces #8404 (but to be applied in new agents only)

I did run that recipe in https://github.com/dmwm/WMCore/issues/8056 to see whether I could encode the sicky FJR and it seems to go through with this fix.

Just for reference, another complement to #8295

### DIFF
--- a/src/python/WMCore/WMException.py
+++ b/src/python/WMCore/WMException.py
@@ -27,8 +27,12 @@ class WMException(exceptions.Exception):
     """
     def __init__(self, message, errorNo=None, **data):
         self.name = str(self.__class__.__name__)
-        exceptions.Exception.__init__(self, self.name,
-                                      message)
+        if hasattr(message, "decode"):
+            # Fix for the unicode encoding issue, see #8056 and #8403
+            # interprets this string using utf-8 codec and ignoring any errors
+            message = message.decode('utf-8', 'ignore')
+
+        exceptions.Exception.__init__(self, self.name, message)
 
         #  //
         # // Init data dictionary with defaults
@@ -40,7 +44,7 @@ class WMException(exceptions.Exception):
         self.data.setdefault("ClassInstance", None)
         self.data.setdefault("FileName", None)
         self.data.setdefault("LineNumber", None)
-        if errorNo == None:
+        if errorNo is None:
             self.data.setdefault("ErrorNr", 0)
         else:
             self.data.setdefault("ErrorNr", errorNo)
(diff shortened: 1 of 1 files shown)


amaltaro:
@ticoann can you please double check it before I proceed with pylint and backport it?

amaltaro:
Just in case, you can find an erratic fjr here:
cmst1@vocms0257:/data/srv/wmagent/current $ cp /data/srv/wmagent/v1.1.8.patch2/install/wmagent/JobCreator/JobCache/pdmvserv_task_BPH-RunIIFall17DRPremix-00037__v1_T_171229_130339_8611/BPH-RunIIFall17DRPremix-00037_0/JobCollection_208364_0/job_1777615/Report.2.pkl .

amaltaro:
test this please

ticoann:
test this please

ticoann:
Alan is this code the only path which could pass to below?

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/Execute.py#L70

If that is the case patch looks fine. Otherwise, I would encode above line.

ticoann:
Alan, I don't think this need to be back ported since we have temp patch. Just for the new agent

amaltaro:
I'm not completely sure, but I think the error message/details only get persisted in the FJR in the code I changed in this PR. I can try running a test tomorrow, but it isn't going to be that simple to get it properly tested :(

amaltaro:
@ticoann I was going to apply the changes we discussed, but it seems WMException is even a better place for that and it should then cover the whole code. If you prefer, I can move it back to the StageOutError class.
Otherwise, I think we can merge it if it passes jenkins (I haven't tested though :))

ticoann:
I was thinking more about not applying this in the place if we don't get the those characters. However it would be more robust in WMException, hope it doesn't have any performance issue.
I will merge it when Jenkins is done.

amaltaro:
Just for reference in case we need to fix it again in the next agent release, I'm not completely sure the decoded error message will be passed all the way until here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Diagnostics/StageOut.py#L30

PR #8412, opened 2018-01-12
Fix if condition in TaskArchiver using workqueue

amaltaro:
Bug inserted in #8354 
That's why workflows were not going to completed on my VM running 1.1.10.pre3 :-)


amaltaro:
Tested and it works fine

ticoann (approved):

PR #8419, opened 2018-01-19
Allow SiteWhitelist for StoreResults creation; better validation for …

### LINKED ISSUE #8407: One can approve and set other parameters all together
Apparently Matteo succeeded setting a StoreResults to `assignment-approved` and set SiteWhitelist parameter altogether (thus bypassing its validation).
This should be of course forbidden.

In addition to that, we might allow SiteWhitelist to be set during a StoreResults creation.

amaltaro:
…user input
Fixes #8407 

* allow SiteWhitelist to be set when creating a StoreResults request. It needs to be passed again during assignment though, as any other request.
* perform a better validation of the user input and only proceed if all parameters are supported for that specific action/transition

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -20,7 +20,7 @@ from WMCore.ReqMgr.DataStructs.RequestStatus import (REQUEST_STATE_LIST,
                                                      REQUEST_STATE_TRANSITION, ACTIVE_STATUS)
 from WMCore.ReqMgr.DataStructs.RequestType import REQUEST_TYPES
 from WMCore.ReqMgr.Utils.Validation import (validate_request_create_args, validate_request_update_args,
-                                            validate_clone_create_args, validateOutputDatasets)
+                                            validate_clone_create_args, validateOutputDatasets, workqueue_stat_validation)
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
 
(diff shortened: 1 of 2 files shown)


src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -375,25 +375,32 @@ def _retrieveResubmissionChildren(self, request_name):
...
     def _handleNoStatusUpdate(self, workload, request_args):
         """
-        only few values can be updated without state transition involved
-        currently 'RequestPriority' and 'total_jobs', 'input_lumis', 'input_events', 'input_num_files'
+        For no-status update, we only support the following parameters:
+         1. RequestPriority
+         2. Global workqueue statistics, while acquiring a workflow
         """
-        if 'RequestPriority' in request_args:
+        if 'RequestPriority' in request_args and len(request_args) == 1:
```
ticoann: Shouldn't this regardless number of request_args?

ticoann:
test this please

ticoann:
test this please

src/python/WMCore/ReqMgr/Service/Request.py
amaltaro: Maybe :) The problem is that we would allow updating the request parameters at any point in the lifetime of a request. With this change, it will fail and only allow RequestPriority. If we want to change more parameters, then the only moment is during assignment.

amaltaro:
@mcremone this is now in testbed and it will go in in production on Feb 6th. I'll update the createStoreResults script in the coming days. Can you please remind me what exactly is the key name for the site white list (data location) that end users are providing in the json template?

mcremone:
here is an example of json as we asked from the users:

http://129.13.101.178/~akhmet/embedding_files.json

PR #8424, opened 2018-01-26
fix exicode 0 when job failed for log Collector

ticoann:


src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -570,6 +570,7 @@ def addError(self, stepName, exitCode, errorType, errorDetails):
 
         stepSection = self.retrieveStep(stepName)
 
+        errorCount = getattr(stepSection.errors, "errorCount", 0)
```
amaltaro: Were you trying to add some other line here? It's the same as the one below.

src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -579,6 +580,7 @@ def addError(self, stepName, exitCode, errorType, errorDetails):
         errDetails.details = errorDetails
 
         setattr(stepSection.errors, "errorCount", errorCount + 1)
+        self.setStepStatus(stepName=stepName, status=exitCode)
```
amaltaro: I wonder how all the other steps and job type work just fine if we were missing this setStepStatus?

src/python/WMCore/WMSpec/Steps/Diagnostics/AlcaHarvest.py
```diff
@@ -41,7 +41,7 @@ def __call__(self, errCode, executor, **args):
         if not os.path.exists(jobRepXml):
             # no report => Error
             msg = "No Job Report Found: %s" % jobRepXml
-            executor.report.addError(50115, "MissingJobReport", msg)
+            executor.report.addError(executor.step._internal_name, 50115, "MissingJobReport", msg)
```
amaltaro: Oh really, it was buggy since a long time because we were passing the exit code as stepName? :)

amaltaro (approved):
Besides one duplicate line, changes look good to me.

I'm still missing something though, how could we not spot these problems before if we were always calling that method without all needed args?!?

src/python/WMCore/WMSpec/Steps/Diagnostics/AlcaHarvest.py
ticoann: Wow, really? that was my first reaction too. :-)

src/python/WMCore/FwkJobReport/Report.py
ticoann: My mistake, I will remove that.

src/python/WMCore/FwkJobReport/Report.py
ticoann: I seems most cases, error status checked by error count not status, but there are one place it is checked by status.

ticoann:
test this please

ticoann:
@amaltaro, Alan I still see failed logCollect job with exit code 0. I updated the patch (2nd commit) with possible place to fix that. I am running the test again.

amaltaro:
test this please

amaltaro:
Seangchan, I found this branch when I was trying to solve this issue:
https://github.com/amaltaro/WMCore/commit/e8b36c44652708db03a901122278afb5ad82a6bc#diff-b870c5345972b4ca6bf3e9c4d7252e95R296

I guess it didn't work though, otherwise I'd have made a PR. Anyways, in case it makes sense to you :)

amaltaro:
@ticoann did you run the DMWM tests with this patch? IF so, we can then add it to the patch2 tag. Otherwise, let me run a few tests before we put it in the agent.

ticoann:
I ran the test but after the 3 commit but that time I couldn't have any cases exit 0 with failure to verify. So I am running the another set of the test.

ticoann:
test this please

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -176,6 +176,7 @@ def execute(self, emulator=None):
                 logging.error("Indefinite hang during edmCopyUtil/stageIn of logArchives")
             except StageOutFailure:
                 logging.error("Unable to stageIn logArchives")
+                raise
```
ticoann: @amaltaro, Should we raise here or ignore? raise means if we fail to get the one of input logArchive fail the logCollect job. If we ignore we can still have logCollectJob successful with missing logArchive file (some or all).

amaltaro: We can ignore it. LogCollect is supposed to fail *only* when we fail to collect ALL jobs tarball. If it succeeds collecting at least one, then the job is supposed to succeed.

src/python/WMCore/WMSpec/Steps/Executors/DeleteFiles.py
```diff
@@ -157,7 +157,6 @@ def deleteOneFile(self, fileForTransfer, manager, waitTime):
             logging.error("Traceback: ")
             logging.error(traceback.format_exc())
             self.report.addError(self.stepName, 1, "StageOutFailure", str(msg))
```
amaltaro: This exit code 1 is completely off here. How about we revive one exit code that was removed a couple of years ago `60313`, but I'd rephrase that to `Failed to delete file already staged in`, apparently that's what it's doing (I guess merge and logcollects).

hufnagel: Failing to delete input files for LogCollect jobs after the other parts of the job are done is not an error we care about. As such, I think any instance of this error is caught upstream and ignored.

Not sure how Cleanup jobs are handled, if we actually fail them or not in cases there is a problem deleting files.

ticoann: You are right Dirk. It seems we always consider that as failure in the code. I guess that was a bug.
For clean up job, I think we should consider as failure but I am not sure whether we retry clean up jobs. Anyway, I will remove that and check the clean up jobs.

PR #8425, opened 2018-01-29
Add all WMRuntime logs to the job logArchive; improve logging

### LINKED ISSUE #8423: Include basic debugging information in the cmsRun stdout
Hostname
dump of environment variables
dump of job classad

amaltaro:
Fixes #8423
Changes are:
* replaced all print statements by logging for the runtime code (well, most of it), including the stage out code
* updated some file handling with context managers
* updated the wmagent job log name from `jobLog.PID.log` to `wmagentJob.log`
* added the `wmagentJob.log` and the condor wrapper logs to the logArchive tarball (NOTE: those logs will be incomplete since the job is still running/logging). I checked a before/after my changes and the size difference seems to be around ~5KB per tarball
* added a few 9999? exit codes to the exceptions list
* copy the full `wmagentJob.log` log back to the schedd, once the job finishes.

Only issue is, for some reason most of the runtime logging is going to the wmgentJob.log AND the condor stderr wrapper log(?) If someone has a hint, I'm all ears. 

TODO: we might get rid of the scramOutput.log files in favor of the wmagentJob.log (?) they usually have no content at all.


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Hmm, this works without any issue... there must be something with that other PR...

ticoann (approved):

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -69,76 +65,79 @@ def execute(self, emulator = None, **overrides):
...
+                numberOfRetries=self.step.retryCount,
+                **overrides)
+
+        # Now we need to find all the reports
+        # The log search follows this structure: ~pilotArea/jobArea/WMTaskSpaceArea/StepsArea
+        # Start looking at the pilot scratch area first, such that we find the condor logs
+        # Then look at the job area in order to find the wmagentJob log
+        # Finally, at the taskspace area to find the cmsRun/FWJR/PSet files
+        pilotScratchDir = os.path.join(self.stepSpace.taskSpace.location, '../../')
+        logFilesForTransfer = self.findFilesInDirectory(pilotScratchDir, matchFiles, ignoredDirs)
```
ticoann: logFilesForTransfe - Is that the files archived?

ticoann:
I can see how condor.stderr gets all the log as well. If not we should propagate wmagentJob.log to agent and parse that to post in wmstats.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
amaltaro: list of files to be archived in the tarball. Let me update it.

amaltaro:
I'm probably missing something, because this ad
`ad['TransferOutput'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]`

which makes HTCondor to create those files comma separated and transfer them back ON_EXIT, is actually creating ONE file only `Report.0.pkl,wmagentJob.log`
when it's supposed to create TWO files, `Report.0.pkl` and `wmagentJob.log`

@bbockelm am I missing something?

amaltaro:
Looking at the CRABServer code, they use this code
`classad.quote(','.join(outputFiles))`

which makes the job classad in condor look like
`TransferOutput = "\"Report.0.pkl,wmagentJob.log\""`

which creates files named as `"Report.0.pkl` and `wmagentJob.log"`...

ticoann:
test this please

amaltaro:
test this please

bbockelm:
@amaltaro - do you have an example of the generated ClassAd in the schedd log?  Particularly, what's in the ClassAd for the case where you try this:

```
ad['TransferOutput'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]
```

That looks like it'd be the correct behavior, assuming `ad` is of type `classads.ClassAd` (and not a dictionary...).

amaltaro:
@bbockelm Brian, I patched one of the testbed agents once again and assigned a new workflow:
```
775.9  1  493 /data/srv/wmagent/v1.1.10.pre5/install/wmagent/JobCreator/JobCache/amaltaro_MonteCarlo_LHE_Jan2018_Validation_180205_221509_5156/Production/JobCollection_504_0/job_1784/condor.775.9.log T2_CH_CERN
cmst1@vocms0262:/data/srv/wmagent/current $ condor_q -af:h TransferOutput
TransferOutput             
Report.0.pkl,wmagentJob.log
...
cmst1@vocms0262:/data/srv/wmagent/current $ condor_q -l 775.9 | grep TransferOutput
TransferOutput = "Report.0.pkl,wmagentJob.log"
```

Let me know if this log above or the condor id is not enough for you.

And yes, we convert ads from python dict to classAd type:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L641

amaltaro:
Sigh... I see both logs back in the schedd now. I have no clue of what happened in my previous tests, because this was the first one I tried out and it did not work! Anyways, I'm updating this patch and then we can follow up with additional tests during the agent validation.

amaltaro:
@ticoann please review again and merge it if it's fine.

amaltaro:
I confirm this `TransferOutput` is still doing weird stuff, likely buggy...

This is the code put in place in this PR:
`ad['TransferOutput'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]`

thus, condor is supposed to:
a) create these 2 files once the payload starts (I noticed a small delay btw)
b) transfer them back to the schedd, once the job has completed.

Looking at their creation (inside the worker node), this was the file created (as can be seen in every single logArchive tarball now):
`../../Report.0.pkl,wmagentJob.log` 

while the files copied back are just fine (and there are two :)):
```
Job_4/wmagentJob.log
Job_4/Report.0.pkl
```

@bbockelm Brian, does it smell like a condor bug to you? Or is it that we missed a subtle detail?

bbockelm:
@amaltaro - I think your step (a) is wrong above.  HTCondor doesn't create the files once the payload starts.  It appears that this is something custom in WMCore's bootstrap script:

https://github.com/dmwm/WMCore/blob/b1270ddd45e88f80e671c027224b31ac8c37f40f/etc/submit.sh#L61-L76

Probably that needs to get updated?  I guess that WMCore has never encountered jobs with more than one output file before?

amaltaro:
I was pretty sure I had read somewhere that those files would be automatically created by HTCondor...
Yes, I think you're write, we need to update the submit.sh to properly handle TransferOutput, good catch!

bbockelm:
I think what you remember is that the schedd will create the output files by default on the submit side (particularly when the command-line-client is used).  This was added about a decade back to give people an early heads-up if they tried to write the output to, for example, an AFS path that the schedd has no permissions to write into.

PR #8426, opened 2018-01-29
Limit long jobs by estimated time limit instead of events per lumi

amaltaro:
Complement to #8398

I made the same changes to EventAwareLumiByWork algo and removed `max_events_per_lumi` from the specs. Tests still to be fixed.


src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/EDITABLE_SPLITTING_PARAM_CONFIG.py
```diff
@@ -5,7 +5,8 @@
              "LumiBased": {"lumis_per_job": '',
                           "halt_job_on_file_boundaries": True},
              "EventAwareLumiBased": {"events_per_job": '',
-                                     "max_events_per_lumi": '',
+                                     "max_events_per_lumi": '',  # has to be removed in ~HG1804
+                                     "job_time_limit": '',
```
amaltaro: @ticoann this is something I'm not sure... requests created before this patch will have max_events_per_lumi, so we should still allow it to be changed, no?
I have to check what happens if we set max_events_per_lumi parameter for a new request (thus only job_time_limit set)(?) I guess nothing, but :-)

src/python/WMCore/WMSpec/StdSpecs/ReReco.py
```diff
@@ -161,8 +161,6 @@ def __call__(self, workloadName, arguments):
                 skimConfig["SkimJobSplitArgs"]["files_per_job"] = int(arguments.get("SkimFilesPerJob%s" % skimIndex, 1))
             elif skimConfig["SkimJobSplitAlgo"] == "EventBased" or skimConfig["SkimJobSplitAlgo"] == "EventAwareLumiBased":
                 skimConfig["SkimJobSplitArgs"]["events_per_job"] = int(arguments.get("SkimEventsPerJob%s" % skimIndex, int((8.0 * 3600.0) / skimConfig["TimePerEvent"])))
-                if skimConfig["SkimJobSplitAlgo"] == "EventAwareLumiBased":
-                    skimConfig["SkimJobSplitAlgo"]["max_events_per_lumi"] = 20000
```
amaltaro: Now I see why we were setting a default value during request creation. That's because we use the "splitting" section when retrieving the job splitting (at least via web UI) and only those attributes found are shown in the web UI. I'm updating this PR

amaltaro:
I did several tests in my VM with requests before the patch, after the patch, changing their splitting via web UI and everything looks fine. I didn't test with this patch applied to the agent though, but I think that is not really important at this stage.

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
```diff
@@ -64,7 +66,7 @@ def algorithm(self, *args, **kwargs):
         # TODO: split across sites two different jobs will read the same lumi
 
         avgEventsPerJob = int(kwargs.get('events_per_job', 5000))
-        lumiEventLimit = int(kwargs.get('max_events_per_lumi', 20000))
+        jobTimeLimit = int(kwargs.get('job_time_limit', self.defaultJobTimeLimit))
```
ticoann: We are not using this plug in now but I guess no harm to change it.

src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/EDITABLE_SPLITTING_PARAM_CONFIG.py
ticoann: Once agents are updated "max_events_per_lumi" will be ignored and if "job_time_limit" is set for old agent that will be ignored as well. The question is whether "max_events_per_lumi" is set always?

ticoann:
change looks fine. Only thing is if this updated first before agent are updated. "max_events_per_lumi" will be all set to default value. Is that desirable?

src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/EDITABLE_SPLITTING_PARAM_CONFIG.py
amaltaro: It seems to be set in every single spec file, during the request creation.

amaltaro:
Good point :( We'll probably have to create a patch for 1.1.8 (and the next release) while we have these hybrid requests in the system, something like:
```
if not kwargs.get('max_events_per_lumi'):
    jobTimeLimit = int(kwargs.get('job_time_limit', self.defaultJobTimeLimit))
etc
```

amaltaro:
test this please

amaltaro:
test this please

src/python/WMCore/JobSplitting/EventAwareLumiByWork.py
amaltaro: Yes, I want to start evaluating this algo. So I think it's a good idea to keep them consistent ;)

amaltaro:
test this please

amaltaro:
jenkins seem to be getting stuck at `Looking for dmwm/WMCore issue 8426`...
I'll just trust the last jenkins run and merge it in order to cut a new release for Lina

Issue #7631, opened 2017-02-16
quantize MaxWallTimeMins


juztas:
I noticed on cms-gwmsmon, that some of new jobs which come with expression:
```
MaxWallTimeMins = WMCore_ResizeJob ? ( EstimatedSingleCoreMins / RequestCpus + 15 ) : OriginalMaxWallTimeMins
```
are not quantized by htcondor and this makes a lot of autoclusters. I might be wrong, but matching looks at MaxWallTimeMins and this makes too many different autoclusters. Maybe quantize to nearest 100?

@bbockelm @vlimant

bbockelm:
Good idea!

Should be able to wrap that to a call to quantize.  I think it's like:

quantize( ...., 180)

[seems that nearest 3 hours should be sufficient, no?]

ericvaandering:
Would it make more sense to move all this quantize stuff into WMCore?

bbockelm:
?

This is the WMCore issue tracker, so I hope that was implied...

ericvaandering:
Well, we don't have a function quantize in WMCore and as far as I know, things like CPU time and memory are quantized in HTCondor, right?

bbockelm:
ah - now I understand your comment.  I thought you meant that the code that produced the `quantize` call in HTCondor should be part of WMCore.   I see you mean that WMCore should do the quantization.

I'm not sure we can quantize within WMCore because the CPU time and memory are a function of the number of CPUs allocated (a runtime decision).

ericvaandering:
OK. We also don't have any references to the HTCondor function named quantize, but I agree that's the thing to do here. @bbockelm you want to make the PR?

ticoann:
> but I agree that's the thing to do here

Sorry I might misunderstood. If quantize is not the simple mod function, all there required parameters accessible  in WMAgent before job submission?

amaltaro:
Fixed by https://github.com/dmwm/WMCore/pull/8429

PR #8429, opened 2018-01-30
Quantize job time and disk usage

### LINKED ISSUE #8174: OriginalMaxWallTimeMins and EstimatedSingleCoreMins should be quantized
Is there any major drawback in having these parameters quantized (say 30 min) ?

amaltaro:
Fixes #8174 

First commit contains the utilitarian function, second contains the actual changes to the job factory.
We could quantize the memory usage too (by 250MB?), but then we need to make sure the MaxRSS follows the same rule (even though the global pool is broken already with its automatic rounding...).

### DIFF
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -16,6 +16,7 @@ except ImportError:
     import pickle
 
 from Utils.Timers import timeFunction
+from Utils.MathUtils import quantize
 from WMComponent.JobCreator.CreateWorkArea import CreateWorkArea
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.DAOFactory import DAOFactory
@@ -108,6 +109,10 @@ def capResourceEstimates(jobGroups, constraints):
 
             j['estimatedJobTime'] = min(j['estimatedJobTime'], constraints['MaxWallTimeSecs'])
             j['estimatedDiskUsage'] = min(j['estimatedDiskUsage'], constraints['MaxRequestDiskKB'])
+
+            # finally, quantize those
+            j['estimatedJobTime'] = quantize(j['estimatedJobTime'], constraints['MinWallTimeSecs'])
+            j['estimatedDiskUsage'] = quantize(j['estimatedDiskUsage'], constraints['MinRequestDiskKB'])
 
     return
 
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -187,9 +187,9 @@ config.JobCreator.jobCacheDir = config.General.workDir + "/JobCache"
 config.JobCreator.defaultJobType = "Processing"
 config.JobCreator.workerThreads = 1
 # glidein restrictions used for resource estimation (per core)
-config.JobCreator.GlideInRestriction = {"MinWallTimeSecs": 1 * 60 * 60, "MaxWallTimeSecs": 45 * 60 * 60,
-                                        # pilot lifetime is usually 48h
-                                        "MinRequestDiskKB": 1 * 1024 * 1024,
+config.JobCreator.GlideInRestriction = {"MinWallTimeSecs": 1 * 3600,  # 1h
+                                        "MaxWallTimeSecs": 46 * 3600,  # pilot lifetime is usually 48h
+                                        "MinRequestDiskKB": 1 * 1024 * 1024,  # 1GB
                                         "MaxRequestDiskKB": 20 * 1024 * 1024}  # site limit is ~27GB
 config.component_("JobSubmitter")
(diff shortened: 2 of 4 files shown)


ticoann:
Thanks Alan, I was creating a patch quantizing directly, but yours looks much better. (To have separate quantize function). I have one comment on the code.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
```diff
@@ -109,6 +110,10 @@ def capResourceEstimates(jobGroups, constraints):
             j['estimatedJobTime'] = min(j['estimatedJobTime'], constraints['MaxWallTimeSecs'])
             j['estimatedDiskUsage'] = min(j['estimatedDiskUsage'], constraints['MaxRequestDiskKB'])
 
+            # finally, quantize those
+            j['estimatedJobTime'] = quantize(j['estimatedJobTime'], constraints['MinWallTimeSecs'])
```
ticoann: Is this enough?  I thought "MaxWallTimeMins" has to be quantized.
```
MaxWallTimeMins = WM
Core_ResizeJob ? ( EstimatedSingleCoreMins / RequestCpus + 15 ) : OriginalMaxWallTimeMins
```
I mean  due to RequestCpus and '+15', it won't have the same quantization you intended.
I am not sure where the 15 comes form if we want to quantize 100's, that should be just removed.

ticoann:
test this please

src/python/WMComponent/JobCreator/JobCreatorPoller.py
amaltaro: I think it's fine. MaxWallTimeMins is built on top of `estimatedJobTime` and the most important is that all jobs reaching the plugin are quantized already. As you noted, it will vary according to the number of cpus, but at least it will always use the same discrete value range.

vlimant: j['estimatedJobTime'] = quantize(j['estimatedJobTime'], constraints['MinWallTimeSecs'])

is quantizing the estimatedJobTime by MinWallTimeSecs= 1 * 3600 , 1h ; sounds a bit steep. Why not adding further parameters WallTimeSecsQuanta ?

vlimant:
can you please explain the reasoning of putting the quantization in the job creation rather than in the submit plugins ? job parameters can be anything we need internally, only when sending to condor do you need this to be quantized, right ?

src/python/WMComponent/JobCreator/JobCreatorPoller.py
amaltaro: I just wanted to make this code more readable and maintainable. 1h seems quite reasonable to me. Maybe a feedback from the SI team would be better @aperezca @bbockelm

amaltaro:
If we apply the quantization at plugin level, then we need to do the same operation over and over to every single job every time we need to refresh the JobSubmitter cache. Thus, it's going to be much more efficient to have it at JobCreator level.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
ticoann: Hmm, according to @juztas post it seem MaxWallTimeMins is used for matching  #7631
So shouldn't that be quantized? Although estimatedJobTime is quantized, won't it change granularity due to division below. (EstimatedSingleCoreMins / RequestCpus + 15) Also I am not sure + 15 is necessary if the quantization is bigger than that.

amaltaro: The only way to get that quantized as well would be using some HTCondor rounding/quantization. AFAIK it has a different algorithm and it isn't very efficient.

I think quantizing the estimatedJobTime will decrease the number of auto-cluster by a lot already (in theory from 2880 to 48 combinations).

Maybe @bbockelm can share his thoughts in here...

bbockelm: Note that `MaxWallTimeMins` is an expression that is evaluated on the fly, not a value.  You can't quantize a priori -- HTCondor has to re-evaluate it.

Also, in terms of clusters and matchmaking, the idle job are all identical until *after* matchmaking is done (when the real value of CPUs is filled in).  So we can effectively ignore all autoclusters that only contain running jobs.

amaltaro:
test this please

amaltaro:
test this please

src/python/WMComponent/JobCreator/JobCreatorPoller.py
bbockelm: Given how much of an approximation all this scheduling is, 1 hour is probably as accurate as I would believe!  I would also accept 2 hours... I suspect there's decreasing returns for higher values.

amaltaro:
test this please

amaltaro:
test this please

ticoann (approved):

PR #8438, opened 2018-02-05
add to filter out missing parent files

ticoann:
I will update to have options in the command line. This is just for quick testing.


ticoann:
replaced by #8593

PR #8440, opened 2018-02-06
Enforce primary dataset to start with a letter - Lexicon

amaltaro:
As reported by Antanas @anorkus , we allow two different regex for the primary dataset.
Questions are:
1) can the PD start with a number (or dot, dash, underscore)?
2) is there any reason to keep both regex (maybe some legacy data dating back to DBS2?)?

@belforte @ericvaandering @yuyiguo @ticoann maybe you know it better, since there was some discussion back a few years in this PR https://github.com/dmwm/WMCore/pull/4973/files

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -19,17 +19,17 @@ from WMCore.WMException import WMException, WMEXCEPTION_START_STR, WMEXCEPTION_E
 # restriction enforced by DBS. for different types blocks.
 # It could have a strict restriction
 # i.e production should end with v[number]
-PRIMARY_DS = {'re': '[a-zA-Z0-9\.\-_]+', 'maxLength': 99}
+PRIMARY_DS = {'re': '^[a-zA-Z][a-zA-Z0-9\-_]*$', 'maxLength': 99}
 PROCESSED_DS = {'re': '[a-zA-Z0-9\.\-_]+', 'maxLength': 199}
 TIER = {'re': '[A-Z\-_]+', 'maxLength': 99}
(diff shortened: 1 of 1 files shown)


amaltaro:
BTW. I'm not fixing those unittests until I have a word on whether it makes sense or not.

ericvaandering:
It would be interesting to see what is in DBS, but I don't think primary or secondary should be allowed to start with ., -, _ for sure and perhaps not a number.

amaltaro:
DBS is using this same function (so both regex ATM):
https://github.com/dmwm/DBS/blob/34efdec694c31c5084abb2ebf3358f134a96ea23/Server/Python/src/dbs/utils/DBSInputValidation.py#L87

yuyiguo:
DBS does not allow to have number, . or _ as the starting part of any names, such as dataset, block or lfn. What Alan changed is not affect to DBS because the primdataset function will do the same thing before and after the patch. DBS checks the primary ds first so the block checking later will not affect the ds.

ticoann:
I think the change is fine. It seems we don't allow primary dataset name start with 0 either.  I think unittest needs to be fixed though. 😄

amaltaro:
I was pretty sure there was a logical OR between those checks ... indeed it won't change anything. I'll fix the unit tests tomorrow and this will go in in the March cmsweb deployment then.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
YESS!!!!!!!!!

ticoann:
So Jenkins works. :-)

ticoann (approved):

PR #8444, opened 2018-02-08
Change requestDisk rounding to base 10

amaltaro:
Just to keep aligned with the current schedd rounding (though that will be removed). And because it looks more clear.


ticoann:
@amaltaro, Alan patch looks good to me. Can I merge? You don't have to create the wmagent branch separately. I will create a branch with this.

amaltaro:
I think we can wait to see whether more patches will be needed and then backport them altogether. It's likely better to branch off of patch1 too (you see, that's why I didn't want to create the wmagent branch ;))

PR #8452, opened 2018-02-13
Replace DEFAULT reqmgr aux docs when requested

### LINKED ISSUE #8450: ReqMgr Aux default config docs have to be updated when there are changes
I was wondering why `job_time_limit` wasn't showing up in the ReqMgr2 UI (in testbed) and why splitting update wasn't working either. The reason is that (some) data structs from 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/

don't get posted after every ReqMgr deployment (or restart). I was going to say we need to push those docs in during the service restart, but it might not be a good idea for all those structs (like CMSSW).

We need to discuss how to address it. Meanwhile, just be aware those changes have to be manually put in reqmgr_aux db

amaltaro:
Fixes #8450

If someone makes a request to put the DEFAULT documents to ReqMgr Aux, just delete and create a new document to make sure all the up-to-date config is there.

These are the deployment changes: https://github.com/dmwm/deployment/pull/585
such that we put these documents in the post step of the reqmgr2 deployment.

### DIFF
--- a/bin/reqmgr-put-default-config
+++ b/bin/reqmgr-put-default-config
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+
+"""
+Helper script called via manage script used to update the default configuration
+documents in the ReqMgr Auxiliary DB.
+
+It's supposed to be called at every ReqMgr2 deployment, such that the splitting,
+permissions, etc are always up-to-date
+"""
+from __future__ import print_function, division
+
+import httplib
+import json
+import os
+import sys
+
+if len(sys.argv) != 2:
+    print("Usage: python reqmgr-put-default-config <hostname>")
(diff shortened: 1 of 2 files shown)


ticoann:
The reason it wasn't replacing Default config for existing config was because this supposed to be used in run time (Ops can change when it is necessary) and config can be added and removed temporary. But in case this doesn't have to be changed often, I think we can always update Default script and replace it when new deployment happen.

ticoann:
test this please

ticoann (approved):

ticoann:
Patch looks good to me, Alan

PR #8456, opened 2018-02-13
Transfer logarchive to eos and made the link on WMStats Job Detail view

ticoann:
There are a few things I want to discuss.

1. Should we make configurable which eos area point to.
2. What should be exact directory structure (I am not completely sure how we get the agent name at this stage)
3. Should attach we put pfn/url in a job report?

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -220,6 +220,7 @@ class Report(object):
         jsonReport["fallbackFiles"] = self.getAllFallbackFiles()
         jsonReport["Campaign"] = self.getCampaign()
         jsonReport["PrepID"] = self.getPrepID()
+        jsonReport["EOSLogURL"] = self.getLogURL()
 
         for stepName in self.listSteps():
             reportStep = self.retrieveStep(stepName)
@@ -1521,3 +1522,17 @@ class Report(object):
                   "JobFeatures": getattr(self.data, 'jobFeatures', {})}
 
         return wnInfo
+
+    def setLogURL(self, url):
+        """
+        Set log url for the this job report.
+        https://eoscmshttp.cern.ch/store/logs/prod/recent/
+        """
+        self.data.logURL = url
+
+    def getLogURL(self):
+        """
+        _getLogURL_
+        Return the log URL
+        """
+        return getattr(self.data, 'logURL', '')
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -229,9 +229,9 @@ class Report(object):
 
             stepTimes = self.getTimes(stepName)
 
-            if stepTimes["startTime"] != None:
+            if stepTimes["startTime"] is not None:
                 stepTimes["startTime"] = int(stepTimes["startTime"])
-            if stepTimes["stopTime"] != None:
+            if stepTimes["stopTime"] is not None:
                 stepTimes["stopTime"] = int(stepTimes["stopTime"])
 
             jsonStep["start"] = stepTimes["startTime"]
--- a/src/python/WMCore/JobStateMachine/ChangeState.py
+++ b/src/python/WMCore/JobStateMachine/ChangeState.py
@@ -416,6 +416,8 @@ class ChangeState(WMObject, WMConnectionBase):
                                   "site": job.get("location", None),
                                   "cms_location": job["fwjr"].getSiteName(),
                                   "exitcode": job["fwjr"].getExitCode(),
+                                  "eos_log_url": job["fwjr"].getLogURL(),
+                                  "worker_node_info": job["fwjr"].getWorkerNodeInfo(),
                                   "errors": errmsgs,
                                   "lumis": inputs,
                                   "outputdataset": outputDataset,
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
@@ -12,7 +12,6 @@ import re
 import signal
 import tarfile
 import time
-import traceback
 
 import WMCore.Storage.FileManager
 import WMCore.Storage.StageOutMgr as StageOutMgr
@@ -47,6 +46,56 @@ class LogArchive(Executor):
         logging.info("Steps.Executors.LogArchive.pre called")
         return None
 
+    def sendLogToEOS(self, overrides, tarBallLocation, useNewStageOutCode):
+        """
+        :param overrides: dictionary for setting the eos lfn.
+        :return: None (copy log archive to eos location.)
+        Don'f fail the job if copy fails
+        """
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = "xrdcp"
+        eosStageOutParams['option'] = "--wma-disablewriterecovery"
+        eosStageOutParams['phedex-node'] = overrides.get('eos-phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = overrides.get('eos-lfn-prefix',
+                                                        "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent")
+
+        if not eosStageOutParams['lfn-prefix']:
+            # if overrides for eos-lfn-prefix is set to None or "", don't copy the log to eos
+            return
+
+        if not useNewStageOutCode:
+            # old style
+            eosmanager = StageOutMgr.StageOutMgr(**eosStageOutParams)
+            eosmanager.numberOfRetries = self.step.retryCount
+            eosmanager.retryPauseTime = self.step.retryDelay
+        else:
+            # new style
+            logging.info("LOGARCHIVE IS USING NEW STAGEOUT CODE For EOS Copy")
+            eosmanager = WMCore.Storage.FileManager.StageOutMgr(
+                retryPauseTime=self.step.retryDelay,
+                numberOfRetries=self.step.retryCount,
+                **eosStageOutParams)
+
+        eosFileInfo = {'LFN': self.getEOSLogLFN(),
+                       'PFN': tarBallLocation,
+                       'PNN': None,
+                       'GUID': None
+                       }
+
+        try:
+            eosmanager(eosFileInfo)
+            eosServerPrefix = eosStageOutParams['lfn-prefix'].replace("root://eoscms.cern.ch//eos/cms",
+                                                                      "https://eoscmshttp.cern.ch")
+            self.report.setLogURL(eosServerPrefix + eosFileInfo['LFN'])
+            self.saveReport()
+        except Alarm:
+            msg = "Indefinite hang during stageOut of logArchive to EOS, ignoring it"
+            logging.error(msg)
+        except Exception as ex:
+            logging.exception("EOS copy failed, lfn: %s. Error: %s", eosFileInfo['LFN'], str(ex))
+
+        return
+
     def execute(self, emulator=None, **overrides):
         """
         _execute_
@@ -58,6 +107,7 @@ class LogArchive(Executor):
             return emulator.emulate(self.step, self.job)
 
         overrides = {}
+        #TODO need to set override using addOverride method in WMStep
         if hasattr(self.step, 'override'):
             overrides = self.step.override.dictionary_()
 
@@ -78,6 +128,7 @@ class LogArchive(Executor):
             "^PSet.pkl$",
             "_condor_std*",  # condor wrapper logs at the pilot top level
         ]
+
         ignoredDirs = ['Utils', 'WMCore', 'WMSandbox']
 
         # Okay, we need a stageOut Manager
@@ -126,9 +177,9 @@ class LogArchive(Executor):
                     'PNN': None,
                     'GUID': None
                     }
-
         signal.signal(signal.SIGALRM, alarmHandler)
         signal.alarm(waitTime)
+
         try:
             manager(fileInfo)
             self.report.addOutputModule(moduleName="logArchive")
@@ -142,22 +193,24 @@ class LogArchive(Executor):
             msg = "Indefinite hang during stageOut of logArchive"
             logging.error(msg)
             self.report.addError(self.stepName, 60404, "LogArchiveTimeout", msg)
-            self.report.persist("Report.pkl")
+            self.saveReport()
             raise WMExecutionFailure(60404, "LogArchiveTimeout", msg)
         except WMException as ex:
             self.report.addError(self.stepName, 60307, "LogArchiveFailure", str(ex))
-            self.report.persist("Report.pkl")
+            self.saveReport()
             raise ex
         except Exception as ex:
             self.report.addError(self.stepName, 60405, "LogArchiveFailure", str(ex))
-            self.report.persist("Report.pkl")
+            self.saveReport()
             msg = "Failure in transferring logArchive tarball\n"
-            msg += str(ex) + "\n"
-            msg += traceback.format_exc()
-            logging.error(msg)
+            logging.exception(msg)
             raise WMException("LogArchiveFailure", message=str(ex))
-
         signal.alarm(0)
+
+        signal.alarm(waitTime)
+        self.sendLogToEOS(overrides, tarBallLocation, useNewStageOutCode)
+        signal.alarm(0)
+
         return
 
     def post(self, emulator=None):
@@ -224,5 +277,20 @@ class LogArchive(Executor):
             LFN = "%s/%s/%s/%i/%s-%i-%s" % (self.altLFN, self.task.getPathName(),
                                             lfnGroup(self.job), self.job.get('retry_count', 0), self.job["name"],
                                             self.job.get('retry_count', 0), tarName)
+
+        return LFN
+
+    def getEOSLogLFN(self):
+        """
+        getEOSLogLFN
+
+        LFNs are messy, do the messy stuff here
+         /eos/cms/store/logs/prod/recent/vlimant_task_TOP-RunIIFall17wmLHEGS-00073__v1_T_180219_102942_9978/TOP-RunIIFall17DRPremix-00053_0/vocms0251.cern.ch-3803-0.log.tar.gz
+        """
+        taskPath = self.task.getPathName().split("/")
+        requestName = taskPath[1]
+        lastTask = taskPath[-1]
+        LFN = "/%s/%s/%s-%s-%s-log.tar.gz" % (requestName, lastTask, self.job.get("agentName",'NA'),
+                                      self.job["id"], self.job.get('retry_count', 0))
 
         return LFN
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
@@ -100,7 +100,6 @@ class LogArchive(Executor):
         """
         _execute_
 
-
         """
         # Are we using emulators again?
         if emulator is not None:
@@ -127,7 +126,7 @@ class LogArchive(Executor):
             "^PSet.py$",
             "^PSet.pkl$",
             "_condor_std*",  # condor wrapper logs at the pilot top level
-        ]
+            ]
 
         ignoredDirs = ['Utils', 'WMCore', 'WMSandbox']
 
--- a/src/couchapps/WMStats/_attachments/js/Views/HTMLList/WMStats.JobDetailList.js
+++ b/src/couchapps/WMStats/_attachments/js/Views/HTMLList/WMStats.JobDetailList.js
@@ -130,7 +130,10 @@ WMStats.namespace('JobDetailList');
                 htmlstr += "</ul>";
             }
             htmlstr += "</li>";
-            htmlstr += "<li>" + WMStats.Utils.expandFormat(jobDoc.logArchiveLFN, "log archive", logArchiveFormat) + "</li>"; 
+            htmlstr += "<li>" + WMStats.Utils.expandFormat(jobDoc.logArchiveLFN, "log archive", logArchiveFormat) + "</li>";
+            if (jobDoc.eos_log_url) {
+                htmlstr += '<li><b>EOS log URL:</b><a href="' + jobDoc.eos_log_url + '"> download log </a></li>';
+            }
             htmlstr += "</ul>";
             htmlstr += "</div>";
         }
--- a/src/couchapps/WMStats/_attachments/js/minified/import-all-t0.min.js
+++ b/src/couchapps/WMStats/_attachments/js/minified/import-all-t0.min.js
@@ -2727,7 +2727,10 @@ WMStats.namespace('JobDetailList');
                 htmlstr += "</ul>";
             }
             htmlstr += "</li>";
-            htmlstr += "<li>" + WMStats.Utils.expandFormat(jobDoc.logArchiveLFN, "log archive", logArchiveFormat) + "</li>"; 
+            htmlstr += "<li>" + WMStats.Utils.expandFormat(jobDoc.logArchiveLFN, "log archive", logArchiveFormat) + "</li>";
+            if (jobDoc.eos_log_url) {
+                htmlstr += '<li><b>EOS log URL:</b><a href="' + jobDoc.eos_log_url + '"> download log </a></li>';
+            }
             htmlstr += "</ul>";
             htmlstr += "</div>";
         }
--- a/src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
+++ b/src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
@@ -2870,7 +2870,10 @@ WMStats.namespace('JobDetailList');
                 htmlstr += "</ul>";
             }
             htmlstr += "</li>";
-            htmlstr += "<li>" + WMStats.Utils.expandFormat(jobDoc.logArchiveLFN, "log archive", logArchiveFormat) + "</li>"; 
+            htmlstr += "<li>" + WMStats.Utils.expandFormat(jobDoc.logArchiveLFN, "log archive", logArchiveFormat) + "</li>";
+            if (jobDoc.eos_log_url) {
+                htmlstr += '<li><b>EOS log URL:</b><a href="' + jobDoc.eos_log_url + '"> download log </a></li>';
+            }
             htmlstr += "</ul>";
             htmlstr += "</div>";
         }
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -120,7 +120,7 @@ def capResourceEstimates(jobGroups, constraints):
 def saveJob(job, workflow, sandbox, wmTask=None, jobNumber=0,
             owner=None, ownerDN=None, ownerGroup='', ownerRole='',
             scramArch=None, swVersion=None, agentNumber=0, numberOfCores=1,
-            inputDataset=None, inputDatasetLocations=None, allowOpportunistic=False):
+            inputDataset=None, inputDatasetLocations=None, allowOpportunistic=False, agentName=''):
     """
     _saveJob_
 
@@ -136,6 +136,7 @@ def saveJob(job, workflow, sandbox, wmTask=None, jobNumber=0,
 
     job['counter'] = jobNumber
     job['agentNumber'] = agentNumber
+    job['agentName'] = agentName
     cacheDir = job.getCache()
     job['cache_dir'] = cacheDir
     job['owner'] = owner
@@ -181,6 +182,7 @@ def creatorProcess(work, jobCacheDir):
         inputDataset = work.get('inputDataset', None)
         inputDatasetLocations = work.get('inputDatasetLocations', None)
         allowOpportunistic = work.get('allowOpportunistic', False)
+        agentName = work.get('agentName', '')
 
         if ownerDN is None:
             ownerDN = owner
@@ -221,7 +223,8 @@ def creatorProcess(work, jobCacheDir):
                     numberOfCores=numberOfCores,
                     inputDataset=inputDataset,
                     inputDatasetLocations=inputDatasetLocations,
-                    allowOpportunistic=allowOpportunistic)
+                    allowOpportunistic=allowOpportunistic,
+                    agentName=agentName)
 
     except Exception as ex:
         # Register as failure; move on
@@ -371,6 +374,7 @@ class JobCreatorPoller(BaseWorkerThread):
         self.defaultJobType = config.JobCreator.defaultJobType
         self.limit = getattr(config.JobCreator, 'fileLoadLimit', 500)
         self.agentNumber = int(getattr(config.Agent, 'agentNumber', 0))
+        self.agentName = getattr(config.Agent, 'hostName', '')
         self.glideinLimits = getattr(config.JobCreator, 'GlideInRestriction', None)
 
         # initialize the alert framework (if available - config.Alert present)
@@ -608,6 +612,7 @@ class JobCreatorPoller(BaseWorkerThread):
                     tempDict['scramArch'] = wmTask.getScramArch()
                     tempDict['jobNumber'] = jobNumber
                     tempDict['agentNumber'] = self.agentNumber
+                    tempDict['agentName'] = self.agentName
                     tempDict['inputDatasetLocations'] = wmbsJobGroup.getLocationsForJobs()
                     tempDict['allowOpportunistic'] = allowOpport
 
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -1101,6 +1101,9 @@ class WMTaskHelper(TreeHelper):
                 stepReport = Report()
                 stepReport.unpersist(reportPath, taskStep)
                 finalReport.setStep(taskStep, stepReport.retrieveStep(taskStep))
+                logURL = stepReport.getLogURL()
+                if logURL:
+                    finalReport.setLogURL(logURL)
             else:
                 msg = "  failed to find it."
                 msg += "Files in the directory are:\n%s" % os.listdir(os.path.join(jobLocation, taskStep))
--- a/src/python/WMCore/WMSpec/WMTask.py
+++ b/src/python/WMCore/WMSpec/WMTask.py
@@ -549,9 +549,9 @@ class WMTaskHelper(TreeHelper):
         splittingParams["trustSitelists"] = self.getTrustSitelists().get('trustlists')
         splittingParams["trustPUSitelists"] = self.getTrustSitelists().get('trustPUlists')
 
-        if "runWhitelist" not in splittingParams.keys() and self.inputRunWhitelist() != None:
+        if "runWhitelist" not in splittingParams.keys() and self.inputRunWhitelist() is not None:
             splittingParams["runWhitelist"] = self.inputRunWhitelist()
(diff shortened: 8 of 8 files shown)


src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -61,6 +61,16 @@ def execute(self, emulator=None, **overrides):
         if hasattr(self.step, 'override'):
             overrides = self.step.override.dictionary_()
 
+        #TODO hard coded - should we get it from config (something like override above
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = "xrdcp"
+        eosStageOutParams['option'] = "--wma-disablewriterecovery"
+        #TODO need to gt the correct prefix and upl
+        eosStageOutParams['phedex-node'] = "T2_CH_CERN"
+        eosStageOutParams['lfn-prefix'] = "root://eoscms.cern.ch//eos/cms"
```
ticoann: We need to find the right place? Should this different from fermi agent to cern agent?

bbockelm: We currently are proposing:

```
root://eoscms.cern.ch//eos/cms/store/logs/prod/recent
```

This should be identical across the agents - but it'd be better to be able to override in an agent configuration file (where `None` indicates "do not stageout").  This way, we can quickly fix any issues post-deploy.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -129,6 +148,11 @@ def execute(self, emulator=None, **overrides):
 
         signal.signal(signal.SIGALRM, alarmHandler)
         signal.alarm(waitTime)
+        try:
+            eosmanager(fileInfo)
+        except:
+            logging.warning("EOS copy failes lfn: %s" % fileInfo['LFN'] )
```
bbockelm: Log exception?

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
ticoann: Thanks Brian, I will make it configurable then.

amaltaro: This will change once every few years, so unless it's takes less than 5min, I think it's completely fine to have it hardcoded. In addition to that, we can't make quick fixes at this stage because it goes inside the workflow sandbox (unless we patch those as well).

ticoann: I think it is relatively easy if we use existing override parameter. (Although I think it is ever set anywhere) I think we should probably set as None first, then turn on for few workflow initially and then make the default value as above, what do you think?

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
ticoann: exception would be better since we can get the traceback. @amaltaro, do you know whether logging writes here?

amaltaro: It should be going to the wmagentJob.log file (and unintentionally to _condor_stderr)

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
amaltaro: How about we leave it as it is, then we find a namespace we can use for tests - just to make sure the whole thing works - then in the end we set it to None and leave the correct configuration commented out waiting for the final green light from CERN IT?

ticoann: That would be good. Thanks Alan.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -58,9 +58,19 @@ def execute(self, emulator=None, **overrides):
             return emulator.emulate(self.step, self.job)
 
         overrides = {}
+        #TODO need to set override using addOverride method in WMStep
```
ticoann: Still need to set up the override value from spec definition. 
and set it using 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMStep.py#L139

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -58,9 +58,19 @@ def execute(self, emulator=None, **overrides):
...
             overrides = self.step.override.dictionary_()
 
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = "xrdcp"
+        eosStageOutParams['option'] = "--wma-disablewriterecovery"
+        eosStageOutParams['phedex-node'] = overrides.get('eos-phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = overrides.get('eos-lfn-prefix',
+                                            "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent")
+
+        if hasattr(self.step, 'eosParams'):
```
amaltaro: this line can go away or be commented out.

bbockelm:
Note: I got confirmation from CERN that an appropriate quota has been setup in `/store/logs/prod/recent`.  So, we should be good-to-go for the stageout location!

ticoann:
Thanks Brian, that was quick.  I wonder how much space we are allow to put there.
Very rough estimate (in case we delete the log right away when workflow is archived) in current production job load is  4 milion jobs *  100K per log (tar file) ~0.5 TB.  If we keep the log for a while we can expect a little more than that.

Also this area can be accessible through some web server, right?

ticoann:
It seems the basic test works. (*)
We still need to set
1.  directory structure for log file. 
2.  HTTPS gateway to the eoscms area. (other wise change log copy to other eos space)
3. Make log location configurable and option for not transferring log file.
4. Need to define deleting scheme (In different patch)

```
(*)
$ xrdfs root://eoscms.cern.ch/ ls -l /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/
-r-- 2018-02-26 17:10:38      164147 /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/113e6387-30ce-4302-8511-2654da27cc9a-0-0-logArchive.tar.gz
-r-- 2018-02-26 17:09:09      163733 /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/113e6387-30ce-4302-8511-2654da27cc9a-15-0-logArchive.tar.gz
-r-- 2018-02-26 17:09:10      161161 /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/113e6387-30ce-4302-8511-2654da27cc9a-16-0-logArchive.tar.gz
-r-- 2018-02-26 17:10:03      162204 /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/113e6387-30ce-4302-8511-2654da27cc9a-2-0-logArchive.tar.gz
-r-- 2018-02-26 17:10:12      162827 /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/113e6387-30ce-4302-8511-2654da27cc9a-23-0-logArchive.tar.gz
-r-- 2018-02-26 17:08:19      165166 /eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_TaskChain_ProdMinBias_sryu_180226_173330_2341/ProdMinBias/0000/0/113e6387-30ce-4302-8511-2654da27cc9a-37-0-logArchive.tar.gz
```

ticoann:
> Subtask name (TOP-RunIIFall17DRPremix-00053_0)

You want to have all the hierarch of the subtask not just last task name,  correct?

In case of file name (as well as whole path), users who looks for the particular log should be able to map from the information they have. Is HTCondor global job ID available in general?

ticoann:
I am gettin following errors today. (It seems it fail to copy log files to eoscms).  Need to investigate.

```
Exception Class: StageOutError
Message: Command exited non-zero, ExitCode:151
Output: stdout: Local File Size is: 
Remote File Size is: 218497
ERROR: Size Mismatch between local and SE

stderr: stat: cannot stat `root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_MonteCarlo_eff_sryu_180226_173312_3177/Production/0000/2/53db7876-979b-48bb-b824-4b08e53a0cb7-1-2-logArchive.tar.gz': No such file or directory
/bin/sh: line 5: [: ==: unary operator expected
 
	ErrorCode : 60311
	ModuleName : WMCore.Storage.StageOutError
	MethodName : __init__
	ErrorType : GeneralStageOutFailure
	ClassInstance : None
	FileName : /srv/job/WMCore.zip/WMCore/Storage/StageOutError.py
	ClassName : None
	Command : LOCAL_SIZE=`stat -c%s "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_MonteCarlo_eff_sryu_180226_173312_3177/Production/0000/2/53db7876-979b-48bb-b824-4b08e53a0cb7-1-2-logArchive.tar.gz"`
echo "Local File Size is: $LOCAL_SIZE"
env XRD_WRITERECOVERY=0 xrdcp --force --nopbar  "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/2/26/sryu_MonteCarlo_eff_sryu_180226_173312_3177/Production/0000/2/53db7876-979b-48bb-b824-4b08e53a0cb7-1-2-logArchive.tar.gz"  "root://eoscms.cern.ch//eos/cms/store/unmerged/logs/prod/2018/2/26/sryu_MonteCarlo_eff_sryu_180226_173312_3177/Production/0000/2/53db7876-979b-48bb-b824-4b08e53a0cb7-1-2-logArchive.tar.gz" 
REMOTE_SIZE=`xrdfs 'eoscms.cern.ch' stat '/eos/cms/store/unmerged/logs/prod/2018/2/26/sryu_MonteCarlo_eff_sryu_180226_173312_3177/Production/0000/2/53db7876-979b-48bb-b824-4b08e53a0cb7-1-2-logArchive.tar.gz' | grep Size | sed -r 's/.*Size:[ ]*([0-9]+).*/\1/'`
echo "Remote File Size is: $REMOTE_SIZE"
if [ $REMOTE_SIZE ] && [ $LOCAL_SIZE == $REMOTE_SIZE ]; then exit 0; else echo "ERROR: Size Mismatch between local and SE"; xrdfs 'eoscms.cern.ch' rm /eos/cms/store/unmerged/logs/prod/2018/2/26/sryu_MonteCarlo_eff_sryu_180226_173312_3177/Production/0000/2/53db7876-979b-48bb-b824-4b08e53a0cb7-1-2-logArchive.tar.gz ; exit 60311 ; fi
	LineNumber : 32
	ErrorNr : 0
	ExitCode : 151

Traceback: 
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutImpl.py", line 216, in __call__
    self.executeCommand(command)

  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutImpl.py", line 94, in executeCommand
    raise StageOutError(msg, Command=command, ExitCode=exitCode)
```

bbockelm:
@ticoann - I would rather just have the last subtask name.  That flattens the directory hierarchy quite a bit.

Your error message is referencing a failure when using the `stat` command line tool against the URL `root://eoscms.cern.ch/...`.  That means the stageout plugin is quite confused and trying to stat a remote file using local tools: maybe it's being fed the wrong arguments?

NOTE: the web view of EOS CMS was enabled today at https://eoscmshttp.cern.ch/store/ (I'm bugging them about the final hostname... consider that possibly transient).  You'll need to auth with your grid certificate.

@vlimant - See the question below about the filename.  From a ES-CMS record, this is sufficient to uniquely identify all attempts.  Is that enough?  Alternately, you'll be able to list the entire subdirectory from the web UI if you want to look for a particular job.

amaltaro:
Having the task name directly under the request name is going to be much easier (and cleaner), I'm just wondering whether we can have some task name clashes, like for merge tasks (I guess not...).

About the filename:
>    WMBS Job ID and start number(e.g., 1.1)
>    HTCondor global job ID (vocms0198.cern.ch#17967727.0#1519661922)
>    HTCondor start count (1).

Using this information from condor will likely require a) fetching more job ads/history and b) maybe database schema.
Why can't we use what is already in the agent schema? Like:
* WMBS Job ID (WMAgent_JobID), also in the WMBS tables or `wmbs_id` from bl_runjob
* Condor job id, `grid_id` from bl_runjob
* agent retry count, `retry_count` from bl_runjob as well.

e.g.: 3803-3638.13-0
Given the request name and the above information, we should be able to find the exact log (and the exact entry in ES-CMS, no?). In addition to that, we rely only on the persisted WMAgent information (single table).
Of course there are corner cases - for both approaches- since we have jobs that never reach the condor queue, sometimes they even failed to be created. Ohh, nevermind, we won't upload any logs to those!

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -129,6 +145,11 @@ def execute(self, emulator=None, **overrides):
 
         signal.signal(signal.SIGALRM, alarmHandler)
         signal.alarm(waitTime)
+        try:
```
amaltaro: How about we move it to the end of this method and set another alarm for the EOS copy (just logging in case any exception is raised - including Alarm)?

amaltaro (changes requested):

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
bbockelm: Using signal.alarm is dicey. Why not push the timeout logic downward?

bbockelm:
Hey Alan,

I can do without the full schedd name (likelihood of collision is very small), but we should include the condor retry number here. That allows us to see failures when there are multiple job starts per WMBS retry.

We could drop the extension for the first try, but I don’t want to override the failed attempts.

Since EOS allows us to list directories via HTTPS, we can find jobs that way.

Brian

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
amaltaro: Are you suggesting to leave the EOS transfer without any timeout (and keep it only for the local logArchive stageout)? 
We still need to have a mechanism to avoid the job hanging on this file transfer, that's why we should have two alarms IMO. We, of course, don't need to wait for 1h * retryNumber, maybe 5 or 10min would suffice.

amaltaro:
> I can do without the full schedd name (likelihood of collision is very small), but we should include the condor retry number here. That allows us to see failures when there are multiple job starts per WMBS retry.

My point was not to fetch additional condor classads. If we fetch 1, then we can fetch a few more.
However, please educate me on why internal condor retries matter? The tarball will only carry information of the last execution anyways and, we still have a unique job mapping, since a new agent retry uses a different condor job id (thus one - if will - can easily correlate ES data with the logArchive).

bbockelm:
> However, please educate me on why internal condor retries matter?

Logging condor retries allows us to figure out why the previous attempts fail, no?

Then again, most of the reasons which cause an internal retry (example: quick memory blowup causes pilot to be killed) will also prevent WMAgent from staging out the log tarball.  So maybe we can put that aside for now.

Thinking about the proposal to do `3803-3638.13-0` -- I think we should do `3803-0` or `3638.13`

So, I guess we end up with LFNs of the form:

```
/store/logs/prod/recent/vlimant_task_TOP-RunIIFall17wmLHEGS-00073__v1_T_180219_102942_9978/TOP-RunIIFall17DRPremix-00053_0/task-vocms0251.cern.ch-3803-0.log.tar.gz
```

?

Good for everyone?

ticoann:
It sounds good to me.
> task-vocms0251.cern.ch-3803-0.log.tar.gz

Do we need 'task-' prefix?

bbockelm:
@ticoann - based on a similar comment from Alan, I already removed the `task` prefix in the example.

ticoann:
One thing (I mentioned in slack) I would like to mention, This whole path doesn't guarantee to be unique.
If agent gets drained (processing workflow A), and some of the workqueue elements of A are not retrieved yet. And the same agent redeployed with new version and pulls down WorkQueue elements of A, Then there is possibility we could have duplicate path.  (I think it is very unlikely,  but it is not 0). I am fine with risking that case unless others have more concerns about that.

amaltaro:
About the sourcePFN issue, can you have a look at the condor wrapper stdout log (in JobArchiver) and check the exact files/paths used for stage out? The destination PFN doesn't seem to be correct (it has dates in there).

ticoann:
Just for the record, getting slightly different error. (sourcePFN is seems to be wrong)

```
Job Name: bd42096a-2258-11e8-9d12-02163e01b4c0-0
WMBS job id: 35
Workflow: sryu_StepChain_DupOutMod_sryu_180307_212343_2797
Task: /sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput
Status: jobfailed
Input dataset:
Site: T1_US_FNAL_Disk
Agent: vocms0193.cern.ch
ACDC URL: https://reqmgr2-dev.cern.ch/couchdb/acdcserver
State Transition
jobfailed: 2018/3/7 (Wed) 22:52:01 UTC, T1_US_FNAL
Exit code: 60307
Retry count: 0
Errors:
logArch1
LogArchiveFailure (Exit Code: 60307)
<@========== WMException Start ==========@>
Exception Class: StageOutFailure
Message: Failure for local stage out:
[Errno 2] No such file or directory: 'root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz'Traceback (most recent call last):
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutMgr.py", line 318, in localStageOut
    impl(protocol, localPfn, pfn, options, checksums)
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutImpl.py", line 208, in __call__
    command = self.createStageOutCommand(sourcePFN, targetPFN, options, checksums)
  File "/srv/job/WMCore.zip/WMCore/Storage/Backends/FNALImpl.py", line 111, in createStageOutCommand
    original_size = os.stat(sourcePFN)[6]
OSError: [Errno 2] No such file or directory: 'root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz'

	TargetPFN : root://cmsdcadisk01.fnal.gov//dcache/uscmsdisk/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz
	ErrorCode : 60311
	ModuleName : WMCore.Storage.StageOutError
	MethodName : __init__
	LFN : /store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz
	ClassInstance : None
	FileName : /srv/job/WMCore.zip/WMCore/Storage/StageOutError.py
	ClassName : None
	Command : stageout-xrdcp-fnal
	LineNumber : 32
	InputPFN : root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz
	Protocol : writexrd
	ErrorNr : 0
	ErrorType : GeneralStageOutFailure

Traceback: 
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutMgr.py", line 318, in localStageOut
    impl(protocol, localPfn, pfn, options, checksums)

  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutImpl.py", line 208, in __call__
    command = self.createStageOutCommand(sourcePFN, targetPFN, options, checksums)

  File "/srv/job/WMCore.zip/WMCore/Storage/Backends/FNALImpl.py", line 111, in createStageOutCommand
    original_size = os.stat(sourcePFN)[6]

<@---------- WMException End ----------@>

```

ticoann:
Alan, don't spend time on it. I am debugging now.

ticoann:
It seems it is woking now. I will check the log file back tomorrow. @amaltaro, Alan, you said you prefer to set Alarm, right?

ticoann:
@amaltaro, Alan I fixed the path. (finally 😄).

https://eoscmshttp.cern.ch/store/logs/prod/recent/sryu_MonteCarlo_LHE_s3yu_180319_203757_9618/Production

Also add the the path in fwjr so dashboard / wmstats can have the direct link. (we still have to prefix https://eoscmshttp.cern.ch/store/logs/prod/recent/ - didn't want to hardcoded this.)

I messed up with path and would like to clean up. Do you know the command?

On minor problem is tar file is crated with relative path starting with grand parent directory, are we always doing that?

```
mac-123267:testTar sryu$ tar xvfP vocms0193.cern.ch-1464-0-log.tar.gz
x ../../_condor_stderr
x ../../job/WMTaskSpace/stageOut1/Report.pkl
x ../../job/WMTaskSpace/cmsRun1/FrameworkJobReport.xml
x ../../job/WMTaskSpace/cmsRun1/Report.pkl
x ../../job/WMTaskSpace/cmsRun1/PSet.pkl
x ../../job/WMTaskSpace/cmsRun1/cmsRun1-stderr.log
x ../../job/WMTaskSpace/cmsRun1/PSet.py
x ../../job/WMTaskSpace/cmsRun1/cmsRun1-stdout.log
x ../../job/WMTaskSpace/cmsRun1/scramOutput.log
x ../../_condor_stdout

**Comment by amaltaro** [comment]
> @amaltaro, Alan I fixed the path. (finally 😄).

Yes, they look good now!

> Also add the the path in fwjr so dashboard / wmstats can have the direct link. (we still have to prefix https://eoscmshttp.cern.ch/store/logs/prod/recent/ - didn't want to hardcoded this.)

I'd say we can make it configurable, but honestly, we aren't going to change it more than once every couple of years (or even more). I'd just hard-code it and make the code simpler.

> I messed up with path and would like to clean up. Do you know the command?

I have no clue (and I'm not sure we have privileges), but I can ask the transfer team tomorrow. Can you please point out what can be deleted?

> On minor problem is tar file is crated with relative path starting with grand parent directory, are we always doing that?

Since a couple of months, when I added more logs into the tarball. I didn't want to make my patch dirty, but in the end every single tarball has weird file paths. Does it cause any problems in the end (besides being weird)? I can update it tomorrow.

Seangchan, I'll go through this patch tomorrow, if I review it right now, it's going to be bad :)

**Comment by ticoann** [comment]
Thanks Alan, I updated the patch with your suggestions.
1. hard coded eos log url
2. add the eos url in job report
3. update wmstats with eos url (only for failed job).

**Comment by ticoann** [comment]
About the cleaning up, any thing under https://eoscmshttp.cern.ch/store/logs/prod/recent/ can be removed but lets request after test is done. 
We also need to have clean up policy so I guess  we would need a permission (with maybe at least cmsweb service cert?), and command later.

File: src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py [hunk 6]
```diff
@@ -127,7 +143,25 @@ def execute(self, emulator=None, **overrides):
...
         signal.signal(signal.SIGALRM, alarmHandler)
+        signal.alarm(waitTime)
+        try:
+            eosmanager(eosFileInfo)
+            eosServerPrefix = eosStageOutParams['lfn-prefix'].replace("root://eoscms.cern.ch//eos/cms",
+                                                                      "https://eoscmshttp.cern.ch")
+            self.report.setLogURL(eosServerPrefix + eosFileInfo['LFN'])
+        except Alarm:
+            msg = "Indefinite hang during stageOut of logArchive to EOS, ignoring it"
+            logging.exception(msg)
```
amaltaro: I think this logging can be `logging.error`

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -127,7 +143,25 @@ def execute(self, emulator=None, **overrides):
...
+        except Exception as ex:
+            logging.exception("EOS copy failes lfn: %s" % eosFileInfo['LFN'])
```
amaltaro: How about `logging.exception("EOS copy failed for lfn: %s. Error %s", eosFileInfo['LFN'], str(ex))`?

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
amaltaro: I think this logging can be `logging.error`

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -152,9 +186,7 @@ def execute(self, emulator=None, **overrides):
             self.report.addError(self.stepName, 60405, "LogArchiveFailure", str(ex))
             self.report.persist("Report.pkl")
             msg = "Failure in transferring logArchive tarball\n"
-            msg += str(ex) + "\n"
-            msg += traceback.format_exc()
-            logging.error(msg)
+            logging.exception(msg)
```
amaltaro: +1 (the import can be removed too)

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -226,3 +258,34 @@ def getLFN(self, tarName):
                                             self.job.get('retry_count', 0), tarName)
 
         return LFN
+
+    def getEOSLogLFN(self):
+        """
+        getLFN
```
amaltaro: _getEOSLogLFN_

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -127,7 +143,25 @@ def execute(self, emulator=None, **overrides):
...
+        eosFileInfo = {'LFN': self.getEOSLogLFN(),
+                       'PFN': tarBallLocation,
+                       'PNN': None,
+                       'GUID': None
+                       }
         signal.signal(signal.SIGALRM, alarmHandler)
+        signal.alarm(waitTime)
+        try:
+            eosmanager(eosFileInfo)
+            eosServerPrefix = eosStageOutParams['lfn-prefix'].replace("root://eoscms.cern.ch//eos/cms",
```
amaltaro: Isn't `eosFileInfo` structure updated in place? I think `eosFileInfo['PFN']` will contain the final file location. We could then pass it to the `setLogURL` and let that method replace the prefix.

src/python/WMCore/JobStateMachine/ChangeState.py
```diff
@@ -416,6 +416,8 @@ def recordInCouch(self, jobs, newstate, oldstate, updatesummary = False):
                                   "site": job.get("location", None),
                                   "cms_location": job["fwjr"].getSiteName(),
                                   "exitcode": job["fwjr"].getExitCode(),
+                                  "eos_log_url": job["fwjr"].getLogURL(),
```
amaltaro: Isn't getLogURL a LogArchive method? Here you're looking for it in a Report object, no? Same for the file above.

amaltaro (changes requested):
A few minor changes and one concern pointed out in the code.

src/python/WMCore/JobStateMachine/ChangeState.py
ticoann: Sorry while I was merging with master branch, made a mistake. Yes these are Report methods.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
ticoann: Thanks I will fix that.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
ticoann: I will fix that.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
ticoann: I will update that.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
ticoann: It is not in place replace. It returns new string.  The reason I put the code here is that I don't wan t to assume the prefix in general although it won't change. I think we have to think about more where is the best place, but since it is the closest place actual location is set, I just put it here.

ticoann:
@amaltaro, Alan I changed code as you suggested.
And add a couple of things.
1. create direct link to download tar file only if url is set.
2. add the condition for eos log transfer (ugly if statements, I might refactor) - created a separate issue for spec file update #8523

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -58,9 +57,22 @@ def execute(self, emulator=None, **overrides):
...
 
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = "xrdcp"
+        eosStageOutParams['option'] = "--wma-disablewriterecovery"
+        eosStageOutParams['phedex-node'] = overrides.get('eos-phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = overrides.get('eos-lfn-prefix',
+                                            "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent")
+
+        if eosStageOutParams['lfn-prefix']:
+            setEOSLog = True
```
amaltaro: honestly, I don't like it :) setEOSLog is always True anyways. This will be - by default - enabled in new agents, so why "control" whether we upload it to EOS or not.
As you said, it's ugly and make the code more unreadable :)

ticoann: I know, but I think we discussed that we can give the hook if we don't want to transfer the file to eos for some reason. (i.e. test workflows, etc). The we we can overwrite overwriteParam value to None. so we don't put unnecessary logs there.

ticoann: If you agree this control is necessary. I will beautify the code. 😄

ticoann:
@amaltaro, Alan I refactored the code. Could you check one more time?

src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -1521,3 +1522,18 @@ def getWorkerNodeInfo(self):
                   "JobFeatures": getattr(self.data, 'jobFeatures', {})}
 
         return wnInfo
+
+    def setLogURL(self, url):
+        """
+        Set log url for the this job report.
+        https://eoscmshttp.cern.ch/store/logs/prod/recent/
+        """
+
```
amaltaro: I'd still remove this blank line :-)

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -47,6 +46,57 @@ def pre(self, emulator=None):
...
+        """
+
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = "xrdcp"
+        eosStageOutParams['option'] = "--wma-disablewriterecovery"
+        eosStageOutParams['phedex-node'] = overrides.get('eos-phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = overrides.get('eos-lfn-prefix',
+                                                        "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent")
+
+        if eosStageOutParams['lfn-prefix']:
```
amaltaro: `if not ...` ?

amaltaro (changes requested):
I think there is still one `if` that you must update :-)
If you do, remove that blank line too please.

Code looks much better now, besides that "if", I think it's good to go. In practice, I don't think we'll ever use this override, since it comes from the spec level (thus, ReqMgr2 level).

ticoann:
> I don't think we'll ever use this override, since it comes from the spec level (thus, ReqMgr2 level)

But test workflow should overwrite this, isn't it. Either not to send or different directory.

amaltaro:
It depends :) Integration tests should upload the tarball to make sure we cover most of the code. Dev tests could be dropped, but since we'll have to add things to either request creation or assignment, I think the load/space with the dev tests is negligible.

src/python/WMCore/WMSpec/Steps/Executors/LogArchive.py
```diff
@@ -47,6 +46,56 @@ def pre(self, emulator=None):
...
+            eosmanager(eosFileInfo)
+            eosServerPrefix = eosStageOutParams['lfn-prefix'].replace("root://eoscms.cern.ch//eos/cms",
+                                                                      "https://eoscmshttp.cern.ch")
+            self.report.setLogURL(eosServerPrefix + eosFileInfo['LFN'])
+            self.report.persist("Report.pkl")
+        except Alarm:
+            msg = "Indefinite hang during stageOut of logArchive to EOS, ignoring it"
+            logging.error(msg)
+        except Exception as ex:
+            logging.exception("EOS copy failes lfn: %s. Error: %s" % (eosFileInfo['LFN'], str(ex)))
```
amaltaro: Seangchan, don't hate me :-) But can you fix this pylint warning too:
 W1201, line 95 in LogArchive.sendLogToEOS: Specify string format arguments as logging function parameters 
?
basically updating this line to
`logging.exception("EOS copy failes lfn: %s. Error: %s", eosFileInfo['LFN'], str(ex))`

amaltaro: updated failes --> failed

ticoann: I hate you Alan. 😄 Actually thank you very much for the review. I always forget that.
Anyway, I still a problem of setting the url in fwjr. I will fix other pylint errors when that is fixed.

ticoann:
@amaltaro, Now it is propagating url correctly in FWJR and WMStats. I am merging it for the testbed deployment. But still please review. If still something need to be updated I will create a new patch

PR #8459, opened 2018-02-14
add log db report for db exception

ticoann:


ticoann:
test this please

amaltaro (changes requested):
Hmm, ideally we should not put any WMCore specific code in the Utils package. Do you see any other place we could add/delete the logdb record?

ticoann:
Yes I agree how about under WMCore/Database?

amaltaro:
Yes, WMCore/Database looks a perfect place

PR #8464, opened 2018-02-16
Remove remaining Teams code

### LINKED ISSUE #7355: Clean up Teams from the system
As discussed with Seangchan yesterday (and initially fixed in #7353), we're no longer supporting a workflow assigned to more than one team name. 

In addition to that, we'll also remove the support of multiple team names defined in the agent configuration. It just causes confusion about `Team` x `Teams` and nobody has used this feature during the last couple of years.

amaltaro:
Fixes #7355
Now should be safe to completely deprecate Teams and keep only string Team around.

I haven't fixed the tests yet, let's see what fails

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -784,7 +784,7 @@ class WorkQueue(WorkQueueBase):
         If resources passed in get work for them, if not available resources
         from get from wmbs.
         """
-        if self.pullWorkConditionCheck() == False:
+        if self.pullWorkConditionCheck() is False:
             return 0
 
         (resources, jobCounts) = self.freeResouceCheck(resources)
@@ -900,7 +900,7 @@ class WorkQueue(WorkQueueBase):
         requestsInfo = self.requestDB.getRequestByNames(reqNames)
         deleteRequests = []
         for key, value in requestsInfo.items():
-            if (value["RequestStatus"] == None) or (value["RequestStatus"] in deletableStates):
+            if (value["RequestStatus"] is None) or (value["RequestStatus"] in deletableStates):
                 deleteRequests.append(key)
 
         return self.backend.deleteWQElementsByWorkflow(deleteRequests)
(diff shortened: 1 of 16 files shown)


amaltaro:
test this please

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, logger=None, dbi=None, **params):
         self.params.setdefault('WMBSUrl', None)  # this will only be set on local Queue
         if self.params.get('WMBSUrl'):
             self.params['WMBSUrl'] = Lexicon.sanitizeURL(self.params['WMBSUrl'])['url']
-        self.params.setdefault('Teams', [])
+        self.params.setdefault('Team', [])
```
ticoann: I would prefer this would be string not list. But I understand that would be many more code changes.

amaltaro: I think I had a look at it today but decided not to change. I'll check that out tomorrow again, I believe it will be this file, the workRestrictions couch list and maybe WorkQueueBackend. I agree it will be much better, it's just more annoying to validate :)

ticoann: Actually it might not be that many change, (Maybe more then below).

This shouldn't be list
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueUtils.py#L133

And workRestrictions  
default string
https://github.com/dmwm/WMCore/pull/8464/files#diff-7e2ed5ba1c031d501f615be69a695be9R23
change to string comparison
https://github.com/dmwm/WMCore/pull/8464/files#diff-7e2ed5ba1c031d501f615be69a695be9R60

But I think in a long run it might be less confusing. Since current code implies Team indicates some times list, but some times string.

amaltaro:
test this please

ticoann:
Alan, patch looks good to me. Do you want me to merge it?

ticoann (approved):

amaltaro:
It looks like we need to patch all the agents under the same domain (cmsweb, testbed, etc). I had workflows assigned to testbed-vocms0192 that were acquired by two other agents :(

Issue #7003, opened 2016-07-22
Replace condor log parsing by condor history


amaltaro:
Since I'm playing with the job tracking, I decided to get rid of that nasty local log parsing - which tries quite badly - to figure out what happened to jobs that vanished from the condor queue.

So I'm using the condor bindings and the `history()` method for the Schedd class, you can see the code here:
https://github.com/amaltaro/WMCore/commit/dc573e56be82f8a57289ec095144ca9207ca6db5

which I tested interactively and I also saw it succeeding once in the agent (vocms0231).

However, it failed the second time that code was performed for one job constraint (and now it fails all the time for ~900 ids). The error reason is `Timeout when waiting for remote host`, which is raised ~20 seconds after querying the schedd.

Here is the tricky part, after googling for a minute, I found out the generator is indeed supposed to return *right away* from the history call, however it still didn't produce/yield any results. It does so when we iterate over it in the `for` loop, which is when the timeout happens.

In case it helps, I locally replaced those 2 lines of `for` (that was using the grouper function) by only a single line like (it's an iterator already anyways)
`            for jobAd in jobiter:`

and then this line is throwing an exception. Example of traceback can be seen here [1]

@ticoann @ericvaandering do you see any problem with that code?
@bbockelm is 20sec a expected delay in a small node with basically no load? is it configurable? do you see any possible optimization to that code such that we can use condor history in production?

[1]
```
2016-07-22 23:15:00,880:140496379688704:INFO:PyCondorPlugin:PyCondor is going to track 1005 jobs
2016-07-22 23:15:00,912:140496379688704:INFO:PyCondorPlugin:Retrieved 19 classAds
2016-07-22 23:15:00,913:140496379688704:INFO:PyCondorPlugin:Found 986 jobs missing in condor. Checking the history.
2016-07-22 23:15:00,928:140496379688704:INFO:PyCondorPlugin:<htcondor.HistoryIterator object at 0x7fc7df5028c0>


2016-07-22 23:15:20,949:140496379688704:ERROR:BossAirAPI:Unhandled exception while tracking jobs for plugin PyCondorPlugin!
Timeout when waiting for remote host
2016-07-22 23:15:21,026:140496379688704:ERROR:BaseWorkerThread:Error in worker algorithm (1):
Backtrace:
  <WMCore.BossAir.StatusPoller.StatusPoller instance at 0x7fc7df48a7a0> BossAirException
[...]

PR #8469, opened 2018-02-20
Bump JobAccountant polling cycle to 5min

amaltaro:
Not really necessary, but it is supposed to mitigate a race condition where the component starts processing completed jobs while the job Report.pkl hasn't been transferred back to the schedd.

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -170,7 +170,7 @@ config.JobAccountant.namespace = "WMComponent.JobAccountant.JobAccountant"
 config.JobAccountant.componentDir = config.General.workDir + "/JobAccountant"
 config.JobAccountant.logLevel = globalLogLevel
 config.JobAccountant.workerThreads = 1
-config.JobAccountant.pollInterval = 60
+config.JobAccountant.pollInterval = 300
 config.JobAccountant.specDir = config.General.workDir + "/JobAccountant/SpecCache"
 
 config.component_("JobCreator")


bbockelm:
Why not fix the race condition?

amaltaro:
Because we don't really know when the output files have been transferred back to the schedd. This goes along the htcondor job monitoring, AFAIK there is no scalable and reliable way to know whether a job gracefully completed or whether it was just removed from the pool

bbockelm:
So, there's five options here:

1.  For JobAccountant, just wait for the job to exit the queue, then look at the output files.  Shouldn't matter in terms of accounting why the job exited the queue.
2.  Enable PER_JOB_HISTORY_DIR.  The schedd will drop a file in a directory corresponding to each job it has finished.
3.  Track job status via the user log.  This can be done in a scalable manner because there's one file per job.
4.  Set the job to stay in queue for N minutes after termination so it is available via `condor_q` equivalents.
5.  Utilize the `condor_history` python API to query for the schedd's history to know why a job has been removed from the queue.  This is what is done for ElasticSearch monitoring and is scalable if you do it, say, once every N minutes from a single thread.

amaltaro:
Thanks for the options, I have some comments/questions though.
1. that's precisely what's done already. Unless "just wait for the job to exit the queue" means something different than: "if condor_q doesn't return your job, it's no longer in the queue"
2. and 3. write more data to the disk, so there will be a I/O penalty here. Would you know how much data it writes for each job? If possible, I don't want to follow any approach that will impact the scalability of the production schedds.
4. probable the simplest option, though it still doesn't guarantee the file will be there. Unless we set it beyond the htcondor timeout itself for transferring files.
5. this is the path I was going to take in the beginning of 2017 or so, when I was told that this is very expensive and not reliable. Has it improved in the newer htcondor versions? See https://github.com/dmwm/WMCore/issues/7003

It looks something simple, but it's quite tricky to get it done properly ;)

bbockelm:
Ok, I think I see the problem.

There's two reasons that the SimpleCondor plugin will mark the job as finished:
1.  The job has entered the `C` state, when HTCondor starts to clean things up.
2.  The job has left the queue and all the job completion hooks have finished (including writing out the log file entries).

I think you there's a lot of edge cases of activities that may be occurring when the job is in `C` state (which is typically less than a second, but can be longer under heavy load).  I would suggest to only transition the job to completion for case 2.

amaltaro:
Yes, these are the two conditions we move a job to completed. Plus, when it's in an error state (aka. Held/Removed/Suspended/etc), this line:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L236

perhaps we should not complete jobs on that state as well, since there might be still some transfers or final htcondor job processing?

Which finally means, we only complete jobs when they are no longer in the queue, right?

bbockelm:
* Not sure what we should do with suspended state - probably map it to running?  Should be hopefully rare as that functionality is not enabled on the global pool.
* Removed: transient state, typically indicates something is being cleaned up.  We should just wait for the job to leave queue.
* Held: This one is tough.  The job is not going to be successful.  We can always just set the job to auto-remove on `Held` state, then treat this like the others.

amaltaro:
> Not sure what we should do with suspended state - probably map it to running? Should be hopefully rare as that functionality is not enabled on the global pool.

If it's not enabled, does it mean we should never see a job in this status in the global pool (and ITB I'd assume)? If we still can see them, then I have a question, after it gets unsuspended, to which state that job is supposed to go? Or it can go to pretty much any state (Held, Running, Idle,...)?

> Removed: transient state, typically indicates something is being cleaned up. We should just wait for the job to leave queue.

agreed!

> Held: This one is tough. The job is not going to be successful. We can always just set the job to auto-remove on Held state, then treat this like the others.

We've put a PeriodicRemove expression some time ago, so we should be fine by just waiting for it to be gone from the queue: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L498

bbockelm:
@amaltaro - the `suspended` state is when a particular evaluation (which we don't set) on the `condor_startd` evaluates to true.  In that case, the processes in the job get a `SIGSTOP`.  Ideally, the job goes back to running -- but it can be preempted and/or held while in that state, so it can basically go to any other state.  I'd just ignore it and not change the job from its prior state.

Ok, sounds like we have some ideas that'll significantly improve job tracking.  Alan, can you file a ticket and follow up?

PR #8470, opened 2018-02-20
add worker_node name in wmarchive

ticoann:


src/python/WMCore/Services/WMArchive/DataMap.py
```diff
@@ -417,9 +417,13 @@ def createArchiverDoc(job, version=None):
         import WMCore
         version = WMCore.__version__
     # append meta data in fwjr
+    wnName = ""
+    if "WorkerNodeInfo" in fwjr:
```
ticoann: We probably won't patch the existing agent but just in case.

src/python/WMCore/Services/WMArchive/DataMap.py
```diff
@@ -417,9 +417,13 @@ def createArchiverDoc(job, version=None):
...
+        wnName = fwjr["WorkerNodeInfo"].get("HostName", "")
+
     newfwjr['meta_data'] = {'agent_ver': version,
                          'host': socket.gethostname().lower(),
-                         'wn_info': fwjr["WorkerNodeInfo"],
+                         'wn_name': wnName,
```
ticoann: @amaltaro, we talk about WorkerNodeName as key but it seems everything else is lower case for meta_data. So I change to wn_name, hope that is OK.

amaltaro (approved):
It looks good! Thanks

ticoann:
test seemed OK, merging it.

PR #8471, opened 2018-02-20
add some machine matrix 

### LINKED ISSUE #8468: Properly fetch Job and Machine Features
As a follow up of #8304. We need to correct how the JOBFEATURES and MACHINEFEATURES are parsed by the runtime code.
While talking to Antonio @aperezca these environment variables/files are not even enforced in CMS yet. If we want to start using them, we should probably go through a deployment/commissioning phase and enforce it (maybe through a SAM test).

FYI, paper pointed out by Antonio: http://hepsoftwarefoundation.org/notes/HSF-TN-2016-02.pdf

ticoann:
fixes #8468

### DIFF
--- a/src/python/WMCore/WMRuntime/Bootstrap.py
+++ b/src/python/WMCore/WMRuntime/Bootstrap.py
@@ -28,6 +28,12 @@ class BootstrapException(WMException):
     """ An awesome exception """
     pass
 
+def readFloatFromFile(filePath):
+    try:
+        with open(filePath, "r") as nf:
+            return float(nf.readline())
+    except:
+        return None
 
 # Copied direct from ProdAgent to find the damn CE name
 def getSyncCE():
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, as we discussed, we won't put this now. (I will run the quick test anyway).
Please review

vkuznet (approved):
+1

src/python/WMCore/WMRuntime/Bootstrap.py
```diff
@@ -28,6 +28,12 @@ class BootstrapException(WMException):
     """ An awesome exception """
     pass
 
+def readFloatFromFile(filePath):
+    try:
+        with open(filePath, "r") as nf:
+            return float(nf.read())
```
amaltaro: AFAIK these files are supposed to have one line only, however, to be safer and protect against bad scripts/feeders, how about we read only the first line (readline)?

src/python/WMCore/WMRuntime/Bootstrap.py
```diff
@@ -231,12 +237,18 @@ def createInitialReport(job, reportName):
     # $MACHINEFEATURES/total_cpu: number of configured job slots
     # $JOBFEATURES/hs06_job: HS06 score available to your job
     # $JOBFEATURES/allocated_cpu: number of allocated slots (=8 in case of a multicore job
+
+    machineFeaturesFile = os.environ.get('MACHINEFEATURES/hs06')
```
amaltaro: the env var is just `MACHINEFEATURES`

src/python/WMCore/WMRuntime/Bootstrap.py
```diff
@@ -231,12 +237,18 @@ def createInitialReport(job, reportName):
...
     report.data.machineFeatures = {}
-    report.data.machineFeatures['HS06_SCORE_HOST'] = os.environ.get('MACHINEFEATURES/hs06')
-    report.data.machineFeatures['TOTAL_CPU'] = os.environ.get('MACHINEFEATURES/total_cpu')
+    if machineFeaturesFile:
+        report.data.machineFeatures['HS06_SCORE_HOST'] = readFloatFromFile("%s/hs06" % machineFeaturesFile)
```
amaltaro: How about we name these metrics after their real name from the files?

amaltaro (changes requested):
A few comments along the code

src/python/WMCore/WMRuntime/Bootstrap.py
ticoann: I was just wondering whether there might be blank lines in front. If every file has one line I can change to readline.

src/python/WMCore/WMRuntime/Bootstrap.py
ticoann: That is why I get empty dict. :-) Thanks

src/python/WMCore/WMRuntime/Bootstrap.py
ticoann: Yes, I think it is better. Thanks Alan for reviewing

ticoann:
merging it since current code is broken, but not posting to WMArchive. (Need a better approach)

vkuznet:
Seangchan, I"m not sure what you meant by your last message, the issue clearly state (DO NOT MERGE in a title) and the issue is not merged :)

ticoann:
I was processing of merging it and remove the title. :-) Now is merged. It is merged since current code is broken. but we will hold on to put in WMArchive. (Unless it is approved).

PR #8476, opened 2018-02-21
fix silent crash

### LINKED ISSUE #8365: JobAccountant terminated but the heart still beats
spotted https://cmsweb.cern.ch/reqmgr2/fetch?rid=vlimant_ACDC0_task_TOP-PhaseIITDRFall17wmLHEGS-00017__v1_T_171121_151106_5101 this by chance, in running since 5 days and 70 jobs created, nothing has ran as far as wmstats/gwmsmon says.
I could not spot anything specifically  wrong with the WQE https://cmsweb.cern.ch/couchdb/workqueue/_design/WorkQueue/_rewrite/element/d7affcd8abd8c053781fbd3688802b2a

ticoann:
fixes #8365

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -106,9 +106,9 @@ class AccountantWorker(WMConnectionBase):
                                                     database=config.ACDC.database)
 
         jobDBurl = sanitizeURL(config.JobStateMachine.couchurl)['url']
-        jobDBName = config.JobStateMachine.couchDBName
-        jobCouchdb = CouchServer(jobDBurl)
-        self.fwjrCouchDB = jobCouchdb.connectDatabase("%s/fwjrs" % jobDBName)
+        self.jobDBName = config.JobStateMachine.couchDBName
+        self.jobCouchdb = CouchServer(jobDBurl)
+        self.fwjrCouchDB = None
         self.localWMStats = WMStatsWriter(config.TaskArchiver.localWMStatsURL, appName="WMStatsAgent")
 
         # Hold data for later commital
@@ -590,6 +590,9 @@ class AccountantWorker(WMConnectionBase):
 
         Associate a logArchive output to its parent job
         """
+        if self.fwjrCouchDB is None:
+            self.fwjrCouchDB = self.jobCouchdb.connectDatabase("%s/fwjrs" % self.jobDBName)
+
         inputFileList = fwkJobReport.getAllInputFiles()
         requestName = task.split('/')[1]
         keys = []
--- a/src/python/WMCore/WorkerThreads/BaseWorkerThread.py
+++ b/src/python/WMCore/WorkerThreads/BaseWorkerThread.py
@@ -95,6 +95,29 @@ class BaseWorkerThread(object):
         """
         logging.error("Calling algorithm on BaseWorkerThread: Override me!")
 
+    def setUpHeartbeat(self, myThread):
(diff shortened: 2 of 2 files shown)


ticoann:
@amaltaro Alan, please review. Still need to be tested

src/python/WMCore/WorkerThreads/BaseWorkerThread.py
```diff
@@ -121,6 +144,9 @@ def initInThread(self, parameters):
         logging.info("Initialising default transaction")
         myThread.transaction = Transaction(myThread.dbi)
 
+        self.setUpHeartbeat()
+        self.setUpLogDB()
```
amaltaro: I'm just not sure whether at this stage the worker threads setup has been done already(?)

src/python/WMCore/WorkerThreads/BaseWorkerThread.py
```diff
@@ -225,6 +233,10 @@ def __call__(self, parameters):
         msg = "Worker thread %s terminated" % str(self)
         logging.info(msg)
 
+        # send heartbeat message that thread is terminated
+        if self.useHeartbeat and self.heartbeatAPI is not None:
```
amaltaro: Ah-haa, I like it! However, it will mask the real exception in the algorithm, see:
https://github.com/ticoann/WMCore/blob/88b40afcdad9bcaef291888fa1b0f88b623468e7/src/python/WMCore/WorkerThreads/BaseWorkerThread.py#L212

src/python/WMComponent/JobAccountant/AccountantWorker.py
```diff
@@ -590,6 +590,9 @@ def associateLogCollectToParentJobsInWMStats(self, fwkJobReport, logAchiveLFN, t
 
         Associate a logArchive output to its parent job
         """
+        if self.fwjrCouchDB is None:
```
amaltaro: I still think this belongs to the init method. And with the changes to the WorkerThread, this is not needed anymore, is it?

amaltaro (changes requested):
A few comments in the code, one that has likely to change.

src/python/WMCore/WorkerThreads/BaseWorkerThread.py
ticoann: Worker thread set up is not done this stage. However, information necessary for heartbeat and logDB setting will be available at this point.  
Actually the error happened in worker node setup. So the heartbeat register didn't happen as you explained before.

src/python/WMCore/WorkerThreads/BaseWorkerThread.py
ticoann: Good point, the problem is there are other exceptions than that code path could happen and we don't seem to report that. I will try to improved to handle this.

src/python/WMComponent/JobAccountant/AccountantWorker.py
ticoann: I was just wondering whether connection failure is temporary problem. If it fails here the component crashes. I thought if it fails in the method it would be retried. I have to double check on that.

ticoann:
@amaltaro, Alan could you review again?

amaltaro (approved):
Seangchan, patch looks good!
Since it touches a very base code, did you manage to test it? We should make sure the components and workers tables have the correct information (names, PIDs, polling cycle)

amaltaro:
test this please

ticoann:
It is tested now.  (stop the couch and restart the JobAccountant - it doesn't crash now).

PR #8481, opened 2018-02-22
fix MaxRSS setting

### LINKED ISSUE #8236: MaxRSS at assignment web ui
With the latest developments related to memory <--> maxrss (where maxRSS gets set according to the Memory requirements), the tasks configuration have the proper settings during the request creation.
Problem is, the form given via assignment web page contains default values and that value comes from the request.schema definitions (which contains the workload default values, not the ones provided within the tasks/steps).

There are at least two ways to fix it:
1. properly build the Memory/MaxRSS value (from the tasks/steps if it's a chained request) and set it at workload (and schema level), during creation
2. remove MaxRSS field from the assignment web ui. Memory isn't there anyways...

I'm in favor of 2 because it's much easier and easier to maintain at a long run.
@ticoann any thoughts?

ticoann:
fixes #8236

1. remove the MaxRSS from request creation value (No longer possible to set the request)
2. allow MaxRSS to be set in assignment different than the Memory value.
3. update Max RSS Request doc in couch

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -426,6 +426,11 @@ class Request(RESTEntity):
         # by default, it contains all unmerged LFNs (used by sites to protect the unmerged area)
         request_args['OutputModulesLFNBases'] = workload.listAllOutputModulesLFNBases()
 
+        # hacky way to update MaxRSS, if MaxRSS is not set but Memory is set Memory becomes MaxRSS value.
+        # that is how it sets in WMTask setJobResourceInformation method
+        if request_args.get('Memory') and request_args.get('MaxRSS') is not None:
+            request_args['Memory'] = request_args['MaxRSS']
+
         # save the spec first before update the reqmgr request status to prevent race condition
         # when workflow is pulled to GQ before site white list is updated
         workload.saveCouch(self.config.couch_host, self.config.couch_reqmgr_db)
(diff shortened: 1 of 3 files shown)


ticoann:
@amaltaro, Alan please review. I still need to test.

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -426,6 +426,11 @@ def _handleAssignmentStateTransition(self, workload, request_args, dn):
         # by default, it contains all unmerged LFNs (used by sites to protect the unmerged area)
         request_args['OutputModulesLFNBases'] = workload.listAllOutputModulesLFNBases()
 
+        # hacky way to update MaxRSS, if MaxRSS is not set but Memory is set Memory becomes MaxRSS value.
+        # that is how it sets in WMTask setJobResourceInformation method
+        if request_args.get('Memory') and request_args.get('MaxRSS') is not None:
```
amaltaro: Honestly, I don't like the way it looks. And your comment made me like it even less :)
I think it can cause more harm then good for cases where we have Memory or MaxRSS also set inside the tasks/steps.

I think it would be better to have some sort of helper function in the StdBase such that we can nicely build the final MaxRSS that goes into the workload_cache (json). The other changes look good.

ticoann: It won't be that bad as it looks. If Memory is int MaxRSS will be int If Memory is dict, MaxRSS will be dict. Getter function would be a little complicated, since it goes through task and returns MaxRSS depending on whether it has the same value or different value by task. (In that case construct the dictionary. I think if we provide getter for this, we need to provide all the getters the same way if it has different values for task.)
I will check how we are doing it.

ticoann:
replaced by #8489

PR #8482, opened 2018-02-22
Handle worker node info when there was no worker node in the job life…

amaltaro:
…time
Complement for #8304 and #8470

There are cases that the job dies before it reaches a worker node, thus we won't have hostName and job/machine features for those cases.

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -1516,8 +1516,8 @@ class Report(object):
         return
 
     def getWorkerNodeInfo(self):
-        wnInfo = {"HostName": self.data.hostName,
-                  "MachineFeatures": self.data.machineFeatures,
-                  "JobFeatures": self.data.jobFeatures}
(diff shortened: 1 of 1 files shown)


amaltaro:
I just did a couple of tests in my VM and the error is gone from the component log. I can also see the job details in wmstats. @ticoann please review and I'll backport and patch all the agents before I call it a day.

ticoann (approved):

ticoann:
Thanks Alan, Sorry it was my fault.

ticoann:
This error is weird though.

Since it is set for boottrap code in Runtime.
https://github.com/dmwm/WMCore/pull/8304/files

Those properties have set in a new agent for all the job report. Is this code cashed in worker node?
Anyway, the patch is safer to be added but I don't understand how this happened.

ticoann:
Sorry, I didn't see your first comment. Now I understand.

amaltaro:
That's the point, we never call Bootstrap for jobs that don't run.

ticoann:
OK I will merge this.

PR #8484, opened 2018-02-23
Remove site from DESIREDSites if task thresholds are 0

### LINKED ISSUE #7181: Multi-psn jobs don't respect all site thresholds
Diego noticed some merge jobs being sent to both CERN and HLT, even though HLT Merge task has 0 threshold (both sites are normal and have site thresholds, but only CERN has Merge thresholds > 0).

amaltaro:
Fixes #7181

Hopefully it won't affect much the component's performance. I'm out of ideas on how to efficiently remove a site from the DESIRED_Sites when one of them is up but its task has 0 thresholds.

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -523,6 +523,27 @@ class JobSubmitterPoller(BaseWorkerThread):
 
         return
 
+    def checkZeroTaskThresholds(self, jobType, siteList):
+        """
+        _checkZeroTaskThresholds_
+
+        Given a job type and a list of sites, remove sites from the list
+        if that site + task has 0 pending thresholds.
+        Returns a new list of sites
+        """
+        newSiteList = []
+        for site in siteList:
+            try:
+                taskPendingSlots = self.currentRcThresholds[site]['thresholds'][jobType]["pending_slots"]
+            except KeyError as ex:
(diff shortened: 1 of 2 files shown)


ticoann:
Are you planning to add this to production new agents?

amaltaro:
I was planning to patch at least the cms@home agent, since it has a special setup and we need this fix in. However, if we can properly test it, then we could indeed patch the agents, it touches only one file in the end. I'm still checking the performance though

bbockelm:
@vlimant - take note of this one.

@amaltaro - this only takes in effect at job submit time, right?  If so, I can't think of possible issues right now.

amaltaro:
> @amaltaro - this only takes in effect at job submit time, right? If so, I can't think of possible issues right now.

that's correct! If someone sets a task-specific thresholds to 0, it has no impact on anything already in the condor queue or somewhere else.

amaltaro:
I was using a key from the other cache data structure. I'll try to merge them altogether in another PR with minor improvements to this code (related to the JobPackage investigation).

amaltaro:
Ok, I tested with HLT Merge thresholds set to 0 and I see that site being taken out of the site list:
```
  DESIRED_Sites = "T2_CH_CERN"
  ExtDESIRED_Sites = "T2_CH_CERN,T2_CH_CERN_HLT"
```

@ticoann please review and merge. I'll come with further changes tomorrow, but in another PR :)

ticoann (approved):

ticoann:
Nice, Alan.

PR #8487, opened 2018-02-24
Switch from UUID1 to UUID4.

bbockelm:
The UUID1 algorithm generates UUIDs based on a host's MAC address
and a timestamp.  While physical MACs are all indeed unique, Docker
containers very commonly use the same MAC address.  In practice, given
the number of sites converting to Docker, this removes a lot of the
uniqueness and has led to observed collisions.

This changes to UUID4, where the UUID is generated entirely from
the RNG.

### DIFF
--- a/src/python/WMCore/Services/StompAMQ/StompAMQ.py
+++ b/src/python/WMCore/Services/StompAMQ/StompAMQ.py
@@ -169,7 +169,7 @@ class StompAMQ(object):
             'metadata': {
                 'timestamp': int(time.time()),
                 'id': id_,
-                'uuid': str(uuid.uuid1()),
+                'uuid': str(uuid.uuid4()),
             }
         }
 
--- a/src/python/WMCore/Services/UUIDLib.py
+++ b/src/python/WMCore/Services/UUIDLib.py
@@ -9,4 +9,4 @@ def makeUUID():
 
     Makes a UUID from the uuid class, returns it
     """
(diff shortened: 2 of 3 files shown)


ticoann:
@amaltaro, Alan if this eliminates the collision, maybe we don't need to put your fix code in WMCore.  I will run the integration test on  whether there is any side effect.

bbockelm:
Failures look unrelated?

(Note - feel free to push to this branch - going to be traveling most of next week and won't have time to work on it...)

amaltaro:
Since UUIDLib class has a single-line method, maybe we can simply remove that and directly use `str(uuid.uuid4)` where it's needed.
I also found this test still using uuid1
test/data/WMStats/generator.py

ticoann:
> Since UUIDLib class has a single-line method, maybe we can simply remove that and directly use str(uuid.uuid4) where it's needed.

Hi Alan, my take is unless it causes the problem, I think it might be better we use the same algorithm throughout the code. (And wrapper function allows to change whole thing at one place, if  we need to change later again). Are you concerning this might cause problem somewhere?

amaltaro:
You have a good argument to keep a one-line class :-)
I think we just update that test file and that's it.

amaltaro:
test this please

ticoann:
It creates the different format of string, but didn't see any side effect.

Issue #8221, opened 2017-10-03
Remove MaxVSize code


amaltaro:
Since we don't use it, we better get it out of the code.

amaltaro:
@hufnagel @bbockelm any reason not to remove it?

hufnagel:
Practical difference between removing it and setting it to 1TB is kind of irrelevant. But if you really are finally too bothered by it, sure, why not ;-).

amaltaro:
I always heard we should not kill jobs based on vsize, that's why I used 1TB as default value :) Just wanted to hear any good arguments for doing or not something. 
The more code, the more maintenance, that's the mean reason behind removing it if unused/useless.

hufnagel:
I cannot foresee any scenario where we would ever go back and want to actually kill on VSize. So in that sense removing it should be fine (and make the code a bit easier to maintain in the long run as you said).

bbockelm:
I would prefer to remove it.  We have `git history` if this turns out to be the wrong decision.

PR #8489, opened 2018-02-26
Remove MaxRSS and MaxVSize from StdBase spec definition 

### LINKED ISSUE #8236: MaxRSS at assignment web ui

amaltaro:
Fixes #8236
Possibly replaces #8481

From the discussion with P&R team:
https://github.com/CMSCompOps/WmAgentScripts/issues/248

we might simply remove those 2 parameters from the top level requests (keeping them only at task level). So users won't (be allowed) set these two parameters. 

MaxRSS is already automatically set to the Memory value (need to check resized jobs)
MaxVSize is just set to a huge number (1TB), and eventually we might just remove it.


amaltaro:
test this please

amaltaro:
@ticoann I'm starting the tests only now, however, feel free to leave your first review.
About the VSize, it's likely going to die: https://github.com/dmwm/WMCore/issues/8221
but for now, we just keep it set to a very high number.

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -81,6 +81,10 @@ def validate_request_create_args(request_args, config, reqmgr_db_service, *args,
     3. convert data from body to arguments (spec instance, argument with default setting)
     TODO: raise right kind of error with clear message
     """
+    ### TODO: backwards compatibility. Remove these 2 lines in ~HG1805
+    request_args.pop('MaxRSS', None)
+    request_args.pop('MaxVSize', None)
+
```
ticoann: nice. :-)

ticoann:
Patch looks good to me. We can clean up a little further (removing setMaxRSS, etc) but we can think about it later.

src/python/WMCore/ReqMgr/Utils/Validation.py
amaltaro: And I'm proud of it LoL . Actually we need it for ACDCs and clones only.

amaltaro:
I pushed a few more line changes in the 2nd commit.
It passes all my tests:
* Memory during creation of every spec, including ACDC
* Memory during assignment of every spec, including ACDC

However, it would be great to have the cleanup also in this PR. Maybe we just skip it for the next month to be on the safe side...

TODO: I still have to make sure that the current production ReqMgr is correctly setting MaxRSS, at least for request creation phase.

amaltaro:
I did again all my tests and I couldn't spot any issues, which looks suspicious...
We can discuss it later today whether we add it or not this week.

src/python/WMCore/ReqMgr/Utils/Validation.py
ticoann: Don't we need before assignment as well. (If wf is created before the update?) Anyway, you cover both cases.

amaltaro: Nope, otherwise Ops will never update their scripts :)
If wf is created after HG1711 release, then it should have the proper Memory/MaxRSS settings.

ticoann (approved):

ticoann (approved):

PR #8494, opened 2018-03-01
Fix WMBSHelper transactions; Warning record if there is no run/lumi map

### LINKED ISSUE #8334: JobCreator not creating final jobs for a specific workflow
This workflow
vlimant_task_SMP-RunIISummer15wmLHEGS-00164__v1_T_170926_193056_9589

is facing the same issues on both vocms0310 and vocms0303 (and maybe submit2).
These agents are being drained for a couple of weeks now, they have 0 jobs queued and/or in the global pool, however, there are still some subscriptions not finished and files available for further processing.

I've put vocms0310 JobCreator component into debug mode, but I can't see anything weird (besides a few things that I still don't know how it works...).

amaltaro:
Fixes #8334

@ticoann Seangchan, I create smaller transactions in WMBSHelper for cases where we have insert ignore. I also fixed that transaction code in the AccountantWorker. Let me know if there is any reason it was like that.

It will also log in the JobCreator when it finds a workflow+subscription with files available but no run/lumi information.

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -153,7 +153,7 @@ class AccountantWorker(WMConnectionBase):
         gc.collect()
         return
 
-    def loadJobReport(self, parameters):
+    def loadJobReport(self, jobReportPath):
         """
         _loadJobReport_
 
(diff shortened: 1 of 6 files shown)


ticoann (approved):

amaltaro:
Basic DMWM tests are Ok. We can leave failure mode tests for the next agent integration tests.

PR #8490, opened 2018-02-26
Remove remaining Teams code

amaltaro:
WMAgent branch version of https://github.com/dmwm/WMCore/pull/8464

PR #8501, opened 2018-03-06
Remove remaining Teams code - 1.1.8 wmagent version

amaltaro:
old wmagent version of #8490


amaltaro:
I don't think we need to merge this one, all old agents are already in drain mode and hopefully they won't be put back up&running.

PR #8504, opened 2018-03-08
Retry schedd/collector query on error

goughes:
I could not find the functionality I needed from the retry package that we have, https://pypi.python.org/pypi/retry.  It just does a set number of retries without a way to manipulate anything between tries.

Not sure I like this solution either though.

This package offers some additional features that might be useful: https://pypi.python.org/pypi/retrying

### DIFF
--- a/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
+++ b/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
@@ -29,6 +29,13 @@ class PyCondorAPI(object):
         """
         try:
             jobs = self.schedd.query(constraint, attr_list)
+            return jobs
+        except Exception:
+            # if there is an error, try to recreate the schedd instance
+            logging.info("Recreating Schedd instance due to query error...")
+            self.schedd = htcondor.Schedd()
+        try:
+            jobs = self.schedd.query(constraint, attr_list)
         except Exception as ex:
             jobs = None  # return None to signalize the query failed
             msg = "Condor failed to fetch schedd constraints for: %s" % constraint
@@ -44,6 +51,14 @@ class PyCondorAPI(object):
         ( ShadowsRunning > 9.700000000000000E-01 * MAX_RUNNING_JOBS) )
         || ( RecentDaemonCoreDutyCycle > 9.800000000000000E-01 )
         """
+        try:
+            scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
+            isOverloaded = scheddAd['CurbMatchmaking'].eval()
+            return isOverloaded
+        except Exception:
+            # if there is an error, try to recreate the collector instance
+            logging.info("Recreating Collector instance due to query error...")
+            self.coll = htcondor.Collector()
         try:
             scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
             isOverloaded = scheddAd['CurbMatchmaking'].eval()


src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -53,6 +62,8 @@ def isScheddOverloaded(self):
             logging.exception(msg)
             # since it failed, assume it's overloaded
             isOverloaded = True
+            # if there is an error, try to recreate the collector instance
+            self.coll = htcondor.Collector()
```
ticoann: @goughes, Erik, in this case doesn't need to be retried?

goughes: Yes, I can update for the collector as well if you think this retry method is ok.

ticoann: Thanks Erik, I think we can do the same thing.

amaltaro:
I have another suggestion which - I think - it's slightly cleaner (or not, it might just be six of one, half a dozen of the other). Anyways, here is it :)
```
def recreate(self, daemonName):
  logging.info("Recreating %s instance due to query error...", daemonName)
  if daemonName == "schedd":
      self.schedd = htcondor.Schedd()
  elif daemonName == "coll":
      self.coll = htcondor.Collector()

### then on those methods, we add something like
    def isScheddOverloaded(self):
...
try:
            scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
            isOverloaded = scheddAd['CurbMatchmaking'].eval()
        except Exception as ex:
            self.recreate("coll")
try:
            scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
            isOverloaded = scheddAd['CurbMatchmaking'].eval()
        except Exception as ex:
            msg = "Condor failed to fetch schedd attributes."
            msg += "Error message: %s" % str(ex)
            logging.exception(msg)
            # since it failed, assume it's overloaded
            isOverloaded = True
```

so, we recreate the daemon for the first failure (and do nothing else), and try it again. If it fails again, then print the exception and move on.

goughes:
I agree it was hard to come up with something clean for this.
You want to return isOverloaded here right?
`
try:
    scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
    isOverloaded = scheddAd['CurbMatchmaking'].eval()
    return isOverloaded
    except Exception as ex:
        self.recreate("coll")
try:
<snip>
`

goughes:
I pushed Alan's method for the Collector part.  Looking at the file now, I think his is cleaner.  Although I didn't include the recreate method.

amaltaro:
Erik, yes, I think returning `isOverloaded` is desirable, it might well mean that the schedd is overloaded and not responding _properly_.

The recreate method is useless indeed, it's just one additional line, so your collector code looks good!

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -27,6 +27,14 @@ def getCondorJobs(self, constraint, attr_list):
         Given a job/schedd constraint, return a list of jobs attributes
         or None if the query to condor fails.
         """
+        try:
+            jobs = self.schedd.query(constraint, attr_list)
+            return jobs
+        except Exception as ex:
+            # if there is an error, try to recreate the schedd instance
+            logging.info("Recreating Schedd instance due to query error...")
+            logging.exception("Error message: %s" % str(ex))
```
amaltaro: I think we can skip the first exception logging.

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -44,6 +52,15 @@ def isScheddOverloaded(self):
...
         || ( RecentDaemonCoreDutyCycle > 9.800000000000000E-01 )
         """
+        try:
+            scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
+            isOverloaded = scheddAd['CurbMatchmaking'].eval()
+            return isOverloaded
+        except Exception as ex:
+            # if there is an error, try to recreate the collector instance
+            logging.info("Recreating Collector instance due to query error...")
+            logging.exception("Error message: %s" % str(ex))
```
amaltaro: same here

amaltaro (changes requested):
It looks good! I still think we can remove those two lines though

ticoann:
@goughes, Erik, could you squash the commit?

goughes:
I will test this patch in my agent tomorrow and let you know if it works as expected.

goughes:
Just tested and the code works fine.

PR #8508, opened 2018-03-12
change splitting call to update

### LINKED ISSUE #8453: Splitting API removing attr not provided when posting new splitting
@ticoann I might have got something wrong, so please correct me if something is wrong.

While testing the `splitting` rest api, it looks like one has to provide all (existent) allowed splitting attributes, otherwise it just gets removed when new settings are posted.

For example, if I post this content:
```
[{u'splitAlgo': u'EventAwareLumiBased',
  u'splitParams': {'job_time_limit': 20000},
  u'taskName': u'/amaltaro_MonteCarloFromGEN_Agent1110_Validation_180213_114610_1192/MonteCarloFromGEN',
  u'taskType': u'Processing'}]
```
the `events_per_job` attribute is gone. I'd expect the update function to update only those attributes provided (maybe it doesn't work for nested dicts though).

The correct content to post would be
```
[{'splitAlgo': u'EventAwareLumiBased',
  'splitParams': {u'events_per_job': 507, 'job_time_limit': 20000},
  'taskName': u'/amaltaro_MonteCarloFromGEN_Agent1110_Validation_180213_131827_7141/MonteCarloFromGEN',
  'taskType': u'Processing'}]
```
which does the correct splitting update.

ticoann:
fixes #8453

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
+++ b/src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py
@@ -247,7 +247,7 @@ class WorkloadSplitting(RESTEntity):
                     msg = "Parameter '%s' is not supported in the algorithm '%s'" % (param, splittingAlgo)
                     raise cherrypy.HTTPError(400, msg)
 
-            helper.setJobSplittingParameters(splittingTask, splittingAlgo, splitParams)
+            helper.setJobSplittingParameters(splittingTask, splittingAlgo, splitParams, updateOnly=True)
 
         # Now persist all these changes in the workload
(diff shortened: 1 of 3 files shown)


ticoann:
@amaltaro, Alan please review

amaltaro (changes requested):
This patch is completely broken, starting with the inexistent method name ;)

ticoann:
Thanks Alan for review. :-) Now the patch is tested.
One thing I am wondering is whether we should separate the update and set REST api. I am not sure whether there is the case we want to wipe out splitting parameters and just inject new one.

PR #8509, opened 2018-03-13
Minor improvements to JobSubmitter; trying to reduce memory footprint

amaltaro:
Honestly, I don't really know whether I managed to provide any actual improvement to this code :-)


amaltaro:
It looks like mariadb/oracle keys are not case sensitive. I'll fix it.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -318,8 +318,7 @@ def refreshCache(self):
 
             # now check for sites in drain and adjust the possible locations
             # also check if there is at least one site left to run the job
-            if len(possibleLocations) == 0:
-                newJob['name'] = loadedJob['name']
+            if len(possibleLocations) == 0:  # FIXME I don't see any of these 3 keys in the picke file?!?
```
amaltaro: @ticoann Seangchan, I had a look at a couple job pickle files today and I couldn't find any of these keys. Am I missing any detail? Or should we just get rid of of them?

amaltaro:
@bbockelm Brian, would you know whether we still need this requirementString (still in use because of HLT):
https://github.com/amaltaro/WMCore/blob/a410888a8f5dc7ff0cac49391ffb0d86f460c5fc/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L109

another question for you, since we're running htcondor 8.6.8, I assume we can simply get rid of of this code too:
https://github.com/amaltaro/WMCore/blob/a410888a8f5dc7ff0cac49391ffb0d86f460c5fc/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L521
?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
ticoann: It only happen a certain case. I don't remember exactly but when submit failed, Ops wanted to know where the jobs supposed to go.

https://github.com/dmwm/WMCore/commit/0920e9857d682cf511eeb5e645ce93382767c6f9#diff-e453fdf0929e6edc3c8e541f17fea9b7

amaltaro: Thanks for the pointer! I'm removing my comment.

amaltaro:
test this please

amaltaro:
@ticoann this patch is also good for review. I ran several workflows and couldn't spot any issues.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -624,63 +606,54 @@ def assignJobLocations(self):
...
 
             # then we're completely done and have our basket full of jobs to submit
             if exitLoop:
                 break
 
-            # start eating through the elder jobs first
-            for job in sorted(self.cachedJobs[jobPrio].values(), key=itemgetter('timestamp')):
-                jobid = job['id']
-                jobType = job['type']
+            # can we assume jobid=1 is older than jobid=3? I think so...
```
ticoann: Does python return keys in order. (I think it is)?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -330,37 +330,28 @@ def refreshCache(self):
...
-                    newJob['name'] = loadedJob['name']
-                    newJob['possibleLocations'] = possibleLocations
+                    newJob['possibleSites'] = possibleLocations
                     badJobs[71102].append(newJob)
                     continue
 
             # try to remove draining sites if possible, this is needed to stop
             # jobs that could run anywhere blocking draining sites
             # if the job type is Merge, LogCollect or Cleanup this is skipped
-            if newJob['type'] not in self.ioboundTypes:
```
ticoann: where does type comes from I don't see original "type" key switch to "task_type" in the patch.

ticoann: Never mind, I found it from below DAO.

wmbs_sub_types.name AS type,

ticoann:
Alan, so memory saving is from mainly removing cachedJob container? Code looks cleaner.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
amaltaro: From what I read, sets are unordered, even though it returns ordered data in my interactive tests. To be safe, I sort them in the line below.

amaltaro:
self.cachedJobs was bad because it was keeping almost half of the information (output from DAO) already present in self.jobDataCache. Still, self.jobDataCache is the worst offender, since we need to keep all the DAO output + some data from the job pickle file in memory.

PR #8512, opened 2018-03-13
fix not closing cursor above the limit

### LINKED ISSUE #8506: fix the returnCursor or remove it
This causes following error when there more than 600 binding is passed (current cms oracle setup)

DatabaseError: (DatabaseError) ORA-01000: maximum open cursors exceeded.

In the comment, it said it is not supported by select - make sure this is true.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/DBCore.py#L86

There is one place in WMCore it uses but there are few places in DBS uses. @yuyiguo 

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/JobFactory.py#L289

ticoann:
fixes #8506

### DIFF
--- a/src/python/WMCore/Database/DBCore.py
+++ b/src/python/WMCore/Database/DBCore.py
@@ -72,6 +72,11 @@ class DBInterface(WMObject):
         resultProxy.close()
         return result
 
+    def _generateCursor(self, sql, binds, connection):
+
+        for bind in binds:
+            yield connection.execute(sql, bind)
+
     def executemanybinds(self, s=None, b=None, connection=None,
                          returnCursor=False):
         """
@@ -96,9 +101,10 @@ class DBInterface(WMObject):
             Trying to select many
             """
             if returnCursor:
-                result = []
-                for bind in b:
-                    result.append(connection.execute(s, bind))
+                if b:
+                    return self._generateCursor(s, b, connection)
+                else:
+                    return []
             else:
                 result = ResultSet()
                 for bind in b:
@@ -119,7 +125,6 @@ class DBInterface(WMObject):
         Return a connection to the engine (from the connection pool)
         """
         return self.engine.connect()
-
 
     def processData(self, sqlstmt, binds={}, conn=None,
                     transaction=False, returnCursor=False):
@@ -156,15 +161,23 @@ class DBInterface(WMObject):
                 #Run single SQL statement for a list of binds - use execute_many()
                 if not transaction:
                     trans = connection.begin()
-                while(len(binds) > self.maxBindsPerQuery):
-                    result.extend(self.processData(sqlstmt, binds[:self.maxBindsPerQuery],
-                                                   conn=connection, transaction=True,
-                                                   returnCursor=returnCursor))
-                    binds = binds[self.maxBindsPerQuery:]
 
-                for i in sqlstmt:
-                    result.extend(self.executemanybinds(i, binds, connection=connection,
-                                                        returnCursor=returnCursor))
+                s = sqlstmt[0]
+
+                if s.lower().endswith('select', 0, 6) and returnCursor:
+                    # for the select statement with returnCursor true, don't worry about sending to many bindings
+                    # it sends one
+                   result = self.executemanybinds(s, binds, connection=connection,
+                                                        returnCursor=returnCursor)
+                else:
+                    while (len(binds) > self.maxBindsPerQuery):
+                        result.extend(self.processData(sqlstmt, binds[:self.maxBindsPerQuery],
+                                                       conn=connection, transaction=True,
+                                                       returnCursor=returnCursor))
+                        binds = binds[self.maxBindsPerQuery:]
+
+                    result.extend(self.executemanybinds(s, binds, connection=connection,
+                                    returnCursor = returnCursor))
                 if not transaction:
                     trans.commit()
             elif len(binds) == len(sqlstmt):


ticoann:
@amaltaro, @bbockelm, Please review.

ticoann:
Still need to test/find out following statement is true. It that is not true, we can have easier/cleaner solution

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/DBCore.py#L86

```
        Can't executemany() selects - so do each combination of binds here instead.
        This will return a list of sqlalchemy.engine.base.ResultProxy object's
        one for each set of binds.
```

src/python/WMCore/Database/DBCore.py
```diff
@@ -96,9 +102,10 @@ def executemanybinds(self, s=None, b=None, connection=None,
             Trying to select many
             """
             if returnCursor:
-                result = []
-                for bind in b:
-                    result.append(connection.execute(s, bind))
+                if b:
+                    return self._generateCursor(s, b, connection)
```
bbockelm: Why a helper function here?  `executemanybinds` isn't exactly overly-long and it seems to be the only user of this two-line method.

src/python/WMCore/Database/DBCore.py
```diff
@@ -120,7 +127,6 @@ def connection(self):
         """
         return self.engine.connect()
 
-
```
bbockelm: Please no extra whitespace movement (and squash the fixup of this).

src/python/WMCore/Database/DBCore.py
```diff
@@ -139,6 +145,7 @@ def processData(self, sqlstmt, binds={}, conn=None,
             # Can take either a single statement or a list of statements and binds
             sqlstmt = self.makelist(sqlstmt)
             binds = self.makelist(binds)
+
```
bbockelm: Unnecessary whitespace movement?

src/python/WMCore/Database/DBCore.py
```diff
@@ -156,15 +163,28 @@ def processData(self, sqlstmt, binds={}, conn=None,
...
-                    result.extend(self.executemanybinds(i, binds, connection=connection,
-                                                        returnCursor=returnCursor))
+                s = sqlstmt[0]
+
+                if s.lower().endswith('select', 0, 6) and returnCursor:
+                    iterableFlag = True
+                else:
+                    iterableFlag = False
+
+                if iterableFlag:
```
bbockelm: Why not remove this variable and just do:

```
if s.lower().endswith('select', 0, 6) and returnCursor:
```

src/python/WMCore/Database/DBCore.py
```diff
@@ -156,15 +163,28 @@ def processData(self, sqlstmt, binds={}, conn=None,
...
+                    # for the select statement with returnCursor true, don't worry about sending to many bindings
+                    # it sends one
+                   result = self.executemanybinds(s, binds, connection=connection,
+                                                        returnCursor=returnCursor)
+                else:
+                    while (len(binds) > self.maxBindsPerQuery):
+                        result.extend(self.processData(sqlstmt, binds[:self.maxBindsPerQuery],
```
bbockelm: I guess I'm uncomfortable with requiring `processData` to know low-level implementation details of `executemanybinds`.  Can you factor out the logic into a shared method?

bbockelm (changes requested):
A few minor items, but the thing that really makes me uncomfortable is the strange intertwining of `executemanybinds` and `processData`.

src/python/WMCore/Database/DBCore.py
ticoann: It seems it doesn't allow return and yield within the same function. (Gives me syntax error) So I had to add a separate helper function.

src/python/WMCore/Database/DBCore.py
ticoann: I will  fix it

src/python/WMCore/Database/DBCore.py
ticoann: I will fix that.

src/python/WMCore/Database/DBCore.py
ticoann: I will fix this

src/python/WMCore/Database/DBCore.py
bbockelm: Hm.  Might be better to consistently yield a generator from here.  Mixing generators and values is a fairly confusing call interface.  How about always using generators?  In that case, we'd have:

```
>>> def test(foo):
...   if foo:
...     for f in foo:
...       yield f
...   else:
...     raise StopIteration()
... 
>>> for t in test([]): print t
... 
>>> for t in test([1,2,3,4]): print t,
... 
1 2 3 4
>>> print test([])
<generator object test at 0x10226eb40>
>>> print test([1,2,3,4])
<generator object test at 0x10226eb40>
```

src/python/WMCore/Database/DBCore.py
ticoann: The problem is executemanybinds returns different type depending on the condition. (Iterable from generator and list). I could change to make return iterable not list but since this function is used everywhere (outside WMCore (DBS) as well), I have to coordinate no code expect list.
Also I am not sure I how I can deal with recursive call with the function returning iterable. (We have to yield again be second yield will return the generator object not the original object yield from generator.

FYI (This recursive call is originally added for the limitation of number of binding variable Oracle accepts - I think it is 500(?))

We can completely refactor code, 
1. This only returns iterable of cursors - too many changes
2. Separate select and insert operation give different executemany function for that.

src/python/WMCore/Database/DBCore.py
ticoann: > Hm. Might be better to consistently yield a generator from here. Mixing generators and values is a fairly confusing call interface. How about always using generators? In that case, we'd have:

I added a few comments below regarding this. (Basically more changes need to be made, including outside WMCore(i.e. DBS) :-))
I will check how bad it is to change this.

ticoann:
Just for the record, getting slightly different error. (sourcePFN is seems to be wrong)

```
Job Name: bd42096a-2258-11e8-9d12-02163e01b4c0-0
WMBS job id: 35
Workflow: sryu_StepChain_DupOutMod_sryu_180307_212343_2797
Task: /sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput
Status: jobfailed
Input dataset:
Site: T1_US_FNAL_Disk
Agent: vocms0193.cern.ch
ACDC URL: https://reqmgr2-dev.cern.ch/couchdb/acdcserver
State Transition
jobfailed: 2018/3/7 (Wed) 22:52:01 UTC, T1_US_FNAL
Exit code: 60307
Retry count: 0
Errors:
logArch1
LogArchiveFailure (Exit Code: 60307)
<@========== WMException Start ==========@>
Exception Class: StageOutFailure
Message: Failure for local stage out:
[Errno 2] No such file or directory: 'root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz'Traceback (most recent call last):
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutMgr.py", line 318, in localStageOut
    impl(protocol, localPfn, pfn, options, checksums)
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutImpl.py", line 208, in __call__
    command = self.createStageOutCommand(sourcePFN, targetPFN, options, checksums)
  File "/srv/job/WMCore.zip/WMCore/Storage/Backends/FNALImpl.py", line 111, in createStageOutCommand
    original_size = os.stat(sourcePFN)[6]
OSError: [Errno 2] No such file or directory: 'root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz'

	TargetPFN : root://cmsdcadisk01.fnal.gov//dcache/uscmsdisk/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz
	ErrorCode : 60311
	ModuleName : WMCore.Storage.StageOutError
	MethodName : __init__
	LFN : /store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz
	ClassInstance : None
	FileName : /srv/job/WMCore.zip/WMCore/Storage/StageOutError.py
	ClassName : None
	Command : stageout-xrdcp-fnal
	LineNumber : 32
	InputPFN : root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/store/unmerged/logs/prod/2018/3/7/sryu_StepChain_DupOutMod_sryu_180307_212343_2797/ProdMinBias/DIGIPROD1MergeRAWSIMoutput/0000/0/bd42096a-2258-11e8-9d12-02163e01b4c0-0-0-logArchive.tar.gz
	Protocol : writexrd
	ErrorNr : 0
	ErrorType : GeneralStageOutFailure

Traceback: 
  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutMgr.py", line 318, in localStageOut
    impl(protocol, localPfn, pfn, options, checksums)

  File "/srv/job/WMCore.zip/WMCore/Storage/StageOutImpl.py", line 208, in __call__
    command = self.createStageOutCommand(sourcePFN, targetPFN, options, checksums)

  File "/srv/job/WMCore.zip/WMCore/Storage/Backends/FNALImpl.py", line 111, in createStageOutCommand
    original_size = os.stat(sourcePFN)[6]

<@---------- WMException End ----------@>

```

amaltaro:
@ticoann I think you're mixing the PRs :)

ticoann:
Sorry wrong issue. :-)

ticoann:
Oracle test failed. Need a further investigation. Also there are ~30 places needs to be changed in DBS code.

ticoann:
replaced by #8568

PR #8515, opened 2018-03-14
Adapt EventsPerJob to a multiple of EventsPerLumi - EventBased only

### LINKED ISSUE #6857: prevent small lumisection in EventBased job splitting
provided eventsperjob and eventsperlumi, the current schema of splitting allows for the last lumi of a job to have a number of events smaller than eventsperlumi. This has an impact on multi-thread processing down the line of such production.

the proposal is to change so that the actual eventsperjob is made an exact multiple of eventsperlumi so that no smaller lumisection  is created

@bendavid

amaltaro:
Fixes #6857

Use cases are:
1) if requestor provides both EpJ and EpL, then nothing gets changed
2) if requestor provides EpJ but *not* EpL, then usual `EpL = EpJ` is applied
3) if requestor provides EpL but *not* EpJ, 8h job is calculate and EpJ is adapted to a multiple of EpL
4) if requestor doesn't provide any of these, 8h job is calculated and usual `EpL = EpJ` is applied.

I'm expecting tests to fail... I also think the workload_cache document won't get updated with the adapted value (that information is uploaded outside of the spec factory, I have to investigate how to update it before it gets persisted)

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/DataProcessing.py
@@ -28,6 +28,7 @@ class DataProcessing(StdBase):
             if self.procJobSplitAlgo == "EventAwareLumiBased":
                 self.procJobSplitArgs["job_time_limit"] = 48 * 3600  # 2 days
             self.procJobSplitArgs["events_per_job"] = self.eventsPerJob
+            arguments['EventsPerJob'] = self.eventsPerJob
         elif self.procJobSplitAlgo == "LumiBased":
             self.procJobSplitArgs["lumis_per_job"] = self.lumisPerJob
         elif self.procJobSplitAlgo == "FileBased":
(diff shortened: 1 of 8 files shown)


ticoann (approved):

ticoann:
Alan, patch looks good to me. What kind of error are you getting?

amaltaro:
I didn't test yet, but I think the workload will have one value and the workload_cache (json doc) will have another. Better not merge it until I get that fixed.

amaltaro:
@ticoann I managed to fix the issue mentioned above. You can see I update `arguments['EventsPerJob']` from the spec factory because we save the workload json in another place, here I think:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L551

If it passes the unit tests and you agree with the code, feel free to merge. Otherwise we can discuss tomorrow and fix/merge it in the coming days.

Issue #8443, opened 2018-02-08
Publish ReqMgr stats to CERN MONIT


amaltaro:
We could start with:
* for workflows in assignment-approved
  * their priority
  * their RequestNumEvents (if starting from scratch)
  * maybe, amount of requests in each campaign (tricky for TaskChains)

ticoann:
Hi Alan,
This information is in wmstats, exactly what statistics  need to be displayed and what is the advantage of putting is CERN Monit?

amaltaro:
Indeed they are :) I usually don't pay attention to that stats box. Still, one has to apply some filters and aggregation by priority is not supported (it is for campaign), and the output isn't very user friendly (either a row or a table).
Having this info in Monit would allow to make a fancy graph :) and have that visualization in a dashboard, together with other overview graphs of central production.

PR #8521, opened 2018-03-19
Fetch monitoring metrics from the ReqMgr monitoring thread

amaltaro:
Supposedly fixing #8443 , but I still need to get my AMQ topic created and those documents properly structured and posted to MonIT.

It needs this deployment patch as well: https://github.com/dmwm/deployment/pull/590
There are also a few tweaks that I need to make in this patch, but in theory it's working :)

This is how these metrics will look like for wmstats and MonIT:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/PR_8521/data_struct

There are changes to the spec files as well, we need to add a new dependency on: py2-stomp

### DIFF
--- a/src/python/WMCore/REST/HeartbeatMonitorBase.py
+++ b/src/python/WMCore/REST/HeartbeatMonitorBase.py
@@ -8,6 +8,9 @@ class HeartbeatMonitorBase(CherryPyPeriodicTask):
         super(HeartbeatMonitorBase, self).__init__(config)
         self.centralWMStats = WMStatsWriter(config.wmstats_url)
         self.threadList = config.thread_list
+        self.postToAMQ = getattr(config, "post_to_amq", False)
+        self.topicAMQ = getattr(config, "topic_amq", None)
+        self.hostPortAMQ = getattr(config, "host_port_amq", None)
 
     def setConcurrentTasks(self, config):
         """
@@ -18,7 +21,7 @@ class HeartbeatMonitorBase(CherryPyPeriodicTask):
     def reportToWMStats(self, config):
         """
         report thread status and heartbeat.
-        Also can report additional mointoring information by rewriting addAdditionalMonitorReport method
+        Also can report additional monitoring information by rewriting addAdditionalMonitorReport method
         """
         self.logger.info("Checking Thread status...")
         downThreadInfo = self.logDB.wmstats_down_components_report(self.threadList)
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
@@ -1,5 +1,187 @@
-from __future__ import (division, print_function)
+from __future__ import division, print_function
+
+import time
+from pprint import pformat
+
 from WMCore.REST.HeartbeatMonitorBase import HeartbeatMonitorBase
+from WMCore.ReqMgr.DataStructs.RequestStatus import ACTIVE_STATUS
+from WMCore.Services.StompAMQ.StompAMQ import StompAMQ
+from WMCore.Services.WMStatsServer.WMStatsServer import WMStatsServer
+
+
+def _getCampaign(propertyValue):
+    """
+    WMStats server can return this argument in three different data types.
+    If there are multiple values, return only the first one.
+    """
+    if not propertyValue:  # Empty string is Ok; it should never be None though
+        return ["NONE"]
+    elif isinstance(propertyValue, basestring):
+        return [propertyValue]
+    # then it's a list
+    return propertyValue
+
+
+def _getRequestNumEvents(propertyValue):
+    """
+    WMStats server can return this argument in three different data types.
+    If there are multiple values, return only the first one.
+    """
+    if not propertyValue:
+        return 0
+    elif isinstance(propertyValue, list):
+        return propertyValue[0]
+    # then it's integer
+    return propertyValue
+
+
+def initMetrics():
+    """
+    Initialize the metrics dictionary
+    :param veryActiveStatus: a small list of active status
+    """
+    results = {"requestsByStatus": {},
+               "requestsByStatusAndCampaign": {},
+               "requestsByStatusAndPrio": {},
+               "requestsByStatusAndNumEvts": {}}
+
+    for st in ACTIVE_STATUS:
+        results["requestsByStatus"].setdefault(st, 0)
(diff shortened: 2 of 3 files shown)


src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
```diff
@@ -1,5 +1,86 @@
...
+        """
+        _buildAssignmentStats_
+        Given a list of requests and their details, parse each of them and extract
+        the information we want.
+        :param requestsInfo: list of request dictionaries
+        :return: summary of requests in a given state with their priority and
+        requested number of events.
+        """
+        stats = {"assignmentByPrio": None, "assignmentByReqEvents": None}
+        requestsByPrio = defaultdict(Counter)
```
amaltaro: change it! json encoding/decoding it creates extras key/value pairs.

amaltaro:
@ticoann if you have some spare minutes, feel free to have a first look at it and suggest better ways of fetching some of this info ;)

ticoann:
HI Alan,  patch looks good to me. However, if you are only getting the statistics on Active request following call might be better than new views in reqmgr. It loads all in the memory but it just uses the cache so it shouldn't be too much overhead.

https://cmsweb.cern.ch/wmstatsserver/data/filtered_requests?mask=RequestStatus&mask=Campaign&mask=RequestPriority&mask=RequestType

Counting is easier with view but if you get different statics you can do it at one time going through the data.

amaltaro:
I knew you'd have a better idea! ;) Using this wmstats API is much cleaner indeed.
BTW, `RequestNumEvents` is returned as a list, it should be just an integer, I might fix that altogether.

ticoann:
Sorry Alan for making you to rework.  If you decided to use the wmstatsserver API, you can use this  wrapper class/method.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WMStatsServer/WMStatsServer.py#L79

amaltaro:
@ticoann can you please review this patch. I still have to fix a few things:
* remove my pformat code
* test the upload to AMQ (when CERN IT answers my ticket, otherwise we can simply pass that code for now)
* maybe move the `postToAMQ` method to the base class (?) If we do so, then I guess we also need to add 'WMCore.Services+' dependency to 'wmc-rest'?

Thanks!

src/python/WMCore/REST/HeartbeatMonitorBase.py
```diff
@@ -8,6 +8,8 @@ def __init__(self, rest, config):
         super(HeartbeatMonitorBase, self).__init__(config)
         self.centralWMStats = WMStatsWriter(config.wmstats_url)
         self.threadList = config.thread_list
+        self.userAMQ = config.user_amq
```
ticoann: You will be adding this in reqmgr2 config, right?

amaltaro:
Yes, check out the first comment in this PR. I have a link to the deployment changes (and needed changes at spec level).

src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
```diff
@@ -1,5 +1,186 @@
...
+            results['requestsByStatus'][status] += 1
+
+            for campaign in _getCampaign(reqInfo["Campaign"]):
+                results["requestsByStatusAndCampaign"][status].setdefault(campaign, 0)
+                results['requestsByStatusAndCampaign'][status][campaign] += 1
+
+            if status not in VERY_ACTIVE_STATUS:
+                continue
+
+            requestPrio = reqInfo['RequestPriority']
```
ticoann: So this one is for all the active status not just VERY_ACTIVE_STATUS, right?

amaltaro: Exactly, though I'm not 100% sure. Still, I think people don't care about request priority for requests in completed, or aborted, or ... closed-out etc, I think...

ticoann:
Patch looks good to me, Alan. Thanks

src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
ticoann: I wonder whether it will be more confusing select certain state for certain condition but not other conditions, we can report everything but it doesn't have to be used.

amaltaro: Yes, I slept on that and I think it's better to don't use a small set of ACTIVE_STATUS, but fetch the same metrics for all of them. Thanks! I'm updating this PR

amaltaro:
@ticoann Seangchan, can you please review it again? Now we use ACTIVE_STATUS for all those metrics.

I also managed to test the upload to AMQ and it seems to be working fine (I could find the docs in Kibana). I also put a code to avoid crashing the thread if there are issues posting docs to AMQ.

Once we merge it, we need to cut a new tag and build new RPMs with py2-stomp, otherwise ReqMgr2 HB thread will fail to import stomp.

Last but not least, I need to check with Lina on how we can share these credentials with her (see deployment PR).

src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
```diff
@@ -1,5 +1,187 @@
...
+            self.uploadToAMQ(allDocs)
+
+        return results
+
+    def buildMonITDocs(self, stats):
+        """
+        Given the statistics that are uploaded to wmstats, create different
+        documents to post to MonIT AMQ (aggregation-friendly docs).
+        """
+        commonInfo = {"version": '0.3',
```
ticoann: where is the version comes from?

ticoann:
Patch looks good to me. Thanks Alan, I will create a new tag with spec update later today.

src/python/WMCore/ReqMgr/CherryPyThreads/HeartbeatMonitor.py
amaltaro: Good question :) I took it from Benjamin's script, however I think we don't really need it. StompAMQ has already a version in the headers. Anyways, we can keep it for now, getting info from CERN IT is a great challenge...

amaltaro:
Seangchan, just in case, you can find the spec changes in my repo in vocms055:/build/amaltaro/cmsdist

ticoann:
Thanks Alan,

PR #8524, opened 2018-03-20
fix unicode encoding failure

### LINKED ISSUE #8403: Another unicode encoding failure

ticoann:
fixes #8403

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -560,7 +560,13 @@ class Report(object):
         errDetails = getattr(stepSection.errors, errEntry)
         errDetails.exitCode = exitCode
         errDetails.type = str(errorType)
-        errDetails.details = errorDetails
+
+        if hasattr(errorDetails, "decode"):
+            # Fix for the unicode encoding issue, #8043
+            # interprets this string using utf-8 codec and ignoring any errors
+            errDetails.details = errorDetails.decode('utf-8', 'ignore')
+        else:
+            errDetails.details = errorDetails
 
         setattr(stepSection.errors, "errorCount", errorCount + 1)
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan, please review.

amaltaro:
We "fixed" it so many times that I don't even know whether it's right or not anymore :P
Seangchan, did you actually run this code against that bad FWJR? Just to make sure this fix really fixes that issue.

If you have that example handy, could you check the performance impact on decoding that string 10k times, for instance. Just to avoid making JobAccountant even fattier.

ticoann:
Hi Alan, I check the one document, but not performance test.
However if you are concerning about the performance, the cheapest option is the temporary fix we had. https://github.com/dmwm/WMCore/pull/8404 since that only does when error occurs.

ticoann:
One thing more concerning is this not Python3 compatible.

amaltaro (approved):
In this case, thumbs up! 
About python3, it passed the checks. Why you say it's not compatible?

ticoann:
> About python3, it passed the checks. Why you say it's not compatible?

I don think python3 string type has decode method. (I might be wrong though)

https://docs.python.org/3/library/stdtypes.html#string-methods

amaltaro:
Hmm, good point. Should we then chase the source of this problem and have a fix py3 compatible? It must be coming from one of the log files in the worker node, so if we can open/parse those logs following the latest attempt, that would be better. If it takes too long, we just merge this one then.

PS.: What surprises me is that it hasn't happened in any of the production nodes...

PR #8526, opened 2018-03-22
Set MaxRetries to 0 for testbed/dev agents

amaltaro:
Following the maxRetries migration to central AuxDB. Notice I added a new parameter to make sure we make a PUT request against the correct url.

FYI @ticoann  We can merge it when we start rolling the new agent release. For now, it will serve our private agents ;)


amaltaro:
It has been tested and it works fine. We need to wait for the next stable release to come out though before merging it.

amaltaro:
@ticoann this one should be ready to go as well (mind the repurpose of `-c` option).

PR #8527, opened 2018-03-23
Make WMBSHelper slightly more verbose when there is input data

amaltaro:
Related - but not fixing - to #8528.
Just so we can have a better idea of where (not exactly though) `WorkQueueManagerWMBSFileFeeder` thread gets busy.


amaltaro:
I've pushed these changes to #8533

PR #8530, opened 2018-03-23
fix multi reject in reqmgr2

### LINKED ISSUE #8391: reject button does not work on the assign page
@bajarang was having difficulties rejecting workflows that belongs to him from the https://cmsweb.cern.ch/reqmgr2/assign page.
Is this supposed to function ?
He was able to do it when reaching directly to https://cmsweb.cern.ch/reqmgr2/fetch?rid=request-bsutar_HLT_newco_RelVal_300155_171207_141000_168 but this is painful when you have multiple workflows to reject.
@ticoann @amaltaro  no emergency, just needs to be fixed

ticoann:
fixes #8391

### DIFF
--- a/src/html/ReqMgr/jinja_templates/assign.tmpl
+++ b/src/html/ReqMgr/jinja_templates/assign.tmpl
@@ -192,12 +192,17 @@ function ProcessRequests(new_status) {
         }
     }
     if (ids.length>0) {
-        // capture form id="assign-settings"
-        var parameters = $('#assign-settings').serializeObject();
-        // capture form id="assign-misc-settings"
-        var misc_args = $('#assign-misc-settings').serializeObject();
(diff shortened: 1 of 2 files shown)


ticoann:
@amaltaro, Alan this is tested. If you cut the new tag for next week. Please include this. Thanks,

amaltaro:
Without investigating the code, why new_status == assigned? If he wants to reject, the workflow goes to rejected status, not to assigned :-)

ticoann:
> Without investigating the code, why new_status == assigned? If he wants to reject, the workflow goes to rejected status, not to assigned :-)

What it does is if assigned status we pass all the parameters (SiteWhitelist, etc), but for reject status it only passes request name and new status. I ran the test, the code should be safe. :smile:

PR #8531, opened 2018-03-23
add override param to reqmgr2

### LINKED ISSUE #8523: Provide parameters for override from Spec definition
Need to set up the override value from spec definition.
and set it using
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMStep.py#L139

ticoann:
fixes #8523

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -1153,7 +1153,11 @@ class StdBase(object):
                      "BlockCloseMaxEvents": {"default": 25000000, "type": int, "validate": lambda x: x > 0},
                      "BlockCloseMaxSize": {"default": 5000000000000, "type": int, "validate": lambda x: x > 0},
                      # dashboard activity
-                     "Dashboard": {"default": "production", "type": str, "validate": activity}
+                     "Dashboard": {"default": "production", "type": str, "validate": activity},
+                     # Override parameters for step (EOS log location, etc
+                     # set to "" string or None for eos-lfn-prefix if you don't want to save the log in eos
+                     "Override": {"default": {"eos-lfn-prefix": "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent"},
+                                  "type": dict},
                      }
         # Set defaults for the argument specification
         StdBase.setDefaultArgumentsProperty(arguments)
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -1380,6 +1380,26 @@ class WMWorkloadHelper(PersistencyHelper):
         """
         return self.data.section_('overrides')
 
+    def setWorkloadOverrides(self, overrides):
+        """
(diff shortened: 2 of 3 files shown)


ticoann:
@amaltaro, Alan could you review, still need to test though.

ticoann:
@amaltaro, Alan. It is tested now.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -1153,7 +1153,11 @@ def getWorkloadAssignArgs():
                      "BlockCloseMaxEvents": {"default": 25000000, "type": int, "validate": lambda x: x > 0},
                      "BlockCloseMaxSize": {"default": 5000000000000, "type": int, "validate": lambda x: x > 0},
                      # dashboard activity
-                     "Dashboard": {"default": "production", "type": str, "validate": activity}
+                     "Dashboard": {"default": "production", "type": str, "validate": activity},
+                     # Override parameters for step (EOS log location, etc
+                     # set to "" string or None for eos-lfn-prefix if you don't want to save the log in eos
+                     "Override": {"default": {"eos-lfn-prefix": "root://eoscms.cern.ch//eos/cms/store/logs/prod/recent"},
```
amaltaro: This override is actually set to all steps, right? I wonder if we aren't overriding the LogCollect jobs as well, such that they use this path instead of the usual CASTOR and EOS?

We might have issues validating this parameter with a dict value, but I'm planning to improve that in the very near future anyways.

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -1380,6 +1380,26 @@ def getWorkloadOverrides(self):
...
+    def setWorkloadOverrides(self, overrides):
+        """
+        _getWorkloadOverrides_
+
+        Get the overrides config section
+        of this workload, creates it if it doesn't exist
+        """
+        
+        if overrides:
+            taskIterator = self.taskIterator()
```
amaltaro: actually, isn't it returning only the top level task? Is this override then propagated to all tasks or only the top level one?

amaltaro:
Seangchan, if my assumptions are wrong and you have already addressed those concerns, I think it's good to go to testbed.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
ticoann: It sets for the all the steps but only affects LogArchive stageout in other case, it will be ignored.
Yes, currently we don't have the proper validation for this.  I can add validation for only "eos-lfn-prefix" parameter.

amaltaro: This is probably not going to be used by anyone, I'd rather add the validation code later on (it's always a tricky business anyways) :)

src/python/WMCore/WMSpec/WMWorkload.py
ticoann: It should iterate all the task not just toplevel.

childTaskIterator is the one iterates only top level.

Issue #8528, opened 2018-03-23
Heartbeat timeouts for WorkQueueManagerWMBSFileFeeder


amaltaro:
As reported here:
https://its.cern.ch/jira/browse/CMSCOMPPR-2534

this WorkQueueManager thread spent around 12h processing/inserting data into the database for some requests.

From vocms0253 logs, it spent around 11h to process this request:
vlimant_ACDC0_task_HIG-RunIIFall17wmLHEGS-00623__v1_T_180322_210305_3387

```
2018-03-22 22:17:58,097:139943359907584:INFO:WorkQueue:Adding WMBS subscription for vlimant_ACDC0_task_HIG-RunIIFall17wmLHEGS-00623__v1_T_180322_210305_3387
2018-03-22 22:18:44,273:139943359907584:INFO:SandboxCreator:Created sandbox vlimant_ACDC0_task_HIG-RunIIFall17wmLHEGS-00623__v1_T_180322_210305_3387-Sandbox.tar.bz2 with size 8241235
2018-03-22 22:18:44,329:139943359907584:INFO:WMBSHelper:Top level subscription created: 9816
2018-03-22 22:18:44,544:139943359907584:INFO:WMBSHelper:Child subscription created: 9817
2018-03-22 22:18:44,554:139943359907584:INFO:WMBSHelper:Child subscription created: 9818
2018-03-22 22:18:44,754:139943359907584:INFO:WMBSHelper:Child subscription created: 9819
2018-03-22 22:18:44,767:139943359907584:INFO:WMBSHelper:Child subscription created: 9820
2018-03-22 22:18:45,054:139943359907584:INFO:WMBSHelper:Child subscription created: 9821
2018-03-22 22:18:45,111:139943359907584:INFO:WMBSHelper:Child subscription created: 9822
2018-03-22 22:18:45,122:139943359907584:INFO:WMBSHelper:Child subscription created: 9823
2018-03-22 22:18:45,139:139943359907584:INFO:WMBSHelper:Child subscription created: 9824
2018-03-22 22:18:45,159:139943359907584:INFO:WMBSHelper:Child subscription created: 9825
2018-03-22 22:18:45,170:139943359907584:INFO:WMBSHelper:Child subscription created: 9826
2018-03-22 22:18:45,185:139943359907584:INFO:WMBSHelper:Child subscription created: 9827
2018-03-22 22:18:45,201:139943359907584:INFO:WMBSHelper:Child subscription created: 9828
[...]

PR #8532, opened 2018-03-25
TEMP wmagent debug patch for WorkQueueManagerWMBSFileFeeder

amaltaro:
Mitigating #8528 
A better attempt to fix it is in #8533, though we can't improve every bit unless we proceed with a code refactoring.
NEVER MERGE! I'm applying it to vocms0257...

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/Add.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/Add.py
@@ -1,12 +1,9 @@
 #!/usr/bin/env python
-
-
-
-
-
 """
 MySQL implementation of AddFile
 """
+
+from Utils.IteratorTools import grouper
 from WMCore.Database.DBFormatter import DBFormatter
 
 class Add(DBFormatter):
@@ -16,18 +13,18 @@ class Add(DBFormatter):
 
     def getBinds(self, files, size, events, cksum, dataset_algo, status, workflowID, inPhedex):
         # Can't use self.dbi.buildbinds here...
-        binds = {}
+        binds = []
         if isinstance(files, basestring):
-            binds = {'lfn': files,
+            bind = {'lfn': files,
                      'filesize': size,
                      'events': events,
                      'dataset_algo': dataset_algo,
                      'status' : status,
                      'workflow': workflowID,
                      'in_phedex': inPhedex}
-        elif isinstance(files, list):
-        # files is a list of tuples containing lfn, size, events, cksum, dataset, status
-            binds = []
+            binds.append(bind)
+        elif isinstance(files, (list, set)):
+            # files is a list of tuples containing lfn, size, events, cksum, dataset, status
             for f in files:
                 binds.append({'lfn': f[0],
                               'filesize': f[1],
@@ -44,6 +41,7 @@ class Add(DBFormatter):
         binds = self.getBinds(files, size, events, cksum, datasetAlgo, status,
                               workflowID, inPhedex)
 
-        self.dbi.processData(self.sql, binds, conn=conn, transaction=transaction)
+        for sliceBinds in grouper(binds, 10000):
+            self.dbi.processData(self.sql, sliceBinds, conn=conn, transaction=transaction)
 
         return
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/AddChecksumByLFN.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/AddChecksumByLFN.py
@@ -7,7 +7,7 @@ MySQL implementation of AddChecksumByLFN
 
 
 
-
+from Utils.IteratorTools import grouper
 from WMCore.Database.DBFormatter import DBFormatter
 
 class AddChecksumByLFN(DBFormatter):
@@ -21,9 +21,10 @@ class AddChecksumByLFN(DBFormatter):
         if bulkList:
             binds = bulkList
         else:
-            binds = {'lfn': lfn, 'cktype': cktype, 'cksum': cksum}
+            binds = [{'lfn': lfn, 'cktype': cktype, 'cksum': cksum}]
 
-        result = self.dbi.processData(self.sql, binds,
-                         conn = conn, transaction = transaction)
+        for sliceBinds in grouper(binds, 10000):
+            self.dbi.processData(self.sql, sliceBinds,
+                                 conn = conn, transaction = transaction)
 
         return
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetLocationByLFN.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetLocationByLFN.py
@@ -6,15 +6,14 @@ MySQL implementation of DBSBuffer.SetLocationByLFN
 """
 
 
-
-
+import logging
+from Utils.IteratorTools import grouper
 from WMCore.Database.DBFormatter import DBFormatter
 
 class SetLocationByLFN(DBFormatter):
     sql = """INSERT IGNORE INTO dbsbuffer_file_location (filename, location)
                SELECT df.id, dl.id
-               FROM dbsbuffer_file df
-               INNER JOIN dbsbuffer_location dl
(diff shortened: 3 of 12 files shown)


amaltaro:
I'm running some tests in testbed and, if everything works fine, I'll push it to at least vocms0257.

vlimant:
what's the GH issue you are trying to fix with this ?

amaltaro:
I updated the description

amaltaro:
I have further changes in another branch and from the logs, this line
https://github.com/amaltaro/WMCore/blob/643b58b826d29a072d64b18ca6d39c451f7e6c8f/src/python/WMCore/WorkQueue/WMBSHelper.py#L580

takes around 1.5h to add 10k file locations in dbsbuffer_file_location (out of 310k...).

I initially thought it could be because locations were added (a few lines above) to DBSBuffer but the transaction wasn't commited, causing "weird" issues to the file location DAO, since locations supposedly could not be found in dbsbuffer. However, this can't be true because we have the same behavior for wmbs files...

src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/Add.py
```diff
@@ -44,6 +41,7 @@ def execute(self, files = None, size = 0, events = 0, cksum = 0,
         binds = self.getBinds(files, size, events, cksum, datasetAlgo, status,
                               workflowID, inPhedex)
 
-        self.dbi.processData(self.sql, binds, conn=conn, transaction=transaction)
+        for sliceBinds in grouper(binds, 10000):
```
ticoann: I don't think this is necessary since we do slice in processData anyway. it would actually add a little more time.

src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/AddChecksumByLFN.py
```diff
@@ -21,9 +21,10 @@ def execute(self, lfn = None, cktype = None, cksum = None, bulkList = None, conn
         if bulkList:
             binds = bulkList
         else:
-            binds = {'lfn': lfn, 'cktype': cktype, 'cksum': cksum}
+            binds = [{'lfn': lfn, 'cktype': cktype, 'cksum': cksum}]
 
-        result = self.dbi.processData(self.sql, binds,
-                         conn = conn, transaction = transaction)
+        for sliceBinds in grouper(binds, 10000):
```
ticoann: same for this one as well. we slice it in processData. If this give the performance boost. That means we need to change to the code in processData. (Maybe using loop instead of recursive call. - Although I don't think that makes big performance difference)

src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetLocationByLFN.py
```diff
@@ -25,6 +24,10 @@ def execute(self, binds, conn = None, transaction = None):
         Expect binds in the form {lfn, pnn}
 
         """
-        self.dbi.processData(self.sql, binds, conn = conn,
-                             transaction = transaction)
+        count = 0
+        for sliceBinds in grouper(binds, 10000):
```
ticoann: same as above.

src/python/WMCore/WMBS/File.py
```diff
@@ -529,20 +529,19 @@ def addFilesToWMBSInBulk(filesetId, workflowName, files, isDBS=True,
...
+    logging.info("AMR building binds for %d wmbs files for workflow %s", len(files), workflowName)
     for wmbsFile in files:
         lfn = wmbsFile['lfn']
-        lfnList.append(lfn)
+        lfnList.add(lfn)
 
         if wmbsFile.get('inFileset', True):
-            if lfn not in fileLFNs:
-                fileLFNs.append(lfn)
+            fileLFNs.add(lfn)
```
ticoann: This make sense as we discussed in chat.

src/python/WMCore/WMBS/MySQL/Files/AddDupsToFileset.py
```diff
@@ -65,14 +63,18 @@ def execute(self, file, fileset, workflow, conn = None, transaction = False):
...
                           "insert_time": timestamp, "workflow": workflow})
-            availBinds.append({"lfn": fileLFN, "fileset": fileset,
-                               "workflow": workflow})
+            availBinds.append({"lfn": fileLFN, "fileset": fileset, "workflow": workflow})
 
-        self.dbi.processData(self.sql, binds, conn = conn,
-                             transaction = transaction)
-        self.dbi.processData(self.sqlAvail, availBinds, conn = conn,
-                             transaction = transaction)
+        for slicedBinds in grouper(binds, 10000):
```
ticoann: The same as above, I don't think slice is needed.

src/python/WMCore/WMBS/MySQL/Files/AddDupsToFileset.py
```diff
@@ -65,14 +63,18 @@ def execute(self, file, fileset, workflow, conn = None, transaction = False):
...
+            self.dbi.processData(self.sql, slicedBinds, conn = conn,
+                                 transaction = transaction)
+        for slicedBinds in grouper(availBinds, 10000):
```
ticoann: The same as above, I don't think slice is needed.

src/python/WMCore/WMBS/MySQL/Files/SetLocationForWorkQueue.py
```diff
@@ -36,14 +37,16 @@ def execute(self, lfns, locations, isDBS=True, conn=None, transaction=None):
...
         if isDBS:
-            self.dbi.processData(self.deleteSQL, binds, conn=conn,
-                                 transaction=transaction)
+            binds = []
+            for lfn in lfns:
+                binds.add({'lfn': lfn})
 
-        self.dbi.processData(self.insertSQL, locations, conn=conn,
-                             transaction=transaction)
+            for sliceBinds in grouper(binds, 10000):
```
ticoann: The same as above, I don't think slice is needed.

src/python/WMCore/WMBS/MySQL/Files/SetLocationForWorkQueue.py
```diff
@@ -36,14 +37,16 @@ def execute(self, lfns, locations, isDBS=True, conn=None, transaction=None):
...
+                self.dbi.processData(self.deleteSQL, sliceBinds, conn=conn,
+                                     transaction=transaction)
+
+        for sliceBinds in grouper(locations, 10000):
```
ticoann: The same as above, I don't think slice is needed.

ticoann:
Alan, I think slicing bind variable is not necessary unless you seed the performance boost with this.
If there is performance boost, I think it should be fixed in DBCore.processData not the all the calls which  calls that.
Also, only change to set on necessary list (with duplicate checking).
I think we should keep it simple (only applying the changes we are sure about the performance increase i.e. list to set when duplicate is not allowed.

amaltaro:
Thanks for the review Seangchan.
I'm _almost_ sure this binds slicing gives us some performance gain, I'll write some performance tests for the next week. One or two of those slicing I'd like to keep specially because we know at which stage of the insertion/update we are, otherwise we only know when it begins and that's it.

About the list to set replacement, I tried to change it only where we have stuff like "insert ignore" etc etc or we check for duplicates in the python code, so we _should_ be safe.

Since the slicing is harmless (to be confirmed), I'm pushing it to vocms0257 now. Thanks again!

src/python/WMCore/WMBS/MySQL/Files/SetLocationForWorkQueue.py
```diff
@@ -36,14 +37,16 @@ def execute(self, lfns, locations, isDBS=True, conn=None, transaction=None):
...
-        binds = []
-        for lfn in lfns:
-            binds.append({'lfn': lfn})
-
         if isDBS:
-            self.dbi.processData(self.deleteSQL, binds, conn=conn,
-                                 transaction=transaction)
+            binds = []
+            for lfn in lfns:
+                binds.add({'lfn': lfn})
```
amaltaro: Darn it! I saw this one failing in vocms0257, it turns out it was supposed to be an append, fixing it.

ticoann:
closing this.

PR #8534, opened 2018-03-26
Avoid couch race condition when updating agent config to aux db

amaltaro:
I know it's not a nice fix, but we have what we have :)
I've seen issues in testbed and in my private VMs, not in production (yet).


amaltaro:
test this please

PR #8535, opened 2018-03-27
Minor performance update to the splitting API

amaltaro:
I noticed a difference of about a second (or two) between standard splitting call and splitting/update_only.
Also remove `lheInputFiles` from the updatable splitting parameters.


amaltaro:
I'm removing the do not merge label, but let's wait for the cmsweb production upgrade. We can then push it in May deployment.

ticoann (approved):

ticoann:
I would say that is major improvement. It is 10x faster. I am surprised it make that much difference. 
Patch looks good to me. Merging it.

PR #8536, opened 2018-03-28
check the PATH for condor_chirp command

ticoann:
@hufnagel, Dirk, is this what you were thinking to get the right path in Singularity container?
I am not sure whether PATH would be set correctly in Singularity. We can find all the path as well but that would take long time (might be better to look for certain paths if we knows some candidate)

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executor.py
+++ b/src/python/WMCore/WMSpec/Steps/Executor.py
@@ -12,6 +12,7 @@ import sys
 import json
 import subprocess
 
+from Utils.FileTools import getFullPath
 from WMCore.FwkJobReport.Report import Report
 from WMCore.WMSpec.WMStep import WMStepHelper
 from WMCore.WMSpec.Steps.StepFactory import getStepEmulator
@@ -175,9 +176,11 @@ class Executor:
             condor_config_dir = os.path.dirname(condor_config)
             condor_chirp_bin = os.path.join(condor_config_dir, 'main/condor/libexec/condor_chirp')
             if not os.path.isfile(condor_chirp_bin):
-                condor_chirp_bin = None
+                # Singularity container might not have CONDOR_CONFIG env.
+                #TODO: It should have fixed from Singularity setting
+                condor_chirp_bin = getFullPath("condor_chirp")
 
-        if condor_chirp_bin:
+        if condor_chirp_bin and os.access(condor_chirp_bin, os.X_OK):
             args = [ condor_chirp_bin, 'set_job_attr_delayed', key, json.dumps(value) ]
             returncode = subprocess.call(args)
 
--- a/test/python/Utils_t/FileTools_t.py
+++ b/test/python/Utils_t/FileTools_t.py
@@ -77,6 +77,12 @@ class testFileTools(unittest.TestCase):
         self.assertEqual(info['Size'], 34)
         return
 
+    def test_getFullPath(self):
+        fullPath = FileTools.getFullPath("cd")
+        #assuming there is path for cd
+        self.assertIsNotNone(fullPath)
+        fullPath = FileTools.getFullPath("this_shouldnt_be")
+        self.assertEqual(fullPath, None)
 
 if __name__ == "__main__":
     unittest.main()
--- a/src/python/Utils/FileTools.py
+++ b/src/python/Utils/FileTools.py
@@ -106,3 +106,15 @@ def findMagicStr(filename, matchString):
         for line in logfile:
             if matchString in line:
                 yield line
+
+def getFullPath(name, envPath="PATH"):
+    """
+    :param name: file name
+    :param envPath: any environment variable specified for path (PATH, PYTHONPATH, etc)
+    :return: full path if it is under PATH env
+    """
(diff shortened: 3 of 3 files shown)


src/python/WMCore/WMSpec/Steps/Executor.py
```diff
@@ -19,6 +19,14 @@
...
 getStepErrorDestination = lambda step: WMStepHelper(step).getErrorDestinationStep()
 
+#TODO this function can me moved to Utils.
+def which(name):
+    for path in os.getenv("PATH").split(os.path.pathsep):
+        full_path = path + os.sep + name
+        if os.path.exists(full_path):
+            return full_path
+    return None
+
```
ticoann: This code doesn't check whether file is executable (assumes condor_chirp is executable).

src/python/WMCore/WMSpec/Steps/Executor.py
```diff
@@ -19,6 +19,18 @@
 getStepName = lambda step: WMStepHelper(step).name()
 getStepErrorDestination = lambda step: WMStepHelper(step).getErrorDestinationStep()
 
+#TODO this function can me moved to Utils.
```
hufnagel: Yeah, that would have been my comment, a function like this really belongs into Utils.

src/python/WMCore/WMSpec/Steps/Executor.py
```diff
@@ -19,6 +19,18 @@
...
+def getFullPath(name):
+    """
+    :param name: file name
+    :return: full path if it is under PATH env
+    """
+    for path in os.getenv("PATH").split(os.path.pathsep):
+        fullPath = path + os.sep + name
```
hufnagel: fullPath = os.path.join(path, name)

src/python/WMCore/WMSpec/Steps/Executor.py
```diff
@@ -175,7 +187,9 @@ def setCondorChirpAttrDelayed(self, key, value):
             condor_config_dir = os.path.dirname(condor_config)
             condor_chirp_bin = os.path.join(condor_config_dir, 'main/condor/libexec/condor_chirp')
             if not os.path.isfile(condor_chirp_bin):
-                condor_chirp_bin = None
+                # Singularity container might not have CONDOR_CONFIG env.
+                #TODO It should have fixed from Singularity setting
+                condor_chirp_bin = getFullPath("condor_chirp")
 
         if condor_chirp_bin:
```
hufnagel: Per your other comment, maybe we should add

and os.access(condor_chirp_bin, os.X_OK)

hufnagel (changes requested):

ticoann:
@hufnagel, Dirk I updated the patch as you suggested. 
@amaltaro, Alan, could you review the patch as well

amaltaro (approved):
Patch looks good, Seangchan.
I'm just wondering whether we could make it slightly more generic, where one also provides the environment variable (e.g. PATH, PYTHONPATH, etc) as parameters to getFullPath.

test/python/Utils_t/FileTools_t.py
```diff
@@ -77,6 +77,11 @@ def test_fileInfo(self):
...
         return
 
-
+    def test_getFullPath(self):
+        fullPath = FileTools.getFullPath("cd")
+        #assuming there is path for cd
+        self.assertIsNotNone(fullPath)
+        fullPath = FileTools.getFullPath("this_shouldnt_be")
+        self.assertEqual(fullPath, None)
 if __name__ == "__main__":
```
amaltaro: we could have a couple of blank new lines above

hufnagel (approved):

ticoann:
Thanks Alan, I updated the patch as you suggested. I will run the test in my vm whether it breaks anything, and if it goes OK. I can patch  cmsgwms-submit6 hepcloud agent.

amaltaro:
Thanks, Seangchan. It looks good!

hufnagel:
Would check with Ops if this is wanted in all agents as well since it affects all sites that use singularity. I'll definitely add it to the Tier0 deployment instructions.

ticoann:
I create the branch patch (#8538) and updated hepcloud agent. (My vm test at least didn't break anything - I am not sure it will fix the problem though since PATH is not guaranteed to set correctly for condor_chirp) If the patch fixes the problem we can apply to that to all agent)

bbockelm:
Hi @ticoann -

CMSSW searches `$PATH` for Singularity and CMSSW-based jobs have been reporting without a problem.

Brian

amaltaro:
> CMSSW searches $PATH for Singularity and CMSSW-based jobs have been reporting without a problem.

well, what's the problem then (if any)? How can we check whether a job is properly reporting from inside CMSSW? I think we can run a workflow in a singularity-based site (Nebraska e.g.), fail all jobs, look up at Kibana and check the step exit code(?)

hufnagel:
It works. Jobs at PSC which have this problem chirp from within CMSSW just fine. With the patch they also chirp the exit codes from the WMCore runtime wrapper.

ticoann:
Thank you for checking, Dirk. I am merging this then.

PR #8538, opened 2018-03-29
#8536 1.1.10_wmagent branch (check the PATH for condor_chirp command)

ticoann:
wmagent branch for #8536


ticoann:
hepcloud agent (cmsgwms-submit6) patched and wiki page is updated.  
Alan I didn't update the trello since it is not applied to production agent. (Also not sure about fixing the problem)

PR #8539, opened 2018-03-29
Publish CMSSW versions via htcondor class ad

### LINKED ISSUE #7786: Providing CMSSW version info to ElasticSearch
Doing analysis in ElasticSearch it is very useful to have the CMSSW version that was used by the job. I'd like to ask that this information is made available for exporting to ElasticSearch.

amaltaro:
Fixes #7786

Create an htcondor class ad to keep all CMSSW versions to be used in the job. Format is a comma separated string following the order as cmsRun1, cmsRun2, cmsRun3, etc.

### DIFF
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -608,7 +608,7 @@ class JobCreatorPoller(BaseWorkerThread):
                     tempDict = {}
                     tempDict.update(processDict)
                     tempDict['jobGroup'] = wmbsJobGroup
-                    tempDict['swVersion'] = wmTask.getSwVersion()
+                    tempDict['swVersion'] = wmTask.getSwVersion(allSteps=True)
                     tempDict['scramArch'] = wmTask.getScramArch()
                     tempDict['jobNumber'] = jobNumber
                     tempDict['agentNumber'] = self.agentNumber
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -372,7 +372,7 @@ class JobSubmitterPoller(BaseWorkerThread):
                        'possibleSites': frozenset(possibleLocations),  # abort and drain sites filtered out
                        'potentialSites': frozenset(potentialLocations),  # original list of sites
                        'scramArch': loadedJob.get("scramArch", None),
-                       'swVersion': loadedJob.get("swVersion", None),
+                       'swVersion': loadedJob.get("swVersion", []),
                        'proxyPath': loadedJob.get("proxyPath", None),
                        'estimatedJobTime': loadedJob.get("estimatedJobTime", None),
                        'estimatedDiskUsage': loadedJob.get("estimatedDiskUsage", None),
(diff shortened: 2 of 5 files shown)


hufnagel:
I would suggest to stay with chirping conventions, i.e.

WMCore_cmsRun1_ScramArch
WMCore_cmsRun2_ScramArch
...

WMCore_cmsRun1_CMSSW
WMCore_cmsRun2_CMSSW
...

hufnagel:
It would be hard to filter on this stuff in ElasticSearch if you throw it all into one field in some weird format.

hufnagel:
Oh, never mind, you don't plan to chirp this but just return it via classadd. Why though ? Wouldn't it be more useful to chirp this ?

amaltaro:
I think what Andrea Sciaba and Brian want (from the issue above) is to have it in both places, CMSSW condor_chirp'ing and having them in the classads/WMCore.

hufnagel:
Ok, classads you can call whatever. You evaluate them in python code usually, so can take them apart. ElasticSearch is limited in it's query choices, a compressed format is very hard to use.

bbockelm:
Actually, ElasticSearch is quite good at handling lists (regardless, we post-process in Python...). I’d prefer to keep it all in one attribute in case we ever need to match on it again.

What attribute name does CRAB use?

hufnagel:
My problem is that different to DESIRED_Sites, the scram arches for the different steps are actually not equivalent, i.e. order matters and also you assume an order that is not part of the attribute itself. Which is to say, cmsRun1 followed by cmsRun2 and cmsRun4 etc is a convention that you need to know when evaluating this list. Plus, we already have the convention of one attribute per step for the exit codes.

hufnagel:
@bbockelm , out of interest, in ES, how would you compare a specific element in a list against a value ?

bbockelm:
But still - in the current form, it’s more flexible as it can potentially be used in matchmaking in the future.  This would be beneficial if we ever had to run at a HPC with limited releases in a container.

There’s no problem having an ordered list here.

The current exit code setup is done as there really isn’t any choice. I think that it’s certainly valid for having a different approach for what’s done at job runtime versus submit time.

hufnagel:
I agree with the matching argument. Still, for queries later, individual values seem better to me. But as long as you can index list elements for comparison you can still do what you want.

Anything speaking against doing both ?

amaltaro:
There are already gazillions of classads and you want to have even more redundant ones? Why don't we keep this single classad here and make the required changes to chirp the CMSSW version from cmsRun? 

About the crab ad name, apparently they aren't using anything generic:
https://github.com/dmwm/CRABServer/blob/258293c55df1dd8c1431e76b9863067e0da64e75/src/python/TaskWorker/Actions/DagmanCreator.py#L658

how about someone asking them to use the same as we have in this PR to facilitate things on the monitoring land?

hufnagel:
As long as we chirp the expanded names I am happy, I don't really care having a classad for it.

amaltaro:
Changes look good. Here is how the classad looks like for a single and multi-step jobs:
```
cmst1@vocms0263:/data/srv/wmagent/current $ condor_q -l 199.125 | egrep 'CMSSW_Versions|WMAgent_RequestName'
CMSSW_Versions = "CMSSW_9_4_2"
WMAgent_RequestName = "amaltaro_TaskChain_NanoAOD_April18_Validation_180403_114124_6394"
cmst1@vocms0263:/data/srv/wmagent/current $ 
cmst1@vocms0263:/data/srv/wmagent/current $ condor_q -l 202.121 | egrep 'CMSSW_Versions|WMAgent_RequestName'
CMSSW_Versions = "CMSSW_7_1_25_patch2,CMSSW_8_0_21,CMSSW_8_0_21"
WMAgent_RequestName = "amaltaro_StepChain_MC_April18_Validation_180403_114150_5243"
```

ticoann (approved):

ticoann:
Thanks Alan,

PR #8542, opened 2018-04-03
Pass PYTHONPATH with the WMCore libraries to the CMSSW environment

amaltaro:
As a follow up of https://its.cern.ch/jira/browse/CMSCOMPPR-2560

Starting in this 10_1_0 release, CMSSW is no longer using PYTHONPATH but instead PYTHON27PATH environment variable. That means we do not have PYTHONPATH in the CMSSW environment anymore (which we used to have for older releases, while running the scram runtime step).

This patch sets PYTHONPATH when it's not available in the scram/cmssw runtime environment.

### DIFF
--- a/src/python/WMCore/WMRuntime/Tools/Scram.py
+++ b/src/python/WMCore/WMRuntime/Tools/Scram.py
@@ -25,13 +25,13 @@ sample usage:
 
 """
 
+import logging
 import os
 import os.path
-import sys
 import subprocess
-import logging
-from PSetTweaks.WMTweak import readAdValues
+import sys
 
+from PSetTweaks.WMTweak import readAdValues
 
 ARCH_TO_OS = {'slc5': ['rhel6'], 'slc6': ['rhel6'], 'slc7': ['rhel7']}
 
@@ -145,10 +145,10 @@ class Scram(object):
 
         """
         result = ""
-        if self.architecture != None:
+        if self.architecture is not None:
             result += "export SCRAM_ARCH=%s\n" % self.architecture
 
-        if self.initialise != None:
+        if self.initialise is not None:
             result += "%s\n" % self.initialise
             result += """if [ "$?" -ne "0" ]; then exit 2; fi\n"""
         return result
@@ -194,7 +194,7 @@ class Scram(object):
         Scram runtime command
 
         """
-        if self.projectArea == None:
+        if self.projectArea is None:
             msg = "Scram Runtime called with Project Area not set"
             self.stdout = msg
             self.stderr = ""
@@ -252,21 +252,20 @@ class Scram(object):
         Run the command in the runtime environment
 
         """
-        if self.projectArea == None:
+        if self.projectArea is None:
             msg = "Scram Project Area not set/project() not called"
             raise RuntimeError(msg)
         if self.runtimeEnv == {}:
             msg = "Scram runtime environment is empty/runtime() not called"
             raise RuntimeError(msg)
         executeIn = runtimeDir
-        if runtimeDir == None:
+        if runtimeDir is None:
             executeIn = self.projectArea
         # if the caller passed a filename and not a filehandle and if the logfile does not exist then create one
         if isinstance(logName, basestring) and not os.path.exists(logName):
-            f = open(logName, 'w')
-            f.write('Log for recording SCRAM command-line output\n')
-            f.write('-------------------------------------------\n')
-            f.close()
+            with open(logName, 'w') as f:
+                f.write('Log for recording SCRAM command-line output\n')
+                f.write('-------------------------------------------\n')
         logFile = open(logName, 'a') if isinstance(logName, basestring) else logName
         bashcmd = "/bin/bash"
         if cleanEnv:
@@ -275,7 +274,7 @@ class Scram(object):
                                 stdout=logFile,
                                 stderr=logFile,
                                 stdin=subprocess.PIPE,
-                               )
+                                )
 
         # Passing the environment in to the subprocess call results in all of
         # the variables being quoted which causes problems for search paths.
@@ -294,11 +293,11 @@ class Scram(object):
         if "PYTHONPATH" not in self.runtimeEnv:
             self.procWriter(proc, 'export PYTHONPATH=%s\n' % os.environ['PYTHONPATH'])
 
-        if os.environ.get('VO_CMS_SW_DIR', None) != None:
+        if os.environ.get('VO_CMS_SW_DIR', None) is not None:
             self.procWriter(proc, 'export VO_CMS_SW_DIR=%s\n' % os.environ['VO_CMS_SW_DIR'])
-        if os.environ.get('OSG_APP', None) != None:
+        if os.environ.get('OSG_APP', None) is not None:
             self.procWriter(proc, 'export VO_CMS_SW_DIR=%s/cmssoft/cms\n' % os.environ['OSG_APP'])
-        if os.environ.get('CMS_PATH', None) != None:
+        if os.environ.get('CMS_PATH', None) is not None:
             self.procWriter(proc, 'export CMS_PATH=%s\n' % os.environ['CMS_PATH'])
         if os.environ.get('_CONDOR_JOB_AD'):
             self.procWriter(proc, 'export _CONDOR_JOB_AD=%s\n' % os.environ['_CONDOR_JOB_AD'])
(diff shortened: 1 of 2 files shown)


davidlange6:
Hi

With only a phone...  PYTHONPATH usage is fine and supported. I think your only problem is that the code you have assumes that PYTHONPATH is set and should instead check first if it is set or not.

Cheers,
David


On 3 Apr 2018, at 14:36, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


As a follow up of https://its.cern.ch/jira/browse/CMSCOMPPR-2560

Starting in this 10_1_0 release, CMSSW is no longer using PYTHONPATH but instead PYTHON27PATH environment variable, so we need to make sure this one also has a path to the WMCore libraries.

@davidlange6<https://github.com/davidlange6>, should we set it with the WMCore path only or add all the PYTHONPATH paths to it as well?

You can view, comment on, or merge this pull request online at:

  https://github.com/dmwm/WMCore/pull/8542

Commit Summary

  *   Add WMCore libraries to PYTHON27PATH

File Changes

  *   M etc/submit.sh<https://github.com/dmwm/WMCore/pull/8542/files#diff-0> (1)
  *   M src/python/WMCore/WMRuntime/Unpacker.py<https://github.com/dmwm/WMCore/pull/8542/files#diff-1> (4)

Patch Links:

  *   https://github.com/dmwm/WMCore/pull/8542.patch
  *   https://github.com/dmwm/WMCore/pull/8542.diff

davidlange6:
In fact you should not use PYTHON27PATH as it won’t work on old releases.


On 3 Apr 2018, at 14:45, David Lange <notifications@github.com<mailto:notifications@github.com>> wrote:

Hi

With only a phone... PYTHONPATH usage is fine and supported. I think your only problem is that the code you have assumes that PYTHONPATH is set and should instead check first if it is set or not.

Cheers,
David


On 3 Apr 2018, at 14:36, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com><mailto:notifications@github.com>> wrote:


As a follow up of https://its.cern.ch/jira/browse/CMSCOMPPR-2560

Starting in this 10_1_0 release, CMSSW is no longer using PYTHONPATH but instead PYTHON27PATH environment variable, so we need to make sure this one also has a path to the WMCore libraries.

@davidlange6<https://github.com/davidlange6>, should we set it with the WMCore path only or add all the PYTHONPATH paths to it as well?

You can view, comment on, or merge this pull request online at:

https://github.com/dmwm/WMCore/pull/8542

Commit Summary

* Add WMCore libraries to PYTHON27PATH

File Changes

* M etc/submit.sh<https://github.com/dmwm/WMCore/pull/8542/files#diff-0> (1)
* M src/python/WMCore/WMRuntime/Unpacker.py<https://github.com/dmwm/WMCore/pull/8542/files#diff-1> (4)

Patch Links:

* https://github.com/dmwm/WMCore/pull/8542.patch
* https://github.com/dmwm/WMCore/pull/8542.diff

amaltaro:
We don't know which variable is going to be used by CMSSW python, thus we need to make sure both environment variables are set and that both contain the WMCore libraries path. Thus, I expect it to work with old releases too (it works with 10_1_0).
Am I missing something?
PS.: Thanks for looking at it from your phone ;)

davidlange6:
No. You do. Cmssw python supports PYTHONPATH just like all python distributions

Cheers,
David


On 3 Apr 2018, at 15:10, Alan Malta Rodrigues <notifications@github.com<mailto:notifications@github.com>> wrote:


We don't know which variable is going to be used by CMSSW python, thus we need to make sure both environment variables are set and that both contain the WMCore libraries path. Thus, I expect it to work with old releases too (it works with 10_1_0).
Am I missing something?
PS.: Thanks for looking at it from your phone ;)

etc/submit.sh
```diff
@@ -160,6 +160,7 @@ $pythonCommand Unpacker.py --sandbox=$SANDBOX --package=JobPackage.pkl --index=$
 cd job
 export WMAGENTJOBDIR=$PWD
 export PYTHONPATH=$PYTHONPATH:$PWD/WMCore.zip:$PWD
+export PYTHON27PATH=$PYTHON27PATH:$PWD/WMCore.zip:$PWD  # CMSSW_10_1_0 no longer uses PYTHONPATH
```
davidlange6: I would suggest just changing 162-163 into:

if [ -z "$PYTHONPATH" ]
then
  export PYTHONPATH=$PWD/WMCore.zip:$PWD
else
  export PYTHONPATH=$PYTHON27PATH:$PWD/WMCore.zip:$PWD
fi

or some improved version of this
[ps, for later - do you really need to add $PWD?]

davidlange6: after more checking it seems my suggestion should not be needed.

smuzaffar:
Ok I think I found the issue. Problem is here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Tools/Scram.py#L273
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Tools/Scram.py#L286

You are starting from a clean env and then added the environment variables returned by SCRAM. As previously scram had return PYTHONPATH (which included DMWM path too) that is why it was working. Now as SCRMA do not return PYTHONPATH so in your sub-shell it is not passed. I would suggest to add something like this
```
if not 'PYTHONPATH' in self.runtimeEnv:
  self.runtimeEnv['PYTHONPATH']=os.environ['PYTHONPATH']
```

amaltaro:
Thanks for this investigation, Shahzad!

> Now as SCRAM does not return PYTHONPATH so in your sub-shell it is not passed.

I'm still not sure this is the actual problem, otherwise this patch would not work as well, since we don't set (again) PYTHON27PATH for the new shell, only at the very top level.

smuzaffar:
this should work. Basically this is what you are doing
1. set PYTHONPATH via submit.sh script
2. Run scram to get the env variables
3. reset the env and then set all env variables which were set by SCRAM 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Tools/Scram.py#L286

Previously SCRAM was setting PYTHONPATH, so in step2 scram was returning cmssw PYTHONPATH + yours (which you had set in step1). Now as scram is not setting PYTHONPATH , so it is not returning it when you run `scram ru -sh`.

trust me this should work :-)

amaltaro:
Yes, you're right Shahzad! The PYTHONPATH is not returned from the scram.runtime() code, only PYTHON27PATH (for the newer release). So when we execute cmsRun1 with the cmssw python, it loads only the libraries found in the latter variable (which never got the WMCore library path added to).

I know you wanted to make sure PYTHON27PATH would have only python2.7 compatible libraries, but shouldn't Scram return PYTHON27PATH with the paths in PYTHONPATH (in case PYTHON27PATH wasn't originally set)? Otherwise the same issue will happen once CMSSW decides to use PYTHON3PATH and so on. In the end, I think it's easier to change it in production than making sure every single user will have a proper python setup :)

smuzaffar:
Alan, scram does not know about PYTHON27PATH or PYTHONPATH and I would not recommend to change scram to treat some env variables differently. 
When we move to PYTHON3 then scram will be instructed (via tool files) to set PYTHON3xPATH variable and python3 should work properly.

If DMWM scripts are setting some environment (e.g. PYTHONPATH) then they should make sure that that env is passed to sub-shell.

amaltaro:
That's another way to say it :) What matters is that the behavior of scram runtime has changed (according to the CMSSW version) and what used to be set is no longer.

BTW, I updated this patch as we discussed. Thanks for helping to fix it!

amaltaro:
test this please

ticoann:
Thanks Alan, Shahzad, patch looks good to me.

ticoann (approved):

belforte:
@ticoann could you kindly apply these changes also in 1.1.2_crab branch and create a new tag 1.1.2.crab6 with them ?
Looks to me that this is an isolated enough change that should not require importing/checking any other of the other 865 (sigh)  commits by which Master is ahead

ticoann:
@belforte, Hi Stefano, this patch is currently testing. Although I don't think we will have the problem. 

We applied production agent with below patch. (Although this patch is more correct approach, below patch works as well)
https://github.com/dmwm/WMCore/pull/8543

Would you want me to apply 8543 instead, in case you don't have time to run the test, or would you prefer wait a couple of days until we validate this patch?

belforte:
@ticoann let's wait a few days and you only do this once (hopefully). We will need to test this on our end anyhow.

ticoann:
@belforte, Hi Stefano, patch is tested now (with wmagent 1.1.12.patch1). So I create the crab branch tag with the patch. 1.1.2.crab6. Please let me know if you find any problem.

belforte:
thanks @ticoann I tested in my VM and it works fine. I will deploy on pre-prod CRAB

PR #8551, opened 2018-04-09
add worker node info to wmstats

### LINKED ISSUE #7908: report the worker node in wmstats
that is an information that must be available and would be very useful in many cases

ticoann:
fixes #7908 (only for 1.1.12.patch version of agents)


ticoann:
It is not tested yet. I will test and report.

PR #8552, opened 2018-04-09
remove alert components

ticoann:
@amaltaro, Alan I left Alert config for the API. It seems it is used for WorkQueue.
If this looks OK. I will create the wmagent branch (bulk branch with other patches).

I will clean  up the actual code for next release.


amaltaro:
Yes, I also kept the top level Alert configuration. I suggest we wait a couple of days before we backport it to the wmagent branch, there will be more patches.

PR #8553, opened 2018-04-09
clean up Alert manager

### LINKED ISSUE #7566: Remove unused component.
AlertGenerator, AlertProcessor. - replaced by LogDB
WorkflowManager - never been used
WMBSService,  - HTTPFrontEnd/ - not used by anyone anymore

ticoann:
fixes #7566 
Need to clean up the deployment script and WMAgentConfig.
zmq should be still kept since it is used in ProcessPool (Although ProcessPool is not explicitly used. 
Check the DBS code.
Still need to test.


amaltaro:
Seangchan, I didn't look for all the ZMQ code, but it seems you removed a bunch in this PR. Didn't you say we need to keep that?

Another comment - better later than never ;) - looking at the AlertGenerator.py and its Pollers class, maybe we could reuse a big slice of that code (removing alerts, signals and thresholds part) and repurpose that component to one that is watching the performance of the components and updating it to MonIT. What do you think?

ticoann:
> Seangchan, I didn't look for all the ZMQ code, but it seems you removed a bunch in this PR. Didn't you say we need to keep that?

It seems ZMQ is used for killing the Process as well as Alert. I remove the Alert part but not in ProcessPool. However I am not sure we are using ProcessPool anywhere as well. (I thought it is used in  JobCreator, JobSubmitter but I didn't see in the code. 

> Another comment - better later than never ;) - looking at the AlertGenerator.py and its Pollers class, maybe we could reuse a big slice of that code (removing alerts, signals and thresholds part) and repurpose that component to one that is watching the performance of the components and updating it to MonIT. What do you think?

That would be fine, but isn't it easier to create a brand new one. Do you see any reusable code?

PR #8555, opened 2018-04-10
Enable agent speed drain

goughes:
Some initial commits for the agent speed drain.  Still need to check thresholds and update the agent configuration.

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
@@ -10,8 +10,9 @@ __all__ = []
 import logging
 from Utils.Timers import timeFunction
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
-from WMCore.Services.ReqMgrAux.ReqMgrAux import isDrainMode
+from WMCore.Services.ReqMgrAux.ReqMgrAux import isDrainMode, ReqMgrAux
 from WMComponent.AgentStatusWatcher.DrainStatusAPI import DrainStatusAPI
+from WMCore.Services.PyCondor.PyCondorAPI import PyCondorAPI
 
 class DrainStatusPoller(BaseWorkerThread):
     """
@@ -27,6 +28,13 @@ class DrainStatusPoller(BaseWorkerThread):
         BaseWorkerThread.__init__(self)
         self.config = config
         self.drainAPI = DrainStatusAPI()
+        self.condorAPI = PyCondorAPI()
+        self.speedDrainConfig = None
+
+        if hasattr(config, "Tier0Feeder"):
+            self.reqAuxDB = None
+        else:
+            self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
 
     @timeFunction
     def algorithm(self, parameters):
@@ -35,7 +43,13 @@ class DrainStatusPoller(BaseWorkerThread):
         """
         if isDrainMode(self.config):
             logging.info("Checking agent drain status...")
-
+            # check to see if the agent hit any speed drain thresholds
+            if self.reqAuxDB:
+                logging.info("Checking speed drain thresholds...")
+                thresholdsHit = self.checkSpeedDrainThresholds()
+                if thresholdsHit:
+                    logging.info("Updating agent configuration for speed drain...")
+                    self.updateAgentSpeedDrainConfig(thresholdsHit)
             try:
                 DrainStatusPoller.drainStats = self.drainAPI.collectDrainInfo()
                 logging.info("Finished collecting agent drain status.")
@@ -54,3 +68,69 @@ class DrainStatusPoller(BaseWorkerThread):
         Return drainStats class variable
         """
         return cls.drainStats
+
+    def updateAgentSpeedDrainConfig(self, thresholdsHit):
+        """
+        Takes a list of speed drain configuration keys and updates the agent configuration
+        """
+        updateConfig = False
+
+        if 'CondorPriority' in thresholdsHit:
+            if self.condorAPI.editCondorJobs(
+                    "JobStatus=?=1 && (CMS_JobType =?= \"Production\" || CMS_JobType =?= \"Processing\")",
+                    "JobPrio", "999999"):
+                logging.info("Condor job priority updated to 999999 for pending Production/Processing jobs.")
+                if not self.speedDrainConfig['CondorPriority']['Enabled']:
+                    self.speedDrainConfig['CondorPriority']['Enabled'] = True
+                    updateConfig = True
+
+        if 'NoJobRetries' in thresholdsHit:
+            if self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "MaxRetries", 0):
+                logging.info("Agent updated max retries to 0.")
+                self.speedDrainConfig['NoJobRetries']['Enabled'] = True
+                updateConfig = True
+
+        if 'EnableAllSites' in thresholdsHit:
+            logging.info("Updating agent to submit to all sites.")
+            # setting this value to True makes JobSubmitterPoller ignore site status
+            self.speedDrainConfig['EnableAllSites']['Enabled'] = True
+            updateConfig = True
+
+        # update the aux db speed drain config with any changes
+        if updateConfig:
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainMode", True)
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainConfig", self.speedDrainConfig)
+
+        return
+
+    def checkSpeedDrainThresholds(self):
+        """
+        Check the current number of jobs in Condor and create a list of agent configuration parameters
+        that need updated for speed draining
+        """
+        validKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
+        enableKeys = []
+
+        # get the current speed drain status
+        self.speedDrainConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainConfig")
+
+        # get condor jobs
+        jobs = self.condorAPI.getCondorJobs("", [])
+        if jobs is None:
+            logging.warning("There was an error querying the schedd.  Not checking speed drain thresholds.")
+            return []
+
+        # loop through the speed drain configuration and make a list of what thresholds have been hit
+        for k, v in self.speedDrainConfig.items():
+            # make sure keys in the speed drain config are valid
+            if k in validKeys and isinstance(v['Threshold'], int) and isinstance(v['Enabled'], bool):
+                # we always want to apply the condor priority change if the threshold is hit
+                if not v['Enabled'] or k == 'CondorPriority':
+                    logging.info("Checking speed drain threshold for %s. ", k)
+                    if len(jobs) < v['Threshold']:
+                        logging.info("Agent will update speed drain configuration for %s. ", k)
+                        enableKeys.append(k)
+            else:
+                logging.warning("Speed drain configuration error for %s.  Please check aux db contents.", k)
+
+        return enableKeys
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -88,6 +88,9 @@ class JobSubmitterPoller(BaseWorkerThread):
         self.condorOverflowFraction = 0.2
         self.ioboundTypes = ('LogCollect', 'Merge', 'Cleanup', 'Harvesting')
 
+        # Used for speed draining the agent
+        self.enableAllSites = False
+
         # Additions for caching-based JobSubmitter
         self.jobsByPrio = {}  # key'ed by the final job priority, which contains a set of job ids
         self.jobDataCache = {}  # key'ed by the job id, containing the whole job info dict
@@ -275,6 +278,9 @@ class JobSubmitterPoller(BaseWorkerThread):
 
         logging.info("Found %s new jobs to be submitted.", len(newJobs))
 
+        if self.enableAllSites:
+            logging.info("Agent is in speed drain mode. Submitting jobs to all possible locations.")
+
         logging.info("Determining possible sites for new jobs...")
         jobCount = 0
         for newJob in newJobs:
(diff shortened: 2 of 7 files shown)


goughes:
@ticoann @amaltaro Take a look at this code for speed draining.  Currently it will detect if the # of condor jobs is under a threshold defined in aux db.  I also added code to update maxRetries to 0 in aux db. These two mechanisms are tested on my agent and are working.

I added the 'Enabled' key to the speed drain config in aux db so I could keep track of what config changes had been executed.  This prevents making the config changes over and over again in the drain thread.

I still need to implement the methods for condor priority increase and the db updates for enabling all sites.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -123,6 +123,18 @@ def updateAgentDrainingMode(self, agentName, drainFlag):
                                    (drainFlag, resp))
             return False
 
+    def updateSpeedDrainConfig(self, agentName, speedDrainConfig):
```
amaltaro: And how about replacing these two new methods by a generic one like `updateWMAgentConfig`

goughes: I add this as updateAgentConfig.  Given a key/value pair, update the appropriate agent aux db configuration.

amaltaro:
@goughes while I review this code, can you please figure out and fix the failing unit test
WMComponent_t.JobSubmitter_t.JobSubmitterCaching_t.JobSubmitterCachingTest:testCaching changed from success to error
?

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -27,6 +29,14 @@ def __init__(self, config):
         BaseWorkerThread.__init__(self)
         self.config = config
         self.drainAPI = DrainStatusAPI()
+        self.condorAPI = PyCondorAPI()
+
+        if hasattr(config, "Tier0Feeder"):
+            self.reqAuxDB = None
+            self.speedDrainConfig = None
```
amaltaro: I think this `self.speedDrainConfig` line can be set above, outside the if-else block.
Then in the algorithm method, just below the "Checking agent drain status" log, we add something like
```
if self.reqAuxDB:
    self.checkSpeedDrainThresholds()
```

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -54,3 +68,46 @@ def getDrainInfo(cls):
         Return drainStats class variable
         """
         return cls.drainStats
+
+    def checkSpeedDrainThresholds(self, speedDrainConfig):
+        """
+
+        """
+        validKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
```
amaltaro: Then here you start with
```
self.speedDrainConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainConfig")
```

Don't forget the docstring :)

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -54,3 +68,46 @@ def getDrainInfo(cls):
...
+        for k, v in speedDrainConfig.items():
+            # make sure keys are valid
+            if k in validKeys and isinstance(v['Threshold'], int) and isinstance(v['Enabled'], bool):
+                # and the drain config change isn't active
+                if not v['Enabled']:
+                    logging.info("Checking threshold for %s", k)
+                    # if number of condor jobs is below the threshold, activate speed drain config
+                    if len(jobs) < v['Threshold']:
+                        updateConfig = True
+                        if k == 'CondorPriority':
```
amaltaro: A minor change, which might make this code more readable. How about setting these actions after this logic that checks whether it is enabled or not and whether it has to be enabled.

Something like after the `for` loop, we could
```
if self.speedDrainConfig['CondorPriority']['Enabled']:
  res = self.condorAPI.editCondorJobs("JobStatus=?=1 && (CMS_JobType =?= \"Production\" || CMS_JobType =?= \"Processing\")", "JobPrio", "999999"):
  if res:
    logging.info("Threshold hit for %s, Updated condor priority to 999999 for all jobs.", k)
  else:
    blah
if self.speedDrainConfig['NoJobRetries']['Enabled']:
```

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -54,3 +68,46 @@ def getDrainInfo(cls):
...
+                        elif k == 'NoJobRetries':
+                            if self.reqAuxDB.updateMaxRetries(self.config.Agent.hostName, 0):
+                                logging.info("Threshold hit for %s. Updated max retries to 0.", k)
+                                speedDrainConfig[k]['Enabled'] = True
+                        elif k == 'EnableAllSites':
+                            logging.info("Updating agent to submit to all sites.")
+                            # setting this value to True makes JobSubmitterPoller ignore site status
+                            speedDrainConfig[k]['Enabled'] = True
+                else:
+                    logging.info("Speed drain configuration is active for %s", k)
```
amaltaro: This is not correct for `CondorPriority`, which is supposed to go through a continuous update (when it's enabled).

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -317,15 +317,18 @@ def refreshCache(self):
...
             if len(possibleLocations) == 0:
                 newJob['fileLocations'] = loadedJob.get('fileLocations', [])
                 newJob['siteWhitelist'] = loadedJob.get('siteWhitelist', [])
                 newJob['siteBlacklist'] = loadedJob.get('siteBlacklist', [])
                 badJobs[71101].append(newJob)
                 continue
-            else:
+
+            # if agent is in speed drain and has hit the threshold to submit to all sites, we can skip the logic to exclude sites
+            if not self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainConfig")['EnableAllSites']['Enabled']:
```
amaltaro: how about we break this line down to 2 such that it's more readable?
We also need to check whether `if self.reqAuxDB:` before we call getWMAgentConfig

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -44,6 +44,30 @@ def getCondorJobs(self, constraint, attr_list):
...
+        success = False
+        try:
+            self.schedd.edit(job_spec, attr, value)
+            success = True
+        except Exception as ex:
+            # edit doesn't distinguish between an error and not matching any jobs
+            # check for this message and assume it just didn't match any jobs
+            if isinstance(ex, RuntimeError) and str(ex) == "Unable to edit jobs matching constraint":
+                success = True
+                msg = "Condor constraint did not match any jobs. "
```
amaltaro: this should be a logging.info only. For the else problem, then I agree with logging.exception

amaltaro (changes requested):
I still have to review ReqMgrAux code, but I made several comments/suggestions for the rest of the files.

amaltaro (changes requested):
Thanks, Erik! Patch is looking much better, there are only a few more adjustments to be done and it should be ready.

goughes:
test this please

ticoann:
test this please

ticoann:
test this please

amaltaro:
test this please

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -740,9 +750,9 @@ def algorithm(self, parameters=None):
         if self.useReqMgrForCompletionCheck:
             # only runs when reqmgr is used (not Tier0)
             self.removeAbortedForceCompletedWorkflowFromCache()
-            agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
-            self.condorFraction = agentConfig.get('CondorJobsFraction', 0.75)
-            self.condorOverflowFraction = agentConfig.get("CondorOverflowFraction", 0.2)
+            self.agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
+            self.condorFraction = self.agentConfig.get('CondorJobsFraction', 0.75)
```
amaltaro: how about making agentConfig a local variable and then
`self.enableAllSites = agentConfig.get("SpeedDrainConfig")['EnableAllSites']['Enabled']`
?

If so, then the code above has to be updated and in the __init__ method we should add
`self.enableAllSites = False`

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -316,15 +317,22 @@ def refreshCache(self):
...
                 newJob['fileLocations'] = loadedJob.get('fileLocations', [])
                 newJob['siteWhitelist'] = loadedJob.get('siteWhitelist', [])
                 newJob['siteBlacklist'] = loadedJob.get('siteBlacklist', [])
                 badJobs[71101].append(newJob)
                 continue
-            else:
+
+            # see if the agent is in speed drain mode for enabling all sites
+            enableAllSites = False
+            if self.reqAuxDB:
```
amaltaro: see comment below

amaltaro (commented):
Erik, I went through this PR again and I have some minor suggestions. Please update this PR and let me know such that I can run some tests too.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -12,6 +12,8 @@
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.Services.ReqMgrAux.ReqMgrAux import isDrainMode
 from WMComponent.AgentStatusWatcher.DrainStatusAPI import DrainStatusAPI
+from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
```
amaltaro: This import can go up 2 lines, in the same line of isDrainMode

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -741,6 +750,7 @@ def algorithm(self, parameters=None):
             # only runs when reqmgr is used (not Tier0)
             self.removeAbortedForceCompletedWorkflowFromCache()
             agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
+            self.enableAllSites = agentConfig.get("SpeedDrainConfig")['EnableAllSites']['Enabled']
```
amaltaro: Erik, please update this line to
```
if agentConfig.get("SpeedDrainConfig"):
    self.enableAllSites = agentConfig["SpeedDrainConfig"]['EnableAllSites']['Enabled']
```
to avoid issues when couch is not available.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -54,3 +69,66 @@ def getDrainInfo(cls):
...
+    def updateAgentSpeedDrainConfig(self, thresholdsHit):
+        """
+        Takes a list of speed drain configuration keys and updates the agent configuration
+        """
+        if 'CondorPriority' in thresholdsHit:
+            if self.condorAPI.editCondorJobs(
+                    "JobStatus=?=1 && (CMS_JobType =?= \"Production\" || CMS_JobType =?= \"Processing\")",
+                    "JobPrio", "999999"):
+                logging.info("Condor job priority updated to 999999 for pending Production/Processing jobs.")
+                self.speedDrainConfig['CondorPriority']['Enabled'] = True
```
amaltaro: Before setting it to True (and updating the config), we should check whether it's false, otherwise it will keep updating this same key over and over.
In addition to that, I think we need to define a `updateConfig = False` at the very first line of this method.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
bbockelm: Just a thought - feel free to commit one-liners directly to Erik's branch.  Given the timezone differences, it often makes things simpler to iterate...

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -333,17 +338,20 @@ def refreshCache(self):
...
+                if newJob['task_type'] not in self.ioboundTypes:
+                    nonDrainingSites = [x for x in possibleLocations if x not in self.drainSites]
+                    if nonDrainingSites:  # if >1 viable non-draining site remove draining ones
+                        possibleLocations = nonDrainingSites
+                    else:
+                        newJob['possibleSites'] = possibleLocations
+                        badJobs[71104].append(newJob)
+                        continue
+            else:
+                logging.info("Agent is in speed drain mode. Submitting jobs to all possible locations.")
```
amaltaro: This logging can't be here since we're iterating over every potential job to be submitted.
I'd place it around L280 
```
if self.enableAllSites:
    logging....
```

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
amaltaro: One line change is enough to break someone else's commit :) However I agree with that if we need to get something in asap

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
goughes: When the thresholds are checked, line 126 of DrainStatusPoller.py should prevent any unnecessary updates for this key

`if not v['Enabled'] or k == 'CondorPriority':`

amaltaro: And that's why CondorPriority is always added to the enableKeys var, provided it's under the threshold. Then the actual couchdb update is done just below L99

goughes: Oh wait a sec... I was looking at EnableAllSites instead of CondorPriority.  You are right.  Let me add the logic.

amaltaro (approved):

amaltaro:
@goughes thanks for all the follow up changes, this PR looks good now and it's ready to go.
Before Seangchan merge it, can you please squash all the real code changes (src/) into one commit and leave test/unittests/pylints in a second commit? Thanks

PR #8557, opened 2018-04-11
revert a change that turned out to not bee needed

hufnagel:


hufnagel:
This will need to go into the 1.1.12 wmagent branch as well.

ticoann:
Thanks Dirk. I will create the 1.1.12_wmagent branch

ticoann:
wmagent branch patch is added in #8558

PR #8563, opened 2018-04-12
upload lumi text file from AlcaHarvest

hufnagel:
AlcaHarvest will produce a luminosity text file with information for the LHC.

Since this is uploaded to a different EOS system (not CMS EOS), refactor the AlcaHarvest code a bit to be able to pass in upload URL for both condition and luminosity output.

Only touches code that is used in the Tier0 (apart from adding a new method in Utilities.py, which is also harmless).

This should be backported to the 1.1.12 wmagent branch.

### DIFF
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -3,6 +3,8 @@
 from __future__ import division, print_function
 
 import subprocess
+import os
+import re
 
 
 def makeList(stringList):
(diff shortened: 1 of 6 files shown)


amaltaro:
Dirk, can you please create a simple unit test for the Utilities function?

hufnagel:
I anyways need to fix some of the existing unit tests. On it.

hufnagel:
Question, can I see the output of the unit tests somewhere ? Just knowing a certain unit test failed doesn't narrow it down much...

amaltaro:
Usually we have a link from the "default" tests, just above. This one doesn;t have it though, let me trigger the tests again.

amaltaro:
test this please

hufnagel:
Ok, that should look good now.

ticoann:
@hufnagel, Do you need a branch (1.1.12_wmagent) patch for this?

hufnagel:
Yep, I labeled it as such :-)

ticoann:
merging it since this only affects agent

PR #8568, opened 2018-04-16
fix: not closing cursor above the limit

### LINKED ISSUE #8506: fix the returnCursor or remove it

ticoann:
fixes #8506
@vkuznet, Valentin, this was the code we discussed at the meeting. Could you take a look?

### DIFF
--- a/src/python/WMCore/Database/DBCore.py
+++ b/src/python/WMCore/Database/DBCore.py
@@ -37,6 +37,10 @@ class DBInterface(WMObject):
         self.engine = engine
         self.maxBindsPerQuery = 500
 
+    def _generateCursor(self, sql, binds):
+        for bind in binds:
+            yield self.engine.execute(sql, bind)
+
     def buildbinds(self, sequence, thename, therest=[{}]):
         """
         Build a list of binds. Can be used recursively, e.g.:
@@ -95,9 +99,10 @@ class DBInterface(WMObject):
             Trying to select many
             """
             if returnCursor:
-                result = []
-                for bind in b:
-                    result.append(connection.execute(s, bind))
+                if b:
+                    return self._generateCursor(s, b)
+                else:
+                    return []
             else:
                 result = ResultSet()
                 for bind in b:
@@ -128,6 +133,7 @@ class DBInterface(WMObject):
 
         """
         connection = None
+        closeConnection = True
         try:
             if not conn:
                 connection = self.connection()
@@ -155,9 +161,13 @@ class DBInterface(WMObject):
                 #Run single SQL statement for a list of binds - use execute_many()
                 if not transaction:
                     trans = connection.begin()
-                for subBinds in grouper(binds, self.maxBindsPerQuery):
-                    result.extend(self.executemanybinds(sqlstmt[0], subBinds,
-                                                        connection=connection, returnCursor=returnCursor))
+                if sqlstmt[0].lower().endswith('select', 0, 6) and returnCursor:
+                    return self.executemanybinds(sqlstmt[0], binds, connection=connection,
+                                                 returnCursor=returnCursor)
+                else:
+                    for subBinds in grouper(binds, self.maxBindsPerQuery):
+                        result.extend(self.executemanybinds(sqlstmt[0], subBinds,
+                                            connection=connection, returnCursor=returnCursor))
 
                 if not transaction:
                     trans.commit()
--- a/test/python/WMCore_t/Database_t/DBCore_t.py
+++ b/test/python/WMCore_t/Database_t/DBCore_t.py
@@ -11,8 +11,37 @@ Unit tests for the DBInterface class
 import unittest
 import logging
 import threading
+import types
 
 from WMQuality.TestInit import TestInit
+
+
+def formatCursor(cursor, size=10):
+    """
+    Fetch the driver cursor directly.
+    Tested only with cx_Oracle.
+    Cursor must be already executed.
+    Use fetchmany(size = default arraysize = 50)
+
+    """
+    if type(cursor.keys) == types.MethodType:
+        keys = [x.lower() for x in cursor.keys()]
+    else:
+        keys = [x.lower() for x in cursor.keys]
+    result = []
+    while True:
+        if not cursor.closed:
+            rows = cursor.fetchmany(size=size)
+            if not rows:
+                cursor.close()
+                break
+            for r in rows:
+                result.append(dict(list(zip(keys, r))))
+        else:
+            break
+    if not cursor.closed:
+        cursor.close()
+    return result
 
 class DBCoreTest(unittest.TestCase):
     def setUp(self):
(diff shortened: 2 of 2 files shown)


ticoann:
Here is the error.

```
ERROR: testProcessDataSeveralBindsWithReturnCursor (__main__.DBCoreTest)
  File "DBCore_t.py", line 272, in testProcessDataSeveralBindsWithReturnCursor
    for cursor in resultCursor:
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/Database/DBCore.py", line 42, in _generateCursor
    yield connection.execute(sql, bind)
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 719, in execute
    return self._execute_text(object, multiparams, params)
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 868, in _execute_text
    statement, parameters
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 891, in _execute_context
    None, None)
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1116, in _handle_dbapi_exception
    exc_info
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/util/compat.py", line 189, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb)
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 885, in _execute_context
    conn = self._revalidate_connection()
  File "/home/dmwm/unittestdeploy/wmagent/1.1.10.pre7/sw/slc7_amd64_gcc630/external/py2-sqlalchemy/0.9.6-comp/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 246, in _revalidate_connection
    raise exc.ResourceClosedError("This Connection is closed")
StatementError: This Connection is closed (original cause: ResourceClosedError: This Connection is closed) u'SELECT column1, column2, column3 FROM test_tablea\n             WHERE column1 = %s AND column2 = %s AND column3 = %s' [(1, 2, 'three')]
```

vkuznet (commented):
Seangchan, I'm not sure if the code will work. The _generateCursor function is a generator. It means that in order to get results from it someone need to iterate over, otherwise it will produce nothing. So, if you pass a connection object there and no-one iterate over this function before connection closed the code will behave as you described. What should be done is that you should not pass connection to it and rather acquire it within this function.

ticoann:
Thank you Valentin, for the reviw. 

Main difference with original code is 
Original code returns the list of cursor and updated code returns the generator.
We just assume for now the caller always iterates over the what is return without using list specific method (i.e. len)

```
# returns list

for bind in b:
       result.append(connection.execute(s, bind))
return result

--->
# returns cursor

def _generateCursor(self, sql, binds, connection):
        for bind in binds:
            yield connection.execute(sql, bind)

return self._generateCursor(s, b, connection)
```

In the unittest, we are iterating the whatever above returns, (list, generator)
First case (list works), second case (generator) fails with above error.
```
for cursor in resultCursor:
    result.extend(formatCursor(cursor))
```

Not sure why (I suspect there is something wrong with below method)

```
def formatCursor(cursor, size=10):
    """
    Fetch the driver cursor directly.
    Tested only with cx_Oracle.
    Cursor must be already executed.
    Use fetchmany(size = default arraysize = 50)

    """
    if type(cursor.keys) == types.MethodType:
        keys = [x.lower() for x in cursor.keys()]
    else:
        keys = [x.lower() for x in cursor.keys]
    result = []
    while True:
        if not cursor.closed:
            rows = cursor.fetchmany(size=size)
            if not rows:
                cursor.close()
                break
            for r in rows:
                result.append(dict(list(zip(keys, r))))
        else:
            break
    if not cursor.closed:
        cursor.close()
    return result
```

ticoann:
FYI, Below test didn't have any problem. (I am probably not understanding how the cursor works)

```
def genIter(b):
    for i in b:
        yield i

def printList(b):
    if len(b) > 2:
        return genIter(b)
    else:
        return b

for d in printList([1,2,3]):
    print(d)

for d in printList([1,2]):
    print(d)
```

vkuznet:
Look at the code flow:
- in processData code acquires the connection at [1]
- for certain conditions you return executemanybinds results [2]
  - the executemanybinds is a generatator since it calls _generateCursor here [3]
- but processData has a finally statement at [4] which explicitly closes the
connection

So, what you end-up with is that connection is closed because of [4], but you
return to the caller list of cursors from [3]. Since cursors were never
called (iterated yet) you get the error.

If you replace generate [3] with list everything will work since you
get all cursors before you closed connection in [4].

As I said generator is lazy object, think of this as placeholder, it will
be called only someone will start iteration or call next().

To fix the problem you either need to iterator over your generator (but it
defeats the purpose) or correctly deal with connections, e.g. acquire it
within _generateCursor code itself since it will stay alive when you
return the generator.


[1] https://github.com/ticoann/WMCore/blob/9e835fc00614f2ae559f77f7d3bb2dac0225020d/src/python/WMCore/Database/DBCore.py#L138
[2] https://github.com/ticoann/WMCore/blob/9e835fc00614f2ae559f77f7d3bb2dac0225020d/src/python/WMCore/Database/DBCore.py#L138
[3] https://github.com/ticoann/WMCore/blob/9e835fc00614f2ae559f77f7d3bb2dac0225020d/src/python/WMCore/Database/DBCore.py#L103
[4] https://github.com/ticoann/WMCore/blob/9e835fc00614f2ae559f77f7d3bb2dac0225020d/src/python/WMCore/Database/DBCore.py#L206

ticoann:
Thank Valentin, Now I understand. It means the cursor is create already, we can access the data even if connection is closed, correct?  
Anyway it means there are all the code using returnCursor True, has to handle the connection outside processData Method.

ticoann:
Valentin, Hmm, something interesting. I put create the connection outside the processData method and close the connection after all data is fetched and cursors are closed. Assuming it would be faster than if I create the new connection and close it for each cursor in _generateCursor as you suggested.
(Although, it is just getting it from the pool and returning to the pool).

> e.g. acquire it
> within _generateCursor code itself since it will stay alive when you
> return the generator.

However, I ran the test for 30K binding, there is no performance difference.

ticoann:
@vkuznet, Valentin, I updated the patch, please review.

amaltaro:
Seangchan, to keep the returned value consistent (when asked for returnCursor), how about changing L105 from
`return []`
to
`return iter(())`
?

Another comment, it took me a while to digest this syntax
`s.lower().endswith('select', 0, 6):`...
can't we just use `startswith('select')` instead (seems to be exactly what it's doing...)? There must have been a reason to make it this way though...

My last comment/question is, when we return a cursor, don't we need to make sure the connection gets closed? I don't see it getting closed in DBCore, since we simply return the cursor.

vkuznet:
Seangchan, I'm not sure what you want to accomplish with these changes, but here is few observations.

The code now seems a mixture of generator and lists, if you want to improve things, e.g. reduce memory footprint, then you better use only generators. For instance, result.extend basically allocates stuff which otherwise you can yield. Or, formatCursor don't really need to create internal result list, instead it can be easily converted to generator. But I don't know who use this class and therefore the migration to generator should be propagated upstream.

ticoann:
test this please

PR #8570, opened 2018-04-17
Scale EventsPerLumi by FilterEfficiency

### LINKED ISSUE #8510: Scale lumisection by filter efficiency
TaskChains can produce very small lumisections when many of the generated events are filtered out.

For example, if 99% of events are removed, a 300 event lumisection at the generator step becomes a 3 event lumisection for DIGI/RECO; the latter has seriously bad CPU efficiencies as we must serialize on lumi boundaries.

After discussion on the CPU efficiency mailing list, I think we've settled on tackling this at the computing level, not within the framework.

In particular, @ticoann and @vlimant suggested that we could apply the scaling here:

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/TaskChain.py#L490-L496

I would suggest only applying the scaling if the input filtered efficiency is a sane value (non-negative, not greater than 100%, etc).

This does imply that we will depend somewhat on upstream to provide sane values for request efficiency.

amaltaro:
Fixes #8510

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/MonteCarlo.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/MonteCarlo.py
@@ -26,6 +26,17 @@ class MonteCarloWorkloadFactory(StdBase):
     Stamp out Monte Carlo workflows.
     """
 
+    def __init__(self):
+        """
+        __init__
+
+        Setup parameters that will be later overwritten in the call,
+        otherwise pylint will complain about them.
+        """
+        super(MonteCarloWorkloadFactory, self).__init__()
+        self.eventsPerLumi = None
+        self.eventsPerJob = None
+
     def buildWorkload(self):
         """
         _buildWorkload_
@@ -95,8 +106,10 @@ class MonteCarloWorkloadFactory(StdBase):
         """
         StdBase.__call__(self, workloadName, arguments)
 
-        # Adjust the events by the filter efficiency
+        # Adjust the events and lumi size by the filter efficiency
         self.totalEvents = int(self.requestNumEvents / self.filterEfficiency)
+        if self.eventsPerLumi:
+            self.eventsPerLumi /= self.filterEfficiency
 
         # We don't write out every event in MC,
         # adjust the size per event accordingly
--- a/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
@@ -329,6 +329,8 @@ class StepChainWorkloadFactory(StdBase):
             taskConf["SplittingAlgo"] = "EventBased"
             taskConf["RequestNumEvents"] = int(requestNumEvts / filterEff)
             taskConf["SizePerEvent"] = self.sizePerEvent * filterEff
+            if taskConf.get("EventsPerLumi"):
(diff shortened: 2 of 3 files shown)


amaltaro:
FYI @anorkus , we'll push it to cmsweb-testbed this week and if everything goes fine, to production on 2nd of May.

amaltaro:
test this please

amaltaro:
@ticoann @vlimant to make sure I got the right changes, can you please review it?
I still have to run some tests, tomorrow.

amaltaro:
Seangchan, one thing just came to my mind regarding the EventsPerLumi in the workload json. Let me explain both scenarios:
1) current scenario: whatever user provides for EventsPerLumi will be saved to the workload json
2) new scenario: EventsPerLumi gets updated based on the filter efficiency

Question is, what happens if Ops clone a request in scenario 2)? Well, EventsPerLumi will be scaled once again based on the filter efficiency, and again and again.

I'd like to keep the workload json with the data provided by the user + default values + any other customization automatically applied by ReqMgr2, and make it consistent with the workload configuration. However, this might be (in this case it is) a problem when cloning requests.

Maybe we can chat later today.

ticoann:
> Question is, what happens if Ops clone a request in scenario 2)? Well, EventsPerLumi will be scaled once again based on the filter efficiency, and again and again.

That is good point. I didn't think about that. 
I have a question: Is modified EventsPerLumi updated in json stored in reqmgr couchdb or just in the spec?
(If it is the first case, I can think of treating this 3 ways)
1. clone and ACDC workflows ignore filter efficiency and take event per lumi. (or when it clones/ACDCes multiply filer efficiency back so it has original user setting EventsPerLumi)
2. We create EventsPerLumiWithoutFilter parameter separately and only allow to set that and EventsPerLumi is generated.

I think option 1.  is a little hacky. but option 2. might require changes outside WMAgent. Who is setting EventsPerLumi now, McM, Unified?

amaltaro:
Seangchan and I just discussed this problem and we see 3 possible solutions:
a) as Seangchan mentioned above, we define two other spec arguments: `EventsPerJobWithoutFilter` and `EventsPerLumiWithoutFilter` and McM can set either EventsPerLumi (scaled by filter) or EventsPerLumiWithoutFilter and so on. This way we can properly distinguish what is what.
b) We leave it as it is, thus McM doing the scaling before injecting into ReqMgr2
c) we keep the workload cache (json) inconsistent with the workload spec (which isnt really an option, we don't want to keep things inconsistent).

If we go for a), then we'll probably have to wait for the next cmsweb upgrade because I'm pretty sure it's not going to be **just** two new arguments.

@vlimant @bbockelm your input would be welcomed here.

amaltaro:
> a) as Seangchan mentioned above, we define two other spec arguments: EventsPerJobWithoutFilter and EventsPerLumiWithoutFilter and McM can set either EventsPerLumi (scaled by filter) or EventsPerLumiWithoutFilter and so on. This way we can properly distinguish what is what.

this is the approach we decided to take, however there is no need to have two EventsPerJob parameter. If user provides `EventsPerLumi`, we don't do anything besides properly setting it in the spec. If user provides `EventsPerLumiWithoutFilter`, then we apply the filter and update `EventsPerLumi` with that value.

PR #8572, opened 2018-04-18
Post global_workqueue data to MonIT

### LINKED ISSUE #8420: Inject global workqueue stats into monit IT
using the stompAMQ and the document structure used by Benjamin for the agent info
https://github.com/stiegerb/wmamon_es/blob/master/post_agentinfo.py#L92

amaltaro:
Fixes #8420 (besides the full list of sites).
Changes are:
* Add all the logic to get the WorkQueue Heartbeat thread to post data to MonIT as well (in addition to wmstats).
* for such cases, the data structure has to be changed such that it can be easily aggregated on the ES end
* when possible (any data returned from the workqueue internal queries), I try to create default docs such that we always have that info (even if 0)

It depends on https://github.com/dmwm/deployment/pull/603 and we need to add a stomp dependency on the workqueue RPM spec file.
And, we have to disable global_workqueue docs from @stiegerb script, once it goes into prod.

### DIFF
--- a/src/python/WMCore/GlobalWorkQueue/CherryPyThreads/HeartbeatMonitor.py
+++ b/src/python/WMCore/GlobalWorkQueue/CherryPyThreads/HeartbeatMonitor.py
@@ -1,34 +1,145 @@
 from __future__ import (division, print_function)
+
 from WMCore.REST.HeartbeatMonitorBase import HeartbeatMonitorBase
 from WMCore.WorkQueue.WorkQueue import globalQueue
+from WMCore.WorkQueue.DataStructs.WorkQueueElement import STATES
+
 
 class HeartbeatMonitor(HeartbeatMonitorBase):
 
+    def __init__(self, rest, config):
+        super(HeartbeatMonitor, self).__init__(rest, config)
+        self.initialStatus = ['Available', 'Negotiating', 'Acquired']
+        self.producer = "global_workqueue"
+        self.docTypeAMQ = "cms_%s_info" % self.producer
+
     def addAdditionalMonitorReport(self, config):
         """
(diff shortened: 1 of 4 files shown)


amaltaro:
WMCore functionality looks Ok, but unfortunately I cannot find any data in kibana:
https://cern.service-now.com/service-portal/view-request.do?n=RQF0996903

I still have to remove one of my loggings, but can you please review it @ticoann 
Benjamin, your input is also welcomed ;)

src/python/WMCore/GlobalWorkQueue/CherryPyThreads/HeartbeatMonitor.py
```diff
@@ -1,34 +1,180 @@
...
+                doc["type"] = "work_site_possibleAAA"
+                doc["status"] = status
+                doc.update(commonInfo)
+                doc.update(item)
+                docs.append(doc)
+
+        self.logger.info("%i docs created to post to MonIT", len(docs))
+        return docs
+
+    def uploadToAMQ(self, docs, producer="global_workqueue"):
```
ticoann: This seems to be exactly the same method as uploadToAMQ in reqmgr2 heartbeat monitor except producer. Maybe we should move this is base class?

amaltaro:
Seangchan, 2nd commit has all the changes you suggested. I'll redeploy it in my VM and run some quick tests. However, I think wmstats will now fail because we'll need to add the stomp dependency to the RPM (otherwise importing StompAMQ will fail).

amaltaro:
Chances look good, so I went ahead and squashed those commits.

ticoann:
Thanks Alan, It looks good to me.  I will merge this when Jenkins comes back.

PR #8573, opened 2018-04-20
fix report location when bootstrap error happens

ticoann:
When job fails in Bootstrap state. FrameWorkJob Report retry number is always 0. 
And WMAgent won't find the Repork.[retry_number].pkl file.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Bootstrap.py#L133
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Bootstrap.py#L141
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Bootstrap.py#L152

This it temporary patch for tier 0 agent. (Need to find the better solution) - Not tested yet.

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -258,9 +258,12 @@ class SimpleCondorPlugin(BasePlugin):
                 continue
 
             reportName = os.path.join(job['cache_dir'], 'Report.%i.pkl' % job['retry_count'])
+            reportNameWithBootstapError =  os.path.join(job['cache_dir'], 'Report.0.pkl')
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan, could you check whether this patch is necessary?
I think it fixes one case if Jobs failing repeatedly in bootstrap. (Although it is a little hacky way)

ticoann:
@amaltaro, Alan is this still valid? I remember you already fixed this and this is not the correct fix.

amaltaro:
We can probably have a better fix for this. I created this issue:
https://github.com/dmwm/WMCore/issues/8716

and I'm closing this PR. Feel free to reopen it if needed ;)

ticoann:
Closing it is good. Thanks Alan

PR #8575, opened 2018-04-20
temp fix for shutil is None

ticoann:
@belforte, Stefano, if this patch is OK. I will create a new tag. (1.1.2.crab7)


belforte:
thanks @ticoann I will try it and let you know

belforte:
It is fine, thanks. And sorry for my misunderstanding of the exception catching.

ticoann:
Thank you very much for testing, Stefano. 1.1.2.crab7 tag is created now.

PR #8583, opened 2018-04-24
fix wmstats report for data update time

### LINKED ISSUE #8347: WMStats should report AgentStatusWatcher instead of AnalyticsDataCollector
I found that AgentStatusWatcher was down (while AnalyticsDataCollector was up), even though wmstats reports an error with `status: Data is not updated: AnalyticsDataCollector Down; `

which basically masks any other component that might be down but not getting reported by AgentStatusWatcher.

ticoann:
fixes #8347

### DIFF
--- a/src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
+++ b/src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
@@ -1,25 +1,23 @@
 """
 Perform cleanup actions
 """
-__all__ = []
-
-
-
 import threading
 import logging
 import time
 
 from Utils.Timers import timeFunction
+from WMCore.WMFactory import WMFactory
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.Database.CMSCouch import CouchMonitor
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue as WorkQueueService
 from WMCore.Services.WMStats.WMStatsWriter import WMStatsWriter
 from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
-from WMComponent.AnalyticsDataCollector.DataCollectAPI import LocalCouchDBData, \
-     WMAgentDBData, combineAnalyticsData, convertToRequestCouchDoc, \
-     initAgentInfo
-from WMComponent.DBS3Buffer.DBSBufferUtil  import DBSBufferUtil
-from WMCore.WMFactory import WMFactory
+from WMComponent.AnalyticsDataCollector.DataCollectAPI import (LocalCouchDBData,
+                                            WMAgentDBData, combineAnalyticsData,
+                                            convertToRequestCouchDoc, initAgentInfo)
+
+from WMComponent.DBS3Buffer.DBSBufferUtil import DBSBufferUtil
+
 
 class AnalyticsPoller(BaseWorkerThread):
     """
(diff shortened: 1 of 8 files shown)


src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
```diff
@@ -142,10 +144,18 @@ def algorithm(self, parameters):
 
             self.localSummaryCouchDB.uploadData(requestDocs)
             logging.info("Request data upload success\n %s request, \nsleep for next cycle" % len(requestDocs))
-            DataUploadTime.setInfo(uploadTime, "ok")
+
+            self.centralWMStatsCouchDB.updateAgentInfoInPlace(self.agentInfo["agent_url"],
+                                                              {"data_last_update": uploadTime, "data_error": "ok"})
 
         except Exception as ex:
             logging.error("Error occurred, will retry later:")
```
amaltaro: How about replacing these error logging by one exception? traceback import can probably be removed then.

src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
```diff
@@ -142,10 +144,18 @@ def algorithm(self, parameters):
...
             logging.error(str(ex))
-            DataUploadTime.setInfo(False, str(ex))
             logging.error("Trace back: \n%s" % traceback.format_exc())
+
+            try:
+                self.centralWMStatsCouchDB.updateAgentInfoInPlace(self.agentInfo["agent_url"], {"data_error": str(ex)})
```
amaltaro: Just trying to understand. If there was an exception above, we immediately post this error to couch wmstats, which is visible through the agent_info view? Does it show up like a component error or something else?

amaltaro (approved):
Patch looks good, Seangchan.
I'm just wondering whether these new key/values updated in place (data_*) will get overwritten/removed by AgentStatusWatcher updates?

I need to eventually look at this component and check what it does :-) It seems it collects some interesting data, which is never used and never looked by anyone though...

ticoann:
> I'm just wondering whether these new key/values updated in place (data_*) will get overwritten/removed by AgentStatusWatcher updates?

That is good point, it will be removed every time AgentStatusWatcher updates. I will fix that.

src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
ticoann: Thanks, I will change that.

src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
ticoann: Yes it will be visible in agent_info view. It should show up "data_error".

ticoann:
@amaltaro, Alan I updated patch, Thanks.

amaltaro:
Seangchan, sorry for the late review. If it's not much work, it might be easier and cleaner to do something as Erik did for the DrainStatusPoller thread, using a class method to access the AnalyticsDataCollector info from AgentStatusWatcher and making sure that that never gets overwritten.

See this code for instance:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py#L51

if you like it, then I think it's worth it changing this PR again :)

ticoann:
Hi Alan, Actually that was the whole point of this patch (It was using class cache). 😄 But it doesn't work since those 2 components doesn't run in the same process, hence it doesn't share memory.

I could use the zmq for communicating between processes but I thought current patch is simpler.

PR #8593, opened 2018-05-01
add api for searching missing parentage

### LINKED ISSUE #8590: Create new API which retrieves parent files by lumi section
Precondition is at least dataset parentage needs be know.  

1. provide file parentage by given file and parent dataset
2. provide file parentage by given block and parent dataset
3. provide file parentage by given dataset and parent dataset

Option 3 will take big performance hit.

ticoann:
fixes #8590 
still need to discuss APIs' functions and signatures. Just initial attempt start discussion.

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -5,7 +5,11 @@ _DBSReader_
 Readonly DBS Interface
 
 """
+from __future__ import print_function, division
+
 import time
+import logging
+import traceback
 from collections import defaultdict
 
 from dbs.apis.dbsClient import DbsApi
@@ -15,6 +19,8 @@ from RestClient.ErrorHandling.RestClientExceptions import HTTPError
 from Utils.IteratorTools import grouper
 from WMCore.Services.DBS.DBSErrors import DBSReaderError, formatEx3
 from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
+from WMCore.Services.pycurl_manager import getdata as multi_getdata
+
 
 
 def remapDBS3Keys(data, stringify=False, **others):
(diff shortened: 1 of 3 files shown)


ticoann:
setting whole dataset parentage may require millions files to download.

1. shouldn't be supported 
2. or internally break down to blocks and returns generator.

Current code doesn't do either but added for information purpose.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -883,3 +885,167 @@ def listDatasetParents(self, childDataset):
...
+
+    def _getParentFilesWithDatasetOrBlock(self, childDataset=None, childBlock=None):
+        """
+
+        :param childDataset:
+        :param childBlock:
+        :return: list of dict [{"child": child_file, "parents": set(parentFiles)}, ...]
+        """
+        if childBlock:
+            childDataset = childBlock.split("/")[1]
```
amaltaro: `childBlock.split("#")[0]`

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -883,3 +885,167 @@ def listDatasetParents(self, childDataset):
...
+        :param childDataset:
+        :param searchFile: flag to search for all the files if it sets true, otherwise only search missing parent blocks
+        :param printFlag: true - print out some debugging statement
+        :return: structure DBS is using to update the parentage. (TODO: need to refine)
+        """
+
+        pDatasets = self.listDatasetParents(childDataset)
+        # print("parent datasets %s\n" % pDatasets)
+        if not pDatasets:
+            # print("No parents dataset found for %s\n" % dataset)
```
amaltaro: how about logging.error("No parent dataset found") or something like that?

amaltaro:
Seangchan, I made just a couple of comments for the moment. Please let me know when it's ready for review.
BTW, maybe we should create an abnormal dataset and use it in unit tests? Let me know and I can help you with that

ticoann:
@vkuznet, Valentin, this is not concurrent call. Could you review the code?

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -883,3 +885,167 @@ def listDatasetParents(self, childDataset):
             msg += "DBSReader.listDatasetParents(%s)\n" % childDataset
             msg += "%s\n" % formatEx3(ex)
             raise DBSReaderError(msg)
+
+    def getParentFilesGivenParentDataset(self, parentDataset, childLFN):
```
vkuznet: This function should be easily parallelizeable, but it requires complete change of its logic and tools. I'll provide here a recipe and pointers and you can look/adjust the code. First, you need pycurl_manager getdata function
```
```
Then you need to construct a list of urls you're going to concurrently send to DBS, e.g.
```
uri = /dbs_uri # replace with main DBS URI
urls = ["%s/parents?run_num=%s&lumi_list=%s" % (uri, f['run_num'], f['lumi_section_num']) for f in fInfo]
```
then you can implement similar logic as shown here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/RequestInfo.py#L150
For example:
```
parentFiles = []
for row in data:
     # request url
     req = row['url'].split('/')[-1]
     # response data
     data = json.loads(row['data'])
     # parse data here and extract file name from response
     parentFiles.append( data['logical_file_name'] ) # replace logical_file_name with whatever key DBS uses in response
```
So, to sum-up.
1. you will not use dbs client APIs since they do not support concurrency
2. you construct list of all urls you want to fetch at once
3. multi_getdata will do proper thing: it will place all calls concurrently and do not saturate DBS server (internally it split input list of urls in to a chunk (default size is 100) and send the chunks).
4. you'll collect results from pycurl and extract responses and construct your output.

vkuznet: @bbockelm , Brian, if I'm not mistaken requests library only provides API to fetch a single request from the server. Therefore to implement concurrency we either need to write asyncio code (but it is only available in python 3.5+) or write custom code via thread pool, etc. Therefore it till take some effort/code which we'll need to mantain. This is what we're doing with pycurl already. Plus pycurl basically calls curl C library, so it's is fast. How requests libraryunderlying logic is implemented I don't know. Seems like a good student project to me to extend requests library and implement concurrent APIs and perform benchmarks wrt pycurl.

ticoann:
Thanks Valentin, Brian for the review and comment. After a brief search in google, it seems what Valentin says is correct about asyncio and request (needs python 3.5+). Since Valentin already made wrapper class for pycurl. https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/pycurl_manager.py

I will try to look into that direction unless there is better way to handle it.

src/python/WMCore/Services/DBS/DBS3Reader.py
bbockelm: Sounds good, thanks for the update!

ticoann:
test this please

ticoann:
I was able to move all the functionality from old script to new script including handling ReReco case.
So deprecated-fix-dbs-parentage is only there for validation purpose. (I will delete that when validation is done for the next release. - So I didn't add documents in detail)

For DBSReader classes, print statement is still there to check the function, we can either make that optional or change to logging.

There is one commented out function for using pycurl which need some change either in dbs api or pycurl_manager frame work. (So please ignore that for now)

ticoann:
test this please

PR #8611, opened 2018-05-09
add getLumis flag

### LINKED ISSUE #8467: allow to ignore lumilists when using FileBased splitting
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/3554/2/1/1/1/2/2/1/2/2/1/1/1/1/1/1/1/1/1.html

this is needed by
https://github.com/dmwm/CRABServer/issues/5645

@ticoann @amaltaro @hufnagel let's use this to converge on the new signature for the SplitterFactory
so that I can make the proper changes to CRAB in
https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/Splitter.py

ticoann:
fixes #8467

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -285,7 +285,7 @@ class DBS3Reader(object):
 
         return
 
-    def listDatasetFileDetails(self, datasetPath, getParents=False, validFileOnly=1):
+    def listDatasetFileDetails(self, datasetPath, getParents=False, getLumis=True, validFileOnly=1):
         """
         TODO: This is completely wrong need to be redone. or be removed - getting dataset altogether
         might be to costly
@@ -324,14 +324,16 @@ class DBS3Reader(object):
                 for p in parents:
                     if p['logical_file_name'] in files:  # invalid files are not there if validFileOnly=1
                         files[p['logical_file_name']]['Parents'].extend(p['parent_logical_file_name'])
-            # get the lumis
-            file_lumis = self.dbs.listFileLumis(block_name=blockName)
-            for f in file_lumis:
-                if f['logical_file_name'] in files:  # invalid files are not there if validFileOnly=1
-                    if f['run_num'] in files[f['logical_file_name']]['Lumis']:
-                        files[f['logical_file_name']]['Lumis'][f['run_num']].extend(f['lumi_section_num'])
-                    else:
-                        files[f['logical_file_name']]['Lumis'][f['run_num']] = f['lumi_section_num']
+
+            if getLumis:
+                # get the lumis
+                file_lumis = self.dbs.listFileLumis(block_name=blockName)
+                for f in file_lumis:
+                    if f['logical_file_name'] in files:  # invalid files are not there if validFileOnly=1
+                        if f['run_num'] in files[f['logical_file_name']]['Lumis']:
+                            files[f['logical_file_name']]['Lumis'][f['run_num']].extend(f['lumi_section_num'])
+                        else:
+                            files[f['logical_file_name']]['Lumis'][f['run_num']] = f['lumi_section_num']
 
         return files
 


ticoann:
@belforte, Stefano, this is what you need, right? I should've put the getLumis parameter at the end after "valid"). But it seem it won't break WMCore and CRABServer code (use keywords parameter). I put it before "valid". If this cause the problem please let me know.

belforte:
+1
Perfect, thanks a lot. Yes parameter order is not relevant. I will test with this, but not immediately as I am traveling to Chicago today.

amaltaro (approved):
Change looks good. It would probably be useful though to have a unit test for this method.

belforte:
I tested this and it works finely for me. Please feel free to merge.
As usual I will need this to merged into 1.1.2_crab branch as well
Thanks a lot

amaltaro:
Stefano, just a side note. Can you plan to move CRAB to the newest stable release in the next month or so? It's not that often that we need to backport patches for CRAB, but CRAB is now using a year old tag and we might have problems backporting patches at any time.

belforte:
Alan, I fully agree.But my hours are what they are :-( Please see slide 10  here
https://indico.cern.ch/event/711343/contributions/2957803/attachments/1628968/2596393/CRAB_Operations-9Apr18.pdf
If someone could make the next stable release only a matter of testing it.. sure, but since there is a known non trivial problem.. I can't make any plan.

If only past developers did not have diverged ... now there are changes in CRAB branch which are not in Master one, in a way for good reasons, but it is not trivial to resolve. People intentionally choose to have different code. Usual.. it seemed a good idea at the time.
I suggest you or someone else does a branch comparison, then we talk again.

amaltaro:
Stefano, I think all those crab branch commits have made it to the master too:
https://github.com/dmwm/WMCore/compare/master...1.1.2_crab
Thanks for the slides, I see it's in your radar already.

belforte:
and yet 1.1.2_crab is 25 commits ahead of master (and 930 behind)
https://github.com/dmwm/WMCore/tree/1.1.2_crab

are those 25 the ones listed at
https://github.com/dmwm/WMCore/compare/master...1.1.2_crab
?

what bothers me mostly is https://github.com/dmwm/WMCore/commit/9efb4c87bd07c3db03981d06a028984211dd7056

why are those listed if as you say "they made it to the master too" ?
How can I check that they are ?

Yes, we could build with WMCore master and test, but CRAB has so many use case, and so few unit tests, that it will be impossible to be confident unless one at least knows where differences are.

amaltaro:
We usually tag the master branch PR in the specific branch PR, it wasn't that case. But here it is the master version of the commit you pointed out:
https://github.com/dmwm/WMCore/pull/7993

Maybe it doesn't show in the master branch because it wasn't cherry-picked? But I don't really know what makes it show up there or not (same commit hash?)

belforte:
Maybe sometimes people did something creative with git.
So I guess I have to checkout master and crab and use diff
to find differences if any in files "which are in the WMCore areas CRAB touches".
Sort of to get a sense of how high is the fall...
Will open a CRABgit issue :-)

ticoann:
It there is conflict between master version and branch version, we have to create new patches for each. May it is that case?

belforte:
let's hope is not. So far the only real thing is Stefano getting scared by the amount of differences.
What should the goal be here ? To use Master in CRAB or some branch ? If I go for master, how do I know what is the tag to use ? As I wrote in the slides, I'd like to stay aligned with you.

amaltaro:
Our goal is to use/share only stable releases. We cut a new CMSWEB stable release every month, but I think CRAB needs to use a WMAgent stable release, which are cut every two months.

We are going to cut a 1.1.14 release as soon as we get CMSWEB upgraded in June, then it takes a week or two to get a ready-for-production release, then we can ping you again (the latest one is 1.1.12.patch4).

ticoann:
Yes I agree that crab should aligned with WMCore master as close as possible. The only difference we need to concern is the library crab uses. (It shouldn't be that many). I think we should do following.

1. create the new CRAB branch from stable WMCore version as Alan mentioned above
2. Stefano can create a new crab version with new WMCore crab branch and test them.
If we get lucky we won't have any problem. :-) If not,
3. Stefano can provide WMCore library CRAB uses,
4. Alan/Seangchan can find the difference between 1.1.2_crab branch and and new crab branch and provide the new patches with that. 
Then go back to 1.

I think it shouldn't be more than 1 iteration ideally. But anyway, it will be useful to have the list what libraries crab uses. That way we will be more careful to update those libraries.

belforte:
thanks guys. Let's be concrete: I could do 1. with WMCore 1.1.12.patch4 ok ?
But then I should use some older crab tag which does not rely on thee lastest commits that you pushed for me in 1.1.2_crab7-8
Or is there a tag I can try which could have everything ? Some 1.1.14.pre ? or plainly Master ?

belforte:
perfect, thanks

belforte:
@ticoann Any news on the new tag you which will contain the patches for 1.1.2.crab.7. 8 ?
I'd guess that it exists, but you forgot to point it out.

belforte:
mayeb this ?
https://github.com/dmwm/WMCore/releases/tag/1.1.14

amaltaro:
@belforte 1.1.14 is the CMSWEB stable tag. I'm in the process of validating WMAgent on tag 1.1.14.patch1 (from this branch: https://github.com/dmwm/WMCore/tree/1.1.14_wmagent), but I have really just started that. I still want to add some more bug fixes for a component or two.

Should we create a new branch for CRAB (1.1.14_crab), cut a new release and you start from there?

belforte:
I was hoping to stop having a crab branch.  I will see how CMSWEB stable woks. I presume that difference in _wmagent branch are mostly (only?) in WMA code, so for me they'd be the same.

amaltaro:
Unless you don't want to have patch releases anymore, but simply work with patches on top of cmsweb stable release (which I think you don't), this is something we cannot run away.
The _wmagent branch might have all sorts of fixes, from component-wise code to external services interfaces.

PR #8613, opened 2018-05-10
Debug ErrorHandler memory crazyness - NEVER MERGE

### LINKED ISSUE #8614: ErrorHandler very high memory footprint while creating ACDCs
vocms0255 was going from 0 to 20GB within 20min or so.

The problem seems to be coming from this query 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMBS/MySQL/Jobs/LoadForErrorHandler.py#L43

when there are several MCFake file jobs to be handled (generating > 100k run/lumi maps).
Apparently we get the run/lumi map for unique fileids, but if we have 100 of them in the same binds, it might take all the memory...

amaltaro:
Fixes #8614
Debugging vocms0255, which is going from 0MB to 20GB within 20min or so...


amaltaro:
Replaced by #8672

PR #8615, opened 2018-05-11
Strip the requirement string and proxy for Volunteer jobs

amaltaro:
It's ugly, I don't like it and I'm not sure it works, but we have what we have...

This was Laurence's recommendation:
```
In the condor map file you map all DNs from the volunteers to an
identity e.g.

GSI "/O=Volunteer Computing/O=CERN/CN=.*" volunteer-node@cern.ch

This is essentially what we are doing now but we need to check that this
identity is only used for the volunteers.

My cluster can use an IGTF host certificate and we can use a different
mapping

***

You then need to ensure that all jobs have the followed requirement set.

Requirements = (AuthenticatedIdentity != volunteer-node@cern.ch).

For the jobs that you want to go to the volunteer machines, you remove
the requirement and strip the proxy.

To make this work we need to do four things.

1) My resources need to have that host certificate.
2) The opportunistic mapping needs to be added to the condor mapfile
3) The Requirement needs to be added to the jobs
4) The proxy striping needs to be extended to include removing the
requirement.
```

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -109,7 +109,8 @@ class SimpleCondorPlugin(BasePlugin):
         # TODO(bbockelm): Remove reqStr once HLT has upgraded.
         self.reqStr = ('((REQUIRED_OS=?="any") || '
                        '(GLIDEIN_REQUIRED_OS =?= "any") || '
-                       'stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS))')
+                       'stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS)) && '
+                       '(AuthenticatedIdentity =!= "volunteer-node@cern.ch")')
         if hasattr(config.BossAir, 'condorRequirementsString'):
             self.reqStr = config.BossAir.condorRequirementsString
 
@@ -118,6 +119,8 @@ class SimpleCondorPlugin(BasePlugin):
         self.x509userproxy = proxy.getProxyFilename()
         self.x509userproxysubject = proxy.getSubject()
         self.x509userproxyfqan = proxy.getAttributeFromProxy(self.x509userproxy)
+        # Remove the x509 ads if the job is matching a volunteer resource
+        self.x509Expr = 'ifThenElse("$$(GLIDEIN_CMSSite)" =?= "T3_CH_Volunteer",undefined,"%s")'
 
         return
 
@@ -467,7 +470,7 @@ class SimpleCondorPlugin(BasePlugin):
 
     def getClusterAd(self):
         """
-        _initSubmit_
+        _getClusterAd_
 
         Return common cluster classad
 
@@ -478,8 +481,6 @@ class SimpleCondorPlugin(BasePlugin):
         ad = classad.ClassAd()
 
         # ad['universe'] = "vanilla"
-        if self.reqStr:
-            ad['Requirements'] = classad.ExprTree(self.reqStr)
         ad['ShouldTransferFiles'] = "YES"
         ad['WhenToTransferOutput'] = "ON_EXIT"
         ad['UserLogUseXML'] = True
@@ -506,10 +507,6 @@ class SimpleCondorPlugin(BasePlugin):
 
         # Customized classAds for this plugin
         ad['DESIRED_Archs'] = "INTEL,X86_64"
-
-        ad['x509userproxy'] = self.x509userproxy
-        ad['x509userproxysubject'] = self.x509userproxysubject
-        ad['x509userproxyfirstfqan'] = self.x509userproxyfqan
 
         ad['Rank'] = 0.0
         ad['TransferIn'] = False
@@ -542,6 +539,15 @@ class SimpleCondorPlugin(BasePlugin):
             ad['Arguments'] = "%s %i" % (os.path.basename(job['sandbox']), job['id'])
 
             ad['TransferOutput'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]
+
+            # Do not define Requirements and X509 ads for Volunteer resources
+            if self.reqStr and "T3_CH_Volunteer" not in job.get('possibleSites'):
+                ad['Requirements'] = classad.ExprTree(self.reqStr)
+
+            ad['x509userproxy'] = classad.ExprTree(self.x509Expr % self.x509userproxy)
+            ad['x509userproxysubject'] = classad.ExprTree(self.x509Expr % self.x509userproxysubject)
+            ad['x509userproxyfirstfqan'] = classad.ExprTree(self.x509Expr % self.x509userproxyfqan)
+
 
             sites = ','.join(sorted(job.get('possibleSites')))
             ad['DESIRED_Sites'] = sites


bbockelm:
To do this "for real", we need to get the HTCondor folks to convert the `x509userproxy` attribute into something that can accept an expression.

Honestly, that *might* work already.  Can you try it?  Might be along the lines of this:

```
x509userproxy = ifThenElse("$$(GLIDEIN_CMSSite)" =?= "T3_US_Volunteer", UNDEFINED, "/path/to/file")
```

amaltaro:
That's smart! I'll try that expression out first.

hufnagel:
Question, where is that expression evaluated ? I just want to make sure it's not inside the pilot, since that wouldn't really help (we want to avoid the pilot having access to the prod certs at all, not just not using them at runtime).

bbockelm:
`$$(GLIDEIN_CMSSite)` is syntax for a "match expression" - it gets filled in on the schedd side based on the information sent by the negotiator as part of the match.

That means this should be present for all activities (job startup, running, shutdown) that involve a worker node.

The pilot is limited by the job's contents -- it cannot ask for files not specified in the job's description.

ddavila0:
The expression Requirements = (AuthenticatedIdentity != volunteer-node@cern.ch) is not likely to work on the CERN pools. Don't know if this is a bug on condor, but I have seen that the AuthenticatedIdentity gets lost when the machine classAd goes from the secondary to the primary collector. 
At the primary collector all machines show the same AuthenticatedIdentity, this corresponds to the AuthenticatedIdentity of the machine hosting the collector. 

This can easily observed if querying the ITB pool:

Querying the primary collector:
condor_status -pool vocms0809 -af AuthenticatedIdentity | sort | uniq -c
     54 condor@cms

Querying the secondary collector on port 9629
condor_status -pool vocms0809:9629 -af AuthenticatedIdentity
pilot04@cms
pilot04@cms

amaltaro:
Thanks, Diego. It seems to be yet another temporary solution then... we might have to hardcode stuff in the plugin and simply trust that nobody is going to condor_qedit jobs (including jobrouter)...

@bbockelm I know you're busy these days, but when you get a spare cycle, please advice. I don't want to get my proxy in a volunteer node (neither Ivan's proxy, since he has the cms:production role).

For the record, these are the job classads after the latest changes (noticed some of those x509 ads don't seem to get the Expression they were supposed to have(???)):
```
[cmsdataops@cmsgwms-submit1 current]$ condor_q -l 293.14 | egrep -i 'DESIRED_Sites|Requirements|X509'
DESIRED_Sites = "T1_US_FNAL,T2_CH_CERN"
Requirements = ((REQUIRED_OS =?= "any") || (GLIDEIN_REQUIRED_OS =?= "any") || stringListMember(GLIDEIN_REQUIRED_OS,REQUIRED_OS)) && (AuthenticatedIdentity != "volunteer-node@cern.ch") && TARGET.OPSYS == "LINUX" && TARGET.ARCH == "X86_64" && TARGET.HasFileTransfer && TARGET.Disk >= RequestDisk && TARGET.Memory >= RequestMemory
x509userproxy = ifThenElse("$$(GLIDEIN_CMSSite)" =?= "T3_CH_Volunteer",undefined,"/data/certs/myproxy.pem")
x509UserProxyEmail = "alan.malta@cern.ch"
x509UserProxyExpiration = 1527227702
x509UserProxyFirstFQAN = "/cms/Role=production/Capability=NULL"
x509UserProxyFQAN = "/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=amaltaro/CN=718748/CN=Alan Malta Rodrigues,/cms/Role=production/Capability=NULL,/cms/Role=NULL/Capability=NULL,/cms/uscms/Role=NULL/Capability=NULL"
x509userproxysubject = "/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=amaltaro/CN=718748/CN=Alan Malta Rodrigues"
x509UserProxyVOName = "cms"
```

```
[cmsdataops@cmsgwms-submit1 current]$ condor_q -l 293.33 | egrep -i 'DESIRED_Sites|Requirements|X509'
DESIRED_Sites = "T3_CH_Volunteer"
Requirements = true && TARGET.OPSYS == "LINUX" && TARGET.ARCH == "X86_64" && TARGET.HasFileTransfer && TARGET.Disk >= RequestDisk && TARGET.Memory >= RequestMemory
x509userproxy = ifThenElse("$$(GLIDEIN_CMSSite)" =?= "T3_CH_Volunteer",undefined,"/data/certs/myproxy.pem")
x509UserProxyEmail = "alan.malta@cern.ch"
x509UserProxyExpiration = 1527227702
x509UserProxyFirstFQAN = "/cms/Role=production/Capability=NULL"
x509UserProxyFQAN = "/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=amaltaro/CN=718748/CN=Alan Malta Rodrigues,/cms/Role=production/Capability=NULL,/cms/Role=NULL/Capability=NULL,/cms/uscms/Role=NULL/Capability=NULL"
x509userproxysubject = "/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=amaltaro/CN=718748/CN=Alan Malta Rodrigues"
x509UserProxyVOName = "cms"
```

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -544,6 +541,15 @@ def getProcAds(self, jobList):
 
             ad['TransferOutput'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]
 
+            # Do not define Requirements and X509 ads for Volunteer resources
+            if self.reqStr and "T3_CH_Volunteer" not in job.get('possibleSites'):
+                ad['Requirements'] = classad.ExprTree(self.reqStr)
+
+            ad['x509userproxy'] = classad.ExprTree(self.x509Expr % self.x509userproxy)
+            ad['x509userproxysubject'] = classad.ExprTree(self.x509Expr % self.x509userproxysubject)
```
amaltaro: and these 2 last lines somehow are not an expression in the condor classads. I guess they are automatically set from the proxy file or something, so classad expr won't work. Still, this one is probably a not a big deal to have in in the volunteer nodes.

amaltaro:
Looking at the very first job running on a volunteer node:
```
x509userproxy = ifThenElse("$$(GLIDEIN_CMSSite)" =?= "T3_CH_Volunteer",undefined,"/tmp/x509up_u31961")
MATCH_EXP_x509userproxy = ifThenElse("T3_CH_Volunteer" =?= "T3_CH_Volunteer",undefined,"/tmp/x509up_u31961")
Requirements = true && TARGET.OPSYS == "LINUX" && TARGET.ARCH == "X86_64" && TARGET.HasFileTransfer && TARGET.Disk >= RequestDisk && TARGET.Memory >= RequestMemory
```

I would say it's working, apparently.

amaltaro:
This seems to be working fine in the CMS@Home agent. Even though, I'm gonna run a few jobs at CERN and FNAL (global pool) before merging it into master.

amaltaro:
Tests in the global pool look Ok and the cms@home agent has been running with this patch for a week or two now. However, I recall we wanted to get rid of the requirements string because it was getting evaluated at 2 different level.

@ddavila0 @bbockelm @aperezca perhaps we should set the different requirements string at schedd level on the CMS@Home node?

bbockelm:
Hi @amaltaro -

Did you confirm that the proxy was actually not transferred in the CMS@Home case?  We need to double check that before moving forward.

I don't understand the comment about the requirements string.  What's different on the CMS@Home node?

Brian

amaltaro:
> Did you confirm that the proxy was actually not transferred in the CMS@Home case? We need to double check that before moving forward.

I did not connect to the worker node myself (because I have no privileges), but the file path was wrong in the classad and it didn't cause the job to fail. Also Laurence and Ivan confirmed that there was no proxy or my DN in the volunteers node.

> I don't understand the comment about the requirements string. What's different on the CMS@Home node?

Some time ago we discussed on removing that reqStr since it's already set at another level of the glideinwms system. We didn't remove it because HLT needed, for some reason that I don't remember.
Since it has that specific bit (AuthenticatedIdentity...) for volunteers node (the one which guarantees the proxy is not shipped together), I wonder whether it's fine to have such expression for all jobs when < 1% of them would really need that expression?

bbockelm:
> Also Laurence and Ivan confirmed that there was no proxy or my DN in the volunteers node.

Ok, that's good enough for me.

> Some time ago we discussed on removing that reqStr since it's already set at another level of the glideinwms system. 

Ok, now I recall what you are talking about.  It's a worthwhile cleanup but let's do this in a separate ticket.

amaltaro:
Ok, thanks Brian. I'm squashing it and pushing it in into the new agent release.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
@ddavila0 @aperezca @jamesletts as mentioned before, this one is going to go to production in the next week.

PR #8618, opened 2018-05-15
add getLumis flag

ticoann:
1.1.2_crab branch for #8611


ticoann:
@belforte, Stefano, 1.1.2.crab8 tag is created

PR #8621, opened 2018-05-16
Add retry count to the job classad

amaltaro:
As requested by Brian @bbockelm 
Please let me know whether we should set it to a generic name such that we can have it in common with analysis.


bbockelm:
@belforte -

It looks like CRAB uses `CRAB_Retry`.  Should we have WMCore use the same thing or just come up with a common attribute?

belforte:
A common attribute sounds better

bbockelm:
Since it's a CMS-specific attribute, I'd prefer to use some sort of prefix.  `CMS_JobRetryCount` is fine by me.

PR #8164, opened 2017-09-13
Remove xmlData from subscribe/delete APIs since it is provided by sub…

ticoann:
…scription/deletion parameter 

replace #8144

PR #8624, opened 2018-05-22
Fix phedex delete call - wmagent version

amaltaro:
1.1.10 wmagent version of #8625 
We missed this delete call when these changes https://github.com/dmwm/WMCore/pull/8164 were applied, and now (well, for half an year, the T0 agent fails to delete blocks...)


amaltaro:
I meant to backport it to 1.1.12 instead of 1.1.10...

PR #8625, opened 2018-05-22
Fix phedex delete call

amaltaro:
Vytas found a problem in the PhEDExInjector component logs where the delete call was failing, due to the changes applied in https://github.com/dmwm/WMCore/pull/8164


amaltaro:
Vytas already tested it on the T0 nodes.

ticoann (approved):

PR #8629, opened 2018-05-27
Temp patch which disables PhEDEx block deletion for T0 agent

amaltaro:
There *might* be a problem with the T0 agent and how it makes block deletion, see discussion in:
https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/1900.html

Honestly, I doubt there is a logic issue since it hasn't been changed for a long time (the last patch was only to fix the delete method signature)

FYI @hufnagel @vytjan and Dima (I don't know his GH user)

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
@@ -174,7 +174,7 @@ class PhEDExInjectorPoller(BaseWorkerThread):
 
             if self.pollCounter == self.subFrequency:
                 self.pollCounter = 0
-                self.deleteBlocks()
+                # self.deleteBlocks()  # skip block deletion until situation gets under control
                 self.subscribeDatasets()
         except HTTPException as ex:
             if hasattr(ex, "status") and ex.status in [502, 503]:


vytjan:
The block deletions were not properly working after this PR [1] went in place. But as Dima said, we have noticed that until recently.

As I understand, the problematic part is that originally, for T0 deletions, the XML was generated using the makePhEDExXMLForBlocks method [2].
Meantime, the [1] changes forced the XML being created with the makePhEDExXMLForDatasets method [3].

This caused the not working deletions for all this time and as well as the dataset level deletions after the patch was deployed earlier this week.

[1]
https://github.com/amaltaro/WMCore/commit/c3bb31fadcf122aed2af2337b632baa4b2cf4a67

 [2]
https://github.com/amaltaro/WMCore/blob/master/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L524

[3]
https://github.com/amaltaro/WMCore/commit/c3bb31fadcf122aed2af2337b632baa4b2cf4a67#diff-73e9dd1a91792c7023eeec392d2e28b3L522

amaltaro:
I see, it seems there is another bug from when PhEDEx and XMLDrop modules were separated.
Can you guys confirm then whether T0 ever creates a dataset level deletion request? If not, then the problem is obvious indeed. I'm working on a real fix ...

hufnagel:
Tier0 only makes block deletion requests. In fact, the underlying WMCore code is specifically written to only make block deletion requests:

https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L514

This whole functionality is nothing Tier0 specific anyways, it's just a different way to handle subscriptions in the WMAgent.

amaltaro:
@drkovalskyi @vytjan now that the T0 situation is under control, can I close this temporary fix?

drkovalskyi:
We are not using it in production yet (in one shape or another).

amaltaro:
Thanks, Dima. Closing it then

PR #8636, opened 2018-05-28
Add retry count to the job classad - wmagent 1.1.10 version

amaltaro:
1.1.10 wmagent version of #8621


amaltaro:
unfortunately we still have a bunch of old agents being drained

Issue #8493, opened 2018-02-28
JobSubmitter memory hungry


amaltaro:
As reported by Diego in this elog
https://cms-logbook.cern.ch/elog/GlideInWMS/6087

there is a wmagent process that is consuming around 18GB of RSS memory on vocms0253. Having a look at the agent, that's JobSubmitter:
```
cmst1@vocms0253:/data/srv/wmagent/current $ grep -r 1352320 install/wmagent/*/Daemon.xml
install/wmagent/JobSubmitter/Daemon.xml:	<ProcessID Value="1352320"/>
cmst1@vocms0253:/data/srv/wmagent/current $ 
cmst1@vocms0253:/data/srv/wmagent/current $ ps -p 1352320 -o pid,ppid,rss,vsize,pcpu,pmem,cmd -ww | grep 1352320
1352320       1 18152828 18588968 3.0 30.1 python /data/srv/wmagent/v1.1.10.patch4/sw/slc7_amd64_gcc630/cms/wmagent/1.1.10.patch4/bin/wmcoreD --start --config=/data/srv/wmagent/v1.1.10.patch4/config/wmagent/config.py
cmst1@vocms0253:/data/srv/wmagent/current $ 
```

This must be the reason why that process sometimes silently dies (or it gets somehow stuck, I don't remember right now) and causes it to be automatically restarted.

For the record, this is the load of that agent right now
```
2018-02-28 21:52:27,445:139740511799040:INFO:JobSubmitterPoller:Found 8795 new jobs to be submitted.

cmst1@vocms0253:/data/srv/wmagent/current $ condor_status --schedd vocms0253.cern.ch
Name              Machine           RunningJobs   IdleJobs   HeldJobs
vocms0253.cern.ch vocms0253.cern.ch       14482      50249          0
                      TotalRunningJobs      TotalIdleJobs      TotalHeldJobs
               Total             14482              50249                  0
```

ticoann:
We can control the memory usage (in some degree) by setting below value lower.
config.JobSubmitter.cacheRefreshSize = 30000
Currently it gets holds 30000 record in the cache, which shouldn't be 18GB. I will restart JobSubmitter whether memory usage goes up again. We might have some memory leak.

ticoann:
After JobAccountant restarted memory usage is around 400MB. So it doesn't seem to be due a cache size. (cache was reduced only 33%). I will monitor for a few more days.

amaltaro:
[...]

PR #8651, opened 2018-06-13
Reduce the BossAir memory footprint

amaltaro:
While investigating #8493 - which I can't reproduce/pinpoint the memory leak via unittests - I saw large spikes (and sort of random) of memory consumption when a site status is changed and all `executing` jobs have to be loaded, the site white list parsed and jobs eventually updated.

Note for myself: when profiling memory usage via unit tests, the component class/algorithm is attached to the same thread running the unit test, that means objects in memory might be from both unit test and the component object instance (so not exactly what the component running in the agent would see...).

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -13,7 +13,6 @@ import logging
 import os.path
 import threading
 from collections import defaultdict, Counter
-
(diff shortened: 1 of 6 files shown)


amaltaro:
test this please

amaltaro:
I have done several tests with this change in and it looks Ok. Still, @ticoann Seangchan please review it  in the next week, especially because it touches BossAir :)

PR #8652, opened 2018-06-13
split STAGEOUT_ERRORS for code 60317. Fix #8405

### LINKED ISSUE #8405: when CRAB ASO timeouts should return exit code 60317
see
https://github.com/dmwm/CRABServer/issues/5364
current code in 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMExceptions.py#L263-L265
returns 60321 which indicates site problem, while ASO timeout can be for other reasons as well. Should use the ad-hoc, already existing 60317

belforte:
Fixes #8405
It turns out to be too much time expensive to set up a test of this. OTOH the only place where the affected lines are used is in CRABServer/ServerUtilities.py. So this is should be pretty safe  on your side.
A different story would be to move such CRAB specific code out or WMCore... too much work as well. I'll get back to it after ASOv2 is rolled out, hopefully we clean code and get rid of whatever makes use of this.


amaltaro:
Do you need it to be backported to the old crab branch (1.1.2)? Or only the new one that we might create tomorrow?

belforte:
new one only. thanks.

PR #8653, opened 2018-06-14
Added exit code 8033 - FileWriteError - to the list of exceptions

amaltaro:
Fixes #8647
Thanks, @belforte !

PR #8654, opened 2018-06-14
Bundle of WMExceptions patches for CRAB

amaltaro:
crab branch version of #8653 and #8405


amaltaro:
@belforte FYI, 1.1.14_crab was forked from 1.1.14 tag and 1.1.14.crab1 tag was created including this patch.

belforte:
thanks Alan, I will build with that and see how it goes

Issue #8648, opened 2018-06-11
vocms0256 has created > 3 million jobs (!)


amaltaro:
There is something weird going on with the production system.
According to grafana, today at 15:05 CERN time there was a spike of requests in ReqMgr2 (jumping from ~4k to ~17k requests in the system), where ~6k of them were in acquired status. (I bet it was a monit glitch, since it's back to normal after 15:15).

Then looking at the agents, it seems vocms0256 managed to acquire and create more than 3M jobs
```
2018-06-11 15:57:47,223:139947075262208:INFO:JobSubmitterPoller:Found 3375907 new jobs to be submitted.
```

having a quick look at wmstats, some of those workflows are
pdmvserv_task_SMP-RunIISummer15wmLHEGS-00224__v1_T_180416_194001_5265
pdmvserv_task_SMP-RunIISummer15wmLHEGS-00225__v1_T_180416_194002_549
pdmvserv_task_SMP-RunIISummer15wmLHEGS-00226__v1_T_180416_194018_271
pdmvserv_task_SMP-PhaseIISummer17GenOnly-00022__v1_T_180607_163200_4994

all MC from scratch AFAICT.

amaltaro:
I still don't know what exactly happened, but vocms0256 has acquired ~1400 WQE units since yesterday:
```
cmst1@vocms0256:/data/srv/wmagent/current $ grep pdmvserv_task_SMP-RunIISummer15wmLHEGS-00224__v1_T_180416_194001_5265 install/wmagent/WorkQueueManager/ComponentLog | grep 'Queuing element' | wc -l
1412
cmst1@vocms0256:/data/srv/wmagent/current $ grep pdmvserv_task_SMP-RunIISummer15wmLHEGS-00224__v1_T_180416_194001_5265 install/wmagent/WorkQueueManager/ComponentLog | grep 'Queuing element' | grep '2018-06-10' | wc -l
263
cmst1@vocms0256:/data/srv/wmagent/current $ grep pdmvserv_task_SMP-RunIISummer15wmLHEGS-00224__v1_T_180416_194001_5265 install/wmagent/WorkQueueManager/ComponentLog | grep 'Queuing element' | grep '2018-06-11' | wc -l
1149
```

I should have looked at the resource thresholds before moving the agent into drain mode.

@thongonary any chance we can abort/clone/reassign this *_5265? otherwise it will just slow this agent and production in general.

amaltaro:
[...]

PR #8655, opened 2018-06-14
TEMP patch to retrieve not more than 50k created jobs for submission

amaltaro:
Mitigates #8648 , wmagent branch
Number of rows is hardcoded to 50k, that's why it's temporary and I'm applying it to vocms0256 only.
Still, we can make the number of rows configurable and then push this in into master too.


amaltaro:
Tested in my testbed agent, so pushing it in in vocms0256

amaltaro:
Even though it's still applied to vocms0256, I think we can close it. The twiki has a pointer to it anyways.

PR #8658, opened 2018-06-16
Configure ErrorHandler to 50 jobs only; bump deploy example to 1.1.14…

amaltaro:
….patch2


amaltaro:
Tested! As mentioned in the src code, once the ErrorHandler issue is sorted out, we need to remove the config tweak from this deployment script.

PR #8659, opened 2018-06-19
Make WorkQueueManagerWMBSFileFeeder slightly more verbose

amaltaro:
I've seen heartbeat alarms for this thread (even in testbed), so we better improve its logging (could be better, but then that's going to be _maybe_ too much verbose).


amaltaro:
There is a tricky business with the logger object, so it gets recorded in the log with another thread name... I'll check if I can get it fixed later.

amaltaro:
@ticoann I also converted some "file XX on location" logs from info to debug, I think that's too verbose. Let me know if'd rather keep that as info for debugging purposes

ticoann (approved):

ticoann:
replaced by #8755

PR #8661, opened 2018-06-22
Limit number of jobs to cache in JobSubmitter; also order by task_id

### LINKED ISSUE #8660: Priority of subsequent jobs in task chains
how complicated would it be to have a mechanism that make a +=(N-1)*1000 priority on the job of task_N of a task chain, with respect to the workflow priority ?
trying to avoid having tasks competing for resource.
(on a suggestion from Tom)

amaltaro:
Fixes #8660

Summary of changes is:
* limit the number of jobs JobSubmitter can retrieve from the database and load in memory/cache
* make this new attribute configurable
* update job type priorities
* use task priority (0 to 5) from wmbs instead of every site in resource control
* make sure DAO always returns the highest priority work first, according to 
 1) task priority; then see https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMBS/CreateWMBSBase.py#L503
 2) workflow priority; then
 3) task id (later tasks go in first)

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -198,6 +198,7 @@ config.JobSubmitter.pollInterval = 120
 config.JobSubmitter.workerThreads = 1
 config.JobSubmitter.jobsPerWorker = 100
 config.JobSubmitter.maxJobsPerPoll = 1000
+config.JobSubmitter.maxJobsToCache = 50000
 config.JobSubmitter.cacheRefreshSize = 30000  # set -1 if cache need to refresh all the time.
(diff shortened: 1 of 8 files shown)


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
@ticoann please review.  I ran the unit tests both in mysql and oracle and everything looks fine. I also run tests on my oracle agent and I see the order of priorities being enforced.

I wasn't going to backport it, but since we hit that issue on vocms0250 (new agent with 1M jobs), I'm doing so and patching that for now.

Just a note, the database changes (wmbs_sub_types) are only applied during deployment, so we either update those rows by hand or just leave it as it is. I'm just leaving that alone for now (in vocms0250).

PR #8663, opened 2018-06-26
Minor improvements for ListThresholdsForCreate

### LINKED ISSUE #8648: vocms0256 has created > 3 million jobs (!)

amaltaro:
Fixes #8648 (eventually)

I just wanted to see whether I really understood this DAO and to make sure my changes don't break anything.
Next step is to look at Acquired LQE (local workqueue db), which I think we can actually look at wmbs subscriptions somehow.

### DIFF
--- a/src/python/WMCore/ResourceControl/MySQL/ListThresholdsForCreate.py
+++ b/src/python/WMCore/ResourceControl/MySQL/ListThresholdsForCreate.py
@@ -71,7 +71,11 @@ class ListThresholdsForCreate(DBFormatter):
                             WHERE wmbs_job.location IS NULL AND
                                   wmbs_job_state.name != 'killed' AND
                                   wmbs_job_state.name != 'cleanout' AND
-                                  wsv.valid = 1) unassigned_jobs ON
+                                  (wsv.valid = 1 OR
+                                   (wsv.valid IS NULL AND NOT EXISTS
+                                    (SELECT wsv2.valid FROM wmbs_subscription_validation wsv2
+                                     WHERE wsv2.subscription_id = wmbs_jobgroup.subscription
+                                     AND wsv2.valid = 1)))) unassigned_jobs ON
                             wl.id = unassigned_jobs.location
                          INNER JOIN wmbs_location_state wls ON
                             wls.id = wl.state
--- a/src/python/WMCore/ResourceControl/MySQL/ListThresholdsForCreate.py
+++ b/src/python/WMCore/ResourceControl/MySQL/ListThresholdsForCreate.py
@@ -3,7 +3,7 @@
 _ListThresholdsForCreate_
 
 Query the database to determine how many jobs are pending so that we can
-determine whether or not to task for more work.
+determine whether or not to ask for more work.
 """
 
 from WMCore.Database.DBFormatter import DBFormatter
@@ -71,11 +71,7 @@ class ListThresholdsForCreate(DBFormatter):
                             WHERE wmbs_job.location IS NULL AND
                                   wmbs_job_state.name != 'killed' AND
                                   wmbs_job_state.name != 'cleanout' AND
-                                  (wsv.valid = 1 OR
-                                   (wsv.valid IS NULL AND NOT EXISTS
-                                    (SELECT wsv2.valid FROM wmbs_subscription_validation wsv2
-                                     WHERE wsv2.subscription_id = wmbs_jobgroup.subscription
-                                     AND wsv2.valid = 1)))) unassigned_jobs ON
+                                  wsv.valid = 1) unassigned_jobs ON
                             wl.id = unassigned_jobs.location
                          INNER JOIN wmbs_location_state wls ON
                             wls.id = wl.state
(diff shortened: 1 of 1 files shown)


amaltaro:
The failing unit test is legit, even though I'm not sure the use case is valid...
The current behavior is that subscriptions without a site white/black list will use the job location assigned to count a pending_job, while the code with these changes will just disconsider any subscriptions without white/black locations (valid is Null).

@ticoann do you know whether we can have a wmbs_subscription where the corresponding wmbs_subscription_validation.valid is Null? I guess not, but...

ticoann:
Thanks Alan, that is smart. The one concern is when there is very slow job creation as you observed. Agent will stop pulling the work until every subscription is finished.  so it won't pull the enough work and work will go to other agents. (However it might be a right behavior if JobCreator has some problems).

amaltaro:
I've done the WorkQueueManager vs JobCreator tests and it works as expected. Not pulling any new work from GQ while there are still subscriptions not handled by JobCreator.

I'm going to try to better understand those queries now and check whether my SQL changes cause more benefit than drawbacks. If we keep that, I still have to update one unit test.

src/python/WMCore/ResourceControl/MySQL/ListThresholdsForCreate.py
```diff
@@ -71,11 +71,7 @@ class ListThresholdsForCreate(DBFormatter):
                             WHERE wmbs_job.location IS NULL AND
                                   wmbs_job_state.name != 'killed' AND
                                   wmbs_job_state.name != 'cleanout' AND
-                                  (wsv.valid = 1 OR
-                                   (wsv.valid IS NULL AND NOT EXISTS
```
amaltaro: I had a look at vocms0254 db and I cannot find anything for `wsv.valid IS NULL AND...`.
Still, running the original sub-query (unassigned_jobs) and this one, I get around 246k rows while the latter gives me 245k rows.

ticoann: you can find  wsv.valid is NULL from the table but only when you do the left outer join. Were you checking that? I am still trying to understand original query

amaltaro:
Seangchan, as we discussed today, I moved the 2nd commit to https://github.com/dmwm/WMCore/pull/8667 and I/we will work on it for the new agent only. I'll bug Kate after CHEP too.

amaltaro:
test this please

ticoann:
test this please

amaltaro:
The failing unit test was explained here:
https://github.com/dmwm/WMCore/pull/8663#issuecomment-400324402

I guess that's not a valid case though (having a subscription without any valid location), so I'll just update the unit test.

src/python/WMCore/ResourceControl/MySQL/ListThresholdsForCreate.py
amaltaro: I'm not sure I got the correct query. However, looking at the code, I see we do not have any update action for wmbs_subscription_validation. So either a subscription has a site white/black list or it doesnt:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WMBSHelper.py#L294

That means, removing this part of the code _should_ indeed be fine

amaltaro:
I was wrong, merge subscriptions don't have an entry in wmbs_sub_validation (so no sites white/black listed). See
```
MariaDB [wmagent]> select count(distinct wmbs_subscription.id) from wmbs_sub_types
    ->   inner join wmbs_subscription on wmbs_subscription.subtype = wmbs_sub_types.id
    -> where name = 'Merge' and wmbs_subscription.id not in 
    ->   (select distinct subscription_id from wmbs_subscription_validation);
+--------------------------------------+
| count(distinct wmbs_subscription.id) |
+--------------------------------------+
|                                  199 |
+--------------------------------------+
1 row in set (0.00 sec)

MariaDB [wmagent]> select count(distinct wmbs_subscription.id) from wmbs_sub_types
    ->   inner join wmbs_subscription on wmbs_subscription.subtype = wmbs_sub_types.id
    -> where name = 'Merge' and wmbs_subscription.id in 
    ->   (select distinct subscription_id from wmbs_subscription_validation);
+--------------------------------------+
| count(distinct wmbs_subscription.id) |
+--------------------------------------+
|                                    0 |
+--------------------------------------+
1 row in set (0.00 sec)
```

I did however a very very quick performance test for the remaining code changes (not the SQL) and it seems to be around 10% faster (and cleaner). So I'm going to roll back the SQL changes and we can go ahead with the formatting changes only. Does it sound good, Seangchan?

ticoann:
test this please

amaltaro:
I give up trying to understand what's wrong with this patch and the failing unit test. Even without a SQL change, it would give us a performance boost, but I'm closing this one.

PR #8668, opened 2018-07-02
fix the view by date

### LINKED ISSUE #8642: ReqMgr2 info API reqmgr_last_injected_request
This API is probably not used for anything but by the ServerMonitor.
Still, we might eventually want to fix this `reqmgr_last_injected_request`, which is clearly wrong.
And there might be other parameters not getting updated as well.

```
{"result": [
 {
  "wmcore_reqmgr_version": "1.1.14.pre4", 
  "reqmgr_db_info": {
    "update_seq": 141019, 
    "disk_size": 387244152, 
    "reqmgr_couch_url": "https://cmsweb-testbed.cern.ch/couchdb", 
    "purge_seq": 0, 
    "doc_count": 13877, 
    "compact_running": false, 
    "db_name": "reqmgr_workload_cache", 
    "doc_del_count": 14, 
    "instance_start_time": "1527764617008404", 
    "committed_update_seq": 141019, 
    "data_size": 384154354, 
    "disk_format_version": 6
  }, 
  "reqmgr_last_injected_request": {
    "rows": [
      {
        "value": {
          "RequestType": "ReReco", 
          "RequestStatus": null, 
          "RequestDate": [
            "2012", 
            "9", 
            "3", 
            "12", 
            "16", 
            "46"
          ], 
          "RequestName": "pkonst_WMSTATS_RERECOSKIMTEST_01_120913_144507_3291"
        }, 
        "id": "pkonst_WMSTATS_RERECOSKIMTEST_01_120913_144507_3291", 
        "key": [
          "2012", 
          "9", 
          "3", 
          "12"
        ]
      }
    ], 
    "total_rows": 13819, 
    "offset": 0
  }
}]}
```

ticoann:
fixes #8642. Still need to be tested.

### DIFF
--- a/src/couchapps/ReqMgr/views/bydate/map.js
+++ b/src/couchapps/ReqMgr/views/bydate/map.js
@@ -1,15 +1,14 @@
 // bydate view - returns all requests sorted by RequestDate
 // curl $COUCHURL/reqmgr_workload_cache/_design/ReqMgr/_view/bydate?descending=true&limit=1
 
-function(doc) 
-{	
-	if (doc.RequestDate && (doc.RequestDate.length > 0)){
-      var dateArray = doc.RequestDate;
-      // get year, month, day, min
-      emit([dateArray[0], dateArray[1], dateArray[2], dateArray[3]],
+function(doc) {
+    if (doc.RequestDate && (doc.RequestDate.length >= 4)){
+        var dateArray = doc.RequestDate;
+        // get year, month, day, hour
+        emit([parseInt(dateArray[0]), parseInt(dateArray[1]), parseInt(dateArray[2]), parseInt(dateArray[3])],
 		                   {"RequestDate": doc.RequestDate,
-		                   "RequestName": doc.RequestName,
-		                   "RequestType": doc.RequestType,
-		                   "RequestStatus": doc.RequestStatus});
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan please review

src/couchapps/ReqMgr/views/bydate/map.js
```diff
@@ -6,7 +6,7 @@ function(doc)
 	if (doc.RequestDate && (doc.RequestDate.length > 0)){
       var dateArray = doc.RequestDate;
       // get year, month, day, min
```
amaltaro: it's actually year, month, day and hour.
Can you also properly indent the if above (and these lines) such that it's more readable?

amaltaro (changes requested):
Since you didn't "officially" ask for my review, I did it myself :-)
Code looks good, I guess we just have a mix of RequestDate data format in old and new requests and that's why it was failing? I also remember Antanas reporting in the very far past that this field wasn't very consistent among all the requests, maybe those outliers are gone now.

Seangchan, a made a minor change request in the code though.

ticoann:
Thanks Alan, I will update the code as you suggested. Yes I am not sure why it is mixed with string.  We could fix all the database instead but ...

PR #8623, opened 2018-05-21
fix error handler memory problem

ticoann:
fixes #8614, replace #8616

PR #8672, opened 2018-07-09
fix memory problem on error handler

ticoann:
replace #8623 
Create only one instance of Run object for the duplicate file. It still creates duplicate files object (WMBSFile) but only one Run object is shared for multiple file object. It might be better to create only one file object as well. I think when the job mask runs I am not sure whether it is needed to have a separate file object. (It seems it doesn't need that though)

### DIFF
--- a/src/python/WMCore/WMBS/MySQL/Jobs/LoadForErrorHandler.py
+++ b/src/python/WMCore/WMBS/MySQL/Jobs/LoadForErrorHandler.py
@@ -108,6 +108,7 @@ class LoadForErrorHandler(DBFormatter):
                                            transaction=transaction)
         fileList = self.formatDict(filesResult)
 
+        noDuplicateFiles = {}
         fileBinds = []
         if fileSelection:
             fileList = [x for x in fileList if x['lfn'] in fileSelection[x['jobid']]]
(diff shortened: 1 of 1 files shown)


ticoann:
It seems #8623 actually works. However, this would me more elegant solution and in case of MCFake file this will give more performance advantage when maxProcessSize is larger with the same memory usage. (If there are so many MCFake failure) I will try to run the test for this patch.

ticoann:

amaltaro:
Patch looks good Seangchan. Unless there are several different workflows with tons of lumis, it should indeed save quite some memory usage.

ticoann:
Hope there is no case like that. :-)

ticoann:
test this please

ticoann:
Hi Alan I ran the test for this. Also tested in submit5 as well.  Just crash the component before it upload can compare the result with the patch and without patch. So I think it is ready to patch the agents. I will create the branch patch.

amaltaro:
Thanks, Seangchan. I'll push this patch in the agents right now then.

amaltaro (approved):

ticoann:
Thanks Alan, Didn't anything blow up?

PR #8676, opened 2018-07-10
set the dataset parentage for StepChain

### LINKED ISSUE #8671: Modify DBSBufferBlock structure and include parent dataset information
in relation to https://github.com/dmwm/DBS/issues/566.
Instead setting dataset parentage separately, pass the parent dataset information in DBSBuffer block.
1. need to check whether the block is generated by StepChain
2. make sure file parentage is not set for the StepChain. (Some StepChain will have the correct parentage when there is only one Step) @amaltaro, Alan can you think about other cases.

ticoann:
fixes #8671

Just initial version. 

1. getting the StepChain parentage from wmstats server. (Since update will occur only active data, this will be more performative than getting the information from WMBS)
2. Rely on the order of dataset reported from filtered query (which will order the output dataset with the same order as Step definition - Need to be sure this is the right parentage relation, if not, we need to have the way to get precise information.)
3. Maintain the cache for parent dataset for every cycle. - May need to keep more than one cycle.
4. Handles the case if only one step in StepChain  - parentage will be set correctly by file as other request type.
5. Pass the parentage information through insertBulkBlock call. (requires DBS api updated https://github.com/dmwm/DBS/issues/566)

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
@@ -49,6 +49,7 @@ class DBSBufferBlock:
                           'primds':               {},   # Dict of primary dataset info
                           'dataset':              {},   # Dict of processed dataset info
                           'file_parent_list':     [],   # List of file parents
+                          'dataset_parent_list':  [],   # List of parent datasets (DBS requires this as list although it only allows one parent)
                           'close_settings':       {}}   # Dict of info about block close settings
 
         self.files        = []
@@ -151,10 +152,13 @@ class DBSBufferBlock:
         # Append to the files list
         self.data['files'].append(fileDict)
 
-        # now add file to data
-        parentLFNs = dbsFile.getParentLFNs()
-        for lfn in parentLFNs:
-            self.addFileParent(child = dbsFile['lfn'], parent = lfn)
+        # If dataset_parent_list is defined don't add the file parentage.
+        # This means it is block from StepChain workflow and parentage of file will be resloved later
+        if not self.data['dataset_parent_list']:
+            # now add file to data
+            parentLFNs = dbsFile.getParentLFNs()
+            for lfn in parentLFNs:
+                self.addFileParent(child = dbsFile['lfn'], parent = lfn)
 
 
         # Do the algo
@@ -217,8 +221,7 @@ class DBSBufferBlock:
 
         Add the parent datasets to the data block
         """
-
-        self.data['ds_parent_list'].append({'parent_dataset': parent})
+        self.data['dataset_parent_list'].append(parent)
         return
 
     def setProcessingVer(self, procVer):
--- a/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
@@ -40,6 +40,7 @@ from WMComponent.DBS3Buffer.DBSBufferUtil import DBSBufferUtil
 from WMCore.Algorithms.MiscAlgos import sortListByKey
 from WMCore.DAOFactory import DAOFactory
 from WMCore.Services.UUIDLib import makeUUID
+from WMCore.Services.WMStatsServer.WMStatsServer import WMStatsServer
 from WMCore.WMException import WMException
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 
@@ -126,6 +127,8 @@ def uploadWorker(workInput, results, dbsUrl):
                 logging.error(msg)
                 logging.error(str(traceback.format_exc()))
                 results.put({'name': name, 'success': "check"})
+            elif 'Missing data when inserting to dataset_parents' in exString:
+                logging.warning("Parent dataset is not inserted yet for block %s", block)
             else:
                 msg = "Error trying to process block %s through DBS.\n" % name
                 msg += exString
@@ -162,6 +165,13 @@ class DBSUploadPoller(BaseWorkerThread):
         # This is slightly dangerous, but DBSUpload depends
         # on DBSInterface anyway
         self.dbsUrl = self.config.DBS3Upload.dbsUrl
+
+        #Tier0 Agent don't need this
+        if hasattr(self.config, "Tier0Feeder"):
+            self.wmstatsServerSvc = None
+        else:
+            wmstatsSvcURL = (self.config.AgentStatusWatcher.centralWMStatsURL).replace("couchdb/wmstats", "wmstatsserver")
+            self.wmstatsServerSvc = WMStatsServer(wmstatsSvcURL)
 
         self.dbsUtil = DBSBufferUtil()
 
@@ -200,6 +210,8 @@ class DBSUploadPoller(BaseWorkerThread):
                                      'dbsuploader_block.json')
 
         self.timeoutWaiver = 1
+
+        self.datasetParentageCache = {}
 
         return
 
@@ -293,6 +305,9 @@ class DBSUploadPoller(BaseWorkerThread):
         """
         try:
             logging.info("Starting the DBSUpload Polling Cycle")
+            # refreshing parentageCache every cycle
+            self.updateDatasetParentageCache()
+            logging.info("Dataset parentage map: %s" % self.datasetParentageCache)
             self.checkBlocks()
             self.loadBlocks()
             self.loadFiles()
@@ -307,6 +322,13 @@ class DBSUploadPoller(BaseWorkerThread):
             msg += str(str(traceback.format_exc()))
             logging.error(msg)
             raise DBSUploadException(msg)
+
+    def updateDatasetParentageCache(self):
+        if self.wmstatsServerSvc:
+            self.datasetParentageCache = self.wmstatsServerSvc.getChildParentDatasetMap()
+        else:
+            self.datasetParentageCache = {}
+        return
 
     def loadBlocks(self):
         """
@@ -341,6 +363,11 @@ class DBSUploadPoller(BaseWorkerThread):
             block = DBSBufferBlock(name=blockInfo['block_name'],
                                    location=blockInfo['origin_site_name'],
                                    datasetpath=blockInfo['datasetpath'])
+
+            parent = self.datasetParentageCache.get(blockInfo['datasetpath'])
+            if parent:
+                block.addDatasetParent(parent)
+                logging.info("load block: Child dataset %s, Parent dataset %s", blockInfo['datasetpath'], parent)
             block.FillFromDBSBuffer(blockInfo)
             blockname = block.getName()
 
@@ -504,6 +531,12 @@ class DBSUploadPoller(BaseWorkerThread):
         newBlock = DBSBufferBlock(name=blockname,
                                   location=location,
                                   datasetpath=datasetpath)
+
+        parent = self.datasetParentageCache.get(datasetpath)
+        if parent:
+            newBlock.addDatasetParent(parent)
+            logging.info("Get block: Child dataset %s, Parent dataset %s", datasetpath, parent)
+
         self.blockCache[blockname] = newBlock
         return newBlock
 
@@ -816,3 +849,5 @@ class DBSUploadPoller(BaseWorkerThread):
 
         # We're done
         return
+
+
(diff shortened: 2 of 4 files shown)


ticoann:
@amaltaro, Alan please review, Not the code yet (the code is not ready yet), but in terms of the logic for how to get the StepChain  dataset parentage.

ticoann:
@vlimant, Thanks Jean-Roch, that was the information I needed. I will update the patch.

amaltaro:
Seangchan, Jean-Roch is right, Step1 isn't necessarily the parent of Step2 output.
However, what you really need to figure out the StepChain parentage are these 3 keys:
https://cmsweb-testbed.cern.ch/wmstatsserver/data/filtered_requests?RequestName=amaltaro_StepChain_Tasks_HG1805_Validation_180426_130526_5459&mask=StepName&mask=KeepOutput&mask=InputStep

thus `StepName`, `InputStep` and `KeepOutput`.

Now I have some comments/questions on your logic:
> getting the StepChain parentage from wmstats server. (Since update will occur only active data, this will be more performative than getting the information from WMBS)

can you show me the query/API you're planning to use from WMStats? Unless we cache some info in the component, this might increase a lot the load on WMStats, given it would have a query per job.

2. has been discussed above.

> 3. Maintain the cache for parent dataset for every cycle. - May need to keep more than one cycle.

I think you just answered part of my question on 1. Still, I'd say this cache can be a long lasting cache (at least a day). The only changes are new requests being added to the cache and old ones leaving it.

ticoann:
Thanks you very much Alan for the information.

About the cache, I think I can increase the duration for long time, in terms of the load, it shouldn't be problem. In the serverside, wmstats maintain the cache for 10 min, and filterted_requests only run against that cache so it would be small and fast. So it is not necessary to do FIFO queue. I was more concerning it will fail when service is not available.

About the parent dataset logic.
1.  check the step > 2,  check KeepOut value and the InputStep, If KeepOut is false move to next step.
2.  If child step has KeepOut true, check parent step with InputStep, whether if parent step has KeepOut = true. (do recursively until we find the KeepOut step is true.
3.  In that case, set the output dataset parentage between child step (find the output datasets using ProcessString and StepName and parent step.

However, this seems overly complicated. Is there any easier way to figure this out?
 Whichever logic we use, I think it should be automatically generated the  parentage relation and add to spec (maybe in assigned statge) instead of calculated every time we need.

ticoann:
Just for the reminder. - I need to catch below error and retry. (This happens when parent dataset is not yet inserted but child dataset try to add parentage.

```
2018-08-01 22:08:02,394:140466190964544:ERROR:DBSUploadPoller:Error trying to process block /RelValProdMinBias/Integ_Test-ProdMinBias_TaskChain_ProdMinBias_TaskChain-v11/GEN-SIM#d9e10c13-d130-448f-ad60-f8460f7bfa0f through DBS.
HTTP Error 412:  Missing data when inserting to dataset_parents. 
2018-08-01 22:08:02,397:140466190964544:ERROR:DBSUploadPoller:Traceback (most recent call last):
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/wmagent/1.1.14.patch6/lib/python2.7/site-packages/WMComponent/DBS3Buffer/DBSUploadPoller.py", line 105, in uploadWorker
    dbsApi.insertBulkBlock(blockDump=block)
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/dbs3-client/3.7.3/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 337, in insertBulkBlock
    result =  self.__callServer("bulkblocks", data=blockDump, callmethod='POST' )
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/dbs3-client/3.7.3/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 194, in __callServer
    self.__parseForException(http_error)
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/dbs3-client/3.7.3/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 221, in __parseForException
    raise HTTPError(http_error.url, data['exception'], data['message'], http_error.header, http_error.body)
HTTPError: HTTP Error 412:  Missing data when inserting to dataset_parents
```

ticoann:
Hmm, here is more serious error which shouldn't happen. (This needs be investigated)
```
2018-08-02 21:56:08,185:140043492792064:ERROR:DBSUploadPoller:Error trying to process block /WJetsToLNu_HT-200To400_TuneCUETP8M1_13TeV-madgraphMLM-pythia8/RunIIFall17NanoAOD-RunIIFall17NanoAOD_TaskChain_NanoAOD_TaskChain-v11/NANOAODSIM#bf41aa33-6abd-4d27-b816-ba4893e20843 through DBS.
HTTP Error 400: insertBulkBlock: dataset and file parentages cannot be in the input at the same time
2018-08-02 21:56:08,187:140043492792064:ERROR:DBSUploadPoller:Traceback (most recent call last):
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/wmagent/1.1.14.patch6/lib/python2.7/site-packages/WMComponent/DBS3Buffer/DBSUploadPoller.py", line 105, in uploadWorker
    dbsApi.insertBulkBlock(blockDump=block)
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/dbs3-client/3.7.3/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 337, in insertBulkBlock
    result =  self.__callServer("bulkblocks", data=blockDump, callmethod='POST' )
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/dbs3-client/3.7.3/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 194, in __callServer
    self.__parseForException(http_error)
  File "/data/srv/wmagent/v1.1.14.patch6/sw/slc7_amd64_gcc630/cms/dbs3-client/3.7.3/lib/python2.7/site-packages/dbs/apis/dbsClient.py", line 221, in __parseForException
    raise HTTPError(http_error.url, data['exception'], data['message'], http_error.header, http_error.body)
HTTPError: HTTP Error 400: insertBulkBlock: dataset and file parentages cannot be in the input at the same time

2018-08-02 21:56:10,131:140043492792064:ERROR:DBSUploadPoller:Error found in multiprocess during process of block /WJetsToLNu_HT-200To400_TuneCUETP8M1_13TeV-madgraphMLM-pythia8/RunIIFall17NanoAOD-RunIIFall17NanoAOD_TaskChain_NanoAOD_TaskChain-v11/NANOAODSIM#bf41aa33-6abd-4d27-b816-ba4893e20843
2018-08-02 21:56:10,131:140043492792064:ERROR:DBSUploadPoller:Error trying to process block /WJetsToLNu_HT-200To400_TuneCUETP8M1_13TeV-madgraphMLM-pythia8/RunIIFall17NanoAOD-RunIIFall17NanoAOD_TaskChain_NanoAOD_TaskChain-v11/NANOAODSIM#bf41aa33-6abd-4d27-b816-ba4893e20843 through DBS.
HTTP Error 400: insertBulkBlock: dataset and file parentages cannot be in the input at the same time
2018-08-02 21:56:10,235:140043492792064:INFO:BaseWorkerThread:DBSUploadPoller took 10.105 secs to execute

```

ticoann:
test this please

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -163,6 +166,13 @@ def __init__(self, config):
         # on DBSInterface anyway
         self.dbsUrl = self.config.DBS3Upload.dbsUrl
 
+        #Tier0 Agent don't need this
+        if hasattr(self.config, "Tier0Feeder"):
+            self.wmstatsServerSvc = None
+        else:
+            wmstatsSvcURL = (self.config.AgentStatusWatcher.centralWMStatsURL).replace("couchdb/wmstats", "wmstatsserver")
```
amaltaro: Seangchan, it looks like this `centralWMStatsURL` attribute is defined for a few components. How about we set it under the General section just to avoid these cross-component attributes? That can be done in a different PR though.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -504,6 +531,12 @@ def getBlock(self, newFile, location, skipOpenCheck=False):
         newBlock = DBSBufferBlock(name=blockname,
                                   location=location,
                                   datasetpath=datasetpath)
+
+        parent = self.datasetParentageCache.get(datasetpath)
+        if parent:
+            newBlock.addDatasetParent(parent)
+            logging.info("Get block: Child dataset %s, Parent dataset %s", blockInfo['datasetpath'], parent)
```
amaltaro: Just to make it consistent, how about using `datasetpath` instead of `blockInfo['datasetpath']`?

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, name, location, datasetpath):
                           'primds':               {},   # Dict of primary dataset info
                           'dataset':              {},   # Dict of processed dataset info
                           'file_parent_list':     [],   # List of file parents
+                          'dataset_parent_list':  [],   # List of parent datasets
```
amaltaro: I believe I mentioned it before, but why parent dataset is a list instead of a simple string?

amaltaro (commented):
I made some comments to the code, it might be that the code is ready though.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
ticoann: As we discussed we will do it in different PR. Since this has to be patched to current production agent.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
ticoann: As we discussed I will just update the comments that dbs requires this as list.

ticoann:
@amaltaro, Alan I updated the patch. This is tested in my vm, please review.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -293,6 +305,9 @@ def algorithm(self, parameters=None):
         """
         try:
             logging.info("Starting the DBSUpload Polling Cycle")
+            # refreshing parentageCache every cycle
+            self.updateDatasetParentageCache()
+            logging.info("Dataset parentage map: %s" % self.datasetParentageCache)
```
amaltaro: This log line can be very very large. Maybe we should convert it to debug or maybe only print the number of parent/child dataset maps.

amaltaro (commented):
Seangchan, I didn't really test it, but as you did, I think we can put it in the agent and then run further tests. There is just one open comment in the code, which I think we should at least change the logging level.

Besides that, this PR should be good to go

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
ticoann: It was just meant for testing. I will make to debug for now.  The number of parent/child give much information

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -341,6 +363,11 @@ def loadBlocks(self):
             block = DBSBufferBlock(name=blockInfo['block_name'],
                                    location=blockInfo['origin_site_name'],
                                    datasetpath=blockInfo['datasetpath'])
+
+            parent = self.datasetParentageCache.get(blockInfo['datasetpath'])
+            if parent:
+                block.addDatasetParent(parent)
+                logging.info("load block: Child dataset %s, Parent dataset %s", blockInfo['datasetpath'], parent)
```
ticoann: I will make this debug as well

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -504,6 +531,12 @@ def getBlock(self, newFile, location, skipOpenCheck=False):
         newBlock = DBSBufferBlock(name=blockname,
                                   location=location,
                                   datasetpath=datasetpath)
+
+        parent = self.datasetParentageCache.get(datasetpath)
+        if parent:
+            newBlock.addDatasetParent(parent)
+            logging.info("Get block: Child dataset %s, Parent dataset %s", datasetpath, parent)
```
ticoann: This one too.

PR #8679, opened 2018-07-11
add priority transition record

### LINKED ISSUE #8677: Record transitions in priority
recording the timestamps and DN of changes in priority would help trace what happened to various requests

ticoann:
fixes #8677 Still need to get the more requirement and test.

### DIFF
--- a/src/couchapps/ReqMgr/updates/updaterequest.js
+++ b/src/couchapps/ReqMgr/updates/updaterequest.js
@@ -5,16 +5,27 @@ function(doc, req) {
     	return [null, "Error: document not found"];
     };
     
-    function updateTransition() {
+    function updateTransition(key) {
+        var keyAllowed = {"RequestStatus": "RequestTransition",
+                          "RequestPriority": "PriorityTransition"};
+        var keyMap = {"RequestStatus": "Status",
+                      "RequestPriority": "Priority"};
+
+        var transitionKey = keyAllowed[key]
+        if (transitionKey === undefined) {
+            return
+        }
+
         var currentTS =  Math.round((new Date()).getTime() / 1000);
         var dn = doc.DN || null;
-        var statusObj = {"Status": doc.RequestStatus, "UpdateTime": currentTS, "DN": dn};
-        
-        if (!doc.RequestTransition) {
-            doc.RequestTransition = new Array();
-            doc.RequestTransition.push(statusObj);
+        var statusObj = {"UpdateTime": currentTS, "DN": dn};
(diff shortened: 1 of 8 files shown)


ticoann:
@amaltaro, Alan, please review. Still need to display information and test.

src/couchapps/ReqMgr/updates/updaterequest.js
```diff
@@ -78,9 +89,7 @@ function(doc, req) {
         doc[key] = newValues[key];
 
         // If key is RequestStatus, also update the transition
-        if (key == "RequestStatus") {
-            updateTransition();
-        }
+        updateTransition(key);
```
amaltaro: Directly calling it is certainly cleaner, but there must be a performance penalty for executing L9-13 for every single key. How about we at least define keyAllowed and keyMap outside that function?

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
```diff
@@ -232,6 +232,28 @@ def state_transition(docs):
     out += '</ul></fieldset>'
     return out
 
+def priority_transition(docs):
+    "create html for priority transition format"
+    if  not isinstance(docs, list):
```
amaltaro: extra space here. Maybe just run the "reformat code" in this module as another commit.

amaltaro (approved):
Code looks good, I just expressed one opinion about the javascript performance though.

Seangchan, I was going to say we need to update it here as well:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/RequestDB/RequestDBWriter.py#L10

but grepping around WMCore, we don't seem to be using it at all. How about you delete it together?

src/couchapps/ReqMgr/updates/updaterequest.js
ticoann: I think L9-13 has to run every single key whether it is inside the function or not. The question is how much does it cost to call the function each time. It was true that the cost of calling function is adding additional baggage before. I think these days Javascript engine is much more efficient.  (Not sure with other language, though)
https://www.codereadability.com/performance-cost-javascript-function-call-and-foreach/

But if you think putting the condition check outside is cleaner, we can move that.

src/python/WMCore/ReqMgr/Web/ReqMgrService.py
ticoann: thanks I will fix that.

ticoann:
Alan, Thank you very much for the review. I will run the test and let you know.

PR #8682, opened 2018-07-16
Allow CRIC to be used by the agent via an environment variable

### LINKED ISSUE #8696: Create a CRIC/SiteDB switch for WMAgent
Create a new configuration parameter for WMAgent and Global Workqueue defining whether we want to use CRIC or SiteDB for site-names and data-processing APIs.
Also add a new configuration parameters for the CRIC URL.

amaltaro:
Fixes #8696
CRIC usage is disabled, by default.

Allow CRIC to be used by the agent by setting an environment variable. This
`export WMAGENT_USE_CRIC=true`
has to be defined in /data/admin/wmagent/env.sh

If one wants to populate the resource-control based on CRIC queries, then the `-X` deploy parameter has to be used.

### DIFF
--- a/bin/wmagent-resource-control
+++ b/bin/wmagent-resource-control
@@ -14,6 +14,7 @@ from WMCore.Configuration import loadConfigurationFile
 from WMCore.ResourceControl.ResourceControl import ResourceControl
 from WMCore.Services.SiteDB.SiteDB import SiteDBJSON
 from WMCore.Services.SiteDB.SiteDBAPI import SiteDBAPI
+from WMCore.Services.CRIC.CRIC import CRIC
 from WMCore.WMInit import connectToDB
 from WMQuality.Emulators.SiteDBClient.MockSiteDBApi import mockGetJSON
 
@@ -100,16 +101,21 @@ def createOptionParser():
     return myOptParser
 
 
-def getCMSSiteInfo(pattern=''):
+def getCMSSiteInfo(pattern):
     """
     _getCMSSiteInfo_
 
     Query SiteDB for the site and SE names matching the pattern.  Return a
     dictionary keyed by site name.
     """
-    print("Querying SiteDB for site information...")
+    if os.getenv("WMAGENT_USE_CRIC", False):
+        cricObj = CRIC()
+    else:
+        cricObj = SiteDBJSON()  # this will be removed anyways
 
-    return SiteDBJSON().PSNtoPNNMap(pattern)
+    mapping = cricObj.PSNtoPNNMap(pattern)
+    print("Retrieved %i maps from %s" % (len(mapping), cricObj['endpoint']))
+    return mapping
 
 
 def addSites(resourceControl, allSites, ceName, plugin, pendingSlots, runningSlots,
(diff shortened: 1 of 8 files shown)


amaltaro:
There are still a few more places where SiteDB is used, but that is under ReqMgr2 service.
@ticoann this is going to be a PITA to review, but can you please have a quick look just to see whether I'm missing any configuration place? It's completely untested at the moment :)

src/python/WMCore/Services/SiteDB/SiteDBAPI.py
```diff
@@ -32,10 +32,12 @@ class SiteDBAPI(Service):
 
     def __init__(self, config={}, logger=None):
         config = dict(config)
-        config['endpoint'] = "https://cmsweb.cern.ch/sitedb/data/prod/"
-        config['logger'] = logger if logger else logging.getLogger()
+        config.setdefault('endpoint', "https://cmsweb.cern.ch/sitedb/data/prod/")
+        config.setdefault('logger', logging.getLogger())
 
         Service.__init__(self, config)
+        logging.info("AMR SiteDB endpoint: %s", config['endpoint'])
```
amaltaro: Remove this!

amaltaro:
@ticoann this patch seems to be working now and I already see WorkQueueManager acquiring work and contacting CRIC instead of Sitedb. 
I made some notes on the initial description, such that I don't forget other patches that need to come along when backporting it and patching the relval agent.
Can you please review it? Thanks

PS.: I'm not sure whether we remove those info("DEBUG...") messages from the CRIC module or not.

bin/wmagent-resource-control
```diff
@@ -96,20 +97,22 @@ def createOptionParser():
...
     myOptParser.add_option("--add-disk", dest="addDisk", default=True, action="store_false",
                            help="Use it to explicitly add the Disk endpoints together with the other T1s")
+    myOptParser.add_option("--use-Cric", dest="useCric", default=False, action="store_true",
+                           help="Query CRIC Service instead of SiteDB")
 
     return myOptParser
 
 
-def getCMSSiteInfo(pattern=''):
+def getCMSSiteInfo(pattern='', cricObj=None):
```
ticoann: This is just my preference, you don't have to change the code. 😄 
Since cricObj cannot be None. function signature could be below.
```
def getCMSSiteInfo(cricObj, pattern='')
```

bin/wmagent-resource-control
```diff
@@ -142,16 +145,14 @@ def addSites(resourceControl, allSites, ceName, plugin, pendingSlots, runningSlo
 
     return
 
-def addPNNs(resourceControl, pattern):
+def addPNNs(resourceControl, pattern, cricObj=None):
```
ticoann: same comment as above

bin/wmagent-resource-control
amaltaro: I agree! I'll update it

ticoann:
Hi Alan, patch looks good to me. Only one thing I am not sure mentioned above.
```
sites = cmsSiteNames()  - this will still use siteDB.
```

Also, I think you should leave the debug statement for now. We can clean up all together later.

Thanks Alan,

amaltaro:
Thank you very much for the review, Seangchan.
I ended up following your suggestion from yesterday and defined an environment variable for all that.
Just to make sure we don't use SiteDB when we don't want to, I've also set a wrong URL for SiteDB, such that it fails any sitedb queries.

amaltaro:
test this please

amaltaro:
@ticoann this has been already tested and the code is actually being used by the relval agent. Please review and if everything is fine, we can merge it in the next Tuesday.

amaltaro:
And new issue created https://github.com/dmwm/WMCore/issues/8845

PR #8685, opened 2018-07-16
clean up TaskArchiver log

### LINKED ISSUE #8684: Clean TaskArchiver log
While investigating Tier0 TaskArchiver problem, It seem that we are logging too many details in the log and actually slows down the process. Need to clean up the log.

ticoann:
fixes #8684

### DIFF
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -161,8 +161,9 @@ class CleanCouchPoller(BaseWorkerThread):
         try:
             logging.info("Creating workload summary")
             finishedwfsWithLogCollectAndCleanUp = DataCache.getFinishedWorkflows()
+            logging.info("Total %s workload summary will be created", len(finishedwfsWithLogCollectAndCleanUp))
             self.archiveSummaryAndPublishToDashBoard(finishedwfsWithLogCollectAndCleanUp)
-            logging.info("%s workload summary is created", len(finishedwfsWithLogCollectAndCleanUp))
+            logging.info("All workload summary docs were uploaded")
 
             logging.info("Cleaning up couch db")
             self.cleanCouchDBAndChangeToArchiveStatus()
@@ -373,7 +374,7 @@ class CleanCouchPoller(BaseWorkerThread):
         deletablewfs = deletableWorkflowsDAO.execute()
 
         # Only delete those where the upload and notification succeeded
-        logging.info("Found %d candidate workflows for deletion: %s", len(deletablewfs), deletablewfs.keys())
+        logging.info("Found %d candidate workflows for deletion.", len(deletablewfs))
         # update the completed flag in dbsbuffer_workflow table so blocks can be closed
         # create updateDBSBufferWorkflowComplete DAO
         if len(deletablewfs) == 0:
@@ -390,7 +391,7 @@ class CleanCouchPoller(BaseWorkerThread):
                 if wfStatus in safeStatesToDelete:
                     wfsToDelete[workflow] = {"spec": spec, "workflows": deletablewfs[workflow]["workflows"]}
                 else:
-                    logging.info("%s is in %s, will be deleted later", workflow, wfStatus)
+                    logging.debug("%s is in %s, will be deleted later", workflow, wfStatus)
 
             except Exception as ex:
                 # Something didn't go well on couch, abort!!!
@@ -410,6 +411,7 @@ class CleanCouchPoller(BaseWorkerThread):
         The input is a dictionary with workflow names as keys, fully loaded WMWorkloads and
         subscriptions lists as values
         """
+        logging.info("Deleting %s workflows by subscription (from disk)", len(workflows))
         for workflow in workflows:
             logging.info("Deleting workflow %s", workflow)
             try:
@@ -447,7 +449,7 @@ class CleanCouchPoller(BaseWorkerThread):
                     # Now delete directories
                     _, taskDir = getMasterName(startDir=self.jobCacheDir,
                                                workflow=wmbsWorkflow)
-                    logging.info("About to delete work directory %s", taskDir)
+                    logging.debug("About to delete work directory %s", taskDir)
                     if os.path.exists(taskDir):
                         if os.path.isdir(taskDir):
                             shutil.rmtree(taskDir)
@@ -500,7 +502,7 @@ class CleanCouchPoller(BaseWorkerThread):
         # so we can skip this if there is a summary already up there
         # TODO: With multiple agents sharing workflows, we will need to differentiate and combine summaries for a request
         if self.workdatabase.documentExists(workflowName):
-            logging.info("Couch summary for %s already exists, proceeding only with cleanup", workflowName)
+            logging.debug("Workload summary for %s already exists, proceeding only with cleanup", workflowName)
             return
 
         # Set campaign
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -207,7 +207,7 @@ class CleanCouchPoller(BaseWorkerThread):
             if spec:
                 self.archiveWorkflowSummary(spec=spec)
                 # Send Reconstruciton performance information to DashBoard
-                if self.dashBoardUrl != None:
+                if self.dashBoardUrl is not None:
                     self.publishRecoPerfToDashBoard(spec)
             else:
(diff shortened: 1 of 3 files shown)


src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -161,8 +161,9 @@ def algorithm(self, parameters=None):
         try:
             logging.info("Creating workload summary")
             finishedwfsWithLogCollectAndCleanUp = DataCache.getFinishedWorkflows()
+            logging.info("Total %s workflow summay will be created", len(finishedwfsWithLogCollectAndCleanUp))
```
amaltaro: typo

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -161,8 +161,9 @@ def algorithm(self, parameters=None):
...
             self.archiveSummaryAndPublishToDashBoard(finishedwfsWithLogCollectAndCleanUp)
-            logging.info("%s workload summary is created", len(finishedwfsWithLogCollectAndCleanUp))
+            logging.info("Workload summary is created")
```
amaltaro: isn't it 'workflow summary'?

amaltaro (changes requested):
comments in the code

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
ticoann: I am not sure we can go with workflow summary. :-) I will updated the patch

ticoann: actually it seems we are using workload summary fot the name.

amaltaro: then workload summary it is ;)

ticoann:
@amaltaro, Alan I added 
``` python
```
MathAlgo, however, I didn't change / -> //. It seems original code was wrong. could you check as well?

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
```diff
@@ -233,7 +233,7 @@ def completeTasks(self, finishedwfs):
                     # Tier-0 case, the agent has to mark it completed
                     if not self.useReqMgrForCompletionCheck:
                         self.requestLocalCouchDB.updateRequestStatus(workflow, "completed")
-                        logging.info("status updated to completed %s", workflow)
+                        logging.debug("status updated to completed %s", workflow)
```
amaltaro: This transition happens only once per workflow, so I think it's always a good idea to keep it info.

src/python/WMCore/Algorithms/MathAlgos.py
```diff
@@ -69,11 +70,9 @@ def getAverageStdDev(numList):
         average = 0.0
     if math.isnan(stdDev) or math.isinf(average) or not decimal.Decimal(str(stdDev)).is_finite():
         stdDev = 0.0
-    if type(stdDev) != float and type(stdDev) != int:
+    if not isinstance(stdDev, float) and not isinstance(stdDev, int):
```
amaltaro: You can make it as `if not isinstance(stdDev, (int, float)):`

src/python/WMCore/Algorithms/MathAlgos.py
```diff
@@ -117,13 +115,12 @@ def createHistogram(numList, nBins, limit):
...
         return histogram
 
-
     histEvents.sort()
     upperBound = max(histEvents)
     lowerBound = min(histEvents)
     if lowerBound == upperBound:
         # This is a problem
-        logging.error("Only one value in the histogram!")
+        logging.debug("Only one value in the histogram!")
```
amaltaro: If it's really a problem as the comment says, then we should make it a warning at least :)

amaltaro (changes requested):
Seangchan, I made several comments along the code.

About your division question, now that you imported division
`/`: makes a float division
`//`: makes a floor division (can return either integer or float though).

So the code seems to be correct, but I think we can remove those `float` casting from a few places, unless we want to make sure to show `1.0` instead of `1` for instance.

src/python/WMCore/Algorithms/MathAlgos.py
ticoann: I am not sure whether this is the problem, next code basically handles this cases. So not sure whether this need to warned.

ticoann:
> About your division question, now that you imported division
> /: makes a float division
> //: makes a floor division (can return either integer or float though).

Yes, so ideally it need to be changed / -> //,  (if there is no float type case is applied) but I think original code doesn't really mean to do that. i.e. math.sqrt(Q / n)

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -410,8 +411,9 @@ def killWorkflows(self, workflows):
         The input is a dictionary with workflow names as keys, fully loaded WMWorkloads and
         subscriptions lists as values
         """
+        logging.debug("Deleting workflows %s by subscription (from disk)", len(workflows))
         for workflow in workflows:
-            logging.info("Deleting workflow %s", workflow)
+            logging.debug("Deleting workflow %s", workflow)
```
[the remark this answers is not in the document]
amaltaro: But that's one time in a workflow lifetime, isn't it? I agree on overwhelming the logs with sometimes useless information, but at the same time it's sometimes interesting to make sure that the component did take some action on something (or at least started...).

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
amaltaro: production agent doesn't actually reach this code, since it uses reqmgr for the status transition. Unless I missunderstood you?

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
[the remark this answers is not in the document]
ticoann: If you insist. :-) I will update the patch. Thanks.

src/python/WMComponent/TaskArchiver/TaskArchiverPoller.py
ticoann: Yes you are right. It is only for Tier0

amaltaro:
Exactly, when there is no float I think the data itself is float already, for those that I checked, the result is meant to be float.

ticoann:
@amaltaro, Alan, I hope I updated all the places you reviewed. Could you check again?

ticoann:
test this please

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -161,8 +161,9 @@ def algorithm(self, parameters=None):
         try:
             logging.info("Creating workload summary")
             finishedwfsWithLogCollectAndCleanUp = DataCache.getFinishedWorkflows()
+            logging.info("Total %s workload summary will be created", len(finishedwfsWithLogCollectAndCleanUp))
             self.archiveSummaryAndPublishToDashBoard(finishedwfsWithLogCollectAndCleanUp)
-            logging.info("%s workload summary is created", len(finishedwfsWithLogCollectAndCleanUp))
+            logging.info("All workload summary docs were uploaded")
```
amaltaro: uploaded or created? Don't we upload them somewhere down the chain?

amaltaro (approved):
Besides those 2 other comments (that I might be wrong), code looks good and is/will be ready to go. Thanks!

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
ticoann: Actually we are uploading the summary in workload_summary couch db here.

src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
```diff
@@ -500,16 +502,15 @@ def archiveWorkflowSummary(self, spec):
         # so we can skip this if there is a summary already up there
         # TODO: With multiple agents sharing workflows, we will need to differentiate and combine summaries for a request
         if self.workdatabase.documentExists(workflowName):
-            logging.info("Couch summary for %s already exists, proceeding only with cleanup", workflowName)
+            logging.debug("Couch summary for %s already exists, proceeding only with cleanup", workflowName)
```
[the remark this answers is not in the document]
ticoann: You are right. Updated.

ticoann:
Thank you, Alan for the review.

PR #8686, opened 2018-07-17
Wrapper for HTCondor job submission

### LINKED ISSUE #8493: JobSubmitter memory hungry

amaltaro:
Fixes #8493 (until the real issue gets fixed in HTCondor, developers are already aware)

As pointed out by @bbockelm and @belforte , CRAB has to do condor submissions with a different user proxy, so they implemented this context manager class which manages proxy/environment and encapsulates the final condor submission/command in a forked process. It turns out it properly manages the memory footprint, which remains stables between different bulk submissions (tested in an ad-hoc script, around 40MB for 1k jobs submitted).

@bbockelm @belforte I'd welcome your review too. I made only minor changes to the wrapper class such that it works for CRAB and WMAgent, hopefully :)

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -11,7 +11,7 @@ import os.path
 import re
 import threading
 import time
-
+import pickle
 import classad
 import htcondor
 
@@ -23,7 +23,7 @@ from WMCore.DAOFactory import DAOFactory
 from WMCore.FwkJobReport.Report import Report
 from WMCore.WMInit import getWMBASE
 from WMCore.Lexicon import getIterMatchObjectOnRegexp, WMEXCEPTION_REGEXP, CONDOR_LOG_FILTER_REGEXP
-
+from WMCore.Services.PyCondor.PyCondorUtils import AuthenticatedSubprocess
 
 class SimpleCondorPlugin(BasePlugin):
     """
@@ -128,8 +128,7 @@ class SimpleCondorPlugin(BasePlugin):
         """
         _submit_
 
-
-        Submit jobs for one subscription
+        Submits jobs to the condor queue
         """
         successfulJobs = []
         failedJobs = []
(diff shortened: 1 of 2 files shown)


amaltaro:
@ticoann I am still not certain where these utils should be placed. I started creating a Utils/SubprocessUtils.py module, but since it has htcondor specific code, I moved it under the Services package. Anyways, I'm open to suggestions to make it easier/cleaner.

amaltaro:
test this please

ticoann:
Thanks Alan for taking care of this. I think current location (under the service) of the file is fine. One thing we probably need to add is try except block on "import htcondor" since not all the application using WMCore/Services will have htcondor dependencies.

belforte:
isn't it a matter of those using WMCore/Service/pycondor ? Those would better get an exceptoin if htcondor can't be imported.

ticoann:
@belforte, I think you are right it might be better to let it fail.  The thing I was worried was if any already existing application using the WMCore.Services classes has import like below. (Although that should be changed anyway)
```
from WMCore.Services import * 
```
But I don't see any of those import in DBS or WmAgentScript. I think throwing exception might be better. Thanks,

amaltaro:
Ok, then I'm not changing anything for now.
BTW, I ran some tests and it seems to be working just fine, component doesn't crash, information in `bl_runjob` table makes sense and memory should be stable, I'm going to submit 20k jobs now just to make sure it remains under control.

@belforte in case you want to update CRAB to start importing/using it from WMCore, please keep in mind this part of the code needs to be tested.

bbockelm:
Excellent, thanks for tackling this @amaltaro !

PR #8687, opened 2018-07-17
StepChain parentage logic

### LINKED ISSUE #8681: Create a map for Step and OutputDataset including child dataset information
Create a map between step and output dataset after assigning the workflow.
1. when workflow is assigned, figure out  mapping between step and output dataset
2. Also parentage between output datasets,  (Using StepName, InputStep, KeepOut parameter)
3. Create the map  [{StepName:  "SomeStep", "OutputDatasets": [], "ParentDataset": parentDataset}, ... ] - this is just sample  data structure. It can be changed.
4. Update the map to spec in couchdb

This will be used by Agent #8676

ticoann:
fixes #8681

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
@@ -53,6 +53,7 @@ class StepChainWorkloadFactory(StdBase):
         # stepMapping is going to be used during assignment for properly mapping
         # the arguments to each step/cmsRun
         self.stepMapping = {}
+        self.stepDatasetParentage = {}
 
     def __call__(self, workloadName, arguments):
         """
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro Alan, this is just initial attempt. Could you check the code. Especially about how to get the output data names. I am getting it from outmodule. But I am pretty sure it is not the correct way. 
Thanks,

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
```diff
@@ -229,7 +230,21 @@ def setupNextSteps(self, task, origArgs):
             parentCmsswStepHelper.keepOutput(parentKeepOutput)
             childKeepOutput = strToBool(taskConf.get('KeepOutput', True))
             childCmsswStepHelper.keepOutput(childKeepOutput)
-            self.setupOutputModules(task, taskConf, currentCmsRun, childKeepOutput)
+            outputMods = self.setupOutputModules(task, taskConf, currentCmsRun, childKeepOutput)
+
+            # set the output dataset parentage
+            if parentKeepOutput and parentStepNumber == "Step1":
+                childStepName = origArgs[parentStepNumber]["StepName"]
```
amaltaro: This is actually parentStepName

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
```diff
@@ -229,7 +230,21 @@ def setupNextSteps(self, task, origArgs):
...
+                # TODO how to get the output dataset if first step has KeepOut == True
+                outputModule = outputMods[origArgs[parentStepNumber]['InputFromOutputModule']]
```
amaltaro: and here it will break if parentStepNumber is Step1, since there is no InputFromOutputModule attr.

amaltaro (commented):
Seangchan, I made a couple of comments in the code but I haven't checked all the changes.
Besides that, we need to create a mapping for the step1 as well. 

Another comment is that we need to update this mapping during workflow assignment, since some parameters can change between creation and assignment.

Let me play with it tomorrow, I think starting with a unit test is going to be easier to understand it :-)

ticoann:
replaced by #8697

PR #8690, opened 2018-07-17
don't use workqueue injection check with T0Agent

### LINKED ISSUE #8215: JobArchiver error message important ?
Testing the latest Tier0 release, is this error important ?

2017-10-02 17:22:27,330:139649875142400:ERROR:JobArchiverPoller:Could not load workQueue'ConfigSection' object has no attribute 'WorkDir'

ticoann:
fixes #8215

### DIFF
--- a/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
+++ b/src/python/WMComponent/JobArchiver/JobArchiverPoller.py
@@ -69,16 +69,16 @@ class JobArchiverPoller(BaseWorkerThread):
             logging.exception(msg)
             raise JobArchiverPollerException(msg)
 
+        self.tier0Mode = hasattr(config, "Tier0Feeder")
+
         try:
-            self.workQueue = queueFromConfig(self.config)
+            if not self.tier0Mode:
+                self.workQueue = queueFromConfig(self.config)
         except Exception as ex:
             msg = "Could not load workQueue"
             msg += str(ex)
             logging.error(msg)
             # raise JobArchiverPollerException(msg)
-
-        self.handleWorkflowInjection = getattr(self.config.JobArchiver,
-                                               'handleInjected', True)
 
         return
 
@@ -295,7 +295,7 @@ class JobArchiverPoller(BaseWorkerThread):
         Mark any workflows that have been fully injected as injected
         """
 
-        if not self.handleWorkflowInjection:
+        if self.tier0Mode:
             logging.debug("Component will not check workflows for injection status")
             return
 


src/python/WMComponent/JobArchiver/JobArchiverPoller.py
```diff
@@ -69,8 +69,12 @@ def __init__(self, config):
             logging.exception(msg)
             raise JobArchiverPollerException(msg)
 
+        self.handleWorkflowInjection = getattr(self.config.JobArchiver,
```
amaltaro: Instead of defining another parameter, how about we use the same logic as this:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/AgentStatusWatcher/ResourceControlUpdater.py#L58
?

amaltaro (changes requested):
Seangchan, besides the comment I made in the code, we also need to update the `markInjected` method:
https://github.com/ticoann/WMCore/blob/bd5fa581944f61c2c7569937ccdf8b70b4115dfe/src/python/WMComponent/JobArchiver/JobArchiverPoller.py#L295

I guess we''l just assume T0 workflows as injected and mark them as so in oracle, basically skipping all that try/except block. (I wonder how T0 works now, this isn't a new change...)

hufnagel:
To point out, Tier0 *does* use the injection flag. It's critical to manage the handover from Repack to PromptReco. We set this ourselves internally from the Tier0. You will need to check for the injected status for Tier0 workflows.

amaltaro:
@hufnagel I didn't get this comment "We set this ourselves internally from the Tier0."
Are you saying a T0 code (not from WMCore) is responsible for marking workflows as injected? Such that that whole method becomes unneeded for the T0 agent?

hufnagel:
The relevant code for marking workflows as injected from the Tier0 is here

https://github.com/dmwm/T0/blob/master/src/python/T0/WMBS/Oracle/Tier0Feeder/MarkWorkflowsInjected.py

and that DAO is then used by

https://github.com/dmwm/T0/blob/master/src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py#L254

https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L1122

amaltaro:
Thanks, Dirk.
Seangchan, it's not crystal clear from Dirk's comment, but we discussed in the office and the summary is that we can check whether it's a T0 agent in the very beginning of the `markInjected` method, if so just return.

ticoann:
Thanks Alan, Dirk, I updated the patch. Please review.

amaltaro (approved):

Issue #8644, opened 2018-06-05
Log original request parameters before ReqMgr default values are added


amaltaro:
We should keep a record of every POST (and PUT) ReqMgr2 request before we apply the default parameters/values to the user request, in addition to what we have already.

ticoann:
@amaltaro, Alan where would you keep this and what will be the usage?

amaltaro:
Darn it! We should get this in for August deployment. I think we can simply log the user input in the usual reqmgr2 log.

The use case is debugging workflows and making sure that what the user claims to have posted to reqmgr2 is actually what he really posted. Example are the EventsPerJob and EventsPerLumi, that we're rounding now in ReqMgr2.

I'm working on it today.

amaltaro:
Fixed by #8691

PR #8691, opened 2018-07-19
Remove duplicate cherrypy logs in the central services

### LINKED ISSUE #8111: Duplicate reqmgr log records
Sorry, my bad that I haven't seen this one before... apparently this module
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/ReqMgr/ReqMgr.py

is writing its log records twice to the reqmgr log file, e.g.:
```
[30/Aug/2017:11:46:35]  INFO: Assign request amaltaro_ReDigi_cmsR ...
then
INFO:cherrypy.error:[30/Aug/2017:11:46:35]  INFO: Assign request amaltaro_ReDigi_cmsR...
```

which means - I think - that logger object has two handlers. I'm not sure I know how to fix it without breaking other stuff :-)

amaltaro:
Fixes #8111 and #8644

The actual change to stop the duplicate log records was just with the `log.screen=False` line, which disables writing log records to stdout. This way we keep only the cherrypy access and error logs (with a format that ain't great, but I couldn't fix it yet). Notice that everything that is not an HTTP request (access) is logged with the error log manager, see documentation in:
https://tools.ietf.org/doc/python-cherrypy3/api/cherrypy._cplogging-module.html

Second change was to make ReqMgr2 slightly more verbose, recording original user parameters, priority and status transitions.
Seangchan, as we discussed, logging for priority change of requests not assigned yet are ugly, since the web UI posts to ReqMgr server the whole dictionary built for the client's browser.

### DIFF
--- a/src/python/WMCore/REST/Main.py
+++ b/src/python/WMCore/REST/Main.py
@@ -8,18 +8,28 @@ up an appropriately configured CherryPy instance. Views are loaded dynamically
 and can be turned on/off via configuration file."""
 from __future__ import print_function
 
-import sys, os, errno, re, os.path, subprocess, socket, time
-import cherrypy, logging, thread, traceback
-import WMCore.REST.Tools
-from WMCore.Configuration import ConfigSection, loadConfigurationFile
+import errno
+import logging
+import os
+import os.path
+import re
+import signal
+import socket
+import sys
+import thread
+import time
+import traceback
 from cStringIO import StringIO
-from optparse import OptionParser
-from cherrypy import Application
-from cherrypy.lib import profiler
-from cherrypy._cplogging import LogManager
-from subprocess import Popen, PIPE
 from glob import glob
-from signal import *
+from optparse import OptionParser
+from subprocess import Popen, PIPE
+
+import cherrypy
+from cherrypy import Application
+from cherrypy._cplogging import LogManager
+from cherrypy.lib import profiler
+
+from WMCore.Configuration import ConfigSection, loadConfigurationFile
 
 #: Terminal controls to switch to "OK" status message colour.
 COLOR_OK = "\033[0;32m"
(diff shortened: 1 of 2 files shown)


amaltaro:
Another issue that I don't think we can fix are those print statements before the logging is setup, given that we disable the screen log, I don't see them in my tests. :(

amaltaro:
@ticoann I still want to do some pylint cleanup, but please review. And if you're concerned about the logging format or something else, we can discuss tomorrow. I think we can wait until Monday for the next testbed

src/python/WMCore/REST/Main.py
```diff
@@ -85,29 +85,28 @@ def access(self):
...
-                   and getattr(request.rfile.rfile, "bytes_read", None)
-                   and request.rfile.rfile.bytes_read) or "-",
-             'b': nout or "-",
-             'T': delta_time,
-             'AS': inheaders.get("CMS-Auth-Status", "-"),
-             'AU': inheaders.get("CMS-Auth-Cert", inheaders.get("CMS-Auth-Host", "")),
-             'AC': getattr(request.cookie.get("cms-auth", None), "value", ""),
-             'f': inheaders.get("Referer", ""),
-             'a': inheaders.get("User-Agent", "") })
+        msg = ('%(t)s %(H)s %(h)s "%(r)s" %(s)s'
```
ticoann: where does this msg is used now? I don't see any place here. Am I missing something?

amaltaro:
@ticoann I just pushed another commit moving the log.screen thing down to the moment we run the server. That way we can get all the initial setup logging out (including a dump of the service config) and still get rid of duplicate logs.
If you think it makes more sense, I can squash those commits.

ticoann:
Thanks Alan, patch looks good to me.

PR #8693, opened 2018-07-19
Fix dup exception handling and print in 8686

amaltaro:
Fix for PR #8686 where I forgot cleaning the file before putting it in a PR.
@ticoann please review


ticoann:
Sorry Alan, I merged that patch too early.

PR #8697, opened 2018-07-23
Create a WMWorkload property with a StepChain steps/parentage mapping

### LINKED ISSUE #8681: Create a map for Step and OutputDataset including child dataset information

amaltaro:
Fixes #8681
Replaces #8687

Creates a large data structure for StepChains containing relationship between steps, parent steps, parent dataset and output dataset. Only available and used for StepChain, such that we can later figure out the correct dataset parentage.

The new property is `stepParentageMapping` and in the near future it can replace `stepMapping`.
A few concepts worthy mentioning:
*) `ParentDataset`: is the real parent dataset of a given list of output datasets, thus it takes the KeepOutput into consideration.
*) `ParentStepNumber`: the direct parent step for the current step (so only looking at InputFromStep).

Also added an update function to make sure the new dataset names get updated into this structure at the end of the assignment step.

Examples of the whole structure can be seen in the comments below.

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
@@ -53,6 +53,7 @@ class StepChainWorkloadFactory(StdBase):
         # stepMapping is going to be used during assignment for properly mapping
         # the arguments to each step/cmsRun
         self.stepMapping = {}
+        self.stepParentageMapping = {}
 
     def __call__(self, workloadName, arguments):
         """
@@ -81,6 +82,9 @@ class StepChainWorkloadFactory(StdBase):
         # Create the first task
         firstTask = self.workload.newTask(taskConf['StepName'])
 
+        # it has to be called before the other steps are created
+        self.createStepMappings(arguments)
+
         # Create a proper task and set workload level arguments
         if isGenerator(arguments):
             self.workload.setDashboardActivity("production")
@@ -98,7 +102,10 @@ class StepChainWorkloadFactory(StdBase):
         if self.stepChain > 1:
             self.setupNextSteps(firstTask, arguments)
 
+        self.createStepParentageMappings(firstTask, arguments)
+
         self.workload.setStepMapping(self.stepMapping)
+        self.workload.setStepParentageMapping(self.stepParentageMapping)
         self.reportWorkflowToDashboard(self.workload.getDashboardActivity())
 
         # Feed values back to save in couch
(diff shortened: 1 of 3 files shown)


src/python/WMCore/WMSpec/StdSpecs/StepChain.py
```diff
@@ -108,6 +115,96 @@ def __call__(self, workloadName, arguments):
...
+            cmsRunNumber = "cmsRun%d" % i
+
+            self.stepParentageMapping.setdefault(stepName, {})
+            self.stepParentageMapping[stepName] = {'StepNumber': stepNumber,
+                                                   'StepCmsRun': cmsRunNumber,
+                                                   'ParentStepName': None,
+                                                   'ParentStepNumber': None,
+                                                   'ParentStepCmsRun': None,
+                                                   'ParentDataset': None,
+                                                   'OutputDatasets': {}}
```
ticoann: Can we change this name to OutputDatasetMap, since top level spec has "OutputDatasets" and wmstatsserver filter will recursively find the that parameter, it will give the mixed results.

ticoann:
Thank you very much Alan, patch looks good. If I understand correctly, this will include steps even if there is no output dataset or parent dataset, right? I will close my patch then.

src/python/WMCore/WMSpec/StdSpecs/StepChain.py
amaltaro: Sure, I'll update it to `OutputDatasetMap` since it's anyways key'ed by the output module.
Only one detail, in theory this property doesn't go to the request json doc, but it's set/get via WMWorkload object only. If we want to put it in the json doc, then I think we need to filter out some of these things and make a simple {'Step1': [out1, out2], 'Step2': [out3], ...}

ticoann: I think it would be useful to have information in reqmgr couchdb. Even though it won't be used in WMAgent directly. But don't worry about make it simple. We will do that at the time we put the data in reqmgr2 db. I mean we don't have to have the same formation in json and in spec.

ticoann: With that, I think it would be useful to add converter method in the workload helper. basically take current format and convert to whatever simpler format desired. (You don't have to add that in this patch. We can discuss the specific format first and I can add later.) Thanks.

amaltaro:
Seangchan, I made one line change to findParentStepWithOuputDataset such that it returns a string instead of list. 
I also renamed OutputDatasets to OutputDatasetMap as you suggested. 
Finally, I made a third unit test to make sure all these mappings are correct.

I still would like to remove `self.stepMapping` and use only the new mapping structure, which has all the information we need.
Then we need to discuss which information exactly we want to put in the request json and its format. How about something like 
`[{'ParentDset': 'blah1', 'OutDset': 'blah2'}, {'ParentDset': 'blah2', 'OutDset': 'blah3'},...]`

or maybe we rename 'OutDset' with 'ChildDset'?

ticoann:
> [{'ParentDset': 'blah1', 'OutDset': 'blah2'}, {'ParentDset': 'blah2', 'OutDset': 'blah3'},...]
> 
> or maybe we rename 'OutDset' with 'ChildDset'?

I think ChildDset is good. This structure is more flexible. But if only have one to many relations between parent and child datasets. we could do  [{'ParentDset': 'blah1', 'ChildDsets': ['blah2']}, .....]
in a way to imply the constraint. I don't have particular preference though.

ticoann:
test this please

amaltaro:
Actually, since we want to map which step outputs what, how about:
```
{'Step1': {'ParentDset': 'blah1', 'ChildDsets': ['blah2']},
 'Step2': {'ParentDset': 'blah2', 'ChildDsets': ['blah3', 'blah4],
...}
```
?

And `None` for no ParentDset or `[]` for no ChildDsets (KeepOutput=False). Note the singular vs plural trick too.

amaltaro:
test this please

amaltaro:
@ticoann I'm merging it and cutting a new release for testbed, but please go through it carefully once more. I'm updating the PR description soon.

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -123,6 +124,42 @@ def setStepProperties(self, assignArgs):
...
+                cmsRunNumber = parentMap[stepName]['StepCmsRun']
+                stepHelper = topLevelTask.getStepHelper(cmsRunNumber)
+                for outputModuleName in stepHelper.listOutputModules():
+                    outputModule = stepHelper.getOutputModule(outputModuleName)
+                    outputDataset = "/%s/%s/%s" % (outputModule.primaryDataset,
+                                                   outputModule.processedDataset,
+                                                   outputModule.dataTier)
+
+                    # now find and replace the old dataset by the new dataset name
+                    oldOutputDset = parentMap[stepName]['OutputDatasetMap'][outputModuleName][0]
```
amaltaro: Apparently an output module in a task/step has only one output dataset. Just wondering whether we should make it a string instead of a list? @ticoann

ticoann: So it does creates only one output modules per step which means parent dataset and child dataset is always one to one (at least within the same Step/TaskChain workflow.)

amaltaro: Actually each step can have several output modules, but everything leads me to think that each PSet/Step output module has only one output dataset. AND, the same output module is not present twice in the same PSet/Step. @vlimant maybe you can confirm it?

amaltaro: Yep, I'll make these changes on the other PR. Thanks!

PR #8698, opened 2018-07-24
scheleton for adding stepchain parentage

ticoann:
replace #8676 
#8700 need to be merged first

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
@@ -152,9 +152,12 @@ class DBSBufferBlock:
         self.data['files'].append(fileDict)
 
         # now add file to data
-        parentLFNs = dbsFile.getParentLFNs()
-        for lfn in parentLFNs:
-            self.addFileParent(child = dbsFile['lfn'], parent = lfn)
+        # If dataset_parent_list is defined, don't add file parentage.
+        # This means block is from StepChain workflow and parentage of file will be resolved later
+        if not self.data['dataset_parent_list']:
+            parentLFNs = dbsFile.getParentLFNs()
+            for lfn in parentLFNs:
+                self.addFileParent(child=dbsFile['lfn'], parent=lfn)
 
 
         # Do the algo
@@ -218,7 +221,7 @@ class DBSBufferBlock:
         Add the parent datasets to the data block
         """
 
-        self.data['ds_parent_list'].append({'parent_dataset': parent})
+        self.data['dataset_parent_list'].append(parent)
         return
 
     def setProcessingVer(self, procVer):
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferDataset.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferDataset.py
@@ -17,9 +17,8 @@ class DBSBufferDataset(WMConnectionBase, dict):
     information
     """
 
-    def __init__(self, path, id = -1,
-                 processingVer = None, acquisitionEra = None, validStatus = None,
-                 globalTag = None, parent = None, prep_id = None):
+    def __init__(self, path, id=-1,processingVer=None, acquisitionEra=None, validStatus=None,
+                 globalTag=None, parent=None, prep_id=None, is_stepchain=0):
         """
         Initialize the stored attributes and
         database connection.
@@ -35,6 +34,7 @@ class DBSBufferDataset(WMConnectionBase, dict):
         self['globalTag'] = globalTag
         self['parent'] = parent
         self['prep_id'] = prep_id
+        self['is_stepchain'] = is_stepchain
         self['subscriptions'] = []
 
     def exists(self):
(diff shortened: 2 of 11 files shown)


ticoann:
@amaltaro, Alan as we discussed I change to get the parentage information from spec instead of wmstats. It is a little more intrusive and complicated. Could you review the logic? I still need to test.

ticoann:
It assumes WMSpec helper has getParentDataset(dataset). I will add that function when your patch is in.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -211,14 +215,14 @@ def addBlockParent(self, parent):
         self.data['block_parent_list'].append({'block_name': parent})
         return
 
-    def addDatasetParent(self, parent):
+    def addDatasetParents(self, parents):
```
amaltaro: Why is it a list if there is only one parent dataset? I'd suggest to call it singular too

src/python/WMComponent/DBS3Buffer/DBSBufferDataset.py
```diff
@@ -17,9 +17,8 @@ class DBSBufferDataset(WMConnectionBase, dict):
     information
     """
 
-    def __init__(self, path, id = -1,
-                 processingVer = None, acquisitionEra = None, validStatus = None,
-                 globalTag = None, parent = None, prep_id = None):
+    def __init__(self, path, id=-1,processingVer=None, acquisitionEra=None, validStatus=None,
+                 globalTag=None, parent=None, prep_id=None, stepChain=0):
```
amaltaro: Minor, but will be helpful in the future :) How about renaming it to `is_stepchain`?

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
amaltaro: Why is it a list if there is only one parent dataset? I'd suggest to change it to singular too

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -152,9 +152,13 @@ def addFile(self, dbsFile, datasetType, primaryDatasetType):
         self.data['files'].append(fileDict)
 
         # now add file to data
-        parentLFNs = dbsFile.getParentLFNs()
-        for lfn in parentLFNs:
-            self.addFileParent(child = dbsFile['lfn'], parent = lfn)
+        # If dataset_parent_list is defined don't add the file parentage.
+        # This means it is block from StepChain workflow and parentage of file will be resloved later
```
amaltaro: typo and I think this blank line can go away. Or you fix it in a later commit.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -341,6 +341,10 @@ def loadBlocks(self):
             block = DBSBufferBlock(name=blockInfo['block_name'],
                                    location=blockInfo['origin_site_name'],
                                    datasetpath=blockInfo['datasetpath'])
+
+            parents = self.getStepChainParentage(blockInfo['datasetpath'])
+            block.addDatasetParents(parents)
```
amaltaro: `addParentDataset`?

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -816,3 +822,16 @@ def checkBlocks(self):
 
         # We're done
         return
+
+    def getStepChainParentage(self, dataset):
+        myThread = threading.currentThread()
+        parentDatasetDAO = self.daoFactory(classname="UpdateBlocks")
```
amaltaro: How about defining this DAO (and actually all the others) inside the __init__ method? And is it really going to be an update operation or just a select? Don't forget you need to create a new DAO for this.

In addition to that - there is no need to fix it in this PR - I just noticed we don't have the transaction handle in the algorithm method. In the last case it's handled in the BaseWorkerThread or something like that, but shouldn't we have the rollback and etc in algorithm?

amaltaro: Haa, you probably meant to write `GetStepChainParentDataset` :)

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
amaltaro: And when it's not a StepChain block, the DAO will return an empty list or something like that, shouldn't we then skip this `block.addDatasetParents` call such that we leave the correct parentage information for non-stepchain requests?

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -511,8 +511,13 @@ def _createDatasetSubscriptionsInDBSBuffer(self):
         existingTransaction = self.beginTransaction()
 
         subInfo = self.wmSpec.getSubscriptionInformation()
+
         for dataset in subInfo:
-            dbsDataset = DBSBufferDataset(path=dataset)
```
amaltaro: Are you sure this is the correct place for fetching and setting this parentage info? To me it seems it will only add IF the workflow was assigned with data subscriptions set.

amaltaro (changes requested):
General idea looks Ok, but can you explain in plain english the logic in the first comment of this PR? I got lost a couple of times.

There are a few things to change, I made several comments along the code.

PR #8707, opened 2018-07-30
revert fix for duplicate log update.

ticoann:
@amaltaro, I reverted the fix, but weird thing is that I don't see duplicate record for all the log. 
We can push this now and investigate more for next release. What do you think?

### DIFF
--- a/src/python/WMCore/REST/Main.py
+++ b/src/python/WMCore/REST/Main.py
@@ -469,7 +469,6 @@ class RESTDaemon(RESTMain):
         self.setup_server()
         self.install_application()
         cherrypy.log("INFO: starting server in %s" % self.statedir)
-        cherrypy.config.update({'log.screen': bool(getattr(self.srvconfig, "log_screen", False))})
         cherrypy.engine.start()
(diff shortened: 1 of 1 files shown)


amaltaro:
The problematic logs used to be those "Create request ..." and "Assign request...", like this one
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Request.py#L560

Logs from both testbed backends (vocms0132 and vocms0731) seem fine to me. The last day I injected/assigned workflows was on July 25 and I see all those log records over there (scanning by eye).

Did you also see a weird behaviour in your VM? Anyhow, if there are no dup records with this patch, I think we can just merge it (even though there is no explanation at all about the dup logs then).

ticoann:
Weird, Now it works without this patch and create a duplicate with this one. I am closing it. But I am not sure why it has cherrypy.error message event though there is no errors.

amaltaro:
I found an openstack link giving some explanation about this cherrypy.error thing. However, you can also see it in my first comment/documentation in this PR: 
https://github.com/dmwm/WMCore/pull/8691#issue-202559993

PR #8715, opened 2018-07-31
make backward compatible for the url change

ticoann:
related to #8712


amaltaro:
Thanks, Seagnchan. Tested in my VM and it works just fine.

Issue #8718, opened 2018-08-01
Remove the "stautus stall > 2 days" alert from wmstats


amaltaro:
With the amount of request we have in the system and priority changes, it's just not reliable (without mentioning that tab has now 2300 requests listed in different categories).

Then I'd move the "failed" box to the very top, followed by the "assigned > 2h". We might review as well the "Site Error" and "Config Error" alerts. Or we just leave them at the bottom of the tab.

PR #8722, opened 2018-08-01
fix inputdatset filter

### LINKED ISSUE #8721: Filter by "input dataset" in wmstats does not work
Or maybe it doesn't work for TaskChain workflows only...
Example in production is this input dataset
/ttH_M125_mtop1735_TuneCP5_13TeV-powheg-pythia8/RunIIFall17DRPremix-PU2017_94X_mc2017_realistic_v11-v1/AODSIM

which belongs to this workflow (which is active)
https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_task_HIG-RunIIFall17MiniAODv2-01000__v1_T_180414_215925_7580

ticoann:
fixes #8721 and #8718

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/DataStruct/WMStats.GenericRequests.js
+++ b/src/couchapps/WMStats/_attachments/js/DataStruct/WMStats.GenericRequests.js
@@ -295,6 +295,9 @@ WMStats.GenericRequests.prototype = {
         if (property == 'request_status') {
             return workflowData[property][workflowData[property].length - 1].status;
         }
+        if (property == 'inputdataset') {
+            return WMStats.Utils.getInputDatasets(workflowData);
+        }
         return workflowData[property];
     },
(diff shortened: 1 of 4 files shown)


ticoann:
Alan, I tested in my vm. I will merge and make the pull request for new testbed. But please review.

amaltaro:
Thank you very much for the quick fix, Seangchan.

Going back to the code, why do we check for the string "None" here
https://github.com/ticoann/WMCore/blob/f5cd9f620f8a1276292081aacf446113e2f17d17/src/couchapps/WMStats/_attachments/js/WMStats.Utils.js#L231
?

Isn't it None and when we dump it into json it becomes null?

ticoann:
I am not sure but there was a case it returned "None".  I don't think this is still the case but I didn't want to risk of breaking this time.

PR #8726, opened 2018-08-07
Use port 8443 to communicate with cmsweb services

### LINKED ISSUE #8565: Migrate machine tools to port 8443
We need to identify machine tools which access cmsweb and start switching them to use port 8443 to access cmsweb.

amaltaro:
Fixes #8565
Replaces #8704 

Instead of changing every single place that defines a cmsweb endpoint, we can simply change it in this base class and make sure the other Services modules inherit from it (there seems to be a few that don't).

### DIFF
--- a/src/python/WMCore/Services/Service.py
+++ b/src/python/WMCore/Services/Service.py
@@ -103,6 +103,10 @@ class Service(dict):
         if not cfg_dict['endpoint'].endswith('/'):
             cfg_dict['endpoint'] = cfg_dict['endpoint'].strip() + '/'
 
+        # setup port 8443 for cmsweb services
+        if cfg_dict['endpoint'].startswith("https://cmsweb"):
+            cfg_dict['endpoint'] = cfg_dict['endpoint'].replace('.cern.ch/', '.cern.ch:8443/', 1)
+
         #set up defaults
         self.setdefault("inputdata", {})
         self.setdefault("cacheduration", 0.5)


src/python/WMCore/Services/Service.py
```diff
@@ -103,6 +103,10 @@ def __init__(self, cfg_dict = None):
         if not cfg_dict['endpoint'].endswith('/'):
             cfg_dict['endpoint'] = cfg_dict['endpoint'].strip() + '/'
 
+        # setup port 8443 for cmsweb services
+        if cfg_dict['endpoint'].startswith("https://cmsweb"):
```
ticoann: how about cmsweb-testbed is it using the same port?

ticoann:
smart. :smile:, Is there any case we are not using Service class?  how about dbs apis?

src/python/WMCore/Services/Service.py
amaltaro: Yes, it has that port enabled too

amaltaro:
Yes, I see quite a few, including DBS3Reader and possibly some legacy modules.
I'll create those changes in another PR, if it's straight forward, then we add to today's agent release, otherwise this change will touch the bulk of the cmsweb requests.

amaltaro:
BTW, I ran the DMWM tests on an agent with this patch and I see data was uploaded to the DQM GUI, workflows properly completed, dbs and phedex stuff also done. I think we're good with this change.

ticoann:
Thanks Alan,

amaltaro:
Before I forget, @belforte @hufnagel if you build T0/CRAB using tag 1.1.15 or newer, this change will be in. So any of those modules under WMCore/Services that inherit from `Service` class, will be talking to CMSWEB services through the port 8443 instead of 443. 

It should be transparent to your services, but just in case :)

belforte:
thanks

belforte:
Hi Alan, I am likely missing something, but looking at what's new there (to get a sense
of what to test) I do not see anything in the way of DBS URL.

belforte:
forgot the URL I have used to compare
https://github.com/dmwm/WMCore/compare/1.1.14_crab...1.1.15_cmsweb

ticoann:
Hi Stefano, you are right. DBS is not included here. We can temporary switch DBS port here by overwriting URLs in WMCore/Service/DBS package, Or wait for DBS client source handle this. I think it is cleaner dbs client handle that. But if you prefer we can add that here temporary.

belforte:
thanks @ticoann. Let's wait. I do not need do to test things now. Only wanted to have a sense of what to look if I build with new tag.

PR #8729, opened 2018-08-08
Workaround for condor submitMany; remove legacy condor parameters fro…

amaltaro:
…m job ads
Follow up from #8493 (see johnkn comments) and reverting the plugin changes added in #8686

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -11,7 +11,6 @@ import os.path
 import re
 import threading
 import time
-import pickle
 import classad
 import htcondor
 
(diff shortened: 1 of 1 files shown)


src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -528,12 +512,6 @@ def getClusterAd(self):
         ad['JobAdInformationAttrs'] = ("JobStatus,QDate,EnteredCurrentStatus,JobStartDate,DESIRED_Sites,"
                                        "ExtDESIRED_Sites,WMAgent_JobID,MachineAttrGLIDEIN_CMSSite0")
 
-        # TODO: remove when 8.5.7 is deployed
```
amaltaro: @bbockelm if you can have a quick look at this code. According to the comment, it would be safe to remove it now since we're running 8.6.7 / 8.6.11

bbockelm: This looks correct.

The ultimate test is to see if the job submitter produces a job with the `JOB_Site` attribute set.  If so, we're good.

amaltaro: And this is what I was going to ask you, thanks for the hint. Unfortunately those jobs don't have this JOB_Site attribute.

I'm rolling back this change and we can get back to it in the near future, if you know what happened here, I'm all ears.

amaltaro:
test this please

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
ticoann: Thanks Alan, I will merge this and if we need the change we will create another issue.

PR #8731, opened 2018-08-09
Replace print by logging in DBS3Reader

amaltaro:
Complement for #8593

### DIFF
--- a/bin/fix-dbs-parentage
+++ b/bin/fix-dbs-parentage
@@ -57,7 +57,7 @@ def getMissingParentSummaryUsingBlock(datasetList, fileFix=False, insertFlag=Fal
     start = int(time.time())
 
     if insertFlag:
-        answer = raw_input("fixing DBS parentage: Are you sure?(y/n): ").lower()
+        answer = input("fixing DBS parentage: Are you sure?(y/n): ").lower()
 
     for dataset in datasetList:
         if isinstance(dataset, dict):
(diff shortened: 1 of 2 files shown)


bin/fix-dbs-parentage
```diff
@@ -57,7 +57,7 @@ def getMissingParentSummaryUsingBlock(datasetList, fileFix=False, insertFlag=Fal
     start = int(time.time())
 
     if insertFlag:
-        answer = raw_input("fixing DBS parentage: Are you sure?(y/n): ").lower()
+        answer = input("fixing DBS parentage: Are you sure?(y/n): ").lower()
```
ticoann: Hmm, I know it is more compatible with Python 3, when I tried it before it didn't work (Aside it evaluates the input).  I will try again.

ticoann:
I left print there since it is only using by script and logging destination is not set there. I will merge this after the validation test.

bin/fix-dbs-parentage
amaltaro: Ahhh, I just noticed it indeed complains in ipython. But if you type "y" instead of y, it works, apparently. So one needs to enclose the input in quotes, which is weird! Is it acceptable, otherwise we'll keep getting this failure in jenkins.

amaltaro:
I see, sorry about that. I flagged this PR as "Do not merge yet" then.

amaltaro:
@ticoann I think you have already fixed these things, didn't you? If not, then please reopen this PR

ticoann:
Thanks Alan, Yes it is merged.

PR #7112, opened 2016-08-26
converting stringIO #6935

SaltumDis:

PR #8733, opened 2018-08-09
[PY3] Replace StringIO by the python3 compatible io module

amaltaro:
Superseeds #7112 (before we forget it)

Given that Adelina had a few modules directly importing `io`, I went ahead and simply imported it instead of trying to import the one available in python2.7.

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -38,7 +38,8 @@ for row in data:
 """
 from __future__ import print_function
 
-import cStringIO as StringIO
+
+# system modules
 import httplib
 import json
 import logging
@@ -46,19 +47,13 @@ import os
 import re
 import subprocess
 import sys
-
+import pycurl
+from io import BytesIO
 try:
     from urllib import urlencode
 except ImportError:
     # PY3
     from urllib.parse import urlencode
-
-# python3
-if sys.version.startswith('3.'):
-    import io
-
-# 3d-party libraries
-import pycurl
 
 
 class ResponseHeader(object):
@@ -198,9 +193,9 @@ class RequestHandler(object):
         curl.setopt(pycurl.URL, str(url))
         if headers:
             curl.setopt(pycurl.HTTPHEADER, \
-                        ["%s: %s" % (k, v) for k, v in headers.items()])
-        bbuf = StringIO.StringIO()
-        hbuf = StringIO.StringIO()
+                    ["%s: %s" % (k, v) for k, v in headers.items()])
+        bbuf = BytesIO()
+        hbuf = BytesIO()
         curl.setopt(pycurl.WRITEFUNCTION, bbuf.write)
         curl.setopt(pycurl.HEADERFUNCTION, hbuf.write)
         if capath:
@@ -417,12 +412,8 @@ def getdata(urls, ckey, cert, headers=None, options=None, num_conn=100, cookie=N
             if cookie and url in cookie:
                 curl.setopt(pycurl.COOKIEFILE, cookie[url])
                 curl.setopt(pycurl.COOKIEJAR, cookie[url])
-            if sys.version.startswith('3.'):
-                bbuf = io.BytesIO()
-                hbuf = io.BytesIO()
-            else:
-                bbuf = StringIO.StringIO()
-                hbuf = StringIO.StringIO()
+            bbuf = BytesIO()
+            hbuf = BytesIO()
             curl.setopt(pycurl.WRITEFUNCTION, bbuf.write)
             curl.setopt(pycurl.HEADERFUNCTION, hbuf.write)
             mcurl.add_handle(curl)
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -7,13 +7,13 @@ independent python structure
 
 """
 
-import StringIO
 import imp
 import inspect
 import json
 import pickle
 import sys
 from functools import reduce
+from io import BytesIO
 
 
 class PSetHolder(object):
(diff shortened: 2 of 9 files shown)


amaltaro:
I don't get this message. I made a few commits/PRs today and the bot didn't complain...

amaltaro:
test this please

amaltaro:
Failure is coming from a base class, that's why so many failures:
```
...
  File "/build/cmsbld/jenkins/workspace/DMWM-PR-unittests/SLICE/2/label/cms-dmwm-cc7/code/src/python/WMCore/Services/Requests.py", line 183, in makeRequest_pycurl
    ckey=ckey, cert=cert, capath=capath)
  File "/build/cmsbld/jenkins/workspace/DMWM-PR-unittests/SLICE/2/label/cms-dmwm-cc7/code/src/python/WMCore/Services/pycurl_manager.py", line 252, in request
    curl.perform()
error: (23, 'Failed writing header')

Standard Error
TypeError: unicode argument expected, got 'str'
```

This needs further investigation. Likely a good candidate for Dario @mapellidario in the coming weeks.

davidlange6:
hi @amaltaro  - from what I read, these changes are not expected to work for python2.

vkuznet:
The way I deal with this problem in DAS for both python2 and python3 is to use this import
```
try:
    import cStringIO as StringIO
except ImportError: # python3
    import io as StringIO
except:
    import StringIO
```
This only solve issue with proper import which will work in both python versions. Then you need to ensure a proper input to StringIO call as it requires in concrete python version, but it is slightly different story. And, when you need to adjust BytesIO you can do this:
```
        if  sys.version.startswith('3.'):
            bbuf = io.BytesIO()
        else:
            bbuf = StringIO.StringIO()
```
Doing this way you don't need explicitly import `io` and will avoid issue with python2/python3.

davidlange6:
Yes - I agree.

On Jun 11, 2020, at 3:26 PM, Valentin Kuznetsov <notifications@github.com<mailto:notifications@github.com>> wrote:



The way I deal with this problem in DAS for both python2 and python3 is to use this import

try:
    import cStringIO as StringIO
except ImportError: # python3
    import io as StringIO
except:
    import StringIO


This only solve issue with proper import which will work in both python versions. Then you need to ensure a proper input to StringIO call as it requires in concrete python version, but it is slightly different story. And, when you need to adjust BytesIO you can do this:

        if  sys.version.startswith('3.'):
            bbuf = io.BytesIO()
        else:
            bbuf = StringIO.StringIO()


Doing this way you don't need explicitly import io and will avoid issue with python2/python3.

amaltaro:
I guess I have to read further about `BytesIO` and `StringIO`, but when in python3, don't we need to know whether we are dealing with a stream of bytes or a text stream; thus calling the correct class?

Futurize claims only one import would suffice, provided we call the correct class to handle the correct data:
https://python-future.org/compatible_idioms.html#stringio

khurtado:
If we are going to depend on http://python-future.org for porting to Py2/3 in the future, should we follow scikit-bio example for StringIO and use:
```
from future.utils.six import StringIO
```
instead?
http://scikit-bio.org/docs/0.1.3/development/py3.html#stringio-and-bytesio

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -418,11 +413,11 @@ def getdata(urls, ckey, cert, headers=None, options=None, num_conn=100, cookie=N
...
             if sys.version.startswith('3.'):
-                bbuf = io.BytesIO()
-                hbuf = io.BytesIO()
+                bbuf = BytesIO()
+                hbuf = BytesIO()
             else:
-                bbuf = StringIO.StringIO()
-                hbuf = StringIO.StringIO()
+                bbuf = StringIO()
+                hbuf = StringIO()
```
khurtado, line 420: Shouldn't this just be removed? Attempting to replace `StringIO.StringIO()` with `io.StringIO` would fail because the former can handle bytes and strings, while the second can't. 
If we have `io` in python2, then `io.BytesIO` should work for both python2 and python3 with pycurl.

amaltaro: You're correct Kenyi. It does fix the unit test I was looking at btw, let's see how it goes with the others.

amaltaro:
Kenyi, I think your scikit-bio got lost with my last commit. So let me reference that documentation/example here again (short and straight to the point!):
http://scikit-bio.org/docs/0.1.3/development/py3.html#stringio-and-bytesio

FYI @mapellidario

amaltaro:
Given all the problems that popped up in this "simple" fix, I guess I'll test it for real in central services and wmagent.
Nonetheless, if you guys could review it, it's even better. @khurtado @mapellidario

mapellidario (commented):
My opinion is far from being the definitive one. I have almost no experience with WMCore and this review should only be intended as a summary of what I found while investigating how to modify py2 code so that it is compatible with py3, too.

The [general suggestion](https://nedbatchelder.com/text/unipain.html#h_pain_relief) is to convert from bytes to unicode as soon as possible, manipulate unicode, then convert back from unicode to bytes as late as possible. This requires on fixing an encoding, keeping in mind that if "ascii" is chosen, it may break when receiving a non-ascii unicode character. Nonetheless, I saw that cStringIO was used before these changes, so I guess you did plan to support only ascii encoding.

I just a quick read of the files that have been changed and it does not seem that we are doing any manipulation on the strings, so we should be fine using bytes instead of unicode. And this PR seems good for me.

I have a question, though. While almost everywhere you set pycurl WRITEFUNCTION to be BytesIO, why is it still StringIO in [WMCore/Services/McM/McM.py](https://github.com/dmwm/WMCore/blob/5021aa96398585ad6e44a6c472638954c890988e/src/python/WMCore/Services/McM/McM.py#L76) ? Am I missing something?

amaltaro:
Dario, thanks for your review. I updated the McM module as well.

Honestly speaking, I'm not sure whether these changes are going to work out of the box in python3. It's very possible that some of them will have to be revisited once we have a working python3 environment.

PR #8737, opened 2018-08-13
clean up out put of the script and add more instruction

ticoann:
Test with old script and compare the results. Just update the some prints statement for comparison.

### DIFF
--- a/bin/fix-dbs-parentage
+++ b/bin/fix-dbs-parentage
@@ -33,9 +33,13 @@ insert options (when missing parentage is found you can specify to insert to DBS
 Method for the search: (By default it searches only for blocks with missing parents first then searches file parentage for those blocks)
 
 -f: search for all the files in the dataset for missing parentage - very expensive to run only use when block has parent
-    but some files in the block doesn't have parents
+    but some files in the block doesn't have parents. This is normally used -d option for only one dataset.
 
 Usage:
+
+Uses block parentage to fix the problem (faster checks block parentage first to fix the missing parentage.)
+If some files has missing parents but some files has parents in the same block, this call doesn't fix those files parentage.
+In that case use -f option (slow)
 
 $manage execute-agent fix-dbs-parentage -s 1/1/2017 -e 1/2/2017 -p --insert
 $manage execute-agent fix-dbs-parentage -p --insert -d /SingleMuon/Run2016D-03Feb2017-v1/MINIAOD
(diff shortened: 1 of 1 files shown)


ticoann:
@amaltaro, Alan please review

bin/fix-dbs-parentage
```diff
@@ -58,29 +62,29 @@ def getMissingParentSummaryUsingBlock(datasetList, fileFix=False, insertFlag=Fal
...
-                print("Aborting the procedure, Nothing updated")
-                exit(1)
-        else:
-            print("Only testing perfomance: No insert")
-            failedBlocks = dbsService.insertMissingParentageForAllFiles(dsName,  filterFilesWithParents=fileFix,
+        if fileFix:
+            failedBlocks = dbsService.insertMissingParentageForAllFiles(dsName, filterFilesWithParents=fileFix,
                                                                         insertFlag=insertFlag)
+        else:
+            failedBlocks = dbsService.fixMissingParentageDatasets(dsName, insertFlag=insertFlag)
```
amaltaro: Seangchan, are these two methods `insertMissingParentageForAllFiles` and `fixMissingParentageDatasets` actually writing to DBS? If so, then I'd suggest to move then somewhere else or rename this `getMissingParentSummaryUsingBlock` method name, such that it's clear what's only reading and what is writing to DBS.

bin/fix-dbs-parentage
```diff
@@ -58,29 +62,29 @@ def getMissingParentSummaryUsingBlock(datasetList, fileFix=False, insertFlag=Fal
 
     if insertFlag:
```
amaltaro: If one doesn't pass `insertFlag`, it will run this method all the way until the insert and fix methods below.

amaltaro (changes requested):
I made a couple of comments in the code.

bin/fix-dbs-parentage
ticoann: It only insert to DBS when insertFlag is set to True. I can update the name.

bin/fix-dbs-parentage
ticoann: Yes but with insertFlag = False. which run the those method without actually inserting to dbs. This is just a test case to check how long it takes and correctness of the parentage.

bin/fix-dbs-parentage
amaltaro: Seangchan, I'd suggest then to update this method getMissingParentSummaryUsingBlock to updateParentageInfo or something like that.

ticoann: I will do that. Thanks,

amaltaro:
Seangchan, is there any difference between those 2 scripts?

ticoann:
One script is for exact same functionality for reqmgr thread (checking by announced, archived data). 
Sorry I wasn't plan to include it here, since it need some refinement and I just created for the test. I will remove it.

amaltaro:
Seangchan, don't forget removing the script you said you didn't mean to add to this PR please. Otherwise changes look good (just make sure that script is executable, if you're planning to run it with $manage)

amaltaro (approved):

PR #8738, opened 2018-08-13
StepChain fix in reqmgr

ticoann:
Initial version for StepChain parentage fix

### DIFF
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
@@ -1,0 +1,87 @@
+from __future__ import (division, print_function)
+import time
+from collections import defaultdict
+
+from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
+from WMCore.Services.RequestDB.RequestDBWriter import RequestDBWriter
+from WMCore.Services.DBS.DBS3Reader import DBS3Reader
+
+
+def getChildDatasetsForStepChainMissingParent(reqmgrDB, status):
+    """
+    :param status: workflow status
+    :return: set of child dataset
+    """
+    results = reqmgrDB.getStepChainDatasetParentageByStatus(status)
+
+    requestsByChildDataset = defaultdict(set)
+
+    for reqName, info in results.items():
+
+        for dsInfo in info.values():
+            if dsInfo["ParentDset"]:
+                for childDS in dsInfo["ChildDsets"]:
+                    requestsByChildDataset[childDS].add(reqName)
+    return requestsByChildDataset
+
+
+class StepChainParentageFixTask(CherryPyPeriodicTask):
+    """
+    Upldates StepChain parentage periodically
+    """
+
+    def __init__(self, rest, config):
+
+        super(StepChainParentageFixTask, self).__init__(config)
+        self.reqmgrDB = RequestDBWriter(config.reqmgrdb_url)
+        self.dbsSvc = DBS3Reader(config.dbs_url, logger=self.logger)
+        self.statusToCheck = ["announced", "normal-archived"]
+
+    def setConcurrentTasks(self, config):
+        """
+        sets the list of functions which
+        """
+        self.concurrentTasks = [{'func': self.fixStepChainParentage, 'duration': config.parentageFixDuration}]
+
+    def fixStepChainParentage(self, config):
+        """
+        Look through the stepchain workflows with ParentageResolved flag is False.
+        Fix the StepChain parentage and update the ParentageResolved flag to True
+        """
+        self.logger.info("Updating parentage for StepChain workflows for ", self.statusToCheck)
+        childDatasets = set()
+        requests = set()
+        requestsByChildDataset = {}
+        for status in self.statusToCheck:
+            reqByChildDS= getChildDatasetsForStepChainMissingParent(self.reqmgrDB, status)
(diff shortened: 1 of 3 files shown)


ticoann:
https://github.com/dmwm/deployment/pull/658/ need to be in first.

src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
```diff
@@ -0,0 +1,61 @@
...
+        sets the list of functions which
+        """
+        self.concurrentTasks = [{'func': self.fixStepChainParentage, 'duration': config.parentageFixDuration}]
+
+    def fixStepChainParentage(self, config):
+        """
+        Look through the stepchain workflows with ParentageResolved flag is False.
+        Fix the StepChain parentage and update the ParentageResolved flag to True
+        """
+        childDatasets = set()
```
amaltaro: I'd suggest first thing to log that this thread is running

src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
```diff
@@ -0,0 +1,61 @@
...
+        requests = set()
+        for status in self.statusToCheck:
+            info = getChildDatasetsForStepChainMissingParent(self.reqmgrDB, status)
+            childDatasets = childDatasets.union(info["ChildDatasets"])
+            requests = requests.union(info["Requests"])
+
+        for childDS in childDatasets:
+            self.dbsSvc.fixMissingParentageDatasets(childDS, insertFlag=True)
```
amaltaro: Then add another log like (check if self.logger is the correct one) 
```
msg = "Fixing dataset parentage for: %s" % childDS
msg += " Counter %d out of %d" % (dsetCount, len(childDatasets))
self.logger.info(msg)
```

such that we can monitor:
*) which datasets it fixes
*) roughly how long it takes
*) and how much work it has got in this cycle

src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
```diff
@@ -0,0 +1,61 @@
...
+            self.logger.info("Fixing %s parentage", childDS)
+
+        for request in requests:
+            self.reqmgrDB.updateRequestProperty(request, {"ParentageResolved": True})
```
amaltaro: Same here, something like "Marking %d requests as ParentageResolved=True"

amaltaro (changes requested):
Patch looks good Seangchan. Nonetheless, I made some comments in the code.

ticoann:
Thanks Alan for all valuable comments. I will update what you suggested. One thing I forgot to add is retry logic, since there might be a case if fails update certain blocks. (That is why the method is returning the failed block list) I will get that and need to retry that as well. To do that we need to have map for child dataset to request name so we won't update the status of those workflows.
It won't retry all the blocks on that dataset, but only the blocks fails to update.

amaltaro:
Good point, I missed that part. Ideally we should try only the blocks that failed, but if that's too much micro-management, just go with the dataset that is retrieved from ReqMgr.

ticoann:
@amaltaro, I updated patch including the failure retry logic. Could you review again? I still need to test the change.

src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
```diff
@@ -0,0 +1,87 @@
...
+        totalChildDS = len(childDatasets)
+        fixCount = 0
+        for childDS in childDatasets:
+            start = int(time.time())
+            failedBlocks = self.dbsSvc.fixMissingParentageDatasets(childDS, insertFlag=True)
+            end = int(time.time())
+            timeTaken = end - start
+            if failedBlocks:
+                self.logger.warning("Failed to fix the parentage for %s will be retried: time took: %s (sec)",
+                                 failedBlocks, timeTaken)
```
amaltaro: I like printing the block name, but depending on the case this `failedBlocks` might print a lot of blocks (like when cmsweb is down). But I think we can watch this thread for a little before changing it, if needed.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -57,12 +57,13 @@ class DBS3Reader(object):
...
     _datatiers = {}
 
-    def __init__(self, url, **contact):
+    def __init__(self, url, logger=None, **contact):
 
         # instantiate dbs api object
         try:
             self.dbsURL = url
             self.dbs = DbsApi(url, **contact)
+            self.logger = logger or logging
```
amaltaro: Isn't it `logging.getLogger()`?

src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
ticoann: I expect the failure of updating parentage is very rare occasion unless there is real problem. (dbs is not responsive, or some code failure).

src/python/WMCore/Services/DBS/DBS3Reader.py
ticoann: I am  always confused how to the log should be set. getLogger create new logger instance from the logging which has original setting. I think it is better to use new instance for each module with name to find which module is logging but I don't think current code will have failure.

PR #8748, opened 2018-08-17
improve the performance of dbs file insert

ticoann:
fixes #8740

PR #8758, opened 2018-08-22
Improvements for workQueueManager - wmagent version

amaltaro:
1.1.14_wmagent version of
#8748 : (only first commit) improve the performance of dbs file insert
#8755 : Make WorkQueueManagerWMBSFileFeeder slightly more verbose

I'm pushing it to vocms0254 only, because it has an urgent pilot request to acquire and it's acquiring some bad ACDC workflows.


ticoann:
Merging this, if you need to apply other agents, we will do that.

PR #8766, opened 2018-08-23
set default logger for DBS3Reader

ticoann:


amaltaro (approved):
Thanks, it looks good!

PR #8771, opened 2018-08-28
fix import and logging for crabserver

ticoann:
@amaltaro, Alan please review. need to used with https://github.com/dmwm/deployment/pull/664

### DIFF
--- a/src/python/WMCore/REST/Main.py
+++ b/src/python/WMCore/REST/Main.py
@@ -29,6 +29,8 @@ from cherrypy import Application
 from cherrypy._cplogging import LogManager
 from cherrypy.lib import profiler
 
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
Thanks for fixing it, Seangchan.

ticoann:
Hmm, somehow with the patch and deployment script it doesn't print out the log at all. Now when I revert back the change and remove the configuration change, it still does not printing the log.

amaltaro:
I don't get it either. But I've seen cases where the first REST API call isn't recorded, but then everything afterwards is. I've seen cases that no matter what calls are made, there are no logs recorded at all.... sigh...
However, the common case is that after the first api call, everything gets properly logged.

PR #8775, opened 2018-08-29
Import Tools in REST.Main

amaltaro:
This import was removed in https://github.com/dmwm/WMCore/pull/8691
unfortunately it's actually used by CRABServer during startup, setting up tools attributes or something like that.
Fix https://github.com/dmwm/WMCore/pull/8771


amaltaro:
I'm merging it and cutting a new tag for testbed.

PR #8776, opened 2018-08-31
send sigterm for memory failure

### LINKED ISSUE #8345: PerformanceMonitor.py only doing soft kill on memory?
Seeing in this workflow now with cmsRun processes reaching >100G in many places at FNAL currently.  The jobs realize this and only do a casual attempt at suicide:

workflow: Request pdmvserv_task_PPD-RunIISummer17PrePremix-00004__v1_T_171113_092409_4193

2017-11-17 03:53:45,590:ERROR:PerformanceMonitor:Attempting to kill step using SIGUSR2
2017-11-17 03:58:45,677:INFO:PerformanceMonitor:Retrieved following performance figures:
2017-11-17 03:58:45,677:INFO:PerformanceMonitor:RSS: 8633852;  VSize: 62504264; PCPU: 308; PMEM: 6.5
2017-11-17 03:58:45,677:ERROR:PerformanceMonitor:Error in CMSSW step cmsRun1
Job has exceeded maxRSS: 6041600
Job has RSS: 8633852

2017-11-17 03:58:45,680:ERROR:PerformanceMonitor:Attempting to kill step using SIGUSR2
2017-11-17 04:03:45,757:INFO:PerformanceMonitor:Retrieved following performance figures:
2017-11-17 04:03:45,757:INFO:PerformanceMonitor:RSS: 10299568;  VSize: 64189000; PCPU: 284; PMEM: 7.7
2017-11-17 04:03:45,758:ERROR:PerformanceMonitor:Error in CMSSW step cmsRun1
Job has exceeded maxRSS: 6041600
Job has RSS: 10299568

...

 

2017-11-17 10:59:29,495:ERROR:PerformanceMonitor:Attempting to kill step using SIGUSR2
2017-11-17 11:04:29,656:INFO:PerformanceMonitor:Retrieved following performance figures:
2017-11-17 11:04:29,656:INFO:PerformanceMonitor:RSS: 102040000;  VSize: 187908936; PCPU: 187; PMEM: 77.2
2017-11-17 11:04:29,657:ERROR:PerformanceMonitor:Error in CMSSW step cmsRun1
Job has exceeded maxRSS: 6041600
Job has RSS: 102040000

We only issue a sigterm on time based fails?  Not eventually on memory?

ticoann:
fixes #8345

### DIFF
--- a/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
+++ b/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
@@ -219,11 +219,13 @@ class PerformanceMonitor(WMRuntimeMonitor):
             msg += "Job has exceeded maxRSS: %s\n" % self.maxRSS
             msg += "Job has RSS: %s\n" % rss
             killProc = True
+            killHard = True
             reason = 'RSS'
         elif self.maxVSize is not None and vsize >= self.maxVSize:
             msg += "Job has exceeded maxVSize: %s\n" % self.maxVSize
             msg += "Job has VSize: %s\n" % vsize
             killProc = True
+            killHard = True
             reason = 'VSZ'
         elif self.hardTimeout is not None and self.softTimeout is not None:
             currentTime = time.time()


ticoann:
I am not sure whether this is right things to do. There must be a reason it doesn't send SIGTERM when originally this is written.

amaltaro:
I think we need to have a mechanism which sends a sigterm only if the previous signal (I forgot which one is that) didn't have any effect. We can assume that if any of those kills is triggered a second time (usually 5min later), then we'd send a sigterm instead. Allowing the job to shut itself for 5min.

hufnagel:
No, sigterm is correct. If you kill a job, you want to see in the job report that it was actually killed. At the moment, a soft memory kill is indistinguishable from a CMSSW job that was configured to process less events. Which is dangerous.

One could argue that this is actually a CMSSW bug, that it should never exit with code 0 when it receives an external signal. But with anything CMSSW, it would take years for such a change to take effect since we have so many workflows using older CMSSW.

The way CMSSW interpretes signals right now, the whole soft/hard kill concept is kind of useless. You always want a hard kill.

Before we can merge this change I would start a larger discussion though or at least announce that this change is coming...

hufnagel:
Alternative that would keep the soft/hard kill concept intact: change soft kill from SIGUSR2 to SIGTERM and hard kill from SIGTERM to SIGKILL

amaltaro:
The only case I've seen a job not gracefully ending was with over consumption of memory and, that's a very small fraction of it. My suggestion is to keep sending the current signals - which we know it works - and simply send SIGTERM for RSS/VSZ, in case a SIGUSR2 has been already sent.

If you want to have a review of these signals and their meaning CMSSW-wise, then I think we can do that in parallel. It will anyways affect all our tools running CMSSW, so it won't be accomplished in a couple of days.

amaltaro (changes requested):
See my comment above. I'd rather keep sending the same signals, but send a hard kill if the soft kill didn't work in the last time this monitor ran.

amaltaro:
Superseeded by #8972 , thanks Seangchan!

PR #8777, opened 2018-08-31
fix wrong link in workload summary

ticoann:
@amaltaro, Alan please review. It seems it doesn't affect things in the surface though, we might not use jquery in the workloadsummary code.

### DIFF
--- a/src/couchapps/WorkloadSummary/templates/histogramByWorkflow.html
+++ b/src/couchapps/WorkloadSummary/templates/histogramByWorkflow.html
@@ -1,8 +1,8 @@
 <html>
   <head>
     <title>{{{workflow}}} Performance Histogram</title>
-    <script src="/_utils/script/json2.js"></script>
-    <script src="/_utils/script/jquery.js"></script>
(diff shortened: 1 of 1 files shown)


amaltaro:
Seangchan, can you please elaborate on the issue? Example? Maybe just create an issue and point this fix to it.

ticoann:
Sorry Alan, her is the case

When you go to any workloadsummary page.

https://cmsweb.cern.ch/couchdb/workloadsummary/_design/WorkloadSummary/_show/histogramByWorkflow/vlimant_ACDC3_task_HIG-RunIIFall17wmLHEGS-01285__v1_T_180831_162104_8438

You can see this javascript error. which means it doesn't load this 2 library. It is due to wrong path for those script. We could load this locally, but cdn has advantage, it will be cached in the browsers, it will have a little performance gain. What I don't know is whether these libraries are actually used.

json2.js:1 Failed to load resource: the server responded with a status of 404 (Not Found)
jquery.js:1 Failed to load resource: the server responded with a status of 404 (Not Found)

PR #8781, opened 2018-09-03
Remove lustre from the ignoredDisks list; added /mnt/ramdisk

amaltaro:
I saw one of the newest agents going into drain because `/mnt/ramdisk` was almost full. We have to of course blacklist that mount point.
I can no longer see lustre mounted on the FNAL agents, so I'm removing that as well.

### DIFF
--- a/src/python/WMCore/Agent/DefaultConfig.py
+++ b/src/python/WMCore/Agent/DefaultConfig.py
@@ -10,7 +10,7 @@ DEFAULT_AGENT_CONFIG = {
     # disk usage percentage for warning
     "DiskUseThreshold": 85,
      # list of disks which shouldn't be included in monitoring for the Threshold
-    "IgnoreDisks": ["/lustre/unmerged"],
+    "IgnoreDisks": ["/mnt/ramdisk"],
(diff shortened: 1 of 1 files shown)


bbockelm:
Sure about this?

If `/mnt/ramdisk` fills up, then HTCondor will stop working on those agents.

amaltaro:
That's a valid point, however if that happens, it's probably too late to take such action (JobSubmitter cache usually has plenty of jobs to be submitted, which would actually make it worse).
Besides that, I'd say it's not very uncommon to see /mnt/ramdisk beyond 85%, which makes switching the agent to drain/undrain unwanted. Last but not least, I don't remember seeing ramdisk exceeding 95% of its quota for quite some time (>= 1.5 year)

bbockelm:
Sounds fine.  Just wanted to point out the potential consequence here!

ticoann (approved):

amaltaro:
Seangchan, it looks like we'll have patch(es) to be pushed to the production agents. So I suggest we keep this GH label and add it to a wmagent PR once we backport all of them.

PR #8786, opened 2018-09-04
Added users-wise CRIC APIs for CRAB

### LINKED ISSUE #8783: SiteDB API's used by CRAB
as requested by @amaltaro I list all SiteDB API's used by CRAB.
This should be useful also to @lecriste with reference to https://github.com/dmwm/CRABServer/issues/5705

CRAB always does:
```
from WMCore.Services.SiteDB.SiteDB import SiteDBJSON
```
and  from there it uses:

https://github.com/dmwm/WMCore/blob/6686ebd5fa728839f228575dc08de2d7b385bc2c/src/python/WMCore/Services/SiteDB/SiteDB.py#L107
https://github.com/dmwm/WMCore/blob/6686ebd5fa728839f228575dc08de2d7b385bc2c/src/python/WMCore/Services/SiteDB/SiteDB.py#L118
https://github.com/dmwm/WMCore/blob/6686ebd5fa728839f228575dc08de2d7b385bc2c/src/python/WMCore/Services/SiteDB/SiteDB.py#L178

the above list includes client and server.

Finally CRAB client also makes a direct call to
https://cmsweb.cern.ch/sitedb/data/prod/whoami
which I am going to replace with a call to CRIC  (maybe start with CRIC as fallback)

In the client we decided to use the direct curl to sitedb so that we have a shell command which user can see in the log and try themselves interactively to help them understand what is a CRAB problem and what is not.

Let me know if you need more.

amaltaro:
Fixes #8783

I ran a quick test and I can query every single CRIC api with this code.

A few notes/actions still to be done:
* instead of having gazillions of different methods doing the same thing, I only ported `PNNstoPSNs`. Stefano, you would need to change the CRAB code such that it calls `PNNstoPSNs` instead of `PNNtoPSN`. Is that fine?
* given that we'll have to make slightly changes to othes importing from WMCore. How about we give a more meaningful name to some of those methods (comments in the code).
* I abstracted most of the caching assumptions from SiteDB. The default caching time is set to 30min (in the super class). Do we really need to have a mechanism in this API to enforce fetching fresh data from CRIC?
* TODO: unit tests and mocking data

FYI @belforte @ticoann Seangchan, feel free to push things in into this PR in case it gets merged only after I'm gone for a week. Or just cherry-pick and start from a new PR.

### DIFF
--- a/src/python/WMCore/Services/CRIC/CRIC.py
+++ b/src/python/WMCore/Services/CRIC/CRIC.py
@@ -29,7 +29,7 @@ class CRIC(Service):
 
     def __init__(self, url=None, logger=None):
         params = {}
-        defaultURL = "https://cms-cric-dev-3.cern.ch/api/cms/site/query"
+        defaultURL = "https://cms-cric.cern.ch/"
         params['endpoint'] = url or defaultURL
         params.setdefault('cacheduration', 3600)
         params.setdefault('accept_type', 'application/json')
@@ -38,14 +38,14 @@ class CRIC(Service):
         Service.__init__(self, params)
         self['logger'].info("DEBUG: Initializing CRIC with url: %s", self['endpoint'])
 
-    def _getResult(self, callname="", args=None):
+    def _getResult(self, uri, callname="", args=None, unflatJson=True):
         """
         Either fetch data from the cache file or query the data-service
         :param metricNumber: a number corresponding to the SSB metric
         :return: a dictionary
         """
         cachedApi = "%s.json" % callname
-        apiUrl = '?json&preset=%s' % callname
+        apiUrl = '%s?json&preset=%s' % (uri, callname)
 
         self['logger'].info('DEBUG: Fetching data from %s, with args %s', apiUrl, args)
         # need to make our own encoding, otherwise Requests class screws it up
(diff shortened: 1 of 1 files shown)


belforte:
thanks. quick comments
- fine with me to have PNNstoPSNs only.
- did you test this whoamI ? I'd be surprised that it works given that it fails with a proxy, see https://its.cern.ch/jira/projects/CRIC/issues/CRIC-33?filter=allopenissues)
- I was looking at getAllCMSNames right now (!) and am wondering if we need it all, it may have been some legacy thing before we switched to PNN/PSN

belforte:
About PNN/PSN conversion, I do not know now how it is done in SiteDB Service, but there is no constraint of unique mapping there, differently from what the comments in the source say.

amaltaro:
```
>>> cric.whoAmI()
[{u'dn': u'/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch', u'login': u'wmagent@vocms0192.cern.ch', u'method': u'X509Cert', u'roles': {u'operator': {u'group': [u'dbs'], u'site': [], u'facility': []}, u'production-operator': {u'group': [u'dataops'], u'site': [], u'facility': []}}, u'name': u'vocms0192.cern.ch Service'}]
>>>
```
:-) It works with the service certificate, at least. I didn't check with proxies.

belforte:
yes, I know that it works when presenting PEM files.

belforte:
@ticoann sure you can merge, but we still need a working PNNtoPSN which this does not provide, see https://github.com/dmwm/WMCore/pull/8786#issuecomment-418379613

ticoann:
Sorry Stefano, Since Alan is on vacation, I am planning to work on that in separate patch. I will create the issue.

amaltaro:
> About PNN/PSN conversion but there is no constraint of unique mapping, differently from what the comments in the source say.

@belforte Stefano, can you please elaborate on this comment? As you can see, the function does *not* translate 1 PNN to 1 PSN, it returns a list of PSNs that match that/those PNN(s).

belforte:
Sounds like I misread the code. Will try again asap. What about an example in the first lines? sent from Stefano's mobileOn Sep 13, 2018 10:10, Alan Malta Rodrigues <notifications@github.com> wrote:
About PNN/PSN conversion but there is no constraint of unique mapping, differently from what the comments in the source say.

@belforte Stefano, can you please elaborate on this comment? As you can see, the function does not translate 1 PNN to 1 PSN, it returns a list of PSNs that match that/those PNN(s).

belforte:
OK @amaltaro you are right of course. It may be useful make the description more clear, that's all. Shall I dare a PR ? What I do not like so much is to use the code to filter our T0_CH_CERN_EXPORT and the tape locations, that should be taken care of by simply not indicating any PSN for those. Or are using PNN->PSN to indicate something different ?

amaltaro:
So this is the current docstring "Given a list of PNNs, return all their PSNs". How would you suggest it to be?

I can fix it, because I left 2 TODO comments in the code. Which actually was a question mainly to you and @ticoann . Since we're moving to CRIC, how about taking the opportunity to give a more meaningful name to those APIs? Nobody likes to change a method name all over a project, but... :-)

amaltaro:
About Export, MSS and Buffer, maybe we could make enable or not that via an argument.
Or we make the method such that it assumes whatever you pass as PNN, you want to get the PSN.
I'm open to suggestions here.

belforte:
I would keep the semantic of this at "lookup the map which we have in our database" and if someone does not like the output, either fix the DB (preferred) or add their own filtering.

belforte:
about the function documentation. What about changing
```
:return: a list with unique PSNs matching those PNNs
```
with
```:return: a list with all PSNs matching those PNNs, allowing for multiple PSNs per PNNs
```

and similarly for the sister function

amaltaro:
> I would keep the semantic of this at "lookup the map which we have in our database" and if someone does not like the output, either fix the DB (preferred) or add their own filtering.

Stefano, are you talking about the method `PNNstoPSNs`?
Honestly, I don't like your message :) If it's an interface to the CRIC service, then it doesn't matter if the database is set by CMS or WLCG. The point is, it queries the CRIC `data-processing` API and retrieves a list of `psn_name` given a list of `phedex_name`. 

I'm creating a new PR in a min and we can follow it up from there.

I see you agreed to update the `getallcmsnames` method name, so I'm making those changes too.

amaltaro:
@belforte could you please confirm whether those user-wise APIs are working properly for CRAB services? Just so I can check it in our checklist for the services integration tests.

belforte:
yes, they do. thanks.

PR #8788, opened 2018-09-05
add edmcopyutil fallback for older cmssw releases, fixes #8665

### LINKED ISSUE #8665: Debug LogCollect jobs for Production jobs in CMS@Home
e.g. ireid_MonteCarlo_eff2_IDR_CMS_Home_180626_171641_2516

hufnagel:
fixes #8665
Completely untested, but you get the idea.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -104,8 +104,14 @@ class LogCollect(Executor):
         useEdmCopyUtil = False
         if result:
             try:
-                if int(result.group(1)) >= 8:
+                cmssw_major = int(result.group(1))
+                if cmssw_major >= 8:
                     useEdmCopyUtil = True
+                elif cmssw_major < 8 and scramArch.startswith('slc6_amd64_'):
+                    useEdmCopyUtil = True
+                    logging.warning("CMSSW too old to support edmCopyUtil, using CMSSW_10_2_3 instead")
+                    cmsswVersion = "CMSSW_10_2_3"
+                    scramArch = "slc6_amd64_gcc700"
             except ValueError:
                 pass
 


ticoann:
@hufnagel, this only affects CMSSW version less than 8, right? We have some of the test workflows using CMSSW_7_X_X, I will run the test and merge this.

hufnagel:
Yes, correct, the new code path should not be active for any newer CMSSW releases.

amaltaro:
Hmm, I'm not sure it's a good idea to have the cmssw and scram hardcoded in all these LogCollect jobs.
Eventually it will get deprecated and removed from CVMFS, and when we notice it it will be too late and ALL sandboxes will have to be patched to a proper release.

bbockelm:
If CMSSW_8 is deprecated and removed, then we hopefully won’t have any sandboxes running CMSSW_7!

amaltaro:
How about we use one of the latest 10_x_y stable releases? This 8xy must be almost 2 years old now.
In addition to that, @hufnagel please add a log line such that we know when the LogCollect job had to use the hardcoded release.

hufnagel:
Why ? The edmCopyUtil is supported for CMSSW_8_X and the CMSSW_8_X series will stay around as long or longer than the CMSSW_7_X series. Is there any technical advantage to going to CMSSW_10_X ?

amaltaro:
Ok, I had a quick chat with David L. and, in short, it doesn't really matter since they don't manage to deprecate releases.
However, between picking 800 or a 10x, he would go with 10_2x. Looking at this list for production releases and their production arch:
https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML

how about we use CMSSW_10_2_3 for slc6_amd64_gcc700 ? I see there is a 10_2_5, but it has a patch release, so either we pick that patch1, or we step back a bit and use 10_2_3.

Dirk, don't forget adding that log line. Thanks

hufnagel:
If we should use CMSSW_10_2_X for every CMSSW_7_X release or older, how about we use it for every CMSSW_10_2_X release or older ? I mean, if we override anyways and the newer is better, lets just override for anything older and take advantage of the new xrootd clients. Also makes it easier to test/validate the override code path.

amaltaro:
Hmm, I don't have a strong opinion on that. However, I'd not override it unless we must to.
Otherwise, in 10 years from now we might find this code trying to use 10_2 that is long gone :-)

amaltaro:
I don't really know what you want. I didn't suggest to change anything on your approach but using a newer CMSSW release. Honestly, we're over discussing something that should be much simpler.

hufnagel:
My point is that we use the original CMSSW release for anything CMSSW_8_X or newer. If we say that CMSSW_8_X is no good for overriding older CMSSW, why would it be good for CMSSW_8_X itself ? It's ok or it isn't. There is zero technical reason I've heard so far why CMSSW_10_2_X is better than CMSSW_8_X. If there were, I should override every other pre-CMSSW_10_2_X release as well.

Anyways, all of this doesn't matter one way or another. I am overriding with CMSSW_10_2_3 and adding the log line to be done with this.

amaltaro:
Technical reason 10_2 is better than 8_0 is the whole software package/clients that come with it. Newer xrootd client, newer gcc, newer openssl, and newer release. Why override old releases with a somehow new'ish release if you can have one even newer(?)

hufnagel:
Well then, so the logic should be:

if cmssw < 10_2_3:
  use 10_2_3
else:
  use cmssw

???

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -106,6 +106,11 @@ def execute(self, emulator=None):
             try:
                 if int(result.group(1)) >= 8:
                     useEdmCopyUtil = True
+                elif int(result.group(1)) < 8 and scramArch.startswith('sl6_amd64_'):
+                    useEdmCopyUtil = True
+                    logging.warning("WARNING: CMSSW too old to support edmCopyUtil, using CMSSW_10_2_3 instead")
```
amaltaro: This `WARNING` in the message body can go away. Logging should take care and record the log level altogether.

amaltaro (commented):
Besides the minor log fix, changes look good. Thanks!

amaltaro:
Hmm, according to the job logs, I don't actually see that code being executed... wait, I found a problem in this patch. Commenting in the code.

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -104,8 +104,14 @@ def execute(self, emulator=None):
         useEdmCopyUtil = False
         if result:
             try:
-                if int(result.group(1)) >= 8:
+                cmssw_major = int(result.group(1))
+                if cmssw_major >= 8:
                     useEdmCopyUtil = True
+                elif cmssw_major < 8 and scramArch.startswith('sl6_amd64_'):
```
amaltaro: `slc6_amd64_` instead of `sl6_amd64_`

amaltaro (changes requested):
Dirk, given that LogCollect always have a CMSSW release configured. Should we also rework that `useEdmCopyUtil` flag such that it will always evaluate to True? Thus, ALL logcollect jobs use the edmCopyUtils.

If not, then you just need to fix the scramArch thing mentioned in the code

hufnagel:
By removing the stagein support from LogCollect jobs, we would explicitly say  that we do not support any non-sl6 releases earlier than CMSSW_8_X. Which is a policy decison, although practically speaking I don't think there are such releases. @bbockelm , opinion ?

Once we have that decision, yes, we could clean up the code and remove the whole stagein code path.

hufnagel:
Not sure though if you want to do this in this ticket or open another one...

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
hufnagel: Working on it...

amaltaro:
It works
```
2018-09-21 10:51:20,121:WARNING:LogCollect:CMSSW too old to support edmCopyUtil, using CMSSW_10_2_3 instead
2018-09-21 10:51:20,121:INFO:LogCollect:Running scram
2018-09-21 10:51:24,585:INFO:LogCollect:Running edmCopyUtil
2018-09-21 10:51:24,591:INFO:Scram:    Invoking command: env X509_USER_PROXY=/srv/myproxy.pem edmCopyUtil /store/unmerged/...
```

@ticoann from my side this can go in.

amaltaro (approved):

ticoann:
Thanks, Alan, Dirk. I am merging this since this is only agent side change.

PR #8793, opened 2018-09-09
catch proxy server error wmagent_branch of #8791

ticoann:
1.1.15_wmagent branch of #8791


ticoann:
[Fix #8793 - DBS proxy error handle](https://trello.com/c/yH6vDRG0/126-fix-8793-dbs-proxy-error-handle)

ticoann:
Now it is applied to all agents.

PR #8799, opened 2018-09-13
Changed a couple of CRIC method/API names

amaltaro:
Complement for #8786

As discussed with @belforte , there is no need anymore to filter out those PNNs (Buffer/MSS/Export), since it's now taken care by the service API itself (actually it doesn't even go to the database anymore).

Besides that, I updated the method name `getAllCMSNames` to `getAllPSNs`. See #8809

### DIFF
--- a/src/python/WMCore/Services/CRIC/CRIC.py
+++ b/src/python/WMCore/Services/CRIC/CRIC.py
@@ -63,17 +63,20 @@ class CRIC(Service):
 
     def whoAmI(self):
         """
+        _whoAmI_
+
         Given the authentication mechanism used for this request (x509 so far),
         return information about myself, like DN/ roles/groups, etc
-        :return: FIXME FIXME a list of dictionary?
+        :return: a list of dictionary
         """
         uri = "/api/accounts/user/query/"
         userinfo = self._getResult(uri, callname='whoami', unflatJson=False)
         return userinfo['result']
 
-    # TODO how about renaming it to userNameToDN???
     def userNameDn(self, username):
         """
+        _userNameDn_
+
         Convert CERN Nice username to DN.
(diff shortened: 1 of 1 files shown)


belforte:
the switch back and forth is not clear to me. AFAIU we need to have two versions of the code, one which calls Service/SiteDB, one which calls Service/Cric. So naming for the two have no need to be the same.
Bu maybe I am confused, as usual.

amaltaro:
Stefano, let me update an opened PR on that subject and I'll point you to that code. We have to of course import both modules, but we should have a flag (at config level) that we can enable/disable CRIC usage. Source code is going to be full of IF statements, unfortunately, but we can clean up after ourselves later on.

belforte:
hmm... passing such flags to CRAB REST which runs on CMSWEB may require deeper changes to the code that I am confident with risking. I am starting to think about jumping the gun, make a CRIC  only CRAB and run with it. All in all it is 100 sites, should not be so difficult to make sure that get correct lists, it is not like the 4K users !

belforte:
hmm... passing such flags to CRAB REST which runs on CMSWEB may require deeper changes to the code that I am confident with risking. I am starting to think about jumping the gun, make a CRIC  only CRAB and run with it. All in all it is 100 sites, should not be so difficult to make sure that get correct lists, it is not like the 4K users !

ticoann:
Patch looks good to me and merging this.

ticoann (approved):

ticoann:
I will create the unittest for this module. As far as config and switch backend forth with PhEDEx. It seem there are arround ~20 places in WMCore need to me changed (not including test code) which doesn't seem to be that bad but I think we need to think whether there will be a cleaner solution.

Issue #8805, opened 2018-09-13
Workflows not defaulting to other sites when a site is in drain


jenimal:
Right now we are having massive failures in Workflows due to having to put Rome into drain because of ongoing site issues at Rome.
The site list for these workflows is fairly long, but they are only attempting to run at Rome, and then failing out, why aren't they just running at the other sites on the list?  Please see this set of workflows for an example... (just starting on the list for today)
pdmvserv_task_EXO-RunIISummer15GS-11399__v1_T_180910_180229_3947		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11398__v1_T_180910_180229_4968		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11394__v1_T_180910_180113_5242		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11396__v1_T_180910_180113_5823		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11382__v1_T_180910_175922_6718		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11383__v1_T_180910_175922_9611		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11385__v1_T_180910_175916_1603		100.0%	
pdmvserv_task_EXO-RunIISummer15GS-11391__v1_T_180910_175915_3895		100.0%

amaltaro:
Jen, first problem is related to an issue Seangchan is working on, see
https://github.com/dmwm/WMCore/pull/8806#issuecomment-421210417

which means, unfortunately vocms0250 has MaxRetries=0, instead the usual number of retries.
I just fixed it in vocms0250 and all the other agents up&running.

Second issue is an inconsistent site report in WMStats. The job detail table properly shows T2_IT_Rome (which is correct!), while the `State Transition` list wrongly reports "T3_US_Colorado".

Can you please clarify when Rome was put into drain?

amaltaro:
Answering my own question, it was put into drain around this CERN timestamp:
2018-09-10 22:33:51

and checking the code again, the one-line change Seangchan has in his PR will fix this issue for good, even though it should no longer happen since I updated all agents configs.

jenimal:
https://ggus.eu/?mode=ticket_info&ticket_id=136548  I updated the ggus ticket when I put it into drain so here is the comment
[...]

PR #8806, opened 2018-09-13
don't apply speed mode when user drain mode is false

### LINKED ISSUE #8803: Couple SpeedDrainMode to the other drain flags
Provided the way we deploy agents, where services and components are started but the agent is kept in `UserDrainMode=True`. We should make sure the speed draining settings are couple to these drain flags (AgentDrainMode and UserDrainMode), such that when Unified takes an agent out of drain, the speed drain flagS are set back to False.

ticoann:
fixes #8803 #8805 
Hi Alan, I was wrong, this has the problem when UserDrainMode change to false, and still SpeedDraining config gets to be applied. I manually change the all agents config. Please review.

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
@@ -29,27 +29,25 @@ class DrainStatusPoller(BaseWorkerThread):
         self.config = config
         self.drainAPI = DrainStatusAPI()
         self.condorAPI = PyCondorAPI()
-        self.speedDrainConfig = None
+        self.agentConfig = {}
+        self.validSpeedDrainConfigKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
 
-        if hasattr(config, "Tier0Feeder"):
-            self.reqAuxDB = None
-        else:
-            self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
+        self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
 
     @timeFunction
     def algorithm(self, parameters):
         """
         Update drainStats if agent is in drain mode
         """
+        logging.info("Running agent drain algorithm...")
+        self.agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
+
         if isDrainMode(self.config):
-            logging.info("Checking agent drain status...")
             # check to see if the agent hit any speed drain thresholds
-            if self.reqAuxDB:
-                logging.info("Checking speed drain thresholds...")
-                thresholdsHit = self.checkSpeedDrainThresholds()
-                if thresholdsHit:
-                    logging.info("Updating agent configuration for speed drain...")
-                    self.updateAgentSpeedDrainConfig(thresholdsHit)
+            thresholdsHit = self.checkSpeedDrainThresholds()
+            if thresholdsHit:
+                logging.info("Updating agent configuration for speed drain...")
+                self.updateAgentSpeedDrainConfig(thresholdsHit)
             try:
                 DrainStatusPoller.drainStats = self.drainAPI.collectDrainInfo()
                 logging.info("Finished collecting agent drain status.")
@@ -60,7 +58,8 @@ class DrainStatusPoller(BaseWorkerThread):
                 msg += str(ex)
                 logging.exception(msg)
         else:
-            logging.info("Agent not in drain mode. Skipping drain check...")
+            logging.info("Agent not in drain mode. Resetting flags and skipping drain check...")
+            self.resetAgentSpeedDrainConfig()
 
     @classmethod
     def getDrainInfo(cls):
@@ -74,33 +73,55 @@ class DrainStatusPoller(BaseWorkerThread):
         Takes a list of speed drain configuration keys and updates the agent configuration
         """
         updateConfig = False
+        condorPriorityFlag = False
+        speedDrainConfig = self.agentConfig.get("SpeedDrainConfig")
 
         if 'CondorPriority' in thresholdsHit:
-            if self.condorAPI.editCondorJobs(
-                    "JobStatus=?=1 && (CMS_JobType =?= \"Production\" || CMS_JobType =?= \"Processing\")",
-                    "JobPrio", "999999"):
-                logging.info("Condor job priority updated to 999999 for pending Production/Processing jobs.")
-                if not self.speedDrainConfig['CondorPriority']['Enabled']:
-                    self.speedDrainConfig['CondorPriority']['Enabled'] = True
-                    updateConfig = True
+            logging.info("Bumping condor job priority to 999999 for Production/Processing pending jobs.")
+            self.condorAPI.editCondorJobs(
+                "JobStatus=?=1 && (CMS_JobType =?= \"Production\" || CMS_JobType =?= \"Processing\")",
+                "JobPrio", "999999")
+            condorPriorityFlag = True
+
+        if condorPriorityFlag != speedDrainConfig['CondorPriority']['Enabled']:
+            # CondorPriority setting is irreversible so the flag only indicates weather
+            # priority is increased or not. It is not checked by other components
+            logging.info("Enabling CondorPriority flag.")
+            speedDrainConfig['CondorPriority']['Enabled'] = condorPriorityFlag
+            updateConfig = True
 
         if 'NoJobRetries' in thresholdsHit:
-            if self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "MaxRetries", 0):
-                logging.info("Agent updated max retries to 0.")
-                self.speedDrainConfig['NoJobRetries']['Enabled'] = True
-                updateConfig = True
+            logging.info("Enabling NoJobRetries flag: Error Handler won't retry the jobs")
+            # ErrorHandler will pick this up and set max retries to 0
+            speedDrainConfig['NoJobRetries']['Enabled'] = True
+            updateConfig = True
 
         if 'EnableAllSites' in thresholdsHit:
-            logging.info("Updating agent to submit to all sites.")
+            logging.info("Enabling EnableAllSites flag: Updating agent to submit to all sites.")
             # setting this value to True makes JobSubmitterPoller ignore site status
-            self.speedDrainConfig['EnableAllSites']['Enabled'] = True
+            speedDrainConfig['EnableAllSites']['Enabled'] = True
             updateConfig = True
 
         # update the aux db speed drain config with any changes
         if updateConfig:
             self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainMode", True)
-            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainConfig", self.speedDrainConfig)
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainConfig", speedDrainConfig)
 
+        return
+
+    def resetAgentSpeedDrainConfig(self):
+        """
+        resetting SpeedDrainMode to False and SpeedDrainiConfig Enabled to False
+        """
+
+        if self.agentConfig.get("SpeedDrainMode"):
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainMode", False)
+            speedDrainConfig = self.agentConfig.get("SpeedDrainConfig")
+            for key, v in speedDrainConfig.items():
+                if key in self.validSpeedDrainConfigKeys and v['Enabled']:
+                    speedDrainConfig[key]['Enabled'] = False
+
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainConfig", speedDrainConfig)
         return
 
     def checkSpeedDrainThresholds(self):
@@ -108,11 +129,10 @@ class DrainStatusPoller(BaseWorkerThread):
         Check the current number of jobs in Condor and create a list of agent configuration parameters
         that need updated for speed draining
         """
-        validKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
         enableKeys = []
 
         # get the current speed drain status
-        self.speedDrainConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainConfig")
+        speedDrainConfig = self.agentConfig.get("SpeedDrainConfig")
 
         # get condor jobs
         jobs = self.condorAPI.getCondorJobs("", [])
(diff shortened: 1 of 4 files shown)


src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -103,6 +104,11 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
 
         return
 
+    def resetAgentSpeedDrainConfig(self):
+        if self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainMode"):
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainMode", False)
```
ticoann: I am not resetting individual speed draining config since it will be updated when drain mode change to true.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -751,7 +751,7 @@ def algorithm(self, parameters=None):
             # only runs when reqmgr is used (not Tier0)
             self.removeAbortedForceCompletedWorkflowFromCache()
             agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
-            if agentConfig.get("SpeedDrainConfig"):
+            if agentConfig.get("UserDrainMode") and agentConfig.get("SpeedDrainMode") and agentConfig.get("SpeedDrainConfig"):
```
amaltaro: I think the last statement is not needed.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -103,6 +104,20 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
...
         return
 
+    def resetAgentSpeedDrainConfig(self):
+        """
+        resetting SpeedDrainMode to False and SpeedDrainiConfig Enabled to False
+        :return:
+        """
+        if self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainMode"):
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainMode", False)
+            speedDrainConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("SpeedDrainConfig")
```
amaltaro: We could use this `validKeys` thing that Erik used below just to avoid retyping the parameter name. Or we could set it as a object attr.

amaltaro (commented):
I made a couple of comments, otherwise changes look good in general.
My concern is where this code is located though. DrainStatusPoller has a quite large polling cycle, while JobSubmitter has a 2min polling cycle.

Maybe we should instead change how we set those parameters, like: if there are 0 jobs in the system, then don't bother about the speedDrain parameters.

And this code still needs to go in, I think. But maybe better off with AgentStatusWatcher? which has a 5min polling.

amaltaro (commented):
You're right , Seangchan. It better stay in the DrainStatusPoller thread.
I reviewed your patch again and there is only one line that can be removed, then the patch should be good to go from my side.

About your question about where things are updated, check this method:
https://github.com/dmwm/WMCore/pull/8555/files#diff-7b56e06b637a9d188fc2a84784d0c57fR72

amaltaro:
Seangchan, we also need to update MaxRetries value. So instead of parsing one by one, how about we
`from WMCore.Agent.DefaultConfig import DEFAULT_AGENT_CONFIG`

and either push the default config to reqmgr_aux, or update only SpeedDrainMode, MaxRetries and SpeedDrainConfig?

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -123,11 +140,13 @@ def checkSpeedDrainThresholds(self):
...
             # make sure keys in the speed drain config are valid
-            if k in validKeys and isinstance(v['Threshold'], int) and isinstance(v['Enabled'], bool):
+            if k in self.validSpeedDrainConfigKeys and isinstance(v['Threshold'], int) and isinstance(v['Enabled'], bool):
                 # we always want to apply the condor priority change if the threshold is hit
                 if not v['Enabled'] or k == 'CondorPriority':
                     logging.info("Checking speed drain threshold for %s. ", k)
-                    if len(jobs) < v['Threshold']:
+                    numJobs = len(jobs)
+                    # Don't change if  there is no jobs
+                    if numJobs > 0 and numJobs < v['Threshold']:
```
amaltaro: Sorry Seangchan, I might have missed another use case yesterday. EnableAllSites should be evaluated even when there are no jobs in the condor queue. Reason is, there might be jobs stuck in the JobSubmitter cache waiting for that site to come back. Even though I'm not completely sure that can happen nowadays...

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -103,12 +105,27 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
...
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainConfig", speedDrainConfig)
+        return
+
     def checkSpeedDrainThresholds(self):
         """
         Check the current number of jobs in Condor and create a list of agent configuration parameters
         that need updated for speed draining
         """
-        validKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
+        self.validSpeedDrainConfigKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
```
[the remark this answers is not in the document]
ticoann: Thank Alan, I removed

ticoann:
> About your question about where things are updated, check this method:
> https://github.com/dmwm/WMCore/pull/8555/files#diff-7b56e06b637a9d188fc2a84784d0c57fR72

Ah now I see. It seems that flag is set after the update for condor priority and max retry but for enable all site is opposite way (it set the flag first and job submitter pulls the flag). I think we need to change that to do the same thing as JobSubmitter. I will update the patch.

As far as job priority edit in schedd, I think we can leave as it is (It is irreversible procedure and I think that is OK)

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
ticoann: I don't think we have to worry that much, but since if this happen it can get stuck long time. I will remove that check. We have other protection anyway. Good catch, Alan.

ticoann:
@amaltaro, Alan, I updated the patch, Could you review one more time? Thanks,

ticoann:
test this please

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -74,21 +76,26 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
         Takes a list of speed drain configuration keys and updates the agent configuration
         """
         updateConfig = False
+        condorPriorityFlag = False
 
         if 'CondorPriority' in thresholdsHit:
             if self.condorAPI.editCondorJobs(
```
amaltaro: Do we actually need to check whether the condor_qedit was successful or not? Why don't we just run the edit jobs and set the Enabled flag to True (if it's still False)?

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
amaltaro: In order to avoid one extra reqmgr call, how about we fetch the whole config first and then check for the keys/values?

src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
```diff
@@ -71,7 +71,11 @@ def __init__(self, config):
             self.maxRetries = self.config.ErrorHandler.maxRetries
         else:
             self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
-            self.maxRetries = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("MaxRetries")
+            agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
+            if agentConfig.get("UserDrainMode") and agentConfig.get("SpeedDrainMode"):
```
amaltaro: SpeedDrainMode is set to true when any one of those 3 metrics goes Enabled. So we also need to check for the SpeedDrainConfig/NoJobRetries/Enabled value.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
```diff
@@ -751,7 +751,7 @@ def algorithm(self, parameters=None):
             # only runs when reqmgr is used (not Tier0)
             self.removeAbortedForceCompletedWorkflowFromCache()
             agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
-            if agentConfig.get("SpeedDrainConfig"):
+            if agentConfig.get("UserDrainMode") and agentConfig.get("SpeedDrainMode"):
```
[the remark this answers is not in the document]
amaltaro: And actually the same comment I made for ErrorHandler. Check SpeedDrainConfig

amaltaro (changes requested):
There are more changes needed, Seangchan. Unless we redefine SpeedDrainMode to be True only when all thresholds are hit.

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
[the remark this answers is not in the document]
ticoann: This one already checks

ticoann:
Thanks Alan, I updated what you suggested.

amaltaro (changes requested):
Either use `self.agentConfig` or `self.speedDrainConfig`

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -74,21 +75,26 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
...
                     "JobPrio", "999999"):
                 logging.info("Condor job priority updated to 999999 for pending Production/Processing jobs.")
-                if not self.speedDrainConfig['CondorPriority']['Enabled']:
-                    self.speedDrainConfig['CondorPriority']['Enabled'] = True
-                    updateConfig = True
+            condorPriorityFlag = True
+
+        if condorPriorityFlag != self.speedDrainConfig['CondorPriority']['Enabled']:
+            # CondorPriory setting is irreversible so the flog only indicated weather priorty is increased or not.
+            # this flag is not checked by other component
```
amaltaro: And how about logging something like 'Enabling CondorPriority bump'

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -74,21 +75,26 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
...
+            self.speedDrainConfig['CondorPriority']['Enabled'] = condorPriorityFlag
+            updateConfig = True
 
         if 'NoJobRetries' in thresholdsHit:
-            if self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "MaxRetries", 0):
-                logging.info("Agent updated max retries to 0.")
-                self.speedDrainConfig['NoJobRetries']['Enabled'] = True
-                updateConfig = True
+            logging.info("Agent updated max retries to 0.")
```
amaltaro: And how about logging something like 'Enabling NoJobRetries flag'

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -74,21 +75,26 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
...
+            # ErrorHandler will pick this up and set max retries to 0
+            self.speedDrainConfig['NoJobRetries']['Enabled'] = True
+            updateConfig = True
 
         if 'EnableAllSites' in thresholdsHit:
             logging.info("Updating agent to submit to all sites.")
```
amaltaro: And how about logging something like 'Enabling EnableAllSites flag'

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -103,16 +109,30 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
 
         return
 
+    def resetAgentSpeedDrainConfig(self):
+        """
+        resetting SpeedDrainMode to False and SpeedDrainiConfig Enabled to False
+        :return:
```
amaltaro: remove this line

amaltaro (commented):
and now I made further comments :-)

ticoann:
@amaltaro, Alan, patch is updated. Hopefully I got all things you pointed.

ticoann:
test this please

amaltaro (commented):
Thanks, Seangchan. Code looks good now, besides one more typo in the log (which I took the opportunity to suggest something slightly better :)).

I think we can update it on this master patch only though, and let the wmagent branch as is.

src/python/WMComponent/ErrorHandler/ErrorHandlerPoller.py
```diff
@@ -71,7 +71,12 @@ def __init__(self, config):
             self.maxRetries = self.config.ErrorHandler.maxRetries
         else:
             self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
-            self.maxRetries = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName).get("MaxRetries")
+            agentConfig = self.reqAuxDB.getWMAgentConfig(self.config.Agent.hostName)
+            if (agentConfig.get("UserDrainMode") and agentConfig.get("SpeedDrainMode") and
+                agentConfig.get("SpeedDrainConfig")['NoJobRetries']['Enabled']):
```
amaltaro: Seangchan, can you also log that the component is using the MaxRetries=0 here too?

src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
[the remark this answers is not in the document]
amaltaro: Nevermind, I see we already have a log at L282 here

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -74,45 +75,65 @@ def updateAgentSpeedDrainConfig(self, thresholdsHit):
...
+    def resetAgentSpeedDrainConfig(self):
+        """
+        resetting SpeedDrainMode to False and SpeedDrainiConfig Enabled to False
+        """
+
+        if self.agentConfig.get("SpeedDrainMode"):
+            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, "SpeedDrainMode", False)
+            speedDrainConfig = self.agentConfig.get("SpeedDrainConfig")
+            for key, v in speedDrainConfig.items():
+                if key in self.validSpeedDrainConfigKey and v['Enabled']:
```
amaltaro: Haaa, there is a typo here. It should be `validSpeedDrainConfigKeys`

amaltaro (changes requested):
You've got more work to make in here :)

amaltaro:
@ticoann I'm not a big fan of it, but since I'd like to cut a patch5 release today. I pushed a commit into your branch with the suggestions I had previously made.
Can you please review them and squash both commits, if you're happy?

I'm going to cherry-pick this 2nd commit and push it on #8815, run another quick test and merge it if everything is ok. Thanks

amaltaro:
I squashed two more changes into the 2nd commit, hopefully that will be all.

amaltaro:
Fixed another issue with JobSubmitterPoller. We should be settle now, time to cherry-pick it and then you can squash them Seangchan.

ticoann:
Thanks Alan, I squashed it and merging it.

PR #8807, opened 2018-09-14
 Try creating reqmgr_aux agent doc once again

### LINKED ISSUE #8804: wmagent-upload-config script should overwrite the current document
Instead of delete and put a new document, we should simply override the whole document in the database (if it exists). Otherwise, from time to time I see a couch race condition between the delete and put and we end up without any document in the reqmgr_aux db.

amaltaro:
Fixes #8804

### DIFF
--- a/bin/wmagent-upload-config
+++ b/bin/wmagent-upload-config
@@ -26,3 +26,9 @@ if __name__ == "__main__":
 
     removeWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName)
     insertWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName, DEFAULT_AGENT_CONFIG)
+
+    # check whether the document was properly inserted (for some reason,
+    # it doesn't get from time to time
+    if not reqMgrAux.getWMAgentConfig(wmConfig.Agent.hostName):
+        print("Agent config not found, trying to insert it again...")
+        insertWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName, DEFAULT_AGENT_CONFIG)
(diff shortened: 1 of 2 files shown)


amaltaro:
I can't really find a nice fix for this problem. And honestly, I don't quite understand what happens with these wmagent_config docs. For instance, the CMSSW_VERSIONS doc is *posted* every cycle, and a new doc is created and contains rev 1.

On the other hand, *posting* a wmagent_config doc (that has been previously deleted) creates a new doc but it contains the previous rev++. AND, if there is already a doc in the database, it fails to update it with a Doc conflict message.

@ticoann can you please have a look at this code. I'm certainly missing something. They both inherit from the same base class, both call the post method, still we have different behaviour :(

bin/wmagent-upload-config
```diff
@@ -26,3 +26,6 @@ if __name__ == "__main__":
 
     removeWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName)
     insertWMAgentConfig(reqMgrAux, wmConfig.Agent.hostName, DEFAULT_AGENT_CONFIG)
+    # for some reason, sometimes the document doesn't get created
+    print("Trying to insert it again...")
```
ticoann: Instead trying again maybe it checks document is inserted, if not insert again.
Anyway, Alan, I will create the new patch, since you have a lot things to do.

amaltaro:
Thanks, Seangchan. That's actually a pretty simple and good idea :-)
I updated the patch and will test it soon.

amaltaro:
@ticoann from my side it's ready to go

ticoann:
test this please

amaltaro:
Issue 8813 has been created for this unit test that is failing.

ticoann (approved):

PR #8808, opened 2018-09-14
Define PU dataset name in WMTask and in the job classad

### LINKED ISSUE #8801: need pileupDataset classad in job
Similar to how the input dataset makes it into the job classads

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L574

we also need information about the pileup dataset. The inability to select on this for job routing purposes hurts us at HPC.

amaltaro:
Fixes #8801

Changes are:
 * assumes each task can have multiple PU samples and multiple PU types
 * persist the PU name(s) under a WMTask object (and in the workload object)
 * pass the PU name(s) all the way from JobCreator to JobSubmitter and the submitter plugin
 * define a new job classad `DESIRED_CMSPileups` with the PU name(s) for that job/task. Leave it undefined if there is no PU dataset.

Note this will only be visible in the job classads for workflows created after this change goes to ReqMgr2.

### DIFF
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -119,7 +119,8 @@ def capResourceEstimates(jobGroups, constraints):
 def saveJob(job, workflow, sandbox, wmTask=None, jobNumber=0,
             owner=None, ownerDN=None, ownerGroup='', ownerRole='',
             scramArch=None, swVersion=None, agentNumber=0, numberOfCores=1,
-            inputDataset=None, inputDatasetLocations=None, allowOpportunistic=False, agentName=''):
+            inputDataset=None, inputDatasetLocations=None, inputPileup=None,
+            allowOpportunistic=False, agentName=''):
     """
     _saveJob_
 
@@ -147,11 +148,11 @@ def saveJob(job, workflow, sandbox, wmTask=None, jobNumber=0,
     job['numberOfCores'] = numberOfCores
     job['inputDataset'] = inputDataset
     job['inputDatasetLocations'] = inputDatasetLocations
+    job['inputPileup'] = inputPileup
     job['allowOpportunistic'] = allowOpportunistic
 
-    output = open(os.path.join(cacheDir, 'job.pkl'), 'w')
-    pickle.dump(job, output, pickle.HIGHEST_PROTOCOL)
-    output.close()
+    with open(os.path.join(cacheDir, 'job.pkl'), 'w') as output:
+        pickle.dump(job, output, pickle.HIGHEST_PROTOCOL)
 
     return
 
@@ -180,6 +181,7 @@ def creatorProcess(work, jobCacheDir):
         numberOfCores = work.get('numberOfCores', 1)
         inputDataset = work.get('inputDataset', None)
         inputDatasetLocations = work.get('inputDatasetLocations', None)
+        inputPileup = work.get('inputPileup', None)
         allowOpportunistic = work.get('allowOpportunistic', False)
         agentName = work.get('agentName', '')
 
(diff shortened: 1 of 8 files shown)


hufnagel:
Not sure about list vs. string. We already know from the feedback we got recently that we can have workflows that use multiple pileup datasets (comparing classical and premixed pileup for validation purposes for instance) and if we presume that eventually these might all run as StepChain from a single task, then a list would be needed.

hufnagel:
It's really not more complicated to have a list, is it ? Does it complicate the lookup in ElasticSearch to have a list parameter ? Or do we need to encode it ourselves as a string with multiple elements ?

@bbockelm ?

amaltaro:
This change does support multiple PU samples in a workflow basis, but not in the same task.
Question now is, can we (or do we want to support) multiple PU datasets in the same StepChain request (StepChain is a single BIG task)?

Before making any further changes, I'll try to come up with a test workflow just to see whether multiple PUs would work all the way from request creation to job execution.

bbockelm:
Hi @hufnagel -

List-vs-string doesn't matter too much to ElasticSearch as we have a script that converts from ClassAds to JSON.

List might be slightly easier in that we may not have to tweak the script at all.

ticoann:
As we discussed in the meeting, since it supports multiple pile up in a workflow, although it is never used as such, we will try to use list unless it requires a lot of code change.

hufnagel (commented):
Looks good. Didn't check every single line of code for typos etc, that'll come out in testing. The method how this was added in looks good though, it's how I would have done it (and how it was done before for other parameters like the allowOpportunistic).

amaltaro:
Ok, I see the classad being set but with undefined value. I might have missed something in JobCreator that would propagate it to JobSubmitter :( I'll investigate it tomorrow.

amaltaro:
Bloody RunJob object... it's fixed now and this is what we get in the job classad:
```
cmst1@vocms0262:/data/srv/wmagent/current $ condor_q -l 1847.32 | grep CMSPileup
DESIRED_CMSPileups = "/MinBias_TuneCUETP8M1_13TeV-pythia8/RunIIWinter15GS-MCRUN2_71_V1-v1/GEN-SIM,/Neutrino_E-10_gun/RunIISpring15PrePremix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v2-v2/GEN-SIM-DIGI-RAW"
```
is it ok?

hufnagel:
Perfect. If there were multiple it would just follow comma separated?

amaltaro:
Yes, the example above has 2 PU samples :)

hufnagel:
Ah, haha, completely missed that ! Perfect then.

hufnagel:
@bbockelm , this will play well with ElasticSearch then, right ?

bbockelm:
Yes, that should work well with ElasticSearch.  Please keep this ticket open until we've confirmed data looks OK there.

As soon as this goes into a test schedd on the Global Pool, we should have records we can examine.

amaltaro:
@bbockelm we usually use ITB schedds for testing these PRs, so it won't show up in ES.
The only testbed schedds connected to the global pool are part of the integration setup.

I'd suggest to get it in (we need to get it in in ReqMgr2 asap), we proceed with the integration tests in global pool, if we have issues, then we can create another patch to fix the data format (which would be only on the plugin side anyways).

amaltaro:
@ticoann from my side it's ready to go.

ticoann:
test this please

ticoann (approved):

ticoann:
Thanks Alan, patch looks good to me.

amaltaro:
FYI., this will only be available for workflows created after it goes into production. I also don't foresee backporting it to the current WMAgent, so we'll have to wait for a new agent release.

PR #8811, opened 2018-09-17
add job type info in the log

ticoann:
add job type on the JobSubmitter report.

Will print something like below. (It will have long line, so maybe we can make this in debugging mode, if you prefer.

"T1_US_FNAL": {
        "Processing": {
            "NoTaskPendingSlot": 113
        }, 
        "Cleanup": {
            "NoTaskRunningSlot": 3780, 
            "submitted": 4
        }
    }

"NoTaskPendingSlot": 113 - This mean there 113 couldn't be submitted due to no pending slot available

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -12,6 +12,7 @@ from __future__ import print_function, division
 import logging
 import os.path
 import threading
+import json
 from collections import defaultdict, Counter
 try:
     import cPickle as pickle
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
Patch looks good, Seangchan.
I'd use `from pprint import pformat` instead of json, I think it's slightly cleaner, but that's really just a diff taste :P

ticoann:
Thanks Alan,

ticoann:
Hi I checked but pformat still prints defaultdict
>>> pformat(a, 1)
"defaultdict(<type 'list'>, {'r': [1]})"

PR #8814, opened 2018-09-18
make crabtaskworker depend on all services, so it gets the new CRIC one

belforte:
@amaltaro @ticoann we need this in order to have CRIC in the rpm when we build CRAB TaskWorker. Can you please review, merge, and include in next tag ?


ticoann:
test this please

ticoann:
Thank Stefano. patch looks good to me. We will create a new tag probably on Thursday.

PR #8817, opened 2018-09-19
CRAB Client does not need SiteDB Service

belforte:
not urgent !!


ticoann:
Thanks Stefano, I will include in this tag.

amaltaro:
@belforte btw, this and the other TW PR were added to 1.1.17.pre2

PR #8824, opened 2018-09-21
Add username parameter

lecriste:
Add `username` parameter in order to make use of it here:
https://github.com/dmwm/CRABServer/blob/631c55d6cc9f5aae1c73085b1e38051d00a4494c/src/python/UserFileCache/RESTFile.py#L101


ticoann:
@lecriste, Thanks, Leonardo, do you need new crab branch with this patch?

ticoann:
@lecriste, How urgent do you need a tag. we are running the validation test with current tag which will go to next Tuesday, do you need a tag before that?

lecriste:
@ticoann, yes I would like to have a tag before next Tuesday if possible.

amaltaro:
@lecriste Leonardo, we've cut a 1.1.17.pre3 tag.

PR #8835, opened 2018-09-28
Add files to ACDC docs even if they have no location

### LINKED ISSUE #8832: ACDC with files = {}
https://cmsweb.cern.ch/couchdb/acdcserver/_design/ACDC/_view/byCollectionName?key=%22vlimant_ACDC0_task_JME-RunIIFall17DRPremix-00009__v1_T_180807_034901_4782%22&include_docs=true&reduce=false

has 

"doc":{"_id":"4efc3f02203a914a9cdc078e59a2f3cb","_rev":"1-6d49de84c2810c5cc4ae48cae383c82e","files":{},"collection_name":"vlimant_ACDC0_task_JME-RunIIFall17DRPremix-00009__v1_T_180807_034901_4782","collection_type":"ACDC.CollectionTypes.DataCollection","fileset_name":"/vlimant_ACDC0_task_JME-RunIIFall17DRPremix-00009__v1_T_180807_034901_4782/JME-RunIIFall17DRPremix-00009_1","timestamp":1533638961.6794159412}}

the files = {} is rather worrisome. The task that failed needs input, has an acdc, but no files to be used, and hence no location to be used ...

amaltaro:
Fixes #8832

Changes are:
* log warning the job id when it cannot be submitted.
* make sure the job report contains an error for any of those submitFailed cases
* return the file information to ErrorHandler even if the file doesn't have any correspondent row in wmbs_file_location table

See further findings in the issue itself, but I think the whole problem comes from the RAL PNN changes (renaming).

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -328,6 +328,7 @@ class JobSubmitterPoller(BaseWorkerThread):
                 newJob['fileLocations'] = loadedJob.get('fileLocations', [])
                 newJob['siteWhitelist'] = loadedJob.get('siteWhitelist', [])
                 newJob['siteBlacklist'] = loadedJob.get('siteBlacklist', [])
+                logging.warning("Input data location doesn't pass the site restrictions for job id: %s", jobID)
                 badJobs[71101].append(newJob)
                 continue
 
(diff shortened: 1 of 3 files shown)


ticoann:
Hi Alan, I am not sure about the problem. Does this problematic file doesn't have the location? Is the input file from dbs (which should have location in PhEDEx) or input file generated intermediately by workflow? 
I think both cases shouldn't happen and if it did we have some other problem. Especially first case it shouldn't happen at all and we should not process further. For the second case, I think we could do do some protection like this patch, but I am not sure whether that can be solved. If the unmerged file is input with no location, does trust site doesn't set to true, right?

amaltaro:
Ok, @ticoann . My tests look good, feel free to review it again and merge.

ticoann (approved):

Issue #8827, opened 2018-09-26
AgentJobInfo not in wmstats info for completed workflow (When Agent is drained)


vlimant:
unified used to get AgentJobInfo from https://cmsweb.cern.ch/wmstatsserver/data/request/pdmvserv_task_HIG-RunIIFall17DRPremix-00809__v1_T_180513_171003_961 , but it is not there anymore for some reason. glitch on that one, or a recent change ?

amaltaro:
If there is no AgentJobInfo information in wmstats, that's because the workflow ran in an agent(s) that has been already shutdown/redeployed. Given that wmstats doesn't keep data for more than ... 0.25 days AFAIR, and the agent that handled it is gone from the system, we do not have this information anymore.

vlimant:
BTW https://cmsweb.cern.ch/wmstatsserver/data/jobdetail/pdmvserv_task_HIG-RunIIFall17DRPremix-00809__v1_T_180513_171003_961 seem to still have lots of useful information as can be seen below

![screenshot from 2018-09-26 15-23-21](https://user-images.githubusercontent.com/4989875/46083167-15891180-c1a1-11e8-9c87-d56a21ac4e29.png)

amaltaro:
Just to make sure everyone is aware of it. This is a BIG!!! problem for agent recycling and deployments. In practice - with the current behavior of workflows and P&R Ops - it means it will take half a year to manage to get an agent redeployed. Which will cause all sort of other issues as data volume, agent maintenance, patching, etc etc etc...

This means we need to work on wmstats and increase the data retention for a reasonable amount of time (days? weeks?) and assume that P&R will take the necessary actions within that time window.

PS.: It has data now because I restarted submit4 (which is completely drained but not wiped out yet), which was handling this workflow.

amaltaro:
BTW, that workflow has been sitting in completed for 37 days without any action. We cannot keep data forever.

ticoann:
> config.TaskArchiver.DataKeepDays = 0.125

This is duration for keeping old job summary doc (which is the same as AgentJobInfo). Job Summary information is updated every 10 min, so it keeps the 3 hour history. 

Job record from given given workflow won't be deleted until workflow is archived. That is why you can recover when you revive the agent.

What we probably could do is changing the deleting algorithm. (Keep the summary record for completed workflow even though it is more than 3 hours passed). I will look at the code what need to be done to keep this when agent is drained.

ticoann:
[...]

PR #8836, opened 2018-09-28
keep only one job summary docs per agent and request

ticoann:
related to #8827 (The code is just skeleton, need to be tested and refined)

This is agent part of the patch, I will create separate patch for wmstats serverside.

1. instead of adding new document for workflow job summary we will update the old document.
  (id is set to [agent-name]-[request-name])
2. This has to be backward compatible.  (Still agent will delete the document every 3 hours from TaskArchiver - new agent doesn't have to do that but if this is patched to current agent it still need that feature) 
3. With new id, central wmstats view still need to work.
4. Agent patch has to be applied before wmstats are updated. (After agents are updated we have to rotate wmstats to remove all the previous job summary document from wmstats)

### DIFF
--- a/src/couchapps/WMStatsAgent/views/agentRequests/map.js
+++ b/src/couchapps/WMStatsAgent/views/agentRequests/map.js
@@ -1,0 +1,8 @@
+function(doc) {
+  if (doc.type === "agent_request" && doc.agent_url) {
+    // this condition check can be removed when there is no historical data is left. Not sure why this check is not working
+    //if (doc['_id'].startsWith(doc.agent_url)) {
+        emit(doc['_id'],  {'rev': doc['_rev']});
+    //}
+  }
+}
--- a/src/python/WMCore/Services/WMStats/WMStatsWriter.py
+++ b/src/python/WMCore/Services/WMStats/WMStatsWriter.py
@@ -55,6 +55,7 @@ def convertToServiceCouchDoc(wqInfo, wqURL):
 
 
 class WMStatsWriter(WMStatsReader):
+
     def __init__(self, couchURL, appName="WMStats", reqdbURL=None, reqdbCouchApp="ReqMgr"):
         # set the connection for local couchDB call
         WMStatsReader.__init__(self, couchURL, appName, reqdbURL, reqdbCouchApp)
@@ -76,6 +77,28 @@ class WMStatsWriter(WMStatsReader):
         for doc in docs:
             self.couchDB.queue(doc)
         return self.couchDB.commit(returndocs=True)
+
+    def bulkUpdateData(self, docs, existingDocs):
+        """
+        :param docs: docs to insert or update
+        :param existingDocs: docs existing in current
+        :return:
+        """
+        if isinstance(docs, dict):
+            docs = [docs]
+        for doc in docs:
+            if doc['_id'] in existingDocs:
+                revList = existingDocs[doc['_id']].split('-')
+                # update the revision number
+                doc['_rev'] = "%s-%s" % (int(revList[0]) + 1, revList[1])
+            else:
+                # just send well formatted revision for the new documents which required by new_edits=False
+                doc['_rev'] = "1-123456789"
+            self.couchDB.queue(doc)
+
+        self.couchDB.commit(new_edits=False)
+        return
+
 
     def insertRequest(self, schema):
         doc = monitorDocFromRequestSchema(schema)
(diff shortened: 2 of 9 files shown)


ticoann:
@amaltaro Alan, I test this. But still need to test various scenarios. When it was patched current running agent. First commit need to be applied to agent, Second commit is on the wmstats serverside.
(Actually view doesn't have to be changed - leave reduce as true) It might be a little cleaner as patch but for the compatibility I can revert back as before only remove the cleaning couch doc code.

ticoann:
test this please

src/couchapps/WMStatsAgent/views/agentRequests/map.js
```diff
@@ -0,0 +1,8 @@
+function(doc) {
+  if (doc.type === "agent_request" && doc.agent_url) {
+    // this condition check can be removed when there is no historical data is left. Not sure why this check is not working
+    //if (doc['_id'].startsWith(doc.agent_url)) {
+        emit(doc['_id'],  {'rev': doc['_rev']});
```
amaltaro: Can't we simply emit `doc['_rev']` as value? instead of emitting a key/value dict.

Second question is regarding the database we store these docs on the agent side. What other types of documents are stored there?

src/python/WMCore/Services/WMStats/WMStatsWriter.py
```diff
@@ -77,6 +78,27 @@ def uploadData(self, docs):
...
+        """
+        :param docs: docs to insert or update
+        :param existingDocs: docs existing in current
+        :return:
+        """
+        if isinstance(docs, dict):
+            docs = [docs]
+        for doc in docs:
+            if doc['_id'] in existingDocs:
+                revList = existingDocs[doc['_id']].split('-')
```
amaltaro: I must be missing something. When we want to update a document in couch, don't we post the data we want (including the current revision number) and couch updates the document and bump the revision number?

src/python/WMCore/Services/WMStats/WMStatsWriter.py
```diff
@@ -77,6 +78,27 @@ def uploadData(self, docs):
...
+                # update the revision number
+                doc['_rev'] = "%s-%s" % (int(revList[0]) + 1, revList[1])
+            else:
+                # just send well formatted revision for the new documents which required by new_edits=False
+                doc['_rev'] = "1-123456789"
```
amaltaro: Same question here. If document doesn't exist, don't we just post it and let the revision to magically appear?

src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
```diff
@@ -20,17 +20,7 @@ def setConcurrentTasks(self, config):
         """
         sets the list of functions which runs concurrently
         """
-        self.concurrentTasks = [{'func': self.cleanUpOldRequests, 'duration': (config.DataKeepDays * 24 * 60 * 60)},
-                                {'func': self.cleanUpArchivedRequests, 'duration': config.archivedCleanUpDuration}]
```
amaltaro: Don't you want to keep this second line?

amaltaro (changes requested):
I made several comments in the code, mostly questions though to help me understand something that I didn't know about couch. One thing needs to be updated though.

src/couchapps/WMStatsAgent/views/agentRequests/map.js
ticoann: Yes we can, I think it is clearer and more flexible if we want to add some other values, but if you prefer we can change to just emit rev number.

Here is docs (view) stored in agents.
https://github.com/dmwm/WMCore/tree/master/src/couchapps/WMStatsAgent/views

src/python/WMCore/Services/WMStats/WMStatsWriter.py
ticoann: Yes it does but with new_edits=False it overwrites it doesn't do that. And that is needed for the bulk update otherwise we have update document one by one. Only highest revision number is showing when document is retrieved without revision number. That is the reason code gets the current revision number and increase by one.

src/python/WMCore/Services/WMStats/WMStatsWriter.py
ticoann: The same answer. We can separate the documents with initial inserting and updating then use new_edits=True, new_edits=False respectively but I don't see any advantages doing that.

src/python/WMCore/WMStats/CherryPyThreads/CleanUpTask.py
ticoann: Actually I didn't remove the second line.
https://github.com/dmwm/WMCore/pull/8836/files#diff-a83e41bb095c827fc02a0bc6f4694b4cR23

PR #8838, opened 2018-10-02
Add required files for agent deployment; improved initial node setup

### LINKED ISSUE #8792: Add Sanity check and more automation on production agent deployment.
https://github.com/dmwm/WMCore/wiki/WMAgent-deployment

1. check whether agent is already registered in wmstats. 
   - if true, check the drain status, if false proceed to next step
      - if drain is not completed abort the procedure
      - else proceed to next step.

2. automatically check schedd is installed and connected to pool, if not abort

3. automatically create the secrete file from template. 
     - set the flavor according to host name (vocmsbox oracle, fermi agent maria db)  - you have option to overwrite this.
     - Fill up could host with host ip address automatically
      - User types user and password  for database. (We can add the validation for db connection or whether it can be wiped out) - This is most critical case if we made a mistake.

4. automatically set the cert and proxy (We need to improve this part)

5. automatically deploy the agent

amaltaro:
Fixes #8792 

Changes are:
* created a `deploy` directory for deployment-related scripts
* moved the restartComponent.sh script over from my repo (hopefully temporary though)
* created a separate script to clean up oracle (instead of during the deployment) before a new agent deployment, to be used with `$manage execute-agent` (planning to execute that when shutting down an agent for good)
* created a prod and a preprod template for the agent secrets file
* created a template (actually I don't think there is anything to change in there) for the env.sh script, with CRIC enabled by default

Changes to the deployment script are:
* moved it under the deploy directory too
* run the basic directories/files/permissions bootstrap, creating then and downloading templates when ran in a new node. One still need to properly set the secrets file with the credentials.
* run a bunch of checks, like
  * copy the certificate and proxy from another (hard-wired) node, if they are not in place
  * check if there are any wmcore/couch/mysql process running
  * if oracle, check whether the schema doesn't have any tables
  * use CRIC if it's set in the env.sh

TODO: create/update wiki with agent redeployment instructions.
TODO: backport only the clean-oracle executable to the agent branch


amaltaro:
Ok, I caught several problems while testing and it should be good to go now.

ticoann (approved):

ticoann:
Alan, patch looks good. One additional check we might consider is check whether agent is exist in wmstats and if agent is not in completed drained status abort the deployment.

amaltaro:
I think we need to actually make it part of the decommissioning of an old agent. Otherwise it will just complicate things when we move to pip/docker installations.

PR #8842, opened 2018-10-04
Properly check whether Services cache is expired or not

### LINKED ISSUE #8841: WMCore Services cache does not work
Re-purposing this issue, original problem reported in this issue is here [1]

What's left to sort out in WMCore is:
*) the Services cache is always considered as expired
**) at least for CRIC, where we define a cache file on the CRIC module, every http request ends up creating 2 cache files, one as defined in the subclass/Services, the second is likely internally defined in the Requests parent class.

Example when querying cric interactively in python, these are the cache files:
/tmp/.wmcore_cache_EmdSXn/cms-cric.cern.ch/15744094587142094_GET_data-processing.json
against (which has some header information as well)
/tmp/.wmcore_cache_EmdSXn/cms-cric.cern.ch/.cache/cms-cric.cern.ch\,api\,cms\,site\,query\,json\&preset\=data-processing\&rcsite_state\=ANY\,abfa15f4a7bc5291c02802dceb166bc7 

given that we can't easily read from the second cache file, I think we should actually not create it at all.

[1]
Creating this GH issue for a better tracking. Some of the comments from the private thread are posted below, from Todor
```
We think we took the problem to a more concrete question here. We noticed that one of our modules (namely DataDiscovery.py)  is calling a CRIC function like that [1] (SiteDB was using this call [2] instead). This module stays in that step for too long time while reading the WmcoreCache file returned by the CRIC instantiation and the output in the cache is enormous, see our previous observations described here [3].
Is there any crucial difference in how those modules (CRIC and SITedb) use the Wmcore Cache?
[1]
https://github.com/dmwm/CRABServer/compare/3.3.1809.rc8...3.3.1810.rc4#diff-c51fab6758de074f1346c596adc0a1d8R69
[2]
https://github.com/dmwm/CRABServer/compare/3.3.1809.rc8...3.3.1810.rc4#diff-c51fab6758de074f1346c596adc0a1d8L75
[3]
https://cms-logbook.cern.ch/elog/Analysis+Operations/2297
```

from Leonardo
```
please find the two cache files here (look for the hidden files):
/afs/cern.ch/user/l/lecriste/public/forAlan/
We also noticed that the clearCache argument is missing in the CRIC module... not sure if it may be related. 
```

amaltaro:
Fixes #8841

Actually it fixes the first issue, which is now correctly checking whether the (Services) cache file is expired or not, based on the `cacheduration`. Before this patch, cache was always considered as expired.

In short, the new behavior is:
* if the cache file is a file-like object, assume it's expired
* if file does not exist, assume it's expired
* otherwise, check if last modification of the file is greater than `cacheduration`, in hours
* (this is not changed, but...) the cache file is only updated if there is an actual query to the data-service
* if cache is dead/expired and the actual query failed, it raises an exception (no matter the `usestalecache` flag value)

### DIFF
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -220,7 +220,7 @@ class Requests(dict):
         if verb != 'GET' and data:
             if isinstance(encoder, (types.MethodType, types.FunctionType)):
                 encoded_data = encoder(data)
-            elif encoder == False:
+            elif encoder is False:
                 # Don't encode the data more than we have to
                 #  we don't want to URL encode the data blindly,
                 #  that breaks POSTing attachments... ConfigCache_t
@@ -417,7 +417,7 @@ class Requests(dict):
         # Set but not found
         if key is None or cert is None:
             raise WMException('Request requires a host certificate and key',
-                                  "WMCORE-11")
+                              "WMCORE-11")
 
         # All looks OK, still doesn't guarantee proxy's validity etc.
         return key, cert
--- a/src/python/WMCore/Services/Service.py
+++ b/src/python/WMCore/Services/Service.py
@@ -25,14 +25,14 @@ Calling refreshCache/forceRefresh will return an open file object, the cache
 file. Once done with it you should close the object.
 
 The service has a default timeout to receive a response from the remote service
-of 30 seconds. Over ride this by passing in a timeout via the configuration
+of 300 seconds. Over ride this by passing in a timeout via the configuration
 dict, set to None if you want to turn off the timeout.
 
 If you just want to retrieve the data without caching use the Requests class
 directly.
 
 The Service class provides two layers of caching:
-    1. Caching from httplib2 is provided via Request, this respectsetag and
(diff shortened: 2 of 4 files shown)


amaltaro:
FYI @hufnagel @yuyiguo @mmascher @titask @todor-ivanov @belforte 

In case you have additional service client APIs inheriting from this Services class. You might want to adapt `cacheduration` according to your needs. Full description is in the first comment (I still have to test it though).

amaltaro:
test this please

hufnagel:
Looked, but couldn't find me using this in the Tier0.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

ticoann (approved):

ticoann:
Hi Alan, Sorry this has conflict after I merged CRIC patch. Could you fix the conflict? Thanks,

amaltaro:
Done! Thanks, SEangchan.

amaltaro:
test this please

PR #8843, opened 2018-10-04
1.1.14_wmagent branch: keep only one job summary docs per agent and request

ticoann:
1.1.14_wmagent branch of #8836


ticoann:
Applied to production agents

PR #8844, opened 2018-10-08
Catch other failures in the DQMUpload step

### LINKED ISSUE #8839: Question about DQMHarvest jobs
Can DQMHarvest jobs fail in the upload to the GUI, but still be considered successful?

Because Data certification team has found that for  2 collision runs 320824 and 321781, some DQM files were not successfully transferred from T0 to DQM GUI. Basically these PDs

1) JetHT PD  for 321781
2) ZeroBias PD for 320824

It was found by Marcel that these jobs ended during the CMSWEB update downtime, one on 2018-08-07 and one on 2018-09-04,  and DQMGUI was probably down during that period which lead to this unsuccessful transfer.

amaltaro:
Fixes #8839

Catch silent errors that don't come from an http upload exception.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
@@ -7,9 +7,9 @@ Implementation of an Executor for a DQMUpload step
 """
 from __future__ import print_function
 
+import logging
 import os
 import sys
-import logging
 import urllib2
 from cStringIO import StringIO
 from functools import reduce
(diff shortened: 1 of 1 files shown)


amaltaro:
Ok, I tested it with sevelal harvesting jobs (some set up to fail) and it seems to be working fine. I didn't manage to exercise the exact problem it solves though: not raising any exception during the upload.

amaltaro:
@hufnagel you might want to have a look at this patch too.

hufnagel:
Looks good per-se. To properly test this, you would have to intentionally break cmsweb (or better the testbed or a completely private instance). Maybe check with the cmsweb team how best to replicate the behavior.

amaltaro:
I managed to test it with cmsweb-dev and it works fine, the only problem is the log.exception that makes the error show twice in the job log. Let me make this fix a bit uglier to avoid the double error in the logs

amaltaro:
It works fine and it's ready to go!

ticoann (approved):

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -150,27 +150,33 @@ def upd(m, data):
...
-                msg += 'Status code: %s\n' % headers.get("Dqm-Status-Code", None)
-                msg += 'Message: %s\n' % headers.get("Dqm-Status-Message", None)
-                msg += 'Detail: %s\n' % headers.get("Dqm-Status-Detail", None)
-                msg += 'Data: %s\n' % str(data)
-                logging.info(msg)
+                msg = '  Status code: %s\n' % headers.get("Dqm-Status-Code", None)
+                msg += '  Message: %s\n' % headers.get("Dqm-Status-Message", None)
+                msg += '  Detail: %s\n' % headers.get("Dqm-Status-Detail", None)
+                msg += '  Data: %s\n' % str(data)
+                if int(headers.get("Dqm-Status-Code", 400)) >= 400:
```
ticoann: I assume it doesn't contain the Dqm-Status-Code considers as failure?

amaltaro: Yes, that's why I put the default to 400. If it doesn't have that header, then - I believe - the DQM server didn't handle that request.

andresfelquintero:
Hello Alan, can we change the DQMUpload.py file on the production machine with this patch?

amaltaro:
@andresfelquintero , Vytas @vytjan told me this patch has already been applied to both production T0 agents.

PR #8821, opened 2018-09-20
add CRIC unittest

ticoann:
fixes #8794

PR #8848, opened 2018-10-11
Added keyword args to the CRIC module

### LINKED ISSUE #8847: how to make CRIC use pycurl ?
New Service.CRIC module defaults to httplib2.
But I need to use pycurl which is more available, e.g. on schedd's, we
had gone through a small campaign in past years to make CRAB use only pycurl,
but this is not now not the case and odd errors are popping up.
How can I CRIC to use pycurl,  in lack of an explicit argument like SiteDB API had ?

see old usage:
https://github.com/dmwm/CRABServer/blob/1243414ef7b120699cf7f769570309b3fb09bdbf/src/python/TaskWorker/Actions/DagmanCreator.py#L831

amaltaro:
Fixes #8847
Also includes the changes made by Seangchan on #8821

In summary:
* accept a dictionary with further configuration key/value pairs to be used for overriding the default parameters of the super classes (Services and Requests)
* added a couple of wrapper functions to the CRIC module to make it cleaner and easier to run unit tests
* also updated the cacheduration, which is actually in hours instead of secs

### DIFF
--- a/src/python/WMCore/Services/CRIC/CRIC.py
+++ b/src/python/WMCore/Services/CRIC/CRIC.py
@@ -27,15 +27,19 @@ class CRIC(Service):
     Class which provides client APIs to the CRIC service.
     """
 
-    def __init__(self, url=None, logger=None):
-        params = {}
-        defaultURL = "https://cms-cric.cern.ch/"
-        params['endpoint'] = url or defaultURL
-        params.setdefault('cacheduration', 3600)
-        params.setdefault('accept_type', 'application/json')
-        params.setdefault('content_type', 'application/json')
-        params['logger'] = logger if logger else logging.getLogger()
-        Service.__init__(self, params)
+    def __init__(self, url=None, logger=None, configDict=None):
+        """
+        configDict is a dictionary with parameters that are passed
+        to the super class
+        """
+        url = url or "https://cms-cric.cern.ch/"
+        configDict = configDict or {}
+        configDict.setdefault('endpoint', url)
+        configDict.setdefault('cacheduration', 1)  # in hours
+        configDict.setdefault('accept_type', 'application/json')
+        configDict.setdefault('content_type', 'application/json')
+        configDict['logger'] = logger if logger else logging.getLogger()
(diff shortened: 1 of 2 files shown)


amaltaro:
test this please

amaltaro:
My tests (and the unit tests) are looking Ok.
@belforte do you need this patch soon? If you want I can merge it and cut a new tag for you, otherwise I'd wait for Seangchan until Monday or so.

belforte:
thanks @amaltaro I an surely wait to get this when most convenient to you.

belforte:
@amaltaro did you test using the new input parameter ? Can you give me an example ?
(add to unittest?)
 I could not find a way to make it work. Both of these give a syntax error:
```
CRIC({"pycurl":True},logger=logger)
CRIC{logger=logger,{"pycurl":True})
```
while this runs, but has no effect (i.e. does not switch from httpliib to curl):
```
CRIC(config={"pycurl":True},logger=logger)
```
because it inserts in the config dictionary for Service `'config':{'pycurl': True}` instead of simply `'pycurl': True`

amaltaro:
Yes I tested it. I also made a unit test for that, see `def testConfig`.
I can change it to take a config dictionary instead, I just had the impression doing it with kwargs was a more pythonic way :)

belforte:
thanks Alan. Sorry for panicking too early :-) All fine.

amaltaro:
@belforte I updated this patch again, such that you can pass `config={...}` when instantiating a CRIC object. I'd actually move the `logger` parameter in there too, but anyways... 
Check the same unit test for an example

belforte:
@amaltaro whatever you like more. As we say in Ital:  too many cooks ruins the cooking

ticoann (approved):

ticoann:
Thanks Alan, I still need to update the unittest code a little bit but I will do so in separate patch.

amaltaro:
Oh, I didn't really review all those unit tests.
I fixed and improved what was left and a new PR was made https://github.com/dmwm/WMCore/pull/8850

PR #8202, opened 2017-09-28
Migrate optparse to argparse.

alexanderrichards:

PR #8849, opened 2018-10-11
Migrate from optparse to argparse

### LINKED ISSUE #7158: Update code using optparse to argparse
optparse is deprecated since py2.7. At least the reqmgr client scripts use it.

amaltaro:
Fixes #7158 
Resuming the work done by Alex on #8202

Very annoying PR, some files can likely be deleted (etc/harvestingInjector).
I also had to replace `raw_input` by `input`, otherwise python3 compatibility tests would start failing. Notice that with that change, we can no longer type only `yes` in the prompt, but actually `"yes"` has to be typed...

### DIFF
--- a/bin/wmagent-delete-couchdb-workflow
+++ b/bin/wmagent-delete-couchdb-workflow
@@ -1,18 +1,22 @@
 #!/usr/bin/env python
+from __future__ import print_function
 
 import os
 import sys
 from argparse import ArgumentParser
-from WMCore.Database.CMSCouch import CouchServer
+
 from WMCore.Configuration import loadConfigurationFile
+from WMCore.Database.CMSCouch import CouchServer
+
 
 def getJobDumpCouchDB(couchDBUrl, jobDBName):
-    jobCouchdb  = CouchServer(couchDBUrl)
-            
+    jobCouchdb = CouchServer(couchDBUrl)
+
     jobsdatabase = jobCouchdb.connectDatabase("%s/jobs" % jobDBName)
     fwjrdatabase = jobCouchdb.connectDatabase("%s/fwjrs" % jobDBName)
-    
-    return jobsdatabase, fwjrdatabase 
+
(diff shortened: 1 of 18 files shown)


amaltaro:
test this please

bin/wmagent-delete-couchdb-workflow
```diff
@@ -26,43 +30,43 @@ def deleteWorkflowFromCouch(workflowName, jobsdatabase, fwjrdatabase):
...
+        fwjrdatabase.delete_doc(id=id, rev=rev)
 
     return len(jobs), len(fjobs)
 
-if __name__ == "__main__":
 
-    answer = raw_input("Warning: This is irreversible. Are you sure to delete workflows (yes, no)? ")
+if __name__ == "__main__":
+    print("Warning: This is irreversible. ")
+    answer = input("Are you sure to delete workflows (yes, no, type it between quotes)? ")
```
ticoann: I think you can explicitly say 'yes', 'no'  like above instruction
```
Type 'Y' or 'N' (type quotes too!)
```

bin/wmagent-unregister-wmstats
```diff
@@ -1,40 +1,42 @@
...
+    (options, args) = parser.parse_known_args()
     if not args:
         agentUrl = ("%s" % wmagentConfig.Agent.hostName)
     else:
         agentUrl = args[0]
-    
-    answer = raw_input("Warning: This is remove agent info record for %s from wmstats (yes, no)?" % agentUrl)
+
+    msg = "Are you sure you want to remove agent info record for %s from wmstats (yes, no, between quotes please)? "
+    answer = input(msg % agentUrl)
```
ticoann: Same as above. (When we move to python3 quotation need to be dropped, right?)

amaltaro:
Yes, when we move to python3, `input` will have the same behavior as `raw_input`, so it won't try to evaluate the input as an expression.
I'll update the message then.

PR #8855, opened 2018-10-17
post wmstats summary directly to central wmstats

### LINKED ISSUE #8851: Recent workflow missing agent data in wmstats
Probably related to this patch #8840

Sharad reported these three prep_ids:
task_EXO-RunIISummer16MiniAODv3-04050
task_EXO-RunIISummer16MiniAODv3-10105
task_EXO-RunIISummer16MiniAODv3-10890

that have data missing in wmstats. I had a look at the first one and its status transition are very recent
```
acquired: 2018-10-13 17:43:09 UTC DN: /DC=ch/DC=cern/OU=computers/CN=dmwm/cmsweb.cern.ch
running-open: 2018-10-13 17:53:29 UTC DN: /DC=ch/DC=cern/OU=computers/CN=dmwm/cmsweb.cern.ch
running-closed: 2018-10-13 17:53:30 UTC DN: /DC=ch/DC=cern/OU=computers/CN=dmwm/cmsweb.cern.ch
completed: 2018-10-14 13:21:44 UTC DN: /DC=ch/DC=cern/OU=computers/CN=dmwm/cmsweb.cern.ch
```

it was processed by vocms0280, which was patched and is apparently working without any issues.
I'm investigating why this document hasn't been uploaded to wmstats...

ticoann:
fixes #8851

This fix contains new view. - Need to come up with temporary WMAgent patch without updating the views in cmsweb couch. Then reapply this patch when cmsweb is updated.

We still need replication since local wmstats db contains docs for failed job as well.

### DIFF
--- a/src/couchapps/WMStats/rewrites.json
+++ b/src/couchapps/WMStats/rewrites.json
@@ -49,6 +49,11 @@
   },
 
   {
+    "from": "_view/byAgentURL",
+    "to": "../WMStatsErl/_view/byAgentURL"
+  },
+
+  {
     "from": "_view/cooledoffRequests",
     "to": "../WMStatsErl/_view/cooledoffRequests"
   },
--- a/src/couchapps/WMStatsAgent/views/agentRequests/map.js
+++ b/src/couchapps/WMStatsAgent/views/agentRequests/map.js
@@ -1,8 +1,0 @@
-function(doc) {
-  if (doc.type === "agent_request" && doc.agent_url) {
-    // this condition check can be removed when there is no historical data is left. Not sure why this check is not working
-    //if (doc['_id'].startsWith(doc.agent_url)) {
-        emit(doc['_id'],  {'rev': doc['_rev']});
-    //}
-  }
-}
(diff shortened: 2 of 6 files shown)


ticoann:
@amaltaro, I check the datasize for include_docs for initial cache update, the data we need to get initially ~250MB and takes around 5 min to get it (in my browser, it would be possibly faster in agents). But I don't think this is good temporary solution. I will try to figure out other way to get the rev number but I am afraid we might have to request production couch view update. I will test this patch first.

ticoann:
test this please

ticoann:
tested in my vm, working OK.

ticoann:
> aren't those in the wmagent_jobdump database?

It is not in job dump db,  it is in  wmagent_summary. It contains 2 types of data, individual job data for failed job and job summary data by request. Size wise individual job data will be much more and bigger. Those can also directly update to the central location but those data is not temporary data so replication might be better. I still think directly updating summary data will be an option if we cannot fix the replication. I think replication problem only due to those summary data.

ticoann:
test this please

PR #8856, opened 2018-10-18
Does not raise DBSReaderError when block has no valid files

### LINKED ISSUE #8846: getDBSSummaryInfo should not raise an exception for blocks with only INVALID files
As discussed in a private thread "Error in DBSReader.listDatasetSummary" with Yuyi and CRAB folks, when we call this WMCore/DBSReader `getDBSSummaryInfo`
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L377

passing a block (dataset would be the same) which only contains invalid files, DBSServer returns an empty list as result. However, our code raises a `DBSReaderError` exception in L395.

It seems to be necessary by the GQ start policies, to make sure that this block is skipped all the way up in the chain.

I need to investigate to see what are the required changes if we stop raising this exception.

amaltaro:
Fixes #8846 

Changes are:
* no longer raise an exception in DBS3Reader getDBSSummaryInfo API when it cannot find any valid file in the block/dataset
* adapted the Workqueue code to handle an empty dictionary instead of the exception
* log every time we find a block where all files are invalid
* log every time we find a block that doesn't pass the runs restrictions
* post rejected work to the LogDB as a warning
  * TODO: make sure this warning remains there until the workflow gets archived (see #7808)

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -378,21 +378,21 @@ class DBS3Reader(object):
         """
         Get dataset summary includes # of files, events, blocks and total size
         """
-        # FIXME: Doesnt raise exceptions on missing data as old api did
         if dataset:
             self.checkDatasetPath(dataset)
         try:
             if block:
                 summary = self.dbs.listFileSummaries(block_name=block, validFileOnly=1)
-            else:  # dataset case dataset shouldn't be None
+            else:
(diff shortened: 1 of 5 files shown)


belforte:
thanks for letting us know @amaltaro . @lecriste please take note.

amaltaro:
@ticoann Seangchan, I think this one is good to go to testbed tomorrow.

ticoann (approved):

PR #8859, opened 2018-10-18
1.1.15_wmagent wmstats replication temp fix

ticoann:
This patch requires WMAgent couchapp update as well.

### DIFF
--- a/src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
+++ b/src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
@@ -36,6 +36,9 @@ class AnalyticsPoller(BaseWorkerThread):
         self.summaryLevel = (config.AnalyticsDataCollector.summaryLevel).lower()
         self.pluginName = getattr(config.AnalyticsDataCollector, "pluginName", None)
         self.plugin = None
+        self.revCache = {} # key: doc id, value doc rev
+        self.loopCount = 0
+        self.crashFlag = False
 
     def setup(self, parameters):
         """
@@ -138,8 +141,27 @@ class AnalyticsPoller(BaseWorkerThread):
             if self.plugin != None:
                 self.plugin(requestDocs, self.localSummaryCouchDB, self.centralRequestCouchDB)
 
-            existingDocs = self.localSummaryCouchDB.getAllAgentRequestRevByID()
-            self.localSummaryCouchDB.bulkUpdateData(requestDocs, existingDocs)
+            if self.loopCount < 2 and not self.revCache:
+                try:
+                    logging.info("Populating cache of revision number...")
+                    start = int(time.time())
+                    self.revCache = self.centralWMStatsCouchDB.getAllAgentRequestFromCentralServer()
+                    end = int(time.time())
+                    logging.info("Cache populated - took %s sec", (end - start))
+                except Exception as ex:
+                    logging.warning("Populating cache failed: %s ", str(ex))
+
+            if self.loopCount >= 2 and not self.revCache:
+                self.crashFlag = True
+                raise Exception("CacheUpload failed")
+
+            notInCacheDoc, notInCacheKey = self.centralWMStatsCouchDB.bulkUpdateDataAnUpdateCache(requestDocs, self.revCache)
+
+            if notInCacheKey:
+                missingDocs = self.centralWMStatsCouchDB.getAllAgentRequestFromCentralServer(notInCacheKey)
+                self.centralWMStatsCouchDB.bulkUpdateDataAnUpdateCache(notInCacheDoc, missingDocs, secondTry=True)
+
+            self.loopCount += 1
 
             logging.info("Request data upload success\n %s request, \nsleep for next cycle", len(requestDocs))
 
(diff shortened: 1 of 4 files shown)


ticoann:
If this works we just need to run agent with this patch until cmsweb upgrade. then revert this patch and create new patch with #8855

src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
```diff
@@ -77,6 +78,12 @@ def setup(self, parameters):
             pluginFactory = WMFactory("plugins", "WMComponent.AnalyticsDataCollector.Plugins")
             self.plugin = pluginFactory.loadObject(classname=self.pluginName)
 
+        # delete all agent-request type doc.
+        self.localSummaryCouchDB.deleteAllAgentRequestDocument()
```
amaltaro: Seangchan, what happens to the job-related documents? Can they be recreated and wmstats repopulated with them?

src/python/WMCore/Services/WMStats/WMStatsWriter.py
```diff
@@ -99,6 +99,38 @@ def bulkUpdateData(self, docs, existingDocs):
...
+                doc['_rev'] = "%s-%s" % (int(revList[0]) + 1, revList[1])
+                self.couchDB.queue(doc)
+                revCache[doc['_id']] = doc['_rev']
+            else:
+                if secondTry:
+                    doc['_rev'] = "1-123456789"
+                    self.couchDB.queue(doc)
+                    revCache[doc['_id']] = doc['_rev']
+                else:
+                    idParts = doc['_id'].split('-')
```
amaltaro: Shouldn't it be `doc['_rev']...`? And what happens if it's the first time we create the document, it doesn't have a revision number yet.

src/python/WMComponent/AnalyticsDataCollector/AnalyticsPoller.py
```diff
@@ -77,6 +78,12 @@ def setup(self, parameters):
...
+    
+        # populate cache: This will take time ~5 min
+        self.revCache = self.centralWMStatsCouchDB.getAllAgentRequestFromCentralServer()
```
amaltaro: How about encapsulating it in a try/except and printing a nice error message if it fails (likely timeout) and assign an empty dict to `self.revCache`?

ticoann:
This seems to be working in my vm. Except the deletion part. I deletes the document with different revision number. I am not sure how it get that number.

ticoann:
@amaltaro, I changed repfilter not to replicate agent_request doc directly instead of rotating the database. So far it seems OK. If I don't see the couch server crash until tomorrow morning. I will update drained production agents tomorrow and one non draining production agent to test out.

ticoann:
I don't see the replication errors anymore. I will apply this patch to all drained agent.
This patch requires WMAgent couchapp update.

amaltaro:
awesome!

ticoann:
Alan, I applied patch now to below.
draining agent (submit6, submit8, vocms0250, vocms0252)
active agent (vocms0280).
(twiki is updated)

I will monitor today and if everything is ok I will apply to all other agent and update trello board.

ticoann:
Now all the agents are updated.

draining agent (submit6, submit8, vocms0250, vocms0252)
active agent (submit7,   vocms0256. vocms0280, vocms0281,  vocms0282).

fingers crossed. :-) I will merge this and create the new patch for when cmsweb is upgraded.

amaltaro:
Thank you very much, Seangchan.

From the list of workflows with missing data in WMStats, provided by Sharad in this ticket https://github.com/dmwm/WMCore/issues/8851

the only one left without any information yet is:
task_EXO-RunIISummer16MiniAODv3-10105

Hopefully it's just a patch that wasn't properly applied to one agent. I'll try to check it this morning

amaltaro:
This workflow
pdmvserv_task_EXO-RunIISummer16MiniAODv3-10105__v1_T_180923_230353_8407

was being worked on by vocms0250, which turns not to have been properly patched (we missed the src code part). It's done, services+agents restarted and couchapps pushed (again).

amaltaro:
vocms0252 was also missing the src/ changes. I updated this patch over there and also double check all the other active agents and I can see their information in wmstats. So we should be good by now.

Sorry if I lost track, but how are we going to handle it in the coming agent release (master branch?) What we need to revert once CMSWEB is upgrade?

Thank you very much for fixing it, Seangchan.

amaltaro:
[Fix #8859 - wmstats replication temp fix](https://trello.com/c/EMbY3SEx/131-fix-8859-wmstats-replication-temp-fix)

ticoann:
Sorry Alan, was there any conflict on vocms0252 or couch app is updated? I remember I applied patch there. When cmsweb is upgrade we will remove the cache and use new view directly. Although the cache refresh is not so bad (took < 1min in most cases), I think we can make it consistent. I will create the patch.

amaltaro:
I didn't check whether the couchapp was updated, only that the javascript code was updated, while the src code wasn't. No problem, it's all sorted now and we can start shutting agents down again (submit6 was the first one ;))

PR #8860, opened 2018-10-19
fix the replication on wmstats

ticoann:
@amaltaro, Alan, I think this is the reason replication didn't work well. To make the replication work, it keeps the revision history. With just rev change it doesn't seem to have revision history stored.

I am testing this in my vm.
1. update the agent with thjs patch.
2. remove the cental wmstats data base
3. replicate the data again.

It seems  replication is very slow. It might be stuck though.I will check back tomorrow morning

### DIFF
--- a/src/python/WMCore/Services/WMStats/WMStatsWriter.py
+++ b/src/python/WMCore/Services/WMStats/WMStatsWriter.py
@@ -90,10 +90,10 @@ class WMStatsWriter(WMStatsReader):
             if doc['_id'] in existingDocs:
                 revList = existingDocs[doc['_id']].split('-')
                 # update the revision number
-                doc['_rev'] = "%s-%s" % (int(revList[0]) + 1, revList[1])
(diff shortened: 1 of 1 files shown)


amaltaro:
Let me try to understand it.

You're actually trying to overwrite the document revision history such that it always has the current doc _rev in the history only, and nothing else. Is that correct? Where did you get this information from?
I found a pdf documentation (page 286)
https://media.readthedocs.org/pdf/couchdb/1.6.1/couchdb.pdf

which says that ids have to be an array of valid ids. According to the code in this PR, I see two issues:
* when the doc exists, you don't update its _rev number
* the way it sets the _revisions attr, you end up with an invalid ids (we don't increment the last bit). E.g.:
`doc['_revisions'] = {"start":int(revList[0]) + 1,"ids":[str(int(revList[1]) + 1), revList[1]]}`
if revision number is "1-1234567890", then that line would evaluate to
`doc['_revisions'] = {"start":2, "ids":[1234567891, 1234567890]}`

where id "1234567891" won't exist. Unless I'm missing something.

ticoann:
integrated to #8851

PR #8863, opened 2018-10-23
Add CRIC flag for central services

### LINKED ISSUE #8834: Create a CRIC/SiteDB switch for central services
Related to #8696, but for services running on CMSWEB, like ReqMgr2 and Global WorkQueue.
In order to make it simple, I think we can use the same mechanism as in the agent, so defining env vars in their config.py.

amaltaro:
Fixes #8834
Requires these deployment changes https://github.com/dmwm/deployment/pull/682

Created a new `WMCORE_USE_CRIC` environment variable for central services (ReqMgr2 and WorkQueue) such that we can switch between CRIC and SiteDB.

### DIFF
--- a/src/python/WMCore/Services/SiteDB/SiteDBAPI.py
+++ b/src/python/WMCore/Services/SiteDB/SiteDBAPI.py
@@ -33,7 +33,7 @@ class SiteDBAPI(Service):
 
     def __init__(self, config={}, logger=None):
         config = dict(config)
-        if os.getenv("WMAGENT_USE_CRIC", False):
+        if os.getenv("WMAGENT_USE_CRIC", False) or os.getenv("WMCORE_USE_CRIC", False):
             # just to make sure we don't use SiteDB anywhere when CRIC flag is true
             config.setdefault('endpoint', "https://BLAH.cern.ch/sitedb/data/prod/")
         else:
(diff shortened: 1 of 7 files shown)


amaltaro:
Ok, I tested this change in central services and again in WMAgent (with SiteDB broken url) and everything works fine. I'm going to remove the 3rd commit that breaks SiteDB and it should be good to go.

ticoann:
test this please

src/python/WMCore/Services/SiteDB/SiteDBAPI.py
```diff
@@ -33,7 +33,7 @@ class SiteDBAPI(Service):
 
     def __init__(self, config={}, logger=None):
         config = dict(config)
-        if os.getenv("WMAGENT_USE_CRIC", False):
+        if os.getenv("WMAGENT_USE_CRIC", False) or os.getenv("WMCORE_USE_CRIC", False):
             # just to make sure we don't use SiteDB anywhere when CRIC flag is true
             config.setdefault('endpoint', "https://BLAH.cern.ch/sitedb/data/prod/")
```
ticoann: So you want to crash the component if SiteDB is used, right? Can we just raise an exception?

ticoann:
Patch looks good  to  me, Alan.

src/python/WMCore/Services/SiteDB/SiteDBAPI.py
amaltaro: It could have been done. I was afraid there could be a place in the code we get an instance of SiteDB without actually using it, thus I didn't want to have an exception for that case. However, it would probably have been a better idea.
If you want I can create another patch for this, but I think it's really no longer used, otherwise I'd have hit this problem in my tests.

PR #8866, opened 2018-10-25
Use CRIC in ReqMgr2 GUI

amaltaro:
Complement to #8863


amaltaro:
test this please

amaltaro:
@ticoann this patch has been tested and I think that was the last place still using SiteDB, hopefully.
However, I added another commit as we discussed on the other PR such that SiteDB raises an exception when it is not supposed to be used (when CRIC env vars are defined).
Please let me know if we should raise another exception or change something.

ticoann:
Patch looks good, Alan. Thanks,

Issue #8864, opened 2018-10-24
Expose login info outside CherryPy


bbockelm:
Hi,

After a discussion with @belforte, we think that all of CRAB's direct usage of CherryPy looks like this:

https://github.com/dmwm/CRABServer/blob/631c55d6cc9f5aae1c73085b1e38051d00a4494c/src/python/UserFileCache/RESTExtensions.py#L38

That is, we look at the thread-local request variable to retrieve the `user` member set here:

https://github.com/dmwm/WMCore/blob/0615bd977edd4ad103db7ea24f461aa794c9a104/src/python/WMCore/REST/Auth.py#L108

Could we add a method to `WMCore.REST.Auth` that is basically:

```
def get_user_info():
    return cherrypy.request.user
```

?

With this, we can remove the direct CherryPy usage from CRABServer and make any transition / upgrade simpler.

Brian

belforte:
I was actually thinking this to happen as part of moving from cmsweb FE to
a (THE ?) new authz. via tokens. So introducing a hook to a WMCore method
sounds great. And also the time to go from group/role to a leaner scheme.

vkuznet:
+1

belforte:
would this be a good time to document what cherrypy.request.user is, beside "a dictionary" ? so that we define what user/group/role means and are led to define how to do authz w/o roles ?

vkuznet:
Well, there is no cherrypy.request.user it is us (CMS) who append user to cherrypy.request. This is done in Auth.py module
https://github.com/dmwm/WMCore/blob/0615bd977edd4ad103db7ea24f461aa794c9a104/src/python/WMCore/REST/Auth.py#L108
which use user_info_from_headers function defined here
https://github.com/dmwm/WMCore/blob/0615bd977edd4ad103db7ea24f461aa794c9a104/src/python/WMCore/REST/Auth.py#L3
which by itself defines the dictionary
```
user = { 'dn': None, 'method': None, 'login': None, 'name': None, 'roles': {} }
```

belforte:
I did not say cherrypy defines this. I am asking CMS to be clear,
if you look more in the code there are places where `'group` key is added, and  where it is used e.g. in CRAB REST side. And while good written source can serve as good documentation at times, it is not IMHO a good reason to avoid properly document things in some less hidden place as well.
[...]

PR #8868, opened 2018-10-25
Add get_user_info helper function to encapsulate underlying technology

vkuznet:
Fix Issue #8864


ticoann:
Thanks Valentin for the patch, there are ~10 places in WMCore which need to replaced by the new function.  I will create a new patch for that.

Issue #6072, opened 2015-07-24
Create a fine-grained permissions to ReqMgr2


amaltaro:
If I looked at the right file
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Auth.py

it seems the only permissions distinctions are for: creating a request (new), approving (assign-approved) and assigning it (assigned).

If someone wants to reject a request, then this person needs to have full privileges (ReqMgr: Admin).

My suggestion would be (for ReqMgr group):
- User: should be able to create, approve and reject workflows. nothing else.
- Operator: should be able to create requests and change their status to anything that is not *-archived
- Admin: full access.

@ticoann does it make sense?

amaltaro:
And two more comments:
 1. We need to keep FacOps: web-service, since this one is used by cmsweb.
 2. The same distinctions/privileges should apply to couch docs, though I'm not sure how hard it is (do we need to create a validate-func to each design doc?).

ticoann:
Hi Alan, Yes I agree what you said.
The plan was, make this configuable for now it is hard coded
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Auth.py 

Base logic are following.
1. Authentication group and roles are not hosted in reqmgr2 unlike reqmgr1. (It uses the groups and roles from sitedb).
2. permission is tied to group and roles as well as request type. (So we define permission for those combination) i.e. you set set specific group, and/or role to approve StoreResult request but that group won't have approve privilege for other types of request.  
(It doesn't have user level and individual request level control, (i.e. only user who create the request can reject request. - I don't think we need that level of control, but we can add it if it is absolutely necessary.)

amaltaro:
Thanks Seangchan! Individual request level control is just too much in my opinion as well.

About permissions being tied to groups, did you mean physics group? If so, IMO we could skip it as well, currently there are only a few people creating workflows, I'd go for a single group, easier to manage.

ticoann:
group and role here is just sitedb group and roles, I am not sure how groups and roles defined there. But it is pretty flexible to define groups and roles, right?

vlimant:
good !

ticoann:
@vlimant, Jean-Roch could you give me the list which one needs which permission to have.

Here is some examples

https://cmsweb-testbed.cern.ch/couchdb/_utils/document.html?reqmgr_auxiliary/PERMISSION_BY_REQUEST_TYPE

[...]

PR #8869, opened 2018-10-25
Grey out status not allowed to humans in ReqMgr2 Gui

amaltaro:
Related to #6072 

This change only affects the ReqMgr2 GUI, forbidding request status transition that a human user is not supposed to make via GUI.

@ticoann in case you want to see it live, I have this patch applied to alancc7-cloud1.


ticoann:
Thanks Alan, I didn't try it but patch looks good.

PR #8870, opened 2018-10-25
Delete LogDB wmstats docs when wfs are archived

### LINKED ISSUE #7808: Report the reason why a workflow failed to be acquired

amaltaro:
Fixes #7808

Changes are:
* when moving a request to archive state, first delete any logDB documents it might have and only then make the state transition
* WMStats LogDBTasks thread was kept around (we might completely remove it in the next month), but it's configuration was adapted such that it only deletes docs older than 90 days (if there will be any)

Deployment change in https://github.com/dmwm/deployment/pull/685

### DIFF
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
@@ -6,6 +6,7 @@ from __future__ import (division, print_function)
 import time
 from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
 from WMCore.ReqMgr.DataStructs.RequestStatus import AUTO_TRANSITION
+from WMCore.Services.LogDB.LogDB import LogDB
 from WMCore.Services.WorkQueue.WorkQueue import WorkQueue
 from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
 from WMCore.Services.WMStatsServer.WMStatsServer import WMStatsServer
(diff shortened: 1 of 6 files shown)


ticoann:
Hi Alan, I have a few questions.
1. Current setting is that the logDB not deleted?
2. It seems the procedure is not fault tolerant (it the deletion fails it doesn't get retried.

amaltaro:
test this please

amaltaro:
Seangchan, please leave your review and I'll check it tomorrow. I'm also running some tests over night and will check the outcome tomorrow morning before merging it. Thanks

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -9,103 +9,16 @@
...
-                try:
-                    reqmgrSvc.updateRequestStatus(reqInfo["RequestName"], nextStatus)
-                    count += 1
-                except Exception as ex:
-                    logger.error("Fail to update %s: %s", reqInfo["RequestName"], str(ex))
-        # convert to start status for the logging purpose
-        initStatus = "aborted" if status == "aborted-completed" else status
-        logger.info("Total %s-archived: %d", initStatus, count)
-
-    return
```
ticoann: Alan, the reason the functions are out side of class is in that way we can create the script easily and run it separately. I think I have script to use that way.  Also it is easier to test and debug. 
However, if you prefer change to class method as in the patch, this would be fine as well.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -117,19 +30,111 @@ def advanceStatus(self, config):
...
+
+        for status, nextStatus in statusTransition.items():
+            inputConditon = {"RequestStatus": [status], "AgentJobInfo": "CLEANED"}
+            for reqInfo in self.wmstatsSvc.getFilteredActiveData(inputConditon, outputMask):
+                reqName = reqInfo["RequestName"]
+                if reqName and (not reqInfo["RequestTransition"] or
+                                        (currentTime - reqInfo["RequestTransition"][-1]["UpdateTime"]) > threshold):
+                    try:
+                        self.logger.info("Deleting %s from LogDB WMStats...", reqName)
+                        res = self.logdb.delete(reqName, agent=False)
```
ticoann: Is this deleting also agent documents?

amaltaro:
Seangchan, basic tests work fine. I reverted changes as you suggested.
I'm merging it now and we can further validate it in testbed.

PR #8875, opened 2018-10-29
Distinguish between bad work units and work that doesnt pass constraints

### LINKED ISSUE #8872: Reject work in GQ should only be account for bad input
With the changes in
https://github.com/dmwm/WMCore/pull/8856

it looks like we report blocks that don't match the Run*list or Lumilist as a rejected/bad work, e.g.:
https://alancc7-cloud1.cern.ch/couchdb/_utils/document.html?wmstats_logdb/8008880e209e6da56487c9c6c4edb871

So, workflows providing LumiList, or RunWhitelist/RunBlacklist should not report blocks that don't match those constraints as a logdb error document (meaning not in rejectedWork).

amaltaro:
Fixes #8872 

I didn't notice `self.rejectedWork` was also used for keeping track of work units (blocks) that don't pass possible work restrictions (block, run, lumi mask). We need to separate those, otherwise there are tons of LogDB request-level warnings that aren't supposed to be there.

Looking into the usefulness of `self.rejectedWork`. Apparently we keep it in the single workqueue_inbox document where we have a complete list of `RejectedInputs` and `ProcessedInputs` such that it can have a more efficient cycle the next time it runs (through `modifyPolicyForWorkAddition`). However, didn't we kill this WorkAddition when we deprecated OpenRunningTimeout?

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/Block.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/Block.py
@@ -122,6 +122,7 @@ class Block(StartPolicyInterface):
                 # Don't duplicate blocks rejected before or blocks that were included and therefore are now in the blacklist
                 continue
             if task.getLumiMask() and blockName not in maskedBlocks:
+                logging.warning("Block %s doesn't pass the lumi mask constraints", blockName)
                 self.rejectedWork.append(blockName)
                 continue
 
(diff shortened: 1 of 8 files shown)


amaltaro:
Seangchan, initial tests look Ok, but I still need to check the outcome more careful. Can you please have a look at it?

If you prefer - and this patch looks good to you - I can merge this and your other PR tomorrow and create a new tag for Lina.

ticoann (approved):

ticoann:
Hi Alan, patch looks good to me. When you test this please merge and create a new tag. I don't have any other patch to include at the moment except deployment patch.

https://github.com/dmwm/deployment/pull/687

Please include above when you make pull request. Thanks.

amaltaro:
Thanks, Seangchan. Tests look good and the only template that has a problem is:
ReReco_Parents.json

with the following logdb entry (just for the record):
```
Request with the following unprocessable input data: ['/Cosmics/Commissioning2015-PromptReco-v1/RECO#2b3acb96-cabe-11e4-91f0-001e67ac06a0']
```

PR #8876, opened 2018-10-29
add delay for key up

ticoann:
@amaltaro Alan, this will some of the performance issue in wmstats.  (won't run the filter function too often).
Tested in vm. Please review.


amaltaro:
I'm not sure I understand this patch, but if you say it will improve the performance, I believe in you :-)
Is it sort of defining a polling cicle for that filter function to run?

ticoann:
Hi Alan, sorry for not describing the patch. Without it, it triggers filtering right after a user types something.
If you want to check all the workflows start with "alan",  It is start filtering when you type a, l, a, n. so even though you don't want to search all the workflows with 'a', it is starting to filter that and waist cpu time especially when there are a lot of data. Patch gives the delay of 0.3 second so it won't start filtering the data unless someone is very slow typist.

amaltaro:
Thanks, Seangchan. That is something very annoying indeed and I even wonder whether it should be something between .5 and 1. seconds.. We shall see how better it gets now ;)

PR #8878, opened 2018-10-30
filter only necessary information for wmstats

### LINKED ISSUE #8831: WMStat not loading in Firefox
this was reported, but I don't see an high priority GH about this

ticoann:
fixes #8831

### DIFF
--- a/src/python/WMCore/WMStats/DataStructs/DataCache.py
+++ b/src/python/WMCore/WMStats/DataStructs/DataCache.py
@@ -1,5 +1,76 @@
 import time
+
 from WMCore.ReqMgr.DataStructs.Request import RequestInfo, protectedLFNs
+
+def _updateFilteredResult(result, key, value, currentExcluisveList, childExclusiveList):
+    """
+
+    :param result: nested dict of original data, this will have pruned result after the function call
+    :param key: key of the dictionary whose value need to be pruned.
+    :param value: dictionary fo the original result[key] with will be pruned
+    :param currentExcluisveList: key level exculive list. if key is in exclusive list it will be pruned.
+    :param childExclusiveList: child level exclusive list, result[key][some_property],
+           childExclusive list will be checked with 'some_property'
+    :return:
+    """
+    if key not in currentExcluisveList:
+        childKeyFlag = False
+        if len(childExclusiveList) > 0:
+            for prop, cExList in childExclusiveList[0].items():
+                if key == prop:
+                    childKeyFlag = True
+                    result[key] = filterExcludeList(value, cExList, childExclusiveList[1:])
+
+            if not childKeyFlag:
+                result[key] = value
+    return
+
+def filterExcludeList(wmstatsCache, exclude, childExclusiveList):
+    """
+
+    :param wmstatsCache: wmstats cache data
+    :param exclude: dict format of {"key": "skip"|"start", "list": list of excluding property]
+    :param childExclusiveList: list of exculding property for child level. it is chained
+           - first item is a child, second item is grand child, etc (it doesn't support multiple child
+
+    :return: filtered data of wmstatsCache
+    """
+    result = {}
+    for k, v in wmstatsCache.items():
+        # when keys are not know ahead and we have to filter next level. i.e. task name, site name as key
+        if exclude["key"] == "skip":
+            result[k] = {}
+            for sk, sv in v.items():
+                _updateFilteredResult(result[k], sk, sv, exclude["list"], childExclusiveList)
+        else:
+            _updateFilteredResult(result, k, v, exclude["list"], childExclusiveList)
+
+    return result
+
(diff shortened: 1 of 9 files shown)


ticoann:
@amaltaro, Alan this is working version, but still need to refine the test and filtering aspect.  Current filtering only gives 20% reduced data on my vm test. It depends on the data in wmstats. 
You can compare
https://reqmgr2-dev.cern.ch/wmstatsserver/data/wmstatsrequestcache 222K
https://reqmgr2-dev.cern.ch/wmstatsserver/data/requestcache  275K

I also set the cache duration to 5 min instead of refreshing the cache every time. I thought it wouldn't matter first since the cache on the server refreshed every 10 min.  But I am suspecting server might keep the compressed data in the memory. (I have to check whether there is any difference on the transfer performance.

Could you review the code?

ticoann:
@amaltaro, Alan please review. I tested in vm. If the code is OK. I will fix the pylin and Python3 compatibility.

src/python/WMCore/WMStats/DataStructs/DataCache.py
```diff
@@ -1,6 +1,61 @@
...
+        if exclude["key"] == "skip":
+            result[k] = {}
+            for sk, sv in v.items():
+                _updateFilteredResult(result[k], sk, sv, exclude["list"], childExclusiveList)
+        else:
+            _updateFilteredResult(result, k, v, exclude["list"], childExclusiveList)
+
+    return result
+
+def wmstatsFilter(wmstatsCache):
```
amaltaro: What's the purpose of key=start or skip?
Why these lists have different format compared to `requestExcudeList`?

src/python/WMCore/WMStats/DataStructs/DataCache.py
```diff
@@ -1,6 +1,61 @@
 import time
+
 from WMCore.ReqMgr.DataStructs.Request import RequestInfo, protectedLFNs
 
+def _updateFilteredResult(result, key, value, curentExcluisveList, childExclusiveList):
```
amaltaro: This needs to be better explained.  Typo in `curentExcluisveList`

amaltaro (commented):
Seangchan, honestly speaking, what was in my mind while reviewing this patch is that it's going to be quite CPU intensive. Not for half megabyte of data, but for half gigabyte it's likely going to deliver a bad performance.

I also made some comments in the code, but I didn't manage to digest all the details and a clear description/comments/doc string would be helpful.

I also see a risky implementation here which is one function calling each other over and over.

src/python/WMCore/WMStats/DataStructs/DataCache.py
ticoann: current db structure is mixed.  There are 2 types of keys. One type is unique values as key, i.e. request name, task name.  These type of key we can't filter it since we are not filtering out specific data.  
Other keys are property name {'task', 'site', etc} which can filter.  I know 'start', 'skip' is the bad name, do you have any suggestion,

amaltaro:
Seangchan, before we keep going down this track. I wonder if we can actually change the information at the source (push less data to wmstats, review what does not belong there, review what nobody uses) and make it cleaner. Have you already evaluated that?

About keys: start|skip. What's the idea, "skip" means we want to skip those properties and "start" means we do not do anything (we keep those properties)?

About
https://reqmgr2-dev.cern.ch/wmstatsserver/data/wmstatsrequestcache

it does not work.

ticoann:
Concerning about CPU usage is legit. I am not sure how much it will affect the performance.  Number of request is about 20K requests when we had trouble. (I think actual data size doesn't really matter since we store this already in the memory in any case (without this patch).  
I don't think this is putting to much stress on cpu though. The call is every 5 minutes. Compare to transferring data and unpacking in the browser, the cpu usage increase is small. 
You can test it with filtered_request api since that api is basically have the same effect.
(by comparing with a lot of filters and without filter) You can see adding more filters and mask actually reduces the time if that reduces the data volume.

https://cmsweb.cern.ch/wmstatsserver/data/filtered_requests

But I know that doesn't mean this will work the same way unless we test it against production data. My vm test has small number of data.
 
There are calling function to other functions (with recursive call). Initially code has many duplicate code and that was the result of refactoring.  I know recursive call is overhead but nested level is shallow. Only 3-4 call will be made.  Still code looks ugly, if there is the cleaner solution, I am all for it.

ticoann:
Problem of filtering from agent is good. I just don't know whether those data is used somewhere else (Unified, etc).  Also basically the same logic need to be applied unless we change the datastructure.
Sorry I edited the "skip", "start" meaning on previous comment, you can see why the name is like that.
 
https://reqmgr2-dev.cern.ch/wmstatsserver/data/wmstatsrequestcache
Should work now, I was testing something else so that patch wasn't in reqmgr2-dev. I just update the patch.

PR #8883, opened 2018-10-30
only look for HTTP/ at beginning of line. Fix #8882

### LINKED ISSUE #8882: pycurl_manager fails when remote server header contains GRIDHTTP_PASSCODE
I have found that when authenticating via a service certificate, CRIC server returns the following header [1], while when using a personal proxy it returns [2].
The problem with [1] is that it breaks pycurl_manager.py code in here
https://github.com/dmwm/WMCore/blob/c23a66cacdf7750121cbf62c87e811252ae7a05c/src/python/WMCore/Services/pycurl_manager.py#L74-L80
because the  line with `GRIDHTTP_PASSCODE` is interpreted as an `HTTP/.1.1. nnn ttttt` one and the code tries to assign the strings following it to status and reason, and thus the line
https://github.com/dmwm/WMCore/blob/c23a66cacdf7750121cbf62c87e811252ae7a05c/src/python/WMCore/Services/pycurl_manager.py#L80
raises an exception since `status` is not a number [3]

This was never noticed because CMSWEB always returnes headers of kind [2] (at least in the calls I saw running CRAB TaskWorker) and WMA deployments/tests always use operator proxy as credential (IIUC).

I am going to ask CRIC developers why the different headers, but for all I know (which is little) header [2] looks legal to me and works if minimally modify pycurl_manager code. So I believe that if would be good to make the parsing less fragile. E.g. changing `row.find('HTTP')` to `row.find('HTTP'/)` is enough !
I can make a PR with that, still looks a bit fragile though, maybe we want to make sure that HTTP is the first string of the header line instead ? So go for `row.find('HTTP')==0` in place of  `row.find('HTTP')!=-1`.
Would be better if someone who knows what HTTP specifications are, gives solid instructions.

More on the history of this bug discovery in this thread: https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2788/1/1/1/1/1.html

[1]
```
HTTP/1.1 200 OK
Date: Tue, 30 Oct 2018 15:23:26 GMT
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips mod_wsgi/3.4 Python/2.7.5 mod_gridsite/2.3.4
Vary: Cookie
X-Frame-Options: SAMEORIGIN
Set-Cookie: csrftoken=nlu6EiwuwEcyHgGaRQ9I6otz6wH88jC7OeRx2jjyAIgYPlvHAuTHDimooskiVSCm; expires=Tue, 29-Oct-2019 15:23:26 GMT; Max-Age=31449600; Path=/
Set-Cookie: sessionid=2vaa554o6cty3lintjvwzd5v00gyjs42; expires=Sun, 04-Nov-2018 15:23:26 GMT; httponly; Max-Age=432000; Path=/
Strict-Transport-Security: max-age=300; includeSubDomains
Set-Cookie: GRIDHTTP_PASSCODE=2c6da9c96efa2ad0farhda; domain=cms-cric.cern.ch; path=/; secure
Content-Length: 37354
Content-Type: application/json
```
[2]
```
HTTP/1.1 200 OK
Date: Tue, 30 Oct 2018 15:23:26 GMT
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips mod_wsgi/3.4 Python/2.7.5 mod_gridsite/2.3.4
Vary: Cookie
X-Frame-Options: SAMEORIGIN
Set-Cookie: csrftoken=VPUDCTFhVVksPki5FvWKDQbCrm96oYdyUeioJBFTYJ8ESJt3pcYemqux6Xw13ub5; expires=Tue, 29-Oct-2019 15:23:26 GMT; Max-Age=31449600; Path=/
Set-Cookie: sessionid=0p59a1gjkjn2o62pzdax0qvevibfi7ng; expires=Sun, 04-Nov-2018 15:23:26 GMT; httponly; Max-Age=432000; Path=/
Strict-Transport-Security: max-age=300; includeSubDomains
Content-Length: 37354
Content-Type: application/json
```

[3]
```
  File "/data/user/WMCore/src/python/WMCore/Services/pycurl_manager.py", line 84, in parse
    self.status = int(status)
ValueError: invalid literal for int() with base 10: 'Set-Cookie:'
```

belforte:
this solves current problem, but as I wrote in dmwm#8882 it would be better that someone who knows what HTTP specs are makes sure that header parsing is done properly.

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -71,7 +71,7 @@ class ResponseHeader(object):
             row = row.replace('\n', '')
             if  not row:
                 continue
-            if  row.find('HTTP') != -1 and \
+            if  row.find('HTTP/') == 0 and \
                 row.find('100') == -1: #HTTP/1.1 100 found: real header is later
                 res = row.replace('HTTP/1.1', '')
                 res = res.replace('HTTP/1.0', '')


amaltaro:
@belforte do you need these changes for CRABServer or only TaskWorker? Just to know whether we need to provide you a new tag for testbed. I haven't looked yet in deep at the problem and your fix though...

belforte:
only for TaskWorker, server REST does not initiate pycurl connections. We can keep hacking a fix on our side for a bit more, no problem.

amaltaro:
@belforte I made a new PR trying to get a more robust check on the response header, including unit tests too. Please review #8900

belforte:
@amaltaro your PR is much better !

PR #8889, opened 2018-11-01
fix cache refresh

ticoann:
Because this bug cache som of the workflow info doesn't get to updated. tested in vm. Updating production agents.

### DIFF
--- a/src/python/WMCore/Services/WMStats/WMStatsWriter.py
+++ b/src/python/WMCore/Services/WMStats/WMStatsWriter.py
@@ -126,16 +126,16 @@ class WMStatsWriter(WMStatsReader):
                     magicStr = ".fnal.gov-"
                     idParts = doc['_id'].split(magicStr)
                     if len(idParts) == 2:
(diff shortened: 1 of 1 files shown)


ticoann:
[Fix #8889 - wmstats cache refresh fix](https://trello.com/c/MaEl3WVe/132-fix-8889-wmstats-cache-refresh-fix)

ticoann:
[Fix #8889 - wmstats cache refresh fix](https://trello.com/c/MaEl3WVe/132-fix-8889-wmstats-cache-refresh-fix)

ticoann:
updated production agents. twiki, trello is updated as well

amaltaro:
Stupid bug! I guess it means we should have this covered by unit tests :-)

ticoann:
Yes, all my fault as well. 
The damage is not that bad. It is only for the newly pulled workflows after patch is 
 upgraded and only for updates not initial insert. Also, when the component restarts it should be updating all the docs. 
Anyway, this is temp patch. All will be gone when the cmsweb is upgraded.

PR #8891, opened 2018-11-02
improve query speed by 30%

ticoann:
@amaltaro, Alan please review. Although this not changing the result of the data but it seems it improves the speed of query 30%. I thought it would be the other way, since previous query doesn't have to reduce. But it seems it is more of data size it is transferring than the computation work on reduce.  I think we should put this on this release since we are suffering from the performance of wmstats.

### DIFF
--- a/src/couchapps/WMStats/_attachments/js/Models/WMStats._RequestModelBase.js
+++ b/src/couchapps/WMStats/_attachments/js/Models/WMStats._RequestModelBase.js
@@ -84,7 +84,7 @@ WMStats._RequestModelBase.prototype = {
     },
 
     _getLatestRequestAgentUrlAndCreateTable: function (overviewData, keys, objPtr) {
-        var options = {"keys": keys, "reduce": false, "descending": true};
+        var options = {"keys": keys, "reduce": true, "group": true, "descending": true};
         WMStats.Couch.view('latestRequest', options,
               function(agentIDs) {
                   objPtr._getRequestDetailsAndTriggerEvent(agentIDs, overviewData, objPtr);
--- a/src/couchapps/WMStats/_attachments/js/minified/import-all-t0.min.js
+++ b/src/couchapps/WMStats/_attachments/js/minified/import-all-t0.min.js
@@ -4028,7 +4028,7 @@ WMStats._RequestModelBase.prototype = {
     },
 
     _getLatestRequestAgentUrlAndCreateTable: function (overviewData, keys, objPtr) {
-        var options = {"keys": keys, "reduce": false, "descending": true};
+        var options = {"keys": keys, "reduce": true, "group": true, "descending": true};
         WMStats.Couch.view('latestRequest', options,
               function(agentIDs) {
                   objPtr._getRequestDetailsAndTriggerEvent(agentIDs, overviewData, objPtr);
--- a/src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
+++ b/src/couchapps/WMStats/_attachments/js/minified/import-all-t1.min.js
@@ -4676,7 +4676,7 @@ WMStats._RequestModelBase.prototype = {
     },
 
     _getLatestRequestAgentUrlAndCreateTable: function (overviewData, keys, objPtr) {
-        var options = {"keys": keys, "reduce": false, "descending": true};
+        var options = {"keys": keys, "reduce": true, "group": true, "descending": true};
         WMStats.Couch.view('latestRequest', options,
               function(agentIDs) {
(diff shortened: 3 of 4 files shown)


ticoann:
So I ran few tests on this. I guess it depends on the network speed.
When I ran on the browser
https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStatsErl/_view/latestRequest?decending=false&reduce=false

took 18 sec with 3 MB data

https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStatsErl/_view/latestRequest?group=true&decending=true&reduce=true

and took 12 sec with 2MB data

But when I ran in vocms machine

first query took 1 sec
and second query 9 sec.

One thing I found is there are duplicate data in production with id

cmsgwms-submit7.fnal.gov-pdmvserv_task_EXO-RunIISummer16MiniAODv3-05478__v1_T_180920_073540_4259

https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStatsErl/_view/latestRequest?startkey=[%22pdmvserv_task_EXO-RunIISummer16MiniAODv3-05478__v1_T_180920_073540_4259%22]&endkey=[%22pdmvserv_task_EXO-RunIISummer16MiniAODv3-05478__v1_T_180920_073540_4259%22,%20{}]&reduce=false

ticoann:
The performance is depending on the network speed so actually without reduce it is faster but since the data size is bigger when reduce is false (which contains duplicate information (id)) total time is longer with slow network. 

https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStatsErl/_view/latestRequest?desending=true&reduce=false
It took 18 sec sec 3 MB data at home with Chrome

https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStatsErl/_view/latestRequest?group=true&decending=true&reduce=true
It took 12 sec with 2 MB data 

But when I ran the same query on vocms box, first query took 1 sec and second one 8 sec

I still think we need to update this patch. 

One thing I am worried is that in production database there is duplicate document. I am not sure why newer revision doesn't overwrite this. 

https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStatsErl/_view/latestRequest?startkey=%5B%22pdmvserv_task_EXO-RunIISummer16MiniAODv3-05478__v1_T_180920_073540_4259%22%5D&endkey=%5B%22pdmvserv_task_EXO-RunIISummer16MiniAODv3-05478__v1_T_180920_073540_4259%22,%20%7B%7D%5D&reduce=false

amaltaro:
Seangchan, I can't make much sense of your examples, your first and second query are linear, 6sec/MB no matter whether it has reduce true or false.

Is it only wmstats using this `_getLatestJobInfo` method? How it uses that info?

ticoann:
The data is contains exactly the same information. But with reduce = false in added duplicate information since it send id as well which is in value field.
```
{"id":"cmsgwms-submit7.fnal.gov-amaltaro_Run2018A-v1-DoubleMuon-17Sep2018_1024p1_181018_184027_2551","key":["amaltaro_Run2018A-v1-DoubleMuon-17Sep2018_1024p1_181018_184027_2551","cmsgwms-submit7.fnal.gov"],"value":{"id":"cmsgwms-submit7.fnal.gov-amaltaro_Run2018A-v1-DoubleMuon-17Sep2018_1024p1_181018_184027_2551","timestamp":1541176615}}
```
But reduce true, this is what it is transfered.

{"key":["amaltaro_Run2018A-v1-DoubleMuon-17Sep2018_1024p1_181018_184027_2551","cmsgwms-submit7.fnal.gov"],"value":{"id":"cmsgwms-submit7.fnal.gov-amaltaro_Run2018A-v1-DoubleMuon-17Sep2018_1024p1_181018_184027_2551","timestamp":1541176615}}

So it always transfer more data when we reduce false although computation is probably cheaper.
We can change the view itself "return null" and use reduce = false. But that is not backward compatible and error prone. (But if you prefer, we can change that way, I will try to run more through test).

ticoann:
I think #8892 is better fix but need more throw testing as I mentioned it above.

amaltaro:
IMO, we could either:
1) merge this one and cut a new tag for Lina for Monday morning (did you check whether there is any other code using that view and using the id/key instead of values?)
2) we wait until December upgrade and merge your other fix

ticoann:
Hi Alan, I think we should do 1. 
This fix is basically no change in previous code. So backward compatible. The other one is might affect some where else in the code (not just in WMCore). 
I will check whether there is any other things need be changed and create a new tag and make the pull request before Monday.

Thanks for reviewing.

PR #8892, opened 2018-11-02
improve performance in wmstats

ticoann:
This patch will reduce the cpu time or requestcache REST api,
1. not calling the reduce = true. (Since it is not doing reduce call time is 9 times faster - However, it is a small portion of whole call, data transfer time is the dominant factor. 
2. by emitting null instead of rev and time in value section, it reduces the data size.
3. reduces the 2 calls to one.

### DIFF
--- a/src/couchapps/WMStatsErl/views/latestRequest/map.erl
+++ b/src/couchapps/WMStatsErl/views/latestRequest/map.erl
@@ -5,9 +5,7 @@ fun({Doc}) ->
     <<"agent_request">> ->
       Workflow = couch_util:get_value(<<"workflow">>, Doc),
       AgentUrl = couch_util:get_value(<<"agent_url">>, Doc),
-      Id = couch_util:get_value(<<"_id">>, Doc),
-      Timestamp = couch_util:get_value(<<"timestamp">>, Doc),
-      Emit([Workflow, AgentUrl], {[{id,Id},{timestamp,Timestamp}]});
+      Emit([Workflow, AgentUrl], null);
     _ -> ok
   end
 end.
--- a/src/python/WMCore/Services/WMStats/WMStatsReader.py
+++ b/src/python/WMCore/Services/WMStats/WMStatsReader.py
@@ -115,8 +115,7 @@ class WMStatsReader(object):
 
         if len(requestNames) > 0:
             requestAndAgentKey = self._getRequestAndAgent(requestNames)
-            jobDocIds = self._getLatestJobInfo(requestAndAgentKey)
-            jobInfoByRequestAndAgent = self._getAllDocsByIDs(jobDocIds)
+            jobInfoByRequestAndAgent = self._getLatestJobInfo(requestAndAgentKey)
         return jobInfoByRequestAndAgent
 
     def _updateRequestInfoWithJobInfo(self, requestInfo):
@@ -221,12 +220,10 @@ class WMStatsReader(object):
         """
         if len(keys) == 0:
             return []
-        options = {}
-        options["reduce"] = True
-        options["group"] = True
+        options = {"include_docs": True}
+        options["reduce"] = False
         result = self._getCouchView("latestRequest", options, keys)
-        ids = [row['value']['id'] for row in result["rows"]]
-        return ids
+        return result
 
     def _getAllDocsByIDs(self, ids, include_docs=True):
         """
--- a/src/couchapps/WMStats/_attachments/js/Models/WMStats._RequestModelBase.js
+++ b/src/couchapps/WMStats/_attachments/js/Models/WMStats._RequestModelBase.js
@@ -84,7 +84,7 @@ WMStats._RequestModelBase.prototype = {
     },
 
     _getLatestRequestAgentUrlAndCreateTable: function (overviewData, keys, objPtr) {
-        var options = {"keys": keys, "reduce": true, "group": true, "descending": true};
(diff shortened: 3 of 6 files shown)


ticoann:
This can be further improved by reducing number of calls in wmstats.
i.e it doesn't have to get requestAgentURL call. I can simply get all the dos in agent_request type doc for wmstats but that would need a more changes.

ticoann:
@amaltaro, Alan, test came out fine. Could you review?

ticoann:
test this please

amaltaro:
@ticoann while I go through this PR, can you please update the description of this PR in the first comment above? Thanks

src/couchapps/WMStatsErl/views/latestRequest/map.erl
```diff
@@ -5,9 +5,7 @@ fun({Doc}) ->
     <<"agent_request">> ->
       Workflow = couch_util:get_value(<<"workflow">>, Doc),
       AgentUrl = couch_util:get_value(<<"agent_url">>, Doc),
-      Id = couch_util:get_value(<<"_id">>, Doc),
-      Timestamp = couch_util:get_value(<<"timestamp">>, Doc),
-      Emit([Workflow, AgentUrl], {[{id,Id},{timestamp,Timestamp}]});
+      Emit([Workflow, AgentUrl], null);
```
amaltaro: Is the `Workflow` really the workflow name here (blah)? Or it's the agent_url + workflow name (vocms0250.cern.ch-blah)?

src/python/WMCore/Services/WMStats/WMStatsReader.py
```diff
@@ -221,12 +220,10 @@ def _getLatestJobInfo(self, keys):
         """
         if len(keys) == 0:
             return []
-        options = {}
-        options["reduce"] = True
-        options["group"] = True
+        options = {"include_docs": True}
```
amaltaro: I was going to ask how can we transfer less data if one of the couch views is asked to return the full document, but I believe the reason is that it was already returned by the call to `_getAllDocsByIDs` that you removed, right?

src/python/WMCore/Services/WMStats/WMStatsReader.py
```diff
@@ -221,12 +220,10 @@ def _getLatestJobInfo(self, keys):
...
+        options["reduce"] = False
         result = self._getCouchView("latestRequest", options, keys)
-        ids = [row['value']['id'] for row in result["rows"]]
-        return ids
+        return result
```
amaltaro: for consistency with the other methods, shouldn't we return something like `result['rows']` instead? Or maybe parse it and return only the tuple requestname and agenturl, dropping the None value(?)

amaltaro (commented):
Seangchan, patch looks good in general and I see only 2 places using that view, so it should be fairly safe. Just for clarity, can you post the output of that view before and after these changes (just a a couple of lines)? Thanks

amaltaro:
test this please

amaltaro:
test this please

src/couchapps/WMStatsErl/views/latestRequest/map.erl
ticoann: It is just workflow name.

src/python/WMCore/Services/WMStats/WMStatsReader.py
ticoann: yes _getAllDocsByIDs using include_docs so the data volume is less by {'_rev', 'timestamp'}

src/python/WMCore/Services/WMStats/WMStatsReader.py
ticoann: _getCouchView handling that format. It has to be fed to other function. So format cannot me changed.

src/couchapps/WMStatsErl/views/latestRequest/map.erl
ticoann: The only deference of the view is value part new view has null, old view has id (workflow + agentURL) and timestamp
```
new view

{"total_rows":15,"offset":0,"rows":[
{"id":"vocms0193.cern.ch-sryu_DQMHarvesting_RunWhitelist_syu_181129_170448_3645","key":["sryu_DQMHarvesting_RunWhitelist_syu_181129_170448_3645","vocms0193.cern.ch"],"value":null},

old view

{"rows":[
{"key":["amaltaro_Run2018A-v1-EGamma-17Sep2018_1024p1_181018_183329_3950","cmsgwms-submit7.fnal.gov"],"value":{"id":"cmsgwms-submit7.fnal.gov-amaltaro_Run2018A-v1-EGamma-17Sep2018_1024p1_181018_183329_3950","timestamp":1543533540}},

```

src/python/WMCore/Services/WMStats/WMStatsReader.py
amaltaro: I don't see _getCouchView changing anything on the output. If it does, then why this function has this line for example
https://github.com/dmwm/WMCore/blob/b25964f7f5eaa57018b414f6273d9220957fa895/src/python/WMCore/Services/WMStats/WMStatsReader.py#L211
?

amaltaro:
I'm merging it and 8933 and putting it into pre3. If there is anything else to be fixed, then we can find it in testbed and make another request.

PR #8898, opened 2018-11-05
change rev to revision for initial revision

ticoann:
@amaltaro, Alan please review. As we discussed, I changed the rev to revision for initial setting as well.
Still need to run the test.


amaltaro:
Do we actually need to set it for new docs?

ticoann:
We do otherwise we have to call separately with new_edit flag = false.

amaltaro (approved):
Thanks, Seangchan. 
Changes look good, even though I still don't fully understand this feature and can't make sense of what's written in the page 244 of the couch documentation:
https://media.readthedocs.org/pdf/couchdb/1.6.1/couchdb.pdf

which explicitly says we need to provide _id and _rev values for bulk_docs update (but hey, where is _rev in all this?!?)

amaltaro:
Closing this one in favor of #8907

Issue #8516, opened 2018-03-15
Add the service certificate under AgentStatusWatcher watch


amaltaro:
The component has all the mechanism to monitor the proxy lifetime already, we should add the service certificate as well and get notifications (error!) when it's about to expire.

khurtado:
Fixed via #8899

PR #8899, opened 2018-11-06
Check for Certificate expiration time and create unit test for AgentStatusWatcher

### LINKED ISSUE #8879: Create AgentStatusWatcher unitest

khurtado:
Fixes  #8879 and #8516
Assumes `X509_USER_CERT` is the service certificate.
Right now, it's giving a warning at 10 days of expiration time and error at 3 days. Can be changed if different values are preferred.

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -45,6 +45,11 @@ class AgentStatusPoller(BaseWorkerThread):
         proxyArgs = {'logger': logging.getLogger()}
         self.proxy = Proxy(proxyArgs)
         self.proxyFile = self.proxy.getProxyFilename()  # X509_USER_PROXY
+        self.userCertFile = self.proxy.getUserCertFilename() # X509_USER_CERT
+        # credential lifetime warning/error thresholds, in days
+        self.credThresholds = {'proxy': {'error': 3, 'warning': 5},
+                               'certificate': {'error': 10, 'warning': 20}}
+
 
         localWQUrl = config.AnalyticsDataCollector.localQueueURL
         self.workqueueDS = WorkQueueDS(localWQUrl)
@@ -108,7 +113,8 @@ class AgentStatusPoller(BaseWorkerThread):
         """
         try:
             agentInfo = self.collectAgentInfo()
-            self.checkProxyLifetime(agentInfo)
+            self.checkCredLifetime(agentInfo, "proxy")
+            self.checkCredLifetime(agentInfo, "certificate")
 
             timeSpent, wmbsInfo, _ = self.collectWMBSInfo()
             wmbsInfo['total_query_time'] = int(timeSpent)
@@ -255,29 +261,41 @@ class AgentStatusPoller(BaseWorkerThread):
 
         return results
 
-    def checkProxyLifetime(self, agInfo):
+    def checkCredLifetime(self, agInfo, credType):
         """
-        Check the proxy lifetime (usually X509_USER_CERT) and raise either
-        a warning or an error if the proxy validity is about to expire.
+        Check the credential lifetime. Usually X509_USER_PROXY or X509_USER_CERT
+        and raise either a warning or an error if the proxy validity is about to expire.
         :param agInfo: dictionary with plenty of agent monitoring information in place.
+        :param credType: credential type, can be: "proxy" or "certificate"
         :return: same dictionary object plus additional keys/values if needed.
         """
-        secsLeft = self.proxy.getTimeLeft(proxy=self.proxyFile)
-        logging.debug("Proxy '%s' lifetime is %d secs", self.proxyFile, secsLeft)
+        if credType == "proxy":
+            credFile = self.proxyFile
+            secsLeft = self.proxy.getTimeLeft(proxy=credFile)
+        elif credType == "certificate":
+            credFile = self.userCertFile
+            secsLeft = self.proxy.getUserCertTimeLeft(openSSL=True)
+        else:
+            logging.error("Unknown credential type. Available options are: [proxy, certificate]")
+            return
 
-        if secsLeft <= 86400 * 3:  # 3 days
-            proxyWarning = True
+        logging.debug("%s '%s' lifetime is %d seconds", credType, credFile, secsLeft)
+
+        daysLeft = secsLeft / (60. * 60 * 24)
+
+        if daysLeft <= self.credThresholds[credType]['error']:
+            credWarning = True
             agInfo['status'] = "error"
-        elif secsLeft <= 86400 * 5:  # 5 days
-            proxyWarning = True
+        elif daysLeft <= self.credThresholds[credType]['warning']:
+            credWarning = True
             if agInfo['status'] == "ok":
                 agInfo['status'] = "warning"
         else:
-            proxyWarning = False
+            credWarning = False
 
-        if proxyWarning:
-            warnMsg = "Agent proxy '%s' must be renewed ASAP. " % self.proxyFile
-            warnMsg += "Its time left is: %.2f hours." % (secsLeft / 3600.)
-            agInfo['proxy_warning'] = warnMsg
+        if credWarning:
+            warnMsg = "Agent %s '%s' must be renewed ASAP. " % (credType, credFile)
+            warnMsg += "Its time left is: %.2f hours;" % (secsLeft / 3600.)
+            agInfo['proxy_warning'] = agInfo.get('proxy_warning', "") + warnMsg
 
         return
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -174,21 +174,31 @@ class Proxy(Credential):
     def cmd_exists(self, cmd):
         return subprocess.call(self.setEnv("type " + cmd), shell=True,
                                stdout=subprocess.PIPE, stderr=subprocess.PIPE) == 0
+    def getUserCertFilename(self):
+        if 'X509_USER_CERT' not in os.environ:
+            ucFilename = '~/.globus/usercert.pem'
+        else:
+            ucFilename = os.environ['X509_USER_CERT']
+        return ucFilename
 
-    def getUserCertEnddate(self, openSSL=True):
+    def getUserCertTimeLeft(self, openSSL=True):
         """
-        Return the number of days until the expiration of the user cert in .globus/usercert.pem or $X509_USER_CERT if set
+        Return the number of seconds until the expiration of the user cert
+        in .globus/usercert.pem or $X509_USER_CERT if set
         Uses openssl by default and fallback to voms-proxy-info in case of problems
         """
-        certLocation = '~/.globus/usercert.pem' if 'X509_USER_CERT' not in os.environ else os.environ['X509_USER_CERT']
+        certLocation = self.getUserCertFilename()
         if openSSL:
-            out, _, retcode = execute_command('openssl x509 -noout -in %s -dates' % certLocation, self.logger,
+            out, _, retcode = execute_command('openssl x509 -noout -in %s -dates'
+                                              % certLocation, self.logger,
                                               self.commandTimeout)
             if retcode == 0:
                 out = out.split('notAfter=')[1]
                 if out[-1] == '\n':
                     out = out[:-1]
-                possibleFormats = ['%b  %d  %H:%M:%S %Y %Z', '%b %d %H:%M:%S %Y %Z']
+
+                possibleFormats = ['%b  %d  %H:%M:%S %Y %Z',
+                                   '%b %d %H:%M:%S %Y %Z']
                 exptime = None
                 for frmt in possibleFormats:
                     try:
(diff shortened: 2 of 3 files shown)


vkuznet:
Why do we need to know expiration of X509_USER_CERT, shouldn't we always use X509_USER_PROXY?

khurtado:
@vkuznet I think since proxies are generated from certificates, knowing in advance when a certificate is about to expire (hence giving you time to get a new one) is better than realizing in-situ that proxies can't get extended anymore because of it.

src/python/WMCore/Credential/Proxy.py
```diff
@@ -174,13 +174,16 @@ def setEnv(self, cmd):
     def cmd_exists(self, cmd):
         return subprocess.call(self.setEnv("type " + cmd), shell=True,
                                stdout=subprocess.PIPE, stderr=subprocess.PIPE) == 0
+    def getUserCertFilename(self):
+        ucFilename = '~/.globus/usercert.pem' if 'X509_USER_CERT' not in os.environ else os.environ['X509_USER_CERT']
```
amaltaro: I'd suggest to expand it to multiple lines in order to make it more readable.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -281,3 +283,32 @@ def checkProxyLifetime(self, agInfo):
             agInfo['proxy_warning'] = warnMsg
 
         return
+
+    def checkUserCertLifetime(self, agInfo):
```
amaltaro: Kenyi, maybe we could reuse the same method as above (you can rename it to checkCredLifetime) and pass a parameter saying whether it's a proxy or certificate, which will be used for the error message too.

The threshold is wired in the code because I didn;t expect it to change too often, I think we can keep that. We could set something like at the very beginning
```
thresholds = {'proxy': {'error': 3, 'warn': 5},
                       'cert': {'error': 10, 'warn': 20}}
```

or something else that comes to your mind.
This line
`agInfo['proxy_warning'] = warnMsg`
has to become
`agInfo['proxy_warning'] += warnMsg`

to make sure we catch both messages in case both proxy and cert are about to expire.

amaltaro (changes requested):
I have made some comments in the code. If that method doesn't become too ugly, I think we should try to reuse as much as possible of that code.

About USER_PROXY and USER_CERT, Kenyi is almost right, however we do not create the proxy out of the service certificate (yet) because service certificates don't carry VO roles (besides the cms one). Thus we have the service certificate used to authenticate to other CMS services (mostly CMSWEB) and the user proxy is shipped with every single job, which allows them to write to the storage production namespace.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
khurtado: Will do, should I follow any convention of time units to show in debug? Right now, we are showing proxy in secs and cert in days.

khurtado: Done.

src/python/WMCore/Credential/Proxy.py
khurtado: Done.

src/python/WMCore/Credential/Proxy.py
```diff
@@ -174,13 +174,17 @@ def setEnv(self, cmd):
     def cmd_exists(self, cmd):
         return subprocess.call(self.setEnv("type " + cmd), shell=True,
                                stdout=subprocess.PIPE, stderr=subprocess.PIPE) == 0
+    def getUserCertFilename(self):
+        ucFilename = ('~/.globus/usercert.pem' if 'X509_USER_CERT'
```
amaltaro: Sorry Kenyi, I meant:
```
if 'X509_USER_CERT' not in os.environ:
    ucFilename = '~/.globus/usercert.pem'
else:
    ucFilename = os.environ['X509_USER_CERT']
return ucFilename
```

It's a very small method, so the "classic" if-else style looks better IMO.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -255,29 +257,44 @@ def collectWMBSInfo(self):
 
         return results
 
-    def checkProxyLifetime(self, agInfo):
+    def checkCredLifetime(self, agInfo, cred_type):
```
amaltaro: Can you rename it to camel case please (credType)?

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -255,29 +257,44 @@ def collectWMBSInfo(self):
...
-        a warning or an error if the proxy validity is about to expire.
+        Check the credential lifetime. Usually X509_USER_PROXY or X509_USER_CERT
+        and raise either a warning or an error if the proxy validity is about to expire.
         :param agInfo: dictionary with plenty of agent monitoring information in place.
+        :param cred_type: credential type, can be: "proxy" or "certificate"
         :return: same dictionary object plus additional keys/values if needed.
         """
-        secsLeft = self.proxy.getTimeLeft(proxy=self.proxyFile)
-        logging.debug("Proxy '%s' lifetime is %d secs", self.proxyFile, secsLeft)
+        thresholds = {'proxy': {'error': 3, 'warning': 5},
```
amaltaro: How about actually defining it in the __init__ method as an instance variable `self.proxyThres = ...`? Second line has a too many leading space.

amaltaro (changes requested):
Thanks, Kenyi.
I still have two minor comments that can be improved.

We try to stick with camel case for argument and variable name (even though there are still lots of inconsistency in the code). If you want to have a look at what we'd like to have in the project, check out this file:
https://github.com/dmwm/WMCore/blob/master/standards/.pylintrc#L91

then you can evaluate your changes (well, the whole file you changed) with:
pylint --rcfile=/Users/amaltaro/Pycharm/WMCore/standards/.pylintrc YOUR_FILE_NAME

I highly recommend running pylint before pushing files upstream (and if possible running a code reformat to make the code as clear as possible)

src/python/WMCore/Credential/Proxy.py
khurtado: Done

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
khurtado: Done.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
khurtado: Done

amaltaro:
Commits need to be squashed too.

khurtado:
@amaltaro Just squashed everything into 1 commit

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -255,29 +260,41 @@ def collectWMBSInfo(self):
...
-            proxyWarning = False
+            credWarning = False
 
-        if proxyWarning:
-            warnMsg = "Agent proxy '%s' must be renewed ASAP. " % self.proxyFile
+        if credWarning:
+            warnMsg = "Agent %s '%s' must be renewed ASAP." % (credType, credFile)
             warnMsg += "Its time left is: %.2f hours." % (secsLeft / 3600.)
-            agInfo['proxy_warning'] = warnMsg
+            agInfo['%s_warning' % credType] = warnMsg
```
amaltaro: This key has to remain `proxy_warning` even if the credential type is a certificate, otherwise this alert won't show up in wmstats (there are further javascript code handling it under the couchapps)

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
```diff
@@ -0,0 +1,103 @@
...
+import threading
+import unittest
+from WMQuality.TestInitCouchApp import TestInitCouchApp as TestInit
+
+from WMComponent.AgentStatusWatcher.AgentStatusPoller import AgentStatusPoller
+
+
+class AgentStatusWatcher_t(unittest.TestCase):
+    """
+    TestCase for WorkQueueManagerTest module
```
amaltaro: Hmm? :)

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    def setUp(self):
+        """
+        setup for test.
+        """
+        self.testInit = TestInit(__file__)
+        self.testInit.setLogging()
+        self.testInit.setDatabaseConnection()
+        #self.testInit.setSchema(customModules=["WMCore.WMBS"],
```
amaltaro: Can we delete it?

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
```diff
@@ -0,0 +1,103 @@
...
+        #                        useDefault=False)
+        #self.testInit.setSchema(customModules=["WMComponent.DBS3Buffer"],
+        #                        useDefault=False)
+        self.reqmonDBName = "wmstat_t"
+        self.localDBName = "wmstat_t_local"
+        self.testInit.setupCouch(self.reqmonDBName, "WMStats")
+        self.testInit.setupCouch(self.localDBName, "WMStats")
```
amaltaro: Is it WMStats or WMStatsAgent for the local agent? Maybe @ticoann  knows it better

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
```diff
@@ -0,0 +1,103 @@
...
+        config.Agent.contact = "test@test.com"
+        config.Agent.teamName = "testTeam"
+        config.Agent.agentName = "testAgentName"
+        config.Agent.useMsgService = False
+        config.Agent.useTrigger = False
+        config.Agent.useHeartbeat = False
+
+        config.component_("AgentStatusWatcher")
+        config.AgentStatusWatcher.namespace = "WMComponent.AgentStatusWatcher.AgentStatusWatcher"
+        config.AgentStatusWatcher.jsonFile = "/dev/null"
```
amaltaro: I need to clean up this jsonFile thing. It was one of those temporary things that end up becoming permanent..

amaltaro (changes requested):
Almost there Kenyi. I put some comments in the code.
I noticed you put test and src changes all in the same commit, it's fine for this PR, but try to separate them in future PRs please.

ticoann:
test this please

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
khurtado: Does `agInfo['proxy_warning']` always exist in the dictionary, or do I need to check for it? E.g.:
```
           if agInfo.get('proxy_warning', None):
                agInfo['proxy_warning'] += warnMsg
            else:
                agInfo['proxy_warning'] = warnMsg
```

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
khurtado: The commented lines you mean? If so, yes. Will do.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -45,6 +45,10 @@ def __init__(self, config):
         proxyArgs = {'logger': logging.getLogger()}
         self.proxy = Proxy(proxyArgs)
         self.proxyFile = self.proxy.getProxyFilename()  # X509_USER_PROXY
+        self.userCertFile = self.proxy.getUserCertFilename() # X509_USER_CERT
+        self.credThresholds = {'proxy': {'error': 3, 'warning': 5},
```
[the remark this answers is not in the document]
khurtado: Thanks, added as a comments

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
khurtado: Changed

test/python/WMComponent_t/AgentStatusWatcher_t/AgentStatusWatcher_t.py
khurtado: One of those copy-paste overlooks :)

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
khurtado: Done.

amaltaro:
@khurtado after changing the Proxy.py module to properly print the error message, this patch has a problem fetching the time left for a service certificate
```
2018-11-15 17:12:49,743:140454046529280:ERROR:Proxy:Command: "voms-proxy-info -file /data/certs/servicecert.pem -timeleft"
	 failed to execute with error: Proxy not found: Private key was not found in the PEM keystore (1 certificate(s) was (were) found).
```

voms-proxy-info cannot be used for certificates with pub/private key in different files, so you might have to create a new method there which fetches info for certificate instead of proxies.

This openssl command works (well, some parsing/conversion is still needed):
`openssl x509 -in /data/certs/servicecert.pem -enddate -noout`

Can you please update this PR? Let me know if you have any questions.

khurtado:
@amaltaro Done.

amaltaro:
Thanks Kenyi, this patch looks good now and tests were Ok. I'm merging/backporting and putting it in a new agent release for today.

amaltaro (approved):

amaltaro:
@belforte just a heads up for these changes to the Proxy module that CRAB relies quite heavily. AFAICS this change should be transparent for CRAB.

src/python/WMCore/Credential/Proxy.py
```diff
@@ -174,42 +174,63 @@ def setEnv(self, cmd):
...
 
         # uses this as a fallback
         timeleft = self.getTimeLeft(proxy=certLocation, checkVomsLife=False)
         if self.retcode:
             raise CredentialException('Cannot get user certificate remaining time with "voms-proxy-info"')
 
-        daystoexp = int(timeleft / (60 * 60 * 24))
+        return timeleft
+
+    def getUserCertEnddate(self, openSSL=True):
```
belforte: wouldn't it be more proper to call this method: getUserCertValidDays ? from and Enddate thing I'd expect the day validity end.

belforte:
Thanks. I checked and agree that changes are not in methods which CRAB calls atm

src/python/WMCore/Credential/Proxy.py
khurtado: @belforte: Makes sense. This is not a new method though, which is why I wanted to preserve the original name for backwards compatibility ( in case there is people depending on this function as-is already).

PR #8900, opened 2018-11-06
Handle HTTP in the pycurl response header

### LINKED ISSUE #8882: pycurl_manager fails when remote server header contains GRIDHTTP_PASSCODE

amaltaro:
Fixes #8882
Superseeds #8883

HTTP header/response format seems to be well defined, looking at:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages

The only real change is:
* Search for `HTTP/\d.\d \d{3}` (e.g. HTTP/1.1 200) at the very beginning of the header line.

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
-#-*- coding: ISO-8859-1 -*-
-#pylint: disable=R0913,W0702,R0914,R0912,R0201
+# -*- coding: ISO-8859-1 -*-
+# pylint: disable=R0913,W0702,R0914,R0912,R0201
 """
 File: pycurl_manager.py
 Author: Valentin Kuznetsov <vkuznet@gmail.com>
@@ -38,16 +38,15 @@ for row in data:
 """
 from __future__ import print_function
 
-# system modules
-import os
-import re
-import sys
 import cStringIO as StringIO
 import httplib
 import json
 import logging
-import urllib
+import os
+import re
 import subprocess
+import sys
+import urllib
 
 # python3
 if sys.version.startswith('3.'):
@@ -56,8 +55,10 @@ if sys.version.startswith('3.'):
 # 3d-party libraries
 import pycurl
 
+
 class ResponseHeader(object):
     """ResponseHeader parses HTTP response header"""
+
     def __init__(self, response):
         super(ResponseHeader, self).__init__()
         self.header = {}
@@ -73,7 +74,7 @@ class ResponseHeader(object):
 
         for row in response.split('\r'):
             row = row.replace('\n', '')
-            if  not row:
+            if not row:
                 continue
             if re.search(startRegex, row):
                 if re.search(continueRegex, row):
@@ -89,14 +90,16 @@ class ResponseHeader(object):
             except:
                 pass
 
+
 class RequestHandler(object):
     """
     RequestHandler provides APIs to fetch single/multiple
     URL requests based on pycurl library
     """
+
     def __init__(self, config=None, logger=None):
         super(RequestHandler, self).__init__()
-        if  not config:
+        if not config:
             config = {}
         self.nosignal = config.get('nosignal', 1)
         self.timeout = config.get('timeout', 30)
@@ -111,11 +114,11 @@ class RequestHandler(object):
             uses a different encoding depending on the HTTP verb
             (either json.dumps or urllib.urlencode)
         """
-        #data is already encoded, just return it
+        # data is already encoded, just return it
         if isinstance(params, basestring):
             return params
 
-        #data is not encoded, we need to do that
+        # data is not encoded, we need to do that
         if verb in ['GET', 'HEAD']:
             if params:
                 encoded_data = urllib.urlencode(params, doseq=doseq)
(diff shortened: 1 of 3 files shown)


amaltaro:
test this please

vkuznet (commented):
Looks good to me, but it would be better if you'll not mix style adjustments with actual fix of functionality, because it is harder to follow.

amaltaro:
Yes, you don't need to look at the second commit Valentin. We always push real changes into the 1st commit only, 2nd is for unit tests, pylint fixed, safe code refactoring. Thanks!

belforte:
thanks. Looks good to me. I closed #8883 Let it be superseeded

amaltaro:
test this please

amaltaro:
I don't get why this WMInit test keeps failing, none of these changes should affect that test.

amaltaro:
test this please

amaltaro:
@belforte is there any easy way you can test these changes (just to see whether it breaks something)?
We still don't use pycurl in WMCore services, so it's a bit problematic to get it properly tested (besides unit tests).

Another option is to merge it and give you a new tag soon, and you keep it in mind when testing a new tag. I'm fine either way.

belforte:
@amaltaro I forgot to mention that I put this  in in preprod CRAB yesterday
https://cms-logbook.cern.ch/elog/Analysis+Operations/2341
I believe that you can merge and tag whenever convenient.

ticoann (approved):

PR #8907, opened 2018-11-07
Use _revisions instead of _rev; create small chunks of docs to update

amaltaro:
Superseeds #8898 

Changes are:
* cherry-picked Seangchan's change from 8898, replacing _rev by _revision
* don't upload all docs in a single HTTP request, bu create small chunks of 100 docs per request (performance varies A LOT, from a few seconds to a few minutes)
* no need for the try/except block, all docs should be well formatted in a new agent
* creates a random revision id for the first time a document is created (to avoid issues if a doc gets deleted in central wmstats)

It appears to have a bad performance, we should keep an eye on this mode of upload and make further changes if needed.

### DIFF
--- a/src/python/WMCore/Services/WMStats/WMStatsWriter.py
+++ b/src/python/WMCore/Services/WMStats/WMStatsWriter.py
@@ -1,6 +1,10 @@
-import WMCore
+import logging
+import random
 import time
 from json import JSONEncoder
+
+import WMCore
+from Utils.IteratorTools import grouper
 from WMCore.Database.CMSCouch import CouchNotFoundError
 from WMCore.Services.WMStats.WMStatsReader import WMStatsReader
 
@@ -55,7 +59,6 @@ def convertToServiceCouchDoc(wqInfo, wqURL):
 
 
 class WMStatsWriter(WMStatsReader):
-
     def __init__(self, couchURL, appName="WMStats", reqdbURL=None, reqdbCouchApp="ReqMgr"):
         # set the connection for local couchDB call
         WMStatsReader.__init__(self, couchURL, appName, reqdbURL, reqdbCouchApp)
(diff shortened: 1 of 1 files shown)


ticoann:
This doesn't seem to work.  It seems for the initial doc, _rev has to be set

amaltaro:
test this please

ticoann:
test this please

ticoann:
Alan, how was the performance test went. (You said initial call take long time 18 sec for 32 requests)

ticoann:
This is for only new inserts, right? After the first insert is it faster (for updating records)?

amaltaro:
Yes, new inserts take quite a while. Only document updates are much faster.

ticoann (approved):

ticoann:
test this please

amaltaro:
Now that we have wiped out the wmstats database, AnalyticsDataCollector became 10x faster. 
Checking a CERN agent (0250), time to upload agent_request docs went down from ~250secs to 16secs (310 docs).
While at FNAL it went down from 250secs to 50secs (~430 docs).

amaltaro:
That was the execution time of the component cycle in general, so including everything. Still, I made an average (by eye) of the last 50 cycles or so.
Yes, DB is much healthier since we applied this patch yesterday, it grew only 13GB over the night.

ticoann:
13G? Before the compaction? Right now it seems it is 2.7G. I expect it keeps growing until we hit the revision limit. current revision is around ~30 revision.

amaltaro:
No no, it went from 332GB (yesterday afternoon) to 345GB (this morning). While the night before it grew ~150GB :)

ticoann:
I see.  That is after you patch the agent, right?

amaltaro:
Yes sir!

ticoann:
Thanks Alan, it looks a lot better than before. Hope it stays that way. :-)

PR #8890, opened 2018-11-02
Adding support for certificates

ddavila0:
With this, you can use certificates instead of username/password to send data

PR #8908, opened 2018-11-07
Adding support to use certificates on StompAMQ

ddavila0:
Superseeds #8890

### DIFF
--- a/src/python/WMCore/Services/StompAMQ/StompAMQ.py
+++ b/src/python/WMCore/Services/StompAMQ/StompAMQ.py
@@ -61,19 +61,24 @@ class StompAMQ(object):
     :param topic: The topic to be used on the broker
     :param host_and_ports: The hosts and ports list of the brokers.
         E.g.: [('agileinf-mb.cern.ch', 61213)]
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

ticoann:
test this please

amaltaro:
I think the tests won't be triggered unless Diego gets added to one of the WMCore/DMWM teams.
Since it's a one time PR, I guess we can just skip that, this module doesn't have any unit tests anyways...

I ran a pylint and code looks good. We can merge it and proceed with further testing afterwards.

ticoann:
Is username and password is not mandatory parameter anymore?

ticoann:
I am merging it but I think we should set the default values if the parameters are optional and validate in the function. (either cert/key or user/pass  has to be set - both cannot be empty)

amaltaro:
Sigh, I was going to reply that that was a good point :-) We could make it all kwargs, this is the only usage in WMCore:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/HeartbeatMonitorBase.py#L67

PR #8935, opened 2018-11-30
WMAgent and WMCore pypi packaging

### LINKED ISSUE #7647: Package WMCore into pypi
I was wanting to float the idea of pushing WMCore into pypi so it can be easily be pulled into other projects. Setuptools already handles it via `python setup.py upload`, but that requires credentials, whch kinda makes `bin/buildrelease.sh` not the right place to put it, since whoever built a release would need to have PyPi credentials.

The other option is to add the hooks into the rpm bot or Jenkins to publish new releases. That's a bit easier since those hosts are secured and could keep the credentials secret.

Anyway, it'd be useful to have WMCore (well, a lot of DMWM's code, honestly) in pypi. If there's no objection, I'd like to put the hooks in to get it pushed up automagically

goughes:
Fixes #7647
These are the changes I have made to be able to build the WMAgent pypi package.  

Documentation updates regarding pypi packaging will be made on this wiki page:
https://github.com/dmwm/WMCore/wiki/DMWM-Packaging-with-PyPi

There are some caveats regarding setup.py that I will explain.

~~The original setup.py (used to build the wmagent rpm) has been renamed to setup_wmcore.py~~
~~A new file setup_wmagent.py was created that is used to build the wmagent pypi package.~~

~~To build the pypi package, I create a symlink setup.py -> setup_wmagent.py~~
~~To build the rpm, the symlink can be changed setup.py -> setup_wmcore.py~~

Regarding versioning, pypi packages need to follow PEP 440: https://www.python.org/dev/peps/pep-0440/
WMCore uses versioning like "1.1.15.pre6", which will get normalized during the build process.

`UserWarning: Normalizing '1.1.15.pre8' to '1.1.15rc8'`

### DIFF
--- a/setup_wmagent.py
+++ b/setup_wmagent.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+# This setup script is used to build the WMAgent pypi package.
+# The version number comes from WMCore/__init__.py and needs to
+# follow PEP 440 conventions
+
+from __future__ import print_function, division
+import os
+import sys
+import imp
+from setuptools import setup
+from setup_build import list_packages, list_static_files, get_path_to_wmcore_root
+
+# Obnoxiously, there's a dependency cycle when building packages. We'd like
+# to simply get the current WMCore version by using
+# from WMCore import __version__
+# But PYTHONPATH isn't set until after the package is built, so we can't
+# depend on the python module resolution behavior to load the version.
+# Instead, we use the imp module to load the source file directly by
+# filename.
+wmcore_root = get_path_to_wmcore_root()
+wmcore_package = imp.load_source('temp_module', os.path.join(wmcore_root,
+                                                             'src',
+                                                             'python',
+                                                             'WMCore',
+                                                             '__init__.py'))
+wmcore_version = wmcore_package.__version__
+
+# Requirements file for pip dependencies
+requirements = "requirements.txt"
+
+
+def parse_requirements(requirements_file):
+    """
+      Create a list for the 'install_requires' component of the setup function
+      by parsing a requirements file
+    """
+
+    if os.path.exists(requirements_file):
+        # return a list that contains each line of the requirements file
+        return open(requirements_file, 'r').read().splitlines()
+    else:
+        print("ERROR: requirements file " + requirements_file + " not found.")
+        sys.exit(1)
+
+
+setup(name='wmagent',
+      version=wmcore_version,
+      maintainer='CMS DMWM Group',
+      maintainer_email='hn-cms-dmDevelopment@cern.ch',
+      package_dir={'': 'src/python/'},
+      packages=list_packages(['src/python/Utils',
+                              'src/python/WMCore',
+                              'src/python/WMComponent',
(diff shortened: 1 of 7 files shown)


src/python/WMCore/__init__.py
```diff
@@ -6,5 +6,5 @@
 
 """
 
-__version__ = '1.1.15.pre6'
+__version__ = '1.1.15.8'
```
ticoann: Where is this version number comes from?

goughes: I need to switch that back to .pre6.  You have to increment the version every time you upload to pypi.  The change isn't necessary here.

amaltaro:
@goughes Erik, apparently there is a conflict with this PR.
Now that Yuyi merged the DBS PRs, can you please run another wmcore/wmagent test with 1.2.1 and make sure everything works fine? Thanks

goughes:
test this please

setup_wmagent.py
```diff
@@ -0,0 +1,61 @@
...
+    else:
+        print("ERROR: requirements file " + requirements_file + " not found.")
+        sys.exit(1)
+
+
+setup(name='wmagent',
+      version=wmcore_version,
+      maintainer='CMS DWMWM Group',
+      maintainer_email='hn-cms-dmDevelopment@cern.ch',
+      package_dir={'':'src/python/'},
```
amaltaro: add a blank space after `:` please (same for the file below)

amaltaro (changes requested):
Erik, changes look good in general. 
However, it fails the python3 check. Besides the `print_function` import, can you please import `division` as well? Jenkins expect `from __future__ import division` in every new python file.

amaltaro:
test this please

amaltaro:
@goughes I just noticed we ended up with 3 requirements txt files now. Sorry if I missed it, but couldn't we use `requirements.txt` in place of `requirements.wmagent.txt`? AFAICS they have the same list.

In addition to that, maybe the wmcore one should have all the wmagent dependencies plus whatever we have in central services spec. What do you think?

goughes:
I thought we were going to keep `requirements.txt` as a full list of requirements for any package that can be built from WMCore (wmagent, reqmgr2, etc.) mainly for documentation purposes. The list of dependencies in `requirements.txt` will grow once we add the requirements for the other packages and diverge from the wmagent one. The more I think about it, I suppose we could even rename `requirements.txt` to something else or potentially remove it altogether.

For the wmcore requirements file, including all dependencies creates problems. For instance, cx-Oracle requires `ORACLE_HOME` to be set and MySQL-Python requires `mysql_config` to be present. So those really should just be in the wmagent requirements file.

amaltaro:
I believe the GH alerts are sent based on the `requirements.txt`, so if we rename or remove it, we might miss the vulnerabilities-wise automatic alerts sent by GH. Are you saying then that:
* requirements.txt: could be a list with all WMCore python dependencies
* requirements_wmagent.txt: list of WMAgent related dependencies
* requirements_wmcore.txt: no requirements at all
?

PR #8938, opened 2018-12-03
move to archived status only depends on the time

### LINKED ISSUE #8894: Transition from aborted-completed to aborted-archived might not happen
With the recent changes made to `agent_summary` (AgentJobInfo in wmstats) and the removal of central cleanup of old wmstats docs, we might have a scenario where a workflow is aborted but the agent never cleans that data (e.g. agent was shutdown/redeployed), thus holding that workflow from being archived.

My suggestion is to proceed as we do with announced to normal-archived transition, use a grace period (maybe 2 days) and archive it whether it still has `AgentJobInfo` or not.

ticoann:
fixes #8894 

Since AgentJobInfo document won't be cleaned anymore, move do archive status only depend on the time. This is not the permanent fix. 
(We need to add the clean up status back to wmstats for each agent) 
But for now this will be OK with increased archive delay for 7 days.

update the delay hours below
https://github.com/dmwm/deployment/pull/703


amaltaro:
Thanks Seangchan. Yes, let's go with 7 days and keep an eye out and ask Lina to update the configuration if needed. I'm cutting a new release, thanks!

PR #8923, opened 2018-11-23
Make some WMAgent/WorkQueue metrics MonIT-friendly

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/8480 and #8928
Requires deployment changes: https://github.com/dmwm/deployment/pull/696
Changes are:
* updated views removing backward compatible changes done in #7554
* no longer dump those monitoring metrics in a local json file
* make sure some keys like job type, schedd status, workqueue element status are always present in the output (for both WMStats and MonIT)
* moved BossAir DAO under BossAir instead of WMBS (RunJobByStatus)
* changed the data structure for a few metrics uploaded to MonIT (affecting global_workqueue too)
* changed the data structure for possible and unique jobs
* take AMQ credentials from the agent secrets file
* fetch worker thread info from wm_workers and push those to MonIT
* renamed WMAgent document `type` pushed to MonIT
* removed reqmgr (one) bits from the deployment script

TODO: write a nice wiki with every single doc type in MonIT and the meaning of each field.

For the record, an empty agent creates around 125 docs to be uploaded to AMQ every cycle.

PR #8940, opened 2018-12-05
Rename a few MonIT document type

amaltaro:
Complement to #8923 

Changes are:
* make sure a few keys are always present in the doc `wma_site_info`
* new doc type `wma_wmbs_state_info` for wmbsCountByState information (to avoid different structure with the same doc type)
* new doc type `wma_summary_info` for the agent summary info
* removed `timestamp` from the payload to keep the same document structure as used in ReqMgr2 and Global WorkQueue


amaltaro:
Quickly tested in vocms0192 and vocms0261 and it looks Ok.
I've also cherry-picked this patch in the 1.1.18_wmagent PR.

ticoann (approved):

PR #8954, opened 2018-12-18
allow other compressions beside gzip for tarball

belforte:
Alan, Seanchan, please consider if you like this PR.
It has been pointed out that CRAB tarball can get significantly smaller by usiing gzip2 instead of gzip. Since modern tar self-detect the compression algorithm, removing 'z' option from untar in here allows to use gzip2 transparently.
ref: https://github.com/dmwm/CRABServer/issues/5729
No hurry to get it in, though.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py
+++ b/src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py
@@ -89,7 +89,7 @@ def UnpackUserTarball():
             logging.info("Fetching tarball %s through xrootd", tarball)
             try:
                 subprocess.check_call(['xrdcp', '-d', '1', '-f', tarball, 'TEMP_TARBALL.tgz'])
-                subprocess.check_call(['tar', 'xzf', 'TEMP_TARBALL.tgz'])
+                subprocess.check_call(['tar', 'xf', 'TEMP_TARBALL.tgz'])
             except subprocess.CalledProcessError:
                 logging.error("Couldn't retrieve/extract file from xrootd")
                 raise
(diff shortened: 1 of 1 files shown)


ticoann:
@belforte, Thank you Stefano. I didn't know that.

belforte:
@ticoann I think the idea is to allow the client to use what is best in a specific use case. For CRAB we need to limit user sandboxes to 100MB and that is tight at times, while we never have problems with tar/untar time, and all jobs have min. 2GB RAM available when they start and need to untar. So for us size is much more relevant. Maybe WMA ha other use cases for this script where it can stick to gzip.

belforte:
@ticoann I amnot sure why you assigned to me. Is that to solicit the answer above, or do you expect me to merge ? I think it is best that Alan is in control of merging.

ticoann:
Thanks Stefano for explanation. So you are saying removing 'z' option will pick compression algorithm automatically? How would you know which one it would pick? Before I think z option is tied to gzip and j option is tied to bzip2. 

Yes, only Alan should merge the request (assigned to Alan).

belforte:
thanks for the benchmark page, I was relying on one user reporting up to x2 smaller sandbox with bzip2 in his situation. I guess that a log depends on the input file(s). I could even let the user pick the compression algorithm so they can experiment which one is smaller.

belforte:
indeed tar detects compression algorightm automatically, regardless of file name, even if man page does not say that clearly. See e.g.
http://www.manpagez.com/info/tar/tar-1.23/tar_128.php
(same text is also in `info tar`, but harder to find)
You can also try `tar tf` on your favorite archive.

-z and -j options still work, the will force the specific algorithm and raise error if a different one is detected in the archive file

amaltaro:
If the compression/decompression time difference wasn't that big (almost 10x bigger), we could even adopt it for the WMCore tarball as well.

belforte:
if the time spent in (un)compressing WMCore tarball has any visible impact, I'd worry no matter what !

amaltaro:
I ran a few test workflows and apparently the userTarball and userFiles are not set in the production jobs. So I'm merging it.

amaltaro (approved):

amaltaro:
In case you want to start playing with it, a tag has been created: 1.1.20.pre1

belforte:
thanks. WIll let you know now it works

belforte:
works as expected. Thanks.

Issue #8916, opened 2018-11-14
DN tracking in change of priority


vlimant:
https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_task_MUO-RunIIAutumn18DR-00005__v1_T_181024_180950_4708

Dima was reporting that the change in priority looked weird.

    110000: 2018-10-24 16:09:50 UTC DN: /DC=ch/DC=cern/OU=computers/CN=pdmvserv/vocms081.cern.ch
    85000: 2018-11-03 16:13:08 UTC DN: /DC=ch/DC=cern/OU=computers/CN=dmwm/cmsweb.cern.ch
    110000: 2018-11-14 10:48:25 UTC DN: /DC=ch/DC=cern/OU=computers/CN=pdmvserv/vocms081.cern.ch

 110000: 2018-11-14 10:48:25 is me changing the priority from the webUI https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_task_MUO-RunIIAutumn18DR-00005__v1_T_181024_180950_4708 

85000: 2018-11-03 16:13:08 is Paolo changing the priority

https://cms-pdmv.cern.ch/mcm/requests?prepid=MUO-RunIIAutumn18DR-00005&page=0&shown=33

Update | 2018-10-30-21-59 | Paolo Gunnellini |  
-- | -- | -- | --
priority | 2018-11-03-17-13 | Paolo Gunnellini | 85000

so something is wrong in recording the DN of who changed the priority

PR #8967, opened 2019-01-16
Set the correct DN for priority transition

### LINKED ISSUE #8778: No DN recorded for priority change from new to new status
Via web UI, without any state change. If the request is in New state, then the box with Priority Transition will not fetch the correct DN, see request (prio 220k)
https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=amaltaro_ACDC_StepChain_ReDigi3_HG1809_Validation_180901_234933_6208

amaltaro:
Fixes #8778 and #8916

We were persisting `DN` to the workload_cache request, and that isn't an allowed attribute.

### DIFF
--- a/src/couchapps/ReqMgr/updates/updaterequest.js
+++ b/src/couchapps/ReqMgr/updates/updaterequest.js
@@ -5,7 +5,7 @@ function(doc, req) {
     	return [null, "Error: document not found"];
     };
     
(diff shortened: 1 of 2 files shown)


amaltaro:
Already ran some basic tests and changes look Ok.

PR #8969, opened 2019-01-16
Fix count of leftover processes in deploy-wmagent

### LINKED ISSUE #8942: 1.1.19.pre5 wmagent deploy script overcounts leftover processes
The following line counts for leftover WMAgent processes before moving forward:
https://github.com/dmwm/WMCore/blob/1.1.19.pre5/deploy/deploy-wmagent.sh#L150

The problem is, this will always produce a result greater than 1 in my tests, because it's grepping for wmagent(matching the deploy-wmagent.sh script itself, and the grep command being executed with the script):

``` 
cmst1    3027742  0.0  0.0 113288  1608 pts/0    S+   22:31   0:00 sh deploy-wmagent.sh -w 1.1.19.pre5 -d HG1811a -t testbed-dev -r comp=comp -c kwmcore.cern.ch
cmst1    3027757  0.0  0.0 112708   960 pts/0    S+   22:31   0:00 grep -E couch|wmcore|mysql|beam
```

For now, I'm using
output=`ps aux | egrep 'couch|wmcore|mysql|beam' |  grep -v deploy-wmagent.sh | wc -l`
instead. Does that look good?

khurtado:
Fix count of leftover processes in deploy-wmagent.sh script in the case one of the argument strings (e.g the machine hostname) for the deployment script itself matches one of the process names we are trying to count (i.e.: couch, wmcore, mysql, beam)
 
Fixes #8942

### DIFF
--- a/deploy/deploy-wmagent.sh
+++ b/deploy/deploy-wmagent.sh
@@ -147,7 +147,7 @@ check_certs()
 check_process()
 {
   echo -n "Checking whether there are any leftover processes ..."
-  output=`ps aux | egrep 'couch|wmcore|mysql|beam' | wc -l`
+  output=`ps aux | egrep 'couch|wmcore|mysql|beam' | grep -v deploy-wmagent.sh | wc -l`
   if [ "$output" -gt 1 ]; then
     echo "  FAILED!\n There are still $output WMCore process running. Quitting!"
     exit 8


amaltaro (approved):
Thanks, Kenyi.

PR #8973, opened 2019-01-21
Support ACDC for recovery workflows

### LINKED ISSUE #8245: ACDC of recovery workflows
https://cmsweb.cern.ch/reqmgr2/fetch?rid=vlimant_ACDC0_recovery-1-fabozzi_Run2016H-v1-BTagCSV-07Aug17_8029__171013_133417_8968

went in failed. it's an ACDC of a recovery.
how hard will it be to make it work ?
FYI @prozober

amaltaro:
Fixes #8245 

Standard ACDC workflows don't explicitly set `CollectionName`, it's actually just set to the parent workload name (the workflow we're creating an ACDC for). While recovery ACDCs do set it, since they create a brand new CollectionName in the ACDCServer. The moment that recovery workflow is used for creating a standard ACDC, it inherits the previous CollectionName and we're screwed, given that the new CollectionName is supposed to be the name of the parent workload.

After writing the paragraph above, it all sounds very confusing, but we have what we have ;)

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -116,6 +116,9 @@ def validate_resubmission_create_args(request_args, config, reqmgr_db_service, *
     response = reqmgr_db_service.getRequestByNames(request_args["OriginalRequestName"])
     originalArgs = response.values()[0]
 
+    ### not a nice fix for #8245, but we cannot inherit the CollectionName attr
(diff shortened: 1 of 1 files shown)


vlimant:
how have you tested this ? looks like a good step forward for @sharad1126 day-to-day operation

amaltaro:
I ran a normal ReReco workflow, then created a recovery ACDC, then a standard ACDC on top of the recovery. Final tests are still needed during the testbed validation phase though.

PR #8974, opened 2019-01-21
StompAMQ changes to integrate with different versions

stiegerb:
- Renamed `make_notification` to `make_notification_list` without changes.
- Added new `make_notification` method that produces a single notification.
- Adding new functionality to pass user-defined metadata.
- No longer add unnecessary `'payload'` subfield to notification body.

We can discuss here how best to proceed. In my opinion, `make_notification_list` is still confusing, as it assigns the same document id, type, and timestamp to all notifications. As in this PR now, it's also not really consistent with the other method, as it still uses the `'payload'` subfield.

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
@@ -550,8 +550,8 @@ class AgentStatusPoller(BaseWorkerThread):
                                 host_and_ports=self.hostPortAMQ,
                                 logger=logging)
 
-            notifications = stompSvc.make_notification(payload=docs, docType=docType,
-                                                       docId=producer, ts=timeS)
+            notifications = [stompSvc.make_notification(payload=doc, docType=docType,
+                                                        ts=timeS) for doc in docs]
 
             failures = stompSvc.send(notifications)
             logging.info("%i docs successfully sent to AMQ", len(notifications) - len(failures))
--- a/src/python/WMCore/Services/StompAMQ/StompAMQ.py
+++ b/src/python/WMCore/Services/StompAMQ/StompAMQ.py
@@ -66,7 +66,7 @@ class StompAMQ(object):
     """
 
     # Version number to be added in header
-    _version = '0.2'
+    _version = '0.3'
 
     def __init__(self, username, password, producer, topic,
                  host_and_ports=None, logger=None, cert=None, key=None):
@@ -91,11 +91,11 @@ class StompAMQ(object):
         :param data: Either a single notification (as returned by
             `make_notification`) or a list of such.
 
-        :return: a list of successfully sent notification bodies
+        :return: a list of notification bodies that failed to send
         """
-        if not isinstance(data, list):
-            self.logger.error("Argument for send method has to be a list, not %s", type(data))
-            return data
+        # If only a single notification, put it in a list
+        if isinstance(data, dict) and 'body' in data:
+            data = [data]
 
         conn = stomp.Connection(host_and_ports=self._host_and_ports)
 
@@ -152,40 +152,55 @@ class StompAMQ(object):
             return body
         return
 
-    def make_notification(self, payload, docType, docId, producer=None, ts=None):
+    def make_notification(self, payload, docType, docId=None, producer=None, ts=None, metadata=None,
+                          dataSubfield="data"):
         """
-        Given a single payload (or a list of them), generate a list
-        of notifications including the specified data.
+        Produce a notification from a single payload, adding the necessary
+        headers and metadata. Generic metadata is generated to include a
+        timestamp, producer name, document id, and a unique id. User can
+        pass additional metadata which updates the generic metadata.
 
-        :param payload: Actual notification data.
-        :param docType: document type for the high level metadata.
-        :param docId: document id representing the notification.
-        :param producer: The notification producer.
-        :param ts: timestamp to be added to each document metadata.
+        If payload already contains a metadata field, it is overwritten.
 
-        :return: a list of notifications with the proper metadata
+        :param payload: Actual data.
+        :param docType: document type for metadata.
+        :param docId: document id representing the notification. If none provided,
+               a unique id is created.
+        :param producer: The notification producer name, taken from the StompAMQ
+               instance producer name by default.
+        :param ts: timestamp to be added to metadata. Set as time.time() by default
+        :param metadata: dictionary of user metadata to be added. (Updates generic
+               metadata.)
+        :param dataSubfield: field name to use for the actual data. If none, the data
+               is put directly in the body. Default is "data"
+
+        :return: a single notifications with the proper headers and metadata
         """
         producer = producer or self._producer
+        umetadata = metadata or {}
         ts = ts or int(time.time())
+        uuid = makeUUID()
+        docId = docId or uuid
 
-        if isinstance(payload, dict):
-            payload = [payload]  # it was a single document
+        headers = {'type': docType,
+                   'version': self._version,
+                   'producer': producer}
 
-        commonHeaders = {'type': docType,
-                         'version': self._version,
-                         'producer': producer}
+        metadata = {'timestamp': ts,
+                    'producer': producer,
+                    '_id': docId,
+                    'uuid': uuid}
+        metadata.update(umetadata)
 
-        docs = []
-        for doc in payload:
-            notification = {}
-            notification.update(commonHeaders)
-            # Add body consisting of the payload and metadata
-            body = {'payload': doc,
-                    'metadata': {'timestamp': ts,
-                                 'id': docId,
-                                 'uuid': makeUUID()}
-                   }
-            notification['body'] = body
-            docs.append(notification)
+        body = {}
+        if dataSubfield:
+            body[dataSubfield] = payload
+        else:
+            body.update(payload)
+        body['metadata'] = metadata
 
-        return docs
+        notification = {}
+        notification.update(headers)
+        notification['body'] = body
+
+        return notification
--- a/src/python/WMCore/REST/HeartbeatMonitorBase.py
+++ b/src/python/WMCore/REST/HeartbeatMonitorBase.py
@@ -71,8 +71,8 @@ class HeartbeatMonitorBase(CherryPyPeriodicTask):
                                 host_and_ports=self.hostPortAMQ,
                                 logger=self.logger)
 
-            notifications = stompSvc.make_notification(payload=docs, docType=self.docTypeAMQ,
-                                                       docId=producer, ts=ts)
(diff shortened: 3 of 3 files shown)


vkuznet:
I would suggest to keep payload for backward compatibility but make it as optional function argument (and be last argument). If provided then you can structure the body of the doc to have payload otherwise leave as is.

leggerf:
+1

vkuznet:
+1

src/python/WMCore/Services/StompAMQ/StompAMQ.py
```diff
@@ -189,3 +189,51 @@ def make_notification(self, payload, docType, docId, producer=None, ts=None):
...
+        :param payload: Actual data.
+        :param docType: document type for metadata.
+        :param docId: document id representing the notification.
+        :param producer: The notification producer name.
+        :param ts: timestamp to be added to metadata.
+        :param metadata: user metadata to be added. (Updates generic metadata.)
+        :param payload_subfield: Toggle whether to put the payload in a subfield
+               or not
+
+        :return: a list of notifications with the proper metadata
```
amaltaro: "a list with a single notification"

src/python/WMCore/Services/StompAMQ/StompAMQ.py
```diff
@@ -189,3 +189,51 @@ def make_notification(self, payload, docType, docId, producer=None, ts=None):
...
+        pass additional metadata which updates the generic metadata.
+
+        If payload already contains a metadata field, it is overwritten.
+
+        :param payload: Actual data.
+        :param docType: document type for metadata.
+        :param docId: document id representing the notification.
+        :param producer: The notification producer name.
+        :param ts: timestamp to be added to metadata.
+        :param metadata: user metadata to be added. (Updates generic metadata.)
```
amaltaro: maybe say it's a dictionary with user metadata.

src/python/WMCore/Services/StompAMQ/StompAMQ.py
```diff
@@ -189,3 +189,51 @@ def make_notification(self, payload, docType, docId, producer=None, ts=None):
...
+        ts = ts or int(time.time())
+
+        headers = {'type': docType,
+                   'version': self._version,
+                   'producer': producer}
+
+        metadata = {'timestamp': ts,
+                    'producer': producer,
+                    'id': docId,
+                    'uuid': makeUUID()}
```
amaltaro: given that `id` is already meant to be unique, I wonder whether we actually have any need for `uuid`?

amaltaro (changes requested):
Patch looks good in general. I made a few comments in the code and in addition to that, how about we bump the module version to 0.3?
https://github.com/dmwm/WMCore/blob/af82fc85a83e5f1eaaaad472f089cf52779e0fb0/src/python/WMCore/Services/StompAMQ/StompAMQ.py#L69

Thanks for creating this PR, Benjamin.

vkuznet:
One more comment. The code now contains both CamelLike and python_like naming convention. Can we converge of using underscore (python way) for all names/variables? For instance, a function `make_notification` takes `docType` and `payload_subfield`.

amaltaro:
Actually we adopted the camelCase convention in WMCore, but as you pointed out, there is a large fraction of the code using both convention...
https://github.com/dmwm/WMCore/blob/master/standards/.pylintrc#L91

Benjamin, there is no need to update all the code (unless we really want to), but please do update the code you're touching ;)

src/python/WMCore/Services/StompAMQ/StompAMQ.py
amaltaro: It doesn't make much difference for me either. @vkuznet @leggerf shall we keep uuid?

amaltaro:
Thanks, Benjamin! Just one more request from my side, can you please squash those commits in a single one? Thanks

leggerf:
I’d say let’s keep it. It doesn’t hurt

stiegerb:
Ok, squashed into one.

stiegerb:
There was still a typo, fixed now, and tested.

vkuznet:
Regarding docId vs uuid. This code does not provide any guarantee that docId will be unique since it is external parameter, i.e. it is up to applications which will call this module. Therefore my suggestion would be to keep uuid just in case because we are making it unique and therefore guarantee that submitted doc will have at least some unique id (either id or uuid).

stiegerb:
One more thing I noticed: we changed the name of the "id" field in the metadata to be "_id". Let me check with MONIT whether that's necessary for it to be picked up by them.

stiegerb:
@vkuznet yes, let's leave it for now. Although MONIT assigns a unique id to any incoming message anyway, so I'm not sure how useful it will be if we don't actually keep track of them anyway.

stiegerb:
Update from MONIT:
- If they find a field called "_id" or "metadata._id" in the data, they do **not** create a unique id. So maybe it would be good to keep ours, just in case.
- Indeed, id fields should be "_id", not "id". I'll update the PR to do so.
- The following list of fields are digested and promoted to metadata in the MONIT message. This means that if we want one of those fields to show up as a regular "data.x" field, we need to put it in a "data" subfield first. Otherwise it will be swallowed:
   - _id
   - availability_zone
   - environment
   - event_timestamp
   - host
   - hostgroup
   - hostname
   - json
   - path
   - producer
   - submitter_environment
   - submitter_hostgroup
   - toplevel_hostgroup
   - timestamp
   - timestamp_format
   - type
   - type_prefix
   - version

So it might be a good idea to keep our "payload" subfield, just renaming it to "data". We can make it configurable, so we can still use "payload" for backwards compatibility.

leggerf:
As it is, I don’t see any use for make_notification_list. What’s the argument to keep it?

amaltaro:
Argument against removing `make_notification_list` is:
* we might have service/nodes metrics that we want to push to MonIT where the document id itself does not matter; we'll never do anything with that same document besides making a time series plot (this is the case for WMAgent/ReqMgr2/Global Workqueue)
* in case of hundreds of thousands or millions of documents, calling `make_notification` for every single document increases the wallclock time by almost two orders of magnitude. You can check it yourself with this example:
https://github.com/amaltaro/ProductionTools/blob/master/timeFunctionCalls.py

`make_notification_list` method has two problems that I acknowledge. a) `producer` is actually overwritten somewhere in the MonIT ecosystem(?) and b) `docId`is wrong and we should replace `id` key/value by `_id: makeUUID()`.

However, if we want to make the code really clean, I think we can remove that method. The services I have pushing data to MonIT don't create more than 5k notifications every cycle, so I expect a performance impact closer to 0.

stiegerb:
@amaltaro In your example you're creating the data inside the loop, and calling time.time() every time. That's why it's so slow. If I rewrite it to just pass the data in the argument, calling the function multiple times is actually like 30% faster. See here: https://gist.github.com/stiegerb/6981b5487cb82483c08c85a5339a4a06

So I would propose to just remove the `make_notification_list` method. And in case no `docId` is provided, let's use the uuid, so users can fall back on that.

The last thing that Alan pointed out yesterday, the `send` method actually only accepts lists now and throws an error if a single notification is passed. I originally wrote it such that it internally converts it to a list with a single entry and sends it. Shall we revert to that or is there a good reason not to? Another reason to allow non-list data would be to allow generator objects.

vkuznet:
Now, when I look at full code I see few flaws in a logic:
- the module has 3 public APIs: send data, make notification and make notification list
- the send data API which handles N documents (where N is large) is very inefficient since our time is O(N) (we loop over every single notification and send one by one), see this line:
https://github.com/dmwm/WMCore/blob/783bd0447fe6e3cd14aa60704bdf1fb9288a75ed/src/python/WMCore/Services/StompAMQ/StompAMQ.py#L120
We should find out if there is a way to send a bulk of docs rather individual ones and adjust code accordingly if such API exists. Moreover the O(N) time is a network time rather the processing time.
- make_notification_list takes docId and then re-use it, see https://github.com/dmwm/WMCore/blob/783bd0447fe6e3cd14aa60704bdf1fb9288a75ed/src/python/WMCore/Services/StompAMQ/StompAMQ.py#L185 I doubt that we need documents with the same ids. Instead, my understanding that such API should not take docId and internally use makeUUID() to generate id of the documents.
- it also does not allow client to choose dataSubfield for documents and it means that it become useless as is if client wants to generate specific dataSubfield docs.
- but as an API it is useful for clients since they can only pass their documents and get notification docs.

stiegerb:
@vkuznet I remember checking whether there was a bulk send method in stomp. As far as I can tell there still isn't. However, I'm not sure it would make a difference, since we send data without waiting for acknowledgments. Note that we already open only a single connection for each list of documents we send. For reference, for the job monitoring data, we are sending bunches of 5000 docs in < 1 second. I don't think we're limited by the processing time inside StompAMQ here.

Shall I go ahead and update the PR to remove the `make_notification_list` method and to take the `_id` as the uuid in case it is not provided?

amaltaro:
> Shall I go ahead and update the PR to remove the make_notification_list method and to take the _id as the uuid in case it is not provided?

Yes, please go ahead.

stiegerb:
Ok, done and rebased to master. For me it's ready to be merged now, but maybe someone should have another look.

vkuznet:
@stiegerb , Benjamin thanks for providing benchmark, it answers my question. The code now looks good to me.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
```diff
@@ -550,8 +550,8 @@ def uploadToAMQ(self, docs, producer, agentUrl, timeS):
                                 host_and_ports=self.hostPortAMQ,
                                 logger=logging)
 
-            notifications = stompSvc.make_notification(payload=docs, docType=docType,
-                                                       docId=producer, ts=timeS)
+            notifications = stompSvc.make_notification_list(payload=docs, docType=docType,
```
amaltaro: Benjamin, can you also update these 2 calls such that they reflect your changes to StompAMQ module?
Something like
```
notifications = []
for doc in docs:
    notifications.append(make_notification(payload=doc, docType=docType, ts=timeS, dataSubfield="payload"))
```

Same for the module below. Thanks

amaltaro:
Thanks, Benjamin. It should work, I'm just not sure I like it being a list comprehension with that for at the end of the 2nd line, it looks like the snippet I shared before is more readable.
Anyways, it's just my opinion and if needed to, I can update this code myself. Thanks

amaltaro (approved):

amaltaro:
Thanks, Benjamin. I ran a quick test and it works just fine.

src/python/WMComponent/AgentStatusWatcher/AgentStatusPoller.py
amaltaro: Sigh.. you missed the "dataSubfield" bit and I didn't catch it on my review either. Fixing it in a new PR

PR #8982, opened 2019-01-29
Make XML parser to always return a list of elements

### LINKED ISSUE #8981: Problem parsing TagCollector xml - fails workflow creation
For some mysterious reason, the `fetch` reqmgr2 API and workflow injection started failing this morning while validating the CMSSW/ScramArch input.

While debugging TagCollector module, I noticed this line (row['project'])
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/TagCollector/TagCollector.py#L75

isn't always a python list. For the case below with a single entry:
```
  <architecture name="slc7_ppc64le_gcc530">
    <project label="CMSSW_8_1_0_pre15" type="Development" state="Announced"/>
  </architecture>
```

row['project'] was actually a flat dict, thus complaining about the string index lookup.

For now I asked Shahzad to temporarily remove that entry (scram/cmssw) - which is there for quite some time now - and we'll need to get back to it.
The best might be to simply convert that xml to another python object type and be done with it.

amaltaro:
Fixes #8981
Current code was also adding duplicate entries to the list (parsing at both start and end events).
Last but not least, added a basic unit test, as it was supposed to have since day 1.

### DIFF
--- a/src/python/WMCore/Services/TagCollector/XMLUtils.py
+++ b/src/python/WMCore/Services/TagCollector/XMLUtils.py
@@ -16,14 +16,15 @@ import xml.etree.cElementTree as ET
 int_number_pattern = re.compile(r'(^[0-9-]$|^[0-9-][0-9]*$)')
 float_number_pattern = re.compile(r'(^[-]?\d+\.\d*$|^\d*\.{1,1}\d+$)')
 
+
 def adjust_value(value):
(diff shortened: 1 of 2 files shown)


amaltaro:
tests are Ok in my VM. Merging it to get a tag out asap

PR #8985, opened 2019-01-29
Delete request workload document if it fails to add spec attachment d…

### LINKED ISSUE #8978: Workflow without reqmgr_workload spec file in couchdb
@vlimant reported failures in one of the unified modules and the apparent problem is that the workflow was created, with the proper reqmgr_workload_cache request doc in the database (couchdb)
https://cmsweb.cern.ch/reqmgr2/data/request?name=pdmvserv_task_SUS-RunIISummer16NanoAODv4-00040__v1_T_190124_181348_3206

but somehow the spec file for the same workflow is missing:
https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/pdmvserv_task_SUS-RunIISummer16NanoAODv4-00040__v1_T_190124_181348_3206/spec

The request status is New, and since it's related to couch, we can't have an atomic operation. However, we should likely invert the order in which these documents get persisted into the database. Having a spec file in there doesn't hurt, but the inverse does.

One more thing, the fetch API fails to find this same workload_cache because it tries to load the spec file for the splitting view
https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_task_SUS-RunIISummer16NanoAODv4-00040__v1_T_190124_181348_3206

To be investigated whether the same happened with workflows in other status...

amaltaro:
…uring creation
Fixes #8978 

I'm not 100% sure it solves the problem reported. We could also retry the upload, but chances that it will fail again are high, so better abort the whole request and make the user start a fresh new injection.

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -557,9 +557,16 @@ class Request(RESTEntity):
             self._update_additional_request_args(workload, request_args)
 
             cherrypy.log("Create request, input args: %s ..." % request_args)
-            workload.saveCouch(request_args["CouchURL"], request_args["CouchWorkloadDBName"],
-                               metadata=request_args)
-            out.append({'request': workload.name()})
+            try:
+                workload.saveCouch(request_args["CouchURL"], request_args["CouchWorkloadDBName"],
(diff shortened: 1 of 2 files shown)


amaltaro:
Very basic tests look Ok too.

PR #8987, opened 2019-01-31
Keep using payload sub-key for MonIT

amaltaro:
I oversaw this detail while reviewing the previous PR. Without the subfield, it breaks the compatibility with the previous docs in MonIT which follow 'data.payload.....'


amaltaro:
Jenkins seem to be stuck. It has been tested, merging it in order to get Lina a new release asap.

PR #8991, opened 2019-02-04
Send job retry number as condor job argument

### LINKED ISSUE #8716: Send a retry count parameter for the job wrapper script
Just as we send the job index, we should also send the retry count such that we can create the job report pkl file as a first thing in the job wrapper.

amaltaro:
Fixes #8716


amaltaro:
tested in my VM

PR #8998, opened 2019-02-05
Record site status changes timestamp and allow grace period for drain…

### LINKED ISSUE #8858: submitfailed only for sites in draining for > 24h
As just discussed with Sharad @sharad1126, in order to decrease the burden on operations in terms of recovering work from sites that go into drain mode (which triggers submitfailed for some specific cases), we should make it a bit smarter and give jobs a grace period before marking them as failed.

Rough logic is:
* when update the site cache in JobSubmitterPoller, set also a timestamp for that status
* if the site is in drain for less than 24h, then just skip that job from the submitter cycle and move to the next one
* else, if the site is in drain for more than 24h and the job can only run there (no IObound jobs), then mark it as submitfailed.

amaltaro:
…ing sites
Fixes #8858 

Changes are:
* there are database schema changes (additional `state_time` column for wmbs_location table)
* record the timestamp for adding a new site to the local database or changing its state
* default grace period for sites in drain set to 2 days. This grace period allows jobs to sit in the Submitter queue either waiting for the site to come back to Normal state OR for the grace period to expire, where JobSubmitter would mark those jobs as submitfailed.

Note that the behavior for sites changing states hasn't been touched. Thus a site being put into drain will keep triggering the job removal IF a job is pending in condor and assigned to only that site


amaltaro:
This seems to be working fine. I'll just make some aesthetic changes and it should be ready to go.

PR #8999, opened 2019-02-05
Added PreMixingModule to MC mix modules

### LINKED ISSUE #8997: Possible misconfiguration with the secondary input
Also related to https://github.com/dmwm/WMCore/issues/8983
and initially reported and discussed over this JIRA ticket: https://its.cern.ch/jira/browse/CMSCOMPPR-4523

I'm still debugging it, but I have strong suspicious on a badly configured pileup task. More news this afternoon.

amaltaro:
Fixes #8997 

@vlimant is this new module type only allowed for MCPileup? Or should it also be supported for DataPileup (L504)?


vlimant:
not too sure about this. the way you have it is good for now IMO. we have to test this of course

amaltaro:
Ok, my test workflow
amaltaro_TaskChain_724_HG1902_Validation_190205_182132_9778 |  

hasn't finished yet, but I do see that it successfully went through the task2 and it's now at the MINIAOD merge step already. I'm backporting this fix and pushing it to all the production agents in a bit.

PR #9014, opened 2019-02-12
Take transient output modules into account when updating task parenta…

amaltaro:
…ge map
cmsweb branch fix of #9015


amaltaro:
tested in my VM

PR #9024, opened 2019-02-15
Add exception message to condor chirp, when present.

### LINKED ISSUE #6939: Chirp exception message
We've had sites who want to explore the ClassAd archive based on a full text search of exception messages.

As part of the WMCore wrapper that currently parses & chirps back the FJR, we should also include the exception message for jobs that fail.

However, with the exception messages, we are likely to hit the 1024 character limit on values passed via Chirp.  Hence, we should extract the exception message, compress it and base64-encode it.  If the encoded version is still greater than 1024 character, then we should recursively repeat the procedure until the encoded text is below the 1024 character threshold.  ElasticSearch can decode things appropriately on the other side.

khurtado:
Fixes #6939

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executor.py
+++ b/src/python/WMCore/WMSpec/Steps/Executor.py
@@ -13,6 +13,7 @@ import json
 import subprocess
 
 from Utils.FileTools import getFullPath
+from Utils.Utilities import zipEncodeStr
 from WMCore.FwkJobReport.Report import Report
 from WMCore.WMSpec.WMStep import WMStepHelper
 from WMCore.WMSpec.Steps.StepFactory import getStepEmulator
@@ -52,7 +53,7 @@ def getStepSpace(stepName):
     return stepSpace
 
 
-class Executor:
+class Executor(object):
     """
     _Executor_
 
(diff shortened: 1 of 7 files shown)


amaltaro:
Kenyi, I was looking into this code today. Maybe you can create a new method like this one:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/FwkJobReport/Report.py#L327

and return `reportError.details` instead. It should contain the error message from the cmsRun FJR, but please verify.

khurtado:
@amaltaro Thanks, that seems like a much better idea. I made one method that returns (exitcode, errorMessage) instead of using stdErr

src/python/WMCore/WMSpec/Steps/Executor.py
```diff
@@ -51,8 +53,39 @@ def getStepSpace(stepName):
         raise RuntimeError(msg)
     return stepSpace
 
+def zipEncodeStr(message, maxLen=5120, compressLevel=9, steps=100, truncateIndicator=" (...)"):
```
amaltaro: Given this function doesn't depend on anything WMCore related, how about we move it under 
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/Utilities.py
?

A unit test testing different scenarios is also very important in this case.

khurtado:
Test this please

khurtado:
test this please

khurtado:
test this please

PR #9026, opened 2019-02-18
Support pileup for StepChain MC top step

### LINKED ISSUE #9025: Add pileup support to StepChain MC from scratch
While working on something else, I noticed the `pileup` section is not created IF the StepChain contains a single step and the first step does not have an input dataset, i.e., if `setupGeneratorTask` is called instead of `setupTask`.

amaltaro:
Fixes #9025

As the title says, handle Pileup datasets for top level steps that do not have any input dataset (and have a single step).


amaltaro:
Tests are Okay.

PR #9030, opened 2019-02-19
Upper limit for MaxWallTimeSecs set to 45h

### LINKED ISSUE #9028: Decrease MaxWallTimeSecs limit in the agents config to 2750min (instead of 2760)
In order to avoid the 25% rounding that would bring this up to 3000mins, thus making it impossible to run in many sites.

amaltaro:
Fixes #9028 
To circumvent a change made to the schedds somewhere in the past, where it rounds MaxWallTimeSecs to the closest 25% (thus 2700 will be 2750).


amaltaro:
Diego already tested it for me.

PR #9037, opened 2019-02-21
Support multiple pileup datasets in StepChain

### LINKED ISSUE #9032: Support multiple pileup in StepChain
Such requests would probably be done with TaskChain instead of StepChain, but this needs to be fixed. Problem is that when we set a pileup configuration for a task, it sets that configuration to all CMSSW steps, thus all cmsRun steps would have the same pileup configuration.

We probably need to pass the cmsRun number to this call
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StepChain.py#L263

and change a few things here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L844

such that it doesn't overwrite previous steps.

amaltaro:
Fixes #9032

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -841,16 +841,23 @@ class StdBase(object):
 
         return
 
-    def setupPileup(self, task, pileupConfig):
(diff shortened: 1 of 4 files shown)


amaltaro:
test this please

amaltaro:
also tested in my VM

PR #9038, opened 2019-02-21
Remove code left for backwards compatibility

amaltaro:
This change should be pretty safe, since those changes were supposed to be removed by May 2018, at most.
For the splitting param change, we might have to manually remove that key from the central couch doc.

We'll need this update to the manage script: https://github.com/dmwm/deployment/pull/726
and then ask Lina to run the following command in one of the ReqMgr2 backends:
`/data/srv/current/config/reqmgr2/manage postdefaultconfig cmsweb-testbed.cern.ch 'I did read documentation'`


amaltaro:
It has also been tested in my VM. There are instructions on my first comment here that need to be carried on once it gets deployed in testbed and production.

PR #9039, opened 2019-02-21
Make type REST api to return json-like instead of xml results

amaltaro:
Return a nicely formatted output when calling the `type` REST API, e.g.:
https://cmsweb.cern.ch/reqmgr2/data/type


amaltaro:
tested in my VM

PR #9048, opened 2019-02-26
Make CRIC module less verbose

amaltaro:
Now that we know it works... it's time to demote some log level and make our life easier when debugging other issues in the logs.
@belforte FYI


belforte:
What I did in crab was to wrap it in a context handler where log level is error only and if it fails, I call again with more verbose mode. Seems a good general strategy to me. I did not go as far as write a generic method to do all of that, sort of "tryWithOptimalLogging"... but if this makes technical sense, would be great to have it in WMCore

amaltaro:
I'm not sure I completely understood you, but... if you have a right away retry logic that would retry that same call, then I see usage for it.

Anyhow, I tagged you such that you're aware we're making it less verbose from now on.

belforte:
thanks. I guess I wanted to say that for CRAB I had made it already : silent, unless needed :-)

PR #9075, opened 2019-03-05
fix unittest schema init, fixes #9062

### LINKED ISSUE #9062: problems with setSchema in unit tests
Initializing the schema in Tier0 unit tests doesn't work anymore. It seems at some we introduced a dependency between the DBSBuffer and WMCore schema that forces us now to initialize the DBSBuffer schema first (which is why such dependencies are a bad idea in general) and only then add the WMCore schema. If I am trying to extend this to include the T0 schema I get problems

        logging.getLogger().setLevel(logging.DEBUG)
        self.testInit.setSchema(customModules = ["WMComponent.DBS3Buffer",
                                                 "WMCore.WMBS",
                                                 "T0.WMBS"],
                                useDefault=False)

and I run the unit test and it only prints out this

DEBUG:root:Tables for WMCore.WMBS created
DEBUG:root:Problem creating database table

ALTER TABLE dataset_locked
                 ADD CONSTRAINT dat_loc
                 FOREIGN KEY (dataset_id)
                 REFERENCES dbsbuffer_dataset(id)
                 ON DELETE CASCADE

(DatabaseError) ORA-00942: table or view does not exist

which is as expected since initializing the WMCore first doesn't work due to the dependency. Problem is, I called it with the DBSBuffer schema first, so why is it trying to initialize the WMCore schema first?

@andresfelquintero should be following up on this with @amaltaro

hufnagel:
Fixes #9062
The current unitest schema init code does not allow to initialize multiple schemas in a defined order. The order you get is essentially random due to the use of a dict to filter the passed list.

### DIFF
--- a/src/python/WMQuality/TestInit.py
+++ b/src/python/WMQuality/TestInit.py
@@ -185,23 +185,25 @@ class TestInit(object):
         if not useDefault:
             defaultModules = []
 
-        # filter out unique modules
-        modules = {}
+        # filter for unique modules
+        modules = []
         for module in (defaultModules + customModules):
-            modules[module] = 'done'
+            if module not in modules:
+                modules.append(module)
 
         try:
-            self.init.setSchema(modules.keys(), params = params)
+            self.init.setSchema(modules, params = params)
         except Exception as ex:
             print(traceback.format_exc())
             raise ex
 
         # store the list of modules we've added to the DB
-        modules = {}
+        # again filter for unique modules
+        modules = []
         for module in (defaultModules + customModules + self.currModules):
-            modules[module] = 'done'
-
-        self.currModules = modules.keys()
+            if module not in modules:
+                modules.append(module)
+        self.currModules = modules
 
         return
 


src/python/WMQuality/TestInit.py
```diff
@@ -185,23 +185,29 @@ def setSchema(self, customModules = [], useDefault = True, params = None):
         if not useDefault:
             defaultModules = []
 
-        # filter out unique modules
-        modules = {}
+        # filter for unique modules
+        seen = set()
```
amaltaro: We don't actually use this `seen` variable. Same comment for below block.

amaltaro (changes requested):
Thanks, Dirk! Comment made in the code, can you update that please?

src/python/WMQuality/TestInit.py
hufnagel: What do you mean, that what I use to filter.

amaltaro: You use `seen` and `modules`, but only `modules` is really used. Thus `seen` can go away and you just update 
`if module not in seen:`
by
`if module not in modules:`

hufnagel: Ok, guess I did my copy&paste from the internet too quickly ;-). Will fix and update the PR.

amaltaro (approved):

amaltaro:
Thanks, Dirk!

PR #9082, opened 2019-03-07
Add possible sites in job submission error reports (SubmitFailed).

### LINKED ISSUE #9047: SubmitFailed not reporting site in wmstats
for pdmvserv_task_HIG-RunIIFall18wmLHEGS-00388__v1_T_181126_083424_8286 for example, job failed with SubmitFailed but the site is not reported in wmstat (while it should to my understanding, @amaltaro ?)

khurtado:
Fixes #9047.

### DIFF
--- a/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
+++ b/src/python/WMComponent/JobSubmitter/JobSubmitterPoller.py
@@ -482,7 +482,8 @@ class JobSubmitterPoller(BaseWorkerThread):
             job['fwjr'] = Report()
             if exitCode in [71102, 71104]:
                 job['fwjr'].addError("JobSubmit", exitCode, "SubmitFailed",
-                                     WM_JOB_ERROR_CODES[exitCode] + ', '.join(job['possibleSites']))
+                                     WM_JOB_ERROR_CODES[exitCode] + ', '.join(job['possibleSites']),
+                                     ', '.join(job['possibleSites']))
             elif exitCode in [71101]:
                 # there is no possible site
                 if job.get("fileLocations"):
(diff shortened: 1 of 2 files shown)


khurtado:
test this please

amaltaro:
Another suggestion would be to create a `setSiteName` method in the Report class and then call it in JobSubmitterPoller under `job['fwjr'].addError...`. No further changes would be needed.

khurtado:
> Another suggestion would be to create a `setSiteName` method in the Report class and then call it in JobSubmitterPoller under `job['fwjr'].addError...`. No further changes would be needed.

Done. I tested briefly before `cmsweb-testbed.cern.ch` went down and it worked fine.

amaltaro (commented):
Kenyi, I actually thought we could leave `addError` untouched and simply add a new public method `setSiteName` to the Report module. Then in JobSubmitterPoller, we make two calls, one to addError and one to setSiteName.

If you agree :), can you also create a simple unit test testing getSiteName and setSiteName?

khurtado:
@amaltaro Ah, I see! Actually, I was actually trying to avoid setting a public method to allow setting `siteName` as much as possible, because in my mind, this is something that should never be done outside the worker node, except of course, during submission errors where the task couldn't reach the WN. Am I overthinking things (or is this not the case)? I can create the public method and do as you say otherwise.

amaltaro:
That's a valid point, Kenyi. In practice it makes no difference whether we have a setter method or we set it from `addError` though.
I think having a setter method would be slightly more friendly, but these changes are so small that I think it's not worth it changing this PR once again.

amaltaro (approved):

PR #9085, opened 2019-03-08
Use `tier0_wmagent` producer for T0 agent uploading data to MonIT

amaltaro:
@vytjan as we discussed, with this change data will be posted to AMQ with `tier0_wmagent` producer name and thus it should end up in a different data source in ES.
It hasn't been tested yet and I believe it won't patch cleanly, I might have to backport this and the previous PR on the same subject. Please let me know.


vytjan:
Thanks @amaltaro , I am going to test this once monitoring guys reply that it's ok to change the producer name to tier0_wmagent. Going to update this here.

amaltaro:
Vytas, as you discussed with the MonIT team, I removed the underscore from the producer name. So instead of `tier0_wmagent`, it will be `tier0wmagent`.

vytjan:
Just replied them. Once they whitelist tier0wmagent producer name and we see that data is in a proper place, I will update this thread.

PR #9086, opened 2019-03-08
Add configuration parameters

vkuznet:
Brian asked to revisit ReqMgr2 MicroService with Unified Transferor. He asked me to setup an instance in read-only mode and provide a demo. I'm adding config parameters which we can use to turn on verbosity level and change the interval of MS cycle.

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Transferor.py
+++ b/src/python/WMCore/MicroService/Unified/Transferor.py
@@ -194,7 +194,11 @@ class UnifiedTransferorManager(object):
         self.config = config
         self.reqmgrAux = ReqMgrAux(self.config.reqmgr2_url)
         self.requests = {}
-        self.reqManager = RequestManager()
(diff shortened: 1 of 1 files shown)


src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -194,7 +194,11 @@ def __init__(self, config=None):
         self.config = config
         self.reqmgrAux = ReqMgrAux(self.config.reqmgr2_url)
         self.requests = {}
-        self.reqManager = RequestManager()
+        group = 'DataOps' if not hasattr(config, "group") else config.group
```
amaltaro: How about
`group = getattr(config, "group", "DataOps")`
?
Same for the others.

vkuznet:
Done, I was not sure if it is applicable to config instance.

PR #9097, opened 2019-03-14
Change RSS to PSS in Performance Monitor

### LINKED ISSUE #8498: Monitor memory utilization based on PSS instead of RSS
Some interesting explanation here: https://elinux.org/Android_Memory_Usage

In short, RSS considers all pages of shared libraries (so there is always an overcounting), while PSS (Proportional) divide that size among all the processes sharing it.

goughes:
Fixes #8498
References to RSS have been updated to PSS in PerformanceMonitor.  There are lots of other refs around in WMTask, StdBase, Watchdog and the WMTask unit tests.  I'd like to give the PerformanceMonitor changes a test, then work on updating the other references.

### DIFF
--- a/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
+++ b/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
@@ -60,6 +60,7 @@ class PerformanceMonitor(WMRuntimeMonitor):
         self.pid = None
         self.uid = os.getuid()
         self.monitorBase = "ps -p %i -o pid,ppid,rss,pcpu,pmem,cmd -ww | grep %i"
+        self.pssMemoryCommand = "awk '/^Pss/ {pss += $2} END {print pss}' /proc/%i/smaps"
         self.monitorCommand = None
         self.currentStepSpace = None
         self.currentStepName = None
@@ -68,7 +69,7 @@ class PerformanceMonitor(WMRuntimeMonitor):
         self.pcpu = []
         self.pmem = []
 
-        self.maxRSS = None
+        self.maxPSS = None
         self.softTimeout = None
         self.hardTimeout = None
         self.logPath = None
@@ -95,7 +96,7 @@ class PerformanceMonitor(WMRuntimeMonitor):
         # Set the steps we want to watch
         self.watchStepTypes = args.get('WatchStepTypes', ['CMSSW', 'PerfTest'])
 
-        self.maxRSS = args.get('maxRSS', None)
+        self.maxPSS = args.get('maxPSS', args.get('maxRSS'))
         self.softTimeout = args.get('softTimeout', None)
         self.hardTimeout = args.get('hardTimeout', None)
         self.numOfCores = args.get('cores', None)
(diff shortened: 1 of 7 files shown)


amaltaro:
@goughes in order to start having it properly filled in the workflow specs - since we'll have a phase where both RSS and PSS will have to be supported - how about we also update the spec with a `maxPSS` in addition to the `maxRSS`?

Places to add that line are:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L239
and
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/WMTask.py#L1199

for WMTask, I suggest we copy this `setMaxRSS` method into `setMaxPSS` and call it from `setMaxRSS`

goughes:
Watchdog can fall back and use the old maxRSS key here:
https://github.com/dmwm/WMCore/blob/a3c47a3351e17b95168d376f6468b3bbee3914fb/src/python/WMCore/WMRuntime/Watchdog.py#L90

When this happens, the new maxPSS key is only set if the job resources change.
https://github.com/dmwm/WMCore/blob/a3c47a3351e17b95168d376f6468b3bbee3914fb/src/python/WMCore/WMRuntime/Watchdog.py#L102

I added a fallback to the maxRSS key in PerformanceMonitor, ensuring the memory check will run if the old key is still in use.
https://github.com/dmwm/WMCore/blob/a3c47a3351e17b95168d376f6468b3bbee3914fb/src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py#L222

amaltaro (approved):
Patch looks good, Erik! I'm going to run a quick test though.

amaltaro:
BTW, can you please squash those commits? Keep one commit for the real fix (src/* changes) and a second one for the test and pylint/pep fixes.

amaltaro:
Erik, I forgot to add it to my review. Can you also update this error message:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMExceptions.py#L156
Thanks

src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
```diff
@@ -182,33 +183,47 @@ def periodicUpdate(self):
...
+        logging.info("PSS: %s; RSS: %s; PCPU: %s; PMEM: %s", smaps_output[0], ps_output[2], ps_output[3], ps_output[4])
 
         msg = 'Error in CMSSW step %s\n' % self.currentStepName
         msg += 'Number of Cores: %s\n' % self.numOfCores
 
-        if self.maxRSS is not None and rss >= self.maxRSS:
-            msg += "Job has exceeded maxRSS: %s\n" % self.maxRSS
-            msg += "Job has RSS: %s\n" % rss
+        if self.maxPSS is not None and pss >= self.maxPSS:
+            msg += "Job has exceeded maxPSS: %s\n" % self.maxPSS
```
amaltaro: @goughes sorry for suggesting it only now, but can you make another small modification here such that the error message is more clear? I'd suggest to add the unit in the end of these 2 lines, e.g.
`msg += "Job has exceeded maxPSS: %s MB\n" % self.maxPSS`

Don't forget to squash the new commit against the first one ;)

amaltaro:
test this please

amaltaro:
Thanks, Erik. I'll fix the unrelated python3 failed check in another PR.

PR #9099, opened 2019-03-15
Add rucio variables to the unittest setup

amaltaro:
After managing to build a new docker image and running it, I noticed I can't import rucio-clients stuff.
I guess this should fix that problem.


amaltaro:
Failure is unrelated to this PR.
I'm merging it, cutting a new release and building new wmagent{,-dev} RPMs such that we can also build a new docker image with the correct docker configuration file. 

FIXME: I've also put a code in place that I'll have to remove before it goes to testbed/production.

PR #9100, opened 2019-03-17
Perform lowering Cms headers

vkuznet:
When testing WMCore code on k8s I found that Cms headers are passed as is, i.e. non lower-case, e.g.  Cms-Auth-Status. But Auth code expects all cms headers in lower-case format. Moreover we found that WMCore in some place also look at `CMS-Auth-XXX` header. To avoid all different type of capitalization of cms auth headers this patch performs lowering all Cms-XXX headers if they are present.

Alan pointed out that JavaScripts used with CouchDB use Camel-Case headers convention. But so far we didn't find any issues with that and therefore leave them as is in JavaScripts.

### DIFF
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -8,6 +8,20 @@ import re
 import zlib
 import base64
 
+def lowerCmsHeaders(headers):
+    """
+    Lower CMS headers in provided header's dict. The WMCore Authentication
+    code check only cms headers in lower case, e.g. cms-xxx-yyy.
+    """
+    lheaders = {}
+    for hkey, hval in headers.items(): # perform lower-case
+        # lower header keys since we check lower-case in headers
+        if hkey.startswith('Cms-') or hkey.startswith('CMS-'):
+            lheaders[hkey.lower()] = hval
+        else:
+            lheaders[hkey] = hval
+    return lheaders
+
 def makeList(stringList):
     """
     _makeList_
--- a/src/python/WMCore/REST/Auth.py
+++ b/src/python/WMCore/REST/Auth.py
@@ -3,6 +3,7 @@ import hashlib
 import hmac
 import re
 
+from Utils.Utilities import lowerCmsHeaders
 
 def get_user_info():
     "Helper function to return user based information of the request"
@@ -15,7 +16,7 @@ def user_info_from_headers(key, verbose=False):
     returns user info object with the data from the headers."""
     # Set initial user information for this request
     log = cherrypy.log
-    headers = cherrypy.request.headers
+    headers = lowerCmsHeaders(cherrypy.request.headers)
     user = {'dn': None, 'method': None, 'login': None, 'name': None, 'roles': {}}
 
     # Reject if request was not authenticated.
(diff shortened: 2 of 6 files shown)


amaltaro:
You might want to have a look at these unit test failures.

vkuznet:
Alan, now it is in your hands. I merged changes from #9112, added dashes, squashed the changes.

vkuznet:
Alan,
I'm not sure why this Unit test depends on this changes:
WMCore_t.WMSpec_t.StdSpecs_t.ReReco_t.ReRecoTest:testMemCoresSettings changed from success to error
and if you point me to the test log we can see what's wrong.

Regarding pylint complains, it seems to me that those are about code which I didn't touch. Do we need to fix those?

amaltaro:
Thanks, Valentin. The ReReco test was a transient issue, likely coming from the sporadically unavailability of DBS.

So, I'm going through the code to make sure it's consistent throughout WMCore.
I'd suggest to apply the same lower case to these `inheaders` here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WebTools/Root.py#L87
Given that we need it for 2 modules under the WebTools package, it's better to write a simple function making this conversion, such that we can write a simple unit test for it as well.

The same applies to this module:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Main.py#L101

It's only about logging, but I think it's worth to spend these extra cycles in order to make the code more consistent.

Last but not least, we also have these couchapps redirect that will likely need the same `toLowerCase()` operation:
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WMStats/shows/redirect.js#L4
https://github.com/dmwm/WMCore/blob/master/src/couchapps/WorkQueue/shows/redirect.js#L4
https://github.com/dmwm/WMCore/blob/master/src/couchapps/couchskel/shows/redirect.js#L4

As we discussed yesterday, this will get merged once the CMSWEB production upgrade is completed (and you fix those above :)).

vkuznet:
Alan, I made code refactoring as you requested. I decided to put new function in Utils.Utilities such that it can be used everywhere in WMCore without circular dependencies (e.g. if we put it in REST then WebTools will depend on REST and vice versa). I also added unit test for it.

But I didn't touched couchapps JS files. The reason is that I think they are used correctly since headers coming from frontend are using Camel case notations, while in python code we only check lower case. But if you insist I can lower headers in JS too.

amaltaro (changes requested):
@vkuznet Valentin, we're still missing a few places here. One is commented out in the code and the second is here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Main.py#L101

About the javascript, let's wait and test it before updating it, maybe we don;t really need to do anything on that side.

amaltaro:
Valentin, can you please also update the description on the first comment in this PR? Just write a short summary of the changes (it's worth mention that we - for the moment - don't see the need to update the javascript code as well).
It helps a bunch when looking at the history and etc.

vkuznet:
Alan,
thanks for pointing out to other CMS-Auth headers, I didn't inspect WMCore code for their presence. I fixed them and squashed the commits. I also updated my first comment on this PR describing the issue as you requested.

amaltaro (approved):

amaltaro:
Thanks, Valentin. It seems to be working fine and I didn't spot any issues on central services and/or on the agent.

PR #9103, opened 2019-03-19
Set retries to 0 for LogArchive to EOS prod/recent

### LINKED ISSUE #9060: Disable retries for logArchive copy to EOS over HTTP
As discussed with Dirk and Steven, stage out of the logArchive tarballs to CERN EOS `/store/logs/prod/recent` is not going to work for the AWS project due to agreements of outbound traffic.

Given that that log area is also not well taken care of, I suggest we remove any retries for that stage out such that the job don't waste time waiting + retrying 3 times, thus decreasing the overall job efficiency.

amaltaro:
Fixes #9060 

Disable retries for the logArch1 step when staging out the logArchive tarball to CERN EOS recent area (the one available over HTTP).

### DIFF
--- a/src/python/WMCore/Storage/StageOutImpl.py
+++ b/src/python/WMCore/Storage/StageOutImpl.py
@@ -185,7 +185,7 @@ class StageOutImpl(object):
         #  //
         # // Create the output directory if implemented
         # //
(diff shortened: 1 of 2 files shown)


amaltaro:
test this please

amaltaro:
Now it works fine.

PR #9107, opened 2019-03-20
Allow not failing job creation in JobSplitting/ EventAwareLumiBased if specified in WMAgent config

### LINKED ISSUE #9053: EventAwareLumiBased algo should not fail job creation for T0 agents
Taking information from ticket https://its.cern.ch/jira/browse/CMSTZ-651, we agreed with @hufnagel that a change on the time estimation for jobSplitting will be useful to avoid the 48 hours limit that is set upon some PromptReco jobs. Specifically we discussed this code https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/EventAwareLumiBased.py#L170 and it is suggested that a different type of jobSplitting should be done just for PromptReco and it should have some optional parameter for this specific type.

khurtado:
Fixes #9053
- Add a new parameter to the `EventAwareLumiBased` algorithm (allowCreationFailure) to allow or avoid job creation failures when estimated job time is higher the job time limit. Defaults to `True`.
- Set the splitting parameter above in the StdSpec of PromptReco for `EventAwareLumiBased` and default it to `False`.

### DIFF
--- a/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
+++ b/src/python/WMCore/JobSplitting/EventAwareLumiBased.py
@@ -37,6 +37,7 @@ class EventAwareLumiBased(JobFactory):
         self.collectionName = None  # Placeholder for ACDC Collection Name, if needed
         # TODO this might need to be configurable instead of being hardcoded
         self.defaultJobTimeLimit = 48 * 3600  # 48 hours
+        self.lumiChecker = None
 
     def algorithm(self, *args, **kwargs):
         """
@@ -59,6 +60,7 @@ class EventAwareLumiBased(JobFactory):
         lumis = kwargs.get('lumis', None)
         applyLumiCorrection = bool(kwargs.get('applyLumiCorrection', False))
         deterministicPileup = kwargs.get('deterministicPileup', False)
+        allowCreationFailure = kwargs.get('allowCreationFailure', True)
 
         timePerEvent, sizePerEvent, memoryRequirement = \
             self.getPerformanceParameters(kwargs.get('performance', {}))
@@ -169,9 +171,10 @@ class EventAwareLumiBased(JobFactory):
                 # and it's only one lumi then ditch that lumi
                 timePerLumi = f['avgEvtsPerLumi'] * timePerEvent
                 if timePerLumi > jobTimeLimit and f['lumiCount'] == 1:
-                    failNextJob = True
-                    stopJob = True
                     lumisPerJob = 1
+                    stopJob = True
+                    if allowCreationFailure:
+                        failNextJob = True
                 elif splitOnFile:
                     # Then we have to split on every boundary
                     stopJob = True
--- a/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
@@ -17,6 +17,15 @@ class PromptRecoWorkloadFactory(DataProcessing):
     Stamp out PromptReco workflows.
     """
 
+    def __init__(self):
+        super(PromptRecoWorkloadFactory, self).__init__()
+        self.procJobSplitArgs = {}
+        self.skimJobSplitArgs = {}
+        self.inputProcessedDataset = None
+        self.inputPrimaryDataset = None
+        self.eventsPerJob = None
+        self.inputDataTier = None
+
     def buildWorkload(self):
         """
         _buildWorkload_
@@ -131,6 +140,7 @@ class PromptRecoWorkloadFactory(DataProcessing):
             self.procJobSplitArgs["events_per_job"] = self.eventsPerJob
             if self.procJobSplitAlgo == "EventAwareLumiBased":
                 self.procJobSplitArgs["job_time_limit"] = 96 * 3600  # 4 days in seconds
+                self.procJobSplitArgs["allowCreationFailure"] = False
         elif self.procJobSplitAlgo == "LumiBased":
             self.procJobSplitArgs["lumis_per_job"] = self.lumisPerJob
         elif self.procJobSplitAlgo == "FileBased":
@@ -142,6 +152,7 @@ class PromptRecoWorkloadFactory(DataProcessing):
                 self.eventsPerJob = int((8.0 * 3600.0) / self.timePerEvent)
             if self.skimJobSplitAlgo == "EventAwareLumiBased":
                 self.skimJobSplitArgs["job_time_limit"] = 48 * 3600  # 2 days
+                self.skimJobSplitArgs["allowCreationFailure"] = False
             self.skimJobSplitArgs["events_per_job"] = self.eventsPerJob
         elif self.skimJobSplitAlgo == "LumiBased":
             self.skimJobSplitArgs["lumis_per_job"] = self.lumisPerJob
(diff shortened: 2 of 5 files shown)


src/python/WMComponent/JobCreator/JobCreatorPoller.py
```diff
@@ -72,6 +72,7 @@ def retrieveJobSplitParams(wmWorkload, task):
     if not task:
         return {"files_per_job": 5}
     else:
+        task.setSplittingParameters(allowCreationFailure=allowCreationFailure)
```
amaltaro: This is going to set a new attribute for this task. I suggest we do not set anything, but simply propagate this component parameter all the way to the job splitting algorithm.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
```diff
@@ -511,7 +513,7 @@ def pollSubscriptions(self):
 
             logging.debug("Going to call wmbsJobFactory for sub %i with limit %i", subscriptionID, self.limit)
 
-            splitParams = retrieveJobSplitParams(wmWorkload, workflow.task)
+            splitParams = retrieveJobSplitParams(wmWorkload, workflow.task, self.allowCreationFailure)
             logging.debug("Split Params: %s", splitParams)
```
amaltaro: Given what I said above, I think we can remove this third parameter and add a line like
`splitParams.update(allowCreationFailure=self.allowCreationFailure)`
or something like that.

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
```diff
@@ -169,9 +171,10 @@ def algorithm(self, *args, **kwargs):
                 # and it's only one lumi then ditch that lumi
                 timePerLumi = f['avgEvtsPerLumi'] * timePerEvent
                 if timePerLumi > jobTimeLimit and f['lumiCount'] == 1:
-                    failNextJob = True
-                    stopJob = True
```
amaltaro: From what I see `stopJob` means the splitting for the current job is completed and we don't need to iterate over another file. So I think this change is Ok, but can you have another look just to be safer?

amaltaro (changes requested):
I put some comments along the code. Please let me know if something is not clear.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
hufnagel: Hm, having only a component parameter means this can only be changed for all jobs running in one instance. We wouldn't be able to just use this for PromtpReco for instance.

Currently we could have the component parameter set to True and then override in the spec the job splitter parameter, right?

src/python/WMComponent/JobCreator/JobCreatorPoller.py
hufnagel: See above. I would prefer to be able to override this per job splitter.

src/python/WMComponent/JobCreator/JobCreatorPoller.py
khurtado: Ahh, I made @amaltaro requested change before reading @hufnagel comment. Should I change it back to the original approach (setting a new attribute for the task)?

amaltaro:
Ok, that changes a bunch! @hufnagel you only want this feature for PromptReco, right?
If so, then we need to change the PromptReco spec file and support a new splitting parameter for EventAwareLumiBase (along the lines Kenyi was working on).

src/python/WMCore/JobSplitting/EventAwareLumiBased.py
khurtado: Yes, this seems okay according to the unit tests (seems to do what I want, it actually doesn't if I  set `stopJob = True` only when `allowCreationFailure = True` )

hufnagel:
Yes, we only want this for PromptReco. For Repack and Express we have similar code directly in the job splitters (since they are Tier0 specific). 

So having this as a job splitter parameter for EventAwareLumiBase  and making the adjustment to the PromptReco spec to set it there is the best (least invasive with least possibility for unwanted side effects) way to address this.

amaltaro:
> Yes, we only want this for PromptReco. For Repack and Express we have similar code directly in the job splitters (since they are Tier0 specific).

This is precisely what this patch would provide you, since you said Repack/Express use a T0-specific job splitting.

Point is, I'd like not to touch the job splitting parameters unless it's really necessary. If we add it to Prompt Reco, then it means we also have to support it for P&R.

hufnagel:
PromptReco spec is only used by Tier0. Nothing else uses it.

amaltaro:
Yes, I know. I was thinking more about having a solid schema for the job splitting algorithms, something like:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/DefaultConfig/EDITABLE_SPLITTING_PARAM_CONFIG.py#L8

but even if we add it to the PromptReco spec, it doesn't necessarily mean we need to add it to this list of editable parameters as well, thus no need to support it for production activities as well.

@khurtado let's take Dirk's approach them. You only need to change the PromptReco spec:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/PromptReco.py

and the EventAwareLumiBased module.

khurtado:
Test this please

khurtado:
Test this please

khurtado:
test this please

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
```diff
@@ -131,6 +140,7 @@ def __call__(self, workloadName, arguments):
             self.procJobSplitArgs["events_per_job"] = self.eventsPerJob
             if self.procJobSplitAlgo == "EventAwareLumiBased":
                 self.procJobSplitArgs["job_time_limit"] = 96 * 3600  # 4 days in seconds
+                self.procJobSplitArgs["allowCreationFailure"] = True
```
amaltaro: False

src/python/WMCore/WMSpec/StdSpecs/PromptReco.py
```diff
@@ -142,6 +152,7 @@ def __call__(self, workloadName, arguments):
                 self.eventsPerJob = int((8.0 * 3600.0) / self.timePerEvent)
             if self.skimJobSplitAlgo == "EventAwareLumiBased":
                 self.skimJobSplitArgs["job_time_limit"] = 48 * 3600  # 2 days
+                self.skimJobSplitArgs["allowCreationFailure"] = True
```
amaltaro: False

amaltaro (changes requested):
Changes look good Kenyi, you just need to switch the default value to False in the PromptReco spec.
Good job with the unit tests too!

khurtado:
@amaltaro Done. Also, I squashed all changes into a single commit.

amaltaro (approved):
Thanks, Kenyi. Changes look good to me. I do have two comments though:
a) next time please keep the test and pylint/reformat related changes in a second separated commit (we can chat about it on slack if you prefer). No need to make further changes here though.
b) can you also update the first comment in this PR with a short description of what it provides. Even though the code is pretty straight forward, it's always good to have a short summary of what's provided in the PR.

@andresfelquintero @vytjan can you patch a T0 replay node and fire a test with this patch in please?

khurtado:
@amaltaro Thanks! Initial PR Comment changed. I forgot about keeping the pylint fixes on a different commit, will do that next time.

amaltaro:
Thanks, Kenyi and Andres.

PR #9112, opened 2019-03-22
Lower Cms headers in FronEndAuth module (used by DBS) [DO NOT MERGE YET]

vkuznet:
Yes another place where headers should be lowered before checking, see #9100.
Also properly distinguish 403 messages since they were alike.

### DIFF
--- a/src/python/WMCore/WebTools/FrontEndAuth.py
+++ b/src/python/WMCore/WebTools/FrontEndAuth.py
@@ -58,6 +58,15 @@ class FrontEndAuth(cherrypy.Tool):
         dict with information about the authorized user."""
         headers = cherrypy.request.headers
         user = cherrypy.request.user
+        # protect code against non-lower case Cms headers (happens with nginx ingress on k8s)
+        lheaders = {}
+        for hkey, hval in headers.items(): # perform lower-case
+            # lower header keys since we check lower-case in headers
+            if hkey.startswith('Cms') or hkey.startswith('CMS'):
+                lheaders[hkey.lower()] = hval
+            else:
+                lheaders[hkey] = hval
+        headers = lheaders
 
         if 'cms-auth-status' not in headers:
             # Non SSL request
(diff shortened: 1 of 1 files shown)


bbockelm:
It’s a pity we have to go this route - HTTP headers are defined to be case-insensitive.  Probably a side-effect of representing them in a dictionary?

vkuznet:
but the current code is blindly check lower case headers regardless how they are passed. It is better to protect code from possible mess up (as I discovered when running nginx+frontend). Plus the "coordination" is required between how code is done in frontend (perl) and how we validate headers (in python).

bbockelm:
Right - as I said, it's a pity that we use the wrong representation (a case-sensitive dictionary) of headers for the request.

amaltaro:
So is the reason that we see these issues only now because the current CMSWEB frontends were/are always adding lower case CMS headers while on the Kubernetes setup it doesn't?

If so, it might be worth communicating it to all the CMSWEB service responsibles such that they can evaluate whether a fix on their service is needed or not.

vkuznet:
Alan, I found that using nginx+frontend setup changes the headers, probably to more standard way (as Camel case). 

Our frontend uses lower-case notations in https://github.com/dmwm/deployment/blob/master/frontend/cmsauth.pm module, but it uses capital "CMS" in its apache configuration, see https://github.com/dmwm/deployment/blob/master/frontend/frontend.conf#L34

So far I didn't change anything in frontend, it is side effect of using it with nginx. I think nginx capitalize first letter in the headers and the cmauth.pm perl is case-insensitive when it extracts them. That's why on frontend everything is fine, but our backend fails one by one. Moreover, I also saw that CherryPy changes headers to capital letters in its request if something fails.

At this moment I can't say for sure at which stack (apache, apache perl, CherryPy) the headers are changed, but our python code should be protected.

As far as you can see the changes are only lower CMS specific headers which python code expects. I doubt it is used anywhere because it is only relevant for authentication but we'll validate this once we'll migrate all services to k8s. At this point we know that ReqXXX, CRAB, DBS are WMCore based services and therefore their authentication/authorization is done by WMCore which I address in this issue and in #9100 . The couch auth model uses lower case, but I checked that I can access just fine. That leave us only with DQM which I didn't test.

amaltaro:
@vkuznet can you please move these changes to this PR #9100, such that it's easier to keep track of these? Thanks (don't forget squashing your commits too)

vkuznet:
port changes from this PR to #9100 , closing this PR

PR #9116, opened 2019-03-25
Fix midnight log rotation for cherrypy threads

### LINKED ISSUE #8780: Log for the ReqMgr2 parentageFixTask thread
Looking at the logs on the backend, I see it uses the date from when the process was started, e.g.
vocms0132/reqmgr2/parentageFixTask-20180831.log 

and it never rotates and/or create a new log file when it gets into a new day. Even more interesting, looking at this file now, I can see entries only for today (e.g. 2018-09-03 ...), while the rest is gone.
If one look at vocms055, the file name is the same but the content is from "2018-08-31"...

Weird thing is, it uses the same base class as statusChangeTasks or auxCacheUpdateTasks, and those are rotating just fine (see vocms0731)...

amaltaro:
Fixes #8780 

Problem was with the (- 7200 secs), depending on the polling cycle of the cherrypy thread, we were not actually updating the `self.baseFilename`, thus there were no new log files created.

### DIFF
--- a/src/python/WMCore/WMLogging.py
+++ b/src/python/WMCore/WMLogging.py
@@ -4,9 +4,9 @@ _WMLogging_
 
 Logging facilities used in WMCore.
 """
-import time
 import logging
 import codecs
+from datetime import date, timedelta
 from logging.handlers import HTTPHandler, RotatingFileHandler, TimedRotatingFileHandler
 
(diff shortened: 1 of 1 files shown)


amaltaro:
tested in my VM.

PR #9124, opened 2019-03-28
Add loggin of incoming requests to RESTModel end-points

vkuznet:
This PR provides ability to dump cherrypy request info before request is processed.

### DIFF
--- a/src/python/WMCore/WebTools/RESTModel.py
+++ b/src/python/WMCore/WebTools/RESTModel.py
@@ -8,6 +8,7 @@ from functools import wraps
 from WMCore.Lexicon import check
 from WMCore.WebTools.WebAPI import WebAPI
 from cherrypy import request, HTTPError
+import time
 import cherrypy
 import traceback
 
@@ -160,7 +161,7 @@ class RESTModel(WebAPI):
 
     def _addMethod(self, verb, methodKey, function, args=[],
                   validation=[], version=1, expires = None,
-                  secured=False, security_params={}):
+                  secured=False, security_params={}, dump_request_info=False):
         """
         Add a method handler to self.methods self.methods, decorate it such that it
         receives sanitised input and is marked as 'restexposed'.
@@ -170,6 +171,26 @@ class RESTModel(WebAPI):
             self.methods[verb] = {}
         @wraps(function)
         def wrapper(*input_args, **input_kwargs):
+            if dump_request_info:
+                request = cherrypy.request
+                headers = dict(request.headers)
+                keysToRemove = ['cookie', 'cms-authn-hmac', 'cms-authz-admin']
+                for key in headers.keys():
+                    if key.lower().startswith('ssl-') or \
+                        key.lower() in keysToRemove:
+                        del headers[key]
+                tst = time.gmtime(time.time())
+                msg = 'REQUEST {} {} {} {} {} [{}] [{}] [{}]'.format(\
+                        time.strftime('[%d/%b/%Y %H:%M:%S GMT]', tst),
+                        request.remote.ip,
+                        request.remote.port,
+                        request.method,
+                        request.path_info,
+                        headers,
+                        request.query_string,
+                        request.params
+                        )
+                cherrypy.log.access_log.info(msg)
             if secured:
                 # set up security
                 security = cherrypy.tools.secmodv2


yuyiguo:
Hi  Valentin, Alan,
I am trying to test this patch. I have a  question/comment.
DBS uses WMCore 1.1.16. The current release is 1.2.0.  There may be a lot changes between the two versions. I did not want to make major changes from WMCore right now.  So I used Valentin PR patched a copy of 1.1.16.  I hope this is OK when going to production.

 The standard spec file likes 
```
%define wmcver 1.1.16
Source0: git://github.com/dmwm/WMCore.git?obj=master/%{wmcver}&export=WMCore&output=/WMCore4%{n}.tar.gz
```
Now with my private repository:
```
Source0: git://github.com/yuyiguo/WMCore.git?obj=1.1.16-monitoring/%{wmcver}&export=WMCore&output=/WMCore4%{n}.tar.gz
```
The question is  that wmcver/1.1.16 is a tag in WMCore repository. Do I need to tag my repository too?
Or I could just remove "/%{wmcver}" from source0 for my private repository?
Thanks,
Yuyi

amaltaro:
Yuyi, you should be able to see that tag in your repo (assuming it's called `origin`) if you do a `git fetch origin`. However, since you have a patch on top of it, you might as well tag that commit with `1.1.16.cmsweb1` in your origin remote.
If you have issues with that, let me know and I can branch off and create that tag for you under WMCore, including this fix (only to that branch).

yuyiguo:
Thanks Alan. I created the tag as suggested.

yuyiguo:
@vkuznet 
Valentin,
I am testing the patch. It works, but explore too much info. Could you please take a look and remove below info from the logging:
'Ssl-Client-Cert'
'Cms-Authz-Admin'
Thanks,
Yuyi

vkuznet:
Done, but I also realized that these changes may affect all services which
relies on RESTModel class. Should we make it optional? If so I suggest
to add `dump_request_info=False` parameter into `_addMethod` function such that
user may control if (s)he wants to dump incoming request. But this will imply
that you'll need to explicitly add this option to those DBS methods which you
want to monitor. The advantage as you can see that you can only enable
this for specific APIs.

yuyiguo:
@vkuznet 
It is fine with me to add the option. But I think it is a good to have it for all the APIs. While you are adding the option, could you also add a timestamp in the message? 
I am using this feature to debug a problem and found it is really handy to have it. I also realized that the request did not have a timestamp. 
Thanks,
Yuyi

vkuznet:
Yuyi, ok I added dump_request_info option with False as default, so it means
that you should enable it in all APIs you want. I also added timestamp in GMT
timezone. Please try it out and report.

yuyiguo:
@vkuznet 
Thanks a lot. I will rebuild DBS and let everyone know my result.
Yuyi

yuyiguo:
@vkuznet 
Hi Valentin,

Every DBS API got a 500 error as in [1] after applied the optional dump_requst_info. DBS log did not show much useful info except this line [2].
[1] {"exception": 500, "message": "Server Error", "type": "KeyError"}
[2]  INFO:cherrypy.access:[22/Apr/2019:18:33:58] dbs3-test2.cern.ch 188.184.94.167 "GET /dbs/dev/global/DBSReader/datatiers HTTP/1.1" 500 Internal Server Error [data: - in 65 out 1272 us ] [auth: OK "/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=yuyi/CN=639751/CN=Yuyi Guo" "" ] [ref: "" "curl/7.57.0" ]

Please take a look. Tomorrow Lina is going to do testbed deployment.
Thanks,
Yuyi

vkuznet:
Tell me which VM/node I can check (where code is deployed) and give me sudo
access that I can debug the code and run it in this VM.

vkuznet:
Yuyi,
please try again, I changed PR (code) to avoid key deletion in headers if it
is not there. I think error occurs since I deleted ssl-XXX keys and then tried
to delete ssl- header again from the skip list.

You need to reapply the PR or add my recent changes [1] by hand in your VM.

[1] https://github.com/dmwm/WMCore/pull/9124/commits/500952878be1e47940a109db6699bd696d0c4a8f

yuyiguo:
Let me try. My VM is dbs3-test2.

yuyiguo:
@vkuznet 
Manually edit the file /data/srv/current/sw.yuyi/slc7_amd64_gcc630/cms/dbs3/3.7.8.2/lib/python2.7/site-packages/WMCore/WebTools/RESTModel.py .

I got the same error as before.
Thanks,
Yuyi

yuyiguo:
@vkuznet 
import time
did not solve the problem.

vkuznet:
Yuyi,
I fixed PR again, it was missing time module, and I tested that logs contains a
message from incoming request. But it does not fix the 500 Server Error
which I think comes not from my code but rather from DBS internals somewhere.

The RESTModel.py is patched now and it dumps input requests header.

I also updated PR with missing import time module.

Best,
Valentin.

vkuznet:
It did since log contains the input request info now. It is something else in
DBS code.

vkuznet:
I see the problem, give me some time

yuyiguo:
OK

vkuznet:
So, the problem was with deleting headers used to dump into log. You requested to delete a few like `cms-authn-hmac` but these headers are used in security module. The issue was that I get an alias to headers dict instead of making a copy, i.e. instead of
```
request = cherrypy.request
headers = request.headers
```
we need
```
request = cherrypy.request
headers = dict(request.headers)
```
The later `dict(...)` call makes a **copy** instead of being an alias.
I fixed code in your VM and PR and now DBS APIs work. Feel free to check.

yuyiguo:
Looks good now. Thanks, Valentin.
I will run unit tests now.

yuyiguo:
@vkuznet @amaltaro @h4d4 
All unit tests past. All looked good to me for tomorrow's cmsweb testbed deployment.

Lina, Alan,
We will need to create a WMCore branch from 1.1.16 tag, then apply Valentin's this patch to the branch in order to deploy it on testbed. I did the same thing in my own repository. I think for cmsweb testbed we should use official repo.    This PR can be merged to the master branch too. I will move to the wmcore master branch in next DBS release, not this one. 

As soon as you give me the branch name and tag  name, I will create the DBS cmsdist PR.

Thanks everyone,
Yuyi

amaltaro:
test this please

h4d4:
@amaltaro Alan, 

since this PR is for WMCore, I suppose that are you the person that can create the branch, right?

src/python/WMCore/WebTools/RESTModel.py
```diff
@@ -170,6 +171,27 @@ def _addMethod(self, verb, methodKey, function, args=[],
             self.methods[verb] = {}
         @wraps(function)
         def wrapper(*input_args, **input_kwargs):
+            if dump_request_info:
+                request = cherrypy.request
+                headers = dict(request.headers)
+                skip = ['cookie', 'cms-authn-hmac', 'cms-authz-admin']
+                for key in headers.keys():
+                    if key.lower().startswith('ssl-'):
```
amaltaro: How about
`if key.lower().startswith('ssl-') or key.lower() in skip:`
?

`skip` also sounds like we can just skip those keys, while you're actually removing them from the headers. Should we rename it to `keysToRemove` or something like that?

amaltaro (changes requested):
Thanks, Valentin and Yuyi for fixing and testing it. I made a suggestion along the code and I also need you to squash those commits before we can merge it.

According to the setup_dependencies, it seems not to be used by CRAB.

amaltaro:
@yuyiguo once Valentin applies the last changes, I'll fork 1.1.16 and apply this patch on top of that.
It's good that you're planning to move to master in the next month. Please let me know if you need any advice on which tag to use when the time comes.

src/python/WMCore/WebTools/RESTModel.py
vkuznet: done

vkuznet:
all commits are squashed now

src/python/WMCore/WebTools/RESTModel.py
amaltaro: I guess you forgot to push your changes. I still see both if statements.

vkuznet: Alan, please check now.

yuyiguo:
@vkuznet @amaltaro 
Thanks for all the effort to bring stuff together. 
@h4d4 I saw your comment on testbed deployment. We will need to have this patched WMCore in order to deploy DBS on testbed. I will make the PR as soon as Alan makes the branch.

amaltaro (approved):

PR #9130, opened 2019-04-04
Fix EventBased ACDC job splitting 

amaltaro:
Fixes #9126 

Trying to fix the root cause of this problem, work is still very incomplete.
I will first find a workaround for the production agents, then I can resume this one.

PR #9131, opened 2019-04-04
Workaround for ACDC from MCFakeFiles - wmagent version

amaltaro:
Supposed to be a wmagent branch version of #9130


amaltaro:
I tested it with different setup:
* single lumi per job
* 3 lumis per job

and in different backend agents (MariaDB x Oracle), and it looks Ok'ish. There are still things to be properly fixed, but this workaround will solve the ACDC for MCFakeFiles, for the moment.

amaltaro:
[Fix #9131 - Workaround for ACDC from MCFakeFiles](https://trello.com/c/wp7lVdRR/153-fix-9131-workaround-for-acdc-from-mcfakefiles)

amaltaro:
@khurtado I'm adding it to a new release and patching the production agents soon. But still, it would be great if you could pass an eye over this patch.

PR #9133, opened 2019-04-05
Migrate to from submitMany() to htcondor.Submit() method.

### LINKED ISSUE #9064: Change object used to submit jobs via SimpleCondorPlugin
As discussed in a private thread with the SI team - outcome of the investigation here [1] - we need to move away of the `htcondor.Schedd.SubmitMany()` object/method and use instead the `htcondor.Submit()`, which would add one extra bit to the requirement string and make sure that jobs land on pilots that provide the minimum Cpus requested in the job classad (resizable jobs).

[1]
Diego Davila reply:
" " "
Hello all,

I managed to reproduce the problem and it seems related to the way jobs are submitted using the python bindings.
When one submits a job using the "submit" method[1] of the schedd, the requirements of the job will not include the "TARGET.Cpus >= RequestCpus" expression. This is how production submits.

On the other hand, if the submission is done using a transaction[2], the Requirements of the job will have the "TARGET.Cpus >= RequestCpus" expression added.

I would expect that [1] and [2] will produce the same result but they don't (adding Todd and Jaime in cc for comments)

The easiest way to fix this would be to manually add the expression "TARGET.Cpus >= RequestCpus" into the Requirements of the job, here[3] I guess

Cheers,

Diego

[1]
import htcondor
import classad

ad =classad.ClassAd()

ad['Executable']  = "sleep.sh"
ad['Arguments']   = "600"
ad['RequestCpus']  = 2

schedd = htcondor.Schedd()
schedd.submit(ad)

[2]
import htcondor
import classad

sub = htcondor.Submit({"executable": "sleep.sh", "arguments": "600", "RequestCpus":"2"})
schedd = htcondor.Schedd()
with schedd.transaction() as txn:
    sub.queue(txn)
" " "

khurtado:
Fixes #9064

Working, but needs the following PR merged before:
https://github.com/cms-sw/cmsdist/pull/4899

and ideally also the one below before (since that one isolates changes needed for WMAgent to work with 8.8.3 bindings without changing anything else in the submission methods).
https://github.com/dmwm/WMCore/pull/9236

and this PRs for the ITB/prod schedds (so we are allowed to set our own x509 attributes at submission time):
https://gitlab.cern.ch/ai/it-puppet-module-vocmshtcondor/merge_requests/7
https://gitlab.cern.ch/ai/it-puppet-module-vocmshtcondor/merge_requests/8

Related:
CRAB team is testing 8.8.3 (the previous 8.8.2 did not have a method they need, but this one should).  
https://github.com/dmwm/CRABServer/issues/5884

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+
 """
 _SimpleCondorPlugin_
 
@@ -136,8 +137,10 @@ class SimpleCondorPlugin(BasePlugin):
         # x509 proxy handling
         proxy = Proxy({'logger': myThread.logger})
         self.x509userproxy = proxy.getProxyFilename()
-        self.x509userproxysubject = proxy.getSubject()
-        self.x509userproxyfqan = proxy.getAttributeFromProxy(self.x509userproxy)
+
+        # These are added now by the condor client
+        #self.x509userproxysubject = proxy.getSubject()
+        #self.x509userproxyfqan = proxy.getAttributeFromProxy(self.x509userproxy)
 
         return
 
@@ -159,14 +162,13 @@ class SimpleCondorPlugin(BasePlugin):
         # Submit the jobs
         for jobsReady in grouper(jobs, self.jobsPerSubmit):
 
-            clusterAd = self.getClusterAd()
-            procAds = self.getProcAds(jobsReady)
+            (sub, jobParams) = self.createSubmitRequest(jobsReady)
 
-            logging.debug("Start: Submitting %d jobs using Condor Python SubmitMany", len(procAds))
+            logging.debug("Start: Submitting %d jobs using Condor Python Submit", len(jobParams))
             try:
-                # 4th argument has to be None otherwise HTCondor leaks the result ads
-                # through it (as of 8.7.x). More info in WMCore/#8729
-                clusterId = schedd.submitMany(clusterAd, procAds, False, None)
+                with schedd.transaction() as txn:
+                    submitRes = sub.queue_with_itemdata(txn, 1, iter(jobParams))
+                    clusterId = submitRes.cluster()
             except Exception as ex:
                 logging.error("SimpleCondorPlugin job submission failed.")
                 logging.exception(str(ex))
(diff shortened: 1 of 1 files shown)


khurtado:
@amaltaro After some tests and failures, I realized the condor python bindings used in the Agent are not gotten from the schedd itself, but as an external dependency, linked to `8.5.7`, which doesn't support the following

```
 __init__( (object)arg1, (str)arg2) -> None
queue_with_itemdata( (Submit)self, (Transaction)txn [, (int)count=0 [, (object)from=None]]) -> SubmitResult :
```
I can probably replace these with older methods (dictionary in `__init__`, ugly loop on schedd transaction), but since we are migrating all schedds to `v8.8` already, I think it makes more sense to update the condor spec below: 
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/condor.spec
Thoughts?

amaltaro:
Kenyi, updating the condor spec to 8.8.x is a good idea indeed. I remember having issues with it the last time I was trying to upgrade that spec to 8.6.x, maybe you'll have more luck than me upgrading it to 8.8 though.

khurtado:
Okay, I created a new issue to upgrade to 8.81/8.8.2 (not sure which one yet, I need either, but I think there were some people waiting for a feature in the bindings included in 8.8.2, will ask):

https://github.com/cms-sw/cmsdist/issues/4877

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -469,192 +469,178 @@ def updateJobInformation(self, workflow, task, **kwargs):
...
-
+            ad['+DESIRED_Sites'] = doubleQuotes(sites)
+     
             sites = ','.join(sorted(job.get('potentialSites')))
-            ad['ExtDESIRED_Sites'] = sites
-
-            ad['CMS_JobRetryCount'] = job['retry_count']
-            ad['WMAgent_RequestName'] = job['request_name']
-
+            ad['+ExtDESIRED_Sites'] = doubleQuotes(sites)
```
bbockelm: Use the native condor string quoting mechanism

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -469,192 +469,178 @@ def updateJobInformation(self, workflow, task, **kwargs):
...
+            ad['+WMAgent_JobID'] = str(job['jobid'])
+            ad['+WMAgent_SubTaskName'] = doubleQuotes(job['task_name'])
+            ad['+CMS_JobType'] = doubleQuotes(job['task_type'])
 
+     
             # Handling for AWS, cloud and opportunistic resources
-            ad['AllowOpportunistic'] = job.get('allowOpportunistic', False)
-
+            ad['+AllowOpportunistic'] = str(job.get('allowOpportunistic', False))
+
```
bbockelm: Fixup the commit to avoid spurious whitespace changes.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -469,192 +469,178 @@ def updateJobInformation(self, workflow, task, **kwargs):
...
+            +PeriodicRemoveReason = PeriodicRemove ? "Job automatically removed for being in Held status" : ""
+            """)
+        sub['executable'] = self.scriptFile
+
+        # Required for global pool accounting
+        sub['accounting_group'] = self.acctGroup
+        sub['accounting_group_user'] = self.acctGroupUser
+ 
+        # Customized classAds for this plugin
+        sub['+DESIRED_Archs'] = '"INTEL,X86_64"'
```
bbockelm: This seems unnecessary.

khurtado:
@bbockelm Condor python bindings question:
When looking into differences in the python bindings for 8.5.7 and 8.8.1/2, I noticed the following query is now different:

``` # 8.5.7
>>> import htcondor
>>> col = htcondor.Collector()
>>> scheddAd = col.locate(htcondor.DaemonTypes.Schedd)
>>> scheddAd['CurbMatchmaking'].eval()
False
```
```
8.8.2
>>> col = htcondor.Collector()
>>> scheddAd = col.locate(htcondor.DaemonTypes.Schedd)
>>> scheddAd['CurbMatchmaking'].eval()
KeyError: 'CurbMatchmaking'
```
So this breaks the following code: https://github.com/dmwm/WMCore/blob/bb6a77ae9dab0f4672b34a2e51088c2c6e2b821d/src/python/WMCore/Services/PyCondor/PyCondorAPI.py#L73-L98

This is on the same schedd, just changing the bindings version. The reason is 8.8.2 only gets the following classads: 
>>> print(scheddAd)

    [
        CondorPlatform = "$CondorPlatform: X86_64-CentOS_7.6 $";
        MyAddress = "<188.185.78.132:4080?addrs=188.185.78.132-4080&noUDP&sock=3931_eae1_3>";
        CondorVersion = "$CondorVersion: 8.8.2 May 02 2019 $";
        MyType = "Scheduler";
        Name = "vocms0192.cern.ch";
        Machine = "vocms0192.cern.ch"
    ]
```
while 8.5.7 prints all classads, including the expression for CurbMatchmaking.
```
>>> print(scheddAd)
<snip>
 CurbMatchmaking = ( ShadowsRunning > 9.700000000000000E-01 * 17000 ) || ( RecentDaemonCoreDutyCycle > 9.800000000000000E-01 );
<snip>
```

Do you know how to get all Schedd Ads using 8.8 python bindings?

**Comment by bbockelm** [comment]
Hi @khurtado -

That's on purpose.  The `locate` query is only supposed to return the minimal information to connect to the daemon.  If you want additional ClassAds, you want to use `query` (with the appropriate projection) to get the full ad.

File: src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py [hunk 4]
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+# pylint: disable=W0221
```
amaltaro: Kenyi, just a standalone comment. I'd suggest not to disable any pylint checks otherwise we don't know what needs to be fixed in the future.

khurtado: @amaltaro Fair enough. 
The pylint message complains about something actually unchanged in this PR, but part of the SimpleCondorPlugin file:
```
Arguments number differs from %s method
```
Which is basically complaining about the number of arguments here:
https://github.com/dmwm/WMCore/blob/e94f8cd05b40b1293866914a79b444b57b2b48af/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L411

vs here:
https://github.com/dmwm/WMCore/blob/e94f8cd05b40b1293866914a79b444b57b2b48af/src/python/WMCore/BossAir/Plugins/BasePlugin.py#L96

I considered the warning harmless enough, which is why I disabled it to make the "warning and error that must be fixed" go away. I could just create and `__init__` section defining `self. raiseEx` so the number of arguments match for example, if that sounds better.

amaltaro:
@khurtado it looks like you need to rebase this branch.
Can you also update the description of this PR (including the reason why we changed from Ad to +Ad :-)).

amaltaro:
@khurtado Kenyi, can you please rebase this PR such that we can proceed with further tests today? Thanks

khurtado:
@amaltaro Changes have been rebased.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -488,197 +489,160 @@ def updateJobInformation(self, workflow, task, **kwargs):
...
+                ad['+DESIRED_CMSPileups'] = classad.quote(str(cmsPileups))
             else:
-                ad['DESIRED_CMSPileups'] = classad.Value.Undefined
-
+                ad['+DESIRED_CMSPileups'] = undefined
             # HighIO and repack jobs
-            ad['Requestioslots'] = 1 if job['task_type'] in ["Merge", "Cleanup", "LogCollect"] else 0
-            ad['RequestRepackslots'] = 1 if job['task_type'] == 'Repack' else 0
-
+            ad['+Requestioslots'] = str(1 if job['task_type'] in ["Merge", "Cleanup", "LogCollect"] else 0)
```
amaltaro: Kenyi, why do we pass it as a string now?

amaltaro (approved):
And I just noticed I hadn't reviewed this PR... it looks good, but I still made a comment in the code, Kenyi.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
khurtado: Because the ad dictionary is used on a different class/object now. Bare classads don't really have the concept of '+', to distinguish between built-in classads like e.g.:`request_memory` and cuztomized ones like `+Requestioslots`, but the Submit class does. So, with ads with `+`  in this case, integers, booleans and classads expressions are passed as strings, and strings are passed via `classads.quote()`.

amaltaro: Gotcha, thanks! It'd be worth to have this explanation in the description of the PR (first comment) for future reference ;)

PR #9140, opened 2019-04-09
MicroService code re-factoring

vkuznet:
This PR revisits MicroService code and made it working with proposed [MS-Transferor](https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor) architecture. So far it does the following:
- talks to ReqMgr2 and gets `acquired` workflows
- talks to Workqueue and gets GQE
- performs `as is` Unified Transferor logic

### DIFF
--- a/setup_dependencies.py
+++ b/setup_dependencies.py
@@ -76,7 +76,7 @@ dependencies = {
                     'WMCore.Wrappers.JsonWrapper.__init__',
                     'WMCore.Wrappers.JsonWrapper.JSONThunker',
                     ],
-        'systems': ['wmc-rest', 'wmc-runtime'],
+        'systems': ['wmc-rest', 'wmc-runtime', 'wmc-database'],
         'statics': [],
     },
     'workqueue': {
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -22,6 +22,7 @@ import urllib
 
 # WMCore modules
 from Utils.CertTools import getKeyCertFromEnv
+from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
 from WMCore.Services.pycurl_manager import RequestHandler
 from WMCore.Services.pycurl_manager import getdata as multi_getdata
 
@@ -42,6 +43,10 @@ class UnifiedConfiguration(object):
         if parameter in self.configs:
             return self.configs[parameter]
         return default
+
+    def set(self, parameter, value):
+        "Set parameter from unified configuration"
+        self.configs[parameter] = value
 
 # static variables used in Unified modules
 uConfig = UnifiedConfiguration()
@@ -86,16 +91,6 @@ def getWorkflow(requestName):
     headers = {'Accept': 'application/json'}
     params = {}
     url = '%s/data/request/%s' % (reqmgrUrl(), requestName)
-    mgr = RequestHandler()
-    res = mgr.getdata(url, params=params, headers=headers, ckey=ckey(), cert=cert())
-    data = json.loads(res)
-    return data.get('result', [])
-
-def getWorkflows(state):
-    "Get list of workflows from ReqMgr2 data-service"
-    url = '%s/data/request' % reqmgrUrl()
-    headers = {'Accept': 'application/json'}
-    params = {'status': state}
     mgr = RequestHandler()
     res = mgr.getdata(url, params=params, headers=headers, ckey=ckey(), cert=cert())
     data = json.loads(res)
@@ -259,60 +254,71 @@ def cert():
     pair = getKeyCertFromEnv()
     return pair[1]
 
-def stucktransferUrl():
+def stucktransferUrl(url=None):
     "Return stucktransfer url"
-    default = 'https://cms-stucktransfers.web.cern.ch/cms-stucktransfers'
-    return uConfig.get('stucktransferUrl', default)
+    if not url:
+        url = 'https://cms-stucktransfers.web.cern.ch/cms-stucktransfers'
+    return uConfig.get('stucktransferUrl', url)
 
-def dashboardUrl():
+def dashboardUrl(url=None):
     "Return dashboard url"
-    default = 'http://dashb-ssb.cern.ch/dashboard/request.py'
-    return uConfig.get('dashboardUrl', default)
+    if not url:
+        url = 'http://dashb-ssb.cern.ch/dashboard/request.py'
+    return uConfig.get('dashboardUrl', url)
 
-def monitoringUrl():
+def monitoringUrl(url=None):
     "Return monitoring url"
-    default = 'http://cmsmonitoring.web.cern.ch/cmsmonitoring'
-    return uConfig.get('monitoringUrl', default)
+    if not url:
+        url = 'http://cmsmonitoring.web.cern.ch/cmsmonitoring'
+    return uConfig.get('monitoringUrl', url)
 
-def dbsUrl():
+def dbsUrl(url=None):
     "Return DBS URL"
-    default = 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader'
-    return uConfig.get('dbsUrl', default)
+    if not url:
+        url = 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader'
+    return uConfig.get('dbsUrl', url)
 
-def reqmgrUrl():
+def reqmgrUrl(url=None):
     "Return ReqMgr2 url"
-    default = 'https://cmsweb.cern.ch/reqmgr2'
-    return uConfig.get('reqmgrUrl', default)
+    if not url:
+        url = 'https://cmsweb.cern.ch/reqmgr2'
+    return uConfig.get('reqmgrUrl', url)
 
-def reqmgrCacheUrl():
+def reqmgrCacheUrl(url=None):
     "Return ReqMgr cache url"
-    default = 'https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache'
-    return uConfig.get('reqmgrCacheUrl', default)
+    if not url:
+        url = 'https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache'
+    return uConfig.get('reqmgrCacheUrl', url)
 
-def phedexUrl():
+def phedexUrl(url=None):
     "Return PhEDEx url"
-    default = "https://cmsweb.cern.ch/phedex/datasvc/json/prod"
-    return uConfig.get('phedexUrl', default)
+    if not url:
+        url = 'https://cmsweb.cern.ch/phedex/datasvc/json/prod'
+    return uConfig.get('phedexUrl', url)
 
-def ssbUrl():
+def ssbUrl(url=None):
     "Return Dashboard SSB url"
-    default = "http://dashb-ssb.cern.ch/dashboard/request.py"
-    return uConfig.get('ssbUrl', default)
+    if not url:
+        url = 'http://dashb-ssb.cern.ch/dashboard/request.py'
+    return uConfig.get('ssbUrl', url)
 
-def agentInfoUrl():
+def agentInfoUrl(url=None):
     "Return agent info url"
-    default = 'https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStats/_view/agentInfo?stale=update_after'
-    return uConfig.get('agentInfoUrl', default)
+    if not url:
+        url = 'https://cmsweb.cern.ch/couchdb/wmstats/_design/WMStats/_view/agentInfo?stale=update_after'
+    return uConfig.get('agentInfoUrl', url)
 
-def mcoreUrl():
+def mcoreUrl(url=None):
     "Return mcore url"
-    default = "http://cmsgwms-frontend-global.cern.ch/vofrontend/stage/mcore_siteinfo.json"
-    return uConfig.get('mcoreUrl', default)
+    if not url:
+        url = 'http://cmsgwms-frontend-global.cern.ch/vofrontend/stage/mcore_siteinfo.json'
+    return uConfig.get('mcoreUrl', url)
 
-def workqueueUrl():
+def workqueueUrl(url=None):
     "Return WorkQueue url "
-    default = 'https://cmsweb.cern.ch/couchdb/workqueue'
-    return uConfig.get('workqueueUrl', default)
+    if not url:
+        url = 'https://cmsweb.cern.ch/couchdb/workqueue'
+    return uConfig.get('workqueueUrl', url)
 
 def workqueueView(view, kwds=None):
     "Return WorkQueue view url"
--- a/src/python/WMCore/MicroService/Unified/RequestInfo.py
+++ b/src/python/WMCore/MicroService/Unified/RequestInfo.py
@@ -9,18 +9,18 @@ from __future__ import print_function, division
 
 # system modules
 import json
-import pickle
 import time
+import pickle
+import logging
 
 from WMCore.MicroService.Unified.Common import uConfig, cert, ckey, \
     reqmgrCacheUrl, dbsUrl, eventsLumisInfo, workflowsInfo, \
     dbsInfo, phedexInfo, reqmgrUrl, elapsedTime, getComputingTime, \
-    getNCopies, teraBytes, workqueueView
+    getNCopies, teraBytes, getWorkflow
 from WMCore.MicroService.Unified.SiteInfo import SiteInfo
 # WMCore modules
 from WMCore.Services.pycurl_manager import RequestHandler
 from WMCore.Services.pycurl_manager import getdata as multi_getdata
-
 
 def findParent(dataset):
     "Helper function to find a parent of the dataset"
@@ -150,6 +150,9 @@ def taskDescending(node, select=None):
 def getRequestWorkflows(requestNames):
     "Helper function to get all specs for given set of request names"
     urls = [str('%s/data/request/%s' % (reqmgrUrl(), r)) for r in requestNames]
+    print("### getRequestWorkflows")
+    for u in urls:
+        print(u)
     data = multi_getdata(urls, ckey(), cert())
     rdict = {}
     for row in data:
@@ -269,7 +272,7 @@ def getMulticore(request):
     return max(mcores)
 
 
-def getSiteWhiteList(request, siteInfo, reqmgrAuxSvc, reqSpecs=None, pickone=False, verbose=True):
+def getSiteWhiteList(svc, request, siteInfo, reqSpecs=None, pickone=False, verbose=True):
     "Return site list for given request"
     lheinput, primary, parent, secondary = getIO(request)
     allowedSites = []
@@ -306,7 +309,11 @@ def getSiteWhiteList(request, siteInfo, reqmgrAuxSvc, reqSpecs=None, pickone=Fal
                       minChildJobPerEvent, rootJobPerEvent, maxBlowUp)
 
     for campaign in getCampaigns(request):
-        campaignConfig = reqmgrAuxSvc.getCampaignConfig(campaign)
+        # for testing purposes add post campaign call
+        # res = svc.reqmgrAux.postCampaignConfig(campaign, {'%s_name' % campaign: {"Key1": "Value1"}})
+        campaignConfig = svc.reqmgrAux.getCampaignConfig(campaign)
+        if isinstance(campaignConfig, list):
+           campaignConfig = campaignConfig[0]
         campSites = campaignConfig.get('SiteWhitelist', [])
         if campSites:
             if verbose:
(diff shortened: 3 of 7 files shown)


src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -259,60 +276,71 @@ def cert():
...
+        url = 'http://dashb-ssb.cern.ch/dashboard/request.py'
+    return uConfig.get('dashboardUrl', url)
 
-def monitoringUrl():
+def monitoringUrl(url=None):
     "Return monitoring url"
-    default = 'http://cmsmonitoring.web.cern.ch/cmsmonitoring'
-    return uConfig.get('monitoringUrl', default)
+    if not url:
+        url = 'http://cmsmonitoring.web.cern.ch/cmsmonitoring'
```
amaltaro: Don't waste time refactoring these simple get methods. Eventually we should convert all of them to service arguments and that's it.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -81,6 +85,19 @@ def phedexInfo(datasets):
             blockNodes[item['name']] = nodes
     return blockNodes
 
+def getRequests(status):
```
amaltaro: This is not using what I suggested here: https://github.com/dmwm/WMCore/issues/9142#issuecomment-482565706

vkuznet: Alan, this is the part of **Unified black box** logic which I didn't touch yet. I want you first to concentrate on parts which I made for first release, and once it is everything in place I'll move on fixing stuff like that.

vkuznet: Ohh, sorry, I thought about something else. I'll fix it.

vkuznet:
Alan, I fixed this API.

amaltaro:
Valentin, I noticed this week that the model we have been discussing was missing one use case. Which is regarding the pileup sample and the need to have it completely available for a workflow (either in one or in scattered among many sites). Which brings me to the conclusion that we cannot have a service only making subscriptions and not monitoring them.

I'm drafting another proposal and will write it down soon, such that we can go through it later today, if you have time.

amaltaro:
@vkuznet I'd suggest to separate the developments you're putting into this PR into a couple of PRs, or maybe even 3. This PR has a) code refactoring, b) further developments for Transferor and c) implementation of TransferMonit, making it harder to validate and thus easier to introduce a bug.

If not too late, try to at least keep those 3 above in different commits.

vkuznet:
Alan,
since both MSTransferor and MSTransferorMonit are integrated it would be hard to
separate this PR because we want to have fully working system at once.
Instead, I suggest to merge this PR and then separate further
development into different streams as you suggested. Moreover it seems
logical since now we can concentrate on individual implementations.
V.

vkuznet:
@amaltaro , Alan please review this PR. I removed WorkQueue pieces from the code as was requested on today's meeting.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+    def monit(self, req_status='staging'):
+        """
+        MSManager monitoring function.
+        It performs transfer requests from staging to staged state of ReqMgr2.
+        For references see
+        https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor
+        """
+        try:
+            # get requests from ReqMgr2 data-service for given statue
+            requestSpecs = self.svc.reqmgr.getRequestByStatus([req_status], detail=True)
```
amaltaro: We don't need request details here (detail=False). We only need to know workflow names in this status, all the rest of the information should be available from the application database containinig a map of workflow -- dataset -- transfer id

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+            requests = [r for item in requestSpecs for r in item.keys()]
+            self.logger.debug('+++ monit found %s requests in %s state' % (len(requests), req_status))
+
+            # nothing to do
+            if not requests:
```
amaltaro: these 3 lines are not really needed, the API should return an empty dict, so the for loop below will be automatically skipped.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+            if rec['tid'] == tid:
+                return rec['request']
+
+    def getTransferStatuses(self, tids):
+        "get transfer ids from backend"
+        # transfer docs on backend has the following form
+        # https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154
+        statuses = []
+        records = self.store.records()
+        store_tids = [r.get('tid', 0) for r in records]
```
amaltaro: You are scanning this `records` structure twice. Remove this line above and anything related to `store_tids`, just find the tid from the records as done in the for loop below.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+        pass
+
+    def getTransferIDs(self):
+        "Get transfer ids from backend"
+        # TODO/Wait: https://github.com/dmwm/WMCore/issues/9198
+        # meanwhile return transfer ids from internal store 
+        tids = [r['tid'] for r in self.store.records()]
+        return tids
+
+    def getRequest(self, tid):
```
amaltaro: It should be the other way around, return transfer id for a given request (sort of as implemented above).

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+        statuses = []
+        records = self.store.records()
+        store_tids = [r.get('tid', 0) for r in records]
+        for tid in tids:
+            if tid in store_tids:
+                for rec in records:
+                    if rec.get('tid', 0) == tid:
+                       statuses.append(rec['status'])
+                       break
+            else:
```
amaltaro: For the record, this else should never get executed. The MS is supposed to persist all transfer ids, so if we cannot find it, we have a bug.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+
+    def getRequest(self, tid):
+        "Return request name for given transfer id"
+        records = self.store.records()
+        for rec in records:
+            if rec['tid'] == tid:
+                return rec['request']
+
+    def getTransferStatuses(self, tids):
+        "get transfer ids from backend"
```
amaltaro: update this doc string

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
-        if self.verbose:
-            print("### sdict", sdict)
+        self.logger.debug("### sdict %s" % sdict)
         tot = len(sdict.keys())
         if not tot:
             return -1
         # return percentage of completion
         return round(float(sum(sdict.values()))/float(tot), 2) * 100
 
+    def checkTransferRequests(self, request):
```
amaltaro: I'm not sure I understand the difference between these 3 methods: getTransferStatuses, checkSubscription and this one. To me they seem to have the same goal, checking the status of the transfer requests from the cache or querying the data-service to get up-to-date status.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+             -d '{"RequestStatus":"staging", "RequestName":"bla-bla"}' \
+             https://xxx.yyy.zz/reqmgr2/data/request
+        """
+        try:
+            rec = self.store.get(request)
+            self.logger.debug('%s changing %s status of record %s to req_status=%s' \
+                % (prefix, request, rec, req_status))
+            if rec['reqStatus'] != req_status:
+                self.svc.reqmgr.updateRequestStatus(request, req_status)
+                self.store.update(request, {'reqStatus': req_status})
```
amaltaro: I'd rather not save the request status in multiple databases. We know sooner or later it goes out of sync and it just gives us more problems. I'd say we don't store request status in this self.store temp db.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -168,37 +384,56 @@ def add(self, requests):
...
                 subscription = PhEDExSubscription(datasets, sites, self.group)
-                if self.verbose:
-                    print("### add subscription", subscription)
+                self.logger.debug("### add subscription %s" % subscription)
                 # TODO: when ready enable submit subscription step
                 # self.phedex.subscribe(subscription)
 
+		# TODO: when we made subscription we change status of request in ReqMgr2
+
     def info(self, request=None):
```
amaltaro: Shouldn't `request` parameter be mandatory here? If it's not provided, what exactly is it returning?

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -168,37 +384,56 @@ def add(self, requests):
...
         return idict
 
     def delete(self, request):
         "Delete request in backend"
         return self.store.delete(request)
 
+class Services(object):
+    "Services class provides access to reqmgr2 services: ReqMgr, ReqMgrAux"
+    __slots__ = ('reqmgrAux', 'reqmgr', 'workqueue')
+    def __init__(self, reqmgrUrl, workqueueUrl, logger=None):
```
amaltaro: Remove workqueueUrl.
BTW, I'd completely remove this class and just get an instance of ReqMgr, ReqMgrAux, wherever you need it (apparently it's used only once for now)

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -168,37 +384,56 @@ def add(self, requests):
...
+        self.reqmgrAux = ReqMgrAux(reqmgrUrl, logger=logger)
+        self.reqmgr = ReqMgr(reqmgrUrl, logger=logger)
+
 class UnifiedTransferorManager(object):
     """
     Initialize UnifiedTransferorManager class
```
amaltaro: Can you improve this doc string with what this class is meant to do? Does it deserve its own python file?

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -168,37 +384,56 @@ def add(self, requests):
...
     """
-    def __init__(self, config=None):
+    def __init__(self, config=None, logger=None):
         self.config = config
-        self.reqmgrAux = ReqMgrAux(self.config.reqmgr2_url)
+        if logger:
+            self.logger = logger
+        else:
+            self.logger = logging.getLogger('reqmgr2ms:UnifiedTransferorManager')
```
amaltaro: How about replacing it by
`logging.getLogger(self.__class__.__name__)`
?

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -168,37 +384,56 @@ def add(self, requests):
...
+            logging.basicConfig()
+        reqmgrUrl = getattr(config, 'reqmgrUrl', 'https://cmsweb.cern.ch/reqmgr2')
+        workqueueUrl = getattr(config, 'workqueueUrl', 'https://cmsweb.cern.ch/couchdb/workqueue')
```
amaltaro: no need for this

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -168,37 +384,56 @@ def add(self, requests):
...
-        self.reqManager = RequestManager(group, dbFileName, interval, verbose)
+        self.svc = Services(reqmgrUrl, workqueueUrl, self.logger)
+        group = getattr(config, 'group', 'DataOps')
+        dbFileName = getattr(config, 'dbFileName', None)
+        interval = getattr(config, 'interval', 3600)
+        self.verbose = getattr(config, 'verbose', False)
+        # update uConfig urls according reqmgr2ms configuration
+        uConfig.set('reqmgrUrl', reqmgrUrl)
+        uConfig.set('reqmgrCacheUrl', getattr(config, 'reqmgrCacheUrl', 'https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache'))
+        uConfig.set('workqueueUrl', getattr(config, 'workqueueUrl', workqueueUrl))
```
amaltaro: no need again

amaltaro (changes requested):
I made many comments along the code.
In addition to that, I have a few comments and concerns to share:
* given that most of the web-services interaction are already implemented in WMCore, it makes sense to use those Services API when possible; however, we need to look out for performance degradation, since those don't support a pool of requests to be made all in parallel as the pycurl does.
* I'm certainly missing some details, but I'm afraid we're not caching many things that could, thus making many unnecessary duplicate calls (e.g., fetching requests from ReqMgr)
* you have a mix of variable naming convention all over this code. Let's stick to camelCase please.

I have to say this code is also still very confusing to me. Reason is likely because we copied over lots of things from unified. So whatever we can make simpler, we should!

setup_dependencies.py
```diff
@@ -75,8 +75,11 @@
         'modules': ['WMCore.Wrappers.__init__',
                     'WMCore.Wrappers.JsonWrapper.__init__',
                     'WMCore.Wrappers.JsonWrapper.JSONThunker',
+                    'WMCore.WorkQueue.__init__',
```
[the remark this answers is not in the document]
vkuznet: yes this can be removed now.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -259,60 +254,71 @@ def cert():
     pair = getKeyCertFromEnv()
     return pair[1]
 
-def stucktransferUrl():
+def stucktransferUrl(url=None):
```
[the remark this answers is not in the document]
vkuznet: since I already made all of these changes, I'll leave them here.

src/python/WMCore/MicroService/Unified/RequestInfo.py
```diff
@@ -9,19 +9,19 @@
...
-import pickle
 import time
+import pickle
+import logging
 
 from WMCore.MicroService.Unified.Common import uConfig, cert, ckey, \
     reqmgrCacheUrl, dbsUrl, eventsLumisInfo, workflowsInfo, \
     dbsInfo, phedexInfo, reqmgrUrl, elapsedTime, getComputingTime, \
-    getNCopies, teraBytes, workqueueView
+    getNCopies, teraBytes, getWorkflow
```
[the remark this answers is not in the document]
vkuznet: I agree.

src/python/WMCore/MicroService/Unified/RequestInfo.py
```diff
@@ -335,41 +342,89 @@ def getSiteWhiteList(request, siteInfo, reqmgrAuxSvc, reqSpecs=None, pickone=Fal
...
+    # TMP: so far we make it simple, i.e. we return dicts for request names
+    if len(requestSpecs):
+        return requestSpecs
+    requestsToProcess = {}
+    for name, campaign in campaigns.items():
+        if name in requestSpecs:
+            requestsToProcess[name] = requestSpecs[name]
+    return requestsToProcess
+
+def requestsInfo_old(svc, state='acquired', requests=None, logger=None, verbose=False):
```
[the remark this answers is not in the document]
vkuznet: I would like to keep this code since it contains original port of Unified logic. Once I'll port it to requestsInfo we'll remove this function.

src/python/WMCore/MicroService/Unified/SiteInfo.py
```diff
@@ -81,13 +81,14 @@ def fetch(self):
...
-            '%s/stuck_1.json' % stucktransferUrl(),
-            '%s/stuck_2.json' % stucktransferUrl(),
-            '%s/stuck_m1.json' % stucktransferUrl(),
-            '%s/stuck_m3.json' % stucktransferUrl(),
-            '%s/stuck_m4.json' % stucktransferUrl(),
-            '%s/stuck_m5.json' % stucktransferUrl(),
-            '%s/stuck_m6.json' % stucktransferUrl(),
+# VK: so far this is unavailable
+#            '%s/stuck_1.json' % stucktransferUrl(),
+#            '%s/stuck_2.json' % stucktransferUrl(),
```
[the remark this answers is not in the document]
vkuznet: ok, will remove

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -72,88 +86,290 @@ def close(self):
...
+    def __init__(self, svc, group='DataOps', dbFileName=None, interval=60, logger=None, verbose=False):
+        if logger:
+            self.logger = logger
+        else:
+            self.logger = logging.getLogger('reqmgr2ms:MSManager')
+            logging.basicConfig()
         self.verbose = verbose
+        if verbose:
+            self.logger.setLevel(logging.DEBUG)
         if not dbFileName:
```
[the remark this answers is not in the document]
vkuznet: no, you're wrong, the DB file name is configuration parameter we can and will make it persistent.

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: I will not do it until we'll settle on couchdb record formats and you'll implement proper APIs. So far it is placeholder code which allow to implement required functionality. Its logic will change once we'll get proper tids and APIs.

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok, I remove this method (i used it for testing purposes).

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: once again, it is placeholder code to make whole workflow working. Once proper APIs will be in place it will be re-written. Moreover, most likely I'll remove internal store as well.

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: so far it is placeholder. We may need all 3 for different purposes:
- getTransferStatuses will fetch doc from couchdb and return transfer statuses
- checkSubscription will ask upstream server
- checkTransferRequest will use both above if necessary

I don't think we should worry right now about not implemented function, instead you should evaluate place where it is used.

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: I agree, but since there is no yet couchdb one I used internal store to test whole system. As I wrote in previous comments it is most likely will be gone once we'll have couchdb one.

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: my oversight, I agree and will remove it.

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: ok

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: removed

src/python/WMCore/MicroService/Unified/Transferor.py
vkuznet: removed

src/python/WMCore/MicroService/Unified/Transferor.py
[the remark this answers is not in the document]
amaltaro: Then there is no need to make this parameter optional. But I guess you want to keep it for testing purposes, for the moment(?)

src/python/WMCore/MicroService/Unified/Transferor.py
amaltaro: I just want to avoid with code that "we may need" that is actually never needed and just adds up to code maintenance and complexity. In general, I'd say don't implement something unless you're sure it will be needed (this is something I've been fighting in WMCore over these many past years).

amaltaro:
@vkuznet did you manage to run the very basic tests here? I understand it depends on many other things, but is the code at least not crashing as it is?

The concerns I expressed in my previous comment are still valid; and we will need to revisit them in the next PR.

I'm merging a bunch of things by tomorrow (cherrypy changes, new request status, pulling workqueue elements from another status) and possibly this one such that we can see how it goes in testbed.

Can you please squash all non **test** changes into the same commit and leave unittests/pylint changes in a second commit? Thanks

test/python/WMCore_t/MicroService_t/Unified_t/Transferor_t.py
```diff
@@ -16,16 +16,18 @@
         RequestStore, RequestManager
```
amaltaro: And that's why jenkins report 7 tests deleted. Because you're importing `RequestManager` here, which you renamed to `MSManager`. Please fix it.

vkuznet:
Alan,
thanks for update. I fixed unix tests flow even though they need separate revision since logic has changed completely. I squashed all commits to move things forward.

Feel free to review/decide tomorrow if you want to merge it.

vkuznet:
Alan,
please merge this PR that I can move forward. Honestly I spent more time on comments, then on actual work. I already made significant clean-up of this code, see
https://gist.github.com/vkuznet/ba5fd2015963ed12a21029c9df02e04f
where I remove local store, I removed APIs you complained, I put clear logic based on proposal. But I want to apply changes once this PR is merged. This will allow me to come back to code and re-use it if necessary.

amaltaro:
I guess you're just not used to have people reviewing your PRs.
Alright, let's put it in. Next PR will be the production-quality one though :) Thanks!

src/python/WMCore/MicroService/Unified/RequestInfo.py
```diff
@@ -306,7 +309,11 @@ def getSiteWhiteList(request, siteInfo, reqmgrAuxSvc, reqSpecs=None, pickone=Fal
                       minChildJobPerEvent, rootJobPerEvent, maxBlowUp)
 
     for campaign in getCampaigns(request):
-        campaignConfig = reqmgrAuxSvc.getCampaignConfig(campaign)
+        # for testing purposes add post campaign call
+        # res = svc.reqmgrAux.postCampaignConfig(campaign, {'%s_name' % campaign: {"Key1": "Value1"}})
+        campaignConfig = svc.reqmgrAux.getCampaignConfig(campaign)
+        if isinstance(campaignConfig, list):
+           campaignConfig = campaignConfig[0]
```
amaltaro: one empty space less here btw. Pylint report gives us many good insights!

vkuznet:
It is not that, I doubt that we both need to waste time on code which is far
from what we need and bother about pre-mature optimization/quality.

Once we achieve the functional code according to proposal then reviewing PR in
terms of code syntax and other stuff makes sense.

amaltaro:
I disagree. Not having a final and functional code doesn't mean we can push in garbage code full of very simple mistakes that can be fixed. Most of code convention, unused imports/vars,  can be fixed with a couple of clicks in an IDE. So the cost is very small to have a production-quality code in the repo. Last but not least, we know that half of those cases we never get back to the code and improve it.

vkuznet:
Alan,
we may have disagreement on how to approach code/repositories. But I value my
and peoples time. As such, while awaiting a simple merge I spent time to adjust
code to new specs. This version is more close to specs, it addressed many issues
you raised.  But since you're resisted to merge intermediate code, I saved it for
on my own and pushed new one forward. Feel free to review again.
Valentin.

PR #9145, opened 2019-04-10
add back copy stageout plugin

### LINKED ISSUE #9144: need for cp stageout plugin
For HPC that do not provide internet access to the worker node, stageout has to happen to the internal file system, which is mounted on the worker nodes.

The CP plugin was retired, so I see two options here. I could write a new CP plugin. Alternatively, the XRDCP plugin understands local file paths for input and output, but it misses the automatic output directory creation. I could add the latter based on parsing the output PFN (only do it if stageout is to a local file system path).

@amaltaro , what do you prefer? Or do you see another option?

hufnagel:
Fixes https://github.com/dmwm/WMCore/issues/9144
Not just the same old Backends/CPImpl, but a rewrite. Added stagein support since this exists in XRDCPImpl, which this new version is based on. Do we even still support stagein? If not maybe we should clean that up, it makes the code much more complicated than it has to be.

The Plugins/CPImpl is similar to the old version, with a few minor tweaks/changes.

@amaltaro , what do you think?

### DIFF
--- a/src/python/WMCore/Storage/Plugins/CPImpl.py
+++ b/src/python/WMCore/Storage/Plugins/CPImpl.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+"""
+_CPImpl_
+
+Implementation of StageOutImpl interface for plain cp
+
+"""
+import os
+import errno
+import shutil
+
+from WMCore.Storage.StageOutImplV2 import StageOutImplV2
+from WMCore.Storage.StageOutError import StageOutFailure
+
+
+class CPImpl(StageOutImplV2):
+    """
+    _CPImpl_
+
+    Implement interface for plain cp command
+
+    """
+
+    def createOutputDirectory(self, targetPFN):
+        targetdir = os.path.dirname(targetPFN)
+        if not os.path.isdir(targetdir):
+            try:
+                os.makedirs(targetdir)
+            except OSError as exc:
+                if exc.errno != errno.EEXIST:
+                    raise
+        return
+
+    def doTransfer(self, fromPfn, toPfn, stageOut, pnn, command, options, protocol, checksum ):
+        self.createOutputDirectory(toPfn)
+        shutil.copy(fromPfn, toPfn)
+        if os.path.getsize(fromPfn) != os.path.getsize(toPfn):
+            raise StageOutFailure("Invalid file size")
+        return toPfn
+
+    def doDelete(self, pfn, pnn, command, options, protocol):
+        os.remove(pfn)
+        return
--- a/src/python/WMCore/Storage/Backends/CPImpl.py
+++ b/src/python/WMCore/Storage/Backends/CPImpl.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+"""
+_CPImpl_
+
+Implementation of StageOutImpl interface for plain cp
+
+Useful for sites with large shared POSIX file systems (HPC)
+"""
+from __future__ import print_function
+
+import os
+import errno
+
+from WMCore.Storage.Registry import registerStageOutImpl
+from WMCore.Storage.StageOutImpl import StageOutImpl
+
+
+class CPImpl(StageOutImpl):
+    """
+    _CPImpl_
+
+    """
+
+    def __init__(self, stagein=False):
+        StageOutImpl.__init__(self, stagein)
+        self.numRetries = 5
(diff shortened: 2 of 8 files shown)


hufnagel:
Completely untested, just wanted to get some early feedback on code structure etc. Also not sure how I would test this.

amaltaro:
Dirk, I think we can also copy back those 3 test changes. BTW, you have pushed in an empty CPImpl at the root directory as well.

hufnagel:
Added back CPImpl unit test as-is and also adjusted the other unit test. Seems to me we might want to get rid of the special test-copy unit test implementation altogether since it isn't really used now anymore.

hufnagel:
@amaltaro , how can I get cmsdmwmbot to actually give me the output of the unit test?

amaltaro:
Given the list of failures in the report.html, you look them up here:
https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/8732/#showFailuresLink

which gives you a link to, for instance:
https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/8732/testReport/junit/WMCore_t.Storage_t.Backends_t.CPImpl_t/CPImplTest/testCreateOutputDirectory_simple/

I'm not sure how up-to-date this is, but it might be helpful as well:
https://github.com/dmwm/WMCore/wiki/Understanding-Jenkins

hufnagel:
Thank, that helps, I see what's wrong now. Would be good if the report that is linked in the standard reply would contain that link...

hufnagel:
Ok, this looks good now. @amaltaro , can you review?

Also, what about the other uses of "test-copy"? Should we clean this up?

amaltaro:
The link is made available via "default" check, just push the "Details" link.
If "test-copy" is no longer used anywhere, then I think we can remove it (but I have the impression there are still a few tests using it).

amaltaro:
BTW, you can ask for reviewers at the top right button "Reviewers". It's easier than to find what is waiting for your review.

hufnagel:
I don't think test-copy is actually used anywhere now. There are a few references left though:

[hufnagel@lxplus717 WMCore]$ grep -r test-copy test
test/python/WMCore_t/WMRuntime_t/Scripts_t/SetupCMSSWPset_t.py:            <command value="test-copy"/>
test/python/WMCore_t/WMSpec_t/samples/site-local-config.xml:    <command value="test-copy"/>
test/python/WMCore_t/Misc_t/site-local-config.xml:    <command value="test-copy"/>
[hufnagel@lxplus717 WMCore]$ grep -r test-copy src
src/python/WMCore/Storage/Backends/UnittestImpl.py:registerStageOutImpl("test-copy", LocalCopyImpl)
src/python/WMCore/Storage/Registry.py:                'test-copy': 'TestLocalCopyImpl',

hufnagel:
Ok, requested a review. Even if we want to get rid of these last "test-copy" reference (which at first glance don't seem to be functional to me), that should happen in a separate PR.

src/python/WMCore/Storage/Plugins/CPImpl.py
```diff
@@ -0,0 +1,43 @@
...
+            try:
+                os.makedirs(targetdir)
+            except OSError as exc:
+                if exc.errno != errno.EEXIST:
+                    raise
+        return
+
+    def doTransfer(self, fromPfn, toPfn, stageOut, pnn, command, options, protocol, checksum ):
+        self.createOutputDirectory(toPfn)
+        shutil.copy(fromPfn, toPfn)
```
amaltaro: This `shutil.copy()` doesn't copy all the file metadata. But looking at the history, that's how this plugin used to be, so I think we can leave it like that.

amaltaro (approved):
Patch looks good, Dirk. Is there any place we can run a few jobs to get it tested before we merge it?

In addition to that, I have two comments:
* when describing your changes in the PR, you can write "Fixes #9144" such that it also closes the issue when it gets merged
* please keep test/* changes and pylint/pep fixes in a different commit than the one with the real src/* changes. No need to change this one, but keep it in mind for future ones please.

hufnagel:
There is no site that uses this as a stageout plugin, that's why I said testing this right now is hard.

amaltaro:
Ok, we shall eventually see whether it works or not. Thanks

PR #9147, opened 2019-04-11
Fixed lumis list in the ACDC docs and added a new version field

### LINKED ISSUE #9126: Possible duplicate lumi sections
Might be legit or not:
https://its.cern.ch/jira/browse/CMSCOMPPR-5251

but let me create this GH issue to keep the investigation status here.

amaltaro:
Fixes #9126
Superseeds #9130 

Summary of changes are:
* added a new field to every single acdc document in order to know how to handle it depending on its version (acdc_version=2)
* if document belongs to the old (current) version, then pop the very last lumi section of the `lumis` list (only the last job of a MCFakeFile would be correct, and in such case we'd be popping a good lumi! C'est la vie!)
* Mask object now considers LastEvent/LastLumi/LastRun as processeable, thus I had to update the `getMax*` methods.
* Mask filterRunLumisByMask method updated to no longer yield a lumi section that isn't really processeable
* EventBased algo updated to properly set FirstEvent/LastEvent and FirstLumi/LastLumi, according to the always True `inclusiveMask` parameter. Thus `setMaxAndSkipEvents` and `setMaxAndSkipLumis` are usually provided with maxvalue - 1
* handle non-sequential list of lumis (and with different number of lumis per job) for a MCFakeFile in the EventBased algorithm

### DIFF
--- a/src/python/WMCore/ACDC/CouchFileset.py
+++ b/src/python/WMCore/ACDC/CouchFileset.py
@@ -7,16 +7,14 @@ Created by Dave Evans on 2010-03-19.
 Copyright (c) 2010 Fermilab. All rights reserved.
 """
 import time
-from WMCore.ACDC.Fileset import Fileset
-from WMCore.Database.CouchUtils import connectToCouch, requireFilesetName
 
 import WMCore.Database.CMSCouch as CMSCouch
-
+from WMCore.ACDC.Fileset import Fileset
 from WMCore.Algorithms.ParseXMLFile import coroutine
-
-from WMCore.DataStructs.Fileset import Fileset as DataStructsFileset
 from WMCore.DataStructs.File import File
+from WMCore.DataStructs.Fileset import Fileset as DataStructsFileset
 from WMCore.DataStructs.Run import Run
+from WMCore.Database.CouchUtils import connectToCouch, requireFilesetName
 
 
 @coroutine
@@ -42,9 +40,9 @@ def filePipeline(targets):
     while True:
         inputDict = (yield)
         newFile = File(
-                lfn=str(inputDict[u'lfn']),
-                size=int(inputDict[u'size']),
-                events=int(inputDict[u'events'])
+            lfn=str(inputDict[u'lfn']),
+            size=int(inputDict[u'size']),
+            events=int(inputDict[u'events'])
         )
         targets['run'].send((newFile, inputDict[u'runs']))
         targets['fileset'].addFile(newFile)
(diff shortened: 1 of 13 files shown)


amaltaro:
@belforte @lecriste Stefano, Leo, is CRAB using the same EventBased.py module we have in WMCore? My last (7th) commit in this PR has changes on how the job mask is defined. I found it to be inconsistent with all the other job splitting algorithms, and the reason seems to be because we use those setter methods (setMaxAndSkipEvents, setMaxAndSkipLumis, etc).

Examples and a better description of the problem can be read in #9126.
This is still work in progress and further tests need to be performed, but I don't think I can run away from either changing EventBased.py or ... Mask.py modules

lecriste:
> @belforte @lecriste Stefano, Leo, is CRAB using the same EventBased.py module we have in WMCore?
Apparently yes:
https://github.com/dmwm/CRABClient/blob/ff5f0abe00601ef85610e72394d2f802cf93271c/src/python/CRABClient/JobType/PrivateMC.py#L101

although I did not find a direct import of that file.

belforte:
sorry I got distracted. Yes we use it. All the splitting is done with WMCore splitting factory.
https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/Splitter.py

OTOH I do not see at first sigh differences in call for EvenBased vs. other kind of splitting,
code simply calls the factory with the splitting algoritthm indicated in the crab config
https://github.com/dmwm/CRABServer/blob/37b80da5e7e8a896509c8d636e30cd42bc4fe982/src/python/TaskWorker/Actions/Splitter.py#L2
https://github.com/dmwm/CRABServer/blob/37b80da5e7e8a896509c8d636e30cd42bc4fe982/src/python/TaskWorker/Actions/Splitter.py#L51-L53
https://github.com/dmwm/CRABServer/blob/37b80da5e7e8a896509c8d636e30cd42bc4fe982/src/python/TaskWorker/Actions/Splitter.py#L56

Maybe CRAB does not use those Mask modules, whatever they are.

I'd rather not try to learn how WMCore Subscription and Factory work if I can avoid,
can we maybe simply provide a concrete example of the `splitparam` argument passed
in there ? It seems the only thing that CRAB sends to WMCore besides the name
of the splitting algorithm.

belforte:
By the way, CRAB only allows split by event for MC generation, so I do not think lumi mask
can be used there. For event processing we always force EventAwareLumiBased (or byFile).
I'd say: go ahead assuming it will work, but let us know when there's something to test.

amaltaro:
Thanks, Stefano and Leonardo.
What I'm trying to accomplish here is to really have an inclusive mask object.

So if we have a job with `FirstEvent:1, LastEvent:10` it process all 10 events, including 1 and 10. Which is different from the current behaviour, where it would process from 1 to 9, skipping the 10th event.
Thus, a job with Mask `FirstEvent:1, LastEvent:1`, means to process only the event 1. Same for Lumi and Run.

I'll let you know once it's good to have a look at. And I'll also have a look at the code you pointed out just to make sure it doesn;t break your side.

belforte:
I believe that CRAB does not use that Mask. Would be among factory arguments,
or should I look in some other pace ?

amaltaro:
Stefano, I grepped the CRABServer repo for every single method in the Mask module and it doesn't match anything.

I'm now doing some minor refactoring to to this EventBased module and I came across a scenario that I'm not sure what's the expected behaviour. Should we create a job if the input file (or MCFakeFile) contains 0 events? I believe not, but then I wonder whether it might then leave a fileset open behind? @hufnagel @belforte

belforte:
that's not a use case we have in CRAB since we only use split by events when there is
no input file.

amaltaro:
test this please

hufnagel:
Yes, you need to create a job for a 0 event input file. You need to preserve the information that run/lumis with the 0 events actually contain 0 events. Plus there could also be other run/lumi information present, like DQM histograms.

amaltaro:
Thanks! 0-event file job creation restored.

amaltaro:
Tests are finally looking good (still a last test to run) and I believe I have covered all possible use cases now, hopefully.
Anyways, it would be great to get another pair of eyes on this PR just to make sure I'm not doing something very stupid. @hufnagel @ericvaandering 

PS.: I will remove my debugging as soon as I'm done with my tests.

amaltaro:
test this please

amaltaro:
DBS instabilities are giving jenkins a hard time (meaning not all calls to DBS are mocked...)

amaltaro:
test this please

amaltaro:
Well, I think it's time to get this one merged. Any future review is still welcome.

PR #9154, opened 2019-04-17
Enable lazy-download in the PSet for jobs with LheInputFiles=True

### LINKED ISSUE #7997: Can we enable lazy download when LheInputFiles is true
It looks like the lhe eos space is exported via xrootd, and therefore we should be able to run pLHE type of workflows at other place than CERN, but having the lazy download should be enabled so as to limit the read.
How does one change that ?

amaltaro:
Fixes #7997 

Besides enabling `lazy-download` when a job requires to read lhe articles from CERN, I've also added some extra WMTweak logging on what tweaks are made to the PSet.py

### DIFF
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -413,6 +413,7 @@ def makeJobTweak(job):
             # If there is a preset lumi in the mask, use it as the first
             # luminosity setting
             if job['mask'].get('FirstLumi', None) != None:
+                logging.info("Setting 'firstLuminosityBlock' attr to: %s", job['mask']['FirstLumi'])
                 result.addParameter("process.source.firstLuminosityBlock",
                                     job['mask']['FirstLumi'])
             else:
@@ -425,6 +426,8 @@ def makeJobTweak(job):
         for secondaryFile in inputFile["parents"]:
             secondaryFiles.append(secondaryFile["lfn"])
 
+    logging.info("Adding %d files to 'fileNames' attr", len(primaryFiles))
+    logging.info("Adding %d files to 'secondaryFileNames' attr", len(secondaryFiles))
     if len(primaryFiles) > 0:
         result.addParameter("process.source.fileNames", primaryFiles)
         if len(secondaryFiles) > 0:
@@ -434,8 +437,8 @@ def makeJobTweak(job):
         # That should have the added protection of not going over 2^32 - 1
         # If there is nothing in the mask, then we fallback to the counter method
         if job['mask'].get('FirstEvent', None) != None:
-            result.addParameter("process.source.firstEvent",
-                                job['mask']['FirstEvent'])
+            logging.info("Setting 'firstEvent' attr to: %s", job['mask']['FirstEvent'])
+            result.addParameter("process.source.firstEvent", job['mask']['FirstEvent'])
         else:
             # No first event information in the mask, raise and error
             raise WMTweakMaskError(job['mask'],
@@ -445,7 +448,9 @@ def makeJobTweak(job):
 
     # event limits
     maxEvents = mask.getMaxEvents()
-    if maxEvents == None: maxEvents = -1
+    if maxEvents is None:
+        maxEvents = -1
+    logging.info("Setting 'maxEvents.input' attr to: %s", maxEvents)
     result.addParameter("process.maxEvents.input", maxEvents)
 
     # We don't want to set skip events for MonteCarlo jobs which have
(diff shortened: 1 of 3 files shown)


amaltaro:
```
Failed to open the file 'root://cmsxrootd-site.fnal.gov//store/lhe/18788/DarkSUSY_mH_125_mGammaD_20_14TeV_cT_1000_madgraph452_bridge224_events100k.lhe.xz'
   Additional Info:
      [a] XrdCl::File::Open(name='root://cmsxrootd-site.fnal.gov//store/lhe/18788/DarkSUSY_mH_125_mGammaD_20_14TeV_cT_1000_madgraph452_bridge224_events100k.lhe.xz', flags=0x10, permissions=0660) => error '[ERROR] Server responded with an error: [3011] No servers are available to read the file.
' (errno=3011, code=400). No additional data servers were found.
```

does it mean we cannot read LHE files through xrootd? Or is it just that CERN storage isn't reliable today?!?

vlimant:
today is not the best day to check against /store/lhe on eos ...

amaltaro:
@Dr15Jones Chris, do you know whether the LHESource module supports fallback protocol for accessing files in case it failed to access it locally/on-site?

I've been trying to run test workflows in many places (FNAL/DESY/Pisa) and apparently there is no fallback attempt to open files, e.g.:
```
An exception of category 'FileOpenError' occurred while
   [1] Constructing input source of type LHESource
   [2] Calling StorageFactory::open()
   [3] Calling File::sysopen()
Exception Message:
Failed to open the file 'file:/cms/store/lhe/18788/DarkSUSY_mH_125_mGammaD_20_14TeV_cT_1000_madgraph452_bridge224_events100k.lhe.xz'
   Additional Info:
      [a] open() failed with system error 'No such file or directory' (error code 2)
```

Dr15Jones:
@amaltaro I’m fairly certain that only ROOT based files can do fallback. @bbockelm can give definitive answer.

bbockelm:
Correct - fallback is implemented as part of `PoolSource` and is not available to the `LHESource`.

amaltaro:
Thanks for the confirmation, Chris and Brian.
Is there any any chance to get it also supported in the LHESource in future releases? Otherwise, we shall close this PR and keep running such workflows only on-site (at CERN).

bbockelm:
This is still useful, even for workflows at CERN.

That said - why is the location of the workflow relevant here?  The workflow just needs to give an EOS PFN to the LHESource instead of a LFN...

amaltaro:
Won't it use the protocol defined under site-local-config.xml in order to access that file (either providing the LFN or PFN)? Meaning that if we have `dcap` or `cp`, it won't work.

We don't ever fill the PSet with the LFN of the input files. It can be done for this special case, but then I need to check how to properly build the PFN for such files (I think we need to load the TFC and run the same logic used for stage out...).

bbockelm:
> Won't it use the protocol defined under site-local-config.xml in order to access that file (either providing the LFN or PFN)? Meaning that if we have dcap or cp, it won't work.

The protocol used by CMSSW is based on the URL it is handed, not `site-local-config.xml`.  So, if you call a protocol `cp` and then result in a URL of the form `root://...`, XRootD is still used.

Further, lazy-download is still used for `cp` cases if the filesystem is FUSE mounted (or another remote FS).

You do raise an interesting point about LHESource and fallback that is a valid question for UL.  But it's a bit out-of-scope for this PR.

amaltaro:
I see! This requires some extra work then. I'll get back to this fix in a couple of days. Thanks!

amaltaro:
Actually, I just noticed we don't touch the `*.lhe` files present in the PSet.py, there is no manipulation during the job runtime.

So we either start manipulating those files --> read, map LFN to PFN, write back to the PSet;
or we simply ask McM to provide the PFN for such lhe files, e .g.:
```
root://eoscms.cern.ch//eos/cms/store/lhe/18788/DarkSUSY_mH_125_mGammaD_20_14TeV_cT_1000_madgraph452_bridge224_events100k.lhe.xz
```
instead of
```
/store/lhe/18788/DarkSUSY_mH_125_mGammaD_20_14TeV_cT_1000_madgraph452_bridge224_events100k.lhe.xz
```

I'm not sure how dangerous it's though to have a PFN in the PSet, given that a change of EOS endpoint would cause all jobs to fail.
Any thoughts on this @vlimant @bbockelm

bbockelm:
Hi Alan,

My observation is there are so many hardcoded variables in the GEN step that having someone hardcoding the EOS endpoint is a worry - but hardly near the top of the worries.

Let's keep it separate from this ticket, however.  We've made good progress and we should get it out into the open.

Brian

amaltaro:
Thanks, Brian.
I managed to get a new PSet configuration in couch which provides the `lhe` files with their PFN instead of LFN and it seems to be working fine.

Given that a new agent is supposed to come out in the coming week, I won't backport this feature to the current agent.

@khurtado can you please review this PR? I'd like to put it in for tomorrow's wmagent tag.

vlimant:
for what is worth, having missed most discussions, sorry, here until now, the fallback should have been implemented in the LHESource to prevent more complications than simplification in configuring workflows

PR #9158, opened 2019-04-23
Port from CRABServer code Throttled classes; add unit test for them

vkuznet:


amaltaro:
test this please

src/python/Utils/Throttled.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self, throttle, user):
+        self.throttle = throttle
+        self.user = user
+
+    def __enter__(self):
+        ctr = self.throttle._incUser(self.user)
+        #self.throttle.logger.debug("Entering throttled function with counter %d for user %s" % (ctr, self.user))
+        if ctr >= self.throttle.getLimit():
+            self.throttle._decUser(self.user)
+            raise ExecutionError("The current number of active operations for this resource exceeds the limit of %d for user %s" % (self.throttle.getLimit(), self.user))
```
amaltaro: This is not a built-in exception.

src/python/Utils/Throttled.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __exit__(self, type, value, traceback):
```
amaltaro: Do we actually need those 3 args? If we do, it would be great to rename `type` to something else.

amaltaro (changes requested):
Valentin, in addition to my comments in the code, can you also fix some of those comments made in the report? Including the python3 import

src/python/Utils/Throttled.py
vkuznet: The `__exit__` python API is used when someone wants to have class works with `with` statement. Therefore it requires 3 arguments, the exception type, value and traceback. You can read more about `__enter__` and `__exit__` implementation here: https://stackoverflow.com/questions/1984325/explaining-pythons-enter-and-exit
But I agree that usage of reserved keyword `type` should be avoid. Instead the argument keywords should be changed to `self, exc_type, exc_value, exc_traceback`. I'll this change.

vkuznet:
Alan,
I submitted first round of changes, we'll review once new report will be available.

vkuznet:
@amaltaro , please review again as we need this to enable throttling for DBS

amaltaro (approved):
I didn't follow the whole code, but changes look Ok.

yuyiguo:
@amaltaro 

Could you commit this patch tp branch 1.1.16_dbs? and tag it too?  I am think that I should on this branch for now since there are a lot of new stuff going on. I should not moving to the WMCore head for now.

yuyiguo:
@vkuznet 
Could you add the time limit on the count? It would be nice to make the time limit configureable.

vkuznet:
Yuyi,
I revisited the code and add my changes in new PR: https://github.com/dmwm/WMCore/pull/9211
Please see an example over there.

yuyiguo:
Thanks Valentin, will try it.

PR #9160, opened 2019-04-23
Allow tag creation from dbs branches

amaltaro:
Also support x.y.z_dbs branches when cutting and releasing a new tag.


amaltaro:
Already tested, merging.

PR #9165, opened 2019-04-26
Fix CampaignConfig document type

### LINKED ISSUE #9164: Campaign configuration couch document with incorrect doc type
The CampaignConfig class is suppose to create couch documents with this doc type:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Auxiliary.py#L280

but instead it defaults to `WMAGENT_CONFIG`, as reported here:
https://github.com/dmwm/WMCore/issues/9142#issuecomment-482697437

amaltaro:
Fixes #9164


amaltaro:
tested in my VM.

PR #9166, opened 2019-04-26
Log job splitting changes in ReqMgr2

amaltaro:
No issue created, but recording the POST parameters in the reqmgr2 log is going to be helpful for any future investigation.


amaltaro:
tested in my VM.

PR #9169, opened 2019-04-29
Fix issue with missing received bytes in custom cherrypy logger (used by DBS)

vkuznet:


vkuznet:
@yuyiguo , please try it out and let us know if it fixes the missing input bytes in cherrypy logger messages.

yuyiguo:
@vkuznet 
Below os the DBS version of Root.py. It seemed that you patch a different one. 
https://github.com/dmwm/WMCore/blob/1.1.16_dbs/src/python/WMCore/WebTools/Root.py

vkuznet:
Yuyi, I always patch master head and not specific branches. If you need a specific branch you can do it on your own based on changes I provided here.

yuyiguo:
@vkuznet 
They are very different code.

vkuznet:
Yuyi,
how this line
https://github.com/dmwm/WMCore/blob/1.1.16_dbs/src/python/WMCore/WebTools/Root.py#L86
is different from this one
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WebTools/Root.py#L92

I understand that they are on different line numbers in your editor. But the code is identical.

What I fixed is replaced `-` with `%(i)s` and define what `i` would mean and passed to the string.

Again, look at the code, take parts which I defined in master head, and put it into your branch.

yuyiguo:
@vkuznet 
Sorry that was my mistake. The difference was from other places before the patch part.

yuyiguo:
@vkuznet 
Here is the output [data:  503n 2 out 508369 us ] 

Not sure why the work "in" was cut half?

yuyiguo:
@vkuznet 
For the get calls, it only show "INFO:cherrypy.access:REQUEST",  the information("INFO:cherrypy.access") after the request processed was missing.

vkuznet:
Yuyi, the missing `in` in `[data: 503n 2 out 508369 us ]` was due to bug in string format. I updated PR, originally I put `%(i)` but it should be `%(i)s`.

vkuznet:
please try now with updated PR.

yuyiguo:
@vkuznet 
How about the missing info for the GET calls?

vkuznet:
it is unclear from your messages what's wrong with GET messages. Please post example of POST and GET messages.

yuyiguo:
@vkuznet 

It works for POST now. GET is missing the info after processed.

 "if not rbytes:" was two times? Not sure which one is the GTE case?

vkuznet:
Yuyi, just post GET and POST messages.

The two if statements because request/body size can be calculated differently. I put all three methods. The GET method does not have a body so I would assume it should get zero rbytes.

yuyiguo:
@vkuznet 
See below example [1] of a GET call to DBS before today's patch. I remove some of the info for security reason. There are two parts: 
The first one "INFO:cherrypy.access:REQUEST " is the request info. This part was added recently.
The second one "INFO:cherrypy.access" is the log of the request served by DBS.  This is the old DBS log since one day. It also has [data: - in 54 out 862 us ] part.

Now we missed the second one after today's patch.


[1]
" INFO:cherrypy.access:REQUEST [28/Apr/2019 00:33:53 GMT] 188.184.93.2 56510 GET /serverinfo 'User-Agent': 'DBSClient/3.7.8/', 'Cms-Authz-Developer': 'group:reqmgr', 'X-Forwarded-Proto': 'https', 'X-Forwarded-Server': 'cmsweb-testbed.cern.ch', 'Content-Type':
 'application/json', 'Cms-Request-Uri': '/dbs/int/global/DBSReader//serverinfo'}] [] [{}]

INFO:cherrypy.access:[28/Apr/2019:02:33:53] vocms0132.cern.ch 188.184.93.2 "GET /dbs/int/global/DBSReader/serverinfo HTTP/1.1" 200 OK [data: - in 54 out 862 us ] [auth: OK " ] [ref: "" "DBSClient/3.7.8/" ]

vkuznet:
Yuyi,
GET request does not have a body and there is no in info. But I updated PR one
more time to protect code from this case when there is such info.

Please test it again.

We may dump bytes info in REQUEST as well I think if we can extract
Content-Length HTTP parameter from cherrypy.request. But for that I prefer to
have separate PR.

vkuznet:
@amaltaro , Alan please review the changes as it requires to see dump of DBS incoming requests info and related to ongoing investigation. It would be nice to have these changes in place for next cmsweb upgrade (my understanding that Yuyi will patch separate branch, but if she did we should keep these changes in master head too).

amaltaro:
@vkuznet when your' PR is ready for review, please use the top-right option "Reviewers" to ask for reviews.

src/python/WMCore/WebTools/Root.py
```diff
@@ -88,14 +88,30 @@ def access(self):
         response = cherrypy.response
         inheaders = lowerCmsHeaders(request.headers)
         outheaders = response.headers
+        rbytes = 0
+        # identify size of body from HTTP Content-Length header
+        if 'Content-Length' in cherrypy.request.headers:
```
amaltaro: Given that you assign `cherrypy.request` to `request`, you can replace all those places like this line and a few others below.

amaltaro (changes requested):
Just some aesthetic changes that can be made, otherwise it looks Ok.

yuyiguo:
@vkuznet 
Valentin,

I was testing on cmsweb-testbed and found that the missing log on GET after processing actually happened before this PR. I don't know when this started since I just realized this yesterday. I did not see any errors in the log. Could you take a look at previous PR ? I will continue to test this PR, but I double the missing log will come back with this updated PR.

vkuznet:
Yuyi, I'm not sure I understand what you're saying. Please rephrase where the problem and/or give examples.

yuyiguo:
@vkuznet 
Sorry Valentin, that was  my reading error.  I did not get the log because I got an " HTTP Error 502: Proxy Error".  This was first time that I got 502 on a cmsweb-testbed. I ran 300 unit tests in a single thread and one of the tests got 502 error. It is wired and not sure what I should look into.  I will check the FE log, but not sure I will find anything useful. 

The unit tests kind of ran slower than I thought. I am not sure why. Will rerun it.

yuyiguo:
Here is the query that I got 502 error. 
DBS server log:
```
INFO:cherrypy.access:REQUEST [30/Apr/2019 17:24:50 GMT] 188.184.94.167 53222 GET /blockdump [{'Cms-Authn-Name': 'Yuyi Guo','Accept': 'application/json', 'X-Forwarded-For': '188.184.94.167', 'Cms-Authn-Login': 'yuyi', 'Host': 'vocms0731.cern.ch:8252', 'User-Agent': 'DBSClient/3.7.8.2/', 'X-Forwarded-Server': 'testbed.cern.ch', 'Content-Type': 'application/json', 'Cms-Request-Uri': '/dbs/int/global/DBSReader/blockdump?block_name=%2Funittest_web_primary_ds_name_24607%2Facq_era_24607-v4609%2FGEN-SIM-RAW%2324607'}] [block_name=%2Funittest_web_primary_ds_name_24607%2Facq_era_24607-v4609%2FGEN-SIM-RAW%2324607] [{'block_name': u'/unittest_web_primary_ds_name_24607/acq_era_24607-v4609/GEN-SIM-RAW#24607'}]
```
As you can see the "block_name" was repeated twice.

Then what I got from front end access log is
```
[30/Apr/2019:19:24:50 +0200] cmsweb-testbed.cern.ch 188.184.94.167 "GET /dbs/int/global/DBSReader/blockdump?block_name=%2Funittest_web_primary_ds_name_24607%2Facq_era_24607-v4609%2FGEN-SIM-RAW%2324607 HTTP/1.1" 502 [data: 7164 in 15988 out 475 body 300088217 us ] [auth: TLSv1.2 ECDHE-RSA"/DC=ch/DC=cern/OU=Organic Units/OU=" "-" ] [ref: "-" "DBSClient/3.7.8.2/" ]
```
This blockdump took 300.088217 seconds and I guess this cause the http 502 error. 
I think I used to get timeout error instead of 502 error.  Any reason for this change? I feel the timeout is more clear message than 502.

vkuznet:
Yuyi,
both issues are not related to this PR.

The first one is related to this PR:
https://github.com/cms-sw/cmsdist/pull/4922
see
https://github.com/dmwm/WMCore/pull/9124/files#diff-27caaae36aa502e2719c31060d1bfa8fR190
and
https://github.com/dmwm/WMCore/pull/9124/files#diff-27caaae36aa502e2719c31060d1bfa8fR190

The first one is request.query_string and second block name is from
request.params

It is up to you to decide if you want to keep both. If one of them should be
dropped we need another PR.

The second issue you reported here, time out, has nothing to do with proposed
changes.

Therefore I don't know if I need to address anything here.

Best,
Valentin.

yuyiguo:
@vkuznet 
Valentin, 
In this example, request.query_string and request.params are the same.  I am not sure when they will be different? 
For the message of timeout error or not, I knew it is not related to this PR.  I mentioned here in case you or @amaltaro would knew the change of error message.

amaltaro:
Yuyi, cmsweb frontend time out has always returned a Proxy Error. We've already discussed that it's not the best error message, and I believe Diego G. explained me the reason behind that (or why they cannot change it). I no longer remember it since it was many years ago. I think we can simply disregard it and take a `502 Proxy Error` as a frontend timeout.

vkuznet:
Yuyi, here is new PR which removes query string:
https://github.com/dmwm/WMCore/pull/9171

And, I agree with Alan regarding `502` error. I'm not sure where it is exactly generated and if it is in internals of CherryPy we better not to touch it, if it is returned by FE we also better not to touch it.

amaltaro:
Yuyi, if you're somewhat happy with this patch, let me know and I can backport it to the dbs branch such that you can cleanly apply this patch and run your final tests.

yuyiguo:
@vkuznet @amaltaro 
Thanks for the explanation. I will do one more testing before we commit it.

amaltaro:
New fix in #9177

PR #9176, opened 2019-05-03
add scram.getStdout method. Fix #9170

### LINKED ISSUE #9170: Make it possible to get output from commands executed with Scram()
see #9113 and #8989 
`Scram()` module in `WMCore/WMRuntime/Tools/Scram.py` was modified to report the output of the command it exectues via logging.info instead of allowing a file name to be passed as input.

https://github.com/dmwm/WMCore/blob/91254cd394ad35b73860fe0964890e06e6164ad3/src/python/WMCore/WMRuntime/Tools/Scram.py#L304-L308
https://github.com/dmwm/WMCore/blob/91254cd394ad35b73860fe0964890e06e6164ad3/src/python/WMCore/WMRuntime/Tools/Scram.py#L361-L363

But CRAB needs to get the command output as an object to process in the code, and in several instances is using the bad practice of directly accessing Scram internal structure via a refernce to `scram.stdout`.

Proposal is to have a proper public method to retrieve the needed information which is now part of the Scram() object. I will make a tested PR for this shortly

**NOTE** there is no point in filling `scram.stderr` at line 363 since command stderr is piped to stdout in line 306  ! But I will leave this small cleanup to WM Developers, if they care.

belforte:
in the end.. a trivial solution :-(

### DIFF
--- a/src/python/WMCore/WMRuntime/Tools/Scram.py
+++ b/src/python/WMCore/WMRuntime/Tools/Scram.py
@@ -389,3 +389,12 @@ class Scram(object):
             self.stdout,
             self.stderr)
         return result
+
+    def getStdout(self):
+        """
+        retrieve stdout of the command executed via the __call__ method\
+        note that in current implementation stderr of command execution
+        is piped to stdout
+        :return: a (possibly long) string
+        """
+        return self.stdout


belforte:
@lecriste FYI
@amaltaro if you accept this, kindly let me know when it is tagged, so we can start build with WMCore > 1.1.20

amaltaro:
Thanks, Stefano. How come we didn't think about it before? :)
Do you need it to be backported to a specific branch?

lecriste:
@amaltaro no need to backport, we use just the master branch.

amaltaro:
Ok, it will be in a new release on the next Tuesday. Please let me know if you need something else before that.

amaltaro:
@lecriste @belforte tag `1.2.2` has been created.

PR #9171, opened 2019-05-01
Remove query_string since it is covered by params in CherryPy.request dump

vkuznet:

PR #9178, opened 2019-05-03
fix missing format string in #9171 - dbs version

amaltaro:
dbs branch of #9177 (first commit only)

@yuyiguo can you please test this bug fix and let me know if we can merge and cut a new dbs tag? Thanks

### DIFF
--- a/src/python/WMCore/WebTools/Root.py
+++ b/src/python/WMCore/WebTools/Root.py
@@ -82,7 +82,6 @@ class WTLogger(LogManager):
         response = cherrypy.response
         inheaders = request.headers
         outheaders = response.headers
-        body = request.body
         # identify size of body from HTTP Content-Length header
         rbytes = int(cherrypy.request.headers.get('Content-Length', 0))
         if not rbytes:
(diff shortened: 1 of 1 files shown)


src/python/WMCore/WebTools/Root.py
```diff
@@ -92,10 +91,10 @@ def access(self):
             # request.rfile.rfile.bytes_read is a custom CMS web
             #  cherrypy patch not always available, hence the test
             rbytes = (getattr(request.rfile, 'rfile', None)
-                 and getattr(request.rfile.rfile, "bytes_read", None)
-                 and request.rfile.rfile.bytes_read) or "-"
+                      and getattr(request.rfile.rfile, "bytes_read", None)
```
yuyiguo: 8 spaces here instead of 4 ?

yuyiguo:
@amaltaro 
Allan, The code here are tested. I hand picked the code to my test area so I missed the missed of "s" format in #9171.  The #9171 is combined of new stuff with #9169 that has the correct string format. I apply #9169 first so my test version was correct. Thank you for finding the missing "s" error in #9171 .

This pr looked good to me, except for a small thing on the code format that I have commented on the code. It should not be any functional issue. Please commit it and give the 1.1.16_dbs branch a new tag.
Thanks,
yuyi

src/python/WMCore/WebTools/Root.py
amaltaro: that's what PyCharm decided to put in place, I guess to keep it aligned with the whole statement.

amaltaro:
@yuyiguo tag `1.1.16.dbs3` should be available. Thanks

yuyiguo:
Thanks Alan. 
I hope this is the final tag needed for May release. Making a new PR to Lina now.

PR #8979, opened 2019-01-25
Add new job classads required for monitoring

amaltaro:
Fixes #8958 (but the TaskType bit)

Still work in progress, and the ads about off/onsite need some thought.

Issue #8958, opened 2019-01-04
Condor job classads required for monitoring


amaltaro:
As a follow up of:
https://its.cern.ch/jira/browse/CMSMONIT-41
https://its.cern.ch/jira/browse/CMSMONIT-61

we need to implement/set the following classads:
* `CMS_WMTool`: with `WMAgent` value
* `CMS_SubmissionTool`: again with `WMAgent`
* `CMS_Type`: likely the same as the dashboard activity used to be. One of the following values are allowed: `(Production, Test, Tier0)`

others that will require a bit more work:
* `CMS_TaskType`: more or less the same logic as used by Benjamin's script. See https://github.com/dmwm/WMCore/issues/8802 for further details
* `CMS_PrimaryInputLocation`: "did this job ask ro local or WAN reads ? (what it got may be different, but we can maybe tell)". Values are: `(Onsite, Offsite, Mixed)`
* `CMS_SecondaryInputLocation`: "did this job ask ro local or WAN reads ? (what it got may be different, but we can maybe tell)". Values are: `(Onsite, Offsite, Mixed)`

Schema is at: https://github.com/dmwm/CMSMonitoring/blob/master/schemas/ClassAds.json

The following jobads are already in place: `CMS_WMTool`, `CMS_SubmissionTool`, `CMS_Type`

amaltaro:
Jean-Roch, confusing in which sense?

I also have a question, are there any campaigns that only read data remotely (either primary or secondary). Or those workflows with TrustPUSitelists/TrustSitelists always have a few sites whitelisted that also contain part of the data? Just trying to see where we have more "offsite" compared to "mixed" or if it's the inverse. It's going to be tricky to get it properly done in the agent...

vlimant:
confusing in the sense that you might end up with a job reading data fully locally and still having CMS_PrimaryInputLocation = Offsite

vlimant:
for campaign settings, in some sub-set of setting, it might be all in one case, but only in some small sub-set

vlimant:
and yes ; to do it right in the agent at job submission time is going to be tricky, as the what can actually happen may be different..
To get it "less wrong", you might need to make it an expression that takes into account the known data location at submission time, like CMS_PrimaryInputLocation = "GlideinSite in CMS_DataLocationSites" (whichever is the actual syntax)

vkuznet:
[...]

PR #9184, opened 2019-05-08
Add WMTool and SubmissionTool job classads

amaltaro:
Add the easy job classads from this ticket https://github.com/dmwm/WMCore/issues/8958

Easy part of #8979 , leaving only what will take more effort and investigation to the other PR.


khurtado (commented):
Looks good. A very minor suggestion for esthetics would be to change the quotes from `'WMAgent'` to `"WMAgent"`, since all the other hardcoded string classads were set that way, just to keep consistency, but not really important though.

amaltaro:
Thanks, Kenyi. I've updated the code.

amaltaro:
Also tested in my VM.

PR #9189, opened 2019-05-09
Provide application function to allow place Root cherrypy server to work with uwsgi

vkuznet:
This patch allows to start WMCore/WebTools/Root.py server under uwsgi. Since Root server requires configuration parameter it can be passed via environment. I defined as `WMCORE_WEBTOOLS_CONFIG`. Then it can be started as following:
```
# setup your own configuration
export WMCORE_WEBTOOLS_CONFIG=/path/config.py
# start Root.py under uwsgi on port 7250
uwsgi --http :7250 --wsgi-file Root.py
```

### DIFF
--- a/src/python/WMCore/WebTools/Root.py
+++ b/src/python/WMCore/WebTools/Root.py
@@ -423,6 +423,25 @@ class Root(Harness):
             server.unsubscribe()
             del cherrypy.servers[name]
 
+def application(environ, start_response):
+    "Helper function to start cherrypy Root under uwsgi"
+    inifile = os.environ.get('WMCORE_WEBTOOLS_CONFIG', '')
+    if not inifile:
+        print("### ERROR: No configuration is specified, please provide it via WMCORE_WEBTOOLS_CONFIG environment variable")
+        sys.exit('No configuration specified')
+    cfg = loadConfigurationFile(inifile)
+
+    component = cfg.Webtools.application
+    workdir = getattr(cfg.Webtools, 'componentDir', '/tmp/webtools')
+    if workdir == None:
+        workdir = '/tmp/webtools'
+    croot = Root(cfg)
+    croot._validateConfig()
+    croot._configureCherryPy()
+    croot._loadPages()
+    croot._makeIndex()
+    cherrypy.tree.mount(croot, '/', None)
+    return cherrypy.tree(environ, start_response)
 
 if __name__ == "__main__":
     parser = ArgumentParser()


yuyiguo:
@vkuznet 
`export WMCORE_WEBTOOLS_CONFIG=/path/config.py`
is the config.py same as 
**https://github.com/dmwm/deployment/blob/master/dbs/DBSGlobalReader.py** ?

amaltaro:
@yuyiguo I suggest you make further changes to the deployment scripts (manage and config), properly test it - exercising different settings - and let us know in case changes are required. Given that we're adding another component to the overall system, I'd suggest to run some benchmarks with and without uWSGI as well.

BTW, it also requires new spec files under COMP cmsdist, such that we can control the uWSGI version (and possibly uwsgitop for further monitoring).

vkuznet:
Yuyi, yes the config I'm referring is the one you use in DBS. So the changes to manage script would be to replace this line https://github.com/dmwm/deployment/blob/master/dbs/manage#L97 with something like:
```
LD_PRELOAD=$JEMALLOC_ROOT/lib/libjemalloc.so WMCORE_WEBTOOLS_CONFIG=$CFGDIR/$1 uwsgi --http :8250 --wsgi-file $DBS3_ROOT/lib/python*/site-packages/WMCore/WebTools/Root.py
```

Please note:
- I don't know yet if we need to use jemalloc, so I will try with and without it
- I think we need to have multiple lines like this for every DBS instance you run where every line will have explicitly proper port number. The problem is that port number is defined in your DBSXXX.py files and uwsgi does not read them but it needs to set port explicitly at command line. Therefore you may add something like this
```
port=`cat $CFGDIR/$1 | grep config.Webtools.port | awk '{print $3}'`
LD_PRELOAD=$JEMALLOC_ROOT/lib/libjemalloc.so WMCORE_WEBTOOLS_CONFIG=$CFGDIR/$1 uwsgi --http :$port --wsgi-file $DBS3_ROOT/lib/python*/site-packages/WMCore/WebTools/Root.py
```

Finally, you need to inspect uwsgi and define proper options/settings, e..g choose number of processes to run, number of threads, etc., taking into account that DBS will run on node which share resources with others.

vkuznet:
Here is new spec for uwsgi: https://github.com/cms-sw/cmsdist/pull/4956

vkuznet:
We can setup certain number of threads and processes which will improve throughtput of the server, and uwsgi allows to export its stats which we can use for Prometheus server and build performance graph. Here is example from uwsgi documentation:
```
uwsgi --http :9090 --wsgi-file foobar.py --master --processes 4 --threads 2 --stats 127.0.0.1:9191
```

amaltaro:
@yuyiguo if you won't have time to follow up on this initial implementation, or if you consider it to be no longer needed, please let me know and I'll close this PR.

PR #9197, opened 2019-05-15
fixed the bug that caused no logs after requests processed.

yuyiguo:


yuyiguo:
@amaltaro 
Alan, Please commit it to 1.1.16-dbs branch.  This fixed the bug you found that did not show log after the request was processed. I will test it again before committing to mast branch.

amaltaro:
@yuyiguo are you sure this fix the problem? I don't see any logical difference, just a different way to implement it. In addition to that, I have other code recommendations to give, but that can wait.

yuyiguo:
@amaltaro 
Alan, 

I tested the code and indeed it fixed the problem. The first try should get the data. The data will be "-" if nothing worked after the two trys. I left the second try there just in case. It should work in the first try. The previous code had the second try first, but it failed without any way to do the first try. That is why we missed entire log for after processing. 
Let me know if any questions. I need this for June release.

amaltaro:
I still don't get it! This line
https://github.com/dmwm/WMCore/pull/9197/files#diff-e0f21be5aa9ce07a539bfc7b6f8cae79R122

is outside any of those `if` statements, so the only way to skip that logging would be if there was a python exception in the lines above that. Were you getting an exception in the logs, or nothing at all?

yuyiguo:
@amaltaro 
Alan,
I did not get an python exception the in the log with the old code and the log was skipped. When I put the try blocks around the code, I saw the exception in the log if I printed it.  With the current try blocks, I am able to protect the code to always go to https://github.com/dmwm/WMCore/pull/9197/files#diff-e0f21be5aa9ce07a539bfc7b6f8cae79R122.
Yuyi

yuyiguo:
@amaltaro 
Alan,
This code was tested. Pleas commit it to bith master and 1.1.16_dbs.
Thanks,
Yuyi

yuyiguo:
@amaltaro 
Alan, Please merge it to 1.1.16_dbs branch too.
Thanks,
yuyi

PR #9199, opened 2019-05-15
Update CherryPy configuration of WebTools/RESTModel via WMCore configuration

vkuznet:
Based on discussion with Brian we may change CherryPy engine configuration dynamically. In particular Brian pointed to `accepted_queue_size` which controls CherryPy queue size. By default it is set to -1 which means unlimited queue. This patch provides ability to setup CherryPy configuration which will be applied to CherryPy engine, i.e. it updates cherrypy config. This is particular important for DBS server where we can apply this configuration

```
config = Configuration()
config.component_('CherryPy')
config.CherryPy.accepted_queue_size = 5 # whatever number we can use to limit the queue size
```

### DIFF
--- a/src/python/WMCore/WebTools/RESTModel.py
+++ b/src/python/WMCore/WebTools/RESTModel.py
@@ -34,6 +34,17 @@ class RESTModel(WebAPI):
         # If not set expire data after 5 mins
         self.defaultExpires = config.default_expires
 
+        # update cherrypy configuration if it is provided in configuration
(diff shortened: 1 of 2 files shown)


vkuznet:
@amaltaro , @bbockelm  please review, it seems I started from wrong branch, I'll fixed later. The code I requested is here: src/python/WMCore/WebTools/RESTModel.py

vkuznet:
Replaced with #9200

PR #9200, opened 2019-05-16
Update CherryPy configuration of WebTools/RESTModel via WMCore configuration

vkuznet:


vkuznet:
@amaltaro , @bbockelm please review this PR.

vkuznet:
Based on discussion with Brian we may change CherryPy engine configuration dynamically. In particular Brian pointed to `accepted_queue_size` which controls CherryPy queue size. By default it is set to -1 which means unlimited queue. This patch provides ability to setup CherryPy configuration which will be applied to CherryPy engine, i.e. it updates cherrypy config. This is particular important for DBS server where we can apply this configuration

```
config = Configuration()
config.component_('CherryPy')
config.CherryPy.accepted_queue_size = 5 # whatever number we can use to limit the queue size
```

amaltaro:
Valentin, this parameter is already supported by WMCore (by both REST and WebTools package).
However, DBS has a slightly different configuration, so we have to parse things in a different way instead of relying on the auto-parsing of the config file.

I have a new PR here: https://github.com/dmwm/WMCore/pull/9201
closing this one

PR #9204, opened 2019-05-17
Convert unicode to bytestring in formatCursor

### LINKED ISSUE #605: MIgration failures
Sorry to add more problems.
There have been a few, persistent, failures to migrate datasets from Global to Phys03. While it makes surely sense that migration can fail when Global can't be read, I'd like to see what exactly went wrong, if nothing else to know when to try again.

I lookeed at DBSmigration logs via vocms055, but they have no useful information, only a series of time stamps. Even when grepping for a given (failed) migration request id, I found only one line liting the Id, but no detail.

Is there maybe a verbosity level that could be changed, temporarely ?

examples of Dataset which failed to migrate:

/NonPrD0_pT-1p2_y-2p4_pp_13TeV_pythia8/RunIILowPUAutumn18DR-102X_upgrade2018_realistic_v15-v1/AODSIM

/DYJetsToLL_M-50_TuneCUETHS1_13TeV-madgraphMLM-herwigpp/RunIISummer16MiniAODv3-PUMoriond17_94X_mcRun2_asymptotic_v3-v2/MINIAODSIM

amaltaro:
Fixes https://github.com/dmwm/DBS/issues/605
Well, another proposal to fix it. This fix goes as close as possible to the SQL/database layer.

I have a couple of concerns though:
* it will apply to every single result calling DBFormatter.formatCursor. So there is some overhead here (complexity O(n^3)!)
* if data is not formatted with one of those methods handling unicode, then there is the risk to pass unicode string data to the next SQL query.

PS.: Oracle returns table/column names in capital, should we apply `.lower()`? I guess not, there wasn't anything in the previous implementation.

### DIFF
--- a/src/python/WMCore/Database/DBFormatter.py
+++ b/src/python/WMCore/Database/DBFormatter.py
@@ -137,7 +137,13 @@ class DBFormatter(WMObject):
                     cursor.close()
                     break
                 for r in rows:
-                    result.append(dict(list(zip(keys, r))))
+                    entry = {}
+                    for index in range(0, len(keys)):
+                        if isinstance(r[index], unicode):
+                            entry[str(keys[index])] = str(r[index])
+                        else:
+                            entry[str(keys[index])] = r[index]
+                    result.append(entry)
             else:
                 break
         if not cursor.closed:


bbockelm:
This breaks the ability to stream results in the future, no?

vkuznet:
Alan, I would vote to NOT do it. This code has plenty of issues by itself (e.g. it should not construct objects and put them in a list, instead it should `yield` results for streaming) and adding additional complexity may have dramatic consequences everywhere (e.g. DBS, T0, etc.)

amaltaro:
@bbockelm I don't think so. Looking at how it's used in DBS, it expands some rows and return it as a generator:
https://github.com/dmwm/DBS/blob/master/Server/Python/src/dbs/dao/Oracle/Block/List.py#L144

@vkuznet see reply above. In addition to that, this fix does NOT change anything but casting unicode to string. ALL the rest keeps exactly the same logic/behavior as before.

amaltaro:
Even the additional complexity... it's just more visible now; it was there before, just hidden by the `zip` function.

vkuznet:
Alan,
I suggest that you benchmark it, take the largest block size and see how much time
will be added if we need to parse 10k-100k+ rows.
V

bbockelm:
Gross.  The whole API is a flurry of object churn, switching between lists and generators.

Honestly - is it better to just either scrap this (introducing a `formatCursorV2` that is all generator-based) or force the caller to declare what behavior they need instead of doing it under the hood?

amaltaro:
Given that we're talking about a hot fix for DBS, I'd add the minimum amount of changes in, the safest the best. Then for the upcoming June release, we try to get an improved version of it.

Yes Valentin, I can do that. I switched gears for the moment and will get back to it later. If @yuyiguo can benchmark it, it's even better.

vkuznet:
@bbockelm , Brian I proposed full redesign of this and other modules in 2015, see https://github.com/dmwm/WMCore/pull/6337 (and before that here https://github.com/dmwm/WMCore/pull/6284) The changes I made based on pure python based iterators instead of using SQLAlchemy. The code was discussed, tested, merged and put under the rug and never get into the main tree.

@amaltaro , Alan I understand desire to do hot-fix but as history shows the proper solution may not appear and be implemented or put in place.

bbockelm:
@vkuznet - indeed, it's worthwhile - and important - to go back to this item.  However, I am reluctant to do this until the dust has settled.

amaltaro:
PR probably no longer needed. We shall work on this issue instead:
https://github.com/dmwm/WMCore/issues/9207

and a possible rework between WMCore/DBS for proper use of generators.

amaltaro:
This was a possible hot fix for DBS and seems to be no longer relevant. Please let me know otherwise such that I can reopen it.

PR #9208, opened 2019-05-20
Replace unicode in the DAO binds by byte string

### LINKED ISSUE #9207: Sanitize SQL input binds against unicode key and values
Not sure we also need to cast the keys, but if we want to make sure it won't cause problem in the future (until we migrate to python3), then we better cast both. Change should go as lower level as possible, somewhere like here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/DBCore.py#L123

Issue discussed in many places, like:
https://github.com/dmwm/DBS/pull/606
https://github.com/dmwm/DBS/issues/605#issuecomment-493223306

By the way, with this change in place, it means we no longer need to sanitize the SQL output results, as done here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Database/DBFormatter.py#L99

amaltaro:
Fixes #9207 

Whenever a DAO gets executed (so `processData` method is called), parse all the binds provided and convert any unicode string key/value to byte string. Any other data type remains untouched.

This has been tested for both WMAgent Oracle and MySQL database backend.
To be tested and discussed within the DBS context and evaluate whether the performance is acceptable.

### DIFF
--- a/src/python/WMCore/Database/DBCore.py
+++ b/src/python/WMCore/Database/DBCore.py
@@ -6,6 +6,7 @@ Core Database APIs
 
 
 """
+import sys
 from copy import copy
 
 from Utils.IteratorTools import grouper
@@ -90,10 +91,8 @@ class DBInterface(WMObject):
         """
 
         s = s.strip()
-        if s.lower().endswith('select', 0, 6):
-            """
-            Trying to select many
-            """
+        if s.lower().startswith('select', 0, 6):
+            ### selecting many
             if returnCursor:
                 result = []
                 for bind in b:
@@ -104,13 +103,13 @@ class DBInterface(WMObject):
                     resultproxy = connection.execute(s, bind)
                     result.add(resultproxy)
                     resultproxy.close()
+        else:
+            ### inserting / updating many
+            result = ResultSet()
+            for sliceBinds in grouper(b, self.maxBindsPerQuery):
+                res = connection.execute(s, sliceBinds)
+                result.add(res)
 
-            return self.makelist(result)
-
-        """
-        Now inserting or updating many
-        """
-        result = connection.execute(s, b)
         return self.makelist(result)
 
     def connection(self):
@@ -120,13 +119,14 @@ class DBInterface(WMObject):
         return self.engine.connect()
 
 
-    def processData(self, sqlstmt, binds={}, conn=None,
-                    transaction=False, returnCursor=False):
+    def processData(self, sqlstmt, binds=None, conn=None, transaction=False,
+                    returnCursor=False, encodeType="ascii"):
         """
         set conn if you already have an active connection to reuse
         set transaction = True if you already have an active transaction
 
         """
+        binds = binds or []
         connection = None
         try:
             if not conn:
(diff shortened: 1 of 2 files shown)


src/python/WMCore/Database/DBCore.py
```diff
@@ -196,3 +197,20 @@ def processData(self, sqlstmt, binds={}, conn=None,
...
+    def convertBinds(self, bindsList, encodeType):
+        """
+        Convert any unicode string key/value to its correspondant byte string.
+        Let an exception be raised in case the string cannot be encoded in ascii.
+        :param bindsList: list of dictionaries corresponding to the binds
+        :return: updates are made in-place, but a reference is returned anyways
+        """
+        for item in bindsList:
+            for k in item:
+                ### TODO: this will have to be re-thought for python3 migration
```
vkuznet: How about this instead of comment
```
import sys
....
if  sys.version.startswith('2.'): # python 2 code
```

amaltaro: I added a check for the major python version, but outside of the loop. Thanks

amaltaro:
And here is a script to evaluate the time overhead for this check:
https://gist.github.com/amaltaro/f782d53df503c8d11c0c38c8a9ccebf7

as we can see, for lists with 100k binds, the overhead is quite small, having or not unicode keys/values.

When we hit 1 million binds though, the picture changes and the overhead goes up to a few seconds.
Not a big deal for WMAgent, hopefully 1M binds is not a common case for DBSServer(?)

vkuznet:
Alan,
that's nice but not realistic yet. Do you mind to change your values to actual
LFNs/dataset/block names and re-run it. In other words does the size of the
string matters for encoding.
V.

amaltaro:
Sure, I put more realistic key/value pairs and re-run those tests. Elapsed time went up by <10%. Have a look at the numbers.

amaltaro:
@yuyiguo @vkuznet please have a further look at it and leave your review. I'm planning to merge it by this Wednesday and release a new tag for Lina.

amaltaro:
@vkuznet @yuyiguo this is what I said in the meeting that is still not merged and needs a follow up.

If Yuyi is planning to update DBS in HG1906 (the current cycle), then we need to get this reviewed and in asap. I understood this was the plan for HG1906, including the update back to the latest SQLAlchemy.

yuyiguo:
@amaltaro 
Alan,
Give me some time to review the code. Too much things to take care right now.
Yuyi

yuyiguo:
@amaltaro 
Alan, I don't think I have time to get this in June release. The throttling stuffs seems more complicated than I thought. I think throttling is important to get DBS stable. I will leave this to July.
Yuyi

amaltaro:
@yuyiguo it would be great if we could resume working on this fix and get some tests performed on your end. Depending on the results, DBS could update to the latest SQLAlchemy (same as in use by WMCore)

vkuznet (commented):
I"m ok with these changes.

vkuznet (approved):

amaltaro:
test this please

amaltaro:
@yuyiguo we don't need to merge it for testbed because none of the CMSWEB WMCore services use it; however, I'm inclined to get this merged within two weeks for the next WMAgent release.

This will have to go in sooner or later and the PR has been inactive for a couple of months now.

amaltaro:
It's been more than 2 years that we have this PR opened, and still no final agreement was reached. Taking into account that DBS is also going to move to Golang technology, I consider this feature change to be no longer relevant.

Please do let me know if it's still relevant.

PR #9210, opened 2019-05-21
Port from CRABServer code Throttled classes; add unit test for them - dbs branch

amaltaro:
dbs branch version of #9158  and #9211

@yuyiguo please let me know if you need anything else, otherwise I'll merge and cut a new dbs tag.

PR #9211, opened 2019-05-21
Code re-factoring and add timerange for user activities under throttling

vkuznet:
After review of throttled code it seems to me that it never worked since `__enter__` and `__exit__` functions kept user counter intact under context clause.

I re-factored code and add the following features:
- add time range (in seconds) for `make_throttled` method which keeps track of user activities within that range, i.e. how many calls users made in last 2 seconds
- increment user call activities within specified time range
- reset user activities when user pass the time range, i.e. if user made N calls and exceed its limit then it will be blocked, but once throttler passed time range of last user call user will be granted again access to throttled API

Here is how it should be used
```
# initialize UserThrottle class with hard limit, i.e. total number of calls it will allow
thr = UserThrottle(limit=5)
# decorate an API with given time range, i.e. time range controls user access to that API within UserThrottle limit

@thr.make_throttled(trange=2)
def api():
    return time.time()
```

The unit test is adjusted too.


bbockelm (commented):
I’m not sure I’m following you - the counter increments at the beginning of the context manager and decrements at exit.  What’s wrong?

vkuznet:
Brian,
if I understood code correctly it is not doing what we wanted. Let's break it
down.

The code increments and decrements local counter when particular
function/API was executed. It means that the counter will go above 1 only
in a case when multiple calls of that function/API happen at the same time
and still executed within context.

While the behavior we discussed with Yuyi should involve monitoring of clients
usage and give fair access to all users of this API. As such I introduced a time
range during which we collect statistics of a given user and prevent API from
over usage from that client. For example, if one client sends 10K requests to
the API within the specified time range, e.g. grid jobs which will query API
almost at the same time, then we can block such usage.

I asked Yuyi to perform tests of both scenarios using concurrency tool. At least
we'll compare both implementations.

Valentin.

belforte:
@vkuznet I am confused by how the time limit is indicated to work. Sounds like if I hit the limit I am allowed to add more after some time even if previous ones have not completed yet (throttling is reset). Do I understand correctly ? And if so, is that really  desirable ?

I am also confused by what the goal is.
I thought that the idea was to avoid someone from doing more than N (<10 ? N<2?) simultaneous calls. 
Putting an overall limit on how much someone can use the system even if nobody else is using it, is a different thing.
I.e. are you and @bbockelm discussing implementation, or policy ?

amaltaro:
> For example, if one client sends 10K requests to the API within the specified time range, e.g. grid jobs which will query API almost at the same time, then we can block such usage.

This seems to be already supported by the throttling mechanism, which considers the amount of parallel requests made, disregarding the time range.

I confess I didn't pay too much attention to the implementation details in the beginning, but now going through the original code, is it correct to say that:
* the counter implemented is based on **user** AND **cherrypy thread** (usage of threading.local())
* that said, there is no global user counter for the application
?

I also see a few points that probably need to be fixed in the original code:
* context manager always executes the __exit__ method, even when an exception is raised, so we decrement user counter once too many (when limit is reached)
* I think returning a `429 Too Many Requests` instead of `500` would make much more sense.
* let's use `logger.setLevel()` and let that `if self.debug` statement go away.

vkuznet:
The original code provides throttling mechanism based on function execution time.
That said, the conditions for throttling are different then time based approach
I proposed in this PR.

In former case the conditions happens when multiple concurrent calls from the
same client access the function while it still executes. If function is short
lived the chances for throttling to happen is minimal.
In later approach the conditions happens based on user pattern during specified
time range, i.e throttling happens if client exceeds a limit based on number
calls in last time interval.

About local/global counters. In former case it is local counter per thread.
In later case it is global counter per throttling instance, but user can have
multiple instances of throttling class and therefore global counter will be
applied per API basis or to entire application (if one instance is used).

Yuyi is testing/benchmarking both implementation and I'll suggest that we'll
for her to finish and publish results. I bet you'll find them interesting.

Based on your interests I think I'll go yet another round with the code and
make both implementation available.

vkuznet:
I put back original implementation and now Throttling module will provide both. Here how we can use them:

**Counter based throttling**
```
from Utils.Throttling import global_user_throttle
@global_user_throttle.make_throttled(trange=2)
def api():
    # define your api logic here
```

**Time range based throttling**
```
# initialize UserThrottleTime class with hard limit, i.e. total number of calls it will allow
from Utils.Throttling import UserThrottleTime
thr = UserThrottleTime(limit=5)
@thr.make_throttled(trange=2)
def api():
    # define your api logic here
```

yuyiguo:
@vkuznet 
I am testing the time ranging based throttling, the counter based is down. Will publish them soon.

Could you make the count/limit configurable in the counter based too?
Thanks,
Yuyi

vkuznet:
> @vkuznet
> I am testing the time ranging based throttling, the counter based is down. Will publish them soon.
> 
> Could you make the count/limit configurable in the counter based too?
> Thanks,
> Yuyi

Yuyi, the counter based one is already have limit configuration, you may use it as
```
from Utils.Throttling import UserThrottle
thr = UserThrottle(limit=5) # adjust if necessary
@thr.make_throttled()
def api():
    # define your api logic here
```

So I think we should remove `global_user_throttle` completely, but to avoid issues and backward compatibility with CRAB codebase I kept it intact. Just instantiate the class you need in DBS and you can apply to `UserThrottle` or `UserThrottleTime` specific limits. Therefore both `limit` and `trange` are now can be obtained from external configuration and passed as values to these classes and `make_throttled` decorator.

yuyiguo:
@vkuznet @amaltaro @h4d4 
Got it. Thanks Valentin.
Alan, Would you please merger this PR to 1.1.16_dbs branch before you leave for the day? Please tag this branch too.

I will get a PR to Lina over the weekend so DBS can be include in the Monday's testbed  deployment. I am still testing the throttling to get the whole picture so we can decide which throttling version to go and what the initial parameters to set.

belforte:
FIne with to to do the one line change in CRAB to avoif  `global_user_throttle` whenever a new version of WMCore is tagged. So feel free.

vkuznet:
Stefano, now it is gone. Feel free to adjust CRAB code once this PR will be
merged. I provided examples of both throttling use cases in docstring.

I'm working with Yuyi to provide full test results for both implementations.
She will post them here once they will be completed.

amaltaro:
@yuyiguo code has been backported, merged and tag 1.1.16.dbs4 created, see #9210

vkuznet:
@amaltaro , Alan can this be merged that I can provide code on top of these changes? @yuyiguo requested to add white/black list to throttle classes and I can do it in separate PR on top of this one once it will be merged into the master.

src/python/Utils/Throttled.py
```diff
@@ -16,41 +36,131 @@
...
         ctr = self.throttle._incUser(self.user)
-        if self.debug:
-            msg = "Entering throttled function with counter %d for user %s" \
-                    % (ctr, self.user)
-            self.throttle.logger.debug(msg)
+	msg = "Entering throttled function with counter %d for user %s" \
+		% (ctr, self.user)
+	self.throttle.logger.debug(msg)
         if ctr >= self.throttle.getLimit():
             self.throttle._decUser(self.user)
```
amaltaro: I still think we should not decrement the counter here. It's already getting decremented when it leaves the context manager. 
E.g.: limit is set to 3 and user put a 4th request in the system, it enters the context manager and gets incremented to 4, then it finds it has exceeded the limit, decrements back to 3, leave the context manager and decrements it back to 2.

Another thing I just noticed, if limit is 3, then we should allow a 3rd call, this line doesn't:
`if ctr >= self.throttle.getLimit():`
it should be just greater than, not greater or equal than.

src/python/Utils/Throttled.py
```diff
@@ -16,41 +36,131 @@
...
+    def __init__(self, throttle, user, trange):
+        self.throttle = throttle
+        self.user = user
+        self.trange = trange
+
+    def __enter__(self):
+        "Define enter method for `with` context"
+        ctr = self.throttle._incUser(self.user, self.trange)
+        if ctr > self.throttle.getLimit():
+            self.throttle._decUser(self.user)
```
amaltaro: same comment as above.

amaltaro (changes requested):
Valentin, please first fix those pylint issues reported (and do not use tabs for indentation).
I have also put a comment in the code.

src/python/Utils/Throttled.py
vkuznet: fixed

src/python/Utils/Throttled.py
vkuznet: not fixed since __exit__ does not decrement counter. Here the decrement call is correct

vkuznet:
Alan, I fixed pylint change and for the record I don't use tabs, but when I port somebody code elsewhere it may have them.

src/python/Utils/Throttled.py
```diff
@@ -16,41 +36,130 @@
...
+    def reset(self, user):
+        "Reset user activities, i.e. access counter and last time of access"
+        self.users[user] = 0
+        self.users_time[user] = time.time()
+
+    def _incUser(self, user, trange=60):
+        "increment user count"
+        with self.lock:
+            self.users.setdefault(user, 0)
+            self.users_time.setdefault(user, time.time())
```
bbockelm: Just a small python suggestion as `setdefault` returns the resulting value in the `dict`

```suggestion
            last_time = self.users_time.setdefault(user, time.time())
```

vkuznet: done

src/python/Utils/Throttled.py
```diff
@@ -16,41 +36,130 @@
...
+            raise cherrypy.HTTPError("429 Too Many Requests", msg)
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        "Define exit method for `with` context"
+        pass
+
+
+class UserThrottleTime(object):
+    """
+    UserThrottle class defines how to handle throttle time range based mechanism.
```
bbockelm: It would be really useful to say what the "time range based mechanism" is.  I think you've captured it in the ticket, but it'll be a bit hard to find the ticket in 5 years when we need it next.

Personally, I remain a bit confused as to the benefit of the time-based mechanism compared to the original one, which intended (bugs notwithstanding) to limit the number of concurrent active queries.

What resources is it protecting?  What access patterns does it limit?  Looking at the code, it appears that with `trange = 2`, then a user can effectively launch as many heavyweight queries as they want in parallel if they match the heavy query with two short queries.  This is because each query start increases the counter by 1 and, if matched with two short queries at the beginning of the time range (so a total increment of 3), these lines:

```
step = int(trange - abs(time.time()-last_time))
self.users[user] -= step
```

will cause the counter to decrement by approximately 4, at which point the `last_time` variable is reset to `time.time()`.

Hence, it appears I can create an infinite stream of heavyweight blocking queries if I'm willing to do short queries in a separate thread to push down the counter by more than 1.

Why was this algorithm implemented as opposed to something more common such as an exponential moving average?

(Again, it's more useful to put the explanations in the class comments ... that's what started the long question in the first place!)

vkuznet:
Brian,
I added description to the module docstring.

The main difference as I see is the following:
- the counter based approach depends on time execution of the underlying API,
i.e. counter incremented/decrement based on time execution of the passed
function
- the time based approach counts total accesses within given time range.

As such the condition of throttling threshold are differ. Per my discussion with
Yuyi originally she was in favor of time based approach that's why I added it.
But, I also advised her to perform proper benchmark tests to justify which
approach is better to apply.

Best,
Valentin.

src/python/Utils/Throttled.py
vkuznet: I replied already below, but comment was not put in this thread.

amaltaro:
Trying to decrease the amount of PRs opened in WMCore...

Perhaps @yuyiguo and @belforte can comment on which algorithm you're using (or want to) such that we can converge.

If my memory doesn't fail me, it looks like each throttling algorithm serves a different purpose:
* one protects against many short requests (quick response time)
* the second protects against not that many but long requests (protection against bad and slow APIs)

It would be great if we could converge. If we see the need to have both in the WMCore repository, that's fine with me as well.

belforte:
So far I am of the opinion that the only problem for backends are the slow queries. If we need to throttle too maby short ones (why, if we can server them ?) I'd think it should be done in the FE. 
But it is not clear to me if the FE will disappear with CMSWEB migration in the k8s world, in which case we  may end up in needing both. In the end that something was never a problem so far, does not mean that it never will.
What's the drawbacl of having both options implemented ? Of courese it needs some good description of the API so one knows what we are asking for.

amaltaro:
@vkuznet Valetin, when it's convenient to you, please resolve these conflicts and I will get this PR merged, it's been lingering around for too long.

amaltaro:
Thanks Valentin (the failed unit test is unrelated and it has been already fixed in another open PR).

@belforte @yuyiguo just a heads up that it's going to show up starting in WMCore 1.4.3.pre3

belforte:
thanks Alan, I have an issue on HOLD to adopt this, will do after we rejoin master WM branch. I want to get rid of 1.3.6.crab ASAP.
Please remove the "Do not merge yet" label from this, so when I look at my issues I am not fooled into waiting

amaltaro:
Good catch, label removed. Thanks

PR #9212, opened 2019-05-21
Add email alerting capability

### LINKED ISSUE #9167: Implement email notification for agent draining status transitions
I think it would be useful if the agent would send an email notification every time:
* an agent is automatically put IN or OUT of drain
* an agent is manually (user) put IN or OUT of drain
* an agent goes IN or OUT of speed drain.

Recipient email address has to be configurable.

goughes:
Fixes  #9167

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -64,12 +64,16 @@ workflowArchiveTimeout = 3600
 # For setting the general log level of the components
 globalLogLevel = 'INFO'
 
+# Contact information
+# Used for email alerting
+contactName = "cms-service-production-admins@cern.ch"
+
 # Nothing beyond this point should need to be changed.
 config = Configuration()
 
 config.section_("Agent")
 config.Agent.hostName = serverHostName
-config.Agent.contact = "cms-comp-ops-workflow-team@cern.ch"
+config.Agent.contact = contactName
 config.Agent.teamName = "REPLACE_TEAM_NAME"
 config.Agent.agentName = "WMAgentCommissioning"
 config.Agent.agentNumber = 0
@@ -296,6 +300,12 @@ config.Alert.controlAddr = "tcp://127.0.0.1:6559"
 
 # mysql*Poller sections were made optional and are defined in the
 # wmagent-mod-config file
+
+# Email alert configuration
+config.section_("EmailAlert")
+config.EmailAlert.toAddr = [contactName] # additional emails can be appended to the list
+config.EmailAlert.fromAddr = "noreply@cern.ch"
+config.EmailAlert.smtpServer = "localhost"
 
 config.component_("AnalyticsDataCollector")
 config.AnalyticsDataCollector.namespace = "WMComponent.AnalyticsDataCollector.AnalyticsDataCollector"
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
@@ -143,7 +143,7 @@ class DrainStatusPoller(BaseWorkerThread):
         # update the aux db speed drain config with any changes
         if updateConfig:
             self.agentConfig['SpeedDrainMode'] = True
-            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, self.agentConfig)
+            self.reqAuxDB.updateWMAgentConfig(self.config.Agent.hostName, self.agentConfig)
 
         return
 
@@ -159,7 +159,7 @@ class DrainStatusPoller(BaseWorkerThread):
                 if key in self.validSpeedDrainConfigKeys and v['Enabled']:
                     speedDrainConfig[key]['Enabled'] = False
 
-            self.reqAuxDB.updateAgentConfig(self.config.Agent.hostName, self.agentConfig)
+            self.reqAuxDB.updateWMAgentConfig(self.config.Agent.hostName, self.agentConfig)
         return
 
     def checkSpeedDrainThresholds(self):
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
@@ -8,12 +8,13 @@ from __future__ import division
 __all__ = []
 
 import logging
+import copy
 from Utils.Timers import timeFunction
 from WMComponent.AgentStatusWatcher.DrainStatusAPI import DrainStatusAPI
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
 from WMCore.Services.PyCondor.PyCondorAPI import PyCondorAPI
-
+from Utils.EmailAlert import EmailAlert
 
 class DrainStatusPoller(BaseWorkerThread):
     """
@@ -31,9 +32,10 @@ class DrainStatusPoller(BaseWorkerThread):
         self.drainAPI = DrainStatusAPI(config)
         self.condorAPI = PyCondorAPI()
         self.agentConfig = {}
+        self.previousConfig = {}
         self.validSpeedDrainConfigKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
-
         self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
+        self.emailAlert = EmailAlert(config)
 
     @timeFunction
     def algorithm(self, parameters):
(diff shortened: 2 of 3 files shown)


amaltaro:
Cool! Does it work as it is? We then have to find out what SMTP server to use for FNAL nodes.

goughes:
I split this into two commits. The first adds the email alert class. The second adds alerting to the drain thread.

I had to modify the drain thread class a bit. The aux db is the authority of drain status of the agent, but I needed to add local variables to track and compare what the agent's current state was. These are used to determine when to alert.

Last thing will be to update the "to" address in the config to an appropriate value.

goughes:
I tested each drain mode case on my agent by entering and exiting both user and agent drain mode and received the alerts. Since my agent had no jobs, during the next drain status poller cycle I received a email saying the agent entered or exited speed drain mode.

etc/WMAgentConfig.py
```diff
@@ -297,6 +297,12 @@
 # mysql*Poller sections were made optional and are defined in the
 # wmagent-mod-config file
 
+# Email alert configuration
+config.section_("EmailAlert")
+config.EmailAlert.toAddr = ["egough@cern.ch"] # additional emails can be appended to the list
```
amaltaro: How about setting "cms-service-production-admins" as `contactAddr` and then assign this variable to this line and also to the L72 (config.Agent.contact)?

src/python/WMCore/Alerts/EmailAlert.py
```diff
@@ -0,0 +1,40 @@
+"""
+EmailAlert - send alerts via email
+"""
+
+from __future__ import division
+import smtplib
+import logging
+
+
+class EmailAlert:
```
amaltaro: Eric, given that it does not rely on any WMCore related code, how about we move it under src/python/Utils ? It would be easier for others to import from WMCore

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -73,6 +78,39 @@ def getDrainInfo(cls):
         """
         return cls.drainStats
 
+    def checkDrainStatusChanges(self):
```
amaltaro: For this method, I have a couple of comments:
a) the agent might have had multiple drain transitions, and in this case we would alert only on the last one evaluated.
b) I think we can reduce this code by iterating over the drain keys and using self.getattr(blah.lower()) or sort of to evaluate things in a loop and send an email alert from inside the loop (such that we catch multiple cases).

The message could even be: `"Agent had a drain status transition. %s = %s" % key, value` or something like that.

amaltaro (changes requested):
@goughes sorry for the belated review. I made a few comments along the code. Can you please update this PR and/or ask further questions in case something is not clear or you have a better idea?
Thanks!

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
goughes: At first I thought we would need to handle multiple transitions, for instance user drain and speed drain. But, these can only be caught if we make another call to aux db somewhere in the DrainStatusPoller algorithm or in checkDrainStatusChanges, which I was trying to avoid.

For example:
1. User sets aux db to UserDrainMode=True when jobs are < Speed drain thresholds
2. DrainStatusPoller algorithm will run and query aux db caching current drain statuses
3. Agent enters speed drain mode and updates aux db

In this case the agent has an out of date version of aux db contents (self.agentConfig) and we can't make the status comparison to alert on. Even if I run checkDrainStatusChanges at the end of the DrainStatusPoller algorithm function, it won't pick up the speed drain changes since my state comparisons used the out of date config.

If it is OK to make an additional call to the aux db in checkDrainStatusChanges, then doing multiple transitions in a single alert is definitely doable. I'll just move that function to the end of the DrainStatusPoller algorithm.

I guess there could be a rare case where the agent would enter both UserDrainMode and AgentDrainMode between DrainStatusPoller runs as well.

amaltaro: I'm not sure I completely understood the case you pointed out, but let me try it from another angle.

We have the local object variable containing the drain statusES of the previous thread run. In the line 44 we fetch data from ReqMgrAux:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py#L44

which then we compare inside this function to see whether there were changes; if there were, create an alert.

Then the missing part is, along the algorithm() executing of this Drain thread, we need to trigger another email in case there are further drain transitions. And last but not least, also update the object variable for the next cycle.

IF we don't trigger an alert in other places where a transition can happen (based on thresholds and etc), then we'd receive an alert one cycle later.

Does it make sense?

goughes:
The code is a bit better now, but like I mentioned above, I would need to query the aux db again in algorithm to be able to trigger one alert for changes to both AgentDrainMode and SpeedDrainMode (or UserDrainMode and SpeedDrainMode). Maybe we can go through this logic on a quick call.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -32,22 +32,26 @@ def __init__(self, config):
...
         self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
+        self.emailAlert = EmailAlert(config)
 
     @timeFunction
     def algorithm(self, parameters):
         """
         Update drainStats if agent is in drain mode
         """
         logging.info("Running agent drain algorithm...")
+        previousConfig = self.agentConfig
```
amaltaro: This is doing a reference copy, not a value copy. You might want to use something like `copy.deepcopy` for this assignment such that a new dict object gets created.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -73,6 +78,28 @@ def getDrainInfo(cls):
...
+        """
+        message = ""
+        drainStatusKeys = ['UserDrainMode', 'AgentDrainMode', 'SpeedDrainMode']
+
+        if not previousConfig:
+            return
+
+        for key in drainStatusKeys:
+            if previousConfig[key] != self.agentConfig[key]:
+                message += "Agent had a drain status transition. %s = %s\n" % (str(key), str(self.agentConfig[key]))
```
amaltaro: How about replacing that dot by "to"? So
`Agent had a drain status transition to: `...

amaltaro: BTW, where is the host name visible? Maybe in the sender? If not, then we should update it to actually:
`agent %s had a drain status transition to: % host`
replace `host` by the var hosting the agent name :)

amaltaro (changes requested):
@goughes Erik, sorry for the late review. Changes are much cleaner now, but... we will always get a notification delayed by 1 thread cycle.

What we could do perhaps is:
* evaluate the changes between new and previous config
* once evaluation is done, assign your new config to the previous one
* run the whole algorithm/logic of the Drain thread
* and just before completing/exiting the cycle, call again that method to compare the final new config against the previous one.
  * here I assume that every update call to ReqMgrAux db, is preceded by a local in-memory update, no?

Thanks

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
goughes: The agent hostname is in the email subject

goughes:
I missed these changes to DrainStatusPoller where you added updates to the local agent config when the speed drain mode changes: 05d0613c0bfd38312c32162fad11df08cc151ec9

I just did the same thing on this branch, which made it easier to be able to notify on all the drain status transitions at one time.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -32,22 +33,26 @@ def __init__(self, config):
...
+        self.emailAlert = EmailAlert(config)
 
     @timeFunction
     def algorithm(self, parameters):
         """
         Update drainStats if agent is in drain mode
         """
         logging.info("Running agent drain algorithm...")
+        # make a copy of the previous agent aux db configuration to compare against later
+        previousConfig = copy.deepcopy(self.agentConfig)
```
amaltaro: Hmm, I see the following scenario:
cycle 1: component goes through the normal cycle
cycle 2: component fails to fetch the agent config from ReqMgrAux
cycle 3: an empty list gets assigned to previousConfig (since it failed to fetch the config in the previous cycle).

In order to properly handle such scenarion, we should then have something like:
```
if self.agentConfig:
    previousConfig = copy.deepcopy(self.agentConfig)
```

Have I missed anything?

amaltaro (changes requested):
Almost there, Erik. I made a couple of suggestions, please go ahead if you agree with them.
I'd say it's also time to squash and rebase those commits. Thanks

goughes:
Thanks for the good comments Alan. The only other idea I had was switching previousConfig to an instance variable instead of passing a local variable around between two methods.

amaltaro:
I had the same thought, but was unsure whether it makes the code better or not... If you feel like pushing yet another commit, then go ahead ;)

amaltaro:
@goughes Erik, what's the status of this fix from your side?

goughes:
Just gave this a final test on a new deploy of 1.2.4.patch1 and things works as expected.

amaltaro (approved):

amaltaro:
I tested it both in a CERN and a FNAL agent and it works like a charm, thanks Erik!

amaltaro:
Hmm, I forgot asking you to squashing the first 2 commits :(

goughes:
Oh well... one commit adds the ability to do email alerts, one commit applies it to the drain thread. Maybe those should have been separate PRs anyways. Too late now! :)

PR #9216, opened 2019-05-23
MS Transferor code re-factoring and clean-up according to proposal specs

vkuznet:

PR #9217, opened 2019-05-24
Dont perform write operations from MS Transferor

amaltaro:
Change introduced in this PR: https://github.com/dmwm/WMCore/pull/9140

Let's make sure central production services don't rely on MS Transferor until it becomes solid.
This transition is taken care by ReqMgr2 CP thread, from this PR: https://github.com/dmwm/WMCore/pull/9209

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Transferor.py
+++ b/src/python/WMCore/MicroService/Unified/Transferor.py
@@ -350,7 +350,8 @@ class MSManager(object):
             self.logger.debug('%s changing %s status of record %s to req_status=%s' \
                 % (prefix, request, rec, req_status))
             if rec.get('reqStatus', None) != req_status:
-                self.svc.reqmgr.updateRequestStatus(request, req_status)
+                ### FIXME: we're still not executing write operations from MS
+                #self.svc.reqmgr.updateRequestStatus(request, req_status)
                 self.store.update(request, {'reqStatus': req_status})
             if req_status == 'staged':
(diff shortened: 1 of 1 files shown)


vkuznet:
Alan, before merging PRs related to MS please coordinate this with me to avoid work duplication. I already submitted another major code-refactoring yesterday in #9216 and such changes should go there rather in master branch.

vkuznet:
I also doubt that such change is useful since it eliminates the purpose of MS code which we should test. Instead of commenting out the code which we need to test we should not use production reqmgr2 instance that's it. But we do need to test all transitions that they're working.

amaltaro:
When we discussed and planned how to roll this to production, we said it would initially only read data and run the transfer logic, without making any transfer. So that line shouldn't be there and it almost made it to testbed (which would cause a bunch of errors in the logs with 2 different components trying to do the same action).

In short:
1. we get all pieces in place
2. we run the service in read mode only
3. we adapt anything that needs to be adapted
4. we enable write (or enable write for something).

I'll update this wiki:
with these deployment details and how we'll push it to production.

vkuznet:
Then let's do it properly, instead of fixing code back and forth,
I've added readOnly external configuration parameter with True as default.
See https://github.com/dmwm/WMCore/pull/9216/commits/f203e862b195d03501f0f5e4c3e5117d63436cfa

This allows to both deploy code, test it and use in production.

PR #9230, opened 2019-05-28
Use ReqMgrAux service to get campaign name for given request

### LINKED ISSUE #9228: Persist the campaign configurations in couch
List of actions to be taken here are:
* fetch all the current campaigns available in the Unified database (or repository?)
* insert every single of them in ReqMgrAux (one time insert for now)
* we already have REST APIs to insert/update/retrieve them, so we're all set on that respect
* create a default campaign configuration such that we can use in testbed and for our tests (whether we use it or not has to be configurable, we do NOT want to fallback to it in production)

However, before taking this issue, we better work on a campaign schema and make sure we have reasonable data stored in the Campaign collection, tracked in this issue:
https://github.com/dmwm/WMCore/issues/9139

vkuznet:
Fixes https://github.com/dmwm/WMCore/issues/9228

#### Status
in development

#### Description
implement how to get campaign configuration for given request

YES

#### Related PRs
https://github.com/dmwm/WMCore/issues/9228

no

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Transferor.py
+++ b/src/python/WMCore/MicroService/Unified/Transferor.py
@@ -198,7 +198,8 @@ class MSManager(object):
 
     def requestCampaign(self, req):
         "Return request campaign"
-        return 'campaign_TODO' # TODO
+        cname = "" # TODO: we should get it somehow
+        return self.svc.reqmgraux.getCampaignConfig(cname)
 
     def requestConfiguration(self, req):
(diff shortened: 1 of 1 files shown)


amaltaro:
Given that there is a conflict now, I suggest we make these changes once we have the campaign documents available in the database (when we'll have to adapt the transferor logic to the new keys anyways).

vkuznet:
not valid anymore, closing

PR #9231, opened 2019-05-28
Replace all print statements with logger

### LINKED ISSUE #9091: Set proper rotating log for ReqMgr2MS and replace print statements
Title says it all.

vkuznet:
Fixes #9091 

#### Status
In development

#### Description
Replace all print statements with logger

YES

#### Related PRs
#9091 

no

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -6,7 +6,7 @@ Original code: https://github.com/CMSCompOps/WmAgentScripts/Unified
 """
 
 # futures
-from __future__ import print_function, division
+from __future__ import division
 
 # system modules
 import os
@@ -15,6 +15,7 @@ import json
 import time
 import math
 import urllib
+import logging
 
 # py2/py3 modules
 # from future import standard_library
@@ -169,8 +170,12 @@ def getEventsLumis(dataset, blocks=None, eventsLumis=None):
     data = eLumis.get(dataset, {'num_event':0, 'num_lumi':0})
     return data['num_event'], data['num_lumi']
 
-def getComputingTime(workflow, eventsLumis=None, unit='h'):
+def getComputingTime(workflow, eventsLumis=None, unit='h', logger=None):
     "Return computing time per give workflow"
+    if not logger:
+        logger = logging.getLogger('reqmgr2ms:Common')
+        logger.setLevel(logging.DEBUG)
+        logging.basicConfig()
     cput = None
 
     if 'InputDataset' in workflow:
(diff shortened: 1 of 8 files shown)


vkuznet:
Alan, please review and merge.

amaltaro (commented):
Valentin, can you please check whether we could use this logging mechanism:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMLogging.py#L37

with the MS? That way we can consolidate the logging format and also avoid code duplication.

It's used by the CherryPy threads, like:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/CherryPyPeriodicTask.py#L27

and the log_file name format is defined like:
https://github.com/dmwm/deployment/blob/master/reqmgr2/config.py#L150

You'll probably want to add a `logLevel=20` (INFO) to the `getTimeRotatingLogger` function such that it's configurable.

vkuznet:
Alan,
how about do this in separate PR. So far you asked to convert print to logger
and I did that. Now you're asking to spend another round to convert from logger
to custom WMCore based one. If you want to consolidate under custom classes
do it in separate PRs.

And, for the record, the more you use or rely on custom/3rd party software stack,
the less portable the code becomes (e.g. I honestly can't see why WMCore needs
to use CherryPy threads instead of threads from a standard library, which by
itself creates an additional work to port REST server from CherryPy to 
another framework). But it is another story.

amaltaro:
If you look at the WMLogging module that I pointed out, you can see it's completely based on the standard library and it doesn't rely on cherrypy. The CP threads thing was just an example.

Converting from print to logging doesn't mean we can start duplicating code, and before you create yet another function to create/return a logger object, I pointed you to a code that's ready to use.

vkuznet:
Alan,
I'm hear you, but proposed WMLogging getTimeRotatingLogger function does
not fit all use-cases.  The reason is that it requires to pass configuration
file and such configuration may not be available in all circumstances. Therefore
it is not standard library behavior which does not require neither custom
configuration or external file to initialize the logger.

For instance, in plenty of auxiliary functions, see
https://github.com/dmwm/WMCore/pull/9231/files#diff-b6d3c0dff667aefd66ffe861a26ca8a8R173
or
https://github.com/dmwm/WMCore/pull/9231/files#diff-2d0f7daa8c620a6ce74e6f76c1543fd4R57
there is no configuration file passed around. But it is desired to have logging
capabilities. I did change the code (which I didn't wrote btw) to have the
logger instead of printouts, but in order to adapt WMLogging we need extra
steps.

Therefore, to address this properly, we need two changes:
- one change for WMLogging to make **optional** logFile such that it will
satisfy standard library behavior and
- changes across MS code to adapt getTimeRotatingLogger function, especially
in places where no configuration files are available.

That's why it is better to make it in separate PR.

Best,
Valentin.

amaltaro:
@vkuznet can you please fix this conflict?

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Ok, @vkuznet . We'll get back to this logging and make it standard in the future. 
I ran some quick tests and everything looks Ok, please squash those commits at your convenience and we can merge it.

amaltaro:
test this please

vkuznet:
squash is done

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

PR #9236, opened 2019-05-31
Make PyCondorAPI compatible with HTCondor v8.8.3

khurtado:
Makes WMCore compatible with HTCondor v8.8.3 python bindings. It also works with the current v8.5.7 binding.

https://github.com/cms-sw/cmsdist/pull/4899

#### Description
After locating the schedd, query it with the proper CurbMatchmaking projection. 

Yes, works with v8.5.7 too.

#### Related PRs
https://github.com/cms-sw/cmsdist/pull/4899

### DIFF
--- a/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
+++ b/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
@@ -79,7 +79,9 @@ class PyCondorAPI(object):
         """
         try:
             scheddAd = self.coll.locate(htcondor.DaemonTypes.Schedd)
-            isOverloaded = scheddAd['CurbMatchmaking'].eval()
+            q = self.coll.query(htcondor.AdTypes.Schedd, 'Name == "%s"' % scheddAd['Name'],
+                                projection=['CurbMatchmaking'])[0]
+            isOverloaded = q['CurbMatchmaking'].eval()
             return isOverloaded
         except Exception:
             # if there is an error, try to recreate the collector instance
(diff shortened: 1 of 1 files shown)


amaltaro:
@khurtado Kenyi, how about the schedd version? Is there any requirement? Or this change relies only on the COMP HTCondor package installed with the agents?

khurtado:
@amaltaro No changes in the schedd are needed. This works with the python bindings on the COMP HTCondor package. Tested with the current  `8.5.7` bindings and the new `8.8.3`  and had no issues querying schedds working with `8.6.11` and `8.8.1`.

amaltaro:
Thanks, Kenyi.

amaltaro:
@khurtado Kenyi, I've just hit the following exception coming from this fix
```
2019-06-20 21:03:54,365:140084854126336:INFO:PyCondorAPI:Recreating Collector instance due to query error...
2019-06-20 21:04:14,442:140084854126336:ERROR:PyCondorAPI:Condor failed to fetch schedd attributes.Error message: list index out of range
  File "/data/srv/wmagent/v1.2.4.pre3-comp/sw.amaltaro/slc7_amd64_gcc630/cms/wmagent/1.2.4.pre3-comp/lib/python2.7/site-packages/WMCore/Services/PyCondor/PyCondorAPI.py", line 93, in isScheddOverloaded
    projection=['CurbMatchmaking'])[0]
IndexError: list index out of range
2019-06-20 21:04:14,443:140084854126336:WARNING:WorkQueueManagerWorkPoller:No work will be pulled: schedd is overloaded
```

is it something that would deserve a better error handling? Or you think the catch-all exception is good enough.
For the record, it happened in vocms0262 while the agent had only ~200 jobs in the condor queue.

khurtado:
@amaltaro The error message seems clear enough to me. I wonder if we should retry at least once more when we find this though, since it seems a transient query error. Like, using the @retry decorator [here](https://pypi.org/project/retrying/), what do you think?

amaltaro:
It's a minor change, but I think your suggestion would make things more clear indeed.

PR #9205, opened 2019-05-17
Add CherryPy stats to WebTools/RESTModel based server

vkuznet:
The code provides ability to expose CherryPy statistics from https://cherrypydocrework.readthedocs.io/pkg/cherrypy.lib.html#module-cherrypy.lib.cpstats

Any web service which depends on WebTools/RESTModel class (e.g. DBS) will now serve CherryPy statistics via /stats endpoint, e.g. `cmsweb.cern.ch/dbs/stats` will provide this. The statistics dict will provide various metrics about underlying cherrypy activity. Here is an example of stats dict:
```
{
    "Cheroot HTTPServer 4517118096": {
        "Accepts": 0,
        "Accepts/sec": 0.0,
        "Bind Address": "('127.0.0.1', 8080)",
        "Bytes Read": -1,
        "Bytes Written": -1,
        "Enabled": false,
        "Queue": 0,
        "Read Throughput": -1,
        "Requests": -1,
        "Run time": -1,
        "Socket Errors": 0,
        "Threads": 10,
        "Threads Idle": 9,
        "Work Time": -1,
        "Worker Threads": {
            "CP Server Thread-10": {
                "Bytes Read": 0,
                "Bytes Written": 0,
                "Read Throughput": 0.0,
                "Requests": 0,
                "Work Time": 0,
                "Write Throughput": 0.0
            },
            "CP Server Thread-11": {
                "Bytes Read": 0,
                "Bytes Written": 0,
                "Read Throughput": 0.0,
                "Requests": 0,
                "Work Time": 0,
                "Write Throughput": 0.0
            },
            "CP Server Thread-12": {
                "Bytes Read": 0,
                "Bytes Written": 0,
                "Read Throughput": 0.0,
                "Requests": 0,
                "Work Time": 0,
                "Write Throughput": 0.0
            },
            "CP Server Thread-3": {
                "Bytes Read": 0,
                "Bytes Written": 0,
                "Read Throughput": 0.0,
                "Requests": 0,
                "Work Time": 0,
                "Write Throughput": 0.0
            },
            "CP Server Thread-4": {
                "Bytes Read": 0,
                "Bytes Written": 0,
                "Read Throughput": 0.0,
                "Requests": 0,
                "Work Time": 0,
                "Write Throughput": 0.0
            },
            "CP Server Thread-5": {
                "Bytes Read": 0,
                "Bytes Written": 0,
                "Read Throughput": 0.0,
                "Requests": 0,
                "Work Time": 0,
[...]

PR #9237, opened 2019-06-03
Enable CherryPy stats for internal HTTP server

vkuznet:
Fixes #<GH_Issue_Number>

#### Status
ready

#### Description
Upon further investigation with zeroed values in CherryPy HTTP server metrics I found that an additional parameter should be turned on in order to enable them. There are two type of stats: one which is controlled by `tools.cpstats.on` flag enabled CherryPy application statistics, while another one which is controlled by `server.statistics` flag is enable internal HTTP server stats.

YES

#### Related PRs
https://github.com/dmwm/WMCore/pull/9205

no

### DIFF
--- a/src/python/WMCore/WebTools/Root.py
+++ b/src/python/WMCore/WebTools/Root.py
@@ -269,6 +269,7 @@ class Root(Harness):
                                         'tools.secmodv2.group': self.secconfig.default.group,
                                         'tools.secmodv2.site': self.secconfig.default.site})
         cherrypy.config.update({'tools.cpstats.on': configDict.get('cpstats', False)})
+        cherrypy.config.update({'server.statistics': configDict.get('cpstats', False)})
         cherrypy.log.error_log.debug('Application %s initialised in %s mode', self.app, self.mode)
         cherrypy.log.access_log.info("Final CherryPy configuration: %s" % pformat(cherrypy.config))
 


vkuznet:
@yuyiguo , Yuyi I found a reason why CherryPy HTTP server metrics are empty in dashboard. Turns out we need an additional flag to enabled in cherrypy configuration. Since it is already late for this release we'll only see application stats. Once this PR will be accepted the CherryPy HTTP server metrics will also appear.

@amaltaro , Alan, I found this only when we enabled stats in DBS and when I build and inspected dashboards. Then it took me a while to find out in cherrypy code where to turn on the server stats. Luckily, it is only configuration change and we can have it in next cmsweb upgrade.

src/python/WMCore/WebTools/Root.py
```diff
@@ -269,6 +269,7 @@ def _configureCherryPy(self):
                                         'tools.secmodv2.group': self.secconfig.default.group,
                                         'tools.secmodv2.site': self.secconfig.default.site})
         cherrypy.config.update({'tools.cpstats.on': configDict.get('cpstats', False)})
+        cherrypy.config.update({'server.statistics': configDict.get('cpstats', False)})
```
amaltaro: Do we want to always, either enable both or don't enable any at all? Otherwise we should read it from a different service parameter.

vkuznet:
From monitoring point of view it is desired to enable both since they provide
full picture of CherryPy application/server. That's why I left the same flag.

amaltaro:
Alright. For reference, here is where it gets documented:
https://cherrypydocrework.readthedocs.io/pkg/cherrypy.html#cherrypy._cpserver.Server.statistics

@yuyiguo I'm not backporting it to the DBS branch because I assume you're working on updating DBS to a newer WMCore version/HEAD. If not, then please consider it, the dbs branch is very old and it's becoming more complicated to backport fixes.

PR #9186, opened 2019-05-09
Open files/urls with context manager

amaltaro:
Let's make sure we properly close file descriptors. I skipped a few cases that were more complicated or under the test/* directory.

PR #9249, opened 2019-06-11
removing the .read from the with open() file open method.

vytjan:
Fixes #9186 

#### Status
ready

#### Description
Fixes the file open method.

YES

#### Related PRs
#9186 

None


vytjan:
@amaltaro Alan, should this be merged or would you like to have some workaround here?

amaltaro:
Good catch, Vytas!

PR #9275, opened 2019-06-26
Add two new APIs: getTrainsferIDs/postTransferIDs to deal with transfer IDs

### LINKED ISSUE #9198: Create a new REST API for the MS Transferor to store/retrieve transfer IDs
As part of the MS Transferor design described in:

we also need to persist information related to workflow and the data placement subscription ids, which will be used by the MS Transferor and TransferMonit applications.

In terms of implementation details, it would be another class/document type under the reqmgr_aux database
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/Auxiliary.py#L246

vkuznet:
Fixes #9198 

#### Status
not-tested

#### Description
Add two new APIs to deal with transfer IDs

We need a new transferids database in CouchDB to store this data.

YES

#### Related PRs
#9198 

no

### DIFF
--- a/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
+++ b/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
@@ -229,6 +229,29 @@ class ReqMgrAux(Service):
         else:
             return self["requests"].delete('%sconfig/%s' % (docType, docName))[0]['result']
 
+    def getTransferInfo(self):
+        """
+        Get transfer document from coudhdb backend. The document has the following form:
+        {
+          "wf_A": [record1, record2, ...],
+          "wf_B": [....],
+        }
+        where each record has the following format:
+        {"timestamp":000, "dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2,3]}
+        transferIDs here either represent PhEDEx subscription or Rucio transfer/rules IDs.
+        """
+        return self._getDataFromMemoryCache('transferinfo')
+
+    def postTransferInfo(self, transferInfo):
+        """
+        Post new transfer document into couchdb backend.
+        :param transferInfo: represent a transferInfo document
+        {
+          "wf_A": [record1, record2, ...],
+          "wf_B": [....],
+        }
+        """
+        return self["requests"].post('transferinfo', transferInfo)[0]['result']
 
 AUXDB_AGENT_CONFIG_CACHE = {}
 


vkuznet:
@amaltaro , this PR introduce new APIs to fetch and store transfer IDs doc. For that we need to create a new DB in couch. I called **transferids**. Please review and make comments. Once this APIs will be in place I'll move forward with using them in MS (see #9276 )

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -229,6 +229,72 @@ def deleteConfigDoc(self, docType, docName):
...
 
+    def getTransferInfo(self):
+        """
+        Get transfer document from coudhdb backend. The document has the following form:
+        {
+          "wf_A": [record1, record2, ...],
+          "wf_B": [....],
+        }
+        where each record has the following format:
+        {"timestamp":000, "dataset":"/a/b/c", "campaignType": "primary", "subscription": subscription}
```
amaltaro: replace `campaignType` by `dataType` and `subscription` by `transferIDs`.
As discussed and agreed here: https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154#file-transferor_format-json-L24

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -229,6 +229,72 @@ def deleteConfigDoc(self, docType, docName):
...
+        subscription here either represent PhEDEx or Rucio subscriptions.
```
amaltaro: `subscription here represents either a PhEDEx subscription ID or a Rucio rule ID`

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -229,6 +229,72 @@ def deleteConfigDoc(self, docType, docName):
...
+        Phedex subscription record:
```
amaltaro: this block with this struct is not needed here.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -229,6 +229,72 @@ def deleteConfigDoc(self, docType, docName):
...
+            "node_bytes": 1542152522,
+            "level": "DATASET",
+            "custodial": "n",
+            "suspended": "n",
+            "percent_files": 3.57142857142857
+          }
+        """
+        return self._getDataFromMemoryCache('transferinfo')
+
+    def postTransferInfo(self, transferInfo):
```
amaltaro: Similar comments as above.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
```diff
@@ -229,6 +229,72 @@ def deleteConfigDoc(self, docType, docName):
...
+            "node_id": 50,
+            "request": 1822579,
+            "node_bytes": 1542152522,
+            "level": "DATASET",
+            "custodial": "n",
+            "suspended": "n",
+            "percent_files": 3.57142857142857
+          }
+        """
+        return self["requests"].post('transferinfo', transferIDs)[0]['result']
```
amaltaro: This API needs to be defined/renamed. We'll get back to it...

amaltaro (changes requested):
It depends on having the proper REST APIs, even though I've made a few comments/requests along the code already.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
vkuznet: Alan, I already had transferIDs, but then used subscription since I was not sure which PhEDEx/Rucio APIs we're going to use. I suggest before making changes here you review and outline explicitly which APIs which should use, either
```
phedex/datasvc/json/prod/requestlist?dataset=/a/b/c
```
which returns ids or
```
phedex/datasvc/json/prod/subscriptions?dataset=/a/b/c
```
which returns subscription details. I think once we will know which APIs to use we'll adjust ReqMgrAux `getTransferInfo` to use appropriate name for the record.

src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py
vkuznet: again please provide APIs which we should use and then we can adjust the docstring.

vkuznet:
Alan, please post which APIs should be used. We need not only APIs how to fetch IDs/subscription but also APIs where we can get info about transfers. The `requestinfo` Phedex API returns IDs, but it is unclear how we can use those IDs, while `subscriptions` API returns details of subscription and its records contains percentage info which we need. For the reference, the Unified code use the later (`subscriptions` API).

vkuznet:
@amaltaro , Alan please review and merge this PR

amaltaro:
This code would need a new `transferinfo` REST API to be created.

Besides that, I'd like to discuss (maybe on the Monday chat) how we'll handle those transfer-related data. My initial idea was the simplest possible, which is:
1. we store transfer-related data in a single document in CouchDB, making it much easier to manage and GET and PUT any new data in.
2. however, for that to work, we'd have to make updates in bulk. E.g., the transferor thread would handle X requests (100?), once transfers are all done, then it would update this document. The **problem** on this approach is that if the thread/node crashes in the request number 99, then we lose everything that has been done up to that point, and the thread needs to run the same thing again in the next cycle.
3. same applies to the monitor thread, it would have to update in-memory the status of X requests/transfers and once all of those were updated, then we trigger a CouchDB document update to persist that information.

In short, besides the eventual document conflicts that we'd hit (and we need to handle that problem), we could as well lose part of the work done by the thread, having to wait until the next cycle to try again.

SO, my second suggestion would be to create one transfer document for each workflow in the system, such that having problems with one request has no effect at all on the overall thread execution.
The drawback on this case is that we need to have additional micro management of the documents, it has more network and REST API calls and it's likely going to be slower.

amaltaro:
Superseeded by https://github.com/dmwm/WMCore/pull/9307

PR #9276, opened 2019-06-26
Changes for transfer IDs logic

### LINKED ISSUE #9196: Create a new MS TransferMonit application
As discussed in this second proposal for the MS Transferor architecture:

we have the need for another application that can monitor data placement subscriptions and advance requests to the next status.

In short, this MS would be in charge of:
* fetching workflows from ReqMgr2 that are in the `staging` status
* fetch the map of workflow / dataset / list of transfer IDs from Couchdb
  * respecting workflows with a timestamp greater than X hours (6h?)
* IF the workflow has no input data at all, simply update its status to `staged`, otherwise, continue the bullets below...
* fetch the unified configuration (or the campaign configuration, whatever we need to use in order to read the Pileup transfer acceptance)
* call PhEDEx/Rucio subscriptions
  * IF all calls for a given workflow succeeded, update its timestamp in the couch map document
  * ELSE, don't update anything and just retry in the next cycle (or update LogDB with the call that failed and the reason...)
* IF pileup is above the completion acceptance in the campaign/unified configuration, then update the request status to `staged`

vkuznet:
Partially fixes #9196 

#### Status
In development

#### Description
It implements logic for MS based on transfer IDs. Here is how it is done:
  - code refactoring
     - `MSCore` is a base class for `MSMonitor`, `MSTransferor` and `RequestInfo`
  - transferor API obtain request info and perform transferRequest
    the transferRequest performs phedex subscription (via subscribe API which
    so far is commented out) and obtain transfer ids via getTransferIds API
    It does this for every dataset in request spec. At the end transfer API
    received status (transfer) records which its store to couchdb via
    `updateTransferInfo` API (which calls apis from 9275)
  - monit API get requests and fetch status (transfer) records from couchdb
    via `getTransferRecords` API (which by itself calls getTransferIds one).
    Then we check completion of transfer via `completion` API. If completed value is
    100 (percent) we are done, otherwise we update new set of records
    in couchdb via updateTransferInfo API.

YES

#### Related PRs
depends on #9198 , #9275, #9307 

no

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSCore.py
+++ b/src/python/WMCore/MicroService/Unified/MSCore.py
@@ -1,0 +1,136 @@
+"""
+File       : MSCore.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
+Description: MSCore class provides core functionality of the MS.
+"""
+# futures
+from __future__ import division, print_function
+
+# system modules
+import time
+
+# WMCore modules
+from WMCore.MicroService.Unified.Common import getMSLogger
+from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
+from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
+from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
+
+
+class MSCore(object):
+    """
+    This class provides core functionality for both
+    MSTransferor and MSMonitor classes.
+    """
+
+    def __init__(self, msConfig, logger=None):
+        """
+        Provides setup for MSTransferor and MSMonitor classes
+
+        :param config: MS service configuration
+        :param logger: logger object (optional)
+        """
+        self.logger = getMSLogger(getattr(msConfig, 'verbose', False), logger)
+        self.msConfig = msConfig
+        self.logger.info(
+            "Configuration including default values:\n%s", self.msConfig)
+
+        self.reqmgr2 = ReqMgr(self.msConfig['reqmgrUrl'], logger=self.logger)
+        self.reqmgrAux = ReqMgrAux(self.msConfig['reqmgrUrl'],
+                                   httpDict={'cacheduration': 60},
+                                   logger=self.logger)
+
+        # eventually will change it to Rucio
+        self.phedex = PhEDEx(httpDict={'cacheduration': 10 * 60},
+                             dbsUrl=self.msConfig['dbsUrl'], logger=self.logger)
+
+    def unifiedConfig(self):
+        """
+        Fetches the unified configuration
+        :return: unified configuration content
+        """
+        res = self.reqmgrAux.getUnifiedConfig(docName="config")
+        if res:
+            if isinstance(res, list):
+                return res[0]
+            return res
+        else:
+            return {}
+
+    def change(self, req, reqStatus, prefix='###'):
+        """
+        Change request status, internally it is done via PUT request to ReqMgr2:
+        curl -X PUT -H "Content-Type: application/json" \
+             -d '{"RequestStatus":"staging", "RequestName":"bla-bla"}' \
+             https://xxx.yyy.zz/reqmgr2/data/request
+        """
+        self.logger.debug(
+            '%s updating %s status to %s', prefix, req['name'], reqStatus)
+        try:
+            if not self.msConfig['readOnly']:
+                self.reqmgr2.updateRequestStatus(req['name'], reqStatus)
+        except Exception as err:
+            self.logger.exception(
+                "Failed to change request status. Error: %s", str(err))
+
+    def updateTransferInfo(self, requestStatuses):
+        "Update transfer ids in backend"
+        tstamp = time.time()
+        # Alan's suggestion to use bulk query
+        # get all docs from CouchDB
+        docs = self.getTransferInfo('ALL_DOCS')
+        for wname, statusRecord in requestStatuses:
+            # obtain records from CouchDB
+            # doc = self.getTransferInfo(wname)
+            # find record in our list of docs
+            # VK: I don't know if it will be faster then fetching the record
+            #     in CouchDB, we need to loop every time to find proper record
+            doc = {}
+            for doc in docs:
+                if wname in doc:
+                    break
+            records = doc.get('transfers', [])
+            records += statusRecord
+            doc['workflowName'] = wname
+            doc['timestamp'] = tstamp
+            doc['transfers'] = records
+            self.reqmgrAux.postTransferInfo(wname, doc)
+
+    def getTransferInfo(self, wname):
+        """
+        Get transfer document from backend. The document has the following form:
+
+        .. doctest::
+
+            {"workflowName": "bla",
+             "timestamp": 123,
+             "transfers": [rec1, rec2, ... ]
+            }
+            where each record has the following format:
+            {"dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2], "completion": 0}
+
+
+        :param wname: workflow name
+        :return: a tranfer ids document
+        """
+        return self.reqmgrAux.getTransferInfo(wname)
+
+    def getTransferIds(self, dataset):
+        """
+        Get transfer ids document for given request name and datasets.
+        :param dataset: dataset name
+        :return: a list of transfer ids
+        """
+        # phedex implementation, TODO: implement Rucio logic when it is ready
+        data = self.phedex.subscriptions(
+            dataset=dataset, group=self.msConfig['group'])
+        self.logger.debug(
+            "### dataset %s group %s", dataset, self.msConfig['group'])
+        self.logger.debug("### subscription %s", data)
+        tids = []
+        vals = []
+        for row in data['phedex']['dataset']:
+            if row['name'] == dataset:
+                for rec in row['subscription']:
+                    vals.append(int(rec['percent_files']))
+                    tids.append(int(rec['request']))
+        return tids, float(sum(vals))/len(vals)
--- a/src/python/WMCore/MicroService/Unified/MSManager.py
+++ b/src/python/WMCore/MicroService/Unified/MSManager.py
@@ -1,11 +1,26 @@
 """
-File       : UnifiedTransferorManager.py
+File       : MSManager.py
 Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
-Description: UnifiedTransferorManager class provides full functionality of the UnifiedTransferor service.
-"""
+             Alan Malta <alan dot malta AT cern dot ch >
+Description: MSManager class provides full functionality of the MSManager service.
+It provides a interface to reqmgr2ms service and should be
+used in service config.py as following
 
+.. doctest::
+
+    # REST interface
+    data = views.section_('data')
+    data.object = 'WMCore.MicroService.Service.RestApiHub.RestApiHub'
+    data.manager = 'WMCore.MicroService.Unified.MSManager.MSManager'
+    data.reqmgr2Url = "%s/reqmgr2" % BASE_URL
+    data.readOnly = False
+    data.verbose = True
+    data.interval = 60
+    data.rucioAccount = RUCIO_ACCT
+    data.dbsUrl = "%s/dbs/%s/global/DBSReader" % (BASE_URL, DBS_INS)
+"""
 # futures
-from __future__ import division
+from __future__ import division, print_function
 
 # system modules
 import time
@@ -13,9 +28,8 @@ import time
 # WMCore modules
 from WMCore.MicroService.Unified.Common import getMSLogger
 from WMCore.MicroService.Unified.MSTransferor import MSTransferor
+from WMCore.MicroService.Unified.MSMonitor import MSMonitor
 from WMCore.MicroService.Unified.TaskManager import start_new_thread
-from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
-from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
 
 
 def daemon(func, reqStatus, interval, logger):
@@ -31,42 +45,48 @@ def daemon(func, reqStatus, interval, logger):
 class MSManager(object):
     """
     Entry point for the MicroServices.
-    This class manages both transferor and monitor services/threads.
+    This class manages both transferor and monitoring services.
     """
 
     def __init__(self, config=None, logger=None):
         """
-        Setup a bunch of things, like:
-         * logger for this service
-         * initialize all the necessary service helpers
-         * fetch the unified configuration from central couch
-         * update the unified configuration with some deployment and default settings
-         * start both transfer and monitor threads
+        Initialize MSManager class with given configuation,
+        logger, ReqMgr2/ReqMgrAux/PhEDEx/Rucio objects,
+        and start transferor and monitoring threads.
         :param config: reqmgr2ms service configuration
         :param logger:
         """
         self.uConfig = {}
         self.config = config
         self.logger = getMSLogger(getattr(config, 'verbose', False), logger)
+        self._msConfig = None  # will be defined by _parseConfig API call
         self._parseConfig(config)
-        self.logger.info("Configuration including default values:\n%s", self.msConfig)
+        self.logger.info(
+            "Configuration including default values:\n%s", self.msConfig)
 
-        self.reqmgr2 = ReqMgr(self.msConfig['reqmgrUrl'], logger=self.logger)
-        self.reqmgrAux = ReqMgrAux(self.msConfig['reqmgrUrl'], httpDict={'cacheduration': 60}, logger=self.logger)
+        # initialize transferor class with assigned status as default
+        self.msTransferor = MSTransferor(self.msConfig, logger=self.logger)
+        # initialize monitoring class with staging status as default
+        self.msMonitor = MSMonitor(self.msConfig, logger=self.logger)
 
-        # transferor has to look at workflows in assigned status
-        self.msTransferor = MSTransferor(self.msConfig, "assigned", logger=self.logger)
-
-        ### Last but not least, get the threads started
+        # Last but not least, get the threads started
         thname = 'MSTransferor'
         self.transfThread = start_new_thread(thname, daemon,
-                                             (self.transferor, 'assigned', self.msConfig['interval'], self.logger))
-        self.logger.debug("### Running %s thread %s", thname, self.transfThread.running())
+                                             (self.transferor,
+                                              'assigned',
+                                              self.msConfig['interval'],
+                                              self.logger))
+        self.logger.debug(
+            "### Running %s thread %s", thname, self.transfThread.running())
 
         thname = 'MSTransferorMonit'
         self.monitThread = start_new_thread(thname, daemon,
-                                            (self.monitor, 'staging', self.msConfig['interval'] * 2, self.logger))
-        self.logger.debug("+++ Running %s thread %s", thname, self.monitThread.running())
+                                            (self.monitor,
+                                             'staging',
+                                             self.msConfig['interval'] * 2,
+                                             self.logger))
+        self.logger.debug(
+            "+++ Running %s thread %s", thname, self.monitThread.running())
 
     def _parseConfig(self, config):
         """
@@ -82,10 +102,15 @@ class MSManager(object):
         self.msConfig['interval'] = getattr(config, 'interval', 5 * 60)
         self.msConfig['readOnly'] = getattr(config, 'readOnly', True)
 
-        self.msConfig['reqmgrUrl'] = getattr(config, 'reqmgr2Url', 'https://cmsweb.cern.ch/reqmgr2')
-        self.msConfig['reqmgrCacheUrl'] = self.msConfig['reqmgrUrl'].replace('reqmgr2', 'couchdb/reqmgr_workload_cache')
-        self.msConfig['phedexUrl'] = getattr(config, 'phedexUrl', 'https://cmsweb.cern.ch/phedex/datasvc/json/prod')
-        self.msConfig['dbsUrl'] = getattr(config, 'dbsUrl', 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader')
+        reqmgr2Url = 'https://cmsweb.cern.ch/reqmgr2'
+        self.msConfig['reqmgrUrl'] = getattr(config, 'reqmgr2Url', reqmgr2Url)
+        self.msConfig['reqmgrCacheUrl'] = \
+            self.msConfig['reqmgrUrl'].replace(
+                'reqmgr2', 'couchdb/reqmgr_workload_cache')
+        phedexUrl = 'https://cmsweb.cern.ch/phedex/datasvc/json/prod'
+        self.msConfig['phedexUrl'] = getattr(config, 'phedexUrl', phedexUrl)
+        dbsUrl = 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader'
+        self.msConfig['dbsUrl'] = getattr(config, 'dbsUrl', dbsUrl)
 
     def transferor(self, reqStatus):
         """
@@ -97,10 +122,11 @@ class MSManager(object):
         """
         startT = time.time()
         self.logger.info("Starting the transferor thread...")
-        self.msTransferor.execute()
-        self.logger.info("Total transferor execution time: %.2f secs", time.time() - startT)
+        self.msTransferor.execute(reqStatus)
+        self.logger.info(
+            "Total transferor execution time: %.2f secs", time.time() - startT)
 
-    def monitor(self, reqStatus='staging'):
+    def monitor(self, reqStatus):
         """
         MSManager monitoring function.
         It performs transfer requests from staging to staged state of ReqMgr2.
@@ -109,50 +135,9 @@ class MSManager(object):
         """
         startT = time.time()
         self.logger.info("Starting the monitor thread...")
-        # First, fetch/update our unified configuration from reqmgr_aux db
-        # Keep our own copy of the unified config to avoid race conditions
-        self.uConfig = self.reqmgrAux.getUnifiedConfig(docName="config")
-        if not self.uConfig:
-            self.logger.warning("Monitor failed to fetch the unified config. Skipping this cycle.")
-            return
-        self.uConfig = self.uConfig[0]
-
-        try:
-            # get requests from ReqMgr2 data-service for given statue
-            # here with detail=False we get back list of records
-            requests = self.reqmgr2.getRequestByStatus([reqStatus], detail=False)
-            self.logger.debug('+++ monit found %s requests in %s state', len(requests), reqStatus)
-
-            requestStatus = {}  # keep track of request statuses
-            for reqName in requests:
-                req = {'name': reqName, 'reqStatus': reqStatus}
-                # get transfer IDs
-                tids = self.getTransferIDs()
-                # get transfer status
-                transferStatuses = self.getTransferStatuses(tids)
-                # get campaing and unified configuration
-                campaign = self.requestCampaign(reqName)
-                conf = self.requestConfiguration(reqName)
-                self.logger.debug("+++ request %s campaing %s conf %s", req, campaign, conf)
-
-                # if all transfers are completed, move the request status staging -> staged
-                # completed = self.checkSubscription(request)
-                completed = 100  # TMP
-                if completed == 100:  # all data are staged
-                    self.logger.debug("+++ request %s all transfers are completed", req)
-                    self.change(req, 'staged', '+++ monit')
-                # if pileup transfers are completed AND some input blocks are completed, move the request status staging -> staged
-                elif self.pileupTransfersCompleted(tids):
-                    self.logger.debug("+++ request %s pileup transfers are completed", req)
-                    self.change(req, 'staged', '+++ monit')
-                # transfers not completed, just update the database with their completion
-                else:
-                    self.logger.debug("+++ request %s transfers are not completed", req)
-                    requestStatus[req] = transferStatuses  # TODO: implement update of transfer ids
-            self.updateTransferIDs(requestStatus)
-        except Exception as err:  # general error
-            self.logger.exception('+++ monit error: %s', str(err))
-        self.logger.info("Total monitor execution time: %.2f secs", time.time() - startT)
+        self.msMonitor.execute(reqStatus)
+        self.logger.info(
+            "Total monitor execution time: %.2f secs", time.time() - startT)
 
     def stop(self):
         "Stop MSManager"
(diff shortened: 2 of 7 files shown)


amaltaro:
As mentioned in the meeting, please create an issue reporting the problem and what's the expected behavior.

vkuznet:
Alan, this PR already listed dependencies:  #9198 , #9275. The only missing bit is you need to decide which DB name to use and which structure (transfer IDs or suscription) we'll deal with. Once settle and we both issues will be merged I can move with this.

amaltaro:
Valentin, I'm having a hard time reviewing this PR. Can you please describe in the first comment box what exactly are these changes? What is the flow of actions executed by MS here? and what has to be discussed/defined/fix ahead.

vkuznet:
let's adjust it once you'll clarify which Phedex APIs to use.

src/python/WMCore/MicroService/Unified/RequestInfo.py
```diff
@@ -461,8 +461,10 @@ def unified(svc, requestRecords, logger):
             ncopies = getNCopies(cput)
 
             attrs = winfo[wname]
-            ndatasets = len(attrs['datasets'])
-            npileups = len(attrs['pileups'])
+            campaign = attrs['campaign']
+            campaignType = 'primary' # TODO: how to define it?
```
amaltaro: That's a good question. I believe it's defined in this method:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/MSTransferor.py#L115

when it's looking for all possible data that we need to transfer (it misses the parent though). So we should not hardcode it but actually see all possible values and get it.

src/python/WMCore/MicroService/Unified/Transferor.py
```diff
@@ -143,79 +167,147 @@ def stop(self):
...
+                for rec in row['subscription']:
+                    # obtain only unfinished transfer ids
+                    if rec['percent_files'] != 100:
+                        tids.append(int(rec['request']))
+        return list(set(tids))
 
-    def requestCampaign(self, req):
-        "Return request campaign"
-        return 'campaign_TODO' # TODO
+    def getStatusRecords(self, reqName):
```
amaltaro: Hmm, we don't really persist a status of the transfers. It will be calculated on the fly based on the transfer document. OR, we need to add another key to the trasnfer document schema to keep track of the completion status of those transfers.

Then we still have to define: "completion", but that's in another time.

amaltaro (changes requested):
Valentin, I've done just an initial review because this PR needs to be rebased (I pushed in a bunch of changes on the last Friday).

Nonetheless, let me list the steps we should take in the monitoring thread, in addition to what's described in the point 6. of this wiki:

The step by step would be:
1. fetch all the workflows in the `staging` status
2. fetch all the workflow transfer documents
2.a. filter those documents out such that we don't check for their completion status every single cycle. So we could do something like `if timestamp - time() > X hours: then check this one`. I'd say every 4h or so.
3. fetch the campaign configuration
4. using the transfer ids, make calls to phedex and get the transfer status
5. calculate the status completion
5.a. if ALL transfer type (primary, secondary, etc) are above the minimum thresholds, then a) update the transfer doc and b) update the request status to `staged`
5.b. if NOT ALL transfers are above the minimum thresholds, then update the transfer doc and proceed to the next workflow
5.c. IF there are no transfers under the workflow transfer doc (it means there is not any type of input data), then move the workflow to `staged`

I'll update the wiki page soon. Please let me know if you have any questions.

vkuznet:
ok, then I'll rebase once you'll merge https://github.com/dmwm/WMCore/pull/9307 and continue working on this

amaltaro:
Done, feel free to proceed with this one. I'm going to ask a new testbed release tomorrow morning, so we might have to postpone it until the next week (if safe changes).

vkuznet:
Alan, do you mind to explain why did you re-factor MS code and what exactly you did.
I see SO MANY changes that I want to understand the entire logic before proceeding. For instance, to preserve previous work I need clear understanding which branch to use to resolve the conflicts, the head or the one I had before. I need to understand which methods/functions to keep, etc.

And, for the future, please elaborate and discuss the changes up-front. I already rewrote this piece so many times that I start having a feeling that I'm wasting my time on this codebase.

amaltaro:
Because the transferor code was looking an all-in-one, where we were piling up methods and more methods without really thinking where things belong to, without reusing code.

I have all that explained in the first comment of this PR:
https://github.com/dmwm/WMCore/pull/9296

where you were tagged. I also asked you to have a closer look once you get back from vacation (it was merged in the past week). Let me know if you still have questions, but we better then follow up on that PR to keep the discussion in the same context.

amaltaro:
I have started the basic description/behaviour of the Monitor thread here:
https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor

but as you know, we might have to touch it once we fully understand the "black box" logic of the transferor service and all the parameters it depends on. Which made me create this option D here:
https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154#file-transferor_format-json-L32

and add an extra field `campaignName`, such that we don;t need to look into a request detail once again in the monitor thread.

vkuznet:
I performed rebase and added/fixed the following:
- I re-factored all mine and Alan's changes to separate MS logic into the following classes:
  - MSCore.py represents core module for MSTransferor, MSMonitor, RequestInfo classes
  - I separated transferor, monitoring and request info into separate classes
  - I cleaned and implemented consistent logic in MSManager class
- I refactored Alan's changes to make code thread safe (especially introduced in RequestInfo and MSManager classes)
- I implemented logic based on option D from transfer records document (see https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154)

I did not (yet) made any effort to complete full logic of transferor/monitoring MS services and suggest that it should be handled in next PR once we agree on these changes and merge them. I also didn't touch yet unified logic and suggest that it should be done after full MS logic will be in place.

amaltaro:
@vkuznet you probably have a broken import here, that's why this avalanche of unit tests failing.

In addition to that, please do not disable pylint checks blindly, otherwise we never know what needs attention in the future or not.

vkuznet:
Alan, I doubt it is broken import since pylint/pycodestyle will capture this. I also checked manually imports and they work fine. Plus I have doubt that such unit test:
```
WMCore_t.Services_t.DBS_t.DBSReader_t.DBSReaderTest:testListDatatiers changed from success to error
```
will be related to MS code. I think failure of unit tests is related by something else. Do you have any log to look what is exactly fails?

Regarding pylint checks, I don't exclude blindly. I excluded those which I believe irrelevant. For instance, here is a typical codes:

```
pylint --help-msg=C0103,R0914,R0912,R1702,R0201,W0703
No config file found, using default configuration
:invalid-name (C0103): *%s name "%s" doesn't conform to %s*
  Used when the name doesn't conform to naming rules associated to its type
  (constant, variable, class...). This message belongs to the basic checker.

:too-many-locals (R0914): *Too many local variables (%s/%s)*
  Used when a function or method has too many local variables. This message
  belongs to the design checker.

:too-many-branches (R0912): *Too many branches (%s/%s)*
  Used when a function or method has too many branches, making it hard to
  follow. This message belongs to the design checker.

:too-many-nested-blocks (R1702): *Too many nested blocks (%s/%s)*
  Used when a function or a method has too many nested blocks. This makes the
  code less understandable and maintainable. This message belongs to the
  refactoring checker.

:no-self-use (R0201): *Method could be a function*
  Used when a method doesn't use its bound instance, and so could be written as
  a function. This message belongs to the classes checker.

:broad-except (W0703): *Catching too general exception %s*
  Used when an except catches a too general exception, possibly burying
  unrelated errors. This message belongs to the exceptions checker.
```
If you want I can remove all pylints, but it is trivial to check, as shown, what codes means.

amaltaro:
Besides C0103, I'd say all the others have a more meaningful impact on the code and can eventually be made better (even C0103 could, after all most of it is new code...). If you disable them, we never get it reported (unless we remember to check the pylints that were disabled).
I'm just saying we don't need to have a perfect score, but if possible not to make it worse...

About the error, if you follow the `default` `Details` link, it brings you to a summary of all the unit tests failing, e.g.:
https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/9231/testReport/junit/WMCore_t.WorkQueue_t.Policy_t.Start_t.Block_t/BlockTestCase/testLumiMask/

so the problem is not an import indeed, but the PhEDex module being mocked at the wrong place (you reshuffled things, so we don't need to mock PhEDEx there anymore). Ping me on slack if you need help walking through those reports.

vkuznet:
Alan, I removed pylint exceptions, I fixed unit test. Please review since I don't to waste time fixing either pylint or irrelevant unit tests (Rucio).

amaltaro:
Has it been tested? I need to request another WMCore tag tomorrow early in the morning and probably won't manage to have a thorough review and test.

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -0,0 +1,52 @@
...
+from WMCore.Services.PhEDEx.PhEDEx import PhEDEx
+
+
+class MSCore(object):
+    """
+    This class provides core functionality for both
+    MSTransferor and MSMonitor classes.
+    """
+
+    def __init__(self, msConfig=None, logger=None):
```
amaltaro: make `msConfig` mandatory and remove the check below

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        Provides setup for MSTransferor and MSMonitor classes
+
+        :param config: MS service configuration
+        :param logger:
```
amaltaro: either add a description or remove this line :)

src/python/WMCore/MicroService/Unified/MSManager.py
```diff
@@ -1,21 +1,34 @@
 """
-File       : UnifiedTransferorManager.py
+File       : MSManager.py
 Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
-Description: UnifiedTransferorManager class provides full functionality of the UnifiedTransferor service.
+Description: MSManager class provides full functionality of the MSManager service.
+It provides a REST interface to reqmgr2ms service and should be
```
amaltaro: The REST interface is actually provided under the Services package, no?

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,83 @@
-#!/usr/bin/env python
 """
-_MSTransferor_
-
-Class to hold the whole logic behind the transferor module
+File       : MSTransferor.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
```
amaltaro: How come you put your name on the module I created? :)

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,83 @@
...
-        return bool(self.uConfig)
+        super(MSTransferor, self).__init__(msConfig, logger)
+        self.reqInfo = RequestInfo(msConfig, logger)
 
-    def execute(self):
+    def execute(self, reqStatus):
         """
         Executes the whole transferor logic
-        :param reqStatus: request status to that matters for this module
+        :param reqStatus: request statue to process
```
amaltaro: typo

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,83 @@
...
-                        # self.change(req, 'staged', '### transferor')
-        except Exception as err:  # general error
-            self.logger.exception('### transferor error: %s', str(err))
+            # compose status record
+            wname = reqName  # in our case workflow name is identical to request name
+            ctype = req['dataType']
+            for dataset, tids in sdict.items():
+                statusRecord = {'timestamp': time.time(), 'dataset': dataset,
+                                'dataType': ctype, 'transferIDs': tids}
+                requestStatuses.setdefaul(wname, []).append(statusRecord)
```
amaltaro: typo

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,83 @@
...
+            if requestSpecs:
+                for _, wfData in requestSpecs[0].items():
                     requestRecords.append(self.requestRecord(wfData))
-        except Exception as err:
+            self.logger.debug(
+                '### transferor found %s requests in %s state',
+                len(requestRecords), reqStatus)
+            # get complete requests information
+            # based on Unified Transferor logic
+            requestRecords = self.reqInfo(requestRecords)
```
amaltaro: You have just reverted the slicing changes I made. 
This code can potentially blow up memory usage on the node, since it collects tons of information for each request and only in the very end it would perform the data transfers. Please put the slicing logic back here.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -126,28 +107,24 @@ def requestRecord(self, wfData):
...
         if datasets and sites:
-            self.logger.debug("### creating subscription for: %s", pformat(req))
-            subscription = PhEDExSubscription(datasets, sites, self.msConfig['group'])
-            # TODO: implement how to get transfer id
-            tid = hashlib.md5(str(subscription)).hexdigest()
+            self.logger.debug(
+                "### creating subscription for: %s", pformat(req))
+            subscription = PhEDExSubscription(
+                datasets, sites, self.msConfig['group'])
+            self.logger.info(
```
amaltaro: Breaking all these lines make the code harder to read IMO. In general, up to 80 chars in the same line is fine, if it goes beyond that, we might either break into multiple lines or not, it depends on how readable it is.

amaltaro (changes requested):
I've performed a spot review and the code looks much better now!
There are a few issues that need to be solved though.

src/python/WMCore/MicroService/Unified/MSCore.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSCore.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSManager.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: hmm, I thought I created MS codebase, you just mess it up :) But since you asked I put your name and will blame you for not working code from now on.

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: Alan, I was trying to merge broken branch and clean-up the conflicts. Please explicitly point out which slicing you're talking about and we can put it back. Better, I suggest to make another PR once this one will be merged since we first need to get proper codebase without conflicts.

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: please be concrete what you want, the code goes above 80 characters, that's why I break it according to pylint suggestion. So either comply with pylint or not but don't ask for both :)

vkuznet:
Alan, I changed code according to your suggestions, but I didn't put slicing. Please provide a pointer and I can added it back. But once again, don't try to get ideal code until we understand its flow. Don't over optimize it until it is working, etc. I DO UNDERSTAND issue with memory consumption but we're still battling on codebase flow.

For the record code is not tested since I'm awaiting your ok with its logic.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,84 @@
...
+                # Once all transfer requests were successfully made, update: assigned -> staging
+                self.logger.debug("### transfer request for %s successfull", reqName)
+                self.change(req, 'staging', '### transferor')
+            # if there is nothing to be transferred (no input at all),
+            # then update the request status once again staging -> staged
+            # self.change(req, 'staged', '### transferor')
 
-        try:
-            reqInfo = RequestInfo(self.msConfig, self.uConfig, self.logger)
-            for reqSlice in grouper(requestRecords, 50):
```
amaltaro: This function is the one doing the slicing `grouper`.

amaltaro (changes requested):
I put a comment in the code pointing you to the slicing logic. I'd suggest to have it back in this PR to avoid having many PR adding/removing the same code, anyways this will have to be properly tested, so we better add those 10 line changes and test it altogether.

vkuznet:
Alan, I put back the slicer, but I really doubt it is proper solution. Instead, we should convert RequestInfo into generator, then it will not consume memory and neither being prompt to amount of records it process. The loop will be one pass across passed records where it will enhance them (if necessary) and yield back new set of records. I can do that code-refactoring later.

So far I suggest to finish this code refactoring and put it in place that we can work on top of it.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,196 @@
...
+            where each record has the following format:
+            {"dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2]}
+
+
+        :param wname: workflow name
+        :return: a tranfer ids document
+        """
+        return self.reqmgrAux.getTransferInfo(wname)
+
+    def getTransferIds(self, dataset):
```
amaltaro: This should be called `getTransferStatus` or something like that. We'll provide a transfer/request ID, and get the transfer status back (method input is wrong as well).

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,196 @@
...
+        # get existing transfer IDs record
+        doc = self.getTransferInfo(wname)
+        # loop over all records and obtain new transfer IDs
+        for rec in doc['transfers']:
+            dataset = rec['dataset']
+            dtype = rec['dataType']
+            # obtain new transfer ids for given request name and dataset
+            tids = self.getTransferIds(dataset)
+            rec = {'dataset': dataset, 'dataType': dtype, 'transferIDs': tids}
+            statusRecords += rec
```
amaltaro: this won't work with list + dict

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,196 @@
...
+            dataset = rec['dataset']
+            ctype = rec['dataType']
+            if ctype == 'pileup':
+                tids = self.getTransferIds(dataset)
+                # if we have any transfer IDs then we still have transfers
+                if tids:
+                    return False
+        return True
+
+    def checkStatusRecords(self, records):
```
amaltaro: I'm still not sure I get the usefulness of this method.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,196 @@
...
+            for reqName in requests:
+                req = {'name': reqName, 'reqStatus': reqStatus}
+                self.logger.debug("+++ request %s", req)
+                # obtain status records from couchdb for given request
+                statusRecords = self.getStatusRecords(reqName)
+                # check status records and obtain completion status
+                statusRecords, completed = self.checkStatusRecords(statusRecords)
+                # if all transfers are completed
+                # move the request status staging -> staged
+                if completed == 100:
```
amaltaro: This is 100 right now, but soon we'll have to compare against the numbers defined in the campaign configuration (and perhaps the unified config too, I don't know yet).

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,196 @@
...
+                    self.logger.debug(
+                        "+++ request %s all transfers are completed", req)
+                    self.change(req, 'staged', '+++ monit')
+                # if pileup transfers are completed AND
+                # some input blocks are completed,
+                # move the request status staging -> staged
+                elif self.pileupTransfersCompleted(reqName):
```
amaltaro: This check has likely to be done together with the input data.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,196 @@
...
+            self.logger.exception('+++ monit error: %s', str(err))
+
+    def getStatusRecords(self, reqName):
+        """
+        Get transfer statuses for given request name
+        transfer docs on backend has the following form
+        https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154
+        So far logic is based on option D of the records
+
+        .. doctest::
```
amaltaro: what 's this `.. doctest::` for?

amaltaro (changes requested):
Valentin, I made further comments in the code and in general it looks Okay. I don't think we need to get a perfect logic for the MSMonitoring here because it's not really the purpose of this PR AFAICS.

Besides those comments, I think we should fetch the transfer records in bulk such that we can avoid hundreds/thousands of calls, so in the MSMonitor, you fetch all the requests in a given status, then you fetch all the transfer. Amount of data should be pretty small (I foresee <20MB).

Please do review the MSMonitor logic though:
https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: It is sphinx placeholder for code (cut-and-paste feature), see https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html?highlight=doctest#doctest-blocks

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: what IS the input data? please be explicit.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: read the code. It does the following:
- read given records and extract dataset
- it gets new transfer IDs (transfer statuses) for a given dataset
- it save them as new record
- it calculates completion value
- it returns new transferIds (transferStatus) and completion value

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: ok, fixed

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: the function literally returns transfer Ids, why do you need to rename it? I don't mind but I want to understand your logic.

And, if we'll provide request ID, then I need to know two things:
- how to get requestID  from WMA request name
- and, I think we should store this requestID in transfer records, otherwise we lost this info and can't use it. So far we store dataset, timstamp, transferIDs

Once you'll provide this info I'll be happy to change the name of API and its input/logic to use requestID

vkuznet:
Alan, regarding your comment:
```
Besides those comments, I think we should fetch the transfer records in bulk such that we can avoid hundreds/thousands of calls, so in the MSMonitor, you fetch all the requests in a given status, then you fetch all the transfer. Amount of data should be pretty small (I foresee <20MB).
```
which bulk operation are you referring to? Please be explicit. The code now uses `getTransferInfo` API from ReqMgrAux. This API has workflow name as an input parameter. If you want to get all records from CouchDB you need to provide appropriate API for that. Otherwise I don't see ANY way to get all records (or do bulk operation). If you're talking about another topic please provide pointer to API(s) which does this bulk operation.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: Done, via this change:
https://github.com/dmwm/WMCore/pull/9276/files#diff-4cd93292a8b855fbad20e8e899adb8b7R46

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: point 2. here lists what "input data" can be:

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: This is exactly what we discussed over that long private thread.
Things from the top of my head - but better getting back there to review what I say - the PhEDEx `subscribe` API returns the request ID (which I also call transfer ID).

As we discussed about the transfer document schema, this request ID is supposed to be persisted in Couch, according to this structure:
https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154#file-transferor_format-json-L32

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: Which is wrong, see the MS Monitor wiki I provided before (point 4).
BTW, this method is apparently doing the same thing as `getStatusRecords`.

amaltaro (changes requested):
Further comments in the code. If something is not clear - after you read the wiki - please let me know.
Otherwise, update the code and give it a very basic test please. If everything is fine, we might be able to request another testbed upgrade tomorrow morning.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: if I read wiki properly the point 2 tells me that input data is a list of datasets (either primary, parent, or pileup datasets). But the point is that you insisted in code to use this call
```
requests = self.reqmgr2.getRequestByStatus([reqStatus], detail=False)
```
see line https://github.com/dmwm/WMCore/pull/9276/files#diff-4cd93292a8b855fbad20e8e899adb8b7R32
This call returns ONLY request names and other information. How can I obtain the input data for a given request? Do I need to make another call? Or, should I change `getRequestByStatus` details to True to get all details? In other words if you want to use input data (list of datasets from request) then we should get them somehow.

Please let me know how you'd like to resolve this.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: Not so fast. Did you ever check what phedex subscriptions API return for given request? Please give it a try, e.g. this url (using dataset input parameter):
https://cmsweb-testbed.cern.ch/phedex/datasvc/xml/prod/subscriptions?dataset=/EG/Run2010A-WZEG-Nov4Skim_v1/RAW-RECO
will return proper subscription detials in no time (call time is 0.016 sec), while this call (using request ID)
https://cmsweb-testbed.cern.ch/phedex/datasvc/xml/prod/subscriptions?request=191805
will return no results in 45 seconds. May be I'm doing something wrong but so far usage of request ID does not seem to work with Phedex subscriptions API

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: And, I insist that we need to get complete status value somehow. I don't mind to remove this method, but you should tell me how to get complete value then?

It is not about persisting records in CouchDB (which is what point 4 is saying) it is about how to find if we completed all transfers. For that we need to take records we put into CouchDB and check them again somehow. The former task (how to get records from Couchdb) is done in `getStatusRecords` while the later task (how to obtain complete status) is done in `checkStatusRecords`. While we obtain a new transfer completion values we can get new transfer IDs (request IDs). That is why checkStatusRecrods is doing both tasks.

vkuznet: Let me rephrase it. Monitor wiki says:
```
using the transfer ids available in the transfer document, make calls to PhEDEx/Rucio in order to get the status of those transfers; then calculate the transfer status completion. 
```
The `checkStatusRecords` is doing exactly this:
- take status records (transfer documents)
- it calls `getTransferIds` APi for each record which by itself
- calls PhEDEx subscriptions API
- find out (calculate) completion value

I dont' understand do you want to get rid of this method, rename it or change its logic.

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: Interesting, I thought we could simply provide the request ID parameter to that API, but indeed it takes a long time to resolve it (even with a newer example/subscription).
It looks like we need to provide either the dataset or the block name, together with the request ID. In that case, we can quickly get a "valid" output, e.g.:
https://cmsweb-testbed.cern.ch/phedex/datasvc/json/prod/subscriptions?request=1857599&block=/RelValH125GGgluonfusion_14/Integ_Test-SC_LumiMask_PhEDEx_HG1908_Val_Alanv11-v11/NANOAODSIM%23*

We might have to ping Natalia on this item once again.

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: That's right, "input data" is a list of datasets.
Getting requests with "detail=false" returns only the request name indeed, but that's why I said we need to then retrieve the transfer record (which is named after the workflow name), that transfer document will have all the input data related information that we need (including transfer IDs).

In short,
1. we retrieve workflow names in the `staging` status
2. then we retrieve its transfer document in order to proceed with the input dataset monitoring/check (note here there is a PS., as mentioned somewhere else in this PR about bulk get)

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: I'd rather not mix status records with transfer records. The way I see it is:
* getTransferIds: retrieve the transfer IDs for a given workflow
* checkStatusRecords (or renamed to checkTransferCompletion): uses the transfer IDs, query PhEDEx and calculate the dataset/transfer completion (can be done in multiple methods if you like).

So I see no use for `getStatusRecords`. In addition to that, `checkStatusRecords` is not supposed to generate any new transfer ID, it's supposed to receive transfer IDs, and treat them.

vkuznet: ok, I did the following:
- renamed `checkStatusRecords` to `completion` method, so far it does very simple operation. I decided to have it if we need to implement more complex algorithm.
- new method takes transfer records and return completion value

amaltaro:
Valentin, about getting the transfer docs in bulk, it can be done using the same `getTransferInfo` API from ReqMgrAux module, by providing `ALL_DOCS` as the document name.

I updated this information here as well:
https://github.com/dmwm/WMCore/wiki/ReqMgr2-apis#retrieve-a-workflow-transfer-document

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: so, what should we do? So far I left everything as is, i.e. I used dataset as input parameter. Until you clarify how this should be done I don't think we need to change anything.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: ok, not it is clear. I changed input parameter of pileup method and implemented basic logic according to wiki (it may not be complete though).

vkuznet:
ok, here is what I've done:
- use bulk call via ALL_DOCS, I'm not convinced that it will be faster and I put proper comment in a code
- I renamed checkRecordsStatus to completion method, and make later accept transfer records to find out completion value
- I renamed statusRecords into transferRecords for better code clarify
- I implemented basic pileup logic based on wiki description (further tweaking may be required)
- I left as is subscriptions call since Phedex API does not provide call via requests only, at the moment I don't see need to change anything until it is clarified
- added completion threshold , read it from configuration

I think at this moment all requests are implemented. Feel free to review again.

amaltaro:
Thanks, Valentin. I'll look into it tomorrow.
Just one more comment. The completion threshold does *not* come from the service configuration. It will either come from the unified config or from the campaign one.

Last but not least, can you please make a nice description in the first comment of this PR explaining what are the changes and so on? Thanks

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,92 @@
...
+                requestStatuses.setdefault(wname, []).append(statusRecord)
 
-    def post(self):
-        """
-        Runs any post tasks before exiting the execution cycle
-        :return:
-        """
-        pass
+        # update/insert requestStatues in couchdb
+        self.updateTransferInfo(requestStatuses)
```
amaltaro: It crashes in this line.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -1,102 +1,92 @@
...
+                        # update: assigned -> staging
                         self.logger.debug("### transfer request for %s successfull", reqName)
                         self.change(req, 'staging', '### transferor')
                         # if there is nothing to be transferred (no input at all),
                         # then update the request status once again staging -> staged
                         # self.change(req, 'staged', '### transferor')
         except Exception as err:  # general error
             self.logger.exception('### transferor error: %s', str(err))
+            # compose status record
+            wname = reqName  # in our case workflow name is identical to request name
```
amaltaro: UnboundLocalError: local variable 'reqName' referenced before assignment

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,190 @@
...
+        :param reqName: name of the request
+        :return: list of of status records
+        """
+        transferRecords = []
+
+        # in our case workflow name is the same as request name
+        wname = reqName
+        # get existing transfer IDs record
+        doc = self.getTransferInfo(wname)
+        # loop over all records and obtain new transfer IDs
```
amaltaro:
```
if not doc:
   self.logger.error("Failed to find a transfer document for request: %s.", wname)
   return transferRecords
```

or something like that. Then we need to skip this workflow on the for loop calling this function.

amaltaro:
Valentin, just fix those problems such that nothing crashes, and we can proceed and merge it.
I'll then work on their logic in another PR.

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSTransferor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: done

vkuznet:
Alan, I fixed issues you pointed out and it require additional code refactoring (I was need to move common methods from MSMonitoring/MSTransferor to MSCore).

vkuznet:
I also fixed reading threshold from unified config and updated description of this issue.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,134 @@
...
+        try:
+            # get requests from ReqMgr2 data-service for given statue
+            # here with detail=False we get back list of records
+            requests = self.reqmgr2.getRequestByStatus([reqStatus], detail=False)
+            self.logger.debug('+++ monit found %s requests in %s state',
+                              len(requests), reqStatus)
+
+            # FIXME: completion threshold should come either from unified
+            # or campaign configuration
+            thr = self.unifiedConfig().get('completedThreshold', 100)
```
amaltaro: please comment this line and hard-wire it to 100 (I'll come back to it tomorrow or so), otherwise...
```
2019-08-01 16:04:03,496:ERROR:MSMonitor: +++ monit error: 'list' object has no attribute 'get'
  File "/data/srv/HG1908d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.1.2.pre3/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSMonitor.py", line 35, in execute
    thr = self.unifiedConfig().get('completedThreshold', 100)
AttributeError: 'list' object has no attribute 'get'
```

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -0,0 +1,134 @@
...
+            tids, completion = self.getTransferIds(dataset)
+            rec = {'dataset': dataset, 'dataType': dtype,
+                   'transferIDs': tids, "completion": completion}
+            transferRecords.append(rec)
+        return transferRecords
+
+    def pileupTransfersCompleted(self, transferRecords):
+        "Check if pileup transfers are completed for given transfer records"
+        # loop over all records and obtain new transfer IDs
+        thr = self.msConfig.get('completedThreshold', 100)
```
amaltaro: same comment here, please hard-wire it to 100. It's anyways wrong, it should use unified config, not the microservice one.

amaltaro (changes requested):
Valentin, there are two more lines to fix. Please just hard-code those things because I want to have this PR in asap. Such that we can start working heavily on the logic.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: done

vkuznet:
Alan, I put thr = 100 in a code and put in comments that we should fix it and provide example.
Feel free to proceed.

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -0,0 +1,130 @@
...
+        # eventually will change it to Rucio
+        self.phedex = PhEDEx(httpDict={'cacheduration': 10 * 60},
+                             dbsUrl=self.msConfig['dbsUrl'], logger=self.logger)
+
+    def unifiedConfig(self):
+        """
+        Fetches the unified configuration
+        :return: unified configuration content
+        """
+        return self.reqmgrAux.getUnifiedConfig(docName="config")
```
amaltaro:
```suggestion
        res = self.reqmgrAux.getUnifiedConfig(docName="config")
        if res:
            return res[0]
        else:
            return {}
```

vkuznet: I fixed it in more conservative way, I added check to res to be list.

In general, return types of WMCore APIs is somewhat fragile and inconsistent, but it is a different story.

vkuznet:
Alan, I submitted the fix for unifiedConfig.

amaltaro:
Thanks, it seems to be working fine now. Once we hear back from Jenkins, I'll merge it and put in a new tag for Lina for testbed tomorrow.

amaltaro (approved):

PR #9281, opened 2019-07-03
Print the complete error message in LogDB

### LINKED ISSUE #9280: LogDB post API doesn't report error details
I've seen it here and there. And now again on DBS3Upload while testing a few things. 
Problem is that an `HTTPException` gets raised and casting it to string yields an empty string.

See small traceback as follows:
```
2019-07-03 10:52:41,601:140686135908096:WARNING:DBSUploadPoller:Failed to fetch parentage map from WMStats, skipping this cycle.Error: Service Unavailable
2019-07-03 10:52:41,618:140686135908096:ERROR:LogDB:LogDBBackend post API failed, error=
2019-07-03 10:52:41,619:140686135908096:INFO:BaseWorkerThread:DBSUploadPoller took 0.060 secs to execute
2019-07-03 10:52:41,628:140686135908096:ERROR:LogDB:Failed to delete doc in LogDB. Reason: Service Unavailable, status: 503
```

So we have to fetch the proper exception attributes and put it in the error message (just as the `delete` API does).

amaltaro:
Fixes #9280 

#### Status
not-tested

#### Description
As the subject says, provide error details when an HTTPException gets raised.

yes

no

### DIFF
--- a/src/python/WMCore/Services/LogDB/LogDB.py
+++ b/src/python/WMCore/Services/LogDB/LogDB.py
@@ -59,6 +59,7 @@ class LogDB(object):
 
     def post(self, request=None, msg="", mtype="comment"):
         """Post new entry into LogDB for given request"""
+        res = 'post-error'
         try:
             if request is None:
                 request = self.default_user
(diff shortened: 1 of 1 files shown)


amaltaro:
Tests went fine.

PR #9283, opened 2019-07-03
Update deployment with the new PUT request behaviour for the agent config

### LINKED ISSUE #9257: Update wmagentconfig call on deploy-wmagent.sh
As a follow up of this PR: https://github.com/dmwm/WMCore/pull/9250

we need to update those curl commands which are PUTing data into wmagentconfig/resource to make sure they either run a in-place update or that we send the whole document/content in the PUT request, otherwise it's going to break the format of those documents right at the deployment.

amaltaro:
Fixes #9257 

#### Status
not-tested

#### Description
Allow the configuration to be overriden in the script to upload the agent configuration to the reqmgr_aux couchdb.
Then adapted the deployment script to use this option.

yes

no

### DIFF
--- a/bin/wmagent-upload-config
+++ b/bin/wmagent-upload-config
@@ -2,17 +2,31 @@
 from __future__ import division, print_function
 
 import os
+import sys
+import json
 from pprint import pformat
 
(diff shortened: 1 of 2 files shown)


amaltaro:
Tested in my VM.

PR #9285, opened 2019-07-04
Add only unique error exit codes to the job FJR 

### LINKED ISSUE #9115: Don't duplicate error report in the FJR
Apparently multicore workflows can add the same exception/error multiple times into the FJR,
e.g.:
amaltaro_ReReco_badBlocks_HG1904_Validation_190320_153739_3035

before persisting it in the FJR, we might compare whether the same exit code + exception type has already occurred in that step, if so, then we just skip it.

amaltaro:
Fixes #9115

#### Status
not-tested

#### Description
When adding an error to the job FJR, check whether that exit code doesn't exist yet, if it does, then don't add it again.

yes

#### Related PRs
no

no


amaltaro:
This seems to be working fine. Further tests can be performed in the next WMAgent release candidate.

PR #9288, opened 2019-07-07
Use logging library in the reqmgr2 client and inject test wfs

amaltaro:
#### Status
not-tested

#### Description
As the subject says, convert logging from print to the logging library.

yes

#### Related PRs
no

no


amaltaro:
It works well.

PR #9296, opened 2019-07-15
Initial refactoring of MS Transferor; continuously fetch and persist the Unified config in Couch

### LINKED ISSUE #9227: Persist the unified configuration in couch
In short, we need to carry the following actions here:
* get the up-to-date Unified configuration and push it to couch
* make it available through a REST API (I think we already have one...)
* update the MS Transferor code, getting rid of the local json file and using the ReqMgrAux unified API.

amaltaro:
Fixes #9227 

#### Status
ready

#### Description
This PR started with the goal to persist and use the unified configuration from central CouchDB. However, there were a few other things that would greatly improve the code readability and maintainability, so I went ahead and made those changes as well. In short:
* remove the unified json file and the global variable carrying that info
* updated the ReqMgr2 `AuxCacheUpdateTasks` CP thread to continuously update the unified configuration in Couch. It runs every hour and fetches the configuration from GH Ops repository (link is configurable)
* a few changes to the `ReqMgrAux` client class, such as:
  * accept a `cacheduration` parameter. We have to use data caching in some places (like unified and campaign configuration);
  * implemented the read/write/update APIs for the unified config
* don't overload the unified configuration with URL values that are not there (thus returning a hard-wired config). Instead, use service deployment values where needed.

I have done a whole reorganization of the code as well (mostly no logic changes), such as:
* use consistent logger format
* created a new entry point `MSManager` for the application, which is supposed to simply manage the application and the transferor and monitor threads
* created a `MSTransferor` class, to keep the high level logic of the transferor thread
  * also sliced the amount of work the transferor thread will execute in parallel to avoid a possible lack of resources (hard-wired to 50 requests at the moment, can be done configurable in the future)
* many changes to the `RequestInfo` module, keeping all request-wise information in there
* hard-wired a few monitoring URLs in the `SiteInfo` module, they should probably be removed eventually. 
  * also removed the mechanism which fetches agent information `agentSites`, I see no need for that at this stage.
* last but not least, removed a bunch of unused functions/methods (or that won't be used for data placement)

No

Yes, it requires these deployment changes:
https://github.com/dmwm/deployment/pull/795


amaltaro:
@vkuznet Valentin, please read the description of this PR carefully (checking the code as well) and let me know if you have any concerns/questions/suggestions. 
I'm going to merge it such that we can move on with other things.

vkuznet:
Alan, please change title of this PR appropriately. It is more about code re-factoring than current title.

PR #9302, opened 2019-07-22
Added transferinfo REST API to access ReqMgrAux transfer related docs

### LINKED ISSUE #9198: Create a new REST API for the MS Transferor to store/retrieve transfer IDs

amaltaro:
Fixes #9198 

#### Status
not-tested

#### Description
Created a new ReqMgr2 `transferinfo` REST API to be used to manage `TRANSFER` type documents under the reqmgr_aux DB.
There was no need to create a couch view, because it had already the precise view I needed (which was buggy though).

Last but not least, I made it possible to fetch all the documents under a given document/config type by providing a document name `ALL_DOCS`.

no, it depends on a new version of ReqMgr2 (and agents need to have the client part).

#### Related PRs
no

no

### DIFF
--- a/src/python/WMCore/ReqMgr/Service/Auxiliary.py
+++ b/src/python/WMCore/ReqMgr/Service/Auxiliary.py
@@ -16,6 +16,7 @@ from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 from WMCore.REST.Server import RESTEntity, restcall, rows
 from WMCore.REST.Tools import tools
 from WMCore.ReqMgr.DataStructs.ReqMgrConfigDataCache import ReqMgrConfigDataCache
+from WMCore.Services.RequestDB.RequestDBReader import RequestDBReader
 
 
 class MissingPostData(RESTError):
@@ -148,6 +149,7 @@ class AuxBaseAPI(RESTEntity):
         RESTEntity.__init__(self, app, api, config, mount)
         # CouchDB auxiliary database name
         self.reqmgr_aux_db = api.db_handler.get_db(config.couch_reqmgr_aux_db)
+        self.reqmgr_aux_db_service = RequestDBReader(self.reqmgr_aux_db, couchapp="ReqMgrAux")
         self.setName()
 
     def setName(self):
@@ -169,7 +171,10 @@ class AuxBaseAPI(RESTEntity):
         """
         try:
             if subName:
-                docName = "%s_%s" % (self.name, subName)
+                if subName.lower() == "all_docs":
+                    return rows(self._getAllDocs())
+                else:
+                    docName = "%s_%s" % (self.name, subName)
             else:
                 docName = self.name
             sw = self.reqmgr_aux_db.document(docName)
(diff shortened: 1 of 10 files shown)


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
@vkuznet can you please have a look at this PR? It implements the REST apis to pull/push transfer documents to couch.

src/python/WMCore/ReqMgr/Service/Auxiliary.py
```diff
@@ -169,7 +171,10 @@ def get(self, subName=None):
         """
         try:
             if subName:
-                docName = "%s_%s" % (self.name, subName)
+                if subName == "ALL_DOCS":
```
vkuznet: To be on a safe side, may be here we should use
```
if subName.lower() == 'all_docs':
```

test/python/WMCore_t/Services_t/ReqMgr_t/ReqMgr_t.py
```diff
@@ -140,7 +140,7 @@ def testRequestSimpleCycle(self):
         self.assertEqual(len(response), 1)
         clonedName = response[0]['request']
         response = self.reqSvc.getRequestByNames(clonedName)
-        self.assertEqual(response[0][clonedName]['TimePerEvent'], 15)
+        self.assertEqual(response[0][clonedName]['TimePerEvent'], 73.85)
```
vkuznet: could you please put comment why this number should be 73.85?

vkuznet (commented):
Alan, I put only two comments, everything else looks fine to me

test/python/WMCore_t/Services_t/ReqMgr_t/ReqMgr_t.py
amaltaro: this value comes from the json template we load in the setUp() step

src/python/WMCore/ReqMgr/Service/Auxiliary.py
amaltaro: Is it a concern that something might change the arguments encoded in the url? Or just to make it easier to people interacting with this API?

vkuznet:
First, it is unclear why capitals should be used.
Second, in curl we use _all_docs (lower case) to access all docs in given
database.

Therefore, I thought it would be beneficial and consistent to use lower case
notation.

vkuznet:
Again, put the comment in a code to avoid question I raised.
Plus it would be beneficial to explain why this value is used, i.e.
why 73.85 and not 1231.2349829304?

amaltaro:
`_all_docs` is used when querying Couchdb directly. That's why I choose all capitals, to not confuse a couch native API with a ReqMgr2 REST API.
Nonetheless, I updated the code to consider all combinations of capital/non-capital.

PR #9273, opened 2019-06-24
Add new parser script to parse Unified Campaigns

vkuznet:
Fixes 

#### Status
ready

#### Description
This PR provides a new script which can be used to parse Unified campaigns either from given input json file or from Unified MongoDB. Here is a typical usage:
```
bin/adhoc-scripts/parseUnifiedCampaigns.py --help
usage: PROG [-h] [--fin FIN] [--verbose VERBOSE] [--url URL] [--dburi DBURI]
            [--dbname DBNAME] [--dbcoll DBCOLL]

optional arguments:
  -h, --help         show this help message and exit
  --fin FIN          Input file
  --verbose VERBOSE  verbosity level
  --url URL          url of server where we should upload campaign configs
  --dburi DBURI      MongoDB URI
  --dbname DBNAME    MongoDB database
  --dbcoll DBCOLL    MongoDB collection

# parse unified campaigns and stored them to external file
./parseUnifiedCampaigns.py --dburi=mongodb://vocms0274.cern.ch:27017 --dbname=unified --dbcoll=campaignsConfiguration > ~/public/UnifiedCampaigns.txt

# parse unified campaigns and push results into couchdb
./parseUnifiedCampaigns.py --dburi=mongodb://vocms0274.cern.ch:27017 --dbname=unified --dbcoll=campaignsConfiguration --url=https://vk-reqmgr.cern.ch/reqmgr2/data

```

#### Is it backward compatible (if not, which system it affects?)
YES

#### Related PRs
#9139 

#### External dependencies / deployment changes
No

PR #9306, opened 2019-07-24
script to fetch and parse campaign configs from Unified MongoDB

### LINKED ISSUE #9228: Persist the campaign configurations in couch

amaltaro:
Partially fixes #9228 , we still have to find a way to continuously update campaign configs in CouchDB.

#### Status
ready

#### Description
Script to access the Unified MongoDB, fetch all the campaign configuration records, parse them (convert from Unified to WMCore schema) and - if needed - to inject the new campaign configuration in ReqMgr auxiliary DB.

Most of the work has been done by Valentin here: https://github.com/dmwm/WMCore/pull/9273
this PR contains a few bug fixes and allows someone to dump a final json file (in case the place we're running this script doesn't have the WMCore library).

yes

#### Related PRs
Yes, continuation of #9273

no

### DIFF
--- a/bin/adhoc-scripts/parseUnifiedCampaigns.py
+++ b/bin/adhoc-scripts/parseUnifiedCampaigns.py
@@ -1,0 +1,301 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# pylint: disable=R0903
+# R0903: too-few-public-methods
+"""
+File       : ParseUnifiedCampaigns.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
+Description: Parse Unified campaigns json record or read them
+from Unified MongoDB. The yielded campaign records are ready
+for CouchDB and follow the campaign schema:
+```
+{
+  "campaign_A": [record1, record2],
+  "campaign_B: [...]
+}
+```
+where individual records have the following structure:
+```
+{
+ "CampaignName": "blah2019",
+ "PrimaryAAA": boolean,
+ "SecondaryAAA": boolean,
+ "SiteWhiteList": ["T1", "T2"],
+ "SiteBlackList": [...],
+ "SecondaryLocation": ["T1", "T2"] # list with sites to hold a whole copy of the dataset
+ "Secondaries": {
+    "datasetA": ["list of sites for SiteWhitelist"],
+    "datasetB": ["list of sites for SiteWhitelist"]
+ },
+ "MaxCopies": integer
+}
+```
+
+Example of execution:
+python parseUnifiedCampaigns.py --dburi=mongodb://localhost:27017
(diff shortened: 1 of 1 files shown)


amaltaro:
@vkuznet please review. I've ran many tests, including the upload of documents to central couch - a total of 313 campaign configuration at this very moment - in this case against my private VM:
https://alancc7-cloud1.cern.ch/couchdb/_utils/database.html?reqmgr_auxiliary

@thongonary could you please spare a couple of minutes and carry out a spot check on the output campaign configuration, I dumped all of them here
http://amaltaro.web.cern.ch/amaltaro/forAlan/output.json

I've made many comparisons already, but just in case I missed something.

thongonary:
Why SecondaryLocation is always blank in campaigns with premix library?

amaltaro:
It's only empty if the secondaries dict doesn't define it (or if it's empty in there). For instance, campaign `RunIISpring16DR80` has the proper values, no?

thongonary:
@amaltaro Most campaigns have SecondaryLocation at the top-level (like `RunIIAutumn18FSPremix`). Also `RunIISpring16DR80` is a very old campaign which was created 3 years ago -- nowadays we reserve `DR` for classical mixing (no secondary location) and `DRPremix` for premixing (with secondary location).

amaltaro:
Good catch, @thongonary !
I was missing the `SecondaryLocation` from the top level dict, since it was not in the `remap` dict.
Fix was pushed in the 2nd commit (which will be squashed later) and a new output json was put in place.

Would you mind having another quick look? :)

thongonary:
Looks good. I'm just wondering whether we should change `SecondaryLocation` to `PremixLocation` to reduce confusion (for the future generation 😄).

amaltaro:
IF we can make the campaign configuration something better and much more understandable (and maintanainable), I'm all for it ;) I'm just reluctant to diverge too much from the original campaign configuration kept/used by Ops/Unified.

I know Sharad has the first attempt in a WmAgentScripts PR.

vkuznet (commented):
Alan, code looks good to me, feel free to proceed.

amaltaro:
Thanks @thongonary and @vkuznet

PR #9317, opened 2019-07-26
Add PartialCopy attribute and fix issue with parsing campaigns.json file

### LINKED ISSUE #9316: Add partial_copy parameter to the WMCore campaign documents and parser
As being discussed on this Ops ticket:
https://github.com/CMSCompOps/WmAgentScripts/issues/435

their campaign configuration will soon(?) have a new parameter called `partial_copy`, which we should also take into account in the campaign docs stored in couch.
In addition to that, we need to update the parser script:
https://github.com/dmwm/WMCore/blob/master/bin/adhoc-scripts/parseUnifiedCampaigns.py

I believe the default value will be `1`, but let's wait for the other ticket to be updated.
Further details of its definition and implementation can be seen on the other ticket too.

vkuznet:
Fixes #9316 

#### Status
ready

#### Description
Added PartialCopy attribute and fixed issue with parsing given campaigns.json file instead of MongoDB data source.

YES

### DIFF
--- a/bin/adhoc-scripts/parseUnifiedCampaigns.py
+++ b/bin/adhoc-scripts/parseUnifiedCampaigns.py
@@ -27,13 +27,16 @@ where individual records have the following structure:
     "datasetA": ["list of sites for SiteWhitelist"],
     "datasetB": ["list of sites for SiteWhitelist"]
  },
- "MaxCopies": integer
+ "MaxCopies": integer,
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
Thanks, Valentin. I'll probably inject the campaign records to testbed by tomorrow.

Issue #9139, opened 2019-04-09
Verify the Campaign APIs and define a schema


amaltaro:
We need to make sure Ops can properly create new Campaign configurations, edit and delete using the REST APIs. Documentation also needs to be created.

On top of it, we need to revisit the Campaign structure and define (and enforce) a schema.

Here is the schema discussed along this thread:
```
{
 "CampaignName": "blah2019",
 "PrimaryAAA": boolean, # default false
 "SecondaryAAA": boolean, # default false
 "SiteWhiteList": ["T1_*", "T2_*"],
 "SiteBlackList": [...],
 "SecondaryLocation": ["list with sites to hold a whole copy of the dataset"],
 "Secondaries": {"datasetA": ["list of sites for SiteWhitelist"], "datasetB": ["list of sites for SiteWhitelist"]},
 "MaxCopies": integer
}
```

where:
* `CampaignName`: string corresponding to the campaign name; which is the top level keys of the campaign.json file.
* `PrimaryAAA`: boolean to flag whether to use AAA for the primary dataset; mapped from the `primary_AAA` top level key.
* `SecondaryAAA`: boolean to flag whether to use AAA for the secondary dataset; mapped from the `secondary_AAA` key, which can be *either* at top level or under the `secondaries` dictionary. If it appears multiple times, we make an OR of the values.
* `SiteWhiteList`: list of sites where the workflow gets assigned to and where the primary and secondary CLASSIC MIX dataset is placed (likely in chunks of data); mapped from the `SiteWhitelist` key which can be AFAIK in multiple places, like: top level dict, under the `parameters` key and under the `secondaries` dictionary. If it appears multiple times, we make an intersection of the values.
* `SiteBlackList`: list of sites where the workflow does NOT get assigned to and where the primary and secondary CLASSIC MIX dataset is NOT placed; mapped from the `SiteBlacklist` key which can be AFAIK in multiple places, like: top level dict, under the `parameters` key and under the `secondaries` dictionary. If it appears multiple times, we make an *union* of the values.
* `SecondaryLocation`: list of sites where the secondary PREMIX dataset has to be placed as a whole (not necessarily also assigned to those). mapped from the `SecondaryLocation` key, which can be *either* at top level or under the `secondaries` dictionary. If it appears multiple times, we make an intersection of the values.
* `Secondaries`: dictionary with a map of allowed secondary datasets and where they are supposed to be placed; mapped from the key top level `secondaries`.
[...]

PR #9325, opened 2019-08-02
Adjust MSMonitoring logic according to the requirements

vkuznet:
Fixes #<GH_Issue_Number>

#### Status
in development

#### Description
Adjust MSMonitoring logic according to requirement document:
https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor

This PR contains also changes to ReqMgrAux to cover additional mocking functions as well as MSMonitoring_t unit test.

MAYBE

No

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSCore.py
+++ b/src/python/WMCore/MicroService/Unified/MSCore.py
@@ -6,9 +6,6 @@ Description: MSCore class provides core functionality of the MS.
 # futures
 from __future__ import division, print_function
 
-# system modules
-import time
-
 # WMCore modules
 from WMCore.MicroService.Unified.Common import getMSLogger
 from WMCore.Services.ReqMgr.ReqMgr import ReqMgr
@@ -67,66 +64,3 @@ class MSCore(object):
             self.reqmgr2.updateRequestStatus(reqName, reqStatus)
         except Exception as err:
             self.logger.exception("Failed to change request status. Error: %s", str(err))
-
-    def updateTransferInfo(self, requestStatuses):
-        "Update transfer ids in backend"
-        tstamp = time.time()
-        # Alan's suggestion to use bulk query
-        # get all docs from CouchDB
-        docs = self.getTransferInfo('ALL_DOCS')
-        for wname, statusRecord in requestStatuses:
-            # obtain records from CouchDB
-            # doc = self.getTransferInfo(wname)
-            # find record in our list of docs
-            # VK: I don't know if it will be faster then fetching the record
-            #     in CouchDB, we need to loop every time to find proper record
-            doc = {}
-            for doc in docs:
-                if wname in doc:
-                    break
-            records = doc.get('transfers', [])
-            records += statusRecord
-            doc['workflowName'] = wname
-            doc['timestamp'] = tstamp
-            doc['transfers'] = records
-            self.reqmgrAux.postTransferInfo(wname, doc)
-
-    def getTransferInfo(self, wname):
-        """
-        Get transfer document from backend. The document has the following form:
-
-        .. doctest::
-
-            {"workflowName": "bla",
-             "timestamp": 123,
-             "transfers": [rec1, rec2, ... ]
-            }
-            where each record has the following format:
-            {"dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2], "completion": 0}
-
-
-        :param wname: workflow name
-        :return: a tranfer ids document
-        """
-        return self.reqmgrAux.getTransferInfo(wname)
-
-    def getTransferIds(self, dataset):
-        """
-        Get transfer ids document for given request name and datasets.
-        :param dataset: dataset name
-        :return: a list of transfer ids
-        """
-        # phedex implementation, TODO: implement Rucio logic when it is ready
-        data = self.phedex.subscriptions(
-            dataset=dataset, group=self.msConfig['group'])
-        self.logger.debug(
-            "### dataset %s group %s", dataset, self.msConfig['group'])
-        self.logger.debug("### subscription %s", data)
-        tids = []
-        vals = []
-        for row in data['phedex']['dataset']:
-            if row['name'] == dataset:
-                for rec in row['subscription']:
-                    vals.append(int(rec['percent_files']))
-                    tids.append(int(rec['request']))
-        return tids, float(sum(vals))/len(vals)
--- a/src/python/WMCore/MicroService/Unified/MSMonitor.py
+++ b/src/python/WMCore/MicroService/Unified/MSMonitor.py
@@ -8,6 +8,10 @@ the transferor monitoring module.
 # futures
 from __future__ import division, print_function
 
+# system modules
+import json
+import time
+
 # WMCore modules
 from WMCore.MicroService.Unified.MSCore import MSCore
 
@@ -17,9 +21,41 @@ class MSMonitor(MSCore):
     MSMonitor class provide whole logic behind
     the transferor monitoring module.
     """
+    def __init__(self, msConfig, logger=None):
+        super(MSMonitor, self).__init__(msConfig, logger)
+        # update interval is used to check records in CouchDB and update them
+        # after this interval, default 6h
+        self.updateInterval = self.msConfig.get('updateInterval', 6*60*60)
+
+    def updateCaches(self):
+        """
+        Fetch some data required for the monitoring logic, e.g.:
+         * all campaign configuration
+         * all transfer records from backend DB
+        :return: True if all of them succeeded, else False
+        """
+        campaigns = self.reqmgrAux.getCampaignConfig("ALL_DOCS")
+        transferRecords = [d for d in self.getTransferInfo('ALL_DOCS')]
+        cdict = {}
+        if not campaigns:
+            self.logger.warning("Failed to fetch campaign configurations")
+        if not transferRecords:
+            self.logger.warning("Failed to fetch transfer records")
+        else:
+            for camp in campaigns:
+                if 'CampaignName' not in camp:
+                    self.logger.warning(
+                        'No CampaignName attribute in campaign dict: %s',
+                        json.dumps(camp))
+                    continue
+                cdict[camp['CampaignName']] = camp
+        return cdict, transferRecords
+
     def execute(self, reqStatus):
         """
-        Executes the MS monitoring logic
+        Executes the MS monitoring logic, see
+        https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor
+
         :param reqStatus: request statue to process
         :return:
         """
@@ -30,110 +66,140 @@ class MSMonitor(MSCore):
             self.logger.debug('+++ monit found %s requests in %s state',
                               len(requests), reqStatus)
 
-            # FIXME: completion threshold should come either from unified
-            # or campaign configuration, e.g.
-            # thr = self.unifiedConfig().get('completedThreshold', 100)
-            thr = 100
+            campaigns, transferRecords = self.updateCaches()
+            if not campaigns or not transferRecords:
+                # then wait until the next cycle
+                msg = "Failed to fetch data from one of the data sources. Retrying again in the next cycle"
+                self.logger.error(msg)
+                return
 
-            requestStatuses = {}  # keep track of request statuses
+            # keep track of request and their new statuses
+            requestsToStage = []
+            # main logic
             for reqName in requests:
-                req = {'name': reqName, 'reqStatus': reqStatus}
-                self.logger.debug("+++ request %s", req)
-                # obtain status records from couchdb for given request
-                transferRecords = self.getTransferRecords(reqName)
-                if not transferRecords:
+                self.logger.debug("+++ request %s", reqName)
+                # obtain transfer records for our request
+                transfers = []
+                for rec in transferRecords:
+                    if reqName == rec['workflowName']:
+                        transfers = rec['transfers']
+                        break
+                if not transfers:
                     continue
-                completion = self.completion(transferRecords)
                 # if all transfers are completed
                 # move the request status staging -> staged
-                if completion == thr:
+                if self.completion(transfers, campaigns):
                     self.logger.debug(
-                        "+++ request %s all transfers are completed", req)
-                    self.change(req, 'staged', '+++ monit')
+                        "+++ request %s all transfers are completed", reqName)
+                    requestsToStage.append(reqName)
                 # if pileup transfers are completed AND
                 # some input blocks are completed,
                 # move the request status staging -> staged
-                elif self.pileupTransfersCompleted(transferRecords):
+                elif self.completion(transfers, campaigns, pileup=True):
                     self.logger.debug(
-                        "+++ request %s pileup transfers are completed", req)
-                    self.change(req, 'staged', '+++ monit')
+                        "+++ request %s pileup transfers are completed", reqName)
+                    requestsToStage.append(reqName)
                 # transfers not completed
                 # just update the database with their completion
                 else:
                     self.logger.debug(
-                        "+++ request %s transfers are %s completed",
-                        req, completion)
-                    # for us workflow name is the same as request name
-                    wname = reqName
-                    requestStatuses[wname] = transferRecords
-            self.updateTransferInfo(requestStatuses)
+                        "+++ request %s transfers are completed", reqName)
+            self.updateTransferInfo(transferRecords)
+            # finally, update statuses for requests
+            for reqName in requestsToStage:
+                self.change(reqName, 'staged', '+++ monit')
         except Exception as err:  # general error
             self.logger.exception('+++ monit error: %s', str(err))
 
-    def getTransferRecords(self, reqName):
+    def completion(self, transfers, campaigns, pileup=None):
         """
-        Get transfer records for given request name from CouchDB.
-        Transfer records on backend has the following form
-        https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154
-        So far logic is based on option D of the records
+        Helper function to calculate completion of given records
+        :param transfers: list of transfers records
+        :param pileup: check pileup completion (optional)
+        :return: completion status
+        """
+        # check completion of all transfers
+        statuses = []
+        transferTypes = ['primary', 'secondary']
+        for rec in transfers:
+            campaign = rec['CampaignName']
+            cdict = campaigns[campaign]
+            thr = cdict.get('PartialCopy', 0)
+            if pileup and rec['dataType'] not in transferTypes:
+                continue
+            if rec['completion'] >= thr:
+                status = 1
+            else:
+                status = 0
+            statuses.append(status)
+        return True if sum(statuses) == len(transfers) else False
+
+    def updateTransferInfo(self, transferRecords):
+        """
+        Update transfer records in backend
+        :param transferRecords: list of transfer records
+        """
+        tstamp = time.time()
+        for doc in transferRecords:
+            transfers = []
+            for rec in doc.get('transfers', []):
+                # obtain new transfer ids and completion for given dataset
+                _, completion = self.getTransferIds(rec['dataset'])
+                # Per Alan request, we'll update only completion and not tids
+                rec.update({'completion': completion})
+                transfers.append(rec)
+            wname = doc['workflowName']
+            doc['lastUpdate'] = tstamp
+            doc['transfers'] = transfers
+            self.reqmgrAux.updateTransferInfo(wname, doc, inPlace=True)
+
+    def getTransferIds(self, dataset):
+        """
+        Get transfer ids document for given request name and datasets.
+        :param dataset: dataset name
+        :return: a list of transfer ids and completion value
+        """
+        # phedex implementation, TODO: implement Rucio logic when it is ready
+        data = self.phedex.subscriptions(
+            dataset=dataset, group=self.msConfig['group'])
+        self.logger.debug(
+            "### dataset %s group %s", dataset, self.msConfig['group'])
+        self.logger.debug("### subscription %s", data)
+        tids = []
+        vals = []
+        for row in data['phedex']['dataset']:
+            if row['name'] == dataset:
+                for rec in row['subscription']:
+                    vals.append(int(rec['percent_files']))
+                    tids.append(int(rec['request']))
+        if not vals:
+            return tids, 0
+        return tids, float(sum(vals))/len(vals)
+
+    def getTransferInfo(self, wname):
+        """
+        Get transfer document from backend. The document has the following form:
 
         .. doctest::
 
             {"workflowName": "bla",
-             "timestamp": 123,
+             "lastUpdate": 123,
              "transfers": [rec1, rec2, ... ]
             }
             where each record has the following format:
-            {"dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2], "completion":0}
+            {"dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2], "completion": 0}
 
-        :param reqName: name of the request
-        :return: list of of status records
+
+        :param wname: workflow name
+        :return: a generator of transfer records
         """
-        transferRecords = []
-
-        # in our case workflow name is the same as request name
-        wname = reqName
-        # get existing transfer IDs record
-        doc = self.getTransferInfo(wname)
-        if not doc:
-            self.logger.error("Failed to find a transfer document for request: %s", wname)
-            return transferRecords
-
-        # loop over all records and obtain new transfer IDs
-        for rec in doc['transfers']:
-            dataset = rec['dataset']
-            dtype = rec['dataType']
-            # obtain new transfer ids for given request name and dataset
-            tids, completion = self.getTransferIds(dataset)
-            rec = {'dataset': dataset, 'dataType': dtype,
-                   'transferIDs': tids, "completion": completion}
-            transferRecords.append(rec)
-        return transferRecords
-
-    def pileupTransfersCompleted(self, transferRecords):
-        "Check if pileup transfers are completed for given transfer records"
-        # FIXME: completion threshold should come either from unified
-        # or campaign configuration, e.g.
-        # thr = self.unifiedConfig().get('completedThreshold', 100)
-        thr = 100
-
-        records = 0
-        transferTypes = ['primary', 'secondary']
-        # loop over all records and obtain new transfer IDs
-        for rec in transferRecords:
-            ctype = rec['dataType']
-            completion = rec['completion']
-            # count records above threshold
-            if ctype in transferTypes and completion > thr:
-                records += 1
-        return True if records == len(transferRecords) else False
-
-    def completion(self, records):
-        """
-        Helper function to calculate completion of given records
-        :param records: list of status records
-        :return: completion value
-        """
-        # may need to implement proper algorithm, so far we take average
-        val = [r['completion'] for r in records]/len(records)
-        return val
+        records = self.reqmgrAux.getTransferInfo(wname)
+        # according to https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor
+        # we may need to select only record since last updated according
+        # to configuration UpdateInterval setting
+        for rec in records:
+            if self.updateInterval:
+                if time.time()-rec['lastUpdate'] > self.updateInterval:
+                    yield rec
+            else:
+                yield rec
--- a/src/python/WMQuality/Emulators/ReqMgrAux/MockReqMgrAux.py
+++ b/src/python/WMQuality/Emulators/ReqMgrAux/MockReqMgrAux.py
@@ -1,5 +1,15 @@
+"""
+MockReqMgrAux class provides mocking methods of ReqMgrAux class
+"""
+# futures
 from __future__ import (division, print_function)
+
+# system modules
+import time
+
+# WMCore modues
 from WMCore.Agent.DefaultConfig import DEFAULT_AGENT_CONFIG
+
 
 class MockReqMgrAux(object):
     """
@@ -7,10 +17,54 @@ class MockReqMgrAux(object):
     """
 
     def __init__(self, *args, **kwargs):
-        print("Using MockReqMgrAux")
+        print("Using MockReqMgrAux", args, kwargs)
+        tstamp = int(time.time())
+        cname = 'testCampaign'
+        cdict = {"CampaignName": cname,
+                 "PrimaryAAA": True,
+                 "SecondaryAAA": False,
+                 "SiteWhiteList": ["T1", "T2"],
+                 "SiteBlackList": [],
+                 "SecondaryLocation": ["T1", "T2"],
+                 "Secondaries": {"datasetA": ["T1", "T2"], "datasetB": ["T1"]},
+                 "MaxCopies": -1,
+                 "PartialCopy": 0
+	}
+        self.campaigns = [cdict]
+        self.transferRecords = []
+        rec = {"dataset":"/a/b/c", "dataType": "primary",
+               "transferIDs": [1,2], "completion": 1, "CampaignName": cname}
+        self.transferRecords = \
+                [{'workflowName': 'test', 'lastupdate':tstamp, 'transfers':[rec]}]
 
     def getWMAgentConfig(self, agentName):
         """
         macking getWMAgentConfig returns default config.
         """
         return DEFAULT_AGENT_CONFIG
+
+    def getCampaignConfig(self, name):
+        "mocking getCampaignConfig method"
+        campaigns = []
+        if name == 'ALL_DOCS':
+            return self.campaigns
+        for camp in self.campaigns:
+            if camp['CampaignName'] == name:
+                campaigns.append(camp)
+        return campaigns
+
+    def getTransferInfo(self, name):
+        "mocking getTransferInfo method"
+        if name == 'ALL_DOCS':
+            return self.transferRecords
+        records = []
+        for rec in self.transferRecords:
+            if rec['workflowName'] == name:
+                records.append(rec)
+        return records
+
+    def updateTransferInfo(self, name, doc, inPlace=False):
+        "mocking updateTransferInfo method"
+        # while mocking we do nothing
+        print('mock method updateTransferInfo: name=%s, doc=%s inPlace=%s' \
+                % (name, doc, inPlace))
--- a/test/python/WMCore_t/MicroService_t/Unified_t/MSMonitor_t.py
+++ b/test/python/WMCore_t/MicroService_t/Unified_t/MSMonitor_t.py
@@ -1,0 +1,88 @@
+"""
+Unit tests for Unified/Monitor.py module
+
+Author: Valentin Kuznetsov <vkuznet [AT] gmail [DOT] com>
+"""
+from __future__ import division, print_function
+
+# system modules
+import json
+import unittest
+
+# WMCore modules
+from WMCore.MicroService.Unified.MSMonitor import MSMonitor
+from WMQuality.Emulators.EmulatedUnitTestCase import EmulatedUnitTestCase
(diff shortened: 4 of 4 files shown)


vkuznet:
Alan, I created this PR as a placeholder to adjust MSMonitoring logic according to wiki page. In order to move forward I need the following input from you:
- the wiki sais in point 3 to obtain campaign names and its configuration
  - do we have multiple campaign names for single request or single?
  - if I understand correctly we should use reqmgrAux `getCampaignConfig` API, right?
 - the wiki sais in point 4a that we should look at transfer types above the minimum completion thresholds as defined in campaign schema
  - where this threshold is defined in campaign schema, i.e. according to recent campaign schema we put in https://github.com/dmwm/WMCore/blob/master/bin/adhoc-scripts/parseUnifiedCampaigns.py#L19  there is no threshold attribute. So the question is where to find it?
  - if we have multiple campaign schema docs, how to use their thresholds for given request? Is it average value or what?

But you also wrote before in https://github.com/dmwm/WMCore/pull/9276 that threshold may come from Unified configuration. So I need clarify here, where exactly should I extract it and how. Please provide pointer to existing unified configuration or campaign schema(s) where this/these value(s) is/are defined.

Once this information will be cleared then I can proceed with final fix for MSMonitor logic. So far I fixed obvious parts.

amaltaro:
Valentin, replies below:
> the wiki sais in point 3 to obtain campaign names and its configuration
>    do we have multiple campaign names for single request or single?

yes, campaign is defined in multiple places in a request (top level or inside a Task/Step). Then we should see where the input data appear on the request and use the correspondent campaign. Unless Unified does something different...

>    if I understand correctly we should use reqmgrAux getCampaignConfig API, right?

Yes, but the get has to be in bulk, not get every single campaign document as needed.

> where this threshold is defined in campaign schema, i.e. according to recent campaign schema we put in https://github.com/dmwm/WMCore/blob/master/bin/adhoc-scripts/parseUnifiedCampaigns.py#L19 there is no threshold attribute. So the question is where to find it?

It should be here: https://github.com/dmwm/WMCore/blob/master/bin/adhoc-scripts/parseUnifiedCampaigns.py#L205 
but there might be something else at unified configuration level.

> if we have multiple campaign schema docs, how to use their thresholds for given request? Is it average value or what?

it's not average. As said above, each schema define the input data somewhere, we then have to see which campaign is defined in that structure. Just an example:
https://github.com/dmwm/WMCore/blob/master/test/data/ReqMgr/requests/DMWM/TC_PreMix.json

defines `MCPileup` at different level, but always under `RunIIAutumn18DRPremix` campaign name, even though the request has other campaign names.

vkuznet:
Alan, your answers do not clarify how to get the threshold. Please be specific:
- does threshold should be taken from campaign or Unified configurations?
- if it comes from campaign then we already have a schema for that, from which part of the schema it comes from? From `partialCopy` parameter? The `partialCopy` parameter 1 means 100%? I need clear definition here since we read PhEDEx subscriptions which is in percentage. So the threshold should be in percentage.
   - it also implies that we should read every time campaigns from ReqMgrAux and look-up threshold under campaign name?
- if it comes from Unified configuration please provide pointer where to read it

Better, if you'll provide an API in ReqMgrAux to use for reading threshold it would be solve these questions. Then I can use the API in MS.

amaltaro:
> does threshold should be taken from campaign or Unified configurations?

this is part of the black box logic, which is still not crystal clear to me.

> if it comes from campaign then we already have a schema for that, from which part of the schema it comes from? From partialCopy parameter? The partialCopy parameter 1 means 100%? I need clear definition here since we read PhEDEx subscriptions which is in percentage. So the threshold should be in percentage.


Yes, from L205, as per the link I sent you (and the parameter is `PartialCopy`).
Yes, 1 == 100%

>     it also implies that we should read every time campaigns from ReqMgrAux and look-up threshold under campaign name?

Yes, that's why we need to set a `cacheduration` when creating a ReqMgrAux object., I believe 1h cache would be good enough.

> Better, if you'll provide an API in ReqMgrAux to use for reading threshold it would be solve these questions. Then I can use the API in MS.

You just need a dict get attribute. If we make an API for that, I foresee a sub-optimal usage of the campaign GET API.

vkuznet:
ok, then I'll proceed with reading campaigns configuration for given workflow name and extracting PartialCopy as threshold.

vkuznet:
One more question. MSMonitoring code gets request names and operates with them. Now, in order to get threshold I need to use `getCampaignConfig` API from ReqMgrAux. This API requires campaign name. The question is how to get campaign names knowing ONLY request name? I can probably fetch all campaign configs via `ALL_DOCS` input parameter (if it works?) but our campaign configuration does not hold request/workflow name in docs.

Another way is to NOT USE `detail=False` parameter when we fetch requests, and instead fetch all details of the request (which may contain campaign name(s)).

Please clarify.

amaltaro:
MSMonitor will heavily depend on the transfer documents stored in Couch (by MSTransferor).
From the document structure:
https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154#file-transferor_format-json-L32

that's where you'll find the campaign name.

And yes, let's fetch all the campaign documents (by providing `ALL_DOCS` as document name) with a single call and then just use it where it's needed.

vkuznet:
Alan, I reviewed changes in this PR and they good to go since none of them depends on changes in your #9330 , i..e all changes are independent.

This PR may only need to adjust default UpdateInterval which should be used. You should tell me what is suitable value, so far I used large number (4 days).

You may review the logic now.

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -76,8 +76,9 @@ def updateTransferInfo(self, requestStatuses):
         "Update transfer ids in backend"
```
amaltaro: Valentin, if this method is updating a transfer document, then I'd expect it not to get fresh new documents; but instead, either receive an updated document or a list of them.

amaltaro: In addition to that, this method is only used by MSMonitor, so we better move it there.

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -95,7 +96,7 @@ def updateTransferInfo(self, requestStatuses):
             doc['transfers'] = records
             self.reqmgrAux.postTransferInfo(wname, doc)
 
-    def getTransferInfo(self, wname):
+    def getTransferInfo(self, wname, since=None):
```
amaltaro: Same comment here, move it over to MSMonitor.
In addition to that, provided that the data loaded isn't too much, we should get all those transfer docs in bulk.

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -110,9 +111,20 @@ def getTransferInfo(self, wname):
...
+        :return: a generator of transfer records
         """
-        return self.reqmgrAux.getTransferInfo(wname)
+        records = self.reqmgrAux.getTransferInfo(wname)
+        # according to https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor
+        # we may need to select only record since last updated according
+        # to configuration UpdateInterval setting
+        for rec in records:
+            if since:
+                interval = self.msConfig.get('UpdateInterval', 4*24*60*60)
```
amaltaro: let's make it every 6h please. And don't forget updating the deployment configuration file.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -30,47 +32,41 @@ def execute(self, reqStatus):
             self.logger.debug('+++ monit found %s requests in %s state',
                               len(requests), reqStatus)
 
-            # FIXME: completion threshold should come either from unified
-            # or campaign configuration, e.g.
-            # thr = self.unifiedConfig().get('completedThreshold', 100)
-            thr = 100
-
             requestStatuses = {}  # keep track of request statuses
+            since = self.msConfig.get('UpdateInterval', 4*24*60*60)
```
amaltaro: should we move this setting to the __init__ method?

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -30,47 +32,41 @@ def execute(self, reqStatus):
...
             for reqName in requests:
                 req = {'name': reqName, 'reqStatus': reqStatus}
                 self.logger.debug("+++ request %s", req)
                 # obtain status records from couchdb for given request
-                transferRecords = self.getTransferRecords(reqName)
+                transferRecords = self.getTransferRecords(reqName, since)
```
amaltaro: See comment I made above. Going workflow by workflow gives us the best memory footprint, but a poor cpu performance. I'd rather have the best cpu performance and an average memory footprint. Thus, let's get data in bulk when possible.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -30,47 +32,41 @@ def execute(self, reqStatus):
...
+                        "+++ request %s transfers are completed", req)
                     # for us workflow name is the same as request name
                     wname = reqName
                     requestStatuses[wname] = transferRecords
             self.updateTransferInfo(requestStatuses)
         except Exception as err:  # general error
             self.logger.exception('+++ monit error: %s', str(err))
 
-    def getTransferRecords(self, reqName):
+    def getTransferRecords(self, reqName, since=None):
```
amaltaro: sort of duplicate of `getTransferInfo`...

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -94,46 +92,35 @@ def getTransferRecords(self, reqName):
...
-                records += 1
-        return True if records == len(transferRecords) else False
-
-    def completion(self, records):
+    def completion(self, transferRecords, pileup=None):
         """
         Helper function to calculate completion of given records
         :param records: list of status records
+        :param pileup: calcualate completion for pileups (optional)
         :return: completion value
```
amaltaro: it's returning boolean

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -94,46 +92,35 @@ def getTransferRecords(self, reqName):
...
         """
-        # may need to implement proper algorithm, so far we take average
-        val = [r['completion'] for r in records]/len(records)
-        return val
+        # check completion of all transfers
+        status = False
+        transferTypes = ['primary', 'secondary']
+        for rec in transferRecords:
+            campaign = rec['campaignName']
+            cdict = self.reqmgrAux.getCampaignConfig(campaign)
```
amaltaro: load all the campaigns first, then just use their data.

amaltaro (changes requested):
Valentin, I have made several comments along the code.

Given that most of my suggestions is to collect data in bulk, I think we should have a mechanism similar to what I implemented in the MSTransferor:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/MSTransferor.py#L46

such that we decide whether we have to spend cpu cycles or not based on the information retrieved.

I've also made a minor revision of the Monitor documentation (same logic, but better described):
https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor

src/python/WMCore/MicroService/Unified/MSCore.py
vkuznet: Alan, apart from moving it from MSCore to MSMonitor (since MSTransferor now don't use this method) what your first comment asked to do? The method get list of documents and post them to Couch. The input list is passed from execute method where it gets list of new transfer records for given request. Please elaborate what you actually want to do with this method. I don't see anything to be done except moving method from MSCore to MSMonitor.

amaltaro:
So, one of the problems is that MSMonitor is fetching the transfer documents multiple times. It only needs to fetch then once per cycle, use them if needed, and update solely two keys (lastUpdate and completion). Here is an example of transfer document:
```
{
   "_id": "TRANSFER_amaltaro_StepChain_MC_Aug2019_Val_190816_143916_9956",
   "_rev": "1-92f0afec0052405b8eec2fb58cac4530",
   "ConfigType": "TRANSFER",
   "lastUpdate": 0,
   "workflowName": "amaltaro_StepChain_MC_Aug2019_Val_190816_143916_9956",
   "transfers": [
       {
           "completion": 0,
           "dataType": "secondary",
           "campaignName": "Aug2019_Val",
           "transferIDs": [
               1111
           ],
           "dataset": "/Neutrino_E-10_gun/RunIISpring15PrePremix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v2-v2/GEN-SIM-DIGI-RAW"
       }
   ]
}
```

Which means, your method `updateTransferInfo` doesn't need to fetch the transfer documents again. In addition to that, you do not pass the `transfers` section only, but you pass the whole transfer document and simply `POST` it to ReqMgrAux.

This
https://github.com/dmwm/WMCore/blob/d8d4a401484ce74c803f58569eb4ca085ad9ebad/src/python/WMCore/MicroService/Unified/MSMonitor.py#L118

is the only place you update the `completion` transfer status, however it never gets executed because you always call that method with ALL_DOCS.

One more thing, I noticed you compare transfer records against a workflow name, keep in mind the workflow name is under record['workflowName'].

amaltaro:
And I have two more comments:
* the `change` call has to be the very last thing to be executed, such that we only update a request status if all the rest has been successfully executed (including updating the transfer document).
* I think it's time to start playing with the phedex calls here and complete the loop

vkuznet:
Alan, thanks for comments. I adjusted code accordingly. But after re-factoring I found that I need further input from you. I put appropriate comment in a code, see https://github.com/dmwm/WMCore/pull/9325/files#diff-4cd93292a8b855fbad20e8e899adb8b7R140

Basically MSMonitor wiki says nothing about competed transfers. What should we do with transfer documents if transfer is completed and we move request to `staged`. Should we discard transfer records in CouchDB? I think so. Therefore, I put appropriate comment in a code but so far disable it. If this is the case we need a `deleteTransferInfo`.

amaltaro:
About the transfer document cleanup, I'm actually considering to leave those in the system for a little while (if the performance impact is acceptable). Maybe only delete them when the request advances its status to something that's not `staging|staged|acquired`.

Anyways, don't bother about couch cleanup because I believe we already have a thread taking care of that. I'll then extend it before we push it to production.

vkuznet:
ok, I removed the part with clean-up. This PR is ready for review.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,9 +20,18 @@ class MSMonitor(MSCore):
     MSMonitor class provide whole logic behind
     the transferor monitoring module.
     """
+    def __init__(self, msConfig, logger=None):
+        super(MSMonitor, self).__init__(msConfig, logger)
+        # update interval is used to check records in CouchDB and update them
+        # after this interval
+        self.updateInterval = self.msConfig.get('UpdateInterval', 6*24*60*60)
```
amaltaro: defaut to 6h, not 6 days

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -30,52 +42,105 @@ def execute(self, reqStatus):
...
-                # obtain status records from couchdb for given request
-                transferRecords = self.getTransferRecords(reqName)
-                if not transferRecords:
+                self.logger.debug("+++ request %s", reqName)
+                # for us workflow name is the same as request name
+                wname = reqName
+                # obtain transfer records for our request
+                transfers = []
+                for rec in transferRecords:
+                    if reqName in rec:
```
amaltaro: Isn't `rec` the transfer record? If so, it doesn't have a key named after the workflow.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -30,52 +42,105 @@ def execute(self, reqStatus):
...
+            if rec['completion'] >= thr:
+                status = True
+        return status
+
+    def getCampaignConfig(self, campaign):
+        "Helper function to get given campaign from backend or cache"
+        if campaign not in self.campaigns:
+            campaigns = self.reqmgrAux.getCampaignConfig("ALL_DOCS")
+            for cdict in campaigns:
+                if cdict not in self.campaigns:
```
amaltaro: Testing if a dict is in another dict? If won't work

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -30,52 +42,105 @@ def execute(self, reqStatus):
...
+            if doc['workflowName'] in completedWorkflows:
+                continue  # we skip completed workflows
+            transfers = []
+            for rec in doc.get('transfers', []):
+                # obtain new transfer ids for given request name and dataset
+                tids, completion = self.getTransferIds(rec['dataset'])
+                rec.update({'transferIDs': tids, 'completion': completion})
+                transfers.append(rec)
+            wname = doc['workflowName']
+            doc['timestamp'] = tstamp
```
amaltaro: We don't use such key in the transfer doc.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -84,56 +149,19 @@ def getTransferRecords(self, reqName):
...
-        # may need to implement proper algorithm, so far we take average
-        val = [r['completion'] for r in records]/len(records)
-        return val
+        records = self.reqmgrAux.getTransferInfo(wname)
+        # according to https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor
+        # we may need to select only record since last updated according
+        # to configuration UpdateInterval setting
+        for rec in records:
+            if self.updateInterval:
+                if time.time()-rec['timestamp'] > self.updateInterval:
```
amaltaro: same here, no `timestamp` key.

amaltaro (changes requested):
Valentin, I made several comments in the code. It wasn't a complete review and I'm pretty sure there are other problems. I suggest you fix those, get this PR tested and if everything works fine, then we can have another review iteration.

BTW, the way you fetch the campaign configuration won't work. You're apparently caching the campaign forever. I'd suggest you once again to look at MSTransferor and how it does the initial cycle setup.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: done

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: changed to lastupdate

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: replaced with lastupdate

vkuznet:
Alan, in order to do proper testing of MSMonitor we need lots of stuff in place, e.g. campaign configurations, transfer records, etc.

I start writing unit tests and found immediately that I can't proceed since no appropriate records are in place. For instance, if I use cmsweb-testbed.cern.ch instance and ask for campaign configurations I got these records:
```
# if I do this call (part of MSMonitor codebase)
# campaigns = self.reqmgrAux.getCampaignConfig("ALL_DOCS")
# I got this record
{"ConfigType": "CAMPAIGN_CONFIG", "flag": false, "_id": "CAMPAIGN_CONFIG_testB", "camp": "testB"}
```
while according to https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/ReqMgrAux/ReqMgrAux.py#L172 I see a totally different structure of campaign records.

Therefore in order to proceed we need all pieces in place somewhere. Better, we need mocking functions for MSTransferor which will populate CouchDB with some records, etc.

Do you mind to organize proper configuration that they will be accessible in testbed or any other instance.

vkuznet:
Alan, I added two things: new methods for MockReqMgrAux and unit test for MSMonitoring where I tested basic functionality of the methods. But as I wrote to perform full test with data you need to populate necessary pieces elsewhere that MSMonitoring code will be able to read campaigns, transfer records, etc.

I'll wait for jenkins test and adjust code if necessary, if passed, then it will be ready again for review.

amaltaro:
Valentin, given that you're the main developer behind these MicroServices, I'd rather help you to get the testing setup in place, otherwise it's a waste of the both of our time going back and forth on review/test/fix over and over again.

Given that you have access to Unified MongoDB instance and know how to fetch/parse the campaigns, I've also added another option to insert them into your own ReqMgrAux instance, see:
https://github.com/dmwm/WMCore/pull/9346

second step is to disable this CP thread:
https://github.com/dmwm/deployment/blob/master/reqmgr2/config.py#L166
otherwise it's going to proceed with request status transition and it might be that MSMonitor doesn't manage to find any in `staging` status

and the last step is to inject test workflows, which you've already been doing.

BTW, I just noticed this PR
https://github.com/dmwm/WMCore/pull/9330/files#diff-24825ae3f77716fa4f3123460f4f52c8L69

went in without the `readOnly` check in the `change()` method. So both MSTransferor and MSMonitor are able to complete the whole cycle (and I might to add that check back before testbed/production).

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -68,53 +65,12 @@ def change(self, reqName, reqStatus, prefix='###'):
...
-            where each record has the following format:
-            {"dataset":"/a/b/c", "dataType": "primary", "transferIDs": [1,2], "completion": 0}
-
-
-        :param wname: workflow name
-        :return: a tranfer ids document
-        """
-        return self.reqmgrAux.getTransferInfo(wname)
 
     def getTransferIds(self, dataset):
```
amaltaro: This one doesn't belong here as well and we should move it to MSMonitor.
In addition to that, it's not getting any transfer ID, but it's actually getting the subscription status for a dataset and we should change things accordingly.
Last but not least, given that the phedex status will be done via GET call, we should try to use `multi_getdata` and get things in parallel.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
+        if not campaigns:
+            self.logger.warning("Failed to fetch the campaign configurations")
+        else:
+            # let's make campaign look-up easier and more efficient
+            self.campaigns = {}
+            for camp in campaigns:
+                if 'campaignName' not in camp:
+                    self.logger.warning(
+                            'No campaignName attribute in campaign dict: %s',
+                            json.dumps(camp))
```
amaltaro: `camp` is a python dict, no need to dump it as a json in the logs.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
+        :return: True if all of them succeeded, else False
+        """
+        campaigns = self.reqmgrAux.getCampaignConfig("ALL_DOCS")
+        if not campaigns:
+            self.logger.warning("Failed to fetch the campaign configurations")
+        else:
+            # let's make campaign look-up easier and more efficient
+            self.campaigns = {}
+            for camp in campaigns:
+                if 'campaignName' not in camp:
```
amaltaro: `CampaignName`. Same comment for a few lines below

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
     the transferor monitoring module.
     """
+    def __init__(self, msConfig, logger=None):
+        super(MSMonitor, self).__init__(msConfig, logger)
+        # update interval is used to check records in CouchDB and update them
+        # after this interval, default 6h
+        self.updateInterval = self.msConfig.get('updateInterval', 6*60*60)
+        self.campaigns = {}
+
+    def updateCaches(self):
```
amaltaro: This should also be fetching all the transfer documents in bulk

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
     def execute(self, reqStatus):
         """
-        Executes the MS monitoring logic
+        Executes the MS monitoring logic, see
+        https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor
+
         :param reqStatus: request statue to process
         :return:
         """
+        if not self.updateCaches():
```
amaltaro: and we better call `updateCaches` after getting the list of requests in `reqStatus`. So somewhere after this line:
https://github.com/dmwm/WMCore/pull/9325/files#diff-4cd93292a8b855fbad20e8e899adb8b7R72

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
-
-            requestStatuses = {}  # keep track of request statuses
+            # get list of all transfer records
+            transferRecords = [d for d in self.getTransferInfo('ALL_DOCS')]
+            # keep track of completed workflows
+            completedWorkflows = []
+            # keep track of request and their new statuses
+            reqStatus = {}
+            # main logic
             for reqName in requests:
```
amaltaro: It might give us a better performance if we iterate over the `transferRecords` and check if the workflow name is in `requests`. If we do so, we need to have a mechanism which reports workflows in staging status (in `requests`) but without a correspondent transfer document though.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
+                self.logger.debug("+++ request %s", reqName)
+                # for us workflow name is the same as request name
+                wname = reqName
+                # obtain transfer records for our request
+                transfers = []
+                for rec in transferRecords:
+                    if reqName == rec['workflowName']:
+                        transfers = rec['transfers']
+                        break
+                if not transfers:
```
amaltaro: This has no distinction between a transfer document without any transfers FROM the lack of a transfer document for a given workflow. We know every workflow should have an associated transfer document, but in case it doesn't, the very least we need to do is to throw an ERROR in the logs.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
+        Helper function to calculate completion of given records
+        :param transfers: list of transfers records
+        :param pileup: check pileup completion (optional)
+        :return: completion status
+        """
+        # check completion of all transfers
+        statuses = []
+        transferTypes = ['primary', 'secondary']
+        for rec in transfers:
+            campaign = rec['campaignName']
```
amaltaro: `CampaignName`

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
                 # some input blocks are completed,
                 # move the request status staging -> staged
-                elif self.pileupTransfersCompleted(transferRecords):
+                elif self.completion(transfers, pileup=True):
                     self.logger.debug(
-                        "+++ request %s pileup transfers are completed", req)
-                    self.change(req, 'staged', '+++ monit')
+                        "+++ request %s pileup transfers are completed", reqName)
+                    completedWorkflows.append(wname)
+                    reqStatus[reqName] = 'staged'
```
amaltaro: It will always be to the same status, no need to keep it as a dict

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
+
+    def updateTransferInfo(self, transferRecords, completedWorkflows):
+        """
+        Update transfer records in backend
+        :param transferRecords: list of transfer records
+        :param completedWorkflows: list of completed workflow names to be removed from backend
+        """
+        tstamp = time.time()
+        for doc in transferRecords:
+            if doc['workflowName'] in completedWorkflows:
```
amaltaro: I think we should still update these transfer records, such that we can go back and see what was the last snapshot MSMonitor had for each of those transfers/workflows.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
+                continue  # we skip completed workflows
+            transfers = []
+            for rec in doc.get('transfers', []):
```
amaltaro: The transfer ID is set in stone by now and it's not responsibility of the MSMonitor to change/update it (well, not at this stage where we don't have transfers troubleshooting). It will only update the transfer completion and `lastUpdate` timestamp.

src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -17,123 +21,170 @@ class MSMonitor(MSCore):
...
-        # may need to implement proper algorithm, so far we take average
-        val = [r['completion'] for r in records]/len(records)
-        return val
+        records = self.reqmgrAux.getTransferInfo(wname)
+        # according to https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor
+        # we may need to select only record since last updated according
+        # to configuration UpdateInterval setting
+        for rec in records:
+            if self.updateInterval:
+                if time.time()-rec['lastupdate'] > self.updateInterval:
```
amaltaro: `lastUpdate`

src/python/WMQuality/Emulators/ReqMgrAux/MockReqMgrAux.py
```diff
@@ -1,16 +1,66 @@
...
     """
 
     def __init__(self, *args, **kwargs):
-        print("Using MockReqMgrAux")
+        print("Using MockReqMgrAux", args, kwargs)
+        tstamp = int(time.time())
+        cname = 'testCampaign'
+        cconf = {"go": True, "labels":["a", "b"],
+                 "overflow": {"PRIM": {}},
+                 "DDMcopies": {},
```
amaltaro: This is not a WMCore campaign configuration!

amaltaro: And a few other keys below don't match our campaign/transfer document schema

amaltaro (changes requested):
Comments made in the code.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: Here I totally lost. In most of the code the convention is to use `lowerCase` instead of `CamelCase` notations. Why on earth you insist to use `CampaignName` instead of `campaignName`. To give you a few examples we used already `dataType`, `transferIDs`, `updateInterval`, `workflowName`. I think you're breaking your own rules and I don't really know where to use certain schema naming convention.

Please clarify?

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: I'm not following here. First we loop over all requests we obtained from ReqMgr2. So, my understanding that each can progress independently from each other. Therefore each can carry its own status. One may be staged and another not. That's why I keep track between request names and their statuses.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: and I think they should be discarded, since if you update them their update timestamp will advance. Then we'll need to introduce another thread which will monitor requests statuses to decide when we should wiped out records from CouchDB instead of just relying on timestamp. Do you really want to add another complexity?

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: I'm not sure that it is fixed? It is up to upstream system (PhEDEx or Rucio) to decide when/how to transfer blocks. Therefore I would expect that transfer ids set will/may change over time. There is no hard to update them either if their are fixed since they will never change. Please clarify your thoughts on this before requesting the change.

src/python/WMCore/MicroService/Unified/MSMonitor.py
vkuznet: Alan, that's why I wrote about. PLEASE BE CONSISTENT in naming convention through the code. You used `CampaignName`, `dataType` and `lastupdate` (see https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154) ALL OF THEM are different naming conventions.

I'll fix it here , but I followed what was in your gist document.

src/python/WMQuality/Emulators/ReqMgrAux/MockReqMgrAux.py
vkuznet: And, where is campaign document schema?

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: You're mixing up the document types. We're dealing mostly with two different documents in CouchDB:
* **campaign**: schema extensively discussed and agreed as in this issue: https://github.com/dmwm/WMCore/issues/9139#issue-430869532
* **transfers**: also discussed in lengthy and schema defined here: https://gist.github.com/amaltaro/72599f995b37a6e33566f3c749143154#file-transferor_format-json-L33

Most of the WMCore couchdb documents follow the capital camel case convention, like workload_cache and workqueue elements. So perhaps the transfer one should be adapted to follow the same as well...

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: It could be. However, in the end you only have two options here:
a) request remains in the same status (doesn't matter what is that)
b) request moves to the next status, which is always `staged`

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: document cleanup will be based on request status only, no need to make it complex. So updating the timestamp or not makes no difference.

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: PhEDEx/Rucio are simply making the data movement that someone/something ordered them to do. So once a request ID is created, PhEDEx/Rucio will consider that and start the data movement. Of course a given data is not associated to only 1 request, but what matters to us is what MicroService is requesting, we can't control what others request.

src/python/WMCore/MicroService/Unified/MSMonitor.py
amaltaro: Sorry, let me update that gist with the correct key name (camelCase).

src/python/WMQuality/Emulators/ReqMgrAux/MockReqMgrAux.py
amaltaro: Right now in the parse script that you created: https://github.com/dmwm/WMCore/blob/master/bin/adhoc-scripts/parseUnifiedCampaigns.py#L197-L206
Until we properly define and enforce it.

amaltaro:
replaced by https://github.com/dmwm/WMCore/pull/9348

PR #9326, opened 2019-08-02
Adjust MSTrasferor logic according to requirements

vkuznet:
Fixes #<GH_Issue_Number>

#### Status
in development

#### Description
Adjust MSTransferor logic according to requirement document
https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor#relying-on-reqmgr2-second-proposal

#### Is it backward compatible (if not, which system it affects?)
MAYBE

#### External dependencies / deployment changes
NO

Issue #8324, opened 2017-11-13
Transferor MS module has to subscribe data according to the SiteWhitelist


amaltaro:
As discussed in a private thread, there are basically 2 options for the transferor module, but Jean-Roch has already confirmed how it works (or how it's supposed to work). Snippets of a private email:
"""
How is the transferor supposed to work
in the new schema. Is it:
1. transferor module will distribute the input data among sites - which fulfil the data
placement requirements - and *only then* Unified assigns that workflow. Thus, based
on where data is available or heading to.
OR
2. Unified first decides which sites the workflow has to run (one could read the workflow
is assigned) and *only then* the transferor module starts making phedex subscriptions
to places already decided by assignor/unified
"""

and this was @vlimant  answer:
"
 I think 2. But then we need a handle (setting back from staging to assignment-approved ?) when the destination is set drain afterwards, so as to re-route transfers."

This means transferor module has to query reqmgr2 for "acquired" workflows and then fetch all those blocks from Global workqueue and create phedex subscriptions.
This schema has to be changed as well: https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-APIs

@vkuznet FYI

amaltaro:
This is apparently how the Transferor microservice has been implemented, but we can review and make sure it's doing the right thing.

amaltaro:
@vlimant Jean-Roch, just to make sure we're still on the same page. Data placement is based on the rules defined within each Campaign, right?
The MicroService has a reference to a unified configuration json
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/Common.py#L37

I guess that used to be the file with all the campaign configuration, correct?

vlimant:
there are both a campaign level and general level configurations. I am not sure which one is the one @vkuznet implemented in UnifiedConfiguration

vkuznet:
According to
https://github.com/CMSCompOps/WmAgentScripts
there are two files:
unifiedConfiguration.json
campaigns.json
(there were third file campains.relval.json but it is gone now).

The former is loaded in utils.py module, and later is not used anywhere.

In MicroService code I used unifiedConfiguration.json since it was used in
utils.py module. And I have a env variable to read campaigns.json but where
to use (or should it be used) should come from JR.

vlimant:
campaigns.json used to be loaded by campaignInfo and used through that interface ; the json file is a sync in GH of the content of campaign configuration in mongodb

[...]

PR #9330, opened 2019-08-08
Complete MSTransferor architecture; subscribe data and create transfer docs

### LINKED ISSUE #9195: Review MS Transferor and adapt it according to the new architecture
As per the second proposal on this wiki (picture still to be updated):

we should review the Transferor/Unified code already available in WMCore and follow the new Transferor design.

In short, the Transferor module is responsible for:
* fetching workflows from ReqMgr2 in the status `assigned`, including their json description
  * make a list of unique campaigns in those requests
* fetch campaign configuration from CouchDB (given a unique list of campaigns)
* fetch the unified configuration from CouchDB
* parse the request json and find out which blocks need to be transferred (using DBS)
* run the Transferor algorithm to decide which data go where (following the SiteWhitelist - SiteBlacklist)
* create all phedex subscriptions - or rucio rules
* classify subscriptions IDs by dataset (inside the same workflow) and persist it in Couchdb
* IF ALL these steps succeeded, update the request status to `staging`
* ELSE, record the error in Couch LogDB and restart from scratch in the next cycle

amaltaro:
Fixes #9195 #8324
Superseeds #9326

#### Status
In development

#### Description
Summary of changes:
* created a `Workflow` class to store a workflow data and provide some simple getters/setters
* converted `findParent` to execute concurrent calls
* implemented `findBlockParents` function to find parent blocks based on a child block
* hard-wired DBS production url to the PhEDEx instance, otherwise subscriptions fail to match data on TMDB...
On MSTransferor:
* use CRIC to map from PSN to PNN (and 1h cache)
* fetch unified config, all campaign documents and the CRIC psn/pnn map for every cycle
* order requests by `RequestPriority`, such that higher priority requests get worked on first
* implemented actual data placement (PhEDEx subscription) for any type of data (primary, secondary and parent); either at dataset or block level
* implemented basic data placement logic based on the request `SiteWhitelist` and the campaign configuration
* create the transfer document in reqmgr_aux db
* update the workflow status, if allowed
On RequestInfo:
* didn't change any logic in there, simply added the data/block lookup methods
* TODO: evaluate all the functionality in there and update/merge what's is really needed, otherwise remove what's not
* last but not least, fixed a bug on the PhEDEx subscribe API such that it supports block level subscriptions too

Things still left TO DO, but probably in another pull request:
* consider RunWhitelist/RunBlacklist/LumiList on the input data discovery
* take `MaxCopies` into account and properly implement it
* subscribe data in chunks of XX TB (unified black box logic)
* use another data-source to monitor how much data is subscribed to each PNN and avoid oversubscribing data
* revisit RequestInfo and remove duplicate code (like that loop making data discovery); and multiple calls to the same API with the same data
* check what exactly has to be used from the unified campaign for the data placement

no

#### Related PRs
no

no

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -1,0 +1,242 @@
+"""
+Workflow class provides all the workflow data
+required by MS Transferor
+"""
+from __future__ import division
+
+
+class Workflow(object):
+    """
+    Class to represent a workflow and some helpers to access
+    its information within MS
+    """
+
+    def __init__(self, reqName, reqData):
+        self.reqName = reqName
+        self.data = reqData
+        self.inputDataset = ""
+        self.parentDataset = ""
+        self.pileupDatasets = set()
+        self.campaigns = set()
+        self.dataCampaignMap = []
+        # these blocks structure will be key'ed by the block name and value'd by the block size
+        self.primaryBlocks = {}
+        self.parentBlocks = {}
+        # sort of duplicate info, but we need to have a way to link input to parent block(s)
+        self.childToParentBlocks = {}
+        # pileup don't need to get resolved into blocks, store only their total size
+        self.secondarySummaries = {}
+
+        self.setDataCampaignMap()
+        self.setInputData()
+
+    def __str__(self):
+        """
+        Write out useful information for this object
+        :return:
+        """
+        res = {'reqName': self.reqName, 'inputDataset': self.inputDataset,
+               'parentDataset': self.parentDataset, 'pileupDatasets': self.pileupDatasets,
+               'campaigns': self.campaigns, 'dataCampaignMap': self.dataCampaignMap}
+        return str(res)
+
+    def getName(self):
+        """
+        Get this request name
+        """
+        return self.reqName
+
+    def getDbsUrl(self):
+        """
+        Get the DbsUrl defined in this request
+        """
+        return self.data['DbsUrl']
+
+    def getSitelist(self):
+        """
+        Get the SiteWhitelist minus the black list for this request
+        """
+        return sorted(list(set(self.data['SiteWhitelist']) - set(self.data['SiteBlacklist'])))
+
+    def setDataCampaignMap(self):
+        """
+        Set the association between input data, data type and campaign name
+        in the format of:
+        {"type": "type of input data", "name": "dataset name", "campaign": "campaign name"}
+        Also set a flat set of campaign names
+        """
+        inputMap = {"InputDataset": "primary", "MCPileup": "secondary", "DataPileup": "secondary"}
+
+        if "TaskChain" in self.data or "StepChain" in self.data:
+            innerDicts = []
+            for i in range(1, self.data.get("TaskChain", self.data.get("StepChain")) + 1):
+                innerDicts.append(self.data.get("Task%d" % i, self.data.get("Step%d" % i)))
+        else:
+            # ReReco and DQMHarvesting
+            innerDicts = [self.data]
+
+        data = {}
+        for item in innerDicts:
+            for key in inputMap:
+                if key in item and item[key]:
+                    # use top level campaign if not available in the inner dict
+                    data[item[key]] = dict(type=inputMap[key], name=item[key],
+                                           campaign=item.get('Campaign', self.data.get('Campaign')))
+            # also create a flat list of campaign names
+            if "Campaign" in item and item["Campaign"]:
+                self.campaigns.add(item["Campaign"])
+
+        self.dataCampaignMap = data.values()
+
+    def getDataCampaignMap(self):
+        """
+        Retrieve map of campaign, dataset, dataset type
+        :return: list of dictionaries
+        """
+        return self.dataCampaignMap
+
+    def getCampaigns(self):
+        """
+        Get a set of campaign names used within this request
+        """
+        return self.campaigns
+
+    def setInputData(self):
+        """
+        Parse the request data and set the primary and secondary datasets
+        """
+        for item in self.getDataCampaignMap():
+            if item["type"] == "primary":
+                self.inputDataset = item["name"]
+            elif item["type"] == "secondary":
+                self.pileupDatasets.add(item["name"])
+
+    def getInputDataset(self):
+        """
+        Get this request's input dataset name
+        """
+        return self.inputDataset
+
+    def getPileupDatasets(self):
+        """
+        Get this request's secondary dataset names
+        """
+        return self.pileupDatasets
+
+    def hasParents(self):
+        """
+        Check whether the request has IncludeParents
+        :return:
+        """
+        if not self.inputDataset:
+            return False
+        return self._getValue("IncludeParents") or False
+
+    def setParentDataset(self, parent):
+        """
+        Set the parent dataset name and update the data/campaign map
+        :param parent: string corresponding to the parent dataset name
+        """
+        self.parentDataset = parent
+        self._updateDataCampaignMap(parent)
+
+    def getParentDataset(self):
+        """
+        Return the parent dataset name
+        :return: parent dataset name string
+        """
+        return self.parentDataset
+
+    def setPrimaryBlocks(self, blocksDict):
+        """
+        Sets a list of primary input blocks taking into consideration
+        the BlockWhitelist and BlockBlacklist
+        :param blocksDict: flat dict of block name and block size
+        """
+        ### FIXME: we must consider run and lumi lists too, eventually
+        blockWhite = self._getValue("BlockWhitelist") or []
+        blockBlack = self._getValue("BlockBlacklist") or []
+        if not blockWhite and not blockBlack:
+            self.primaryBlocks = blocksDict
+        else:
+            for name, size in blocksDict.items():
+                if blockBlack and name in blockBlack:
+                    continue
+                if blockWhite and name in blockWhite:
+                    self.primaryBlocks[name] = size
+
+    def getPrimaryBlocks(self):
+        """
+        Retrieve list of input primary blocks
+        """
+        return self.primaryBlocks
+
+    def setSecondarySummary(self, dsetName, dsetSize):
+        """
+        Sets the secondary dataset name and its total size, in bytes
+        :param dsetName: string with the secondary dataset name
+        :param dsetSize: integer with the secondary dataset size
+        """
+        self.secondarySummaries[dsetName] = dsetSize
+
+    def getSecondarySummary(self):
+        """
+        Retrieve list of input secondary datasets and their sizes
+        """
+        return self.secondarySummaries
+
+    def setParentBlocks(self, blocksDict):
+        """
+        Sets a list of parent input blocks taking.
+        NOTE: this list is solely based on the parent dataset, without
+        considering what are the actual input primary blocks
+        :param blocksDict: flat dict of block name and block size
+        """
+        self.parentBlocks = blocksDict
+
+    def getParentBlocks(self):
+        """
+        Retrieve list of input parent blocks
+        """
+        return self.parentBlocks
+
+    def setChildToParentBlocks(self, blocksDict):
+        """
+        Sets a relationship between input primary block and its parent block(s)
+        :param blocksDict: dict key'ed by the primary block, with a list of parent blocks
+        """
+        if len(self.getPrimaryBlocks()) == len(blocksDict):
+            # then there is no block white/black list, assign data by reference
+            self.childToParentBlocks = blocksDict
+        else:
+            # then save parents only for actual input primary blocks
+            for block in blocksDict:
+                if block in self.getPrimaryBlocks():
+                    self.childToParentBlocks[block] = blocksDict[block]
+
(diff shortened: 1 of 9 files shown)


vkuznet (commented):
Alan, this PR reveals variety of problems, let me briefly stress a few:
- having O(N^4), O(N^5) O(n^6) loops is unacceptable, e.g. getParentDataset, getParentBlocks, etc.
- even though I provided a way to fetch data from cmsweb services concurrently via multi_getdata you diverged from this approach back to sequential model by looping over workflow records and obtaining sequentially information about datasets/blocks, etc., see clear flaw in logic getParentDatasts. Instead of sequential model, you should FIRST get list of ALL datasets and fetch their info and then decompose this info into appropriate structures.
- Your implementation of getParentDataset, getInputDataBlocks, getParentBlocks are not thread-safe and can't be parallelized since you're doing in-place changes to the records. This is the way to trouble.
Overall it seems that data-structures are not optimal neither from code performance point of view nor from operational point of view. Dealing with nested dicts means that every time a dict should change a new memory allocation should be made. My bet that amount of RAM this code will require will be very large.

Also, I don't recall that we had any discussion how to do this logic and I don't remember it was written anywhere. My suggestion that you should write down required logic and we should discuss how to implement it efficiently.

amaltaro:
Let me start saying that I have this PR marked as "in development", so changes are still happenning and you shouldn't waste your time reviewing something that might still change a bunch. 

In addition to that, I've put many TODO/FIXME here and in the source code as well as a reminder that I need to get back to those. This PR is meant to get all the parts in place, not necessarily in their final version or most performing implementation.

Anyways, I can answer a few of those comments,
> having O(N^4), O(N^5) O(n^6) loops is unacceptable, e.g. getParentDataset, getParentBlocks, etc.

I agree! However, `getParentDatasets` dataset is O(N^3), in practice it's O(N^2) because the upper loop is just setting different DBS instances (in production it's always DBS prod).

`getParentBlocks` is horrible indeed. I can make it less complex, but for that I'll have to convert a flat list of blocks (dictionaries) into a flat dictionary. Getting the parent blocks is something quite complex already, and if one wants to track the total size of those blocks and make sure to keep their parent association, it makes things even harder. If you really know what I'm trying to do there, you know it can't look nice and be simple.

> even though I provided a way to fetch data from cmsweb services concurrently via multi_getdata you diverged from this approach back to sequential model by looping over workflow records and obtaining sequentially information about datasets/blocks, etc., see clear flaw in logic getParentDatasts. Instead of sequential model, you should FIRST get list of ALL datasets and fetch their info and then decompose this info into appropriate structures.

Please read the code again then. AFAICS, all the required service interaction is done via `multi_getdata` (see the Common module changes). The internal processing of data is still done sequentially though, as it has always been. Otherwise, be specific and point me to the code where you see sequential external calls.

> Your implementation of getParentDataset, getInputDataBlocks, getParentBlocks are not thread-safe and can't be parallelized since you're doing in-place changes to the records. This is the way to trouble.

That data is only accessed by the MSTransferor thread. If you want to run MSTransferor on multi-threading mode, than there is a whole new design to make, starting at the MSCore. If you meant something else, please elaborate.

> Overall it seems that data-structures are not optimal neither from code performance point of view nor from operational point of view. Dealing with nested dicts means that every time a dict should change a new memory allocation should be made. My bet that amount of RAM this code will require will be very large.

Make an specific suggestion then. Again, it's another tradeoff between data structure (and memory) vs processing complexity. For instance, having a list in place of a dict, might require yet another level of `for` loop...

> Also, I don't recall that we had any discussion how to do this logic and I don't remember it was written anywhere. My suggestion that you should write down required logic and we should discuss how to implement it efficiently.

There is nothing new on the logic of the transferor component, see documentation here:

vkuznet:
Alan, to answer your questions I think we need written description of what you're trying to do in a first place. I understand why you want to get list of blocks for datasets in given request. But I don't understand why do you need parent blocks? Where parent datasets/blocks are used in a system? Why do you need this info? Is it responsibility of Transferor? At a surface transferor task is to transfer datasets/blocks to sites. Why does he need to know relationships between datasets/blocks? Should it be covered from outside of Transferor logic.

The problem can be further simplified. If I understand correctly a request contains a set of datasets (any type of them, i.e. parents or not). And, we need to know blocks for them. And, we have series of workflow names. Moreover, I may assume that some subset of datasets/blocks remains the same  across periods of MS cycles. That implies that we don't need to resolve all workflows/datasets/blocks all the time and rather need to have a local cache. So, instead of sequential operations for every request we need to do the following:
- get all transfer records docs from CouchDB which will give us all workflow names and all datasets
- we compose full list of datasets for all workflows which we need to resolve to blocks, here we resolve dataset parents (if necessary) to add parents to the list
- then we can make SINGLE multi_getdata call (which will execute DBS queries concurrently) to get list of blocks
  - this can be further optimized at DB level once we know our full input set of parameters
- we cache info as necessary
- we use all transfer Records docs to map back datasets/blocks to their associative workflows

It means that we need several flat data-structures:
- transfer records (as shown in your gist plan D)
- a dataset blocks associations object per DBS instance

Having this two structures around and updating records accordingly will keep them under control and avoid having nested records. I don't see ANY reason why do you need 3 (!) mostly identical methods getParentDatasets, getInputDataBlocks, getParentBlocks. At the end what actually doing is getting list of blocks for a given dataset list. I would replace all three single single function which takes list of datasets and returns dataset-blocks object (dictionary). I also don't understand why do you need to find block sizes?

To sum-up, it is better to see written document of what you're trying to do then parse a code trying to understand its logic. And, there is nothing written in [1] about datasets/blocks relationships (i.e. parentage) in MS code and neither about additional attributes (i.e. block sizes) you extract from DBS calls.

I made an effort in #9326 and #9325 to cover exactly the points of [1] and [2], and used described logic, but your PR goes to much fine-granularity level which is not yet present in these documents.

[1] https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor
[2] https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Monitor

amaltaro:
@vkuznet I don't expect any other major changes on this PR, so you can start reviewing and I'll follow up on your comments tomorrow morning.
I updated the first description with all the changes that got in; which completes the whole MSTransferor architecture.

Most of the unified logic has still to be revisited/fixed though, here I use the very basic information to get something useful as output. There are also a bunch of FIXME and TODO that needs to be revisited, some are my debugging that will be removed before merging it, others depend on the actual unified logic.

vkuznet:
Alan, I asked specific questions which you still avoid to answer. Here my list:
- why do you need parents at dataset level
- why do you need blocks parents
- why do you need block sizes

While, I reviewing the code I suggest that you do additional benchmarks and re-factoring.  In particular, I'm interested how much RAM will new Workflow class hold for every single request. In other words what is a total RAM usage of all Workflow classes for N requests MS needs to deal with on a regular basis. A simple optimization which immediately can be applied is to strip off all dataset names in block names since they represent useless information, i.e. if you have dataset-block association you only need one string for dataset and then store only block IDs since every block name is dataset#blockID. This will save quite a bit of memory for every single Workflow instance (I would expect that a single dataset can spawn 100-1k blocks not to account for parents).

amaltaro:
About data parentage, it's needed every time a request defines `IncludeParents:True`, where in addition to that file/block being processed, it also processes the parent data (same run/lumi on the parent data).

> why do you need parents at dataset level

because then we can collect all the blocks in that parent dataset and also fetch their block size (will be used for subscriptions in chunks)

> why do you need blocks parents
> why do you need block sizes

reply above.

> While, I reviewing the code I suggest that you do additional benchmarks and re-factoring. In particular, I'm interested how much RAM will new Workflow class hold for every single request.

Yes, I can get that info. But it will really be just an example, not necessarily representative since the workflow module will also hold input data information, and each workflow can vary quite a bit.

> A simple optimization which immediately can be applied is to strip off all dataset names in block names since they represent useless information, i.e. if you have dataset-block association you only need one string for dataset and then store only block IDs since every block name is dataset#blockID. This will save quite a bit of memory for every single Workflow instance (I would expect that a single dataset can spawn 100-1k blocks not to account for parents).

That's a good suggestion and, we might have to implement it if memory footprint is too large. Hopefully not, it's another small complexity on top of tons of small complexities and minor details.

However, we do have chunks of workflows being processed at a time, if the python garbage/object collector is smart enough, it should keep memory usage within a few hundred MBs.

vkuznet:
More about block usage. Why do you care about them? Are you saying that transfers should be done on sites for some subset of blocks? The way phedex subscription takes a dataset and my understanding it will transfer all blocks from a dataset, so this is why I'm asking why do you need blocks? Do you really need to to bookkeeping on block level if transfer takes a dataset? In other words what is a purpose of having blocks which belong to a single dataset on a different sites or yet another rephrase, do you (WMA) benefit from having incomplete dataset at a single site (only subset of dataset blocks)? If there is no reason to keep blocks at a different sites I don't a reason to look-up block info at all, i.e. we still get all dataset (and its parents) and then ask phedex/rucio to make subscription to them, why do we need to care about blocks in workflow class?

amaltaro:
Definitely! Chunks of work are distributed by global workqueue in a block basis, such that we can distribute pieces of a dataset (block) to different sites and agents, where the agent would then create jobs to process those blocks against those sites only. Of course the dataset needs to be completely available on Disk (the sum of all blocks), but not necessarily completely available at the same location.

If we had to transfer a dataset and, we wanted to process a workflow in 5 different sites, it means we would have to have 5 copies of the dataset. CMS **cannot** that disk data volume, so we need to implement a smart data placement.

I'm going to make another section in the wiki page explaining this data placement logic as I see it at the moment. It might still change once we open the black box though

amaltaro:
Valentin, you can find further information about data placement here:
https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor#data-placement-model

it's not a complete description of the data placement business (we'll need to read unified code for that), but should clarify several items.

About the `Workflow` object size, it will vary on the workflow type/description, on the input data to be used, and on the amount of blocks in each input data. However, here are some in-memory size:
* object without any block list (so object created solely with the workflow description): ~46kB
* same object, but after expanding and storing the input blocks (255 blocks in total): ~155kB

Let me know if you need further information. I'll clean any debug left in the code and see if there is anything else that we badly need to get fixed here.

amaltaro:
Valentin, as we discussed today. I ran some basic performance evaluation of the MSTransferor running with this patch in, and processing a set of 12 workflows as defined in the DMWM templates.
1) first time I had PhEDEx subscriptions disabled and the total transferor cycle completed in ~21secs
2) second time I had PhEDEx subscriptions enabled (plus the code to get custom object sizes) and the total time was ~51secs.

The large difference comes basically from 2 items: a) `getParentDatasets()` took 22.1s against 0.1s from the first time (yes, resolving the same dataset), and b) the phedex subscription takes something 0 < time < 3. This wallclock time is based on the whole cycle, thus
* fetching unified configuration and ALL campaign documents from couch
* fetching the PSN to PNN map from CRIC
* run the transferor logic and resolve input data
* put data subscriptions
* store a transfer couch record for each workflow
* change the request status in reqmgr2.

All in all, numbers look pretty good to me.

I can't say the same about memory footprint though, there was a peak of 211220 RSS (kilobytes AFAIK). So a group of _bad_ workflows could go up to 2GB or so...

Inspecting the `Workflow` objects, I saw one of them with a very large size
```
Workflow amaltaro_StepChain_MC_Aug2019_Val_190813_222151_7320 size: 18630442
```

so almost 20MB alone. It contains a very big premix pileup data (/Neutrino) and such pileup samples are always placed as a whole in the storage. So we might check whether it's a premixing PU (AFAIK there is no robust way but looking at the dataset name pattern) and not resolve the blocks and their sizes, that would save A LOT of memory.

amaltaro:
All the previous commits have been squashed.
Now the 2nd commit contains the following changes:
* don't fetch blocks for `secondary` datasets
* fetch total dataset size for `secondary` datasets
* make phedex subscription at dataset level for `secondary` datasets

With that change, we brought RSS memory footprint to almost half (131204 kB). However total time was awful because the `blocksummaries` call took 224secs (apparently not normal behaviour though). 

Anyways, let's see what Yuyi suggests to use in order to fetch a dataset size.

amaltaro:
I ran another battery of tests today and this time the tranferor module took 42secs in total. Still, the `blocksummaries` call is responsible for 75% of that time.

At this point I'm done with my changes and tests for this PR. We have to decide now - advice from @yuyiguo would be greatly appreciated - on what's the best approach/API to retrieve the total size of a dataset.

vkuznet:
Alan, thanks for benchmarking. I need to admit that I expected worth performance numbers. The DBS call time is unavoidable and we should work closely with Yuyi to optimize this part. The memory footprint is large, but I already mentioned that you may reduce it by not storing block names strings and use only block UIDs and then have a dataset-block mapping to get block names. Therefore I'm ok to merge this as is for the first implementation and we already discuss other ways to improve the codebase, e.g. thread-pool.

amaltaro:
Thanks, Valentin. I'll get it merged as soon as jenkins get back.
We can also start creating unit tests for the Workflow class, but I'll take care of that in another PR not to block you on the MSMonitor development.

PR #9342, opened 2019-08-15
Allow clients to scan JSON object

vkuznet:
Fixes issue(s) with memory consumption in #9330

#### Description
Changing JSON representation of output to allow clients to scan JSON documents instead of loading them via `json.load` function. This technique will be extremely useful when dealing with large JSON objects, e.g. output from DBS, for instance, blocks API. The client code can read the output line by line and extract records.

YES, but it changes the json output will be modified. Instead of object written as one line, e.g.
```
[{"bla":1}, {"bla":2}]
```
the output will yield records one per line, e.g. the same output but formatted differently
```
[
{"bla":1}
,
{"bla":2}
]
```

#### Related PRs
#9330 

no

### DIFF
--- a/src/python/WMCore/WebTools/RESTFormatter.py
+++ b/src/python/WMCore/WebTools/RESTFormatter.py
@@ -29,15 +29,15 @@ class RESTFormatter(TemplatedPage):
         TemplatedPage.__init__(self, config)
 
     def genstreamer(self, data):
-        yield "["
+        yield "[\n"
         firstItem  = True
         for item in data:
             if not firstItem:
-                yield ","
+                yield "\n,\n"
             else:
                 firstItem = False
             yield json.dumps(item)
-        yield "]"
+        yield "\n]"
 
     def json(self, data):
         if isinstance(data, GeneratorType):


vkuznet:
Finally, I managed to write a program to demonstrate usage of memory when using `json.load` API vs using stream. The program can be found [here](https://gist.github.com/vkuznet/9a82367a759b23bf43bd978a719878a8). What it does is the following:
- it takes JSON data (which we obtain via API call)
- it dumps memory usage using sys.getsizeof and psutil tools of json memory usage
- it creates a string stream out of JSON according to proposed newline solution in this PR
- it dumps memory usage of string stream
- it re-create data from string stream and verifies that created data is the same as loaded via `json.load`
- it dumps memory usage of final JSON object obtained via string stream

Here is a numbers I obtained from this blocks API call, i.e. I downloaded json data from this URL and run the prorgram:
https://cmsweb.cern.ch/dbs/prod/global/DBSReader/blocks?dataset=/JetHT/Run2012C-v1/RAW&detail=true

```
Memory usage for object: json
object  memory: 4280
total   memory: 0 (0.0)
used    memory: 1286144 (1.3MB)
RSS     memory: 1548288 (1.5MB)
VMS     memory: 1273856 (1.3MB)
percent memory: 0.0

Memory usage for object: stream
object  memory: 136
total   memory: 0 (0.0)
used    memory: 278528 (278.5KB)
RSS     memory: 286720 (286.7KB)
VMS     memory: 225280 (225.3KB)
percent memory: 0.0

Memory usage for object: json out of stream
object  memory: 4280
total   memory: 0 (0.0)
used    memory: 749568 (749.6KB)
RSS     memory: 1318912 (1.3MB)
VMS     memory: 786432 (786.4KB)
percent memory: 0.0
```

So, the pure `json.load` takes 1.3MB used, 1.5MB of RSS and 1.3MB of VMS memory. The string stream itself consumes: 278KB, 286KB and 225KB of used, RSS and VMS memory, respectively. And, final json object created out of the stream only consumes: 740KB, 1.3MB and 786 KB of used, RSS adn VMS memory, respectively. Therefore, the benefit of using string stream approach is almost 2x better in terms of memory consumption then pure `json.load` for large object. Please note that data I used only had 470 block records. For larger data we'll have more pronounced savings in memory usage.

You can also see from https://gist.github.com/vkuznet/9a82367a759b23bf43bd978a719878a8#file-mem_test_blocks-py-L73 how easy to parse string stream to create JSON objects. The benefit comes that memory allocated only for a single JSON record within parsed stream and not for entire list of records.

amaltaro:
I didn't manage to look at the test script yet, but since these changes look Okay and it's backwards compatible, let's merge it and follow up on what's necessary.

amaltaro:
Valentin, I had a look at your script and made further tests/changes to it, you can see it in:
https://github.com/amaltaro/ProductionTools/blob/master/memStreaming.py

changes are basically:
* final json object size has to be the same between the single loaded and the streamed loaded (otherwise it doesn't make sense)
* I used a WMCore/Utils library to get the correct size of the object (`getsizeof` doesn't work with nested structure)
* print more meaningful memory statistics, like PSS and USS (however depending on the architecture PSS isn't available).

The input data used has 25145 blocks, from this API call:
https://cmsweb.cern.ch/dbs/prod/global/DBSReader/blocks?dataset=/Neutrino_E-10_gun/RunIISpring15PrePremix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v2-v2/GEN-SIM-DIGI-RAW&detail=true

and here are the numbers I got:
```
Memory usage for object: nothing
object  memory: 16
Total   memory: 0 (0.0)
RSS     memory: 11231232 (11.2MB)
PSS     memory: 9645056 (9.6MB)
USS     memory: 9441280 (9.4MB)

Memory usage for object: json
object  memory: 99456844
Total   memory: 0 (0.0)
RSS     memory: 115167232 (115.2MB)
PSS     memory: 113581056 (113.6MB)
USS     memory: 113377280 (113.4MB)

Memory usage for object: stream
object  memory: 136
Total   memory: 0 (0.0)
RSS     memory: 2428928 (2.4MB)
PSS     memory: 2457600 (2.5MB)
USS     memory: 2457600 (2.5MB)

Memory usage for object: json out of stream
object  memory: 99456844
Total   memory: 0 (0.0)
RSS     memory: 95903744 (95.9MB)
PSS     memory: 96055296 (96.1MB)
USS     memory: 96055296 (96.1MB)
```

so even though the final data object has the same size, somehow the process memory consumption went down by 10-20%. 
So, do I understand it right that the only beneft would be to transfer this data between server and client? Such that json can be loaded in chunks on the client side without waiting for all the data to be available (end of file or so).

vkuznet:
Alan,
I don't have time to look at your code right now, but I'll give you feedback
about memory savings.

The savings are on a client side, not on server (DBS) one.
To understand this you need better understanding of how python memory allocation
works. I'll send you a link later, but in short, python allocates memory
in chunks and allocates it in advance, i.e. more than it is actually required.
And, then it keeps it around "just in case". That's why long running
processes (like services) usually have large memory footprint.

So, in first case in your the PSS is 113MB, while in streaming case it is 96MB.
Even though the savings is not big it will be accumulated over time in a client
especially if it runs as a server (MS case) and need to process records
over long time.

In json parsing case python allocates memory according to object it parses, so
the bigger number of blocks more and more python will need to have at its
disposal. The RAM will grow up to a certain threshold. This is what the client
(in this case MS server) will see. For the first blocks python will allocate
N-MB, if any other object will require more python will add more memory. But
eventually it will stop because it will have enough memory.

In streaming case, python will allocate only chunk of memory needed to parse
single block record in a JSON. Then it will re-use it since all blocks records
will be identical in size. But it will store then in final list of blocks
which means it will need to allocate then accordingly. Therefore the client (in
this case MS server) memory will be gradually increase to a certain threshold.

The difference in two cases is that json parser will jump in memory allocation
and such jump may not be optimal. While in streaming case memory will be
allocated more precise.

I can't say precisely the savings amount and it will require running MS
over long period of time and see how memory will be used. But using streaming
technique it will be much smoother then using json.load (this will become
even more important in multi-threaded environment).

Valentin.

PR #9344, opened 2019-08-16
Make DbsUrl homogeneous in the json template

amaltaro:
#### Status
not-tested

#### Description
While testing the microservices, I noticed we split DBS calls to different DBS instances because some templates add the slash symbol at the end, while others don't. Make them uniform to take the best advantage of the MicroServices.

yes

#### Related PRs
no

no


amaltaro:
test this please

PR #9346, opened 2019-08-19
Accept also WMCore-like campaign json configuration

### LINKED ISSUE #9345: Support input file with WMCore-schema for unified campaigns
**Impact of the new feature**
Just to make our life easier during the MicroServices commissioning, given that we might not have the WMCore/ReqMgrAux setup in our node (or even pymongo).

`parseUnifiedCampaigns.py` script accepts a file as input, but that file is supposed to be the campaign.json configuration maintained by the Ops team, which means campaigns don't follow the schema implemented for WMCore.
 
So my suggestion is to add another option, such that we can pass in an input json file (result of this same script when we want to dump the output to a file), be able to move the file around, and properly inject campaigns (and a few testing campaigns as well).

**Additional context**

amaltaro:
Fixes #9345

#### Status
ready

#### Description
Support another campaign file as input, but this one already compliant to the WMCore schema.
Once you have the final json with the WMCore campaign configurations (e.g., output,.json), you can 
```
source /data/srv/current/sw/slc7_amd64_gcc630/cms/reqmgr2/1.2.5.pre7/etc/profile.d/init.sh
python parseUnifiedCampaigns.py --cin=output.json --url=https://alancc7-cloud2.cern.ch/reqmgr2 --verbose=10
```

yes

#### Related PRs
Yes, complement to #9306 

no

### DIFF
--- a/bin/adhoc-scripts/parseUnifiedCampaigns.py
+++ b/bin/adhoc-scripts/parseUnifiedCampaigns.py
@@ -3,7 +3,7 @@
 # pylint: disable=R0903
 # R0903: too-few-public-methods
 """
-File       : ParseUnifiedCampaigns.py
+File       : parseUnifiedCampaigns.py
 Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
 Description: Parse Unified campaigns json record or read them
 from Unified MongoDB. The yielded campaign records are ready
@@ -32,11 +32,17 @@ where individual records have the following structure:
 }
 ```
 
-Example of execution:
+***** Example of execution *****
 python parseUnifiedCampaigns.py --dburi=mongodb://localhost:27017
        --dbname=unified --dbcoll=campaignsConfiguration --verbose=10 --fout=output.json
+
 or, parsing campaigns.json file
 python parseUnifiedCampaigns.py --fin=/path/WmAgentScripts/campaigns.json
+
+or yet parsing a json file filled of campaign configurations already compliant
+to the WMCore schema:
+python parseUnifiedCampaigns.py --fin=wmcore_campaign.json --url=https://alancc7-cloud2.cern.ch/reqmgr2
+         --verbose=10 --testcamp
 """
 from __future__ import print_function, division
 
@@ -47,18 +53,18 @@ import logging
 import os
 
 # pymongo modules
-from pymongo import MongoClient
+try:
+    from pymongo import MongoClient
+except ImportError:
+    print("pymongo library not found. Faking it")
+    MongoClient = None
 
 # WMCore modules
 try:
     from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
 except ImportError:
     print("WMCore/ReqMgrAux environemnt not found. Faking it")
-
-
-    class ReqMgrAux():
-        def __init__(self, url, httpDict, logger):
-            pass
+    ReqMgrAux = None
 
 
 class OptionParser(object):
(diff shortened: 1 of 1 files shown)


bin/adhoc-scripts/parseUnifiedCampaigns.py
```diff
@@ -72,6 +78,9 @@ def __init__(self):
         self.parser.add_argument("--fout", action="store",
                                  dest="fout", default="",
                                  help="Output json file with all the campaigns created")
+        self.parser.add_argument("--cin", action="store",
```
vkuznet: I doubt that you need extra parameter for file input. Instead (see below), I suggest to add parameter which will specify action, e.g. add-test-schemas.

bin/adhoc-scripts/parseUnifiedCampaigns.py
```diff
@@ -251,11 +260,37 @@ def upload(mgr, campRec):
...
+    """
+    Insert a few campaigns configuration used for testing purposes
+    NOTE: the campaign name might change according to the tests you're running.
+    :param mgr: ReqMgrAux object
+    """
+    if not mgr:
+        return
+
+    msg = "\nShould we also insert a few testing campaigns (y/n between quotes): "
+    resp = input(msg)
```
vkuznet: I always against inputs from stdin since they lead to problems during scripting. Plus here the input can't be ignored by script input parameters. My suggestion either provide a flag as input parameter to this function along with adding this flag to argument list, or (Better) get rid of input at all and rely on arguments passed to the script. For instance, 
```
parseUnififiedCampaigns.py --add-test-campaigns
```
this is MUCH clear, more robust to user inputs and does not require input from the user interactively

bin/adhoc-scripts/parseUnifiedCampaigns.py
```diff
@@ -276,14 +311,24 @@ def main():
             key = proxy
         hdict = {'cert': cert, 'key': key, 'pycurl': True}
         mgr = ReqMgrAux(opts.url, hdict, logger=logger)
-    if opts.dburi:
+    if opts.cin:
```
vkuznet: I suggest to rely on another option, e.g. add-test-campaigns here and re-use opts.fin as input file option. So this code can be replaced as
```
if opts.test_campaigns:
```

bin/adhoc-scripts/parseUnifiedCampaigns.py
```diff
@@ -276,14 +311,24 @@ def main():
...
+        # then simply use this json file where campaign configurations are
+        # already in the WMCore format
+        with open(opts.cin) as jo:
```
vkuznet: If you add test_campaigns, then you should reuse opts.fin here. And, don't use 2 letters for code variables. Python style always suggest to use 3 or more. Better use `istream` which is much clear then `jo`.

bin/adhoc-scripts/parseUnifiedCampaigns.py
```diff
@@ -276,14 +311,24 @@ def main():
...
+            data = json.load(jo)
+        print("About to insert %d campaign documents into CouchDB..." % len(data))
+        for rec in data:
+            upload(mgr, rec)
+        insertTestCampaigns(mgr)
+        sys.exit(0)
```
vkuznet: Do you really need sys.exit here? this is inside of if block already and no other action should be executed.

amaltaro:
Thanks for the suggestions, Valentin. I think I've got all of them in, if you have a spare time, please have another look. I put in a fragile way of figuring whether input json contains unified or wmcore campaign schema, but anyways we don't plan to use this for too long, so that should suffice.

vkuznet (approved):

PR #9340, opened 2019-08-13
Add services into configuration to separate MSTransferor/MSMonitor

vkuznet:
Fixes #9339 

#### Status
tested with unit test

#### Description
Add 3 new configuration parameters:
- services should specify list of services to run, e.g. 'transferor' or 'transferor,monitor' to run both services, if not specified we'll run both services
- transferorInterval, new value to specify transferor polling interval
- monitorInterval, new value to specify monitor polling interval

I make it backward compatible with single interval option which is assigned to both intervals if not specified.

I also make it safe to start transferor/monitor execute and threads since they should only be started if appropriate objects are initialized.

#### Is it backward compatible (if not, which system it affects?)
YES

#### Related PRs
This PR complements discussion of #9330 

#### External dependencies / deployment changes
no

PR #9349, opened 2019-08-21
Support MSTransferor/MSMonitor separation into different processes

### LINKED ISSUE #9339: Separate MSTransferor and MSMonitor in different process
**Impact of the new feature**
ReqMgr2MS

Once ReqMgr2MS hit production, it will be under heavy load and we better allocate a different CPU for each of those threads.
With the thread/service separation, we also need to take into consideration the log file that each of them is going to write to.

Start two different process, one for MSTransferor and one for MSMonitor.

**Additional context**
Changes will also be required to the deployment repository.

amaltaro:
Fixes #9339 
Superseeds #9340 

#### Status
not-tested

#### Description
Only start the thread defined in the deployment configuration. If nothing is defined, don't run anything.

in terms of workload, yes.

#### Related PRs
Yes, cherry-picked Valentin's changes on #9340 

Yes, deployment changes coming up soon...

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSManager.py
+++ b/src/python/WMCore/MicroService/Unified/MSManager.py
@@ -56,37 +56,35 @@ class MSManager(object):
         :param config: reqmgr2ms service configuration
         :param logger:
         """
-        self.uConfig = {}
         self.config = config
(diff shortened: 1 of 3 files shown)


amaltaro:
This is also good to go. Documentation needs to be updated with the new REST URI

PR #9356, opened 2019-08-27
Update json templates with IncludeParents True; fix Harvesting on StepChain

amaltaro:
#### Status
not-tested

#### Description
Update the StepChain template such that the file upload to the DQMGui succeeds.
Also updated all the templates processing the parent dataset - since the current one has been DELETED - so they now use the same input dataset, release and configuration.

yes

no


amaltaro:
Works fine.

PR #9358, opened 2019-08-27
Create Harvest task with the same CMSSWVersion as the DQMMerge task

### LINKED ISSUE #9357: Harvest task using top level CMSSWVersion in StepChain
**Impact of the bug**
ReqMgr2

**Describe the bug**
After fixing the StepChain harvest template in:
https://github.com/dmwm/WMCore/pull/9356

I noticed that the Harvesting job ran with the first CMSSWVersion value in the spec (the top level one), which might not be compatible with the CMSSWVersion used to produce the DQM/DQMIO dataset.

**How to reproduce it**
Injecting the Integration/SC_ReDigi_Harvest.json in the PR#9356 and looking into the logs for the Harvesting job, where one can see in `wmagentJob.log` file this line:
```
2019-08-26 16:23:00,528:INFO:Scram:    Invoking command: /cvmfs/cms.cern.ch/slc7_amd64_gcc700/cms/cmssw/CMSSW_10_6_1/external/slc7_amd64_gcc700/bin/python2.7 -m WMCore.WMRuntime.ScriptInvoke WMTaskSpace.cmsRun1 SetupCMSSWPset
```

**Expected behavior**
Ideally the StepChain template should know which Step is responsible for producing the DQM/DQMIO datasets, and then create the Harvest task using that same CMSSWVersion.
A second approach - that should always work as well, even though not ideal - would be to re-use the same CMSSWVersion as the one used by the last Step in a StepChain workflow.
  
amaltaro:
Fixes #9357 

#### Status
not-tested

#### Description
Pass the correct CMSSWVersion/ScramArch to the Harvest task (which will use exactly the same one used for merging the DQM/DQMIO files).
BTW, this issue wasn't specific to StepChain only.

no, only applies to newly created workflows

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -717,7 +717,8 @@ class StdBase(object):
                                    uploadProxy=self.dqmUploadProxy,
                                    periodic_harvest_interval=self.periodicHarvestInterval,
                                    doLogCollect=doLogCollect,
-                                   dqmHarvestUnit=self.dqmHarvestUnit)
+                                   dqmHarvestUnit=self.dqmHarvestUnit,
+                                   cmsswVersion=cmsswVersion, scramArch=scramArch)
 
         # only in the very end, in order to get it in for the children tasks as well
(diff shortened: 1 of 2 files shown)


amaltaro:
Works fine!

PR #9360, opened 2019-08-28
Enable pycurl by default for all Service client APIs

### LINKED ISSUE #9359: Switch from httplib to pycurl in order to support the K8s setup
**Impact of the new feature**

**Additional context**

amaltaro:
Fixes #9359 
Superseeds https://github.com/dmwm/WMCore/pull/6360

#### Status
tested

#### Description
As the subject says, use pycurl library instead of httplib in any of the WMCore client Service APIs, including those interfacing with Couch, that have been previously missed.

Additional important changes are:
* integrated the `pycurl` module with the `Requests` one, such that the proper headers and encoding/decoding can be applied from the sub classes.
* consolidate HTTP requests setup for both httplib and pycurl library usage
* support compression algorithms `Accept-Encoding` in the pycurl module
* changed the signature of a few methods in the `pycurl` module

I have also executed some performance/concurrency tests using the `hey` tool and the pycurl version seems to be around 50% faster (GET requests). 
Workflow creation and assignment (POST requests) seems to be almost 50% slower though. But for this case we can't isolate WMCore from the other services (DBS/PhEDEx), so there might be some side effect there...

yes

#### Related PRs
First attempt made 4 years ago, here: https://github.com/dmwm/WMCore/pull/6360
It now includes changes proposed in this PR: Now it includes https://github.com/dmwm/WMCore/pull/9386 

no

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -46,7 +46,11 @@ import os
 import re
 import subprocess
 import sys
-import urllib
+try:
+    from urllib import urlencode
+except ImportError:
+    # PY3
+    from urllib.parse import urlencode
 
 # python3
 if sys.version.startswith('3.'):
@@ -109,12 +113,14 @@ class RequestHandler(object):
         self.maxredirs = config.get('maxredirs', defaultOpts['MAXREDIRS'])
         self.logger = logger if logger else logging.getLogger()
 
-    def encode_params(self, params, verb, doseq):
+    def encode_params(self, params, verb, doseq, encode):
         """ Encode request parameters for usage with the 4 verbs.
-            Assume params is alrady encoded if it is a string and
+            Assume params is already encoded if it is a string and
             uses a different encoding depending on the HTTP verb
             (either json.dumps or urllib.urlencode)
         """
+        if not encode:
+            return params
         # data is already encoded, just return it
         if isinstance(params, basestring):
             return params
@@ -122,7 +128,7 @@ class RequestHandler(object):
         # data is not encoded, we need to do that
         if verb in ['GET', 'HEAD']:
             if params:
-                encoded_data = urllib.urlencode(params, doseq=doseq)
+                encoded_data = urlencode(params, doseq=doseq)
             else:
                 return ''
         else:
@@ -135,7 +141,7 @@ class RequestHandler(object):
 
     def set_opts(self, curl, url, params, headers,
                  ckey=None, cert=None, capath=None, verbose=None,
-                 verb='GET', doseq=True, cainfo=None, cookie=None):
+                 verb='GET', doseq=True, encode=False, cainfo=None, cookie=None):
         """Set options for given curl object, params should be a dictionary"""
         if not (isinstance(params, (dict, basestring)) or params is None):
             raise TypeError("pycurl parameters should be passed as dictionary or an (encoded) string")
@@ -144,11 +150,21 @@ class RequestHandler(object):
         curl.setopt(pycurl.CONNECTTIMEOUT, self.connecttimeout)
         curl.setopt(pycurl.FOLLOWLOCATION, self.followlocation)
         curl.setopt(pycurl.MAXREDIRS, self.maxredirs)
+
+        # also accepts encoding/compression algorithms
+        if headers and headers.get("Accept-Encoding"):
+            if isinstance(headers["Accept-Encoding"], basestring):
+                curl.setopt(pycurl.ENCODING, headers["Accept-Encoding"])
+            else:
+                logging.warning("Wrong data type for header 'Accept-Encoding': %s",
+                                type(headers["Accept-Encoding"]))
+
         if cookie and url in cookie:
             curl.setopt(pycurl.COOKIEFILE, cookie[url])
             curl.setopt(pycurl.COOKIEJAR, cookie[url])
 
-        encoded_data = self.encode_params(params, verb, doseq)
+        encoded_data = self.encode_params(params, verb, doseq, encode)
+
 
         if verb == 'GET':
             if encoded_data:
@@ -198,7 +214,7 @@ class RequestHandler(object):
         if cert:
             curl.setopt(pycurl.SSLCERT, cert)
         if verbose:
-            curl.setopt(pycurl.VERBOSE, 1)
+            curl.setopt(pycurl.VERBOSE, True)
             curl.setopt(pycurl.DEBUGFUNCTION, self.debug)
         return bbuf, hbuf
 
@@ -232,11 +248,11 @@ class RequestHandler(object):
 
     def request(self, url, params, headers=None, verb='GET',
                 verbose=0, ckey=None, cert=None, capath=None,
-                doseq=True, decode=False, cainfo=None, cookie=None):
+                doseq=True, encode=False, decode=False, cainfo=None, cookie=None):
         """Fetch data for given set of parameters"""
         curl = pycurl.Curl()
-        bbuf, hbuf = self.set_opts(curl, url, params, headers,
-                                   ckey, cert, capath, verbose, verb, doseq, cainfo, cookie)
+        bbuf, hbuf = self.set_opts(curl, url, params, headers, ckey, cert, capath,
+                                   verbose, verb, doseq, encode, cainfo, cookie)
         curl.perform()
         if verbose:
             print(verb, url, params, headers)
@@ -267,10 +283,12 @@ class RequestHandler(object):
         return header, data
 
     def getdata(self, url, params, headers=None, verb='GET',
-                verbose=0, ckey=None, cert=None, doseq=True, cookie=None):
+                verbose=0, ckey=None, cert=None, doseq=True,
+                encode=False, decode=False, cookie=None):
         """Fetch data for given set of parameters"""
         _, data = self.request(url=url, params=params, headers=headers, verb=verb,
-                               verbose=verbose, ckey=ckey, cert=cert, doseq=doseq, cookie=cookie)
+                               verbose=verbose, ckey=ckey, cert=cert, doseq=doseq,
+                               encode=encode, decode=decode, cookie=cookie)
         return data
 
     def getheader(self, url, params, headers=None, verb='GET',
@@ -287,7 +305,8 @@ class RequestHandler(object):
         for params in parray:
             curl = pycurl.Curl()
             bbuf, hbuf = \
-                self.set_opts(curl, url, params, headers, ckey, cert, verbose, cookie=cookie)
+                self.set_opts(curl, url, params, headers, ckey=ckey, cert=cert,
+                              verbose=verbose, cookie=cookie)
             multi.add_handle(curl)
             while True:
                 ret, num_handles = multi.perform()
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -46,6 +46,7 @@ import os
 import re
 import subprocess
 import sys
+
 try:
     from urllib import urlencode
 except ImportError:
--- a/test/python/WMCore_t/MicroService_t/MicroService_t.py
+++ b/test/python/WMCore_t/MicroService_t/MicroService_t.py
@@ -7,10 +7,11 @@ Author: Valentin Kuznetsov <vkuznet [AT] gmail [DOT] com>
 from __future__ import division, print_function
 
 import json
-import cherrypy
 import unittest
 
+import cherrypy
 from WMCore_t.MicroService_t import TestConfig
+
 from WMCore.MicroService.Service.RestApiHub import RestApiHub
 from WMCore.MicroService.Unified.Common import cert, ckey
 from WMCore.Services.pycurl_manager import RequestHandler
@@ -20,10 +21,11 @@ class ServiceManager(object):
     """
     Initialize ServiceManager class
     """
+
     def __init__(self, config=None):
         self.config = config
         self.state = None
-        self.appname = 'test' # keep it since it is used by XMLFormat(self.app.appname))
+        self.appname = 'test'  # keep it since it is used by XMLFormat(self.app.appname))
 
     def status(self, **kwargs):
         "Return current status about our service"
@@ -36,8 +38,10 @@ class ServiceManager(object):
         self.state = kwargs.get('state', None)
         return {'state': self.state}
 
+
 class MicroServiceTest(unittest.TestCase):
     "Unit test for MicroService module"
+
     def setUp(self):
         "Setup MicroService for testing"
         self.app = ServiceManager()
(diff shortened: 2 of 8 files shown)


amaltaro:
Yes, Valentin. Unless we have other modules making http calls which do not rely on any of the Services client.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
I'll keep this one open for reference when we implement the proper integration with WMCore/Services super class.

amaltaro:
@vkuznet given that this PR touches the pycurl module, you might want to have a look at it.
@belforte Stefano, AFAIK all CRAB* tools rely on the pycurl module, so you might as well want to have a look and/or get something tested with this PR in.

Most of my tests are now done, but I still have to run a final test with all the pylint changes in as well.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -144,11 +151,17 @@ def set_opts(self, curl, url, params, headers,
         curl.setopt(pycurl.CONNECTTIMEOUT, self.connecttimeout)
         curl.setopt(pycurl.FOLLOWLOCATION, self.followlocation)
         curl.setopt(pycurl.MAXREDIRS, self.maxredirs)
+
+        # also accepts encoding/compression algorithms
+        if headers and headers.get("Accept-Encoding"):
+            curl.setopt(pycurl.ENCODING, headers["Accept-Encoding"])
```
vkuznet: this may fail if headers data-type will not be properly set. I suggest get extract headers, verify proper data-type and then call setopt.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -198,8 +211,9 @@ def set_opts(self, curl, url, params, headers,
         if cert:
             curl.setopt(pycurl.SSLCERT, cert)
         if verbose:
-            curl.setopt(pycurl.VERBOSE, 1)
+            curl.setopt(pycurl.VERBOSE, True)
             curl.setopt(pycurl.DEBUGFUNCTION, self.debug)
+        curl.setopt(pycurl.VERBOSE, True)
```
vkuznet: I'm not sure you want VERBOSE level all the time, that why it was under `if verbose` block.

test/python/WMCore_t/MicroService_t/MicroService_t.py
```diff
@@ -7,10 +7,11 @@
 from __future__ import division, print_function
 
 import json
-import cherrypy
 import unittest
 
+import cherrypy
```
vkuznet: I'm not sure why do you need cherrypy here since all applied changes do not use cherrypy.

vkuznet (commented):
Alan, since you put lots of stuff into this PR it is harder to review. I only checked pycurl_manager and added my comments.

src/python/WMCore/Services/pycurl_manager.py
amaltaro: Haaa, I was looking for where the verbosity was coming from. Thanks, it looks like this line sneaked in.

src/python/WMCore/Services/pycurl_manager.py
amaltaro: Did you mean the "Content-Type" header? There is no such "data-type".
However, honestly speaking I simply used the same header logic that we have been using for httplib requests, in the super classes Requests/Service.

test/python/WMCore_t/MicroService_t/MicroService_t.py
amaltaro: Don't spend time looking at test/* changes (and/or pylint changes). As you can see, this is just an automatic reformat of this module.

src/python/WMCore/Services/pycurl_manager.py
vkuznet: So, ask yourself a question what is a data type of `headers["Accept-Encoding"]` Is it an integer, dict, list? You're passing it to setopt and if data-type of return value will match the expected one you'll get an error here. My suggestion is to do this:
```
value = headers.get("Accept-Encoding" ,"") # here you should put proper default (data-type value), I expect it should be string
if isinstance(value, str):
   curl.setopt(pycurl.ENCODING, value)
```

amaltaro: Gotcha! I thought you were speaking about the http request itself and its composition.
Yes, input data sanitization can be done. I'll update it.

amaltaro:
Tested both the httplib and pycurl lib usage, for WMAgents and central services and things are working fine. I've also put some performance notes in my first comment.

I'm going to merge it, cut a new tag and ask Valentin to push it to the kubernetes cluster.

belforte:
@amaltaro is there now a wmcore tag with this in as per your Oct 2 comment ? 1.2.7 maybe ?

amaltaro:
Yep! And if you want our latest production release, please use `1.2.8.pre3` (or wait for a few hours and we should cut a `1.2.8` by the end of the day).

belforte:
good. testing 1.2.7 in k8s cmsweb-test cliuster now. Will build again with 1.2.8 when this has been used a bit moire, so far, so good, services run and one submission went through.

belforte:
Sorry @amaltaro but at first sight this is not backward compatible and TaskWorker fails.

Somehow the request to encode parameters from dict. form like `{'getstatus': 'QUEUED', 'limit': 6, 'workername': 'stefanovm'}` to urlencoded string like `getstatus=QUEUED&limit=6&workername=stefanovm` was default=True before, but it is default=False now. Or at least my params above are not encoded and I get an exception about concatenating dict and string at this line:
https://github.com/dmwm/WMCore/blob/2ec701cd7785545663461ceb536e462deee0ecfa/src/python/WMCore/Services/pycurl_manager.py#L172

So far I know that this happens because the flag encode here is false since it is not passed among arguments to
https://github.com/dmwm/WMCore/blob/2ec701cd7785545663461ceb536e462deee0ecfa/src/python/WMCore/Services/pycurl_manager.py#L143-L145

belforte:
looking deeper, the problem was introduced with this commit
https://github.com/dmwm/WMCore/commit/4181fef43d5147f223fceca3e7a777524c0005cd
and the relevant change is in the screenshot below
Screenshot from 2019-11-12 17-24-08
while before `self.encode_params` was "just called", now it is passed the encode flag which defaults to False and which was not existing before.
Indeed same commit also brings
Screenshot from 2019-11-12 17-32-27


I guess I will add all the encode=True whic are needed now and were not before.

Let's hope I can find all relevant placed.

I hate that a bit at a time it seems I have to learn every line of CRAB code.

Why do you keep making non-compatible changes ?

belforte:
OTOH... When I put on my list "make sure that parameters for the URL needs to be URLencoded" I was feeeling a bit silly. Passing a dictionary and asking to encode it...
Especially given that the pycurl methos in WMCore says that it supports both
a dictionary and an urlencodded string !

Could you consider to add in
https://github.com/dmwm/WMCore/blob/2ec701cd7785545663461ceb536e462deee0ecfa/src/python/WMCore/Services/pycurl_manager.py#L143
something like
```
       if isinstance(params, (dict)): encode=True
```

And by the way, there should be no harm in calling `encode_params` anyhow since it is already coded to deal with urlencoded data (by returning them) !

All in all I think I should not change CRAB this time, but wait for WMCore 1.2.9

vkuznet:
Stefano, don't rush with conclusions. Different systems may had different defaults. What was in CRAB by default does not mean it was by default in DBS, etc. Passing dictionary and asking to encode is not a strong argument since it depends on HTTP method, for GET it is valid request, but not for POST where dictionary may be preserved.

I rather want to investigate **all** use cases before introducing a new default. For that I think we should get Alan's opinion because the details here applies to entire WMCore which is shared across different projects, e.g. DBS, ReqMgr2, WMArchive, CRAB, etc.

belforte:
I understand the point, for me it is reverting to the OLD default, not introducing a new one. But maybe other systems had a different default when using httplib and Alan tried to be backward compatible for them.
I am sure you understand my point as well, would be great if WMCore could tell if encode is needed or not and do, what's needed. Please see this as well :
https://github.com/dmwm/WMCore/blob/2ec701cd7785545663461ceb536e462deee0ecfa/src/python/WMCore/Services/pycurl_manager.py#L129-L139

But you know much more then me on this. So I defer. Let me know what to do when you reach the conclusion.

amaltaro:
Apologies for giving you this extra work, Stefano. This problem is actually more complex than it looks like. If one uses the `pycurl_manager` module in an standalone mode, then there was nothing wrong with it.

However, the `WMCore/Services` package follows a hierarchy where the wrapper services APIs (e.g. DBS) is a sub-class of the `Services` module, which then uses either "plain Requests" or JSON requests (and their encoding/decoding defined). That means the sub-class is the one responsible for setting up the whole environment. Inferring those things at the `pycurl` layer used to work for most of the cases, but at least not for the PhEDEx wrapper. 
In addition to that, we had multiple (and sometimes different) encoding/decoding when using Services wrapper modules with PyCURL.

So this PR is basically using the pycurl library as we used to do with httplib, leaving encoding/decoding to the upper layer, see:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L153-L170

amaltaro:
This was a very deep change and hard to make it compatible with every single client of WMCore, but I'm pretty sure we can put more effort in the future and either try not to break compatibility, or be very clear when something is likely going to break.

belforte:
Ciao @amaltaro . Indeed if you had told me about the change, at least I would have not spent a couple hours trying to find where the problem was.

Looking ahead now, I have the usual problem that this was not my code and reverse-engineer all of CRAB is a pain. We were working with a pycurl API which did not have the Encode switch,  so my calls never say if encoding will be needed or not, now I have to add it. I worry about doing it right ! 
Can you guys help ? Should I assume as @vkuznet indicated that params need encoding for GET but not for POST ? Or ..what ?

belforte:
I created an ad-hoc issue in CRAB GIT. Let's follow up in there about what to do.
https://github.com/dmwm/CRABServer/issues/5984
The action requested in this (test new pycurl in CRAB) was done.
Now have to make it work

vkuznet:
Well, let's break down series of calls:
- a client invoke WMCore request method which is defined here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L153
- the `makeRequest` calls `self.makeRequest_pycurl(uri, data, verb, headers)` defined here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L165
- the `makeRequest_pycurl` calls `self.reqmgr.request(uri, data, headers, verb=verb,ckey=ckey, cert=cert, capath=capath)` defined here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L182
- the `self.reqmgr.request` is defined in pycurl_manager here
https://github.com/dmwm/WMCore/blob/2ec701cd7785545663461ceb536e462deee0ecfa/src/python/WMCore/Services/pycurl_manager.py#L250
and it has default parameters `doseq=True, encode=False, decode=False` which are passed to `set_opts`. So the `set_opts` receives by default `encode=False` which is not passed in WMCore/Services/Requests.py at all.

Therefore it is a flaw in WMCore/Services/Requests.py code which does not provide an optional parameter to allow encoding or not input data. Instead, the WMCore/Services/Requests.py relies on `encoder=True` option which performs encoding of parameters internally and pass it to pycurl. It is done here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Requests.py#L160
Therefore pycurl_manager always take intact url and data passed to it.

Since I don't know how CRAB uses WMCore Requests I can't comment from here. But it seems to me that WMCore Requests either should provide an optional parameter `encode` which is passed to pycurl manager or CRAB should use `encoder=True` in its code which by the way is set to True by default.

Therefore it is up to CRAB to decide either use WMCore/Service/Requests.py which provide proper default or bypass it use pycurl_manager directly. If we don't want to change pycurl_manager default parameters then I suggest Stefano should write its own wrapper where it calls pycurl_manager with whatever parameters he wants and then use his wrapper in CRAB code.

belforte:
Thanks @vkuznet , give me a bit do digest that !  but the bottom line is that currebtly there is NO place in CRAB where code indicates it encoding should be used or not, it was done inside WMCore/pycurl_manager.py and worked. Now whether I call that directly or via Services/Requests is immaterial, since I still need to put code on my end to decide if to encode or not.  I can sort out the calling stack, what bother me is to be sure that I do the correct HTTP calls, since I have no experience and am entering the age where learning new stuff gets harder and harder.

amaltaro:
I also don't know whether CRAB is using pycurl from Requests or in an standalone way.

Regarding Valentin's last comment, almost everything is correct except by 
```
Therefore it is a flaw in WMCore/Services/Requests.py code which does not provide an optional parameter to allow encoding or not input data
```

That's exactly  why we have `encoder` and `decoder` for, where one can either:
* leave the default `encoder=True`: which means, use the default encoder function as defined by my subclass (DBS/PhEDEx/etc)
* set `encoder=False`: which means, my data has been already encoded or don't need anything else
* set it to whatever encoding function you want (urlencode, json.loads, etc): so `encoder` is not a boolean, but actually a function to be applied to your input data.

Hope it clarifies the code.

belforte:
I'd really like to discuss the CRAB side in the CRAB issue:
https://github.com/dmwm/CRABServer/issues/5984

where I already sumamrized what CRAB does.
My question is very simple, and it is:
**WHEN SHOULD I SET ENCODE=TRUE AND WHEN ENCODE=FALSE ?**
Given that this is a new argument in the call to pycurl_manager which we did not have before !
Yeah.. I know.. I have to dig the old version of pycurl_manager from git, figure out what exactly was doing, and reproduce it in my code. I would have loved if such thing had been done by you culminating in a nice sujmmary of differences and/or a "how to use".... but yeah... I know.. we still still be good friends.

vkuznet:
And answer was said:
- you use encode=True if you want your parameters to be encoded and encode=False otherwise.

The encoded parameters is usually used in GET calls, while POST calls usually do not perform parameters encoding. I say usually here since it depends on use cases, e.g.
- in GET calls you need to encode when a server accept positional argument, but for REST interfaces you do not encode parameters, but often construct a proper path
- in POST calls you do not encode since your server may expect paramaters in specific format, e.g. JSON, but it is not mandatory either and there are servers which expect encoded parameters in POST calls

Therefore the full answer is: it depends on your concrete use case, i.e. on server which expect your call. What CRAB use I have no idea, to which services it talks I have no idea. Therefore I can't say/tell you what should be your defaults. But I understand that chosen defaults now may not coincide with defaults used before. That's why it should be application specific and explicit usage of arguments should be always enforced to avoid potential problems with default choices.

belforte:
Sure. And I totally and fully agree. And I DO NOT KNOW EITHER !!! I am not the person who wrote CRAB. I am only trying to keep it running. The goal here and now is simply to make it work like before, good or bad that it was. I thank you for not having given up in trying to educate me on how to write code, but would rather not rewrite CRAB now.

> Therefore the full answer is: it depends on your concrete use case, i.e. on server which expect your call. What CRAB use I have no idea, to which services it talks I have no idea. Therefore I can't say/tell you what should be your defaults. But I understand that chosen defaults now may not coincide with defaults used before. That's why it should be application specific and explicit usage of arguments should be always enforced to avoid potential problems with default choices.

PR #9365, opened 2019-09-03
Use pycurl library by default; urlencode PhEDEx body request

### LINKED ISSUE #9359: Switch from httplib to pycurl in order to support the K8s setup

amaltaro:
Fixes #9359
Superseeds #9360
Superseeds https://github.com/dmwm/WMCore/pull/6360

#### Status
ready

#### Description
* for any sub-class of `WMCore/Services/Requests` (or `WMCore/Services/Services`), use by default the `pycurl` library instead of httplib.
  * don't encode the body data again within the pycurl module
* set the `Content-Length` HTTP header for PUT/POST requests within the pycurl module
* given that the PhEDEx service class relies on `Requests` instead of `JSONRequests`, use `urlencode` to encode the body data (if any)

The pycurl module is still missing a better integration with the Services/Requests super class though, as proposed in this PR: https://github.com/dmwm/WMCore/pull/9360

yes

#### Related PRs
Old and initial proposed fix in: https://github.com/dmwm/WMCore/pull/6360

no

### DIFF
--- a/src/python/WMCore/Services/PhEDEx/PhEDEx.py
+++ b/src/python/WMCore/Services/PhEDEx/PhEDEx.py
@@ -1,7 +1,11 @@
 import json
 import logging
+try:
+    from urllib import urlencode
+except ImportError:
+    # PY3
+    from urllib.parse import urlencode
 from xml.dom.minidom import parseString
-
 from WMCore.Services.PhEDEx import XMLDrop
 from WMCore.Services.Service import Service
 
@@ -47,6 +51,8 @@ class PhEDEx(Service):
         if clearCache:
             self.clearCache(ifile, args, verb=verb)
 
+        if args:
+            args = urlencode(args, doseq=True)
         fobj = self.refreshCache(ifile, callname, args, verb=verb)
         result = fobj.read()
         fobj.close()
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -74,7 +74,7 @@ class Requests(dict):
         if not idict:
             idict = {}
         dict.__init__(self, idict)
-        self.pycurl = idict.get('pycurl', None)
+        self.pycurl = idict.get('pycurl', True)
         self.capath = idict.get('capath', None)
         if self.pycurl:
             self.reqmgr = RequestHandler()
@@ -150,15 +150,14 @@ class Requests(dict):
         Wrapper around request helper functions.
         """
         if self.pycurl:
-            result = self.makeRequest_pycurl(uri, data, verb, incoming_headers,
-                                             encoder, decoder, contentType)
+            result = self.makeRequest_pycurl(uri, data, verb, incoming_headers, contentType)
         else:
             result = self.makeRequest_httplib(uri, data, verb, incoming_headers,
                                               encoder, decoder, contentType)
         return result
 
     def makeRequest_pycurl(self, uri=None, params={}, verb='GET',
-                           incoming_headers={}, encoder=True, decoder=True, contentType=None):
+                           incoming_headers={}, contentType=None):
         """
         Make HTTP(s) request via pycurl library. Stay complaint with
         makeRequest_httplib method.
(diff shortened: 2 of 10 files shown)


amaltaro:
test this please

amaltaro:
Now that I sorted the unit tests, it's time to test it over the WMCore services for real.

src/python/WMCore/Services/PhEDEx/PhEDEx.py
```diff
@@ -32,7 +36,7 @@ def __init__(self, httpDict=None, responseType="json", logger=None,
         Service.__init__(self, httpDict)
 
     def _getResult(self, callname, clearCache=False,
-                   args=None, verb="POST"):
+                   args='', verb="POST"):
```
vkuznet: Since args (if provided) is a list data type, you should not change its data-type here. The None is correct default for that.

src/python/WMCore/Services/Requests.py
```diff
@@ -301,7 +299,7 @@ def encode(self, data):
         """
         encode data into some appropriate format, for now make it a string...
         """
-        return urllib.urlencode(data, doseq=1)
+        return urllib.urlencode(data, doseq=True)
```
vkuznet: If you look at urllib.urlencode documentation you'll see the following:
```
Help on function urlencode in module urllib:

urlencode(query, doseq=0)
    Encode a sequence of two-element tuples or dictionary into a URL query string.
```
So doseq is defined as integer, so don't change types when it is not required.

test/python/WMCore_t/Services_t/CRIC_t/CRIC_t.py
```diff
@@ -35,13 +35,13 @@ def testConfig(self):
         self.assertEqual(self.myCRIC['cacheduration'], 1)
         self.assertEqual(self.myCRIC['accept_type'], 'application/json')
         self.assertEqual(self.myCRIC['content_type'], 'application/json')
-        self.assertEqual(self.myCRIC['requests'].pycurl, None)
+        self.assertEqual(self.myCRIC['requests'].pycurl, True)
 
-        newParams = {"cacheduration": 100, "pycurl": True}
+        newParams = {"cacheduration": 100, "pycurl": False}
```
vkuznet: I'm not sure why you changed here from True to False, just double check the logic.

test/python/WMCore_t/Services_t/Requests_t.py
```diff
@@ -133,7 +133,7 @@ def testRecoveryFromConnRefused(self):
         """Connections succeed after server down"""
         import socket
         self.rt.stop()
-        req = Requests.Requests(self.urlbase, {'req_cache_path': self.cache_path})
+        req = Requests.Requests(self.urlbase, {'req_cache_path': self.cache_path, 'pycurl': False})
```
vkuznet: again, I'm not sure if False is appropriate default here since you're moving towards using pycurl everywhere.

vkuznet (commented):
Alan, I put only minor comments, the logic itself seems good to me.

src/python/WMCore/Services/PhEDEx/PhEDEx.py
amaltaro: If you look for the encoded_data downstream (at Requests or pycurl_manager level), you'll see that the default value is an empty string. But switching it back to None won't harm either...

src/python/WMCore/Services/Requests.py
amaltaro: It's actually a boolean in python3. In python2, it's not clear, they simply cite "if the optional parameter doseq is evaluates to True". So it's actually safer and forward compatible if we set it to a boolean.

test/python/WMCore_t/Services_t/CRIC_t/CRIC_t.py
amaltaro: It's just a test. The default config has it `True`, so I test a non-default config where it's False

test/python/WMCore_t/Services_t/Requests_t.py
amaltaro: Usually you shouldn't bother reviewing the test changes. But I can answer this one too, the test below (testSecureOddPort_with_pycurl) tests with pycurl library, while this one is meant to test httplib.

amaltaro:
Tests look good on central services and on the agent. Merging

PR #9376, opened 2019-09-13
Make data location update task more verbose

### LINKED ISSUE #9375: Stale GQE sitting in Available
**Impact of the bug**
Global Workqueue

**Describe the bug**
As briefly discussed over slack, the `stuckElementsInfo` CouchDB view was/is reporting an incorrect pileup location for some elements. For instance:
https://cmsweb.cern.ch/couchdb/_utils/document.html?workqueue/014fad4a30799c90ae31edb5f9c00139

which is available at FNAL, but the element reports the PU dataset to be available at DESY + CERN. 

Last element update time dates back to May 10 as well:
```
"updatetime": 1557510816.722089,
```

pointing to a problem with the global workqueue `locationUpdateTask` cherrypy thread.

This is the code responsible for updating elements with an up-to-date list of PNNs:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/DataLocationMapper.py#L192

**How to reproduce it**

**Expected behavior**
It's expected that GQEs sitting in Available status go through a full location resync for all possible input data: primary, secondary and parent dataset. Full resync is quite heavy and has to happen only a few times per day (current setting seems to be every 2h).

amaltaro:
Fixes #9375 

#### Status
ready

#### Description
Honestly, I'm not sure it fixes the problem reported in the GH issue, but what goes in here is:
* fixes data location update for multiple parent blocks within the same GQE
* fixes data location update for multiple pileup datasets within the same GQE
* remove `fullResync` flag; full resync will be the standard operation mode now (with a polling cycle of 6 hours though)
* remove `datasetSearch` flag (given that we automatically find out whether it's a dataset or block)
* last but not least, make things slightly more verbose

yes

#### Related PRs
none

Yes, deployment changes here: https://github.com/dmwm/deployment/pull/807

### DIFF
--- a/src/python/WMCore/GlobalWorkQueue/CherryPyThreads/LocationUpdateTask.py
+++ b/src/python/WMCore/GlobalWorkQueue/CherryPyThreads/LocationUpdateTask.py
@@ -1,5 +1,5 @@
 from __future__ import (division, print_function)
-
+from time import time
 from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
 from WMCore.WorkQueue.WorkQueue import globalQueue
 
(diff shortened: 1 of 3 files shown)


amaltaro:
tests are Okay on my VM.

PR #9385, opened 2019-09-23
Workaround for bug in HTCondor related to customized job attributes when late materialization is enabled

### LINKED ISSUE #9384: +Attributes break submission when late materialization is enabled
Using +Attribute, as done [HERE](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L516-L522) works fine, except when late materialization is enabled.

Description of the issue below:
https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=7261

**How to reproduce it**
Adding `max_materialize = N` in the submission method.

Solution: Change "+" with "My." when creating the submission requests.

khurtado:
https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=7261

Using '+Attribute' for customized job attributes when late materialization is
enabled breaks submission.
We need to use 'My.Attribute' instead.

Fixes #9384

#### Status
Tested in the ITB

#### Description
Change of syntax: Replace '+' with 'My.' This is needed in order to workaround a bug when late materialization (not used yet) is enabled

YES

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -513,53 +513,53 @@ class SimpleCondorPlugin(BasePlugin):
             if self.reqStr and "T3_CH_Volunteer" not in job.get('possibleSites'):
                 ad['Requirements'] = self.reqStr
 
-            ad['+x509userproxy'] = classad.quote(self.x509userproxy)
+            ad['My.x509userproxy'] = classad.quote(self.x509userproxy)
             sites = ','.join(sorted(job.get('possibleSites')))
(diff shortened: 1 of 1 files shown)


amaltaro (commented):
Change looks good, however I wonder what's the impact of these changes on the job monitoring? Does it actually change the classads name, or is it just changed internally in condor?

khurtado:
@amaltaro My understanding is that this is just changed internally in condor. The classad name for +A and My.A is always just "A" in the end.

amaltaro:
Thanks for confirming it, Kenyi. I noticed the same behaviour in a test I executed today.

PR #9386, opened 2019-09-24
Enable pycurl by default also for CouchDB related modules

### LINKED ISSUE #9359: Switch from httplib to pycurl in order to support the K8s setup

amaltaro:
Fixes #9359 

#### Status
not-tested

#### Description
There were these two modules that were explicitly setting `pycurl` to False, thus avoid the default value set in the Requests.py module.

yes

#### Related PRs
Complement to https://github.com/dmwm/WMCore/pull/9365

none


vkuznet:
+1

amaltaro:
Superseeded by #9360

PR #9401, opened 2019-10-18
Wrap around Rucio write APIs for data registration

### LINKED ISSUE #9068: Extend Rucio Service class with data injection/subscription API
Once we have basic read Rucio operations like:
* checking whether a dataset exist in Rucio
* data discovery/location in Rucio

we can then start writing the WMCore APIs for registering CMS files/blocks/dataset and creating subscriptions (rules).

Related to: https://its.cern.ch/jira/browse/CMSRUCIO-97
and 
https://its.cern.ch/jira/browse/CMSRUCIO-37

Reference of writing operations:
https://github.com/dmwm/CMSKubernetes/blob/master/docker/rucio-sync/scripts/BlockSyncer.py#L78

amaltaro:
Fixes #9068 

#### Status
In development

#### Description
Will be filled later

yes

#### Related PRs
See further examples and usage in the CMSRucio module developed by the CRAB team: https://github.com/dmwm/CRABServer/pull/5954

none

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -8,11 +8,11 @@ CMS Workload Management system (and migration from PhEDEx).
 from __future__ import division, print_function, absolute_import
 
 import logging
-
+from copy import deepcopy
 from rucio.client import Client
-from rucio.common.exception import AccountNotFound, DataIdentifierNotFound,\
-    AccessDenied
-
+from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied,
+                                    DataIdentifierAlreadyExists, DuplicateContent,
+                                    UnsupportedOperation, FileAlreadyExists, RuleNotFound)
 from WMCore.WMException import WMException
 
 
@@ -28,6 +28,13 @@ class Rucio(object):
     """
     Service class providing additional Rucio functionality on top of the
     Rucio client APIs.
+
+    Just a reminder, we have different naming convention between Rucio
+    and CMS, where:
+     * CMS dataset corresponds to a Rucio container
+     * CMS block corresponds to a Rucio dataset
+     * CMS file corresponds to a Rucio file
+    We will try to use Container -> CMS dataset, block -> CMS block.
     """
 
     def __init__(self, acct, hostUrl=None, authUrl=None, configDict=None):
@@ -42,24 +49,28 @@ class Rucio(object):
         :param configDict: dictionary with extra parameters
         """
         configDict = configDict or {}
-        params = configDict.copy()
-        params.setdefault('ca_cert', None)
-        params.setdefault('auth_type', None)
-        params.setdefault('creds', None)
-        params.setdefault('timeout', 600)
-        params.setdefault('user_agent', 'wmcore-client')
+        self.logger = configDict.pop("logger", logging.getLogger())
 
-        self.logger = params.get("logger", logging.getLogger())
+        self.rucioParams = deepcopy(configDict)
+        self.rucioParams.setdefault('account', acct)
+        self.rucioParams.setdefault('rucio_host', hostUrl)
+        self.rucioParams.setdefault('auth_host', authUrl)
+        self.rucioParams.setdefault('ca_cert', None)
+        self.rucioParams.setdefault('auth_type', None)
+        self.rucioParams.setdefault('creds', None)
+        self.rucioParams.setdefault('timeout', 600)
+        self.rucioParams.setdefault('user_agent', 'wmcore-client')
+
         # yield output compatible with the PhEDEx service class
-        self.phedexCompat = params.get("phedexCompatible", True)
+        self.phedexCompat = self.rucioParams.get("phedexCompatible", True)
 
         msg = "WMCore Rucio initialization with acct: %s, host: %s, auth: %s" % (acct, hostUrl, authUrl)
-        msg += " and these extra parameters: %s" % params
+        msg += " and these extra parameters: %s" % self.rucioParams
         self.logger.info(msg)
         self.cli = Client(rucio_host=hostUrl, auth_host=authUrl, account=acct,
-                          ca_cert=params['ca_cert'], auth_type=params['auth_type'],
-                          creds=params['creds'], timeout=params['timeout'],
-                          user_agent=params['user_agent'])
+                          ca_cert=self.rucioParams['ca_cert'], auth_type=self.rucioParams['auth_type'],
+                          creds=self.rucioParams['creds'], timeout=self.rucioParams['timeout'],
+                          user_agent=self.rucioParams['user_agent'])
         clientParams = {}
         for k in ("host", "auth_host", "auth_type", "account", "user_agent",
                   "ca_cert", "creds", "timeout", "request_retries"):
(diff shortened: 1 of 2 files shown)


amaltaro:
I have done further changes to this patch and managed to get some things tested, it's looking good from my side.

Once I get further rights in Rucio/RSE, I still need to test adding replicas, closing a block and creating a basic rule.

It's likely in a good shape for a code review now, Eric and Todor.

amaltaro:
Only missing bit is now creating a Rucio rule. Eric, would you suggest some specific test (that won't cause any problems) for a rule for:
```
contName = '/RelValProdMinBias/DMWM_Test-ProdMinBias_TaskChain_ProdMinBias_Feb2020_Val_AlanPrivv2-v11/GEN-SIM'
scope = 'user.amaltaro'
```

in the `int` instance.

ericvaandering:
You can put a rule on where you just placed it. Or you can also put a rule to move it to T1_US_FNAL_Disk_Test.

I'll be curious to see what happens with the user.amaltaro scope. Hopefully @dciangot planned for that and our LFN to PFN conversion works for it.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -8,11 +8,10 @@
 from __future__ import division, print_function, absolute_import
 
 import logging
-
+from copy import deepcopy
 from rucio.client import Client
 from rucio.common.exception import AccountNotFound, DataIdentifierNotFound,\
```
ericvaandering: Prettier way to do this is use () around the exception names. Preferred now too.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -229,3 +239,246 @@ def getPFN(self, nodes=None, lfns=None, destination=None, protocol='srmv2', cust
...
+        except DataIdentifierAlreadyExists:
+            self.logger.debug("Block name already exists in Rucio: %s", name)
+            response = True
+        except Exception as ex:
+            self.logger.error("Exception creating block: %s. Error: %s", name, str(ex))
+
+        # then attach this block recently created to its container
+        if response and attach:
+            container = name.split('#')[0]
+            response = self.attachDIDs(kwargs.get('rse'), container, name, scope)
```
ericvaandering: You can get an error here if the block already existed and was already attached to the container (I think).

ericvaandering: Yeah, you catch a similar error in the next section.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -229,3 +239,246 @@ def getPFN(self, nodes=None, lfns=None, destination=None, protocol='srmv2', cust
...
+
+        # TODO: test to make sure 'state' is a valid argument
+        response = False
+        try:
+            # add_replicas(rse, files, ignore_availability=True)
+            response = self.cli.add_replicas(rse, files, ignoreAvailability)
+        except Exception as ex:
+            self.logger.error("Failed to add replicas for: %s and block: %s. Error: %s", files, block, str(ex))
+
+        if response:
```
ericvaandering: I'm not sure the logic here makes sense to try to attach DIDs every time a replica is made. 

In general, the way I'd think about it is 
- tell Rucio which DIDs exist and how they are related
- tell Rucio where the replicas of files are

So attaching files to block is part of the first part.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -229,3 +239,246 @@ def getPFN(self, nodes=None, lfns=None, destination=None, protocol='srmv2', cust
...
+        kwargs.setdefault('account', self.rucioParams.get('account'))
+        kwargs.setdefault('locked', False)
+        kwargs.setdefault('notify', 'N')
+        kwargs.setdefault('purge_replicas', False)
+        kwargs.setdefault('ignore_availability', False)
+        kwargs.setdefault('ask_approval', False)
+        kwargs.setdefault('asynchronous', False)
+        kwargs.setdefault('priority', 3)
+
+        if isinstance(names, basestring):
```
ericvaandering: This is not going to work in python3. Is there another way?

ericvaandering (changes requested):
In general it looks good. My only substantive comment in on the logic of did attachment.

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: will update it

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: Yes, the error handling is done on the `attachDIDs` method

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: Ok, this might be where I get confused.
When we add a new replica (which is a file at a given location), doesn't it meet your second bullet comment?

In short, we add a replica also providing their RSE availability, then we attach those replicas to a block.

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: I will check if it's not a list instead ;)

amaltaro:
Thanks, Eric. I applied changes as you suggested, the only bit still opened is related to the replica logic (I replied to your comment in there).

ericvaandering:
Right, but replicas are not attached to blocks, LFNs are.

Sent from a mobile device.

On Feb 7, 2020, at 11:08, Alan Malta Rodrigues <notifications@github.com> wrote:

﻿

@amaltaro commented on this pull request.

In src/python/WMCore/Services/Rucio/Rucio.py<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_dmwm_WMCore_pull_9401-23discussion-5Fr376504661&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=EHaoB-POFWGrYFvPXoj1bQ&m=QZXWFsYbwULEnLFShHYAiVXcV5V0ZjM4G9rWaadWiCQ&s=BXnOij-q0gD1Rq2OEmBMclMy9AtbTirL2DwDA_iOQFY&e=>:

> +        :return: a boolean to represent whether it succeeded or not
+        """
+        if isinstance(files, dict):
+            files = [files]
+        for item in files:
+            item['scope'] = scope
+
+        # TODO: test to make sure 'state' is a valid argument
+        response = False
+        try:
+            # add_replicas(rse, files, ignore_availability=True)
+            response = self.cli.add_replicas(rse, files, ignoreAvailability)
+            self.logger.error("Failed to add replicas for: %s and block: %s. Error: %s", files, block, str(ex))
+
+        if response:


Ok, this might be where I get confused.
When we add a new replica (which is a file at a given location), doesn't it meet your second bullet comment?

In short, we add a replica also providing their RSE availability, then we attach those replicas to a block.

amaltaro:
Within the WM system, I don't think we will ever have a LFN that does not have a location (besides some non-EDM T0 files). That's why when we create a replica through `createReplica`, we also pass those LFNs to be attached to a block (by the end of the same method).

Are you suggesting to have another method simply to add a new did (a LFN in this case)?

BTW, the more I read this code the more I think I should rename those methods from `createxxx` to `addxxx`. Will do so soon.

amaltaro:
@dciangot Diego, you might want to have a look/review this PR as well. I'm planning to get it in today/tomorrow.

amaltaro:
Now that I managed to test the rule creation (and fixed a problem there), I decided to implement a `getRule` API as well.
Things have been tested and seem to be working fine.

@ericvaandering could you please clarify my previous comment? AFAICS there is nothing else to be fixed, but please let me know if I'm wrong.

ericvaandering:
One could imagine a mode in which you are creating the structure of the files before you are ready to create the replicas. 

It probably doesn't matter on a technical level, it's just a question of how you think about the objects and their relations. And separating file creation and attachment from replica creation is how Rucio thinks about these concepts.

amaltaro:
Yes, it indeed makes sense for the data management tool. 
We might revisit it in the future once all this becomes more mature in the WM context.

Thanks for looking into it, Eric.

amaltaro:
FYI @belforte . It will be in `1.3.1.pre1` wmcore tag.

belforte:
Thanks Alan. @dciangot please take note.

amaltaro:
@todor-ivanov can you please merge this PR and cut a new tag `1.3.1.pre1`? Thanks

todor-ivanov (approved):
good to go

PR #9416, opened 2019-11-12
Fetch quota from Detox; fetch usage from Rucio/PhEDEx; input primary/…

### LINKED ISSUE #9372: Rely on PhEDEx (or something else) for the storage management within MicroServices
**Impact of the new feature**
ReqMgr2 MSTransferor and likely MSMonitor

As part of the Unified logic for input data placement, we need to have a mechanism to reliably monitor:
* how much storage is available at each PNN for the `DataOps` group
* how much data is commited to each PNN

such that it can drive the location selection logic for input data placement (also part of the Unified black box, as requested by issue #9138 )

Create the client API under the `Services` package and integrate it to the MicroServices where needed. Certainly in the MSTransferor algorithm to see which PNNs are eligible to receive data or not.

It could be handled through reviewing the Transferor logic in Unified and integrating whatever hasn't yet...

**Additional context**
See latest changes in this PR: https://github.com/dmwm/WMCore/pull/9330
and also some hints as discussed in this ticket: https://github.com/CMSCompOps/WmAgentScripts/issues/439#issuecomment-522857388

amaltaro:
…parent block placement made by chunks

Fixes #9372 

#### Status
In development

#### Description
There are many changes in this PR, like:
* fetch the PhEDEx group quota from Detox (FRAGILE!!! it needs to go away as soon as possible)
* given a PhEDEx group, fetch the storage usage from PhEDEx; or
  * given a Rucio account, fetch the account usage and quota (limit) from Rucio
* a configurable quota usage has been set to 80%
* PNNs with less than 1TB of available space are skipped from the data placement (should probably be made configurable)
* Workflow API to create chunks of blocks as evenly as possible - size is not fixed and it depends on the number of PNNs to receive data - it also adds the parent blocks to their correspondent chunks (primary + parent)
* fixed logger for the data structures
* retry up to 2 times to update the local MSTransferor caches (required for running the whole cycle)

this logic has also been put in place:
* update the internal cache of storage usage for every single input data placement;
* re-evaluate sites out-of-quota between every single data placement;

what needs further thought is:
* what to do with chunks that are too big for the space/quota available? Do we skip the whole workflow data placement? Do we oversubscribe it? Do we distribute the extra load among other PNNs?

No (it has new features)

#### Related PRs
none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSTransferor.py
+++ b/src/python/WMCore/MicroService/Unified/MSTransferor.py
@@ -4,6 +4,9 @@ Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
              Alan Malta <alan dot malta AT cern dot ch >
 Description: MSTransferor class provide whole logic behind
 the transferor module.
+
+This is NOT a thread-safe module, even though some internal
+tasks might be extended to multi-threading in the future.
 """
 # futures
 from __future__ import division, print_function
@@ -16,11 +19,26 @@ from retry import retry
 
 # WMCore modules
 from Utils.IteratorTools import grouper
-from WMCore.MicroService.Unified.Common import getDetoxQuota
+from WMCore.MicroService.Unified.Common import gigaBytes
 from WMCore.MicroService.Unified.MSCore import MSCore
 from WMCore.MicroService.Unified.RequestInfo import RequestInfo
+from WMCore.MicroService.Unified.RSEQuotas import RSEQuotas
 from WMCore.Services.CRIC.CRIC import CRIC
 from WMCore.Services.PhEDEx.DataStructs.SubscriptionList import PhEDExSubscription
+
+
+def newTransferRec(dataIn):
+    """
+    Create a basic transfer record to be appended to a transfer document
+    :param dataIn: dictionary with information relevant to this transfer doc
+    :return: a transfer record dictionary
+    """
+    transferRecord = {"dataset": dataIn['name'],
+                      "dataType": dataIn['type'],
+                      "transferIDs": set(),  # casted to list before going to couch
+                      "campaignName": dataIn['campaign'],
+                      "completion": [0.0]}
+    return transferRecord
 
 
 class MSTransferor(MSCore):
@@ -36,13 +54,19 @@ class MSTransferor(MSCore):
         """
         super(MSTransferor, self).__init__(msConfig, logger)
         # url for fetching the storage quota
-        self.msConfig.setdefault("detoxUrl",
-                                 "http://t3serv001.mit.edu/~cmsprod/IntelROCCS/Detox/SitesInfo.txt")
-        self.msConfig.setdefault("quotaUsage", 0.8)  # use only 80% of the quota
+        detoxUrl = self.msConfig.get("detoxUrl",
+                                     "http://t3serv001.mit.edu/~cmsprod/IntelROCCS/Detox/SitesInfo.txt")
+        quotaFraction = self.msConfig.get("quotaUsage", 0.8)  # use only 80% of the quota
         # now a PhEDEx group name, then the Rucio account name in the near future
-        self.msConfig.setdefault("quotaAccount", "DataOps")
+        dataAcct = self.msConfig.get("quotaAccount", "DataOps")
+        # minimum available storage to consider a resource good for receiving data
+        minimumThreshold = self.msConfig.get("minimumThreshold", 1 * (1000 ** 4))  # 1TB
 
+        self.rseQuotas = RSEQuotas(detoxUrl, dataAcct, quotaFraction, useRucio=hasattr(self, "rucio"),
+                                   logger=logger, verbose=self.msConfig.get('verbose'),
+                                   minimumThreshold=minimumThreshold)
         self.reqInfo = RequestInfo(msConfig, logger)
+
         self.cric = CRIC(logger=self.logger)
         self.inputMap = {"InputDataset": "primary",
                          "MCPileup": "secondary",
@@ -50,14 +74,10 @@ class MSTransferor(MSCore):
         self.uConfig = {}
         self.campaigns = {}
         self.psn2pnnMap = {}
-        # information about RSE/PNNs and their quota/usage
-        self.nodesUsage = {}
-        # list of RSE/PNNs with less than 1TB available, skipped from data placement
-        self.fullRSEs = set()
         self.dsetCounter = 0
         self.blockCounter = 0
 
-    @retry(tries=2, delay=2, jitter=2)
+    @retry(tries=3, delay=2, jitter=2)
     def updateCaches(self):
         """
         Fetch some data required for the transferor logic, e.g.:
@@ -67,23 +87,27 @@ class MSTransferor(MSCore):
          * all campaign configuration
          * PSN to PNN map from CRIC
         """
-        self._getStorageQuota()
-        self._getStorageUsage()
+        self.logger.info("Updating RSE/PNN quota and usage")
+        self.rseQuotas.fetchStorageQuota()
+        self.rseQuotas.fetchStorageUsage(getattr(self, "rucio", self.phedex))
+        self.rseQuotas.evaluateQuotaExceeded()
+        if not self.rseQuotas.getNodeUsage():
+            raise RuntimeWarning("Failed to fetch storage usage stats")
 
+        self.logger.info("Updating all local caches...")
         self.dsetCounter = 0
         self.blockCounter = 0
-        self.fullRSEs = set()
         self.uConfig = self.unifiedConfig()
         campaigns = self.reqmgrAux.getCampaignConfig("ALL_DOCS")
         self.psn2pnnMap = self.cric.PSNtoPNNMap()
         # also clear the RequestInfo pileup cache
         self.reqInfo.clearPileupCache()
         if not self.uConfig:
-            raise RuntimeWarning("Failed to fetch the unified configuration. Retrying...")
+            raise RuntimeWarning("Failed to fetch the unified configuration")
         elif not campaigns:
-            raise RuntimeWarning("Failed to fetch the campaign configurations. Retrying...")
+            raise RuntimeWarning("Failed to fetch the campaign configurations")
         elif not self.psn2pnnMap:
-            raise RuntimeWarning("Failed to fetch PSN x PNN map from CRIC. Retrying...")
+            raise RuntimeWarning("Failed to fetch PSN x PNN map from CRIC")
         else:
             # let's make campaign look-up easier and more efficient
             self.campaigns = {}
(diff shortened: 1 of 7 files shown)


amaltaro:
@vkuznet @todor-ivanov you guys might want to have a look at it. I still have a few prints here and there that I need to remove, but all the logic seems to be in place.

I still want to run a new - clean - test, if everything goes fine, I'd like to get it merged on Monday to proceed with other issues/developments.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -326,3 +360,105 @@ def createTransferDoc(self, reqName, transferRecords):
         msg += "Error: %s" % resp
         self.logger.error(msg)
         return False
+
+
+    def _getStorageQuota(self):
```
vkuznet: This method is not thread safe since it redefines member data self.nodeUsage all the time. I don't see **ANY** reason why you should keep it this way since you can easily convert to return local copy of nodeUsage dict to upstream caller and I suggest to do this way.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -326,3 +360,105 @@ def createTransferDoc(self, reqName, transferRecords):
...
+            # convert from TB to bytes
+            self.nodesUsage.setdefault(pnn, {})
+            self.nodesUsage[pnn] = dict(quota=int(quota) * (1024 ** 4),
+                                        bytes_limit=0,
+                                        bytes=0,
+                                        bytes_remaining=0,
+                                        quota_avail=0)
+
+
+    def _getStorageUsage(self):
```
vkuznet: This method is not thread safe either. Please convert it to accept nodeUsage dict and return an updated version.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -326,3 +360,105 @@ def createTransferDoc(self, reqName, transferRecords):
...
+                quota = self.nodesUsage[item['name']]['quota']
+                self.nodesUsage[item['name']].update({'bytes_limit': quota,
+                                                      'bytes': item['group'][0]['dest_bytes'],
+                                                      'bytes_remaining': quota - item['group'][0]['dest_bytes']})
+
+        self.logger.info("Summary of the initial quotas:")
+        for site, info in self.nodesUsage.items():
+            self.logger.info("  %s : %s", site, info)
+
+    def _evaluateQuotaExceeded(self):
```
vkuznet: This method is not thread safe. Please pass to it nodeUsage dict and return fullRSes

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -41,38 +50,47 @@ def __init__(self, msConfig, logger=None):
...
         self.psn2pnnMap = {}
+        # information about RSE/PNNs and their quota/usage
+        self.nodesUsage = {}
+        # list of RSE/PNNs with less than 1TB available, skipped from data placement
+        self.fullRSEs = set()
         self.dsetCounter = 0
         self.blockCounter = 0
 
+    @retry(tries=3, delay=2, jitter=2)
     def updateCaches(self):
```
vkuznet: This method is not thread safe, please convert it to make it thread safe. It should get info about caches and return it back to upstream code.

vkuznet (changes requested):
Alan, I had serious concerns about diverging code from thread safe route. You're making code monolithic that it will hurt its maintenance, performance, etc in a long run. Please make an effort and address issues with non thread-safe parts of caches. Keep caches local and use pass data around instead of rely on single instance state.

amaltaro:
This is a valid concern, and as we discussed in the previous PRs you have reviewed, I don't foresee running the whole application in multithreading - but maybe a few tasks within it. 
That said, I usually tend to use object instance variables because I see them much easier to read and maintain than passing many variables in and out of methods over and over.
One or two of your comments indeed make sense to return a reference to a new object, I'll update this PR later.

vkuznet:
Alan, please put **explicit** docstring to the codebase that it is not thread
safe and point out to methods which are not thread safe. I rather prefer to
have explicit notices in codebase to discover this easily then go through
lots of debugging when it will be necessary.

amaltaro:
@vkuznet can you please have another look here?
Changes regarding your first review are:
* a few bug fixes here and there
* and all the quota-related logic moved to its own module

src/python/WMCore/MicroService/Unified/RSEQuotas.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __str__(self):
+        """
+        Write out useful information for this object
+        :return: a stringified dictionary
+        """
+        res = {'detoxUrl': self.detoxUrl, 'dataAcct': self.dataAcct,
+               'useRucio': self.useRucio, 'quotaFraction': self.quotaFraction,
+               'minimumSpace': self.minimumSpace}
+        return str(res)
```
vkuznet: Never use str for getting string representation of the dict. You should always use json.dump. The former will produce improper json and we had so much headache to catch such errors.

src/python/WMCore/MicroService/Unified/RSEQuotas.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        return self.availableRSEs
+
+    def getOutOfSpaceRSEs(self):
+        """
+        Return a list of out-of-space RSE/PNNs
+        """
+        return self.outOfSpaceNodes
+
+    def fetchStorageQuota(self):
```
vkuznet: As I requested, I want explicit **NOT THREAD SAFE** comment in all methods which fall into this category. This will help to identify them easily.

src/python/WMCore/MicroService/Unified/RSEQuotas.py
```diff
@@ -0,0 +1,204 @@
...
+            self.nodeUsage.setdefault(pnn, {})
+            # convert from TB to bytes
+            self.nodeUsage[pnn] = dict(quota=int(quota) * (1000 ** 4),
+                                       bytes_limit=0,
+                                       bytes=0,
+                                       bytes_remaining=0,
+                                       quota_avail=0)
+        self.logger.info("Storage quota filled from Detox information")
+
+    def fetchStorageUsage(self, dataSvcObj):
```
vkuznet: Please add explicit **NOT THREAD SAFE** comment for this method docstring

src/python/WMCore/MicroService/Unified/RSEQuotas.py
```diff
@@ -0,0 +1,204 @@
...
+                quota = self.nodeUsage[item['name']]['quota']
+                self.nodeUsage[item['name']].update({'bytes_limit': quota,
+                                                     'bytes': item['group'][0]['dest_bytes'],
+                                                     'bytes_remaining': quota - item['group'][0]['dest_bytes']})
+
+        self.logger.info("Summary of the initial quotas:")
+        for site, info in self.nodeUsage.items():
+            self.logger.info("  %s : %s", site, info)
+
+    def evaluateQuotaExceeded(self):
```
vkuznet: Please add explicit **NOT THREAD SAFE** comment for this method docstring

src/python/WMCore/MicroService/Unified/RSEQuotas.py
```diff
@@ -0,0 +1,204 @@
...
+            quotaAvail = info['quota'] * self.quotaFraction
+            info['quota_avail'] = min(quotaAvail, info['bytes_remaining'])
+            if info['quota_avail'] < self.minimumSpace:
+                self.outOfSpaceNodes.add(rse)
+            else:
+                self.availableRSEs.add(rse)
+
+        self.logger.info("List of RSE's out of quota: %s", self.outOfSpaceNodes)
+
+    def updateNodeUsage(self, node, dataSize):
```
vkuznet: Please add explicit **NOT THREAD SAFE** comment for this method docstring

src/python/WMCore/MicroService/Unified/RSEQuotas.py
```diff
@@ -0,0 +1,204 @@
...
+It can also communicate with other data management tools, like
+Detox, Rucio and PhEDEx.
+"""
+from __future__ import division, print_function
+
+from WMCore.MicroService.Unified.Common import getDetoxQuota
+from WMCore.MicroService.Unified.Common import getMSLogger
+
+
+class RSEQuotas(object):
```
vkuznet: If I understood logic correctly you update info about specific rse and keep internal dict of them.
Why not to make a class which just keep info about single RSE instead. Then you can take whole logic of nodeUsage out of this class and use it in upper codebase making RSE updates concurrent/thread-safe if you want. The main problem is self.nodeUsage here, if you take it out and keep in another class then all RSE instances can be updated concurrently.

vkuznet (changes requested):

src/python/WMCore/MicroService/Unified/RSEQuotas.py
amaltaro: That's the correct understand of this module, Valentin.
I'm not sure I followed your suggestion, especially regarding rse information lookup and update. In the end we would have to have a set of such objects (one for each RSE), but then we would have to scan every object in order to find the RSE that we want to (and then the update can be made).

Another possibility, as I quickly mentioned in the meeting, would be to add a locking system around the local cache updates. I'd need to educate myself on the best way to make it thread safe (maybe using the threading.Queue feature...). Still, that might have to wait for another iteration on the near future.

src/python/WMCore/MicroService/Unified/RSEQuotas.py
amaltaro: It goes to a plain text log anyways. But it's a valid concern and I will get it updated.

src/python/WMCore/MicroService/Unified/RSEQuotas.py
amaltaro: As mentioned in the meeting, this is mentioned at the very top of the module and I don't plan to write it to every single method not thread safe :)

src/python/WMCore/MicroService/Unified/RSEQuotas.py
amaltaro: Valentin, I'm keeping this method override as is because if I have a set object in there, json.dumps fails to serialize.
And if someone wants to get a json representation of this data, then we should implement the proper method. The `__str__` is mostly a human readable string.

amaltaro:
I had to fix a few other issues and everything looks good now.
Valentin, you requested a few changes to this PR, and I hope I managed to clarify those. If I didn't, please either reply here or even better, make a new GH issue and I can get back to it.

Anyhow, it's time to move on with this. I'm merging it now in order to cut a new release.

PR #9433, opened 2019-11-20
Adding an error message for missing PU dataset.

### LINKED ISSUE #9431: Not catching exceptions with error code 8029 for missing PU file on the Site.
**Impact of the bug**
All the agents. 

**Describe the bug**
We are missing the exit code 8029 in 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMExceptions.py 

which comes from the CMSSW framework as:
```
Fatal Exception (Exit Code: 8029)
An exception of category 'NoSecondaryFiles' occurred while
   [1] Constructing module: class=MixingModule label='mix'
Exception Message:
RootEmbeddedFileSequence no input files specified for secondary input source.
```

So we do not catch any exceptions with that error code and no relevant actions could be taken to recover from it, neither automatic nor manual. This is an exit code falling into the range of CMSSW errors. 
``` 
Range(7000 - 9000)
    cmsRun (CMSSW) exit codes. These codes may depend on specific CMSSW version
    https://github.com/cms-sw/cmssw/blob/CMSSW_5_0_1/FWCore/Utilities/interface/EDMException.h#L26
```
The meaning of this error is that the job landed on a worker node (site) that does not host a single block of the pileup dataset.

A proper message should be added, so that at least the operators can have a better clue for the origin of the error. Error description might be the same as the one from CMSSW:
```RootEmbeddedFileSequence no input files specified for secondary input source.```

**How to reproduce it**
Reproducible with a workflow at a site without the relevant PU dataset. 

**Expected behavior**
Provide the proper Error message.

No

### PR DESCRIPTION
**Pull request by todor-ivanov**
Fixes #9431

#### Status
 not-tested

#### Description
Added an error message for CMSSW exit code 8029, caused by a missing PU dataset at the site. 

YES

#### Related PRs
NO

NO

### CODE REVIEW

File: src/python/WMCore/WMExceptions.py [hunk 1]
```diff
@@ -131,6 +131,7 @@
                       8026: "NotFound (something other than a product or dictionary not found).",
                       8027: "FormatIncompatibility.",
                       8028: "FileOpenError with fallback.",
+                      8029: "RootEmbeddedFileSequence no input files specified for secondary input source.",
```
amaltaro: Thong, how about we update this error message to:
`The job configuration has no input files specified for secondary input source`?

todor-ivanov: Ok, it is changed now.

amaltaro (approved):
It looks good now, Todor. Thanks

PR #9438, opened 2019-11-27
Close PhEDEx blocks in smaller chunks of location and dataset

### LINKED ISSUE #9437: Break closure of phedex blocks into smaller chunks
**Impact of the bug**
WMAgent (PhEDExInjector)

**Describe the bug**
Apparently we put too much data into the same PhEDEx call to close blocks here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L375

it's one call per location, and that location might have many many blocks from multiple datasets.
The problem is, if there is a problem with any of those blocks (e.g., the current state of vocms0253 which has a **dataset** closed in PhEDEx, see [1]), the whole request would fail and no blocks can be closed...

**How to reproduce it**
none

**Expected behavior**
My suggestion is to break the http requests per location and dataset, something like
for location in locations:
  for dataset in datasets:
    make a phedex request

Yes, it will increase the amount of PhEDEx calls, but still it's going to be 1 call per block that we close, which should be just fine for the phedex data-service.

[1]
```
2019-11-27 22:23:08,945:140168429426432:WARNING:Service:The cachefile /data/srv/wmagent/v1.2.6.patch1/install/wmagent/PhEDExInjector/.wmcore_cache/.wmcore_cache_31961/requests/cmsweb.cern.ch:8443/-7685143179613178687_POST_inject does n
ot exist and the service at https://cmsweb.cern.ch:8443/phedex/datasvc/json/prod/inject is unavailable - it returned 400 because Bad Request with result: injectData error: dataset /RadionToWW_narrow_M-6500_TuneCUETP8M1_13TeV-madgraph-p
ythia8/RunIISummer16MiniAODv3-PUMoriond17_94X_mcRun2_asymptotic_v3-v1/MINIAODSIM is closed\n

2019-11-27 22:23:08,946:140168429426432:ERROR:PhEDExInjectorPoller:PhEDEx block close failed with HTTPException: 400 injectData error: dataset /RadionToWW_narrow_M-6500_TuneCUETP8M1_13TeV-madgraph-pythia8/RunIISummer16MiniAODv3-PUMorio
nd17_94X_mcRun2_asymptotic_v3-v1/MINIAODSIM is closed\n
```

amaltaro:
Fixes #9437 

#### Status
not-tested

#### Description
Iterate over the locations and then over each dataset, creating a phedex request (to close blocks) for each tuple of location+dataset. It will of course increase the number of http requests made to PhEDEx, on the other hand, errors will be more isolated.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
+++ b/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
@@ -351,7 +351,7 @@ class PhEDExInjectorPoller(BaseWorkerThread):
 
         migratedBlocks = self.getMigrated.execute()
 
-        for siteName in migratedBlocks.keys():
+        for siteName in migratedBlocks:
             # SE names can be stored in DBSBuffer as that is what is returned in
             # the framework job report.  We'll try to map the SE name to a
             # PhEDEx node name here.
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
```diff
@@ -351,7 +351,7 @@ def closeBlocks(self):
 
         migratedBlocks = self.getMigrated.execute()
 
-        for siteName in migratedBlocks.keys():
```
todor-ivanov: @amaltaro since you are not changing the way how you are initializing && calling the upper level object [1], why do you want to change the way you iterate here. The only reason I could imagine is later down in this same PR [2]. You obviously want to have a new view on the dictionary, but even with that, keeping the explicit reference to migratedBlocks.keys() here should not break the rest of the code. Have you tested this?  

[1] https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py#L140

[2]
https://github.com/dmwm/WMCore/pull/9438/commits/70567a7dc6dd462f970008751f18a1a886dd2ad3#diff-73e9dd1a91792c7023eeec392d2e28b3R375

todor-ivanov (commented):
Besides the one comment pasted inline, all the rest looks good.

src/python/WMComponent/PhEDExInjector/PhEDExInjectorPoller.py
amaltaro: Iterating through the keys of a dictionary can be either accomplished by `for k in dict.keys():` or `for k in dict:`, but the first call returns a list while the second yields the key on the fly. It uses the same syntax in python 3 too, but the result is different.

todor-ivanov: Thanks @amaltaro!

todor-ivanov (approved):
All here  looks fine to me.

amaltaro:
Thanks, Todor. Given that it does not affect central services, I'm merging it now. Code very similar to this one is already running in the agents anyways.

PR #9444, opened 2019-11-28
Keep different datatier caches for different DBS URLs

### LINKED ISSUE #9442: Datatier validation does not use the DbsUrl from the workload config
**Impact of the bug**
ReqMgr2

**Describe the bug**
The request dictionary contains a `DbsUrl` key, which is supposed to be used for the input data (plus parent and pileup) check, and also against any other validation needed against DBS (like the datatiers).
With the current problems in DBS testbed, I noticed that even if we point DbsUrl to the production instance, it still fails the datatier validation because it cannot find existent datatiers in the dbs api response.

The problem seems to be here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Utils/Validation.py#L263

which uses a single cache name, no matter the DbsUrl provided. Thus, once a cache gets created, it will always use the same until it gets expired...

**How to reproduce it**
Sort of using this gist
https://gist.github.com/amaltaro/2c546c9a251d12b2f7bfa3d04eb56e34

**Expected behavior**
We should have a different cache structure for each DBS instance. Such that we can fetch data from the correct cache.
That means we need to name the caches after the dbs url. We can likely hash the dbsurl and append it to the current name.

Follow up of: https://github.com/dmwm/WMCore/pull/9355

amaltaro:
Fixes #9442 

#### Status
not-tested

#### Description
This PR started by simply setting a different cache name for the datatier validation (based on the dbs url). That code gets executed in ReqMgr2 during the workflow assignment.

However, I had to make further changes like:
* removed the `_datatiers` class attribute from DBS3Reader (and its caching logic)
* replaced the DBS3Reader static methods by an standalone function (to get the datatiers)
* make DBS3Reader depend on the non standard `retry` library

yes

#### Related PRs
none

I grepped the CRABServer repository and couldn't find any usage for those 2 static methods that got removed in this PR.

### DIFF
--- a/src/python/WMCore/Cache/GenericDataCache.py
+++ b/src/python/WMCore/Cache/GenericDataCache.py
@@ -83,6 +83,7 @@ class GenericDataCache(object):
         elif not isinstance(memoryCache, MemoryCacheStruct):
             raise CacheWithWrongStructException(cacheName)
         else:
+            logging.info("Creating generic cache named: %s", cacheName)
             GenericDataCache._dataCache[cacheName] = memoryCache
 
     @staticmethod
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -3,8 +3,7 @@ ReqMgr request handling.
 
 """
 from __future__ import print_function
-from retry import retry
-
+from hashlib import md5
 from WMCore.Lexicon import procdataset
 from WMCore.REST.Auth import authz_match
 from WMCore.ReqMgr.Auth import getWritePermission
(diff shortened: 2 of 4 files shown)


amaltaro:
@belforte @lecriste FYI., this PR makes changes to our DBS3Reader wrapper API. You can see what changes in my first comment.
In addition to that, I scanned the CRABServer repo but couldn't find those being used, so hopefully you won't have any surprises here.

amaltaro:
It has been tested in my VM. I'm merging it such that I can hand a new tag for cmsweb-testbed deployment. However, Todor and Kenyi, feel free to make any comments here and if needed I can fix it in the coming days.

lecriste:
Ok @amaltaro, let us know when we can test this in cmsweb-testbed.

belforte:
I looked at the changed and agree that it should be fully transparent for CRAB code. I think we can test as part or routine updating to new WMCore releases.

PR #9453, opened 2019-12-06
Update JobSubmitter for docker support

goughes:
#### Status
Tested

#### Description
We need to modify the location of Unpacker.py for agents running in Docker containers. Since the schedd runs on the host, this script must exist somewhere on the host and be available through bind mounts in the container.

Uses an "isDocker" configuration parameter to differentiate between Docker and the standard VM WMAgent deployment and change the Unpacker.py path. The isDocker variable is updated to True when a container starts via https://github.com/dmwm/CMSKubernetes/blob/master/docker/wmagent/run.sh.

Yes

#### Related PRs
NA

NA

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -100,10 +100,16 @@ class SimpleCondorPlugin(BasePlugin):
 
         self.packageDir = None
 
-        if os.path.exists(os.path.join(getWMBASE(),
-                                       'src/python/WMCore/WMRuntime/Unpacker.py')):
-            self.unpacker = os.path.join(getWMBASE(),
-                                         'src/python/WMCore/WMRuntime/Unpacker.py')
+        # if agent is running in a container, Unpacker.py must come from a directory
+        # on the host so the condor schedd can see it
+        # config.General.workDir should always be bind mounted to the container
+        if getattr(config.Agent, "isDocker", False):
+            unpackerPath = os.path.join(config.General.workDir + "/Docker/WMRuntime/Unpacker.py")
+        else:
+            unpackerPath = os.path.join(getWMBASE(), 'src/python/WMCore/WMRuntime/Unpacker.py')
+
+        if os.path.exists(unpackerPath):
+            self.unpacker = unpackerPath
         else:
             self.unpacker = os.path.join(getWMBASE(),
                                          'WMCore/WMRuntime/Unpacker.py')
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -80,6 +80,7 @@ config.Agent.agentNumber = 0
 config.Agent.useMsgService = False
 config.Agent.useTrigger = False
 config.Agent.useHeartbeat = True
+config.Agent.isDocker = False
 
 config.section_("General")
 config.General.workDir = workDirectory
--- a/src/python/WMQuality/Emulators/EmulatorSetup.py
+++ b/src/python/WMQuality/Emulators/EmulatorSetup.py
@@ -48,6 +48,7 @@ def _wmAgentConfig(configFile):
     config.Agent.agentName = "WMAgentCommissioning"
     config.Agent.useMsgService = False
     config.Agent.useTrigger = False
+    config.Agent.isDocker = False
 
     # BossAir setup
     config.section_("BossAir")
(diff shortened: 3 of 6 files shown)


goughes:
I think this is good to merge, but I won't have time to test a final time until after the winter recess. Let's leave it until the new year.

goughes:
I tested this with a normal VM deployment using the 1.2.8 release. It works fine, I think we could do some additional cleanup of this code if desired.

What I found was that the Unpacker location gets set to `/data/srv/wmagent/v1.2.8/sw/slc7_amd64_gcc630/cms/wmagent/1.2.8/lib/python2.7/site-packages/WMCore/WMRuntime/Unpacker.py`. I checked on some other production agents and it gets set the same there as well.

That means the Unpacker path on the else block here doesn't really exist. And I don't see how it could ever exist.
https://github.com/dmwm/WMCore/blob/b970c2b1be1cbaea0f0f893ad348cffc3bf3ad99/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L108-L109

So the path on L114 will always get set.
https://github.com/dmwm/WMCore/blob/b970c2b1be1cbaea0f0f893ad348cffc3bf3ad99/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L113-L114

We could probably just simplify this to have an if/else statement to switch between the Docker path `os.path.join(config.General.workDir + "/Docker/WMRuntime/Unpacker.py")` and the VM path `os.path.join(getWMBASE(), 'WMCore/WMRuntime/Unpacker.py')`

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -100,10 +100,16 @@ def __init__(self, config):
...
-        if os.path.exists(os.path.join(getWMBASE(),
-                                       'src/python/WMCore/WMRuntime/Unpacker.py')):
-            self.unpacker = os.path.join(getWMBASE(),
-                                         'src/python/WMCore/WMRuntime/Unpacker.py')
+        # if agent is running in a container, Unpacker.py must come from a directory
+        # on the host so the condor schedd can see it
+        # config.General.workDir should always be bind mounted to the container
+        if config.Agent.isDocker:
+            unpackerPath = os.path.join(config.General.workDir + "/Docker/WMRuntime/Unpacker.py")
+        else:
```
amaltaro: Eric, would you know what `getWMBASE()` is when running from a container?

amaltaro: Oh, and did you test both of these cases?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -100,10 +100,16 @@ def __init__(self, config):
...
         self.packageDir = None
 
-        if os.path.exists(os.path.join(getWMBASE(),
-                                       'src/python/WMCore/WMRuntime/Unpacker.py')):
-            self.unpacker = os.path.join(getWMBASE(),
-                                         'src/python/WMCore/WMRuntime/Unpacker.py')
+        # if agent is running in a container, Unpacker.py must come from a directory
+        # on the host so the condor schedd can see it
+        # config.General.workDir should always be bind mounted to the container
+        if config.Agent.isDocker:
```
amaltaro: One more comment here, how about changing it to:
`if getattr(config.Agent, "isDocker", False):`
?
With such change, you don't necessarily need to update those test files/configs

amaltaro (commented):
Erik, I made a few comments in the code. Otherwise patch looks good to me.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
goughes: Would you prefer I leave out the updates to the unit test files then?

amaltaro: Given that they are already in place, it doesn't harm to leave them now.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
goughes: I tested both the Docker and "standard" VM deployment paths

goughes:
Did a quick retest with getattr change to verify submission still works. Things look good from my end.

amaltaro:
Thanks, Erik

PR #9457, opened 2019-12-12
Do not transfer block/dataset if already available within the SiteWhitelist

### LINKED ISSUE #9436: MSTransferor: Evaluate data location before proceeding with data placement
**Impact of the new feature**
MS Transferor

Before we start making data placement (read, phedex subscription) in the MSTransferor service, we should fetch the current input data location (likely blockreplicas API) and, not transfer any blocks that are already available in one or more of the workflow site whitelist.
Not to be done in this issue, but we might consider the `MaxCopies` parameter in the future, such that we can have a more intelligent data placement.

Once migrated to Rucio, this will no longer be needed because the rule should be smart enough to handle this case.

If data (block/dataset) is already completely available in a subset of the site whitelist, do:
* remove those blocks/dataset from the data to be transferred
* remove those sites (PNNs) from the list of storage to receive data. OR, sum up the amount of data in this PNN and use the remainder chunk size for extra blocks.

none

**Additional context**
none

amaltaro:
Fixes #9436 

#### Status
not-tested

#### Description
In short, don't transfer any input data if it's already available in a sub-set of the SiteWhitelist.

At length, changes are:
* added a new configurable parameter `minPercentCompletion`, used to evaluate whether a dataset is fully available at a location or not (default is 99%)
* bypass any input data placement for `StoreResults` and `Resubmission` workflows
* changed the format of `primaryBlocks` within a Workflow object such that it holds the block location too
* changed the format of `parentBlocks` within a Workflow object such that it holds the block location too
* changed the format of `secondarySummaries` within a Workflow object such that it holds the dataset location too
* use `blockreplicas` PhEDEx API to fetch the pileup dataset size (getPileupDatasetSizes)
* use `blockreplicas` PhEDEx API to fetch the parent/primary block size and location, assuming that `complete=y & subscribed=y & group=DataOps` (getBlockReplicasAndSize)
* use `subscriptions` PhEDEx API to fetch DATASETs subscribed and their locations, assuming that `group=DataOps & percent_min=99` (getPileupSubscriptions)
* removed the in memory object caching for pileup dataset size. Rely on the Requests module cache.
* added `MSTransferor/checkDataPlacement` method to evaluate which input data is already available at the SiteWhitelist-SiteBlacklist; any data found is removed from the data placement

TODO:
* use the equivalent Rucio APIs (if service is configured to use Rucio)
* deal with multiple copies

different logic, but works with whatever is already in the system

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -23,6 +23,7 @@ class Workflow(object):
         self.inputDataset = ""
         self.parentDataset = ""
         self.pileupDatasets = set()
+
         self.campaigns = set()
         self.dataCampaignMap = []
         # these blocks structure will be key'ed by the block name and value'd by the block size
@@ -30,7 +31,7 @@ class Workflow(object):
         self.parentBlocks = {}
         # sort of duplicate info, but we need to have a way to link input to parent block(s)
         self.childToParentBlocks = {}
-        # pileup don't need to get resolved into blocks, store only their total size
+        # pileup don't need to get resolved into blocks, store only their total size and location
         self.secondarySummaries = {}
 
         self.setDataCampaignMap()
@@ -94,6 +95,11 @@ class Workflow(object):
         {"type": "type of input data", "name": "dataset name", "campaign": "campaign name"}
         Also set a flat set of campaign names
         """
+        if self.data['RequestType'] in ("StoreResults", "Resubmission"):
+            self.logger.info("Request type %s does not support input data placement", self.data['RequestType'])
+            self.dataCampaignMap = []
+            return
+
         inputMap = {"InputDataset": "primary", "MCPileup": "secondary", "DataPileup": "secondary"}
 
         if "TaskChain" in self.data or "StepChain" in self.data:
@@ -194,7 +200,9 @@ class Workflow(object):
         """
         Sets a list of primary input blocks taking into consideration
         the BlockWhitelist and BlockBlacklist
-        :param blocksDict: flat dict of block name and block size
+        Data is in the form of:
+        {"block_name": {"blockSize": 1234,
+                        "locations": [list of locations]}}
         """
         self.primaryBlocks = blocksDict
 
@@ -204,13 +212,20 @@ class Workflow(object):
         """
         return self.primaryBlocks
 
-    def setSecondarySummary(self, dsetName, dsetSize):
+    def setSecondarySummary(self, dsetName, dsetSize, locations=None):
         """
-        Sets the secondary dataset name and its total size, in bytes
+        Create a summary of the pileup dataset, with its total data size
+        and locations where the whole dataset is subscribed and available
         :param dsetName: string with the secondary dataset name
         :param dsetSize: integer with the secondary dataset size
+        :param locations: locations hosting this dataset in full (and subscribed)
+        Data is in the form of:
+        {"dataset_name": {"dsetSize": 1234,
+                          "locations": [list of locations]}}
         """
-        self.secondarySummaries[dsetName] = dsetSize
+        self.secondarySummaries.setdefault(dsetName, {})
+        self.secondarySummaries[dsetName]['dsetSize'] = dsetSize
+        self.secondarySummaries[dsetName]['locations'] = locations or []
 
     def getSecondarySummary(self):
         """
@@ -223,7 +238,8 @@ class Workflow(object):
         Sets a list of parent input blocks and their size.
         NOTE: this list is solely based on the parent dataset, without
         considering what are the actual input primary blocks
-        :param blocksDict: flat dict of block name and block size
+        {"block_name": {"blockSize": 1234,
+                        "locations": [list of locations]}}
         """
         self.parentBlocks = blocksDict
 
@@ -278,13 +294,13 @@ class Workflow(object):
           * and a list integers, which references the total size of each chunk in
             the list above (same order).
         """
-        if numChunks <= 1:
+        if numChunks == 1:
             thisChunk = set()
             thisChunk.update(self.getPrimaryBlocks().keys())
-            thisChunkSize = sum(self.getPrimaryBlocks().values())
+            thisChunkSize = sum([blockInfo['blockSize'] for blockInfo in self.getPrimaryBlocks().values()])
             if self.getParentDataset():
                 thisChunk.update(self.getParentBlocks().keys())
-                thisChunkSize += sum(self.getParentBlocks().values())
+                thisChunkSize += sum([blockInfo['blockSize'] for blockInfo in self.getParentBlocks().values()])
             # keep same data structure as multiple chunks, so list of lists
             return [thisChunk], [thisChunkSize]
 
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -6,7 +6,7 @@ Original code: https://github.com/CMSCompOps/WmAgentScripts/Unified
 """
 
 # futures
-from __future__ import division
+from __future__ import division, print_function
 
 import json
 import logging
@@ -66,6 +66,9 @@ def dbsInfo(datasets, dbsUrl):
 
     for row in data:
         dataset = row['url'].split('=')[-1]
+        if row['data'] is None:
+            print("FAILURE: dbsInfo for %s. Error: %s %s" % (dataset, row.get('code'), row.get('error')))
+            continue
         rows = json.loads(row['data'])
         blocks = []
         size = 0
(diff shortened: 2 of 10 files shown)


amaltaro:
@vkuznet Valentin, could you please review this PR? Two things to note:
* Rucio APIs/calls will be done in a different PR
* as you know, we haven't explored multithreading for these application and they are not thread safe

src/python/WMCore/MicroService/DataStructs/Workflow.py
```diff
@@ -278,13 +294,13 @@ def getChunkBlocks(self, numChunks=1):
           * and a list integers, which references the total size of each chunk in
             the list above (same order).
         """
-        if numChunks <= 1:
+        if numChunks == 1:
             thisChunk = set()
             thisChunk.update(self.getPrimaryBlocks().keys())
-            thisChunkSize = sum(self.getPrimaryBlocks().values())
+            thisChunkSize = sum([blockInfo['blockSize'] for blockInfo in self.getPrimaryBlocks().values()])
```
vkuznet: Here I suggest to make it safe and replace `blockInfo["blockSize"]` with `blockInfo.get("blockSize", 0)` and even better would be to add if statement to test that blockInfo really exists and is a dict type.

src/python/WMCore/MicroService/DataStructs/Workflow.py
```diff
@@ -278,13 +294,13 @@ def getChunkBlocks(self, numChunks=1):
...
             if self.getParentDataset():
                 thisChunk.update(self.getParentBlocks().keys())
-                thisChunkSize += sum(self.getParentBlocks().values())
+                thisChunkSize += sum([blockInfo['blockSize'] for blockInfo in self.getParentBlocks().values()])
```
vkuznet: same argument as above.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -80,12 +80,39 @@ def dbsInfo(datasets, dbsUrl):
...
+    sizeByDset = {}
+    if not datasets:
+        return sizeByDset
+
+    urls = ['%s/blockreplicas?dataset=%s' % (phedexUrl, dset) for dset in datasets]
+    data = multi_getdata(urls, ckey(), cert())
+
+    for row in data:
+        dataset = row['url'].split('=')[-1]
+        rows = json.loads(row['data'])
```
vkuznet: here you should account that row may not contain 'data' or be dict type and add a check before asking `json.loads` to parse the stream. The parsing may fail too, so wrap it up with appropriate try/except block.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -80,12 +80,39 @@ def dbsInfo(datasets, dbsUrl):
...
+        sizeByDset.setdefault(dataset, 0)  # flat dict in the format of blockName: blockSize
+        for item in rows['phedex']['block']:
```
vkuznet: this is phedex specific and you if you're planning on parsing Rucio data you may put this part in function which will properly handle both phedex/rucio data.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -101,10 +128,43 @@ def phedexValidBlocks(datasets, phedexUrl):
         rows = json.loads(row['data'])
         dsetBlockSize.setdefault(dataset, {})  # flat dict in the format of blockName: blockSize
         for item in rows['phedex']['block']:
-            dsetBlockSize[dataset].update({item['name']: item['bytes']})
+            block = {item['name']: {'blockSize': item['bytes'], 'locations': []}}
```
vkuznet: again, phedex specific part, you may consider write it in more generic way to handle both phedex and rucio streams of data.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -101,10 +128,43 @@ def phedexValidBlocks(datasets, phedexUrl):
...
+    if not datasets:
+        return locationByDset
+
+    url = "%s/subscriptions?group=%s&percent_min=%s&dataset=%s"
+    urls = [url % (phedexUrl, group, percentMin, dset) for dset in datasets]
+    data = multi_getdata(urls, ckey(), cert())
+
+    for row in data:
+        dataset = row['url'].rsplit('=')[-1]
+        rows = json.loads(row['data'])
```
vkuznet: again there is no error conditioning checks here. And part below is phedex specific.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -176,16 +178,63 @@ def getRequestRecords(self, reqStatus):
 
         return orderedRequests
 
+    def checkDataPlacement(self, wflow):
+        """
+        Check which data is already in place (in the workflow SiteWhitelist)
+        :param wflow: workflow object
+        :return: don't know yet
```
vkuznet: Since I don't know what you expect this method to do, but I think you should decomposed it into pieces. One method should check a single dataset placement. Then this method should get list of datasets and ask previous method to check their placements. The return would be either list of dict of datasets along with their checks.

src/python/WMCore/MicroService/Unified/RequestInfo.py
```diff
@@ -233,41 +230,54 @@ def getParentDatasets(self, workflows):
                 if wflow.hasParents() and wflow.getInputDataset() in parentageMap:
                     wflow.setParentDataset(parentageMap[wflow.getInputDataset()])
 
+
+    def getSecondaryDatasets(self, workflows):
```
vkuznet: This method modifies workflows *in place* which is always dangerous business and not thread-safe. Instead takes the workflows and return a new list with proper parameter sets. In place changes can be so tricky to debug and not to mention that entire codebase should *knows* about it and have proper logic.

vkuznet (changes requested):
Alan, I made my review and once again I want to warn you that creation of not thread-safe code can lead to issues in a long run. Please re-consider every time you create a code as not thread-safe.

vkuznet:
Alan, I submitted my review already and I don't know why you requested it second time. Feel free to go through all unresolved comments.

amaltaro:
I hadn't officially requested it through github. It's fine, you can discard the second notification. I'm going through your comments now and will get them answered.

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: Ok, I see now that `data` can actually be `None` if the request fails:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/pycurl_manager.py#L469
I'll make sure `data` evaluates to True before doing anything.
About the return data type, I understand it will only change if we change the request headers. Provided we keep making the same requests to the same API, I think it's pretty safe not to check it (otherwise the code becomes quite unreadeable as well with so many try/except blocks)

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: I'm actually planning to make a separate function for the equivalent API in Rucio. It's more lines of code, but it looks cleaner and easier to cleanup things once we have completely migrated to Rucio.

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: Yes, the rucio part will be made in a separate PR because this one has already too many changes. I'm planning to make separate functions for Rucio calls (code is pretty small anyways)

src/python/WMCore/MicroService/DataStructs/Workflow.py
amaltaro: I think this can be guaranteed from the origin of this information (getBlockReplicasAndSize)

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: I applied those changes to all functions in this module

src/python/WMCore/MicroService/Unified/Common.py
vkuznet: Alan, since you pointed out to pycurl_manager code, the returned message contains url, data and error and code. So you probably need to check for error/code and treat response correctly.

src/python/WMCore/MicroService/Unified/MSTransferor.py
amaltaro: I updated the docstring to make it more clear. This method is pretty straightforward though and all the information is already available in the `Workflow` object. 
It simply does an intersection of the site list with the block/dataset location, if it evaluates to True, block/dataset gets popped out of the structure/dict.

src/python/WMCore/MicroService/Unified/RequestInfo.py
amaltaro: I see, let me try to break it down into two pieces. There will be 3 other places requiring the same changes too.

amaltaro:
@vkuznet Valentin, the third commit has some of the suggestions you made. I'm going to run a quick test now. Please have another look

vkuznet (commented):
The changes are ok, I only have concern about changing data in place, but it is your choice.

amaltaro:
Yes, in the end that boils down to a tradeoff between resource usage (more memory to allocate, more cpu cycles, more LOC) to a tricky code. Hopefully I won't regret that in the coming months.
Merging it and getting it out of my plate.

PR #9473, opened 2020-01-10
Properly use the status/info MS REST APIs

### LINKED ISSUE #9472: ReqMgr2MS: use the status REST API to return a summary of the latest cycle
**Impact of the new feature**
MicroServices

none

Right now the `status` API returns basically a status:OK. I'd suggest to keep record of the actions/problems/time/etc executed in the latest cycle of the service and return it in the `status` API.
This data is supposed to come from the memory and we won't have any history, but just a snapshot.

Still to be decided, but the information we can return could be:
* start time (UTC): last time when the thread got executed
* end time (UTC): when the thread ended
* execution time: difference between end and start time.
* advanced workflows: list of workflows advanced to the next status
* problematic workflows: workflows that failed (for whatever reason) to move on
* status: status of the service
* microservice: name of the microservice
* wmcore_version: version of the wmcore package
* microservice_version: version of the microservice
* what else?!

Another possibility could be to persist that summary in a couch document. But again, it would be just a snapshot, so I think the in-memory option is sound.

**Additional context**
This is important to know how the service is behaving and have a very basic information of what it has done.

amaltaro:
Fixes #9472 

#### Status
not-tested

#### Description
* replaced the `data` REST API by two other APIs: `status` and `info`
* GET `status` REST API returns mostly the same data. However, one can provide the microservice name `?service=transferor` (or `monitor`) to fetch further information about the last cycle of the service.
* GET `info` REST API returns mostly the same data as the status API. However, one can provide a request name`?request=REQUEST_NAME` to fetch the transfer document for that request.
* fixed the POST calls. For now it has the same behaviour as the `info` API.
* created another module `ServicesSummary` to keep some of the data structures used within the MicroServices. Keys not yet enforced though.
* changed MSTransferor to use PUT method when creating a new transfer info document, such that it goes through if a document already exists (POST method would return a document conflict and the request would be forever stuck).

Likely yes

#### Related PRs
none

It requires these deployment changes: https://github.com/dmwm/deployment/pull/828

It also requires changes to cmsdist. `reqmgr2ms.spec` needs to depend on `py2-future` now.


vkuznet:
Alan, if you plan to make these changes may be you may consider to provide `/metrics` end-point. This is common convention for Prometheus exporters which I can set for MS service. Then all of the metrics will appear in Prometheus and we can build dashboard about MS status, set alerts, etc. This name is only a convention, it can be any other name too. I just want to mention this.

amaltaro:
Valentin, given that that api is not essential and/or needed by the end user, I'll tackle it in a different pull request later in the next week. Thanks for the suggestion though.

amaltaro:
@vkuznet I'm still running some final tests, but it would be great if you could have a look at this PR before we start the weekend. If everything goes well, I'm likely merging it still tonight. Thanks

src/python/WMCore/MicroService/Service/RestApiHub.py
```diff
@@ -32,4 +32,5 @@ def __init__(self, app, config, mount):
         cherrypy.log("MicroService entire configuration:\n%s" % Configuration.getInstance())
         cherrypy.log("MicroService REST configuration subset:\n%s" % config)
 
-        self._add({"data": Data(app, self, config, mount)})
+        self._add({"status": Data(app, self, config, mount),
```
amaltaro: I have the impression this is what causes the duplicate records in the logging. Still trying to understand it though...

vkuznet:
Alan,
sorry for being late, I'm on a road and without internet. I did a brief review and everything looks good to me.

amaltaro:

I was investigating the source of the duplicate records in the log, and it turns out that `RestApiHub` doesn't only create an instance of the REST API, but it also runs the whole application (MSManager).

In other words, for each each API defined here (using the same `Data` class):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Service/RestApiHub.py#L35
a whole application gets executed in this line (`MSManager`):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Service/Data.py#L52

Thus, the duplicate records is actually multiple instances of the same application!

I need to think what's the best way to solve it, which could be:
* separating the application from the REST service/apis
* using a different module for every REST api
* something else...

amaltaro:
Problem solved in the 3rd commit by defining `Data` as a singleton class. Tests look fine, but I'll x-check again tomorrow morning.

It also requires changes to the spec file, which needs to now `Requires: py2-future`

amaltaro:
Tests look good and I'm merging it to move on.

PR #9480, opened 2020-01-13
ReqMgr2MS more configurable: # of workflows; status transition; data transfers

### LINKED ISSUE #9479: Extend the ReqMgr2MS configuration to support different write flags and # of workflows
**Impact of the new feature**
ReqMgr2MS

We need to have a way to control the amount of workflows being processed by MSTransferor within the same cycle; and also have a flag allowing microservices to make a workflow status change or not.

The goal of these changes is that we run microservices in production, allowing it to:
* create the transfer info document
* update the workflow status
* but NOT create any data transfer requests

So, we need to extend the service configuration (deployment repo) such that we can control:
1. number of workflows MSTransferor can process within the same cycle (not to create too many data transfers in one shot)
2. split the `readOnly` configuration into 2 different configurations
 * `enableStatusTransition`: to allow microservices to update the status of workflows or not (default to False)
 * `enableDataTransfer`: to allow microservices to subscribe data or not (default to False)

I think we can let microservices create the transfer info couch document by default, there is likely no need to make that one configurable...

none

**Additional context**
none

amaltaro:
Fixes #9479 

#### Status
ready

#### Description
It supports different service configuration parameters, such as:
* `readOnly` has been removed
* `enableStatusTransition` decides whether MSTransferor/MSMonitor are allowed to request status transition
* `enableDataTransfer` decides whether MSTransferor can make data transfer requests (PhEDEx subscriptions and/or Rucio rules)
* `limitRequestsPerCycle` limit MSTransferor service to this amount of workflows per cycle (still obeying to the workflow priorities).

nope, new configuration attributes.

#### Related PRs
None

Yes, these deployment changes: https://github.com/dmwm/deployment/pull/829

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSTransferor.py
+++ b/src/python/WMCore/MicroService/Unified/MSTransferor.py
@@ -77,6 +77,8 @@ class MSTransferor(MSCore):
         self.msConfig.setdefault("minPercentCompletion", 99)
         # minimum available storage to consider a resource good for receiving data
         self.msConfig.setdefault("minimumThreshold", 1 * (1000 ** 4))  # 1TB
+        # limit MSTransferor to this amount of requests per cycle
+        self.msConfig.setdefault("limitRequestsPerCycle", 500)
 
         self.rseQuotas = RSEQuotas(self.msConfig['detoxUrl'], self.msConfig["quotaAccount"],
                                    self.msConfig["quotaUsage"], useRucio=self.msConfig["useRucio"],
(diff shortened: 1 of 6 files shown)


amaltaro:
Tests are looking Okay. @vkuznet could you please review these changes? Thanks

amaltaro:
Valentin, please feel free to make any comments to this PR. I'm merging it now because I need to cut another release for testbed.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -335,6 +337,9 @@ def makeTransferRequest(self, wflow):
                         self.dsetCounter += 1
                     else:
                         self.blockCounter += len(blocks)
+                else:
+                    self.logger.info("FAKE data subscription. Default transfer id set to: '1111'")
+                    transRec['transferIDs'].add("1111")  # any fake number
```
vkuznet: I"m not sure why do you need fake numbers at all? If there is no transfer it should not be any IDs either. Better you should put `None` as transferIDs which resolve any ambiguity. Then, in other parts of the code, you should check that transferIDs should be not a `None` which is the only right criteria for IDs. All other IDs should be considered as possible values.

vkuznet (commented):
I only had one comment about fake IDs

src/python/WMCore/MicroService/Unified/MSTransferor.py
amaltaro: I think I had a fake ID in order to test the whole cycle between Transferor and Monitor. At this stage they are indeed no longer needed, especially because we're about to set things in write mode.
That field is supposed to contain a list of transfer IDs, so if there is none, it's better to set it to an empty string.
Thanks for your comments though

PR #9483, opened 2020-01-14
Skip input data placement if data has availability AND AAA is enabled

### LINKED ISSUE #9482: Handle data placement in ReqMgr2MS for workflows with AAA enabled
**Impact of the new feature**
ReqMgr2MS

I believe we don't have any special treatment for workflows assigned with AAA enabled (either for primary/parent data or for the pileup dataset). So we need to have such logic implemented in the Microservices to support this use case.

It might need further thought, but in summary:
* we need to make sure all the required data is available on disk at one storage, at least
* there is no need to subscribe the input data to one of the SiteWhitelist location. It does not mean we cannot pick one of those though.

Am I missing anything else? @thongonary @sharad1126 

Not do anything and just assume data is in place

**Additional context**
none

amaltaro:
Fixes #9482 

#### Status
not-tested

#### Description
With this patch, MSTransferor takes into consideration the AAA flags (`TrustSitelists`, `TrustPUSitelists`) before actually performing any input data placement.

In short:
* if `TrustSitelists` is enabled, check whether the input blocks are available somewhere on disk. If so, skip that given block from the data placement
* if `TrustPUSitelists` is enabled, check whether the input dataset is available somewhere on disk. If so, skip that dataset from the data placement

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -65,6 +65,15 @@ class Workflow(object):
         """
         return self.data['RequestType']
 
+    def getReqParam(self, param):
+        """
+        Return a top level parameter for this request.
+        Read: parameters internal to Task/Steps are not looked up.
+        """
+        if param not in self.data:
+            self.logger.warning("Request parameter '%s' not found in the workflow %s", param, self.getName())
+        return self.data.get(param)
+
     def getSitelist(self):
         """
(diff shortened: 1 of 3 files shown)


amaltaro:
Valentin, Todor, can one of you have a look at this pull request? Tests went fine and I have other developments waiting for this to go in ;) Thanks

vkuznet (approved):

PR #9485, opened 2020-01-14
Disconsider phedex group for input data existence check

### LINKED ISSUE #9484: Automatic input data existence check should disconsider the phedex group
**Impact of the new feature**
ReqMgr2MS

While checking my private tests on the MicroServices, I noticed that some blocks were at a given disk location, but it wasn't seen by the service. It turns out we match the `blockreplicas` output against a given PhEDEx group (in this case `DataOps`).

After chatting with Sharad, he confirmed that Unified doesn't care whether the dataset/block is subscribed under a PhEDEx group different than `DataOps`. If data is available, this is all that matters to proceed with the workflow assignment/execution. 

In short, we should remove the phedex group check when evaluating which replicas are already in place. If data is complete and subscribe, skip it from the data transfer.

Which is different than the data placement. For that, we do have to define the `DataOps` group, always.

**Additional context**
Problem spotted with this DMWM/TaskChain_NanoAOD template.

amaltaro:
Fixes #9484 

#### Status
not-tested

#### Description
During the MSTransferor input data placement, we first check for data existence such that we don't need to subscribe all the input. With this patch, the input data don't need to be necessarily subscribed under `DataOps`.

Usual input data placement made by MSTransferor will keep subscribing data under `DataOps` though, that remains untouched.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -112,14 +112,16 @@ def getPileupDatasetSizes(datasets, phedexUrl):
     return sizeByDset
 
 
-def getBlockReplicasAndSize(datasets, phedexUrl, group):
+def getBlockReplicasAndSize(datasets, phedexUrl, group=None):
     """
     Given a list of datasets, find all their blocks with replicas
-    available, i.e., blocks that have valid files to be processed,
-    keeping the block size and blocks completed and subscribed
+    available (thus blocks with at least 1 valid file), completed
+    and subscribed.
+    If PhEDEx group is provided, make sure it's subscribed under that
+    same group.
     :param datasets: list of dataset names
     :param phedexUrl: a string with the PhEDEx URL
-    :param group: PhEDEx group name
+    :param group: optional PhEDEx group name
     :return: a dictionary in the form of:
     {"dataset":
         {"block":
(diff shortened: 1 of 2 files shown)


amaltaro:
@vkuznet @todor-ivanov another simple change that affects microservices. Could you please have a look at it? Tests went fine in my VM.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -146,8 +148,11 @@ def getBlockReplicasAndSize(datasets, phedexUrl, group):
         for item in rows['phedex']['block']:
             block = {item['name']: {'blockSize': item['bytes'], 'locations': []}}
             for repli in item['replica']:
-                if repli['complete'] == 'y' and repli['subscribed'] == 'y' and repli['group'] == group:
-                    block[item['name']]['locations'].append(repli['node'])
+                if repli['complete'] == 'y' and repli['subscribed'] == 'y':
+                    if not group:
+                        block[item['name']]['locations'].append(repli['node'])
+                    elif repli['group'] == group:
```
vkuznet: what will happen if group is not `None` but also does not match of replica group?

amaltaro: In that case locations will remain an empty list (L149). Which means, MSTransferor must make a data subscription later in the execution.

amaltaro:
test this please

amaltaro:
Tests went fine. Valentin, please let me know if you have further questions.

PR #9489, opened 2020-01-16
Add initial sleep to CP threads; improve logging

### LINKED ISSUE #9488: Insufficient logging in DataCacheUpdate reqmon thread
**Impact of the bug**
WMStats threads, at least

**Describe the bug**
Lack of logging making it impossible for any reasonable debugging.

**How to reproduce it**
none

**Expected behavior**
When implementing these catch-all exceptions, we should at least log the traceback. Otherwise there is no way to find out which module/line is actually hitting a problem.

Current log of one of the testbed reqmon nodes (DataCacheUpdate thread)
```
2020-01-15 20:11:50,488:INFO:LogDB:<LogDB(url=https://cmsweb-testbed.cern.ch/couchdb/wmstats_logdb, identifier=wmstats-vocms0731.cern.ch, agent=1)>
2020-01-15 20:12:00,773:ERROR:DataCacheUpdate:argument of type 'NoneType' is not iterable
2020-01-15 20:17:09,544:ERROR:DataCacheUpdate:argument of type 'NoneType' is not iterable
```

amaltaro:
Fixes #9488 

#### Status
ready

#### Description
Started this patch by simply printing the traceback in the cherrypy threads, when reasonable.
In addition to that:
* try to log when the thread starts and when it's about to end
* added a random 30 to 90 secs initial sleep before the thread runs for the first time (not to start all of them at the same time; and to give dependencies some extra time to get started)

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/REST/CherryPyPeriodicTask.py
+++ b/src/python/WMCore/REST/CherryPyPeriodicTask.py
@@ -1,11 +1,14 @@
-'''
+"""
+
 Created on Jul 31, 2014
 
 @author: sryu
-'''
+"""
 from __future__ import print_function, division
 import cherrypy
 import traceback
+from random import randint
+from time import sleep
 from WMCore.WMLogging import getTimeRotatingLogger
 
 from threading import Thread, Condition
@@ -13,7 +16,6 @@ from threading import Thread, Condition
 class CherryPyPeriodicTask(object):
 
     def __init__(self, config):
-
         """
         BaseClass which can set up the concurrent task using cherrypy thread.
         WARNING: This assumes each task doesn't share the object.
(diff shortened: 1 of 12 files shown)


amaltaro:
test this please

amaltaro:
Changes look good and it's ready to go.

For the record, I think we could time every cherrypy thread from the super class, instead of implementing that logic inside each thread.

amaltaro:
I don't like this patch. It gives me all the components red upon wmstats startup. A better way needs to be implemented; or we define that some threads have no need to sleep at startup (like hearbeat and dataCache)...

src/python/WMCore/REST/CherryPyPeriodicTask.py
```diff
@@ -84,6 +86,10 @@ def stop(self):
         self.wakeUp.release()
 
     def run(self):
+        firstSleep = randint(30,90)
```
todor-ivanov: Could it get some processes out of sync?

todor-ivanov (commented):
I agree with both of your previous comments @amaltaro. 
The timing is definitely better to be started one level up, but then I think it will count the random sleep too. 
Did you prove that  heartbeat is the only thread that needs to be excluded from those to be delayed at startup? Besides that all the rest looks good to me.

todor-ivanov (approved):
It is up to you to decide either to fix the aforementioned in this PR or to create another one.

amaltaro:
Old and no longer relevant fix. Closing this one out.

PR #9492, opened 2020-01-16
Put MicroServices to sleep for a few seconds before its first execution

### LINKED ISSUE #9491: Add initial sleep to the MicroServices before executing the threads
**Impact of the new feature**
ReqMgr2MS

Depending on the order services get started, it might be that Transferor and Monitor fail to execute their first cycle.

Implement an initial sleep time during the threads initialisation.

Make sure dependencies are started in the correct order :)

**Additional context**
none

amaltaro:
Fixes #9491 

#### Status
ready

#### Description
Sleep between 30 and 90 seconds before start executing the MSMonitor and/or MSTransferor threads.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSManager.py
+++ b/src/python/WMCore/MicroService/Unified/MSManager.py
@@ -27,6 +27,7 @@ from __future__ import division, print_function
 # system modules
 from time import sleep
 from datetime import datetime
+from random import randint
 
 # WMCore modules
 from WMCore.MicroService.Unified.Common import getMSLogger
@@ -75,8 +76,7 @@ class MSManager(object):
                                                   'assigned',
                                                   self.msConfig['interval'],
                                                   self.logger))
-            self.logger.debug(
-                "### Running %s thread %s", thname, self.transfThread.running())
+            self.logger.info("Running %s thread: %s", thname, self.transfThread.running())
 
         # initialize monitoring module
         if 'monitor' in self.services:
@@ -87,8 +87,7 @@ class MSManager(object):
                                                  'staging',
                                                  self.msConfig['interval'],
                                                  self.logger))
-            self.logger.debug(
-                "+++ Running %s thread %s", thname, self.monitThread.running())
+            self.logger.info("Running %s thread: %s", thname, self.monitThread.running())
 
     def _parseConfig(self, config):
         """
@@ -114,6 +113,11 @@ class MSManager(object):
         For references see
         https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor
         """
+        if not self.statusTrans:
+            firstSleep = randint(30, 90)
+            self.logger.info("MSTransferor thread initialized. Sleeping %d secs before its 1st cycle",
+                             firstSleep)
+            sleep(firstSleep)
         startTime = datetime.utcnow()
         self.logger.info("Starting the transferor thread...")
         res = self.msTransferor.execute(reqStatus)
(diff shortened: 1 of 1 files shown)


src/python/WMCore/MicroService/Unified/MSManager.py
```diff
@@ -114,6 +113,11 @@ def transferor(self, reqStatus):
         For references see
         https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor
         """
+        if not self.statusTrans:
+            firstSleep = randint(30, 90)
+            self.logger.info("MSTransferor thread initialized. Sleeping %d secs before its 1st cycle",
+                             firstSleep)
+            sleep(firstSleep)
```
vkuznet: I really doubt it will solve the problem, in fact it will introduce a problem when time interval will exceed provided range and service will fail to run. I provide my comments in https://github.com/dmwm/WMCore/issues/9491 and I think the issue should be treated differently.

It should be no firstSleep at all. Instead both Transferor and Monitor should be able to *ping* an external service before the start-up. If service is not responding the code should introduce incremental sleep with possibility to give up if external service is not responding after certain iterations (which can be configurable parameter). The initial interval should be specified in configuration, instead of hardcoded here.

src/python/WMCore/MicroService/Unified/MSManager.py
```diff
@@ -129,6 +133,11 @@ def monitor(self, reqStatus):
         For references see
         https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Transferor
         """
+        if not self.statusMon:
+            firstSleep = randint(30, 90)
+            self.logger.info("MSMonitor thread initialized. Sleeping %d secs before its 1st cycle",
+                             firstSleep)
+            sleep(firstSleep)
```
vkuznet: the same argument here.

vkuznet (changes requested):
The best way to deal with this is the following:
- add `def probe()` method to MSCore and implement it properly in MSTransferor and MSMonitor
- the probe method should contact necessary external service and return status True or False which classes can use to decide if they can contact the external services
- one way to implement is to get simple HTTP info response from Couch/ReqMgr2/Phedex/DBS or whatever services MS will need to contact. It is cheap HTTP call but it can provide information about upstream services
- if you're parnoid then you can add probe call to every method os MS classes which will ensure that it can talk to services. Of course it does not guarantee that service during method operation may not go down, but it will ensure that service was indeed ready before the operation.

I strongly suggest to remove the initial sleep since it will potentially only confuse and mis-lead in a long run, especially if external service will take longer to wake up with respect to provided time intervals for sleep function.

I also suggest to implement incremental sleep with probe calls and configure it externally such that you can easily adjust service behavior without modifying its codebase.

amaltaro:
That's a valid argument Valentin. However it also has it's cons and there are cases that we really don't need anything "sophisticated".
As I already said, this is only needed for the startup of the service, when we are likely deploying and starting a bunch of things in parallel. Of course it could be extended to every cycle though (if cycles are not too short).

In addition to that, a random initial sleep is also part of the WMAgent and all components/threads follow the same logic (which dates before I actually joined the team). Which helps not to start a bunch of (sometimes heavy) threads all in parallel.

Last but not least, the polling cycle here is 10min, so if one of the services is not up and we need to wait for the next cycle, the delay added is negligible.

Still, if you insist on your approach, can you please create a PR and run some basic tests? I'm definitely not requesting anything for testbed before tomorrow 10am CERN time. Thanks

vkuznet:
Alan, I provided comments based on *best practices* and experience. You can treat them as you like. My goal here is to provide guidance and demonstrate consequences of concrete implementation.

Said that, it is your call, not mine. I provided the review and I don't have time for implementation, and especially for last moment request before the deadline. If you want to add sleep intervals feel free to add it, I'll not stop you. Code-wise everything is fine, and my comments only show why I consider this as not ideal approach.

amaltaro:
Thanks, Valentin. We should have more time during the year to polish these microservices. A sub-optimal fix will do for now, otherwise we might not meet the deadline for all these developments.

amaltaro:
Potentially no longer relevant and/or something that I won't be getting back to in the near future. So I am closing this one out.

PR #9494, opened 2020-01-20
Enable late materialization factory in condor

khurtado:
Enable late materialization with a huge materializion parameter number until job splitting is properly adapted for this feature.
Note: Even if all jobs are materialized, this
already enables the late materialiation factory in condor, so it requires HTCondor 8.9.2+ in the schedd in order to handle proxy certificates properly during the submission requests.

#### Status
In testing

YES

Schedd: HTCondor v8.9.2+


todor-ivanov:
Not merging it yet. We are  waiting for the latest tests with HTCondor 8.9.5. And we need this version deployed on the schedds (both prod/testbed), before we can push it in WMagent. Should go to the next WMAgent release and should be put from the very beginning.

PR #9496, opened 2020-01-21
Fix info REST API for ReqMgr2MS; changed kwarg to detail

### LINKED ISSUE #9495: ReqMgr2MS 'info' API returning 500 Internal Server Error
**Impact of the bug**
ReqMgr2MS

**Describe the bug**
I'm still investigating where and what exactly the error is. However, it looks like that if we request `info` for a given request that has already been handled by MSTransferor, it returns a `500 Internal Server Error` to the end user. There are no traces in the logs besides the DEBUG log where the transferinfo document gets requested.
 
**How to reproduce it**
Assign a workflow, wait MSTransferor to process it, then make a GET request to:
https://alancc7-cloud1.cern.ch/ms-transferor/data/info?request=amaltaro_StepChain_DupOutMod_Jan2020_Val_200121_121530_5515

**Expected behavior**
This API should return a very short summary of the service, and the TRANSFER document for the request provided (if it exists).

All I see in the logs is:
```
2020-01-21 12:26:39,760:INFO:MSManager: Total transferor execution time: 208.38 secs
2020-01-21 12:27:21,991:DEBUG:Service: getData: 
	url: transferinfo/amaltaro_StepChain_DupOutMod_Jan2020_Val_200121_121530_5515
	verb: GET
	incoming_headers: {}
	data: {}

```

amaltaro:
Fixes #9495 

#### Status
not-tested

#### Description
Problem is that we were trying to add a key/value to a list object.
Summary of changes is:
* the `status` API accepts the following keyword argument `detail` (instead of `service`); it's meant to be a boolean (but string variations are also accepted)
* the `info` API will always return these two extra keys (`request`, `transferDoc`); being null in case a request is not provided and/or a transfer document cannot be found in couchdb. Otherwise, `transferDoc` will have the same content as it's currently in couch database.
* updated `lastUpdate` field in the transfer document to be an integer instead of float.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Service/Data.py
+++ b/src/python/WMCore/MicroService/Service/Data.py
@@ -90,7 +90,10 @@ class Data(with_metaclass(Singleton, RESTEntity, object)):
                'microservice': self.mgr.__class__.__name__}
 
         if kwds.get('API') == "status":
-            res.update(self.mgr.status(kwds.get("service")))
+            detail = kwds.get("detail", True)
+            if detail not in (True, "true", "True", "TRUE"):
+                detail = False
+            res.update(self.mgr.status(detail))
         elif kwds.get('API') == "info":
             res.update(self.mgr.info(kwds.get("request")))
         return results(res)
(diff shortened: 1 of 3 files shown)


amaltaro:
Tests look good!

amaltaro:
tests went fine. Feel free to leave your review, Todor.

PR #9503, opened 2020-01-23
Fix to block white/black list in MSTransferor

### LINKED ISSUE #9502: MS: do we consider block white/black list in MSTransferor
**Impact of the bug**
ReqMgr2MS

**Describe the bug**
I am actually not sure it's a bug or not. It might be that code is already in place and this issue is not valid; anyhow, just so I don't forget.
Do we consider BlockWhitelist and BlockBlacklist for input data placement of workflows without any lumi and run lists? This code basically:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/RequestInfo.py#L355-L363

**How to reproduce it**

**Expected behavior**

amaltaro:
Fixes #9502 

#### Status
tested

#### Description
Summary of changes are:
* when setting the list of primary blocks `setPrimaryBlocks`: consider the block white/black list
* when providing a list of input primary/parent blocks to `_handleInputDataInfo`: only apply the block white/black list on the primary blocks (otherwise it would result in an empty list for the parent list)

All these changes are very much annoying because we need to be careful with variable assignment by reference and by value...

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -5,6 +5,7 @@ required by MS Transferor
 from __future__ import division, print_function
 
 import operator
+from copy import copy, deepcopy
 from WMCore.DataStructs.LumiList import LumiList
 from WMCore.MicroService.Unified.Common import getMSLogger, gigaBytes
 
@@ -213,7 +214,16 @@ class Workflow(object):
         {"block_name": {"blockSize": 1234,
                         "locations": [list of locations]}}
         """
-        self.primaryBlocks = blocksDict
+        blockWhite = self.getBlockWhitelist()
+        if blockWhite:
+            for block in blockWhite:
+                if block in blocksDict:
+                    self.primaryBlocks[block] = copy(blocksDict[block])
+        else:
+            self.primaryBlocks = deepcopy(blocksDict)
+
+        for block in self.getBlockBlacklist():
+            self.primaryBlocks.pop(block, None)
 
     def getPrimaryBlocks(self):
         """
@@ -263,16 +273,14 @@ class Workflow(object):
         Sets a relationship between input primary block and its parent block(s)
         This method guarantees that only parent blocks with valid replicas are
         kept around (and later transferred)
+        The child block list is also final, meaning all the run/block/lumi lists
+        have already been applied.
         :param blocksDict: dict key'ed by the primary block, with a list of parent blocks
         """
         # flat list with the final parent blocks
         parentBlocks = set()
         # remove parent blocks without any valid replica (only invalid files)
         for child, parents in blocksDict.items():
-            if child not in self.getPrimaryBlocks():
-                # then we don't need this child+parent data
-                continue
-
             for parent in list(parents):
                 if parent not in self.getParentBlocks():
                     # then drop this block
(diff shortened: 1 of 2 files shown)


amaltaro:
Initial tests look ok. Valentin/Todor, can one of you please have a look at this patch? Thanks

src/python/WMCore/MicroService/DataStructs/Workflow.py
```diff
@@ -213,7 +214,16 @@ def setPrimaryBlocks(self, blocksDict):
         {"block_name": {"blockSize": 1234,
                         "locations": [list of locations]}}
         """
-        self.primaryBlocks = blocksDict
+        blockWhite = self.getBlockWhitelist()
+        if blockWhite:
+            for block in blockWhite:
+                if block in blocksDict:
+                    self.primaryBlocks[block] = blocksDict[block]
```
vkuznet: Two lines below you do a deepcopy of blocksDict, while here you take the assingment. Since I don't know what blocksDict[block] data type is it may be dangerous in case if it is dict type.

vkuznet (commented):
Alan, I don't really know the data structure this code expects and operates with. But from the content my understanding that it only needs to have a block lists.

For that it is much better to have internal structure as dict where keys will be block names and values those keys will integers. Doing this way you still maintain O(1) look-up but simplify memory management. For example:
```
self.primaryBlocks = {'block#123':1, 'block#234':1, 'block#345': 1}
```
will be easy to maintain and will provide full functionality, while from the code I see that you do a copy of probably much more complex structure `blocksDict`.

src/python/WMCore/MicroService/DataStructs/Workflow.py
amaltaro: That's a good comment! blocksDict[block] is actually a flat dict. So indeed it's safer to do a `copy` here and a `deepcopy` below; even though I did some local tests and they didn't have the same id.

amaltaro:
Yes, block : blocksize was the initial structure I had in mind. It turns out we need to run data location discovery as well and, given that that's done in bulk, I'd rather keep them stored somewhere for late processing/check.
So the current structure is the one mentioned in the doc string:
```
{"block_name": {"blockSize": 1234,
                        "locations": [list of locations]}}
```

src/python/WMCore/MicroService/DataStructs/Workflow.py
todor-ivanov: @amaltaro , I would agree with Valentine's comment here. It is still unclear to me, why in that 'if' there are two different methods for copying. Why not, just to stay on the safe side, use deep copy in both cases? Especially since the data structure is not defined in the current function, but comes as an argument from this call [1] (I believe). It would be safer if something else, but a flat dict is passed to `setPrimaryBlocks`. And in the case when an actual flat dict is passed, it should still produce a legitimate copy. Otherwise it will be an easy error to make - changing the data structure itself, wherever it is defined, and forgetting to update how you copy here. 

[1]
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/RequestInfo.py#L310

todor-ivanov (commented):
Besides the remark I made in the previous comment, all the rest looks fine to me.

src/python/WMCore/MicroService/DataStructs/Workflow.py
amaltaro: Thanks Todor and Valentin. I have changed that to a deepcopy in a PR that I have opened now. It's indeed a dangerous business when selective copying/removing things from such structures. 
We could make it cleaner by defining two different data structures, but that would cause a >25% increase in the memory usage.

PR #9506, opened 2020-01-24
Properly deal with RunWhitelist/RunBlacklist within MS

### LINKED ISSUE #9501: MS: properly handle SiteBlacklist and RunBlacklist
**Impact of the bug**
ReqMgr2MS

**Describe the bug**
The way it has been implemented:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/DataStructs/Workflow.py#L77

`getSitelist` and `getRunlist` would not return the black list, in case the white list is empty. Thus causing transfer of unneeded input data.

**How to reproduce it**
none

**Expected behavior**
We need to make sure that the black lists are properly evaluatde and block/run matching those are actually taken out of the data placement transfers. 

none

amaltaro:
Fixes #9501 

#### Status
tested

#### Description
Supposedly a simple fix, but NO!
In short, this properly deals with cases where only RunBlacklist is provided. In addition to that, the following changes have been made when dealing with workflows that have either run or lumi lists:
* make bulk calls whenever it's possible. Like, to list runs in blocks; and to list run/lumis in blocks
* catch DBS call errors and remove the corresponding workflow from the MSTransferor cycle
* raise exceptions from the Common functions, such that we can catch and take proper actions with the workflows
* fixed bug when a secondary is already in place and gets removed from the list of secondary datasets.

yes

#### Related PRs
Yes, reverted a few things put in on this PR: https://github.com/dmwm/WMCore/pull/9503

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -81,11 +81,18 @@ class Workflow(object):
         """
         return sorted(list(set(self.data['SiteWhitelist']) - set(self.data['SiteBlacklist'])))
 
-    def getRunlist(self):
+    def getRunWhitelist(self):
         """
-        Get the RunWhitelist minus the black list for this request
+        Get the RunWhitelist for this request
         """
-        res = set(self._getValue('RunWhitelist', [])) - set(self._getValue('RunBlacklist', []))
+        res = set(self._getValue('RunWhitelist', []))
+        return sorted(list(res))
+
+    def getRunBlacklist(self):
+        """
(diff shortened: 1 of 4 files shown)


amaltaro:
test this please

amaltaro:
Tests went fine (I checked all kinds of input data types, multiple of them in the same cycle, and so on).
The performance has also improved a lot with this fix.

amaltaro:
I'm merging this one because I need to address other issues involving the same modules. Feel free to leave your comments.

todor-ivanov (approved):
looks good to me.

PR #9511, opened 2020-01-27
Catch data-service queries that failed and retry workflow in the next cycle

### LINKED ISSUE #9493: MS: Retry workflows with data lookup failures in MSTransferor
**Impact of the bug**
ReqMgr2MS MSTransferor

**Describe the bug**
While checking the ms-transferor logs, I noticed that there was a workflow that had a timeout failure while discovering blocks for a given dataset and list of runs, and still it went through the cycle, got a transfer document created in couchdb and advanced to the next status (staging).

**How to reproduce it**
Failed one of those data discoveries call in the Common module.

**Expected behavior**
Catch any exception (including those during the data discovery) and make sure that the corresponding workflow is removed from the cycle. That same workflow needs to be retried again in the next cycle of the service.

Here are some logs from ms-transferor-20200120.log
```
2020-01-20 14:30:41,870:INFO:RequestInfo: ### unified processing 2 requests
2020-01-20 14:30:41,870:DEBUG:RequestInfo: ### getParentDatasets: 0.0 sec
2020-01-20 14:30:41,870:INFO:RequestInfo: Fetching pileup dataset sizes for 1 datasets against PhEDEx: https://cmsweb.cern.ch/phedex/datasvc/json/prod
2020-01-20 14:30:42,029:INFO:RequestInfo: Fetching pileup dataset location for 1 datasets against PhEDEx: https://cmsweb.cern.ch/phedex/datasvc/json/prod
2020-01-20 14:30:42,098:DEBUG:RequestInfo: ### getSecondaryDatasets: 0.2 sec
2020-01-20 14:30:42,099:INFO:RequestInfo: Fetching block info for 1 datasets against PhEDEx: https://cmsweb.cern.ch/phedex/datasvc/json/prod
2020-01-20 14:30:42,287:INFO:RequestInfo: Fetching blocks matching a list of runs for amaltaro_ReReco_11_HG2002_Val_200120_142045_5789
FAILURE: getDbsBlocksRun for /SingleMuon/Run2018D-v1/RAW. Error: 28 Operation timed out after 270280 milliseconds with 0 bytes received
2020-01-20 14:35:12,569:DEBUG:RequestInfo: ### getInputDataBlocks: 270.5 sec
2020-01-20 14:35:12,569:DEBUG:RequestInfo: ### getParentChildBlocks: 0.0 sec
2020-01-20 14:35:12,569:DEBUG:RequestInfo: ### total time for unified method: 270.7 sec
...
2020-01-20 14:35:12,571:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 has 0 initial blocks from getPrimaryBlocks
2020-01-20 14:35:12,571:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 has 0 final blocks from getPrimaryBlocks
2020-01-20 14:35:12,572:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 has 0 initial blocks from getParentBlocks
2020-01-20 14:35:12,572:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 has 0 final blocks from getParentBlocks
2020-01-20 14:35:12,572:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 with 0 initial secondary dataset to be transferred
2020-01-20 14:35:12,572:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 with 0 final secondary dataset to be transferred
2020-01-20 14:35:12,572:INFO:MSTransferor: Request amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 does not have any further data to transfer
2020-01-20 14:35:12,572:INFO:MSTransferor: Transfer requests successful for amaltaro_ReReco_11_HG2002_Val_200120_142045_5789. Summary: []
2020-01-20 14:35:12,796:INFO:ReqMgrAux: Update in-place: False for transfer doc: amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 was successful.
2020-01-20 14:35:12,797:INFO:MSTransferor: Transfer document successfully created in CouchDB for: amaltaro_ReReco_11_HG2002_Val_200120_142045_5789
2020-01-20 14:35:12,797:INFO:MSCore: MSTransferor updating amaltaro_ReReco_11_HG2002_Val_200120_142045_5789 status to: staging
```

### LINKED ISSUE #9500: MS: skip workflow if the pileup size/location requests failed
**Impact of the bug**
ReqMgr2MS

**Describe the bug**
In case one of the PhEDEx/Rucio calls fail in the MSTransferor, the workflow would go through with some default values set in the RequestInfo module.

**How to reproduce it**
none

**Expected behavior**
We need to be able to know when a data-service call has failed, such that we can skip workflows with that dataset from the current cycle, retrying again in the next cycle.
Here's one example:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/RequestInfo.py#L278

if we don't manage to calculate the pileup dataset size and/or find the current locations, then we need to retry that dataset (and its workflows) in the next cycle.

none

amaltaro:
Fixes #9500 
Fixes #9493 

#### Status
tested

#### Description
In short, this PR defines a mechanism to identify calls that failed to get responded by the server (or came back with an invalid result); with that, we can also identify which workflows are affected by them and pop those workflows out of the cycle (they will be retried in the next cycle).

Changes are:
* when parsing the output of bulk calls, set the value to `None` in case the call failed and/or returned invalid data
* in `RequestInfo` module, we need to identify those calls (by the input parameter), then scan all the workflows looking for that input parameter. Workflow matching that will get removed from the MSTransferor cycle and will be retried later.
* if the bulk call is made for a single workflow, we can just raise an exception and catch it upstream, removing the concerning workflow.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -91,6 +91,7 @@ def getPileupDatasetSizes(datasets, phedexUrl):
     :param datasets: list of dataset names
     :param phedexUrl: a string with the PhEDEx URL
     :return: a dictionary of datasets and their respective sizes
+    NOTE: Value `None` is returned in case the data-service failed to serve a given request.
     """
     sizeByDset = {}
     if not datasets:
@@ -102,13 +103,19 @@ def getPileupDatasetSizes(datasets, phedexUrl):
     for row in data:
         dataset = row['url'].split('=')[-1]
         if row['data'] is None:
-            print("FAILURE: getPileupDatasetSizes for %s. Error: %s %s" % (dataset, row.get('code'),
-                                                                          row.get('error')))
+            print("Failure in getPileupDatasetSizes for dataset %s. Error: %s %s" % (dataset,
+                                                                                     row.get('code'),
+                                                                                     row.get('error')))
+            sizeByDset.setdefault(dataset, None)
             continue
         rows = json.loads(row['data'])
         sizeByDset.setdefault(dataset, 0)  # flat dict in the format of blockName: blockSize
-        for item in rows['phedex']['block']:
-            sizeByDset[dataset] += item['bytes']
+        try:
+            for item in rows['phedex']['block']:
+                sizeByDset[dataset] += item['bytes']
+        except Exception as exc:
+            print("Failure in getPileupDatasetSizes for dataset %s. Error: %s" % (dataset, str(exc)))
+            sizeByDset[dataset] = None
     return sizeByDset
 
 
@@ -128,6 +135,7 @@ def getBlockReplicasAndSize(datasets, phedexUrl, group=None):
             {"blockSize": 111, "locations": ["x", "y"]}
         }
     }
+    NOTE: Value `None` is returned in case the data-service failed to serve a given request.
     """
     dsetBlockSize = {}
     if not datasets:
@@ -139,21 +147,26 @@ def getBlockReplicasAndSize(datasets, phedexUrl, group=None):
     for row in data:
         dataset = row['url'].split('=')[-1]
         if row['data'] is None:
-            print("FAILURE: getBlockReplicasAndSize for %s. Error: %s %s" % (dataset,
-                                                                            row.get('code'),
-                                                                            row.get('error')))
+            print("Failure in getBlockReplicasAndSize for dataset %s. Error: %s %s" % (dataset,
+                                                                                       row.get('code'),
+                                                                                       row.get('error')))
+            dsetBlockSize.setdefault(dataset, None)
             continue
         rows = json.loads(row['data'])
         dsetBlockSize.setdefault(dataset, {})
-        for item in rows['phedex']['block']:
-            block = {item['name']: {'blockSize': item['bytes'], 'locations': []}}
-            for repli in item['replica']:
-                if repli['complete'] == 'y' and repli['subscribed'] == 'y':
-                    if not group:
-                        block[item['name']]['locations'].append(repli['node'])
-                    elif repli['group'] == group:
-                        block[item['name']]['locations'].append(repli['node'])
-            dsetBlockSize[dataset].update(block)
+        try:
+            for item in rows['phedex']['block']:
+                block = {item['name']: {'blockSize': item['bytes'], 'locations': []}}
+                for repli in item['replica']:
+                    if repli['complete'] == 'y' and repli['subscribed'] == 'y':
+                        if not group:
+                            block[item['name']]['locations'].append(repli['node'])
+                        elif repli['group'] == group:
+                            block[item['name']]['locations'].append(repli['node'])
+                dsetBlockSize[dataset].update(block)
+        except Exception as exc:
+            print("Failure in getBlockReplicasAndSize for dataset %s. Error: %s" % (dataset, str(exc)))
+            dsetBlockSize[dataset] = None
     return dsetBlockSize
 
 
@@ -165,7 +178,8 @@ def getPileupSubscriptions(datasets, phedexUrl, group="DataOps", percentMin=99):
     :param phedexUrl: a string with the PhEDEx URL
     :param group: PhEDEx group
     :param percent_min: only return subscriptions that are this complete
-    :return: a dictionary of datasets and a list of their location
+    :return: a dictionary of datasets and a list of their location.
+    NOTE: Value `None` is returned in case the data-service failed to serve a given request.
     """
     #FIXME: implement the same logic for Rucio
     locationByDset = {}
@@ -179,15 +193,20 @@ def getPileupSubscriptions(datasets, phedexUrl, group="DataOps", percentMin=99):
     for row in data:
         dataset = row['url'].rsplit('=')[-1]
         if row['data'] is None:
-            print("FAILURE: getPileupSubscriptions for %s. Error: %s %s" % (dataset,
-                                                                           row.get('code'),
-                                                                           row.get('error')))
+            print("Failure in getPileupSubscriptions for dataset %s. Error: %s %s" % (dataset,
+                                                                                      row.get('code'),
+                                                                                      row.get('error')))
+            locationByDset.setdefault(dataset, None)
             continue
         rows = json.loads(row['data'])
         locationByDset.setdefault(dataset, [])
-        for item in rows['phedex']['dataset']:
-            for subs in item['subscription']:
-                locationByDset[dataset].append(subs['node'])
+        try:
+            for item in rows['phedex']['dataset']:
+                for subs in item['subscription']:
+                    locationByDset[dataset].append(subs['node'])
+        except Exception as exc:
+            print("Failure in getPileupSubscriptions for dataset %s. Error: %s" % (dataset, str(exc)))
+            locationByDset[dataset] = None
     return locationByDset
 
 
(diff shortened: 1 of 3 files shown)


amaltaro:
Valentin, Todor, I'm still running some tests, but things are looking okay.

Please review this PR and feel free to suggest a better/smarter way to avoid some copy/pasting as done in this PR. I failed to come up with one because each bulk call requires a different verification...

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -247,27 +266,36 @@ def getFileLumisInBlock(blocks, dbsUrl, validFileOnly=1):
...
-            dataset = item['this_block_name'].split("#")[0]
+        try:
+            if dataset in parentsByBlock and parentsByBlock[dataset] is None:
+                # then one of the block calls has failed, keep it failed!
+                continue
             parentsByBlock.setdefault(dataset, {})
-            parentsByBlock[dataset].setdefault(item['this_block_name'], set())
-            parentsByBlock[dataset][item['this_block_name']].add(item['parent_block_name'])
+            for item in rows:
+                parentsByBlock[dataset].setdefault(item['this_block_name'], set())
```
vkuznet: This can be one line, e.g.
```
parentsByBlock[dataset].setdefault(item['this_block_name'], set()).add(item['parent_block_name'])
```

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -139,21 +147,26 @@ def getBlockReplicasAndSize(datasets, phedexUrl, group=None):
...
-            block = {item['name']: {'blockSize': item['bytes'], 'locations': []}}
-            for repli in item['replica']:
-                if repli['complete'] == 'y' and repli['subscribed'] == 'y':
-                    if not group:
-                        block[item['name']]['locations'].append(repli['node'])
-                    elif repli['group'] == group:
-                        block[item['name']]['locations'].append(repli['node'])
-            dsetBlockSize[dataset].update(block)
+        try:
+            for item in rows['phedex']['block']:
```
vkuznet: Alan,
this block will consume lots of RAM since you have nested dict allocations all over, i.e. dsetBlockSize is a dict which holds block dict which holds yet another (attribute) dicts.

Instead, you can flat it out to simplify the data management. For example, dsetBlockSize can be a flat dict which holds only list of blocks for dataset key. Then you'll have separate flat dicts one to make block-size association and another for block-location one. Therefore instead of one nested structure you'll have 3 flat one. Moreover you can introduce a caching layer for all of them since eventually a dataset and blocks will be fixed. Since you're operating over and over with dataset/blocks your cache will be quite stable. If you want to use this route I suggest that you create separate MSCache class to handle the caches (and again it can be a separate MS micro-service).

vkuznet (commented):
Codebase is fine, but I provided my comments about further optimization through MSCache with flat dict structures.

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: Valentin, just to make I understood your comment. You suggest to replace the single dict structured like:
```
{"dataset_A": {"blockSize": 123, "locations": [pnn1, pnn2]},
                      {"blockSize": 434, "locations": [pnn2, pnn3]},
 "dataset_B": {...},
},
```

by 3 data structures. First with a list of blocks by dataset:
```
{"dataset_A": [block1, block2],
 "dataset_B": [block3, block5]}
```
then size per block
```
{"block1": 1234,
 "block2": 454}
```
and finally 
```
{"block1": [pnn1, pnn2],
 "block2": [pnn2, pnn3, ...]}
```

This makes us repeat the block name 3 times (well, original + 2 repetitions). While the current approach doesn't repeat it. I totally agree though that it's much cleaner to deal with it in different data structures, but I really don't see how it will save memory consumption.

What we could do is to create a class and constrain it with slots, that should save us some (good?) amount of memory RAM. However, we can't make it generic, thus it would require a different class for each data structure.

For the caching, that's a good point. However, depending on the workflows that we are dealing with, I fear that the memory footprint might be quite high. Keeping a cache around will just make it worse.

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: That line is too ugly to my taste ;)

todor-ivanov (approved):
I cannot add anything more than Valentin's comments here.

src/python/WMCore/MicroService/Unified/Common.py
vkuznet: Alan, yes I'm against nested using dicts, especially when there is NO ANY documentation about their schema.

The idea of using dedicated class with slots to hold block info is actually not to bed. In fact, it will eliminate storing block names 3 times such that you will only need one flat dict {dataset: [BlockObject, BlockObject]}. The BlockObject will hold block name, size and list of sites. The usage of cache is up to you, but I bet having it around may be useful.

amaltaro: Check the docstring! ;)
Let me see how much extra work such object will give, if it doesn't make the rest of the logic too hard, I'll push it in.

amaltaro:
So, before proceeding with changes to the code, I decided to create a simple script to simulate the same amount of datasets/blocks we would have in MS. Here is the source code using a class with slots versus standard dict objects:
https://github.com/amaltaro/ProductionTools/blob/master/testMemory.py

and the results is an improvement of ~7% with class+slots, for 1k datasets each with 100 blocks in it. 

Here is their memory report:
```
(PY2) AlanNewMacPro:ProductionTools amaltaro$ cat dicts.log 
Filename: testMemory.py

Line #    Mem usage    Increment   Line Contents
================================================
    53     25.8 MiB     25.8 MiB   @profile(stream=profileFp)
    54                             def createDicts(numDsets):
    55                                 """Create all the necessary objects and submit to condor"""
    56     25.8 MiB      0.0 MiB       data = {}
    57    107.2 MiB      0.0 MiB       for i in range(numDsets):
    58    107.1 MiB      0.0 MiB           dataset = randomString(100)
    59    107.1 MiB      0.0 MiB           data.setdefault(dataset, {})
    60                                     # 100 blocks
    61    107.2 MiB      0.0 MiB           for b in range(100):
    62    107.2 MiB      0.0 MiB               blockName = randomString(125)
    63    107.2 MiB      0.0 MiB               block = {blockName: {"blockSize": random.randint(1e5, 1e7), "locations": []}}
    64    107.2 MiB      0.0 MiB               for site in range(3):
    65    107.2 MiB      0.0 MiB                   block[blockName]["locations"].append(randomString(20))
    66    107.2 MiB      0.0 MiB               data[dataset].update(block)
    67    107.2 MiB      0.0 MiB       print("%s dataset dicts created" % numDsets)
```

versus slots
```
(PY2) AlanNewMacPro:ProductionTools amaltaro$ cat slots.log 
Filename: testMemory.py

Line #    Mem usage    Increment   Line Contents
================================================
    34     38.5 MiB     38.5 MiB   @profile(stream=profileFp)
    35                             def createSlots(numDsets):
    36                                 """Create all the necessary objects and submit to condor"""
    37     38.5 MiB      0.0 MiB       data = {}
    38    114.8 MiB      0.0 MiB       for i in range(numDsets):
    39    114.8 MiB      0.0 MiB           dataset = randomString(100)
    40    114.8 MiB      0.0 MiB           data.setdefault(dataset, [])
    41                                     # 100 blocks
    42    114.8 MiB      0.0 MiB           for b in range(100):
    43    114.8 MiB      0.0 MiB               blockName = randomString(125)
    44    114.8 MiB      0.0 MiB               locations = []
    45    114.8 MiB      0.0 MiB               for site in range(3):
    46    114.8 MiB      0.0 MiB                   locations.append(randomString(20))
    47    114.8 MiB      0.0 MiB               block = MSBlock(blockName, random.randint(1e5, 1e7), locations)
    48    114.8 MiB      0.0 MiB               data[dataset].append(block)
    49    114.8 MiB      0.0 MiB       print("%s dataset objects created" % numDsets)
```

Unless you can point me a mistake in that script (besides block no longer being a key), I think this improvement will have to wait a little longer given the small improvement.

amaltaro:
test this please

amaltaro:
There are all sorts of network issues today. Given that the last test was successful, and the changes I made today only: remove one comment from the src code; and added a new unit test. I'm merging it to hand it to Imran.

PR #9542, opened 2020-02-13
Consider the RelVal phedex group for pileup dataset location

amaltaro:
Fixes #<GH_Issue_Number>

#### Status
not-tested

#### Description
RelVal workflows rely on data subscribed under the `RelVal` PhEDEx group. MicroServices does the pileup location against only the `DataOps` group.

We should make it configurable for the next month, but for now hard-wiring it will let the RelVal workflows go through.

We don't really patch CMSWEB services, instead we cut a new release, build a new RPM, test it in testbed and only then we push hot fixes to production. However, given that it's a new service, I'm trying to discuss it with @muhammadimranfarooqi to see if we can patch `reqmgr2ms` on vocms0742 instead...

yes

#### Related PRs
no

no

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -186,7 +186,7 @@ def getPileupSubscriptions(datasets, phedexUrl, group="DataOps", percentMin=99):
     if not datasets:
         return locationByDset
 
-    url = "%s/subscriptions?group=%s&percent_min=%s&dataset=%s"
+    url = "%s/subscriptions?group=%s&group=RelVal&percent_min=%s&dataset=%s"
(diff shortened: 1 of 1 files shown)


amaltaro:
For the record, we have applied this change to the vocms0742:reqmgr2ms

A proper fix needs to be implemented though, but this patch is not supposed to change anymore, actually, I think we could even merge it.

amaltaro:
@todor-ivanov as we just discussed, this needs to go in ASAP. It needs to get deployed to testbed and the upcoming production cmsweb cluster. Sorry for getting late to it

todor-ivanov:
Hi @amaltaro,  
Merging and cutting a new tag now.

PR #9552, opened 2020-03-12
Correct reference to blocksize in PrimaryBlocks && Catch exception while calling makeTransferRequest in MSTrasferor

todor-ivanov:
Fixes #9551

#### Status
not-tested

#### Description
Fixing the bug with the wrong reference to the 'blockSize' when trying to get PrimaryBlocks in Workflow.py and also catching any exception that  may be thrown while calling makeTransferRequest in MSTransferor.py. The exception is handled in a per workflow basis and only for that function call, so that the cycle is not interrupted and also later in the same iteration the success/failure counters could be updated.    

#### Is it backward compatible (if not, which system it affects?)
YES

#### Related PRs
None

#### External dependencies / deployment changes
None

PR #9560, opened 2020-03-16
Fix MSTransferor logic for block distribution

### LINKED ISSUE #9551: MSTransferor interrupted due to bad datastructure reference
**Impact of the bug**
reqmgr2ms

**Describe the bug**
We started seeing those errors [6] in the MSTransferor component. The reason is a broken reference to a value in the following data structure [1]. This happens in this line here [2]. For how it should be done see the next section. 

The second part of this bug is that the exception is not caught and the cycle is interrupted whenever it hits. In the current situation this happens in the very first workflow from every batch of 100 workflows because of the bad reference to one of the fields in the data structure, but even after this is fixed, if a future exception of the kind is not handled properly here, a malformed value returned by the external service (in this case Phedex) can still lead to interruption of the work of the internal component (in this case MSTransferor).

The third part of the bug is that it also prevents the transition of the workflow from 'Assigned' to 'Staging', because this is supposed to be happening in the same component too. The status transition is called here [3], but the chain breaks and the MSTransferor is interrupted just few lines above - here[4]. Luckily we still keep the code of the old function that was used to do that transition [5] and we can revert to it for the moment (the switch is configurable). 

[1]
https://github.com/dmwm/WMCore/blob/f7e830a573d50fb1d7240797f18d809f994b934d/src/python/WMCore/MicroService/DataStructs/Workflow.py#L221-L222

[2]
https://github.com/dmwm/WMCore/blob/f7e830a573d50fb1d7240797f18d809f994b934d/src/python/WMCore/MicroService/DataStructs/Workflow.py#L333

[3]
https://github.com/dmwm/WMCore/blob/f7e830a573d50fb1d7240797f18d809f994b934d/src/python/WMCore/MicroService/Unified/MSTransferor.py#L189-L190
https://github.com/dmwm/WMCore/blob/f7e830a573d50fb1d7240797f18d809f994b934d/src/python/WMCore/MicroService/Unified/MSCore.py#L63-L74

[4]
https://github.com/dmwm/WMCore/blob/f7e830a573d50fb1d7240797f18d809f994b934d/src/python/WMCore/MicroService/Unified/MSTransferor.py#L183

[5]
https://github.com/dmwm/WMCore/pull/9508
https://github.com/dmwm/WMCore/pull/9508/commits/372dafdf11d5f8b26cc4112ae9b241934058a1a1

[6]
```
2020-03-11 20:18:20,918:INFO:MSTransferor: Handling data subscriptions for request: pdmvserv_task_HIG-RunIISummer16NanoAODv6-02991__v1_T_200305_220949_2007
2020-03-11 20:18:20,919:INFO:MSTransferor:   final list of PSNs to be use: [u'T1_DE_KIT', u'T1_ES_PIC', u'T1_FR_CCIN2P3', u'T1_RU_JINR', u'T1_UK_RAL', u'T1_US_FNAL', u'T2_BE_IIHE', u'T2_CH_CERN', u'T2_DE_DESY', u'T2_DE_RWTH', u'T2_ES_CIEMAT', u'T2_FR_GRI
F_IRFU', u'T2_FR_GRIF_LLR', u'T2_FR_IPHC', u'T2_IT_Bari', u'T2_IT_Legnaro', u'T2_UK_London_Brunel', u'T2_UK_SGrid_RALPP', u'T2_US_Caltech', u'T2_US_MIT', u'T2_US_Nebraska', u'T2_US_Purdue', u'T2_US_UCSD', u'T2_US_Vanderbilt', u'T2_US_Wisconsin']
2020-03-11 20:18:20,919:INFO:MSTransferor: List of out-of-space RSEs dropped for 'pdmvserv_task_HIG-RunIISummer16NanoAODv6-02991__v1_T_200305_220949_2007' is: set([u'T1_ES_PIC_Disk', u'T2_US_UCSD', u'T2_CH_CERN', u'T1_FR_CCIN2P3_Disk', u'T2_US_Wisconsin'
, u'T2_DE_DESY', u'T2_DE_RWTH', u'T2_FR_GRIF_IRFU', u'T2_US_Caltech', u'T1_DE_KIT_Disk', u'T2_UK_London_Brunel', u'T2_US_Purdue', u'T1_RU_JINR_Disk', u'T2_UK_SGrid_RALPP', u'T2_US_Vanderbilt', u'T2_US_Nebraska', u'T1_UK_RAL_Disk', u'T1_US_FNAL_Disk'])
2020-03-11 20:18:20,965:ERROR:MSManager: MS daemon error: unsupported operand type(s) for +: 'int' and 'dict'
  File "/data/srv/beHG2003k/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.1.pre3/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSManager.py", line 42, in daemon
    func(reqStatus)
  File "/data/srv/beHG2003k/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.1.pre3/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSManager.py", line 119, in transferor
    res = self.msTransferor.execute(reqStatus)
  File "/data/srv/beHG2003k/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.1.pre3/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 183, in execute
    success, transfers = self.makeTransferRequest(wflow)
  File "/data/srv/beHG2003k/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.1.pre3/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 319, in makeTransferRequest
    for blocks, dataSize, idx in self._decideDataDestination(wflow, dataIn, len(nodes)):
  File "/data/srv/beHG2003k/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.1.pre3/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 465, in _decideDataDestination
    listBlockSets, listSetsSize = wflow.getChunkBlocks(numNodes)
  File "/data/srv/beHG2003k/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.1.pre3/lib/python2.7/site-packages/WMCore/MicroService/DataStructs/Workflow.py", line 333, in getChunkBlocks
    chunkSize = sum(item[1] for item in sortedPrimary) // numChunks
TypeError: unsupported operand type(s) for +: 'int' and 'dict'

```

**How to reproduce it**

**Expected behavior**

Instead of trying to fetch and sum the value like that:
```
chunkSize = sum(item[1] for item in sortedPrimary)
``` 
it should be referred like that:
```
chunkSize = sum(item[1]['blockSize'] for item in sortedPrimary)
```

Also the exception needs to be caught and handled properly so the rest of the cycle is not interrupted.

amaltaro:
Fixes #9551 
Superseeds #9552 

#### Status
tested

#### Description
It contains a few bug fixes, like:
* a bug reading the data structure (accessing a dict instead of `blockSize`)
* bug dealing with parent x child relationship
* decrease number of chunks (sites) when there are less blocks than sites to distribute data
* fix bug where transfer IDs might be skipped from the transfer doc (when a chunk has 0 blocks)

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -291,9 +291,9 @@ class Workflow(object):
             for parent in list(parents):
                 if parent not in self.getParentBlocks():
                     # then drop this block
-                    parents.remove()
+                    parents.remove(parent)
             self.childToParentBlocks[child] = blocksDict[child]
-            parentBlocks = parentBlocks | blocksDict[child]
+            parentBlocks = parentBlocks | set(blocksDict[child])
 
         # Now remove any parent block that don't need to be transferred
         for block in list(self.getParentBlocks()):
@@ -330,7 +330,12 @@ class Workflow(object):
 
         # create a descendant list of blocks according to their sizes
         sortedPrimary = sorted(self.getPrimaryBlocks().items(), key=operator.itemgetter(1), reverse=True)
-        chunkSize = sum(item[1] for item in sortedPrimary) // numChunks
+        if len(sortedPrimary) < numChunks:
+            msg = "There are less blocks than chunks to create. "
+            msg += "Reducing numChunks from %d to %d" % (numChunks, len(sortedPrimary))
+            self.logger.info(msg)
+            numChunks = len(sortedPrimary)
+        chunkSize = sum(item[1]['blockSize'] for item in sortedPrimary) // numChunks
 
         self.logger.info("Found %d blocks and the avg chunkSize is: %s GB",
                          len(sortedPrimary), gigaBytes(chunkSize))
@@ -347,20 +352,26 @@ class Workflow(object):
                 if not sortedPrimary or idx >= len(sortedPrimary):
                     # then all blocks have been distributed
                     break
-                elif thisChunkSize + sortedPrimary[idx][1] <= chunkSize:
+                elif not thisChunkSize:
+                    # then this site/chunk is empty, assign a block to it
                     thisChunk.add(sortedPrimary[idx][0])
-                    thisChunkSize += sortedPrimary[idx][1]
+                    thisChunkSize += sortedPrimary[idx][1]['blockSize']
+                    sortedPrimary.pop(idx)
+                elif thisChunkSize + sortedPrimary[idx][1]['blockSize'] <= chunkSize:
+                    thisChunk.add(sortedPrimary[idx][0])
+                    thisChunkSize += sortedPrimary[idx][1]['blockSize']
                     sortedPrimary.pop(idx)
                 else:
                     idx += 1
-            blockChunks.append(thisChunk)
-            sizeChunks.append(thisChunkSize)
+            if thisChunk:
+                blockChunks.append(thisChunk)
+                sizeChunks.append(thisChunkSize)
 
         # now take care of the leftovers... in a round-robin style....
         while sortedPrimary:
             for chunkNum in range(numChunks):
                 blockChunks[chunkNum].add(sortedPrimary[0][0])
-                sizeChunks[chunkNum] += sortedPrimary[0][1]
+                sizeChunks[chunkNum] += sortedPrimary[0][1]['blockSize']
                 sortedPrimary.pop(0)
                 if not sortedPrimary:
                     break
@@ -384,7 +395,7 @@ class Workflow(object):
             # of blocks within the chunk and update the chunk size as well
             blockChunks[chunkNum].update(parentSet)
             for parent in parentSet:
-                sizeChunks[chunkNum] += parentsSize[parent]
+                sizeChunks[chunkNum] += parentsSize[parent]['blockSize']
         self.logger.info("Created %d primary+parent data chunks out of %d chunks",
                          len(blockChunks), numChunks)
         self.logger.info("    with chunk size distribution: %s", sizeChunks)
--- a/src/python/WMCore/MicroService/Unified/MSCore.py
+++ b/src/python/WMCore/MicroService/Unified/MSCore.py
@@ -69,7 +69,7 @@ class MSCore(object):
                 self.logger.info('%s updating %s status to: %s', prefix, reqName, reqStatus)
                 self.reqmgr2.updateRequestStatus(reqName, reqStatus)
             else:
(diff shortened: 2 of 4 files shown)


amaltaro:
test this please

src/python/WMCore/MicroService/DataStructs/Workflow.py
```diff
@@ -347,9 +352,14 @@ def getChunkBlocks(self, numChunks=1):
...
                     # then all blocks have been distributed
                     break
-                elif thisChunkSize + sortedPrimary[idx][1] <= chunkSize:
+                elif not thisChunkSize:
+                    # then this site/chunk is empty, assign a block to it
                     thisChunk.add(sortedPrimary[idx][0])
-                    thisChunkSize += sortedPrimary[idx][1]
+                    thisChunkSize += sortedPrimary[idx][1]['blockSize']
+                    sortedPrimary.pop(idx)
+                elif thisChunkSize + sortedPrimary[idx][1]['blockSize'] <= chunkSize:
```
todor-ivanov: IIUC here you are iterating through the sortedPrimary list (until it's exhausted) but you are making the checks towards the chunkSize - basically you check how well you are filling a second list. Did you check if under some strange condition with some weird block size distribution, we do not end up with some leftovers - I mean not being able to fit in all the blocks into the available chunks?

todor-ivanov (commented):
Only one comment inside the code. If you think we have nothing to worry about that, then all the rest looks perfect to me.

src/python/WMCore/MicroService/DataStructs/Workflow.py
amaltaro: Yes, I check against the target chunk size (`chunkSize`), while we find blocks that fit in there, we append blocks to the same site.

About the leftovers, that's a good point and I did foresee such cases. Logic is a few lines below (# now take care of the leftovers...)

amaltaro:
I've done something that should be done since the beginning, implement a few unit tests for some of those methods. With that, I managed to spot two more issues. I'm going to test it all again.

amaltaro:
test this please

amaltaro:
This is good to go from my side.

todor-ivanov (approved):
I think It is good to go.

PR #9566, opened 2020-03-18
Use absolute numbers for lumis and completion in the workflow completion

### LINKED ISSUE #9565: Fix rounding on script to check workflow completion
**Impact of the bug**
ad-hoc scripts in WMAgent

**Describe the bug**
This does not affect WMAgent at all, it's just one of those scripts we can use for a daily debugging.
When a workflow is processing a reasonable amount of lumi sections, the rounding of lumi completion can be misleading.

**How to reproduce it**
none

**Expected behavior**
Increase the decimal digits rounded when calculating the lumi completion; round down if needed.

none

amaltaro:
Fixes #9565 

#### Status
ready

#### Description
Print absolute number of lumi sections and the real division result.

yes

#### Related PRs
none

none


todor-ivanov (approved):
it is ok

PR #9540, opened 2020-02-11
New REST API parentlocks; new WMStats CP thread for resolving parentage

amaltaro:
Fixes #7651 

#### Status
In development

#### Description
This PR provides the following:
* a new wmstatsserver REST API `parentlocks`, which should return a list of active parent datasets in the system
* a new WMStats CherryPy thread `BuildParentLock`. Responsable for parsing the active data cached in WMStats, and running the DBS parentage discovery
* an extra `DataCache` property to store a list of parent datasets, to be served through the REST `parentlocks`

Basic test looks ok, e.g.:
https://alancc7-cloud2.cern.ch/wmstatsserver/data/parentlocks

What deserves extra work are those corner cases where the data cache might be empty.
In other words, for WMStats unavailability and/or redeployment, there is a small time window that we would not have data populated in the cache, thus yielding an empty result for either the `parentlocks` or the `globallocks` REST API.

In addition to that, I also _think_ we should extend the `globallocks` to consider workflows sitting in `assignment-approved` status.

#### Is it backward compatible (if not, which system it affects?)
No, it's new!

#### Related PRs
none

#### External dependencies / deployment changes
It requires these deployment changes:
https://github.com/dmwm/deployment/pull/850

PR #9567, opened 2020-03-18
Move parentlocks API from reqmon to reqmgr2

### LINKED ISSUE #9561: Move parent dataset lock list to reqmgr aux db
**Impact of the new feature**
ReqMgr2, WMStats

To build a list of parent datasets to lock, we are running a parent lock cherrypy thread on multiple backend servers. Some failure scenarios (mainly with DBS queries) could create cases where the parent locks are different across each backend. 

Move the cherrypy thread to a single backend. The thread will update and store a parent lock list in the reqmgr auxiliary database. The WMStats `parentlocks` API will query the single copy that exists in aux db.

We considered some, but this is the best option.

Need to revert this deployment change: https://github.com/dmwm/deployment/pull/867

goughes:
Fixes #9561 

#### Status
not-tested

#### Description
This PR moves the parentlocks API from wmstats to reqmgr2. This allows us to keep track of a single list of parent datasets, instead of tracking and serving the list in multiple backends.

YES

#### Related PRs
Reference and some documentation: https://github.com/dmwm/WMCore/wiki/Data-Locking-in-WMCore
#9540 

Requires this deployment update: https://github.com/dmwm/deployment/pull/876


src/python/WMCore/WMStats/CherryPyThreads/BuildParentLock.py
```diff
@@ -28,14 +30,19 @@ def setConcurrentTasks(self, config):
...
         """
         # use this boolean to signal whether there were datasets that failed
         # to get their parentage resolved
         incompleteParentage = False
 
         setDsets = set()
         setParents = set()
+        dictParents = {}
         self.logger.info("Executing parent lock cherrypy thread")
+        if DataCache.isEmpty():
```
goughes: When reqmon starts, the data cache is not populated by the time the build parent lock thread runs. If the cache is empty, the thread will overwrite whatever is in the aux db with a blank list. This is bad. Bailing out of the thread if the cache is empty can prevent this, but there might be another way.

goughes:
Some error scenarios I have tested:

For the BuildParentLock thread, if couchdb is down then 503 error is returned and the thread runs again the next cycle. When couchdb comes back the updates to the parent locks document proceed as normal.

For couchdb, if the document is missing in couchdb a 404 error is returned and if couchdb is down a 500 error is returned.

src/python/WMCore/ReqMgr/Service/Auxiliary.py
```diff
@@ -317,11 +318,34 @@ class UnifiedConfig(AuxBaseAPI):
...
+        """
+        Return entire self.name document
+        subName is subcategory of document which is added as postfix string
+        """
+        try:
+            sw = self.reqmgr_aux_db.document(self.name)
+        except CouchNotFoundError:
+            raise NoSuchInstance
+
+        return rows(sw['parentlocks'])
```
goughes: I might need a check to see if this 'parentlocks' key exists to be safe.

src/python/WMCore/WMStats/CherryPyThreads/BuildParentLock.py
```diff
@@ -52,13 +59,15 @@ def fetchIncludeParentsRequests(self, config):
...
 
         if not incompleteParentage:
-            DataCache.setParentDatasetList(list(setParents))
-            self.logger.info("Parentage lookup complete and cache renewed")
+            dictParents['parentlocks'] = list(setParents)
+            self.reqmgrAux.updateParentLocks(dictParents)
+            self.logger.info("Parentage lookup complete and auxiliary database updated")
         else:
             # then don't replace any data for the moment, simply add new parents
             previousData = set(DataCache.getParentDatasetList())
```
amaltaro: This should actually be something like `self.reqmgrAux.getParentLocks()`, thus this client get method needs to be implemented as well.
DataCache object will only be used to filter out workflows requiring IncludeParents, nothing else.

amaltaro (changes requested):
Erik, I have made some comments along the code.

I'm actually most concerned about the usage of the DataCache object. If we make this cherrypy thread a reqmgr2 (instead of wmstats) thread, then it won't be attached to a process populating the DataCache (given that that's a task of the wmstats process/application).

Perhaps, we could actually populate the DataCache in this reqmgr2 CP thread, and then resolve the parentage (all in the same cycle). Please let me know if you have another suggestion.

src/python/WMCore/WMStats/CherryPyThreads/BuildParentLock.py
goughes: Oof, yeah. Missed that line completely.

src/python/WMCore/ReqMgr/Service/Auxiliary.py
goughes: If we want to simply return a list of datasets and not the whole couch document, it is not obvious to me how to do that with the get method from the super class.

goughes:
> populate the DataCache in this reqmgr2 CP thread

Do you mean populate it and resolve the parents in an existing thread like AuxCacheUpdateTasks? Or make a new reqmgr2 CP thread for this?

src/python/WMCore/ReqMgr/Service/Auxiliary.py
amaltaro: I think it's fine to return the whole document. Even better if the document contains a `timestamp` field, such that one can know when the document is stale.

amaltaro:
Erik, sorry for giving you extra work with the specification changes. Please have a look here:
https://github.com/dmwm/WMCore/issues/9561#issuecomment-601608062
I think this makes much more sense. Please let me know what your thoughts are.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

src/python/WMCore/ReqMgr/CherryPyThreads/BuildParentLock.py
```diff
@@ -0,0 +1,102 @@
...
+        filterKeys = ['assignment-approved', 'assigned', 'staging', 'staged', 'acquired',
+                      'running-open', 'running-closed', 'force-complete', 'completed']
+        results = reqDB._getCouchView("requestsincludeparents", {}, filterKeys)
+        for row in results["rows"]:
+            dataset = row["value"]
+            setDsets.add(dataset)
+
+        # check to see if any changes have been made
+        # this will be true where new datasets are added and old datasets are no longer needed
+        if setDsets ^ set(self.datasetToParentMap.keys()):
```
amaltaro: It's more like a taste comment, but isn't it more clear to use different `!=`?

src/python/WMCore/ReqMgr/CherryPyThreads/BuildParentLock.py
```diff
@@ -0,0 +1,102 @@
...
+        dictParents = {}
+
+        self.logger.info("Executing parent lock cherrypy thread")
+
+        # query couch view to find datasets for workflows requiring parent datasets
+        # only returning requests with the statuses in filterKeys
+        reqDB = RequestDBReader(config.reqmgrdb_url)
+        filterKeys = ['assignment-approved', 'assigned', 'staging', 'staged', 'acquired',
+                      'running-open', 'running-closed', 'force-complete', 'completed']
+        results = reqDB._getCouchView("requestsincludeparents", {}, filterKeys)
```
amaltaro: should we add a try/except block here, in case couch is down at that moment.

src/python/WMCore/ReqMgr/CherryPyThreads/BuildParentLock.py
```diff
@@ -0,0 +1,102 @@
...
+
+        setDsets = set()
+        setParents = set()
+        dictParents = {}
+
+        self.logger.info("Executing parent lock cherrypy thread")
+
+        # query couch view to find datasets for workflows requiring parent datasets
+        # only returning requests with the statuses in filterKeys
+        reqDB = RequestDBReader(config.reqmgrdb_url)
```
amaltaro: How about moving this initialization to the `__init__` method?

amaltaro: same for `filterKeys` below

src/python/WMCore/ReqMgr/CherryPyThreads/BuildParentLock.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.logger.info("Found %d unique datasets requiring the parent dataset", len(setDsets))
+        if auxDbUpdateRequired:
+            self.logger.info("Found new parent dataset locks to update.")
+            # reset dataset to parent dataset map
+            self.datasetToParentMap = {}
+            # look up parent datasets via DBS
+            for dset in setDsets:
+                try:
+                    res = self.dbs.listDatasetParents(dset)
```
amaltaro: This is actually the call I planned to have cached in memory. Parent dataset will not change, so we resolve it once and use it from the cache while the process exists.
Which also means L68 (resetting the map variable) should be removed, such that we "never" cleanup that structure. Memory footprint should be negligible.

src/python/WMCore/ReqMgr/CherryPyThreads/BuildParentLock.py
```diff
@@ -0,0 +1,102 @@
...
+
+            if not incompleteParentage:
+                dictParents['parentlocks'] = list(setParents)
+                self.reqmgrAux.updateParentLocks(dictParents)
+                self.logger.info("Parentage lookup complete and auxiliary database updated.")
+            else:
+                # then don't replace any data for the moment, simply add new parents
+                previousData = self.reqmgrAux.getParentLocks()
+                # check to see if response from aux db has been populated
+                if 'parentlocks' in previousData[0]:
```
amaltaro:
```suggestion
                if previousData and 'parentlocks' in previousData[0]:
```

src/python/WMCore/ReqMgr/CherryPyThreads/BuildParentLock.py
```diff
@@ -0,0 +1,102 @@
...
+                    incompleteParentage = True
+                    continue
+                self.logger.info("Resolved parentage for: %s", res)
+                if res:
+                    setParents.add(res[0]['parent_dataset'])
+                    self.datasetToParentMap[dset] = res[0]['parent_dataset']
+
+            if not incompleteParentage:
+                dictParents['parentlocks'] = list(setParents)
+                self.reqmgrAux.updateParentLocks(dictParents)
```
amaltaro: Shouldn't it be something like:
```
if self.reqmgrAux.updateParentLocks(dictParents):
    self.logger.info("Parentage lookup complete and auxiliary database updated.")
else:
    self.logger.warning("Failed to update parentage document. Using stale data until the next cycle.")
```
?

That function returns a boolean to say whether it succeeded or not.

amaltaro (changes requested):
Erik, sorry for the late review. I've made a few comments along the code.

amaltaro (approved):

amaltaro:
Thanks, Erik. Patch looks good now.

PR #9574, opened 2020-03-23
Cleanup unneeded transfer info couch documents

### LINKED ISSUE #9478: Implement automatic cleanup of TRANSFER documents from ReqMgrAux
**Impact of the new feature**
ReqMgr2MS

Given that every request being handled by MSTransferor will result in one transfer info document in reqmgr_aux couch db, we need to have a mechanism that automatically deletes those documents, otherwise MSTransferor/MSMonitor might suffer of performance issues in the future.

We need to decide at which status we should delete those files. For the moment, I think it will be useful to keep those documents for a bit longer in the system, such that we can get back to them if needed and check what were the transfer requests created.
Thus, I'd suggest to cleanup those documents once the workflow is done processing. I.e., if the workflow is NOT in `('staging', 'staged', 'acquired', 'running-open', 'running-closed')`, then the transfer info document can be deleted.

By the way, we should probably create a new cherrypy thread to take care of this cleanup. Running it twice a day is more than enough.

In the future, when this service is robust and stable, we can start deleting those transfer documents as soon as the workflow starts running.

**Additional context**
none

amaltaro:
Fixes #9478 

#### Status
not-tested

#### Description
Expand the ReqMgr2 `CouchDBCleanup` cherrypy thread to also clean up the transfer documents inserted into `reqmgr_auxiliary` db. That task is meant to run every 12h.
It will only delete documents if the workflow is in a status **different** than
```
["assigned", "staging", "staged", "acquired", "running-open", "running-closed"]
```

I've also added a fix to the DELETE calls, such that a `Content-Length` header get set in case there is no body data.

yes

#### Related PRs
none

Deployment changes: https://github.com/dmwm/deployment/pull/875

### DIFF
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/CouchDBCleanup.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/CouchDBCleanup.py
@@ -1,43 +1,85 @@
-'''
+"""
 Created on Aug 13, 2014
 
 @author: sryu
-'''
+"""
 from __future__ import (division, print_function)
 
-from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
+from WMCore.ACDC.CouchService import CouchService
 from WMCore.Lexicon import splitCouchServiceURL
+from WMCore.REST.CherryPyPeriodicTask import CherryPyPeriodicTask
+from WMCore.Services.ReqMgrAux.ReqMgrAux import ReqMgrAux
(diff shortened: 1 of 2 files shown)


amaltaro:
I have fixed a couple of issues and things seem to be working fine now.
Erik and/or Todor, could you please review it? Thanks

amaltaro:
I'm merging it to get a new testbed deployment this morning. Don't hesitate to leave your review though.

PR #9548, opened 2020-02-21
when userName argument is present. Use it as myproxy credential. Fix #9544

belforte:
Fixes #9544 

#### Status
ready

#### Description
use existing userName argument of Credential.py to pass a predefined myproxy credential username as alternative to creating one from the hash of the DN and the obscure myproxyAccount variable which pointed to the CRAB rest hostname.

#### Is it backward compatible (if not, which system it affects?)
YES 

#### Related PRs
None
#### External dependencies / deployment changes
None

PR #9577, opened 2020-03-25
Simplify my proxy fix 9544

### LINKED ISSUE #9544: simplify myproxy credential name
**Impact of the new feature**
only affects CRAB.

currently submission from client  needs to go through same REST host as indicated in TaskWorker config. With multiple REST hosts and TW's in docker/k8s era this is not working. Also the numerical hash of the DN makes it always difficult to tell what's happening w/o adding a real extrasecurity value.

credentail in myproxy should have a simple username, e.g. we can pass
to the WMCore Credential service a string like "CRAB" and credential name will be
CRAB_belforte etc. 

no other way is possible

**Additional context**

belforte:
Fixes #9544 

#### Status
ready

#### Description
1. if username is given in proxycfg, use it also for `logonRenewMyProxy()`
2. removes some fake error logging from `vomsExtensionRenewal`
Those functions are only used  by CRAB. You can safely merge and kindly add in next tag so that we can build a new TaskWorker with this and test in preprod.
I tested already in my VM.

YES 

#### Related PRs
#9548 was a partial fix for same issue

None

### DIFF
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -309,6 +309,42 @@ class Proxy(Credential):
 
         return uName
 
+    def getMyproxyUsernameForCRAB(self):
+        """
+        :return: username (a string) to be used when myproxy-* command is called with the option
+               -l | --username        <username> Username for the delegated proxy
+
+        CRAB (and only CRAB as of March 2020) needs to obtain from myproxy a proxy from a credential
+        uploaded by another user. So that it can submit jobs and data transfers authenticated
+        with the user proxy. For this a special credential is uploaded in myproxy server by
+        CRAB Client which is connected to a username associated to the user and has a
+        list of authorized retrievers as a list of DN's of CRAB TaskWorkers maintained in
+        central CRAB configuration.
+        The username is passed to myproxy-* command via the "-l" option
+        Two different algorithms for defining this username have been used in CRAB:
+        1. the hash of the user DN + the fqdn of the CRAB REST host
+        2. the user CERN primary account username + the _CRAB string
+        During spring 2020 CRAB migrates from 1. to 2. For a smooth migration the
+        new client needs to upload both credentials and the TW will try 2. and fall back to 1.
+        Only after all tasks submited with old client are gone from the system, can we change
+        to support only 2.
+        The reasons to change from 1. to 2. are to make the username readable (helps support) and
+        to allow using different REST hosts (helps in K8s world). The reasons for the complicated
+        recipe in 1. are unknown, aside some security by obscurity attempt.
+        The caller decides if the call to myproxy-* done by this module will use 1. or 2. via
+        the userName key in the dictionary passed as argument to Proxy() at __init__ time
+           - if the dictionary contains the key 'userName', algorithm 2 is used
+           - if the dictionary does not have it, algorithm 1. is used
+        """
+        if self.userName:
+            self.logger.debug("using %s as credential login name", self.userName)
+            username = self.userName
+        else:
+            self.logger.debug(
+                "Calculating hash of %s for credential name" % (self.userDN + "_" + self.myproxyAccount))
+            username = sha1(self.userDN + "_" + self.myproxyAccount).hexdigest()
+        return username
+
     def checkAttribute(self, proxy=None):
         """
         Check attributes from a proxy file.
@@ -372,16 +408,10 @@ class Proxy(Credential):
                 'rfc' if self.rfcCompliant else 'old', self.myproxyServer)
 
             if nokey is True:
-                if self.userName:
-                    self.logger.debug("using %s as credential login name", self.userName)
-                    credname = self.userName
-                else:
-                    self.logger.debug(
-                        "Calculating hash of %s for credential name" % (self.userDN + "_" + self.myproxyAccount))
-                    credname = sha1(self.userDN + "_" + self.myproxyAccount).hexdigest()
+                myproxyUsername = self. getMyproxyUsernameForCRAB()
                 myproxyDelegCmd = 'export GT_PROXY_MODE=%s ; myproxy-init -d -n -s %s -x -R \'%s\' -x -Z \'%s\' -l \'%s\' -t 168:00 -c %s' \
                                   % ('rfc' if self.rfcCompliant else 'old', self.myproxyServer, self.serverDN, \
-                                     self.serverDN, credname, self.myproxyValidity)
+                                     self.serverDN, myproxyUsername, self.myproxyValidity)
             elif serverRenewer and len(self.serverDN.strip()) > 0:
                 serverCredName = sha1(self.serverDN).hexdigest()
                 myproxyDelegCmd += ' -x -R \'%s\' -Z \'%s\' -k %s -t 168:00 -c %s ' \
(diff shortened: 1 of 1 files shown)


belforte:
Note: all failures reported by Jenkins are due to code parts which were not touched in this PR

src/python/WMCore/Credential/Proxy.py
```diff
@@ -628,11 +635,18 @@ def vomsExtensionRenewal(self, proxy, voAttribute='cms'):
             self.logger.error("Error while checking retrieved proxy timeleft for %s" % proxy)
             return
 
+        # timeLeft indicates how maby seconds the proxy is still valid for
```
amaltaro: there are a few typos in these comments, can you please fix them.

src/python/WMCore/Credential/Proxy.py
```diff
@@ -586,10 +586,17 @@ def logonRenewMyProxy(self, proxyFilename=None, credServerName=None):
         # Note that this is saved in a temporary file with the pid appended to the filename. This way we will avoid adding many
         # signatures later on with vomsExtensionRenewal in case of multiple processing running at the same time
         tmpProxyFilename = proxyFilename + '.' + str(os.getpid())
+
+        if self.userName:
```
amaltaro: given that I see the same logic in different places in this module, how about we create a method to return the `credname` value? Just call that method every where you need to resolve the user credentials in the same format as: `sha1(self.userDN + "_" + self.myproxyAccount).hexdigest()`

amaltaro (changes requested):
Stefano, please see the comments I have made in the code.

src/python/WMCore/Credential/Proxy.py
belforte: OK. Also moved some lines around so comments are close to code they refer to.

amaltaro:
test this please

src/python/WMCore/Credential/Proxy.py
belforte: @amaltaro I started with the idea of making minimal modifications.

I do not like to introduce a `getCredname` method for something which is not exactly
a credential or anyhow adding a method I can't properly document.
Fact is that clearly different people added code to `Proxy.py` in uncoordinated way,
at times w/o fully understanding what they were doing. Hopefully one day we
will deprecate X509 and this stuff will all go. Hoping that day to be not too far,
I tried to avoid a major overhaul.

But  now that I learnt the inside of this and you made me think more about it, what about 
the following which is a truly minimal change to WMCore Proxy ?

we use `self.userName`  from the argument dictionary passed to
`Proxy.py` as "username" whenever a `myproxy-*` command is used which requires
the `-l username` option (only CRAB use that).
And leave it to the calling CRAB code to decide if to pass a human string or a hash of the DN+ the mysterious `self.myproxyAccount` which when you read this code you have no
clue what it should be ?

AFAIK only CRAB uses `-l username` option in myproxy-* (since it is only used when an indentity switch is needed) and nobody else was using the self.userName, at least not in WMCore repository.  Will also avoid using the name `credname` since this is not really the name of a credential, actually myproxy-* has a credential name argument, via the option `-k`.
I am not sure why the original developers used `-l username` instead of `-k credname` 

This requires to modify the changes I have done to CRAB already and a new round of test from my side, so do not hold your breath... but if you like it more, I'll do.

belforte: sorry for very logg comment, let me summarize:
if I do what you suggested it will still be a non-understandable hack, which nobody but a CRAB expert can maintain/use, while saving only ~10 lines of code.
Since Proxy.py  will not be there in Run4 for sure, and hopefully gone sooner, I don't see the gain, but I can do it easily.

The alternative is more work for me, but a way to push some (but still not all) CRAB internals knowledge from Proxy.py back into CRAB land.

All in all, having slept on it, I'd rather do what you suggested as it gives me a place where to put a bunch of comments to help others to understand what's going on.

amaltaro: Stefano, your idea looks much clean indeed. However, it will also depend on how many clients/places you will have to apply such change. If not too much and you're willing to do so, please go ahead.

BTW, I don't think any of the WMCore services use this module. Even though we could, in order to renew my proxy that gets shipped with the jobs. Anyways, that's not important at the moment and I do not foresee investing time on that.

belforte: looks like comments crossed. I'go for the simplest.
You will never need to use myproxy to refresh proxy for jobs, condor does it. As long as you have a single user, you only need to keep one valid proxy in each schedd, which I think you do already with the other methods in this class.

belforte: I decided to follow initial suggestion, but use a method name which makes it clear it is a hack for CRAB and did my best to document it.
Hope you like it @amaltaro 

Do not merge yet. I need to test it now.

amaltaro:
@belforte I assume you haven't completely tested it yet, right?

belforte:
right, sorry. kid at home and all that..

belforte:
This now ready to go from my side.

belforte:
shall I remove the Do not merge yet label ?

src/python/WMCore/Credential/Proxy.py
```diff
@@ -309,6 +309,42 @@ def getUserName(self, proxy=None):
...
+        the userName key in the dictionary passed as argument to Proxy() at __init__ time
+           - if the dictionary contains the key 'userName', algorithm 2 is used
+           - if the dictionary does not have it, algorithm 1. is used
+        """
+        if self.userName:
+            self.logger.debug("using %s as credential login name", self.userName)
+            username = self.userName
+        else:
+            self.logger.debug(
+                "Calculating hash of %s for credential name" % (self.userDN + "_" + self.myproxyAccount))
```
amaltaro, line 344: I don't remember which pylint complaint is triggered here. A better logging format would be:
```
"Calculating hash of %s for credential name", self.userDN + "_" + self.myproxyAccount)
```

No need to update this patch for this line only though.

src/python/WMCore/Credential/Proxy.py
```diff
@@ -372,16 +408,10 @@ def delegate(self, credential=None, serverRenewer=False, nokey=False):
...
 
             if nokey is True:
-                if self.userName:
-                    self.logger.debug("using %s as credential login name", self.userName)
-                    credname = self.userName
-                else:
-                    self.logger.debug(
-                        "Calculating hash of %s for credential name" % (self.userDN + "_" + self.myproxyAccount))
-                    credname = sha1(self.userDN + "_" + self.myproxyAccount).hexdigest()
+                myproxyUsername = self. getMyproxyUsernameForCRAB()
```
amaltaro, line 411: please fix the extra blank space between `self` and and the method name

src/python/WMCore/Credential/Proxy.py
```diff
@@ -403,14 +433,8 @@ def getMyProxyTimeLeft(self, proxy=None, serverRenewer=False, nokey=False):
...
             if nokey is True and serverRenewer is True:
-                if self.userName:
-                    self.logger.debug("using %s as credential login name", self.userName)
-                    credname = self.userName
-                else:
-                    self.logger.debug(
-                        "Calculating hash of %s for credential name" % (self.userDN + "_" + self.myproxyAccount))
-                    credname = sha1(self.userDN + "_" + self.myproxyAccount).hexdigest()
-                checkMyProxyCmd = 'myproxy-info -l %s -s %s' % (credname, self.myproxyServer)
+                myproxyUsername = self. getMyproxyUsernameForCRAB()
```
amaltaro, line 436: same extra blank space here

amaltaro (changes requested):
Stefano, sorry for the delay on this review.
Changes look good in general, but there are a few comments in the code that deserve to be fixed.

Once you make those changes, can you also please squash all these commits into a single one?

belforte:
oh my.. how did I get those blank spaces in ?
@amaltaro, how do I squash other then while merging ? sorry..

src/python/WMCore/Credential/Proxy.py
belforte: done

belforte:
I am getting lost in git rebase and company, maybe also becasue github is having problems atm.
May I close this and make a new PR with one single commit ?

PR #9578, opened 2020-03-25
Use available secondary dataset regardless of which group its subscribed to

### LINKED ISSUE #9576: MS: skip check on PhEDEx group for secondary dataset
**Impact of the new feature**
ReqMgr2MS

Yes, we have just noticed that there are 6 workflows stuck in `assigned` status. Reason being that those workflows, e.g.:
cmsunified_task_HIN-pPb816Spring16GS-00160__v1_T_200322_043001_9388

are requesting a secondary dataset that is available at the site whitelisted:
/ReggeGribovPartonMC_EposLHC_PbP_4080_4080_DataBS/pPb816Spring16GS-MB_80X_mcRun2_pA_v4-v2/GEN-SIM

but its phedex group is `heavy-ions`.

After discussing this issue with the P&R team, over the #wmcore-support slack channel, the agreement is that the microservices should disconsider the PhEDEx group which the secondary data is subscribed to. I.e., if the dataset is available at one of the necessary sites, no matter which group, skip the data placement and move the workflow on to the next status.

The other alternative would be to change the dataset group subscription from whatever to `DataOps`.

**Additional context**
This needs to go in for the upcoming upgrade.

amaltaro:
Fixes #9576 

#### Status
not-tested

#### Description
When retrieving the secondary dataset subscription, no longer filter by PhEDEx group. Keep the functionality in place though.
This means that MSTransferor will not create any data placement requests if the pileup dataset is available under other groups, including the `local` group.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -190,7 +190,8 @@ def getPileupSubscriptions(datasets, phedexUrl, group=None, percentMin=99):
         url = "%s/subscriptions?group=%s" % (phedexUrl, group)
         url += "&percent_min=%s&dataset=%s"
     else:
-        url = "%s/subscriptions?percent_min=%s&dataset=%s" % phedexUrl
+        url = "%s/subscriptions?" % phedexUrl
+        url += "percent_min=%s&dataset=%s"
     urls = [url % (percentMin, dset) for dset in datasets]
 
     data = multi_getdata(urls, ckey(), cert())
--- a/src/python/WMCore/MicroService/Unified/Common.py
+++ b/src/python/WMCore/MicroService/Unified/Common.py
@@ -170,24 +170,29 @@ def getBlockReplicasAndSize(datasets, phedexUrl, group=None):
     return dsetBlockSize
 
 
-def getPileupSubscriptions(datasets, phedexUrl, group="DataOps", percentMin=99):
+# FIXME: implement the same logic for Rucio
+def getPileupSubscriptions(datasets, phedexUrl, group=None, percentMin=99):
     """
     Provided a list of datasets, find dataset level subscriptions where it's
     as complete as `percent_min`.
     :param datasets: list of dataset names
     :param phedexUrl: a string with the PhEDEx URL
-    :param group: PhEDEx group
+    :param group: optional string with the PhEDEx group
     :param percent_min: only return subscriptions that are this complete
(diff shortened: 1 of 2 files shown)


amaltaro:
Tested and it's good to go.

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -170,24 +170,29 @@ def getBlockReplicasAndSize(datasets, phedexUrl, group=None):
...
     if not datasets:
         return locationByDset
 
-    url = "%s/subscriptions?group=%s&group=RelVal&percent_min=%s&dataset=%s"
-    urls = [url % (phedexUrl, group, percentMin, dset) for dset in datasets]
+    if group:
+        url = "%s/subscriptions?group=%s" % (phedexUrl, group)
+        url += "&percent_min=%s&dataset=%s"
+    else:
+        url = "%s/subscriptions?percent_min=%s&dataset=%s" % phedexUrl
```
todor-ivanov: This one here shouldn't work. It must produce:
```
TypeError: not enough arguments for format string
```
And wouldn't it be better to use the `"{}".format(var)` syntax?

src/python/WMCore/MicroService/Unified/Common.py
```diff
@@ -190,7 +190,8 @@ def getPileupSubscriptions(datasets, phedexUrl, group=None, percentMin=99):
         url = "%s/subscriptions?group=%s" % (phedexUrl, group)
         url += "&percent_min=%s&dataset=%s"
     else:
-        url = "%s/subscriptions?percent_min=%s&dataset=%s" % phedexUrl
+        url = "%s/subscriptions?" % phedexUrl
+        url += "percent_min=%s&dataset=%s"
```
todor-ivanov, line 194: I see here, you have fixed that error from the previous commit. So sorry for my previous comment then. But it would be better if we squash commits like that - and then only the final version of the code will be reviewed (leaving aside the late review itself, of course - for which I can blame only myself ;) )

todor-ivanov (commented):
looks  good to me

todor-ivanov (commented):
The comment I made earlier is fixed in the next commit. So all the rest looks perfect to me.

src/python/WMCore/MicroService/Unified/Common.py
amaltaro: I agree, Todor. I overlooked the number of commits. It should have been squashed before getting merged.

PR #9593, opened 2020-03-30
Limit MSTransferor in terms of successful workflows

### LINKED ISSUE #9592: MS: do not apply priority order to workflows without any input data
**Impact of the bug**
ReqMgr2MS

**Describe the bug**
Only today I noticed that, in a bad scenario with many sites out of quota, we can hit the situation where most of the 500 workflows (set via `limitRequestsPerCycle`) will not go through in the ReqMgr2MS cycle, being retried in the next cycle. 

This will delay the other workflows sitting in `assigned` state, even if they have no input data to be placed. Reason for that is that we order workflows by priority, if high priority workflows don't manage to go to `staging`, then low prio will never be evaluated.

**How to reproduce it**
Steps can be extracted from the description above.

**Expected behavior**
Workflows without any kind of input data should not wait for the normal priority cycle to pick them up.
In other words, low priority workflows should have a chance to be evaluated as well. If they don't require any input dataset, there is no reason to keep holding them under the `assigned` status.

I see two options here:
a) every other cycle we go an extra mile and process beyond the `limitRequestsPerCycle` number of workflows, but only move forward with those without any input data.
b) every cycle, we consider `limitRequestsPerCycle` extra workflows not requiring any sort of input data.

It's not clear to me which approach will be the best (especially in terms of implementation) at this moment.

none

amaltaro:
Fixes #9592 

#### Status
not-tested

#### Description
One of the MSTransferor configuration parameters is:
`limitRequestsPerCycle`
which is used to limit how many workflows MSTransferor can deal with in each cycle.

With this patch, we extend a bit the definition of that parameter to the number of successful workflows handled by MSTransferor. Thus, always targeting to properly handle (and moving them to the next status) that amount of workflows in every cycle (currently set to 500). Avoiding workflow starvation in the MSTransferor queue.

Note, that check has been placed under the slice level. So it could be that MSTransferor successfully updates a few more workflows in a cycle. Not a big deal though.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSTransferor.py
+++ b/src/python/WMCore/MicroService/Unified/MSTransferor.py
@@ -146,6 +146,7 @@ class MSTransferor(MSCore):
         :param reqStatus: request status to process
         :return:
         """
+        counterWorkflows = 0
         counterFailedRequests = 0
         counterSuccessRequests = 0
         summary = dict(TRANSFEROR_REPORT)
@@ -155,7 +156,6 @@ class MSTransferor(MSCore):
             msg = "  retrieved %s requests. " % len(requestRecords)
             msg += "Service set to process up to %s requests per cycle." % self.msConfig["limitRequestsPerCycle"]
             self.logger.info(msg)
-            requestRecords = requestRecords[:self.msConfig["limitRequestsPerCycle"]]
         except Exception as err:  # general error
             msg = "Unknown exception while fetching requests from ReqMgr2. Error: %s", str(err)
             self.logger.exception(msg)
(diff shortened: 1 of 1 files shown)


amaltaro:
I'm planning to add it on tomorrow's CMSWEB request.
So, Erik and Todor, if it's not too late for you, could you please have a look at this PR? I still have to test it.

amaltaro:
Tests went fine as well. Feel free to leave a review if you will.

goughes (commented):
Just did a quick review and I don't see any issues here.

todor-ivanov (commented):
Looks good to me.

Issue #9600, opened 2020-04-02
Enable NANOAOD/NANOAODSIM only publication to Rucio


amaltaro:
**Impact of the new feature**
WMAgent

**Is your feature request related to a problem? Please describe.**
As discussed this week, once WMAgent has the ability to publish data to Rucio - via its RucioInjector component - we would like to start first injecting NANOAOD and NANOAODSIM datatiers into Rucio, any other data tier should keep going to PhEDEx.

**Describe the solution you'd like**
NANOAOD and NANOAODSIM datatiers data has to be injected against Rucio, anything else goes to PhEDEx. This should be configurable BTW.

We will also need to implement some logic for dataset subscription, given that some NANO will be available through PhEDEx and some in Rucio. The agreement we had with Eric and L2s is that we do not need to bother about NANO dataset subscriptions (either done via WMAgent, or ReqMgr2MS); Rucio/DM team will take care of it.

**Describe alternatives you've considered**
We still need to decide whether we will be able to patch agents (possibly ending up with part of a dataset in Rucio and part in PhEDEx). 

**Additional context**
This issue depends on: https://github.com/dmwm/WMCore/issues/9597
so, we cannot start injecting data into Rucio until we figure that meta-data information out.

amaltaro:
As discussed in https://github.com/dmwm/WMCore/pull/9617
whatever is NOT banned on the Rucio component, we need to make sure it's banned on the PhEDExInjector. There should never be the case where both components try to update the same data/row tables (even though I don't think it would be a disaster...).

PR #9617, opened 2020-04-06
Created new component RucioInjector

### LINKED ISSUE #9065: Repurpose PhEDExInjector towards the Rucio service
Or we create a brand new RucioInjector component inheriting some of the PhEDExInjector functionality. Internal logic might change quite a bit, so we might write a new component from scratch and have a switch to either use RucioInjector or PhEDExInjector.

amaltaro:
Fixes #9065 

#### Status
In development

#### Description
Created a new WMAgent component called `RucioInjector`. In more details, this PR provides:
* RucioInjector configuration section, where we can disable the component; disable block rules creation; disable rules creation for a list of datatiers; etc
* moved ALL the DAOs from the PhEDExInjector package to the RucioInjector one
* keep 24h cache for containers and blocks already injected into Rucio
* create a simple block rule (`rse=RSE & grouping=DATASET`) for every block inserted into Rucio
* retry (in the next cycle) files that failed to have their dbsbuffer state updated
* close blocks that have already been migrated to DBS
* create standard container rules as configured in the workflows. Rule is like `rse=RSE & grouping=ALL`
* added meta-data to rules creation (basic info for now)
* NOTE: we have dropped the validation of data location against valid PNNs (now RSEs)

In addition to the new component, I can list:
* implemented a Utils MemoryCache module for data caching in memory
* extended the Rucio wrapper API to also support:
  * listing the full history of rules for a given container/block
  * and a method to delete a given rule

TODO: check error handling, like an operation succeeding in Rucio but failing in dbsbuffer.
TODO: understand the full logic of block deletion and reimplement it

no

#### Related PRs
none

Yes, these deployment changes: https://github.com/dmwm/deployment/pull/885
Deployment changes available starting in `HG2005a`


amaltaro:
@ericvaandering @nsmith- Eric, Nick, would you mind having a general look at these changes.
The initial comment should list most of the changes that went in.

I have just added some meta data to rules/blocks as well. It will let a workflow running to see whether that works or not.

I still have to think how to properly implement the data deletion used by T0 though...

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,426 @@
...
+        return newBlocks
+
+    # TODO: this will likely go away once the phedex to rucio migration is over
+    def _isBlockTierAllowed(self, blockName):
+        """
+        Check the block datatier against the list of data tiers that we want
+        to get pinned in Rucio through a block rule
+        :return: boolean whether we can or not insert a rule
+        """
+        for skipTier in self.skipRulesForTiers:
```
ericvaandering, line 220: You can simplify this a bunch:
endBlock = 
endTier = 
if endTier in self.skipRules: return False

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,426 @@
...
+        for block, rse in blocksAndRSEs.items():
+            if not self._isBlockTierAllowed(block):
+                logging.info("Component configured to skip block rule for: %s", block)
+                continue
+            # FIXME TODO REMOVE hack to test RSEs
+            rseName = "%s_Test" % rse
+            # DATASET = replicates all files in the same block to the same RSE
+            resp = self.rucio.createReplicationRule(block, rseExpression="rse=%s" % rseName,
+                                                    account=self.rucioAcct, grouping="DATASET",
+                                                    comment="WMCore automatic block rule",
```
ericvaandering, line 249: Can you say something like "WMAgent production site"

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,426 @@
...
+        _deleteBlocks_
+        Find deletable blocks, then decide if to delete based on:
+        Is there an active subscription for dataset or block ?
+          If yes => set deleted=2
+          If no => next check
+        Has transfer to all destinations finished ?
+          If yes => request block deletion, approve request, set deleted=1
+          If no => do nothing (check again next cycle)
+        """
+        raise NotImplementedError
```
ericvaandering, line 341: Here you should be able to delete the rules you made on the blocks if you can find them or kept track of them. If a another rule was made to move the container to another site and it hasn't happened yet, Rucio is smart enough to know this and will not clean up the replicas

ericvaandering: In other words I don't think you even need to find the deletable blocks.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def deleteBlocksPhEDExCalls(self, location, blocksToDelete):
```
ericvaandering, line 343: I'm guessing this is just here to show logic of what we do now?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,426 @@
...
+        return
+
+    # TODO: this will likely go away once the phedex to rucio migration is over
+    def _isContainerTierAllowed(self, dsetName):
+        """
+        Check the container datatier against the list of data tiers that we want
+        to get pinned in Rucio through a container rule
+        :return: boolean whether we can or not insert a rule
+        """
+        for skipTier in self.skipRulesForTiers:
```
ericvaandering, line 378: Same comment, more idiomatic python could be used.

ericvaandering (commented):
The overall logic looks good. I think the block deletions are a question we should resolve still. And you can't put more in the metadata field because there is nothing more in "uninjectedData"?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Definitely! Let me fix it.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Good idea!

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Yes, I will address the rule tracking in a different issue. For now, I'm leaving just a placeholder for rules deletion. I have to review what triggers the T0 rule deletion as well. I believe what we will need is to just check whether the DID has any other rule on it, no?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Yes, I'm deleting it before pushing it in. Logic remains in PhEDExInjector, if needed.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: thanks!

amaltaro:
Nick/Eric, from the discussion we had this week. I have done a few tests providing meta-data and my findings are:
* Rucio CLI documentation says it has to be a python dictionary, but actually it needs to be a json string (json.dumps)
* the server returns an error (server_error) if we provide metadata during block creation
* it seems to work fine for rule creation for blocks

Fresh examples for rules with comments and metadata information are:
```
2020-04-08 22:03:51,681:139798235088640:INFO:RucioInjectorPoller:Block rule created for /DYJetsToLL_Pt-50To100_TuneCUETP8M1_13TeV-amcatnloFXFX-pythia8/DMWM_Test-StepChain_MC_April2020_Val_Privv4-v20/LHE#cb61210c-5644-42ed-96ec-0d47eec812a0 under rule id: [u'de3a83f598d7438c89843554f5f80f88']
2020-04-08 22:03:51,832:139798235088640:INFO:RucioInjectorPoller:Block rule created for /DoubleMuon/DMWM_Test-ReReco_badBlocks_April2020_Val_Privv4-v11/DQMIO#75e04dae-a7bf-4288-a189-c287281dfcd6 under rule id: [u'dc25779d41824245b9989560544538db']
```

todor-ivanov (approved):
Only one minor comment from my side in the MemoryCache.py file. For the RucioInjectorPoller.py,  Eric covered it much better than I ever could. Nothing more to add there. The rest looks good to me.

src/python/Utils/MemoryCache.py
```diff
@@ -0,0 +1,85 @@
...
+        elif isinstance(self._cache, list) and isinstance(inputItem, (list, set)):
+            # extend another list or set into a list
+            self._cache.extend(inputItem)
+        elif isinstance(self._cache, list):
+            # add a simple object (integer, string, etc) to a list
+            self._cache.append(inputItem)
+        elif isinstance(self._cache, dict) and isinstance(inputItem, dict):
+            self._cache.update(inputItem)
+        else:
+            msg = "Cache data type: %s, item to be added with type: %s" % (type(self._cache), type(inputItem))
```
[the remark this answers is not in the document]
amaltaro: Let me try to make this error message slightly more clear.

src/python/Utils/MemoryCache.py
```diff
@@ -0,0 +1,85 @@
...
+    def addItemToCache(self, inputItem):
+        """
+        Adds new item(s) to the cache, without resetting its expiration.
+        NOTE that it will only work for list, set and dictionary types.
+        :param inputItem: additional item to be added to the current cached data
+        """
+        if isinstance(self._cache, set) and isinstance(inputItem, (list, set)):
+            # extend another list or set into a set
+            self._cache.update(inputItem)
+        elif isinstance(self._cache, set):
```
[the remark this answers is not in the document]
amaltaro: That's a good comment Todor, I didn't want to add a check there because them we limit the object types to be stored in this cache (like, we can no longer store complex objects)

amaltaro:
Todor, you raised a valid concern about that cache module. We have two options here:
1) constrain the object types we can store in the cache (but make the module safer)
2) allow free flexibility for the data types, but having a user-mistake susceptible code

I decided for option 2), but giving it further thought, I might change it to 1) and if the user has an object unsupported, he has to write his own cache...

amaltaro:
@todor-ivanov from your comment, I modified MemoryCache to support only a small set of object types.

etc/WMAgentConfig.py
```diff
@@ -356,3 +356,18 @@
...
+config.component_("RucioInjector")
+config.RucioInjector.namespace = "WMComponent.RucioInjector.RucioInjector"
+config.RucioInjector.componentDir = config.General.workDir + "/RucioInjector"
+config.RucioInjector.logLevel = globalLogLevel
+config.RucioInjector.enabled = False
+config.RucioInjector.pollInterval = 300
+config.RucioInjector.pollIntervalRules = 43200
+config.RucioInjector.cacheExpiration = 24 * 60 * 60  # one day
+config.RucioInjector.createBlockRules = True
+config.RucioInjector.useTestRSE = ""  # set it to "_Test" in a test node
```
nsmith-, line 369: Maybe better to call it `.RSEPostfix` ?

nsmith-: (Or just make it boolean with the postfix handled in code)

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,390 @@
...
+        self.setStatus = daofactory(classname="DBSBufferFiles.SetPhEDExStatus")
+        self.setBlockClosed = daofactory(classname="SetBlockClosed")
+
+    @timeFunction
+    def algorithm(self, parameters):
+        """
+        _algorithm_
+
+        Poll the database for uninjected files and attempt to inject them into
+        PhEDEx.
```
nsmith-, line 116: Probably all this could use `s/phedex/rucio/i` treatment

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,390 @@
...
+
+            # now create a Rucio rule for every single block (and update local cache)
+            blocksAdded = self.insertBlockRules(blocksAdded)
+            if self.blocksCache.isCacheExpired():
+                self.blocksCache.setCache(blocksAdded)
+            else:
+                self.blocksCache.addItemToCache(blocksAdded)
+
+            # finally, create file replicas
+            self.insertReplicas(uninjectedFiles)
```
nsmith-, line 149: @ericvaandering is this how the sync works? by injecting the files *after* creating the rule? I would think its before but am not sure.

etc/WMAgentConfig.py
```diff
@@ -356,3 +356,18 @@
...
+config.RucioInjector.skipRulesForTiers = ["NANOAOD", "NANOAODSIM"]
```
nsmith-, line 370: This somehow seems backwards to me, maybe I am missing something. I expect that we *would* be making rules for injected NanoAOD and *not* for any other data tiers as those are expected to be handled by PhEDEx for now, where the sync daemon would keep phedex and rucio in agreement about where it is. (we can still inject replicas for these I think, as the sync script will just see the work it was going to do done already and not do any harm, i.e. there's a race between WMCore and sync but either one winning produces the same state)

So I would think we want something like `makeRulesForTiers = ["NANOAOD", "NANOAODSIM"]` and the list grows as we progress until it is short-circuited at some point to always true once we no longer inject anything into phedex.

nsmith-: Now, aside from all that, there is the question whether WMCore needs to make a rule at all. In theory, no since a rucio subscription can do that. However I think that ATLAS puts a rule with something like 1w lifetime on output so we might want to keep that option open.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,390 @@
...
+        If the insertion succeeds, also switch their database state to injected.
+
+        :param uninjectedData: dictionary with blocks as key, and RSEs as value
+        """
+        # FIXME: I think we need a different data struct from the database
+        # this method is very expensive O(n^4)
+        logging.info("Preparing to insert replicas into Rucio...")
+
+        for location in uninjectedData.keys():
+            rseName = location + self.testRSEs
```
nsmith-, line 264: I wonder, does the actual job wrapper have a mechanism to optionally do stageout to the test RSE? (it is a different LFN-to-PFN algo) Otherwise this function is lying to rucio about physical location of replicas.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,390 @@
...
+        _deleteBlocks_
+        Find deletable blocks, then decide if to delete based on:
+        Is there an active subscription for dataset or block ?
+          If yes => set deleted=2
+          If no => next check
+        Has transfer to all destinations finished ?
+          If yes => request block deletion, approve request, set deleted=1
+          If no => do nothing (check again next cycle)
+        """
+        # FIXME: figure out the proper logic for rule block deletion
```
nsmith-, line 333: @ericvaandering I think this is a good spot for rule chaining right? We never have to explicitly delete but simply attach the child rule so that the original rule keeping the data at its original placement site is removed.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -0,0 +1,390 @@
...
+
+            # finally, create file replicas
+            self.insertReplicas(uninjectedFiles)
+
+            # now close blocks already uploaded to DBS
+            self.closeBlocks()
+
+            if self.lastRulesExecTime + self.pollRules <= int(time.time()):
+                self.deleteBlocks()
+                self.insertContainerRules()
```
nsmith-, line 156: If we are making block rules what is the container rule for?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
ericvaandering: The Sync makes the rules after, as I recall.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
ericvaandering: I have not looked at Rule Chaining...

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
ericvaandering: Container rule is, I assume the data placement for the whole CMS dataset to put it at a single site. But I do wonder how you can do this in a single agent since this is really a systemwide thing.

etc/WMAgentConfig.py
amaltaro: I could change it, but I understand these Test RSEs are going away in a couple of weeks from now, no? @ericvaandering

amaltaro: Nick, I just had a chat with Eric and apparently we have no use case for such _Test RSEs. I'm going to remove this code.

etc/WMAgentConfig.py
amaltaro: Nick, what I had understood from our discussion was:
* WMCore will inject only NANOAOD/NANOAODSIM data into Rucio (container + block + replicas)
* and rules - for those 2 datatiers - would be sorted by the DM team, given that we won't know how much of our container (cms dataset) is/was injected into PhEDEx and how much into Rucio.

BTW, the ban on data injection is not yet in place. I have it implemented only for rules for now. I actually want to work on that front in a different PR, see this issue:
https://github.com/dmwm/WMCore/issues/9600

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Here I first create a rule because I want to make sure there is a rule protecting a block before any replicas get created.
Otherwise, I foresee a day where we could have:
* block is created, then replicas are added
* then the component crashes for the whole weekend (meaning, no rules created)
* what happens to those replicas then?!?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Eric is correct. The container rule needs to be created whenever a workflow is configured to make a cms dataset level subscription. It happens at any point in the lifetime of a workflow (usually in the first 24h of the workflow, once it goes to running-open/running-closed status).

Good comment about multiple subscriptions btw. Either PhEDExInjector has the necessary logic to see if there is already a dataset subscription for that. Or we duplicate subscriptions (harmless)

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Yes, all the tests I have ran over the last 72h were a lie :P
As mentioned in another comment, I'm removing this logic.

amaltaro:
@nsmith- Nick, thank you for your comments. I think I replied to all of them.
Just for the record, this PR is meant to provide the basic functionality of a Rucio injector component.

I still have other issues created for filtering out datatiers, massaging rule creation/deletion, etc.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
nsmith-: Actually I think it would be good to expand this and add the test site stageout function to the job wrapper. Then we could have a complete test system including all sites!

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
nsmith-: I think you should do it as sync: any replicas that have no rules are not possible to be touched by reaper: a replica must have a tombstone set to be reaped and it is only set once all rules are removed. If no rule ever is created, the replica lives on.

etc/WMAgentConfig.py
nsmith-: I don't think there is any need for a ban on data injection, its a race between the sync script that always end up with the same result. @ericvaandering thoughts?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
nsmith-: Does this only get set for a single RSE? If you make a container rule on multiple RSEs then it would duplicate the whole dataset..
But here actually we can simplify the logic a lot with rucio: the desire is to have two copies of the output data during production for resilience and this is trivial with rucio if we define a single container rule (with grouping = DATASET) for the whole workflow with an RSE expression matching "all active production sites" and a replicas count = 2. Each source block that is injected will count towards one of the two and rucio will manage placing another as appropriate.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: I can change that. Which brings me to the next question. Should I create a block rule as soon as we have 1 file/replica in the block? Or only when we are about to close the block (or add the last replica)?

etc/WMAgentConfig.py
amaltaro: There will be a moment that we will keep running PhEDExInjector and RucioInjector. Whatever one component does, the other cannot. Reason is because we keep the state of blocks/files in the local database and we cannot (should not) have two different components trying to do the same database updates.
That's why, IMO, whatever we decide to inject into Rucio, we need to NOT inject to PhEDEX - at least from WMAgent point of view.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
nsmith-: Well let me rephrase: I think this and the sync code should be consistent. I am not 100% sure which order is better, although at least we have operation experience with rule-after-replica. Things that influence the decision:

- How long it takes rucio to detect changes to DID replicas and move a rule from DONE to REPLICATING
- How a rule behaves when unsatisfiable because no replica exists

etc/WMAgentConfig.py
nsmith-: Ok that is reasonable.
For the rules, do I understand that its possible during the transition that part of a CMS dataset will be injected in PhEDEx and part in Rucio? Will we still create the Rucio container (CMS dataset) rule?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Ok, keeping the functionality doesn't hurt. You might want to open a new GH issue to expand on this test site stageout though.

nsmith-:
Ok, I didn't mean for this to snowball but its very convenient to discuss requirements with a skeleton code strawman to attack :)

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: single RSE. IF that's how we want to run production in the Rucio era, then we need to make sure there is a GH with the feature requirements.

etc/WMAgentConfig.py
amaltaro: Yes, it's possible. We have workflows that remain in running-* state in the system for many months. So it could easily have blocks created/injected against PhEDEx, while others are in a new agent version injecting things into Rucio.
Here indeed we would have a race between sync and rucio injected data. However, RucioInjector handles nicely existent containers and just move on with block/replicas injection.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
nsmith-: Maybe I don't understand this part of the code, but I thought this is called once per block update? I would expect the container to be created once for the whole workflow, is it handled here somehow once so that another block processing at a different site does not create also a container rule at that site?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Let's go with your operation experience then. I'm updating the code.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: `insertContainerRules` is called once every many cycles of the component. It also only operates on output datasets (rucio containers) that have been configured at workflow level. Once the agent has created the container level rule, we persist that information in the local database and it's no longer triggered again.

About multiple agents, that's a fair point. What happens if the same rule gets created multiple times (same rucio acct, same container, same rse expression)? I believe it will return duplicate rule.
AFAICS, we need to have a smarter logic in place. I put this issue in place:
https://github.com/dmwm/WMCore/issues/9621

feel free to discuss anything that I missed over there.

amaltaro:
No worries Nick, better now than once it's in production ;)
BTW, I have made further changes to the code as per our discussion, including the order swap of rule creation and replica creation.

nsmith-:
Ok, I regret not clicking an actual review so that I can resolve comments, but I think everything I brought up is either settled or spun into an issue so I'm happy.

amaltaro:
Tests went fine, a few follow up issues were created. So I'm merging it now.
It needs to be backported to the wmagent branch though, I just want to get another fix in place before I start cherry-picking those.

PR #9629, opened 2020-04-11
Introducing the DDM Service Wrapper.

### LINKED ISSUE #9607: MS: create a Services wrapper module for DDM
**Impact of the new feature**
ReqMgr2MS (and anything else using the Services package)

Unified talks to two different data management technologies for the output data placement. PhEDEx for relval output and DDM for standard production samples (because DDM does an intelligent data placement)

Given that we will have to use DDM REST APIs for a few months, we should implement a wrapper around those APIs in order to make its usage cleaner and straight forward.

One API that we definitely need to use is `/registry/request/copy`. No need to implement anything else, unless we find it to be important for this migration.

I have already asked Benedikt for further information. I'll update this description once I have further documentation on that API.

or create a wrapper around all the 3 data management tools: DDM, PhEDEx and Rucio

**Additional context**
Unified has this logic implemented in `_pass_to_dynamo`, here:
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L511

todor-ivanov:
Fixes #9607 

#### Status
In development

#### Description
A wrapper providing simple interface to the DMM Service. Due to the different interfaces to the DMM system, this module also defines a set of default templates for the different request types made  to DDM. It provides the ability to create a request to dynamo per dataset, or to aggregate all the similar requests into a single call.    

YES 

#### Related PRs
No

No

### DIFF
--- a/src/python/WMCore/Services/DDM/DDM.py
+++ b/src/python/WMCore/Services/DDM/DDM.py
@@ -1,0 +1,313 @@
+from __future__ import (division, print_function)
+
+import json
+import logging
+
+from WMCore.Services.Service import Service
+from pprint import pformat
+
+
+class DDMReqTemplate(dict):
+    """
+    A simple class defining a set of request templates for DDM, with default
+    structures representing the different POST calls expected by the modules
+    here:
+        https://github.com/SmartDataProjects/dynamo/tree/master/lib/web/modules
+    """
+
+    def __init__(self, api, **kwargs):
+        """
+        api:
+            String representing the type of request to be sent. In case the api
+            is missing or unrecognized the created object will be an ordinary
+            dictionary.
+        example:
+            ddmReq = DDMReqTemplate('copy',
+                                     item=['/LQLQToTop..._M-1200/NANOAODSIM'],
+                                     site=['T2_CH_CERN', 'T2_US_MIT'])
+        NOTE: we do not enforce the use of api parameter, but we may change this
+        """
+        super(DDMReqTemplate, self).__init__(**kwargs)
+
+        self.api = api
+
+        # tempTuple format: [(key, defaultValue, type), ....()]
+        # :item: list of datasets
+        # :site: list of destination sites
+        # :n: Number of copies
+        # :cache: cache the request on the DDM side for quicker copypoll calls
+        # :request_id: The Request id assigned by DDM
+        if api == 'copy':
+            tempTuple = [('item', [], list),
+                         ('site', ['T2_*', 'T1_*_Disk'], list),
+                         ('group', 'DataOps', str),
+                         ('n', None, int),
+                         ('cache', None, str)]
+
+        elif api == 'pollcopy':
+            tempTuple = [('request_id', None, int),
+                         ('item', None, list),
+                         ('site', None, list),
+                         ('status', None, str),
+                         ('user', None, str)]
+
+        elif api == 'cancelcopy':
+            tempTuple = [('request_id', None, int)]
+
+        else:
+            msg = "ERROR: Unsupported API: {}".format(api)
+            raise ValueError(msg)
+
+        # buildTemplate:
+        template = dict()
+        for tup in tempTuple:
+            template.update({tup[0]: tup[1]})
+
+        # check if we can fit all the arguments provided through **kwargs
+        for kw, arg in kwargs.iteritems():
+            found = False
+            typeok = False
+            for tup in tempTuple:
+                if kw == tup[0]:
+                    found = True
+                    if isinstance(kwargs[kw], tup[2]):
+                        typeok = True
+                        template[kw] = kwargs[kw]
+            if not found:
+                msg = "ERROR: Unrecognized parameter: {}: {}".format(kw, kwargs[kw])
+                raise KeyError(msg)
+            if not typeok:
+                msg = "ERROR: Wrong type: {} for parameter: {}: {}".format(
+                    type(kwargs[kw]),
+                    kw,
+                    kwargs[kw])
+                raise TypeError(msg)
+
+        self.update(template)
+
+    def strip(self):
+        """
+        _strip_
+
+        Method used to drop all the fields with value None. DDM does not
+        handle those quite well, and they better be removed from the request.
+        """
+
+        for key in list(self):
+            if self[key] is None:
+                self.pop(key)
+
+    def isEqual(self, ddmReq, exclude=None):
+        """
+        _isEqual_
+
+        Equivalence method.
+        :exclude:
+            A string representing the key to be excluded from the comparison.
+        :return:
+            Bool
+        """
+        # first we check the apis - we cannot compare apples to oranges:
+        if self.api != ddmReq.api:
+            return False
+
+        # in case no key have been excluded it is a normal dict comparison
+        if not exclude:
+            return self == ddmReq
+
+        # The final check: we iterate over the keys of the two objects in order
+        # to be sure we do not leave any extra key which is present in one of
+        # the objects but not in the other
+        allKeys = set(self.keys()) | set(ddmReq.keys())
+        for key in allKeys:
+            if key == exclude:
+                continue
+            try:
+                if ddmReq[key] != self[key]:
+                    return False
+            except KeyError:
+                return False
+        return True
+
+
+class DDM(Service):
+    """
+    Class which provides client APIs to the DDM service.
+
+    Usage example:
+        api = 'copy'
+        ddmReq = DDMReqTemplate(api)
+        ddm = DDM()
+        ddm.makeRequest(ddmReq, api)
+    """
+
+    def __init__(self,
+                 url=None,
+                 logger=None,
+                 configDict=None,
+                 enableDataPlacement=True):
+        """
+        configDict:
+            Dictionary with parameters that are passed to the super class.
+        enableDataPlacement:
+            If `False` just create the request templates without sending them.
+        """
+        url = url or "https://dynamo.mit.edu"
+        configDict = configDict or {}
+        configDict.setdefault('endpoint', url)
+        configDict.setdefault('cacheduration', 1)  # in hours
+        configDict.setdefault('accept_type', 'application/json')
+        configDict.setdefault('content_type', 'application/json')
+        configDict['logger'] = logger if logger else logging.getLogger()
+        super(DDM, self).__init__(configDict)
+        self.enableDataPlacement = enableDataPlacement
+        self['logger'].debug("Initializing DDM with url: %s", self['endpoint'])
+
+    def _getResult(self, ddmReq, apiUrl, callname):
+        """
+        Either fetch data from the cache file or query the data-service
+        :param metricNumber: a number corresponding to the SSB metric
+        :return: a dictionary
+
+        """
+
+        # TODO: to make this cache not per call but per DDM object instance
+        cachedApi = "%s.json" % callname
+
+        # striping the None values from the request before sending it to dynamo
+        ddmReq.strip()
+        msg = "INFO: sending data to DDM: "
+        msg += "API: {}: ddmReq: {}".format(ddmReq.api, pformat(ddmReq))
+        self['logger'].debug(msg)
+
+        # The Request class expects this to be a dict, otherwise it tries to
+        # encode the inputdata once again and inserts parasite information and
+        # breaks the call to dynamo.
+        ddmReq = dict(ddmReq)
+        data = self.refreshCache(cachedApi,
+                                 apiUrl,
+                                 inputdata=ddmReq,
+                                 verb='POST')
+        results = data.read()
+        data.close()
+
+        results = json.loads(results)
+        return results
+
+    def makeRequest(self, ddmReq):
+        """
+        _makeRequest_
+
+        Makes a SINGLE request to DDM
+
+        :ddmReq: Request template an instance of DDMReqTemplate class.
+        :return: Dictionary containing the result from DDM
+        """
+
+        # try to figure out the api from the ddmReq template
+        api = ddmReq.api
+
+        if not api:
+            msg = "WARNING: skip sending data to DDM: "
+            msg += "API: {}: ddmReq: {}\n".format(ddmReq.api, pformat(ddmReq))
+            return None
+
+        apiUrl = "registry/request/{}".format(api)
+        callname = "DDM-{}".format(api)
+
+        if not self.enableDataPlacement:
+            return ddmReq
+
+        try:
+            # TODO: here/getResult to send the request parameters to the logger
+            result = self._getResult(ddmReq, apiUrl, callname)
+        except Exception as e:
+            msg = "ERROR: sending data to DDM: \n"
+            msg += "API: {}\n".format(ddmReq.api)
+            msg += "ddmReq: {}\n".format(pformat(ddmReq))
+            msg += "ERROR: {}".format(str(e))
+            self['logger'].debug(msg)
+            return None
+        return result
+
+    def makeAggRequests(self, ddmReqList, aggKey=None):
+        """
+        _makeAggRequest_
+
+        Makes several Aggregated requests to DDM. Takes the list of objects,
+        instances of ddmReqTemplate and aggregates it based on a selected key.
+        It iterates through the list and compares all to all. Uses the
+        ddmReqTemplate internal Equivalence method for comparison with the
+        aggregation key excluded. If a match is found the two requests are
+        squeezed into one. The final result should be a new list of unique
+        ddmRequests each one of them should contain all similar ddmRequests
+        from the initial list with the values of the aggregation key accumulated
+        in a list under the same key name. Creates a request per ddmReqTemplate
+        from the list with aggregated templates.
+        NOTE:
+            In this method the api is always fetched from the template.
+        TODO:
+            So far it is working basically for the 'copy' API and due to that
+            the value of the aggregation key has to be a dictionary so no
+            explicit check needs to be done for the type. In case we decide to
+            use it we should develop this to the end with the proper type check
+            for that key
+
+        :ddmReqList: List of objects, instances of DDMReqTemplate class.
+        :aggKey:    String representing the Aggregation key.
+        :return:     List of Dictionaries containing the results from DDM
+
+        """
+        if not ddmReqList or not isinstance(ddmReqList, list):
+            return None
+
+        ddmReqListAgg = []
+
+        # populate the first element in the aggregated list
+        ddmReqListAgg.append(ddmReqList.pop())
+
+        # feed the rest
+        while len(ddmReqList) != 0:
+            ddmReq = ddmReqList.pop()
+            found = False
+            for aggReq in ddmReqListAgg:
+                if ddmReq.isEqual(aggReq, aggKey):
+                    # Check if the two objects are not references to one and the
+                    # same object. Only then copy the values of the aggKey,
+                    # otherwise we will enter an endless cycle.
+                    if ddmReq is not aggReq and aggKey is not None:
+                        for i in ddmReq[aggKey]:
+                            aggReq[aggKey].append(i)
+                    found = True
+                    del(ddmReq)
+                    break
+            if not found:
+                ddmReqListAgg.append(ddmReq)
+
+        if not self.enableDataPlacement:
+            return ddmReqListAgg
+
+        results = []
+        for ddmReq in ddmReqListAgg:
+            api = ddmReq.api
+            if not api:
+                msg = "WARNING: skip sending data to DDM: "
+                msg += "API: {}: ddmReq: {}\n".format(ddmReq.api, pformat(ddmReq))
+                self['logger'].debug(msg)
+                results.append(None)
+                continue
+
+            # here to be noticed - we create a cache file per request
+            apiUrl = "registry/request/{}".format(api)
+            callname = "DDM-{}".format(api)
+            try:
+                result = self._getResult(ddmReq, apiUrl, callname)
+                results.append(result)
+            except Exception as e:
+                msg = "ERROR: sending data to DDM: "
+                msg += "API: {}: ddmReq: {}\n".format(ddmReq.api, pformat(ddmReq))
+                msg += "ERROR: {}".format(str(e))
+                self['logger'].debug(msg)
+                print(str(e))
+                results.append(None)
+        return results
--- a/src/python/WMCore/Services/DDM/__init__.py
+++ b/src/python/WMCore/Services/DDM/__init__.py
@@ -1,0 +1,8 @@
+#!/usr/bin/env python
+"""
+_DDM.DDM_
+
+API for calling DDM service
+
+"""
+__all__ = ['DDM', 'DDMReqTemplate']
--- a/test/python/WMCore_t/Services_t/DDM_t/DDM_t.py
+++ b/test/python/WMCore_t/Services_t/DDM_t/DDM_t.py
@@ -1,0 +1,248 @@
+#!/usr/bin/env python
+"""
+Test case for DDM
+"""
+from __future__ import print_function, division
+
+import unittest
+
+from nose.plugins.attrib import attr
+
+from WMCore.Services.DDM.DDM import DDM, DDMReqTemplate
+from WMQuality.Emulators.EmulatedUnitTestCase import EmulatedUnitTestCase
+
+
+class DDMReqTemplateTest(EmulatedUnitTestCase):
+    """
+    Unit tests for DDM Request templates
+    """
+
+    def __init__(self, methodName='runTest'):
+        super(DDMReqTemplateTest, self).__init__(methodName=methodName)
+
+    def setUp(self):
+        """
+        Setup for unit tests
+        """
+        super(DDMReqTemplateTest, self).setUp()
+        self.myddmReq = DDMReqTemplate('copy')
+        self.myddmReq['item'] = ['/LQLQToTopMuTopTau_M-1200_TuneCP5_13TeV_pythia8/RunIIFall17NanoAODv5-PU2017_12Apr2018_Nano1June2019_102X_mc2017_realistic_v7-v1/NANOAODSIM']
+
+    def testConstructor(self):
+        # Test construct with default values:
+        expectedDdmReq = {'group': 'DataOps',
+                          'item': [],
+                          'site': ['T2_*', 'T1_*_Disk'],
+                          'n': None,
+                          'cache': None}
+        self.ddmReq = DDMReqTemplate('copy')
+        self.assertEqual(self.ddmReq, expectedDdmReq)
+
+        # Test bad API:
+        with self.assertRaises(ValueError):
+            self.ddmReq = DDMReqTemplate('coppy')
+
+        # Test bad request keys types:
+        with self.assertRaises(TypeError):
+            self.ddmReq = DDMReqTemplate('copy',
+                                         item='String instead of List')
+        with self.assertRaises(TypeError):
+            self.ddmReq = DDMReqTemplate('copy',
+                                         item=[],
+                                         site='String instead of List')
+        with self.assertRaises(TypeError):
+            self.ddmReq = DDMReqTemplate('copy',
+                                         item=[],
+                                         site=[],
+                                         group=['List instead of String'])
+        with self.assertRaises(TypeError):
+            self.ddmReq = DDMReqTemplate('copy',
+                                         item=[],
+                                         site=[],
+                                         group='',
+                                         n='String instead of Int')
+        with self.assertRaises(TypeError):
+            self.ddmReq = DDMReqTemplate('copy',
+                                         item=[],
+                                         site=[],
+                                         group='',
+                                         n=1,
+                                         cache=['List instead of String'])
+        # Test unsupported keys:
+        with self.assertRaises(KeyError):
+            self.ddmReq = DDMReqTemplate('copy',
+                                         unsupported='')
+
+    def testStrip(self):
+        expectedDdmReq = {'group': 'DataOps',
+                           'item': ['/LQLQToTopMuTopTau_M-1200_TuneCP5_13TeV_pythia8/RunIIFall17NanoAODv5-PU2017_12Apr2018_Nano1June2019_102X_mc2017_realistic_v7-v1/NANOAODSIM'],
+                           'site': ['T2_*', 'T1_*_Disk']}
+        self.myddmReq.strip()
+        self.assertEqual(self.myddmReq, expectedDdmReq)
+
+    def testIsEqual(self):
+        ddmReq1 = DDMReqTemplate('copy', item=['DataSet1'],     site=['T2_*', 'T1_*_Disk'])
+        ddmReq2 = DDMReqTemplate('copy', item=['DiffDataSet1'], site=['T2_*', 'T1_*_Disk'])
+        ddmReq3 = DDMReqTemplate('copy', item=['DataSet1'],     site=['T2_CH_CERN'])
+
+        # Test full (dictionary like) match - no key is excluded from the comparison:
+        self.assertFalse(ddmReq1.isEqual(ddmReq2))
+        self.assertFalse(ddmReq1.isEqual(ddmReq3))
+        self.assertFalse(ddmReq2.isEqual(ddmReq3))
+
+        # Test with 'item' key excluded from the comparison:
+        self.assertTrue(ddmReq1.isEqual(ddmReq2, 'item'))
+        self.assertFalse(ddmReq1.isEqual(ddmReq3, 'item'))
+        self.assertFalse(ddmReq2.isEqual(ddmReq3, 'item'))
+
+        # Test compare requests with equal keys, different APIS:
+        ddmReq0 = DDMReqTemplate('pollcopy', item=['DataSet1'], site=['T2_*', 'T1_*_Disk'])
+        self.assertFalse(ddmReq0.isEqual(ddmReq1))
+        self.assertFalse(ddmReq0.isEqual(ddmReq1, 'item'))
+        self.assertFalse(ddmReq0.isEqual(ddmReq2, 'item'))
+        self.assertFalse(ddmReq0.isEqual(ddmReq3, 'item'))
+
+
+class DDMTest(EmulatedUnitTestCase):
+    """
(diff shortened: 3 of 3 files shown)


src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+            done for the type. In case we decide to use it we should develop
+            this to the end with the proper type checks.
+
+        :ddmReqList: List of objects, instances of DDMReqTemplate class.
+        :aggrKey:    String representing the Aggregation key.
+        :return:     List of Dictionaries containing the results from DDM
+
+        """
+        # TODO: to put a check for aggrKey == None
+        ddmReqListAgg = []
```
amaltaro, line 200: how about starting this variable name without `ddm`? It's very easy to get lost in the block below :)

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        _makeAggrRequest_
+
+        Makes several Aggregated requests to DDM. Takes the list of objects,
+        instances of ddmReqTemplate and aggregates it based on a selected key.
+        Uses the ddmReqTemplate internal Equivalence method for comparison with
+        the aggregation key excluded. The final result should be a list of unique
+        ddmRequests each one of them containing all equal ddmRequests from the
+        initial list with the values of the aggregation key accumulated in a list
+        under the same key. Creates a request per ddmReqTemplate from the list
```
amaltaro, line 184: I'm not sure I understand these aggregated requests where one DDM call is made for each template?
For instance, is it meant to add many datasets under the same `copy` call? If so, would those many datasets be injected in a single HTTP call? How about the response, do we get a single request id?

amaltaro (changes requested):
Todor, this was a partial review. However, I have made several comments/questions along the code.
You might want to implement unit tests for the `DDQMReqTemplate` as well, since it does not need to talk to DDM and can be easily tested.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+
+        Method used to drop all the fields with value None. DDM does not
+        handle those quite well, and they better be removed from the request.
+        """
+
+        for key in list(self):
+            if not self[key]:
+                self.pop(key)
+
+    def isEqual(self, ddmReq, exclude=None):
```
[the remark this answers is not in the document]
todor-ivanov: Hi @amaltaro , I was thinking of doing so, but in this case I need to exclude the aggregation key from the check since the objects will always differ on that one, so I need to pass it as a parameter to the check somehow. And with operators I do not know how  to do it. At the end decided to go for the simple but working solution.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+    def strip(self):
+        """
+        _strip_
+
+        Method used to drop all the fields with value None. DDM does not
+        handle those quite well, and they better be removed from the request.
+        """
+
+        for key in list(self):
+            if not self[key]:
```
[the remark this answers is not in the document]
todor-ivanov: Totally correct. I will change that.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+
+class DDMReqTemplate(dict):
+    """
+    A simple class defining a set of request templates for DDM, with default
+    structures representing the different POST calls expected by the modules
+    here:
+        https://github.com/SmartDataProjects/dynamo/tree/master/lib/web/modules
+    """
+
+    def __init__(self, api=None, *arg, **kw):
```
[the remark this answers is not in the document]
todor-ivanov: Agrred again.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+                return False
+        return True
+
+class DDM(Service):
+    """
+    Class which provides client APIs to the DDM service.
+
+    Usage example:
+        api = 'copy'
+        ddmReq = DDMReqTemplate(api)
```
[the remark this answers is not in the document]
todor-ivanov: it was the first implementation, but I changed it because it constrains you a lot. The idea is that you may create and manage the objects outside of the wrapper and deal with them through the logic of the service you are building while this one to stay a generic wrapper. Also you can keep track of the status outside of this code, through Mongodb or so. My original idea was(and still is) to build the list of workflows outside catch store the parameters for output in MongoDB marking them with to_DDM = true status or a separate field with more than a bool value like DDM_status = (pending | in progress| done)  and then with a simple cycle outside this code to go and create a list of ddmRequests with their relevant output paramenters and simply pass this list to makeAggrrequest and let it do the rest.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,222 @@
...
+        configDict.setdefault('cacheduration', 1)  # in hours
+        configDict.setdefault('accept_type', 'application/json')
+        configDict.setdefault('content_type', 'application/json')
+        configDict['logger'] = logger if logger else logging.getLogger()
+        super(DDM, self).__init__(configDict)
+        self['logger'].debug("Initializing DDM with url: %s", self['endpoint'])
+
+    def _getResult(self, ddmReq, apiUrl, callname=""):
+
+        cachedApi = "%s.json" % callname
```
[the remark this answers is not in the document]
todor-ivanov: I have made this exact same comment few lines bellow in the code itself. Agree, we need to think if we can make benefit of a cache file per DDM instance. Because I want to have one instance of this outside this code and simply feed it with ddmRequstTemplate objects. This was my original idea, but....

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Yes @amaltaro , the Answer to all those question is Yes. And we get an object (json) with the status of all requested transfers in separate fields. 
It is a single call per aggregates DDM request. I will try to make the doc string more clear.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: I will try to do something about it.

todor-ivanov:
Thank you @amaltaro, Alan! I will address your comments in my next commit.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: And forgot to mention. This way we can also get rid of the usage of Phedex for the Relvals, and do everything through DDM instead. (we can put a site list in the request template and it will effectively be equivalent to a Phedex request, IIUC). It will make also the transition to Rucio much easier after that, because we will move the logic from a single place.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done + few more variable name changes.

todor-ivanov: I also changed the cycle. It was not behaving as expected. Now it is doing the job right.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Just to be more precise about what we receive from DDM, here to the current message, I attach one such aggregated request with the relative response from the server.
[ddm.log](https://github.com/dmwm/WMCore/files/4466951/ddm.log)

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: I think we will not have time to make it perfect - I suggest not to change this one. Resolving this conversation now. @amaltaro if you disagree, let me know.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: Now all the parameters for the request may be passed directly to the constructor: 
```
ddmReq = DDMReqTemplate(api='pollcopy', request_id=46633)
```

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: The docstring was updated too.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
amaltaro: If functionality and latency is the same between a PhEDEx and DDM request, then I think we could indeed use DDM for all this output data placement. I'll write to Benedikt.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+    """
+
+    def __init__(self, api=None, **kwargs):
+        """
+        api:
+            String representing the type of request to be sent. In case the api
+            is missing or unrecognized the created object will be an ordinary
+            dictionary.
+        example:
+            ddmReq = DDMReqTemplate('copy',
```
amaltaro, line 25: Should we validate the data type for those keyword arguments as well? For instance, can I set `site` to a string, or it has to be a list?

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
amaltaro: Looking how PhEDEx wrapper handles this, it calls the `refreshCache` and that one sorts the correct cache file name via `cacheFileName` method. So, I think `cachedApi` can be replaced by `callname`.

BTW, why `callname` has a default value?

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        if not api:
+            api = ddmReq.api
+
+        if not api:
+            msg = "WARNING: skip sending data to DDM: "
+            msg += "API: {}: ddmReq: {}\n".format(ddmReq.api, pformat(ddmReq))
+            return None
+
+        apiUrl = "registry/request/{}".format(api)
+        callname = "DDM-{}".format(api)
```
amaltaro, line 198: Is it for caching purposes? If so, then the `apiUrl` would suffice in my opinion.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+
+        Makes a SINGLE request to DDM
+
+        :ddmReq: Request template an instance of DDMReqTemplate class.
+        :api:    The DDM api to be queried.
+        :return: Dictionary containing the result from DDM
+        """
+
+        # try to figure out the api from the ddmReq template
+        if not api:
```
amaltaro, line 189: Perhaps we could make the API name a mandatory argument, and then here we validate it against the `ddmReq.api` attribute?

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        """
+
+        # TODO: to make this cache not per call but per DDM object instance
+        cachedApi = "%s.json" % callname
+
+        # striping the None values from the request before sending it to dynamo
+        ddmReq.strip()
+        msg = "INFO: sending data to DDM: "
+        msg += "API: {}: ddmReq: {}".format(ddmReq.api, pformat(ddmReq))
+        self['logger'].info(msg)
```
amaltaro, line 161: Maybe this logging level has to be debug before we merge it?

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+
+        if not api:
+            msg = "WARNING: skip sending data to DDM: "
+            msg += "API: {}: ddmReq: {}\n".format(ddmReq.api, pformat(ddmReq))
+            return None
+
+        apiUrl = "registry/request/{}".format(api)
+        callname = "DDM-{}".format(api)
+
+        result = None
```
amaltaro, line 200: no need for this line

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+            the aggregation key is always a dictionary so no explicit check is
+            done for the type. In case we decide to use it we should develop
+            this to the end with the proper type checks.
+
+        :ddmReqList: List of objects, instances of DDMReqTemplate class.
+        :aggKey:    String representing the Aggregation key.
+        :return:     List of Dictionaries containing the results from DDM
+
+        """
+        if not isinstance(ddmReqList, list) or len(ddmReqList) == 0:
```
amaltaro, line 240:
```suggestion
        if not ddmReqList or not isinstance(ddmReqList, list):
```

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        squeezed into one. The final result should be a new list of unique
+        ddmRequests each one of them should contain all similar ddmRequests
+        from the initial list with the values of the aggregation key accumulated
+        in a list under the same key name. Creates a request per ddmReqTemplate
+        from the list with aggregated templates.
+        NOTE:
+            In this method the api is always fetched from the template.
+        TODO:
+            So far it is working basically for the 'copy' API and due to that
+            the aggregation key is always a dictionary so no explicit check is
```
amaltaro, line 231: "the aggregation key is always a dictionary" makes be believe that `aggKey` is a dictionary, and not a string as stated below.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        # feed the rest
+        while len(ddmReqList) != 0:
+            ddmReq = ddmReqList.pop()
+            for aggReq in ddmReqListAgg:
+                if ddmReq.isEqual(aggReq, aggKey):
+                    # Check if the two objects are not references to one and the
+                    # same object. Only then copy the values of the aggKey,
+                    # otherwise we will enter an endless cycle.
+                    if ddmReq is not aggReq and aggKey is not None:
+                        for i in ddmReq[aggKey]:
```
amaltaro, line 257: what if `ddmReq[aggKey]` has a string value?

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        ddmReqListAgg = []
+
+        # populate the first element in the aggregated list
+        ddmReqListAgg.append(ddmReqList.pop())
+
+        # feed the rest
+        while len(ddmReqList) != 0:
+            ddmReq = ddmReqList.pop()
+            for aggReq in ddmReqListAgg:
+                if ddmReq.isEqual(aggReq, aggKey):
```
amaltaro, line 252: If objects are equal, doesn't it mean that all key/value pairs are the same? If so, we can simply drop `ddmReq` and move to the next object.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+                    # Check if the two objects are not references to one and the
+                    # same object. Only then copy the values of the aggKey,
+                    # otherwise we will enter an endless cycle.
+                    if ddmReq is not aggReq and aggKey is not None:
+                        for i in ddmReq[aggKey]:
+                            aggReq[aggKey].append(i)
+                    del(ddmReq)
+                    break
+            if 'ddmReq' in locals():
```
amaltaro, line 261: This check is too nasty!

amaltaro (changes requested):
Thanks for applying some of those changes, Todor. I went through this PR once again and I still have many questions/comments/requests made along the code.
In addition to those, I think we should convert the DDMReqTemplate aggregation logic into a function, such that we can have a unit test for that too.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Yes, This comment is to stress exactly this fact. That it is working so far only for keys, which have dictionary values (and as we already checked for 'item' DDM does accept dictionary). If we decide to make it functional for other aggkeys we should have an iteration with DDM people to ask which are the keys in a request that can be lists and which cannot.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: It will break - ergo the previos comment.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+
+class DDMReqTemplate(dict):
+    """
+    A simple class defining a set of request templates for DDM, with default
+    structures representing the different POST calls expected by the modules
+    here:
+        https://github.com/SmartDataProjects/dynamo/tree/master/lib/web/modules
+    """
+
+    def __init__(self, api=None, **kwargs):
```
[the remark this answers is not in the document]
todor-ivanov: I was thinking the same. Now as I know your opinion, will make the change. Thanks.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        _isEqual_
+
+        Equivalence method.
+        :exclude:
+            a key to be excluded from the comparison.
+        :return:
+            Bool
+        """
+        # first we check the apis - we cannot compare apples to oranges:
+        if not self.api == ddmReq.api:
```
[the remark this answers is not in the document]
todor-ivanov: Agree

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+        for key in list(self):
+            if self[key] is None:
+                self.pop(key)
+
+    def isEqual(self, ddmReq, exclude=None):
+        """
+        _isEqual_
+
+        Equivalence method.
+        :exclude:
```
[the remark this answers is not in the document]
todor-ivanov: For the current check (and function) it doesn't matter. It is simply a key which is excluded from the check. But it does matter for the calling function as you already noticed in the comment bellow.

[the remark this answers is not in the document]
todor-ivanov: Actually following your comment, I made an explicit check and found a silly bug (which Ishould have gotten myself). If it is passed a s a string here the check `key in exclude'  will give a false positive if key is a substring of excluded. Thanks for mentioning it here! This needs a fix!

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,286 @@
...
+            return False
+
+        # in case no key have been excluded it is a normal dict comparison
+        if not exclude:
+            return self == ddmReq
+
+        # The final check: we iterate over the two objects separately in order
+        # to be sure we do not leave any extra key which is present in one of
+        # the objects but not in the other
+        for key in self:
```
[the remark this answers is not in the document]
todor-ivanov: Excellent idea. I like it.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: In case you are not going for an aggregated Request it can be a string too. It will be sent to DDM as it is. And since DDM can digest both - lists and strings, it really doesn't matter. But I am more for sticking to a common practice everywhere, and would say we better send it always as a list. So a check here doesn't hurt.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: The default  value is because I am still for the idea to have a common cache for the whole module but not in a per request basis (of course it must be in the __init__ then). But if I do not make it work like that, then Yes, apiUrl makes more sense here. And of course having a default value her will be an error then - because we will overlap caches in case of a wrong call to the function.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Actually exactly the opposite. While using this module during the tests, I figurred out that having the 'api' in the template and fetching it from there, makes things much more easy to use and understandable (as it is done for the aggregates requests). I would rather drop this argument, or if we keep it, we may keep it just for completeness of the interface - but it will only open the door for future errors, because at some point one can make the mistake to mix a template and a call for a different apis.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: agreed

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: It is a leftover from some tests... what a shame :) Thanks  for catching that!

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: agree

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: This is exactly what is happening few lines bellow, but before we drop the object, we need to add the values from the aggregation key of the object under check to the list of values in the aggregation key of the object that it matched.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: We must check somehow that the object was not already matched, aggregated and dropped from the `for` cycle above. And if there was not a match found for it, this means it still exists here as a new and unique object and needs to be added as such at the end of the aggregated list, so that the next cycle it will be included for checks with the rest of the objects. If you have a better idea how to do this check I am open for suggestions.

amaltaro: I'm not sure how "pythonic" it is to use the locals function in the code (besides for debugging purposes)... So, I would actually use a flag to signalize whether we need to append the object or not.

src/python/WMCore/Services/DDM/DDM.py
amaltaro: I meant, if objects are **equal**, there is no aggregation to be done, objects are duplicate and we only need to keep one.

todor-ivanov: No, it is not true in the current case. In the current case the equality is checked through the `isEqual()` internal function, which returns `True` if the objects are similar on all other fields/keys, but the excluded one. They differ on that one, and this is the aggregation key in our case. And that's why I needed to develop this method at first place so that we can have a match like that:  Two objects are announced equal if they match on all other but the aggregation key, then we merge the values from the aggregation key in the new one and drop the old object.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: It will add one more line  and achieving the same result, but will make the code much clear and obvious while reading. I agree. WiIl change that too.

todor-ivanov:
Thanks for the Review @amaltaro . I will work on all the comments now. And will provide the unit test with one of the following commits.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done. Every parameter in the template is now checked against its expected type and ValueError is raised in case of mismatch.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: Done. The object keys are now compared with the excluded key as strings instead.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done. The api is now fetched from the Template only.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done. Explicit checks for key types was added in the constructor. See above.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: I tested it with apiUrl and it breaks with:
```
[Errno 2] No such file or directory: '/tmp/tivanov/.wmcore_cache_yq79dB/dynamo.mit.edu/5284144785438926808_POST_DDM-registry/request/pollcopy.json'
```
I am keeping it as it is for the moment.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: Done. Remove default value fro `callname`.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
amaltaro: I guess you forgot pushing in these changes(?)

amaltaro (changes requested):
Todor, thanks for changing what we have previously discussed.
I have done another check and made a few more comments/requests as well. Feel free to also mark no longer discussions as resolved.

I noticed you have pushed the test changes in the same commit. For future pull requests, please keep them in a different commit.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: No I did not forget. I was just keeping the default as None and then immediately checking and refusing a None value - which is not needed of course. I Just changed it.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,309 @@
...
+        example:
+            ddmReq = DDMReqTemplate('copy',
+                                     item=['/LQLQToTop..._M-1200/NANOAODSIM'],
+                                     site=['T2_CH_CERN', 'T2_US_MIT'])
+        NOTE: we do not enforce the use of api parameter, but we may change this
+        """
+        super(DDMReqTemplate, self).__init__(**kwargs)
+
+        self.api = api
+        if api is None:
```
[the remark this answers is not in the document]
todor-ivanov: Done. See above.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,309 @@
...
+            msg = "ERROR: Missing API parameter"
+            raise ValueError(msg)
+
+        # tempTuple format: [(key, defaultValue, type), ....()]
+        if api == 'copy':
+            tempTuple = [('item', [], list),
+                         ('site', ['T2_*', 'T1_*_Disk'], list),
+                         ('group', 'DataOps', str),
+                         ('n', None, int),
```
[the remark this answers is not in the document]
todor-ivanov: The default values are not type-checked so the `None` values here do not break the checks bellow. This is done in order to always have a static structure of the template with the unneeded fields set to `None` until the very last moment when they are dropped just before sending them to DDM.
   
`n` - should be the number of copies we request. About the `cache` - we  actually did not get a  

`cache` - I'd better just cite the answer from DDM developers:

```
Finally, cache=y as parameter puts it into another table of dynamo to cache them, which dynamo looks at every two hours. This was done because somehow without the cache=y, at some point the call to the API took very long for unified and this was a way around it. I would like to try it without the cache=y first, but it's definitely an option too.
```

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,309 @@
...
+                         ('user', None, str)]
+
+        elif api == 'cancelcopy':
+            tempTuple = [('request_id', None, int)]
+
+        else:
+            msg = "ERROR: Unsupported API: {}".format(api)
+            raise ValueError(msg)
+
+        # buildTemplate:
```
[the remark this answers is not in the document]
todor-ivanov: I was thinking about that before and I was playing around with stripping the `None` values as early as possible, but this can break things in two ways. First the needed properties to the request may be added on a later stage, And I would like to have it done not as an update to the dictionary but just with an overwrite of an existing placeholder (we may need to enforce type checking if a request parameter is added later). The second is it can easily break the isEqual method... because one may try to aggregate on a key which was striped earlier, so it will exclude  that key from the check and will generate a false positive. That's the reason why (as mentioned in my comment to the previous conversation), we must keep the static structure of this template (with the `None` values in it) until the ferry last moment just before we send it to DDM. Ergo we cannot simplify the code here.

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,309 @@
...
+                    break
+            if not found:
+                ddmReqListAgg.append(ddmReq)
+
+        if not self.enableDataPlacement:
+            return ddmReqListAgg
+
+        results = []
+        for ddmReq in ddmReqListAgg:
+            api = ddmReq.api
```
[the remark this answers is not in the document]
todor-ivanov: Not exactly. And more precisely because of line 294. See my answer inline bellow.

src/python/WMCore/Services/DDM/DDM.py
```diff
@@ -0,0 +1,309 @@
...
+            if not api:
+                msg = "WARNING: skip sending data to DDM: "
+                msg += "API: {}: ddmReq: {}\n".format(ddmReq.api, pformat(ddmReq))
+                self['logger'].debug(msg)
+                results.append(None)
+                continue
```
todor-ivanov, line 294: This line here splits the actions from `makeRequest' in two steps, because in case of an error with some of the requests while iterating through the list, we do not want to exit the `for` cycle completely - if it happens we will drop the rest of the requests from the list and will never notice (I wouldn't like to debug such a silent error).

test/python/WMCore_t/Services_t/DDM_t/DDM_t.py
```diff
@@ -0,0 +1,249 @@
...
+"""
+Test case for DDM
+"""
+from __future__ import print_function, division
+
+import unittest
+
+from nose.plugins.attrib import attr
+
+from WMCore.Services.DDM.DDM import DDM
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
todor-ivanov: Done

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
todor-ivanov: @amaltaro I am not resolving this conversation here until we do not take a final decision. Please feel free to do so if you think it is better to continue the discussion offline.

todor-ivanov:
@amaltaro Alan, please take a look at the final result.

src/python/WMCore/Services/DDM/__init__.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+"""
+_DDM.DDM_
+
+API for calling DDM service
+
+"""
+__all__ = ['DDM', 'DDMReqTemplate']
```
amaltaro, line 8: The way I see it it's actually not really needed because both classes are "public". This might be the reason for the poor quality jenkins report...

amaltaro:
Todor, for future PRs, please go through the detailed report provided by jenkins:
"""
Details at https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/9936/artifact/artifacts/PullRequestReport.html
"""
and fix those simple mistakes, like:
 W0611, line 9: Unused attr imported from nose.plugins.attrib 
 C0325, line 282: Unnecessary parens after 'del' keyword
Line 84, E241 multiple spaces after ',' 
etc

Especial attention to new files, they should be as clean as possible from the very beginning.

src/python/WMCore/Services/DDM/DDM.py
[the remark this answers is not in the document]
amaltaro: How come we can have a request template without having defined an api? From the template construction, it seems not to be possible.

src/python/WMCore/Services/DDM/DDM.py
amaltaro: That's why we have try/except blocks for ;)

amaltaro (approved):
Thanks for applying those changes, Todor.
I think there are still things that would deserve to be improved (or further discussed here), but honestly speaking, I don't want to overdo this DDM module since it will only be used for a short transition period. If needed, we can follow it up with extra PRs.
Thanks

PR #9637, opened 2020-04-15
Validate KeepOutput settings in TaskChain spec

### LINKED ISSUE #9636: Validate KeepOutput for the last task in a TaskChain request
**Impact of the bug**
ReqMgr2

**Describe the bug**
The `KeepOutput` task parameter can be used to define whether we want to save (inject into DBS and PhEDEx) the output of a given task or not.
It turns out one can inject a TaskChain workflow where the last task in the request sets it to false. Which is pointless, such tasks should actually get removed from the workflow.

**How to reproduce it**
Creating a taskchain workflow, where the last task in the request has a `KeepOutput=False`.

**Expected behavior**
This needs to be added to the TaskChain spec validation - during workflow creation - and if the last task sets it to false, then a validation exception should be raised and workflow creation should fail; of course, with a clear error message.

none

amaltaro:
Fixes #9636

#### Status
not-tested

#### Description
Validate the usage of KeepOutput in a TaskChain workflow and fail workflow creation if the last task is set to drop its output data.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StepChain.py
@@ -539,8 +539,8 @@ class StepChainWorkloadFactory(StdBase):
 
         lastStep = "Step%s" % schema['StepChain']
         if not strToBool(schema[lastStep].get('KeepOutput', True)):
-            msg = "Dropping the output of the last step is prohibited.\n"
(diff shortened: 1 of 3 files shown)


todor-ivanov (approved):
Looks good to me

amaltaro:
Thanks for the review, Todor. The latest update was just improving the error message.

Issue #7936, opened 2017-06-15
CC7: AgentStatusWatcher fails to get proxy information (symbol lookup error)


amaltaro:
I finally managed to deploy wmagent on the CC7 nodes, after manually installing the following packages (I still have to ask Daniel to puppetize them):
`yum -y install libXft-devel libXpm-devel libXext-devel mesa-libGLU-devel perl-Switch perl-Env perl-Thread-Queue`

and services/components run without any apparent issues.

However, looking at the AgentStatusWatcher logs, I see it's failing to retrieve the timeleft for the user proxy (debug):
```
2017-06-15 14:11:55,225:140049327441664:DEBUG:Proxy:Executing : 
 command : voms-proxy-info -file PROXYFILE -timeleft
 output : 
 error: voms-proxy-info: symbol lookup error: voms-proxy-info: undefined symbol: _ZN8vomsdataC1ESsSs

 retcode : 127
2017-06-15 14:11:55,225:140049327441664:ERROR:Proxy:Error while checking proxy timeleft for PROXYFILE
```

the interesting part is that it works just fine if the agent environment is not sourced.

Google suggests checking the shared libraries linked from voms-proxy-info with 
```
ldd `which voms-proxy-info`
```

which then shows several libraries being used from cms comp stack

amaltaro:
And I forgot to mention that I don't know even from where to start debugging this issue :(

davidlt:
I guess you are using `voms-clients-cpp` and something is wrong. lxplus7.cern.ch uses `voms-clients-java`, you cold try using it.

davidlt:
Correction boths are installed on lxplus7.cern.ch, but JAVA clients are used (in my case).

amaltaro:
The executable is the same
`/bin/voms-proxy-info`

how do I check whether it uses cpp or java?

then the libraries being used by the standard `voms-proxy-info` are:
```
cmst1@vocms0192:/afs/cern.ch/user/a/amaltaro $ ldd `which voms-proxy-info`
	linux-vdso.so.1 =>  (0x00007ffd921e6000)
	libvomsapi.so.1 => /lib64/libvomsapi.so.1 (0x00007fe505061000)
	libssl.so.10 => /lib64/libssl.so.10 (0x00007fe504df3000)
	libcrypto.so.10 => /lib64/libcrypto.so.10 (0x00007fe504a08000)
	libexpat.so.1 => /lib64/libexpat.so.1 (0x00007fe5047de000)
[...]

PR #9642, opened 2020-04-16
Move checkProxy script under WMCore; script to renew proxy

### LINKED ISSUE #9641: Proxy renewal failing in the CERN agents
**Impact of the bug**
WMAgents

**Describe the bug**
We noticed a couple of CERN agents started failing to renew my proxy - which is done through a cron entry - and over the last few days, there has been a couple more agents failing as well. Debugging shows a problem with java, likely related to some sort of update that hit those nodes in the last days..

**How to reproduce it**
Running the command manually works fine, either with or without the agent environment (!)

**Expected behavior**
The proxy renewal should go through. Ideally it would be better to have a component taking care of it, but we can simply fix it for now and keep the same renewal model.

After piping the error to a log, here is the problem:
```
A credential has been received for user amaltaro in /data/certs/mynewproxy.pem.
Error occurred during initialization of VM
java.lang.Error: Properties init: Could not determine current working directory.
        at java.lang.System.initProperties(Native Method)
        at java.lang.System.initializeSystemClass(System.java:1168)
```

amaltaro:
Fixes #9641 

#### Status
not-tested

#### Description
Changes are:
* moved the checkProxy script from my repo over to WMCore (used to check the proxy validity and notify if below a given timeleft)
* converted the cron entry to renew my proxy into a bash script (using the agent's libraries by the way)
* updated the deployment script accordingly

yes

#### Related PRs
none

### DIFF
--- a/deploy/checkProxy.py
+++ b/deploy/checkProxy.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python
+"""
+Script to check the proxy validity and notify users in case it's close
+to expiration.
+
+Original author: direyes@cern.ch
+"""
+from __future__ import division, print_function
+
+import getopt
+import os
+import re
+import subprocess
+import sys
+
+
+def main(argv):
+    """
+    This is an utility that checks the proxy validity and sends an email in
+    case the time left is lesser than --time.
+    Arguments:
+    --proxy     : Proxy file location (Default: X509_USER_PROXY)
+    --myproxy   : Check the log term proxy in myproxy (Default: True)
+    --mail      : Mail where notification should be sent (Default: USER).
+    --send-mail : <True|False> Send mail notification (Default: True).
+    --time      : Minimun time left [hours]. It should be an integer. (Default: 48 h).
+    --verbose   : Print output messages.
+    --help      : Print this menu.
+
+    Example: python /data/amaltaro/checkProxy.py --proxy /data/certs/myproxy.pem
+    --time 72 --send-mail True --mail alanmalta@gmail.com,alan.malta@cern.ch --verbose
+    """
+
+    valid = ["proxy=", "myproxy=", "mail=", "send-mail=", "time=", "verbose", "help"]
+    ### // Default values
+    proxy = os.getenv('X509_USER_PROXY')
+    myproxy = False
+    verbose = False
+    mail = os.getenv('USER')
+    sendMail = True
+    time = 3
+    host = os.getenv('HOSTNAME')
+
+    try:
+        opts, _args = getopt.getopt(argv, "", valid)
+    except getopt.GetoptError as ex:
+        print("Options: {}\n\nException: {}".format(main.__doc__, str(ex)))
+        sys.exit(1)
+
+    ### // Handle arguments given in the command line
+    for opt, arg in opts:
+        if opt == "--help":
+            print(main.__doc__)
+            sys.exit(0)
+        if opt == "--proxy":
+            proxy = arg
+            if proxy.startswith("~/"):
+                proxy = os.getenv('HOME') + proxy[1:]
+            if not os.path.exists(proxy):
+                print("Proxy File does not exist")
+                sys.exit(2)
+        if opt == "--mail":
+            mail = arg
+        if opt == "--myproxy":
+            myproxy = arg
+        if opt == "--send-mail":
+            sendMail = arg
+        if opt == "--time":
+            time = int(arg)
+            if time < 1:
+                print("Invalid time format. Check the options: {}".format(main.__doc__))
+                raise sys.exit(3)
+        if opt == "--verbose":
+            verbose = True
+
+    command = ["voms-proxy-info", "-file", str(proxy)]
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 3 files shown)


deploy/checkProxy.py
```diff
@@ -0,0 +1,162 @@
...
+            print('Proxy information: {}'.format(proxyInfo))
+        timeLeft = []
+        for line in proxyInfo:
+            if line.find('timeleft') > -1:
+                dateReg = re.compile('\d{1,3}[:/]\d{2}[:/]\d{2}')
+                timeLeft = dateReg.findall(line)[0]
+                timeLeft = timeLeft.split(':')[0]
+                continue
+    else:
+        msg = "Valid proxy file in %s not found. " % host
```
todor-ivanov, line 128: I would change this message to:
"No valid proxy was found in %s"

deploy/renew_proxy.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+credname=CREDNAME
+source /data/admin/wmagent/env.sh
+source /data/srv/wmagent/current/apps/wmagent/etc/profile.d/init.sh
+myproxy-get-delegation -v -l amaltaro -t 168 -s 'myproxy.cern.ch' -k $credname -n \
+  -o /data/certs/mynewproxy.pem && voms-proxy-init -rfc -voms cms:/cms/Role=production -valid 168:00 \
```
todor-ivanov, line 6: In this line we are still using voms-proxy-init through the system paths as they are configured by default. Which in many cases (especially for the CERN machines) points to voms-client-java. [1]

[1]
```
[vocms0252]$ ll `which voms-proxy-init`
lrwxrwxrwx. 1 root root 33 Nov 17  2018 /usr/bin/voms-proxy-init -> /etc/alternatives/voms-proxy-init
[vocms0252]$ ll /etc/alternatives/voms-proxy-init
lrwxrwxrwx. 1 root root 25 Nov 17  2018 /etc/alternatives/voms-proxy-init -> /usr/bin/voms-proxy-init3
[vocms0252]$ cat /usr/bin/voms-proxy-init3
#!/bin/sh
VOMS_CLIENTS_JAVA_OPTIONS=${VOMS_CLIENTS_JAVA_OPTIONS:-"-XX:+UseSerialGC -Xmx16m"}
java $VOMS_CLIENTS_JAVA_OPTIONS -cp $(build-classpath voms-clients-java voms-api-java canl-java bcpkix bcprov commons-cli commons-io) org.italiangrid.voms.clients.VomsProxyInit "$@"
```

todor-ivanov (approved):
Having it all done the way it is suggested in this PR is much better than the current implementation. It improves the situation a lot. Although, even with a dedicate utilitarian script we still do not avoid the risk of eventual crash of the Java voms-clients with the next system upgrade(due to a silent java version change etc.). Please take a look at the comment I made directly in the issue. This is a problem already faced before, with a proven solution.

deploy/checkProxy.py
```diff
@@ -0,0 +1,162 @@
...
+    for line in proxyInfo:
+        message += "%s\n" % line
+    # echo %msg | mail -s 'HOST proxy status' MAIL_ADD
+    command = " echo \"%s\" | " % (message)
+    command += "mail -s '%s: Proxy status'" % (host)
+    command += " %s" % (mail)
+
+    if verbose:
+        print("Running email command: {}".format(command))
+    retCode = os.system(command)
```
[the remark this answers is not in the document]
amaltaro: Yes, I considered that library too. But then this script becomes dependent on the libraries we ship with WMAgent, which might bring other problems too.
I have just made the necessary changes to use the EmailAlert, but then I started getting an `undefined symbol` problem.

deploy/checkProxy.py
amaltaro: Changing

amaltaro:
Todor, I believe I already tried to change to the c++ version of the voms-client in the past, and failed.
In this PR, I also tried to use the EmailAlert library, but then we hit other problems that are likely related to this issue:
https://github.com/dmwm/WMCore/issues/7936

So, I'm keeping it as it is for now, I don't think it's worth the effort at the moment. Feel free to create another issue (and the final fix), if you know how exactly this can be improved.
Thanks for your comments though.

todor-ivanov:
Thanks @amaltaro for letting me know about your past effort. That was good to read. Yes, it is logical that with the C clients we introduce dependency on library versions from the comp environment. And I can see the other eventual version misalignment may obviously come from yet another package that lives outside comp - HTCondor. We of course can spare the effort of just moving the pain from one place to another and we may continue living with the java version.

todor-ivanov (commented):
looks good to me

amaltaro:
AgentStatusWatcher is the right place to have these checks. I actually don't know why it hasn't been implemented that way yet, likely because there is one of these problems already seen. 
Still, that's the correct way of fixing it!

amaltaro:
Sigh...of course FNAL and CERN nodes have different version of the voms-proxy-init. And now the proxy creation fails with
```
A credential has been received for user amaltaro in /data/certs/mynewproxy.pem.
voms-proxy-init: symbol lookup error: voms-proxy-init: undefined symbol: _ZN8vomsdataC1ESsSs
```
on the FNAL nodes.

PR #9648, opened 2020-04-17
RucioInjector has to close only those blocks injected into Rucio

### LINKED ISSUE #9647: RucioInjector trying to close blocks injected into PhEDEx
**Impact of the bug**
WMAgent

**Describe the bug**
Sigh, from my testbed agent, it looks like I missed the datatier check on the `closeBlocks()` method. Which means, RucioInjector might try to close blocks that have actually being injected into PhEDEx, see:
```
2020-04-15 01:42:04,886:140658649351936:ERROR:RucioInjectorPoller:Failed to close block: /RelValQCD_FlatPt_15_3000HS_13/CMSSW_10_6_1_patch1-SC_ReDigi_Harvest_Agent133_Val_Alanv11-v11/GEN-SIM#af031cd3-55d5-44be-aedc-743feca2704d. Will retry again later.
```

It's not a terrible thing, because eventually PhEDExInjector will close those blocks and they won't be retried by RucioInjector. Yet, unneeded data service calls and other issues.

**How to reproduce it**
standard agent deployment with RucioInjector enabled

**Expected behavior**
The component should apply the same rules as specified by its configuration parameter:
```
config.RucioInjector.listTiersToInject = []
```
and only try to close blocks for data tiers that are listed in that parameter.

none

amaltaro:
Fixes #9647 

#### Status
not-tested

#### Description
Rookie mistake! The same check we have for injecting data into Rucio, we need it to close blocks as well. So, verify that the data tier is allowed before trying to close blocks from the RucioInjector component. 
If it's not allowed, then we can assume the block has been inserted into PhEDEx instead.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -347,6 +347,8 @@ class RucioInjectorPoller(BaseWorkerThread):
         ### FIXME the data format returned by this DAO
         for location in migratedBlocks:
             for container in migratedBlocks[location]:
+                if not self._isContainerTierAllowed(container, checkRulesList=False):
+                    continue
                 for block in migratedBlocks[location][container]:
                     if self.rucio.closeBlockContainer(block):
                         self.setBlockClosed.execute(block)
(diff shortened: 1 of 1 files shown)


amaltaro:
Sigh... it helps if I enable the component when running tests over the weekend....

todor-ivanov (approved):
Looks good to me

amaltaro:
This fix is actually not okay. Reason is that there is no distinction between tiers to inject vs tiers to create rules, in `_isContainerTierAllowed`. I will have to separate those two, because there can be actually 4 combinations between them, not only 2...

amaltaro:
Merging it to make a new release

PR #9659, opened 2020-04-23
Consider PU location for top level MC task

### LINKED ISSUE #9658: Top level task must consider pileup location of subsequent tasks
**Impact of the bug**
WMAgent

**Describe the bug**
I still cannot believe this is the current buggy behaviour...
Apparently, when the agent acquires a TaskChain WQE without any input dataset, it creates the top level jobs against all the sites in the SiteWhitelist, even if the subsequent tasks require a pileup dataset.

**How to reproduce it**
Injecting a TaskChain workflow, without any input dataset and a pileup dataset in the Task2.
Assign it to a site containing the pileup and another without it.

**Expected behavior**
Top level jobs should be created against the intersection(SiteWhitelist - SiteBlacklist, PileupLocation). 

Problem reported in this issue: https://its.cern.ch/jira/browse/CMSCOMPPR-12155
Error message in WMStats is:
```
    Fatal Exception (Exit Code: 8029)

        An exception of category 'NoSecondaryFiles' occurred while
           [1] Constructing module: class=MixingModule label='mix'
        Exception Message:
        RootEmbeddedFileSequence no input files specified for secondary input source.
```

amaltaro:
Fixes #9658

#### Status
ready

#### Description
When adding MCFakeFile files into `wmbs_file_details` table, consider the common locations of the WQE instead of setting every single PNN as a candidate for that file.

In other words, when fetching a WQE without any input dataset, but with pileup dataset, set the MCFakeFile location to the same location as of the pileup data, such that we ensure that jobs will only be submitted to sites hosting some of the pileup data (be it a top level task or the subsequent ones).

NOTE: we should probably make the same changes to top level tasks **with* input data (or ACDC data). I'm not making those changes right now because I believe we would have negative side effects. E.g., misassignment generating an empty file location; e.g., ACDC collections with much less locations than it should... and anything else that still didn't come to my mind. 

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WorkQueue/WMBSHelper.py
+++ b/src/python/WMCore/WorkQueue/WMBSHelper.py
@@ -153,7 +153,8 @@ class WMBSHelper(WMConnectionBase):
     Interface between the WorkQueue and WMBS.
     """
 
-    def __init__(self, wmSpec, taskName, blockName=None, mask=None, cachepath='.'):
+    def __init__(self, wmSpec, taskName, blockName=None, mask=None,
+                 cachepath='.', commonLocation=None):
         """
         _init_
 
@@ -169,6 +170,7 @@ class WMBSHelper(WMConnectionBase):
         self.topLevelFileset = None
         self.topLevelSubscription = None
         self.topLevelTaskDBSBufferId = None
+        self.commonLocation = commonLocation or []
 
         self.mergeOutputMapping = {}
 
@@ -185,7 +187,6 @@ class WMBSHelper(WMConnectionBase):
         self.setFileAddChecksum = self.daofactory(classname="Files.AddChecksumByLFN")
         self.addFileAction = self.daofactory(classname="Files.Add")
         self.addToFileset = self.daofactory(classname="Files.AddDupsToFileset")
-        self.getLocations = self.daofactory(classname="Locations.ListSites")
         self.getLocationInfo = self.daofactory(classname="Locations.GetSiteInfo")
 
         # DAOs from DBSBuffer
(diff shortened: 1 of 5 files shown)


amaltaro:
squashing

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -348,18 +349,14 @@ def addMCFakeFile(self):
...
-            try:
-                siteInfo = self.getLocationInfo.execute(site, conn=self.getDBConn(),
-                                                        transaction=self.existingTransaction())
-                if not siteInfo:
-                    self.logger.info('Skipping MonteCarlo injection to site "%s" as unknown to wmbs' % site)
-                    continue
-                locations.add(siteInfo[0]['pnn'])
-            except Exception as ex:
-                self.logger.error('Error getting storage element for "%s": %s' % (site, str(ex)))
+        siteInfo = self.getLocationInfo.execute(conn=self.getDBConn(),
```
todor-ivanov, line 354: @amaltaro I have only one question. Why did you take the siteInfo query out of the 'try: - except:' block?

src/python/WMCore/WorkQueue/WMBSHelper.py
```diff
@@ -348,18 +349,14 @@ def addMCFakeFile(self):
...
+                                                transaction=self.existingTransaction())
+        for site in siteInfo:
+            if site['pnn'] in self.commonLocation:
```
todor-ivanov, line 357: What happens if sefl.commonLocation is `[]` [1]. Do we skip submission? Or we are assuring that for those workflows we always provide this parameter to the call? Do we have such a check outside the WMBSHelper class (where it is actually instanced)? The non `None` value here seems strange to me to be the only way of communicating the fact that we are running on the particular type of workflows which need the additional check here.
   
https://github.com/dmwm/WMCore/pull/9659/commits/964437b2a607f63074e3f46ded6e2758b4cb65f7#diff-af5cb316c0b66783a9f738da4bc8c83aR173

todor-ivanov (approved):
I had Only one comment inline. The rest looks good.

src/python/WMCore/WorkQueue/WMBSHelper.py
amaltaro: It should be caught upstream, without breaking the whole cycle, but only this one element.
I also removed it, because as you can see in the old line 352, it wasn't covered by any try/except block as well.

src/python/WMCore/WorkQueue/WMBSHelper.py
amaltaro: If `self.commonLocation` is empty, then work will failed to get acquired in WMBS.
However, there shouldn't be such cases, otherwise it would mean that WQEs would never leave the global workqueue.

todor-ivanov: My point exactly. My question was regarding this part - 'there shouldn't be such cases'. Is there an explicit check for that outside this class? I am sure you already anticipated that kind of situation, but thought would be faster to ask instead of digging into the code myself.

amaltaro: This piece of code is actually getting work from the local queue (LQE) and adding it to the WMBS. That means, if the LQE cannot fulfill all the data locations, work cannot be materialized in jobs (WMBS insertion). In other words, there is no explicit check for self.commonLocation, but I do thing the check on common locations for WQEs cover it.

todor-ivanov: That's fine with me. It doesn't matter where the check is done, as long as you are sure we do not create like a memory leak behavior in terms of LQEs ... I mean to start accumulating elements in the LQ which never get submitted for real. All the rest is good I believe. Resolving this conversation now so we can continue with more important stuff. Thank you @amaltaro  !

amaltaro:
Thank you for your comments, Todor.

I still think input data and pileup data location should not be correlated in WMBS, only for building the DESIRED_Sites lists. Otherwise we have an impact in the ACDC documents, where data location would actually be their intersection with the pileup location.
Anyways, that would be a big change. I just wanted to register it here.

PR #9672, opened 2020-05-02
Always create block-level rules for blocks inserted by RucioInjector

amaltaro:
Fixes #9654

#### Status
not-tested

#### Description
After further discussion, we concluded that there is no need to have block-level rule creation configurable. Thus, if RucioInjector is inserting files/blocks in Rucio, let it also create the block-level rules.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #9667, opened 2020-04-29
Added "project" metadata for block and container objects

amaltaro:
Fixes #9655

#### Status
not-tested

#### Description
This pull request injects deployment changes, where there will be a different RucioInjector parameter configuration for agents dealing with different workflow activities.

So, this PR uses the Rucio supported `project` DID metadata block and container creation, in short:
* `Production`: for central production activities/workflows
* `RelVal`: for Release Validation activities/workflows
* `Tier0`: for Tier0 production activities/workflows
* `Test`: for any test/dev/replay activities/workflows
* `User`: to be used by Analysis

Validation of the `project` value has also been added to the Rucio Services wrapper, such that we can ensure that data created through those APIs will be validated and consistent.

TODO: the Tier0 team will need to make the necessary changes on their side as well.

#### Is it backward compatible (if not, which system it affects?)
Yes, as a component functionality. But not due to a new DID information.

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #9666, opened 2020-04-29
Properly deal with MSS/Tape container rule creation

amaltaro:
Fixes #9638 

#### Status
ready

#### Description
Summary of changes are:
* Convert *_MSS endpoints to *_Tape
* added `activity="Production Output"` to container and block level rules (as per this request: https://github.com/dmwm/WMCore/issues/9654#issuecomment-620793995)
* raise `WMRucioException` exception from the Services/Rucio wrapper if we get a `rucio.AccessDenied` exception in the `createReplicationRule` method
* if container-level rule creation fails with `WMRucioException`, then retry with `ask_approval=True`

It's important to highlight that both `Custodial` and `NonCustodial` subscriptions can go through.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #9673, opened 2020-05-04
Bundle patch for 1.3.3.patch3

amaltaro:
wmagent branch version of:
 Added "project" metadata for block and container objects #9667 
 Properly deal with MSS/Tape container rule creation #9666 
 Always create block-level rules for blocks inserted by RucioInjector #9672 

When applying this patch to the production agents, we need to manually:
* add this configuration line to the config.py
```
config.RucioInjector.metaDIDProject = 'Production'
```
* or this line for the relval agent:
```
config.RucioInjector.metaDIDProject = 'RelVal'
```

further configuration will be required once we decide to enable RucioInjector in production.


amaltaro:
[Fix #9673 - Bundle patch for 1.3.3.patch3](https://trello.com/c/LW0MYcWo/204-fix-9673-bundle-patch-for-133patch3)

PR #9682, opened 2020-05-07
Print warning in Executor if condor_chirp is not found.

khurtado:
Fixes #9674

#### Status
Ready

#### Description
Print warning in Executor if condor_chirp is not found. This will be seen in the JobCache of the JobCreator.

#### Is it backward compatible (if not, which system it affects?)
YES

PR #9689, opened 2020-05-11
Fix finding condor_chirp for containers 

### LINKED ISSUE #9688: Fix condor_chirp
**Impact of the bug**
If CONDOR_CONFIG is not in the environment (containers case now), we don't use `condor_chirp`

**Describe the bug**
We don't not even try to look for `condor_chirp` in the environment if `CONDOR_CONFIG` is not found

**How to reproduce it**
Any workflow running inside containers that don't set `CONDOR_CONFIG`, like T2_CH_CERN

**Expected behavior**
We should not depend on `CONDOR_CONFIG` to find `condor_chirp` in the environment.

** Additional comments**
Basically, we fixed the warning issues from below:
https://github.com/dmwm/WMCore/issues/9674

Even though `condor_chirp` is now available, `CONDOR_CONFIG` is not with new implementation. This makes sense because we won't always find file inside the container, or the file PATH variables, so we should probably still leave that as legacy, in case we ever need the non-container support, but look for `condor_chirp` in the environment independently of that environment variable.

khurtado:
Fixes #9688

#### Status
 ready

#### Description
Do not only rely on `CONDOR_CONFIG` being present in the environment to find `condor_chirp`.

YES

#### Related PRs
Goes along with the already merged PR: https://github.com/dmwm/WMCore/pull/9682

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executor.py
+++ b/src/python/WMCore/WMSpec/Steps/Executor.py
@@ -181,15 +181,17 @@ class Executor(object):
             value = zipEncodeStr(value, maxLen=maxLen)
 
         # construct condor_chirp binary location from CONDOR_CONFIG
+        # Note: This works when we do not use containers.
         condor_chirp_bin = None
         condor_config = os.getenv('CONDOR_CONFIG', None)
         if condor_config:
             condor_config_dir = os.path.dirname(condor_config)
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
+1
Thanks Kenyi!

PR #9692, opened 2020-05-13
Load wmstats couch views with stale=update_after in WMStats Service wrappers

amaltaro:
Fixes #9680 

#### Status
not-tested

#### Description
Instead of loading wmstats views in different ways. Always go through via `_getCouchView()`, which by default would add `stale=update_after` view option (unless provided a different value).

This `update_after` means: return whatever data that is already indexed in the view at the moment (thus, it allows for a quick response). Once CouchDB serves the request, a view indexing is triggered (in case there are new documents), i.e., the view data is updated.

Documentation for 1.6.1: https://docs.couchdb.org/en/1.6.1/api/ddoc/views.html

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
This is another fix: https://github.com/dmwm/WMCore/pull/9684

#### External dependencies / deployment changes
none

PR #9684, opened 2020-05-08
Break up WMStats DataCache update into multiple calls

amaltaro:
Fixes #9680 

#### Status
under testing

#### Description
Summary of changes is:
* better classify which workflow statuses we should fetch job information from wmstats (using two new static list of statuses);
* make smaller slices (hardcoded to 5k ATM) when fetching job information for a workflow/agent tuple (via `latestRequest` view)
* when fetching workflows by status from ReqMgr2, make a separate call per status (given that it has `include_docs=true`, we better break this call into smaller pieces)
* make it more verbose such that we know where exactly it fails

Note: I see many timeouts from this simple wmstats view: `requestAgentUrl`. It's like 5MB of data and it should definitely not timeout...

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #9702, opened 2020-05-19
Hotfixes for reqmon 1.3.4.cmsweb1

amaltaro:
1.3.4_cmsweb branch version of:
 Break up WMStats DataCache update into multiple calls #9684 
 Load wmstats couch views with stale=update_after in WMStats Service wrappers #9692 

It provides basically two changes:
* DataCacheUpdate thread logic with more granular requests (instead one fat request fetching tons of data in the same call)
* by default, WMStats view calls will be made with `stale=update_after`, such that view requests are non-blocking.

### DIFF
--- a/src/python/WMCore/ReqMgr/DataStructs/RequestStatus.py
+++ b/src/python/WMCore/ReqMgr/DataStructs/RequestStatus.py
@@ -83,6 +83,25 @@ ACTIVE_STATUS = ["new",
                  "aborted-completed",
                  "rejected"]
 
+### WMSTATS_JOB_INFO + WMSTATS_NO_JOB_INFO is meant to be equal to ACTIVE_STATUS
+WMSTATS_JOB_INFO = ["running-open",
+                    "running-closed",
+                    "force-complete",
+                    "completed",
+                    "closed-out"]
+
+WMSTATS_NO_JOB_INFO = ["new",
+                       "assignment-approved",
+                       "assigned",
+                       "staging",
+                       "staged",
+                       "acquired",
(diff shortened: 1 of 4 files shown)


amaltaro:
@todor-ivanov Todor, can you please have one last look at this backported version PR? I'll try to get Imran deploying it to testbed still this evening, so I'll need to get it merged within 30min or so.

amaltaro:
Todor, I'll keep the master PRs opened for now. If there is any non-critical issue, we can follow up over there. If there are critical issues, then I'll cut yet another _cmsweb tag. Merging it now to get this thing rolling.

amaltaro:
Here is the deployment change btw: https://github.com/dmwm/deployment/pull/904

As one can see, default value is `false`, so the thread is not going to fetch job-level data. I tested it in testbed and it works fine. I have updated the config by setting `getJobInfo=True` and now I can see job information in wmstats testbed.

Once the production storm is gone and we feel confident with central services again, we can then switch that flag to True as well.

amaltaro:
It has been deployed to CMSWEB production (in the form of patching `reqmon`, to avoid too much work on the HTTP team).
The agents took less than an hour to replicate their documents back to central wmstats. Views indexing took a bit longer, and it got very slow at two or three points, but it seems to be mostly normal now.

For the record, wmstats db has a volume of 4.5GB, while the views are 8.5GB at the moment (which is a bad sign btw, views should not be bigger than the database itself!)

todor-ivanov (commented):
Looks good to me

PR #9706, opened 2020-05-21
Use DELETE method to delete single couch documents

### LINKED ISSUE #9705: Delete Couchdb documents with the DELETE method
**Impact of the new feature**
CouchDB interactions

Not a problem at all, but it might be cleaner and easier to debug if we see DELETE methods in the couchdb logs, instead of a PUT method which wipes out most of the content of the document.

Instead of making another PUT request with `_deleted=true` against the couchdb databases, use the DELETE method to delete single documents.
In case there are many documents to delete, then making a POST call against the `_bulk_docs` API might be faster (to be tested though).
 
not change anything, since everything works fine at the moment.

**Additional context**
none

amaltaro:
Fixes #9705 

#### Status
not-tested

#### Description
As the subject says, delete couchdb documents via DELETE http method, instead of PUTting/POSTing the same document with `_deleted=true`.

TODO: evaluate `_bulk_docs` API for many deletions against many DELETE calls.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Database/CMSCouch.py
+++ b/src/python/WMCore/Database/CMSCouch.py
@@ -422,9 +422,13 @@ class Database(CouchDBRequests):
         """
         Immediately delete a document identified by id and rev.
         """
-        doc = self.document(id, rev)
-        doc.delete()
-        return self.commitOne(doc)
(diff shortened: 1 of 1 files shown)


todor-ivanov (approved):
Looks good to me

PR #9710, opened 2020-05-27
Distribute WMStats couch views among different design docs

amaltaro:
Fixes #9708 

#### Status
not-tested

#### Description
Distribute the WMStats/WMStatsErl views into different design docs, organized as:
```
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl/_view/requestAgentUrl"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl1/_view/byAgentURL"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl2/_view/allWorkflows"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl3/_view/jobsByStatusWorkflow"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl4/_view/cooledoffRequests"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl5/_view/time"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl6/_view/requestHistory"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl7/_view/agentInfo"
src/couchapps/WMStats/rewrites.json:    "to": "../WMStatsErl7/_view/latestRequest"
```

Views will have to be completely rebuilt.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
Need deployment changes: https://github.com/dmwm/deployment/pull/908

PR #9713, opened 2020-05-31
add new WMStatsErl ddocs to the reqmon package dependency

### LINKED ISSUE #9708: Separate CouchDB views into different design docs
**Impact of the new feature**
CouchDB

It's related to the current performance problems with the wmstats database.

Separate the wmstats views (and views for other dbs) into different design docs, such that views can be {re-}indexed in separated javascript process; and to avoid reindexing the all the views in case we have changes/add/delete any other view.
It is still not clear whether we gain multi-processing performance if views are written natively in erlang, but it's worth a test. 

Can just live with what we have right now.

**Additional context**
Further references can be found at:
https://stackoverflow.com/questions/40750008/couchdb-one-design-doc-with-multiple-views-vs-multiple-design-docs-with-split
and
https://www.dimagi.com/blog/what-every-developer-should-know-about-couchdb/

amaltaro:
Fixes #9708 

#### Status
not-tested

#### Description
Add WMStatsErl* ddocs as a dependency during build time.

yes

#### Related PRs
Complement to: #9710

None


amaltaro:
I don't manage to build private RPMs for this branch, so I'll have to test it after it gets merged :(

amaltaro:
Yep, it works!

PR #9715, opened 2020-06-01
Fix wrong number of requests by status retrieved from Reqmgr2

### LINKED ISSUE #9714: MSOutput: Wrong number of requests by status retrieved from Reqmgr2 in log message
**Impact of the bug**
Reqmgr2ms

**Describe the bug**
There is a wrong number of requests by status in the following log message [1] in MSOutput micro service. It always shows the full length of the `requestRecords` dictionary but not the length of the result from the current status.

[1]
https://github.com/dmwm/WMCore/blob/d1db5986c6b09f09f96ef38cf4f9d549373edd98/src/python/WMCore/MicroService/Unified/MSOutput.py#L185 

**How to reproduce it**
An inspection of the logs from the current implementation is enough.

**Expected behavior**
The log message should reflect the numbers from the current query regarding every status separately.

N/A

todor-ivanov:
Fixes #9714 
#### Status
ready

#### Description
Recalculate the number of requests by status as a difference between the initial length of the `requestRecords` dictionary and the length of the updated one. 

YES

#### Related PRs
N/A

NO

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -181,9 +181,10 @@ class MSOutput(MSCore):
         try:
             requestRecords = {}
             for status in reqStatus:
+                numRequestRecords = len(requestRecords)
                 requestRecords.update(self.getRequestRecords(status))
                 msg = "{}: Retrieved {} requests in status {} from ReqMgr2. ".format(self.currThreadIdent,
-                                                                                     len(requestRecords),
+                                                                                     len(requestRecords) - numRequestRecords,
                                                                                      status)
                 self.logger.info(msg)
         except Exception as err:  # general error


src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -181,9 +181,10 @@ def _executeProducer(self, reqStatus):
         try:
             requestRecords = {}
             for status in reqStatus:
+                numRequestRecords = len(requestRecords)
                 requestRecords.update(self.getRequestRecords(status))
```
amaltaro, line 185: Isn't it easier to just separate this line into two, a) fetch requests, b) update requestRecords. In between them, you print the amount of requests in that specific status. This avoids the arithmetics with dict/lists.

amaltaro (commented):
Todor, I made a simple suggestion that - for my taste - makes the code cleaner. Please have a look and apply it if you agree.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I know that in the code it looks much cleaner... but then you need save the output with the enormous list of `requestRegcords` from the given new `status` in another variable and you need to to live with that variable up until the end of the whole call to`_executeProducer()`, while in the case how I made it, one should never save anything out of the scope of `update()` and since the `len()` builtin is already of complexity of O(1) you basically have the best performance with the minimum memory footprint here. So my opinion is we should bare with the little bit unreadable code here for the sake of performance.

amaltaro: You have a good point. I'm curious to know whether in practice this is what happens though (in case you have profiled it). Otherwise, the code is good to go indeed.

amaltaro (approved):

PR #9717, opened 2020-06-03
MSOutput implementation for 'RelVal' workflows.

### LINKED ISSUE #9608: MS: perform output data placement for completed RelVal workflows
**Impact of the new feature**
ReqMgr2MS

As stated in the document, there are two types of output dataset placement:
* relval datasets
* standard datasets
this issue is meant to address **only** the RelVal one, which currently uses PhEDEx.

NOTE: this implementation has to have all the hooks/placeholders for Rucio.

Fetch workflows sitting in the following status:
`'completed', 'closed-out', 'announced'`

filter their output datasets `OutputDatasets`, and make a phedex dataset placement for:
* priority='normal', approve=True, group='RelVal'.

We need to implement the datatier map as well, as documented in the wiki page.

Persist the workflow name, output datasets, and maybe transfer id somewhere in MongoDB. 

The logic of what has to be handled and what has been already handled might need further thought.

**Additional context**
none

todor-ivanov:
Fixes #9608 

#### Status
ready

#### Description
Implements the Output data placement for RelVal workflows. The basic logic is simple it builds the destination map for the RelVal workflows in the Producer thread and later in the Consumer just uses it to create a series of transfers requests aggregated per dataset name and destination eg. A workflow with the following destination map [1] would get those 3 transfer requests created [2]    

NOTE: Before deployment of this one the MongoDB needs to be wiped out.

[1]
```
u'destination': [u'T2_CH_CERN', u'T1_US_FNAL_Disk'],
u'destinationOutputMap': [{u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/GEN-SIM'],
                           u'destination': [u'T2_CH_CERN',
                                            u'T1_US_FNAL_Disk']},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/GEN-SIM-DIGI-RAW'],
                           u'destination': [u'T2_CH_CERN',
                                            u'T1_US_FNAL_Disk']},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/GEN-SIM-RECO'],
                           u'destination': [u'T2_CH_CERN',
                                            u'T1_US_FNAL_Disk']},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlMinBias-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-HcalCalHBHEMuonFilter-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-EcalESAlign-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlZMuMu-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlMuonIsolated-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-SiStripCalMinBias-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlJpsiMuMu-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-SiPixelCalSingleMuon-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-MuAlOverlaps-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlUpsilonMuMu-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
                           u'destination': []},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/DQMIO'],
                           u'destination': [u'T2_CH_CERN']},
                          {u'datasets': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/MINIAODSIM'],
                           u'destination': [u'T2_CH_CERN']}],

```

[2]
```
2020-06-03 11:01:00,319:INFO:MSOutput: Making transfer subscriptions for nwickram_RVCMSSW_11_1_0_pre7NuGun_200511_194429_2673
2020-06-03 11:01:00,320:WARNING:MSOutput: No data found in ddmResults for nwickram_RVCMSSW_11_1_0_pre7NuGun_200511_194429_2673. Either dry run mode or broken transfer submission to DDM. ddmResults: 
[{'cache': None,
  'group': 'RelVal',
  'item': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/MINIAODSIM',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/DQMIO'],
  'n': None,
  'site': [u'T2_CH_CERN']},
 {'cache': None,
  'group': 'RelVal',
  'item': [u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlUpsilonMuMu-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-MuAlOverlaps-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-SiPixelCalSingleMuon-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlJpsiMuMu-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-SiStripCalMinBias-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlMuonIsolated-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlZMuMu-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-EcalESAlign-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-HcalCalHBHEMuonFilter-111X_mcRun3_2021_realistic_v4-v1/ALCARECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-TkAlMinBias-111X_mcRun3_2021_realistic_v4-v1/ALCARECO'],
  'n': None,
  'site': []},
 {'cache': None,
  'group': 'RelVal',
  'item': [u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/GEN-SIM-RECO',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/GEN-SIM-DIGI-RAW',
           u'/RelValNuGun/CMSSW_11_1_0_pre7-111X_mcRun3_2021_realistic_v4-v1/GEN-SIM'],
  'n': None,
  'site': [u'T2_CH_CERN', u'T1_US_FNAL_Disk']}]

```

Yes

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
+++ b/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
@@ -81,10 +81,10 @@ class MSOutputTemplate(dict):
             ('isTakenBy', None, (str, unicode)),
             ('OutputDatasets', None, list),
             ('destination', None, list),
-            ('destinationOutputMap', None, dict),
-            ('campaignOutputMap', None, dict),
-            ('transferStatus', None, unicode),
-            ('transferIDs', None, int),
+            ('destinationOutputMap', None, list),
+            ('campaignOutputMap', None, list),
+            ('transferStatus', None, (str, unicode)),
+            ('transferIDs', None, list),
             ('numberOfCopies', None, int)]
 
         self.docTemplate = docTemplate
@@ -231,11 +231,11 @@ class MSOutputTemplate(dict):
             return True
         return False
 
-    def updateDoc(self, myDoc):
+    def updateDoc(self, myDoc, throw=False):
         """
-        Method to be used for updating the document fields from a dictionary
+        A method to be used for updating the document fields from a dictionary
         """
-        if self._checkAttr(self.docTemplate, myDoc, throw=False, update=False, **myDoc):
+        if self._checkAttr(self.docTemplate, myDoc, throw=throw, update=False, **myDoc):
             self.update(myDoc)
             return True
         return False
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -284,41 +284,80 @@ class MSOutput(MSCore):
             #    but an iterator of length 'stride' and then we should be doing:
             #    for workflow in workflows:
             if isinstance(workflow, MSOutputTemplate):
-                if workflow['isRelVal']:
-                    pass
-                else:
-                    try:
+                ddmReqList = []
+                try:
+                    if workflow['isRelVal']:
+                        for dMap in workflow['destinationOutputMap']:
+                            try:
+                                ddmRequest = DDMReqTemplate('copy',
+                                                            item=dMap['datasets'],
+                                                            n=workflow['numberOfCopies'],
+                                                            site=dMap['destination'],
+                                                            group='RelVal')
+                            except KeyError as ex:
+                                # NOTE:
+                                #    If we get to here it is most probably because the 'site'
+                                #    mandatory field to the DDM request is missing (due to an
+                                #    'ALCARECO' dataset or similar). Since this is expected
+                                #    to happen a lot, we'd better just log a warning and continue
+                                msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
+                                msg += "Error: {}".format(ex)
+                                self.logger.warning(msg)
+                                continue
+                            ddmReqList.append(ddmRequest)
+                    else:
+                        # FIXME:
+                        #    We need to create the campaignMap and use it for
+                        #    creating the requests for the nonRelVal workflows
+                        #    it should be also the case when we migrate to Rucio
                         ddmRequest = DDMReqTemplate('copy',
                                                     item=workflow['OutputDatasets'],
                                                     n=workflow['numberOfCopies'],
                                                     site=workflow['destination'])
-                    except Exception as ex:
-                        msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
-                        msg += "Error: {}".format(ex)
-                        self.logger.exception(msg)
-                        return workflow
+                        ddmReqList.append(ddmRequest)
+                except Exception as ex:
+                    msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
+                    msg += "Error: {}".format(ex)
+                    self.logger.exception(msg)
+                    return workflow
 
-                    try:
-                        # In the message bellow we may want to put the list of datasets too
-                        msg = "Making transfer subscriptions for {}".format(workflow['RequestName'])
-                        self.logger.info(msg)
-                        ddmResult = self.ddm.makeRequest(ddmRequest)
-                    except Exception as ex:
-                        msg = "Could not make transfer subscription for Workflow: {}".format(workflow['RequestName'])
-                        msg += "Error: {}".format(ex)
-                        self.logger.exception(msg)
-                        return workflow
+                try:
+                    # In the message bellow we may want to put the list of datasets too
+                    msg = "Making transfer subscriptions for {}".format(workflow['RequestName'])
+                    self.logger.info(msg)
+                    ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
+                except Exception as ex:
+                    msg = "Could not make transfer subscription for Workflow: {}".format(workflow['RequestName'])
+                    msg += "Error: {}".format(ex)
+                    self.logger.exception(msg)
+                    return workflow
 
+                ddmStatusList = ['new', 'activated', 'completed', 'rejected', 'cancelled']
+                transferIDs = []
+                transferStatusList = []
+                for ddmResult in ddmResultList:
                     if 'data' in ddmResult.keys():
-                        self.docKeyUpdate(workflow,
-                                          transferStatus=deepcopy(ddmResult['data'][0]['status']),
-                                          transferIDs=deepcopy(ddmResult['data'][0]['request_id']))
-                        return workflow
-                    else:
-                        msg = "No data found in ddmResult for %s. Either dry run mode or " % workflow['RequestName']
-                        msg += "broken transfer submission to DDM. "
-                        msg += "ddmResult: \n%s" % pformat(ddmResult)
-                        self.logger.warning(msg)
+                        id = deepcopy(ddmResult['data'][0]['request_id'])
+                        status = deepcopy(ddmResult['data'][0]['status'])
+                        transferStatusList.append({'transferID': id,
+                                                   'status': status})
+                        transferIDs.append(id)
+
+                if transferStatusList and all(map(lambda x:
+                                                  True if x['status'] in ddmStatusList else False,
+                                                  transferStatusList)):
+                    self.docKeyUpdate(workflow,
+                                      transferStatus='done',
+                                      transferIDs=transferIDs)
+                    return workflow
+                else:
+                    self.docKeyUpdate(workflow,
+                                      transferStatus='incomplete')
+                    msg = "No data found in ddmResults for %s. Either dry run mode or " % workflow['RequestName']
+                    msg += "broken transfer submission to DDM. "
+                    msg += "ddmResults: \n%s" % pformat(ddmResultList)
+                    self.logger.warning(msg)
+                return workflow
 
             elif isinstance(workflow, (list, set, CommandCursor)):
                 ddmRequests = {}
@@ -446,12 +485,24 @@ class MSOutput(MSCore):
                 # - which are not already taken or
                 # - a transfer subscription have never been done for them and
                 # - avoid retrying workflows in the same cycle
+                # NOTE:
+                #    Once we are running the service not in a dry run mode we may
+                #    consider adding and $or condition in mQueryDict for transferStatus:
+                #    '$or': [{'transferStatus': None},
+                #            {'transferStatus': 'incomplete'}]
+                #    So that we can collect also workflows with partially or fully
+                #    unsuccessful transfers
                 currTime = int(time())
                 treshTime = currTime - self.msConfig['interval']
-                mQueryDict = {'isTaken': False,
-                              'transferStatus': None,
-                              '$or': [{'lastUpdate': None},
-                                      {'lastUpdate': {'$lt': treshTime}}]}
+                mQueryDict = {
+                    '$and': [
+                        {'isTaken': False},
+                        {'$or': [
+                            {'transferStatus': None},
+                            {'transferStatus': 'incomplete'}]},
+                        {'$or': [
+                            {'lastUpdate': None},
+                            {'lastUpdate': {'$lt': treshTime}}]}]}
                 try:
                     pipeLine.run(mQueryDict)
                 except KeyError as ex:
@@ -503,17 +554,21 @@ class MSOutput(MSCore):
         #    to set a destructive function at the end of the pipeline
         # NOTE:
         #    To discuss the collection names
+        # NOTE:
+        #    Here we should never use docUploader with `update=True`, because
+        #    this will erase the latest state of already existing and fully or
+        #    partially processed documents by the Consumer pipeline
         self.logger.info("Running the msOutputProducer ...")
         msPipelineRelVal = Pipeline(name="MSOutputProducer PipelineRelVal",
                                     funcLine=[Functor(self.docTransformer),
                                               Functor(self.docKeyUpdate, isRelVal=True),
-                                              Functor(self.docInfoUpdate),
+                                              Functor(self.docInfoUpdate, pipeLine='PipelineRelVal'),
                                               Functor(self.docUploader, self.msOutRelValColl),
                                               Functor(self.docCleaner)])
         msPipelineNonRelVal = Pipeline(name="MSOutputProducer PipelineNonRelVal",
                                        funcLine=[Functor(self.docTransformer),
                                                  Functor(self.docKeyUpdate, isRelVal=False),
-                                                 Functor(self.docInfoUpdate),
+                                                 Functor(self.docInfoUpdate, pipeLine='PipelineNonRelVal'),
                                                  Functor(self.docUploader, self.msOutNonRelValColl),
                                                  Functor(self.docCleaner)])
         # TODO:
@@ -587,17 +642,90 @@ class MSOutput(MSCore):
         for key, value in kwargs.items():
             try:
                 msOutDoc.setKey(key, value)
+                msOutDoc.updateTime()
             except Exception as ex:
                 msg = "Cannot update key {} for doc: {}\n".format(key, msOutDoc['_id'])
                 msg += "Error: {}".format(str(ex))
                 self.logger.warning(msg)
         return msOutDoc
 
-    def docInfoUpdate(self, msOutDoc):
+    def docInfoUpdate(self, msOutDoc, pipeLine=None):
         """
         A function intended to fetch and fill into the document all the needed
         additional information like campaignOutputMap etc.
         """
+
+        # Fill the destinationOutputMap first
+        if msOutDoc['isRelVal']:
+            destinationOutputMap = []
+            wflowDstSet = set()
+            updateDict = {}
+            for dataset in msOutDoc['OutputDatasets']:
+                _, dsn, procString, dataTier = dataset.split('/')
+                destination = set()
+                if dataTier != "RECO" and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                if dataTier == "GEN-SIM":
+                    destination.add('T1_US_FNAL_Disk')
+                if dataTier == "GEN-SIM-DIGI-RAW":
+                    destination.add('T1_US_FNAL_Disk')
+                if dataTier == "GEN-SIM-RECO":
+                    destination.add('T1_US_FNAL_Disk')
+                if "RelValTTBar" in dsn and "TkAlMinBias" in procString and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                if "MinimumBias" in dsn and "SiStripCalMinBias" in procString and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                dMap = {'datasets': [dataset],
+                        'destination': list(destination)}
+                destinationOutputMap.append(dMap)
+                wflowDstSet |= destination
+
+            # here we try to aggregate the destination map per destination
+            aggDstMap = []
+
+            # populate the first element in the aggregated list
+            aggDstMap.append(destinationOutputMap.pop())
+
+            # feed the rest
+            while len(destinationOutputMap) != 0:
+                dMap = destinationOutputMap.pop()
+                found = False
+                for aggMap in aggDstMap:
+                    if set(dMap['destination']) == set(aggMap['destination']):
+                        # Check if the two objects are not references to one and the
+                        # same object. Only then copy the values of the dMap,
+                        # otherwise we will enter an endless cycle.
+                        if dMap is not aggMap:
+                            for i in dMap['datasets']:
+                                aggMap['datasets'].append(i)
+                        found = True
+                        del(dMap)
+                        break
+                if not found:
+                    aggDstMap.append(dMap)
+
+            # finally reassign the destination map with the aggregated one
+            destinationOutputMap = aggDstMap
+
+            wflowDstList = list(wflowDstSet)
+            updateDict['destination'] = wflowDstList
+            updateDict['destinationOutputMap'] = destinationOutputMap
+            try:
+                msOutDoc.updateDoc(updateDict, throw=True)
+                # msg = "%s: %s: 'msOutDoc': %s"
+                # self.logger.debug(msg,
+                #                   self.currThreadIdent,
+                #                   pipeLine,
+                #                   pformat(msOutDoc))
+            except Exception as ex:
+                msg = "%s: %s: Could not update the additional information for "
+                msg += "'msOutDoc' with '_id': %s \n"
+                msg += "Error: %s"
+                self.logger.exception(msg,
+                                      self.currThreadIdent,
+                                      pipeLine,
+                                      msOutDoc['_id'],
+                                      str(ex))
         return msOutDoc
 
     def docUploader(self, msOutDoc, dbColl, update=False, keys=None, stride=None):
@@ -632,7 +760,7 @@ class MSOutput(MSCore):
             if not keys:
                 keys = []
 
-            # update only the requested keyes:
+            # update only the requested keys:
             if update and keys:
                 updateDict = {}
                 for key in keys:
--- a/src/python/WMCore/Services/DDM/DDM.py
+++ b/src/python/WMCore/Services/DDM/DDM.py
@@ -43,6 +43,7 @@ class DDMReqTemplate(dict):
                          ('group', 'DataOps', str),
                          ('n', None, int),
                          ('cache', None, str)]
+            required = ['item', 'site']
 
         elif api == 'pollcopy':
             tempTuple = [('request_id', None, int),
@@ -50,9 +51,11 @@ class DDMReqTemplate(dict):
                          ('site', None, list),
                          ('status', None, str),
                          ('user', None, str)]
+            required = ['request_id']
 
         elif api == 'cancelcopy':
             tempTuple = [('request_id', None, int)]
+            required = ['request_id']
 
         else:
             msg = "ERROR: Unsupported API: {}".format(api)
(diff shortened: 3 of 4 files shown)


src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
```diff
@@ -231,11 +231,11 @@ def setRelVal(self, isRelVal):
...
         return False
 
-    def updateDoc(self, myDoc):
+    def updateDoc(self, myDoc, throw=False):
         """
-        Method to be used for updating the document fields from a dictionary
+        A method to be used for updating the document fields from a dictionary
         """
-        if self._checkAttr(self.docTemplate, myDoc, throw=False, update=False, **myDoc):
+        if self._checkAttr(self.docTemplate, myDoc, throw=throw, update=False, **myDoc):
```
amaltaro, line 238: Can you please remind me why we would want NOT to throw an exception if the key is not in the template or not of the same data type?

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,56 @@ def makeSubscriptions(self, workflow):
...
+                    msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
+                    msg += "Error: {}".format(ex)
+                    self.logger.exception(msg)
+                    return workflow
 
+                try:
+                    # In the message bellow we may want to put the list of datasets too
+                    msg = "Making transfer subscriptions for {}".format(workflow['RequestName'])
+                    self.logger.info(msg)
+                    ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
```
amaltaro, line 313: Imagine that this `makeAggRequests` actually makes 3 calls to DDM. Where 1 fails and the other 2 return 200 response. How do you deal with the request that failed here? From L322, I'd say the code won't even noticed it, provided at least 1 request went through.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -503,17 +518,21 @@ def msOutputProducer(self, requestRecords):
         #    to set a destructive function at the end of the pipeline
         # NOTE:
         #    To discuss the collection names
+        # NOTE:
```
amaltaro, line 521: I don't get this comment. You do add self.docUploader to the pipeline here. What am I missing?

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -593,11 +612,56 @@ def docKeyUpdate(self, msOutDoc, **kwargs):
                 self.logger.warning(msg)
         return msOutDoc
 
-    def docInfoUpdate(self, msOutDoc):
+    def docInfoUpdate(self, msOutDoc, pipeLine=None):
```
amaltaro, line 615: this new argument seems to be used only for the exception code. How about we print the whole document in the exception (where one can see whether it was a RelVal or standard workflow.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -593,11 +612,56 @@ def docKeyUpdate(self, msOutDoc, **kwargs):
...
+
+        # Fill the destinationOutputMap first
+        if msOutDoc['isRelVal']:
+            destinationOutputMap = []
+            wflowDstSet = set()
+            updateDict = {}
+            for dataset in msOutDoc['OutputDatasets']:
+                _, dsn, procString, dataTier = dataset.split('/')
+                destination = set()
+                if dataTier != "RECO" and dataTier != "ALCARECO":
```
amaltaro, line 629: We should check how Unified deals with RECO and ALCARECO datasets.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -593,11 +612,56 @@ def docKeyUpdate(self, msOutDoc, **kwargs):
...
+                if dataTier == "GEN-SIM-RECO":
+                    destination.add('T1_US_FNAL_Disk')
+                if "RelValTTBar" in dsn and "TkAlMinBias" in procString and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                if "MinimumBias" in dsn and "SiStripCalMinBias" in procString and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                dMap = {'datasets': [dataset],
+                        'destination': list(destination)}
+                destinationOutputMap.append(dMap)
+                wflowDstSet = wflowDstSet | destination
```
amaltaro, line 644:
```suggestion
                wflowDstSet.add(destination)
```
is cleaner and clearer.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -593,11 +612,56 @@ def docKeyUpdate(self, msOutDoc, **kwargs):
...
+                    destination.add('T1_US_FNAL_Disk')
+                if dataTier == "GEN-SIM-RECO":
+                    destination.add('T1_US_FNAL_Disk')
+                if "RelValTTBar" in dsn and "TkAlMinBias" in procString and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                if "MinimumBias" in dsn and "SiStripCalMinBias" in procString and dataTier != "ALCARECO":
+                    destination.add('T2_CH_CERN')
+                dMap = {'datasets': [dataset],
+                        'destination': list(destination)}
+                destinationOutputMap.append(dMap)
```
amaltaro, line 643: As we discussed over slack, this will add one dict per dataset, while we should aggregate datasets going to the same destination. Change should be fairly simple, so I'd rather have it implemented in this PR.

amaltaro (changes requested):
Changes look good in general, Todor. However, I have made a few suggestions in the code that would be important to get addressed here.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,56 @@ def makeSubscriptions(self, workflow):
...
-                        self.docKeyUpdate(workflow,
-                                          transferStatus=deepcopy(ddmResult['data'][0]['status']),
-                                          transferIDs=deepcopy(ddmResult['data'][0]['request_id']))
-                        return workflow
-                    else:
-                        msg = "No data found in ddmResult for %s. Either dry run mode or " % workflow['RequestName']
-                        msg += "broken transfer submission to DDM. "
-                        msg += "ddmResult: \n%s" % pformat(ddmResult)
-                        self.logger.warning(msg)
+                        transferStatus.append(deepcopy(ddmResult['data'][0]['status']))
```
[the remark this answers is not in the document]
todor-ivanov: @amaltaro  it is a good question. I need to check. And actually going through the Dynamo code here is this list I find on their side: [1]. So it should be one of:
```
['', 'new', 'activated', 'completed', 'rejected', 'cancelled']
```

[1]
https://github.com/SmartDataProjects/dynamo/blob/911f90c88702c9f7ed7a885744028c5705543e88/lib/dataformat/request.py#L18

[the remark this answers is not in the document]
amaltaro: honestly, given that we do not plan to keep monitoring these output data placement requests. I'd get rid of it completely.

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
todor-ivanov: It is a method supposed to be used from outside the class, so since here we do not know how the user may want to handle those exceptions (if at all), I left the option for him/her to chose to either throw them or not. In general if it does, and the exception is not handled on the next step it will break the current run cycle. And there are cases where failing one and only one document is preferable than breaking the whole thread. (of course it got me into troubles when I tested it, because if not checked the exit status from this method, it may lead to silent errors  - but this is usually a responsibility of whom ever is about to call it).

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I may need your help here with that. I cannot find anything else than what I have already implemented. Maybe a fresh look on the problem from another person, can reveal things I am constantly overlooking.

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
amaltaro: This method is actually a bit confusing. From the name, one can say nothing is supposed to get written, but just checked.
The "user" here seems to be the module itself, and there shouldn't be any other users. So I would actually implement what we need and what we will use, instead of providing flexibility that will never get used.

This can be addressed in a different PR though (setting it as unresolved).

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Agreed and Done.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,56 @@ def makeSubscriptions(self, workflow):
...
+                    if workflow['isRelVal']:
+                        for dMap in workflow['destinationOutputMap']:
+                            ddmRequest = DDMReqTemplate('copy',
+                                                        item=dMap['datasets'],
+                                                        n=workflow['numberOfCopies'],
+                                                        site=dMap['destination'],
+                                                        group='RelVal')
+                            ddmReqList.append(ddmRequest)
+                    else:
                         ddmRequest = DDMReqTemplate('copy',
```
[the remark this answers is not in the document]
todor-ivanov: Agreed && Done.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Yes, I know, but could not find a way out of this that satisfies me enough. The alternative I was considering was having the transfer status as a list of dictionaries like the following: [1]. This will generally lead to information repetition, but will help if we need to do a reverse map of transfer Id to reqest status per workflow, and eventually in the next step when we start aggregate for many workflows..... but there (once we cross the boundaries of the workflow) things get messy anyway. What do you think?

[1]
```
transferstatus: [{'transferId': transferId1,
                  'status': status1},
                {'transferId': transferId2,
                  'status': status2},
...] 
```

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: The comment is not regarding the `docUploader()` function itself. It is why it is added with the default value of the `update=False` parameter. It is just a reminder, because I was quite close earlier this morning to fool myself  and just put `update=True` here so that the newly generated `destinationMap` can be updated without the need for deleting the whole MongoDB during deployment time. But this would break the logic for the Consumer thread then.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: The document is already fully logged in debug  mode at the end of the consumer pipeline..... Which is not exactly relevant to the current message, but simply pointing out that there are already way to many times one and the same document will be fully dumped in the log. I am more for actually making our way of handling exceptions from inside the Pipeline so we can log them properly and in a consistent manner or at least to go through the whole code and make all the messages follow the same schema, like: [1]. I lately find it quite difficult to relate some of the old messages that are not following [1] to the proper module pipeline and workflow. 
[1]
```
"%s: %s: Some message regarding msOutDoc with '_id': %s" % (self.currThreadIdent, pipelineName, msOutDoc['_id'])
```

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
todor-ivanov: It was develpped in a quite early stage and the usage slightly changed afterwards. I would say if we start messing with that this may shake the ground for the whole Class, since it is constantly referred from any update method from the Class itself.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Done

todor-ivanov:
@amaltaro I think that my final commit adressess almost all of your comments please take a look at it, and give me your opinion, so I can squash the two commits before we merge.

todor-ivanov:
Test this please

todor-ivanov:
Test this please

todor-ivanov:
Test this please

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,80 @@ def makeSubscriptions(self, workflow):
...
+                try:
+                    if workflow['isRelVal']:
+                        for dMap in workflow['destinationOutputMap']:
+                            try:
+                                ddmRequest = DDMReqTemplate('copy',
+                                                            item=dMap['datasets'],
+                                                            n=workflow['numberOfCopies'],
+                                                            site=dMap['destination'],
+                                                            group='RelVal')
+                            except KeyError as ex:
```
amaltaro, line 297: I would actually put a check for it, instead of assuming that the KeyError comes from the lack of value for the `site` attribute. Unless, the only possibility to raise such KeyError would come from that.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,80 @@ def makeSubscriptions(self, workflow):
...
-                        msg += "broken transfer submission to DDM. "
-                        msg += "ddmResult: \n%s" % pformat(ddmResult)
-                        self.logger.warning(msg)
+                        id = deepcopy(ddmResult['data'][0]['request_id'])
+                        status = deepcopy(ddmResult['data'][0]['status'])
+                        transferStatusList.append({'transferID': id,
+                                                   'status': status})
+                        transferIDs.append(id)
+
+                if transferStatusList and all(map(lambda x:
```
amaltaro, line 346: replace map built-in by list comprehension (as reported in the jenkins results)

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,80 @@ def makeSubscriptions(self, workflow):
...
+                    msg = "Could not make transfer subscription for Workflow: {}".format(workflow['RequestName'])
+                    msg += "Error: {}".format(ex)
+                    self.logger.exception(msg)
+                    return workflow
 
+                ddmStatusList = ['new', 'activated', 'completed', 'rejected', 'cancelled']
+                transferIDs = []
+                transferStatusList = []
+                for ddmResult in ddmResultList:
                     if 'data' in ddmResult.keys():
```
amaltaro, line 339: Will a DDM http request always return the `data` key?
I still wonder if we could hit the same issue as I reported above, some tests going through, some not, and we marking the workflow as done.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,80 @@ def makeSubscriptions(self, workflow):
...
+                    msg = "Making transfer subscriptions for {}".format(workflow['RequestName'])
+                    self.logger.info(msg)
+                    ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
+                except Exception as ex:
+                    msg = "Could not make transfer subscription for Workflow: {}".format(workflow['RequestName'])
+                    msg += "Error: {}".format(ex)
+                    self.logger.exception(msg)
+                    return workflow
 
+                ddmStatusList = ['new', 'activated', 'completed', 'rejected', 'cancelled']
```
amaltaro, line 335: these status don't belong in here.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -284,41 +284,80 @@ def makeSubscriptions(self, workflow):
...
-                        self.docKeyUpdate(workflow,
-                                          transferStatus=deepcopy(ddmResult['data'][0]['status']),
-                                          transferIDs=deepcopy(ddmResult['data'][0]['request_id']))
-                        return workflow
-                    else:
-                        msg = "No data found in ddmResult for %s. Either dry run mode or " % workflow['RequestName']
-                        msg += "broken transfer submission to DDM. "
-                        msg += "ddmResult: \n%s" % pformat(ddmResult)
-                        self.logger.warning(msg)
+                        id = deepcopy(ddmResult['data'][0]['request_id'])
```
amaltaro, line 340: isn't it an immutable object (either integer or string)? No need for deepcopy it (same for the line below)

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: BTW, you do not check the response from DDM, just that it returned any known status.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -587,17 +642,90 @@ def docKeyUpdate(self, msOutDoc, **kwargs):
...
-    def docInfoUpdate(self, msOutDoc):
+    def docInfoUpdate(self, msOutDoc, pipeLine=None):
         """
         A function intended to fetch and fill into the document all the needed
         additional information like campaignOutputMap etc.
         """
+
+        # Fill the destinationOutputMap first
+        if msOutDoc['isRelVal']:
+            destinationOutputMap = []
```
amaltaro, line 660: and this logic should go under the template IMO. We provide a reqmgr2 document, we get the final mongodb document.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -587,17 +642,90 @@ def docKeyUpdate(self, msOutDoc, **kwargs):
...
+                dMap = {'datasets': [dataset],
+                        'destination': list(destination)}
+                destinationOutputMap.append(dMap)
+                wflowDstSet |= destination
+
+            # here we try to aggregate the destination map per destination
+            aggDstMap = []
+
+            # populate the first element in the aggregated list
+            aggDstMap.append(destinationOutputMap.pop())
```
amaltaro, line 687: Way too much complicated way to aggregate destinations/datasets. One could use a dict with frozenset keys. Once done, we could then expand that to a list and be done with it (like what I did to the campaign/output map).

amaltaro (changes requested):
Thanks for the extended effort Todor. I still think there is a bunch of changes and/or improvements to be made here.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I implicitly made it to be the case with [1]. There is only one other place where a `KeyError`  is generated [2], but we can never hit it because we generate the keys in this call by hand, only the values are taken from the document but they may generate only `TypeError`.

[1]
https://github.com/dmwm/WMCore/blob/296b118889cb05ad35e486f4873b56b4d1c7168d/src/python/WMCore/Services/DDM/DDM.py#L105

[2]
https://github.com/dmwm/WMCore/blob/296b118889cb05ad35e486f4873b56b4d1c7168d/src/python/WMCore/Services/DDM/DDM.py#L85

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: The list comprehension is ready now. 
> you do not check the response from DDM

The DDM object always returns a response. Whether it will contain [data] depends on the run mode. If no Subscriptions are made for real the returned object is the DDMrequest itself (no [data] key here) and if the run mode is real then the actual result from the request is returned

amaltaro:
please squash

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: If the request is actually done the 'data' field will be there (unless DDM changes its interface to the outer world). Here is an example [1]. And if in case of connectivity problems or something like that, the response is incomplete and we fail to fetch the 'data' - even then we cannot mark the Workflow as done because it will fail to accumulate the statuses few lines bellow, so the Workflow will be retired on the next cycle.


[1]
```
{u'data': [{u'dn': u'/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch',
            u'first_request': u'2020-05-25T14:20:15 UTC',
            u'group': u'DataOps',
            u'item': [u'/LQLQToTopMuTopTau_M-1200_TuneCP5_13TeV_pythia8/RunIIFall17NanoAODv5-PU2017_12Apr2018_Nano1June2019_102X_mc2017_realistic_v7-v1/NANOAODSIM'],
            u'last_request': u'2020-05-25T14:20:33 UTC',
            u'n': 1,
            u'request_count': 2,
            u'request_id': 47373,
            u'site': [u'T1_UK_RAL_Disk',
                      u'T2_PL_Swierk',
                      u'T2_RU_INR',
                      u'T2_BE_IIHE',
                      u'T2_US_MIT',
...
                      u'T1_IT_CNAF_Disk'],
            u'status': u'new',
            u'user': u'dmwm_alandev@cmsweb.cern.ch'}],
 u'message': u'',
 u'result': u'OK'}
```

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Do you mean that the place is wrong or the very fact that I refer to them in anyway? If the former I am all open for ideas.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I just wanted to stay on the safe side - it shouldn't break if it is or not an immutable, though.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Actually my understanding is exactly the opposite. My  origina idea was to have as minimum additional methods as possible in the Template and the ones that are there to be related only to validation checks and general components updates of the current object - in order to minimize the Memory footprint again. Because every additional method is about to be carried with the object through out it's whole lifetime. and once we start working in batches of Workflows at a time.... it may become noticeable.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Need to admit - never tried it before.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Anyway I just added few  additional checks.

PR #9719, opened 2020-06-03
Create a map of output datasets by campaign for the MSOutput component

### LINKED ISSUE #9718: MS: build map of output dataset to campaign name
**Impact of the new feature**
MSOutput

Not a problem, but as discussed in this issue:
https://github.com/dmwm/WMCore/issues/9610

we should build a map of output datasets for each campaign defined in a workflow. This map will be used to properly subscribe the output datasets, according to the campaign configuration.

Build a map of output datasets associated to each workflow campaign. The final result must be in this structure:
```
"campaignOutputMap": [{"campaignName": "blah", "datasets": ["list of datasets"]},
                                          {"campaignName": "blah", "datasets": ["list of datasets"]}, etc],
```
Eventually we need to look up at which information is actually used from the campaign configuration. It looks like Unified uses a bunch of default values, instead of using information from the config level.

**Additional context**
none

amaltaro:
Fixes #9718 

#### Status
not-tested

#### Description
When MSOutputTemplate is called to convert a ReqMgr2 request dictionary into our MSOutput internal object, perform a map of output datasets in each campaign and persist this information in the MSOutput object.
Note that StepChain and TaskChain requests can have many campaign names, while all the others are single campaign'ed requests.

no, in the sense of being a new feature.

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
+++ b/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
@@ -82,7 +82,7 @@ class MSOutputTemplate(dict):
             ('OutputDatasets', None, list),
             ('destination', None, list),
             ('destinationOutputMap', None, dict),
-            ('campaignOutputMap', None, dict),
+            ('campaignOutputMap', None, list),
             ('transferStatus', None, unicode),
             ('transferIDs', None, int),
             ('numberOfCopies', None, int)]
@@ -113,6 +113,9 @@ class MSOutputTemplate(dict):
         # if any object attribute is passed as a **kwargs parameter then
         # overwrite the equivalent parameter which was coming with the doc
         self._checkAttr(docTemplate, myDoc, update=True, throw=True, **kwargs)
+
+        if doc is not None:
+            self._setCampMap(doc, myDoc)
 
         # enforce a full check on the final document
         self._checkAttr(docTemplate, myDoc, update=False, throw=True, **myDoc)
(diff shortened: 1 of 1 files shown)


amaltaro:
@todor-ivanov this still needs to be tested, and possibly integrated into the changes you've been working on. Could you please review and, we either merge it before your PR, such that you can test it; or you cherry pick this PR and test it with the rest of your changes?

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
```diff
@@ -240,11 +243,31 @@ def updateDoc(self, myDoc):
...
+                campOutputMap[campName].extend(doc["ChildDsets"])
+            # now just write the final data structure out
+            for campName, dsetList in campOutputMap.items():
+                finalMap.append(dict(campaignName=campName, datasets=dsetList))
+        else:
+            # ReReco and StoreResults
+            finalMap.append(dict(campaignName=doc["Campaign"], datasets=doc["OutputDatasets"]))
+
+        # finally, update this object
+        self.update({"campaignOutputMap": finalMap})
```
todor-ivanov, line 270: Please do not go for direct update of the object without passing the regular check of self consistency. Please use the `_checkAttr()`  internal method in a way similar to [1] if you plan to have this as a 'public' method exposed to the outer world. And always return `True` or `False` as an outcome of the function. The Alternative is if you plan this one to be used only as an internal method and rely on the following similar check that is happening in the constructor later [2], then please rename it to `_setCampMap()`. But still return a value from the function. We may also consider renaming the parameter to `myDoc` so to be more clear that while in the function we are referring to the document passed as an argument to the function but not to any other (just for better readability).



[1]
https://github.com/dmwm/WMCore/pull/9719/commits/fca0a0893a7d7ac719a8e8c9d9a53ddd6ecf118a#diff-40f3495cc16a13ae8fd7449995945ce1R222

[2]
https://github.com/dmwm/WMCore/pull/9719/commits/fca0a0893a7d7ac719a8e8c9d9a53ddd6ecf118a#diff-40f3495cc16a13ae8fd7449995945ce1R121

todor-ivanov (changes requested):
Alan @amaltaro , this is awesome. With a single push you just took it two steps closer to the end of the module. Thanks a lot! I made some comments inline, please take a short look and give me your feed back.

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
```diff
@@ -87,6 +87,9 @@ def __init__(self, doc=None, **kwargs):
             ('transferIDs', None, int),
             ('numberOfCopies', None, int)]
 
+        if doc is not None:
```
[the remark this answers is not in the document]
amaltaro: Todor, I cannot see the diff link. Can you please provide me the line number?

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
amaltaro: I'm going to use the "private" method name then. About the boolean outcome, I think it makes the code less readable and unnecessarily more complex. I'm in favor of simply enforcing a schema here (or raise exceptions upstream). Anyways, this can be followed up in the coming days/weeks.

amaltaro:
Thanks Todor. I have pushed the changes you suggested in a second commit (also fixed a bug in the code).

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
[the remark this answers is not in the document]
todor-ivanov: Sorry Alan, obviously the diff is not the right method of citing code :) Here is a permanent link:
[1]
 https://github.com/dmwm/WMCore/blob/fca0a0893a7d7ac719a8e8c9d9a53ddd6ecf118a/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py#L107

src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
todor-ivanov: This seems reasonable to me too.

todor-ivanov:
Thanks  @amaltaro ! Once I am done with this change here [1] I will cherrypick this one on top of [1] and will test the final result. I will post the result her so we can merge this one too. 


[1]
 https://github.com/dmwm/WMCore/pull/9717

todor-ivanov (approved):
It is good. Needs to be squashed before merge, though.

Issue #9676, opened 2020-05-06
Cannot inject work against PNNless resources


amaltaro:
**Impact of the bug**
WMAgent (either getting stuck in WorkQueueManager, or completely skipping such opportunistic resources)

**Describe the bug**
The CMS@Home agent got redeployed yesterday, and with that a new issue got exposed.
It's related to the fix made in this PR:
https://github.com/dmwm/WMCore/pull/9659/files#diff-7841c008d28dc37b9c039eef99e13610R434

which is supposed to only send top level jobs where there is a common data location (primary + secondary + parent).
The side effect is, it validates the `commonSites` against the information in CRIC (thus mapping PSN to PNN), which of course will fail for most of the opportunistic resources, given that they do not have a storage endpoint and instead rely on other sites remote stage out.

**How to reproduce it**
Assign a MC workflow against one opportunistic resource (T3_CH_Volunteer, T3_US_NERSC, etc). Of course, the agent needs to have those resources in the resource-control.

**Expected behavior**
We need to revisit which CRIC APIs to use, such that:
* this work would still go through to the opportunistic resources
* and that the agent would automatically add such resources to the local database (thus, also considering PNNless resources)

**Additional context and error message**
I have manually applied changes to the vocms0267 (CMS@Home agent) such that work can go through..

Traceback from WorkQueueManager:
```
2020-05-05 20:33:08,791:140536444503808:INFO:WorkQueueBackend:Getting up to 100 available work from http://vocms0267.cern.ch:5984
2020-05-05 20:33:08,807:140536444503808:INFO:WorkQueue:Got 1 elements matching the constraints
2020-05-05 20:33:08,908:140536444503808:INFO:WorkQueue:Running WMBS preparation for ireid_TC_SLC7_IDR_CMS_Home_200505_201537_9548 with ParentQueueId 63dbbb9f098d5119fc8bf38123f3ecba,
  with common location []
2020-05-05 20:33:08,962:140536444503808:INFO:Fileset:Fileset created: ireid_TC_SLC7_IDR_CMS_Home_200505_201537_9548-SinglePiE50HCAL_pythia8_2018_GenSimFull-50cb422cbd94edda4fec6dd36fe0d621
...
2020-05-05 20:33:12,022:140536444503808:ERROR:WorkQueue:Failed to create subscription for ireid_TC_SLC7_IDR_CMS_Home_200505_201537_9548 with block name None
Error: <@========== WMException Start ==========@>
[...]

PR #9729, opened 2020-06-09
Use PSN as PNN for opportunistic diskless resources - wmagent branch

amaltaro:
wmagent branch version of #9676 

I'm pushing this patch in in submit6 and submit7, which have been crashing since I inserted the NERSC/OSG sites in the resource control.

### DIFF
--- a/src/python/WMCore/Services/CRIC/CRIC.py
+++ b/src/python/WMCore/Services/CRIC/CRIC.py
@@ -168,11 +168,12 @@ class CRIC(Service):
                 self["logger"].debug("No PSNs for PNN: %s" % pnn)
         return list(psns)
 
-    def PSNstoPNNs(self, psns):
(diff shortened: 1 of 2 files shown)


amaltaro:
This
```
2020-06-09 05:44:40,516:140350260205312:INFO:WorkQueue:Running WMBS preparation for dmytro_task_TOP-RunIIFall17wmLHEGS-00419__v1_T_200608_173739_3793 with ParentQueueId 9c5b65750f64613ac8e7257c9359c683,
  with common location [u'T3_US_NERSC']
```

is something that I'm afraid it can cause issues to ACDC workflows, because documents would get uploaded with NERSC as a file location. The same would apply to any other site that does not have any storage defined as `phedex-node` in CRIC...

amaltaro:
The last agents running 1.3.3.patchX are about to go away from the system, so I'm closing this PR.

PR #9731, opened 2020-06-09
[py2py3] futurize on WMCore/src/python/Utils

### LINKED ISSUE #9730: [py2py3] Utils should be compatible with both py2 and py3
**Impact of the new feature**

WMCore/src/python/Utils

Make Utils package compatible with both pyhton2 and python3

Run futurize on all the modules of the Utils package and check if any manual change is required to pass the unit tests.

mapellidario:
Fixes #9730 

#### Status
ready

#### Description

Run futurize on the modules in the Utils package

The following files have been automatically modified with futurize.
Only some minor manual tweaks have been necessary to make all the
unit tests pass.

	modified:   src/python/Utils/CertTools.py
	modified:   src/python/Utils/EmailAlert.py
	modified:   src/python/Utils/ExtendedUnitTestCase.py
	modified:   src/python/Utils/IteratorTools.py
	modified:   src/python/Utils/MathUtils.py
	modified:   src/python/Utils/MemoryCache.py
	modified:   src/python/Utils/Pipeline.py
	modified:   src/python/Utils/ProcessStats.py
	modified:   src/python/Utils/Signals.py
	modified:   src/python/Utils/Throttled.py
	modified:   src/python/Utils/Timers.py
	modified:   src/python/Utils/Utilities.py

A log of the changes is available at https://cms-dmwm-test.docs.cern.ch/py2py3/porting_log/000_9731_utils/

yes

#### Related PRs

None

### DIFF
--- a/src/python/Utils/CertTools.py
+++ b/src/python/Utils/CertTools.py
@@ -1,4 +1,5 @@
 from __future__ import division, print_function
+from builtins import str
 import os
 
 
--- a/src/python/Utils/EmailAlert.py
+++ b/src/python/Utils/EmailAlert.py
@@ -3,11 +3,12 @@ EmailAlert - send alerts via email
 """
 
 from __future__ import division
+from builtins import str, object
 import smtplib
 import logging
 
 
-class EmailAlert:
+class EmailAlert(object):
     """
     A simple class to send alerts via email
     """
--- a/src/python/Utils/ExtendedUnitTestCase.py
+++ b/src/python/Utils/ExtendedUnitTestCase.py
@@ -4,6 +4,7 @@ Unit testing base class with our extensions
 """
 
 from __future__ import (division, print_function)
+from future.utils import viewitems
 
 import copy
 import unittest
@@ -23,7 +24,7 @@ class ExtendedUnitTestCase(unittest.TestCase):
         """
 
         def traverse_dict(dictionary):
-            for key, value in dictionary.iteritems():
+            for key, value in viewitems(dictionary):
                 if isinstance(value, dict):
                     traverse_dict(value)
                 elif isinstance(value, list):
@@ -39,7 +40,7 @@ class ExtendedUnitTestCase(unittest.TestCase):
             theList.sort()
             return
 
-        if type(expected_obj) != type(actual_obj):
+        if not isinstance(expected_obj, type(actual_obj)):
             self.fail(msg="The two objects are different type and cannot be compared: %s and %s" % (
             type(expected_obj), type(actual_obj)))
 
--- a/src/python/Utils/IteratorTools.py
+++ b/src/python/Utils/IteratorTools.py
@@ -1,9 +1,10 @@
 #! /usr/bin/env python
 from __future__ import division, print_function
 
+from future.utils import viewitems
+from builtins import str, map
 import collections
-from itertools import islice
-from itertools import chain
+from itertools import islice, chain
 
 def grouper(iterable, n):
     """
@@ -30,7 +31,7 @@ def nestedDictUpdate(d, u):
     Code from Alex Matelli
     http://stackoverflow.com/questions/3232943/update-value-of-a-nested-dictionary-of-varying-depth
     """
-    for k, v in u.iteritems():
+    for k, v in viewitems(u):
         if isinstance(v, collections.Mapping):
             r = nestedDictUpdate(d.get(k, {}), v)
             d[k] = r
@@ -38,16 +39,16 @@ def nestedDictUpdate(d, u):
             d[k] = u[k]
     return d
 
-def convertFromUnicodeToStr(data):
+def convertFromUnicodeToBytes(data):
     """
     code fram
     http://stackoverflow.com/questions/1254454/fastest-way-to-convert-a-dicts-keys-values-from-unicode-to-str
     """
-    if isinstance(data, basestring):
-        return str(data)
+    if isinstance(data, str):
+        return data.encode('utf-8')
     elif isinstance(data, collections.Mapping):
-        return dict(list(map(convertFromUnicodeToStr, data.iteritems())))
+        return dict(list(map(convertFromUnicodeToBytes, viewitems(data))))
     elif isinstance(data, collections.Iterable):
-        return type(data)(list(map(convertFromUnicodeToStr, data)))
+        return type(data)(list(map(convertFromUnicodeToBytes, data)))
     else:
         return data
--- a/src/python/Utils/MathUtils.py
+++ b/src/python/Utils/MathUtils.py
@@ -4,6 +4,8 @@ Module containing mathematical and physics utils
 """
 from __future__ import division, print_function
 
+from builtins import int
+from past.builtins import basestring
 from math import ceil
 
 
@@ -15,11 +17,11 @@ def quantize(inputVal, quanta):
     """
     if isinstance(inputVal, basestring):
         inputVal = float(inputVal)
-    elif not isinstance(inputVal, (int, float, long)):
+    elif not isinstance(inputVal, (int, float)):
         msg = "Input value has to be either int or float, not %s" % (type(inputVal))
         raise ValueError(msg)
 
-    if isinstance(quanta, (basestring, float, long)):
+    if isinstance(quanta, (basestring, int, float)):
         quanta = int(float(quanta))
     elif not isinstance(quanta, int):
         msg = "Quanta value has to be either int or float, not %s" % (type(quanta))
--- a/src/python/Utils/ProcessStats.py
+++ b/src/python/Utils/ProcessStats.py
@@ -29,6 +29,8 @@ upteima and cpu/mem/threads information.
 from __future__ import print_function
 from __future__ import division
 
+from future.utils import viewitems
+
 # system modules
 import os
 import sys
@@ -83,7 +85,7 @@ def threadStack():
     tdict = {}
     id2name = dict([(th.ident, th.name) for th in threading.enumerate()])
     threads = []
-    for tid, stack in sys._current_frames().items():
+    for tid, stack in viewitems(sys._current_frames()):
         tdict = {"thead": id2name.get(tid, ""), "thead_id": tid}
         stacklist = []
         for filename, lineno, name, line in traceback.extract_stack(stack):
--- a/src/python/Utils/Signals.py
+++ b/src/python/Utils/Signals.py
@@ -10,7 +10,9 @@ running program will receive SIGUSR1 signal. E.g., in unix shell
 just do
 shell# kill -s SIGUSR1 <pid>
 """
+# future
 from __future__ import print_function, division
+from future.utils import viewitems
 
 # system modules
 import sys
@@ -26,7 +28,7 @@ def dumpthreads(isignal, iframe):
     print("DAS stack, signal=%s, frame=%s" % (isignal, iframe))
     id2name = dict([(th.ident, th.name) for th in threading.enumerate()])
     code = []
-    for tid, stack in sys._current_frames().items():
+    for tid, stack in viewitems(sys._current_frames()):
         code.append("\n# Thread: %s(%d)" % (id2name.get(tid,""), tid))
         for filename, lineno, name, line in traceback.extract_stack(stack):
             code.append('File: "%s", line %d, in %s' % (filename, lineno, name))
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -1,7 +1,10 @@
 #! /usr/bin/env python
 
 from __future__ import division, print_function
+from future.utils import viewitems
 
+from builtins import str
+from past.builtins import basestring
 import subprocess
 import os
 import re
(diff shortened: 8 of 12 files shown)


amaltaro:
test this please

src/python/Utils/EmailAlert.py
```diff
@@ -3,11 +3,13 @@
 """
 
 from __future__ import division
+from builtins import str
```
amaltaro, line 6: Please group these imports.

src/python/Utils/ProcessStats.py
```diff
@@ -83,7 +83,7 @@ def threadStack():
     tdict = {}
     id2name = dict([(th.ident, th.name) for th in threading.enumerate()])
     threads = []
-    for tid, stack in sys._current_frames().items():
+    for tid, stack in list(sys._current_frames().items()):
```
amaltaro, line 86: can't cast to list here

src/python/Utils/MathUtils.py
```diff
@@ -15,11 +16,11 @@ def quantize(inputVal, quanta):
     """
     if isinstance(inputVal, basestring):
         inputVal = float(inputVal)
-    elif not isinstance(inputVal, (int, float, long)):
+    elif not isinstance(inputVal, (int, float)):
         msg = "Input value has to be either int or float, not %s" % (type(inputVal))
         raise ValueError(msg)
 
-    if isinstance(quanta, (basestring, float, long)):
+    if isinstance(quanta, (basestring, float, int)):
```
amaltaro, line 23: I'm afraid this might break something. Many of the MySQL queries return long data.

src/python/Utils/Signals.py
```diff
@@ -26,7 +26,7 @@ def dumpthreads(isignal, iframe):
     print("DAS stack, signal=%s, frame=%s" % (isignal, iframe))
     id2name = dict([(th.ident, th.name) for th in threading.enumerate()])
     code = []
-    for tid, stack in sys._current_frames().items():
+    for tid, stack in list(sys._current_frames().items()):
```
amaltaro, line 29: same problem casting to a list.

src/python/Utils/Utilities.py
```diff
@@ -18,7 +20,7 @@ def lowerCmsHeaders(headers):
     code check only cms headers in lower case, e.g. cms-xxx-yyy.
     """
     lheaders = {}
-    for hkey, hval in headers.items(): # perform lower-case
+    for hkey, hval in list(headers.items()): # perform lower-case
```
amaltaro, line 23: same problem casting to a list.

amaltaro (changes requested):
Dario, I have made some comments along the code that will require attention.
We need to see how to deal with the `long` data type as well, given that it's heavily used in MySQL based agents.

Question about the fixer applied to dictionaries. Is that automatically applied by futurize?

src/python/Utils/ProcessStats.py
mapellidario: In this PR I did not change anything by hand. Every change was made automatically by futurize.

Why do you suggest that this can not be done?

The [future doc](http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list) have some examples that look similar to this. Nonetheless, this does not seem to be efficient at all.

As a matter of fact, it seems that future has a way to select the best memory-efficient in both py2 and py3. See
* http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items
* http://python-future.org/what_else.html#dict
* https://stackoverflow.com/a/10190228

src/python/Utils/MathUtils.py
mapellidario: This is strange.

In py3 the int type is like the py2 long type. Futurize let us import the backported py3 int type to py2, so that using int in py2 with futurize import is like using long type in py2 without futurize.

I do not know why this import is not done by default by futirize.

References:
* http://python-future.org/compatible_idioms.html#long-integers
* http://python-future.org/what_else.html#int

mapellidario:
test this please

mapellidario:
test this please

src/python/Utils/CertTools.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import division, print_function
+from builtins import str
```
amaltaro, line 2: It is not clear to me why we need it. Is it to keep the same byte string casting in py2 and py3, thus avoiding unicode strings in python3?

src/python/Utils/ExtendedUnitTestCase.py
```diff
@@ -23,7 +24,7 @@ def assertContentsEqual(self, expected_obj, actual_obj, msg=None):
         """
 
         def traverse_dict(dictionary):
-            for key, value in dictionary.iteritems():
+            for key, value in iteritems(dictionary):
```
amaltaro, line 27: Dario, this documentation:
https://python-future.org/_modules/future/utils.html

seems to suggest to only use `iteritems` if we have python< 2.7. Perhaps we should instead use `viewitems, viewkeys, viewvalues`?

src/python/Utils/MathUtils.py
```diff
@@ -15,11 +17,11 @@ def quantize(inputVal, quanta):
     """
     if isinstance(inputVal, basestring):
         inputVal = float(inputVal)
-    elif not isinstance(inputVal, (int, float, long)):
+    elif not isinstance(inputVal, (int, float)):
         msg = "Input value has to be either int or float, not %s" % (type(inputVal))
         raise ValueError(msg)
 
-    if isinstance(quanta, (basestring, float, long)):
+    if isinstance(quanta, (basestring, int, float)):
```
amaltaro, line 24: no need for `int` here.

src/python/Utils/MathUtils.py
```diff
@@ -15,11 +17,11 @@ def quantize(inputVal, quanta):
     """
     if isinstance(inputVal, basestring):
         inputVal = float(inputVal)
-    elif not isinstance(inputVal, (int, float, long)):
```
amaltaro, line 18: removing `long` here scares me. Perhaps we should override the long data type like:
```
if sys.version_info.major == 3:
  long = int
```
? Just an idea, I don't really know what the best way to approach it.

amaltaro (changes requested):
Dario, sorry for the very belated review. I have a few questions placed in the code, and likely required changes (e.g. itertools is in several places, but I only made one comment).
Please let me know if anything needs clarification.

mapellidario:
I am sorry, described the changes in the commit message, maybe it is better to write the reasoning for the manual changes in the PR comments (i.e. *here* ) instead.


#### PR commit message

Dictionaries: replaced mydict.items() with iteritems(mydict), from future
Resources:
  * http://python-future.org/compatible_idioms.html#dictionaries
  * http://python-future.org/what_else.html#dict
  * https://stackoverflow.com/questions/21783763/in-python-2-what-are-the-performance-trade-offs-for-viewvalues-viewitems-ov
Files
  * Utils/ExtendedUnitTestCase.py
  * Utils/IteratorTools.py
  * Utils/ProcessStats.py
  * Utils/Signals.py
  * Utils/Utilities.py

Int & Long
Resources:
  * http://python-future.org/compatible_idioms.html#long-integers
  * http://python-future.org/what_else.html#int
Files:
  * Utils/MathUtils.py

#### further comments

The approach that futurize has in dealing with ints and longs is why I dropped long in Utils/MathUtils.py

Whether we prefer iteritems or viewitems it is a matter of style. I made a quick research on the topic and found the stackoverflow that I attached to the commit message. The main difference seems that one automatically updates when looping over it, while the other is static. I opted for the one that remains static as a conservative approach, to avoid something like

```
for i in view_of_mylist:
  view_of_mylist.append(something)
```

that could cause not to ever exiting the loop. Maybe this is a silly concern, but I prefer not to loop over dynamic content. As always, the last call is yours, and if we think it is safe for us to loop over the view (which seem to be dynamic), then we can definitely do so.

But in the first link cited in the stackoverflow answer (which is [this](https://stackoverflow.com/a/10190228)) they say that vies may be a bit faster. It is up to you to decide what you prefer.

amaltaro:
Dario, thanks for providing this extra documentation. About iteritems vs viewitems, it's a somehow complicated decision.
It looks like `viewitems` provides the same behaviour between py2 and py3 (returning a dict_items), while `iteritems` returns a `dictionary-itemiterator`, which means you can only consume it once. If I understand it correctly, `iteritems` would provide us with a smaller memory footprint in python2, while `viewitems` would have to load the set-like objects in memory at once.

I think having the same behaviour between py2/py3 is important to have as soon as possible. Given that viewitems seems to perform better (even though a test with an empty dict isn't a valid test!), I'm inclined to think that `viewitems` will be a better choice.

mapellidario:
Perfect, thanks for the review!
I added the `do not merge` label, will rebase this PR, apply the suggested changes and request for the review again

src/python/Utils/CertTools.py
mapellidario: I think futurize automatically added it since it saw that CertTools contained `str()`.

We have to decide what the function should return:

```
[...]
    key = cert = '/tmp/x509up_u' + str(os.getuid())
    if os.path.exists(key):
        return key, cert
```

Right now it is returning two bytes strings. Futurize made the assumption that we want to return two unicode strings and added `from builtins import str`. 

I suggest that even if we want to return bytes, we use `from builtins import bytes` and cast `os.getiud` to bytes, so that the string type will be uniform in python2 and python3.

Some more tests with unicode and bytes strings in py2 are available at https://cms-dmwm-test.docs.cern.ch/py2py3/cheatsheet/strings_future_py2/#unicode-bytes-basestring

amaltaro: What if we change it from:
```
key = cert = '/tmp/x509up_u' + str(os.getuid())
```
to 
```
key = cert = '/tmp/x509up_u{}'.format(os.getuid())
```
?

mapellidario (commented):
I  added some comments (I am sorry for not having published these earlier yesterday)

mapellidario:
test this please

src/python/Utils/EmailAlert.py
```diff
@@ -3,11 +3,12 @@
 """
 
 from __future__ import division
+from builtins import str, object
 import smtplib
 import logging
 
 
-class EmailAlert:
+class EmailAlert(object):
```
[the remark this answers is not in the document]
amaltaro: If dropping python2 support means more than:
* deleting all the `from __future__ import ...`
* deleting all the `from future import ...`
* deleting all the `from builtins import ...`
so, things that are clearly coming from one of those imports. Then yes, I fully support creating a checklist for python2 support removal.

src/python/Utils/IteratorTools.py
```diff
@@ -30,24 +31,24 @@ def nestedDictUpdate(d, u):
...
+    for k, v in viewitems(u):
         if isinstance(v, collections.Mapping):
             r = nestedDictUpdate(d.get(k, {}), v)
             d[k] = r
         else:
             d[k] = u[k]
     return d
 
-def convertFromUnicodeToStr(data):
+def convertFromUnicodeToBytes(data):
```
amaltaro, line 42: Dario, I believe you did, but let me confirm it. Did you grep the whole code for other places calling/importing this function?

amaltaro (approved):
Changes look good to me, Dario. I made a few comments in the code though.
To be on the safe side, I suggest we only merge it once the final CMSWEB tag has been created.

src/python/Utils/IteratorTools.py
mapellidario: Yep, I did grep all the code and did not find usage of this function.

```
> grep -r "convertFromUnicodeToStr" .
Binary file ./src/python/Utils/IteratorTools.pyc matches
./src/python/Utils/IteratorTools.py:def convertFromUnicodeToStr(data):
./src/python/Utils/IteratorTools.py:        return dict(list(map(convertFromUnicodeToStr, data.iteritems())))
./src/python/Utils/IteratorTools.py:        return type(data)(list(map(convertFromUnicodeToStr, data)))
>
```

src/python/Utils/EmailAlert.py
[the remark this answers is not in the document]
mapellidario: Well, having `from builtins import object` in python3 is not going to cause any problem, it is only pedantic and verbose. but I can create such a py2 drop checklist!

src/python/Utils/MathUtils.py
mapellidario: If the plan is to drop python2, having a line that screams `past` was not so appealing :)

The "price" that we have to pay is to get accustomed to the new `int` type that is going to be "long" also in py2, which does not seem terrible. So I took the liberty to choose option 1. I am glad that you agree :)

mapellidario:
I squashed the commits into a single commit. All the changes here involved "real" code, no specific "test" code in this PR

amaltaro:
Sorry Dario, I overlooked the "Pre-python 2.6 idioms..." from the report file, can you please get that fixed too (and squashed)?

mapellidario:
Ok, done! Sorry for not having changed it before. We did not have a discussion about what style we should follow and I changed only what has been deemed necessary. 

The change suggested by the report seems to come from `2to3`, which likes to update some older idioms to more recent alternatives, despite the oldest alternative being still valid. https://docs.python.org/3/library/2to3.html#2to3fixer-idioms

On a same note, would you like me to go thru some of the errors and warning from pylint?

amaltaro:
Thanks Dario. About the issue I reported, I think this is the check done in jenkins:
https://github.com/dmwm/WMCore/wiki/Understanding-Jenkins#python-future-code-quality-checks

Regarding refactoring, we can likely leave those to a future opportunity and focus mainly on the python3 support (unless we spot something that might be too fragile).

PR #9735, opened 2020-06-11
[PY3] Make StringIO and httplib compatible with python3

amaltaro:
Fixes #
(no issue created yet)

#### Status
In development

#### Description
Make some imports compatible between python2 and python3, such as:
* httplib --> http.client
* StringIO --> io.StringIO

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -6,8 +6,11 @@ Record a set of tweakable parameters from a CMSSW Configuration in a CMSSW
 independent python structure
 
 """
+try:
+    from StringIO import StringIO ## for Python 2
+except ImportError:
+    from io import StringIO ## for Python 3
 
-import StringIO
 import imp
 import inspect
 import json
@@ -394,7 +397,7 @@ class PSetTweak:
 
 
             jsoniser = JSONiser()
-            jsoniser.dejson(json.load(StringIO.StringIO(jsonContent)))
+            jsoniser.dejson(json.load(StringIO(jsonContent)))
 
             for param, value in jsoniser.parameters.items():
                 self.addParameter(param , value)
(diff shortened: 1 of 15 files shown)


amaltaro:
@mapellidario Dario, given that the StringIO part has already been taken care of; and you have other PRs open for PSetTweaks and soon WMCore.Cache. Do you think you could create a new PR taking over a few other changes proposed in this PR?

I'm specifically talking about the http module:
https://python-future.org/compatible_idioms.html#http-module

mapellidario:
Python futurize deals with `httplib` -> `http.client` with the stage2 fixer `futurize -f libfuturize.fixes.fix_future_standard_library $path` . 

Dealing with all standard library in just one fixer, it would also change 

* `cPickle` to `pickle`
* `urlparse` to `urllib.parse` 
* `thread` to `_thread`
* `from commands import getoutput` to `from subprocess import getoutput`

The files affected are 37, mainly in WMCore, but also a handful in WMComponent.

Do we still want to update httplib only? It would require some manual work, that would have to be replicated for cPickle, urlparse, thread and subprocess, while we could just rely as much as we can on the automated fixers changing and testing a single module at a time.

amaltaro:
Dario, as we discussed over slack, please proceed with the automated fixer, but revert the cPickle changes for now (until we can properly profile/evaluate what is the performance penalty in migrating to the non C implementation of pickle).

mapellidario:
Dear Alan, in the last weeks we opened the issue #9761 and its relative PR #9762 dealing with the migration of python std library as a whole, and not only `httplib` to `http.client`. Moreover, migrating the use of strings may require some more extra attention and may be dealt with a gradual approach.

What would you think about closing this PR without merging?

amaltaro:
I agree Dario. Thanks to your and David's work. Closing

PR #9740, opened 2020-06-12
[py2py3] futurize on WMCore/src/python/PSetTweaks

### LINKED ISSUE #9739: [py2py3] PSetTweaks should be compatible with both py2 and py3
**Impact of the new feature**

WMCore/src/python/PSetTweaks

Make PSetTweaks compatible with both python2 and python3

Run futurize on all the modules of the Utils package and check if any manual change is required to pass the unit tests.

mapellidario:
Fixes #9739 

#### Status
ready

#### Description
Run futurize on the modules in the PSetTweaks package

The log of the commands and tests run is available at https://cms-dmwm-test.docs.cern.ch/py2py3/porting_log/001_9740_psettweaks/

yes

#### Related PRs
none

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -7,14 +7,19 @@ independent python structure
 
 """
 
+from future import standard_library
+standard_library.install_aliases()
+
+from builtins import object, map, range
+from past.builtins import basestring
+from future.utils import viewitems, viewvalues
+
 import imp
 import inspect
 import json
 import pickle
 import sys
 from functools import reduce
-from io import BytesIO
-
 
 class PSetHolder(object):
     """
@@ -31,7 +36,7 @@ class PSetHolder(object):
 #  //
 # // Assistant lambda functions
 #//
-childPSets = lambda x: [ value for value in x.__dict__.values()
+childPSets = lambda x: [ value for value in viewvalues(x.__dict__)
                          if value.__class__.__name__ == "PSetHolder" ]
 childParameters = lambda p, x: [ "%s.%s" % (p,i) for i in  x.parameters_ ]
 
@@ -61,7 +66,7 @@ def psetIterator(obj):
 
 
 
-class PSetLister:
+class PSetLister(object):
     """
     _PSetLister_
 
@@ -92,7 +97,7 @@ class PSetLister:
         self.queue.pop(-1)
 
 
-class JSONiser:
+class JSONiser(object):
     """
     _JSONiser_
 
@@ -139,7 +144,7 @@ class JSONiser:
         queue = ".".join(self.queue)
         for param in params:
             self.parameters["%s.%s" % (queue, param)]  = dictionary[param]
-        for key, value in dictionary.items():
+        for key, value in viewitems(dictionary):
             if isinstance(value, dict):
                 self.queue.append(key)
                 self.dejson(dictionary[key])
@@ -152,7 +157,7 @@ class JSONiser:
 
 
 
-class PSetTweak:
+class PSetTweak(object):
     """
     _PSetTweak_
 
@@ -394,9 +399,9 @@ class PSetTweak:
 
 
             jsoniser = JSONiser()
-            jsoniser.dejson(json.load(BytesIO(jsonContent)))
+            jsoniser.dejson(json.loads(jsonContent))
 
-            for param, value in jsoniser.parameters.items():
+            for param, value in viewitems(jsoniser.parameters):
                 self.addParameter(param , value)
 
 
@@ -411,6 +416,6 @@ def makeTweakFromJSON(jsonDictionary):
     jsoniser = JSONiser()
     jsoniser.dejson(jsonDictionary)
     tweak = PSetTweak()
-    for param, value in jsoniser.parameters.items():
+    for param, value in viewitems(jsoniser.parameters):
         tweak.addParameter(param , value)
     return tweak
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -7,7 +7,10 @@ Note: This can be used within the CMSSW environment to act on a
 process/config but does not depend on any CMSSW libraries. It needs to stay like this.
 
 """
-from __future__ import print_function
+from __future__ import print_function, division
+
+from builtins import map, range, str, object
+from future.utils import viewitems, viewkeys
 
 import logging
 import os
@@ -119,7 +122,7 @@ def lfnGroup(job):
     default both to 0. The result will be a 5-digit string.
     """
     modifier = str(job.get("agentNumber", 0))
-    jobLfnGroup = modifier + str(job.get("counter", 0) / 1000).zfill(4)
+    jobLfnGroup = modifier + str(job.get("counter", 0) // 1000).zfill(4)
     return jobLfnGroup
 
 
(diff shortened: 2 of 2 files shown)


src/python/PSetTweaks/PSetTweak.py
```diff
@@ -394,9 +400,9 @@ def unpersist(self, filename, formatting=None):
 
 
             jsoniser = JSONiser()
-            jsoniser.dejson(json.load(StringIO.StringIO(jsonContent)))
+            jsoniser.dejson(json.load(io.StringIO(jsonContent.decode("utf-8"))))
```
amaltaro, line 403: I dealt with this line differently in:
https://github.com/dmwm/WMCore/pull/8733/files#diff-b600d610cf56591819a4df56f024cb45R397

and honestly, I'm not sure whether we should deal with that as a byte streaming, or decoding a text object as a unicode string.

mapellidario: Uh, now I see. You are totally right on this, my solution is a waste of resources. 
What should I do? Add a commit that is the same as yours so that the PR, in case they will be both merged, will not collide? Should I open the PR without any change to this line?

amaltaro: I'm actually not sure I'm right ;)
You deal with this following the sandwich approach, which sooner or later will need to be implemented (first, we need to define which encoding we use in the repo such that we do not need to start guessing how to encode/decode things).

src/python/PSetTweaks/PSetTweak.py
```diff
@@ -7,7 +7,13 @@
 
 """
 
-import StringIO
+from future import standard_library
+standard_library.install_aliases()
+from builtins import map
```
amaltaro, line 12: I suggest we group these imports

src/python/PSetTweaks/PSetTweak.py
```diff
@@ -139,7 +145,7 @@ def dejson(self, dictionary):
         queue = ".".join(self.queue)
         for param in params:
             self.parameters["%s.%s" % (queue, param)]  = dictionary[param]
-        for key, value in dictionary.items():
+        for key, value in list(dictionary.items()):
```
amaltaro, line 148: why cast it to a list if it's actually returning both keys and values?

src/python/PSetTweaks/PSetTweak.py
amaltaro: For now, I'd suggest you to rebase your PR and simply remove this fix.

src/python/PSetTweaks/PSetTweak.py
```diff
@@ -394,9 +400,9 @@ def unpersist(self, filename, formatting=None):
 
 
             jsoniser = JSONiser()
-            jsoniser.dejson(json.load(StringIO.StringIO(jsonContent)))
+            jsoniser.dejson(json.load(io.StringIO(jsonContent.decode("utf-8"))))
 
-            for param, value in jsoniser.parameters.items():
+            for param, value in list(jsoniser.parameters.items()):
```
amaltaro, line 405: Same here (casting a tuple to list). And a couple other places in this PR.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -8,7 +8,13 @@
 
 """
 from __future__ import print_function
+from __future__ import division
 
+from builtins import map
+from builtins import str
```
amaltaro, line 14: I'd suggest to group all these imports too.

src/python/PSetTweaks/WMTweak.py
```diff
@@ -119,7 +125,7 @@ def lfnGroup(job):
     default both to 0. The result will be a 5-digit string.
     """
     modifier = str(job.get("agentNumber", 0))
-    jobLfnGroup = modifier + str(job.get("counter", 0) / 1000).zfill(4)
+    jobLfnGroup = modifier + str(old_div(job.get("counter", 0), 1000)).zfill(4)
```
amaltaro, line 128: I'd actually use the division from futurize (this place is expecting an integer division `//`)

amaltaro (changes requested):
@mapellidario Dario, besides rebasing your PR with the changes that went in in #8733, and the comments I made along the code. Here are a few generic comments/questions:
* do we need to import `object from builtins` for every class directly inheriting from it? If so, then we might want to run a single PR updating every single file in the repository.
* when iterating over a dict, we only cast to `list` if we ask for `keys()` or `values()`, right?
* rule of thumb, let's try to group imports as much as we can.

src/python/PSetTweaks/PSetTweak.py
mapellidario: In this particular case, we are lucky: `json.loads`, which deserializes a JSON document into a python dictionary parsing, can have as input either a bytes string or a unicode string in both py2 and py3.

This means that we can just use
```
jsoniser.dejson(json.loads(jsonContent))
```

References
* https://docs.python.org/2.7/library/json.html#json.loads
* https://docs.python.org/3/library/json.html#json.loads

If we were not so lucky (if we had to deal with a library not as smart as `json`), we could have followed the suggestions from [future cheatsheet about py2 and py3 compatible file opening syntaxes](http://python-future.org/compatible_idioms.html#file-io-with-open), so that we could have been sure that `jsonContent` was a handle to bytes strings and that it could have been a proper input for `io.BytesIO()` in both py2 and py3.

mapellidario:
> do we need to import `object from builtins` for every class directly inheriting from it? If so, then we might want to run a single PR updating every single file in the repository.

This seems like a reasonable suggestion. I added it to the list of PRs that I should open.
It should be the `libpasteurize.fixes.fix_newstyle` fix, described at https://github.com/PythonCharmers/python-future/blob/master/src/libpasteurize/fixes/fix_newstyle.py

> when iterating over a dict, we only cast to `list` if we ask for `keys()` or `values()`, right?

I guess this is some strange futurize choice to have consistent behaviour in py2 and py3. As we already discussed in #9731 , we could use the `view*` operators from `builtins` to iterate efficiently over dictionaries in both py2 and py3.

> rule of thumb, let's try to group imports as much as we can.

Roger

src/python/PSetTweaks/WMTweak.py
mapellidario: Do you mean that we could just use the `//` operator, as suggested by [futurize cheatsheet](https://python-future.org/compatible_idioms.html#division)

src/python/PSetTweaks/PSetTweak.py
amaltaro: Perfect!

src/python/PSetTweaks/PSetTweak.py
```diff
@@ -31,7 +37,7 @@ def __init__(self, psetName):
 #  //
 # // Assistant lambda functions
 #//
-childPSets = lambda x: [ value for value in x.__dict__.values()
+childPSets = lambda x: [ value for value in list(x.__dict__.values())
```
amaltaro, line 40: shall we try to use `viewvalues` instead?

amaltaro (approved):
Dario, this is looking pretty good to me. I made one comment in the code though, if you feel like it's worth updating that, then please go ahead.

src/python/PSetTweaks/PSetTweak.py
mapellidario: Oops, you are definitely correct, I must have missed that. thank you very much!

amaltaro (approved):
It looks good to me. Just need to get those commits squashed.

PR #9755, opened 2020-06-18
Implement a check for 'excludeDataTier' && Skip transfer submission for workflows with empty outputDataset list or with all data tiers excluded.

### LINKED ISSUE #9612: MS: support a configurable list of data-tiers to be skipped from the output data placement
**Impact of the new feature**
ReqMgr2MS

as we heard over the last weeks, NANO data tiers will be the first one to be managed by Rucio. And we have decided not to make any subscriptions for those tiers once they start getting injected into Rucio.

Create a service configuration parameter (list data type) with a list of data tiers that the microservice should skip from the output data placement. Default value has to be:
`['NANOAOD', 'NANOAODSIM']`

but evaluated to `[]` if Rucio is still not enabled in the Microservice.

Having this configurable via a REST API might be preferred. But then we need to implement access control...

**Additional context**
none

### LINKED ISSUE #9727: MSOutputConsumer fails with KeyError for DQMHarvest workflows
**Impact of the bug**
ReqMgr2MS (MSOutputConsumer)

**Describe the bug**
DQMHarvest workflows do not have any output datasets, which makes MSOutput to fail when preparing the output data placement.

**How to reproduce it**
Run a DQMHarvest workflow until it completes

**Expected behavior**
No output data placement should be performed, and the mongodb record needs to be marked as `transferStatus="done"` (or whatever final status we decided to use here), such that it's not picked by this thread in the next cycle. 

Fresh traceback:
```
2020-06-06 05:05:31,310:ERROR:MSOutput: Could not create DDMReq for Workflow: amaltaro_DQMHarvesting_RunWhitelist_Agent112_Validation_170413_130804_3940Error: "ERROR: Missing Mandatory Fields: ['item'] for: {'item': [], 'cache': None, 'group': 'DataOps', 'site': ['T2_*', 'T1_*_Disk'], 'n': None}"
  File "/data/srv/beHG2006h/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.5.pre5/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSOutput.py", line 317, in makeSubscriptions
    site=workflow['destination'])
  File "/data/srv/beHG2006h/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.5.pre5/lib/python2.7/site-packages/WMCore/Services/DDM/DDM.py", line 106, in __init__
    raise KeyError(msg)
KeyError: "ERROR: Missing Mandatory Fields: ['item'] for: {'item': [], 'cache': None, 'group': 'DataOps', 'site': ['T2_*', 'T1_*_Disk'], 'n': None}"
2020-06-06 05:05:31,311:INFO:MSOutput: MSOutputConsumer:140182742312704@vocms0731.cern.ch: PipelineNonRelVal: Processed 'msOutDoc' with '_id': amaltaro_DQMHarvesting_RunWhitelist_Agent112_Validation_170413_130804_3940.
2020-06-06 05:05:31,311:DEBUG:MSOutput: {u'Campaign': u'Agent112_Validation',
 u'OutputDatasets': [],
 u'RequestName': u'amaltaro_DQMHarvesting_RunWhitelist_Agent112_Validation_170413_130804_3940',
 u'_id': u'amaltaro_DQMHarvesting_RunWhitelist_Agent112_Validation_170413_130804_3940',
 u'campaignOutputMap': [{u'campaignName': u'Agent112_Validation',
                         u'datasets': []}],
 u'creationTime': 1591290411,
 u'destination': None,
 u'destinationOutputMap': None,
 u'isRelVal': False,
 u'isTaken': False,
 u'isTakenBy': None,
 u'lastUpdate': 1591412731,
 u'numberOfCopies': None,
 u'transferIDs': None,
 u'transferStatus': None}
2020-06-06 05:05:31,312:INFO:MSOutput:
```

todor-ivanov:
Fixes #9612 
Fixes #9727 

#### Status
ready

#### Description
We need to have a way for excluding some data Tiers from being managed by the Microservices until we fully switch to Rucio. So we need a configurable flag to exclude them from the data placement logic.

In order to achieve that I had to consolidate the  logic how we do the actual transfer subscription in the Consumer Thread, so that we always make the subscriptions based on the destinationOutputMap for all types of workflows, ergo I had to build it for both RelVal && NonRelval workflows. 

Having the aforementioned changes logically allows  (and basically requires) to fix the broken submissions for workflows with empty `OutputDtaset` lists.

The list of changes this PR introduces is as follows:
   * Implement a check for 'excludeDataTier' 
   * Always create destinationOutputMap 
   * Always use destinationOutput map for transfer subscriptions 
   * Skip transfer submission for workflows with empty outputDataset list or with all data tiers excluded.
 
YES

#### Related PRs
NO

NO

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -82,7 +82,7 @@ class MSOutput(MSCore):
         self.msConfig.setdefault("defaultGroup", "DataOps")
         self.msConfig.setdefault("enableAggSubscr", True)
         self.msConfig.setdefault("enableDataPlacement", False)
-        self.msConfig.setdefault("excludeDataTier", ['NANOAOD'])
+        self.msConfig.setdefault("excludeDataTier", ['NANOAOD', 'NANOAODSIM'])
         self.msConfig.setdefault("rucioAccount", 'wma_test')
         self.msConfig.setdefault("mongoDBUrl", 'mongodb://localhost')
         self.msConfig.setdefault("mongoDBPort", 8230)
@@ -288,34 +288,30 @@ class MSOutput(MSCore):
                 ddmReqList = []
                 try:
                     if workflow['isRelVal']:
-                        for dMap in workflow['destinationOutputMap']:
-                            try:
-                                ddmRequest = DDMReqTemplate('copy',
-                                                            item=dMap['datasets'],
-                                                            n=workflow['numberOfCopies'],
-                                                            site=dMap['destination'],
-                                                            group='RelVal')
-                            except KeyError as ex:
-                                # NOTE:
-                                #    If we get to here it is most probably because the 'site'
-                                #    mandatory field to the DDM request is missing (due to an
-                                #    'ALCARECO' dataset or similar). Since this is expected
-                                #    to happen a lot, we'd better just log a warning and continue
-                                msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
-                                msg += "Error: {}".format(ex)
-                                self.logger.warning(msg)
-                                continue
-                            ddmReqList.append(ddmRequest)
+                        group = 'RelVal'
                     else:
-                        # FIXME:
-                        #    We need to create the campaignMap and use it for
-                        #    creating the requests for the nonRelVal workflows
-                        #    it should be also the case when we migrate to Rucio
-                        ddmRequest = DDMReqTemplate('copy',
-                                                    item=workflow['OutputDatasets'],
-                                                    n=workflow['numberOfCopies'],
-                                                    site=workflow['destination'])
+                        group = 'DataOps'
+
+                    for dMap in workflow['destinationOutputMap']:
+                        try:
+                            ddmRequest = DDMReqTemplate('copy',
+                                                        item=dMap['datasets'],
+                                                        n=workflow['numberOfCopies'],
+                                                        site=dMap['destination'],
+                                                        group=group)
+                        except KeyError as ex:
+                            # NOTE:
+                            #    If we get to here it is most probably because the 'site'
+                            #    mandatory field to the DDM request is missing (due to an
+                            #    'ALCARECO' dataset from a Relval workflow or similar).
+                            #    Since this is expected to happen a lot, we'd better just
+                            #    log a warning and continue
+                            msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
+                            msg += "Error: {}".format(ex)
+                            self.logger.warning(msg)
+                            continue
                         ddmReqList.append(ddmRequest)
+
                 except Exception as ex:
                     msg = "Could not create DDMReq for Workflow: {}".format(workflow['RequestName'])
                     msg += "Error: {}".format(ex)
@@ -324,9 +320,20 @@ class MSOutput(MSCore):
 
                 try:
                     # In the message bellow we may want to put the list of datasets too
-                    msg = "Making transfer subscriptions for {}".format(workflow['RequestName'])
-                    self.logger.info(msg)
-                    ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
+                    msg = "Making transfer subscriptions for %s"
+                    self.logger.info(msg, workflow['RequestName'])
+
+                    if ddmReqList:
+                        ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
+                    else:
+                        # NOTE:
+                        #    Nothing else to be done here. We mark the document as
+                        #    done so we do not iterate through it multiple times
+                        msg = "Skip submissions for %s. Either all data Tiers were "
+                        msg += "excluded or there were no Output Datasets at all."
+                        self.logger.warning(msg, workflow['RequestName'])
+                        self.docKeyUpdate(workflow, transferStatus='done')
+                        return workflow
                 except Exception as ex:
                     msg = "Could not make transfer subscription for Workflow: {}".format(workflow['RequestName'])
                     msg += "Error: {}".format(ex)
@@ -504,6 +511,10 @@ class MSOutput(MSCore):
                         {'$or': [
                             {'lastUpdate': None},
                             {'lastUpdate': {'$lt': treshTime}}]}]}
+
+                # FIXME:
+                #    To redefine those exceptions as MSoutputExceptions and
+                #    start using those here so we do not mix with general errors
                 try:
                     pipeLine.run(mQueryDict)
                 except KeyError as ex:
@@ -658,13 +669,29 @@ class MSOutput(MSCore):
         """
 
         # Fill the destinationOutputMap first
-        if msOutDoc['isRelVal']:
-            destinationOutputMap = []
-            wflowDstSet = set()
-            updateDict = {}
-            for dataset in msOutDoc['OutputDatasets']:
-                _, dsn, procString, dataTier = dataset.split('/')
-                destination = set()
+        destinationOutputMap = []
+        wflowDstSet = set()
+        updateDict = {}
+        for dataset in msOutDoc['OutputDatasets']:
+            _, dsn, procString, dataTier = dataset.split('/')
+            # NOTE:
+            #    Data tiers that have been configured to be excluded will never
+            #    enter the destinationOutputMap
+            if dataTier in self.msConfig['excludeDataTier']:
+                # msg = "%s: %s: "
+                # msg += "Data Tier: %s is blacklisted. "
+                # msg += "Skipping dataset placement for: %s:%s"
+                # self.logger.info(msg,
+                #                  self.currThreadIdent,
+                #                  pipeLine,
+                #                  dataTier,
+                #                  msOutDoc['RequestName'],
+                #                  dataset)
+                continue
+
+            destination = set()
+
+            if msOutDoc['isRelVal']:
                 if dataTier != "RECO" and dataTier != "ALCARECO":
                     destination.add('T2_CH_CERN')
                 if dataTier == "GEN-SIM":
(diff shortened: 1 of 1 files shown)


src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -324,9 +320,30 @@ def makeSubscriptions(self, workflow):
...
+                    self.logger.info(msg, workflow['RequestName'])
+
+                    # NOTE:
+                    #    If ddmReqList is empty it means that no destinationOutputMap
+                    #    has been created for this workflow, either because of
+                    #    all output dataTiers for this workflow have been excluded
+                    #    or because the workflow had no output datasets at all
+                    #    (eg. a Harvesting workflow)
+                    #
+                    # NOTE:
```
amaltaro, line 333: I think these 2 notes could be dropped. Maybe just write inside of the `else` block that we mark the document as `done` because there is nothing else to be done, just move on. The warning message already explains the rest.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -658,13 +679,29 @@ def docInfoUpdate(self, msOutDoc, pipeLine=None):
...
-            wflowDstSet = set()
-            updateDict = {}
-            for dataset in msOutDoc['OutputDatasets']:
-                _, dsn, procString, dataTier = dataset.split('/')
-                destination = set()
+        destinationOutputMap = []
+        wflowDstSet = set()
+        updateDict = {}
+        for dataset in msOutDoc['OutputDatasets']:
+            _, dsn, procString, dataTier = dataset.split('/')
```
amaltaro, line 686:
```suggestion
            dataTier = dataset.split('/')[-1]
```

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -658,13 +679,29 @@ def docInfoUpdate(self, msOutDoc, pipeLine=None):
...
+            # NOTE:
+            #    Data tiers that have been configured to be excluded will never
+            #    enter the destinationOutputMap
+            if dataTier in self.msConfig['excludeDataTier']:
+                # msg = "%s: %s: "
```
amaltaro, line 691: I suggest to log the dataset name getting excluded and mention it's a data tier blacklisted.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -677,57 +714,69 @@ def docInfoUpdate(self, msOutDoc, pipeLine=None):
...
+        aggDstMap = []
+
+        # populate the first element in the aggregated list
+        if len(destinationOutputMap) != 0:
+            aggDstMap.append(destinationOutputMap.pop())
 
-            # here we try to aggregate the destination map per destination
-            aggDstMap = []
+        # feed the rest
+        while len(destinationOutputMap) != 0:
```
amaltaro, line 741: I think this can be addressed in a much cleaner way with dictionary. The key will take care of the dataset aggregation by site(s).

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -677,57 +714,69 @@ def docInfoUpdate(self, msOutDoc, pipeLine=None):
...
-                                      self.currThreadIdent,
-                                      pipeLine,
-                                      msOutDoc['_id'],
-                                      str(ex))
+        wflowDstList = list(wflowDstSet)
+        updateDict['destination'] = wflowDstList
+        updateDict['destinationOutputMap'] = destinationOutputMap
+        try:
+            msOutDoc.updateDoc(updateDict, throw=True)
+            # msg = "%s: %s: 'msOutDoc': %s"
```
amaltaro, line 766: These comments can likely go away.

amaltaro (changes requested):
These changes are looking good. I have made a few comments suggesting some minor changes though.

I also think we should move the workflow dataset/location logic under the output template module:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py

It's cleaner and easier to maintain the workflow logic at the same place.

amaltaro:
Todor, please create some unit tests as well to validate the workflow handling. I'm mostly concerned about the logic for all those datasets/campaign/location mapping, but feel free to extend it as you like. The more unit tests the better ;)

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I was afraid of having thousands of those in a single run. But OK, I will log it.

amaltaro: Ohh, if it is on the Producer thread, then yes, we will have hundreds of messages to the same workflow. With that in mind, perhaps we should then build the MSOutput document without any restrictions, and leave the possible "datatier filter" to be handled by the consumer thread.

todor-ivanov: Yes, we can expose that thing into the Consumer though, but then we will deviate from the original idea - to have as little conditions as possible there, but to have everything predefined in the document instead. This was the concept I was relying on when I was consolidating the logic for both `RelVal`s and `NonRelval`s. Another reason why I put this in the Producer was that the Consumer does not iterate thorough the output datasets any more. It deals with the destinationOutputMap directly and builds the transfer requests based on the already aggregated dataset lists there.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: done

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: done

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: done

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I am not sure about that. The campaign map is also a list, having them in the same format is good. Also changing the format here will require a change of the database document format and a whole new testing and validation of how we build the map and how we use it etc. etc.. Not that it is quite a big test but will require time we do not have it right now.

todor-ivanov:
Thanks @amaltaro I worked on the comments you posted inline and comited some changes already. I am creating few Unit tests right now.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Actually this immediately introduced a BUG, because `dsn` and `procString` are used few lines bellow :). Here is the pylint report: [1] I am reverting this one.

[1]
```
E0602, line 703 in MSOutput.docInfoUpdate: Undefined variable 'dsn'
E0602, line 703 in MSOutput.docInfoUpdate: Undefined variable 'procString'
E0602, line 705 in MSOutput.docInfoUpdate: Undefined variable 'dsn'
E0602, line 705 in MSOutput.docInfoUpdate: Undefined variable 'procString' 
```

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -288,34 +288,30 @@ def makeSubscriptions(self, workflow):
...
+                    for dMap in workflow['destinationOutputMap']:
+                        try:
+                            ddmRequest = DDMReqTemplate('copy',
+                                                        item=dMap['datasets'],
+                                                        n=workflow['numberOfCopies'],
+                                                        site=dMap['destination'],
+                                                        group=group)
+                        except KeyError as ex:
+                            # NOTE:
+                            #    If we get to here it is most probably because the 'site'
```
amaltaro, line 304: Wait, don't we make a DDM request without any destination for standard workflows? Which means, `site` will only have a non-None value for RelVal samples (that match the data tier)

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -324,9 +320,20 @@ def makeSubscriptions(self, workflow):
...
                 try:
                     # In the message bellow we may want to put the list of datasets too
-                    msg = "Making transfer subscriptions for {}".format(workflow['RequestName'])
-                    self.logger.info(msg)
-                    ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
+                    msg = "Making transfer subscriptions for %s"
+                    self.logger.info(msg, workflow['RequestName'])
+
+                    if ddmReqList:
+                        ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
```
amaltaro, line 327: I'm pretty sure this question has already been asked in your previous PR, but I don't remember the answer.
What happens if one of the DDM requests fail?

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: You're right, sorry!

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: It's not an easy decision. There are pros and cons on both approaches. I think we can assume that these list of tiers to be skipped won't be around for more than 2 or 3 months, so it's something that we will soon be able to get rid of in the code.
I'd suggest to make a DEBUG level record them, just in case.

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: Dict would be a temporary data structure. Just so you can aggregate datasets by site. In the end you would then need to convert that dict to a list of dicts, as defined in the document schema.

However, this destination output map really does not belong to MSOutput and it should be moved under the msOutTemplate module. That class should contain all the logic for a workflow <--> msOutDoc (including the RelVal logic. Can you please move it there and get unit tests created?

amaltaro (changes requested):
Besides unit tests, I still think we need another iteration on this PR.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: No, For standard workflows, the fact that we do or do not define the `site` here or in the the producer thread, does not change anything - the Wildcard list of `['T1_*', 'T2_*']` is set by default from the DDM module anyway [1]. But when it comes to `RelVal`s we can get into the trouble of a workflow with a zeroed site list because of: [2]. This check does not come with this PR I was experiencing that error back before I start working on unifying the submission logic for Relvals and NonRelvals,  and that's why it is put there. 


[1]
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/Services/DDM/DDM.py#L42

[2]
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/MSOutput.py#L668
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/MicroService/Unified/MSOutput.py#L676
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/MicroService/Unified/MSOutput.py#L678

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: It is an atomic call to the the external function `makeAggrequests()` and the aggregation is happening inside that function. If this call fails an exception would be raised. If it succeeds then it will return a list of the statuses of all the transfer requests that have been made - one entry in the result list per aggregated submission [1], if any of them fails the relevant entry in the list will be `None` [2].

Then all the statuses are converted accumulated in a status dictionary [3] and later converted to to `True` `False` values. At the end  if any of the there was even one transfer that have been failed then the Workflow will not be marked as done and will be retried on the next run [4].

[1]
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/Services/DDM/DDM.py#L327 

[2]
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/Services/DDM/DDM.py#L334

[3]
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/MicroService/Unified/MSOutput.py#L339-L345

[4]
https://github.com/dmwm/WMCore/blob/6c7032b85841a535247026ba70b48686fccd2838/src/python/WMCore/MicroService/Unified/MSOutput.py#L347-L352

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: Okay, I see you have also updated the wiki and that this is actually the default in Dynamo for cases that a destination site does not get provided by the client. We might have to touch this default in the future, when going to Rucio, but we can likely live it like that for now.

> But when it comes to RelVals we can get into the trouble of a workflow with a zeroed site list because of: [2]

I'm not sure I understood it. Are you saying that same Relval output datasets do not have any location, because it's by design that we do not transfer some of them? What is the problem with that?

todor-ivanov:
> However, this destination output map really does not belong to MSOutput and it should be moved under the msOutTemplate module

This is something that I am still not convinced in, though. I am trying to keep the MSOutputTemplate class as small ass possible. Any additional method that we add there is an overhead in memory consumption. When it comes to several hundred documents won't be an issue, but when we start dealing with several thousands just for a single call to Reqmgr, then it could become noticeable. I am trying to see your point here btw, and generally speaking I would say it is not without a sense what you are saying here. Maybe we should through more stuff in that class.

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: Cool. I'm resolving it then

amaltaro:
That's a valid concern. IMO, the code would be better structured with all the workflow logic in there, but if the memory overhead is substantial, then I fully support to have it in MSOutput. 

I remember you had some memory profiling in your first PRs for MSOutput, if that is something that would take you less than 30min or so, then I'd be interested in seeing the memory effect. Otherwise we leave it to another occasion and move on.

amaltaro:
Todor, Imran is going to run a testbed deployment this morning. So I'll add this PR in.
We still have work left for MSOutput, so whatever needs to be addressed here can be taken care in your next PR ;) Thanks

todor-ivanov:
Thanks Alan!

PR #9759, opened 2020-06-19
MSTransferor: support input data placement with Rucio

### LINKED ISSUE #9461: Create Rucio APIs for MicroServices with multi pycurl
**Impact of the new feature**
MicroServices

The MS Transferor implementation is not yet fully supporting Rucio. One of the reasons is because we still have to find out what are the exact APIs that best suit our needs.

Implement the Common wrapper APIs for Rucio requests, using the concurrent pycurl `getdata`
function such that we can speed up any REST API requests.
If there are no Rucio APIs providing the information that is needed, e.g.: given a dataset, return its size, we might have to fallback to DBS (likely meaning a slower response too).
 
We could use the Services/Rucio module, but then we cannot take advantage of the concurrent requests. In the future we can reorganize those functions under the `Common.py` module and make it at least more clear and reuse as much code as we can.

**Additional context**
This issue is meant to address what didn't get implemented in this PR: https://github.com/dmwm/WMCore/pull/9457

### LINKED ISSUE #9725: Transferor MS should be able to get data location from Rucio
We have this issue https://its.cern.ch/jira/browse/CMSRUCIO-119 which I think makes sense for the next round of transitions, which is that all data location goes through Rucio. Are there other parts of the microservices which should do the same? Alternatively are there parts of Unified which are not in MicroServices which would need to be modified?
 

@todor-ivanov ? @amaltaro ?

amaltaro:
Fixes #9725 
Fixes #9461

#### Status
In development

#### Description
Summary of changes are:
* implemented a Rucio wrapper based on pycurl for concurrent requests; only a few very needed APIs implemented so far;
* when listing Rucio replication rules, it could be that rules in state STUCK and SUSPENDED will not be considered and a new rule can be created;
* updated MSCore `__init__` method to support sub-classes to provide service names that they do not want to be initialized;
* updated MSCore to initialize a Rucio object;
* support RSE quotas update based on Rucio (using quota limit and usage)
* added all the necessary logic in MSTransferor to deal with either PhEDEx-based or Rucio-based input data placement;
  * including check on the RSE quota and usage
  * relying on a new Rucio account: `wmcore_transferor`
  * logic for the pileup is: if the container is not locked anywhere within the SiteWhitelist (with that Rucio acct), then we make a new rule with grouping=ALL against one specific RSE
  * logic for the primary is: find all the blocks and make a rule for all of them against all RSEs to be used (grouping=DATASET)
  * logic for the primary + parent is: create chunks of primary + parent and make a rule against one specific RSE (other chunks go to other RSEs), using grouping=DATASET
* Rucio rules are created in asynchronous mode, `ask_approval` always False;
* Rucio rules - in non production server/mode - will contain a lifetime of 24h;
* Rucio RSE quota and rucio rules are created with the python client APIs; container size, list of blocks and their sizes and current rules locking data is done via RESTful APIs;
* renew the Rucio token if it's to expiry within 30min;
* Todor/Kenyi/I are notified on large data placement;
* on the Rucio wrapper, support `lifetime` and `asynchronous` parameters for replication rule creation;

In addition to that, there is:
* fixed a bug in MSMonitor when counting workflows that had to be skipped;
* fixed a bug, inconsistent progress between Rucio and PhEDEx transfers. Now their progress is expressed in percentage, thus between 0 and 100;
* added a debug line to print some extra information about rules being evaluated

It depends on the following new configuration parameters:
* `useRucio`: to enable/disable Rucio as a DM service (if disabled, PhEDEx is used)
* `rulesLifetime`: defines an expiration time for all rules created by MSTransferor. Production rules (against the production Rucio server) have no expiration time.

yes

#### Related PRs
none

Deployment changes: 
https://github.com/dmwm/deployment/pull/924
https://github.com/dmwm/deployment/pull/942

### DIFF
--- a/src/python/WMCore/MicroService/Tools/PycurlRucio.py
+++ b/src/python/WMCore/MicroService/Tools/PycurlRucio.py
@@ -1,0 +1,403 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Module based on the pycurl_manager module, implementing a few
+functions to call to the Rucio RESTful APIs, leveraging the
+pycurl concurrency.
+"""
+from __future__ import print_function, division, absolute_import
+
+from future import standard_library
+standard_library.install_aliases()
+
+import datetime
+import json
+import logging
+import re
+
+from urllib.parse import quote, unquote
+
+from Utils.CertTools import getKeyCertFromEnv
+from WMCore.Services.pycurl_manager import RequestHandler
+from WMCore.Services.pycurl_manager import getdata as multi_getdata
+
+### Amount of days that we wait for stuck rules to be sorted
+### After that, the rule is not considered and a new rule is created
+STUCK_LIMIT = 7  # 7 days
+
+
+def ckey():
+    "Return user CA key either from proxy or userkey.pem"
+    pair = getKeyCertFromEnv()
+    return pair[0]
+
+
+def cert():
+    "Return user CA cert either from proxy or usercert.pem"
+    pair = getKeyCertFromEnv()
+    return pair[1]
+
+
+def parseNewLineJson(stream):
+    """
+    Parse newline delimited json streaming data
+    """
+    for line in stream.split("\n"):
+        if line:
+            yield json.loads(line)
+
+
+def getRucioToken(rucioAuthUrl, rucioAcct):
+    """
+    Provided a Rucio account, fetch a token from the authentication server
+    :param rucioAuthUrl: url to the rucio authentication server
+    :param rucioAcct: rucio account to be used
+    :return: an integer with the expiration time in EPOCH
+    """
+    params = {}
+    headers = {"X-Rucio-Account": rucioAcct}
+
+    url = '%s/auth/x509' % rucioAuthUrl
+    logging.info("Requesting a token to Rucio for account: %s, against url: %s", rucioAcct, rucioAuthUrl)
+    mgr = RequestHandler()
+    res = mgr.getheader(url, params=params, headers=headers, ckey=ckey(), cert=cert())
+    if res.getReason() == "OK":
+        userToken = res.getHeaderKey('X-Rucio-Auth-Token')
+        tokenExpiration = res.getHeaderKey('X-Rucio-Auth-Token-Expires')
+        logging.info("Retrieved Rucio token valid until: %s", tokenExpiration)
+        # convert the human readable expiration time to EPOCH time
+        tokenExpiration = datetime.datetime.strptime(tokenExpiration, "%a, %d %b %Y %H:%M:%S %Z")
+        tokenExpiration = int(tokenExpiration.strftime('%s'))
+        return userToken, tokenExpiration
+
+    raise RuntimeError("Failed to acquire a Rucio token. Error: {}".format(res.getReason()))
+
+
+def renewRucioToken(rucioAuthUrl, userToken):
+    """
+    Provided a user Rucio token, check it's lifetime and extend it by another hour
+    :param rucioAuthUrl: url to the rucio authentication server
+    :param rucioAcct: rucio account to be used
+    :return: a datetime.datetime object with the new token lifetime
+    """
+    params = {}
+    headers = {"X-Rucio-Auth-Token": userToken}
+
+    url = '%s/auth/validate' % rucioAuthUrl
+    logging.info("Renewing the Rucio token...")
+    mgr = RequestHandler()
+    res = mgr.getdata(url, params=params, headers=headers, ckey=ckey(), cert=cert())
+    try:
+        newExpiration = eval(res)['lifetime']
+    except Exception as exc:
+        raise RuntimeError("Failed to renew Rucio token. Response: {} Error: {}".format(res, str(exc)))
+    return newExpiration
+
+
+def getPileupContainerSizesRucio(containers, rucioUrl, rucioToken, scope="cms"):
+    """
+    Given a list of containers, find their total size in Rucio
+    :param containers: list of container names
+    :param rucioUrl: a string with the Rucio URL
+    :param rucioToken: a string with the user rucio token
+    :param scope: a string with the Rucio scope of our data
+    :return: a flat dictionary of container and their respective sizes
+    NOTE: Value `None` is returned in case the data-service failed to serve a given request.
+    NOTE: Rucio version of getPileupDatasetSizes()
+    """
+    sizeByDset = {}
+    if not containers:
+        return sizeByDset
+
+    headers = {"X-Rucio-Auth-Token": rucioToken}
+
+    urls = ['{}/dids/{}/{}?dynamic=anything'.format(rucioUrl, scope, cont) for cont in containers]
+    logging.info("Executing %d requests against Rucio for the container size", len(urls))
+    data = multi_getdata(urls, ckey(), cert(), headers=headers)
+
+    for row in data:
+        container = row['url'].split('/dids/{}/'.format(scope))[1]
+        container = container.replace("?dynamic=anything", "")
+        if row['data'] is None:
+            msg = "Failure in getPileupContainerSizesRucio for container {}. Response: {}".format(container, row)
+            logging.error(msg)
+            sizeByDset.setdefault(container, None)
+            continue
+        response = json.loads(row['data'])
+        try:
+            sizeByDset.setdefault(container, response['bytes'])
+        except KeyError:
+            msg = "getPileupContainerSizesRucio function did not return a valid response for container: %s. Error: %s"
+            logging.error(msg, container, response)
+            sizeByDset.setdefault(container, None)
+            continue
+    return sizeByDset
+
+
+def listReplicationRules(containers, rucioAccount, grouping,
+                         rucioUrl, rucioToken, scope="cms"):
+    """
+    List all the replication rules for the input filters provided.
+    It builds a dictionary of container name and the locations where
+    they have a rule locking data on, with some additional rule state
+    logic in the code.
+    :param containers: list of container names
+    :param rucioAccount: string with the rucio account
+    :param grouping: rule grouping string, only "A" or "D" are allowed
+    :param rucioUrl: string with the Rucio url
+    :param rucioToken: string with the Rucio token
+    :param scope: string with the data scope
+    :return: a flat dictionary key'ed by the container name, with a list of RSEs as value
+    NOTE: Value `None` is returned in case the data-service failed to serve a given request.
+    """
+    locationByContainer = {}
+    if not containers:
+        return locationByContainer
+    if grouping not in ["A", "D"]:
+        raise RuntimeError("Replication rule grouping value provided ({}) is not allowed!".format(grouping))
+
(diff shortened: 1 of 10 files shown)


src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -415,45 +391,143 @@ def makeTransferRequest(self, wflow):
         self.rseQuotas.evaluateQuotaExceeded()
         return success, response
 
-    def _subscribeData(self, subscriptionObj, wflowName, dsetName):
+    def makeTransferRucio(self, wflow, dataIn, subLevel, blocks, dataSize, nodes, nodeIdx):
```
amaltaro, line 394: @todor-ivanov FYI Todor, this is more or less what you will need for the output data placement

todor-ivanov: Thanks @amaltaro  I am looking at your code now.

amaltaro:
For some reason, the `import cert, ckey` from PycurlRucio does not want to work, I get the same import error in my container too:
```
  File "/home/dmwm/unittestdeploy/wmagent/1.3.6.pre4/sw/slc7_amd64_gcc630/external/py2-nose/1.3.7-comp3/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/dmwm/unittestdeploy/wmagent/1.3.6.pre4/sw/slc7_amd64_gcc630/external/py2-nose/1.3.7-comp3/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/dmwm/unittestdeploy/wmagent/1.3.6.pre4/sw/slc7_amd64_gcc630/external/py2-nose/1.3.7-comp3/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/MicroService_t/DataStructs_t/Workflow_t.py", line 8, in <module>
    from WMCore.MicroService.DataStructs.Workflow import Workflow
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/MicroService/DataStructs/Workflow.py", line 10, in <module>
    from WMCore.MicroService.Unified.Common import getMSLogger, gigaBytes
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/MicroService/Unified/Common.py", line 18, in <module>
    from WMCore.MicroService.Tools.PycurlRucio import getPileupDatasetSizesRucio, getPileupSubscriptionsRucio, \
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/MicroService/Tools/PycurlRucio.py", line 20, in <module>
    from WMCore.MicroService.Unified.Common import cert, ckey
ImportError: cannot import name cert
```

it must be missing some subtle detail of the import behavior.
As a workaround, I copied those 2 methods over to the PycurlRucio module.

@todor-ivanov I believe there will be further changes to this PR. But just in case you can have a look at these changes by tomorrow morning, it would be great! Thanks

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -680,45 +650,143 @@ def makeTransferRequest(self, wflow):
...
+            aboveWarningThreshold = self.msConfig.get('warningTransferThreshold') > 0. and \
+                                    dataSize > self.msConfig.get('warningTransferThreshold')
+            if aboveWarningThreshold and rule['ask_approval'] is False:
+                rule['ask_approval'] = True
+
+            # Then make the data subscription, for real!!!
+            res = self.rucio.createReplicationRule(dids, rseExpr, **rule)
+            if res:
+                self.logger.info("Rule replication created under id: %s", res[0])
+                transferId = res[0]
```
todor-ivanov, line 695: Now this one I consider it dangerous. Taking only the first ruleId. One situation where it won't work for sure (especially after the changes I made here [1]) is, if we face a `DuplicateRule` exception internally in the `createReplicationRule` method. Then it for sure will return a list of `ruleIds`, but they will have also historical meaning and one needs to make additional parsing to check which is the one that he actually needs.

[1]
https://github.com/dmwm/WMCore/pull/9772/commits/6bfc5cbb6e797a27ad441a6556fdf61b5fd4db4a#diff-9ac27fcf16ebc6591102e67cfa2e1212R517

src/python/WMCore/MicroService/Tools/PycurlRucio.py
```diff
@@ -0,0 +1,267 @@
...
+        if row['data'] is None:
+            msg = "Failure in getPileupSubscriptionsRucio for dataset {} and block {}".format(dataset, block)
+            msg += " Error: {} {}".format(row.get('code'), row.get('error'))
+            logging.info(msg)
+
+            locationByDset[dataset] = None
+            continue
+        if locationByDset.get(dataset, "non-None") is None:
+            # then one of the block requests failed, skip the whole dataset
+            continue
```
todor-ivanov, line 172: Shouldn't we log our actions here too before we just continue?

todor-ivanov (changes requested):
Alan, I made few minor comments in the code. Most of them could really be skipped, but one or two may be worth taking care of or at least double checked if not changed.

src/python/WMCore/MicroService/Tools/PycurlRucio.py
amaltaro: It will get logged at the `RequestInfo` level, but I agree that it's a good idea to log which query failed here. Log level a few lines above needs to be changed to error as well.

amaltaro: Actually, this `if` will only be executed if the previous one has been evaluated to true (row[data] is None), for the same dataset. This means, one of the block level calls failed for this dataset, and in that case, we simply have to mark that dataset as "unresolved" for the moment and retry in the next cycle.

src/python/WMCore/MicroService/Unified/MSCore.py
```diff
@@ -36,16 +35,12 @@ def __init__(self, msConfig, logger=None):
...
-            # self.rucio = Rucio(self.msConfig['rucioAccount'], configDict={"logger": self.logger})
-            self.phedex = PhEDEx(httpDict={'cacheduration': 0.5},
-                                 dbsUrl=dbsUrl, logger=self.logger)
+        if self.msConfig['useRucio']:
+            self.rucio = Rucio(self.msConfig['rucioAccount'], configDict={"logger": self.logger})
         else:
-            self.phedex = PhEDEx(httpDict={'cacheduration': 0.5},
-                                 dbsUrl=dbsUrl, logger=self.logger)
+            # hard code it to production DBS otherwise PhEDEx subscribe API fails to match TMDB data
+            dbsUrl = "https://cmsweb.cern.ch/dbs/prod/global/DBSReader"
```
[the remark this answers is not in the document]
amaltaro: I think I have tested it in the past with testbed DBSReader and it was giving me some failures (likely the reason for this short comment above).
It could go in the config as well, but given that it won't change between different instances of the service, I decided simply to hard-wire it here.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -618,60 +618,30 @@ def makeTransferRequest(self, wflow):
...
-                    # and update some instance caches
+                if self.msConfig['useRucio']:
+                    success, transferId = self.makeTransferRucio(wflow, dataIn, subLevel,
+                                                                 blocks, dataSize, nodes, idx)
+                else:
+                    success, transferId = self.makeTransferPhedex(wflow, dataBlocks, dataIn,
+                                                                  subLevel, dataSize, nodes, idx)
+                if not success:
+                    # stop any other data placement for this workflow
+                    break
```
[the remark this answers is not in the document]
amaltaro: I was going to say it's already logged from the makeTransferRucio/Phedex methods, but it's a good idea to log it here as well.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -680,45 +650,143 @@ def makeTransferRequest(self, wflow):
...
+        :param wflow: the workflow object
+        :param dataIn: short summary of the data to be placed
+        :param subLevel: subscription level (dataset/container or block)
+        :param blocks: list of blocks to be subscribed (or None if dataset level)
+        :param dataSize: amount of data being placed by this rule
+        :param nodes: list of nodes/RSE
+        :param nodeIdx: index of the node/RSE to be used in the replication rule
+        :return: a boolean flagging whether it succeeded or not, and the rule id
+        """
+        subLevel = "ALL" if subLevel == "dataset" else "DATASET"
```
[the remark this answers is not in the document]
amaltaro: `subLevel` is a subscription level in PhEDEx, which can be either `block` or at `dataset` level.
In Rucio, this is called as `grouping` level, which also mandates how data is transferred/grouped in the same destination. I'll add a comment here too.

src/python/WMCore/MicroService/Unified/MSTransferor.py
```diff
@@ -680,45 +650,143 @@ def makeTransferRequest(self, wflow):
...
+        if self.msConfig.get('enableDataTransfer', True):
+            # Force request-only subscription
+            # to any data transfer going above some threshold (do not auto-approve)
+            aboveWarningThreshold = self.msConfig.get('warningTransferThreshold') > 0. and \
+                                    dataSize > self.msConfig.get('warningTransferThreshold')
+            if aboveWarningThreshold and rule['ask_approval'] is False:
+                rule['ask_approval'] = True
+
+            # Then make the data subscription, for real!!!
+            res = self.rucio.createReplicationRule(dids, rseExpr, **rule)
```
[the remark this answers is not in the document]
amaltaro: It's probably a better idea indeed. Especially because I am not sure whether this AccessDenied is a general issue related to the account/credentials we are using (this is what I assumed in the first place), or whether it could be related to RSE's that we cannot write to (or approve things)... I'll update it too

src/python/WMCore/MicroService/Unified/MSTransferor.py
amaltaro: That's a very good point. I'm updating the code.

amaltaro:
Todor, thanks for your review. Following your review, I have made further changes in my last commit. Please have another look, and if you agree, mark those conversations as resolved.

todor-ivanov:
Thanks @amaltaro I took a quick look at your new commit and I think you addressed everything I mentioned. I think the changes are good to go now.

amaltaro:
@nsmith- and @ericvaandering I think I have finally converged on this PR and tested the most common use cases.
I'm not asking for your review because it became very big, but feel free to go through the initial description at the very top and ask/suggest anything else that you might consider important. 

I might merge it today, but feel free to leave questions/comments at any time and I can follow up on those in a different PR, if needed.

amaltaro:
All right, it should be tested enough by now...

PR #9762, opened 2020-06-25
[py2py3] Applied both futurize stage2 fixers related to supporting python3 stdlib only

### LINKED ISSUE #9761: [py2py3] WMCore should depend on python3 standard library
**Impact of the new feature**

all dmwm/WMCore python code

Right now WMCore python2 code depends on python2 standard library.
As part of the py2 to py2py3 migration, WMCore should depend on python3 standard library.

We should use futurize (in particular the stage2 fixer `libfuturize.fixes.fix_future_standard_library`) to update the WMCore code so that it depends on python3 standard library. 

When running WMCore with python2, the python3 standard library is available thanks to [futurize](http://python-future.org/_modules/future/standard_library.html), in particular adding at the beginning of every affected module

``` 
from future import standard_library
standard_library.install_aliases()
```

mapellidario:
This PR is one of the PRs related to the issue #9761 .

#### Status

not-tested 

#### Description

This PR is about reducing the dependency of WMCore on python2 stdlib, by using specific futurize fixers that allow supporting python3 stdlib.

##### fixers

The stage2 fixers that have been applied are

* `libfuturize.fixes.fix_future_standard_library`
* `libfuturize.fixes.fix_future_standard_library_urllib`

(the complete list is available at http://python-future.org/automatic_conversion.html)

The urllib fixer selects the "easiest option" described in the [official futurize documentation](http://python-future.org/compatible_idioms.html#urllib-module).

We applied both the fixers at the same time because 
despite futurize having a separate fixer for urllib, it would have been impossible 
to completely decouple the two fixers.
If we run `fix_future_standard_library` on a module containing httplib
(that should be converted to http.client) that also contains
`import urlparse`, then `urlparse` is be translated to `urllib.parse`.

Hence, `fix_future_standard_library_urllib` fixers has been applied in the same
PR as `fix_future_standard_library`.

##### no cPickle for the time being

As discussed with Alan in #9735, no changes have been applied to cPickle for performance issues. More informations are available at https://cms-dmwm-test.docs.cern.ch/py2py3/porting/#pickle .
All the automatic changes applied by futurize that translate `cPickle` in `pickle` have been manually discarded.

Achtung! The file `./src/python/WMCore/WMSpec/Persistency.py` depends on both urllib and cPickle. 
No change has been applied to this module. If we want to update urllib to the python3 version while using python2 cPickle,
we could import urllib with

```python
from future.moves.urllib.parse import urlparse, urlencode
from future.moves.urllib.request import urlopen, Request
from future.moves.urllib.error import HTTPError
```

Since this is not consistent with how we imported urllib in other modules, this option has not been applied yet. 

##### log

The log of the PR will be available at https://cms-dmwm-test.docs.cern.ch/py2py3/porting_log/004_stdlib/ (right now no tests have been run and the log is pretty empty)

The purpose of the py2->py2py3 migration is to be as transparent as possible. However, some changes can have some pitfalls. 

The main change that can have some impact on client code is `urllib.urlopen`in py2 to `urllib.request.urlopen` in py2py3 with futurize. This affects the following files:

* src/python/WMCore/Cache/WMConfigCache.py
* src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py

Since `urllib2.urlopen` in python2 behaves in the same way as `urllib.request.urlopen` in python3 (backported to python2 with futurize), then only these two modules may require changes on client code.

What to do if you rely on one of these files:

* check how you deal with HTTP error.  

This is because the python2 version of [urllib.urlopen](https://docs.python.org/2/library/urllib.html#urllib.urlopen) does not raise an exception in case of HTTP error, but it returns valid html containing the description of the error. The python3 version of [urllib.request.urlopen](https://docs.python.org/3.8/library/urllib.request.html#urllib.request.urlopen) raise a URLError in case of a HTTP error. The same behavior is replicated to the functions that are based on `urlopen`, such as `urlretrieve`. If you rely on a file or an string containing html returned from one of the aforementioned modules and you check for errors inspecting the return value, then you need to adapt the error checking with catching an exception, as we did in [this unit test](https://github.com/dmwm/WMCore/pull/9762/files#diff-42be78d8853c84d0c995ae1d05fd0a22). 

Do not hesitate to contact @mapellidario if something is not clear.

#### Related PRs

The related PRs are linked in #9761.

This PR partially supersedes #9735 , in particular in dealing with httplib -> http.client

### DIFF
--- a/bin/wmagent-couchapp-init
+++ b/bin/wmagent-couchapp-init
@@ -5,6 +5,9 @@ wmagent-couchapp-init
 """
 from __future__ import print_function, division
 
+from future import standard_library
+standard_library.install_aliases()
+
 import argparse
 import os
 import shutil
@@ -12,11 +15,7 @@ import sys
 import tempfile
 import urllib
 
-try:
-    from urlparse import urlsplit
-except ImportError:
-    # PY3
-    from urllib.parse import urlsplit
+from urllib.parse import urlsplit
 
 from couchapp.commands import push as couchapppush
 from couchapp.config import Config
--- a/src/python/WMCore/Cache/WMConfigCache.py
+++ b/src/python/WMCore/Cache/WMConfigCache.py
@@ -5,13 +5,15 @@ _WMConfigCache_
 Being in itself a wrapped class around a config cache
 """
 
+from future.utils import with_metaclass
+from future import standard_library
+standard_library.install_aliases()
+
 import hashlib
 import logging
 import traceback
-import urllib
+import urllib.request
 from contextlib import closing
-
-from future.utils import with_metaclass
 
 import WMCore.GroupUser.Decorators as Decorators
 from Utils.Patterns import Singleton
@@ -432,7 +434,7 @@ class ConfigCache(WMObject):
 
         """
         # The newConfig parameter is a URL suitable for passing to urlopen.
-        with closing(urllib.urlopen(newConfig)) as f:
+        with closing(urllib.request.urlopen(newConfig)) as f:
             configString = f.read(-1)
         configMD5 = hashlib.md5(configString).hexdigest()
 
--- a/src/python/WMCore/DataStructs/LumiList.py
+++ b/src/python/WMCore/DataStructs/LumiList.py
@@ -13,11 +13,14 @@ This code began life in COMP/CRAB/python/LumiList.py
 """
 
 
+from future import standard_library
+standard_library.install_aliases()
+
 import copy
 import itertools
 import json
 import re
-import urllib2
+import urllib.request
 from contextlib import closing
 
 class LumiList(object):
@@ -61,7 +64,7 @@ class LumiList(object):
                 self.compactList = json.load(jsonFile)
         elif url:
             self.url = url
-            with closing(urllib2.urlopen(url)) as jsonFile:
+            with closing(urllib.request.urlopen(url)) as jsonFile:
                 self.compactList = json.load(jsonFile)
         elif lumis:
             runsAndLumis = {}
--- a/src/python/WMQuality/WebTools/RESTClientAPI.py
+++ b/src/python/WMQuality/WebTools/RESTClientAPI.py
@@ -1,13 +1,12 @@
+from future import standard_library
+standard_library.install_aliases()
+
 import hashlib
 import hmac
-import urllib
-from httplib import HTTPConnection
+import urllib.parse
 
-try:
-    from urlparse import urlparse
-except ImportError:
-    # PY3
-    from urllib.parse import urlparse
+from http.client import HTTPConnection
+
 from WMCore.WebTools.Page import make_rfc_timestamp
 
 
@@ -29,15 +28,15 @@ def makeRequest(url, values=None, verb='GET', accept="text/plain",
 
     data = None
     if verb == 'GET' and values:
-        data = urllib.urlencode(values, doseq=True)
+        data = urllib.parse.urlencode(values, doseq=True)
     elif verb != 'GET' and values:
         # needs to test other encoding type
         if contentType == "application/x-www-form-urlencoded":
-            data = urllib.urlencode(values)
+            data = urllib.parse.urlencode(values)
         else:
             # for other encoding scheme values assumed to be encoded already
             data = values
-    parser = urlparse(url)
+    parser = urllib.parse.urlparse(url)
     uri = parser.path
     if parser.query:
         uri += "?" + parser.query
--- a/test/python/WMCore_t/WebTools_t/REST_t.py
+++ b/test/python/WMCore_t/WebTools_t/REST_t.py
@@ -7,10 +7,12 @@ Unit tests for checking RESTModel works correctly
 TODO: duplicate all direct call tests to ones that use HTTP
 """
 
+from future import standard_library
+standard_library.install_aliases()
+
 import unittest
 import logging
-import urllib2
-import urllib
+import urllib.request, urllib.error
 import json
 
 from cherrypy import HTTPError
@@ -138,23 +140,33 @@ class RESTTest(RESTBaseUnitTest):
 
     def testException(self):
         """
+        testException
+
         list takes a single integer argument, querying with a string
         """
         url = self.urlbase + 'list?int=a'
-        self.assertRaises(urllib2.HTTPError, urllib2.urlopen, url)
-        # urllib2,urlopen raise the error but not urllib.urlopen
+        self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, url)
+        try:
+            urllib.request.urlopen(url)
+        except urllib.error.HTTPError as e:
+            self.assertEqual(e.code, 400)
+            self.assertEqual(e.reason, u'Bad Request')
+            self.assertEqual(e.msg, u'Bad Request')
+            exception_data = json.loads(e.read())
+            self.assertEqual(exception_data['type'], 'HTTPError')
+            self.assertEqual(exception_data['message'], 'Invalid input: Input arguments failed sanitation.')
+
         url = self.urlbase + 'list1?int=a'
-        expected_data = {"exception": 400, "type": "HTTPError", "message": "Invalid input: Arguments added where none allowed"}
-        urllib_data = urllib.urlopen(url)
-        if self.do_production:
-            #production mode returns 403 error
-            self.assertEqual(urllib_data.getcode(), 403)
-        else:
-            response_data = urllib_data.read()
-            response_data = json.loads(response_data)
-            self.assertEqual(response_data['type'], expected_data['type'])
-            self.assertEqual(response_data['message'], expected_data['message'])
-            self.assertEqual(urllib_data.getcode(), 400)
+        self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, url)
+        try:
+            urllib.request.urlopen(url)
+        except urllib.error.HTTPError as e:
+            self.assertEqual(e.code, 400)
+            self.assertEqual(e.reason, u'Bad Request')
+            self.assertEqual(e.msg, u'Bad Request')
+            exception_data = json.loads(e.read())
+            self.assertEqual(exception_data['type'], 'HTTPError')
+            self.assertEqual(exception_data['message'], 'Invalid input: Arguments added where none allowed')
 
     def testList(self):
         verb ='GET'
@@ -367,7 +379,7 @@ class RESTTest(RESTBaseUnitTest):
     def testAuthentication(self):
         verb ='PUT'
         url = self.urlbase + 'list1'
-        urllib_data = urllib.urlopen(url)
+        urllib_data = urllib.request.urlopen(url)
         self.assertEqual(urllib_data.getcode(), 403)
 
         # pass proper role
--- a/bin/adhoc-scripts/checkDsetFileCount.py
+++ b/bin/adhoc-scripts/checkDsetFileCount.py
@@ -4,24 +4,25 @@ Script meant to be used to check the status of a dataset (or lfn) against DBS
 and PhEDEx. It also prints any discrepancy between those 2 data management tools.
 """
 from __future__ import print_function, division
+from future import standard_library
+standard_library.install_aliases()
 
-import httplib
+import http.client
 import json
 import os
 import sys
-import urllib
-import urllib2
+import urllib.request, urllib.parse
+from urllib.error import HTTPError, URLError
 from argparse import ArgumentParser
-from urllib2 import HTTPError, URLError
 
 main_url = "https://cmsweb.cern.ch"
 phedex_url = main_url + "/phedex/datasvc/json/prod/"
 dbs_url = main_url + "/dbs/prod/global/DBSReader/"
 
 
-class HTTPSClientAuthHandler(urllib2.HTTPSHandler):
+class HTTPSClientAuthHandler(urllib.request.HTTPSHandler):
     def __init__(self):
-        urllib2.HTTPSHandler.__init__(self)
+        urllib.request.HTTPSHandler.__init__(self)
         self.key = os.getenv("X509_USER_PROXY")
         self.cert = os.getenv("X509_USER_PROXY")
 
@@ -29,11 +30,11 @@ class HTTPSClientAuthHandler(urllib2.HTTPSHandler):
         return self.do_open(self.getConnection, req)
 
     def getConnection(self, host, timeout=300):
-        return httplib.HTTPSConnection(host, key_file=self.key, cert_file=self.cert)
+        return http.client.HTTPSConnection(host, key_file=self.key, cert_file=self.cert)
 
 
 def get_content(url, params=None):
-    opener = urllib2.build_opener(HTTPSClientAuthHandler())
+    opener = urllib.request.build_opener(HTTPSClientAuthHandler())
     try:
         if params:
             response = opener.open(url, params)
@@ -55,7 +56,7 @@ def phedex_info(dataset):
     Query blockreplicas PhEDEx API to retrieve detailed information
     for a specific dataset
     """
-    api_url = phedex_url + "blockreplicas" + "?" + urllib.urlencode([('dataset', dataset)])
+    api_url = phedex_url + "blockreplicas" + "?" + urllib.parse.urlencode([('dataset', dataset)])
     phedex_summary = json.loads(get_content(api_url))
     return phedex_summary
 
@@ -65,9 +66,9 @@ def dbs_info(dataset):
     Queries 2 DBS APIs to get both summary and detailed information
     """
     dbs_out = {}
-    api_url = dbs_url + "blocksummaries" + "?" + urllib.urlencode({'dataset': dataset})
+    api_url = dbs_url + "blocksummaries" + "?" + urllib.parse.urlencode({'dataset': dataset})
     dbs_out['blocksummaries'] = json.loads(get_content(api_url))
-    api_url = dbs_url + "files" + "?" + urllib.urlencode({'dataset': dataset}) + "&detail=1"
+    api_url = dbs_url + "files" + "?" + urllib.parse.urlencode({'dataset': dataset}) + "&detail=1"
     dbs_out['files'] = json.loads(get_content(api_url))
     return dbs_out
 
--- a/bin/adhoc-scripts/setrequeststatus.py
+++ b/bin/adhoc-scripts/setrequeststatus.py
@@ -4,7 +4,9 @@ This script can be used to update the status of a request in ReqMgr2.
 """
 from __future__ import print_function, division
 
-import httplib
+from future import standard_library
+standard_library.install_aliases()
+import http.client
 import json
 import os
 import sys
@@ -20,7 +22,7 @@ def setStatus(url, workflow, newstatus):
     else:
         encodedParams = json.dumps({"RequestStatus": newstatus})
 
-    conn = httplib.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
+    conn = http.client.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
     conn.request("PUT", "/reqmgr2/data/request/%s" % workflow, encodedParams, headers)
     resp = conn.getresponse()
     if resp.status != 200:
@@ -37,7 +39,7 @@ def getStatus(url, workflow):
     headers = {"Content-type": "application/json",
                "Accept": "application/json"}
 
-    conn = httplib.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
+    conn = http.client.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
     urn = "/reqmgr2/data/request/%s" % workflow
     conn.request("GET", urn, headers=headers)
     res = conn.getresponse()
--- a/bin/adhoc-scripts/summaryWMStatsFailures.py
+++ b/bin/adhoc-scripts/summaryWMStatsFailures.py
@@ -6,7 +6,9 @@ per task
 """
 from __future__ import print_function, division
 
-import httplib
+from future import standard_library
+standard_library.install_aliases()
+import http.client
 import json
 import os
 import sys
@@ -16,7 +18,7 @@ from pprint import pformat
 def getWMStatsData(workflow):
     url = 'cmsweb.cern.ch'
     headers = {"Accept": "application/json", "Content-type": "application/json"}
-    conn = httplib.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
+    conn = http.client.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
     urn = "/wmstatsserver/data/request/%s" % workflow
     conn.request("GET", urn, headers=headers)
     r2 = conn.getresponse()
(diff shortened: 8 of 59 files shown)


src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -432,7 +434,7 @@ def addConfig(self, newConfig, psetHash=None):
 
         """
         # The newConfig parameter is a URL suitable for passing to urlopen.
-        with closing(urllib.urlopen(newConfig)) as f:
+        with closing(urllib.request.urlopen(newConfig)) as f:
```
mapellidario, line 437: @amaltaro, this line causes the regression https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/10166/testReport/WMCore_t.Cache_t.WMConfigCache_t/testWMConfigCache/testB_addingConfigsAndTweaks/

This is likely a dramatic change in urllib, that apparently does not support file path anymore in py3. more information at https://cms-dmwm-test.docs.cern.ch/py2py3/cheatsheet/urllib_py2/

Some more investigation is required about this matter.

mapellidario: The change is far less dramatic than expected. New version of `urllib` require the specifier `file:` to be present in front of a local file path.

Specifier syntax
* Relative path: `file:./relative/path`
* Absolute path: `file://absolute/path`

What can we do here? I updated the urllib_py2 reference, but in short we could do something like

```
## should work in both python 2 and python3
from future import standard_library
standard_library.install_aliases()
import urllib.request
from contextlib import closing
import os

my_path = "./hello_ascii.txt"
my_url = "file://" + urllib.quote(os.path.abspath(my_path)) # escapes nasty characters

with closing(urllib.request.urlopen(my_url)) as f:
    print f.read()
```

This require some manual changes, but they are not going to be wasted time

mapellidario:
@amaltaro @todor-ivanov, since the interface of urllib changed in the new py3 version, we may need to change a unit test, which is now failing.

In particular, the unit test [WMCore_t.WMRuntime_t.Scripts_t.UnpackUserTarball_t.UnpackUserTarballTest.testBadUrl ](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/10228/testReport/junit/WMCore_t.WMRuntime_t.Scripts_t.UnpackUserTarball_t/UnpackUserTarballTest/testBadUrl/) fails after updating the stdlib and is a regression.

The cause of the regression is at line 60/64 in `src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py`

```python
-   retriever = urllib.urlretrieve
+  retriever = urllib.request.urlretrieve
```

The failing unit test, which is [here](https://github.com/dmwm/WMCore/blob/77a203f17c5955782c27e94e1d8add1915eef3d3/test/python/WMCore_t/WMRuntime_t/Scripts_t/UnpackUserTarball_t.py#L140), calls `retriever("http://home.fnal.gov/~ewv/not-there.txt")`. As the name implies, http://home.fnal.gov/~ewv/not-there.txt does not exist and it return a 404 error. The problem arises since the behavior of `urllib.urlretrieve` changed in python3.
* python2: if the URL does not exist, a file is created and its content is "404 error"
* python3: if the URL does not exist, an exception is raised

Right now, the unit test is written so that it passes if a RuntimeError is raised when the function `UnpackUserTarball` tries to extract as a tar archive the temp file that contains "404 error" (being a text file and not an archive, it fails and that error is catched and the unit test passes). 

After changing the stdlib to the new version, when `UnpackUserTarball`runs, the new version of urllib.request.urlretrieve raises a HTTPError, which is not caught by `assertRaises(RuntimeError, UnpackUserTarball)` and the unit test fails.

We could change the failing unit test to check if `UnpackUserTarball` raises an HTTPError instead of a RuntimeError.

More synthetic tests are available at https://cms-dmwm-test.docs.cern.ch/py2py3/cheatsheet/urllib_py2/#urllib-urlretrieve

amaltaro:
Can't we pass multiple exception types to the assertRaises?

mapellidario:
Yes, we can. But I do not expect that particular unit test to catch RuntimeError after we update `src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py` with the new the stdlib.

amaltaro:
You're right, I misunderstood you.

Have you checked all the places calling `UnpackUserTarball`? Just to make sure the code is prepared to handle a new exception (HTTPError once this fix is made).

mapellidario:
Yep. Apart from [the module itself](https://github.com/dmwm/WMCore/blob/f1985b795455d125d60a0fd30a8af457d85468c9/src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py#L135) and its [unit tests](https://github.com/dmwm/WMCore/blob/f1985b795455d125d60a0fd30a8af457d85468c9/test/python/WMCore_t/WMRuntime_t/Scripts_t/UnpackUserTarball_t.py#L16), `UnpackUserTarball` appears in [the CMSSW executor in WMSpec/Steps](https://github.com/dmwm/WMCore/blob/f1985b795455d125d60a0fd30a8af457d85468c9/src/python/WMCore/WMSpec/Steps/Executors/CMSSW.py#L425), in which the change in the exception raised does not seem to be a concern. Am I correct?

amaltaro:
test this please

test/python/WMCore_t/WebTools_t/REST_t.py
```diff
@@ -141,11 +143,11 @@ def testException(self):
         list takes a single integer argument, querying with a string
         """
         url = self.urlbase + 'list?int=a'
-        self.assertRaises(urllib2.HTTPError, urllib2.urlopen, url)
+        self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, url)
         # urllib2,urlopen raise the error but not urllib.urlopen
         url = self.urlbase + 'list1?int=a'
         expected_data = {"exception": 400, "type": "HTTPError", "message": "Invalid input: Arguments added where none allowed"}
-        urllib_data = urllib.urlopen(url)
+        urllib_data = urllib.request.urlopen(url)
```
mapellidario, line 150: Dear @amaltaro and @todor-ivanov,
This line causes the last unit test failure (actually this is the last "regression", we have other failures that have been present before this PR) that we have: https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/10248/testReport/WMCore_t.WebTools_t.REST_t/RESTTest/testException/

The cause seems to be the same as what we encountered with the previous unit test: `urllib.request.urlopen` raises an exception when there is a http error, instead of return data containing the details of the error.

This unit test is designed to check that the data that urlopen return in case of failures is what we expect.
This is now failing because this line (L150 - `urllib_data = urllib.request.urlopen(url)`) raises an exception before we check it.

if we change this unit test in thew following way:

```python
    def testException(self):
        """
        list takes a single integer argument, querying with a string
        """
        url = self.urlbase + 'list?int=a'
        self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, url)
        # urllib2,urlopen raise the error but not urllib.urlopen
        url = self.urlbase + 'list1?int=a'
-        expected_data = {"exception": 400, "type": "HTTPError", "message": "Invalid input: Arguments added where none allowed"}
+        # expected_data = {"exception": 400, "type": "HTTPError", "message": "Invalid input: Arguments added where none allowed"}
-        urllib_data = urllib.request.urlopen(url)
+        self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, url)
+        # urllib_data = urllib.request.urlopen(url)
-        if self.do_production:
-            #production mode returns 403 error
-            self.assertEqual(urllib_data.getcode(), 403)
-        else:
-            response_data = urllib_data.read()
-            response_data = json.loads(response_data)
-            self.assertEqual(response_data['type'], expected_data['type'])
-            self.assertEqual(response_data['message'], expected_data['message'])
-            self.assertEqual(urllib_data.getcode(), 400)
+        # if self.do_production:
+        #     #production mode returns 403 error
+        #     self.assertEqual(urllib_data.getcode(), 403)
+        # else:
+        #     response_data = urllib_data.read()
+        #     response_data = json.loads(response_data)
+        #     self.assertEqual(response_data['type'], expected_data['type'])
+        #     self.assertEqual(response_data['message'], expected_data['message'])
+        #     self.assertEqual(urllib_data.getcode(), 400)
```
Then the unit test succeeds because we catch the exception. This is not the best option because we only catch the exception, while the purpose of this unit test seems to be having a deeper introspection of the failure.

What can we do? What about something like the following?

```python
        # test 1: only check the exception that is raised
        self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, url)
        # test 2: checks the same exception as above, but introspects also the error code and message
        try:
            urllib.request.urlopen(url)
        except urllib.error.HTTPError as e:
            print("testException", str(e.hdrs), e.url)
            self.assertEqual(e.code, 400)
            self.assertEqual(e.msg, u'Bad Request')
            exception_data = json.loads(e.read())
            self.assertEqual(exception_data['type'], 'HTTPError')
            self.assertEqual(exception_data['message'], 'Invalid input: Arguments added where none allowed')
```

amaltaro: Dario, this code snippet you provided looks good to me. It keeps the same test case. Thanks

mapellidario (commented):
I checked that every time that we switched from using py2 `urllib.urlopen` (or its derived `urllib.urlretrieve`) to py3 `urllib.request.urlopen` (or its derived `urllib.request.urlretrieve`) the new way of throwing exceptions in case of errors instead of writing the details of the error in the response does not cause problems.

We still have something to discuss, especially in `src/python/WMCore/Cache/WMConfigCache.py`.

P.S. Remember that py2 `urllib2.urlopen` behaves in the same way as py3 `urllib.request.urlopen`.

src/python/WMCore/DataStructs/LumiList.py
```diff
@@ -61,7 +64,7 @@ def __init__(self, filename=None, lumis=None, runsAndLumis=None, runs=None, comp
                 self.compactList = json.load(jsonFile)
         elif url:
             self.url = url
-            with closing(urllib2.urlopen(url)) as jsonFile:
+            with closing(urllib.request.urlopen(url)) as jsonFile:
```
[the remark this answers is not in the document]
mapellidario: This is used extensively in dmwm/WMCore. In case of error, it may just be easier to have everything fail where it should instead going through our code and check if the error can be recovered and fail otherwise.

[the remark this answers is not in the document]
mapellidario: Changing L437 of WMConfigCache can have impacts on
https://github.com/cms-PdmV/wmcontrol/blob/1cd4db611d167f73ed4f9ae109ed360a3392eaae/modules/wma.py#L272
for example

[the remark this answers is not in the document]
mapellidario: After some thought, I suggest not to change this line. The behavior is not going to change from  py2 `urllib2` to py3 `urllib`.

bin/wmagent-couchapp-init
```diff
@@ -5,17 +5,16 @@ wmagent-couchapp-init
...
 import shutil
 import sys
 import tempfile
 import urllib
 
-try:
-    from urlparse import urlsplit
-except ImportError:
-    # PY3
     from urllib.parse import urlsplit
```
amaltaro, line 18: indentation

src/python/WMQuality/WebTools/RESTClientAPI.py
```diff
@@ -1,13 +1,13 @@
...
-
-try:
-    from urlparse import urlparse
-except ImportError:
-    # PY3
-    from urllib.parse import urlparse
+import urllib.parse
+
+from http.client import HTTPConnection
+from urllib.parse import urlparse
```
amaltaro, line 9: How about removing this line and updating the line which uses urlparse to `urllib.parse.urlparse...`?

amaltaro:
@mapellidario Dario, these changes look good in general. I have made a couple of comments in the code though.
Just to confirm, with this PR, we stop depending on urllib2, right?

This is a very intrusive change (likely getting merged in the coming days), so we better notify other teams building services on top of WMCore, like @yuyiguo @belforte @ddaina @germanfgv @vkuznet

amaltaro (changes requested):
(now posting it as part of the review...)
@mapellidario Dario, these changes look good in general. I have made a couple of comments in the code though.
Just to confirm, with this PR, we stop depending on urllib2, right?

This is a very intrusive change (likely getting merged in the coming days), so we better notify other teams building services on top of WMCore, like @yuyiguo @belforte @ddaina @germanfgv @vkuznet

amaltaro:
Dario, from the documentation:
https://python-future.org/standard_library_imports.html

I'm worried about this paragraph:
"""
Limitation: Note that the http-based backports do not currently support HTTPS (as of 2015-09-11) because the SSL support changed considerably in Python 3.x. If you need HTTPS support, please use this idiom for now:
"""
since most of our HTTP calls are based on HTTPS. Perhaps you have more information? Otherwise we can also test it on our VMs once this patch gets merged, no big deal.

belforte:
thanks for the heads-up @amaltaro but if this is really intrusive, and I suspect it is, we'll need somethig more. Like a list of what's changed in our dependencies, whether code change in the client (our) side is needed and when/how we can test things. No hurry,  but honestly don't expect me to to do anything until this is more clear.

mapellidario:
#### how to import urllib 

> since most of our HTTP calls are based on HTTPS. Perhaps you have more information? Otherwise we can also test it on our VMs once this patch gets merged, no big deal.

I think that testing it in our VMs is always nice. Nonetheless, I have a queston. is it possible that our Jenkins tests have not cought this sort of error? Is this because in every test external services are mocked in such a way that https is not used? 

In any case, I find comforting  the follwing lines

"If you need HTTPS support, please use this idiom for now:"
```
from future.moves.urllib.request import urlopen
```

#### depending on urllib2

`urllib2` is available on python2 only. This PR is based on the futurize fixer that change all the dependencies on the standard library so that our code is compatible with both py2 and py3. As such, part of the purpose of this PR is removing the dependency on urllib2.

Maybe it is worth stressing that `urllib` interface in python3 is based on `urllib2` in python2, so i do not expect problems in the migration `py2-urllib2 -> py2p3-urllib`. I worry that `py2-urllib -> py2p3-urllib` may be a bit more troublesome, but the unit tests passed, so I am pretty confident that the changes are consistent, at least in a py2 environment. As always, only testing will give us the definitive answer.

#### Notify the other teams about the changes

The change to `src/python/WMCore/Cache/WMConfigCache.ConfigCache.addConfig` is likely to be the more intrusive, as stated in https://github.com/dmwm/WMCore/pull/9762#discussion_r453560009. However, `WMConfigCache` underwent some other major updates in #9818 as well. I included the changes about urllib in that PR also since I thought it would have been merged before this one. Long story short, both this PR and #9818 have the potential to cause some problems to other teams.

belforte:
@mapellidario I am a bit confused. CRAB does not use `WMConfigCache` e.g. but we rely on WMCore for all https calls. So what you do about urllib, including possible exceptions, affects us.
IIUC 59 files are changed in this PR, why do you point out `WMConfigCache` specifically ?
The  one I mostly worry about is `pycurl_manager.py`, but may not need the only thing which changes. 

BTW, in light of recent incident with attempt to move CRABClient to import HTTPException from http.client and lack of py2-futurize in some (many?) CMSSW versions, I am now worried about how to keep CRAB Client working since it relies on WMCore too to make https calls to CMSWEB.
Maybe we should/could have an ad-hoc discussion with also Shazad ?

mapellidario:
I worried about WMConfigCache because I saw it is used in https://github.com/cms-PdmV/wmcontrol/blob/1cd4db611d167f73ed4f9ae109ed360a3392eaae/modules/wma.py#L272 and because we had problems with WMConfigCache.ConfigCache.addConfig in our unit tests. I can go through all the changes that we made in this PR and have an exhaustive list of the possibly braking changes, but it should not be too long. For example, changes to pycurl_manager should be pretty transparent.

I have to admit that I only peeked the usage of WMConfigCache by PdmV when scrolling some code that was linked during a meeting, I have not systematically gone through client code to check how our code is used, I am sorry.

belforte:
that's OK @mapellidario . We simply need to touch base at some point.
E.g. I need to import HTTPException in CRAB code as well, so we better do it in the same way.
Even if signature of WMCore Services does not change, there can be things to know and to take care of.

amaltaro:
@mapellidario how about we use the PR description section "Is it backward compatible" to provide some insights about possible (or necessary) changes on other external services (like, raises a different exception; requires X data type; etc)?

bin/wmagent-couchapp-init
mapellidario: Oops, my bad. Having a second pair of eyes going through the changes is always useful!

src/python/WMQuality/WebTools/RESTClientAPI.py
mapellidario: Definitely tidier this way, thanks!

mapellidario:
As suggested by Alan, I added the information about the effects on client code in the "Is backward compatible?" section of the PR message. Let me know if it is not clear

amaltaro:
test this please

amaltaro:
@mapellidario can you please resolve those 2 unit tests?

amaltaro:
test this please

mapellidario:
Dear Alan, I am sorry if this message will get a bit too long.

I took into consideration your suggestion to consider the python-future documentation suggestion at https://python-future.org/standard_library_imports.html

> Limitation: Note that the http-based backports do not currently support HTTPS (as of 2015-09-11)
because the SSL support changed considerably in Python 3.x. If you need HTTPS support, please use this idiom for now:

I changed how we import `urllib.request` from

```python
from future import standard_library
standard_library.install_aliases()

import urllib.request
```

to

```python
import future.moves.urllib.request
```

After this change, two unit tests started failing _again_ . 

TL;DR I propose to revert those changes and to go back to what we were doing before. I explain why in the following paragraphs.

### urllib, a mess

The py2 first version of this libary is https://docs.python.org/2/library/urllib.html .

It seems that they were not satisfied on how `urllib.urlopen()` worked (they based it on the class `urllib.URLOpener`), so they https://docs.python.org/2/library/urllib2.html that simply contains an improved version of `urlopen()` (based on the class`urllib2.Request`) and almost nothing more. This version of `urlopen()` returns a `URLError` in case of an http error, instead of simply creating a file that contains the information of the error in text form.

Then python3 came and they refactored `urllib`:

* "The urllib module has been split into parts and renamed in Python 3 to urllib.request, urllib.parse, and urllib.error" (from [urllib py2 docs](https://docs.python.org/2/library/urllib.html) )
* "Also note that the urllib.request.urlopen() function in Python 3 is equivalent to `urllib2.urlopen()` and that `urllib.urlopen()` has been removed (from urllib.request in python3 - Ed. )." (again from [urllib py2 docs](https://docs.python.org/2/library/urllib.html)). 
* In python3 `urllib.request` both `urllib.request.Request` (upon which the "new" `urlopen` is build) and `urllib.request.URLOpen` (upon which the old `urlopen` was build) are available. The latter behaves like the former (meaning that in case of an http error an exception is raised) but it likely to be [deprecated in the future](https://docs.python.org/3/library/urllib.request.html#urllib.request.URLopener)
* `urllib.urlopen()` in python2 has been deprecated in favor of `urllib2.urlopen()`

Note also that (from the [urllib py2 docs](https://docs.python.org/2/library/urllib.htm) )

> Warning: For Python versions earlier than 2.7.9, urllib does not attempt to validate the server certificates of HTTPS URIs. Use at your own risk! 

what about `urlretrieve()` (used to "Copy a network object denoted by a URL to a local file" ):

* in python2 `urllib`, it seems very to have the same foundations of `urlopen()` (same way of dealing with errors), it seems to have only a different interface/purpose.
* is has not been implemented in python2 `urllib2`
* It has been ported to [python3 urllib](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve) under as a "Legacy interface" that may be deprecated in the future. I assume it is build upon `urllib.request.URLOpener`

### how python-future deals with urllib

The last version of `futurize` (the automatic translation tool) uses

```python
from future import standard_library
standard_library.install_aliases()

import urllib.request
```

[install_aliases()](https://github.com/PythonCharmers/python-future/blob/28b336151e991ddb8d8a4fb48c48f0e2b9784f55/src/future/standard_library/__init__.py#L443) does nothing in py3 and imports `urllib` from `future.backports` in python2. This behaviour is the same since [25 july 2015](https://github.com/PythonCharmers/python-future/commit/617bddf00e59071b5cc6ba570a1114db01e82946). This backports are the py3 to py2 translation of py3 `urllib`. They are supposed as similar to py3 `urllib` as possible and work smoothly in py2.

As Alan noticed, the [docs](https://python-future.org/standard_library_imports.html#aliased-imports) state that `future.backports.urllib` does not support https in the requests, and suggest using `future.moves.urllib.request` instead. [This](https://github.com/PythonCharmers/python-future/blob/master/src/future/moves/urllib/request.py) mainly moves into the global scope what is in `urllib.request` for py3 and what is in `urllib` and `urllib2` in py2.
The comment has been made in september 2015, some months after the change to `install_aliases()` has been made.

However, [future.backports.urllib](https://github.com/PythonCharmers/python-future/blob/master/src/future/backports/urllib/request.py) has been changed up to 2018 and it includes a [HTTPSHandler](https://github.com/PythonCharmers/python-future/blob/28b336151e991ddb8d8a4fb48c48f0e2b9784f55/src/future/backports/urllib/request.py#L1323).

### what happens in our code

I focused on `src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py` because in this file the change in bahaviour of `urlopen()` is visible.

Before this PR, in the function `UnpackUserTarball()` a tar file was downloaded with py2 `urllib.URLOpen` or `urllib.urlretrieve()` (which is still based on `urllib.URLOpen`) depending on some parameters. In `test/python/WMCore_t/WMRuntime_t/Scripts_t/UnpackUserTarball_t.py` the unit test `testBadUrl` has been written to check if a HTTP error occurred. Since `UnpackUserTarball()` used `urllib.URLOpen`, in case of a HTTP error, instead of creating the requested file, a file with the same name is created and it contains the details of the error, `tar` tries to unpack the tarball but fails because it is a text file, not a tarball, and a Runtime error is raised. This behaviour was asserted in the unit test.

`futurize` suggested to change the import statements adding

```python
from future import standard_library
standard_library.install_aliases()
```

This effectively changes the version of `URLOpen/urlretrieve` from the py2 `urllib` one to the py3 `urllib.request.URLOpen/urlretrieve` backported to py2 one. This means that after running futurize, in case of an HTTP error a URLError is raised and no file is created. This change is likely not to have any breaking effect on our code or on client code (a RuntimeError exception is replaced by a URLError), but we notice this with our unit test. We changed our unit test and everything was working fine.

Then Alan found the incriminated sentence in python future documentation stating that the backported version of future does not support https and I changed the import of `urllib` to

```python
import future.moves.urllib.request
```

Since in py2 `urllib2` there is no `URLOpen/urlretrieve`, in python2 the function that is called is the une from py2 `urllib.URLopen/urlretrieve`. This makes `UnpackUserTarball()` behave as if futurize has not been run over `UnpackUserTarball.py`. This would make `UnpackUserTarball.py` have different behaviour in python2 and python3 environments, because py2 `urllib.URLOpen` behaves differently from the backported version of py3 `urllib.request.URLOpen`

Since I would rather prefer having the same behaviour (at least as much as possible, which means as much as the technology let us) in py2 and in py3, I would prefer using the backported version of py3 `urllib`, which means sticking with

```python
from future import standard_library
standard_library.install_aliases()
```

Since it is not clear to me who to trust, whether the documentation is outdated and not the backported version of urllib supports https or whether the HTTPSHandler is not enough to have full https support, then I would prefer to test this backport. In case the support for https is a problem, then we think about a different solution.

I am sorry if this is a bit long.

amaltaro (approved):
Everything looks good to me. There is one enhancement (no longer the need to catch an ImportError), but we can deal with that in a different PR.
I fully support the idea of having the same behavior in py2 and py3, so I agree with you as well!

PR #9764, opened 2020-06-25
Couple primary/parent block distribution to where the PU dataset is

### LINKED ISSUE #9745: Model for secondary data placement
**Impact of the new feature**
MSTransferor

Yes, it's related to this JIRA ticket:
https://its.cern.ch/jira/browse/CMSCOMPPR-12263

In short, depending on the workflow setup, we might end up with unprocesseable GQEs due to the lack of intersection between primary and secondary data location.
Primary dataset gets distributed in a block-basis all over the SiteWhitelist (sites with quota available); while the PU gets distributed at the dataset level, and we only care about having at least 1 replica of it.

It isn't clear what the best approach is. We could subscribe the pileup data in a block basis too, but then we might risk the physics content of the samples if too many jobs get to execute in a site holding a small fraction of it, where the same PU events would be used much more than if we had the whole dataset available.

Another possibility would be to make multiple replicas of the pileup dataset.

Yet another possibility would be to transfer the primary data only to places holding (part of) the pileup dataset.

Or perhaps we could overwrite the Trust flags post-assignment, to make sure that data can be pulled by the agents and data read remotely.

Any other options that I haven't thought yet?

UPDATE (post discussion): I think that will be the way to go. Make sure the PU dataset is available at at least one site within the SiteWhitelist, and put ALL the primary/parent blocks at the same location(s) that contains the PU. Workflows will take longer to run, but it's probably better than replicating 100TB or more of classic PU... 

a few above

**Additional context**
https://its.cern.ch/jira/browse/CMSCOMPPR-12263

amaltaro:
Fixes #9745 

#### Status
not-tested

#### Description
This changes the current data placement model we had implemented in MSTransferor, thus causing this large amount of changes, complex and delicate. Honestly, I don't like very much this implementation, it's very complicated!

In short, for workflows with pileup data, make sure that the input primary/parent blocks are placed under the same location hosting the pileup dataset(s) (unless AAA flags are enabled).

In plain english, these are the use cases:
1. no pileup dataset
	--> transfer primary within sitewhitelist
2. one pileup dataset
	--> transfer primary where the pileup is, within the sitewhitelist and campaign configuration
3. multiple pileup datasets
	--> all pileup datasets must be available in the same PNN within the sitewhitelist and campaign configuration
    --> we also try to transfer the smallest pileup sample over to where the largest one is already available
    --> then transfer all the input blocks to the same PNN
4. one pileup dataset with secondaryAAA
	--> if pileup is available in a disk PNN, do nothing; else, put it in 1 site within the sitewhitelist and campaign config
5. multiple pileup datasets with secondaryAAA
	--> both pileup datasets need to be available in a disk PNN; else, put it in 1 site within the sitewhitelist and campaign config

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/DefaultStructs.py
+++ b/src/python/WMCore/MicroService/DataStructs/DefaultStructs.py
@@ -30,6 +30,7 @@ TRANSFEROR_REPORT = dict(start_time=0,
                          nodes_out_of_space=None,
                          success_request_transition=0,
                          failed_request_transition=0,
+                         problematic_requests=0,
                          num_datasets_subscribed=0,
                          num_blocks_subscribed=0)
 
--- a/src/python/WMCore/MicroService/DataStructs/Workflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/Workflow.py
@@ -24,6 +24,7 @@ class Workflow(object):
         self.inputDataset = ""
         self.parentDataset = ""
         self.pileupDatasets = set()
+        self.pileupRSEList = set()
 
         self.campaigns = set()
         self.dataCampaignMap = []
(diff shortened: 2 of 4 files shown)


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Imran has been waiting for our request for too long. So I'm merging it and requesting a new testbed deployment this morning.
A review would still be much appreciated.

PR #9766, opened 2020-06-25
Fix python version evaluation in Configuration module

### LINKED ISSUE #9765: Wrong evaluation of python version in the Configuration module
**Impact of the bug**
WMCore services in general (only if running in python3 environment)

**Describe the bug**
As Stefano pointed out, we had a bug in this line:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Configuration.py#L18

which doesn't evaluate to True when running with a python3 binary.

**How to reproduce it**
None

**Expected behavior**
Fix the bug, such that configuration object can be loaded in either python versions.

none

amaltaro:
Fixes #9765 

#### Status
ready

#### Description
And this is what happens when you do not have a python3 environment while porting your code to python3...
Thanks to @belforte for catching this issue.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Configuration.py
+++ b/src/python/WMCore/Configuration.py
@@ -14,8 +14,10 @@ import os
 import sys
 import traceback
 
-# PY3 compatibility
-if sys.version_info == 3:  # PY3 Remove when python 3 transition complete
+PY3 = sys.version_info.major == 3
+
+# PY3 compatibility (can be removed once python2 gets dropped)
+if PY3:
     basestring = str
     unicode = str
     long = int


mapellidario (approved):
This looks like a perfect solution for the time being.

How well this will integrate with modules that import Configuration is yet to be seen and can be postponed to the moment in which we will actually hit this, as with any other problem that require manual change

amaltaro:
Thanks Dario!
Stefano, I'll let you know as soon as we have a tag with this fix in (probably by tomorrow).

belforte:
@amaltaro thanks, but do not bother. I have tried to make CRABClient run in python3 and issues
are more and deeper. Can't be done as a rush one kickj after another. I was too optimistic that the current py3 compatibility checks in Jenkins were enough. I will wait until you have consolidated some progress in WMCore. I'd rather not try to "fix" pycurl_manager myself e.g.  ref: https://github.com/dmwm/CRABClient/issues/4925

PR #9767, opened 2020-06-26
Remove duplicated REQUIRED_OS logic and leftovers from x509 proxy removal in volunteer sites

### LINKED ISSUE #9736: Some CMS@Home jobs end up with the wrong HTCondor requirements
**Impact of the bug**
This bug impacts CMS@Home jobs.  They end up with a requirement that
(AuthenticatedIdentity =!= "volunteer-node@cern.ch") which prevents them from being sent to Volunteer machines.  These jobs sit in the "pending" queue for five days until a time-out leads to their resubmission, usually with the desired requirements.  We use cmsweb-testbed.cern.ch and the agent is vocms0267.cern.ch.

**Describe the bug**
CMS@Home jobs should have the HTCondor requirements
( TARGET.Arch == "X86_64" ) && ( TARGET.OpSys == "LINUX" ) && ( TARGET.Disk >= RequestDisk ) && ( TARGET.Memory >= RequestMemory ) && ( TARGET.Cpus >= RequestCpus ) && ( TARGET.HasFileTransfer )

However, some end up with
( ( ( REQUIRED_OS is "any" ) || ( GLIDEIN_REQUIRED_OS is "any" ) || stringListMember(GLIDEIN_REQUIRED_OS,REQUIRED_OS) ) && ( AuthenticatedIdentity isnt "volunteer-node@cern.ch" ) ) &&    ( TARGET.Arch == "X86_64" ) && ( TARGET.OpSys == "LINUX" ) && ( TARGET.Disk >= RequestDisk ) && ( TARGET.Memory >= RequestMemory ) && ( TARGET.Cpus >= RequestCpus ) && ( TARGET.HasFileTransfer )

The added requirements would appear to come from
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py#L135

Initially, we thought that jobs with the added requirements were those which had failed the condor limit on retries and been passed back to WMAgent, which resubmitted them with the addition, but now we have evidence that in fact some of the initial job submissions already have them.  Jobs with the added requirements stay in the pending queue, and are not distributed on request to Volunteer machines running BOINC tasks.  If these are the only jobs in the queue, the request fails and the BOINC task terminates with an error.  This then has consequences for the Volunteer (failed tasks generate no credit, and repeated failures lead to an automatic decrease in the daily quota of tasks the machine can run).

There is a five-day timeout in the pending queue, after which WMAgent resubmits the job.  As far as we can tell these resubmissions usually have the correct job requirements, as after five days jobs in the particular workflow start running again, and the pending queue decreases.  This five-day delay is somewhat excessive for our requirements (we aim to have jobs that run for 1-2 hours), and puts constraints on our workflow submissions.

**How to reproduce it**
Submit a workflow to T3_CH_Volunteer and observe its behaviour with WMStats:
python inject-test-wfs.py -u https://cmsweb-testbed.cern.ch -m DMWM -f TC_SLC7.json  -c IDR_CMS_Home -r IDR_CMS_Home -t cms-home -a DMWM_Test -p CMS_Home_IDRv5d -s "T3_CH_Volunteer"

**Expected behavior**
We expect that production jobs submitted to T3_CH_Volunteer will all be able to run on Volunteer machines.  Note that post-production jobs (merge, log-collect, etc.) run on a VM farm on T3_CH_CMSAtHome; even if these jobs had the added requirement they would still run as T3_CH_CMSAtHome is not "volunteer-node@cern.ch".

khurtado:
Fixes #9736

#### Status
not-tested

#### Description
Remove duplicated REQUIRED_OS logic, since this is integrated in T2_CH_CERN_HLT already.

Also, AuthenticatedIdentity logic is no longer needed,
as we take care of removing x509 proxies in the volunteer schedds
natively with condor job transforms (which happen before the materialization of jobs in the local queue)

Set self.reqStr to None, unless defined via
config.BossAir.condorRequirementsString

YES

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
@@ -127,17 +127,11 @@ class SimpleCondorPlugin(BasePlugin):
         # Required for global pool accounting
         self.acctGroup = getattr(config.BossAir, 'acctGroup', "production")
         self.acctGroupUser = getattr(config.BossAir, 'acctGroupUser', "cmsdataops")
-
-        # Build a requirement string.  All CMS resources match DESIRED_Sites on the START
-        # expression side; however, there are currently some resources (T2_CH_CERN_HLT)
-        # that are missing the REQUIRED_OS logic.  Hence, we duplicate it here.
-        # TODO(bbockelm): Remove reqStr once HLT has upgraded.
-        self.reqStr = ('((REQUIRED_OS=?="any") || '
-                       '(GLIDEIN_REQUIRED_OS =?= "any") || '
-                       'stringListMember(GLIDEIN_REQUIRED_OS, REQUIRED_OS)) && '
-                       '(AuthenticatedIdentity =!= "volunteer-node@cern.ch")')
+ 
         if hasattr(config.BossAir, 'condorRequirementsString'):
             self.reqStr = config.BossAir.condorRequirementsString
+        else:
+            self.reqStr = None
 
         # x509 proxy handling
         proxy = Proxy({'logger': myThread.logger})
@@ -515,8 +509,13 @@ class SimpleCondorPlugin(BasePlugin):
             ad['Arguments'] = "%s %i %s" % (os.path.basename(job['sandbox']), job['id'], job["retry_count"])
             ad['transfer_output_files'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]
 
-            # Do not define Requirements and X509 ads for Volunteer resources
-            if self.reqStr and "T3_CH_Volunteer" not in job.get('possibleSites'):
+            # Dictionary keys need to be consistent across all jobs within the same 
+            # clusterId when working with queue_with_itemdata()
+            # Initialize 'Requirements' to an empty string for all jobs.
+            # See issue: https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=7715 
+            ad['Requirements'] = ''
+            # Do not define custom Requirements for Volunteer resources
+            if self.reqStr is not None:
                 ad['Requirements'] = self.reqStr
 
             ad['My.x509userproxy'] = classad.quote(self.x509userproxy)


khurtado:
@amaltaro This basically removes `self.reqStr`, as it is no longer needed neither for REQUIRED_OS logic in HLT nor for x509 proxy removal from volunteer sites.

khurtado:
@amaltaro Just changed this a little bit so setting `condorRequirementsString` is still supported (HTCondor gave me a way to solve what I reported in ticket [7715](https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=7715)).

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
```diff
@@ -515,7 +509,12 @@ def getJobParameters(self, jobList):
...
             ad['transfer_output_files'] = "Report.%i.pkl,wmagentJob.log" % job["retry_count"]
 
-            # Do not define Requirements and X509 ads for Volunteer resources
+            # Dictionary keys need to be consistent across all jobs within the same 
+            # clusterId when working with queue_with_itemdata()
+            # Initialize 'Requirements' to an empty string for all jobs.
+            # See issue: https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=7715 
+            ad['Requirements'] = ''
+            # Do not define custom Requirements for Volunteer resources
             if self.reqStr and "T3_CH_Volunteer" not in job.get('possibleSites'):
```
amaltaro, line 518: Should we simply set it if `self.reqStr is not None`?

amaltaro:
@khurtado about this sentence "...as we take care of removing x509 proxies in the volunteer schedds
natively with condor job router rules." is there any possibility that a job would actually start running before the JobRouter tweaking kicks in?

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
khurtado: `ad['Requirements']` needs to be set always. This is what I recently learned from the HTCondor ticket. The itemdata object in condor creates a "table", so if we set a classad in a job, all the other jobs within the same clusterId need to have it set to something too, even if it's just an empty string, as in this case.

khurtado:
@amaltaro No, we use job transforms (performed by the schedd) to remove the x509 proxy. The transformation is executed before the materialization of the job in the queue.

src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py
amaltaro: I see. What I wanted to say was something like:
```
ad['Requirements'] = ''
if self.reqStr is not None:
                ad['Requirements'] = self.reqStr
```

Which means, if the agent is instructed to define a requirement string, just define it regardless of the sites in the whitelist. It's cleaner and provides a better performance, but I do agree that, one day in future, it might cause us problems. Just an idea, I don't really have any preference here.

khurtado: Gotcha! Yes, I agree it is cleaner. Whoever defines `reqStr` in the future can exclude a specific site in the expression anyway (e.g.: `(GLIDEIN_CMSSite =!= "T3_CH_Volunteer") && (more conditions)`), so I will follow your suggestion.

amaltaro:
Thanks Kenyi!

amaltaro (approved):

PR #9770, opened 2020-06-29
Validate/fail workflow assignment for improper use of TrustSitelists/TrustPUSitelists

### LINKED ISSUE #6999: TrustSitelists flag does not work for workflow with no input
Andrew had issues with a TaskChain request:
https://cms-logbook.cern.ch/elog/Workflow+processing/25003

where the second task jobs were all failing with submitfailed (71101 exit code). That means there are no `possiblePSN` in the job pickle file.

The problem seems to come from the fact that the workflow was assigned with TrustSitelists=True, which simply uses the site white and black lists to figure out top level tasks location:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/JobFactory.py#L178

Looking at the splitting parameters passed to the JobFactory, for the first task:

```
{'algorithm': 'EventBased', 'events_per_lumi': 1000, 'siteWhitelist': [u'T1_US_FNAL'], 'events_per_job': 1000, 'siteBlacklist': [], 'deterministicPileup': False, 'lheInputFiles': False, 'trustSitelists': True, 'performance': {'sizePerEvent': 1234.0, 'memoryRequirement': 3000.0, 'timePerEvent': 0.1}}
```

coming from the workload: `....ZTTFS_13.constraints.sites.whitelist = [u'T1_US_FNAL']`

on the other hand, the second task uses the following parameters:

```
{'algorithm': 'LumiBased', 'siteWhitelist': [], 'lumis_per_job': 10, 'deterministicPileup': False, 'siteBlacklist': [], 'trustSitelists': True, 'performance': {'sizePerEvent': 1234.0, 'memoryRequirement': 3000.0, 'timePerEvent': 0.1}}
```

coming from the workload: `...MINIAODMCUP15FS.constraints.sites.whitelist = []`

Now I see two options to fix this issue:
a) properly set the sites.whitelist attribute for all processing/production jobs (I'm afraid it can break unified job routing/overflow)
b) fail request assignment for such cases

@hufnagel @ticoann maybe you have some input on this issue?

amaltaro:
Fixes #6999 

#### Status
tested

#### Description
This contains the ReqMgr2-side changes for a proper evaluation of the Trust flags during the workflow assignment.
In short:
* if the workflow does not have an input dataset, then `TrustSitelists` cannot be set to True (workflow assignment will fail otherwise)
* if the workflow does not have a pileup dataset, then `TrustPUSitelists` cannot be set to True (also failing the workflow assignment otherwise)

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -1870,6 +1870,12 @@ class WMWorkloadHelper(PersistencyHelper):
         it affects where secondary jobs are meant to run.
         The pileup flag has to be set for all the tasks in the workload.
         """
+        if inputFlag is True and not self.listInputDatasets():
+            msg = "Setting TrustSitelists=True for workflows without input dataset is forbidden!"
+            raise RuntimeError(msg)
+        if pileupFlag is True and not self.listPileupDatasets():
+            msg = "Setting TrustPUSitelists=True for workflows without pileup dataset is forbidden!"
+            raise RuntimeError(msg)
         for task in self.getAllTasks(cpuOnly=True):
             if task.isTopOfTree():
                 task.setTrustSitelists(inputFlag, pileupFlag)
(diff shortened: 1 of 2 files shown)


todor-ivanov (approved):
Looks good to me

amaltaro:
I completely overlooked the unit test failures. Will get it fixed in another PR today.

PR #9772, opened 2020-06-30
Add support for Rucio in MSOutput Consumer thread && Handle Duplicate Rule error from Rucio client.

### LINKED ISSUE #9771: Add Support for Rucio in MSOutput
**Impact of the new feature**
 MicroServices

We need to add support for the Rucio Data Management System in the MSOutput Module. The change should happen only in the consumer thread where the actual transfer subscription is happening. The implementation should not require change of the document format in MongoDB.

Add a simple condition in `makeSubscriptions` method in the consumer thread and implement similar logic as the one for  DDM 

None

**Additional context**

todor-ivanov:
Fixes #9771 

#### Status
tested 

#### Description
Adding Support for Rucio to MSOutput. The change is only in the Consumer Thread. 

YES 

#### Related PRs
No

No

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -24,6 +24,7 @@ from WMCore.MicroService.DataStructs.DefaultStructs import OUTPUT_CONSUMER_REPOR
 from WMCore.MicroService.Unified.MSCore import MSCore
 from WMCore.Services.DDM.DDM import DDM, DDMReqTemplate
 from WMCore.Services.CRIC.CRIC import CRIC
+from WMCore.Services.Rucio.Rucio import Rucio
 from Utils.EmailAlert import EmailAlert
 from Utils.Pipeline import Pipeline, Functor
 from WMCore.Database.MongoDB import MongoDB
@@ -51,6 +52,18 @@ class EmptyResultError(MSOutputException):
         else:
             self.myMessage = "EmptyResultError."
         super(EmptyResultError, self).__init__(self.myMessage)
+
+
+class UnsupportedError(MSOutputException):
+    """
+    A MSOutputException signalling an unsupported mode for a function or method.
+    """
+    def __init__(self, message=None):
+        if message:
+            self.myMessage = "UnsupportedError: %s"
+        else:
+            self.myMessage = "UnsupportedError."
+        super(UnsupportedError, self).__init__(self.myMessage)
 
 
 class MSOutput(MSCore):
@@ -109,9 +122,13 @@ class MSOutput(MSCore):
         self.msOutDB = MongoDB(**msOutDBConfig).msOutDB
         self.msOutRelValColl = self.msOutDB['msOutRelValColl']
         self.msOutNonRelValColl = self.msOutDB['msOutNonRelValColl']
-        self.ddm = DDM(url=self.msConfig['ddmUrl'],
-                       logger=self.logger,
-                       enableDataPlacement=self.msConfig['enableDataPlacement'])
+        if self.msConfig['defaultDataManSys'] == 'DDM':
+            self.ddm = DDM(url=self.msConfig['ddmUrl'],
+                           logger=self.logger,
+                           enableDataPlacement=self.msConfig['enableDataPlacement'])
+        elif self.msConfig['defaultDataManSys'] == 'Rucio':
+            self.rucio = Rucio(self.msConfig['rucioAccount'],
+                               configDict={"logger": self.logger})
 
     @retry(tries=3, delay=2, jitter=2)
     def updateCaches(self):
@@ -269,7 +286,7 @@ class MSOutput(MSCore):
         # NOTE:
         #    Here is just an example construction of the function. None of the
         #    data structures used to visualise it is correct. To Be Updated
- 
+
         if self.msConfig['defaultDataManSys'] == 'DDM':
             # NOTE:
             #    We always aggregate per workflow here (regardless of enableAggSubscr)
@@ -320,8 +337,7 @@ class MSOutput(MSCore):
 
                 try:
                     # In the message bellow we may want to put the list of datasets too
-                    msg = "Making transfer subscriptions for %s"
-                    self.logger.info(msg, workflow['RequestName'])
+                    self.logger.info("Making transfer subscriptions for %s", workflow['RequestName'])
 
                     if ddmReqList:
                         ddmResultList = self.ddm.makeAggRequests(ddmReqList, aggKey='item')
@@ -330,7 +346,8 @@ class MSOutput(MSCore):
                         #    Nothing else to be done here. We mark the document as
                         #    done so we do not iterate through it multiple times
                         msg = "Skip submissions for %s. Either all data Tiers were "
-                        msg += "excluded or there were no Output Datasets at all."
+                        msg += "excluded or there were no Output Datasets at all. "
+                        msg += "Marking this workflow as `done`."
                         self.logger.warning(msg, workflow['RequestName'])
                         self.docKeyUpdate(workflow, transferStatus='done')
                         return workflow
@@ -388,17 +405,112 @@ class MSOutput(MSCore):
                     #    reconstructing and returning the same type of object
                     #    as the one that have been passed to the current call.
                     pass
+                # FIXME:
+                msg = "Not yet implemented mode with workflows of type %s!\n" % type(workflow)
+                msg += "Skipping this call"
+                self.logger.error(msg)
+                raise NotImplementedError
             else:
                 msg = "Unsupported type %s for workflows!\n" % type(workflow)
                 msg += "Skipping this call"
                 self.logger.error(msg)
+                raise UnsupportedError
 
         elif self.msConfig['defaultDataManSys'] == 'PhEDEx':
             pass
 
         elif self.msConfig['defaultDataManSys'] == 'Rucio':
-            pass
+            if isinstance(workflow, MSOutputTemplate):
+                self.logger.info("Making transfer subscriptions for %s", workflow['RequestName'])
 
+                rucioResultList = []
+                if workflow['destinationOutputMap']:
+                    for dMap in workflow['destinationOutputMap']:
+                        try:
+                            copies = workflow['numberOfCopies'] if workflow['numberOfCopies'] else 1
+                            # NOTE:
+                            #    Once we get rid of DDM this rseExpression generation
+                            #    should go in the Producer thread
+                            if workflow['isRelVal']:
+                                if dMap['destination']:
+                                    rseUnion = '('+'|'.join(dMap['destination'])+')'
+                                else:
+                                    # NOTE:
+                                    #    If we get to here it is most probably because the destination
+                                    #    in the destinationOutputMap is empty (due to an
+                                    #    'ALCARECO' dataset from a Relval workflow or similar).
+                                    #    Since this is expected to happen a lot, we'd better just
+                                    #    log a warning and continue
+                                    msg = "No destination provided. Avoid creating transfer subscription for "
+                                    msg += "Workflow: %s : Dataset Names: %s"
+                                    self.logger.warning(msg, workflow['RequestName'], dMap['datasets'])
+                                    continue
+                                rseExpression = rseUnion + '&cms_type=real&rse_type=DISK'
+                                # NOTE:
+                                #    The above rseExpression should resolve to something similar to:
+                                #    (T2_CH_CERN|T1_US_FNAL_Disk)&cms_type=real&rse_type=DISK
+                                #    where the first part is a Union of all destination sites and
+                                #    the second part is a general constraint for those to be real
+                                #    entries but not `Test` or `Temp` and we also target only sites
+                                #    marked as `Disk`
+                            else:
+                                rseExpression = '(tier=2|tier=1)&cms_type=real&rse_type=DISK'
+                                # NOTE:
+                                #    The above rseExpression should target all T1_*_Disk and T2_*
+                                #    sites, where the first part is a Union of those Tiers and
+                                #    the second part is a general constraint for those to be real
+                                #    entries but not `Test` or `Temp` and we also target only sites
+                                #    marked as `Disk`
+
+                            if self.msConfig['enableDataPlacement']:
+                                rucioResultList.append(self.rucio.createReplicationRule(dMap['datasets'],
+                                                                                        rseExpression,
+                                                                                        copies=copies))
+                            else:
+                                msg = "DRY-RUN:: The effective Rucio submission would look like: \n"
+                                msg += "account: %s \n"
+                                msg += "dids: %s \n"
+                                msg += "rseExpression: %s\n"
+                                msg += "copies: %s\n"
+                                self.logger.warning(msg,
+                                                    self.msConfig['rucioAccount'],
+                                                    pformat(dMap['datasets']),
+                                                    rseExpression,
+                                                    copies)
+                        except Exception as ex:
+                            msg = "Could not make transfer subscription for Workflow: %s\n:%s"
+                            self.logger.exception(msg, workflow['RequestName'], str(ex))
+                            return workflow
+                else:
+                    # NOTE:
+                    #    Nothing else to be done here. We mark the document as
+                    #    done so we do not iterate through it multiple times
+                    msg = "Skip submissions for %s. Either all data Tiers were "
+                    msg += "excluded or there were no Output Datasets at all. "
+                    msg += "Marking this workflow as `done`."
+                    self.logger.warning(msg, workflow['RequestName'])
+                    self.docKeyUpdate(workflow, transferStatus='done')
+                    return workflow
+
+                transferIDs = rucioResultList
+                if self.msConfig['enableDataPlacement']:
+                    self.docKeyUpdate(workflow,
+                                      transferStatus='done',
+                                      transferIDs=transferIDs)
+                return workflow
+
+            elif isinstance(workflow, (list, set, CommandCursor)):
+                # FIXME:
+                msg = "Not yet implemented mode with workflows of type %s!\n" % type(workflow)
+                msg += "Skipping this call"
+                self.logger.error(msg)
+                raise NotImplementedError
+
+            else:
+                msg = "Unsupported type %s for workflows!\n" % type(workflow)
+                msg += "Skipping this call"
+                self.logger.error(msg)
+                raise UnsupportedError
         # NOTE:
         #    if we are about to implement this through a pipeline we MUST not
         #    return the result here but the WHOLE document with updated fields
@@ -542,7 +654,7 @@ class MSOutput(MSCore):
 
         wfCounterTotal = sum(wfCounters.values())
         return wfCounterTotal
- 
+
     def msOutputProducer(self, requestRecords):
         """
         A top level function to drive the upload of all the documents to MongoDB
@@ -712,7 +824,7 @@ class MSOutput(MSCore):
                 #    Once we migrate to Rucio we should change those defaults to
                 #    whatever is the format in Rucio (eg. referring a subscription
                 #    rule like: "store it at a good site" or "Store in the USA" etc.)
-                destination.add('T1_*')
+                destination.add('T1_*_Disk')
                 destination.add('T2_*')
 
             dMap = {'datasets': [dataset],
@@ -740,7 +852,7 @@ class MSOutput(MSCore):
                         for i in dMap['datasets']:
                             aggMap['datasets'].append(i)
                     found = True
-                    del(dMap)
+                    del dMap
                     break
             if not found:
                 aggDstMap.append(dMap)
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -9,8 +9,10 @@ from __future__ import division, print_function, absolute_import
 
 import logging
 from copy import deepcopy
+from pprint import pformat
+
 from rucio.client import Client
-from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied,
+from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied, DuplicateRule,
                                     DataIdentifierAlreadyExists, DuplicateContent,
                                     UnsupportedOperation, FileAlreadyExists, RuleNotFound)
 from WMCore.WMException import WMException
@@ -478,9 +480,82 @@ class Rucio(object):
         except AccessDenied as ex:
             msg = "AccessDenied creating DID replication rule. Error: %s" % str(ex)
             raise WMRucioException(msg)
+        except DuplicateRule as ex:
+            # NOTE:
+            #    The unique constraint is per did and it is checked against the tuple:
+            #    rucioAccount + did = (name, scope) + rseExpression
+            # NOTE:
+            #    This exception will be thrown by Rucio even if a single Did has
+            #    a duplicate rule. In this case all the rest of the Dids will be
+            #    ignored, which in general should be addressed by Rucio. But since
+            #    it is not, we should break the list of Dids and proceed one by one
+            # NOTE:
+            #    This thing here may be slow, because it will wait for Rucio to
+            #    return the history of rules per every Did, but no shorter path exists
+            msg = "A duplicate rule for: \naccount: %s \ndids: %s \nrseExpression: %s.\n"
+            self.logger.info(msg,
+                             kwargs['account'],
+                             pformat(dids),
+                             rseExpression)
+
+            ruleIds = []
+            didsDup = []
+            for did in dids:
+                try:
+                    response = self.cli.add_replication_rule([did], copies, rseExpression, **kwargs)
+                    for ruleId in response:
+                        ruleIds.append(ruleId)
+                    self.logger.debug("Per did ruleIds: %s", ruleIds)
+                except DuplicateRule:
+                    didsDup.append(did)
+
+            ruleHistory = self.listRuleHistory(didsDup)
+            self.logger.debug("Rule History: %s\n", pformat(ruleHistory))
+
+            for did in ruleHistory:
+                for didHist in did['did_hist']:
+                    ruleIds.append(didHist['rule_id'])
+            ruleIds = list(set(ruleIds))
+            self.logger.debug("ruleIds: %s\n", ruleIds)
+            return ruleIds
         except Exception as ex:
             self.logger.error("Exception creating rule replica for data: %s. Error: %s", names, str(ex))
         return response
+
+    def listRuleHistory(self, dids):
+        """
+        _listRuleHistory_
+
+        A function to return a list of historical records of replication rules
+        per did.
+        :param dids: a list of dids of the form {'name': '...', 'scope: '...'}
+
+        The returned structure looks something like:
+        [{'did': {'name': 'DidName',
+                  'scope': 'cms'},
+          'did_hist': [{u'account': u'wma_test',
+                        u'created_at': datetime.datetime(2020, 6, 30, 1, 34, 51),
+                        u'locks_ok_cnt': 9,
+                        u'locks_replicating_cnt': 0,
+                        u'locks_stuck_cnt': 0,
+                        u'rse_expression': u'(tier=2|tier=1)&cms_type=real&rse_type=DISK',
+                        u'rule_id': u'1f0ab297e4b54e1abf7c086ac012b9e9',
+                        u'state': u'OK',
+                        u'updated_at': datetime.datetime(2020, 6, 30, 1, 34, 51)}]},
+         ...
+         {'did': {},
+          'did_hist': []}]
+        """
+        fullHistory = []
+        for did in dids:
+            didHistory = {}
+            didHistory['did'] = did
+            didHistory['did_hist'] = []
+            # check the full history of the current did
+            for hist in self.cli.list_replication_rule_full_history(did['scope'], did['name']):
+                didHistory['did_hist'].append(hist)
+            fullHistory.append(didHistory)
+        return fullHistory
 
     def listContent(self, name, scope='cms'):
         """


src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                    for dMap in workflow['destinationOutputMap']:
+                        try:
+                            msg = "Making transfer subscriptions for %s"
+                            self.logger.info(msg, workflow['RequestName'])
+                            copies = workflow['numberOfCopies'] if workflow['numberOfCopies'] else 1
+                            # NOTE:
+                            #    Once we get rid of DDM this rseExpression generation
+                            #    should go in the Producer thread
+                            if workflow['isRelVal']:
+                                pnnExpr = '('+'|'.join(["pnn=%s" % pnn for pnn in dMap['destination']])+')'
```
amaltaro, line 420: Is there any difference between `pnn` and `rse` in this expression? RucioInjector component injects data against an `rse`, so we might want to keep things consistent. Just for clarity, can you also expand that expression as a comment in plain english?

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                                rseExpression = pnnExpr + '&cms_type=real&rse_type=DISK'
+                            else:
+                                rseExpression = '(tier=2|tier=1)&cms_type=real&rse_type=DISK'
```
amaltaro, line 423: A comment explaining this expression would be useful too.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                            scope = 'cms'
```
amaltaro, line 424: no need for this scope here, it has a default value defined in the method already.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                            if self.msConfig['enableDataPlacement']:
+                                rucioResultList.append(self.rucio.createReplicationRule(dMap['datasets'],
+                                                                                        rseExpression,
+                                                                                        copies=copies,
+                                                                                        scope=scope))
+                            else:
+                                msg = "Dry run mode. The effective Rucio submission would look like: \n"
```
amaltaro, line 431: To keep it consistent with the MSTransferor code, perhaps make it big, like `DRY-RUN: blah blah`, e.g.:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/MSCore.py#L72

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                                msg += "account: %s \n"
+                                msg += "dids: %s \n"
+                                msg += "rseExpression: %s\n"
+                                msg += "copies: %s\n"
+                                self.logger.warning(msg,
+                                                    self.msConfig['rucioAccount'],
+                                                    pformat([{'scope': scope, 'name': did} for did in dMap['datasets']]),
```
amaltaro, line 438: this scope is also not needed here. I don't think production will ever use anything but `cms`

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -478,6 +480,44 @@ def createReplicationRule(self, names, rseExpression, scope='cms', copies=1, **k
...
+            #    This exception will be thrown by Rucio even if a single Did has
+            #    a duplicate rule. In this case all the rest of the Dids will be
+            #    ignored, which in general should be addressed by Rucio. But since
+            #    it is not, we should break the list of Dids and proceed one by one
+            # NOTE:
+            #    This thing here may be slow, because it will wait for Rucio to
+            #    return the history of rules per every Did, but no shorter path exists
+            msg = "A duplicate rule for: \naccount: %s \ndids: %s \nrseExpression: %s\ncopies: %s\n"
+            ruleHistory = []
+            for did in dids:
```
amaltaro, line 494: This we quickly discussed over slack. We should have a separate method for dealing with the rules history.
In addition to that, can you please add a comment to what a duplicate rule is? AFAIK, it's considered as a unique key of:
rucio account + did + rse expression + scope

If this is correct, then I don't think we need to break down the list of dids and fetch the history for every single of them. We would either have a duplicate rule (for all of them), or for none of them.
@nsmith- might be able to clarify it though.

amaltaro (changes requested):
Todor, I have made many comments along the code. Mostly aesthetic comments though.
For the new Rucio method, given that it's mostly a GET request, I'd suggest to bring it with a unit test as well (using a real NANO data, for the moment).

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -109,9 +110,13 @@ def __init__(self, msConfig, mode, logger=None):
         self.msOutDB = MongoDB(**msOutDBConfig).msOutDB
         self.msOutRelValColl = self.msOutDB['msOutRelValColl']
         self.msOutNonRelValColl = self.msOutDB['msOutNonRelValColl']
-        self.ddm = DDM(url=self.msConfig['ddmUrl'],
-                       logger=self.logger,
-                       enableDataPlacement=self.msConfig['enableDataPlacement'])
+        if self.msConfig['defaultDataManSys'] == 'DDM':
```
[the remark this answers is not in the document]
todor-ivanov: I agree with you and as we discussed yesterday on a chat this may happen only once this PR is merged: #9759

Because we are going to be using the same flag and the Rucio client will be instantiated on the master thread for both services: MSTransferror and MSOutput. Otherwise if I try to re-implement it here again, we will get into git conflicts that need to be resolved later and we are risking to loose one of the two PRs (depending which one is merged first), as it happened the last time.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -109,9 +110,13 @@ def __init__(self, msConfig, mode, logger=None):
...
+            self.ddm = DDM(url=self.msConfig['ddmUrl'],
+                           logger=self.logger,
+                           enableDataPlacement=self.msConfig['enableDataPlacement'])
+        elif self.msConfig['defaultDataManSys'] == 'Rucio':
```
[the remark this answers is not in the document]
todor-ivanov: Same comment as above.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
             pass
 
         elif self.msConfig['defaultDataManSys'] == 'Rucio':
-            pass
+            if isinstance(workflow, MSOutputTemplate):
+                msg = "Making transfer subscriptions for %s"
```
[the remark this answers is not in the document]
todor-ivanov: I find it more readable like this, and it is also following the same pattern as above (for the DDM data management system ): 
https://github.com/dmwm/WMCore/blob/b8326f9e07ef9cb59e126dafb0008f72c8767845/src/python/WMCore/MicroService/Unified/MSOutput.py#L323-L324

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                self.logger.info(msg, workflow['RequestName'])
 
+                rucioResultList = []
+                if workflow['destinationOutputMap']:
+                    for dMap in workflow['destinationOutputMap']:
+                        try:
+                            msg = "Making transfer subscriptions for %s"
```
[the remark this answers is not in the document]
todor-ivanov: Actually as  I think twice I think this one is either a leftover or I wanted to log the current list of datasets from the destination map for which the rule is about to be created. I will remove it for the moment, and if we want to keep that much log information, I can add it back with `dMap` details instead.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I haven't notice any difference between them both so far. My only concern was that the `rse` may be more tricky to combine with additional requirements in the expression as ```  + '&cms_type=real&rse_type=DISK'``` And I may be totally wrong here btw. @nsmith- can you confirm ?

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: The explanation for both of them was added as comment in the code.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I know are never supposed to use anything else but the default 'cms' value comming from the method definition, which I cannot access from here, but I need it in order to build the porpper log message for the DRY_RUN mode, bellow. That's why I passed it explicitly here too, so I can reuse it later in the message.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Agreed. Done!

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: I think it doesn't hurt to see the transfer submission, printed exactly the same way as it would be sent to Rucio if the run mode was `REAL`. That's why I put it here and above too.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                                                    copies)
+                        except Exception as ex:
+                            msg = "Could not make transfer subscription for Workflow: %s\n:%s"
+                            self.logger.exception(msg, workflow['RequestName'], str(ex))
+                            return workflow
+                else:
+                    # NOTE:
+                    #    Nothing else to be done here. We mark the document as
+                    #    done so we do not iterate through it multiple times
+                    msg = "Skip submissions for %s. Either all data Tiers were "
```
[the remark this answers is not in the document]
todor-ivanov: Agreed. Done!

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,69 @@ def makeSubscriptions(self, workflow):
...
+                self.docKeyUpdate(workflow,
+                                  transferStatus='done',
+                                  transferIDs=transferIDs)
+                return workflow
+
+            elif isinstance(workflow, (list, set, CommandCursor)):
+                # FIXME
+                pass
+            else:
+                msg = "Unsupported type %s for workflows!\n" % type(workflow)
```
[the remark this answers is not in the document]
todor-ivanov: I think this will not help too much, because it is not related to the workflow type, but rather to destingush the way/mode in which the `makeTransferSubscription` method is called - either for a single workflow or for a set of workflows. This line here should never be hit at  runtime. If we end up here this is a definite sign for an error in the code. I would rather think of rising an exception here. Your opinion is quite valuable here, so please do share it.

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: Yes as discussed earlier I am working on that. 

And yes, the unique constrain you mention is correct with just one minor but really important detail. We do not submit per single `did` but rather a whole list of `dids`. And this constrain looked like an Oracle constrained to me (one of the exceptions I received was kind of pointing in that direction), and since it seems to be a general `unique` constrain it is raising the `DuplicateRule` exception regardless of the number of the `dids` in the submission request violating it. Which immediately skips the rule creation for the ones that are in the same list but are not violating this requirement. and masks that quite well.  Which immediately turns into a silent bug. So Yes, I am quite sure we must break the list in pieces here. I tested  this  several times in many different configurations.My oppinion is that  this think should be addressed in Rucio though, because asking the full history per every single did from our side may result in prolonged queries.

nsmith-: indeed the tuple `rucio account + did (=name, scope) + rse expression` is the unique key for rules

src/python/WMCore/MicroService/Unified/MSOutput.py
nsmith-: Please use `rse=`.
`pnn` is only used for the syncing script in the case the RSE name != PNN name (essentially only true for `*_Tape` as pnn=`*_MSS`) and it might go away eventually

nsmith-: Really, you don't even need `rse=`, for example:
```bash
$ rucio list-rses --expression "rse=T2_CH_CERN|rse=T2_US_MIT"
T2_CH_CERN
T2_US_MIT
$ rucio list-rses --expression "T2_CH_CERN|T2_US_MIT"
T2_CH_CERN
T2_US_MIT
```

todor-ivanov: ok! Thanks @nsmith- Switching to `rse` then. I think I tried to add the additional constrain directly to SiteNames Unions and it was raising an exception for malformed `rseExpression`. I will give it a second chance though.

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: Yes, but the `did` is still single not plural :) and we are in the situation I explained above. Could this ever be addressed in Rucio or it would be too complicate and we just keep it as it is. The way I handle this is doing the job anyway. We are just risking a little bit of longish queries.

nsmith-: Ok, I see. If you provide multiple DIDs, it just loops over them and makes a rule for each. It may be that, given one DID in the list hits the duplicate rule exception, the others are processed as normal, but I am not sure.

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
amaltaro: I hope I can finish that PR tomorrow, but if you want to not depend on that change, you can reuse (copy/paste) it from there and have it implemented here. Anyway is fine with me though.
Note that you will need to change the ms-output configuration with similar changes as I applied here:
https://github.com/dmwm/deployment/pull/924/files

Once you have a deployment PR, please link it from the description of this PR.

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
amaltaro: Yes, and I use it too. However, I usually write it in multiple lines if we would end up with a line too long. For this case, the line is fairly short
```
self.logger.info("Making transfer subscriptions for %s", workflow['RequestName'])
```
and having all the print related statement in a single line looks better (in addition to counting to less LOC).

This would also be consistent with the way logging has been implemented in WMCore over the last few years. The more uniformity we can agree in the code, the better IMO.

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: That's fine! If you want to show the scope in the DRY-RUN mode, this will have the same effect than writing `cms` in the log message though...

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: I actually wanted to save those cpu cycles spent in the for loop. So one could something like:
```
msg += "scope: %s \n"
msg += "dids: %s \n" etc etc
```

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
amaltaro: That's a good idea! Raise an exception with a clear message, such that placement for this workflow fails - without affecting the rest of the workflows in the pipeline.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -397,8 +402,82 @@ def makeSubscriptions(self, workflow):
...
+                    return workflow
 
+                transferIDs = rucioResultList
+                if self.msConfig['enableDataPlacement']:
+                    self.docKeyUpdate(workflow,
+                                      transferStatus='done',
+                                      transferIDs=transferIDs)
+                return workflow
+
+            elif isinstance(workflow, (list, set, CommandCursor)):
```
amaltaro, line 474: I thought I had made a comment on this block if code. Can you please elaborate a bit here Todor? Is this another case that should never happened and we better go with raising an exception?

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: Todor, please correct me if I misunderstood your comment.
When a list of dids is passed to the Rucio server, Rucio does NOT deal with them in a single transaction? Meaning that some dids might succeed to have a rule created, while others not. Have you actually seen this behaviour in your tests?
If that's correct, then that's a serious problem that needs to be addressed on the server side. Otherwise clients have to implement all these extra checks in order to get their multi-dids rules going through.

The rucio slack channel might be a good place to confirm this behaviour. Can you please confirm it with the experts before we take any decision here?

amaltaro (commented):
Todor, I have made further comments along the code.

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: First, let me apologize for the longish comment I am about to make, but I want to clear any misunderstanding.  
Hi @nsmith- This should be the correct final result, Yes. But it does not happen like that. Even if a single `did` violates the `unique` constrain the whole batch is dropped. This I demonstrate few lines bellow. While @amaltaro, I think `Ricio Client` behaves exactly the opposite of what you say - it for sure tries to deal with the whole list in a single transaction, and that's exactly why it fails the hole list, but not only the ones who are the actual intruders.  

Let me paste some small snippet of code here [1].  And the output from it [2]. While creating the list of `dids`, I chose two datasets that already have previously created rules for them, and one dataset for which I am sure there is none. As you can see, at the end, the whole transaction is dropped, the `DuplicateRule` Exception is thrown for the whole call and the last dataset is left behind, without any notice that it is actually fine to have a rule created for it. So the caller of the method just skips the whole run presuming there are already rules created for his datasets (may be from previous runs or what so ever) and continues with his life, and there is the bug - the last dataset is left without a rule and unnoticed and no rule creation is repeated - ergo 'lost'.



[1]
```
In [35]: from rucio.client import Client
    ...: try:
    ...:     account = 'wma_test'
    ...:     rcl = Client(account=account)
    ...:     scope = 'cms'
    ...:     dids=[{'name': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11/GEN-SIM-RECO', 'scope': scope},
    ...:           {'name': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-DigiFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11/GEN-SIM-DIGI-RAW', 'scope': scope},
    ...:           {'name': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_HG1803_Validation_TEST_Alan_v15-v11/GEN-SIM-RECO', 'scope': scope}]
    ...:     rseExpression = '(tier=2|tier=1)&cms_type=real&rse_type=DISK'
    ...:     copies = 1
    ...:     rcl.add_replication_rule(dids, copies, rseExpression)
    ...: except DuplicateRule as ex:
    ...:     msg = "\nA duplicate rule for: \naccount: %s \ndids: %s \nrseExpression: %s\ncopies: %s\n"
    ...:     ruleHistory = []
    ...:     for did in dids:
    ...:         didHistory = {}
    ...:         didHistory['did'] = did['name']
    ...:         didHistory['did_hist'] = []
    ...:         for hist in rcl.list_replication_rule_full_history(scope, did['name']):
    ...:             didHistory['did_hist'].append(hist)
    ...:         ruleHistory.append(didHistory)
    ...:     msg += "Rule History: \n%s\n"
    ...:     logger.debug(msg,
    ...:                  account,
    ...:                  pformat(dids),
    ...:                  rseExpression,
    ...:                  copies,
    ...:                  pformat(ruleHistory))
```


[2]
```
DEBUG:rucio.client.baseclient:no auth_type passed. Trying to get it from the environment variable RUCIO_AUTH_TYPE and config file.
DEBUG:rucio.client.baseclient:no creds passed. Trying to get it from the config file.
DEBUG:rucio.client.baseclient:no ca_cert passed. Trying to get it from the config file.
DEBUG:rucio.client.baseclient:got token from file
DEBUG:dogpile.lock:value creation lock <dogpile.cache.region._LockWrapper object at 0x7f64801ffd90> acquired
DEBUG:dogpile.lock:Calling creation function for previously expired value
DEBUG:dogpile.lock:Released creation lock
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): cmsrucio-int.cern.ch:80
DEBUG:urllib3.connectionpool:http://cmsrucio-int.cern.ch:80 "POST /rules/ HTTP/1.1" 409 None
DEBUG:urllib3.connectionpool:http://cmsrucio-int.cern.ch:80 "GET /rules/cms/%2FRelValH125GGgluonfusion_13%2FCMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11%2FGEN-SIM-RECO/history HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:http://cmsrucio-int.cern.ch:80 "GET /rules/cms/%2FRelValH125GGgluonfusion_13%2FCMSSW_8_1_0-DigiFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11%2FGEN-SIM-DIGI-RAW/history HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:http://cmsrucio-int.cern.ch:80 "GET /rules/cms/%2FRelValH125GGgluonfusion_13%2FCMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_HG1803_Validation_TEST_Alan_v15-v11%2FGEN-SIM-RECO/history HTTP/1.1" 200 0
DEBUG:__main__:
A duplicate rule for: 
account: wma_test 
dids: [{'name': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11/GEN-SIM-RECO',
  'scope': 'cms'},
 {'name': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-DigiFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11/GEN-SIM-DIGI-RAW',
  'scope': 'cms'},
 {'name': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_HG1803_Validation_TEST_Alan_v15-v11/GEN-SIM-RECO',
  'scope': 'cms'}] 
rseExpression: (tier=2|tier=1)&cms_type=real&rse_type=DISK
copies: 1
Rule History: 
[{'did': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11/GEN-SIM-RECO',
  'did_hist': [{u'account': u'wma_test',
                u'created_at': datetime.datetime(2020, 6, 30, 1, 34, 51),
                u'locks_ok_cnt': 9,
                u'locks_replicating_cnt': 0,
                u'locks_stuck_cnt': 0,
                u'rse_expression': u'(tier=2|tier=1)&cms_type=real&rse_type=DISK',
                u'rule_id': u'1f0ab297e4b54e1abf7c086ac012b9e9',
                u'state': u'OK',
                u'updated_at': datetime.datetime(2020, 6, 30, 1, 34, 51)},
               {u'account': u'wma_test',
                u'created_at': datetime.datetime(2020, 6, 30, 1, 36, 37),
                u'locks_ok_cnt': 9,
                u'locks_replicating_cnt': 0,
                u'locks_stuck_cnt': 0,
                u'rse_expression': u'(pnn=T2_CH_CERN|pnn=T1_US_FNAL_Disk)&cms_type=real&rse_type=DISK',
                u'rule_id': u'd1c29da28d6a409daf477e5934f754ce',
                u'state': u'OK',
                u'updated_at': datetime.datetime(2020, 6, 30, 1, 36, 37)}]},
 {'did': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-DigiFullPU_2017PU_TaskChain_PUMCRecyc_Agent114_Validation_TEST_Alan_v18-v11/GEN-SIM-DIGI-RAW',
  'did_hist': [{u'account': u'wma_test',
                u'created_at': datetime.datetime(2020, 6, 30, 1, 34, 51),
                u'locks_ok_cnt': 18,
                u'locks_replicating_cnt': 0,
                u'locks_stuck_cnt': 0,
                u'rse_expression': u'(tier=2|tier=1)&cms_type=real&rse_type=DISK',
                u'rule_id': u'00e40fca61b5461e88b7bc235a47f531',
                u'state': u'OK',
                u'updated_at': datetime.datetime(2020, 6, 30, 1, 34, 52)},
               {u'account': u'wma_test',
                u'created_at': datetime.datetime(2020, 6, 30, 1, 36, 37),
                u'locks_ok_cnt': 18,
                u'locks_replicating_cnt': 0,
                u'locks_stuck_cnt': 0,
                u'rse_expression': u'(pnn=T2_CH_CERN|pnn=T1_US_FNAL_Disk)&cms_type=real&rse_type=DISK',
                u'rule_id': u'718bf04af115451f868f67c6ca71a1a5',
                u'state': u'OK',
                u'updated_at': datetime.datetime(2020, 6, 30, 1, 36, 37)}]},
 {'did': u'/RelValH125GGgluonfusion_13/CMSSW_8_1_0-RecoFullPU_2017PU_TaskChain_PUMCRecyc_HG1803_Validation_TEST_Alan_v15-v11/GEN-SIM-RECO',
  'did_hist': []}]
```

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: No, Actually it is something that must happen. As we discussed during the previous weeks, since we are not anticipating for this service to to become multithreaded ever, the only plan for scaling this if needed was to start  feeding the `Consumer` thread instead of a single document at a time in a set of documents at a time. And since this function needs to change behavior based on the way it is called this is actually how it recognizes it's run mode.  

And I checked, but rising an exception is actually a bad idea - it breaks the pipeline and needs to be handled outside - which leads to not needed complications. As I previously said this, is not expected to change during runtime with some workflows or any change of running conditions. This is to distinguish the actual run mode - how  the pipeline/thread is feeded.

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: And I checked, but rising an exception is actually a bad idea - it breaks the pipeline and needs to be handled outside - which leads to not needed complications. As I previously said this, is not expected to change during runtime with some workflows or any change of running conditions. This is to distinguish the actual run mode - how  the pipeline/thread is feeded

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: It works [1]. I am going to use it like that now. Thanks @nsmith- 

[1]
```
In [45]: for i in rcl.list_rses('(T2_CH_CERN|T1_US_FNAL_Disk)&cms_type=real&rse_type=DISK'):
    ...:     pprint(i)

{u'rse': u'T2_CH_CERN'}
{u'rse': u'T1_US_FNAL_Disk'}
```

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: I will deal with that later today.

todor-ivanov:
@amaltaro Thank you for your review. I think one way or another I addressed all your comments, except the one related to the   `UseRucio` flag. I will do it later today. In the meantime you may want to take a look at my latest commit - I am keeping them separate for the moment, so you can look at the new changes. I will squash them before we merge.

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: I see. This is definitely a bug on the Rucio side. Would you mind creating a GH issue for them under: https://github.com/rucio/rucio ?

Regarding our real use case, the only scenario that I can think of where we would have some dids with rules, and some not (within the same rule uniqueness), is that a partial rule gets created for a sub-set of the dids that we provide. This can only happen in the following situations:
* if the client makes one HTTP call for each did in the list of dids (thus, single transaction on the server side)
* if the server breaks the request into multiple database transactions, one for each did

As Rucio relies heavily on Oracle, I'd say they probably implemented all the dids rule creation in a single transaction. Still, we should contact them to make sure that we only implemented what is needed. Please check their slack channel today during CERN time: https://github.com/rucio/rucio#getting-support

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: And what does it mean having a `CommandCursor` object type? Is it a generator object yielding data straight from Mongo?

Instead of this `pass` statement, perhaps we should then add something like:
```
raise NotImplementedError("This block needs to be implemented to support bulk workflow operations")
```

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
amaltaro: Please update it.

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: Please update it as well.

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: I think this can be updated as well (then check if pformat still needs to be imported)

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
amaltaro: I see. I think we should log the content of `workflow` then, in addition to its type. Unless it's already done in the previous blocks of code.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -478,10 +480,79 @@ def createReplicationRule(self, names, rseExpression, scope='cms', copies=1, **k
...
+            #    rucioAccount + did = (name, scope) + rseExpression
+            # NOTE:
+            #    This exception will be thrown by Rucio even if a single Did has
+            #    a duplicate rule. In this case all the rest of the Dids will be
+            #    ignored, which in general should be addressed by Rucio. But since
+            #    it is not, we should break the list of Dids and proceed one by one
+            # NOTE:
+            #    This thing here may be slow, because it will wait for Rucio to
+            #    return the history of rules per every Did, but no shorter path exists
+            msg = "A duplicate rule for: \naccount: %s \ndids: %s \nrseExpression: %s\ncopies: %s\n"
```
amaltaro, line 495: This should be logged before we try to check the rule history (I also think we should mention that the rule history will be checked for X dids. Please make it INFO level

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -478,10 +480,79 @@ def createReplicationRule(self, names, rseExpression, scope='cms', copies=1, **k
...
+            for did in dids:
+                try:
+                    response = self.cli.add_replication_rule([did], copies, rseExpression, **kwargs)
```
amaltaro, line 498: Don't we need to save all of the response rules id? Then we only list the rule history for those that really have a DuplicateRule.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -478,10 +480,79 @@ def createReplicationRule(self, names, rseExpression, scope='cms', copies=1, **k
...
+                except DuplicateRule:
+                    pass
+            ruleHistory = self.listRuleHistory(dids)
```
amaltaro, line 501: And this probably deserves a `checkRuleExists` method :-D

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: I wouldn't touch this piece here. Trying to log something you do not know what it is or how long it is, is not a good idea. Especially when you know it is a clear signal that the method is misused. This should stay as a clear message to the programmer that the call is wrong. It will not help us logging workflow contents, because in this situation the worklflow won't be relevant at all (and it will be broken for any workflow), it may just create obscure message and distract the one who is looking at the logs from the important fact - this is a signal that the method is called in a mode it is not designed to work. Or even worse trying to log something of an unknown type may break it in unexpected way and again miss the important message.

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Done. But I kept `pformat` - otherwise the message get completely unstructured.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Yes this is  a generator coming from MongoDB. 
But I think rising such an error is an unneeded complication.  We just need to implement the code for working on a set of workflows.

todor-ivanov: Upon a second thought, I see your point here and I like the idea. I am implementing the additional class and adding this exception here. Just to mention: this will break the whole module and would require someone to come and take care of the error. But since it should not happen at runtime, we may be safe.

todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: Ok! Lets break the whole module in case we run into this one. Mybe there is some point of fully attracting the attention of the developers like that. I implemented a `UnsupportedError` exception, similarly to the `NotImplementedError` one.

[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: I thought this was exactly when it is logged, but yes, I see it is just the message that is created at the proper place but the logging is happening down bellow. I am fixing this now.

todor-ivanov: Done

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: I do not think so. We will not benefit from this, only a bunch of `if` conditions would come into the code which would make it harder to read. If it turns out that we suffer too much performance we can think of doing so, but even then I foresee not that much of optimization at the end.

todor-ivanov: Ok, again :) I made few gentle touches to the code and we have this optimization now. Coming with my next commit.

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: I wouldn't try to make it too perfect though ;)

src/python/WMCore/Services/Rucio/Rucio.py
nsmith-: Why not loop through the DIDs on the client side for safety?
You can catch the duplicate rule exception there, at the expense of more API calls.

todor-ivanov: Hi @nsmith- , Yes we are doing exactly that, and it works. It is simply slower doing it on the client side, because it waits for getting the full history data per `did` and then it (eventually) parses it. This could be much faster doing it on the server side. And also it may be more clear to the other users too, if they could have a response from the server giving the information for which of the `dids` a rule was created and for which not.

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
todor-ivanov: @amaltaro  are you quite far from finalizing this PR https://github.com/dmwm/WMCore/pull/9759 . I see we are not overlapping in files changed in our PRs and I would like to keep it like that. So It would be much cleaner if we merge your PR first and then I rebase  mine on top of that - this  way we will avoid eventual git conflicts, that would require manual resolution.

todor-ivanov:
@amaltaro could you please look at the Final version of this PR.  And let me mention about the unit tests, before you started asking where are they :) Last night I spent some time trying to setup and develop those but because of MongoDB dependencies this task turned into a non trivial one. For testing any of the methods inside the `MSOutput` Class, I need an instance of it, but then we need the MongoDb to be up and running in the container for the unit tests - which is not the case for the moment. And the whole thing is breaking with [1]. And even more, I have no idea how would this behave when the Unit test is about to be run by Jenkins. So I did not push whatever I have so far because it will simply continue breaking. 





[1]
```
[dmwm@a4e07b4a650f Unified_t]$ nosetests MSOutput_t.py
/home/dmwm/unittestdeploy/wmagent/1.3.6.pre4/sw/slc7_amd64_gcc630/external/py2-cryptography/2.4.2-comp2/lib/python2.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py:163: CryptographyDeprecationWarning: OpenSSL version 1.0.1 is no longer supported by the OpenSSL project, please upgrade. A future version of cryptography will drop support for it.
  utils.CryptographyDeprecationWarning
E
ERROR: Test the getCampaignConfig method
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/MicroService_t/Unified_t/MSOutput_t.py", line 47, in setUp
    self.ms = MSOutput(self.msConfig, 'MSOutputConsumer')
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/MicroService/Unified/MSOutput.py", line 134, in __init__
    self.msOutDB = MongoDB(**msOutDBConfig).msOutDB
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/Database/MongoDB.py", line 47, in __init__
    raise ex
ServerSelectionTimeoutError: localhost:8230: [Errno 111] Connection refused
-------------------- >> begin captured logging << --------------------
root: DEBUG: Reading mocked SiteDB data from the file /home/dmwm/wmcore_unittest/WMCore/test/python/../data/Mock/SiteDBMockData.json 
root: INFO: Configuration including default values:
...

root: DEBUG: Initializing CRIC with url: https://cms-cric.cern.ch/
root: ERROR: Could not connect to MongoDB server: mongodb://localhost
localhost:8230: [Errno 111] Connection refusedGiving up Now.
--------------------- >> end captured logging << ---------------------

Ran 1 test in 30.127s

FAILED (errors=1)
```

src/python/WMCore/Services/Rucio/Rucio.py
todor-ivanov: @amaltaro The question was forwarded to Rucio Support Team by mail.

amaltaro:
Pylint check/report has a few bold issues that need to be sorted.
Todor, I don't really see a need to override the `NotImplementedError` builtin exception as well. Why can't you simply
```
raise NotImplementedError("implement bulk requests in docUpdater blah blah")
```
?

That's an unneeded complexity in my opinion. I haven't gone through all your comments yet btw, but either we get it done ASAP (within 50min, at most), or Imran won't make another CMSWEB deployment.

todor-ivanov:
I just removed my version of this Exception.

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: thanks, I suggest we leave it open for historical reference!

src/python/WMCore/MicroService/Unified/MSOutput.py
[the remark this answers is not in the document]
amaltaro: Yes, I'd suggest to leave this one open and revisit it once I'm done with 9759, which is not going to be today.

amaltaro (approved):
Thanks Todor. These changes look good to me. However, you just got yourself 3 issues to be created:
* create unit tests for MSOutput
* create unit tests for MSOutputTemplate
* integrate mongodb in our unit test setup (including docker image)

Can you please get those issues created? Thanks

PR #9782, opened 2020-06-30
check for skipped steps in cmsRun exit code chirp, fixes #9769

### LINKED ISSUE #9769: ignore skipped cmsRun steps for setting aggregate Chirp_WMCore_cmsRun_ExitCode
**Impact of the bug**
ElasticSearch job records are harder to interpret, making debugging harder 

**Describe the bug**
When we get a cmsRun failure in a StepChain step, all later steps are skipped, as they should. Skipping a step has a special exit code though (99108), so the later steps all have that exit code. The aggregate cmsRun exit code Chirp_WMCore_cmsRun_ExitCode takes the last step cmsRun exit code though, so for such jobs it always shows 99108, which is not very useful.

**How to reproduce it**
Look at any failed step chain jobs in ElasticSearch.

**Expected behavior**
For debugging it would be much better to have the aggregate cmsRun exit code contain the last "meaningful" cmsRun exit code and ignore any skipped steps.

hufnagel:
Fixes #9769 

#### Status
not-tested

#### Description
Check the overall (aggregate) cmsRun exit code chirp for previously chirped exit codes. If the exit code is already set to a non-zero value, don't do the chirp. This way we keep as the overall (aggregate) value the last meaningful exit code. Same logic for the overall (aggregate) exception message. 

YES

#### Related PRs
None

Only affects job runtime behavior and values reported to job monitoring.

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executor.py
+++ b/src/python/WMCore/WMSpec/Steps/Executor.py
@@ -169,6 +169,30 @@ class Executor(object):
         """
         return None
 
+    def getCondorChirpAttrDelayed(self, key):
+        """
+        _getCondorChirpAttrDelayed_
+
+        Util to call condor_chirp and retrieve the value for the given key
+
+        """
+        value = None
+        condor_chirp_bin = self.getCondorChirpBin()
+        if condor_chirp_bin and os.access(condor_chirp_bin, os.X_OK):
+            args = [condor_chirp_bin, 'get_job_attr_delayed', key]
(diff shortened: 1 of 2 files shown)


hufnagel:
What's the next step on this?

amaltaro:
Perhaps @todor-ivanov or @khurtado could have a look at run some tests. By a quick glance, I have the feeling that some of this code is a duplicate of what Kenyi has already implemented, I might be wrong though.

PR #9785, opened 2020-07-02
reduce default number of pycurl connections to 50

### LINKED ISSUE #9752: Reduce number of DBS concurrent requests in MSTransferor pycurl
**Impact of the new feature**
MSTransferor

DBS is tightening up the throttling limits in DBSReader Global (decreasing it from 10 to 5 requests per backend/DN now):
https://github.com/dmwm/deployment/pull/919

and we better decrease the number of concurrent requests made against DBS as well, to avoid getting the service in an endless failure mode created by the concurrency/throttling in place.

Ideally, the number of concurrent requests should be made as a configuration parameter, which would propagate all the way down to the method actually creating the pycurl requests.

Extra: we should print the amount of requests we have to make and the number of concurrent requests getting created.

If too much work to make it configurable, I think it's okay to have those limits hardwired in the source code.

**Additional context**
none

amaltaro:
Fixes #9752 

#### Status
ready

#### Description
Change the default number of concurrent requests to 50. Mostly because we want to avoid getting throttled by DBS.
Eventually, we should make it configurable in the MicroServices though...

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -368,7 +368,7 @@ def cern_sso_cookie(url, fname, cert, ckey):
     proc.wait()
 
 
-def getdata(urls, ckey, cert, headers=None, options=None, num_conn=100, cookie=None):
+def getdata(urls, ckey, cert, headers=None, options=None, num_conn=50, cookie=None):
     """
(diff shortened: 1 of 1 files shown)


todor-ivanov (approved):
Looks good to me

PR #9818, opened 2020-07-09
[py2py3] Migrate WMCore/Cache/*

### LINKED ISSUE #9806: Migrate WMCore/Cache/* to be compatible with both python2 and python3
**Impact of the new feature**

This involves migrating `WMCore.Cache.WMConfigCache` and `WMCore.Cache.GenericDataCache` all their dependencies. The dependencies list is obtained with [pydeps](https://github.com/thebjorn/pydeps) and is attached at the end of this issue.

Make  `WMCore.Cache.*` compatible with both python2 and python3. 

Run futurize on `WMCore.Cache.*` and their dependencies and then apply the necessary manual changes, starting from the lowest-level modules up to the higher level ones.

**Additional context**

Some more information can be found [on our mkdocs documentation](https://cms-dmwm-test.docs.cern.ch/py2py3/porting_log/003_wmcore_cache/)

This is the dependency diagram (this cxan be obtained repliocating the instructions about pydeps [on dmwm mkdocs documentation](https://cms-dmwm-test.docs.cern.ch/py2py3/tools/pydeps/) )

src_python_WMCore_Cache_WMConfigCache_yesmissing

mapellidario:
Fixes #9806 

#### Status

Done

ACHTUNG! Some of the changes in this PR will conflict with #9762. They are likely to be dealt with in #9762, since it has a lower priority and will likely be merged after this PR is done.

#### Description

Futurize is run on WMCore.Cache.WMConfigCache and the modules that it depends on.

Manual changes are applied to fix what futurize is not able to migrate:

1. Dictionaries

- replaced mydict.$operator() with view$operator(mydict)

  This is a temporary change and will not be needed anymore
  after dropping python2 support. This is used for performance
  reasons when running python2 with idioms that are compatibles
  with python3.

  Resources:
  * http://python-future.org/compatible_idioms.html#dictionaries
  * http://python-future.org/what_else.html#dict

- used a `iter*` operator only once in a loop in which the dictionary that was being looped over is updated

- used the guidelines at
  http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list
  for the cases in which we want to check if a string is
  among the **keys** of a dictionary. no need to use viewkeys for performance reasons.
  we can just use `list(mydict)` or `for key in mydict:` in both py2 and py3.

2. bytes and unicode strings

JSONThunker.py

- In python3 the meaning of `str()` in a return statement changed.
  The code now behaves in the same way as before, but the names
  used to check the string type have been updated to python3 nomenclature

3. urllib

WMConfigCache.py, `urllib.urlopen` changed to `urllib.request.urlopen` by futurize.

- `urllib.request.urlopen` behaves as `urllib2.urlopen`, throwing an
  exception in case of HTTP error, instead of simply returning valid
  data containing the details of the error.
  ACHTUNG! This may require changes to code that use
  `WMCore.Cache.WMConfigCache.ConfigCache.addConfig`
- `urllib.request.urlopen` behaves as `urllib2.urlopen`, requiring to
  specify the `file:` scheme if you want to open a file in a local
  filesystem. This required changing `WMConfigCache_t` unit tests.

Migration plan

step1
- [x] src/python/WMCore/WMException.py

step2
- [x] src/python/WMCore/Lexicon.py
- [x] src/python/WMCore/Algorithms/Permissions.py
- [x] src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
- [x] src/python/WMCore/Services/pycurl_manager.py
- [x] src/python/Utils/CertTools.py

step 3
- [x] src/python/WMCore/Services/Requests.py

step 4
- [x] src/python/WMCore/Database/CMSCouch.py
- [x] src/python/WMCore/GroupUser/Decorators.py

step5
- [x] src/python/WMCore/GroupUser/CouchObject.py

step 6
- [x] src/python/WMCore/GroupUser/Group.py

step 7
- [x] src/python/WMCore/GroupUser/User.py
- [x] src/python/Utils/Patterns.py
- [x] src/python/WMCore/DataStructs/WMObject.py

step 8
- [x] src/python/WMCore/Cache/WMConfigCache.py

yes

requires python-future.org

### DIFF
--- a/src/python/WMCore/Cache/WMConfigCache.py
+++ b/src/python/WMCore/Cache/WMConfigCache.py
@@ -5,14 +5,16 @@ _WMConfigCache_
 Being in itself a wrapped class around a config cache
 """
 
-from future.utils import with_metaclass
 from future import standard_library
 standard_library.install_aliases()
+import urllib.request, urllib.parse, urllib.error
+from builtins import str
+from future.utils import with_metaclass
+from future.utils import viewkeys, viewvalues
 
 import hashlib
 import logging
 import traceback
-import urllib.request
 from contextlib import closing
 
 import WMCore.GroupUser.Decorators as Decorators
@@ -269,12 +271,12 @@ class ConfigCache(WMObject):
         """
         try:
             self.document = self.database.document(id=configID)
-            if 'owner' in self.document.keys():
+            if 'owner' in self.document:
                 self.connectUserGroup(groupname=self.document['owner'].get('group', None),
                                       username=self.document['owner'].get('user', None))
-            if '_attachments' in self.document.keys():
+            if '_attachments' in self.document:
                 # Then we need to load the attachments
-                for key in self.document['_attachments'].keys():
+                for key in viewkeys(self.document['_attachments']):
                     self.loadAttachment(name=key)
         except CouchNotFoundError as ex:
             msg = "Document with id %s not found in couch\n" % (configID)
@@ -301,7 +303,7 @@ class ConfigCache(WMObject):
         attach = self.database.getAttachment(self.document["_id"], name)
 
         if not overwrite:
-            if name in self.attachments.keys():
+            if name in self.attachments:
                 logging.info("Attachment already exists, so we're skipping")
                 return
 
@@ -400,7 +402,7 @@ class ConfigCache(WMObject):
         Retrieve the dataset information for the config in the ConfigCache.
         """
         psetTweaks = self.getPSetTweaks()
-        if not 'process' in psetTweaks.keys():
+        if 'process' not in psetTweaks:
             raise ConfigCacheException("Could not find process field in PSet while getting output modules!")
         try:
             outputModuleNames = psetTweaks["process"]["outputModules_"]
@@ -548,14 +550,14 @@ class ConfigCache(WMObject):
                 # Something's gone wrong with trying to open the configCache
                 msg = "Error in getting output modules from ConfigCache during workload validation.  Check ConfigCache formatting!"
                 raise ConfigCacheException("%s: %s" % (msg, str(ex)))
-            for outputModule in outputModuleInfo.values():
+            for outputModule in viewvalues(outputModuleInfo):
                 dataTier = outputModule.get('dataTier', None)
                 filterName = outputModule.get('filterName', None)
                 if not dataTier:
                     raise ConfigCacheException("No DataTier in output module.")
 
                 # Add dataTier to duplicate dictionary
-                if not dataTier in duplicateCheck.keys():
+                if dataTier not in duplicateCheck:
                     duplicateCheck[dataTier] = []
                 if filterName in duplicateCheck[dataTier]:
                     # Then we've seen this combination before
--- a/src/python/WMCore/Database/CMSCouch.py
+++ b/src/python/WMCore/Database/CMSCouch.py
@@ -12,6 +12,9 @@ from __future__ import print_function, division
 
 from future import standard_library
 standard_library.install_aliases()
+from builtins import str, object
+from future.utils import viewitems
+import urllib.request, urllib.parse, urllib.error
 
 import base64
 import hashlib
@@ -19,7 +22,6 @@ import logging
 import re
 import time
 import traceback
-import urllib.parse
 from datetime import datetime
 from http.client import HTTPException
 
@@ -201,7 +203,7 @@ class Database(CouchDBRequests):
             data[label] = int(time.time())
         else:
             for doc in data:
-                if label not in doc.keys():
+                if label not in doc:
                     doc[label] = int(time.time())
         return data
 
@@ -523,7 +525,7 @@ class Database(CouchDBRequests):
         options = options or {}
         keys = keys or []
         encodedOptions = {}
-        for k, v in options.iteritems():
+        for k, v in viewitems(options):
             # We can't encode the stale option, as it will be converted to '"ok"'
             # which couch barfs on.
             if k == "stale":
@@ -558,7 +560,7 @@ class Database(CouchDBRequests):
         options = options or {}
         keys = keys or []
         encodedOptions = {}
-        for k, v in options.iteritems():
+        for k, v in viewitems(options):
             encodedOptions[k] = self.encode(v)
 
         if keys:
@@ -600,7 +602,7 @@ class Database(CouchDBRequests):
         options = options or {}
         keys = keys or []
         encodedOptions = {}
-        for k, v in options.iteritems():
+        for k, v in viewitems(options):
             encodedOptions[k] = self.encode(v)
 
         if keys:
--- a/src/python/WMCore/Services/Requests.py
+++ b/src/python/WMCore/Services/Requests.py
@@ -12,6 +12,12 @@ from __future__ import division, print_function
 from future import standard_library
 standard_library.install_aliases()
 
+from future import standard_library
+standard_library.install_aliases()
+
+from builtins import str, bytes, object
+from future.utils import viewvalues
+
 import base64
 import logging
 import os
@@ -45,11 +51,6 @@ except ImportError:
     # Mock ServerNotFoundError since we don't want that WMCore depend on httplib2 using pycurl
     class ServerNotFoundError(Exception):
         pass
-
-# Python3 compatibility
-if sys.version.startswith('3.'):  # PY3 Remove when python 3 transition complete
-    basestring = str
-
 
 def check_server_url(srvurl):
     """Check given url for correctness"""
@@ -194,9 +195,7 @@ class Requests(dict):
         if verb == 'GET' and data:
             uri = "%s?%s" % (uri, data)
 
-        # PY3 needed for compatibility because str under futurize is not a string. Can be just str in Py3 only
-        # PY3 Don't let futurize change this
-        assert isinstance(data, (str, basestring)), \
+        assert isinstance(data, (str, bytes)), \
             "Data in makeRequest is %s and not encoded to a string" % type(data)
 
         # And now overwrite any headers that have been passed into the call:
@@ -224,7 +223,7 @@ class Requests(dict):
             # & retry. httplib2 doesn't clear httplib state before next request
             # if this is threaded this may spoil things
             # only have one endpoint so don't need to determine which to shut
-            for con in conn.connections.values():
+            for con in viewvalues(conn.connections):
                 con.close()
             conn = self._getURLOpener()
             # ... try again... if this fails propagate error to client
@@ -262,7 +261,7 @@ class Requests(dict):
                    "User-Agent": "WMCore.Services.Requests/v002",
                    "Accept": self['accept_type']}
 
-        for key in self.additionalHeaders.keys():
+        for key in self.additionalHeaders:
             headers[key] = self.additionalHeaders[key]
         # And now overwrite any headers that have been passed into the call:
         # WARNING: doesn't work with deplate so only accept gzip
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -40,6 +40,10 @@ from __future__ import print_function
 from future import standard_library
 standard_library.install_aliases()
 
+from builtins import str, range, object
+from past.builtins import basestring
+from future.utils import viewitems
+
 # system modules
 import json
 import logging
@@ -51,7 +55,6 @@ import pycurl
 from io import BytesIO
 import http.client
 from urllib.parse import urlencode
-
 
 class ResponseHeader(object):
     """ResponseHeader parses HTTP response header"""
@@ -214,7 +217,7 @@ class RequestHandler(object):
         curl.setopt(pycurl.URL, str(url))
         if headers:
             curl.setopt(pycurl.HTTPHEADER, \
-                    ["%s: %s" % (k, v) for k, v in headers.items()])
+                    ["%s: %s" % (k, v) for k, v in viewitems(headers)])
         bbuf = BytesIO()
         hbuf = BytesIO()
         curl.setopt(pycurl.WRITEFUNCTION, bbuf.write)
@@ -411,14 +414,14 @@ def getdata(urls, ckey, cert, headers=None, options=None, num_conn=50, cookie=No
     for _ in range(num_conn):
         curl = pycurl.Curl()
         curl.fp = None
-        for key, val in options.items():
+        for key, val in viewitems(options):
             curl.setopt(getattr(pycurl, key), val)
         curl.setopt(pycurl.SSLKEY, ckey)
         curl.setopt(pycurl.SSLCERT, cert)
         mcurl.handles.append(curl)
         if headers:
             curl.setopt(pycurl.HTTPHEADER, \
-                        ["%s: %s" % (k, v) for k, v in headers.items()])
+                        ["%s: %s" % (k, v) for k, v in viewitems(headers)])
 
     # Main loop
     freelist = mcurl.handles[:]
--- a/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
+++ b/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
@@ -1,19 +1,10 @@
 from __future__ import print_function
 
+from builtins import str, bytes, int
+from future.utils import viewitems
+
 import sys
 import types
-
-# PY3 compatibility
-# Not clear this function will work under python3 either.
-# nosetests Requests_t.py:testJSONRequests.testSet2 is a minimal test that fails on python2
-# The problem is that isinstance checks on this object in python2 fail if included
-if sys.version.startswith('3.'):
-    from builtins import object
-
-    # Make types compatible, can also be removed once transition is over
-    long = int
-    basestring = str
-
 
 class _EmptyClass(object):
     pass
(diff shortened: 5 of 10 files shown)


src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
```diff
@@ -145,7 +133,7 @@ def handleDictThunk(self, toThunk):
         toThunk = self.checkRecursion(toThunk)
         special = False
         tmpdict = {}
-        for k, v in toThunk.iteritems():
+        for k, v in iteritems(toThunk):
             if type(k) == type(int):
```
mapellidario, line 137: Dear @amaltaro and @todor-ivanov, 
I am a bit baffled by this line. it is comparing the type of a key with `type(int)`, which is just `type`.

It seems that the type of the key is used to prepend a string to the key itself in a temporary dictionary. If the key is an int, we prepend a `_i:` and shortly aftre if it is a float we prepend `_f:`.
I would replace `if type(k) == type(int):` with `if isinstance(k, int)`.

Are my intuitions correct or do we really want to check if a key is of type `type`?

mapellidario:
Dear @alan and @todor, have a look at [last commit's changes](https://github.com/dmwm/WMCore/pull/9818/commits/310ad5cb6588a421b565b27b257a038c7bd87195). 

```python
-        elif isinstance(toThunk, dict):
+        elif type(toThunk) is dict:
```

This fixes a lot of regressions found by jenkins in a previous commit, for example [WMCore_t.WorkQueue_t.Policy_t.Start_t.ResubmitBlock_t/ResubmitBlockTest/testSingleChunksSplit](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/10264/testReport/WMCore_t.WorkQueue_t.Policy_t.Start_t.ResubmitBlock_t/ResubmitBlockTest/testSingleChunksSplit/).

The incriminated previous commit was something that I considered rather safe, 
but it is clear that I have to reconsider myself. It was about some suggestions 
from [the jenkins module that checks for regressions to python versions older
than 2.6](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/10263/artifact/artifacts/PullRequestReport.html). 
As usual, it suggests using isinstance instead of comparing the type directly.

This can, and actually did, break things.

Further investigation revealed that when I run the test

```bash
python setup.py test --buildBotMode=true --reallyDeleteMyDatabaseAfterEveryTest=true --testCertainPath=test/python/WMCore_t/WorkQueue_t/Policy_t/Start_t/ResubmitBlock_t.py
```

With this version of JSONThunker (which raises an exception to catch the stdout log)

```python
        # src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
        # L 250
        elif type(toThunk) is dict:
            print("CHECKPOINT", type(toThunk))
            temp = self.handleDictThunk(toThunk)
            raise Exception("dummy")
            return self.handleDictThunk(toThunk)
```

we get

```plaintext
-------------------- >> begin captured stdout << ---------------------
CHECKPOINT <type 'dict'>

--------------------- >> end captured stdout << ----------------------
```

While if we run it with

```python
        # src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
        # L 250
        elif isinstance(toThunk, dict):
            print("CHECKPOINT", type(toThunk))
            temp = self.handleDictThunk(toThunk)
            raise Exception("dummy")
            return self.handleDictThunk(toThunk)
```

we get instead

```plaintext
-------------------- >> begin captured stdout << ---------------------
CHECKPOINT <type 'dict'>
CHECKPOINT <class 'WMCore.Database.CMSCouch.Document'>
CHECKPOINT <type 'dict'>
CHECKPOINT <type 'dict'>

--------------------- >> end captured stdout << ----------------------
```

If I have to guess, it seems that since `isinstance` checks for inheritance too,
some objects that are derived from `dict` would pass this check. Since this
causes problems in some unit test, I would not apply this change.

This change was not suggested by futurize, and after seeing these errors I understand
why they did not include this in their fixers.


We applied some similar changes (`type() is ` -> `isinstance`) in #9731 . The unit tests did not catch any problem with those changes. There may be some chance that those changes could create some problems, even though my instinct suggest it is going to be remote, since we do not changed the code manipulating some nasty objects, it seemed more traditional stuff. In any case, we should keep an eye on that.

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
amaltaro: That's a trick question. According to google - and my understanding - here we are checking whether `k` is one of the basic data types or a class definition. The type of those objects is `type` ...
isinstance seems to accept subclasses as well, while type only compares against the direct object type.

With that said, I have the felling that this code:
```
elif type(k) == type(float):
...
```

will never evaluate to True, otherwise the if above (type(int)) would have too.

Here I'm not sure as well, but I think we could replace:
```
if type(k) == type(int):
```
by
```
isinstance(k, type) 
```

Some references that I found:
https://realpython.com/python-metaclasses/#type-and-class
and 
https://note.nkmk.me/en/python-type-isinstance/#:~:text=source%3A%20type.py-,With%20isinstance(),an%20instance%20of%20any%20type.

mapellidario: I would agree with your solution if there was a single check. However here we have something like

```python
for k, v in iteritems(toThunk):
    if type(k) == type(int):
        tmpdict['_i:%s' % k] = self._thunk(v)
    elif type(k) == type(float):
         tmpdict['_f:%s' % k] = self._thunk(v)
    else:
         tmpdict[k] = self._thunk(v)
```
and it seems like the author was trying to put in the dictionary key some information about the value type, but erroneously put a `type()` in front of the comparison type.

If we use `isinstance(k, type) ` we would match both the `int` and the `float` case, wouldn't we?

This being said, I also have the same feeling that those lines now never match True. `git blame` brought me to 10y ago, when this file was created. If in all these years this change to the keys has been made, should we even change it now?

todor:
@mapellidario Please note that the contributor to this repo is @todor-ivanov, you are tagging the wrong person. @todor is my account and I am not contributing to this repo, so maybe you will want to remove me and not granting permissions.

amaltaro:
I fixed the mention to Todor and myself in your comment above, Dario.
Apologies for the mistake, Todor.

mapellidario:
Something strange is happening. I added a new commit yo my feature branch, rebased my feature branch on dmwm/master, pushed with `--force` to my remote but github did not seem to have noticed it.

If I open https://github.com/mapellidario/WMCore/tree/py2py3-wmconfigcache it says ` This branch is 14 commits ahead of dmwm:master. `, while here it says that there are 13 commits. I will wait until tomorrow, then I will try to squash commits and push again

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
amaltaro: I think there is something very much wrong with this code. However, as we discussed over slack, let's keep it as it is, file a new GH issue reporting this problem and move on. In a couple of months from now, we can get back to it.

amaltaro:
Yes Dario, try rebasing and squashing everything (that does not involve tests), and it might clear up. Honestly, I'm not sure I understand the problem you reported here.

mapellidario:
I think that it simply took about 1h for github to register [my last force-push event](https://github.com/dmwm/WMCore/pull/9818#event-3549401924) in this PR. All good what turns out to be good.

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
mapellidario: I created the issue #9852 related to this.

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -432,7 +437,7 @@ def addConfig(self, newConfig, psetHash=None):
 
         """
         # The newConfig parameter is a URL suitable for passing to urlopen.
-        with closing(urllib.urlopen(newConfig)) as f:
+        with closing(urllib.request.urlopen(newConfig)) as f:
```
mapellidario, line 440: As the commit message states:

>   ACHTUNG! This may require changes to code that use
  `WMCore.Cache.WMConfigCache.ConfigCache.addConfig`

For example [cms-PdmV/wmcontrol/modules/wma.py](https://github.com/cms-PdmV/wmcontrol/blob/1cd4db611d167f73ed4f9ae109ed360a3392eaae/modules/wma.py#L272)

When merging this, we must state that if someone depends on __some specific exception__ that this function may raise in case of error, then they need to update their code. If they do not care about exceptions, then they should be fine.

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -267,12 +272,12 @@ def loadByID(self, configID):
         """
         try:
             self.document = self.database.document(id=configID)
-            if 'owner' in self.document.keys():
+            if 'owner' in list(self.document):
```
amaltaro, line 275: Doesn't the `in` operator work in both py2 and py3 (without the need to cast to list)? Like
`if 'owner' in self.document:`

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -267,12 +272,12 @@ def loadByID(self, configID):
...
                 self.connectUserGroup(groupname=self.document['owner'].get('group', None),
                                       username=self.document['owner'].get('user', None))
-            if '_attachments' in self.document.keys():
+            if '_attachments' in list(self.document):
```
amaltaro, line 278: same as above

src/python/WMCore/Services/Requests.py
```diff
@@ -269,7 +268,7 @@ def encodeParams(self, data, verb, incomingHeaders, encoder, contentType):
                    "User-Agent": "WMCore.Services.Requests/v002",
                    "Accept": self['accept_type']}
 
-        for key in self.additionalHeaders.keys():
+        for key in self.additionalHeaders:
```
amaltaro, line 271: should we use viewkeys here?

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
```diff
@@ -145,7 +133,7 @@ def handleDictThunk(self, toThunk):
         toThunk = self.checkRecursion(toThunk)
         special = False
         tmpdict = {}
-        for k, v in toThunk.iteritems():
+        for k, v in iteritems(toThunk):
```
amaltaro, line 136: Dario, was there any reason for the iteritems choice (instead of viewitems)?

amaltaro (commented):
Dario, besides a few comments - mostly against dictionary handling - I made in the code. Can you please review all the other dictionaries updated in this PR, I saw a few others that I have not made any comments.

Otherwise, this code looks good to me and we could try to push it to testbed in the beginning of the coming week. I let @todor-ivanov decide though. Thanks

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
mapellidario: You are right!
I fear that when editing this I had the impression that it was not safe to loop over a dynamic list of `toThunk`'s items, but now that I have another look at it `toThunk` is modified outside of the loop over its items and it is safe to use `viewitems`. I am sorry, I do not remember why I have been so conservative. Thanks for the thorough review!

mapellidario:
test this please

mapellidario:
I re-run the unit tests because in my VM [UUID_t/UUIDTest/testUUID](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/10294/testReport/junit/WMCore_t.Services_t.UUID_t/UUIDTest/testUUID/) was not failing. 
Now no unit test is failing. Good!

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -299,7 +304,7 @@ def loadAttachment(self, name, overwrite=True):
         attach = self.database.getAttachment(self.document["_id"], name)
 
         if not overwrite:
-            if name in self.attachments.keys():
+            if name in list(self.attachments):
```
amaltaro, line 307: How about
```suggestion
            if name in self.attachments:
```
?

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -398,7 +403,7 @@ def getOutputModuleInfo(self):
         Retrieve the dataset information for the config in the ConfigCache.
         """
         psetTweaks = self.getPSetTweaks()
-        if not 'process' in psetTweaks.keys():
+        if not 'process' in list(psetTweaks):
```
amaltaro, line 406: How about
```suggestion
        if 'process' not in psetTweaks:
```
?

src/python/WMCore/Cache/WMConfigCache.py
```diff
@@ -546,14 +551,14 @@ def validate(self, configID):
...
-            for outputModule in outputModuleInfo.values():
+            for outputModule in viewvalues(outputModuleInfo):
                 dataTier = outputModule.get('dataTier', None)
                 filterName = outputModule.get('filterName', None)
                 if not dataTier:
                     raise ConfigCacheException("No DataTier in output module.")
 
                 # Add dataTier to duplicate dictionary
-                if not dataTier in duplicateCheck.keys():
+                if not dataTier in list(duplicateCheck):
```
amaltaro, line 561: How about
```suggestion
                if dataTier not in duplicateCheck:
```
?

src/python/WMCore/Database/CMSCouch.py
```diff
@@ -198,7 +203,7 @@ def timestamp(self, data, label=''):
             data[label] = int(time.time())
         else:
             for doc in data:
-                if label not in doc.keys():
+                if label not in list(doc):
```
amaltaro, line 206: How about
```suggestion
                if label not in doc:
```
?

src/python/WMCore/Services/Requests.py
```diff
@@ -201,9 +202,7 @@ def makeRequest_httplib(self, uri, data, verb, headers):
         if verb == 'GET' and data:
             uri = "%s?%s" % (uri, data)
 
-        # PY3 needed for compatibility because str under futurize is not a string. Can be just str in Py3 only
-        # PY3 Don't let futurize change this
-        assert isinstance(data, (str, basestring)), \
+        assert isinstance(data, (str,bytes)), \
```
amaltaro, line 205:
```suggestion
        assert isinstance(data, (str, bytes)), \
```
extra space

amaltaro (changes requested):
@mapellidario Dario, I have made a few other suggestions in the code.
Besides that, can you please clarify how the `install_aliases` gets added to these modules? Was it done manually or using futurize?

src/python/WMCore/Cache/WMConfigCache.py
mapellidario: You are definitely right, thank you very much!

src/python/WMCore/Cache/WMConfigCache.py
mapellidario: This is definitely more clear, thanks!

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
amaltaro: Thanks!

amaltaro (commented):
Dario, changes look good to me. However, I've made a couple of comments which I think might need attention.
Please also squash these commits (keeping test/* in a separate commit).

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
```diff
@@ -1,20 +1,11 @@
 from __future__ import print_function
 
+from builtins import str, bytes, int
```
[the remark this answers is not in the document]
mapellidario: You are correct. I do not know how it is that the change got lost, whether futurize did not add ir or whether I lost it in a refactoring. I will investigate later.
Anyway, It has an effect only if we use `__next__ / next` or `__str__`, which does not seem to be the case for the `JSONThunker` class.

more details on the backported `newobject` at: https://github.com/PythonCharmers/python-future/blob/master/src/future/types/newobject.py

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -38,6 +38,12 @@
 """
 from __future__ import print_function
 
+from future import standard_library
+standard_library.install_aliases()
+
+from builtins import str, range, object
+from past.builtins import basestring
```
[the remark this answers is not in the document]
mapellidario: Yes, you are correct. Discarding the use of `basestring` is something that we may want to do. 
However, at this early stage I preferred simply importing `from past.builtins import basestring
` to reduce the number of changes needed to run in a py3 environment. I would leave this to a future iteration, but only if it is ok with you.

mapellidario:
test this please

src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py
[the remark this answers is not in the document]
mapellidario: For future reference: adding `from builtins import object` here breaks about 170 unit tests. To be investigated at a later moment

amaltaro:
@mapellidario as you've foreseen, there are conflicts here that will need to be manually resolved. Let me know once things are good to go.

amaltaro (approved):

PR #9833, opened 2020-07-11
Revert "Bundle patch for WMAgent 1.3.6.patch1"

amaltaro:
Reverts dmwm/WMCore#9812

Actually it's not going to revert the whole PR, but only the PY3 related changes, so:
[py2py3] futurize on WMCore/src/python/PSetTweaks #9740
[py2py3] futurize on WMCore/src/python/Utils #9731

the other two fixes should remain in the branch.

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -7,19 +7,14 @@ independent python structure
 
 """
 
-from future import standard_library
-standard_library.install_aliases()
-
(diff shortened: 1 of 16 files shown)


amaltaro:
[Fix #9833 - Revert "Bundle patch for WMAgent 1.3.6.patch1"](https://trello.com/c/n0XPPeq4/213-fix-9833-revert-bundle-patch-for-wmagent-136patch1)

PR #9848, opened 2020-07-15
[py2py3] WMQuality.TestInit - Remove dependency on MySQL-python

### LINKED ISSUE #9805: Remove/replace MySQL-python library by something else that supports python3
**Impact of the new feature**
WMCore in general (and perhaps other services depending on it)

Yes, this package `MySQL-python` does not support python3. Actually, its last commit dates back from 2014...

Either completely deprecate this package from the COMP cmsdist repo:
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/py2-mysqldb.spec

and update all the other specs depending on it. Or if this functionality is needed in WMCore (or other DMWM services), we need to find a library that supports python3 (and probably works still in python2), like `PyMySQL`:
https://pypi.org/project/PyMySQL/

There is no alternative.

**Additional context**
Issue initially created here: https://github.com/cms-sw/cmsdist/issues/6025

mapellidario:
#### Status
ready

#### Description
This is related to #9805

We want not to depend anymore on MySQL-python, since it does
not support python3.

This is the only occurrence of this package left in dmwm/WMCore

We used this library only to import a MySQL specific exception. The specific exception is being replaced by a more generic one.

YES

This make dmwm/WMCore not dependent on `MySQL-python`

### DIFF
--- a/src/python/WMQuality/TestInit.py
+++ b/src/python/WMQuality/TestInit.py
@@ -20,8 +20,6 @@ import tempfile
 import threading
 import traceback
 
-from _mysql_exceptions import OperationalError
-
 from WMCore.Agent.Configuration import Configuration
 from WMCore.Agent.Configuration import loadConfigurationFile
 from WMCore.WMException import WMException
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
Thanks Dario!

PR #9855, opened 2020-07-17
Fix HEAD request via pycurl_manager; added a few extra getters

### LINKED ISSUE #9854: HEAD requests do not work with the pycurl_manager module
**Impact of the bug**
WMCore in general

**Describe the bug**
I don't remember the exact problem now, but either it fails or it does not return any information. The problem is with the positional arguments.

**How to reproduce it**
Make a HEAD request through pycurl_manager module

**Expected behavior**
HEAD requests should return only the response HEADER; and it should work!

none

amaltaro:
Fixes #9854 

#### Status
ready

#### Description
In short:
* fix the pycurl_manager HEAD requests
* added a few getter methods to help parsing the response headers

yes

#### Related PRs
Taking parts of the big work implemented here: https://github.com/dmwm/WMCore/pull/9759

none

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -90,6 +90,30 @@ class ResponseHeader(object):
             except:
                 pass
 
+    def getReason(self):
+        """
+        Return the HTTP request reason
+        """
+        return self.reason
+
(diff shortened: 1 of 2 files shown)


todor-ivanov (approved):
looks good to me

PR #9863, opened 2020-08-11
Script to print a summary of LQEs stuck in Available for a given workflow

### LINKED ISSUE #8724: Tools to monitor stuck workflows
1. monitor given workflow is in the correct status.
   - Help to  debug figure out jobs are not submitted due the the lower priority or some problem.
   - Help to debug figure out when jobs are not pulled from global queue 

report how many jobs are in the queue with higher priority by site.

amaltaro:
Partially fixes #8724

#### Status
ready

#### Description
When completing the agent draining procedure, it's quite common to find a few remaining workflows stuck in the agent. Usually those get stuck because DataLocationMapper updates the data location of the input data, causing the workflow (or work unit) to no longer have a common SiteWhitelist + data locations.

This script - to be ran against a given workflow name - will print a summary of the `Available` local workqueue elements and will suggest a course of action. The output should be provided to the PnR team, who will need to act on that workflow.

yes

#### Related PRs
none

none

### DIFF
--- a/bin/adhoc-scripts/checkStuckLQE.py
+++ b/bin/adhoc-scripts/checkStuckLQE.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Script to be executed with the WMAgent environment, does you first need to execute:
+source apps/wmagent/etc/profile.d/init.sh
+
+Provided a workflow name in the command line, it will find all the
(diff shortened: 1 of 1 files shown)


amaltaro:
FYI @todor-ivanov , I think this is one of the items we would like Tsveta to work on btw.

todor-ivanov (approved):
Looks like it will be quite useful. Thanks!

PR #9865, opened 2020-08-11
Update ParentageResolved flag for StepChains without parent as well

### LINKED ISSUE #9864: ParentageResolved flag needs to be updated for workflows with no parents too
**Impact of the bug**
ReqMgr2

**Describe the bug**
As reported in this JIRA:
https://its.cern.ch/jira/browse/CMSCOMPPR-16710

there are workflows that are not marked as parentage resolved, thus blocking Unified from announcing workflows.

**How to reproduce it**
Run StepChain workflow producing only one dataset from scratch, and move it to closed-out.

**Expected behavior**
The ReqMgr2 cherrypy thread should go through all the StepChain workflows and mark them with:
```
ParentageResolved = true
```
when there is no parentage information to be updated, besides the normal operation of the thread.

Example of workflow producing - from scratch - only a MINIAODSIM dataset, thus no parentage information needs to be updated:
https://cmsweb.cern.ch/reqmgr2/fetch?rid=cmsunified_task_EXO-RunIIFall18wmLHEGS-02833__v1_T_200220_091412_4231

amaltaro:
Fixes #9864 

#### Status
ready

#### Description
Make sure every single StepChain workflow gets updated once they reach a status beyond `closed-out`, even if their output datasets do not have a parent dataset.

With this change, the number of DBS queries will increase (by a few hundred a day), because it has to ask DBS for the parent dataset. It will then print a warning like:
```
2020-08-11 15:51:17,986:WARNING:DBS3Reader:No parent dataset found for child dataset /TWZToLL_thad_Wlept_5f_DR_TuneCP5_PSweights_13TeV-amcatnlo-pythia8/RunIIFall17MiniAODv2-PU2017_12Apr2018_94X_mc2017_realistic_v14-v2/MINIAODSIM
```
and move on with the request spec update.

yes

#### Related PRs
none

This needs to get deployed in CMSWEB.

### DIFF
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/StepChainParentageFixTask.py
@@ -21,9 +21,8 @@ def getChildDatasetsForStepChainMissingParent(reqmgrDB, status):
     for reqName, info in results.items():
 
         for dsInfo in info.values():
-            if dsInfo["ParentDset"]:
-                for childDS in dsInfo["ChildDsets"]:
-                    requestsByChildDataset[childDS].add(reqName)
+            for childDS in dsInfo["ChildDsets"]:
+                requestsByChildDataset[childDS].add(reqName)
     return requestsByChildDataset
 
 


todor-ivanov (approved):
Looks good to me

PR #9871, opened 2020-08-14
PY3: Properly deal with strings in REST/Main

### LINKED ISSUE #9794: MSOutput crashes initializing Rucio
**Impact of the bug**
ReqMgr2MS - MSOutput

**Describe the bug**
Once updated the ms-output configuration to:
```
data.defaultDataManSys = "Rucio"
```

the ms-output service fails to get started. Traceback will be provided below.

**How to reproduce it**
Deploy reqmgr2ms with tag HG2007h and start MSOutput.

**Expected behavior**
The service should start without any problems, nothing else ;)

Traceback from ms-output logs in vocms0731
```
...
2020-07-03 13:35:32,866:WARNING:MongoDB: Collection '('msOutRelValColl', <pymongo.operations.IndexModel object at 0x7feb0e6674b0>)' Already exists in database 'msOutDB'
2020-07-03 13:35:32,868:WARNING:MongoDB: Collection '('msOutNonRelValColl', <pymongo.operations.IndexModel object at 0x7feb0e6674b0>)' Already exists in database 'msOutDB'
2020-07-03 13:35:32,868:INFO:Rucio: WMCore Rucio initialization with acct: wma_test, host: None, auth: None and these extra parameters: {'auth_type': None, 'account': 'wma_test', 'timeout': 600, 'ca_cert': None, 'rucio_host': None, 'auth_
host': None, 'creds': None, 'user_agent': 'wmcore-client'}
2020-07-03 13:35:32,869:DEBUG:baseclient: no auth_type passed. Trying to get it from the environment variable RUCIO_AUTH_TYPE and config file.
2020-07-03 13:35:32,869:DEBUG:baseclient: no creds passed. Trying to get it from the config file.
2020-07-03 13:35:32,870:DEBUG:baseclient: no ca_cert passed. Trying to get it from the config file.
2020-07-03 13:35:32,870:DEBUG:baseclient: get a new token
2020-07-03 13:35:32,874:DEBUG:connectionpool: Starting new HTTPS connection (1): cmsrucio-auth-int.cern.ch:443
2020-07-03 13:35:32,988:DEBUG:connectionpool: https://cmsrucio-auth-int.cern.ch:443 "GET /auth/x509 HTTP/1.1" 401 126
  File "/data/srv/beHG2007h/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.6.pre6/bin/wmc-httpd", line 3, in <module>
    main()
  File "/data/srv/beHG2007h/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.6.pre6/lib/python2.7/site-packages/WMCore/REST/Main.py", line 596, in main
    server.start_daemon()
  File "/data/srv/beHG2007h/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.2.6.pre6/lib/python2.7/site-packages/WMCore/REST/Main.py", line 426, in start_daemon
    traceback.print_exc(file=trace)
  File "/data/srv/beHG2007h/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/traceback.py", line 233, in print_exc
    print_exception(etype, value, tb, limit, file)
  File "/data/srv/beHG2007h/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/traceback.py", line 124, in print_exception
    _print(file, 'Traceback (most recent call last):')
  File "/data/srv/beHG2007h/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/traceback.py", line 13, in _print
    file.write(str+terminator)
TypeError: unicode argument expected, got 'str'
[03/Jul/2020:13:35:33]  WATCHDOG: server exited with exit code 1
```

amaltaro:
Fixes #9794 

#### Status
ready

#### Description
I'm still debugging the rucio initialization problem, but there seems to be no issue.
This PR fixes the unicode / string issue reported in the traceback. The traceback is a python2 string, so we need to use BytesIO instead of StringIO.

yes

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/REST/Main.py
+++ b/src/python/WMCore/REST/Main.py
@@ -20,7 +20,7 @@ import thread
 import time
 import traceback
 from argparse import ArgumentParser
-from io import StringIO
+from io import BytesIO, StringIO
 from glob import glob
 from subprocess import Popen, PIPE
 from pprint import pformat
@@ -422,7 +422,10 @@ class RESTDaemon(RESTMain):
             self.run()
         except Exception as e:
             error = True
-            trace = StringIO()
+            if sys.version_info[0] == 2:
+                trace = BytesIO()
+            else:
+                trace = StringIO()
             traceback.print_exc(file=trace)
             cherrypy.log("ERROR: terminating due to error: %s" % trace.getvalue())
 


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

src/python/WMCore/REST/Main.py
```diff
@@ -422,7 +422,7 @@ def start_daemon(self):
             self.run()
         except Exception as e:
             error = True
-            trace = StringIO()
+            trace = BytesIO()
```
mapellidario, line 425:
```suggestion
            import sys
            [...]
            if sys.version[0] == '2':
                trace = BytesIO()
            elif sys.version[0] == '3':
                tracce = StringIO()
```

mapellidario (changes requested):
Dear Alan, the change that you propose would work in python2 only, it would brake in python3.

I did some tests with`traceback` and `io` at

* py2: https://cms-dmwm-test.docs.cern.ch/py2py3/porting/traceback_py2/
* py3: https://cms-dmwm-test.docs.cern.ch/py2py3/porting/traceback_py3/

Why is this the case? Prerequisites

1. `StringIO.StringIO` constructor accepts both unicode and bytes strings. Available only in py2
1. `io.StringIO` constructor accepts _only_ unicode strings in both py2 and py3 (backported in py2 from py3. This is why "String" in `StringIO` (unicode sequence) differs from the type `str` (bytes sequence))
1. `io.BytesIO` constructor accepts _only_ bytes strings in both py2 and py3 (backported in py2 from py3)
1. `traceback.print_exc(file=trace)` in py2 internally does directly `trace.write(stacktrace)`, where `stacktrace` is a `str`, which in py2 is a sequence of bytes.
1. `traceback.print_exc(file=trace)` in py3 internally writes to the file with `print(line, file=trace)`. python3 [print](https://docs.python.org/3/library/functions.html#print) converts all the inputs to `str` before writing to the buffer, and in python3 `str` is a sequence of unicode codepoints.

What can be done:
* py2: use `trace = StringIO.StringIO()`
* py2: use `trace = io.BytesIO()`
* py3: use `trace = io.StringIO()`

Some more tests with `io.StringIO` and `io.BytesIO`

* https://cms-dmwm-test.docs.cern.ch/py2py3/porting/strings_io_py2/
* https://cms-dmwm-test.docs.cern.ch/py2py3/porting/strings_io_py3/

amaltaro:
Thank you very much for this deep review, Dario.
An alternative to your suggestion, how about we import different classes but with the same name (such that all the changes are made at import level instead of in the middle of the code)? Something like:
```
if sys.version[0] == '2':
                from io import BytesIO as StringIO
else:
                from io import StringIO
```

mapellidario:
Yes, that would work, too. 

However, we can consider this an exercise for the gradual migration of WMCore. What if we have module A.py with ` from io import BytesIO as StringIO`, and module B.py that imports A.py but require `from io import StringIO` ? Renaming modules is fine if we have isolated modules, but I fear that in the long run it could cause problems. Moreover, since during the migration we will have to deal with such errors a lot, I would prefer to be 100% sure of the type of a variable reading only the line where it is defined. I fear that if we rename modules, then I will make some mess since I forget to check at the beginning of the file if an imported module has been renamed. I know that the expense is less beautiful code, but in this case I would trade less beauty for more clarity. 

This being said, code style is a matter of opinions, and your opinion always has the priority

amaltaro:
I did some local tests and it would only be a problem if we had (not recommended!) imports like:
```
from A import *
```

then those modules would be at the main scope of the imported modules.
While, if one does imports like
```
import A
import B
```
this would be completely fine and `StringIO` (or the masked BytesIO) would be in the scope of the module A (or module B).

The motivation I had to have this py2/py3 logic in the imports section is that when deprecating py2, it would be mostly a matter of imports, without touching the internals of the code.
However, I do agree with you that - if we have problems due to this masked importing - things will be much harder to debug. I'm going to apply your suggestion soon. Thanks again Dario!

amaltaro:
@mapellidario I have updated this PR following your suggestion, could you please review it once again?

In addition to this, I guess we should revisit all the usage of BytesIO and implement the same logic we have in this PR. Would you see any possible exception to that?

mapellidario (approved):
This looks perfect to me!
I know that this may look like a bit cumbersome, but it the best that we can do while maintaining py2 and py3 correctness and code clarity. 
Thanks Alan!

mapellidario:
Well, I would be a bit more conservative and would not jump into applying the same logic everywhere. The solution that we found for this case depends on the interface of `traceback`, which maintained `str` in both py2 and py3 despite its change of meaning. A different library could choose a different approach and the only way to be sure is to check it case by case. 
Since dealing with strings is the most difficult and troublesome part of the migration, I would leave every aspect of it to the gradual migration. If it were not for your open PR, I would not have suggested any change to this module right now.

amaltaro:
Sounds good. Thanks for your feedback, Dario.
This has been tested on my VM (py2, of course) and it works fine.

Issue #9849, opened 2020-07-15
Use setuptools instead of distutils for PyPI builds


goughes:
**Impact of the new feature**
We use distutils to build pypi things currently. Setuptools is a more current approach for this.
Python packaging guide recommends setuptools: https://packaging.python.org/guides/tool-recommendations/

**Is your feature request related to a problem? Please describe.**
No problem, but we should base our pypi things off of the current accepted standard for packaging.
Distutils will go away at some point: https://bugs.python.org/issue41282

**Describe the solution you'd like**
Update all setup scripts to use setuptools instead of distutils.

**Describe alternatives you've considered**
None

amaltaro:
Closed by https://github.com/dmwm/WMCore/pull/9874
Erik, please reopen it if I'm wrong.

Issue #9822, opened 2020-07-09
Create an organized structure for PyPI packaging 


goughes:
**Impact of the new feature**
This will help arrange the PyPI components of WMCore, allowing us to build the required packages without making a mess of the repo with multiple requirements files and setup.py scripts.

**Is your feature request related to a problem? Please describe.**
We need to package at least 6-7 things, which could lead to lots of duplication of files. Let's make sure we do this in a sane way.
Currently there are two setup scripts for wmcore and wmagent, which have the same content except for one line. 

**Describe the solution you'd like**
Be able to build and upload all the required packages to PyPI with a minimal footprint in the WMCore repo.

amaltaro:
Closed by https://github.com/dmwm/WMCore/pull/9874
Erik, please reopen this issue if I'm wrong.

PR #9874, opened 2020-08-18
PyPI packaging for ReqMgr2

### LINKED ISSUE #8587: Migrate ReqMgr2 to pypi
As Brian suggested I'm opening up a ticket to start campaign of migration DMWM tools to pypi. This one goes to ReqMgr2. And we need similar to other sub-systems of WMCore.

There are several steps which need to be accomplished:
- write proper setup.py
- write requirements.txt file for app in question
- test/deploy app via pip install -r requirements.txt on local node, e.g. `pip install -r requirements.txt`, verify that dependencies works with app code, if necessary adjust app code to align with newest versions of dependencies
- upload app to pypi

The [requirements](https://pip.readthedocs.io/en/1.1/requirements.html) file can be composed from a list of requirements of app spec files, see [reqmgr2.spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/reqmgr2.spec#L11).

goughes:
Fixes #8587 #9849 #9822 

#### Status
In development

#### Description
PyPI updates to create an organized structure to build many packages from the WMCore repo. Uses a single setup_template.py to create the necessary setup.py build scripts. Uses a single requirements.txt file that is parsed to create per package requirements files for build and upload. These files are manipulated by tools/build_pypi_packages.sh.

As an example of the new method, ReqMgr2 can be packaged and uploaded to PyPI.

Also, does not use distutils. 

YES

#### Related PRs
Other PRs here: https://github.com/dmwm/WMCore/projects/8

None


setup_template.py
```diff
@@ -43,19 +56,18 @@ def parse_requirements(requirements_file):
...
-                              'src/python/WMComponent',
-                              'src/python/WMQuality',
-                              'src/python/PSetTweaks']),
-      data_files=list_static_files(),
+      packages=packages,
+      py_modules=py_modules,
+      data_files=data_files,
       install_requires=parse_requirements(requirements),
+      maintainer='CMS DMWM Group',
+      maintainer_email='hn-cms-dmDevelopment@cern.ch',
```
amaltaro, line 67: I guess this address could be updated to:  hn-cms-wmdevelopment

amaltaro (commented):
Erik, it wasn't a thorough review, but changes are looking good.
I wonder if the requirements.txt file is prepared to multiple WMCore services, or is it only ready for either wmcore/wmagent/reqmgr2 at the moment?

goughes:
Currently this only supports wmcore, wmagent and reqmgr2. If you think this process for building packages looks good, I can update the requirements.txt file with the other services as well. The build script would need an update as well to be more streamlined.

In my testing, this method produces packages with similar contents for packages, modules, and data files. In some cases, there are additional files in the bin directory (regmgr2 for instance), and I have not figured out how they get there. Also, there are some init scripts that are missing from the etc dir that get adding during the RPM build process. These might need to change slightly for the pypi deployment procedure.

amaltaro:
Here's how those files show up under the bin directory (and a few other things done during the RPM build):
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/reqmgr2.spec#L22-L39

Could you please point me again to the procedure - considering these changes - to build and upload a pypi package? Just so I can see the full picture here.

goughes:
The procedure is the same as before. https://github.com/dmwm/WMCore/wiki/DMWM-Packaging-with-PyPi

Reqmgr2 is on PyPI now if you want to install to see how the directory structure looks in a virtual environment: https://pypi.org/project/reqmgr2

goughes:
I worked out the issue with including all files in the bin dir. The environment scripts that are placed in etc/profile.d in the RPMs should be handled during the creation of the Docker image for each service. The current method for creating the env scripts is specific to a cmsweb release, which is not known when we package and upload to pypi.

amaltaro (approved):
Changes look good to me Erik and I see no problems with this procedure.

I have two comments though, with this PR, we no longer need the `.pypirc` file, right? If so, could you please update this documentation:
https://github.com/dmwm/WMCore/wiki/DMWM-Packaging-with-PyPi

Second comment was regarding the WMCore scripts under the `bin` directory, when installed via pip. I see it only has 3 or 4 wmcore scripts, while our RPM-based installation has a bunch of them. I *think* all the scripts required to run reqmgr2 are already in place though, just something to keep in mind if we fail to run the service in the near future.

amaltaro:
Is it ready to be merged? Or you still want to get something else in?

PR #9881, opened 2020-08-26
MS: Track Rucio rule progress in MSMonitor

### LINKED ISSUE #9851: MSMonitor: distinguish between PhEDEx and Rucio data placement
**Impact of the new feature**
ReqMgr2MS (MSMonitor)

There will be a moment that MSMonitor will have to update the status of input data transfers that have been made against PhEDEx, and others against Rucio. However, a given workflow will always be under the same DM system.

We need to implement some logic in MSMonitor to properly distinguish a PhEDEx transfer request from a Rucio rule.
I initially thought that we could actually update the transfer document schema, with an extra key/value pair but that's is likely too much - and it would only be needed for some short period of time - so we can likely check the data type (or how long the transfer ID is). PhEDEx has an integer request ID, while Rucio has a hash string. 

none

**Additional context**
none

amaltaro:
Fixes #9851 

#### Status
ready

#### Description
This PR provides the ability to track Rucio rule progress. The logic for the rucio rule is basically:
* if state == OK, then the rule progress is 100% (all files are where they are meant to be)
* else, progress depends on locks count such as: `locks_ok / total_files_to_be_locked`

In addition to that, it brings:
* skip workflow from the current cycle if the DM system request failed or if an invalid response was received (workflow progress update is retried in the next cycle)
* some logging improvements

yes

#### Related PRs
none

Yes, the following deployment/configuration changes:
https://github.com/dmwm/deployment/pull/939
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/21
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/22

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSMonitor.py
+++ b/src/python/WMCore/MicroService/Unified/MSMonitor.py
@@ -14,6 +14,7 @@ import time
 # WMCore modules
 from WMCore.MicroService.DataStructs.DefaultStructs import MONITOR_REPORT
 from WMCore.MicroService.Unified.MSCore import MSCore
+from WMCore.Services.Rucio.Rucio import Rucio
 
 
 class MSMonitor(MSCore):
@@ -27,6 +28,10 @@ class MSMonitor(MSCore):
         # update interval is used to check records in CouchDB and update them
         # after this interval, default 6h
         self.updateInterval = self.msConfig.get('updateInterval', 6 * 60 * 60)
+        self.rucio = Rucio(acct=self.msConfig['rucioAccount'],
+                           hostUrl=self.msConfig['rucioUrl'],
+                           authUrl=self.msConfig['rucioAuthUrl'],
+                           configDict={"logger": self.logger, "user_agent": "WMCore-MSMonitor"})
 
     def updateCaches(self):
         """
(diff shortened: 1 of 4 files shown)


src/python/WMCore/MicroService/Unified/MSMonitor.py
```diff
@@ -206,6 +216,47 @@ def _getTransferstatus(self, dataset, requestList):
...
+            # if we query by dataset and the subscription was at block level,
+            # we get an empty response. So always wildcard the block parameter
+            data = self.rucio.getRule(reqId)
+            if not data:
+                # failed to retrieve data for this rule id,
+                # TODO: should we raise an exception and work on it in the next cycle?
+                completion.append(0)
+                continue
+
+            if data['state'] == "OK":
```
amaltaro, line 248: @ericvaandering Eric, could you please confirm whether this is all we need to track the rucio rule progress?
In short:
state=OK means the rule is 100% completed
otherwise locks_ok / sum(all the other 3 locks) gives me the progress of this yet incomplete rule.

Does it cover it properly?
Thanks!

ericvaandering: Yes. Still looking. Can you change reqID to ruleID? Requests are another thing in Rucio and I'd rather not have anyone get confused. I presume request list stays as is since it can be PhEDEx or Rucio request.

amaltaro: Sure, I'm updating this. Eventually I will stop mixing up those words with different meanings ;)

amaltaro:
Basic tests have been performed with both PhEDEx and Rucio and from my side it's ready to go.
Eric, if you feel like having another look into this PR, please go for it!

amaltaro:
Basic tests have been done here.

PR #9884, opened 2020-08-28
Rucio pick site

### LINKED ISSUE #9841: MSOutput: decide which custodial site to use for data placement
**Impact of the new feature**
MSOutput

It looks like the we have overlooked the custodial output data placement. Custodial means that data is subscribed to Tape/MSS storage endpoints.

Right now, the output data placement in MicroServices does a data placement against any disk RSE (either DDM or Rucio will decide what the final location will be). So we need to change it to actually perform a custodial data placement (where MS needs to decide the final Tape location for DDM; with luck, Rucio can define it itself).

We need to go through the checkor code:
https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/checkor.py#L1009

and make/extract an algorithm for how the custodial site is defined within Unified. Once we know how it's done, we can either implement it in MSOutput or create a better way of identifying the final location.

Note that the Unified configuration is also used to define what goes or not to Tape (and maybe a few other parameters).

In short, this issue requires three actions:
1. study how custodial location is selected by Unified - AND make a clear specification in this issue
2. implement those requirements in the MSOutput 
3. evaluate whether we really need to have standard non-custodial output data placement done by MSOutput (given that workflows are already assigned with NonCustodial parameters, where WMAgent takes care of the output data placement).

Given that WMAgents *can* perform the output data placement - if workflows are assigned as such - it might be that the current MSOutput implement could actually be dropped, and no nonCustodial data placement needs to be performed.

**Additional context**
none

ericvaandering:
Fixes #9841 (or part of it)

#### Status
Ready

#### Description
Pick a custodial site. This does work against production which has quota values set on Disk sites. It can also work anywhere with equal weights when setting rseAttribute=None

YES

#### Related PRs
None

No. When transition to python3 is done, a function can be remove

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -8,15 +8,15 @@ CMS Workload Management system (and migration from PhEDEx).
 from __future__ import division, print_function, absolute_import
 
 import logging
+import random
 from copy import deepcopy
 from pprint import pformat
-
 from rucio.client import Client
 from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied, DuplicateRule,
                                     DataIdentifierAlreadyExists, DuplicateContent,
                                     UnsupportedOperation, FileAlreadyExists, RuleNotFound)
-from WMCore.WMException import WMException
 
+from WMCore.WMException import WMException
 
 RUCIO_VALID_PROJECT = ("Production", "RelVal", "Tier0", "Test", "User")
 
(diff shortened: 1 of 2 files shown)


amaltaro:
@ericvaandering Eric, could you please provide some basic unit test to test and demostrate its functionality? If you don't manage to test things yourself, feel free to provide the code and I can test it for you (or just push it in and wait for jenkins).

Another question I had was whether both the integration and the production Rucio servers contain the same RSE attributes? Or at least the default `ddm_quota`, which will be extremely important to develop/test things against the int server.

ericvaandering:
Alan, here is a minimal unit test. That should do it.

amaltaro (approved):

amaltaro:
Thanks Eric!

PR #9889, opened 2020-09-03
Rucio wrapper API to find where data is locked and available

### LINKED ISSUE #9888: Workqueue data location has to use only data locked by MSTransferor
**Impact of the new feature**
Global and local workqueue

The current global workqueue data location logic - for Rucio - relies solely on the block replicas available, which means that might be locked by anyone/anything else, thus data can vanish at any moment.

Now that we have the correct workflow in place for MSTransferor, we should adopt a similar logic in global workqueue for the input data location, so the idea is to list replicas (blocks/containers) locked by our own system/MSTransferor, thus checking for rules locking data under our `wmcore_transferor` account. Use only those locks as data location.

Not an alternative, but just something that **must** be considered during the transition. Once we switch WMCore to Rucio for the input data placement, there will be thousands of WQE already created under PhEDEx subscriptions. Which means data location based on Rucio will not be able to find any locks under `wmcore_transferor`, thus setting that WQE location as empty.
This use case needs to be properly discussed with the experts. Perhaps we could turn all the PhEDEx subscriptions under DataOps into Rucio rules under `wmcore_transferor` account?

**Additional context**
Initial integration of Rucio into global/local workqueue has been done here:
https://github.com/dmwm/WMCore/pull/9389

### LINKED ISSUE #9895: MSTransferor: input data placement for Harvesting workflows need to be done at container level
**Impact of the new feature**
MSTransferor

If we make the data placement at the block-level, global workqueue fails to find a replication rule for the input dataset (container-level), which means the workqueue element would have the input data location marked as an empty list.

Given that Harvesting workflows require the whole input dataset to be available at one given location (workqueue element has a container granularity, thus single element), we should actually create Rucio rules at the container level, instead of placing rules for all the input dataset blocks.

An alternative would be to, look for replication rules for all the blocks of a container, in addition to the replication rules for the container itself.

**Additional context**
Some logs for a block-level data placement through Rucio:
```
2020-09-07 19:28:07,262:INFO:DataLocationMapper:Fetching location from Rucio...
2020-09-07 19:28:07,262:INFO:DataLocationMapper:AMR fetching data location for: /NoBPTX/Run2016F-23Sep2016-v1/DQMIO
2020-09-07 19:28:07,263:INFO:Rucio:AMR kwargs : {'account': 'wmcore_transferor',
 'name': u'/NoBPTX/Run2016F-23Sep2016-v1/DQMIO',
 'scope': 'cms'}
2020-09-07 19:28:07,291:INFO:Rucio:Data location for /NoBPTX/Run2016F-23Sep2016-v1/DQMIO from single RSE locks and available at: []
2020-09-07 19:28:07,292:INFO:DataLocationMapper:/NoBPTX/Run2016F-23Sep2016-v1/DQMIO: []
```

when all the blocks are already available at the final destination, even though we still see an empty list here...

amaltaro:
Fixes #9888
Fixes #9895

#### Status
ready

#### Description

Summary of changes is:
* moved the `rucioAccount` attribute from config.General to config.WorkQueueManager.queueParams
* updated agent rucio account to: `wmcore_transferor` (RucioInjector keeps using the wma_* accts)
* DQMHarvest workflows will have the whole input dataset placed under the same destination (at dataset/container level)
* Rucio wrapper API to check whether a DID is a container or not
* Rucio wrapper API to perform the data location and locking (tightly coupled to rules created by MSTransferor)
* updated WorkQueue DataLocationMapper thread to properly use the Rucio object for workqueue elements data location
* update StartPolicyInterface to properly build container-level location with Rucio

Description of the 3 most important use cases is:
**Use case 1:** my workqueue element contains an input block, I need to find out in which RSEs this block is currently available, AND which RSEs have this block locked by my MSTransferor rucio account.
**Use case 2:** my workqueue element contains a pileup dataset, so I need to find all RSEs hosting at least one block of the pileup, AND which RSEs have this block locked by my MSTransferor rucio account.
**Use case 3:** my workqueue element contains a DQM/DQMIO dataset (container), I need to find which RSE contains the whole container (all blocks), AND which RSEs have this container locked by MSTransferor rucio account.
  * for DQM/DQMIO, we could change how the rule is created and make rules on the container DID, instead of passing all the blocks DID to the rule

The logic implemented is as follows (using the `getDataLockedAndAvailable` method):
**For a block:**
1. list all its replication rules, provided a DID + SCOPE + GROUPING + ACCOUNT
2. single RSE rules (state=OK), are considered a valid location
3. for multi RSE rules, we keep the rule_id in a list and check `get_dataset_locks` output, data AVAILABLE under one of those rule IDs is considered as a valid location
4. **FIXME**: should we get the parent of this block and check for container level rules?

**For a container:**
1. same steps 1) and 2) above
2. if there are multi RSE rules, call `_getContainerLockedAndAvailable`. Which discover all the block names in that container
3. then for each block in the container, perform step 3) above
4. depending on the grouping required, deal different with the list of RSEs for each block
   1. ALL: we want to have the whole dataset in the same location (like Harvesting workflows, DQM/DQMIO containers)
   2. DATASET: we only want to know which RSEs contain at least one block of the container (workflows reading secondary data)

yes

#### Related PRs
none

none

### DIFF
--- a/bin/wmagent-mod-config
+++ b/bin/wmagent-mod-config
@@ -111,8 +111,6 @@ def modifyConfiguration(config, **args):
     if hasattr(config, "General"):
         config.General.central_logdb_url = args["central_logdb_url"]
         config.General.centralWMStatsURL = args["wmstats_url"]
-        # Both PhEDExInjector and WorkQueueManager use it, so we better make it general
-        config.General.rucioAccount = args["rucio_account"]
         # t0 agent doesn't have reqmgr2_url
         if "reqmgr2_url" in args:
             config.General.ReqMgr2ServiceURL = args["reqmgr2_url"]
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -88,7 +88,6 @@ config.General.logdb_name = logDBName
 config.General.central_logdb_url = "need to get from secrets file"
 config.General.ReqMgr2ServiceURL = "ReqMgr2 rest service"
 config.General.centralWMStatsURL = "Central WMStats URL"
-config.General.rucioAccount = "OVERWRITE_BY_SECRETS"
 
 config.section_("JobStateMachine")
 config.JobStateMachine.couchurl = couchURL
@@ -135,6 +134,8 @@ config.WorkQueueManager.queueParams["ParentQueueCouchUrl"] = "https://cmsweb.cer
 config.WorkQueueManager.queueParams["QueueURL"] = "http://%s:5984" % (config.Agent.hostName)
 config.WorkQueueManager.queueParams["WorkPerCycle"] = 200  # don't pull more than this number of elements per cycle
 config.WorkQueueManager.queueParams["QueueDepth"] = 0.5  # pull work from GQ for only half of the resources
+config.WorkQueueManager.queueParams["rucioAccount"] = "wmcore_transferor"  # account for data locks
+
 
 config.component_("DBS3Upload")
 config.DBS3Upload.namespace = "WMComponent.DBS3Buffer.DBS3Upload"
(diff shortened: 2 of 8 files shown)


amaltaro:
bulk of the development and tests has been done. Further very very careful look and test is probably a good idea, but it should be ready to go.

PR #9897, opened 2020-09-07
Fix Rucio rule creation when there are duplicate rules

### LINKED ISSUE #9621: Properly deal with duplicated Rucio rules
**Impact of the new feature**
WMAgent

When making the final output dataset subscription - as we currently do for phedex - we need to have a mechanism in place to figure out when the cms dataset (container) already has such rule created by a different agent (for cases where the same workflow is dealt with by multiple agents)

In case we start persisting the rule id in the database, we need to make another rucio call to retrieve the rules for a given DID. Then we need to compare all the rules available against:
* rucio account
* DID
* RSE expression
such that we can find - or not - a pre-existent rule.

IF we do not persist the rule id, then we could simply catch the duplicate rule exception `DuplicateRule` and return a positive response, such that the component can mark that as done and move on.

alternative described above

**Additional context**
none

amaltaro:
Fixes #9621 

#### Status
ready

#### Description
Use `list_did_rules` instead of `list_replication_rule_full_history` API to look for rules locking a given DID (I believe the latter lists also rules long deleted). 
In addition to that, compare DID + Scope + Account + RSE expression to find what the duplicate rule IDs are.

yes

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -10,7 +10,6 @@ from __future__ import division, print_function, absolute_import
 import logging
 import random
 from copy import deepcopy
-from pprint import pformat
 from rucio.client import Client
 from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied, DuplicateRule,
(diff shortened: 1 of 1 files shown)


amaltaro:
tests went fine, merging.

PR #9900, opened 2020-09-08
Get block replicas info using the Rucio bulk API

### LINKED ISSUE #9893: Use Rucio list_dataset_replicas with a list of DIDs instead of single DID calls
**Impact of the new feature**
WMCore in general

Now that Rucio supports listing dataset replicas in bulk, as reported in this issue:
https://github.com/rucio/rucio/issues/2459

made available through this client API `list_dataset_replicas_bulk`.
We should update the Rucio wrapper `getReplicaInfoForBlocks` method and make HTTP calls in bulk, instead of making single block HTTP calls.

Once we have a list of blocks to list their replicas, use the `list_dataset_replicas_bulk` client API to retrieve replicas for all of them with a single HTTP call.
This change should be transparent, but we better tag Stefano once a solution gets implemented.

not touch anything and keep hitting the Rucio server with many "unneeded" calls.

**Additional context**
none

amaltaro:
Fixes #9893 

#### Status
tested

#### Description
Summary of changes is:
* `getReplicaInfoForBlocks` method no longer supports the `deep` parameter (even though nothing will fail)
* when fetching blocks in a container, there is an extra call to check whether the input DID is a container or not
  * and this `isContainer` method now raises a `WMRucioException` exception if the DID is unknown to Rucio
* listing dataset replicas (our `getReplicaInfoForBlocks` wrapper method) does this operation in bulk now, single API call with all the block names (using `list_dataset_replicas_bulk`)
NOTE: It requires rucio-clients >= 1.23.0
yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -253,31 +253,37 @@ class Rucio(object):
         elif 'block' in kwargs:
             blockNames = [kwargs['block']]
 
-        # FIXME: make bulk requests once https://github.com/rucio/rucio/issues/2459 gets fixed
         if isinstance(kwargs.get('dataset', None), (list, set)):
             for datasetName in kwargs['dataset']:
                 blockNames.extend(self.getBlocksInContainer(datasetName, scope=kwargs['scope']))
         elif 'dataset' in kwargs:
             blockNames.extend(self.getBlocksInContainer(kwargs['dataset'], scope=kwargs['scope']))
 
-        for blockName in blockNames:
-            replicas = []
-            response = self.cli.list_dataset_replicas(kwargs['scope'], blockName,
-                                                      deep=kwargs['deep'])
-            for item in response:
-                # same as complete='y' used for PhEDEx (which is always set within WMCore)
-                if item['state'].upper() == 'AVAILABLE':
-                    replicas.append(item['rse'])
-            result.append({'name': blockName, 'replica': list(set(replicas))})
+        inputDids = []
+        for block in blockNames:
+            inputDids.append({"scope": kwargs["scope"], "type": "DATASET", "name": block})
+
+        resultDict = {}
+        for item in self.cli.list_dataset_replicas_bulk(inputDids):
+            resultDict.setdefault(item['name'], [])
+            if item['state'].upper() == 'AVAILABLE':
+                resultDict[item['name']].append(item['rse'])
 
         if self.phedexCompat:
-            # convert plain node list to list of nodes dict
-            for block in result:
-                replicas = []
-                for node in block['replica']:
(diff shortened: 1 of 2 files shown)


amaltaro:
@belforte Stefano, this PR has a few changes to the Rucio wrapper that might affect you (especially that exception). Please let me know if you have any concerns. Description has been updated at the very top.

dciangot:
@amaltaro thanks, this can be useful, the part of the code involved in monitoring replica status still has to be ported to use WMCore class though. So we are going to use directly this new version.

amaltaro:
Thanks @dciangot Diego. Did I understand it right then that - right now - there should be no impact on CRAB with this change?
BTW, I just added a note about the rucio-client dependency on the PR description.

dciangot:
Yes correct, as far as I know, we should not use it, at least directly. I
don't know if it has any consequence on other wmcore Rucio api though.

Il gio 10 set 2020, 10:11 Alan Malta Rodrigues <notifications@github.com>
ha scritto:

amaltaro:
If I don't hear anything else, it will get merged tomorrow morning.

belforte:
@dciangot shouldn't we be using this in DBSDataDiscovery already ? If there's a new exception, is it catched in some other WMCore layer, or passed to CRAB ? Why do you mention the transfer monitoring component here ?
@amaltaro please go ahead and merge and we will test it and adapt as needed.

Of course we will not use it unless it is ported to 1.3.6.crab branch.
But we need to get serious about aligning CRAB to top of WMore repo again, I am sure you are fed up with this multibranch stuff.

This is in the relecant CRAB line:
https://github.com/dmwm/CRABServer/blob/0a592f9e3fb70d030e9d2255374b1235d9ac1f24/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L192
i.e.
```
                locations = rucioClient.getReplicaInfoForBlocks(scope=scope, block=list(blocks))
```
we catch any exception, log it, and then generate our own errors, so if you  raise a exceptions should be harmless.

dciangot:
mmm, right @belforte . I keep forgetting that part sorry, anyhow yes, should be harmless.

belforte:
@amaltaro maybe my misunderstanding (please ignore me then), but the list of kwargs in
https://github.com/amaltaro/WMCore/blob/7ebbd9c911be0ac1c78e78ae0bcb5cdf98c7d58d/src/python/WMCore/Services/Rucio/Rucio.py#L225-L244
appears to have little to do with the way we actually call this method

amaltaro:
You're completely right, Stefano. I overlooked the docstring, which is basically meaningless at this point. The bulk API only supports 3 parameters (name, scope, type), so most of those kwargs (legacy from PhEDEx) won't ever get evaluated. Nothing will break though, if you keep providing them in the API call.

PR #9905, opened 2020-09-11
Initial changes to PSet Tweaking.

### LINKED ISSUE #9808: Gather list of PSet tweaks applied by WMAgent during job runtime
**Impact of the new feature**
WMAgent

Yes, it's related to the tightly coupling of WMCore, WMControl and CMSSW, and the complications affecting the support of multiple python versions (in CMSSW and WMCore) and job configuration (Pset) tweaks applied during runtime.

This issue is meant to be an investigation of what exactly is tweaked by WMAgent - during the job runtime; what the source of the information is (things from the workflow description? things from the WMAgent job description? things based on the site and/or storage? things related to the data to be accessed? etc)

I think the result of this issue should be something like:
* CMSSW/PSet parameter tweaked:
* data format
* data source

Once we have an (almost) exhaustive list, we can then have further discussions and see whether this PSet tweaking would be better maintained in a different place (like in the CMSSW itself), such that the agent would only need to call that script, passing all the arguments to it, and be done.

The other alternative is what we have right now, which is an auto-discover of the CMSSW latest python version, and then executing the PSetTweak script to perform all the tweaks with the own CMSSW python version.

**Additional context**
This is likely a good start for this: https://github.com/dmwm/WMCore/blob/master/src/python/PSetTweaks/WMTweak.py#L22
another place is: 
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py

khurtado:
Fixes #9808

#### Status
In development 

#### Description
Split SetupCMSSWPSet in 2 stages (one that depends solely on WMCore libraries, and another that depends on the CMSSW framework)

### DIFF
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -7,10 +7,7 @@ Note: This can be used within the CMSSW environment to act on a
 process/config but does not depend on any CMSSW libraries. It needs to stay like this.
 
 """
-from __future__ import print_function, division
-
-from builtins import map, range, str, object
-from future.utils import viewitems, viewkeys
+from __future__ import print_function
 
 import logging
 import os
@@ -122,7 +119,7 @@ def lfnGroup(job):
     default both to 0. The result will be a 5-digit string.
     """
     modifier = str(job.get("agentNumber", 0))
-    jobLfnGroup = modifier + str(job.get("counter", 0) // 1000).zfill(4)
+    jobLfnGroup = modifier + str(job.get("counter", 0) / 1000).zfill(4)
     return jobLfnGroup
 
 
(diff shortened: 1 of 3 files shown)


src/python/PSetTweaks/WMTweak.py
```diff
@@ -7,10 +7,7 @@
 process/config but does not depend on any CMSSW libraries. It needs to stay like this.
 
 """
-from __future__ import print_function, division
-
-from builtins import map, range, str, object
```
amaltaro, line 12: Kenyi, just a random question. Why are you removing the python3 code modernization changes here?

khurtado: @amaltaro It was an unintentional removal ;). My changes were based on a version that dates before merging the python3 code modernization, apparently. I will have that updated eventually.

PR #9910, opened 2020-09-12
MSOutput: skip output data placement for Resubmission request type

### LINKED ISSUE #9870: Skip ACDC workflows in MSOutput
**Impact of the new feature**
MicroServices

This is not yet a strictly identified problem, but can easily become one as soon as we start MSOutput in production and subsequently re-modify/re-purpose it to deal with custodial transfer requests too. The major problem would become for the disk replica requests though. Since the initial disk transfer request is expected to come immediately once the first agent finishes a block creation, we expect that a transfer rule would already exist at the time the ACDC for a broken request has been created. This of course will not cover the cases when everything fails even before a single block has been completed and the ACDC is more or less a full remake of the already existing request.     

Skip all workflows where `'RequestType' == 'Resubmission'` for the full MSOutput service.

Check for a transfer request / Rucio Rule existence for every dataset on every iteration. This would generate enormous load.    

**Additional context**
N/A

amaltaro:
Fixes #9870 

#### Status
not-tested

#### Description
Given that the Resubmission `OutputDatasets` will be either the same or a subset of the original/first/parent workflow, there is no need to perform output data placement for Resubmission (aka ACDC) workflows. Actually, it's even safer not to, otherwise we could end up with multiple dataset replicas, even though the goal was a single replica.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -589,6 +589,12 @@ class MSOutput(MSCore):
         self.logger.info("Producing MongoDB record for workflow: %s", msOutDoc["RequestName"])
         updatedOutputMap = []
         for dataItem in msOutDoc['OutputMap']:
+            if msOutDoc['RequestType'] == "Resubmission":
+                # make sure not to subscribe the same datasets multiple times, even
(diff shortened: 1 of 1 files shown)


amaltaro:
It works like a charm!

PR #9911, opened 2020-09-12
MSOutput: support custodial/tape data placement

### LINKED ISSUE #9842: MSOutput: perform custodial data placement
**Impact of the new feature**
ReqMgr2MS

MSOutput needs to have the ability to make custodial data placement as well, which means, make transfer requests against MSS/Tape endpoints.

This development depends on:
https://github.com/dmwm/WMCore/issues/9841

then we should resume the data placement implementation and adapt whatever is needed to subscribe data against the long term storage (MSS/Tape).

For DDM data transfers, it seem we need to find out the final location in the MSOutput, thus requesting DDM to move data under a specific location.
For Rucio data transfers, it could be that we just need to provide a Tape expression, and let Rucio handle the final location. Needs to be discussed and investigated though. 

The general policy we discussed during the Rucio transition meeting was as follows:
   a. Use the API / Eric’s logic to pick the tape RSE in MSOutput as  mentioned in this issue: https://github.com/dmwm/WMCore/issues/9842#issuecomment-672194619 . The MSOutput should decide where the final destination (the site to which the tape request is made) for the dataset based on the quota and usage for all those. 
   b. Create the Rucio rule / DDM transfer request for the output datasets (rucio container) before the workflow gets announced (thus, when it's in `closed-out` state).
   c. The Rucio rule should be deleted manually (eventually by Nick)

none

**Additional context**
To be worked once this one has been implemented: https://github.com/dmwm/WMCore/issues/9841

amaltaro:
Fixes #9842 

#### Status
ready

#### Description
This PR makes MSOutput able to also perform the custodial (tape) data placement. Features in place are:
* tape dataset placement supports PhEDEx and Rucio;
* dataset placement is skipped if the tier is blacklisted in the MicroService configuration;
* dataset placement is skipped for tiers blacklisted in the Unified configuration;
* transfer request / rule creation is always made against a single RSE/node (with a single copy)
  * the same RSE endpoint gets used by the other datasets within the same workflow
* a workflow is only completely done once all the Disk + Tape placement has succeeded
* dry-run mode supported as well;
* service summary report unified (data structures were the same anyways);
* update the local caches for both MSOutput threads (Consumer and Producer);
* removed `"&cms_type=real&rse_type=DISK"` from the RelVal RSE expression. We always set the specific RSE name anyways, so that expression won't make any difference;
* FIXME: tape destination placeholder created (and hardcoded to FNAL). To be fixed in #9841 

no, new feature

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -9,18 +9,19 @@ the Output data placement in WMCore MicroServices.
 from __future__ import division, print_function
 
 # system modules
+from http.client import HTTPException
 from pymongo import IndexModel, ReturnDocument, errors
 from pprint import pformat
 from threading import current_thread
 from retry import retry
 
 # WMCore modules
-from WMCore.MicroService.DataStructs.DefaultStructs import OUTPUT_PRODUCER_REPORT
-from WMCore.MicroService.DataStructs.DefaultStructs import OUTPUT_CONSUMER_REPORT
+from WMCore.MicroService.DataStructs.DefaultStructs import OUTPUT_REPORT
 from WMCore.MicroService.Unified.MSCore import MSCore
 from WMCore.Services.DDM.DDM import DDM, DDMReqTemplate
 from WMCore.Services.CRIC.CRIC import CRIC
 from Utils.EmailAlert import EmailAlert
+from WMCore.Services.PhEDEx.DataStructs.SubscriptionList import PhEDExSubscription
 from Utils.Pipeline import Pipeline, Functor
 from WMCore.Database.MongoDB import MongoDB
 from WMCore.MicroService.DataStructs.MSOutputTemplate import MSOutputTemplate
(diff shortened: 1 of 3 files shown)


amaltaro:
Basic tests done, including some real call against DDM and PhEDEx (rucio account needs to be sorted). So, it's good enough to go.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -301,11 +288,8 @@ def makeSubscriptions(self, workflow):
 
                 # overwrite the current DiskDestination by something that will work for DDM
```
nsmith-, line 289: So unlike Unified, the output disk site choice is resolved here? Because as I understand, Unified just forwards the request to dynamo (with a generic or `null` list of sites) in https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/closor.py#L585 (perhaps @dr-stringfellow can confirm) If so, we should review how the disk site is chosen.

nsmith- (commented):
I took a look and it makes sense to me

src/python/WMCore/MicroService/Unified/MSOutput.py
amaltaro: Actually, the disk destination is decided in this block:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/MSOutput.py#L755-L791
but noticed that the destination is already in the format of an RSE Expression (Rucio compliant). 

In this section of the code, we are making data transfers against DDM, so we need to massage a bit the list of sites (for RelVals, it does not change). For standard workflows, we simply pass T1 Disk and T2s, and leave the final decision to DDM.

PR #9916, opened 2020-09-15
MSOutput: new DatasetSize param for the mongo record; fetch container size from DM systems

### LINKED ISSUE #9913: MSOutput: calculate total datasets size within a workflow and use it for the destination selection
**Impact of the new feature**
ReqMgr2MS - MSOutput

Not yet, but the tape output data placement does not consider the total output data volume when deciding which destination to be used. Which means, we risk trying to transfer more data than what a site could hold. This is a not a problem for non-custodial / disk data placement, since their destination will be decided post-transfer request, by the data management system itself.

There are a few things to be addressed with this issue, like:
* change the mongodb record schema to also hold a `DatasetSize` (or ContainerSize) in the `OutputMap` parameter;
* in the Producer thread, get the container size (in bytes) from either PhEDEx or Rucio. In the most efficient way, maybe using the pycurl-based APIs;
* in the Consumer thread, sum up the dataset size of all output datasets that need to go to tape, and use that value to properly decide which tape destination to choose (`_getTapeDestination` method).

The one implemented at the moment, just assume data can fit in the remaining tape storage.

**Additional context**
none

amaltaro:
Fixes #9913 

#### Status
ready

#### Description
In short, adds the container size to the mongodb record, such that we can calculate the total number of bytes for all the output datasets to be pinned on a tape storage.
In more details:
* added `DatasetSize` to the `OutputMap` attribute of the mongodb records
* in the Producer thread, retrieve the container size either from Rucio or PhEDEx (sequentially, for now)
* in the Consumer thread, find all the output datasets that need to be transferred to tape, and sum up their sizes;
* once the workflow total size (in bytes) is known, pass it to the method picking the final custodial destination. If quota (or remaining space) is not enough to host all those containers, that RSE endpoint is dropped from the selection.

no, new mongodb parameter

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
+++ b/src/python/WMCore/MicroService/DataStructs/MSOutputTemplate.py
@@ -127,6 +127,7 @@ class MSOutputTemplate(dict):
         Template format:
             {'Campaign': u'RunIIAutumn18DRPremix',
              'Dataset': u'/Pseudoscalar2HDM_MonoZLL_mScan_mH-500_ma-300/DMWM-TC_PreMix_khurtado_TC_PreMix-v11/AODSIM',
+             'DatasetSize': 0,
              'Copies': 1,
              'DiskDestination': "",
              'TapeDestination': "",
@@ -137,6 +138,7 @@ class MSOutputTemplate(dict):
         outMapTemplate = [
             ('Campaign', "", (str, unicode)),
             ('Dataset', "", (str, unicode)),
+            ('DatasetSize', 0, int),
             ('Copies', 1, int),
             ('DiskDestination', "", (str, unicode)),
             ('TapeDestination', "", (str, unicode)),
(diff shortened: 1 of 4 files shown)


amaltaro:
Everything works well with this patch and performance results are better than what I expected for a sequential resolution of the container size.
```
** For Rucio - even though most of the DIDs were not available in the production database**
2020-09-15 15:28:32,646:INFO:MSOutput: MSOutputProducer: Total 1194 requests processed from the streamer. 
2020-09-15 15:28:32,719:INFO:MSManager: Total outputProducer execution time: 20 secs
```

```
** For PhEDEx - even though the NANOAOD{,SIM} was not available in TMDB **
2020-09-15 15:44:16,756:INFO:MSOutput: MSOutputProducer: Total 1328 requests processed from the streamer. 
2020-09-15 15:44:16,881:INFO:MSManager: Total outputProducer execution time: 147 secs
```

I'm going to merge it and patch MSOutput in CMSWEB later today (as already discussed with Imran).

amaltaro:
Production MSOutput (on vocms0742) is running with this patch in.

PR #9930, opened 2020-09-22
Construct the PU location via rucio considering MSTransferor locks

### LINKED ISSUE #9894: Update PileupFetcher to use Rucio to build the list of replicas and block location
**Impact of the new feature**
WMAgent

PileupFetcher uses PhEDEx to build a dictionary of:
* blocks and their current location
* blocks and a list of all their files/replicas

as we decommission PhEDEx, the same functionality needs to be ported to Rucio.

Support Rucio within PileupFetcher, such that the pileup dataset location can be performed through Rucio.
This means, we need to implement:
* given a container name, return all its blocks and their current location (where data available and locked by WMCore)
* given a container name, return all its blocks and their files/replicas

The piece of code building the list of files need to be reviewed though. We fetch the list from DBS, which then gets overwritten by the list of files retrieved from PhEDEx (likely taking care of the invalid/no-longer-existent files).
It's also not clear to me why we need to know the amount of events in each block..

Not resolve the list of files in Rucio, rely on that information from DBS (and maybe filter by `validOnly`).

**Additional context**
none

amaltaro:
Fixes #9894 

#### Status
ready

#### Description
Previous logic in WMAgent for resolving the pileup location (the pileupconf.json construction) was simply relying on the current block location, without considering who locked the data.

This PR changes that functionality to something similar to what is done at WorkQueue level, such that we only use data that has been locked by MSTransferor.

I also have two somehow unrelated changes here:
* `FileList` will contain a flat list of file names (instead of {"logical_file_name": <the file name>})
* `FileList` is no longer sorted. I couldn't see any reason to sort it. Besides, at SetupCMSSWPSet level we shuffle all those files..

no, between old and new agents. But this isn't a valid use case.

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -60,6 +60,32 @@ def weightedChoice(choices):
     assert False, "Shouldn't get here"
 
 
+def isTapeRSE(rseName):
+    """
+    Given an RSE name, return True if it's a Tape RSE (rse_type=TAPE), otherwise False
+    :param rseName: string with the RSE name
+    :return: True or False
+    """
+    # NOTE: a more reliable - but more expensive - way to know that would be
+    # to query `get_rse` and evaluate the rse_type parameter
+    return rseName.endswith("_Tape")
+
+
+def dropTapeRSEs(listRSEs):
+    """
+    Method to parse a list of RSE names and return only those that
+    are not a rse_type=TAPE, so in general only Disk endpoints
+    :param listRSEs: list with the RSE names
+    :return: a new list with only DISK RSE names
+    """
+    diskRSEs = []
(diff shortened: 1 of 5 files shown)


amaltaro:
@belforte Stefano, I should have implemented it yesterday. But in case it's not too late, I have made 2 useful utilitarian functions in the Rucio wrapper module. Check this PR out if you want.

belforte:
thanks Alan. Not too late. I will adopt https://github.com/dmwm/CRABServer/issues/6165

amaltaro:
@khurtado FYI, this PR will slightly change the format of our pileupconf.json data structure (FileList will be a flat list, instead of a list of dicts).

amaltaro:
Unit test tests most of this functionality, a real test is still required though, once we have more data in Rucio int database

Issue #9632, opened 2020-04-14
Propagate total number of lumis expected to the job runtime


amaltaro:
**Impact of the new feature**
ReqMgr2, WMAgent

**Is your feature request related to a problem? Please describe.**
This has been discussed within the run-dependent group and it's needed to better support run-period era distribution in MC workflows.
From the google document, this is the important piece:
```
in the WMA infrastructure, percolation of the total number LSs requested for a dataset, and of the job-specific LS index to all production jobs;
rescoping of LS transitions to run transitions in the digitization step in CMSSW.

Currently, for production requests starting from GEN (which are the largest part of the request with the current production approach), the LS index is already tracked by WMA guaranteeing its uniqueness for each job. WMA developers consider the propagation of the additional parameter storing the total # of LS requested a minor development. What might require more thinking is how to handle the recovery jobs injected manually for problematic workflows.
```

**Describe the solution you'd like**
Global workqueue calculates the estimated total number of lumi sections to be produced in a workflow, that information gets added to the high level description of the workflows as `TotalInputLumis`.

Thus, WMAgent has to pick that information - which in rare cases, is not available - and add it to the job baggage, such that it can be loaded during the job runtime and added to the PSet tweak. I'm not sure whether there is a CMSSW parameter already available for that or not(?) Perhaps we can use
```
process.source.totalLumisInWorkflow
```
? I believe it needs to be supported by the CMSSW release as well, otherwise it will fail when cmsRun loads the PSet configuration.

**Describe alternatives you've considered**
In theory, none.

**Additional context**
none

PR #9933, opened 2020-09-23
WIP: Squash changes to RucioInjectorPoller.deleteBlocks

### LINKED ISSUE #9623: Implement the T0 block deletion logic in RucioInjector
**Impact of the new feature**
WMAgent

**Additional context**

germanfgv:
Fixes #9623

#### Status
In developmen

#### Description
Implemented Tier0 block deletion logic described in issue #9632. Identifies deletable blocks but does not remove any rules. Using getDataLockedAndAvailable() to check block transfer status.

YES

#### Related PRs
No

No


germanfgv:
I'm not sure about the usage of getDataLockedAndAvailable(), as it appears to me a block status will only be 'OK' when the whole container has been transferred. Is this true? In that case, which method should I use to check block level transfer status?

germanfgv:
listDataRules() does not list rules of parent DIDs, so the implementation must change. I'll create a wrapper to list parent DIDs and then search for their rules.

PR #9935, opened 2020-09-24
Rucio list parent dids

germanfgv:
Builds towards fixing #9623 

#### Status
ready

#### Description
Wrapper for Rucio API's list_parent_dids()

YES

#### Related PRs
As stated in #9933, we need to list parent DIDs for each block

No

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -622,6 +622,23 @@ class Rucio(object):
             self.logger.error("Exception listing rules history for data: %s. Error: %s", name, str(ex))
         return list(res)
 
+    def listParentDIDs(self, name, scope='cms'):
+        """
+        _listParentDID__
+
+        List the parent block/container of the specified DID.
+        :param name: data identifier (either a block or a file name)
+        :param scope: string with the scope name
+        :return: a list with dictionary items
+        """
+        res = []
+        try:
+            res = self.cli.list_parent_dids(scope, name)
+        except Exception as ex:
+            self.logger.error("Exception listing parent DIDs for data: %s. Error: %s", name, str(ex))
+        return list(res)
+
+
     def getRule(self, ruleId, estimatedTtc=False):
         """
         _getRule_


amaltaro:
@germanfgv just to make sure we have the correct concept of parent in the context of Rucio.
* a file has a block parent, basically the block which the file belongs to
* a block has a container (or multiple, if we start creating customized containers), basically the container which the block belongs to
is that correct?

In that case, if I call this API `list_parent_dids` with a block name. Does it look for all the rules locking the parent container? Or it lists all rules of the parent container AND the rules for this block?

germanfgv:
> @germanfgv just to make sure we have the correct concept of parent in the context of Rucio.
> 
> * a file has a block parent, basically the block which the file belongs to
> * a block has a container (or multiple, if we start creating customized containers), basically the container which the block belongs to
>   is that correct?
> 
> In that case, if I call this API `list_parent_dids` with a block name. Does it look for all the rules locking the parent container? Or it lists all rules of the parent container AND the rules for this block?

`list_parent_dids` does not look for rules, it looks for parent DIDs. If you call this API with a block name, it will return a list of dictionaries, each one describing a parent container. If you call this API with a file name, it will return a dictionary describing the parent block (not sure if a file can belong to multiple blocks).

In order to list all the rules locking the parent container of a block, I will use listParentDIDs to get the parent container and then use listDataRules to get its rules.

amaltaro:
Sorry, I don't know how I came up with rules here(!)
Your answer clarifies my question though. Thanks

germanfgv:
I could have created a method using both list_parent_dids and  list_did_rules to directly get the parents's rules, but I decided against that. My reasoning was that being able to list parent DIDs may be useful for other purposes in RucioInjector.py, so I created listParentDIDs. @amaltaro Do you agree with this?

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -622,6 +622,23 @@ def listDataRulesHistory(self, name, scope='cms'):
...
+        :param name: data identifier (either a block or a file name)
+        :param scope: string with the scope name
+        :return: a list with dictionary items
+        """
+        res = []
+        try:
+            res = self.cli.list_parent_dids(scope, name)
+        except Exception as ex:
+            self.logger.error("Exception listing parent DIDs for data: %s. Error: %s", name, str(ex))
+        return list(res)
```
amaltaro, line 639: Are you casting it to a list because it returns a generator?

amaltaro (approved):
@germanfgv code looks good to me. And I agree in separating these APIs for dids and rules too.
I asked a question in the code though, other than that it's good to go from my side.

Actually, I would ask you to implement a unit test for it. But since I'm working on a few other unit tests for this Rucio wrapper, I can take care of that.

src/python/WMCore/Services/Rucio/Rucio.py
germanfgv: Yes, it does return a generator

amaltaro:
Thanks German

PR #9940, opened 2020-09-25
Add new module to convert cp stats into prometheus format suited for CMS monitoring

### LINKED ISSUE #9939: Provide proper metrics for web based services
**Impact of the new feature**
Based on recent incident [INC2542764](https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC2542764) with low service performance of Python based services on k8s it would be extremely helpful to provide proper set of metrics from service stack. Since all cmsweb Python based services (DBS, ReqMgr2, CRAB etc.) are based on WMCore codebase it is natural to implement service metrics in WMCore rather in upstream services. This will allow proper monitoring of web services.

Extend [ProcessStats.py](https://github.com/dmwm/WMCore/blob/master/src/python/Utils/ProcessStats.py) metrics to include the following
- number of requests
- status code stats
- response time of request
- number of connections
- number of open file descriptors
- if DB is in use DB related stats

Please see this incident discusssions [INC2542764](https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC2542764)

None, currently we run our services without proper knowledge how they behave, what is their throughput and how the perform under production load

**Additional context**
Once metrics are added to [ProcessStats.py](https://github.com/dmwm/WMCore/blob/master/src/python/Utils/ProcessStats.py) we need to modify WMCore WebTools and REST interface to add `/metrics` or `/status` end-points and enable this by default that it would not be required in upstream code to change anything in order to have them. Once they are available we can setup proper exporter (code should be developed in CMS Monitoring group) and integrate it into Prometheus metrics for each individual service.

vkuznet:
Fixes #9939 

#### Status
ready

#### Description
I provide a new module which can be integrated into WMCore stack. It provides useful set of functions to flatten CherryPy stats, make static schema and provide them in format suitable for Prometheus server.

I suggest to add an additional `metrics` endpoint to complement `stats` one shown in here
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Server.py#L700
with the following
```
@expose
def metrics(self):
       name = self.__class__.__name__
       return prom_metrics(name)
```

This will provide ability to scrape all cherrypy metrics from Prometheus for any DMWM/REST server. Please refer to docstring of provided module for format description/example.

YES

this module uses cherrypy lib to get cherrypy metrics

### DIFF
--- a/src/python/Utils/CPMetrics.py
+++ b/src/python/Utils/CPMetrics.py
@@ -1,0 +1,267 @@
+#!/usr/bin/env python3
+"""
+File       : CPMetrics.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
+Description: This module provide functions to flatten cherrypy stats
+and provide them in Prometheus format for scraping.
+
+Below we present three different outputs:
+    - CherryPy stats default output
+    - flatten structure of stats
+    - prometheus output for stats
+
+### Defautl CherryPy metrics output
+
+{
+    "Cheroot HTTPServer 4388603856": {
+    "Accepts": 0,
+    "Accepts/sec": 0.0,
+    "Bind Address": "('127.0.0.1', 8080)",
+    "Bytes Read": -1,
+    "Bytes Written": -1,
+    "Enabled": false,
+    "Queue": 0,
+    "Read Throughput": -1,
+    "Requests": -1,
+    "Run time": -1,
+    "Socket Errors": 0,
+    "Threads": 20,
+    "Threads Idle": 19,
+    "Work Time": -1,
+    "Worker Threads": {
+        "CP Server Thread-10": {
+        "Bytes Read": 0,
+        "Bytes Written": 0,
+        "Read Throughput": 0.0,
+        "Requests": 0,
+        "Work Time": 0,
+        "Write Throughput": 0.0
+        },
+        .....
+    },
+    "Write Throughput": -1
+    },
+    "CherryPy Applications": {
+        "Bytes Read/Request": 0.0,
+        "Bytes Read/Second": 0.0,
+        "Bytes Written/Request": 0.0,
+        "Bytes Written/Second": 0.0,
+        "Current Requests": 0,
+        "Current Time": 1601039589.916887,
+        "Enabled": true,
+        "Requests": {},
+        "Requests/Second": 0.0,
+        "Server Version": "17.4.2",
+        "Start Time": 1601039576.541158,
+        "Total Bytes Read": 0,
+        "Total Bytes Written": 0,
+        "Total Requests": 0,
+        "Total Time": 0,
+        "Uptime": 13.375718116760254
+    }
+}
+
+### flatten structure via flattenStats function
+{
+"cherrypy_app_requests_second": 0.0,
+"cherrypy_app_uptime": 1.7454230785369873,
+"cherrypy_app_total_time": 0,
+"cherrypy_app_current_time": 1601040713.369842,
+"cherrypy_app_bytes_read_second": 0.0,
+"cherrypy_app_requests": {},
+"cherrypy_http_server_queue": 0,
+"cherrypy_app_bytes_written_second": 0.0,
+"cherrypy_app_total_requests": 0,
+"cherrypy_http_server_write_throughput": -1,
+"cherrypy_app_enabled": true,
+"cherrypy_app_start_time": 1601040711.624408,
+"cherrypy_http_server_threads": 20,
+"cherrypy_http_server_work_time": -1,
+"cherrypy_app_bytes_read_request": 0.0,
+"cherrypy_http_server_bytes_read": -1,
+"cherrypy_http_server_accepts_sec": 0.0,
+"cherrypy_server_worker_threads": [
+   {"thread_name": "cp_server_thread_3", "read_throughput": 0.0,
+   "work_time": 0, "write_throughput": 0.0, "bytes_written": 0,
+   "bytes_read": 0, "requests": 0}, ....],
+"cherrypy_http_server_run_time": -1,
+"cherrypy_http_server_bind_address": "('127.0.0.1', 8080)",
+"cherrypy_app_server_version": "17.4.2",
+"cherrypy_http_server_enabled": false,
+"cherrypy_http_server_socket_errors": 0,
+"cherrypy_app_bytes_written_request": 0.0,
+"cherrypy_app_current_requests": 0,
+"cherrypy_app_total_bytes_read": 0,
+"cherrypy_http_server_bytes_written": -1,
+"cherrypy_http_server_threads_idle": 19,
+"cherrypy_http_server_requests": -1,
+"cherrypy_app_total_bytes_written": 0,
+"cherrypy_http_server_read_throughput": -1,
+"cherrypy_http_server_accepts": 0}
+
+### prometheus exporter structure provided via promMetrics function
+
+# HELP cherrypy_app_requests_second
+# TYPE cherrypy_app_requests_second gauge
+cherrypy_app_requests_second 0.0
+# HELP cherrypy_app_uptime
+# TYPE cherrypy_app_uptime gauge
+cherrypy_app_uptime 2.03766894341
+# HELP cherrypy_app_total_time
+# TYPE cherrypy_app_total_time counter
+cherrypy_app_total_time 0
+# HELP cherrypy_app_current_time
+# TYPE cherrypy_app_current_time gauge
+cherrypy_app_current_time 1601043839.47
+# HELP cherrypy_app_bytes_read_second
+# TYPE cherrypy_app_bytes_read_second gauge
+cherrypy_app_bytes_read_second 0.0
+# HELP cherrypy_http_server_queue
+# TYPE cherrypy_http_server_queue counter
+cherrypy_http_server_queue 0
+# HELP cherrypy_app_bytes_written_second
+# TYPE cherrypy_app_bytes_written_second gauge
+cherrypy_app_bytes_written_second 0.0
+# HELP cherrypy_app_total_requests
+# TYPE cherrypy_app_total_requests counter
+cherrypy_app_total_requests 0
+# HELP cherrypy_http_server_write_throughput
+# TYPE cherrypy_http_server_write_throughput counter
+cherrypy_http_server_write_throughput -1
+# HELP cherrypy_server_worker_threads
+# TYPE cherrypy_server_worker_threads histogram
+cherrypy_server_worker_threads{thread_name="cp_server_thread_3",...} 1
+# HELP cherrypy_app_start_time
+# TYPE cherrypy_app_start_time gauge
+cherrypy_app_start_time 1601043837.44
+# HELP cherrypy_http_server_threads
+# TYPE cherrypy_http_server_threads counter
+cherrypy_http_server_threads 20
+# HELP cherrypy_http_server_work_time
+# TYPE cherrypy_http_server_work_time counter
+cherrypy_http_server_work_time -1
+# HELP cherrypy_app_bytes_read_request
+# TYPE cherrypy_app_bytes_read_request gauge
+cherrypy_app_bytes_read_request 0.0
+# HELP cherrypy_http_server_bytes_read
+# TYPE cherrypy_http_server_bytes_read counter
+cherrypy_http_server_bytes_read -1
+# HELP cherrypy_http_server_accepts_sec
+# TYPE cherrypy_http_server_accepts_sec gauge
+cherrypy_http_server_accepts_sec 0.0
+# HELP cherrypy_http_server_run_time
+# TYPE cherrypy_http_server_run_time counter
+cherrypy_http_server_run_time -1
+# HELP cherrypy_http_server_socket_errors
+# TYPE cherrypy_http_server_socket_errors counter
+cherrypy_http_server_socket_errors 0
+# HELP cherrypy_app_bytes_written_request
+# TYPE cherrypy_app_bytes_written_request gauge
+cherrypy_app_bytes_written_request 0.0
+# HELP cherrypy_app_current_requests
+# TYPE cherrypy_app_current_requests counter
+cherrypy_app_current_requests 0
+# HELP cherrypy_app_total_bytes_read
+# TYPE cherrypy_app_total_bytes_read counter
+cherrypy_app_total_bytes_read 0
+# HELP cherrypy_http_server_bytes_written
+# TYPE cherrypy_http_server_bytes_written counter
+cherrypy_http_server_bytes_written -1
+# HELP cherrypy_http_server_threads_idle
+# TYPE cherrypy_http_server_threads_idle counter
+cherrypy_http_server_threads_idle 19
+# HELP cherrypy_http_server_requests
+# TYPE cherrypy_http_server_requests counter
+cherrypy_http_server_requests -1
+# HELP cherrypy_app_total_bytes_written
+# TYPE cherrypy_app_total_bytes_written counter
+cherrypy_app_total_bytes_written 0
+# HELP cherrypy_http_server_read_throughput
+# TYPE cherrypy_http_server_read_throughput counter
+cherrypy_http_server_read_throughput -1
+# HELP cherrypy_http_server_accepts
+# TYPE cherrypy_http_server_accepts counter
+cherrypy_http_server_accepts 0
+"""
+
+# system modules
+import json
+
+# WMCore dependencies
+from Utils.Utilities import decodeBytesToUnicode
+
+
+def promMetrics(data, exporter):
+    """
+    Provide cherrypy stats prometheus metrics for given exporter name.
+    """
+    metrics = flattenStats(data)
+    if isinstance(metrics, str):
+        metrics = json.loads(metrics)
+    # the following keys will be skipped
+    skip = [
+        'cherrypy_app_enabled',
+        'cherrypy_http_server_bind_address',
+        'cherrypy_app_requests',
+        'cherrypy_app_server_version',
+        'cherrypy_http_server_enabled']
+    # our prometheus data representation
+    pdata = ""
+    for key, val in metrics.items():
+        if key in skip:
+            continue
+        # add exporter name as a prefix for each key
+        key = '{}_{}'.format(exporter, key)
+        mhelp = "# HELP {}\n".format(key)
+        if isinstance(val, list):
+            mtype = "# TYPE {} histogram\n".format(key)
+            pdata += mhelp
+            pdata += mtype
+            for wdict in val:
+                entries = []
+                for kkk, vvv in wdict.items():
+                    entries.append('{}="{}"'.format(kkk, vvv))
+                entry = "{%s}" % ','.join(entries)
+                pdata += "{}{} 1\n".format(key, entry)
+        elif isinstance(val, (str, tuple)):
+            continue
+        else:
+            pdata += mhelp
+            if isinstance(val, int):
+                mtype = "# TYPE {} counter\n".format(key)
+            if isinstance(val, float):
+                mtype = "# TYPE {} gauge\n".format(key)
+            pdata += mtype
+            pdata += "{} {}\n".format(key, val)
+    return pdata
+
+
+def flattenStats(cpdata):
+    "Flatten cherrypy stats to make them suitable for monitoring"
+    if isinstance(cpdata, str) or isinstance(cpdata, bytes):
+        cpdata = json.loads(decodeBytesToUnicode(cpdata))
+    data = {}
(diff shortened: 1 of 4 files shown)


amaltaro (changes requested):
Not really a review of the code you proposed, but a couple of things caught my attention:
* please use lower camelCase for the variable and function names (prom_metrics --> promMetrics)
* given that this is going to be a common Utils module, please provide docstrings as completed as you can (there is no model defined yet, but it looks like we are adopting reStructured Text)
* can you also please create unit test(s) for this module

Can you please take care of the comments above? Thanks

vkuznet:
Alan, I added requested changes, please review at your convenience.

src/python/Utils/CPMetrics.py
```diff
@@ -0,0 +1,305 @@
...
+        if key.lower().find('cherrypy') != -1:
+            for kkk, vvv in val.iteritems():
+                nkey = 'cherrypy_app_%s' % kkk
+                nkey = nkey.lower().replace(" ", "_").replace("/", "_")
+                data[nkey] = vvv
+        if key.lower().find('cheroot') != -1:
+            for kkk, vvv in val.iteritems():
+                if kkk == 'Worker Threads':
+                    wdata = []
+                    for wkey, tval in vvv.iteritems():
```
todor-ivanov, line 258: Same comment as above regarding the variable names for `wkey` and `tval`. Since here it already becomes clear there is a third level in the CherryPy data structure we are considering, I would, for sure, make those variable names more descriptive.

src/python/Utils/CPMetrics.py
```diff
@@ -0,0 +1,305 @@
...
+                else:
+                    nkey = 'cherrypy_http_server_%s' % kkk
+                    nkey = nkey.lower().replace(" ", "_").replace("/", "_")
+                    data[nkey] = vvv
+    return data
+
+
+class MetricsServer(object):
+    """
+    Example of metrics server which can server cherrypy metrics
```
todor-ivanov, line 276: A typo here. Should be:
`     Example of metrics server which can serve cherrypy metrics`

todor-ivanov (approved):
Valya, I made few minor comments in the code. But, if you think those are too picky, just ignore them, we can definitely live with the code as it is. 
All the rest looks good to me!

src/python/Utils/CPMetrics.py
```diff
@@ -0,0 +1,305 @@
...
+            if isinstance(val, float):
+                mtype = "# TYPE {} gauge\n".format(key)
+            pdata += mtype
+            pdata += "{} {}\n".format(key, val)
+    return pdata
+
+
+def flattenStats(cpdata):
+    "Flatten cherrypy stats to make them suitable for monitoring"
+    if isinstance(cpdata, str):
```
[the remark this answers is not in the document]
vkuznet: Todor, it is each to check, it work fine with python3
```
Python 3.7.9 (default, Sep  6 2020, 13:20:25)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
>>> s="bla"
>>> isinstance(s, str)
True
```

src/python/Utils/CPMetrics.py
```diff
@@ -0,0 +1,305 @@
...
+        cpdata = json.loads(cpdata)
+    data = {}
+    for key, val in cpdata.iteritems():
```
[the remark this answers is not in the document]
vkuznet: Since we don't own CP stats data structure we cannot guarantee that it will be sustainable. It is their API at the end. But so far the parser I provided works with current CP stats structure.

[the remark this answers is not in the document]
vkuznet: regarding keys/vals, I changed them as you requested

src/python/Utils/CPMetrics.py
vkuznet: ok, replaced to workerKey and threadValue.

src/python/Utils/CPMetrics.py
vkuznet: done

vkuznet:
@todor-ivanov  please review again as I resolved all requested suggestions.

src/python/Utils/CPMetrics.py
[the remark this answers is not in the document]
mapellidario: In this case, we may be luckier that expected and no modification may be required.

#### Fact 1

`json.loads()` is very permissive in the type of its input.

1. [python2](https://docs.python.org/2/library/json.html#json.loads): `str` or `unicode`
2. [python3](https://docs.python.org/3/library/json.html#json.loads): `str`, `bytes` or `bytearray`

If  `cpdata` is of type `str`, then it is valid input for `json.loads()` in both python2 and python3.

`json.dumps()` return a `str` in both [py2](https://docs.python.org/2/library/json.html#json.dumps) (meaning a sequence of bytes here) and [py3](https://docs.python.org/3/library/json.html#json.dumps) (meaning a sequence of unicode codepoints). 

#### Fact 2

I am not an expert of `cherrypy`, but looking at [pypi](https://pypi.org/project/CherryPy/) the developers state that 

1. version [17.4.2](https://pypi.org/project/CherryPy/17.4.2/) of cherrypy is the last compatible with both py2 (py2.7) and py3 (py3.0->py3.3)
2. version [18.6.0](https://pypi.org/project/CherryPy/18.6.0/) of cherrypy (latest at time fo this comment) is compatible with py3 only (>= py3.5)

In version [17.4.2](https://github.com/cherrypy/cherrypy/blob/c3c2cec9a74f2c9fe71849c9aff5ab0e7f5743ec/cherrypy/lib/cpstats.py#L679), `cherrypy.lib.cpstats.StatsPage.data()` returns `json.dumps()`. If we run this in python2, then `cpdata` is of type `str` (meaning a sequence of bytes) and it is valid input for `json.loads()`.

In version [18.6.0](https://github.com/cherrypy/cherrypy/blob/b2938bc92570a1a04aa4336412e1824b9fe455eb/cherrypy/lib/cpstats.py#L677), `cherrypy.lib.cpstats.StatsPage.data()` returns `json.dumps().encode('utf-8')`. If we run this in python3, then `cpdata` is of type `str` (meaning a sequence of unicode codepoints) and is valid input for `json.loads()`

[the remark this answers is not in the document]
todor-ivanov: Thanks @vkuznet and @mapellidario !

todor-ivanov (approved):
Hi Valya, I think it is looking good now. Just to clear my own understanding. From your comment in the PR I read that provided the code like that it is not referred anywhere outside. So we should make an explicit call to it from here: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Server.py#L700

`
@expose
def metrics(self):
       name = self.__class__.__name__
       return prom_metrics(name)
 `

If I am correct it would be nice to have it called in this same PR so that  we avoid making too many PRs for the  same functionality. @amaltaro, Please take a look and tell me if you like to have the code merged with or without being referred in this PR.

vkuznet:
Todor,
it is fine from my side to add changes to REST Server in this PR. But I think either you or Alan should decide on it. Here is a proposed changes (in diff format):
```
diff --git a/src/python/WMCore/REST/Server.py b/src/python/WMCore/REST/Server.py
index b3a3b8a73..450040e49 100644
--- a/src/python/WMCore/REST/Server.py
+++ b/src/python/WMCore/REST/Server.py
@@ -15,6 +15,7 @@ from cherrypy.lib import cpstats
 from WMCore.REST.Error import *
 from WMCore.REST.Format import *
 from WMCore.REST.Validation import validate_no_more_input
+from Utils.CPMetrics import promMetrics

 try:
     from cherrypy.lib import httputil
@@ -360,6 +361,11 @@ class RESTFrontPage:
         "Return CherryPy stats dict about underlying service activities"
         return cpstats.StatsPage().data()

+    @expose
+    def metrics(self):
+        "Return CherryPy stats metrics in Prometheus format"
+        name = self.__class__.__name__
+        return promMetrics(name)


 ######################################################################
```

amaltaro:
Valentin, just to let you know that this PR is on my todo list for the coming days. I still think there is either something missing or a misunderstanding with this proposal, so I'll have to test a few things myself first.

vkuznet:
@amaltaro , I incorporated your changes from #9961 over here. I removed cherrypy deps and TestMetricsServer from CPMetrics.py code. But it would be nice to keep TestMetricsServer some place that users can easily see how to construct required end-point and test it easily. I created this file but did not include it in this PR. If you think it is relevant let me know where to put it, otherwise we may just skip it.
```
#!/usr/bin/env python
"""
File       : cp_metrics.py
Description: Example of Python REST Server to provide cherrypy stats as prometheus metrics
"""



import cherrypy
from cherrypy.lib import cpstats

class MetricsServer(object):
    """
    Example of metrics server which can serve cherrypy metrics
    in Prometheus format
    """
    def metrics(self):
        "metrics end-point for prometheus to scrape"
        data = cpstats.StatsPage().data()
        pdata = promMetrics(data, 'test')
        return pdata
    metrics.exposed = True

    def index(self):
        "default index end-point"
        return "Hello World"
    index.exposed = True


def TestMetricsServer():
    "Test MetricsServer function"
    cherrypy.root = MetricsServer()
    cherrypy.config.update({'server.socket_port': 8080,
                            'server.thread_pool': 20,
                            'environment': 'production',
                            'log.screen': True,
                            'log.error_file': "crab.log"})
    conf = {'/': {'tools.staticdir.root': os.getcwd()}}
    cherrypy.quickstart(MetricsServer(), '/', config=conf)

if __name__ == '__main__':
    TestMetricsServer()

```

vkuznet:
@amaltaro , I found where stats and metrics can be plugged into WebTools server. Looking at DBS code it inherits WebTools RESTModel, which by itself inherits from WebAPI.py. The whole chain of dependencies is the following:
- https://github.com/dmwm/DBS/blob/master/Server/Python/src/dbs/web/DBSReaderModel.py#L69
- https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WebTools/RESTModel.py#L24
- https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WebTools/WebAPI.py

So, I added stats and metrics to WebAPI and exposed them. Doing this way all DBS servers (inherited from WebTools) will get these end-points.

Please review.

vkuznet:
I had a look at pylint report and all of the failures related to WebAPI.py codebase which I didn't touched. I don't think it is my responsibility to go through them and include in this PR. Please let me know how you'll treat it.

vkuznet:
I had a look at Alan's VM and even though I can access provided end-points via https calls they are not accessible from localhost and plain http:
```
# https requests works fine
https://alancc7-cloud1.cern.ch/reqmgr2/data/metrics

# http request is not allowed giving 403 forbidden page
http://localhost:8246/reqmgr2/data/metrics
```

Such strict constrains make entire work a waste since prometheus does not care about cmsweb auth x509 schema and all metrics should be accessible through normal localhost request. Therefore, unless I miss something or unless you tell me where to change REST server code to allow http requests from localhost I don't see a point how it can be used because REST server requires cmsweb x509 authentication for all exposed methods.

todor-ivanov (approved):
This version seems to properly incorporate both the initial idea and Alan's requirements. Thanks @vkuznet and @amaltaro. 

Valya, just a minor confirmation here. Those lines, you did test them for real with a Prometheus instance, right? Or it was just the code from the WMcore side that you tested?

vkuznet:
Todor, I tested all changes with TestMetricsServer, the code I provided in original PR content or I put it in this thread, see https://github.com/dmwm/WMCore/pull/9940#issuecomment-702801325

But I also commented that I can't test it properly with WMCore services since there is no localhost access to them and the only access it through external frontend which requires X509 authentication, e.g. `https://alancc7-cloud1.cern.ch/reqmgr2/data/metrics`. The URL works with curl and my certificates, but since Prometheus does not use x509 authentication I can't use/test it. And, call to `http://localhost:8246/reqmgr2/data/metrics` is forbidden on the host where it runs.

todor-ivanov:
Thanks @vkuznet !

amaltaro:
@vkuznet Valentin, trying to move forward with this PR. Are you able to talk to Imran and try to figure out how the ServerMonitor manages to make localhost calls to every single service under CMSWEB?

amaltaro:
@vkuznet Valentin, I think these localhost queries are made with some specific http headers. Have you figured that out or have you found a way to make it working with prometheus?

vkuznet:
Alan, the [ServerMonitor](https://github.com/dmwm/deployment/blob/master/admin/ServerMonitor#L27) and any calls to localhost of WMCore services requires proper hmac file and CMS headers. This is constrain of WMCore REST code, see [Auth.py](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Auth.py).  WIthout CMS headers all calls to WMCore REST based services is prohibited. As such Prometheus can't talk to local host of the service using /metrics end-point since it does not have any knowledge of our authz schema. And, I doubt that we should modify prometheus to handle CMS headers. I rather want to see exception in WMCore REST  implementation which will allow calls from localhost.

vkuznet:
test this please

vkuznet:
@amaltaro I revisited this PR and adjusted it to pass unit tests, etc. It still shows plenty of pylint errors in [src/python/WMCore/REST/Server.py]() but I doubt I should fix them The REST server codebase is outside the scope of this PR and since it has (by itself) so many different issues with pylint check I don't want to fix them here.

I would appreciate if you finally, after two years :), will review this PR as we plan to include it as part of https://github.com/dmwm/WMCore/issues/9939 which we claim on Q2/22, and let me know if any other action items we need on this.

amaltaro:
@vkuznet thanks for ping'in on this.
I left a couple of comments that we need to update in the code, and this will allow us to see how unit tests behave after such a long time. Please make sure to keep src/ changes separated from test.

Once you make this new commit, can you please rebase this branch to see the actual changes? Thanks

amaltaro (commented):
My comment above was made at the same time as the review request, so please see my feedback above.

test/python/Utils_t/CPMetrics_t.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        return
+
+    def testFlattenStats(self):
+        """
+        Test the flattenStats function
+        """
+        data = flattenStats(self.testData)
+        expect = {"cherrypy_app_bytes_read_request":0.0, "cherrypy_http_server_a":1}
+        self.assertEqual(data, expect)
```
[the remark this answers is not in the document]
vkuznet: fixed

vkuznet:
@amaltaro , I rebased the code. Please proceed with review.

src/python/Utils/CPMetrics.py
```diff
@@ -0,0 +1,267 @@
...
+cherrypy_app_total_bytes_written 0
+# HELP cherrypy_http_server_read_throughput
+# TYPE cherrypy_http_server_read_throughput counter
+cherrypy_http_server_read_throughput -1
+# HELP cherrypy_http_server_accepts
+# TYPE cherrypy_http_server_accepts counter
+cherrypy_http_server_accepts 0
+"""
+# python3 compatibility
+from __future__ import division
```
amaltaro, line 187: Sorry, I missed this. We should target python3 only, so please remove this import and make sure that division is behaving properly without it.

test/python/Utils_t/CPMetrics_t.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+"""
+Unittests for CPMetrics functions
+"""
+
+from __future__ import division, print_function
```
amaltaro, line 6: Same comment here. We no longer require such compatibility changes.

src/python/Utils/CPMetrics.py
```diff
@@ -0,0 +1,267 @@
...
+
+# system modules
+import json
+
+
+def promMetrics(data, exporter_name):
```
amaltaro, line 193: please stick to lower camel case for variable names.

amaltaro (changes requested):
Thanks for these changes, Valentin. I left a few comments along the code (to remove py2/py3 compatibility import).

Before this can go through, please:
* check the pylint and pycodestyle for the new modules. Those should be as clean as possible.
* unit tests are likely unstable. So we can just trigger the tests again to see if they succeed
* commits need to be correctly squashed

src/python/Utils/CPMetrics.py
vkuznet: done

src/python/Utils/CPMetrics.py
vkuznet: done

test/python/Utils_t/CPMetrics_t.py
vkuznet: done

vkuznet:
Alan, pylint score is 10 for new code and its unit test. All others are old code which I don't want to touch in this PR. The unit tests are unstable. Please review once you'll get a chance.

amaltaro (approved):
This looks good to me.
Before we get it merged, can you please deploy these changes to your test9 cluster and run a basic curl test of the new REST API?

vkuznet:
@amaltaro , after testing on test9 with curl I found that additional changes to support `bytes` I/O is required (new python CP server should return bytes for its end-points). So I added separte commit https://github.com/dmwm/WMCore/pull/9940/commits/3813d7eb6def6cca6a859f81946329ef28354e01 Please review this commit and then I can squash it.

src/python/Utils/CPMetrics.py
```diff
@@ -237,6 +237,8 @@ def flattenStats(cpdata):
     "Flatten cherrypy stats to make them suitable for monitoring"
     if isinstance(cpdata, str):
         cpdata = json.loads(cpdata)
```
amaltaro, line 239: To load this json you need to have unicode data. I would suggest to use this utilitarian method instead:
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/Utilities.py#L203

Can you please replace this if/else by the following line:
```
cpdata = json.loads(decodeBytesToUnicode(cpdata))
```
?

amaltaro (changes requested):
Valentin, please  see my partial review on your latest changes.

src/python/Utils/CPMetrics.py
vkuznet: done

src/python/WMCore/REST/Server.py
```diff
@@ -374,7 +374,8 @@ def metrics(self):
         """
         Return CherryPy stats following the prometheus metrics structure
         """
-        return promMetrics(cpstats.StatsPage().data(), self.app.appname)
+        metrics = promMetrics(cpstats.StatsPage().data(), self.app.appname)
+        return bytes(merics, 'utf-8')
```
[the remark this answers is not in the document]
vkuznet: done

vkuznet:
Alan, I made necessary changes, please let me know.

vkuznet:
Now commits are squashed too.

amaltaro:
The Jenkins CI report is your friend, please take advantage of it ;)
```
Unit tests added
    Utils_t.CPMetrics_t.CPMetricsTests:testFlattenStats was added with status error. Must be fixed
    Utils_t.CPMetrics_t.CPMetricsTests:testPromMetrics was added with status error. Must be fixed
```

Once you are done with code changes, please run another test in your central services setup testing reqmgr2 endpoints like: stats, metrics and info. Please let me know how that goes.

vkuznet:
Alan, test is fixed (adding encodeUnicodetoBytes still requires a check) and code is tested on test9, e.g.
```
scurl -s https://cmsweb-test9.cern.ch/ms-output/data/metrics | head
# HELP ms-output_cherrypy_http_server_accepts
# TYPE ms-output_cherrypy_http_server_accepts counter
ms-output_cherrypy_http_server_accepts 0
# HELP ms-output_cherrypy_http_server_accepts_sec
# TYPE ms-output_cherrypy_http_server_accepts_sec gauge
ms-output_cherrypy_http_server_accepts_sec 0.0
# HELP ms-output_cherrypy_http_server_bytes_read
# TYPE ms-output_cherrypy_http_server_bytes_read counter
ms-output_cherrypy_http_server_bytes_read -1
# HELP ms-output_cherrypy_http_server_bytes_written
```

amaltaro (approved):
Looks good to me, thanks!

mapellidario:
From a first quick pick, this is beautiful! We had the idea of doing something similar and will likely steal from this in the future :)
Fyi: @novicecpp

vkuznet:
FYI @amaltaro , @mapellidario , @novicecpp I made corresponding monitoring ticket: https://its.cern.ch/jira/browse/CMSMONIT-514 to request CherryPy dashboard. Please follow up monitoring ticket and request your dashboard for your python based service.

PR #9936, opened 2020-09-24
Fix Rucio data location method; support dropping Tape RSEs

amaltaro:
Fixes #9932
Fixes #9937

#### Status
ready

#### Description
Created a new logic for `getDataLockedAndAvailable`, deprecating the old method. New logic only depends on the rule ids (given a set of filter when listing replication rules) and on the output of dataset_locks that match one of those rules. The docstrings have a better explanation of that. Added a warning message to the old methods to be soon deprecated.

In addition to that, allow `returnTape` argument to be passed such that `_Tape` locations can be skipped from the current input location.

Last but not least, there is a new wrapper API which receives a child DID (usually a block), and lists all the rules for the parent DID(s).

There are a few unit tests to test those functionalities. Those are marked as integration right now because we are not able to authenticate to that account with the jenkins DN.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
Complement to: https://github.com/dmwm/WMCore/pull/9889

#### External dependencies / deployment changes
none

PR #9942, opened 2020-09-28
Delete blocks using using getParentContainerRules and getReplicaInfoForBlocks

### LINKED ISSUE #9623: Implement the T0 block deletion logic in RucioInjector

germanfgv:
Fixes #9623

#### Status
In development

#### Description
<Description of the changes proposed.>

YES

#### Related PRs
This builds upon PR #9936

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -370,7 +370,53 @@ class RucioInjectorPoller(BaseWorkerThread):
           If no => do nothing (check again next cycle)
         """
         # FIXME: figure out the proper logic for rule block deletion
+        # TODO: Update metadata after deleting block rules
         logging.info("Starting deleteBlocks methods --> IMPLEMENT-ME!!!")
+        logging.info("Starting deleteBlocks methods --> I'M ON IT...")
+
+        #Get list of blocks that can be deleted
+        blockDict = self.findDeletableBlocks.execute(transaction=False)
+
+        if not blockDict:
+            logging.info("No blocks were found")
+            return
+
+        deletableBlocks = []
+		# Obtain list of target RSE for each block
+        for blockName in blockDict:
+            rules = self.rucio.getParentContainerRules(name=blockName, scope=self.scope, account=self.rucioAcct)
+
+            #
+            requestedRSEs = []
+            requestedCopies = 0
+            for rule in rules:
+                if rule['account'] == self.rucioAcct:
+                    requestedCopies += rule['copies']
+                    requestedRSEs.extend(self.rucio.evaluateRSEExpression(rule['rse_expression'], useCache=True, returnTape=True))
+
+            #Get RSE in which the block is available
+            availableRSEs = self.rucio.getReplicaInfoForBlocks(block=blockName)
+
+            #Count how many of the copies were requested by this account
+            actualCopies = len([copie for copie in availableRSEs if copie in requestedRSEs])
+
+            #If enough requested copies are available, block can be deleted
+            if requestedCopies <= actualCopies:
+                deletableBlocks.append(blockName)
+
+
+
+        for block in deletableBlocks:
+            rules = self.rucio.listDataRules(block, self.scope)
+            for rule in rules:
+                if rule['account'] == self.rucioAcct:
+                    if self.rucio.deleteRule(rule['id']):
+                        logging.info("Successfully deleted block: %s.", block)
+                    else:
+                        logging.error("Failed to delete block: %s. Will retry again later.", block)
+
+
+        return
 
     # TODO: this will likely go away once the phedex to rucio migration is over
     def _isContainerTierAllowed(self, containerName, checkRulesList=True):
--- a/bin/adhoc-scripts/checkStuckLQE.py
+++ b/bin/adhoc-scripts/checkStuckLQE.py
@@ -53,6 +53,7 @@ def commonDataLocation(element):
     :param element: workqueue element object
     :return: list with the common **data** location
     """
+    commonLoc = set()
     if element['PileupData']:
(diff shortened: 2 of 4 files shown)


src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,7 +370,53 @@ def deleteBlocks(self):
           If no => do nothing (check again next cycle)
         """
         # FIXME: figure out the proper logic for rule block deletion
+        # TODO: Update metadata after deleting block rules
```
amaltaro, line 373: You want to update the block metadata once a rule deletion happens? Is it somehow a requirement?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,7 +370,53 @@ def deleteBlocks(self):
...
+        blockDict = self.findDeletableBlocks.execute(transaction=False)
+
+        if not blockDict:
+            logging.info("No blocks were found")
+            return
+
+        deletableBlocks = []
+		# Obtain list of target RSE for each block
+        for blockName in blockDict:
+            rules = self.rucio.getParentContainerRules(name=blockName, scope=self.scope, account=self.rucioAcct)
```
amaltaro, line 387: For this block of code, I think we could first map the block to a container name, perhaps something like:
```
{"container_name_A": ["block_A", "block_B"], "container_name_B": [], ...}
```
such that we resolve the parent rules only once per cycle, having it only 1 or 1M blocks under the same container.

Even a better improvement would be to keep a containerRules cache (<= 1 hour), which contains parent container names and a list of replication rules. This will only be useful if we run this deleteBlocks method a few times every hour though.

Last but not least, if you provide an `account` keyword arg, then it *should* return only rules matching that account. Which means, your `if` statement on L393 isn't really necessary.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,7 +370,53 @@ def deleteBlocks(self):
...
+            #
+            requestedRSEs = []
+            requestedCopies = 0
+            for rule in rules:
+                if rule['account'] == self.rucioAcct:
+                    requestedCopies += rule['copies']
+                    requestedRSEs.extend(self.rucio.evaluateRSEExpression(rule['rse_expression'], useCache=True, returnTape=True))
+
+            #Get RSE in which the block is available
+            availableRSEs = self.rucio.getReplicaInfoForBlocks(block=blockName)
```
amaltaro, line 398: If you map blocks to a container, as suggested in my comment above, you could then pass a list of those blocks (within the same container) to this method. Such that it makes a single HTTP requests for a bunch of blocks.
IF this chunk of blocks might become too big though (> 200 blocks), then we might need to apply some slicing on the list.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,7 +370,53 @@ def deleteBlocks(self):
...
+
+            #Count how many of the copies were requested by this account
+            actualCopies = len([copie for copie in availableRSEs if copie in requestedRSEs])
```
amaltaro, line 401: Perhaps something like
```
actualCopies = len(set(requestedRSEs) & set(availableRSEs))
```
?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,7 +370,53 @@ def deleteBlocks(self):
...
+
+            #If enough requested copies are available, block can be deleted
+            if requestedCopies <= actualCopies:
+                deletableBlocks.append(blockName)
+
+
+
+        for block in deletableBlocks:
+            rules = self.rucio.listDataRules(block, self.scope)
```
amaltaro, line 410: Not for your pull request. But I've been considering to change the Rucio client API we use within `listDataRules`, we could use `list_replication_rules`, which would take some filters and give us a clean answer. Not sure whether there is any performance penalty though...

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,7 +370,53 @@ def deleteBlocks(self):
...
+            for rule in rules:
+                if rule['account'] == self.rucioAcct:
+                    if self.rucio.deleteRule(rule['id']):
+                        logging.info("Successfully deleted block: %s.", block)
```
amaltaro, line 414: I guess we also want to persist a successful deletion in the oracle database, right?

amaltaro (changes requested):
Thanks for sharing your work, German. I've made a few comments along the code that will need your attention.
Can you please also rebase this PR such that it clears itself from the other already merged PR?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I'll start adding the container-blocks mapping. I think deleteBlocks will probably run several times and hour, so caching container rules also seems like a good idea. 

I'll delete the if statement. It was necessary before you added `getParentContainerRules` and I forgot to remove it afterwards.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: yes, this makes more sense.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Yes, this is what I meant before with "metadata". Maybe I was using the wrong term.

PR #9944, opened 2020-09-28
Drop Tape RSEs from the MSTransferor input location

### LINKED ISSUE #9943: Drop Tape RSEs from the MSTransferor input data placement logic
**Impact of the bug**
ReqMgr2MS - MSTransferor

**Describe the bug**
There are two methods performing basically the same logicPSN to PNN logic, and _Tape endpoints are only filtered on the method no longer used (_getFinalPNNs). We need to decide which method to use and make sure _Tape RSE endpoints will not be considered as a valid location.

**How to reproduce it**
none

**Expected behavior**
Properly drop _Tape RSEs from the current data location (and from possible destinations for new input data).

none

amaltaro:
Fixes #9943 

#### Status
ready

#### Description
Do not use `_Tape` RSEs as a valid current location; and don't use them as a destination for input data placement either.

yes

#### Related PRs
Bugfix to: https://github.com/dmwm/WMCore/pull/9759

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSTransferor.py
+++ b/src/python/WMCore/MicroService/Unified/MSTransferor.py
@@ -907,30 +907,6 @@ class MSTransferor(MSCore):
             self.logger.info("Have whole PU dataset: %s (%s GB)", dsetName, gigaBytes(dsetSize))
             yield blockList, dsetSize, idx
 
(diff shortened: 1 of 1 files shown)


amaltaro:
Tests went fine, merging.

PR #9947, opened 2020-09-29
Flag to disable PhEDExInjector; setup to enable all datatiers in RucioInjector

### LINKED ISSUE #9946: Add flag to disable PhEDExInjector; use empty listTiersToInject to inject all tiers into Rucio
**Impact of the new feature**
WMAgent

When we decide to fully enable RucioInjector for all datatiers, it looks like we would have to list all of them in the agent configuration, which then gets read by the RucioInjector component via:
```
        self.listTiersToInject = config.RucioInjector.listTiersToInject
```

So I suggest to set this `listTiersToInject` parameter to an empty list, together with `RucioInjector.enabled` it should be treated as if we wanted to inject all the datatiers in Rucio.

A few things:
* make a flag to disable PhEDExInjector (one could also remove the component section from the config file and restart the agent)
* change RucioInjectorPoller to inject any datatiers IF:
  * the component is set to enabled; and
  * the list of tiers to inject is empty

none

**Additional context**
none

amaltaro:
Fixes #9946

#### Status
ready

#### Description
This PR provides:
* a PhEDExInjector parameter to disable the component: `enabled` (default to `True`). If the component is disabled, it keeps running but not doing anything (returning on the 1st line of the algorithm method)
* change how `listTiersToInject` parameter is handled in RucioInjector. If the component is enabled, and that list is empty, then it means the component is supposed to inject every single data produced by the agent against the Rucio system.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -24,20 +24,23 @@ from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 
 
 ### TODO: remove this function once PhEDExInjector is out of the game
-def filterDataByTier(rawData, allowedTiers):
+def filterDataByTier(rawData, listTiersToInject):
     """
     This function will receive data - in the same format as returned from
     the DAO - and it will pop out anything that the component is not meant
     to inject into Rucio.
     :param rawData: the large dict of location/container/block/files
-    :param allowedTiers: a list of datatiers that we want to inject
+    :param listTiersToInject: a list of datatiers that we want to inject
     :return: the same dictionary as in the input, but without dataset structs
              for datatiers that we do not want to be processed by this component.
     """
+    if not listTiersToInject:
+        # then we are actually allowed to insert anything into Rucio
+        return rawData
     for location in rawData:
         for container in list(rawData[location]):
             endTier = container.rsplit('/', 1)[1]
-            if endTier not in allowedTiers:
+            if endTier not in listTiersToInject:
                 logging.debug("Container %s not meant to be injected by RucioInjector", container)
                 rawData[location].pop(container)
     return rawData
@@ -107,8 +110,11 @@ class RucioInjectorPoller(BaseWorkerThread):
         self.testRSEs = config.RucioInjector.RSEPostfix
         self.filesToRecover = []
 
-        logging.info("Component configured to only inject data for data tiers: %s",
-                     self.listTiersToInject)
+        if not self.listTiersToInject:
+            logging.info("Component configured to inject all the data tiers")
+        else:
+            logging.info("Component configured to only inject data for data tiers: %s",
+                         self.listTiersToInject)
         logging.info("Component configured to skip container rule creation for data tiers: %s",
                      self.skipRulesForTiers)
         logging.info("Component configured to create block rules: %s", self.createBlockRules)
@@ -246,6 +252,8 @@ class RucioInjectorPoller(BaseWorkerThread):
         be handled by this component, thus block-level rule creation as well
         :return: True if the component can proceed with this block, False otherwise
         """
+        if not self.listTiersToInject:
+            return True
         endBlock = blockName.rsplit('/', 1)[1]
         endTier = endBlock.split('#')[0]
         if endTier not in self.listTiersToInject:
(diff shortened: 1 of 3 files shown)


amaltaro:
This PR seems to be working fine. @todor-ivanov can you please review just in case I missed something?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -34,6 +34,9 @@ def filterDataByTier(rawData, allowedTiers):
     :return: the same dictionary as in the input, but without dataset structs
              for datatiers that we do not want to be processed by this component.
     """
+    if not allowedTiers:
```
todor-ivanov, line 37: I do not like this magical silent metamorphosis of the value name from `listTiersToInject` to `allowedTiers`. Those are linked only on line 171 in the current commit: [1]. This makes the code hard to track. I would suggest this value to remain consistent with it's configuration name [2] wherever it is passed to a function so that it's meaning remains unambiguous throughout the code.

[1]
https://github.com/dmwm/WMCore/blob/a3624ea191c980555b0464aae68866c1bf231c35/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py#L171

[2]
https://github.com/dmwm/WMCore/blob/a3624ea191c980555b0464aae68866c1bf231c35/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py#L88

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -246,6 +252,8 @@ def _isBlockTierAllowed(self, blockName):
         be handled by this component, thus block-level rule creation as well
         :return: True if the component can proceed with this block, False otherwise
         """
+        if not self.listTiersToInject:
```
todor-ivanov, line 255: This will work only if `self.listTiersToInject` is set. Meaning that we should guarantee a default value of `[]` in the constructor. Otherwise it will raise a `NameError` exception.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -385,7 +393,7 @@ def _isContainerTierAllowed(self, containerName, checkRulesList=True):
         :return: True if the component can proceed with this container, False otherwise
         """
         endTier = containerName.rsplit('/', 1)[1]
-        if endTier not in self.listTiersToInject:
+        if self.listTiersToInject and endTier not in self.listTiersToInject:
```
todor-ivanov, line 396: Same as above.

todor-ivanov (changes requested):
Alan, I made a quick look at the PR and I think there are few things you need to double check. Please find the comments inline.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: It's set in the init method

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Honestly, I do not know what's the most pythonic way to pass these variables. I know that pylint complains if we pass a variable named `blah` to a function with a parameter named the same (thus `blah`).

It can be easier to track indeed, I'll update it.

amaltaro:
Thanks for the review Todor. Code has been updated as per your suggestion.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
todor-ivanov: Thanks

todor-ivanov (approved):
Alan, I think things are fine. I made a clarification inline about what I mean by 'default value'. But anyway it is up to you to decide. For me it is fine even like that. Because I think the constructor will break anyway in case the value is not set in the config, so there will be another error before we hit `NamError` exception in this check.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Default value is already set in the WMAgent configuration template:
https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py#L373

if someone screws up that variable (either by removing it from the config or something else), the component will crash, as any of the other component attributes.

This attribute will no longer be necessary when we completely migrate to Rucio. Having said that, I think changing this PR again isn't worthwhile.

amaltaro:
Thanks for the review, Todor.

PR #9951, opened 2020-09-29
MSMonitor: deal with ZeroDivision exception for large rules in INJECT state

### LINKED ISSUE #9950: MSMonitor fails to calculate rule progress: division by zero
**Impact of the bug**
ReqMgr2MS - MSMonitor

**Describe the bug**
There is one specific workflow which we hit this exception:
```
2020-09-29 23:54:53,133:ERROR:MSMonitor: Unknown exception checking workflow amaltaro_TaskChain_Prod_Sept2020_Val_200929_232610_2012. Error: division by zero
```

apparently the locks_* parameters are all 0. It's likely related to the current state of the rule `state=INJECT`, because this is a very large container (500TB).

**How to reproduce it**
Make a container rule for a very large container (perhaps > 10k blocks)

**Expected behavior**
If we cannot calculate the rule progress, then we should set its progress to 0%.
There might be room for improvement on the error message as well.

Error from the service log:
```
2020-09-29 23:44:52,078:DEBUG:MSMonitor: Rule ID: 4a2f8d276dbc40a98f2ea2a3582f2076, DID: /Cosmics/Commissioning2015-v1/RAW#487c3a70-b47d-11e4-8732-02163e00d625, state: OK, grouping: DATASET, rse_expression: T1_US_FNAL_Disk
2020-09-29 23:44:52,078:DEBUG:MSMonitor: Checking transfers for: amaltaro_TaskChain_Prod_Sept2020_Val_200929_232610_2012
2020-09-29 23:44:52,098:ERROR:MSMonitor: Unknown exception checking workflow amaltaro_TaskChain_Prod_Sept2020_Val_200929_232610_2012. Error: division by zero
```

amaltaro:
Fixes #9950 

#### Status
ready

#### Description
For very large container (and maybe for not so large too), there is a delay on the Rucio side to calculate the amount of locks expected by the rule. If MSMonitor checks the progress of such rules in the very initial state, it could be that we hit a ZeroDivisionError exception. In such cases, consider the rule completion as 0% and print the whole rule dictionary.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSMonitor.py
+++ b/src/python/WMCore/MicroService/Unified/MSMonitor.py
@@ -165,7 +165,7 @@ class MSMonitor(MSCore):
                 doc['lastUpdate'] = tstamp
             except Exception as exc:
                 msg = "Unknown exception checking workflow %s. Error: %s"
-                self.logger.error(msg, doc['workflowName'], str(exc))
+                self.logger.exception(msg, doc['workflowName'], str(exc))
                 skippedWorkflows.append(doc['workflowName'])
         return skippedWorkflows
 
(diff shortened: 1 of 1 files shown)


todor-ivanov (approved):
This one looks pretty good.

amaltaro:
Thanks Todor!

PR #9954, opened 2020-09-30
Use dataset_replicas for the container data location

### LINKED ISSUE #9953: Optimize Rucio-based dataset/pileup data location
**Impact of the new feature**
Global and local workqueue

With my latest changes to the Rucio logic for data location here:
https://github.com/dmwm/WMCore/pull/9936

the code was streamlined, but at the same time it became a much more expensive operation, given that now we compare every single block lock against the list of container-level rules locking that data.
  
If we want to make it simple, then we should look only at the container-level rules without checking where blocks (rucio datasets) are locked, thus avoiding to loop over all the blocks. The problem with this approach would be if we have rules made against multiple RSE, the only way to know where our data is locked is either checking the locks with `get_dataset_locks`, or listing the replicas location with `list_dataset_replicas`, but the latter does not contain the rule id (even though it accepts requests in bulk...). For MSTransferor-made rules, it wouldn't be a problem.

Another option - that might become a bug in the future - would be to use only the single RSE rules (well, len(rses) == len(copies)), this would be the easiest option and should work for the near future....

If the performance with `list_dataset_replicas_bulk` is acceptable, then we might compare those multi RSE locations agains the locations returned from this method, making then the final container location.

**Additional context**
Here is a log from global workqueue to resolve a ~25k blocks pileup dataset (~500TB):
```
2020-09-30 08:43:59,758:INFO:DataLocationMapper:Found 1 unique pileup data to update location
2020-09-30 08:43:59,786:INFO:DataLocationMapper:/Neutrino_E-10_gun/RunIISpring15PrePremix-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v2-v2/GEN-SIM-DIGI-RAW, setting location to: [u'T1_US_FNAL']
2020-09-30 08:43:59,786:INFO:DataLocationMapper:Updating 1 elements for Pileup location update
2020-09-30 08:43:59,873:INFO:LocationUpdateTask:LocationUpdateTask took 462.979 secs and updated 1 non-unique elements
```

amaltaro:
Fixes #9953

#### Status
not-tested

#### Description
Remake once again the container data location logic to use `list_dataset_replicas_bulk` instead of `get_dataset_locks` API,
which gives us a speed up by a factor of 3 or 4 (when there are multiple RSE rules), still very expensive though, went down from ~450secs to ~120secs for a large premix container (25k blocks / 125k files).
In case there are no multiple RSE rules, the speed up is by two orders of magnitude.

The possible flaw on this logic is that we no longer match dataset_locks rules against the container level rules; instead we match dataset_replicas RSEs against the RSEs from multiple RSE container-level rules. In other words, there is a chance that we use a given RSE as data location, but it's actually not locked by WMCore... I guess.

The docstrings explains every single detail of this logic.

Just a note, we might have to make the same changes for the pileup data location from WMAgent, which right now uses a different Rucio wrapper method and still relies on the heavy dataset_locks

yes

#### Related PRs
Change the logic implemented here: https://github.com/dmwm/WMCore/pull/9936

none

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -14,6 +14,8 @@ from rucio.client import Client
 from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied, DuplicateRule,
                                     DataIdentifierAlreadyExists, DuplicateContent, InvalidRSEExpression,
                                     UnsupportedOperation, FileAlreadyExists, RuleNotFound)
+
+from Utils.IteratorTools import grouper
 from Utils.MemoryCache import MemoryCache
 from WMCore.WMException import WMException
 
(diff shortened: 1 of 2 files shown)


nsmith-:
So the main bottleneck is having to issue `get_dataset_locks` for each block in a large CMS dataset-container. And the way around appears to be not to cross-check the locks, so the effect is that if a replica _could_ match a rule (i.e. its RSE is in the RSE expression of the rule) it is considered protected and useable. There is some possibility that the replica is not locked by the rule.

amaltaro:
Exactly! I guess I'll leave this PR open such that we can keep discussing any other possibilities over the next days (like implementing the bulk API on the server; making concurrent requests with threading, etc). And I will create another PR with only the step-1 logic, which means, evaluating single RSE rules and if possible, skipping the get_dataset_locks calls. Hopefully it won't be slow enough to impact the overall WM performance.

nsmith-:
I had hoped that using [list_replica_locks](https://rucio.readthedocs.io/en/latest/api/rule.html#rucio.client.ruleclient.RuleClient.list_replica_locks), despite being file-level, would work better since you instantly get the distribution of replicas for a given rule, and we already have effective filtering for the rules. Sadly, I have yet to have this API return successfully for the very large (125k file) dataset you are using in your test.

PR #9955, opened 2020-10-01
Avoid resolving container location with the dataset_locks, whenever possible

### LINKED ISSUE #9953: Optimize Rucio-based dataset/pileup data location

amaltaro:
Fixes #9953
Superseeds #9954 (even though that might still be useful)

#### Status
ready

#### Description
The docstring provides a very detailed explanation.
In short, this PR changes the logic for container data location such that it tries to figure the container location by simply looking at the rules, in case there are rules where len(copies) == len(rses), we use that RSE as a container location.
If rules are made against many RSEs, then we need to list all the blocks in the container and ask for their dataset_locks, which is the heavy operation we must avoid as much as possible.

Important note: this will change the current behaviour of pileup input, in the sense that only container-wide location will be considered. Thus, transferring a single block to a given location will not be enough to unstuck a GQE waiting for the pileup to be available!

yes

#### Related PRs
First attempt to fix it in (still under scrutiny): https://github.com/dmwm/WMCore/pull/9954

none

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -1071,11 +1071,25 @@ class Rucio(object):
         This method retrieves all the locations where a given container DID is
         currently available and locked (note that, by default, it will not
         return any Tape RSEs). The logic is as follows:
-          1. list the container-level replication rules
-          2. list all the blocks in the container
-          3. for each block, check where it's are locked and available (state=OK),
-             matching one of the replication rule ids discovered in the previous steps
-          4. the input `grouping` specified decides how the list of final RSEs are built
+          1. for each container-level replication rule, check
+            i. if the rule state=OK and if the number of copies is equals to
+             the number of RSEs. If so, that is one of the container locations
+            ii. elif the rule state=OK, then consider that rule id (and its RSEs) to
+             be evaluated against the dataset locks
+            iii. otherwise, don't use the container rule
+          2. if there were no rules with multiple RSEs, then return the RSEs from step 1
+          3. else, retrieve all the blocks in the container and build a block-based dictionary
+            by listing all the dataset_locks for all the blocks:
+            i. if the lock state=OK and the rule_id belongs to one of those multiple RSE locks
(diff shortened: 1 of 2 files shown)


amaltaro:
this is looking good now, even though I see a possible degradation on the Rucio server now :-)

PR #9959, opened 2020-10-02
Delete blocks using container-block mapping

### LINKED ISSUE #9623: Implement the T0 block deletion logic in RucioInjector

germanfgv:
Fixes #9623

**Status**
In development

**Description**
Remake the method so now a container dictionary is used to keep track of blocks and container-level rules, so listDataRules is never used twice for the same container. It also allows us to use getReplicaInfoForBlocks() in bulk. After this, the method checks if the number of available replicas of each block is at least as many as requested by current account. If that's the case, the block can be deleted. listDataRules is then used to retrieve block-level rules that are going to be deleted. Finally, database is updated to reflect the blocks that were deleted.

The method does not use cache to store container replication rules because this method is intended to be executed around once every hour.

It would be very useful to be able to use listDataRules in bulk. Right now this method makes an HTTP request for each container and each deleted block.

**Is it backward compatible (if not, which system it affects?)**
YES

**Related PRs**
This builds upon PR #9936 and change the logic PR #9942 
Requires `listDataRules` changes applied on https://github.com/dmwm/WMCore/pull/9999 (thus a backport to the agent branch is required)

**External dependencies / deployment changes**
None

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -102,7 +102,8 @@ class RucioInjectorPoller(BaseWorkerThread):
         self.rucio = Rucio(acct=self.rucioAcct,
                            hostUrl=config.RucioInjector.rucioUrl,
                            authUrl=config.RucioInjector.rucioAuthUrl,
-                           configDict={'logger': self.logger})
+                           configDict={'logger': self.logger,
+                                       'phedexCompatible': False})
 
         # metadata dictionary information to be added to block/container rules
         # cannot be a python dictionary, but a JSON string instead
@@ -379,15 +380,85 @@ class RucioInjectorPoller(BaseWorkerThread):
         """
         _deleteBlocks_
         Find deletable blocks, then decide if to delete based on:
-        Is there an active subscription for dataset or block ?
-          If yes => set deleted=2
-          If no => next check
         Has transfer to all destinations finished ?
-          If yes => request block deletion, approve request, set deleted=1
+          If yes => Delete rules associated with the block, set deleted=1
           If no => do nothing (check again next cycle)
         """
-        # FIXME: figure out the proper logic for rule block deletion
-        logging.info("Starting deleteBlocks methods --> IMPLEMENT-ME!!!")
+        logging.info("Checking if there are block rules to be deleted...")
+
+        # Get list of blocks that can be deleted
+        blockDict = self.findDeletableBlocks.execute(transaction=False)
+
+        if not blockDict:
+            logging.info("No candidate blocks found for rule deletion")
+            return
+
+        logging.info("Found %d candidate blocks for rule deletion", len(blockDict))
+
+        blocksToDelete = []
+        containerDict = {}
+        # Populate containerDict, assigning each block to its correspondant container
+        for blockName in blockDict:
+            container = blockDict[blockName]['dataset']
+            # If the container is not in the dictionary, create a new entry for it
+            if container not in containerDict:
+                # Set of sites to which the container needs to be transferred
+                sites = blockDict[blockName]['sites']
+                containerDict[container] = {'blocks': [], 'rse': sites}
+            containerDict[container]['blocks'].append(blockName)
+
+        for contName in containerDict:
+            cont = containerDict[contName]
+
+            # Checks if the container is not requested in any sites.
+            # This should never be triggered, but better safe than sorry
+            if not cont['rse']:
+                logging.warning("No rules for container: %s. Its blocks won't be deleted.", contName)
+                continue
+
+            try:
+                # Get RSE in which each block is available
+                availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
+            except Exception as exc:
+                msg = "Failed to get replica info for blocks in container: %s.\n" % contName
+                msg += "Will retry again in the next cycle. Error: %s" % str(exc)
+                logging.error(msg)
+                continue
+
+            for blockRSEs in availableRSEs:
+                # If block is available at every RSE its container needs to be transferred, the block can be deleted
+                blockSites = set(blockRSEs['replica'])
+                if cont['rse'].issubset(blockSites):
+                    blocksToDelete.append(blockRSEs['name'])
+
+        # Delete agent created rules locking the block
+        binds = []
+        logging.info("Going to delete %d block rules", len(blocksToDelete))
+        for block in blocksToDelete:
+            try:
+                rules = self.rucio.listDataRules(block, scope=self.scope, account=self.rucioAcct)
+            except WMRucioException as exc:
+                logging.warning("Unable to retrieve replication rules for block: %s. Will retry in the next cycle.", block)
+            else:
+                if not rules:
+                    logging.info("Block rule for: %s has been deleted by previous cycles", block)
+                    binds.append({'DELETED': 1, 'BLOCKNAME': block})
+                    continue
+                for rule in rules:
+                    deletedRules = 0
+                    if self.rucio.deleteRule(rule['id']):
+                        logging.info("Successfully deleted rule: %s, for block %s.", rule['id'], block)
+                        deletedRules += 1
+                    else:
+                        logging.warning("Failed to delete rule: %s, for block %s. Will retry in the next cycle.", rule['id'], block)
+                if deletedRules == len(rules):
+                    binds.append({'DELETED': 1, 'BLOCKNAME': block})
+                    logging.info("Successfully deleted all rules for block %s.", block)
+
+
+        self.markBlocksDeleted.execute(binds)
+        logging.info("Marked %d blocks as deleted in the database", len(binds))
+        return
 
     # TODO: this will likely go away once the phedex to rucio migration is over
     def _isContainerTierAllowed(self, containerName, checkRulesList=True):
--- a/src/python/WMComponent/RucioInjector/Database/MySQL/MarkBlocksDeleted.py
+++ b/src/python/WMComponent/RucioInjector/Database/MySQL/MarkBlocksDeleted.py
@@ -20,6 +20,10 @@ class MarkBlocksDeleted(DBFormatter):
              """
 
     def execute(self, binds, conn=None, transaction=False):
+
+        if not binds:
+            return
+
         self.dbi.processData(self.sql, binds, conn=conn,
                              transaction=transaction)
 


amaltaro (commented):
@germanfgv German, could you please update the initial description of this PR (according to the template)?
Once you have changes that you would like to be reviewed, please let me know.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -378,7 +378,62 @@ def deleteBlocks(self):
...
+                if rule['account'] == self.rucioAcct:
+                    cont['copies'] += rule['copies']
+                    cont['rse'].extend(self.rucio.evaluateRSEExpression(rule['rse_expression'], useCache=True, returnTape=True))
+
+            #Get RSE in which the block is available
+            availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
+            
+            for rses in availableRSEs:
+                #Count how many of the copies were requested by this account
+                actualCopies = len(set(rses['replica']) & cont['rse'])
```
germanfgv, line 417: This should be `set(rses['replica']) & set(cont['rse']))`. I fixed it during local pylint tests but must have reversed the changes by accident.

germanfgv:
@amaltaro This code is ready for you to review. Please let me know if you have any comments.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+            #Retrieve requested copies and RSEs
+            for rule in rules:
+                if rule['account'] == self.rucioAcct:
+                    cont['copies'] += rule['copies']
+                    cont['rse'].extend(self.rucio.evaluateRSEExpression(rule['rse_expression'], useCache=True, returnTape=True))
+
+            #Get RSE in which the block is available
+            availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
+
+            for rses in availableRSEs:
```
amaltaro, line 406: How about giving a more meanigful name to this `rses` variable, like `blockRSEs`?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+                #Count how many of the copies were requested by this account
+                actualCopies = len(set(rses['replica']) & set(cont['rse']))
+
+                #If enough requested copies are available, block can be deleted
+                if cont['copies'] <= actualCopies:
```
amaltaro, line 411: This would delete blocks with 0 copies.
Can you please remind me again what is the logic we discussed here? Is it that the container needs to have at least one rule and that the block needs to be AVAILABLE on at least one RSE?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+                    deletableBlocks.append(rses['name'])
+
+        #Delete agent ccreated rules locking the block
```
amaltaro, line 414: typo

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+        binds = []
+        for block in deletableBlocks:
```
amaltaro, line 416: Before deleting, log something like
```
logging.info("Going to delete %d blocks", deletableBlocks)
```

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+            rules = self.rucio.listDataRules(block, self.scope)
```
amaltaro, line 417: I think there might be a case where the rule has been properly deleted, but information wasn't persisted on the database. So I would add a protection like:
```
if not rules:
    logging.info("Block rule for: %s has been deleted by previous cycles", block)
    binds.append({'DELETED': 1, 'BLOCKNAME': block})
```

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+            for rule in rules:
+                if rule['account'] == self.rucioAcct:
+                    if self.rucio.deleteRule(rule['id']):
+                        logging.info("Successfully deleted block: %s.", block)
```
amaltaro, line 421: Please be explicit on these logging records, like
```
logging.info("Successfully deleted rule: %s, for block: %s.", rule, block)
```

same for the logging.error line below (and please, make it a warning level instead of error).

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+                        binds.append({'DELETED': 1, 'BLOCKNAME': block})
+                    else:
+                        logging.error("Failed to delete block: %s. Will retry again later.", block)
+        self.markBlocksDeleted.execute(binds)
+
```
amaltaro, line 426:
```suggestion
logging.info("Successfully marked %d blocks as deleted in the database", len(binds))
```

amaltaro (changes requested):
German, that's a good start for this feature. I have made a few comments that require further follow up.
In addition to that, please look at the jenkins report (mostly pep8 I think) and fixed those lines reported matching your changes.

About the `listDataRules`, I think I can change the underlying rucio API used. That would make this implementation slightly easier. However, please do not wait for that in order to start making further changes here.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+          If yes => Delete rules associated with the block, set deleted=1
           If no => do nothing (check again next cycle)
         """
-        # FIXME: figure out the proper logic for rule block deletion
-        logging.info("Starting deleteBlocks methods --> IMPLEMENT-ME!!!")
+        #Get list of blocks that can be deleted
+        blockDict = self.findDeletableBlocks.execute(transaction=False)
+
+        if not blockDict:
+            logging.info("No blocks were found")
```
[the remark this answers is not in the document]
germanfgv: I updated the logging message according to your recommendations

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+            return
+
+        deletableBlocks = []
+        containerDict = {}
+		#Populate containerDict, assigning each block to its correspondant container
```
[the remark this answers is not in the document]
germanfgv: Indentation fixed.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+        for blockName in blockDict:
+            containerList = self.rucio.listParentDIDs(blockName, scope=self.scope)
+            for cont in containerList:
+                if cont['name'] not in containerDict:
+                    containerDict[cont['name']] = {'blocks':[], 'rse':[], 'copies':0}
```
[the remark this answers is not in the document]
germanfgv: Sorry, I was checking Pylint reports but not Pep8. White-spaces where added.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Fixed

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Logging message added.

amaltaro:
@germanfgv German, can you please rebase this PR such that changes merged in
https://github.com/dmwm/WMCore/pull/9963

are considered? Once you rebase it, please update your code to look for DID rules by providing the rucio account as well (one less `if` statement we need to keep in the component).

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,61 @@ def deleteBlocks(self):
...
+
+        if not blockDict:
+            logging.info("No blocks were found")
+            return
+
+        deletableBlocks = []
+        containerDict = {}
+		#Populate containerDict, assigning each block to its correspondant container
+        for blockName in blockDict:
+            containerList = self.rucio.listParentDIDs(blockName, scope=self.scope)
```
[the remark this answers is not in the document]
germanfgv: `findDeletableBlocks` was already returning dataset name. I'm now using that info instead of the Rucio call.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I changed the variable name to the one you suggested

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: We will need either 1 or 2 copies of each block (2 for most RAW data, 1 for other datatiers). I'm (maybe dangerously) assuming that all the necessary container-level rules have been created. If that's the case, cont['copies'] will be equal to the number of copies Tier0 needs.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I added this protection

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I improved logging according to your suggestions

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Logging info added

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: With phedex, we get the destination sites from `dbsbuffer_dataset_subscription.site`. Is there going to be an equivalent column for Rucio?

amaltaro: it should remain the same. I'm planning to make changes as backward compatible as possible

germanfgv: I'll keep using `dbsbuffer_dataset_subscription.site` then. I'll also add a protection and a warning message in case a particular container has no requested copies.

amaltaro:
@germanfgv is it ready for another review? Or are you still working on other items from the first review?

germanfgv:
Hi @amaltaro I think all the initial concerns have been tackled. Is there something else you thinks needs to be modified?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,67 @@ def deleteBlocks(self):
...
+
+        blocksToDelete = []
+        containerDict = {}
+        # Populate containerDict, assigning each block to its correspondant container
+        for blockName in blockDict:
+            container = blockDict[blockName]['dataset']
+            if container not in containerDict:
+                # Set of sites to wich the container is needs to be transferred
+                sites = blockDict[blockName]['sites']
+                containerDict[container] = {'blocks': [], 'rse': sites}
```
amaltaro, line 394: I have two questions on this block of code:
1. is it possible that some blocks are in one set of RSEs, and other blocks are in different RSEs (blocks belonging to the same container)?
2. what is this list of sites returned from the T0 database. Is it meant to be where the block has been placed (via a container rule)?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,67 @@ def deleteBlocks(self):
...
+            if not cont['rse']:
+                logging.warning("No rules for container: %s. It's blocks won't be deleted.", contName)
+                continue
+
+            # Get RSE in which each block is available
+            availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
+
+            for blockRSEs in availableRSEs:
+                # If block is available at every RSE its container needs to be transferred, the block can be deleted
+                if cont['rse'].issubset(blockRSEs['replica']):
```
amaltaro, line 411: Isn't `blockRSEs['replica']` a list here? I guess you'll need to cast it to a set object.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,67 @@ def deleteBlocks(self):
...
+            cont = containerDict[contName]
+
+            # Checks if the container is not requested in any sites.
+            # This should never be triggered, but better safe than sorry
+            if not cont['rse']:
+                logging.warning("No rules for container: %s. It's blocks won't be deleted.", contName)
+                continue
+
+            # Get RSE in which each block is available
+            availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
```
amaltaro, line 407: It might be useful to wrap this call with a try/except, such that we don't lose the whole cycle in case of issues while talking to Rucio.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,67 @@ def deleteBlocks(self):
...
+                # If block is available at every RSE its container needs to be transferred, the block can be deleted
+                if cont['rse'].issubset(blockRSEs['replica']):
+                    blocksToDelete.append(blockRSEs['name'])
+
+        # Delete agent created rules locking the block
+        binds = []
+        logging.info("Going to delete %d blocks", len(blocksToDelete))
+        for block in blocksToDelete:
+            rules = self.rucio.listDataRules(block, scope=self.scope, account=self.rucioAcct)
+            if not rules:
```
amaltaro, line 419: This might be a problem in the future, because the `listDataRules` API catches Rucio exceptions and return an empty list. So it could be that: a) there are no real rules for that block anymore; or b) the query fails and we proceed marking the block as deleted.

The only option I see would be to stop catching that exception in the Rucio wrapper and let it propagate to whoever is making the call. But that would be a single wrapper API with a different behaviour... so I'm afraid I don't have any good suggestion for now.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,67 @@ def deleteBlocks(self):
...
+                logging.info("Block rule for: %s has been deleted by previous cycles", block)
+                binds.append({'DELETED': 1, 'BLOCKNAME': block})
+                continue
+            for rule in rules:
+                if self.rucio.deleteRule(rule['id']):
```
amaltaro, line 424: We should make sure we only mark blocks as deleted if ALL block-level rules are properly deleted. Maybe a counter for deleted rules vs rules to delete.

amaltaro (changes requested):
@germanfgv these changes are looking good. There are still a few items that need attention though.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: 1. All blocks within a container are being transferred to the same sites, but some blocks may finish transferring earlier than others.
2. Yes, it is the list of sites the container/dataset is being transferred to.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Fixed this, and added another clause so every block with more than 2 replicas gets deleted.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I added try/except. I'm catching a general exception as the other methods in the component or should I import RucioException from RucioAPI?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I added the counter

germanfgv:
@amaltaro Can you take another look at the state of the code?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,84 @@ def deleteBlocks(self):
...
-          If yes => set deleted=2
-          If no => next check
         Has transfer to all destinations finished ?
-          If yes => request block deletion, approve request, set deleted=1
+          If yes => Delete rules associated with the block, set deleted=1
           If no => do nothing (check again next cycle)
         """
-        # FIXME: figure out the proper logic for rule block deletion
-        logging.info("Starting deleteBlocks methods --> IMPLEMENT-ME!!!")
+        logging.info("Preparing to create block rules into Rucio...")
```
amaltaro, line 377: How about "Checking if there are block rules to be deleted..."?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,84 @@ def deleteBlocks(self):
...
+            except Exception as exc:
+                msg = "Failed to get replica info for blocks in container: %s.\n" % contName
+                msg += "Will retry again in the next cycle. Error: %s" % str(exc)
+                logging.error(msg)
+                continue
+
+            for blockRSEs in availableRSEs:
+                # If block is available at every RSE its container needs to be transferred, the block can be deleted
+                blockSites = set(blockRSEs['replica'])
+                if cont['rse'].issubset(blockSites) or len(blockSites & cont['rse']) >= 2:
```
amaltaro, line 421: This logic isn't very clear to me. Trying to expand it in plain english:
* if block is available in a subset of the container locations, then delete it (even if block is available in a single location - do we want this behaviour?)
* or, if block location and container location have at least 2 common RSEs, delete it (does it cover the parked data, which requires a single archival copy?)
Is that correct?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,84 @@ def deleteBlocks(self):
...
+                    blocksToDelete.append(blockRSEs['name'])
+
+        # Delete agent created rules locking the block
+        binds = []
+        logging.info("Going to delete %d blocks", len(blocksToDelete))
```
amaltaro, line 426: Not to cause a misinterpretation from the logs, how about "Going to delete %d block rules"?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: I think the best will be to change `listDataRules` to raise an `WMRucioException` exception in case that call fails. You catch it here then and handle it accordingly.

amaltaro (changes requested):
@germanfgv German, the code is almost final IMO. There is just one dependency though and I'll get it fixed/merged today. There are a few other comments and/or items to be fixed along the code. Thanks

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: The conditions are the following:
-The first condition is: The container locations is a subset of the RSEs in which the block is available. If this condition is met, it means the block has been transferred to every site to which a container level rule has been created, therefore, it can be removed. This is the way PhEDExInjector checks if the block can be deleted. 
- The second condition checks if the block is available in at least 2 of the sites to which its container was subscribed. If this condition is met, the block is in enough sites so it can be deleted.

Taking into account that I added the second condition, the first condition will only make a difference in the case of blocks belonging to containers that are only subscribed to 1 site. This can cover a few of the parked data scenarios, but it's probably best to remove the first condition and create a check to deal specifically with parked data. I'm not sure how to do that in a not-hardcoded-way

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -370,15 +370,84 @@ def deleteBlocks(self):
...
+                else:
+                    logging.warning("Failed to delete rule: %s, for block %s.", rule['id'], block)
+
+            if deletedRules == len(rules):
+                binds.append({'DELETED': 1, 'BLOCKNAME': block})
+                logging.info("Successfully deleted every rule for block %s.", block)
+            else:
+                logging.warning("Some rules for block %s were not deleted. Will retry in the next cycle.", block)
+
+        self.markBlocksDeleted.execute(binds)
```
germanfgv, line 448: Need whether `binds` is empty or not

amaltaro: maybe add it to the class carrying the database operation.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: After discussion with @drkovalskyi , It is clear that we only need the first condition. This method will assume that the container rules meet the Tier0 requirements, and each block will be deleted if and only if it has been transferred to every requested destination.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Should I assume that there is only going to be 1 block level rule for each block? because, if that's the case, then I can remove the  `for` loop that goes through the rules

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Yes, it looks good. And L405 protects you against containers without any data placement!

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: In general, yes. But we better deal with it as a list to be on the safe side.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I added a protection MarkBlocksDeleted.py

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -379,15 +379,85 @@ def deleteBlocks(self):
...
+                availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
+            except Exception as exc:
+                msg = "Failed to get replica info for blocks in container: %s.\n" % contName
+                msg += "Will retry again in the next cycle. Error: %s" % str(exc)
+                logging.error(msg)
+                continue
+
+            for blockRSEs in availableRSEs:
+                # If block is available at every RSE its container needs to be transferred, the block can be deleted
+                blockSites = set(blockRSEs['replica'])
```
germanfgv, line 429: @amaltaro this assumes `rucio.phedexCompat = False`. Should I make it compatible with phedex?

amaltaro: better not make it compatible. PhEDEx related things will eventually get deleted from the code, so that would give you more work (and it would probably be less efficient)

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: German, can you please fix this logging message?

germanfgv: I fixed the message

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -379,15 +379,85 @@ def deleteBlocks(self):
...
+                    blocksToDelete.append(blockRSEs['name'])
+
+        # Delete agent created rules locking the block
+        binds = []
+        logging.info("Going to delete %d block rules", len(blocksToDelete))
+        for block in blocksToDelete:
+            try:
+                rules = self.rucio.listDataRules(block, scope=self.scope, account=self.rucioAcct)
+            except WMRucioException as exc:
+                pass
```
amaltaro, line 440: German, can you please add an logging.warning error message here? not to pass this failed call quietly.

amaltaro (changes requested):
German, I think we only need one more logging - see comment along the code - and we are good to go.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I added a warning message:
"Unable to retrieve replication rules for block: %s. Will retry in the next cycle."

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: Right now, I think the value of `rucio.phedexCompat` cannot be changed from any of the config files (Not from config.py nor rucio.cfg). Should I add this functionality?

amaltaro: Oh, now I understand what you say. Yes, we should create the Rucio instance with `phedexCompatible=False`, to make sure that the code above will behave properly.
Can you please update the Rucio object construction as above? No need to make anything configurable. That flag is mostly for external services that needed transparent changes.

amaltaro (approved):
Looks good to me now. Is there anything else that you're trying to push in?

germanfgv:
After a couple of replays, this seems to be working as expected. I think this is ready to be merged.

amaltaro:
Thank you very much again, German.

PR #9961, opened 2020-10-02
Added prom_metrics REST API for prometheus-like metrics

### LINKED ISSUE #9939: Provide proper metrics for web based services

amaltaro:
Fixes #9939
Superseeds https://github.com/dmwm/WMCore/pull/9940

#### Status
In development

#### Description
This PR provides two commits:
 * 1st: with an utilitarian module to convert CherryPy stats to an structure like those provided by Prometheus (flat structure)
 * 2nd: a new REST API `prom_metrics` created under WMCore/REST, which should provide the same output as stats, but in a different format.

Even though the complexity of the `flattenStats` function is O(n^4)!!, the request takes only 1ms more to get served. So I think it's still okay to keep this very complex code around.

NOTE-1: the WebTools package hasn't been touched because I'm not sure where the change has to go and how I could test it.
NOTE-2: need to implement the same logic for MicroServices (probably in a different PR).
NOTE-3: if the code is Okay, a proper unit test needs to be implemented to catch releases incompatibilities.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/Utils/CPMetrics.py
+++ b/src/python/Utils/CPMetrics.py
@@ -1,0 +1,267 @@
+#!/usr/bin/env python
+"""
+File       : CPMetrics.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
+             Alan Malta Rodrigues <alanmalta AT gmail dot com>
+Description: This module provides functions to flatten CherryPy stats
+and provide them in Prometheus format for scraping.
+
+Below we present three different outputs:
+    - CherryPy stats default output
+    - flatten structure of stats
+    - prometheus output for stats
+
+### Defautl CherryPy metrics output
+
+{
+    "Cheroot HTTPServer 4388603856": {
+    "Accepts": 0,
+    "Accepts/sec": 0.0,
+    "Bind Address": "('127.0.0.1', 8080)",
+    "Bytes Read": -1,
+    "Bytes Written": -1,
+    "Enabled": false,
+    "Queue": 0,
+    "Read Throughput": -1,
+    "Requests": -1,
+    "Run time": -1,
+    "Socket Errors": 0,
+    "Threads": 20,
+    "Threads Idle": 19,
+    "Work Time": -1,
+    "Worker Threads": {
+        "CP Server Thread-10": {
+        "Bytes Read": 0,
+        "Bytes Written": 0,
+        "Read Throughput": 0.0,
+        "Requests": 0,
+        "Work Time": 0,
+        "Write Throughput": 0.0
+        },
+        .....
+    },
+    "Write Throughput": -1
+    },
+    "CherryPy Applications": {
+        "Bytes Read/Request": 0.0,
+        "Bytes Read/Second": 0.0,
+        "Bytes Written/Request": 0.0,
+        "Bytes Written/Second": 0.0,
+        "Current Requests": 0,
+        "Current Time": 1601039589.916887,
+        "Enabled": true,
+        "Requests": {},
+        "Requests/Second": 0.0,
+        "Server Version": "17.4.2",
+        "Start Time": 1601039576.541158,
+        "Total Bytes Read": 0,
+        "Total Bytes Written": 0,
+        "Total Requests": 0,
+        "Total Time": 0,
+        "Uptime": 13.375718116760254
+    }
+}
+
+### flatten structure via flattenStats function
+{
+"cherrypy_app_requests_second": 0.0,
+"cherrypy_app_uptime": 1.7454230785369873,
+"cherrypy_app_total_time": 0,
+"cherrypy_app_current_time": 1601040713.369842,
+"cherrypy_app_bytes_read_second": 0.0,
+"cherrypy_app_requests": {},
+"cherrypy_http_server_queue": 0,
+"cherrypy_app_bytes_written_second": 0.0,
+"cherrypy_app_total_requests": 0,
+"cherrypy_http_server_write_throughput": -1,
+"cherrypy_app_enabled": true,
+"cherrypy_app_start_time": 1601040711.624408,
+"cherrypy_http_server_threads": 20,
+"cherrypy_http_server_work_time": -1,
+"cherrypy_app_bytes_read_request": 0.0,
+"cherrypy_http_server_bytes_read": -1,
+"cherrypy_http_server_accepts_sec": 0.0,
+"cherrypy_server_worker_threads": [
+   {"thread_name": "cp_server_thread_3", "read_throughput": 0.0,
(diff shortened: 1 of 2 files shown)


amaltaro:
@vkuznet  Valentin, I think this PR addresses everything that you needed in your PR #9940. Could you please confirm?

I noticed one weird output on the prometheus-like metrics, e.g.:
```
# HELP reqmgr2_cherrypy_server_worker_threads
# TYPE reqmgr2_cherrypy_server_worker_threads histogram
reqmgr2_cherrypy_server_worker_threads{thread_name="cp_server_thread_30",read_throughput="0.0",work_time="0",write_throughput="0.0",bytes_written="0",bytes_read="0",requests="0"} 1
```
that number `1` at the very end. I fail to see where that comes from. Could you please check the code and point where the problem is?

It would be great if you could check the exact output of those APIs under:
https://alancc7-cloud1.cern.ch/reqmgr2/data/stats
and
https://alancc7-cloud1.cern.ch/reqmgr2/data/prom_metrics

I believe you already have access to my VM. Thanks

vkuznet:
@amaltaro , I don't understand why did you made a new PR? Anyway, few things:
- please do not change name of exposed endpoint. I used `/metrics` endpoint name because it is default for all exporters used by prometheus, while your invention to `prom_metrics` break the rule and will force to use special configuration parameter which I rather prefer to avoid
- you skipped unit test I provided, see https://github.com/dmwm/WMCore/blob/29e5086bd4251e44b85842e265a855a7688afc45/test/python/Utils_t/CPMetrics_t.py
- I don't understand why did you changed input parameters of metrics API either, i.e. you pass cherrypy dict to metrics API. Doing so you intentionally ask Python to allocate additional memory for it since it passed it by value (extra copy).
Please fix all of the above.

Meanwhile, your question about histogram syntax. It requires two parameters the name (with attributes) and a value. Value 1 correspond that it exists. There is nothing wrong here.

Regarding how to test these things. I suggest that you follow this recipt:
- download static executable of prometheus from https://prometheus.io/download/
- create the following config file `prometheus.yml` somewhere. You may adjust `localhost:8080` to whatever your cherrypy server is using (do not change `localhost:9090` which is default prometheus port):
```
# my global config
global:
  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
  # scrape_timeout is set to the global default (10s).

# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

# A scrape configuration containing exactly one endpoint to scrape:
# Here it's Prometheus itself.
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: 'prometheus'

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
    - targets: ['localhost:9090']
  - job_name: 'cherrypy'
    static_configs:
    - targets: ['localhost:8080']
```
- start prometheus as following
```
./prometheus --config.file=prometheus.yml
```
- open whatever host you use on port 9090 and check the prometheus web UI which will contain your metrics. If you're busy, lazy, or unable to do that I can login to your VM and set it up. So far I don't have access to your VM `alancc7-cloud1` and if you need my help over there please provide me the access.

amaltaro:
So, I made a new PR because there are a few things which I wanted to change in yours, and I figured it would be more efficient if I had done those myself, such as:
* Test code is not supposed to go into the source code, as present in your PR;
* I understood you were proposing to have a CheeryPy stats parser, not invoker + parser (see next comment);
* having the cherrypy stats call in the Utilitarian function:
https://github.com/dmwm/WMCore/pull/9940/files#diff-7b548d90012d34adcc512c981750b322R201
looks pretty bad to me. If there will be a REST endpoint serving such requests - through CherryPy engine - we should simply use that data from there and use the util function simply to 
* data/memory allocation is in the scale of a few Kb, I don't really think it's a concern. I hope we won't have multiple clients hammering this API, so more reason not to be concerned;

About the endpoint, I did not know prometheus monitoring was hard-wired to `metrics`, I can of course change it back to metrics.

This PR is not final (proper unit tests need to be created to validate the cherrypy stats structure, and the parser). I just wanted to get back to you to check whether this is what you were trying to get from WMCore.

You are right! I can't allocate more time on setting up this test environment and getting it tested through prometheus. It might be easier if you take care of that yourself on your PR (please take the changes I made here). If you do so, let me know and I can close this PR. Otherwise I can try to get back to it in the next week or two.

vkuznet:
@amaltaro , I know you're busy and it's fine if I'll do it. As I said I need access to your VM, then I can setup things properly for prometheus over there. Feel free to close this PR and I'll incorporate whatever you'll tell me:
- ok, I'll pas around data from cherrypy
- I'm not sure I understood your concern about test, I put it into `test/python/Utils_t/CPMetrics_t.py` in my PR, please clarify this
If anything else needs to be done, please put it in my PR as comments/suggestions and I can proced.

vkuznet:
@amaltaro , Alan I put your changes back to #9940 , please review over there.

amaltaro:
@vkuznet about access to my VM, please check slack.

vkuznet (approved):
Found this one which was still pending under my review. All changes are fine.

amaltaro:
Changes have been incorporated into the initial PR, so I am closing this one out.

PR #9978, opened 2020-10-09
Add getTransferInfo to MSOutput and info request to MSManager.

### LINKED ISSUE #9613: MS: complete MSOutput REST APIs implementation
**Impact of the new feature**
ReqMgr2MS

No solution yet, it's just a placeholder for a final implementation of the MSOutput REST APIs.

One of them has to provide the status of the service (to be used for basic monitoring);
the second has to be a functional API, for end users, possibly fetching data from MongoDB.

none

**Additional context**
none

todor-ivanov:
Fixes #9613 

#### Status
ready

#### Description
Provides the function `getTransferInfo`, which searches and reads a document from MongoDB in all collections related to MSOUtput Service. It is called only from MSManager in order to serve the `info?request=...` API for the service.  In this way the full information recorded into MongoDB related to a request is provided and so that all request transfer records can be tracked.

YES

#### Related PRs
N/A

N/A

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSManager.py
+++ b/src/python/WMCore/MicroService/Unified/MSManager.py
@@ -231,6 +231,8 @@ class MSManager(object):
                 transferDoc = self.msMonitor.reqmgrAux.getTransferInfo(reqName)
             elif 'transferor' in self.services:
                 transferDoc = self.msTransferor.reqmgrAux.getTransferInfo(reqName)
+            elif 'output' in self.services:
+                transferDoc = self.msOutputProducer.getTransferInfo(reqName)
             if transferDoc:
                 # it's always a single document in Couch
                 data['transferDoc'] = transferDoc[0]
--- a/src/python/WMCore/MicroService/Unified/MSOutput.py
+++ b/src/python/WMCore/MicroService/Unified/MSOutput.py
@@ -1018,7 +1018,31 @@ class MSOutput(MSCore):
                 self.logger.exception(msg)
                 raise ex
         else:
-            self.logger.info("Query: '%s' did not return any records from MongoDB", mQueryDict)
+            self.logger.info("%s Query: '%s' did not return any records from MongoDB", dbColl.name, mQueryDict)
+
+    def getTransferInfo(self, reqName):
+        """
+        Searches and reads a document from MongoDB in all collections related to
+        the MSOutput service. It is supposed to be called only by MSManager e.g.:
+           transferDoc = self.msOutputProducer.getTransferInfo(reqName)
+        And the output to be served to the REST interface, so that all request
+        transfer records can be tracked.
+        :param reqName: The name of the request to be searched for
+        :return: a list of all msOutDocs with the record if it exists or None otherwise
+        """
+
+        mQueryDict = {'RequestName': reqName}
+        result = []
+        stripKeys = ['_id']
+
+        for dbColl in [self.msOutNonRelValColl, self.msOutRelValColl]:
+            doc = dbColl.find_one(mQueryDict)
+            if doc:
+                for key in stripKeys:
+                    doc.pop(key, None)
+                result.append(doc)
+                break
+        return result
 
     def docCleaner(self, doc):
         """


src/python/WMCore/MicroService/Unified/MSManager.py
```diff
@@ -231,6 +231,8 @@ def info(self, reqName=None):
                 transferDoc = self.msMonitor.reqmgrAux.getTransferInfo(reqName)
             elif 'transferor' in self.services:
                 transferDoc = self.msTransferor.reqmgrAux.getTransferInfo(reqName)
```
amaltaro, line 233: I would like to sanitize the request name here, by calling:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L500
but we can deal with that in a different issue if you prefer.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -1018,7 +1018,27 @@ def getDocsFromMongo(self, mQueryDict, dbColl, limit=1000):
...
+           transferDoc = self.msOutputProducer.getTransferInfo(reqName)
+        And the output to be served to the REST interface, so that all request
+        transfer records can be tracked.
+        :param reqName: The name of the request to be searched for
+        :return: a list of all msOutDocs with the record if it exists or None otherwise
+        """
+        mQueryDict = {'RequestName': reqName}
+        result = []
+        stripKeys = ['_id', 'producerDoc']
+        for dbColl in [self.msOutRelValColl, self.msOutNonRelValColl]:
```
amaltaro, line 1036: On this method, I have a few ideas that would improve the performance of this API:
* return the document as soon as a document is retrieved (and in that case, we should start with the non-relval collection)
* instead of calling getDocsFromMongo method (which does a conversion to MSOutputTemplate), how about we go straight with `self.msOutNonRelValColl.find(mQueryDict)`?

If performance difference is negligible, then there is no need to change anything.

amaltaro (commented):
Thanks Todor. Changes look good to me, however, I made a couple of comments that might be interesting to consider.

src/python/WMCore/MicroService/Unified/MSManager.py
todor-ivanov: Done!

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSManager.py
todor-ivanov: Actually, @amaltaro, this raises a question though. Should we return a `None` transferDoc in case of a bad requestName expression (maybe this could be beneficial if the API is used by another automated interface but not only by a User) or should we return the error so that to inform the User about his mistake?

amaltaro: Todor, looking a bit more into this, I think what would make more sense is to validate the GET parameters at the parent class (Service/Data) and return an error like:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py#L35
in case of bad request (request name not matching the lexicon).

That work goes orthogonal to these changes though, so I think you can remove the current request name validation here and we deal with that in a different issue/PR. What do you think?

todor-ivanov: OK , Alan@amaltaro. I dropped the reqName validation from `getTransferInfo`

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -28,6 +28,7 @@
 from WMCore.Database.MongoDB import MongoDB
 from WMCore.MicroService.DataStructs.MSOutputTemplate import MSOutputTemplate
 from WMCore.WMException import WMException
+from WMCore.Lexicon import requestName as checkRequestName
```
amaltaro, line 31: This should be removed as well.

src/python/WMCore/MicroService/Unified/MSOutput.py
```diff
@@ -1018,7 +1019,30 @@ def getDocsFromMongo(self, mQueryDict, dbColl, limit=1000):
...
+        result = []
+        stripKeys = ['_id']
+
+        for dbColl in [self.msOutNonRelValColl, self.msOutRelValColl]:
+            doc = dbColl.find_one(mQueryDict)
+            if doc:
+                for key in stripKeys:
+                    doc.pop(key, None)
+                result.append(doc)
+                return result
```
amaltaro, line 1045: This should return result in case no documents are found, otherwise it's by default, returning None.
I think we can have something like:
```
                result.append(doc)
                break
        return result
```

amaltaro (changes requested):
Todor, please find a couple more comments along the code.

src/python/WMCore/MicroService/Unified/MSOutput.py
todor-ivanov: done

amaltaro (approved):
Thanks Todor, this looks good now.
Can you please make a new GH issue for the input parameter sanitization in Microservices? Thanks

PR #9980, opened 2020-10-11
Disable PhEDExInjector; Enable all tiers to be injected in Rucio

amaltaro:
No issue created

#### Status
ready

#### Description
This changes the default Data Management system for output data, allowing all the datatiers to be injected into Rucio; at same time completely disabling PhEDExInjector (component still running, but not taking any actions).

Note that this configuration remains:
```
config.RucioInjector.skipRulesForTiers = ["NANOAOD", "NANOAODSIM"]
```
meaning, the agent is not allowed to create container-level rules - if specified in the workflow - for those 2 datatiers.

yes

#### Related PRs
none

none

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -162,7 +162,7 @@ config.PhEDExInjector.namespace = "WMComponent.PhEDExInjector.PhEDExInjector"
 config.PhEDExInjector.componentDir = config.General.workDir + "/PhEDExInjector"
 config.PhEDExInjector.logLevel = globalLogLevel
 config.PhEDExInjector.maxThreads = 1
-config.PhEDExInjector.enabled = True
+config.PhEDExInjector.enabled = False
 config.PhEDExInjector.subscribeDatasets = True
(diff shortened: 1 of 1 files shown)


etc/WMAgentConfig.py
```diff
@@ -371,7 +371,7 @@
 config.RucioInjector.createBlockRules = True
 config.RucioInjector.RSEPostfix = False  # enable it to append _Test to the RSE names
 config.RucioInjector.metaDIDProject = "Production"
-config.RucioInjector.listTiersToInject = ["NANOAOD", "NANOAODSIM"]  # []
+config.RucioInjector.listTiersToInject = []  # ["NANOAOD", "NANOAODSIM"]
 config.RucioInjector.skipRulesForTiers = ["NANOAOD", "NANOAODSIM"]
```
amaltaro, line 375: Note we are still going to skip container-level rules for these 2 datatiers.

amaltaro:
Noo! I forgot to backport this patch...

PR #9981, opened 2020-10-11
Increase polling cycle for WorkQueueManagerLocationPoller to 1h

### LINKED ISSUE #9979: Increase polling cycle for WorkQueueManagerLocationPoller thread
**Impact of the new feature**
WMAgent

All the WorkQueueManager threads follow the same polling cycle, which isn't ideal for some of those threads, like `WorkQueueManagerLocationPoller`

Instead of using the default 180 seconds polling cycle, we should make `WorkQueueManagerLocationPoller` run less often. An interval of 30 or 60min should be more than enough. Especially now that we are still trying to find the best optimisations for data location through Rucio.

I probably mentioned this alternative somewhere else, but we could actually get rid of this thread if we made sure that elements acquired by the agent would have their WMBS preparation done right away (thus bypassing the local queue logic).

**Additional context**
none

amaltaro:
Fixes #9979 

#### Status
ready

#### Description
Make a new component configuration attribute `dataLocationInterval` to set a different polling cycle for the `WorkQueueManagerLocationPoller` thread, changing it from 3min to 1h.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/WorkQueueManager/WorkQueueManager.py
+++ b/src/python/WMComponent/WorkQueueManager/WorkQueueManager.py
@@ -5,17 +5,17 @@ WorkQueuemanager component
 Runs periodic tasks for WorkQueue
 """
 from __future__ import print_function
+
 import threading
 
-from WMCore.Agent.Harness import Harness
-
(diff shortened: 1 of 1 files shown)


amaltaro:
tested, merging.

PR #9988, opened 2020-10-13
Treat NANOAOD/NANOAODSIM as any other datatier in the output data placement logic

amaltaro:
Fixes #9986 

#### Status
ready

#### Description
As the title says, stop excluding NANOAOD tiers from the output data placement (both on WMAgent and MSOutput).

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
Deployment changes: https://github.com/dmwm/deployment/pull/965

PR #9989, opened 2020-10-13
Skip Custodial and make generic Disk NonCustodial data placement in RucioInjector

### LINKED ISSUE #9639: Repurpose workflow output data placement settings to deal with Rucio rules
**Impact of the new feature**
ReqMgr2, WMAgent

Once we migrate away of PhEDEx, there will be a few workflow level parameters (set during workflow assignment) that will no longer make sense, given that those were parameters meant for a PhEDEx subscription request, example:
```
    "CustodialSites": ["T1_US_FNAL_MSS"],
    "NonCustodialSites": ["T1_US_FNAL_Disk"],
    "AutoApproveSubscriptionSites": ["T1_US_FNAL_Disk"],
    "SubscriptionPriority": "Normal",
    "CustodialSubType": "Replica",
    "NonCustodialSubType": "Move",
    "CustodialGroup": "FacOps",
    "NonCustodialGroup": "AnalysisOps",
```

We should update the workflow spec files, by either:
* removing those parameters that will no longer be used
* rename any other parameters that will still be necessary

and adapt WMAgent (including the dbsbuffer_dataset_subscription table) to be able to read the new parameters, if any. Note that this needs to be made in a backward compatible way, given that workflows can remain in the system for a year or more.

UPDATE: this module `PhEDEx.DataStructs.SubscriptionList` is used to validate those workflow parameters. It should either be refactored/renamed considering the Rucio system; or it should be completely removed (and its unit tests).

none

**Additional context**
none

amaltaro:
Fixes #9639 

#### Status
ready

#### Description
Summary of changes is:
* created two new RucioInjector component configuration parameters:
  * `containerDiskRuleParams`: additional parameters overriding the default container-level rule attributes, only for Disk data placement
  * `containerDiskRuleRSEExpr`: very generic RSE expression to be used for container-level Disk rules
* if it's a T0 agent, then proceed with standard output data placement; otherwise, proceed with central production data placement, which has the following peculiarities:
  * Custodial data placement is bypassed and the container is marked as subscribed in the relational database;
  * NonCustodial data placement overrides settings at the workflow spec, where destination is set to any T1+T2 real Disk endpoint;  number of copies is set to 2, with a grouping=DATASET, and a rule weight is added as well (with "ddm_quota" value).

yes

#### Related PRs
To be merged after: https://github.com/dmwm/WMCore/pull/9988

None

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -84,6 +84,8 @@ class RucioInjectorPoller(BaseWorkerThread):
         self.pollRules = config.RucioInjector.pollIntervalRules
         self.lastRulesExecTime = 0
         self.createBlockRules = config.RucioInjector.createBlockRules
+        self.containerDiskRuleParams = config.RucioInjector.containerDiskRuleParams
+        self.containerDiskRuleRSEExpr = config.RucioInjector.containerDiskRuleRSEExpr
         self.skipRulesForTiers = config.RucioInjector.skipRulesForTiers
         self.listTiersToInject = config.RucioInjector.listTiersToInject
         if config.RucioInjector.metaDIDProject not in RUCIO_VALID_PROJECT:
@@ -109,6 +111,13 @@ class RucioInjectorPoller(BaseWorkerThread):
 
         self.testRSEs = config.RucioInjector.RSEPostfix
         self.filesToRecover = []
+
+        # output data placement has a different behaviour between T0 and Production agents
+        if hasattr(config, "Tier0Feeder"):
+            logging.info("RucioInjector running on a T0 WMAgent")
+            self.isT0agent = True
+        else:
+            self.isT0agent = False
 
         if not self.listTiersToInject:
             logging.info("Component configured to inject all the data tiers")
@@ -283,7 +292,7 @@ class RucioInjectorPoller(BaseWorkerThread):
             rseName = "%s_Test" % item['pnn'] if self.testRSEs else item['pnn']
             # DATASET = replicates all files in the same block to the same RSE
             resp = self.rucio.createReplicationRule(item['blockname'],
-                                                    rseExpression="rse=%s" % rseName, **kwargs)
+                                                    rseExpression=rseName, **kwargs)
             if resp:
                 msg = "Block rule created for block: %s, at: %s, with rule id: %s"
                 logging.info(msg, item['blockname'], item['pnn'], resp[0])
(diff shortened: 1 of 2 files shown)


amaltaro:
Tests on the testbed agent went fine.

amaltaro:
Backported and merged in 1.4.1.patch2

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -415,6 +427,67 @@ def insertContainerRules(self):
...
+            kwargs.update(self.containerDiskRuleParams)
+
+            rseExpr = self.containerDiskRuleRSEExpr
+            if self.testRSEs:
+                rseExpr = rseExpr.replace("cms_type=real", "cms_type=test")
+            logging.info("Creating container rule for %s against RSE %s", container, rseExpr)
+            try:
+                resp = self.rucio.createReplicationRule(container,
+                                                        rseExpression=rseExpr, **kwargs)
+            except Exception as exc:
```
todor-ivanov, line 460: Looking at the code of `createReplicationRule` function, it seems that the only Exception that can be raised there and captured here is `AccessDenied` from [1]. All the rest are masked in [2], including the `DuplicateRule` few lines above. So the question is should we retry indefinitely/infinitely here expecting something to happen all by itself on the Rucio side to fix the error and give the access in question or it is worth alarming the WMCore team after few retries here that, something is wrong.   

[1]
https://github.com/dmwm/WMCore/blob/51b89389493932391080d9d4e1ad95e9356fd3bb/src/python/WMCore/Services/Rucio/Rucio.py#L505

[2]
https://github.com/dmwm/WMCore/blob/51b89389493932391080d9d4e1ad95e9356fd3bb/src/python/WMCore/Services/Rucio/Rucio.py#L533

todor-ivanov (approved):
@amaltaro, I made a single comment in the code. But given that the change already went into the wmagent branch the question may easily be left behind. All the rest looks good to me.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: This code is now only dealing with Disk data placement, through a very generic rse expression. So if we hit an `AccessDenied`, it should happen only once in life, until the CMS Rucio team fixes it. Given that the impact would be very clear too - not in the form of a WMAgent alert thoguh - I decided not to deal with that specific exception in this code (it remains in the T0 logic though).

That's a good question though, and **ideally**, we should eventually retry actions a few times (over a few different cycles) until a hard failure is triggered (possibly triggering a notification/alert too). There are many places that this behaviour would be very helpful.

PR #9994, opened 2020-10-15
Properly validate RequestPriority parameter

### LINKED ISSUE #9992: RequestPriority validation only works during the workflow creation
**Impact of the bug**
ReqMgr2

**Describe the bug**
`RequestPriority` spec parameter is supposed to be integer, according to its specification here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1033-L1034

however, I just spotted a few requests in the system with real values, example:
https://cmsweb.cern.ch/reqmgr2/fetch?rid=cmsunified_ACDC2_task_TOP-RunIISummer19UL17wmLHEGEN-00071__v1_T_201010_132628_5975
```
    244840.32: 2020-10-10 11:26:28 UTC DN: /DC=ch/DC=cern/OU=computers/CN=unified/vocms0268.cern.ch
```

**How to reproduce it**
I guess one can try even from the Gui for a workflow in active status.

**Expected behavior**
Such values should not be allowed and the HTTP request - to update the workflow priority - should fail. In other words, the spec parameter validation function needs to be properly applied, including the data type.

I'm still not sure whether this can cause other problems, perhaps at job submission or something like that...

amaltaro:
Fixes #9992 

#### Status
ready

#### Description
Make sure the `RequestPriority` spec parameter is properly validated, regardless of which state the workflow is.
For clarity: it must be of the `integer` type and be in the range of `[0, 1e6]`

Yes (well, we need to chase those requests with float value, otherwise an ACDC might fail to be created...)

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/ReqMgr/Utils/Validation.py
+++ b/src/python/WMCore/ReqMgr/Utils/Validation.py
@@ -13,6 +13,7 @@ from WMCore.ReqMgr.DataStructs.RequestStatus import check_allowed_transition, ST
 from WMCore.ReqMgr.Tools.cms import releases, architectures, dashboardActivities
 from WMCore.Services.DBS.DBS3Reader import getDataTiers
 from WMCore.WMFactory import WMFactory
+from WMCore.WMSpec.StdSpecs.StdBase import StdBase
 from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
 from WMCore.WMSpec.WMWorkloadTools import loadSpecClassByType, setArgumentsWithDefault
 from WMCore.Cache.GenericDataCache import GenericDataCache, MemoryCacheStruct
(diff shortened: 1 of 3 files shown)


amaltaro:
Tests went fine!

src/python/WMCore/ReqMgr/Utils/Validation.py
```diff
@@ -63,15 +64,31 @@ def validate_request_update_args(request_args, config, reqmgr_db_service, param)
...
+    Validate the RequestPriority argument against its definition
+    in StdBase
+    :param reqArgs: dictionary of user request arguments
+    :return: nothing, but raises an exception in case of an invalid value
+    """
+    if 'RequestPriority' in reqArgs:
+        reqPrioDefin = StdBase.getWorkloadCreateArgs()['RequestPriority']
+        try:
+            # Web GUI sends it as a string
+            reqArgs['RequestPriority'] = int(reqArgs['RequestPriority'])
```
todor-ivanov, line 83: Wouldn't that be more clear if one just does the following here:

```
if not isinstance(reqArgs['RequestPriority'], reqPrioDefin['type']):
    raise InvalidSpecParameterValue(msg)
```
We need to check for the type only and not for the possibility to cast the object to a given type, because we never do so down the code. And we will also avoid hardcoding a `int` type check in the middle of the code while we have the object type defined from StdBase. I think we should take it from the upper level object otherwise we risk to have objects that silently transform from one type to another, which in my opinion is a bug.

todor-ivanov (changes requested):
@amaltaro  I made one comment in the code. Please take a look at it. Thanks!

src/python/WMCore/ReqMgr/Utils/Validation.py
amaltaro: Good catch. I meant to write:
```
reqArgs['RequestPriority'] = reqPrioDefin['type'](reqArgs['RequestPriority'])
```

About simply checking the data type, it would be easier if we were already running this software in python3. Since we are still in py2, any check for string is just another complication in the near future.

todor-ivanov (approved):
looks good to me
Thanks @amaltaro

amaltaro:
@todor-ivanov sorry for bugging you again here. I didn't manage to reproduce the string type on my VM, neither via script, nor via Gui. So I made the change to simply check for the data type. Can you please have another look? Thanks

todor-ivanov (approved):
Looks perfect to me. Thanks!

amaltaro:
Thanks!

PR #9999, opened 2020-10-19
If listDataRules API fail, raise a WMRucioException

### LINKED ISSUE #9998: Re-raise exception in Rucio listDataRules API
**Impact of the new feature**
WMCore in general (and any other project using that API)

As part of the block-level rule deletion logic, we need to have a clear way to distinguish between
1. server failed to serve the user request (some sort of exception)
2. server properly served the user request, with an empty list of rules

So, re-raise an exception to the caller in case listDataRules fail to be served.

If an exception is hit in the `listDataRules` Rucio wrapper API, re-raise a `WMRucioException` to the caller, with the proper error message, of course!

Leave it as it is, but then it becomes harder to deal with those 2 possible cases...

**Additional context**
Needed for the development of: https://github.com/dmwm/WMCore/pull/9959

amaltaro:
Fixes #9998

#### Status
not-tested

#### Description
It changes the behaviour of the `listDataRules` Rucio wrapper API, which used to always return a list type object. Now it can also raise a `WMRucioException` exception if the http request fails.

Yes, in general

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -607,12 +607,13 @@ class Rucio(object):
         """
         kwargs["name"] = name
         kwargs.setdefault("scope", "cms")
-        res = []
         try:
             res = self.cli.list_replication_rules(kwargs)
-        except Exception as ex:
-            self.logger.error("Exception listing rules for data: %s and kwargs: %s. Error: %s",
-                              name, kwargs, str(ex))
+        except Exception as exc:
+            msg = "Exception listing rules for data: {} and kwargs: {}. Error: {}".format(name,
+                                                                                          kwargs,
+                                                                                          str(exc))
+            raise WMRucioException(msg)
         return list(res)
 
     def listDataRulesHistory(self, name, scope='cms'):


src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -607,12 +607,13 @@ def listDataRules(self, name, **kwargs):
...
         try:
             res = self.cli.list_replication_rules(kwargs)
-        except Exception as ex:
-            self.logger.error("Exception listing rules for data: %s and kwargs: %s. Error: %s",
-                              name, kwargs, str(ex))
+        except Exception as exc:
+            msg = "Exception listing rules for data: {} and kwargs: {}. Error: {}".format(name,
+                                                                                          kwargs,
+                                                                                          str(exc))
+            raise WMRucioException(msg)
```
todor-ivanov, line 616: I am leaving the comment here, which is related to an old piece of code referenced here, but I cannot make the comment on the exact place the definition happens. Anyway, since we are about to raise those `WMRucioException' exceptions here and there, I would suggest to have a little bit more "by the standard' definition on line 23 [1]. The bare minimum I would say should be to have at least an `__init__` method, instead of just `pass`. Something like:
```
class WMRucioException(WMException):
    """
    _WMRucioException_
    Generic WMCore exception for Rucio
    """
    def __init__(self, message):
        self.myMessage = "WMRucioException: %s" % message
        super(WMRucioException, self).__init__(self.myMessage)
```

[1]
https://github.com/dmwm/WMCore/blob/ea35a220e9cd1277f52afb2db730e6a155ba7323/src/python/WMCore/Services/Rucio/Rucio.py#L23

todor-ivanov (approved):
@amaltaro  The comment I left in the code is related to an older piece of code. If you want to take a look that would be fine, if not... just ignore it. The current change looks good to me.

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: My understanding is, since we do not define a constructor `__init__` method here, when we create an instance of WMRucioException, it actually goes to the parent class constructor class (thus executing the WMException `__init__` method). Which is exactly what you're asking for here.

todor-ivanov: Yes, your understanding is correct and yes what I was talking about is 'almost' the same. There is one slight difference though. Notice the line:
`        self.myMessage = "WMRucioException: %s" % message` 
Which basically adds the child exception signature prefixing the message passed to the parent exception class. Which is kind of a simple way to reveal the exception hierarchy. But anyway it is not important, I already approved your change so no need to waste time on that.

amaltaro: Oh, I see. Yes, it isn't clear to me how useful it would be to have the exception type as part of the error message (especially for the "end user"). I think the most important is to have a clear error message (and be able to distinguish them, in case there are possibly many of them). That's a valid comment though, thanks.

amaltaro:
Thanks for the review, Todor.

amaltaro:
BTW, this needs to be backported to the agent branch, otherwise German won't be able to properly use his delete-blocks patch in the upcoming T0 release.

PR #10006, opened 2020-10-22
Deal with T0 output data subscription in RucioInjector

### LINKED ISSUE #9925: Support data subscriptions defined in the T0 machinery
**Impact of the new feature**
WMCore - WMAgent - T0

Every single workflow spec carries output data subscription information, which gets inserted into WMAgent's relational database, and then fetched and executed by the PhEDExInjector component.

Our initial plan for the Rucio migration was that, we would bypass those parameters in central production workflows (such that a generic non-custodial data placement gets done). Unfortunately this cannot be done for the T0 agent, which relies heavily on output data placement made by the agent and it exercises subscription code that isn't triggered in any other agents. Here is an example:
https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L933

A very detailed solution is still to be further discussed and specified.

However, it looks like we will have to implement the same output data placement functionality in the RucioInjector component. It will require some phedex-based attributes to be mapped to Rucio. Or maybe we could even redesign all those attributes and make them meaningful for the Rucio era.

none

**Additional context**
This GH issue touches the same subject, but from the central production view:
https://github.com/dmwm/WMCore/issues/9639

amaltaro:
Fixes #9925

#### Status
ready

#### Description
In short, it uses the information from `dbsbuffer_dataset_subscription` - by skipping a few columns only important to PhEDEx - and it creates container-level rules for those.

Expanding a bit further:
* no longer check whether a datatier is allowed to have a rule or not;
* kept the same functionality for central production data placement, thus
  * Tape placement is skipped and container is marked as "subscribed" in the local database
  * Disk has the RSE and a few other rule arguments override (copies, weight, grouping, etc)
* T0 Tape data placement (from the location defined in the database) is created as configured at the workflow level (using `T0 Tape` activity);
* T0 Disk data placement (from the location defined in the database) is created as configured at the workflow level (using `T0 Export` activity), plus the extra key/value pairs defined at the component configuration parameter `containerDiskRuleParams`
* rules are always tried first with `ask_approval=False`, if it fails, we set it to `True` and try again.

My original plan was to keep two different functions to deal with production and t0 output data placement, but code duplication is a bit too much...

NOTE: this is a simple port of PhEDEx subscription to Rucio rules. Real fix to the workflow ouput subscription schema, database schema changes, Tier0Feeder configuration, still needs to be done.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -410,75 +410,21 @@ class RucioInjectorPoller(BaseWorkerThread):
 
     def insertContainerRules(self):
         """
-        _insertContainerRules_
-        Poll the database for datasets meant to be subscribed and create
-        a container level rule to replicate all files to a given RSE
+        Polls the database for containers meant to be subscribed and create
+        a container level rule to replicate all the files to a given RSE.
+        It deals with both Central Production and T0 data rules, which require
+        a different approach, such as:
+          * Production Tape/Custodial data placement is skipped and data is marked as transferred
+          * Production Disk/NonCutodial has a generic RSE expression and some rules override
+            from the agent configuration (like number of copies, grouping and weight)
+          * T0 Tape is created as defined, with a special rule activity for Tape
+          * T0 Disk is created as defined, with a special rule activity for Disk/Export
         """
-        if self.isT0agent:
-            return self._insertContainerRulesT0()
-
         logging.info("Starting insertContainerRules method")
 
-        # FIXME also adapt the format returned by this DAO
-        # Check for completely unsubscribed datasets
-        # in short, files in phedex, file status in "GLOBAL" or "InDBS", and subscribed=0
-        unsubscribedDatasets = self.getUnsubscribedDsets.execute()
-
-        # Keep a list of subscriptions to tick as subscribed in the database
-        subscriptionsMade = []
-
-        # Create the subscription objects and add them to the list
-        # The list takes care of the sorting internally
-        for subInfo in unsubscribedDatasets:
-            container = subInfo['path']
-            rseExpr = subInfo['site']
-            # we skip Custodial/MSS/Tape data placement
-            if subInfo['custodial'] in [1, 'y']:
-                logging.info("Bypassing custodial container rule for container: %s and RSE: %s",
-                             container, rseExpr)
-                subscriptionsMade.append(subInfo['id'])
-                continue
-
-            if not self._isContainerTierAllowed(container):
-                logging.info("Component configured to skip container rule for: %s", container)
-                subscriptionsMade.append(subInfo['id'])
-                continue
-
-            kwargs = dict(ask_approval=False, activity="Production Output",
-                          account=self.rucioAcct, grouping="ALL",
-                          comment="WMAgent automatic container rule", meta=self.metaData)
-            # top it up with component-level parameters
-            kwargs.update(self.containerDiskRuleParams)
-
-            rseExpr = self.containerDiskRuleRSEExpr
-            if self.testRSEs:
-                rseExpr = rseExpr.replace("cms_type=real", "cms_type=test")
-            logging.info("Creating container rule for %s against RSE %s", container, rseExpr)
-            try:
-                resp = self.rucio.createReplicationRule(container,
-                                                        rseExpression=rseExpr, **kwargs)
-            except Exception as exc:
-                msg = "Failed to create container rule for: %s" % container
-                msg += "\nWill retry again in the next cycle. Error: %s" % str(exc)
-                continue
-            if resp:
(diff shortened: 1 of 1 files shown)


amaltaro:
I still have to test this patch (will have to backport it though), but it would be great to hear your opinion, German @germanfgv
If you are planning to have the ability to set a lifetime for rules created in replays, then I'd have to make some minor changes to this patch.

todor-ivanov (approved):
Looks good to me

amaltaro:
Tested this as a T0 and Production agent and everything looks fine.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -526,5 +487,25 @@ def _insertContainerRulesT0(self):
...
+        """
+        if not self.isT0agent and not rseName.endswith("_Tape"):
+            return "Production Output"
+        elif self.isT0agent and rseName.endswith("_Tape"):
+            return "T0 Tape"
+        elif self.isT0agent:
+            return "T0 Export"
+        else:
+            msg = "This code should never be reached. Report it to the developers. "
+            msg += "Trying to create container rule for WMAgent: {}, with RSE name: {}".format(self.isT0agent,
```
germanfgv, line 509: Is the first argument here right? At this point we know `self.isT0agent` is false.

amaltaro: @germanfgv did you mean L505? If not, please point me to the exact line or statement you're referring to.

germanfgv: I'm referring to L509. The Exception message.

amaltaro: You're right. Let me update this code.

germanfgv:
> I still have to test this patch (will have to backport it though), but it would be great to hear your opinion, German @germanfgv
> If you are planning to have the ability to set a lifetime for rules created in replays, then I'd have to make some minor changes to this patch.

Yes, the ability to set lifetime would be very useful fur us. Doing it through config.py is perfectly fine for us.

amaltaro:
@germanfgv German, I will make the necessary changes in this PR then. Just keep in mind that you will have to change the default configuration of:
```
config.RucioInjector.containerDiskRuleParams = {"weight": "ddm_quota", "copies": 2, "grouping": "DATASET"}
```
to something like
```
config.RucioInjector.containerDiskRuleParams = {"lifetime": 7 * 24 * 60 * 60}
```
in your replay instance.

And likely a change like:
```
config.RucioInjector.containerDiskRuleParams = {}
```
in the T0 production node.

If you think that's reasonable, I'll make this modification right away.

germanfgv:
> @germanfgv German, I will make the necessary changes in this PR then. Just keep in mind that you will have to change the default configuration of:
> 
> ```
> config.RucioInjector.containerDiskRuleParams = {"weight": "ddm_quota", "copies": 2, "grouping": "DATASET"}
> ```
> 
> to something like
> 
> ```
> config.RucioInjector.containerDiskRuleParams = {"lifetime": 7 * 24 * 60 * 60}
> ```
> 
> in your replay instance.
> 
> And likely a change like:
> 
> ```
> config.RucioInjector.containerDiskRuleParams = {}
> ```
> 
> in the T0 production node.
> 
> If you think that's reasonable, I'll make this modification right away.

I think this is reasonable. This is not a parameter That I think we will be changing frequently.

amaltaro:
@germanfgv 2nd commit contains the modifications we have just discussed. Can you please have another look at those changes? 

I'd also suggest you to create a T0 ticket to make sure the WMAgent configuration is properly updated during agent deployments, according to what we discussed above. Otherwise the agent might create improper rules...

Are you planning to run a replay with this patch. Or you need a new tag for that?

germanfgv:
> @germanfgv 2nd commit contains the modifications we have just discussed. Can you please have another look at those changes?
> 
> I'd also suggest you to create a T0 ticket to make sure the WMAgent configuration is properly updated during agent deployments, according to what we discussed above. Otherwise the agent might create improper rules...
> 
> Are you planning to run a replay with this patch. Or you need a new tag for that?

I'm running a replay with this PR to check if it is ok. I'll create a ticket concerning the deployment changes that need to me made to both production and replay procedure.

germanfgv (commented):
Everything looks fine for me. I ran 2 replays over the weekend. For the first replay the component behaved as expected. It created Container rules according to the specification of the Replay configuration file and deleted the rules of the blocks that had been transferred to their final destinations. The second replay had a different configuration and it is still running, but everything seems fine so far.

amaltaro:
Great!  @germanfgv  Have you also checked those "special" rule attributes defined for the replay instance?
I guess we can wait for the completion of your 2nd replay, before this gets merged. So please let us know how that goes.

PR #10008, opened 2020-10-22
[py2py3] Migrate WMCore/Credential

### LINKED ISSUE #10007: [py2py3] Migrate WMCore/Credential
**Impact of the new feature**

This involves migrating WMCore.Credential. All its dependencies have already been migrated.

Make WMCore.Credential compatible with both py2 and py3.

Run futurize on WMCore.Credential and then apply manual changes, if necessary.

mapellidario:
Fixes #10007 

#### Status

- src: ready
- test: ready

#### Description

Run futurize and then apply manual changes to WMCore/Credential

Summary

- Division: keep an eye on `old_div`, it may be dropped in favor of `//`
  - src/python/WMCore/Credential/Proxy.py
  - test/python/WMCore_t/Credential_t/Proxy_t.py
  - test/python/WMCore_t/Credential_t/MyProxy_t.py

- str : `form builtins import str` is used extensivvely in this PR
  - src/python/WMCore/Credential/Proxy.py
  - src/python/WMCore/Credential/SimpleMyProxy.py

- list and iterators: using backported version of `filter` and `range`, as described in the [python future docs](http://python-future.org/compatible_idioms.html#lists-versus-iterators)
  - src/python/WMCore/Credential/Proxy.py

Missing
- grouping import from builtins

It should be. However, as obtained from the [direct dependency graph](https://cms-dmwm-test.docs.cern.ch/py2py3/data/wmcore_gznzsw1eblR_direct_group_l2.txt), any bug would affect directly

```
 'WMCore_Credential': ['WMCore_BossAir',
                       'WMCore_Credential',
                       'WMComponent_AgentStatusWatcher'],
```

### DIFF
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -4,6 +4,9 @@ _Proxy_
 Wrap gLite proxy commands.
 """
 
+from __future__ import division
+from builtins import filter, str, range
+
 import contextlib
 import copy
 import os
@@ -64,7 +67,7 @@ def destroyListCred(credNameList=None, credTimeleftList=None, logger=None, timeo
     credTimeleftList = credTimeleftList or {}
     cleanCredCmdList = []
 
-    for credIdx in xrange(len(credNameList)):
+    for credIdx in range(len(credNameList)):
         hours, minutes, seconds = credTimeleftList[credIdx]
         timeleft = int(hours) * 3600 + int(minutes) * 60 + int(seconds)
         if timeleft == 0:
@@ -229,7 +232,7 @@ class Proxy(Credential):
         Uses openssl by default and fallback to voms-proxy-info in case of problems
         """
         timeleft = self.getUserCertTimeLeft(openSSL)
-        daystoexp = int(timeleft / (60. * 60 * 24))
+        daystoexp = int(timeleft // (60. * 60 * 24))
 
         return daystoexp
 
@@ -682,7 +685,7 @@ class Proxy(Credential):
 
         vomsValid = '00:00'
         if vomsTime > 0:
-            vomsValid = "%d:%02d" % (vomsTime / 3600, (vomsTime - (vomsTime / 3600) * 3600) / 60)
+            vomsValid = "%d:%02d" % (int(vomsTime // 3600), int((vomsTime % 3600) // 60))
         self.logger.debug('Requested voms validity: %s' % vomsValid)
 
         ## set environ and add voms extensions
--- a/src/python/WMCore/Credential/Credential.py
+++ b/src/python/WMCore/Credential/Credential.py
@@ -4,7 +4,8 @@ _Credential_
 Parent of all credentials class.
 Childs class should implement methods of this class.
 """
-class Credential:
+from builtins import object
+class Credential(object):
     """
     An abstract credential
     """
--- a/src/python/WMCore/Credential/SimpleMyProxy.py
+++ b/src/python/WMCore/Credential/SimpleMyProxy.py
@@ -12,6 +12,7 @@
 # Mattia Cinquilli <mcinquil@cern.ch>
 # on 2013/04/14 for integration on CRAB services.
 
+from builtins import str
 import logging
 import re
 import socket
--- a/src/python/WMCore/Credential/SimpleMyProxy.py
+++ b/src/python/WMCore/Credential/SimpleMyProxy.py
@@ -280,5 +280,5 @@ if __name__ == '__main__':
             import traceback
 
             traceback.print_exc()
-        logger.error("Error:" + str(e))
+        logger.error("Error: %s", str(e))
         sys.exit(1)
--- a/test/python/WMCore_t/Credential_t/MyProxy_t.py
+++ b/test/python/WMCore_t/Credential_t/MyProxy_t.py
@@ -5,6 +5,7 @@ Test the basic MyProxy operations.
 You need to source your UI before running these tests.
 The user Proxy and MyProxy is initialized in testCreateMyProxy method and they are used by the remaining tests.
 """
+from __future__ import division
 
 import unittest
 import os
@@ -91,7 +92,7 @@ class MyProxyTest(unittest.TestCase):
         self.proxy.renewMyProxy( proxy = self.proxyPath )
         time.sleep( 5 )
         timeLeft = self.proxy.getMyProxyTimeLeft( proxy = self.proxyPath )
-        self.assertEqual(int(timeLeft) / 3600, 167)
+        self.assertEqual(int(int(timeLeft) // 3600), 167)
 
     @attr("integration")
     def testRenewMyProxyForServer( self ):
@@ -102,7 +103,7 @@ class MyProxyTest(unittest.TestCase):
         self.proxy.renewMyProxy( proxy = self.proxyPath, serverRenewer = True )
         time.sleep( 5 )
         timeLeft = self.proxy.getMyProxyTimeLeft( proxy = self.proxyPath, serverRenewer = True )
-        self.assertEqual(int(timeLeft) / 3600, 167)
+        self.assertEqual(int(int(timeLeft) // 3600), 167)
 
     @attr("integration")
     def testMyProxyEnvironment(self):
--- a/test/python/WMCore_t/Credential_t/MyProxy_t.py
+++ b/test/python/WMCore_t/Credential_t/MyProxy_t.py
@@ -11,10 +11,7 @@ import unittest
 import os
 import logging
 import logging.config
-import socket
 import time
-import tempfile
-import subprocess
 
 from nose.plugins.attrib import attr
 from WMCore.Credential.Proxy import Proxy, myProxyEnvironment
--- a/test/python/WMCore_t/Credential_t/Proxy_t.py
+++ b/test/python/WMCore_t/Credential_t/Proxy_t.py
@@ -13,9 +13,7 @@ import unittest
 import os
 import logging
 import logging.config
-import socket
 import time
-import tempfile
 import subprocess
 
 from nose.plugins.attrib import attr
@@ -68,7 +66,7 @@ class ProxyTest(unittest.TestCase):
         if vomsProxyInfoCall.wait() != 0:
             return None
 
-        (stdout, stderr) = vomsProxyInfoCall.communicate()
+        stdout, _ = vomsProxyInfoCall.communicate()
         return stdout[0:-1]
 
     def getUserAttributes(self):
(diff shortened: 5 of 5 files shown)


amaltaro (changes requested):
Dario, we have been adopting python3 division behaviour through `__future__`. If I'm missing something, please enlighten me, otherwise we should replace `old_div` by the suggestion I made in the code. Thanks

todor-ivanov:
@mapellidario 
I tend to disagree with the following statement:

>Division: keep an eye on old_div, it may be dropped in favor of //

What I mean is, blindly substituting an operator may lead to an unexpected surprises. We all know the only difference between those two operators in Python2 and Python3, but let me just for clarity I paste the following table bellow:

```
   1 │|------------------------------------------------------------------------+------------------------------------------------------------------------|                                     
   2 │| (python2.venv) tivanov@tot-t590:tmp$ python                            | (python2.venv) tivanov@tot-t590:tmp$ python                            |                                     
   3 │| Python 2.7.16 (default, Oct 10 2019, 22:02:15)                         | Python 2.7.16 (default, Oct 10 2019, 22:02:15)                         |                                     
   4 │| [GCC 8.3.0] on linux2                                                  | [GCC 8.3.0] on linux2                                                  |                                     
   5 │| Type "help", "copyright", "credits" or "license" for more information. | Type "help", "copyright", "credits" or "license" for more information. |                                     
   6 │|                                                                        |                                                                        |                                     
   7 │| >>> from past.utils import old_div                                     | >>> from __future__ import division                                    |                                     
   8 │| >>> 5/2                                                                | >>> 5/2                                                                |                                     
   9 │| 2                                                                      | 2.5                                                                    |                                     
  10 │| >>> 5/2.0                                                              | >>> 5/2.0                                                              |                                     
  11 │| 2.5                                                                    | 2.5                                                                    |                                     
  12 │| >>> 5//2                                                               | >>> 5//2                                                               |                                     
  13 │| 2                                                                      | 2                                                                      |                                     
  14 │| >>> 5//2.0                                                             | >>> 5//2.0                                                             |                                     
  15 │| 2.0                                                                    | 2.0                                                                    |                                     
  16 │|------------------------------------------------------------------------+------------------------------------------------------------------------|                                     
```

Notice the (expected, but still subtle) difference between lines 9 and 15 in the two cases with and without the `__future__` module division. I mean if we just change the division with the one from future (`old_div` -> `__future__`) is one thing and could be considered  safe enough, but if we couple the change also with a an operator substitution (`/` -> `//`), then we suddenly complicate the 'transition matrix' and we can run into troubles due to over complication (and we may miss a place or two where we forget to take this into account).

mapellidario:
Dear Alan and Todor, thanks for your review.

In this PR we face the subtleties of division for the first time in WMCore and in some cases your experience can make the difference.

### why i chose `old_div`

I take `src/python/WMCore/Credential/Proxy.py` as an example. Right now in this file we do not `from __future__ import division`, so we have to assume that every division behaves like the old python2 division. (I think that the new kind of division has not been imported here because there may be a couple of places in this file where just importing a new kind of division may break things, as todor pointed out in his comment)

An example of simple change is at L232
```python
- daystoexp = int(timeleft / (60. * 60 * 24)) 
+ daystoexp = int(timeleft // (60. * 60 * 24))
```
Here it is obvious that `daystoexp` should be an integer, and despite `int(3.9) == 3` in both python2 and python3 (meaning that the lossy conversion from float to int keeps only the integer part, without any rounding up), using the  `//` operator makes this more explicit.

A different example is L685
```python
-  vomsValid = "%d:%02d" % (vomsTime / 3600, (vomsTime - (vomsTime / 3600) * 3600) / 60)
+ vomsValid = "%d:%02d" % (old_div(vomsTime, 3600), old_div((vomsTime - (old_div(vomsTime, 3600)) * 3600), 60))
```

vomsValid is later used as an argument to call the `voms-proxy-init`, so getting this string wrong can break the `voms-proxy-init` command. I have been as much conservativeand lazy  as possible by using `from past.utils import old_div` (this is provided by python future), which is a function that behaves like python2 division both in python2 and in python3.

I am sorry for the comment "Division: keep an eye on old_div, it may be dropped in favor of //", my intentions are not clear at all. I will try to be more clear in my PR messages next time. What I meant is that we may want to have a look at the usages of the division in this PR, understand where the result is intended to be a float (and use `/` after importing `from __future__ import division`) and where it is intended to be an integer (and use `int(a // b)`). This refactoring is not necessary, we can always be dependent on python-future. Since it is not necessary, I have just opted for the lazy options and kept the current situation by importing `old_div` so that the behaviour is consistent in python3 also.

### example of refactoring

With this is mind, we may want to have a second look at L685, I guess that `vomsTime` is in seconds, but the argument of `voms-proxy-init` is `${hours}:${minutes, two digits}"`.
This would mean that `vomsTime / 3600` is in hours (and this abuses the python2 division by loosing the fraction part and keeping only the decimal part when two integers are involved). If this is true, this means that in order to have consistent python2 and python3 behavior we have two options:

```python
- vomsTime / 3600
+ old_div(vomsTime, 3600)
# OR
- vomsTime / 3600
+ int(vomsTime // 3600)
```

With the same assumption as before, this `(vomsTime - (vomsTime / 3600) * 3600) / 60` seems like a way of computing the remaining minutes. `vomsTime / 3600` is what is already in the hours, `(vomsTime / 3600) * 3600` converts the what has already been put in the hours into seconds, `(vomsTime - (vomsTime / 3600) * 3600)` this keeps only the seconds that did not end up in the hours count, then we divide this by 60 to convert the remaining seconds in minutes (again, this is an abuse of the python2 lossy division between integers, that discards the fractional part). With this in mind, and in order to have consistent behaviour in python2 and in python3, we have some options

```python
- (vomsTime - (vomsTime / 3600) * 3600) / 60
+ int((vomsTime - int(vomsTime // 3600) * 3600) // 60)
# or
- (vomsTime - (vomsTime / 3600) * 3600) / 60
* int((vomsTime % 3600) // 60)
```

### conclusions

I can go through the PR again and try get rid of `old_div` and use `/ // %` to get consistent behavior in python2 aand pythhon3 instead. I am sorry for having been  bit lazy and conservative at the first round

src/python/WMCore/Credential/Proxy.py
```diff
@@ -3,6 +3,8 @@
 _Proxy_
 Wrap gLite proxy commands.
 """
+from builtins import filter, str, range
+from past.utils import old_div
```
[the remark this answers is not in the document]
mapellidario: `from past.utils import old_div`: this is a function that behaves like python2 division in both python2 and pythhon3. I chose it to have consistent behaviur, driven by laziness and being conservative (do the minimal set of changes possible, without refactoring that may break the code). This import has no side effects.

`from __future__ import division`: this make the binary operator `/`to behave like python3 division in both python2 and python3, with the consequences highlighted by todor in his comment

todor-ivanov:
Hi @mapellidario I understand  what you mean and in general I would be quite supportive in your suggestion, like making an implicit floor operation hidden behind the behavior of the division function itself to become more obvious and rely on everybody's knowledge about the operator behavior. But in  this case as you can see now we suggest a shortcut - like just substituting the operator, and later we forget that this change affected this particular piece of code and then we decide to have a full switch to the `__ future__` division .... Not telling this is going to create a problem, but I'd rather not try to remember a past change until the actual transition happens and would directly move to the new division behavior and avoid the operator substitution at all. This is not a strong disagreement with your suggestion... more like ... expressing myself something like : "I would do it the other way around". :)

mapellidario:
Ok, so I will add `from __future__ import division`, but this is going to break lines like 

```python
 vomsValid = "%d:%02d" % (vomsTime / 3600, (vomsTime - (vomsTime / 3600) * 3600) / 60)
```

So, if you want me to add `from __future__ import division`, then I will have to make all the necessary adjustments/refactoring. I agree that this is the cleanest solution, so new commits will come in the afternoon. 

Thanks again!

amaltaro (approved):

amaltaro:
Thank you very much Dario.
FYI @belforte Stefano, something to keep in mind analysis services start using WMCore >= 1.4.2.pre7 (py2/py3 compatible code).

src/python/WMCore/Credential/Proxy.py
```diff
@@ -682,7 +685,7 @@ def vomsExtensionRenewal(self, proxy, voAttribute='cms'):
 
         vomsValid = '00:00'
         if vomsTime > 0:
-            vomsValid = "%d:%02d" % (vomsTime / 3600, (vomsTime - (vomsTime / 3600) * 3600) / 60)
+            vomsValid = "%d:%02d" % (int(vomsTime // 3600), int((vomsTime % 3600) // 60))
```
amaltaro, line 688: For the record, am I wrong or you actually made a bug fix here?

belforte: doesn't python have any standard lib to convert seconds to hours:min format ? Learning proper use of // vs old_div vs % is surely good, but should not be needed here.

belforte:
> FYI @belforte Stefano, something to keep in mind analysis services start using WMCore >= 1.4.2.pre7 (py2/py3 compatible code).

I am not sure I understand. Sorry my being thick. Do you simply mean "beware when you will try to get rid of separate CRAB branch that there will be changes in Credential as well" ?

BTW, since we are on topic, I think we'll spend November on K8s, Rucio and NewPublisherForCrab and I'll start looking at py2topy3 for Crab lin a month or so. If @amaltaro  or @mapellidario have time to think about the matter in the meanwhile, the critical part IMHO is this: https://github.com/dmwm/CRABClient/issues/4940  Especially in light of the thing that CRAB Client should keep working also in Py2 environmenr for those CMSSW releases which will stay there. See also [this](https://github.com/dmwm/CRABServer/issues/6204#issuecomment-714770309) comment from KevinL

amaltaro:
> I am not sure I understand. Sorry my being thick. Do you simply mean "beware when you will try to get rid of separate CRAB branch that there will be changes in Credential as well" ?

Yes, just something to keep in mind when you upgrade imported WMCore to the latest production release/HEAD in the future.

src/python/WMCore/Credential/Proxy.py
mapellidario: We could use `time` or `datetime`, but I can not find a way to use `strftime` to print the hours without overloading into the day field if `%H > 23`. For example

```python
# This is ok
>>> import time
>>> time.strftime('%H:%M', time.gmtime(10000))
'02:46'
>>> str(datetime.timedelta(seconds=10000))
'2:46:40'

# this is less ok
>>> time.strftime('%H:%M', time.gmtime(100000))
'03:46'
>>> str( datetime.timedelta(seconds=100000) )
'1 day, 3:46:40'
```

If our proxies expire in less than 24 hours than this is a valid option, even though it feels like a bit binding.

Did you have something else in mind? Am I missing a `strftime` parameter?

PR #10011, opened 2020-10-22
[py2py3] WMCore/Algorithms - migration

mapellidario:
Fixes #10009 

#### Status

- src: ready
- test: not necessary

#### Description

Run futurize on WMCore/Algorithms

Changes:

str: `from builtins import str` used in
- src/python/WMCore/Algorithms/MathAlgos.py: used in `decimal.Decimal(str(stdDev)).is_finite()`
- src/python/WMCore/Algorithms/SubprocessAlgos.py

list and iterators: range
- src/python/WMCore/Algorithms/MathAlgos.py

[Iterate over dictionariy keys](http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items) used in:
- src/python/WMCore/Algorithms/MiscAlgos.py

[iterate over dictionary items](http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items) used in:
- src/python/WMCore/Algorithms/ParseXMLFile.py

oct: `from builtins import oct`
- src/python/WMCore/Algorithms/Permissions.py

#### Is it backward compatible (if not, which system it affects?)

It should be. However, any bug would affect directly, 

```
'WMCore_Algorithms': ['WMComponent_TaskArchiver',
                       'WMCore_DataStructs',
                       'WMComponent_DBS3Buffer',
                       'WMCore_FwkJobReport',
                       'WMCore_WorkQueue',
                       'WMCore_ACDC',
                       'WMCore_Algorithms',
                       'WMCore_WMSpec',
                       'WMCore_Services',
                       'WMCore_Storage',
                       'WMCore_WMRuntime'],
```

#### External dependencies / deployment changes

requires python-future

PR #10012, opened 2020-10-22
[py2py3] Migrate WMCore/DataStructs

### LINKED ISSUE #10010: [py2py3] Migrate WMCore/DataStructs
**Impact of the new feature**

This involves migrating WMCore.DataStructs. Since this directory depends on WMCore.Algorithms, this issue has to be addressed _after_ #10009 

Make WMCore.DataStructs compatible with both py2 and py3.

Run futurize on WMCore.DataStructs and then apply all the necessary manual changes.

mapellidario:
Fixes #10010 

#### Status

- src: ready
- test: ready

#### Description

Run futurize on WMCore/DataStructs and applied some manual changes.

Since WMCore/DataStructs depends on WMCore/Algorithms. this PR is continuously rebases onto #10011 .

Open questions

DASDocument
- `dict` keyword abused?

Canges

[make a list of dictionary keys](http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list)
- src/python/WMCore/DataStructs/DASDocument.py
- src/python/WMCore/DataStructs/JobGroup.py
- src/python/WMCore/DataStructs/LumiList.py

[iterate over dictionary items and values](http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items):
- src/python/WMCore/DataStructs/LumiList.py
- src/python/WMCore/DataStructs/Run.py
- test/python/WMCore_t/DataStructs_t/LumiList_t.py
- test/python/WMCore_t/DataStructs_t/MathStructs_t/ContinuousSummaryHistogram_t.py

[list of dictionary items](http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list)
- src/python/WMCore/DataStructs/Run.py
- src/python/WMCore/DataStructs/WorkUnit.py

basestring: ` from past.builtins import basestring `
- src/python/WMCore/DataStructs/File.py

str: `from builtins import str,`
- src/python/WMCore/DataStructs/Fileset.py
- src/python/WMCore/DataStructs/LumiList.py
- src/python/WMCore/DataStructs/MathStructs/SummaryHistogram.py
- src/python/WMCore/DataStructs/Workflow.py
- test/python/WMCore_t/DataStructs_t/Fileset_t.py
- test/python/WMCore_t/DataStructs_t/LumiList_t.py
- test/python/WMCore_t/DataStructs_t/Subscription_t.py

[list and iterators](http://python-future.org/compatible_idioms.html#lists-versus-iterators): map, range, zip
- src/python/WMCore/DataStructs/Fileset.py
- src/python/WMCore/DataStructs/Job.py
- src/python/WMCore/DataStructs/LumiList.py
- src/python/WMCore/DataStructs/Mask.py
- src/python/WMCore/DataStructs/Subscription.py
- test/python/WMCore_t/DataStructs_t/JobPackage_t.py
- test/python/WMCore_t/DataStructs_t/Job_t.py
- test/python/WMCore_t/DataStructs_t/LumiList_t.py
- test/python/WMCore_t/DataStructs_t/MathStructs_t/ContinuousSummaryHistogram_t.py
- test/python/WMCore_t/DataStructs_t/MathStructs_t/DiscreteSummaryHistogram_t.py
- test/python/WMCore_t/DataStructs_t/Subscription_t.py

division: old_div
- src/python/WMCore/DataStructs/MathStructs/ContinuousSummaryHistogram.py
- test/python/WMCore_t/DataStructs_t/MathStructs_t/ContinuousSummaryHistogram_t.py

File
- `from past.builtins import basestring`

Fileset
- from builtins import str, map

Job
- from builtins import map, range

JobPackage
- migration from cPickle to pickle have been set back

LumiList
- dictionaries http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items and http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list
- list and iterators: zip, range: http://python-future.org/compatible_idioms.html#lists-versus-iterators
- from builtins import str
- from builtins import object
- from builtins import next

Mask
- dict iteration
- from bultins import range + list(range()) : http://python-future.org/compatible_idioms.html#range

Run
- viewitems
- dictionary items as list: http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list

Subscription
- range

Workflow
- from builtins import str

WorkUnit
- from future.utils import listitems - TODO investigate `listitmes(self)`

ContinuousSummary
- keep `old_div` not to break anything

Summary histogram
- from builtins import str

It should be. However, as obtained from the [direct dependency graph](https://cms-dmwm-test.docs.cern.ch/py2py3/data/wmcore_gznzsw1eblR_direct_group_l2.txt), any bug would affect directly

```
 'WMCore_DataStructs': ['WMComponent_TaskArchiver',
                        'WMCore_WMBS',
                        'WMComponent_JobSubmitter',
                        'WMCore_JobStateMachine',
                        'WMCore_JobSplitting',
                        'WMCore_WebTools',
                        'WMCore_DataStructs',
                        'WMCore_Cache',
                        'WMComponent_DBS3Buffer',
                        'WMCore_FwkJobReport',
                        'WMCore_WorkQueue',
                        'WMCore_MicroService',
                        'WMCore_ACDC',
                        'WMCore_Database',
                        'WMCore_WMSpec',
                        'WMCore_Services',
                        'WMCore_WMRuntime'],
```

requires python-future

### DIFF
--- a/src/python/WMCore/DataStructs/DASDocument.py
+++ b/src/python/WMCore/DataStructs/DASDocument.py
@@ -13,7 +13,7 @@ class DASDocument(dict):
         """
         Make an empty dictionary representing a DAS document
         """
-        dict = {"request_timestamp": 0,
+        dasdoc = {"request_timestamp": 0,
                 "request_url": "",
                 "request_method": "GET",
                 "request_params": {},
@@ -22,8 +22,8 @@ class DASDocument(dict):
                 "response_checksum": "",
                 "request_api": "",
                 "call_time": 0}
-        self.keys = dict.keys()
-        self.setdefault(service, dict)
+        self.keys = list(dasdoc)
+        self.setdefault(service, dasdoc)
 
     def compare(self, dict):
         """
--- a/src/python/WMCore/DataStructs/Mask.py
+++ b/src/python/WMCore/DataStructs/Mask.py
@@ -10,6 +10,7 @@ job in two ways:
 
 """
 
+from builtins import range
 from WMCore.DataStructs.Run import Run
 
 
@@ -139,7 +140,7 @@ class Mask(dict):
         if not isinstance(lumis, list):
             lumis = list(lumis)
 
-        if not run in self['runAndLumis'].keys():
+        if run not in self['runAndLumis']:
             self['runAndLumis'][run] = []
 
         self['runAndLumis'][run].append([min(lumis), max(lumis)])
@@ -167,7 +168,7 @@ class Mask(dict):
             # ALWAYS TRUE
             return True
 
-        if not run in self['runAndLumis'].keys():
+        if run not in self['runAndLumis']:
             return False
 
         for pair in self['runAndLumis'][run]:
@@ -208,7 +209,7 @@ class Mask(dict):
                 if pair[0] == pair[1]:
                     maskLumis.add(pair[0])
                 else:
-                    maskLumis = maskLumis.union(range(pair[0], pair[1] + 1))
+                    maskLumis = maskLumis.union(list(range(pair[0], pair[1] + 1)))
 
             filteredLumis = set(runDict[runNumber].lumis).intersection(maskLumis)
             if len(filteredLumis) > 0:
--- a/src/python/WMCore/DataStructs/MathStructs/ContinuousSummaryHistogram.py
+++ b/src/python/WMCore/DataStructs/MathStructs/ContinuousSummaryHistogram.py
@@ -8,6 +8,7 @@ Created on Nov 20, 2012
 
 @author: dballest
 """
+from __future__ import division
 import math
 
 from WMCore.DataStructs.MathStructs.SummaryHistogram import SummaryHistogram
--- a/src/python/WMCore/DataStructs/File.py
+++ b/src/python/WMCore/DataStructs/File.py
@@ -5,6 +5,7 @@ _File_
 Data object that contains details for a single file
 
 """
+from past.builtins import basestring
 __all__ = []
 
 from WMCore.DataStructs.Run import Run
--- a/src/python/WMCore/DataStructs/Fileset.py
+++ b/src/python/WMCore/DataStructs/Fileset.py
@@ -7,6 +7,8 @@ Data object that contains a set of files
 """
 
 from __future__ import print_function
+from builtins import str, map
+
 from operator import itemgetter
 from WMCore.DataStructs.WMObject import WMObject
 __all__ = []
--- a/src/python/WMCore/DataStructs/Job.py
+++ b/src/python/WMCore/DataStructs/Job.py
@@ -6,6 +6,8 @@ Data object that describes a job
 
 """
 
+from builtins import map, range
+
 __all__ = []
 
 
--- a/src/python/WMCore/DataStructs/JobGroup.py
+++ b/src/python/WMCore/DataStructs/JobGroup.py
@@ -128,9 +128,7 @@ class JobGroup(WMObject):
         """
         This just gets a length for either dict or list objects
         """
-        if isinstance(obj, dict):
-            return len(obj.keys())
-        elif isinstance(obj, list):
+        if isinstance(obj, (dict, list)):
             return len(obj)
         else:
             return 0
--- a/src/python/WMCore/DataStructs/LumiList.py
+++ b/src/python/WMCore/DataStructs/LumiList.py
@@ -13,6 +13,10 @@ This code began life in COMP/CRAB/python/LumiList.py
 """
 
 
+from builtins import zip, range, next, str, object
+
+from future.utils import viewitems
+
 from future import standard_library
 standard_library.install_aliases()
 
@@ -77,12 +81,12 @@ class LumiList(object):
         if isinstance(runsAndLumis, list):
             queued = {}
             for runLumiList in runsAndLumis:
-                for run, lumis in runLumiList.items():
+                for run, lumis in viewitems(runLumiList):
                     queued.setdefault(run, []).extend(lumis)
             runsAndLumis = queued
 
         if runsAndLumis:
-            for run in runsAndLumis.keys():
+            for run in runsAndLumis:
                 runString = str(run)
                 lastLumi = -1000
                 lumiList = runsAndLumis[run]
@@ -104,7 +108,7 @@ class LumiList(object):
                 self.compactList[runString] = [[1, 0xFFFFFFF]]
 
         if compactList:
-            for run in compactList.keys():
+            for run in compactList:
                 runString = str(run)
                 if compactList[run]:
                     self.compactList[runString] = compactList[run]
@@ -124,9 +128,9 @@ class LumiList(object):
             if len(runs) <= 0 or len(lumis) != len(runs):
                 raise RuntimeError('Improper format for wmagentFormat. # of lumi lists must match # of runs')
 
-            for run, lumiString in itertools.izip(runs, lumis):
+            for run, lumiString in zip(runs, lumis):
                 runLumis = lumiString.split(',')
-                if not str(run) in self.compactList:
+                if str(run) not in self.compactList:
                     self.compactList[str(run)] = []
                 if len(runLumis) % 2:
                     raise RuntimeError('Improper format for wmagentFormat. Lumis must be in pairs')
@@ -139,7 +143,7 @@ class LumiList(object):
 
         # Compact each run and make it unique
 
-        for run in self.compactList.keys():
+        for run in self.compactList:
             newLumis = []
             for lumi in sorted(self.compactList[run]):
                 # If the next lumi starts inside or just after the last just change the endpoint of the first
@@ -210,8 +214,8 @@ class LumiList(object):
 
     def __or__(self, other):
         result = {}
-        aruns = self.compactList.keys()
-        bruns = other.compactList.keys()
+        aruns = list(self.compactList)
+        bruns = list(other.compactList)
         runs = set(aruns + bruns)
         for run in runs:
             overlap = sorted(self.compactList.get(run, []) + other.compactList.get(run, []))
@@ -274,7 +278,7 @@ class LumiList(object):
         Return the list of pairs representation
         """
         theList = []
-        runs = self.compactList.keys()
+        runs = list(self.compactList)
         runs.sort(key=int)
         for run in runs:
             lumis = self.compactList[run]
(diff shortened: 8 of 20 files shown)


src/python/WMCore/DataStructs/DASDocument.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, service):
         """
         Make an empty dictionary representing a DAS document
         """
-        dict = {"request_timestamp": 0,
```
mapellidario, line 16: @amaltaro This change was not suggested by futurize, it was mine.
Is the redefinition of the reserved keyword `dict` intentional? It may not cause any problem, but safe is better than sorry.
I also took the liberty of renaming it to `dasdoc`, please suggest a new and better name if you do not like this one.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
finally :-)

mapellidario:
Yey! thanks!

Looks like there a lot of suggestions from pylint. Do you want me to tackle the more obvious ones?

amaltaro:
If it's bold, super safe and straight forward, yes please go for it. If all of them take more than 15min of your time though, then we might reserve it for a future cleanup (one day I'll try to clean as many of those as I can!)

mapellidario:
I had a look and only a handful of those are simple and safe, many other could easily break something if I am not super careful, such as changing the name of a keyword argument in a class member. Let's say that it is a bit too risky at this stage

klannon:
Regarding pylint, I think this is a good example of something where we need to be intentional about how we proceed.  It could be a good thing to work on, but it's a big task and we need to balance it against other priorities.  I've put it in the list of possible priorities for 2021 under "WM maintenance."  Maybe it would be better under another category, but at least it's on our list.  When we review our quarterly priorities, we can decide if we want to break it into appropriately sized chunks and tackle put them in the to-do list.

src/python/WMCore/DataStructs/DASDocument.py
amaltaro: This module is not used anywhere (and apparently not by any of the external services either). We can keep these changes, and I will delete this module in another PR. Or we delete it in another PR, and you remove these changes.
I assume the first option gives you less work.

src/python/WMCore/DataStructs/Mask.py
```diff
@@ -139,7 +140,7 @@ def addRunAndLumis(self, run, lumis=None):
         if not isinstance(lumis, list):
             lumis = list(lumis)
 
-        if not run in self['runAndLumis'].keys():
+        if not run in self['runAndLumis']:
```
amaltaro, line 143: For future PRs, I'd suggest the construction something not in, e.g.:
```
if run not in self['runAndLumis']:
```

No need to update this PR only for these 2 cases though (including the one below)

src/python/WMCore/DataStructs/MathStructs/ContinuousSummaryHistogram.py
```diff
@@ -8,6 +8,8 @@
 
 @author: dballest
 """
+from __future__ import division
+from past.utils import old_div
```
amaltaro, line 12: How about we use only future division here?

amaltaro (commented):
Dario, these changes look good to me. I have made a couple of suggestions/questions in the code though, which might trigger further changes here.

Note-1: DataStructs/DASDocument will be deleted in the coming days.
Note-2: I never know the exact differences between viewitems and listitems. Is it that listitems expands all the keys/values in memory (like the py2 items), while viewitems is sort of generating keys/values on the fly?

mapellidario:
### future and dictionaries

The main idea behind future's management of the access to dictionaries is to have consistent behavior in py2 and py3.

Future provides two kinds of operators for accessing dictionaries

1. operators to iterate over a dictionary
  - [examples - future documentation](https://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items)
  - [implementation - python-future github](https://github.com/PythonCharmers/python-future/blob/28b336151e991ddb8d8a4fb48c48f0e2b9784f55/src/future/utils/__init__.py#L269)
  - I chose to use `for keys in mydict:`, `future.utils.viewvalues` and `future.utils.viewitems`, since in the implementation they suggest `future.utils.viewvalues` instead of `future.utils.itervalues` (this last option is what they have in the documentation, but I trust the docstrings in the implementation more than the documentation). 
  - `future.utils.view*` functions behave like `view*` in python2 and like `.items()` and `.values()` in python3, to have the same behavior is py2 and py3.

2. operators to retrieve a list of data from a dictionary
  - [examples - future doc](https://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list)
  - [implementation - python-future github](https://github.com/PythonCharmers/python-future/blob/28b336151e991ddb8d8a4fb48c48f0e2b9784f55/src/future/utils/__init__.py#L704)
  - I chose to use `keylist = list(mydict)`, `future.utils.listvalues` and `future.utils.listitems`
  - `future.utils.listvalues` and `future.utils.listitems` make sure that `.values()` and `.items()` iterators are expanded in py3, so that py3 behaves like python2

It is possible that there would not be any downside in passing iterators around in py3 instead of lists, but maybe it is better to investigate this after the modernization to py2 and py3 compatible code is complete. Do you agree?

Nonetheless, I used `future.utils.listitems` in DataStructs/Run.py

```python
- value += hash(frozenset(self.eventsPerLumi.items()))
+ value += hash(frozenset(listitems(self.eventsPerLumi)))
```

and having a second look at these lines I realize that `frozenset` would expand the iterator over the items in python3 as well, so using `listitems` may be redundant in this case.

### hash

While testing the behavior `hash(frozenset())`, I noticed that in python3 
`hash(frozenset({"hello": "world"}.items()))` returns a different result every time a new python3 interpreter is started. This matches with what the [python docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__) says in the "Note". Meanwhile, [python2 docs](https://docs.python.org/2/reference/datamodel.html#object.__hash__) does not have such a "note" and the hash is always the same.

Would this be a problem for us? Do we rely on some hashes to be the same every time a program is run?

src/python/WMCore/DataStructs/Mask.py
mapellidario: you are correct, I promise I have explicitely added this to the list of checks before requesting a review. :sweat_smile: 
I made these changes.

src/python/WMCore/DataStructs/MathStructs/ContinuousSummaryHistogram.py
mapellidario: I left old_div because I was not sure how to change it. I tried to use python3 division, let's see if jenkins agrees with my intuition

mapellidario:
test this please (in my docker UUID_t.py is  passing both tests)

mapellidario:
test this please

src/python/WMCore/DataStructs/MathStructs/ContinuousSummaryHistogram.py
mapellidario: It seems that jenkins liked the changes!

amaltaro:
Thanks for this explanation, Dario. I fully agree with having a consistent behaviour between python2 and python3 stack, and we should - whenever possible - to opt for the python3 behaviour.

The second concern I always have with python dictionaries is that we should make sure to:
* use `listitems` if we need to update the dictionary while iterating over it (since listitems expands the dictionary into a pair of lists with the keys and values - thus, memory hungry!)
* use `viewitems` or `iteritems` if we are concerned about memory footprint and changes to the dictionary itself are not made while iterating through it.

> and having a second look at these lines I realize that frozenset would expand the iterator over the items in python3 as well, so using listitems may be redundant in this case.

I guess I would leave the `listitems` in the code, it somehow seems more ...clear/complete.

> Would this be a problem for us? Do we rely on some hashes to be the same every time a program is run?

I think with the current changes, we guarantee that the dictionary hash will be the same either in py2 or py3, right? I had to google for it, and this hash method seems to be quite important, including to evaluate objects equality (and other things that I either did not fully understand or have not read carefully).

mapellidario:
`listitems` vs `viewitems` / `iteritems`: sure, you are correct. 

> we guarantee that the dictionary hash will be the same either in py2 or py3

I do not think that it is possible to have consistent hashing results between py2 and py3 using the standard `hash()` function. In python 3 it is always going to use salting that make the hash consistent only inside the same python interpreter process.

```bash
> python2.7
Python 2.7.18 (default, Apr 20 2020, 20:30:41) 
[GCC 9.3.0] on linux2
>>> hash(frozenset({"hello": "world"}.items()))
-1980536423468450193
> python2.7
Python 2.7.18 (default, Apr 20 2020, 20:30:41) 
[GCC 9.3.0] on linux2
>>> hash(frozenset({"hello": "world"}.items()))
-1980536423468450193

> python3
Python 3.8.6 (default, Sep 25 2020, 09:36:53) 
[GCC 10.2.0] on linux
>>> hash(frozenset({"hello": "world"}.items()))
3543227547196071832
> python3
Python 3.8.6 (default, Sep 25 2020, 09:36:53) 
[GCC 10.2.0] on linux
>>> hash(frozenset({"hello": "world"}.items()))
-4045192331877773067
```

If we want the hashes to be the same in both py2 and py3, we have to look for an alternative, for example using a specific hash function from hashlib:

```bash
> python2.7
Python 2.7.18 (default, Apr 20 2020, 20:30:41) 
[GCC 9.3.0] on linux2
>>> import hashlib
>>> hashlib.sha256(unicode({"hello": "world"}).encode("utf-8")).hexdigest()
'08b8bdfc96c88412e3e4c9ce471899037be4f44b4d737fbdb85407bb3e66a331'

> python3
Python 3.8.6 (default, Sep 25 2020, 09:36:53) 
[GCC 10.2.0] on linux
>>> import hashlib
>>> hashlib.sha256(bytes(str({"hello": "world"}), "utf-8")).hexdigest()
'08b8bdfc96c88412e3e4c9ce471899037be4f44b4d737fbdb85407bb3e66a331'
```

amaltaro (approved):
Thanks Dario. I just noticed some pylint warning and even errors that we should address (unused imports, and undefined stuff). Those are unrelated to your changes though, so I'll just pretend I have not seen those and we get them fixed in the future ;)

PR #10028, opened 2020-10-28
a simple rucio.getPfn

belforte:
Fixes #9873 

#### Status
ready

#### Description
add a rucio.getPFN method for use by CRAB (which resolves an LFN to a PFN)

#### Is it backward compatible
Yes, it is new node. Nothing is removed.

#### Related PRs
#9977 earlier attempt by @dciangot to address same issue. To be closed.

#### External dependencies / deployment changes
None

#### References in CRAB repository
https://github.com/dmwm/CRABServer/issues/6102

PR #10029, opened 2020-10-28
a simple rucio.getPfn. Fix #9873

### LINKED ISSUE #9873: Integrate LFN2PFN for Rucio
**Impact of the new feature**
At least CRAB

Integrate  https://rucio.readthedocs.io/en/latest/api/rse.html#rucio.client.rseclient.RSEClient.lfns2pfns into Rucio WMCore class

**Additional context**
CRAB ticket: https://github.com/dmwm/CRABServer/issues/6102

belforte:
Port to Master branch of  #10028  Fixes #9873  

#### Status
ready

#### Description
add a rucio.getPFN method for use by CRAB (which resolves an LFN to a PFN)

#### Is it backward compatible
Yes, it is new node. Nothing is removed.

#### Related PRs
 #10028

None

### DIFF
--- a/test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py
+++ b/test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py
@@ -23,6 +23,7 @@ PUBLOCK = "/WPhi_2e_M-10_H_TuneCP5_madgraph-pythia8/RunIIAutumn18NanoAODv6-Nano2
 DSET2 = "/Mustar_MuG_L10000_M-3750_TuneCP2_13TeV-pythia8/RunIIFall17NanoAODv6-PU2017_12Apr2018_Nano25Oct2019_102X_mc2017_realistic_v7-v1/NANOAODSIM"
 BLOCK2 = "/Mustar_MuG_L10000_M-3750_TuneCP2_13TeV-pythia8/RunIIFall17NanoAODv6-PU2017_12Apr2018_Nano25Oct2019_102X_mc2017_realistic_v7-v1/NANOAODSIM#50aecec1-d5a8-4756-9ee4-f93995c5b524"
 
+
 # integration containers with with 2 blocks
 # DSET3 = "/NoBPTX/Integ_Test-ReReco_LumiMask_HG1812_Validation_TEST_Alan_v13-v11/AOD"
 # DSET4 = "/NoBPTX/Integ_Test-ReReco_LumiMask_HG1812_Validation_TEST_Alan_v13-v11/DQMIO"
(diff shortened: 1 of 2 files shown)


belforte:
@amaltaro how can I see the unit test output or run it myself so I can fix it ? Jenkins log only says
```
    WMCore_t.Services_t.Rucio_t.Rucio_t.RucioTest:testGetPFN changed from success to error
```
IIUC that this is current test, failure is not a surprise ! But you menationed you had something already... I'd like to start from that.
https://github.com/dmwm/WMCore/blob/fa4111f3e3ddeba76eee611669ed7a185b8bd988/test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py#L236-L240

The pylint complaint is abount code which I did not touch, I let you decide if to fix it or keep it.

amaltaro:
That unit test was expected to fail. Right now it expects a call to `getPFN` to raise an exception, which is no longer the case now that you have implemented that method. Here is an initial unit test to test that method
```
    def testGetPFN(self):
        """
        Test `getPFN` method
        """
        resp = self.myRucio.getPFN(site="T2_CH_CERN_Test", lfns="/store")
        self.assertItemsEqual(resp, {u'/store': u'gsiftp://eoscmsftp.cern.ch:2811/eos/cms/store/test/rucio/int/cms//store'})

        # request below should raise an exception or something, according to the docstring...
        resp = self.myRucio.getPFN(site="T2_CH_CERN_Test", lfns="wrong_lfn")
        self.assertItemsEqual(resp, {})
```

About the pylint issues, they seem to be all "expected". But if there are new ones reported for the lines you have changed, it would be great if you could get them fixed.

About interpreting and debugging jenkins unit tests, this seems to be up-to-date:
https://github.com/dmwm/WMCore/wiki/Understanding-Jenkins#how-to-interpret-jenkinss-comments-on-your-pull-request

you should be able to test those by running the WMAgent container, IF Rucio authentication was working properly from the container... something changed a few weeks ago and I still didn't manage to get that fixed. It will be easier if you give me some examples of `getPFN` calls and I write the unit tests for you. Otherwise, you can write unit tests and wait for Jenkins results too.

belforte:
sorry, but I can't also learn to run WMA Container. WIll leave to you. I will add some unit test though.

test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py
```diff
@@ -237,7 +237,26 @@ def testGetPFN(self):
         """
         Test `getPFN` method
         """
-        self.assertRaises(NotImplementedError, self.myRucio.getPFN)
+        # simplest call
+        resp = self.myRucio.getPFN(site="T2_CH_CERN", lfns="/store")
```
amaltaro, line 241: Stefano, quick comment here. This Rucio object is connected to the Rucio integration server, which seems to be unable to resolve LFN x PFN for the main RSE names. I don't know the reason neither whether it's expected or not, however, could you please update your RSE names to actually their `_Test` correspondents? Thanks

belforte: thanks Alan, but...
- I found in test log: getPFN() takes at least 3 arguments (1 given) 
   so maybe something else is going on (too)
- this lnf2pfn resolution fails a lot also in main server and I pissed off.
- I do not know how to use test server so that I can first test things in my python shell, and I have no confidence that "all _Test" sites will be there. For testing protocol (scheme in Rucio) only a few sites can do.

test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py
```diff
@@ -237,7 +237,26 @@ def testGetPFN(self):
         """
         Test `getPFN` method
         """
-        self.assertRaises(NotImplementedError, self.myRucio.getPFN)
+        # simplest call
+        resp = self.client.getPFN(site="T2_CH_CERN", lfns="/store")
```
amaltaro, line 241: You're actually using the wrong object here. Please replace `self.client` by `self.myRucio` in all these calls below.

belforte: I had `self.myRucio` before and got the `getPFN() takes at least 3 arguments (1 given)`   I am quite confused.
Anyhow I am doing as you suggest.  Let me also comment all my additional things and only leave your example. I need to work on CRAB code.

amaltaro:
Latest error reported by jenkins is:
```
invalid syntax (Rucio_t.py, line 274)
```

Don't spend more time on this though. I'm going to pull your branch and make those unit tests to work, then I get back to you and you push those changes in.

belforte:
thanks Alan, but I could not leave with with such stupid typo ! Anyhow, you can also edit the file in my PR in github
save some push/pull etc.

amaltaro:
While I test it, I get even different results. One time it returns gsiftp, another returns root:, even though my test did not change...

belforte:
from CERN ? hard to test a non-deterministic thing. Maybe that's the drawback of connecting to a test server where people change stuff.  We are trying to write a Rucio test, not a WMCore one.

amaltaro:
Right! While we write these wmcore unit tests, it's a great opportunity to ensure we are getting the expected response back from the server, especially for "freshly" commissioned services like Rucio.

Stefano, I have made 2 different unit tests, one that will keep running in every jenkins test; and a second one marked as `integration` that can only be run manually, this one pointing to the production server though. Here is the code:
http://amaltaro.web.cern.ch/amaltaro/forStefanoB/Rucio_t.py

can you curl/wget it and push it into this branch? Please also take your time to go through those tests and make sure the outcomes satisfies the requirements you have discussed with the CMS Rucio team. I particularly spotted 2 issues in there, tagged as FIXME. And please, adapt those comments/etc as you like.

belforte:
I updated my branch with that, so code does not get lost. Will look at details and possibly change later on a let you know. OTOH, I guess that when you will  change something else and get a unit test for Rucio failing, you will not loose sleep on it !

belforte:
@amaltaro about this 
```python
        ### FIXME: shouldn't this LFN fail the input regex?!
        resp = self.myRucio.getPFN(site="T2_CH_CERN_Test", lfns="")
```
I believe that the explanation is that current Rucio's lfn2pfn implementation simply appies the TFC from that side as currently defined, so it really depends on the kind of regexp syntax they used, which is why different sites give mixed result (works/does not work) with same test patterns. That's whay I decided to stick with `lfn=/store/afile` as simplest possible yet that everybody accepts. I'd say that testing site TFC's is out of scope, nor am I in favor of adding a check on our side. All in all if in the future Ops manages to have all sites lfn2pfn happen via a simple prefix (the plan), the query with `lfn=""` may actually be the best way to use this ! Let me know if you agree with replacing the FIXME with some succint explnation of the above.
Then I believe we can close this until actual problems arise.

amaltaro:
Of course, you're the expert here ;) I thought Rucio had its own regex on the input files before anything else (TFC rules) were applied. It makes sense, thanks for this explanation Stefano.

belforte:
it will be a good day when we can stop using this and simply set Rucio rules !

belforte:
@amaltaro I am done here. All yours. Unit tests fail, but it looks to me to be not in this file.

amaltaro:
Thanks Stefano. Let me ask you then a final change. Can you please squash all the test commits into a single one? Such that this PR provides 2 commits; 1st for the real changes (src/*), 2nd commit for unittests/pylint changes (usually under test/*).

belforte:
ah.. sure. sorry should have thought of that myself

amaltaro (approved):
Perfect, thanks!

PR #10031, opened 2020-10-29
Ignore availability for block level rules against the origin site/RSE

### LINKED ISSUE #10030: Block level rules should be created with ignore_availability=True
**Impact of the new feature**
WMAgent

Yes, sites could become ineligible and not accept rule creation for some period of time. Meaning, the agents will keep failing to create block-level rules against those RSEs and - I guess - it could even cause that data to be transferred somewhere else while the workflow is still active in the system.

WMAgent block level rules - created against the origin RSE, where data is being produced - should be made with `ignore_availability=True` by default, such that it can bypass sites in downtime (availability_write: False, in Rucio words).

For our own education, from Eric: "In other words, rules that are meant to reflect reality about where data is already should be “True”. Rules that seek to move data should be “false”. Unless you really want to queue up data to be transferred “whenever” a site comes back."

none

**Additional context**
This has been discussed with Eric V over slack.

amaltaro:
Fixes #10030

#### Status
not-tested

#### Description
These block level rules are created to lock data that is getting produced at a given "origin" site/RSE. Which means, data is already available at the storage. So, ignore RSEs tagged as unavailable/downtime and make sure our block level rules can get created through the `ignore_availability=True`.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -289,7 +289,7 @@ class RucioInjectorPoller(BaseWorkerThread):
                 continue
             kwargs = dict(activity="Production Output", account=self.rucioAcct,
                           grouping="DATASET", comment="WMAgent automatic container rule",
-                          meta=self.metaData)
+                          ignore_availability=True, meta=self.metaData)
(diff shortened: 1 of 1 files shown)


todor-ivanov (approved):
looks good to me

amaltaro:
Thanks Todor!

PR #10034, opened 2020-10-29
List of json templates that can undergo input data placement

### LINKED ISSUE #10033: Create a subset of workflow templates that can have input placed by MSTransferor
**Impact of the new feature**
WMCore

Our integration tests became much harder now that we use Rucio and there are differences between the Int and Prod instances. So I think it would be useful if we knew exactly which test json templates can be injected with MSTransferor fully enabled (including data placement).

Create a new command line parameter (`--transfer` or something like that) which can be used to specify that we only want to inject workflows that are safe to have input data placement done. Which means, we also have to hardwire a list of template names in the `inject-test-wfs.py` script.
 
I have already identified good templates that would transfer only a few GB of data (maybe a few tens of GB).

**Additional context**
none

amaltaro:
Fixes #10033 

#### Status
not-tested

#### Description
New command line parameter `--transfer` created to signalize that we want to inject only those workflows that have a very small input data requirement, thus which can have a rule created to make their data placement.

yes

#### Related PRs
none

none

### DIFF
--- a/test/data/ReqMgr/inject-test-wfs.py
+++ b/test/data/ReqMgr/inject-test-wfs.py
@@ -91,6 +91,8 @@ def parseArgs():
                         help='ProcessingVersion for assignment')
     parser.add_argument('-i', '--injectOnly', action='store_true', default=False,
                         help='Only injects requests but do not assign them')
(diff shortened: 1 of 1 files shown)


amaltaro:
Works like a charm. FYI @todor-ivanov

PR #10036, opened 2020-10-29
Script to parse recursively cms-sw/cmsdist spec files

### LINKED ISSUE #10035: Automated check of py3 support in WMCore spec dependencies
**Impact of the new feature**

We could use `caiusepython3` to check if all the third-party python dependencies of WMCore support python3.

We need a script that parses a spec file [from cmsdist](https://github.com/cms-sw/cmsdist/tree/comp_gcc630) and produces a `requirements.txt` file that contains the python modules imported from the spec file and all its dependencies.
This `requirements.txt` could then be analyzed with `caniusepython3` to check if the WMCore packages depend on spec files that are not python3 ready.

mapellidario:
Fixes #10035

#### Status

ready

#### Description

We provide a script that parses a spec file and produces a requirements.txt files with all the python packegs in the spec file itself and all its dependencies

yes

This can be run from our laptop, so it does not require any change in the build and deploy process

### DIFF
--- a/bin/adhoc-scripts/ParseSpecCmsswdist.py
+++ b/bin/adhoc-scripts/ParseSpecCmsswdist.py
@@ -1,0 +1,295 @@
+"""Script that parses a cmssw-dist spec files and returns a requirements.txt
+
+Intro
+=====
+
+This script may have multiple purposes depending on the parameters it is given.
+
+Non-recursive
+=============
+
+-d    Path of the spec's files directory (e.g. /path/to/cmssw-dist )
+-f    Spec file filename (e.g. `wmagentpy3.spec`)
+
+This is the default behavior. The script reads the specfile, parses the
+``Requires:`` field selecting the ``py2-*.spec`` or ``py3-*.spec`` entries,
+tries to guess the name of the package and adds it to the ``requirements.txt``
+with the specified version.
+
+Recursive
+=========
+
+The purpose of this behavior is to automatically check if the **python**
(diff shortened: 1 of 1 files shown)


amaltaro (commented):
Dario, from our discussion in one of the WMCore meetings. I think it was suggested not to get dependencies recursively. Should we update this script to only consider top level python libraries and then provide your procedure to deploy those services and get a `pip freeze` with all the inner dependencies?

mapellidario:
I updated the script so that by default it is not recursive. 

I would still keep the recursive behavior, it reminded me that the current `wmagentpy3.spec` still indirectly depends on `nats-client` (from `cmsmonitoring`), `restkit` (from `couchapp`) and `python-cjson` (from `dbs3-pycurl-client.spec`).

Moreover, I added some functions to guess the package name from `%setup`, `%define pip_name` or `%define downloadn` spec files directives. Manual changes to the output requirements.txt should not be necessary anymore.

amaltaro:
> I would still keep the recursive behavior, it reminded me that the current wmagentpy3.spec still indirectly depends on nats-client (from cmsmonitoring), restkit (from couchapp) and python-cjson (from dbs3-pycurl-client.spec).

We'll cross that bridge once we get to it ;)

Thanks Dario.

amaltaro:
Actually, please take care of all the pylint/pep issues reported. New files should come as clean as possible.

amaltaro:
Transient problem with DBS. Thanks for sorting all of those Dario.

PR #10041, opened 2020-10-29
MSTransferor: resolve RSE expression for pileup location

### LINKED ISSUE #9969: Consider rules in INJECT/REPLICATING state for pileup location at MSTransferor level
**Impact of the new feature**
ReqMgr2MS - MSTransferor

To avoid over-replicating the pileup dataset - in case the campaign configuration allows it - we should consider those pileup container rules sitting in state INJECT (rule created but not yet considered by Rucio) and REPLICATING as valid location for the pileup container.

The current MSTransferor logic is here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Tools/PycurlRucio.py#L137

and we should add those RSEs from rules in state INJECT or REPLICATING (any other state to be considered?), of course we should log such cases as well.

Besides accepting those as data location, we would have to keep their rule id such that MSMonitor can keep evaluating it and only release the workflow once that rule has been satisfied. Passing this rule id from `listReplicationRules` function all the way down to where transferIDs are persisted is going to be a challenge...

If we could implement it assuming MSTransferor will only run in Rucio mode, it would be much simpler...

**Additional context**
none

### LINKED ISSUE #10038: Pileup location RSE expressions are not evaluated in MSTransferor
**Impact of the bug**
ReqMgr2MS - MSTransferor

**Describe the bug**
When checking the pileup location through the PycurlRucio module, we use the RSE expression - from the rule data - as is, trusting that only MSTransferor would be making such rules, thus always single RSE rule with the correct RSE name.
This is very dangerous and we should always resolve the RSE expression in a list of RSEs.

**How to reproduce it**
Use the Pycurl `listReplicationRules` function.

**Expected behavior**
We need to resolve the RSE expression value, taken from the rule information via parameter `rse_expression`, such that we have the correct RSE name for the pileup location.
There are many ways to resolve it, like:
* sending a GET request to the `rses` RESTful API as soon as we have the RSE expression
* collect all the RSE expressions and make another concurrent call via a new Pycurl based API
* resolve the RSEs using the WMCore wrapper module, which has the ability to cache RSE expressions

I'm leaning towards the last option, which would be more maintainable and cleaner. The performance might not be optimal, but I do not think it will have a large contribution on the overall service execution time. 

none

amaltaro:
Fixes #10038 
Fixes #9969 (not really a fix though, more like a minor enhancement)

#### Status
ready

#### Description
* Improved logging for pileup rules evaluated by MSTransferor;
* STUCK rules with NO_SOURCES are never used as a valid location;
* Resolve the pileup RSE expressions in MSTransferor by using the MSTransferor Rucio object, which has an in-memory cache for 12h.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Tools/PycurlRucio.py
+++ b/src/python/WMCore/MicroService/Tools/PycurlRucio.py
@@ -147,7 +147,8 @@ def listReplicationRules(containers, rucioAccount, grouping,
     :param rucioUrl: string with the Rucio url
     :param rucioToken: string with the Rucio token
     :param scope: string with the data scope
-    :return: a flat dictionary key'ed by the container name, with a list of RSEs as value
+    :return: a flat dictionary key'ed by the container name, with a list of RSE
+      expressions that still need to be resolved
     NOTE: Value `None` is returned in case the data-service failed to serve a given request.
     """
     locationByContainer = {}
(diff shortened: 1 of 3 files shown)


amaltaro:
Works fine too!

PR #10042, opened 2020-10-30
Fix pileup data location in global workqueue

### LINKED ISSUE #10040: Global WorkQueue using the wrong 'grouping' for pileup data location
**Impact of the bug**
Global WorkQueue

**Describe the bug**
While testing workflows in my VM, I noticed that global workqueue was creating elements which had an empty list of RSEs for the pileup data, even though the container rule had recently being satisfied.
Upon global workqueue restart, the LocationUpdateTask cherrypy thread gets executed and those elements get updated with the correct pileup location.

Both threads use the same method `getDataLockedAndAvailable`, but they provide a different set of arguments. For the workqueue, here is where the pileup gets resolved:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/Policy/Start/StartPolicyInterface.py#L255

while the LocationUpdateTask thread calls this:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/DataLocationMapper.py#L99

**How to reproduce it**
inject any workflow that requires pileup data.

**Expected behavior**
Pileup data location has to work during the workqueue element creation. Which means, the `grouping="DATASET"` key/value pair has to be removed (this internally uses the containe ruler RSEs for rules in the OK state).

There are no error messages, just the wrong information in the GQ element.

amaltaro:
Fixes #10040 

#### Status
ready

#### Description
MSTransferor does not make pileup container level rules with grouping=DATASET, thus when a workqueue element is getting created, pileups always have an empty location until the LocationUpdateTask thread kicks in and update those elements (that thread does not filter by grouping type!).

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WorkQueue/Policy/Start/StartPolicyInterface.py
+++ b/src/python/WMCore/WorkQueue/Policy/Start/StartPolicyInterface.py
@@ -252,7 +252,7 @@ class StartPolicyInterface(PolicyInterface):
         for dbsUrl in datasets:
             for datasetPath in datasets[dbsUrl]:
                 if hasattr(self, "rucio"):
-                    locations = self.rucio.getDataLockedAndAvailable(name=datasetPath, grouping="DATASET",
+                    locations = self.rucio.getDataLockedAndAvailable(name=datasetPath,
                                                                      account=self.args['rucioAcct'])
                 else:
                     locations = set()


amaltaro:
Works fine!

PR #10044, opened 2020-11-02
Purge data when deleting block-level rules. Fix naming of tape RSEs t…

germanfgv:
Purge data when deleting block-level rules. Fix naming of tape RSEs taken from database

#### Status
ready

#### Description
Tier0 needs data to be deleted as soon as possible. Because of this, the purge_replicas option is now used when deleting block-level rules. 

On the other hand,I added a naming fix, in case database uses `_MSS` as suffix for tape sites. The suffix will be replaced by `_Tape`.

YES 
#### Related PRs
This fixes an issue #9959

This does not require deployment changes


amaltaro:
@germanfgv I guess you need it to be backported to the wmagent branch too, right?

amaltaro (approved):

germanfgv:
> @germanfgv I guess you need it to be backported to the wmagent branch too, right?

This would be nice, but is not necessary. We can patch our agents.

PR #10051, opened 2020-11-05
Use dataset_locks to find pileup block locations in WMAgent

amaltaro:
Hack PR to allow P&R team to run some I/O tests with only a sub-set of the pileup blocks locally available.
So, this changes the current behaviour which is based only on the container level rules, to also consider all the dataset_locks for every single block in the pileup container.

This patch is getting applied to the `backfill` agent (vocms0285).


amaltaro:
This hack has been ported to the latest production branch: 1.4.3

PR #10025, opened 2020-10-27
Bundle of patches for WMAgent 1.4.1.patch4

amaltaro:
wmagent branch version of a bunch of patches:
* Use list_replication_rules Rucio API to list rules #9963 
* If listDataRules API fail, raise a WMRucioException #9999 
* Delete blocks using container-block mapping #9959 
* Deal with T0 output data subscription in RucioInjector #10006 
* Only add replicas if parent block has been properly added/attached into Rucio #10024

PR #10067, opened 2020-11-16
Removing file status condition from GetUnsubscribedDatasets query

### LINKED ISSUE #10064: RucioInjector  checking unnecessary conditions to create Container Rules
**Impact of the bug**
WMAgent

**Describe the bug**
It's not really a bug, but an outdated behavior. RucioInjectorPoller will create rules only for containers whose blocks have been uploaded to DBS. Apparently, this behavior was needed with PhEDExInjector, but makes no sense for Rucio as it causes an unnecessary delays in transfers.

**How to reproduce it**
Create a new container whose blocks are not uploaded to DBS, and define a subscription for it on the database. The subscription/rule will not be created as the status of files within the container will be `bsbuffer_file.status = 'NOTUPLOAD'`

**Expected behavior**
RucioInjector should create container rules as soon as the container subscription/rule is properly defined in the database. i.e:
`dbsbuffer_dataset_subscription.subscribed = 0`
`dbsbuffer_dataset.path != 'bogus'`

**Additional context**
None

germanfgv:
Fixes #10064 >

#### Status
Ready

#### Description
As discussed in #10064, I removed the requirement for files to be uploaded to DBS from GetUnsubscribedDatasets query . This will allow to create container rules earlier. 

No, It is not compatible with systems that use PhEDEx injector (WMCore version < 1.4.1)

#### Related PRs
#10025

### DIFF
--- a/src/python/WMComponent/RucioInjector/Database/MySQL/GetUnsubscribedDatasets.py
+++ b/src/python/WMComponent/RucioInjector/Database/MySQL/GetUnsubscribedDatasets.py
@@ -32,7 +32,6 @@ class GetUnsubscribedDatasets(DBFormatter):
                INNER JOIN dbsbuffer_file ON
                  dbsbuffer_algo_dataset_assoc.id = dbsbuffer_file.dataset_algo
              WHERE dbsbuffer_dataset_subscription.subscribed = 0 AND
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

amaltaro (approved):
Changes look good to me, German. Thanks!
PS.: The unit test failure is unrelated and it has been fixed in another PR that is getting merged today/tomorrow.

PR #10072, opened 2020-11-17
Evaluate all WQEs when local resources are available

### LINKED ISSUE #10066: Local WorkQueue not exhausting local resources when pulling work from GQ
**Impact of the bug**
WMAgent

**Describe the bug**
With the change provided by this PR:
https://github.com/dmwm/WMCore/pull/9679

we started limiting the amount of workqueue elements that are retrieved by the agents; which is very need to have a performant system and avoid wasting many network/cpu cycles.

It turns out that sometimes all those 200 elements (by default) retrieved from Couchdb, do not pass the site and job resource restrictions, performed in the `availableWork` method:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WorkQueueBackend.py#L324

I'm still to fully understand what happens, but I think the GQEs pulled from couch have a lower priority than what we have already in the agent, which makes that/those GQEs to be skipped in the cycle (over and over).

**How to reproduce it**
not simple to reproduce with a small amount of work

**Expected behavior**
There are a few options to address this, such as:
1) ensure that whatever is pulled from CouchDB, is fully replicated and accepted by the agent (likely a change in the workRestrictions.js javascript code)
2) pull more than X (200 by default) elements from CouchDB, but only accept the first X elements to be worked on
3) keep pulling elements from CouchDB (using the offset option, not sure it works with descending though...), until X elements have been accepted by the agent.

For 1, we would have to overcommit elements at the javascript layer (knowing the local resources); for 2, we would need to find the best and reliable number of elements to be pulled down.
 
Example of WorkQueueManager log from vocms0251:
```
2020-11-16 10:54:39,972:140297023502080:INFO:WorkQueueBackend:Retrieved 200 elements from workRestrictions list for: https://cmsweb.cern.ch/couchdb/workqueue
2020-11-16 10:54:39,979:140297023502080:INFO:WorkQueueBackend:Current siteJobCounts:
2020-11-16 10:54:39,979:140297023502080:INFO:WorkQueueBackend:    T1_RU_JINR : {80000: 1000}
2020-11-16 10:54:39,979:140297023502080:INFO:WorkQueueBackend:    T2_US_MIT : {85000: 1000}
2020-11-16 10:54:39,989:140297023502080:INFO:WorkQueueBackend:And 0 elements passed location and siteJobCounts restrictions for: https://cmsweb.cern.ch/couchdb/workqueue
```

even though there are plenty of resources available, according to the local db.

amaltaro:
Fixes #10066 

#### Status
tested

#### Description
In very short, this fixes an issue where a low priority workflow doesn't manage to run, even though it's assigned to resources that have free resources in the agents.

In more details, this PR provides:
* clear distinction between fetching LQE to be processed from fetching LQE to calculate the resources available/allocated (freeResouceCheck + calculateAvailableWork)
* `calculateAvailableWork()` fetches all the local workqueue elements in a single call. Removed most of its logging and unneeded complexity
* `availableWork()` fetches global workqueue elements in slices of 1k elements, until all the elements have been exhausted, or until the component has accepted the configured number of elements (default=200). Example, if we have 2500 elements in Available status, we would:
  *) make first HTTP request and retrieve elements in the range of [0, 1000]
  *) then make another HTTP request and retrieve elements in the range of [1000, 2000]
  *) then make a third HTTP request and retrieve elements in the range of [2000, 3000] (actually [2000, 2500]
  *) and a final HTTP request is made, to retrieve elements in the range of [3000, 4000], which would return an empty list and exit the loop.

In addition to that:
* stop returning the threshold from `availableWork`, it is anyways unchanged between the method call and response
* removed unused `wfs` parameter from `availableWork` method
* log when an element is being accepted, and against which "site" it was (even though it can spawns jobs against multiple sites)
* rest of the logic is untouched, higher priority elements that have been created first, go first!

YES (but could be made incompatible with the javascript code...)

#### Related PRs
none

For reference, CouchDB documentation for couch views:
https://docs.couchdb.org/en/1.6.1/api/ddoc/views.html?highlight=views%20stale

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueueBackend.py
+++ b/src/python/WMCore/WorkQueue/WorkQueueBackend.py
@@ -321,95 +321,55 @@ class WorkQueueBackend(object):
             except CouchNotFoundError:
                 pass
 
-    def availableWork(self, thresholds, siteJobCounts, team=None, wfs=None,
-                      excludeWorkflows=None, numElems=9999999):
+    def calculateAvailableWork(self, thresholds, siteJobCounts):
         """
-        Get work which is available to be run
-
-        Assume thresholds is a dictionary; keys are the site name, values are
-        the maximum number of running jobs at that site.
-
-        Assumes site_job_counts is a dictionary-of-dictionaries; keys are the site
-        name and task priorities.  The value is the number of jobs running at that
-        priority.
-
-        It will pull work until it reaches the number of elements configured (numElems).
-        Since it's also used for calculating free resources, default it to "infinity"
-
-        Note: this method will be called with no limit of work elements when it's simply
-        calculating the resources available (based on what is in LQ), before it gets work
-        from GQ
+        A short version of the `availableWork` method, which is used only to calculate
+        the amount of work already available at the local workqueue.
+        :param thresholds: a dictionary key'ed by the site name, values representing the
+            maximum number of jobs allowed at that site.
+        :param siteJobCounts: a dictionary-of-dictionaries key'ed by the site name; value
+            is a dictionary with the number of jobs running at a given priority.
+        :return: a tuple with the elements accepted and an overview of job counts per site
         """
-        self.logger.info("Getting up to %d available work from %s", numElems, self.queueUrl)
-        self.logger.info("  for team name: %s", team)
-        self.logger.info("  for wfs: %s", wfs)
-        self.logger.info("  with excludeWorkflows: %s", excludeWorkflows)
-        self.logger.info("  for thresholds: %s", thresholds)
-
-        excludeWorkflows = excludeWorkflows or []
+        # NOTE: this method can be less verbose as well
         elements = []
-        sortedElements = []
-
-        # We used to pre-filter sites, looking to see if there are idle job slots
-        # We don't do this anymore, as we may over-allocate
-        # jobs to sites if the new jobs have a higher priority.
-
         # If there are no sites, punt early.
         if not thresholds:
             self.logger.error("No thresholds is set: Please check")
-            return elements, thresholds, siteJobCounts
+            return elements, siteJobCounts
+
+        self.logger.info("Calculating available work from queue %s", self.queueUrl)
 
         options = {}
         options['include_docs'] = True
         options['descending'] = True
-        options['num_elem'] = numElems
         options['resources'] = thresholds
-        if team:
-            options['team'] = team
-        if wfs:
-            result = []
-            for i in xrange(0, len(wfs), 20):
-                options['wfs'] = wfs[i:i + 20]
-                data = self.db.loadList('WorkQueue', 'workRestrictions', 'availableByPriority', options)
-                result.extend(json.loads(data))
-        else:
-            result = self.db.loadList('WorkQueue', 'workRestrictions', 'availableByPriority', options)
-            result = json.loads(result)
-            if not result:
-                self.logger.info("No available work or it did not pass work/data restrictions for: %s ",
-                                 self.queueUrl)
-            else:
-                self.logger.info("Retrieved %d elements from workRestrictions list for: %s",
-                                 len(result), self.queueUrl)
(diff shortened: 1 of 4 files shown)


amaltaro:
test this please

amaltaro:
This has been tested in testbed and it seems to be working just fine. I guess I'm going to apply this patch to one production agent and keep watching it for 30min or so, before pushing it in in the other agents.

amaltaro:
FYI., I found other confusing code around (WorkQueueManager runs in multi-threading against multiple queues, making things hard to debug) and I'm going to make another commit today. In any case, feel free to look into this one.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
```diff
@@ -321,112 +321,127 @@ def deleteElements(self, *elements):
...
+                    sortedElements.append(element)
+            sortedElements.sort(key=lambda element: element['CreationTime'])
+            sortedElements.sort(key=lambda element: element['Priority'], reverse=True)
+
+            for element in sortedElements:
+                if numElems <= 0:
+                    msg = "Reached maximum number of elements to be accepted, "
+                    msg += "configured to: {}, from queue: {}".format(len(elements), self.queueUrl)
+                    self.logger.info(msg)
+                    breakOut = True  # get out of the outer loop as well
```
todor-ivanov, line 487: One question here:  Do you account for or is there some other kind of protection for the fact that if you have exhausted the "Maximum number of elements  to accept" counter while you are in the middle of the current slice of elements fetched from the Queue, then on your next run you are about to skip the rest of the elements that are in the current slice and not yet considered at all. I am not sure this would ever be a problem, though.

todor-ivanov (approved):
All the changes seem to be fine. There is only one question inline. If you want you may elaborate on it in the comment directly. Other than that I do not see any other obstacles here.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
amaltaro: Yes, we break out of this loop right away and the first thing to be checked in the `while` loop is whether this flag `breakOut` is True or not. If true, then the whole loop going through the elements is over.

amaltaro:
Right, time to work on the best things in WMCore, WorkQueue unit tests...

amaltaro:
@todor-ivanov the first commit has already been reviewed by you. I decided to make this code a bit more clear with what is pulled from what and for which purpose. Those changes have gone in in the 2nd commit.

amaltaro:
Todor, feel free to leave your review in here too. I'm merging it to hand it over to Imran.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
```diff
@@ -321,112 +321,197 @@ def deleteElements(self, *elements):
...
-        for i in result:
-            element = CouchWorkQueueElement.fromDocument(self.db, i)
-            # make sure not to acquire work for aborted or force-completed workflows
-            if element['RequestName'] in excludeWorkflows:
-                msg = "Skipping aborted/force-completed workflow: %s, work id: %s"
-                self.logger.info(msg, element['RequestName'], element._id)
-            else:
-                sortedElements.append(element)
-        # sort elements to get them in priority first and timestamp order
+        options['num_elem'] = 9999999  # magic number!
```
todor-ivanov, line 347: What is the magic of this number?

todor-ivanov (commented):
@amaltaro I had only one question inline. And the other comment was going to make was about the trade off between code and call repetitions, but this one we already discussed in a private chat few days ago.  
The changes look good to me.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
amaltaro: just a number that we should never reach in the coming 10 (or maybe 100) years or so ;) such that all the elements from workqueue are returned in the same call.

PR #10080, opened 2020-11-18
Add zeroCounters() function.

### LINKED ISSUE #10078: Cleaned and Archived counters not zeroed in MSRuleCleaner
**Impact of the bug**
MSRuleCleaner

**Describe the bug**
Our initial idea was  to initiate the workflow counters for 'Cleaned' and 'Archived' workflows only once during the thread creation and then accumulate there while the service is running. But it does not give the correct information, because of two reasons: 
1. We are still just cleaning and not archiving the workflows with the MSRuleCleaner, that's why every each one of them sits in the queue for at least 3 days and will be fetched by the service on every run during all those days, ergo will reenter the counters multiple times.
2. We do not need the total amount of workflows treated since the initial MSRuleCleaner thread start, but we rather need only the numbers from the last pulling cycle.

**How to reproduce it**
No need to - it is a request for a change of the current behavior

**Expected behavior**
Accumulate the numbers only from the last polling cycle into the Cleanup and Archived counters. 

N/A

todor-ivanov:
Fixes #10078 

#### Status
ready

#### Description
The current change introduces  a simple function which should be called from the execute method so that the counters can be zeroed on every pulling cycle. 

YES

#### Related PRs
NO

NO

### DIFF
--- a/src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
+++ b/src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
@@ -108,8 +108,14 @@ class MSRuleCleaner(MSCore):
         # Initialization of the 'cleaned' and 'archived' counters:
         self.wfCounters = {'cleaned': {},
                            'archived': 0}
+
+    def resetCounters(self):
+        """
+        A simple function for zeroing the cleaned and archived counters.
+        """
         for pline in self.cleanuplines:
             self.wfCounters['cleaned'][pline.name] = 0
+        self.wfCounters['archived'] = 0
 
     def execute(self, reqStatus):
         """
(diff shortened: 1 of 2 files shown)


src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -108,8 +108,14 @@ def __init__(self, msConfig, logger=None):
         # Initialization of the 'cleaned' and 'archived' counters:
         self.wfCounters = {'cleaned': {},
                            'archived': 0}
+
+    def zeroCounters(self):
```
amaltaro, line 112: Given that you have to sort that unit test out, let me say that `resetCounters` is a more natural name ;)

amaltaro (changes requested):
One comment in the code and the unit test to be looked at. Thanks

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
And also unit test fixed

PR #10100, opened 2020-11-24
Split properly archived and force archived counters.

### LINKED ISSUE #10082: Split ProperlyArchived and ForceArchived counters in MSRuleCleaner
**Impact of the bug**
MSRuleCleaner

**Describe the bug**
For some of the workflows in the queue (so far only ACDCs) we are setting the`ForceArchive` flag to True at  the very begining so that they bypass the whole rest logic of the service and go straight to archival. The reason for that is because we consider the ACDC would not have its won set of Rucio rules but would rather benefit from the already created ones for the main request submission. But we do not distinguished between which flag triggered the archival: either `IsClean` or `ForceArchive`. The result is that the those workflows are contributing to the 'properly archived' counters of the service which accumulate bigger number than the 'properly cleaned' counters, which makes the report from the service be ambiguous  and missleading[1].

[1]  
```
2020-11-19 01:48:31,909:INFO:MSRuleCleaner: Workflows cleaned by pipeline: plineMSTrCont: 197
2020-11-19 01:48:31,910:INFO:MSRuleCleaner: Workflows cleaned by pipeline: plineMSTrBlock: 197
2020-11-19 01:48:31,910:INFO:MSRuleCleaner: Workflows cleaned by pipeline: plineAgentCont: 299
2020-11-19 01:48:31,910:INFO:MSRuleCleaner: Workflows cleaned by pipeline: plineAgentBlock: 299
2020-11-19 01:48:31,910:INFO:MSRuleCleaner: Workflows archived: 380
2020-11-19 01:48:31,910:INFO:MSRuleCleaner: 
Number of processed workflows: 380.
Number of properly cleaned workflows: 299.
Number of archived workflows: 380.
```

**How to reproduce it**

**Expected behavior**
To account for which flag triggered the archival in two different counters and include them into the report.

N/A

todor-ivanov:
Fixes #10082 

#### Status
ready

#### Description
Splits the archived counter in two separate counters:
* One counting the number of workflows that are having the `Forcearchived` flag set to `True` (such as Resubmissions) - those workflows basically bypass the whole logic of MSRuleCleaner and are sent directly for archival
* The other is counting the workflows that have actually passed through the cleaning pipelines and have reached the archival step (and have been eventually properly archived)

YES

#### Related PRs
No

No


amaltaro:
Todor, I haven't reviewed the whole PR because there is still a unit test failing. However, why don't we use a flat structure for `self.wfCounters`, something like:
```
self.wfCounters = {'cleaned': {},
                              'properArchived': 0,
                              'forceArchived': 0}}
```
?

And about naming, I'd suggest `normalArchived` instead of `properArchived`.

todor-ivanov:
Hi @amaltaro,
About the naming convention - I agree, your suggestion seems to be better than what I came up with. 
About the data structure I believe having them separated as `cleaned` and `archived` counters makes things look a little bit better organized, because in they are related to two consequent and well defined steps in the code. 

The change about the renaming is on its way. I am about to fix the Unittest   in a separate commit - it is because of the change of the names of those exact counters. Please review the following change I am about to make so I can push the fix for the unit test at the end of the PR.

amaltaro:
>  About the data structure I believe having them separated as cleaned and archived counters makes things look a little bit better organized, because they are related to two consequent and well defined steps in the code.

the key name already contains `Archived` in it, so it looks clear enough to me. Having a nested dictionary to be used as a counter seems unneeded and error prone to me as well. Anyhow, whatever you prefer.

src/python/WMCore/MicroService/DataStructs/DefaultStructs.py
```diff
@@ -68,7 +68,8 @@
                           error="",
                           total_num_requests=0,
                           clean_num_requests=0,
-                          archived_num_requests=0,
+                          proper_archived_num_requests=0,
```
amaltaro, line 71: should we rename it here as well to `normal_archived...`?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -144,17 +146,20 @@ def execute(self, reqStatus):
 
         # Call _execute() and feed the relevant pipeline with the objects popped from requestRecords
         try:
-            totalNumRequests, cleanNumRequests, archivedNumRequests = self._execute(requestRecords)
+            totalNumRequests, cleanNumRequests, normalArchivedNumRequests, forceArchivedNumRequests = self._execute(requestRecords)
```
amaltaro, line 149: I wonder why this method returns 
```
cleanNumRequests, normalArchivedNumRequests, forceArchivedNumRequests
```

if there is already an object attribute counter for that? Shouldn't it simply return `totalNumRequests` and the rest of the information is available in `self.wfCounters`?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -144,17 +146,20 @@ def execute(self, reqStatus):
...
             msg = "\nNumber of processed workflows: %s."
             msg += "\nNumber of properly cleaned workflows: %s."
-            msg += "\nNumber of archived workflows: %s."
+            msg += "\nNumber of properly archived workflows: %s."
```
amaltaro, line 152: Maybe here too? Something like "Number of normally archived..."

amaltaro (changes requested):
Todor, I'd suggest to update the default structure as well to keep things consistent.
Please have a look at the comments made in the code.

src/python/WMCore/MicroService/DataStructs/DefaultStructs.py
todor-ivanov: Agreed

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: At the beginning I did it for consistency of how this was done before and then I found it quite useful in the unit tests to have all what I need returned in a tuple here. Because there we do not call the object from `MSManager`, but we rather create an instance for `MSRulecleaner` and we deal with it directly. Not saying it is impossible for the unit test to be fixed too, in case we change the returned value here though, but it will for sure require more than a change of a single number in the tuple, as I did for fixing it this time.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Ok

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -195,9 +200,11 @@ def _execute(self, reqRecords):
         for pline in self.cleanuplines:
             msg = "Workflows cleaned by pipeline: %s: %d"
             self.logger.info(msg, pline.name, self.wfCounters['cleaned'][pline.name])
-        archivedNumRequests = self.wfCounters['archived']
-        self.logger.info("Workflows archived: %d", self.wfCounters['archived'])
-        return totalNumRequests, cleanNumRequests, archivedNumRequests
+        normalArchivedNumRequests = self.wfCounters['archived']['normalArchived']
+        forceArchivedNumRequests = self.wfCounters['archived']['forceArchived']
+        self.logger.info("Workflows properly archived: %d", self.wfCounters['archived']['normalArchived'])
```
amaltaro, line 205: You've missed this line, but it's not a big deal.

amaltaro:
@todor-ivanov can you please squash those commits? If it's ready to go, please remove the do not merge label as well.

amaltaro:
As reminded over slack, next time keep unit tests in its own commit please.

PR #10105, opened 2020-11-24
Manage rule approval requirements for tape sites

### LINKED ISSUE #10104: Manage approval requirements while creating container rules
**Impact of the new feature**
WMAgent
Some Rucio tape sites require rules to undergo the approval process before starting data transfers. Right now, WMAgent autoapproves every container rule it creates and this causes problems for site admins.

WMAgent should identify which sites require approval, and created the rules with the necessary flags.

Adding --ask-approval flag to every Tape rule was discarted as an option because it will likely cause delays and extra work for sites without an automated approval process.

germanfgv:
Fixes #10104 

#### Status
ready

#### Description
Adds requireApproval method to Rucio wrapper. This method returns whether a site needs rule approval according to site attributes. RucioInjector uses this new method during the container rule creation process to add the --ask-approval flag to according to site needs.

YES

#### Related PRs
Builds upon #10006

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -539,6 +539,16 @@ class RucioInjectorPoller(BaseWorkerThread):
             elif self.testRSEs:
                 rseName = "%s_Test" % rseName
 
+            #Checking whether we need to ask for rule approval
+            try:
+                if self.rucio.requiresApproval(rseName):
+                    ruleKwargs['ask_approval'] = True
+            except WMRucioException as exc:
+                msg = str(exc)
+                msg += "\nUnable to check approval requirements. Will retry again in the next cycle."
+                logging.error(msg)
+                continue
+
             logging.info("Creating container rule for %s against RSE %s", container, rseName)
             logging.debug("Container rule will be created with keyword args: %s", ruleKwargs)
             try:
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -13,7 +13,7 @@ from copy import deepcopy
 from rucio.client import Client
 from rucio.common.exception import (AccountNotFound, DataIdentifierNotFound, AccessDenied, DuplicateRule,
                                     DataIdentifierAlreadyExists, DuplicateContent, InvalidRSEExpression,
-                                    UnsupportedOperation, FileAlreadyExists, RuleNotFound)
+                                    UnsupportedOperation, FileAlreadyExists, RuleNotFound, RSENotFound)
 from Utils.MemoryCache import MemoryCache
 from WMCore.WMException import WMException
 
@@ -788,6 +788,22 @@ class Rucio(object):
 
         choice = weightedChoice(rsesWithWeights)
         return choice
+
+    def requiresApproval(self, rse):
+        """
+        _requiresApproval_
+
+        Returns wether or not the specified RSE requires rule approval.
+        :param res: string containing the Rucio RSE name
+        :returns: True if the RSE requires approval to write (rule property)
+        """
+        try:
+            attrs = self.cli.list_rse_attributes(rse)
+        except RSENotFound as exc:
+            msg = "Error retrieving attributes for RSE: {}. Error: {}".format(rse, str(exc))
+            raise WMRucioException(msg)
+        return attrs.get('requires_approval', False)
+
 
     def isContainer(self, didName, scope='cms'):
         """
--- a/test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py
+++ b/test/python/WMCore_t/Services_t/Rucio_t/Rucio_t.py
@@ -439,6 +439,15 @@ class RucioTest(EmulatedUnitTestCase):
         self.assertTrue(len(resp) == 2)
         self.assertTrue(resp[1] is True or resp[1] is False)
 
+    def testRequiresApproval(self):
+        """
+        Test the `pickRSE` method
+        """
+        resp = self.myRucio.requiresApproval("T1_UK_RAL_Tape_Test")
+        self.assertTrue(resp)
+        resp = self.myRucio.requiresApproval("T1_FR_CCIN2P3_Tape_Test")
+        self.assertFalse(resp)
+
     def testIsTapeRSE(self):
         """
         Test the `isTapeRSE` utilitarian function


germanfgv:
@amaltaro Is there a problem with automatic testing?

amaltaro:
Those RSEs don't exist in the Rucio Integration server, yet. You need to append _Test to them...
Review is coming soon.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -789,6 +796,26 @@ def pickRSE(self, rseExpression='rse_type=TAPE\cms_type=test', rseAttribute='ddm
...
 
+    def requiresApproval(self, rse):
+        """
+        _requiresApproval_
+
+        Returns wether or not the specified RSE requires rule approval.
+        :param res: Rucio RSE name
+        :returns: True if the RSE requires approval to write (rule property)
+        """
+
```
amaltaro, line 807: German, let me suggest to remove this blank line. According to the PEP convention, there shouldn't be blank lines after the docstring (and I agree that its absence makes the code more readable). If you want, see:
https://www.python.org/dev/peps/pep-0257/

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -789,6 +796,26 @@ def pickRSE(self, rseExpression='rse_type=TAPE\cms_type=test', rseAttribute='ddm
...
+        _requiresApproval_
+
+        Returns wether or not the specified RSE requires rule approval.
+        :param res: Rucio RSE name
+        :returns: True if the RSE requires approval to write (rule property)
+        """
+        
+        try:
+            attrs = self.cli.list_rse_attributes(rse)
+        except RSENotFound as exc:
```
amaltaro, line 810: How about we don't throw this exception, instead just make a log error record, and return a sensible default value as you do in the line below?

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -789,6 +796,26 @@ def pickRSE(self, rseExpression='rse_type=TAPE\cms_type=test', rseAttribute='ddm
...
+            msg = "RSE not found in Rucio: {}. Error: {}".format(rse, str(exc))
+            raise WMRucioRSENotFound(msg)
+
+        requiresApproval = attrs.get('requires_approval', False)
```
amaltaro, line 814: I'm inclined to think that having a default value to `True` would be safer, no?

amaltaro (changes requested):
If you agree with not throwing that exception, then the new exception can be removed as well.

src/python/WMCore/Services/Rucio/Rucio.py
germanfgv: The thing is, sites that don't require approval don't have a `requires_approval` attribute. At least I have never seen 'requires_approval=False'. It is either 'requires_approval=True' or not 'requires_approval' at all.

amaltaro: Gotcha! Then default to `False` makes sense.

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: And changing my mind here, perhaps we can through the generic WMRucioException with the proper error message, instead of creating a new one.

germanfgv: For this we may use `True` as default. I was not sure of what should be the response to trying to create a rule to an un-existent site. I'll make the changes as you suggest.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -538,6 +538,16 @@ def insertContainerRules(self):
                     rseName = rseName.replace("cms_type=real", "cms_type=test")
             elif self.testRSEs:
                 rseName = "%s_Test" % rseName
+            
+            #Checking whether we need to ask for rule approval
+            try:
```
germanfgv, line 543: @amaltaro according to discussion over `requiresApproval` method, I'm adding this try clause here. I guess I could instead call the method inside the previously existing try clause (ln 554) but I like it better this way. What do you think?

amaltaro: This try/except here looks good. Most of the error message is already encoded in the exception though, so the error message here can likely be shorter and append str(exc) etc.

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -789,6 +789,24 @@ def pickRSE(self, rseExpression='rse_type=TAPE\cms_type=test', rseAttribute='ddm
         choice = weightedChoice(rsesWithWeights)
         return choice
 
+    def requiresApproval(self, rse):
+        """
+        _requiresApproval_
+
+        Returns wether or not the specified RSE requires rule approval.
+        :param res: Rucio RSE name
```
amaltaro, line 797: can you mention the data type here as well? E.g.: "string with the Rucio RSE name"

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -789,6 +789,24 @@ def pickRSE(self, rseExpression='rse_type=TAPE\cms_type=test', rseAttribute='ddm
...
+        :returns: True if the RSE requires approval to write (rule property)
+        """
+        try:
+            attrs = self.cli.list_rse_attributes(rse)
+        except RSENotFound as exc:
+            msg = "Error retrieving attributs for RSE: {}. Error: {}".format(rse, str(exc))
+            raise WMRucioException(msg)
```
amaltaro, line 804: I'd replace the else: ... by something like:
```
return attrs.get('requires_approval', False)
```

if we got to this point, then there was no exception.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I shortened the message

germanfgv:
@amaltaro I think I have addressed your concerns. Also, I tested this with Tier0 agent and seems ti be working as expected. I think this is ready to be merged. Do you (or @todor-ivanov ) have additional comments?

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -789,6 +789,22 @@ def pickRSE(self, rseExpression='rse_type=TAPE\cms_type=test', rseAttribute='ddm
...
+        _requiresApproval_
+
+        Returns wether or not the specified RSE requires rule approval.
+        :param res: string containing the Rucio RSE name
+        :returns: True if the RSE requires approval to write (rule property)
+        """
+        try:
+            attrs = self.cli.list_rse_attributes(rse)
+        except RSENotFound as exc:
+            msg = "Error retrieving attributs for RSE: {}. Error: {}".format(rse, str(exc))
```
amaltaro, line 803: Can you please fix this typo "attributs"?

amaltaro (changes requested):
German, changes look good to me. Can you just fix that typo pointed out in the code?

src/python/WMCore/Services/Rucio/Rucio.py
germanfgv: Done!

amaltaro (approved):

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: Thanks!

PR #10109, opened 2020-11-26
Add MSRuleCleanerStandalone.py

### LINKED ISSUE #10097: Delete old Block and Container Level rules at Rucio
**Impact of the new feature**
Rucio migration

While being in the transition period during the Rucio migration we have accumulated some amount of block and container level rules under the `wma_prod` and eventually the `wmcore_transferror` accounts. The current issue is more likely related only to some operational procedures needed to be followed in order to clean those, rather than to any kind of development. 

One option is to use the already existing `MSRuleCleaner` service, change few parameters like the workflow status we consider in the service, and then we use the whole logic there for the cleaning.   

N/A

**Additional context**
N/A

todor-ivanov:
Fixes #10097 

#### Status
 ready

#### Description
Adds a new script which is supposed to be used for calling the MSRuleCleaner as a standalone executable in order to be used for cleaning all the old Rucio rules created by the agents during the transition period from Phedex to Rucio. We have  to fetch all the workflows in statuses `['abordet-archived', 'rejected-archived', 'normal-archived' ]`, and the list of workflows to work with needs to be build from the CouchDB views instead of reqmgr. We also need to add the equivalent `*-archived` statuses at the relevant places, so that the already existing logic in MSRulecleaner may cover the workflows which are already in archived state.  

YES

#### Related PRs
NO

NO

### DIFF
--- a/bin/adhoc-scripts/MSRuleCleanerStandalone.py
+++ b/bin/adhoc-scripts/MSRuleCleanerStandalone.py
@@ -89,7 +89,7 @@ argParser.add_argument('-st', '--reqStatus',
                        default=[])
 
 argParser.add_argument('-erm', '--enableRealMode',
-                       type=bool,
+                       action='store_true',
                        required=False,
                        help="Enable RealRun mode",
                        dest='enableRealMode',
--- a/bin/adhoc-scripts/MSRuleCleanerStandalone.py
+++ b/bin/adhoc-scripts/MSRuleCleanerStandalone.py
@@ -1,7 +1,10 @@
+#!/usr/bin/python
+
 import sys
 import os
 import time
 import logging
+import argparse
 
 from pprint import pformat
 from itertools import izip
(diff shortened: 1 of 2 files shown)


bin/adhoc-scripts/MSRuleCleanerStandalone.py
```diff
@@ -0,0 +1,77 @@
...
+
+# https://cmsweb-testbed.cern.ch/couchdb/reqmgr_workload_cache/_design/ReqMgr/_view/bystatusandtime?descending=false&startkey=%5B%22aborted-archived%22,1588338001%5D&endkey=%5B%22aborted-archived%22,1606224631%5D
+
+couchDB = CouchDB(dbname='reqmgr_workload_cache',
+                  url=msConfig['couchDBUrl'],
+                  size=100000,
+                  ckey=os.getenv("X509_USER_KEY", "Unknown"),
+                  cert=os.getenv("X509_USER_CERT", "Unknown"))
+timeFormat = "%b %d %H:%M:%S %Y"
+startTimeStr = "May 1 15:00:00 2020"
```
amaltaro, line 42: Nick provided the timestamp for the first `wma_prod` created in Rucio, please check slack.

bin/adhoc-scripts/MSRuleCleanerStandalone.py
```diff
@@ -0,0 +1,77 @@
...
+    reqRecords = {}
+    for status in reqStatus:
+        logger.info("Fetching requests in status: %s", status)
+        design = 'ReqMgr'
+        view = 'bystatusandtime?descending=false'
+        view += '&startkey=["%s",%s]' % (status, startTimeSec)
+        view += '&endkey=["%s",%s]' % (status, endTimeSec)
+        view += '&include_docs=true'
+        reqList.extend([req['doc'] for req in couchDB.loadView(design, view)['rows']])
+        reqNames.extend([req['RequestName'] for req in reqList])
```
amaltaro, line 65: perhaps print the number of requests retrieved in each status as well, just so we have an idea of how huge is this thing going to be.

amaltaro (commented):
Todor, I couldn't spot any issues with this script, besides the comments made in the code.

As an outcome of this scrip, besides deleting all the `wma_prod` rules on the output datasets, I think it would be useful to have two lists:
1) list of workflows that got all their wma_prod rules successfully deleted (or which had no rules at all)
2) list of workflows that had failures while deleting the rules (or any other issue in the cycle)

Such that, if needed, we could run this script again on a sub-set of the workflows in that date range.

PS.: Don't worry in making things as perfect as possible, this script will be executed once and be forgotten. What matters is that it gets the job done in a reasonably efficient way ;)

bin/adhoc-scripts/MSRuleCleanerStandalone.py
```diff
@@ -0,0 +1,77 @@
...
+
+from WMCore.MicroService.Unified.MSRuleCleaner import MSRuleCleaner
+from WMCore.Database.CMSCouch import Database as CouchDB
+
+FORMAT = "%(asctime)s:%(levelname)s:%(module)s:%(funcName)s(): %(message)s"
+logging.basicConfig(stream=sys.stdout, format=FORMAT, level=logging.DEBUG)
+logger = logging.getLogger()
+
+# Service config
+msConfig = {"enableRealMode": False,
```
amaltaro, line 17: Don't forget this configuration will have to be updated as well.

bin/adhoc-scripts/MSRuleCleanerStandalone.py
todor-ivanov: Done

bin/adhoc-scripts/MSRuleCleanerStandalone.py
todor-ivanov: I'd rather make a set of a runtime arguments and this one to be part of them. Working on that  at this very moment.

bin/adhoc-scripts/MSRuleCleanerStandalone.py
todor-ivanov: This should also go to the set of runtimie argument I am talking about in the comment bellow.

todor-ivanov:
Thanks for your review @amaltaro 
>  1.  list of workflows that got all their wma_prod rules successfully deleted (or which had no rules at all)
>   2. list of workflows that had failures while deleting the rules (or any other issue in the cycle)

I was thinking the same but it will require changes not to the adhock script but in MSRulecleaner itself, which I wanted to avoid to the best I could. But anyway, I agree this is important. Let me see what I can think of.

bin/adhoc-scripts/MSRuleCleanerStandalone.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -195,13 +200,29 @@ def _execute(self, reqRecords):
             totalNumRequests += 1
             if self._checkClean(wflow):
                 cleanNumRequests += 1
+                cleanReqList.append(wflow['RequestName'])
```
amaltaro, line 203: Should I assume this `cleanReqList` will also contain those requests flagged as ForceArchive=True?

amaltaro:
It looks good to me Todor.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: No, they are accounted separately. And because we are speaking about ACDCs here, that's why those are checked explicitly for there type and accumulated in a separate counter as `Resubmissions`

todor-ivanov:
Closing this PR now. It is not supposed to ever be merged.

PR #10112, opened 2020-11-27
Temporarily stop staging out log tarball to CASTOR

### LINKED ISSUE #10111: LogCollect jobs failing to stage out to CASTOR
**Impact of the bug**
WMAgent

**Describe the bug**
Same issue as reported here: https://github.com/dmwm/WMCore/issues/10110

but this one is supposed to be used only for the very short term, where we need to decide what's the best workaround to avoid wasting grid resources. In other words, this issue is supposed to result in a fix for the wmagent branch only (and be applied to all the production agents).

**How to reproduce it**
Run any LogCollect job

**Expected behavior**
Given that LogCollect jobs are also responsible for cleaning up the /store/unmerged/logs logArchive tarballs, the best option at the moment would be to remove the stage out to CASTOR, perform only the one to CERN EOS and let those jobs to cleanup the logArchive files.

There is one problem with this approach though, which is the need to patch every single workflow sandbox as well, otherwise the fix won't have any effects on workflows that have already been acquired by the agents.

Another alternative would be to set LogCollect jobs to `submitfailed` in JobSubmitter, such that we don't even try to run them. Drawback is that logArchives will remain at the sites until someone/something else cleans them up.

amaltaro:
Fixes #10111 

#### Status
ready

#### Description
<Description of the changes proposed.>

* stop trying to write log tarballs to CERN CASTOR
* write log tarball only to CERN EOS, and make that mandatory (thus, fail the job if it fails too)
* delete all the input files assigned to the job (if the stage out step succeeds)

NOTE: this is not a permanent change. It will be used until CERN CTA implementation is completed and available in production.

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -68,13 +68,6 @@ class LogCollect(Executor):
         # Set wait to over an hour
         waitTime = overrides.get('waitTime', 3600 + (self.step.retryDelay * self.step.retryCount))
 
-        # hardcode CERN Castor T0_CH_CERN_MSS stageout parameters
-        castorStageOutParams = {}
-        castorStageOutParams['command'] = overrides.get('command', "xrdcp")
-        castorStageOutParams['option'] = overrides.get('option', "--wma-cerncastor")
-        castorStageOutParams['phedex-node'] = overrides.get('phedex-node', "T2_CH_CERN")
-        castorStageOutParams['lfn-prefix'] = overrides.get('lfn-prefix', "root://castorcms.cern.ch//castor/cern.ch/cms")
-
         # hardcode CERN EOS T2_CH_CERN stageout parameters
         eosStageOutParams = {}
         eosStageOutParams['command'] = overrides.get('command', "xrdcp")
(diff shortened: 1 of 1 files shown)


amaltaro:
[Fix #10112 - temporarily stop staging out log tarball to CASTOR](https://trello.com/c/VovBEwtU/234-fix-10112-temporarily-stop-staging-out-log-tarball-to-castor)

amaltaro:
This has been tested in my VM, either for new workflows and also in old workflows that had their sandboxes patched.

amaltaro:
All the production nodes (plus relval and backfill) have been patched and WorkQueueManager restarted.

In addition to that, I have patched ALL the sandboxes (still running on submit4 though), such that workflows already in the system stop failing LogCollect jobs. This is how we can patch/fix the workflow sandboxes:
https://raw.githubusercontent.com/amaltaro/ProductionTools/master/fix_sandbox.sh

running from the /data/srv/wmagent/current. Output log is sandbox_fix.log

PS.: I've fixed things in the wrong order on submit3, instead of applying the patch first + restart of the component, I started by fixing the workflow sandboxes. That means, there could be some workflows that were acquired during the patching of the sandboxes, and I'm not planning to try to find and fix those.
PPS.: LogCollect failures should stop by tomorrow evening or Sunday.

todor-ivanov (commented):
Sorry for the belated review @amaltaro 
The changes looks good to me.

amaltaro:
I knew I was missing something in the WMAgent release candidate ... porting this fix to 1.4.3_wmagent branch as well.

Issue #10115, opened 2020-11-30
[py2py3] Migration at level scr/python/A/B/C - slice 4


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 4th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 4th bulk of migration units 
at src/python/A/B/C level.

```plaintext
                           name  nfiles  gradual  future_pm
 WMCore_Storage_SiteLocalConfig       1     True          3
               WMCore_REST_Test       1     True          8
        WMCore_WMConnectionBase       1     True          3
     WMQuality_REST_ServerSetup       1     True          5
             WMCore_REST_Server       1     True         38
          WMCore_WMStats_WebGui       1     True          0
        WMCore_Services_Service       1     True          3
      WMCore_Services_WMArchive       2     True         32
  WMCore_Services_WMStatsServer       1     True          3
            WMCore_Services_DDM       1     True          5
```

#### Is your feature request related to a problem? Please describe.

The end goal is to have WMCore py2 and py3 compatible.

#### Describe the solution you'd like

Run futurize on the first bulk of units 
and then apply the necessary manual changes.

PR #10103, opened 2020-11-24
[py2py3] Migration at level scr/python/A/B/C - slice 2

mapellidario:
Fixes #10098 

#### Status

ready

#### Related PRs

This PR is rebased on `master`.

* Previous migration PR: #10057 (slice1, issue #10055 ). 
* Next migration PR: #10117 (slice3, issue #10114 )

#### Description

Run futurize and some manual changes on the first batch of src/python/A/B/C.

No notable changes or errors have been encountered during this PR.

#### Is it backward compatible (if not, which system it affects?)

It should be. No possible causes for errors have been spot by unit tests. 

#### External dependencies / deployment changes

Requires python-future in both py2 and py3 environments.

Issue #10098, opened 2020-11-23
[py2py3] Migration at level scr/python/A/B/C - slice 2


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. This is the second slice in which the migration has been divided into. We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), with summary information on top.

#### Impact of the new feature

We start migrating the second bulk of migration units at src/python/A/B/C level, which will include:

```plaintext
                                 name  #files  gradual  changes (+ and -)
          WMCore_ACDC_CollectionTypes       1     True          0
           WMCore_Agent_DefaultConfig       1     True          0
                 WMCore_Configuration       1     True         13
             WMCore_Database_Dialects       1     True          0
          WMCore_GroupUser_Decorators       1     True          0
                        WMCore_ReqMgr       2     True          0
                WMCore_Services_HTTPS       1     True         16
              WMCore_WMSpec_Utilities       1     True          0
 WMCore_WorkQueue_WorkQueueExceptions       1     True          0
         WMCore_FwkJobReport_FileInfo       1     True          2
            WMCore_MicroService_Tools       1     True          5
                WMCore_Services_Rucio       1     True         14
         WMCore_WebTools_FrontEndAuth       1     True          2
               WMCore_ACDC_Collection       1     True          0
                  WMCore_ACDC_Fileset       1     True          0
        WMQuality_Emulators_ReqMgrAux       1     True          1
             WMCore_Agent_ConfigDBMap       1     True          0
    WMCore_Storage_TrivialFileCatalog       1     True          7
      WMCore_WMSpec_ConfigSectionTree       1     True          6
        WMCore_WebTools_DefaultConfig       1     True          0
    WMQuality_Emulators_DefaultConfig       1     True          0
    WMQuality_Emulators_EmulatorSetup       1     True          0
                    WMCore_DAOFactory       1     True          3
          WMCore_Database_Transaction       1     True          0
  WMCore_JobSplitting_SplitterFactory       1     True          0
                 WMCore_WebTools_Page       1     True          3
            WMCore_Database_DBCreator       1     True          9
         WMCore_ResourceControl_MySQL      13     True          6
              WMCore_ThreadPool_MySQL       3     True          0
           WMCore_WMBS_CreateWMBSBase       1     True          2
[...]

PR #10057, opened 2020-11-06
[py2py3] Migration at level src/python/A/B/C - start

mapellidario:
Fixes #10055

#### Status

Ready

#### Related PRs

* Next migration PR: #10103 (slice2, issue #10098  )

#### Description

Run futurize and some manual changes on the first batch of src/python/A/B/C.

Overview of the changes:
- ` from builtins import str `
- ` from builtins import object `
- iterate over dictionary valus and items : `viewitems`, `viewvalues`

Overview based on file:
* `src/python/WMCore/Database/ResultSet.py`
  * L39, `r.keys()`:  `r` is not a dict, it is of type `sqlalchemy.engine.result.RowProxy` -> list(r) does not return a list of keys. Do not change that line with futurize!

#### Is it backward compatible (if not, which system it affects?)

It should be. Possible causes of errors:

* `src/python/WMCore/Services/UUIDLib.py` returns a `future.types.newstr.newstr` instead of a py2 `str`. The test `test/python/WMCore_t/Services_t/UUID_t.py` has been updated accordingly

#### External dependencies / deployment changes

requires python-future

#### Open questions (to be addressed in follow up work)

`src/python/WMCore/REST/Validation.py `

- is `_check_str` still needed? the docstring states "convert unicode to str (bytes) first to support cherrypy 3.2.2"

PR #10117, opened 2020-11-30
[py2py3] Migration at level scr/python/A/B/C - slice 3

### LINKED ISSUE #10114: [py2py3] Migration at level scr/python/A/B/C - slice 3
#### Intro

This is the 3rd slice in which the migration has been divided into. 

We start migrating the 3rd bulk of migration units 
at src/python/A/B/C level.

```plaintext
                     WMQuality_Emulators_AnalyticsDataCollector       1     True          8
                                             WMCore_REST_Format       1     True         13
                                              WMCore_REST_Tools       1     True          0
                                          WMCore_Services_MonIT       1     True          2
                                            WMCore_Services_McM       1     True          2
                                  WMCore_WebTools_Documentation       1     True          0
                                       WMCore_WebTools_Masthead       1     True          0
                                  WMCore_WebTools_RESTFormatter       1     True          5
                            WMCore_WebTools_SecureDocumentation       1     True          2
                                        WMCore_WebTools_Welcome       1     True          0
                               WMQuality_WebTools_RESTClientAPI       1     True          2
                                               WMCore_REST_Main       1     True          6
 WMComponent_AnalyticsDataCollector_DataCollectorEmulatorSwitch       1     True          3
                          WMComponent_DBS3Buffer_DBSBufferBlock       1     True         11
                          WMComponent_JobSubmitter_JobSubmitAPI       1     True          0
                        WMComponent_JobStatusLite_DefaultConfig       1     True          0
                     WMComponent_WorkQueueManager_DefaultConfig       1     True          0
                                              WMCore_Agent_Flow       2     True         30
```

mapellidario:
Fixes #10114 

#### Status

Ready

#### Related PRs

This PR is rebased on `master`

* Previous migration PR: #10103 (slice2, issue #10098 ). 
* Next migration PR: #10165 (slice4, issue #10115 )

#### Description

This needs to be merged _after_ #10103 (slice2, issue #10098) , as it is rebased upon the branch used in that PR.

Notable changes: 
- `src/python/WMCore/REST/Format.py`
  * it uses xml.sax.saxutils [py2 doc](https://docs.python.org/2/library/xml.sax.utils.html) [py 3.8 doc](https://docs.python.org/3.8/library/xml.sax.utils.html), this requires some care with strings management
  * it uses `types.FileType`, which is not available in python3 anymore. Our solution is: instead of checking if `stream` is a file, we can check if `stream` has the desired property that we need. In this case, `stream` is passed as an argument to [cherrypy.lib.file_generator](https://github.com/cherrypy/cherrypy/blob/2a8aaccd649eb1011382c39f5cd93f76f980c0b1/cherrypy/lib/__init__.py#L64), which only uses `file.read()`, so instead of checking if `stream` is a file, we can check if `stream` as a `.read()` attribute. (suggestion from [stackoverflow](https://stackoverflow.com/a/59506892) )

- `src/python/WMCore/REST/Format.py`
  - some changes on how strings are handled was necessary. This can cause problems that are very hard to catch with jenkins CI only and will require extensive tests in testbed.

No changes to `test/python/*`

#### Related PRs

This PR should be merged after #10098 

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
@@ -5,6 +5,7 @@ DBSBufferBlock
 This is a block object which will be uploaded to DBS
 """
 
+from builtins import object
 import time
 import logging
 import copy
@@ -24,7 +25,7 @@ class DBSBufferBlockException(WMException):
 
 
 
-class DBSBufferBlock:
+class DBSBufferBlock(object):
     """
     _DBSBufferBlock_
 
@@ -129,7 +130,7 @@ class DBSBufferBlock:
         fileDict['auto_cross_section'] = 0.0
 
         # Do the checksums
-        for cktype in dbsFile['checksums'].keys():
+        for cktype in dbsFile['checksums']:
             cksum = dbsFile['checksums'][cktype]
             if cktype.lower() == 'cksum':
                 fileDict['check_sum'] = cksum
@@ -455,14 +456,14 @@ class DBSBufferBlock:
         self.startTime = blockInfo.get('creation_date')
         self.inBuff    = True
 
-        if 'status' in blockInfo.keys():
+        if 'status' in blockInfo:
             self.status = blockInfo['status']
             if self.status == "Pending":
                 self.data['block']['open_for_writing'] = 0
 
             del blockInfo['status']
 
-        for key in blockInfo.keys():
+        for key in blockInfo:
             self.data['block'][key] = blockInfo.get(key)
 
     def convertToDBSBlock(self):
@@ -489,7 +490,7 @@ class DBSBufferBlock:
         for key in self.data:
             if key in keyToRemove:
                 continue
-            elif key in dbsBufferToDBSBlockKey.keys():
+            elif key in dbsBufferToDBSBlockKey:
                 block[dbsBufferToDBSBlockKey[key]] = copy.deepcopy(self.data[key])
             else:
                 block[key] = copy.deepcopy(self.data[key])
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
@@ -103,9 +103,9 @@ class DBSBufferBlock(object):
             msg =  "Duplicate file inserted into DBSBufferBlock: %i\n" % (dbsFile['id'])
             msg += "Ignoring this file for now!\n"
             logging.error(msg)
-            logging.debug("Block length: %i" % len(self.files))
+            logging.debug("Block length: %i", len(self.files))
             l = sorted([x['id'] for x in self.files])
-            logging.debug("First file: %s    Last file: %s" % (l[0], l[-1]))
+            logging.debug("First file: %s    Last file: %s", l[0], l[-1])
             return
 
         for setting in self.data['close_settings']:
@@ -299,7 +299,7 @@ class DBSBufferBlock(object):
 
         try:
             if datasetName[0] == '/':
-                junk, primary, processed, tier = datasetName.split('/')
+                _, primary, processed, tier = datasetName.split('/')
             else:
                 primary, processed, tier = datasetName.split('/')
         except Exception:
--- a/src/python/WMCore/REST/Format.py
+++ b/src/python/WMCore/REST/Format.py
@@ -1,5 +1,8 @@
 from __future__ import print_function
 
+from builtins import str, bytes, object
+from future.utils import viewitems
+
 import hashlib
 import json
 import types
@@ -46,9 +49,14 @@ class RESTFormat(object):
         # to a single string. Convert files to 1MB chunks.
         if stream is None:
             stream = ['']
-        elif isinstance(stream, basestring):
+        elif isinstance(stream, (str, bytes)):
             stream = [stream]
-        elif isinstance(stream, types.FileType):
+        elif hasattr(stream, "read"):
+            # types.FileType is not available anymore in python3,
+            # using it raises pylint W1624.
+            # Since cherrypy.lib.file_generator only uses the .read() attribute
+            # of a file, we simply check if stream.read() is present instead.
+            # https://github.com/cherrypy/cherrypy/blob/2a8aaccd649eb1011382c39f5cd93f76f980c0b1/cherrypy/lib/__init__.py#L64
             stream = cherrypy.lib.file_generator(stream, 512 * 1024)
 
         return self.stream_chunked(stream, etag, *self.chunk_args(stream))
@@ -118,13 +126,15 @@ class XMLFormat(RESTFormat):
         """Render an object `obj` into XML."""
         if isinstance(obj, type(None)):
             result = ""
-        elif isinstance(obj, (unicode, str)):
+        elif isinstance(obj, str):
             result = xml.sax.saxutils.escape(obj).encode("utf-8")
+        elif isinstance(obj, bytes):
+            result = xml.sax.saxutils.escape(obj)
         elif isinstance(obj, (int, float, bool)):
             result = xml.sax.saxutils.escape(str(obj)).encode("utf-8")
         elif isinstance(obj, dict):
             result = "<dict>"
-            for k, v in obj.iteritems():
+            for k, v in viewitems(obj):
                 result += "<key>%s</key><value>%s</value>" % \
                   (xml.sax.saxutils.escape(k).encode("utf-8"),
                    XMLFormat.format_obj(v))
@@ -316,14 +326,17 @@ class PrettyJSONHTMLFormat(PrettyJSONFormat):
         """Render an object `obj` into HTML."""
         if isinstance(obj, type(None)):
             result = ""
-        elif isinstance(obj, (unicode, str)):
+        elif isinstance(obj, str):
             obj = xml.sax.saxutils.quoteattr(obj)
+            result = "<pre>%s</pre>" % obj if '\n' in obj else obj
+        elif isinstance(obj, bytes):
+            obj = xml.sax.saxutils.quoteattr(str(obj, "utf-8"))
             result = "<pre>%s</pre>" % obj if '\n' in obj else obj
         elif isinstance(obj, (int, float, bool)):
             result = "%s" % obj
         elif isinstance(obj, dict):
             result = "<ul>"
-            for k, v in obj.iteritems():
+            for k, v in viewitems(obj):
                 result += "<li><b>%s</b>: %s</li>" % (k, PrettyJSONHTMLFormat.format_obj(v))
             result += "</ul>"
         elif is_iterable(obj):
@@ -411,7 +424,7 @@ class RawFormat(RESTFormat):
             etag.invalidate()
             raise
 
-class DigestETag:
+class DigestETag(object):
     """Compute hash digest over contents for ETag header."""
     algorithm = None
 
--- a/src/python/WMCore/WebTools/RESTFormatter.py
+++ b/src/python/WMCore/WebTools/RESTFormatter.py
@@ -7,6 +7,9 @@ appropriate format and sets the CherryPy header appropriately.
 
 Could add YAML via http://pyyaml.org/
 """
+
+from builtins import str, bytes
+
 import json
 from types import GeneratorType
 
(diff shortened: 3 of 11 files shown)


amaltaro:
@mapellidario Dario, I'm making some changes to the jenkins configuration. Hopefully nothing will break, but we never know ;)

mapellidario:
The warning from `pylint py3k`

```plaintext
    src/python/WMCore/REST/Error.py
        W1645, line 74 in RESTError.__str__: Exception.message removed in Python 3
```

is inherited from #10055, it is not related to this PR. Anyway, it is a false positive because we manually added the `.message` attribute to the RESTError class.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -455,14 +456,14 @@ def FillFromDBSBuffer(self, blockInfo):
         self.startTime = blockInfo.get('creation_date')
         self.inBuff    = True
 
-        if 'status' in blockInfo.keys():
+        if 'status' in blockInfo.keys():  # did not modernize, lack of test
```
amaltaro, line 459:
```suggestion
        if 'status' in blockInfo:
```
can it be something like that?

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -455,14 +456,14 @@ def FillFromDBSBuffer(self, blockInfo):
...
             self.status = blockInfo['status']
             if self.status == "Pending":
                 self.data['block']['open_for_writing'] = 0
 
             del blockInfo['status']
 
-        for key in blockInfo.keys():
+        for key in blockInfo.keys():  # did not modernize, lack of test
```
amaltaro, line 466:
```suggestion
        for key in blockInfo:
```
should be enough, I guess.

src/python/WMComponent/TaskArchiver/DataCache.py
```diff
@@ -1,12 +1,13 @@
...
 
     @staticmethod
     def getFinishedWorkflows():
-        return DataCache._finishedWFs;
+        return DataCache._finishedWFs
 
     @staticmethod
     def setFinishedWorkflows(workflowLists):
-        DataCache._finishedWFs = workflowLists;
+        DataCache._finishedWFs = workflowLists
```
amaltaro, line 13: good stuff!

src/python/WMCore/Agent/Daemon/Create.py
```diff
@@ -74,14 +75,14 @@ def daemonize(stdout='/dev/null', stderr=None, stdin='/dev/null',
     # Open file descriptors and print start message
     if not stderr:
         stderr = stdout
-    si = file(stdin, 'r')
-    so = file(stdout, 'a+')
-    se = file(stderr, 'a+', 0)
+    si = open(stdin, 'r')
```
amaltaro, line 78: Hmm, has this been rebased on top of: https://github.com/dmwm/WMCore/pull/10057
instead of the PR you mention at the top level description?

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -6,6 +6,8 @@
 
 from __future__ import print_function, division
 
+from builtins import str, object
+from past.builtins import basestring
```
amaltaro, line 10: If possible, I would not import from past.

test/python/WMCore_t/MicroService_t/Tools_t/PycurlRucio_t.py
```diff
@@ -91,7 +93,7 @@ def testGetBlocksAndSizeRucio(self):
         self.assertTrue(len(resp) == 3)
         self.assertTrue(CONT2 in resp)
         self.assertTrue(len(resp[CONT4]) > 3)
-        self.assertItemsEqual(resp[CONT4][BLOCK].viewkeys(), ["blockSize", "locations"])
+        self.assertItemsEqual(resp[CONT4][BLOCK].keys(), ["blockSize", "locations"])
```
amaltaro, line 96: I had the impression this wouldn't work on py3. Instead we should cast it to `list(my_dict)`, no?

amaltaro (changes requested):
Dario, not a great review because this PR seems to have been rebased on top of: https://github.com/dmwm/WMCore/pull/10057
making the review process a bit more complicated. Nonetheless, you have a few comments/suggestions/questions along the code.

test/python/WMCore_t/MicroService_t/Tools_t/PycurlRucio_t.py
mapellidario: This is part of #10103, so i will mark it as resolved here. In any case, your suggestion is correct! Though, it is not broken in py3: `.keys()` in py3 is the same as `.viewkeys()` in py2. But if we use `.keys()` in both py2 and py3, it is going to be a list in py2 and a generator in py3. In this particular case it does not seem to be a problem since py2 is using a generator already, but it always nice to be consistent and have the same behavior in both py2 and py3.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
mapellidario: Yeah! you are correct. I saw that `.get()` function and imagined that `blockInfo` was some exotic class, but only now I realize that it is the safe way of accessing a the value given the key in a normal python dict. Thanks! (corrected in last push)

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
mapellidario: This was inherited from #10057, I will take care of it in that PR. This one has been cleaned from the relics of #10057 and #10103 and is ready to be review again

src/python/WMCore/REST/Format.py
```diff
@@ -46,9 +49,14 @@ def __call__(self, stream, etag):
         # to a single string. Convert files to 1MB chunks.
         if stream is None:
             stream = ['']
-        elif isinstance(stream, basestring):
+        elif isinstance(stream, str) or isinstance(stream, bytes):
```
amaltaro, line 52: This comparison can be made against a tuple, something like
```suggestion
        elif isinstance(stream, (str, bytes)):
```

src/python/WMCore/WebTools/RESTFormatter.py
```diff
@@ -66,26 +69,31 @@ def to_string(self, data):
             return self.json(data)
         if isinstance(data, dict) or isinstance(data, list):
             return json.dumps(data)
-        return str(data)
+        # Do not be misguided by the function name, this must return bytes, 
+        # otherwise the unit test
+        # testSupportedFormat (WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest)
+        # fails with "ValueError: Page handlers MUST return bytes. 
+        # Use tools.encode if you wish to return unicode."
+        return bytes(str(data), "utf-8")
```
amaltaro, line 77: Honestly, I'm not sure whether `bytes` is really the data type that we need to return here. 
I would expect something like `data.encode('utf-8')` to be enough, which would still be a byte string but not preceded by `b'`, right? I'm not even sure this comment makes sense though.

I fear I won't ever have a full understanding of strings in python...

amaltaro (changes requested):
Dario, there is one minor update that would be great to have (mentioned in the code). And another random comment. Thanks

src/python/WMCore/WebTools/RESTFormatter.py
mapellidario: #### .encode() vs bytes()

As you mentioned, `str(data).encode("utf-8")` would be the same as `bytes(str(data), "utf-8")` and we can use it if you prefer! They behave exactly the same in both python3 and modernized (with python future) python2


```python
> python3
Python 3.9.1 (default, Dec  8 2020, 07:51:42) 
[GCC 10.2.0] on linux
>>> data = {"key1": 2, "key2": 3.14}
>>> str(data).encode("utf-8")
b"{'key1': 2, 'key2': 3.14}"
>>> type(str(data).encode("utf-8"))
<class 'bytes'>
>>> bytes(str(data), "utf-8")
b"{'key1': 2, 'key2': 3.14}"
>>> type(bytes(str(data), "utf-8"))
<class 'bytes'>
```

```python
Python 2.7.13 (default, Sep 14 2017, 23:43:58) 
[GCC 4.9.2] on linux2
>>> from builtins import str, bytes
>>> data = {"key1": 2, "key2": 3.14}
>>> str(data).encode("utf-8")
b"{'key2': 3.14, 'key1': 2}"
>>> type(str(data).encode("utf-8"))
<class 'future.types.newbytes.newbytes'>
>>> bytes(str(data), "utf-8")
b"{'key2': 3.14, 'key1': 2}"
>>> type(bytes(str(data), "utf-8"))
<class 'future.types.newbytes.newbytes'>

```

I think that `b'` is only something that is used by the `__repr__` if the class  `bytes` to remind us that we are not dealing with a sequence of unicode but with a sequence of bytes.


#### do we really need to return bytes here?

When we use python2 interpreter, yes, it seems so. 

If we have `in WMCore/WebTools/RESTFormatter.RESTFormatter.to_string()`

```python
from builtins import str
[...]
    def to_string(self, data):
        if isinstance(data, GeneratorType):
            return self.json(data)
        if isinstance(data, dict) or isinstance(data, list):
            return json.dumps(data)
        return str(data)
```

then `python setup.py test --buildBotMode=true --reallyDeleteMyDatabaseAfterEveryTest=true --testCertainPath=test/python/WMCore_t/WebTools_t/RESTFormat_t.py` fails with

```plaintext
  File "/home/dmwm/unittestdeploy/wmagent/1.4.0-comp/sw/slc7_amd64_gcc630/external/py2-cherrypy/17.4.0-comp/lib/python2.7/site-packages/cherrypy/_cprequest.py", line 630, in respond
    self._do_respond(path_info)
  File "/home/dmwm/unittestdeploy/wmagent/1.4.0-comp/sw/slc7_amd64_gcc630/external/py2-cherrypy/17.4.0-comp/lib/python2.7/site-packages/cherrypy/_cprequest.py", line 689, in _do_respond
    response.body = self.handler()
  File "/home/dmwm/unittestdeploy/wmagent/1.4.0-comp/sw/slc7_amd64_gcc630/external/py2-cherrypy/17.4.0-comp/lib/python2.7/site-packages/cherrypy/_cprequest.py", line 794, in __set__
    raise ValueError(self.unicode_err)
ValueError: Page handlers MUST return bytes. Use tools.encode if you wish to return unicode.
```

It seems that cherrypy in python2 requires `to_string()` to return py2-`str` (which is bytes).

I could simply remove `from builtins import str`, but I assumed (hoped, bet) that cherrypy would require `to_string()` to return bytes also in python3, so I chaged it to

```python
from builtins import str
[...]
    def to_string(self, data):
        if isinstance(data, GeneratorType):
            return self.json(data)
        if isinstance(data, dict) or isinstance(data, list):
            return json.dumps(data)
        return bytes(str(data), "utf-8") # or also `str(data).encode("utf-8")`
```

This is an assumption, it may as weel be that cherrypy uses requires `to_string()` to return `str()` in both py2 (which means bytes) and py3 (which means unicode). I have no definitive answer now, we would need a py3 environment to test my  assumption, but cherrypy code have the same error in their py3 code

https://github.com/cherrypy/cherrypy/blob/c524a0da73a85ec4dd4ae6b1512c25854b19c8de/cherrypy/_cprequest.py#L779

So I would return bytes in `to_string()` also in py3.

src/python/WMCore/REST/Format.py
mapellidario: You are correct! I pushed this change in a new commit

src/python/WMCore/WebTools/RESTFormatter.py
amaltaro: You're right. It's hard to be sure unless we have the python3 environment.
Do you think it would be helpful to add a # TODO, or # FIXME saying that this needs to be verified when running with python3? We have so many in the code, that maybe it won't make much difference...

mapellidario: I fear the moment that we will be hit by a storm of errors and having a place to start looking for possible causes would be of much help. I think that such a flag would be useful! Or we can even have a new column in the python3 project in which we put possibly risky closed issues.
Or would you prefer to open new issues to keep track of these? (I fear that we would open a toooon of issues :sweat_smile:  )

amaltaro: I think all these options would be good...perhaps the GH issue would have a "bigger impact" and easier to track(?) Please follow your taste here, anything works for me ;)

mapellidario: Created #10222 to track this discussion

amaltaro:
Thanks Dario! Can you please squash the src commits?

amaltaro:
@mapellidario explicitly pinging you here, in case you missed it. Once we merge this and the 10187 PR, I will cut a new release and start run further tests.

mapellidario:
Sorry! The `git push` command remained in my mind and was not transferred to my keyboard!

PR #10153, opened 2020-12-04
Add archival logic to MSRuleCleaner.

### LINKED ISSUE #10018: Move workflow archival from ReqMgr2 CherryPy to MSRuleCleaner
**Impact of the new feature**
MicroServices

Once we have the MSRuleCleaner developed It will be the service to take care  of the final workflow status transition to `{normal, aborted, rejected}-archived`, which means we will need to remove the logic from the CherryPy thread which is currently dealing with that. The place where the change should happen is here  [1].

[1]
https://github.com/dmwm/WMCore/blob/5961d2229b1e548e58259c06af154f33bce36c68/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py#L101
 

Let MSRuleCleaner take care of this status transition. and remove the code from StatusChangeTasks.py

No allternative

**Additional context**
No

todor-ivanov:
Fixes #10018 

#### Status
ready

#### Description
Moves the workflows archival logic from the CherryPy thread to the MSRuleCleaner. This is implemented in two steps:
   * Cleans all the LogDB records for the current workflow
   * Makes the needed request status transition in reqmgr   

YES

#### Related PRs
Needs some more configuration parameters such as:
`wmstatsUrl`, 
`logDBUrl`,
`logDBReporter`,
`archiveDelayHours`

The related deployment config PRs:
https://github.com/dmwm/deployment/pull/1010
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/54
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/55

NO


todor-ivanov:
Alan @amaltaro please go ahead and take a look, while I prepare a fix for the Unit test that I expect to break with this PR, so we may discuss the so proposed solution here.  Thanks in advance!

todor-ivanov:
test this please

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -60,9 +64,14 @@ def __init__(self, msConfig, logger=None):
         self.msConfig.setdefault("rucioWmaAccount", "wma_test")
         self.msConfig.setdefault("rucioMStrAccount", "wmcore_transferor")
         self.msConfig.setdefault('enableRealMode', False)
+
         self.mode = "RealMode" if self.msConfig['enableRealMode'] else "DryRunMode"
         self.emailAlert = EmailAlert(self.msConfig)
         self.curlMgr = RequestHandler()
+        self.wmstatsSvc = WMStatsServer(self.msConfig["wmstatsUrl"], logger=self.logger)
+        self.logDB = self.logDB = LogDB(self.msConfig["logDBUrl"],
```
amaltaro, line 72: buggy :)

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
         wflow['IsClean'] = self._checkClean(wflow)
         return wflow
 
+    def _checkLogDBClean(self, wflow):
```
amaltaro, line 373: I would call it `isLogDBClean`. The docstring needs some working IMO.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
+
+    def findTargStatus(self, wflow):
+        """
+        Find the proper targeted archival status
+        :param  wflow:      A MSRuleCleaner workflow representation
+        :return wflow:
+        """
+        # Check the available status transitions before we decide the final status
+        targetStatusList = RequestStatus.REQUEST_STATE_TRANSITION.get(wflow['RequestStatus'], [])
+        targetStatusReg = re.compile(r'.*archived')
```
amaltaro, line 406: How about making it an object variable and compile it only once?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
+        A function to be used for cleaning all the records related to a workflow in logDB.
+        :param wflow:       A MSRuleCleaner workflow representation
+        :return:            Bool
+        """
+        cleanStatus = False
+        try:
+            self.logger.info("Deleting %s from LogDB WMStats...", wflow['RequestName'])
+            res = self.logDB.delete(wflow['RequestName'], agent=False)
+            if res == 'delete-error':
+                self.logger.error("Failed to delete logDB docs for wflow: %s", wflow['RequestName'])
```
amaltaro, line 424: With a `delete-error` response, you still report a cleanStatus=True here.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
+            cleanStatus = True
+        except Exception as ex:
+            msg = "General Exception while cleaning LogDB records for wflow: %s : %s"
+            self.logger.exception(msg, wflow['RequestName'], str(ex))
+        return cleanStatus
+
+    def _checkArchDelayExpired(self, wflow):
+        """
+        A function to check Archival Expiration Delay based on the inromation
```
amaltaro, line 433: There is a typo here. And even though it's just a docstring, I'd suggest not to user capitals where it's not needed, like "Archival Expiration Delay"

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
+        returned by WMStatsServer regarding the time of the last request status transition
+        :param wflow:      MSRuleCleaner workflow representation
+        :return:           Bool
+        """
+        archDelayExpired = False
+        currentTime = int(time.time())
+        threshold = self.msConfig['archiveDelayHours'] * 3600
+        outputMask = ['RequestTransition']
+        inputConditon = {'RequestName': wflow['RequestName']}
```
amaltaro, line 442: This will make one call per workflow, but we should actually make a single call per status, and simply go through the workflows.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -70,6 +70,8 @@ def docSchema(self):
             ('ParentageResolved', True, bool),
             ('PlineMarkers', None, list),
             ('IsClean', False, bool),
+            ('IsLogDBClean', False, bool),
+            ('IsArchDelayExpired', False, bool),
```
amaltaro, line 74: I would completely write this name, such as `IsArchivalDelayExpired`

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
+        return wflow
+
+    def _archive(self, wflow):
+        """
+        A function to make the actual request status transition during the
+        archival step.
+        :param wflow:       A MSRuleCleaner workflow representation
+        :return:            Bool
+        """
+        archiveStatus = False
```
amaltaro, line 467: I somehow have the feeling that we are overloading the usage of "private" methods in this module, basically every method has its own private method to execute the action. IMO, I'd move this logic under the `archive` method itself.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
+            msg = "Could not determine which archival status to target for: %s" % wflow['RequestName']
+            raise MSRuleCleanerArchival(msg)
+        elif not wflow['IsLogDBClean']:
+            if not self._cleanLogDB(wflow):
+                msg = "Could not clean logDB for: %s" % wflow['RequestName']
+                raise MSRuleCleanerArchival(msg)
+        elif not wflow['IsArchDelayExpired']:
+            msg = "Archival delay period has not yet expired for: %s." % wflow['RequestName']
+            raise MSRuleCleanerArchival(msg)
+        elif not self.msConfig['enableRealMode']:
```
amaltaro, line 497: I guess you simply want to log a DRY-RUN thing here?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,139 @@ def setClean(self, wflow):
...
-        # NOTE: check allowed status transitions with:
-        #       https://github.com/dmwm/WMCore/blob/5961d2229b1e548e58259c06af154f33bce36c68/src/python/WMCore/ReqMgr/DataStructs/RequestStatus.py#L171
+        # Make all the needed checks before trying to archive
         if not (wflow['IsClean'] or wflow['ForceArchive']):
             msg = "Not properly cleaned workflow: %s" % wflow['RequestName']
             raise MSRuleCleanerArchival(msg)
+        elif not wflow['TargetStatus']:
+            msg = "Could not determine which archival status to target for: %s" % wflow['RequestName']
+            raise MSRuleCleanerArchival(msg)
+        elif not wflow['IsLogDBClean']:
```
amaltaro, line 490: I don't understand why raising exceptions here. Same for the two block belows (IsArchDelayExpired and enableRealMode). Shouldn't we simply log (if not already) an error and move to the next workflow?

amaltaro (changes requested):
Todor, it's going in the right direction.
However, I have made some minor and major comments along the code.

In addition to that, I'd suggest the following:
* docstrings (especially the return: documentation) need to be worked further. I'm not sure we need that extra spacing as well..
* please make a 2nd commit - in this same PR - with the archival removal from the ReqMgr2 CherryPy thread. This has to be an atomic operation to avoid issues, and the way to have it is by providing everything in the same PR.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: @amaltaro excuse my ignorance here ;), but please do elaborate!

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Ok, sorry again. I will need some more info about what exactly do you think needs to be worked on in the docstring. I will try to make it as clear as possible. 
About the name: I was following a pattern here and all private methods that are not supposed to be called independently are returning a `bool` variable instead of a `wflow` object and all of them I was prefixing them with the same substring `_check....` But if you think I should break the pattern I will rename it.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: I was also thinking the same and hesitating initially. But now I am sure. 
Done now.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: That was a good catch... how did I miss that myself?!
Let me see how I can fix it

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Thanks! 
Fixed

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Ok, this one actually took me longer than I would like spend but anyway I think I managed to solve it in the best possible manner. I had to sacrifice memory footprint for run time complexity, but at the end I figured out that we can afford it.  @amaltaro please take a look and give me your opinion.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Actually no. If we do not leave the function properly with raised exception here, then the workflow is going to get archived. 
But if you simply mean that the Error message should to be changed to "DRY-RUN  mode .... ", then I agree, that should be better.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: As I started to explain in the previous section we must exit the function in order to avoid actual status transition to one  of `*-archival` statuses. And the best way to do this is with an exception, because that way we can signal the upper level function not to update any of the relevant counters regarding the failing worklow.

todor-ivanov:
@amaltaro just take a look at the currently proposed changes. Mostly speaking about the in memory approach for the status transition history. This is not the final version yet. There are still the changes to the cherrypy thread to come.

todor-ivanov:
test this please

todor-ivanov:
test this please

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: repeated assignment

amaltaro (changes requested):
Todor, this is a very incomplete review because I wanted to give you some guidance on the way you are fetching workflow information here.

The `_buildStatusHistory` method is actually not necessary, because the information retrieved there is a subset of what it already retrieves through `getRequestRecords` method.
However, I would actually change the service we fetch workflow information from here, instead of using reqmgr2, I'd suggest to use wmstatsserver which has all that cached. In addition to that, it looks like you are only interested in 4 or 5 workflow properties, so maybe we could use this `getFilteredActiveData` wmstats method and fetch only the information we need. This MS polling cycle is very large, so it won't be too much of a load on the wmstatsserver.

Ping me on slack if you'd like to discuss it further.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: I see. However, we should distinguish the following two use cases:
1) workflow cannot be archived because there was an unexpected problem (e.g. wrong status, not all docs successfully cleaned, etc)
2) workflow is in an expected state (e.g. archival delay hasn't been met, service is in dry-run).

In addition to that, errors that are not really errors shouldn't be logged as error. What do you think?

todor-ivanov:
@amaltaro Alan, Please take a look at this PR again. I managed to also catch two bugs and cover them with my latest commit. There will still be two unittests that are about to break, but this is because I need to figure out how to deal with the LogDB session inside the unit test setup. Anyway, please have a look.

ps. In case it will be easier fro you just tell me and I will sqaush all the commits in two sets - one for the feature changes and one for the unit tests (this PR kind of became heavy to track already)

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Aaaand... this is a silly typo!!!  Fixing it now.
Thanks!

amaltaro:
>... but this is because I need to figure out how to deal with the LogDB session inside the unit test setup ..

Todor, I think I have already struggled with LogDB in unit tests as well. Try setting it to `None` (or have a look at other unit tests)

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: This was remade yet again. To use `reqmgr` instead of `wmstatsServer` as @amaltaro was suggesting, and is actually a better looking piece now. Resolving this conversation now.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: I totally agree. And the only way out of this I see if we put a second exception there and (re)name them both accordingly (Honestly speaking I do not like the name of the current exception either - it does not give any clue that this is actually an exception or an error of some kind)

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -70,7 +70,10 @@ def docSchema(self):
             ('ParentageResolved', True, bool),
             ('PlineMarkers', None, list),
             ('IsClean', False, bool),
-            ('ForceArchive', False, bool)]
+            ('IsLogDBClean', False, bool),
+            ('IsArchivalDelayExpired', False, bool),
```
amaltaro, line 74: Given that `IsArchivalDelayExpired` depends solely on the request dictionary, I'm inclined to say that moving that code over to this module would make things more clear.

amaltaro (changes requested):
Sorry Todor, I forgot publishing half of the review I had made yesterday. Here it goes a few more comments and suggestions

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: In general I tend to create a separate method when a separate call just to the check method is expected to be made somewhere else into the code. In all other cases, I do agree with you that having the logic enclosed inside the function that is about to change the flag status itself is the best. I will see what I can do with the `_checkLogDB` method.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,124 @@ def setClean(self, wflow):
...
+        """
+        A function to set the 'IsLogDBClean' flag based on the presence of any
+        records in LogDB for the current workflow.
+        :param  wflow:      A MSRuleCleaner workflow representation
+        :return wflow:
+        """
+        wflow['IsLogDBClean'] = self._checkLogDBClean(wflow)
+        return wflow
+
+    def findTargStatus(self, wflow):
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,124 @@ def setClean(self, wflow):
...
+
+    def _cleanLogDB(self, wflow):
+        """
+        A function to be used for cleaning all the records related to a workflow in logDB.
+        :param wflow:       A MSRuleCleaner workflow representation
+        :return:            Bool
+        """
+        cleanStatus = False
+        try:
+            self.logger.info("Deleting %s from LogDB WMStats...", wflow['RequestName'])
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,124 @@ def setClean(self, wflow):
...
+            res = self.logDB.delete(wflow['RequestName'], agent=False)
+            if res == 'delete-error':
+                self.logger.error("Failed to delete logDB docs for wflow: %s", wflow['RequestName'])
+                return cleanStatus
+            self.setLogDBClean(wflow)
```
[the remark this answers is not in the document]
todor-ivanov: So in this case we do not fetch the document from a couchdb view, but from the logDb instance itself, and that's  why I was expecting the change to be kind of instant. The additional check was done only because I did not want to have some leftovers behind, but I agree - In case we are  sure that once the deletion call is properly sent then the documents are 100% about to be deleted, then yes this additional call to the database would be redundant.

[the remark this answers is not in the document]
todor-ivanov: BTW I am about to make the change you requested here. Lets see how it goes.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +370,124 @@ def setClean(self, wflow):
...
+        threshold = self.msConfig['archiveDelayHours'] * 3600
+        try:
+            lastTransitionTime = wflow['RequestTransition'][-1]['UpdateTime']
+            if lastTransitionTime and (currentTime - lastTransitionTime) > threshold:
+                archDelayExpired = True
+        except KeyError:
+            self.logger.debug("Could not find status transition history for %s", wflow['RequestName'])
+        return archDelayExpired
+
+    def setArchDelayExpired(self, wflow):
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: At first glance I was about to agree with you immediately. But once I actually tried to implement it, I immediately ran into the need to import the `msConfig` object into `MSRuleCleanerWflow.py`, which makes things a little bit more complicate. And this made me immediately remember that we decided to keep this one as simple/generic as possible, but not putting a lot of weight to this class similar to `MSOutputTemplate.py`.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py

todor-ivanov:
Hi @amaltaro, Please go ahead and review the final version of this PR. I think I have addressed all the comments from the previous iteration.  And also the unit tests must be fixed now.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: Thanks!

amaltaro (commented):
Todor, code is looking almost ready to me. I had a few minor comments along the code and one - about cleaning LogDB - that might require some further work, but isn't a showstopper though.

In addition to that, please have a look at the PR report from jenkins. There are a few very easy improvements that can be made here (not only the bold ones).

Last but not least, it came to my mind today that we might want to decrease the polling cycle in the future (for a quicker workflow archival). I wonder if we make Rucio calls even if the archival delay hasn't been reached yet?

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: I do agree with you that keeping as little as possible in the workflow representation object would be kind of a cleaner approach, but I am not quite sure how much of an impact would that have on the overall memory footprint from a single run. Because we are basically re-initializing the wflow object in place with every new workflow (see here [1]). And the memory should be release right after we get out of scope - which I expect to happen once the `_dispathWflow` function returns a value. And given the fact that we are feeding it a wflow at a time I thought the increased memory consumption while keeping the whole transition history of a single workflow would be neglectable compared to the enormous and detailed list of workflows returned by `reqmgr` and which we keep in memory until the end of the whole cycle (this is when this one gets out of scope). Of course all that  needs to be confirmed with a full run of a proper profiler, which I need to admit I haven't done - so there is a possibility that I may be wrong here. 

So I would suggest we  continue now as it have been already tested, and we may come back to this once we start thinking (eventually) about feeding the whole service in a different way. Will keep the conversation not resolve though - just for a future reminder.   

[1]
https://github.com/dmwm/WMCore/blob/875c3c5a48d0d82df947cdf13466c469c33ca27a/src/python/WMCore/MicroService/Unified/MSRuleCleaner.py#L188-L190

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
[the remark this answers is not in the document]

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -139,6 +160,7 @@ def execute(self, reqStatus):
             requestRecords = {}
             for status in reqStatus:
                 requestRecords.update(self.getRequestRecords(status))
+                # self._buildStatusHistory(status)
```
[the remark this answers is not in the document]
todor-ivanov: Correct. Thanks!

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -60,9 +73,14 @@ def __init__(self, msConfig, logger=None):
         self.msConfig.setdefault("rucioWmaAccount", "wma_test")
         self.msConfig.setdefault("rucioMStrAccount", "wmcore_transferor")
         self.msConfig.setdefault('enableRealMode', False)
+
         self.mode = "RealMode" if self.msConfig['enableRealMode'] else "DryRunMode"
         self.emailAlert = EmailAlert(self.msConfig)
         self.curlMgr = RequestHandler()
+        self.targetStatusReg = re.compile(r'.*archived')
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +384,123 @@ def setClean(self, wflow):
...
+        if not logDBRecords:
+            cleanStatus = True
+        return cleanStatus
+
+    def setLogDBClean(self, wflow):
+        """
+        A function to set the 'IsLogDBClean' flag based on the presence of any
+        records in LogDB for the current workflow.
+        :param  wflow:      A MSRuleCleaner workflow representation
+        :return wflow:
```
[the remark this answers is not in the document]
todor-ivanov: Agreed and Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -358,23 +384,123 @@ def setClean(self, wflow):
...
-
-        # Check the available status transitions before we decide the final status
-        targetStatusList = RequestStatus.REQUEST_STATE_TRANSITION.get(wflow['RequestStatus'], [])
-        self.logger.info("targetStatusList: %s", targetStatusList)
+            raise MSRuleCleanerArchivalSkip(msg)
+        if not wflow['TargetStatus']:
+            msg = "Could not determine which archival status to target for: %s" % wflow['RequestName']
+            raise MSRuleCleanerArchivalError(msg)
+        if not wflow['IsLogDBClean']:
+            if not self._cleanLogDB(wflow):
```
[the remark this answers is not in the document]
todor-ivanov: Agreed, I was feeling the same way while putting it here, but couldn't come up with a better place though. I will see what I can do about it now.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
amaltaro: And here is where you contradict yourself :-), compared to your previous comment:
"""
And this made me immediately remember that we decided to keep this one as simple/generic as possible, but not putting a lot of weight to this class similar to MSOutputTemplate.py.
"""

I try to push for workflow-related logic inside the workflow class, and the argument you give me is that you're concerned with the memory footprint...

The approach I would adopt is to add all the workflow logic in the same place, same as I've done for the workflow object used in MSTransferor:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/DataStructs/Workflow.py

Another good candidate for this would be a helper function that internally evaluates whether the workflow can be archived, e.g.:
```
if myWflowObj.canArchive():
    blah
```
and then that `canArchive` method would be implemented in this MSRuleWflow class and evaluate all the flags we need to evaluate.

Anyhow, I just wanted to make clear that my proposal was to inject workflow logic to the workflow struct class, not to make it as simple as possible and leave logic to be implemented elsewhere. As you said, this code works and it has been tested, let's move on!

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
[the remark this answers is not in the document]
todor-ivanov: Done now.

todor-ivanov:
@amaltaro , I just commited one possible way of reshuffling some function calls so that we have the `_cleanLogDB()` called outside the archival method. Please take a look and let me know if you like it, so I can squash 'em all ;)

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: Agreed!

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: Marking it as resolved! Thanks

amaltaro (approved):

amaltaro:
Thanks for all the follow up, Todor.

Issue #10116, opened 2020-11-30
[py2py3] Migration at level scr/python/A/B/C - slice 5


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 5-th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 5-th bulk of migration units 
at src/python/A/B/C level.

```plaintext

                                    name  nfiles  gradual  future_pm
           WMCore_Services_UserFileCache       1     True          2
                 WMCore_Services_WMAgent       1     True          0
                    WMCore_Services_WMBS       1     True          0
                  WMCore_Storage_Execute       1     True          1
             WMCore_Storage_StageOutImpl       1     True          3
           WMCore_Storage_StageOutImplV2       1     True          3
                    WMCore_WMBS_WMBSBase       1     True          0
          WMCore_WorkQueue_WorkQueueBase       1     True          3
        WMCore_WebTools_DBSRESTFormatter       1     True          0
                    WMCore_REST_Services       1     True          0
 WMComponent_DBS3Buffer_DBSBufferDataset       1     True          0
               WMCore_Agent_HeartbeatAPI       1     True          1
                  WMCore_Services_ReqMgr       1     True          3
                        WMCore_WMBS_File       1     True          9
                     WMCore_WMBS_Fileset       1     True          1
                        WMCore_WMBS_Mask       1     True          6
                    WMCore_WMBS_Workflow       1     True          2
   WMCore_JobStateMachine_ConfigureState       1     True          0
                    WMCore_WMBS_WorkUnit       1     True          0
              WMCore_Database_CouchUtils       1     True          1
        WMCore_JobStateMachine_SummaryDB       1     True         12
                  WMCore_Services_FWJRDB       1     True          6
                        WMCore_WMLogging       1     True          0
               WMCore_WMSpec_Persistency       1     True         11
              WMCore_GroupUser_Interface       1     True          6
                  WMCore_GroupUser_Group       1     True          0
                   WMCore_GroupUser_User       1     True          0
              WMCore_MicroService_WebGui       1     True          0
                    WMCore_Services_CRIC       1     True          2
          WMQuality_Emulators_CRICClient       2     True          4
[...]

PR #10165, opened 2020-12-08
[py2py3] Migration at level scr/python/A/B/C - slice 4

### LINKED ISSUE #10115: [py2py3] Migration at level scr/python/A/B/C - slice 4

mapellidario:
Fixes #10115 

#### Status

In development 

#### Related PRs

This PR is rebased on `master`

* Previous migration PR: #10117 (slice3, issue #10114  ). 
* Next migration PR: #10262 (slice5, issue #10116 )

#### Description

##### Notable change to `src/python/WMCore/Services/Service.py`

Simply adding `from builtins import str` breaks some unit tests in `test/python/WMCore_t/Services_t/Service_t.py`.

In particular the problems arise from the following lines combined with `from builtins import str`:

* https://github.com/dmwm/WMCore/blob/2e11c4ba1c280e448895dc5141e5903b647bca10/src/python/WMCore/Services/Service.py#L296
* https://github.com/dmwm/WMCore/blob/2e11c4ba1c280e448895dc5141e5903b647bca10/src/python/WMCore/Services/Service.py#L303

If I do not import `from builtins import str`, then everything is fine.

I checked the type of `data` in the unit tests and it is always py2 `str` (this means that it is a string composed of a sequence bytes) when it is not a dict/list, so doing `str(a)` in py2 has no effect. 
When I import `from builtins import str` it tries to convert a py2 `str` (sequence of bytes) to a py3 `str` (sequence of unicode codepoints), and in case the string contains non-ascii characters this conversion breaks. 

If I remove the conversion, then the unit test in py2 passes again. This should be a suitable solution for both py2 and py3. 

Only partially related: we should follow [futurize guide](https://python-future.org/compatible_idioms.html#file-io-with-open) to open files in a way that is compatible with both py2 and py3. [EDIT]

### DIFF
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -1,4 +1,7 @@
 from __future__ import (division, print_function)
+from builtins import str
+from past.builtins import basestring
+from future.utils import viewitems, listvalues, listitems
 
 import copy
 import socket
@@ -147,7 +150,7 @@ def _changeToFloat(value):
 
 def _validateTypeAndSetDefault(sourceDict, stepDefault):
     # check primitive time and remvoe if the values is composite type.
-    for key, value in sourceDict.items():
+    for key, value in listitems(sourceDict):  # ACHTUNG! dict size changes while iterating
         if key not in stepDefault and value in [[], {}, None, "None"]:
             del sourceDict[key]
 
@@ -158,11 +161,11 @@ def _validateTypeAndSetDefault(sourceDict, stepDefault):
 
 
 def changePerformanceStruct(perfDict):
-    return [{"pName": prop, "value": _changeToFloat(value)} for prop, value in perfDict.items()]
+    return [{"pName": prop, "value": _changeToFloat(value)} for prop, value in viewitems(perfDict)]
 
 
 def changeToList(aDict):
-    return [{"prop": prop, "value": value} for prop, value in aDict.items()]
+    return [{"prop": prop, "value": value} for prop, value in viewitems(aDict)]
 
 
 def convertInput(inputList):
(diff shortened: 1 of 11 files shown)


src/python/WMCore/Services/WMArchive/DataMap.py
```diff
@@ -219,10 +222,10 @@ def convertOutput(outputList):
                 del outDict[field]
 
         for field in WMARCHIVE_CONVERT_TO_LIST:
-            if field in outDict and isinstance(outDict[field], basestring):
+            if field in outDict and (isinstance(outDict[field], str) or isinstance(outDict[field], bytes)):
```
amaltaro, line 225: Same comment here, you can compare against a tuple of types: isinstance(outDict[field], (str, bytes))

amaltaro (changes requested):
These changes look good, Dario. There is just one minor change that can be made to the code, could you please update it? Thanks

src/python/WMCore/Services/WMArchive/DataMap.py
mapellidario: Awlays correct! I just pushed a new commit!

amaltaro:
Thanks Dario! Can you please squash commits 1 and 3? Then it's good to go as well.

amaltaro (approved):

PR #10168, opened 2020-12-10
[py2py3] Apply changes from `pylint --py3k` to files involved in #9818

### LINKED ISSUE #10167: [py2py3] Apply changes from `pylint --py3k` to files involved in #9818
**Impact of the new feature**

Files involved in #9818 

#9818 was not checked against `pylint --py3k`. Its report is

```
************* Module src.python.WMCore.WMException
W:117,12: Indexing exceptions will not work on Python 3 (indexing-exception)
```

We should then modernize `WMCore/WMException` following [pylint doc](https://pylint.pycqa.org/en/latest/technical_reference/features.html):

> Indexing exceptions will not work on Python 3 Indexing exceptions will not work on Python 3. Use exception.args[index] instead.

Simply change the way that the exception arguments are accessed.

There is not much we ccan do if we want to run the same code in python3

**Additional context**

Follow-up to #9818

mapellidario:
Fixes #10167 

#### Status

ready

#### Related PRs

[EDIT]: since we are pushing #10187 forward, I would set this PR aside and close this without merging when we merge #10187

This PR should be merged **after** #10174 

This is a follow-up of #9818

#### Description

#9818 was not checked against pylint --py3k, whose report on the files changed by that PR is

```
************* Module src.python.WMCore.WMException
W:117,12: Indexing exceptions will not work on Python 3 (indexing-exception)
```

We should then modernize WMCore/WMException following pylint doc:

>    Indexing exceptions will not work on Python 3 Indexing exceptions will not work on Python 3. Use exception.args[index] instead.

yes

This does not require any change in external dependencies.

### DIFF
--- a/test/python/WMCore_t/WMException_py2py3_t.py
+++ b/test/python/WMCore_t/WMException_py2py3_t.py
@@ -72,7 +72,9 @@ class WMExceptionTest(unittest.TestCase):
         data['key1'] = 'value1'
         data['key2'] = 3.14159
         exception.addInfo(**data)
+        exception.addInfo(Protocol="myprotocol", LFN="lfn", TargetPFN="pfn")
         self.logger.debug("String version of exception: %s", str(exception))
+        self.logger.debug("XML version of exception: %s", exception.xml())
 
     def testExceptionUnicode0(self):
         """


mapellidario:
Hello @amaltaro !

The PR now contains a change that makes `pylint --py3k` happy.

### Why `pylint --py3k` complains in the first place

In python2 it was possible to access to exception arguments directly with

```python
catch Exception as e:
  print e.message
  if len(e >= 0):
    print "first argument:", e[0]
    print "other arguments:", e[1:]
```

This was changed in python3 and it now is ( [following the docs](https://docs.python.org/3/tutorial/errors.html#handling-exceptions) )

```python3
catch Exception as e:
  print(e.message)
  if len(e.args >= 0):
    print("first argument:", e.args[0])
    print("other arguments:", e.args[1:])
```

`pylint --py3k` is pretty clever and it recognizes that at this 
line https://github.com/dmwm/WMCore/blob/875c3c5a48d0d82df947cdf13466c469c33ca27a/src/python/WMCore/WMException.py#L117 
we are accessing an exception argument in the old fashion way, since `WMException` inherits from `Exception`. It then suggests

```python
-  self[key] = value
+  self.args[key] = value
```

### Why `pylint --py3k` suggestion is not the right fit for us

`pylint --py3k` is clever, but it did not realize that we added a getter and a setter 
to our `WMException` class:

https://github.com/dmwm/WMCore/blob/875c3c5a48d0d82df947cdf13466c469c33ca27a/src/python/WMCore/WMException.py#L103

This means that `self[key] = value` does not actually change exception arguments in out case, but it has been redefined to change `WMException.data` dictionary. So pylint warning is a false positive.

### My workaround

In order to remove the warning, we can directly and more explicitely set `WMException.data` dictionary in `WMException.addInfo()`, instead of using `WMException.__setitem__()` as a proxy method. This is what the change in this PR does.

If you want to check that the resulting exception is the same, you can add `raise` at the end of the tests in `test/python/WMCore_t/WMException_t.py` so that it prints the logs with the exception message.

The ultimate question is: Is this change necessary? No, it is not. We can simply ignore this warning from pylint and python3 code will not break (hopefully, I could not check this statement)

What do you think? Would you prefer not changing anything and ignore pylint suggestion or apply this change?

amaltaro:
@mapellidario is this PR still relevant?

mapellidario:
I would delete this after we merge #10187 which supersedes this, if this is fine with you

mapellidario:
I rebased this PR on master and, after fixing a small conflict, the only change introduced by this PR disappeared, since it was already in #10187. I close this PR. 
Thanks Alan for the feedback and the housekeeping!

Issue #9852, opened 2020-07-16
JSONThunker - strange dictionary key type comparison


mapellidario:
Rather than a bug, this is something strange that we stumbled upon during the `WMCore.Cache` migration, PR #9818.

**Impact of the bug**

This may affect

* `src/python/WMCore/Services/Requests.py`
* `src/python/WMCore/WebTools/Page.py`
* `src/python/WMCore/WebTools/RESTFormatter.py`

**Describe the bug**

In `src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.JSONThunker.handleDictThunk`, there is the following for loop over the function argument `toThunk` (some lines have been redacted, the original version of the function is available [here](https://github.com/dmwm/WMCore/blob/7584caed5b7a85a03d4008586a0a667346f91d55/src/python/WMCore/Wrappers/JsonWrapper/JSONThunker.py#L149)):

```python
def handleDictThunk(self, toThunk):
    toThunk = self.checkRecursion(toThunk)
    tmpdict = {}
    for k, v in iteritems(toThunk):
        if type(k) == type(int): # this is one of the incriminated lines
            tmpdict['_i:%s' % k] = self._thunk(v)
        elif type(k) == type(float): # this is the other incriminated line
            tmpdict['_f:%s' % k] = self._thunk(v)
        else:
            tmpdict[k] = self._thunk(v)
     [...]
```

Both in python2 and python3, `type(int)` and `type(float)` is simply `type`, so that these checks are passed only in case the key `k` of the argument `toThunk` is of type `type`.

However, peeking inside what would happen if the condition were true, it seems that the type of the key is being used to add a prefix to the key itself in `tmpdict`. It is noticed that when the type is `type(int)` the prefix is `_i:` and when is `type(float)` is it `_f:`. This suggests that maybe the intentions were to write

```python
 def handleDictThunk(self, toThunk):
     toThunk = self.checkRecursion(toThunk)
     tmpdict = {}
     for k, v in iteritems(toThunk):
-        if type(k) == type(int): 
+        if type(k) is int: 
             tmpdict['_i:%s' % k] = self._thunk(v)
-        elif type(k) == type(float): 
+        elif type(k) is float: 
             tmpdict['_f:%s' % k] = self._thunk(v)
         else:
             tmpdict[k] = self._thunk(v)
      [...]
```

**Additional context and error message**

[...]

PR #10172, opened 2020-12-11
[py2py3] Fix pre-python2.6 idiom: use "isinstance()" instead of "type() == " 

### LINKED ISSUE #10171: [py2py3] Fix pre-python2.6 idiom: use "isinstance()" instead of "type() == "
**Impact of the new feature**

Likely to involve a large portion of the codebase.

The "python3 compatibility report" section of the Jenkins report often suggests the following change, in order not to use a pre-py2.6 idiom:

```python
- if type(variable) == mytype:
+ if isinstance(variable, mytype):
```

Since these are almost always safe changes, they can be tackled all at once.

We should apply the changes suggested by the "python3 compatibility report" regarding pre-py2.6 idioms once and for all.

**Additional context**

This issue was suggested in the comments section of #10103

mapellidario:
Fixes #10171 

#### Status

Ready

#### Related PR

This PR is independent form other PRs

#### Description

Run `futurize -f idioms` on `src/python`.

Manual changes

* `type([])` changed to `list`
* `type(1)` changed to `int`
* `type.ListType` changed to `list`

Changes involving `type(dict)`, which can cause problems (despite no problem has been caught by jenkins CI)
* `src/python/WMCore/WMSpec/Steps/BuildTools.py`
* `src/python/WMCore/WMSpec/ConfigSectionTree.py`

##### What has been kept back and not included in this PR

* `JSONThunker` had some problems in the past with this fixers. Moreover, #9852 is still open.
* `WMException`: this change caused a dramatic change in behavior that is being addressed in separate PRs, such as #10174, #10187, #10189

Yes, it should be. any possibly breaking changes will be reported here.

This does not rely on third party libraries

### DIFF
--- a/src/python/WMCore/WMRuntime/ProcessMonitor.py
+++ b/src/python/WMCore/WMRuntime/ProcessMonitor.py
@@ -123,7 +123,7 @@ class ExecProcess(ChildProcess):
             Either accepts a list of arguments which are passed to execvp
             OR     accepts a string which is passed to bash and shell-expanded
         """
-        if (type(arguments) != type.ListType):
+        if not isinstance(arguments, list):
             # we got passed a string, pass it to a shell
             self.args[0] = 'bash'
             self.args[1] = '-c'
@@ -145,9 +145,8 @@ class PythonProcess(ChildProcess):
     def __init__(self):
         self.target = None
 
-    def setTarget(self,newtarget):
-        if ((type(newtarget) != FunctionType) and
-            (type(newtarget) != LambdaType)):
+    def setTarget(self, newtarget):
+        if not isinstance(newtarget, (FunctionType, LambdaType)):
             raise RuntimeError("PythonProcess requires a function for target")
 
         self.target = newtarget
--- a/src/python/WMCore/WMSpec/ConfigSectionTree.py
+++ b/src/python/WMCore/WMSpec/ConfigSectionTree.py
@@ -227,7 +227,7 @@ def format(value):
     format a value as python
     keep parameters simple, trust python...
     """
-    if type(value) == str:
+    if isinstance(value, str):
         value = "\'%s\'" % value
     return str(value)
 
@@ -238,14 +238,14 @@ def formatNative(value):
     Like the format function, but allowing passing of ints, floats, etc.
     """
 
-    if type(value) == int:
+    if isinstance(value, int):
         return value
-    if type(value) == float:
+    if isinstance(value, float):
         return value
-    if type(value) == list:
+    if isinstance(value, list):
         return value
-    if type(value) == dict:
-        return dict
+    if isinstance(value, dict):
+        return value
     else:
         return format(value)
 
@@ -398,7 +398,7 @@ class TreeHelper(object):
         adds an arbitrary value as a dictionary.  Can have multiple values
         """
 
-        if not type(value) == dict:
+        if not isinstance(value, dict):
             raise Exception("TreeHelper.addValue passed a value that was not a dictionary")
 
         for key in value:
--- a/src/python/WMCore/WebTools/WebAPI.py
+++ b/src/python/WMCore/WebTools/WebAPI.py
@@ -73,7 +73,7 @@ class WebAPI(DatabasePage):
 #            dict = self.runMethod(args[0], kwargs)
 #            return dict
             res = self.runMethod(args[0], kwargs)
-            if  type(res) is str:
+            if  isinstance(res, str):
                 res = eval(res)
             return res
         else:
@@ -122,7 +122,7 @@ class WebAPI(DatabasePage):
         """
         try:
             res = self.runMethod(args[0], kwargs)
-            if  type(res) is str:
+            if  isinstance(res, str):
                 res = eval(res)
             return res
         except:
@@ -138,7 +138,7 @@ class WebAPI(DatabasePage):
 #            dict = self.runMethod(args[0], kwargs)
 #            return dict
             res = self.runMethod(args[0], kwargs)
-            if  type(res) is str:
+            if  isinstance(res, str):
                 res = eval(res)
             return res
         else:
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetBlock.py
@@ -18,7 +18,7 @@ class SetBlock(DBFormatter):
 
         """
 
-        if type(lfn) == list:
+        if isinstance(lfn, list):
             binds = []
             for entry in lfn:
                 binds.append({'block': blockName, 'filelfn': entry})
--- a/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetStatus.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/DBSBufferFiles/SetStatus.py
@@ -14,7 +14,7 @@ class SetStatus(DBFormatter):
     sql = "UPDATE dbsbuffer_file SET status = :status WHERE lfn = :lfn"
 
     def execute(self, lfns, status, conn = None, transaction = None):
-        if type(lfns) != list:
+        if not isinstance(lfns, list):
             lfns = [lfns]
 
         bindVars = []
--- a/src/python/WMCore/BossAir/MySQL/NewState.py
+++ b/src/python/WMCore/BossAir/MySQL/NewState.py
@@ -28,7 +28,7 @@ class NewState(DBFormatter):
         """
 
         binds = []
-        if type(states) == str:
+        if isinstance(states, str):
             binds = {'name': states}
         for state in states:
             binds.append({'name': state})
--- a/src/python/WMCore/Services/PhEDEx/DataStructs/PhEDExDeletion.py
+++ b/src/python/WMCore/Services/PhEDEx/DataStructs/PhEDExDeletion.py
@@ -22,9 +22,9 @@ class PhEDExDeletion(object):
         """
         Initialize PhEDEx deletion with default values
         """
-        if type(datasetPathList) == str:
+        if isinstance(datasetPathList, str):
             datasetPathList = [datasetPathList]
-        if type(nodeList) == str:
+        if isinstance(nodeList, str):
             nodeList = [nodeList]
 
         self.datasetPaths = set(datasetPathList)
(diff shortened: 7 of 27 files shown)


mapellidario:
After private discussion with Alan, we decided that `WMException` would be left out of this PR, since we are dealing with it in  #10174, #10187, #10189. (The PR opening comment has been updated accordingly)

src/python/WMCore/WMRuntime/ProcessMonitor.py
```diff
@@ -146,8 +146,8 @@ def __init__(self):
         self.target = None
 
     def setTarget(self,newtarget):
-        if ((type(newtarget) != FunctionType) and
-            (type(newtarget) != LambdaType)):
+        if ((not isinstance(newtarget, FunctionType)) and
```
amaltaro, line 149: Let me suggest something much cleaner here:
```suggestion
        if not isinstance(newtarget, (FunctionType, LambdaType)):
```

src/python/WMCore/WMSpec/ConfigSectionTree.py
```diff
@@ -227,7 +227,7 @@ def format(value):
     format a value as python
     keep parameters simple, trust python...
     """
-    if type(value) == str:
+    if isinstance(value, str):
```
amaltaro, line 230: Won't we get ourselves into big trouble if value is unicode? I mean, don't we need to have future imports here?

src/python/WMCore/WebTools/WebAPI.py
```diff
@@ -73,7 +73,7 @@ def json(self, *args, **kwargs):
 #            dict = self.runMethod(args[0], kwargs)
 #            return dict
             res = self.runMethod(args[0], kwargs)
-            if  type(res) is str:
+            if  isinstance(res, str):
```
amaltaro, line 76: Same comment here. What if `res` is unicode?

amaltaro (changes requested):
Dario, this PR looks pretty good to me. I have requested only one change, and I have a couple of concerns along the code as well.

src/python/WMCore/WMRuntime/ProcessMonitor.py
mapellidario: I like this very much! Changed!

src/python/WMCore/WMSpec/ConfigSectionTree.py
mapellidario: What makes this PR contain trivial changes is that `type(value) == str` and `isinstance(value, str)` are supposed to give the same result, whatever is the type of `value`. 

We already have https://github.com/dmwm/WMCore/blob/979f610f0e62b61bcdc4cea68c2b2726a6d26d5c/src/python/WMCore/WMSpec/ConfigSectionTree.py#L13
so that in this line `str` is already a sequence of unicode codepoints. 

But there is no such thing as a "trivial" change and quirks are hidden everywhere. If value is a sequence of unicode codepoints, then in py3 it is of type `str`, so that both `type(value) == str` and `isinstance(value, str)` will likely return the same result in very situation. 
Meanwhile, in py2 it can get a bit confusing. This few lines summarize what could happen

```plaintext
Python 2.7.13 (default, Sep 14 2017, 23:43:58) 
[GCC 4.9.2] on linux2
>>> a = u'ciao'
>>> type(a)
<type 'unicode'>
>>> from builtins import str // str becomes <class 'future.types.newstr.newstr'>
>>> type(a) == str
False
>>> isinstance(a, str)
True
```

It is possible that `type(a) == str` incorrectly returns `False` when `a` is a string of type unicode codepoints sequence.

Then, you can wonder "what if value is a standard py2 string of type `str`?" Well, I just realized we have been a bit imprecise in #10103. Before adding `from builtins import str`, L230 would have matched standard py2 `str` and added `'` around them (unicode strings were not changed). After adding it, only the new type of py2 `str` introduced by future would get this (unicode strings would be changed, bytes strings would not). We changed the behavior of our code without noticing. Impact of this change: the only place where this change could have affected us is in https://github.com/dmwm/WMCore/blob/979f610f0e62b61bcdc4cea68c2b2726a6d26d5c/src/python/WMCore/WMSpec/WMTask.py#L631 , but `WMTask_t` has no unit tests for it. If we want to be sure this did not cause problems we would have to continue following where this function is used downstream (such as checking `./src/python/WMCore/JobSplitting/Generators/GeneratorFactory.py`).
We have to options here: remove `from builtins import str` and continue formatting `str` in both py2 and py3, despite these are two different things, or properly format both unicode and bytes strings in both py2 and py3. 

On a sidenote: Doesn't the line https://github.com/dmwm/WMCore/blob/979f610f0e62b61bcdc4cea68c2b2726a6d26d5c/src/python/WMCore/WMSpec/ConfigSectionTree.py#L248 feel wrong, does it?

src/python/WMCore/WebTools/WebAPI.py
mapellidario: Here the situation is a bit easier. We do not import `from builints import str`, so `unicode` would not be matched in py2 both with `type(res) is str` and `isinstance(res, str)`. This change does not change behaviour

However, `src/python/WMCore/WebTools/WebAPI.py` is planned to be modernized in slice 7 and we will have to investigate how to modernize this properly. So this is a problem for (not so remote) future us!

src/python/WMCore/WMSpec/ConfigSectionTree.py
amaltaro: > We have to options here: remove from builtins import str and continue formatting str in both py2 and py3, despite these are two different things, or properly format both unicode and bytes strings in both py2 and py3.

I'm not sure what's the best way to proceed here, and I would like to avoid pulling actions from other slices as much as we can (likely properly dealing with unicode and byte strings here, as you suggested at the very end).
I guess a third option would be not to change anything else and see how tests go(?) 

> On a sidenote: Doesn't the line 

It does! I think we should change it to `return value` to keep it consistent with lists, for instance. However, I don't think we have ever had such data types there, otherwise things would have likely had broken already.

src/python/WMCore/WebTools/WebAPI.py
amaltaro: Gotcha!

amaltaro (commented):
Dario, thanks for the follow up.
It's isn't clear to me whether further changes should be made here. Please let me know if you consider it ready to go or if you want to make further changes to the ConfigSectionTree module. Thanks

src/python/WMCore/WMSpec/ConfigSectionTree.py
mapellidario: >  I would like to avoid pulling actions from other slices as much as we can

I totally agree. What I propose is that I open a new issue about "possible bug introduced in #10103 " ans I put it in python3 project in the column "WMCore py2->py2py3: bugs and follow up". I can spend a couple of hours to see what the impact of that bug could be and then we try to make the bug happen when we cut the test deploy.

> However, I don't think we have ever had such data types there, otherwise things would have likely had broken already.

yeah, it's been sitting there since 11 years! I like your suggestion of changing it eventually!

amaltaro: Having an extra GH issue with a possible problem is a good idea, indeed.

mapellidario:
I know I have not been too clear in answering to your comment

### Problem in ConfigSectionTree

Before #10103 we had

```python
def formatNative(value):
    if isinstance(value, int):
        return value
    if isinstance(value, float):
        return value
    if isinstance(value, list):
        return value
    if isinstance(value, dict):
        return value # this was actually "return dict", but it is outside of our scope now
    else:
        return format(value)

def format(value):
    if isinstance(value, str):
        value = "\'%s\'" % value
    return str(value)
```

Since we run this code in py2 only, only py2-`str` would get surrounded in `\'`, not unicode strings, and the return type of this function would be py2-`str` (i.e. bytes string)

In #10103 we introduced `from builtins import str`. This means that right now, when we run the code in python2, py2-`str` are not surrounded by `\'`, instead unicode strings are, and the return type of the function is unicode, not bytes.

### Implications

Is this change a real problem? I am not sure. `formatNative` seems to be used only in `WMCore/WMSpec/WMTask.WMTaskHelper.getGeneratorSettings` in

https://github.com/dmwm/WMCore/blob/08c1eea1a3245ebf956438e0a1ab1b79e338ff40/src/python/WMCore/WMSpec/WMTask.py#L631

Which is then used in `WMCore/JobSplitting/Generators/GeneratorManager.GeneratorManager` and `WMCore/JobSplitting/Generators/GeneratorFacotory.GeneratorFactory.makeGenerators`.

### How to proceed

We have a couple of options and I am not sure which is the best, since it is a bit hard for me to graps how the "pythonisedDict" is used in the Generators.

In any case, I can create a new issue and we move this discussion there, so that it is not lost.

#### 1. Leave it as is (surround unicode and return unicode)

and investigate in testbed the impact of this problem.

(This may not be the best option)

#### 2. Surround both unicode and bytes, return unicode

We could change 

```python
- from builtins import str, object
+ from builtins import str, bytes, object

def format(value):
-    if isinstance(value, str):
+    if isinstance(value, (str, bytes)):
        value = "\'%s\'" % value
+    return str(value)
```

(This is a nice option)

#### 3. Make `ConfigSectionTree` behave the same as before #10103 (surround bytes and return bytes)

We could change 

```python
- from builtins import str, object
+ from builtins import str, bytes, object

def format(value):
-    if isinstance(value, str):
+    if isinstance(value, bytes):
        value = "\'%s\'" % value
-    return str(value)
+    return bytes(value)
```

This may not be what we want, though, especially in python3.

#### 4. Try the "native string" approach

This is inspired by the Fedora's Conservative Py3 Porting Guide https://portingguide.readthedocs.io/en/latest/strings.html#the-native-string

The idea is: often, when we use `str` in py2, we are just consistently using `str` and we may be better off to just forget about unicode,bytes and other amenities and just keep using `str` in both py2 and py3, even though it means different things.

when run from py2 interpreter:
- surround bytes, return bytes
when run from py3 interpreter:
- surround unicode, return unicode

(This is a nice option)

src/python/WMCore/WMSpec/ConfigSectionTree.py
mapellidario: (I made a new commit with `return dict` -> `return value`)

amaltaro (changes requested):
Thanks Dario. We better discuss that possible issue in the new issue you created. You further commits look good to me, can you please squash them in the first one? Thanks

amaltaro (approved):

PR #9868, opened 2020-08-12
Improve py2 compatibility of WMException __init__

mapellidario:
Fixes #9862 

#### Status
ready

#### Description

Small change to `WMException.__init__` in order to improve its handling of unicode error messages in python2

#### Is it backward compatible (if not, which system it affects?)
yes

#### External dependencies / deployment changes
This makes WMException depend on python future

PR #10174, opened 2020-12-12
 [py2py3] Fix potential bug in WMException

### LINKED ISSUE #10173: [py2py3] Fix potential bug in WMException
**Impact of the bug**

**Describe the bug**

The line 

```plaintext
```

**How to reproduce it**

**Expected behavior**

mapellidario:
Fixes #10173 

#### Status

In development. Solution to current problem found, but more investigation and more unit tests are necessary.

#### Description

`WMException` 

* implementing unicode sandwich: convert to unicode all the incoming data, such as `message` and `data`.
* `pylint --py3k` warning addressed in #10168 (next PR to review ;) )
* "fixes for newer python3 idioms" addressed in #10172 (the Pr that should be review after #10168)

`WMException_t` 

* `from builtins import str, bytes`
* Tries to instantiate the exception object with as much unicode characters as possible, in exception message and data (both in keys and values)

As always, it is intended to be, but it impossible to guarantee that these chagnes are perfect. We will list here what we think can cause problems.

#### Related PRs

This work was initially inside #10172 , but has been split from because

* #10172 is intended to contain straightforward changes (pre-py2.6 idioms, from futurize stage 1) on  broad range of file
* this PR involves strings (not a straightforward change)
* #10173 is direct consequence of some negligence in #9868 , which has been discovered by chance while working on #10172 

### DIFF
--- a/src/python/WMCore/WMException.py
+++ b/src/python/WMCore/WMException.py
@@ -6,7 +6,7 @@ General Exception class for WM modules
 
 """
 
-from builtins import bytes
+from builtins import str, bytes
 from future.utils import viewitems
 
 import inspect
@@ -29,9 +29,11 @@ class WMException(Exception):
 
     def __init__(self, message, errorNo=None, **data):
         self.name = str(self.__class__.__name__)
-        if type(message) == bytes:
+        if isinstance(message, bytes):
             # Fix for the unicode encoding issue, see #8056 and #8403
             # interprets this string using utf-8 codec and ignoring any errors
+            # unicode sandwich: convert sequence of bytes strings to
+            # unicode codepoints strings as soon as possible
             message = message.decode('utf-8', 'ignore')
 
         Exception.__init__(self, self.name, message)
@@ -52,7 +54,7 @@ class WMException(Exception):
             self.data.setdefault("ErrorNr", errorNo)
 
         self._message = message
-        self.data.update(data)
+        self.addInfo(**data)
 
         #  //
         # // Automatically determine the module name
@@ -62,7 +64,7 @@ class WMException(Exception):
                 frame = inspect.currentframe()
                 lastframe = inspect.getouterframes(frame)[1][0]
                 excepModule = inspect.getmodule(lastframe)
-                if excepModule != None:
+                if excepModule is not None:
                     modName = excepModule.__name__
                     self.data['ModuleName'] = modName
             finally:
@@ -83,7 +85,7 @@ class WMException(Exception):
         # // ClassName if ClassInstance is passed
         # //
         try:
-            if self.data['ClassInstance'] != None:
+            if self.data['ClassInstance'] is not None:
                 self.data['ClassName'] = self.data['ClassInstance'].__class__.__name__
         except Exception:
             pass
@@ -104,6 +106,12 @@ class WMException(Exception):
         """
         make exception look like a dictionary
         """
+        # unicode sandwich: convert sequence of bytes strings to
+        # unicode codepoints strings as soon as possible
+        if isinstance(key, bytes):
+            key = key.decode("utf-8", "ignore")
+        if isinstance(value, bytes):
+            value = value.decode("utf-8", "ignore")
         self.data[key] = value
 
     def addInfo(self, **data):
@@ -114,6 +122,13 @@ class WMException(Exception):
         exception instance
         """
         for key, value in viewitems(data):
+            # unicode sandwich: convert sequence of bytes strings to
+            # unicode codepoints strings as soon as possible
+            # assumption: value is not iterable (list, dict, tuple, ...)
+            if isinstance(key, bytes):
+                key = key.decode("utf-8", "ignore")
+            if isinstance(value, bytes):
+                value = value.decode("utf-8", "ignore")
             self[key] = value
         return
 
@@ -153,7 +168,7 @@ class WMException(Exception):
         strg += self.traceback
         strg += '\n'
         strg += WMEXCEPTION_END_STR
-        if type(strg) == bytes:
+        if isinstance(strg, bytes):
             # Fix for the unicode encoding issue, #8043
             strg = strg.decode('utf-8', 'ignore')
         return strg
--- a/test/python/WMCore_t/WMException_t.py
+++ b/test/python/WMCore_t/WMException_t.py
@@ -7,6 +7,8 @@ General test for WMException
 
 """
 from __future__ import print_function, division
+from builtins import str, bytes
+
 import logging
 import unittest
 from WMCore.WMException import WMException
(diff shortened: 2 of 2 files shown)


src/python/WMCore/WMException.py
```diff
@@ -114,6 +116,13 @@ def addInfo(self, **data):
         exception instance
         """
         for key, value in viewitems(data):
+            # unicode sandwich: convert sequence of bytes strings to
+            # unicode codepoints strings as soon as possible
+            # assumption: value is not iterable (list, dict, tuple, ...)
```
mapellidario, line 121: @amaltaro is this assumption reasonable?

What I mean is that if `data` is like 

```python
data = {
key: 3.14,
key1: "string"
}
```
then we are ok, but if it can be
```python
data = {
top_key: 3.14,
top_key1: {
  nested_key : "string"
}
}
```

then we can incur in the same bug. The solution would be changing `self.addInfo()` so that it is recursive and can traverse a generic nested dictionary looking for bytes strings and converting them to unicode strings. This would add some complexity and it is why I preferred not to go down this path without checking with you before

amaltaro: I suspect `data.values()` will always be a string, but I'm not 100% sure. 
I guess there is no need to make it recursive, if we happen to have another issue like this, then we know where implementation is required.

src/python/WMCore/WMException.py
```diff
@@ -114,6 +116,13 @@ def addInfo(self, **data):
...
+            if isinstance(key, bytes):
+                key = key.decode("utf-8")
```
amaltaro, line 123: Should we add the `ignore` option here (and below)? Just so we can be sure that non-utf-8 compliant data won't break this?

test/python/WMCore_t/WMException_t.py
```diff
@@ -21,13 +23,27 @@ def setUp(self):
...
         self.logger = logging.getLogger('WMExceptionTest')
 
+        self.test_data = {
+            'key-ascii': 'value-1',       # unicode (ascii): unicode (ascii)
+            'key-ascii-b': '√@ʟυℯ-1',     # unicode (ascii): unicode (non-ascii)
+            'key-ascii-c': u'√@ʟυℯ-1',    # unicode (ascii): unicode (non-ascii)
+            'key-ascii-d': bytes('√@ʟυℯ-1', 'utf-8'),    # unicode (ascii): bytes (of non-ascii)
+            'ḱℯƴ-unicode-a': 'ṽ@łυ℮-2',   # unicode (non-ascii): unicode (non-ascii)
+            u'ḱℯƴ-unicode-b': 'ṽ@łυ℮-2',  # unicode (non-ascii): unicode (non-ascii)
+            bytes('ḱℯƴ-unicode-c', 'utf-8'): 'ṽ@łυ℮-2',  # bytes (of non-ascii): unicode (non-ascii)
```
amaltaro, line 40: Can you add another test case where either the key or value (or both) is: `'\x95'` :-D

amaltaro (changes requested):
Dario, it's looking pretty good. However, I still think this code can crash and I made a suggestion along the code.

test/python/WMCore_t/WMException_t.py
mapellidario: Aha, I get your point. '\x95' does not seem to be a valid unicode-encoded sequence of bytes, so you hope to break the decodings with this and force me to add the "ignore" argument. Nice, thanks!

mapellidario:
Alan suggested to insert malformed bytes into the test data, for example including byte '\x95' which is not a valid unicode-encoded byte.

With the test data

```python
        self.test_data = {
            'key-ascii': 'value-1',       # unicode (ascii): unicode (ascii)
            'key-ascii-b': '√@ʟυℯ-1',     # unicode (ascii): unicode (non-ascii)
            'key-ascii-c': u'√@ʟυℯ-1',    # unicode (ascii): unicode (non-ascii)
            'key-ascii-d': bytes('√@ʟυℯ-1', 'utf-8'),    # unicode (ascii): bytes (of non-ascii)
            'ḱℯƴ-unicode-a': 'ṽ@łυ℮-2',   # unicode (non-ascii): unicode (non-ascii)
            u'ḱℯƴ-unicode-b': 'ṽ@łυ℮-2',  # unicode (non-ascii): unicode (non-ascii)
            bytes('ḱℯƴ-unicode-c', 'utf-8'): 'ṽ@łυ℮-2',  # bytes (of non-ascii): unicode (non-ascii)
            'ḱℯƴ-unicode-d': 'value-\x95',  # unicode (of non-ascii): unicode (invalid byte)
            'key-\x95': 'ṽ@łυ℮-2',  # unicode (invalid byte): unicode (non-ascii)
            'key3': 3.14159,
            }
```

The string output is

```plaintext
        key-ascii-c : √@ʟυℯ-1
        ModuleName : WMCore_t.WMException_t
        MethodName : testExceptionUnicode0
        key- : ṽ@łυ℮-2 # IGNORED MALFORMED BYTE HERE
        key-ascii-d : √@ʟυℯ-1
        key3 : 3.14159
        ḱℯƴ-unicode-d : value- # IGNORED MALFORMED BYTE HERE
        ḱℯƴ-unicode-b : ṽ@łυ℮-2
        ḱℯƴ-unicode-c : ṽ@łυ℮-2
        ClassInstance : None
        FileName : /home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/WMException_t.py
        ClassName : None
        key-ascii-b : √@ʟυℯ-1
        ḱℯƴ-unicode-a : ṽ@łυ℮-2
        LineNumber : 77
        ErrorNr : 100
        key-ascii : value-1
```

The xml output is

```plaintext
[...]
<Key>
key-</Key>
<Value>
ṽ@łυ℮-2</Value>
[...]
<Key>
ḱℯƴ-unicode-d</Key>
<Value>
value-</Value>
```

src/python/WMCore/WMException.py
mapellidario: I see why you want to add this, I agree with you, thanks!

src/python/WMCore/WMException.py
```diff
@@ -114,6 +116,13 @@ def addInfo(self, **data):
...
         """
         for key, value in viewitems(data):
+            # unicode sandwich: convert sequence of bytes strings to
+            # unicode codepoints strings as soon as possible
+            # assumption: value is not iterable (list, dict, tuple, ...)
+            if isinstance(key, bytes):
+                key = key.decode("utf-8", "ignore")
+            if isinstance(value, bytes):
+                value = value.decode("utf-8", "ignore")
             self[key] = value
```
amaltaro, line 126:
```suggestion
            self.data[key] = value
```
And here we have a hard bug! I think the code I suggest here is the correct one.

amaltaro (changes requested):
Dario, I think there is one line that needs to be fixed. See comment along the code.

src/python/WMCore/WMException.py
mapellidario: You forsee the future, that is the topic of #10168

mapellidario: Here is my comment on this exact line: https://github.com/dmwm/WMCore/pull/10168#issuecomment-743386445

mapellidario: (We continued this conversation on slack. we are not including this change in this PR, we are postponing this to #10168 )

amaltaro:
Thanks Dario for all the offline discussion on this!

PR #10179, opened 2020-12-14
Decouple CMSSW/WMAgent python versions from PSet tweaking step

### LINKED ISSUE #10181: Perform PSet tweaking by using CMSSW external scripts
**Impact of the new feature**
WMAgent pset tweaking

To disentangle WMAgent from the CMSSW framework

Move methods that depend on CMSSW to CMSSW itself, make external calls instead.

khurtado:
Fixes #10181

#### Status
Ready for review

#### Description
This PR decouples CMSSW and WMCore python versions.
Short description:
This re-implements the pset tweaking step not to directly depend on the CMSSW framework (i.e.: importing FWCore), but rather makes external calls to scripts maintained in CMSSW in order to read and tweak the pset.

Long description: 
The Pset tweaking happens in the [SetupCMSSWPSet](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py) pre-script, which will now run outside the Scram environment, using its own comp python version, rather than the CMSSW python version. Modifications to the Pset (direct or through a json dictionary) are done via external calls to scripts living in cmssw-wm-tools (see below for github repository link), which do use the CMSSW python version inside the scram environment (python 2 or 3).

For this, WMTweak adds the parameters to be changed using the cms types (originally as strings to save it in the json file, then evaluated to the actual cms types in cms-wm-tools, for the tweaks).
Methods that were changing the PSet directly here, are now done via several scripts supported by cmssw in cmssw-wm-tools too.
Dummy PSet for unit tests was changed in order to fix some unit tests that were broken after these changes. Some of the checks in the unit tests were changed a bit because now we have limitations reading the final PSet (since we run SetupCMSSWPSet outside the CMSSW environment), so we can check on standard types (e.g.: integers), but not cms types directly.

List of changes:
- PSetTweak: 
Added a "simplejson" format, which returns the json document in the following format:
```
{"process.maxEvents.input": 1200, "process.source.firstRun": 1, "process.source.firstLuminosityBlock": 59965}
```
- WMTweak:
Specifies cms types in the tweak json file, that are interpreted later by the cmssw-wm-tools scripts.
Additionally, the methods for creating the tweaks:  makeJobTweak, makeOutputTweak, makeTaskTweak now modify a tweak object, rather than returning a new one. 

- SetupCMSSWPSet
Calls external scripts in cmssw-wm-tools in order to modify the PSet.

- Scram
Make sure we don't pass WMAgent libraries when invoking CMSSW python.

- CMSSW Executor
Call the SetupCMSSWPset pre-script outside the scram environment (so, it's executed using the comp python version) instead.

Additional information here:
https://github.com/dmwm/WMCore/wiki/List-of-PSet-tweaks-applied-by-WMAgent-during-job-runtime

Yes

https://github.com/cms-sw/cmssw-wm-tools

### DIFF
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -527,10 +527,6 @@ def makeOutputTweak(outMod, job, result):
         lfn = "%s/%s/%s.root" % (lfnBase, lfnGroup(job), modName)
         result.addParameter("process.%s.logicalFileName" % modName, lfn)
 
-    # TODO: Nice standard way to meddle with the other parameters in the
-    #      output module based on the settings in the section
-    logging.info("mod result = {0}".format(result.jsondictionary()))
-
     return
 
 
--- a/src/python/PSetTweaks/WMTweak.py
+++ b/src/python/PSetTweaks/WMTweak.py
@@ -377,13 +377,12 @@ def decomposeConfigSection(csect):
     return decomposer.parameters
 
 
-def makeTaskTweak(stepSection):
+def makeTaskTweak(stepSection, result):
     """
     _makeTaskTweak_
 
     Create a tweak for options in the task that apply to all jobs.
     """
-    result = PSetTweak()
 
     # GlobalTag
     if hasattr(stepSection, "application"):
@@ -391,21 +390,20 @@ def makeTaskTweak(stepSection):
             if hasattr(stepSection.application.configuration, "pickledarguments"):
                 args = pickle.loads(stepSection.application.configuration.pickledarguments)
                 if 'globalTag' in args:
-                    result.addParameter("process.GlobalTag.globaltag", args['globalTag'])
+                    result.addParameter("process.GlobalTag.globaltag",  "customTypeCms.string('%s')" % args['globalTag'])
                 if 'globalTagTransaction' in args:
-                    result.addParameter("process.GlobalTag.DBParameters.transactionId", args['globalTagTransaction'])
+                    result.addParameter("process.GlobalTag.DBParameters.transactionId",  "customTypeCms.untracked.string('%s')" % args['globalTagTransaction'])
 
-    return result
+    return
 
 
-def makeJobTweak(job):
+def makeJobTweak(job, result):
     """
     _makeJobTweak_
 
     Convert information from a WMBS Job object into a PSetTweak
     that can be used to modify a CMSSW process.
     """
-    result = PSetTweak()
     baggage = job.getBaggage()
 
     # Check in the baggage if we are processing .lhe files
@@ -421,7 +419,7 @@ def makeJobTweak(job):
             if job['mask'].get('FirstLumi', None) != None:
                 logging.info("Setting 'firstLuminosityBlock' attr to: %s", job['mask']['FirstLumi'])
                 result.addParameter("process.source.firstLuminosityBlock",
-                                    job['mask']['FirstLumi'])
+                                    "customTypeCms.untracked.uint32(%s)" % job['mask']['FirstLumi'])
             else:
                 # We don't have lumi information in the mask, raise an exception
                 raise WMTweakMaskError(job['mask'],
@@ -435,16 +433,16 @@ def makeJobTweak(job):
     logging.info("Adding %d files to 'fileNames' attr", len(primaryFiles))
     logging.info("Adding %d files to 'secondaryFileNames' attr", len(secondaryFiles))
     if len(primaryFiles) > 0:
-        result.addParameter("process.source.fileNames", primaryFiles)
+        result.addParameter("process.source.fileNames", "customTypeCms.untracked.vstring(%s)" % primaryFiles)
         if len(secondaryFiles) > 0:
-            result.addParameter("process.source.secondaryFileNames", secondaryFiles)
+            result.addParameter("process.source.secondaryFileNames", "customTypeCms.untracked.vstring(%s)" % secondaryFiles)
     elif not lheInput:
         # First event parameter should be set from whatever the mask says,
         # That should have the added protection of not going over 2^32 - 1
         # If there is nothing in the mask, then we fallback to the counter method
         if job['mask'].get('FirstEvent', None) != None:
             logging.info("Setting 'firstEvent' attr to: %s", job['mask']['FirstEvent'])
-            result.addParameter("process.source.firstEvent", job['mask']['FirstEvent'])
+            result.addParameter("process.source.firstEvent", "customTypeCms.untracked.uint32(%s)" % job['mask']['FirstEvent'])
         else:
             # No first event information in the mask, raise and error
             raise WMTweakMaskError(job['mask'],
@@ -457,7 +455,7 @@ def makeJobTweak(job):
     if maxEvents is None:
         maxEvents = -1
     logging.info("Setting 'maxEvents.input' attr to: %s", maxEvents)
-    result.addParameter("process.maxEvents.input", maxEvents)
+    result.addParameter("process.maxEvents", "customTypeCms.untracked.PSet(input=cms.untracked.int32(%s))"% maxEvents)
 
     # We don't want to set skip events for MonteCarlo jobs which have
     # no input files.
@@ -465,18 +463,18 @@ def makeJobTweak(job):
     if firstEvent != None and firstEvent >= 0 and (len(primaryFiles) > 0 or lheInput):
         if lheInput:
             logging.info("Setting 'skipEvents' attr to: %s", firstEvent - 1)
-            result.addParameter("process.source.skipEvents", firstEvent - 1)
+            result.addParameter("process.source.skipEvents", "customTypeCms.untracked.uint32(%s)" % (firstEvent - 1))
         else:
             logging.info("Setting 'skipEvents' attr to: %s", firstEvent)
-            result.addParameter("process.source.skipEvents", firstEvent)
+            result.addParameter("process.source.skipEvents", "customTypeCms.untracked.uint32(%s)" % firstEvent)
 
     firstRun = mask['FirstRun']
     if firstRun != None:
-        result.addParameter("process.source.firstRun", firstRun)
+        result.addParameter("process.source.firstRun", "customTypeCms.untracked.uint32(%s)" % firstRun)
     elif not len(primaryFiles):
         # Then we have a MC job, we need to set firstRun to 1
         logging.debug("MCFakeFile initiated without job FirstRun - using one.")
-        result.addParameter("process.source.firstRun", 1)
+        result.addParameter("process.source.firstRun", "customTypeCms.untracked.uint32(1)")
 
     runs = mask.getRunAndLumis()
     lumisToProcess = []
@@ -490,7 +488,7 @@ def makeJobTweak(job):
 
     if len(lumisToProcess) > 0:
         logging.info("Adding %d run/lumis mask to 'lumisToProcess' attr", len(lumisToProcess))
-        result.addParameter("process.source.lumisToProcess", lumisToProcess)
+        result.addParameter("process.source.lumisToProcess", "customTypeCms.untracked.VLuminosityBlockRange(%s)" % lumisToProcess)
 
     # install any settings from the per job baggage
     procSection = getattr(baggage, "process", None)
@@ -499,21 +497,27 @@ def makeJobTweak(job):
 
     baggageParams = decomposeConfigSection(procSection)
     for k, v in viewitems(baggageParams):
+        if isinstance(v, str):
+            v =  "customTypeCms.untracked.string(%s)" % v
+        elif isinstance(v, int):
+            v =  "customTypeCms.untracked.uint32(%s)" % v
+        elif isinstance(v, list):
+            v =  "customTypeCms.untracked.vstring(%s)" % v
         result.addParameter(k, v)
 
-    return result
+    return
 
 
-def makeOutputTweak(outMod, job):
+def makeOutputTweak(outMod, job, result):
     """
     _makeOutputTweak_
 
     Make a PSetTweak for the output module and job instance provided
 
     """
-    result = PSetTweak()
     # output filenames
     modName = str(getattr(outMod, "_internal_name"))
+    logging.info("modName = %s", modName)
     fileName = "%s.root" % modName
 
     result.addParameter("process.%s.fileName" % modName, fileName)
@@ -525,8 +529,9 @@ def makeOutputTweak(outMod, job):
 
     # TODO: Nice standard way to meddle with the other parameters in the
     #      output module based on the settings in the section
+    logging.info("mod result = {0}".format(result.jsondictionary()))
 
-    return result
+    return
 
 
 def readAdValues(attrs, adname, castInt=False):
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -5,165 +5,65 @@ Create a CMSSW PSet suitable for running a WMAgent job.
 
 """
 from __future__ import print_function
-from future.utils import viewitems
+from builtins import next, object
 
 import json
 import logging
 import os
 import pickle
-import random
 import socket
-import re
-
-import FWCore.ParameterSet.Config as cms
 
 from PSetTweaks.PSetTweak import PSetTweak
-from PSetTweaks.WMTweak import applyTweak, makeJobTweak, makeOutputTweak, makeTaskTweak, resizeResources
+from PSetTweaks.WMTweak import  makeJobTweak, makeOutputTweak, makeTaskTweak, resizeResources
 from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig
 from WMCore.Storage.TrivialFileCatalog import TrivialFileCatalog
 from WMCore.WMRuntime.ScriptInterface import ScriptInterface
-from WMCore.WMRuntime.Tools.Scram import isCMSSWSupported, isEnforceGUIDInFileNameSupported
+from WMCore.WMRuntime.Tools.Scram import Scram
 
 
-def fixupGlobalTag(process):
+def factory(module, name):
     """
-    _fixupGlobalTag_
-
-    Make sure that the process has a GlobalTag.globaltag string.
-
-    Requires that the configuration already has a properly configured GlobalTag object.
+    _factory_
+    Function to return a dummy module name when a module
+    is not available
 
     """
-    if hasattr(process, "GlobalTag"):
-        if not hasattr(process.GlobalTag, "globaltag"):
-            process.GlobalTag.globaltag = cms.string("")
-    return
+
+    class DummyClass(object):
+        """
+        _DummyClass_
+        Dummy class to return when a cms class cannot be imported 
+
+        """
+        def __init__(self, module, name='', *args, **kwargs):
+            self.__module = module
+            self.__name = name
+            self.__d = dict()
+
+        def __setitem__(self, key, value):
+            self.__d[key] = value
+
+        def __getitem__(self, item):
+            return self.__d[item]
+
+        def __call__(self, *args, **kwargs):
+            pass
+
+        def __repr__(self):
+            return "{module}.{name}".format(module=self.__module, name=self.__name)
+
+    return DummyClass
 
 
-def fixupGlobalTagTransaction(process):
+class Unpickler(pickle.Unpickler):
     """
-    _fixupGlobalTagTransaction_
-
-    Make sure that the process has a GlobalTag.DBParameters.transactionId string.
-
-    Requires that the configuration already has a properly configured GlobalTag object
-
-    (used to customize conditions access for Tier0 express processing)
-
-    """
-    if hasattr(process, "GlobalTag"):
-        if not hasattr(process.GlobalTag.DBParameters, "transactionId"):
-            process.GlobalTag.DBParameters.transactionId = cms.untracked.string("")
-    return
-
-
-def fixupFirstRun(process):
-    """
-    _fixupFirstRun_
-
-    Make sure that the process has a firstRun parameter.
-
-    """
-    if not hasattr(process.source, "firstRun"):
-        process.source.firstRun = cms.untracked.uint32(0)
-    return
-
-
-def fixupLastRun(process):
-    """
-    _fixupLastRun_
-
-    Make sure that the process has a lastRun parameter.
+    _Unpickler_
+    Use this when loading a PSet pickle
+    and a specific module is not available (i.e.: FWCore modules)
 
     """
-    if not hasattr(process.source, "lastRun"):
-        process.source.lastRun = cms.untracked.uint32(0)
-    return
-
-
-def fixupLumisToProcess(process):
-    """
-    _fixupLumisToProcess_
-
-    Make sure that the process has a lumisToProcess parameter.
-
-    """
-    if not hasattr(process.source, "lumisToProcess"):
-        process.source.lumisToProcess = cms.untracked.VLuminosityBlockRange()
-    return
-
-
-def fixupSkipEvents(process):
-    """
-    _fixupSkipEvents_
-
-    Make sure that the process has a skip events parameter.
-
-    """
-    if not hasattr(process.source, "skipEvents"):
-        process.source.skipEvents = cms.untracked.uint32(0)
-    return
-
-
-def fixupFirstEvent(process):
-    """
-    _fixupFirstEvent_
-
-    Make sure that the process has a first event parameter.
-
-    """
-    if not hasattr(process.source, "firstEvent"):
-        process.source.firstEvent = cms.untracked.uint32(0)
-    return
-
-
-def fixupMaxEvents(process):
-    """
-    _fixupMaxEvents_
-
-    Make sure that the process has a max events parameter.
-
-    """
-    if not hasattr(process, "maxEvents"):
-        process.maxEvents = cms.untracked.PSet(input=cms.untracked.int32(-1))
-    if not hasattr(process.maxEvents, "input"):
-        process.maxEvents.input = cms.untracked.int32(-1)
-    return
-
-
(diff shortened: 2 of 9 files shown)


khurtado:
Test this please

khurtado:
Test this please

amaltaro:
test this please

amaltaro:
test this please

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -383,136 +360,53 @@ def handlePileup(self):
...
+        self.logger.info("Generating json for CMSSW pileup script")
+        try:
+            with open(newJsonPileupConfig, 'wb') as f:
+                json.dump(newPileupDict, f)
+        except Exception as ex:
+            self.logger.exception("Error writing out process filelist json:")
+            raise
+
+        procScript = "cmssw_handle_pileup.py"
+        cmd = "%s --input_pkl %s --output_pkl %s --pileup_dict %s" % (
```
amaltaro, line 398: I was going to suggest to skip this call if we do not have any files to be added as secondary files. However, it would break the current logic which is: always overwrite the list of secondary files.

With that said, do we need to provide a fix to the cmssw-tools for those cases we are providing an empty dictionary as input?

src/python/PSetTweaks/WMTweak.py
```diff
@@ -435,16 +433,16 @@ def makeJobTweak(job):
     logging.info("Adding %d files to 'fileNames' attr", len(primaryFiles))
     logging.info("Adding %d files to 'secondaryFileNames' attr", len(secondaryFiles))
     if len(primaryFiles) > 0:
-        result.addParameter("process.source.fileNames", primaryFiles)
+        result.addParameter("process.source.fileNames", "customTypeCms.untracked.vstring(%s)" % primaryFiles)
         if len(secondaryFiles) > 0:
```
amaltaro, line 437: This code - that was there even before your changes - is somehow surprising to me. I think this `if` statement for secondary files should be at the same upper level as the one for `primaryFiles`. Otherwise, what happens to a job without primary files but with pileup files? Or is it a case that is impossible to happen?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -14,242 +14,167 @@
 import socket
 import re
```
amaltaro, line 16: I'm not sure it can cause any problems, but I noticed the file permissions for this file changed from 0755 to 0644.

amaltaro (commented):
Kenyi, it's not a review, but just a spot check trying to debug that secondary files problem that you saw in one of those tests. I left a few comments along the code.

amaltaro:
@khurtado Kenyi, can you please fix the conflicts in this PR?

khurtado:
please test this

khurtado:
test this please

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: It should not be a problem because this is called by another python script

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -5,190 +5,119 @@
...
+
+        Create scram environment
+
+        """
+        scram = Scram(
+            version=self.getCmsswVersion(),
+            directory=self.stepSpace.location,
+            architecture=self.getScramVersion(),
+            initialise=self.step.data.application.setup.softwareEnvironment
+        )
```
amaltaro, line 99: Can we add an extra log info level here reporting the CMSSW version and ScramArch being used?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -5,190 +5,119 @@
...
+        return scram
+
+    def scramRun(self, cmdArgs):
+        """
+        _scramRun_
+
+        Run command inside scram environment
+
+        """
+        self.logger.info("ScramRun command args: %s", cmdArgs)
```
amaltaro, line 112: Should we change this log to something like `Running CMSSW WM tools with args:`? Unless this method can execute things different than cmssw-wm-tools scripts(?)
Similar change would apply in case it returns with a code != 0

amaltaro (commented):
Kenyi, changes are looking good in general, even though it's quite challenging to provide you a good review, especially on the SetupCMSSWPset with all those changes :)

I have a few general comment/questions/requests though:
* how about we create a log record for every fixer (cmssw-wm-tools) that we call?
* do you think there would be value in defining the set of wm-tools scripts called from this module? Either in a format of variable instances (self.edmPset....) or via a dictionary mapping a given key to the actual script name to be executed (e.g. edm_pset_pickler.py).
* similar to that `fixupDict`, do you think it would be useful to have a map (of at least what we know) of the pset attribute to the pset data type? Something like:
```
attrDataTypeDict = {'process.maxEvents.input', "customTypeCms.untracked.int32(1)", ...}
```
it could be done in a future PR as well, if useful.

There are likely required changes to make it compatible with python3, but I think it could be addressed in a different PR as well.

Sorry for a very non-precise review!

src/python/PSetTweaks/WMTweak.py
```diff
@@ -499,21 +497,27 @@ def makeJobTweak(job):
 
     baggageParams = decomposeConfigSection(procSection)
     for k, v in viewitems(baggageParams):
+        if type(v) == type(''):
```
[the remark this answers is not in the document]
khurtado: Yes! I thought I changed them for `isinstance` but I forgot to change them on WMTweak. I can take care of that.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -5,190 +5,119 @@
...
-    Make sure that the process has a firstRun parameter.
-
-    """
-    if not hasattr(process.source, "firstRun"):
-        process.source.firstRun = cms.untracked.uint32(0)
-    return
-
-
-def fixupLastRun(process):
+def factory(module, name):
```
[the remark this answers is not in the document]
khurtado: This is used with the customed `Unpickler` class. Basically, we can load the PSet without the CMSSW environment to read some key items, and this `factory` stuff will replace the cms objects with a dummy class object. So, we can see the whole PSet structure but can't read cms type values. I'm using this in the loadPSet() method

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: Technically, it can run any command inside the scram environment. Currently,  it  looks like this:

```
  2021-04-09 13:52:18,916:INFO:SetupCMSSWPset:ScramRun command args: edm_pset_pickler.py --input /tmpscratch/users/khurtado/work/validation/relVals/job/WMTaskSpace/cmsRun1/PSet.py --output_pkl /tmpscratch/users/khurtado/work/validation/relVals/job/WMTaskSpace/cmsRun1/PSet.pkl
```

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: Yes, good idea.  I can add that.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -5,190 +5,119 @@
...
-
-def fixupGlobalTagTransaction(process):
-    """
-    _fixupGlobalTagTransaction_
-
-    Make sure that the process has a GlobalTag.DBParameters.transactionId string.
+from WMCore.WMRuntime.Tools.Scram import Scram
 
-    Requires that the configuration already has a properly configured GlobalTag object
+FORMAT = '%(asctime)-15s %(clientip)s %(user)-8s %(message)s'
```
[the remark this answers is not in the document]
khurtado: Just check on this and they are not needed. I will remove these lines

amaltaro:
Thanks for rebasing and squashing those commits, @khurtado . Now I'm actually able to patch 1.4.7.patch2 with this PR. Applying it to a fresh installation in vocms0262 and running further tests.

amaltaro:
@khurtado we haven't yet run all the test workflows, but I see at least 3 of them with 100% failure rate in vocms0262 (alancc7-cloud2). Could you please have a look at them?

khurtado:
Yes, I will take a look. It seems an issue with the PYTHONPATH environment
when calling the preScript.

khurtado:
@amaltaro I just pushed a fix to the stepChain issue. Could you resubmit the workflows that were failing (except for the scl5 workflow, of course)

khurtado:
@amaltaro I fixed fixed 3 issues in total:
- StepChain workflows issue
- skipPileupEvents bug
- runLimits bug

These are affecting:
- All StepChain workflows
- amaltaro_TaskChain_PUMCRecyc_Agent147PSet_Val_210422_153202_9942
- amaltaro_TaskChain_LumiMask_multiRun_Agent147PSet_Val_210422_153149_640

When you get the chance, could you please patch the agent with these changes and resubmit these workflows?

amaltaro:
@khurtado Kenyi, thanks for providing these fixes. I have just applied your commits 3 to 5 to vocms0262, and injected all the `Integration` workflows. You can search for workflows matching `_Agent147PSet_Val_210426_`.

I'm going to clone and assign one of those 4 python3 workflows as well, just to be on the safe side.

khurtado:
Awesome! Thank you.

amaltaro:
@khurtado Kenyi, tests are all looking good now. Can I ask you to squash the last 3 commits into the first one please?

BTW, there are 2 or 3 workflows which didn't manage to run a single job since yesterday. You might want to have a look if there is anything we can do to get them into running in condor. Thanks

khurtado:
@amaltaro Done! W.r.t the workflows, one seems to have FNAL as the only site whitelisted and we know that takes like 3 days. But the other one is assigned to both CERN and FNAL, so I don't know why it's not running at CERN at this point.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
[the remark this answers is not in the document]
amaltaro: Gotcha, thanks!

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
amaltaro: I guess the answer is no, nothing else is needed, right?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -384,136 +370,56 @@ def handlePileup(self):
...
-        # going to be only two types of pileup input datasets: "data" or "mc"
-        # unlike all previous places where pileupType handled in a flexible
-        # way as specified in the configuration passed by the user, here are
-        # the two pileupTypes hardcoded: and we are going to add the "mc"
-        # datasets to "MixingModule"s and only add the "data" datasets to the
-        # "DataMixingModule"s
-
-        # if the user in the configuration specifies different pileup types
-        # than "data" or "mc", the following call will not modify anything
-        self._processPileupMixingModules(pileupDict, PhEDExNodeName, dataMixModules, "data")
```
amaltaro, line 405: Kenyi, do you know if in the new mode to tweak pileup no longer needs the previous distinction of MixingModule and DataMixingModule (mapped from our workflow attributes MCPileup and DataPileup)?

Honestly, I never understood those well enough. Maybe @klannon and/or @davidlange6 would have some input here(?)

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -384,136 +370,56 @@ def handlePileup(self):
...
-                    if requestedPileupType == 'data':
-                        if getattr(self.jobBag, 'skipPileupEvents', None) is not None:
-                            # For deterministic pileup, we want to shuffle the list the
-                            # same for every job in the task and skip events
-                            random.seed(self.job['task'])
-                            self.logger.info("Skipping %d pileup events for deterministic data mixing",
-                                             self.jobBag.skipPileupEvents)
-                            inputTypeAttrib.skipEvents = cms.untracked.uint32(
-                                int(self.jobBag.skipPileupEvents) % eventsAvailable)
-                            inputTypeAttrib.sequential = cms.untracked.bool(True)
```
amaltaro, line 466: I might have missed it in the code, is this `sequential` attribute covered in your PR?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -384,136 +370,56 @@ def handlePileup(self):
         siteConfig = loadSiteLocalConfig()
         PhEDExNodeName = siteConfig.localStageOut["phedex-node"]
         self.logger.info("Running on site '%s', local PNN: '%s'", siteConfig.siteName, PhEDExNodeName)
+        jsonPileupConfig = os.path.join(self.stepSpace.location, "pileupconf.json")
 
-        pileupDict = self._getPileupConfigFromJson()
-
-        # 2011-02-03 according to the most recent version of instructions, we do
```
amaltaro, line 390: If we don't drop the multiple pileup types, I think it would be useful to keep some of these docstrings around.
Same comment for a few other docstring blocks below (like the one about AAA, etc).

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -384,136 +370,56 @@ def handlePileup(self):
...
-                    # Shuffle according to the seed above or randomly
-                    random.shuffle(inputTypeAttrib.fileNames)
-                    self.logger.info("Added %s events from the pileup blocks", eventsAvailable)
-
-                    # Handle enforceGUIDInFileName for pileup
-                    self.handleEnforceGUIDInFileName(inputTypeAttrib)
-
-        return
-
-    def _getPileupMixingModules(self):
```
amaltaro, line 476: Similar to the comment above, do you think this module adds some value and we should retain some of this information somewhere? Either as a docstring, wiki, cmssw-tools documentation, etc?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -728,89 +687,83 @@ def __call__(self):
...
+            self.logger.info("DEBUG: Now in the none scenario to load PSET")
             try:
                 self.loadPSet()
             except Exception as ex:
                 self.logger.exception("Error loading PSet:")
                 raise ex
 
+        dummyTweak = PSetTweak()
+        procScript = "edm_pset_tweak.py"
+        dummypsetTweakJson = os.path.join(self.stepSpace.location, "PSetTweak_dummy.json")
```
amaltaro, line 700: Can you please clarify - through a docstring - what this tweak_dummy object is?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -728,89 +687,83 @@ def __call__(self):
...
+        dummyTweak.persist(dummypsetTweakJson, formatting='simplejson')
+
+        cmd = "%s --input_pkl %s --output_pkl %s --json %s" % (
+            procScript,
+            os.path.join(self.stepSpace.location, self.configPickle),
+            os.path.join(self.stepSpace.location, "PSet_before.pkl"),
```
amaltaro, line 706: What is this `PSet_before.pkl` for? If it's the original PSet and we keep it in the worker node, then it will be part of the logArchive tarball, which can increase its size significantly in case it has a large pileup dataset information (injected all the way from McM).

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
amaltaro: We leave it as it is then. Thanks

amaltaro (commented):
@khurtado Kenyi, code is looking good but I still have a few comments/questions inline. Which might or not require changes to the code.

There are some python3 incompatible changes here, and a few pylint/aesthetic changes to be made. However, the changes already in place are so large that I think we should carry those in a future PR. FYI @mapellidario

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -5,190 +5,119 @@
 
 """
 from __future__ import print_function
-from future.utils import viewitems
```
mapellidario, line 8:
```suggestion
from builtins import next, object
```

These will come handy later

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -5,190 +5,119 @@
...
-        process.maxEvents = cms.untracked.PSet(input=cms.untracked.int32(-1))
-    if not hasattr(process.maxEvents, "input"):
-        process.maxEvents.input = cms.untracked.int32(-1)
-    return
-
 
-def fixupFileNames(process):
-    """
-    _fixupFileNames_
+    class DummyClass:
```
mapellidario, line 33:
```suggestion
    class DummyClass(object):
```

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -272,32 +205,48 @@ def fixupProcess(self):
         #   dataset.dataTier
         #   dataset.filterName
         if hasattr(self.process, "outputModules"):
-            outputModuleNames = list(self.process.outputModules)
-        else:
+            outputModuleNames = self.process.outputModules.keys()
```
mapellidario, line 208:
```suggestion
            outputModuleNames = list(self.process.outputModules)
```

In python3 `dict.keys()` returns a [dictionary view object](https://docs.python.org/3.8/library/stdtypes.html#dict-views), that behaves like a generator. In order to have a list in `outputModuleNames`, we have to materialize this generator with `list()`. Eventually, `list(mydict)` is the same as `list(mydict.keys())`, we can use the one that we prefer.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -272,32 +205,48 @@ def fixupProcess(self):
...
+        elif hasattr(self.process, "outputModules_"):
             outputModuleNames = self.process.outputModules_()
+        elif hasattr(self.process, "_Process__outputmodules"):
+            outputModuleNames = self.process._Process__outputmodules.keys()
```
mapellidario, line 212:
```suggestion
            outputModuleNames = list(self.process._Process__outputmodules)
```

Same as above

mapellidario (changes requested):
Hello! The py3 compatibility of these changes is pretty good! 

I have only a couple of suggestions and it may be quicker for us to discuss them here instead of opening a new PR just for modernizing this PR. Is this ok with you, @khurtado?

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: `I guess you missed this?` I removed it, but it seems it made it back somehow (during the squash?)
`I guess the answer is no, nothing else is needed, right?` Right.
As for pileup, I think it's still using the same logic here:
https://github.com/cms-sw/cmssw-wm-tools/blob/master/bin/cmssw_handle_pileup.py#L88-L91

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: It is mentioned in cmssw-wm-tools here:
https://github.com/cms-sw/cmssw-wm-tools/blob/9f91d3f5cd6463761578f8eeb76dd2f2fc474cdc/bin/cmssw_handle_pileup.py#L17-L23

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: Correct, nothing else should be needed.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: Agreed. I will reincorporate the docstrings

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: Ah! I actually don't need this. This is something I added when I was debugging some issues. I will remove it.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
khurtado: As mentioned above, I will just remove this, as it is not needed.

khurtado:
> Hello! The py3 compatibility of these changes is pretty good!
> 
> I have only a couple of suggestions and it may be quicker for us to discuss them here instead of opening a new PR just for modernizing this PR. Is this ok with you, @khurtado?

@mapellidario Sounds good to me. I just added your changes in a new commit. Let me know if anything else is needed.

khurtado:
@amaltaro I added a couple of commits to the PR to address your comments and Dario's python3 related change requests.

amaltaro:
@khurtado Thanks Kenyi. Could you please squash commits 3 and 4 into the 1st one? I will likely cherry-pick these for the branch 1.4.7_wmagent to run a final validation before pushing it in for RelVal (likely on Monday/Tuesday).

@mapellidario just to make your re-review easier, here is the current 4th commit where Kenyi made the python3 changes you requested: https://github.com/dmwm/WMCore/pull/10179/commits/3f2ba535d5d76cdbf95a1cea1b321b33c8531711
AFAIK, even if he squashes those, you can still access this commit if you have the correct url, thus pasted above ;)

khurtado:
@amaltaro Done.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
davidlange6: I definitely did not understand this logic. - no. . I just copied what was in wmcore to preserve functionality.. It seemed overly complicated to me at the time.

amaltaro: Thanks for the pointer! I have that repository in my bookmarks now.

src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -705,20 +668,35 @@ def __call__(self):
         """
         self.logger.info("Executing SetupCMSSWPSet...")
         self.jobBag = self.job.getBaggage()
+        self.configPickle = getattr(self.step.data.application.command, "configurationPickle", "PSet.pkl")
+        self.psetFile = getattr(self.step.data.application.command, "configuration", "PSet.py")
+        self.scram = self.createScramEnv()
 
         scenario = getattr(self.step.data.application.configuration, "scenario", None)
         if scenario is not None and scenario != "":
+            self.logger.info("DEBUG: I'm in scenario")
```
amaltaro, line 677: Should we remove this line?

amaltaro:
@khurtado Kenyi, I had a quick look at the replay logs and I believe we are passing a unicode string for the `repack` jobs. Here is the error message/traceback extracted from the logs provided by Jhonatan:
```
2021-05-02 08:37:57,869:INFO:Scram:    Invoking command: cmssw_wm_create_process.py --output_pkl /srv/job/WMTaskSpace/cmsRun1/PSet.pkl --funcname repack --funcargs /srv/job/WMTaskSpace/cmsRun1/process_funcArgs.json
2021-05-02 08:38:47,324:INFO:Scram:Subprocess stdout was:
Failed to create a repack process.
  File "/cvmfs/cms.cern.ch/share/overrides/bin/cmssw_wm_create_process.py", line 102, in <module>
    main()
  File "/cvmfs/cms.cern.ch/share/overrides/bin/cmssw_wm_create_process.py", line 93, in main
    process=create_process(args, func_args)
  File "/cvmfs/cms.cern.ch/share/overrides/bin/cmssw_wm_create_process.py", line 40, in create_process
    raise ex
ValueError: write_L1Accept_RAW.root is not a valid <class 'FWCore.ParameterSet.Types.string'>
```

While we keep working on this master branch PR, I'm going to cherry-pick these commits to `1.4.7_wmagent` such that we can start working on the relval agent validation. I think the potential changes to fix it for T0 will be isolated enough, so we can just move on with our plan discussed a couple of weeks ago. Please let me know what your thoughts are.

amaltaro:
Given how critical and deep these changes are in our system, I'd suggest to define a more meaningful commit message as well.

amaltaro:
@khurtado Kenyi, I was mistaken during our chat. This PR still needs to be updated with a better PR/commit message. Can you please update it before your day is over? Thanks

khurtado:
@amaltaro I changed the title and tried to detail the changes more. Let me know if I'm missing anything

amaltaro:
Thanks @khurtado . Could you please reword the commit message as well? Thanks

khurtado:
@amaltaro Ah, I forgot that. I just reworded the commit message.

khurtado:
As discussed with @amaltaro on Monday, we are pushing this PR now that the commit message and description has been updated.

amaltaro:
Thank you very much, Kenyi!

PR #10152, opened 2020-12-04
Catch UnicodeDecodeError in JobAccountant

amaltaro:
Fixes #10151 

#### Status
ready

#### Description
Catch UnicodeDecodeError exception when loading a job FWJR in JobAccountant, and create a fresh/empty FJR in place.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #10187, opened 2020-12-15
 [py2py3] Fix potential bug in WMException - again

### LINKED ISSUE #10173: [py2py3] Fix potential bug in WMException

mapellidario:
Fixes #10173 

#### Status
ready

#### Related PRs

This PR completes the work started in #10174.

This PR also supersedes #10168 , since this involve `self[key]` -> `self.data[key]` (it is not compulsory to have this change in this PR, we can keep this in a separate PR)

An alternative to this PR is #10189 

*Cons*

This PR will likely be not backward compatible. Some changes like in https://github.com/dmwm/WMCore/commit/04b7b03e1d025a9ce595f1d29b49cde62468ab36 will be needed. We fixed all the problems that were caught by unit tests, but it is possible that some other problems will be discovered in testbed only.

*Pro*

This PR makes WMException ready to run in py3

#### Description

In #10174 we focused only on how to get WMException work for modernized code. But WMCore is not fully modernized yet, and it can break if it is called from non-modernized code. We have to follow a different approach. 

The problem arises since in python2 
- `str()` asks for a string of type sequence of bytes (non-modernized code)
- while if we use `from builtins import str`, then `str()` asks for a string of type sequence of unicode codepoints. (modernized code)

Since our current `__str__()` returns unicode codepoints, due to this line,

https://github.com/dmwm/WMCore/blob/08c1eea1a3245ebf956438e0a1ab1b79e338ff40/src/python/WMCore/WMException.py#L173

when we call `str(myexception)` from non-modernized python2 code `str()` tries to encode the unicode string with the default encoding (which is `ascii` by default) and it breaks. We would need to force `utf-8` encoding when `str()` is called from a non-modernized py2 module, but this would require changing some environment variable and would be time-consuming to deploy and check. 

When we use a modernized py2 module, `str()` expects that `__str__` returns a unicode string, so everything works.

The solution to make WMException compatible with both modernized and non-modernized code has been inspired from [stackoverflow](https://stackoverflow.com/a/57485167) and involves a careful separation of `__str__`, `__unicode__` and `__repr__`. Moreover, this solution seems to be python3-compatible.

### DIFF
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -3,8 +3,9 @@
 from __future__ import division, print_function
 from future.utils import viewitems
 
-from builtins import str
+from builtins import str, bytes
 from past.builtins import basestring
+
 import subprocess
 import os
 import re
@@ -189,6 +190,38 @@ def getSize(obj):
                 need_referents.append(obj)
         objects = get_referents(*need_referents)
     return size
+
+
+def decodeBytesToUnicode(value, errors="strict"):
+    """
+    Accepts an input "value" of generic type.
+
+    If "value" is a string of type sequence of bytes (i.e. in py2 `str` or
+    `future.types.newbytes.newbytes`, in py3 `bytes`), then it is converted to
+    a sequence of unicode codepoints.
+
+    This function is useful for cleaning input data when using the
+    "unicode sandwich" approach, which involves converting bytes (i.e. strings
+    of type sequence of bytes) to unicode (i.e. strings of type sequence of
+    unicode codepoints, in py2 `unicode` or `future.types.newstr.newstr`,
+    in py3 `str` ) as soon as possible when recieving input data, and
+    converting unicode back to bytes as late as possible.
+    achtung!:
+    - converting unicode back to bytes is not covered by this function
+    - converting unicode back to bytes is not always necessary. when in doubt,
+      do not do it.
+    Reference: https://nedbatchelder.com/text/unipain.html
+
+    py2:
+    - "errors" can be: "strict", "ignore", "replace",
+    - ref: https://docs.python.org/2/howto/unicode.html#the-unicode-type
+    py3:
+    - "errors" can be: "strict", "ignore", "replace", "backslashreplace"
+    - ref: https://docs.python.org/3/howto/unicode.html#the-string-type
+    """
+    if isinstance(value, bytes):
+        return value.decode("utf-8", errors)
+    return value
 
 
 # TODO: remove this function once we have completely migrated to Rucio
--- a/src/python/WMCore/WMException.py
+++ b/src/python/WMCore/WMException.py
@@ -14,8 +14,11 @@ import logging
 import sys
 import traceback
 
-WMEXCEPTION_START_STR = "<@========== WMException Start ==========@>"
-WMEXCEPTION_END_STR = "<@---------- WMException End ----------@>"
+from Utils.Utilities import decodeBytesToUnicode
+from WMCore.Configuration import PY3
+
+WMEXCEPTION_START_STR = str("<@========== WMException Start ==========@>")
+WMEXCEPTION_END_STR = str("<@---------- WMException End ----------@>")
 
 
 class WMException(Exception):
@@ -29,12 +32,11 @@ class WMException(Exception):
 
     def __init__(self, message, errorNo=None, **data):
         self.name = str(self.__class__.__name__)
-        if isinstance(message, bytes):
-            # Fix for the unicode encoding issue, see #8056 and #8403
-            # interprets this string using utf-8 codec and ignoring any errors
-            # unicode sandwich: convert sequence of bytes strings to
-            # unicode codepoints strings as soon as possible
-            message = message.decode('utf-8', 'ignore')
+
+        # Fix for the unicode encoding issue, see #8056 and #8403
+        # interprets this string using utf-8 codec and ignoring any errors.
+        # using unicode sandwich pattern
+        message = decodeBytesToUnicode(message, "ignore")
 
         Exception.__init__(self, self.name, message)
 
@@ -95,6 +97,8 @@ class WMException(Exception):
             self.traceback = "\n".join(traceback.format_tb(sys.exc_info()[2]))
         except Exception:
             self.traceback = "WMException error: Couldn't get traceback\n"
+        # using unicode sandwich pattern
+        self.traceback = decodeBytesToUnicode(self.traceback, "ignore")
 
     def __getitem__(self, key):
         """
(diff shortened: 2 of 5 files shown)


mapellidario:
The two failures at `test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py` are unexpected. I will have a better look tomorrow, with some fresh energies.

mapellidario:
I changed the order of the first two commits to better reflect the history of my investigation, so that what happened can be more clear to those who read.

### overview of the PR's branch

1. the commit https://github.com/dmwm/WMCore/pull/10187/commits/0d43317ffbe5bac6f99a9bd92e7dd64aa4fc6319 is about new tests for `WMException.py`, which represent how it handles exception rasied by both modules compatible with py2 only and modules modernized to py2py3.
  - the test `test/python/WMCore_t/WMException_py2_t.py` fails (in accordance to alan's observations in #10152 ).
2. the commit https://github.com/dmwm/WMCore/pull/10187/commits/574a6ad0550f8201c8339cb0e6a34e183439f994 is a thorough update of WMException to make it compatible with python2 and python3.
  - the test `test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py:CMSSW_t.testB_ExecuteNonZeroExit` fails because `src/python/WMCore/WMSpec/Steps/Diagnostic/CMSSW.py` can not append `unicode` to `future.types.newbytes.newbytes` in py2, while it was able to append `unicode` to `str`.
3. the commit https://github.com/dmwm/WMCore/pull/10187/commits/9785e159abd0edaacbbed4db21fe1a7949fb4502 is a modernization of `src/python/WMCore/WMSpec/Steps/Diagnostic/CMSSW.py`, so that when it calls `str(excepInst)` it retrieves a unicode string (which can be used to append `unicode` to)
  - all the unit tests succeed now

### description of why `CMSSW_t` failed

Modernizing WMException meant changing what is returned when we call `str(myexception)`. If WMException supports py2 only, then this should `str(myexception)` is python2 type `str` (which is a sequence of bytes), which follows all the py2 rules. One of these rules is that when we concatenate `str` to `unicode` in py2, then the interpreter gets clever and encodes the unicode with the default `ascii` encoding so that is can concatenate two `str` objects. 

If we modernize WMException, then when we call `str(myexception)` from a module that supports py2 only,  it is `future.types.newbytes.newbytes` (which is again a sequence of bytes). But there is a catch: `future.types.newbytes.newbytes` has the same interface of pythhon3 type `bytes`, which forbid the python interpreter to get clever and raises an exception if we try to append `unicode` to `future.types.newbytes.newbytes`. While `future.types.newbytes.newbytes` is equivalent to py2 `str`, the fact that it mimics py3 `bytes` behavior means that some operation that we give for granted (like appending `unicode` to `str`) are now forbid. 

### comment

Modernizing `WMException.py` in the proper way (i.e. change it so that it is compatible properly with both py2 and py3) meant changing its interface, which broke the module `src/python/WMCore/WMSpec/Steps/Diagnostic/CMSSW.py` since it was not modernized yet. 

While in this case all other unit tests succeed, it is possible that 
- there are other similar problems related to WMException interface change. It is possible that there is another module that is not modernized, that has something like `str(myexception)` and that this is not covered by a unit test. We could hunt such occurrences, but this would mean going outside the boundary of current slices and starting modernizing code that we do not plan to address in this moment.
- the very same kind of failure will appear again in the future, not concerning WMException, but maybe some other basic module.

Lesson 1: modernizing very basic modules can break the slice boundaries.

Why did we decide to approach the gradual migration in this risky way? (and I take full responsibility for this) Short answer: because I did not know that it was risky in the first place. I assumed that having a module compatible with both py2 and py3, would mean that its interface would not change in behavior and that it could have been safely imported and used by modules supporting py2 only. This is not the case, though.

Moreover, if we imagine to have 
- a very basic module `A.py` and let's say that this is our migration slice 1
- two modules `B1.py` and `B2.py` that both `import A`, let's call this slice 2
- a module `C.py` that `import `B2.py`, let's call this slice 3
If we start the migration from `A`, then move to `B1` and `B2`, we would be able to run B1 with pythhon3 before migrating `C`. This was driven by the hope that this was the best option to have some modules that fully support py3 _before_ completing all the migration slices. The hidden hypothesis that I assumed was that having module `A` (or let's say slice 1) supporting py2 and py3 would not break `B1`/`B2`/`C` in py2. 

What we observed is that migrating `WMException`, which was migrated before and independently of every other slice, broke `Diagnostics/CMSSW.py`, which is in slice 15. This makes me wonder if starting from slice 1 (the most basic modules) is a viable strategy or if we should  completely change the approach and start from what now is in the last slice (the least basic slices), then move to the more delicate ones. This approach is explored in a different PR

amaltaro:
Dario, as we discussed, I'm applying this patch to my VM and dev WMAgent to see how a few test workflows go. Will update it again in a day or two.

amaltaro (approved):
Code looks good and well tested to me. Shall we merge it and put it in a new tag and keep testing it for the coming week or two?

mapellidario:
Yes! This would be wonderful! Thank you very much!

amaltaro:
How about we squash the src commits then?

mapellidario:
Squashed!

mapellidario:
The errors in `pylint py3k` report provided by jenkins are in `WMException_py2.py`, which is meant to emulate a non-modernized module that calls `WMException` and it is not meant to be run by a python3 interpreter.
The other warnings surprise me since they involve files are not involved in this PR.

mapellidario:
Test again please

amaltaro:
test this please

amaltaro:
@mapellidario I guess the reported pylint py3k issues reported are expected, right?
```
    test/python/WMCore_t/WMException_py2_t.py
        E1610, line 41: Non-ascii bytes literals not supported in 3.x
        E1610, line 44: Non-ascii bytes literals not supported in 3.x
        W1612, line 68 in WMExceptionTest.testExceptionUnicode1: unicode built-in referenced
        W1612, line 72 in WMExceptionTest.testExceptionUnicode1: unicode built-in referenced
    test/python/WMCore_t/WMException_py2py3_t.py
        W1612, line 97 in WMExceptionTest.testExceptionUnicode1: unicode built-in referenced
```

If so, shall we proceed and merge it?

mapellidario:
Your trained eye spotted a minor issue: 

```plaintext
test/python/WMCore_t/WMException_py2py3_t.py
        W1612, line 97 in WMExceptionTest.testExceptionUnicode1: unicode built-in referenced
```

We have two new unit tests that supersedes the one we had before:
* `WMException_py2_t.py`
  - it simulates a non-modernized python2 module that imports `WMException` and that it is meant to be run from py2 interpreter only
  - all the `pylint --py3k` errors and warnings are expected in this module
* `WMException_py2py3_t.py` 
  - it simulates a modernized module that is meant to be run from either python2 and python3
  - `pylint --py3k` was right in its complaints! I did not notice because I only run it from py2 interpreter. It is better not to use `unicode()` in a module that is supposed to be modernized!

Thanks for implementing the new section in jenkins report, it's very helpful!

I updated the commit with the test changes to integrate `pylint --py3k` suggestion

amaltaro:
Thanks Dario!

PR #10189, opened 2020-12-16
[py2py3] Fix potential bug in WMException - No modernization of WMException

### LINKED ISSUE #10173: [py2py3] Fix potential bug in WMException

mapellidario:
Fixes #10173

#### Status

Ready

#### Related PRs

This PR is an alternative to #10187, which is preferred to this PR.

This PR also supersedes #10168 , since this involve `self[key]` -> `self.data[key]` (it is not compulsory to have this change in this PR, we can keep this in a separate PR)

_Pro_

This PR strives to be as backward compatible as possible

_Cons_

WMException would not be modernized and will not run in python 3. This PR is intended to be a temporary patch only.

#### Description

1. the first commit is a cherry-pick of the first commit of #10187, it is about new tests for `WMException.py` to check how it handles exceptions raised by both modules compatible with py2 only and modules modernized to py2py3.
    - the test `test/python/WMCore_t/WMException_py2_t.py` fails (in accordance to Alan's observations in #10152 ).
2. The second commit is an alternative to the second commit of #10187. This commit does not modernize WMException, but makes it compatible with both modules that are modernized and modules that support py2 only.
    - the test `test/python/WMCore_t/WMException_py2_t.py` succeeds again, `test/python/WMCore_t/WMException_py2py3_t.py` does not fail, `test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py` tests do not fail. Everything is fine.
3. The third commit will **not** be needed if and when merging this PR, but it is a test. Following the schema presented in [this comment](https://github.com/dmwm/WMCore/pull/10187#issuecomment-746711709), it mimics migrating `B2` (that we called slice 2) before migrating `A` (that we called slice 1).
    - no tests fail at this stage
 4. The fourth commit is the full modernization of `WMException`. This will **not** be needed if and when merging this PR. It mimics the modernization of `A` after having modernized `B2`. 
     - no tests fail at this stage. Or better , some code that depends on WMException and that is not covered by unit tests may break, but no unit tests in the scope our experiment break, which mean that `test/python/WMCore_t/WMException_*_t.py` and `test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py` do not fail.

This a new approach to the migration:
- when we find a bug in a module that supports py2 only, we fix the bug keeping the support for py2 only
- we start the migration from the last slice, in our case from the 26th
- then we proceed with the previous one, and so on.

This has the following advantages:
- keeping the code fully py2 compatible (or at least it should be, but there is no evidence at this moment that would suggest me otherwise) so that we can keep running our code in py2 while we modernize WMCore
- by updating before the "client" code and only then the more "basic"/"core" one, we should not encounter errors such as in #10187 .

This new approach does not have any evident drawback. 

This PR requires python-future only, no changes are required.

### DIFF
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -3,8 +3,9 @@
 from __future__ import division, print_function
 from future.utils import viewitems
 
-from builtins import str
+from builtins import str, bytes
 from past.builtins import basestring
+
 import subprocess
 import os
 import re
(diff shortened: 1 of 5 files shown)


mapellidario:
We decided that #10187 is the best approach to tackle #10173 for us at this moment. I am closing this PR.

PR #10194, opened 2020-12-17
Fix pileup location with multi-RSE rules

### LINKED ISSUE #10193: Pileup with multi-RSE rules can have a wrong block location
**Impact of the bug**
WMAgent

**Describe the bug**
This bug has been exposed while I was trying to hack one of the agents to let the P&R team to run a backfill workflow with container only partially available.
Problem affects WMAgent, through PileupFetcher module, when looking for pileup availability and locks and producing the usual pileupconf.json file. And the origin of the problem is at this line:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Rucio/Rucio.py#L1039

which assigns `finalRSEs` to result by reference. Which means, as soon as we add any other RSE to the `result` variable, it gets reflect to all the other blocks that point to that memory address.

**How to reproduce it**
Resolve pileup location for a container having a rule locking the data at multiple RSEs.

**Expected behavior**
Block location should be properly built, considering only the container-level rules and dataset_locks matching those rules.
Thus, this block needs fixing:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Rucio/Rucio.py#L1046-L1051

none

amaltaro:
Fixes #10193 

#### Status
ready

#### Description
This fixes the way blocks location were added to the `result` structure, avoiding to use the same memory reference (content) for every single block. Silly variable by value vs variable by reference mistake!

yes

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/Services/Rucio/Rucio.py
+++ b/src/python/WMCore/Services/Rucio/Rucio.py
@@ -1044,11 +1044,13 @@ class Rucio(object):
         # if we got here, then there is a third step to be done.
         # List every single block lock and check if the rule belongs to the WMCore system
         for blockName in result:
+            blockRSEs = set()
             for blockLock in self.cli.get_dataset_locks(scope, blockName):
                 if isTapeRSE(blockLock['rse']):
                     continue
                 if blockLock['state'] == 'OK' and blockLock['rule_id'] in multiRSERules:
-                    result[blockName].add(blockLock['rse'])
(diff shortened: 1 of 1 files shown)


amaltaro:
Tested both via unit tests (after local changes) and on the backfill agent.

todor-ivanov (approved):
Looks good to me

PR #10208, opened 2021-01-05
Fix indentation in WebTools/Root

### LINKED ISSUE #10207: Unexpected indentation in Root.py
**Impact of the bug**
DBS migrate monitoring was impacted.
**Describe the bug**
Please see https://github.com/dmwm/DBS/issues/644
**How to reproduce it**
Please see https://github.com/dmwm/DBS/issues/644

**Expected behavior**
no "Unexpected indentation" error

DBS currently is using WMCore 1.3.2. If the latest version is compatible or no major change affect to DBS , we can move to the current version. Otherwise, please provide new tag that are build from 1.3.2.

amaltaro:
Fixes #10207 

#### Status
not-tested

#### Description
Fix indentation and also make that rbytes assignment more clear, to me at least.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WebTools/Root.py
+++ b/src/python/WMCore/WebTools/Root.py
@@ -96,16 +96,17 @@ class WTLogger(LogManager):
         if not rbytes:
             try:
                 # request.rfile.rfile.bytes_read is a custom CMS web
-                #  cherrypy patch not always available, hence the test
-                rbytes = (getattr(request.rfile, 'rfile', None)
-                        and getattr(request.rfile.rfile, "bytes_read", None)
-                        and request.rfile.rfile.bytes_read) or "-"
+                # cherrypy patch not always available, hence the test
+                if request.rfile.rfile.bytes_read:
+                    rbytes = request.rfile.rfile.bytes_read
+                else:
+                    rbytes = "-"
             except:
                 try:
                     # this will work only when body is read from request
                     rbytes = cherrypy.request.body.fp.bytes_read
                 except:
-                    rbytes = "-"        
+                    rbytes = "-"
         msg = ('%(t)s %(H)s %(h)s "%(r)s" %(s)s'
                + ' [data: %(i)s in %(b)s out %(T).0f us ]'
                + ' [auth: %(AS)s "%(AU)s" "%(AC)s" ]'
--- a/test/python/WMCore_t/MicroService_t/MicroService_t.py
+++ b/test/python/WMCore_t/MicroService_t/MicroService_t.py
@@ -8,11 +8,7 @@ from __future__ import division, print_function
 from future.utils import viewitems
 
 import unittest
-import time
-import random
 import cherrypy
-import gzip
-import json
 from WMCore_t.MicroService_t import TestConfig
 from WMCore.MicroService.Service.RestApiHub import RestApiHub
 from WMCore.MicroService.Tools.Common import cert, ckey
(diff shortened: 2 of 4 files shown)


src/python/WMCore/WebTools/Root.py
```diff
@@ -93,16 +93,17 @@ def access(self):
         if not rbytes:
             try:
                 # request.rfile.rfile.bytes_read is a custom CMS web
-                #  cherrypy patch not always available, hence the test
-                rbytes = (getattr(request.rfile, 'rfile', None)
-                        and getattr(request.rfile.rfile, "bytes_read", None)
-                        and request.rfile.rfile.bytes_read) or "-"
+                # cherrypy patch not always available, hence the test
+                if request.rfile.rfile.bytes_read:
+                    rbytes = request.rfile.rfile.bytes_read
```
vkuznet, line 98: Alan, I don't think it is safe to assume that request may have rfile, and rfile may have bytes_read data. I think getattr is better approach. But what you probably need is the following:
```
rfile = None
if hasattr(request, "rfile"):
   rfile = request.rfile
  if hasattr(rfile, "rfile"):
     rfile = rfile.rfile
bytes_read = None
if rfile and hasattr(rfile, "bytes_read"):
   bytes_read = getattr(rfile, "bytes_read")
if bytes_read:
   rbytes = bytes_read
else:
   rbytes = "-"
```

I suggest to provide proper unit test to pass different structure of request, e.g. if it is real cherrypy request, if it is None, etc.

src/python/WMCore/WebTools/Root.py
```diff
@@ -93,16 +93,17 @@ def access(self):
         if not rbytes:
             try:
                 # request.rfile.rfile.bytes_read is a custom CMS web
-                #  cherrypy patch not always available, hence the test
-                rbytes = (getattr(request.rfile, 'rfile', None)
-                        and getattr(request.rfile.rfile, "bytes_read", None)
-                        and request.rfile.rfile.bytes_read) or "-"
+                # cherrypy patch not always available, hence the test
+                if request.rfile.rfile.bytes_read:
```
vkuznet, line 97: This is not safe as it will throw exception if any nested member data does not exist. You cannot use if for that. Instead, you should walk through every structure and identify if there is specific data you request in it, see my suggestion below.

vkuznet (changes requested):

src/python/WMCore/WebTools/Root.py
amaltaro: That nested getattr only make things unclear. Given that we are inside a try/except block, we can just try to access that attribute directly, which will fall into the except block if it does not exist.

src/python/WMCore/WebTools/Root.py
vkuznet: It is fine with me, I just pointed out to potential problem. Anyway, I still suggest to add proper unit test to test all possible scenarios how to access rbytes value and check that no identation error will be thrown.

amaltaro:
Thanks for your comments, @vkuznet . It looks like the WebTools/Root unit tests were broken (missing an import). I fixed that and added another unit test and everything passes in my container.

The interesting part is, if I fix the unit tests and ran them against master, all of them succeed and there are no exceptions (neither indentation errors), so I'm not totally sure whether this is going to fix the problem reported...

@yuyiguo Yuyi, any chance you could try to apply this patch to one of the cmsweb-testbed (or dev) pods and see whether it clears those errors?

amaltaro:
test this please

amaltaro:
I haven't yet managed to find a way around those 2 microservice unit tests failing now, even though nothing was changed on them...

@yuyiguo Yuyi, I don't plan to make any other changes to this source code, so feel free to test it and let me know whether it solves or not the problem.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
I just can't get these unit tests to work, no matter what I do... I might simply mark them as integration and move on, 12 new unit tests passing, 2 "deleted" ...

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
I did not manage to get the unit test working with these changes, and given that it is stale since 2021/2022, I won't bother with this PR anymore. Closing this out

PR #10210, opened 2021-01-07
Add 2021 validation campaigns.

### LINKED ISSUE #10209: Missing validation campaign configurations for 2021
**Impact of the bug**
Validation

**Describe the bug**
We have the test/validation campaign names hardcoded in the `parseUnifiedCampaigns.py` script and the those for the current year are missing.

**How to reproduce it**
Any injection with a campaign name such as `HG_21**_Val` would fail to move workflows beyond the `assigned` status.  

**Expected behavior**
Have the validation campaign identifiers for the whole year at least

todor-ivanov:
Fixes #10209

#### Status
ready

#### Description
Adding the testing/validation campaign identifiers to the `parseUnifiedCampaigns.py` script.

YES


amaltaro (approved):
Thanks Todor.

PR #10079, opened 2020-11-18
Keep only ruleID instead of the full Rucio rule.

todor-ivanov:
Fixes #10077 

#### Status
ready

#### Description
With this change we drop the full Rucio rule that we were keeping  in the workflow representation, and we are just preserving the ruleID. 

#### Is it backward compatible (if not, which system it affects?)
YES

#### Related PRs
NO

#### External dependencies / deployment changes
NO

PR #10224, opened 2021-01-14
Clean input data rules

### LINKED ISSUE #10017: Remove input data placement rules for data no longer needed by active workflows
**Impact of the new feature**
MicroServices

We need to clean the Rucio data placement Rule created by MSTranferor. For the purpose we are planing to use the MSRuleCleaner service in a way similar to what we already have for the output data rules, but also taking into account all the global locks for the datasets coming from `MSTransferror`. More details on the logic follow in the next section.

To Implement/plug the following logic into `MSRuleCleaner`:

***Globally - on every poling cycle:***
   * Build the `globalLocks` lists:
      * Fetch the global locks list from wmstat   
      * Fetch Parentage locks list from reqmgr
      * Make an union of them both

***Locally - in the `MSTr*` pipelines (per workflow):***
   * Build the lists of input datasets rules:
      *   Check the `InputDataset` from the workflow representation
      * if: datasets do not  exist in the `globalLocks` list:
         * Fetch the Rucio rules, which belong to `wmcore_transferor` acct
         * Include them into `RulesToClean` at the workflow data structure
      * else: Log.
   * Build the list of all Parent datasets rules:
      * Check `includeParents` flag
      * Resolve Parent datastes names via DBS: [1]
      * if: datasets do not exist in the `globalLocks` list:
          * Fetch the Rucio rules, which belong to `wmcore_transferor` acct 
          * Include them into `RulesToClean` at the workflow data structure
      * else:  Log
   * Build the list of Secondary datasets rules:
      * Check for `{MC,Data}Pileup` from the workflow representation (to be noted - they may be not only as a top level keys in the workflow description  coming from `reqmgr2`)
      * if: datasets do not  exist in the `globalLocks` list:
         * Fetch the Rucio rules, which belong to `wmcore_transferor` acct
         * Include them into `RulesToClean` at the workflow data structure
      * else: Log
   * Clean everything from the `RulesToClean` field from the workflow data structure
   * Mark the workflow as clean following the general logic for the rest of the micro service

***New fields to be added to the workflow representation data structure:***
* `IncludeParents`
* `{MC,Data}Pileup`
* `InputDataset`

[1]
https://github.com/dmwm/WMCore/blob/ebfc5159d0245ad41be1701be450b106425cfcc1/src/python/WMCore/Services/DBS/DBS3Reader.py#L777

No Alternative have been discussed so far

**Additional context**
https://github.com/dmwm/WMCore/wiki/Data-Locking-in-WMCore
https://github.com/dmwm/WMCore/wiki/Workflow-creation-and-assignment-definition

todor-ivanov:
Fixes #10017 

#### Status
Ready

#### Description
Fetches the global and parent locks lists from both WMStatsServer and ReqMgr and implements the rest of the logic explained in the issue itself

YES 

#### Related PRs
https://github.com/dmwm/deployment/pull/1035
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/72
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/73
 
No


src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +132,38 @@ def __init__(self, msConfig, logger=None):
...
+        'parentLocks' from request manager. Stores/updates the unified set in
+        the 'globalLocks' instance variable. Returns the resultant unified set.
+        :return: A union set of the 'globalLocks' and the 'parentLocks' lists
+        """
+        globalLocks = set()
+        parentLocks = set()
+        self.logger.info("Fetching globalLocks list from wmstats server.")
+        try:
+            uri = 'globallocks'
+            globalLocks.update(set(self.wmstatsSvc._getResult(uri, clearCache=True, verb="GET")))
```
amaltaro, line 149: Todor, instead of a direct call to that private method. I would suggest to implement a public API on the WMStatsServer module and use it. If the server returns a flat list, you might even cast it to a set() to be returned.

One more comment on this line, I suppose you want to simply refresh the list of locks (not to update it based on the previous cycle), right?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +132,38 @@ def __init__(self, msConfig, logger=None):
...
+        except Exception as ex:
+            msg = "Failed to refresh GlobalLocks list for the current poling cycle. "
+            msg += "Reusing the list (if any) from the previous run. Error: %s"
+            self.logger.warning(msg, str(ex))
+            globalLocks.update(self.globalLocks)
+        self.logger.info("Fetching parentLocks list from wmstats server.")
+        try:
+            uri = 'parentlocks/ALL_DOCS'
+            result = self.reqmgr2._getResult(uri, clearCache=True, verb="GET")
```
amaltaro, line 158: Exactly the same comment goes here.

In addition to that, the URI is wrong. The REST endpoint is just `parentlocks`, e.g.:
https://cmsweb-testbed.cern.ch/reqmgr2/data/parentlocks

amaltaro (commented):
Todor, even though this PR is in an initial phase, I have made a couple of comments to be considered along the code.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Hi @amaltaro I agree this is good to have a dedicated  API  for that, but I am trying to avoid touching major components of a central service with a midmonth patch. We must for sure fix this for the next release. The current method is working well so far.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Same reply, though. 
I tested it and strangely enough it works with both those URIs (with and without `/ALL_DOCS`). Nevertheless I am fixing it as you requested here.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: Experience tells me that something non-ideal that goes to production, remains in production!
So I'd rather have it right since the beginning.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: Same counter-reply here, I'd be very happy to see a proper method to fetch the list of parentlocks.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: @amaltaro I totally agree with you here, and I am about to develop those two methods of course. In the meantime you may also start reviewing the rest of the proposed changes. I will write back here once I am ready with these two.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Same as above.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +573,60 @@ def getMSOutputTransferInfo(self, wflow):
             wflow['TransferDone'] = True
         return wflow
 
-    def getRucioRules(self, wflow, gran, rucioAcct):
+    def setParentDatasets(self, wflow):
+        """
+        Used to resolves parent datasets for a workflow.
```
amaltaro, line 578: `resolve`

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +573,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+                wflow['ParentDataset'].append(parentDataset[childDataset[0]])
+                wflow['ParentDataset'] = list(set(wflow['ParentDataset']))
+                msg = "Found parent %s for input dataset %s for workflow: %s "
+                self.logger.debug(msg, parentDataset, wflow['InputDataset'], wflow['RequestName'])
+            else:
+                msg = "Could not find parent for input dataset: %s for workflows: %s"
+                self.logger.info(msg, wflow['InputDataset'], wflow['RequestName'])
+        return wflow
+
+    def getRucioRules(self, wflow, dataSpec, gran, rucioAcct):
```
amaltaro, line 595: Just an idea here. How about we drop the usage of this `dataSpec` parameter and actually implement a map (could be in the `__init__` method, or even here) like:
```
mapRuleType = {"wmcore_output": ["OutputDatasets"],
                           "wmcore_transferor": ["InputDataset", "MCPileup", "DataPileup", "ParentDataset"]}
```

and deal with all input data rules in one execution of this method. 
This way, you do not need to call this same method with multiple parameters in the pipeline definition.

amaltaro (changes requested):
Todor, as we discussed over slack, I'm concerned with the apparent complexity of the workflow data structure object and how complex it became to parse it. I have made a few comments along the code for your consideration. I'll try to provide you with another suggestion (source code) on how that could be accomplish in a cleaner and more maintainable way, at least from my perspective.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -7,6 +7,129 @@
...
+    # NOTE: extDoc is an auxiliary data structure used to accumulate all the
+    #       variables matches found in the workflow description document
+    extDoc = {}
+    for tup in docTemplate:
+        extDoc[tup[0]] = {'keyName': tup[0],
+                          'values': list(),
+                          'default': tup[1],
+                          'type': tup[2]}
+
+    def paramFinder(wfObj):
```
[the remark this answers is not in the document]
todor-ivanov: I do not think so. That's the meaning of a closure - to preserve an executable dependent entity inside the scope of the closure so that the enclosed function can refer to it while still in the scope, and eventually reach the values from its previous call. In this case this is the accumulated `extDoc`, which lives outside the enclosed function - in the closure rather than inside the enclosed function, where it is only updated with the newly fetched values during the recursive calls.
An alternative approach to this, which you may consider more readable, would be to develop a class with a call method. 
I will provide the alternative in one of my next commits.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -7,6 +7,129 @@
...
+                        in the workflow description
+    :return:            Dictionary with all the values found from the template
+                        into the workflow description
+    """
+
+    # Create the closure
+    wfParser = parseClosure(docTemplate)
+
+    # Parse the workflow description
+    extDoc = wfParser(wfDescr)
```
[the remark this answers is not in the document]
todor-ivanov: Same as above.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -81,25 +83,34 @@ def __init__(self, msConfig, logger=None):
         self.logDB = LogDB(self.msConfig["logDBUrl"],
                            self.msConfig["logDBReporter"],
                            logger=self.logger)
+        self.wmstatsSvc = WMStatsServer(self.msConfig['wmstatsUrl'], logger=self.logger)
 
         # Building all the Pipelines:
         pName = 'plineMSTrCont'
         self.plineMSTrCont = Pipeline(name=pName,
                                       funcLine=[Functor(self.setPlineMarker, pName),
+                                                Functor(self.setParentDatasets),
```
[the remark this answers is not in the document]
todor-ivanov: I do not see an error here.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -152,6 +195,8 @@ def execute(self, reqStatus):
         self.currThreadIdent = self.currThread.name
         self.updateReportDict(summary, "thread_id", self.currThreadIdent)
         self.resetCounters()
+        self.getGlobalLocks()
+        self.logger.debug("GlobalLocks: %s", pformat(self.globalLocks))
```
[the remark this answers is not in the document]
todor-ivanov: That was for my own debugging, yes. I will remove this log message.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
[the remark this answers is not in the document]
todor-ivanov: Here it is @amaltaro This is the alternative approach to the closure. Please let me know if you like it or not.
https://github.com/dmwm/WMCore/pull/10224/commits/2a9fc816050792f927ecc6f1d935cd5b82312265

p.s. I am about to keep this commit as a separate on for the time being and will continue working on the rest of your suggested changes so if the hash of the commit I pasted above changes I will try to update it here too.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -7,6 +7,126 @@
 from __future__ import division, print_function
 
 from copy import deepcopy
+from itertools import chain
```
amaltaro, line 10: Don't forget updating this to use the WMCore Utils package, as I mentioned in the previous review.

amaltaro:
Todor, I failed to find our discussion regarding the workflow parsing, so I make my comment here.

Yes, the current code looks better IMO. There were other useful comments there that might have been lost though, and further improvements could be done (like, not traversing the workflow description for keys that belong only to the MSRuleCleaner object). Please make representative unit tests to make sure it will work as expected.

todor-ivanov:
Thanks for commenting @amaltaro , I am working on those useful comments in this very moment. I will continue developing on top of the so proposed change then.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -7,6 +7,129 @@
...
+    for keyName, data in extDoc.items():
+        if len(data['values']) == 0:
+            extDoc[keyName] = deepcopy(data['default'])
+        elif len(data['values']) == 1:
+            extDoc[keyName] = deepcopy(data['values'][0])
+        elif len(data['values']) > 1:
+            if data['type'] is bool:
+                extDoc[keyName] = any(data['values'])
+            elif data['type'] is list:
+                extDoc[keyName] = list(set(chain(*data['values'])))
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +573,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+        """
+        Used to resolves parent datasets for a workflow.
+        :param  wflow:   A MSRuleCleaner workflow representation
+        :return:         The workflow object
+        """
+        if wflow['InputDataset'] and wflow['IncludeParents']:
+            childDataset = wflow['InputDataset']
+            parentDataset = findParent(childDataset, self.msConfig['dbsUrl'])
+            if parentDataset:
+                wflow['ParentDataset'].append(parentDataset[childDataset[0]])
```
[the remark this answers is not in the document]
todor-ivanov: Done - the `InputDataset` is now a solid string together with  'ParentDataset'.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +573,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+                wflow['ParentDataset'] = list(set(wflow['ParentDataset']))
+                msg = "Found parent %s for input dataset %s for workflow: %s "
+                self.logger.debug(msg, parentDataset, wflow['InputDataset'], wflow['RequestName'])
+            else:
+                msg = "Could not find parent for input dataset: %s for workflows: %s"
+                self.logger.info(msg, wflow['InputDataset'], wflow['RequestName'])
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Thanks

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,38 @@ def __init__(self, msConfig, logger=None):
...
+            uri = 'parentlocks'
+            result = self.reqmgr2._getResult(uri, clearCache=True, verb="GET")
+            parentLocks.update(set(result[0]['parentlocks']))
+        except Exception as ex:
+            msg = "Failed to refresh parentLocks list for the current poling cycle. "
+            msg += "Reusing the list (if any) from the previous run (if any). Error: %s"
+            self.logger.warning(msg, str(ex))
+            parentLocks.update(self.globalLocks)
+        self.globalLocks = globalLocks | parentLocks
+        return self.globalLocks
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,38 @@ def __init__(self, msConfig, logger=None):
...
+        parentLocks = set()
+        self.logger.info("Fetching globalLocks list from wmstats server.")
+        try:
+            uri = 'globallocks'
+            globalLocks.update(set(self.wmstatsSvc._getResult(uri, clearCache=True, verb="GET")))
+        except Exception as ex:
+            msg = "Failed to refresh GlobalLocks list for the current poling cycle. "
+            msg += "Reusing the list (if any) from the previous run. Error: %s"
+            self.logger.warning(msg, str(ex))
+            globalLocks.update(self.globalLocks)
```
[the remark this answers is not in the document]
todor-ivanov: Agreed. This indeed could be dangerous. Let me see how I could skip the whole run from this point.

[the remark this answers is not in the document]
todor-ivanov: Done. Skipping the whole cycle if any part of the globalLocks list fails to be refreshed - either globalLocks or ParentLocks.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,38 @@ def __init__(self, msConfig, logger=None):
...
+        self.logger.info("Fetching parentLocks list from wmstats server.")
+        try:
+            uri = 'parentlocks'
+            result = self.reqmgr2._getResult(uri, clearCache=True, verb="GET")
+            parentLocks.update(set(result[0]['parentlocks']))
+        except Exception as ex:
+            msg = "Failed to refresh parentLocks list for the current poling cycle. "
+            msg += "Reusing the list (if any) from the previous run (if any). Error: %s"
+            self.logger.warning(msg, str(ex))
+            parentLocks.update(self.globalLocks)
```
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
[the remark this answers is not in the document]
todor-ivanov: Done

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Not a bad idea BTW. But I think the way it is implemented right now is more flexible, though. The map will always tie the behavior of that function to the account we are calling it with. And as we saw with this exact function in just this PR, there are indeed, cases when we need more freedom. One comes to mind immediately - splitting the RelVals under a different account and taking rule metadata into consideration.

todor-ivanov:
@amaltaro Alan, I think I addressed all your comments. Please take a look. I expect some unit tests to break and I am about to deal with them in the meantime.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
```diff
@@ -72,6 +72,16 @@ def _createQuery(self, queryDict):
 
         return args.rstrip('&')
 
+    def getParentLocks(self):
+        """
+        _getParentLocks_
```
amaltaro, line 77: How about elaborating on this docstring? Specifying the return format is also useful IMO.

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
```diff
@@ -65,6 +65,15 @@ def _createQuery(self, queryDict):
 
         return args.rstrip('&')
 
+    def getGlobalLocks(self):
+        """
```
amaltaro, line 69: How about elaborating on this docstring? Specifying the return format is also useful IMO.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,34 @@ def __init__(self, msConfig, logger=None):
...
+        Fetches the list of 'globalLocks' from wmstats server and the list of
+        'parentLocks' from request manager. Stores/updates the unified set in
+        the 'globalLocks' instance variable. Returns the resultant unified set.
+        :return: A union set of the 'globalLocks' and the 'parentLocks' lists
+        """
+        globalLocks = set()
+        parentLocks = set()
+        self.logger.info("Fetching globalLocks list from wmstats server.")
+        try:
+            globalLocks.update(set(self.wmstatsSvc.getGlobalLocks()))
```
amaltaro, line 157: I'm not sure there is any performance/memory difference, but we do not need to update a set, we only need to assign it. In other words, I'm more in favor of something like:
```
globalLocks = set(self.wmstatsSvc.getGlobalLocks())
```

and there is no need to initialize it in the beginning of the method.

Same comment goes to the `parentLocks` block below.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,34 @@ def __init__(self, msConfig, logger=None):
...
+        except Exception as ex:
+            msg = "Failed to refresh global locks list for the current polling cycle. Error: %s\n "
```
amaltaro, line 159: Can you please keep the error message in the same line? So, removing the line breaker `\n` from it.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,34 @@ def __init__(self, msConfig, logger=None):
...
+            msg += "Skipping this polling cycle."
+            self.logger.error(msg, str(ex))
+            raise ex
+        self.logger.info("Fetching parentLocks list from wmstats server.")
```
amaltaro, line 163: `... from reqmgr2 server.`

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -130,6 +141,34 @@ def __init__(self, msConfig, logger=None):
...
+        try:
+            parentLocks.update(set(self.reqmgr2.getParentLocks()))
+        except Exception as ex:
+            msg = "Failed to refresh parent locks list for the current poling cycle. Error: %s\n"
```
amaltaro, line 167: Same comment here.
Can you please keep the error message in the same line? So, removing the line breaker `\n` from it.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -167,6 +205,8 @@ def execute(self, reqStatus):
 
         # Call _execute() and feed the relevant pipeline with the objects popped from requestRecords
         try:
+            self.getGlobalLocks()
+            # self.logger.debug("GlobalLocks: %s", pformat(self.globalLocks))
```
amaltaro, line 209: TODO: reminder to remove this debugging line before merging this pr.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +568,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+        """
+        Used to resolve parent datasets for a workflow.
+        :param  wflow:   A MSRuleCleaner workflow representation
+        :return:         The workflow object
+        """
+        if wflow['InputDataset'] and wflow['IncludeParents']:
+            childDataset = wflow['InputDataset']
+            parentDataset = findParent([childDataset], self.msConfig['dbsUrl'])
+            if parentDataset:
+                wflow['ParentDataset'].append(parentDataset[childDataset])
```
amaltaro, line 581: This line seems unnecessary.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +568,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+                wflow['ParentDataset'] = list(set(wflow['ParentDataset']))
+                msg = "Found parent %s for input dataset %s for workflow: %s "
+                self.logger.debug(msg, parentDataset, wflow['InputDataset'], wflow['RequestName'])
+            else:
+                msg = "Could not find parent for input dataset: %s for workflows: %s"
```
amaltaro, line 586: How about `...input dataset: %s in workflow: %s`

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +568,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+        :return:         The workflow object
+        """
+        if wflow['InputDataset'] and wflow['IncludeParents']:
+            childDataset = wflow['InputDataset']
+            parentDataset = findParent([childDataset], self.msConfig['dbsUrl'])
+            if parentDataset:
+                wflow['ParentDataset'].append(parentDataset[childDataset])
+                wflow['ParentDataset'] = list(set(wflow['ParentDataset']))
+                msg = "Found parent %s for input dataset %s for workflow: %s "
+                self.logger.debug(msg, parentDataset, wflow['InputDataset'], wflow['RequestName'])
```
amaltaro, line 584: I suggest to convert this log level to INFO. It's very rare to have such workflows, and it's an important piece of information.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: That's correct. However, the future relval use case will have a different behavior solely on the rucio account used. It's not related at all to the `dataSpec` parameter here. 
Besides, if we change and remove this `dataSpec` parameter, we only need to call this function once in the pipeline, making the code cleaner and less error prone.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +568,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+        :param dataSpec: Data specification. Could be one of:
+                         [primary, secondary, parent, output]
         :return:         The workflow object
         """
         currPline = wflow['PlineMarkers'][-1]
-        # Find all the output placement rules created by the agents
-        for dataCont in wflow['OutputDatasets']:
+
+        # Determine the proper container list we are about to work with:
+        contList = None
```
amaltaro, line 604: This line could be removed.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,18 +568,60 @@ def getMSOutputTransferInfo(self, wflow):
...
+
+        # Find all the data placement rules created by the components:
+        for dataCont in contList:
+            if checkGlobalLocks and dataCont in self.globalLocks:
+                msg = "Found %s dataset: %s in GlobalLocks. NOT considering it for filling the "
+                msg += "RulesToClean list for both container and block level Rules for workflow: %s!"
+                self.logger.info(msg, dataSpec, dataCont, wflow['RequestName'])
+                continue
             if gran == 'container':
                 for rule in self.rucio.listDataRules(dataCont, account=rucioAcct):
```
amaltaro, line 626: Here I think it would be useful to know the DID and the amount of rules found to be deleted. Can you please add an info log level to this?

Same comment for the `block` granularity below, but we can log just a summary, e.g.:
```
self.logger.info("Found %d block-level rules to be deleted for container %s", blah blah)
```

amaltaro (changes requested):
Todor, I have made some more comments along the code.

Regarding MSRuleCleanerWflow module, I have not yet fully reviewed it because I'd like to see first a few unit tests testing those method's functionality. The two already existent are certainly not enough. I'd advice to make method-based unit tests as well, in addition to those testing the whole parsing in one shot.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
todor-ivanov: This  was  method created _Ad exemplum_ :  
https://github.com/dmwm/WMCore/blob/a3e1e6b623a8d8dcbafd381d252f1215658f21ad/src/python/WMCore/Services/WMStatsServer/WMStatsServer.py#L80-L86

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: No, it is not unnecessary

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Just to mention, but we used to have that  information. Now, we don't. Remember [1] ?

[1]
https://github.com/dmwm/WMCore/pull/10079

todor-ivanov: Sorry I should have said: "We used to log that  information". It is still in the rule fetched from Rucio, though, but anyway.

todor-ivanov: You have the log message now.
Done.

src/python/WMCore/Services/ReqMgr/ReqMgr.py
todor-ivanov: You got it.
Done

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: My bad, I just noticed the line below is again on the `ParentDataset` key.

However, looking at the findParent implementation, the line below (making unique elements) is not really needed, since it returns a single parent dataset for every children.
By the way, there is a flaw here because it does not account for an exception in the parentage resolution (value = None).

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Ok @amaltaro I would like to drag your attention to one particular commit here. I implemented the so requested map in this commit [1]. Need to make a strong note here: I really do not think this is the correct imlementation.  And here is why:
   * This way the code looks really ugly and cumbersome to me.
   * The function complexity rose from `O(n)` to O(n^2) because of the nested for cycles. Need to mention though, this is just in theory, because in practice we expect for only one of the rucio accounts to have more than a single container list to search through, but nonetheless this is not an excuse in my opinion.
   * Now we have hard coded an entity such as an account as a string in the code, which is supposed to be only configuration dependent. And not only that but we tied a function behavior to it, meaning any new need for expanding the set of values of the configuration parameter (in this case the set of rucio accounts we consider into the service), we need to go into the code and make a relevant change of the so hard coded list. (an error I already experienced, while forgetting about the existence of `wma_test` account in parallel with the `wma_prod` account). This was absolutely not the case up to now. 
   * Which leads to another dependence we introduce here - dependence of the code itslef on the deployment variation (int or prod). 
   * There are a lot of peculiarities and specific custom cases that needed to be taken into account in a bunch of `ifs` which make it actually much more error prone than what it was before. The very proof of that  is the fact that while I had to modify it that way I have stumbled on at least 15 exceptions, all for different hidden under water stones like the one I already mentioned
   * Thoroughgoing this heavier logic inside the function instead of keeping it up in the set of function calls we make from the pipeline is exactly the opposite of the whole idea of constructing the service the way it is constructed at the first place. More precisely, to expose the logic as high as possible in the construction, while keeping the functions to be as minimalistic as possible and loaded with as little as possible additional logic.

So all that said, I would like to have your honest opinion, do you still like the variant with the account mapping inside the function more.

[1]
https://github.com/dmwm/WMCore/pull/10224/commits/d1b396017b1086c117b6f5ade72702637196ac3f

amaltaro: > This way the code looks really ugly and cumbersome to me.
> The function complexity rose from O(n) to O(n^2) because of the nested for cycles. Need to mention though, this is just in theory, because in practice we expect for only one of the rucio accounts to have more than a single container list to search through, but nonetheless this is not an excuse in my opinion.

That's correct. However, the outer `for` loop is either a single or four elements, so in practice it's basically O(n). Now you only call this method once though, instead of the previous 3 calls with different `dataSpec`.

> Now we have hard coded an entity such as an account as a string in the code, which is supposed to be only configuration dependent. And not only that but we tied a function behavior to it, meaning any new need for expanding the set of values of the configuration parameter (in this case the set of rucio accounts we consider into the service), we need to go into the code and make a relevant change of the so hard coded list. (an error I already experienced, while forgetting about the existence of wma_test account in parallel with the wma_prod account). This was absolutely not the case up to now.

I don't think we need to hard code the account. I have made some changes on top of your latest commit such that you can see it. In addition to that, the `dataSpec` parameters were hard-coded as well... And if we need to consider a new rucio account, the source code has to change anyways, even before this PR was created.

> Which leads to another dependence we introduce here - dependence of the code itslef on the deployment variation (int or prod).

This should be automatically taken care of with the `self.msConfig` object.

> Thoroughgoing this heavier logic inside the function instead of keeping it up in the set of function calls we make from the pipeline is exactly the opposite of the whole idea of constructing the service the way it is constructed at the first place. More precisely, to expose the logic as high as possible in the construction, while keeping the functions to be as minimalistic as possible and loaded with as little as possible additional logic.

I can't really evaluate the tradeoff here between a) making many function calls with different arguments (expected arguments in the method) and executing the correct logic based on that; or b) making one function call with less arguments, and deciding the logic to be executed according an object attribute.
The logic though, is pretty small, the whole method source code is less than 40 lines (and only because you decided to deal with containers and blocks in different ways, otherwise it would be even smaller ;)).

As said above, I took your code (4th commit) and made some modifications on top of it, which I consider to simplify things a bit and might look better to you now:
```
...
        currPline = wflow['PlineMarkers'][-1]

        # Create the container list to the rucio account map and set the checkGlobalLocks flag.
        mapRuleType = {self.msConfig['rucioWmaAccount']: ["OutputDatasets"],
                       self.msConfig['rucioMStrAccount']: ["InputDataset", "MCPileup",
                                                           "DataPileup", "ParentDataset"]}
        if rucioAcct == self.msConfig['rucioMStrAccount']:
            checkGlobalLocks = True
        else:
            checkGlobalLocks = False

        # Find all the data placement rules created by the components:
        for dataType in mapRuleType[rucioAcct]:
            for dataCont in wflow[dataType]:
                self.logger.debug("getRucioRules: dataCont: %s", pformat(dataCont))
                if checkGlobalLocks and dataCont in self.globalLocks:
...
```

If you still think your initial implementation looks more robust, then please drop the 4th commit and I apologize for wasting your time.

todor-ivanov: @amaltaro your code is breaking [2], just as I expected, because you are not taking into account the fact that not all of those data fields are lists any more. The `InputDataset` is a str. Notice that the code is trying to call rucio for every single letter in the dataset name. I was homogenizing (on puropuse) the structure of all those fields to lists in an earlier stage of this review, but you claimed this was bad [1]), I did the cahnge relentlesly, but from there on, we need to carry and remember that difference on the following steps. So I will have to think of a way out of this here at place.

[1]
https://github.com/dmwm/WMCore/pull/10224#discussion_r572978751

[2]

```
...
021-02-12 17:40:54,339:DEBUG:MSRuleCleaner:getRucioRules(): getRucioRules: dataCont: u'1'
2021-02-12 17:40:54,347:DEBUG:connectionpool:_make_request(): http://cmsrucio-int.cern.ch:80 "GET /rules/?scope=cms&account=wmcore_transferor&name=1 HTTP/1.1" 200 None
2021-02-12 17:40:54,348:DEBUG:MSRuleCleaner:getRucioRules(): getRucioRules: dataCont: u'/'
2021-02-12 17:40:54,359:DEBUG:connectionpool:_make_request(): http://cmsrucio-int.cern.ch:80 "GET /rules/?scope=cms&account=wmcore_transferor&name=%2F HTTP/1.1" 200 None
2021-02-12 17:40:54,360:DEBUG:MSRuleCleaner:getRucioRules(): getRucioRules: dataCont: u'R'
2021-02-12 17:40:54,369:DEBUG:connectionpool:_make_request(): http://cmsrucio-int.cern.ch:80 "GET /rules/?scope=cms&account=wmcore_transferor&name=R HTTP/1.1" 200 None
2021-02-12 17:40:54,370:DEBUG:MSRuleCleaner:getRucioRules(): getRucioRules: dataCont: u'E'
2021-02-12 17:40:54,378:DEBUG:connectionpool:_make_request(): http://cmsrucio-int.cern.ch:80 "GET /rules/?scope=cms&account=wmcore_transferor&name=E HTTP/1.1" 200 None
2021-02-12 17:40:54,379:DEBUG:MSRuleCleaner:getRucioRules(): getRucioRules: dataCont: u'C'
2021-02-12 17:40:54,389:DEBUG:connectionpool:_make_request(): http://cmsrucio-int.cern.ch:80 "GET /rules/?scope=cms&account=wmcore_transferor&name=C HTTP/1.1" 200 None
2021-02-12 17:40:54,389:DEBUG:MSRuleCleaner:getRucioRules(): getRucioRules: dataCont: u'O'
...

2021-02-12 17:43:54,496:ERROR:MSRuleCleaner:_dispatchWflow(): plineMSTrBlock: General error from pipeline. Workflow: tivanov_TaskChain_InclParents_HG2102_Val_210209_001757_1463. Error:  
An unknown exception occurred.
Details: no error information passed (http status code: 504 ('gateway_timeout',)). 
Will retry again in the next cycle.
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/WMCore/MicroService/Unified/MSRuleCleaner.py", line 315, in _dispatchWflow
    pline.run(wflow)
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 140, in run
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 140, in <lambda>
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 72, in __call__
    return self.run(obj)
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 75, in run
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/WMCore/MicroService/Unified/MSRuleCleaner.py", line 642, in getRucioRules
    blocks = self.rucio.getBlocksInContainer(dataCont)
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/WMCore/Services/Rucio/Rucio.py", line 229, in getBlocksInContainer
    if not self.isContainer(container):
  File "/afs/cern.ch/user/t/tivanov/work/private/WMCoreDev.d/WMCore/src/python/WMCore/Services/Rucio/Rucio.py", line 824, in isContainer
    response = self.cli.get_did(scope=scope, name=didName)
  File "/afs/cern.ch/work/t/tivanov/private/WMCoreDev.d/WMCore.venv/lib/python2.7/site-packages/rucio/client/didclient.py", line 394, in get_did
    raise exc_cls(exc_msg)
RucioException: An unknown exception occurred.
Details: no error information passed (http status code: 504 ('gateway_timeout',))
```

amaltaro: Right! This logic considers the docSchema template:
https://github.com/dmwm/WMCore/pull/10224/files#diff-414c04814f1133dc334c7718010c11e70809389d23ec30e9d00d5e8874b84bf5R192

I guess you missed updating it to `str` over there then? Right now it's still meant to be of a `list` type.

amaltaro: Ohh, I was looking at the docstring instead of the source code. Without reading the rest of the code, I don't know what is the expected data type for those new parameters. 
However, I think we define call InputDataset/ParentDataset as a string, MCPileup/DataPileup as a list (of strings).

todor-ivanov: Hi @amaltaro, At least for ParentDataaset I would prefer to keep it as a list. It is already a list and is used as such in other places already too. Also, when we resolve the parent datasets from DBS the function `findParent()` returns a list. 

And I actually, do not see a reason why should we have some of the fields which we use to store dataset names in strings and some in lists. I know that those you suggest to be strings if we keep them as lists they can never become lists longer than a single element, but this I would say is not that bad. If we keep juggling with two different types of structures then in every function we have to take this into account, while if we convert them all to lists from the very beginning then the rest of the code simplifies extremely - we just aggregate the total list of datasets a function must work on, just by chaining all the already constructed lists. Or even better, we do not need to change a function behavior just because it was called to work on `wflow['InputDataset']` instead of `wflow['MCPileup']` or `wflow['OutputDatasets']`. The other way forces us to account for these data structure differences based on the type of input data every time we try to iterate through these workflow keys related to keeping dataset names.

todor-ivanov: Hi @amaltaro I just force pushed your suggested code change plus a little modification, which we should apply every time if we keep the data structures indifferent forms. Please take a look again.  I am fine anyway.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: I knew which line you were talking about. Again, it is needed, and not because the `findParents()` returns more than one dataset, but because we may call it twice, and we do not want to accumulate in the 'ParentDataset' list the same dataset which will show up twice.

todor-ivanov: Hi @amaltaro , About the returned `value=None`, I just made a commit that should address this.

amaltaro:
> Regarding MSRuleCleanerWflow module, I have not yet fully reviewed it because I'd like to see first a few unit tests testing those method's functionality. The two already existent are certainly not enough. I'd advice to make method-based unit tests as well, in addition to those testing the whole parsing in one shot.

@todor-ivanov let me try to be very specific on what I think we should have implemented in the unit tests here. So far we only have 2 tests for the MSRuleCleanerWflow module:
* with an empty request dictionary (thus testing default values)
* a simple TaskChain request

if we want to make sure the workflow parsing code works as expected, we should test at least a few more use cases, such as:
* a dictionary with multiple pileup datasets (even better if there are duplicates as well);
* a flat dictionary (like a ReReco request);
* a dictionary requesting parents

can you please implement those tests? Please avoid using those json templates, because whenever we change them, we have to update the test code as well. Instead, simply use the relevalent key/value pairs to exercise that code.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -73,7 +209,12 @@ def docSchema(self):
             ('IsLogDBClean', False, bool),
             ('IsArchivalDelayExpired', False, bool),
             ('ForceArchive', False, bool),
-            ('RequestTransition', [], list)]
+            ('RequestTransition', [], list),
+            ('IncludeParents', False, bool),
+            ('DataPileup', None, (str, unicode)),
```
amaltaro, line 214: And I just noticed these default values don't match the default values in the unit test `testTaskChainDefaults`.
DataPileup, MCPileup and ParentDataset are inconsistent.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -54,7 +182,15 @@ def docSchema(self):
...
             'IsClean': False
-            'ForceArchive', False]
+            'IsLogDBClean': False,
+            'IsArchivalDelayExpired': False,
+            'ForceArchive': False,
+            'RequestTransition': [],
+            'IncludeParents': False
+            'DataPileup': [],
+            'MCPileup': [],
+            'InputDataset': 'None',
```
amaltaro, line 192: Did you mean `None` instead of `'None'`?

todor-ivanov: Yes, indeed.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: Yes, I have not yet started paying attention to the final version of the unit tests. I was about to do that in a later stage. Once we agree on the implementation of all the rest. Because we kind of depend on that for the unit tests step.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: I had to force push one more commit in adjacent to the above mentioned. I was not handling the exception I was raising and the standard long error message from WMCoreException was flooding the logs.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,30 +575,75 @@ def getMSOutputTransferInfo(self, wflow):
...
+        :param  wflow:   A MSRuleCleaner workflow representation
+        :return:         The workflow object
+        """
+        if wflow['InputDataset'] and wflow['IncludeParents']:
+            childDataset = wflow['InputDataset']
+            parentDataset = findParent([childDataset], self.msConfig['dbsUrl'])
+            # NOTE: If findParent() returned None then the DBS service failed to
+            #       resolve the request (it is considered an ERROR outside WMCore)
+            if parentDataset is None:
+                msg = "Failed to resolve parent dataset for: %s in wflow: %s" % (childDataset, wflow['RequestName'])
```
amaltaro, line 590: I would suggest to write the `wflow` word as a whole, thus `workflow`.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: Just to make sure I understand your point, are you saying that we can call this method `setParentDatasets` multiple times for the same workflow within the same cycle? That's the only explanation to have a list here.

The `findParent` function returns a string as value:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/Common.py#L651

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -34,6 +34,17 @@
 from Utils.Pipeline import Pipeline, Functor
 from WMCore.WMException import WMException
 from WMCore.Services.LogDB.LogDB import LogDB
+from WMCore.Services.WMStatsServer.WMStatsServer import WMStatsServer
+from WMCore.MicroService.Unified.Common import findParent
+
+
+class MSRuleCleanerResolveParentError(WMException):
+    """
+    resolve Parent Exception Class for MSRuleCleaner Module in WMCore MicroServices
```
amaltaro, line 43: I feel bad saying it, but you won't know unless someone says it...
Can I ask you to use capital letters only where it makes sense? One day we shall have it all automatically documented and available, and we all like reading things nicely written ;)

To be clear, my sentence would be:
```
WMCore exception class for the MSRuleCleaner module, in WMCore MicroServices,
used to signal if an error occurred during parent dataset resolution step.
```

This should be considered anywhere, be it a comment, docstring, log record, etc.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,30 +575,75 @@ def getMSOutputTransferInfo(self, wflow):
...
+                       self.msConfig['rucioMStrAccount']: ["InputDataset", "MCPileup",
+                                                           "DataPileup", "ParentDataset"]}
+        if rucioAcct == self.msConfig['rucioMStrAccount']:
+            checkGlobalLocks = True
+        else:
+            checkGlobalLocks = False
+
+        # Find all the data placement rules created by the components:
+        for dataType in mapRuleType[rucioAcct]:
+            for dataCont in wflow[dataType] if isinstance(wflow[dataType], list) else [wflow[dataType]]:
```
amaltaro, line 624: Isn't it possible that `wflow[dataType]` is an empty string? Thus getting into this for loop and failing with a rucio query for container `""`?

By the way, I think this line would be more readable if we break it down in multiple lines.

amaltaro (commented):
Todor, I have made some further comments. I guess they are all minor and I think you can proceed and finish this PR off.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Yes, this i exactly what I am saying. 

>The findParent function returns a string as value:

`findParent` returns either a dictionary with keys names set as the child dataset name and values set as the primary dataset (the value could eventually be None if the child dataset has no parent discovered), or `None` instead of a dictionary in the cases when the DBS service fails to answer the query. I hope I am not missing something here...

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: No, it should never be an empty string. If it turns out to be such it means either the key value from the top level workflow representation coming from reqmgr is an empty string (which should never happen) or `findParent` call returned an empty string (which is also not possible) or some of the keys that we call this function with has a default value as `''` in `MSRuleCleanerWflow`, which is also not the case. BUT.... while speaking of that, I just realized (and tested it), we may end up with a `None` value instead of an empty string. Which I think could be equally bad regarding the Rucio call. But again if this happens I would say this is due to a different bug - the conditions under which it may happen is only if someone configures a workflow to have `IncludeParents` flag set to `True` and a value of `None` for `InputDataset`. I need to admit though haven't check if `StdBase` allows this.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: I think we agree on that. Being explicit, findParent() is meant to return data formatted as:
```
{"child_dataset_nameA": "parent_dataset_name",
 "child_dataset_nameB": "parent_dataset_nameX", etc}
```

with such data, these 2 lines you have in the current code:
```
wflow['ParentDataset'].append(parentDataset[childDataset])
wflow['ParentDataset'] = list(set(wflow['ParentDataset']))
```

could be converted to one line:
```
wflow['ParentDataset'] =[parentDataset[childDataset]]
```

Your comment also made me realize a mistake on the way the code parses errors/exceptions in the findParent() function. Shouldn't this if statement:
```
if parentDataset is None:
```

be actually converted to:
```
if not parentDataset.get(childDataset):
```
?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
amaltaro: >if someone configures a workflow to have IncludeParents flag set to True and a value of None for InputDataset

there is a protection against it at the Specs level.

todor-ivanov: And BTW, I see no easy way to protect the code here from such workflow configuration. What should I see do here? (this is supposed to be a rhetorical question ;) ) But joke aside, such a workflow configuration should be treated as a separate use case for sure, much further in the logic not here.

I hope I am not missing any other possibility. If you think of one ,now is the moment for us to take precautions.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: > `wflow['ParentDataset'] = [parentDataset[childDataset]]`

Agreed

> `if parentDataset is None:`
This line I wrote because like that , because I wanted to threat separately the cases when `findParent` returns just a   `None` value instead of the normal dictionary, which is a sign for a DBS service error, but not a workflow misconfiguration or a WMCore error. In those cases we should skip this workflow and postpone its treatment/cleaning until next poling cycle.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: I think this  one is working and in the shape you wanted it. I am about to resolve this conversation.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: Done

todor-ivanov:
Hi @amaltaro, Is there anything else you thing needs to be taken care of regarding this PR?

amaltaro:
Hi Todor, I haven't seen another review request, so I assumed you were still working on this PR. I will have another look into it soon then, meanwhile, can you please update the PR description and create those deployment changes everywhere it's needed?

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
amaltaro: Isn't the DataPileup/MCPileup/ParentDataset default value supposed to be a list instead of None?

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,30 +575,75 @@ def getMSOutputTransferInfo(self, wflow):
...
 
+    def setParentDatasets(self, wflow):
+        """
+        Used to resolve parent datasets for a workflow.
+        :param  wflow:   A MSRuleCleaner workflow representation
+        :return:         The workflow object
+        """
+        if wflow['InputDataset'] and wflow['IncludeParents']:
+            childDataset = wflow['InputDataset']
+            parentDataset = findParent([childDataset], self.msConfig['dbsUrl'])
```
amaltaro, line 586: Todor, haven't we discussed that this API will return:
```
parentDataset = {"pileup_dset_name": None}
```
in case of errors?

In such case, the if statement below would fail to catch those errors and would actually append `None` to a list of parent datasets.

amaltaro (changes requested):
Todor, I think this code is almost ready to go. I made two comments in the code though which I think we need to update the code for.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: @amaltaro I need to understand, what are the conditions under which the following line would be hit [1], and what is the difference from the other error here [2], so I can see if I should treat them separately. 

[1]
https://github.com/dmwm/WMCore/blob/e049d6cf05e49e99843226874a6587c7c760adf0/src/python/WMCore/MicroService/Unified/Common.py#L646

[2]
https://github.com/dmwm/WMCore/blob/e049d6cf05e49e99843226874a6587c7c760adf0/src/python/WMCore/MicroService/Unified/Common.py#L654

amaltaro: I believe a DBS/CMSWEB unavailable would fall under the case [1]. Case isn't clear to me, I guess only if the data object isn't iterable or something like that.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: I think everything is as it should be, you just forget about this:

https://github.com/dmwm/WMCore/pull/10224/files/2f4a6275b7e0af14e3f42aabbca6f40470503ca9#diff-414c04814f1133dc334c7718010c11e70809389d23ec30e9d00d5e8874b84bf5R149-R152

amaltaro: That's precisely my point. This data structure defines the schema of the workflow object used throughout MSRuleCleaner. By just looking at this, I'd expect those keys to have a value equal to `None`, not an empty list.

Does the whole logic - besides that explicit code you just referred to - change if you set default value to `[]` instead of None? If not and we get that for free, then I'd suggest to update the default value to make it consistent with the outcome of this workflow parser.

todor-ivanov: It should work, but what would happen is that this conversion from a string to list will happen silently in the workflow parser class here: [1] instead of explicitly in the three lines I previously mentioned. I have tested it both ways, and now I do not remember which exactly but there was a corner case in which the silent conversion of the data structure was not preferable. So I preferred in the `-wfParser()` method to make the best effort to revert the structure back to the original one as described in the document template (which reflects the values as we expect them from reqmgr), and only then make the explicit conversion to lists only for those I really want and make them obvious in the `init()` method of MSRuleCleanerWflow.

[1]
https://github.com/dmwm/WMCore/pull/10224/files/2f4a6275b7e0af14e3f42aabbca6f40470503ca9#diff-414c04814f1133dc334c7718010c11e70809389d23ec30e9d00d5e8874b84bf5R110-R111

todor-ivanov: Actually, I just tested it again.... And not even a corner case but a rather general one. It tries to create a list of all letters present in all the MCPilup dataset names it can find and explodes.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
todor-ivanov: Done

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
amaltaro: Okay, let it be then. Thanks for checking this out Todor.

src/python/WMCore/MicroService/Unified/MSRuleCleaner.py
```diff
@@ -528,30 +575,75 @@ def getMSOutputTransferInfo(self, wflow):
...
+            # NOTE: If findParent() returned None then the DBS service failed to
+            #       resolve the request (it is considered an ERROR outside WMCore)
+            if parentDataset.get(childDataset, None) is None:
+                msg = "Failed to resolve parent dataset for: %s in workflow: %s" % (childDataset, wflow['RequestName'])
+                raise MSRuleCleanerResolveParentError(msg)
+            elif parentDataset:
+                wflow['ParentDataset'] = [parentDataset[childDataset]]
+                msg = "Found parent %s for input dataset %s in workflow: %s "
+                self.logger.info(msg, parentDataset, wflow['InputDataset'], wflow['RequestName'])
+            else:
```
amaltaro, line 596: This else block is no longer needed. But that's a minor and I won't ask you to update this code once again.

amaltaro (approved):
Code looks good to me. Thanks

amaltaro:
@todor-ivanov Todor, given that this PR was merged without the relevant deployment changes in place. I would really appreciate if you can update this PR description pointing to all the deployment/configuration related changes in other repositories. Thanks

todor-ivanov:
Hi @amaltaro, You have them.

PR #10226, opened 2021-01-14
add option for preloader for xrdcp stageout

hufnagel:
For stageout from certain HPC we want to have a preloader in front of every call to a TCP application to tunnel the traffic. Allow to configure this in SITECONF.

#### Status
ready

#### Description
Adds an option for a preloader in front of every xrdcp and xrdfs call that can be specified in the site-local-conf stageout section.

YES (only sites that set the new option will see any effect)

#### Related PRs
NONE

NONE

### DIFF
--- a/src/python/WMCore/Storage/Backends/XRDCPImpl.py
+++ b/src/python/WMCore/Storage/Backends/XRDCPImpl.py
@@ -26,6 +26,7 @@ class XRDCPImpl(StageOutImpl):
         StageOutImpl.__init__(self, stagein)
         self.numRetries = 5
         self.retryPause = 300
+        self.xrdfsCmd = "xrdfs"
 
     def createSourceName(self, protocol, pfn):
         """
@@ -61,6 +62,7 @@ class XRDCPImpl(StageOutImpl):
         parser.add_argument('--wma-cerncastor', action='store_true')
         parser.add_argument('--wma-old', action='store_true')
         parser.add_argument('--wma-disablewriterecovery', action='store_true')
+        parser.add_argument('--wma-preload')
         args, unknown = parser.parse_known_args(options.split())
 
         # strip out WMAgent specific options
@@ -78,7 +80,7 @@ class XRDCPImpl(StageOutImpl):
             if args.wma_cerncastor:
                 targetPFN += "?svcClass=t0cms"
 
-        useChecksum = (checksums != None and 'adler32' in checksums and not self.stageIn)
+        useChecksum = (checksums is not None and 'adler32' in checksums and not self.stageIn)
 
         xrdcpExec = "xrdcp"
         if args.wma_old:
@@ -115,7 +117,14 @@ class XRDCPImpl(StageOutImpl):
         if args.wma_disablewriterecovery:
             copyCommand += "env XRD_WRITERECOVERY=0 "
 
-        copyCommand += "%s --force --nopbar " % xrdcpExec
+        if args.wma_preload:
+            xrdcpCmd = "%s %s" % (args.wma_preload, xrdcpExec)
+            self.xrdfsCmd = "%s xrdfs" % args.wma_preload
+        else:
+            xrdcpCmd = xrdcpExec
+            self.xrdfsCmd = "xrdfs"
+
+        copyCommand += "%s --force --nopbar " % xrdcpCmd
 
         if copyCommandOptions:
             copyCommand += "%s " % copyCommandOptions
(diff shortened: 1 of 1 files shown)


src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -133,18 +138,19 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
 
         (_, host, path, _) = self.splitPFN(remotePFN)
 
+        removeCommand = ""
```
amaltaro, line 141: This line seems to not be necessary.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -115,7 +117,10 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
         if args.wma_disablewriterecovery:
             copyCommand += "env XRD_WRITERECOVERY=0 "
 
-        copyCommand += "%s --force --nopbar " % xrdcpExec
+        if args.wma_preload:
+            self.preload = "%s " % args.wma_preload
+
+        copyCommand += "%s%s --force --nopbar " % (self.preload, xrdcpExec)
```
amaltaro, line 123: I think all these commands would look cleaner if we add an extra space in every single command, instead of stucking two format strings altogether. So, something like:
```
copyCommand += "%s %s --force --nopbar " % (self.preload, xrdcpExec)
```
and the same to all the other changes below.

That also means there is no need for the extra space above in L121.

amaltaro (commented):
Thanks for providing this functionality, Dirk.

Code looks good to me, however, if you also agree with me that a command like (or source code):
```
"%s xrdfs %s rm %s"
```

goes much better with the eyes/brain, than
```
"%sxrdfs %s rm %s"
```

then please update this code.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
hufnagel: If you add the extra space you break the unit tests since the "no preloader set" option now looks different from before (it has the empty space).

hufnagel: Only other option I see is to have a conditional for every line that includes the preloader and make one version with preloader and one version without.

hufnagel: Or we merge ignoring the unit test failures since after the merge the unit tests should work again (the version with the empty space becomes the new default).

amaltaro: Perhaps we could return the stage out command like:
```
return copyCommand.strip()
```
? That should resolve the unit test failures.

hufnagel:
Unit tests pass and I think this looks cleaner than before.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
```diff
@@ -115,7 +117,14 @@ def createStageOutCommand(self, sourcePFN, targetPFN, options=None, checksums=No
         if args.wma_disablewriterecovery:
             copyCommand += "env XRD_WRITERECOVERY=0 "
 
-        copyCommand += "%s --force --nopbar " % xrdcpExec
+        if args.wma_preload:
+            xrdcpCmd = "%s %s" % (args.wma_preload, xrdcpExec)
+            self.xrdfsCmd = "%s xrdfs" % args.wma_preload
+        else:
+            xrdcpCmd = xrdcpExec
+            self.xrdfsCmd = "xrdfs"
```
amaltaro, line 125: This line isn't really needed.

amaltaro (commented):
Thanks Dirk. Code looks better indeed. I left only one more comment for improvement.

I'm applying this patch - as is - to the hepcloud agent (cmssrv217) just so we can speed up the testing process.

src/python/WMCore/Storage/Backends/XRDCPImpl.py
hufnagel: It's redundant, yes. Makes the code a bit easier to read and understand IMO.

amaltaro: Still it's an unneeded code duplication. Anyways, not a big deal.

amaltaro:
I remember you saying it works fine, even though there was a unrelated issue at those resources. So I'm merging it.
Please let me know if any follow up is required.

PR #10250, opened 2021-02-02
[py2py3] pycurl_manager - better handling of curl.setopts() input arguments

### LINKED ISSUE #10249: [py2py3] pycurl_manager - better handling of curl.setopts() input arguments
**Impact of the new feature**

WMCore.Services.pycurl_manager

When modernizing `pycurl_manager` in #9818 , I overlooked the pycurl API documentation and relied only on the unit tests.

While investigating `pycurl_manager` for https://github.com/dmwm/CRABServer/issues/6374 I realized that the input of `curl.setopts()` should be a bytes string, while we are not passing a unicode string. It is not a problem until we only pass URLs that do not contain non-ascii characters, as stated in the [pycurl docs](http://pycurl.io/docs/7.43.0.3/curlobject.html?highlight=setopt#pycurl.Curl.setopt), but this is not the optimal long term solution. 

We should not use `curl.setopts(..., str(url))` in py3 (never) and in py2 (only when importing `from builtins import str`, which is our current case). We should make sure that url is passed as a bytes encoded string. Maybe we can add the inverse of https://github.com/dmwm/WMCore/blob/ebfc5159d0245ad41be1701be450b106425cfcc1/src/python/Utils/Utilities.py#L195 to Utils/Utilities.py and call it `encodeUnicodeToBytes` and use it here.

We can run with our current code and hope that our URLs will never contain non-ascii characters (a bit risky)

**Additional context**

[This discussion](https://github.com/dmwm/CRABServer/issues/6374) may contain some more information.

mapellidario:
Fixes #10249 

#### Status
ready

#### Description

Quickly encode the `url` variable if it is a futurized string (if it is unicode, then) before passing it to `curl.setopt()`.

Should be better backwrd compatible than what we have now, which breaks old pycurl versions and urls that contain non-ascii characters.

Same as before

### DIFF
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -224,6 +224,34 @@ def decodeBytesToUnicode(value, errors="strict"):
     return value
 
 
+def encodeUnicodeToBytes(value, errors="strict"):
+    """
+    Accepts an input "value" of generic type.
+
+    If "value" is a string of type sequence of unicode (i.e. in py2 `unicode` or
+    `future.types.newstr.newstr`, in py3 `str`), then it is converted to
+    a sequence of bytes.
+
+    This function is useful for encoding output data when using the
+    "unicode sandwich" approach, which involves converting unicode (i.e. strings
+    of type sequence of unicode codepoints) to bytes (i.e. strings of type
+    sequence of bytes, in py2 `str` or `future.types.newbytes.newbytes`,
+    in py3 `bytes`) as late as possible when passing a string to a third-party
+    function that only accepts bytes as input (pycurl's curl.setop is an
+    example).
+    py2:
+    - "errors" can be: "strict", "ignore", "replace", "xmlcharrefreplace"
+    - ref: https://docs.python.org/2/howto/unicode.html#the-unicode-type
+    py3:
+    - "errors" can be: "strict", "ignore", "replace", "backslashreplace", 
+      "xmlcharrefreplace", "namereplace"
+    - ref: https://docs.python.org/3/howto/unicode.html#the-string-type
+    """
+    if isinstance(value, str):
+        return value.encode("utf-8", errors)
+    return value
+
+
 # TODO: remove this function once we have completely migrated to Rucio
 def usingRucio():
     """
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -44,6 +44,7 @@ from builtins import str, range, object
 from past.builtins import basestring
 from future.utils import viewitems
 
+
 # system modules
 import json
 import logging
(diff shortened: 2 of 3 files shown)


belforte:
great Dario ! I'll give this a try, but would not be surprised if we hit same problem with another option.
We may still want a more solid solution for our environment, but being able to run things will allow to progress on the more general CRAB vs WMCore/master vs python3.

belforte:
I found that I need this change as well (could possibly be coded in a better way). But have not managed to exercise all calls yet.

```
         if headers:
-            curl.setopt(pycurl.HTTPHEADER, \
-                    ["%s: %s" % (k, v) for k, v in viewitems(headers)])
+            headersForCurl = ["%s: %s" % (k, v) for k, v in viewitems(headers)]
+            newHeaders = [ h.encode("utf-8") for h in headersForCurl ]
+            curl.setopt(pycurl.HTTPHEADER, newHeaders)


```

belforte:
status update. WIth
1. this changes in pycurl_manger
2. future installed via pip in system python
3. no use of /cvmfs/..COMP
I can at least submit jobs (i.e. crab taskworker and famous DAGMAN on schedd run).
Dario, I suggest that you keep this open until we have managed to exercise all pycurl calls, in case there's some other options which need to be cast into a list of bytes.

amaltaro:
I'm about to test it with WMAgent as well.
Dario, if you need to change something, please add it as an extra commit (for my own information, testing the first 3), which is what you already do I think ;)

amaltaro:
If you don't ask for my review, I ask it myself :-P

amaltaro (approved):
Changes look good to me, Dario! I'll soon check my tests and if everything went fine, it's ready to be merged.

Should we merge it with 3 commits, such as:
1. providing the auxiliary function in Utilities
2. providing the pycurl fix
3. providing the unit test changes
?

If it's too late for that, then can you please squash the src code changes?

mapellidario:
@amaltaro I did not ask for your review because stefano suggested waiting since it is possible that his tests will spot some other problems.

In any case, I can rewrite the git history as you like!

belforte:
I tested as much as I could. Ref:  https://github.com/dmwm/CRABServer/issues/6373#issuecomment-773110744 and https://github.com/dmwm/CRABServer/issues/6374#issuecomment-773108441
From my POV you can merge and tag whenever you are comfortable. If later on we find some corner case, we'll open a new issue.

amaltaro:
WMAgent-side tests went fine as well. Thanks for the prompt fix, Dario.

amaltaro:
@belforte Stefano, tag `1.4.6.pre2` contains this fix in. Please check out the CHANGES file to see what else went in in that tag: https://github.com/dmwm/WMCore/blob/master/CHANGES

belforte:
thanks @amaltaro and @mapellidario  everything looks good. I will run with this tag now.

PR #10252, opened 2021-02-03
Trap SSL_connect error and Service Unavailable in DBSUploadPoller

### LINKED ISSUE #10000: DBS3Upload crashes with HTTP Error 503
**Impact of the bug**
WMAgent

**Describe the bug**
DBS3Upload crashes if it receives a response like:
```
<center><h1>503 Service Temporarily Unavailable</h1></center>
```
while talking to CMSWEB services, in this case retrieving the parentage chain from wmstatsserver.

**How to reproduce it**
not sure

**Expected behavior**
Catch this specific error message:
```
503 Service Temporarily Unavailable
```
when accessing wmstatsserver and simply skip that components cycle, which will be retried again a few minutes later.

Similar to issue: https://github.com/dmwm/WMCore/issues/9123

I don't remember seeing this problem with the production cluster, but it happens quite often with the CMS@Home agent. Traceback is:
```
2020-10-17 17:53:41,216:139720193722112:INFO:DBSUploadPoller:Starting the DBSUpload Polling Cycle
2020-10-17 17:53:41,265:139720193722112:WARNING:Service:The cachefile /data/srv/wmagent/v1.4.1.patch1/install/wmagent/DBS3Upload/.wmcore_cache/.wmcore_cache_31961/requests/cmsweb-testbed.cern.ch:8443/15744094587142094_GET_filtered_requ
ests?RequestType=StepChain&ParentageResolved=False&mask=ChainParentageMap does not exist and the service at https://cmsweb-testbed.cern.ch:8443/wmstatsserver/data/filtered_requests?RequestType=StepChain&ParentageResolved=False&mask=Cha
inParentageMap is unavailable - it returned 503 because Service Temporarily Unavailable with result: <html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>nginx/1.17.10</center>
</body>
</html>


2020-10-17 17:53:41,349:139720193722112:ERROR:BaseWorkerThread:Error in worker algorithm (1):
Backtrace:
  <WMComponent.DBS3Buffer.DBSUploadPoller.DBSUploadPoller object at 0x7f132bcc12d0> <@========== WMException Start ==========@>
Exception Class: DBSUploadException
Message: Unknown failure while fetching parentage map from WMStats. Error: url=https://cmsweb-testbed.cern.ch:8443/wmstatsserver/data/filtered_requests?RequestType=StepChain&ParentageResolved=False&mask=ChainParentageMap, code=503, rea
son=Service Temporarily Unavailable, headers={'Content-Length': '198', 'Server': 'nginx/1.17.10', 'Connection': 'close', 'Date': 'Sat, 17 Oct 2020 15:53:41 GMT', 'Content-Type': 'text/html', 'CMS-Server-Time': 'D=3087 t=160295002126097
6'}
        ModuleName : WMComponent.DBS3Buffer.DBSUploadPoller
        MethodName : updateDatasetParentageCache
        ClassInstance : None
        FileName : /data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMComponent/DBS3Buffer/DBSUploadPoller.py
        ClassName : None
        LineNumber : 357
        ErrorNr : 0

Traceback: 
  File "/data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMComponent/DBS3Buffer/DBSUploadPoller.py", line 345, in updateDatasetParentageCache
    self.datasetParentageCache = self.wmstatsServerSvc.getChildParentDatasetMap()

  File "/data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMCore/Services/WMStatsServer/WMStatsServer.py", line 105, in getChildParentDatasetMap
    results = self.getFilteredActiveData(filter, mask)

  File "/data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMCore/Services/WMStatsServer/WMStatsServer.py", line 92, in getFilteredActiveData
    return self._getResult(callname, verb="GET")

  File "/data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMCore/Services/WMStatsServer/WMStatsServer.py", line 43, in _getResult
    verb=verb, contentType=contentType)

  File "/data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMCore/Services/Service.py", line 208, in refreshCache
    self.getData(cachefile, url, inputdata, incoming_headers, encoder, decoder, verb, contentType)

  File "/data/srv/wmagent/v1.4.1.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.4.1.patch1/lib/python2.7/site-packages/WMCore/Services/Service.py", line 323, in getData
    raise he
```

amaltaro:
Fixes #10000 

#### Status
ready

#### Description
To avoid having DBS3Upload crashing all the time that there are instabilities with the CMSWEB cluster (like now with the frontends), catch the following 2 exceptions/errors and gracefully skip the component cycle:
* "Service Temporarily Unavailable"
* "OpenSSL SSL_connect"

the second error might be too generic, but if we have any serious SSL issues in the agent, everything else should break.

In addition to that, you can see that I removed those 2 functions that were apparently not used anywhere in WMCore nor in T0.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
@@ -48,38 +48,6 @@ from WMCore.WMException import WMException
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 
 
-def createConfigForJSON(config):
-    """
-    Turn a config object into a dictionary of dictionaries
-
-    """
-
-    final = {}
-    for sectionName in config.listSections_():
-        section = getattr(config, sectionName)
-        if hasattr(section, 'dictionary_'):
-            # Create a dictionary key for it
-            final[sectionName] = createDictionaryFromConfig(section)
-
-    return final
-
-
-def createDictionaryFromConfig(configSection):
-    """
-    Recursively create dictionaries from config
-
-    """
-
-    final = configSection.dictionary_()
-
-    for key in final.keys():
-        if hasattr(final[key], 'dictionary_'):
-            # Then we can turn it into a dictionary
-            final[key] = createDictionaryFromConfig(final[key])
-
-    return final
-
-
 def uploadWorker(workInput, results, dbsUrl):
     """
     _uploadWorker_
@@ -141,11 +109,35 @@ def uploadWorker(workInput, results, dbsUrl):
     return
 
 
+def isPassiveError(exceptionObj):
+    """
+    This function will parse the exception object and report whether
+    the error message corresponds to a soft or hard error (hard errors
+    are supposed to let the component crash).
+    :param exceptionObj: any exception object
+    :return: True if it's a soft error, False otherwise
+    """
+    passException = True
+    passiveErrorMsg = ['Service Unavailable', 'Service Temporarily Unavailable',
+                       'Proxy Error', 'Error reading from remote server',
+                       'Connection refused', 'timed out', 'Could not resolve',
+                       'OpenSSL SSL_connect: SSL_ERROR_SYSCALL']
+
+    excReason = getattr(exceptionObj, 'reason', '')
+    for passiveMsg in passiveErrorMsg:
+        if passiveMsg in excReason:
+            break
+        elif passiveMsg in str(exceptionObj):
+            break
+    else:
+        passException = False
+    return passException
+
+
 class DBSUploadException(WMException):
     """
     Holds the exception info for
     all the things that will go wrong
-
     """
 
 
@@ -344,19 +336,16 @@ class DBSUploadPoller(BaseWorkerThread):
         try:
             self.datasetParentageCache = self.wmstatsServerSvc.getChildParentDatasetMap()
         except Exception as ex:
-            reason = getattr(ex, 'reason', '')
-            msg = 'Failed to fetch parentage map from WMStats, skipping this cycle. '
-            msg += 'Exception: %s. Reason: %s. ' % (type(ex).__name__, reason)
-            if 'Service Unavailable' in reason or 'Proxy Error' in reason or \
-                            'Error reading from remote server' in reason:
-                pass
-            elif 'Connection refused' in str(ex) or 'timed out' in str(ex) or 'Could not resolve' in str(ex):
-                msg += 'Error: %s' % str(ex)
+            excReason = getattr(ex, 'reason', '')
+            errorMsg = 'Failed to fetch parentage map from WMStats, skipping this cycle. '
+            errorMsg += 'Exception: {}. Reason: {}. Error: {}. '.format(type(ex).__name__,
+                                                                        excReason, str(ex))
+            if isPassiveError(ex):
+                logging.warning(errorMsg)
             else:
-                msg = "Unknown failure while fetching parentage map from WMStats. Error: %s" % str(ex)
-                raise DBSUploadException(msg)
-            logging.warning(msg)
-            myThread.logdbClient.post("DBS3Upload_parentMap", msg, "warning")
+                errorMsg += 'Hit a terminal exception in DBSUploadPoller.'
+                raise DBSUploadException(errorMsg)
+            myThread.logdbClient.post("DBS3Upload_parentMap", errorMsg, "warning")
             success = False
         else:
             myThread.logdbClient.delete("DBS3Upload_parentMap", "warning", this_thread=True)
@@ -376,7 +365,7 @@ class DBSUploadPoller(BaseWorkerThread):
         # Load them if we don't have them
         blocksToLoad = []
         for block in openBlocks:
-            if block['blockname'] not in self.blockCache.keys():
+            if block['blockname'] not in self.blockCache:
                 blocksToLoad.append(block['blockname'])
 
         # Now load the blocks
--- a/test/python/WMComponent_t/DBS3Buffer_t/DBSUpload_t.py
+++ b/test/python/WMComponent_t/DBS3Buffer_t/DBSUpload_t.py
@@ -20,13 +20,21 @@ from WMComponent.DBS3Buffer.DBSBufferBlock import DBSBufferBlock
 from WMComponent.DBS3Buffer.DBSBufferDataset import DBSBufferDataset
 from WMComponent.DBS3Buffer.DBSBufferFile import DBSBufferFile
 from WMComponent.DBS3Buffer.DBSBufferUtil import DBSBufferUtil
-from WMComponent.DBS3Buffer.DBSUploadPoller import DBSUploadPoller
+from WMComponent.DBS3Buffer.DBSUploadPoller import DBSUploadPoller, isPassiveError
 from WMCore.DAOFactory import DAOFactory
 from WMCore.DataStructs.Run import Run
 from WMCore.Services.UUIDLib import makeUUID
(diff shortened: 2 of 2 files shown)


todor-ivanov:
Thanks @amaltaro! Is it still in `not-tested` state or you just forget to update the comment. Asking, because If the later, then I should proceed with the review.

amaltaro:
@todor-ivanov it's been tested via unit tests. I'm now running the final test before it gets added to the final new production release.
So, please feel free to review it.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -141,11 +109,40 @@ def uploadWorker(workInput, results, dbsUrl):
...
+                       'OpenSSL SSL_connect: SSL_ERROR_SYSCALL']
+
+    excReason = getattr(exceptionObj, 'reason', '')
+    errorMsg = 'Failed to fetch parentage map from WMStats, skipping this cycle. '
+    errorMsg += 'Exception: {}. Reason: {}. Error: {}. '.format(type(exceptionObj).__name__,
+                                                                excReason, str(exceptionObj))
+    for passiveMsg in passiveErrorMsg:
+        if passiveMsg in excReason:
+            break
+        elif passiveMsg in str(exceptionObj):
```
todor-ivanov, line 134: Matching a string inside the whole (string converted) exception object only to identify the type of Error, seems to me kind of error prone. I am not sure if this is a common practice, though. Mybe it is just me not being familiar with it.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -141,11 +109,40 @@ def uploadWorker(workInput, results, dbsUrl):
...
+    passException = True
+    passiveErrorMsg = ['Service Unavailable', 'Service Temporarily Unavailable',
+                       'Proxy Error', 'Error reading from remote server',
+                       'Connection refused', 'timed out', 'Could not resolve',
+                       'OpenSSL SSL_connect: SSL_ERROR_SYSCALL']
+
+    excReason = getattr(exceptionObj, 'reason', '')
+    errorMsg = 'Failed to fetch parentage map from WMStats, skipping this cycle. '
+    errorMsg += 'Exception: {}. Reason: {}. Error: {}. '.format(type(exceptionObj).__name__,
+                                                                excReason, str(exceptionObj))
```
todor-ivanov, line 130: I am not quite confident it is a good idea of having this message taken out of  `DBSUploadPoller` and putting it here. I would preserve the messages related to some kind of exceptions inside the object instance of the above class, while this function here sounds to me a decent one to be considered as a generic function for testing passive vs. hard errors coming from a module. I can already foresee a place or two I could benefit from it. Putting this message here constrains its usage to only one particular type of calls and at the same time cuts significant information in the class it was taken from.

todor-ivanov (changes requested):
Hi @amaltaro In general the code is having the right structure, but there are two things I am not comfortable with. Please take a look at the comments inline.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
amaltaro: I don't think it's a good practice, but given that we do not have different exception types (or return code) for these exceptions, the only way we can prevent the component from crashing from transient issues in the system, is by parsing the error message returned in the HTTP call.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -141,11 +109,40 @@ def uploadWorker(workInput, results, dbsUrl):
...
+    for passiveMsg in passiveErrorMsg:
+        if passiveMsg in excReason:
+            break
+        elif passiveMsg in str(exceptionObj):
+            break
+    else:
```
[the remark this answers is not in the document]
amaltaro: This is actually a for-else loop. It looks weird to my eyes, but it's a valid construction. If we complete the for loop without breaking out of it, the else block gets executed (which means, none of our error messages matched the actual http error).

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
amaltaro: To make it more generic, we would have to move it to a different place, and accept the list of errors as a parameter.
Moving the error message back to the class object implementation is feasible, but I'm trying to avoid code duplication; thus just profiting of some exception parsing that is already done in this function. Let me see if it makes sense to change it.

todor-ivanov: Reading the list of errors as a parameter is not a bad idea indeed.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
todor-ivanov: Aren't all those expected to be visible inside the Exception Reason. And am I correct in the observation that in the current list of errors you consider, there are not only WMCore exceptions but also exceptions from libraries with regular use all over the place like SSL etc. Meaning we are not in full control of where the Error string would appear.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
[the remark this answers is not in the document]
todor-ivanov: Reading the trend of `elif` statements from the previous lines, it is not that hard to fool oneself it is a conclusion statement in the above `for` control structure which is jut missing indentation. Me, obviously, easily fell into that trap :)

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
amaltaro: Exception reason contains the string with the error message. IF, our request response comes back with a header.reason attribute:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/pycurl_manager.py#L285

however, I think there could be other problems when executing `curl.perform()` which would actually raise an exception right there, thus potentially not having the `reason` attribute.
Reading the traceback I posted in the GH issue, this is exactly what happened two days ago.

amaltaro:
@todor-ivanov Todor, can you please have another look? I have pushed commits 3 and 4 with the error message change that you suggested. If you prefer this approach, I will then squash my commits. Thanks

todor-ivanov (approved):
Things look much better now, with the error message back to the place where it belongs.
Thanks @amaltaro, all the rest looks fine to me.

amaltaro:
Tests went fine, even though I don't think we hit any frontend issues and/or instabilities...

PR #10261, opened 2021-02-04
Make WMAgent job wrapper slightly more verbose; fix one error message

### LINKED ISSUE #10259: wrong errror message/exit code
**Impact of the bug**
WMA

**Describe the bug**
**RATS** I put wrong link to code in initial posting. EDITING IT NOW.
misleading/unclear error message from
https://github.com/dmwm/WMCore/blob/6362413ccbaeed76f9d09e33e11b5dab24a13eef/etc/submit.sh#L140-L142
this message is printed when the previous code could not decide if to use $VO_CMS_SW_DIR or
$OSG_APP or $CVMFS. Which to begin with should never happen, i.e. a failures would have been caught by previous code. And if it would happen "job environment does not contain init.sh" it likely useless and misleading. Bootstrap failures are only triggered when a job happens to run a on a broken machine (CVMFS mount failures e.g.) and unless they are very specific and we print attempted commands and full stdout/stderr, they are usually useless for finding/fixing the actual problem.

**How to reproduce it**
Need to run a job on a broken WN to trigger this. 

**Expected behavior**
IMHO the full `else` clause should be removed.
The code which actually intializes python looks good to me.
Maybe the part which source cmsset_default.sh, which now only prints
```
======== WMAgent CMS environment load starting at Thu Feb  4 00:06:37 GMT 2021 ========
WMAgent bootstrap: WMAgent thinks it found the correct CMSSW setup script
======== WMAgent CMS environment load finished at Thu Feb  4 00:06:37 GMT 2021 ========
```
could also say WHERE it found CMSSW ($VO_CMS_SW_DIR|$OSG_APP|$CVMFS)

amaltaro:
Fixes #10259 

#### Status
ready

#### Description
Be explicit on the worker node setup style that the job is running on (LCG, OSG, CVMFS) and create a record in the log for that.
In addition to that, fix the error message in case a python installation area cannot be located.

yes

#### Related PRs
none

none

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -96,19 +96,23 @@ echo -e "======== WMAgent validate arguments finished at $(TZ=GMT date) ========
 echo "======== WMAgent CMS environment load starting at $(TZ=GMT date) ========"
 if [ -f "$VO_CMS_SW_DIR"/cmsset_default.sh ]
 then  #   LCG style --
+    echo "WN with a LCG style environment, thus using VO_CMS_SW_DIR=$VO_CMS_SW_DIR"
     . $VO_CMS_SW_DIR/cmsset_default.sh
 elif [ -f "$OSG_APP"/cmssoft/cms/cmsset_default.sh ]
 then  #   OSG style --
+    echo "WN with an OSG style environment, thus using OSG_APP=$OSG_APP"
     . $OSG_APP/cmssoft/cms/cmsset_default.sh CMSSW_3_3_2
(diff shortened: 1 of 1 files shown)


amaltaro:
I haven't tested it yet. But could you please confirm whether it addresses all the points that you mentioned in your issue, Stefano. @belforte

belforte (approved):
Perfect, thanks !

amaltaro:
Tested on the agent as well. Merging

PR #10277, opened 2021-02-11
[py2py3] Migration at level scr/python/A/B/C - slice 6

mapellidario:
Fixes #10127 

#### Status
In development

#### Related PRs

* Previous migration PR: #10262 (slice5, issue #10116)
* Next migration PR: #10284  (slice7, issue #10131 )

#### Description

Run futurize and some manual changes on the first batch of src/python/A/B/C.

Use of "native-string" strategy:
- `src/python/WMCore/JobSplitting/JobFactory.py`: no `from bultins import str` here

division:
- `src/python/WMCore/JobSplitting/ParentlessMergeBySize.py`: assuming that 
    ```python
    self.currentJob.addResourceEstimates(jobTime = jobTime, disk = (jobSize+largestFile)/1024)
    ```
    can accept float arguments

#### Is it backward compatible (if not, which system it affects?)

It should be. (Any possible cause for errors will we reported here)

#### External dependencies / deployment changes

Requires python-future in both py2 and py3 environments.

Issue #10132, opened 2020-12-03
[py2py3] Migration at level scr/python/A/B/C - slice 8


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 8-th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 8-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
                                         name  nfiles  gradual  future_pm
               WMCore_ProcessPool_ProcessPool       1     True         11
                   WMQuality_TestInitCouchApp       1     True          3
 WMQuality_REST_RESTBaseUnitTestWithDBBackend       1     True          0
                  WMCore_MicroService_Service       2     True          4
                          WMCore_Services_DBS       5     True         31
```

#### Is your feature request related to a problem? Please describe.

The end goal is to have WMCore py2 and py3 compatible.

#### Describe the solution you'd like

Run futurize on the first bulk of units 
and then apply the necessary manual changes.

Issue #10127, opened 2020-12-02
[py2py3] Migration at level scr/python/A/B/C - slice 6


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 6-th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 6-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
                                       name  nfiles  gradual  future_pm
       WMComponent_DBS3Buffer_DBSBufferUtil       1     True         20
                   WMCore_ACDC_CouchFileset       1     True          3
                   WMCore_ACDC_CouchService       1     True          1
             WMCore_JobSplitting_JobFactory       1     True         14
  WMCore_JobSplitting_ParentlessMergeBySize       1     True         10
 WMCore_JobSplitting_SiblingProcessingBased       1     True          2
         WMCore_JobSplitting_SplitFileBased       1     True          4
           WMCore_JobSplitting_TwoFileBased       1     True          2
        WMCore_JobSplitting_WMBSMergeBySize       1     True          6
                            WMCore_WMBS_Job       1     True          8
                   WMCore_WMBS_Subscription       1     True          8
            WMCore_JobSplitting_MergeBySize       1     True          0
               WMCore_JobSplitting_RunBased       1     True          7
              WMCore_JobSplitting_SizeBased       1     True          2
                  WMCore_Database_MySQLCore       1     True          2
             WMCore_JobSplitting_FixedDelay       1     True          0
           WMCore_JobSplitting_MinFileBased       1     True          2
   WMCore_JobSplitting_TwoFileAndEventBased       1     True          0
                   WMCore_Storage_DeleteMgr       1     True          6
                WMCore_ACDC_CouchCollection       1     True          0
```

#### Is your feature request related to a problem? Please describe.

The end goal is to have WMCore py2 and py3 compatible.

#### Describe the solution you'd like

Run futurize on the first bulk of units 
and then apply the necessary manual changes.

PR #10284, opened 2021-02-15
[py2py3] Migration at level scr/python/A/B/C - slice 7

### LINKED ISSUE #10131: [py2py3] Migration at level scr/python/A/B/C - slice 7
#### Intro

This is the 7-th slice in which the migration has been divided into. 

We start migrating the 7-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
        WMCore_ACDC_DataCollectionService       1     True         12
 WMCore_JobSplitting_EventAwareLumiByWork       1     True          7
           WMCore_JobSplitting_EventBased       1     True          4
                    WMCore_BossAir_RunJob       1     True          6
                     WMCore_WMBS_JobGroup       1     True          1
                WMCore_Database_DBFactory       1     True         25
             WMCore_WebTools_DatabasePage       1     True          0
                   WMCore_WebTools_WebAPI       1     True          2
                  WMCore_WebTools_RESTApi       1     True          7
                            WMCore_WMInit       1     True          4
                WMCore_WebTools_RESTModel       1     True         16
              WMCore_WebTools_NestedModel       1     True          7
                       WMQuality_TestInit       1     True          2
```

mapellidario:
Fixes #10131 

#### Status
Ready

#### Related PRs

* Previous migration PR: #10277 (slice6, issue #10127 )
* Next migration PR: #10289  (slice8, issue #10132 )

#### Description

Strings, using "native string" approach:
* `src/python/WMCore/ACDC/DataCollectionService.py`: did not use `from builtins improt str`
  - Possible problem for future us: here `str()` is used both to append an exception message to a string (which is something that is easier to manage) and to format the key of a dictionary (which can cause problems if we search among the dictionary keys at a later time). If we will have a problem with strings in this file, we will likely have to change how they work in one or both these usages.
  - Solution: I decide to be conservative: I did not change the behavior, but I decoupled these two usages of `str()`. I decided to format the exception with `ex.__str__()`, so that we can tweak the dictionary keys as we want/need at a later time without worrying of breaking the exception.
* `src/python/WMCore/JobSplitting/EventAwareLumiByWork.py`: same as file above
* `src/python/WMCore/WebTools/RESTApi.py`: similar as above
* `src/python/WMCore/WMBS/JobGroup.py`: not modernized `__str__()`

Python3 compatibility report:
* we use `round` in `src/python/WMCore/JobSplitting/EventAwareLumiByWork.py`. The behavior of `round` changed from py2 to py3, but I do not think that this is going to cause problems for us and consider the python3 warning a "false positive" for us. [EDIT] proposed change to `float(round())` in #10440 

It should be.

However:
* `src/python/WMCore/WebTools/RESTModel.py`: contains some `str`/`bytes` massaging due to oracle/mariadb mismatching in how they deal with unicode. This may cause some troubles. Check this with py3. Tracked in #10399

### DIFF
--- a/src/python/WMCore/ACDC/DataCollectionService.py
+++ b/src/python/WMCore/ACDC/DataCollectionService.py
@@ -7,6 +7,9 @@ Created by Dave Evans on 2010-07-30.
 Copyright (c) 2010 Fermilab. All rights reserved.
 """
 
+from builtins import next
+from future.utils import viewitems, listvalues
+
 import logging
 import threading
 from operator import itemgetter
@@ -68,7 +71,7 @@ def mergeFilesInfo(chunkFiles):
         _mergeRealDataRunLumis(mergedFiles)
 
     logging.info(" ... resulted in %d unique files.", len(mergedFiles))
-    return mergedFiles.values()
+    return listvalues(mergedFiles)
 
 
 def _isRunMaskDuplicate(run, lumis, runLumis):
@@ -100,7 +103,7 @@ def _mergeRealDataRunLumis(mergedFiles):
             runLumis[item['run_number']].extend(item['lumis'])
         # now write those back to the original data structure
         mergedFiles[fname]['runs'] = []
-        for run, lumis in runLumis.items():
+        for run, lumis in viewitems(runLumis):
             mergedFiles[fname]['runs'].append({'run_number': run,
                                                'lumis': list(set(lumis))})
     return
@@ -120,7 +123,7 @@ def fixupMCFakeLumis(files, acdcVersion):
 
     For more info, see issue: https://github.com/dmwm/WMCore/issues/9126
     """
-    for fname, fvalues in files.items():
+    for fname, fvalues in viewitems(files):
         if fname.startswith('MCFakeFile') and acdcVersion < 2:
             for run in fvalues['runs']:
                 if len(run['lumis']) > 1:
@@ -225,7 +228,7 @@ class DataCollectionService(CouchService):
         for row in results["rows"]:
             files = row["doc"].get("files", [])
             fixupMCFakeLumis(files, row['doc'].get("acdc_version", 1))
-            filesInfo.extend(files.values())
+            filesInfo.extend(listvalues(files))
 
         # second lfn sort
         filesInfo.sort(key=lambda x: x["lfn"])
@@ -419,7 +422,7 @@ class DataCollectionService(CouchService):
                     allRuns[run["run_number"]] = []
                 allRuns[run["run_number"]].extend(run["lumis"])
 
-        for run in allRuns.keys():
+        for run in allRuns:
             lumis = []
             lumis.extend(set(allRuns[run]))
             lumis.sort()
--- a/src/python/WMCore/JobSplitting/EventBased.py
+++ b/src/python/WMCore/JobSplitting/EventBased.py
@@ -6,6 +6,8 @@ Event based splitting algorithm that will chop a fileset into
 a set of jobs based on event counts
 """
 
+from builtins import str
+from builtins import range
 import logging
 import traceback
 from math import ceil
@@ -89,7 +91,7 @@ class EventBased(JobFactory):
                                         self.subscription.workflowName(), self.subscription['id'])
                     for f in fileList:
                         lumiDict = fileLumis.get(f['id'], {})
-                        for run in lumiDict.keys():
+                        for run in lumiDict:
                             f.addRun(run=Run(run, *lumiDict[run]))
 
             for f in fileList:
--- a/src/python/WMCore/WebTools/RESTApi.py
+++ b/src/python/WMCore/WebTools/RESTApi.py
@@ -48,7 +48,7 @@ class RESTApi(WebAPI):
                                  'version': 1}})
 
         # TODO: implement HEAD & TRACE
-        self.supporttypes = self.formatter.supporttypes.keys()
+        self.supporttypes = list(self.formatter.supporttypes)
 
     def _set_model(self, config):
         """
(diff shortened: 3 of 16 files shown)


src/python/WMCore/ACDC/DataCollectionService.py
```diff
@@ -172,7 +175,9 @@ def failedJobs(self, failedJobs, useMask=True):
                 workflow = job['workflow']
                 taskName = job['task']
             except KeyError as ex:
-                msg = "Missing required, non-standard key %s in job in ACDC.DataCollectionService" % (str(ex))
+                # use ex.__str__() instead of str(ex) to decouple str() from
+                # what is used in getLumiWhitelist for whiteList[str(run)]
```
amaltaro, line 179: I'm not sure I understand this comment. What's the relationship of `getLumiWhitelist` with stringfying an exception?

src/python/WMCore/JobSplitting/EventBased.py
```diff
@@ -89,7 +91,7 @@ def algorithm(self, *args, **kwargs):
                                         self.subscription.workflowName(), self.subscription['id'])
                     for f in fileList:
                         lumiDict = fileLumis.get(f['id'], {})
-                        for run in lumiDict.keys():
+                        for run in list(lumiDict.keys()):
```
amaltaro, line 94:
```
for run in list(lumiDict):
```
?

src/python/WMCore/WebTools/RESTApi.py
```diff
@@ -107,25 +107,25 @@ def _buildResponse(self, args=[], kwargs={}):
         except HTTPError as h:
             # If something raises an HTTPError assume it's something that should
             # go to the client
-            response.status = h[0]
-            for kwarg in kwargs.keys():
+            response.status = h.args[0]
```
amaltaro, line 110: should we make sure `args[X]` is used **everywhere** we have an HTTPError exception in WMCore? Is it all included in this PR?

amaltaro (commented):
Dario, I think this code is ready to go. However, I made 3 comments along the code that you might consider required to update.

Can you please educate me on the usage of `str(exception_object)` vs `exception_object.__str__()`? Should we start adopting it everywhere in the code?

src/python/WMCore/ACDC/DataCollectionService.py
mapellidario: Ok, this is an nice topic for discussion and I have not made a decision yet.

The point is that here we use `str()` to do two different things:
1. getting a string from an exception object
2. converting `run` into something that can be hashed and used as a dictionary key

Point 2. can be problematic, since, as Stefano found out, having a mismatch between the type used to create the dictionary keys and what we compare them against can cause problems. This is true especially in python3 where bytes are not automatically encoded into unicode, for example

```plaintext
Python 3.9.0 (default, Oct 13 2020, 20:14:06) 
[GCC 8.3.0] on linux
>>> data = {"hello": 1}
>>> data["hello"]
1
>>> data[b"hello"]
KeyError: b'hello'
```

This is why `whiteList[str(run)]` is likely to cause some that will require some more fiddling with strings in this module. However, we also use `str()` to format exception, so if we were to start fiddling with `from builtins import str` this would get different types of strings from exception and may cause to append bytes to unicode. This would happen because vanilla py2 `str(ex)` calls `ex.__str__()`, which we mapped to `__as_bytes` in WMException,  while if we use `from builtins import str` then `str(ex)` calls `ex.__unicode__()` which we mapped to `__as_unicode` in WMException. This convoluted method allows us to use `"my string" + str(ex)` in both py2 and py3, at the expense of not allowing us to use `from builtins import str; "my string" + str(ex)` in py2.

To avoid this two to usages of `str()` to affect each other in the future if we find some problems in `whiteList[str(run)]`, we can use `ex.__str__()`. 

This is me being super-paranoic. But I agree that *right now* we do not need such change, that we can postpone this to a later time, that if i can spot such a issue right now I will spot it also if and when we hape problems with `whiteList[str(run)]`. 

So I am reverting this change, always a pleasure to have a chat! Thanks for making the right questions every time!

src/python/WMCore/JobSplitting/EventBased.py
mapellidario: Uh, you are right! also `for run in lumiDict` should be ok. Let me check if I left this on purpose (strange, i should have added a comment) or if I just missed this change.
Thanks again!

src/python/WMCore/WebTools/RESTApi.py
mapellidario: Yeah, we should make sure that this is the case or this is will fail in py3! 

This is part of the `pylint --py3k` report though and it is `W1624`, so I would continue to address this one slice at a time instead of having one PR just for this type of change. I can run pylint on all the codebase to find all of the occurences though if you prefer!

src/python/WMCore/JobSplitting/EventBased.py
mapellidario: Ok, no unit test broke, so I really missed this one too! thanks again! :)

mapellidario:
In the last report from jenkins [DMWM-WMCore-PR-test/11277](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/11277/), the section about `pylint --py3k` contains warnings about the following files, but some of those are not involved in this PR and I will ignore them

```plaintext
- [x] src/python/WMCore/JobSplitting/EventAwareLumiByWork.py : this slice (we are not likely to suffer from any problem about this, considering as false positive)
- [x] test/python/Utils_t/Utilities_t.py : past slice (false positive warning)
- [ ] test/python/WMCore_t/WorkQueue_t/Policy_t/Start_t/Block_t.py: slice 11
- [ ] src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py: slice 16
- [ ] src/python/WMCore/WorkQueue/WorkQueue.py: slice 17
- [ ] test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py: slice 17
- [ ] test/python/WMCore_t/WMSpec_t/StdSpecs_t/StepChain_t.py: slice 22
```

Why is this happening?

mapellidario:
Test this please

mapellidario:
Good, jenkins `pylint --py3k` looks like it should now!

src/python/WMCore/WebTools/RESTApi.py
amaltaro: No worries Dario, let's stick to the plan and we deal with those when we find them in the respective slices. Thanks

amaltaro:
Thanks for providing these updates, Dario. Can you please squash the 3rd commit into the first one?

PR #10289, opened 2021-02-16
[py2py3] Migration at level scr/python/A/B/C - slice 8

### LINKED ISSUE #10132: [py2py3] Migration at level scr/python/A/B/C - slice 8

mapellidario:
Fixes #10132 

#### Status
In development

#### Related PRs

* Previous migration PR: #10284  (slice7, issue #10131 )
* Next migration PR: #10290  (slice9, issue #10133 )

#### Description

Strings:
* `src/python/WMCore/Services/DBS/DBS3Reader.py`: required some re-organization of the strings. This should be tested thoroughly! 
  - tracking in #10293

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -7,6 +7,11 @@ Readonly DBS Interface
 """
 from __future__ import print_function, division
 
+from builtins import object, str, bytes
+from future.utils import viewitems
+
+from Utils.Utilities import encodeUnicodeToBytes, decodeBytesToUnicode
+
 import logging
 from collections import defaultdict
 
@@ -37,8 +42,8 @@ def remapDBS3Keys(data, stringify=False, **others):
                'block_name': 'BlockName', 'lumi_section_num': 'LumiSectionNumber'}
 
     mapping.update(others)
-    formatFunc = lambda x: str(x) if stringify and isinstance(x, unicode) else x
-    for name, newname in mapping.iteritems():
+    formatFunc = lambda x: encodeUnicodeToBytes(x) if stringify else x
+    for name, newname in viewitems(mapping):
         if name in data:
             data[newname] = formatFunc(data[name])
     return data
@@ -212,10 +217,8 @@ class DBS3Reader(object):
                 )
         """
         # Pointless code in python3
-        if isinstance(block, str):
-            block = unicode(block)
-        if isinstance(dataset, str):
-            dataset = unicode(dataset)
+        block = decodeBytesToUnicode(block)
+        dataset = decodeBytesToUnicode(dataset)
 
         try:
             if block:
@@ -516,7 +519,7 @@ class DBS3Reader(object):
             for fp in f['parent_logical_file_name']:
                 childByParents[fp].append(f['logical_file_name'])
 
-        parentsLFNs = childByParents.keys()
+        parentsLFNs = list(childByParents)
 
         if len(parentsLFNs) == 0:
             msg = "Error in "
@@ -578,7 +581,7 @@ class DBS3Reader(object):
         """
 
         singleBlockName = None
-        if isinstance(fileBlockNames, basestring):
+        if isinstance(fileBlockNames, (str, bytes)):
             singleBlockName = fileBlockNames
             fileBlockNames = [fileBlockNames]
 
@@ -637,8 +640,7 @@ class DBS3Reader(object):
              "Files" : { LFN : Events },
              "IsOpen" : True|False}
         """
-        if isinstance(fileBlockName, str):
-            fileBlockName = unicode(fileBlockName)
+        fileBlockName = decodeBytesToUnicode(fileBlockName)
 
         if not self.blockExists(fileBlockName):
             msg = "DBSReader.getFileBlockWithParents(%s): No matching data"
--- a/src/python/WMCore/Services/DBS/ProdException.py
+++ b/src/python/WMCore/Services/DBS/ProdException.py
@@ -46,29 +46,29 @@ class ProdException(exceptions.Exception):
         #  //
         # // Automatically determine the module name
         #//  if not set
-        if self['ModuleName'] == None:
+        if self.data['ModuleName'] == None:
             frame = inspect.currentframe()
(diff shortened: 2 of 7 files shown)


mapellidario:
Test this please

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -212,10 +217,8 @@ def listRunLumis(self, dataset=None, block=None):
                 )
         """
         # Pointless code in python3
-        if isinstance(block, str):
-            block = unicode(block)
-        if isinstance(dataset, str):
-            dataset = unicode(dataset)
+        block = decodeBytesToUnicode(block)
```
amaltaro, line 220: This code does not make any sense to me! It's also not consistent in this module, e.g.: `listRuns`.
Dario, I'd suggest to completely remove these lines

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -637,8 +640,7 @@ def getFileBlockWithParents(self, fileBlockName):
              "Files" : { LFN : Events },
              "IsOpen" : True|False}
         """
-        if isinstance(fileBlockName, str):
```
amaltaro, line 640: same here, I'd suggest to remove these lines.

src/python/WMCore/Services/DBS/ProdException.py
```diff
@@ -6,8 +6,7 @@
 
 """
 
-
-
+from future.utils import viewitems
```
amaltaro, line 9: Just an observation, I could be wrong, but I think this code isn't used at all (perhaps since the migration to DBS3!)

amaltaro (changes requested):
Dario, I made a couple of comments in the code which I believe it needs updating. However, please do let me know if I missed something.

This DBS3Reader module is sort of a DBS2 + DBS3 functionality, making things very confusing and possible outdated... Hopefully things will work out of the box here ;)

src/python/WMCore/Services/DBS/DBS3Reader.py
mapellidario: Alan, my job is to make sure that the code that we have right now running in py2 continue to run in the same way in py2 and will continue to be the same in py3.

You are the one that 5y ago added in [this commit](https://github.com/dmwm/WMCore/commit/b50c47968417f7b5321cd595be365a9040a3a6d2) the lines

```python
        # Pointless code in python3
        if isinstance(block, str):
            block = unicode(block)
        if isinstance(dataset, str):
            dataset = unicode(dataset)
```

to `listRunLumis` and not to `listRuns`. Do not complain with me about your inconsistencies.

Moreover, this code in py2 makes sure that if `block` and `dataset` are strings of type sequence of bytes (because `str` in vanilla py2, without any imports from future, means a string of type bytes), then they are converted to unicode strings. `block = decodeBytesToUnicode(block) ` does exactly the same and I had to use it not to rely on the `unicode` keyword, since it is not available in py3 anymore. If you want this code to behave the same in py3 as it is behaving now in py2, this is the only way. 

If you consider that these lines that you added 5 years ago are no longer necessary, then you remove them. It is not my job to refactor your code and decide if it is necessary to explicitely convert to unicode the input of `listRunLumis`, while it is not necessary in the case of `listRuns`. I have to trust that past-you added them for a reason. It is impossible for me to check every time which type a function argument is. I trust that if you decided to convert it from bytes to unicode in py2 explicitely, then
1. you need to continue doing that while we have py2 running
2. you are likely to need it in py3 also

src/python/WMCore/Services/DBS/ProdException.py
mapellidario: `Services/DBS/DBS3Reader.py` imports `from WMCore.Services.DBS.DBSErrors import DBSReaderError`. `DBSErrors.DBSReaderError` inherits from `DBSErrors.DataMgmtError`, which then inherits from `ProdException`

mapellidario:
In any case, I am super-worried of how DBS3Reader manages strings, this is why I created #10293. This will need follow-up, especially when we will switch to py3

src/python/WMCore/Services/DBS/DBS3Reader.py
amaltaro: You're completely right, Dario! It was an observation (not a complaint by the way) that goes beyond the scope of this code modernization. I need to policy myself and keep focus, sorry about that.
Having a quick look at that commit, it was simply a shift of lines. So the code was already there and now there is no way to know the reason.
Anyways, if we find problems in this module, perhaps removing unnecessary string handling will ease our life. We'll see.

src/python/WMCore/Services/DBS/DBS3Reader.py
amaltaro: marking as resolved (not relevant to be frank)

src/python/WMCore/Services/DBS/ProdException.py
amaltaro: Thanks for looking this up, Dario. I will stop making such unrelated observations ;)

src/python/WMCore/Services/DBS/DBS3Reader.py
mapellidario: Perfect! So, should we keep these changes in this PR, come back to this module with py3 unit tests and eventually moving this discussion to #10293, or did I misunderstood you?

amaltaro: Yes, I think we keep the compatibility changes you provided. If we have issues with this module, then I'd say the initial point to look at things might be the previous (and now modernized) unicode casting. I'll make a note to the GH issue you have opened though just to make sure this information doesn't get lost. Thanks again Dario!

amaltaro (approved):

PR #10292, opened 2021-02-16
Properly set rucio.cfg in our docker unittest setup

### LINKED ISSUE #10291: Default rucio.cfg file needs to be updated in our wmagent(-dev) docker image
**Impact of the bug**
WMAgent (docker image)

**Describe the bug**
With the removal of PhEDEx code (or switch flag between PhEDEx and Rucio), lots of problems got exposed with our unit tests. One of them is that, in a few places in WMCore, we create an instance of the `Rucio` wrapper object with default values, where the default values are taken from the rucio.cfg file. That means, we need to make sure the rucio.cfg file (linked through RUCIO_HOME environment variable) contains reasonable default urls.

**How to reproduce it**
Run a wmagent docker container and:
```
source env_unittest.sh
$manage start-services
cd wmcore_unittest/WMCore/test/python/WMCore_t/WorkQueue_t/
nosetests WMBSHelper_t.py:WMBSHelperTest.testDuplicateFileInsert
```

which should fail with [1]

**Expected behavior**
Automatically take the Rucio variables defined in the unittests configuration file:
https://github.com/dmwm/WMCore/blob/master/test/deploy/WMAgent_unittest.secrets

and update the rucio.cfg file accordingly, such that when we hit a line like
```
rucioObj = Rucio("my account name")
```
it uses that auth_url and host_url as defined inside the rucio.cfg.

Problem exposed in: https://github.com/dmwm/WMCore/pull/10283

[1]
```
ClientProtocolNotSupported: Client protocol not supported.
Details: '' not supported
```

amaltaro:
Fixes #10291 

#### Status
ready

#### Description
This PR fixes a problem with our docker unittest setup, when a Rucio connection is not parameterize, but instead it relies on the rucio.cfg file.
There are two changes here:
* change default Rucio urls to the production instance. Once the integration service becomes accessible off-site, we can roll it back;
* update the rucio.cfg file - used by WMAgent unittests - whenever we source the env_unittest.sh script. It reads the rucio urls from the sec rets configuration file.
 
yes

#### Related PRs
none

none


amaltaro:
These changes have been tested with my container. Will merge it, but it will only take effect once we build a new WMAgent image with this change in (likely in 1.4.6.pre6).

PR #10296, opened 2021-02-17
fix bug in selecting stagein for ul logcollect, fixes #10294

### LINKED ISSUE #10294: LogCollect uses stageIn when it shouldn't
**Impact of the bug**
LogCollect uses stagein to retrieve logArchives from storage, possibly overloading storage

**Describe the bug**
LogCollect uses stageIn when it should use edmCopyUtil. The problem was that 2 years ago the cutoff release in this check

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py#L103

was updated to a newer release and now the SL6 only check doesn't work anymore because we have releases older that are available as SL7. This includes some releases used in the UL Campaigns Gen steps.

Fix is easy, just add another conditional check for SL7 scram archs with the appropriate overrides.

hufnagel:
The detection logic when to use edmCopyUtil and when to use stageIn for LogCollect jobs has a hole for UL CMSSW releases, effect is that it uses stageIn when it really should be using edmCopyUtil.

Fixes #10294

#### Status
not-tested

#### Description
Adds check for older SL7 releases, fixes a bug introduced with the last change 2 years ago

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -103,6 +103,11 @@ class LogCollect(Executor):
         useEdmCopyUtil = True
         if isCMSSWSupported(cmsswVersion, "CMSSW_10_4_0"):
             pass
+        elif scramArch.startswith('slc7_amd64_'):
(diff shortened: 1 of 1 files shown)


amaltaro:
Thanks for providing this fix, Dirk. I will backport it to the agent branch and make sure it gets deployed in production soon.

PR #10295, opened 2021-02-17
[py2py3] Migration at level scr/python/A/B/C - slice 10

mapellidario:
Fixes #10134 

#### Status

Ready

#### Related PRs

* Previous migration PR: #10290 (slice9, issue #10133)
* Next migration PR: #10299  (slice11, issue #10135  )

#### Description

Run futurize and some manual changes on the first batch of src/python/A/B/C.

`round`:
- `src.python.WMCore.Services.LogDB.LogDBBackend`
  - proposed change from `round()` to `float(round())` in #10440 

#### Is it backward compatible (if not, which system it affects?)

It should be. 

No major potential problems was found. Let's hope for the best.

#### External dependencies / deployment changes

Requires python-future in both py2 and py3 environments.

PR #10299, opened 2021-02-18
[py2py3] Migration at level scr/python/A/B/C - slice 11 

### LINKED ISSUE #10135: [py2py3] Migration at level scr/python/A/B/C - slice 11
#### Intro

This is the 11-th slice in which the migration has been divided into. 

We start migrating the 11-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
src/python/WMQuality/Emulators/DataBlockGenerator/
src/python/WMQuality/Emulators/DBSClient/
src/python/WMCore/Services/TagCollector/
src/python/WMCore/Services/WMStats/
src/python/WMCore/Storage/Backends/
src/python/WMCore/Storage/Plugins/
src/python/WMCore/WorkQueue/Policy/
src/python/WMCore/WMSpec/WMStep.py

in which
 WMQuality_Emulators_DataBlockGenerator       2     True          6
          WMQuality_Emulators_DBSClient       5     True         16
              WMCore_Services_TagCollector       2     True          1
                WMCore_Storage_Backends      10     True          9
                 WMCore_Storage_Plugins      10     True          7

removed from this slice after plan changed:
                 WMCore_Services_SiteDB       2     True          2
            WMCore_ThreadPool_Oracle       3     True          0
              WMCore_Services_RequestDB       2     True          4
                  WMCore_Services_LogDB       4     True         12
                 WMCore_Services_PhEDEx       4     True         21
```

mapellidario:
Fixes #10135 

#### Status
ready

#### Related PRs

* Previous migration PR: #10295  (slice10, issue #10134 )
* Next migration PR: #10307  (slice12, issue #10136 )

#### Description

"native string" approach used in:
- `src/python/WMCore/WorkQueue/Policy/Start/Block.py`: did not import `str` from future

Other changes that require some comments:
-`src/python/WMCore/WorkQueue/Policy/Start/StartPolicyInterface.py`: here `str()` is used extensively in py2, so I would like to change it as little as possible. This is why I decided to import `from builtins import str as newstr, bytes` to use `(newstr, bytes)` instead of `basestring`, but without affecting all the extensive use of `str()` that likely (and hopefully!) will not require any change when switching to py3


mapellidario:
Damn! The unit test that is failing is a weird error that unsurprisingly emerges from string handling.

The unit tests from `test/python/WMCore_t/Services_t/DBS_t/DBSReader_t.py` make use of `src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py`. In both `MockDbsApi.listFileArray` and `MockDbsApi.listFileLumiArray` we are currently using (in py2):

```python
def listFileLumiArray/listFileArray(...)
        if 'logical_file_name' in kwargs and len(kwargs['logical_file_name']) > 1:
            origArgs = copy.deepcopy(kwargs)
            returnDicts = []
            for lfn in kwargs['logical_file_name']:
                origArgs.update({'logical_file_name': [unicode(lfn)]})
                returnDicts.extend(self.genericLookup(**origArgs))
            return returnDicts
...
    def genericLookup(self, *args, **kwargs):
        ...
        if kwargs:
            signature = '%s:%s' % (self.item, sorted(viewitems(kwargs)))
        ...
        return mockData[self.url][signature]
```

In py3 we can not rely on `unicode`, so we need to ged rid of it. We normally would simply use `from builtins import str` and then replace `unicode(lfn)` with `str(lfn)`. 
Unfortunately, in `genericLookup` we render this to `signature`, and `unicode(lfn)` becomes `u"something"`, while `str(lfn)` becomes simply `"something"`. So, if we use `str(lfn)` we get a slightly different `signature`, which causes `mockData[self.url][signature]` to fail with a `KeyError`

#### logs

code

```python
        except KeyError as e:
            print(signature in mockData[self.url])
            print(signature.encode("utf8") in mockData[self.url])
            print(mockData[self.url])
            print(type(signature), signature)
            raise KeyError("DBS mock API could not return data for method %s, args=%s, and kwargs=%s (URL %s)." %
                           (self.item, args, kwargs, self.url))
```

logs

```plaintext
False
False
...
u"listFileLumiArray:[('logical_file_name', [u'/store/relval/CMSSW_6_2_0_pre4/SingleMu/RECO/PRE_61_V1_RelVal_mu2012A-v1/00000/7C2C57AD-F591-E211-85C4-0026189438F4.root'])]": [{u'lumi_section_num': [690, 710, 508, 507], u'logical_file_name': u'/store/relval/CMSSW_6_2_0_pre4/SingleMu/RECO/PRE_61_V1_RelVal_mu2012A-v1/00000/7C2C57AD-F591-E211-85C4-0026189438F4.root', u'run_num': 191226}],
...
<type 'str'> listFileLumiArray:[('logical_file_name', ['/store/data/Commissioning2015/Cosmics/RAW/v1/000/238/545/00000/C47FDF25-2ECF-E411-A8E2-02163E011839.root'])]
```

Problem: `mockData[self.url]` contain keys of the form `u"listFileLumiArray:[('logical_file_name', [u'something '])]"`, but with `str(lfn)` we create keys of the form `u"listFileLumiArray:[('logical_file_name', ['something'])]"`

I need to think about this a bit, and then ask some help to @amaltaro :sob:

mapellidario:
The problem here arises because we mock DBS data by using `json.load` to de-serialize the file `./test/data/Mock/DBSMockData.json` into a python dictionary. In python 2, `json.JSONEncoder` imports strings in the json file as unicode strings in python dictionary, see [official documentation](./test/data/Mock/DBSMockData.json). This is where unicode strings come from in our py2 unit tests.

No unicode strings are used by DBS, though, so this seems to involve unit tests only.

We have multiple possible solutions:
- continue using `unicode()` without modernization in py2, switching to py3 `str()` only when running in py3 -> no modernization in py2!
- change how json import strings with something like [this suggestion from stack overflow](https://stackoverflow.com/a/33571117)
- convert all the `u"something"` to `"something"` manually _after_ creating the mock data dictionary.

The first option is the easiest, requiring only minimal changes. After all this is a modernization problem with unit testing, which means that can be postponed when we have a py3 environment without causing any breaking changes in our code

amaltaro:
>The first option is the easiest, requiring only minimal changes. After all this is a modernization problem with unit testing, which means that can be postponed when we have a py3 environment without causing any breaking changes in our code

Go for it Dario!

mapellidario:
> Go for it Dario! 

Already in :smirk: 
I am still applying the final touches, but this PR should be ready for reviewing soon

mapellidario:
@amaltaro I am not sure if the failing unit test is my fault and if I try to run it on my laptop it hangs, should we ask jenkins to run the tests again?

src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py
```diff
@@ -35,6 +38,16 @@
 mockData['https://cmsweb-prod.cern.ch/dbs/prod/global/DBSReader'] = mockDataGlobal
 mockData['https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSReader'] = mockData03
 
+def _unicode(data):
```
amaltaro, line 41: Dario, how about we create a GH issue asking for this function (and its references) to be removed once we have services running in python3 only?

amaltaro (approved):
It looks good to me as well, thanks Dario.

I made one comment in the code though, for something that we can definitely clean up once in python3 land.

And another comment, that goes beyond the scope of these changes, is that I've seen one or two places in this PR that we could use something like:
```
iterkeys(myDict).next()
```
to avoid large memory footprint when expanding dictionary keys. Comment that might be useful in the future.

src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py
mapellidario: Yeah, this is a nice suggestion! I can also create a new tag for such issues, like `clean py3`, `followup py3` or whatever else you like

mapellidario: created #10322 for addressing this

mapellidario:
Are you suggesting 

```python
- list(myDict)[0]
+ iterkeys(myDict).next()
```

?

I totally agree! This is a very nice solution! I will keep it in mind, thanks!

mapellidario:
I could not resist and included your suggestion into this PR as well, was too good to ignore it, sorry!
I added two commits with just these changes. if and when you approve them, I squash them!

amaltaro:
> Are you suggesting
> - list(myDict)[0]
> + iterkeys(myDict).next()

exactly! And I see you have even found a better variation of it in your 3rd commit.

Unfortunately, python3 (3.8) does not seem to like neither of the following two constructions:
```
iterkeys(myDict).next()
or
iter(myDict).next()
```

The py2/py3 compatible way to make this change seems to be:
```
next(iter(myDict))
```
if you do not see any problem with this approach, can you please update this PR once again?

mapellidario:
I noticed the same thing and fixed the last two commits

amaltaro (approved):
Everything looks good to me, Dario. Feel free to squash the extra commits accordingly, and we can merge it. Thanks again!

mapellidario:
For future-me: python-future provides a wrapper around `next` that tries to use `__next__` of the object it is called onto. It's likely that we will not need this `future's `next`, but it could be a valid option in some cases. [code docstring](https://github.com/PythonCharmers/python-future/blob/master/src/future/builtins/newnext.py)

PR #10307, opened 2021-02-24
[py2py3] Migration at level scr/python/A/B/C - slice 12

### LINKED ISSUE #10136: [py2py3] Migration at level scr/python/A/B/C - slice 12
#### Intro

This is the 12-th slice in which the migration has been divided into. 

We start migrating the 12-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
     WMCore_WorkQueue_WorkQueue       1     True         20
WMCore_MicroService_DataStructs       4    False         76
```

mapellidario:
Fixes #10136 

#### Status

ready

#### Related PRs

* Previous migration PR: #10299  (slice11, issue #10135 )
* Next migration PR: #10310  (slice13, issue #10137  )

#### Description

Other changes that require some comments:

- `src/python/WMCore/WorkQueue/WorkQueue.py`: here `str()` is used extensively in py2 to simply format log messages, so I would like to use the "native string" approach and not touch it. However, we also need to compare a variable with both unicode and bytes strings types. This is why I decided to decouple the import `from builtins import str as newstr, bytes` to use `(newstr, bytes)` instead of `basestring`, but without affecting all the extensive use of `str()` that likely will not require any change when switching to py3

It should be. However,
- `src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py` checks if the python2 types `str` or `unicode` are in a list. If the list is not built properly, this is likely to fail. There is no workaround, we can only modernize this code and fix accordingly where the list `data['type']` is built from. This may require some testing that goes beyond unit tests.
  - tracked in #10323 

### DIFF
--- a/src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
+++ b/src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
@@ -5,6 +5,8 @@ Description: Provides a document Template for the MSRuleCleaner MicroServices
 
 # futures
 from __future__ import division, print_function
+from builtins import object, str, bytes
+from future.utils import viewitems
 
 from copy import deepcopy
 from Utils.IteratorTools import flattenList
@@ -57,10 +59,10 @@ class WfParser(object):
             for value in wfObj:
                 self._paramFinder(value)
         if isinstance(wfObj, dict):
-            for key, value in wfObj.items():
+            for key, value in viewitems(wfObj):
                 self._paramFinder(value)
-            for key in self.extDoc.keys():
-                if key in wfObj.keys():
+            for key in self.extDoc:
+                if key in wfObj:
                     self.extDoc[key]['values'].append(deepcopy(wfObj[key]))
 
     def _wfParse(self):
@@ -99,7 +101,7 @@ class WfParser(object):
         """
 
         # Convert back the so aggregated extDoc to the original structure:
-        for keyName, data in self.extDoc.items():
+        for keyName, data in viewitems(self.extDoc):
             if len(data['values']) == 0:
                 self.extDoc[keyName] = deepcopy(data['default'])
             elif len(data['values']) == 1:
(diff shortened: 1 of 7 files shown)


src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
```diff
@@ -120,8 +122,8 @@ def _wfParse(self):
                     self.extDoc[keyName] = {}
                     for item in data['values']:
                         self.extDoc[keyName].update(item)
-                elif (isinstance(data['type'], tuple) and (str in data['type'] or unicode in data['type'])) or \
-                     (data['type'] is str or data['type'] is unicode):
+                elif (isinstance(data['type'], tuple) and (bytes in data['type'] or str in data['type'])) or \
```
amaltaro, line 125: rhetorical comment, what a horrible line!

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -12,6 +12,9 @@
 
 from __future__ import division, print_function
 
+from builtins import str as str_, bytes
```
amaltaro, line 15: I've seen you using `from builtins import str as newstr` in slice 11. Should we adopt the same pattern in here as well?

amaltaro (changes requested):
Dario, I think there is one minor change that we can make to keep things more consistent. If you make that change, feel free to squash it with the first commit.

src/python/WMCore/WorkQueue/WorkQueue.py
mapellidario: Uh, you are correct, I definitely lacked consistency here! fixed!

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
mapellidario: I also lit up a candle when I pushed this, praying it does not break. Oh, let me create an issue to track this, we should definitely explicitely check that this work in py3.

mapellidario: Created #10323 to track this

src/python/WMCore/WorkQueue/WorkQueue.py
mapellidario: Pushed new commit with this change

amaltaro (approved):
Thanks for the prompt fix, Dario. Please squash those commits and it's good to go.

src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py
todor-ivanov: A rhetorical reply. Thank you @amaltaro

PR #10308, opened 2021-02-24
Add AlertManager API

### LINKED ISSUE #10234: MicroServices throw UnboundLocalError when sending email alerts from kubernetes/docker containers
**Impact of the bug**
MSTransferor (but it will apply to any system trying to send alerts via SMTP)

**Describe the bug**
This issue is meant to report/address two issues:
1) some missing configuration on our kubernetes/docker setup which does not allow to create a connection on the host:port
2) and likely deal with this alert failure as a soft failure, still allowing the workflow to be moved to staging.

The second issue isn't really a problem at the moment, but it makes MSTransferor to have a "silent" behaviour, in the sense that: when the service fails to send that alert notification, the workflow in question is skipped from that cycle - even though the Rucio rule creation has already happened; then in the next cycle, MSTransferor finds an existent rule (likely in INJECTED/REPLICATING status), so the service assumes data is already available and simply moves that workflow to `staging` without persisting any rule id to be monitored by MSMonitor. Thus, MSMonitor will bypass this workflow right away because there are no rules to be monitored.

**How to reproduce it**
trigger email notification in a pod

**Expected behavior**
Microservices - or any other WMCore service - should be able to send email notifications.

Regarding the MSTransferor behaviour, I think we can log the exception, make a record with the content that was supposed to be sent via email (I think MSTransferor already does that), and move on with the workflow processing as if there were no problems sending the email.

Some sections of the MSTransferor log under
amaltaro@vocms0750:/cephfs/product/dmwm-logs $ less ms-transferor-20210121-ms-transferor-7744c99cd8-5p6m8.log 
```
2021-01-21 09:41:53,542:INFO:MSTransferor: Checking secondary data location for request: haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987, TrustPUSitelists: False, request white/black list PNNs: set([u'T2_CH_CE
RN'])
2021-01-21 09:41:53,543:INFO:MSTransferor: it has secondary: /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_realistic_v11_L1v1-v2/PREMIX, total size: 697625.14 GB, current disk locations: set([u'T1_FR_CCIN2P3_Disk', 
u'T1_US_FNAL_Disk'])
2021-01-21 09:41:53,543:INFO:MSTransferor: secondary: /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_realistic_v11_L1v1-v2/PREMIX will need data placement!!!
2021-01-21 09:41:53,543:INFO:MSTransferor: Finding final pileup destination for request: haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987
2021-01-21 09:41:53,543:INFO:MSTransferor:   found a PSN list: set([u'T2_CH_CERN']), which maps to a list of PNNs: set([u'T2_CH_CERN'])
2021-01-21 09:41:53,543:INFO:MSTransferor: Handling data subscriptions for request: haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987
2021-01-21 09:41:53,543:INFO:MSTransferor: Have whole PU dataset: /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_realistic_v11_L1v1-v2/PREMIX (697625.14 GB)
2021-01-21 09:41:53,558:DEBUG:connectionpool: http://cms-rucio.cern.ch:80 "GET /rses/?expression=T2_CH_CERN HTTP/1.1" 200 None
2021-01-21 09:41:53,612:DEBUG:connectionpool: http://cms-rucio.cern.ch:80 "GET /rses/T2_CH_CERN/attr/ HTTP/1.1" 200 None
2021-01-21 09:41:53,613:INFO:MSTransferor: Placing whole container, picked RSE: T2_CH_CERN out of an RSE list: T2_CH_CERN
2021-01-21 09:41:53,613:INFO:MSTransferor: Creating rule for workflow haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987 with 1 DIDs in container /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_
realistic_v11_L1v1-v2/PREMIX, RSEs: T2_CH_CERN, grouping: ALL
2021-01-21 09:41:53,664:DEBUG:connectionpool: http://cms-rucio.cern.ch:80 "POST /rules/ HTTP/1.1" 201 None
2021-01-21 09:41:53,664:INFO:MSTransferor: Rules successful created for /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_realistic_v11_L1v1-v2/PREMIX : [u'a383a3c0a23348b8a489b9296a09e454']
2021-01-21 09:41:53,665:ERROR:EmailAlert: Error sending alert email.
Details: [Errno 99] Cannot assign requested address
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.4.pre5/lib/python2.7/site-packages/Utils/EmailAlert.py", line 37, in send
    smtp = smtplib.SMTP(self.serverName)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/smtplib.py", line 256, in __init__
    (code, msg) = self.connect(host, port)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/smtplib.py", line 316, in connect
    self.sock = self._get_socket(host, port, self.timeout)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/smtplib.py", line 291, in _get_socket
    return socket.create_connection((host, port), timeout)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/external/python/2.7.13-comp/lib/python2.7/socket.py", line 575, in create_connection
    raise err
error: [Errno 99] Cannot assign requested address
2021-01-21 09:41:53,667:ERROR:MSTransferor: Unknown exception while making Transfer Request for haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987       Error: local variable 'smtp' referenced before assignment
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.4.pre5/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 207, in execute
    success, transfers = self.makeTransferRequest(wflow)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.4.pre5/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 625, in makeTransferRequest
    blocks, dataSize, nodes, idx)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.4.pre5/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 723, in makeTransferRucio
    self.notifyLargeData(aboveWarningThreshold, transferId, wflow.getName(), dataSize, dataIn)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.4.pre5/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSTransferor.py", line 748, in notifyLargeData
    self.emailAlert.send(emailSubject, emailMsg)
  File "/data/srv/HG2101d/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.4.pre5/lib/python2.7/site-packages/Utils/EmailAlert.py", line 43, in send
    smtp.quit()
UnboundLocalError: local variable 'smtp' referenced before assignment
```

and in the subsequent MSTransferor cycle, this is what happened to that workflow:
```
2021-01-21 09:53:39,553:INFO:MSTransferor: Checking secondary data location for request: haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987, TrustPUSitelists: False, request white/black list PNNs: set([u'T2_CH_CERN'])
2021-01-21 09:53:39,553:INFO:MSTransferor: it has secondary: /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_realistic_v11_L1v1-v2/PREMIX, total size: 697625.14 GB, current disk locations: set([u'T1_FR_CCIN2P3_Disk', u'T2_CH_CERN', u'T1_US_FNAL_Disk'])
2021-01-21 09:53:39,553:INFO:MSTransferor: secondary dataset: /Neutrino_E-10_gun/RunIISummer19ULPrePremix-UL18_106X_upgrade2018_realistic_v11_L1v1-v2/PREMIX already in place. Common locations with site white/black list is: set([u'T2_CH_CERN'])
2021-01-21 09:53:39,553:INFO:MSTransferor: Request haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987 does not have any further data to transfer
2021-01-21 09:53:39,553:INFO:MSTransferor: Transfer requests successful for haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987. Summary: []
2021-01-21 09:53:39,656:INFO:ReqMgrAux: Update in-place: False for transfer doc: haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987 was successful.
2021-01-21 09:53:39,656:INFO:MSTransferor: Transfer document successfully created in CouchDB for: haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987
2021-01-21 09:53:39,656:INFO:MSCore: MSTransferor updating haozturk_task_PPS-RunIISummer19UL18pLHEGEN-00007__v1_T_210121_092904_8987 status to: staging
```

### LINKED ISSUE #10340: Email alerts do not work from k8s pods
**Impact of the bug**
Any system trying to send alerts via SMTP is impacted.

**Describe the bug**
Services running in k8s pods cannot send email alerts

Additional detailed information can be found here: https://github.com/dmwm/WMCore/issues/10234

**How to reproduce it**
trigger email notification in a pod

**Expected behavior***
Any WMCore service should be able to send an alert when necessary. Create a replacement for alerting in k8s.

goughes:
Fixes https://github.com/dmwm/WMCore/issues/10340

#### Status
API has been tested manually, need another commit (or PR) to replace email alerting into the microservices with AlertManager API calls

#### Description
This adds a module to provide alerting capabilities via the MONIT AlertManager API. 

YES

#### Related PRs
NA

Specific documentation on this: https://github.com/dmwm/WMCore/wiki/Prometheus-AlertManager-wrapper-APIs
JIRA ticket for the CMS Monitoring team: https://its.cern.ch/jira/browse/CMSMONIT-363

### DIFF
--- a/src/python/Utils/Timers.py
+++ b/src/python/Utils/Timers.py
@@ -7,6 +7,7 @@ from __future__ import print_function, division, absolute_import
 
 from builtins import object
 import time
+from datetime import tzinfo, timedelta
 
 
 def timeFunction(func):
@@ -52,3 +53,47 @@ class CodeTimer(object):
         runtime = end - self.start
         msg = '{label} took {time} seconds to complete'
         print(msg.format(label=self.label, time=runtime))
+
+
+class LocalTimezone(tzinfo):
+    """
+    A required python 2 class to determine current timezone for formatting rfc3339 timestamps
+    Required for sending alerts to the MONIT AlertManager
+    Can be removed once WMCore starts using python3
+
+    Details of class can be found at: https://docs.python.org/2/library/datetime.html#tzinfo-objects
+    """
+
+    def __init__(self):
+        super(LocalTimezone, self).__init__()
+        self.ZERO = timedelta(0)
+        self.STDOFFSET = timedelta(seconds=-time.timezone)
+        if time.daylight:
+            self.DSTOFFSET = timedelta(seconds=-time.altzone)
+        else:
+            self.DSTOFFSET = self.STDOFFSET
+
+        self.DSTDIFF = self.DSTOFFSET - self.STDOFFSET
+
+    def utcoffset(self, dt):
+        if self._isdst(dt):
+            return self.DSTOFFSET
+        else:
+            return self.STDOFFSET
+
+    def dst(self, dt):
+        if self._isdst(dt):
+            return self.DSTDIFF
+        else:
+            return self.ZERO
+
+    def tzname(self, dt):
+        return time.tzname[self._isdst(dt)]
+
+    def _isdst(self, dt):
+        tt = (dt.year, dt.month, dt.day,
+              dt.hour, dt.minute, dt.second,
+              dt.weekday(), 0, 0)
+        stamp = time.mktime(tt)
+        tt = time.localtime(stamp)
+        return tt.tm_isdst > 0
--- a/src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
+++ b/src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
@@ -1,0 +1,105 @@
+"""
+
+AlertManagerAPI - send alerts to MONIT AlertManager via API calls
+"""
+
+from __future__ import division
+from datetime import timedelta, datetime
+import socket
+import json
+import logging
+
+from WMCore.Services.pycurl_manager import RequestHandler
+from Utils.Timers import LocalTimezone
+
+
+class AlertManagerAPI(object):
+    """
+    A class used to send alerts via the MONIT AlertManager API
+    """
+
+    def __init__(self, alertManagerUrl, logger=None):
+        self.alertManagerUrl = alertManagerUrl
+        # sender's hostname is added as an annotation
+        self.hostname = socket.gethostname()
+        self.mgr = RequestHandler()
+        self.ltz = LocalTimezone()
+        self.headers = {"Content-Type": "application/json"}
+        self.validSeverity = ["high", "medium", "low"]
+        self.logger = logger if logger else logging.getLogger()
+
+    def sendAlert(self, alertName, severity, summary, description, service, tag="wmcore", endSecs=600, generatorURL=""):
+        """
+        :param alertName: a unique name for the alert
+        :param severity: low, medium, high
+        :param summary: a short description of the alert
+        :param description: a longer informational message with details about the alert
+        :param service: the name of the service firing an alert
+        :param tag: a unique tag used to help route the alert
+        :param endSecs: how many minutes until the alarm is silenced
+        :param generatorURL: this URL will be sent to AlertManager and configured as a clickable "Source" link in the web interface
+
+        AlertManager JSON format reference: https://www.prometheus.io/docs/alerting/latest/clients/
+        [
+          {
+            "labels": {
+            "alertname": "<requiredAlertName>",
+            "<labelname>": "<labelvalue>",
+            ...
+          },
+            "annotations": {
+              "<labelname>": "<labelvalue>",
+            ...
+          },
+            "startsAt": "<rfc3339>", # optional, will be current time if not present
+            "endsAt": "<rfc3339>",
+            "generatorURL": "<generator_url>" # optional
+          },
+        ]
+        """
+
+        if not self._isValidSeverity(severity):
+            return False
+
+        request = []
+        alert = {}
+        labels = {}
+        annotations = {}
+
+        # add labels
+        labels["alertname"] = alertName
+        labels["severity"] = severity
+        labels["tag"] = tag
+        labels["service"] = service
+        alert["labels"] = labels
+
+        # add annotations
+        annotations["hostname"] = self.hostname
+        annotations["summary"] = summary
+        annotations["description"] = description
+        alert["annotations"] = annotations
+
+        # In python3 we won't need the LocalTimezone class
+        # Will change to d = datetime.now().astimezone() + timedelta(seconds=endSecs)
+        d = datetime.now(self.ltz) + timedelta(seconds=endSecs)
+        alert["endsAt"] = d.isoformat("T")
+        alert["generatorURL"] = generatorURL
+
+        request.append(alert)
+        # need to do this because pycurl_manager only accepts dict and encoded strings type
+        params = json.dumps(request)
+
+        res = self.mgr.getdata(self.alertManagerUrl, params=params, headers=self.headers, verb='POST')
+
+        return res
+
+    def _isValidSeverity(self, severity):
+        """
+        Used to check if the severity of the alert matches the valid levels: low, medium, high
+        :param severity: severity of the alert
+        :return: True or False
+        """
+        if severity not in self.validSeverity:
+            logging.critical("Alert submitted to AlertManagerAPI with invalid severity: %s", severity)
+            return False
+        return True


src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,123 @@
...
+      },
+        "startsAt": "<rfc3339>", # optional, will be current time if not present
+        "endsAt": "<rfc3339>",
+        "generatorURL": "<generator_url>"
+      },
+    ]
+    """
+
+    def __init__(self, configDict, alertName, severity, labels, end, generatorURL, annotations):
+        self.alertManagerUrl = configDict.get("alertManagerUrl", "http://cms-monitoring.cern.ch:30093/api/v1/alerts")
```
vkuznet, line 51: Erik, please do not hardcode anything. The code is in public repository and this leads to leaking our infrastructure details to entire world (the internal ports we use).

Instead, I suggest that code should throw exception if such parameter is not provided.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def send(self):
+        headers = {"Content-Type": "application/json"}
+        request = []
+        alert = {}
+
+        self.labels["alertname"] = self.alertName
+        alert["labels"] = self.labels
+        # add the sender's hostname as an annotation
+        self.annotations["hostname"] = socket.gethostname()
```
vkuznet, line 67: the socket operations is not cheap and hostname will not change in instance of this object. Therefore, it is better to initialize it once in ctor

vkuznet (changes requested):
And, I don't have strong preference for time function to be part of this codebase. You can split it from this code into separate module, which will be easy to test and share with other codebase if necessary.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
goughes: My bad, I had that there for my testing w/o the config object. Removed it.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,123 @@
...
+        mgr = RequestHandler()
+        res = mgr.getdata(self.alertManagerUrl, params=params, headers=headers, verb='POST')
+
+        return res
+
+# Below is a required python 2 class to determine current timezone
+# https://docs.python.org/2/library/datetime.html#tzinfo-objects
+# Everything below can be removed when we move to python3
+
+ZERO = timedelta(0)
```
amaltaro, line 90: Erik, I have some general comments on this time/timezone handling. Can we move it under:
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/Timers.py

and perhaps add these global variables inside an `__init__` method of the LocalTimeZone?

Given that it's a timezone information, it won't change during the process/service execution, so we could likely calculate it only once - in the `__init__` method of the `AlertManagerAPI` class and make that an object data (self.tz).

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,123 @@
...
+      },
+        "startsAt": "<rfc3339>", # optional, will be current time if not present
+        "endsAt": "<rfc3339>",
+        "generatorURL": "<generator_url>"
+      },
+    ]
+    """
+
+    def __init__(self, configDict, alertName, severity, labels, end, generatorURL, annotations):
+        self.alertManagerUrl = configDict.get("alertManagerUrl")
```
amaltaro, line 51: Should we convert this configDict configuration object into a plain string url?

amaltaro (changes requested):
Erik, the way I see this API being used is:
* we initialize it when initializing a service that can use it (thus, all the defaults that won't change, should be set as an object variable - including the RequestHandler object);
* when we want to trigger an alert, we call the already initialized AlertManagerAPI object and send the alert; 
  * in the alert itself, we could have a call to another function to put the timestamp in the required standard
  * I think it would be important to also have some input data validation/sanitization and ensure that the user provided values supported by AM (like the severity in [low, medium, high]

In short, it would be a very similar usage compared to the current `EmailAlert` module.
Please let me know if I'm missing something here. Once we converge on the logical changes, I'll come back here with some aesthetic suggestions ;)

vkuznet:
+1

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,94 @@
...
+            ...
+          },
+            "startsAt": "<rfc3339>", # optional, will be current time if not present
+            "endsAt": "<rfc3339>",
+            "generatorURL": "<generator_url>" # optional
+          },
+        ]
+        """
+
+        validSeverity = ["high", "medium", "low"]
```
amaltaro, line 57: How about making it an object attribute in the init method (self.validSeverity)? Perhaps we could even have a method like:
```
def _isSeverityValid(self, severity):
    if severity not in validSeverity:
        logging.critical("Alert submitted to AlertManagerAPI with invalid severity: %s", severity)
        return False
    return True
```

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,94 @@
...
+from WMCore.Services.pycurl_manager import RequestHandler
+from Utils.Timers import LocalTimezone
+
+
+class AlertManagerAPI(object):
+    """
+    A class used to send alerts via the MONIT AlertManager API
+    """
+
+    def __init__(self, alertManagerUrl):
```
amaltaro, line 21: We should accept an optional logger object here, if nothing is provided, we get the root logger. Something like:
```
self.logger = logger if logger else logging.getLogger()
```

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,94 @@
...
+        self.alertManagerUrl = alertManagerUrl
+        # sender's hostname is added as an annotation
+        self.hostname = socket.gethostname()
+        self.mgr = RequestHandler()
+        self.ltz = LocalTimezone()
+
+    def send(self, alertName, severity, summary, description, tag="wmcore", service="", endSecs=600, generatorURL=""):
```
amaltaro, line 28: Perhaps, we could give an even more explicit name to this method, e.g. `sendAlert`. What do you think?
The docstring is also missing the tag parameter. 
Last comment here is actually a question, as I look at it, I wonder if we should simply make `service` a mandatory argument?

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    A class used to send alerts via the MONIT AlertManager API
+    """
+
+    def __init__(self, alertManagerUrl):
+        self.alertManagerUrl = alertManagerUrl
+        # sender's hostname is added as an annotation
+        self.hostname = socket.gethostname()
+        self.mgr = RequestHandler()
+        self.ltz = LocalTimezone()
```
amaltaro, line 26: How about creating another object attribute `self.generatorURL` with the alertmanager front page as default? Then on the `send` method, we default that to None, and if no value is provided, we fallback to the dafault value in the instance.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,94 @@
...
+          },
+        ]
+        """
+
+        validSeverity = ["high", "medium", "low"]
+        if severity not in validSeverity:
+            logging.critical("Alert submitted to AlertManagerAPI with invalid severity: %s", severity)
+            return
+
+        headers = {"Content-Type": "application/json"}
```
amaltaro, line 62: I would move this `headers` to the init method as well. It's not going to change anyways.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # add annotations
+        annotations["hostname"] = self.hostname
+        annotations["summary"] = summary
+        annotations["description"] = description
+        alert["annotations"] = annotations
+
+        # In python3 we won't need the LocalTimezone class
+        # Will change to d = datetime.now().astimezone() + timedelta(seconds = endSecs)
+        d = datetime.now(self.ltz) + timedelta(seconds = endSecs)
```
amaltaro, line 83: Given that it's a brand new module, we should clean all the pylints, if possible.
The report has something for you on this line ;)

amaltaro (changes requested):
Erik, code is looking good. However, I have made several suggestions in the code; and a few questions as well.

Documentation of the new class is good as well, but I think we can enhance it a bit more (like, we could mention which data type is expected for those args in the send() method). This is especially important because these classes under Utils are commonly shared with other DMWM projects.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
goughes: Service being mandatory is probably a good idea. We could go one step further and make a list of "valid" services to check against, since the service name will be used to route alerts.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
goughes: I'm not sure about setting `generatorURL` to the AlertManager page. Unless this is sent in the body of the email, it would just be a link to the AlertManager page that is clickable from the AlertManager page.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
amaltaro: I like the idea of validating the service against a list of known and expected services (properly capitalized or not). The only problem will be if others, e.g. CRAB, decide to use this same module. 
Ideally the list of valid service names should be configurable, but I fail to see how to enforce that ... and the only feasible logic I see is hard-wiring it in the source code and updating it in the future, if needed. What are your thoughts here?

src/python/Utils/Timers.py
```diff
@@ -52,3 +53,47 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    """
+    A required python 2 class to determine current timezone for formatting rfc3339 timestamps
+    https://docs.python.org/2/library/datetime.html#tzinfo-objects
+    Required for sending alerts to the MONIT AlertManager
+    Can be removed once WMCore starts using python3
+    """
+
+    def __init__(self):
+        super(LocalTimezone, self).__init__()
+        self.ZERO = timedelta(0)
```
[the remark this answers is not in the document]
goughes: I think we should never use this class except for alerts and we should forget about it once we can use the builtin functions from python3. I included a link to the python docs that include this class as an example.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
goughes: For now we should probably just trust ourselves to use the correct names instead of putting in additional checks. Something to think about if we find we are unable to handle that.

src/python/WMCore/Services/AlertManager/AlertManagerAPI.py
amaltaro: I assume this `generatorURL` is optional then? We can simply pass the default empty string from the code below and everything would work, right?

goughes: Yes, I tested with a blank string and things work. Alert shows up with no "Source" on the AM page.

amaltaro (approved):
Code looks good to me, Erik.

I still have a couple of questions/comments though:
* could you please file another GH issue asking for the removal of this LocalTimezone class and the one line update in the alert manager API (post python3 migration)? Just so we don't forget it.
* do we need Valentin's help in setting up things on the alerting system? We can also follow this up offline or on slack.
* I think we could likely persist some of the content in your gdoc in a WMCore wiki; also providing some guidelines for using this API and properly defining alerts within WMCore. If you prefer, I can draft the first version of such wiki, please let me know.

vkuznet:
@amaltaro , I suggest that for all your monitoring needs to open CMS Monitoring Jira ticket. This way any of us (me, Ceyhun, Federica) can work on it. All of has access to configuration files for AM and can add appropriate settings for your use-cases. Once again, we only need to know which tag, emails, slack channels should be used.

vkuznet:
Just for the record. We do have very handy `alert` CLI tool which is available on cvmfs, see `/cvmfs/cms.cern.ch/cmsmon/alert`. You may use it to look-up all your alerts, just run it with `-help` option and you can see how it works. Here are just few examples:
```
# Get all alerts based on multi filters. Ex service=GGUS, severity=high:
            alert -service=GGUS -severity=high

# Get all alerts of specific service/severity/tag/name. Ex GGUS/high/cms/ssb-OTG0058113:
            alert -service=GGUS
            alert -severity=high
            alert -tag=cms
            alert -name=ssb-OTG0058113
```
Therefore, once you confirm your configuration and settle down on tags, etc., you may start using it for your alerts.

The alerts can be viewed in plain (human friendly) and json formats. The tool can be used on any linux, and inside or outside of CERN network (for later I just committed the update and it will require a valid token).

I always prefer CLI solution over web based one and our alerting infrastructure provides that.

amaltaro:
Thanks Valentin. I added this information to our WMCore wiki as well: https://github.com/dmwm/WMCore/wiki/Prometheus-AlertManager-wrapper-APIs#listing-alerts-with-cli-tools

vkuznet:
ok, and to clarify one common question. The AlertManager does not have API to delete alerts, instead it is achieved in different ways. If alert was fired by Prometheus and its rules come back to normal Prometheus takes care of its resolution. For manually inserted alerts (like your use case), the alert can be deleted by simply adjusting its EndsAt timestamp to current time or time in a past. Our `alert` tool will do that (I submitted corresponding PR https://github.com/cms-sw/cmsdist/pull/6750). Or, you can use your AM wrapper to do this job. The way it is done is following. First, you fetch the alert JSON, then adjust its EndsAt value and POST it back to AM. In other words AM only has GET and POST APIs.

Also, I included in my PR a new tool called token-manager, which will allow to handle (renew) access tokens. With such token AM can be accessed from outside of CERN, i.e. you can inject, inspect or update alerts from distributed workflows. But this is a topic of separate discussion.

PR #10309, opened 2021-02-25
Validate status transition for cascade option

### LINKED ISSUE #10217: Forbidden request status transition from aborted-archived to closed-out
**Impact of the bug**
ReqMgr2

**Describe the bug**
While debugging MSRuleCleaner, we have found a strange workflow status transition for:
https://cmsweb.cern.ch/reqmgr2/fetch?rid=cmsunified_ACDC4_Run2017H-v1_LowEGJet_09Aug2019_UL2017_LowPU_200901_093632_7002

which should be forbidden, according to the allowed status transition:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/ReqMgr/DataStructs/RequestStatus.py#L55

**How to reproduce it**
We can try it via RESTful API.

**Expected behavior**
We need to *always* validate the status transition and make sure it complies with the allowed status transition defined in the ReqMgr2 DataStructs (no matter how the status transition is triggered, via web UI or RESTful API).

I'm also in contact with Hasan to see why Unified did such transition and whether there is any valid use case for that.

logs are likely no longer available.

amaltaro:
Fixes #10217 

#### Status
ready

#### Description
This PR provides the following changes:
* update a ReqMgr2 CouchDB view (need to make sure it gets properly pushed in in CMSWEB, after the migration to k8s)
* before updating the children workflows status, check whether the status transition is allowed. If not, create a record in the logs and move to the next workflow.

yes

#### Related PRs
none

none

### DIFF
--- a/src/couchapps/ReqMgr/views/childresubmissionrequests/map.js
+++ b/src/couchapps/ReqMgr/views/childresubmissionrequests/map.js
@@ -5,6 +5,6 @@
  */
 function(doc){
     if(doc.InitialTaskPath){
-        emit(doc.InitialTaskPath.split("/")[1], null);
+        emit(doc.InitialTaskPath.split("/")[1], doc.RequestStatus);
     }
 }
--- a/src/python/WMCore/ReqMgr/Service/Request.py
+++ b/src/python/WMCore/ReqMgr/Service/Request.py
@@ -18,8 +18,8 @@ from WMCore.REST.Auth import get_user_info
 
 from WMCore.ReqMgr.DataStructs.ReqMgrConfigDataCache import ReqMgrConfigDataCache
 from WMCore.ReqMgr.DataStructs.RequestError import InvalidSpecParameterValue
-from WMCore.ReqMgr.DataStructs.RequestStatus import (REQUEST_STATE_LIST,
-                                                     REQUEST_STATE_TRANSITION, ACTIVE_STATUS)
+from WMCore.ReqMgr.DataStructs.RequestStatus import (REQUEST_STATE_LIST, REQUEST_STATE_TRANSITION,
+                                                     ACTIVE_STATUS, check_allowed_transition)
 from WMCore.ReqMgr.DataStructs.RequestType import REQUEST_TYPES
 from WMCore.ReqMgr.Utils.Validation import (validate_request_create_args, validate_request_update_args,
                                             validate_clone_create_args, validateOutputDatasets,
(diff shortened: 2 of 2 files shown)


amaltaro:
Tests went fine in my VM, but given that tomorrow is already Friday, I'm inclined to push it in only in the next cycle (beginning of April).

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -475,19 +483,32 @@ def _handleOnlyStateTransition(self, workload, request_args, dn):
...
+            msg = "Workflow {} has {} ".format(workload.name(), len(childrenNamesAndStatus))
+            msg += "children workflows to have a status transition to: {}".format(req_status)
+            cherrypy.log(msg)
+            for childInfo in childrenNamesAndStatus:
+                if check_allowed_transition(childInfo['value'], req_status):
+                    cherrypy.log('Updating request status for {} to {}.'.format(childInfo['id'], req_status))
+                    self.reqmgr_db_service.updateRequestStatus(childInfo['id'], req_status, dn)
+                else:
+                    msg = "Status transition from {} to {} ".format(childInfo['value'], req_status)
+                    msg += "not allowed for workflow: {}, skippint it!".format(childInfo['id'])
```
todor-ivanov, line 508: `skippint` -> `skipping` ;)

src/couchapps/ReqMgr/views/childresubmissionrequests/map.js
```diff
@@ -5,6 +5,6 @@
  */
 function(doc){
     if(doc.InitialTaskPath){
-        emit(doc.InitialTaskPath.split("/")[1], null);
+        emit(doc.InitialTaskPath.split("/")[1], doc.RequestStatus);
```
todor-ivanov, line 8: Just for my education. Who is capturing the so 'emitted' event by this java script? And is the set of actions it is about trigger changed somehow when the RequestStatus is included in it, or this is added here just for clarity?

src/python/WMCore/ReqMgr/Service/Request.py
```diff
@@ -475,19 +483,32 @@ def _handleOnlyStateTransition(self, workload, request_args, dn):
...
+                    cherrypy.log(msg)
+        # then update the original/parent workflow status in couchdb
+        cherrypy.log('Updating request status for {} to {}.'.format(workload.name(), req_status))
         report = self.reqmgr_db_service.updateRequestStatus(workload.name(), req_status, dn)
```
todor-ivanov, line 512: I know this line is not part of this commit, but just a a general comment - I do not like the attitude of triggering actions during assignments. BTW I am aware nothing can be done here in this regard, so this comment is not a request for a change.

todor-ivanov (commented):
Looks good to me. Especially after @amaltaro mentioned it is already tested and working. I just made a comment or two, which can even be ignored.

src/couchapps/ReqMgr/views/childresubmissionrequests/map.js
amaltaro: This is the code for the so-called CouchDB views, which can contain either a map or a map and reduce function.
It's provided through the CouchDB HTTP APIs, like:
https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/_design/ReqMgr/_view/childresubmissionrequests?key=%22cmsunified_ACDC1_r-0-Run2017E_DoubleMuon_UL2017_MiniAODv2_210120_130131_8395%22

Seeing the response of this API should make it clear what the change will be. Please let me know otherwise ;)

src/python/WMCore/ReqMgr/Service/Request.py
amaltaro: Good catch! Fixing it soon!

src/python/WMCore/ReqMgr/Service/Request.py
amaltaro: Don't see it as triggering an action, but as calling a function and retrieving the returned response ;)

todor-ivanov (approved):
looks good

PR #10327, opened 2021-03-04
Add check and different path for wmagent component.

### LINKED ISSUE #10326: Fix path differences between WMAgent and WMCore components in patchComponent.py
**Impact of the bug**
The process of patching different components

**Describe the bug**
Not exactly a bug, but instead a call for a late fix of an already existing `TODO` comment. 
The script for patching different components `bin/patchComponent.py` so far was working only for central services because of the differences of the root path between WMCore central services and WMAgent deployments.   

**How to reproduce it**
Try to patch a wmagent with the currently existing script

**Expected behavior**
The script should recognize the type of component it is about to patch and apply the relevant changes to the `rootPath` variable.

No

todor-ivanov:
Fixes #10326 

#### Status
ready

#### Description
Adds a check if the current component to be patched is wmagent and if so applies the relevant changes to the `rootPath`

YES 

#### Related PRs
NO

NO

### DIFF
--- a/bin/patchComponent.sh
+++ b/bin/patchComponent.sh
@@ -5,8 +5,8 @@ usage()
     echo -ne "\nA simple script to facilitate component patching\n"
     echo -ne "and to decrease the development && testing turnaround time.\n"
     echo -ne "Usage: \n"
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
Looks good to me, thanks Todor.

Issue #10323, opened 2021-03-03
[py2py3] test str/bytes use in WMCore (wildcard issue)


mapellidario:
**Impact of the bug**

It should involve `src/python/WMCore/MicroService/DataStructs/MSRuleCleanerWflow.py` only

**Describe the bug**
We have not observed any bug at the moment, but this is very likely to create some problems.

(add here all the checks about str/bytes that did not fell into another issue)

- [ ]  `MicroService/DataStructs/MSRuleCleanerWflow.py` 

    In #10307 we made the following changes

    ```python
    -                elif (isinstance(data['type'], tuple) and (str in data['type'] or unicode in data['type'])) or \
    -                     (data['type'] is str or data['type'] is unicode):
    +               elif (isinstance(data['type'], tuple) and (bytes in data['type'] or str in data['type'])) or \
    +                    (data['type'] is bytes or data['type'] is str):
    ```

- [ ] #10321 `src/python/WMCore/JobStateMachine/ChangeState.py`
- [ ] #10321 `src/python/WMCore/WMSpec/WMTask.py`
- [ ] #10329 `src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py`

**Expected behavior**
These are supposed to behave the same in py2 and py3 as he previous lines, but there is a somewhat high probability that they will not. We should make sure that these lines behave properly when we have a py3 environment.

amaltaro:
The 4 modules mentioned above have all been tested and are already in production as well. I think it's safe to close this issue, and if a problem is found in the near future, we can always either reopen it or create a new specific issue.
Closing it.

PR #10321, opened 2021-03-02
[py2py3] Migration at level scr/python/A/B/C - slice 16

mapellidario:
Fixes #10140

#### Status
ready

#### Related PRs

* Previous migration PR: #10319  (slice15, issue #10139  )
* Next migration PR: #10329  (slice17, issue #10141 )

#### Description

Run futurize and some manual changes on the first batch of src/python/A/B/C.

Notable changes:
* `src/python/WMCore/JobStateMachine/ChangeState.py`: used native-string approach. should be mostly safe, but i am not 100% sure. tracked in #10323
* `src/python/WMCore/WMSpec/WMTask.py`: used native-string approach. should be mostly safe, but i am not 100% sure. Tracked in #10323 

#### Is it backward compatible (if not, which system it affects?)

It should be. (Any possible cause for errors will we reported here)

#### External dependencies / deployment changes

Requires python-future in both py2 and py3 environments.

Issue #10142, opened 2020-12-03
[py2py3] Migration at level scr/python/A/B/C - slice 18


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 18-th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 18-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
                                             name  nfiles  gradual  future_pm
                    WMCore_JobSplitting_LumiBased       1     True         21
    WMComponent_AgentStatusWatcher_DrainStatusAPI       1     True          1
WMComponent_AnalyticsDataCollector_DataCollectAPI       1     True         12
       WMComponent_JobAccountant_AccountantWorker       1     True          5
          WMCore_JobSplitting_EventAwareLumiBased       1     True          5
                              WMCore_ReqMgr_Tools       1     True          1
                      WMCore_JobSplitting_Harvest       1     True         14
                        WMCore_ReqMgr_ReqMgrCouch       1     True          1
        WMCore_WorkQueue_WorkQueueReqMgrInterface       1     True          4
            WMCore_WorkerThreads_BaseWorkerThread       1     True          2
                               WMCore_WMBS_Oracle     225     True          5
                    WMCore_JobSplitting_FileBased       1     True          6
                             WMCore_DQMCat_Oracle       1     True          0
                 WMCore_REST_CherryPyPeriodicTask       1     True          2
                       WMCore_WMStats_DataStructs       1     True          8
                               WMCore_ReqMgr_Auth       1     True          0
              WMCore_WMRuntime_DashboardInterface       1     True          1
                 WMQuality_Emulators_PhEDExClient       1     True          3
                  WMQuality_Emulators_RucioClient       1     True          1
                         WMCore_Storage_StoreFail       1     True          4
                                WMCore_WMBS_T0AST       1     True          0
```

#### Is your feature request related to a problem? Please describe.

The end goal is to have WMCore py2 and py3 compatible.

#### Describe the solution you'd like

Run futurize on the first bulk of units 
and then apply the necessary manual changes.

Issue #10140, opened 2020-12-03
[py2py3] Migration at level scr/python/A/B/C - slice 16


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 16-th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 16-th bulk of migration units 
at src/python/A/B/C level.

```plaintext
                               name  nfiles  gradual  future_pm
 WMCore_JobStateMachine_ChangeState       1    False          3
WMCore_WorkQueue_DataLocationMapper       1    False         22
      WMCore_FwkJobReport_ReportEmu       1    False          6
         WMCore_Storage_FileManager       1    False          7
      WMCore_WMSpec_WMWorkloadTools       1    False          4
               WMCore_WMSpec_WMTask       1    False         28
          WMCore_Services_WorkQueue       1    False          4
      WMCore_FwkJobReport_XMLParser       1    False          4
          WMCore_BossAir_BossAirAPI       1    False         25
  WMCore_WorkQueue_WorkQueueBackend       1    False         12
```

#### Is your feature request related to a problem? Please describe.

The end goal is to have WMCore py2 and py3 compatible.

#### Describe the solution you'd like

Run futurize on the first bulk of units 
and then apply the necessary manual changes.

PR #10329, opened 2021-03-04
[py2py3] Migration at level scr/python/A/B/C - slice 17

### LINKED ISSUE #10141: [py2py3] Migration at level scr/python/A/B/C - slice 17
#### Intro

This is the 17-th slice in which the migration has been divided into. 

We start migrating the 17-th bulk of migration units 
at src/python/A/B/C level.

 
```plaintext
       WMCore_WorkQueue_WorkQueueUtils       1     True          0
              WMCore_WMSpec_WMWorkload       1    False         24
             WMCore_Services_ReqMgrAux       1    False          0
WMCore_ResourceControl_ResourceControl       1    False          1
            WMCore_Storage_StageOutMgr       1    False         10
            WMCore_FwkJobReport_Report       1    False         23
           WMCore_WorkQueue_WMBSHelper       1    False         13
                WMCore_WMRuntime_Tools       1    False          5
       WMCore_WMRuntime_SandboxCreator       1    False          4
             WMCore_WMRuntime_Watchdog       1     True          3
              WMCore_WMRuntime_Scripts       2     True          5
 WMComponent_JobCreator_CreateWorkArea       1     True         10
```

mapellidario:
Fixes #10141 

#### Status

Ready 

#### Related PRs

* Previous migration PR: #10321  (slice16, issue #10140)
* Next migration PR: #10331  (slice18, issue #10142  )

#### Description

"native string" approach
- `src/python/WMCore/FwkJobReport/Report.py`: imported str from builtins, but used only in a narrow case. 
  - `addError`: refactored. See #8403 for previous work
  - refactored heavily
- `src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.p`: there is only one line where this may be problematic: `inputTypeAttrib.fileNames.append(str(fileLFN))`, but I do not think that there is any incentive to try to change it at this stage. tracked at #10323 

### DIFF
--- a/src/python/WMCore/FwkJobReport/Report.py
+++ b/src/python/WMCore/FwkJobReport/Report.py
@@ -6,6 +6,8 @@ _Report_
 Framework job report object.
 """
 from __future__ import print_function
+from builtins import str as newstr, bytes, range, object
+from future.utils import viewitems, listitems
 
 import logging
 import math
@@ -14,6 +16,7 @@ import sys
 import time
 import traceback
 
+from Utils.Utilities import encodeUnicodeToBytes, decodeBytesToUnicode
 from WMCore.Configuration import ConfigSection
 from WMCore.DataStructs.File import File
 from WMCore.DataStructs.Run import Run
@@ -88,7 +91,7 @@ def addAttributesToFile(fileSection, **attributes):
 
     Add attributes to a file in the FWJR.
     """
-    for attName, attValue in attributes.items():
+    for attName, attValue in viewitems(attributes):
         setattr(fileSection, attName, attValue)
     return
 
@@ -197,7 +200,7 @@ class Report(object):
                 continue
 
             jsonPerformance[reportSection] = getattr(perfSection, reportSection).dictionary_()
-            for key in jsonPerformance[reportSection].keys():
+            for key in jsonPerformance[reportSection]:
                 val = jsonPerformance[reportSection][key]
                 if isinstance(val, float):
                     if math.isinf(val) or math.isnan(val):
@@ -447,7 +450,7 @@ class Report(object):
 
         # Now we need to eliminate the optional and non-primitives:
         # runs, parents, branches, locations and datasets
-        keyList = aFile.keys()
+        keyList = list(aFile)
 
         fileRef.section_("runs")
         if "runs" in aFile:
@@ -511,7 +514,7 @@ class Report(object):
         fileRef = getattr(srcMod.files, fileSection)
         srcMod.files.fileCount += 1
 
-        keyList = attrs.keys()
+        keyList = list(attrs)
 
         fileRef.section_("runs")
         if "runs" in attrs:
@@ -544,7 +547,7 @@ class Report(object):
         newFile = getattr(analysisFiles, label)
         newFile.fileName = filename
 
-        for x, y in attrs.items():
+        for x, y in viewitems(attrs):
             setattr(newFile, x, y)
 
         analysisFiles.fileCount += 1
@@ -563,7 +566,7 @@ class Report(object):
         removedFiles.section_(label)
         newFile = getattr(removedFiles, label)
 
-        for x, y in attrs.items():
+        for x, y in viewitems(attrs):
             setattr(newFile, x, y)
 
         self.report.cleanup.removed.fileCount += 1
@@ -599,15 +602,19 @@ class Report(object):
         errDetails.type = str(errorType)
 
         try:
-            if hasattr(errorDetails, "decode"):
-                # Fix for the unicode encoding issue, #8043
-                # interprets this string using utf-8 codec and ignoring any errors
-                errDetails.details = errorDetails.decode('utf-8', 'ignore')
+            if isinstance(errorDetails, newstr):
+                errDetails.details = errorDetails
+            elif isinstance(errorDetails, bytes):
+                errDetails.details = decodeBytesToUnicode(errorDetails, 'ignore')
             else:
-                # Then cast it to string and decode it
-                errorDetails = str(errorDetails)
-                errDetails.details = errorDetails.decode('utf-8', 'ignore')
+                errDetails.details = newstr(errorDetails)
         except UnicodeEncodeError as ex:
+            msg = "Failed to encode the job error details for job ID: %s." % self.getJobID()
+            msg += "\nException message: %s\nOriginal error details: %s" % (str(ex), errorDetails)
+            logging.error(msg)
+            msg = "DEFAULT ERROR MESSAGE, because it failed to UTF-8 encode the original message."
+            errDetails.details = msg
+        except UnicodeDecodeError as ex:
             msg = "Failed to decode the job error details for job ID: %s." % self.getJobID()
             msg += "\nException message: %s\nOriginal error details: %s" % (str(ex), errorDetails)
             logging.error(msg)
@@ -786,7 +793,7 @@ class Report(object):
         for run in runList:
             lumis = getattr(fileRef.runs, run)
             if isinstance(lumis, dict):
-                newRun = Run(int(run), *lumis.items())
+                newRun = Run(int(run), *listitems(lumis))
             else:
                 newRun = Run(int(run), *lumis)
             newFile.addRun(newRun)
@@ -1496,7 +1503,7 @@ class Report(object):
         error = None
         files = self.getAllFilesFromStep(step=stepName)
         for f in files:
-            if 'adler32' not in f.get('checksums', {}).keys():
+            if 'adler32' not in f.get('checksums', {}):
                 error = f.get('lfn', None)
             elif f['checksums']['adler32'] is None:
                 error = f.get('lfn', None)
--- a/src/python/WMCore/Storage/StageOutMgr.py
+++ b/src/python/WMCore/Storage/StageOutMgr.py
@@ -8,6 +8,8 @@ Based of RuntimeStageOut.StageOutManager, that should probably eventually
 use this class as a basic API
 """
 from __future__ import print_function
+from builtins import object
+from future.utils import viewitems
 
 import logging
 # If we don't import them, they cannot be ever used (bad PyCharm!)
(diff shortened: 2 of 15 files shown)


src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -599,16 +601,11 @@ def addError(self, stepName, exitCode, errorType, errorDetails, siteName=None):
...
-                errDetails.details = errorDetails.decode('utf-8', 'ignore')
-            else:
-                # Then cast it to string and decode it
-                errorDetails = str(errorDetails)
-                errDetails.details = errorDetails.decode('utf-8', 'ignore')
+            errDetails.details = newstr(errorDetails)
         except UnicodeEncodeError as ex:
             msg = "Failed to decode the job error details for job ID: %s." % self.getJobID()
+            # if the `try` fails in py3, then the following line will also
+            # fail in python3 and there is not much that we can do
```
mapellidario, line 608: Alan, we can remove this comment if you think it is irrelevant!

test/python/WMCore_t/FwkJobReport_t/Report_t.py
```diff
@@ -584,6 +588,27 @@ def testExitCode(self):
...
+        self.assertItemsEqual(report.getStepExitCodes(stepName="cmsRun1"), {99999, 12345, 123})
+        self.assertEqual(report.getStepErrors(stepName="cmsRun1")['errorCount'], 3)
+
+        # and once again, but different type and details (which does not matter) - testing unicode handling
+        report.addError(stepName="cmsRun1", exitCode=12345, errorType=encodeUnicodeToBytes("⊥ℯṧтÅÅ"), errorDetails=encodeUnicodeToBytes("тℯṧт"))
+        self.assertEqual(report.getExitCode(), 12345)
+        self.assertEqual(report.getStepExitCode(stepName="cmsRun1"), 12345)
+        self.assertItemsEqual(report.getStepExitCodes(stepName="cmsRun1"), {99999, 12345, 123})
+        self.assertEqual(report.getStepErrors(stepName="cmsRun1")['errorCount'], 3)
+
```
mapellidario, line 611: Trying to pass both unicode and bytes strings containing non-ascii chracters into `Report.addError`, both in py2 and py3

src/python/WMCore/FwkJobReport/Report.py
amaltaro: Is it because it will fail to stringify `errorDetails` in the error message?

src/python/WMCore/FwkJobReport/Report.py
```diff
@@ -599,16 +601,11 @@ def addError(self, stepName, exitCode, errorType, errorDetails, siteName=None):
...
         try:
-            if hasattr(errorDetails, "decode"):
-                # Fix for the unicode encoding issue, #8043
-                # interprets this string using utf-8 codec and ignoring any errors
-                errDetails.details = errorDetails.decode('utf-8', 'ignore')
-            else:
-                # Then cast it to string and decode it
-                errorDetails = str(errorDetails)
-                errDetails.details = errorDetails.decode('utf-8', 'ignore')
+            errDetails.details = newstr(errorDetails)
```
amaltaro, line 604: Dario, can you educate me here why we no longer need to decode the "text" provided in the errorDetails? Would it work with those weird bytes/words mentioned in the previous GH issues (like `0xff` or `\x95`)?

This code has given us so much headache in the past, that I'm very concerned of having regressions.

src/python/WMCore/Storage/StageOutMgr.py
```diff
@@ -188,7 +190,7 @@ def __call__(self, fileToStage):
         Use call to invoke transfers
 
         """
-        lastException = ""
+        lastException = None
```
amaltaro, line 193: Is this change causing the following error to be reported in the pylint section:
```
 E0702, line 253 in StageOutMgr.__call__: Raising NoneType while only classes or instances are allowed 
```
?
If so, I'd then suggest to switch it back to empty string. AFAICT there is no difference

amaltaro (approved):
Dario, these changes look good to me. However, I have asked a few questions in the code for a few cases that I didn't manage to understand.

src/python/WMCore/FwkJobReport/Report.py
mapellidario: I do feel your headache! The original discussion was in #8403

What you are doing here is

```plaintext
if errorDetails is bytes or unicode 
  # however only bytes should have decode attribute
  # if errorDetails is unicode and it contains only ascii, then it is decoded back to itself
  errDetails.details = unicode(errorDetails)
else
  # here we can have dict, list, ...
  errorDetails = bytes(errorDetails)
  errDetails.details = unicode(errorDetails)
```

My approach would be

```plaintext
errDetails.details = unicode(errorDetails)
```

Regarding malformed bytes, you are correct. Those would break `newstr(errorDetails)`, but that is why we have the try/catch clause! However, now that I stare at it, I am puzzled. It catches "UnicodeEncodeError", but the message states "Failed to **decode** the job error details for job ID: %s.". 

Maybe you are correct, I should give this another try!

src/python/WMCore/Storage/StageOutMgr.py
amaltaro: I see. There was already a pylint report on that same line.
In this case, feel free to leave the original code `""` or set it to the standard exception that you suggest. That default value will never be raised anyways :)

src/python/WMCore/FwkJobReport/Report.py
mapellidario: The plot thickens!

I opted for

```python
from builtins import str as newstr, bytes
...
        try:
            if isinstance(errorDetails, newstr):
                errDetails.details = errorDetails
            elif isinstance(errorDetails, bytes):
                errDetails.details = errorDetails.decode('utf8', 'ignore')
            else:
                errDetails.details = newstr(errorDetails)
``` 

but even if `errorDetails` were py2-`str` (here also `bytes`), then it would fail with 
> UnicodeEncodeError: 'ascii' codec can't encode characters in position 2-5: ordinal not in range(128)
which made no sense because I was trying to decode, not encode bytes.

I after a full hour, I removed the `try` and realized that it was failing here https://github.com/dmwm/WMCore/blob/595a9934aef084b3aaf5ca93c55a145c405edb43/src/python/WMCore/Configuration.py#L141. The bytes was successfully converted to unicode, but then was converted back to bytes with ascii encoding by `configuration.py`

I then decided to change `Configuration` a bit to ease or near-future work with using `Utils.Utilities.encodeUnicodeToBytes()`. This is a temporary fix, we will likely be better of changing this when switching to py3.

Moreover this raises a question: do we really need to decode to unicode in `addError`, if then it would be converted to bytes anyway by `Configuration`?

src/python/WMCore/Storage/StageOutMgr.py
mapellidario: Perfect, thank you!

mapellidario:
(I just noticed that `pylint --py3k` complains about something in `Configuration.py`. I would ignore it and postpone that work to #9973 )

test/python/WMCore_t/FwkJobReport_t/Report_t.py
mapellidario: Improved in later commit, marking this as resolved

mapellidario:
Eventually, I am the one who changes opinion more often!

Since you (rightfully!) prefer to sanitize the error details as soon as it enters `WMCore/FwkJobReport/Report.addError`, and since the simplest way to do so would be to decode the bytes string into unicode and encode it back to bytes ignoring errors along the way, I decided that we can also make `WMCore/FwkJobReport/Report.addError` compatible with py3 and have it send unicode into `WMCore/Configuration.ConfigSection.__setattr__`. In this way, when we will modernize `WMCore/Configuration`, we will not need to change `Report.addError` as well.

One small treat: instead of ignoring errors, we could replace the errors with the unicode character `�` so that it is clear in the logs (to these error details end up in logs?) when and where there is some malformed byte.

In this moment `Configuration` is not modernized and I changed it only a bit to make it encode the incoming unicode with `utf8` and not `ascii`, using `Utils.Utilities.encodeUnicodeToBytes`.

I added to the unit test an error detail that is neither unicode nor bytes, it is a `dict`. I am not sure if we ever would encounter this, but better be safe then sorry. I also added malformed bytes in the dictionary. `newstr` does what it is supposed to and does not fail us!

```plaintext
test
test
test
test
testAA
1 тℯṧт
2 тℯṧт �
3 тℯṧт
4 тℯṧт
b'{'\xd0\xbd\xce\xb5\xca\x9f\xca\x9f\xe2\x97\x8e': 3.14159}'
b'{'\xd0\xbd\xce\xb5\xca\x9f\xca\x9f\xe2\x97\x8e \x95': '\xe1\xba\x98\xc3\xb8\xd2\x91\xe2\x84\x93\xe2\x88\x82 \x95'}'

```

Notice that the questionmark now is not the print statement that does not know what to do with a malformed byte, it is exactly that unicode character, which replaced the malformed byte in `errDetails.details = decodeBytesToUnicode(errorDetails, 'replace')`.

Notice also that the dictionary key and value containing the utf8-invalid byte `\x95` did not cause any error and the byte `\x95` is carried along. I am not sure if this will be the same in py3, we will have to wait to know this answer.

amaltaro:
> One small treat: instead of ignoring errors, we could replace the errors with the unicode character � so that it is clear in the logs (to these error details end up in logs?) when and where there is some malformed byte.

@mapellidario Dario, IMO we should simply drop those malformed characteres. It will usually be part of a long (or very long) error information that happened during job runtime and I don't see much value in keeping the question mark char in place.
What do you think?

mapellidario:
Yeah sure! I mean, the last call is always up to you! I made a new commit replacing 'replace' with 'ignore' and squashed the review commits! Now this is all for you to review (again)!

mapellidario:
@amaltaro you are the one that should make the last call! I replaced `replace` with `ignore`!
I pushed two new commits that contain all the changes after your first review, these are ready to be review as well! Thanks again for the feedback!

amaltaro:
Thanks Dario. Last 2 commits look good to me as well. Can you please squash them accordingly? Thanks

amaltaro (approved):

amaltaro:
Thanks Dario.

PR #10339, opened 2021-03-08
allow to run REST/Main in frontend via env.var. fix #10266

### LINKED ISSUE #10266: make it easy to run WMCore REST in debug mode
**Impact of the new feature**
every develops who uses WMCore REST will benefit

At times I really need to run CRABServer (i.e. a REST application based on WCORE/REST) in the foreground, in single thread mode, with all messages coming to console and with full `pdb` access.
Maybe WMCore devbelopers already have a trick to do this ? If so, please point me to it !
Otherwise... keep reading.

we have a recipe to do this which works since years, which simply requires commenting out severl lines,
but it is annoying to edit code all of the times:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/Crab3OperatorDebugging#To_run_the_crabserver_in_debug_m

So I propose to add an ENV variable switch inside Main.py to skip code as needed.
It should be pretty simple and safe. I can provide a tested PR if you like the idea.

having a "ready to swap MainForDbg.py" at hand, but I need to create it again every time Main,py is changed
Also, to avoid copying in/out of the container all of the times, I should likely hide a script inside CRABServer build which accomplishes this on demand. Ugly to say the least !!!

belforte:
Fixes #10266

#### Status
Ready. Tested and in use by Stefano for CRABServer debugging

#### Description
when the enviromental variable DONT_DAEMONIZE_REST is set and equal to "True" the REST will run in foreground with output to console. Otherwise it behaves exactly like now.

YES

### DIFF
--- a/src/python/WMCore/REST/Main.py
+++ b/src/python/WMCore/REST/Main.py
@@ -387,41 +387,48 @@ class RESTDaemon(RESTMain):
 
     def start_daemon(self):
         """Start the deamon."""
+        # This REST server can be run in the frontend i.e. interactively (e.g. to use pdb)
+        # by setting an env. var. via: export DONT_DAEMONIZE_REST=True
+        # if the variable is missing or set to any other value, code starts normally as a daemon
+        daemonize = os.getenv('DONT_DAEMONIZE_REST', 'False') != 'True'
 
-        # Redirect all output to the logging daemon.
-        devnull = open(os.devnull, "w")
-        if isinstance(self.logfile, list):
-            subproc = Popen(self.logfile, stdin=PIPE, stdout=devnull, stderr=devnull,
-                            bufsize=0, close_fds=True, shell=False)
-            logger = subproc.stdin
-        elif isinstance(self.logfile, str):
-            logger = open(self.logfile, "a+", 0)
-        else:
-            raise TypeError("'logfile' must be a string or array")
-        os.dup2(logger.fileno(), sys.stdout.fileno())
-        os.dup2(logger.fileno(), sys.stderr.fileno())
-        os.dup2(devnull.fileno(), sys.stdin.fileno())
-        logger.close()
-        devnull.close()
-
-        # First fork. Discard the parent.
-        pid = os.fork()
-        if pid > 0:
-            os._exit(0)
-
-        # Establish as a daemon, set process group / session id.
         os.chdir(self.statedir)
-        os.setsid()
 
-        # Second fork. The child does the work, discard the second parent.
-        pid = os.fork()
-        if pid > 0:
-            os._exit(0)
+        if daemonize:
+            # Redirect all output to the logging daemon.
+            devnull = open(os.devnull, "w")
+            if isinstance(self.logfile, list):
+                subproc = Popen(self.logfile, stdin=PIPE, stdout=devnull, stderr=devnull,
+                                bufsize=0, close_fds=True, shell=False)
+                logger = subproc.stdin
+            elif isinstance(self.logfile, str):
+                logger = open(self.logfile, "a+", 0)
+            else:
+                raise TypeError("'logfile' must be a string or array")
+            os.dup2(logger.fileno(), sys.stdout.fileno())
+            os.dup2(logger.fileno(), sys.stderr.fileno())
+            os.dup2(devnull.fileno(), sys.stdin.fileno())
+            logger.close()
+            devnull.close()
 
-        # Save process group id to pid file, then run real worker.
-        with open(self.pidfile, "w") as pidObj:
-            pidObj.write("%d\n" % os.getpgid(0))
+            # First fork. Discard the parent.
+            pid = os.fork()
+            if pid > 0:
+                os._exit(0)
 
+            # Establish as a daemon, set process group / session id.
+            os.setsid()
+
+            # Second fork. The child does the work, discard the second parent.
+            pid = os.fork()
+            if pid > 0:
+                os._exit(0)
+
+            # Save process group id to pid file, then run real worker.
+            with open(self.pidfile, "w") as pidObj:
+                pidObj.write("%d\n" % os.getpgid(0))
+
+        # following code is executed both in daemon and not daemon mode
         error = False
         try:
             self.run()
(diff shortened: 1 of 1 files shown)


belforte:
@amaltaro those pylint complains come from part of the code which I did not touch. Odd that they did not came up in earlier changes. Fixes to some is all in all taste and style dependent  W0702( No exception type(s) specified ) and W0212 may be intentional, but I do not know to be able to tell and this maybe a good case for `it works, dont' fix it`

amaltaro:
@belforte Stefano, thanks for providing this fix. Unfortunately we cannot yet enforce any style/pylint recommendations, because as you pointed out, they are provided for the whole module, not only for those lines that you touched. I will request a couple of aesthetic fixes in the code though.

src/python/WMCore/REST/Main.py
```diff
@@ -387,41 +387,48 @@ def kill_daemon(self, silent=False):
 
     def start_daemon(self):
         """Start the deamon."""
+        # REST can be run in the frontend i.e. interactively e.g. to use pdb
+        # by setting the env. var.: DONT_DAEMONIZE_REST=True
+        # if the variable is missing or se to any other value, start noramlly as a daemon
```
amaltaro, line 392: can you please fix a few typos in this line?
And in the line below, add a missing space that has been likely reported by jenkins, thus something like:
```
daemonize = os.getenv('DONT_DAEMONIZE_REST', 'False') != 'True'
```

src/python/WMCore/REST/Main.py
```diff
@@ -449,34 +456,36 @@ def run(self):
...
-                if os.path.exists(pidfile):
-                    with open(pidfile) as fd:
-                        pid = int(fd.readline())
-                    os.remove(pidfile)
-                    cherrypy.log("WATCHDOG: killing slave server %d" % pid)
-                    try:
-                        os.kill(pid, 9)
-                    except:
-                        pass
+        daemonize = os.getenv('DONT_DAEMONIZE_REST','False') != 'True'
```
amaltaro, line 459: same here, can you please add the missing white space before the `'False'`

belforte:
thanks for spotting those Alan. Should be OK now.

src/python/WMCore/REST/Main.py
belforte: thanks for spotting those Alan. Should be OK now.

amaltaro:
@belforte Stefano, I had a look at this module and it looks like we could accomplish the same result, but instead of using an environment variable, we could define a new attribute in the service configuration object. The optional service configuration would look like:
```
main.foreground = False
```

and the code change would go under the `__init__` method of the class RESTMain (around L170), something like:
```
        self.foreground = getattr(config.main, "foreground", False)
```

Sorry for proposing a different solution, but this seem to be a more solid approach IMO. Of course, most of your changes will still be needed here. What do you think?

belforte:
it is a matter of what's convenient. For debugging CRABserver I have no service configuration to play wtih, at least none which I am comfortable or used to. Also I found that working inside a K8s pod it is better to avoid "files" since
pod can be restarted, ad you loose all changed. But surely other people in other contexts have different needs.

Do you refer to this configuration file, by the way ?
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/prod/crabserver/config.py
it is untouched since >5 years and I have no desire to put my hands in it !

I think I see your point, but am not enthusiastic about editing that file for testing.
I have now wrapped the env. var. implementation in a /data/start.sh which works well in our pod [1]

It could be changed of course, but  a script which edit a configuration is more complex (and fragile)
than a script that sets an env. var.  

But the goal here is to make it easy and quick to debug. What are you and others used to do ?
E.g. I see nothing wrong with having multiple ways to toggle the mode according to the way various people
are used to work,   Debugging RESTs is a pain anyhow, my fix e.g. does not remove the multiple
threads, if you get multiple calls while in `pdb` the screen output becomes a mess.

I was mostly worried about touching code the least possible amount

[1]
```
[root@crabserver-6f4865849f-mzp65 data]# /data/start.sh
Please set how you want to start crabserver (add -c or -g option).

Usage example: ./start.sh -c | -g [-d]
	-c start current crabserver instance
	-g start crabserver instance from GitHub repo
	-d start crabserver in debug mode. Option can be combined with -c or -g
[root@crabserver-6f4865849f-mzp65 data]# 
```

belforte:
P.S. this is not a mode in which you would want to start a service and let it run. It is only useful when you are logged inside the pod and want to trace execution or try hacks. Easyness of switching it on/off is important.
Anyhow, I will adapt to whatever you decide to do.

amaltaro:
>... prod/crabserver/config.py

Yes Stefano, that's the service configuration that I was talking about. Fear not, we have been changing it in WMCore almost in a monthly basis :-D In the end, it's basically an object with many attributes that is passed around the REST layer to provide further key/value arguments.

> But the goal here is to make it easy and quick to debug. What are you and others used to do ?

the few times I had to debug it, I did it the hard way with debugging statements added to the source code. Honestly, I need to educate myself on using `pdb`, but I see Todor has been using it quite often.

Anyhow, if you have the whole setup already working, I'm not going to ask you to adapt this code to something that I think to be slightly more robust. I'm mostly concerned of someone forgetting an env variable defined somewhere, and start the REST server in a different and unexpected mode. Which is hard to spot from an environment variable, very simple from the service configuration angle though.

amaltaro:
@belforte can you please squash these 2 commits into a single one?

belforte:
>  I'm mostly concerned of someone forgetting an env variable defined somewhere, and start the REST server in a different and unexpected mode
that's where K8s helps. OTOH when running in foreground it is hard to exit the session w/o stopping the process ! I mostly use this when: 1. desperate and 2. in a tight start/stop/edit/start/look/hack/stop/start loop.
But now I'll need to put my hands in CRAB REST to remove crabcache and I just can't work w/o `pdb`!!
Thanks.  Squash coming.

belforte:
squashed !

amaltaro (approved):
Thanks, Stefano.
Do you need it backported to a different branch?

PR #10350, opened 2021-03-11
allow to run REST/Main in frontend via env.var. fix #10266 - crab branch

### LINKED ISSUE #10266: make it easy to run WMCore REST in debug mode

amaltaro:
1.4.6_crab branch version of https://github.com/dmwm/WMCore/pull/10339

FYI @belforte

### DIFF
--- a/src/python/WMCore/REST/Main.py
+++ b/src/python/WMCore/REST/Main.py
@@ -387,41 +387,48 @@ class RESTDaemon(RESTMain):
 
     def start_daemon(self):
         """Start the deamon."""
+        # This REST server can be run in the frontend i.e. interactively (e.g. to use pdb)
+        # by setting an env. var. via: export DONT_DAEMONIZE_REST=True
+        # if the variable is missing or set to any other value, code starts normally as a daemon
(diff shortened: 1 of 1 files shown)


belforte:
thanks. Look good.

PR #10351, opened 2021-03-11
enforce oracle variable type in bind. Fix #10335 - crab branch

### LINKED ISSUE #10335: wrong binds types passed from REST/Server.py to cx_oracle due to py23 changes
**Impact of the bug**
CRAB (which AFAIK is the only user of WMCore/REST as a REST server on top of Oracle DB)

**Describe the bug**
It is tricky to describe since it is not fully understood yet. 
This is mostly a **heads up** in case @mapellidario has already an idea.

Please take this as a general report aimed at getting your (e.g. @amaltaro @mapellidario ) help in 
defining a good strategy for CRAB migration. I do not like to stick with an old 1.3.crab branch longer,
but the fact that we are the only user of this class, which has no testing in WMCore fwk, makes things tricky
https://developer.oracle.com/dsl/tuininga-cx-oracle.html

For a while @ddaina and myself have been chasing the fact that just some queries at times were taking O(minutes) in preprod,
while all queries were fast on our dev DB instance after we moved to use WMCore with py23 compat (1.4.6.pre2).
We ruled out a change in dependencies (cx_oracle e.g.) and could confirm by simply picking python from
a different GH tag inside a built image.
But faced with :
1. old image is fast on prod, preprod, dev
2. new image is slow on prod, preprod but fast on dev
we could not isolate problem to our code, rather than DB.

Eventually Katarzyna Maria Dziedziniewicz-Wojcik (aka Kate) notices that Oracle optimizer was using full table scale,
in which case the much smaller table size in "dev" DB may explain the difference.

Log inspection showed that queries like
`UPDATE blah WHERE id = :var` with `kwbinds={'var',:'somestring'}`
where instead being executed as
`UPDATE blah WHERE id = :var` with `kwbinds={'var',:u'somestring'}`
the unicode in the `WHERE` is making Oracle fall to full table scan (do you remember the DBS vs. new SQLAlchemy issue 2 in spring 2019 ?).

I stepped through things with pdb, and the time is lost exactly in
https://github.com/dmwm/WMCore/blob/935ef0de5404c0fa10a41b62bede25233baa6c3f/src/python/WMCore/REST/Server.py#L2098
or the equivalent
https://github.com/dmwm/WMCore/blob/935ef0de5404c0fa10a41b62bede25233baa6c3f/src/python/WMCore/REST/Server.py#L2098
But I can't follow code inside the `c.execute` 

Long story short, and much confusion having  flown and still reigning in my head, Kate has now
indicated that we can force the type of the bind variable with code like

```
binds = ()
k='id'
kwbinds = {k:'2b69a73edc024c216dc285fdee8b9044a9c3de784355fa341104eec2'}
sizes = {k: cx_Oracle.STRING}
cur.setinputsizes(**sizes)
#cur.execute(None, id = s)
cur.execute(None, *binds, **kwbinds)
```
I still have to test this, but a more general solution may be needed, since at times
the varaibles used in the `WHERE` are NUMBERS not all VARCHAR.
Kate suggested to look here : https://developer.oracle.com/dsl/tuininga-cx-oracle.html , I still have to digest that.

It is also very much possible that all of this disappears by simply running with python3 !!

**How to reproduce it**
tricky

**Expected behavior**
Oracle should never use full table scan and queries should be completed in msec

belforte:
Fixes #10335

#### Status
ready

#### Description
forces all bind variables of type basestring (string, unicode) to be cx_Oracle.STRING
to deal with unicode introduced to make code py23 comaptible.
It will ONLY be used in 1.4.6_crab branch until we move to python3.
Nothing should be needed when using cx_oracle in python3.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/REST/Server.py
+++ b/src/python/WMCore/REST/Server.py
@@ -8,6 +8,7 @@ import re
 import signal
 import string
 import time
+import cx_Oracle
 from collections import namedtuple
 from functools import wraps
 from threading import Thread, Condition
@@ -2095,7 +2096,25 @@ class DatabaseRESTApi(RESTApi):
         trace and cherrypy.log("%s execute: %s %s" % (trace, binds, kwbinds))
         if request.db['type'].__name__ == 'MySQLdb':
             return c, c.execute(sql, kwbinds)
-        return c, c.execute(None, *binds, **kwbinds)
+        sizes={}
+        # binds is a list of dictionaries, but somehow passing it by pointer wraps it as a 1-value tupla
+        if binds:
+            for bDict in binds[0]:
+                for k in bDict.keys():
+                    if isinstance(bDict[k], basestring):
+                        sizes[k] = cx_Oracle.STRING
+        # kwbinds is a dictionary (or lists if needed)
+        if kwbinds:
+            for k in kwbinds.keys():
+                if isinstance(kwbinds[k], basestring):
+                    sizes[k]=cx_Oracle.STRING
+        if sizes:
+            c.setinputsizes(**sizes)
+        t1 = time.time()
+        result=c.execute(None, *binds, **kwbinds)
+        elapsed = time.time() -t1
+        trace and cherrypy.log("%s ==== COMPLETED in %s sec ====" % (trace, elapsed))
+        return c, result
 
     def executemany(self, sql, *binds, **kwbinds):
         """Execute a SQL statement many times with bind variables.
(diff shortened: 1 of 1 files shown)


belforte:
After this and https://github.com/dmwm/WMCore/pull/10350 are merged, please create a tag in 1.4.6_crab branch

amaltaro:
@belforte I understand it has already been tested and it doesn't need any further review, right?

belforte:
right. yes it has been tested. Reviews are always welcome, but I would not know how to test it more. We ran tens of thousands of queries with actual CRAB tasks and timing is better than before. In any case it will not contaminate WMCore master :-) (even if CRAB is the only user of the Class which I changed)

amaltaro:
I would have mostly aesthetic suggestions and a few questions to better understand this code (e.g., if you have the same parameter/column name in both positional and keyword arguments, you would cast them to the same data type, even if you wanted them to be different).

Anyhow, if further changes are needed for this branch, we can just do it.

amaltaro:
@belforte @ddaina 1.4.6.crab1 tagged. Have fun!

belforte:
>  (e.g., if you have the same parameter/column name in both positional and keyword arguments, you would cast them to the same data type, even if you wanted them to be different

yeah, that's one of the reasons why I have little confidence in this change and want to keep it quarantined. In CRAB case it is OK because it always use keyword arguments, but surely it does not have the generality of the original. I do not fully understand the original, and I do not have simple ways to test use cases other than my own. So I hope in a couple of months we can throw it away.

THANKS !

PR #10355, opened 2021-03-16
Add py3-cmsmonitoring to the list of python3 requirements

### LINKED ISSUE #10305: Build cmsmonitoring package on python3
**Impact of the new feature**
WMCore (and others) services depending on cmsmonitoring tools

Yes, we need to be able to pull the correct cmsmonitoring (RPM) according to the python stack we are going to use.

I think the best option we have for now would be to split the current cmsmonitoring spec:
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/cmsmonitoring.spec

in two specs, one for python2 and one for python3 (according to the convention used in the cmsdist repo).

So, it means we have to create a brand new `cmsmonitoring` spec for the python3 stack as well.

Note that, if we rename the cmsmonitoring spec to py2-cmsmonitoring, we need to update all the other cmsdist specs that depend on that package.

Not rename the `cmsmonitoring.spec`, and simply create a new `py3-cmsmonitoring` with the correct dependencies.

**Additional context**
Issue spotted and mentioned here: #10302 (comment)

amaltaro:
Fixes #10305

#### Status
ready

#### Description
I just noticed our list of python3 packages didn't have the py3-cmsmonitoring dependency, so here it goes.

yes

#### Related PRs
Complement to: https://github.com/cms-sw/cmsdist/pull/6714

none

### DIFF
--- a/requirements_py3.txt
+++ b/requirements_py3.txt
@@ -29,3 +29,4 @@ pylint==2.6.0
 pymongo==3.10.1
 retry==0.9.2
 sphinx==1.6.3
+CMSMonitoring==0.3.4


amaltaro:
Now with the correct pypi name. Good to go.

PR #10362, opened 2021-03-22
[py2py3] Migration at level scr/python/A/B/C - slice 22

mapellidario:
Fixes #10146 

#### Status
In development 

#### Related PRs

* Previous migration PR: #10361  (slice21, issue #10145)
* Next migration PR: #10363 (slice23, issue #10147  )

#### Description

Run futurize and some manual changes on the first batch of src/python/A/B/C.

"Native String" approach:
- `src/python/WMCore/WMSpec/StdSpecs/ReReco.py`: here we use `str()` to format an object to a string and then replace it into a string, `realArg = argument.replace("#N", str(skimIndex))`. This should be mostly safe!
- `src/python/WMCore/WMSpec/StdSpecs/StdBase.py`:  Some problems may arise here. `str()` use to format the keys of a dictionary (tracked in #10377).

    ```python
        newSplitArgs = {}
        for argName in splitArgs:
            newSplitArgs[str(argName)] = splitArgs[argName]
        procTask.setSplittingAlgorithm(splitAlgo, **newSplitArgs)
    ```
    and it is also used to format a string that is then looped through:

   ```python
    stringRunNumber = str(self.runNumber).zfill(9)
    runSections = [stringRunNumber[i:i + 3] for i in range(0, 9, 3)]
    ```

- `src/python/WMCore/WMSpec/StdSpecs/TaskChain.py`:  Some problems may arise here. `str()` use to format the keys of a dictionary (tracked in #10377).

    ```python
        procMergeTasks = {}
        for outputModuleName in modulesToMerge:
            ...
            procMergeTasks[str(outputModuleName)] = mergeTask
    ```

`pylint --py3k` report:
- warn us about the use of `round`. we have `multiplier = int(round(ePerJob / ePerLumi))`, so we should be safe here. I consider this warning as a false positive

#### Is it backward compatible (if not, which system it affects?)

It should be. (Any possible cause for errors will we reported here)

#### External dependencies / deployment changes

Requires python-future in both py2 and py3 environments.

Issue #10148, opened 2020-12-03
[py2py3] Migration at level scr/python/A/B/C - slice 24


mapellidario:
#### Intro

Our purpose is to continue with the gradual migration. 
This is the 24-th slice in which the migration has been divided into. 
We have split migration units at level `src/python/A/B/C`.

A detailed plan of the migration at this depth is available 
at the dmwm mkdocs [detailed plan page](https://cms-dmwm-test.docs.cern.ch/py2py3/plan_detailed/), 
with summary information on top.

#### Impact of the new feature

We start migrating the 24-th bulk of migration units 
at src/python/A/B/C level.
```plaintext
                                     name  nfiles  gradual  future_pm
WMComponent_TaskArchiver_CleanCouchPoller       1     True         44
        WMComponent_JobCreator_JobCreator       1     True          0
    WMComponent_TaskArchiver_TaskArchiver       1     True          0
                     WMCore_BossAir_MySQL      19     True          8
                      WMCore_ReqMgr_Utils       1     True          9
                        WMCore_WMStats_T0       2     True          1
         WMQuality_Emulators_ReqMgrClient       1     True          0
                    WMCore_ReqMgr_Service       6     True         23
```

#### Is your feature request related to a problem? Please describe.

The end goal is to have WMCore py2 and py3 compatible.

#### Describe the solution you'd like

Run futurize on the first bulk of units 
and then apply the necessary manual changes.

PR #10363, opened 2021-03-22
[py2py3] Migration at level scr/python/A/B/C - slice 23

### LINKED ISSUE #10147: [py2py3] Migration at level scr/python/A/B/C - slice 23
#### Intro

This is the 23rd slice in which the migration has been divided into. 

We start migrating the 23rd bulk of migration units 
at src/python/A/B/C level.

```plaintext
                    WMCore_BossAir_Plugins       5     True         28
WMComponent_AnalyticsDataCollector_Plugins       2     True          5
          WMComponent_RetryManager_PlugIns       7     True          2
                      WMCore_WMSpec_Makers       5     True         18
              WMQuality_Emulators_WMAgents       3     True          0
                 WMCore_WMRuntime_Monitors       4     True          5
       WMQuality_Emulators_WMSpecGenerator       6     True         13
            WMCore_JobSplitting_Generators      12     True         16
                    WMCore_WMStats_Service       4     True          2
   WMComponent_JobCreator_JobCreatorPoller       1     True          6
```

mapellidario:
Fixes #10147 

#### Status
In development 

#### Related PRs

* Previous migration PR: #10362  (slice22, issue #10146 )
* Next migration PR: #10371 (slice24, issue #10148 )

#### Description

"Native String" approach (use `str` in both py2 and py3, hoping that it stays consistent within the same runtime):
- `src/python/WMCore/BossAir/Plugins/MockPlugin.py`: using `jobid = str(jj['id'])` to append a str to `tmpname = report.cmsRun1.output.output.files.file0.lfn.split('.root')[0]`. Not a problem if it is kept as a `str` in both py2 and py3
- `src/python/WMCore/BossAir/Plugins/SimpleCondorPlugin.py`: 
  - format input of `classad.quote(input)`, which should accept a `str` input in py3 as well [as they claim](https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/classad.html#classad.quote), but better be safe than sorry and check this.
  - format a ton of keys of a dict that is appended to `JobParameters`, such as `ad['My.OriginalCpus'] = str(origCores)`. If this is used only by us in `SimpleCondorPlugin.createSubmitRequest` and not by WMCore clients, we may be in a good spot.
  - tracked in #10378

### DIFF
--- a/src/python/WMComponent/AnalyticsDataCollector/Plugins/PluginInterface.py
+++ b/src/python/WMComponent/AnalyticsDataCollector/Plugins/PluginInterface.py
@@ -6,6 +6,7 @@ Base class for AnalyticsDataCollector plug-ins
 
 """
 
+from builtins import object
 class PluginInterface(object):
     """Interface for policies"""
     def __init__(self, **args):
--- a/src/python/WMComponent/AnalyticsDataCollector/Plugins/Tier0Plugin.py
+++ b/src/python/WMComponent/AnalyticsDataCollector/Plugins/Tier0Plugin.py
@@ -10,6 +10,8 @@ Created on Nov 2, 2012
 @author: dballest
 """
 
+from builtins import filter
+
 import re
 import threading
 import traceback
@@ -225,7 +227,7 @@ class Tier0Plugin(PluginInterface):
         clean unused task lists
         """
         self.logger.debug('Cleaning up task cache')
-        for workflow in self.taskCache.keys():
+        for workflow in list(self.taskCache):
             if workflow not in reportedWorkflows:
                 self.taskCache.pop(workflow)
         return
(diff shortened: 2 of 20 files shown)


amaltaro (approved):
It looks good to me, Dario.

Regarding `SimpleCondorPlugin`, I was going to say it's only used by us in WMAgent, but T0 agent is also using it. However, T0 agent uses almost all the same components as we do, so I believe data will be propagated all the way to SimpleCondorPlugin in the same way it gets propagated in WMAgent. We shall see ...
 
BTW, thanks for those notes in the description. They will be very handy if we have issues and need to track it down in these slices!

mapellidario:
Tracking the investigations about `SimpleCondorPlugin` in #10378

PR #10371, opened 2021-03-23
[py2py3] Migration at level scr/python/A/B/C - slice 24 

### LINKED ISSUE #10148: [py2py3] Migration at level scr/python/A/B/C - slice 24

mapellidario:
Fixes #10148 

#### Status
ready

#### Related PRs

* Previous migration PR: #10363  (slice23, issue #10147)
* Next migration PR: #10379  (slice25, issue #10149  )

#### Description

"Native string" approach:
- `src/python/WMComponent/TaskArchiver/CleanCouchPoller.py`: 
  - `str()` is used to format dict strings, such as `count = str(row['key'][1]) ; workflowData['retryData'][taskName][count] = row['value']` or `stepFailures[exitCode]['runs'][str(run.run)].append(l)`
  - this should not be a big problem, but we'd better check this: tracked in #10377
- `src/python/WMCore/ReqMgr/Service/RequestAdditionalInfo.py`: 
  - `str()` used in 
      ```python
      class WorkloadConfig(RESTEntity):
      ...
          @restcall(formats=[('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
          @tools.expires(secs=-1)
          def get(self, name):
              helper = WMWorkloadHelper()
              ...
              return str(helper.data)
      ```
      Which is directly used in `src/python/WMCore/ReqMgr/Service/RestApiHub.py`, which is not clear to me how it  works. It is possible that we need to return bytes at all costs here, we will have to check this.
  - tracked in #10376

Division:
- `src/python/WMComponent/TaskArchiver/CleanCouchPoller.py`: using the new division and it should be safe. 
    ```python
        summarySize = len(json.dumps(workflowData)) / 1024
        if summarySize > 6 * 1024:  # 6MB
            msg = "Workload summary for %s is too big: %d Kb. " % (workflowName, summarySize)
        ...
        logging.info("About to commit %d Kb of data for workflow summary for %s", summarySize, workflowName)

    ```
 
`pylint --py3k`: it complains about using `myexception.message` instead of `str(myexception)`. This is not a problem in this case, because `.message` is used only if `hasattr(myexception, "message")`. (I would rather refactor such behavior only if we spot an issue. Since this is not the case at the moment, I would not touch it)

### DIFF
--- a/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
+++ b/src/python/WMComponent/TaskArchiver/CleanCouchPoller.py
@@ -1,6 +1,10 @@
 """
 Perform cleanup actions
 """
+from __future__ import division
+from builtins import range
+from future.utils import viewitems
+
 from future import standard_library
 standard_library.install_aliases()
 
@@ -361,7 +365,7 @@ class CleanCouchPoller(BaseWorkerThread):
 
             workflowDict = self.centralRequestDBReader.getStatusAndTypeByRequest(requestNames)
 
-            for request, value in workflowDict.items():
+            for request, value in viewitems(workflowDict):
                 if value[0].endswith("-archived"):
                     self.cleanAllLocalCouchDB(request)
                     numDeletedRequests += 1
@@ -422,7 +426,7 @@ class CleanCouchPoller(BaseWorkerThread):
                 # Get the task-workflow ids, sort them by ID,
                 # higher ID first so we kill
                 # the leaves of the tree first, root last
-                workflowsIDs = workflows[workflow]["workflows"].keys()
+                workflowsIDs = list(workflows[workflow]["workflows"])
                 workflowsIDs.sort(reverse=True)
 
                 # Now go through all tasks and load the WMBS workflow objects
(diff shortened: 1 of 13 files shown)


amaltaro (approved):
It looks good to me, thanks Dario.

It contains some important changes to our ReqMgr2 REST APIs; in addition to a few BossAir DAOs. 

I'm creating a new GH issue such that we can create a simple script to help us validating all our RESTful APIs, which shall be extremely helpful in the upcoming central services validation.

PR #10394, opened 2021-03-31
Set output datasets to VALID in DBS3 before announcing a standard workflow

### LINKED ISSUE #9837: MSOutput: set output datasets to VALID in DBS3 before announcing a standard workflow
**Impact of the new feature**
ReqMgr2MS (MSOutput)

This has been discussed with Todor and Sharad, and it looks like we could accommodate a few more actions in the MSOutput, such that these functionalities can be deprecated in Unified.
 
In addition to the output data placement performed by MSOutput (for wfs in closed-out and announced), we should also mark all those datasets as `VALID` in DBS3.
The DBS client API is:
```
dbsapi = DbsApi(url=url3)
dbsapi.updateDatasetType(dataset=DATASET_NAME, dataset_access_type='VALID')
```

but if we decide to run requests concurrently, then we could use the equivalent REST API (ask Yuyi for details).

The best order of actions, in my opinion, would be:
1. mark output datasets as VALID
2. if previous step was successful, then perform the output data placement

NOTE: RelVal workflows are not announced in an standalone mode, but in batches. In order to ease this transition, we should only mark output datasets as VALID for standard workflows; Unified will keep taking care of RelVals for a little longer.

Keep workflow announcement in Unified

**Additional context**
Here we have a more complete description of the workflow announcement process:
https://github.com/dmwm/WMCore/issues/8921
but some of those have already been implemented, and others might be slightly modified.

haozturk:
Fixes #9837 

#### Status
DBS API calls are tested locally in a WMAgent env. (1.4.3.patch2), however the rest of the changes is not tested

#### Description
This PR includes the migration of one feature from Unified to WMCore: Setting output datasets to VALID in DBS3 before announcing a standard workflow.

This feature is implemented as a part of the MSOutput pipeline. In this PR `setDBSStatus` is the first step of the pipeline (before making actual output data placement) and also note that this functionality works for both RelVal and non-RelVal workflows.

Check https://github.com/dmwm/deployment/pull/1044  for the configuration of this feature. Note that `enableDbsStatusChange` is False, right now. Because, we need to coordinate for disabling the feature in Unified and enabling in WMCore. Soon, I'll come up a PR to disable it in Unified.

yes

#### Related PRs
Not for now - soon a related PR in https://github.com/CMSCompOps/WmAgentScripts will come.

DBS

@amaltaro @todor-ivanov As I mentioned, the feature is not tested properly, it would be great if you can help me testing it.

### DIFF
--- a/src/python/WMCore/MicroService/MSOutput/MSOutput.py
+++ b/src/python/WMCore/MicroService/MSOutput/MSOutput.py
@@ -109,6 +109,9 @@ class MSOutput(MSCore):
         for endpoint, quota in viewitems(self.msConfig['tapePledges']):
             self.tapeStatus[endpoint] = dict(quota=quota, usage=0, remaining=0)
 
+        self.dbs3Writer = DBS3Writer(url=self.msConfig["dbsReadUrl"],
+                                     writeUrl=self.msConfig["dbsWriteUrl"])
+
         msOutIndex = IndexModel('RequestName', unique=True)
         msOutDBConfig = {
             'database': 'msOutDB',
@@ -260,36 +263,27 @@ class MSOutput(MSCore):
             msg += "It needs to be of type: MSOutputTemplate"
             raise UnsupportedError(msg)
 
-        dbs3Writer = DBS3Writer(url=self.msConfig["dbsReadUrl"],
-                                writeUrl=self.msConfig["dbsWriteUrl"])
-
-        # if anything fail along the way, set it back to "pending"
-        dbsUpdateStatus = "done"
+        # if anything fail along the way, set it back to False
+        dbsUpdateStatus = True
         for dMap in workflow['OutputMap']:
 
-            if self.msConfig['enableDbsStatusChange']:
+            res = self.dbs3Writer.setDBSStatus(dataset=dMap["Dataset"],
+                                               status=self.msConfig['dbsStatus']["valid"])
 
-                res = dbs3Writer.setDBSStatus(dataset=dMap["Dataset"],
-                                              status=self.msConfig['dbsStatus']["valid"])
-
-                if res:
-                    dMap["DBSStatus"] = self.msConfig['dbsStatus']["valid"]
-                else:
-                    # There is at least one dataset whose dbs status update is unsuccessful
-                    dbsUpdateStatus = "pending"
-
+            if res:
+                dMap["DBSStatus"] = self.msConfig['dbsStatus']["valid"]
             else:
-                msg = "DRY-RUN DBS: DBS status change for DID: {}, ".format(dMap['Dataset'])
-                self.logger.info(msg)
+                # There is at least one dataset whose dbs status update is unsuccessful
+                dbsUpdateStatus = False
 
         # Finally, update the MSOutput template document with either partial or
         # complete dbs statuses
         self.docKeyUpdate(workflow, OutputMap=workflow['OutputMap'])
         workflow.updateTime()
-        if dbsUpdateStatus == "done":
+        if dbsUpdateStatus:
             self.logger.info("All the DBS status updates succeeded for: %s. Marking it as 'done'",
                              workflow['RequestName'])
-            self.docKeyUpdate(workflow, DBSUpdateStatus='done')
+            self.docKeyUpdate(workflow, DBSUpdateStatus=True)
         else:
             self.logger.info("DBS status updates partially successful for: %s. Keeping it 'pending'",
                              workflow['RequestName'])
@@ -521,7 +515,7 @@ class MSOutput(MSCore):
                                                  Functor(self.docCleaner)])
 
         wfCounterTotal = 0
-        mQueryDict = { "$or": [ { "TransferStatus": "pending" }, { "DBSUpdateStatus": "pending" } ] }
+        mQueryDict = { "$or": [ { "TransferStatus": "pending" }, { "DBSUpdateStatus": False } ] }
         pipeCollections = [(msPipelineRelVal, self.msOutRelValColl),
                            (msPipelineNonRelVal, self.msOutNonRelValColl)]
         for pipeColl in pipeCollections:
--- a/src/python/WMCore/MicroService/MSOutput/MSOutput.py
+++ b/src/python/WMCore/MicroService/MSOutput/MSOutput.py
@@ -25,9 +25,7 @@ from WMCore.Database.MongoDB import MongoDB
 from WMCore.MicroService.MSOutput.MSOutputTemplate import MSOutputTemplate
 from WMCore.WMException import WMException
 from WMCore.Services.AlertManager.AlertManagerAPI import AlertManagerAPI
-
-#DBS modules
-from dbs.apis.dbsClient import DbsApi
+from WMCore.Services.DBS.DBS3Writer import DBS3Writer
 
 
 class MSOutputException(WMException):
@@ -251,35 +249,6 @@ class MSOutput(MSCore):
             self.logger.exception(msg)
             self.updateReportDict(summary, "error", msg)
 
-    def getDBSStatus(self, dataset):
-        """
-        The function to get the DBS status of outputs
-        :param dataset: dataset name
-        :return: DBS status of the given dataset
-        """
-
-        dbsReadUrl = self.msConfig["dbsReadUrl"]
-        dbsApi = DbsApi(url=dbsReadUrl)
-
-        response = None
-        try:
-            response = dbsApi.listDatasets(dataset=dataset, dataset_access_type='*', detail=True)
-        except Exception as ex:
-            msg = "Exception while getting the status of following dataset on DBS: {} ".format(dataset)
-            msg += "Error: {}".format(str(ex))
-            self.logger.exception(msg)
-
-        if response:
-            dbsStatus = response[0]['dataset_access_type']
-            isAllowedStatus = dbsStatus in self.msConfig["allowedDbsStatuses"]
-
-            if isAllowedStatus:
-                return dbsStatus
-            else:
-                raise Exception("This is not an allowed DBS status: {}".format(str(dbsStatus)))
-        else:
-            return None
-
     def setDBSStatus(self, workflow):
         """
         The function to set the DBS status of outputs as VALID
@@ -292,7 +261,7 @@ class MSOutput(MSCore):
             raise UnsupportedError(msg)
 
         dbsWriteUrl = self.msConfig["dbsWriteUrl"]
-        dbsApi = DbsApi(url=dbsWriteUrl)
+        dbs3Writer = DBS3Writer(dbsWriteUrl)
 
         # if anything fail along the way, set it back to "pending"
         dbsUpdateStatus = "done"
@@ -300,18 +269,11 @@ class MSOutput(MSCore):
 
             if self.msConfig['enableDbsStatusChange']:
 
-                try:
-                    dbsApi.updateDatasetType(dataset=dMap["Dataset"],
-                                             dataset_access_type=self.msConfig['dbsStatus']["valid"])
-                except Exception as ex:
-                    msg = "Exception while setting the status of following dataset on DBS: {} ".format(dMap["Dataset"])
-                    msg += "Error: {}".format(str(ex))
-                    self.logger.exception(msg)
+                res = dbs3Writer.setDBSStatus(dataset=dMap["Dataset"],
+                                              status=self.msConfig['dbsStatus']["valid"])
 
-                dbsStatus = self.getDBSStatus(dMap["Dataset"])
-
-                if dbsStatus == self.msConfig['dbsStatus']["valid"]:
-                    dMap["DBSStatus"] = dbsStatus
+                if res:
+                    dMap["DBSStatus"] = self.msConfig['dbsStatus']["valid"]
                 else:
                     # There is at least one dataset whose dbs status update is unsuccessful
                     dbsUpdateStatus = "pending"
--- a/src/python/WMCore/MicroService/MSOutput/MSOutput.py
+++ b/src/python/WMCore/MicroService/MSOutput/MSOutput.py
@@ -26,6 +26,9 @@ from WMCore.MicroService.MSOutput.MSOutputTemplate import MSOutputTemplate
 from WMCore.WMException import WMException
 from WMCore.Services.AlertManager.AlertManagerAPI import AlertManagerAPI
 
+#DBS modules
+from dbs.apis.dbsClient import DbsApi
+
 
 class MSOutputException(WMException):
     """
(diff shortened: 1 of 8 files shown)


amaltaro:
@haozturk thanks for providing this new feature. 

I have one general comment/request though, could you please move the getter/setter methods to our DBS3 wrapper module:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L843

and use our wrapper instead of making direct calls with the `DbsApi` object? I'm sorry that it wasn't made explicit in the GH issue.

haozturk:
@amaltaro should I move the setter function to https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBSWriterObjects.py as it looks like that DBS3Reader module is intended to include only readonly functions?

amaltaro:
That DBSWriterObjects.py module has just a set of functions, not really a class implemented.

Given that DBS3Reader.py already has a couple of write functions, it would be fine to add it there as well. However, if we really want to to the right thing, then I'd suggest creating a DBS3Writer.py function that inherits from DBS3Reader, something like this:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WMStats/WMStatsWriter.py#L61

haozturk:
Okay, I create a `DBS3Writer` class by inheriting from `DBS3Reader`. Thanks Alan

haozturk:
@amaltaro I've moved the DBS API calls to their dedicated classes with 2 new unit tests. Looks like there are problems with two unit tests:
```
WMCore_t.Services_t.DBS_t.DBSWriter_t.DBSWriterTest:testSetDBSStatus
WMCore_t.Services_t.LogDB_t.LogDB_t.LogDBTest:test_heartbeat
```
I pass all unit tests on my testing environment, can you check what's wrong with them, please?

Regarding the changes within MSOutput, as I mentioned earlier, I could not test them due to lack of unit test setup for MSOutput. Those changes requires testing in some way.

amaltaro:
test this please

amaltaro:
@haozturk thanks for providing these changes Hasan. I'm going to review this PR soon, but just wanted to point you to the unit test failure logs:
https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/11613/testReport/junit/WMCore_t.Services_t.DBS_t.DBSWriter_t/DBSWriterTest/testSetDBSStatus/

as you can see, the jenkins bot certificate has no privileges to write to DBS, which makes sense. So we should probably try to mock that write call as well.

haozturk:
Ah, thanks for pointing out that Alan, that makes sense. Let me revise that unit test.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
```diff
@@ -248,6 +249,53 @@ def _executeConsumer(self, summary):
...
+        The function to set the DBS status of outputs as VALID
+        :param workflow: a MSOutputTemplate object workflow
+        :return: the MSOutputTemplate object itself (with the necessary updates in place)
+        """
+        if not isinstance(workflow, MSOutputTemplate):
+            msg = "Unsupported type object '{}' for workflows! ".format(type(workflow))
+            msg += "It needs to be of type: MSOutputTemplate"
+            raise UnsupportedError(msg)
+
+        dbs3Writer = DBS3Writer(url=self.msConfig["dbsReadUrl"],
```
amaltaro, line 263: I would suggest to initialize it in the `__init__` method instead.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
```diff
@@ -248,6 +249,53 @@ def _executeConsumer(self, summary):
...
+                                writeUrl=self.msConfig["dbsWriteUrl"])
+
+        # if anything fail along the way, set it back to "pending"
+        dbsUpdateStatus = "done"
+        for dMap in workflow['OutputMap']:
+
+            if self.msConfig['enableDbsStatusChange']:
```
amaltaro, line 270: I guess I would remove this option (and the dry-run mode). When we have it implemented, I believe there is no reason to disable this feature in MSOutput, right?

src/python/WMCore/MicroService/MSOutput/MSOutputTemplate.py
```diff
@@ -102,6 +102,7 @@ def docSchema(self):
                            'Copies': 1,
                            ...}],
                     "TransferStatus": "pending"|"done,
+                    "DBSUpdateStatus": "pending"|"done,
```
amaltaro, line 105: I guess I would make it a boolean, thus something like:
```
"DBSStatusUpdated": False|True,
```
but then it would require more changes from your side. So I leave it up to you.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -971,3 +971,32 @@ def getParentDatasetTrio(self, childDataset):
...
         return parentFrozenData
+
+    def getDBSStatus(self, dataset):
+        """
+        The function to get the DBS status of outputs
+        :param dataset: dataset name
+        :return: DBS status of the given dataset
+        """
+
+        allowedDbsStatuses = ["VALID", "INVALID", "PRODUCTION"]
```
amaltaro, line 982: a dataset can have other statuses. If I'm not wrong, DELETED and DEPRECATED as well. However, I don't think we need to have a list of valid statuses, we just use whatever is provided by DBS.

src/python/WMCore/Services/DBS/DBS3Writer.py
```diff
@@ -0,0 +1,58 @@
...
+    def setDBSStatus(self, dataset, status):
+        """
+        The function to set the DBS status of an output dataset
+        :param dataset: Dataset name
+        :return: True if operation is successful, False otherwise
+        """
+
+        try:
+            self.dbs.updateDatasetType(dataset=dataset,
+                                     dataset_access_type=status)
```
amaltaro, line 47: indentation needs fixing. I would suggest to remove the blank line at 44 as well.

amaltaro: BTW, we should fetch the output of this call and log the error in case it happens; or return it to the caller module.
By parsing the output of that call, we can also get rid of of the `getDBSStatus` implementation in the lines below.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
```diff
@@ -248,6 +249,53 @@ def _executeConsumer(self, summary):
...
+
+                res = dbs3Writer.setDBSStatus(dataset=dMap["Dataset"],
```
amaltaro, line 272: Note that my comments on the DBS3Writer module will impact this logic.

Another comment, I failed to see where this `self.msConfig['dbsStatus']` gets set in this module. Perhaps I would suggest to define an object variable like:
```
self.dbsOutputStatus = "VALID"
```
and simply refer to it here.

amaltaro (changes requested):
@haozturk Hasan, I made a non-exhaustive review of this PR, and I would suggest you to look at those such that we can have another clear review once those changes are in place. I hope that's fine with you.

Just a note/concern: this PR will break the MSOutputTemplate schema, so we need to be careful and make these changes backward compatible with the documents already in MongoDB.

@todor-ivanov might have other comments/requests/concerns as well.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
haozturk: I put it to make the switch from Unified to WMCore easy. So, once the PR is good to merge, we can merge it and put it to production with `enableDbsStatusChange`  `False` and simply it will not take any action but to print some logs. Once we're ready to switch from Unified to WMCore, we can simply change this parameter to True without touching the source code. Does it make sense?

src/python/WMCore/Services/DBS/DBS3Writer.py
haozturk: I also thought that but the function returns `None` if the call is successful and throws an exception o/w. I would prefer it to return True or False respectively. Since it does not do that, I handled it with `getDBSStatus` calls to make it more intuitive (according to me :) ) . We can also handle it:
- Return True if `updateDatasetType` returns None
- Return False o/w

Which one do you prefer?

src/python/WMCore/MicroService/MSOutput/MSOutput.py
haozturk: I also made it configurable: https://github.com/dmwm/deployment/pull/1044/files#diff-3a773e3fbfb5ff69f5fd806e7a4ab54a7dfaaf1f080e7e96ac1df9e73589c2ffR115 It's also possible to set it within the module. Which one do you prefer?

src/python/WMCore/MicroService/MSOutput/MSOutput.py
amaltaro: I think we could enable this DBS status change in MSOutput and keep both this service and Unified updating the DBS status for a day or two (I'm almost sure it's harmless and nothing fails). Then we have time to stop Unified.

At this stage, dry-run mode is just an over-complication of the code IMO.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
amaltaro: Oh, I missed the deployment changes. Thank for pointing it out.
I think it's fine to keep it in the service configuration file. Nonetheless, I would set a default value to "VALID" in this module, in case the configuration does not have it.

src/python/WMCore/Services/DBS/DBS3Writer.py
amaltaro: Hmm, too bad it returns None in case the call is successful..

>    Return True if updateDatasetType returns None
>    Return False o/w

This is my preference. Given that - if there is an exception - the error would likely be on the server side, I would suggest to change the log level from exception to error, such that we do not pollute the service logs with unnecessary information.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
haozturk: Should be fine now

src/python/WMCore/MicroService/MSOutput/MSOutput.py
haozturk: That makes sense to me. This option is removed in the latest commit.

src/python/WMCore/MicroService/MSOutput/MSOutputTemplate.py
haozturk: Good suggestion, this is applied in the last commit.

src/python/WMCore/Services/DBS/DBS3Reader.py
haozturk: As agreed, this extra check is also removed.

src/python/WMCore/Services/DBS/DBS3Writer.py
haozturk: I applied your suggestions Alan, but `setDBSStatus` still returns True if the API call is successful, False otherwise. This looks quite intuitive to me. Do you agree?

src/python/WMCore/MicroService/MSOutput/MSOutput.py
haozturk: Since deployment has the value, I did not set a default value in this module to avoid duplicates. If you think, it's more suitable to keep the configuration in MSOutput module instead of deployment repository, that's also fine, I can change it if you'd like.

src/python/WMCore/MicroService/MSOutput/MSOutput.py
```diff
@@ -248,6 +252,44 @@ def _executeConsumer(self, summary):
...
+
+        # if anything fail along the way, set it back to False
+        dbsUpdateStatus = True
+        for dMap in workflow['OutputMap']:
+
+            res = self.dbs3Writer.setDBSStatus(dataset=dMap["Dataset"],
+                                               status=self.msConfig['dbsStatus']["valid"])
+
+            if res:
+                dMap["DBSStatus"] = self.msConfig['dbsStatus']["valid"]
```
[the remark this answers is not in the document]
amaltaro: BTW, do you know what happens if you set a VALID dataset to VALID once again? I believe DBS won't complain, right?

In addition to that, should we take into consideration datasets that have not been produced in the workflow, thus not being available in the DBS database?

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -971,3 +971,25 @@ def getParentDatasetTrio(self, childDataset):
...
         return parentFrozenData
+
+    def getDBSStatus(self, dataset):
+        """
+        The function to get the DBS status of outputs
+        :param dataset: dataset name
+        :return: DBS status of the given dataset
+        """
+
+        response = None
```
amaltaro, line 982:
```suggestion
        dbsStatus = None
```

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -971,3 +971,25 @@ def getParentDatasetTrio(self, childDataset):
...
+            response = self.dbs.listDatasets(dataset=dataset, dataset_access_type='*', detail=True)
+        except Exception as ex:
+            msg = "Exception while getting the status of following dataset on DBS: {} ".format(dataset)
+            msg += "Error: {}".format(str(ex))
+            self.logger.exception(msg)
+
+        if response:
+            dbsStatus = response[0]['dataset_access_type']
+            self.logger.info("%s is %s", dataset, dbsStatus)
+            return dbsStatus
```
amaltaro, line 993: and here we could replace this if/else block by something like:
```
if response:
    dbsStatus = response[0]['dataset_access_type']
    self.logger.info("%s is %s", dataset, dbsStatus)
return dbsStatus
```

src/python/WMCore/Services/DBS/DBS3Writer.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
+"""
+_DBSReader_
```
amaltaro, line 3: Wrong copy/pasting? :-D You might just remove it as well.

amaltaro (changes requested):
Hasan, these changes are looking good. However, I have made some comments along the code. I have a suggestion for this DBS status related schema, which I believe to be a simplified version of what is currently implemented.

I wonder why you changed the DBSMockData.json as well? Did we have the wrong parentage information in there? That information is fed directly from DBS, so I would be surprised if we had such issue. In addition to that, that json file has been updated this week, so you might want to rebase your PR to get the latest changes.

Thank you very much for providing these changes, Hasan. And apologies for the delay.

PR #10434, opened 2021-04-05
Remove unused tweak file

### LINKED ISSUE #9878: Clean up SetupCMSSWSetup step from unused applyPSetTweak method and check were tweak from ConfigCache is used at otherwise
**Impact of the new feature**
[CMSSWStepSetup](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L770-L772) tries to apply an inexistent method when psetTweak is `not None` in the configuration.

This `psetTweak` parameter is not set anywhere and is always `None` though, which is why we never get an exception message here at present.
This is the file that CMSSWFetcher fetches from ConfigCache [HERE](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Fetchers/CMSSWFetcher.py#L57) though, meaning we are apparently not using the psetTweak file created by TweakMaker and written in ConfigCache at least in this runtime script. Rather, we the script uses the WMLoad and  WMBS Job objects to create the tweaks that are applied at runtime.

This is related to #9808.

We should make 100% sure this isn't used neither here nor in the T0 and simply get rid of this part. Where is the [pset tweak from ConfigCache](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/Steps/Fetchers/CMSSWFetcher.py#L57) used? Do we still need this? Can we get rid of it?

Alternative is leaving this code as is, which is the current situation

**Additional context**
You can find some notes here:
https://github.com/dmwm/WMCore/wiki/List-of-PSet-tweaks-applied-by-WMAgent-during-job-runtime#notes-on-the-code

khurtado:
Fixes #9878

#### Status
not-tested

#### Description
Removes unused JSON tweak file

YES

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Fetchers/CMSSWFetcher.py
+++ b/src/python/WMCore/WMSpec/Steps/Fetchers/CMSSWFetcher.py
@@ -54,7 +54,3 @@ class CMSSWFetcher(FetcherInterface):
                 configCache = ConfigCache(cacheUrl, cacheDb)
                 configCache.loadByID(configId)
                 configCache.saveConfigToDisk(targetFile = fileTarget)
-                tweak = TweakAPI.makeTweakFromJSON(configCache.getPSetTweaks())
-                if tweak:
-                    tweakFile = "%s/%s" % (stepPath, tweakTarget)
-                    tweak.persist(tweakFile, "json")


khurtado:
@amaltaro Sorry I left this pending... I just removed the supposedly unused tweak file from the code. Can we inject a workflow with this test? I don't think the full suite of workflows is needed, only a few to make sure nothing crashes.

amaltaro:
test this please

amaltaro:
@khurtado Kenyi, I lost track whether this has already been tested or not, in one of our tests. Would you know?

khurtado:
@amaltaro If I remember correctly, you did apply this patch in one of the first tests we did on April. Nothing strange was seen.

amaltaro:
Thanks Kenyi! To be on the safe side, I ran a few more workflows with this patch applied to both reqmgr2 and WMAgent, a couple of workflows have already completed without any issues. Hopefully it's not used by T0, but we will find this out in the near future ;)

amaltaro:
I overlooked the failure in that unittest, which just popped up in a different PR that I'm testing:
https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/11693/testReport/junit/WMCore_t.WMSpec_t.Steps_t.Fetchers_t.CMSSWFetcher_t/URLFetcherTest/testA_BasicFunction/

@khurtado could you please fix that unit test and make a new PR? Thanks

PR #10438, opened 2021-04-06
Update CMS WM overview schematic

### LINKED ISSUE #10437: Update the WM schematic overview
**Impact of the new feature**
Documentation

Our current WM schematic is outdated.

Update the WM schematic overview considering the latest changes in our infrastructure, such as:
* PhEDEx is gone
* dashboard is gone
* MSRuleCleaner needs to be added
* MongoDB needs to be added.

Add the origin diagram to GH as well, not only the png image.

**Additional context**
none

amaltaro:
Fixes #10437 

#### Status
ready

#### Description
Remove the outdated Production schematic in favor of a freshly created one. In short, new schematic contains all the current microservices available in production, PhEDEx and Dashboard have been dropped.

Note: diagram made with draw.io tool, which can be used either on the browser or as a desktop app.

yes

#### Related PRs
none

none


todor-ivanov (changes requested):
Thanks @amaltaro. I like the new organization of this diagram, however I have few things to comment, which you may consider before you merge:
1. The aspect ratio of the previous picture was much better
2. I like the idea of having a standardized database icon (makes it much more distinguishable), but I would be happier if the set of connections are better visualized (e.g. MongoDb) is not connected at all
3. MongoDB in the current case is not separate as a centralized database even though it is installed separately, but for the moment it is serving onlu MSOutput and is reached only through its REST, so I would rather argue this one belongs to the micro services section above.
4. I would like to have  a minimum visualization of the complexity of CouchDB + the replication channels. More precisely it must have at the very least two more (smaller) database icons reflecting the database instances: one in the wmagents areas and one in the central services area
5. It would be good if we can express the act that WMAgent is a component separate and living inside an HTCondor Schedd (meaning - I would add an icon for the schedd connected to the agent visualizing/stressing that WMAgents are  currently a two component systems
6. I like the arrows from the previous graphics more:
     - They are more thick and sharp - making them more expressive
     - They were colored - making much easier to read the connection sets and the information flow
     - 
7.   It would be also great if we can visualize the information flow with those arrows - meaning adding  a second direction to the channels that  are bidirectional and set the correct information direction for the unidirectional channels (e.g. CRIC to WMAgents should be reversed, since we just read info from there)
8. I think we should add also the MariaDB/Oracle database instance in the WMAgents area
9. I would use different colors for CRIC and RUCIO similarly to Unified, since those are two different systems serving two different purposes for CMS - one serving as the CMS information system and other as the data management systems
10. Same comment for DBS  

There were some more here, but I think if we can handle those at the very least the picture would deliver much more information.

todor-ivanov:
Hi @amaltaro  some more came to mind:

11. I would change the title to: "CMS Workflow Management System"
12. Regarding those arrows, I would hold to strict pattern of using them - which should be: from component to component and create a channel per component, instead of having communication channels visualized in per component on one end and a whole service area on the other. I am talking about the micro services (the source of information channel is connected to the whole area) to all others  (the other end is referring to separate components).
13. I would color all of those channels differently and make them run separately in parallel to each other so that on a single glimpse one grasp a components relations without the need to follow a set of arrows joining a common communication channel at some place and later on separating to reach another component leaving the reader with a big dose of uncertainty whether this  arrow is related to the communication channel of the component he is currently looking at.    
15. I would also reflect (similarly to the cmsweb area) the inner service area connections inside a given service area. (i.e. again micro services - no need for all connections, but at least the most important ones)

amaltaro:
> 1. The aspect ratio of the previous picture was much better.

If I change the aspect ration, I will have to stretch the image in a non-proportional way. What I'm doing though is to increase it's resolution to 300DPI (instead of 100). If you have any other specific suggestion, please let me know.

> 2. I like the idea of having a standardized database icon (makes it much more distinguishable), but I would be happier if the set of connections are better visualized (e.g. MongoDb) is not connected at all

Yes, I'm trying to decrease the number of connections in order to make the diagram cleaner. Of course, we lose some details with that. I can try to replace the "general" connectors by very specific service connections/dependencies and we see whether that becomes too visually polluted or not.

> 3. MongoDB in the current case is not separate as a centralized database even though it is installed separately, but for the moment it is serving onlu MSOutput and is reached only through its REST, so I would rather argue this one belongs to the micro services section above.

Good point, I will update the diagram.

> 4. I would like to have a minimum visualization of the complexity of CouchDB + the replication channels. More precisely it must have at the very least two more (smaller) database icons reflecting the database instances: one in the wmagents areas and one in the central services area

I'd say the replication is actually covered by the "general" grey connector between CMSWEB and Pool of Agents (with the CouchDB logo). As replied on your point 2, I can create specific connectors and see if the diagram is still clear enough.
Another option would be to create 2 diagrams, one with the services "overview", one with services overview + dependencies/flow.

> 5. It would be good if we can express the act that WMAgent is a component separate and living inside an HTCondor Schedd (meaning - I would add an icon for the schedd connected to the agent visualizing/stressing that WMAgents are currently a two component systems

Adding a HTCondor logo is a good idea indeed. But I would not say "WMAgents are a two component systems". HTCondor is just another (critical) dependency in our system, but it isn't a WMCore product (thus we cannot say it's part of the WMAgent software/service).

> 6. I like the arrows from the previous graphics more:
>   They are more thick and sharp - making them more expressive
>   They were colored - making much easier to read the connection sets and the information flow

I think rounded arrows make it easier to see where it goes, especially when there are tens of them around. Sharp arrows are not clear whether there was a curve or not.
The colored arrows were used for the WMCore MS communication/dependency. I can retrieve those, but then I better come up with a small legend as well.

> 7. It would be also great if we can visualize the information flow with those arrows - meaning adding a second direction to the channels that are bidirectional and set the correct information direction for the unidirectional channels (e.g. CRIC to WMAgents should be reversed, since we just read info from there)

I feel like the previous diagram had some inconsistencies, or when I made it I considered something else. The way this one is built is basically a source/destination (dependency) flow. E.g. CRIC does not do anything with WMCore services, but actually our services depend on that service and GET information from it.

> 8. I think we should add also the MariaDB/Oracle database instance in the WMAgents area

It can be done for MariaDB, Oracle lives outside though. I'll make some changes on this.

> 9. I would use different colors for CRIC and RUCIO similarly to Unified,

Will update it as well.

todor-ivanov:
Thanks @amaltaro
>  I can try to replace the "general" connectors by very specific service connections/dependencies and we see whether that becomes too visually polluted or not.

I think this is the right way of doing it.And I was already voting for this exact approach in my second set of comments I just sent to you.

> Another option would be to create 2 diagrams, one with the services "overview", one with services overview + dependencies/flow.

If it does not hurt this may help to compare the end result, but I am confident that visualizing the information flow and component dependencies won't make the diagram too heavy to read, rather the opposite.

> HTCondor is just another (critical) dependency in our system, but it isn't a WMCore product (thus we cannot say it's part of the WMAgent software/service)

In some degree similar to DBS and CRIC, but we still visualize them in the given area. I'd say,  by putting HTCondor schedd there we won't claim we own the service, but we should express that dependency I think. It also reflects that fact that those two components does live in a single entity (machine or container), which I think is the most important to be stressed here.

 > but then I better come up with a small legend as well.

That would be great, indeed.

>The way this one is built is basically a source/destination (dependency) flow. E.g. CRIC does not do anything with WMCore services, but actually our services depend on that service and GET information from it.

This is one way of thinking of it, I agree. But then you will need one set of information flow arrows and one set of actuators correlations, which in my opinion is the thing that would make the information too much.

> It can be done for MariaDB, Oracle lives outside though. I'll make some changes on this.

Good point, but I think we can put also Oracle outside the WMAgent area similarly to MongoDB and others (but not inside CMSWeb area - just purely independent)

amaltaro:
Even though it's boring to make all of this, that was a pretty good review and I like the initial outcome, thanks Todor.

I have just pushed a second commit - with a v2 png file - where I considered some of your comments. To me, it looks much better now. Could you please have a look at that one and let me know what you think? I have not done anything with the connectors yet, will probably work on a v3 for that... but just in case I don't manage to finish it before the presentation at 3pm, it might be better to present the v2 schematic on those slides.

todor-ivanov:
Thanks @amaltaro !

Indeed, things look much better now in v2. I spotted a single error there, though. Unified does not communicate with the GlideinWMS Global pool directly (it has no connection to the collector or to the Factory itself), but it rather acts on the pool through the schedds, which should say - its arrow must point to the agents area.

>  but just in case I don't manage to finish it before the presentation at 3pm, it might be better to present the v2 schematic on those slides.

Agreed on that.

amaltaro:
Thanks Todor, I have the unified arrow fixed now. I also reworded the technologies sentence, even though I still don't like it 100%...

amaltaro:
Todor, 3rd commit has a new png version where the generic large arrows have been replaced by specific connectors. It doesn't look as bad as I expected, so it might indeed be the way to go. Please let me know what your thoughts are.

todor-ivanov:
Thanks  @amaltaro. it indeed looks better. Did you try to add some colors to the different component connections, because we  have not yet gotten rid of the effect of smearing the communication channels in a common line?

amaltaro:
I have separated the flow communication/dependency as of (legend on top left side):
* blue connectors for WMCore initiated
* black connectors for any other dependency/external service.

If it's good enough, I'll cleanup this PR and get it merged.

todor-ivanov:
Thanks @amaltaro looks good to me now. 
We can spend some more effort also in the future, but let's not waste any more time on that right now and concentrate on more important things! Thanks  again!

todor-ivanov (approved):
Thanks @amaltaro 
Looks good to me

PR #10454, opened 2021-04-16
Create symbolic links in deploy-wmagent

### LINKED ISSUE #9805: Remove/replace MySQL-python library by something else that supports python3

todor-ivanov:
Fixes #9805 

#### Status
ready

#### Description
In order to be able to test different combinations of rpm packages and dependencies, we need to have the ability to build and upload them in our personal repositories in `cmsrep` [1]. While deploying from such a repository (adding the option `-r comp=comp.<user>` to the deployment script) the deployment area changes from `/data/srv/wmagent/current/sw` to /data/srv/wmagent/current/sw.<user>`, this needs to be properly reflected in the set of symbolic links we create during the installation process. 

Another change that this PR introduces is a symbolic link fixing the different path to the manage script embedded in the `$manage` variable when it comes to deployment  of the `wmagentpy3` package (adding the option `--py3` to the deployment script)

[1]
http://cmsrep.cern.ch/cgi-bin/repos/comp.tivanov/slc7_amd64_gcc630

YES

#### Related PRs

### DIFF
--- a/deploy/deploy-wmagent.sh
+++ b/deploy/deploy-wmagent.sh
@@ -277,9 +277,7 @@ set +e
 
 echo -e "\n*** Creating wmagent symlinks ***"
 cd $CURRENT_DIR
-
-[[ sw${REPO##comp=comp} != sw ]] && [[ -d sw${REPO##comp=comp} ]] &&  ln -s sw${REPO##comp=comp} sw
-ln -s ../sw/${WMA_ARCH}/cms/${RPM_NAME}/${WMA_TAG} apps/wmagent
+ln -s ../sw${REPO##comp=comp}/${WMA_ARCH}/cms/${RPM_NAME}/${WMA_TAG} apps/wmagent
 ln -s ../config/${RPM_NAME} config/wmagent
 
 cd -
--- a/deploy/deploy-wmagent.sh
+++ b/deploy/deploy-wmagent.sh
@@ -275,9 +275,13 @@ for step in prep sw post; do
 done
 set +e
 
-echo -e "\n*** Creating wmagent symlink ***"
+echo -e "\n*** Creating wmagent symlinks ***"
 cd $CURRENT_DIR
+
+[[ sw${REPO##comp=comp} != sw ]] && [[ -d sw${REPO##comp=comp} ]] &&  ln -s sw${REPO##comp=comp} sw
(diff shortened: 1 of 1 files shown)


amaltaro (commented):
Thanks for providing this patch, Todor.

However, I wonder if we really need the symlink for `sw`? Is it related to the init.sh script which adds libraries to the environment? That setup.sh/init.sh works fine for python2-based specs, so I would expect it to be the same for python3 as well.

todor-ivanov:
Hi @amaltaro, 
Yes, one of the places where this link is needed is exactly the sourcing of `init.sh`, but it is not related to the python3 migration only. This is something which is related to the fact that whenever you deploy from your personal repository (e.g. `comp.amaltaro`) the deployment area directory name suddenly changes from `sw` to `sw.amaltaro` regardless of the python version. So anything which relates to `sw` does not work equally bad in both python2 and python3.

amaltaro:
It has always been like this. In other words, there is nothing that needs to be fixed IMO (if you saw anything breaking, then please provide further details). It's just a matter of explicitly demonstrating from where your software has been deployed from.

todor-ivanov:
Thanks for your comment @amaltaro 
> It's just a matter of explicitly demonstrating from where your software has been deployed from.

No one questions the reason behind this naming convention. It is simply that the rest of the environment we run the service in is not compatible with this. And the link is the way to fix this. 

I did not preserve the logs, but IIRC even the deployment process of a MySQL based agent was having troubles to complete, without applying this patch to the deployment script.

todor-ivanov:
Hi @amaltaro I just pushed the shortest possible fix tho this softlink creation as we discussed yesterday, which should make this point to the right deployment even for deployments from private repositories. Would you take a look please?

amaltaro (approved):
Thanks Todor, it looks pretty good now!

PR #10458, opened 2021-04-20
Retrieve only a summary of the condor jobs in DrainStatusPoller

### LINKED ISSUE #8779: AgentStatusWatcher silently dying
It happened twice in vocms0250 this morning. The process gets killed by something
Component:AgentStatusWatcher Not Running

and logs become quiet, of course
```
2018-09-03 10:24:33,839:139914991240960:INFO:AgentStatusPoller:WMBS data collected in: 10 secs
2018-09-03 10:24:34,421:139914991240960:INFO:AgentStatusPoller:Local WorkQueue data collected in: 0 secs
2018-09-03 10:24:35,250:139914991240960:INFO:BaseWorkerThread:AgentStatusPoller took 28.950 secs to execute
```
and nothing else

### LINKED ISSUE #10170: AgentStatusWatcher errors due to interrupted polling cycle of DrainStatusPooler
**Impact of the bug**
WMAgent

**Describe the bug**
Recently we started noticing in the agents that `AgentStatausWatcher` goes in red. The usual cause is a stuck and not updated thread - `DrainStatusPooler`. No clue what is causing it yet. We only noticed that the sub component in question just stops following its regular pooling cycle, which in general should be once per hour. A single log record for a healthy sub component run is given in [1].

[1]
```
2020-12-10 04:15:25,947:139645241743104:INFO:DrainStatusPoller:Running agent drain algorithm...
2020-12-10 04:16:14,845:139645241743104:INFO:DrainStatusPoller:Checking speed drain threshold for NoJobRetries. 
2020-12-10 04:16:14,846:139645241743104:INFO:DrainStatusPoller:Checking speed drain threshold for EnableAllSites. 
2020-12-10 04:16:14,846:139645241743104:INFO:DrainStatusPoller:Checking speed drain threshold for CondorPriority. 
2020-12-10 04:16:22,538:139645241743104:INFO:DrainStatusPoller:Finished collecting agent drain status.
2020-12-10 04:16:22,538:139645241743104:INFO:DrainStatusPoller:Drain stats: {'workflows_completed': False}
2020-12-10 04:16:22,538:139645241743104:INFO:BaseWorkerThread:DrainStatusPoller took 56.591 secs to execute
```
 
**How to reproduce it**
Unknown - the cause is yet not figured out.

**Expected behavior**
`DrainStatusPooler` to follow its regular polling cycle

N/A

amaltaro:
Fixes #8779 

#### Status
ready

#### Description
Query condor schedd only once, and ask for a summary of the condor jobs (instead of two queries, for idle and running jobs). It should also drastically decrease the memory footprint, since it no longer returns all the job classads.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusAPI.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusAPI.py
@@ -21,6 +21,7 @@ class DrainStatusAPI(object):
         self.localBackend = WorkQueueBackend(config.WorkQueueManager.couchurl)
         self.dbsUtil = DBSBufferUtil()
         self.condorAPI = PyCondorAPI()
+        self.condorStates = ("Running", "Idle")
 
     def collectDrainInfo(self):
         """
@@ -51,15 +52,13 @@ class DrainStatusAPI(object):
         Check idle and running jobs in Condor
         """
         results = {}
-        queries = [["1", "idle"], ["2", "running"]]
-
-        for query in queries:
-            jobs = self.condorAPI.getCondorJobs("JobStatus=="+query[0], [])
+        jobs = self.condorAPI.getCondorJobsSummary()
+        for state in self.condorStates:
             # if there is an error, report it instead of the length of an empty list
-            if jobs is None:
-                results[query[1]] = "unknown (schedd query error)"
+            if not jobs:
+                results[state.lower()] = None
             else:
-                results[query[1]] = len(jobs)
+                results[state.lower()] = int(jobs[0].get(state))
 
         return results
 
--- a/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
+++ b/src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
@@ -38,6 +38,7 @@ class DrainStatusPoller(BaseWorkerThread):
         self.validSpeedDrainConfigKeys = ['CondorPriority', 'NoJobRetries', 'EnableAllSites']
         self.reqAuxDB = ReqMgrAux(self.config.General.ReqMgr2ServiceURL)
         self.emailAlert = EmailAlert(config.EmailAlert.dictionary_())
+        self.condorStates = ("Running", "Idle")
 
     @timeFunction
     def algorithm(self, parameters):
@@ -55,7 +56,6 @@ class DrainStatusPoller(BaseWorkerThread):
             return
 
         try:
-
             # see if the agent is in drain mode
             if self.agentConfig["UserDrainMode"] or self.agentConfig["AgentDrainMode"]:
                 # check to see if the agent hit any speed drain thresholds
@@ -170,15 +170,16 @@ class DrainStatusPoller(BaseWorkerThread):
         that need updated for speed draining
         """
         enableKeys = []
+        # first, update our summary of condor jobs
+        totalJobs = self.getTotalCondorJobs()
+        if totalJobs is None:
+            msg = "Cannot check speed drain because there was an error fetching job summary from HTCondor."
+            msg += " Will retry again in the next cycle."
+            logging.warning(msg)
+            return []
 
         # get the current speed drain status
         speedDrainConfig = self.agentConfig.get("SpeedDrainConfig")
-
-        # get condor jobs
-        jobs = self.condorAPI.getCondorJobs("", [])
-        if jobs is None:
-            logging.warning("There was an error querying the schedd.  Not checking speed drain thresholds.")
-            return []
 
         # loop through the speed drain configuration and make a list of what thresholds have been hit
         for k, v in viewitems(speedDrainConfig):
@@ -187,10 +188,26 @@ class DrainStatusPoller(BaseWorkerThread):
                 # we always want to apply the condor priority change if the threshold is hit
                 if not v['Enabled'] or k == 'CondorPriority':
                     logging.info("Checking speed drain threshold for %s. ", k)
-                    if len(jobs) < v['Threshold']:
+                    if totalJobs < v['Threshold']:
                         logging.info("Agent will update speed drain configuration for %s. ", k)
                         enableKeys.append(k)
             else:
                 logging.warning("Speed drain configuration error for %s.  Please check aux db contents.", k)
 
         return enableKeys
+
+    def getTotalCondorJobs(self):
+        """
+        Retrieve a summary of the jobs in condor and return an absolute number
+        of the jobs in Idle and Running states.
+        :return: returns an integer with the total number of jobs, or None if it failed.
+        """
+        jobs = self.condorAPI.getCondorJobsSummary()
+        if not jobs:
+            return None
+
+        results = 0
+        if jobs:
+            for state in self.condorStates:
+                results += int(jobs[0].get(state))
+        return results
--- a/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
+++ b/src/python/WMCore/Services/PyCondor/PyCondorAPI.py
@@ -9,7 +9,6 @@ from __future__ import print_function, division
 from builtins import str, object
 from past.builtins import basestring
 import logging
-
 try:
     # This module has dependency with python binding for condor package (condor)
     import htcondor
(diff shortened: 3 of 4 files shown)


amaltaro:
I'm testing this patch on submit3, so far so good. Thread execution time went down from ~55secs to just under a second as well.
I also have the impression that this might fix the issue: https://github.com/dmwm/WMCore/issues/8779 , since we no longer retrieve and load all the job classsads in memory.

src/python/WMComponent/AgentStatusWatcher/DrainStatusAPI.py
```diff
@@ -46,20 +46,26 @@ def checkWorkflows(self):
...
+        for state in states:
+            if jobs is None and totalOnly:
+                return None
             # if there is an error, report it instead of the length of an empty list
-            if jobs is None:
-                results[query[1]] = "unknown (schedd query error)"
+            elif jobs is None:
+                results[state.lower()] = "unknown (schedd query error)"
+            elif totalOnly:
+                results += int(jobs[0].get(state))
```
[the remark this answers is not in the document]
todor-ivanov: And actually this should be exactly the way we should go. Because this will also be a softer query to the schedd, since it does not try to return the full list of classads. Here could be one example [1] how this is usually used in  the command line options and in the python bindings respectively. And calling the current function with `TotalOnly=True` is again a full new unique query, which in no way benefits from the fact that you unify the list  of states we query for in a single call.  So we gain nothing by keeping it under the same function, instead of confusion. 

[1]
https://htcondor.readthedocs.io/en/latest/man-pages/condor_q.html
Looking at option `-total`.  And what we should try with python bindings is giving that option as a parameter to the `htcondor.schedd.query()` method. like:

https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/htcondor.html#htcondor.Schedd.query
https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/htcondor.html#htcondor.QueryOpts.SummaryOnly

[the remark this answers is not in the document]
todor-ivanov: Since we are on the topic, I believe the large memory footprint is due to the fact that in the `condorAPI` we use `htcondor.Schedd.query()` instead of `htcondor.schedd.xquery()`. It is mentioned in the warning in the description of the `query()` method in the HTCondor documentation cited above.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -174,8 +174,8 @@ def checkSpeedDrainThresholds(self):
         # get the current speed drain status
         speedDrainConfig = self.agentConfig.get("SpeedDrainConfig")
 
-        # get condor jobs
-        jobs = self.condorAPI.getCondorJobs("", [])
+        # get a summary of the condor jobs
+        jobs = self.drainAPI.checkCondorStates(totalOnly=True)
```
todor-ivanov, line 178: If my previous comment is to be taken into account, this call here should simply be converted to a call of a separate method, instead to a call of the same method with additional flag.

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -22,22 +22,31 @@ def __init__(self):
         self.schedd = htcondor.Schedd()
         self.coll = htcondor.Collector()
 
-    def getCondorJobs(self, constraint, attr_list):
+    def getCondorJobs(self, constraint='true', attrList=None, opts="SummaryOnly"):
```
todor-ivanov, line 25: looking at this `opts="SummaryOnly"`, I realize you were already aware of the option path I was previously talking about. And it is quite good you added it here, so I would suggest to use it in order to implement my suggestion above. Thanks for that one -  I really like it!

todor-ivanov (changes requested):
Thanks @amaltaro for this PR ! These suggestions of yours are for sure improving the things a lot. I have put few comments inline though, which I think  you should go through and maybe consider my suggestions.

amaltaro:
Repurposing this PR to actually fixing https://github.com/dmwm/WMCore/issues/8779 instead of #10170, since I fail to fix the latter.

amaltaro:
@todor-ivanov Todor, after your review, I have pushed two other commits that should be addressing your concerns. I still have to squash them and run a final test, but it would be great to have your feedback before I do so. Thanks

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
```diff
@@ -170,27 +170,44 @@ def checkSpeedDrainThresholds(self):
...
+        of the jobs in Idle and Running states.
+        :return: returns an integer with the total number of jobs, or None if it failed.
+        """
+        jobs = self.condorAPI.getCondorJobsSummary()
+        if not jobs:
+            return None
+
+        results = 0
+        if jobs:
+            for state in self.condorStates:
```
todor-ivanov, line 211: Do we care only about `Idle` and `Running` jobs here? What about the rest of the states? If we are about to drain a schedd my understanding is that we should suppose that it is free of not only `Idle` && `Running` jobs but also the ones in `Hold`. And there could be many reasons for a job to be in `Hold`, some of which may even require some debugging.

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -18,34 +17,64 @@
...
+            https://htcondor.readthedocs.io/en/v8_9_7/apis/python-bindings/api/htcondor.html#htcondor.QueryOpts
+        :return: returns an iterator to the job classads
+        """
+        attrList = attrList or []
+        # if option parameter is invalid, default it to the standard behavior
+        opts = getattr(htcondor.htcondor.QueryOpts, opts, "Default")
+        msg = "Querying condor schedd with params: constraint=%s, attrList=%s, limit=%s, opts=%s"
+        logging.info(msg, constraint, attrList, limit, opts)
+        try:
+            return self.schedd.xquery(constraint, attrList, limit, opts=opts)
```
todor-ivanov, line 71: This is great!

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -18,34 +17,64 @@
...
+        except Exception:
+            self.recreateSchedd()
+
+        # if we hit another exception, let it be raised up in the chain
+        return self.schedd.xquery(constraint, attrList, limit, opts=opts)
```
todor-ivanov, line 76: This one too!

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
```diff
@@ -18,34 +17,64 @@
...
+        self.schedd = htcondor.Schedd()
 
-        Given a job/schedd constraint, return a list of jobs attributes
-        or None if the query to condor fails.
+    def getCondorJobsSummary(self):
+        """
+        Retrieves a job summary from the HTCondor Schedd object
+        :return: a list of classads representing the matching jobs, or None if failed
         """
+        ### NOTE: SummaryOnly does not work with xquery method
```
todor-ivanov, line 40: I know in this case it doesn't matter if the background query is `query` or `xquery` because we do expect only the total numbers to be provided  by HTcondor, but just for my curiosity, is it documented in the HTCondor the fact that this option is not supported with `xquery` or you had to discover it yourself?

todor-ivanov (approved):
Thanks @amaltaro! You did address the changes requested. There are a comment or two I've put inline, but those are just clarifying questions. The code look perfectly good as it is.

src/python/WMComponent/AgentStatusWatcher/DrainStatusPoller.py
amaltaro: Yes, here we should consider only Idle/Running. Because depending on their total number, we bump their priority or not. So we do not want to affect this decision by jobs lingering around in the `Held` (5) or `Removed` (3) states.
It might be useful though in the drain stats... let me check.

src/python/WMCore/Services/PyCondor/PyCondorAPI.py
amaltaro: I haven't seen any documentation on that, so I had to find it the hard way.
If it isn't too expensive to file an HTCondor issue, I might do so (even though we are not using their latest version...)

amaltaro:
Thanks for your review and comments, Todor.
Looking again at the output of the SummaryOnly option:
```
[[ SchedulerHeld = 0; SchedulerRemoved = 0; SchedulerRunning = 0; SchedulerIdle = 0; ServerTime = 1619512078; Held = 0; AllusersSchedulerRemoved = 0; AllusersSchedulerCompleted = 0; ErrorCode = 0; SchedulerCompleted = 0; AllusersSchedulerJobs = 0; AllusersRemoved = 0; SchedulerJobs = 0; Running = 2716; AllusersHeld = 0; AllusersSchedulerHeld = 0; Removed = 0; AllusersCompleted = 0; Suspended = 0; AllusersJobs = 4690; AllusersRunning = 2716; MyType = "Summary"; AllusersSchedulerIdle = 0; Jobs = 4690; AllusersIdle = 1974; Idle = 1974; AllusersSuspended = 0; AllusersSchedulerRunning = 0; Completed = 0 ]]
```

we could have a minor improvement to the drain stats, by also providing the total number of `Jobs` in the queue. But then it requires changes to the couchapps as well. I'd rather leave that for another moment...

amaltaro:
I did some quick tests and things look fine. I have also decided not to backport it to the current production release, since it does not fix the draining issue. Thanks for the review Todor.

PR #10474, opened 2021-04-29
Properly evaluate stuck rules in MSTransferor

### LINKED ISSUE #10473: [MSTransferor] TypeError exception when evaluating STUCK rules
**Impact of the bug**
MSTransferor

**Describe the bug**
This issue popped up today while talking to Rucio Int, and it's not clear to me whether it could show up in production as well.

There is a data type mismatch when evaluating a rule that is in the STUCK state, which could have `stuck_at=None`, thus raising an exception when getting the difference between its creation and the moment it got stuck. See traceback (with an additional debugging put in place) here [1]

**How to reproduce it**
Not sure

**Expected behavior**
There are two items to be fixed here:
1) `stuck_at` could be None. The code needs to handle it (still figuring out the details)
2) we need to properly convert those string date/time to something that we can operate/subtract

[1]
```
2021-04-29 17:38:20,955:INFO:PycurlRucio: AMR item: {u'locks_ok_cnt': 0, u'source_replica_expression': None, u'weight': None, u'purge_replicas': False, u'rse_expression': u'T1_US_FNAL_Disk', u'updated_at': u'Thu, 29 Apr 2021 15:19:30 UTC', u'meta': None, u'child_rule_id': None, u'id': u'3a4f4e10708c425cb6858855667a43a4', u'ignore_account_limit': False, u'locks_stuck_cnt': 8, u'locks_replicating_cnt': 0, u'notification': u'NO', u'copies': 1, u'comments': u'WMCore MSTransferor input data placement', u'split_container': False, u'priority': 3, u'state': u'STUCK', u'scope': u'cms', u'subscription_id': None, u'ignore_availability': False, u'stuck_at': None, u'error': u'RequestErrMsg.MISMATCH_SCHEME:RequestState.MISMATCH_SCHEME', u'eol_at': None, u'expires_at': u'Fri, 30 Apr 2021 13:15:42 UTC', u'did_type': u'CONTAINER', u'account': u'wmcore_transferor', u'locked': False, u'name': u'/RelValMinBias_14TeV/CMSSW_11_2_0_pre8-112X_mcRun3_2024_realistic_v10_forTrk-v1/GEN-SIM', u'created_at': u'Thu, 29 Apr 2021 13:15:42 UTC', u'activity': u'Production Input', u'grouping': u'ALL'}
2021-04-29 17:38:20,955:ERROR:PycurlRucio: listReplicationRules function did not return a valid response for container: /RelValMinBias_14TeV/CMSSW_11_2_0_pre8-112X_mcRun3_2024_realistic_v10_forTrk-v1/GEN-SIM.Server responded with: unsupported operand type(s) for -: 'NoneType' and 'unicode'
Error: {"id": "3a4f4e10708c425cb6858855667a43a4", "subscription_id": null, "account": "wmcore_transferor", "scope": "cms", "name": "/RelValMinBias_14TeV/CMSSW_11_2_0_pre8-112X_mcRun3_2024_realistic_v10_forTrk-v1/GEN-SIM", "did_type": "CONTAINER", "state": "STUCK", "error": "RequestErrMsg.MISMATCH_SCHEME:RequestState.MISMATCH_SCHEME", "rse_expression": "T1_US_FNAL_Disk", "copies": 1, "expires_at": "Fri, 30 Apr 2021 13:15:42 UTC", "weight": null, "locked": false, "locks_ok_cnt": 0, "locks_replicating_cnt": 0, "locks_stuck_cnt": 8, "source_replica_expression": null, "activity": "Production Input", "grouping": "ALL", "notification": "NO", "stuck_at": null, "purge_replicas": false, "ignore_availability": false, "ignore_account_limit": false, "priority": 3, "comments": "WMCore MSTransferor input data placement", "child_rule_id": null, "eol_at": null, "split_container": false, "meta": null, "created_at": "Thu, 29 Apr 2021 13:15:42 UTC", "updated_at": "Thu, 29 Apr 2021 15:19:30 UTC"}
  File "/data/srv/HG2105c/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.8.pre3/lib/python2.7/site-packages/WMCore/MicroService/Tools/PycurlRucio.py", line 192, in listReplicationRules
    timeDiff = item['stuck_at'] - item['created_at']
TypeError: unsupported operand type(s) for -: 'NoneType' and 'unicode'
```

amaltaro:
Fixes #10473 

#### Status
not-tested

#### Description
Summary of changes is:
* properly handle stuck rules with `stuck_at=None`. It considers the rule to be stuck since its creation.
* correctly convert the date/time string to EPOCH time before operating over them
* last but not least, refactored logic for calculating how long a rule has been stuck: `utcTimeNow - stuckAt`

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/MicroService/Tools/PycurlRucio.py
+++ b/src/python/WMCore/MicroService/Tools/PycurlRucio.py
@@ -50,6 +50,18 @@ def parseNewLineJson(stream):
             yield json.loads(line)
 
 
+def stringDateToEpoch(strDate):
+    """
+    Given a date/time in the format of:
+        'Thu, 29 Apr 2021 13:15:42 UTC'
+    it returns an integer with the equivalent EPOCH time
+    :param strDate: a string with the date and time
+    :return: the equivalent EPOCH time (integer)
+    """
+    timestamp = datetime.datetime.strptime(strDate, "%a, %d %b %Y %H:%M:%S %Z")
+    return int(timestamp.strftime('%s'))
+
+
 def getRucioToken(rucioAuthUrl, rucioAcct):
     """
     Provided a Rucio account, fetch a token from the authentication server
@@ -69,8 +81,7 @@ def getRucioToken(rucioAuthUrl, rucioAcct):
         tokenExpiration = res.getHeaderKey('X-Rucio-Auth-Token-Expires')
         logging.info("Retrieved Rucio token valid until: %s", tokenExpiration)
         # convert the human readable expiration time to EPOCH time
-        tokenExpiration = datetime.datetime.strptime(tokenExpiration, "%a, %d %b %Y %H:%M:%S %Z")
-        tokenExpiration = int(tokenExpiration.strftime('%s'))
+        tokenExpiration = stringDateToEpoch(tokenExpiration)
         return userToken, tokenExpiration
 
     raise RuntimeError("Failed to acquire a Rucio token. Error: {}".format(res.getReason()))
@@ -188,16 +199,24 @@ def listReplicationRules(containers, rucioAccount, grouping,
                         logging.warning(msg)
                         continue
 
-                    timeDiff = item['stuck_at'] - item['created_at']
-                    if int(timeDiff.days) > STUCK_LIMIT:
+                    # then calculate for how long it's been stuck
+                    utcTimeNow = int(datetime.datetime.utcnow().strftime('%s'))
+                    if item['stuck_at']:
+                        stuckAt = stringDateToEpoch(item['stuck_at'])
+                    else:
+                        # consider it to be stuck since its creation
+                        stuckAt = stringDateToEpoch(item['created_at'])
+
+                    daysStuck = (utcTimeNow - stuckAt) // (24 * 60 * 60)
+                    if daysStuck > STUCK_LIMIT:
                         msg = "Container {} has a STUCK rule for {} days (limit set to: {}).".format(container,
-                                                                                                     timeDiff.days,
+                                                                                                     daysStuck,
                                                                                                      STUCK_LIMIT)
                         msg += " Not going to use it! Rule info: {}".format(item)
                         logging.warning(msg)
                         continue
                     else:
-                        msg = "Container {} has a STUCK rule for only {} days.".format(container, timeDiff.days)
+                        msg = "Container {} has a STUCK rule for only {} days.".format(container, daysStuck)
                         msg += " Considering it for the pileup location"
                         logging.info(msg)
                 else:
--- a/test/python/WMCore_t/MicroService_t/Tools_t/PycurlRucio_t.py
+++ b/test/python/WMCore_t/MicroService_t/Tools_t/PycurlRucio_t.py
@@ -9,7 +9,7 @@ import unittest
 
 from WMCore.MicroService.Tools.PycurlRucio import (getRucioToken, parseNewLineJson,
                                                    getPileupContainerSizesRucio, listReplicationRules,
-                                                   getBlocksAndSizeRucio)
+                                                   getBlocksAndSizeRucio, stringDateToEpoch)
 
 CONT1 = "/Pseudoscalar2HDM_MonoZLL_mScan_mH-500_ma-300/DMWM_Test-TC_PreMix_Agent122_Validation_Alanv1-v11/MINIAODSIM"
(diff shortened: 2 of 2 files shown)


amaltaro:
Just tested it in testbed and it's working fine now :-D

src/python/WMCore/MicroService/Tools/PycurlRucio.py
```diff
@@ -188,16 +199,24 @@ def listReplicationRules(containers, rucioAccount, grouping,
                         logging.warning(msg)
                         continue
 
-                    timeDiff = item['stuck_at'] - item['created_at']
-                    if int(timeDiff.days) > STUCK_LIMIT:
+                    # then calculate for how long it's been stuck
+                    utcTimeNow = int(datetime.datetime.utcnow().strftime('%s'))
+                    if item['stuck_at']:
+                        stuckAt = stringDateToEpoch(item['stuck_at'])
```
todor-ivanov, line 205: I see the benefits from the attempt to calculate the time difference in seconds, but what I am concerned here is that the method `stringDateToEpoch()` does not have any protection from receiving `None` as an argument [1]. And it seems to me this was  the original cause of the problem. So I would say here we will still get the Exception. 

BTW Do we know the conditions under which the `item['stuck_at']` value becomes `None`?

[1]
```
In [4]: stringDateToEpoch(None)
TypeError                                 Traceback (most recent call last)
<ipython-input-4-ced2c9ac0b87> in <module>()
----> 1 stringDateToEpoch(None)

<ipython-input-1-a4fbc704f9e9> in stringDateToEpoch(strDate)
      7     :return: the equivalent EPOCH time (integer)
      8     """
----> 9     timestamp = datetime.datetime.strptime(strDate, "%a, %d %b %Y %H:%M:%S %Z")
     10     return int(timestamp.strftime('%s'))

TypeError: strptime() argument 1 must be string, not None
```

src/python/WMCore/MicroService/Tools/PycurlRucio.py
```diff
@@ -188,16 +199,24 @@ def listReplicationRules(containers, rucioAccount, grouping,
...
+                    else:
+                        # consider it to be stuck since its creation
+                        stuckAt = stringDateToEpoch(item['created_at'])
+
+                    daysStuck = (utcTimeNow - stuckAt) // 86400
```
todor-ivanov, line 210: I would rather write this as:

```
daysStuck = (utcTimeNow - stuckAt) // (24*60*60)
```

todor-ivanov (changes requested):
Thanks @amaltaro for the promptly provided fix. There is one significant comment I left inline. I think we should discuss if it would be an actual problem or not. And if we decide we are safe, then I would approve the PR.

src/python/WMCore/MicroService/Tools/PycurlRucio.py
amaltaro: That's what the `if` statement at L204 is supposed to do. 
Given that `created_at` seems to be a property of the rule creation process, I believe that one would always have a non-None value, hence that parameter is not really protected from a None value. Different than `stuck_at`, which is likely initialized with None value.

src/python/WMCore/MicroService/Tools/PycurlRucio.py
amaltaro: That can be done. Let me update it.

amaltaro:
Thanks for your review Todor. Code has been updated, could you please have another look? I will then request Imram to push it to testbed to avoid hitting it in production as well. @todor-ivanov

src/python/WMCore/MicroService/Tools/PycurlRucio.py
todor-ivanov: I see, that would do the job. It seems we are safe to continue here, but I am curious and we should forward an inquiry to Rucio folks, asking how those values are initialized and changed throughout the transfer life time.

todor-ivanov (approved):
Looks good to me.

src/python/WMCore/MicroService/Tools/PycurlRucio.py
amaltaro: Pushed the wrong button, for the record: https://github.com/rucio/rucio/issues/4592

PR #10484, opened 2021-05-08
Replace mox dependency with mock.

### LINKED ISSUE #7071: Convert mox to mox3
mox3 should be compatible with python 2.7 and 3.5. I think @SaltumDis is already looking at this. Part of #7067

khurtado:
Fixes #7071

#### Status
 ready

#### Description
Replaces mox dependency with mock.
Remove deprecated testMkdir test

Yes

It removes the no longer supported mox dependecy and uses mock instead.

### DIFF
--- a/test/python/WMCore_t/Storage_t/Plugins_t/SRMV2Impl_t.py
+++ b/test/python/WMCore_t/Storage_t/Plugins_t/SRMV2Impl_t.py
@@ -1,259 +1,92 @@
 from __future__ import print_function
-from builtins import object
-import os
 import logging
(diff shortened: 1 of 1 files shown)


mapellidario (approved):
I trust your judgement in what you removed/added and everything else looks perfect! Keeping a third-party library dependency for just one unit test was more of maintenance burden than an advantage and I see no problems with py3 support at the moment.

Thank you Kenyi!

amaltaro (approved):
Thanks Kenyi, it looks good to me as well (and much cleaner!).

Issue #9633, opened 2020-04-14
Crash in StepChain jobs can result in misnamed output files


amaltaro:
**Impact of the bug**
WMAgent (seen in submit3 so far)

**Describe the bug**
It still needs further debugging, but it looks like that if a stepchain job fails in a specific way (or some specific exit code), its job report returns with the standard local output file names, such as:
```
      "/store/unmerged/Run3Summer19DRPremix/QCD_Pt-15_IsoTrkFilter_Pt-30_TuneCP5_14TeV-pythia8/GEN-SIM-RECO/2021ScenarioRECO_106X_mcRun3_2021_realistic_v3-v3/230001/RECOSIMoutput.root",
```

which is wrong and should never be inserted into the database. Also because as soon as this job exits the worker node, that file is no longer available.

As soon as we get 2 jobs failing the same way, we hit a duplicate LFN in the database, which then crashes JobAccountant.

**How to reproduce it**
Still to be investigated

**Expected behavior**
Such file names should never get returned to the agent. There are many ways to see/resolve this issue though:
1) transient file names in a StepChain job should not appear in the job report
2) it should return with a local worker node name path (instead of the usual LFN/PFN)
3) we might have to add some logic to JobAccountant to skip such files from database insertion (they anyways will represent failed jobs)

**Additional context and error message**
Log dump from the script to remove duplicate files in JobAccountant can be found here:

amaltaro:
The system seems to be full of such problems since - at least - the past week. Here is the json unpickled version of the job reported processed by JobAccountant in submit3:
http://amaltaro.web.cern.ch/amaltaro/forWMCore/Issue_9633/badPickle.json

other files, for that same job, can also be found under the same directory.

Having a quick look at the wmagentJob.log, there is something very very suspicious:
```
2021-04-09 07:49:47,781:INFO:PerformanceMonitor:PSS: 3577811; RSS: 3625756; PCPU: 759; PMEM: 2.7
2021-04-09 07:52:15,003:INFO:Report:addOutputFile method called with outputModule: MINIAODSIMoutput, aFile: None
2021-04-09 07:52:15,004:INFO:Report:addOutputFile method fileRef: , whole tree: {}
2021-04-09 07:52:15,004:INFO:Report:addOutputFile method called with outputModule: RECOSIMoutput, aFile: None
2021-04-09 07:52:15,005:INFO:Report:addOutputFile method fileRef: , whole tree: {}
[...]

PR #10486, opened 2021-05-13
JobAccountant workaround for StepChain jobs with duplicate files - wmagent branch

amaltaro:
Workaround for #9633 (real fix still needs to be created against master)

#### Status
ready (tested in submit7)

#### Description
Please read the last comment in the GH issue for further details on the actual root cause of this problem.

With this PR, we do not inject output files into the database if they came without any location in the FJR.
Which is the case for StepChain jobs that crash in the worker node, thus not actually staging those files out to the storage area (and getting an unique file id).

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMComponent/JobAccountant/AccountantWorker.py
+++ b/src/python/WMComponent/JobAccountant/AccountantWorker.py
@@ -454,6 +454,8 @@ class AccountantWorker(WMConnectionBase):
                                                 conn=self.getDBConn(),
                                                 transaction=self.existingTransaction())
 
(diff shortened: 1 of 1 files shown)


amaltaro:
[Fix #10486 - JobAccountant workaround for StepChain jobs with duplicate files](https://trello.com/c/p3JzRAf0/256-fix-10486-jobaccountant-workaround-for-stepchain-jobs-with-duplicate-files)

amaltaro:
This patch has been pushed in in all 1.4.7 production agents.

amaltaro:
A few patches have been accumulated, so I'm going to make a patch4 release.

PR #10518, opened 2021-05-27
[py2py3] new variables Utils.PythonVersion.PY2/3

### LINKED ISSUE #10509: [py2py3] Add variable Utils.Utilities.PY3
**Impact of the new feature**
WMCore

We should provide an easily accessible boolean variable to determine if the runtime is py2 or py3. 

Instead of importing from WMCore/Configuration, we can add this variable to Utils/Utilities.py

We can always use `import sys: if sys.version_info[0] == 3:` in every module, instead of importing it from Utils.Utilities.

This is a valid option, but can get a bit verbose, such as in 

https://github.com/dmwm/WMCore/blob/1a9ec04202c3db836d5a2dd555e686e5845dc214/test/python/Utils_t/MemoryCache_t.py#L32-L44

mapellidario:
Fixes #10509 

#### Status
Ready

#### Description
As Alan suggested: I added a new module under Utils which provides two booleans `PY2` and `PY3`.

TODO
- [ ] we should make sure that in the Release Changelog we specify that we no longer provide the variable `PY3` from `Configuration.py` but from `Utils.PythonVersion`. Maybe someone else is using it.

YES

Nope

### DIFF
--- a/src/python/Utils/PythonVersion.py
+++ b/src/python/Utils/PythonVersion.py
@@ -1,0 +1,10 @@
+"""
+Easily get the version of the python interpreter at runtime
+"""
+
+from __future__ import division # Jenkins CI
+
+import sys
+
+PY3 = sys.version_info[0] == 3
+PY2 = sys.version_info[0] == 2
--- a/src/python/Utils/FileTools.py
+++ b/src/python/Utils/FileTools.py
@@ -11,9 +11,9 @@ import stat
 import subprocess
 import time
 import zlib
-import sys
 
 from Utils.Utilities import decodeBytesToUnicode
+from Utils.PythonVersion import PY3
 
 
 def calculateChecksums(filename):
@@ -52,7 +52,7 @@ def calculateChecksums(filename):
     if len(cksumStdout) != 2 or int(cksumStdout[1]) != filesize:
         raise RuntimeError("Something went wrong with the cksum calculation !")
 
-    if sys.version_info[0] == 3:
+    if PY3:
         # using native-string approach. convert from bytes to unicode in
         # python 3 only.
         cksumStdout[0] = decodeBytesToUnicode(cksumStdout[0])
--- a/src/python/WMCore/Configuration.py
+++ b/src/python/WMCore/Configuration.py
@@ -11,10 +11,9 @@ Module dealing with Configuration file in python format
 
 import imp
 import os
-import sys
 import traceback
 
-PY3 = sys.version_info[0] == 3
+from Utils.PythonVersion import PY3
 
 # PY3 compatibility (can be removed once python2 gets dropped)
 if PY3:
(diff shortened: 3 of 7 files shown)


src/python/Utils/PythonVersion.py
```diff
@@ -0,0 +1,8 @@
+"""
+Easily get the version of the python interpreter at runtime
+"""
+
```
amaltaro, line 4: Python compatibility report requests
```
```
to be added. We might discuss the usefulness of this tomorrow btw

amaltaro (changes requested):
This looks pretty good, Dario. 

I made one comment along the code that we need to fix to avoid getting failures all the time, and can you also sort those unused imports reported in bold in the pylint section?
Not this one though:
 W0611, line 39: Unused import WMCore.REST.Tools 
otherwise it causes problems to CRAB.

PS.: feel free to squash them in their proper commits.

src/python/Utils/PythonVersion.py
mapellidario: Yeah, we can discuss this. I would remove such check, I think that `pylint --py3k` already warns us when we have a division in py2

mapellidario: It should be W1619 at https://pylint.pycqa.org/en/latest/technical_reference/features.html

mapellidario: For example, if we remove `from __future__ import division` from `SiteInfo.py` we get

```plainmtext
[dmwm@02f2a819e6dc WMCore]$ pylint --py3k src/python/WMCore/MicroService/Unified/SiteInfo.py
No config file found, using default configuration
************* Module src.python.WMCore.MicroService.Unified.SiteInfo
W: 12, 0: import missing `from __future__ import absolute_import` (no-absolute-import)
W: 52,37: division w/o __future__ statement (old-division)
W:357,35: division w/o __future__ statement (old-division)

------------------------------------------------------------------
Your code has been rated at 9.91/10 (previous run: 9.97/10, -0.06)
```

While if we remove it from `PerformanceMonitor.py` we get

```plaintext
[dmwm@02f2a819e6dc WMCore]$ pylint --py3k src/python/WMCore/WMRuntime/Monitors/PerformanceMonitor.py
No config file found, using default configuration
************* Module src.python.WMCore.WMRuntime.Monitors.PerformanceMonitor
W: 12, 0: import missing `from __future__ import absolute_import` (no-absolute-import)

-----------------------------------
Your code has been rated at 9.94/10
```

Because `PerformanceMonitor.py`casts the first argument to `float` and does not create problems with the old dfivision

amaltaro: Thanks for this detailed information, Dario. Here is a new issue to get it removed in the coming weeks:
https://github.com/dmwm/WMCore/issues/10519

amaltaro (approved):

amaltaro:
It's good to go now. Thank you for your prompt action, Dario.

Issue #10422, opened 2021-04-01
[py2py3] WMCore should pass all the unit tests run in py3 environment


mapellidario:
**Impact of the new feature**
Whole WMCore should be affected

**List**

* [x] PsetTweaks, #9739, #9740:
  - nothing needed
* [x] Utils, #9730, #9731
  - #10500 
* [ ] WMConfigCache, #9806, #9818
  - #10522
* [ ] WMCore/WMFactory - #9970, #9971
* [ ] WMQuality/CherrypyTestInit - #9972
* [ ] WMCore/Configuration - #9973, #9974
* [ ] WMCore/Credential - #10007, #10008
* [ ] WMCore/Algorithms - #10009, #10011
* [ ] WMCore/DataStructs - #10010, #10012
  - This is not fixed, but we have a specific issue to deal with this: #10169 
* [ ] slice 1 - #10055, #10057
* [ ] slice 2 - #10098, #10103
* [ ] slice 3 - #10114, #10117 
* [ ] slice 4 - #10115, #10165 
* [ ] slice 5 - #10116, #10262 
* [ ] slice 6 - #10127, #10277 
* [ ] slice 7 - #10131, #10284 
* [ ] slice 8 - #10132, #10289 
* [ ] slice 9 - #10133, #10290
* [ ] slice 10 - #10134, #10295
* [ ] slice 11 - #10135, #10299
* [ ] slice 12 - #10136, #10307
* [ ] slice 13 - #10137, #10310 
* [ ] slice 14 - #10138, #10313
* [ ] slice 15 - #10139, #10319
* [ ] slice 16 - #10140, #10321
* [ ] slice 17 - #10141, #10329
* [ ] slice 18 - #10142, #10331
* [ ] slice 19 - #10143, #10348
* [ ] slice 20 - #10144, #10349
* [ ] slice 21 - #10145, #10361
* [ ] slice 22 - #10146, #10362
* [ ] slice 23 - #10147, #10363 
* [ ] slice 24 - #10148, #10371 
* [ ] slice 25 - #10149, #10379 
* [ ] slice 26 - #10150, 

(list taken from #10440)

**Is your feature request related to a problem? Please describe.**
We testes the changes contained in the modernizazion slices ( [PRs with tag py2py3slice](https://github.com/dmwm/WMCore/pulls?q=label%3A%22py2py3+slice%22+is%3Apr) ) with the current jenkins setup, which has python2 only unit tests.

[...]

PR #10522, opened 2021-05-27
[py2py3] fix unittests from WMCoreConfigCache

### LINKED ISSUE #10528: [py2py3] Check unittests from WMCore/Cache in py3
#### Intro

#### Details

This issue is about checking that the unittests from WMCore/Cache,
issue #9806 and PR #9818, do not fail.

```
test/python/WMCore_t/Cache_t/GenericDataCache_t.py
test/python/WMCore_t/Cache_t/WMConfigCache_t.py
test/python/WMCore_t/DataStructs_t/WMObject_t.py
test/python/WMCore_t/Database_t/CMSCouch_t.py
test/python/WMCore_t/GroupUser_t/CouchObject_t.py
test/python/WMCore_t/Lexicon_t.py
test/python/WMCore_t/Services_t/Requests_t.py
test/python/WMCore_t/Services_t/pycurl_manager_t.py
test/python/WMCore_t/Wrappers_t/JsonWrapper_t/JSONThunker_t.py

```

mapellidario:
Fix #10528

#### Status
Ready

#### Related PRs
This is the first of a (hopefully not too long) series of PRs for closing #10422 .

This is a followup of #9818 

##### open questions

`Lexicon.py`
- [ ] we load log files in memory, then we scan them looking for patterns
https://github.com/dmwm/WMCore/blob/50f6097838bd5b6591551fe9b006e06086083af5/src/python/WMCore/Lexicon.py#L728-L737
  - we use [mmap](https://docs.python.org/3/library/mmap.html) to load the file, which behaves like a stream of bytes.  However, we expect to extract native strings (bytes in py2, unicode in py3) from the logs https://github.com/dmwm/WMCore/blob/50f6097838bd5b6591551fe9b006e06086083af5/test/python/WMCore_t/Lexicon_t.py#L873
  - is it there any reason for using `mmap`? can we use a `io.StringIO` buffer in py3? is it enough to use the std lib to load the fle to memory? I opted for the latter option (std library file `.read()`)

##### comments

The py3 unittests that fail are either changes from error to failure and vice versa (not really a problem now, i guess) or unstable unittests. Some of these instabilities derive from the fact that we take for granted the order of some lists in the unittest itself, so they should be easy to fix. Some other are a bit more tricky to fix. In any case, we list there the list of unstable unittests in python3 at [this page](https://github.com/dmwm/WMCore/wiki/setup-wmcore-unittest#unstable-tests-in-py3) of the dmwm/WMCore documentation.

Summary of the changes, grouped by file.

`src/python/WMCore/Database/CMSCouch.py`
- [x] Changed internal working of `Database.addAttachment`. The problem arises from `base64.b64encode()` that returns bytes in both py2 and py3 -> this does not allow the use of "native string" approach. we had to decode its result to unicode only in py3. We decided to change `CMSCouch` in such a way that `CMSCouch_t` required no change, so that all the code that calls `CMSCouch` does not need to do the decoding in py3 and need no change.

WMExceptions_t
- [x] commented some lines, because
  - in python3 keywords of function arguments must be (unicode) strings. bytes are not allowed.
  - `bytes(my_exception)` unexpectedly calls `exception__getitem__()`
  - no impact on our code, just updated the unittest, so that the WMException API has a py3-compatible showcase
 
JSONThunker
- [x] in `_unthunk`, convert unicode to bytes only in py2

`src/python/WMCore/Services/Requests.py`
- [x] `Requests.decodeResult` now decodes bytes strings to unicode in py3. The effect is that `Requests.makeRequest`'s first return value now is a unicode string, as expected by unittest https://github.com/dmwm/WMCore/blob/50f6097838bd5b6591551fe9b006e06086083af5/test/python/WMCore_t/Services_t/Requests_t.py#L353
  - make sure to communicate this change in the API for CRAB and others that use this module.

WMConfigCache
- [x] in PY3, `ConfigCache.addConfig()` stores the string that holds the config file as a unicode string, and not as a bytes string

Yes

nope

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -14,8 +14,6 @@ from builtins import str
 from future.utils import viewvalues
 
 import io
-import logging
-import mmap
 import re
 
 from urllib.parse import urlparse, urlunparse
@@ -727,11 +725,5 @@ def getStringsBetween(start, end, source):
 
 def getIterMatchObjectOnRegexp(filePath, regexp):
     with io.open(filePath, 'r', encoding='utf8', errors='ignore') as f:
-        try:
-            mm = mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ)
-        except ValueError:
-            logging.warning("Cannot mmap empty file: %s", filePath)
-        else:
-            for m in re.finditer(regexp, mm):
-                yield m
-            mm.close()
+        for m in re.finditer(regexp, f.read()):
+            yield m
--- a/src/python/WMCore/Cache/WMConfigCache.py
+++ b/src/python/WMCore/Cache/WMConfigCache.py
@@ -19,6 +19,8 @@ from contextlib import closing
 
 import WMCore.GroupUser.Decorators as Decorators
 from Utils.Patterns import Singleton
+from Utils.Utilities import decodeBytesToUnicode
+from Utils.PythonVersion import PY3
 from WMCore.DataStructs.WMObject import WMObject
 from WMCore.Database.CMSCouch import CouchNotFoundError
 from WMCore.Database.CMSCouch import CouchServer, Document
@@ -439,6 +441,8 @@ class ConfigCache(WMObject):
         with closing(urllib.request.urlopen(newConfig)) as f:
             configString = f.read(-1)
         configMD5 = hashlib.md5(configString).hexdigest()
+        if PY3:
+            configString = decodeBytesToUnicode(configString)
 
         self.document['md5_hash'] = configMD5
         self.document['pset_hash'] = psetHash
--- a/src/python/WMCore/Database/CMSCouch.py
+++ b/src/python/WMCore/Database/CMSCouch.py
@@ -9,10 +9,12 @@ http://wiki.apache.org/couchdb/API_Cheatsheet
 NOT A THREAD SAFE CLASS.
 """
 from __future__ import print_function, division
+from builtins import str as newstr, bytes as newbytes, object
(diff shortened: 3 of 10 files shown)


amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
Now running one final test, with the official image and jenkins with final configuration...

src/python/WMCore/Lexicon.py
```diff
@@ -727,11 +725,5 @@ def getStringsBetween(start, end, source):
 
 def getIterMatchObjectOnRegexp(filePath, regexp):
     with io.open(filePath, 'r', encoding='utf8', errors='ignore') as f:
-        try:
```
amaltaro, line 730: I honestly do not know what I decided to implement this with `mmap`, after all, whatever files we read, we load it into memory...
Still, just a note because I do not know whether its removal can cause issues in the near future.

amaltaro (approved):
It looks good to me, thanks Dario.
I made one comment in the code though for something that we might have to come back in the future, it's not clear to me ATM.

As we chatted over slack yesterday, there are a few unstable unit tests flipping on/off:
https://github.com/dmwm/WMCore/wiki/setup-wmcore-unittest#unstable-tests-in-py3

I will try to implement it in jenkins in the coming days.

PR #10562, opened 2021-06-01
[py2py3] fix unittests from  WMCore/Datastructs in py3

### LINKED ISSUE #10531: [py2py3] Check unittests from WMCore/Datastructs in py3
#### Intro

#### Details

This issue is about checking that the unittests from WMCore/Datastructs,
issue #10010 and PR #10012, do not fail.

```
test/python/WMCore_t/DataStructs_t/File_t.py
test/python/WMCore_t/DataStructs_t/Fileset_t.py
test/python/WMCore_t/DataStructs_t/Job_t.py
test/python/WMCore_t/DataStructs_t/JobGroup_t.py
test/python/WMCore_t/DataStructs_t/LumiList_t.py
test/python/WMCore_t/DataStructs_t/Mask_t.py
test/python/WMCore_t/DataStructs_t/MathStructs_t/ContinuousSummaryHistogram_t.py
test/python/WMCore_t/DataStructs_t/Run_t.py
test/python/WMCore_t/DataStructs_t/Subscription_t.py
test/python/WMCore_t/DataStructs_t/WorkUnit_t.py
test/python/WMCore_t/DataStructs_t/Workflow_t.py

```

mapellidario:
Fixes #10531 

#### Status
In development

#### Description

Changes
- [x] Again, another adjustment to how `WMCore.DataStructs.Run.Run`s are compared (i know that the motivations behind these changes may not be self-evident, I made a comment below with some investigations to justify these changes)
- [x] `pickle.dump()` and `pickle.load()` require the file object to be opened in `bytes` mode also in py3. (`JobPackage.py`)
- [x] minor changes to the unittests (usual `assertItemsEqual` -> `assertCountEqual`))

yes

#### Related PRs
This is a follow-up of #10012 

nope

### DIFF
--- a/src/python/WMCore/DataStructs/Run.py
+++ b/src/python/WMCore/DataStructs/Run.py
@@ -49,13 +49,24 @@ class Run(WMObject):
 
     def __lt__(self, rhs):
         """
-        Compare on run # first, then by lumis as a list is compared
+        Compare on run # first, then by lumis as a list, then by events in each lumi
         """
+        # check run number
         if self.run != rhs.run:
             return self.run < rhs.run
+        # if same run number, check list of lumi sections
         if sorted(self.eventsPerLumi.keys()) != sorted(rhs.eventsPerLumi.keys()):
             return sorted(self.eventsPerLumi.keys()) < sorted(rhs.eventsPerLumi.keys())
-        return self.eventsPerLumi < rhs.eventsPerLumi
+        # if same list of lumis, check events in each lumi section
+        for lumiNumber in sorted(self.eventsPerLumi):
+            if self.eventsPerLumi[lumiNumber] == rhs.eventsPerLumi.get(lumiNumber):
+                continue
+            else: 
+                return self.eventsPerLumi[lumiNumber] < rhs.eventsPerLumi.get(lumiNumber)
+        # if same run number, same lumi sections list, 
+        # same events in each lumi section:
+        # then the runs are equal and __lt__ should return false
+        return False
 
     def __le__(self, other):
         return self.__lt__(other) or self.__eq__(other)
--- a/src/python/WMCore/DataStructs/JobPackage.py
+++ b/src/python/WMCore/DataStructs/JobPackage.py
@@ -36,7 +36,7 @@ class JobPackage(WMObject, dict):
 
         Pickle this object and save it to disk.
         """
-        with open(fileName, 'w') as fileHandle:
+        with open(fileName, 'wb') as fileHandle:
             pickle.dump(self, fileHandle, -1)
         return
 
@@ -47,7 +47,7 @@ class JobPackage(WMObject, dict):
         Load a pickled JobPackage object.
         """
         self.clear()
-        with open(fileName, 'r') as fileHandle:
+        with open(fileName, 'rb') as fileHandle:
             loadedJobPackage = pickle.load(fileHandle)
         self.update(loadedJobPackage)
         return
--- a/test/python/WMCore_t/DataStructs_t/Run_t.py
+++ b/test/python/WMCore_t/DataStructs_t/Run_t.py
@@ -102,82 +102,74 @@ class RunTest(unittest.TestCase):
         comparision and sorting
 
         """
+        def cmp_runs(run_x, run_y):
+            return [
+                run_x < run_y, run_x <= run_y,
+                run_x == run_y, run_x != run_y,
+                run_x > run_y, run_x >= run_y,]
+
         run1 = Run(1, 1, 2, 3)
         run2 = Run(2, 4, 5, 6)
         run3 = Run(3, 7, 8, 9)
 
-        runlist = sorted([run2, run3, run1])
-        self.assertEqual(runlist[0], run1)
-        self.assertEqual(runlist[1], run2)
-        self.assertEqual(runlist[2], run3)
-
-        comparisons_0 = [
-            run1 < run2, run1 <= run2,
-            run1 == run2, run1 != run2,
-            run1 > run2, run1 >= run2,
-            ]
-        self.assertListEqual(comparisons_0,
-                             [True, True,
-                              False, True,
-                              False, False])
+        runlist = list(sorted([run2, run3, run1]))
+        self.assertListEqual(runlist, [run1, run2, run3])
+        self.assertListEqual(cmp_runs(run1, run2),
+                             [True, True, False, True, False, False])
 
         run4 = Run(4, 1, 2, 3)
         run5 = Run(4, 4, 5, 6)
         run6 = Run(4, 7, 8, 9)
 
-        runlist = sorted([run5, run6, run4])
-        self.assertEqual(runlist[0], run4)
-        self.assertEqual(runlist[1], run5)
-        self.assertEqual(runlist[2], run6)
+        runlist = list(sorted([run5, run6, run4]))
+        self.assertListEqual(runlist, [run4, run5, run6])
 
-        runlist = sorted([run5, run1, run3, run4, run6, run2])
-        self.assertEqual(runlist[0], run1)
-        self.assertEqual(runlist[1], run2)
-        self.assertEqual(runlist[2], run3)
-        self.assertEqual(runlist[3], run4)
-        self.assertEqual(runlist[4], run5)
-        self.assertEqual(runlist[5], run6)
+        runlist = list(sorted([run5, run1, run3, run4, run6, run2]))
+        self.assertListEqual(runlist, [run1, run2, run3, run4, run5, run6])
 
-        run7 = Run(9, 1, 2, 3)
-        run8 = Run(9, 1, 2, 3)
-        run9 = Run(9, 2)
-        run10 = Run(9, 3, 4, 5)
-        run11 = Run(9, 8, 9, 10)
+        run7  = Run(9, 1, 2)
+        run8  = Run(9, 1, 2, 3)
+        run9  = Run(9, 1, 2, 3)
+        run10 = Run(9, 1, 2, 3, 4)
+        run11 = Run(9, 1, 2, 4)
+        run12 = Run(9, 1, 4)
+        run13 = Run(9, 2)
+        run14 = Run(9, 3, 4, 5)
+        run15 = Run(9, 8, 9, 10)
 
-        runlist = sorted([run10, run8, run11, run7, run9])
-        self.assertEqual(runlist[0], run7)
-        self.assertEqual(runlist[1], run8)
-        self.assertEqual(runlist[2], run9)
-        self.assertEqual(runlist[3], run10)
-        self.assertEqual(runlist[4], run11)
+        runlist = list(sorted([run10, run15, run8,  run11, run13,
+                               run7,  run9,  run14, run12]))
+        self.assertListEqual(runlist,
+                            [run7, run8, run9, run10, run11,
+                             run12, run13, run14, run15])
 
-        comparisons_1 = [
-            run10 < run11, run10 <= run11,
-            run10 == run11, run10 != run11,
-            run10 > run11, run10 >= run11,
-            ]
-        self.assertListEqual(comparisons_1,
-                             [True, True,
-                              False, True,
-                              False, False])
-        comparisons_2_same = [
-            run7 < run7, run7 <= run7,
-            run7 == run7, run7 != run7,
-            run7 > run7, run7 >= run7,
-            ]
-        self.assertListEqual(comparisons_2_same,
-                             [False, True,
-                              True, False,
-                              False, True])
-        comparisons_3_same = [
-            run7 < run8, run7 <= run8,
-            run7 == run8, run7 != run8,
-            run7 > run8, run7 >= run8,
-            ]
-        self.assertListEqual(comparisons_3_same,
-                             [False, True,
-                              True, False,
-                              False, True])
+        self.assertListEqual(cmp_runs(run14, run15),
+                             [True, True, False, True, False, False])
+        self.assertListEqual(cmp_runs(run8, run8),
+                             [False, True, True, False, False, True])
+        self.assertListEqual(cmp_runs(run8, run9),
+                             [False, True, True, False, False, True])
+
+        run20 = Run(666, [(1, 11), (2, 22), (3, 33)])
+        run21 = Run(666, [(1, 11), (2, 22), (3, 33)])
+        run22 = Run(666, [(1, 11), (2, 22), (3, 32)])
+        run23 = Run(666, [(1, 11), (2, 22), (3, 34)])
+        run24 = Run(666, [(1, 11), (2, 21), (3, 33)])
+        run25 = Run(666, [(1, 11), (2, 21), (3, 34)])
+        run26 = Run(666, [(1, 11), (2, 23), (3, 32)])
+
+        self.assertListEqual(cmp_runs(run20, run21),
+                             [False, True, True, False, False, True])
+        self.assertListEqual(cmp_runs(run20, run22),
+                             [False, False, False, True, True, True])
+        self.assertListEqual(cmp_runs(run20, run23),
+                             [True, True, False, True, False, False])
+        self.assertListEqual(cmp_runs(run20, run24),
+                             [False, False, False, True, True, True])
+        self.assertListEqual(cmp_runs(run20, run25),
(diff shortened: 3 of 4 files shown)


mapellidario:
Making `Run.__lt__()` compatible with py3 required some exploration of how dictionary comparison works in py2 and replicating it explicitly in py3. 

### proposed changes

This is the implementation of Run sorting (`-`: py2-only, `+`: is the py3 compatible version prposed in this PR)

```diff
    def __lt__(self, rhs):
        """
        Compare on run # first, then by lumis as a list is compared
        """
        if self.run != rhs.run:
            return self.run < rhs.run
        if sorted(self.eventsPerLumi.keys()) != sorted(rhs.eventsPerLumi.keys()):
            return sorted(self.eventsPerLumi.keys()) < sorted(rhs.eventsPerLumi.keys())
-        return self.eventsPerLumi < rhs.eventsPerLumi # this line breaks in py3
+        for self_events, rhs_events in zip(self.eventsPerLumi.values(), rhs.eventsPerLumi.values()):
+            if self_events == rhs_events:
+               continue
+            else: 
+                return self_events < rhs_events
+        return False
```

### py2 dictionaries comparisons

I paste here my findings for future reference

```python
import sys
print sys.version_info # sys.version_info(major=2, minor=7, micro=16, releaselevel='final', serial=0)

d0 = {1: 11, 2: 22, 3: 33}

# equal
d1 = {1: 11, 2: 22, 3: 33}

# same keys, diff content
d2 =  {1: 11, 2: 22, 3: 32}
d3 =  {1: 11, 2: 22, 3: 34}
d4 = {1: 11, 2: 21, 3: 33}
d5 = {1: 11, 2: 21, 3: 34}
d6 = {1: 11, 2: 23, 3: 32}
d7 = {1: 11, 2: 23, 3: 34}

# one added key OR one missing key
d8 = {1: 11, 2: 22, 3: 33, 4: 44}
d9 = {1: 11, 2: 21, 3: 33, 4: 44}
d10 = {1: 11, 2: 23, 3: 33, 4: 44}
d11 = {1: 11, 2: 22}
d12 = {1: 11, 2: 21}
d13 = {1: 11, 2: 23}

# two missing keys AND one added key
d14 = {1: 11, 4: 22}
d15 = {1: 11, 4: 21}
d16 = {1: 11, 4: 23}

# one missing key AND one added key
d17 = {1: 11, 2: 22, 4: 32}
d18 = {1: 11, 2: 22, 4: 33}
d19 = {1: 11, 2: 22, 4: 44}
d20 = {1: 11, 2: 21, 4: 44}
d21 = {1: 11, 2: 23, 4: 44}

def cmp_py2(d1):
    cmppy2 = [d0 < d1, d0 <= d1, d0 == d1, d0 != d1, d0 > d1, d0 >= d1]
    return cmppy2 == cmp_key(d1), cmppy2, cmp_key(d1)

def cmp_key(d1):
    k0 = sorted(d0.keys())
    k1 = sorted(d1.keys())
    return [k0 < k1, k0 <= k1, k0 == k1, k0 != k1, k0 > k1, k0 >= k1]

print 
print cmp_py2(d1)   # (True, [False, True, True, False, False, True], [False, True, True, False, False, True])
print 
print cmp_py2(d2)   # (False, [False, False, False, True, True, True], [False, True, True, False, False, True])
print cmp_py2(d3)   # (False, [True, True, False, True, False, False], [False, True, True, False, False, True])
print cmp_py2(d4)   # (False, [False, False, False, True, True, True], [False, True, True, False, False, True])
print cmp_py2(d5)   # (False, [False, False, False, True, True, True], [False, True, True, False, False, True])
print cmp_py2(d6)   # (False, [True, True, False, True, False, False], [False, True, True, False, False, True])
print cmp_py2(d7)   # (False, [True, True, False, True, False, False], [False, True, True, False, False, True])
print 
print cmp_py2(d8)   # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])
print cmp_py2(d9)   # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])
print cmp_py2(d10)  # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])
print cmp_py2(d11)  # (True, [False, False, False, True, True, True], [False, False, False, True, True, True])
print cmp_py2(d12)  # (True, [False, False, False, True, True, True], [False, False, False, True, True, True])
print cmp_py2(d13)  # (True, [False, False, False, True, True, True], [False, False, False, True, True, True])
print  
print cmp_py2(d14)  # (False, [False, False, False, True, True, True], [True, True, False, True, False, False])
print cmp_py2(d15)  # (False, [False, False, False, True, True, True], [True, True, False, True, False, False])
print cmp_py2(d16)  # (False, [False, False, False, True, True, True], [True, True, False, True, False, False])
print  
print cmp_py2(d17)  # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])
print cmp_py2(d18)  # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])
print cmp_py2(d19)  # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])
print cmp_py2(d20)  # (False, [False, False, False, True, True, True], [True, True, False, True, False, False])
print cmp_py2(d21)  # (True, [True, True, False, True, False, False], [True, True, False, True, False, False])

```

#### from d2 to d7

froom `d2` to `d7` are what we should focus on, because in py2 we got to ` return self.eventsPerLumi < rhs.eventsPerLumi` if and only if the two runs have the same list of lumi-sections. This is what does not work in py3 and that we have to replicate in py3.

In particular, we should notice that `cmp_py2(d4)` gives the same result as `cmp_py2(d5)`, while `cmp_py2(d6)` gives the same result as `cmp_py2(d7)`. This means that py2 sorts the keys, then compares the values key by key, starting from the lower key.

This is tested in `Run_t.py:RunTest.testC` with runs from `run20` to `run26`.

#### from d8 to d21

I explored what happens to dictionary sorting in py2 when the dictionaries have different keys. Some weird stuff happen, for example with `d0` and `d16`.

However, if we consider only what happens when comparing the keys of the dictionaries, then nothing out of the ordinary comes up. This is tested in `Run_t.py:RunTest.testC` with runs from `run7` to `run15`.

vkuznet:
@mapellidario two issues here:
- performance impact when dicts are large, did you benchmark how much time such comparison will take if dicts has lots of keys and the size of dicts is large?
- another way to comparing dicts (or its parts) is to take a hash from dict representation (but it requires an ordered list of keys anyway). Here is again the performance matter.

mapellidario:
@vkuznet I did not make any benchmark for a couple of reasons.

#### 1: is this really a problem?

These are the changes to be applied [here](https://github.com/dmwm/WMCore/blob/2cdaccab4628c49e6769bc1064eb6712919b6cfb/src/python/WMCore/DataStructs/Run.py#L50).

```diff
      def __lt__(self, rhs):
          if self.run != rhs.run:
              return self.run < rhs.run
          if sorted(self.eventsPerLumi.keys()) != sorted(rhs.eventsPerLumi.keys()):
              return sorted(self.eventsPerLumi.keys()) < sorted(rhs.eventsPerLumi.keys())
-        return self.eventsPerLumi < rhs.eventsPerLumi
+        for (_, self_events), (_, rhs_events) in zip(sorted(viewitems(self.eventsPerLumi)), sorted(viewitems(rhs.eventsPerLumi))):
+            if self_events == rhs_events:
+                continue
+            else: 
+                return self_events < rhs_events
+        return False
```

What I am adding, is
- sorting two dictionaries by values. This means getting two lists of `(key, value)` tuples. Since in a python dictionary the keys are unique, we only access the first item of each tuple for sorting the list of items of the dictionaries. In any case, `value` is an int (if i am not mistaken), nothing huge that we move around.
- for every element, there is a O(1) operation

I do not consider this to be terrible, since a couple of lines above we already sort the keys of the dictionaries. Twice per each dictionary, without even caching the result. If it were paramount to keep the cpu time low, i would expect that we at least cache the list of sorted keys. 

In any case, if we hit the part that I added, we already sorted each dictionary twice, so in the worst case scenario it would be a +50% in cpu time, assuming that py2 implementation of `dict1 < dict2` is instantaneous (hardly to be the case). I can make a quick benchmark to make sure that my approach is not terribly slower than `dict1 < dict2` to improve this worst case scenario estimate.

However, how often does it happen that we compare two `Run`s with the same list of lumi sections? This is the real question that we should should ask ourselves and I feel like I am not in the best place to answer this question. Every synthetic benchmark that I can write will never be production-like.

#### 2: dmwm/WMCore priorities

Our priority is to make dmwm/WMCore work in py3, even if this comes with performance drawbacks. If you are worried about the performances, we can open an issue and have a second look at this very changes at a later stage.

mapellidario:
Valentin, you planted a seed in the back of my mind and by the end of the workday it grew into a full plant. I had to make a synthetic benchmark to measure this.

The benchmark is in my branch `10531-fix-benchmark`, which contains a couple of commits more than the branch used in this PR (with both what we have in master and my proposed changes, and with the benchmark itself): https://github.com/mapellidario/WMCore/compare/10531-fix...10531-fix-benchmark .

Comparing ten times two `Run`s with `300_000` lumi sections (each with a random number of events between 1 and 1e9) with the comparison we have now (`py2-compatible`) and my current approach (`py2py3-compatible-v0`) takes

| (results in seconds) | python2 | python3 |
|-|-|-|
| py2-compatible | 4.5 | `-` |
| py2py3-compatible-v0 | 20 | 10 |

You were right, my approach is significantly slower :grimacing: . I will go back to the drawing board and think about something new. 

Thanks for pointing this out!

vkuznet:
I doubt that you'll achieve decent performance with python itself, what you'll realize at the end is necessity to write C wrapper to battle such problems. I dealt with many dict issues many years ago when DAS was written in python. I even wrote few C-wrappers which you may find [here](https://github.com/dmwm/DAS/blob/master/src/python/DAS/utils/utils.py#L1227-L1240) and associated [C-wrapper](https://github.com/dmwm/DAS/blob/master/src/python/DAS/extensions/dict_handler.c). At the end, I gave up with python and non-structured (no schema) CMS data and switched to GoLang for everything. Since WMCore deals with many dicts and there are a lot of non-structured data (where quite often there is no schema in dicts since keys are not defined but dynamically assigned) you'll always struggle with this and other issues because of quite large size of dicts. And, the only solution for you would be to start learning and writing C-wrappers for Python functions (which by itself far from trivial).

My suggestion still stays, try to create a hash of the dict and compare the hashes. To speed things up you may look at DAS code which provides [genkey](https://github.com/dmwm/DAS/blob/master/src/python/DAS/utils/utils.py#L709-L730) and [das_hash](https://github.com/dmwm/DAS/blob/master/src/python/DAS/extensions/das_hash.c) C-wrapper. You may generate a hash of the dict using this method and compare the two.

mapellidario:
With the new approach [1] (let's call it `py2py3-compatible-v1`) I get

| (results in seconds, slower = better) | python2 | python3 |
|-|-|-|
| py2-compatible | 4.5 | `-` |
| py2py3-compatible-v1 | 8.5 | 3.7 |

This is in line with the +100% for the worst case scenario I had in mind earlier. ( The problem was that `sorted(dict.values())` is way slower than `sorted(dict.keys())`, shame on me that I assumed they were similar in our case. ). 

However, the performance drop appears in py2 only! The new approach in py3 is **faster** than the old approach in py2, so I would thank the python core developers and call this a win.

[1]

```python
    def __lt__(self, rhs):
        ....
        for self_key, rhs_key in zip(sorted(self.eventsPerLumi), sorted(rhs.eventsPerLumi)):
            if self.eventsPerLumi[self_key] == rhs.eventsPerLumi[rhs_key]:
                continue
            else: 
                return self.eventsPerLumi[self_key] < rhs.eventsPerLumi[rhs_key]
        ...
```

mapellidario:
> My suggestion still stays, try to create a hash of the dict and compare the hashes. To speed things up you may look at DAS code which provides genkey and das_hash C-wrapper. You may generate a hash of the dict using this method and compare the two.

Thank you Valentin for the suggestion! I do agree with everything you just said, however, with the time scale that we are dealing with at the moment, I think we just have to leave with this at the moment. I would leave to Kevin or Alan the decision of whether we should invest time in improving the performances of `Run.py`. Perfect topic for discussion for the monday meeting.

klannon:
@mapellidario, I would agree with both your focus on completing an implementation before optimizing, and also on your conclusion that since the Py3 version is comparable (even slightly better) than the existing Py2 version, it's good enough to move forward and **FINISH** the Py3 migration.  Once we complete the Py3 migration, we can add to the brainstorming list an analysis of code performance to identify in which components optimization the benefits of optimization would outweigh the costs.

test/python/WMCore_t/DataStructs_t/File_t.py
```diff
@@ -32,8 +34,8 @@ def testDefinition(self):
         self.assertEqual(testFile['size'], 0)
         self.assertEqual(testFile['events'], 0)
         self.assertEqual(testFile['checksums'], {})
-        self.assertItemsEqual(testFile['parents'], {})
-        self.assertItemsEqual(testFile['locations'], {})
+        self.assertCountEqual(testFile['parents'], {}) if PY3 else self.assertItemsEqual(testFile['parents'], {})
```
amaltaro, line 37: Please be honest ;) Would you like to update it to the override version (that you provided today in a different PR):
```
if PY3:
```
? You have already made all the hard changes, so I'm happy to deal with this in the future as well.

src/python/WMCore/DataStructs/Run.py
```diff
@@ -49,13 +49,24 @@ def __ne__(self, rhs):
...
         """
+        # check run number
         if self.run != rhs.run:
             return self.run < rhs.run
+        # if same run number, check list of lumi sections
         if sorted(self.eventsPerLumi.keys()) != sorted(rhs.eventsPerLumi.keys()):
             return sorted(self.eventsPerLumi.keys()) < sorted(rhs.eventsPerLumi.keys())
-        return self.eventsPerLumi < rhs.eventsPerLumi
+        # if same list of lumis, check events in each lumi section
+        for self_key, rhs_key in zip(sorted(self.eventsPerLumi), sorted(rhs.eventsPerLumi)):
```
amaltaro, line 61: In case you decide to update the assert check, and given the discussion raised in this PR, I might have a minor suggestion:
```
for lumiNumber in sorted(self.eventsPerLumi):
    if self.eventsPerLumi[lumiNumber] == rhs.eventsPerLumi.get(lumiNumber):
        continue
    else:
        return self.eventsPerLumi[lumiNumber] < rhs.eventsPerLumi.get(lumiNumber)
```

I do not know whether it works though. Neither if it would improve the performance :-D

PS.: keep in mind we use lower camelCase for variable names (even though you will find many modules with a mix of conventions).

amaltaro (approved):
Thanks for the very detailed description and (performance) checks, Dario. This code looks good to me and it's ready to go. But I left a couple of comments along the code in case you want to have one last look into it. 
I'm happy to merge it and move forward as well, just let me know.

test/python/WMCore_t/DataStructs_t/File_t.py
mapellidario: I do agree with you, I'd better fix this as you say, thanks Alan!

mapellidario: (pushed in the last commits, already squashed)

src/python/WMCore/DataStructs/Run.py
mapellidario: This is not a minor suggestion, this is a supernice suggestion! Pushed in the last commits, squashed into the previous ones

mapellidario:
I updated the code with @amaltaro's suggestions. I already squashed the commits so that this is ready to be merged (I promise I haven't added anything that wasn't' suggested/approved :smirk: )

With Alan's suggestion (`py2py3-compatible-v2`), we got a further boost

| (results in seconds, slower = better) | python2 | python3 |
|-|-|-|
| py2-compatible | 4.5 | `-` |
| py2py3-compatible-v2 | 6.0 | 2.7 |
| py2py3-compatible-v1 | 8.5 | 3.7 |
| py2py3-compatible-v0 | 20 | 10 |

Thanks @vkuznet for raising the concern in the first place and thanks Alan for providing the final implementation!

amaltaro:
Nicely done, Dario! Thanks

PR #10563, opened 2021-06-01
[py2py3] Adapt to renaming of assertItemsEqual to assertCountEqual in py3

### LINKED ISSUE #10514: [py2py3] assertItemsEqual is not supported in Python3
**Impact of the bug**
Probably unit tests only, but it could be masking other real problems

**Describe the bug**
I've seen a few python3 unit tests failing with this error:
```
AttributeError: 'MyTestCase' object has no attribute 'assertItemsEqual'
```

quick google suggests this attribute is no longer supported in python3 (or it might have been replaced by something else).
 
**How to reproduce it**
```
nosetests test/python/WMCore_t/Services_t/WMStats_t/DataStructs_t/RequestInfoCollection_t.py
```

**Expected behavior**
We should find a way to consistently validate the unit tests in both python2 and python3, so finding an equivalent replacement for `assertItemsEqual`. If we cannot use a single function/attribute common in both python versions, then we might have to find specific python solutions.

Full error log
```
ERROR: some very basic unit tests for the RequestInfo class
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/Services_t/WMStats_t/DataStructs_t/RequestInfoCollection_t.py", line 151, in testRequestInfo
    self.assertItemsEqual(reqInfoObj.getTasks(), {})
AttributeError: 'MyTestCase' object has no attribute 'assertItemsEqual'
```

mapellidario:
Fixes #10514 

#### Status
Ready

#### Description
`assertItemsEqual` has been renamed `assertCountEqual` in py3. Sometimes it's possible to use alternatives, such as `assertListEqual`, but that is not always the case. 

I changed my mind: there are too many occurrences of `assertItemsEqual`, it would take a superlong time to fix them all, we can just use

```python
# ../Module_t.py
class ModuleTest(unittest.TestCase):
    def setUp(self):
    ...
        if PY3:
```

and we should be good to go

yes

#### Related PRs
This changes have already been carried out in some other PRs related to #10422, but i do not have a list of them yet

nope

### DIFF
--- a/test/python/Utils_t/MemoryCache_t.py
+++ b/test/python/Utils_t/MemoryCache_t.py
@@ -16,53 +16,57 @@ class MemoryCacheTest(unittest.TestCase):
     unittest for MemoryCache functions
     """
 
+    def setUp(self):
+        if PY3:
+            self.assertItemsEqual = self.assertCountEqual
+
     def testBasics(self):
         cache = MemoryCache(1, [])
-        self.assertCountEqual(cache.getCache(), []) if PY3 else self.assertItemsEqual(cache.getCache(), [])
+        self.assertItemsEqual(cache.getCache(), [])
         cache.setCache(["item1", "item2"])
-        self.assertCountEqual(cache.getCache(), ["item1", "item2"]) if PY3 else self.assertItemsEqual(cache.getCache(), ["item1", "item2"])
+        self.assertItemsEqual(cache.getCache(), ["item1", "item2"])
         # wait for cache to expiry, wait for 2 secs
         sleep(2)
         self.assertRaises(MemoryCacheException, cache.getCache)
         cache.setCache(["item4"])
         # and the cache is alive again
-        self.assertCountEqual(cache.getCache(), ["item4"]) if PY3 else self.assertItemsEqual(cache.getCache(), ["item4"])
+        self.assertItemsEqual(cache.getCache(), ["item4"])
 
     def testCacheSet(self):
         cache = MemoryCache(2, set())
-        self.assertCountEqual(cache.getCache(), set()) if PY3 else self.assertItemsEqual(cache.getCache(), set())
+        self.assertItemsEqual(cache.getCache(), set())
         cache.setCache(set(["item1", "item2"]))
-        self.assertCountEqual(cache.getCache(), ["item1", "item2"]) if PY3 else self.assertItemsEqual(cache.getCache(), ["item1", "item2"])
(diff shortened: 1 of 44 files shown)


mapellidario:
@amaltaro I am not sure that you like this solution, so I ask for your opinion with a comment instead of using the "official" button "request a review". Any suggestion is very welcome!

amaltaro:
Honestly, when I read your question, the first thing that came to my mind was a method override, just like you did. It looks much cleaner than the previous approach - checking whether it's a PY3 test right before the assert. Yes, this is a neat solution IMO!

mapellidario:
Good, thanks! 

The other approach would be to find and replace `assertItemsEqual` to `assertCountEqual`, then use

```
# ../Module_t.py
class ModuleTest(unittest.TestCase):
    def setUp(self):
    ...
        if PY2:
            self.assertCountEqual = self.assertItemsEqual
``` 

This is the usual trade-off: more job now or more job for future-us

amaltaro:
Given the current circumstances, the approach you have implemented looks more appealing to me. 
If you are ready with these changes, feel free to squash them and fire a review request. Thanks

amaltaro (approved):
I believe these 4 python3 unit tests are actually unstable:
```
WMCore_t.Configuration_t.ConfigurationTest:testF changed from success to failure
WMCore_t.Services_t.Rucio_t.Rucio_t.RucioTest:testGetRule changed from error to failure
WMCore_t.WMBS_t.JobSplitting_t.EventAwareLumiByWork_t.EventAwareLumiByWorkTest:testNoFileSplitNoHardLimit changed from success to failure
WMCore_t.WMBS_t.JobSplitting_t.EventAwareLumiByWork_t.EventAwareLumiByWorkTest:test_NotEnoughEvents changed from error to failure
```

unless you know it for sure, I suggest we keep an eye on those in the coming days and properly flag them as unstable, if needed.

Code looks perfect to me! Thanks Dario.

mapellidario:
`WMCore_t.WMBS_t.JobSplitting_t.EventAwareLumiByWork_t.EventAwareLumiByWorkTest:testNoFileSplitNoHardLimit` should be in the list of unstable unittests, while the others changed from error to failure or vice versa. I have to admit that I am not paying too much attention to these cases

amaltaro:
Let's run a few tests during the day today and update the list of unstable py3 tests then. I will take care of it, but please provide me with more unstable test names once you spot them.

PR #10579, opened 2021-06-08
[py2py3] fix  unittests from Slice 3 in py3

### LINKED ISSUE #10534: [py2py3] Check unittests from Slice 3 in py3
#### Intro

#### Details

This issue is about checking that the unittests from Slice 3,
issue #10114 and PR #10117, do not fail.

```
test/python/WMCore_t/REST_t/Format_t.py
test/python/WMCore_t/Services_t/McM_t/McM_t.py
test/python/WMCore_t/REST_t/Main_t.py

```

Moreover
- [x] `test/python/WMCore_t/WebTools_t/RESTFormat_t.py`

mapellidario:
Fixes #10534

#### Status
Ready

#### Description

`src/python/WMCore/WebTools/Root.py`
- [x] changes configuration of cherrypy so that py2 and py3 interfaces are equivalent
  - source: [stackoverflow](https://stackoverflow.com/a/55039512/5313348)
  - some more info for future-us: `'tools.encode.encoding': 'utf-8',` may be necessary in the future if we have problems with encodings in cherrypy

`src/python/WMCore/REST/Test.py`
- [x] force `/dev/urandom` to be read as bytes in py3, too

yes

#### Related PRs
This is a followup to #10117 

nope

### DIFF
--- a/src/python/WMCore/REST/Test.py
+++ b/src/python/WMCore/REST/Test.py
@@ -61,7 +61,7 @@ def fake_authz_key_file(delete=True):
     :returns: Instance of :class:`~.NamedTemporaryFile`, whose *data*
       attribute contains the HMAC signing binary key."""
     t = NamedTemporaryFile(delete=delete)
-    with open("/dev/urandom") as fd:
+    with open("/dev/urandom", "rb") as fd:
         t.data = fd.read(20)
     t.write(t.data)
     t.seek(0)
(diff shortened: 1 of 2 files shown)


amaltaro (approved):
I don't quite understand the `text_only` option for the `ResponseEncoder` cherrypy object, and document is quite sparse as well.

But it fixes a few unit tests, so I'd suggest to get it merged and see how these things go once deployed (I might be wrong, but I think this Root module is actually only used by CRAB services).
Thanks, Dario.

PR #10581, opened 2021-06-08
[py2py3] Fix unittests from Slice 10 in py3

### LINKED ISSUE #10541: [py2py3] Check unittests from Slice 10 in py3
#### Intro

#### Details

This issue is about checking that the unittests from Slice 10,
issue #10134 and PR #10295, do not fail.

```
test/python/WMCore_t/WorkQueue_t/DataStructs_t/CouchWorkQueueElement_t.py
test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElement_t.py
test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElementsSummary_t.py

test/python/WMCore_t/Services_t/LogDB_t/LogDB_t.py

test/python/WMCore_t/ReqMgr_t/DataStructs_t/Request_t.py

```

mapellidario:
Fixes #10541 

#### Status
ready

#### Description

`src/python/WMCore/WorkQueue/DataStructs/CouchWorkQueueElement.py`
- [x] in `ele[key] > merged_value[key]` we have that `ele[key]` sometimes is `None`. In py2 `None` is always less than an int, but in py3 comparing None to a int gives a TypeError. 

`src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py`
- [x] convert to bytes the input of `md5.update()` in py3 only
  - the implementation may not be elegant, any suggestion is appreciated!
Mainly some changes about string handling in `src/python/WMCore/WorkQueue/DataStructs/`, 

`test/python/WMCore_t/WorkQueue_t/DataStructs_t/`
- [x] fixed a couple of unstable unittests

yes

#### Related PRs
This is a follow-up to #10295 

nope


src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
```diff
@@ -150,18 +152,25 @@ def id(self):
         # Assume md5 is good enough for now
         myhash = md5()
         spacer = ';'  # character not present in any field
-        myhash.update(self['RequestName'] + spacer)
```
amaltaro, line 153: A possible suggestion I would have would be to have a clear separation of how this `id` is calculated in python2 and in python3, something like:
```
if PY2:
    same code as before
else:
    new code
```

I believe it would be more clear, however it would have more lines of code and some (almost) code duplication.
So, in the end I think this approach looks better (but if you see more value on the approach I mentioned, go for it!).

test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElement_t.py
```diff
@@ -189,7 +190,10 @@ def testPossibleSites(self):
         """
         # test element ala MonteCarlo
         ele = WorkQueueElement(SiteWhitelist=["T1_IT_CNAF", "T2_DE_DESY"])
-        self.assertEqual(possibleSites(ele), ["T1_IT_CNAF", "T2_DE_DESY"])
+        if PY3:
```
amaltaro, line 193: We should not really waste time with these, but in the future, I would recommend the following the previous override approach:
```
if PY3:
```
and replace these `self.assertEqual` testing lists to `self.assertItemsEqual`.

amaltaro (approved):
It looks good to me, Dario. I made a couple of comments along the code, but nothing requiring further changes IMO. Could you please squash these commits accordingly? Thanks

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
mapellidario: Have a look at the new proposal, this should be a bit better, I hope

test/python/WMCore_t/WorkQueue_t/DataStructs_t/WorkQueueElement_t.py
mapellidario: > and replace these self.assertEqual testing lists to self.assertItemsEqual

yeah, I know. This is what I should have done, but i felt like that if in py2 we used `assertEqual` than we really wanted to check the order. If you say that we do not really care about the order, then I am happy of using `assertItemsEqual` in all those instances!

mapellidario: (changed!)

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
amaltaro: That's a very original idea! However, I'm not very comfortable with this function change and I think it can add some confusion as well, especially because those 2 methods have become very popular and largely used in our system (I believe not yet by external services).

Actually, why don't we get a third opinion here? I like it very much, but I also wanted to keep it as simple as possible, which might not be the case anymore (or it's all in my head ;)). @khurtado @todor-ivanov could one of you please share your opinion with us?

khurtado: I like the `encodeUnicodeToBytes` method! It looks very similar to the HTCondor `classad.quote` and `classad.unquote` methods to handle "condor strings" to python strings/bytes and the other way around. I would probably prefer to hide the `condition` parameter in the method and import `Utils.PythonVersion`  in `Utils.Utilities` and use that information directly for the the encodeUnicodeToBytes method, if possible.

mapellidario: At the beginning `encodeUnicodeToBytes` / `decodeBytesToUnicode` are conceived to give the same behavior in both py2 and py3 and we are already relying on this consistency, so i would not hide any check on PY2/PY3 inside them (examples: there are multiple occasions where we are using these function outside of `if PY3:`).

Here we want something different, we do not want consistency, we want to perform the encoding only with PY3 without being too-verbose. I understand that changing a function that is widespread may not be nice, what about a new couple of functions `encodeUnicodeToBytesConditional` / `decodeBytesToUnicodeConditional`, as just pushed?

Better name or different proposals are appreciated!

amaltaro: This looks better in my opinion. Can you please just add a docstring to those new function saying something like: "condition has to be an expression evaluating to True/False, usually PY2 or PY3 though"?

mapellidario: Sure! squashed and pushed.

amaltaro (approved):
It looks good to me, thanks Dario!

PR #10580, opened 2021-06-08
[py2py3] Check unittests from Slice 8 in py3

mapellidario:
Fixes #10539 

#### Status

There is a broken PR, but we decided to merge this anyways 

##### open questions

- [ ] open problem with `test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py:WorkQueueTest.testGlobalDatasetSplitting` in py2
  - caused by my new json file. Maybe I should ask Alan to provide me with a new one

##### changes

`DBSReader_t`
- [x] `nosetests test/python/WMCore_t/Services_t/DBS_t/DBSReader_t.py:DBSReaderTest.testListFilesInBlockWithParents` fails in py3, because in `mockData[self.url][signature]` looks for a `signature` that is the srting `listFileArray:[('detail', True), ('logical_file_name', ['/store/data/Commissioning2015/Cosmics/RAW/v1/000/238/545/00000/C47FDF25-2ECF-E411-A8E2-02163E011839.root']`, while in DBSMockData.json we have keys like `"listFileArray:[('detail', True), ('logical_file_name', [u'/store/data/Commissioning2015/Cosmics/RAW/v1/000/238/545/00000/C47FDF25-2ECF-E411-A8E2-02163E011839.root'])]":`. Notice that the only difference is that the file path is prepended with a `u''`
  - is this a problem of the DsbMockApi only or would this be the behaviour of DBS too? 

`src/python/WMCore/ProcessPool/ProcessPool.py`
- [x] fixed some problems with strings that caused ` nosetests test/python/WMCore_t/ProcessPool_t/ProcessPool_t.py:ProcessPoolTest.testA_ProcessPool` to fail immediatly
- [x]  understand why it fails with `ImportError: No module named builtins`

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
This is follow-up to #10289 

#### External dependencies / deployment changes
nope

PR #10585, opened 2021-06-09
Fixes for slice 13: file open mode; pickle operations; stdout/stderr

### LINKED ISSUE #10544: [py2py3] Check unittests from Slice 13 in py3
#### Intro

#### Details

This issue is about checking that the unittests from Slice 13,
issue #10137 and PR #10310, do not fail.

```
test/python/WMCore_t/WMSpec_t/Steps_t/StepFactory_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Template_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Builder_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Builders_t/StageOut_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Builders_t/CMSSW_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/URLFetcher_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/PileupFetcher_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/CMSSWFetcher_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/DeleteFiles_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/StageOut_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py

test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/LogArch_t.py
```

Make sure that the problem presented in #10569 is fixed, have a look at #10570 for suggestions in case you are stuck.

amaltaro:
Fixes #10544 

#### Status
ready

#### Description
From the original list of affected unit tests, these modules had some tests failing:
```
test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/PileupFetcher_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/CMSSWFetcher_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/DeleteFiles_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/StageOut_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py
test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/LogArch_t.py
```

First commit addresses all the problems encountered to get tests succeeding for: `test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/CMSSW_t.py`, in the least intrusive possible way (hopefully!). However, suggestions and/or a careful review will be very welcome.

Fixed `test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/PileupFetcher_t.py` by not converting unicode/text strings to bytes in DBS3Reader. Basically reverting a change that was made in this slice https://github.com/dmwm/WMCore/pull/10289 fixing possibly unicode issues (so, it could be that I'm making a big mistake here!).
It also required a new DBSMockData.json file, but it has just been merged via #10580 

Test for `test/python/WMCore_t/WMSpec_t/Steps_t/Fetchers_t/CMSSWFetcher_t.py` was actually broken for a longer time (also for python2). I have it fixed as well, and the source code was cleaned up a bit.

There are 3 tests failing for `test/python/WMCore_t/WMSpec_t/Steps_t/Executors_t/LogArch_t.py`, but they are actually marked as integration, so I'm not even trying to fix them now.

Last commit - tests still running - fixes unit tests that were actually marked as `integration` for `Executors_t/StageOut_t`, but I figured it would be good to have them testing our code. It required changes only to the test itself.

I still have to properly squash those commits, but I'd rather have a review first such that things can be shown in a more organized and grouped way.

yes

#### Related PRs
none

none


amaltaro:
test this please

amaltaro:
@mapellidario changes to the StageOut_t are being tested by jenkins right now, but I just wanted to give you this heads up such that we can try to get it merged before 4pm today. If we manage to, we can then get it into our next baseline tests to be created at 4. If large changes are required and we cannot make it, it's not a big deal!

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -42,7 +42,7 @@ def remapDBS3Keys(data, stringify=False, **others):
                'block_name': 'BlockName', 'lumi_section_num': 'LumiSectionNumber'}
 
     mapping.update(others)
-    formatFunc = lambda x: encodeUnicodeToBytes(x) if stringify else x
+    formatFunc = lambda x: decodeBytesToUnicode(x) if stringify else x
```
mapellidario, line 45: From a quick scan, this is what worries me the most. 

In any case, the solution in #10580 should work here. Before #10289 we converted unicode to bytes and we were using py2. I assumed this was the case, for py3 also, but i ignored our beloved "native string" approach. so we can use `encodeUnicodeToBytes(x, condition=PY2 and stringify)` here, so that we keep the same behavior in py2.

Forcing unicode in py2 also is not causing any unittest to fail, but feels risky to me. So, i propose we merge #10580, then we merge this PR and we should be good to go

mapellidario: since i am not sure yet about how to properly deal with a problem with workqueue in #10580, If you want to quickly merge this I would just replace this line with 

```diff
+from Utils.PythonVersion import PY2
...
- formatFunc = lambda x: decodeBytesToUnicode(x) if stringify else x
+ formatFunc = lambda x: encodeUnicodeToBytesConditional(x, condition=PY2 and stringify)
```

amaltaro: Okay, I have recreated the list of DBS mock data and just pushed it in (without updating this lambda function). Let's first see how it goes. Thanks for your comments, Dario.

amaltaro:
Reminder: rename these unit tests: WMCore_t.WMSpec_t.Steps_t.Executors_t.StageOut_t.otherStageOutTexst

@mapellidario now that we have your changes in, it's ready for another review (note that unit test results are still using the baseline results from an hour ago, that's why your tests from 10580 will show up here as well).

src/python/WMCore/Services/TagCollector/TagCollector.py
```diff
@@ -60,7 +63,7 @@ def _getResult(self, callname="", clearCache=False,
 
         # Note cFile is just the base name pattern, args 
         # are also considered for the end filename in the method below
-        f = self.refreshCache(cFile, callname, args, encoder=encoder,
+        f = self.refreshCache(cFile, callname, args, encoder=encoder, decoder=decodeBytesToUnicode,
```
mapellidario, line 66: Do we need this in py2 also? 
Because for py3 we should already have it decoded to unicode by https://github.com/dmwm/WMCore/blob/79868333b62c26252035bf2fb57fd480f6c6ec66/src/python/WMCore/Services/Requests.py#L308-L309 
Do you remember why you added this?

src/python/WMCore/WMRuntime/SandboxCreator.py
```diff
@@ -162,8 +165,12 @@ def makeSandbox(self, buildItHere, workload):
...
             # Add a dummy module for zipimport testing
             (handle, dummyModulePath) = tempfile.mkstemp()
-            os.write(handle, "#!/usr/bin/env python\n")
-            os.write(handle, "print('ZIPIMPORTTESTOK')\n")
+            if PY2:
+                os.write(handle, "#!/usr/bin/env python\n")
+                os.write(handle, "print('ZIPIMPORTTESTOK')\n")
+            else:
+                os.write(handle, b"#!/usr/bin/env python\n")
+                os.write(handle, b"print('ZIPIMPORTTESTOK')\n")
```
mapellidario, line 173: I do not remember where an when, but i have already spent time on `tempfile.mkstemp` and it's a pity that they force the input to be [bytes in py3](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp), too.

In any case, if you have a fetish for brevity, and since this is one of those occasions where we want to force the output to be bytes both in py2 and py3, we can opt for

```suggestion
            os.write(handle, decodeUnicodeToBytes("#!/usr/bin/env python\n"))
            os.write(handle, decodeUnicodeToBytes("print('ZIPIMPORTTESTOK')\n"))
```

But i understand that this is not necessary. as always the final about code style is up to you!

mapellidario (approved):
Wow! A ton of nice changes here! Thanks Alan!

I only made a couple of comments, but nothing critical!

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -10,7 +10,8 @@
 from builtins import object, str, bytes
 from future.utils import viewitems
 
-from Utils.Utilities import encodeUnicodeToBytesConditional, decodeBytesToUnicode
+from Utils.PythonVersion import PY2
```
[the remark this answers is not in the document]
amaltaro: I will remove it. It must be a problem from the PR rebase. Thanks for spotting it.

src/python/WMCore/Services/TagCollector/TagCollector.py
amaltaro: It was necessary to fix the `CMSSW_t` unit test. This is the change that made all those WMSpec_t/StdSpecs_t unit tests to work as well.

Hmm, looking at the code you pointed out, it should indeed be doing the same job. I might be wrong, but I believe the code is entering in this condition:
https://github.com/dmwm/WMCore/blob/79868333b62c26252035bf2fb57fd480f6c6ec66/src/python/WMCore/Services/Requests.py#L306

thus simply converting things to string. If I set `decoder=False` on this module, I think it would fix the problem, even though the result is exactly the same than passing an explicit decoder.

src/python/WMCore/WMRuntime/SandboxCreator.py
amaltaro: Applying, but the correct method name `encodeUnicodeToBytes` :-P

amaltaro:
Thank you very much for this review, Dario. I have just pushed another commit with some of the changes you suggested. If you are happy with those, I will squash everything and get it merged.

src/python/WMCore/WMRuntime/SandboxCreator.py
mapellidario: ahaha, you are correct, it did't sound right but i was too lazy to check :grimacing:

src/python/WMCore/Services/TagCollector/TagCollector.py
mapellidario: The more I think about this the more I like your approach and the more I want to remove these lines  https://github.com/dmwm/WMCore/blob/79868333b62c26252035bf2fb57fd480f6c6ec66/src/python/WMCore/Services/Requests.py#L308-L309, so i would stick with your option here.

If I have to guess, when running `CMSSW_t`, your result was `bytes` and `decoder` was set to `True` in some keyword argument, so your strings ended up being unicode strings prepended with a `b'content of the original bytes string'`. In these cases my `if PY3:  result = decodeBytesToUnicode(result)` does not do much, and we should specify the decode that we want case by case.

Let me see if i can why i added those lines and if we can get rid of them!

mapellidario: Ok, i added it in #10522, let's see what we can do

mapellidario: https://github.com/dmwm/WMCore/blob/79868333b62c26252035bf2fb57fd480f6c6ec66/test/python/WMCore_t/Services_t/Requests_t.py#L348-L353

This is why I added `if PY3: result = decodeBytesToUnicode(result)`: if we make requests with `decoder=False, in py2 we have a bytes string `"html"` and we look for it into the result of `makeRequest`, while in py3 `"html"` is unicode. In py3, we can 
1. convert the result from `makeRequest` to unicode by default even if no decode is set (which means adding `if PY3: result = decodeBytesToUnicode(result)`, as done in #10522 ), or 
2. we can never use `decoder=False`, or 
3. we always decode every time after calling `makeRequest` if necessary (maybe not the smartest option, since we already have the possibility of using the `decoder` argument).

I have no strong preferences, but maybe 2. with

```diff
- out = req.makeRequest('/', decoder=False)
+ out = req.makeRequest('/', decoder=lambda x: decodeBytesToUnicodeConditional(x, condition=PY3))
```

is what make more sense to me right now (i can change mind pretty quickly, though)

mapellidario (approved):
Looks perfect, thanks Alan!

amaltaro:
Thanks for the very useful hints and quick review, Dario.

src/python/WMCore/Services/TagCollector/TagCollector.py
mapellidario: I think that this may cause some problems. Currently, `self.decode(result)` is returning `result.__str__()`, so if `result` is an object (let's say a dictionary), we get its string representation. This should not happen to us in `Requests.makeRequest` (because for example we are not decoding from strings to objects the responses from `pycurl_manager`), but it's something that may happen in the future.
Moreover, i think that returning unicode in py2 may cause a lot of troubles. 

However, I like your idea of combining the `decoder` argument with `decodeBytesToUnicode`, so i propose

```python
elif decoder is not False:
    result = decodeBytesToUnicodeConditional(result, condition=PY3)
    result = self.decode(result)
```

In this way
- py2, decoder not False, result is bytes: it is not touched and we return bytes (which is `str` and wwe are happy)
- py2, decoder not False, result is object: `result = self.decode(result)` convert it to bytes
- py3, decoder not False, result is bytes: it is converted to unicode, untouched by `result = self.decode(result)`
- py3, decoder not false, result is object: untouched by `decodeBytesToUnicode`, converted to unicode by result = self.decode(result)

We have been seeing a lot of problems in py3 (for example i just noticed that `test/python/WMCore_t/Services_t/Requests_t.py:TestRequests.testSecureWithProxy_with_pycurl` is failing in py3) because when we have `decoder` not False, `result = self.decode(result)` is prepending it with the characters `b''` and returning a unicode string. My new approach should be more robust. I can create a new issue and provide this fix, so we see how it behaves against unittests

mapellidario: Just a reminder for the two of us: our last approach (for example the one in this commit https://github.com/dmwm/WMCore/pull/10588/commits/886209371d875a5f2f5f7c3fe5e654d639872fec ) will make all uses of `makeRequest(..., decoder=False)` fail in py3. it seems to me that we are using `decoder=False` only in some unittests, so we should not really be affected by this choice

PR #10590, opened 2021-06-11
[py2py3] Fix unittests from Slice 5 in py3

### LINKED ISSUE #10536: [py2py3] Check unittests from Slice 5 in py3
#### Intro

#### Details

This issue is about checking that the unittests from Slice 5,
issue #10116 and PR #10262, do not fail.

```
# broken
test/python/WMCore_t/Services_t/FWJRDB_t/FWJRDBAPI_t.py

#checked
test/python/WMCore_t/Services_t/UserFileCache_t/UserFileCache_t.py
test/python/WMCore_t/Storage_t/Execute_t.py
test/python/WMCore_t/Storage_t/StageOutImpl_t.py
test/python/WMComponent_t/DBS3Buffer_t/DBSBufferDataset_t.py
test/python/WMCore_t/WMBS_t/File_t.py
test/python/WMCore_t/WMBS_t/Fileset_t.py
test/python/WMCore_t/WMBS_t/Workflow_t.py
test/python/WMCore_t/WMBS_t/WorkUnit_t.py
test/python/WMCore_t/Database_t/CouchUtils_t.py
test/python/WMCore_t/WMLogging_t.py
test/python/WMCore_t/WMSpec_t/Persistency_t.py
test/python/WMCore_t/GroupUser_t/Interface_t.py
test/python/WMCore_t/GroupUser_t/Group_t.py
test/python/WMCore_t/GroupUser_t/User_t.py
test/python/WMCore_t/Services_t/CRIC_t/CRIC_t.py
test/python/WMCore_t/Storage_t/Registry_t.py
test/python/WMCore_t/Database_t/DBCore_t.py
test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
test/python/WMCore_t/Services_t/ReqMgr_t/ReqMgr_t.py

```

mapellidario:
Fixes #10536 

#### Status
ready - needs rebasing onto #10595 

##### open

- [x] rebasing

##### fixed

- [x] `src/python/WMCore/Services/UserFileCache/UserFileCache.py`: convert to bytes input of hashlib
- [x] proposal for more useful logs from `pycurl_manager` when connections fails. useful in case of 500 errors.
- [x] `test/python/WMCore_t/WMBS_t/File_t.py:FileTest.testGetParentLFNs`: `TypeError: '<' not supported between instances of 'File' and 'File'`
  - based on `__eq__()`, built `__lt__()` and the other ordering operators for `src/python/WMCore/WMBS/File.py.File.getParentsLFNs()`
- [x]  `test/python/WMCore_t/Services_t/ReqMgr_t/ReqMgr_t.py` fixed with some changes (among cherrypicking some from #10595)

    however, this uinittest hangs in both py2 and py3 at the line https://github.com/dmwm/WMCore/blob/6cab2cbec356c63f9c175ac21995dc199ea0ad5d/test/python/WMCore_t/Services_t/ReqMgr_t/ReqMgr_t.py#L139

- [x] add again change (see comments below)
    ```diff
    -    result = "".join(result)
    +    result = b"".join(result)
    ```

yes

#### Related PRs
follow up to #10262 

nope

### DIFF
--- a/src/python/WMCore/DataStructs/File.py
+++ b/src/python/WMCore/DataStructs/File.py
@@ -5,7 +5,7 @@ _File_
 Data object that contains details for a single file
 
 """
-from past.builtins import basestring
+from builtins import str, bytes
 __all__ = []
 
 from WMCore.DataStructs.Run import Run
@@ -16,6 +16,11 @@ class File(WMObject, dict):
     """
     _File_
     Data object that contains details for a single file
+
+    TODO 
+    - use the decorator `from functools import total_ordering` after
+      dropping support for python 2.6
+    - then, drop __ne__, __le__, __gt__, __ge__
     """
 
     def __init__(self, lfn="", size=0, events=0, checksums=None,
@@ -94,7 +99,7 @@ class File(WMObject, dict):
         eq = False
         if isinstance(rhs, type(self)):
             eq = self['lfn'] == rhs['lfn']
-        elif isinstance(rhs, basestring):
+        elif isinstance(rhs, (str, bytes)):
             eq = self['lfn'] == rhs
         return eq
 
@@ -104,6 +109,27 @@ class File(WMObject, dict):
     def __hash__(self):
         thisHash = self['lfn'].__hash__()
         return thisHash
+
+    def __lt__(self, rhs):
+        """
+        Sort files based on lexicographical ordering of the value connected
+        to the 'lfn' key
+        """
+        eq = False
+        if isinstance(rhs, type(self)):
+            eq = self['lfn'] < rhs['lfn']
+        elif isinstance(rhs, (str, bytes)):
+            eq = self['lfn'] < rhs
+        return eq
+
+    def __le__(self, other):
+        return self.__lt__(other) or self.__eq__(other)
+
+    def __gt__(self, other):
+        return not self.__le__(other)
+
+    def __ge__(self, other):
+        return not self.__lt__(other)
 
     def json(self, thunker=None):
         """
@@ -125,7 +151,7 @@ class File(WMObject, dict):
                     "parents": []}
 
         for parent in self["parents"]:
-            if isinstance(parent, basestring):
+            if isinstance(parent, (str, bytes)):
                 # Then for some reason, we're passing strings
                 # Done specifically for ErrorHandler
                 fileDict['parents'].append(parent)
--- a/src/python/WMCore/REST/Format.py
+++ b/src/python/WMCore/REST/Format.py
@@ -1,6 +1,8 @@
 from __future__ import print_function
 
 from builtins import str, bytes, object
+from Utils.PythonVersion import PY3
+from Utils.Utilities import encodeUnicodeToBytes
 from future.utils import viewitems
 
 import hashlib
@@ -435,7 +437,7 @@ class DigestETag(object):
     def update(self, val):
         """Process response data `val`."""
         if self.digest:
-            self.digest.update(val)
+            self.digest.update(encodeUnicodeToBytes(val))
 
     def value(self):
         """Return ETag header value for current input."""
@@ -476,14 +478,14 @@ def _stream_compress_deflate(reply, compress_level, max_chunk):
         pending.append(chunk)
         npending += len(chunk)
         if npending >= max_chunk:
-            part = z.compress("".join(pending)) + z.flush(zlib.Z_FULL_FLUSH)
+            part = z.compress(encodeUnicodeToBytes("".join(pending))) + z.flush(zlib.Z_FULL_FLUSH)
             pending = []
             npending = 0
             yield part
 
     # Crank the compressor one more time for remaining output.
     if npending:
-        yield z.compress("".join(pending)) + z.flush(zlib.Z_FINISH)
+        yield z.compress(encodeUnicodeToBytes("".join(pending))) + z.flush(zlib.Z_FINISH)
 
 # : Stream compression methods.
 _stream_compressor = {
@@ -612,6 +614,8 @@ def stream_maybe_etag(size_limit, etag, reply):
 
     # OK, respond with the buffered reply as a plain string.
     res.headers['Content-Length'] = size
-    result = "".join(result)
+    # TODO investigate why `result` is a list of bytes strings in py3
+    # The current solution seems to work in both py2 and py3
+    result = b"".join(result) if PY3 else "".join(result)
     assert len(result) == size
     return result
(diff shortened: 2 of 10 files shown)


src/python/WMCore/DataStructs/File.py
```diff
@@ -105,6 +110,27 @@ def __hash__(self):
         thisHash = self['lfn'].__hash__()
         return thisHash
 
+    def __lt__(self, rhs):
```
amaltaro, line 113: Nice!

src/python/WMCore/REST/Format.py
```diff
@@ -612,6 +613,6 @@ def stream_maybe_etag(size_limit, etag, reply):
 
     # OK, respond with the buffered reply as a plain string.
     res.headers['Content-Length'] = size
-    result = "".join(result)
+    result = b"".join(result)
```
amaltaro, line 616: Dario, just a question out of curiosity. Would you remember or still have the error that we were getting without this conversion to byte string?

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -291,8 +291,8 @@ def request(self, url, params, headers=None, verb='GET',
                 data = self.parse_body(bbuf.getvalue(), decode)
         else:
             data = bbuf.getvalue()
-            msg = 'url=%s, code=%s, reason=%s, headers=%s' \
-                  % (url, header.status, header.reason, header.header)
+            msg = 'url=%s, code=%s, reason=%s, headers=%s, result=%s' \
```
amaltaro, line 294: +1

src/python/WMCore/Storage/Execute.py
```diff
@@ -21,7 +22,10 @@ def runCommand(command):
 
     """
     # capture stdout and stderr from command
-    child = Popen(command, shell=True, bufsize=1, stdin=PIPE, close_fds=True)
+    if PY3:
+        child = Popen(command, shell=True, bufsize=1, stdin=PIPE, close_fds=True, encoding='utf8')
```
amaltaro, line 26: Just a note: pylint2 complains with:
```
E1123: Unexpected keyword argument 'encoding' in constructor call
```

To me, it's fine like this. But if you want to add a comment to at least one of these commands, that would be fine either.

amaltaro (approved):
These changes look pretty good to me, Dario.
I have made a few comments/questions along the code though, no mandatory changes I believe. Once it's been rebased, it should be ready to go. Thanks

src/python/WMCore/DataStructs/File.py
mapellidario: I have to admit that i had to speculate on how to sort these `DataStructs/Files`, but based on `__eq__()` checking only `'lfn'` seemded like a reasonable option

src/python/WMCore/Services/pycurl_manager.py
mapellidario: perfect! i had the fear that you did not want to print `data` because we risk to fill some logs in case of some weird errors, but i was stuck on a failing unittest for 1h before realizing that just printing that would have given me the stacktrace I needed.

src/python/WMCore/Storage/Execute.py
mapellidario: yeah, the python2 pylint does not skip the code under the `if PY3`. I will add a comment in the code, thanks for the suggestion!

src/python/WMCore/REST/Format.py
mapellidario: This is a great question. I have to admit I do not remember why I added this change (and i did not take note of this). I think it's just quicker if i remove this change and we see what jenkins has to say

src/python/WMCore/Services/pycurl_manager.py
amaltaro: If it fails, I believe amount of data will be fairly small. So, I think it's a good tradeoff. But definitely something to keep in mind in case things go wild :-D

src/python/WMCore/REST/Format.py
mapellidario: Without this change, the unittest `test/python/WMCore_t/ReqMgr_t/Service_t/Auxiliary_t.py:AuxiliaryTest.testInfo` (and also other from `Auxiliary_t.py`) fails with

```plaintext
cherrypy.error: INFO: [15/Jun/2021:16:25:23]  SERVER OTHER ERROR __builtins__.TypeError 7afe8c458b6fe6294785b341006200c8 (sequence item 0: expected str instance, bytes found)
cherrypy.error: INFO: [15/Jun/2021:16:25:23]    Traceback (most recent call last):
cherrypy.error: INFO: [15/Jun/2021:16:25:23]      File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/REST/Server.py", line 726, in default
cherrypy.error: INFO: [15/Jun/2021:16:25:23]        return self._call(RESTArgs(list(args), kwargs))
cherrypy.error: INFO: [15/Jun/2021:16:25:23]      File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/REST/Server.py", line 842, in _call
cherrypy.error: INFO: [15/Jun/2021:16:25:23]        return stream_maybe_etag(apiobj.get('etag_limit', self.etag_limit), etagger, reply)
cherrypy.error: INFO: [15/Jun/2021:16:25:23]      File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/REST/Format.py", line 616, in stream_maybe_etag
cherrypy.error: INFO: [15/Jun/2021:16:25:23]        result = "".join(result)
cherrypy.error: INFO: [15/Jun/2021:16:25:23]    TypeError: sequence item 0: expected str instance, bytes found
cherrypy.access: INFO: [15/Jun/2021:16:25:23] 028e0d827607 127.0.0.1 "GET /reqmgr2/data/info HTTP/1.1" 500 Internal Server Error [data: 198 in 2661 out 90484 us ] [auth: - "" "" ] [ref: "" "WMCore.Services.Requests/v002" ]
```

mapellidario: So, we can investigate why we have such bytes moving around, but i am not sure that we can simplify anything here and i would just keep this change for the time being. What do you think?

amaltaro: Yes, I agree. I would suggest though to put a FIXME or TODO comment saying that for some reason `result` is a byte string and it fails in python3, such that we can investigate it in the future and try to move to sequence/normal strings.
If we want to be extremely conservative, we could even do something like:
if PY3:
  then do the byte thing, otherwise don't do anything 
else:
  previous code.

Anything is good with me.

mapellidario: I added a TODO and an `if PY3` to that line.

(With the risk of repeating myself and stating obvious facts, in py2 `b"hello"` is the same as `"hello"`, so we added the `if PY3` for the sole purpose of easily getting back to this change)

amaltaro (approved):
Thanks for making these changes, Dario. Can you please rebase these commits accordingly?

amaltaro:
Thanks, Dario.

PR #10605, opened 2021-06-17
Replace built-in hash function by hashlib.sha1

### LINKED ISSUE #10160: [py2py3] investigate effect of `hash()` differences between py2 and py3
**Impact of the new feature**

WMCore in general, and perhaps other services depending on it.

In dmwm/WMCore we rely on the standard `hash()` function, but its behavior has changed in py3. While in python2 the `hash()` of an object is guaranteed to be the always the same (if an object does not change) even after the python interpreter itself is restarted, this is not the case anymore in python3. The python3 `hash()` doc states

> By default, the __hash__() values of str and bytes objects are “salted” with an unpredictable random value. Although they remain constant within an individual Python process, they are not predictable between repeated invocations of Python.

We need to investigate if this can cause some errors. For example, if we store on a database the hash of an object obtained with `hash()` to check if it is the same across different python interpreters, then we need to refactor our code.

Quick links to `hash()` documentation
- [py2 hash doc](https://docs.python.org/2/reference/datamodel.html#object.__hash__): 
- [py2 hash doc](https://docs.python.org/3/reference/datamodel.html#object.__hash__)

Quick links to the use of `hash()` in `dmwm/WMCore/*`

* [src/python/WMCore/DataStructs/Run.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMCore/DataStructs/Run.py#L135)
* [src/python/WMCore/DataStructs/WorkUnit.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMCore/DataStructs/WorkUnit.py#L75)
* [src/python/WMCore/Services/Service.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMCore/Services/Service.py#L107)
* [src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py#L38)
* [src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py#L106)

Same process comparison, should not cause any problem:
* [test/python/WMCore_t/DataStructs_t/WorkUnit_t.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/test/python/WMCore_t/DataStructs_t/WorkUnit_t.py#L99)

How to reproduce:

```bash
# should print always the same value
docker run python:2 python -c 'print(hash(frozenset({"hello": "world"}.items())))'

# should print a different value every time
docker run python:3 python -c 'print(hash(frozenset({"hello": "world"}.items())))'
```

Investigate usage of `hash()` in WMCore and establish if the hashes are used to compare objects outside the boundaries of a python interpreter, for example between two concurrent interpreters or saving a hash on a database and using it at a later time.

The solution would require to use consistent hashing functions, such as the ones provided by `hashlib`, as it is already done in the following examples

* [src/python/WMCore/Services/LogDB/LogDBBackend.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMCore/Services/LogDB/LogDBBackend.py#L24)
* [src/python/WMQuality/WebTools/RESTClientAPI.py](https://github.com/dmwm/WMCore/blob/2b1af3e52c033f101880459556a6bac624f6ff89/src/python/WMQuality/WebTools/RESTClientAPI.py#L96)

Reproduce: how to get consistent hash

```bash
# py2
docker run python:2 python -c 'import hashlib;print(hashlib.sha256(unicode({"hello": "world"}).encode("utf-8")).hexdigest())'
> 08b8bdfc96c88412e3e4c9ce471899037be4f44b4d737fbdb85407bb3e66a331
# py3
docker run python:3 python -c 'import hashlib;print(hashlib.sha256(bytes(str({"hello": "world"}), "utf-8")).hexdigest())'
> 08b8bdfc96c88412e3e4c9ce471899037be4f44b4d737fbdb85407bb3e66a331
```

**Additional context**

Original discussion in [PR 10012](https://github.com/dmwm/WMCore/pull/10012#issuecomment-731164254).

### LINKED ISSUE #10543: [py2py3] Check unittests from Slice 12 in py3
#### Intro

#### Details

This issue is about checking that the unittests from Slice 12,
issue #10136 and PR #10307, do not fail.

```
test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py
```

manually added, files have been renamed in #10357. there is no need to wait until slice 25 to fix these, though
```
test/python/WMCore_t/MicroService_t/MSOutput_t/MSOutputTemplate_t.py
test/python/WMCore_t/MicroService_t/MSRuleCleaner_t/MSRuleCleanerWflow_t.py
test/python/WMCore_t/MicroService_t/MSTransferor_t/Workflow_t.py
```

#### analysis of the failures

Only failures in py2
1. `test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py:WorkQueueTest.testWMBSInjectionStatus`
2.  `test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py:WorkQueueTest.testGlobalBlockSplitting`

broken py3: 

- [ ] `testWhiteList` - F - `line 566, in testWhiteList
    self.assertEqual(len(work), 17 + 19)
AssertionError: 28 != 36`
- [ ] `testProcessingWithContinuousSplitting` - F - `self.assertEqual(len(self.localQueue), 35)
AssertionError: 30 != 35`
- [ ] `test/python/WMCore_t/WorkQueue_t/WorkQueue_t.py:WorkQueueTest.testProcessing` : FAIL `line 500, in testProcessing
    self.assertEqual(processedBlocks, 17)
AssertionError: 12 != 17`

- [ ] `testBlackList` - `line 546, in testBlackList
    self.assertEqual(len(work), 17 + 19)
AssertionError: 33 != 36`

amaltaro:
Fixes #10543 
Fixes #10160

#### Status
ready

#### Description
This PR contains 3 different changes:
* `WorkQueueBackend`: given that we sort the workqueue elements being acquired - by creation time, then by priority - multiple times, I decided to put that in a separate function such that it's easier to test that functionality (and avoid surprises as we upgrade the python version)
* second change deals with the different behavior of the built-in `hash()` function between python2 and python3. Without investigating much, we decided to adopt `hashlib.sha1()` which is a reliable, reproducible in both python versions, and reasonably cheap (still more expensive than the previous hash function though).
* third change is touching the unit tests only

NOTE: the `MockPhEDExApi.py` module wasn't touched because it's supposed to be removed very soon (or is no longer used).

yes

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/DataStructs/Run.py
+++ b/src/python/WMCore/DataStructs/Run.py
@@ -7,9 +7,12 @@ container representing a run, and its constituent lumi sections and event counts
 """
 
 from __future__ import print_function
-
 from future.utils import viewitems, listitems
 
+import hashlib
+from builtins import str as newstr, bytes as newbytes
+from Utils.PythonVersion import PY3
+from Utils.Utilities import encodeUnicodeToBytesConditional, encodeUnicodeToBytes
 from WMCore.DataStructs.WMObject import WMObject
 
 
@@ -142,10 +145,25 @@ class Run(WMObject):
     def __hash__(self):
         """
         Calculate the value of the hash
+
+        NOTE: Python2 maxint is:
+        > python -c 'import sys; print(sys.maxint)'
+        9223372036854775807
+        so we cannot use the full range of the hexadecimal hash code because
+        it could cause an integer overflow. This is the maximum slice/value we
+        can safely use:
+        > int(15 * "f", base=16)
+        1152921504606846975
         """
-        value = self.run.__hash__()
-        value += hash(frozenset(listitems(self.eventsPerLumi)))  # Hash that represents the dictionary
-        return value
+        if isinstance(self.run, (newstr, newbytes)):
+            value = encodeUnicodeToBytesConditional(self.run, condition=PY3)
+        else:
+            value = encodeUnicodeToBytesConditional(str(self.run), condition=PY3)
+        hashValue = hashlib.sha1(value)
+        # Generate immutable sorted list of lumis
+        frozenEvents = str(sorted(listitems(self.eventsPerLumi), key=lambda x: x[0]))
+        hashValue.update(encodeUnicodeToBytes(frozenEvents))
+        return int(hashValue.hexdigest()[:15], 16)
 
     @property
     def lumis(self):
--- a/src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py
+++ b/src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py
@@ -6,10 +6,14 @@ Unpack the user tarball and put it's contents in the right place
 """
 from __future__ import print_function
 from future import standard_library
+
+from Utils.Utilities import encodeUnicodeToBytes
+
 standard_library.install_aliases()
 
 import logging
 import os
+import hashlib
 import re
 import shutil
 import subprocess
@@ -35,7 +39,8 @@ def setHttpProxy(url):
 
     proxyList = re.findall('\(proxyurl=([\w\d\.\-\:\/]+)\)', output)
     if 'loadbalance=proxies' in output:
-        proxy = proxyList[hash(url) % len(proxyList)]
+        urlHash = int(hashlib.sha1(encodeUnicodeToBytes(url)).hexdigest()[:15], 16)
+        proxy = proxyList[urlHash % len(proxyList)]
     else:
         proxy = proxyList[0]
     os.environ['http_proxy'] = proxy
--- a/src/python/WMQuality/Emulators/RucioClient/MockRucioApi.py
+++ b/src/python/WMQuality/Emulators/RucioClient/MockRucioApi.py
@@ -4,19 +4,19 @@
 Version of WMCore/Services/Rucio intended to be used with mock or unittest.mock
 """
 from __future__ import print_function, division
-from future.utils import listitems
 # from builtins import object # avoid importing this, it beraks things
 
 import json
 import logging
 import os
+import hashlib
 
 from WMCore.Services.DBS.DBS3Reader import DBS3Reader, DBSReaderError
 from WMCore.Services.Rucio.Rucio import WMRucioException, WMRucioDIDNotFoundException
 from WMCore.WMBase import getTestBase
 from WMQuality.Emulators.DataBlockGenerator.DataBlockGenerator import DataBlockGenerator
 
-from Utils.PythonVersion import PY2
+from Utils.PythonVersion import PY2, PY3
 from Utils.Utilities import encodeUnicodeToBytesConditional
 
 PROD_DBS = 'https://cmsweb-prod.cern.ch/dbs/prod/global/DBSReader'
@@ -58,9 +58,12 @@ class MockRucioApi(object):
         :return: a fake list of sites where the data is
         """
         logging.info("%s: Calling mock sitesByBlock", self.__class__.__name__)
-        if hash(block) % 3 == 0:
+        block = encodeUnicodeToBytesConditional(block, condition=PY3)
+        # this algorithm gives same results in both python versions
+        blockHash = int(hashlib.sha1(block).hexdigest()[:8], 16)
+        if blockHash % 3 == 0:
             sites = ['T2_XX_SiteA']
-        elif hash(block) % 3 == 1:
+        elif blockHash % 3 == 1:
             sites = ['T2_XX_SiteA', 'T2_XX_SiteB']
         else:
             sites = ['T2_XX_SiteA', 'T2_XX_SiteB', 'T2_XX_SiteC']
--- a/src/python/WMCore/DataStructs/WorkUnit.py
+++ b/src/python/WMCore/DataStructs/WorkUnit.py
@@ -11,9 +11,11 @@ from __future__ import absolute_import, division, print_function
 from future.utils import listitems
 
 import sys
+import hashlib
 import time
 from functools import total_ordering
 
+from Utils.Utilities import encodeUnicodeToBytes
 from WMCore.DataStructs.WMObject import WMObject
 
 
@@ -71,8 +73,16 @@ class WorkUnit(WMObject, dict):
         """
         Hash function for this dict.
         """
-
-        return hash(frozenset(listitems(self)))
+        # Generate an immutable sorted string representing this object
+        # NOTE: the run object needs to be hashed
+        immutableSelf = []
+        for keyName in sorted(self):
+            if keyName == "run_lumi":
+                immutableSelf.append((keyName, hash(self[keyName])))
+            else:
+                immutableSelf.append((keyName, self[keyName]))
+        hashValue = hashlib.sha1(encodeUnicodeToBytes(str(immutableSelf)))
+        return int(hashValue.hexdigest()[:15], 16)
 
     def json(self, thunker=None):
         """
--- a/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
+++ b/src/python/WMCore/WMSpec/Steps/Fetchers/PileupFetcher.py
@@ -9,11 +9,13 @@ from future.utils import viewitems
 
 import datetime
 import os
+import hashlib
 import shutil
 import time
 import logging
 from json import JSONEncoder
 import WMCore.WMSpec.WMStep as WMStep
+from Utils.Utilities import encodeUnicodeToBytes
 from WMCore.Services.DBS.DBSReader import DBSReader
 from WMCore.Services.Rucio.Rucio import Rucio
 from WMCore.WMSpec.Steps.Fetchers.FetcherInterface import FetcherInterface
@@ -91,10 +93,10 @@ class PileupFetcher(FetcherInterface):
         fileName = ""
         for pileupType in stepHelper.data.pileup.listSections_():
             datasets = getattr(getattr(stepHelper.data.pileup, pileupType), "dataset")
-            fileName += ("_").join(datasets)
+            fileName += "_".join(datasets)
         # TODO cache is not very effective if the dataset combination is different between workflow
-        # here is possibility of hash value collision
-        cacheFile = "%s/pileupconf-%s.json" % (self.cacheDirectory(), hash(fileName))
+        cacheHash = hashlib.sha1(encodeUnicodeToBytes(fileName)).hexdigest()
+        cacheFile = "%s/pileupconf-%s.json" % (self.cacheDirectory(), cacheHash)
         return cacheFile
 
     def _getStepFilePath(self, stepHelper):
--- a/src/python/WMCore/WorkQueue/WorkQueueBackend.py
+++ b/src/python/WMCore/WorkQueue/WorkQueueBackend.py
@@ -6,6 +6,7 @@ Interface to WorkQueue persistent storage
 """
 
 from builtins import object
+
 from future.utils import viewitems
 
 import json
@@ -34,6 +35,19 @@ def formatReply(answer, *items):
                 result.append(item)
                 break
     return result, errors
+
+
+def sortAvailableElements(elementsList):
+    """
+    Given a list of workqueue elements dictionary, this function will
+    sort them in place, first by their creation time; secondly by their
+    priority. In other words, higher priority and older requests will
+    be first in the list
+    :param elementsList: a list of elements dictionary
+    :return: nothing, list is updated in place
+    """
+    elementsList.sort(key=lambda element: element['CreationTime'])
+    elementsList.sort(key=lambda element: element['Priority'], reverse=True)
 
 
 class WorkQueueBackend(object):
@@ -360,8 +374,7 @@ class WorkQueueBackend(object):
         for item in result:
             element = CouchWorkQueueElement.fromDocument(self.db, item)
             sortedElements.append(element)
-        sortedElements.sort(key=lambda element: element['CreationTime'])
-        sortedElements.sort(key=lambda element: element['Priority'], reverse=True)
+        sortAvailableElements(sortedElements)
 
         for element in sortedElements:
             commonSites = possibleSites(element)
@@ -476,8 +489,7 @@ class WorkQueueBackend(object):
                     self.logger.info(msg, element['RequestName'], element._id)
                 else:
                     sortedElements.append(element)
-            sortedElements.sort(key=lambda element: element['CreationTime'])
-            sortedElements.sort(key=lambda element: element['Priority'], reverse=True)
+            sortAvailableElements(sortedElements)
 
             for element in sortedElements:
                 if numElems <= 0:
--- a/test/python/WMCore_t/DataStructs_t/Run_t.py
+++ b/test/python/WMCore_t/DataStructs_t/Run_t.py
@@ -200,6 +200,15 @@ class RunTest(unittest.TestCase):
         s.add(run9)
         s.add(run10)
         s.add(run11)
+        self.assertEqual(run7, run8)
+        self.assertNotEqual(run7, run9)
+        self.assertNotEqual(run7, run10)
+
+    def testNewHash(self):
+        run1 = Run(1, 1, 2, 3)
+        self.assertEqual(hash(run1), 79038151938585768)
+        run666 = Run(666, [(1, 111), (3, 33), (2, 22)])
+        self.assertEqual(hash(run666), 21721583371131476)
 
 
 if __name__ == '__main__':
--- a/test/python/WMCore_t/WorkQueue_t/Policy_t/Start_t/Block_t.py
+++ b/test/python/WMCore_t/WorkQueue_t/Policy_t/Start_t/Block_t.py
@@ -440,7 +440,7 @@ class BlockTestCase(EmulatedUnitTestCase):
             for unit in units:
                 pileupData = unit["PileupData"]
                 self.assertEqual(len(pileupData), 1)
-                self.assertItemsEqual(listvalues(pileupData)[0], ['T2_XX_SiteA', 'T2_XX_SiteB'])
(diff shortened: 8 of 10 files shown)


src/python/WMCore/DataStructs/Run.py
```diff
@@ -143,8 +146,14 @@ def __hash__(self):
...
-        value += hash(frozenset(listitems(self.eventsPerLumi)))  # Hash that represents the dictionary
+        if isinstance(self.run, (newstr, newbytes)):
+            value = encodeUnicodeToBytesConditional(self.run, condition=PY3)
+        else:
+            value = encodeUnicodeToBytesConditional(str(self.run), condition=PY3)
+        value = int(hashlib.sha1(value).hexdigest()[:8], 36)
+        # frozenset ensures the same order in python2 and python3, even though it's unordered(?)
+        frozenEvents = encodeUnicodeToBytesConditional(str(frozenset(listitems(self.eventsPerLumi))), condition=PY3)
+        value += int(hashlib.sha1(frozenEvents).hexdigest()[:8], 36)
         return value
```
amaltaro, line 157: All this complicated code to make sure that the hashid is the same between python2 and python3 (and different executions of the python interpreter).

In addition to that, we need to avoid two exceptions in Python3, they are:
* `TypeError: Unicode-objects must be encoded before hashing`
* `TypeError: object supporting the buffer API required`. A set/frozenset cannot be passed as input value.

There is still 2 problems with the unit tests, because sometimes run (or lumi) can be either an integer or a long object, thus changing the final outcome of this hash function. I still do not know how to gracefully deal with those, besides making sure that we are always dealing with python3 `int` data object.
This is one of the unit tests that fail with the current code changes:
```
WMComponent_t.DBS3Buffer_t.DBSBufferFile_t.DBSBufferFileTest:testAddRunSet changed from success to failure
```

@mapellidario I do not want you to investigate this issue, but in case you have any suggestion from the top of your mind, I'm all ears! ;) (including suggestions to make the current code better!)

mapellidario: Hello! I had some fun this morning. https://github.com/dmwm/WMCore/compare/93a110619...mapellidario:10605-suggestion (i left some commented lines in the code to ease reproducibility of the benchmarks)

Some considerations
- i think we were suing `frozenset` because a list is not hash-able because it's mutable, so you can not do `hash(mylist)`. What we did was converting the list into a frozen set (which is an immutable set), then take the hash of that.
- if we want to have the same hash between py2 and py3, then converting the dict `Run.eventsPerLumi` to a string seems like a reasonable option, but we need to make sure that the string representation is the same in py2 and py3. If we do, we fix `test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py:DBSBufferFileTest.testAddRunSet`
  1. we need to make sure that the pairs (lumisection, events) are sorted, and we can sort on the lumisection number
  2. we need to remove the `L` from the string in py2
- We can do a similar thing for `self.run`

There is a catch: our new approach that aims at being consistent among py2 and py3 takes about 5 times longer. 

Table: how long (in seconds) it takes to hash the run `run1 = Run(1, 1, 2, 3)` 100'000 times with old approach (`hash()`) and with our new approach (`hashlib.sha1`)
|        | hash() | hashlib.sha1() |
|-|-|-|
| py2 | 0.17    | 1.22      |
| py3 | 0.20    | 1.03     |

amaltaro: Thank you very much for this investigation and performance numbers, Dario.

I actually found a way not to return `long` types anymore from the MySQL database. The recommended way, by passing a new `conv=dict` argument when creating the database connection - via `connect` method, did not work. I did not manage to find a way to override the converters when using SQLAlchemy.

However, what I managed to test and got successful results was a patch to the MySQL converters module - in the `py2-mysqldb.spec` here:
https://github.com/farcepest/MySQLdb1/blob/MySQLdb-1.2.4b4/MySQLdb/converters.py#L127

with the following changes:
```
[dmwm@9a960e526008 ~]$ diff -rup converters.py converters.py.new 
--- converters.py	2021-06-18 15:50:20.548763000 +0200
+++ converters.py.new	2021-06-18 15:53:01.234707000 +0200
@@ -143,12 +143,12 @@ conversions = {
     set: Set2Str,
     FIELD_TYPE.TINY: int,
     FIELD_TYPE.SHORT: int,
-    FIELD_TYPE.LONG: long,
+    FIELD_TYPE.LONG: int,
     FIELD_TYPE.FLOAT: float,
     FIELD_TYPE.DOUBLE: float,
     FIELD_TYPE.DECIMAL: float,
     FIELD_TYPE.NEWDECIMAL: float,
-    FIELD_TYPE.LONGLONG: long,
+    FIELD_TYPE.LONGLONG: int,
     FIELD_TYPE.INT24: int,
     FIELD_TYPE.YEAR: int,
     FIELD_TYPE.SET: Str2Set,
```

This is what gives us the best performance, since we do not need to cast data that has already been casted by the database driver.

Python2 int is very large, and I do not think we would have issues with runs/lumis/events numbers too large (we limit them at different places).:
```
(PY2) alannewmacpro:cmsdist amaltaro$ python -c 'import sys; print(sys.maxint)'
9223372036854775807
```

It looks almost like an "extreme" solution to me :), but WMAgent (and perhaps T0) seem to be the only current clients of the MySQLdb py2 library. And, soon'ish we won't even be using it. Please let me know what your thoughts are.

mapellidario: That's great news! Every time we avoid playing with string I am happy.

In any case, I would explicitely use `sorted` instead of `frozenset`, since it's not clear to me how reporoducible `frozenset()` ordering is. If i have to guess, being a set, it' likely not reproducible, or is likely to change in the future.

And I would also add something similar to the unittest in my suggestion branch, to check that we are consistently getting the hash that we expect

amaltaro: I see. So we would sort the list/set and cast it to string I guess (making it immutable). Is that what you wanted to say?

Sure, please keep your unit test there and I will take it once I test a new build of py2-mysqldb.

mapellidario: Yes! That's correct, and I will try to articulate why.

If we use frozenset

```plaintext
frozenEvents = 
encodeUnicodeToBytesConditional( # if it is py3, convert the string to bytes
  str(                                             # convert the frozenset into its string representation
    frozenset(                                # convert the list into an immutable frozenset
      listitems(self.eventsPerLumi) # from the dict, return a list of tuples (lumisection, events)
    )
  )
, condition=PY3) 
```

With this, we are casting the data to a string, but since frozenset is a set and has no defined order, we are not guaranteed which tuple `(lumisection, events)` will come first. If we have a run `run1 = Run(1, 1, 2, 3)`, then it is possible that one time we get `[(1: None), (2: None), (3: None)]`, but we can get as well `[(3: None), (1: None), (2: None)]`. Using the frozenset and checking the hash, we always get different hashes for the same `run1` [1], indeed.

If we replace the `frozenset` with `sorted` we have reproducible results.

```plaintext
frozenEvents = 
encodeUnicodeToBytesConditional( # if it is py3, convert the string to bytes
  str(                                             # convert the frozenset into its string representation
    sorted(                                     # returns a sorted list
      listitems(self.eventsPerLumi) # from the dict, return a list of tuples (lumisection, events)
      , key=lambda x: x[0]               # sorts the list of tuples based on the first item, the lumisection
    )
  )
, condition=PY3) 
```

[1]: 

```plaintext

[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 969116890346 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(2, None), (3, None), (1, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 357614472732 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(3, None), (1, None), (2, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 969116890346 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(2, None), (3, None), (1, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 969116890346 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(2, None), (3, None), (1, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 969116890346 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(2, None), (3, None), (1, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 969116890346 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(2, None), (3, None), (1, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 357614472732 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(3, None), (1, None), (2, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)
[dmwm@028e0d827607 WMCore]$ n test/python/WMCore_t/DataStructs_t/Run_t.py:RunTest.testRunHash
F
FAIL: testRunHash (WMCore_t.DataStructs_t.Run_t.RunTest)
  File "/home/dmwm/wmcore_unittest/WMCore/test/python/WMCore_t/DataStructs_t/Run_t.py", line 207, in testRunHash
    self.assertEqual(hash(run1), 651310213912)
AssertionError: 969116890346 != 651310213912
-------------------- >> begin captured stdout << ---------------------
frozenset({(2, None), (3, None), (1, None)})

--------------------- >> end captured stdout << ----------------------

Ran 1 test in 0.001s

FAILED (failures=1)

```

amaltaro:
test this please

amaltaro:
Nice! The last unit test report (right above) has been executed with the patched mysqldb spec:
https://github.com/cms-sw/cmsdist/pull/7044

Will keep testing it a bit further.

src/python/WMCore/DataStructs/Run.py
amaltaro: After a tough debugging and being lost for a while, I figured that I am still getting wrong results here and there because I am misusing these functions. Here is an example:
```
In [4]: int("a2b"[:2], base=16)
Out[4]: 162
```

So, we cannot get the first 16 chars and expect to have an integer of length 16 (very very silly me!). I think I have found a way out, and I will make sure there is a clear comment in the code as well.

src/python/WMCore/DataStructs/Run.py
```diff
@@ -142,10 +145,25 @@ def __delitem__(self, key):
...
+        """
+        if isinstance(self.run, (newstr, newbytes)):
+            value = encodeUnicodeToBytesConditional(self.run, condition=PY3)
+        else:
+            value = encodeUnicodeToBytesConditional(str(self.run), condition=PY3)
+        hashValue = hashlib.sha1(value)
+        # Generate immutable sorted list of lumis
+        frozenEvents = str(sorted(listitems(self.eventsPerLumi), key=lambda x: x[0]))
+        hashValue.update(encodeUnicodeToBytes(frozenEvents))
+        return int(hashValue.hexdigest()[:15], 16)
```
mapellidario, line 166: I like this! I have to say that `base=36` seemed a bit arbitrary, but having using `hexdigest()` and converting to an int in base 16 seems much more reasonable to me!

Moreover, thanks for investigating the limits of the int in py2!

src/python/WMQuality/Emulators/RucioClient/MockRucioApi.py
```diff
@@ -58,9 +58,12 @@ def sitesByBlock(self, block):
         :return: a fake list of sites where the data is
         """
         logging.info("%s: Calling mock sitesByBlock", self.__class__.__name__)
-        if hash(block) % 3 == 0:
+        block = encodeUnicodeToBytesConditional(block, condition=PY3)
+        # this algorithm gives same results in both python versions
+        blockHash = int(hashlib.sha1(block).hexdigest()[:8], 16)
```
amaltaro, line 63: This one I'm planning to leave with 8 chars slice only because otherwise there will be a bunch of tests that I will have to hunt down. Anyways, it's only in the mock module as well :-)

src/python/WMCore/WMRuntime/Scripts/UnpackUserTarball.py
```diff
@@ -35,7 +39,9 @@ def setHttpProxy(url):
 
     proxyList = re.findall('\(proxyurl=([\w\d\.\-\:\/]+)\)', output)
     if 'loadbalance=proxies' in output:
-        proxy = proxyList[hash(url) % len(proxyList)]
+        # NOTE: hash function behaves differently between Py2 and Py3, it might be a problem here!
```
amaltaro, line 42: Self note, when squashing these commits, this comment should be removed.

mapellidario (approved):
It looks very good! thanks Alan for the prompt changes and thorough investigation!

src/python/WMQuality/Emulators/RucioClient/MockRucioApi.py
mapellidario: I was about to ask you about this, and your motivation is sound! 
if the unittests pass and have no problem, then there is no need to change this and change all the unittests again

src/python/WMCore/DataStructs/Run.py
amaltaro: Yes, I somehow thought it would result in an smaller integer value 🤦

amaltaro:
Thank you for your review, Dario. I'm going to squash these commits properly and get it merged.

src/python/WMCore/DataStructs/Run.py
mapellidario: (comment for future me, i know you discovered this already:) ) `int()` always give back the base 10 integer, so if you provide a string and specify it is in base 36, then it will give you back a larger integer

```
# python3
>>> x = 13245907823409857290879087
>>> int(str(x),10)
13245907823409857290879087
>>> int(str(x),36)
876957221859903199400449539217648595047
```

PR #10610, opened 2021-06-21
Make output module ordering consistent between Py2/Py3

amaltaro:
Fixes #10553 

#### Status
ready

#### Description
This PR fixes all the `StdSpecs_t` unit tests failures caused by slice 22.

First commit orders a few list objects used during the workflow creation, such that we can predict the order of workflow/subscription/fileset to be created by the agent.

Second commit fixes all the StdSpecs_t specs unit tests that were failing (those checking for the whole map of workflow+subscription+fileset+id)

Third commit removes a couple of unit tests from the unstable list.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #10612, opened 2021-06-22
Create workload sandbox efficiently

### LINKED ISSUE #10611: Do not attempt to create the Workload sandbox multiple times
**Impact of the new feature**
WMAgent

While fixing unit tests involving WMBSHelper module, I noticed that the workflow sandbox creation is called many times for the same workflow (same amount of subscriptions in the workflow), here is where we attempt to create it:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/WMBSHelper.py#L278

It will, of course, only go through that function completely in the first call. Subsequent calls will get out of the function as soon as they have verified that the sandbox/workload.pkl file already exists on disk. 

Avoid calling `createSandbox()` unnecessarily.
 
Things work well as they are, so we can decide to spend those extra CPU cycles and keep things as they are.

**Additional context**
none

amaltaro:
Fixes #10611 

#### Status
ready

#### Description
This will ensure that we call the sandbox creation, for a given workflow, only once. Thus saving some CPU cycles and a couple of syscalls.

yes

#### Related PRs
Changes already in in https://github.com/dmwm/WMCore/pull/10610 , but I'd like to get this one merged first.

none

### DIFF
--- a/src/python/WMCore/WorkQueue/WMBSHelper.py
+++ b/src/python/WMCore/WorkQueue/WMBSHelper.py
@@ -267,15 +267,18 @@ class WMBSHelper(WMConnectionBase):
 
         return
 
-    def _createSubscriptionsInWMBS(self, task, fileset, alternativeFilesetClose=False):
+    def _createSubscriptionsInWMBS(self, task, fileset, alternativeFilesetClose=False,
+                                   createSandbox=True):
         """
         __createSubscriptionsInWMBS_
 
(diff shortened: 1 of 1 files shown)


amaltaro:
Todor, feel free to leave your review and/or concerns. I'm merging it because the other PR #10610 needs it.

todor-ivanov (commented):
Looks good to me

PR #10619, opened 2021-06-23
Dump pickle with py2-comatible protocol - WMCore-wide

### LINKED ISSUE #10609: WMAgent fails to load spec pickle file created in Python3
**Impact of the bug**
ReqMgr2/WorkQueue and WMAgent

**Describe the bug**
Workflows created in Python3 WMCore central services version `1.5.0.pre3`, running on Python2 WMAgent 1.4.7.patch3, fail in the agent's WorkQueueManager because the agent isn't able to load the pickle file retrieved from ReqMgr2 (actually it comes from the workqueue database, but it should be exactly the same!). Error message can be seen in [1]

**How to reproduce it**
See scenario and versions in the comment above

**Expected behavior**
We need to adopt a common pickle protocol that will work and provide the same behaviour in both python2 and python3 version. Which means, WorkQueueManager should not fail to load the workflow spec file when a workflow is being acquired.

NOTE: even though the problem happens in WMAgent, we should make sure this issue is addressed before Python3 central services are deployed. Such that we can avoid spec creation incompatibilities.

Related to: https://github.com/dmwm/WMCore/issues/9128

[1]
```
2021-06-21 22:44:45,141:140647810840320:INFO:WorkQueueManagerWorkPoller:Obtained 3 unit(s) of work
2021-06-21 22:44:45,250:140647810840320:INFO:WorkQueueManagerWorkPoller:Splitting new work
2021-06-21 22:44:45,250:140647810840320:INFO:WorkQueue:fixing conflict...
2021-06-21 22:44:45,340:140647810840320:ERROR:WorkQueueManagerWorkPoller:Error during split: unsupported pickle protocol: 4
  File "/data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMComponent/WorkQueueManager/WorkQueueManagerWorkPoller.py", line 114, in processWork
    self.queue.processInboundWork()
  File "/data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMCore/WorkQueue/WorkQueue.py", line 1138, in processInboundWork
    inbound_work = self.backend.getElementsForSplitting()
  File "/data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMCore/WorkQueue/WorkQueueBackend.py", line 111, in getElementsForSplitting
    wmspec.load(self.parentCouchUrlWithAuth + "/%s/spec" % ele['RequestName'])
  File "/data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMCore/WMSpec/Persistency.py", line 78, in load
    self.data = pickle.loads(data[0])
ValueError: unsupported pickle protocol: 4
2021-06-21 22:44:45,341:140647810840320:INFO:WorkQueueManagerWorkPoller:Splitting finished
2021-06-21 22:44:45,341:140647810840320:INFO:BaseWorkerThread:WorkQueueManagerWorkPoller took 0.501 secs to execute
```

mapellidario:
Fixes #10609

#### Status
Stopped development here, kept as a reference

#### Description

As described in [python doc](https://docs.python.org/3/library/pickle.html#data-stream-format), we can select the pickle protocol version when we dump it. In this case, we can dump pickle with protocol version `1` (EDIT: I later discovered that the py2 default is protocol 0) which is the default in our py2 environment, so that we can load it from py2 also. Quick reference of pickle protocols at [1]

More thorough investigation of pickle protocol compatibility is available at this branch https://github.com/mapellidario/WMCore/compare/f7739cb..10609-investigation (compared against the commit that was on dmwm/WMCore/master at the time of creating the branch). The result of the investigation is
- from py2 to py3: if you have bytes strings (which are the py2 default strings) containing non-ascii characters of if you have some bytes sequence that does not represent a string (such as a pickle file itself, an image, whatever), then you may need to juggle a bit with string encodings OR with keys retrieval when loading from py3
- from py3 to py2: when you unpickle an object, make sure that it belongs to a class that inherits from `object` (or from python-future's `from builtins import object`, otherwise pickle.load fails when instantiating the object since it cannot call `object.__new__(cls)`

We should not be affected by these problems, so with our current knowledge, I suggest to ignore these suggestions. The investigation has been proved useful to pin down potential pitfalls and to find solutions to those, so that we have a clear path to follow in case of unexpected issues arising from our pickle validation.

My proposed solution is quite brutal: use pickle protocol 1 (we should use 0 instead, that is the py2 default) in all WMCore, apart from in `JobCreatorPoller.py` where we were asking for the highest protocol available (and in py2 that is protocol `2`). I did not spend time investigating where we dump pickle that can be read from py2 and where we dump pickle from py3 that will only be read from py3 (where we could benefit using even higher protocols, like the py3 default).

#### validation

- [x] unittest
- [ ] Try and replicate #10609 

With respect to how we run in py2, I selected the highest protocol available in py2 (protocol 2 is the highet for both pickle and cPickle in py2), so we should not see any degradation. 

However, it is not clear to me what is the effect of dumping and loading pickles in py3 with protocol 2.

#### Related PRs
no PRs related, that i recall

nope

#### Further information

[1]
From the link to python pickle documentation,

-    Protocol version 0 is the original “human-readable” protocol and is backwards compatible with earlier versions of Python.

-    Protocol version 1 is an old binary format which is also compatible with earlier versions of Python.

-    Protocol version 2 was introduced in Python 2.3. It provides much more efficient pickling of new-style classes. Refer to PEP 307 for information about improvements brought by protocol 2.
-    Protocol version 3 was added in Python 3.0. It has explicit support for bytes objects and cannot be unpickled by Python 2.x. This was the default protocol in Python 3.0–3.7.

-    Protocol version 4 was added in Python 3.4. It adds support for very large objects, pickling more kinds of objects, and some data format optimizations. It is the default protocol starting with Python 3.8. Refer to PEP 3154 for information about improvements brought by protocol 4.

-    Protocol version 5 was added in Python 3.8. It adds support for out-of-band data and speedup for in-band data. Refer to PEP 574 for information about improvements brought by protocol 5.


mapellidario:
The unittest `test/python/WMCore_t/FwkJobReport_t/ReportEmu_t.py:ReportEmuTest.testMerge` fails during `setUp`. In particular `configuration.py.pythonise_()` is having troubles reading `self.pickledarguments`, likely set in `src/python/WMCore/WMSpec/Steps/Templates/CMSSW.py`: 

```plaintext
('DM debug - attr', 'pickledarguments', '\x80\x02}q\x00.')
```

mapellidario:
Another reminder that the only way to be sure about something is to try it.

#### problem in py2 when using protocol=2 everywhere

`test/python/WMCore_t/FwkJobReport_t/ReportEmu_t.py:ReportEmuTest.testMerge` failed during `setUp()`. The problem arises from this line

https://github.com/dmwm/WMCore/blob/f7739cba3f487056f8d82819eeba98c5b145130f/test/python/WMCore_t/FwkJobReport_t/ReportEmu_t.py#L39

that called `src/python/WMCore/WMSpec/Steps/Templates/CMSSW.py.CMSSWStepHelper.setDataProcessingConfig()`, which contains

https://github.com/dmwm/WMCore/blob/f7739cba3f487056f8d82819eeba98c5b145130f/src/python/WMCore/WMSpec/Steps/Templates/CMSSW.py#L159



But later we call 

https://github.com/dmwm/WMCore/blob/f7739cba3f487056f8d82819eeba98c5b145130f/test/python/WMCore_t/FwkJobReport_t/ReportEmu_t.py#L40

which relies on `src/python/WMCore/Configuration.py.ConfigSection.pythonise()`, which contains

https://github.com/dmwm/WMCore/blob/f7739cba3f487056f8d82819eeba98c5b145130f/src/python/WMCore/Configuration.py#L251-L254

where we call `formatAsString(getattr(self, 'pickledargument'))`, which means `formatAsAtring(pickle.dumps(args, protocol=<1 or 2>))`.

In python2, 
- `myName` can be a unicode string, 
- `pickle.dumps(args)` is the same as `pickle.dumps(args, protocol=1)`, which returns a bytes string containing ascii characters
- `pickle.dumps(args, protocol=2)` returns a bytes sequence, but this is binary data that do not correspond to unicode codepoints. This can not be decoded to unicode
- since `pickle.dumps()` returns a bytes stream, `formatAsString()` simply prepend and append a `'`.

When py2 encounters `"%s.%s = %s" % ...`, it converts automatically all the bytes sequences to unicode strings assuming that they contain unicode-encoded characters, but this is not the case for `formatAsAtring(pickle.dumps(args, protocol=2))` and it fails with a unicode decode error. If we force protocol1, then we have no problems.

#### Workaround

The safest approach would be continue using protocol 1, not protocol 2. Same as current behavior in py2, should be safe.

amaltaro:
Dario, sorry for giving you this tough issue to be investigated, the outcome is great though!

I'm not sure I fully understood your last comment, so here is my question.
*) Are we changing the default Python2 pickle protocol? Can you please remind me which one is that?

Once you are somehow confident with these changes, I can test it in my private setup where:
* pickle file is created in python3
* and read in python2 (I could test reading it in python3 as well, with some extra effort)

mapellidario:
Well, in my py2 docker `pickle.dumps(..., protocol=1)` gives the same as `pickle.dumps(...)`, this is why I say that i am "not changing py2 behavior".

However, the variable ['pickle.DEFAULT_PROTOCOL'](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL) is available from py3 only, so I have to use the conditional "should be safe". 

I am not 100% sure that we are using protocol=1 in production, but it's veeery likley

vkuznet:
The usual way to solve such kind of problems is to write wrapper around I/O for different protocol versions. Such wrapper can deal with different protocol versions and delivery forward compatible one for on-going workflows. In other words, it can support of reading protocols v1, v2, etc. and provide whatever version to py2/py3 stack (v1/v2). It allows the abstract of protocols and will provide common set of APIs to WMCore. Of course, it will require to change usage of python pickle module and replace it with custom class/wrapper.

amaltaro:
Dario, I managed to test the current changes provided in this PR, and it works fine with:
* central services in python3 (1.5.0.pre3) + this patch
* WMAgent in python2 (1.4.7.patch3) + this patch manually applied

which is actually great, it means reqmgr2 + globalworkqueue + mstransferor + msmonitor are already functional in python3 (!)

I still have to educate myself on the pickle (+protocol) implementation in Py2 and Py3, but maybe creating our own wrapper will be the way forward indeed, as Valentin suggests.

I will ping you on slack tomorrow morning, and maybe we can have a chat on this.

mapellidario:
After a chat with Alan, we convened that he opens a new PR where he leaves the default protocol everywhere apart from a couple of critical cases, namely:

- Persistency
- step/template/cmssw.py

In this way we use and old protocol only where it is absolutely needed.

We are keeping this issue open as a reference, but we do not think that next I will work on this PR in the near future

amaltaro:
For the record, or in case it's needed again soon, here:
https://amaltaro.web.cern.ch/amaltaro/forWMCore/Issue_10609/

we have a pickle dump of 4 workflows (almost equal, besides timestamp and request name) using Python3 and 4 different protocols.

amaltaro:
@mapellidario Dario, would you remember why we left it open? Maybe because first we wanted to check the latest pickle changes with a fully functional Py3 WMAgent instance? I couldn't see anything explicit in this thread.

mapellidario:
If I remember correclty, we had a voice chat a monday morning before the WMCore meeting where we discussed this topic. You used this PR to spot all the uses of `pickle.dump`, as a guideline for #10628. We decided to leave it open a little longer in case some unexpected problem with the pickle dump protocol version emerged even after your PR. Since "no news is a good news", i guess that we did not spot such problem and we can close this. 

Did I miss anything? (in case, i leave to you the honor of closing this)

amaltaro:
Thanks for the recap, Dario.
Let me try to get some Python3 tests done this week, if I don't manage to, then we close it by the end of this week.

amaltaro:
Note: if we really stick to python3.6, for `slc6_*` workflows, as discussed with David and Shahzad and proposed in this PR:
https://github.com/dmwm/WMCore/pull/10726

We will then have to force python3 to adopt protocol=4 instead of 5, while python2 agent could use the highest protocol (2).

From the official pickle documentation:
"""
Protocol version 4 was added in Python 3.4. It adds support for very large objects, pickling more kinds of objects, and some data format optimizations. It is the default protocol starting with Python 3.8. Refer to PEP 3154 for information about improvements brought by protocol 4.

Protocol version 5 was added in Python 3.8. It adds support for out-of-band data and speedup for in-band data. Refer to PEP 574 for information about improvements brought by protocol 5.
"""

amaltaro:
As I keep testing Py3 WMAgent, I get more convinced that your initial approach was the way to go :-D

amaltaro:
This PR still has some very important information, but given that there has been a couple of PRs superseeding this, and that all the services are now running in python3, I would suggest to close it. 
@mapellidario please do re-open it if you consider something is still left to be done with this. Thanks

mapellidario:
Nono, I wanted to mention you could close this as well, but you have been faster than me! :)

Issue #10410, opened 2021-03-31
MSUnmerged: create get methods for fetching unmerged files from the CC tool


amaltaro:
**Impact of the new feature**
ReqMgr2MS

**Is your feature request related to a problem? Please describe.**
It concerns a new microservice to carry out further cleanup of the sites' unmerged area (removal of files that are no longer needed/locked in the system).

**Describe the solution you'd like**
The expected solution from this issue is to implement the MSUnmerged logic for:
* retrieving a summary for all RSEs from the Consistency Check tool. That summary contains the number of unmerged files, empty directories, and others. The URL to be used is: https://cmsweb.cern.ch/rucioconmon/WM/stats
* retrieve a gzip file with a list of all unmerged files in a given RSE. AFAIK it contains one LFN entry per line. The URL to be used is (replacing the RSE name accordingly): https://cmsweb.cern.ch/rucioconmon/WM/files?rse=T2_US_UCSD
* last but not least, also add the necessary functionality to retrieve a list of all active output LFNs from WMStats (note that it will include the merged and unmerged LFNs, so we should filter it out and keep only the relevant information in memory). Reference for this API can be found in: https://github.com/dmwm/WMCore/wiki/wmstatsserver-api#retrieving-a-list-of-protected-lfns

**Describe alternatives you've considered**
none

**Additional context**
Depends on https://github.com/dmwm/WMCore/issues/10402

Some more info from an already closed issue that would be needed here:
https://github.com/dmwm/WMCore/issues/10411#issuecomment-845809843
 
I've been provided with the following link, with regards to how unmerged cleanup is currently performed by Dynamo:
http://t3serv001.mit.edu/~bmaier/ddm/consistency_config.json

especial attention to: IgnoreDirectories and `ListDeletable, expanding here just in case that link goes away:
```
  "IgnoreDirectories": [
    "SAM",
    "HCTest",
    "HCtest",
    "GenericTTbar",
    "EWKWPlus2Jets_WToLNu_M-50_14TeV-madgraph-pythia8",
    "VBFToHHTo4B_SM_14TeV-madgraph-pythia8",
    "WToLNu_1J_14TeV-madgraphMLM-pythia8",
    "WZTo3LNu_1Jets_14TeV-madgraphMLM-pythia8",
    "TTGamma_SingleLeptFromTbar_TuneCUETP8M1_14TeV-madgraph-pythia8",
[...]

PR #10621, opened 2021-06-24
Building the MSUnmerged construction around functional pipelines and configurable RSE lists.

### LINKED ISSUE #10599: MSUnmerged: Create scalable configuration K8 deployment
**Impact of the new feature**
MSUnmerged

In the initial implementation we are going to create `MSUnmerged` service stateless - no background database, no bookkeeping ... etc. In order to be able to scale it as simple as possible e want to have a small function called "GranularityParser" (or similar) to run first in every pod, which should read per pod limits (amount of work to be done by a single pod) from a common MSUnmerged configuration file. This piece of code should decide on which subset of Tiers and Sites(RSEs) the current pod is about to run, without overlapping work with other pods [see the attached file].

There are two possible solutions:

a) Create some communication channel between the pods - either by a shared filesystem or a bookkeeping system with a background database.

b) In case we have a unique identifier per pod (which does not change upon restart) - we may put a hardwired configuration for all the pods directly into the config telling them - pod1 is going to work on T1s, pod2 on T2s(subset 1), pod3 on T2s(subset 2) etc. .. But for that we would need to have that  unique and non volatile identifier per pod in advance.

So our preference is to go with option b) for the time being.

The method suggested by Valentin is to develop `MSUnmerged` in such a way that it would be able to start with the proper set of input parameters, e.g as `--tiers` && `--rseSubset`. And then we should create a dedicated `*.yaml` [1] file per every flavor (combination of `--tiers` && `--rseSubset`) of pods we want to have, so that we may run the full set of pods needed to serve the full list of RSEs we have in the system.
WMCore_MSUnmerged
 
[1]
https://github.com/dmwm/CMSKubernetes/blob/d4e4f2d2a5aada091ed7399a4ab1c0b277b7eab8/kubernetes/cmsweb/services/reqmgr2ms-unmerged.yaml

**Additional context**

todor-ivanov:
Fixes #10599 

#### Status
ready

#### Description
In order to achieve the scalability goals we have put for the service we need to split the initial list of RSEs to work on by Pod. So we  decided to have the that as a per K8 pod configurable `rseExpression` which is supposed to produce the desired subset of RSEs when executed in a call to Rucio. For this `rseExpression` we have prepared a placeholder in the MSConfig, which is supposed to be substituted with the `rseExppression` provided by the .yaml file for the pod at run time right before service execution. 

This PRalso provides  the basic logic for running the load through a functional pipeline, while accumulating counters and providing summary at the end of the service polling cycle     

YES

#### Related PRs
https://github.com/dmwm/CMSKubernetes/pull/612
https://github.com/dmwm/deployment/pull/1059/commits
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/89
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/90

We depend on the rest of the construction of the K8 pods and their setup.

### DIFF
--- a/src/python/WMCore/MicroService/MSManager.py
+++ b/src/python/WMCore/MicroService/MSManager.py
@@ -50,11 +50,11 @@ def daemon(func, reqStatus, interval, logger):
         sleep(interval)
 
 
-def daemonOpt(func, interval, logger):
+def daemonOpt(func, interval, logger, *args, **kwargs):
     "Daemon to perform given function action for all request in our store"
     while True:
         try:
-            func()
+            func(*args, **kwargs)
         except Exception as exc:
             logger.exception("MS daemon error: %s", str(exc))
         sleep(interval)
@@ -251,7 +251,7 @@ class MSManager(object):
         self.logger.info("Total ruleCleaner execution time: %d secs", res['execution_time'])
         self.statusRuleCleaner = res
 
-    def unmerged(self):
+    def unmerged(self, *args, **kwargs):
         """
         MSManager unmerged function.
         It cleans the Unmerged area of the CMS LFN Namespace
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
@@ -11,11 +11,15 @@ RSE basis.
 # futures
 from __future__ import division, print_function
 
+from pprint import pformat
+
 # WMCore modules
 from WMCore.MicroService.DataStructs.DefaultStructs import UNMERGED_REPORT
 from WMCore.MicroService.MSCore import MSCore
+from WMCore.MicroService.MSUnmerged.MSUnmergedRSE import MSUnmergedRSE
 # from WMCore.Services.AlertManager.AlertManagerAPI import AlertManagerAPI
 from WMCore.WMException import WMException
+from Utils.Pipeline import Pipeline, Functor
 
 
 class MSUnmergedException(WMException):
@@ -42,12 +46,21 @@ class MSUnmerged(MSCore):
 
         self.msConfig.setdefault("verbose", True)
         self.msConfig.setdefault("interval", 60)
-        self.msConfig.setdefault('limitRSEsPerInstance', 100)
-        self.msConfig.setdefault('limitTiersPerInstance', ['T1', 'T2', 'T3'])
-        self.msConfig.setdefault("rucioAccount", "FIXME_RUCIO_ACCT")
-        self.alertServiceName = "ms-unmerged"
+        # self.msConfig.setdefault('limitRSEsPerInstance', 100)
+        # self.msConfig.setdefault('limitTiersPerInstance', ['T1', 'T2', 'T3'])
+        # self.msConfig.setdefault("rucioAccount", "FIXME_RUCIO_ACCT")
+        self.msConfig.setdefault("rseExpr", "*")
         # TODO: Add 'alertManagerUrl' to msConfig'
+        # self.alertServiceName = "ms-unmerged"
         # self.alertManagerAPI = AlertManagerAPI(self.msConfig.get("alertManagerUrl", None), logger=logger)
+
+        # Building all the Pipelines:
+        pName = 'plineUnmerged'
+        self.plineUnmerged = Pipeline(name=pName,
+                                      funcLine=[Functor(self.cleanFiles)])
+        # Initialization of the deleted files counters:
+        self.rseCounters = {}
+        self.plineCounters = {}
 
     def execute(self):
         """
@@ -57,6 +70,8 @@ class MSUnmerged(MSCore):
         # start threads in MSManager which should call this method
         summary = dict(UNMERGED_REPORT)
 
+        self.resetCounters()
+
         try:
             rseList = self.getRSEList()
             self.updateReportDict(summary, "total_num_rses", len(rseList))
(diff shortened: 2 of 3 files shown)


todor-ivanov:
@amaltaro there is only two PRs that  need to be added in the deployment and service_config repositories so that we can put the `RSEEXPR` variable in the `config-unmerged.py` file, but this one needs to wait for merging 4 previous commits related to adding the `config-unmerged.py` itself to the repositories. So I would say you may start your review and once those are there I'll put the link in the comment at the proper section.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -57,27 +70,144 @@ def execute(self):
...
+            #         self.logger.debug(msg, fileUnmerged, rse['name'], str(ex))
+            rse['isClean'] = self._checkClean(rse)
+        except Exception as ex:
+            msg = "Error while cleaning RSE: %s"
+            msg += "Will retry in the next cycle. Err: %s"
+            self.logger.debug(msg, rse['name'], str(ex))
+
+        return rse
+
+    def _checkClean(self, rse):
```
amaltaro, line 179: What are your plans if an RSE is marked as not cleaned? My point is, perhaps we could have a different evaluation here to avoid trying to delete 10k files when only a handful of them were missed.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -57,27 +70,144 @@ def execute(self):
...
+                totalNumFiles += rse['counters']['totalNumFiles']
+                numFilesDeleted += rse['counters']['numFilesDeleted']
+            except Exception as ex:
+                msg = "%s: General error from pipeline. RSE: %s. Error:  \n%s. "
+                msg += "\nWill retry again in the next cycle."
+                self.logger.exception(msg, self.plineUnmerged.name, rse['rse'], str(ex))
+                continue
+        return totalNumRses, totalNumFiles, numRsesCleaned, numFilesDeleted
+
+    def filterRseList(self, rseList):
```
amaltaro, line 145: Doesn't the method `getRSEList` does all the job already? If not, how are you considering to use this one?

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
```diff
@@ -0,0 +1,49 @@
...
+        "dirs":  {"allUnmerged": [],
+                  "nonEmpty": [],
+                  "empty": []}
+        }
+        """
+        myDoc = {
+            "name": rseName,
+            "delInterface": "",
+            "isClean": False,
+            "counters": {"totalNumFiles" : 0,
```
amaltaro, line 37: I like this schema!
Just one question on the counters/totalNumFiles. Is it meant to be total number of files eligible for deletion? Or total number of files in the unmerged area (meaning, not necessarily eligible for deletion)?

amaltaro (commented):
These changes are looking good, Todor. I have left some suggestions/questions in the code for your consideration though.

src/python/WMCore/MicroService/MSManager.py
```diff
@@ -50,11 +50,11 @@ def daemon(func, reqStatus, interval, logger):
         sleep(interval)
 
 
-def daemonOpt(func, interval, logger):
+def daemonOpt(func, interval, logger, *args, **kwargs):
     "Daemon to perform given function action for all request in our store"
     while True:
         try:
-            func()
+            func(*args, **kwargs)
```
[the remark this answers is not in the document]
todor-ivanov: Not exactly needed, but this is the general function signature how it should look like. And this is something I regret not doing it in the commit introducing that function, so I decided to push it now.

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
```diff
@@ -0,0 +1,49 @@
...
+    A minimal RSE information representation to serve the needs
+    of the MSUnmerged Micro Service.
+    """
+    def __init__(self, rseName, **kwargs):
+        super(MSUnmergedRSE, self).__init__(**kwargs)
+
+        """
+        Document Schema:
+        {
+        "name": "RSEName",
```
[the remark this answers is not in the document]
todor-ivanov: If you think the doc schema is too much I would just remove the whole docstring. It serves no purpose for a standard `__init__` method.

src/python/WMCore/MicroService/MSManager.py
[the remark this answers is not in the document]
todor-ivanov: sorry we are talking about the function call here ... but still the same the generalization statement applies equally strong to both  definition and call.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: I was thinking about applying the the rest of the load constraints constraints here, those like `self.msConfig.limitRSEsPerInstance` etc. We can skip it if we want, or preserve  the placeholder for whatever constrain comes next.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Actually I plan to do exactly nothing, since  we are going to stumble on this RSE on a next cycle only once the results of the current iteration is reflected everywhere down the chain. So the next time we'll have to deal with this same RSE we won't iterate through the already deleted files. 

It may happen that we delete all the files put for deletion for the current cycle, but in case the RSE was  too polluted and number of files have exceded the limit of deletions per RSE fora  a single cycle the ... well... this is definitely an RSE we may not consider clean and will have to iterate through it again start digesting what ever was an excess from the previous run. 

Or an equivalent scenario - we have some broken files to the level no deletion can happen even though they are present in the namespace. Then ... whell this is something that would need a manual intervention and untill that   happens the RSE shouldn't be considered clean either. And we need to know somehow about this case. I've not yet figured it out how to identify and signal it properly though. Any suggestions are welcome.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -57,27 +70,144 @@ def execute(self):
...
+        for pline in self.plineCounters:
+            self.plineCounters[pline.name]['deletedSuccess'] = 0
+            self.plineCounters[pline.name]['deletedFail'] = 0
+
     def getRSEList(self):
         """
-        Queries CRIC for the proper RSE lists to iterate through.
-        return: List of RSEs.
+        Queries Rucio for the proper RSE list to iterate through.
+        :return: List of MSUnmergedRSE objects.
```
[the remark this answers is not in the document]
todor-ivanov: Which is exactly correct. 
This was  a docstring reflecting the return from this method  in a previous version when I was building the list of `MSUnmergedRSE` objects directly here. but I left that behind - because it means that   once we  start accumulating  all those huge lists of deletions and information in the objects we will have to keep them until the exit. I'd rather prefer to generate a simple list of identifiers (RSE names) here and pass it up and generate those objects right before I through them in the pipeline - this way, once the `MSUnmergedRSE` object goes through, the only leftover should be the counters.... 

Thanks for catching this docstring inconsistency. I will fix it.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -57,27 +70,144 @@ def execute(self):
...
         """
-        # NOTE: To filter out the granularity (Tier*) level.
         rseList = []
-
+        try:
+            for rse in self.rucio.evaluateRSEExpression(self.msConfig['rseExpression']):
```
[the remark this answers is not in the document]
todor-ivanov: that is correct too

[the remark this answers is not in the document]
todor-ivanov: The explanation why I was iterating through this list like that  was given above. I may fix it even though for the sake of staying  on the safe side the fix should include a `deep copy`.. which is going to do exactly the same as this cycle is doing.

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
todor-ivanov: Good question. This is a sign that the name chosen was not expressive enough, so it does not contain the needed amount of information in it. I should be total files eligible for deletion. It is used in the `_checkClean` method it is also returned to the report document (the string there tells it) ... but ye the name is maybe not the best. - I was following the general login in the report document and didn't think that would lead to the current confusion.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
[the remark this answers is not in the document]
todor-ivanov: Done. Works just fine without `deepcopy` too.

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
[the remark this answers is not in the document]
todor-ivanov: done

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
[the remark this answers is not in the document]
todor-ivanov: Fixed

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -57,27 +70,144 @@ def execute(self):
...
+        :return:    Bool: True if all files found have been deleted, False otherwise
+        """
+        return rse['counters']['totalNumFiles'] == rse['counters']['numFilesDeleted']
+
+    def resetCounters(self):
+        """
+        A simple function for zeroing the deleted files counters.
+        """
+
+        for rse in self.rseCounters:
```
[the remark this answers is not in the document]
todor-ivanov: Not exactly... this function was intended to not only clears the counters, but it should also initializes them if not existent. As you can see the very top in the `__init__` those counters are empty dictionaries. But the place this method is called is not the best we do not have the `rseList` yet ... etc. We need to have a quick discussion offline .. on the strategy how to use those, because they are actually duplicates of the already existing counters which are passed between function calls, but I like keeping them in the object more BTW.

todor-ivanov:
Thanks  for the review @amaltaro! I just fixed the things that  were an immediate change. but for some of the rest I may need your input, because there are several paths that   may be followed there. Please read my replies to the comments inline. Thanks again.

amaltaro:
@todor-ivanov can you please clean the pep8 warning reported above?

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
amaltaro: I think I would rely on the expression only. If a POD is defined to work in a set of RSEs, we better make sure that very same set is cleaned up by the same POD.
Regarding placeholders for a _possible_ development, you know what my opinion is on that. Just ditch it, please ;)

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
amaltaro: My suggestion is to actually define a fraction where an RSE is considered clean. Like, if >75% (or 90%?) of the deletable files managed to be deleted, then we should consider that RSE clean, otherwise we will simply add extra load on our microservice and on the RSE storage.

I fear that the microservice will spend a long time trying to delete some final problematic files, and this is something that we should avoid IMO.

To be explicit, what I'm trying to avoid here is to iterate over a list of 100k files, in a given RSE, when you only need to delete 500 of them. I would say, in such cases, it would be better to wait for a new dump of the consistency check tool.

Just some thoughts. The True/False might not be enough for us.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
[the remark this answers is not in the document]
amaltaro: Oh, I see. You're correct!

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
amaltaro: As we discussed offline, this is meant to be the total amount of deletable files (files that pass the WMStats filter). I gave you a suggestion this afternoon, but you could add a comment above and it should be enough as well.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: thanks @amaltaro 

> To be explicit, what I'm trying to avoid here is to iterate over a list of 100k files, in a given RSE, when you only need to delete 500 of them. 

As I mentioned in my previous reply, this effect should be mitigated by the `limitFilesPerRSE` configuration parameter. 

> I would say, in such cases, it would be better to wait for a new dump of the consistency check tool.

And again my point exactly - when I said:  

> we are going to stumble on this RSE on a next cycle only once the results of the current iteration is reflected everywhere down the chain

So we  are on the same page. Only that all these changes I have foreseen in the next issue, where the actual per RSE load and work would be estimated and executed, At that moment we will have a better view the situation and how to organize those limits. 

And, yes we can think of a non boolean resolution of the question what actions to be taken towards a partially cleaned RSEs.

todor-ivanov: BTW we should make a note and link the next issue https://github.com/dmwm/WMCore/issues/10410 to this very comment so we  do not forget this.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: removing it :)

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
todor-ivanov: note made

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: done

todor-ivanov:
Thanks for your replies @amaltaro I implemented the changes you requested. Could you please take another look.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
amaltaro: > As I mentioned in my previous reply, this effect should be mitigated by the limitFilesPerRSE configuration parameter.

The only way I can think of, to mitigate it, would be if we kept the list of files that we have successfully deleted persisted somewhere. Otherwise we have no idea which of the 100k files (from the example above) have been already deleted or not.
Maybe something to be considered when we attach this service to a database.

All right, let's keep discussing it in your next pull request then, once you implement the real file deletion logic.

Thanks for this follow up, Todor.

amaltaro (approved):
Thanks for providing further changes, Todor. Everything looks good to me, please squash the commits (and amend the commit message if necessary).
I made a note on all the 3 deployment change pull requests as well, where a typo needs to be fixed. Thanks!

todor-ivanov:
Thanks @amaltaro
I just squashed them. And also fixed the misalignment of variable names in `msConfig['rseExpression']` here vs. `data.rseExpr` in the `config-unmerged.py` file.
And here are all the commits fixing the typo in the config files too:

https://github.com/dmwm/deployment/pull/1061/commits/03d85bf7b96e6c707ae01d95d49aa81c84cc3cdf

https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/90/diffs?commit_id=f3fd8f0fdb63269122dc8649951c71a89d8a2395

https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/91/diffs?commit_id=d26dcff6c77f13debdfde208763519795a129ab2

amaltaro:
Thank you very much, Todor.

PR #10615, opened 2021-06-22
First draft of chagnes to generate Lexicon regexp in JSON data-format

vkuznet:
Fixes #10614 

#### Status
In development

#### Description
Change code base to provide Lexicon regexp in JSON data-format. The code is independent from external libraries and can be used as simple as following:
```
python Lexicon.py > lexicon.json
```

#### Is it backward compatible (if not, which system it affects?)
YES

I adjusted the code such that it is fully backward compatible, and yet can be used outside of WMCore.

#### External dependencies / deployment changes
None

PR #10622, opened 2021-06-24
Second draft of changes to generate Lexicon regexp in JSON data-format

### LINKED ISSUE #10614: Generate Lexicon regexp in portable data-format (JSON)
**Impact of the new feature**
Generate Lexicon regexp in portable data-format (e.g. JSON) to be used in other programming languages (GoLang).

In order to maintain single Lexicon regexp set which can be used across multiple languages I suggest to adjust Lexicon.py module to generate lexicon regexp into portable data-format, e.g. JSON.

Adjust Lexicon.py module to generate lexicon regexp in JSON

the alternatives include implementation of CMS lexicon regexp in different languages which can diverge eventually.

**Additional context**
This feature request is based on discussion about DBS GoLang implementation.

vkuznet:
Fixes #10614 

#### Status
In development

#### Description
Change code base to provide Lexicon regexp in JSON data-format. The code is independent from external libraries and can be used as simple as following:
```
python Lexicon.py > lexicon.json
# and it produces the following format, the length -1 means no restrictions
[
{"name": <object-name>}, "regexp": [list of regexps], "length": int},
...
]
```

YES

I adjusted the code such that it is fully backward compatible, and yet can be used outside of WMCore.

#### Related PRs
#10615 

None

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -840,7 +840,7 @@ def getIterMatchObjectOnRegexp(filePath, regexp):
         for m in re.finditer(regexp, f.read()):
             yield m
 
-def regexp_json():
+def dump_regexp(fname=None):
     """
     Return all regexp expression used in Lexicon. The format of lexicon JSON is the following:
     [
@@ -853,41 +853,66 @@ def regexp_json():
     blockLength = 4 + PRIMARY_DS['maxLength'] + PROCESSED_DS['maxLength'] + BLOCK_STR['maxLength']
     datasetLength = 3 + PRIMARY_DS['maxLength'] + PROCESSED_DS['maxLength']
     rdict = [
-        {"name": "logical_file_name", "regexp": [r'{}'.format(i) for i in lfn("", True)], "length": lfnLength},
-        {"name": "block_name", "regexp": [r'{}'.format(i) for i in block("", True)], "length": blockLength},
-        {"name": "dataset", "regexp": [r'{}'.format(i) for i in dataset("", True)], "length": datasetLength},
-        {"name": "processed_ds_name", "regexp": [r'{}'.format(i) for i in procdataset("", True)], "length": -1},
-        {"name": "user", "regexp": [r'{}'.format(i) for i in DBSUser("", True)], "length": -1},
-        {"name": "search_block", "regexp": [r'{}'.format(i) for i in searchblock("", True)], "length": -1},
-        {"name": "search_dataset", "regexp": [r'{}'.format(i) for i in searchdataset("", True)], "length": -1},
-        {"name": "searchstr", "regexp": [r'{}'.format(i) for i in searchstr("", True)], "length": -1},
-        {"name": "namestr", "regexp": [r'{}'.format(i) for i in namestr("", True)], "length": -1},
-        {"name": "site_tier", "regexp": [r'{}'.format(i) for i in sitetier("", True)], "length": TIER['maxLength']},
-        {"name": "job_range", "regexp": [r'{}'.format(i) for i in jobrange("", True)], "length": -1},
-        {"name": "cms_name", "regexp": [r'{}'.format(i) for i in cmsname("", True)], "length": -1},
-        {"name": "country_code", "regexp": [r'{}'.format(i) for i in countrycode("", True)], "length": -1},
-        {"name": "identifier", "regexp": [r'{}'.format(i) for i in identifier("", True)], "length": -1},
-        {"name": "global_tag", "regexp": [r'{}'.format(i) for i in globalTag("", True)], "length": -1},
-        {"name": "publish_dataset_name", "regexp": [r'{}'.format(i) for i in publishdatasetname("", True)], "length": -1},
-        {"name": "user_processed_dataset", "regexp": [r'{}'.format(i) for i in userprocdataset("", True)], "length": -1},
-        {"name": "physics_group", "regexp": [r'{}'.format(i) for i in physicsgroup("", True)], "length": -1},
-        {"name": "processed_ds_name", "regexp": [r'{}'.format(i) for i in procversion("", True)], "length": PROCESSED_DS['maxLength']},
-        {"name": "processed_string", "regexp": [r'{}'.format(i) for i in procstring("", True)], "length": -1},
-        {"name": "processed_string_t0", "regexp": [r'{}'.format(i) for i in procstringT0("", True)], "length": -1},
-        {"name": "acquisition_name", "regexp": [r'{}'.format(i) for i in acqname("", True)], "length": -1},
-        {"name": "campaign", "regexp": [r'{}'.format(i) for i in campaign("", True)], "length": -1},
-        {"name": "primary_ds_name", "regexp": [r'{}'.format(i) for i in primdataset("", True)], "length": PRIMARY_DS['maxLength']},
-        {"name": "task_step_name", "regexp": [r'{}'.format(i) for i in taskStepName("", True)], "length": TASK_STEP_NAME['maxLength']},
-        {"name": "hn_name", "regexp": [r'{}'.format(i) for i in hnName("", True)], "length": -1},
-        {"name": "lfn_base", "regexp": [r'{}'.format(i) for i in lfnBase("", True)], "length": -1},
-        {"name": "user_lfn", "regexp": [r'{}'.format(i) for i in userLfn("", True)], "length": -1},
-        {"name": "cmssw_version", "regexp": [r'{}'.format(i) for i in cmsswversion("", True)], "length": -1},
-        {"name": "request_name", "regexp": [r'{}'.format(i) for i in requestName("", True)], "length": -1},
-        {"name": "couch_url", "regexp": [r'{}'.format(i) for i in couchurl("", True)], "length": -1},
-        {"name": "activity", "regexp": [r'{}'.format(i) for i in activity("", True)], "length": -1},
-        {"name": "primary_ds_type", "regexp": [r'{}'.format(i) for i in primaryDatasetType("", True)], "length": -1},
+        {"name": "logical_file_name", "patterns": [r'{}'.format(i) for i in lfn("", True)], "length": lfnLength},
+        {"name": "block_name", "patterns": [r'{}'.format(i) for i in block("", True)], "length": blockLength},
+        {"name": "dataset", "patterns": [r'{}'.format(i) for i in dataset("", True)], "length": datasetLength},
+        {"name": "processed_ds_name", "patterns": [r'{}'.format(i) for i in procdataset("", True)], "length": -1},
+        {"name": "user", "patterns": [r'{}'.format(i) for i in DBSUser("", True)], "length": -1},
+        {"name": "search_block", "patterns": [r'{}'.format(i) for i in searchblock("", True)], "length": -1},
+        {"name": "search_dataset", "patterns": [r'{}'.format(i) for i in searchdataset("", True)], "length": -1},
+        {"name": "searchstr", "patterns": [r'{}'.format(i) for i in searchstr("", True)], "length": -1},
+        {"name": "namestr", "patterns": [r'{}'.format(i) for i in namestr("", True)], "length": -1},
+        {"name": "site_tier", "patterns": [r'{}'.format(i) for i in sitetier("", True)], "length": TIER['maxLength']},
+        {"name": "job_range", "patterns": [r'{}'.format(i) for i in jobrange("", True)], "length": -1},
+        {"name": "cms_name", "patterns": [r'{}'.format(i) for i in cmsname("", True)], "length": -1},
+        {"name": "country_code", "patterns": [r'{}'.format(i) for i in countrycode("", True)], "length": -1},
+        {"name": "identifier", "patterns": [r'{}'.format(i) for i in identifier("", True)], "length": -1},
+        {"name": "global_tag", "patterns": [r'{}'.format(i) for i in globalTag("", True)], "length": -1},
+        {"name": "publish_dataset_name", "patterns": [r'{}'.format(i) for i in publishdatasetname("", True)], "length": -1},
+        {"name": "user_processed_dataset", "patterns": [r'{}'.format(i) for i in userprocdataset("", True)], "length": -1},
+        {"name": "physics_group", "patterns": [r'{}'.format(i) for i in physicsgroup("", True)], "length": -1},
+        {"name": "processed_ds_name", "patterns": [r'{}'.format(i) for i in procversion("", True)], "length": PROCESSED_DS['maxLength']},
+        {"name": "processed_string", "patterns": [r'{}'.format(i) for i in procstring("", True)], "length": -1},
+        {"name": "processed_string_t0", "patterns": [r'{}'.format(i) for i in procstringT0("", True)], "length": -1},
+        {"name": "acquisition_name", "patterns": [r'{}'.format(i) for i in acqname("", True)], "length": -1},
+        {"name": "campaign", "patterns": [r'{}'.format(i) for i in campaign("", True)], "length": -1},
+        {"name": "primary_ds_name", "patterns": [r'{}'.format(i) for i in primdataset("", True)], "length": PRIMARY_DS['maxLength']},
+        {"name": "task_step_name", "patterns": [r'{}'.format(i) for i in taskStepName("", True)], "length": TASK_STEP_NAME['maxLength']},
+        {"name": "hn_name", "patterns": [r'{}'.format(i) for i in hnName("", True)], "length": -1},
+        {"name": "lfn_base", "patterns": [r'{}'.format(i) for i in lfnBase("", True)], "length": -1},
+        {"name": "user_lfn", "patterns": [r'{}'.format(i) for i in userLfn("", True)], "length": -1},
+        {"name": "cmssw_version", "patterns": [r'{}'.format(i) for i in cmsswversion("", True)], "length": -1},
+        {"name": "request_name", "patterns": [r'{}'.format(i) for i in requestName("", True)], "length": -1},
+        {"name": "couch_url", "patterns": [r'{}'.format(i) for i in couchurl("", True)], "length": -1},
+        {"name": "activity", "patterns": [r'{}'.format(i) for i in activity("", True)], "length": -1},
+        {"name": "primary_ds_type", "patterns": [r'{}'.format(i) for i in primaryDatasetType("", True)], "length": -1},
     ]
+    # if we are given output file name we'll dump json to it
+    if fname:
+        with open(fname, 'w') as ostream:
+            ostream.write(json.dumps(rdict))
+        return
+    # otherwise we'll return json to upstream
     return json.dumps(rdict)
 
+def load_regexp(fname, convert_to_dict=False):
+    """
+    Load regexp Lexicon file. It is reverse function to dump_regexp.
+    It either return patterns Lexicon list, see dump_regexp data-format, or
+    convert it to dictionaries. One dictionary contains keys and patterns,
+    while another keys and lengths.
+    """
+    data = []
+    with open(fname, 'r') as istream:
+        data = json.load(istream)
+    if not convert_to_dict:
+        return data
+    rdict = {}
+    ldict = {}
+    for item in data:
+        rdict[item['name']] = item['patterns']
+        ldict[item['name']] = item['length']
+    return rdict, ldict
+
 if __name__ == '__main__':
-    print(regexp_json())
+    print(dump_regexp())
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -838,7 +838,7 @@ def getIterMatchObjectOnRegexp(filePath, regexp):
     "provides iter matching object"
     with io.open(filePath, 'r', encoding='utf8', errors='ignore') as f:
         for m in re.finditer(regexp, f.read()):
-            yield 
+            yield m
 
 def regexp_json():
     """
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -15,10 +15,20 @@ from future.utils import viewvalues
 
 import io
 import re
+import json
 
 from urllib.parse import urlparse, urlunparse
 
-from WMCore.WMException import WMException, WMEXCEPTION_START_STR, WMEXCEPTION_END_STR
+# this try block provides ability to load this module without WMCore dependency, e.g.
+# when we need to regenerate lexicon dict
+try:
+    from WMCore.WMException import WMException, WMEXCEPTION_START_STR, WMEXCEPTION_END_STR
+except:
+    WMEXCEPTION_START_STR = str("<@========== WMException Start ==========@>")
+    WMEXCEPTION_END_STR = str("<@---------- WMException End ----------@>")
+    def WMException(msg):
+        print(msg)
+
 
 # restriction enforced by DBS. for different types blocks.
 # It could have a strict restriction
@@ -63,7 +73,7 @@ CONDOR_LOG_FILTER_REGEXP = re.compile(r"%s|%s" % (CONDOR_LOG_REASON_FILTER, COND
                                       re.DOTALL)
 
 
-def DBSUser(candidate):
+def DBSUser(candidate, return_regexp=False):
     """
     create_by and last_modified_by in DBS are in several formats. The major ones are:
     1. DN that was mostly used in DBS2: example /DC=org/DC=doegrids/OU=People/CN=Lothar A.T. Bauerdick 301799;
@@ -76,6 +86,8 @@ def DBSUser(candidate):
     r2 = r'^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_\']*$'
     r3 = r'^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_]*@[a-zA-Z0-9/][a-zA-Z0-9/\.\-_]*$'
 
+    if return_regexp:
+        return [r1, r2, r3]
     errorMsg = "DBSUser candidate: %s doesn't match any of the following regular expressions:\n" % candidate
     try:
         return check(r1, candidate)
@@ -94,25 +106,29 @@ def DBSUser(candidate):
         raise AssertionError(errorMsg)
 
 
-def searchblock(candidate):
+def searchblock(candidate, return_regexp=False):
     """
     A block name with a * wildcard one or more times in it.
     """
     regexp = r"^/(\*|[a-zA-Z\*][a-zA-Z0-9_\*]{0,100})(/(\*|[a-zA-Z0-9_\.\-\*]{1,199})){0,1}(/(\*|[A-Z\-\*]{1,99})(#(\*|[a-zA-Z0-9\.\-_\*]){0,100}){0,1}){0,1}$"
+    if return_regexp:
+        return [regexp]
     return check(regexp, candidate)
 
 
 SEARCHDATASET_RE = r'^/(\*|[a-zA-Z\*][a-zA-Z0-9_\*\-]{0,100})(/(\*|[a-zA-Z0-9_\.\-\*]{1,199})){0,1}(/(\*|[A-Z\-\*]{1,50})){0,1}$'
 
 
-def searchdataset(candidate):
+def searchdataset(candidate, return_regexp=False):
     """
     A dataset name with a * wildcard one or more times in it. Only the first '/' is mandatory to use.
     """
+    if return_regexp:
+        return [SEARCHDATASET_RE]
     return check(SEARCHDATASET_RE, candidate)
 
 
-def searchstr(candidate):
+def searchstr(candidate, return_regexp=False):
     """
     Used to check a DBS input that searches for names in dbs. Note block name, dataset name, file name have their own
     searching string.
@@ -120,48 +136,67 @@ def searchstr(candidate):
     letters, numbers, periods, dashes, underscores
 
     """
+    regexp = r'^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$'
+    if return_regexp:
+        return [regexp]
     if candidate == '':
         return candidate
-    return check(r'^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$', candidate)
+    return check(regexp, candidate)
 
 
-def namestr(candidate):
+def namestr(candidate, return_regexp=False):
     """
     Any input used in DBS and not defined here should pass namestr check.
     No white space found in names in DBS production and allowed to elimate input like "Drop table table1".
     letters, numbers, periods, dashes, underscores,#,/
 
     """
+    regexp = r'^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_/#]*$'
+    if return_regexp:
+        return [regexp]
     if candidate == '' or not candidate:
         return candidate
-    return check(r'^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_/#]*$', candidate)
+    return check(regexp, candidate)
 
 
-def sitetier(candidate):
-    return check("^T[0-3]", candidate)
+def sitetier(candidate, return_regexp=False):
+    regexp = "^T[0-3]"
+    if return_regexp:
+        return [regexp]
+    return check(regexp, candidate)
 
 
-def jobrange(candidate):
+def jobrange(candidate, return_regexp=False):
     """ Specifies a numbers/range of jobs separated by a comma.
         A range is composed by two numbers separated my minus
         For example valid candidates are either 1 or 1,2 or 3-6,5,7-8
         It is like when you specifies which pages to print in Word
     """
-    return check("^\d+(-\d+)?(,\d+(-\d+)?)*$", candidate)
+    regexp = "^\d+(-\d+)?(,\d+(-\d+)?)*$"
+    if return_regexp:
+        return [regexp]
+    return check(regexp, candidate)
 
 
-def cmsname(candidate):
+def cmsname(candidate, return_regexp=False):
     """
     Candidate must pass the CMS name pattern. Thus:
      * good candidates: T2_UK_SGrid or T2_UK_SGrid_Bristol
      * bad candidates: T2, T2_UK
     """
+    regexp = "^T[0-3]_[A-Z]{2}((_[A-Za-z0-9]+)+)$"
+    if return_regexp:
+        return [regexp]
     candidate = candidate.rstrip('_')
-    return check("^T[0-3]_[A-Z]{2}((_[A-Za-z0-9]+)+)$", candidate)
+    return check(regexp, candidate)
 
 
-def countrycode(candidate):
+def countrycode(candidate, return_regexp=False):
+    "country code check"
     # TODO: do properly with a look up table
+    regexp = "^[A-Z]{2}$"
+    if return_regexp:
+        return [regexp]
     return check("^[A-Z]{2}$", candidate)
 
 
@@ -178,9 +213,12 @@ def _blockStructCheck(candidate):
     return parts
 
 
-def block(candidate):
+def block(candidate, return_regexp=False):
     """assert if not a valid block name"""
 
+    if return_regexp:
+        regexp = '/{}/{}/{}#{}'.format(PRIMARY_DS['re'], PROCESSED_DS['re'], TIER['re'], BLOCK_STR['re'])
+        return [regexp]
     parts = _blockStructCheck(candidate)
 
     primDSCheck = check(r"%s" % PRIMARY_DS['re'], parts[1], PRIMARY_DS['maxLength'])
@@ -191,32 +229,42 @@ def block(candidate):
     return (primDSCheck and procDSCheck and tierCheck and blockCheck)
 
 
-def identifier(candidate):
+def identifier(candidate, return_regexp=False):
     """ letters, numbers, whitespace, periods, dashes, underscores """
-    return check(r'[a-zA-Z0-9\s\.\-_]{1,100}$', candidate)
+    regexp = r'[a-zA-Z0-9\s\.\-_]{1,100}$'
+    if return_regexp:
+        return [regexp]
+    return check(regexp, candidate)
 
 
-def globalTag(candidate):
+def globalTag(candidate, return_regexp=False):
     """ Identifier plus colons """
-    return check(r'[a-zA-Z0-9\s\.\-_:]{1,100}$', candidate)
+    regexp = r'[a-zA-Z0-9\s\.\-_:]{1,100}$'
+    if return_regexp:
+        return [regexp]
+    return check(regexp, candidate)
 
 
 DATASET_RE = r'^/[a-zA-Z0-9\-_]{1,99}/[a-zA-Z0-9\.\-_]{1,199}/[A-Z\-]{1,50}$'
 
 
-def dataset(candidate):
+def dataset(candidate, return_regexp=False):
     """ A slash followed by an identifier,x3 """
+    if return_regexp:
+        return [DATASET_RE]
     return check(DATASET_RE, candidate)
 
 
 PROCDATASET_RE = r'[a-zA-Z][a-zA-Z0-9_]*(\-[a-zA-Z0-9_]+){0,2}-v[0-9]*$'
 
 
-def procdataset(candidate):
+def procdataset(candidate, return_regexp=False):
     """
     Check for processed dataset name.
     letters, numbers, dashes, underscores.
     """
+    if return_regexp:
+        return [PROCDATASET_RE, PROCESSED_DS['re']]
     if not candidate or candidate.startswith('None'):
         raise AssertionError("ProcDataset cannot be empty or start with None.")
 
@@ -225,7 +273,10 @@ def procdataset(candidate):
     return (commonCheck and prodCheck)
 
 
-def publishdatasetname(candidate):
+def publishdatasetname(candidate, return_regexp=False):
+    "Publish dataset name check"
+    if return_regexp:
+        return [r'%(publishdataname)s$' % userProcDSParts]
     if candidate == '' or not candidate:
         return candidate
     return check(r'%(publishdataname)s$' % userProcDSParts, candidate, 100)
@@ -234,11 +285,13 @@ def publishdatasetname(candidate):
 USERPROCDATASET_RE = r'%(groupuser)s-%(publishdataname)s-%(psethash)s$' % userProcDSParts
 
 
-def userprocdataset(candidate):
+def userprocdataset(candidate, return_regexp=False):
     """
     Check for processed dataset name of users.
     letters, numbers, dashes, underscores.
     """
+    if return_regexp:
+        return [USERPROCDATASET_RE]
     if candidate == '' or not candidate:
         return candidate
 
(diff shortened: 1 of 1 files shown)


vkuznet:
@yuyiguo , @amaltaro , @todor-ivanov , @klannon This PR contains second draft proposal to use common Lexicon data-format. After the initial evaluation I think this draft does not have any downsides and fully backward compatible with existing Lexicon. It also generalize better then original PR #10615 , since the format now contains static attribute names and include length of regexp objects. My vote would be to use this proposal.

Please review and provide your feedback.

amaltaro:
Valentin, thanks for providing these changes.

I have a slightly different opinion on how it should be implemented though, which is: the current "dumpRegex" changes look fine to me, but I would only use it to generate the Lexicon rules. In other words, once we have the JSON files, the changes provided here are no longer relevant.

We do have to add logic to load the json file in Lexicon though, to be done as soon as any import Lexicon is done.

Regarding the flat JSON format, with static key names, it's indeed more organized and clear, but then it requires iterating over that dictionary every time we want to execute one of the Lexicon functions. Here is where I would be inclined to have a "dynamic" key name such that we can have direct access to whatever rule we want. If we stick to the flat JSON - which I'm pretty sure it's your preference - then we could create a helper function to lookup at the rule we're looking for, which wouldn't be too bad, besides wasting extra CPU cycles.

vkuznet:
Alan, please say explicitly what do you want to implement (if any). From your message I think you want:
- a function which reads the given json lexicon file and provide dict like look-up of any key

Please note, the static key JSON has nothing to do dynamic keys. The former represent the data, the later is helper function which can be provided. We may have a function which reads the JSON file, then build the dict once and everybody can use. There is no extra CPU cycles involved. 

Please reply with concrete action you want to see in this ticket.

amaltaro:
>     a function which reads the given json lexicon file and provide dict like look-up of any key

That's exactly what I had in mind.

> Please note, the static key JSON has nothing to do dynamic keys. The former represent the data, the later is helper function which can be provided. We may have a function which reads the JSON file, then build the dict once and everybody can use. There is no extra CPU cycles involved.

Not necessarily, if you simply load your JSON into a python dictionary, you stick to that format (which can or cannot have dynamic keys).
But you just mention something that I had not thought about, indeed we can have a flat and static JSON file, load and parse it into a function (global variable?) and then provide the dictionary key look-up.

vkuznet:
@amaltaro, that's exactly my point, the data format should independent from language. They allow to exchange data among applications and different implementations, therefore they should be always generic. If application wants specific structure then the data loaded from generic format can be converted to that structure.

Ok, I'll post an update with load function and a dict you so much desire, but I'll preserve the independent data-format.

vkuznet:
Here is second iteration. Now it contains two functions:
- `dump_regexp` to dump Lexicon data into stdout or given file name
- `load_regexp` to load Lexicon data from JSON file and either return it or provide two dictionaries (one which contains key and patterns, and another which provides keys and lengths).

The upstream code can decide how to keep loaded data, it can loaded into an object (within a class), or global variable. Here is an example:
```
# dump the lexicon data to stdout
python Lexicon.py
[{"name": "logical_file_name", "patterns": [...], "length": 500], ...

# dump lexicon data to a file
python Lexicon.py > /tmp/lexicon.json

# load the Lexicond data from a file
from Lexicon import load_regexp
fname = '/tmp/lexicon.json'
data = load_regexp(fname)
print(data)
[{'name': 'logical_file_name', 'patterns': [...], 'length': 500}, ...]
rdict, ldict = load_regexp(fname, convert_to_dict=True)
print(ldict)
{'logical_file_name': 500, ...}
print(rdict)
{'logical_file_name': [...], ...}
```

Please review and let me know if it satisfies your view

vkuznet:
@amaltaro , @klannon, @yuyiguo  could you please revisit this PR and associated feature request https://github.com/dmwm/WMCore/issues/10614 that I can apply changes and move forward with dbs2go migration to use common lexicon. The PR is already 6 weeks old and I was expecting that at least it would be properly reviewed during this time that we can move forward.

amaltaro:
Sorry for the delay, Valentin.

My opinion is that we should not push this code to production as is. Instead, I'm in favor of getting a dump of all the regular expressions that we need, and adapt the Lexicon code to use those patterns from the external file. In other words, I would not push to production this parameter `return_regexp`

vkuznet:
@amaltaro , could you please convert your opinion to practical set of actions. So far I don't know if it is feasible to make backward compatible code with what you're suggesting. For instance, plenty of function from Lexicon.py has hardcoded regexp, e.g. `def searchblock`. How you can feed into that function? How you can generalize `def lfn` function from external file, it has quite complex flow of if statement, where majority of regexp are dynamically created by the code? I want to be realistic and provide changes which will allow the transition without major interruption and code-refactoring of multiple places (who knows where exactly in WMCore/DBS these functions are used). The PR I provided can achieve both goals, i.e. preserve current behavior Lexicon.py file and provide ability to dump regexp's into external file.

amaltaro:
@vkuznet Valentin, what I have in mind is something like.

First: 
### Generate a json file like (basically with your current changes in place):
```
[{"functionName": "blahA", "length": 123, "regexSet": ('^[a-z]', '^[A-Z]', 'blah')},
 {"functionName": "blahB", "length": None, "regexSet": ('^[A-Z]', 'blah')},
...]
```

Second: 
### Load rules in memory at global scope and change its structure for easy look-up (inside Lexicon.py)
```
REGEX_RULES = {}
with open("./Lexicon_rules.json") as fp:
    for item in json.load(fp):
        REGEX_RULES[item.pop("functionName")] = {}
        REGEX_RULES[item[functionName]].update(item)
```

Third:
### Update the Lexicon source code to look like:
```
def searchblock(candidate):
    rules = REGEX_RULES["searchblock"]
    for item in rules:
        # NOTE: we could iterate through the list of regex values, even if the function is meant to have a single regex
        # at least we would guarantee that whatever changes made in the future to the json file, would be taken into account here
        return check(item["regexSet"][0], candidate, maxLength=item["length"])

def lfnBase(candidate):
    errorMsg = "LFN candidate: %s doesn't match any of the following regular expressions:\n" % candidate
    rules = REGEX_RULES["lfnBase"]
    for item in rules:
        for regex in item["regexSet"]:
            try:
                return check(regex, candidate, maxLength=item["length"])
            except AssertionError:
                errorMsg += "  {}, with length: {}\n".format(regex, item["length"])
        raise AssertionError(errorMsg)
```

See my comment in the snippet above as well. It's basically a tradeoff between robustness vs simplicity/performance.
What do you think? Do you see possible possible improvements?

vkuznet:
@amaltaro I doubt it can work in general way. Here few flaws in your proposal:
- where is default location of lexicon.json file? Is it current directory, env variable or should be provided externally?
- some regexp contradict each other, e.g. processed dataset is defined as independent at line 27 (see https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L27) and as part of `lfnParts` dict, see https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py#L36. And, those regexp are different. And, neither belong to specific function. Therefore, I can't conclude from your proposal how to combine them in general form since it seems to me it requires different structures, with and without function name, independent and part of another structure
- you're mixing function and global variables, e.g. searchblock is a function, while lfnBase is a module (global) variable. The latter, is used in other functions to construction some regexp. 
- the current codebase uses mixture of global variables and functions, e.g. there are two functions `searchblock` vs `block`. The former uses internally defied regexp, while latter relies on global variables. But both represent a block regexp patterns. How to do a separation of patterns in this case?
- does the code should be backward compatible, I have no idea where it is used (it can be used in CRAB for instance) and I have no idea which functions are used. Therefore, if I strip off all hard-coded regexp from the code and I can't properly assign them into different parts (see above example with processed dataset regexp) I don't know if code will behave correctly for different clients.
- I also don't see any needs to keep function name as part of data-structure since regexp does not represent functions, they represent the data. For instance, there is a SEARCHDATASET_RE, there is searchstr and dataset functions, but all of them represent different dataset patterns. How these variables and functions are used I have no idea.

Bottom line, I don't want to do a third iteration until clear idea is outlined how to represent regexp's. In my view they should represent the data, e.g. list of regexp for all possible dataset values, etc. Since the function defined for a specific use-cases, e.g. searchst vs dataset functions, they use different subset of regexp set.

I think it should be only two APIs: how to load lexicon.json and how to generate it. Then, all function should be cleaned up to use a set of regexp associated with metadata parameters, a block name. Although this will require to change functions and will  break backward compatibility, it will allow to have clear definitions of regexp as a list of allowed expressions. Therefore, I suggest you provide further clarification how to resolve all aforementioned issues.

amaltaro:
>I think it should be only two APIs: how to load lexicon.json and how to generate it. Then, all function should be cleaned up to use a set of regexp associated with metadata parameters, a block name. Although this will require to change functions and will break backward compatibility, it will allow to have clear definitions of regexp as a list of allowed expressions. Therefore, I suggest you provide further clarification how to resolve all aforementioned issues.

Valentin, your sentence is a summary of what I am proposing.
And yes, I'm planning to have this backwards compatible, thus not affecting anything importing from Lexicon. The only and sole exception would be if someone is actually importing a regex pattern instead of the validate function; then I agree it would not be backwards compatible (however, that would likely be a wrong usage of the Lexicon module!).

The comment I made above should answer some of your concerns. In addition to that, if the same regular expression is used by multiple validate functions, we just need to define that regex for both functions.

I agree that regex represents data (or a standard for the data), however, we need to have a way to map which standard is used by which function, hence my suggestion for that json structure.

vkuznet:
@amaltaro please provide provide concrete plan, what you proposed is not complete since I pointed out to its flaws. Therefore, before implementing stuff I need the following:
- should we decide on default location of Lexicon.json path, should be take from env, should it be provided to the script
- should script load lexicon.json during `import lexicon` or not? If yes, see above.
- if function coverage is required, then we need to define proper data-format for lexicon.json, e.g.
```
# proposal 1 (data driven, as I define it, but it provides a way to extract from full list of patterns the ones required for function names)
[{"name": "dataset", "patterns": [ full set of regexp], length: int}..., "functions": [{"funcname": string, "funcindexes": [ ids of patterns to use in funcname], ...] ]

# proposal 2 (function driven, as you define it):
[{"name": "dataset", "patterns": [ list of regexp used in func], length: int}..., "funcname": string]

# other proposals ....
```
The proposal 1 above can be used to generalize the code and json, while in proposal 2 it is not possible.

I suggest that you at least clarify these two main issues before starting another implementation.

amaltaro:
>@amaltaro please provide provide concrete plan, what you proposed is not complete since I pointed out to its flaws. Therefore, before implementing stuff I need the following:

@vkuznet please do not get stuck waiting for my reply. If effort is low enough and you think you can propose something, please go ahead and bring this up for discussion/review. I count on you and the rest of the team to come up with plans as well ;)

> * should we decide on default location of Lexicon.json path, should be take from env, should it be provided to the script

Having an environment variable isn't a bad idea (LEXICON_RULES_PATH ?), but I would not enforce it. And if not set, I would just try to load it like `with open(".lexicon_rules.json") as ...`, and if that fails, then we should raise an `ImportError` or something like that.

> * should script load lexicon.json during import lexicon or not? If yes, see above.

Yes, the way I project it is that someone will import Lexicon - just as currently done - and this will bring in the complete validate functions (which will internally access the local module rules cache). Let me know if I'm missing something here though.
For DMWM projects that do not want to have our Lexicon implementation, I assume they would simply pull the lexicon_rules.json file when building/packaging their system. Right? 

> * if function coverage is required, then we need to define proper data-format for lexicon.json, e.g.

Can you please clarify what is your `name` key/value pair? Your proposal #2 is basically the same as what I proposed a day or two ago, besides the `name` key/value. I'd like to have a KISS principle here, thus, one rule line/item for each function. If the same regex is used by multiple functions, it's more clear to me to have that defined under each "function's rules".

Thanks for following this up, Valentin.

vkuznet:
@amaltaro I provided new PR with new format. Please have a look at #10764 . The PR only contains code changes, but does not yet provide `lexicon.json` file which I'll generate later if we agree on code structure.

amaltaro:
Okay, so the `name` attribute is meant to refer which information those rules are meant to, e..: "dataset", "request name", "lfn" and so on. I like this idea to characterize the rule being defined!

However, I have to disagree with your structure above, which I paste here as well:
```
[
{
  "name": "dataset", 
  "patterns": [ "^/[0-9a-zA-Z]{1,99}.*", "/[0-9a-zA-Z]{1,99}/[0-9a-zA-Z]{1,99}/[0-9a-zA-Z]{1,99}", ...], 
  "length": 200.
  "functions": [
       {"name": "searchdataset", "indexes": [0],
       {"name": "dataset", "indexes": [0,1],

   ]
},
...
}
```

and the reason is basically because it does not look simple! It's a very nested structure, where indexes **matter** and it will be more error prone whenever someone needs to update the json content.

I would like to proceed with your proposal #2, which is basically the same structure I suggested in one of my first comments, plus the `name` key/value pair. We can also rename `name` to `attribute`, or `attributeType`, since `name` is something extremely generic and vague.

Last but not least, I also agree with your suggestion of `-1` for no length limit (contrary to my None suggestion). Thanks

amaltaro:
Oh, on top of my reply above, we should make `funcName` a basic string, instead of a list.

vkuznet:
@amaltaro , the structure I proposed is not complex and not nested. It is well defined schema.  The tool will parse it to provide representation of lexicon suitable for specific use-case, e.g. WMCore or Go based applications. Please see [load_lexicon](https://github.com/dmwm/WMCore/pull/10764/files#diff-e440b57eb48eae90be13e652cfaef026928f310a182a799d1d92ccfeaf5e5bf5R65) in PR #10764. It is not complex at all and can return you any number of representations. With such structure the lexicon code becomes very trivial and very generic.

Anyway, the proposal #2 cannot be generalized. here it is for clarify
```
[{"name": "dataset", "patterns": [ list of regexp used in func], length: int, "funcname": string}, ...]
```
And, the reason it can't be generalized across applications because it is function driver (which defeats the purpose of lexicon which you agree should be data driven), i.e. each dict correspond to function used in Lexicon.py. For instance, if we have two functions associated with some attribute `name`, then we'll have two different entries, e.g.
```
[
   {"name": "dataset", "patterns: ListOne, "function": "searchdataset"},
   {"name": "dataset", "patterns: ListTwo, "function": "dataset"},

]
```
But both correspond to the same `name` dataset. So to get full list of regexp for our dataset we need to merge ListOne and ListTwo, and make a set. This is because the format is function driven and not data driven (as I proposed).

This is **NOT** the case I use in Go where I don't use python function, and I don't need them. I need full set of regexp rules. Of course, I can resolve this in Go by writing additional code to get all function entries for a given `name` and then make a union of regexp rules. As such this proposal is Python based, moreover is python based on existing Lexicon.py codebase and has not represent **generic** list of regexp rules.

As I said please decide on **global** meaning of lexicon, is it function driven or data driven. Then, the format should reflect this decision. If you worry about adding entries to lexicon.json, then we can provide a tool for that, e.g.
```
update_lexicon -pattern "regex-pattern" -function "funcname" -length <optional int>
```
Such tool can update lexicon.json accordingly to avoid manual mistakes.

Since we decide how to make lexicon to be generic I still think it should has data driven format. The functions are legacy artifact of Python code and only required for backward compatibility. Otherwise they have no real meaning since applications should only care about regexp rules.

Anyway, I will still leave final decision to you and will adjust code accordingly. Please evaluate properly the decision (not only from Python point of view but rather how we can use new lexicon format outside of WMCore). Once you decide about it please post final format to avoid confusion and I'll use it and adjust PR #10764 accordingly.

amaltaro:
Valentin, if we come back to the original goal and requirements for this new feature, which were:
* separate Lexicon data from the Lexicon source code/implementation
* having a portable format to any other project in any language
* (bonus) have all these changes made in a backwards compatible mode

then all these proposals can meet those requirements.

The same way WMCore/Lexicon can adapt to parse the nested structure you propose, the DBS Go implementation does for the proposal with the validate function name, which I'm advocating to be adopted. The main reason I am against your schema is that I know it will cause issues in the future, that function vs indexes in a list is evil and will definitely hurt us at some point!

If we wanted to make this schema really generic, we should actually ditch the function name from the schema and completely refactor WMCore/Lexicon accordingly. However, I do not think it's worth all that effort.

Anyways, CRAB also uses Lexicon heavily, so perhaps Stefano wants to provide some input on this discussion; and hopefully we can move forward here. @belforte

amaltaro:
@belforte Stefano, if you have any opinion on this matter, please do share it with us (sorry for the long thread/discussion as well). Otherwise, let's converge on a final implementation by our next team chat, Valentin.

belforte:
@amaltaro @vkuznet thanks for keeping me in the loop.
IMHO it would be important to preserve the current inteface, e.g.
```
 from WMCore import Lexicon
        Lexicon.lfn(testLfn)  # will raise if testLfn is not a valid lfn
```
that's what I'd call 'backward compatibility' and I'd like to have.
How `Lexicon.lfn` internally gets the regexp to use in the check, whether it is from reading JSON, or having it hardcoded, or using a cached dictionary... I do not care.
I guess there'd be no problem in changing the above to something like:
```
 from WMCore import Lexicon2
        Lexicon = Lexicon2()
        Lexicon.lfn(testLfn)  # will raise if testLfn is not a valid lfn
```
in case having an object instantiation helps for internal caching.

I do not believe that I can give helpful comment on "how to implement".

amaltaro:
@todor-ivanov your input is always welcome, please don't wait for an explicit request and feel free to share your thoughts on this as well ;)

todor-ivanov:
Thanks for mentioning @amaltaro . I was going through them both the code changes and the comments, and in all that I can mainly agree on few (high level) remarks. 
 * It is really important to preserve the backwards compatibility (well expressed by @belforte )
 * The flat dictionary structure does seem to be a good thing in this case 

Other than that... I cannot find anything which is not looking good in the code.

amaltaro:
@vkuznet Sorry for the delay getting back to this, Valentin.

It looks like we have a consensus on the requirements for this feature, but not yet on the schema for the rules.

So, I would like to proceed according to the comments provided in this comment: https://github.com/dmwm/WMCore/pull/10622#issuecomment-899240578
with the extra data name attribute that you suggested. 

So, the final **flat** JSON rules file would look like (I'm suggesting `attrName` instead of `name`, but we can discuss it as well):
```
[{"attrName": "dataset", "functionName": "blahA", "length": 123, "regexSet": ('^[a-z]', '^[A-Z]', 'blah')},
 {"attrName": "lfn", "functionName": "blahB", "length": None, "regexSet": ('^[A-Z]', 'blah')},
...]
```

the client load these rules can then:
a) either decide to build the data rule map based on the `functionName` (our case in WMCore Lexicon)
b) or decide to build the data rule map based on the `attrName` (likely the case for DBS2Go)

Can you implement these changes (either in this PR, or in one of your other PRs)? Valentin, please also take into consideration the recent changes to Lexicon, provided in this PR:
https://github.com/dmwm/WMCore/pull/10799

Once we have the new set of changes, we will be closer to converging into the final details and getting these changes in for the October release.

belforte:
should be the proper time to make sure this is also fixed: https://github.com/dmwm/WMCore/issues/10810

amaltaro:
> should be the proper time to make sure this is also fixed: #10810

This is going to be a big change, and mistakes can be easily done. So, I would rather have the new model in place first (which can be tested by all the unit tests we have - without a single line change), and only then work on the new GH issue. I'm happy to take that issue, once these changes are behind us ;)

vkuznet:
@amaltaro before going to implementation (I really would like to avoid writing 3rd time and then we'll decide to rewrite it again), one more detail you should clarify. How in your flat proposal you address multiple functions for the same attribute. For instance, current Lexicon.py has `def searchdataset` and `def dataset`, as well as `def searchblock` and `def block`. In my view they are redundant since searchdataset (searchblock) is part of dataset (block), respectively. I only worry that it will lead to confusion and inconsistencies. For instance, if someone update searchdataset regexp but does not update dataset one. In your function based approach it will lead to wrong set of regexp since searchdataset function may have some regexp which are not present in dataset one. In attribute approach I can take the union and it is not a big deal, but I do think functions are irrelevant. Don't take me wrong I'll implement it, but at least I want to show you the potential source of errors in the format you propose. Much better approach is to use attribute based format, i.e. a dataset has only one set of regexp and functions should use them all for backward compatibility, i.e. a search dataset is a dataset at the end which is allowed in DBS DB.

belforte:
could (part of) the confusion be due to the fact that DBS3 has to allow more freedom in format of things to search for than things to write, since we had to allow for stuff written in the more laxed DBS2 times to be still accessible ?

vkuznet:
@belforte  the confusion is in function usage. I don't know who uses `searchdataset` and who `dataset`, but to me they refer to one and only one entity in DB, a `dataset`. As such a single entity should have only one set of regexp. I understand the legacy of this codebase, but I would like to move forward with correct data format which represent DB entities. The legacy functions should in my opinion exists but refer to a single, well defined set of regexp based on attributes (entities).

belforte:
+1

amaltaro:
>Don't take me wrong I'll implement it, but at least I want to show you the potential source of errors in the format you propose. Much better approach is to use attribute based format, i.e. a dataset has only one set of regexp and functions should use them all for backward compatibility, i.e. a search dataset is a dataset at the end which is allowed in DBS DB.

The way I see it, the attribute based approach would have the same issue if you want to call it a `dataset` attribute. We **cannot** allow a dataset in the old (DBS2) construction to be inserted into the database. So, a union of those rules into the same attribute shouldn't be done.

For these quite old backwards compatibility code and rules, I think we will have to keep dedicate attributes to them. Even though those are all `dataset` in the database, we need to have a way to distinguish them between "before DBS3" and "during/post DBS3". 

Perhaps those cases would deserve an attribute/entity like `read_dataset` and `write_dataset` instead?

vkuznet:
well, `read_dataset` and `write_dataset` make sense since they reflect attribute and operations, and I'm fine with them, but how you map them to existing functions, e.g. `def searchdataset` and `def dataset`, the same question will go to other attributes too. Since it is unclear to me what function should represent I'm picky about them. I don't mind to implement them, but I don't yet understand the meaning of all functions used in Lexicon. To me the schema should be about DB entities and not about functions. Once, we agree on which attributes should be presented in `lexicon.json` I'll need to know the mapping of them to existing functions. So far the logic of `Lexicon.py` is far from clear.
Here are the existing functions:
- `dataset` vs `searchdataset` vs `searchstr`
- `block` vs `searchblock`
- `procdataset` vs `userprocdata` vs `procstring` vs `procstringT0`
- `lfn` vs `userLfn`

I suggest that you clarify the meaning of those and how we should construct regexp for them and more **importantly** why do we need to distinguish these regexp patterns. At the end each item above represent a single entity in DBS (even DBS2 vs DBS3 the dataset has identical meaning, even though the conventions for them may have changed). I would understand if you'll say `dataset` is for write dataset and `searchdataset` is for read dataset, but look at others, e.g. `proxXXX` functions and try to clarify their meaning.

I would like to avoid to have data representation based on function which are obsolete and no longer use, especially since other use-cases, e.g. dbs2go, does not use them. I need clarity how we define regexp. So far the suggested format does not provide that.

amaltaro:
I will have a look at those and try to clarify them. However, some of them are not used by WMCore services at all, and many have been implemented even before my arrival in the WM system.
Yes, I completely agree that code no longer used should be removed (and inconsistencies should be cleared out). And it's great that you are trying to tidy this up with this new schema. I will get back here once I have some information to share. Thanks!

amaltaro:
Actually, giving it a second thought, I'd like to propose the following steps:
1) do not change any functionality of the Lexicon APIs; besides the separation of rules and code (this should be the 1st commit)
2) if it passes all unit tests, then it gives us a high level of confidence that no (hidden) mistakes have been made;
3) then we re-evaluate these confusing functions and their rules and try to refactor Lexicon such that we can make it more clear and remove possibly obsolete code (2nd commit)
4) updates to unit tests (3rd commit)

My main concern is that a bug can be injected and remain in the system for many years, until we can spot it (like the one Stefano spotted this week!). Lexicon is used across many projects, so I'd rather play as safe as possible here (in addition to the initial backwards compatibility discussed many times in this thread).

belforte:
I like this because 3. will take some time as we look at code that none of us wrote and try to figure out if that is still needed or anyhow useful. Cleanup is good, but time consuming, the more you want to clean, the more you need to check.
One think I sort of "remember" is that originally there was the idea that Production wanted to make sure they enforce more strict rules for dataset and files formats than what DBS allows, while DBS was more accommodating to let CRAB users adopt more human readable and meaningful names. Things like "for user files, make sure that the username is part of the path, while there is no user name for production", or "make sure users can't create a LFN which Production may want to use one day".
One may surely argue why people did not write two Lexicons, one for DBS and one for WM components...  such arguments usually end with "at the time it seemed the least work".

In a way WMCore/Lexicon is much more than a handy representation of DBS schema constraints.

vkuznet:
@amaltaro your last comment does not address the main and primary issue, which is the data format to use and how this data format is related to the code and DB. I gave you clear examples that the format you propose has flaws, i.e. having multiple entries with the same attribute but different functions may lead to inconsistencies and it is unclear how to make union/intersections of regexps across such records/functions. I still believe that only attribute names, lengths, and list of regexps should be presented in a final format. Then different functions which refer to the same attribute, e.g. `def searchdataset` and `def dataset`, or `def searchblock` and `def block`, or `def lfn` and `def userLfn` will use the attribute's regexp list and check that given value satisfies them. I do agree that we may separate read/write entities, such that we'll have `read_dataset` and `write_dataset` which will represent operations, e.g. we can use `read_dataset` in `def searchdataset` and `write_dataset` in `def dataset`. This makes sense to me. But please resolve code ambiguities where multiple functions exists for the same attribute. In particular the following cases should be clarified:
- `def searchdataset` vs `def dataset` vs `def searchstr` vs `def namestr`, all of the refer to dataset pattern and use different regexps, but in DBS/DB we only have **one** dataset entity
- `def searchblock` vs `def block` refer to block usage and use different patterns, but in DBS/DB we **only** have one block entity
- `def lfn` vs `def userLfn` refer to LFN usage and use different patterns while in DBS/DB we **only** have one lfn entity
- `def procdataset` vs `def userprocdata` vs `def procstring` vs `def procstringT0 refer to a processing dataset patterns while in DBS/DB we have only **one** entitiy
Since these attributes in DB represented by single entity I need to know which regexp to apply to them in GoLang code. So far reading and using `Lexicon.py` functions I have no idea.

So, please clarify your plan with concrete steps, e.g. for your item #1 which data-format we should use. If you still insists on using function name in data format please clarify **how** should I translate and use it in GoLang code where there is no such functions and there is no way to call them. At the end the code represents access to the DB and regexp applies to the DB entities, and the code itself has no meaning to them, what matters is only set of regexps we need to know to call/use in a code.

amaltaro:
Please read my previous comment again. If you still cannot make sense of it, ping me on slack and we can have a quick zoom chat. Sorry to say it, but at this point, this discussion is just becoming a wastage of time!

amaltaro:
Valentin and I had a chat and the course of actions discussed is:
* Valentin will get a map of DBS REST endpoints, operation (read/write), and the Lexicon API(s) used
* Alan will create a new json dump of the rules and separate code from lexicon rules
* Then, we will start analysing these ambiguous cases and see what are the stakeholders, where and how it's used, and whether it's still required.

@belforte Stefano, would you be able to check which of the CRAB* services are depending on Lexicon APIs? Are those services using the function/APIs, or is it also using the regular expressions defined in the Lexicon.py module? Once I know which services (and repository) to look at, I can collect a map of APIs used and try to make sense what their use is for.

belforte:
**CRAB** does 2 things.
1. uses the functions to check names, in particular code equivalent to:
```python
from WMCore import Lexicon
Lexicon.lfn(testLfn)
Lexicon.userLfn(testLfn)
Lexicon.userprocdataset(outputDatasetTagToCheck)
Lexicon.primdataset(datasetNameToCheck)
```
the lines where those are used are:
  * https://github.com/dmwm/CRABServer/blob/11daa05d7d0866a1dbafbc67960cb70327501e76/src/python/TaskWorker/Actions/DagmanCreator.py#L246
  * https://github.com/dmwm/CRABServer/blob/11daa05d7d0866a1dbafbc67960cb70327501e76/src/python/TaskWorker/Actions/DagmanCreator.py#L274
  * https://github.com/dmwm/CRABServer/blob/11daa05d7d0866a1dbafbc67960cb70327501e76/src/python/CRABInterface/RESTUserWorkflow.py#L119
  * https://github.com/dmwm/CRABServer/blob/11daa05d7d0866a1dbafbc67960cb70327501e76/src/python/CRABInterface/RESTUserWorkflow.py#L199


2. import the regexp's themselves 
 * in here to print diagnostic information for users when we reject a dataset name
   * https://github.com/dmwm/CRABServer/blob/11daa05d7d0866a1dbafbc67960cb70327501e76/src/python/CRABInterface/RESTUserWorkflow.py#L131
 * in here it is more perverse, and as the #TODO in the code says already, it is an historical bad piece of code waiting for being "done right".
   * https://github.com/dmwm/CRABServer/blob/11daa05d7d0866a1dbafbc67960cb70327501e76/src/python/CRABInterface/Regexps.py#L1-L15
   *  The use case here is the need to provide REGEXPs to the REST input parameter validation in https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Validation.py

vkuznet:
The usage of `Lexicon.py` and regexp in DBS is a mix bag. The good part is that everything seems to be located in a single python DBS module: https://github.com/dmwm/DBS/blob/master/Server/Python/src/dbs/utils/DBSInputValidation.py, the bad news is that DBS uses a mix of functions from `Lexicon.py` and its own. For instance, a `dataset` used for DBS reader relies on `searchdataset` function from `Lexicon.py` if given value contains a pattern `*`, otherwise it uses its internal regexp coming from its function `reading_dataset_check` (see lines 65-68. The same applies to block, lfns, procds. From what I see in DBS code its reading patterns mostly comes from DBS code functions and sometimes from `Lexicon.py`, while write patterns comes from `Lexicon.py`. For instance, in lines 319-331. The first block is for writing and it uses `validationFunction` (which come from `Lexicon.py`, and their definitions is at line 240), while for reading patterns it uses its own functions like `reading_dataset_check`, `reading_block_check` (which are defined in DBS module itself).

This leads me to conclusion that the following functions from `Lexicon.py` are used for writing operations:
- `block`, `dataset`, `lfn`, `primdataset`, `procversion`, `acqname`, `globalTag`, `validateUrl`, `DBSUser`, `dataset`

while for reading operations DBS relies on mixture of its own functions and functions from `Lexicon.py`, e.g. `searchstr`, `namestr` (from `Lexicon.py`) and `reading_dataset_check`, `reading_block_check`, etc. from DBS code.

So, it will be fun to deduce from this code a set of regexp lists :) But is is clear that we need two set of regexp rules. One set of rules for DBS reader (DBS GET APIs), and another set of rules for writing operations (DBSWriter, i.e. DBS POST APIs). I'm fine to create two different JSON files which will cover read/write operations and use them in a future. This also suggest that the function I listed above for writing operations is what DMWM/WMCore/CRAB tools should use, while all reading regexp should be correctly identified via DBS codebase and merged with few `Lexicon.py` functions (like `searchstr`, `namestr` and `reading_xxx_check` DBS functions).

vkuznet:
While awaiting your input no how to proceed, I adjusted `dbs2go` code to use independent lexicon rules based on proposed JSON formats. If you are interested here are the rules:
- [`lexicon_writer.json`](https://github.com/vkuznet/dbs2go/blob/master/static/lexicon_writer.json)
- [`lexicon_reader.json`](https://github.com/vkuznet/dbs2go/blob/master/static/lexicon_reader.json)
And, I provided two additional json files (for each writer/reader rules) which are used in negative/positive unit tests. In other words the negative json shows wrong examples, while positive contains correct values for tested rules. The files can be found here:
- [`lexicon_reader_negative.json`](https://github.com/vkuznet/dbs2go/blob/master/static/lexicon_reader_negative.json)
- [`lexicon_reader_positive.json`](https://github.com/vkuznet/dbs2go/blob/master/static/lexicon_reader_positive.json)
- [`lexicon_writer_negative.json`](https://github.com/vkuznet/dbs2go/blob/master/static/lexicon_writer_negative.json)
- [`lexicon_writer_negative.json`](https://github.com/vkuznet/dbs2go/blob/master/static/lexicon_writer_negative.json)

For the record, the output of `dbs2go` unit tests looks like this (each test is easy to follow and see what regexp match was applied to each passed value). With such mechanism we can extensively tests the lexicon rules against any DBS entity. With such mechanism you can easily add any lexicon rules and we'll only need to update JSON files. The Python code should load JSON files and use them in functions used by WMCore/CRAB tools.
```
make test-lexicon
cd test && LD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DYLD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DBS_LEXICON_FILE=../static/lexicon_writer.json DBS_LEXICON_SAMPLE_FILE=../static/lexicon_writer_positive.json go test -v -run LexiconPositive
=== RUN   TestLexiconPositive
2021/10/11 15:09:28 logical_file_name=/store/user/one/two/three/file123.root matched with ^/store/(temp/)*(user|group)/(([a-zA-Z0-9\.]+)|([a-zA-Z0-9\-_]+))/([a-zA-Z][a-zA-Z0-9\-_]*)/(([a-zA-Z0-9\-_]+)/)+([a-zA-Z0-9\-_]+).root$
2021/10/11 15:09:28 logical_file_name=/store/lhe/123/file123.lhe.xz matched with ^/store/lhe/([0-9]+)/([a-zA-Z0-9\-_]+).lhe(.xz){0,1}$
2021/10/11 15:09:28 dataset=/ZMM/Summer11-DESIGN42_V11_428_SLHC1-v1/GEN-SIM matched with ^/[a-zA-Z0-9\-_]{1,99}/[a-zA-Z0-9\.\-_]{1,199}/[A-Z\-]{1,50}$
2021/10/11 15:09:28 block_name=/ZMM/Summer11-DESIGN42_V11_428_SLHC1-v1/GEN-SIM#8420a3e7-96cf-48b1-a6a8-f9abcd3de8ef matched with ^/[a-zA-Z][a-zA-Z0-9\-_]+/[a-zA-Z0-9\.\-_]+/[A-Z\-_]+#[a-zA-Z0-9\.\-_]+
2021/10/11 15:09:28 processed_ds_name=123-bla-123 matched with ^[a-zA-Z0-9\.\-_]+
2021/10/11 15:09:28 data_tier_name=RAW matched with ^([A-Z]{3,100})+$
2021/10/11 15:09:28 processing_version=123 matched with ^[0-9]+$
2021/10/11 15:09:28 global_tag=123 matched with ^[a-zA-Z0-9\s\.\-_:]{1,100}$
2021/10/11 15:09:28 create_by=123bak matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:28 create_by=123bak matched with ^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_\']*$
2021/10/11 15:09:28 acquisition_era_name=someEra matched with [a-zA-Z][a-zA-Z0-9_]*$
2021/10/11 15:09:28 physics_group=123 matched with ^([a-zA-Z0-9\-_]+)$
2021/10/11 15:09:28 last_modified_by=123blka matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:28 last_modified_by=123blka matched with ^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_\']*$
2021/10/11 15:09:28 performed positive lexicon test against ../static/lexicon_writer_positive.json
--- PASS: TestLexiconPositive (0.01s)
PASS
ok  	github.com/vkuznet/dbs2go/test	0.922s
cd test && LD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DYLD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DBS_LEXICON_FILE=../static/lexicon_writer.json DBS_LEXICON_SAMPLE_FILE=../static/lexicon_writer_negative.json go test -v -run LexiconNegative
=== RUN   TestLexiconNegative
2021/10/11 15:09:30 performed negative lexicon test against ../static/lexicon_writer_negative.json
--- PASS: TestLexiconNegative (0.01s)
PASS
ok  	github.com/vkuznet/dbs2go/test	0.428s
cd test && LD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DYLD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DBS_LEXICON_FILE=../static/lexicon_reader.json DBS_LEXICON_SAMPLE_FILE=../static/lexicon_reader_positive.json go test -v -run LexiconPositive
=== RUN   TestLexiconPositive
2021/10/11 15:09:33 dataset=/ZMM/Summer11-DESIGN42_V11_428_SLHC1-v1/GEN-SIM matched with ^/([a-zA-Z0-9]+[a-zA-Z0-9\-\._]*)/([a-zA-Z]+[a-zA-Z0-9_\-\.]*)/([A-Z_\-\.]+)$
2021/10/11 15:09:33 dataset=/ZMM/Summer11-DESIGN42_V11_428_SLHC1-v1/GEN-SIM matched with ^/(\*|[a-zA-Z\*][a-zA-Z0-9_\*\-]{0,100})(/(\*|[a-zA-Z0-9_\.\-\*]{1,199})){0,1}(/(\*|[A-Z\-\*]{1,50})){0,1}$
2021/10/11 15:09:33 block_name=/ZMM/Summer11-DESIGN42_V11_428_SLHC1-v1/GEN-SIM#8420a3e7-96cf-48b1-a6a8-f9abcd3de8ef matched with ^/([a-zA-Z0-9]+[a-zA-Z0-9\-\._]*)/([a-zA-Z]+[a-zA-Z0-9_\-\.]*)/([A-Z_\-\.]+)#([a-fA-F0-9\-\.]+)$
2021/10/11 15:09:33 block_name=/ZMM/Summer11-DESIGN42_V11_428_SLHC1-v1/GEN-SIM#8420a3e7-96cf-48b1-a6a8-f9abcd3de8ef matched with ^/(\*|[a-zA-Z\*][a-zA-Z0-9_\*]{0,100})(/(\*|[a-zA-Z0-9_\.\-\*]{1,199})){0,1}(/(\*|[A-Z\-\*]{1,99})(#(\*|[a-zA-Z0-9\.\-_\*]){0,100}){0,1}){0,1}$
2021/10/11 15:09:33 logical_file_name=/store/user/one/two/three/file123.root matched with ^/store/([A-Za-z0-9._\-/])+/([A-Za-z0-9._\-])+(.root|.lhe(.xz){0,1})$
2021/10/11 15:09:33 physics_group=123 matched with ^([a-zA-Z0-9\-_]+)$
2021/10/11 15:09:33 primary_ds_name=123-bla-123 matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:33 processed_ds_name=123-bla-123 matched with (^[a-zA-Z0-9]+[a-zA-Z0-9_\-]*)$
2021/10/11 15:09:33 processed_ds_name=123-bla-123 matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:33 data_tier_name=RAW matched with ^([A-Z]{3,100})+$
2021/10/11 15:09:33 acquisition_era_name=someEra matched with [a-zA-Z][a-zA-Z0-9_]*$
2021/10/11 15:09:33 processing_version=123 matched with ^[0-9]+$
2021/10/11 15:09:33 global_tag=123 matched with ^[a-zA-Z0-9\s\.\-_:]{1,100}$
2021/10/11 15:09:33 global_tag=123 matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:33 create_by=123bak matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:33 create_by=123bak matched with ^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_\']*$
2021/10/11 15:09:33 last_modified_by=123blka matched with ^[a-zA-Z0-9/%*][a-zA-Z0-9/\.\-_%*/#]*$
2021/10/11 15:09:33 last_modified_by=123blka matched with ^[a-zA-Z0-9/][a-zA-Z0-9/\.\-_\']*$
2021/10/11 15:09:33 performed positive lexicon test against ../static/lexicon_reader_positive.json
--- PASS: TestLexiconPositive (0.01s)
PASS
ok  	github.com/vkuznet/dbs2go/test	0.439s
cd test && LD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DYLD_LIBRARY_PATH=`cat /Users/vk/Work/Languages/Go/gopath/src/github.com/vkuznet/dbs2go/oci8.pc | grep "libdir=" | sed -e "s,libdir=,,"` DBS_LEXICON_FILE=../static/lexicon_reader.json DBS_LEXICON_SAMPLE_FILE=../static/lexicon_reader_negative.json go test -v -run LexiconNegative
=== RUN   TestLexiconNegative
2021/10/11 15:09:36 performed negative lexicon test against ../static/lexicon_reader_negative.json
--- PASS: TestLexiconNegative (0.01s)
PASS
ok  	github.com/vkuznet/dbs2go/test	0.347s
```

amaltaro:
@belforte @vkuznet thanks for providing all these details. I think it will take a couple of weeks before I can resume working on this. Once that happens, I will read carefully your messages and get back to you in case of questions.

vkuznet:
@amaltaro please revisit your review before end of the year that we can plan and adjust things.

klannon:
@vkuznet: Alan is currently on Holiday through middle of next week.  I also know that he has a fairly busy rest of the month coming up for personal reasons.  Is it necessary to finish this in 2021 or can it be deferred until early 2022?

vkuznet:
@klannon it is not strictly necessary, but Alan mentioned the review in couple of weeks more than a month ago. Said that, due to lack of response I proceed with my version of lexicon schema in DBS Go implementation and it means that eventually we'll diverse on this subject between different systems which I would like to avoid. Therefore, I don't want to post-pone this issue for too long.

klannon:
After a quick chat with @vkuznet, added this to the 2022 Q1 to-do list so that we don't lose track of it, since it seems unlikely that we'll manage with the end of year craziness and other Q4 2021 priorities.

amaltaro:
Valentin, given that DBS already has its own lexicon logic/format, and you mentioned this no longer to be a critical piece for DBS, I'd suggest to work on this in a low priority this quarter. I'd like to proceed with the remaining implementation myself, but if I do that, it won't happen before march or so.

belforte:
@amaltaro drop me a line when Lexicon.py changes, in case I need to change the way CRAB uses it.

PR #10635, opened 2021-06-28
[py2py3] Fix AgentStatusWatcher subprocess.Popen calls 

amaltaro:
Fixes #10634 

#### Status
ready

#### Description
In python3, subprocess.Popen returns a byte-like object. With this PR, we convert that output to unicode and then do the required string parsing.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #10639, opened 2021-06-28
DBFormatter formatDict/formatList to native string instead of bytes

### LINKED ISSUE #10638: [py2py3] Make DBFormatter return native strings instead of bytes
**Impact of the bug**
Mostly WMAgent (and externals)

**Describe the bug**
Some places in WMCore actually expect a python3 string data type, instead of bytes, as it's been reported in this comment:
https://github.com/dmwm/WMCore/issues/10634#issuecomment-869847004

StompAMQ was also failing in AgentStatusWatcher with:
```
2021-06-28 18:20:05,645:140147165435648:ERROR:StompAMQ:Notification: {'type': 'cms_wmagent_info', 'version': '0.3', 'producer': 'wmagent'} (type=<class 'dict'>) not send, error: Object of type bytes is not JSON serializable
```

So, I believe it would be better to actually format the DAOs results to native string instead of encoding its output to bytes.

**How to reproduce it**
Run WMAgent Python3 `1.5.0.pre3`, with some patches on top of it to allow components to be started.

**Expected behavior**
Instead of casting strings/bytes to byte strings in the DBFormatter `formatDict` and `formatList`, I'd suggest to actually decode the byte/string output to unicode (native string in python2 and python3).
It fixes at least the problem reported here: https://github.com/dmwm/WMCore/pull/10635#issuecomment-869897102

I think the same change should actually get extended to the other methods formatting DAOs results. But since we only have the current bytes casting in those two methods, I think it's a good start.

None

amaltaro:
Fixes #10638 

#### Status
ready (tested in WMA Py3)

#### Description
When formatting the output of the relational database queries, decode bytes to native strings in python3. At the same time, stop encoding python2 native strings to bytes.

yes, hopefully!

#### Related PRs
Cherry-picked one of the commits from https://github.com/dmwm/WMCore/pull/10635

None

### DIFF
--- a/src/python/WMCore/Database/DBFormatter.py
+++ b/src/python/WMCore/Database/DBFormatter.py
@@ -6,14 +6,15 @@ Holds a bunch of helper methods to format input and output of sql
 interactions.
 """
 
-from builtins import str, zip, range
+from builtins import str, bytes, zip, range
 
 import datetime
 import time
 import types
 
+from Utils.Utilities import decodeBytesToUnicodeConditional
 from WMCore.DataStructs.WMObject import WMObject
-from Utils.PythonVersion import PY2
+from Utils.PythonVersion import PY3
 
 
 class DBFormatter(WMObject):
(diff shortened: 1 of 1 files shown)


amaltaro:
Dario, we already had a chat on this subject and we do not see any blocker. So I'm merging it now, but please do leave any comments/concerns that you might have.

mapellidario (commented):
Hello @amaltaro! I am super sorry for not having review this before. 

First of all, thanks for finding the courage for making these changes! These lines always left me with an unsatisfactory feeling, but I never managed to properly improve them.

I made some comments along the code. It is super-possible that i mis-judged your intentions and that I am missing some precious information about what you are trying to achieve here. Maybe we should also consider spending 30min improving `test/python/WMCore_t/Database_t/DBFormatter_t.py:DBFormatterTest.testBFormatting` to make sure we are covering all the possible cases. Furthermore, I think that updating the unittest to reflect what we are trying to solve here is worth more than every _a priori_ discussion.

What do you think? Have you already thoroughly tested these changes and found that everything is still working in py2?

PR #10640, opened 2021-06-28
Add RucioConMon API to WMCore.Services.

### LINKED ISSUE #10410: MSUnmerged: create get methods for fetching unmerged files from the CC tool

todor-ivanov:
Fixes #10410 

#### Status
Ready

#### Description
This PR is intended to provide all the needed interfaces to external systems so that the `MSUnmerged` micro service may fetch the data for:
 * Statistics for all RSEs from `Rucio Consistency Monitor`, together with update timestamps
 * List of files in unmerged area per RSE
 * Lists of protected files from the RSE's unmerged area
 * Default delete interface for the RSE
 * etc.
  
YES

#### Related PRs
We need to put the Url to the Rucio Consistency Monitor in the service config:

<TODO - to put the links to the `service_configs` and deployment repositories here>


src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+# coding=utf-8
+"""
+Rucio Consistency Monitor Service class developed on top of the native WMCore
+Service APIs providing custom output and handling as necessary for the
+CMS Workload Management system
+"""
+
+from __future__ import division, print_function, absolute_import
+
+from urllib.parse import urlencode
```
amaltaro, line 11: Since our code is meant to be dual stack, I'd suggest to have this import fixed with the future standard_library.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        Init method for the RucioConMon Class
+        """
+        configDict = configDict or {}
+        configDict.setdefault('endpoint', url)
+        configDict.setdefault('cacheduration', 1)  # in hours
+        configDict.setdefault('accept_type', 'application/json')
+        configDict.setdefault('content_type', 'application/json')
+        configDict['logger'] = logger if logger else logging.getLogger()
+        super(RucioConMon, self).__init__(configDict)
```
amaltaro, line 34: With this super `__init__` at this position, you can be potentially overriding all the defaults and configurations set above. Can you please move it to one of the first lines in this method, and only then define this class extra arguments?

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        self['logger'].debug("Initializing RucioConMon with url: %s", self['endpoint'])
+
+    def _getResult(self, uri, callname="", clearCache=True, args=None):
```
amaltaro, line 37: Shouldn't we actually only clear the cache upon explicit request? If it's been expired, it should anyways make a new call to the web-service.

You might as well improve this docstring a little bit. For instance, what args data type is supposed to be?

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        :return:    A dictionary
+        """
+
+        # NOTE: Unlike the common case we are not using the callname for building
+        #       apiUrl. we  are using it only for giving proper name of the cache
+        #       file.
+        # NOTE: The 'callname' should not contain '/', otherwise the cache is
+        #       tried to be created with a a non existing subdirectory structure
+
+        cachedApi = "%s.json" % callname
```
amaltaro, line 50: Shouldn't this cache name also consider the arguments? Such that a call to `unmerged&rse=SiteA` doesn't hit the same cache as of `unmerged&rse=SiteB`?

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        results = json.loads(results)
+        return results
+
+    def _getResultZipped(self, uri, callname="", clearCache=True, args=None):
+        """
+        This method is retrieving a zipped file from the uri privided, instead
+        of the normal json
+        :param uri: The endpoint uri
+        :return:    A dictionary
+        """
```
amaltaro, line 73: How about defining a `raise NotImplemented` or something like that? Please check the documentation, I do not remember whether that is the correct exception name.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        :return:    A list of unmerged files for the RSE in question
+        """
+        # NOTE: The default API provided by Rucio Consistency Monitor is in a form of a
+        #       zipped file/stream. Currently we are using the newly provided json API
+        #       But in in case we figure out the data is too big we may need to
+        #       implement the method with the zipped API and use disc cache for
+        #       reading/streaming from file. This will prevent any set arithmetic
+        #       in the future.
+        if not zipped:
+            uri = "WM/files?rse=%s&format=json" % rseName
```
amaltaro, line 99: I would suggest to define the format of the output data to be defined in the `__init__` method, such that it's an object variable. I thing the PhEDEx Service wrapper has something similar to that.

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
```diff
@@ -124,3 +124,14 @@ def getChildParentDatasetMap(self, requestType="StepChain", parentageResolved=Fa
                         for childDS in childParentDS["ChildDsets"]:
                             childParentMap[childDS] = childParentDS["ParentDset"]
         return childParentMap
+
+    def getProtectedLFNs(self):
+        """
+        _getProtectedLFNs_
+
+        A method to be used for fetching a list of all protected lfns from WMStatServer
+        :param unmergedOnly: If True return only the lfns from the unmerged area
```
amaltaro, line 133: This docstring line is invalid. I would also suggest to remove the extra spacing from the line below.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        :return:          dict of the result
+        """
+        uri = "/WM/stats"
+        rseStats = self._getResult(uri, callname='stats')
+        return rseStats
+
+    def getRSEUnmerged(self, rseName, zipped=False):
+        """
+        Gets the list of all unmerged files in an RSE
+        :param rse: The RSE whose list of unmerged files to be retrieved
```
amaltaro, line 89: Wrong docstring.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,110 @@
...
+        of the normal json
+        :param uri: The endpoint uri
+        :return:    A dictionary
+        """
+
+    def getRSEStats(self):
+        """
+        Gets the latest statistics from the RucioConMon, together with the last
+        update timestamps for all RSEs known to CMS Rucio
+        :param callname:  name of the call
```
amaltaro, line 79: Wrong docstring.

amaltaro (changes requested):
Todor, I left some comments along the code.

In addition to those, can you please clear the pycodestyle warning provided in the report (at least for the brand new files that you have been providing)?

I also think it would be positive to have some basic unit tests for some of the methods you have provided here. Something simple, just to have some coverage without taking too much effort to get implemented.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: I would disagree, though.
This `zipped` interface is a thing that is specific and provided only for this `uri` and nowhere else  in the Rucio Connection Monitor. The reason for that is quite trivial - the files are simply created in .zip format. This is not a real compressed stream. So having it as an object variable would be not only misleading, but also a source of errors.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: This is how it is defined in CRIC Service [1]. So if it is wrong her .. well then it is wrong there too.
[1]
https://github.com/dmwm/WMCore/blame/eeb8ab2c7083d95a1dc50879d48287f508b33709/src/python/WMCore/Services/CRIC/CRIC.py#L35-L47

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -115,33 +150,27 @@ def _execute(self, rseList):
...
-        totalNumRses = 0
-        totalNumFiles = 0
-        numRsesCleaned = 0
-        numFilesDeleted = 0
 
-        # Call the workflow dispatcher:
-        for rse in rseList:
+        pline = self.plineUnmerged
+        self.resetCounters(plineName=pline.name)
+        self.plineCounters[pline.name]['totalNumRses'] = len(rseList)
```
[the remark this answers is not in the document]
todor-ivanov: It is not that we have multiple pipelines, it is a patter that I was following. One has two major types of objects driving the service and each one has its counters - one is the RSE the other is the pipeline. The level of hierarchy in both cases is the same. The same also with the timers. It would be far from trivial to change this structure at this stage, because those are updated from several places, and the debugging would take quite a bit of time which we do not have.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -173,20 +202,146 @@ def _checkClean(self, rse):
...
+        if rseName not in self.rseTimestamps:
+            self.rseTimestamps[rseName] = {'prevStartTime': 0.0,
+                                           'startTime': 0.0,
+                                           'prevEndtime': 0.0,
+                                           'endTime': 0.0}
+        if start:
+            self.rseTimestamps[rseName]['prevStartTime'] = self.rseTimestamps[rseName]['startTime']
+            self.rseTimestamps[rseName]['startTime'] = currTime
+        if end:
+            self.rseTimestamps[rseName]['prevEndtime'] = self.rseTimestamps[rseName]['endtime'] = currTime
```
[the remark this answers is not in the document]
todor-ivanov: Good catch. Fixing it now.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: fixed

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: ok

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: Same as above.
https://github.com/dmwm/WMCore/blob/eeb8ab2c7083d95a1dc50879d48287f508b33709/src/python/WMCore/Services/CRIC/CRIC.py#L56

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: That's a good idea! doing it now.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: done

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: Done

todor-ivanov:
Thanks  for the review @amaltaro  I addressed your comments. Please take another look!

src/python/WMCore/Services/RucioConMon/RucioConMon.py
amaltaro: That's a good point! You're right, Todor.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
amaltaro: That's what happens when you review something on Saturday 1am... I thought the setdefault was on the self object, not on configDict.
You're correct, Todor. It's good like this.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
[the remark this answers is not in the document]
amaltaro: Yes, the RSE counter does make sense to be like that.
The pipeline one, is just an extra complication and making the code harder to read and maintain, given that there are no multiple pipelines and the service counters are always going to be in a flat structure.

I agree though not to make this change, it can potentially insert a bug if not carefully done and tested.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
amaltaro: It's not impossible that they are both wrong. This can probably be verified with unit tests, just to make sure we are not using the wrong cache file. Can you double check it in the next week as you work in the next PRs, Todor?

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -12,23 +14,40 @@
...
 from WMCore.MicroService.MSCore import MSCore
 from WMCore.MicroService.MSUnmerged.MSUnmergedRSE import MSUnmergedRSE
+from WMCore.Services.RucioConMon.RucioConMon import RucioConMon
+from WMCore.Services.WMStatsServer.WMStatsServer import WMStatsServer
 # from WMCore.Services.AlertManager.AlertManagerAPI import AlertManagerAPI
 from WMCore.WMException import WMException
 from Utils.Pipeline import Pipeline, Functor
 
+import random
+from future import standard_library
```
amaltaro, line 30: This standard_library change was added to the wrong module. It should have been on the one importing urllib.parse. And please make sure it's right below the `__future__` import.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
amaltaro: Not yet, see my other comment please.

amaltaro (changes requested):
Thanks for the prompt fixes, Todor. There seems to be one missing piece before we can merge it, I left a comment in the code.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: So far the `callname` is used as a unique identifier to the cache name and I am sure we are not overlapping them since I constructed them. For the time being I think this is a good enough.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: fixed

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: fixed

todor-ivanov:
Thank you for your previous review @amaltaro. Can you please take another look.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
```diff
@@ -0,0 +1,113 @@
...
+from __future__ import division, print_function, absolute_import
+from future import standard_library
+
+from urllib.parse import urlencode
+
+import json
+import logging
+
+from WMCore.Services.Service import Service
+standard_library.install_aliases()
```
amaltaro, line 18: I believe the order of these imports will matter. So this `install_aliases()` line should go right in the next line after you import it (and before you import urllib.parse).

amaltaro (changes requested):
Todor, I left another comment along the code that needs to be taken care of.
And since you will have to make yet another change to this PR, please fix the `MSUnmergedRSE.py` pycodestyle report (as just mention over slack as well).

Thanks for providing these fixes.

src/python/WMCore/Services/RucioConMon/RucioConMon.py
todor-ivanov: > So this install_aliases() line should go right in the next line after you import it (and before you import urllib.parse)

I'll fix it as you say, and this is how it was until earlier, but just to warn you - this will break another pylint requirement : 
> E402 module level import not at top of file

amaltaro: The only way to work around that one would be disabling it, but I usually don't disable any pylint rule. So it's fine like this.

amaltaro:
@todor-ivanov please squash these commits and it's ready to go. Thanks!

amaltaro:
To avoid wasting more time on this, I'm merging it like this.
Thanks!

PR #10628, opened 2021-06-28
Force pickle protocol=0 for the spec creation to be compatible with Py2/Py3

amaltaro:
Superseeds https://github.com/dmwm/WMCore/pull/10619 (please see whole discussion/investigation in this PR)
Fixes #10609 

#### Status
Ready

#### Description
As discussed with Dario this morning, trying a variation of the current implementation in #10619
If it passes the unit tests, then we can run some real tests within WMCore.

NOTE: this patch is required for the Python2 `1.4.7.patchX` agents:
```
cmst1@vocms0262:/data/srv/wmagent/current $ diff -rup /data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMCore/BossAir/Plugins/SimpleCondorPlugin.py{.bkp,}
--- /data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMCore/BossAir/Plugins/SimpleCondorPlugin.py.bkp	2021-06-24 19:30:23.860759021 +0200
+++ /data/srv/wmagent/v1.4.7.patch3/sw/slc7_amd64_gcc630/cms/wmagent/1.4.7.patch3/lib/python2.7/site-packages/WMCore/BossAir/Plugins/SimpleCondorPlugin.py	2021-06-24 19:31:50.394372378 +0200
@@ -605,7 +605,7 @@ class SimpleCondorPlugin(BasePlugin):
             requiredOSes = self.scramArchtoRequiredOS(job.get('scramArch'))
             ad['My.REQUIRED_OS'] = classad.quote(requiredOSes)
             cmsswVersions = ','.join(job.get('swVersion'))
-            ad['My.CMSSW_Versions'] = classad.quote(cmsswVersions)
+            ad['My.CMSSW_Versions'] = classad.quote(str(cmsswVersions))
```

#### Is it backward compatible (if not, which system it affects?)
yes, hopefully!

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #10644, opened 2021-06-29
WMAgent 1.4.7 patch for workflow specs made in python3

amaltaro:
Does not fix any issue, but it's required for https://github.com/dmwm/WMCore/pull/10628

#### Status
not-tested

#### Description
Even though we kept the same protocol=0 for the workload spec pickling, there seems to have still some differences between a Python2 vs Python3 made spec pickle (classical issues with bytes / strings / unicode).

So, this `1.4.7_wmagent` branch patch is required before we upgrade central services to Python3. I expect it to be "harmless" for the current python2-made specs that we have in the system; while it would be a must for specs still to be created by python3 services.

UPDATE: there are fixes for the DQMSaver and also chainedProcessing (StepChain), where Python3-made spec files come with unicode strings instead of byte strings.

NOTE: we cannot yet say whether this same fix is required against the master branch.

yes

#### Related PRs
None

Production agents must be patched before #10628 gets deployed in CMSWEB.

### DIFF
--- a/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
+++ b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
@@ -16,7 +16,7 @@ import socket
 import re
 
 import FWCore.ParameterSet.Config as cms
-
+from Utils.Utilities import encodeUnicodeToBytes
 from PSetTweaks.PSetTweak import PSetTweak
 from PSetTweaks.WMTweak import applyTweak, makeJobTweak, makeOutputTweak, makeTaskTweak, resizeResources
 from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig
@@ -353,14 +353,14 @@ class SetupCMSSWPset(ScriptInterface):
         # first, create an instance of TrivialFileCatalog to override
         tfc = TrivialFileCatalog()
         # check the jobs input files
-        inputFile = ("../%s/%s.root" % (self.step.data.input.inputStepName,
-                                        self.step.data.input.inputOutputModule))
+        inputFile = "../%s/%s.root" % (self.step.data.input.inputStepName,
+                                       self.step.data.input.inputOutputModule)
         tfc.addMapping("direct", inputFile, inputFile, mapping_type="lfn-to-pfn")
         tfc.addMapping("direct", inputFile, inputFile, mapping_type="pfn-to-lfn")
 
         fixupFileNames(self.process)
         fixupMaxEvents(self.process)
-        self.process.source.fileNames.setValue([inputFile])
+        self.process.source.fileNames.setValue([encodeUnicodeToBytes(inputFile)])
         self.process.maxEvents.input.setValue(-1)
 
         tfcName = "override_catalog.xml"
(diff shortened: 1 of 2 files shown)


src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py
```diff
@@ -638,7 +638,7 @@ def handleCondorStatusService(self):
         if isCMSSWSupported(self.getCmsswVersion(), "CMSSW_7_6_0"):
             self.logger.info("Tag chirp updates from CMSSW with step %s", self.step.data._internal_name)
             self.process.add_(cms.Service("CondorStatusService",
-                                          tag=cms.untracked.string("_%s_" % self.step.data._internal_name)))
+                                          tag=cms.untracked.string("_%s_" % encodeUnicodeToBytes(self.step.data._internal_name))))
```
mapellidario, line 641: @amaltaro are you 100% sure of this change? because in py3 it would actually add `b''` to the string

```plaintext
> python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
>>> d = "hello world"
>>> "_%s_" % d.encode()
"_b'hello world'_"
```

mapellidario: If this was intended to remove the `u''` in py2, then we may use `encodeUnicodeToBytesConditional(..., condition=PY2)`

amaltaro: This patch is not being provided against the master branch. It's only a patch for the current - quite old - agents that we have in production. We do not even have the `encodeUnicodeToBytesConditional` function on this branch :-D

so, I believe it's safe enough. I did not manage to go that far in the Python3 WMAgent, so I do not know right now whether something "similar" would be required against master.

mapellidario: Uh! you are correct! If this is actually only intended to be run in py2, It's safe! thanks Alan! Sorry for bugging :)

amaltaro:
DQMSaver code also had problems with unicode string being provided as input, so I'm encoding it to bytes as well (new 2nd commit).

amaltaro:
...this is worrisome... 3rd commit has unicode to bytes fix for the chained processing PSet setup.

Exception message was:
```
RuntimeError: Error invoking script for step WMTaskSpace.cmsRun4
withe Script module: SetupCMSSWPset
wrong type being extended to container vstringDetails:
  File "/srv/job/WMCore.zip/WMCore/WMRuntime/ScriptInvoke.py", line 110, in <module>
  File "/srv/job/WMCore.zip/WMCore/WMRuntime/ScriptInvoke.py", line 81, in invoke
  File "/srv/job/WMCore.zip/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py", line 783, in __call__
    self.handleChainedProcessing()
  File "/srv/job/WMCore.zip/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py", line 363, in handleChainedProcessing
    self.process.source.fileNames.setValue([inputFile])
```

amaltaro:
This patch has been applied to all the agents connected to the production system (and perhaps to all the testbed agents as well).

amaltaro:
[Fix #10644 - WMAgent 1.4.7 patch for workflow specs made in python3](https://trello.com/c/yjS7puA1/257-fix-10644-wmagent-147-patch-for-workflow-specs-made-in-python3)

PR #10650, opened 2021-06-30
Fixing unittests in WMCore_t/Services_t/pycurlFileUpload_t.py and WebTools_t/REST_t.py

mapellidario:
Related to #10422 

#### Status
ready

#### Description
Fixing unittests in 

- [x] test/python/WMCore_t/Services_t/pycurlFileUpload_t.py
  - ACHTUNG: in py3 we get bytes back, not unicode
  - [x] the response contains bytes, not unicode. is this going to be a problem for us? (we should discuss). -> updated docstring
- [x] test/python/WMCore_t/BossAir_t/MockPlugin_t.py
- [x] test/python/WMCore_t/WebTools_t/REST_t.py
  - [x] testGoodEcho
  - [x] testDAOBased
  - [x] testException
    - caused by https://github.com/dmwm/WMCore/blob/8c73e7073fd8928326884e3d32fd9c806dcb7429/src/python/WMCore/WebTools/RESTApi.py#L147 not picking https://github.com/dmwm/WMCore/blob/8c73e7073fd8928326884e3d32fd9c806dcb7429/src/python/WMCore/WebTools/RESTFormatter.py#L25, but picking xml instead
    - breakpoints: RESTApi.py: L115, L150. RESTFormatter.py L92, RESTModel.py L285.
  - [x] testList 
    - (simply undoing some of the changes made in debugging)
  - [x] testListTypeArgs
    - ACHTUNG: in py3 we get bytes back, not unicode 
    - [x] the response contains bytes, not unicode. is this going to be a problem for us? (we should discuss) -> updated docstring
  - [x] testSanitisePass
  - [x] testSanitisePassHTTP
  - [x] testPing: i run it 100 times on my machine and every time it passed. removing from unstable list
- [x] test/python/WMCore_t/WebTools_t/RESTFormat_t.py
  - [x] WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest.testGenerator
  - [x] WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest.testNoArgMethods
    - [ ] hacky solution, but real solution would require much moer effort and i am not sure it is worth the time right now
  - [x] WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest.testReturnFormat
    - [ ] hacky solution, but real solution would require much moer effort and i am not sure it is worth the time right now

yes

#### Related PRs
nope

nope

### DIFF
--- a/src/python/WMCore/WebTools/RESTFormatter.py
+++ b/src/python/WMCore/WebTools/RESTFormatter.py
@@ -20,9 +20,10 @@ from WMCore.Wrappers.JsonWrapper.JSONThunker import JSONThunker
 
 class RESTFormatter(TemplatedPage):
     def __init__(self, config):
-        self.supporttypes = {'application/xml': self.xml,
+        self.supporttypes = {
+                    'text/json': self.json,
+                    'application/xml': self.xml,
                    'application/atom+xml': self.atom,
-                   'text/json': self.json,
                    'text/x-json': self.json,
                    'application/json': self.json,
                    'text/html': self.to_string,
--- a/test/python/WMCore_t/Services_t/pycurlFileUpload_t.py
+++ b/test/python/WMCore_t/Services_t/pycurlFileUpload_t.py
@@ -44,7 +44,7 @@ class PyCurlRESTServer(RESTBaseUnitTest):
         self.assertEqual(open(uploadedFilename).read(), open(fileName).read())
         # delete the uploaded file
         os.remove(uploadedFilename)
-        self.assertTrue('Success' in res)
+        self.assertTrue(b'Success' in res)
 
     def testFailingFileUpload(self):
         """
--- a/test/python/WMCore_t/WebTools_t/RESTFormat_t.py
+++ b/test/python/WMCore_t/WebTools_t/RESTFormat_t.py
@@ -18,6 +18,8 @@ from WMQuality.WebTools.RESTBaseUnitTest import RESTBaseUnitTest
 from WMQuality.WebTools.RESTClientAPI import methodTest
 from WMQuality.WebTools.RESTServerSetup import DefaultConfig
 
+from Utils.PythonVersion import PY3
+
 class RESTFormatTest(RESTBaseUnitTest):
 
     def initialize(self):
@@ -74,7 +76,9 @@ class RESTFormatTest(RESTBaseUnitTest):
                          output={'code':200, 'data':'{"a": "a%", "b": "b"}'})
 
         url = self.urlbase + 'list?input_int=a&input_str=a'
-        expected_data = '''{"exception": 400, "message": "Invalid input: Input data failed validation.", "type": "HTTPError"}'''
+        expected_data_py2 = '{"exception": 400, "message": "Invalid input: Input data failed validation.", "type": "HTTPError"}'
+        expected_data_py3 = '{"exception": 400, "type": "HTTPError", "message": "Invalid input: Input data failed validation."}'
+        expected_data = expected_data_py3 if PY3 else expected_data_py2
         methodTest('GET', url, accept=return_type,
                          output={'code':400, 'data':expected_data})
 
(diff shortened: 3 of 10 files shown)


src/python/WMCore/WebTools/RESTFormatter.py
```diff
@@ -20,9 +20,10 @@
 
 class RESTFormatter(TemplatedPage):
     def __init__(self, config):
-        self.supporttypes = {'application/xml': self.xml,
+        self.supporttypes = {
+                    'text/json': self.json,
+                    'application/xml': self.xml,
```
mapellidario, line 25: the order here matters! otherwise `testException` fails in py3

test/python/WMCore_t/WebTools_t/RESTFormat_t.py
```diff
@@ -74,7 +76,9 @@ def testReturnFormat(self):
                          output={'code':200, 'data':'{"a": "a%", "b": "b"}'})
 
         url = self.urlbase + 'list?input_int=a&input_str=a'
-        expected_data = '''{"exception": 400, "message": "Invalid input: Input data failed validation.", "type": "HTTPError"}'''
+        expected_data_py2 = '{"exception": 400, "message": "Invalid input: Input data failed validation.", "type": "HTTPError"}'
+        expected_data_py3 = '{"exception": 400, "type": "HTTPError", "message": "Invalid input: Input data failed validation."}'
+        expected_data = expected_data_py3 if PY3 else expected_data_py2
         methodTest('GET', url, accept=return_type,
                          output={'code':400, 'data':expected_data})
```
mapellidario, line 83: I know this is not an elegant solution, but having a proper fix would require either a significant change of `methodTest`, ar writing a new `methodTest`. Would it be worth it?

test/python/WMCore_t/Services_t/pycurlFileUpload_t.py
```diff
@@ -44,7 +44,7 @@ def testFileUpload(self):
         self.assertEqual(open(uploadedFilename).read(), open(fileName).read())
         # delete the uploaded file
         os.remove(uploadedFilename)
-        self.assertTrue('Success' in res)
+        self.assertTrue(b'Success' in res)
```
mapellidario, line 47: This means that there is a change in the interface. `self.requestHandler.uploadFile` returns a list of bytes string in both py2 and py3 and not a list of "native" string (bytes in py2, uniode in py3)

mapellidario: Nothing to discuss really, just something that we should be aware of and maybe report somewhere. in the function docstring? in the PR description?

test/python/WMCore_t/WebTools_t/REST_t.py
```diff
@@ -360,7 +360,7 @@ def testListTypeArgs(self):
         # 2 positional args (e.g. url/arg1/arg2)
         url = self.urlbase + 'listTypeArgs?aList=1'
         response = makeRequest(url=url)
-        assert response[1] == 200 and response[0] == "[1]", \
+        assert response[1] == 200 and response[0] == b"[1]", \
```
mapellidario, line 363: This means that there is a change in the interface. `makeRequest` returns a list whose first position contains a bytes string in both py2 and py3 and not a "native" string (bytes in py2, uniode in py3).

Nothing to discuss really, just something that we should be aware of and maybe report somewhere. in the function docstring? in the PR description?

test/python/WMCore_t/Services_t/pycurlFileUpload_t.py
amaltaro: Yes, it's probably a good idea to mention it in the method docstring. Can you please do so?

amaltaro (approved):
Changes look good to me, thanks Dario! I have made only one comment which might be interesting to add an extra comment to the docstring, but it can be skipped as well.

I'm concerned about these methods returning byte strings in both py2 and py3. I wish python3 would be returning native string... but I fail to think what would be our best options here.

Feel free to squash those commits accordingly. Thanks

test/python/WMCore_t/Services_t/pycurlFileUpload_t.py
mapellidario: thanks! added to the docstring!

test/python/WMCore_t/WebTools_t/REST_t.py
mapellidario: updated the docstring

mapellidario:
I squashed the commits!

Just a quick comment: i updated the docstrings, for example in https://github.com/dmwm/WMCore/blob/4c4d2472d05fc4ec4ab4fab49ba2da4ddee1cb3f/src/python/WMCore/Services/Requests.py#L464 and by having a quick look at that method it is clear that the return value is in bytes so maybe changing the docstring is a bit redundant. We are getting the value of a `BytesIO` stream. If we want to change it, we can. But there is always my conservative side repeating "if aint broke dont fix it".

amaltaro:
>  and by having a quick look at that method it is clear that the return value is in bytes so maybe changing the docstring is a bit redundant. We are getting the value of a BytesIO stream. If we want to change it, we can. But there is always my conservative side repeating "if aint broke dont fix it".

Hehe, indeed! I think it's safe to wait for the python2 code to be gone from WMCore, then it will likely be easier to start tackling all this bytes/string/unicode war.

Thanks for squashing these!

PR #10665, opened 2021-07-06
Fix broken variable reference in TestMonitor

### LINKED ISSUE #10448: Variable `file` out of scope - pylint complains - not critical
**Impact of the bug**
`src/python/WMCore/WMRuntime/Monitors/TestMonitor.py`, so i assume this is not critical.

**Describe the bug**
While running `pylint --py3k` working on #10440, pylint complained about the line

https://github.com/dmwm/WMCore/blob/dd63a901a9b929d83cbe83e9a2f9004e0901a90c/src/python/WMCore/WMRuntime/Monitors/TestMonitor.py#L155

with the warning `file built-in referenced (file-builtin)`. However, (as far as i understand) here we are using the variable `file` out of scope (it is not defined in the class `TestMonitor`), so pylint thinks we are using the python2 reserved keyword `file` (which is not a reserved keyword anymore in py3) and complains.

I propose to define the variable file as a `TestMonitor` class member, and maybe also change its name to a non-reserved keyword while we are at it. Maybe `self.cmsswFile` is what we should use in that line instead of `file`?

**How to reproduce it**
Run
```bash
pylint --py3k -d W1618,W1619 src/python/WMCore/WMRuntime/Monitors/TestMonitor.py
```

**Expected behavior**
We would like pylint not to complain (not too much, at least)

amaltaro:
Fixes #10448 

#### Status
ready

#### Description
The issue itself is well explained. This fixes a broken reference to a non-existent variable (or out of scope). The logical thing to me is to simply call it as provided in this PR.

Note, that this code is likely not being exercised, otherwise it would have broken long ago.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WMRuntime/Monitors/TestMonitor.py
+++ b/src/python/WMCore/WMRuntime/Monitors/TestMonitor.py
@@ -152,7 +152,7 @@ class TestMonitor(WMRuntimeMonitor):
 
         # Check for events
         if self.cmsswFile:
-            run, event = searchForEvent(file)
+            run, event = searchForEvent(self.cmsswFile)
             if run and event:
                 # Then we actually found something, otherwise do nothing
                 # Right now I don't know what to do


mapellidario (approved):
Aha! So my instincts were good! Thank you very much Alan for checking and fixing this! :)

amaltaro:
You gave this issue "for free" to me, I couldn't resist getting it fixed ;-) Thanks for the review, Dario.

PR #10666, opened 2021-07-06
Add logic for cutting the path depth as provided by RucioConMon

### LINKED ISSUE #10657: MSUnmerged: Create unmerged files fitering logic
**Impact of the new feature**
MSUnmerged

In the process of creating the MSUnmerged micro service, we figured out we cannot benefit from python set operations in the filtering out the protected LFNs (fetched from wmstats) fromf the list of unmerged LFNs fetched from the Rucio Consistency Monitor per RSE. The reason is the different depth in the LFN tree at which the two services keep information:

WMStats:
```
/store/unmerged/Run2016B/JetHT/MINIAOD/ver2_HIPM_UL2016_MiniAODv2-v2
```

Rucio Consistency Monitor:
```
/store/unmerged/Run2016B/JetHT/MINIAOD/ver2_HIPM_UL2016_MiniAODv2-v2/140000/388E3DEF-9F15-D04C-B582-7DD036D9DD33.root
```

 So we will have to iterate through the whole list item by item.

A simple but fast enough method to search through those lists. In preference with the smallest memory footprint possible, because of the enormous amount of information we are about to manipulate. Just for reference: The records for the unmerged present only at one site -  T2_US_Wisconsin:
```
$ ll -h MSUnmerged.log.T2_US_Wisconsin
28M Jul  2 11:00 MSUnmerged.log.T2_US_Wisconsin
 
$ wc -l MSUnmerged.log.T2_US_Wisconsin
143283 MSUnmerged.log.T2_US_Wisconsin
```
`*` There is some minor % of these records coming from script debugging information but it is far less than 1% 

**Additional context**

todor-ivanov:
Fixes #10657 

#### Status
Ready

#### Description
In the current PR we propose the initial implementation of the logic for filtering all the paths provided by Rucio Consistency Monitor in a way that they become suitable for  set operations in order to apply the protected LFNs coming  from the WMStats interface.

YES

#### Related PRs
NO

NO


src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -95,6 +97,9 @@ def __init__(self, msConfig, logger=None):
         self.rseConsStats = {}
         self.protectedLFNs = []
 
+        # The basic /store/unmerged regular expression:
+        self.regStoreUnmerged = re.compile("^/store/unmerged/.*$")
```
amaltaro, line 101: Should we ensure that the path to be deleted will always match a given depth in the directory structure?
E.g., something like `re.compile("^/store/unmerged/*/*/*/.*$")`?

todor-ivanov: I'd rather say no. Because there are stuff that are recorded like SAM tests outputs and so on which are certainly not reaching the depth to what we'd define here... and even from our workflows [1]. In the expression you suggest I suspect are mixed wildcards + regexp, but anyway it can be worked out of course I've got the idea. And actually initially I did try with the following regular expression `r'^/store/unmerged/.+/.+/.+/.+$'` which should be the one expressing the depth correctly, but if yout think twice it is exactly equivalent to just this one:  `'^/store/unmerged/.+$'. So things more or less lead to the generic one. For dealing with the path depth there is the dictionary slicing `[:6]` down the code, which should always provide us the maximum depth to be reached. 


[1]
```
'/store/unmerged/Phase2HLTTDRSummer20ReRECOMiniAOD/DYToLL_M-50_TuneCP5_14TeV-pythia8/FEVT',
'/store/unmerged/Phase2Spring21DRMiniAOD/TTbar_TuneCP5_14TeV-pythia8/FEVTDEBUGHLT',
'/store/unmerged/Phase2Spring21DRMiniAOD/TTbar_TuneCP5_14TeV-pythia8/MINIAODSIM',
'/store/unmerged/Run2016B/BTagCSV/MINIAOD',
'/store/unmerged/Run2016B/BTagMu/MINIAOD',
'/store/unmerged/Run2016B/DoubleEG/AOD',
'/store/unmerged/Run2016B/DoubleEG/MINIAOD',
'/store/unmerged/Run2016B/DoubleMuonLowMass/MINIAOD',
'/store/unmerged/Run2016B/HTMHT/MINIAOD',                           
  ...
 '/store/unmerged/RunIISummer20UL18wmLHEGEN/ttHJetToNonbb_M125_TuneCP5_13TeV_amcatnloFXFX_madspin_pythia8/GEN',
 '/store/unmerged/RunIISummer20UL18wmLHEGEN/ttHToNonbb_M125_TuneCP5_13TeV-powheg-pythia8/GEN',
 '/store/unmerged/RunIISummer20UL18wmLHEGEN/ttH_HToInvisible_M125_TuneCP5_13TeV_powheg_pythia8/GEN',
 '/store/unmerged/SAM/testSRM/SAM-cms-lvs-gridftp.hep.wisc.edu',
 '/store/unmerged/SAM/testSRM/SAM-cmssrm.hep.wisc.edu',
 '/store/unmerged/SnowmassWinter21GEN/GJet_Pt-20to40_DoubleEMEnriched_MGG-8toInf_TuneCP5_14TeV-pythia8/GEN'
```

todor-ivanov: And actually I just spotted I was cutting the path one level higher than I should have. Fixing it with my next commit.

todor-ivanov:
Before I continue, I wanted to put some statistics on how those aggregations to a higher level in the tree work:
If I run it for all sites known to Rucio (but unfortunately some are missing in Rucio Consistency Monitor - see bellow) I get a quite fast runtime to fetch all lists aggregate paths and split by category: 

```
(WMCore.venv3) [ms-unmerged-standalone]$ time ipython init.py 2>&1 > MSUnmerged.log

real	0m54.649s
user	0m47.566s
sys	0m2.697s
``` 
With service final stats as:
```
Total number of processed RSEs: 83.
Total number of files to be deleted: 5180903.
Number of RSEs cleaned: 0.
Number of files deleted: 0.
```

As of how much we  benefit in terms of memory per object, here is some dump of the most consuming functions (and taking T1_US_FNAL as the worst case scenario site, because T2_CH_CERN is still not present):

```
Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
   238     79.2 MiB     79.2 MiB           1       @profile
   239                                             def getUnmergedFiles(self, rse):
   ...
   258    285.4 MiB    206.3 MiB           1           allUnmerged = self.rucioConMon.getRSEUnmerged(rse['name'])
   ...
   270    139.4 MiB   -146.1 MiB           1           return rse
```
```
Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
   317    139.4 MiB    139.4 MiB           1       @profile
   318                                             def purgeRseObj(self, rse, dumpRSE=False):
   ...
   329    143.5 MiB      4.1 MiB           1               self.logger.debug(msg, pformat(rse))
   330     98.1 MiB    -45.4 MiB           1           rse.clear()
   331     98.1 MiB      0.0 MiB           1           return rse
```

And assuming the memory re usability works for us provided by the python interpreter for for free, it looks quite promising. Even though the end memory consumption for the whole run provided by the default `resource` module states:
```
2021-07-07 11:39:44,918:DEBUG:init:resCons(): PostExec: 
usertime = 4725.818949 
systime = 616.364484 
mem = 577.60546875 mb
```
I would rather impute this to Python garbage collector runs rather than to a bad memory management per object as seen above. We of course may try to investigate further if there are not any left references to the `RSE` object after leaving the pipeline, which may prevent its sole deletion (even though an empty one) by the garbage collector, which may also explain the slow climb in memory for the whole script. In any case this is quite affordable number. 

Need to say though, this is in the situation where I aggregate all the lists for deletions from file level granularity to directory level granularity and throwing the original on a very early stage. If a fallback mechanisms is to be developed for sites not supporting recursive deletions, I would suggest to query for the list of files again, only for those `RSE`s instead of keeping them for every site through the full lifetime of every `RSE` object.

todor-ivanov:
As  mentioned in the previous comment here is the full list of `RSE`s missing from Rucio Consistency Monitor (either `stats` or `unmerged` files interface): [1]. The `T3`s I believe we  may ignore, but there are few essential `T2`s and `T1`s that seem to be missing. Maybe @nsmith- or @ericvaandering may know, what should be done, here. That could be a bug in my script too. 

[1]
```
KeyError: 'T0_CH_CERN_Disk'
KeyError: 'T1_FR_CCIN2P3_Disk'
KeyError: 'T1_RU_JINR_Disk'
KeyError: 'T1_UK_RAL_Disk'
KeyError: 'T2_BE_UCL'
KeyError: 'T2_CH_CERN'
KeyError: 'T2_ES_IFCA'
KeyError: 'T2_RU_ITEP'
KeyError: 'T2_UK_London_IC'
KeyError: 'T2_US_Caltech'
KeyError: 'T3_BG_UNI_SOFIA'
KeyError: 'T3_CH_CERN_OpenData'
KeyError: 'T3_CH_PSI'
KeyError: 'T3_FR_IPNL'
KeyError: 'T3_IT_MIB'
KeyError: 'T3_IT_Trieste'
KeyError: 'T3_KR_KISTI'
KeyError: 'T3_KR_KNU'
KeyError: 'T3_KR_UOS'
KeyError: 'T3_MX_Cinvestav'
KeyError: 'T3_TW_NCU'
KeyError: 'T3_TW_NTU_HEP'
KeyError: 'T3_US_Baylor'
KeyError: 'T3_US_Brown'
KeyError: 'T3_US_CMU'
KeyError: 'T3_US_Colorado'
KeyError: 'T3_US_FNALLPC'
KeyError: 'T3_US_MIT'
KeyError: 'T3_US_NotreDame'
KeyError: 'T3_US_Princeton_ICSE'
KeyError: 'T3_US_PuertoRico'
KeyError: 'T3_US_Rice'
KeyError: 'T3_US_TAMU'
KeyError: 'T3_US_UMD'
KeyError: 'T3_US_UMiss'
```

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: fixed

todor-ivanov:
All tests went ok for standalone run against both single and multiple RSEs. All Unit tests are finally stable. I am merging that PR, because I want to continue with the next issue for MSUnmerged. 
@amaltaro @khurtado, please feel free to leave your comments anyway.

test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
```diff
@@ -0,0 +1,181 @@
...
+
+def getTestFile(partialPath):
+    """
+    Returns the absolute path for the test json file
+    """
+    normPath = os.path.normpath(os.path.join(os.path.dirname(__file__), '..', '..', '..', '..'))
+    return os.path.join(normPath, partialPath)
+
+
+class RucioConMonEmul(object):
```
amaltaro, line 25: @todor-ivanov Todor, once things are running smoothly in production, I would suggest to convert this class to a real RucioConMon emulator under: https://github.com/dmwm/WMCore/tree/master/src/python/WMQuality/Emulators

Could you please make a GH issue for this such that we do not lose track of it? Thanks

PR #10671, opened 2021-07-08
Add lfn2pfn resolve && deletion logic.

todor-ivanov:
Fixes #10660 

#### Status
Ready

#### Description
With the current PR the full deletion logic for the MSUnmerged microeservice is proposed. In order to execute the set of deletions for a given RSE we follow the steps bellow:
 * Fetch the default deletion interface for the RSE from Rucio and fill it in the `MSUnmergedRSE['delInterface']` field
 * Execute the deletions by: 
    * iterating through all `lfns` to be deleted and reassembling every `pfn` by concatenating the `MSUnmergedRSE['delInterface']` + `lfn`
    * execute the `gfal rm -r pfn`
    * collect counters
  
#### Is it backward compatible (if not, which system it affects?)
YES 

#### Related PRs
https://github.com/dmwm/WMCore/pull/10677
https://github.com/cms-sw/cmsdist/pull/7148

#### External dependencies / deployment changes
Introduces a new external library dependency: `Ipython` and few more configuration parameters. The relevant PR to come soon.

We may need to mark here the external dependency from `gfal2-python`, which must already been tagged in one of the previous MSUnmerged issues.

PR #10677, opened 2021-07-20
Add IPython dependency for reqmgr2ms.

### LINKED ISSUE #10676: MSUnmerged: Add external dependencies
**Impact of the new feature**
MSUnmerged

We have introduced yet another dependency for MSUnmerged. This is from the IPython.lib. This is supposed to be rsolved witht he current issue

 * Create the relevant PR in cmsdist introducing the external dependency
 * Add the external dependency in the pip packages install list of WMCore

No

**Additional context**
No

todor-ivanov:
Fixes #10676 

#### Status
ready

#### Description
Adding dependency from IPython because of reqmgr2ms and more precisely MSUnmerged.

YES

#### Related PRs
https://github.com/cms-sw/cmsdist/pull/7148

Yes - it is needed because we use IPython.lib in one of our micro services.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,3 +30,4 @@ Sphinx==1.3.5                 # wmagent,reqmgr2,acdcserver,global-workqueue,reqm
 SQLAlchemy==1.3.3             # wmagent
 stomp.py==4.1.15              # wmagent
 CMSCouchapp==1.2.10           # wmagent
+IPython==5.10.0               # reqmgr2ms
--- a/requirements_py3.txt
+++ b/requirements_py3.txt
@@ -32,3 +32,4 @@ retry==0.9.2
 sphinx==1.6.3
 CMSMonitoring==0.3.4
 CMSCouchapp==1.3.2
+IPython==7.25.0


amaltaro:
@todor-ivanov can you please elaborate on why iPython is needed in reqmgr2ms and whether there was any alternative for that? Unless it's really needed, we should avoid increasing our image size and our software stack, it's just more maintenance in the long run.

Can you please also point which WMCore PR introduced that dependency (the library being used, not the requirements).

todor-ivanov:
@amaltaro this is the PR which introduces the requirement:
https://github.com/dmwm/WMCore/pull/10671

And the reason is because we want to have a major dump of the the structure of an RSE element with only few  element  per key instead of a full dump. This is trivial with IPython.lib.pretty while it requires quite a lot of additional coding to achieve if we want to implement it with the default pformat library

amaltaro:
Thanks, Todor. Since it's only a single line using it, and for a debug logging level, I would suggest to actually update the WMCore code such that we do not depend on it. It's always an extra burden to maintain those...

Spot checks in my inbox make me believe you are already replacing it and also removing this dependency from our stack, right? If not, can you please do so? even if that costs us a few more lines of code.

todor-ivanov:
it is already done @amaltaro :
https://github.com/dmwm/WMCore/pull/10679/files

PR #10684, opened 2021-07-21
change fallback for missing system xrootd-client, fixes #10796

### LINKED ISSUE #10796: fallback option in XRDCP stageout plugin too old
**Impact of the bug**
Sites without a locally installed xrootd client will potentially not be able to stageout (depends what xrdcp options are used, it definitely doesn't work for stageout to FNAL storage).

**Describe the bug**
The XRDCP stageout plugin has a fallback option when it doesn't find a locally installed xrdcp. It currently used a 5 years old xrootd-client that is deployed in the CMS COMP cvmfs. When the fallback recently triggered on FNAL Wilson, I found out that the xrdcp version is too old to support the stageout options we use for FNAL (specifically the -p option, which seems to be needed for FNAL to automatically create new directories if they don't exist yet).

**How to reproduce it**
I can't directly reproduce it anymore since FNAL Wilson had an xrootd client installed on the worker nodes.

**Expected behavior**
If the XRDCP stageout plugin doesn't find xrdcp in the path, it should fallback to a usable xrootd client installation from cvmfs.

hufnagel:
Fixes https://github.com/dmwm/WMCore/issues/10796

#### Status
Not tested and will be hard to test, since I don't have any resources anymore without a system installed xrootd client (even the resources that initially triggered this development have since been updated). 

#### Description
In testing on the FNAL Wilson HPC cluster I observed that the fallback option for xrdcp stageout, using the 4.0.4 xrootd client from COMP, doesn't work anymore. The client is too old to support some options we later added for xrdcp stageout at some sites. There is no newer client in the COMP repo (or a CMS repo), but OSG provides one as part of their worker node client.

YES (should affect no CMS site since the changed code path is basically never executed at fully commissioning CMS sites)

### DIFF
--- a/src/python/WMCore/Storage/Backends/XRDCPImpl.py
+++ b/src/python/WMCore/Storage/Backends/XRDCPImpl.py
@@ -87,32 +87,18 @@ class XRDCPImpl(StageOutImpl):
             xrdcpExec = "xrdcp-old"
 
         # check if xrdcp(-old) and xrdfs are in path
-        # fallback to xrootd 4.0.4 from COMP externals if not
         xrootdInPath = False
         if any(os.access(os.path.join(path, xrdcpExec), os.X_OK) for path in os.environ["PATH"].split(os.pathsep)):
             if any(os.access(os.path.join(path, "xrdfs"), os.X_OK) for path in os.environ["PATH"].split(os.pathsep)):
                 xrootdInPath = True
 
+        # if not check if OSG cvmfs repo is available
+        #   this likely only works on RHEL7 x86-64 (and compatible) OS, but this represent most of our resources
+        #   and this is a fallback (i.e. stageout would fail if we don't try something)
         if not xrootdInPath:
-            # COMP software can be in many place, check all of them
-            cmsSoftDir = os.environ.get("VO_CMS_SW_DIR", None)
-            if not cmsSoftDir:
-                cmsSoftDir = os.environ.get("OSG_APP", None)
-                if cmsSoftDir:
-                    cmsSoftDir = os.path.join(cmsSoftDir, "cmssoft/cms")
-                else:
-                    cmsSoftDir = os.environ.get("CVMFS", None)
-
-            if cmsSoftDir:
-
-                initFiles = []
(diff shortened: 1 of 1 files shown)


hufnagel:
Can we have this tested again? I don't see a reason why the initial checks failed.

amaltaro:
Sure, I'm going to trigger a new test with my next comment. @hufnagel meanwhile, could you please create the relevant GH issue which this PR is supposed to fix?

amaltaro:
test this please

hufnagel:
Added a GH issue with the bug report. Let me also look at the failures.

amaltaro:
Thanks, Dirk. Is this:
https://github.com/dmwm/WMCore/issues/10796
the issue you are addressing with this PR?

hufnagel:
Yes, that's the one. Have looked at the unit test failures, but nothing immediately jumps out, i.e. I *think* the changes I made shouldn't affect the tests that now show error. Is there a way to get the full unit test logs, not just what the Details link shows?

amaltaro:
Yes, have a look at this section, which gives you a brief explanation on how to check those failures in more details:
https://github.com/dmwm/WMCore/wiki/Understanding-Jenkins#unit-tests

amaltaro:
@hufnagel Dirk, I would like very much to replace this ancient XRD fallback by something not so old.
Would you be willing to looking and fixing the unit tests? I think the change would only be required to this line:
https://github.com/dmwm/WMCore/blob/master/test/python/WMCore_t/Storage_t/Backends_t/XRDCPImpl_t.py#L90

with the addition of what your PR provides, thus something like `source ....`.

hufnagel:
The xrdcp stageout unit tests run at CERN and never included the fallback (since it's simply not triggered there).

So I don't know why the fail here. I would have to see an actual error message for the unit tests jenkins runs to track this down (and I am not setup anymore to run this interactively, would take some time to get that setup back).

amaltaro:
This has been superseeded by https://github.com/dmwm/WMCore/pull/11117
Closing this one! Thanks Dirk!

PR #10687, opened 2021-07-22
Account for failed deletions on a per directory basis

### LINKED ISSUE #10686: MSUnmerged: Improve accounting of deletions
**Impact of the new feature**
MSUnmerged

In the process of testing the service we've got to realize that we are accounting mostly for successful deletions and not logging the failed ones. We also may try to find  way to measure the volume to be released on every iteration, by an additional series of gfal calls to the RSE, once we have the lists for deletions built. 

 * Account not only for successful, but also for failed deletions on a per directory basis
 * implement a series of `gfal-ls`-like calls to the RSE once the lists of files to be deleted are build or in the process of deletion  (but beware not to consume the file generators too early) so that we may have meaningful assessment of the volume released on the RSE

None

**Additional context**
Non

### LINKED ISSUE #10933: Cancelling deletions permanently after reaching limit files per RSE per cycle
**Impact of the bug**
MSUnmerged

**Describe the bug**
Since we have implemented the stateful version of the service we started to preserve counters among the runs trough the background database. One of the counters which was also preserved was `rse['counters']['filesToDelete']` which is later used in the following line [1] to check if the RSE reached its configuration limit of files to be deleted on a single cycle. This one was  working perfectly fine untill now because on every run the counters were zeroed at the beginning of the MSUnmerged polling cycle. Now the counters are zeroed with the polling cycle of RucioConMon. Which means once the configuration limit is reached the  RSE will never be retried.  

[1]
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py#L307

**How to reproduce it**
Set a low configuration limit of `limitFilesPerRSE` and monitor the service run for two consecutive polling cycles. The service would try to delete exactly the number of files from the limit value and will try to do it only during the first run.   

**Expected behavior**
The service should always retry to delete files from the RSE on every service polling cycle until the full amount of files marked of deletions is reached.

The solutions possible are exactly two:

 * Temporary fix - remove the limit per RSE (set the value in MSConfig: `data.limitFilesPerRSE = -1` - the minor disadvantage would be a prolonged run of one of the pods when hitting a large RSE like T2_CH_CERN 
 * Permanent fix - do not consult a globally preserved value for entering the cycle, but move it as  a local one instead.
     * This is already implemented/fixed in the following PR: https://github.com/dmwm/WMCore/pull/10687 with this commit:  https://github.com/dmwm/WMCore/pull/10687/commits/28aa18911d5b6e17df016c0500655c4cc884202d but we need to also apply the temporary fix until the permanent one is merged. 

todor-ivanov:
Fixes #10686 
Fixes #10933

#### Status
Ready

#### Description
With the current PR we are trying to improve the accounting for failed deletions, while at the same time not increase the amount  of log lines, which is already too high.  

Here are some of the most important changes which it introduces:

 * Handle exceptions per slice instead of the whole RSE - which avoids a single exception from one deletion slice to interrupt the cleaning of the whole RSE && 
 * Count gfall Errors per run - This would provide us with a good summary of what are the most common errors an RSE is experiencying so we can try to address them through GGUS tickets to the site or some other way && 
 * Keep list of deleted success dir - this improves the accounting of already deleted directories && 
 * Do not retry successfully deleted dirs  - Fixes broken counters and minimizes the unneeded operations per RSE per iteration && 
 * Add baseDirPfn check  - this way we avoid errors related to files existing at the toplevel of the current directory which could be the case with files in the following path  `/store/unmerged/SAM/ `  (we may also think of excluding this path too) && 
 * Fix exception handling during recursive iteration in _purgeTree().
 
Because this PR introduces quite  some new fields in the RSE record in the database before deploying we will have to wipe out all the records otherwise we will experience a long series of exceptions because of the missing keys in the old RSE records. And this would continue at least until the next Rucio Consistency Monitor poling cycle when the records for all RSE would be refreshed anyways. 
 
And here is how would look like the output of the info REST API for an RSE which is having all those new fields: [1]  [2]

[1]
https://cmsweb-testbed.cern.ch/ms-unmerged/data/info?rse=T2_US_Florida&detail=True

[2]
https://cmsweb-testbed.cern.ch/ms-unmerged/data/info?rse=T2_RU_INR&detail=True

```
{"result": [
 {
  "wmcore_version": "1.5.7.pre5",
  "microservice_version": "1.5.7.pre5",
  "microservice": "MSManager",
  "query": "rse=T2_RU_INR&detail=True",
  "rseData": [
    {
      "name": "T2_RU_INR",
      "pfnPrefix": "srm://grse001.inr.troitsk.ru:8446/srm/managerv2?SFN=/dpm/inr.troitsk.ru/home/cms",
      "isClean": false,
      "timestamps": {
        "rseConsStatTime": 0.0,
        "prevStartTime": 1642773881.8954148,
        "startTime": 1642778382.9562628,
        "prevEndTime": 1642773882.0129983,
        "endTime": 1642778541.2820504
      },
      "counters": {
        "totalNumFiles": 76,
        "totalNumDirs": 26,
        "dirsToDelete": 2,
        "filesToDelete": 0,
        "filesDeletedSuccess": 0,
        "filesDeletedFail": 0,
        "dirsDeletedSuccess": 0,
        "dirsDeletedFail": 2,
        "gfalErrors": {
          "No such file or directory": 3
        }
      },
      "dirs": {
        "allUnmerged": [],
        "toDelete": [
          "/store/unmerged/SAM/se_gsiftp_20220111_193202_etf-01.wrt",
          "/store/unmerged/SAM/testSRM/SAM-grse001.inr.troitsk.ru/lcg-util"
        ],
        "protected": [
          "/store/unmerged/RunIISummer20UL16NanoAODAPVv9/TTGamma_Dilept_TuneCP5CR2_13TeV-madgraph-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
          "/store/unmerged/RunIISummer20UL16NanoAODv2/GluGluToHiggs0Mf05ph0continToZZTo2e2nu_M125_10GaSM_TuneCP5_13TeV-mcfm701-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v15-v1",
          "/store/unmerged/RunIISummer20UL18NanoAODv2/GluGluToHiggs0Mf05ph0continToZZTo2mu2tau_M125_GaSM_TuneCP5_13TeV-mcfm701-pythia8/NANOAODSIM/106X_upgrade2018_realistic_v15_L1v1-v1",
          "/store/unmerged/RunIISummer20UL16NanoAODAPVv9/QCD_HT300to500_TuneCP5_PSWeights_13TeV-madgraphMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
          "/store/unmerged/RunIISummer20UL17NanoAODv2/HToSSTo2Mu2Hadrons_MH125_MS0p6_ctauS100_TuneCP2_13TeV-powheg-pythia8/NANOAODSIM/106X_mc2017_realistic_v8-v1",
          "/store/unmerged/RunIISummer20UL16MiniAODv2/HbToJPsiMuMu_3MuFilter_TuneCP5_13TeV-pythia8-evtgen/MINIAODSIM/106X_mcRun2_asymptotic_v17-v2",
          "/store/unmerged/RunIISummer20UL16MiniAODAPVv2/QCD_HT300to500_TuneCP5_PSWeights_13TeV-madgraphMLM-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
          "/store/unmerged/RunIISummer20UL16NanoAODv2/HbToJPsiMuMu_TuneCP5_13TeV-pythia8-evtgen/NANOAODSIM/106X_mcRun2_asymptotic_v15-v1",
          "/store/unmerged/RunIISummer20UL16MiniAODv2/QCD_Pt-20to30_EMEnriched_TuneCP5_13TeV-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_v17-v2",
          "/store/unmerged/RunIISummer20UL16NanoAODv9/QCD_HT300to500_TuneCP5_PSWeights_13TeV-madgraphMLM-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v1",
          "/store/unmerged/RunIISummer20UL17MiniAODv2/GluGluToHiggs0PHToZZTo2e2nu_M125_GaSM_TuneCP5_13TeV-mcfm701-pythia8/MINIAODSIM/106X_mc2017_realistic_v9-v2",
          "/store/unmerged/RunIISummer20UL16NanoAODAPVv2/QCD_Pt-1000_MuEnrichedPt5_TuneCP5_13TeV-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_preVFP_v9-v1",
          "/store/unmerged/RunIISummer20UL16NanoAODv2/VBF_LFV_HToEMu_M125_TuneCH3_13TeV_powheg_herwig7/NANOAODSIM/106X_mcRun2_asymptotic_v15_ext1-v1",
          "/store/unmerged/RunIISummer20UL16NanoAODv9/HbToJPsiMuMu_3MuFilter_TuneCP5_13TeV-pythia8-evtgen/NANOAODSIM/106X_mcRun2_asymptotic_v17-v2",
          "/store/unmerged/RunIISummer20UL18NanoAODv9/HbToJPsiMuMu_3MuFilter_TuneCP5_13TeV-pythia8-evtgen/NANOAODSIM/106X_upgrade2018_realistic_v16_L1v1_ext1-v1",
          "/store/unmerged/RunIISummer20UL18NanoAODv9/TT_Mtt-700to1000_TuneCP5_13TeV-powheg-pythia8/NANOAODSIM/20UL18JMENano_106X_upgrade2018_realistic_v16_L1v1-v1",
          "/store/unmerged/RunIISummer20UL17NanoAODv9/GluGluToHiggs0PHToZZTo2e2nu_M125_GaSM_TuneCP5_13TeV-mcfm701-pythia8/NANOAODSIM/106X_mc2017_realistic_v9-v2",
          "/store/unmerged/RunIISummer20UL16MiniAODAPVv2/TTGamma_Dilept_TuneCP5CR2_13TeV-madgraph-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
          "/store/unmerged/RunIISummer20UL16MiniAODAPVv2/QCD_Pt-20To30_MuEnrichedPt5_TuneCP5_13TeV-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
          "/store/unmerged/RunIISummer20UL17NanoAODv9/GluGluToRadionToHHTo2G2ZTo2G4Q_M-600_TuneCP5_PSWeights_narrow_13TeV-madgraph-pythia8/NANOAODSIM/106X_mc2017_realistic_v9-v1",
          "/store/unmerged/RunIISummer20UL18NanoAODv9/TT_Mtt-700to1000_TuneCP5_13TeV-powheg-pythia8/NANOAODSIM/106X_upgrade2018_realistic_v16_L1v1-v1",
          "/store/unmerged/RunIISummer20UL16NanoAODAPVv9/QCD_Pt-20To30_MuEnrichedPt5_TuneCP5_13TeV-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
          "/store/unmerged/RunIISummer20UL17NanoAODv9/XtoAAto4G_X400A6_TuneCP5_13TeV-madgraphMLM-pythia8/NANOAODSIM/106X_mc2017_realistic_v9-v1",
          "/store/unmerged/RunIISummer20UL16NanoAODv9/QCD_Pt-20to30_EMEnriched_TuneCP5_13TeV-pythia8/NANOAODSIM/106X_mcRun2_asymptotic_v17-v2"
        ],
        "deletedSuccess": [],
        "deletedFail": [
          "/store/unmerged/SAM/se_gsiftp_20220111_193202_etf-01.wrt",
          "/store/unmerged/SAM/testSRM/SAM-grse001.inr.troitsk.ru/lcg-util"
        ]
      }
    }
  ]
}]}

```

NO

#### Related PRs
None

Needs complete refresh/deletion of the MSUnmerged MongoDBCollection

### DIFF
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
@@ -257,7 +257,7 @@ class MSUnmerged(MSCore):
         """
 
         pline = self.plineUnmerged
-        self.resetCounters(plineName=pline.name)
+        self.resetServiceCounters(plineName=pline.name)
         self.plineCounters[pline.name]['totalNumRses'] = len(rseList)
 
         for rseName in rseList:
@@ -282,7 +282,7 @@ class MSUnmerged(MSCore):
         return self.plineCounters[pline.name]['totalNumRses'], \
             self.plineCounters[pline.name]['totalNumFiles'], \
             self.plineCounters[pline.name]['rsesCleaned'], \
-            self.plineCounters[pline.name]['deletedSuccess']
+            self.plineCounters[pline.name]['filesDeletedSuccess']
 
     # @profile
     def cleanRSE(self, rse):
@@ -301,13 +301,16 @@ class MSUnmerged(MSCore):
             self.logger.exception(msg)
             raise MSUnmergedPlineExit(msg)
 
+        filesToDeleteCurrRSE = 0
+
         # Start cleaning one directory at a time:
         for dirLfn, fileLfnGen in rse['files']['toDelete'].items():
-            if self.msConfig['limitFilesPerRSE'] < 0 or \
-               rse['counters']['filesToDelete'] < self.msConfig['limitFilesPerRSE']:
+            if dirLfn in rse['dirs']['deletedSuccess']:
+                self.logger.info("RSE: %s, dir: %s already successfully deleted.", rse['name'], dirLfn)
+                continue
 
-                # First increment the dir counter:
-                rse['counters']['dirsToDelete'] += 1
+            if self.msConfig['limitFilesPerRSE'] < 0 or \
+               filesToDeleteCurrRSE < self.msConfig['limitFilesPerRSE']:
 
                 # Now we consume the rse['files']['toDelete'][dirLfn] generator
                 # upon that no values will be left in it. In case we need it again
@@ -328,36 +331,58 @@ class MSUnmerged(MSCore):
                         filePfn = rse['pfnPrefix'] + fileLfn
                         pfnList.append(filePfn)
 
-                rse['counters']['filesToDelete'] += len(pfnList)
+                filesToDeleteCurrRSE += len(pfnList)
                 msg = "\nRSE: %s \nDELETING: %s."
                 msg += "\nPFN list with: %s entries: \n%s"
                 self.logger.debug(msg, rse['name'], dirLfn, len(pfnList), twFormat(pfnList, maxLength=4))
 
                 if self.msConfig['enableRealMode']:
-                    try:
-                        # execute the actual deletion in bulk - full list of files per directory
+                    # execute the actual deletion in bulk - full list of files per directory
 
-                        deletedSuccess = 0
-                        for pfnSlice in list(grouper(pfnList, self.msConfig["filesToDeleteSliceSize"])):
+                    deletedSuccess = 0
+                    for pfnSlice in list(grouper(pfnList, self.msConfig["filesToDeleteSliceSize"])):
+                        try:
                             delResult = ctx.unlink(pfnSlice)
                             # Count all the successfully deleted files (if a deletion was
                             # successful a value of None is put in the delResult list):
-                            deletedSuccess += sum([1 for pfnStatus in delResult if pfnStatus is None])
+                            self.logger.debug("RSE: %s, Dir: %s, delResult: %s",
+                                              rse['name'], dirLfn, pformat(delResult))
+                            for gfalErr in delResult:
+                                if gfalErr is None:
+                                    deletedSuccess += 1
+                                else:
+                                    errMessage = os.strerror(gfalErr.code)
+                                    rse['counters']['gfalErrors'].setdefault(errMessage, 0)
+                                    rse['counters']['gfalErrors'][errMessage] += 1
+                        except Exception as ex:
+                            msg = "Error while cleaning RSE: %s. "
+                            msg += "Will retry in the next cycle. Err: %s"
+                            self.logger.exception(msg, rse['name'], str(ex))
 
-                        self.logger.debug("RSE: %s, Dir: %s, deletedSuccess: %s",
-                                          rse['name'], dirLfn, deletedSuccess)
-                        rse['counters']['deletedSuccess'] = deletedSuccess
+                    self.logger.info("RSE: %s, Dir: %s, filesDeletedSuccess: %s",
+                                      rse['name'], dirLfn, deletedSuccess)
+                    rse['counters']['filesDeletedSuccess'] += deletedSuccess
 
-                        # Now clean the whole branch
-                        self.logger.debug("Purging dirEntry: %s:\n", dirPfn)
-                        purgeSuccess = self._purgeTree(ctx, dirPfn)
-                        if not purgeSuccess:
-                            msg = "RSE: %s Failed to purge nonEmpty directory: %s"
-                            self.logger.error(msg, rse['name'], dirPfn)
-                    except Exception as ex:
-                        msg = "Error while cleaning RSE: %s. "
-                        msg += "Will retry in the next cycle. Err: %s"
-                        self.logger.warning(msg, rse['name'], str(ex))
+                    # Now clean the whole branch
+                    self.logger.debug("Purging dirEntry: %s:\n", dirPfn)
+                    purgeSuccess = self._purgeTree(ctx, dirPfn)
+                    if purgeSuccess:
+                        rse['dirs']['deletedSuccess'].add(dirLfn)
+                        rse['counters']['dirsDeletedSuccess'] = len(rse['dirs']['deletedSuccess'])
+                        # if dirLfn in rse['dirs']['toDelete']:
+                        #     rse['dirs']['toDelete'].remove(dirLfn)
+                        if dirLfn in rse['dirs']['deletedFail']:
+                            rse['dirs']['deletedFail'].remove(dirLfn)
+                        msg = "RSE: %s  Success deleting directory: %s"
+                        self.logger.info(msg, rse['name'], dirPfn)
+                    else:
+                        rse['dirs']['deletedFail'].add(dirLfn)
+                        rse['counters']['dirsDeletedFail'] = len(rse['dirs']['deletedFail'])
+                        msg = "RSE: %s Failed to purge directory: %s"
+                        self.logger.error(msg, rse['name'], dirPfn)
+            else:
+                msg = "RSE: %s reached limit of files per RSE to be deleted. Skipping directory: %s. It will be retried on the next cycle."
+                self.logger.warning(msg, rse['name'], dirLfn)
         rse['isClean'] = self._checkClean(rse)
 
         return rse
@@ -369,44 +394,60 @@ class MSUnmerged(MSCore):
         :param ctx:  The gfal2 context object
         :return:     Bool: True if it managed to purge everything, False otherwise
         """
-        successList = []
+        # NOTE: It deletes only directories and does not try to unlink any file.
+
+        # First test if baseDirPfn is actually a directory entry:
+        try:
+            entryStat = ctx.stat(baseDirPfn)
+            if not stat.S_ISDIR(entryStat.st_mode):
+                self.logger.error("The base pfn: %s is not a directory entry.", baseDirPfn)
+                return False
+        except gfal2.GError as gfalExc:
+            if gfalExc.code == errno.ENOENT:
+                self.logger.warning("MISSING baseDir: %s", baseDirPfn)
+                return True
+            else:
+                self.logger.error("FAILED to open baseDir: %s: gfalException: %s", baseDirPfn, str(gfalExc))
+                return False
 
         if baseDirPfn[-1] != '/':
             baseDirPfn += '/'
 
+        # Second recursively iterate down the tree:
+        successList = []
         for dirEntry in ctx.listdir(baseDirPfn):
             if dirEntry in ['.', '..']:
                 continue
             dirEntryPfn = baseDirPfn + dirEntry
             try:
                 entryStat = ctx.stat(dirEntryPfn)
-            except gfal2.GError:
-                e = sys.exc_info()[1]
-                if e.code == errno.ENOENT:
-                    self.logger.error("MISSING dirEntry: %s", dirEntryPfn)
-                    successList.append(False)
-                    return all(successList)
+            except gfal2.GError as gfalExc:
+                if gfalExc.code == errno.ENOENT:
+                    self.logger.warning("MISSING dirEntry: %s", dirEntryPfn)
+                    successList.append(True)
                 else:
-                    self.logger.error("FAILED dirEntry: %s", dirEntryPfn)
-                    raise
+                    self.logger.error("FAILED to open dirEntry: %s: gfalException: %s", dirEntryPfn, str(gfalExc))
+                    successList.append(False)
+
             if stat.S_ISDIR(entryStat.st_mode):
                 successList.append(self._purgeTree(ctx, dirEntryPfn))
 
+        # Finally remove the baseDir:
         try:
+            self.logger.debug("RM baseDir: %s", baseDirPfn)
             success = ctx.rmdir(baseDirPfn)
             # for gfal2 rmdir() exit status of 0 is success
             if success == 0:
                 successList.append(True)
             else:
                 successList.append(False)
-            self.logger.debug("RM baseDir: %s", baseDirPfn)
-        except gfal2.GError:
-            e = sys.exc_info()[1]
-            if e.code == errno.ENOENT:
-                self.logger.error("MISSING baseDir: %s", baseDirPfn)
+        except gfal2.GError as gfalExc:
+            if gfalExc.code == errno.ENOENT:
+                self.logger.warning("MISSING baseDir: %s", baseDirPfn)
+                successList.append(True)
             else:
-                self.logger.error("FAILED basedir: %s", baseDirPfn)
-                raise
+                self.logger.error("FAILED to remove baseDir: %s: gfalException: %s", baseDirPfn, str(gfalExc))
+                successList.append(False)
         return all(successList)
 
     def _checkClean(self, rse):
@@ -416,7 +457,7 @@ class MSUnmerged(MSCore):
         :param rse: The RSE to be checked
         :return:    Bool: True if all files found have been deleted, False otherwise
         """
-        return rse['counters']['filesToDelete'] == rse['counters']['deletedSuccess']
+        return rse['counters']['dirsToDelete'] == rse['counters']['dirsDeletedSuccess']
 
     def consRecordAge(self, rse):
         """
@@ -443,7 +484,7 @@ class MSUnmerged(MSCore):
                 self.updateRSETimestamps(rse, start=False, end=True)
                 raise MSUnmergedPlineExit(msg)
             else:
-                msg += "But the RSE has NOT been cleaned during the last Rucio Consistency Monitor polling cycle."
+                msg += "But the RSE has NOT been fully cleaned during the last Rucio Consistency Monitor polling cycle."
                 msg += "Retrying cleanup in the current run."
                 self.logger.info(msg)
         if not isConsDone:
@@ -468,6 +509,9 @@ class MSUnmerged(MSCore):
                 msg += "Giving up now. The whole cleanup process will be retried for this RSE on the next run."
                 msg += "Duplicate deletion retries may cause error messages from false positives and wrong counters during next polling cycle."
                 raise MSUnmergedPlineExit(msg)
+
+        # Before returning the RSE update Consistency StatTime:
+        rse['timestamps']['rseConsStatTime'] = self.rseConsStats[rseName]['end_time']
         return rse
 
     # @profile
@@ -492,7 +536,6 @@ class MSUnmerged(MSCore):
         """
         rse['files']['allUnmerged'] = self.rucioConMon.getRSEUnmerged(rse['name'])
         for filePath in rse['files']['allUnmerged']:
-            rse['counters']['totalNumFiles'] += 1
             # Check if what we start with is under /store/unmerged/*
             if self.regStoreUnmergedLfn.match(filePath):
                 # Cut the path to the deepest level known to WMStats protected LFNs
@@ -501,6 +544,9 @@ class MSUnmerged(MSCore):
                 if self.regStoreUnmergedLfn.match(dirPath):
                     # Add it to the set of allUnmerged
                     rse['dirs']['allUnmerged'].add(dirPath)
+
+        rse['counters']['totalNumFiles'] = len(rse['files']['allUnmerged'])
+        rse['counters']['totalNumDirs'] = len(rse['dirs']['allUnmerged'])
         return rse
 
     def _cutPath(self, filePath):
@@ -547,7 +593,7 @@ class MSUnmerged(MSCore):
 
         # The following check may seem redundant, but better stay safe than sorry
         if not (rse['dirs']['toDelete'] | rse['dirs']['protected']) == rse['dirs']['allUnmerged']:
-            rse['counters']['toDelete'] = -1
+            rse['counters']['dirsToDelete'] = -1
             msg = "Incorrect set check while trying to estimate the final set for deletion."
             raise MSUnmergedPlineExit(msg)
 
@@ -579,10 +625,10 @@ class MSUnmerged(MSCore):
         if not dirFilterIncl and not dirFilterExcl:
             for dirName in rse['dirs']['toDelete']:
                 rse['files']['toDelete'][dirName] = genFunc(dirName, rse['files']['allUnmerged'])
-            rse['counters']['dirsToDeleteAll'] = len(rse['files']['toDelete'])
+            rse['counters']['dirsToDelete'] = len(rse['files']['toDelete'])
+            self.logger.info("RSE: %s: %s", rse['name'], twFormat(rse, maxLength=8))
             return rse
 
-        self.logger.info("rse: %s", twFormat(rse, maxLength=8))
         # If we are here, then there are service filters...
         for dirName in rse['dirs']['toDelete']:
             # apply exclusion filter
@@ -603,8 +649,12 @@ class MSUnmerged(MSCore):
                     rse['files']['toDelete'][dirName] = genFunc(dirName, rse['files']['allUnmerged'])
                     break
 
+        # Now apply the filters back to the set in rse['dirs']['toDelete']
+        rse['dirs']['toDelete'] = set(rse['files']['toDelete'].keys())
+
         # Update the counters:
-        rse['counters']['dirsToDeleteAll'] = len(rse['files']['toDelete'])
+        rse['counters']['dirsToDelete'] = len(rse['files']['toDelete'])
+        self.logger.info("RSE: %s: %s", rse['name'], twFormat(rse, maxLength=8))
         return rse
 
     def getPfn(self, rse):
@@ -690,21 +740,28 @@ class MSUnmerged(MSCore):
         :return:      rse
         """
         rseName = rse['name']
-        self.resetCounters(rseName=rseName)
+        self.resetServiceCounters(rseName=rseName)
         self.rseCounters[rseName]['totalNumFiles'] = rse['counters']['totalNumFiles']
-        self.rseCounters[rseName]['deletedSuccess'] = rse['counters']['deletedSuccess']
-        self.rseCounters[rseName]['deletedFail'] = rse['counters']['deletedFail']
+        self.rseCounters[rseName]['totalNumDirs'] = rse['counters']['totalNumDirs']
+        self.rseCounters[rseName]['filesDeletedSuccess'] = rse['counters']['filesDeletedSuccess']
+        self.rseCounters[rseName]['filesDeletedFail'] = rse['counters']['filesDeletedFail']
+        self.rseCounters[rseName]['dirsDeletedSuccess'] = rse['counters']['dirsDeletedSuccess']
+        self.rseCounters[rseName]['dirsDeletedFail'] = rse['counters']['dirsDeletedFail']
+        self.rseCounters[rseName]['gfalErrors'] = rse['counters']['gfalErrors']
 
         self.plineCounters[pName]['totalNumFiles'] += rse['counters']['totalNumFiles']
-        self.plineCounters[pName]['deletedSuccess'] += rse['counters']['deletedSuccess']
-        self.plineCounters[pName]['deletedFail'] += rse['counters']['deletedFail']
+        self.plineCounters[pName]['totalNumDirs'] += rse['counters']['totalNumDirs']
+        self.plineCounters[pName]['filesDeletedSuccess'] += rse['counters']['filesDeletedSuccess']
+        self.plineCounters[pName]['filesDeletedFail'] += rse['counters']['filesDeletedFail']
+        self.plineCounters[pName]['dirsDeletedSuccess'] += rse['counters']['dirsDeletedSuccess']
+        self.plineCounters[pName]['dirsDeletedFail'] += rse['counters']['dirsDeletedFail']
         self.plineCounters[pName]['rsesProcessed'] += 1
         if rse['isClean']:
             self.plineCounters[pName]['rsesCleaned'] += 1
 
         return rse
 
-    def resetCounters(self, rseName=None, plineName=None):
+    def resetServiceCounters(self, rseName=None, plineName=None):
         """
         A simple function for zeroing the service counters.
         :param rseName:   RSE Name whose counters to be zeroed
@@ -716,8 +773,12 @@ class MSUnmerged(MSCore):
             if rseName not in self.rseCounters:
                 self.rseCounters[rseName] = {}
             self.rseCounters[rseName]['totalNumFiles'] = 0
-            self.rseCounters[rseName]['deletedSuccess'] = 0
-            self.rseCounters[rseName]['deletedFail'] = 0
+            self.rseCounters[rseName]['totalNumDirs'] = 0
+            self.rseCounters[rseName]['filesDeletedSuccess'] = 0
+            self.rseCounters[rseName]['filesDeletedFail'] = 0
+            self.rseCounters[rseName]['dirsDeletedSuccess'] = 0
+            self.rseCounters[rseName]['dirsDeletedFail'] = 0
+            self.rseCounters[rseName]['gfalErrors'] = {}
             return
 
         # Resetting Just the pline counters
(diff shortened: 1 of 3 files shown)


todor-ivanov:
While doing some extra tests with realRun mode at  T2_US_Wisconsin I've just found the following error: [1]. This means some of the `gfalError` messages, which I am using as key values, are having `dotted` format hence are not suitable to be written to MongoDB. I'll figure a way out tomorrow morning. In the meantime @amaltaro please feel free to review the rest. It has been tested already. 


[1]
```
2022-01-20 23:28:51,163:INFO:MSUnmerged:uploadRSEToMongoDB(): RSE: T2_US_Wisconsin Writing rse data to MongoDB.
2022-01-20 23:28:51,196:ERROR:MSUnmerged:_execute(): plineUnmerged: General error from pipeline. RSE: T2_US_Wisconsin. Error: The dotted field 'globus_ftp_client: the server responded with an error 451 451-GlobusError: v=1 c=INTERNAL_ERROR  451-GridFTP-Errno: 5  451-GridFTP-Reason: System error in unlink (host=g27n19.hep.wisc.edu, user=cmsprod, path=/store/unmerged/RunIISummer20UL16MiniAOD/ZpMuMu_MZp-0p0005_Df-0p0251_TuneCP5_13TeV-madgraph-pythia8/MINIAODSIM/darkfactor_0p0251_106X_mcRun2_asymptotic_v13-v2/2540000/4D92B796-F97F-7F4F-87CD-15086CCF8289.root)  451-GridFTP-Error-String: Input/output error  451 End.   ' in 'counters.gfalErrors.globus_ftp_client: the server responded with an error 451 451-GlobusError: v=1 c=INTERNAL_ERROR  451-GridFTP-Errno: 5  451-GridFTP-Reason: System error in unlink (host=g27n19.hep.wisc.edu, user=cmsprod, path=/store/unmerged/RunIISummer20UL16MiniAOD/ZpMuMu_MZp-0p0005_Df-0p0251_TuneCP5_13TeV-madgraph-pythia8/MINIAODSIM/darkfactor_0p0251_106X_mcRun2_asymptotic_v13-v2/2540000/4D92B796-F97F-7F4F-87CD-15086CCF8289.root)  451-GridFTP-Error-String: Input/output error  451 End.   ' is not valid for storage. Will retry again in the next cycle.
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py", line 266, in _execute
    pline.run(MSUnmergedRSE(rseName))
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 140, in run
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 140, in <lambda>
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 72, in __call__
    return self.run(obj)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/Utils/Pipeline.py", line 75, in run
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py", line 839, in uploadRSEToMongoDB
    rse.writeRSEToMongoDB(self.msUnmergedColl, fullRSEToDB=fullRSEToDB, overwrite=overwrite, retryCount=self.msConfig['mongoDBRetryCount'])
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore/src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py", line 173, in writeRSEToMongoDB
    result = collection.replace_one(self.mongoFilter, updateFields, upsert=True)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/collection.py", line 930, in replace_one
    collation=collation, session=session),
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/collection.py", line 856, in _update_retryable
    _update, session)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/mongo_client.py", line 1491, in _retryable_write
    return self._retry_with_session(retryable, func, s, None)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/mongo_client.py", line 1384, in _retry_with_session
    return func(session, sock_info, retryable)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/collection.py", line 852, in _update
    retryable_write=retryable_write)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/collection.py", line 823, in _update
    _check_write_command_response(result)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/helpers.py", line 221, in _check_write_command_response
    _raise_last_write_error(write_errors)
  File "/afs/cern.ch/user/t/tivanov/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/helpers.py", line 203, in _raise_last_write_error
    raise WriteError(error.get("errmsg"), error.get("code"), error)
pymongo.errors.WriteError: The dotted field 'globus_ftp_client: the server responded with an error 451 451-GlobusError: v=1 c=INTERNAL_ERROR  451-GridFTP-Errno: 5  451-GridFTP-Reason: System error in unlink (host=g27n19.hep.wisc.edu, user=cmsprod, path=/store/unmerged/RunIISummer20UL16MiniAOD/ZpMuMu_MZp-0p0005_Df-0p0251_TuneCP5_13TeV-madgraph-pythia8/MINIAODSIM/darkfactor_0p0251_106X_mcRun2_asymptotic_v13-v2/2540000/4D92B796-F97F-7F4F-87CD-15086CCF8289.root)  451-GridFTP-Error-String: Input/output error  451 End.   ' in 'counters.gfalErrors.globus_ftp_client: the server responded with an error 451 451-GlobusError: v=1 c=INTERNAL_ERROR  451-GridFTP-Errno: 5  451-GridFTP-Reason: System error in unlink (host=g27n19.hep.wisc.edu, user=cmsprod, path=/store/unmerged/RunIISummer20UL16MiniAOD/ZpMuMu_MZp-0p0005_Df-0p0251_TuneCP5_13TeV-madgraph-pythia8/MINIAODSIM/darkfactor_0p0251_106X_mcRun2_asymptotic_v13-v2/2540000/4D92B796-F97F-7F4F-87CD-15086CCF8289.root)  451-GridFTP-Error-String: Input/output error  451 End.   ' is not valid for storage.
2022-01-20 23:28:51,202:INFO:MSUnmerged:execute(): 
```

todor-ivanov:
> I'll figure a way out 

We do have the error codes, but need to find the proper mapping between their human readable representation and the error codes  from `gfal`'s point point of view.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -369,29 +392,53 @@ def _purgeTree(self, ctx, baseDirPfn):
...
             if dirEntry in ['.', '..']:
                 continue
             dirEntryPfn = baseDirPfn + dirEntry
             try:
                 entryStat = ctx.stat(dirEntryPfn)
-            except gfal2.GError:
+            except gfal2.GError as gfalExc:
                 e = sys.exc_info()[1]
                 if e.code == errno.ENOENT:
                     self.logger.error("MISSING dirEntry: %s", dirEntryPfn)
```
todor-ivanov, line 430: @amaltaro I'd like to draw your attention to this exact error. 

What it says in the general case is: if we try to stat a missing directory while browsing the structure, well it means something wrong has happened in the meantime - since the moment we started the iteration. And hence this should be a legit error (it follows now the same logic as it is in the default `gfal-ls` or `gfal-rm` APIs). But in our case, since we are working all the time with an outdated list of files and directories coming from the RucioConsitencyMonitor (and in the worst case scenario we may end up with an age of up to 1 week), many of those directories may have gone by the time we  try to clean them, either because of a cleanup job has deleted them or because of a previous run of the unmerged service.. etc. So in this case I would think of considering this as  a 'previous' success (because the directory is already gone - which is our goal actually) rather than an error and would call it `True` in the next line. I would move this `error` message here just to `warning` and simply  continue. This way will avoid repeatedly retrying the deletions of all the files which has been there a week ago - meaning if we to not account for the fact that the directory is already gone (even though it may have not been us during the current cycle who removed it), then every time all the thousands of files which are still persisting in the consistency lists would trigger a series of deletions  doomed to failure every polling cycle.  

Please let me know what you think?

todor-ivanov:
> We do have the error codes, but need to find the proper mapping between their human readable representation and the error codes from `gfal`'s point point of view.

After I've spent some time trying to find the proper mapping in the `gfal` codebase myself, I decided the ask the developers.   Here is the relevant issue:

https://github.com/cern-fts/gfal2-python/issues/3

todor-ivanov:
With may latest  commit I suggest a solution to the problem with the accounting of the `gfalError` codes by mapping them to the possix codes. And here is the question (with some more details and concerns) sent to the `cern-fts` developers for clarification if this is the best approach possible. 

https://github.com/cern-fts/gfal2-python/issues/3#issuecomment-1018438381

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
amaltaro: Hi Todor, yes, I agree. If we cannot delete content in a given directory because the directory no longer exists, I think it's a good compromise to say that deletion of that directory is successful.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -328,36 +331,56 @@ def cleanRSE(self, rse):
...
+                        rse['dirs']['deletedSuccess'].add(dirLfn)
+                        rse['counters']['dirsDeletedSuccess'] = len(rse['dirs']['deletedSuccess'])
+                        if dirLfn in rse['dirs']['toDelete']:
+                            rse['dirs']['toDelete'].remove(dirLfn)
+                        if dirLfn in rse['dirs']['deletedFail']:
+                            rse['dirs']['deletedFail'].remove(dirLfn)
+                        msg = "RSE: %s  Success deleting directory: %s"
+                        self.logger.info(msg, rse['name'], dirPfn)
+                    else:
+                        rse['dirs']['deletedFail'].add(dirLfn)
```
amaltaro, line 379: Do we need to protect against multiple insertions of the same dirLfn (between diff cycles)?

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -328,36 +331,56 @@ def cleanRSE(self, rse):
...
+                        rse['counters']['dirsDeletedFail'] = len(rse['dirs']['deletedFail'])
+                        msg = "RSE: %s Failed to purge directory: %s"
+                        self.logger.error(msg, rse['name'], dirPfn)
```
amaltaro, line 382: Do we know why it failed to be purged? What is `purgeSuccess` when it fails? If it's just a boolean, then there is nothing that we can do indeed.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -369,29 +392,53 @@ def _purgeTree(self, ctx, baseDirPfn):
...
         successList = []
 
+        # First test if baseDirPfn is actually a directory entry:
+        try:
+            entryStat = ctx.stat(baseDirPfn)
+        except gfal2.GError as gfalExc:
+            e = sys.exc_info()[1]
+            if e.code == errno.ENOENT:
+                self.logger.error("MISSING baseDir: %s", baseDirPfn)
+                successList.append(False)
```
amaltaro, line 406: I would suggest to replace these two lines by a simple `return False`. Idem for the else block below.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -369,29 +392,53 @@ def _purgeTree(self, ctx, baseDirPfn):
...
         """
+        # NOTE: It deletes only directories and does not try to unlink any file.
+
         successList = []
 
+        # First test if baseDirPfn is actually a directory entry:
+        try:
+            entryStat = ctx.stat(baseDirPfn)
+        except gfal2.GError as gfalExc:
+            e = sys.exc_info()[1]
```
amaltaro, line 403: This line we discussed over zoom yesterday. If you managed to find out that manipulating the `entryStat` object is enough, without this sys.exc_info call, then I would be in favor of that.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -369,29 +392,53 @@ def _purgeTree(self, ctx, baseDirPfn):
...
             try:
                 entryStat = ctx.stat(dirEntryPfn)
-            except gfal2.GError:
+            except gfal2.GError as gfalExc:
                 e = sys.exc_info()[1]
                 if e.code == errno.ENOENT:
                     self.logger.error("MISSING dirEntry: %s", dirEntryPfn)
                     successList.append(False)
                     return all(successList)
                 else:
```
amaltaro, line 433: Todor, I see that these blocks are manipulating a file/directory similar to what is done in the method `_purgeTree`. How about we create auxiliary methods for those, such that we can reuse code and even get some simple unit tests done for this?

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -716,17 +773,24 @@ def resetCounters(self, rseName=None, plineName=None):
             if rseName not in self.rseCounters:
                 self.rseCounters[rseName] = {}
             self.rseCounters[rseName]['totalNumFiles'] = 0
```
amaltaro, line 775: I have the impression we could simplify this `resetCounters` method. We repeat key names many times here, increasing the chances for a mistake. Could you please check whether anything can be done to enhance it?

amaltaro (changes requested):
Todor, this PR has lots of changes and the initial description is quite helpful, thanks!
I finally managed to have a look at the code and I left some comments/questions for your consideration.
As we discussed yesterday, please provide unit tests for the MSUnmergedRSE module (and anything else that is simple enough).
For future changes, you might want to keep them temporarily in different commits to make subsequent reviews easy.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Fixed!

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -301,13 +301,16 @@ def cleanRSE(self, rse):
...
         for dirLfn, fileLfnGen in rse['files']['toDelete'].items():
-            if self.msConfig['limitFilesPerRSE'] < 0 or \
-               rse['counters']['filesToDelete'] < self.msConfig['limitFilesPerRSE']:
+            if dirLfn in rse['dirs']['deletedSuccess']:
+                self.logger.info("RSE: %s, dir: %s already successfully deleted.", rse['name'], dirLfn)
+                continue
 
-                # First increment the dir counter:
-                rse['counters']['dirsToDelete'] += 1
+            if self.msConfig['limitFilesPerRSE'] < 0 or \
```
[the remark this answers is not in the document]
todor-ivanov: the message should actually be: 
"RSE: %s reached limit of files per RSE to be deleted. The rest will be tried on the nex cycle"

And done now.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -328,36 +331,56 @@ def cleanRSE(self, rse):
...
-                    except Exception as ex:
-                        msg = "Error while cleaning RSE: %s. "
-                        msg += "Will retry in the next cycle. Err: %s"
-                        self.logger.warning(msg, rse['name'], str(ex))
+                            self.logger.debug("RSE: %s, Dir: %s, delResult: %s",
+                                              rse['name'], dirLfn, pformat(delResult))
+                            for gfalErr in delResult:
+                                if gfalErr is None:
+                                    deletedSuccess += 1
+                                elif os.strerror(gfalErr.code) not in rse['counters']['gfalErrors']:
```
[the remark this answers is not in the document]
todor-ivanov: a good one!

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -328,36 +331,56 @@ def cleanRSE(self, rse):
...
+                                    rse['counters']['gfalErrors'][os.strerror(gfalErr.code)] = 1
+                                else:
+                                    rse['counters']['gfalErrors'][os.strerror(gfalErr.code)] += 1
+                        except Exception as ex:
```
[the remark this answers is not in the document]
todor-ivanov: done

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -328,36 +331,56 @@ def cleanRSE(self, rse):
...
+                            msg = "Error while cleaning RSE: %s. "
+                            msg += "Will retry in the next cycle. Err: %s"
+                            self.logger.warning(msg, rse['name'], str(ex))
+
+                    self.logger.debug("RSE: %s, Dir: %s, filesDeletedSuccess: %s",
```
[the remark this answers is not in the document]
todor-ivanov: done

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -328,36 +331,56 @@ def cleanRSE(self, rse):
...
+                                      rse['name'], dirLfn, deletedSuccess)
+                    rse['counters']['filesDeletedSuccess'] += deletedSuccess
+
+                    # Now clean the whole branch
+                    self.logger.debug("Purging dirEntry: %s:\n", dirPfn)
+                    purgeSuccess = self._purgeTree(ctx, dirPfn)
+                    if purgeSuccess:
+                        rse['dirs']['deletedSuccess'].add(dirLfn)
+                        rse['counters']['dirsDeletedSuccess'] = len(rse['dirs']['deletedSuccess'])
```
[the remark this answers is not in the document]
todor-ivanov: > Does it mean that that pfnSlice is actually a list of file paths within the same directory?

Correct! You have all the list generators working inside one part of the tree only and you have this line L307, which makes it go one dir at a time.   

>  Shouldn't we sum up the dirs deletedSuccess instead of having a direct assignment of that value here?

No, because this is the exact bug we are having right now in production. Since we use the database for state preservation in between the runs we already have a number in the field from previous runs and adding the current successful deletions we risk to make it only increase infinitely and not reflecting the proper state. And since the `rse['counters']['dirsDeletedSuccess']` is a set I assume this to me more accurate because we know exactly what we have removed from sure. And the operation complexity of `len(set)` is still `O(1)`, I think we are not losing anything here, but rather simply staying on the safe side.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: No need to: `rse['dirs']['deletedFail']` is a a `set`

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: it is just a `bool` indeed. The reason for failing is just logged into the exceptions handling during the process. If we were to develop a different behavior driven by the cause of failure, I admit the code would be definitely more precise, but also quite more complicate, because with `gfal` and all those many protocols to be used in the background it may become quite a lot to handle. I would rather leave it for `gfal` to do its excellent job in that regard, while we just count the final successful operations, and use the log messages for debugging the failures.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: already the case

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Alan, I do not get your comment here. The line you are commenting on is still inside the `_purgeTree` method. And yes those three blocks they do handle similar errors, but they are doing different operations:
> \# First test if baseDirPfn is actually a directory entry:

which is doing a `ctx.stat()` operation on the top level entry point.   
...

> \# Second recursively iterate down the tree:

which is doing `ctx.listdir()` and the actual recursion
...

> \# Finally remove the baseDir:

which is doing the actual removal `ctx.rmdir()`

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Already did it in the past, and couldn't figure out a way out of this. The repetition comes from the two different levels of counters we keep:
* The service counters which report at the end of every polling cycle and are preserved in the MSUnmerged instance
* The RSE counters that we need to keep track about in the database.

This is a legacy because we did not start with the stateful design from the very beginning. I am not telling there is no space for improvement though, but this would definitely cause changes in several other places in the code too.

amaltaro (changes requested):
Thanks for this follow up, Todor.
There are only a few open questions/comments left, and once we have some basic unit tests in place, I think it would be ready to go.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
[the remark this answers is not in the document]
todor-ivanov: > Isn't this rse['counters']['dirsDeletedSuccess'] a counter of directories for the whole RSE?

This is True.

>  If so, then I think we should be accumulating directory counters during a cleanup cycle, no?

This is also True.

> Instead of only counting the sub-tree for a given directory (a sub-set of all the directories to be cleaned up).

This is not what is happening here.  
Since we only accumulate in (and never remove from) the `rse['dirs']['deletedSuccess']` set, then every time we refer to its length, we also account for all previous successful deletions on the cost of a simple `O(1)` operation. And because we are working with a set here we avoid adding/counting the same directory twice in the counters. Something that would definitely happen if for some bizarre reason we iterate twice through an already deleted directory (of course we do try to take care of that not to happen but ... one never knows) and then because we consider the already missing directories for success if we blindly count  the current success and add it to the previous counter then we would count the same deletion twice. Hence the bug living in production right now. The counters due to a similar behavior only grow indefinitely.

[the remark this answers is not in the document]
todor-ivanov: And I see where the confusion may have com - I've made a typo in my previous comment which may have mislead you:
> And since the rse['counters']['dirsDeletedSuccess'] ...

Should have been 
> And since the rse['dirs']['deletedSuccess'] ...

Which is the set accumulating the names of all sucessfully purged branches from the tree. And  not only a sub-tree. 

Apologies for the mistake.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Sorry @amaltaro I still do not see you point. These 3 blocks are doing 3 different things and are similar (but not completely identical) only on the way how they handle the Gfal exception (under the same circumstances one of them return direct bool and the other two accumulate in a list). Are you suggesting me to create  a function only for handling these exceptions and the pass to it
1. the gfal exception
2. an external list etc 

If so I do not think this would be a good idea that would make us juggling with many conditions ..... 
I did try in the past to go with a single cycle for the whole recursion.... kind of achievable .. but then things get not that  obvious what is happening at the bottom. This separation of three steps seemed the most accurate to me.

amaltaro: Todor, you are right. It's not trivial and you are also using the stat response to trigger a recursive call, which I missed when I looked at this code. Sorry for the noise.

amaltaro:
@todor-ivanov we are having a testbed deployment tomorrow. But I think we can get this in on Thursday morning if we manage to get the final bits in this PR sorted. 
Unless there is a critical WM issue tomorrow, I'd suggest you to give 100% attention to this such that we can finish it in time.

todor-ivanov:
Hi @amaltaro The unit test for `MSUnmergedRSE` is in place. Jenkins is going to fail that until we have the `mongomock` library deployed [1]. I have tested it in my environment and all is working fine her is the outcome: [2].  
I am about to squash all the commits now. Please take another look at your convinience. 

[1]
https://github.com/mongomock/mongomock

[2]
```
(WMCore.venv3) [tivanov@tivanov-unit02 WMCore_t]$ nosetests MicroService_t/MSUnmerged_t/MSUnmergedRSE_t.py
..
Ran 2 tests in 0.002s

OK

```

test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmergedRSE_t.py
```diff
@@ -0,0 +1,69 @@
...
+import mongomock
+
+from pymongo import IndexModel
+
+from WMCore.MicroService.MSUnmerged.MSUnmerged import MSUnmerged, MSUnmergedRSE
+
+
+class MSUnmergedRSETest(unittest.TestCase):
+    """ Unit test for MSUnmergedRSE module """
+    # pylint: disable=W0212,C0301
```
amaltaro, line 19: Todor, can you please remove these pylint messages disabled here? If it's not simple to fix them, it's better to let them to get publish such that one day we can come back to them.

amaltaro (approved):
Todor, I am approving it, but please fix that line that I commented on (unit test).
I have made this change to the cmsdist as well:
https://github.com/cms-sw/cmsdist/pull/7598

can you please add it to this PR description?

If I fail to come back to this before you start working tomorrow morning, can you please:
* merge this PR
* ask Imran to merge cmsdist 7598 (if the build is successful)
* cut WMCore 2.0.0.pre2
* ask for a new testbed deployment
* communicate with Imran to clean the MongoDB collections in order to prepare for this new version of MSUnmerged
?

Thanks!

todor-ivanov:
Thanks @amaltaro  I am on it

amaltaro:
As Todor and myself noticed, these new unit tests were not executed by jenkins. My suspicious is that due to the missing python library in the unit test setup, it was actually skipped.

Now that we have specs up-to-date, I triggered unit tests in another random PR and we can see that unit tests have been properly considered and are succeeding:
https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12755/artifact/artifacts/PullRequestReport.html

So we are all good!

PR #10693, opened 2021-07-28
Keep workflows in running-open if they still have WQEs to be pulled by the agents

### LINKED ISSUE #10656: Requests in running-closed status whose WQEs are not in Running state, thus stuck requests
**Impact of the bug**
WMAgent, Workqueue

**Describe the bug**
In the P&R meeting, while trying to understand why some high priority requests are not being completed for weeks, we discovered that there are some workflows whose WQEs are in Available state, thus not acquired by agent. Since they are in running-closed state, by definition, all WQEs should be in state in Running or further according to [1]

Here are some requests: (Click on `Workqueue` in the following pages to check WQEs)
1. https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_B2G-RunIIAutumn18NanoAODv7-02608
2. https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_EXO-RunIISummer20UL16NanoAOD-00011
3. https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_B2G-RunIIFall17NanoAODv7-02817
and more.

Here is the list of "Late" workflows. [Link](https://its.cern.ch/jira/issues/?jql=project%20%3D%20CMSCOMPPR%20AND%20status%20in%20(Open%2C%20%22In%20Progress%22%2C%20Reopened)%20AND%20labels%20%3D%20Late%20ORDER%20BY%20updated%20DESC%2C%20lastViewed%20DESC) I haven't checked all of them, but probably many of them suffers from this issue.

[1] https://github.com/dmwm/WMCore/wiki/Request-Status#running-closed 

**How to reproduce it**
Check the requests above

**Expected behavior**
Jobs must have been created for each WQE by `running-closed` status.

None

amaltaro:
Fixes #10656 

#### Status
ready

#### Description
This PR provides a change of the `running-open` and `running-closed` request status definition (not sure whether I should call it a bug or not), where the current behaviour is:
* `running-open`: as long as a workflow has at least one WQE in `Acquired` status, the workflow would remain in `running-open`. Regardless of whether all the rest is still Available, or it's all Done. In other words, at least one WQE is yet to be processed.
* `running-closed`: if the `running-open` case above is not fulfilled... as long as a workflow has at least one WQE in `Running` status, the workflow would remain in `running-closed`. Regardless of whether all the rest is still Available, or it's all Done. In other words, at least 1 WQEs is being processed.

and the new behaviour - with these changes - will be:
* `running-open`: if at least one GQE has been already pulled by an agent (thus, in Acquired or beyond) AND there are still WQEs waiting in GQ (Available/Negotiating/Acquired, thus not yet in WMBS).
* `running-closed`: when all of the workqueue elements are at least in Running status, thus no workqueue elements left in GQ (Available/Negotiating/Acquired).

NOTE: second commit has a temporary fix to avoid attempts from `running-closed` to `running-open` (transition would not happen, but we would see tons of exceptions in the logs), which will happen to many active workflows in the production system. That workaround should be removed in ~6 months from now (mentioned in the code as well)

TODO: wiki https://github.com/dmwm/WMCore/wiki/Request-Status needs to be updated

No, in the sense that some workflows currently in `running-closed` status would actually be considered as `running-open`, according to the new logic.

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
+++ b/src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
@@ -56,6 +56,11 @@ def moveForwardStatus(reqmgrSvc, wfStatusDict, logger):
                 reqmgrSvc.updateRequestStatus(wf, stateFromGQ)
                 logger.info("%s in %s moved to %s", wf, status, stateFromGQ)
                 continue
+            # FIXME: remove this elif once the current active workflows are out
+            # of the system. January 2022 should be good (HG2201)
+            elif stateFromGQ == "running-open" and status == "running-closed":
+                logger.warning("%s in %s, but it should actually be in %s", wf, status, stateFromGQ)
+                continue
 
             try:
                 i = nextStatus.index(stateFromGQ)
--- a/src/python/WMCore/Services/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/Services/WorkQueue/WorkQueue.py
@@ -15,47 +15,46 @@ def convertWQElementsStatusToWFStatus(elementsStatusSet):
     :returns: request status
 
     Here is the mapping between request status and it GQE status
-    1. acquired:  if all the GQEs are either Available or Negotiating -  Work is still in GQ but not LQ
-    2. running-open: if at least one of the GQEs are in Acquired status - at least some work is in LQ
-    3. running-closed: if all the GQEs are in Running or complted status
-                       no Available Negotiating or Acquired status. all the work is in WMBS db (in agents)
-    4. completed: if all the GQEs are in 'Done', 'Canceled' status.
-                   - all work is finsed in wmbs (excluding cleanup, logcollect)
-    5. failed: if all the GQEs are in Failed status. If the workflow has multiple GQEs and only a few are
-               in Failed status, then just follow the usual request status.
+    1. acquired:  all the GQEs are either Available or Negotiating.
+        Work is still in GQ, but not LQ.
+    2. running-open: at least one of the GQEs are in Acquired status.
+    3. running-closed: all the GQEs are in Running or beyond status.
+        No Available/Negotiating/Acquired status, all the work is in WMBS db (in agents)
+    4. completed: all the GQEs are in a final status, like Done/Canceled/Failed.
+        All work is finished in WMBS (excluding cleanup and logcollect)
+    5. failed: all the GQEs are in Failed status. If the workflow has multiple GQEs
+        and only a few are in Failed status, then just follow the usual request status.
 
-    CancelRequest status treated as transient status.
+    NOTE: CancelRequested status is a transient status and it should not trigger
+    any request status transition (thus, None gets returned).
     """
-    if len(elementsStatusSet) == 0:
+    if not elementsStatusSet:
         return None
 
+    forceCompleted = set(["CancelRequested"])
     available = set(["Available", "Negotiating", "Failed"])
     acquired = set(["Acquired"])
     running = set(["Running"])
-    completed = set(['Done', 'Canceled', "Failed"])
+    runningOpen = set(["Available", "Negotiating", "Acquired"])
+    runningClosed = set(["Running", "Done", "Canceled"])
+    completed = set(["Done", "Canceled", "Failed"])
     failed = set(["Failed"])
 
-    if elementsStatusSet == acquired:
+    if forceCompleted <= elementsStatusSet:  # at least 1 WQE in CancelRequested
+        return None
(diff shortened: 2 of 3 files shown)


amaltaro:
Even though the current request status transition looks inconsistent to me, I'm honestly not very happy with the changes proposed here. It will swap our workflow distribution, making many workflows to actually remain in `running-open` for a longer time, and a smaller fraction remaining in `running-closed` - those that are supposedly close to completion.

This `running-open` status was added back then when we had DBS2, such that new workqueue elements could be created while a workflow was already being processed in the system. That means, global workqueue might get overloaded with these changes, trying to look up for new data and create new elements. I have a hard time assessing how much it's going to be.

@todor-ivanov @haozturk your input on what is proposed here would be greatly appreciated. Perhaps @klannon as well.

haozturk:
> It will swap our workflow distribution, making many workflows to actually remain in running-open for a longer time, and a smaller fraction remaining in running-closed - those that are supposedly close to completion.

Why do you think this is an issue? I think the important point is to have distinct differences between different states, so that we can easily understand how a workflow is performing. The existing definitions do not give us much information. Besides, it is confusing and it took quite some time to understand what it actually is. I agree that w/ the new scheme we'll not see much of an improvement in terms of understanding of workflow's progress, but at least it's more explicit and easy to understand. Btw, do we have a finite state machine showing the state transitions of WQEs? I am lacking some knowledge there.

Anyhow, this is currently unrelated to the critical unmerged issue, and I think we don't need to rush. I'd propose taking our time and coming up w/ the best scheme which will benefit us the most.

amaltaro:
Sorry for missing these questions, Hasan.

> Why do you think this is an issue?

Because there is some logic in global/local workqueue to keep looking up input data for workflows in `running-open` state. AFAIK this is a legacy code from the time we had DBS2; but it could be that it's still functional for cases where a workflow is processing data that is still being produced (like RAW data from T0).

> Btw, do we have a finite state machine showing the state transitions of WQEs?

I didn't find any state machine defining the states and the allowed transition. The WQE states can be found here though:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py#L14-L15

If the running-open workflows wouldn't cause too much unnecessary load in the system - and I think they will - this would be a good candidate for the upcoming CMSWEB release.

@todor-ivanov please let me know if you have any thoughts on this item.

todor-ivanov:
Hi @amaltaro , Things does look more explicit to me with this change. On the other side we  should avoid any kind of additional load to GQ, but if a workflow still has data to proceed it should be logical to stay in the work queue. Whether that queue should be the global or the local depends on your second statement:
> it could be that it's still functional for cases where a workflow is processing data that is still being produced (like RAW data from T0).

Because if data is still produced... there is no way to know in advance how many WQE would eventually be created in the GQ, so in this case it is inevitable to keep it there.

amaltaro:
I still have to check a few things, specially the `running-open` functionality. But I'm heavily considering in adding this for the upcoming release. Todor/Hasan, please let me know if you have any comments and/or concerns with this.

amaltaro:
test this please

amaltaro:
I missed this patch yesterday, but I'd rather push this in this month than waiting and piling more things for the next months.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
```diff
@@ -56,6 +56,11 @@ def moveForwardStatus(reqmgrSvc, wfStatusDict, logger):
                 reqmgrSvc.updateRequestStatus(wf, stateFromGQ)
                 logger.info("%s in %s moved to %s", wf, status, stateFromGQ)
                 continue
+            # FIXME: remove this elif once the current active workflows are out
```
todor-ivanov, line 59: This is good to have this FIXME here, but I'd also create a separate issue and let it sit open until the time comes.

todor-ivanov (approved):
Thanks @amaltaro! 
Changes look good to me.

src/python/WMCore/ReqMgr/CherryPyThreads/StatusChangeTasks.py
amaltaro: That's a good idea. Here is an issue to keep track of this: https://github.com/dmwm/WMCore/issues/10891

amaltaro:
Thanks for the review, Todor. I have done some quick tests on my VM as well and everything looks good. We should look at the testbed logs though to see whether it finds other corner cases that we fail to have in our private VM.

amaltaro:
Wiki - linked in the initial description - has been updated accordingly.

PR #10695, opened 2021-07-28
Performance improvements for WorkQueueManagerCleaner thread

### LINKED ISSUE #8477: WorkQueueManagerCleaner is very slow
And so it raises a heartbeat alarm.
I don't know exactly what it's doing to be honest, but I see it taking a few hours per cycle, on vocms0250
```
2018-02-21 01:18:34,045:140688431773440:INFO:BaseWorkerThread:WorkQueueManagerCleaner took 12439.739 secs to execute
2018-02-21 04:47:52,849:140688431773440:INFO:BaseWorkerThread:WorkQueueManagerCleaner took 12377.496 secs to execute
2018-02-21 07:00:58,592:140688431773440:INFO:BaseWorkerThread:WorkQueueManagerCleaner took 7805.157 secs to execute
etc etc
```

probably there is some room for improvements here. Or we slice the amount of work it has to do. Or ...

amaltaro:
Fixes #8477 

#### Status
ready

#### Description
Some performance improvements for the WorkQueueManagerCleaner agent thread.
The main clearly wrong and inefficient code was related to the `wmbsSubscriptionStatus` call, which was executed for every single workflow in the agent, even though it collects data for all the workflows in the system. So, instead of calling it for every workflow, call it before getting into the for loop and just use that information to sync workqueue elements with the WMBS information.

Also added some important logs in INFO level, such that we have a clue of what the thread is doing.

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -703,14 +703,14 @@ class WorkQueue(WorkQueueBase):
         return len(work)
 
     def status(self, status=None, elementIDs=None,
-               dictKey=None, syncWithWMBS=False, loadSpec=False,
+               dictKey=None, wmbsInfo=None, loadSpec=False,
                **filters):
         """
         Return elements in the queue.
 
         status, elementIDs & filters are 'AND'ed together to filter elements.
         dictKey returns the output as a dict with the dictKey as the key.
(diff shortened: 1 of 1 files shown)


amaltaro:
The PhEDEx-related code will be removed in its own issue/PR.

todor-ivanov (approved):
looks good to me

amaltaro:
Thanks, Todor.

PR #10696, opened 2021-07-29
Provide a configurable weight value for MSTransferor Rucio rules

### LINKED ISSUE #10617: [MS] Create input data placement rules with a configurable weight expression
**Impact of the new feature**
MSTransferor

Yes, in addition to the extra complexity that we need to introduce to the WMCore microservice code, it can also result in non-ideal and unbalanced data distribution.

We should start providing a `weight` expression to the rucio creation call. The initial idea would be to use `ddm_quota` as the weight expression, but this is likely going to change in the coming weeks/months, so we should make this weight expression configurable via MS configuration.

Thus, changes to both service configuration and WMCore source code are required here.

None

**Additional context**
None

amaltaro:
Fixes #10617 

#### Status
not-tested

#### Description
With this PR, we start creating input data placement Rucio rules with a default weight expression:
```
weight="ddm_quota"
```

this weight expression has been made configurable through the service configuration though.

yes

#### Related PRs
None

It requires the following configuration/deployment changes:
https://github.com/dmwm/deployment/pull/1073

Note that it contains also a couple of MSUnmerged configuration changes, as discussed with Todor.

### DIFF
--- a/src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
+++ b/src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
@@ -80,6 +80,8 @@ class MSTransferor(MSCore):
         # Send warning messages for any data transfer above this threshold.
         # Set to negative to ignore.
         self.msConfig.setdefault("warningTransferThreshold", 100. * (1000 ** 4))  # 100TB
+        # weight expression for the input replication rules
+        self.msConfig.setdefault("rucioRuleWeight", 'ddm_quota')
 
         quotaAccount = self.msConfig["rucioAccount"]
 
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
```diff
@@ -670,20 +672,22 @@ def makeTransferRucio(self, wflow, dataIn, subLevel, blocks, dataSize, nodes, no
...
-                'comment': 'WMCore MSTransferor input data placement'}
+        ruleAttrs = {'copies': 1,
+                     'activity': 'Production Input',
+                     'lifetime': self.msConfig['rulesLifetime'],
+                     'account': self.msConfig['rucioAccount'],
+                     'grouping': subLevel,
+                     'meta': {'workflow_group': wflow.getWorkflowGroup()},
+                     'comment': 'WMCore MSTransferor input data placement'}
+        # add a configurable weight value
+        ruleAttrs["weight"] = self.msConfig['rucioRuleWeight']
```
todor-ivanov, line 683: Why not in the `ruleAttrs={}` definition itself?

todor-ivanov (approved):
@amaltaro,  I left one minor comment inline, but do not feel like changing it. It will work perfectly fine as it is right now. So I think the change you suggest here is looking already good enough.

src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
amaltaro: Good question! Let me quickly update it. Thanks

amaltaro:
Thanks for the prompt review, Todor.

PR #10698, opened 2021-07-29
Import micro services on demand in MSManager.

### LINKED ISSUE #10697: MSUnmerged: Fix imports on a MicroService basis for the MSManager
**Impact of the bug**
MSUnmerged

**Describe the bug**
In the process  of developing MSUnmerged we had to introduce the dependency from the `gfal2` library, which is installed ont by the regular `rpm` based methods we use for the rest of our services and not from the cmsrepository, but it is using the standard OS `yum` based deployment mechanisms. This implies a requirement to add some more extra and dangerous sources to the `$PYTHONPATH`, like the OS path, for the running service. In order to isolate this only for this micro service  we created a separate image called `reqmgr2ms-unmerged` in docker hub, which inherits from `reqmgr2ms` and has the required (dangerous) change. A consequence we did overlook was that MSManager imports all the micro services definitions at the very beginning,  and is actually breaking all the rest of the micro services, which do not have the `gfal` library installed. All details and some log dumps from the installation process and suggested solutions are preserved in the following few comments: 
* https://github.com/dmwm/WMCore/issues/10682#issuecomment-888466001
* https://github.com/dmwm/WMCore/issues/10682#issuecomment-888511296

**How to reproduce it**
Try to deploy any micro service in K8 

**Expected behavior**
To have isolated dependencies for every 

todor-ivanov:
Fixes #10697 

#### Status
ready

#### Description
In order to fix the MicroService general dependency from `gfal2`problem we we suggest to import the different micro services code only when it is needed.

YES 

#### Related PRs
No

No

### DIFF
--- a/src/python/WMCore/MicroService/MSManager.py
+++ b/src/python/WMCore/MicroService/MSManager.py
@@ -32,11 +32,6 @@ from collections import deque
 
 # WMCore modules
 from WMCore.MicroService.Tools.Common import getMSLogger
-from WMCore.MicroService.MSTransferor.MSTransferor import MSTransferor
-from WMCore.MicroService.MSMonitor.MSMonitor import MSMonitor
-from WMCore.MicroService.MSOutput.MSOutput import MSOutput
-from WMCore.MicroService.MSRuleCleaner.MSRuleCleaner import MSRuleCleaner
-from WMCore.MicroService.MSUnmerged.MSUnmerged import MSUnmerged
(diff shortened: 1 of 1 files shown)


amaltaro (approved):
This looks good to me, Todor.

And this was likely the cause for the problem I mentioned here:
https://github.com/dmwm/WMCore/pull/10671#issuecomment-886482640

PR #10699, opened 2021-07-29
MSOutput: do not fail workflow if there are containers unknown to Rucio

amaltaro:
Fixes #10368 

#### Status
not-tested

#### Description
This PR provides two changes:
* bug-fix to producing/consuming workflows where not all the output datasets have been produced and published to DBS/Rucio. In the current state, MSOutput keeps trying to produce them over and over, forever!
* create an alert (email + slack notification) via AlertManager Prometheus system, for each workflow with their list of missing output datasets.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

PR #10717, opened 2021-08-03
MSOutput: skip inexistent containers in the Tape data placement

### LINKED ISSUE #10368: MSOutput - unhandled exception in _getDatasetSize()
**Impact of the bug**
MSOutput

**Describe the bug**
While trying to determine dataset size if the service for any reason fails to do so (e.g. fails to get DID info from Rucios etc.) we do not handle the exception [1] but just let it reach the upper level pipeline exception handler, while proper actions  should be taken at place. In result we pollute the logs with exceptions, which in many cases could be just predictable/expected behavior.   

[1]
https://github.com/dmwm/WMCore/blob/e45a4425c5437d194d0d02814b7011c752c33548/src/python/WMCore/MicroService/MSOutput/MSOutput.py#L696

[2]
```
2021-03-23 01:14:48,521:ERROR:MSOutput: MSOutputProducer Pipeline Possibly broken read from ReqMgr2 API or other. Err: 'bytes'. Continue to the next document.
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSOutput.py", line 545, in msOutputProducer
    msPipeline.run(request)
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/Utils/Pipeline.py", line 140, in run
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/Utils/Pipeline.py", line 140, in <lambda>
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/Utils/Pipeline.py", line 72, in __call__
    return self.run(obj)
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/Utils/Pipeline.py", line 75, in run
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSOutput.py", line 622, in docInfoUpdate
    bytesSize = self._getDatasetSize(dataItem['Dataset'])
  File "/data/srv/HG2103g/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.4.6.pre9/lib/python2.7/site-packages/WMCore/MicroService/Unified/MSOutput.py", line 695, in _getDatasetSize
    return didInfo["bytes"]
KeyError: 'bytes'
```

**How to reproduce it**
Just look at the MSOutput  logs

**Expected behavior**
In case the situation is expected, we should handle the exception at the place where it happens (even if we need to create a new class of exceptions), and propagate only the unexpected ones upstream.
 
N/A

amaltaro:
Fixes #10368 (again!)

#### Status
ready

#### Description
When evaluating whether a container can be placed at a Tape endpoint, take into consideration the container size, which will say whether the container has been produced or not.

YES

#### Related PRs
Complement to https://github.com/dmwm/WMCore/pull/10699

None

### DIFF
--- a/src/python/WMCore/MicroService/MSOutput/MSOutput.py
+++ b/src/python/WMCore/MicroService/MSOutput/MSOutput.py
@@ -797,11 +797,11 @@ class MSOutput(MSCore):
 
     def canDatasetGoToTape(self, dataItem, workflow):
         """
-        This function evaluates whether a dataset can be passed to the
+        This function evaluates whether a container can be passed to the
         Data Management system, considering the following configurations:
-          1) list of blacklisted tiers in the MicroService configuration
-          2) list of white listed tiers bypassing the Unified configuration
-          3) list of black and white listed tiers in the Unified config
+          1) list of tiers banned in the MicroService configuration
+          2) list of tiers allowed to bypass the Unified configuration
+          3) list of allowed and banned tiers in the Unified config
         :param dataItem: output map dictionary present in the MongoDB record
         :param workflow: MSOutputTemplate object retrieved from MongoDB
         :return: True if the dataset is allowed to pass, False otherwise
@@ -815,6 +815,9 @@ class MSOutput(MSCore):
             return False
         if workflow['RequestType'] == "Resubmission":
             # their parent/original workflow will take care of all the data placement
+            return False
+        if int(dataItem['DatasetSize']) == 0:
+            # container hasn't been produced in the workflow
             return False
 
         if dataItem['TapeRuleID']:


amaltaro:
I have changed the MSOutput code in k8s production, with this 3 lines change, and restarted the service inside the POD.
This is now working as expected.

I am not planning to cut a new release and image with this fix though. Will leave the fix in MSOutput production - until that POD gets restarted - and push this in in the next CMSWEB cycle.

amaltaro:
It turns out the amount of email notifications is too big with the current settings and the current service design.

I have received at least 5 emails for workflow:
pdmvserv_RVCMSSW_12_0_0_pre5BuMixing_14_210730_160440_5107

and the main reason is that we keep producing workflow documents as long as the request is sitting in closed-out/announced status, thus relying on the MongoDB unique key mechanism to skip workflow documents duplication. 

Problem is, mongodb document insertion is the last step when producing documents in the Producer thread.. need to think in a cheap way to stop it.

amaltaro:
New issue #10722 has been created to address the multiple notifications.

amaltaro:
Just a reminder: this patch has been directly applied to the ms-output POD in production. If the POD is restarted, this fix will be removed as well. When we create a deploy a new image, the fix will then be permanent (next CMSWEB release). Merging.

PR #10720, opened 2021-08-03
Implement py3 sorting for JobSplitting

### LINKED ISSUE #10627: Broken unittests in WMCore_t/WMBS_t/JobSplitting_t/ in py3
**Impact of the bug**
Some unittests in `WMCore_t/WMBS_t/JobSplitting_t` are failing in py3 and not in py2.

**Describe the bug**
The following unittests are failing in py3 only and this may mean a regression caused by our modernization

- [ ] `test/python/WMCore_t/WMBS_t/JobSplitting_t/EventBased_t.py`
- [ ] `test/python/WMCore_t/WMBS_t/JobSplitting_t/ParentlessMergeBySize_t.py`
- [ ] `test/python/WMCore_t/WMBS_t/JobSplitting_t/RunBased_t.py`
- [ ] `test/python/WMCore_t/WMBS_t/JobSplitting_t/SplitFileBased_t.py`
- [ ] `test/python/WMCore_t/WMBS_t/JobSplitting_t/WMBSMergeBySize_t.py`

**How to reproduce it**
run the unittests

**Expected behavior**
the unittests should pass

Check the failing unittests in py3 at [this webpage from jenkins](https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCorePy3-UnitTests/lastBuild/)

goughes:
Fixes #10627

#### Status
ready

#### Description
Implements the necessary changes to allow for python3 sorting based on comparison functions

YES

#### Related PRs
NA

NA

### DIFF
--- a/src/python/WMCore/JobSplitting/JobFactory.py
+++ b/src/python/WMCore/JobSplitting/JobFactory.py
@@ -8,6 +8,7 @@ from builtins import range
 
 import logging
 import threading
+import operator
 
 from WMCore.DAOFactory import DAOFactory
 from WMCore.DataStructs.WMObject import WMObject
@@ -242,7 +243,11 @@ class JobFactory(WMObject):
             logging.debug("About to load files by DAO")
             fileset = self.subscription.availableFiles(limit=self.limit, doingJobSplitting=True)
 
-        for fileInfo in fileset:
+        # Reverse sorting this set by location is required to match how sorting was done in py2
+        # Unittests that rely on this sorting method
+        # - WMCore_t.WMBS_t.JobSplitting_t.FileBased_t.FileBasedTest:testSiteWhiteBlacklist
+        # - WMCore_t.WMBS_t.JobSplitting_t.FileBased_t.FileBasedTest:testLocationSplit
+        for fileInfo in sorted(fileset, key=operator.itemgetter('locations'), reverse=True):
 
             if self.trustSitelists:
                 locSet = frozenset(set(['AAA']))
--- a/src/python/WMCore/JobSplitting/ParentlessMergeBySize.py
+++ b/src/python/WMCore/JobSplitting/ParentlessMergeBySize.py
@@ -8,6 +8,7 @@ from __future__ import division
 
 import time
 import threading
+from functools import cmp_to_key
 
 from WMCore.WMBS.File import File
 from WMCore.DataStructs.Run import Run
@@ -78,7 +79,7 @@ class ParentlessMergeBySize(JobFactory):
             self.newGroup()
 
         self.newJob(name = self.getJobName())
-        mergeFiles.sort(fileCompare)
+        mergeFiles.sort(key=cmp_to_key(fileCompare))
 
         jobSize = 0
         largestFile = 0
@@ -123,7 +124,7 @@ class ParentlessMergeBySize(JobFactory):
         mergeJobFiles    = []
         earliestInsert   = 999999999999999
 
-        mergeableFiles.sort(fileCompare)
+        mergeableFiles.sort(key=cmp_to_key(fileCompare))
 
         for mergeableFile in mergeableFiles:
             if mergeableFile["file_size"] > self.maxMergeSize or \
--- a/src/python/WMCore/JobSplitting/SplitFileBased.py
+++ b/src/python/WMCore/JobSplitting/SplitFileBased.py
@@ -9,6 +9,7 @@ single job for each merge unit.
 """
 
 import threading
+from functools import cmp_to_key
 
 from WMCore.DAOFactory import DAOFactory
 from WMCore.DataStructs.Run import Run
@@ -58,12 +59,10 @@ def sortedFilesFromMergeUnits(mergeUnits):
     Given a list of merge units sort them and the files that they contain.
     Return a list of sorted WMBS File structures.
     """
-    mergeUnits.sort(mergeUnitCompare)
-
+    mergeUnits.sort(key=cmp_to_key(mergeUnitCompare))
     sortedFiles = []
     for mergeUnit in mergeUnits:
-        mergeUnit["files"].sort(fileCompare)
-
+        mergeUnit["files"].sort(key=cmp_to_key(fileCompare))
         for file in mergeUnit["files"]:
             newFile = File(id=file["file_id"], lfn=file["file_lfn"],
                            events=file["file_events"])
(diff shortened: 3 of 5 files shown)


mapellidario:
Hi @goughes, i had a quick look at the changes and they look terrific! :) Before I send the proper review, can I ask you if you had a look at

- WMCore_t.WMBS_t.JobSplitting_t.FileBased_t.FileBasedTest:testSiteWhiteBlacklist changed from success to failure
- WMCore_t.WMBS_t.JobSplitting_t.FileBased_t.FileBasedTest:testLocationSplit
- WMCore_t.WMBS_t.JobSplitting_t.RunBased_t.EventBasedTest:testMultipleRunsCombine changed from success to failure
- WMCore_t.WMBS_t.JobSplitting_t.EventBased_t.EventBasedTest:test100EventMultipleSite

? These have been marked as unstable likely because of non reproducible sorting. Thanks!

goughes:
Indeed, testing this locally shows these tests do not sort reliably. I will take a closer look.

goughes:
I added some debug statements to the code and found that `testSiteWhiteBlacklist` and `testLocationSplit` do not call any of the code I changed. In py2 these tests are successful and stable. There must be some other place that is sorting (or not sorting) these.

goughes:
The fileset that comes back here does not get sorted consistently in py3

https://github.com/dmwm/WMCore/blob/7e6a0e5a8c9b75a8f8045c18d89cc362ff12dba0/src/python/WMCore/JobSplitting/JobFactory.py#L245

I committed a change that fixes the unit tests, but it is not a good change. :stuck_out_tongue: I'm getting into unfamiliar WMCore territory here... Does the order of this fileset actually matter? If it does not, then I can update the unit tests so they don't depend on the jobGroups always getting sorted one way. If it does, then we should discuss further. @mapellidario @amaltaro

goughes:
test this please

goughes:
test this please

amaltaro:
@goughes regarding your question on the JobFactory fileset order. I was actually looking at that code today, so something is still fresh in my mind :)
What do you think about sorting this `fileIDs` list in this line: https://github.com/dmwm/WMCore/blob/7e6a0e5a8c9b75a8f8045c18d89cc362ff12dba0/src/python/WMCore/JobSplitting/JobFactory.py#L366
?

It's meant to be a list of integers, and it would give us a consistent behaviour between python versions.

goughes:
Sorting on L366 of JobFactory would work for only some of the cases due to the logic here:

https://github.com/dmwm/WMCore/blob/7e6a0e5a8c9b75a8f8045c18d89cc362ff12dba0/src/python/WMCore/JobSplitting/JobFactory.py#L237

Which (at least for the unit test case) creates the fileset from the Subscription class:

https://github.com/dmwm/WMCore/blob/7e6a0e5a8c9b75a8f8045c18d89cc362ff12dba0/src/python/WMCore/DataStructs/Subscription.py#L147

So we would need to sort there as well. Or even here?

https://github.com/dmwm/WMCore/blob/7e6a0e5a8c9b75a8f8045c18d89cc362ff12dba0/src/python/WMCore/DataStructs/Fileset.py#L67

In any of these cases, I need to sort specifically with `reverse=True` to satisfy the unit test. So I think the unit test will updated as well.

amaltaro:
@goughes Erik, I just noticed one of the problems I solved here: https://github.com/dmwm/WMCore/pull/10739
you had already resolved in this PR :) It might be worth it discussing it in our chat today.

amaltaro:
From Dario's comment in: https://github.com/dmwm/WMCore/pull/10720#issuecomment-892021617
only the RunBased_t is really failing with something else than sorting the input files/fileset. So I'd suggest to remove those 3 other unit tests from the unstable list (if they are all marked as unstable):

@goughes Erik, can you please remove the relevant tests from the unstable list in:
https://github.com/dmwm/WMCore/blob/master/test/etc/UnstableTests.txt
?

Once that's done, I believe this is ready to be merged, right? I will then try to put it in in a new release tonight.

goughes:
test this please

goughes:
test this please

goughes:
Here is a list of unstable tests after my last commit

- `WMCore_t.WMBS_t.JobSplitting_t.RunBased_t.EventBasedTest:testMultipleRunsCombine`
- `WMCore_t.WMBS_t.JobSplitting_t.EventBased_t.EventBasedTest:test100EventMultipleSite`

I will add some comments and squash my second commit shortly.

mapellidario:
Regarding `WMCore_t.WMBS_t.JobSplitting_t.EventBased_t.EventBasedTest:test100EventMultipleSite`, I see going from failure to success consistently in last jenkins reports. Did i miss anything?

goughes:
@mapellidario I was able to verify the `WMCore_t.WMBS_t.JobSplitting_t.RunBased_t.EventBasedTest:testMultipleRunsCombine
` test via local testing. It definitely is unstable. But, for `test100EventMultipleSite` I was relying on the output from Jenkins due to a cert error running the test locally. Maybe I misinterpreted all the results. Are you able to run this test manually with your setup?

goughes:
I fixed my issue with the `test100EventMultipleSite` unit test and verified it runs with consistent success locally. I removed this test from the unstable list with my last commit. I think this one is ready for review @mapellidario

mapellidario (approved):
Thanks Erik! This looks great! Using [functools.cmp_to_key](https://docs.python.org/3/library/functools.html#functools.cmp_to_key) is a very nice solution, definitely better than the custom hacky function i would have come up with! :)
And also sorry for having sidetracked you by adding a couple more unittests into the mix

amaltaro:
Erik, this code looks good to me, however, let me ask you a question.
AFAIU this functools code is a way to implement sorting compatible with both python versions, right? So I wonder if we could change the code from, for instance:
```
         mergeableFiles.sort(key=cmp_to_key(fileCompare)) if PY3 else mergeableFiles.sort(fileCompare)
```
to simply
```
         mergeableFiles.sort(key=cmp_to_key(fileCompare))
```
thus, not worrying about which python version we are working on.


I have executed unit tests for `WMCore_t/WMBS_t/JobSplitting_t/ParentlessMergeBySize_t.py`, with this variation, and all the PY2 and PY3 relevant unittests are successful. I could be missing something though(?)

goughes:
@amaltaro you are correct, this logic is not needed. Thanks for the suggestion! It has been removed with my most recent commit and unit tests look good.

amaltaro:
Wonderful! Thank you very much for the prompt follow up!

Issue #10555, opened 2021-05-29
[py2py3] Check unittests from Slice 24 in py3


mapellidario:
#### Intro

We need to make sure that the work done in the migration slices
is py3 compatible. In particular, we should make sure that the
unittests that are involved in each slice do not fail.

The schedule for this validation / bug-fixing campaign is kept at #10422

#### Details

This issue is about checking that the unittests from Slice 24,
issue #10148 and PR #10371, do not fail.
A non-exhaustive list of the ut that need to pass is

```
test/python/WMCore_t/ReqMgr_t/Utils_t/Validation_t.py
test/python/WMComponent_t/JobCreator_t/JobCreator_t.py
test/python/WMComponent_t/TaskArchiver_t/TaskArchiver_t.py
test/python/WMCore_t/ReqMgr_t/Service_t/Auxiliary_t.py

```

PR #10724, opened 2021-08-04
[py2py3] fix JobCreator_t.py

mapellidario:
related to #10555 

#### Status
Ready

#### Description
This fixes a unittest from #10555, but should not close it because TaskArchiver_t has not been fied yet.

The solution was simply a quick change in how a pickle is loaded in the unittest itself, nothing serious.

yes

#### Related PRs
nope

nope


amaltaro (approved):
Perfect! Thanks, Dario.

Issue #10642, opened 2021-06-29
Deploy SLC7 Python3 packages in CVMFS


amaltaro:
**Impact of the new feature**
WMAgent (grid runtime)

**Is your feature request related to a problem? Please describe.**
As we move WMAgents to python3 stack, we need to make sure that the runtime code/environment uses the same python interpreter/setup.
WARNING: This needs to be carefully done though, the way submit.sh is implemented right now, it would start using the new python3 version as soon as it's visible in CVMFS.

**Describe the solution you'd like**
Contact Shahzad and ask if he can deploy the following RPMs (latest RPMs at this very moment are):
external+py3-future+0.18.2-1-1.slc7_amd64_gcc630.rpm
external+python3+3.8.2-comp-1-1.slc7_amd64_gcc630.rpm
from: http://cmsrep.cern.ch/cgi-bin/repos/comp/slc7_amd64_gcc630?C=M;O=D

into this CVMFS path: /cvmfs/cms.cern.ch/COMP/slc7_amd64_gcc630/

**Describe alternatives you've considered**
Not really an alternative, but I wonder how we are going to deal with SLC6 workflows running from the new Python3 WMAgent?!?!

**Additional context**
This issue is a pre-requirement for: https://github.com/dmwm/WMCore/issues/10641

amaltaro:
This has been discussed in a private thread with Shahzad and a few others, subject: "Assistance needed for python3 in SLC6".

Shahzad deployed the following COMP RPMs:
 external+python3+3.8.2-comp-1-1.slc7_amd64_gcc630.rpm
 external+py3-future+0.18.2-1-1.slc7_amd64_gcc630.rpm

at this path: `/cvmfs/cms.cern.ch/COMP/slc7_amd64_gcc630/`. Deployment to CVMFS happened at ~13:00 CERN time, on August 5th.

Regarding those same packages for slc6_* ScramArch, we do not have our own COMP future and python3 packages. Shahzad and David L. suggested creating a symlink to those packages in one of the CMSSW releases. These:
```
/cvmfs/cms.cern.ch/COMP/slc6_amd64_gcc700/external/python3/3.6.4
/cvmfs/cms.cern.ch/COMP/slc6_amd64_gcc700/external/py2-future/0.16.0
```
paths have been made available for us as well. This slc6_* setup still needs to be tested though, since we do not run our software against those package versions.

amaltaro:
[...]

PR #10726, opened 2021-08-05
Created new job wrapper script for Python3 environment

### LINKED ISSUE #10641: Adapt the job wrapper script (submit.sh) to load a python3 environment
**Impact of the new feature**
WMAgent

As we move WMAgent to python3 stack, we need to assure that the agent and the runtime code are both executed with the same environment, thus python3 environment needs to be loaded within the worker nodes (from CVMFS).

The final solution is not yet clear to me, and it could be that we have a very short solution; and then - in a later stage - work on something more robust and longer term.

This code: https://github.com/dmwm/WMCore/blob/master/etc/submit.sh#L134-L156 needs to be adapted such that we source the COMP python3 and py3-future packages, when running jobs from a Python3 WMAgent. We could either hard-wire the python installation area; or we could keep the current logic in place, which looks for the latest python version installation.

WARNING: we cannot install python3 in CVMFS yet because - I think - it will immediately cause a massive failure in production. I'm making a new GH issue for that and will update this one soon.

To keep dual stack compatibility, we might want to look into one of the following options:
* for Py3 agent, we use a different script, e.g.: `submit_py3.sh`
* we pass a new argument to the submit.sh script, flagging whether it's suppose to run in a Python2 or Python3 environment.

**Additional context**
None

amaltaro:
Fixes #10641

#### Status
ready

#### Description
This PR adds a new job wrapper script to be used for Py3 WMAgents, such that we load the correct python3 and py3-future libraries during runtime in the worker node.

If the job is requesting for slc6_* ScramArch, we do not have our own COMP packages deployed in CVMFS, but a symlink that Shahzad created to some specific CMSSW version. More info at: https://github.com/dmwm/WMCore/issues/10642

Note that the job wrapper is exactly the same as submit.sh, only difference is regarding the python and future libraries that we source.

NO (meant to only be used in Py3 WMAgent (and T0 agent, in the future)

#### Related PRs
None

None

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -8,7 +8,14 @@ WMAgent Configuration
 Sample WMAgent configuration.
 """
 import os
+from Utils.PythonVersion import PY3
 from WMCore.Configuration import Configuration
+
+# job wrapper script
+if PY3:
+    submitScript = "etc/submit_py3.sh"
+else:
+    submitScript = "etc/submit.sh"
 
 # The following parameters may need to be changed.
 serverHostName = "HOSTNAME OF WMAGENT MACHINE"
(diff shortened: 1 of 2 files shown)


amaltaro:
This has been tested with a Py3 WMAgent (1.5.2.pre1) and things seem to be working properly. Any comments or concerns are welcome at any time though.

PR #10731, opened 2021-08-06
PY3: Create job package with pickle protocol=4 for Python3 agents

### LINKED ISSUE #10730: [py2py3] Use pickle protocol=4 for job package in Python3 agents
**Impact of the bug**
WMAgent

**Describe the bug**
As we commission WMAgent in python3, it looks like we are not going to have a consistent python3 version between `rhel6` and `rhel7` OS, for:
* `rhel6`, thus slc6_*: we are planning to actually use the CMSSW python3.6.4
* `rhel7`, thus slc7_*: we are going to use our own COMP python3.8.2

Python 3.6.4 does not support pickle protocol=5, so we cannot simply create pickle files with the standard protocol in python 3.8.2. Otherwise it fails to be loaded during job runtime with:
```
Failed to load JobPackage:/srv/job/WMSandbox/JobPackage.pcl
unsupported pickle protocol: 5
```

**How to reproduce it**
Run a test workflow with Py3 WMAgent >= 1.5.1.pre6 (plus a few extra patches)

**Expected behavior**
My suggestion is to evaluate the python version when creating the job package in JobCreator
* if it's a python2, then simply create it with the highest protocol available
* if it's a python3 env, then create the pickle files with protocol 4

This PR contains a lot of information: https://github.com/dmwm/WMCore/pull/10619#issuecomment-894079860
And here is the job wrapper changes: https://github.com/dmwm/WMCore/pull/10726

amaltaro:
Fixes #10730 

#### Status
ready

#### Description
Be explicit with the pickle protocol used in JobCreator, which creates the job package pickle files. Such that:
* in a Py2 WMAgent, set it to protocol 2
* in a Py3 WMAgent, set it to protocol 4 (because runtime code might use python3.6 instead of python3.8)

Once we manage to move all the runtime code to python3.8+, we can remove this tweak and just use whatever is the HIGHEST_PROTOCOL.

YES

#### Related PRs
Somehow related to this job wrapper implementation: https://github.com/dmwm/WMCore/pull/10726

None

### DIFF
--- a/src/python/Utils/PythonVersion.py
+++ b/src/python/Utils/PythonVersion.py
@@ -5,6 +5,13 @@ Easily get the version of the python interpreter at runtime
 from __future__ import division # Jenkins CI
 
 import sys
+import pickle
 
 PY3 = sys.version_info[0] == 3
 PY2 = sys.version_info[0] == 2
+
+# We need to keep compatibility between multiple python versions
+# Further details at: https://github.com/dmwm/WMCore/pull/10726
+# For PY2: set highest protocol to 2
+# For PY3: set highest protocol to 4 (compatible with python3.6 and python3.8)
+HIGHEST_PICKLE_PROTOCOL = pickle.HIGHEST_PROTOCOL if PY2 else 4
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -12,12 +12,14 @@ import os
 import os.path
 import threading
 
+
 try:
     import cPickle as pickle
 except ImportError:
     import pickle
 
 from Utils.Timers import timeFunction
+from Utils.PythonVersion import HIGHEST_PICKLE_PROTOCOL
 from Utils.MathUtils import quantize
 from WMComponent.JobCreator.CreateWorkArea import CreateWorkArea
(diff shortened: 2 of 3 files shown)


mapellidario (approved):
Looks good, thanks Alan!

amaltaro:
Thanks for the prompt review, Dario. Unfortunately, it was an ill fix and I wasn't fixing the real issue. Second commit should resolve that problem, but not to waste your time, another review request will only come once I see it fixed in my tests ;)

mapellidario:
Don't worry! When we decided not to use a fixed the pickle protocol version in all WMCore in all the python runtimes, we knew that this try and error approach would have been the only sane way to set the protocol versions correctly (or at least to set them in a way that doesn't break everything). I completely trust your tests and your instincts!

amaltaro:
Tested and it now works well. Of course, there are other errors during runtime, but those deserve a different GH issue :-)
Let me squash these 2 commits. Dario, in case you want to look only at the 2nd commit, here it is: https://github.com/dmwm/WMCore/pull/10731/commits/d04e5c0380bc368cea18a89eeaf4a8b45821ac4d

mapellidario (approved):
Looks even better than before, Thanks Alan!

amaltaro:
Thanks, Dario. Merging it to make a new, clean and fresh release.

PR #10742, opened 2021-08-09
[py2py3] Modernize all code in WMCore around new pickle API

### LINKED ISSUE #10732: [py2py3] pickle load/dump should always read bytes
**Impact of the bug**
WMCore in general

**Describe the bug**
`pickle.dump(object, file)` and `pickle.load(file)` require `file` to be  of type bytes and we do not enforce it everywhere in our code yet. The list of potential problematic usage of pickle is at [1]

**How to reproduce it**
We have not any specific bugs, i just looked for `pickle` in our code and tried to spot problematic uses.

**Expected behavior**
We should always pass bytes to pickle.

[1]

https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/src/python/PSetTweaks/PSetTweak.py#L370-L371

https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/src/python/PSetTweaks/PSetTweak.py#L399-L400

https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/src/python/WMCore/BossAir/Plugins/MockPlugin.py#L40-L42

https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/src/python/WMCore/BossAir/Plugins/MockPlugin.py#L167-L173

Here we should also fix how we interface with `base64`, since `base64.encodestring()` `base64.decodestring()` are not available in py3
https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/src/python/WMCore/ThreadPool/ThreadPool.py#L147-L149

https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/src/python/WMCore/ThreadPool/ThreadSlave.py#L176

https://github.com/dmwm/WMCore/blob/2b41a0cc1fc635c71ce790b6d2433d3b31da7304/test/python/WMCore_t/BossAir_t/BossAir_t.py#L318-L319

https://github.com/dmwm/WMCore/blob/74bd93ca5ebd50c7b84e72223074ef7945dfce21/test/python/WMCore_t/Misc_t/WMAgent_t.py#L448-L450

mapellidario:
Fixes #10732

#### Status
Ready

#### Description

The changes fall into two categories:

- make sure that when loading a pickle, we read from a bytes stream in both py2 and py3. make sure that when dumping a pickle, we write to a bytes stream in both py2 and py3. We are lucky, we are only reading/writing to files (no bytesio streams, for example), so this was as easy as using `"wb"`/`"rb"` where neeeded.

- make sure that who used the result of `pickle.dumps` is expecting bytes. make sure that the input of `pickle.loads` is bytes. This was down to checking the input/output types of `pickle` and `base64` loader/dumper encoder/decoder.

yes, it should continue behaving as expected in py2

#### Related PRs
We already make changes related to pickle, but this is the first extensive search into what we missed

nope

### DIFF
--- a/src/python/PSetTweaks/PSetTweak.py
+++ b/src/python/PSetTweaks/PSetTweak.py
@@ -367,7 +367,7 @@ class PSetTweak(object):
             with open(filename, "w") as handle:
                 handle.write(self.jsonise())
         if formatting == "pickle":
-            with open(filename, "w") as handle:
+            with open(filename, "wb") as handle:
                 pickle.dump(self, handle)
         if formatting == "simplejson":
             with open(filename, "w") as handle:
@@ -396,7 +396,7 @@ class PSetTweak(object):
             raise RuntimeError(msg)
 
         if formatting == "pickle":
-            with open(filename, 'r') as handle:
+            with open(filename, 'rb') as handle:
                 unpickle = pickle.load(handle)
             self.process.__dict__.update(unpickle.__dict__)
 
--- a/src/python/WMCore/BossAir/Plugins/MockPlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/MockPlugin.py
@@ -37,7 +37,7 @@ def processWorker(myinput, tmp):
             if jj['cache_dir'].count("Production/LogCollect") > 0:
                 if lcreport is not None:
                     lcreport.task = "/" + taskName + "/Production/LogCollect"
-                    with open(outfile, 'w') as f:
+                    with open(outfile, 'wb') as f:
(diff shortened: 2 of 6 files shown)


amaltaro (approved):
And we indeed got 1 unit test fixed, thanks Dario!

PR #10753, opened 2021-08-16
Investigate effect of RESTFormatter.to_string return type

### LINKED ISSUE #10222: [py2py3] investigate WMCore/WebTools/RESTFormatter.to_string() return type in py3
**Impact of the new feature**

It should affect only the internals cherrypy when running WMCore from a py3 environment, [this function](https://github.com/cherrypy/cherrypy/blob/c524a0da73a85ec4dd4ae6b1512c25854b19c8de/cherrypy/_cprequest.py#L775) in particular.

In #10117  we modernized ` src/python/WMCore/WebTools/RESTFormatter.py ` as follows

```python
+ from builtins import str, bytes
    def to_string(self, data):
        if isinstance(data, GeneratorType):
            return self.json(data)
        if isinstance(data, dict) or isinstance(data, list):
            return json.dumps(data)
-        return str(data)
+        return bytes(str(data), "utf-8") #  equivalent to `str(data).encode("utf-8")`
```

as discussed in [this comment in PR 10117](https://github.com/dmwm/WMCore/pull/10117#discussion_r555779338). This is based on the assumption that the internals of cherrypy use bytes string in py3 also, which is only guessed by looking at [cherrypy py3 code](https://github.com/cherrypy/cherrypy/blob/c524a0da73a85ec4dd4ae6b1512c25854b19c8de/cherrypy/_cprequest.py#L775), this has not been tested yet.

We should test this assumption in dev, testbed and/or pre-production.

If cherrypy requires `RESTFormatter.to_string() ` to return a bytes string in py3 also, then we are good to go. If it requires unicode string in py3, then we need to change our code.

mapellidario:
Fixes #10222

#### Status
I think we are ready!

#### Description
Trying not to force `RESTFormatter.to_string` to return bytes.

We have not seen any problem related to this, but some investigation is always desirable.

- [x] how does other unittests react to this cange?
  - no new failures in both py2 and py3
  - This is strange: i am not able to reproduce the error described in #10117 that sparked the opening of #10222
- [ ] why is `RESTFormat_t.RESTFormatTest.testSupportedFormat` not failing? (this is the only unittest that uses `RESTFormatter.to_string` in `RESTFormat_t.py`)

Stacktrace of when `RESTFormatter.RESTFormatter.to_string` is called [1]

Options

1. "if aint broke, dont fix it"
2. since the method is called `to_string`, since we are moving to py3, we can just return py3 `str`. Since the unittests do not fail, this should be a viable option

**Solution**: Since we have not been able to replicate the error, we drop the conversion in the last line, we force every type that is not `GeneratorType`, `list`, `dict`, `bytes` to be stringyfied with `str(data)` (unicode in both py2 and py3), we  properly decode bytes to unicode and remove the comment 

```plaintext
        # Do not be misguided by the function name, this must return bytes, 
        # otherwise the unit test
        # testSupportedFormat (WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest)
        # fails with "ValueError: Page handlers MUST return bytes. 
        # Use tools.encode if you wish to return unicode."
```

yes

#### Related PRs
See this comment: https://github.com/dmwm/WMCore/pull/10117#discussion_r555779338

nope

#### more info

[1]

How is `RESTFormatter.RESTFormatter.to_string` called, obtained with `pdb` when running `RESTFormat_t.RESTFormatTest.testSupportedFormat`:
1. RESTApi.RESTApi.default https://github.com/dmwm/WMCore/blob/8c73e7073fd8928326884e3d32fd9c806dcb7429/src/python/WMCore/WebTools/RESTApi.py#L79
2. RESTApi.RESTApi._buildResponse https://github.com/dmwm/WMCore/blob/8c73e7073fd8928326884e3d32fd9c806dcb7429/src/python/WMCore/WebTools/RESTApi.py#L104
3. RESTApi.RESTApi._formatResponse https://github.com/dmwm/WMCore/blob/8c73e7073fd8928326884e3d32fd9c806dcb7429/src/python/WMCore/WebTools/RESTApi.py#L149
4. RESTFormatter.RESTFormatter.format  https://github.com/dmwm/WMCore/blob/8c73e7073fd8928326884e3d32fd9c806dcb7429/src/python/WMCore/WebTools/RESTFormatter.py#L91
5. RESTFormatter.RESTFormatter.to_string

### DIFF
--- a/src/python/WMCore/WebTools/RESTFormatter.py
+++ b/src/python/WMCore/WebTools/RESTFormatter.py
@@ -8,7 +8,7 @@ appropriate format and sets the CherryPy header appropriately.
 Could add YAML via http://pyyaml.org/
 """
 
-from builtins import str, bytes
+from builtins import str
 
 import json
 from types import GeneratorType
@@ -17,6 +17,9 @@ from cherrypy import response, HTTPError, request
 
 from WMCore.WebTools.Page import TemplatedPage, _setCherryPyHeaders
 from WMCore.Wrappers.JsonWrapper.JSONThunker import JSONThunker
+
+from Utils.Utilities import decodeBytesToUnicodeConditional
+from Utils.PythonVersion import PY3
 
 class RESTFormatter(TemplatedPage):
     def __init__(self, config):
@@ -66,16 +69,21 @@ class RESTFormatter(TemplatedPage):
                                 path = request.path_info)
 
     def to_string(self, data):
+        """
+        Since json.dumps returns unicode in py3 and bytes in py2 (it behaves
+        returning "native str" in both versions), then we do the same with all
+        the data that is not GeneratorType, dict nor list by calling str(data)
+        Moreover, we need to properly decode bytes.
+
+        :returns: "native str" (unicode in py3, bytes in py2)
+        """
         if isinstance(data, GeneratorType):
             return self.json(data)
         if isinstance(data, dict) or isinstance(data, list):
             return json.dumps(data)
-        # Do not be misguided by the function name, this must return bytes, 
-        # otherwise the unit test
-        # testSupportedFormat (WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest)
-        # fails with "ValueError: Page handlers MUST return bytes. 
-        # Use tools.encode if you wish to return unicode."
-        return bytes(str(data), "utf-8")
+        if not isinstance(data, bytes):
+            return str(data)
+        return decodeBytesToUnicodeConditional(data, condition=PY3)
 
     def format(self, data, datatype, expires):
         response_data = ''


src/python/WMCore/WebTools/RESTFormatter.py
```diff
@@ -75,7 +75,8 @@ def to_string(self, data):
         # testSupportedFormat (WMCore_t.WebTools_t.RESTFormat_t.RESTFormatTest)
         # fails with "ValueError: Page handlers MUST return bytes. 
         # Use tools.encode if you wish to return unicode."
-        return bytes(str(data), "utf-8")
+        # return bytes(str(data), "utf-8")
+        return str(data)
```
amaltaro, line 79: Dario, how about a safe conversion to str like `decodeBytesToUnicode`. Or maybe we execute that only in PY3, and in py2 we do this `str(data)`?

mapellidario: Yeah yeah i know. we can do all these things. My purpose here was just to check if and why that explicit (and with retrospect rather ugly) conversion to bytes was necessary for dmwm/WMCore. Since no unittest failed with this change, i would simply remove the conversion and live without. As you mentioned in a private chat, e should also wait for CRAB confirmation

mapellidario: Ah sorry! you proposed a `how about a safe conversion to str` actually, replacing `str()` altogether. I do not think it is really feasible, what if `data` is not a string? I think the purpose of this method was stringyfying lists,dictionaries, sets, etc to a str, so i would keep this `str()`

amaltaro: Dario, to avoid bad surprises, how about the following then:
```
if not isinstance(data, bytes):
    return str(data)
return decodeBytesToUnicode(data)
```

this would make the method consistent, as provided from your changes, and also avoid possible issues if we try to convert a bytes to string. If you agree, then this would likely close that other issue as well :-D

mapellidario: ok! works for me! fix coming!

amaltaro:
Thank you for the prompt fix, Dario.

PR #10757, opened 2021-08-17
Fix Unittests in Api_t.Tester

### LINKED ISSUE #10719: Broken unittests in Api_t.Tester.* in py3
**Impact of the bug**

Some unittests are failing in `Api_t.Tester`. It is not clear to me yet if this has a broader impact.

**Describe the bug**

I checked a couple of the failures in the jenkins report and they all look the same: the line in `cheroot/test/webtest.py` 

```python
old_settings = termios.tcgetattr(fd)
```

fails with: `(25, 'Inappropriate ioctl for device')`

However, when running in my docker the unittest `test/python/WMCore_t/REST_t/Api_t.py:Tester.test_accept_good_gif` fails at the line https://github.com/dmwm/WMCore/blob/65b2e5cb0323b77103382fe07bc52cc0bc49a72b/src/python/WMCore/REST/Format.py#L621, which is a different failures wrt Jenkins.

**How to reproduce it**
Just have a look at the failing unittests in the [jenkins report](https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCorePy3-UnitTests/lastBuild/testReport/junit/Api_t/Tester/).

**Expected behavior**
the unittests should pass, as in py2.

This issue is related to #10422

mapellidario:
Fixes #10719 

#### Status
problems solved. we should discuss if my current solutions are the best solutions possible

##### status of this PR

I replicated the error reported by jenkins using the pdb debugger provided by vscode, when i use the docker environment directly from its shell i get a different failure. This was sort to be expected, since the error in jenkins emerges from `termios`. lets see what i can do.

- [x] REST/Format -> test_good_gif, test_good_gif2
  - this is caused by https://github.com/dmwm/WMCore/blob/88b70a21208c1b4b849a122336c600210a71c487/src/python/WMCore/REST/Format.py#L621
  - we expect the content of the request to be `'GIF89a\x01\x00\x01\x00\xc2\x82\x00\x01\xc2\x99"\x1e\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00,\x00\x00\x00\x00\x01\x00\x01\x00\x02\x03\x02\x08\t\x00;'`, whose `len()` is 53 and matches `Content-Length` http header says, but the content we retrieve contains twice the byte `\xc2`, making the length of the content that we retrieve longer, which fails the assertion.
  - The problem was that in `test/python/WMCore_t/REST_t/Api_t.py` we define `gif_bytes = ('some' 'thing')` (which becomes `'something'`). We are effectively chaining py3 `str`. but this should be bytes! defining instead `gif_bytes = (b'some' b'thing')` fixes the unittest. the two spurious bytes were added when converting (inadvertently, i am not even sure where this happened) gif_bytes from py3 str to py3 bytes.
- [x] REST/Validation -> test_multi_throw10, test_multi_throw5a
  1. [x] problem 1
    - this is caused by https://github.com/dmwm/WMCore/blob/88b70a21208c1b4b849a122336c600210a71c487/src/python/WMCore/REST/Validation.py#L41
    - `TypeError: cannot use a string pattern on a bytes-like object`
    - `_validate_one` is calling `_validate_str` (which expects bytes input), which calls `_check_str` where we match a bytes string with a str regex. This requires proper investigation
  2. [x] problem 2
    - Log: [1]
    - the solution is not so smooth as i would like.
    - Breakpoints for `pdb`: REST/Format.py L246, L542, L579, REST/Server.py L726, L842
    - Stacktrace: Server 726, Server 739->842, Format 549-> 579, (some logging stuff of cherrypy), Format 535->542  (yields first element for iteration at Format 239->246), if `chunk` at 543 is not bytes, then cherrypy crashes
    - eventually caused by https://github.com/dmwm/WMCore/blob/88b70a21208c1b4b849a122336c600210a71c487/src/python/WMCore/REST/Format.py#L543 retuurning a py3 `str` instead of `bytes` 
- [x] test_multi_nothrow and others unittests

##### How to investigate the failures

this is not trivial, since if we have a bug in the cherrypy server, if we assert the status of a request with `cheroot.test.webtest.WebCase.assertStatus` then it [calls termios.tcgetattr()](https://github.com/cherrypy/cheroot/blob/ba1bc4b0afc287de80b2b7449756e56f8a5e44dd/cheroot/test/webtest.py#L77). What happens in

1. jenkins: we get `termios.error:  (25, 'Inappropriate ioctl for device')` (because the shell that jenkins gives to cherrypy is not interactive and cherrypy can not get any key from it)
2. shell open in dev docker environment: cherrypy reports the stacktrace of the failure in its logs, but then hangs and it needs to be killed with something like `ps auc | grep python3 | awk '{print $2}' | xargs kill -9` from another shell opened in the same docker container.
3. using pdb through vscode inside the dev docker environment: same as jenkins.

In order to run the unittests ffrom you docker shell with for example `python3 setup.py test --buildBotMode=true --reallyDeleteMyDatabaseAfterEveryTest=true --testCertainPath=test/python/WMCore_t/REST_t/Api_t.py`, then you need to add `test/python/WMCore_t/REST_t/` to the pythonpath, because REST/Main.py imports Api_t.py. You can either change the shell environment v ariable `$PATH`, or you can append that directory into `sys.path`, as in [2].

moreover, if you need/want to use the vscode debugger, you need to make the unittests discoverable by vscode (by adding a properly placed `__init__` file), as in [2]. Beware! this change makes jenkins crash so badly that it does not even send the report in the comments and starts again the unittests. Use this change only on your local environment, do not push this!

yes

#### Related PRs
nope

nope

#### other info

[1] 

```plaintext
[18/Aug/2021:16:22:37] ENGINE ValueError('WSGI Applications must yield bytes')
  File "/home/dmwm/unittestdeploy/wmagent/1.4.9.pre6-comp2/sw/slc7_amd64_gcc630/external/py3-cheroot/8.2.1/lib/python3.8/site-packages/cheroot/server.py", line 1280, in communicate
    req.respond()
  File "/home/dmwm/unittestdeploy/wmagent/1.4.9.pre6-comp2/sw/slc7_amd64_gcc630/external/py3-cheroot/8.2.1/lib/python3.8/site-packages/cheroot/server.py", line 1083, in respond
    self.server.gateway(self).respond()
  File "/home/dmwm/unittestdeploy/wmagent/1.4.9.pre6-comp2/sw/slc7_amd64_gcc630/external/py3-cheroot/8.2.1/lib/python3.8/site-packages/cheroot/wsgi.py", line 147, in respond
    raise ValueError('WSGI Applications must yield bytes')
ValueError: WSGI Applications must yield bytes
```

[2]

```diff
From 5894ae84c8a3211a9fd53623e3077b3d01c92edf Mon Sep 17 00:00:00 2001
From: Dario Mapelli <mapelli.dario@gmail.com>
Date: Tue, 17 Aug 2021 14:50:00 +0200
Subject: [PATCH] replicate in docker the unittest error reported by jenkins in
  WMCore_t/REST_t/Api_t.py

---
 test/python/WMCore_t/REST_t/Api_t.py    | 2 ++
 test/python/WMCore_t/REST_t/__init__.py | 0
 2 files changed, 2 insertions(+)
 create mode 100644 test/python/WMCore_t/REST_t/__init__.py

diff --git a/test/python/WMCore_t/REST_t/Api_t.py b/test/python/WMCore_t/REST_t/Api_t.py
index 1b88b29f80..85cb1e4674 100644
--- a/test/python/WMCore_t/REST_t/Api_t.py
+++ b/test/python/WMCore_t/REST_t/Api_t.py
@@ -3,6 +3,7 @@
 import re
 import zlib
 from multiprocessing import Process
+import sys
 
 import cherrypy
 from cherrypy import response
@@ -245,6 +246,7 @@ def test_multi_throw10(self):
             assert b["result"][i][1] == i
 
 def setup_server():
+    sys.path.append("/".join(__file__.split("/")[:-1]))
     srcfile = __file__.split("/")[-1].split(".py")[0]
     setup_dummy_server(srcfile, "Root", authz_key_file=FAKE_FILE, port=PORT)
 
diff --git a/test/python/WMCore_t/REST_t/__init__.py b/test/python/WMCore_t/REST_t/__init__.py
new file mode 100644
index 0000000000..e69de29bb2
```

### DIFF
--- a/src/python/WMCore/REST/Format.py
+++ b/src/python/WMCore/REST/Format.py
@@ -537,10 +537,10 @@ def _etag_tail(head, tail, etag):
     Sets ETag header at the end to value of `etag` if it's defined and
     yields a value."""
     for chunk in head:
-        yield chunk
+        yield encodeUnicodeToBytes(chunk)
 
     for chunk in tail:
-        yield chunk
+        yield encodeUnicodeToBytes(chunk)
 
     etagval = (etag and etag.value())
     if etagval:
--- a/src/python/WMCore/REST/Validation.py
+++ b/src/python/WMCore/REST/Validation.py
@@ -1,10 +1,12 @@
-from builtins import str, bytes
-from past.builtins import basestring
+from builtins import str as newstr, bytes as newbytes
 
 from WMCore.REST.Error import *
 import math
 import re
 import numbers
+
+from Utils.Utilities import decodeBytesToUnicodeConditional, encodeUnicodeToBytesConditional
+from Utils.PythonVersion import PY3, PY2
 
 def return_message(main_err, custom_err):
     if custom_err:
@@ -21,7 +23,7 @@ def _arglist(argname, kwargs):
         return val
 
 def _check_rx(argname, val, custom_err = None):
-    if not isinstance(val, basestring):
+    if not isinstance(val, (newstr, newbytes)):
         raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
     try:
         return re.compile(val)
@@ -30,30 +32,24 @@ def _check_rx(argname, val, custom_err = None):
 
 def _check_str(argname, val, rx, custom_err = None):
     """
-    convert unicode to str (bytes) first to support cherrypy 3.2.2
     This is not really check val is ASCII.
-    """
-    if isinstance(val, str):
-        try:
-            val = bytes(val, "utf-8")
-        except:
-            raise InvalidParameter(return_message("Invalid '%s' parameter %s %s" % (argname, val, type(val)), custom_err))
-    if not isinstance(val, basestring) or not rx.match(val):
-        raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
-    return val
+    2021 09: we are now using version 17.4.0 -> we do not need to convert to 
+    bytes here anymore, we are using a recent verison of cherrypy.
+    We merged the funcionality of _check_str and _check_ustr into a single function
 
-def _check_ustr(argname, val, rx, custom_err = None):
-    if isinstance(val, bytes):
-        try:
-            val = str(val, "utf-8")
-        except:
-            raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
-    if not isinstance(val, basestring) or not rx.match(val):
-        raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
+    :type val: str or bytes (only utf8 encoded string) in py3, unicode or str in py2
+    :type rx: regex, compiled from native str (unicode in py3, bytes in py2)
+    """
+    val = decodeBytesToUnicodeConditional(val, condition=PY3)
+    val = encodeUnicodeToBytesConditional(val, condition=PY2)
+    # `val` should now be a "native str" (unicode in py3, bytes in py2)
+    # here str has not been redefined. it is default `str` in both py2 and py3.
+    if not isinstance(val, str) or not rx.match(val):
+        raise InvalidParameter(return_message("Incorrect '%s' parameter %s %s" % (argname, type(val), val), custom_err))
     return val
 
 def _check_num(argname, val, bare, minval, maxval, custom_err = None):
-    if not isinstance(val, numbers.Integral) and (not isinstance(val, basestring) or (bare and not val.isdigit())):
+    if not isinstance(val, numbers.Integral) and (not isinstance(val, (newstr, newbytes)) or (bare and not val.isdigit())):
         raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
     try:
         n = int(val)
@@ -66,7 +62,7 @@ def _check_num(argname, val, bare, minval, maxval, custom_err = None):
         raise InvalidParameter(return_message("Invalid '%s' parameter" % argname, custom_err))
 
 def _check_real(argname, val, special, minval, maxval, custom_err = None):
-    if not isinstance(val, numbers.Number) and not isinstance(val, basestring):
+    if not isinstance(val, numbers.Number) and not isinstance(val, (newstr, newbytes)):
         raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
     try:
         n = float(val)
@@ -114,9 +110,10 @@ def validate_str(argname, param, safe, rx, optional = False, custom_err = None):
     successful the string is copied into `safe.kwargs` and the value
     is removed from `param.kwargs`.
 
-    Use `validate_ustr` instead if the argument string might need to
-    be converted from utf-8 into unicode first. Use this method only
-    for inputs which are meant to be bare strings.
+    Accepts both unicode strings and utf8-encoded bytes strings as argument 
+    string.
+    Accepts regex compiled only with "native strings", which means  str in both 
+    py2 and py3 (unicode in py3, bytes of utf8-encoded strings in py2)
 
     If `optional` is True, the argument is not required to exist in
     `param.kwargs`; None is then inserted into `safe.kwargs`. Otherwise
@@ -125,22 +122,20 @@ def validate_str(argname, param, safe, rx, optional = False, custom_err = None):
 
 def validate_ustr(argname, param, safe, rx, optional = False, custom_err = None):
     """Validates that an argument is a string and matches a regexp,
-    once converted from utf-8 into unicode.
+    During the py2->py3 modernization, _check_str and _check_ustr have been 
+    merged into a single function called _check_str.
+    This function is now the same as validate_str, but is kept nonetheless 
+    not to break our client's code.
 
     Checks that an argument named `argname` exists in `param.kwargs`
     and it is a string which matches regular expression `rx`. If
     successful the string is copied into `safe.kwargs` and the value
     is removed from `param.kwargs`.
 
-    Use `validate_str` instead if the argument string should always be
-    a bare string. This one automatically converts everything into
-    unicode and expects input exclusively in utf-8, which may not be
-    appropriate constraints for some uses.
-
     If `optional` is True, the argument is not required to exist in
     `param.kwargs`; None is then inserted into `safe.kwargs`. Otherwise
     a missing value raises an exception."""
-    _validate_one(argname, param, safe, _check_ustr, optional, rx, custom_err)
+    _validate_one(argname, param, safe, _check_str, optional, rx, custom_err)
 
 def validate_num(argname, param, safe, optional = False,
                  bare = False, minval = None, maxval = None, custom_err = None):
--- a/src/python/WMCore/REST/Error.py
+++ b/src/python/WMCore/REST/Error.py
@@ -1,4 +1,4 @@
-from builtins import str
+from builtins import str as newstr
 import random, cherrypy
 
 class RESTError(Exception):
@@ -258,11 +258,11 @@ def report_rest_error(err, trace, throw):
                      % (err.http_code, err.app_code, err.message,
                         getattr(err.errobj, "__module__", "__builtins__"),
                         err.errobj.__class__.__name__,
-                        err.errid, err.instance, str(err.errobj).rstrip(),
+                        err.errid, err.instance, newstr(err.errobj).rstrip(),
                         sql, binds, kwbinds, offset))
         for line in err.trace.rstrip().split("\n"): cherrypy.log("  " + line)
-        cherrypy.response.headers["X-REST-Status"] = str(err.app_code)
-        cherrypy.response.headers["X-Error-HTTP"] = str(err.http_code)
+        cherrypy.response.headers["X-REST-Status"] = newstr(err.app_code)
+        cherrypy.response.headers["X-Error-HTTP"] = newstr(err.http_code)
         cherrypy.response.headers["X-Error-ID"] = err.errid
         report_error_header("X-Error-Detail", err.message)
         report_error_header("X-Error-Info", err.info)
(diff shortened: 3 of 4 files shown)


mapellidario:
test this please

mapellidario:
test this please

src/python/WMCore/REST/Validation.py
```diff
@@ -35,11 +37,17 @@ def _check_str(argname, val, rx, custom_err = None):
...
             raise InvalidParameter(return_message("Invalid '%s' parameter %s %s" % (argname, val, type(val)), custom_err))
-    if not isinstance(val, basestring) or not rx.match(val):
+    if not isinstance(val, basestring):
         raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
+    try:
+        if not rx.match(val):
+            raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
+    except TypeError:  # if we match bytes against a string pattern
+        if not rx.match(decodeBytesToUnicode(val)):
+            raise InvalidParameter(return_message("Incorrect '%s' parameter" % argname, custom_err))
```
mapellidario, line 50: The problem that we try to solve here is: sometimes `val` is bytes, but `rx` is a regex pattern compiled from unicode. Since i could not find a way of programmatically knowing the type of the string used to compile the pattern, what i lazily propose here is:

assumptions
- if `val` is unicode, we convert it to bytes
- if `val` is not bytes, we raise InvalidParameter

options, knowing that `val` is bytes:
1. if rx is a pattern from bytes, then `rx.match(val)` returns a boolean correctly. if this is not True, then we raise an InvalidParameter exception.
2. if rx is a pattern from unicode, `rx.match(val)` raises a TypeError exception, which is catched below by `except TypeError`. now `rx.match(decodeBytesToUnicode(val))` returns a valid boolean. if this is not True, then we raise and InvalidParameter exception.

It is possible that there is a simpler and better way to deal with this, i am open to suggestions!

When we want to check if a bytes string matches a unicode pattern, i assume that the bytes string can be safely decoded to unicode. This is not true in general (one may have a bytes stream containing data from an image or root file), but i assume that in this case we wouldn't run `validate_str()` on such a stream in py2. Or at least, we would pass a pattern that is explicitly of bytes type.

amaltaro: Dario, sorry for suggesting another approach, but what if we enforce `val` to be a py2/py3 string object instead of bytes? You said the regex `rx` is a py3 string, which makes me think that we should actually expect a py3 string `val` as well. I could very well be missing something here.

amaltaro (commented):
I have made some comments along the code. I still do not know whether those changes are good to go or not, even though they fix a bunch of unit tests!
Before merging it, I think running central services with this patch is a must. I will do so once we agree on a final candidate (which might be this one).

src/python/WMCore/REST/Format.py
```diff
@@ -537,10 +537,10 @@ def _etag_tail(head, tail, etag):
     Sets ETag header at the end to value of `etag` if it's defined and
     yields a value."""
     for chunk in head:
-        yield chunk
+        yield encodeUnicodeToBytes(chunk)
```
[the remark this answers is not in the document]
mapellidario: Yep, you understood correctly. It fails with log at [1]. It is possible that this is related to the WSGI server wrapper that we are using, it is possible that the old version of cherrypy was doing this conversion internally silently, while the new one does less silent job. but i am not sure, we may need to investigate this claim.

src/python/WMCore/REST/Validation.py
mapellidario: your suggestion seems the most reasonable one to me! i do not know-remember why i did not go for it, let me try!

mapellidario: Here we go: `convert unicode to str (bytes) first to support cherrypy 3.2.2` is in the docstring, this is why i did not go for that route

amaltaro: I couldn't find anything clear on the CP documentation, besides some bug fixes for response header and body. 
We are now in version 17.4.0, if you think it's worth trying the string object approach, I'm all for it. In the end, whatever approach we take here, a reasonable "real" test will be required.

mapellidario: I am currently trying to see what happens if I assume that `val` and `rx` are unicode. If this works, then `_check_ustr()`/`validate_ustr()` become redundant. They do not seem to be used anywhere by WMCore, but maybe someone else could be relying on those.

mapellidario:
@amaltaro I had yet another look at `src/python/WMCore/REST/Validation.py`, in particular at `_check_str()`.

Our "native string" approach worked so well that in `_check_str()`
- py2: we assumed `val` and `rx` to be py2-`str` (bytes) (return and error otherwise)
- py3: if we assume that `val` and `rx` are py3-`str` (unicode), then the unittests are happy
- We can use the same code to do both these two jobs.

However, in py2 we had `validate_str()` to validate py2-`str` , which called `_check_str()`, and `validate_ustr()` to validate unicode, which called `_check_ustr()`. Since it seems that we can modernize `_check_str()` to deal with unicode in py3 and bytes in py2, since `validate_ustr()` doesn't seem to be used in WMCore, we can think about 

1. simply removing `validate_ustr()` in order to avoid confusion, or
2. removing `validate_ustr()`, but making sure that `validate_str` calls `_check_ustr()` in py2 if it receives unicode in py2 and calls `_check_bytes()` (which we need to write from scratch) in py3 if it receives bytes in py3. or
3. keep `validate_str()` and `validate_ustr()` as they are, but change in WMCore code which one we call based on `if PY3`
4. I am open to suggestions! :)

amaltaro:
Thanks for this analysis and for coming up with many options, Dario.

I am not 100% confident on what proposal is going to be the best, but I'm inclined to the option 1:
> 1. simply removing validate_ustr() in order to avoid confusion, or

a possible drawback from this is that we lose the ability to "mark" byte strings from unicode. 

The `_check_bytes` option is good as well, but I fear it would be confusing while we have things running in dual stack.

Sorry for not providing you with a precise option. Hopefully this comment will help you choosing what to implement :-D
Once you have changes in place, I'm happy to run tests with your patch in place.

mapellidario:
As always, I change opinion in the same moment i get back to the code, no matter of how much time I spend thinking about a proper solution beforehand.

The current approach:

- keep separate `validate_str` and `validate_ustr`, not to break the downstream/client code
- both these two functions call `_check_str`, that as arguments accepts
  - `rx`: regex compiled from "native string"
  - `val`: is a string, but can be both unicode and bytes in both py2 and py3
- `_check_str` converts `val` in "native str" if necessary, than matches it against the regex

This approach should cover all the cases that we may face while maintaining dual-stack compatibility. 

There are only two things that may "break" this code:

- regex compiled from bytes in py3, regex compiled from unicode in py2
- bytes containing something that is not utf8-encoded string in py3

I say `"break"` and not `break` because when we have malformed inputs we always raise something, in some fringe cases that i am not sure if will ever happen we may not raise `InvalidParameter` but something related to `UnicodeDecodeError` or `TypeError`. Is this a problem? I do not think so.

@amaltaro, if you agree, we are ready for some tests. I would prefer not to squash the commits before the tests, i would like to be able to revert some changes if needed. thanks! :)

amaltaro:
@mapellidario thanks Dario, I'm going to run some tests with the changes provided here.

amaltaro:
I patched central services, injected a few workflows and things seem to be running Okay. Shall we squash these commits and get this merged? @mapellidario

mapellidario:
If the test are ok, then i am more than happy to squash and merge this! thanks Alan! :)

mapellidario:
@amaltaro This is ready to be merged! thanks, sorry for making you wait!

amaltaro:
Thanks, Dario. 
Please consider these changes whenever you and @belforte work in a new CRABServer release. There could be some surprises there that we did not manage to spot with the very basic WMCore tests that I performed.

belforte:
no worries, I am very well aware that http call args validation is where issues in CRABServer REST start, and oracle bindings is where they hopefully ends.
The question for @amaltaro  is rather...
how long, and why does this code need to be 'dual stack' ? I am happy to stick with old branch until we are ready to run the REST in  python3 and get rid of all compatibility trick.
Do you really want to run REST servers with python2 ? Or can we start looking at a pure-py3 code here ? Sorry @mapellidario for all work which may be thrown away at that point, but... the less code, the better.

amaltaro:
@belforte problem lies with the external dependencies! WMCore must remain dual-stack as long as we have other services - that depend on WMCore - running in Python2. Some examples that come to my mind are:
* CRABServer
* TaskWorker (?)
* T0WMADataSvc
* DBSServer

As soon as nobody requires python2 WMCore code, we can then start dropping python2 (and the compatibility layer).

belforte:
I fear a chicken and egg problem. CRABServer and TaskWorker may be easier to make them run in py3 if we use a WMCore branch which is fully py3. Maybe DBS and T0 could use same approach to stick to a frozen WMCore release until they have a version which runs in py3 with newer WMCore ?

belforte:
the question is: what is the chance that large, critical changes may need to be backported to the old py2 WMCore tag before the dependent tools are ready to use the new py3-only tags ?
For CRAB it was "zero" (with the day-after wisdom of course, initially one always fears).

belforte:
Do DBS and T0 have a roadmap and timeline for py3 migration ?

amaltaro:
I'm all for dropping python2 support in WMCore. The only problem is that it's not only deleting python2 lines of code, again it will take a reasonable amount of effort and we have to properly plan for that.
Maybe we should schedule a discussion for the first week of October and try to come up with concrete plan?

belforte:
+1
at least we need to identify stakeholders. DBS ? T0 ? Others ? What exactly are their plans and timelines ? Seems that WMCore and CRAB can sort out things pretty well, thanks also to Dario being such a convenient bridge :-) but I have no idea what is happening in other areas.

PR #10773, opened 2021-08-25
MSOutput: only produce documents for workflows not yet in the database

### LINKED ISSUE #10722: [MS] Stop producing the same workflow record multiple times in MSOutput
**Impact of the bug**
MSOutput

**Describe the bug**
With the deployment of this alerting new feature: https://github.com/dmwm/WMCore/pull/10699, the MSOutputProducer thread became very annoying with these duplicate notifications being created many times for the same workflow.

We need to come up with a solution to stop re-producing the same workflow record in the producer thread.
 
**How to reproduce it**
Having a workflow with >= 1 output dataset not registered in Rucio.

**Expected behavior**
Given that the load of MSOutput is pretty small, I would suggest to first search for a workflow id in the MongoDB database. If it's found, then we should skip that workflow from the Producer thread.

We should also evaluate possible performance improvements when implementing it, for instance:
* is it better to make a mongo query for each workflow?
* or we should actually make search queries in bulk? If so, is there a reasonable slice size (10? 100? 1000?)

Some extra background information can be found at: https://github.com/dmwm/WMCore/pull/10717#issuecomment-892409399

amaltaro:
Fixes #10722 

#### Status
ready

#### Description
Refactored how the **producer** thread of MSOutput process workflows eligible for output data placement. Instead of relying on an in-memory cache and in the MongoDB mechanism to check for unique document ids, changed it to actually fetch all the requests (only their names) inserted in MongoDB in the last X secs (default to last 6 months, to have a very relaxed value and give people enough time to spot workflows stuck being produced by MSOutput). We filter those out and then only produce mongo records for workflows not yet in the database.

Second commit is actually removing the memory cache mechanism that had been initially implemented. But I think we will squash those commits before merging it.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/MicroService/MSManager.py
+++ b/src/python/WMCore/MicroService/MSManager.py
@@ -28,7 +28,6 @@ from __future__ import division, print_function
 from builtins import object
 from time import sleep
 from datetime import datetime
-from collections import deque
 
 # WMCore modules
 from WMCore.MicroService.Tools.Common import getMSLogger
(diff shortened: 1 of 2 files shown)


todor-ivanov (approved):
This looks pretty good @amaltaro , Thanks !

PR #10781, opened 2021-09-01
Fix spec validation for Multicore/Memory/Streams/TpE

### LINKED ISSUE #10741: Problems with ACDCs whose memory setting was overwritten
**Impact of the bug**
The component which determines/monitors the memory threshold of requests, e.g. whoever enforces the 2000MB maxPSS in this example 
```
PerformanceKill (Exit code: 50660)
Error in CMSSW step cmsRun1
Number of Cores: 1
Job has exceeded maxPSS: 2000 MB
Job has PSS: 2192 MB
```

**Describe the bug**
For the past couple of weeks, I've realized that some ACDCs whose memory setting was overwritten are having issues. I see two issues in particular:

1. **Inconsistency between globally defined `Memory` parameter and the `Memory` defined within tasks.**

This is explained in the following issue in detail: https://github.com/CMSCompOps/WmAgentScripts/issues/891

2. **Memory setting of the ACDC workflow is not respected by the Performance Monitor.**

For instance, the following request's 1st task was submitted w/ 2GB per core, but it was not enough:
```
PerformanceKill (Exit code: 50660)
Error in CMSSW step cmsRun1
Number of Cores: 1
Job has exceeded maxPSS: 2000 MB
Job has PSS: 2223 MB
```
Then ACDCs were submitted w/ 2.3GB and 3GB of memory, but again the jobs were failed since they exceeded the previous threshold of 2GB of memory - so, the new 2.3GB and 3GB thresholds were not respected.

Requests: https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_SUS-RunIIFall17FSPremix-00218 

The original workflow: https://cmsweb.cern.ch/reqmgr2/fetch?rid=pdmvserv_task_SUS-RunIIFall17FSPremix-00218__v1_T_210703_150317_8449
Error: https://cms-unified.web.cern.ch/cms-unified/report/pdmvserv_task_SUS-RunIIFall17FSPremix-00218__v1_T_210703_150317_8449

First ACDC w/ 2300MB memory: https://cmsweb.cern.ch/reqmgr2/fetch?rid=cmsunified_ACDC0_task_SUS-RunIIFall17FSPremix-00218__v1_T_210803_121213_7312
Error: https://cms-unified.web.cern.ch/cms-unified/report/cmsunified_ACDC0_task_SUS-RunIIFall17FSPremix-00218__v1_T_210803_121213_7312

2nd ACDC w/ 3GB memory: https://cmsweb.cern.ch/reqmgr2/fetch?rid=cmsunified_ACDC1_task_SUS-RunIIFall17FSPremix-00218__v1_T_210805_143615_5865
Error: https://cms-unified.web.cern.ch/cms-unified/report/cmsunified_ACDC1_task_SUS-RunIIFall17FSPremix-00218__v1_T_210805_143615_5865  

**How to reproduce it**
Resubmitting the ACDCs mentioned above with the same configuration should reproduce the issue.

**Expected behavior**
We (P&R) should be able to set a custom memory value for a specific task and the new threshold for the new jobs of that task should be adjusted accordingly. As I explained above, there are two problems at the moment. 

1. Having 2 different memory settings (global one and nested one) is the first reason for confusion. It should be clear for us how to set the memory of a given task. With a dictionary defined globally or an absolute number defined within tasks.
2. Once we agree on this interface and propagate this information to the jobs, this should be applied.  

UPDATE (Alan): assignment validation function would fail with such dict value. 

Alan - please let me know if this long issue should be separated into two. Since the two issues that I reported here could be related, I kept them in the same issue. If this is not correct, I can separate things.

amaltaro:
Fixes #10741 

UPDATE:
* 1st commit has already been backported in this PR: https://github.com/dmwm/WMCore/pull/10783
* unfortunately it wasn't sufficient, so a 2nd commit was created to fix the remaining issues. It's been backported in this PR: https://github.com/dmwm/WMCore/pull/10787

#### Status
ready

#### Description
With the migration to python3, comparisons between dict and integer/float no longer work.

This PR refactors how `Multicore`, `Memory`, `EventStreams` and `TimePerEvent` are validated during workflow creation AND workflow assignment.

There are no changes to the format of data provided for those parameters, and their definition is made at the StdBase and it's overriden from the children classes when necessary. In short, we can provide a single value e.g.:
```
"Multicore": 4
```
or (for TaskChain/StepChain/Resubmission), we can also provide:
```
"Multicore": {"TaskName_1": 4, "TaskName_2": 16, etc etc}
```

When the former is provided in the creation of a Resubmission workflow (e.g Multicore=4), this will apply to all the tasks throughout the workflow.

NOTE: such settings provided at the Task or Step level (inner dict), take no effect! Only top level ones.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
@@ -5,12 +5,12 @@ _Resubmission_
 Resubmission module, this creates truncated workflows
 with limited input for error recovery.
 """
-
 from Utils.Utilities import makeList
 from WMCore.Lexicon import couchurl, identifier, cmsname, dataset
 from WMCore.WMSpec.StdSpecs.StdBase import StdBase
 from WMCore.WMSpec.WMWorkload import WMWorkloadHelper
-from WMCore.WMSpec.WMWorkloadTools import loadSpecClassByType, validateArgumentsCreate
+from WMCore.WMSpec.WMWorkloadTools import (loadSpecClassByType, validateArgumentsCreate,
+                                           checkMemCore, checkEventStreams, checkTimePerEvent)
 
 
 class ResubmissionWorkloadFactory(StdBase):
@@ -41,13 +41,21 @@ class ResubmissionWorkloadFactory(StdBase):
         # override a couple of parameters, if provided by user
         if 'RequestPriority' in arguments:
             helper.setPriority(arguments["RequestPriority"])
-        if arguments['OriginalRequestType'] != 'TaskChain' or isinstance(arguments['Memory'], dict):
+        if arguments['OriginalRequestType'] == 'TaskChain':
+            # Meaning: only call the setter method if values are in a dict format
+            if isinstance(arguments['Memory'], dict):
+                helper.setMemory(arguments['Memory'])
+                helper.setupPerformanceMonitoring(softTimeout=arguments.get("SoftTimeout"),
+                                                  gracePeriod=arguments.get("GracePeriod"))
+            if isinstance(arguments['Multicore'], dict):
+                helper.setCoresAndStreams(arguments['Multicore'], arguments.get("EventStreams", 0))
+            if isinstance(arguments.get('TimePerEvent'), dict):
+                helper.setTimePerEvent(arguments.get("TimePerEvent"))
+        else:
             helper.setMemory(arguments['Memory'])
             helper.setupPerformanceMonitoring(softTimeout=arguments.get("SoftTimeout"),
                                               gracePeriod=arguments.get("GracePeriod"))
-        if arguments['OriginalRequestType'] != 'TaskChain' or isinstance(arguments['Multicore'], dict):
             helper.setCoresAndStreams(arguments['Multicore'], arguments.get("EventStreams", 0))
-        if arguments['OriginalRequestType'] != 'TaskChain' or isinstance(arguments.get('TimePerEvent'), dict):
             helper.setTimePerEvent(arguments.get("TimePerEvent"))
 
         return helper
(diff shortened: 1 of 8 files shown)


haozturk:
Hi @amaltaro To be on the same page, can we make sure the new format? From the description, I understand that both integer and dictionary is supported. 
- What will happen if we provide a single integer for the resubmission of a taskchain? Are all tasks going to run with that configuration?
- AFAIU, the changes in this PR affect the validation of the parameters defined in the global level. Does this mean that the values given inside Tasks/Steps will have no effect?

Once, we agree on this, I'll do corresponding changes on Unified side.

amaltaro:
@haozturk Hi Hasan, I found an issue this morning while testing these changes in pre-production. The usual struggle of top-level and inner-level configurations... I'm investigating how to properly fix that, but any of these changes go very deep in the software and are risky to be done as a "hot-fix".

> * What will happen if we provide a single integer for the resubmission of a taskchain? Are all tasks going to run with that configuration?

As of now, a single integer would not update the workload spec settings. However, the standard behaviour would be to set that single value to all the Production/Processing tasks (no merge, logcollect, cleanup, harvesting). This is how it works for the other spec types (ReReco, StepChain, TaskChain).

> AFAIU, the changes in this PR affect the validation of the parameters defined in the global level. Does this mean that the values given inside Tasks/Steps will have no effect?

For resubmission, yes. The current implementation only considers values defined at the top level. Note that it only affects those 4 parameters provided in the initial description of this PR.

Anyhow, please do not start implementing anything on the Unified side, because there are further changes that I need to make before this can be ready to go to production.

amaltaro:
I believe this to be in a final state now. I'm cutting another reqmgr2 release for pre-production soon.
@haozturk @todor-ivanov please let me know if you have any questions/concerns.

src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
```diff
@@ -74,25 +83,44 @@ def getWorkloadCreateArgs():
...
+            TimePerEvent, Memory, Multicore, EventStreams
+        :param arguments: full set of arguments from creation+assignment definitions
+        :param userArgs: solely the key/value pair values provided by the client
+        :return: nothing, updates are made in place
+        """
+        specialArgs = ("TimePerEvent", "Memory", "Multicore", "EventStreams")
+        argsDefinition = self.getWorkloadCreateArgs()
+        for arg in specialArgs:
+            if arg not in userArgs:
+                arguments[arg] = argsDefinition[arg]["default"]
```
amaltaro, line 112: From the recent validation in testbed:
* We should not set the default value (null) for `ReReco` requests.
* Memory and TimePerEvent for StepChain and ReReco should be a single value
* We should also update the json specification accordingly (meaning, if Memory for Task1 was overwritten to 1GB, the Task1 dictionary should have the correct value). HOWEVER, this one is definitely a different issue.

amaltaro:
This has been tested in my VM and in testbed and no issues were found. There are a few aspects that should be improved, but I already talked to Hasan and we agreed that we can wait for the September upgrade to have those final tweaks in place.
I also left a comment along the code, with what needs to be addressed.

I'm merging it now because these changes actually just got deployed to CMSWEB production (using ReqMgr2 `1.5.1.cmsweb2`). I will follow the leftovers with a new GH issue/PR.

PR #10789, opened 2021-09-06
Keep Resubmission spec and json data consistent

### LINKED ISSUE #10788: Make Resubmission json and workload spec consistent
**Impact of the bug**
ReqMgr2

**Describe the bug**
This is meant to be a follow up of: https://github.com/dmwm/WMCore/pull/10781 ,
where there is an inconsistency between the workload json document (stored in reqmgr_workload_cache) and its spec file (workload object, seen under the "Config" option in ReqMgr2).

As mentioned in the PR above, there are 3 issues to be solved here:
* We should not set the default value (null) for ReReco requests.
* TimePerEvent for StepChain and ReReco should be validated as a single value
* We should also update the json specification accordingly (meaning, if Memory for Task1 was overwritten to 1GB, the Task1 dictionary should have the correct value).

**How to reproduce it**
Make ACDCs, either overriding or not some of those parameters

**Expected behavior**
The JSON request view should be consistent with what the workload object has. This is meant to be done only for Resubmission workflows. Other request types should be addressed in a different issue/PR.

None

amaltaro:
Fixes #10788 

#### Status
ready

#### Description
The only logic change - regarding the recent changes to this Resubmission spec - is keeping the JSON vs the Workload config data in sync, thus, iterating over the Tasks/Steps (when existing) and updating those accordingly to the new parameters provided by the user.

The behaviour implemented in this PR was also quickly discussed with Hasan, and in summary, this is how it will behave during Resubmission workflow **creation**:
* ACDC of ReReco: resource parameters `Memory | TimePerEvent | Multicore | EventStreams` are either of type int or float, so regardless whether they are overwritten or not, both the JSON and the workload object will always contain the up-to-date and consistent value;
* ACDC of TaskChain: resource parameters `Memory | TimePerEvent | Multicore | EventStreams` can be either int/float or a dictionary.
  * If no value is provided, the top level parameter will be set to None (workload object isn't touched);
  * If value is provided as int/float, it gets applied to both top and ALL inner tasks;
  * If value is provided as a dictionary, top level will contain the dictionary, and the tasks will be updated after their keyname/value as provided by the user;
* ACDC of StepChain: resource parameters `Memory | TimePerEvent` have simple type (int/float), while `Multicore | EventStreams` can be either int/float or a dictionary. 
  * If `Memory | TimePerEvent` is provided as a dictionary, workflow creation fails;
  * If no value is provided, `Multicore | EventStreams` top level parameter will be set to None, while Memory | TimePerEvent remains untouched (workload object isn't touched).
  * If `Multicore | EventStreams` is provided as int/float, it gets applied to both top and ALL inner steps. If value is provided as a dictionary, top level will contain the dictionary, and the steps will be updated after their keyname/value as provided by the user (NOTE that the workload object has a bug #10791).

YES

#### Related PRs
Complement to #10781 

None

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/Resubmission.py
@@ -39,11 +39,15 @@ class ResubmissionWorkloadFactory(StdBase):
         helper.ignoreOutputModules(self.ignoredOutputModules)
 
         # override a couple of parameters, if provided by user
-        if 'RequestPriority' in arguments:
+        # Note that if it was provided by the user, then it's already part of the arguments too
+        if "RequestPriority" in self.userArgs:
             helper.setPriority(arguments["RequestPriority"])
-        helper.setMemory(arguments['Memory'])
-        helper.setCoresAndStreams(arguments['Multicore'], arguments.get("EventStreams", 0))
-        helper.setTimePerEvent(arguments.get("TimePerEvent"))
+        if "Memory" in self.userArgs:
+            helper.setMemory(arguments["Memory"])
+        if "Multicore" in self.userArgs or "EventStreams" in self.userArgs:
+            helper.setCoresAndStreams(arguments["Multicore"], arguments.get("EventStreams", 0))
+        if "TimePerEvent" in self.userArgs:
+            helper.setTimePerEvent(arguments.get("TimePerEvent"))
 
         return helper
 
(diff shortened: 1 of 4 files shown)


amaltaro:
And voilà! Finally, this is working as expected. Time to squash it, update the description, and fire up a review request.

haozturk:
Thanks a lot Alan! I read the description and it all looks good to me.

Issue #10401, opened 2021-03-31
Add support for GPU parameters at StepChain spec level


amaltaro:
**Impact of the new feature**
ReqMgr2

**Is your feature request related to a problem? Please describe.**
It's a new project to support GPU processing in central production workflows, where WM system will be the bridge between Offline/CMSSW and the grid resources made available to CMS (through the glideinWMS layer).

**Describe the solution you'd like**
This GH issue depends on #10388, which contains all the specific details to be implemented in this ticket as well.

Then, the solution expected from this issue is to support the same `RequiresGPU` and `GPUParams` parameters at the StepChain spec. Those 2 parameters should also be supported at the Step dictionary level. In case the Step dictionary is missing these 2 parameters, they should be inherited from the top level specification (just as it's done for Multicore).

Note that the same validation should also apply to the Step level parameters.

**Describe alternatives you've considered**
Not to validate the **optional** parameters.

**Additional context**
Major discussion happened here: https://github.com/cms-sw/cmssw/pull/33057
and description of these parameters: https://github.com/cms-sw/cmssw/pull/33057#issuecomment-810914270

Related to #11410

amaltaro:
@khurtado we might just update this GH issue (from the action item created in today's WMCore meeting).

amaltaro:
As discussed in different venues, StepChain requests are not suitable for different resource requirements among the different steps.
Having said that, the only option we have at the moment is to enforce the whole StepChain workflow to require GPU, whenever any of its steps requires GPU.

Issue #10400, opened 2021-03-31
Add support for GPU parameters at TaskChain spec level


amaltaro:
**Impact of the new feature**
ReqMgr2

**Is your feature request related to a problem? Please describe.**
It's a new project to support GPU processing in central production workflows, where WM system will be the bridge between Offline/CMSSW and the grid resources made available to CMS (through the glideinWMS layer).

**Describe the solution you'd like**
Detailed specification has been provided in this GH issue: https://github.com/dmwm/WMCore/issues/10388

which this one actually depends on.

Then, the solution expected from this issue is to support the same `RequiresGPU` and `GPUParams` parameters at the TaskChain spec. Those 2 parameters should also be supported at the Task dictionary level. In case the Task dictionary is missing these 2 parameters, they should be inherited from the top level specification (just as it's done for Multicore).

Note that the same validation should also apply to the Task level parameters.

**Describe alternatives you've considered**
Not to validate the **optional** parameters.

**Additional context**
Major discussion happened here: https://github.com/cms-sw/cmssw/pull/33057
and description of these parameters: https://github.com/cms-sw/cmssw/pull/33057#issuecomment-810914270

PR #10799, opened 2021-09-10
Add support and validation to GPU in StdBase/ReReco

### LINKED ISSUE #10388: Add support for GPU parameters at ReReco spec level
**Impact of the new feature**
ReqMgr2

It's a new project to support GPU processing in central production workflows, where WM system will be the bridge between Offline/CMSSW and the grid resources made available to CMS (through the glideinWMS layer).

There will be a series of such tickets, such that we can try to break work down in even parts.

In short, we need to support new workflow spec attributes for the ReReco/DataProcessing spec modules.

Now expanding on the important details:
* create a new spec attribute named `RequiresGPU`, with one of the possible 3 string values:
  * `forbidden`: must not use GPUs (**default**)
  * `optional`: can use GPU, if possible
  * `required`: must use GPUs
* create another spec attribute named `GPUParams`, which will contain all the necessary information to match/leverage GPU resources. It's meant to be JSON encoded python dictionary (thus of `string` type). There are 3 required and 3 optional parameters. The required parameters are:
  * `GPUMemory`: integer with the amount of memory, in Megabytes (MB). Validate as `> 0`. E.g.: 8000
  * `CUDACapabilities`: a list of short strings (could be real though). Validation should ensure at least one item in the list and matching this regex: `r"^\d+.\d$"`. E.g.: ["7.5", "8.0"]
  * `CUDARuntime`: a short string with the runtime version (could be real though). Validated against this regex: `r"^\d+.\d+$"`. E.g.: "11.2"
The list of optional parameters are:
  * `GPUName`: a string with the GPU name. Validate against `<= 100 chars`. E.g. "Tesla T4", "Quadro RTX 6000";
  * `CUDADriverVersion`: a string with the CUDA driver version. Validated against this regex: `r"^\d+.\d+\d+$"`E.g. "460.32.03"
  * `CUDARuntimeVersion`: a string with the CUDA runtime version. Validated against this regex: `r"^\d+.\d+\d+$"`E.g. "11.2.152"
* another validation is, if `RequiresGPU` is `optional` or `required`, then we should enforce `GPUParams` to contain the 3 *required* parameters with correct values;
* same way if `RequiresGPU` is `forbidden`, then `GPUParams` should be an empty string (FIXME: not sure we should enforce it(?))
* all this validation should happen only at workflow creation, which means those two new attributes are NOT allowed to change after the workflow is created. 

Just as an example, here is how `GPUParams` required parameters would look like:
```
'{"GPUMemory": 8000, "CUDARuntime": "11.2", "CUDACapabilities": ["7.5", "8.0"]}'
```

and this is an example with the full set of parameters:
```
'{"GPUMemory": 8000, "CUDARuntime": "11.2", "CUDACapabilities": ["7.5", "8.0"], "GPUName": "Tesla T4", "CUDADriverVersion": "460.32.03", "CUDARuntimeVersion": "11.2.152"}'
```

Not to validate the **optional** parameters.

**Additional context**
Major discussion happened here: https://github.com/cms-sw/cmssw/pull/33057
and description of these parameters: https://github.com/cms-sw/cmssw/pull/33057#issuecomment-810914270

amaltaro:
Fixes #10388 

#### Status
ready

#### Description
This PR implements GPU functionality within WMCore (only at the request-level, job level will be done in a different issue/PR).
Summary of changes is:
*  New request spec parameters called 'RequiresGPU' and 'GPUParams', where:
  * `RequiresGPU`: can be one of these values `("forbidden", "optional", "required")`, with default value to `forbidden`, thus not using GPUs
  * `GPUParams`: a dictionary JSON encoded, with a default value to None JSON encoded. It must be provided if RequiresGPU=optional or RequiresGPU=required.

List of **mandatory** parameters within `GPUParams` is:
* `GPUMemoryMB` (renamed from GPUMemory !): integer greater than 0
* `CUDACapabilities`: a list of string values. Each value must match the `CUDA_VERSION_REGEX` regular expression and max length.
* `CUDARuntime`: a string value matching the `CUDA_VERSION_REGEX` constraints.

And a list of the 3 **optional** parameters is:
* `GPUName`: a string value with less than 100 chars
* `CUDADriverVersion`: a string value matching the `CUDA_VERSION_REGEX` constraints.
* `CUDARuntimeVersion`: a string value matching the `CUDA_VERSION_REGEX` constraints.

NOTE: full support in TaskChain and StepChain is going to be done in a different GH issue/pull request, but the bulk of the development is already provided in this PR.

NO (new feature!)

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -10,11 +10,12 @@ from __future__ import print_function, division
 
 from future import standard_library
 standard_library.install_aliases()
-from builtins import str
-from future.utils import viewvalues
+from builtins import str, bytes
+from future.utils import viewvalues, viewkeys
 
 import io
 import re
+import json
 
 from urllib.parse import urlparse, urlunparse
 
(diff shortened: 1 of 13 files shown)


amaltaro:
copying a few names from the GH issue discussion @justinasr @mrceyhun @fwyzard @srimanob @hufnagel 

This is still a work in progress, but I wanted to draw your attention that this is happening and this is the current candidate implementation to go into the Workload Management system, hopefully in the next week.

The most important input/feedback that I would like to get from you is, whether you see any inconsistencies or use cases that are not covered with the current schema (data type + length + regular expression). Thank you very much!

amaltaro:
The baseline development to support GPUs within WMCore is provided in this PR, which also supports those new workflow arguments at the ReReco spec type. TaskChain and StepChain will be addressed in different issues/PR.

I'm going to run some extra tests in the next hours, and if everything goes fine, I will request these changes to be deployed in cmsweb-testbed today, run a final validation, and deploy this to production as well on Thursday.

amaltaro:
Looking at a real request JSON, I see this:
```
  "RequiresGPU": "forbidden",
  "GPUParams": "\"\"",
```

which doesn't look great with those scaped chars. Maybe we should default it to encoded `None` instead (thus, `'null'`). Checking...

amaltaro:
I'm going to need this code merged, such that I can resume working on its implementation for TaskChain https://github.com/dmwm/WMCore/issues/10400 and StepChain https://github.com/dmwm/WMCore/issues/10401

Basic tests went fine in my VM. I'm going to squash the 5th commit with the 1st one, and from my side it's good to go.
However, Todor, please feel free to leave your review. If something needs to be addressed, I can do so in the other(s) PR to be created today touching this GPU code.

src/python/WMCore/Lexicon.py
```diff
@@ -710,6 +711,103 @@ def activity(candidate):
...
+      * optional: GPUName (str), CUDADriverVersion (str), CUDARuntimeVersion (str)
+    :param candidate: a JSON encoded data to be validated
+    :return: True if validation succeeded, False or exception otherwise
+    """
+    mandatoryArgs = set(["GPUMemoryMB", "CUDARuntime", "CUDACapabilities"])
+    optionalArgs = set(["GPUName", "CUDADriverVersion", "CUDARuntimeVersion"])
+    try:
+        data = json.loads(candidate)
+    except Exception:
+        raise AssertionError("GPUParams is not a valid JSON object")
```
todor-ivanov, line 729: Sorry for the question @amaltaro , but why should we raise Assertion error here? It seems strange to me to get a general exception and raise assertion error instead and from the message it even seems like one related to a specific case related to the data structure itself.

src/python/WMCore/Lexicon.py
```diff
@@ -710,6 +711,103 @@ def activity(candidate):
...
+    # are there unknown arguments in the data provided?
+    unknownArgs = paramSet - mandatoryArgs - optionalArgs
+    if unknownArgs:
+        msg = "GPUParams contains arguments that are not supported. Args provided: {}, ".format(paramSet)
+        msg +="while mandatory args are: {} and optional args are: {}".format(mandatoryArgs, optionalArgs)
+        raise AssertionError(msg)
+    return _gpuInternalParameters(data)
+
+
+CUDA_VERSION_REGEX = {"re": r"^\d+\.\d+(\.\d+)?$", "maxLength": 100}
```
todor-ivanov, line 752: if this one  is about to be in the global scope, I'd say we should move it at the top of the file similar to others defined on lines 27-31

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -851,6 +851,31 @@ def setCoresAndStreams(self, cores, nStreams, initialTask=None):
...
 
+    def setGPUSettings(self, requiresGPU, gpuParams, initialTask=None):
+        """
+        Setter method for the workload GPU parameters.
+        It's responsible for setting whether GPUs are required or not; and
+        which GPU parameters to be used for that. This is done for every
+        task of this spec.
+        :param requiresGPU: string defining whether GPUs are needed. For TaskChains, it
+            could be a dictionary key'ed by the taskname.
+        :param gpuParams: GPU settings. A JSON encoded object, from either a None object
```
todor-ivanov, line 862: What happens if it is a json encoded  from None. I can guess, of course, but it is a little bit obscure.

src/python/WMCore/WMSpec/WMTask.py
```diff
@@ -1468,6 +1468,42 @@ def setNumberOfCores(self, cores, nStreams):
 
         return
 
+    def setTaskGPUSettings(self, requiresGPU, gpuParams):
+        """
+        Setter method for the GPU settings, applied to this Task object and
+        all underneath CMSSW type step object.
+        :param requiresGPU: string defining whether GPUs are needed. For TaskChains, it
+            could be a dictionary key'ed by the taskname.
+        :param gpuParams: GPU settings. A JSON encoded object, from either a None object
```
todor-ivanov, line 1477: Same comment as the one given bellow.

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -851,6 +851,31 @@ def setCoresAndStreams(self, cores, nStreams, initialTask=None):
...
+        if not requiresGPU:
+            return
+
+        if initialTask:
+            taskIterator = initialTask.childTaskIterator()
+        else:
+            taskIterator = self.taskIterator()
+
+        for task in taskIterator:
+            task.setTaskGPUSettings(requiresGPU, gpuParams)
```
todor-ivanov, line 875: The following comment is just for my personal education. (I am pretty sure you have already double checked this).
Isn't it a repeated recursion, when combined with the one from `setTaskGPUSettings()` on line: https://github.com/dmwm/WMCore/pull/10799/files#diff-81efe0a8bcf6b4cb2d5ee526c24a027563fa129b414de2aca75e78f1b38acbf1R1503

src/python/WMCore/Lexicon.py
amaltaro: Lexicon checks either return False or raise an `AssertionError` in case of failures during the input data validation.
This is also the standard behaviour of the `check` function in this module.

src/python/WMCore/Lexicon.py
amaltaro: That could be done as well. However, looking at this module, you can see that the regular expression is usually defined closer to the function that will consume it, so I just kept the consistency.
Anyhow, this will be refactored once a decision is made on how to separate Lexicon logic from lexicon rules/regex.

todor-ivanov (approved):
Thanks  for this quite big PR @amaltaro . From all what I managed to grasp in this whole picture it looks good to me. Of course I cannot catch possible errors better than the ones you already have with the tests you've done. I just left few comments inline related to some clarity of the code while reading. None of them is related to a possible problems, so they are  not blockers in any case.

src/python/WMCore/WMSpec/WMWorkload.py
amaltaro: This:
```
In [2]: json.dumps(None)
Out[2]: 'null'
```
which will be the default value in the specs/workflows. Also defined in the StdBase spec file.

src/python/WMCore/WMSpec/WMWorkload.py
amaltaro: That's a good question! And indeed it's confusing!
The WMWorkload `taskIterator` method only iterate through the top level tasks. Each workload usually (always?) has a single top level task.
While the recursion in WMTask iterates through all sub-tasks. Example, from a Processing task, we have a Merge sub-task, then we also have a LogCollect and Cleanup sub-tasks.

amaltaro:
Thank you very much for this review, Todor.
I totally agree that this is too big for having a decent code review. That's why I have separated the TaskChain and StepChain cases in their own issues, otherwise it would be even bigger and more complex.

Hopefully the manual tests and unit tests will be enough to make sure this code behaves ;)

amaltaro:
Here is a fairly good documentation for these GPU developments: https://github.com/dmwm/WMCore/wiki/GPU-Support

PR #10801, opened 2021-09-13
Fix Issue creating tmp files

### LINKED ISSUE #10800: Issue creating /tmp/cmst0/.wmcore_cache file
**Impact of the bug**
T0 Agent not able to create/run any job/workflow using WMCore agent 1.5.2

**Describe the bug**
T0 use the Tier0Feeder component to create the sandbox. When Tier0 started to test a new release based con WMCore 1.5.2, Tier0Feeder component was unable to start properly given the following error:
```
2021-09-13 06:55:27,603:140108589012736:WARNING:Service:The cachefile /tmp/cmst0/.wmcore_cache_xlpsm5ls/cmssdt.cern.ch/-3895757188456107843_GET__SDT_cgi-bin_ReleasesXML_ does not exi
st and the service at https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML/ raised a FileNotFoundError(2, 'No such file or directory') when accessed
2021-09-13 06:55:27,673:140108589012736:ERROR:StdBase:About to raise exception <@========== WMException Start ==========@>
Exception Class: WMSpecFactoryException
Message: [Errno 2] No such file or directory: '/tmp/cmst0/.wmcore_cache_xlpsm5ls/cmssdt.cern.ch/-3895757188456107843_GET__SDT_cgi-bin_ReleasesXML_'
        ClassName : None
        ModuleName : WMCore.WMSpec.WMWorkloadTools
        MethodName : _validateArgFunction
        ClassInstance : None
        FileName : /data/tier0/srv/wmagent/3.0.0/sw.jamadova/slc7_amd64_gcc630/cms/t0/3.0.0/lib/python3.8/site-packages/WMCore/WMSpec/WMWorkloadTools.py
        LineNumber : 105
        ErrorNr : 0
```

**How to reproduce it**
Deploying the T0 (3.0.0) agent and start all the components of the agent or only Tier0Feeder.

**Expected behavior**
The /tmp/cmst0/.wmcore_cache file must be correctly created and finally Tier0Feeder will be able to create the sandbox

Issue is **only** affecting when T0 use WMCore 1.5.2.

jhonatanamado:
Fixes #10800 

#### Status
Tested with a Tier0-replay using WMCore agent 1.5.2

#### Description
While Tier0 started to test the new python3 version of WMCore, the component Tier0Feeder that is in charge of create the sandbox was not able to start properly. This due that the temp file within the following path `/tmp/cmst0/.wmcore_cache_xxxxxx was not created.

To guarantee that the file is properly created, the path where the file will be written needs to be create before creating the file. _My hypothesis is that file manipulation/creation changes from py2 to py3_ .

No. Dual stack DMWM services may be affected 

#### Related PRs
No

Relies on `pathlib` library from python3

### DIFF
--- a/src/python/WMCore/Services/Service.py
+++ b/src/python/WMCore/Services/Service.py
@@ -294,11 +294,22 @@ class Service(dict):
                     cachefile.write(data)
                     cachefile.seek(0, 0)  # return to beginning of file
                 else:
-                    with open(cachefile, 'w') as f:
-                        if isinstance(data, dict) or isinstance(data, list):
-                            f.write(json.dumps(data))
-                        else:
-                            f.write(data)
+                    # Ensure that the path for /tmp file is created before
+                    # creating the file
+                    pathfile=cachefile.split("/")
+                    pathfile="/"+"/".join(pathfile[1:-1])
+                    if not os.path.exists(pathfile):
+                        os.makedirs(pathfile, exist_ok=True)
+                    try:
+                        # Raise excpetion if file/path is not created
+                        with open(cachefile, 'w') as f:
+                            if isinstance(data, dict) or isinstance(data, list):
+                                f.write(json.dumps(data))
+                            else:
+                                f.write(data)
+                    except EnvironmentError as EnvError:
+                        msg = ' raised a %s when accessed' % EnvError.__repr__()
+                        self['logger'].error(msg)
 
 
         except (IOError, HttpLib2Error, HTTPException) as he:


amaltaro:
@jhonatanamado after looking into this once again, I feel that we should actually try to understand the root cause of this problem.
This module is used for all of the http requests made in WMCore, and there will be a performance penalty with the current implementation (besides other problems raised in the unit tests).

I'm inclined to believe that there is something special with the T0/user/VM machinery, which causes this problem. We did not see this in any of the production agents, including when deploying Py3 WMAgent in a brand new node.

Perhaps you and @todor-ivanov could investigate this further?

mapellidario:
From a quick look, i think many of the failing unittests in py2 are caused by the fact that `pathlib` is not available in that environment. Until we support py2, i would suggest using plain old `os.path` instead, that should work in both py2 and py3.

Sorry for the not-requested comment! :grimacing: :)

jhonatanamado:
Hi Alan, I used `os.makedirs` function to give dual-stack support. But I will schedule a short session with @todor-ivanov to check if something is failing with the configuration of the T0 vms.

src/python/WMCore/Services/Service.py
```diff
@@ -294,11 +294,22 @@ def getData(self, cachefile, url, inputdata=None, incoming_headers=None,
...
                 else:
-                    with open(cachefile, 'w') as f:
-                        if isinstance(data, dict) or isinstance(data, list):
-                            f.write(json.dumps(data))
-                        else:
-                            f.write(data)
+                    # Ensure that the path for /tmp file is created before
+                    # creating the file
+                    pathfile=cachefile.split("/")
+                    pathfile="/"+"/".join(pathfile[1:5])
```
germanfgv, line 300:
```suggestion
                    pathfile="/"+"/".join(pathfile[1:-1])
```
this would be a bit more reliable, e. g. the path for container's caché is a bit different, so the harcoded `5` doesn't work

jhonatanamado: Thanks for the suggestion @germanfgv. I added the changes. Lets keep using this PR to test and deploy the T0 agents but Im still investigating the root of this issue.

PR #10809, opened 2021-09-17
PhysicsSkim issue

### LINKED ISSUE #10806: Tier0 is not able to find physics_skims using the agent 1.5.2
**Impact of the bug**
Tier0 agent. Tier0Feeder is down due that is not able to find physics_skims for prompt-reco. Not able to lunch promt-reco and replay fails.

**Describe the bug**
With the T0 agent based on WMCore 1.5.2, the Tier0Feeder component is not able to find physics skims for promt-reco. If the dataset doesn't have physics_skims in the configuration as follows:
```
DATASETS = ["Cosmics"]
 
for dataset in DATASETS:
    addDataset(tier0Config, dataset,
               do_reco=True,
               write_reco=False,
               write_aod=True,
               write_miniaod=False,
               write_dqm=True,
               alca_producers=["SiStripCalCosmics","SiPixelCalCosmics","TkAlCosmics0T","MuAlGlobalCosmics"],
               #physics_skims=["CosmicSP", "CosmicTP", "LogError", "LogErrorMonitor"],
               timePerEvent=0.5,
               sizePerEvent=155,
               scenario=cosmicsScenario)
```
Tier0 is able to run prompt-reco but if physics_skims appear in the configuration the T0 agent is down and none of the promtreco workflows will run
**How to reproduce it**
Deploy T0 agent (3.0.0) based on WMCore 1.5.2

**Expected behavior**
Promptreco workflows should be able to run including the physics_skims parameter

Error message from the Componentlog is as follows:
`ERROR:StdBase:About to raise exception Can't find physics skim LogError in CMSSW_12_0_0`

jhonatanamado:
Fixes #10806 

#### Status
Tested with a replay

#### Description
During the final step of migrating the T0 agent to python3 and based on WMCore agent 1.5.2, The replay fails due that Tier0Feeder was not able to find physics_skims associated to CMSSW. The changes are basically two. One related to the `subprocess.Popen` due that in python3 is needed to decode the bytes from standard output to strings. Also to use CMSSW 12_X_X is needed to modify the `command` to use and get proper results from `getSkimDataTier`

MAYBE. Need clarification if this PR will break dual-stack support for other DMWM services

#### Related PRs
No

No

### DIFF
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -10,6 +10,9 @@ from builtins import range, object
 
 import logging
 import json
+
+from Utils.PythonVersion import PY3
+from Utils.Utilities import decodeBytesToUnicodeConditional
 from Utils.Utilities import makeList, makeNonEmptyList, strToBool, safeStr
 from WMCore.Cache.WMConfigCache import ConfigCache, ConfigCacheException
 from WMCore.Lexicon import couchurl, procstring, activity, procversion, primdataset, gpuParameters
@@ -142,7 +145,8 @@ class StdBase(object):
         import subprocess
 
         p = subprocess.Popen("/cvmfs/cms.cern.ch/common/cmsos", stdout=subprocess.PIPE, shell=True)
-        cmsos = p.communicate()[0].strip()
+        cmsos = p.communicate()[0]
+        cmsos = decodeBytesToUnicodeConditional(cmsos, condition=PY3).strip()
 
         scramBaseDirs = glob.glob("/cvmfs/cms.cern.ch/%s*/cms/cmssw/%s" % (cmsos, cmsswVersion))
         if not scramBaseDirs:
@@ -154,12 +158,17 @@ class StdBase(object):
         command += "cd %s\n" % scramBaseDirs[0]
         command += "eval `scramv1 runtime -sh`\n"
 
-        command += """python -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
+        if PY3:
+            command += """python3 -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
+        else:
+            command += """python -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
+
         command += """dataTier = getSkimDataTier("%s")\n""" % skim
-        command += """if dataTier:\n\tprint dataTier.value()'"""
+        command += """if dataTier:\n\tprint(dataTier.value())'"""
 
         p = subprocess.Popen(command, stdout=subprocess.PIPE, shell=True)
-        dataTier = p.communicate()[0].strip()
+        dataTier = p.communicate()[0]
+        dataTier = decodeBytesToUnicodeConditional(dataTier, condition=PY3).strip()
 
         if dataTier == "None":
             dataTier = None


src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -142,7 +142,7 @@ def skimToDataTier(cmsswVersion, skim):
         import subprocess
 
         p = subprocess.Popen("/cvmfs/cms.cern.ch/common/cmsos", stdout=subprocess.PIPE, shell=True)
-        cmsos = p.communicate()[0].strip()
+        cmsos = p.communicate()[0].decode('utf-8').strip()
```
amaltaro, line 145: I'd suggest to use the str/bytes utils that we have been adopting in the code with the python migration, so something like:
```suggestion
        cmsos = p.communicate()[0]
        cmsos = decodeBytesToUnicodeConditional(cmsos, condition=PY3).strip()
```

Keep in mind you need to import them as well:
```
from Utils.PythonVersion import PY3
from Utils.Utilities import decodeBytesToUnicodeConditional
```

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -154,12 +154,13 @@ def skimToDataTier(cmsswVersion, skim):
...
-        command += """python -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
+        command += """python3 -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
         command += """dataTier = getSkimDataTier("%s")\n""" % skim
-        command += """if dataTier:\n\tprint dataTier.value()'"""
+        command += """if dataTier:\n\tprint (dataTier.value())'"""
 
+        logging.info("command= %s" % command)
         p = subprocess.Popen(command, stdout=subprocess.PIPE, shell=True)
-        dataTier = p.communicate()[0].strip()
+        dataTier = p.communicate()[0].decode('utf-8').strip()
```
amaltaro, line 163: Same comment as above, regarding the adoption of `decodeBytesToUnicodeConditional`

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -154,12 +154,13 @@ def skimToDataTier(cmsswVersion, skim):
         command += "cd %s\n" % scramBaseDirs[0]
         command += "eval `scramv1 runtime -sh`\n"
 
-        command += """python -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
```
amaltaro, line 157: and here we could keep the dual compatibility by
```
if PY3:
   ### create the python3 command
else:
  ### create the python2 command
```

amaltaro (changes requested):
Thanks for coming up with this solution, Jhonatan. I left a few comments along the code that should guarantee the code works in both py2 and py3. Please let me know if you have any questions or need any further help. Thanks

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
amaltaro: Or actually, something like:
```
        whichPython = "python3" if PY3 else "python"
        command += """%s -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n""".format(whichPython)
        command += """dataTier = getSkimDataTier("%s")\n""" % skim
        command += """if dataTier:\n\tprint(dataTier.value())'"""
```
should be enough. Could you please give this a try?

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
jhonatanamado: Done

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
jhonatanamado: First option works, but using `whichPython = "python3" if PY3 else "python"` fails again with the same issue.

amaltaro: Hmm, maybe because we would have to decode that string to bytes(?)
Anyhow, sorry for giving you another suggestion, but I think this code can be enhanced with something like:
```
        if PY3:
            command += """python3 -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
        else:
            command += """python -c 'from Configuration.StandardSequences.Skims_cff import getSkimDataTier\n"""
        command += """dataTier = getSkimDataTier("{}")\n""".format(skim)
        command += """if dataTier:\n\tprint(dataTier.value())'"""
```
Can you please provide these changes before running a replay to test it? Just to make sure you got the correct formatting in place. Thanks

amaltaro: @jhonatanamado Jhonatan, you are missing the white space between `print` and `datatier`, which has been provided in the discussion above. Can you please fix it?

amaltaro:
@jhonatanamado in addition to the comment made along the code, there is a conflict with this branch that needs to be fixed. One way to fix that is by rebasing your branch on top of the upstream recent changes. Here is a good reference that I have bookmarked: https://steveklabnik.com/writing/how-to-squash-commits-in-a-github-pull-request

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
jhonatanamado: Ups sorry! I will fixed asap

amaltaro:
test this please

amaltaro (approved):

amaltaro:
Changes look good to me, thanks Jhonatan!

amaltaro:
@jhonatanamado Jhonatan, here is a wmagent branch version of this patch:
https://github.com/dmwm/WMCore/pull/10816

which you should be able to cleanly apply on T0.

PR #10811, opened 2021-09-17
Propagate GPU requirements to job creation and submission

### LINKED ISSUE #10393: Propagate workflow GPU parameters to the job specification/classads
**Impact of the new feature**
WMAgent

Once this issue is addressed: https://github.com/dmwm/WMCore/issues/10388
the next step is to propagate the workflow level GPU parameters all the way down to the job creation, job submission and job classad.

In short, we need to propagate the workflow GPU configuration all the way to JobCreator, to be picked up by JobSubmitter and finally passed to the SimpleCondorPlugin, which should properly fill the condor job classad (and expressions where needed).

Trying to expand on the job classads (still needs further discussion with the SI team):
* workflow/task/step `RequiresGPU` mapped to job classad `RequiresGPU`, with one of the possible 3 string values:
  * `forbidden`: must not use GPUs (**default**)
  * `optional`: can use GPU, if possible
  * `required`: must use GPUs
* workflow/task/step `GPUMemoryMB` mapped to job classad `GPUMemoryMB`, string type (just as OriginalMemory is), default value to 0.
* workflow/task/step `CUDACapabilities` mapped to job classad `CUDACapabilities`, a comma separated string (just as DESIRED_Sites is), default value to empty string (should default value be `undefined` instead?);
* workflow/task/step `CUDARuntime` mapped to job classad `CUDARuntime`, string type, default to empty string (should default be `undefined` instead?)

The job matchmaking logic would be something like:
* if (`RequestGPU` == "forbidden") || (`RequestGPU` == "optional"), then forget GPUs in the matchmaking
* if `RequestGPU` == "required", then:
  * job's `GPUMemoryMB` <= node's `GPUMemoryMB`
  * at least one job's `CUDACapabilities` in the node's `CUDACapabilities`
  * job's `CUDARuntime` among those in the node's `CUDACompatibleRuntimes`

This description might be updated as we discuss it further with the SI team and converge on the best classad naming convention and data type.

This issue considers only the **required** parameters. In the future, we can also explore supporting the 3 optional spec parameters at job description level.

**Additional context**
 Major discussion happened here: cms-sw/cmssw#33057

amaltaro:
Fixes #10393 

#### Status
ready

#### Description
This PR implements the GPU handling on the WMAgent side, thus:
* fetch the workflow - actually the task - GPU requirements from the spec
* create the job pickle files with those GPU requirements in JobCreator
* load those GPU requirements in the JobSubmitter 
* define the necessary condor job classads in SimpleCondorPlugin

In terms of condor job description, it has been implemented as:
* `RequiresGPU`: a string that can have one of the following 3 values: ("forbidden", "optional", "required")
  * UPDATE: it will be an integer, either 1 (for required) or 0 (for optional/forbidden)
* `GPUMemoryMB`: undefined if not set in the workflow, otherwise a string object
* `CUDACapabilities`: undefined if not set in the workflow, otherwise a classad.quote object with capabilities comma-separated
  * UPDATE: Antonio suggested to use the default condor attribute, thus renamed to `CUDACapability`
* `CUDARuntime`: undefined if not set in the workflow, otherwise a classad.quote object

NOTE-1: note the custom logic for figuring out whether a task requires GPU or not (for cases with multiple cmsRun steps and with different settings). In short, `required` has precedence over `optional`, which has precedence over `forbidden`
NOTE-2: note that, if different GPUParams are specified for different steps in the same task; we will use the first one with valid key/value pairs (thus, which is not None).
NOTE-3: we are going to start supporting only the mandatory requirements. The remaining 3 optional arguments will be supported in the future, once we gather some experience with the current setup.

YES

#### Related PRs
Workflow specification provided in: https://github.com/dmwm/WMCore/pull/10799

None

### DIFF
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -126,7 +126,8 @@ def saveJob(job, workflow, sandbox, wmTask=None, jobNumber=0,
             owner=None, ownerDN=None, ownerGroup='', ownerRole='',
             scramArch=None, swVersion=None, agentNumber=0, numberOfCores=1,
             inputDataset=None, inputDatasetLocations=None, inputPileup=None,
-            allowOpportunistic=False, agentName=''):
+            allowOpportunistic=False, agentName='', requiresGPU='forbidden',
+            gpuRequirements=None):
     """
     _saveJob_
 
@@ -156,6 +157,8 @@ def saveJob(job, workflow, sandbox, wmTask=None, jobNumber=0,
     job['inputDatasetLocations'] = inputDatasetLocations
     job['inputPileup'] = inputPileup
     job['allowOpportunistic'] = allowOpportunistic
+    job['requiresGPU'] = requiresGPU
+    job['gpuRequirements'] = gpuRequirements
 
     with open(os.path.join(cacheDir, 'job.pkl'), 'wb') as output:
         pickle.dump(job, output, HIGHEST_PICKLE_PROTOCOL)
@@ -190,6 +193,8 @@ def creatorProcess(work, jobCacheDir):
         inputPileup = work.get('inputPileup', None)
         allowOpportunistic = work.get('allowOpportunistic', False)
         agentName = work.get('agentName', '')
+        requiresGPU = work.get('requiresGPU', 'forbidden')
+        gpuRequirements = work.get('gpuRequirements', None)
 
         if ownerDN is None:
             ownerDN = owner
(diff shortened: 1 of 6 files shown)


amaltaro:
Now that I have found your GH user name: @aperezca and @mmascher

Marco, Antonio, could you please check the SimpleCondorPlugin changes and confirm whether the choosen job classads and their data type is correct? 

Once I have your confirmation, I'm going to run a functional test to validate whether all the pieces are in place (without considering if the job runs or not).

amaltaro:
Evertyhing seems to be working fine. Once my tests are completed, I will squash it and it's ready to be merged.
PS.: I want to work on another issue that will touch exactly the same code, so I'd rather not have it lingering around unmerged for too long.

amaltaro:
This has been tested and jobs are properly created with the new parameters. Documentation has been updated as well, explicitly this section: https://github.com/dmwm/WMCore/wiki/GPU-Support#gpu-job-description-in-wmcore-x-glideinwms

I'm happy to follow up on any future comments, but things need to keep moving and I'm going to merge it now. Thanks!

PR #10839, opened 2021-09-29
[py2py3] make WMCore/Configuration.py suitable for CRABClient

mapellidario:
Investigation related to https://github.com/dmwm/CRABClient/issues/4974

#### Status
In development

#### Description

CRABClient needs to be shipped with CMSSW environment that do not ship `python-future`, and CRABClient relies on `WMCore.Configuration`. I am investigating if it is feasible to remove the dependency on python-future from WMCore.Configuration. I do not think this is the correct way to go, i am opening this PR just to have an opinion from jenkins. After it, we can discuss this approach openly. A full run of jenkins unittests is worth 100h of speculation.

As always: it should be, but we will only discover with certainty after we tested it

#### Related PRs
not yet

nope (WMCore in general will continue to depend on `python-future`!)

### DIFF
--- a/src/python/Utils/Utilities.py
+++ b/src/python/Utils/Utilities.py
@@ -6,6 +6,8 @@ from future.utils import viewitems
 from builtins import str, bytes
 from past.builtins import basestring
 
+from Utils.PythonVersion import PY3
+
 import subprocess
 import os
 import re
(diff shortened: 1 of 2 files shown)


mapellidario:
This experiment failed badly! We will go back to the original plan, closing this. sorry for the noise!

PR #10850, opened 2021-10-04
add two new CMSSW exit code: 8034 and 8501

belforte:
reference:
https://github.com/cms-sw/cmssw/blob/3abab3eebee2b7c3a8ff31fb491470b0834d53ec/FWCore/Utilities/interface/EDMException.h#L68-L70


amaltaro:
@belforte Stefano, I assume you do not want to put a few words more in there to make it a more meaningful error message? 8034 is probably clear enough as is, but the second looks generic to me. Anyways, it's all up to you!

belforte:
I just copied the description from CMSSW repo. You surely well know the story about the GUID check. But for the other i have no clue. Details if any belongs in FWCore repository IMHO.

PR #10673, opened 2021-07-19
first commit to enable archs != x86_64

tommasoboccali:
Fixes https://github.com/dmwm/WMCore/issues/10840

First attempt for supporting ARCHS != x86_64

See the presentations @ DRP: https://indico.cern.ch/event/1049288/

#### Status
tested on a local environment

#### Description
just 2 changes: 
- submit.sh: instead of using *_amd64_*. the middle part of the COMP repository is taken from `uname -m`
- SimpleCondorPlugin.py: for worker nodes NOT using x86_64, the ARCH requirement in Condor cannot be left blank (otherwise the WMA ARCH is used). A function to prepare the proper requirement is proposed. 

Please note that most probably this will not work with CMS@HOME on PPC (while the current behavior for cms@home on x86 should be ok) since in that case additional requirements are already present, and I did not want to touch them.

#### Is it backward compatible (if not, which system it affects?)
In my setup it is (based on 1.4.7.patch5). Should be checked for consistency on a release closer to master

#### Related PRs
None

#### External dependencies / deployment changes
no

PS: this is NOT including the changes needed for multi arch (running on x84 OR ppc, for example); this will be done later after discussions with the DMWM developers

PR #10853, opened 2021-10-05
Support different architectures during job submission and runtime

### LINKED ISSUE #10840: Add support to Power PC 64 workflows and jobs
**Impact of the new feature**
ReqMgr2 and WMAgent

Related to: https://github.com/dmwm/WMCore/issues/10674

but one step before, in the sense that first we need to have support to `ppc64le` architectures. Next step, which is requested in 10674, would be to support a mix of x86 and ppc64 architectures within the same task/jobs.

We should ensure that workflows requesting `ppc64le` architecture can be properly created and that the job description requirements properly specify this architecture.
A check list of expected solutions from this issue is:
* support `ppc64le` workflows in ReqMgr2 (ensure the request description passes the key/value validation functions)
* ensure that the condor job description is properly requesting the correct architecture (I believe it only specifies the OS, not the architecture)
* ensure that the job runtime code loads the needed libraries from the correct scramArch
 
None

**Additional context**
Some initial solution has already been provided in: https://github.com/dmwm/WMCore/pull/10673

amaltaro:
Superseeds https://github.com/dmwm/WMCore/pull/10673
Fixes #10840 

#### Status
not-tested

#### Description
Summary of changes is:
* support different CPU architectures at the job wrapper (actually, CVMFS is only setup for `ppc64le` and `x86_64`. Architecture `aarch64` will only be supported in the future)
  * this means, CVMFS scripts will be sourced accordingly
* added a map of ScramArch to architecture in the BasePlugin
  * note that there is no default architecture, in case a job does not specify ScramArch (only known case is Cleanup, which will then allow it to run in any architecture)
  * note that the result must be a single architecture, so some precedence logic goes embedded
* overwrite the `Requirements` expression string according to the architecture that we want to run that job.  

UPDATE: The SI team discussed this with the HTCondor developers, and indeed `TARGET.Arch` in the requirements is the way to use such resources.

maybe

#### Related PRs
None

None

### DIFF
--- a/etc/submit.sh
+++ b/etc/submit.sh
@@ -121,12 +121,17 @@ echo -e "======== WMAgent CMS environment load finished at $(TZ=GMT date) ======
 
 
 echo "======== WMAgent COMP Python bootstrap starting at $(TZ=GMT date) ========"
-# First, decide which COMP ScramArch to use based on the required OS
+# First, decide which COMP ScramArch to use based on the required OS and Architecture
+THIS_ARCH=`uname -m`  # if it's PowerPC, it returns `ppc64le`
+if [ "$THIS_ARCH" = "x86_64" ]
+then
+    THIS_ARCH="amd64"
+fi
 if [ "$REQUIRED_OS" = "rhel7" ];
 then
-    WMA_SCRAM_ARCH=slc7_amd64_gcc630
+    WMA_SCRAM_ARCH=slc7_${THIS_ARCH}_gcc630
 else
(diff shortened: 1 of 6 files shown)


spigad:
Hi @amaltaro  just spoke with @tommasoboccali we agreed to go ahead so far we've not things to add. 

And about multi-arch I agree we need further discussions ..
thanks Daniele

todor-ivanov (approved):
Thanks @amaltaro. There is just a single question inline. You may want to take a look at it.

PR #10859, opened 2021-10-08
Update wmagent-deploy script to deploy Py3 WMAgent by default; update to cycle 1.5.4

### LINKED ISSUE #10858: Make Python3 the default option for WMAgent deployment
**Impact of the new feature**
WMAgent deploy script

No, but as we have been running Py3 WMAgent in a solid state, and the Py2 stack is approaching its end of life. We should make sure that python3 is the default value whenever possible.

Update the deploy-wmagent.py script and make the python3 RPM the default choice, thus renaming the `py3` option as `py2`.
In addition to that, we can also update the deploy scripts with the latest stable release going under validation right now.

None

**Additional context**
None

amaltaro:
Fixes #10858 

#### Status
ready

#### Description
If someone wants to deploy WMAgent in python2 stack, a `--py2` option must be provided in the command line. Otherwise, it will deploy a Py3 WMAgent by default.

Also update the CMSWEB and WMCore tag to the latest stable release.

I would suggest to also apply the DBS3 client patch with the following command line:
```
curl https://patch-diff.githubusercontent.com/raw/dmwm/DBS/pull/655.patch | patch -p 4 -d sw/slc7_amd64_gcc630/cms/py3-dbs3-pycurl-client/3.16.2000/lib/python3.8/site-packages/
```

yes

#### Related PRs
None

None

### DIFF
--- a/bin/adhoc-scripts/parseUnifiedCampaigns.py
+++ b/bin/adhoc-scripts/parseUnifiedCampaigns.py
@@ -282,15 +282,11 @@ def insertTestCampaigns(mgr):
                    'SecondaryLocation': ["T1_US_FNAL", "T2_CH_CERN"],
                    'SiteBlackList': [], 'SiteWhiteList': ["T1_US_FNAL", "T2_CH_CERN"]}
 
-    testCamp = ("CMSSW_10_6_1_Step3", "CMSSW_10_6_1_patch1_Step1", "CMSSW_10_6_1_patch1_Step2",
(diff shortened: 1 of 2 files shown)


todor-ivanov (commented):
Thanks @amaltaro. The changes look good to me. But I wouldn't remove the `-3|--py3` option though. Having a python3 deployment as a default is the right way to go indeed, but it is not good to remove the option parameter I think.

PR #10890, opened 2021-11-03
Add validation to the SubRequestType parameter

### LINKED ISSUE #10282: Add spec creation validation for the SubRequestType parameter
**Impact of the new feature**
ReqMgr2

Given that we started using this field to take decisions in different parts of the system (mostly microservices as far as I remember), we better add some validation and make sure the data type and the value is as expected.

Here is its definition: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1035

Add a validation function for the `SubRequestType` field. It needs to be evaluated during request creation, and request creation should fail if the provided value does not pass the validate function.

The supported values for this parameter are (still discussing with Justinas/PdmV the last bits):
```
["MC", "ReDigi", "Pilot", "RelVal", "HIRelVal", "ReReco"]
``` 
if value is not provided, it should default to an empty string.

Not add this validation and suffer the consequences in the future.

**Additional context**
none

amaltaro:
Fixes #10282 

#### Status
ready

#### Description
The `SubRequestType` base spec parameter has been adopted already by a couple of Microservices, and we are going to extend its use in the framework, so we better start validating these values to make sure that the expected values are provided.

If request is created with an invalid value, it fails workflow creation.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/Lexicon.py
+++ b/src/python/WMCore/Lexicon.py
@@ -702,6 +702,15 @@ def primaryDatasetType(candidate):
     raise AssertionError("Invalid primary dataset type : %s should be 'mc' or 'data' or 'test'" % candidate)
 
 
+def subRequestType(candidate):
+    subTypes = ["MC", "ReDigi", "Pilot", "RelVal", "HIRelVal", "ReReco", ""]
+    if candidate in subTypes:
+        return True
+    # to sync with the check() exception when it doesn't match
+    msg = "Invalid SubRequestType value: '{}'. Allowed values are: {}".format(candidate, subTypes)
+    raise AssertionError(msg)
+
+
 def activity(candidate):
     dashboardActivities = ['reprocessing', 'production', 'relval', 'tier0', 't0',
                            'harvesting', 'storeresults', 'integration',
--- a/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
+++ b/src/python/WMCore/WMSpec/StdSpecs/StdBase.py
@@ -15,8 +15,9 @@ from Utils.PythonVersion import PY3
 from Utils.Utilities import decodeBytesToUnicodeConditional
 from Utils.Utilities import makeList, makeNonEmptyList, strToBool, safeStr
 from WMCore.Cache.WMConfigCache import ConfigCache, ConfigCacheException
-from WMCore.Lexicon import couchurl, procstring, activity, procversion, primdataset, gpuParameters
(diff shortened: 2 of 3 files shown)


amaltaro:
Tested in my VM. It has to be considered when we resume working on the Lexicon refactoring.

src/python/WMCore/Lexicon.py
```diff
@@ -702,6 +702,15 @@ def primaryDatasetType(candidate):
     raise AssertionError("Invalid primary dataset type : %s should be 'mc' or 'data' or 'test'" % candidate)
 
 
+def subRequestType(candidate):
+    subTypes = ["MC", "ReDigi", "Pilot", "RelVal", "HIRelVal", "ReReco", ""]
```
todor-ivanov, line 706: Does this set of subRequestTypes consider also all the possible (if any) sub types for T0?

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
```diff
@@ -1052,7 +1053,7 @@ def getWorkloadCreateArgs():
                      "RunNumber": {"default": 0, "type": int},
                      "RobustMerge": {"default": True, "type": strToBool},
                      "Comments": {"default": ""},
-                     "SubRequestType": {"default": ""},  # used only(?) for RelVals
+                     "SubRequestType": {"default": "", "validate": subRequestType},
```
todor-ivanov, line 1056: Are we sure if T0 does not need to validate the subrequest type it wont break here?

todor-ivanov (approved):
Thanks for this changes  @amaltaro. I did approve them, but also left two comments inline, just to get sure that while coding you considered those cases I am talking about. Feel free to ignore them if you think they are irrelevant.

src/python/WMCore/Lexicon.py
amaltaro: AFAIK T0 does not use it. If they do, we can then extend it in the future.

src/python/WMCore/WMSpec/StdSpecs/StdBase.py
amaltaro: This validation will be triggered for every single workflow creation. But the default value will ensure it's backwards compatible with those not using it.

amaltaro:
Thanks for this review, Todor. Release 1.5.5.pre4 coming in in a few minutes.

PR #10899, opened 2021-11-10
Update request state diagram with archival done by MS

### LINKED ISSUE #10898: Request state transition diagram needs to be updated with MS archiving
**Impact of the bug**
WMCore documentation

**Describe the bug**
With the implementation of MSRuleCleaner, we decided that that service would also be responsible for the workflow archival.

**How to reproduce it**
none

**Expected behavior**
Workflow archival is no longer performed by ReqMgr2, but by MSRuleCleaner, which does the cleaning of no longer needed rucio rules and finally move the workflow to the next status (aborted-archived, rejected-archived, normal-archived).

none

amaltaro:
Fixes #10898 

#### Status
ready

#### Description
Updated only the transition to *-archived state, which is now (actually for a few months) performed by MSRuleCleaner instead of ReqMgr2.

yes

#### Related PRs
none

none


todor-ivanov:
Hi @amaltaro Sorry for commenting without been asked for a review. I do find the diagram perfect, thanks for updating it. What I also noticed though, is we are missing two requests statuses from the wiki page referring to the diagram [1] - `staging` and `staged`.

[1]
https://github.com/dmwm/WMCore/wiki/Request-Status

todor-ivanov (approved):
All looks perfect @amaltaro :)

amaltaro:
Thanks for pointing out to what was written in the wiki. I somehow was looking only at the diagram.

PR #10902, opened 2021-11-19
Adding --deep option to getReplicaInfoForBlocks

### LINKED ISSUE #10901: Add --deep option to getReplicaInfoForBlocks
**Impact of the new feature**
WMAgent, T0Agent, MicroServices? 

The current version of `RucioInjectorPooler.deleteBlocks()` relies on Rucio DB's `collection_replicas` to be properly syncrhonized. Given the recent sync issues in CMS rucio server we would like to have the option of using the more reliable (although costly) `--deep` flag for dataset queries.

We would like to add an attribute to RucioInjector, stating the value of the deep flag (`default=False`). That value should be passed to the method [Rucio.getReplicaInfoForBlocks()](https://github.com/dmwm/WMCore/blob/4b6676a4350addd15b6e9abe48a4e41e5217d920/src/python/WMCore/Services/Rucio/Rucio.py#L239). If `deep=True`, the method should set the deep flag when querying for dataset replicas.

**Additional context**
Currently, the agent uses `list_dataset_replicas_bulk` which does not currently support --deep flag. With Rucio issue [#5024](https://github.com/rucio/rucio/issues/5024) I requested that option to be added to the API. In the mean time, `getReplicaInfoForBlocks` cam make use of the regular `list_dataset_replicas`

germanfgv:
Fixes #10901 

#### Status
ready

#### Description
Added `deep` option to  [Rucio.getReplicaInfoForBlocks()](https://github.com/dmwm/WMCore/blob/4b6676a4350addd15b6e9abe48a4e41e5217d920/src/python/WMCore/Services/Rucio/Rucio.py#L239). The value of the flag can be changed in RucioInjector config.

Also added slicing to [RucioInjectorPooler.deleteBlocks()](https://github.com/dmwm/WMCore/blob/4b6676a4350addd15b6e9abe48a4e41e5217d920/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py#L325) to avoid  problems in case of a high number of deletable datasets.

YES

#### Related PRs
NO

NO

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -21,6 +21,7 @@ from WMCore.DAOFactory import DAOFactory
 from WMCore.Services.Rucio.Rucio import Rucio, WMRucioException, RUCIO_VALID_PROJECT
 from WMCore.WMException import WMException
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
+from Utils.IteratorTools import grouper
 
 
 class RucioInjectorException(WMException):
@@ -77,6 +78,9 @@ class RucioInjectorPoller(BaseWorkerThread):
                            hostUrl=config.RucioInjector.rucioUrl,
                            authUrl=config.RucioInjector.rucioAuthUrl,
                            configDict={'logger': self.logger})
+
+        self.useDsetReplicaDeep = getattr(config.RucioInjector, "useDsetReplicaDeep", False)
+        self.delBlockSlicesize = getattr(config.RucioInjector, "delBlockSlicesize", 100)
 
         # metadata dictionary information to be added to block/container rules
         # cannot be a python dictionary, but a JSON string instead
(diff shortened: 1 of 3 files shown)


src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -78,6 +79,9 @@ def __init__(self, config):
                            authUrl=config.RucioInjector.rucioAuthUrl,
                            configDict={'logger': self.logger})
 
+        self.deep = getattr(config.RucioInjector, "deepFlag", False)
```
todor-ivanov, line 82: What about changing the name of this flag to something more descriptive. like `useDsetReplicaDeep` ?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -78,6 +79,9 @@ def __init__(self, config):
...
+        self.sliceSize = getattr(config.RucioInjector, "sliceSize", 100)
```
todor-ivanov, line 83: Same  here. Something like: `blockSliceSize` or even `delBlockSlicesize`?

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -343,67 +347,72 @@ def deleteBlocks(self):
...
-                    if self.rucio.deleteRule(rule['id'], purgeReplicas=True):
-                        logging.info("Successfully deleted rule: %s, for block %s.", rule['id'], block)
-                        deletedRules += 1
-                    else:
-                        logging.warning("Failed to delete rule: %s, for block %s. Will retry in the next cycle.", rule['id'], block)
-                if deletedRules == len(rules):
-                    binds.append({'DELETED': 1, 'BLOCKNAME': block})
-                    logging.info("Successfully deleted all rules for block %s.", block)
 
+            self.markBlocksDeleted.execute(binds)
```
todor-ivanov, line 413: It seems to me this piece needs to be one  level up, otherwise it would account only for blocks deleted  during the last slice. Need to double check that one  myself once again.

todor-ivanov: no need to take actions on this one, I'll come back to this again later

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -268,10 +268,17 @@ def getReplicaInfoForBlocks(self, **kwargs):
             inputDids.append({"scope": kwargs["scope"], "type": "DATASET", "name": block})
 
         resultDict = {}
-        for item in self.cli.list_dataset_replicas_bulk(inputDids):
-            resultDict.setdefault(item['name'], [])
-            if item['state'].upper() == 'AVAILABLE':
-                resultDict[item['name']].append(item['rse'])
+        if kwargs['deep']:
```
todor-ivanov, line 271: please reflect the addition in the deep argument in the docstring of the method

todor-ivanov (changes requested):
Hi @germanfgv, Thank you very much for this PR! I managed to take a quick look and no major  issues fond so far. But I think there is one  place we need to get clear/double check about before merging, in order not to miss marking some blocks as deleted.  And also maybe a flag renaming may be is not a bad idea - -comments inline.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: I think you are right. If I keep this here, it'll  work well if there are no issues. But, if the routine is interrupted at some point, the DB will be left outdated. I'll fix it.

todor-ivanov: thanks @germanfgv

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -343,67 +347,72 @@ def deleteBlocks(self):
...
+                    msg += "Will retry again in the next cycle. Error: %s" % str(exc)
+                    logging.error(msg)
+                    continue
 
-            for blockRSEs in availableRSEs:
-                # If block is available at every RSE its container needs to be transferred, the block can be deleted
-                blockSites = set(blockRSEs['replica'])
-                if cont['rse'].issubset(blockSites):
-                    blocksToDelete.append(blockRSEs['name'])
+                for blockRSEs in availableRSEs:
```
jhonatanamado, line 380: hi @germanfgv , Do you think is it possible to check for not all the `availableRSEs`? Just the RSEs that fits the following expression:
`'tier=0|tier=1|T2_CH_CERN'`

amaltaro:
test this please

src/python/WMCore/Services/Rucio/Rucio.py
amaltaro: And you probably need a default value. Something like:
```
kwargs.setdefault("deep", False)
```

src/python/WMCore/Services/Rucio/Rucio.py
```diff
@@ -268,10 +268,17 @@ def getReplicaInfoForBlocks(self, **kwargs):
...
+            for block in blockNames:
```
amaltaro, line 272: For the sake of avoiding confusion and keeping consistency, I think it would be better to iterate through `inputDids` instead of `blockNames`. What do you think?

Instead of passing `deep=True`, I would suggest to replace it by `deep=kwargs['deep']`.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
amaltaro: Maybe even `useDsetReplicaDeepFlag`

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -343,67 +347,72 @@ def deleteBlocks(self):
...
 
-            try:
-                # Get RSE in which each block is available
-                availableRSEs = self.rucio.getReplicaInfoForBlocks(block=cont['blocks'])
-            except Exception as exc:
-                msg = "Failed to get replica info for blocks in container: %s.\n" % contName
-                msg += "Will retry again in the next cycle. Error: %s" % str(exc)
-                logging.error(msg)
-                continue
+                try:
```
amaltaro, line 371: I am actually concerned with all these logical changes here, especially because we don't have tests for the deletion logic. Can't you keep most of the existent logic and change only this try except to use the block slicing parameter? It would become something like:
```
availableRSEs = self.rucio.getReplicaInfoForBlocks(block=sliceOfBlocks, deep=self.deep)
```
and then you refer to this `sliceOfBlocks` in the code below, when required.

amaltaro (changes requested):
German, I left some comments along the code for your consideration. The RucioInjectorPoller was only partially reviewed, since I have a suggestion that might trigger a substantial amount of changes/revert.

In addition to that, can you please update the unit test(s) for the `getReplicaInfoForBlocks` method? Such that we are properly testing the use of the `deep` flag as well? Thanks

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: The logic here hasn't changed. I'm only using the `sliceOfBlocks` to populate the containers dict. Everything else does exactly the same as before, but using the smaller container dictionary.

src/python/WMCore/Services/Rucio/Rucio.py
germanfgv: Ok, it makes sense.

src/python/WMCore/Services/Rucio/Rucio.py
germanfgv: I'll add the docstring and default value.

todor-ivanov: thanks @germanfgv

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -341,69 +345,74 @@ def deleteBlocks(self):
...
-            for blockRSEs in availableRSEs:
-                # If block is available at every RSE its container needs to be transferred, the block can be deleted
-                blockSites = set(blockRSEs['replica'])
-                if cont['rse'].issubset(blockSites):
-                    blocksToDelete.append(blockRSEs['name'])
+                for blockRSEs in availableRSEs:
+                    # If block is available at every RSE its container needs to be transferred, the block can be deleted
+                    blockSites = set(blockRSEs['replica'])
+                    logging.debug("BlockName: %s", blockRSEs['name'])
+                    logging.debug("Needed: %s / Available: %s", (str(cont['rse']), str(blockSites) ))
```
todor-ivanov, line 384: This is an error. You either use a string substitution operator and a tuple while creating the log message, or put all sub strings needed in the logger's variable argument list and let it do the string expansion for you. So this one should be one  of those:
```
logging.debug("Needed: %s / Available: %s" % (str(cont['rse']), str(blockSites) ))
```
or
```
logging.debug("Needed: %s / Available: %s", str(cont['rse']), str(blockSites))
```

todor-ivanov:
Thanks for the changes  @germanfgv. I noticed few errors which are also notified in the Pylint tests.  Please take a look at the latest comment I made inline and also of the few of errors mentioned in the Pylint test details. More particularly the ones in Bold need to be addressed. At least the ones coming from your  change. 

There are also two more unresolved conversations above(one for simplifying the logic with @amaltaro's suggestion the other from me about accounting for failed deletions). Those I'd like to keep unresolved, so we may refer to them later and keep an eye or address in a separate issue  if needed.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
germanfgv: yes,  pylint was complaining abour the first, so I  tried to change it to the second and missed the parenthesis. I'll fix it.

PR #10905, opened 2021-11-24
Use siteLocalConfig in logCollect jobs to specify output site

### LINKED ISSUE #10903: SiteLocalConfig support in LogCollect jobs
**Impact of the new feature**
WMAgent, T0 WMAgent, WMStats, MicroServices (?)

By default, LogCollect jobs copy job logs to T2_CH_CERN. The only way to modify this behavior is to provide stage out params via the `overrides` section in the step's configuration. All the parameters required to configure the stage out process can be obtain by loading a SiteLocalConfig file. Nonetheless, it is currently not possible to directly use such file on LogCollect jobs.

This would be particularly useful for T0, as we intend to isolate our operations from T2_CH_CERN as much as possible.

Provide means to use SiteLocalConfig to specify the output data placement for LogCollect.

Use the current method it's ok, although not ideal. 

**Additional context**

germanfgv:
Fixes #10903

#### Status
ready

#### Description
Add possibility of specifying stage out parameters via siteLocalConfig.
Add setters and getters for steps override section

YES

### DIFF
--- a/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
@@ -22,6 +22,7 @@ from WMCore.Storage.StageOutMgr import StageOutMgr
 from WMCore.WMRuntime.Tools.Scram import Scram, getSingleScramArch, isCMSSWSupported
 from WMCore.WMSpec.Steps.Executor import Executor
 from WMCore.WMSpec.Steps.WMExecutionFailure import WMExecutionFailure
+from WMCore.Storage.SiteLocalConfig import loadSiteLocalConfig
 
 
 class LogCollect(Executor):
@@ -65,6 +66,9 @@ class LogCollect(Executor):
         overrides = {}
         if hasattr(self.step, 'override'):
             overrides = self.step.override.dictionary_()
+            if overrides.get('logRedirectSiteLocalConfig', None):
+                siteCfg = loadSiteLocalConfig()
+                overrides.update(siteCfg.localStageOut)
 
         # Set wait to over an hour
         waitTime = overrides.get('waitTime', 3600 + (self.step.retryDelay * self.step.retryCount))


src/python/WMCore/WMSpec/Steps/Template.py
```diff
@@ -62,6 +63,22 @@ def environment(self):
...
+    def setOverrides(self, **overrides):
+        """
+        _setOverrides_
+        set a overrides dictionary
+        """
+        for key, value in viewitems(overrides):
+            setattr(self.data.override, key, value)
+        return   
+
+    def getOverrides(self):
```
todor-ivanov, line 75: @germanfgv I think this method is already redefinition of an already existing one:
https://github.com/dmwm/WMCore/blob/0d667606fed0166f304f48c2fc85b9c3b32cc0f9/src/python/WMCore/WMSpec/WMStep.py#L170

it must already be present into the object becuase you inevitably inherit from `WMStep`

src/python/WMCore/WMSpec/Steps/Template.py
```diff
@@ -62,6 +63,22 @@ def environment(self):
         """
         return self.data.environment
 
+    def setOverrides(self, **overrides):
```
todor-ivanov, line 66: @germanfgv  I think (similarly as bellow) an equivalent of this method is already present into the object:
 https://github.com/dmwm/WMCore/blob/0d667606fed0166f304f48c2fc85b9c3b32cc0f9/src/python/WMCore/WMSpec/WMStep.py#L157

Of course the already existing one is having the unpleasant need to call it for setting every single override parameter one by one. But if you insist on having the option to pass those `overrides` on a single pas as a dictionary (which I personally find quite useful), please  move it to the proper place. i.e. here:

https://github.com/dmwm/WMCore/blob/0d667606fed0166f304f48c2fc85b9c3b32cc0f9/src/python/WMCore/WMSpec/WMStep.py#L168

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -63,18 +64,24 @@ def execute(self, emulator=None):
         overrideCatalog = getattr(self.step.application, 'overrideCatalog', None)
 
         overrides = {}
+        params = {}
         if hasattr(self.step, 'override'):
             overrides = self.step.override.dictionary_()
+            if 'siteLocalConfig' in overrides.keys():
```
todor-ivanov, line 70: @germanfgv I understand and agree with you this is the best way of setting the parameter through which we can overwrite the logcollect destination. But since this config section is not used solely by T0. (unified is also using it), what I would suggest (and what I was actually trying to express during the meeting) is to set an additional Bool flag  notifying the trigger of the new component behavior like: 

```
if 'redirectLogcollect' in overrides.keys() and overrides['redirectLogcollect']: 
```
(suggesting the flag to trigger the new component behavior to be called `redirectLogcollect`)

todor-ivanov: and then with an additional parameter to notify the method to be used for this redirection: 
```
if 'redirectLogcollect' in overrides.keys() and overrides['redirectLogcollect']: 
    if 'useSiteLocalConfig' in overrides.keys() and overrides['useSiteLocalConfig']: 
        < ... new behaviour here.. >
    elif 'useStaticConfig' in overrides.keys():
        < ...some new  behavior here triggered by the 'useStaticConfig' parameter... > 
         ...
```

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -63,18 +64,24 @@ def execute(self, emulator=None):
...
         # Set wait to over an hour
         waitTime = overrides.get('waitTime', 3600 + (self.step.retryDelay * self.step.retryCount))
 
         # hardcode CERN EOS T2_CH_CERN stageout parameters
         eosStageOutParams = {}
-        eosStageOutParams['command'] = overrides.get('command', "xrdcp")
-        eosStageOutParams['option'] = overrides.get('option', "--wma-disablewriterecovery")
-        eosStageOutParams['phedex-node'] = overrides.get('phedex-node', "T2_CH_CERN")
-        eosStageOutParams['lfn-prefix'] = overrides.get('lfn-prefix', "root://eoscms.cern.ch//eos/cms")
+        eosStageOutParams['command'] = params.get('command', "xrdcp")
```
todor-ivanov, line 81: and in the light of  those changes  this variable  name is no longer relevant, since once we merge this  PR `eos` is not going to be the sole destination for logcollect

todor-ivanov: a better name would be: `logtStageoutParams` @amaltaro what do you think here?

todor-ivanov: and this would also require some additional changed in the component's structure further down: look at the comments bellow

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -63,18 +64,24 @@ def execute(self, emulator=None):
...
+        eosStageOutParams['option'] = params.get('option', "--wma-disablewriterecovery")
+        eosStageOutParams['phedex-node'] = params.get('phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = params.get('lfn-prefix', "root://eoscms.cern.ch//eos/cms")
 
         try:
             eosStageOutMgr = StageOutMgr(**eosStageOutParams)
```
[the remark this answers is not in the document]
todor-ivanov: Also few more log messages down (at which I cannot comment because are not part of this commit obviously) need to be changed. Those were strictly relying on the fact that the logcollect was actually hardwired to CERN and with this change now they become quite misleading. And all that said, the proper log message may need to be triggered again based on the bool vairable triggering the redirection.

todor-ivanov (changes requested):
Hi @germanfgv It took me a while to re-implement the configuration changes  suggested here manually and test  few stuff. Based on my tests I've  added few comments inline. Plese take a look and let me know. I am ready to help, if you need me, at any time.

src/python/WMCore/WMSpec/Steps/Template.py
germanfgv: You are right, I missed this.

src/python/WMCore/WMSpec/Steps/Template.py
germanfgv: You are right, I missed this.

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
germanfgv: @todor-ivanov With the current master-branch code, it is already possible to redirect the logs. The current method is simply to specify the stageout params in the overrides dictionary, without bolean flags (the presence the parameters is used as the _flag_). The introduction of `redirectLogcollect` would modify the behavior for other users of this code. 

What I am doing here is to use the presence of `siteLocalConfig` in the overrides dict to trigger the new behaviour (The params dict is not necessary and I will remove it). This should in no way modify the regular behaviour, as other users will not set the `siteLocalConfig` key in theri `overrides dictionary`. Am I missing sometnhing?

src/python/WMCore/WMSpec/Steps/Executors/LogCollect.py
```diff
@@ -65,6 +66,9 @@ def execute(self, emulator=None):
         overrides = {}
         if hasattr(self.step, 'override'):
             overrides = self.step.override.dictionary_()
+            if 'siteLocalConfig' in overrides.keys() and overrides['siteLocalConfig']:
```
amaltaro, line 69:
```suggestion
            if overrides.get('siteLocalConfig', None):
```

amaltaro (changes requested):
German, can you please link the required changes on the T0 system to trigger this behavior? Are you using the spec `OverrideCatalog` creation argument?

todor-ivanov:
Thanks @germanfgv

todor-ivanov (approved):
The change looks good to me. Thanks @germanfgv

PR #10906, opened 2021-11-24
Adapt MSUnmerged to sync RSE data with MongoDB properly

### LINKED ISSUE #10797: MSUnmerged: Move from stateless to stateful design
**Impact of the new feature**
MSUnmerged

In order to:
* Have better accounting on the service and 
* Avoid some side effects due to external glitches and 
* Minimize memory footprint
* Improve reports of the RSEs by the service REST interfaces 

We decided  to move from a stateless to a stateful design. 

For the purpose, the steps that needs to be performed are as follows: 
 * A database to be deployed - MongoDb would be the first chose, since we already have a MonoDb server instance running into a dedicated pod in K8 (so far serving MSOutput only). 
 * Create the relevant databases and collections.
 * Create/develop the interface needed to start uploading/reading the RSE objects into and out from the database
 * Improve the reports by fetching the results and RSE status from the database

Continue running stateless.

**Additional context**
None

todor-ivanov:
Fixes #10797 

#### Status
ready

#### Description
The current PR is supposed to resolve few  issues which were introduced with adding MongoDB as a backend database for the service, which are making the service not backward compatible:
 * After preserving the timestamps into the database we are getting into the situation that a RSE would be retried only once in a through out a span of a single Rucio Consistency Monitor poling cycle. now we should be able to retry the RSE in case it was not previously completely cleaned
 * If we have a fresh new record for the RSE in the Rucio Consistency Monitor and we do not wipe out all the provious file records for that RSE in MongoDB, then we are about to start accumulating old records. With this PR we are resetting the object if a brand new record is present in Rucio Consistency Monitor.

NO

#### Related PRs
No additional configurations PRs are required, but it is worth mentioning the previous PR:
https://github.com/dmwm/WMCore/pull/10895

### DIFF
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
@@ -29,6 +29,7 @@ except ImportError:
     gfal2 = None
 
 from pymongo import IndexModel
+from pymongo.errors  import NotMasterError
 
 # WMCore modules
 from WMCore.MicroService.DataStructs.DefaultStructs import UNMERGED_REPORT
@@ -109,6 +110,7 @@ class MSUnmerged(MSCore):
         self.msConfig.setdefault("mongoDBUrl", 'mongodb://localhost')
         self.msConfig.setdefault("mongoDBPort", 27017)
         self.msConfig.setdefault("mongoDB", 'msUnmergedDB')
+        self.msConfig.setdefault("mongoDBRetryCount", 3)
         self.msConfig.setdefault("mongoDBReplicaset", None)
 
         msUnmergedIndex = IndexModel('name', unique=True)
@@ -163,6 +165,9 @@ class MSUnmerged(MSCore):
         # The basic /store/unmerged regular expression:
         self.regStoreUnmergedLfn = re.compile("^/store/unmerged/.*$")
         self.regStoreUnmergedPfn = re.compile("^.+/store/unmerged/.*$")
+
+        # log msConfig
+        self.logger.info("msConfig: %s", pformat(self.msConfig))
 
     # @profile
     def execute(self):
@@ -415,7 +420,7 @@ class MSUnmerged(MSCore):
 
     def consRecordAge(self, rse):
         """
-        A method to heck the duration of the consistency record for the RSE
+        A method to check the duration of the consistency record for the RSE
         :param rse: The RSE to be checked
         :return:    rse or raises MSUnmergedPlineExit
         """
@@ -431,15 +436,38 @@ class MSUnmerged(MSCore):
         isConsNewer = self.rseConsStats[rseName]['end_time'] > self.rseTimestamps[rseName]['prevStartTime']
         if not isConsNewer:
             msg = "RSE: %s With old consistency record in Rucio Consistency Monitor. " % rseName
-            msg += "Skipping it in the current run."
-            self.logger.info(msg)
-            raise MSUnmergedPlineExit(msg)
+            if 'isClean' in rse and rse['isClean']:
+                msg += "And the RSE has been cleaned during the last Rucio Consistency Monitor polling cycle."
+                msg += "Skipping it in the current run."
+                self.logger.info(msg)
+                self.updateRSETimestamps(rse, start=False, end=True)
+                raise MSUnmergedPlineExit(msg)
+            else:
+                msg += "But the RSE has NOT been cleaned during the last Rucio Consistency Monitor polling cycle."
+                msg += "Retrying cleanup in the current run."
+                self.logger.info(msg)
         if not isConsDone:
             msg = "RSE: %s In non-final state in Rucio Consistency Monitor. " % rseName
             msg += "Skipping it in the current run."
             self.logger.warning(msg)
+            self.updateRSETimestamps(rse, start=False, end=True)
             raise MSUnmergedPlineExit(msg)
 
+        if isConsNewer and isConsDone:
+            # NOTE: If we've got to this point then we have a brand new record for
+            #       the RSE in RucioConMOn and we are then about to start a new RSE cleanup
+            #       so we will need to wipe out all but the timestamps from both
+            #       the current object and the MongoDB record for the object.
+            msg = "RSE: %s With new consistency record in Rucio Consistency Monitor. " % rseName
+            msg += "Resetting RSE and starting a fresh cleanup process in the current run."
+            self.logger.info(msg)
+            try:
+                rse.resetRSE(self.msUnmergedColl, keepTimestamps=True, retryCount=self.msConfig['mongoDBRetryCount'])
+            except NotMasterError:
+                msg = "Could not reset RSE to MongoDB for the maximum of %s mongoDBRetryCounts configured." % self.msConfig['mongoDBRetryCount']
+                msg += "Giving up now. The whole cleanup process will be retried for this RSE on the next run."
+                msg += "Duplicate deletion retries may cause error messages from false positives and wrong counters during next polling cycle."
+                raise MSUnmergedPlineExit(msg)
         return rse
 
     # @profile
@@ -554,6 +582,7 @@ class MSUnmerged(MSCore):
             rse['counters']['dirsToDeleteAll'] = len(rse['files']['toDelete'])
             return rse
 
+        self.logger.info("rse: %s", twFormat(rse, maxLength=8))
         # If we are here, then there are service filters...
         for dirName in rse['dirs']['toDelete']:
             # apply exclusion filter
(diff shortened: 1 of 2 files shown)


src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -431,15 +431,32 @@ def consRecordAge(self, rse):
         isConsNewer = self.rseConsStats[rseName]['end_time'] > self.rseTimestamps[rseName]['prevStartTime']
         if not isConsNewer:
             msg = "RSE: %s With old consistency record in Rucio Consistency Monitor. " % rseName
-            msg += "Skipping it in the current run."
-            self.logger.info(msg)
-            raise MSUnmergedPlineExit(msg)
+            if rse['isClean']:
```
amaltaro, line 434: With the current definition of `isClean`, doesn't it mean that the service will keep trying to delete files until it manages to delete every single file that is listed as deletable? I'm afraid we need to redefine how we consider an RSE as clean or not.

Another option, which is likely better, would be to add a new key/value pair to the MSUnmergedRSE object to hold the original timestamp - when the file dump was collected - provided by Rucio ConMon. Such that we end up with an expression like:
```
isConsNewer = self.rseConsStats[rseName]['end_time'] > self.rseTimestamps[rseName]['dumpTStamp']
```

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -431,15 +431,32 @@ def consRecordAge(self, rse):
...
+                msg += "And the RSE has been cleaned during the last Rucio Consistency Monitor pooling cycle."
```
amaltaro, line 435: `pooling` --> `polling`. Same for a few lines below.

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
```diff
@@ -74,13 +82,17 @@ def buildMongoProjection(self, fullRSEToDB=False):
             mongoProjection.update({"files": True})
         return mongoProjection
 
-    def readRSEFromMongoDB(self, collection):
+    def readRSEFromMongoDB(self, collection, useProjection=False):
```
amaltaro, line 85: Don't forget updating the docstring please.

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
```diff
@@ -145,3 +157,22 @@ def writeRSEToMongoDB(self, collection, fullRSEToDB=False, overwrite=False):
...
+        """
+        resetDoc = self.defaultDoc()
+
+        if keepTimestamps:
+            resetDoc['timestamps'] = self['timestamps']
+        if keepCounters:
+            resetDoc['counters'] = self['counters']
+
+        self.update(resetDoc)
+        writeResult = self.writeRSEToMongoDB(collection, fullRSEToDB=True, overwrite=True)
```
amaltaro, line 177: Isn't `fullRSEToDB` meant to be parameterized? It's not clear to me why it's set to True here.

amaltaro (changes requested):
Todor, I had a quick review and left some comments for your consideration.

In addition to that, I wonder if this is actually a required change for the upcoming CMSWEB release?

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: thanks ! :)

vkuznet: It is always dangerous to use constructs like `if rse['isClean']` since you never know rse data type at run-time, it is much safer to use the following `if rse and 'isClean' in rse`. Here you always check that rse exists, and then check if certain key exists in a directionary.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: With the current definition of `isClean` we refresh/consider as new the RSE cleanup status on every RucioConMon polling cycle. Meaning if RucioConMon provides a new dump of the RSE, we reset everything - counters and cleanup status (keeping only the timestamps/updating them as usual), and we retry a fresh cleanup regarding the new RSE situation. So if there have been some leftovers from previous Runs those must have been detected by RucioConMon and will get into the our next iteration. If we have cleaned the RSE last time. We should consider it unclean now and retry again - in case there is nothing to be deleted it will enter immediately in clean status, if there is something new - it will be deleted together with leftovers (if any).

The idea of preserving the RucioConMon timestamp is not bad, let me check what I can do to put it there.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: thanks @vkuznet ,  I was brave here because we make sure on several other places that we do have a common, well defined and universal `rse` data structure, which should always include that field. But you are correct, staying on the safe side and following good practice principles, this should be written the way you suggest here. I will fix that.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Hi @amaltaro, I did implement this and as simple as it sounds from the line you suggested, at the end it became a quite horrible change, because I also had to check the status of the RucioConMon record for the RSE not only here  but also in the `updateRSETimestamps` method, so we can be sure to update the record for this timestamp in MongoDB only if the RucioConMon has done a complete and successful scan of the RSE, basically re implementing the whole `consRecordAge` method, and at the end we will get exactly the same result. I think this is going to make things more error prone than simplifying them.

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
todor-ivanov: done

src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
todor-ivanov: And I have just found a bug which caused the the transformation of the data type for some fields when we writing/resetting the RSE record to MongoDB, e.g. 
```
In [67]: rse.resetRSE(msUnmergedColl)
Out[67]: True

In [68]: rse
Out[68]: 
{'name': 'T2_DE_DESY',
 'pfnPrefix': None,
 'isClean': False,
 'timestamps': {'rseConsStatTime': 0.0,
  'prevStartTime': 0.0,
  'startTime': 0.0,
  'prevEndTime': 0.0,
  'endTime': 0.0},
 'counters': {'totalNumFiles': 0,
  'dirsToDeleteAll': 0,
  'dirsToDelete': 0,
  'filesToDelete': 0,
  'deletedSuccess': 0,
  'deletedFail': 0},
 'files': {'allUnmerged': [],
  'toDelete': {},
  'protected': {},
  'deletedSuccess': [],
  'deletedFail': []},
 'dirs': {'allUnmerged': set(), 'toDelete': set(), 'protected': set()},
 '_id': ObjectId('61a0e052b713a7692b8475f9')}

In [69]: msUnmergedColl.find_one(rse.mongoFilter)
Out[69]: 
{'_id': ObjectId('61a0e052b713a7692b8475f9'),
 'name': 'T2_DE_DESY',
 'pfnPrefix': None,
 'isClean': False,
 'timestamps': {'rseConsStatTime': 0.0,
  'prevStartTime': 0.0,
  'startTime': 0.0,
  'prevEndTime': 0.0,
  'endTime': 0.0},
 'counters': {'totalNumFiles': 0,
  'dirsToDeleteAll': 0,
  'dirsToDelete': 0,
  'filesToDelete': 0,
  'deletedSuccess': 0,
  'deletedFail': 0},
 'files': {'allUnmerged': [],
  'toDelete': [],
  'protected': [],
  'deletedSuccess': [],
  'deletedFail': []},
 'dirs': {'allUnmerged': [], 'toDelete': [], 'protected': []}}
```

Notice the type of all the fields which are of filter/generator type (e.g. `rse['files']['toDelete']`). The fix is in place already, but this one was tricky to debug and I was just lucky to notice it before merging the PR.

todor-ivanov:
Thanks for your reviews @amaltaro and @vkuznet .  I've addressed what I feel safe enough to do at this very moment. I have  tested extensively the logic involved. Also found a bug and fixed that one too. I will try to merge this tonight because I'd like to push this with the tag I am about to cut any moment now. Please feel free to leave comments if you happen to look at it again after I have merged it.

PR #10908, opened 2021-11-24
Add replicaset opotion to the MongoDB wrapper.

### LINKED ISSUE #10907: MongoDB: Add possibility to connect to replica set
**Impact of the new feature**
MongoDB 

As we discovered  few hours ago, when one tries to connect to a high availiability setup of MongoDB (which is the case for the newly setup MongoDBAsAService) usually it is not a connection to a single server, but rather connection  to a replica set. If one connects to just a single server of such a setup it is possible to do normal search queries, but if the client is connected to a 'nonPrimary` server of the cluster write operations are not permitted and the error to be received is: [1] 

 While there is a fairly easy way to describe the whole cluster configuration to the Mongo Client we do not forward that information to the wrapper. As explained in the MongoDB documentation: [replica_set connection](https://pymongo.readthedocs.io/en/stable/examples/high_availability.html#id1)

We can do it either through the wrapper  and passing the replica name as a client parameter and then connect as normally:
```
client = MongoClient(server, port, replicaset=replicaset)
``` 
Or sending everything through the configuration only  directly into the url link  and not touching the wrapper code at all. Something like: 
```
client = MongoClient('mongodb://localhost:27017,localhost:27018/?replicaSet=replicaset')
```
 
Add the possibility to pass a full cluster description to the MongoDB wrapper we are using in WMCore.

If we decide to move to connections to MongoDBAsAService we have no alternative but to pass the information to the client correctly. 

**Additional context**
[1]
```
NotMasterError: not master, full error: {'topologyVersion': {'processId': ObjectId('61964003a108c9aadd686164'), 'counter': 5}, 'ok': 0.0, 'errmsg': 'not master', 'code': 10107, 'codeName': 'NotWritablePrimary', '$clusterTime': {'clusterTime': Timestamp(1637749346, 1), 'signature': {'hash': ..........$', 'keyId': 7031874994229477378}}, 'operationTime': Timestamp(1637749346, 1)}

```

todor-ivanov:
Fixes #10907 

#### Status
ready

#### Description
With the current PR we add the ability to connect to a named replicaset instead to a single MongoDB server. as explained in the issue we have two options of doing this either describing the whole cluster server by server in the MongoDB url as:
```
client = MongoClient('mongodb://localhost:27017,localhost:27018/?replicaSet=replicaset')
```
or just connecting to the load balanced name and provide the replicaset name so that  in case the connection has happened to a secondary instead of a primary backend server, the MonogoDB client may resolve the situation during write operations:
```
client = MongoClient(server, port, replicaset=replicaset)
```  

YES
#### Related PRs
The commits to service configuration for adding the option for MSUnmerged: 
[preprod](https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/123/diffs?commit_id=f7d7118623aefe1f70e13f15900166355cfe7892)
and 
[prod](https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/122/diffs?commit_id=781b8294bb740c520261b6654d9d6be93ccac167)
 

None

### DIFF
--- a/src/python/WMCore/Database/MongoDB.py
+++ b/src/python/WMCore/Database/MongoDB.py
@@ -19,10 +19,12 @@ class MongoDB(object):
                  database=None,
                  server=None,
                  port=None,
+                 replicaset=None,
                  create=False,
                  collections=None,
                  testIndexes=False,
-                 logger=None):
+                 logger=None,
+                 **kwargs):
         """
         :databases:   A database Name to connect to
         :server:      The server url (see https://docs.mongodb.com/manual/reference/connection-string/)
@@ -40,7 +42,10 @@ class MongoDB(object):
         self.port = port # 8230
         self.logger = logger
         try:
-            self.client = MongoClient(self.server, self.port)
+            if replicaset:
+                self.client = MongoClient(self.server, self.port, replicaset=replicaset, **kwargs )
+            else:
+                self.client = MongoClient(self.server, self.port, **kwargs)
             self.client.server_info()
         except Exception as ex:
             msg = "Could not connect to MongoDB server: %s\n%s" % (self.server, str(ex))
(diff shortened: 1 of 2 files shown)


todor-ivanov:
@amaltaro I am merging this one because I need it for few other issues. Please feel free to leave your comments later.

PR #10911, opened 2021-12-03
Files for GitHub Acton CI/CD

vkuznet:
This resolves feature request https://github.com/dmwm/WMCore/issues/10921

#### Status
In development

#### Description
This PR provides skeleton bulid.yml and Dockerfile which are used by GitHub Action. These file later should be extended to cover full build for all WMCore services. The build.yaml is already provide a release tar ball and Dockerfile already create an image, but it will require additional work to include all dependencies via system packages (apt-get or yum) and PyPi ones.

yes

#### Related PRs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,90 @@
+name: Build
+
+on:
+  push:
+    tags:
+      - '*.*.*'
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ '3.x', 'pypy-3.7' ]
+    steps:
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: |
+          #python3 -m pip install -r requirements_py3.txt
+          python3 setup.py install_system -s reqmgr2ms --prefix=install
+          tar cfz reqmgr2ms.targz $PWD/install
+
+    - name: Get the Ref
+      id: get-ref
+      uses: ankitvgupta/ref-to-tag-action@master
+      with:
+        ref: ${{ github.ref }}
+        head_ref: ${{ github.head_ref }}
+
+    - name: Build image
+      run: |
+        sed -i -e "s,ENV TAG=.*,ENV TAG=${{steps.get-ref.outputs.tag}},g" Dockerfile
+        cat Dockerfile
+        docker build . --tag registry.cern.ch/cmsweb/reqmgr2ms
+
+    - name: Login to registry.cern.ch
+      uses: docker/login-action@v1.6.0
+      with:
+        registry: registry.cern.ch
+        username: ${{ secrets.CERN_LOGIN }}
+        password: ${{ secrets.CERN_TOKEN }}
+
+    - name: Publish image to registry.cern.ch
+      uses: docker/build-push-action@v1
+      with:
+        username: ${{ secrets.CERN_LOGIN }}
+        password: ${{ secrets.CERN_TOKEN }}
+        registry: registry.cern.ch
+        repository: cmsweb/reqmgr2ms
+        tag_with_ref: true
+
+    - name: Create Release
+      id: create_release
+      uses: actions/create-release@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: ${{ github.ref }}
+        release_name: Release ${{ github.ref }}
+        draft: false
+        prerelease: false
+
+    - name: Upload amd64 binary
+      id: upload-auth-proxy-server
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ steps.create_release.outputs.upload_url }}
+        asset_path: ./reqmgr2ms.tar.gz
+        asset_name: reqmgr2ms.tar.gz
+        asset_content_type: application/octet-stream
+
+#     - name: Push new image to k8s
+#       run: |
+#         curl -ksLO https://raw.githubusercontent.com/vkuznet/imagebot/main/imagebot.sh
+#         sed -i -e "s,COMMIT,${{github.sha}},g" -e "s,REPOSITORY,${{github.repository}},g" -e "s,NAMESPACE,wma,g" -e "s,TAG,${{steps.get-ref.outputs.tag}},g" -e "s,IMAGE,registry.cern.ch/cmsweb/reqmgr2ms,g" -e "s,SERVICE,reqmgr2ms,g" -e "s,HOST,${{secrets.IMAGEBOT_URL}},g" imagebot.sh
+#         chmod +x imagebot.sh
+#         cat imagebot.sh
+#         sh ./imagebot.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
       run: |
           #python3 -m pip install -r requirements_py3.txt
           python3 setup.py install_system -s reqmgr2ms --prefix=install
-          tar cfz reqmgr2ms.targz $PWD/install
+          tar cfz reqmgr2ms.tar.gz $PWD/install
 
     - name: Get the Ref
       id: get-ref
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,23 @@
-FROM ubuntu
+FROM registry.cern.ch/cmsweb/cmsweb-base:20211201-stable as cmsweb
+FROM debian:stable-slim
 MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
 
 ENV WDIR=/data
+ENV USER=_dmwm
 WORKDIR $WDIR
 
 # tag to use
-ENV WTAG=1.5.5
 RUN apt-get update && \
     apt-get install -y curl git python3 python3-distutils python3-pip libcurl4-openssl-dev libssl-dev \
-    libjemalloc-dev apache2-utils
+    libjemalloc-dev apache2-utils sudo
 RUN git clone https://github.com/dmwm/WMCore.git
-WORKDIR /data/WMCore
-#RUN python3 setup.py install_system -s reqmgr2ms --prefix=/data/install
+WORKDIR $WDIR/WMCore
+# ENV WTAG=`grep version src/python/WMCore/__init__.py | awk '{print $3}' | sed -e "s,',,g"`
+ENV WTAG=1.5.6
 RUN git checkout tags/$WTAG -b build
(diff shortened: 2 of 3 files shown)


vkuznet:
@amaltaro, @klannon, @todor-ivanov , @goughes  using this CI/CD workflow and its Docker file, I was able to build and run reqmgr2 service on recent Ubuntu Linux OS where I manually applies the following changes:
- add dbs3 dependencies, see https://github.com/dmwm/WMCore/pull/10919
- fix `#!/usr/bin/env python` issue to use python3
- add proper config, static dir, etc. which are required to start `wmc-httpd` daemon

I did not test any DB accesses, or any APIs, and only paid attention to start of `wmc-httpd` daemon. The image I got was about 820MB. Therefore, if you want to proceed with CI/CD we need to resolve aforementioned issue, and we should probably provide test config file to run through CI/CD. So far I suggest to review and merge #10919 , then I can fix `#!/usr/bin/env python` via pattern substitution, and you should separately provide me test config which I can put into this workflow. Then, we'll have fully working CI/CD in place via GitHub Actions.

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ubuntu
+MAINTAINER Valentin Kuznetsov vkuznet@gmail.com
+
+ENV WDIR=/data
+WORKDIR $WDIR
+
+# tag to use
+ENV WTAG=1.5.5
```
amaltaro, line 8: How about automatically extracting the tag from here: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/__init__.py#L9 ?

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+      - '*.*.*'
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ '3.x', 'pypy-3.7' ]
```
amaltaro, line 15: When python `3.x` is defined here, do you know which python version exactly is targeted?

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+on:
+  push:
+    tags:
+      - '*.*.*'
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
amaltaro, line 12: Any chance we could change this build to a Red-Hat flavor OS distribution?

.github/workflows/build.yml
```diff
@@ -0,0 +1,90 @@
...
+        architecture: x64
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: |
+          #python3 -m pip install -r requirements_py3.txt
+          python3 setup.py install_system -s reqmgr2ms --prefix=install
+          tar cfz reqmgr2ms.targz $PWD/install
```
amaltaro, line 31: isn't there a typo here? Where `reqmgr2ms.targz` should be `reqmgr2ms.tar.gz`?

amaltaro (changes requested):
This is great to see, Valentin.

My review can be found along the code, but I also have some other - more generic - questions:
1) should we rename the build.yml file to something like build_reqmgr2ms.yml?
2) can you please create a GH issue following our bug/feature templates and clarifying what is expected to be resolved/implemented? If you feel like multiple GH issues are required to break the work in smaller pieces, please do so.
3) unless it's not possible, we should make sure that automatic builds are made for the same environment that we use to run our services (OS, architecture, dependencies, etc).

Could you also please try to summarize what the final goal you have in mind is? Is it to build all the WMCore services whenever a new tag is created, upload the docker image to gitlab registry, deploy it somewhere, and run a basic check to see whether things are "not breaking"?
Thanks

Dockerfile
vkuznet: ok, I can try, but I need to know if the tag is properly updated in master/main github branch. I can move this part down when I do `git clone` and then rebase to whatever tag will be provided by __init__.py file.

.github/workflows/build.yml
vkuznet: it is latest version available for given distribution

.github/workflows/build.yml
vkuznet: why? really, I need to know why? The WMCore is python based code, as such it can run on anything where Python runs. What are so specific about RedHat that WMCore needs? I find much easier to use Debian based distributions than RedHat.

vkuznet:
Alan, I'll answer your questions as you numbered them:
1. we may have different build yml files and their names and number we can setup later once main workflow will be in place and agreed upon
2. fine, I'll make one
3. what do you mean by the same environment? do you need RPM based procedure? I doubt it is possible since
    * I'm not sure CC7 OS is supported, and I doubt we should use it as it many years old
    * what do you mean by architecture for python based framework, the version of gcc we use? If so, I also doubt it is possible since the version we use dose not come with the OS (neither with CC7 or any other) since it is CMS custom environment. As such the architecture here is only amd64 (x86 on Intel)
    *   Dependencies: my understanding that we should use as much as possible from system stack, and this is what I do in Dockerfile, but for build itself (which only creates Python tar ball it is irrelevant

The overall goal is to fully automate deployment procedure of WMCore servers, i.e. build, tests, creation of release, uploading to PyPi, building the image, deploying the image to cmsweb clusters, release validation. The modern way to work is to use personal laptop (running your favorite OS) to develop code, and only commit/tag your code base. After that CI/CD will take care of building/testing/uploading and deploying all services to cmsweb. This is how I manage now dozen of other services including (DAS, WMArchive, dbs2go, CMSSW popularity, auth proxy server, etc.). It is zero deployment procedure, i.e. I only tag my code and it appears running on k8s cluster.

.github/workflows/build.yml
yuyiguo: Asking the same question. I tried 3.7 with "runs-on: ubuntu-latest" and got error as "error::Version 3.7 with arch x64 not found%0AAvailable versions:%0A%0A".

vkuznet: @yuyiguo what do you mean by saying "I tried 3.7", did you explicitly call this version, did you login to your docker image and verified that it is installed? Did you tried `python3`? There are so many possibilities here. But, if I use `act` to get `ubuntu` image, it provides me with `ghcr.io/catthehacker/ubuntu`, and if I login into it and run python I get this:
```
python
Python 3.8.10 (default, Sep 28 2021, 16:10:42)
```
So, as I wrote before, this is whatever latest 3.X is shipped with whatever image you use.

.github/workflows/build.yml
vkuznet: somehow my previous comment was duplicated and then wiped out. Let me repeat. 

Why do you need RD Linux? What is special about it for WMCore use-case? In other words, does it required to run WMCore services? The choice of Linux OS is not trivial, and I do understand that at CERN we use CC distribution due to many reason (which are irrelevant here). What is relevant does WMCore require specific Linux OS? If so, which one? How do you foresee to move with its upgrades, etc.? In my experience working with Debian based distros is much more pleasant and easier to maintain, then with RD ones. They don't use RPMs, and rely on more flexible `apt` based packaging. Since WMCore services are pure python ones, it should not be matter which Linux OS we need to use for it.

.github/workflows/build.yml
vkuznet: In other words, the concrete python version really depends on image to be in use. The latest images uses the most recent python version, which is always updated when new latest image is available. In other words, if I run today latest image it would be version 3.X1, while in a month from now it may be 3.X2, etc. To have fixed version it is desired to specify concrete image tag.

.github/workflows/build.yml
amaltaro: I think it's extremely important to keep the dev/test environment as similar as to those used to run services in production. Given that both VMs and kubernetes clusters are based on Red Hat distribution (Cent OS / Fedora), it's important to test services with the same OS, when possible.

It's all Linux, yes, but still there are differences of the base packages made available, their versions, and even what is available in each repository (as you could see while working with the requirements_py3.txt file).

.github/workflows/build.yml
amaltaro: This is another example of keeping an environment as similar as the one used in production. Even an apparently minor difference of python 3.7 vs 3.8 version can cause and/or mask issues that we would only spot once the service is deployed in cmsweb.

In addition to that, we should have full control of the environment that we use for building this. "Latest" here can change at any time.

.github/workflows/build.yml
amaltaro: Sorry if I missed anything, but why has this been marked as resolved?

.github/workflows/build.yml
vkuznet: @amaltaro in k8s the nodes are CC, but what you run within them are docker images which can be anything. My question remains, what exactly WMCore services users from RedHat/CC/Fedora? All dependencies are python packages, and unless you give me specific example, I don't see any requirements to run service on specific Linux OS. For instance, I can run WMCore service on macOS and the only question is if service requires access to DB which require DB specific libs. As such what matters if ORACLE provides the libs for given Linux OS.

amaltaro: I could be wrong, but it looks like most of our docker images have libraries other than python packages. Here is just one example:
https://github.com/dmwm/CMSKubernetes/blob/master/docker/cmsweb-base/Dockerfile#L8-L11
but cmsweb-base image also has plenty of non-python packages installed.

.github/workflows/build.yml
yuyiguo: @vkuznet 
jobs:
  build-n-publish:
    name: Build
    runs-on: ubuntu-latest
    steps:
    ...
    - name: Set up Python 3.x
      uses: actions/setup-python@v2
      with:
        python-version: 3.x
Now I got python 3.10 installed in my docker. My docker is running on my mac. but you have 3.8. So these are changing with different docker/config. Are there a place one can check for python version under the version of ubuntu , actions/setup-python and docker combination? I will the discussion to DBSClient CI/CD issues next.

vkuznet: @yuyiguo I do not control images provided by upstream. I think everything new xxxx-latest image is created it has up to date software packages. You can use docker command itself and pull out any image, then you can login into image and check which python version it has.

.github/workflows/build.yml
vkuznet: @amaltaro , as I stated already, WMCore is python based application, as such in my view it does not and should not depend on OS. Docker allows multi-stage builds where I can borrow some stuff from one image and put it into another. As such I can use certs from cmsweb and copy them into my final image.

vkuznet: @klannon I totally agree, but if we do I rather prefer to have full meeting slot dedicated to this discussion, and not last X minutes. There are so many technical details that it is worth time to discuss them properly. I suggest to start meeting with this topic and if time left go to the usual business with news, etc.

amaltaro (commented):
As discussed along the code, we are going to get back to this and have a general discussion on what exactly is expected out of these CI/CD setup, what the requirements are, what is the target environment, etc. 

We can agree on a time slot in our first team meeting in 2022, maybe even a special slot since there is a lot to be discussed.

amaltaro:
@vkuznet Valentin, this work started back in December 2021 and I believe it has already been covered by the latest developments from Q4/2022 and Q1/2023. Shall we close this PR?

vkuznet:
yes, I'll close this issue

PR #10918, opened 2021-12-14
[py3] CredentialProxy - convert to bytes the input of hashlib.sha1()

### LINKED ISSUE #10912: Credential/Proxy is not fully py3 compatible
TypeError: Unicode-objects must be encoded before hashing
in

https://github.com/dmwm/WMCore/blob/15067b9916fbd01b6599d5329aa6825940654bf6/src/python/WMCore/Credential/Proxy.py#L263-L264

mapellidario:
Fixes #10912 

#### Status

Ready

Tested on CRABServer `test1` instance:
- The task [211214_133519:dmapelli_crab_20211214_143516](https://cmsweb-test1.cern.ch/crabserver/ui/task/211214_133519%3Admapelli_crab_20211214_143516) finished successfully

#### Description

We need to make sure that the input of `hashlib.sha1()` is always bytes, in both py2 and py3.

This PR is based on: https://github.com/belforte/WMCore/commit/1be8e3c564f39f67eaa948a758a82b0b734811c8

fyi: @belforte 

yes

#### Related PRs
none

none

### DIFF
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -19,7 +19,7 @@ from hashlib import sha1
 from WMCore.Credential.Credential import Credential
 from WMCore.WMException import WMException
 from Utils.PythonVersion import PY3
-from Utils.Utilities import decodeBytesToUnicode
+from Utils.Utilities import decodeBytesToUnicode, encodeUnicodeToBytes
 
 
 def execute_command(command, logger, timeout, redirect=True):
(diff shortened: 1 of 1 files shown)


belforte:
about testing, please be aware that this code is only used by CRAB TaskWorker and (from now on) only in python3. WCredetial/Proxy is effectively made of two parts, one used by CRAB, one by WMA & C. even if they are under same umbreall interface, there is almost no method shared. I.e. the testing we did already in CRAB is good, IMHO. And of course we'll do more.

amaltaro:
Does it mean it can be merged? Or you'd rather run more tests first?
Would you also need a new tag? Does it need to be backported to any given branch?

mapellidario:
I run some tests and they are looking good. We do not need any backporting, this can stay in `dmwm/WMCore/master`.

> so were thinking of waiting for all of those to be fully sorted out before messing with your already very busy schedule.

yes, we temporarily switch to using our own private branch of WMCore for building python3 crabserver images exactly for this purpose. While we are finalizing the migration, we benefit from using `dmwm/WMCore/master` branch and applying our own logging and fixes on top. We will take care of how we build crab, no new action items on Alan and the rest of wmcore team!

> assuming it is same code as I already used in my branch (Dario?)

Not really. I used `encodeUnicodeToBytes` instead, to be python2 compatible (which i assume is still a requirement for WMCore, even if we are the only users of some parts of its code), and i changed the input of `hashlib.sha()` in every place where it is used, not only the couple of places where you added `.encode()`

amaltaro (approved):
Looks good to me! Thanks Dario and Stefano!

PR #10923, opened 2021-12-16
Add getStatsFromMongoDB to MSUnmerged.

### LINKED ISSUE #10417: MSUnmerged: expose basic cleanup stats via REST API
**Impact of the new feature**
MSUnmerged

It concerns a new microservice to carry out further cleanup of the sites' unmerged area (removal of files that are no longer needed/locked in the system).

Since we are now having the remote deletions through `gfal` implemented,  we would like to have some statistics exposed per  RSE about the cleaning process. We are currently bound to the pooling cycle of the RucioConmon service, which is once per week, so the information and results of the consecutive reruns of our MSUnmerged service would be accumulated through out this period in the background database (MongoDB) and displayed  through appropriate REST APIs developed to the service. 

In order to retrieve data for a given RSE, we need to call the `info` REST API and provide the RSE name via `rse=RSE_name` query string. If `detail=true` is provided, then it also returns directory related information (what is protected, what is deleted, etc). Detailed information is disabled by default.

Create the desired Methods and APIs to translate the supported REST calls to the correct MongoDB queries and display the information through the `info` and `status` interfaces to the service. 

**Additional context**
Depends on https://github.com/dmwm/WMCore/issues/10411

todor-ivanov:
Fixes #10417 
#### Status
Ready

#### Description
With the current PR we are providing some additional interfaces to the service REST APIs in order to  expose aggregated data from MongoDB per RSE or per RSE Type. E.g. 

https://cmsweb-testbed.cern.ch/ms-unmerged/data/info?rse=T2_DE_DESY

or

https://cmsweb-testbed.cern.ch/ms-unmerged/data/info?rse_type=t1

Here follows a simple example of the output (the long lists in the dictionary are trimmed): 

```
"result": [
 {
  "wmcore_version": "1.5.6.pre1",
  "microservice_version": "1.5.6.pre1",
  "microservice": "MSManager",
  "query": "rse=T2_US_Wisconsin",
  "rseData": [{
    "name": "T2_US_Wisconsin",
    "pfnPrefix": "gsiftp://cms-lvs-gridftp.hep.wisc.edu:2811/hdfs/store/test/rucio/int",
    "isClean": false,
    "timestamps": {
      "rseConsStatTime": 0.0,
      "prevStartTime": 1639732931.3478053,
      "startTime": 1639734096.9867024,
      "prevEndTime": 1639732936.3460612,
      "endTime": 1639734102.1553724
    },
    "counters": {
      "totalNumFiles": 780538,
      "dirsToDeleteAll": 731,
      "dirsToDelete": 16082,
      "filesToDelete": 780538,
      "deletedSuccess": 0,
      "deletedFail": 0
    },
    "dirs": {
      "allUnmerged": [],
      "toDelete": [
        "/store/unmerged/RunIISummer20UL16wmLHEGENAPV/GJets_DoubleEMEnriched_PtG-20MGG-40To80_TuneCP5_13TeV-madgraphMLM-pythia8/GEN/106X_mcRun2_asymptotic_preVFP_v8-v1",
        "/store/unmerged/RunIISummer20UL16MiniAODAPVv2/WWTo4Q_4f_TuneCP5_13TeV-amcatnloFXFX-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v2",
        "/store/unmerged/RunIISummer20UL16MiniAOD/HbToJPsiMuMu_3MuFilter_TuneCP5_13TeV-pythia8-evtgen/MINIAODSIM/106X_mcRun2_asymptotic_v13-v1",
        "/store/unmerged/RunIISummer20UL16RECOAPV/HbToJPsiMuMu_TuneCP5_13TeV-pythia8-evtgen/AODSIM/106X_mcRun2_asymptotic_preVFP_v8-v1",
        "/store/unmerged/RunIISummer20UL17MiniAODv2/TTToHminusToCS_M-155_TuneCP5_13TeV-amcatnlo-pythia8/MINIAODSIM/106X_mc2017_realistic_v9-v2",
        "/store/unmerged/RunIISummer20UL16MiniAODAPVv2/WJetsToLNu_0J_TuneCP5_13TeV-amcatnloFXFX-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_preVFP_v11-v1",
        ...
              ],
      "protected": []
    }]
  }
}]}
```

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/MicroService/MSManager.py
+++ b/src/python/WMCore/MicroService/MSManager.py
@@ -32,7 +32,7 @@ from datetime import datetime
 # WMCore modules
 from WMCore.MicroService.Tools.Common import getMSLogger
 from WMCore.MicroService.TaskManager import start_new_thread
-
+from Utils.Utilities import strToBool
 
 def daemon(func, reqStatus, interval, logger):
     "Daemon to perform given function action for all request in our store"
@@ -295,8 +295,7 @@ class MSManager(object):
         :param reqName: request name
         :param kwargs: other arguments that can be provided for different
             microservices. Examples:
-            rse_type: string with the type of the RSEs to fetch (MSUnmerged)
-                      Meaningful values at the moment are: "t1", "t2t3" and "t2t3us"
+            rse: string with the name of the RSE (e.g 'T2_DE_DESY')
         :return: data transfer information for this request
         """
         if 'ruleCleaner' in self.services or 'unmerged' in self.services:
@@ -317,8 +316,10 @@ class MSManager(object):
                 data['transferDoc'] = transferDoc[0]
 
         if 'unmerged' in self.services:
-            if kwargs.get("rse_type"):
-                data = {"rse_type": kwargs.get("rse_type")}
+            detail = strToBool(kwargs.get('detail', False))
+
+            if kwargs.get("rse"):
+                data = self.msUnmerged.getStatsFromMongoDB(detail=detail, rse=kwargs['rse'])
         return data
 
     def delete(self, request):
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
@@ -771,6 +771,34 @@ class MSUnmerged(MSCore):
             raise MSUnmergedPlineExit(msg)
         return rse
 
+    def getStatsFromMongoDB(self, detail=False, **kwargs):
+        """
+        Auxiliary method to serve the APIs of the REST calls for the service.
+        Implements various queries to MongoDB based on the parameters passed.
+        :param detail:           Bool marking if additional details must be queried from the database
+                                 (e.g. putting the 'dirs' field into the mongo projection)
+        :param rse=rseName:      String representing the RSE name to query the database for
+        :return:                 Dictionary with the type of query + the data returned from the database
+        """
+        data = {}
+        if kwargs.get('rse'):
+            data["query"] = 'rse=%s&detail=%s' % (kwargs['rse'], detail)
+            mongoProjection = {"_id": False,
+                               "name": True,
+                               "pfnPrefix": True,
+                               "isClean": True,
+                               "timestamps": True,
+                               "counters": True}
+            if detail:
+                mongoProjection["dirs"]  = True
+
+            rseList = kwargs['rse'] if isinstance(kwargs['rse'], list) else [kwargs['rse']]
+            data["rseData"] = []
+            for rseName in rseList:
+                mongoFilter = {'name': rseName}
+                data["rseData"].append(self.msUnmergedColl.find_one(mongoFilter, projection=mongoProjection))
+        return data
+
     # @profile
     def getRSEList(self):
         """
--- a/src/python/WMCore/MicroService/Service/Data.py
+++ b/src/python/WMCore/MicroService/Service/Data.py
@@ -32,7 +32,7 @@ from Utils.Patterns import Singleton
 from WMCore.REST.Server import RESTEntity, restcall
 from WMCore.REST.Tools import tools
 # from WMCore.REST.Validation import validate_rx, validate_str
-from WMCore.REST.Format import JSONFormat
+from WMCore.REST.Format import JSONFormat, PrettyJSONFormat
 
 
 def results(res):
@@ -79,7 +79,7 @@ class Data(with_metaclass(Singleton, RESTEntity, object)):
                 return False
         return True
 
-    @restcall(formats=[('application/json', JSONFormat())])
+    @restcall(formats=[('text/plain', PrettyJSONFormat()), ('application/json', JSONFormat())])
     @tools.expires(secs=-1)
     def get(self, **kwds):
         """


amaltaro (commented):
Todor, this review was done here and on slack discussions in parallel, so some comments might no longer be relevant, but here they are:
* we do not have any aggregation mechanism at the moment. The `rse_type` is only meant to target a different replica/POD in cmsweb. So I would suggest to remove any mention to it to avoid confusions. If you have a use case for aggregated data, I would suggest to track that in a new GH issue.
* I think it would be extremely important to fetch only a summary of RSE stats. Having said that, I think the default `info` query should not return the `dirs` data structure. While if the client provides `&detail=true`, it would then return the same thing as detail=false plus the `dirs` data structure. I left a comment along the code on this as well.

Regarding this `detail` query string, other microservices use the same mechanism. Please have a look at the wiki describing it to keep things consistent.

amaltaro:
@todor-ivanov please also update the MSUnmerged REST documentation here (where relevant): https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-APIs#msunmerged-rest-apis

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -771,6 +771,30 @@ def uploadRSEToMongoDB(self, rse, fullRSEToDB=False, overwrite=True):
...
+        data = {}
+        if kwargs.get('rse'):
+            data["query"] = 'rse=%s' % kwargs['rse']
+            mongoProjection = {"_id": False,
+                               "name": True,
+                               "pfnPrefix": True,
+                               "isClean": True,
+                               "timestamps": True,
+                               "counters": True,
+                               "dirs": True}
```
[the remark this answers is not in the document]
todor-ivanov: Agreed.
Working on that

todor-ivanov:
Thanks for the review @amaltaro Please take a look at the latest  commit it addresses your all comments I think.

src/python/WMCore/MicroService/MSManager.py
```diff
@@ -318,10 +319,15 @@ def info(self, reqName=None, **kwargs):
                 data['transferDoc'] = transferDoc[0]
 
         if 'unmerged' in self.services:
+            if kwargs.get("detail"):
```
amaltaro, line 322: I would suggest importing one of our own utilities from (instead of distutil):
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/Utilities.py#L64

So the code could be something like:
```
detail = strToBool(kwargs.get('detail', False))
```

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -771,24 +771,25 @@ def uploadRSEToMongoDB(self, rse, fullRSEToDB=False, overwrite=True):
             raise MSUnmergedPlineExit(msg)
         return rse
 
-    def getStatsFromMongoDB(self, **kwargs):
+    def getStatsFromMongoDB(self, detail=False, **kwargs):
         """
         Auxiliary method to serve the APIs of the REST calls for the service.
         Implements various queries to MongoDB based on the parameters passed.
         :param rse=rseName:      String representing the RSE name to query the database for
-        :param rse_type=rseType: String representing the RSE type to query the database for
```
amaltaro: and `detail` here could be part of the kwargs as well. 
Regardless of what you select to implement here, please update the docstring.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -771,24 +771,25 @@ def uploadRSEToMongoDB(self, rse, fullRSEToDB=False, overwrite=True):
...
         data = {}
         if kwargs.get('rse'):
-            data["query"] = 'rse=%s' % kwargs['rse']
+            data["query"] = 'rse=%s&detail=%s' % (kwargs['rse'], detail)
             mongoProjection = {"_id": False,
                                "name": True,
                                "pfnPrefix": True,
                                "isClean": True,
                                "timestamps": True,
-                               "counters": True,
```
amaltaro: shouldn't we explicitly set the projection to `False` if no details are required? Maybe could be accomplished through:
```
....
"dirs": detail,
...
```

amaltaro (changes requested):
Thanks for the quick update. Please find some comments inline.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Sorry @amaltaro  this is not how MongoDB works unfortunately. You can imagine my dissapointment when I found out that the only field of the document allowd to be set as `False` in the projection is `_id` otherwise you get the following exception [1] once you try to use the projection.  And speaking of which there is one more place where this needs to be fixed [2]. Which was actually the reason to not being able to use the mongoProjection to its full power when I was testing it back then. 

[2]
https://github.com/dmwm/WMCore/blob/39e1ecdfd581c497aed0d00de873d66d7ee448e4/src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py#L80


[1]
```
In [87]: mongoProjection
Out[87]: 
{'_id': False,
 'name': True,
 'pfnPrefix': True,
 'isClean': True,
 'timestamps': True,
 'counters': True,
 'dirs': False}

In [88]: msUnmergedColl.find_one({'name': rse['name']}, mongoProjection)
...
OperationFailure: Projection cannot have a mix of inclusion and exclusion.
```

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: I'd prefer to keep it with a default value here, so we protect ourselves of unneeded huge dumps in the browser of enormous RSE's like CERN or FNAL - a mistake I have already done several times myself while testing it. 
I'll update the docstring.

todor-ivanov:
Hi @amaltaro I have addressed the set of your remarks. And I also fixed a behavior which I was not very pleased of earlier - and that is not taking into account if there are more than one RSEs present into the REST call. Now this is fixed too, so queries of the sort: https://tivanov-unit02.cern.ch/ms-unmerged/data/info?rse=T2_US_Wisconsin&rse=T2_CH_CERN must produce the following result: 

```
{"result": [
 {
  "wmcore_version": "1.5.6.pre1",
  "microservice_version": "1.5.6.pre1",
  "microservice": "MSManager",
  "query": "rse=['T2_US_Wisconsin', 'T2_CH_CERN']&detail=False",
  "rseData": [
    {
      "name": "T2_US_Wisconsin",
      "pfnPrefix": "gsiftp://cms-lvs-gridftp.hep.wisc.edu:2811/hdfs/store/test/rucio/int",
      "isClean": false,
      "timestamps": {
        "rseConsStatTime": 0.0,
        "prevStartTime": 1639804256.731806,
        "startTime": 1639805118.2272134,
        "prevEndTime": 1639804261.4009178,
        "endTime": 1639805122.9460645
      },
      "counters": {
        "totalNumFiles": 206108,
        "dirsToDeleteAll": 845,
        "dirsToDelete": 5915,
        "filesToDelete": 206108,
        "deletedSuccess": 0,
        "deletedFail": 0
      }
    },
    {
      "name": "T2_CH_CERN",
      "pfnPrefix": "gsiftp://eoscmsftp.cern.ch:2811/eos/cms/store/test/rucio/int",
      "isClean": false,
      "timestamps": {
        "rseConsStatTime": 0.0,
        "prevStartTime": 1639698610.379427,
        "startTime": 1639702833.4908555,
        "prevEndTime": 1639698887.8020303,
        "endTime": 1639703109.4869418
      },
      "counters": {
        "totalNumFiles": 4318625,
        "dirsToDeleteAll": 1869,
        "dirsToDelete": 9345,
        "filesToDelete": 4319280,
        "deletedSuccess": 0,
        "deletedFail": 0
      }
    }
  ]
}]}
```

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -771,6 +771,36 @@ def uploadRSEToMongoDB(self, rse, fullRSEToDB=False, overwrite=True):
...
+                               "counters": True}
+            if detail:
+                mongoProjection["dirs"]  = True
+
+            rseList = kwargs['rse'] if isinstance(kwargs['rse'], list) else [kwargs['rse']]
+            data["rseData"] = []
+            for rseName in rseList:
+                mongoFilter = {'name': rseName}
+                data["rseData"].append(self.msUnmergedColl.find_one(mongoFilter, projection=mongoProjection))
+        if kwargs.get('rse_type'):
```
amaltaro, line 800: If we have a call like: `rse=Alan&rse_type=Blah&detail=False`
it would result in:
```
data["query"] = "rse_type=Blah"
```
dropping the previous content, while what we should do is actually to concatenate them through a `&` char.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -771,6 +771,36 @@ def uploadRSEToMongoDB(self, rse, fullRSEToDB=False, overwrite=True):
             raise MSUnmergedPlineExit(msg)
         return rse
 
+    def getStatsFromMongoDB(self, detail=False, **kwargs):
+        """
+        Auxiliary method to serve the APIs of the REST calls for the service.
+        Implements various queries to MongoDB based on the parameters passed.
+        :param detail:           Bool marking if additional details must be queried from the database
+                                 (e.g. putting the 'dirs' field into the mongo projection)
+        :param rse=rseName:      String representing the RSE name to query the database for
```
amaltaro, line 780: I dislike making another comment on this docstring, but it's still not consistent with the source code. Please make your best to provide correct docstrings for code freshly implemented.

src/python/WMCore/MicroService/MSManager.py
```diff
@@ -317,8 +318,12 @@ def info(self, reqName=None, **kwargs):
                 data['transferDoc'] = transferDoc[0]
 
         if 'unmerged' in self.services:
+            detail = strToBool(kwargs.get('detail', False))
+
             if kwargs.get("rse_type"):
```
amaltaro, line 323: Since there are other comments in the PR, let me make the following as well.
It just ocurred to me that we could replace these 4 lines here by:
```
data = self.msUnmerged.getStatsFromMongoDB(detail=detail, rse_type=kwargs.get('rse_type'), rse=kwargs.get('rse'))
```

amaltaro (changes requested):
It's looking good to me, Todor. However, I think there is still at least one minor fix that need to be made concerning the query information. The others would be good to change as well, but they don't change any functionality or behavior.

src/python/WMCore/MicroService/MSManager.py
todor-ivanov: And this changes the logic completely, by making all queries passed at once and hence no separation between the interfaces is achieved. Which was exactly the goal in the first place when I was calling the function to make the queries to the database once per API. This way we can maintain a decent set of well separated calls to the databases providing well defined results. While the other way around one would need to invent a really complicate call to the database to satisfy a REST call of the type: `rse=T2_CH_CERN&rse_type=t2&rse=T3_US_NERSC&detail=true`

BTW I already tried to mix the two calls `rse_type=...&rse=...` and actually the best result achieved was the values returned by the last query to the database to overwrite the others. So long sotrry short, mixing the APIs in a single calls from the top level method would complicate the call to the database itself quite a lot and would prevent us from keeping short and well understood filters, as it is rigt now.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: I do not get this comment.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: Sorry @amaltaro I tend to disagree here. Those two interfaces simply cannot live in a single call to thee database (or at least not in a simple one). Those are serving completely different purposes. So mixture of the sort simply makes things unneeded over complicated. The rest call `rse_type=....` would result (eventually in the future) in an aggregation across all the RSEs of the given type through out the whole database and providing results based on completely different projection than the one for the `rse=....` call. On the other hand the call `rse=...` is simply fetching a single document (eventually could be repeated multiple times and the result displayed in a list) and displaying it according to the projection we already discussed above. So this mixture of calls if not completely impossible in theory is at least not going to produce results of similar type and shape, so I'd not support these two calls together.  I'd rather instruct the users that  the better way to fetch those tow difernt types of info simply by making two separate REST calls to the relevant API.

todor-ivanov: BTW this:
> If we have a call like: rse=Alan&rse_type=Blah&detail=False
> it would result in:
>
> data["query"] = "rse_type=Blah"

Is completely true, and I kind of like this behavior more (which I already explained why). And having the interfaces  work separately was kind of a desired effect rather than a defect.

amaltaro:
Todor, I think we need to take a step back on this development and first properly describe what is the expected product out of the issue being worked on here. The original description dates from April, when we had a different MSUnmerged design in mind. For future issues, I'd suggest to first review the issue description and describe and/or update anything that is not clear before actually starting the implementation, this will make development and review easier.

Now talking about the expected outcome, from my point of view:
* implement a REST API where we can retrieve some basic statistics concerning a given RSE
  * it's been implemented through the `info` API with a query string `rse=my_rse_name`
  * it supports the `detail=boolean` query string, to fetch directory-related information as well
  * and I see you added support to multiple RSEs in the same call, which is a plus
* I would suggest to not expose any internal queries back to the client (so your `query` key, could either be dropped, or you simply populate it with the REST query strings (as mentioned in one of my comments)
* the `rse_type` query string is, right now, only a mechanism to be able to reach a specific ms-unmerged replica in the kubernetes setup (which actually only work in testbed, not in production due to cluster architecture). So I would suggest to:
  * either drop it completely
  * or leave it for future development and ideas. I do not think we need to make placeholders though. If we come up with a specific use case that would add value to it, we make a new feature request and implement it.

What do you think? Please let me know if I missed anything. Can you please also replicate this to the GH issue itself (update the original description)?

todor-ivanov:
Hi @amaltaro Thanks for the feedback:

> implement a REST API where we can retrieve some basic statistics concerning a given RSE

>   it's been implemented through the info API with a query string rse=my_rse_name
>   it supports the detail=boolean query string, to fetch directory-related information as well
>   and I see you added support to multiple RSEs in the same call, which is a plus

This is all done.

> I would suggest to not expose any internal queries back to the client (so your query key, could either be dropped, or you simply populate it with the REST query strings (as mentioned in one of my comments)

This is exactly the case already. It visualizes only the REST call for which the result is about to be displayed. It has nothing to do with the MongoDB query. 

 > either drop it completely

And this is now done in my latest commit. Please take a look once again. Thanks in advance!

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
amaltaro: Now that rse_type is no longer used in the code, the docstring is consistency. Thanks

amaltaro (approved):
Thanks Todor. It's looking good to me now.
Before we merge it, can you please:
* squash commits accordingly
* update the GH issue, as requested in my previous comment
* update the REST API wiki, if needed.
?

todor-ivanov:
Hi @amaltaro Please find the squashed commits and the updated  giithub issue reflecting the current state.  Thanks!

amaltaro (approved):
Thanks Todor. It all looks good now!

PR #10928, opened 2021-12-16
Fix logic which drops input blocks to be transferred

### LINKED ISSUE #10927: Test json template TC_PY3_Data_LumiList.json triggering 126TB of input data placement
**Impact of the bug**
WMAgent tests

**Describe the bug**
I just injected the full battery of DMWM and Integration tests into cmsweb-testbed - running WMCore 1.5.6.pre1 with dbs3-client 4.0.4 - and a MSTransferor alert has been triggered [1]. I'm going to delete those rucio rules first, then I will start investigating why such LumiList template is considering so many input blocks for the data placement.

**How to reproduce it**
Inject and assign this template in testbed: https://github.com/dmwm/WMCore/blob/master/test/data/ReqMgr/requests/Integration/TC_PY3_Data_LumiList.json

**Expected behavior**
That template provides a single run via a `LumiList` parameter and I doubt that 95 input blocks would match it. Something must have gotten broken with the latest MSTransferor release, which could well be the new dbs3-client as well. Further investigation is required though...

[1]
```
description = Workflow: amaltaro_TC_PY3_Data_LumiList_Dec2021_Val_211216_193826_9374 has a large amount of data subscribed: 126.232069696652 TB, for primary data: /DoubleMuon/Run2018D-v1/RAW.
hostname = ms-transferor-6745f5c54b-48dqq
summary = [MS] Large pending data transfer under request id: {'c4489f11c18349169ad39df96de9f762', '6ca3f4db2cfa4c84b55b3aa9d8f97aea', 'b6bcaace0c9942c6a6e2356ec5c275fe', '4e07236eda394a1db97d0583a0c3a251', '41c2eeef2b24475ca4d64051ab769dbd', ....
```

amaltaro:
Fixes #10927 

#### Status
ready

#### Description
If we have a workflow, with either run or lumi list, and we happen to drop all the blocks from the initial input list, the initial list of blocks is kept and used for input data placement.

With this patch, if a workflow drops all the previously matched input blocks, it will pass an empty dictionary upstream which should not trigger any input data placement.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/MicroService/MSTransferor/RequestInfo.py
+++ b/src/python/WMCore/MicroService/MSTransferor/RequestInfo.py
@@ -341,26 +341,30 @@ class RequestInfo(MSCore):
         """
         finalBlocks = {}
         dbsUrl = wflow.getDbsUrl()
-        runWhite = wflow.getRunWhitelist()
-        runBlack = set(wflow.getRunBlacklist())
+        runAllowedList = wflow.getRunWhitelist()
+        runForbiddenList = set(wflow.getRunBlacklist())
         lumiList = wflow.getLumilist()
+        # if there is no filter on the input data, simply return it
+        if not (lumiList or runAllowedList or runForbiddenList):
+            return self._removeZeroSizeBlocks(blocksDict)
+
         if lumiList:
             # LumiList has precedence over RunWhitelist
-            runWhite = []
+            runAllowedList = []
             for run in lumiList.getRuns():
-                runWhite.append(int(run))
-            runWhite = list(set(runWhite))
-        if runWhite:
+                runAllowedList.append(int(run))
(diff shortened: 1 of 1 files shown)


amaltaro:
test this please

todor-ivanov (approved):
Thanks for those changes @amaltaro  They look OK to me.  I have only one minor comment, which is not a show stopper here. It is about the new naming of the variables. I would rather call them: `runAllowedList` and `runForbiddenList` so that while reading later  in the code one does not get the wrong impression it is a single valued item. The `List` suffix still marks the plural in the variable's nature :) But both ways would do I think. As you find it better.

amaltaro:
@todor-ivanov thanks for the review, Todor. Indeed renaming those variables will be better. I have just updated the code.

PR #10930, opened 2021-12-17
Remove trailing slash from DBS URL

### LINKED ISSUE #10929: Sanitize DbsUrl and remove trailing slash character
**Impact of the new feature**
Any WMCore service not relying on DbsApi client

As we migrated to the Go-based DBSServer in testbed, a new issue has been exposed where DBS calls like
```
/dbs/int/global/DBSReader//serverinfo
```
not longer work because that resource is now in a new location, e.g.:
```
/dbs/int/global/DBSReader/serverinfo
```

That means, we need to make sure that the correct url is constructed in WMCore.

In order to avoid url constructions with double slash (see example above), we should sanitize the DbsUrl and strip an ending slash after DBSReader or DBSWriter.

First place is in the WMSpec factory, when validating and accepting workflow arguments:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMSpec/StdSpecs/StdBase.py#L1032

Second place is in the MicroServices, when using the pycurl module or anything different than DBS3Reader or DbsApi

Third place is in global/local workqueue (since it might be processing old workflows with possibly a trailing slash).

And any other place where DBSServer calls are made.

To enable follow redirect in the HTTP python library that we are using, however, it would double our HTTP calls and would not work for HTTP methods different than GET.

**Additional context**
Similar to this: https://github.com/dmwm/DBSClient/issues/55

amaltaro:
Fixes #10929 

#### Status
not-tested

#### Description
In addition to removing the trailing `/` from any explicit DBS url in WMCore, this patch will make sure that the correct url (without trailing slash) gets persisted in the system/CouchDB.

YES

#### Related PRs
Related issue/PRs in the DBSClient repository as well: https://github.com/dmwm/DBSClient/issues/55

None


amaltaro:
test this please

amaltaro:
test this please

todor-ivanov (approved):
These look good to me @amaltaro 

Thanks for finding and fixing them all!

amaltaro:
Thanks for the review, Todor. Tests look good at ReqMgr2 and WorkQueue level.

PR #10937, opened 2021-12-22
Improve WMCore framework metrics

### LINKED ISSUE #10936: Improve WMCore framework metrics
**Impact of the new feature**
To setup better monitoring of WMCore service process it is desired to improve its metrics. This may include but not limited to the following items:
- total HTTP request time
- time spend on DB
- process metrics like cpu/mem, etc.

It will provide better way to monitor WMCore processes

Improve metrics in WMCore logging and feed them to monitoring infrastructure

**Additional context**

vkuznet:
Fixes https://github.com/dmwm/WMCore/issues/10936

#### Status
in development

#### Description
Improve process metrics

YES

it relies on [psuti](https://pypi.org/project/psutil/)l 3rd party library

### DIFF
--- a/src/python/Utils/ProcessStats.py
+++ b/src/python/Utils/ProcessStats.py
@@ -42,8 +42,41 @@ import traceback
 
 try:
     import psutil
+    PSUTIL = True
 except ImportError:
+    PSUTIL = False
     pass
+
+class ProcessInfo(object):
+    def __init__(self):
+        self.psutil = PSUTIL
+        self.process = psutil.Process(os.getpid())
+        self.init_rss = self.rss()
+        self.init_vms = self.vms()
+
+    def rss(self):
+        "Provide process memory value"
+        return self.process.memory_info().rss if self.psutil else 0
+
+    def cpu(self):
+        "Provide process cpu value"
+        return self.process.cpu_percent(interval=None) if self.psutil else 0
+
+    def vms(self):
+        "Provide process virtual memory value"
+        return self.process.memory_info().vms if self.psutil else 0
+
+    def connections(self):
+        "Provide process conenctions"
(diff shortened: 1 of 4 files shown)


vkuznet:
@amaltaro I re-used existing structure of Utils.ProcessStats module and add additional class to report psutil metrics. Then, I plug it in into REST WMCore frameworks and report them in logging entries. The output of look like this:
```
[22/Dec/2021:09:48:23] vkair 127.0.0.1 "GET /reqmgr2/data/info HTTP/1.1" 200 OK [data: 93 in 1115 out 2345169 us ] [auth: - "" "" ] [ref: "" "curl/7.80.0" ] [stats: cpu=0.0 rss=34807808 vms=4489146368 inc_rss=8261632 inc_vms=42156032 connections tot=2 lis=1 est=1]
```
where you can see an additional stats part reporting current usage of process metrics. Any suggestions are welcome. So far I provided reporting in logging and next step would be improve ProcessStats to report these and other metrics to MONIT (via Prometheus scraping).

amaltaro (changes requested):
This has been implemented as part of the request access_log, thus for every single HTTP call. 
In my opinion this is an overkill and not how it should be deployed in production. This will definitely affect the performance of every HTTP request as well, given that process stats will be retrieved for every single thread of the CherryPy thread pool.

Instead, we should use the `stats` endpoint and poll it at a meaningful interval.

vkuznet:
Yes, you're right, and we have both `stats` and `metrics` (from https://github.com/dmwm/WMCore/pull/9940 if we merge). So, I think we can drop this PR once #9940 will be in place.

amaltaro:
After merging https://github.com/dmwm/WMCore/pull/9940, this is no longer relevant. Closing.

PR #10938, opened 2021-12-22
Check that sdir exists on a file system

### LINKED ISSUE #10941: Check existence of directory location while accessing it
**Impact of the bug**
While listing files in directory the provided directory should exists, otherwise python will through excception.

**Describe the bug**
while accessing`os.listdir(sdir)` the `sdir` should exist

**How to reproduce it**
```
>>> import os
>>> os.listdir("ljkljdflkg")
OSError: [Errno 2] No such file or directory: 'ljkljdflkg'
```

**Expected behavior**
Check sdir before accessing it with `os.listdir` function

vkuznet:
Fixes https://github.com/dmwm/WMCore/issues/10941

#### Status
ready

#### Description
during testing reqmgr service within local environment I found that code tries to access non-existing area. To protect it from failure I patched the code to check that such area exists.

YES

### DIFF
--- a/src/python/WMCore/ReqMgr/Web/ReqMgrService.py
+++ b/src/python/WMCore/ReqMgr/Web/ReqMgrService.py
@@ -413,9 +413,10 @@ class ReqMgrService(TemplatedPage):
     def update_scripts(self, force=False):
         "Update scripts dict"
         if force or abs(time.time() - self.sdict['ts']) > self.sdict_thr:
-            for item in os.listdir(self.sdir):
-                with open(os.path.join(self.sdir, item), 'r') as istream:
-                    self.sdict[item.split('.')[0]] = istream.read()
+            if os.path.isdir(self.sdir):
+                for item in os.listdir(self.sdir):
+                    with open(os.path.join(self.sdir, item), 'r') as istream:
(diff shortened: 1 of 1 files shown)


amaltaro:
@vkuznet as mentioned in the contribution guidelines:
https://github.com/dmwm/WMCore/blob/master/CONTRIBUTING.rst
please create a GH issue to report and track this problem.

vkuznet:
@amaltaro , I'll be happy to do it if instructions will be clarified, e.g. according to CONTRIUTIONS.rst, it should be an issue which can be in form of patch or feature. But, if I click on New issue it provides "bug report" (which I choose for this PR) or "New feature. Please adjust your documentation accordingly to options you provide in New issue. I think I did a correct thing by choosing bug report in this particular case.

vkuznet:
@amaltaro I figure out that what you mention as patch is called bug report. So, I created bug report and adjusted this PR description. Sorry for confusion on my side.

amaltaro:
test this please

amaltaro (approved):
I think we could be more informative on those directories and when there is a problem with them, but given that this is barely used (I guess these scripts are actually not used at all), it's good enough like this. Thanks Valentin.

PR #10939, opened 2021-12-22
Add support for IAM token

### LINKED ISSUE #10940: Add support for TOKEN based authentication
**Impact of the new feature**
We should provide support for token based authentication with external services. We gradually move all cmsweb service to new authentication mechanism and x509 authentication will be phase out.

The x509 authentication will be dropped in a near by future and all clients should adopt token based authentication.

adjust HTTP request handlers and any other pieces within WMCore services to support TOKEN based authentication;

None
**Additional context**

vkuznet:
Fixes https://github.com/dmwm/WMCore/issues/10940

This PR provides support to include IAM token to HTTP requests. The token can be obtained via separate process, e.g. via curl, `create-iam-token.sh` script or [oidc-agent](https://indigo-iam.github.io/v/v1.7.0/docs/tasks/user/getting-a-token/), and then can be used by any HTTP client. We added appropriate `Authorization` HTTP header to `pycurl_manager`. The token itself can be provided either via file name or environment variable.

For more information please consult [Tokens in WMCore](https://github.com/dmwm/WMCore/wiki/Tokens-in-WMCore) wiki page.

#### Status
ready

#### Description
We provide a patch to pycurl manager to setup appropriate HTTP header if token is provided via configuration or environment. And, `TokenManager` module which provides different ways to obtain IAM token, either via `readToken` function or `TokenManager` class.

YES

#### Related PRs
Here is PR for wmcore/wmagent specs: https://github.com/cms-sw/cmsdist/pull/7647

None

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,3 +41,4 @@ retry~=0.9.2                  # wmcore,wmagent,wmagent-devtools,reqmgr2,reqmon,g
 rucio-clients~=1.25.5         # wmcore,wmagent,global-workqueue,reqmgr2ms
 Sphinx~=4.0.3                 # wmcore,wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
 SQLAlchemy~=1.3.3             # wmcore,wmagent
+PyJWT~=2.3.0                  # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms,wmcore
--- a/src/python/Utils/TokenManager.py
+++ b/src/python/Utils/TokenManager.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#pylint: disable=W0212
+"""
+File       : TokenManager.py
+Author     : Valentin Kuznetsov <vkuznet AT gmail dot com>
+Description: IAM token manager
+             by default it relies on the following environment/parameters:
+             - IAM_TOKEN is either file name or actual token value
+             - https://cms-auth.web.cern.ch/jwk is default CMS IAM provider
+             - https://wlcg.cern.ch/jwt/v1/any is default for audience parameter
+"""
+
+# system modules
+import os
+import ssl
+import time
+import logging
+import traceback
+
+# third part library
+try:
+    import jwt
+except ImportError:
+    traceback.print_exc()
+    jwt = None
+
+from Utils.Utilities import encodeUnicodeToBytes
+
+# prevent "SSL: CERTIFICATE_VERIFY_FAILED" error
+# this will cause pylint warning W0212, therefore we ignore it above
+ssl._create_default_https_context = ssl._create_unverified_context
+
+
+def readToken(name=None):
+    """
+    Read IAM token either from environment or file name
+    :param name: ether file name containing token or environment name which hold the token value.
+    If not provided it will be assumed to read token from IAM_TOKEN environment.
+    :return: token or None
+    """
+    if name and os.path.exists(name):
+        token = None
+        with open(name, 'r', encoding='utf-8') as istream:
+            token = istream.read()
+        return token
+    if name:
+        return os.environ.get(name)
+    return os.environ.get("IAM_TOKEN")
+
+
+def tokenData(token, url="https://cms-auth.web.cern.ch/jwk", audUrl="https://wlcg.cern.ch/jwt/v1/any"):
+    """
+    inspect and extract token data
+    :param token: token string
+    :param url: IAM provider URL
+    :param audUrl: audience string
+    """
+    if not token or not jwt:
+        return {}
+    if isinstance(token, str):
+        token = encodeUnicodeToBytes(token)
+    jwksClient = jwt.PyJWKClient(url)
+    signingKey = jwksClient.get_signing_key_from_jwt(token)
+    key = signingKey.key
+    headers = jwt.get_unverified_header(token)
+    alg = headers.get('alg', 'RS256')
+    data = jwt.decode(
+        token,
+        key,
+        algorithms=[alg],
+        audience=audUrl,
+        options={"verify_exp": True},
+        )
+    return data
+
+
+def isValidToken(token):
+    """
+    check if given token is valid or not
+
+    :param token: token string
+    :return: true or false
+    """
+    tokenDict = {}
+    tokenDict = tokenData(token)
+    exp = tokenDict.get('exp', 0)  # expire, seconds since epoch
+    if not exp or exp < time.time():
+        return False
+    return True
+
+
+class TokenManager():
+    """
+    TokenManager class handles IAM tokens
+    """
+
+    def __init__(self,
+                 name=None,
+                 url="https://cms-auth.web.cern.ch/jwk",
+                 audUrl="https://wlcg.cern.ch/jwt/v1/any",
+                 logger=None):
+        """
+        Token manager reads IAM tokens either from file or env.
+        It caches token along with expiration timestamp.
+        By default the env variable to use is IAM_TOKEN.
+        :param name: string representing either file or env where we should read token from
+        :param url: IAM provider URL
+        :param audUrl: audience string
+        :param logger: logger object or none to use default one
+        """
+        self.name = name
+        self.url = url
+        self.audUrl = audUrl
+        self.expire = 0
+        self.token = None
+        self.logger = logger if logger else logging.getLogger()
+        try:
+            self.token = self.getToken()
+        except Exception as exc:
+            self.logger.exception("Failed to get token. Details: %s", str(exc))
+
+    def getToken(self):
+        """
+        Return valid token and sets its expire timestamp
+        """
+        if not self.token or not isValidToken(self.token):
+            self.token = readToken(self.name)
+        tokenDict = {}
+        try:
+            tokenDict = tokenData(self.token, url=self.url, audUrl=self.audUrl)
+            self.logger.debug(tokenDict)
+        except Exception as exc:
+            self.logger.exception(str(exc))
+            raise
+        self.expire = tokenDict.get('exp', 0)
+        return self.token
+
+    def getLifetime(self):
+        """
+        Return reamaining lifetime of existing token
+        """
+        return self.expire - int(time.time())
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -60,6 +60,7 @@ from urllib.parse import urlencode
 
 from Utils.Utilities import encodeUnicodeToBytes, decodeBytesToUnicode
 from Utils.PortForward import portForward, PortForward
+from Utils.TokenManager import TokenManager
 
 
 def decompress(body, headers):
@@ -165,6 +166,11 @@ class RequestHandler(object):
         self.followlocation = config.get('followlocation', defaultOpts['FOLLOWLOCATION'])
         self.maxredirs = config.get('maxredirs', defaultOpts['MAXREDIRS'])
         self.logger = logger if logger else logging.getLogger()
+        self.tokenLocation = config.get('iam_token_file', '')
+        if self.tokenLocation:
+            self.tmgr = TokenManager(self.tokenLocation)
+        else:
+            self.tmgr = None
 
     def encode_params(self, params, verb, doseq, encode):
         """ Encode request parameters for usage with the 4 verbs.
@@ -245,6 +251,11 @@ class RequestHandler(object):
                 curl.setopt(pycurl.POSTFIELDS, encoded_data)
         else:
             raise Exception('Unsupported HTTP method "%s"' % verb)
+
+        if self.tmgr:
+            token = self.tmgr.getToken()
+            if token:
+                headers['Authorization'] = 'Bearer {}'.format(token)
 
         if verb in ('POST', 'PUT'):
             # only these methods (and PATCH) require this header
--- a/test/python/WMCore_t/Services_t/pycurl_manager_t.py
+++ b/test/python/WMCore_t/Services_t/pycurl_manager_t.py
@@ -5,8 +5,10 @@ Unit test for pycurl_manager module.
 from __future__ import division
 
 import gzip
+import os
 import tempfile
 import unittest
+import traceback
 from Utils.CertTools import getKeyCertFromEnv
 from WMCore.Services.pycurl_manager import \
         RequestHandler, ResponseHeader, getdata, cern_sso_cookie, decompress
@@ -167,6 +169,33 @@ class PyCurlManager(unittest.TestCase):
         # Kubernetes cluster responds with a different Server header
         serverHeader = res.getHeaderKey("Server")
         self.assertTrue(serverHeader.startswith("nginx/") or serverHeader.startswith("CherryPy/") or serverHeader.startswith("openresty/"))
+
+    def testToken(self):
+        """
+        Test setting up token header
+        """
+        iam_token = "eyJraWQiOiJyc2ExIiwiYWxnIjoiUlMyNTYifQ.eyJ3bGNnLnZlciI6IjEuMCIsInN1YiI6IjZjYTE3YzMyLTU0YWUtNDUzYy04YjQ1LTYyNTFkMjhlOTRhMSIsImF1ZCI6Imh0dHBzOlwvXC93bGNnLmNlcm4uY2hcL2p3dFwvdjFcL2FueSIsIm5iZiI6MTY0NTU1MDExMCwic2NvcGUiOiJhZGRyZXNzIHBob25lIG9wZW5pZCBvZmZsaW5lX2FjY2VzcyBwcm9maWxlIGVkdXBlcnNvbl9zY29wZWRfYWZmaWxpYXRpb24gZWR1cGVyc29uX2VudGl0bGVtZW50IGVtYWlsIHdsY2ciLCJpc3MiOiJodHRwczpcL1wvY21zLWF1dGgud2ViLmNlcm4uY2hcLyIsImV4cCI6MTY0NTU1MzcxMCwiaWF0IjoxNjQ1NTUwMTEwLCJqdGkiOiJkMWE3NzI0MC02Mjk5LTRiM2MtYmY5ZS0yOGNmYzgxY2ZjZDciLCJjbGllbnRfaWQiOiI0N2EwYjZkMC1mMzBlLTQ2OGItYmMwYy01MWVlNmE5Nzg2ODAifQ.Rpw6Mk_QzxtQVcbS-2OeVYGNMzyO08W540Sv3Nda2x3UJmPBRK4lnriRgSzMNTGP4y51rC5exbpf970MpJrjPaWqDhFWC--T6hxPAHhxHUxMTkXF5hUGXMLoNkCN5yR4eBJSAbgrgYJJwBcY5lMdCQ0BH5kQAUL6XRr3kvCrQQ0"
+        os.environ['IAM_TOKEN'] = iam_token
+        url = 'https://cmsweb-testbed.cern.ch'
+        params = {}
+        headers = {}
+        try:
+            # this call requires that our url (cmsweb-testbed) will
+            # accept token based header, therefore we wrap it up in
+            # try/except block to prevent from failing until
+            # token based auth will be in place
+            res = self.mgr.getheader(url, params=params, headers=headers, ckey=self.ckey, cert=self.cert)
+            if isinstance(self.mgr.request_headers, dict):
+                auth_header = self.mgr.request_headers.get('Authorization')
+                self.assertTrue(auth_header != "", True)
+            # the following test should be true only if we have valid token
+            # TODO: so far I don't know how to test this since it requires
+            # to obtain valid token
+            # self.assertTrue(auth_header == 'Bearer {}'.format(iam_token), True)
+        except Exception as exc:
+            # do not use print since it cause E1601 ("print statement used"), see
+            # https://github.com/PyCQA/pylint/issues/437
+            traceback.print_exc()
 
 if __name__ == "__main__":
     unittest.main()


vkuznet:
@amaltaro , @todor-ivanov  could you please review this PR. The changes are trivial and if merged we can start testing usage of WMCore services with token based authentication.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -61,6 +61,14 @@
 from Utils.PortForward import portForward, PortForward
 
 
+def get_token(env=None):
```
amaltaro, line 64: Valentin, I believe at some point we will want to have different operations for tokens (like, renewing, checking validity, etc) and maybe even different technologies. Right?

In that case, I would suggest to create a new module under the `Utils` package and have all the token-based implementation in there.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -242,6 +257,7 @@ def set_opts(self, curl, url, params, headers,
         if verbose:
             curl.setopt(pycurl.VERBOSE, True)
             curl.setopt(pycurl.DEBUGFUNCTION, self.debug)
+        self.request_headers = headers
```
amaltaro, line 260: Given that this attribute is not used anywhere but here in the `__init__` method, it looks like this line could be removed.

amaltaro (changes requested):
Valentin, in addition to the comments made along the code, I have a more general question/comment on how these tokens will be used, like:
* once we load a token in the `RequestHandler` object, I think we should keep using it instead of making the os.environ call for every HTTP call that goes through pycurl_manager.
* with that said, we would need to have a mechanism to also update our token because it will eventually expire, right?
* adding support to reading this token from a file might be a good idea

Last but not least, the `pycurl_manager` module also contains some functions. We need to make sure `getdata` function is covered with this implementation.

In terms of other HTTP modules, we use this one for uploading a root file - from within the worker node - to the DQM Gui:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/HTTPS/HTTPSAuthHandler.py
you might want to have a look at it as well.

vkuznet:
@amaltaro thanks for comments. We're working on setting up never expired token, similar to how we handle proxy for x509. Once new service will be in place, the token will be updated on regular basis and placed in a file. The token validity is defined by IAM provider and as far as I saw it is valid for 1hour. Therefore, I'll adjust code to do the following:
- add token manager to provide valid token
   - it will read token from default env/file
   - it will check validity of the token
   - if necessary it will re-read token again (if token is expired)
- initialize token manager at ctor

I'll put new code in Utils as you suggested and then use it here and pycurl manager.

src/python/WMCore/Services/pycurl_manager.py
vkuznet: the request headers can be required in different clients, e.g. DBSClient, and should be set accodringly. This is why it was set here. The are initially declared in ctor and set appropriate for every HTTP request.

vkuznet:
@amaltaro , I provided a first draft of TokenManager which so far can only read and validate token. Later, we may discuss and add new functionality how to obtain new token in TokenManager. Please bare this in your review. I explored several python libraries and decided to stick to [pyjwt](https://github.com/jpadilla/pyjwt/) recommended on [jwt.io](https://jwt.io/libraries). This library provides ability to inspect token and does not have much dependencies. This implies that it should be added to requirements of WMCore.

vkuznet:
@amaltaro , @todor-ivanov I am not sure I understand the following pylint error `[E1601, line 72 in tokenData: print statement used]()`. According to quick Google search it is false positive, see [here](https://github.com/PyCQA/pylint/issues/437)

amaltaro:
@vkuznet I haven't looked into this PR, but please replace the print statement by the logging library. This is an effort that we made a few years ago and kept the print statement only where we didn't manage to get a logger object working well.

amaltaro:
@vkuznet Valentin, could you please work on the cmsdist specs to get this new python library in the stack? Once you have that working, please link it to the first description of this PR. I haven't looked into these details, but I believe the massive failures are coming from the import error for the new library.

vkuznet:
@amaltaro I made necessary changes to replace print with logger. I checked that all pylint issues are gone. But the jenkins tests are failing due to the following:
```
...
  File "/build/cmsbld/jenkins/workspace/DMWM-WMCorePy3-PR-unittests/SLICE/0/label/cms-dmwm-cc7/code/src/python/WMCore/Services/Requests.py", line 80, in __init__
    self.reqmgr = RequestHandler()
name 'RequestHandler' is not defined
```
It seems to me that in Jenkins there is an issue with environment since `RequestHandler` is defined in `WMCore/Services/Requests.py` as following:
```
try:
    from WMCore.Services.pycurl_manager import RequestHandler, ResponseHeader
except ImportError:
    pass
```
Therefore, it seems to me that there is an issue with loading pycurl_manager which is part of this PR. Since there is not any printout in import error section we don't know what is the actual problem. Locally, I can easily load pycurl_manager and its pylint scores 10/10. If you want I can change `Requests.py` to add import error printout to understand the issue.

vkuznet:
@amaltaro , please be specific which spec file you're talking about? I adjusted code base to handle missing `jwt` import, hopefully this will fix failed unit tests. I think we should add separate PR for requirements.txt and the spec, but I need to know which spec(s) do you use these days.

vkuznet:
Here is new PR for specs: https://github.com/cms-sw/cmsdist/pull/7647

vkuznet:
Now, majority of unit tests failures are gone. The rest 5 failed tests are not related to this PR since they report issues with workflows which have nothing to do with proposed changes. I suggest that @amaltaro should look at those failure to decide what it is.

vkuznet:
@amaltaro I rebased this PR against master and added `pyjwt` dependency here too. Please review again.

amaltaro (changes requested):
Valentin, I have a few comments along the code.

In addition to that, I have some more general comments/questions:
* can you please improve those docstrings? This is going to be a common and shared library, so we better make it clear to anyone that starts looking into these modules.
* second, I wonder why you created those functions in the TokenManager module? Don't you think it's better to make it methods of the TokenManager class?
* you have encapsulated a few calls in try/except blocks. I believe it's only for the tests to pass, right? If that's the case, we could merge the cmsdist changes such that we get this library in our stack, and run these tests again (without try/except).

I will likely have further feedback for you, but it's probably better to work these out first. Thanks

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,132 @@
...
+
+class TokenManager():
+    """
+    TokenManager class handles IAM tokens
+    """
+    def __init__(self, name=None, url=None, aud=None, verbose=None, logger=None):
+        """
+        Token manager reads IAM tokens either from file or env.
+        It caches token along with expiration timestamp.
+        By default the env variable to use is IAM_TOKEN.
```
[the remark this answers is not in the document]
vkuznet: done

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        self.name = name
+        self.verbose = verbose
+        self.url = url
+        self.aud = aud
+        self.expire = 0
+        self.token = None
+        self.logger = logger if logger else logging.getLogger()
+        try:
```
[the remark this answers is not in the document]
vkuznet: we don't have yet mechanism how to create token as code only reads existing one. Once we'll add code to create tokens then we'll add real failure procedure

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -213,6 +219,14 @@ def set_opts(self, curl, url, params, headers,
         else:
             raise Exception('Unsupported HTTP method "%s"' % verb)
 
+        try:
+            token = self.tmgr.getToken()
```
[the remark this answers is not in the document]
vkuznet: It really depends how we decide to handle token creation. It is not the same as X509. Current procedure is to have separate process/daemon which will handle token creation. it is more like ssh-agent rather the voms-proxy-init. In former case, you deploy separate token agent which handle the tokens. In latter, you really obtain a new proxy.

vkuznet:
@amaltaro to answer your questions:
- I added/improved docstrings
- the functions are separate from TokenManager class since they are only need to read tokens. Depending how you'll manage tokens the TokenManager may create new tokens or refresh it, etc. Therefore, someone who will use this library can use either function, e.g. only import function to read token from env, or have TokenMamanger to handle token entirely.
- try/except added if infrastructure does not provide any token and code should function as is. I think we still need it until we'll setup infrastructural pieces in place
- token may or may not be presented in HTTP request, e.g. we still have period when we'll need to work with x509, therefore we still need try/except block regardless of tokens during the transition period
- token management is different from x509, as I wrote IAM provide token daemon similar to ssh agent, and it will handles the tokens for you. We only need to read token which will be managed by external software. Said that, we may also need to add functionality to obtain tokens in env which will not have token manager, that's why we may need and extend TokenManager.
- I removed requests_headers and let jenkins complain if it finds missing dependencies. The DBSClient relies on RestClient which by itself uses pycurl manager and headers are dumped in HTTPError.

amaltaro:
Valentin, side question, in addition to the `oidc-agent` service that we can install on these nodes to provide us with tokens. Would there be another option using the current software stack (or with this pyjwt library)?

vkuznet:
Tokens can be obtained in a different way and from different providers. For instance, the auth proxy server (aps) I'm pushing to replace our apache stack is capable to work with tokens from CERN SSO and IAM. As such, there are different ways to get tokens. For instance, someone can get CERN SSO token via web interface, e.g.  on aps server someone can visit `/token` end-point and you'll get your token. If you're using IAM, you can use either oidc-token, or you may request client credentials to get token from IAM directly (e.g. we created token manager pod for k8s which is doing exactly that and it will keep fresh token for all services in k8s infrastructure). I think using client credentials (which is client id and client secret) it is possible to use 3rd party library, e.g. pyjwt, to obtain token, but it will require coding and management on your infrastructure (someone will need to securely handle client credentials).

But I think ultimately we'll rely on oidc agent to provide us tokens to reduce overhead of management them. In k8s, we already have token manager for that which works with IAM. I also provide token manager as part of aps codebase (but it is not python it is written in Go), and it works with aps+CERN SSO.

amaltaro:
test this please

amaltaro (changes requested):
This is not a full review, but could you please look into these failing unit tests?

I had a very quick look at the `oidc-agent` service as well and the installation procedure, and I do not see any RPM installation instructions. Can't this complicate deployment of this in a non-containerized environment? 

WMAgents can run in different OS distributions as well - FNAL vs CERN nodes - and this is another item to be taken into consideration.

Having said that, unless it requires a substantial amount of effort, I would say that implementing our own code to fetch a token would be the best (we should maintain one credential for CERN agents, one for FNAL agents). This is more or less what has been done for Rucio here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Tools/PycurlRucio.py#L65-L108

vkuznet:
regarding rpm based install for token tools, can it be as simple as `yum search oidc` and then `yum install oidc...`? Regarding Rucio token, it is not the same as OIDC one (as far as I know), therefore you will still need to manage tokens. Said that, on k8s we already do that and everything is ready for services to use it. For WMAgent we can provide a daemon too, but I think you just need to try first yum approach. Please read [install instructions](https://indigo-dc.gitbook.io/oidc-agent/installation/install) which basically confirms what I just wrote:
```
CentOS 7
sudo yum install \
      libcurl-devel \
      libsodium-devel \
      libsodium-static \
      libmicrohttpd-devel \
      libseccomp-devel \
      libsecret-devel \
      libqrencode-devel
```

vkuznet:
test this please

vkuznet:
@amaltaro , I looked at failed tests and they are related to Unauthorized CouchDB access, see https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/12936/testReport/junit/WMComponent_t.ErrorHandler_t.ErrorHandler_t/ErrorHandlerTest/testD_Exhausted/
Did you change anything for couch access in jenkins. I have no idea why proposed changes can affect couch. The only change is to pycurl_manager which is within tr/except block which pass if token is not present. Therefore it can't affect anything. The `TokenManager` class is independent from the existing WMCore and can't affect either. Therefore, I don't see any logical reason how the proposed code affects the CouchDB access. Any clue?

amaltaro:
The only change that went in in the last week'ish was this:
https://github.com/dmwm/WMCore/pull/11045

and it did not touch ErrorHandler code. Are you able to reproduce this error when testing with WMCore container unit test setup? Or it only shows up in jenkins?

vkuznet:
Alan, do you have any documentation how to setup "WMCore container unit test setup"? I never did it and have no idea how to do it. Moreover, you should also explain how to do it with and without #11045 PR.

vkuznet:
Alan, after lots of digging and following a different documentation [wiki page](https://github.com/dmwm/WMCore/wiki/setup-wmcore-unittest#1-using-the-wmagent-docker-container) I setup WMCore container where I executed failed test, and it is failing like this:

```
# I patched WMCore with changes I provide in this PR

# I started WMCore container

# I run one of the tests 
python3 /home/dmwm/wmcore_unittest/WMCore/test/python/WMComponent_t/ErrorHandler_t/ErrorHandler_t.py
....
http.client.HTTPException: url=http://localhost:6994/job_summary/_design/WMStatsAgent/_update/jobSummaryState/7b07d232-b161-4419-9507-54a42e2f6b70?newstate=jobcooloff&timestamp=1648584707, code=404, reason=Object Not Found, headers={'Server': 'CouchDB/1.6.1 (Erlang OTP/R16B03-1)', 'Date': 'Tue, 29 Mar 2022 20:11:47 GMT', 'Content-Type': 'application/json', 'Content-Length': '41', 'Cache-Control': 'must-revalidate'}, result=b'{"error":"not_found","reason":"missing"}\n'

  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/JobStateMachine/ChangeState.py", line 178, in propagate
    self.recordInCouch(jobs, newstate, oldstate, updatesummary)
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/JobStateMachine/ChangeState.py", line 304, in recordInCouch
    self.jsumdatabase.makeRequest(uri=updateUri, type="PUT", decode=False)
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/Database/CMSCouch.py", line 135, in makeRequest
    self.checkForCouchError(getattr(e, "status", None),
  File "/home/dmwm/wmcore_unittest/WMCore/src/python/WMCore/Database/CMSCouch.py", line 153, in checkForCouchError
    raise CouchNotFoundError(reason, data, result)
WMCore.Database.CMSCouch.CouchNotFoundError: CouchNotFoundError - reason: Object Not Found, data: None result: None
```

The output seems identical to what jenkins provide, therefore I would conclude that something is related to CouchDB setup in a container. Please advise how to proceed forward with it.

vkuznet:
Alan, please respond to this ticket on how to move forward with tests. I provided output from WMCore container as well.

vkuznet:
@amaltaro any suggestions how to move forward with it? I'm awaiting your feedback on how to proceed with this PR.

amaltaro:
test this please

amaltaro:
Out of the 5 new failing unit tests, only this:
```
WMComponent_t.TaskArchiver_t.TaskArchiver_t.TaskArchiverTest:testA_BasicFunctionTest
```

fails when running from the docker image. The other 4 tests are successful, with or without your patch.
Jenkins is complaining about:
```
CouchUnauthorisedError - reason: Unauthorized, data: {} result: None
```

I don't think these failures are coming from this PR, but it needs further investigation. One thing I can say, our unit tests aren't behaving well for a few weeks now (we keep seeing the "tests added", even though it's been weeks that those have been added).

vkuznet:
ok, from the error message it seems it is related to Couch access and most likely expired certificate in jenkins. But since tests do not tell which Couch it contacts I can't verify that. If you know the CouchDB address we may check FE logs.

amaltaro:
Besides a some services/calls that are not mocked, our unit tests are pretty much self container. Meaning, we deploy MariaDB and CouchDB at the localhost, run the unit tests, and clean up everything.

I'm adding a debugging statement in the jenkins setup, let me trigger tests once again.

amaltaro:
test this please

amaltaro:
For the record, this log has a few CouchUnauthorized errors:
https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCorePy3-PR-unittests/SLICE=9,label=cms-dmwm-cc7/lastBuild/consoleText

@vkuznet can you please add a print statement above this line:
https://github.com/dmwm/WMCore/blob/eba0a315ed973616357e231976f7092adcb6b2e6/src/python/WMCore/JobStateMachine/ChangeState.py#L92

printing the value of `self.config.JobStateMachine.couchurl`? Push this in as a new commit please.

vkuznet:
test this please

vkuznet:
Alan, tests are now passing. There are two warnings about unit tests but I don't know if you want to fix them. I put explicitly few comments about the code and we need to decide how to run unit tests when we'll handle real tokens. Feel free to make suggestions.

amaltaro:
@vkuznet Valentin, I am going to review it once again tomorrow. For now, could you please squash your commits (actually when rebasing it, you can simply delete the last 2 commits and force push it). But please do whatever GH workflow works for you.

vkuznet:
For completeness here are details about new token manager services on k8s:
- we created a new token manager services which is based on a regular cron job
- it creates new token on periodic interval and update a given file (which hold the token) on k8s cluster
- the new service is deployed on all cmsweb clusters
All details can be found in this ticket: https://its.cern.ch/jira/browse/CMSKUBERNETES-129

If DMWM wants to adapt this strategy outside of k8s infrastructure, e.g. on WMAgent nodes, then the following set of actions are required:
- install `oidc` token based tools based on set of instructions from IAM, see [here](https://indigo-iam.github.io/v/v1.7.0/docs/tasks/user/getting-a-token/). The RPMs or other installation methods are available.
- register your services or application with IAM provider and obtain its credentials (client_id and secret)
- run the following command on a node to obtain token on regular interval
```
# adjust client_id and client_secret accordingly to your setup
export client_id=`cat /etc/client/client_id`
export client_secret=`cat /etc/client/client_secret`

# obtain new token using client credentials
curl -s -d grant_type=client_credentials -d scope="profile" -u ${client_id}:${client_secret} https://cms-auth.web.cern.ch/token | jq -r '.access_token' > /tmp/token
```

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,149 @@
...
+    if name and os.path.exists(name):
+        return readTokenFile(name)
+    return readTokenEnv(name)
+
+def readTokenEnv(env=None):
+    """
+    Get IAM token from environment
+    :param env: environment name to read the token, if not provided we'll use IAM_TOKEN
+    """
+    if env:
```
amaltaro, line 45: It can be refactored to:
```
    if env:
        return os.environ.get(env)
    return os.environ.get("IAM_TOKEN")
```

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,149 @@
...
+        return os.environ.get(env)
+    if os.environ.get("IAM_TOKEN"):
+        return os.environ.get("IAM_TOKEN")
+    return None
+
+def readTokenFile(fname=None):
```
amaltaro, line 51: `fname` should be camel case --> `fName`. Please rework the docstring as well.

src/python/WMCore/Services/pycurl_manager.py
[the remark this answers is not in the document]
amaltaro: Yes, what I am saying is that if the system is defined to use tokens, it should use it instead of the x509 certificates, even if that token is expired!

So, I would basically take this code out of the try/except block.

amaltaro (changes requested):
@vkuznet Valentin, this is another partial review because I think there are a few things worth changing first. 

I made some comments along the code, but regarding those functions in the TokenManager module, the `getToken` function is so small that I think we should  merge `readTokenEnv` and `readTokenFile`. Which will also minimize code duplication.

Given that TokenManager is a new code, please run an code reformat on it such that it comes out as clean as possible. For instance, spacing between functions are meant to be 2 blank lines.

Regarding pycurl_manager, I am in favor of: if the client specifies that it wants to use tokens, then requests should either be made with it, or fail. From the code implemented, I think it would fallback to the certificates, no?

vkuznet:
@amaltaro I addressed all issues you raised except merging `readTokenEnv` and `readTokenFile` functions. I think they should be kept separately since there may be different use-cases for them. We still don't know mode of operations we'll end-up with and both `env` and `file` based ones are possible. As such, I still think we need both functions to use for different purposes.

amaltaro:
> @amaltaro I addressed all issues you raised except merging readTokenEnv and readTokenFile functions. I think they should be kept separately since there may be different use-cases for them. We still don't know mode of operations we'll end-up with and both env and file based ones are possible. As such, I still think we need both functions to use for different purposes.

Thanks Valentin. And I agree with that! Sorry for not being clear, what I meant was to merge those two functions and move their code under the `getToken` function, which already has some of that logic. This will make the code more clear and less redundant IMO.

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    headers = jwt.get_unverified_header(token)
+    alg = headers.get('alg', 'RS256')
+    if not aud: # use WLCG
+        aud = "https://wlcg.cern.ch/jwt/v1/any"
+    data = jwt.decode(
+        token,
+        key,
+        algorithms=[alg],
+        audience=aud,
+        options={"verify_exp": True},
```
amaltaro, line 70: can you please add a comment on what this `verify_exp` means and is used for?

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+    return data
+
+def validToken(token):
```
amaltaro, line 74: Perhaps calling it `isTokenValid` would be better?

Somehow unrelated to this, but we should also have a function/method that returns the lifetime of a given token (such that we can get a new one *before* we see it expired)

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    validate token expiration timestamp
+    :param token: token string
```
amaltaro, line 77: docstring needs some rework.

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    :return: token or None
+    """
+    if name and os.path.exists(name):
+        token = None
+        with open(name) as istream:
+            token = istream.read()
+        return token
+    if name:
+        return os.environ.get(name)
+    return os.environ.get("IAM_TOKEN")
```
amaltaro, line 43: I'd suggest to actually suggest a global constant with this default environment variable name, e.g. (at the top of the module):
```
DEFAULT_TOKEN_ENV="IAM_TOKEN"
```
what do you think?

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    TokenManager class handles IAM tokens
+    """
+    def __init__(self, name=None, url=None, audUrl=None, logger=None):
+        """
+        Token manager reads IAM tokens either from file or env.
+        It caches token along with expiration timestamp.
+        By default the env variable to use is IAM_TOKEN.
+        :param name: string representing either file or env where we should read token from
+        :param url: IAM provider URL, by default it is https://cms-auth.web.cern.ch/jwk
```
amaltaro, line 101: Same comment as the one above, for these default urls.

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+        if not self.token or not validToken(self.token):
+            self.token = getToken(self.name)
+        tokenDict = {}
+        try:
+            tokenDict = tokenData(self.token, url=self.url, aud=self.aud)
+            self.logger.debug(tokenDict)
+        except Exception as exc:
+            self.logger.exception(str(exc))
+            raise
+        self.expire = tokenDict.get('exp', 0)
```
amaltaro, line 129: Along the comment made above - for creating a new function - now that I look into this, it makes sense to me to have a method called `getExpirationTime` or `getLifetime` to read this attribute.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -165,6 +166,12 @@ def __init__(self, config=None, logger=None):
         self.followlocation = config.get('followlocation', defaultOpts['FOLLOWLOCATION'])
         self.maxredirs = config.get('maxredirs', defaultOpts['MAXREDIRS'])
         self.logger = logger if logger else logging.getLogger()
+        self.request_headers = None
```
amaltaro, line 169: This seems not to be necessary. You might want to remove the TMP commented out code as well (from 100 lines below).

amaltaro (changes requested):
Valentin, thanks for making those changes.

I have further comments for your consideration.

In addition to those, I have a few more general comments:
1) is there any reason to define both functions and class/methods for the TokenManager? Do you see any problems in actually moving those functions as a TokenManager method? IMO, the code looks more organized and clear when we have methods defined in a class; and we also can profit from the logger object and have proper logging in place.
2) please run a code reformat in the new module (TokenManager.py). pep8 is failing to run in jenkins, but I see a few places that we could have enhance and avoid suggestions from jenkins (like double spaces between modules). Together with this, unless there is really no better option, I'd say that variables with 3 chars only are usually not the most meaningful and cleaner option.
3) we can break things down in smaller parts, but I think that having the capability of creating a token without any external service/daemon would be extremely useful (and reduce dependencies on our stack/jenkins/dev setup).

todor-ivanov (commented):
Hi @vkuznet for providing all this construct. This functionality is really needed. I have left a comment or two inline, but in general they all go in the same direction, which is: 
I think having the IAM token provided through the environment would be the best approach. And from the two possible options:
* One - the whole token stored in an environment variable
* The other - the whole token stored in a file and pointed to by an environment variable holding the filename
I'd say I am more in favor of the later. But having them both implemented and having the flexibility to chose of course would be the best. 

Thanks!

requirements.txt
```diff
@@ -41,3 +42,4 @@ retry~=0.9.2                  # wmcore,wmagent,wmagent-devtools,reqmgr2,reqmon,g
 rucio-clients~=1.25.5         # wmcore,wmagent,global-workqueue,reqmgr2ms
 Sphinx~=4.0.3                 # wmcore,wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
 SQLAlchemy~=1.3.3             # wmcore,wmagent
+pyjwt~=2.3.0                  # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
```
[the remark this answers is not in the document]
vkuznet: ok

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    if name and os.path.exists(name):
+        token = None
+        with open(name) as istream:
+            token = istream.read()
+        return token
+    if name:
+        return os.environ.get(name)
+    return os.environ.get("IAM_TOKEN")
+
+def tokenData(token, url=None, aud=None):
```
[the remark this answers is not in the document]
vkuznet: fine

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,130 @@
...
+    inspect and extract token data
+    :param token: token string
+    :param url: IAM provider URL, by default it is https://cms-auth.web.cern.ch/jwk
+    :param aud: audience string, by default it is https://wlcg.cern.ch/jwt/v1/any
+    """
+    if not token or not jwt:
+        return {}
+    if isinstance(token, str):
+        token = bytes(token, 'utf-8')
+    if not url: # use CMS IAM
```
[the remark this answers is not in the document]
vkuznet: yes

src/python/Utils/TokenManager.py
vkuznet: ok

src/python/Utils/TokenManager.py
vkuznet: no, it should be `isValidToken` since we read token and it can be mailformed, and only if we read it correctly we verify its expiration timestamp. So, let's not change it.

src/python/Utils/TokenManager.py
vkuznet: to Alan: why do you want to invent and introduce yet another global variable, I always against all globals.
to Todor: why do you want to complicate handling of different use-case by adding more code? In current implementation you can either read token from env or from file using the same function. In your proposal, not only we need two separate functions here, but also on a client side you'll need to call them independently. Why do you want to duplicate behavior while gaining nothing?

src/python/WMCore/Services/pycurl_manager.py
vkuznet: done

src/python/Utils/TokenManager.py
vkuznet: it is parameter of jwt linbrary to check token validity

vkuznet:
I adjusted code based on your comments, but you also asked several questions which I answer below:
Alan's questions:
- class vs functions, the `TokenManager` class will handle token. We said since we don't know yet which behavior we want to token manager we keep it as a class. On a contrary, if client wants to read a token it does not need entire class and can use simple function. Once we understand behavior we want we can adjust to one or another. But, since functions does not require any class member data it is simpler and less costly to keep it as simple function.
- pep8 autoformat is applied now
- if we want to manage tokens within python code then you'll need:
   - add code to obtain token (this can be done via curl call) but for that you'll need to manage client's credentials
   - you'll need to extend `TokenManager` to do this job (which so far is done by external tool)
   - you'll need to maintain TokenManager objects/instances within all different applications which will require token instead of using token oidc tools
   - bottom line, if you want independence from external tool you'll basically need to write its functionality and maintain it. I doubt it is productive. We got tools exactly for that, e.g. you don't manage you ssh keys, you don't need to manage tokens, just use existing tools

Todor's questions:
- I doubt that support of both use-cases (env and file) via separate code base will make code easier, you'll need to implement both use-case within token manager module and on a client side, you'll need to decide which method should come first, which second, what to do if one fail, etc. By unifying both your client does not care how token will be provided, i.e. either in env or in file. Moreover, file can be updated by external daemon, we may have different use-cases, e.g. in k8s we'll provide you a file which will be updated so you don't need to manage it, while on WMA you may decide to use env and obtain token in a different. way. Bottom line, so far you did not provide any benefits that separation will be easier to maintain.

vkuznet:
@amaltaro , @todor-ivanov , I replied to your comments. Please review this PR again and let's move it forward.

todor-ivanov:
Thanks @vkuznet! 
I did try to take a look at the latest changes of yours, but it seems a huge merge is obscuring the picture: https://github.com/dmwm/WMCore/pull/10939/commits/d99704fa846463daeb361791ecbadf6cce72a316 
Is this due to a merge of `master` branch  on top of a  `development` branch or due to a rebase? This may also affect the master branch history as well.

vkuznet:
@todor-ivanov , at some point I rebased it to master and this is probably what you see, but this PR does not show any of that, i.e. I can only see my changes, `Files changed 4` which reflects only my changes. Can you be more specific about [d99704f](https://github.com/dmwm/WMCore/pull/10939/commits/d99704fa846463daeb361791ecbadf6cce72a316) as it is not shown in `Files changed` on github.

todor-ivanov:
Hi @vkuznet sorry, .. indeed it is not shown at the 'Files changed` tab. I was looking at the link from the github mail I received  and it was pointing to the full rebase diff .... which was terrifying at a first glance. I am looking at the new changes now.

todor-ivanov (approved):
Looks good to me.
Thanks @vkuznet

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,135 @@
...
+    :param token: token string
+    :return: true or false
+    """
+    tokenDict = {}
+    try:
+        tokenDict = tokenData(token)
+    except:
+        traceback.print_exc()
+        raise
+    exp = tokenDict.get('exp', 0)
```
amaltaro, line 89: Can you please add a comment on what this `exp` attribute is supposed to be?
I guess it stands for `expires`, with an integer EPOCH time of when it expires?

src/python/Utils/TokenManager.py
amaltaro: Another reminder for the default urls (similar to the one provided in the function above).

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,135 @@
...
+        self.name = name
+        self.url = url
+        self.audUrl = audUrl
+        self.expire = 0
+        self.token = None
+        self.logger = logger if logger else logging.getLogger()
+        try:
+            self.token = self.getToken()
+        except Exception as exc:
+            self.logger.exception(str(exc))
```
amaltaro, line 119: Can you add some more context to this error message? Something like:
```
self.logger.exception("Failed to get token. Details: %s", str(exc))
```
?

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,135 @@
...
+except ImportError:
+    traceback.print_exc()
+    jwt = None
+
+# prevent "SSL: CERTIFICATE_VERIFY_FAILED" error
+# this will cause pylint warning W0212, therefore we ignore it above
+ssl._create_default_https_context = ssl._create_unverified_context
+
+
+def getToken(name=None):
```
amaltaro, line 29: Now that I am getting to the end of this review. I see that we have this `getToken` function but the class below also has a `getToken` method. Even though we need a `TokenManager` object (self) to call the `getToken` method, I think it would be less confusing to rename this function to actually `readToken`.
What do you think?

src/python/Utils/TokenManager.py
amaltaro: Yes, what I wanted to say is that we should have something like:
```
def getLifetime(self):
    return self.expire - int(time.time()) 
```

The use case here is that we want to renew this token BEFORE it gets expired. So we need to have a mechanism to constantly poll how long it's still valid for.

src/python/WMCore/Services/pycurl_manager.py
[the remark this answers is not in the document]
amaltaro: Resolving this comment as you have already implemented the suggested changes, thanks.

amaltaro (changes requested):
Valentin, please find further comments along the code.
I also wanted to mention that we have integrated `pyjwt` into our jenkins setup, in case you want to try more realistic tests.

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Get IAM token either from environment or file name
+    :param name: ether file name containing token or environment name which hold the token value.
+    If not provided it will be assumed to read token from IAM_TOKEN environment.
+    :return: token or None
+    """
+    if name and os.path.exists(name):
+        token = None
+        with open(name) as istream:
```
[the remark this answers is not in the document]
vkuznet: done

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,135 @@
...
+            token = istream.read()
+        return token
+    if name:
+        return os.environ.get(name)
+    return os.environ.get("IAM_TOKEN")
+
+
+def tokenData(token, url=None, audUrl=None):
```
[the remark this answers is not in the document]
vkuznet: done

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    inspect and extract token data
+    :param token: token string
+    :param url: IAM provider URL, by default it is https://cms-auth.web.cern.ch/jwk
+    :param audUrl: audience string, by default it is https://wlcg.cern.ch/jwt/v1/any
+    """
+    if not token or not jwt:
+        return {}
+    if isinstance(token, str):
```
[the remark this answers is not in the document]
vkuznet: done

src/python/Utils/TokenManager.py
vkuznet: done

src/python/Utils/TokenManager.py
vkuznet: done

src/python/Utils/TokenManager.py
vkuznet: done

src/python/Utils/TokenManager.py
amaltaro: Thanks Valentin, I'm resolving this as well.

src/python/Utils/TokenManager.py
```diff
@@ -0,0 +1,139 @@
...
+        return False
+    return True
+
+
+class TokenManager():
+    """
+    TokenManager class handles IAM tokens
+    """
+
+    def __init__(self, name=None, url=None, audUrl=None, logger=None):
```
amaltaro, line 98:
```suggestion
    def __init__(self, name=None, url="https://cms-auth.web.cern.ch/jwk",
                        audUrl="https://wlcg.cern.ch/jwt/v1/any", logger=None):
```

this is what I am suggesting in my comment below (maybe the spacing in my comment is not correct, so please consider it as well).

src/python/Utils/TokenManager.py
amaltaro: Resolving this as well, since you implemented this method below. Thanks

src/python/Utils/TokenManager.py
amaltaro: Regarding this comment, please actually consider the one made in the `__init__` method signature.

src/python/Utils/TokenManager.py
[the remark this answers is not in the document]
amaltaro: Valentin, you can remove the `isinstance` check in the line above because it's already performed within the `encodeUnicodeToBytes` function.

amaltaro (changes requested):
Valentin, I had another look at this and it's looking good now, almost ready to go.
I resolved a few comments that you have already worked on but were not automatically closed by GH, and left 2 or 3 comments that are worth it.

Once you have those changes in place, feel free to rebase and squash these commits. If it's not too late, please keep the test changes in a separate commit.

requirements.txt
```diff
@@ -41,3 +41,4 @@ retry~=0.9.2                  # wmcore,wmagent,wmagent-devtools,reqmgr2,reqmon,g
 rucio-clients~=1.25.5         # wmcore,wmagent,global-workqueue,reqmgr2ms
 Sphinx~=4.0.3                 # wmcore,wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
 SQLAlchemy~=1.3.3             # wmcore,wmagent
+pyjwt~=2.3.0                  # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms,wmcore
```
amaltaro, line 44: Luckily I just looked this up in Pypi and noticed that this name isn't correctly written. As far as I know, Pypi is case sensitive, so please update it to `PyJWT`, from: https://pypi.org/project/PyJWT/

amaltaro:
@vkuznet Valentin, can you also please give a brief summary of what is provided in this PR and what kind of token management it allows now? It'd be useful to link the WMCore token wiki as well: https://github.com/dmwm/WMCore/wiki/Tokens-in-WMCore

FYI @belforte @mapellidario this change is likely going in in the next release and it depends on a new python 3rd-party library called: `PyJWT`, which has been already integrated to `cmsdist` repo, but our wmcore/crab specs still need to have this dependency added to the Requires list.

Last but not least, these changes - especially to pycurl_manager - should be backwards compatible, so even if we are not using tokens and don't have `PyJWT` library, nothing should break. Valentin can confirm this though.

requirements.txt
vkuznet: done

src/python/Utils/TokenManager.py
vkuznet: done

vkuznet:
@amaltaro , thanks for suggestions I adjusted the codebase accordingly. Regarding PR description. I updated it too. But this PR does not provide any token management, we only provide methods (either via function or class) to read token, but we do not manage tokens. Pointer to wiki has been added.

amaltaro:
Thank you, Valentin. Yes, by management I meant that we have some utilities to load in a token and check expiration time. But I agree this is only a small set of managing tokens.

If you are ready with these changes, please:
1) squash the commits accordingly
2) remove the no longer relevant GH labels
3) update the status `In development` to `ready` or `done`. 
4) and fire a final request review.

vkuznet:
test this please

amaltaro (approved):
Thanks, Valentin.

belforte:
thanks for keeping us in the loop. Appreciated.

PR #10959, opened 2022-01-26
Send alerts to prometheus when a workflow fails to be processed.

### LINKED ISSUE #10215: Create monitoring or alerts for workflows failing in MSTransferor
**Impact of the new feature**
MSTransferor

If a workflow fails to be processed by MSTransferor, the only way to spot those - and see what the error is - is by scanning the service logs from CMSWEB. Thus, we should make it easier to access this information, either with some sort of monitoring; or via push notifications/alerts.

A final solution isn't well defined yet. However, a few options that I could list now are:
a) push details to the LogDB, in central CouchDB, via `error logs` WMStats tab
b) send email alerts. Need to make sure not to spam too much with duplicated alerts
c) maybe, setup something in MonIT
d) or perhaps, provide this information through some REST API (still, someone would have to spot those workflows stuck in assigned)

any other suggestions?

The error description does not necessarily need to be a log of the workflow processing in MSTransferor, but a clear message identifying why the workflow does not move forward in the service.

A few described above, now we need to agree on one of them.

**Additional context**
none

khurtado:
Fixes #10215

#### Status
In development

#### Description
Send alerts to prometheus when a workflow fails to be processed.

YES

### DIFF
--- a/src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
+++ b/src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
@@ -199,8 +199,7 @@ class MSTransferor(MSCore):
                 self.checkPUDataLocation(wflow)
                 if wflow.getSecondarySummary() and not wflow.getPURSElist():
                     # then we still have pileup to be transferred, but with incorrect locations
-                    msg = "Workflow: %s cannot proceed due to some PU misconfiguration. Check previous logs..."
-                    self.logger.critical(msg, wflow.getName())
+                    self.alertPUMisconfig(wflow.getname())
                     # FIXME: this needs to be logged somewhere and workflow be set to failed
                     counterProblematicRequests += 1
                     continue
@@ -212,8 +211,9 @@ class MSTransferor(MSCore):
                     success, transfers = self.makeTransferRequest(wflow)
                 except Exception as ex:
                     success = False
-                    msg = "Unknown exception while making Transfer Request for %s " % wflow.getName()
-                    msg += "\tError: %s" % str(ex)
+                    self.alertUnknownTransferError(wflow.getName())
+                    msg = "Unknown exception while making transfer request for %s " % wflow.getName()
+                    msg = "\tError: %s" % str(ex)
                     self.logger.exception(msg)
                 if success:
                     self.logger.info("Transfer requests successful for %s. Summary: %s",
@@ -225,6 +225,7 @@ class MSTransferor(MSCore):
                         counterSuccessRequests += 1
                     else:
                         counterFailedRequests += 1
+                        self.alertTransferCouchDBError(wflow.getname())
                 else:
                     counterFailedRequests += 1
             # it can go slightly beyond the limit. It's evaluated for every slice
@@ -729,7 +730,7 @@ class MSTransferor(MSCore):
                     self.logger.info("Rules successful created for %s : %s", dataIn['name'], res)
                     transferId.update(res)
                     # send an alert, if needed
-                    self.notifyLargeData(aboveWarningThreshold, transferId, wflow.getName(), dataSize, dataIn)
+                    self.alertLargeInputData(aboveWarningThreshold, transferId, wflow.getName(), dataSize, dataIn)
                 else:
                     self.logger.error("Failed to create rule for %s, will retry later", dids)
                     success = False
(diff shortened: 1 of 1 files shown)


src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
```diff
@@ -201,6 +201,15 @@ def execute(self, reqStatus):
                     # then we still have pileup to be transferred, but with incorrect locations
                     msg = "Workflow: %s cannot proceed due to some PU misconfiguration. Check previous logs..."
                     self.logger.critical(msg, wflow.getName())
+                    # Send alert to prometheus
```
amaltaro, line 204: Kenyi, I am inclined to say that the content of this alert should be encapsulated in a method/function, something similar to `notifyLargeData`. Such that this logic does not get too much polluted with the alert information. Maybe we could call it something like `def alertPUMisconfig`.

amaltaro (changes requested):
Kenyi, these changes are looking good. However, I think these alerts will be better organized if we actually have a wrapper function/method for each of them. We might even want to create a new module called something like `MSTransferorAlerts` and encapsulate all this information in there.

If you feel like refactoring the `notifyLargeData` method as well, that would be even better :-D

What do you think?

src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
```diff
@@ -225,8 +242,27 @@ def execute(self, reqStatus):
...
+                        alertName = "{}: Transfer document error posting to CouchDB. Workflow: {}".format(self.alertServiceName,
+                                                                                         wflow.getname())
+                        alertSeverity = "high"
+                        alertSummary = "[MSTransferor] Transfer document could not be created in CouchDB."
+                        alertDescription = "Workflow: {}, failed request  due to error posting to CouchDB".format(wflow.getName())
+                        self.sendAlert(alertName, alertSeverity, alertSummary, alertDescription,
+                                       self.alertServiceName)
                 else:
                     counterFailedRequests += 1
+                    msg = "Unknown exception while making Transfer Request for %s " % wflow.getName()
```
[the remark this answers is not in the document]
khurtado: Ahh, you are right! I will just get rid of this alert, then

khurtado:
@amaltaro : Okay, I encapsulated the alerts in different methods. Could you please check the second commit?

src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
```diff
@@ -738,6 +738,59 @@ def makeTransferRucio(self, wflow, dataIn, subLevel, blocks, dataSize, nodes, no
...
+        self.logger.critical(alertDescription)
+
+    def alertUnknownTransferError(self, workflowName):
+        """
+        Send alert to Prometheus with unknown transfer error
+        """
+        alertName = "{}: Transfer request error. Workflow: {}".format(self.alertServiceName,
+                                                                         workflowName)
+        alertSeverity = "high"
+        alertSummary = "[MSTransferor] Unknown exception while making Transfer Request."
```
amaltaro, line 773: I think we do not need the initial capital letters here. It could be "transfer request", no? Same comment for the line below.

Kenyi, in addition to that, this log exception is already executed on the caller function. Perhaps we can leave it only there and remove it from this method?

src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
```diff
@@ -738,6 +738,59 @@ def makeTransferRucio(self, wflow, dataIn, subLevel, blocks, dataSize, nodes, no
...
+        self.sendAlert(alertName, alertSeverity, alertSummary, alertDescription,
+                       self.alertServiceName)
+        msg = "Unknown exception while making Transfer Request for %s " % wflow.getName()
+        self.logger.exception(alertDescription)
+
+    def alertTransferCouchDBError(self, workflowName):
+        """
+        Send alert to Prometheus with CouchDB transfer error
+        """
+        alertName = "{}: Transfer document error posting to CouchDB. Workflow: {}".format(self.alertServiceName,
```
amaltaro, line 784: How about `{}: Failed to create a transfer document in CouchDB for workflow: {}`?

src/python/WMCore/MicroService/MSTransferor/MSTransferor.py
```diff
@@ -738,6 +738,59 @@ def makeTransferRucio(self, wflow, dataIn, subLevel, blocks, dataSize, nodes, no
...
+                                                                         workflowName)
+        alertSeverity = "high"
+        alertSummary = "[MSTransferor] Transfer document could not be created in CouchDB."
+        alertDescription = "Workflow: {}, failed request  due to error posting to CouchDB".format(workflowName)
+        self.sendAlert(alertName, alertSeverity, alertSummary, alertDescription,
+                       self.alertServiceName)
+        self.logger.warning(alertDescription)
+
+
     def notifyLargeData(self, aboveWarningThreshold, transferId, wflowName, dataSize, dataIn):
```
amaltaro, line 794: Kenyi, since you will have to make some changes to this PR, can I ask you to provide an enhancement for this function as well? :-D
Could you please rename it to `alertLargeInputData`? Just so it's more consistent with the other alerts.

amaltaro (changes requested):
These changes are looking much better Kenyi! I left a few comments along the code for your consideration though (I think I made single comments instead of a review, sorry!).

khurtado:
> These changes are looking much better Kenyi! I left a few comments along the code for your consideration though (I think I made single comments instead of a review, sorry!).

Alright! Just made the changes.

amaltaro (approved):

amaltaro:
@khurtado this is looking good now. Can you please squash those 3 commits into a single one? Thanks

khurtado:
@amaltaro Done!

amaltaro:
Thanks Kenyi!

PR #10961, opened 2022-01-26
DBS3Upload: check error message from the response body attribute

### LINKED ISSUE #10960: DBS3Upload not properly parsing error messages for blocks already inserted
**Impact of the bug**
WMAgent

**Describe the bug**
While validating the latest candidate WMAgent release: 1.5.7, I noticed that DBS3Upload is failing to identify that a given block has already been inserted into DBS server, thus retrying to insert that same block infinitely. Error code can be found at [1]

**How to reproduce it**
Mark an already inserted block as not in DBS, such that DBS3Upload will attempt to insert it in the next cycle.

**Expected behavior**
DBS3Upload component should parse the HTTP response object and properly identify that that block has already been inserted into DBS server, thus proceeding with the normal workflow of block injection into the server (marking that block as inserted in the relational database).

[1]
```
2022-01-26 21:54:51,636:139908283959040:INFO:DBSUploadPoller:About to insert block /RelValProdMinBias/Agent157_Val-Agent157_Val_OLD_Alanv3-v22/GEN-SIM#5927f844-ecd8-4a18-9eca-3f4a1a37567f
2022-01-26 21:54:51,702:139908283959040:ERROR:DBSUploadPoller:Error trying to process block /RelValProdMinBias/Agent157_Val-Agent157_Val_OLD_Alanv3-v22/GEN-SIM#5927f844-ecd8-4a18-9eca-3f4a1a37567f through DBS. Error: HTTP Error 400: Bad R
equest
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-client/4.0.5/lib/python3.8/site-packages/dbs/apis/dbsClient.py", line 488, in __parseForException
    raise HTTPError(http_error.url, data['exception'], data['message'], http_error.header, http_error.body)
RestClient.ErrorHandling.RestClientExceptions.HTTPError: HTTP Error 400: DBSBlockInsert/insertBlock. Block /RelValProdMinBias/Agent157_Val-Agent157_Val_OLD_Alanv3-v22/GEN-SIM#5927f844-ecd8-4a18-9eca-3f4a1a37567f already exists.

  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/wmagentpy3/1.5.7/lib/python3.8/site-packages/WMComponent/DBS3Buffer/DBSUploadPoller.py", line 85, in uploadWorker
    dbsApi.insertBulkBlock(blockDump=block)
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-client/4.0.5/lib/python3.8/site-packages/dbs/apis/dbsClient.py", line 605, in insertBulkBlock
    result =  self.__callServer("bulkblocks", data=blockDump, callmethod='POST' )
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-client/4.0.5/lib/python3.8/site-packages/dbs/apis/dbsClient.py", line 448, in __callServer
    self.__parseForException(http_error)
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-client/4.0.5/lib/python3.8/site-packages/dbs/apis/dbsClient.py", line 490, in __parseForException
    raise http_error
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-client/4.0.5/lib/python3.8/site-packages/dbs/apis/dbsClient.py", line 445, in __callServer
    self.http_response = method_func(self.url, method, params, data, request_headers)
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-pycurl/3.17.7/lib/python3.8/site-packages/RestClient/RestApi.py", line 42, in post
    return http_request(self._curl)
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/py3-dbs3-pycurl/3.17.7/lib/python3.8/site-packages/RestClient/RequestHandling/HTTPRequest.py", line 62, in __call__
    raise HTTPError(effective_url, http_code, http_response.msg, http_response.raw_header, http_response.body)
RestClient.ErrorHandling.RestClientExceptions.HTTPError: HTTP Error 400: Bad Request
```

amaltaro:
Fixes #10960 

#### Status
ready

#### Description
It isn't clear why we have this new behaviour now, but it could be from the new version of dbs3-client/pycurl.

When looking up at the error message, use the `body` attribute of the HTTP response object, which would have something like:
```
{"exception": 400, "message": "DBSBlockInsert/insertBlock. Block /RelValProdMinBias/Agent157_Val-Agent157_Val_OLD_Alanv3-v22/GEN-SIM#d6eb571b-f65e-4fb3-af49-ceb16b55a14e already exists.", "type": "HTTPError"}
```

different than the stringyfication of the exception object, which was giving us only this message:
```
400: Bad Request
```

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
@@ -85,7 +85,7 @@ def uploadWorker(workInput, results, dbsUrl):
             dbsApi.insertBulkBlock(blockDump=block)
             results.put({'name': name, 'success': "uploaded"})
         except Exception as ex:
-            exString = str(ex)
+            exString = str(getattr(ex, "body", ex))
             if 'Block %s already exists' % name in exString:
                 # Then this is probably a duplicate
                 # Ignore this for now


amaltaro:
This patch has been already tested in testbed (WMA 1.5.7) and it works like a charm.

vkuznet (approved):
Even though the change is trivial I doubt it is correct. You should never rely on body to treat the errors. The typical example is this. In DBS Python server the exceptions send as dicts. In Go Server all results, including errors are send as lists. Moreover, the server should be free to change its error representation at any time which implies that client should be able to treat errors correctly. The Body part is only for information and information can be changed over time. Therefore, the code may break depending how you read the body, e.g. if exception contains word "body" or not (for example in Go there is no concept of exceptions even though I followed DBS Python way to create error messages). The more reliable way to check for errors is to read HTTP code instead of the body. Even though I'll approve this change I suggest to pay attention to such exception handling. I understand that it may require lots of changes in different places to use HTTP code instead of body, but this is correct way to do it.

amaltaro:
Thanks for this review, Valentin. The big problem here is that HTTP error code is `400`, nothing else. And this is the same error code that we would get for different problems.
What I understand from your message is that, in order to have a solid fix for this, we should agree on a set of error codes that can be returned from the code, and adapt our applications for that. Is that correct? If so, I fully support it, but as you said, it's probably something for the future.

vkuznet:
@amaltaro , yes, we should use different HTTP error codes, but their set is limited. You can easily see all of them [here](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). What probably should be done is standardization of DBS error codes and DBS error message(s). For instance, each DBS error message should have:
- DBS error code
- DBS message
- DBS api
etc.

With such convention we can come up with standard parser on a client side such that application will know exactly what each error message is and how to parse it. I already have some structure in new DBS server and it can be improved further once we made a decision about standard error format/codes.

amaltaro:
Ok, I suggest we follow this up on this issue that I just created: https://github.com/dmwm/WMCore/issues/10962
I do not think we have room to deal with it this quarter though, so I have already pre-assigned it for the next quarter.

todor-ivanov (approved):
Thanks @amaltaro 
The change indeed looks trivial and I have nothing more to add other than what has already been discussed with Valentin.

Issue #10954, opened 2022-01-24
Jenkins: remove Python2 pylint and unit tests from our CI setup


amaltaro:
**Impact of the new feature**
Jenkins for WMCore

**Is your feature request related to a problem? Please describe.**
As confirmed in our team meeting today - and announced in the O&C weekly chat on Jan 12th - WMCore support to python2 will be stopped starting this year, thus we should update Jenkins accordingly.

**Describe the solution you'd like**
The expected outcome for this issue is the following:
* removal of the Python2 unit tests against master/HEAD
* removal of the Python2 unit tests executed for PRs
* removal of the Python2 pylint checks
* update of the WMCore documentation (development workflow and jenkins instructions).

**Describe alternatives you've considered**
None

**Additional context**
None

PR #10963, opened 2022-01-27
Aggregate DBS results from Go server

### LINKED ISSUE #10956: Use of DBS filelumis RESTful API needs to be made compatible
**Impact of the new feature**
MSTransferor, but there could be other parts of the code (including DBS3Reader)

While debugging why our LumiList-based templates don't manage to pass the `assigned` request status, I noticed that MSTransferor is actually failing to run the lumi list comparison. The reason for that is that Go and the Python implementation of the `filelumis` RESTful API yield data in different format.

I believe this is only a problem because we make a direct call to the REST endpoint, without using the dbs3-client. Nonetheless, it needs to be verified as well.

In MSTransferor, this line is actually raising an exception:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/MSTransferor/RequestInfo.py#L409

as one can interactive check with:
```
>>> fileMask = LumiList(runsAndLumis={'123': 333})
  File "/data/srv/wmagent/v1.5.7/sw/slc7_amd64_gcc630/cms/wmagentpy3/1.5.7/lib/python3.8/site-packages/WMCore/DataStructs/LumiList.py", line 96, in __init__
    for lumi in sorted(int(l) for l in lumiList):
TypeError: 'int' object is not iterable
```

The MSTransferor code needs to be updated such that it can deal with both response formats, allowing for a transparent migration of the DBS server. Here is the location calling `filelumis`:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Tools/Common.py#L247

we still need to investigate whether this very same DBS API is called in different places, and we need to ensure that our code will behave properly, regardless of the DBSServer flavor.

In addition to that, we should also investigate whether any changes are required for the `DBS3Reader` module.

**Additional context**
The python-based server does an aggregation of the lumis grouped by run+lfn, example:
https://cmsweb.cern.ch/dbs/prod/global/DBSReader/filelumis?validFileOnly=1&block_name=/DoubleMuon/Run2018D-v1/RAW%23590119e4-3060-4c9b-9b38-2b16f6844bab

while the Go implementation yields a row for each lfn+run+lumi information:
https://cmsweb-testbed.cern.ch/dbs/int/global/DBSReader/filelumis?validFileOnly=1&block_name=/DoubleMuon/Run2018D-v1/RAW%23590119e4-3060-4c9b-9b38-2b16f6844bab

vkuznet:
Fixes #10956

#### Status
ready

#### Description
Use DBSClient aggregators to aggregate relevant parts from DBS Go server APIs.

YES

#### Related PRs

new changes depends on DBSClient. If it is not desired we can take appropriate functions from it and put to WMCore. So far changes are minimal and since WMCore relies on DBSClient it should not be a problem with this dependency.

### DIFF
--- a/src/python/WMCore/MicroService/Tools/Common.py
+++ b/src/python/WMCore/MicroService/Tools/Common.py
@@ -25,11 +25,30 @@ from Utils.CertTools import getKeyCertFromEnv
 from WMCore.Services.pycurl_manager import RequestHandler
 from WMCore.Services.pycurl_manager import getdata as multi_getdata
 
+# DBS agregators
+from dbs.apis.dbsClient import aggRuns, aggFileLumis
+
 
 # static variables
 STEP_PAT = re.compile(r'Step[0-9]')
 TASK_PAT = re.compile(r'Task[0-9]')
 
+
+def isEmptyResults(row):
+    """
+    _isEmptyResults_
+
+    Evaluates whether row data contains empty result set
+    :return: bool
+    """
+    if 'data' not in row:
+        raise Exception("provided result dict does not contain 'data' key")
+    # if code is not present in row it means it was success (HTTP status code 200)
+    code = int(row.get('code', 200))
+    data = row['data']
+    if (code >= 200 and code < 400) and data in (None, []):
+        return True
+    return False
 
 def getMSLogger(verbose, logger=None):
     """
@@ -63,7 +82,7 @@ def dbsInfo(datasets, dbsUrl):
 
     for row in data:
         dataset = row['url'].split('=')[-1]
-        if row['data'] is None:
+        if isEmptyResults(row):
             print("FAILURE: dbsInfo for %s. Error: %s %s" % (dataset, row.get('code'), row.get('error')))
             continue
         rows = json.loads(row['data'])
@@ -232,7 +251,7 @@ def getBlocksByDsetAndRun(datasetName, runList, dbsUrl):
 
     for row in data:
         dataset = row['url'].rsplit('=')[-1]
-        if row['data'] is None:
+        if isEmptyResults(row):
             msg = "Failure in getBlocksByDsetAndRun for %s. Error: %s %s" % (dataset,
                                                                              row.get('code'),
                                                                              row.get('error'))
@@ -261,12 +280,13 @@ def getFileLumisInBlock(blocks, dbsUrl, validFileOnly=1):
 
     for row in data:
         blockName = unquote(row['url'].rsplit('=')[-1])
-        if row['data'] is None:
+        if isEmptyResults(row):
             msg = "Failure in getFileLumisInBlock for block %s. Error: %s %s" % (blockName,
                                                                                  row.get('code'),
                                                                                  row.get('error'))
             raise RuntimeError(msg)
         rows = json.loads(row['data'])
+        rows = aggFileLumis(rows) # adjust to DBS Go server output
         runLumisByBlock.setdefault(blockName, [])
         for item in rows:
             runLumisByBlock[blockName].append(item)
@@ -288,7 +308,7 @@ def findBlockParents(blocks, dbsUrl):
     for row in data:
         blockName = unquote(row['url'].rsplit('=')[-1])
         dataset = blockName.split("#")[0]
-        if row['data'] is None:
+        if isEmptyResults(row):
             print("Failure in findBlockParents for block %s. Error: %s %s" % (blockName,
                                                                               row.get('code'),
                                                                               row.get('error')))
@@ -322,12 +342,13 @@ def getRunsInBlock(blocks, dbsUrl):
     data = multi_getdata(urls, ckey(), cert())
     for row in data:
         blockName = unquote(row['url'].rsplit('=')[-1])
-        if row['data'] is None:
+        if isEmptyResults(row):
             msg = "Failure in getRunsInBlock for block %s. Error: %s %s" % (blockName,
                                                                             row.get('code'),
                                                                             row.get('error'))
             raise RuntimeError(msg)
         rows = json.loads(row['data'])
+        rows = aggRuns(rows) # adjust to DBS Go server output
         runsByBlock[blockName] = rows[0]['run_num']
     return runsByBlock
 
@@ -366,7 +387,7 @@ def eventsLumisInfo(inputs, dbsUrl, validFileOnly=0, sumOverLumi=0):
     data = multi_getdata(urls, ckey(), cert())
     for row in data:
         data = unquote(row['url'].split('=')[-1])
-        if row['data'] is None:
+        if isEmptyResults(row):
             print("FAILURE: eventsLumisInfo for %s. Error: %s %s" % (data,
                                                                     row.get('code'),
                                                                     row.get('error')))
@@ -585,7 +606,7 @@ def findParent(datasets, dbsUrl):
 
     for row in data:
         dataset = row['url'].split('=')[-1]
-        if row['data'] is None:
+        if isEmptyResults(row):
             print("Failure in findParent for dataset %s. Error: %s %s" % (dataset,
                                                                           row.get('code'),
                                                                           row.get('error')))
--- a/test/python/WMCore_t/MicroService_t/Tools_t/Common_t.py
+++ b/test/python/WMCore_t/MicroService_t/Tools_t/Common_t.py
@@ -7,7 +7,8 @@ from __future__ import division, print_function
 
 import unittest
 
-from WMCore.MicroService.Tools.Common import dbsInfo, getEventsLumis, findParent
+from WMCore.MicroService.Tools.Common import \
+        dbsInfo, getEventsLumis, findParent, isEmptyResults
 
 
 class CommonTest(unittest.TestCase):
@@ -47,6 +48,19 @@ class CommonTest(unittest.TestCase):
         parents = findParent(self.child, self.dbsUrl)
         self.assertEqual(parents[self.child[0]], '/SingleElectron/Run2016B-v2/RAW')
 
+    def test_isEmptyResults(self):
+        row = {'code': 200, 'data': None}
+        result = isEmptyResults(row)
+        self.assertEqual(result, True)
+        row = {'code': 200, 'data': []}
+        result = isEmptyResults(row)
+        self.assertEqual(result, True)
+        row = {'code': 400, 'data': None}
+        result = isEmptyResults(row)
+        self.assertEqual(result, False)
+        row = {'code': 400, 'data': []}
+        result = isEmptyResults(row)
+        self.assertEqual(result, False)
 
 if __name__ == '__main__':
     unittest.main()


vkuznet:
@amaltaro , I tested these changes manually in stand-along script using both Py and Go server. For completeness below you can find imy stand-along test script I used for testing. There are only two APIs which required aggregation `/runs` and `/filelumis`. All others APIs used by `multi_getdata` do not require aggregation.

In the script below, I manually check aggregated results for specific block/lfn/lumis/runs. Since produced results are `dict` type and aggregated lists are not ordered only `runs` tests can be compared in a code, while `file_lumis` one needs to be checked manually (this is what I did):
```
#!/usr/bin/env python3
#-*- coding: utf-8 -*-
import json
from urllib.parse import quote, unquote

# WMCore modules
from WMCore.MicroService.Tools.Common import ckey, cert

# DBS agregatros
from dbs.apis.dbsClient import aggRuns, aggFileLumis


def test_file_lumis(dbsUrl, blocks):
    validFileOnly = 1
    runLumisByBlock = {}
#     urls = ['%s/filelumis?validFileOnly=%d&block_name=%s' % (dbsUrl, validFileOnly, quote(b)) for b in blocks]
    urls = ['%s/filelumis?block_name=%s' % (dbsUrl, quote(b)) for b in blocks]
    # limit it to 10 concurrent calls not to overload DBS
    data = multi_getdata(urls, ckey(), cert(), num_conn=10)

    for row in data:
        blockName = unquote(row['url'].rsplit('=')[-1])
        if row['data'] is None:
            msg = "Failure in getFileLumisInBlock for block %s. Error: %s %s" % (blockName,
                                                                                 row.get('code'),
                                                                                 row.get('error'))
            raise RuntimeError(msg)
        rows = json.loads(row['data'])
        rows = aggFileLumis(rows) # adjust to DBS Go server outpu
        runLumisByBlock.setdefault(blockName, [])
        for item in rows:
            runLumisByBlock[blockName].append(item)
    return runLumisByBlock

def test_runs(dbsUrl, blocks):

    runsByBlock = {}
    urls = ['%s/runs?block_name=%s' % (dbsUrl, quote(b)) for b in blocks]

    for row in data:
        blockName = unquote(row['url'].rsplit('=')[-1])
        if row['data'] is None:
            msg = "Failure in getRunsInBlock for block %s. Error: %s %s" % (blockName,
                                                                            row.get('code'),
                                                                            row.get('error'))
            raise RuntimeError(msg)
        rows = json.loads(row['data'])
        rows = aggRuns(rows) # adjust to DBS Go server outpu
        runsByBlock[blockName] = rows[0]['run_num']
    return runsByBlock

def main():
    dbsUrlPy = "https://cmsweb.cern.ch/dbs/prod/global/DBSReader"
    dbsUrlGo = "https://cmsweb-testbed.cern.ch/dbs/int/global/DBSReader"
    blocks = [
        "/ZeroBias/Run2010B-Dec4ReReco_v1/AOD#0c30731e-8447-49bb-998f-2df1e4501bab",
        "/ZeroBias/Run2010B-Dec4ReReco_v1/AOD#738f25ea-c5ec-4db2-8ac4-a4a257bf77f1",
        "/ZeroBias/Run2010B-Dec4ReReco_v1/AOD#8c5bff23-fc9a-4d83-8eb2-456ceb267cc0"
    ]

    print("compare results from runs API")
    resPy = test_runs(dbsUrlPy, blocks)
    resGo = test_runs(dbsUrlGo, blocks)
    if resPy != resGo:
        print("results from DBS servers are incompatible")

    print("compare results from filelumis API")
    resPy = test_file_lumis(dbsUrlPy, blocks)
    print("results from DBS Py server")
    print(resPy)
    resGo = test_file_lumis(dbsUrlGo, blocks)
    print("results from DBS Go server")
    print(resGo)
    if resPy != resGo:
        print("results from DBS servers are incompatible")

if __name__ == '__main__':
    main()

```

vkuznet:
During these tests I also found that I need to fix DBS Go server to handle `validFileOnly` flag properly, see https://github.com/dmwm/dbs2go/issues/3

amaltaro (changes requested):
Thanks for providing these changes, Valentin. I wasn't sure whether we should import those functions from DBS itself, but it seems to be the best option and we keep the same implementation in multiple places.

Regarding the unit test failures, it looks like you introduce some funky chars in this PR:
```
Exception Class: WMInitException
Message: Critical error while attempting to delete entire DB!
(exceptions.LookupError) unknown encoding: utf8mb4
[SQL: SELECT CAST('test plain returns' AS CHAR(60)) AS anon_1]Traceback (most recent call last):
```

if you need further instructions to investigate the unit tests results, please refer to: https://github.com/dmwm/WMCore/wiki/Understanding-Jenkins#unit-tests

There is one missing part in this PR as well, concerning the comment I made in the GH issue itself (None vs []):
https://github.com/dmwm/WMCore/issues/10956#issuecomment-1022623983

vkuznet:
@amaltaro my code changes seems very trivial, and I have no idea where you get the unit tests failure you pointed out, I scan Py3 Unit tests and don't see message you provide. I look at Jenkins and don't see my PR there. Please provide specific link how you grab the error (sorry I need some guidance with Jenkins as I didn't use them often).

amaltaro:
@vkuznet Valentin, from the wiki I pointed out, this is the important section:
"""
If you do have a failure and want to see the unit test output

    The detailed report will tell you which tests failed, but not the details of how
    If you click on "Details" under the default status box/row
    then click on Test Result
    and finally find the failing test you are interested in and click on it. This will give you all the info that nose captured while it was running your test.
"""

just ping me on slack if you want to walk through those together though.

vkuznet:
@amaltaro , sorry I re-read this paragraph multiple times and it does not clarify anything for me. Which Details link it referring to? on github? or on jenkins? On github the details point to Py3 unit tests [page](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12734/artifact/artifacts/PullRequestReport.html#unittestspy3). On Jenkins there is no Details if I visit https://cmssdt.cern.ch/dmwm-jenkins/view/DMWM/. Sorry but documentation should provide more details about how to access the info. I'm on Slack now and unfortunately you are not. Let's try to converge on Slack.

vkuznet:
After I figure out how to access proper Jenkins build page I see failed test, but I have no idea how to find those which are new. Can it be done or should developer come at every single one to scan for errors? The page reports (74 failures / +3) which I read as 3 new failed tests, but I have no idea which ones.

For example, if I click on first one (WMCore_t.WMBS_t.Job_t.JobTest.testJobCacheDir) I see this error:
```
root: DEBUG: Using SQLAlchemy v.1.3.3
root: INFO: Instantiating base WM DBInterface
root: ERROR: Critical error while attempting to delete entire DB!
(exceptions.LookupError) unknown encoding: utf8mb4
[SQL: SELECT CAST('test plain returns' AS CHAR(60)) AS anon_1]Traceback (most recent call last):
```
but from test details I see that it is python 2.7 build, should we care? The tests itself is not related `Common.py` module I'm changing. Therefore I can conclude that this failure has nothing to do with changes I provide.

The only useful insight comes from [Py3 Unit tests](https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12734/artifact/artifacts/PullRequestReport.html#unittestspy3) page where in Python2 pylint summary it claims:
```
E0611, line 29: No name 'aggRuns' in module 'dbs.apis.dbsClient'
E0611, line 29: No name 'aggFileLumis' in module 'dbs.apis.dbsClient'
```
which tells me that DBSClient version is wrong since it does not provide this functions. Which brings me to the question which env is used in Unit tests and how I can verify that DBSClient there is up to date?

vkuznet:
Scanning further, I see another [test](https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/12734/testReport/junit/nose.failure/Failure/runTest/) which clearly indicates the issue:
```
  File "/build/cmsbld/jenkins/workspace/DMWM-PR-unittests/SLICE/0/label/cms-dmwm-cc7/deploy/1.5.7.pre4-comp/sw/slc7_amd64_gcc630/external/python/2.7.13-comp2/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
....
  File "/build/cmsbld/jenkins/workspace/DMWM-PR-unittests/SLICE/0/label/cms-dmwm-cc7/code/src/python/WMCore/MicroService/Tools/Common.py", line 29, in <module>
    from dbs.apis.dbsClient import aggRuns, aggFileLumis
ImportError: cannot import name aggRuns
```
So, there are two issues here:
- the unit test shows it uses py2.7 build and as such new DBSClient is not compatible
- the test fail to load aggregated functions which means that Jenkins env does not have proper version of DBSClient

I don't know how to setup env in Jenkins and how to force through github PR that it should use new version of DBSClient, but this seems to be real issue with failed Py2/Py3 unit tests. The Py2 unit tests will fail by default since new DBSClient is only python3 based, while Py3 unit tests fail because it does not have new version of DBSClient.

@amaltaro , please let me know how to resolve this issue and setup proper env in Jenkins. Or, I can take DBSClient codebase and put into WMCore (and this will eliminate loading DBSClient module), but this will lead to code duplication in WMCore and DBSClient repos.

amaltaro:
@vkuznet that's a good point! This PR looks good in terms of Python3 checks:
```
Python3 Unit tests: succeeded
- 2 changes in unstable tests
```
and the failures are coming from the Python2 stack, which we did not update to the latest dbs3-client, since those developments were only made for Python3 stack.

I think it's better if I get this issue sorted our right away: https://github.com/dmwm/WMCore/issues/10954
to allow people to focus on what matters :-D

amaltaro:
Valentin, I decided to test this patch in my VM and I noticed that MSTransferor was no longer able to serve HTTP requests:
```
[28/Jan/2022:00:46:41]  SERVER OTHER ERROR __builtins__.AttributeError 3b7f9bc697fb2f877e0026c1d0c50dc8 ('Data' object has no attribute 'mgr')
[28/Jan/2022:00:46:41]    Traceback (most recent call last):
[28/Jan/2022:00:46:41]      File "/data/srv/HG2201e/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.5.7.pre5/lib/python3.8/site-packages/WMCore/REST/Server.py", line 728, in default
[28/Jan/2022:00:46:41]        return self._call(RESTArgs(list(args), kwargs))
[28/Jan/2022:00:46:41]      File "/data/srv/HG2201e/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.5.7.pre5/lib/python3.8/site-packages/WMCore/REST/Server.py", line 809, in _call
[28/Jan/2022:00:46:41]        obj = apiobj['call'](*safe.args, **safe.kwargs)
[28/Jan/2022:00:46:41]      File "/data/srv/HG2201e/sw/slc7_amd64_gcc630/cms/reqmgr2ms/0.5.7.pre5/lib/python3.8/site-packages/WMCore/MicroService/Service/Data.py", line 90, in get
[28/Jan/2022:00:46:41]        'microservice': self.mgr.__class__.__name__}
[28/Jan/2022:00:46:41]    AttributeError: 'Data' object has no attribute 'mgr'
```

it turns out that, this PR adds a new dependency to ReqMgr2MS (all microservices). So, the spec file also needs to be updated with this new dependency:
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/reqmgr2ms.spec

vkuznet:
@amaltaro I'm not sure I understand the reason why adding functions defined in DBSClient to the code brings dependency on ReqMgr2MS for the code. It seems to me that dependency chain is not properly set among RPMs. At most the changes should depend on py3-pycurl (because DBSClient uses it) and nothing else.

amaltaro:
This import:
```
 from dbs.apis.dbsClient import aggRuns, aggFileLumis
```

comes from the DBSClient repo:
https://github.com/dmwm/DBSClient/blob/main/src/python/dbs/apis/dbsClient.py#L109

Hence, we need to add `py3-dbs3-client` to the `Requires` dependency of reqmgr2ms.spec file.

amaltaro:
test this please

vkuznet:
@amaltaro , I addressed `row['data'] is None` issue with additional changes. Please review.

amaltaro (changes requested):
Valentin, please see some comments along the code. It looks like I missed those PhEDEx-based APIs, they should no longer be in our baseline code.

Having a second look at this result None/[], I might have misunderstood the code, sorry. I think None is actually defined by the pycurl_manager module, when the HTTP is not successful (code 2xx, 3xx), to be verified though. That means, we should distinguish between `None` and `[]`, because `[]` is a valid response and it generally means there was no match on the server/database side, right?

vkuznet:
Alan, I pushed changes according to your review.

Now, about `None` vs `[]`. You are right about the difference, but you're wrong it is `pycurl_manager` feature. It is what backend server returns. If you want to differentiate between successful call and failure the code should use HTTP error code. For instance
- successful HTTP request will return 200 HTTP code, but it may return `None` or `[]` depending on server implementation
- failed HTTP request will return non 200 HTTP code, e.g. 400, and backend server may return you either no results or empty list.

The code you have does not differentiate between success and failure, it only looks at empty results. You should provide me more insight what code actually suppose to do. For instance,
```

    for row in data:
        dataset = row['url'].split('=')[-1]
        if isEmptyResults(row):
            print("FAILURE: dbsInfo for %s. Error: %s %s" % (dataset, row.get('code'), row.get('error')))
            continue
```
Does this snippet care about HTTP failures? Or, does it only need to check if there is no results in response? The printout is incorrecvt since it prints `FAILURE` while HTTP code may be 200.

I made changes only to match existing codebase but didn't change its logic.

vkuznet:
@amaltaro I added check for code key. Please review now.

amaltaro:
Changes look good to me, Valentin. Can you please make a unit test for that new function and commit it in a separate commit? With that, it should be ready to be merged.

vkuznet:
@amaltaro , I added unit test and slightly adjust `isEmptyResults` to account for missing `code` key. Please review again. Changes are squashed already.

src/python/WMCore/MicroService/Tools/Common.py
```diff
@@ -25,12 +25,34 @@
...
+def isEmptyResults(row):
+    """
+    _isEmptyResults_
+
+    Evaluates whether row data contains empty result set
+    :return: bool
+    """
+    if 'data' not in row:
+        raise Exception("provided result dict does not contain 'data' key")
+    code = 200 # default HTTP status code
```
amaltaro, line 46: Valentin, just a final suggestion that will make this code look better. How about replacing L46-L50 by:
```
# if code is not present in row it means it was success (HTTP status code 200)
code = int(row.get('code', 200))
```

amaltaro (commented):
@vkuznet I have made one final suggestion in the code. If you agree, please update the code and squash and it's ready to go.

PS.: For future PRs, please keep unit test changes in a separate commit.

src/python/WMCore/MicroService/Tools/Common.py
vkuznet: done

amaltaro (approved):
Thanks Valentin! It looks good to me!

amaltaro:
And I merged it without having a reference to the required cmsdist spec changes with the new dependency on dbs3-client. Which of course, was only spotted when deploying services in cmsweb-testbed.

Here is the relevant cmsdist PR (including a few other changes as well): https://github.com/cms-sw/cmsdist/pull/7600

PR #10976, opened 2022-02-02
Update requirements.txt and PyPI build scripts for py3 stack

### LINKED ISSUE #10955: Adjust build_pypi_packages.sh and requirements.txt to handle py3 stack
**Impact of the bug**
With migration to python3 we need to adjust [build_pypi_packages.sh](https://github.com/dmwm/WMCore/blob/master/tools/build_pypi_packages.sh) and its usage of `requirements.txt`  to satisfy py3 dependencies. There are two files in WMCore repo, `requirements.txt` and `requirements_py3.txt`, and `build_pypi_packages.sh` uses the former one.

**Describe the bug**
In order to use `build_pypy_packages.sh` in CI/CD builds we need to fix its usage of `requirements.txt` . I don't know which dependencies (`requirements.txt` or `requirements_py3.txt`) is a proper set and I suggest to leave now only one file. Moreover, I suggest to remove explicit versions of packages, the `==` condition, and/or replace it with `>=` ones. For example, replace `mock==4.0.3` to `mock` or `mock>=4.0.3`. This is required to resolve potential conflicts of dependency packages during pip install.

**How to reproduce it**
There is no way to reproduce it since I only point out to two different files `requirements.txt` and `requirements_py3.txt`, and ask to converge them to a single one. But, someone can use pip command to see dependency conflicts, e.g.
```
# setup python env
python3 -m venv venv

# source env
source venv/bin/activate

# update pip
pip install --upgrade pip

# install dependencies
pip install -r requirements_py3.txt
...
Collecting pycurl==7.43.0.3
  Using cached pycurl-7.43.0.3.tar.gz (215 kB)
  Preparing metadata (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /data/users/vk/tmp/venv/bin/python3 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py'"'"'; __file__='"'"'/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/valya/pip-pip-egg-info-jy14zk6u
       cwd: /tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/
  Complete output (22 lines):
    File "/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py", line 229, in configure_unix
      stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    File "/usr/lib64/python3.6/subprocess.py", line 729, in __init__
      restore_signals, start_new_session)
    File "/usr/lib64/python3.6/subprocess.py", line 1364, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'curl-config': 'curl-config'

    File "<string>", line 1, in <module>
    File "/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py", line 944, in <module>
      ext = get_extension(sys.argv, split_extension_source=split_extension_source)
    File "/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py", line 606, in get_extension
      ext_config = ExtensionConfiguration(argv)
    File "/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py", line 101, in __init__
      self.configure()
    File "/tmp/valya/pip-install-54dkfn3w/pycurl_b64a8d6a1b074e97a09c5a871124b7e7/setup.py", line 233, in configure_unix
      raise ConfigurationError(msg)
  __main__.ConfigurationError: Could not run curl-config: [Errno 2] No such file or directory: 'curl-config': 'curl-config'
  ----------------------------------------
WARNING: Discarding https://files.pythonhosted.org/packages/ac/b3/0f3979633b7890bab6098d84c84467030b807a1e2b31f5d30103af5a71ca/pycurl-7.43.0.3.tar.gz#sha256=6f08330c5cf79fa8ef68b9912b9901db7ffd34b63e225dce74db56bb21deda8e (from https://pypi.org/simple/pycurl/). Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Could not find a version that satisfies the requirement pycurl==7.43.0.3 (from versions: 7.18.2, 7.19.0, 7.19.0.1, 7.19.0.2, 7.19.0.3, 7.19.3, 7.19.3.1, 7.19.5, 7.19.5.1, 7.19.5.2, 7.19.5.3, 7.21.5, 7.43.0, 7.43.0.1, 7.43.0.2, 7.43.0.3, 7.43.0.4, 7.43.0.5, 7.43.0.6, 7.44.0, 7.44.1)
ERROR: No matching distribution found for pycurl==7.43.0.3

```

Here, we got dependency conflicts due to strict versions in `requirements_py3.txt`

**Expected behavior**
`pip install -r requirements.txt` command should work, where `requirements.txt` should be common set of dependencies (most likely for py3 we need content of `requirements_py3.txt` file).

Related to https://github.com/dmwm/WMCore/issues/10922,  https://github.com/dmwm/WMCore/issues/10921

goughes:
Fixes #10955 

#### Status
In development

#### Description
This PR updates requirements.txt to include only requirements needed for py3 WMCore services and the new `wmagent-devtools` package. It also updates build_pypi_packages.sh to build the packages.

* Updated operators from `==` to `~=`.
* Removed `dbs3-pycurl` since it is a dependency of `dbs3-client`. 
* Removed `dbs3-client` as a dependency for wmagent-devtools since it was not in the [spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/wmcorepy3-devtools.spec).
* Added `gfal2-python` for reqmgr2ms since it was in the [spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/reqmgr2ms.spec)
* Added `mysqlclient` for wmagent since it was in the [spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/py3-mysqlclient.spec) 

NO, cannot build py2 PyPI packages for any WMCore service after merge.

#### Related PRs
None

None

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,32 +1,43 @@
 # All dependencies needed to run WMCore software
+#
 # This file is parsed by tools/build_pypi_packages.sh to create requirements.txt
 # files for each piece of software built from WMCore and uploaded to PyPI
 # Format:
 # PackageName==X.Y.Z          # <comma separated list of WMCore software needing the package>
+#
+# Different version specifiers are used to ensure compatibility and to avoid potential python conficts
+# == when exact version of the package is required
+# >= when a newer version of the package is required
+# ~= when we need compatibility of the version, e.g. stay within major version of the package
+# for more details please refer to official Python documentation, see
+# https://www.python.org/dev/peps/pep-0440/#version-specifiers
 
-Cheetah==2.4.0                # wmagent,reqmgr2,reqmon
-CherryPy==17.4.0              # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
-CMSMonitoring>=0.3.4          # wmagent,reqmgr2,reqmon,reqmgr2ms
-coverage==4.5.4               # wmagent
-cx-Oracle==5.2.1              # wmagent
-dbs-client==3.13.1            # wmagent,reqmgr2,reqmon,global-workqueue
-decorator==3.4.2              # wmagent
-funcsigs==1.0.2               # wmagent
-future==0.18.2                # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
-httplib2==0.19.0              # wmagent,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
-Markdown==3.0.1               # wmagent
-mock==2.0.0                   # wmagent
-MySQL-python==1.2.5           # wmagent
-nose2==0.9.2                  # wmagent
-psutil==5.6.6                 # wmagent,reqmgr2,reqmon,global-workqueue
-pycurl==7.43.0.3              # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
-pymongo==3.10.1               # reqmgr2ms
-pyOpenSSL==18.0.0             # wmagent
-pyzmq==17.1.2                 # wmagent
-retry==0.9.1                  # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
-rucio-clients==1.23.0         # wmagent,global-workqueue,reqmgr2ms
-setuptools==39.2.0            # wmagent
-Sphinx==1.3.5                 # wmagent,reqmgr2,acdcserver,global-workqueue,reqmgr2ms
-SQLAlchemy==1.3.3             # wmagent
-stomp.py==4.1.15              # wmagent
-CMSCouchapp==1.2.10           # wmagent
+Cheetah3~=3.2.6.post2         # wmagent,reqmgr2,reqmon
+CherryPy~=17.4.0              # wmagent,wmagent-devtools,reqmgr2,reqmon,global-workqueue,reqmgr2ms
+CMSCouchapp~=1.3.4            # wmagent
+CMSMonitoring~=0.3.4          # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
+coverage~=5.4                 # wmagent,wmagent-devtools
+cx-Oracle~=7.3.0              # wmagent
+dbs3-client~=4.0.7            # wmagent,reqmgr2,reqmon,global-workqueue
+future~=0.18.2                # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
+gfal2-python~=1.11.0.post3    # reqmgr2ms
+httplib2~=0.19.0              # wmagent,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
+htcondor~=8.9.7               # wmagent
+Jinja2~=3.0.1                 # wmagent
+memory-profiler~=0.58.0       # wmagent-devtools
+mock~=4.0.3                   # wmagent,wmagent-devtools
+mox3~=1.1.0                   # wmagent-devtools
+mysqlclient~=2.0.3            # wmagent
+nose~=1.3.7                   # wmagent-devtools
+nose2~=0.10.0                 # wmagent-devtools
+pep8~=1.7.1                   # wmagent-devtools
+psutil~=5.8.0                 # wmagent,wmagent-devtools,reqmgr2,reqmon,global-workqueue
+pycurl~=7.43.0.6              # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
+pylint~=2.7.0                 # wmagent-devtools
+pymongo~=3.10.1               # wmagent-devtools,reqmgr2ms
+pyOpenSSL~=18.0.0             # wmagent
+pyzmq~=19.0.2                 # wmagent
+retry~=0.9.2                  # wmagent,wmagent-devtools,reqmgr2,reqmon,global-workqueue,reqmgr2ms
+rucio-clients~=1.25.5         # wmagent,global-workqueue,reqmgr2ms
+Sphinx~=4.0.3                 # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
+SQLAlchemy~=1.3.3             # wmagent
--- a/tools/build_pypi_packages.sh
+++ b/tools/build_pypi_packages.sh
@@ -14,7 +14,7 @@ set -x
 # package passed as parameter, can be one of PACKAGES or "all"
 TOBUILD=$1
 # list of packages that can be built and uploaded to pypi
-PACKAGES="wmagent wmcore reqmon reqmgr2 reqmgr2ms global-workqueue acdcserver"
+PACKAGES="wmagent wmagent-devtools wmcore reqmon reqmgr2 reqmgr2ms global-workqueue acdcserver"
 PACKAGE_REGEX="^($(echo $PACKAGES | sed 's/\ /|/g')|all)$"
 
 if [[ -z $TOBUILD ]]; then
--- a/setup_dependencies.py
+++ b/setup_dependencies.py
@@ -62,6 +62,28 @@ dependencies = {
                     'etc+'
                     ],
     },
+    'wmagent-devtools': {
+        'packages': ['WMCore+',
+                     'WMComponent+',
+                     'WMQuality+',
+                     'PSetTweaks+',
+                     'Utils+'],
+        'modules': [],
+        'systems': [],
+        'statics': ['src/couchapps+',
+                    'src/css+',
+                    'src/html+',
+                    'src/javascript+',
+                    'src/templates+',
+                    'bin+',
+                    'deploy+',
+                    'doc+',
+                    'etc+',
+                    'test+',
+                    'standards+'
+                    'tools+'
+                    ],
+    },
     'reqmgr2': {
         'packages': ['WMCore.ReqMgr+',
                      'WMCore.Services+',


tools/build_pypi_packages.sh
```diff
@@ -14,7 +14,7 @@ set -x
 # package passed as parameter, can be one of PACKAGES or "all"
 TOBUILD=$1
 # list of packages that can be built and uploaded to pypi
-PACKAGES="wmagent wmcore reqmon reqmgr2 reqmgr2ms global-workqueue acdcserver"
+PACKAGES="wmagent wmagent-devtools wmcore reqmon reqmgr2 reqmgr2ms global-workqueue acdcserver"
```
amaltaro, line 17: Did you want to specify `wmcorepy3-devtools` here? And I think we should also replace `wmagent` by `wmagentpy3`.

We could discuss whether we should convert all our Py3 based changes to the standard packages that we used to have in Py2, but I guess that can wait for a future opportunity.

requirements.txt
```diff
@@ -4,29 +4,32 @@
...
-pyOpenSSL==18.0.0             # wmagent
-pyzmq==17.1.2                 # wmagent
-retry==0.9.1                  # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
-rucio-clients==1.23.0         # wmagent,global-workqueue,reqmgr2ms
-setuptools==39.2.0            # wmagent
-Sphinx==1.3.5                 # wmagent,reqmgr2,acdcserver,global-workqueue,reqmgr2ms
-SQLAlchemy==1.3.3             # wmagent
-stomp.py==4.1.15              # wmagent
-CMSCouchapp==1.2.10           # wmagent
+Cheetah3~=3.2.6.post2         # wmagent,reqmgr2,reqmon
```
amaltaro, line 7: Erik, is this list exactly the same one that we had in the requirements_py3.txt file? Can you please confirm whether the only differences are:
* a sorted list now
* and the version comparison specifier
?

amaltaro (changes requested):
Erik, besides those comments/questions along the code, can you please also update this file:
https://github.com/dmwm/WMCore/blob/master/CONTRIBUTING.rst
?

It mentions requirements_py3.txt file (and I just noticed that the first paragraph has wrongly formatted inner bullet points, if you could fix those as well, it would be great!).

Even though these are the most flexible changes, I fear that we lose the ability to reproduce the exact same environment with these changes. So, it's a tradeoff of avoiding incompatibilities versus no longer guaranteeing the exact same environment. Right?

requirements.txt
```diff
@@ -4,29 +4,32 @@
...
+Jinja2~=3.0.1                 # wmagent
+memory-profiler~=0.58.0       # wmagent-devtools
+mock~=4.0.3                   # wmagent,wmagent-devtools
+mox3~=1.1.0                   # wmagent-devtools
+mysqlclient~=2.0.3            # wmagent
+nose~=1.3.7                   # wmagent-devtools
+nose2~=0.10.0                 # wmagent-devtools
+pep8~=1.7.1                   # wmagent-devtools
+psutil~=5.8.0                 # wmagent,wmagent-devtools,reqmgr2,reqmon,global-workqueue
+pycurl~=7.43.0.3              # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
```
amaltaro, line 27: We should update this pycurl version if possible, and keep it in sync with dbs3-client, but we better have another issue to track some of those updates. Just saying it now to share this and make sure others can remember to get back to this :-D

tools/build_pypi_packages.sh
goughes: Hmm, I thought about this. I don't know how I feel about making a PyPI package name with the python version in it. Perhaps we can use these "Classifiers" instead to specify what version of Python is required? https://pypi.org/classifiers/

Doing it now I don't think would cause an issue since we don't use the current py2 PyPI packages for anything (like wmagent).

requirements.txt
goughes: There are some additional changes:
* Removed `dbs3-pycurl` since it is a dependency of `dbs3-client`. 
* Removed `dbs3-client` as a dependency for wmagent-devtools since it was not in the [spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/wmcorepy3-devtools.spec).
* Added `gfal2-python` for reqmgr2ms since it was in the [spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/reqmgr2ms.spec)
* Added `mysqlclient` for wmagent since it was in the [spec](https://github.com/cms-sw/cmsdist/blob/comp_gcc630/py3-mysqlclient.spec) 

I will add these changes to the PR description as well.

requirements.txt
goughes: I fixed it. :)

requirements.txt
amaltaro: Wonderful! Thanks for catching those up!

tools/build_pypi_packages.sh
amaltaro: Oh, that's a good point! I somehow thought that this list was referring to the spec names that we have in cmsdist.
If it's only about the packages that we upload to PyPi, I completely agree with you.

In case you need an example of classifiers, here is what I did for the CMSCouchApps PyPi package:
[github.comamaltaro/CMSCouchApps](https://github.com/amaltaro/couchapp/blob/master/setup.py#L31)

vkuznet:
To avoid potentials questions in a future, and to remind ourselves I suggest to add proper explanation into requirements.txt about all possible combinations `==`, `>=` and `~=`. I also advise to add proper references to python documentation about each rule. For instance:
```
# we use different version specifiers to avoid potential python conficts
# == when exact version of the package is required
# >= when specific or new version of the package is required
# ~= when we need compatibility of the version, e.g. stay within major version of the package
# for more details please refer to official Python documentation, see
# https://www.python.org/dev/peps/pep-0440/#version-specifiers
```

With such proper comment on top of requirements.txt we will keep a reminder to ourself how to move forward with package versions.

vkuznet (approved):

amaltaro (approved):
These changes are looking good to me, thanks Erik!

I think there is one last bit to be discussed/decided though. In this setup, do we want to:
* provide a full environment for WMAgent + Central services
* and provide a full dev/test environment as well (like a few extra libraries that are only related to code quality like mock/pylint/etc)
?

If it's all in one, then I think it's still missing a library or two (e.g. mongomock).

vkuznet:
@amaltaro my view on this is the following:
- venv captures and provides everything we need from Python point of view to run things as production or not, therefore I suggest to make venv build/deploy script be flexible to support all environments. For instance, `deploy.sh production "wmagent reqmgr2"` command will install wmagent, reqmgr2 in production environment, while `deploy.sh development all` will install all packages as development environment. Here I only refer to `deploy.sh` as deployment script, but its actual name is irrelevant.
- we should separate venv from non-Python dependencies, i.e. if MariaDB or CouchDB is required it can and should be installed separately since they are independent from WMCore codebase. Such separation will allow to accomplish first item easily, and in fact it the "environment" may be boundaries can be vanishing. For example, I do all my development on my mac laptop. As such, in a past I installed independently from CMS the following services: MySQL, ORACLE, CouchDB, MongoDB, all monitoring tools. Since I can install them using external binaries, I don't need anything from WMCore to take care of them, and I only can clone WMCore and run any service as pure python application. Therefore, the only requirements in my view is presence of Python itself and Python libraries required by WMCore, e.g. CherryPy. Everything else can be run separately. Said that, what is required is set of scripts how to setup and inject schema files, and data which are required for WMCore services
- we also should separate authentication from venv, i.e. to develop things I don't need CMSWEB environment and I can access anything through `http://localhost:port`. As such, authentication layer is only required when you put things into production and it should be independent from WMCore deployment. Said that, the auth layer should also be easy to install in local env. therefore I push hard to switch from apache FEs to auth-proxy-server (which is a static executable which can work in VM, in your local laptop and across OSes). We should not be bound to requirement to have full featured CMSWEB VM with full stack in order to develop and test our services.

tools/build_pypi_packages.sh
amaltaro: @goughes Erik, I actually think `wmagent-devtools` will fail to build, since it's not defined in the setup_dependencies.py.
And I also have a question here, what is the difference between `wmagent-devtools` and `wmcore` packages?

requirements.txt
```diff
@@ -1,32 +1,43 @@
...
-SQLAlchemy==1.3.3             # wmagent
-stomp.py==4.1.15              # wmagent
-CMSCouchapp==1.2.10           # wmagent
+Cheetah3~=3.2.6.post2         # wmagent,reqmgr2,reqmon
+CherryPy~=17.4.0              # wmagent,wmagent-devtools,reqmgr2,reqmon,global-workqueue,reqmgr2ms
+CMSCouchapp~=1.3.4            # wmagent
+CMSMonitoring~=0.3.4          # wmagent,reqmgr2,reqmon,global-workqueue,reqmgr2ms
+coverage~=5.4                 # wmagent,wmagent-devtools
+cx-Oracle~=7.3.0              # wmagent
+dbs3-client~=4.0.5            # wmagent,reqmgr2,reqmon,global-workqueue
```
amaltaro, line 21: Erik, please update this version to `4.0.6` instead.

amaltaro (changes requested):
Valentin, I agree with almost everything that you mentioned above, if not all. However, we should keep things separated not to deviate too much from the expected development requested in the GH issue.

@goughes Erik, I have a couple of comments left along the code that needs attention, and it should be ready to go.

Next step then - which can be done either here or in a new GH ticket - is to flag python libraries that are not required for running production services (thus, only required for our dev setup). I think we could make something similar to what you have already implemented for wmagent | reqmgr2 | etc. What do you think?

tools/build_pypi_packages.sh
goughes: @amaltaro I added the `wmagent-devtools` package. The `wmcore` package was only focused on source code (python, css, js, etc.). I added some additional directories to the `wmagent-devtools` that I thought would be appropriate. Let me know if you think we should add/remove anything.

amaltaro: It looks good to me, Erik! Once we start diving into these PyPi packages, we might extend and/or polish anything that we find. Thanks

amaltaro (approved):

amaltaro:
@goughes I just noticed there is a conflict with ` requirements_py3.txt`. You probably need to rebase this PR. In case you need, here is a very good reference for that: https://steveklabnik.com/writing/how-to-squash-commits-in-a-github-pull-request

vkuznet (changes requested):

goughes:
Fixed the conflict and bumped dbs3-client to 4.0.7. I think this is gtg now @amaltaro.

amaltaro:
Thanks Erik! This was the only leftover that Valentin requested as well, so it's good to go now! Thanks!

PR #10987, opened 2022-02-10
Remove files from mongoProjection

### LINKED ISSUE #10981: MSUnmerged: Wrong input filters applied when reading from MongoDB
**Impact of the bug**
MSUnmeged

**Describe the bug**
Not exactly a bug since we are already running quite safe with the current code, which already has a workaround for this. 
When reading from MongoDB we cannot apply filters of the sort:
```
mongoProjection = {
    "_id": False,
    "name": True,
    ...
    "files": False
}
``` 

Because we are getting this result: 
```
In [3]: from WMCore.MicroService.MSUnmerged.MSUnmergedRSE import MSUnmergedRSE

In [4]: rse = MSUnmergedRSE('T2_US_Wisconsin')

In [5]: rse.readRSEFromMongoDB(msUnmergedColl, useProjection=True)
OperationFailure                          Traceback (most recent call last)
<ipython-input-5-52e7cdc3e352> in <module>
----> 1 rse.readRSEFromMongoDB(msUnmergedColl, useProjection=True)

~/WMCoreDev.d/WMCore/src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py in readRSEFromMongoDB(self, collection, useProjection)
     99         if useProjection:
    100             mongoProjection = self.buildMongoProjection()
--> 101             mongoRecord = collection.find_one(self.mongoFilter, projection=mongoProjection)
    102         else:
    103             mongoRecord = collection.find_one(self.mongoFilter)

...

~/WMCoreDev.d/WMCore.venv3/lib64/python3.6/site-packages/pymongo/helpers.py in _check_command_response(response, msg, allowable_errors, parse_write_concern_error)
    157 
    158             msg = msg or "%s"
--> 159             raise OperationFailure(msg % errmsg, code, response)
    160 
    161 

OperationFailure: Cannot do exclusion on field files in inclusion projection

```

And the reason is because we are hitting some MongoDB functionality constraints here. It simply does not allow to have `exclusion` fields inside `inclusion` filters. In those cases the only field we are allowed to exclude is `_id`, which is basically what we need to exclude for sure. While if we want to skip some of the rest, they simply must not be present into the filter. Here is some discussion on the topic from the MongoDb comunity with some more  details [1]

[1]
https://www.mongodb.com/community/forums/t/projection-does-not-allow-exclusion-inclusion-together/31756
 

**How to reproduce it**
Just execute the steps from the code snippet above, trying to read from MongoDB with a complex `inclusion` filter which is having more than a single `exclusion` at the `_id` field. 

**Expected behavior**
Create the mongoDB filter properly so it could be applied at read and write operations with equivalent result without failing any of them. 

None

todor-ivanov:
Fixes #10981

#### Status
Ready

#### Description
We are fixing the  `OperationFailure: Cannot do exclusion on field files in inclusion projection` by simply removing the `'files:' False`  field from the mongo projection. It will be added few lines bellow for the cases when it is needed. 

YES 

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py
@@ -82,7 +82,6 @@ class MSUnmergedRSE(dict):
             "isClean" : True,
             "timestamps": True,
             "counters": True,
-            "files": False,
             "dirs": True}
         if fullRSEToDB:
             mongoProjection.update({"files": True})


vkuznet (approved):

amaltaro (approved):
Thanks Todor. I understand we do not need to make a hot-fix with this for MSUnmerged, right?

PR #10989, opened 2022-02-10
Add parameters to rucio datasets (block-level rules)

### LINKED ISSUE #10558: Use config file to specify additional parameters for block-level rules
**Impact of the new feature**
WMAgent and T0 WMAgent

Tier0 Replay data is useful for a limited amount of time and can be usually deleted after a few weeks (depending on the kind of test). If the block-level rules are not deleted, junk data continues to accumulate. As disk storage is a premium resource for T0 operation, we sould like to ensure that replay data gets deleted after a safe amount of time. 

We would like te possibility to specify additional rule parameters for block rules, similarly of how it is already possible to do so for Container rules with the `config.RucioInjector.containerDiskRuleParams` attribute. With this, we could add a costumized lifetime value to our T0 replay rules without affecting production. Also, this would allow for the modification of other rule parameters in case the need arises.

A different solution would involve the use of acron to delete rules after an specific amount of time. While this option is viable, It would not allow us to modify the lifetime value for specific replays.

jhonatanamado:
Fixes #10558 

#### Status
Tested with a replay

#### Description
Add different parameters to rucio datasets via WMAgentConfig. This new parameter called `config.RucioInjector.datasetRuleParams` will allow T0 agent to modify as needed the attributes for block-level rules. Will be extremely useful for our account `tier0_replay` to add a lifetime parameter due that sometimes during our testing procedures with new configurations the block-level rules are not been deleted.
The T0 production agent will add the following entry to the config file `config.RucioInjector.datasetRuleParams = {}` while in our testing environment we will add `config.RucioInjector.datasetRuleParams = {'lifetime': 14 * 24 * 60 * 60}`. This will help the cleaning procedure in our operational site `T0_CH_CERN_Disk` from only replay data without affecting the logic of deleting block-level rules in production and its easy to overwrite this parameter using our usual deployment script.
As an example, the following two block-level rules. The first one as usual and the second one with the changes proposed in this PR
```
Id:                         8b7e5141e23c4388a7c97c30ecaea41a
Account:                    tier0_replay
Scope:                      cms
Name:                       /L1Accept/Tier0_REPLAY_2021-v1/RAW#965d5fe0-7117-47f0-a3df-b73cd439fe9e
RSE Expression:             T0_CH_CERN_Disk
Copies:                     1
State:                      OK
Locks OK/REPLICATING/STUCK: 10/0/0
Grouping:                   DATASET
Expires at:                 None
Locked:                     False
Weight:                     None
Created at:                 2022-02-10 15:11:51
Updated at:                 2022-02-10 15:17:00
Error:                      None
Subscription Id:            None
Source replica expression:  None
Activity:                   Production Output
Comment:                    WMAgent automatic container rule
Ignore Quota:               False
Ignore Availability:        True
Purge replicas:             False
Notification:               NO
End of life:                None
Child Rule Id:              None
```

```
Id:                         9a9d91b238de40f09fa794a3f23f0c60
Account:                    tier0_replay
Scope:                      cms
Name:                       /L1Accept/Tier0_REPLAY_2021-v1/RAW#e9aec287-c790-41bf-a6da-4d8fe5f991f7
RSE Expression:             T0_CH_CERN_Disk
Copies:                     1
State:                      OK
Locks OK/REPLICATING/STUCK: 10/0/0
Grouping:                   DATASET
Expires at:                 2022-02-17 05:27:48
Locked:                     False
Weight:                     None
Created at:                 2022-02-10 05:27:48
Updated at:                 2022-02-10 05:33:14
Error:                      None
Subscription Id:            None
Source replica expression:  None
Activity:                   Production Output
Comment:                    WMAgent automatic container rule
Ignore Quota:               False
Ignore Availability:        True
Purge replicas:             False
Notification:               NO
End of life:                None
Child Rule Id:              None
```

MAYBE

#### Related PRs
T0 PR https://github.com/dmwm/T0/pull/4643

None

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -361,6 +361,7 @@ config.RucioInjector.createBlockRules = True
 config.RucioInjector.RSEPostfix = False  # enable it to append _Test to the RSE names
 config.RucioInjector.metaDIDProject = "Production"
 config.RucioInjector.containerDiskRuleParams = {"weight": "ddm_quota", "copies": 2, "grouping": "DATASET"}
+config.RucioInjector.blockRuleParams = {}
 # this RSEExpr below might be updated by wmagent-mod-config script
 config.RucioInjector.containerDiskRuleRSEExpr = "(tier=2|tier=1)&cms_type=real&rse_type=DISK"
 config.RucioInjector.rucioAccount = "OVER_WRITE_BY_SECRETS"
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -62,6 +62,7 @@ class RucioInjectorPoller(BaseWorkerThread):
         self.lastRulesExecTime = 0
         self.createBlockRules = config.RucioInjector.createBlockRules
         self.containerDiskRuleParams = config.RucioInjector.containerDiskRuleParams
+        self.blockRuleParams = config.RucioInjector.blockRuleParams
         self.containerDiskRuleRSEExpr = config.RucioInjector.containerDiskRuleRSEExpr
         if config.RucioInjector.metaDIDProject not in RUCIO_VALID_PROJECT:
             msg = "Component configured with an invalid 'project' DID: %s"
@@ -239,6 +240,7 @@ class RucioInjectorPoller(BaseWorkerThread):
                           ignore_availability=True, meta=self.metaData)
             rseName = "%s_Test" % item['pnn'] if self.testRSEs else item['pnn']
             # DATASET = replicates all files in the same block to the same RSE
+            kwargs.update(self.blockRuleParams)
             resp = self.rucio.createReplicationRule(item['blockname'],
                                                     rseExpression=rseName, **kwargs)
             if resp:
--- a/test/python/WMComponent_t/RucioInjector_t/RucioInjectorPoller_t.py
+++ b/test/python/WMComponent_t/RucioInjector_t/RucioInjectorPoller_t.py
@@ -77,6 +77,7 @@ class RucioInjectorPollerTest(EmulatedUnitTestCase):
         config.RucioInjector.RSEPostfix = False  # enable it to append _Test to the RSE names
         config.RucioInjector.metaDIDProject = "Production"
         config.RucioInjector.containerDiskRuleParams = {"weight": "ddm_quota", "copies": 2, "grouping": "DATASET"}
+        config.RucioInjector.blockRuleParams = {}
         config.RucioInjector.containerDiskRuleRSEExpr = "(tier=2|tier=1)&cms_type=real&rse_type=DISK"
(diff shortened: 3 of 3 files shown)


etc/WMAgentConfig.py
```diff
@@ -361,6 +361,7 @@
 config.RucioInjector.RSEPostfix = False  # enable it to append _Test to the RSE names
 config.RucioInjector.metaDIDProject = "Production"
 config.RucioInjector.containerDiskRuleParams = {"weight": "ddm_quota", "copies": 2, "grouping": "DATASET"}
+config.RucioInjector.datasetRuleParams = {}
```
amaltaro, line 364: Can you please rename it to `blockRuleParams`? The dataset word is extremely confusing when it comes to Rucio vs CMS.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -239,7 +240,12 @@ def insertBlockRules(self):
                           ignore_availability=True, meta=self.metaData)
             rseName = "%s_Test" % item['pnn'] if self.testRSEs else item['pnn']
             # DATASET = replicates all files in the same block to the same RSE
-            resp = self.rucio.createReplicationRule(item['blockname'],
+            if self.isT0agent:
+                kwargs.update(self.datasetRuleParams)
```
amaltaro, line 244: I'd suggest to remove this if/else. By default it will be set to an empty dictionary, so it's harmless to any system unless the agent configuration is modified.

test/python/WMComponent_t/RucioInjector_t/RucioInjectorPoller_t.py
```diff
@@ -77,6 +77,7 @@ def createConfig(self):
         config.RucioInjector.RSEPostfix = False  # enable it to append _Test to the RSE names
         config.RucioInjector.metaDIDProject = "Production"
         config.RucioInjector.containerDiskRuleParams = {"weight": "ddm_quota", "copies": 2, "grouping": "DATASET"}
+        config.RucioInjector.datasetRuleParams = {}
```
amaltaro, line 80: same here.

amaltaro (changes requested):
@jhonatanamado this looks like a neat change! However, could you please consider those comments I made along the code? Thanks!

etc/WMAgentConfig.py
jhonatanamado: Sure, I tried to follow the same convention as the previous lines, but indeed, is extremely confusing when it comes to Rucio-CMS

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
jhonatanamado: Ok, I just wanted to be extra cautious not trying to break any other system.

amaltaro (approved):
Thank you for providing these changes, Jhonatan. It looks good to me!

PR #10990, opened 2022-02-11
Fix DBS3Reader crossCheck APIs; updated unit tests and mock APIs/data

### LINKED ISSUE #10988: DBS3Reader crossCheck/crossCheckMissing API incorrectly parsing a dictionary
**Impact of the bug**
WMCore in general (even though it seems not to be used by anyone)

**Describe the bug**
When executing those methods, the return data type for `listFileArray` is actually a dictionary, while the source code parses it as if it was a list data type, thus raising an exception.

**How to reproduce it**
None

**Expected behavior**
We need to properly extract a flat list of lfns out of the `listFileArray`, both in the `crossCheck` and `crossCheckMissing` module APIs.

None

amaltaro:
Fixes #10988 

#### Status
not-tested

#### Description
Summary of changes is:
* First commit: a few bug-fixes for DBS3Reader APIs. Mostly related to how data is parsed and correcting mistakes of lists/sets/dicts.
* Second commit: added a few extra DBS APIs to be mocked and updated our json mock data (only global affected)
* Third commit: a few updates to the unit tests, many new tests created.
* Fourth commit: ad-hoc script to compare prod against testbed data (TODO: to be removed before it gets merged)

For the record, after updating the additional APIs and args, this is how we update the json mock data:
```
[dmwm@211f83ccdead DBSClient]$ python3 MakeDBSMockFiles.py 
```

YES

#### Related PRs
None

Requires dbs3-client 4.0.7


src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -724,18 +728,20 @@ def listDatasetLocation(self, datasetName):
             raise DBSReaderError(msg)
 
         if not blocksInfo:  # no data location from dbs
-            return list()
+            return list(locations)
```
todor-ivanov, line 731: Isn't that a minor change of behavior? Before the list returned has been empty.

todor-ivanov (approved):
Thank you @amaltaro The changes look good to me. I left a single question inline, for which I am sure you already have thought of. It is only for clarification.

src/python/WMCore/Services/DBS/DBS3Reader.py
amaltaro: It will still be an empty list. `locations` is an empty set here, so we simply cast it to list. I made this change just to keep it more consistent with the same return at the very end of this method.

amaltaro:
From all the new unit tests created here, only:
```
[WMCore_t.Services_t.DBS_t.DBSReaderReal_t.DBSReaderTest:testGetParentDatasetTrio was added with status error. Must be fixed]()
```
fails! Which is expected, since we need to have the new dbs3-client version in our stack.

While this:
```
[WMCore_t.Services_t.DBS_t.DBSReader_t.DBSReaderTest:testGetParentDatasetTrio was added with status success]()
```
only succeeded because it's mocking the data service calls and data.

Now that we know testbed and production are behaving consistent with the remaining APIs, I can remove the 4th commit (DBSReaderReal_t.py). Will do so in the coming hours.

src/python/WMCore/Services/DBS/DBS3Reader.py
todor-ivanov: thanks @amaltaro, that makes sense.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -339,7 +339,9 @@ def crossCheck(self, datasetPath, *lfns):
         Return the list of lfns that are in the dataset
 
         """
-        allLfns = self.dbs.listFileArray(dataset=datasetPath, validFileOnly=1, detail=False)
+        allLfns = []
+        for fileDict in self.dbs.listFileArray(dataset=datasetPath, validFileOnly=1, detail=False):
```
vkuznet, line 343: The call to `self.dbs.listFileArray` may fail and throw exception, as such the for loop here will break on exception. I don't know if you want to do it this way or if you actually want to put this API call into `try/except` block. I will do the later and throw proper exception as you do in other parts of the code, see line 144.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -352,7 +354,9 @@ def crossCheckMissing(self, datasetPath, *lfns):
         are *not* known by DBS
 
         """
-        allLfns = self.dbs.listFileArray(dataset=datasetPath, validFileOnly=1, detail=False)
+        allLfns = []
+        for fileDict in self.dbs.listFileArray(dataset=datasetPath, validFileOnly=1, detail=False):
```
vkuznet, line 358: same comment here, it will be better to put `self.dbs.listFileArray` into `try/except` block

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -724,18 +728,20 @@ def listDatasetLocation(self, datasetName):
             raise DBSReaderError(msg)
 
         if not blocksInfo:  # no data location from dbs
-            return list()
+            return list(locations)
 
         for blockInfo in blocksInfo:
-            locations.update(blockInfo['origin_site_name'])
+            locations.add(blockInfo['origin_site_name'])
```
vkuznet, line 734: here code is not protected against missing key in a dictionary. I think it will be better to do the following:
```
site = blockInfo.get('origin_site_name', None)
if site:
   locations.add(site)
```

src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py
```diff
@@ -88,6 +88,8 @@ def listFileLumiArray(self, **kwargs):
         if 'logical_file_name' in kwargs and len(kwargs['logical_file_name']) > 1:
             origArgs = copy.deepcopy(kwargs)
             returnDicts = []
+            if isinstance(kwargs['logical_file_name'], str):
```
vkuznet, line 91: what if `logical_file_name` will not be in kwargs? I think it is better to do like `kwargs.get('logical_file_name', None)`.

src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py
```diff
@@ -88,6 +88,8 @@ def listFileLumiArray(self, **kwargs):
...
+                kwargs['logical_file_name'] = [kwargs['logical_file_name']]
```
vkuznet, line 92: an additional comment would be welcome here. Why do you convert it to the list? And, again code is not protected if kwargs will not have such key in it.

test/python/WMCore_t/Services_t/DBS_t/DBSReaderReal_t.py
```diff
@@ -0,0 +1,485 @@
...
+#class DBSReaderTest(EmulatedUnitTestCase):
+class DBSReaderTest(unittest.TestCase):
+
+    def setUp(self):
+        """
+        _setUp_
+
+        Initialize the PhEDEx API to point at the test server.
+        """
+        self.endpoint = 'https://cmsweb-prod.cern.ch/dbs/prod/global/DBSReader'
```
vkuznet, line 42: Please, try to avoid hard-coded values. Per long discussion in a past we may convert back to `cmsweb.cern.ch`. I suggest to do the following:
```
self.endpoint = os.environ.get('DBS_PROD', 'https://cmsweb-prod.cern.ch/dbs/prod/global/DBSReader')
self.testbed = os.environ.get('DBS_TESTBED', 'https://cmsweb-testbed.cern.ch/dbs/int/global/DBSReader')
```
Doing this way you allow to setup env variables for your tests and use it with ANY DBS deployment instances, and yet use the one you put in a code as defaults.

test/python/WMCore_t/Services_t/DBS_t/DBSReaderReal_t.py
```diff
@@ -0,0 +1,485 @@
...
+
+    def testListDatatiers(self):
+        """
+        listDatatiers returns all datatiers available
+        """
+        self.dbs = DBSReader(self.endpoint)
+        prod = self.dbs.listDatatiers()
+        self.dbsTest = DBSReader(self.testbed)
+        testbed = self.dbsTest.listDatatiers()
+        print(prod)
```
vkuznet, line 67: It will be much more clear if you'll add to print statement which instance results come from, e.g.
```
print("DBS {}\nresults {}".format(self.endpoint, prod))
```

test/python/WMCore_t/Services_t/DBS_t/DBSReaderReal_t.py
```diff
@@ -0,0 +1,485 @@
...
+        print(testbed)
```
vkuznet, line 68: same here, please put
```
print("DBS {}\nresults {}".format(self.testbed, testbed))
```

test/python/WMCore_t/Services_t/DBS_t/DBSReaderReal_t.py
```diff
@@ -0,0 +1,485 @@
...
+    def testListPrimaryDatasets(self):
+        """
+        listPrimaryDatasets returns known primary datasets
+        """
+        self.dbs = DBSReader(self.endpoint)
+        prod = self.dbs.listPrimaryDatasets('JetMETTau*')
+        self.dbsTest = DBSReader(self.testbed)
+        testbed = self.dbsTest.listPrimaryDatasets('JetMETTau*')
+        print(prod)
+        print(testbed)
```
vkuznet, line 95: replace it with
```
print("DBS {}\nresults {}".format(self.endpoint, prod))
print("DBS {}\nresults {}".format(self.testbed, testbed))
```

test/python/WMCore_t/Services_t/DBS_t/DBSReaderReal_t.py
```diff
@@ -0,0 +1,485 @@
...
+    def testMatchProcessedDatasets(self):
+        """
+        matchProcessedDatasets returns known processed datasets
+        """
+        self.dbs = DBSReader(self.endpoint)
+        prod = self.dbs.matchProcessedDatasets('Jet', 'RAW', 'Run2011A-v1')
+        self.dbsTest = DBSReader(self.testbed)
+        testbed = self.dbsTest.matchProcessedDatasets('Jet', 'RAW', 'Run2011A-v1')
+        print(prod)
+        print(testbed)
```
vkuznet, line 108: replace it with
```
print("DBS {}\nresults {}".format(self.endpoint, prod))
print("DBS {}\nresults {}".format(self.testbed, testbed))
```

test/python/WMCore_t/Services_t/DBS_t/DBSReaderReal_t.py
```diff
@@ -0,0 +1,485 @@
...
+
+        prod = prod[0]
+        testbed = testbed[0]
+        # remove keys that will be difference no matter what
+        for keyName in ("last_modified_by", "last_modification_date", "LumiList"):
+            prod.pop(keyName)
+            testbed.pop(keyName)
+
+        print(prod)
+        print(testbed)
```
vkuznet, line 277: replace it with
```
print("DBS {}\nresults {}".format(self.endpoint, prod))
print("DBS {}\nresults {}".format(self.testbed, testbed))
```

vkuznet (changes requested):

src/python/WMCore/Services/DBS/DBS3Reader.py
amaltaro: I did not want to make too many changes in here, outside of the scope of the original issue, but I think it's a fair request. I will get this code updated soon.

src/python/WMCore/Services/DBS/DBS3Reader.py
amaltaro: I was going to say the line above `if not blocksInfo` should give enough protection, but I think there is a minor improvement that can be done here.

src/python/WMQuality/Emulators/DBSClient/MockDbsApi.py
amaltaro: the key presence is already verified a few lines above.

amaltaro:
@vkuznet @todor-ivanov thanks for your review. I have addressed some of those comments in my 5th and 6th commits, could you please have another look at them?

Regarding `DBSReaderReal_t.py`, I am not even bothering to go through the comments because I only wanted to test DBS testbed x production. Now that we know everything is working properly, I will actually remove it from this PR (remove the 4th commit). For future reference, here it is: https://github.com/dmwm/WMCore/pull/10990/commits/63eb6af72b636707e9151cb94bdc94aaf8e7dce4

vkuznet (approved):

amaltaro:
Thanks for the reviews! Merging

PR #10993, opened 2022-02-14
Add rucioConMonErrors counter.

### LINKED ISSUE #10992: MSUnmerged: Missing records in MongoDB for RSEs sitting in non-final state at Rucio Consistency Monitor
**Impact of the bug**
MSUnmergred

**Describe the bug**
Again not exactly a bug but following the findings here:  https://github.com/dmwm/WMCore/issues/10982#issuecomment-1036385522
we figured out that missing any record in MongoDB for all those RSEs in non-final or error state at Rucio Consistency Monitor is quite uncomfortable and time consuming when it comes to debugging, while keeping a single document for each one of them at MongoDB is not a big deal. So we decided to have them recorded and properly tracked in our database too.

**How to reproduce it**
Currently: Try to count the records in MongoDB would give a different number than the number of RSEs returned by queering Rucio with the general filter we are using for MSUnmerged. 

```
In [9]: mongoClt.client.msUnmergedDBProd.msUnmergedColl.count({})
Out[9]: 46

rucio list-rses --rses 'rse_type=DISK&cms_type=real' |wc -l
88
```

The missing records are due to early exit from the pipeline, before we  make any record in the database - see more debug information in the related issue cited above

**Expected behavior**
Preserve records for all the RSEs, regardless of the type of error.

### LINKED ISSUE #10994: MSUnmerged: Simplify service counters logic
**Impact of the new feature**
MSUnmerged

As already mentioned in several places we need to simplify the service counters logic. We are  still accumulating stuff in the service memory, which is a legacy from the time when we were not having the background database. It is not only uncomfortable to juggle with all those places to accumulate stuff it is also extremely error prone and is making any additional change related to service accounting logic extremely difficult in time.  A typical example are the latest two issues where we had to add new places  for accounting external services Errors : [1] && [2]

[1]
https://github.com/dmwm/WMCore/issues/10686

[2]
https://github.com/dmwm/WMCore/issues/10992
  

Get rid of the service's in memory counters per RSE and read them only from the background database. Keep only the pipeline counters, since they are  used for reporting the current run statistics in the REST interface. 

Keep using the old (error prone) logic.

**Additional context**
None

todor-ivanov:
Fixes #10992 
Fixes #10994 

#### Status
Ready

#### Description
With the current PR we are going to persist at MongoDB any RSE state as we find it in Rucio Consistency Monitor + The skip reason for the RSEs with old consistency record. 

YES

#### Related PRs
None

None


todor-ivanov:
During the process of testing those changes I've found that having all those places to accumulate numbers for accounting service work and errors, makes our life extremely difficult and dangerous. So I have created the following issue [1], which I intend to solve with this same PR in one of my following commits. 


[1]
https://github.com/dmwm/WMCore/issues/10994

amaltaro (approved):
Thanks Todor, it looks good to me!
Do we need to dump the database once again before deploying it in CMSWEB (for the new `rucioConMonStatus` attribute)?

todor-ivanov:
Hi @amaltaro, thanks for merging it. I just saw you had an additional question here:

> Do we need to dump the database once again before deploying it in CMSWEB 

No in this case we do not need to wipe out the whole database. The new field is not tried to be accessed anywhere,  before the document is updated at the database. This was what was driving the need for complete database deletion last time.

PR #10996, opened 2022-02-14
Remove verbose=True from getMSLogger calls

### LINKED ISSUE #10317: Set microservices log level to INFO instead of DEBUG; adapt source code accordingly
**Impact of the new feature**
All microservices

When we set the microservice loglevel to DEBUG level, this setting propagates throughout the whole rucio-clients dependency chain, generating lots of annoying and useless records in the logs.
 
A few actions/solutions are expected from this GH issue, such as:
* update the microservices `deployment` configuration with the INFO log level
* update the microservices `config_services` configuration with the INFO log level
* review ALL microservices source code and move useful logging records from DEBUG to INFO level, leaving only very deep (and with large content) logging records in DEBUG mode.

Do nothing and live with a very polluted log.

**Additional context**
Example of annoying logs come from `lock` and `connectionpool` modules, and probably others.

vkuznet:
Fixes #10317

#### Status
need confirmation from @amaltaro 

#### Description
remove `verbose=True` in getMSLogger, the level should be applied by passing proper logger instance.

MAYBE

#### Related PRs
https://github.com/dmwm/deployment/pull/1124

### DIFF
--- a/src/python/WMCore/MicroService/TaskManager.py
+++ b/src/python/WMCore/MicroService/TaskManager.py
@@ -110,7 +110,7 @@ class UidSet(object):
 class Worker(threading.Thread):
     """Thread executing worker from a given tasks queue"""
     def __init__(self, name, taskq, pidq, uidq, logger=None):
-        self.logger = getMSLogger(verbose=True, logger=logger)
+        self.logger = getMSLogger(logger=logger)
         threading.Thread.__init__(self, name=name)
         self.exit = 0
         self.tasks = taskq
@@ -160,7 +160,7 @@ class TaskManager(object):
 
     """
     def __init__(self, nworkers=10, name='TaskManager', logger=None):
-        self.logger = getMSLogger(verbose=True, logger=logger)
+        self.logger = getMSLogger(logger=logger)
         self.name = name
         self.pids = set()
         self.uids = UidSet()
--- a/src/python/WMCore/MicroService/Tools/Common.py
+++ b/src/python/WMCore/MicroService/Tools/Common.py
@@ -400,7 +400,7 @@ def getEventsLumis(dataset, dbsUrl, blocks=None, eventsLumis=None):
 
 def getComputingTime(workflow, eventsLumis=None, unit='h', dbsUrl=None, logger=None):
     "Return computing time per give workflow"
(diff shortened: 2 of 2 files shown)


vkuznet:
The changes to gitlab `services_configs` repo can be applied once https://github.com/dmwm/deployment/pull/1124 is approved.

amaltaro:
@vkuznet thanks for providing these options, but the last bullet in the expected outcome - from the GH ticket description - states:
```
review ALL microservices source code and move useful logging records from DEBUG to INFO level, leaving only very deep (and with large content) logging records in DEBUG mode.
```

and this is the hard part. We need to review every single log record from all the microservices, and decide which ones should be logged all the time (INFO) and which ones are really meant to be debug mode only. I am afraid to that that the best way to deal with those would be to have Todor and myself looking into those microservices and swapping logs, e.g.::
```
self.logger.debug()
```
to
```
self.logger.info()
```
when relevant.

vkuznet:
Alan, I understand, and I only wanted to help with what I can. In fact, I want to suggest a better approach to logging rather then simply swap INFO/DEBUG. In DBS GO server I use verbosity level, e.g. 0,1,2,3 such that I can change it in configuration and at run time and get more verbose information. Here it is a similar use-case. You actually need flexibility, i.e. increase the logging level during debugging phase and reduce it in normal operation. As such for MS services (since they are web services) you may add an extra end-point where you can send verbosity level (via POST request) and it will dynamically change web server behavior. Think about it.

Meanwhile, I provided necessary PRs and relevant info which you can use.

amaltaro (changes requested):
Just to "silence" this review request.

As already discussed in this PR, we need to review all the log records throughout the microservices and decide which ones should be defined as `logger.debug()` and `logger.info()`. Such that when we move log level to INFO, we do not miss important information currently set as logger.debug.

I suggest to either Todor or myself to look into it, while looking into the production logs in parallel and assess what is the minimum log that we need to have in the service logs (INFO level). We might want to come up with an extra PR to complement this one.

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

amaltaro:
test this please

Issue #8478, opened 2018-02-21
CleanCouchPoller is very slow


amaltaro:
And so it triggers the heartbeat alarm. It has been executing the same cycle for 680min now (>10h!!!).
Maybe we can improve or slice the amount of work done in every cycle such that it doesn't take so long in busy cycles.

amaltaro:
We need to break the amount of work this thread can handle into smaller pieces such that we avoid the false heartbeat alarms. Hitting us right now on vocms0250.

amaltaro:
Just an update on this issue, since I see TaskArchiver heartbeat timeouts in WMStats (1.3.3.patch1).

Checking the component log (seems to be stuck creating a bunch of workload summary):
```
2020-06-10 22:26:07,766:140398655698688:INFO:CleanCouchPoller:Total 478 workload summary will be created
2020-06-10 22:27:17,491:140398733711104:INFO:BaseWorkerThread:TaskArchiverPoller took 97.275 secs to execute
2020-06-11 01:08:06,500:140398733711104:ERROR:LogDB:LogDBBackend delete API failed, error=(28, 'Operation timed out after 857179 milliseconds with 0 bytes received')
```

the ps command gives me this
```
[cmsdataops@cmsgwms-submit6 current]$ ps -p 700668 -o pid,ppid,rss,vsize,pcpu,pmem,cmd -ww | grep 700668
 700668       1 19107904 19519780 5.4 14.5 python /data/srv/wmagent/v1.3.3.patch1/sw/slc7_amd64_gcc630/cms/wmagent/1.3.3.patch1/bin/wmcoreD --restart --components=PhEDExInjector,DBS3Upload,TaskArchiver
[cmsdataops@cmsgwms-submit6 current]$
```

which is a memory footprint of ~18GB, also confirmed from the top command:
```
[cmsdataops@cmsgwms-submit6 current]$ top -p 700668

top - 01:15:43 up 72 days, 15:36,  1 user,  load average: 9.40, 8.63, 8.52
Tasks:   1 total,   0 running,   1 sleeping,   0 stopped,   0 zombie
%Cpu(s): 22.6 us,  5.8 sy,  0.0 ni, 70.8 id,  0.1 wa,  0.0 hi,  0.7 si,  0.0 st
KiB Mem : 13158487+total, 14494432 free, 72428624 used, 44661812 buff/cache
KiB Swap: 33553404 total, 31308400 free,  2245004 used. 58529396 avail Mem 

[...]

Issue #4656, opened 2013-05-31
JobAccountant is slow


dballesteros7:
Known by those who came before, and to be remembered by those who will follow. 

The JobAccountant is the most hard working component in the system but it needs a tune-up. In a high load MC agent I discovered backlogs as big as 2M jobs.

There must be some improvements we can do to this guy.

amaltaro:
Getting back to this a few years later :) Indeed it does some heavy work and apparently it's performance is in the order of 10k jobs per hour (really?!? what a let down)

vocms0126 accumulated a backlog of 170k jobs, and we can see it took around 18h to go through them.
```
2017-09-07 11:11:44,292:140109560870656:INFO:JobAccountantPoller:Found 170089 completed jobs
2017-09-08 05:31:02,054:140109560870656:INFO:JobAccountantPoller:Found 82627 completed jobs
```

amaltaro:
@ticoann since Erik G. is waiting for the new draining status update. Maybe this one would be a good candidate to get him involved?

amaltaro:
Reopening because this has to be fixed, eventually.

amaltaro:
The impact of very short jobs (<=1h), something like 15k jobs completed within an hour:
```
cmst1@vocms0253:/data/srv/wmagent/current $ grep 'completed jobs' install/wmagent/JobAccountant/ComponentLog | tail -n20
2018-08-06 12:14:14,414:139642164897536:INFO:JobAccountantPoller:Found 357543 completed jobs
2018-08-08 01:59:39,016:139642164897536:INFO:JobAccountantPoller:Found 569940 completed jobs
2018-08-10 09:53:06,364:139642164897536:INFO:JobAccountantPoller:Found 645218 completed jobs
```

klannon:
Despite the vintage of this issue, @amaltaro confirms this is an issue that needs to be tackled before HL-LHC.

PR #10998, opened 2022-02-15
Remove PFNs from a final document we send to WMArchive

### LINKED ISSUE #10879: Big jsons uploaded to WMArchive
**Impact of the bug**
WMAgent, WMArchive

**Describe the bug**
While debugging missing monitoring information from ES the Monit team has discovered some quite big json files uploaded to WMarchive and ES [1]. The document's size has been of the order of  2MBs. The documents themselves seem to be sane. Their origin are few workflows which tend to create huge LFNArrays for the input files for the job itself. This  is a behavior on which we need more feedback from other groups like P&R and PDMv in order to estimate the expected future frequency of those. 

In the meantime there was a temporary solution applied by Monit side  so that  we  may continue to use the monitoring properly (all is well explained in the relevant ticket [1]), but we cannot rely on that in the long term. So this ticket is expected to be a  follow up on the decisions taken on our side to revisit the information which we upload to WMArchive and ES.  

[1]
https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC2961279

**How to reproduce it**
N/A 
In order to reproduce  those huge documents we need some feedback so we can relate to a speciffic type of workflows/campaigns.

**Expected behavior**
Documents uploaded to WMArchive and ES to be in the order of kilobytes rather than megabytes.

N/A

vkuznet:
Fixes #10879 

#### Status
ready

#### Description
Remove PFNArray from final WMArchive document

YES, since we do not fill out PFNArray part

#### Related PRs

### DIFF
--- a/src/python/WMCore/Services/WMArchive/DataMap.py
+++ b/src/python/WMCore/Services/WMArchive/DataMap.py
@@ -1,6 +1,5 @@
 from __future__ import (division, print_function)
 from builtins import str
-from past.builtins import basestring
 from future.utils import viewitems, listvalues, listitems
 
 import copy
@@ -13,18 +12,17 @@ WMARCHIVE_TOP_LEVEL_REMOVE_OUTSIDE_LAYER = ["steps"]
 # convert data format under stpes["cmsRun1"/"logArch1"/"stageOut1"]["output"]
 WMARCHIVE_REMOVE_OUTSIDE_LAYER = ["checksum", "dataset"]
 # convert to list from str
-WMARCHIVE_CONVERT_TO_LIST = ["OutputPFN", "lfn"]
+WMARCHIVE_CONVERT_TO_LIST = ["lfn"]
 
-WMARCHIVE_DATA_MAP = {"OutputPFN": "outputPFNs", "inputPath": "inputDataset",
-                      "lfn": "outputLFNs", "input": "inputLFNs", "inputpfns": "inputPFNs"}
+WMARCHIVE_DATA_MAP = {"inputPath": "inputDataset",
+                      "lfn": "outputLFNs", "input": "inputLFNs"}
 
-WMARCHIVE_REMOVE_FIELD = ["InputPFN", "pfn", "user_dn", "user_vogroup", "user_vorole"]
+WMARCHIVE_REMOVE_FIELD = ["user_dn", "user_vogroup", "user_vorole"]
 WMARCHIVE_COMBINE_FIELD = {"outputDataset": ["primaryDataset", "processedDataset", "dataTier"]}
 
 WMARCHIVE_LFN_REF_KEY = ["lfn", "files"]
-WMARCHIVE_PFN_REF_KEY = ["pfn"]
-WMARCHIVE_FILE_REF_KEY = {"LFN": WMARCHIVE_LFN_REF_KEY,
-                          "PFN": WMARCHIVE_PFN_REF_KEY}
+WMARCHIVE_PFN_REF_KEY = []
+WMARCHIVE_FILE_REF_KEY = {"LFN": WMARCHIVE_LFN_REF_KEY}
 
 ERROR_TYPE = {'exitCode': int}
 
@@ -94,7 +92,6 @@ STEP_DEFAULT = {  # 'name': '',
         # 'guid': '',
         # 'inputDataset': '',
         'inputLFNs': [],
-        'inputPFNs': [],
         # TODO change to empty string from None
         # 'location': '',
         # 'merged': False,
@@ -102,7 +99,6 @@ STEP_DEFAULT = {  # 'name': '',
         # 'output_module_class': '',
         # 'outputDataset': '',
         'outputLFNs': [],
-        'outputPFNs': [],
         # 'prep_id': '',
         # 'processingStr': '',
         # 'processingVer': -1,
@@ -124,6 +120,23 @@ STEP_DEFAULT = {  # 'name': '',
     # 'stop': 1454569736
 }
 
+
+def cleanStep(idict):
+    """
+    perform clean-up of PFNs attributes in place for given dictionary
+
+    :param idict: a FWJR report dictionary
+    :return: a dictionary without PFNs
+    """
+    for step in ['input', 'output']:
+        data = idict.get(step, {})
+        for key, values in data.items():
+            for elem in values:
+                for skip in ['pfn', 'InputPFN', 'OutputPFN', 'inputpfns']:
+                    if skip in elem:
+                        del elem[skip]
+            data[key] = values
+    return idict
 
 def combineDataset(dataset):
     dataset["outputDataset"] = "/%s/%s/%s" % (dataset.pop("primaryDataset"),
@@ -300,12 +313,12 @@ def convertStepValue(stepValue):
 
     return stepValue
 
-
 def convertSteps(steps):
     stepList = []
     for key, value in viewitems(steps):
         stepItem = {}
         stepItem['name'] = key
+        value = cleanStep(value)
         stepItem.update(convertStepValue(value))
         _validateTypeAndSetDefault(stepItem, TOP_LEVEL_STEP_DEFAULT)
         stepList.append(stepItem)


vkuznet:
@amaltaro based on my understanding of the codebase we only need trivial change. So far I only removed PFN part of selected keys which in fact remove PFNArray and PFNArrayRef attributes in a final document. I don't know if you want to remove LFNArray too, the change will be trivial as well.

amaltaro (commented):
@vkuznet Valentin, can you please provide a diff of the final document with and without this patch? There might be unit tests to help you with that.

When I looked at this code, I found it very confusing and sometimes the same key is present in multiple level of the final dict/json.

vkuznet:
The diff is tricky since the data is randomly generated, but here are the examples of docs with and without PFNs:
- [fwjr-with-pfns.json](https://gist.github.com/vkuznet/a46c3b11a92c42ce4cd24a7522ab51b2)
- [fwjr-no-pfns.json](https://gist.github.com/vkuznet/a01202e432f07bfc252c0cc249c3d7bf)

amaltaro:
This data format conversion looks blurred to me. Reason why I started looking at the unit tests to make more sense of the outcome document.

For instance, the LFNArrayRef/PFNArrayRef seem to be like an index of attributes from the LFNArray/PFNArray lists. But then, when the later is shorter than the former how do you know which attribute type is missing in the output.

Also looking into the two json files that you shared, I see some strange output for the nested "pfn" key. Now it lists the full PFN instead of its index. So the current fix doesn't seem to make it too much better.

amaltaro:
@vkuznet I just wanted to check whether there is any information that you need from me here? Looking at my previous reply, I think there are still some inconsistencies in the final document that it would be better to understand and get it fixed.

vkuznet:
@amaltaro , I added new function to remove all PFNs, please see final doc [here](https://gist.github.com/vkuznet/a46c3b11a92c42ce4cd24a7522ab51b2). Now the code is ready for review again.

vkuznet:
@amaltaro could you please review it again, as I wrote I adjusted code to completely remove PFNs, you can see it in new [fwjr-no-pfns.json](https://gist.github.com/vkuznet/a01202e432f07bfc252c0cc249c3d7bf) JSON file I produced with new code.

amaltaro (changes requested):
Valentin, thanks for providing this new approach. 

However, even though it does the job, it doesn't look an efficient way to deal with this. Reason is, we actually run the whole logic mapping FJR information into a WMArchive document - expanding everything in memory - to only then go back to the WMArchive document and remove key/value pairs that we do not need.

The proper way to fix it IMO would be, to convert only the necessary information into a WMArchive document. This will save a lot of memory and CPU cycles (and to be frank, it will likely become a less complex logic).

vkuznet:
@amaltaro , do you have quantitative confirmation about excessive memory and CPU footprint? We run with PFNs for years and I never heard anything bother you or other in terms of submission. My point is that I tried a simple approach without spending too much time on optimization. If you have specific numbers to show that it takes that much CPU and that much memory I think it is not worth the effort. If you still insist to make it in a proper way, I'll do it, but it will take time and change of algorithm/code logic.

amaltaro:
Unfortunately I do not! I also believe that collecting such metrics could actually take more time than the actual implementation :)

Please don't take me as a picky person, I am just trying to avoid to end up with issues like these:
https://github.com/dmwm/WMCore/issues/8478
https://github.com/dmwm/WMCore/issues/4656

where things are functional, but fail to scale in specific situations.

A few weeks ago I looked into this code myself, and indeed it's quite complex and hard to follow. Which is another reason I would be in favor of actually not doing unnecessary things and make it cleaner and more readable.

vkuznet:
Alan, ok, I added new function `cleanStep` which removes provided by FWJR pfn related fields. The produced document no longer have PFNs, but I'm not sure if it is sufficient to address problems you're referring since FWJR which is passed to WMCore/Services/WMArchive/DataMap.py still may contain PFNs attribute in FWJR document. I have no idea where exactly FWJR is created and who is passing it to this module. Please review and let me know if it is sufficient. Otherwise, please point me to the place where in WMCore FWJR is constructed such that we can remove PFNs (if necessary) from there.

amaltaro (commented):
>Otherwise, please point me to the place where in WMCore FWJR is constructed such that we can remove PFNs (if necessary) from there.

There is nothing that needs to be done with the FJR. It's only the source information and the data structure that is parsed in order to produce a WMArchive job document.

Having said that, the optimal change here is to change the "converter" code, such that it disregard input LFNs and PFNs. In other words, instead of cleaning these attributes from the WMArchive job document, we should actually make sure that they don't even get (temporary) created in the document.

vkuznet:
Alan, I hope you read my changes, they do exactly what you said. I added new function `convertStep` which is called from `convertSteps` which by itself is called from `convertToArchiverFormat` (see https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WMArchive/DataMap.py#L320) which by itself takes given FWJR. Unless you clarify your statement further I see no further action I need to do with this PR. Please spend time to read changes and code itself. I only left `cleanup` function which I introduced earlier in case if you want to do cleanup step at the end. This function can be removed once you review the proposed changes.

vkuznet:
Alan, in addition what I wrote yesterday, I need to say that you're mixing two different issues here. One, which is issue of not sending large docs to MONIT is addressed in this PR, the second issue is (possible) large memory footprint you pointed out. Because current code does `deepcopy` at line https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WMArchive/DataMap.py#L318 from a given FWJR the memory issue is still present. As I pointed out already several times if you want to address the memory issue we either need to:
- remove `copy.deepcopy` call, but in this case I don't know what will happen with provided FWJR after data send to MONIT, i.e. if we remove certain things from provided dict and later they are used by something else in WMCore this will be wrong, or
- we should modify FWJR creation process to avoid putting into its doc PFNs

Therefore, we need to separate these two issues, and if memory footprint is bothering you I suggest to open different ticket for it and it should be handled from the view of handling FWJR docs.

src/python/WMCore/Services/WMArchive/DataMap.py
```diff
@@ -13,18 +13,17 @@
 # convert data format under stpes["cmsRun1"/"logArch1"/"stageOut1"]["output"]
 WMARCHIVE_REMOVE_OUTSIDE_LAYER = ["checksum", "dataset"]
 # convert to list from str
-WMARCHIVE_CONVERT_TO_LIST = ["OutputPFN", "lfn"]
```
amaltaro, line 16: I don't know if this information could be consumed by anyone, but I'd say `OutputPFN` should remain in the WMA doc, no?

src/python/WMCore/Services/WMArchive/DataMap.py
```diff
@@ -300,12 +297,26 @@ def convertStepValue(stepValue):
...
 
+def cleanStep(idict):
+    """
+    perform clean-up of PFNs attributes in place for given dictionary
+    """
+    for step in ['input', 'output']:
+        data = idict.get(step, {})
+        for key, values in data.items():
+            for elem in values:
+                for skip in ['pfn', 'InputPFN', 'OutputPFN', 'inputpfns']:
```
amaltaro, line 308: Shouldn't this be defined in the WMARCHIVE_REMOVE_FIELD variable? Maybe not, because it's actually removing keys from the FJR, right? Still, this would be better to be defined at the top.

amaltaro (changes requested):
Yes, if lfn information is no longer in the WMArchive document, then it does solve the problem reported in the GH issue. However, to me it looks more like a workaround than a real fix. 

What I would expect from this code (not necessarily your changes) is:
* load FJR -> perform fields selection and convert to WMArchive schema -> create WMArchive doc

instead, we have either (your 1st and 2nd attempt):
* load FJR -> perform fields selection and convert to WMArchive schema -> create WMArchive doc -> delete unnecessary fields -> create final WMArchive doc; OR
* load FJR -> delete specific fields -> perform fields selection and convert to WMArchive schema -> create WMArchive doc

Don't you think the very top/former model is THE correct way to fix it? 

This whole code is hard to read and lack of unit tests make these changes even more challenging (well, there is 1 "unit" test that checks the whole module in shot). That's why I was advocating for an approach that would go straight to the final result - likely making it less complex.

I will try to make sense of this by Monday morning...

vkuznet:
Alan, I think you contradict yourself. Originally, you asked to avoid unnecessary memory consumption by which can only be done before the load FWJR step. Once you load it, the memory is already allocated and consumer. Therefore, your desired approach:
```
load FJR -> perform fields selection and convert to WMArchive schema -> create WMArchive doc
```
is identical to
```
load FJR -> delete specific fields -> perform fields selection and convert to WMArchive schema -> create WMArchive doc
```
because the PFNs already exists in FWJR as far as I can tell. Don't you think so? Both these above approaches will be identical in terms of memory consumption since `deepcopy` will take your RAM by the time it will load FJR.

I don't think that full re-write is answer here. In fact, the second commit I made does exactly what you ask for, it does not delete specific fields explicitly, it skips them, i.e. it is making perform fields selection and convert to WMArchive schema. Once again, please read what `def cleanStep` is doing and if you still believe that it deletes specific fields point me how.

Anyway, I think this issue require a chat since it seems to me stuck with understanding how to address it.

src/python/WMCore/Services/WMArchive/DataMap.py
amaltaro: Maybe move it up in the module under FWJR_REMOVE_STEP_FIELD?

amaltaro:
> because the PFNs already exists in FWJR as far as I can tell. Don't you think so? Both these above approaches will be identical in terms of memory consumption since deepcopy will take your RAM by the time it will load FJR.

Now that I spent 30min looking into this code, my conclusion is that it's not worth it to go deeper. If we do so, then this should be refactored.

Yes, I agree that everything is already loaded in memory and duplicated with the deepcopy, whatever we do now will remain in memory until things go out of scope (next cycle).

Valentin, can you please look into the comments made along the code, update it if required, and provide me with a new dump of the outcome (as you've done initially in this PR)?

vkuznet:
Alan, I clean-up code, i.e. removed `cleanup` function I committed originally, moved (per your request) `cleanStep` function to top of the file (after module definitions), fixed docstring. The new doc can be seen [here](https://gist.github.com/vkuznet/a01202e432f07bfc252c0cc249c3d7bf) where I posted first dict output from the following command:
```
python test/python/WMCore_t/Services_t/WMArchive_t/DataMap_t.py
```
Please note, this command generates several docs which are dumped to the stdout. I checked them all for pfn appearance which is no longer there and put to the gist the first dict.

Please review the code again.

vkuznet:
Alan, do you need anything else on this issue?

amaltaro (commented):
Thanks for providing these further details, Valentin.

Looking into your gist/dict, this is the reason why I really think we should rework this document conversion:
```
  "LFNArray": [
    "/store/unmerged/logs/prod/2016/2/4/sryu_TaskChain_Data_wq_testt_160204_061048_5587/RECOCOSD/0000/0/7d7d41dc-cb02-11e5-833c-02163e00efd5-88-0-logArchive.tar.gz",
    "/store/unmerged/CMSSW_7_0_0_pre11/Cosmics/ALCARECO/DtCalib-RECOCOSD_TaskChain_Data_pile_up_test-v1/00000/ECCFE421-08CB-E511-9F4C-02163E017804.root",
    "/store/data/Run2011A/Cosmics/RAW/v1/000/160/960/E8099605-8853-E011-A848-0030487A18F2.root",
    "/store/unmerged/CMSSW_7_0_0_pre11/Cosmics/ALCARECO/MuAlCalIsolatedMu-RECOCOSD_TaskChain_Data_pile_up_test-v1/00000/9665EB21-08CB-E511-9F4C-02163E017804.root"
  ],
  "LFNArrayRef": [
    "outputLFNs",
    "lfn",
    "skippedFiles",
    "inputLFNs",
    "fallbackFiles"
  ],
```

even if that means a refactoring of the code.

If I understand it correctly - and this output reminds me a couple of the old PhEDEx APIs - where `LFNArrayRef` is an ordered list of attributes to be used as a reference for the `LFNArray` list, such that:
* LFNArray[0] should be a `outputLFNs`
* LFNArray[1] should be a `lfn`
* LFNArray[2] should be a `skippedFiles`; and so on.

which clearly is not clear at all(!)

Maybe the expected outcome should be something like:
```
  "LFNArray": [
    "/store/unmerged/logs/prod/2016/2/4/sryu_TaskChain_Data_wq_testt_160204_061048_5587/RECOCOSD/0000/0/7d7d41dc-cb02-11e5-833c-02163e00efd5-88-0-logArchive.tar.gz",
    "/store/unmerged/CMSSW_7_0_0_pre11/Cosmics/ALCARECO/DtCalib-RECOCOSD_TaskChain_Data_pile_up_test-v1/00000/ECCFE421-08CB-E511-9F4C-02163E017804.root",
    "/store/unmerged/CMSSW_7_0_0_pre11/Cosmics/ALCARECO/MuAlCalIsolatedMu-RECOCOSD_TaskChain_Data_pile_up_test-v1/00000/9665EB21-08CB-E511-9F4C-02163E017804.root"
  ],
  "LFNArrayRef": [
    "outputLFNs",
  ],
```

thus, reporting only the output files for this job (2 ALCARECO files and the logArchive).

vkuznet:
Alan, this getting longer and longer list of changes you desire to have which have nothing to do with original goal of this PR. If you agreed that it is not possible to address the memory footprint due to `deepcopy` I really suggest to come back to my original `cleanup` function which just strip out PFNs from final document. Everything else, if necessary, should go in my opinion into separate PRs/issues. I don't want to waste time on refactoring of something which we may or may not use, and making gigantic changes and refactoring code to the issues which far beyond of the scope of original problem on a monitoring side. Don't you agree?

My original code was only stripping off PFNs from final doc, how the rest of the structure is handling should be a business of this PR.

amaltaro:
Valentin, my last comment does not have anything to do with memory footprint. In short, it says that the current - and/or with this patch in - schema uploaded to WMArchive is not functional. In addition to that, I still see the input LFN in your dictionary.

From the original issue description - which had some discussion and was still evolving - this:
https://github.com/dmwm/WMCore/issues/10879#issuecomment-997928400

was one of the expected outcome/behavior for that issue.

vkuznet:
Alan, we can't just strip out LFNs since your pointer does not address all use-cases. For instance, primary reason for keeping LFNArray is to provide logArchive look-up on HDFS via Spark job. As such, I have no idea which LFNs are required for this use-case. In addition to that I have no clue how many types of LFNs FWJR contains. Does your list:
```
    "outputLFNs",
    "lfn",
    "skippedFiles",
    "inputLFNs",
    "fallbackFiles"
```
represents all possible LFNs created, used in WMCore workflows?

Therefore, if you want to address this properly please provide the following:
- full schema of FWJR, especially whatever attributes related to LFNs
- please outline use-cases which we should support, e.g. does `LFNArrayRef` should only contain `outputLFNs` or any other kind which will be required by use-cases we need to support
  - use cases should cover existing WMCore and data-ops ones which (again) I have no idea where to get
- does `LFNArray ` should keep all used LFNs or only subset, if later, which one

In my view the LFNs and PFNs presence in WMArchive should be addressed in different PRs. This PR only address removal of PFNs which will already reduce WMArchive size. Even though it is partial solution it is better to have it first, while addressing LFNs may require much broader discussion among different groups and supporting different use-cases.

vkuznet:
@amaltaro based on today's discussion please provide clear guidelines how to proceed with this PR. So far, it removes PFNs. But our discussion, to avoid ambiguities, it would be easy to remove LFNs as well. If this is the case, please say so and I'll adjust the PR accordingly.

amaltaro:
We have many FJR samples in here:
https://github.com/dmwm/WMCore/tree/master/test/python/WMCore_t/Services_t/WMArchive_t/FWJRSamples

which all date back from 6 years ago. They should be fairly representative, even though I wanted to see if we could have one with `skippedFiles`.

In any case, would you be able to provide us with the following samples:
* ProcessingSuccessFwjr.json
* MergeSuccessFwjr.json

converted to WMArchive format with a) this patch applied; and b) without this patch? So, 4 json files in total.

vkuznet:
Here we go:
- no patch
  - [MergeSuccessFwjr-pfns.json](https://gist.github.com/vkuznet/d30a3618fc318649bee3bf866d3c56d0)
  - [ProcessingSuccessFwjr-pfns.json](https://gist.github.com/vkuznet/8b26415c7899154c7b0b3b0acdb35e86)
- with current PR patch
  - [MergeSuccessFwjr-nopfns.json](https://gist.github.com/vkuznet/a6d18dd88946f12da6ac3d2326879c6e)
  - [ProcessingSuccessFwjr-nopfns.json](https://gist.github.com/vkuznet/f766e07671b5e89c612f038839e13cbb)

vkuznet:
And, here is test code to produce the aforementioned data:
```
#!/usr/bin/env python

import sys
import json
from WMCore.Services.WMArchive.DataMap import createArchiverDoc

fwjr = json.load(open(sys.argv[1]))
doc = {'id': 123, 'doc': fwjr}
data = createArchiverDoc(doc)
print(json.dumps(data))
```
The code can be run as following (assuming you save it as `test-wma.py`):
```
./test-wma.py test/python/WMCore_t/Services_t/WMArchive_t/FWJRSamples/MergeSuccessFwjr.json | jq > MergeSuccessFwjr-pfns.json
```

amaltaro:
Valentin, as discussed on Monday, this is indeed removing the PFNs from the output document and it's good to go. Eventually we should review how the list of files (input/output, lfns/pfns) is structured in the WMArchive document.

We should backport this fix to 2.0.2_wmagent and patch all the production agents running the latest version.

PR #11002, opened 2022-02-21
make sure that created voms proxy is 2048 bits

### LINKED ISSUE #11005: WMCore/Credential creates 1024-bit proxies
**Impact of the bug**
services using WMCore/Credential create 1024 bits proxies, which are not good anymore.
There is no celear evidence of problems since other services (HTCondor e.g.) create new proxies
with proper length, but lately some FTS xfers are demonstrated to have 1024-bit proxy
which result in failed transers, see this thread
https://cms-talk.web.cern.ch/t/repost-fts-proxy-issues-it-uses-1024-bits/6827
and even if the full proxy chain in those FTS transfers is not understood, having a fully 2048 bit
chain on our side can only help, if nothing else, to avoid confusion

**Describe the bug**
current code forces a 1024 bit proxy when applying VOMS extensions  to credential retrieved from myproxy
https://github.com/dmwm/WMCore/blob/fbca8729b8aaabb646e87139453dfac70eae7f70/src/python/WMCore/Credential/Proxy.py#L701

**How to reproduce it**
no need

**Expected behavior**
proxies should be 2048 bit to be accepted by CentOS8 servers

belforte:
Fixes https://github.com/dmwm/WMCore/issues/11005

with reference to discussion in https://cms-talk.web.cern.ch/t/repost-fts-proxy-issues-it-uses-1024-bits/6827

### DIFF
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -698,7 +698,7 @@ class Proxy(Credential):
         cmdList = []
         cmdList.append('env')
         cmdList.append('X509_USER_PROXY=%s' % proxy)
-        cmdList.append('voms-proxy-init -noregen -voms %s -out %s -bits 1024 -valid %s %s'
+        cmdList.append('voms-proxy-init -noregen -voms %s -out %s -bits 2048 -valid %s %s'
                        % (voAttribute, proxy, vomsValid, '-rfc' if isRFC  else ''))
         cmd = ' '.join(cmdList)
         msg, _, retcode = execute_command(self.setEnv(cmd), self.logger, self.commandTimeout)
--- a/src/python/WMCore/Credential/Proxy.py
+++ b/src/python/WMCore/Credential/Proxy.py
@@ -698,6 +698,7 @@ class Proxy(Credential):
         cmdList = []
         cmdList.append('env')
         cmdList.append('X509_USER_PROXY=%s' % proxy)
+        # make sure to make a 2048-bit proxy (for CentOS 8 based servers)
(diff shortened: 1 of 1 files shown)


belforte:
I have tested the -bits 2048 option in CRAB TaskWorker and works as expected. As noted in that CMSTalks thread, this makes no difference in the end for CRAB since HTCondor creates new proxies on sched and WN's and makes them 2048 bits already, but it will help mind sanity to have all proxy files with same key length.

amaltaro:
@belforte thanks for providing this fix.
Could you please create a GH issue that this PR is meant to resolve with a brief summary? If you need help with the templates, here you can find a short documentation on that: https://github.com/dmwm/WMCore/blob/master/CONTRIBUTING.rst#how-to-contribute-to-wmcore

Once you have it created, please update the PR description as well such that the issue gets automatically fixed once this gets merged.

belforte:
@amaltaro note that changing PR description is not sufficient to automatically close an issue. The PR must contain a commit which says "fix #...." or similar. I can't change the comment for past commits, but am looking into a new one

belforte:
Done

amaltaro:
@belforte sorry for the extra work, Stefano. The PR template started with something like:
```
Fixes https://github.com/dmwm/WMCore/issues/11005
```
which does the job without the need for a special commit message (I just updated this description accordingly).

Now we have a new issue, the double commit :-( Would you mind squashing those into a single one not to pollute the commit history? If needed, here is a reference that I always recommend: https://steveklabnik.com/writing/how-to-squash-commits-in-a-github-pull-request

belforte:
would have been easier if I had started on my fork instead of having edited the code in the browser as a shortcut. let's start over

PR #11004, opened 2022-02-21
fix uninitialized templates variable in workflow injection script

### LINKED ISSUE #11003: workflow injection script has an uninitialized variable
**Impact of the bug**
Can't inject workflows

**Describe the bug**
Script fails due to the presence of a code path where the templates variable is uninitialized before use

hufnagel:
Fixes #11003

#### Status
tested

#### Description
Just added a one-liner to initialize templates in the affected code path


amaltaro:
Thanks for the bugfix, Dirk.

PR #11007, opened 2022-02-22
Add mockMongoDB to MSUnmerged and MongoDB Class init optoins

### LINKED ISSUE #10978: MSUnmerged: Broken MongodB related unittests
**Impact of the bug**
MSUnmerged

**Describe the bug**
Not exactly a bug. Until now we were having a quite unstable setup of MongoDB related unit tests for both `MSUnmerged` and `MSOutput` as well. Since we now know that the `mogomock` library [1] is doing the job of simulating a real database connection quite well [2] we need to fix these [3] tests too.  

[1]
https://github.com/mongomock/mongomock

[2]
https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12755/artifact/artifacts/PullRequestReport.html

[3]
https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12754/testReport/junit/MSUnmerged_t/MSUnmergedTest/testCutPath/

https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12754/testReport/junit/MSUnmerged_t/MSUnmergedTest/testFilterInclExclDirectories/

https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12754/testReport/junit/MSUnmerged_t/MSUnmergedTest/testFilterExclDirectories/

https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12754/testReport/junit/MSUnmerged_t/MSUnmergedTest/testFilterInclDirectories/

https://cmssdt.cern.ch/dmwm-jenkins/view/All/job/DMWM-WMCore-PR-test/12754/testReport/junit/MSUnmerged_t/MSUnmergedTest/testPlineUnmerged/

**How to reproduce it**
All MongoDB related unit tests for MSUnmerged are failing due to lack of MongoDB server at the nodes executing the tests.

**Expected behavior**
Have a decently constructed MongoDB emulation so that  all the tests pass.

N/A

todor-ivanov:
Fixes #10978 

#### Status
Ready

#### Description
With the current PR, we add one additional option to both MSUnmerged and MongoDB classes so that we can request the creation of a mocked MongoDB client for the Services' unit tests. The default value for this parameter is set to `False`. 

YES

#### Related PRs
None

UPDATE: depends on this new spec: https://github.com/cms-sw/cmsdist/pull/7598

### DIFF
--- a/src/python/WMCore/Database/MongoDB.py
+++ b/src/python/WMCore/Database/MongoDB.py
@@ -9,6 +9,7 @@ from __future__ import division, print_function
 from builtins import str, object
 
 from pymongo import MongoClient, errors, IndexModel
+import mongomock
 
 
 class MongoDB(object):
@@ -24,6 +25,7 @@ class MongoDB(object):
                  collections=None,
                  testIndexes=False,
                  logger=None,
+                 mockMongoDB=False,
                  **kwargs):
         """
         :databases:   A database Name to connect to
@@ -41,8 +43,12 @@ class MongoDB(object):
         self.server = server # '127.0.0.1'
         self.port = port # 8230
         self.logger = logger
+        self.mockMongoDB = mockMongoDB
         try:
-            if replicaset:
+            if mockMongoDB:
+                self.client = mongomock.MongoClient()
+                self.logger.info("NOTICE: MongoDB is set to use mongomock, instead of real database.")
+            elif replicaset:
                 self.client = MongoClient(self.server, self.port, replicaset=replicaset, **kwargs )
             else:
                 self.client = MongoClient(self.server, self.port, **kwargs)
@@ -156,7 +162,8 @@ class MongoDB(object):
         """
         try:
             setattr(self, db, self.client[db])
-            self._dbTest(db)
+            if not self.mockMongoDB:
+                self._dbTest(db)
         except errors.ConnectionFailure as ex:
             msg = "Could not connect to MongoDB server for database: %s\n%s\n" % (db, str(ex))
             msg += "Giving up Now."
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
@@ -112,6 +112,7 @@ class MSUnmerged(MSCore):
         self.msConfig.setdefault("mongoDB", 'msUnmergedDB')
         self.msConfig.setdefault("mongoDBRetryCount", 3)
         self.msConfig.setdefault("mongoDBReplicaset", None)
+        self.msConfig.setdefault("mockMongoDB", False)
 
         msUnmergedIndex = IndexModel('name', unique=True)
         msUnmergedDBConfig = {
@@ -121,6 +122,7 @@ class MSUnmerged(MSCore):
             'replicaset': self.msConfig['mongoDBReplicaset'],
             'logger': self.logger,
             'create': True,
+            'mockMongoDB': self.msConfig['mockMongoDB'],
             'collections': [('msUnmergedColl', msUnmergedIndex)]}
 
         mongoClt = MongoDB(**msUnmergedDBConfig)
--- a/test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
+++ b/test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
@@ -67,7 +67,8 @@ def getBasicRSEData():
             "/store/unmerged/express/prod/2020/1/12/log8.tar",
             "/store/unmerged/express/prod/2020/1/12/log9.tar",
             "/store/unmerged/alan/prod/2022/1/12/log10.tar"]
-    rse = {"counters": {"dirsToDeleteAll": 0},
+    rse = {"name": "T2_TestRSE",
+           "counters": {"dirsToDeleteAll": 0},
            "dirs": {"allUnmerged": set(),
                     "toDelete": {},
                     "protected": []},
@@ -124,7 +125,8 @@ class MSUnmergedTest(unittest.TestCase):
                          'wmstatsUrl': 'https://cmsweb-testbed.cern.ch/wmstatsserver',
                          'dirFilterIncl': [],
                          'dirFilterExcl': [],
-                         'emulateGfal2': True}
+                         'emulateGfal2': True,
+                         'mockMongoDB': True}
 
         self.creds = {"client_cert": os.getenv("X509_USER_CERT", "Unknown"),
                       "client_key": os.getenv("X509_USER_KEY", "Unknown")}
@@ -140,7 +142,7 @@ class MSUnmergedTest(unittest.TestCase):
         self.msUnmerged = MSUnmerged(self.msConfig)
         self.msUnmerged.rucioConMon = RucioConMonEmul()
         self.msUnmerged.wmstatsSvc = WMStatsServerEmul()
-        self.msUnmerged.resetCounters()
+        self.msUnmerged.resetServiceCounters()
         self.msUnmerged.rucio = Rucio.Rucio(self.msConfig['rucioAccount'],
                                             hostUrl=self.rucioConfigDict['rucio_host'],
                                             authUrl=self.rucioConfigDict['auth_host'],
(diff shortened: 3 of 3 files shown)


src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -87,12 +87,12 @@ class MSUnmerged(MSCore):
     """
 
     # @profile
-    def __init__(self, msConfig, logger=None):
+    def __init__(self, msConfig, logger=None, mockMongoDB=False):
```
amaltaro, line 90: How about reading this key `mockMongoDB` from the `msConfig` object? Then you can replace it when calling the super class as well.

src/python/WMCore/Database/MongoDB.py
```diff
@@ -156,7 +161,8 @@ def _dbConnect(self, db):
         """
         try:
             setattr(self, db, self.client[db])
-            self._dbTest(db)
+            if not self.mockMongoDB:
```
amaltaro, line 164: Todor, I'd suggest to move this check for `self.mockMongoDB` to the first line(s) in the `_dbTest` method.

amaltaro (commented):
These changes are looking pretty good Todor. I left a couple of comments with possible minor improvements for your consideration though.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: I was thinking about this possibility, too. I'll fix it.

src/python/WMCore/Database/MongoDB.py
todor-ivanov: Actually, this would make the code less readable. The `_dbTest`  method has no return statement, it is acting with the database with a `command` that is not supported by mongomock [1]. And actually we do not need to test the database at all while working with `mongomock`, so I'd prefer to not even call the method.  


[1]
```
NotImplementedError: command is a valid Database method but is not implemented in Mongomock yet  
```

todor-ivanov:
Thank you for your review @amaltaro 
I have addressed your comments. Could you please take another look.

src/python/WMCore/Database/MongoDB.py
amaltaro: Yes, hence my suggestion to check it at the very first line of that method, to skip it when it's not actually needed (it's being emulated). 
To me, having that check at the very beginning of the method that is supposed to act differently according to the real vs emulation mode would make it more readable. This nested try/except block makes it not a the best spot IMO.
Code is still fully functional though, so if you feel like keeping it like that, let's keep it.

vkuznet (changes requested):
I strongly advise to stop using dynamic keys in dictionaries. This practice leads to plenty of issues, from growth of memory to non generalize code to parse such data. Unless I hear strong argument why do you need dynamic keys in a schema I would vote to not approve the request.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -121,6 +122,7 @@ def __init__(self, msConfig, logger=None):
             'replicaset': self.msConfig['mongoDBReplicaset'],
             'logger': self.logger,
             'create': True,
+            'mockMongoDB': self.msConfig['mockMongoDB'],
```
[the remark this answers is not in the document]
todor-ivanov: Every configuration parameter which enters msConfig has its default value set at object creation time. The line to set the default value is at:
https://github.com/dmwm/WMCore/pull/11007/files#diff-5794b61bc9d37524abcbd5ec373f318ead63e018f0bfb54ab66011106f8a8540R115

So the protection you suggest here is redundant.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -121,6 +122,7 @@ def __init__(self, msConfig, logger=None):
             'replicaset': self.msConfig['mongoDBReplicaset'],
```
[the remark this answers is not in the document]
todor-ivanov: same as above

test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
```diff
@@ -157,23 +159,46 @@ def testPlineUnmerged(self):
...
+                                    "filesDeletedSuccess": 0,
+                                    "filesDeletedFail": 0,
+                                    "dirsDeletedSuccess": 0,
+                                    "dirsDeletedFail": 0,
+                                    "gfalErrors": {}},
+                       'files': {'allUnmerged': mock.ANY,
+                                 'deletedFail': set(),
+                                 'deletedSuccess': set(),
+                                 'protected': {},
+                                 'toDelete': {'/store/unmerged/Phase2HLTTDRSummer20ReRECOMiniAOD/DYToLL_M-50_TuneCP5_14TeV-pythia8/FEVT/FlatPU0To200_pilot_111X_mcRun4_realistic_T15_v1-v2': mock.ANY,
```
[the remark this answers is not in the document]
todor-ivanov: Hi @vkuznet So there is argumentation behind the document schema (which is here BTW: https://github.com/dmwm/WMCore/blob/14e381381dd4639a26fb61807c2f04cd93cf9807/src/python/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py#L40-L69) . The dynamic keys are needed as a place holder for the generators of filters to the final list of files to be deleted. We have extremely long lists per RSE and duplicating the `toDelete` lists inside the object was only leading to extreme memory footprint. Instead we preserve the full list only at one place and we dynamically create filters at runtime based on the current conditions (what is and what is not protected), then we keep only the filters. You see it 'hardcoded' here as such only because this is a unit test with predefined data for which I have tested in advance what should be the name to be expect here. In the real code those filters are generated at runtime, hence the dynamic names of the keys to hold them.

src/python/WMCore/Database/MongoDB.py
todor-ivanov: Agreed, but the method that is supposed to act differently based on the `mockMongoDb` flag is not `_dbTest`, but rather `_dbConnect`, because we are not actually connecting to any database we are mocking it instead.

test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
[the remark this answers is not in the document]
vkuznet: Please show me concrete example. And, justify why do you need to keep LFNs per RSE in single document? Why not to use separate document for that and then refer (a la foreign key in RDBMS world) to it. What are those generators and filters? And, which duplication you're talking about. Sorry, so far nothing is clear to me how you convert your static schema to dynamic one, by avoiding some duplication via dynamic keys in a dictionary.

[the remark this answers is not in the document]
vkuznet: And, why do you keep schema within specific class? Can you define it separately and load it in a class? Here we are having similar situation to `Lexicon.py` issue, which now we're trying to share among different code bases and it is extremely difficult since of embedded nature of the schema. The schema files defines the data. As such they should not be part of concrete implementation and it is better to keep it separately to allow to have different implementation for that particular schema.

src/python/WMCore/Database/MongoDB.py
amaltaro: I am suggesting an alternative to the code you are proposing. If you say it's only `_dbConnect` that should not be called when mocking MongoDB, then it's probably better to update this code, no?

test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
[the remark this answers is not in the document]
amaltaro: Given that this comment refers to a test module, and that it goes beyond the scope of this PR, I would suggest to follow this up offline. Or in one of our future reviews of the system, services, etc. What do you all think?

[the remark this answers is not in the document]
todor-ivanov: Completely agree with you here @amaltaro

src/python/WMCore/Database/MongoDB.py
todor-ivanov: I said  `_dbConnect` should behave differently when called with and without mocking. But it still needs to be called, because it sets the database attribute to the object on line 163. This is an attribute of the objects that  other pieces of the code rely on. The difference is that when mocking the database then no database test is needed, so no need to call the `_dbTest` method. I am sure this is the right place for this check.

amaltaro: Then the `_dbConnect` comment is meaningless for what we are discussing here, which is what the best place is for checking whether mongodb is being mocked or not. Anyhow, it's time to move on.

amaltaro (changes requested):
Todor, please see my comments along the code. Once you push that change, feel free to squash commits accordingly and it should be ready to go. Thanks.

todor-ivanov:
Thanks @amaltaro.  The relevant log message was added and the commits squashed!

amaltaro (approved):

amaltaro:
Thanks Todor! As discussed over slack, please make sure to keep test/aesthetic changes *separated* in different commits from the actual logical source code changes.

PR #11010, opened 2022-02-24
Add pyjwt dependency

### LINKED ISSUE #10940: Add support for TOKEN based authentication

vkuznet:
Fixes #10940, #10939

#### Status
ready

#### Description
Add necessary dependency on PyJWT library for token based authentication

YES

#### Related PRs
#10939 

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,3 +41,4 @@ retry~=0.9.2                  # wmagent,wmagent-devtools,reqmgr2,reqmon,global-w
 rucio-clients~=1.25.5         # wmagent,global-workqueue,reqmgr2ms
 Sphinx~=4.0.3                 # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms
 SQLAlchemy~=1.3.3             # wmagent
+pyjwt~=2.3.0                  # wmagent,wmagent-devtools,reqmgr2,reqmon,acdcserver,global-workqueue,reqmgr2ms


vkuznet:
I found that `requirements_py3.txt` is no longer in master branch, but I still had it in my local area. Therefore, I removed original commit to this file to avoid conflict. Now the change only applied to `requirements.txt` file.

amaltaro:
Given that this new library is introduced in the PR https://github.com/dmwm/WMCore/pull/10939 , I would say that that PR should be providing these requirements changes as well. Once we merge that PR, then we get a new dependency in the stack as well (like an atomic operation ;))

vkuznet:
ok, I added pyjwt dependency in this commit https://github.com/dmwm/WMCore/pull/10939/commits/97e85176b8b1cc7ddb3b8e72d86ae2987a1f3463 to https://github.com/dmwm/WMCore/pull/10939

amaltaro:
Given that this change has also been provided here: https://github.com/dmwm/WMCore/pull/10939
with the actual baseline code to using tokens, I think it's better if we close this PR.

@vkuznet please reopen it in case I am missing anything.

PR #10780, opened 2021-08-31
CouchDB 3 Migration: Remove all_or_nothing flag

phenomax:
#### Status
not-tested

#### Description
As part of the CouchDB Migration to 3.x ( #8853 ), the `all_or_nothing` flag has to be removed because 
it's no longer present (see https://github.com/apache/couchdb/issues/1235 for a discussion).

#### Is it backward compatible (if not, which system it affects?)
NO

#### External dependencies / deployment changes
It depends on CouchDB 3.x

PR #11011, opened 2022-02-25
[CouchDB] Remove CouchDB all_or_nothing option for bulk API

### LINKED ISSUE #11009: CouchDB no longer supports all_or_nothing option for bulk_docs API
**Impact of the new feature**
WMCore in general (likely used by local/global workqueue and wmstats)

This `all_or_nothing` bulk operation option has been removed in CouchDB 2.0, see:
https://docs.couchdb.org/en/latest/whatsnew/2.0.html#upgrade-notes

As we migrate to CouchDB 3.x, we need to update our code such that we can keep remain as close as possible to an atomic transaction, even when multiple documents are created or updated.

According to this documentation: https://docs.couchdb.org/en/latest/api/database/bulk-api.html#db-bulk-docs
we can make a nested dictionary/json with all the documents that we want to create or update in CouchDB. This should bring us closer to the `all_or_nothing` deprecated functionality (if not the same behavior).

So, the expected changes with this ticket are:
* remove all the places defining this `all_or_nothing` flag
* and update our CMSCouch code performing bulk operations such that it provides the correct data format to the `bulk_docs` API, a list of documents to be created/updated.

Max provided an initial proposal for fixing this, in: https://github.com/dmwm/WMCore/pull/10780

**Additional context**
A sub-task of https://github.com/dmwm/WMCore/issues/8853

amaltaro:
Fixes #11009 
Superseeds https://github.com/dmwm/WMCore/pull/10780

#### Status
ready

#### Description
CouchDB `all_or_nothing` option is no longer available for bulk APIs. This PR stops using that option.
Note that if there is a conflict with one or more of the documents being updated, the conflict is only resolved if a callback function is passed over.

YES

#### Related PRs
Superseeds https://github.com/dmwm/WMCore/pull/10780
by adding a few extra checks in the tests and other minor improvements.

Many other changes have been provided in this PR: https://github.com/dmwm/WMCore/pull/11001

Changes required for CouchDB 3.x, but can also be merged while we are at CouchDB 1.6.

### DIFF
--- a/src/python/WMCore/WorkQueue/WorkQueueBackend.py
+++ b/src/python/WMCore/WorkQueue/WorkQueueBackend.py
@@ -158,7 +158,6 @@ class WorkQueueBackend(object):
         self.insertWMSpec(units[0]['WMSpec'])
         newUnitsInserted = []
         for unit in units:
-
             # cast to couch
             if not isinstance(unit, CouchWorkQueueElement):
                 unit = CouchWorkQueueElement(self.db, elementParams=dict(unit))
@@ -171,10 +170,11 @@ class WorkQueueBackend(object):
             if unit._couch.documentExists(unit.id):
                 self.logger.info('Element "%s" already exists, skip insertion.' % unit.id)
                 continue
-            else:
-                newUnitsInserted.append(unit)
+
+            newUnitsInserted.append(unit)
             unit.save()
-            unit._couch.commit(all_or_nothing=True)
+            # FIXME: this is not performing bulk request, but single document commits(!)
+            unit._couch.commit()
 
         return newUnitsInserted
 
--- a/src/python/WMCore/Database/CMSCouch.py
+++ b/src/python/WMCore/Database/CMSCouch.py
@@ -279,7 +279,7 @@ class Database(CouchDBRequests):
         its internal retval
 
         key, value pairs can be used to pass extra parameters to the bulk doc api
-        See http://wiki.apache.org/couchdb/HTTP_Bulk_Document_API
+        See https://docs.couchdb.org/en/latest/api/database/bulk-api.html#db-bulk-docs
 
         TODO: restore support for returndocs and viewlist
 
--- a/test/python/WMCore_t/Database_t/CMSCouch_t.py
+++ b/test/python/WMCore_t/Database_t/CMSCouch_t.py
@@ -283,13 +283,13 @@ class CMSCouchTest(unittest.TestCase):
         self.assertEqual(answer[0]['ok'], True)
         self.assertEqual(answer[1]['error'], 'conflict')
 
-        # all_or_nothing mode ignores conflicts
+        # bulk commit with one conflict unresolved
         self.db.queue(Document(id = "2", inputDict = doc))
         self.db.queue(Document(id = "2", inputDict = {'foo':1234, 'bar':456}))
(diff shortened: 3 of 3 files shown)


src/python/WMCore/WorkQueue/WorkQueueBackend.py
```diff
@@ -158,23 +158,23 @@ def insertElements(self, units, parent=None):
...
         for unit in units:
-
             # cast to couch
             if not isinstance(unit, CouchWorkQueueElement):
                 unit = CouchWorkQueueElement(self.db, elementParams=dict(unit))
 
+            if unit._couch.documentExists(unit.id):
+                self.logger.info('Element "%s" already exists, skip insertion.' % unit.id)
+                continue
+
```
todor-ivanov, line 168: Hi @amaltaro I seem to fail grasping the need of reordering of these checks (The one above and the `if parent` one below.) Could you please shed some light?

todor-ivanov (approved):
Thanks @amaltaro The change looks good to me. I had only one comment inline. It is not requesting any change though, it is just for clarification purposes.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
amaltaro: I moved this check up, because when the document already exists, we do not (re-)insert it into CouchDB, neither we return this document/object back from this method. So, it looks to me the sooner we check and skip this document, the better. 
Of course, it could be a problem if this list of documents is still used later in the chain somehow, which I am not 100% sure.

amaltaro: But I am going to revert it just to reduce the amount of changes required for the couch migration. Thanks Todor.

todor-ivanov: Thanks Alan! Just to note, I was not requesting a change to it, I was asking more like... for me to understand it.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
```diff
@@ -171,10 +170,11 @@ def insertElements(self, units, parent=None):
             if unit._couch.documentExists(unit.id):
                 self.logger.info('Element "%s" already exists, skip insertion.' % unit.id)
                 continue
-            else:
-                newUnitsInserted.append(unit)
+
+            newUnitsInserted.append(unit)
             unit.save()
-            unit._couch.commit(all_or_nothing=True)
+            # FIXME: this is not performing bulk request, but single document commits(!)
```
todor-ivanov, line 176: Why is it needed to change the behavior here, from `all_or_nothing=True` to single document uploads?

todor-ivanov (approved):
Thanks @amaltaro I had only one question inline. But the changes look good to me.

src/python/WMCore/WorkQueue/WorkQueueBackend.py
todor-ivanov: Ok, let me answer  myself, you've already explained it all in the issue comments.
> CouchDB all_or_nothing option is no longer available for bulk APIs

Sorry for the noise! :)

vkuznet (approved):
Sorry for delay, I think I missed the email with this request.

amaltaro:
And there is nothing else to be done here, merging.

PR #11015, opened 2022-02-28
Copy DQM ROOT file to EOS and register entry in DQMUpload executor

### LINKED ISSUE #10287: The new DQM GUI file management
**Impact of the new feature**
This request affects all systems that are responsible for harvested DQM data being uploaded to the DQM GUIs. This includes T0 processed DQM data and RelVal/reprocessing DQM data.

We're deploying a new, upgraded version of the DQM GUI tool. The procedure which notifies the DQM GUI about the new DQM files is different in a new version. We would like this new procedure to be used along side the the old, `visDQMUpload` based DQM file upload.

Now, the DQM data is uploaded to the DQM GUIs using a tool called `visDQMUpload`. New procedure requires this process to be split into two stages: 

* Making a copy of the new DQM file to EOS.
* Making an HTTP POST request to the new DQM GUI to notify it about the presence of a new file.

If required, a facade could be provided by us (DQM) that would have exactly the same interface as `visDQMUpload`. In such case, we would only like you to call the facade script (`visDQMUpload_new`) alongside the old one. 

No viable, future proof alternatives were found.

**Additional context**
Bellow is a diagram that represents the current Offline DQM file movement:

current-dqm-diagram

Bellow is a diagram that represents the desired Offline DQM file movement, after the changes mentioned in this request:

new-dqm-diagram2

khurtado:
Fixes #10287

#### Status
New changes being tested

#### Description
This implements the new DQM Gui method described in #10287

- Copy DQM root file to EOS
- Register entry to new DQMGUI: Note this does not upload any files. It registers one root file per call, right after the file has been uploaded to EOS.  
- Delete files uploaded to EOS if registration fails
- Use value of forceRunNumber for multiRun, to replicate current behavior by old visDQMGUI

YES

### DIFF
--- a/src/python/WMCore/WMExceptions.py
+++ b/src/python/WMCore/WMExceptions.py
@@ -184,7 +184,12 @@ WM_JOB_ERROR_CODES = {-1: "Error return without specification.",
                       60409: "Timeout in stage out of log files during log collection (WMAgent).",  # (WMA)
                       60450: "No output files present in the report",  # (WMA)
                       60451: "Output file lacked adler32 checksum (WMAgent).",  # (WMA)
+                      70317: "Indefinite hangout while staging out DQM root file to EOS",  # (WMA)
                       70318: "Failure in DQM upload.",
+                      70319: "HTTP POST to DQM register url failed.",  # (WMA)
+                      70320: "Error while retrieving run number from job definition.",  # (WMA)
+                      70321: "Indefinite hang while deleting file from EOS.",  # (WMA)
+                      70322: "General failure while deleting file from EOS.",  # (WMA)
                       70452: "No run/lumi information in file (WMAgent).",
                       71101: "No sites are available to submit the job because the location of its input(s) do not pass the site whitelist/blacklist restrictions (WMAgent).",
                       71102: "The job can only run at a site that is currently in Aborted state (WMAgent).",
--- a/src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
+++ b/src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
@@ -23,6 +23,7 @@ from mimetypes import guess_type
 try:
     # python2
     import urllib2
+
     HTTPError = urllib2.HTTPError
     OpenerDirector = urllib2.OpenerDirector
     Request = urllib2.Request
@@ -30,6 +31,7 @@ try:
 except:
     # python3
     import urllib.request
+
     HTTPError = urllib.error.HTTPError
     OpenerDirector = urllib.request.OpenerDirector
     Request = urllib.request.Request
@@ -46,7 +48,7 @@ from WMCore.Algorithms.Alarm import Alarm, alarmHandler
 from WMCore.Storage.DeleteMgr import DeleteMgr, DeleteMgrError
 from WMCore.Storage.FileManager import DeleteMgr as NewDeleteMgr
 import WMCore.Storage.StageOutMgr as StageOutMgr
-import WMCore.WMRuntime.Bootstrap as Bootstrap
+import time
 
 class DQMUpload(Executor):
     """
@@ -310,10 +312,13 @@ class DQMUpload(Executor):
 
         return (result.headers, data)
 
-    def _uploadToEOS(self, analysisFile, stepLocation):
+    def _uploadToEOS(self, stepLocation, analysisFile):
         """
-        upload  an analysis file to EOS
-        :return: boolean with upload status
+        Upload  an analysis file to EOS
+
+        :param stepLocation: Path location for step
+        :param analysisFile: analysisFile from WMCore step report
+        :return: bool, with upload status
         """
         overrides = {}
         if hasattr(self.step, 'override'):
@@ -323,7 +328,7 @@ class DQMUpload(Executor):
         eosStageOutParams['option'] = overrides.get('option', "--wma-disablewriterecovery")
         eosStageOutParams['phedex-node'] = overrides.get('eos-phedex-node', "T2_CH_CERN")
         eosStageOutParams['lfn-prefix'] = overrides.get('eos-lfn-prefix',
-                                                         "root://eoscms.cern.ch/%s" % self.registerEOSPrefix)
+                                                        "root://eoscms.cern.ch/%s" % self.registerEOSPrefix)
 
         # Switch between old and newstageOut manager.
         # Use old by default
@@ -340,15 +345,15 @@ class DQMUpload(Executor):
             # new style
             logging.info("DQMUpload is using new StageOut code for EOS Copy")
             eosmanager = WMCore.Storage.FileManager.StageOutMgr(
-                    retryPauseTime=self.retryDelay,
-                    numberOfRetries=self.retryCount,
-                    **eosStageOutParams)
+                retryPauseTime=self.retryDelay,
+                numberOfRetries=self.retryCount,
+                **eosStageOutParams)
 
         eosFileInfo = {'LFN': self.getAnalysisFileLFN(analysisFile),
-                       'PFN': os.path.join(stepLocation, 
+                       'PFN': os.path.join(stepLocation,
                                            os.path.basename(analysisFile.fileName)),
-                       'PNN': None,
-                       'GUID': None
+                       'PNN': None,  # Assigned in StageOutMgr
+                       'GUID': None  # Assigned in StageOutMgr
                        }
 
         msg = "Writing DQM root files to CERN EOS with retries: %s and retry pause: %s"
@@ -369,16 +374,23 @@ class DQMUpload(Executor):
     def getAnalysisFileLFN(self, analysisFile):
         """
         Construct an LFN for an analysisFile
+
+        :param analysisFile: analysisFile from WMCore step report
+        :return: str, analysis file LFN
         """
         # If lfn start with '/', make it relative to it
         lfnFile = os.path.relpath(analysisFile.lfn, start='/')
         lfn = os.path.join(self.registerLFNBase, lfnFile)
-    
+
         return lfn
 
     def _register(self, registerURL, args):
         """
         POST request to register URL
+
+        :param registerURL: str, url for registration
+        :param args: dict, POST arguments
+        :return: result object from opening  the request
         """
 
         ident = "WMAgent python/%d.%d.%d" % sys.version_info[:3]
@@ -407,7 +419,7 @@ class DQMUpload(Executor):
             result = opener.open(datareq)
 
         return result
-    
+
     def _registerAnalysisFile(self, stepName, analysisFile):
         """
         Register a file to new DQM GUi
@@ -416,13 +428,16 @@ class DQMUpload(Executor):
             [{"dataset": "/a/b/c", "run": "123456", "lumi": "0", "file": "/eos/cms/store/group/comm_dqm/DQMGUI_data/location/file.root", "fileformat": 1}]
         Set lumis to 0, as this is reproducing current per Run based root files
         - https://github.com/dmwm/WMCore/issues/10287#issuecomment-1052558061
-        DQM root files produced by Harvesting prosessing are type 1 (plain ROOT)
+        DQM root files produced by Harvesting processing are type 1 (plain ROOT)
         - https://github.com/cms-DQM/dqmgui#file-formats
+
+        :param stepName: str, Name of the step we are running through
+        :param analysisFile: WMCore.Configuration.ConfigSection, Analysis file in this step from WMCore step report
+        :return:
         """
         # Get task description
-        job = Bootstrap.loadJobDefinition()
-        jobBag = job.getBaggage()
-        datasetName = job.get('inputDataset', None)
+        jobBag = self.job.getBaggage()
+        datasetName = self.job.get('inputDataset', None)
 
         # Get runNumber, depending on the harvesting job mode (byRun vs multiRun)
         multiRun = getattr(jobBag, "multiRun", False)
@@ -431,14 +446,14 @@ class DQMUpload(Executor):
             runNumber = forceRunNumber
         else:
             try:
-                runNumber = list(job['mask']['runAndLumis'].keys())[0]
+                runNumber = list(self.job['mask']['runAndLumis'].keys())[0]
             except Exception as ex:
                 msg = "Error while retrieving run number from job definition:\n %s" % str(ex)
-                raise WMExecutionFailure(70320, "DQMUploadFailure", msg)    
+                raise WMExecutionFailure(70320, "DQMUploadFailure", msg)
 
         args = {}
         args['file'] = self.registerEOSPrefix + \
-                self.getAnalysisFileLFN(analysisFile)
+                       self.getAnalysisFileLFN(analysisFile)
         args['dataset'] = datasetName
         args['run'] = runNumber
         args['lumi'] = 0
@@ -449,10 +464,10 @@ class DQMUpload(Executor):
         msg += " => Filename: %s\n" % args['file']
         logging.info(msg)
 
-        try:
-            logException = True
-            for registerURL in self.registerURL.split(';'):
-                result = self._register(registerURL, [args])
+        for numRetry in range(self.retryCount + 1):
+            try:
+                logException = True
+                result = self._register(self.registerURL, [args])
                 msg = '  Status code: %s\n' % result.getcode()
                 if result.getcode() >= 400 or result.getcode() is None:
                     logException = False
@@ -460,13 +475,16 @@ class DQMUpload(Executor):
                 else:
                     msg = 'HTTP POST to register url finished succesfully with response:\n' + msg
                     logging.info(msg)
-        except Exception as ex:
-            msg = 'HTTP POST to register url failed! Error:\n%s' % str(ex)
-            if logException:
-                logging.exception(msg)
-            else:
-                logging.error(msg)
-            raise WMExecutionFailure(70319, "DQMUploadFailure", msg)
+                    break
+            except Exception as ex:
+                if numRetry == self.retryCount:
+                    msg = 'HTTP POST to register url failed! Error:\n%s' % str(ex)
+                    if logException:
+                        logging.exception(msg)
+                    else:
+                        logging.error(msg)
+                    raise WMExecutionFailure(70319, "DQMUploadFailure", msg)
+                time.sleep(self.retryDelay)
 
         return
 
@@ -474,6 +492,9 @@ class DQMUpload(Executor):
         """
         Delete a file uploaded to EOS.
         I.e.: If we upload a DQM root file but registration failed for some reason
+
+        :param eosFileLFN: str, LFN file
+        :return:
         """
         overrides = {}
         if hasattr(self.step, 'override'):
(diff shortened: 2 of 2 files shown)


khurtado:
@amaltaro I think this *almost is ready for review. What is pending before the code review is getting a new permanent EOS area, managed by used, rather than the unmerged area.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -50,6 +56,14 @@ class DQMUpload(Executor):
 
     """
 
+    def __init__(self):
+        super(DQMUpload, self).__init__()
+        self.retryDelay = 300
+        self.retryCount = 3
+        self.registerLFNBase = '/store/unmerged/DQMGUI'
+        self.registerEOSPrefix = '/eos/cms'
+        self.registerURL = 'https://cmsweb-testbed.cern.ch/dqm/offline-test-new/api/v1/register'
```
amaltaro, line 65: Can we derive this URL from the url used for uploading the root file to the DQM Gui?

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+        """
+        overrides = {}
+        if hasattr(self.step, 'override'):
+            overrides = self.step.override.dictionary_()
+        eosStageOutParams = {}
+        eosStageOutParams['command'] = overrides.get('command', "xrdcp")
+        eosStageOutParams['option'] = overrides.get('option', "--wma-disablewriterecovery")
+        eosStageOutParams['phedex-node'] = overrides.get('eos-phedex-node', "T2_CH_CERN")
+        eosStageOutParams['lfn-prefix'] = overrides.get('eos-lfn-prefix',
+                                                         "root://eoscms.cern.ch/%s" % self.registerEOSPrefix)
```
amaltaro, line 326: It seems not to be indented with the line above(?)

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+            logging.info("DQMUpload is using new StageOut code for EOS Copy")
+            eosmanager = WMCore.Storage.FileManager.StageOutMgr(
+                    retryPauseTime=self.retryDelay,
+                    numberOfRetries=self.retryCount,
+                    **eosStageOutParams)
+
+        eosFileInfo = {'LFN': self.getAnalysisFileLFN(analysisFile),
+                       'PFN': os.path.join(stepLocation, 
+                                           os.path.basename(analysisFile.fileName)),
+                       'PNN': None,
```
amaltaro, line 350: Should we fill this PNN out? Or it gets filled during the underlying stage out modules?

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+        """
+        # If lfn start with '/', make it relative to it
+        lfnFile = os.path.relpath(analysisFile.lfn, start='/')
+        lfn = os.path.join(self.registerLFNBase, lfnFile)
+    
+        return lfn
+
+    def _register(self, registerURL, args):
+        """
+        POST request to register URL
```
amaltaro, line 381: Please update this docstring as well.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+    
+    def _registerAnalysisFile(self, stepName, analysisFile):
+        """
+        Register a file to new DQM GUi
+        - https://github.com/cms-DQM/dqmgui
+        HTTP request body:
+            [{"dataset": "/a/b/c", "run": "123456", "lumi": "0", "file": "/eos/cms/store/group/comm_dqm/DQMGUI_data/location/file.root", "fileformat": 1}]
+        Set lumis to 0, as this is reproducing current per Run based root files
+        - https://github.com/dmwm/WMCore/issues/10287#issuecomment-1052558061
+        DQM root files produced by Harvesting prosessing are type 1 (plain ROOT)
```
amaltaro, line 419: typo in `processing`

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+        - https://github.com/cms-DQM/dqmgui#file-formats
+        """
+        # Get task description
+        job = Bootstrap.loadJobDefinition()
```
amaltaro, line 423: Kenyi, could you please check whether the job description is already available in this object? From:
https://github.com/dmwm/WMCore/blob/ee8c9fe76301997917e938b998393760893de7cf/src/python/WMCore/WMSpec/Steps/Executor.py#L93

it might be that we already have it under `self.job`

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
amaltaro: BTW, I think we should merge this code with what we have under the `upload` method to avoid duplication. Maybe create a new method named like `_setupHttp`.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+            runNumber = forceRunNumber
+        else:
+            try:
+                runNumber = list(job['mask']['runAndLumis'].keys())[0]
+            except Exception as ex:
+                msg = "Error while retrieving run number from job definition:\n %s" % str(ex)
+                raise WMExecutionFailure(70320, "DQMUploadFailure", msg)    
+
+        args = {}
+        args['file'] = self.registerEOSPrefix + \
```
amaltaro, line 440: It might be safer to use `os.path.join` to avoid issues with double or lack of slashes. The line looks short enough as well, maybe put it all in the same line.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+        args['fileformat'] = 1
+
+        msg = "HTTP Upload is about to start:\n"
+        msg += " => URL: %s\n" % self.registerURL
+        msg += " => Filename: %s\n" % args['file']
+        logging.info(msg)
+
+        try:
+            logException = True
+            for registerURL in self.registerURL.split(';'):
```
amaltaro, line 454: We should contact the DQM Gui and see whether the file registration should be done against a single server (for each workflow). If so, I'm very much in favor of removing this. Actually, it was added to the DQMUpload step many many years ago, I am not even sure whether it's still required and useful to anyone.

If we can remove the multi url manipulation, then we can likely try to remove this `logException` flag as well.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+                result = self._register(registerURL, [args])
+                msg = '  Status code: %s\n' % result.getcode()
+                if result.getcode() >= 400 or result.getcode() is None:
+                    logException = False
+                    raise Exception(msg)
+                else:
+                    msg = 'HTTP POST to register url finished succesfully with response:\n' + msg
+                    logging.info(msg)
+        except Exception as ex:
```
amaltaro, line 463: Given that we have a retry logic for the root file stage out, I think we should recycle the same logic (or something similar) to this POST requests. If we get this far, then we should try as hard as we can not to fail this job.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+        else:
+            # new style
+            logging.info("DQMUpload is using new StageOut code to delete files from EOS")
+            manager = NewDeleteMgr(retryPauseTime=self.retryDelay,
+                                   numberOfRetries=self.retryCount,
+                                   **eosStageOutParams)
+
+        logging.info("Deleting LFN: %s", eosFileLFN)
+        eosFileInfo = {'LFN': eosFileLFN,
+                           'PFN': None, # PFNs are assigned in the Delete Manager
```
amaltaro, line 510: Indentation needs to be fixed.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+                           'PNN': None, # PNNs are assigned in the Delete Manager
+                           'StageOutCommand': None}
+        try:
+            manager(fileToDelete=eosFileInfo)
+        except Alarm:
+            msg = "Indefinite hang while deleting file from EOS"
+            logging.error(msg)
+            raise WMExecutionFailure(70321, "DQMUploadStageOutTimeout", msg)
```
amaltaro, line 518: to avoid a long and unreadable traceback, I'd suggest to complete this raise line with `from None`. Same for the block below.

amaltaro (changes requested):
Kenyi, I left some comments along the code. And we might want to contact the DQM team as well to see what we have to support in terms of urls (do we need to support multi urls?)

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+
+        return eosFileInfo['LFN']
+
+    def getAnalysisFileLFN(self, analysisFile):
+        """
+        Construct an LFN for an analysisFile
+        """
+        # If lfn start with '/', make it relative to it
+        lfnFile = os.path.relpath(analysisFile.lfn, start='/')
+        lfn = os.path.join(self.registerLFNBase, lfnFile)
```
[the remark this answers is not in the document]
khurtado: I tried that first, but was surprised with the `self.registerLFNBase` being ignored if `analysisFile.lfn` started with `/`:

```
>>> os.path.join("/root", "/hello")
'/hello'
```

which is why I had to make sure the `analysisFile.lfn` has a relative form

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+            logging.error(msg)
+            raise WMExecutionFailure(70322, "DQMUploadFailure", msg)
+
+        return
+
+    def uploadToEOSAndRegister(self, stepName, stepLocation, analysisFile):
+        """
+        Copy DQM Root files to EOS and register files to new DQMGUI
+        """
+        eosFileLFN = self._uploadToEOS(analysisFile, stepLocation)
```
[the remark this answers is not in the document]
khurtado: 1. But `_uploadToEOS` finishes with a try/except block already, so it throws an exception if anything fails here:

https://github.com/dmwm/WMCore/blob/ee8c9fe76301997917e938b998393760893de7cf/src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py#L356-L365
2.  Sounds good to me, I will switch the position of the arguments

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: Do you meant `step.upload.URL`? Aren't they independent?

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: It's taken care by the `StageOutMgr`. Maybe I can add a comment with that?

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: I thought about that at first but even though both are POST requests, the `upload` method actually uploads file data content, while this doesn't, so I made them separated to distinguish the original purpose of each. What do you think?

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: Good catch! I will check

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: Same as one comment above, I would need to remove the starting `/` from the second argument in order to use `os.path.join`, otherwise we get:

```
>>> os.path.join("/eos/cms","/store/user/something")
'/store/user/something'
```

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: Good point! I just left a comment in the issue asking the DQM folks.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: Ah, I wonder how that happened. Thanks, will do

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
khurtado: That's fair. I will implement some retrial feature for this method

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +309,229 @@ def upload(self, url, args, filename):
...
+        except Exception as ex:
+            msg = "General failure while deleting file. Error: %s" % str(ex)
+            logging.error(msg)
+            raise WMExecutionFailure(70322, "DQMUploadFailure", msg)
+
+        return
+
+    def uploadToEOSAndRegister(self, stepName, stepLocation, analysisFile):
+        """
+        Copy DQM Root files to EOS and register files to new DQMGUI
```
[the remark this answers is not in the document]
khurtado: Done

khurtado:
@amaltaro I think I addressed most of the comments and left comments on the ones I didn't

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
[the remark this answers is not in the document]
amaltaro: and this is a surprise to me as well!

amaltaro (commented):
Kenyi, sorry for the delay reviewing this. Next time, please refresh the review request through GH such that we can properly filter it.

While reviewing it, it just occurred to me that T0 also uses this code. So we need to verify whether T0 would need to use this same mechanism; or whether we need to work on workflow configuration that could enable/disable this new DQM feature. Can you please follow this up with the DQM team?

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
amaltaro: Maybe we could have something like `def _getHttpsOpener` with the initial common code? It it becomes too ugly, then we keep these into 2 different methods.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +311,253 @@ def upload(self, url, args, filename):
...
+                    logging.info(msg)
+                    break
+            except Exception as ex:
+                if numRetry == self.retryCount:
+                    msg = 'HTTP POST to register url failed! Error:\n%s' % str(ex)
+                    if logException:
+                        logging.exception(msg)
+                    else:
+                        logging.error(msg)
+                    raise WMExecutionFailure(70319, "DQMUploadFailure", msg)
```
amaltaro, line 486: Please add this new error code and message to the WMExceptions.py module as well.

src/python/WMCore/WMSpec/Steps/Executors/DQMUpload.py
```diff
@@ -293,3 +311,253 @@ def upload(self, url, args, filename):
...
+                msg = '  Status code: %s\n' % result.getcode()
+                if result.getcode() >= 400 or result.getcode() is None:
+                    logException = False
+                    raise Exception(msg)
+                else:
+                    msg = 'HTTP POST to register url finished succesfully with response:\n' + msg
+                    logging.info(msg)
+                    break
+            except Exception as ex:
+                if numRetry == self.retryCount:
```
amaltaro, line 480: Maybe we should have something like:
```
if numRetry < self.retryCount:
    print "registration failed blah blah"
    sleep
else:
   logger.exception(blah)
```
what do you think?

amaltaro (changes requested):
@khurtado Kenyi, here goes a fresh review of the current changes. In addition to the comments made along the code, I'd like to mention:
1. Please update the initial PR description.
2. I think it would be extremely useful to have a short summary of what DQMUpload module does at the top of the module, including all the steps it takes (since it no longer does a simple upload of the root file).
3. We also need to clarify whether T0 workflows will have to follow the same new design (have you heard anything from the DQM team?)
4. Did we agree on a final map of urls? If so, it needs to be implemented as well.

khurtado:
Hi @amaltaro , to answer your questions:

1. Do you mean the first commit, or this PR GH description?
2. Agreed, Done!
3. Yes, it's the same machinery for T0, answer 1 in: https://github.com/dmwm/WMCore/issues/10287#issuecomment-1086730224
4. Done

amaltaro:
> 1. Do you mean the first commit, or this PR GH description?

The initial PR description at the very top, here: https://github.com/dmwm/WMCore/pull/11015#issue-1154640531
However, having a second look at it, it looks good to me. Perhaps can you expand on the granularity used for the file registration in the new DQM Gui? Is it one HTTP call for each run? For each root file?

src/python/WMCore/WMExceptions.py
```diff
@@ -184,7 +184,12 @@
                       60409: "Timeout in stage out of log files during log collection (WMAgent).",  # (WMA)
                       60450: "No output files present in the report",  # (WMA)
                       60451: "Output file lacked adler32 checksum (WMAgent).",  # (WMA)
+                      70317: "Indefinite hangout while staging out DQM root file to EOS",  # (WMA)
                       70318: "Failure in DQM upload.",
+                      70319: "HTTP POST to register url failed.",  # (WMA)
```
amaltaro, line 189: I'd suggest to rephrase it to ""HTTP POST to DQM register url failed.""

amaltaro (approved):
Kenyi, other than the minor exception suggestion, it looks good to me.

khurtado:
> Kenyi, other than the minor exception suggestion, it looks good to me.
@amaltaro Thanks! I just made that last change. 
Note we still need to have a successful test with the new changes. It is currently failing while registering the file, but it looks like it's due to cmsweb related issue with the new mapping rather than this code. See my last comment below:

https://github.com/dmwm/WMCore/issues/10287#issuecomment-1119818542

PR #11016, opened 2022-03-01
Swap NotMasterError with NotPrimaryError.

### LINKED ISSUE #11012: MSUnmerged: NotMasterError renamed to NotPrimaryError
**Impact of the bug**
MSUnmerged

**Describe the bug**
This is a future bug. Currently we do not experience it, because we are still using old enough version of MongoDB such that the breaking change has not been introduced yet.

While working on some manual imports of few WMCore modules deployed with `pypi` I've stumbled on the following error: [1]. This is a change in behavior introduced  with `pymongo` version 4.0. The pymongo error `NotMasterError` has been deprecated and the new one to be used should be `NotPrimaryError` as recommended at [2]. This is a non-backward compatible change on pymongo side about which we need to pay attention, once we upgrade to `pymongo` version 4.0 and above. We are currently using  version 3.10.1: [3]. One thing needs to be mentioned though, as long as we are using the RPM based deployment we are  safe, but once we move to `pypi` packaging and deployment we need to have it fixed, because at 
https://github.com/dmwm/WMCore/blob/8a6a17e54470cdae5be75209cd9ac3b94c6dbdbb/requirements.txt#L37 
 we are already using `~=` as a requirement string, this immediately places us at the latest available version (currently 4.0.1) because it has no conflicts with any other module deployed.  
  

[1]
```
In [1]: from WMCore.Micrrom WMCore.MicroService.MSUnmerged.MSUnmergedRSE import MSUnmergedRSE

ImportError                               Traceback (most recent call last)
<ipython-input-1-9daca6c75ccf> in <module>
----> 1 from WMCore.MicroService.MSUnmerged.MSUnmergedRSE import MSUnmergedRSE

~/WMCoreDev.d/WMCoreTest.venv3/lib64/python3.6/site-packages/WMCore/MicroService/MSUnmerged/MSUnmergedRSE.py in <module>
      4 """
      5 
----> 6 from pymongo.errors  import NotMasterError
      7 # from pymongo.results import results as MongoResults
      8 

ImportError: cannot import name 'NotMasterError'
```

[2]
https://pymongo.readthedocs.io/en/stable/changelog.html?highlight=3.10#breaking-changes-in-4-0
```
 * Removed pymongo.errors.NotMasterError. Use [pymongo.errors.NotPrimaryError](https://pymongo.readthedocs.io/en/stable/api/pymongo/errors.html#pymongo.errors.NotPrimaryError) instead.
```

[3]
https://github.com/cms-sw/cmsdist/blob/comp_gcc630/py3-pymongo.spec

**How to reproduce it**
```
In [1]: pymongo.version
Out[1]: '4.0.1'

In [2]: from WMCore.Micrrom WMCore.MicroService.MSUnmerged.MSUnmergedRSE import MSUnmergedRSE 
```

**Expected behavior**
Safe import of modules depending on pymongo.

None

todor-ivanov:
Fixes #11012 

#### Status
Ready

#### Description
The current PR is intended to handle the newly introduced exception `NotPrimaryError` on the `pymongo` client side, which is about to substitute the old one  `NotMasterError`, starting with `pymongo` version 4.01 . We have tested and proved that on the server side the latest version is already supporting both the old and the new clients, so no errors in exception handling are originating from client vs server version mismatch.

NO

#### Related PRs
There is about to be a cmsdist related PR to upgrade the pymongo rpm package:
https://github.com/cms-sw/cmsdist/pull/7654

Yes.


amaltaro:
Todor, don't we have to update any of the exception handling in our baseline source code? I had the impression it required a pymongo bump and a WMCore change as well.

todor-ivanov:
Hi @amaltaro  This
https://github.com/dmwm/WMCore/pull/11016/files#diff-cd601ffcba537b986d36fc945d0d34012052b4c198c3fbd14335e01579849b85R141

is the only strictly WMCore related change, which is related to a change of  a deprecated method in pymongo 4.0.1. But this one is there since a while (with a deprecation warning) . The new method (`client.list_database_names()`) exists even in pymongo 3.10.1, see [1]. So this change is not going to break anything in other services using `pymongo`. The rest of the code changes affect only MSUnmerged micro service.


[1]
```
In [2]: msUnmergedDB.command({'buildInfo':1})['version']
Out[2]: '5.0.3'

In [3]: import pymongo

In [4]: pymongo.version
Out[4]: '3.10.1'

In [5]: msUnmergedDB.client.database_names()
~/WMCore.venv3/bin/ipython:1: DeprecationWarning: database_names is deprecated. Use list_database_names instead.
  #!~/WMCore.venv3/bin/python3
Out[5]: ['admin', 'config', 'local', 'msUnmergedDBPreProd', 'msUnmergedDBProd']

In [6]: msUnmergedDB.client.list_database_names()
Out[6]: ['admin', 'config', 'local', 'msUnmergedDBPreProd', 'msUnmergedDBProd']

```

amaltaro (commented):
Todor, it looks like I forgot to scroll down and missed those baseline changes.
While these changes look good to me, I would say it cannot be merged until we cover all the 3 MongoDB use cases that we have in WMCore (as mentioned a minute ago in the GH issue itself).

vkuznet (approved):

todor-ivanov:
Thank you @vkuznet !

BTW I think I have somehow closed that issue by mistake. Reopening it with the current comment.

amaltaro (changes requested):
Just making sure my review - and discussion - gets posted to the PR instead of the issue itself: https://github.com/dmwm/WMCore/issues/11012#issuecomment-1055751018

I think it would be important to have a fully functional change. And given that it was mentioned it would be a near future issue, we might want to decide to come back to this in the very near future as well (hopefully with the VM setup no longer valid).

amaltaro:
Okay, I tried to upgrade MongoDB in COMP, but it's not that straight forward.

Let's get this merged then, but before that, we should check why those unit tests are reported as deleted (I think we need to first upgrade pymongo version, I will get that cmsdist PR merged and then we can come back to this later today or tomorrow).

amaltaro:
test this please

amaltaro:
This looks better now.

Note that this will likely break our MSOutput and MSUnmerged services when running on RPM-based MongoDB setup. @todor-ivanov can you please update the initial description with this note as well? Thanks

amaltaro (approved):

PR #11023, opened 2022-03-04
Allow MongoDB to be instantiated without mongomock library

### LINKED ISSUE #11022: MSOutput/MSUnmerged fails to get an instance of the MongoDB object
**Impact of the bug**
MSOutput and MSUnmerged

**Describe the bug**
With the changes and the new `mongomock` library provided in this PR: https://github.com/dmwm/WMCore/pull/11007

microservices that depend on the MongoDB class will fail to run because the MicroServices spec does not have this new dependency, as can be seen in the reqmgr2ms.spec: https://github.com/cms-sw/cmsdist/blob/comp_gcc630/reqmgr2ms.spec

making the service to fail to start the REST manager and serve RESTful requests, see traceback in [1]

**How to reproduce it**
Deploy reqmgr2ms 2.0.1.pre3 or 2.0.1.pre4 and start reqmgr2ms services.

**Expected behavior**
The `mongomock` library should not be requested to run microservices (be it production service or not). It's only necessary for the unit tests. So, my suggestion is to enclose that import in a try/except clause and simply bypass it when the library is not found.

[1]
```
ERROR initializing MicroService REST module.
  File "/data/srv/HG2203a-939cb157ff04435ac3077ba2f0a76194/sw.amaltaro/slc7_amd64_gcc630/cms/reqmgr2ms/1.0.1.pre3-082edd71c26e8df9fb2c59ca16cef404/lib/python3.8/site-packages/WMCore/MicroService/Service/Data.py", line 58, in __init__
    self.mgr = getattr(module, cname)(config)
  File "/data/srv/HG2203a-939cb157ff04435ac3077ba2f0a76194/sw.amaltaro/slc7_amd64_gcc630/cms/reqmgr2ms/1.0.1.pre3-082edd71c26e8df9fb2c59ca16cef404/lib/python3.8/site-packages/WMCore/MicroService/MSManager.py", line 107, in __init__
    from WMCore.MicroService.MSOutput.MSOutput import MSOutput
  File "/data/srv/HG2203a-939cb157ff04435ac3077ba2f0a76194/sw.amaltaro/slc7_amd64_gcc630/cms/reqmgr2ms/1.0.1.pre3-082edd71c26e8df9fb2c59ca16cef404/lib/python3.8/site-packages/WMCore/MicroService/MSOutput/MSOutput.py", line 25, in <module>
    from WMCore.Database.MongoDB import MongoDB
  File "/data/srv/HG2203a-939cb157ff04435ac3077ba2f0a76194/sw.amaltaro/slc7_amd64_gcc630/cms/reqmgr2ms/1.0.1.pre3-082edd71c26e8df9fb2c59ca16cef404/lib/python3.8/site-packages/WMCore/Database/MongoDB.py", line 12, in <module>
    import mongomock
ModuleNotFoundError: No module named 'mongomock'
```

amaltaro:
Fixes #11022 

#### Status
ready

#### Description
Don't crash the service that is importing MongoDB if it does not have the `mongomock` library in the path, which is actually only needed for unit tests.

YES

#### Related PRs
Complement for: https://github.com/dmwm/WMCore/pull/11007

None

### DIFF
--- a/src/python/WMCore/Database/MongoDB.py
+++ b/src/python/WMCore/Database/MongoDB.py
@@ -5,28 +5,24 @@ Description: Provides a wrapper class for MongoDB
 
 # futures
 from __future__ import division, print_function
-
 from builtins import str, object
 
+try:
+    import mongomock
+except ImportError:
+    # this library should only be required by unit tests
+    mongomock = None
+
 from pymongo import MongoClient, errors, IndexModel
-import mongomock
 
 
 class MongoDB(object):
     """
     A simple wrapper class for creating a connection to a MongoDB instance
     """
-    def __init__(self,
-                 database=None,
-                 server=None,
-                 port=None,
-                 replicaset=None,
-                 create=False,
-                 collections=None,
-                 testIndexes=False,
-                 logger=None,
-                 mockMongoDB=False,
-                 **kwargs):
+    def __init__(self, database=None, server=None, port=None, replicaset=None,
+                 create=False, collections=None, testIndexes=False,
+                 logger=None, mockMongoDB=False, **kwargs):
         """
         :databases:   A database Name to connect to
         :server:      The server url (see https://docs.mongodb.com/manual/reference/connection-string/)
@@ -44,6 +40,10 @@ class MongoDB(object):
         self.port = port # 8230
         self.logger = logger
         self.mockMongoDB = mockMongoDB
+        if mockMongoDB and mongomock is None:
+            msg = "You are trying to mock MongoDB, but you do not have mongomock in the python path."
+            self.logger.critical(msg)
+            raise ImportError(msg)
         try:
             if mockMongoDB:
                 self.client = mongomock.MongoClient()


src/python/WMCore/Database/MongoDB.py
```diff
@@ -5,11 +5,15 @@
...
 # futures
 from __future__ import division, print_function
-
 from builtins import str, object
 
+try:
+    import mongomock
+except ImportError:
+    # this library should only be required by unit tests
+    mongomock = None
```
todor-ivanov, line 14: That is a good workaround, but if we go that way we should then walk to the end. 

We do know that it is supposed to be used only in the unit tests. But, since the class itself provides the option for an object to be created with `mockMongoDB=True`, then  if someone, who is missing the `mongomock` library, does try to benefit from the flag, this would generate the following error on line [53](https://github.com/dmwm/WMCore/blob/4d64d4aab684a7a18b852fba5ab52dba7479e2d3/src/python/WMCore/Database/MongoDB.py#L53) :

```
In [1]: mongomock = None

In [2]: client = mongomock.MongoClient()
AttributeError                            Traceback (most recent call last)
<ipython-input-2-233abe3373bf> in <module>
----> 1 client = mongomock.MongoClient()

AttributeError: 'NoneType' object has no attribute 'MongoClient'
```

In this case I would suggest to add some extra protection on line in question, like:
```
In [7]: if mockMongoDB and mongomock is not None:
   ...:     self.client = mongomock.MongoClient()

```

todor-ivanov (changes requested):
Thanks @amaltaro  for your quick fix. I have left one comment inline, which I think is worth taking a look.

src/python/WMCore/Database/MongoDB.py
amaltaro: I actually wanted it to crash indeed, since it would be a setup that people should never even try. Nothing as a big error in our face :)
But I think I can do something slightly better here, still along the same line of thought. New change coming up in a few minutes.

src/python/WMCore/Database/MongoDB.py
```diff
@@ -48,6 +40,10 @@ def __init__(self,
         self.port = port # 8230
         self.logger = logger
         self.mockMongoDB = mockMongoDB
+        if mockMongoDB and mongomock is None:
+            msg = "You are trying to mock MongoDB, but you do not have mongomock in the python path."
+            self.logger.critical(msg)
+            raise ImportError(msg)
```
todor-ivanov, line 46: ok, this is sound.
Agreed!

todor-ivanov (approved):
Thanks @amaltaro this looks perfect now.

amaltaro:
Thanks Todor. Code rebased and will get merged soon.

PR #11024, opened 2022-03-07
MSOutput: read RelVal output data policy from configuration 

### LINKED ISSUE #10106: Problems in RelVal Output Data Placement
**Impact of the bug**
RelVal Output Data Placement

**Describe the bug**
MSOutput handles the output data placement of RelVal output datasets with the same logic that Unified applied before [1]. There are more than one problem with this logic:
1. Output datasets whose data tier is `ALCARECO` are skipped and they are not placed anywhere.
2. The conditionals are redundant - i.e. it selects `T2_CH_CERN` as a destination if the data tier is not `RECO` and `ALCARECO`, then it checks if the data tier is `GEN-SIM`, which is totally redundant. `GEN-SIM` is already placed to `T2_CH_CERN` in the first conditional.
3. As this is a very old logic, I do not see any justification behind this. For instance, why do the datasets go to `T2_CH_CERN` if they are not `RECO` and `ALCARECO`?

[1] https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Unified/MSOutput.py#L638-L659

**How to reproduce it**
An example from the log:
```
2020-11-21 00:44:02,517:WARNING:MSOutput: RelVal dataset: /RelValZpTT_1500_14/CMSSW_11_2_0_pre9-TkAlUpsilonMuMu-112X_mcRun3_2021_realistic_v11-v1/ALCARECO without any destination
```

**Expected behavior**
A new policy should be prepared for RelVal output data placement. Firstly, `ALCARECO` datasets are not placed anywhere for months and therefore they are not used by anyone for months. This makes me think that maybe they are not necessary and we should stop producing it?

@amaltaro @todor-ivanov @nsmith- FYI

amaltaro:
Fixes #10106 

#### Status
ready

#### Description
This PR provides the following:
* remove the hard-coded RelVal output logic in place of a configurable schema (defined within the MSOutput service configuration)
* validate the policy data structure against is data type, RSE names, and datatier names
* each RSE destination will get a copy of the output data

In addition to this, it also removes an obsolete use of the `tapePledges` configuration attribute.

YES

#### Related PRs
None

Deployment changes: https://github.com/dmwm/deployment/pull/1130
services_config: https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/131 and https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/130

### DIFF
--- a/src/python/WMCore/MicroService/MSOutput/RelValPolicy.py
+++ b/src/python/WMCore/MicroService/MSOutput/RelValPolicy.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Module to store the RelVal output data placement policy and
+make decisions based on that
+"""
+from __future__ import print_function, division
+
+from copy import deepcopy
+import json
+from WMCore.MicroService.Tools.Common import getMSLogger
+from WMCore.WMException import WMException
+
+
+class RelValPolicyException(WMException):
+    """
+    General exception to be raised when a flaw is found in the RelVal
+    output data placement policy
+    """
+    pass
+
+
+class RelValPolicy():
+    """
+    This module will contain the RelVal output data placement policy, where
+    destinations will be decided according to the dataset datatier.
+
+    It's supposed to hold a policy driven by dataset datatier, and it's
+    data structure looks like:
+    [{"datatier": "tier_1", "destinations": ["rse_name_1", "rse_name_2"]},
+     {"datatier": "tier_2", "destinations": ["rse_name_2"]},
+     {"datatier": "default", "destinations": ["rse_name_3"]}]
+
+    the 'default' key matches the case where a datatier is not specified
+    in the policy.
+    """
+
+    def __init__(self, policyDesc, listDatatiers, listRSEs, logger=None):
+        """
+        Given a policy data structure - as a list of dictionaries - it
+        will validate the policy, the datatiers and RSEs defined in it,
+        and it will convert the policy into a flat dictionary for easier
+        data lookup.
+        :param policyDesc: list of dictionary items with the output rules
+        :param listDatatiers: flat list of existent datatiers in DBS
+        :param listRSEs: flat list of existent Disk RSEs in Rucio
+        :param logger: logger object, if any
+        """
+        self.origPolicy = deepcopy(policyDesc)
+
+        self.logger = getMSLogger(verbose=False, logger=logger)
+
+        self._validatePolicy(policyDesc, listDatatiers, listRSEs)
+        self.dictPolicy = self._convertPolicy(policyDesc)
+
+    def __str__(self):
+        """
+        Stringify this object, printing the original policy
+        """
+        objectOut = dict(originalPolicy=self.origPolicy, mappedPolicy=self.dictPolicy)
+        return json.dumps(objectOut)
+
+    def _validatePolicy(self, policyDesc, validDBSTiers, validDiskRSEs):
+        """
+        This method validates the overall policy data structure, including:
+         * internal and external data types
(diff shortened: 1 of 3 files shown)


amaltaro:
@haozturk @bbilin could you please review the RelVal output data placement policy defined in this PR: https://github.com/dmwm/deployment/pull/1130

I am planning to change the implementation of this PR, but I would appreciate if you could confirm that the RelVal policy is sound. Thanks

bbilin:
@amaltaro sorry for the latency in replying.

For the moment we are running all the relval chains at CERN, so there is no need to add FNAL to the rules. We would hence suggest for all the tiers you mention to be stored only at CERN.

Many thanks,

B. for PdmV

src/python/WMCore/MicroService/MSOutput/RelValPolicy.py
```diff
@@ -0,0 +1,106 @@
...
+        :param policyDesc: list of dictionaries with the policy definition
+        :param validDBSTiers: list with existent DBS datatiers
+        :param validDiskRSEs: list with existent Rucio Disk RSEs
+        :return: raise an exception if any validation fails
+        """
+        if not isinstance(policyDesc, list):
+            msg = "The RelVal output data placement policy is not in the expected data type. "
+            msg += "Type expected: list, while the current data type is: {}. ".format(type(policyDesc))
+            msg += "This critical ERROR must be fixed."
+            raise RelValPolicyException(msg) from None
```
todor-ivanov, line 58: Once this is raised during the parameter validation process here, is there a place in the `MSOutput` module code to handle this exception, or it is left to be treated as a general exception and just the proper messages to be logged?

src/python/WMCore/MicroService/MSOutput/RelValPolicy.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # policy must have a default/fallback destination for datatiers not explicitly listed
+        hasDefault = False
+        for item in policyDesc:
+            # validate the datatier
+            if not isinstance(item['datatier'], str):
+                msg = "The 'datatier' parameter must be a string, not {}.".format(type(item['datatier']))
+                raise RelValPolicyException(msg) from None
```
todor-ivanov, line 66: Same question as above. And also for the few similar cases bellow.

todor-ivanov (changes requested):
Thanks for this PR @amaltaro. I do like the idea  for separating the whole RelVal policy generation as a separate class. I've left few comments inline, which may be worth taking a look. None of them is a showstopper. Mostly requests for adding few lines for code documentation.

src/python/WMCore/MicroService/MSOutput/RelValPolicy.py
amaltaro: In case it fails, it's supposed to fail the MSOutput object initialization. Meaning, the service will crash until someone spots the problem.

amaltaro:
Thanks for the review, Todor. I think I either answered or fixed the comments/requests you made.

If it looks good to you, could you please buy us some time and:
* merge this PR
* cut a new WMCore tag (I guess it's 2.0.1.pre5)
* make a new cmsdist pull request with that tag
* request that cmsdist PR - and all of the deployment/services_config - in the usual HG2203 requests gitlab ticket
?

src/python/WMCore/MicroService/MSOutput/RelValPolicy.py
todor-ivanov: That's fair. 
Thanks

todor-ivanov (approved):
This is looking pretty good.
Thanks @amaltaro

haozturk:
perfect, thanks Alan!

PR #11031, opened 2022-03-10
Fix MSTransferor typo calling getName method

amaltaro:
Fixes #11030 

#### Status
ready

#### Description
The title says it all, correct the method name called for the `Workflow` object, which should be `getName()`.

#### Is it backward compatible (if not, which system it affects?)
YES

#### Related PRs
Complement to https://github.com/dmwm/WMCore/pull/10959

#### External dependencies / deployment changes
None

PR #11032, opened 2022-03-10
MSTransferor: fix positional and kwargs passed to sendAlert API

amaltaro:
Fixes #11031
Actually, it fixes the error reported in this comment: https://github.com/dmwm/WMCore/pull/11031#issuecomment-1064014828

#### Status
ready

#### Description
The current code was actually assigning `endSecs` to the `tag` kwarg in the method definition:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/AlertManager/AlertManagerAPI.py#L32

with this fix, we pass `endSecs` as a keyword argument such that parameters are properly passed through.

YES

#### Related PRs
None

None


todor-ivanov (approved):
Looking good!
Thanks  @amaltaro

PR #11033, opened 2022-03-10
Prevent excessive calls for agent config from ReqMgrAux database

### LINKED ISSUE #10860: Excessive number of calls to reqmgr_auxiliary couch db
**Impact of the bug**
WMAgent/CouchDB

**Describe the bug**
While checking central couchdb logs to investigate some alarms that have been popping up over the last couple of hours (vocms0742), I noticed many many calls to the agent configuration in `reqmgr_auxiliary`, sometimes multiple calls within the same second.

**How to reproduce it**
No clue

**Expected behavior**
We need to first investigate who is making all these calls. I believe it's WMAgent (we have 3 or 4 components that rely on the agent configuration), however it should be one call per cycle, so one call every few minutes (and some places have a cache of many minutes).

We need to make sure that the agent configuration is fetched only once per component cycle. If needed, some caching in the ReqMgrAux wrapper should be created.

Couchdb logs in vocms0742 should list something like:
```
... GET /reqmgr_auxiliary/WMAGENT_CONFIG_vocms0243.cern.ch 200
```

goughes:
Fixes #10860 

#### Status
Tested

#### Description
Updates the `_getResult` method of the ReqMgrAux API to not clear its disk cache by default. Also, sets a 5 minute cache duration for agent polling threads to prevent multiple queries.

Reduces queries for the agent configuration by ~30% for an empty agent over a 10 minute period. Unknown how this translates to an agent filled with jobs. 

Also moves calls to `isDrainMode` from inside a for loop in JobArchiverPoller. This is likely the largest cause of excessive calls from agents full of work.

YES

#### Related PRs
NA

NA

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -91,6 +91,8 @@ config.General.logdb_name = logDBName
 config.General.central_logdb_url = "need to get from secrets file"
 config.General.ReqMgr2ServiceURL = "ReqMgr2 rest service"
 config.General.centralWMStatsURL = "Central WMStats URL"
+# ReqMgrAux disk cache duration (in hours), set to 5 minutes: 5 / 60 = 0.083
+config.General.ReqMgrAuxCacheDuration = 0.083
 
 config.section_("JobStateMachine")
 config.JobStateMachine.couchurl = couchURL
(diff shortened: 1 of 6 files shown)


amaltaro (changes requested):
Erik, these changes look good to me. However, I would suggest a small variation of this as follows:
* I think we can default this cache to 5min, for all those components
* create a new `General` WMAgent configuration attribute which will contain the cache duration, e.g. something like `config.General.ReqMgrAuxCacheDuration = 0.083 (please add a comment explaining this valid and unit)
*  read this value in from the components with a default value, e.g. `var = getattr(config.General, "ReqMgrAuxCacheDuration", 5 / 60)`, such that it doesn't fail unit tests.

Here is the place to change this configuration: https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py#L93

amaltaro (approved):
This looks good to me, thanks Erik! Could you please squash these commits? Thanks

goughes:
Got these squashed, thanks @amaltaro!

amaltaro:
Thanks Erik!

PR #11040, opened 2022-03-16
Set processing version as integer

### LINKED ISSUE #10844: Ensure correct data-type values in DBS injection requests
**Impact of the new feature**

None

**Additional context**

vkuznet:
Fixes #10844 

#### Status
not-tested

#### Description
In DBS processing version is defined as integer. During WMA integration tests with new DBS Go server I identified the following error:
```
Go struct field ProcessingEra.processing_era.processing_version of type int64
```
which means that processing version was not passed as integer. Indeed, I found that in bulkblocks API call it passed as following:
```
"processing_era": {"processing_version": "11", "create_by": "WMAgent", "description": ""}
```
I looked up all possible `def setProcessingVer` functions and adjusted their code to ensure int data type. Unless if it is set by other method I think these changes should ensure that it is converted to integer data-type. Please note, I did not put any checks on provided values and therefore if `int(...)` function will fail it will cause the exception. We will need to run integration tests again to see if we either fixed the issue or will get an exception.

MAYBE

#### Related PRs

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
@@ -232,10 +232,17 @@ class DBSBufferBlock(object):
         Set the block's processing version.
         """
         # compatibility statement for old style proc ver (still needed ?)
-        if procVer.count("-") == 1:
-            self.data["processing_era"]["processing_version"] = procVer.split("-v")[1]
+        pver = procVer
+        if isinstance(pver, str) or isinstance(pver, bytes):
+            procVer = int(pver)
+        elif isinstance(pver, int):
+            procVer = pver
+        elif pver == None:
+            procVer = 0
         else:
-            self.data["processing_era"]["processing_version"] = procVer
+            msg = "Provided procVer=%s of type %s cannot be converted to int" % (procVer, type(procVer))
+            raise Exception(msg)
+        self.data["processing_era"]["processing_version"] = procVer
 
         self.data["processing_era"]["create_by"] = "WMAgent"
         self.data["processing_era"]["description"] = ""
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
@@ -387,8 +387,16 @@ class DBSBufferFile(WMBSBase, WMFile):
 
         Set the era
         """
-
-        self['processingVer'] = ver
+        if isinstance(ver, str) or isinstance(ver, bytes):
+            procVer = int(ver)
+        elif isinstance(ver, int):
+            procVer = ver
+        elif ver == None:
+            procVer = 0
+        else:
+            msg = "Provided procVer=%s of type %s cannot be converted to int" % (ver, type(ver))
+            raise Exception(msg)
+        self['processingVer'] = procVer
         return
 
     def setAcquisitionEra(self, era):
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -949,6 +949,9 @@ class WMWorkloadHelper(PersistencyHelper):
 
         Change the processing version for all tasks in the spec and then update
         all of the output LFNs and datasets to use the new processing version.
+
+        :param processingVersions: can be any data-type but it is set from StdBase
+        which performs the input data sanitization/type already.
         """
         stepNameMapping = self.getStepMapping()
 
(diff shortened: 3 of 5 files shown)


src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -232,10 +232,19 @@ def setProcessingVer(self, procVer):
         Set the block's processing version.
         """
         # compatibility statement for old style proc ver (still needed ?)
-        if procVer.count("-") == 1:
-            self.data["processing_era"]["processing_version"] = procVer.split("-v")[1]
+        pver = procVer
```
amaltaro, line 235: Valentin, I would suggest to drop this compatibility statement. It's been many years and we definitely do not have any of those ancient workflows still active in the system.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -232,10 +232,19 @@ def setProcessingVer(self, procVer):
...
+        if isinstance(pver, str) or isinstance(pver, bytes):
+            if procVer.count("-") == 1:
+                pver = procVer.split("-v")[1]
+            procVer = int(pver)
+        elif isinstance(pver, int):
+            procVer = pver
+        elif pver == None:
```
amaltaro, line 242: Here you change the behavior. I'd suggest not to default to anything other than what comes from the database.

amaltaro: Actually, I suggest to simply assign the value that it receives from the database (thus keeping basically the same logic, without the ancient compatibility code). I'm going to suggest a schema change - later in this review - that will ensure the correct data type.

src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
```diff
@@ -387,8 +387,17 @@ def setProcessingVer(self, ver):
 
         Set the era
         """
-
-        self['processingVer'] = ver
+        procVer = 0 # type declaration
```
amaltaro, line 390: this code change is likely not necessary (with the database schema change).

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -956,7 +956,18 @@ def setProcessingVersion(self, processingVersions):
             task.setProcessingVersion(processingVersions, stepChainMap=stepNameMapping)
 
         self.updateLFNsAndDatasets()
-        self.data.properties.processingVersion = processingVersions
+        procVer = 0 # type declaration
+        if isinstance(processingVersions, str) or isinstance(processingVersions, bytes):
```
amaltaro, line 960: ProcessingVersion is already validated at the WMSpec/StdBase module, there is no need for all these changes either.

amaltaro (changes requested):
Valentin, most of these changes are actually not necessary.

As we discussed yesterday, the best way to address it would be with a database schema change, here:
https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/DBS3Buffer/MySQL/Create.py#L31

and in its equivalent Oracle Create module (replace MySQL by Oracle in that path and you will find the schema creation for Oracle).

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
vkuznet: this I agree may go, I only preserved the original logic, will drop

src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
vkuznet: I explicitly added type declaration not only for clarity but also to declare default value. It is bad practice to define new variable within if statements and my intention was to have proper definition. Please let me know explicitly if you disagree and I'll remove it then.

amaltaro: I am in favor of validating the data type as it "enters" the system. Given the python duck typing nature, we would have to implement such code ALL over WMCore if we want to ensure data type, which in many cases will only unnecessarily increase the complexity of the system.

In this case, we already have two places enforcing the data type:
* at the spec level (WMSpec/StdBase)
* and at the database schema one

so I'd rather trust that those are enough for this check.

src/python/WMCore/WMSpec/WMWorkload.py
amaltaro: Yes, it could be anything **in** this method. But this setter is called from StdBase, which performs the input data sanitization/type already.
This is exactly what I mentioned a minute ago, which I would like to avoid checking data type every time that data flows between functions/methods. We should put a check at the very top level, and then all the bottom ones would be correct.

Of course, it does not give us protection for someone, one day, making a mistake and overwriting it along the code to a different data type... (or maybe it does at db level?)

src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
vkuznet: ok, done

src/python/WMCore/WMSpec/WMWorkload.py
vkuznet: ok, done, and I added appropriate docstring about StdBae.

vkuznet:
Alan, all tests passed, I adjusted code to your comments, please review it again.

vkuznet:
this seems to be duplicate of #11043

amaltaro:
Let me close this one out to avoid confusions then. Please reopen it if I closed the wrong one.

PR #11043, opened 2022-03-17
Adjust processing version to be integer

### LINKED ISSUE #10844: Ensure correct data-type values in DBS injection requests

vkuznet:
Fixes #10844 

#### Status
not tested

#### Description
Adjust processing version to be integer data type similar to DBS.

MAYBE

#### Related PRs
#11040 

### DIFF
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
@@ -231,12 +231,14 @@ class DBSBufferBlock(object):
 
         Set the block's processing version.
         """
-        # compatibility statement for old style proc ver (still needed ?)
-        if procVer.count("-") == 1:
-            self.data["processing_era"]["processing_version"] = procVer.split("-v")[1]
-        else:
-            self.data["processing_era"]["processing_version"] = procVer
-
+        pver = procVer or 0
+        try:
+            pver = int(pver)
+        except TypeError:
+            msg = "Provided procVer=%s of type %s cannot be converted to int" \
+                    % (procVer, type(procVer))
+            raise TypeError(msg) from None
+        self.data["processing_era"]["processing_version"] = pver
         self.data["processing_era"]["create_by"] = "WMAgent"
         self.data["processing_era"]["description"] = ""
         return
--- a/src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
@@ -387,8 +387,14 @@ class DBSBufferFile(WMBSBase, WMFile):
 
         Set the era
         """
-
-        self['processingVer'] = ver
+        procVer = ver or 0
+        try:
+            procVer = int(procVer)
+        except TypeError:
+            msg = "Provided ver=%s of type %s cannot be converted to int" \
+                    % (ver, type(ver))
+            raise TypeError(msg) from None
+        self['processingVer'] = procVer
         return
 
     def setAcquisitionEra(self, era):
--- a/src/python/WMComponent/DBS3Buffer/MySQL/UpdateDataset.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/UpdateDataset.py
@@ -26,10 +26,16 @@ class UpdateDataset(DBFormatter):
               WHERE id = :id
            """
     def execute(self, datasetId,
-                processingVer = None, acquisitionEra = None,
+                processingVer = 0, acquisitionEra = None,
                 validStatus = None, globalTag = None,
                 parent = None, prep_id = None,
                 conn = None, transaction = False):
+
+        # to be backward compatible with various places in WMCore codebase which
+        # assign processingVer to None default value. In DBS and WMCore databases
+        # the processsing version should be integer data type
+        if processingVer is None:
+            processingVer = 0
 
         bindVars = {"procVer" : processingVer,
                     "acqEra" : acquisitionEra,
--- a/src/python/WMCore/WMSpec/WMWorkload.py
+++ b/src/python/WMCore/WMSpec/WMWorkload.py
@@ -949,6 +949,9 @@ class WMWorkloadHelper(PersistencyHelper):
 
         Change the processing version for all tasks in the spec and then update
         all of the output LFNs and datasets to use the new processing version.
+
+        :param processingVersions: can be any data-type but it is set from StdBase
+        which performs the input data sanitization/type already.
         """
         stepNameMapping = self.getStepMapping()
 
--- a/test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
+++ b/test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
@@ -947,7 +947,7 @@ class DBSBufferFileTest(unittest.TestCase):
                                configContent="MOREGIBBERISH")
         testFileA.setDatasetPath("/Cosmics/CRUZET09-PromptReco-v1/RECO")
         testFileA.setValidStatus(validStatus="VALID")
-        testFileA.setProcessingVer(ver="ProcVer")
+        testFileA.setProcessingVer(ver="123")
         testFileA.setAcquisitionEra(era="AcqEra")
         testFileA.setGlobalTag(globalTag="GlobalTag")
         testFileA.setDatasetParent(datasetParent="Parent")
--- a/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/Create.py
@@ -28,7 +28,7 @@ class Create(DBCreator):
             """CREATE TABLE dbsbuffer_dataset (
                  id              INTEGER      AUTO_INCREMENT,
                  path            VARCHAR(500) COLLATE latin1_general_cs NOT NULL,
-                 processing_ver  VARCHAR(255),
+                 processing_ver  INTEGER      NOT NULL,
                  acquisition_era VARCHAR(255),
                  valid_status    VARCHAR(20),
                  global_tag      VARCHAR(255),
--- a/src/python/WMComponent/DBS3Buffer/MySQL/FindDASToUpload.py
+++ b/src/python/WMComponent/DBS3Buffer/MySQL/FindDASToUpload.py
@@ -44,11 +44,6 @@ class FindDASToUpload(DBFormatter):
         """
         datasetalgos = []
         for result in results:
-
-            # compatibility statement for old style proc ver (still needed ?)
-            if result['procver'].count("-") == 1:
-                result['procver'] = result['procver'].split("-v")[1]
-
             datasetalgos.append( { 'DatasetPath' : result['datasetpath'],
                                    'AcquisitionEra' : result['acquera'],
                                    'ProcessingVer' : result['procver'] } )
--- a/src/python/WMComponent/DBS3Buffer/Oracle/Create.py
+++ b/src/python/WMComponent/DBS3Buffer/Oracle/Create.py
@@ -27,7 +27,7 @@ class Create(DBCreator):
             """CREATE TABLE dbsbuffer_dataset (
                  id                INTEGER        NOT NULL,
                  path              VARCHAR2(500)  NOT NULL,
-                 processing_ver    VARCHAR2(255),
+                 processing_ver    INTEGER        NOT NULL,
                  acquisition_era   VARCHAR2(255),
                  valid_status      VARCHAR2(20),
                  global_tag        VARCHAR2(255),
--- a/test/python/WMComponent_t/JobAccountant_t/JobAccountant_t.py
+++ b/test/python/WMComponent_t/JobAccountant_t/JobAccountant_t.py
@@ -1005,7 +1005,7 @@ class JobAccountantTest(EmulatedUnitTestCase):
         datasetInfo = myThread.dbi.processData("SELECT * FROM dbsbuffer_dataset")[0].fetchall()[0]
         self.assertEqual(datasetInfo[1],
                          "/Mu/IansMagicMushroomSoup-T0Test-AnalyzeThisAndGetAFreePhD-PreScaleThingy10-v9_29_pre14replaythingy_v5/AOD")
-        self.assertEqual(datasetInfo[2], "9")
+        self.assertEqual(datasetInfo[2], 9) # processing version, should be integer
         self.assertEqual(datasetInfo[3], "IansMagicMushroomSoup")
         self.assertEqual(datasetInfo[4], "Production")
         self.assertEqual(datasetInfo[5], "GT:Super")


vkuznet:
@amaltaro , I don't know what's wrong with github, but it does not update ability to merge automatically this PR. the only changes I put is to the schema. The jenkins says `GitHub pull request #11043 of commit 7c588dfc1194ce2a49a9593311559cb1ce0845f7, has merge conflicts.` but I don't see any issue with merge on github. Any clues?

vkuznet:
Alan, if possible, can you please create from your end the PR with this changes and we may then close this one. I don't know why github stuck here and I want to check if schema changes will lead to something in jenkins tests.

amaltaro:
Let's give it some time. I think GH is going through some changes (I saw new features around) and maybe things didn't go as they expected.

vkuznet:
Alan, it looks now things with github get back to normal and now I see that PR can be merged. Could you please trigger the jenkins test.

amaltaro:
can you try trigerring it yourself with a comment in this PR as `test this please` (nothing else!). You should have permissions to do so.

vkuznet:
test this please

vkuznet:
Alan, the good news is that I have no longer see failures in DBS3Upload due to wrong data-type. So, the change to the schema is correct one. The bad news is that I found that in https://github.com/dmwm/WMCore/blob/master/src/python/WMComponent/DBS3Buffer/MySQL/UpdateDataset.py#L29 the `processingVer` is passed with a wrong data-type (`None`) which causes failure in unit tests now. I fixed it and resubmit changes, let's see how unit tests will go now. We may have other places where code passes wrong data-type.

vkuznet:
Alan, I was trying to converge on minimum set of changes due to wrong data-type and turns out that in order to pass all unit tests I was needed changed both to schema and the ones I introduced in https://github.com/dmwm/WMCore/pull/11040. I merged later over here and now all tests ares passed, and I no longer have issue with WMA (even though now I face a different issue with file parentage which is not related to these changes). So I think we good to go, and I would appreciate if you'll make a final review.

As you said, we may need clean setup of WMA and this can be done once we agree on this PR. If you accept this one we may safely close #11040 as it duplicates the changes now.

vkuznet:
With these changes now I see successful data injection into new DBS server:
```
[2022-03-21 08:21:03.455144197 +0000 UTC m=+222376.335625738] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 11460 in 2 out] [remoteAddr: 188.184.72.217:47943] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 412.51213ms proxy-resp: 0]
```
The `bulkblocks` API happily returns 200 HTTP repsonse.

vkuznet:
Finally, after many iterations I got completed WMAgent workflow, see https://cmsweb-testbed.cern.ch/reqmgr2/fetch?rid=wmagent_TC_6Tasks_Scratch_test_dbs2go_writer_220322_140254_8786 All bulkblocks API calls to new DBS Go writer are successful
```
[2022-03-22 14:14:57.16858942 +0000 UTC m=+4404.505114946] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 13955 in 2 out] [remoteAddr: 10.100.147.128:61518] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 4.228858337s proxy-resp: 0]
[2022-03-22 14:46:50.957306578 +0000 UTC m=+6318.293832110] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 8697 in 2 out] [remoteAddr: 188.185.121.219:35863] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 4.429559957s proxy-resp: 0]
[2022-03-22 15:00:17.794534604 +0000 UTC m=+7125.131060135] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 8042 in 2 out] [remoteAddr: 188.185.121.219:21816] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 2.236467875s proxy-resp: 0]
[2022-03-22 15:02:01.162415884 +0000 UTC m=+7228.498941407] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 8439 in 2 out] [remoteAddr: 10.100.147.128:12674] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 945.864915ms proxy-resp: 0]
[2022-03-22 15:02:01.162638386 +0000 UTC m=+7228.499163921] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 8307 in 2 out] [remoteAddr: 10.100.147.128:42517] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 945.571729ms proxy-resp: 0]
[2022-03-22 15:20:28.547742938 +0000 UTC m=+8335.884268449] logging.go:228: HTTP/1.1 200 POST /dbs2go-writer/bulkblocks [data: 13449 in 2 out] [remoteAddr: 10.100.6.64:7820] [X-Forwarded-For: 188.184.88.70] [X-Forwarded-Host: cmsweb-testbed.cern.ch] [auth: no-TLS cipher-none "/DC=ch/DC=cern/OU=computers/CN=wmagent/vocms0192.cern.ch" wmagent@vocms0192.cern.ch X509Cert] [ref: "-" "DBSClient/Unknown/"] [req: 1.567981098s proxy-resp: 0]
```
And, this proves that integration test between WMAgent and new Go based DBSWriter is doing fine.

Alan, at this point I don't see any further need to adjust this PR and will await on your approval.

Meanwhile, I sent you and others another email on my finding with logic of `bulkblocks` API which may cause racing conditions (by design), but this is a separate story from this PR.

test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
```diff
@@ -947,7 +947,7 @@ def testProperties(self):
                                configContent="MOREGIBBERISH")
         testFileA.setDatasetPath("/Cosmics/CRUZET09-PromptReco-v1/RECO")
         testFileA.setValidStatus(validStatus="VALID")
-        testFileA.setProcessingVer(ver="ProcVer")
+        testFileA.setProcessingVer(ver="123")
```
amaltaro, line 950: shouldn't it be integer?

src/python/WMCore/WMSpec/WMWorkload.py
```diff
@@ -949,14 +949,25 @@ def setProcessingVersion(self, processingVersions):
...
+        which performs the input data sanitization/type already.
         """
         stepNameMapping = self.getStepMapping()
 
         for task in self.taskIterator():
             task.setProcessingVersion(processingVersions, stepChainMap=stepNameMapping)
 
         self.updateLFNsAndDatasets()
-        self.data.properties.processingVersion = processingVersions
+        if isinstance(processingVersions, int):
```
amaltaro, line 962: Did you make this change because the argument name suggests it could be multiple processing versions?
I think this is a valid case for TaskChain workflows where diff procVer can be defined for different tasks.

Unless there is a strong reason for this change, I think we should remove these changes.

src/python/WMComponent/DBS3Buffer/MySQL/UpdateDataset.py
```diff
@@ -26,11 +26,17 @@ class UpdateDataset(DBFormatter):
...
-                processingVer = None, acquisitionEra = None,
+                processingVer = 0, acquisitionEra = None,
                 validStatus = None, globalTag = None,
                 parent = None, prep_id = None,
                 conn = None, transaction = False):
 
+        # to be backward compatible with various places in WMCore codebase which
+        # assign processingVer to None default value. In DBS and WMCore databases
+        # the processsing version should be integer data type
+        if processingVer == None:
```
amaltaro, line 37:
```suggestion
        if processingVer is None:
```

src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
```diff
@@ -387,8 +387,16 @@ def setProcessingVer(self, ver):
 
         Set the era
         """
-
-        self['processingVer'] = ver
+        if isinstance(ver, str) or isinstance(ver, bytes):
```
amaltaro, line 390: I think I commented on this code in the other PR.
When it gets to this level, processing version has been extensively tested. Maybe the only thing that we would need to keep here is something like:
```
if ver is None:
    procVer = 0
```

in case this information isn't loaded from the database.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
```diff
@@ -231,12 +231,17 @@ def setProcessingVer(self, procVer):
 
         Set the block's processing version.
         """
-        # compatibility statement for old style proc ver (still needed ?)
-        if procVer.count("-") == 1:
-            self.data["processing_era"]["processing_version"] = procVer.split("-v")[1]
+        pver = procVer
```
amaltaro, line 234: I think I commented on this code in the other PR.
When it gets to this level, processing version has been extensively tested. Maybe the only thing that we would need to keep here is something like:
```
if ver is None:
    procVer = 0
```

in case this information isn't loaded from the database.

amaltaro (changes requested):
@vkuznet Valentin, I left a few comments along the code.

I think you mentioned that some of those changes are needed to pass the unit tests, but it's still not clear to me why we would need them (unless the unit test is bypassing the Spec validation and testing only things from the Task object level and beyond).

Could you please make those changes I suggested/asked in a different commit? Such that we can see how jenkins is going to behave (and if we eventually find out that we need to remove these changes, we could simply remove the last commit). Thanks

test/python/WMComponent_t/DBS3Buffer_t/DBSBufferFile_t.py
vkuznet: I think we need to test code which should handle string input since there is no data-type checking anywhere in a code and moreover we know that different types can be passed to the code.

src/python/WMCore/WMSpec/WMWorkload.py
vkuznet: Alan, the way how this code is structured it requires a check on provided `processingVersions`. The allowed data-types (as far as I can tell) is `int, dict or list`. Therefore any other type should be ignore and proper exception should be thrown. For instance, if someone will pass the string it will be wrong since I don't know full chain of data manipulation through entire WMCore code. I still suggest to keep this change to avoid potential issue somewhere else.

src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
vkuznet: I disagree since python does not provide a way to check input data-type. I don't rely on your word that provided input is and will be extensively tested. By definition someone can simply pass a string to this code just calling this method. And, I don't understand why you are so resistant against checking provided data-type. I discovered exactly this issue that somewhere in WMCore someone passed a wrong data-type and I'm not convinced that changing schema will prevent from doing this again form any other place in a WMCore code. I think **IT IS SAFE** to check explicitly for correct data type and only allow the one which is expected. The code does not add anything new and only provide additional checks.

src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
vkuznet: The same argument here again. It is much safer to check explicitly the provided data type and avoid potential issue with some caller which will not depend on database type definition.

vkuznet:
Alan, I'm not sure I understand your request to provide changes in separate commit. Please look at 6 failed jenkins tests above in this ticket which shows how tests failed. If you are interesting how the code can fail I can simply initialize the class and call `setProcessingVer` with string argument. I don't think that removing type checking make this code more generic, on a contrary it will allow the same problem which we're trying to fix here to exist in a code since by definition of Python different data-types can be passed to `setProcessingVer` as input values. The changes I propose prevent this from happening regardless of database definition, and therefore **guarantee** the correct data-type assignment to processing version.

src/python/WMComponent/DBS3Buffer/DBSBufferFile.py
amaltaro: Yes, I know it's safer. I am just trying to find a good balance between being safe and keeping code as simple as possible.

I don't think it's impossible to build a DBSBufferFile/DBSBufferBlock without relying on the DAO information, even though WMAgents do so based on DAOs. But okay, let's keep these extra checks here then. I would like to suggest a different approach though, which relies less on data type checking (I'd avoid using anything that checks for bytes/str, as it gave us a nightmare over the past years).

Here is a suggestion for this code:
``` 
ver = ver or 0
try:
    ver = int(ver)
except TypeError:
    msg = "Provided procVer=%s of type %s cannot be converted to int" % (ver, type(ver))
    raise TypeError(msg)
```

or if we do not want to "massage" the exception message, we could remove the try/except as well.
What do you think? Could you please update this code accordingly?

amaltaro (changes requested):
Thanks for your comments, Valentin. I made a few more suggestions, and then I think it should be ready to go.

amaltaro (changes requested):
And I think we need one more change. Thanks for the quick turnaround.

amaltaro:
PS.: I am removing the link to those projects because that is meant to be done only for GH issues.

vkuznet:
Alan, tests are ok, but I don't know what triggered pylint warnings (even though they seem unrelated to changes I made). Please review again.

amaltaro:
Because pylint check works now :-D I will review this again soon.

amaltaro:
@vkuznet Valentin, I think there are pylint records that are worth looking into. For example:
```
src/python/WMComponent/DBS3Buffer/DBSBufferBlock.py
    W0707, line 240 in DBSBufferBlock.setProcessingVer: Consider explicitly re-raising using the 'from' keyword
    W0707, line 310 in DBSBufferBlock.setDataset: Consider explicitly re-raising using the 'from' keyword 
```

Python3 makes exceptions and traceback much more verbose. Given that you already have a proper error message implemented, when we raise an exception from within an exception block, we should actually implement it like:
```
raise TypeError(msg) from None
```

such that the exception that caused it isn't raised. You can find [https://stefan.sofa-rockers.org/2020/10/28/raise-from/](here) some very brief and clear explanation of this.

In general, if any of those **bold** pylint records have been injected by these changes, could you please revisit and clean them up?

Regarding the multiple commits, I think it pollutes the file history if we have a commit that adds something, and the next commit removes it. So, I'd be in favor of squashing both commits. For the record, [this](https://github.com/dmwm/WMCore/pull/11043/commits/37cdce4dfb95b5caf81f990c37aecdc8feeccb4c) link should remain alive and point to the WMWorkload changes that you mentioned, even after you squash and we merge it.

vkuznet:
ok, squashed and fixed those pylint changes which reflect the changes (but I don't touch others which pop up from previous code). Don't tell me that you don't picky :) Please review again and hopefully we can merge it.

amaltaro (approved):
It looks good to me, thanks Valentin!

You are right, there is no need to fix pylint issues from previous commits, however, if you decide to do so - even for only a few of them, like the easy and safe ones - please do so in a second commit. Saying this only for future reference (it is also in the contribution guidelines)

amaltaro:
For the record, you can apply this patch to WMAgents (1.5.7.patchXXX) with this command (with cmst1 user):
```
curl https://patch-diff.githubusercontent.com/raw/dmwm/WMCore/pull/11043.patch | patch -p3 -d apps/wmagentpy3/lib/python3.8/site-packages/
```

then restart the components to make sure that the new code is picked up:
```
$manage stop-agent
and
$manage start-agent
```

PR #11050, opened 2022-03-24
Add getProtectedLFNsT0 to WMStatsServer && Call it from MSUnmerged.

### LINKED ISSUE #10873: MSUnmerged: Start querying  T0 WMStatsServer too
**Impact of the new feature**
MSUnmerged

As T0 are on the path of enabling some T1 sites as part of their resource we need to make some adjustments to MSUnmerged in order to have it synchronized  with the new type of workflows, that  are about to be showing up at the sites. More precisely we are expecting to have workflows at T1 sites about which the production WMStatsServer knows nothing about and hence there are no records for them in protectedLFN lists. For that we will have to start fetching also the protectedLFn list from T0 WMStatsServer too. 

Start querying the T0 WMStatsServer in the initial step and build the proper pretectedLFN list as  a merge from the lists provided by both Production and T0 WMStatsServers. 

None

**Additional context**
None

todor-ivanov:
Fixes #10873 

#### Status
Ready

#### Description
With the changes from the current MSUnmerged starts querying the `protectedlfns` from `T0 WMStatsServer` (the actual interface is `t0_reqmon`) and creates a union from the sets `protectedLFNs` (fetched from `Production WMStatsServer`) and `protectedLFNsT0` (fetched from `T0 WMStatsServer`)  

The behavior is configurable through:
```
msConfig['wmstatsUrlT0'] 
msConfig['enableT0WMStats'] 
```

NOTE: 
The `t0_recmon` functionality is yet to be implemented. The development is tracked in this GH issue: https://github.com/dmwm/WMCore/issues/10874 

YES

#### Related PRs

https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/136
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/merge_requests/137

None

### DIFF
--- a/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
+++ b/src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
@@ -18,7 +18,6 @@ from time import time
 import random
 import re
 import os
-import sys
 import errno
 import stat
 try:
@@ -100,6 +99,7 @@ class MSUnmerged(MSCore):
         self.msConfig.setdefault("skipRSEs", [])
         self.msConfig.setdefault("rseExpr", "*")
         self.msConfig.setdefault("enableRealMode", False)
+        self.msConfig.setdefault("enableT0WMStats", False)
         self.msConfig.setdefault("fullRSEToDB", False)
         self.msConfig.setdefault("gfalLogLevel", 'normal')
         self.msConfig.setdefault("dirFilterIncl", [])
@@ -142,6 +142,7 @@ class MSUnmerged(MSCore):
         self.rucioConMon = RucioConMon(self.msConfig['rucioConMon'], logger=self.logger)
 
         self.wmstatsSvc = WMStatsServer(self.msConfig['wmstatsUrl'], logger=self.logger)
+        self.wmstatsSvcT0 = WMStatsServer(self.msConfig['wmstatsUrlT0'], logger=self.logger)
 
         # Building all the Pipelines:
         pName = 'plineUnmerged'
@@ -163,7 +164,7 @@ class MSUnmerged(MSCore):
 
         # Initialization service common data structures:
         self.rseConsStats = {}
-        self.protectedLFNs = []
+        self.protectedLFNs = set()
 
         # The basic /store/unmerged regular expression:
         self.regStoreUnmergedLfn = re.compile("^/store/unmerged/.*$")
@@ -185,12 +186,25 @@ class MSUnmerged(MSCore):
         try:
             self.protectedLFNs = set(self.wmstatsSvc.getProtectedLFNs())
             # self.logger.debug("protectedLFNs: %s", pformat(self.protectedLFNs))
-
             if not self.protectedLFNs:
-                msg = "Could not fetch the protectedLFNs list from WMStatServer. "
+                msg = "Could not fetch the protectedLFNs list from Production WMStatServer. "
                 msg += "Skipping the current run."
                 self.logger.error(msg)
                 return summary
+
+            protectedLFNsT0 = set()
+            if self.msConfig['enableT0WMStats']:
+                msg = "WMStatsServer.getProtectedLFNs for T0 is not yet implemented."
+                raise NotImplementedError(msg)
+                # protectedLFNs = set(self.wmstatsSvcT0.getProtectedLFNs())
+                if not protectedLFNsT0:
+                    msg = "Could not fetch the protectedLFNs list from T0 WMStatServer. "
+                    msg += "Skipping the current run."
+                    self.logger.error(msg)
+                    return summary
+
+            self.protectedLFNs = self.protectedLFNs | protectedLFNsT0
+
         except Exception as ex:
             msg = "Unknown exception while trying to fetch the protectedLFNs list from WMStatServer. Error: {}".format(str(ex))
             self.logger.exception(msg)
--- a/test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
+++ b/test/python/WMCore_t/MicroService_t/MSUnmerged_t/MSUnmerged_t.py
@@ -123,6 +123,8 @@ class MSUnmergedTest(unittest.TestCase):
                          'services': ['unmerged'],
                          'verbose': True,
                          'wmstatsUrl': 'https://cmsweb-testbed.cern.ch/wmstatsserver',
+                         'wmstatsUrlT0': "https://cmsweb-testbed.cern.ch/t0_reqmon",
+                         'enableT0WMStats': False,
                          'dirFilterIncl': [],
                          'dirFilterExcl': [],
                          'emulateGfal2': True,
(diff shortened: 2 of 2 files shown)


vkuznet (approved):
@todor-ivanov changes are fine with me

todor-ivanov:
Thanks @vkuznet

todor-ivanov:
test this please

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -185,12 +187,24 @@ def execute(self):
...
             if not self.protectedLFNs:
-                msg = "Could not fetch the protectedLFNs list from WMStatServer. "
+                msg = "Could not fetch the protectedLFNs list from Production WMStatServer. "
                 msg += "Skipping the current run."
                 self.logger.error(msg)
                 return summary
+
+            self.protectedLFNsT0 = set()
+            if self.msConfig['enableT0WMStats']:
+                self.protectedLFNsT0 |= set(self.wmstatsSvcT0.getProtectedLFNsT0())
```
amaltaro, line 198: I guess you don't want to have an OR here, right? Otherwise you will keep accumulating LFNs as long as the service is running.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -185,12 +187,24 @@ def execute(self):
...
+                # self.logger.debug("protectedLFNsT0: %s", pformat(self.protectedLFNsT0))
+                if not self.protectedLFNsT0:
+                    msg = "Could not fetch the protectedLFNs list from T0 WMStatServer. "
+                    msg += "Skipping the current run."
+                    self.logger.error(msg)
+                    return summary
+
+            self.protectedLFNs |= self.protectedLFNsT0
```
amaltaro, line 206: IMO, this `|=` operator is easy to miss when reading this code. So I would suggest to expand it to:
```
            self.protectedLFNs = self.protectedLFNs | self.protectedLFNsT0
```

It's a matter of taste though, if you prefer to keep the current code, that's fine as well.

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
```diff
@@ -132,3 +132,16 @@ def getProtectedLFNs(self):
         """
         callname = 'protectedlfns'
         return self._getResult(callname, verb="GET")
+
+    def getProtectedLFNsT0(self):
```
amaltaro, line 136: I could be wrong here, but I think this method is not going to be needed. We will use the same `getProtectedLFNs` method, but this object will be pointing to a different endpoint.

amaltaro (changes requested):
Todor, please find a few comments/questions along the code. Thanks

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: One can never jump over line 196 which is zeroing the protectedLFNsT0 set by default. Both operators `|=` and  `=` would work equally well here. I'll change it, no problem.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
todor-ivanov: yes, I agree it is a little bit more readable. I'll change that one too.

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
todor-ivanov: Well, I did foresee such an approach (eventually). But until we resolve issue https://github.com/dmwm/WMCore/issues/10874 we cannot tell for sure. Because it may turn out the only way to collect what we need from T0 to be trough a set of queries to WMBS, which may change the method signature...etc. Until then I need the separate method, so I can clearly signal `NotImplementedError`. For the time being I did not want to touch the original method for doing so. Once we know how to properly fetch the list from T0 we  can easelly merge those two methods into one. But I'd leave that to be resolved in the other GH issue, rather than here.

todor-ivanov:
Thanks @amaltaro for your review. I did address your comments. And in addition to that I found out that such an extra level of protection as resetting the `protectedLFNs`  list on every call to the `execute`method not only having it zeroed in the `__init__` and then relying on the returned value from `WMStatserver` to be `[]`  is a good thing to have also for the Production's `protectedLFNs` list  similarly to the T0's  `protecteionLFNST0` list. Because `MSManager` creates the `MSUnmerged` object just once whle creating the thread, and is calling the `execute` method from the thread itself on every polling  cycle. 
The changes are in my last commit. Please take another look.

amaltaro (changes requested):
Thanks for the prompt changes, Todor. Please see my comments along the code.

In addition to those, please check out the pylint report and clean those records in bold.

src/python/WMCore/MicroService/MSUnmerged/MSUnmerged.py
```diff
@@ -183,14 +185,27 @@ def execute(self):
 
         # fetch the protectedLFNs list
         try:
+            self.protectedLFNs = set()
```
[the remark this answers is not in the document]
todor-ivanov: I am not sure I understand why we involve the garbage collector here, this line is only to assure we start the cycle with a fresh zeroed `self.protectedLFNs` set. Which if we fail to update on the very next line then we just interrupt the  cycle two lines bellow. It will introduce no new objects in memory, because `self.protectedLFNs` is an attribute to the object.  

The way we were running right now was by relying that in case anything bad happens with the call to the central services,  `WMStatserver` would do one of the followin two things - either raise an error or return an empty list. I have put that line there to be safe than sorry, because an eventual bug introduced in WMStatServer, may lead to change of behavior on its side and then we could be in a big trouble here.

todor-ivanov:
@amaltaro please take another look

src/python/WMCore/Services/WMStatsServer/WMStatsServer.py
```diff
@@ -132,3 +132,4 @@ def getProtectedLFNs(self):
         """
         callname = 'protectedlfns'
         return self._getResult(callname, verb="GET")
+
```
amaltaro, line 135: How about removing this, such that this module is not even part of this PR?

amaltaro (approved):
It looks good to me! Thanks Todor.

amaltaro:
Actually, you missed those pylint reports in bold, e.g.:
```
W0707, line 537 in MSUnmerged.consRecordAge: Consider explicitly re-raising using the 'from' keyword
W1201, line 806 in MSUnmerged.getRSEFromMongoDB: Use lazy % formatting in logging functions
W1201, line 817 in MSUnmerged.uploadRSEToMongoDB: Use lazy % formatting in logging functions
W0707, line 823 in MSUnmerged.uploadRSEToMongoDB: Consider explicitly re-raising using the 'from' keyword 
```

Next time you work on MSUnmerged, please resolve those issues.

PR #11056, opened 2022-03-28
Add usage of gzip encoding in DBS3Buffer

### LINKED ISSUE #11055: Adopt data compression in DBSBuffer
**Impact of the new feature**
We should add data (gzip) compression in DBSBuffer since new DBS server support it

The data injection to DBS can contain large JSON payloads, for large dataset/blocks the size of JSON can reach O(100MB). With gzip compression this size can be significantly reduced (by factor of 20 from preliminary benchmarks). Therefore, it will reduce network IO between client and DBS server

Add `Content-Encoding: gzip` HTTP header to DBSBuffer HTTP requests. This is sufficient to solve the problem. Please note, that this will only be supported in new DBS Go server and will not work with Python server.

None

**Additional context**
This request is only applicable to DBS Go server which we should setup first on our back-ends.

vkuznet:
Fixes #11055

#### Status
not-tested

#### Description
Add ability to use gzip encoding when uploading data to DBS server

YES

#### Related PRs

Requires new DBSClient, version 4.0.7 or above.

### DIFF
--- a/etc/WMAgentConfig.py
+++ b/etc/WMAgentConfig.py
@@ -152,6 +152,9 @@ config.DBS3Upload.pollInterval = 100
 config.DBS3Upload.dbsUrl = "OVERWRITE_BY_SECRETS"
 config.DBS3Upload.primaryDatasetType = "mc"
 config.DBS3Upload.dumpBlock = False  # to dump block meta-data into a json file
+# set DbsApi requests to use gzip enconding, thus sending compressed data
+# please change to True when new DBSWriter Go server will be in place
+config.DBS3Upload.gzipEncoding = False
 
 config.section_("DBSInterface")
 config.DBSInterface.DBSUrl = globalDBSUrl
--- a/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
+++ b/src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
@@ -50,17 +50,22 @@ from WMCore.WMException import WMException
 from WMCore.WorkerThreads.BaseWorkerThread import BaseWorkerThread
 
 
-def uploadWorker(workInput, results, dbsUrl):
+def uploadWorker(workInput, results, dbsUrl, gzipEncoding=False):
     """
     _uploadWorker_
 
     Put JSONized blocks in the workInput
     Get confirmation in the output
+
+    :param workInput: work input data
+    :param results: output results dictionary
+    :param dbsUrl: url of DBS server to use
+    :param gzipEncoding: specify if we should use gzipEncoding
     """
 
     # Init DBS Stuff
     logging.debug("Creating dbsAPI with address %s", dbsUrl)
-    dbsApi = DbsApi(url=dbsUrl)
+    dbsApi = DbsApi(url=dbsUrl, useGzip=gzipEncoding)
 
     while True:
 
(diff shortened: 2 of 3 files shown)


amaltaro (changes requested):
Valentin, thanks for the prompt feedback and action on this item.
However, the right way to deal with it is by defining a new component parameter in our default configuration:
https://github.com/dmwm/WMCore/blob/master/etc/WMAgentConfig.py#L147

which is then picked up in this component from the Configuration object.

todor-ivanov (changes requested):
Thanks @vkuznet 
I'd second Alan's suggestion. it would be better to stick to configuration parameters going through the configuration module rather than through environment variables.

vkuznet:
ok, now you have as configuration parameter

todor-ivanov (approved):
Thanks @vkuznet  for the prompt fix!
This looks good to me!

vkuznet:
Alan, the failed unit test `[E1123, line 202 in DBSUploadPoller.__init__: Unexpected keyword argument 'gzip' in constructor call]()` implies that on jenkins you don't have newer version of dbsClient. As I provided in description of this ticket the version 4.0.7 or above is required. Could you please check and post here which version of dbsClient is used in jenkins and if necessary upgrade it to recent version.

amaltaro:
Jenkins has WMAgent `2.0.1`, which is the same what we have in the docker image. Looking into the container, I see version 4.0.7 in there:
```
[dmwm@7de784512c7b WMArchive_t]$ ls /home/dmwm/unittestdeploy/wmagent/current/sw/slc7_amd64_gcc630/cms/py3-dbs3-client/4.0.7/
```

vkuznet:
ok, thanks, I found the issue.

src/python/WMComponent/DBS3Buffer/DBSUploadPoller.py
```diff
@@ -191,7 +196,8 @@ def __init__(self, config):
         self.datasetType = getattr(self.config.DBS3Upload, "datasetType", "PRODUCTION")
         self.primaryDatasetType = getattr(self.config.DBS3Upload, "primaryDatasetType", "mc")
         self.blockCount = 0
-        self.dbsApi = DbsApi(url=self.dbsUrl)
+        self.gzipEncoding = getattr(self.config.DBS3Upload, 'gzipEncoding', False)
+        self.dbsApi = DbsApi(url=self.dbsUrl, useGzip=self.gzipEncoding)
```
amaltaro, line 200: @vkuznet Valentin, it looks like we do not need to use gzip when getting this instance. It's only used for checking whether a block exists in the DBS server. I'd say not using it for this use case will reduce the overall latency for such checks (no need to compress before and after send).

etc/WMAgentConfig.py
```diff
@@ -152,6 +152,7 @@
 config.DBS3Upload.dbsUrl = "OVERWRITE_BY_SECRETS"
 config.DBS3Upload.primaryDatasetType = "mc"
 config.DBS3Upload.dumpBlock = False  # to dump block meta-data into a json file
+config.DBS3Upload.gzipEncoding = False # swith to True when new DBSWriter will be in place
```
amaltaro, line 155: typo. And how about documenting it like "set DbsApi requests to use gzip enconding, thus sending compressed data"?

src/python/WMQuality/Emulators/DBSClient/DBS3API.py
```diff
@@ -21,11 +21,16 @@ class DbsApi(object):
...
     """
-    def __init__(self, url):
+    def __init__(self, url, **kwds):
         """
         _init_
 
         Store the url where this dbs is "located"
+
+        :param url: url of the DBS server
+        :param kwds: optional positional DBSApi parameters which are irrelevant
```
amaltaro, line 31: these are actually keyword arguments.

amaltaro (changes requested):
Valentin, this looks good in general, but I left a couple of suggestions along the code
Could you please update the initial description as well (and state that it requires Go-based DBSWriter before this new feature can be enabled)? Thanks

vkuznet:
Alan, I adjusted code based on your feedback. Tests are fine. It is ready for review again.

amaltaro (approved):
Looks good to me, thanks!

PR #11059, opened 2022-03-29
FIx default value for processing version

### LINKED ISSUE #10844: Ensure correct data-type values in DBS injection requests

vkuznet:
Fixes #10844 

#### Status
ready

#### Description
Change default values of processingVer in various places of WMCore codebase

MAYBE

#### Related PRs
Complement to #11043


vkuznet:
Alan, turns out WMCore has plenty of `processingVer` defaults pointed to `None`. I scan WMCore code for them, based on jenkins test failures and I think I identified the final set, unless the default values of `processingVer` at a different lines. Please include this PR in jenkins tests and we will see how it is going.

amaltaro (approved):
Thanks, Valentin! We are back to 45 permanent failures in the oracle baseline unit tests.

PR #11060, opened 2022-03-29
Add support to EL8 jobs

### LINKED ISSUE #11051: Add support to RHEL8 workflows/jobs
**Impact of the new feature**
ReqMgr2 / WMAgent

Following an email exchange with Shahzad, they would like to start running tests with RHEL8 production workflows as soon as possible.

In short, WMCore systems should support workflow creation and assignment as well as job creation and execution under RHEL8 OS (or variations of it, like EL8, Alma8, CS8, etc).

A **non-exhaustive** list of places that we need to verify and/or update if required, is:
1. submit_py3.sh job wrapper
2. the Scram.py module, which should have a map of ScramArch to OSes
3. SimpleCondorPlugin

UPDATE: as provided by Shahzad below, the list of new archs to be supported is:
```
(el|cc|cs|alma)N_(amd64|ppc64le|aarch64)_gccV
```
(where N>=8 and V>=10).

None

**Additional context**
In order to properly test it, we need to identify:
* which CMSSW release we could use for this test. Apparently CMSSW_12_3_0_preX releases under `cs8_amd64_gcc10`
* which ScramArch are we going to use for such workflows? Is it going to be only `slc8_*`, or should we consider the other variations?
* and whether there are RHEL8 pilot containers that we could use (thus, being able to run such workflows anywhere)

amaltaro:
Fixes #11051 

#### Status
ready

#### Description
Provides a map of ScramArch to OS for the EL8 flavors.

For completeness, this map is consumed in this BossAir base class:
https://github.com/dmwm/WMCore/blob/fa033faa8ef12bc4f218433d8b3f5346a50d4d74/src/python/WMCore/BossAir/Plugins/BasePlugin.py#L131

YES

#### Related PRs
None

None

### DIFF
--- a/test/python/WMCore_t/WMRuntime_t/Tools_t/Scram_t.py
+++ b/test/python/WMCore_t/WMRuntime_t/Tools_t/Scram_t.py
@@ -107,7 +107,10 @@ class Scram_t(unittest.TestCase):
     def testArchMap(self):
         self.assertItemsEqual(OS_TO_ARCH['rhel6'], ['slc5', 'slc6'])
         self.assertItemsEqual(OS_TO_ARCH['rhel7'], ['slc7'])
+        self.assertItemsEqual(OS_TO_ARCH['rhel8'], ['cc8', 'cs8', 'alma8'])
         self.assertItemsEqual(ARCH_TO_OS['slc6'], ['rhel6'])
+        self.assertEqual(len(ARCH_TO_OS), 6)
+        self.assertItemsEqual(ARCH_TO_OS['slc7'], ['rhel7'])
         self.assertItemsEqual(ARCH_TO_OS['slc7'], ['rhel7'])
 
     def testScramArchParsing(self):
--- a/src/python/WMCore/WMRuntime/Tools/Scram.py
+++ b/src/python/WMCore/WMRuntime/Tools/Scram.py
@@ -39,7 +39,10 @@ from Utils.PythonVersion import PY3
 from Utils.Utilities import encodeUnicodeToBytesConditional, decodeBytesToUnicodeConditional
 
 SCRAM_TO_ARCH = {'amd64': 'X86_64', 'aarch64': 'aarch64', 'ppc64le': 'ppc64le'}
-ARCH_TO_OS = {'slc5': ['rhel6'], 'slc6': ['rhel6'], 'slc7': ['rhel7']}
+ARCH_TO_OS = {'slc5': ['rhel6'],
+              'slc6': ['rhel6'],
+              'slc7': ['rhel7'],
+              'cc8': ['rhel8'], 'cs8': ['rhel8'], 'alma8': ['rhel8']}
 OS_TO_ARCH = {}
 for arch, oses in viewitems(ARCH_TO_OS):
     for osName in oses:


test/python/WMCore_t/WMRuntime_t/Tools_t/Scram_t.py
```diff
@@ -107,7 +107,10 @@ def testC(self):
     def testArchMap(self):
         self.assertItemsEqual(OS_TO_ARCH['rhel6'], ['slc5', 'slc6'])
         self.assertItemsEqual(OS_TO_ARCH['rhel7'], ['slc7'])
+        self.assertItemsEqual(OS_TO_ARCH['rhel8'], ['cc8', 'cs8', 'alma8'])
```
vkuznet, line 110: I don't know if this is generic enough, I rather think you should apply sort to final list to get consistent results all the time. But this should be do to all checked lists (this and on line 108 and may be others if they exist in a code). It really depends on a code which make these lists.

todor-ivanov (approved):
looks good to me

test/python/WMCore_t/WMRuntime_t/Tools_t/Scram_t.py
mapellidario: I had your very same worry about the sorting of lists in tests when I was doing the py3 migration.

I think that Alan's changes are as safe as possible. Inside the [doc](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual) it is specified that this test already sorts internally the two lists. Yes, this was renamed in python3. Yes, i think that [assertCountEqual](https://docs.python.org/3.10/library/unittest.html#unittest.TestCase.assertCountEqual) is even a worse name for such an assertion. 

I am sorry for intruding here, Stefano asked me to adapt CRAB code to use what you will develop in this PR and stumbled upon this comment.

amaltaro: Thanks, Dario. Yes, Valentin. In short, we do need to sort these lists to run a comparison.

amaltaro:
I ran some DMWM tests, and edited one json template to use:
```
            "CMSSWVersion": "CMSSW_12_3_0_pre6",
            "ScramArch": "cs8_amd64_gcc10",
``` 
and I see that the correct `rhel8` REQUIRED_OS condor job ad has been defined.

Once things are all set at the SI layer, we will probably have to apply this patch to the RelVal agent such that further EL8 tests can be performed.

amaltaro:
Thanks for the review.

smuzaffar:
@amaltaro , can you please also add `el8` to the list?

amaltaro:
> @amaltaro , can you please also add el8 to the list?

yet another flavor?!? Sure, thanks for bringing this up Shahzad! @belforte FYI

smuzaffar:
@amaltaro , in yesterday O&C meeting I proposed to call the new arch `el8` instead of having different flavors `alma8 or cs8 or rhel8` :-) so in the end we might just need `el8` but for now keep all of these in.

amaltaro:
Yes, I saw a few question marks in the final slide but missed the conclusions, I guess.

Agreed, let's keep it around for a little longer and we get back to this in a couple of months, once everything is solid. Thanks Shahzad.

PR #11061, opened 2022-03-31
MSTransferor: do not bypass Pileup rules in INJECT/REPLICATING state

### LINKED ISSUE #11035: MSTransferor: rule ids for INJECT/REPLICATING rules must be passed to MSMonitor
**Impact of the bug**
MSTransferor

**Describe the bug**
Hasan reported a workflow that was bypassed straight to `staging` in MSTransferor, and the reason for that is that its pileup dataset already had an input data rule in status REPLICATING. In such cases, we simply consider data to be on the way and advance the workflow to the next status (staging), without keeping record of the rule_ids still unsatisfied. Which causes the same workflow(s) to be bypassed by MSMonitor as well. 

If the workflow is marked as `TrustPUSitelists=true`, then it would be bypassed by global workqueue as well, which could result in many jobs failing to read secondary data in case not all files are available at any given Disk endpoint.

**How to reproduce it**
Assign a workflow with pileup - with secondary AAA enabled - where the pileup only has rules in INJECT or REPLICATING status. 

**Expected behavior**
I think the best way to address it would be:
* if input rule is in status INJECT or REPLICATING, we need to keep that/those rule_id and persist it in the transfer document
* in terms of MSTransferor, we do not require any new input rules for such dataset and allow it to move to `staging`, in case all of the other requirements are met.

Here is an example: https://cmsweb.cern.ch/ms-transferor/data/info?request=cmsunified_task_TSG-Run3Summer21wmLHEGS-00035__v1_T_220309_215550_5271

amaltaro:
Fixes #11035 

#### Status
not-tested

#### Description
If a pileup has a Rucio rule in one of these states: INJECT or REPLICATING, then it should not consider as having the data already available and a new rule needs to be created.

In the case it fails with a `DuplicateRule` exception, then `createReplicationRule` API will automatically find the rule_id for the existent rule and it will be returned to MSTransferor, which then persists this rule_id in the transfer document to be monitored by MSMonitor.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/MicroService/Tools/PycurlRucio.py
+++ b/src/python/WMCore/MicroService/Tools/PycurlRucio.py
@@ -164,6 +164,8 @@ def listReplicationRules(containers, rucioAccount, grouping,
     :return: a flat dictionary key'ed by the container name, with a list of RSE
       expressions that still need to be resolved
     NOTE: Value `None` is returned in case the data-service failed to serve a given request.
+    NOTE-2: Available rule states can be found at:
+    https://github.com/rucio/rucio/blob/16f39dffa1608caa0a1af8bbc0fcff2965dccc50/lib/rucio/db/sqla/constants.py#L180
     """
     locationByContainer = {}
     if not containers:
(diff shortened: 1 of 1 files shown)


todor-ivanov (approved):
Thanks @amaltaro 
Looks good.

PR #11067, opened 2022-04-05
Assign the correct activity for Tape_Test RSEs

### LINKED ISSUE #11046: Wrong activity for subscriptions to Tape_Test sites
**Impact of the bug**
T0 Agent and probably production WMAgent

**Describe the bug**
During the transfers test, subscribing data to `T0_CH_CERN_Tape_Test` we spotted that a wrong activity was assigned to multiple rules.. For example the following:
```
rucio rule-info 9fadb5aa83114e9485fd7e8a43c6d1b1
Id:                         9fadb5aa83114e9485fd7e8a43c6d1b1
Account:                    tier0_prod
Scope:                      cms
Name:                       /D16/TransferTest2022-v3/RAW
RSE Expression:             T0_CH_CERN_Tape_Test
Copies:                     1
State:                      OK
Locks OK/REPLICATING/STUCK: 3488/0/0
Grouping:                   ALL
Expires at:                 None
Locked:                     True
Weight:                     None
Created at:                 2022-02-21 14:45:38
Updated at:                 2022-02-23 01:51:27
Error:                      None
Subscription Id:            None
Source replica expression:  None
Activity:                   T0 Export
Comment:                    T0 WMAgent automatic container rule
Ignore Quota:               False
Ignore Availability:        False
Purge replicas:             False
Notification:               NO
End of life:                None
Child Rule Id:              None
```

**How to reproduce it**
Replay subscribing data to any Tape_Test site

**Expected behavior**
Activity to all Tape sites even for the test sites should be `T0 Tape` 

None

jhonatanamado:
Fixes #11046 

#### Status
Tested with a replay

#### Description
Add the proper activity for the T0-Agent when is subscribing data to Tape_Test rses. Small changes on how this code recognize a Tape_Test RSE. With this changes, the rule replica subscribed to T0_CH_CERN_Tape_Test gives the following:

```
rucio rule-info 0035c69b588045769195f241df939406
Id:                         0035c69b588045769195f241df939406
Account:                    tier0_replay
Scope:                      cms
Name:                       /ExpressCosmics/Tier0_REPLAY_2022-v98/RAW
RSE Expression:             T0_CH_CERN_Tape_Test
Copies:                     1
State:                      REPLICATING
Locks OK/REPLICATING/STUCK: 0/1/0
Grouping:                   ALL
Expires at:                 None
Locked:                     True
Weight:                     None
Created at:                 2022-04-05 03:33:03
Updated at:                 2022-04-05 03:33:55
Error:                      None
Subscription Id:            None
Source replica expression:  None
Activity:                   T0 Tape
Comment:                    T0 WMAgent automatic container rule
Ignore Quota:               False
Ignore Availability:        False
Purge replicas:             False
Notification:               NO
End of life:                None
Child Rule Id:              None
```

Maybe

#### Related PRs
None

None

### DIFF
--- a/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
+++ b/src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
@@ -465,7 +465,7 @@ class RucioInjectorPoller(BaseWorkerThread):
                               grouping="ALL",
                               comment=ruleComment,
                               meta=self.metaData)
-            if not rseName.endswith("_Tape"):
+            if not rseName.endswith(("_Tape", "_Tape_Test")):
                 # add extra parameters to the Disk rule as defined in the component configuration
                 ruleKwargs.update(self.containerDiskRuleParams)
 
@@ -529,7 +529,7 @@ class RucioInjectorPoller(BaseWorkerThread):
         """
         if not self.isT0agent and not rseName.endswith("_Tape"):
             return "Production Output"
-        elif self.isT0agent and rseName.endswith("_Tape"):
+        elif self.isT0agent and rseName.endswith(("_Tape", "_Tape_Test")):
             return "T0 Tape"
         elif self.isT0agent:
             return "T0 Export"


src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -465,7 +465,7 @@ def insertContainerRules(self):
                               grouping="ALL",
                               comment=ruleComment,
                               meta=self.metaData)
-            if not rseName.endswith("_Tape"):
+            if "Tape" not in rseName:
```
amaltaro, line 468: This is a risky check, and given that we never produce data at a `Tape` endpoint, I would suggest to revert this change.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
```diff
@@ -529,7 +529,7 @@ def _activityMap(self, rseName):
         """
         if not self.isT0agent and not rseName.endswith("_Tape"):
             return "Production Output"
-        elif self.isT0agent and rseName.endswith("_Tape"):
+        elif self.isT0agent and "Tape" in rseName:
```
amaltaro, line 532: I would suggest a check like:
``` 
elif self.isT0agent and (rseName.endswith("_Tape") or rseName.endswith("_Tape_Test"):
``` 

People are very creative and I wouldn't be surprised to, one day, see a Disk endpoint with `Tape` in its name :-D

amaltaro (changes requested):
Thanks for providing this patch, Jhonatan. Please see a couple of comments along the code.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
todor-ivanov: I can second [Alan's](@amaltaro) suggestion here.

src/python/WMComponent/RucioInjector/RucioInjectorPoller.py
todor-ivanov: Agreed, the way how this check has been implemented originally is better.

todor-ivanov (changes requested):
Thanks [Jhonatan](@jhonatanamado)
I have basically agreed with Alan's suggestions here too. Please take a look the comments inline.

todor-ivanov (approved):
Thanks @jhonatanamado 
looks good to me

amaltaro (approved):
@jhonatanamado this is even a better implementation than the one suggested. I wasn't aware we could check it against a tuple! Thanks

PR #11069, opened 2022-04-06
Use Content-Encoding to automatically decode body in RequestHandler

### LINKED ISSUE #11066: Wrong gzip compression recognized/identified by header
**Impact of the bug**
Reqmgr2, WorkQueue , maybe some others as well

**Describe the bug**
While validating the April release we have noticed an error coming from the newly introduced `decompress` function at the `pycurl_manager` module. This [[1]](https://github.com/dmwm/WMCore/issues/11036) is the PR with which we have included this function. It serves to decompress zipped streams through HTTP, which are identified by proper negotiation between the server and the client about the data content compression through the HTTP headers. It looks like for some cases the content's compression has been wrongly set to `gzip` [2].  More on the way how the identification of the proper compression happens is discussed in the issue/PR above. One thing is suspicious in the log message for the current bug - at the end it says: 
```
Getting from Cache due to: CouchNotFoundError - reason: Object Not Found, data: {} result: None     
```
which may be a sign for reading from a cache file which is not having the proper compression. 

**How to reproduce it**
Not clear yet. But if it is related to reading gzip data from the cache it might be by trying to provide the proper header from the client and then try to read  the data from an old cache file, which is not having the contents compressed (Still to be confirmed).

**Expected behavior**

[1]
https://github.com/dmwm/WMCore/issues/11036

[2]

[05/Apr/2022:02:49:24]  Updating request "tivanov_ReReco_Parents_HG2204_Val_220401_123539_5532" with these user-provided args: {'total_jobs': 0, 'input_events': 0, 'input_lumis': 0, 'input_n
um_files': 0, 'RequestName': 'tivanov_ReReco_Parents_HG2204_Val_220401_123539_5532'}
ERROR:root:While processing decompress function with headers: {'Date': 'Tue, 05 Apr 2022 00:49:24 GMT', 'Server': 'CouchDB/1.6.1 (Erlang OTP/R16B03-1)', 'Set-Cookie': 'cms-node=624b9214001f3
6ffe1331bdb2ae875f748469a9f8732fb707323f8b02b6822441ae6cccea426fbbd05c2c42b;path=/;secure;httponly', 'ETag': '"W1AVV/xAOj0Xndl63XAMIA=="', 'Content-Type': 'application/json', 'Content-MD5': 
'W1AVV/xAOj0Xndl63XAMIA==', 'Content-Length': '17414', 'Content-Encoding': 'gzip', 'Cache-Control': 'must-revalidate', 'Accept-Ranges': 'none', 'CMS-Server-Time': 'D=12028 t=1649119764746333
'}, we were unable to decompress gzip content. Details: Not a gzipped file (b'cc')
  File "/data/srv/HG2204b/sw/slc7_amd64_gcc630/cms/reqmgr2/2.0.2.pre2/lib/python3.8/site-packages/WMCore/Services/pycurl_manager.py", line 80, in decompress
    return gzip.decompress(body)
  File "/data/srv/HG2204b/sw/slc7_amd64_gcc630/external/python3/3.8.2-comp/lib/python3.8/gzip.py", line 548, in decompress
    return f.read()
  File "/data/srv/HG2204b/sw/slc7_amd64_gcc630/external/python3/3.8.2-comp/lib/python3.8/gzip.py", line 292, in read
    return self._buffer.read(size)
  File "/data/srv/HG2204b/sw/slc7_amd64_gcc630/external/python3/3.8.2-comp/lib/python3.8/gzip.py", line 479, in read
    if not self._read_gzip_header():
  File "/data/srv/HG2204b/sw/slc7_amd64_gcc630/external/python3/3.8.2-comp/lib/python3.8/gzip.py", line 427, in _read_gzip_header
    raise BadGzipFile('Not a gzipped file (%r)' % magic)
gzip.BadGzipFile: Not a gzipped file (b'cc')
Getting from Cache due to: CouchNotFoundError - reason: Object Not Found, data: {} result: None

amaltaro:
Fixes #11066 

#### Status
ready

#### Description
If `Accept-Encoding` is provided in the HTTP headers, then set `pycurl.ACCEPT_ENCODING` with it such that libcurl can automatically decode the HTTP response body according to the `Content-Encoding` received in the headers. Note that we do not set `Accept-Encoding` again at the `pycurl.HTTPHEADER` (I did not test whether it's a problem or not).

Important information at this documentation (also linked in the pycurl source code):
https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

Apparently we are not able to do the same automatic procedure for multi curl, so that one still relies in the recently created `decompress` function.

UPDATE: we can set both `ACCEPT_ENCODING` and `HTTPHEADER` with the Accept-Encoding value and the automatic decoding of the response body remains functional. But given that `headers` could be a dictionary from upper classes, we better not change the original dictionary (so I am keeping the deepcopy around).

YES

#### Related PRs
Refactor some of the changes in: https://github.com/dmwm/WMCore/pull/11041

Specific dependencies on pycurl version and libcurl build, but it seems to be fine for our COMP setup.

### DIFF
--- a/src/python/WMCore/Services/pycurl_manager.py
+++ b/src/python/WMCore/Services/pycurl_manager.py
@@ -46,6 +46,7 @@ from future.utils import viewitems
 
 
 # system modules
+import copy
 import json
 import gzip
 import logging
@@ -81,7 +82,8 @@ def decompress(body, headers):
     except Exception as exc:
         logger = logging.getLogger()
         msg = "While processing decompress function with headers: %s, " % headers
-        msg += "we were unable to decompress gzip content. Details: %s" % str(exc)
+        msg += "we were unable to decompress gzip content. Details: %s. " % str(exc)
+        msg += "Considering response body as uncompressed."
         logger.exception(msg)
         return body
 
@@ -202,16 +204,20 @@ class RequestHandler(object):
         curl.setopt(pycurl.FOLLOWLOCATION, self.followlocation)
         curl.setopt(pycurl.MAXREDIRS, self.maxredirs)
 
-        # also accepts encoding/compression algorithms
-        if headers and headers.get("Accept-Encoding"):
-            if isinstance(headers["Accept-Encoding"], basestring):
-                curl.setopt(pycurl.ENCODING, headers["Accept-Encoding"])
+        # If ACCEPT_ENCODING is set to the encoding string, then libcurl
+        # will automatically decode the response object according to the
+        # Content-Enconding received
+        # More info: https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html
+        thisHeaders = copy.deepcopy(headers)
+        if thisHeaders and thisHeaders.get("Accept-Encoding"):
+            if isinstance(thisHeaders["Accept-Encoding"], basestring):
+                curl.setopt(pycurl.ACCEPT_ENCODING, thisHeaders.pop("Accept-Encoding"))
             else:
                 logging.warning("Wrong data type for header 'Accept-Encoding': %s",
-                                type(headers["Accept-Encoding"]))
+                                type(thisHeaders["Accept-Encoding"]))
         else:
             # add gzip encoding by default
-            curl.setopt(pycurl.ENCODING, 'gzip')
+            curl.setopt(pycurl.ACCEPT_ENCODING, 'gzip')
 
         if cookie and url in cookie:
             curl.setopt(pycurl.COOKIEFILE, cookie[url])
@@ -242,15 +248,15 @@ class RequestHandler(object):
 
         if verb in ('POST', 'PUT'):
             # only these methods (and PATCH) require this header
-            headers["Content-Length"] = str(len(encoded_data))
+            thisHeaders["Content-Length"] = str(len(encoded_data))
 
         # we must pass url as a string data-type, otherwise pycurl will fail with error
         # TypeError: invalid arguments to setopt
         # see https://curl.haxx.se/mail/curlpython-2007-07/0001.html
         curl.setopt(pycurl.URL, encodeUnicodeToBytes(url))
-        if headers:
+        if thisHeaders:
             curl.setopt(pycurl.HTTPHEADER, \
-                [encodeUnicodeToBytes("%s: %s" % (k, v)) for k, v in viewitems(headers)])
+                [encodeUnicodeToBytes("%s: %s" % (k, v)) for k, v in viewitems(thisHeaders)])
         bbuf = BytesIO()
         hbuf = BytesIO()
         curl.setopt(pycurl.WRITEFUNCTION, bbuf.write)
@@ -275,12 +281,11 @@ class RequestHandler(object):
         """Debug callback implementation"""
         print("debug(%d): %s" % (debug_type, debug_msg))
 
-    def parse_body(self, data, headers, decode=False):
+    def parse_body(self, data, decode=False):
         """
         Parse body part of URL request (by default use json).
         This method can be overwritten.
         """
-        data = decompress(data, headers)
         if decode:
             try:
                 res = json.loads(data)
@@ -316,7 +321,7 @@ class RequestHandler(object):
             if verb == 'HEAD':
                 data = ''
             else:
-                data = self.parse_body(bbuf.getvalue(), header.header, decode)
+                data = self.parse_body(bbuf.getvalue(), decode)
         else:
             data = bbuf.getvalue()
             data = decompress(data, header.header)


amaltaro:
Keeping in mind that I haven't fully tested this change and I only have ~15min of logs - with this patch applied - in my VM, the gzip errors I was seeing in the ReqMgr2 logs are now gone.

@todor-ivanov if it's sound to you, then please get it merged, cut a new WMCore release (`2.0.2.pre3`?) and make a new cmsdist PR to @muhammadimranfarooqi as the morning begins at CERN. Hopefully by noon we can then confirm whether it's fully functional or not and schedule the CMSWEB upgrade for the next day (Thursday).

@vkuznet even if it's merged when you get to it, please do let us know your thoughts.

src/python/WMCore/Services/pycurl_manager.py
```diff
@@ -202,16 +204,20 @@ def set_opts(self, curl, url, params, headers,
...
+            if isinstance(thisHeaders["Accept-Encoding"], basestring):
+                curl.setopt(pycurl.ACCEPT_ENCODING, thisHeaders.pop("Accept-Encoding"))
             else:
                 logging.warning("Wrong data type for header 'Accept-Encoding': %s",
-                                type(headers["Accept-Encoding"]))
+                                type(thisHeaders["Accept-Encoding"]))
         else:
             # add gzip encoding by default
-            curl.setopt(pycurl.ENCODING, 'gzip')
+            curl.setopt(pycurl.ACCEPT_ENCODING, 'gzip')
```
todor-ivanov, line 220: I would say it have been kind of a long shot to enforce a `gzip` encoding by default in the previous version. Using  `accept_enoding` as the default header and letting curl to decide I agree is the better way to go for the defaults.

todor-ivanov (approved):
Thanks [Alan](@amaltaro) for providing this fix.
This looks good to me. I know we are in a hurry to fix the release, but since this is a generic module used all over the place, I'd also wait [Valentin's](@vkuznet) opinion before we merge and proceed.

todor-ivanov:
> I know we are in a hurry to fix the release, but since this is a generic module used all over the place, I'd also wait [Valentin's](@vkuznet) opinion before we merge and proceed.

Sorry [Alan](@amaltaro), I have read your last comment upon sending my review. I will merge and release the tag now. [Valentin](@vkuznet) please add  all you think is also needed here and we will try to react ASAP and include everything in this fix.

src/python/WMCore/Services/pycurl_manager.py
vkuznet: I would agree, but honestly I was not aware that pycurl library is capable of dealing with encoding. If we knew this earlier I'll structure my initial gzip PR differently.  The changes seems correct to me.

vkuznet (commented):
I was not aware of pycurl feature to handle accept-encoding, otherwise I'll structure my gzip changes different. The changes seems fine to me but since you did not change multirequest function it will break again when this function will be called (since multirequest also calls set_opts where you added the changes). We need to make a second  review of the codebase and align it properly to use pycurl encoding feature.

todor-ivanov:
thanks  [Valentin](@vkuznet) 
> you did not change multirequest function it will break again when this function will be called

how do youthink would be the best action here, could you provide the exact line  of change or should we wait for [Alan](@amaltaro) to login for the day, or should I proceed with the change. In any case the sooner we have the additional fix the quicker I'll cut the next tag and we redeploy in testbed. 

Thanks

vkuznet:
I think now we need to remove this line: https://github.com/amaltaro/WMCore/blob/cc4546f8519448ca27d7242a839374275db6ccbc/src/python/WMCore/Services/pycurl_manager.py#L388 which calls decompress since you changed set_opt function to let pycurl to handle the encoding. As such I would expect that returned buffer will be already decoded. You may try to remove this line and test it when accepting gzip content.

amaltaro:
That's a good catch, Valentin. I am fixing it soon.

@todor-ivanov if the current fix is working fine in testbed, we can announce the upgrade for tomorrow. I just checked the code, and we do not use `multirequest` method anywhere in WMCore. So even if we get another tag in a couple of hours from now, we don't need to wait for it before we assess whether services are fully functional or not.

todor-ivanov:
thanks for checking  @amaltaro 
I am resubmitting some validation workflows and will just let them start progressing their status in reqmgr, and will forward the validation document to @muhammadimranfarooqi  so he can properly announce the deployment for tomorrow.

todor-ivanov:
I have double checked again with the workflows being progressed enough so they have passed through both reqmgr and workqueue at least and all seems to be fine. I am sending the validation document to Imran, now.

PR #11088, opened 2022-04-12
Enhance logic to map ScramArch to OS

### LINKED ISSUE #11081: Cleanup jobs requesting an empty OS at the job classad
**Impact of the bug**
WMAgent

**Describe the bug**
While validating these job wrapper changes:
https://github.com/dmwm/WMCore/pull/11077

I noticed that Cleanup jobs actually define an empty string for the required OS:
```
  CMS_JobType = "Cleanup"
  REQUIRED_OS = ""
```

**How to reproduce it**
Create a workflow and check the job classads for a Cleanup job.

**Expected behavior**
Cleanup jobs don't rely on any CMSSW release, so ScramArch is actually not defined for this step.
When converting the job/step requirements into job classads, we should make sure that `REQUIRED_OS` has a correct value, which is either `any` or the `rhelX` string.

None

amaltaro:
Fixes #11081 

#### Status
ready

#### Description
The previous logic was checking whether the first part of the ScramArch was `in` the ScramArch defined in the logic, which of course can fail if we test a substring like `"el8" in "el888_amd64_gcc630"`. Of course, it's very likely to become a problem, but I think it's worth changing it.

This PR changes it and now we check for equality.

YES

#### Related PRs
Meant to fix a typo in this PR: https://github.com/dmwm/WMCore/pull/11083, but now it became a minor refactoring.

None

### DIFF
--- a/src/python/WMCore/BossAir/Plugins/BasePlugin.py
+++ b/src/python/WMCore/BossAir/Plugins/BasePlugin.py
@@ -6,7 +6,7 @@ Base class for BossAir plugins
 """
 
 from builtins import object, str, bytes
-from future.utils import viewitems, viewvalues
+from future.utils import viewvalues
 
 from Utils.Utilities import decodeBytesToUnicode
 from WMCore.WMException import WMException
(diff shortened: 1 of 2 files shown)


amaltaro:
FYI @mapellidario Dario, if I am not wrong, you were working with this code (or the one in Scram.py).

src/python/WMCore/BossAir/Plugins/BasePlugin.py
```diff
@@ -130,27 +130,25 @@ def updateSiteInformation(self, jobs, siteName, excludeSite):
...
         """
+        Matches a ScramArch - or a list of it - against a map of Scram
+        to Operating System
 
-        Args:
-            scramArch: string or list of scramArches that are acceptable for the job
-
-        Returns:
-            string to be matched for OS requirements for job
+        :param scramArch: string of list of scramArches defined for a given job
```
smuzaffar, line 136: may be it should be "string or list of ...." :-)

smuzaffar:
looks good

amaltaro:
Thanks for your review, Shahzad.

PR #11096, opened 2022-04-13
Remove cPickle support

### LINKED ISSUE #7110: Remove try/except import cPickle
This is after we have comfortably migrated to python3. cPickle will no longer exist and there is no need to try to import it before falling back to pickle.

goughes:
Fixes #7110

#### Status
Not tested

#### Description
Since we no longer support python2, remove attempts to import cPickle and just import pickle.

Only works with python3!

#### Related PRs
NA

NA

### DIFF
--- a/src/python/WMComponent/JobCreator/JobCreatorPoller.py
+++ b/src/python/WMComponent/JobCreator/JobCreatorPoller.py
@@ -11,12 +11,7 @@ import logging
 import os
 import os.path
 import threading
-
-
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
+import pickle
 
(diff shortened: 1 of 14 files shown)


amaltaro (changes requested):
@goughes it looks good to me. However, can you also please remove the cpickle reference from this file:
src/python/WMCore/WMSpec/Makers/Handlers/MakeJobSlave.py
?

goughes:
Removed the cPickle reference from MakeJobSlave.py with my latest push. Thanks for catching that @amaltaro!

amaltaro (approved):
Thanks Erik! It looks good to go.

PR #10330, opened 2021-03-05
Forward all machine generated traffic to port 8443 for cmsweb services.

todor-ivanov:
Fixes #10119 

#### Status
needs some more testing

#### Description

In order to finalize the forwarding of the whole machine (NOT human) generated traffic with destination `https://cmsweb.*.cern.ch` to port 8443 we made an extensive investigation on the pieces of code and APIs left to still query the standard 443 port, which is now reserved for human generated traffic only. The results are commented in the issue itself, but here are the two major changes to WMCore/WMAgent related services that needs to be introduced with the current PR:
   * Because at some places we import Requests and `JSONRequests` classes directly but not through the `WMCore.Services.Service` we need  to add a check and a url mangling logic so that we can forward all traffic with destination cmsweb to `https://cmsweb.*.cern.ch:8443`   
   * Even though the CouchDB server port is set in the configuration to port  5984. We still see some queries asking for port 443 with source client `"CouchDB/1.6.1"`. These are the folwoing two APIs:
      * /couchdb/wmstats/...  
      * /couchdb/workqueue/...
They are coming from the following module  `WMCore.Database.CMSCouch` and whoever imports it. This is basically related to the logic used for automated database replication between the agents and the central services.  

#### Is it backward compatible (if not, which system it affects?)
YES

#### Related PRs
The relevant change to the top level class  `WMCore.Services.Service` is here:
https://github.com/dmwm/WMCore/pull/8726

#### External dependencies / deployment changes
**IMPORTANT**
It needs a change in the WMAgent secrets file. The change should look like [1]. And will take effect after agent deployment. It should result in the following configuration changes in the agent config itself [2].

[1]
```
$ diff -u /data/admin/wmagent/WMAgent.secrets /data/admin/wmagent/WMAgent.secrets~
--- /data/admin/wmagent/WMAgent.secrets	2021-03-19 06:53:46.743574324 +0100
+++ /data/admin/wmagent/WMAgent.secrets~	2021-03-19 06:51:33.385626269 +0100
@@ -7,8 +7,8 @@
-GLOBAL_WORKQUEUE_URL=https://cmsweb-testbed.cern.ch:8443/couchdb/workqueue
-WMSTATS_URL=https://cmsweb-testbed.cern.ch:8443/couchdb/wmstats
+GLOBAL_WORKQUEUE_URL=https://cmsweb-testbed.cern.ch/couchdb/workqueue
+WMSTATS_URL=https://cmsweb-testbed.cern.ch/couchdb/wmstats
```
[...]

PR #11099, opened 2022-04-15
speed up listDatasetFileDetails API

### LINKED ISSUE #11098: Improve speed of DBS3Reader APIs
**Impact of the new feature**
Current implementation of DBS3Reader API relies on sequential access of data from DBS. Even though it works just fine for heavy populated datasets/blocks it can be very slow especially when we require to fetch information from lots of blocks, e.g. parentage information. This ticket should address the speed of DBS3Reader APIs by refactoring codebase to take advantage of concurrency in API calls

For heavily populated datasets (with lots of blocks) I found that certain APIs, e.g. `listDatasetFileDetails` is very slow, see https://github.com/dmwm/dbs2go/issues/5

Refactor code to take advantage of concurrent (parallel) execution of APIs

**Additional context**
See this ticket: https://github.com/dmwm/dbs2go/issues/5

vkuznet:
Fixes #11098 

#### Status
ready

#### Description
Use concurrent features to speed up `listDatasetFileDetails` API. In my local setup I achieve speed up by factor of 5 or more for a specific dataset. Here are benchmark numbers for using `/VBF1Parked/Run2012D-v1/RAW` dataset. This dataset contains 594 blocks. The current codebase takes approximately 190 seconds to fetch its parents and file lumis. Using proposed solution I achieve the following numbers:
- 37 seconds using 50 concurrent tasks
- 41 seconds using 100 concurrent tasks
- 90 seconds using 10 concurrent tasks

Internally, I used `requests` python library instead of `pycurl_manager.py`. The latter is not suitable for concurrent execution since it is not thread safe as it holds global curl object. As such curl options are set on first tasks, but can't be changed (since code set them up) in others until first task is completed. The `requests` library does not depend on global curl object and curl library itself, and it is suitable for concurrent execution of multiple URL calls.

This PR only addresses speed up of single  `listDatasetFileDetails` DBS3Reader API but other APIs where multiple (sequential) calls to DBS are made can benefit from this approach.

YES

#### Related PRs
https://github.com/dmwm/dbs2go/issues/5

relies on Python `requests` library

### DIFF
--- a/src/python/WMCore/Services/DBS/DBS3Reader.py
+++ b/src/python/WMCore/Services/DBS/DBS3Reader.py
@@ -23,6 +23,8 @@ from retry import retry
 from Utils.IteratorTools import grouper
 from Utils.PythonVersion import PY2
 from WMCore.Services.DBS.DBSErrors import DBSReaderError, formatEx3
+from WMCore.Services.DBS.DBSUtils import dbsListFileParents, dbsListFileLumis, \
+    dbsBlockOrigin, dbsParentFilesGivenParentDataset
 
 
 ### Needed for the pycurl comment, leave it out for now
@@ -72,17 +74,28 @@ class DBS3Reader(object):
     General API for reading data from DBS
     """
 
-    def __init__(self, url, logger=None, **contact):
+    def __init__(self, url, logger=None, parallel=None, **contact):
+        """
+        DBS3Reader constructor
+
+        :param url: url of DBS server
+        :param logger: logger to be used by this class
+        :param parallel: optional parameter to specify parallel execution of some APIs
+        You may pass any true value, e.g. True or 1. The parallel APIs are:
+        listDatasetFileDetails, listFileBlockLocation, getParentFilesGivenParentDataset
+        :param contact: optional parameters to pass to DbsApi class
+        """
 
         # instantiate dbs api object
         try:
             self.dbsURL = url.replace("cmsweb.cern.ch", "cmsweb-prod.cern.ch")
             self.dbs = DbsApi(self.dbsURL, **contact)
             self.logger = logger or logging.getLogger(self.__class__.__name__)
+            self.parallel = parallel
         except Exception as ex:
             msg = "Error in DBSReader with DbsApi\n"
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
     def _getLumiList(self, blockName=None, lfns=None, validFileOnly=1):
         """
@@ -103,7 +116,7 @@ class DBS3Reader(object):
             msg = "Error in "
             msg += "DBSReader.listFileLumiArray(%s)\n" % lfns
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         lumiDict = {}
         for lumisItem in lumiLists:
@@ -128,7 +141,7 @@ class DBS3Reader(object):
             msg = "Error in "
             msg += "DBS server is not up: %s" % self.dbsURL
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
     def listPrimaryDatasets(self, match='*'):
         """
@@ -143,7 +156,7 @@ class DBS3Reader(object):
         except Exception as ex:
             msg = "Error in DBSReader.listPrimaryDataset(%s)\n" % match
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         result = [x['primary_ds_name'] for x in result]
         return result
@@ -160,7 +173,7 @@ class DBS3Reader(object):
         except dbsClientException as ex:
             msg = "Error in DBSReader.listProcessedDatasets(%s)\n" % primary
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         for dataset in datasets:
             dataset = remapDBS3Keys(dataset, processed_ds_name='Name')
@@ -195,7 +208,7 @@ class DBS3Reader(object):
         except dbsClientException as ex:
             msg = "Error in DBSReader.listRuns(%s, %s)\n" % (dataset, block)
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
         for x in results:
             runs.extend(x['run_num'])
         return runs
@@ -228,7 +241,7 @@ class DBS3Reader(object):
         except dbsClientException as ex:
             msg = "Error in DBSReader.listRuns(%s, %s)\n" % (dataset, block)
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         # send runDict format as result, this format is for sync with dbs2 call
         # which has {run_number: num_lumis} but dbs3 call doesn't return num Lumis
@@ -254,7 +267,7 @@ class DBS3Reader(object):
         except dbsClientException as ex:
             msg = "Error in DBSReader.listProcessedDatasets(%s)\n" % primary
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         result = [x['dataset'].split('/')[2] for x in result]
         return result
@@ -307,6 +320,25 @@ class DBS3Reader(object):
             files[f['logical_file_name']]['Lumis'] = {}
             files[f['logical_file_name']]['Parents'] = []
 
+        # parallel execution for listFileParents and listFileLumis APIs
+        if self.parallel:
+            if getParents:
+                block_parents = dbsListFileParents(self.dbsURL, blocks)
+                for blockName, parents in block_parents.items():
+                    for p in parents:
+                        if p['logical_file_name'] in files:  # invalid files are not there if validFileOnly=1
+                            files[p['logical_file_name']]['Parents'].extend(p['parent_logical_file_name'])
+            if getLumis:
+                block_file_lumis = dbsListFileLumis(self.dbsURL, blocks)
+                for blockName, file_lumis in block_file_lumis.items():
+                    for f in file_lumis:
+                        if f['logical_file_name'] in files:  # invalid files are not there if validFileOnly=1
+                            if f['run_num'] in files[f['logical_file_name']]['Lumis']:
+                                files[f['logical_file_name']]['Lumis'][f['run_num']].extend(f['lumi_section_num'])
+                            else:
+                                files[f['logical_file_name']]['Lumis'][f['run_num']] = f['lumi_section_num']
+            return files
+
         # Iterate over the blocks and get parents and lumis
         for blockName in blocks:
             # get the parents
@@ -344,7 +376,7 @@ class DBS3Reader(object):
         except Exception as exc:
             msg = "Error in DBSReader.crossCheck({}) with {} lfns.".format(datasetPath, len(lfns))
             msg += "\nDetails: {}\n".format(formatEx3(exc))
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
         setOfAllLfns = set(allLfns)
         setOfKnownLfns = set(lfns)
         return list(setOfAllLfns.intersection(setOfKnownLfns))
@@ -363,7 +395,7 @@ class DBS3Reader(object):
         except Exception as exc:
             msg = "Error in DBSReader.crossCheckMissing({}) with {} lfns.".format(datasetPath, len(lfns))
             msg += "\nDetails: {}\n".format(formatEx3(exc))
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
         setOfAllLfns = set(allLfns)
         setOfKnownLfns = set(lfns)
         knownFiles = setOfAllLfns.intersection(setOfKnownLfns)
@@ -384,7 +416,7 @@ class DBS3Reader(object):
         except Exception as ex:
             msg = "Error in DBSReader.getDBSSummaryInfo(%s, %s)\n" % (dataset, block)
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         if not summary:  # missing data or all files invalid
             return {}
@@ -411,7 +443,7 @@ class DBS3Reader(object):
         except dbsClientException as ex:
             msg = "Error in DBSReader.listFileBlocks(%s)\n" % dataset
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         result = [x['block_name'] for x in blocks]
 
@@ -435,7 +467,7 @@ class DBS3Reader(object):
             msg = "Error in "
             msg += "DBSReader.blockExists(%s)\n" % fileBlockName
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         if len(blocks) == 0:
             return False
@@ -454,7 +486,7 @@ class DBS3Reader(object):
         """
         if not self.blockExists(fileBlockName):
             msg = "DBSReader.listFilesInBlock(%s): No matching data"
-            raise DBSReaderError(msg % fileBlockName)
+            raise DBSReaderError(msg % fileBlockName) from None
 
         try:
             files = self.dbs.listFileArray(block_name=fileBlockName, validFileOnly=validFileOnly, detail=True)
@@ -462,7 +494,7 @@ class DBS3Reader(object):
             msg = "Error in "
             msg += "DBSReader.listFilesInBlock(%s)\n" % fileBlockName
             msg += "%s\n" % formatEx3(ex)
-            raise DBSReaderError(msg)
+            raise DBSReaderError(msg) from None
 
         if lumis:
             lumiDict = self._getLumiList(blockName=fileBlockName, validFileOnly=validFileOnly)
@@ -487,7 +519,7 @@ class DBS3Reader(object):
         """
         if not self.blockExists(fileBlockName):
             msg = "DBSReader.listFilesInBlockWithParents(%s): No matching data"
-            raise DBSReaderError(msg % fileBlockName)
+            raise DBSReaderError(msg % fileBlockName) from None
 
         try:
             # TODO: shoud we get only valid block for this?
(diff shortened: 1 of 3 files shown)


vkuznet:
@amaltaro , @todor-ivanov , @klannon this PR speed up single (most expensive) DBS3Reader API by factor of 5 or more. The code shows that concurrent execution of calls to upstream server can provide significant improvement. I suggest that you study this pattern which can be applied to different parts of WMCore codebase where multiple calls to upstream server (DBS, Rucio, etc.) will be required. Even thought it is not in Q2 plan I decided to put this forward as I obtained significant improvements can be made which entire system can benefit from and (more importantly) will allow to work with large datasets and avoid timeouts on CMSWEB.

vkuznet:
@amaltaro , on my second commit I got weird unit test failure which I'm 100% sure is not related to my changes but I want to point you to it since it may show how this unit test became unstable. Please see https://cmssdt.cern.ch/dmwm-jenkins/job/DMWM-WMCore-PR-test/13026/testReport/junit/WMCore_t.Services_t.UUID_t/UUIDTest/testUUID/ which shows:
```
'bcb3' == 'bcb3' : Second component of UUID the same bcb3 != bcb3
```
I think it is random error since UUID may be close enough during random process generation. Moreover, the message of test is wrong. I suggest that we address this unit issue separately. I'll trigger test again.

vkuznet:
test this please

amaltaro (commented):
Valentin, even though this looks like a great performance boost, I do not think this is the right implementation.

Why don't we use the pycurl_manager module for this? We already have some DBS concurrent APIs implemented in this module (inherited from your work porting Unified features into WMCore):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Tools/Common.py

and I think the best way forward would be to actually create a new module under:
WMCore/Services/DBS/youNameIt.py

and put this code in there, if possible relying on pycurl_manager only instead of `requests`.
If you see a need to use the straight forward `requests` library, then I think we could have a chat first before putting all this effort in.

Please let me know your thoughts (BTW, it's a holiday here, so you might hear back from me only on Monday).

vkuznet:
@amaltaro I explained in description that `pycurl_manager.py` can't be used in concurrent futures since it locks curl object. The only way to use `pycurl_manager` is to use `multi_getdata` and it is on my todo list.

vkuznet:
ok, I refactored the code using `pycurl_manager.py` `multi_getdata` and results are even more impressive. I achieve 10+ improvements for this API. Here is new structure:
- WMCore/Services/DBS/DBSUtils.py contains new common code
- WMCore/Services/DBS/DBS3Reader.py now can use either approach, tested with `requests` and `pycurl_manager.py` libraries
- I borrowed code from WMCore/MicroService/Tools/Common.py which seems to be common between these two code-bases, therefore I suggest to re-factor the code again to put this common code elsewhere. Please see appropriate comment in WMCore/Services/DBS/DBSUtils.py

Bottom line, I would like to keep around concurrent.features code since it demonstrates new paradigm on how concurrent programming should be done. This concept can be applied to different parts of WMCore which may benefit from it. The `requests` library used in original codebase is just an example and I suggest to keep it around, but it is not required now after my second implementation with `multi_getdata` from `pycurl_manager.py`

vkuznet:
New best value I achieved with `multi_getdata` is 16 seconds, which is almost factor of 12 better than current implementation based on sequential programming pattern.

vkuznet:
I re-factor the code and generalize it. The new API `dbsParallelApi` will allow to be used by different APIs in DBS3Reader (see full list in https://github.com/dmwm/WMCore/issues/11098). I added new attribute `parallel` into constructor of DBS3Reader such that class can execute either linear (sequential) or concurrent (parallel) DBS APis workflows. The code now can be further expanded to speed-up other APIs.

vkuznet:
I rebased this PR since we merged #11101

src/python/WMCore/Services/DBS/DBSUtils.py
```diff
@@ -0,0 +1,149 @@
...
+def dbsListFileParents(dbsUrl, blocks, logger):
+    """
+    Concurrent counter part of DBS listFileParents API
+
+    :param dbsUrl: DBS URL
+    :param blocks: list of blocks
+    :param logger: logger
+    :return: list of file parents
+    """
+    urls = ['%s/fileparents?block_name=%s' % (dbsUrl, b.replace('#', '%23')) for b in blocks]
```
amaltaro, line 28: Instead of replacing string, I'd suggest to stick to the standard way of encoding urls (based on urllib). In this case, you want to use `urllib.parse.quote_plus`.

src/python/WMCore/Services/DBS/DBSUtils.py
```diff
@@ -0,0 +1,149 @@
...
+def dbsListFileLumis(dbsUrl, blocks, logger=None):
+    """
+    Concurrent counter part of DBS listFileLumis API
+
+    :param dbsUrl: DBS URL
+    :param blocks: list of blocks
+    :param logger: logger
+    :return: list of file lumis
+    """
+    urls = ['%s/filelumis?block_name=%s' % (dbsUrl, b.replace('#', '%23')) for b in blocks]
```
amaltaro, line 43: Same comment as the one above.

src/python/WMCore/Services/DBS/DBSUtils.py
```diff
@@ -0,0 +1,149 @@
...
+def dbsBlockOrigin(dbsUrl, blocks, logger=None):
+    """
+    Concurrent counter part of DBS files API
+
+    :param dbsUrl: DBS URL
+    :param blocks: list of blocks
+    :param logger: logger
+    :return: list of block origins for a given parent lfns
+    """
+    urls = ['%s/blockorigin?block_name=%s' % (dbsUrl, b.replace('#', '%23')) for b in blocks]
```
amaltaro, line 58: Same as the comment above.

src/python/WMCore/Services/DBS/DBSUtils.py
```diff
@@ -0,0 +1,149 @@
...
+    :param logger: logger object
+    :return: list of parent files for given file info object
+    """
+    urls = []
+    for fileInfo in fInfo:
+        run = fileInfo['run_num']
+        lumis = urllib.parse.quote_plus(str(fileInfo['lumi_section_num']))
+        url = '%s/files?dataset=%s&run_num=%s&lumi_list=%s' \
+                % (dbsUrl, parentDataset, run, lumis)
+        if ':8443' not in url:
```
amaltaro, line 80: I might be wrong, but I think this switch has already been covered by the portForward decorator, no?

src/python/WMCore/Services/DBS/DBSUtils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    rdict = {}
+    for row in data:
+        url = row['url']
+        if uKey:
+            key = urlParams(url).get(uKey)
+        else:
+            key = url
+        code = int(row.get('code', 200))
+        if code != 200:
```
amaltaro, line 123: Shouldn't we consider fail http requests only if the status code is >= 400?

amaltaro (changes requested):
Valentin, this is looking pretty good in general.

However, I left some comments along the code and I would like to see unit tests created for some of those functions in the DBSUtils module. Those actually making calls to the DBS server might have to be skipped though (unless we mock things).

Valentin, once you update it, I will have a final and thorough review.

src/python/WMCore/Services/DBS/DBSUtils.py
vkuznet: that's fine

src/python/WMCore/Services/DBS/DBSUtils.py
vkuznet: this code does not use any decorators, and no, I checked, the port was not changed, that's why I added the change.

src/python/WMCore/Services/DBS/DBSUtils.py
vkuznet: this is an edge use-case since non 200 codes does not return any data and may return plain string, e.g. the code 301 will return "Moved permanently", and code below explicitly extracts the `data` which is constructed by `multi_getdata` function from `pycurl_manager`. I would leave it as is because it explicitly will show non-data message.

vkuznet:
@amaltaro I added one _**integration**_ test to test functionality of `listDatasetFileDetails` DBS API through sequential and parallel execution. I found small dataset (in terms of blocks/files) in production DBS DB and used for the test. Please review and let me know if it is appropriate.

src/python/WMCore/Services/DBS/DBSUtils.py
amaltaro: Then I'd suggest to use the exact same mechanism that Todor implemented in the last year, as you can see here:
https://github.com/dmwm/WMCore/pull/10330/files

in short:
```
portForwarder = PortForward(8443)
url = portForwarder(url)
```

src/python/WMCore/Services/DBS/DBSUtils.py
amaltaro: If my memory doesn't fail me, I've seen cases where we get successfully get actual data and the return code is != 200. Perhaps we should then consider at least codes 200-200 as successful? From:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

src/python/WMCore/Services/DBS/DBSUtils.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+_DBSUtils_
+
+set of common utilities for DBS3Reader
+
+"""
+import json
+import urllib
+from urllib.parse import urlparse
+from urllib.parse import parse_qs, quote_plus
```
amaltaro, line 11: These two imports should be grouped.

test/python/WMCore_t/Services_t/DBS_t/DBSUtils_t.py
```diff
@@ -0,0 +1,67 @@
...
+class DBSUtilsTest(unittest.TestCase):
+
+    def testUrlParams(self):
+        """
+        urlParams should return dictionary of URL parameters
+        """
+        url = 'http://a.b.com?d=1&f=bla'
+        results = urlParams(url)
+        self.assertTrue(results.get('d'), 1)
+        self.assertTrue(results.get('f'), 'bla')
```
amaltaro, line 24: Maybe testing `results` with:
```
self.assertCountEqual(results, {'d': ['1'], 'f': ['bla']})
```
would be cleaner and better?

Can you please also add a test case for an url with multiple values for the same query string (e.g. d=1&d=2&f=bla)?

test/python/WMCore_t/Services_t/DBS_t/DBSUtils_t.py
```diff
@@ -0,0 +1,67 @@
...
+        return
+
+
+    def testGetParallel(self):
```
amaltaro, line 28: It's great to have a test like this, even if it's not run in jenkins. At least we can make sure our code is sound!

To avoid triggering this by every single jenkins run, can you please:
```
from nose.plugins.attrib import attr
```

and decorate this test with:
```
@attr("integration")
```
?

amaltaro (changes requested):
Thank you for these changes, Valentin. I have some extra comments for your consideration.

src/python/WMCore/Services/DBS/DBSUtils.py
vkuznet: fine

test/python/WMCore_t/Services_t/DBS_t/DBSUtils_t.py
vkuznet: ok

test/python/WMCore_t/Services_t/DBS_t/DBSUtils_t.py
vkuznet: done

vkuznet:
@amaltaro , I adjusted all comments you made except codes. The codes 200-300 have special meaning, e.g. 201 Created  is used to create resource, and we are not creating anything, 202 Accepted The request has been received but not yet acted upon, so when we query the data we should receive the data, etc. Please do not mix codes with actual behavior. Each code mans something and our action ONLY to receive the data which falls into code 200. All other codes are not appropriate for our action. That's why code should not use them as correct condition.

vkuznet:
@amaltaro , @todor-ivanov what is holding us here? I addressed all issues you asked for, and already other PR changed DBS3Reader that I was need to resolve conflict.

amaltaro (changes requested):
Valentin, you can find a couple of comments along the code.

In addition to those, I have some other remarks to make:
* maybe we should rename the DBSUtil module to something more meaningful, e.g.: DBSParallel, or DBSPycurl, or DBSConc? Just to give a clear distinction that its underlying library is different than DBS3Reader
* logger parameter in those DBSUtil functions is not consistent. Sometimes it's mandatory, others it's optional. From a different angle, maybe it could be resolved by converting those functions into methods of a class, defining the logger object only once when it gets instantiated. Just an idea though...
* the DBSUtil functions are not consistent as well in the way they got developed. For instance, `dbsListFileParents` and other 2 only return data, while `dbsParentFilesGivenParentDataset` replicates the whole logic from DBS3Reader. I would stick to one model only, so either we (re-)implement everything in the DBSUtil function, or we only retrieve data and send it not processed back to DBS3Reader
* these changes are not compliant with the guidelines. New modules should not have any pylint/pep8/pycodestyle issues, unless there is no way around of it (as mentioned [here](https://github.com/dmwm/WMCore/blob/master/CONTRIBUTING.rst#automatic-tests))
* there is a mix of variable names as well (included in this PR, what was there already does not need to be touched), e.g. `block_parents`.

It's good that Erik is working on the pep8/pycodestyle, so we can at least get these reports now and fix it before human review have to spend time on it.

test/python/WMCore_t/Services_t/DBS_t/DBSUtils_t.py
```diff
@@ -0,0 +1,75 @@
...
+			"num_lumi": 428
+		  }
+		]
+
+		The parallel call should perform better then sequential while results
+        should remain the same.
+
+        """
+        url = "https://cmsweb.cern.ch/dbs/prod/global/DBSReader"
+        reader1 = DBS3Reader(url, logger=None, parallel=False, aggregate=True)
```
amaltaro, line 61: Valentin, I like how you made this integration test. Can you please do the same for the other APIs that you offer a concurrent option (please keep them marked as integration, it's really only to make sure all these changes are providing the same result and to help testing whenever we need to change it again).

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -307,6 +310,25 @@ def listDatasetFileDetails(self, datasetPath, getParents=False, getLumis=True, v
             files[f['logical_file_name']]['Lumis'] = {}
             files[f['logical_file_name']]['Parents'] = []
 
+        # parallel execution for listFileParents and listFileLumis APIs
+        if self.parallel:
```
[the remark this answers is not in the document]
vkuznet: Your changes do two things:
- they modify existing code base, since you add new loop to create block_parents dict
- they create new dict (which adds to the memory consumption)

In my version, I did not change existing codebase.

It is a choice and I don't have strong preference in one way or another. I leave it up to you to decide. Bottom line, the concurrent/parallel construct in a code quite often require new redesign and it is hard to align them with sequential programming patterns.

vkuznet:
Alan, let me answer your questions in order they appear. For clarify I'll repeat the list:
- maybe we should rename the DBSUtil module to something more meaningful, e.g.: DBSParallel, or DBSPycurl, or DBSConc? Just to give a clear distinction that its underlying library is different than DBS3Reader
   - VK: Please be specific, I choose name using my view, if you want to name it differently please specify how, I should not guess which name out of your suggestion is more appropriate
- logger parameter in those DBSUtil functions is not consistent. Sometimes it's mandatory, others it's optional. From a different angle, maybe it could be resolved by converting those functions into methods of a class, defining the logger object only once when it gets instantiated. Just an idea though...
   - VK: indeed, logger parameter seems inconsistent, I only added based on existing DBS3Reader code. But after your review I decided to completely remove it as it is not used by parallel code at all
- the DBSUtil functions are not consistent as well in the way they got developed. For instance, dbsListFileParents and other 2 only return data, while dbsParentFilesGivenParentDataset replicates the whole logic from DBS3Reader. I would stick to one model only, so either we (re-)implement everything in the DBSUtil function, or we only retrieve data and send it not processed back to DBS3Reader
   - VK: this is side-effect of code review, originally I placed code within DBS3Reader which then you requested to put into new module. As such code was copied to new module and inherited original naming convention used in DBS3Reader
- these changes are not compliant with the guidelines. New modules should not have any pylint/pep8/pycodestyle issues, unless there is no way around of it (as mentioned [here](https://github.com/dmwm/WMCore/blob/master/CONTRIBUTING.rst#automatic-tests))
   - VK: fair enough, will improve
- there is a mix of variable names as well (included in this PR, what was there already does not need to be touched), e.g. block_parents.
   - VK: I made changes consistent with names used by DBS3Reader. I don't know what is better to have consistency with code surrounding my changes or be strict with guidelines to make new code. In latter case, new code will be inconsistent with existing code. I am open to suggestion here. If you insist to make new code based on guidelines I'll adjust new variables but at the end we'll end-up with a mix of naming conventions. Even your proposal to adjust code violates naming conventions. Bottom line, guidelines are not a panacea to all cases.

vkuznet:
test this please

src/python/WMCore/Services/DBS/DBS3Reader.py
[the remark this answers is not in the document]
amaltaro: The memory concern is a good point. However, if we use the concurrent version, then we will end up with that large dictionary anyways.

As anything, there are pros and cons and it's hard to decide whether a) making a transparent change now is better than b) making the code maintenance easier for future changes. Given that you have already coded and tested it, I'd say we keep the current implementation.

vkuznet:
@amaltaro , I made necessary changes and replied to all your comments (some of them were unclear and I left my questions). Meanwhile, I run code through `autopep8`, and verified that Jenkins reports 10 score for new module. I also complement the code with integration tests. Please review again.

src/python/WMCore/Services/DBS/DBS3Reader.py
```diff
@@ -72,17 +74,18 @@ class DBS3Reader(object):
     General API for reading data from DBS
     """
 
-    def __init__(self, url, logger=None, **contact):
+    def __init__(self, url, logger=None, parallel=None, **contact):
```
amaltaro, line 77: Can you please create a docstring for this class and specify what `parallel` is for?

amaltaro (approved):
Valentin, this looks good to me. I'd suggest to add the docstring I mentioned in the code though.

In addition to that, please:
1) next time, please keep in mind that test/* changes should NOT go together with the source code changes, i.e., they must be provided in different commits;
2) please squash these commits in a single one (if test was separated, we should squash them in 2 commits instead).

Thanks

src/python/WMCore/Services/DBS/DBS3Reader.py
vkuznet: done

vkuznet:
@amaltaro now you have it squashed and rebased.

amaltaro:
test this please

amaltaro (approved):

todor-ivanov (commented):
looks good to me

PR #11101, opened 2022-04-18
move ckey/cert functions to Utils.CertTools

### LINKED ISSUE #11098: Improve speed of DBS3Reader APIs

vkuznet:
Partially Fixes #11098 (moving cert/ceky function will allow to share code between MicroServices and DBS services)

#### Status
ready

#### Description
Move functions `ckey` and `cert` into `Utils.CertTools` which is generic place to import from other places. This will allow to share and use this code in different Services, like MicroServices and DBS.

YES

#### Related PRs
See #11099 , if applied then I can safely remove these functions from #11099  and use common code.

### DIFF
--- a/src/python/Utils/CertTools.py
+++ b/src/python/Utils/CertTools.py
@@ -3,6 +3,18 @@ from builtins import str
 import os
 
 
+def ckey():
+    "Return user CA key either from proxy or userkey.pem"
+    pair = getKeyCertFromEnv()
+    return pair[0]
+
+
+def cert():
+    "Return user CA cert either from proxy or usercert.pem"
+    pair = getKeyCertFromEnv()
+    return pair[1]
+
+
 def getKeyCertFromEnv():
     """
     gets key and certificate from environment variables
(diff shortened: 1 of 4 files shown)


vkuznet:
@amaltaro this PR is ready for review. One note though. The py3k lint checks fail with [round built-in referenced]() which is explained here https://stackoverflow.com/questions/59533912/what-is-the-alternative-to-built-in-round. I think we should not degrade about py2 compatibility and the [pylint documentation](https://pycodequ.al/docs/pylint-messages/w1633-round-builtin.html) suggests that it is only for backward compatibility, and therefore it should be ignored or it is better to adjust pylint configuration to skip this test.

todor-ivanov (approved):
Thanks @vkuznet 
The code looks good to me. 
And also thanks for fixing two of the old standing pylint warnings.

amaltaro (changes requested):
This looks good to me, Valentin. However, given you are already on this, could you please update the same in this module:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/MicroService/Tools/PycurlRucio.py#L32-L38
?

Keep in mind there is an import that we can likely remove as well.

vkuznet:
Alan, I corrected `PycurlRucio.py` as well. Please review again.

amaltaro (approved):
Thanks Valentin, it looks good to me. Regarding pylint 3k, Erik and I had a brief chat and we also think it could be removed (AFAIR it checks whether our py2 code is compatible with py3, in some sense).

PR #11119, opened 2022-04-28
Properly handle OpenRunningTimeout in WorkQueue

### LINKED ISSUE #11048: Run3 Nano production while data taking
**Impact of the new feature**
WMAgent, ReqMgr2(?), WorkQueue(?)

T0 is not able generate NanoAOD as part of PromptReco workflows. Adding that feature would involve several time consuming developments. Instead of that, we are considering the possibility of having production workflows that generate NanoAOD during data taking.

T0 creates MiniAOD datasets and continually adds blocks to them during data taking. We want to have running production workflows that will create NanoAOD out of the growing MiniAOD datasets. A more detailed description of the feature can be found in Jira: [CMSTZ-969](https://its.cern.ch/jira/browse/CMSTZ-969)

The alternative would be to modify PromptReco spec to generate the NanoAOD in T0. This has the complication that T0 currently cannot handle flat N-tuples files.

**Additional context**
CMS-T0 Jira discussion: [CMSTZ-969](https://its.cern.ch/jira/browse/CMSTZ-969)

amaltaro:
Fixes #11048 

#### Status
ready

#### Description
This pull request provides all the pieces to support OpenRunningTimeout workflows in GQ, allowing freshly created input blocks to be created as workqueue elements in GQ.

A detailed description of the changes provided here is:
* overall improvements to log records
* default GQ inbox elements (1 per workflow) to be OpenForNewData=True
* default workqueue elements `CreationTime` and `TimestampFoundNewData` to use the current time (instead of 0)
* encapsulates work cancelation in try/except (since there is another thread doing the cleanup)
* `closeWork` method refactored to only close inbox work (now used only in GQ), according to:
  * if inbox element has OpenRunningTimeout=0, it gets closed
  * if it's been more than X secs since the last element has been discovered, it gets closed too
  * no longer takes a specific workflow as input, but lists all workflows open for new data and verify each one of them
  * no longer look for open blocks in DBS
* `addWork` method refactored to add new elements to workflows with workqueue inbox opened for new data (only used by GQ)

In the module WorkQueue/WorkQueueReqMgrInterface.py, changes are:
* reorganized tasks executed in the __call__ method, such that:
  * first it cancels elements for aborted/force-completed workflows
  * secondly, it closes requests with open inbox elements (if constraints are fulfilled)
  * third, it tries to add new work to open requests
  * fourth, it splits/queues work for new requests (workflows sitting in staged status)
* removed `report` method, which was already no longer used (since we started making status transition only in the ReqMgr2 cherrypy thread)

TODO: fix input data placement in MSTransferor such that the whole container has a rule. But there is a separate issue for that.

YES

#### Related PRs
None

None

### DIFF
--- a/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
+++ b/src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
@@ -5,7 +5,7 @@ A dictionary based object meant to represent a WorkQueue element
 """
 
 from future.utils import viewitems, viewvalues
-
+import time
 from hashlib import md5
 
 from Utils.Utilities import encodeUnicodeToBytesConditional
@@ -98,16 +98,16 @@ class WorkQueueElement(dict):
         self.setdefault('NumOfFilesAdded', 0)
         # Mask used to constrain MC run/lumi ranges
         self.setdefault('Mask', None)
-        # is new data being added to the inputs i.e. open block with new files or dataset with new closed blocks?
+        # is new data being added to the workflow (e.g., growing input datasets)
         self.setdefault('OpenForNewData', False)
-        # When was the last time we found new data (not the same as when new data was split), e.g. An open block was found
-        self.setdefault('TimestampFoundNewData', 0)
+        # EPOCH time for when the last block was found and added to the workflow
+        self.setdefault('TimestampFoundNewData', int(time.time()))
         # Trust initial input dataset location only or not
         self.setdefault('NoInputUpdate', False)
         # Trust initial pileup dataset location only or not
         self.setdefault('NoPileupUpdate', False)
-        # set the creation time for wq element, need for sorting
-        self.setdefault('CreationTime', 0)
+        # EPOCH time with the creation time for this wq element, needed for sorting
+        self.setdefault('CreationTime', int(time.time()))
         # set to true when updated from a WorkQueueElementResult
         self.modified = False
 
--- a/src/python/WMCore/WorkQueue/WorkQueue.py
+++ b/src/python/WMCore/WorkQueue/WorkQueue.py
@@ -607,6 +607,7 @@ class WorkQueue(WorkQueueBase):
             else:
                 self.logger.debug('Not deleting "%s" as it is %s', request.id, request['Status'])
 
+    # NOTE: this function is not executed by local workqueue
     def queueWork(self, wmspecUrl, request=None, team=None):
         """
         Take and queue work from a WMSpec.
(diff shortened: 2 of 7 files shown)


amaltaro:
And this is looking good enough to request a review, even though real tests are still running. Please read carefully the initial description, which should make this review not terribly hard.

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
```diff
@@ -101,13 +101,13 @@ def __init__(self, **kwargs):
         # is new data being added to the inputs i.e. open block with new files or dataset with new closed blocks?
         self.setdefault('OpenForNewData', False)
         # When was the last time we found new data (not the same as when new data was split), e.g. An open block was found
-        self.setdefault('TimestampFoundNewData', 0)
+        self.setdefault('TimestampFoundNewData', int(time.time()))
```
vkuznet, line 104: From the change it is unclear what is convention about stored timestamps, do you intend to store local time or stick to UTC. If former, the change is fine, but if you'd like to have common time zone for all deployed instances then UTC will be much better choice.

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
```diff
@@ -101,13 +101,13 @@ def __init__(self, **kwargs):
...
         # Trust initial input dataset location only or not
         self.setdefault('NoInputUpdate', False)
         # Trust initial pileup dataset location only or not
         self.setdefault('NoPileupUpdate', False)
         # set the creation time for wq element, need for sorting
-        self.setdefault('CreationTime', 0)
+        self.setdefault('CreationTime', int(time.time()))
```
vkuznet, line 110: again, local or UTC?

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -640,23 +641,61 @@ def queueWork(self, wmspecUrl, request=None, team=None):
         work = self.processInboundWork(inbound, throw=True)
         return len(work)
 
-    def addWork(self, requestName, rucioObj=None):
+    def addWork(self, inboundElem, rucioObj=None):
         """
         Check and add new elements to an existing running request,
         if supported by the start policy.
+
```
vkuznet, line 648: please add reference where to find inboundElem schema. The code in this method uses different attributes and it is unclear which schema for this data structure is used.

src/python/WMCore/WorkQueue/WorkQueueReqMgrInterface.py
```diff
@@ -48,48 +47,53 @@ def __init__(self, **kwargs):
...
         except Exception as ex:
-            errorMsg = "Error caught during RequestManager pull"
-            self.logger.exception("%s: %s", errorMsg, str(ex))
+            self.logger.exception("Generic error while canceling work. Details: %s", str(ex))
 
-        try:  # get additional open-running work
-            self.logger.info("adding new element to open requests")
+        try:  # Close requests that no longer need to be open for new data
+            self.logger.info("Closing open requests")
+            workClosed = queue.closeWork()
```
vkuznet, line 62: what will happen if this indeed will throw an exception. Does the queue will have orphan requests? How takes care of that? In other words how queue is managed. From the code changes it is unclear since it is passed as an input.

vkuznet (commented):
I am now familiar with entire logic of WQ and only paid attention to the code changes.

src/python/WMCore/WorkQueue/DataStructs/WorkQueueElement.py
amaltaro: This is EPOCH time, so it should be a reference to "January 1, 1970, 00:00:00 (UTC)". I can add a comment in the line above though. Same reply for the question.

goughes:
test this please

amaltaro:
test this please

amaltaro:
There seems to be an issue/intervention with the Rucio Integration server, hence all these failures. I will test it again later in the day.

amaltaro:
test this please

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -818,95 +858,60 @@ def pullWork(self, resources=None):
...
-                    # if it is successful remove previous error
-                    self.logdb.delete(element.id, "error", this_thread=True)
-                else:
-                    msg = "ChildElement is empty for element id %s: investigate" % element.id
-                    self.logdb.post(element.id, msg, "error")
-                    # self.logdb.upload2central(element.id)
-                    self.logger.error(msg)
+            # Check if the delay has passed
+            foundNewDataTime = element.get('TimestampFoundNewData', 0)
+            if (currentTime - foundNewDataTime) > openRunningTimeout:
```
vkuznet, line 895: this block requires further clarification. For instance:
- if `openRunningTimeout ` is zero
- then, the condition will be always meet
- and, the same `element.id` will be inserted to `workflowsToClose`

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -818,95 +858,60 @@ def pullWork(self, resources=None):
...
+                self.logger.info("Workflow %s has expired OpenRunningTimeout. Queuing to be closed.",
+                                 element['RequestName'])
+                workflowsToClose.append(element.id)
```
vkuznet, line 898: should you check that given `element.id` is not already in `workflowsToClose`?

src/python/WMCore/WorkQueue/WorkQueue.py
amaltaro: Good point! I should code this with an if-elseif instead, to make sure only one condition is met.
I noticed I made a mistake in this line as well:
https://github.com/dmwm/WMCore/pull/11119/files#diff-c7498b20d45ccef9c8d6c59b3107f8e58c93d01e34decfd2b1fa4fec62ec915cR886

I should be getting that key from `element`, not `policy`.

src/python/WMCore/WorkQueue/WorkQueue.py
amaltaro: Each workflow has only one inbox element, so it's pretty safe. Unless we make a mistake and consider the same workflow multiple times, as I have in the current code. Updating it. Thanks for spotting it.

amaltaro:
@vkuznet my last commit has a fix to the code that we were discussing in your last comments.

vkuznet (approved):

amaltaro:

src/python/WMCore/WorkQueue/WorkQueue.py
```diff
@@ -818,95 +858,57 @@ def pullWork(self, resources=None):
...
-                    self.logdb.post(element.id, msg, "error")
-                    # self.logdb.upload2central(element.id)
-                    self.logger.error(msg)
+        workflowsToCheck = self.backend.getInboxElements(OpenForNewData=True)
+        self.logger.info("Retrieved a list of %d open workflows", len(workflowsToCheck))
+        workflowsToClose = []
+        currentTime = time.time()
+        for element in workflowsToCheck:
+            # fetch attributes from the inbox workqueue element
+            openRunningTimeout = element.get('OpenRunningTimeout', 0)
```
amaltaro, line 885: This last minute change actually broke this functionality. The WorkQueue element data structure does not have this parameter... 
I will get it fixed in: https://github.com/dmwm/WMCore/pull/11135

amaltaro:
It turns out this was only supported for ReReco specs, given that we do not pass `OpenRunningTimeout` to the method `setWorkQueueSplitPolicy` in specs different than ReReco.

I will address it with https://github.com/dmwm/WMCore/pull/11135

